PR #1337, opened 2017-03-24
make docs more clear about how to work with the current api

SantoDE:
cc @containous/traefik :)

### DIFF
--- a/traefik.sample.toml
+++ b/traefik.sample.toml
@@ -1006,43 +1006,43 @@
 #
 # Optional
 #
-#[rancher]
+# [rancher]
 
 # Default domain used.
 # Can be overridden by setting the "traefik.domain" label on an service.
 #
 # Required
 #
-#domain = "rancher.localhost"
+# domain = "rancher.localhost"
 
 # Enable watch Rancher changes
 #
 # Optional
 # Default: true
 #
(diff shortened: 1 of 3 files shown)


timoreimann:
Should we update the sample file too?

emilevauge:
Ping @SantoDE ;)

ldez:
@SantoDE Do you have time to update the sample file?

SantoDE:
Hey @timoreimann & @ldez ,

done ;-)

ldez:
@SantoDE  can you change the PR from v1.2 to master

SantoDE:
@ldez done :)

timoreimann:
Update of `traefik.sample.toml` still outstanding?

SantoDE:
Ah, yeah! :D Sorry. we have to many examples 👼 

done @timoreimann

traefik.sample.toml
```diff
@@ -996,43 +996,43 @@
 #
 # Optional
 #
-#[rancher]
```
timoreimann: I think we want to leave all ranger configuration parameters commented out, as with the other providers in the sample file.

AFAIU, the idea is that users copy the file and uncomment just what they need.

See also #1455.

PR #1320, opened 2017-03-20
Make Traefik health checks label-configurable with Marathon.

timoreimann:
For the two existing health check parameters (path [formerly known as URL] and interval), we add support for Marathon labels.

Changes in detail:

- Extend the Marathon provider and template.
- Refactor Server.loadConfig to reduce duplication.
- Refactor the healthcheck package slightly to accommodate the changes and allow extending by future parameters.
- Update documentation.

Fixes #1232.

PR #1338, opened 2017-03-24
Add global health check interval parameter.

timoreimann:
The new parameter allows to set a health check interval valid for all backends. Custom values set per provider may override the global one.

@containous/traefik Please review.

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -269,6 +269,27 @@ Supported filters:
 # attempts = 3
 ```
 
+## Health check configuration
+```toml
+# Enable custom health check options.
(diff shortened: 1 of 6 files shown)


timoreimann:
The PR currently builds on #1320. Will be rebased once downstream gets merged.

timoreimann:
This isn't working yet because `time.Duration` values are not parsed correctly. We require #1342.

timoreimann:
Everything's working as expected now.

ldez:
@timoreimann Need rebase 😉

PR #1339, opened 2017-03-24
Fix regex with PathStrip

### LINKED ISSUE #928: Regex not working with PathStrip
Hi,

The doc states that PathStrip works like Path but strips matched path.

But actually Path does work with regex where PathStrip **doesn't**:

For example: 
```
docker run -d -p 8085:80 --name prefix --label "traefik.enable"="true" --label "traefik.frontend.rule"='Path: /configuration/{category}/{id:[0-9]+}' tutum/hello-world

curl 127.0.0.1/configuration/test/123
<html>
<head>
        <title>Hello world!</title>
        <link href='http://fonts.googleapis.com/css?family=Open+Sans:400,700' rel='stylesheet' type='text/css'>
        <style>
        body {
                background-color: white;
                text-align: center;
                padding: 50px;
                font-family: "Open Sans","Helvetica Neue",Helvetica,Arial,sans-serif;
        }

        #logo {
                margin-bottom: 40px;
        }
        </style>
</head>
<body>
        <img id="logo" src="logo.png" />
        <h1>Hello world!</h1>
        <h3>My hostname is 18ba062df939</h3>    </body>
</html>

docker rm -f prefix
```

```
docker run -d -p 8086:80 --name prefixstrip --label "traefik.enable"="true" --label "traefik.frontend.rule"='PathStrip: /configuration/{category}/{id:[0-9]+}' tutum/hello-world

curl 127.0.0.1/configuration/test/123
404 page not found

docker rm -f prefixstrip
```

seguins:
Fixes #928

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -104,9 +104,11 @@ Following is the list of existing matcher rules along with examples:
 - `HostRegexp: traefik.io, {subdomain:[a-z]+}.traefik.io`: Match request host. It accepts a sequence of literal and regular expression hosts.
 - `Method: GET, POST, PUT`: Match request HTTP method. It accepts a sequence of HTTP methods.
 - `Path: /products/, /articles/{category}/{id:[0-9]+}`: Match exact request path. It accepts a sequence of literal and regular expression paths.
-- `PathStrip: /products/, /articles/{category}/{id:[0-9]+}`: Match exact path and strip off the path prior to forwarding the request to the backend. It accepts a sequence of literal and regular expression paths.
+- `PathStrip: /products/`: Match exact path and strip off the path prior to forwarding the request to the backend. It accepts a sequence of literal paths.
+- `PathStripRegex: /articles/{category}/{id:[0-9]+}`: Match exact path and strip off the path prior to forwarding the request to the backend. It accepts a sequence of literal and regular expression paths.
 - `PathPrefix: /products/, /articles/{category}/{id:[0-9]+}`: Match request prefix path. It accepts a sequence of literal and regular expression prefix paths.
-- `PathPrefixStrip: /products/, /articles/{category}/{id:[0-9]+}`: Match request prefix path and strip off the path prefix prior to forwarding the request to the backend. It accepts a sequence of literal and regular expression prefix paths. Starting with Traefik 1.3, the stripped prefix path will be available in the `X-Forwarded-Prefix` header.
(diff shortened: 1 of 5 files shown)


emilevauge:
@seguins thanks for your contribution. I think we should avoid adding a dependency to `github.com/containous/mux` in `middlewares/`. Do you think you could add some interface tricks into this?

seguins:
I think we can create an interface for the matching, and move the `github.com/containous/mux` part to the `main` package: 
```go
type PrefixMatcher interface {
	Match(r *http.Request) (bool, string)
}
```

<details>

The `stripPrefix` file cloud be: 

```go
package middlewares

import (
	"net/http"
)

type PrefixMatcher interface {
	Match(r *http.Request) (bool, string)
}

// StripPrefix is a middleware used to strip prefix from an URL request
type StripPrefix struct {
	Handler http.Handler
	matcher PrefixMatcher
}

// NewStripPrefix builds a new StripPrefix given a handler and prefixes
func NewStripPrefix(handler http.Handler, matcher PrefixMatcher) *StripPrefix {
	stripPrefix := StripPrefix{Handler: handler, matcher: matcher}
	return &stripPrefix
}

func (s *StripPrefix) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	if match, prefix := s.matcher.Match(r); match == true {
		r.URL.Path = r.URL.Path[len(prefix):]
		r.RequestURI = r.URL.RequestURI()
		s.Handler.ServeHTTP(w, r)
		return
	}
	http.NotFound(w, r)
}

// SetHandler sets handler
func (s *StripPrefix) SetHandler(Handler http.Handler) {
	s.Handler = Handler
}
```

</details>

---

For you information, I have checked other files in middlewares, and I found that [handlerSwitcher.go](https://github.com/containous/traefik/blob/08c1871c98fa86e76d5a2e07e550cc1720a02445/middlewares/handlerSwitcher.go#L6) and [routes.go](https://github.com/containous/traefik/blob/08c1871c98fa86e76d5a2e07e550cc1720a02445/middlewares/routes.go#L8) already use `github.com/containous/mux`.

@emilevauge do you think we should do this change ?

emilevauge:
@seguins 
You right, let's keep the dep to `github.com/containous/mux`.
Would you consider creating a `PathStripRegex` instead of reusing `PathStrip` ? I fear that adding regex check in `PathStrip` will slow down the process for all other requests.

seguins:
@emilevauge, I've pushed a version with the regex matching separated.

ldez:
Great job @seguins! Could you fix this https://travis-ci.org/containous/traefik/jobs/226814163#L827 ? Thanks.

seguins:
Fixed 😃

Issue #1322, opened 2017-03-21
accept4: too many open files


IncSW:
### What version of Traefik are you using (`traefik version`)?
```
Version:      v1.2.0-rc2
Codename:     morbier
Go version:   go1.7.5
Built:        2017-03-01_01:03:26PM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml...)?
Docker container with `traefik:morbier-alpine`
`MaxIdleConnsPerHost` 200
`ulimit -n` 65536

Had problem like a #127 #157.
With ~100rps, after ~24h takes limit.
`server.go:2317: http: Accept error: accept tcp [::]:80: accept4: too many open files; retrying in 1s`.

emilevauge:
@IncSW Could you give use the whole configuration (flags, toml) ?

IncSW:
```toml
defaultEntryPoints = ["http"]
MaxIdleConnsPerHost = 200

accessLogsFile = "/var/log/access.log"

[web]
address = ":8080"
  [web.statistics]
    RecentErrors = 10
  [web.metrics.prometheus]
    Buckets = [0.1, 0.3, 1.2, 5.0]

[entryPoints]
  [entryPoints.http]
  address = ":80"

[frontends]
  [frontends.gateway-frontend]
  backend = "gateway-backend"
  passHostHeader = true
    [frontends.gateway-frontend.routes.host]
    rule = "Host:gateway.domain.com"
  [frontends.frontend-frontend]
  backend = "frontend-backend"
  passHostHeader = true
    [frontends.frontend-frontend.routes.host]
    rule = "Host:domain.com"

[file]
[backends]
  [backends.gateway-backend]
    [backends.gateway-backend.servers.server]
    url = "http://gateway:80"
  [backends.frontend-backend]
    [backends.frontend-backend.servers.server]
    url = "http://frontend:80"
```

emilevauge:
If you have the possibility to try commenting
```
  [web.statistics]
    RecentErrors = 10
  [web.metrics.prometheus]
    Buckets = [0.1, 0.3, 1.2, 5.0]
```
and then
```
accessLogsFile = "/var/log/access.log"
```
That would be awesome :)

IncSW:
It doesn't work.

timoreimann:
[...]

PR #1340, opened 2017-03-24
Add IdleConnTimeout to Traefik's http.server settings

bparli:
Without enforcing a timeout Traefik is susceptible to resource leakage, particularly when deployed as a public facing proxy exposed to the Internet.

Make sure IdleTimeout is configurable via the traefik.toml file

Set the default to be 180 seconds

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -69,6 +69,8 @@
 
 # IdleTimeout: maximum amount of time an idle (keep-alive) connection will remain idle before closing itself.
 # This is set to enforce closing of stale client connections.
+# Can be provided in a format supported by [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration) or as raw
+# values (digits). If no units are provided, the value is parsed assuming seconds.
 #
 # Optional
 # Default: "180s"
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -62,6 +62,14 @@
 #
 # ProvidersThrottleDuration = "5"
 
+# IdleTimeout: maximum amount of time an idle (keep-alive) connection will remain idle before closing itself.
+# This is set to enforce closing of stale client connections.
+#
+# Optional
+# Default: "180s"
+#
+# IdleTimeout = "360s"
+
 # If non-zero, controls the maximum idle (keep-alive) to keep per-host.  If zero, DefaultMaxIdleConnsPerHost is used.
 # If you encounter 'too many open files' errors, you can either change this value, or change `ulimit` value.
 #
@@ -1637,7 +1645,7 @@ RefreshSeconds = 15
 
 ```
 
-Items in the dynamodb table must have three attributes: 
+Items in the dynamodb table must have three attributes:
 
 
 - 'id' : string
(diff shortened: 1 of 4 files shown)


bparli:
Also, I think this should address #1322

timoreimann:
@bparli thanks for the contribution -- a pretty important one I should say.

Do I see correctly that we need to wait for #1342 before we can merge this?

timoreimann:
I'm also pretty confident this PR will

fix #1322

bparli:
@timoreimann IMO this isn't blocked by #1342; that was just an issue I noticed when testing this.  This setting will be much easier for folks to configure after that issue is fixed though.

timoreimann:
@bparli:

>@timoreimann IMO this isn't blocked by #1342; that was just an issue I noticed when testing this.  This setting will be much easier for folks to configure after that issue is fixed though.

IIUC, configuring is still possible because users may specify (nanoseconds) integers. One thing we have to take into consideration though is that if we use a `time.ParseDuration`-based approach in the future, [specifications missing a unit are going to be rejected](https://play.golang.org/p/md_zcRpk76). What this means is that accepting unit-less specs now will force us to continue providing support for it unless we want to make a breaking change.

WDYT?

bparli:
Thats a fair point, @timoreimann.  My thinking was its already broke for the ProvidersThrottleDuration setting anyway, but you're right, better and cleaner to fix #1342 first.  And it looks like you've already picked that one up.

errm:
LGTM, but waiting for #1342 does make sense ...

bparli:
Sure thing @timoreimann, added some content to toml.md and examples.md.  I also put ProvidersThrottleDuration in the examples.md for good measure.

docs/toml.md
```diff
@@ -67,6 +67,14 @@
 #
 # ProvidersThrottleDuration = "2s"
 
+# IdleTimeout: maximum amount of time an idle (keep-alive) connection will remain idle before closing itself.
+# This is set to enforce closing of stale client connections.
+#
```
timoreimann: Could you please copy the description of the valid input values [from here](https://github.com/containous/traefik/blame/7d256c9bb98b7786e972cacde02574619bbba21e/docs/toml.md#L13-L15)?

bparli:
Np, good catch to match the ProvidersThrottleDuration language

timoreimann:
Ah sorry, can't merge right away, we need one more pair of maintainer eyes.

@emilevauge could you have a look please?

Yggdrasil:
🥇 Thanks for reporting *and* fixing this @bparli. This issue led to some production downtime for us until we understood what was happening. We've resorted to restarting Traefik twice a week as a workaround. I'll be soooo happy when this is released.

bparli:
My pleasure, happy to contribute

Yggdrasil:
Would it be possible to mention this fix in the Changelog for 1.3.0? I couldn't find it in the RC's and I think it's worth a mention to inform those impacted by the leak.

Below's a sweet example of the difference this fix has made for us on one of our public facing webservers. On this server you can see memory usage climbing from the increasing number of open sockets, until Traefik's restarted. On May 16th around midnight we deployed v1.3.0-rc1, after which memory usage has been stable around 30MB. 😍 

schermafbeelding 2017-05-19 om 16 43 09

ldez:
@Yggdrasil the changelog is fixed. 😃  See #1642.

PR #1343, opened 2017-03-24
Improve rancher provider handling of service and container health states

### LINKED ISSUE #1253: Rancher provider does not correctly handle unhealthy containers
Traefik: v1.2.0-rc2 / morbier
Rancher: v1.4.1

**docker-compose.yml**
```yaml
version: '2'
services:
  traefik:
    image: traefik:v1.2.0-rc2
    stdin_open: true
    tty: true
    ports:
    - 80:80/tcp
    - 443:443/tcp
    - 8080:8080/tcp
    command:
    - --web
    - --rancher
    - --rancher.domain=development.cluster.rancher.foo.com
    - --rancher.endpoint=https://ui.cluster.rancher.foo.com
    - --rancher.accesskey=...
    - --rancher.secretkey=...
    - --logLevel=DEBUG
    labels:
      io.rancher.container.pull_image: always
      io.rancher.scheduler.global: 'true'
```

### What did you do?
Upgraded a service via the Rancher UI, enabling the `Start before Stopping` toggle. This service is configured with a health check.

The 'old' containers should be removed from the backend config when they are stopped by Rancher.

### What did you see instead?
Containers which have been stopped by Rancher remain in the backend config until they are **removed,** leading to failed requests.

kelchm:
This pull request aims to fix some basic issues which impact the usability of the rancher provider. During my testing with the rancher provider I found that services which did not have a healthy state were removed from the traefik config. In practice we should not care about the health of a service, but only about the health of that services individual containers. As a result I've made the following changes:

1. By default, we no longer filter services by their `HealthState`. A modified version of this functionality may be enabled using the 'EnableServiceHealthFilter' configuration toggle. This is particularly useful for 'advanced' configurations such as doing blue/green deployments.
2. A `RefreshSeconds` configuration parameter has been added (much like in the ECS provider) for configuring how often the Rancher API is queried.
2. Both containers and services are filtered by a combination of their `HealthState` and `State` due to the behavior of the Rancher API.
3. `healthy` and `upgrading-healthy` are considered to be a healthy `HealthState` for containers and services.
4.  `active`, `updating-active` and `upgraded` are considered to be a healthy `State` for services.
5. `running` and `updating-running` are considered to be a healthy `State` for containers.

If anyone has any feedback or test results please feel free to share them with me here or in #rancher on the Traefik Slack.

Fixes #1253

### DIFF
--- a/provider/rancher/rancher.go
+++ b/provider/rancher/rancher.go
@@ -21,11 +21,6 @@ import (
 	rancher "github.com/rancher/go-rancher/client"
 )
 
-const (
-	// RancherDefaultWatchTime is the duration of the interval when polling rancher
-	RancherDefaultWatchTime = 15 * time.Second
-)
-
 var (
 	withoutPagination *rancher.ListOpts
 )
@@ -34,12 +29,14 @@ var _ provider.Provider = (*Provider)(nil)
 
 // Provider holds configurations of the provider.
 type Provider struct {
-	provider.BaseProvider `mapstructure:",squash"`
-	Endpoint              string `description:"Rancher server HTTP(S) endpoint."`
-	AccessKey             string `description:"Rancher server access key."`
-	SecretKey             string `description:"Rancher server Secret Key."`
-	ExposedByDefault      bool   `description:"Expose Services by default"`
-	Domain                string `description:"Default domain used"`
+	provider.BaseProvider     `mapstructure:",squash"`
+	Endpoint                  string `description:"Rancher server HTTP(S) endpoint."`
+	AccessKey                 string `description:"Rancher server access key."`
+	SecretKey                 string `description:"Rancher server Secret Key."`
+	ExposedByDefault          bool   `description:"Expose Services by default"`
+	Domain                    string `description:"Default domain used"`
+	RefreshSeconds            int    `description:"Polling interval (in seconds)"`
+	EnableServiceHealthFilter bool   `description:"Filter services with unhealthy states and health states."`
 }
 
 type rancherData struct {
@@ -47,6 +44,7 @@ type rancherData struct {
 	Labels     map[string]string // List of labels set to container or service
 	Containers []string
 	Health     string
+	State      string
 }
 
 func init() {
@@ -56,7 +54,7 @@ func init() {
 }
 
 func (r rancherData) String() string {
-	return fmt.Sprintf("{name:%s, labels:%v, containers: %v, health: %s}", r.Name, r.Labels, r.Containers, r.Health)
+	return fmt.Sprintf("{name:%s, labels:%v, containers: %v, health: %s, state: %s}", r.Name, r.Labels, r.Containers, r.Health, r.State)
 }
 
 // Frontend Labels
@@ -261,7 +259,7 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 
 			if p.Watch {
 				_, cancel := context.WithCancel(ctx)
-				ticker := time.NewTicker(RancherDefaultWatchTime)
+				ticker := time.NewTicker(time.Second * time.Duration(p.RefreshSeconds))
 				pool.Go(func(stop chan bool) {
 					for {
 						select {
@@ -384,6 +382,7 @@ func parseRancherData(environments []*rancher.Environment, services []*rancher.S
 			rancherData := rancherData{
 				Name:       environment.Name + "/" + service.Name,
 				Health:     service.HealthState,
+				State:      service.State,
 				Labels:     make(map[string]string),
 				Containers: []string{},
 			}
@@ -393,11 +392,8 @@ func parseRancherData(environments []*rancher.Environment, services []*rancher.S
 			}
 
 			for _, container := range containers {
-				for key, value := range container.Labels {
-
-					if key == "io.rancher.stack_service.name" && value == rancherData.Name {
-						rancherData.Containers = append(rancherData.Containers, container.PrimaryIpAddress)
-					}
+				if container.Labels["io.rancher.stack_service.name"] == rancherData.Name && containerFilter(container) {
+					rancherData.Containers = append(rancherData.Containers, container.PrimaryIpAddress)
 				}
 			}
 			rancherDataList = append(rancherDataList, rancherData)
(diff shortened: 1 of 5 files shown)


kelchm:
@SantoDE, @nuclon I'd be interested to see your feedback on this.

One area I have not tested yet is how these changes affect containers that lack a health check.

errm:
LGTM, but is there a way to cover this case with a unit test?

kelchm:
@errm, which case specifically? For containers without a health check I need to actually test the behavior of the Rancher API.

timoreimann:
@kelchm with your PR, `parseRancherData()` now seems to skip containers from the result set that are missing certain health states. Ideally we can verify this through the calling `Provide()` method; since this will likely mean we'll have to jump through a number of hoops, however, I think it's also okay to test `parseRancherData()` directly.

errm:
Yeah I would be fine with testing `parseRancherData()` directly...

SantoDE:
Hey @kelchm,

as mentioned already, I like the idea where this PR is going to :) I'm also fine with migrating checks from the service to a container.

However, as mentioned twice already, we need some unit test to cover it ;)

kelchm:
I'm going to need to rework some of my original plan here -- it turns out that the Rancher API returns a `healthState` of `"healthy"` in situations we wouldn't expect. For example, if a stack is stopped any containers within that stack's services still return a `healthState` of `"healthy"`. I've opened an issue to confirm that this is the expected behavior of the Rancher API:  rancher/rancher/issues/8354

My thinking is that we will likely need to use a combination of factors to determine first if a service should be included in the config and then second if a given container should be included for that service.

jwentworth:
I think this is close to fixing what I see as the root cause of the issue. I think we need to add a check on the container's State (not just HealthState). I would say we'd only want to include containers that have a HealthState of "healthy" and a State of "running". I don't think I'd include containers that are being upgraded or initializing into the LB until they reach a healthy state, and as you mentioned before a healthy HealthState alone isn't enough to determine if a container should be included in the load balancer.

From my investigating the API and my use cases I think those are all the criteria that are needed if using the API (the metadata service deals with things a little differently, but this isn't using that)

kelchm:
@SantoDE, I think this is ready for your review. I think the only thing remaining are some minor updates to the documentation.

jwentworth:
This looks good to me, haven't been able to test it yet personally but this looks like it'd resolve the issues I'm seeing!

martinbaillie:
I've had a quick test this afternoon and it LGTM. A well-needed enhancement, thanks @kelchm.

Now it could be my use case bias, but this seems like it should be default functionality - I would say `EnableServiceHealthFilter` should default to true. As it stands, I'm seeing my Traefik instance include stopped containers as active backends (so long as they have the label). Thoughts?

kelchm:
@martinbaillie, containers which have a `HealthState` which is not 'healthy' or 'updating healthy' and/or a `State` which is not 'running' or 'updating-running' _should_ be filtered by containerFilter. I haven't seen the issue you are describing, but it's definitely possible I missed something.

Can you grab the API response for the container(s) you are seeing this behavior on? I'd be curious what the `State` and `HealthState` are.

martinbaillie:
@kelchm sorry I may have confused things. I meant it works perfectly with your PR enabled. I was questioning whether it should default to true rather than false by default, because without your PR enabled I saw stopped containers appear in the backend list which is undesirable default behaviour.

kelchm:
@martinbaillie, There are two types of filters, [containerFilter ](https://github.com/uqf/traefik/blob/a3333027f1620282dd416f25c8d7aa120b16e796/provider/rancher.go#L443)which filters at the container level and [serviceFilter](https://github.com/uqf/traefik/blob/a3333027f1620282dd416f25c8d7aa120b16e796/provider/rancher.go#L457) which filters at the service level.

I have `serviceFilter` disabled by default because it implies a more 'advanced' use case, such as replacing an existing stack with a new stack rather than upgrading the services within a stack. The `containerFilter` which is **always** on should prevent stopped (or otherwise unhealthy) containers from being brought in as backends even when `EnableServiceHealthFilter ` is disabled.

martinbaillie:
Thanks @kelchm that explains it. I was struggling to come up with a use case for when this would be desirable behaviour.

provider/rancher.go
```diff
@@ -456,9 +466,18 @@ func (provider *Rancher) serviceFilter(service rancherData) bool {
 		return false
 	}
 
-	if service.Health != "" && service.Health != "healthy" {
-		log.Debugf("Filtering unhealthy or starting service %s", service.Name)
-		return false
+	// Only filter services by Health (HealthState) and State is EnableServiceHealthFilter is true
```
SantoDE: typo I guess? 

> if EnableServiceHealthFilter is true... ?

kelchm: Good eye @SantoDE, fixed!

SantoDE:
Hey @kelchm ,

a) tests are failing :'(
b) you have to rebase again ;-)

Thanks for your work! :)

ldez:
@kelchm Could you squash your commits? Thanks.

provider/rancher/rancher.go
```diff
@@ -393,11 +392,8 @@ func parseRancherData(environments []*rancher.Environment, services []*rancher.S
 			}
 
 			for _, container := range containers {
-				for key, value := range container.Labels {
-
-					if key == "io.rancher.stack_service.name" && value == rancherData.Name {
-						rancherData.Containers = append(rancherData.Containers, container.PrimaryIpAddress)
-					}
+				if containerFilter(container) && container.Labels["io.rancher.stack_service.name"] == rancherData.Name {
```
emilevauge: It would be even better to write `if container.Labels["io.rancher.stack_service.name"] == rancherData.Name && containerFilter(container) {`, but meh :)

kelchm:
@emilevauge, @ldez done 👍

traefik.sample.toml
```diff
@@ -1046,13 +1046,26 @@
 #
 # Watch = true
 
+# Polling interval (in seconds)
+#
+# Optional
+#
+RefreshSeconds = 15
```
ldez: Could you comment this line.

traefik.sample.toml
```diff
@@ -1046,13 +1046,26 @@
...
 # Default: true
 #
 # ExposedByDefault = false
 
+# Filter services with unhealthy states and health states
+#
+# Optional
+# Default: false
+#
+EnableServiceHealthFilter = false
```
ldez: Could you comment this line.

traefik.sample.toml
kelchm: Fixed.

traefik.sample.toml
kelchm: Fixed.

Issue #1243, opened 2017-03-07
IP address + exported port instead of endpoint from marathon


lcottereau:
### What version of Traefik are you using (`traefik version`)?

* Version:      v1.2.0-rc2
* Codename:     morbier
* Go version:   go1.7.5
* Built:        2017-03-01_01:13:09PM
* OS/Arch:      linux/amd64

### What is your environment & configuration (arguments, toml...)?
Linux RHEL 7.2
#### my configuration file
```toml
    debug = true
    accessLogsFile = "log/access.log"
    logLevel = "INFO"
    defaultEntryPoints = ["http"]
    [entryPoints]
       [entryPoints.http]
       address = "infra-q-i-mes01:8008"
    [web]
    address = ":8009"
    [web.statistics]
       RecentErrors = 10
    [marathon]
    endpoint = "http://infra-q-i-mes01:8080"
    watch = true
    domain = "infra-q-i-mes01"
```

#### The deployed app details in Marathon :
```
    Host: infra-q-i-mes01
    IP Addresses: 10.16.0.3
    Ports: [31855]
    Endpoints: infra-q-i-mes01:31855
    Service Discovery: n/a
    Status: Started
    Staged at: 03/03/2017 à 18:26:26
    Started at: 03/03/2017 à 18:26:27
    Version: 2017-03-03T17:13:55.405Z
    Health: Healthy
    Mesos details: link
```
#### The deployed app configuration
Notice the 2 traefik labels
```json
    {
      "id": "trace",
      "cmd": null,
      "cpus": 1,
      "mem": 512,
      "disk": 0,
      "instances": 1,
      "container": {
        "docker": {
          "image": "sysadm-reg/assemblage/trace:1.4-20170303.164750-23",
          "network": "BRIDGE",
          "parameters": [],
          "portMappings": [
            {
              "containerPort": 8080,
              "protocol": "tcp",
              "name": "tomcat",
              "labels": null
            }
          ],
          "forcePullImage": true
        },
        "type": "DOCKER",
        "volumes": []
      },
      "env": {},
      "healthChecks": [
        {
          "protocol": "HTTP",
          "path": "/trace/api/health"
        }
      ],
      "labels": {
        "traefik.frontend.rule": "PathPrefix:/trace",
        "traefik.backend.loadbalancer.sticky": "true"
      },
      "uris": [ "file:///etc/catsa-anonymous-puller.credentials.tar.gz" ],
      "upgradeStrategy": {
[...]

PR #1345, opened 2017-03-26
[Marathon] Detect proper hostname automatically.

diegooliveira:
The IP-Per-Task feature changed the behavior for
clients without this configuration (using the task IP instead
of task hostname). This patch make the new behavior available
just for Mesos installation with IP-Per-Task enabled. It also
make it possible to force the use of task's hostname.

### DIFF
--- a/traefik.sample.toml
+++ b/traefik.sample.toml
@@ -599,6 +599,16 @@
 #
 # keepAlive = "10s"
 
+# By default, a task's IP address (as returned by the Marathon API) is used as 
+# backend server if an IP-per-task configuration can be found; otherwise, the
+# name of the host running the task is used.
+# The latter behavior can be enforced by enabling this switch.
+# 
+# Optional
+# Default: false
+#
+# forceTaskHostname: false 
+
 ################################################################
 # Mesos configuration backend
 ################################################################
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -969,6 +969,16 @@ domain = "marathon.localhost"
 # Default: "10s"
 #
 # keepAlive = "10s"
+
+# By default, a task's IP address (as returned by the Marathon API) is used as 
+# backend server if an IP-per-task configuration can be found; otherwise, the
+# name of the host running the task is used.
+# The latter behavior can be enforced by enabling this switch.
+# 
+# Optional
+# Default: false
+#
+# forceTaskHostname: false 
 ```
 
 Labels can be used on containers to override default behaviour:
--- a/provider/marathon/marathon.go
+++ b/provider/marathon/marathon.go
@@ -482,12 +482,9 @@ func processPorts(application marathon.Application, task marathon.Task) (int, er
 	portIndexLabel, ok := (*application.Labels)[labelPortIndex]
 	if ok {
 		var err error
-		portIndex, err = strconv.Atoi(portIndexLabel)
-		switch {
-		case err != nil:
-			return 0, fmt.Errorf("failed to parse port index label: %s", err)
-		case portIndex < 0, portIndex > len(ports)-1:
-			return 0, fmt.Errorf("port index %d must be within port range (0, %d)", portIndex, len(ports)-1)
+		portIndex, err = parseIndex(portIndexLabel, len(ports))
+		if err != nil {
+			return 0, fmt.Errorf("cannot use port index to select from %d ports: %s", len(ports), err)
 		}
 	}
 	return ports[portIndex], nil
@@ -543,7 +540,8 @@ func (p *Provider) getBackendServer(task marathon.Task, applications []marathon.
 			log.Errorf("Found %d task IP addresses but missing IP address index for Marathon application %s on task %s", numTaskIPAddresses, application.ID, task.ID)
 			return ""
 		}
-		ipAddressIdx, err := strconv.Atoi(ipAddressIdxStr)
+
+		ipAddressIdx, err := parseIndex(ipAddressIdxStr, numTaskIPAddresses)
 		if err != nil {
 			log.Errorf("Cannot use IP address index to select from %d task IP addresses for Marathon application %s on task %s: %s", numTaskIPAddresses, application.ID, task.ID, err)
 			return ""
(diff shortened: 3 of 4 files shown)


errm:
Looks like the unit tests are failing on this change...

traefik.sample.toml
```diff
@@ -586,6 +586,14 @@
 #
 # keepAlive = 10
 
+# Force the use of the task's hostname in the backend server address even in a
```
timoreimann: I think we should be a bit more elaborative here and describe the default behavior. I'm thinking of something like the following:

_By default, a task's IP address (as returned by the Marathon API) is used as backend server if an IP-per-task configuration can be found; otherwise, the name of the host running the task is used.
The latter behavior can be enforced by enabling this switch._

Please also copy the description into the `docs/toml.md` file.

timoreimann:
hey @diegooliveira, any chance to get an update on the PR? Thanks.

emilevauge:
@diegooliveira tests are still failing.

jensendw:
@emilevauge it looks like the tests are passing again.  Do you think this will be merged in before the next release?

provider/marathon_test.go
```diff
@@ -107,7 +110,7 @@ func TestMarathonLoadConfig(t *testing.T) {
 				"backend-test": {
 					Servers: map[string]types.Server{
 						"server-test": {
-							URL:    "http://127.0.0.1:80",
+							URL:    "http://localhost:80",
```
timoreimann: Was there a need to change this?

provider/marathon_test.go
```diff
@@ -1,10 +1,13 @@
 package provider
 
 import (
+	"encoding/json"
 	"errors"
 	"reflect"
 	"testing"
 
+	"fmt"
```
timoreimann: Can you sort the imports alphabetically?

provider/marathon_test.go
```diff
@@ -336,27 +339,37 @@ func TestMarathonLoadConfig(t *testing.T) {
...
-		fakeClient.AssertExpectations(t)
-		if c.expectedNil {
-			if actualConfig != nil {
-				t.Fatalf("Should have been nil, got %v", actualConfig)
-			}
-		} else {
-			// Compare backends
-			if !reflect.DeepEqual(actualConfig.Backends, c.expectedBackends) {
-				t.Fatalf("expected %#v, got %#v", c.expectedBackends, actualConfig.Backends)
+		t.Run(fmt.Sprintf("Running case: %s", appID), func(t *testing.T) {
```
timoreimann: I don't think you need the "Running case" prefix: it's clear we run a case when being logged.

provider/marathon_test.go
```diff
@@ -336,27 +339,37 @@ func TestMarathonLoadConfig(t *testing.T) {
...
+				ExposedByDefault: true,
+				marathonClient:   fakeClient,
 			}
-			if !reflect.DeepEqual(actualConfig.Frontends, c.expectedFrontends) {
-				t.Fatalf("expected %#v, got %#v", c.expectedFrontends, actualConfig.Frontends)
+			actualConfig := provider.loadMarathonConfig()
+			fakeClient.AssertExpectations(t)
+			if c.expectedNil {
+				if actualConfig != nil {
+					t.Fatalf("Should have been nil, got %v", actualConfig)
```
timoreimann: Please say "config should have been nil, got [...]".

provider/marathon_test.go
```diff
@@ -336,27 +339,37 @@ func TestMarathonLoadConfig(t *testing.T) {
...
+				}
+			} else {
+				// Compare backends
+				if !reflect.DeepEqual(actualConfig.Backends, c.expectedBackends) {
+					expected, _ := json.Marshal(c.expectedBackends)
+					actual, _ := json.Marshal(actualConfig.Backends)
+					t.Fatalf("expected\t %s\n, \tgot %s\n", expected, actual)
```
timoreimann: This should be an `t.Errorf` IMHO so that we can still see the frontend assertion even if the backend one fails.

provider/marathon_test.go
```diff
@@ -1434,3 +1447,76 @@ func TestMarathonGetSubDomain(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			expected: "sample.com",
+		},
+	}
+
+	for _, app := range applications {
+		t.Run(fmt.Sprintf("running %s", app.application.ID), func(t *testing.T) {
```
timoreimann: No "running" here too.

timoreimann: Can we parallelize the tests? If so, please call `t.Parallel()` here and do `app := app` before the call to `t.Run` to capture the loop variable.

provider/marathon_test.go
timoreimann: Can we parallelize the tests? If so, please call `t.Parallel()` here.

provider/marathon_test.go
```diff
@@ -1434,3 +1447,76 @@ func TestMarathonGetSubDomain(t *testing.T) {
...
+				IPAddresses: []*marathon.IPAddress{
+					{
+						IPAddress: "192.168.0.1",
+					},
+				},
+			}
+
+			actual := provider.getBackendServer(task, applications)
+			if actual != app.expected {
+				t.Errorf("App %s, expected %q, got %q", task.AppID, app.expected, actual)
```
timoreimann: No need to include the app ID in the message: we already get that context by the first parameter to `t.Run`.

timoreimann: Please swap the expected / got order: the actual value should go first (for consistency reasons and because that's Go-typical).

timoreimann:
Hey @diegooliveira, do you have time for a final update? Would be great to get this feature merged before the code freeze.

timoreimann:
Wanted to make sure this important PR makes it into the upcoming version 1.3 release, so I carried the PR "in-place". Here's what I did:

- Addressed my own comments.
- Completed the unit tests to cover all error paths.
- A bit of refactoring.
- Rebased the PR onto #1474 which already refactored some of the production and test codes.

@diegooliveira I hope all of this is okay with you.

timoreimann:
Fixes #1243.

timoreimann:
@containous/traefik please review.

diegooliveira:
@timoreimann nice work, sorry for the delay, busy month. The addition of spew was a nice touch.

kopax:
Hi, I have hostname detection issue with traefik 2.1, could this fix have been removed in v2?

PR #1350, opened 2017-03-27
Use TOML-compatible duration type.

### LINKED ISSUE #1342: Time.duration config settings not being read in correctly from Traefik.toml
Through the latest commit (240b2be)

Go 1.8, logLevel = "DEBUG"

### What did you do?
Tried starting Traefik with ProvidersThrottleDuration set in Traefik.toml.  I tried:

- `ProvidersThrottleDuration = 5`

- `ProvidersThrottleDuration = "5"`

- `ProvidersThrottleDuration = "5s"`

Traefik starts with the Global configuration indicating ProvidersThrottleDuration is set to 5 seconds.  By default this shows 2 seconds (`"ProvidersThrottleDuration":2000000000`).  I think it should be `"ProvidersThrottleDuration":5000000000`

### What did you see instead?
- The first setting sets it to 5 ns (`"ProvidersThrottleDuration":5`)

- The second setting is actually in the file as an example.  But both the second and third fail with: `Error reading TOML config file /Users/bparli/go/src/github.com/containous/traefik/traefik.toml : toml: cannot load TOML value of type string into a Go integer`

timoreimann:
This PR introduces a new `time.Duration`-based type capable of deserializing durations in TOML specs. Without providing a unit specifier, it defaults to seconds for backwards compatibility reasons and the fact that most of our duration specifiers are at the order of seconds.

~~TODO:  Switch `glide.yaml` back to github.com/containous/flaeg once containous/flaeg#34 merges.~~ Done.

Fixes #1342.

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -9,13 +9,15 @@
 # Global configuration
 ################################################################
 
-# Timeout in seconds.
-# Duration to give active requests a chance to finish during hot-reloads
+# Duration to give active requests a chance to finish during hot-reloads.
+# Can be provided in a format supported by [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration) or as raw
+# values (digits). If no units are provided, the value is parsed assuming
+# seconds.
 #
 # Optional
-# Default: 10
+# Default: "10s"
 #
-# graceTimeOut = 10
+# graceTimeOut = "10s"
 
 # Enable debug mode
 #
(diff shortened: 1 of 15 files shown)


timoreimann:
@containous/traefik Please review.

timoreimann:
@vdemeester:

>SGTM but this would mean ppl that have this option specified as an int 10 would need to update their configuration, right ?

No, precisely not: The custom duration type assumes second scale if you don't specify a unit.

See also the description in the PR, the docs, and [the source of truth](https://github.com/containous/flaeg/pull/34/files#diff-af4299a1937f6eada4d1dec66083d437R152). :)

vdemeester:
@timoreimann thanks 👼 sorry I'm not fully awake this morning it seems 😂

timoreimann:
@vdemeester haha no worries. Nothing that an extra shot of caffeine can't fix. :-)

dtomcej:
LGTM, But I might make it clearer in the docs that not providing a unit will result in the value being parsed as seconds. Other proxies use raw values as milliseconds, so that would be a distinction worth making.

timoreimann:
@dtomcej I tried to document that the value defaults to seconds without a unit. Do you think it should be clearer? If so, could you provide a better phrasing please?

docs/toml.md
```diff
@@ -56,11 +57,13 @@
 # Backends throttle duration: minimum duration in seconds between 2 events from providers
 # before applying a new configuration. It avoids unnecessary reloads if multiples events
 # are sent in a short amount of time.
+# Can be provided in a format supported by [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration) or as digits
+# only. In the latter case, seconds are assumed.
```
dtomcej: `or as digits only` -> `or as raw values (digits)`
`In the latter case, seconds are assumed.` -> `If no units are provided, the value is parsed assuming seconds.`

timoreimann: @dtomcej Updated the docs.

Issue #1075, opened 2017-01-27
etcd kv store with username and password


kleinsasserm:
I like to use an etcd kv store for traefik and I would like to use the username and password (authentication) feature of etcd. I'am not able to find a documentation for a custom etcd.tmpl where it may be possible to configure the etcd credentinals.

Is it currently possible to configure user/pass for etcd or will I have to use confd to manage the config in kv store?

Thanks!

tcolgate:
Have sumitted #1357

ldez:
Closed by #1357

PR #1357, opened 2017-03-28
Add libkv Username and Password

tcolgate:


tcolgate:
This lets you configure the username/password params in libkv. This is definitely useful for etcd, but may not be obvious for others. The struct embedding makes it hard to hide in the other KV providers where it may not be so relevant.

timoreimann:
Refs #1075.

tcolgate:
@emilevauge Docs updated.

tcolgate:
@emilevauge rebased on master

Issue #30, opened 2015-09-29
More advanced middlewares


lucmichalski:
Hi guys,

Hope you are all well !

Here are the 
- Transform Request/Response
- Managing templates for frontends body transformation
- Concurrent sub-requests to nest json responses in a middleware

Cheers,

seeder:
So i just tried putting traefik in front of Jenkins ... no joy, it needs some response rewriting ( in Location header for example )
Great project, makes stuff cleaner than any apache/haproxy/nginx frankensteins monsters, but needs a bit more feature oomph exposed

EmileVauge:
@seeder could you give me a concrete exemple? I already have some implementation ideas, but I'm still not completely satisfied by those ones ^^

seeder:
@EmileVauge https://wiki.jenkins-ci.org/display/JENKINS/Jenkins+says+my+reverse+proxy+setup+is+broken

lucmichalski:
For my part, it would be great to have a rewrite module like on Vulcand (https://vulcand.io/proxy.html#rewrites-and-redirects).

My use case is to embbed Traefik inside a docker container in order to map several services:

Current use (with a file generated during the build of the container):
/service_1 going to loadbalanced ports or different front-ends
/service_2 going to one specific front-end

Wished use of Traefik:
/api/v1/service1/function1
/api/v1/service1/function2
/api/v1/service2/

mathieu-pousse:
+1 would be very nice !

ekozan:
@emilevauge Yop, may be can we discuss about many middlewares like auth, logs, etc ... for example I want to use the next feature of Coreos/Dex ( groups ) for make an oauth2 login on some admin applications on my k8s cluster. the real question is "How we can pass the options to the middleware"

sandstrom:
Transforming request/response would be great! Ability to adjust headers (add/remove/edit) would probably go a long way.

sybrandy:
Hello,

[...]

PR #1370, opened 2017-03-31
[WIP] Add basic plugin support

### LINKED ISSUE #1048: Custom Middlewares?
Apologize if this is the wrong forum, Is there support for or would be a good feature to support custom injectable middlewares. Traefik looks amazing, we are using Fabio from ebay now and are looking at mantl and found Traefik.

### LINKED ISSUE #1336: enable custom plugins/middlewares for Traefik
After seeing the [Go1.8 new plugin feature](https://golang.org/pkg/plugin/) I though that this could help a lot o people to add specific functionalities to Traefik.

Instead of building/compiling/shipping a custom-made version of Traefik to enable a custom functionality it would be possible to write way simpler custom-made middlewares with this approach, doesn't it ? 

Try imagine creating a package that receives the request at a parameter without having to recompile the whole Traefik repository just to add a small change. Does it sounds like a middleware ? Because for me it is ! It's just a go1.8-plugin-based-middleware !

What do you guys think ?

emilevauge:
This PR is a proposal for adding basic plugin support in Traefik using the new Golang 1.8 plugin mechanism.

Right now, only `Middleware` plugin type is available.
Build your plugin using:

```sh
go build -buildmode=plugin -o sample.so sample.go
```

Then add in you Traefik toml file:

```toml
[[plugins]]
path = "sample.so"
```

Please, give your feedback on this basic implementation :)

/cc @containous/traefik 

Fixes #1336
Fixes #1048
Part of #30

Signed-off-by: Emile Vauge <emile@vauge.com>

### DIFF
--- a/configuration.go
+++ b/configuration.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/containous/flaeg"
 	"github.com/containous/traefik/acme"
+	"github.com/containous/traefik/plugin"
 	"github.com/containous/traefik/provider"
 	"github.com/containous/traefik/types"
 )
@@ -40,6 +41,7 @@ type GlobalConfiguration struct {
 	IdleTimeout               flaeg.Duration          `description:"maximum amount of time an idle (keep-alive) connection will remain idle before closing itself."`
 	InsecureSkipVerify        bool                    `description:"Disable SSL certificate verification"`
 	Retry                     *Retry                  `description:"Enable retry sending request if network error"`
+	Plugins                   plugin.Plugins          `description:"External plugins definition"`
 	Docker                    *provider.Docker        `description:"Enable Docker backend"`
 	File                      *provider.File          `description:"Enable File backend"`
 	Web                       *WebProvider            `description:"Enable Web backend"`
@@ -470,6 +472,7 @@ func NewTraefikConfiguration() *TraefikConfiguration {
 			MaxIdleConnsPerHost:       200,
 			IdleTimeout:               flaeg.Duration(180 * time.Second),
 			CheckNewVersion:           true,
+			Plugins:                   plugin.Plugins{},
 		},
 		ConfigFile: "",
 	}
--- a/plugin/manager.go
+++ b/plugin/manager.go
@@ -1,0 +1,50 @@
+package plugin
+
+import (
+	"fmt"
+	goplugin "plugin"
+)
+
+// Manager is in charge of instantiating plugins and storing those in memory
+type Manager struct {
+	middlewares []*Middleware
+}
+
+// NewManager builds a new manager
+func NewManager() *Manager {
+	return &Manager{
+		middlewares: []*Middleware{},
+	}
+}
+
+// Load loads a plugin
+func (m *Manager) Load(plugin Plugin) error {
+	p, err := goplugin.Open(plugin.Path)
+	if err != nil {
+		return fmt.Errorf("error opening plugin: %s", err)
+	}
+	loader, err := p.Lookup("Load")
+	if err != nil {
+		return fmt.Errorf("error in plugin Lookup: %s", err)
+	}
+	load, ok := loader.(func() interface{})
+	if !ok {
+		return fmt.Errorf("plugin from %+v does not implement Load() interface{} function", plugin)
+	}
+	instance := load()
+	if instance == nil {
+		return fmt.Errorf("plugin from %+v does not implement Load() interface{} function", plugin)
+	}
+	if middleware, ok := instance.(Middleware); ok {
+		// if Middleware, add to middleware list
+		m.middlewares = append(m.middlewares, &middleware)
+	} else {
+		return fmt.Errorf("plugin from %+v does not implement any plugin interface", plugin)
+	}
+	return nil
+}
+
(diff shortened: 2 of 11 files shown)


plugin/plugin.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// Plugins defines a set of Plugin
+type Plugins []*Plugin
+
+//Set []*Plugin
+func (p *Plugins) Set(str string) error {
+	exps := strings.Split(str, ",")
+	if len(exps) == 0 {
+		return fmt.Errorf("Bad Plugin format: %s", str)
```
timoreimann: Bad -> bad

plugin/manager.go
```diff
@@ -0,0 +1,44 @@
+package plugin
+
+import (
+	"fmt"
+	goplugin "plugin"
+)
+
+// Manager is in charge of instanciating plugins and storing those in memory
+type Manager struct {
+	middlewares []*Middleware
```
timoreimann: Does this have to be a slice of pointers, or could it be simplified to a slice of values?

plugin/plugin.go
```diff
@@ -0,0 +1,42 @@
...
+	"strings"
+)
+
+// Plugin defines a plugin configuration in traefik
+type Plugin struct {
+	Path string
+}
+
+// Plugins defines a set of Plugin
+type Plugins []*Plugin
```
timoreimann: Slice of values possible?

plugin/plugin.go
```diff
@@ -0,0 +1,42 @@
...
+
+//Set []*Plugin
```
timoreimann: GoDoc should be expanded here and below.

plugin/plugin.go
```diff
@@ -0,0 +1,42 @@
...
+func (p *Plugins) String() string { return fmt.Sprintf("%+v", *p) }
+
+//SetValue sets []*Plugin into the parser
+func (p *Plugins) SetValue(val interface{}) {
+	*p = Plugins(val.(Plugins))
+}
+
+// Type exports the Plugins type as a string
+func (p *Plugins) Type() string {
+	return fmt.Sprint("plugins")
```
timoreimann: No need to use `fmt.Sprint`, can return a string value directly.

configuration.go
```diff
@@ -38,6 +39,7 @@ type GlobalConfiguration struct {
 	MaxIdleConnsPerHost       int                     `description:"If non-zero, controls the maximum idle (keep-alive) to keep per-host.  If zero, DefaultMaxIdleConnsPerHost is used"`
 	InsecureSkipVerify        bool                    `description:"Disable SSL certificate verification"`
 	Retry                     *Retry                  `description:"Enable retry sending request if network error"`
+	Plugins                   plugin.Plugins          `description:"External plugins definition"`
```
timoreimann: I'd amend that it currently supports middlewares only.

plugin/manager.go
```diff
@@ -0,0 +1,44 @@
+package plugin
+
+import (
+	"fmt"
+	goplugin "plugin"
+)
+
+// Manager is in charge of instanciating plugins and storing those in memory
```
emilevauge: fixed

plugin/manager.go
emilevauge: The whole question is: do we allow stateful plugins?

plugin/plugin.go
emilevauge: yep, done ;)

plugin/plugin.go
emilevauge: Done

plugin/plugin.go
emilevauge: Hum, don't know why this happened ^^

configuration.go
emilevauge: I will give more details in the documentation. I prefer to keep flags description simple :)

plugin/manager.go
```diff
@@ -0,0 +1,44 @@
...
+	if err != nil {
+		return fmt.Errorf("Error in plugin Lookup: %s", err)
+	}
+	instance := loader.(func() interface{})()
+	middleware, ok := instance.(Middleware)
+	if ok {
+		// if Middleware, add to middleware list
+		m.middlewares = append(m.middlewares, &middleware)
+	} else {
+		return fmt.Errorf("Plugin from %+v does not implement any plugin interface", plugin)
```
emilevauge: done

emilevauge:
@timoreimann thanks for your review :)

>Should we expand the docs?

Of course! That's why I added `[WIP]` in the description, just wanted to validate the implementation first :)

>We have no single test. Should we add some?

I will add an integration test.

>I wonder if this will affect building Traefik locally on Mac machines (which I use) where plugins are not supported yet. Will need to test.

Hum. I hope this will not be an issue... Looking forward to get the results of your tests.

emilevauge:
OK, bad news... We can't load an external plugin if Traefik is built with `CGO_ENABLED=0`, and we **really** need this to build a statically linked golang executable to run in a Docker container.
https://github.com/golang/go/issues/19569

emilevauge:
To be continued #1865

PR #1374, opened 2017-04-02
Add Path Replacement Rule

ssttevee:
Replaces the old path with the new one and adds the old path to the `X-Replaced-Path` header.

My use case is to map to a FAAS web triggers (i.e. google cloud functions, aws lambda).

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -84,6 +84,7 @@ Modifier rules only modify the request. They do not have any impact on routing d
 Following is the list of existing modifier rules:
 
 - `AddPrefix: /products`: Add path prefix to the existing request path prior to forwarding the request to the backend.
+- `ReplacePath: /serverless-path`: Replaces the path and adds the old path to the `X-Replaced-Path` header. Useful for mapping to AWS Lambda or Google Cloud Functions.
 
 ### Matchers
 
--- a/middlewares/replace_path.go
+++ b/middlewares/replace_path.go
@@ -1,0 +1,20 @@
+package middlewares
+
+import (
+	"net/http"
+)
+
+// ReplacePath is a middleware used to replace the path of a URL request
+type ReplacePath struct {
+	Handler http.Handler
+	Path    string
+}
+
+// ReplacedPathHeader is the default header to set the old path to
+const ReplacedPathHeader = "X-Replaced-Path"
+
+func (s *ReplacePath) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	r.Header.Add(ReplacedPathHeader, r.URL.Path)
+	r.URL.Path = s.Path
+	s.Handler.ServeHTTP(w, r)
+}
--- a/middlewares/replace_path_test.go
+++ b/middlewares/replace_path_test.go
@@ -1,0 +1,44 @@
+package middlewares_test
+
+import (
+	"net/http"
+	"testing"
+
+	"github.com/containous/traefik/middlewares"
+)
+
+func TestReplacePath(t *testing.T) {
+	const replacementPath = "/replacement-path"
+
+	paths := []string{
+		"/example",
+		"/some/really/long/path",
+	}
+
+	for _, path := range paths {
+		t.Run(path, func(t *testing.T) {
+			var newPath, oldPath string
+			handler := &middlewares.ReplacePath{
+				Path: replacementPath,
+				Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					newPath = r.URL.Path
+					oldPath = r.Header.Get("X-Replaced-Path")
+				}),
+			}
+
+			req, err := http.NewRequest("GET", "http://localhost"+path, nil)
+			if err != nil {
+				t.Error(err)
+			}
+
+			handler.ServeHTTP(nil, req)
+			if newPath != replacementPath {
+				t.Fatalf("new path should be '%s'", replacementPath)
+			}
+
+			if oldPath != path {
+				t.Fatalf("old path should be '%s'", path)
+			}
+		})
+	}
+}
--- a/server/rules.go
+++ b/server/rules.go
@@ -75,6 +75,13 @@ func (r *Rules) pathStrip(paths ...string) *mux.Route {
 	return r.route.route
 }
 
+func (r *Rules) replacePath(paths ...string) *mux.Route {
+	for _, path := range paths {
+		r.route.replacePath = path
+	}
+	return r.route.route
+}
+
 func (r *Rules) addPrefix(paths ...string) *mux.Route {
 	for _, path := range paths {
 		r.route.addPrefix = path
@@ -116,6 +123,7 @@ func (r *Rules) parseRules(expression string, onRule func(functionName string, f
 		"Headers":         r.headers,
 		"HeadersRegexp":   r.headersRegexp,
 		"AddPrefix":       r.addPrefix,
+		"ReplacePath":     r.replacePath,
 	}
 
 	if len(expression) == 0 {
--- a/server/server.go
+++ b/server/server.go
@@ -65,6 +65,7 @@ type serverRoute struct {
 	route         *mux.Route
 	stripPrefixes []string
 	addPrefix     string
+	replacePath   string
 }
 
 // NewServer returns an initialized Server.
@@ -803,6 +804,14 @@ func (server *Server) wireFrontendBackend(serverRoute *serverRoute, handler http
 		handler = &middlewares.StripPrefix{
 			Prefixes: serverRoute.stripPrefixes,
 			Handler:  handler,
+		}
+	}
+
+	// path replace
+	if len(serverRoute.replacePath) > 0 {
+		handler = &middlewares.ReplacePath{
+			Path:    serverRoute.replacePath,
+			Handler: handler,
 		}
 	}
 


docs/basics.md
```diff
@@ -82,6 +82,7 @@ Frontends can be defined using the following rules:
 - `HostRegexp: traefik.io, {subdomain:[a-z]+}.traefik.io`: Adds a matcher for the URL hosts. It accepts templates with zero or more URL variables enclosed by `{}`. Variables can define an optional regexp pattern to be matched.
 - `Method: GET, POST, PUT`: Method adds a matcher for HTTP methods. It accepts a sequence of one or more methods to be matched.
 - `Path: /products/, /articles/{category}/{id:[0-9]+}`: Path adds a matcher for the URL paths. It accepts templates with zero or more URL variables enclosed by `{}`.
+- `PathReplace: /serverless-path`: Replaces the path and adds the old path to the `X-Replaced-Path` header. Useful for mapping to AWS Lambda or Google Cloud Functions.
```
timoreimann: I'm not sure if it was a conscious decision, but we seem to have a pattern that `Path*` match a path and optionally modify it (e.g., `Path`, `PathPrefix`, `PathPrefixStrip`), while `*Path` do modifications only (`AddPrefix`). Regardless of how deliberate this naming choice was, I like it. :-)

Your new rule seems to fall into the second category. If that was your intention too, then I'd suggest to call it `ReplacePath` for naming consistency reasons.

middlewares/pathReplace.go
```diff
@@ -0,0 +1,23 @@
...
+
+// PathReplace is a middleware used to replace the path of a URL request
+type PathReplace struct {
+	Handler http.Handler
+	Path    string
+}
+
+func (s *PathReplace) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	r.Header.Add("X-Replaced-Path", r.URL.Path)
+	r.URL.Path = s.Path
```
timoreimann: A few questions:

1. Is it intentional that you replace the entire path?
1. Shouldn't we check if the path we intend to replace actually exists in `r.URL.Path` and only replace / add the header if it does?
1. What's the desired behavior if `s.Path` is shorter than `r.URL.Path`, i.e., we cannot replace it completely?

server.go
```diff
@@ -66,6 +66,7 @@ type serverRoute struct {
 	route         *mux.Route
 	stripPrefixes []string
 	addPrefix     string
+	pathReplace   string
```
timoreimann: Should be called `replacePath` if you follow my naming reasoning above.

rules.go
```diff
@@ -74,6 +74,13 @@ func (r *Rules) pathStrip(paths ...string) *mux.Route {
 	return r.route.route
 }
 
+func (r *Rules) pathReplace(paths ...string) *mux.Route {
+	for _, path := range paths {
+		r.route.addPrefix = path
```
timoreimann: You mean `pathReplace` / `replacePath` here.

middlewares/pathReplace.go
ssttevee: Yes, the idea is to replace the entire path so that all paths will be forwarded to a single path on the backend.

For example, if I'm using Google Cloud Functions, I want both `http://example.com/` and `http://example.com/some/long/url.html` to be replaced with `/some-function`, then I can simulate having multiple paths on `https://some-project.cloudfunctions.net/some-function`.

Regarding your third question, I'm not sure why we would ever be unable to replace the path completely.

ssttevee:
I agree with your suggestion to rename it, however I think you're missing the point of my changes. My reply to your comment further explains the idea.

timoreimann:
Thanks, I got a clearer picture now. Makes total sense to me.

If you could do the rename and add a test or two, I'd be glad. :-)

ssttevee:
I have made the changes, but I feel like one might want to customize the header that it is set to too.

middlewares/replacePath.go
```diff
@@ -0,0 +1,26 @@
...
+
+func (s *ReplacePath) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	r.Header.Add(ReplacedPathHeader, r.URL.Path)
+	r.URL.Path = s.Path
+	r.RequestURI = r.URL.RequestURI()
+	s.Handler.ServeHTTP(w, r)
+}
+
+// SetHandler sets handler
+func (s *ReplacePath) SetHandler(Handler http.Handler) {
```
emilevauge: Is it needed (apart from tests) ?

middlewares/replacePath_test.go
```diff
@@ -0,0 +1,42 @@
...
+		oldPath = r.Header.Get("X-Replaced-Path")
+	}))
+
+	paths := []string{
+		"/example",
+		"/some/really/long/path",
+	}
+
+	for _, path := range paths {
+		req, err := http.NewRequest("ANY", "http://localhost"+path, nil)
```
timoreimann: Please use subtests and pass `path` to `t.Run`. Right now, if a test fails, there's no way to tell which one did.

middlewares/replacePath_test.go
```diff
@@ -0,0 +1,42 @@
+package middlewares_test
+
+import (
+	"github.com/containous/traefik/middlewares"
```
timoreimann: Please move Github imports below stdlib ones, separated by a blank line.

middlewares/replacePath_test.go
timoreimann: Unless `ANY` is some kind of special place holder accepted by Go/HTTP, I'd rather like to use `GET`. This will be safe to work in the future too.

rules.go
```diff
@@ -74,6 +74,13 @@ func (r *Rules) pathStrip(paths ...string) *mux.Route {
 	return r.route.route
 }
 
+func (r *Rules) replacePath(paths ...string) *mux.Route {
+	for _, path := range paths {
+		r.route.addPrefix = path
```
timoreimann: What about https://github.com/containous/traefik/pull/1374/files/12e21cd2e9c5c22c8aa2683339b08a70b62e4095#diff-c06ec64a1e1b34f38b2152c5a5169d7c?

server.go
```diff
@@ -791,6 +792,14 @@ func (server *Server) wireFrontendBackend(serverRoute *serverRoute, handler http
 		}
 	}
 
+	// path replace
+	if len(serverRoute.replacePath) > 0 {
```
timoreimann: This definitely needs a test.

middlewares/replacePath.go
ssttevee: I was wondering the same thing for the `AddPrefix` handler. I thought it might be part of an interface so I copied it anyways.

I'll remove it since you don't think it's needed either.

server.go
ssttevee: there is no existing test file for server.go, should I make one to only test this part?

timoreimann: Please do.

ssttevee: actually, does this function even need to be a method of `Server`? it doesn't access any of it's fields.

timoreimann: Yeah good catch, we could convert it from a method to a function right away. Would simplify testing.

ldez:
@ssttevee do you the time to update this PR before the upcoming feature freeze for the 1.3 release?

ssttevee:
I couldn't think of a good way to test the `wireFrontendBackend` method, but everything else should be done.

middlewares/replace_path_test.go
```diff
@@ -0,0 +1,44 @@
...
+	paths := []string{
+		"/example",
+		"/some/really/long/path",
+	}
+
+	for _, path := range paths {
+		t.Run(path, func(t *testing.T) {
+			req, err := http.NewRequest("GET", "http://localhost"+path, nil)
+			if err != nil {
+				t.Error(err)
```
timoreimann: This should be a `t.Fatal`.

middlewares/replace_path_test.go
```diff
@@ -0,0 +1,44 @@
...
+			}
+
+			handler.ServeHTTP(nil, req)
+			if newPath != replacementPath {
+				t.Errorf("new path should be \"%s\"", replacementPath)
```
timoreimann: I suggest to use single quotes (`'`) to avoid escaping.

middlewares/replace_path_test.go
```diff
@@ -0,0 +1,44 @@
...
+			}
+
+			if oldPath != path {
+				t.Errorf("old path should be \"%s\"", path)
```
timoreimann: Single quotes preferred here too.

middlewares/replace_path_test.go
```diff
@@ -0,0 +1,44 @@
...
+	const replacementPath = "/replacement-path"
+
+	var newPath, oldPath string
+	handler := &middlewares.ReplacePath{
+		Path: replacementPath,
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			newPath = r.URL.Path
+			oldPath = r.Header.Get("X-Replaced-Path")
+		}),
+	}
```
timoreimann: Please move the block from line 13 to 20 into the test loop body in order to make sure the tests work okay even if we run them in parallel.

middlewares/replace_path.go
```diff
@@ -0,0 +1,21 @@
...
+	Path    string
+}
+
+// ReplacedPathHeader is the default header to set the old path to
+const ReplacedPathHeader = "X-Replaced-Path"
+
+func (s *ReplacePath) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	r.Header.Add(ReplacedPathHeader, r.URL.Path)
+	r.URL.Path = s.Path
+	r.RequestURI = r.URL.RequestURI()
```
timoreimann: I'm not too deep into this part of HTTP, but is it legitimate to change this?

This is what [net/http](https://golang.org/pkg/net/http/#Request) says:

```
// RequestURI is the unmodified Request-URI of the
// Request-Line (RFC 2616, Section 5.1) as sent by the client
// to a server. Usually the URL field should be used instead.
// It is an error to set this field in an HTTP client request.
RequestURI string
```

WDYT?

ssttevee: I did find it a bit strange too, but iirc, it does not directly affect a client request.

I'm probably wrong, but at one point there might have been some code that read r.RequestURI` to use for the path to use for the backend request. That would also explain why the same expression elsewhere in the code base.

ssttevee: In the `github.com/vulcand/oxy`, when creating a new forwarder, it is used for `r.Opaque` which in http(s) requests has no effect.

ssttevee:
After investigating, I've concluded that setting `r.RequestURI` is unnecessary (maybe even erroneous). [My reasoning is in this comment.](https://github.com/containous/traefik/pull/1374#discussion_r113162101)

It will probably be best to remove all instances of this from the rest of code base to avoid future contributors copying it.

middlewares/replace_path.go
timoreimann: :+1:

timoreimann:
@ssttevee one last request: could you squash your commits for git hygiene reasons?

ssttevee:
it's done!

aantono:
I believe I found a possible bug, or at least a not intuitive behavior when `ReplacePath` rule is preceded by `PathPrefixStrip`
Basically because `PathPrefixStrip` is a Matcher, but also happens to be a Modifier as well, when combined with `ReplacePath`, like `PathPrefixStrip:/management;ReplacePath:/health`, due to the order in which the Modifiers are applied in [`server.go`](https://github.com/containous/traefik/blob/master/server/server.go#L800), the above rule-set does not behave as expected.  

While one would expect that we match the request to `/management` url, strip the `/management` prefix, followed by a replace with `/health`.  This is NOT, however, what is happening.  The matcher does the job correctly and matches the request on `/management` and sends it down the correct `Handler` chain, but that's when the issue creeps in.  

Because `ReplacePath` Modifier gets executed first in the `Handler` chain, it replaces `/management` with `/health`, and when `StripPrefix` Modifier gets called, it is trying to match the Path with `/management` prefix, and obviously failing to do so, thus returning `404-Not Found`. 

If the Rule is written as `PathPrefixStrip:/management;ReplacePath:/management/health`, then everything works, but the actual invocation URL Path ends up being NOT `/management/health` but `/health`, thus making this configuration be misleading.

Should I open a new Issue to track this improvement?

timoreimann:
@aantono thanks for the report.

Please do file a new bug report and link it here. Appreciated!

aantono:
Done... I'll try to see if I can make PR with a fix tomorrow.
My current thinking is to try to move the `ReplacePath` block in `server.go` to be the first in the chain, above the `AddPrefix` - https://github.com/containous/traefik/blob/master/server/server.go#L778
Thoughts?

timoreimann:
@aantono sounds reasonable on first sight.

We should try to add a few tests to make sure the known combinations work as expected.

IceXPR:
I still have the problem with the route replacements in the containous/traefik image.

IceXPR:
@aantono What version of Traefik did you use for the workaround you mentioned (PathPrefixStrip:/management;ReplacePath:/management/health)?

aantono:
That was the 1.3-RC1 (the one I used was downloaded from the releases page)

PR #1381, opened 2017-04-04
log XFF and XFP if present

mildis:
Check if X-Forwarded-For and X-Forwarded-Proto are present in the header.
Is so, overwrite host and proto in the logs.


emilevauge:
@mildis thanks  a lot for your contribution :)
The logger middleware is currently being rewritten here: #1485.
Do you mind joining us in this PR to discuss if adding this feature is doable ?

PR #1208, opened 2017-03-01
Merge v1.2.0 rc2 master

emilevauge:
This PR merge `v1.2` branch into master.
@timoreimann, this one is for you ;)

PR #1204, opened 2017-03-01
Prepare release v1.2.0 rc2

emilevauge:
/cc @containous/traefik

PR #1383, opened 2017-04-04
Merge v1.2.1 master

emilevauge:
This PR merges v1.2.1 back into master.
There is an issue, commits from `Add v1.2 codename` to `Merge pull request #1204 from containous/prepare-release-v1.2.0-rc2` should not be in this PR. Those are already in master thanks to https://github.com/containous/traefik/pull/1208. It seems that the problem comes from PR https://github.com/containous/traefik/pull/1208 where there is no merge commit and commit hashes [are](https://github.com/containous/traefik/pull/1383/commits/7a7992a6390d9d7ef9ebd781b3e8d19f83316a27) [different](https://github.com/containous/traefik/commit/7a164ed401f08407f64aa76634656243ee7e91ba) :thinking: 
~~So,~~

~~1. either I leave it like this (with some duplicates in the history)~~
~~2. either I can rebase interactively and drop commits from `Add v1.2 codename` to `Merge pull request #1204 from containous/prepare-release-v1.2.0-rc2`. The history will look better, but I fear that the next merge will end to the same issue :'(~~

~~WDYT @containous/traefik ?~~

Finally, what I did:

1. on branch `v1.2`,  `git rev-list --no-merges --reverse 9b24e13..fba3db5 > /tmp/rebase`: list all commits without merge commits

2. On branch `merge-v1.2.1-master` (from master), `cat /tmp/rebase | git cherry-pick --stdin`


timoreimann:
This is pretty much impossible to review due to the sheer number of lines changed. I suggest we merge and deal with any potential problems in master.

If I remember our agreement correctly, we are going to do cherry-picking from release branches into master in the future. In that case, it's okay to not try to "save the merge" here and go with the cherry-picking solution already (which will presumably make future merges of the release branch into master harder, but we won't have to do that again).

emilevauge:
@containous/traefik your review is required to merge :)

emilevauge:
OMG it's green, I can finally merge this PR!

>410 files changed?

We all love vendoring ;)

Issue #1390, opened 2017-04-05
Marathon backend with CNI Plugin doesn't work


lusche:
### What is your environment & configuration (arguments, toml...)?
We run an marathon/mesos Cluster and try to use Traefik as LB solution. Unfortunately both the mesos and the marathon backend didn't work as aspected. Maybe i use it wrong. 

The Environment:
- CentOS Linux release 7.3.1611 (Core)
- docker-engine.x86_64 1.11.2-1.el7.centos
- mesos 1.1.0
- marathon-1.4.1
- Traefik version v1.2.1 built on 2017-03-27_03:19:40PM
- CNI Calico (version v1.1.0, build 882dd008)

traefik.toml
```toml
debug = true

################################################################
# Web configuration backend
################################################################
[web]
address = ":8080"

[mesos]
endpoint = "zk://mesos3.domain.xyz:2181,mesos4.domain.xyz:2181,mesos5.domain.xyz:2181/mesos"
domain = "mesos.localhost"
ExposedByDefault = true
# IPSources = "docker"
IPSources = "mesos"

[marathon]
endpoint = "http://mesos3.domain.xyz:8080,mesos4.domain.xyz:8080,mesos5.domain.xyz:8080"
domain = "marathon.localhost"
marathonLBCompatibility = true
exposedByDefault = true
groupsAsSubDomains = true
```

### What did you do?
We run two tasks / Services to test this setup. One as docker container and a second one as Universal containerizer.

docker:
``` json
{
    "container": {
        "type": "DOCKER",
        "docker": {
            "forcePullImage": true,
            "image": "hub.binky.domain.xyz/trashcan/fixed:0.1",
            "privileged": false,
            "network": "USER",
            "portMappings": [
              {
                "containerPort": 80,
                "protocol": "tcp",
                "name": null,
                "labels": null
              }
            ]
        }
    },
    "cpus": 0.1,
    "id": "/traefik/docker-1",
    "instances": 1,
    "minimumHealthCapacity": 1,
    "acceptedResourceRoles": ["*"],
    "mem": 128,
    "labels": {
      "HAPROXY_0_VHOST": "docker-1.traefik.binky.domain.xyz",
      "HAPROXY_GROUP": "internal"
    },    
    "ipAddress": {
        "networkName": "my-calico-net",
        "labels": {
          "app": "test",
          "group": "development"
        }
    }
}
```

universal containerizer:
```json
{
    "container": {
        "type": "MESOS",
        "docker": {
[...]

PR #1398, opened 2017-04-07
Fix default timeouts for Marathon provider.

timoreimann:
The timeouts were given without a unit, which caused nanoseconds scale to be applied when we switched the type from `int` to `flaeg.Duration`.

Fixes a regression introduced by #1350.

Discovered along #1390.


timoreimann:
Marking WIP because I still want to add an integration test.

timoreimann:
I went down a huge rabbit hole trying to add an integration test. I decided to carve out the test into a separate PR #1406 and reduced this one.

@emilevauge can we approve the PR without the extended test for now? I'd like the fix to be in first, will continue to work on the test.

timoreimann:
Ping @emilevauge

trecloux:
LGTM

Issue #1379, opened 2017-04-04
client-go 1.5 not compatible with kubernetes 1.6


patte:
### What version of Traefik are you using (`traefik version`)?
1.2.1 (or 1.2.0-rc2)

### What is your environment & configuration (arguments, toml...)?
Kubernetes Cluster (v. 1.6.1)

### What did you do?
Deploy Traefik as described in the docs. Because my cluster is self hosted, I use `hostNetwork: true` see config

### What did you expect to see?
Traefik up and running happily... as it was the case with Kubernetes 1.5.3

### What did you see instead?
No backend / frontend showing up in the dashboard. Pages not serving. Errors in the logs. [client-go](https://github.com/kubernetes/client-go) tells me, client-go 1.5 is not compatible with Kubernetes 1.6. True?

### If applicable, please paste the log output in debug mode (`--debug` switch)
kubectl logs traefik-ingress-controller-vn21d
```
ERROR: logging before flag.Parse: E0404 07:57:02.404571       1 reflector.go:214] github.com/containous/traefik/vendor/k8s.io/client-go/1.5/tools/cache/reflector.go:109: Failed to list *v1.Service: the server does not allow access to the requested resource (get services)
ERROR: logging before flag.Parse: E0404 07:57:02.404589       1 reflector.go:214] github.com/containous/traefik/vendor/k8s.io/client-go/1.5/tools/cache/reflector.go:109: Failed to list *v1.Endpoints: the server does not allow access to the requested resource (get endpoints)
ERROR: logging before flag.Parse: E0404 07:57:02.404719       1 reflector.go:214] github.com/containous/traefik/vendor/k8s.io/client-go/1.5/tools/cache/reflector.go:109: Failed to list *v1beta1.Ingress: the server does not allow access to the requested resource (get ingresses.extensions)
```

traefik.yaml
```
# We run this with hostnetwork as a workaround for
# hostPort not beeing available with CNI networking
# https://github.com/kubernetes/kubernetes/issues/23920
# https://github.com/kubernetes/kubernetes/issues/31307
# https://github.com/containernetworking/cni/issues/46
kind: DaemonSet
apiVersion: extensions/v1beta1
metadata:
  name: traefik-ingress-controller
  labels:
    k8s-app: traefik-ingress-lb
    kubernetes.io/cluster-service: "true"
spec:
  selector:
    matchLabels:
      k8s-app: traefik-ingress-lb
  template:
    metadata:
      labels:
[...]

PR #1368, opened 2017-03-31
License 2017, Træfɪk => Træfik

emilevauge:
This PR updates license year to 2017 and replaces  `Træfɪk` by `Træfik` in documentation.

Signed-off-by: Emile Vauge <emile@vauge.com>

PR #1404, opened 2017-04-08
Add documentation for k8s RBAC configuration

aolwas:
This PR add documentation and manifest example to use Traefik with k8s 1.6+ and RBAC enabled

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -67,6 +67,64 @@ To deploy Træfɪk to your cluster start by submitting the deployment to the clu
 ```sh
 kubectl apply -f examples/k8s/traefik.yaml
 ```
+### Role Based Access Control configuration (optional)
+
+Kubernetes introduces [Role Based Access Control (RBAC)](https://kubernetes.io/docs/admin/authorization/) in 1.6+ to allow fine-grained control
+of Kubernetes resources and api.
+
+If your cluster is configured with RBAC, you need to authorize Traefik to use
+kubernetes API using ClusterRole, ServiceAccount and ClusterRoleBinding resources:
+
+```yaml
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: traefik-ingress-controller
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+      - services
+      - endpoints
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - extensions
+    resources:
+      - ingresses
+    verbs:
+      - get
+      - list
+      - watch
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: traefik-ingress-controller
+  namespace: kube-system
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: traefik-ingress-controller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: traefik-ingress-controller
+subjects:
+- kind: ServiceAccount
+  name: traefik-ingress-controller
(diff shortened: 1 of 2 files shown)


docs/user-guide/kubernetes.md
```diff
@@ -67,6 +67,64 @@ To deploy Træfɪk to your cluster start by submitting the deployment to the clu
...
 ```
+### Role Based Access Control configuration (optional)
+
+Kubernetes introduces [Role Based Access Control (RBAC)](https://kubernetes.io/docs/admin/authorization/) in 1.6+ to allow fine-grained control
+of Kubernetes resources and api.
+
+If your cluster is configured with RBAC, you need to authorize Træfɪk to use
+kubernetes API using ClusterRole, ServiceAccount and ClusterRoleBinding resources:
+
+```yaml
```
timoreimann: If the example is identical to the inline definition, could we just point to the example instead of repeating it?

(This will avoid getting out of sync.)

docs/user-guide/kubernetes.md
```diff
@@ -67,6 +67,64 @@ To deploy Træfɪk to your cluster start by submitting the deployment to the clu
 ```sh
 kubectl apply -f examples/k8s/traefik.yaml
 ```
+### Role Based Access Control configuration (optional)
+
+Kubernetes introduces [Role Based Access Control (RBAC)](https://kubernetes.io/docs/admin/authorization/) in 1.6+ to allow fine-grained control
+of Kubernetes resources and api.
+
+If your cluster is configured with RBAC, you need to authorize Træfɪk to use
```
timoreimann: Aren't we using the normal spelling for Traefik these days  instead of the special ae combo? @emilevauge?

emilevauge: Indeed, @aolwas could you replace `Træfɪk` by `Træfik` please ?

aolwas: I used this spelling because I found it elsewhere in the document. Do you want me to also fix the spelling in the whole document ?

docs/user-guide/kubernetes.md
aolwas: by inline definition, you mean the official k8s documentation ?

docs/user-guide/kubernetes.md
emilevauge: @aolwas thanks, but this is already done in https://github.com/containous/traefik/pull/1368 :)

docs/user-guide/kubernetes.md
timoreimann: No I mean the part starting in line 67. It seems to be identical to the example file you also included in this PR.

errm: its a pain, but I feel there is some value in having things both inline (so the documentation can be read) and as a file so the example is runnable...

The ideal thing would be to have something inline the contents of a file into the rendered documentation, I don't know if the tool we are using supports that though?

timoreimann: Unfortunately, it [does not seem to be supported](https://github.com/github/markup/issues/172#issuecomment-33241601).

Personally, I don't mind the flow break. After all, that's what hyperlinking is all about. :-)

That said, I'm not overly enthusiastic on this point. A bit of duplication is acceptable for me here.

aolwas: @timoreimann You're right, the code in the documentation embedded as is in the exemple file. I asked myself if I should provide only the example file but it seems to me more confortable for the reader to have directly the code in the doc. 

Tell me what you think is the best.

timoreimann: I'm okay with keeping the inline spec, especially since you and @errm are in favor of it. Happy to let myself get overruled by the majority. :-)

As soon as you address [the other point](https://github.com/containous/traefik/pull/1404#discussion_r111105543), we can merge.

timoreimann:
Tests green, merging!

timoreimann:
Refs #1379.

Issue #156, opened 2016-01-08
More detailed metrics


keis:
It would be great if traefik could provide some more detailed metrics, e.g
- Response time quantile to avoid skew from one or two outliers. (like how [prometheus](http://prometheus.io/docs/practices/histograms/#quantiles) does it)
- Breakdown by backend
  - request count
  - response time

thawkins:
Definatly +1 on this.... 

I would also add response status to that list.

keis:
Most of the heavy lifting for this is done by the logger middleware now. It should be pretty simple to add some prometheus counters to that. Not sure if that's the best place for doing it or if prometheus is the right tool. I do quite like it though :)

samber:
I think most interesting metrics are:

- per entrypoint:
  - Number of active connections

- per backend type (consul catalog, file, etcd, kubernetes...):
  - Number of successful backend updates
  - Number of failed backend updates
  - number catched updates
  - number effective reload

- per service:
  - circuit breaker:
    - number of backends available
    - number failing backends

  - frontend:
    - response time (including Traefik processing)
    - number successful requests (1xx, 2xx, 3xx)
    - number failing requests (4xx, 5xx)
    - request size
    - response size

  - per backend:
    - response time (excluding Traefik processing)
    - number successful requests (1xx, 2xx, 3xx)
    - number failing requests (4xx, 5xx)
    - retry count
    - circuit breaker:
      - is available (bool)
      - expression (example: NetworkErrorRatio() > 0.5)
      - number network error

Prometheus is probably the more "trendly" time series db. IMO, that's why we should integrate a ```/metrics``` route in the web provider.

Prometheus can aggregate backends stats to get avg or sum of each metrics, then adding "total" counts would only be used for ```/health```requests.

thawkins:
+1 on prometheus integration,  prometheus will be out of the box on ge next release of kubenetes. Having good integration would be excellent.

samber:
Last week, I made a standalone Prometheus exporter: https://github.com/iadvize/traefik-exporter

I will combine that work with the Traefik binary very soon.

As example of limitation with an external exporter, we cannot use Prometheus summaries to mesure the duration of 1% slowest requests.

therc:
[...]

PR #1405, opened 2017-04-09
Add more prometheus metrics

jereksel:
This MR implement more metrics mentioned [here](https://github.com/containous/traefik/issues/156#issuecomment-221536251).

It implements more detailed backend metrics (it now includes backend server), frontend metrics (see below) and config reload metrics.

Unfortunatelly there doesn't seem to be a clean way to get more information about frontend without rewriting route management.

I'm also not sure about frontend/backend times. I've included debug start/stop messages for backend and frontend and these are the results:

```
backend start
backend start
frontend start
DEBU[2017-04-09T21:03:27+02:00] Round trip: http://httpbin.org/, code: 200, duration: 165.885424ms 
frontend stop
backend stop
backend stop
```
Which is very weird because backend is invoked before frontend and I don't know how to add time metrics that will correctly calculate frontend and backend response times.


timoreimann:
@jereksel you cannot change vendored code directly; you have to make a PR against upstream and afterwards revendor the updated library in Traefik (i.e., use glide to bump our referenced version).

For the time being though, I think it's okay to pretend this is the oxy change we want, do the review, and eventually update containous/oxy (our fork of vulcand/oxy) should we all agree to the suggested path.

server.go
```diff
@@ -219,17 +230,22 @@ func (server *Server) listenProviders(stop chan bool) {
...
 				return
 			}
+			var labels []string
 			server.defaultConfigurationValues(configMsg.Configuration)
 			currentConfigurations := server.currentConfigurations.Get().(configs)
 			jsonConf, _ := json.Marshal(configMsg.Configuration)
 			log.Debugf("Configuration received from provider %s: %s", configMsg.ProviderName, string(jsonConf))
 			if configMsg.Configuration == nil || configMsg.Configuration.Backends == nil && configMsg.Configuration.Frontends == nil {
 				log.Infof("Skipping empty Configuration for provider %s", configMsg.ProviderName)
+				labels = []string{"state", "failed", "type", configMsg.ProviderName}
```
timoreimann: The only parameter that changes between the different occasions where we assign the slice is the state. Can we initialize the slice with the other parameters in line 236 right away and just complete the state in the subsequent lines, respectively?

middlewares/transformer.go
```diff
@@ -0,0 +1,20 @@
+package middlewares
+
+import (
+	"net/http"
+)
+
+//Transformer is a hacky way to get metrics middleware with frontend
```
timoreimann: Not sure I get why we need the transformer. Could you explain please?

middlewares/metrics.go
```diff
@@ -19,13 +24,41 @@ type Metrics interface {
 // given Metrics implementation to expose and monitor Traefik metrics
 type MetricsWrapper struct {
 	Impl Metrics
+
+	//frontend/backend
```
timoreimann: The comment should maybe be a bit more elaborate.

Also, space between slash and first word.

ldez:
@jereksel do you the time to update this PR before the upcoming feature freeze for the 1.3 release?

middlewares/metrics.go
```diff
@@ -34,9 +67,21 @@ func NewMetricsWrapper(impl Metrics) *MetricsWrapper {
...
+	var state string
+	if prw.StatusCode() < 400 {
+		state = "Successful"
+	} else {
+		state = "Failing"
+	}
+
+	labelsStatus := []string{"state", state, "method", r.Method, "name", m.getName(r), "type", m.typ}
 	m.Impl.getReqsCounter().With(labels...).Add(1)
+	m.Impl.getReqsStatusCounter().With(labelsStatus...).Add(1)
```
jereksel: Sure, I was basing it on rancher-exporter that splits these metrics. But having them together is indeed better idea.

middlewares/metrics.go
```diff
@@ -19,13 +24,41 @@ type Metrics interface {
 // given Metrics implementation to expose and monitor Traefik metrics
 type MetricsWrapper struct {
 	Impl Metrics
+
+	//frontend/backend
+	typ     string
```
jereksel: I thought about naming it `type`, but it's reserved keyword. Phase sound good thought.

jereksel:
Sure, I'll try to fix this MR today.

middlewares/transformer.go
```diff
@@ -0,0 +1,20 @@
...
+	next    http.Handler
+	metrics *MetricsWrapper
+}
+
+//NewTransformer creates new proxy for given Metrics and Handler
+func NewTransformer(next http.Handler, metrics *MetricsWrapper) *Transformer {
+	return &Transformer{next, metrics}
+}
+
+func (sb *Transformer) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
```
jereksel: I copied it from another file :smile:

jereksel:
@Idez @timoreimann Maybe I'll scrap this and redo it with https://github.com/containous/traefik/pull/1408/ and https://github.com/containous/traefik/pull/1485/

ldez:
@jereksel simply rebase your branch, like that you'll already have the content of the #1408

timoreimann:
Yep, rebase is the way to go.

We have also restructured/moved some packages. (For instance, `/server.go` now lives in `/server/server.go`.) My git client got a bit confused by that and tried to merge files into `integration`. I managed to solve this by specifying the rename threshold like this: `git rebase -X find-renames=25 upstream/master`.

jereksel:
I think I'll wait for #1485 - I think I have an idea how to combine it with Prometheus metrics.

PR #1406, opened 2017-04-10
Improve Marathon integration tests.

timoreimann:
This is a continuation of #1398.

The tests do not succeed yet, apparently because the external network isn't applied.

### DIFF
--- a/integration/marathon_test.go
+++ b/integration/marathon_test.go
@@ -1,53 +1,129 @@
 package integration
 
 import (
+	"fmt"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/containous/traefik/integration/try"
+	"github.com/containous/traefik/types"
+	marathon "github.com/gambol99/go-marathon"
 	"github.com/go-check/check"
 	checker "github.com/vdemeester/shakers"
 )
 
+const (
+	containerNameMesosSlave = "mesos-slave"
+	containerNameMarathon   = "marathon"
+)
+
 // Marathon test suites (using libcompose)
-type MarathonSuite struct{ BaseSuite }
+type MarathonSuite struct {
+	BaseSuite
+	marathonURL string
+}
 
 func (s *MarathonSuite) SetUpSuite(c *check.C) {
 	s.createComposeProject(c, "marathon")
 	s.composeProject.Start(c)
 
-	// FIXME Doesn't work...
-	//// "github.com/gambol99/go-marathon"
-	//config := marathon.NewDefaultConfig()
-	//
-	//marathonClient, err := marathon.NewClient(config)
-	//if err != nil {
-	//	c.Fatalf("Error creating Marathon client. %v", err)
-	//}
-	//
-	//// Wait for Marathon to elect itself leader
-	//err = try.Do(30*time.Second, func() error {
-	//	leader, err := marathonClient.Leader()
-	//
-	//	if err != nil || len(leader) == 0 {
-	//		return fmt.Errorf("Leader not found. %v", err)
-	//	}
-	//
-	//	return nil
-	//})
-	//
-	//c.Assert(err, checker.IsNil)
+	marathonIPAddr := s.composeProject.Container(c, containerNameMarathon).NetworkSettings.IPAddress
+	c.Assert(marathonIPAddr, checker.Not(checker.HasLen), 0)
+	s.marathonURL = "http://" + marathonIPAddr + ":8080"
+
+	// Wait for Marathon readiness prior to creating the client so that we
+	// don't run into the "all cluster members down" state right from the
+	// start.
+	err := try.GetRequest(s.marathonURL+"/v2/leader", 1*time.Minute, try.StatusCodeIs(http.StatusOK))
+	c.Assert(err, checker.IsNil)
+
+	// Add entry for Mesos slave container IP address in the hosts file so
+	// that Traefik can properly forward traffic.
+	// This is necessary as long as we are still using the docker-compose v1
+	// spec. Once we switch to v2 or higher, we can have both the test/builder
+	// container and the Mesos slave container join the same custom network and
+	// enjoy DNS-discoverable container host names.
+	mesosSlaveIPAddr := s.composeProject.Container(c, containerNameMesosSlave).NetworkSettings.IPAddress
+	c.Assert(mesosSlaveIPAddr, checker.Not(checker.HasLen), 0)
+	err = s.extendDockerHostsFile(containerNameMesosSlave, mesosSlaveIPAddr)
+	c.Assert(err, checker.IsNil)
 }
 
-func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
-	cmd, _ := s.cmdTraefik(withConfigFile("fixtures/marathon/simple.toml"))
+// extendDockerHostsFile extends the hosts file (/etc/hosts) by the given
+// host/IP address mapping if we are running inside a container.
+func (s *MarathonSuite) extendDockerHostsFile(host, ipAddr string) error {
+	const hostsFile = "/etc/hosts"
+
+	// Determine if the run inside a container. The most reliable way to
+	// do this is to inject an indicator, which we do in terms of an
+	// environment variable.
+	// (See also https://groups.google.com/d/topic/docker-user/JOGE7AnJ3Gw/discussion.)
+	if os.Getenv("CONTAINER") == "DOCKER" {
+		// We are running inside a container -- extend the hosts file.
+		file, err := os.OpenFile(hostsFile, os.O_APPEND|os.O_WRONLY, 0600)
+		if err != nil {
+			return err
+		}
+		defer file.Close()
+
+		if _, err = file.WriteString(fmt.Sprintf("%s\t%s\n", ipAddr, host)); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (s *MarathonSuite) TestConfigurationUpdate(c *check.C) {
+	// Start Traefik.
+	file := s.adaptFile(c, "fixtures/marathon/simple.toml", struct {
+		MarathonURL string
+	}{s.marathonURL})
+	defer os.Remove(file)
+	cmd, output := s.cmdTraefik(withConfigFile(file))
 	err := cmd.Start()
 	c.Assert(err, checker.IsNil)
 	defer cmd.Process.Kill()
 
-	// TODO validate : run on 80
-	// Expected a 404 as we did not configure anything
-	err = try.GetRequest("http://127.0.0.1:8000/", 500*time.Millisecond, try.StatusCodeIs(http.StatusNotFound))
-
+	// Wait for Traefik to turn ready.
+	err = try.GetRequest("http://127.0.0.1:8000/", 2*time.Second, try.StatusCodeIs(http.StatusNotFound))
 	c.Assert(err, checker.IsNil)
+
+	// Prepare Marathon client.
+	config := marathon.NewDefaultConfig()
+	config.URL = s.marathonURL
+	client, err := marathon.NewClient(config)
+	c.Assert(err, checker.IsNil)
+
+	// Show the Traefik log if any assertion fails. If the entire test runs
+	// to a successful completion, we flip the flag at the very end and don't
+	// display anything.
+	showTraefikLog := true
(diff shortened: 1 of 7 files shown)


timoreimann:
According to @vdemeester, it's because the version of libcompose we use does not support docker-compose v2 yet.

timoreimann:
This PR is blocked because we need to first update libcompose (through libkermit) in order to have docker-compose version 2 which brings custom networking support. Otherwise, we cannot talk to the Marathon instances from the integration tests.

Updating libcompose is non-trivial, once again due to glide and our two-glide-manifests setup.

timoreimann:
Note to myself: Also update go-marathon integration dependency.

timoreimann:
Turns out [networking aliasing isn't properly working in libcompose](https://github.com/docker/libcompose/issues/461), which is a blocker for the required inter-container communication on the Marathon docker-compose manifest.

Our very own @vdemeester is on it. :-)

timoreimann:
Currently being show-stopped by https://github.com/docker/libcompose/issues/484. Sending all bug reports directly to @vdemeester. :)

ldez:
How you have retry since https://github.com/docker/libcompose/pull/483?

timoreimann:
@ldez I did retry, and that helped us move forward a bit. However, I then ran into  docker/libcompose#484. That's still the latest blocker.

timoreimann:
Trying if switching to host network helps. Can only debug this properly on the CI.

timoreimann:
Host networking didn't help because we can't (easily?) reach Marathon running on the host from inside our build container.

Fortunately, I finally got this right by switching back to using the docker-compose v1 spec. The key advantage being that in v1, both our builder container and all compose services share the same bridged network which allows them to talk to each other.

That said, the change doesn't come for free: We now have to determine the IP addresses for container we want to speak with (Marathon, Mesos slave) and inject them at the appropriate locations. For the Traefik binary itself, this means we have to add an entry to `/etc/hosts` (unless we're running outside of a container where we don't want to mess with the hosts file) since the Marathon provider uses the task host name as the backend server. I tried tweaking the Marathon configuration to force a different mode, but everything I did ended up being other as least as complicated or too remote from how Marathon in actually used in the real world.

So overall, the approach isn't pretty. On the bright side, it will become obsolete once we can switch to the docker-compose v2 spec where all containers can share the custom network and DNS discoverability is available. If anyone has any further idea for simplification, please step forward. :)

Note that I kept the libcompose dependency upgrade in even though it's not strictly needed anymore: Since we should strive moving to version of libcompose that eventually supports v2 to the extent we need it, I think it's okay to keep it in.

timoreimann:
@containous/traefik PTAL!

integration/marathon_test.go
```diff
@@ -48,6 +75,60 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
 	// TODO validate : run on 80
 	// Expected a 404 as we did not configure anything
 	err = try.GetRequest("http://127.0.0.1:8000/", 500*time.Millisecond, try.StatusCodeIs(http.StatusNotFound))
+	c.Assert(err, checker.IsNil)
+}
+
+func (s *MarathonSuite) TestConfigurationUpdate(c *check.C) {
+	marathonURL := "http://" + s.getContainerIPAddr(c, containerNameMarathon) + ":8080"
+	fmt.Printf("Using Marathon URL %s\n", marathonURL)
```
Juliens: We really need this log ?

integration/marathon_test.go
```diff
@@ -48,6 +75,60 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
...
+
+	// Wait for Marathon readiness prior to creating the client so that we
+	// don't run into the "all cluster members down" state right from the
+	// start.
+	fmt.Println("Waiting for Marathon to become ready")
```
Juliens: We really need this log ?

integration/marathon_test.go
```diff
@@ -48,6 +75,60 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
...
+	app := marathon.NewDockerApplication().
+		Name("/whoami").
+		CPU(0.1).
+		Memory(32).
+		AddLabel(types.LabelFrontendRule, "PathPrefix:/service")
+	app.Container.Docker.Bridged().
+		Expose(80).
+		Container("emilevauge/whoami")
+
+	fmt.Println("Deploying test application")
```
Juliens: We really need this log ?

integration/marathon_test.go
```diff
@@ -48,6 +75,60 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
...
+	deploy, err := client.UpdateApplication(app, false)
+	c.Assert(err, checker.IsNil)
+	fmt.Println("Waiting for Deployment to complete")
```
Juliens: We really need this log ?

integration/marathon_test.go
```diff
@@ -48,6 +75,60 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
...
+	c.Assert(client.WaitOnDeployment(deploy.DeploymentID, 2*time.Minute), checker.IsNil)
 
+	fmt.Println("Querying application via Traefik")
```
Juliens: We really need this log ?

integration/marathon_test.go
```diff
@@ -48,6 +75,60 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
...
+	c.Assert(err, checker.IsNil)
+	defer cmd.Process.Kill()
+
+	// Prepare Marathon client.
+	config := marathon.NewDefaultConfig()
+	config.URL = marathonURL
+	client, err := marathon.NewClient(config)
+	c.Assert(err, checker.IsNil)
+
+	showTraefikLog := true
```
Juliens: you always want to show traefik logs ?

integration/marathon_test.go
```diff
@@ -48,6 +75,60 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
...
+
+	file := s.adaptFile(c, "fixtures/marathon/simple.toml", struct {
+		MarathonURL string
+	}{marathonURL})
+	defer os.Remove(file)
+	cmd, output := s.cmdTraefik(withConfigFile(file))
+	err = cmd.Start()
+	c.Assert(err, checker.IsNil)
+	defer cmd.Process.Kill()
+
```
Juliens: Wait for traefik is Up and Running ?

timoreimann: Good catch -- added.

integration/marathon_test.go
timoreimann: See the last line of the test method: I flip the flag if the test runs to completion successfully so that we don't show the logs if everything went fine.

timoreimann:
Addressed all comments -- PTAL again.

integration/marathon_test.go
ldez: maybe you can add a small comment to explains that in the code.

integration/marathon_test.go
```diff
@@ -1,42 +1,69 @@
...
-	//
-	//	if err != nil || len(leader) == 0 {
-	//		return fmt.Errorf("Leader not found. %v", err)
-	//	}
-	//
-	//	return nil
-	//})
-	//
-	//c.Assert(err, checker.IsNil)
+func (s *MarathonSuite) getContainerIPAddr(c *check.C, name string) string {
```
timoreimann: I can move it if you like; just to be sure though: did you notice that I need it once for the Mesos slave and another time for Marathon? That'd be 4 lines, so I wanted to avoid code duplication. (In fact, the function would be reusable if we turned `s.composeProject` into a function parameter, which would allow us to apply it in a lot of other tests as well.)

Personally, I also find `getContainerIPAddr(c, name)` more readable and comprehendible compared to `s.composeProject.Container(c, name).NetworkSettings.IPAddress`.

integration/marathon_test.go
```diff
@@ -48,6 +75,63 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
...
+	c.Assert(err, checker.IsNil)
+}
+
+func (s *MarathonSuite) TestConfigurationUpdate(c *check.C) {
+	marathonURL := "http://" + s.getContainerIPAddr(c, containerNameMarathon) + ":8080"
+
+	// Wait for Marathon readiness prior to creating the client so that we
+	// don't run into the "all cluster members down" state right from the
+	// start.
+	err := try.GetRequest(marathonURL+"/ping", 1*time.Minute, try.StatusCodeIs(http.StatusOK))
```
timoreimann: Good idea (on both). Done.

integration/marathon_test.go
```diff
@@ -48,6 +75,63 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
...
+		AddLabel(types.LabelFrontendRule, "PathPrefix:/service")
+	app.Container.Docker.Bridged().
+		Expose(80).
+		Container("emilevauge/whoami")
+
+	// Deploy the test application.
+	deploy, err := client.UpdateApplication(app, false)
+	c.Assert(err, checker.IsNil)
+	// Wait for deployment to complete.
+	c.Assert(client.WaitOnDeployment(deploy.DeploymentID, 2*time.Minute), checker.IsNil)
```
timoreimann: Cut in half. (Although a higher timeout doesn't really hurt us in the regular case where things go fast and possibly saves us if there's higher latency for whatever reason.)

integration/marathon_test.go
```diff
@@ -48,6 +75,63 @@ func (s *MarathonSuite) TestSimpleConfiguration(c *check.C) {
...
 
+	// Query application via Traefik.
+	err = try.GetRequest("http://127.0.0.1:8000/service", 1*time.Minute, try.StatusCodeIs(http.StatusOK))
```
timoreimann: Cut down to 30 seconds.

timoreimann:
I also just realized that the `TestSimpleConfiguration` test in its current form is pretty useless: The Marathon configuration is actually broken in that test as I don't replace the `{{.MarathonURL}}` template parameter there. And yet, Traefik starts up successfully (because it's always trying hard to do its best).

Simplifying the configuration even further doesn't really give us any automatic guarantee. As there's no way to reliably tell if we got a 404 because we haven't added a Marathon application yet, I removed the test. It doesn't add any value to me. WDYT?

ldez:
@timoreimann I agree `TestSimpleConfiguration` is useless. I think you can remove this test (which tests nothing)

integration/marathon_test.go
ldez: I understand you can create a function `getContainerIPAddr` in `integration_test.go`

```golang
func (s *BaseSuite) getContainerIPAddr(c *check.C, name string) string {
	ipAddr := s.composeProject.Container(c, name).NetworkSettings.IPAddress
	c.Assert(ipAddr, checker.Not(checker.HasLen), 0)
	return ipAddr
}
```

But I think this refactor must be done in another PR.

timoreimann:
@ldez @Juliens PTAL again. All comments should be addressed.

integration/marathon_test.go
```diff
@@ -1,53 +1,128 @@
...
+
+	// Add entry for Mesos slave container IP address in the hosts file so
+	// that Traefik can properly forward traffic.
+	// This is necessary as long as we are still using the docker-compose v1
+	// spec. Once we switch to v2 or higher, we can have both the test/builder
+	// container and the Mesos slave container join the same custom network and
+	// enjoy DNS-discoverable container host names.
+	mesosSlaveIPAddr := s.composeProject.Container(c, containerNameMesosSlave).NetworkSettings.IPAddress
+	c.Assert(mesosSlaveIPAddr, checker.Not(checker.HasLen), 0)
+	c.Assert(s.extendDockerHostsFile(containerNameMesosSlave, mesosSlaveIPAddr), checker.IsNil)
```
Juliens: I prefer when we split the call and the assert (more readable for me, but I can live with it)
```golang
err := s.extendDockerHostsFile(containerNameMesosSlave, mesosSlaveIPAddr)
c.Assert(err, checker.IsNil)
```

timoreimann: Separating the call and the check would actually be more consistent with our existing style.

Updated.

PR #1293, opened 2017-03-14
Configurable access logger with JSON support

rickb777:
This is a new access logger derived from (and replacing) the old one. It exists in its own accesslog package under middlewares and has similar-ish net/http filters to the existing solution. Its first concern is to implement the existing extended Common Log Format behaviour, to provide a clean upgrade path for existing users.

But it's now split into pluggable parts, the important part being its formatter (the CLF formatter being the default). A new JSON formatter exists.

The file writing part is termed the log appender and is slightly enhanced: it allows buffered writing, or gzipped writing, and/or asynchronous writing via a channel/goroutine.

The final area of change is to the configuration: new settings are needed to control the new features. This is done with a new AccessLog type in types.go.

Notice that there are now three middlewares:

the first is the 'main' logger that tracks each request, builds log events, and despatches these to the log appender;
the second inserts knowledge about the frontend into the log-data context;
the third inserts knowledge about the backend into the log-data context.
There had previously been an error in tracking the frontend data: it operated on the incorrect assumption that there is a 1:N mapping from frontends to backends.

PR #1408, opened 2017-04-10
New access logger

rjshep:
As requested in [#1293](https://github.com/containous/traefik/pull/1293), splitting new access logger into smaller commits to aid PR process.

This first PR adds the new access logger alongside the existing one, allowing information to be captured but at this stage not output.  Additional functionality will follow in subsequent PRs.


middlewares/accesslog/logger.go
```diff
@@ -0,0 +1,147 @@
...
+	"sync/atomic"
+	"time"
+)
+
+type key int
+
+const (
+	// DataTableKey is the key within the request context used to
+	// store the Log Data Table
+	DataTableKey key = iota
```
timoreimann: Is there a particular reason why we use 0/iota as the key here? Why not a more meaningful, human-readable string?

To the least, this would make debugging easier.

middlewares/accesslog/logger.go
```diff
@@ -0,0 +1,147 @@
...
+
+	logDataTable := &LogData{core, req.Header, nil, nil}
+	core[StartUTC] = now
+	core[StartLocal] = now.Local()
+
+	r2 := req.WithContext(context.WithValue(req.Context(), DataTableKey, logDataTable))
+
+	var crr *captureRequestReader
+	if req.Body != nil {
+		crr = &captureRequestReader{req.Body, 0}
```
timoreimann: Again, my preference goes towards a key-indexed definition.

middlewares/accesslog/logger.go
```diff
@@ -0,0 +1,147 @@
...
+	logDataTable.DownstreamResponse = crw.Header()
+	l.logTheRoundTrip(logDataTable, crr, crw)
+}
+
+// Close closes the Logger (i.e. the file etc).
+func (l *LogHandler) Close() error {
+	return nil
+}
+
+func silentSplitHostPort(value string) (string, string) {
```
timoreimann: Can we name the return parameters `host` and `port`, respectively, for documentary reasons?

middlewares/accesslog/logdata.go
```diff
@@ -0,0 +1,124 @@
...
+	allCoreKeys[RequestLine] = struct{}{}
+	allCoreKeys[OriginStatusLine] = struct{}{}
+	allCoreKeys[DownstreamStatusLine] = struct{}{}
+	allCoreKeys[GzipRatio] = struct{}{}
+	allCoreKeys[StartLocal] = struct{}{}
+	allCoreKeys[Overhead] = struct{}{}
+}
+
+// CoreLogData holds the fields computed from the request/response.
+type CoreLogData map[string]interface{}
```
timoreimann: I notice we have to do quite a bit of type casting due to the `interface{}` mapee.

Couldn't we just use a strongly typed struct instead of a map?

middlewares/accesslog/logger.go
```diff
@@ -0,0 +1,147 @@
...
+		core[RequestContentSize] = crr.count
+	}
+
+	core[DownstreamStatus] = crw.Status()
+	core[DownstreamStatusLine] = fmt.Sprintf("%03d %s", crw.Status(), http.StatusText(crw.Status()))
+	core[DownstreamContentSize] = crw.Size()
+	if original, ok := core[OriginContentSize]; ok {
+		o64 := original.(int64)
+		if o64 != crw.Size() {
+			// n.b divide-by-zero is tolerated here and dealt with elsewhere as appropriate
```
timoreimann: What do you mean by "is tolerated here"? Wouldn't the code panic if `crw.Size()` turned out to be zero?

middlewares/accesslog/save_backend.go
```diff
@@ -0,0 +1,71 @@
...
+	next        http.Handler
+	backendName string
+}
+
+// NewSaveBackend creates a SaveBackend handler.
+func NewSaveBackend(next http.Handler, backendName string) http.Handler {
+	return &SaveBackend{next, backendName}
+}
+
+func copyHeaders(src http.Header) http.Header {
```
timoreimann: Our vulcand/oxy dependency already [brings a header copying functionality](https://github.com/containous/traefik/blob/d8c21639f79573f5e0e48bf509bf296f7916c0d9/vendor/github.com/vulcand/oxy/utils/netutils.go#L109). Not sure if you can use it though since it does not override headers.

middlewares/logger_test.go
```diff
@@ -64,7 +66,15 @@ func TestLogger(t *testing.T) {
 		},
 	}
 
-	logger.ServeHTTP(&logtestResponseWriter{}, r, LogWriterTestHandlerFunc)
+	// Temporary - until new access logger is fully implemented
+	// create the data table and populate frontend and backend
+	core := make(accesslog.CoreLogData)
+	logDataTable := &accesslog.LogData{core, r.Header, nil, nil}
```
timoreimann: Keyed index.

middlewares/accesslog/capture_response_writer.go
```diff
@@ -0,0 +1,55 @@
...
+	if f, ok := crw.rw.(http.Flusher); ok {
+		f.Flush()
+	}
+}
+
+func (crw *captureResponseWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	if h, ok := crw.rw.(http.Hijacker); ok {
+		return h.Hijack()
+	}
+	return nil, nil, fmt.Errorf("Not a Hijacker: %T", crw.rw)
```
rjshep: Changed

middlewares/accesslog/logger.go
rjshep: Changed to string - more descriptive as you say.

middlewares/accesslog/logger.go
```diff
@@ -0,0 +1,147 @@
...
+// data as the request passes through the middleware chain.
+func GetLogDataTable(req *http.Request) *LogData {
+	return req.Context().Value(DataTableKey).(*LogData)
+}
+
+func (l *LogHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next http.HandlerFunc) {
+	now := time.Now().UTC()
+	core := make(CoreLogData)
+
+	logDataTable := &LogData{core, req.Header, nil, nil}
```
rjshep: Changed

middlewares/accesslog/logger.go
rjshep: Changed

middlewares/accesslog/logger.go
rjshep: Changed

middlewares/accesslog/logger.go
rjshep: You're right it would panic.  I've changed the code to avoid divide by zero.

middlewares/accesslog/save_backend.go
rjshep: Code changed to use the existing header copying function.

middlewares/accesslog/logdata.go
rjshep: We went with the map so as to minimise the amount of effort needed to add additional log data in the future.

rjshep:
I've made the requested modifications and addressed the other comments.

In terms of tests,  yes there are more to come with the rest of the changes which will introduce an appender for the new logger as well as formatters for JSON and the common logging format.

The change to the existing access logger to get frontend and backend name is covered by an existing test modified in this PR.

middlewares/accesslog/save_backend.go
```diff
@@ -0,0 +1,63 @@
+package accesslog
+
+import (
+	"fmt"
+	"github.com/vulcand/oxy/utils"
```
timoreimann: Please group the non-stdlib imports at the bottom, separated by a blank line.

middlewares/logger_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"runtime"
 	"testing"
 
+	"context"
```
timoreimann: Please move this import right below the other stdlib ones and use the blank line to separate from the non-stdlib imports following.

middlewares/accesslog/logdata.go
timoreimann: Could you elaborate please? How would a strongly typed struct increase the effort? What extra work would be needed that the map does not demand?

Personally, I tend to accept a bit more effort if it brings us compile-time safety. I don't see the full picture yet to understand the different trade-offs though.

middlewares/accesslog/save_backend.go
rjshep: Done

middlewares/logger_test.go
rjshep: Done

rjshep:
The intention for the new access logger is to provide a configurable output format, where the format, ordering and naming of the fields is based on user-supplied configuration.  In order to do this simply the map was chosen so that, based on configuration, we can simply lookup the necessary values to output.

I appreciate that, as this PR doesn't include the log appenders, it's not easy to see why having the map is preferred.

If we choose a strongly typed struct we would (I believe) need to use either reflection or a large switch/case statement in order to achieve the same goal.  So I guess the question is whether this is preferable to using the existing map?

If based on the above you would still like to replace the map with a struct I'll look at doing that, though my preference is to retain the map :-)

timoreimann:
@rjshep thanks for elaborating. I'm okay with keeping the map for now. Let's revisit the decision once the follow-up PRs are coming in and enable a better understanding as you said.

middlewares/accesslog/capture_response_writer.go
```diff
@@ -0,0 +1,55 @@
+package accesslog
+
+import (
+	"bufio"
+	"fmt"
+	"net"
+	"net/http"
+)
+
```
emilevauge: Could you add this here ?
```
var (
	_ http.ResponseWriter = &captureResponseWriter{}
	_ http.Hijacker       = &captureResponseWriter{}
	_ http.Flusher        = &captureResponseWriter{}
	_ http.CloseNotifier  = &captureResponseWriter{}
)
```

rjshep: Added :-)

rjshep:
In terms of only adding the access log middleware when access logging is enabled in config, I think that is a good idea.  If it's okay with you I propose to make that change in the next PR which will replace the current access logger with the new one?

emilevauge:
@rjshep 
>I propose to make that change in the next PR which will replace the current access logger with the new one?

If you can make this PR in the next days, I'm not against that (we will freeze the code for v1.3 soon).

rjshep:
@emilevauge Yes I can submit the next PR in a few days.

middlewares/accesslog/logger.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (l *LogHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next http.HandlerFunc) {
+	now := time.Now().UTC()
+	core := make(CoreLogData)
+
+	logDataTable := &LogData{Core: core, Request: req.Header}
+	core[StartUTC] = now
+	core[StartLocal] = now.Local()
+
+	r2 := req.WithContext(context.WithValue(req.Context(), DataTableKey, logDataTable))
```
ldez: can you use a more explicit variable naming for `r2`

middlewares/accesslog/logger.go
```diff
@@ -0,0 +1,144 @@
...
+		r2.Body = crr
+	}
+
+	core[RequestCount] = nextRequestCount()
+	if req.Host != "" {
+		core[RequestAddr] = req.Host
+		core[RequestHost], core[RequestPort] = silentSplitHostPort(req.Host)
+	}
+	// copy the URL without the scheme, hostname etc
+	u := &url.URL{
```
ldez: can you use a more explicit variable naming for `u`

middlewares/accesslog/logger.go
```diff
@@ -0,0 +1,144 @@
...
+		Path:       req.URL.Path,
+		RawPath:    req.URL.RawPath,
+		RawQuery:   req.URL.RawQuery,
+		ForceQuery: req.URL.ForceQuery,
+		Fragment:   req.URL.Fragment,
+	}
+	us := u.String()
```
ldez: can you use a more explicit variable naming for `us`

middlewares/accesslog/logger.go
```diff
@@ -0,0 +1,144 @@
...
+	host, port, err := net.SplitHostPort(value)
+	if err != nil {
+		return value, "-"
+	}
+	return host, port
+}
+
+func usernameIfPresent(u *url.URL) string {
+	username := "-"
+	if u.User != nil {
```
ldez: can you use a more explicit variable naming for `u`

middlewares/accesslog/save_backend.go
```diff
@@ -0,0 +1,64 @@
...
+	next        http.Handler
+	backendName string
+}
+
+// NewSaveBackend creates a SaveBackend handler.
+func NewSaveBackend(next http.Handler, backendName string) http.Handler {
+	return &SaveBackend{next, backendName}
+}
+
+func (sb *SaveBackend) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
```
ldez: can you use a more explicit variable naming for `rw` and `r`

middlewares/accesslog/save_backend.go
```diff
@@ -0,0 +1,64 @@
...
+	next         http.Handler
+	frontendName string
+}
+
+// NewSaveFrontend creates a SaveFrontend handler.
+func NewSaveFrontend(next http.Handler, frontendName string) http.Handler {
+	return &SaveFrontend{next, frontendName}
+}
+
+func (sb *SaveFrontend) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
```
ldez: can you use a more explicit variable naming for `rw` and `r`

middlewares/accesslog/logger.go
rjshep: Done

middlewares/accesslog/logger.go
rjshep: Done

middlewares/accesslog/logger.go
rjshep: Done

middlewares/accesslog/logger.go
rjshep: Done

middlewares/accesslog/save_backend.go
rjshep: I've left this the same as it's consistent with all of the other middlewares :-)

middlewares/accesslog/save_backend.go
rjshep: I've left this the same as it's consistent with all of the other middlewares :-)

rjshep:
@emilevauge @ldez @timoreimann Hi all, is there anything further required before this PR can be merged?  I'm keen to press on with the next PR.

timoreimann:
As far as my review is concerned, I'm fine with the status quo.

PR #1413, opened 2017-04-11
Exclude vendor directories from Makefile SRCS variable.

timoreimann:
While it did not hurt to include those when gofmt'ing, it's not needed and may turn out to become a source of problems in the future should `SRCS` be used more extensively.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ TRAEFIK_ENVS := \
 	-e VERSION \
 	-e CODENAME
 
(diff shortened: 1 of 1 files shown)


timoreimann:
@containous/traefik PTAL.

timoreimann:
Looks like we already excluded vendor directories in the 1.2 branch recently merged in. Thus, we don't need this PR anymore.

Closing.

PR #1414, opened 2017-04-11
Fix wrong url with multiple rancher environments

bseng:
Scenario:
Traefik running inside rancher with admin api key for multiple environments.

In rancher I've got 2 environments.
image

In both environment there's a service with the same name.
image
image

When I'm using traefik with accesskey and secret option there is only one frontend url get created.
image

The issue is because environments on rancher ui is not environment in rancher api, but is project in api.
https://forums.rancher.com/t/api-key-for-all-environments/279/9
image

So this fix is to look at project instead of environment in api.

### DIFF
--- a/provider/rancher.go
+++ b/provider/rancher.go
@@ -286,11 +286,14 @@ func (provider *Rancher) Provide(configurationChan chan<- types.ConfigMessage, p
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


ldez:
Due to a huge refactor of the Rancher provider, the changes in this PR are been reported to #1563.

emilevauge:
Closing this one, superseded in #1563.

Issue #1207, opened 2017-03-01
Feature request: "sticky sessions" based on configured header


Slijkhuis:
Traefik supports sticky sessions using cookies like this:

```
[backends]
  [backends.backend1]
    [backends.backend1.loadbalancer]
      sticky = true
```

However, this requires both the client and Treafik to keep track of the _TRAEFIK_BACKEND-cookie.
Internal load balancers and stateless APIs don't always like cookies.

Sometimes, clients just need their requests to go to the same server when the same header is sent. Imagine a backend that holds a lot of user-specific data in memory. It's a waste to have this data in memory on every backend-server. The client may send a header like "X-Balance-User-Id: xxxxxxx". Based on that header the load balancer should then redirect the user to the same backend-server every time.

As an example, in HAProxy you'd use this configuration:

```
backend your_backend
  balance hdr(X-Your-Header)
```

I propose to add a parameter that specifies what kind of stickiness you want. For example:

```
# For cookies:
[backends]
  [backends.backend1]
    [backends.backend1.loadbalancer]
      stickyfunc = "cookie"
      sticky = true

# For a specific header:
[backends]
  [backends.backend1]
    [backends.backend1.loadbalancer]
      stickyfunc = "hashhdr(X-Your-Header)"
      sticky = true
```

Kind of the same way that `extractorfunc` categorizes requests based on a header.

Will such a feature be added?

kleinsasserm:
I would like to throw in #988 at this issue. Yes, there are situations where such configuration would be helpful. But for the first shot wouldn't be an ip address based load balancing enough? A simple hash of the source ip-address mapped to the backend would be a possibility. As I wrote in #988, maybe one parameter would be enough to manage this. If sticky is not set then no sticky sessions at all. If sticky is set, one and only one option is possible. eg:
```
sticky = "cookie" OR
sticky = "ipaddress" OR
sticky = "...." 
```

bcorijn:
If I understand correctly, this would also be useful for people who prefer not to "leak" backend IP's through the sticky sessions cookie, by being able to specify their own cookie?

kleinsasserm:
Yes, that's correct.

MichaelErmer:
@kleinsasserm no ip based load balancing would not be sufficient as all requests from a customers server would be balance to the same server.

kleinsasserm:
Is I wrote:
```
sticky = "cookie" OR
sticky = "ipaddress" OR
sticky = "...." 
```
[...]

PR #1418, opened 2017-04-11
Convert LoadBalancer.Sticky to a string

indrekj:
Previously it had only true/false value which indicated the use of
cookie based sticky sessions. There however can be other types of sticky
sessions (e.g. ip address, custom headers, custom functions). Using a
string to choose a type is more appropriate here.

Using "true" value will still default to "cookie" for backwards
compatibility.

Should be useful for https://github.com/containous/traefik/issues/1207

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -237,7 +237,7 @@ For example:
 [backends]
   [backends.backend1]
     [backends.backend1.loadbalancer]
-      sticky = true
+      sticky = "cookie"
 ```
 
 Healthcheck URL can be configured with a relative URL for `healthcheck.URL`.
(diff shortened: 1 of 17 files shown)


indrekj:
Correct. I do have a plan to submit a PR for Custom Header based routing but that has turned out to be a bit more complicated that I anticipated.

timoreimann:
@indrekj Gotcha.

I'd prefer to keep this PR open until you're fairly confident your follow-up PR is ready. That way, we make sure it truly fits the need by having a proper, real-world "consumer".

indrekj:
Closing PR as I'm far from implementing other sticky types. Can always reopen later.

PR #1421, opened 2017-04-11
Move make cross binary to Travis deploy step

emilevauge:
Currently `make crossbinary` takes ~ 20 min on Travis, I propose to move `make crossbinary` & `make image` from `after_success` to `before_deploy`

Signed-off-by: Emile Vauge <emile@vauge.com>


emilevauge:
:heart_eyes: 
travis

PR #1436, opened 2017-04-13
Updating Kubernetes tests to properly test missing endpoints code path

### LINKED ISSUE #1307: Update Kubernetes tests to properly handle endpoints not existing
In #1295 I introduced a temporary work around, this issue is for removing that and cleaning up the tests that break.

The code that originally causes the issue is this:
```
func (c clientMock) GetEndpoints(namespace, name string) (*v1.Endpoints, bool, error) {
	if c.apiEndpointsError != nil {
		return &v1.Endpoints{}, false, c.apiEndpointsError
	}

	for _, endpoints := range c.endpoints {
		if endpoints.Namespace == namespace && endpoints.Name == name {
			return endpoints, true, nil
		}
	}

	return &v1.Endpoints{}, true, nil
}
```

The final return has true as the second value being returned even if no endpoints for the service were found. This means the tests never hit the code path for `!exists` here: https://github.com/containous/traefik/blob/master/provider/kubernetes.go#L240

Instead what happens is all the tests that don't specify endpoints go through and hit the code path for `len(endpoints.Subsets) == 0`.

We need to update the mock `GetEndpoints` function to return `false` for the second return value when no endpoints are found, fix all the tests that break, and then ensure that there is a test that tests the current code path being used.

Here are the tests that fail when the `GetEndpoints` function returns correctly: TestHostlessIngress, TestLoadMultipleNamespacedIngresses, TestLoadNamespacedIngresses, TestGetPassHostHeader, TestOnlyReferencesServicesFromOwnNamespace, TestLoadIngresses

Related to this is defining and possibly changing what happens when endpoints are not found. Currently when a service is not found the referencing frontend is removed. However if no endpoints are found we still keep the frontend. Should we be removing the frontend if no endpoints are found?

Regner:
This fixes #1307

### DIFF
--- a/provider/kubernetes/kubernetes_test.go
+++ b/provider/kubernetes/kubernetes_test.go
@@ -264,10 +264,6 @@ func TestLoadIngresses(t *testing.T) {
 			},
 			"bar": {
 				Servers: map[string]types.Server{
-					"2": {
-						URL:    "http://10.0.0.2:802",
-						Weight: 1,
-					},
 					"https://10.15.0.1:8443": {
 						URL:    "https://10.15.0.1:8443",
 						Weight: 1,
@@ -498,12 +494,7 @@ func TestGetPassHostHeader(t *testing.T) {
 	expected := &types.Configuration{
 		Backends: map[string]*types.Backend{
 			"foo/bar": {
-				Servers: map[string]types.Server{
-					"1": {
-						URL:    "http://10.0.0.1:801",
-						Weight: 1,
-					},
-				},
+				Servers:        map[string]types.Server{},
 				CircuitBreaker: nil,
 				LoadBalancer: &types.LoadBalancer{
 					Sticky: false,
@@ -610,12 +601,7 @@ func TestOnlyReferencesServicesFromOwnNamespace(t *testing.T) {
 	expected := &types.Configuration{
 		Backends: map[string]*types.Backend{
 			"foo": {
-				Servers: map[string]types.Server{
-					"1": {
-						URL:    "http://10.0.0.1:80",
-						Weight: 1,
-					},
-				},
+				Servers:        map[string]types.Server{},
 				CircuitBreaker: nil,
 				LoadBalancer: &types.LoadBalancer{
 					Sticky: false,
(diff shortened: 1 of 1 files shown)


Regner:
As a heads up when I run the provider tests locally I get the following failed test:
```
--- FAIL: TestDockerLoadDockerServiceConfig (0.00s)
	docker_test.go:3124: expected map[string]*types.Backend{"backend-foo-service":(*types.Backend)(0xc42062cc90)}, got map[string]*types.Backend{"backend-foo":(*types.Backend)(0xc4206576b0)}
```

This PR doesn't touch anything relating to that test. Curious if it fails in Travis. Haven't looked into why but I am running from head.

provider/kubernetes_test.go
```diff
@@ -499,12 +495,7 @@ func TestGetPassHostHeader(t *testing.T) {
 	expected := &types.Configuration{
 		Backends: map[string]*types.Backend{
 			"foo/bar": {
-				Servers: map[string]types.Server{
-					"1": {
-						URL:    "http://10.0.0.1:801",
-						Weight: 1,
-					},
-				},
+				Servers:        map[string]types.Server{},
```
timoreimann: What's the reason we are able to remove / empty a bunch of server definitions?

timoreimann: Oh of course, it's because we now model missingness correctly, right?

Regner: Yea, see this line: https://github.com/containous/traefik/pull/1436/files#diff-4c0e9bf481017297ab14b697157b2675R2079

Regner:
One of the annoying things was at first I just left the `Servers:        map[string]types.Server{},` line out, and the tests failed but looked 100% identical according to the output but were actually different.

timoreimann:
@Regner :+1:

In the referencing issue, you also mention the fact that we keep frontends dangling when no Endpoints can be found. Do you know what the effect of that is? And should we address it in this PR or a different one? (I'm open to both paths.)

Regner:
I don't know what the effect of that is and am unsure how that should be handled. I think it needs some discussion first so should probably be a different issue and PR.

timoreimann:
@Regner makes sense to me.

timoreimann:
@Regner could you rebase and fix the conflicts? We recently moved all providers into separate sub-packages. Sorry for the inconveniences!

PR #1440, opened 2017-04-15
Few refactoring around the docker provider

vdemeester:
This is a big one, sorry for that 😛 (and this is just the beginning). This one does 2 things, separated in 2 commits

- Move the `docker` provider to its own package (`provider/docker`). I intend to do that for all of the providers.
- Refactor some test to be less noisy using builders. This is just the first step… they are still too noisy 😛 

Note that there is still some flakyness in the unit tests 😱 

🦁


provider/docker/builder_test.go
```diff
@@ -0,0 +1,69 @@
...
+		NetworkSettings: &docker.NetworkSettings{
+			NetworkSettingsBase: docker.NetworkSettingsBase{},
+		},
+	}
+
+	for _, op := range ops {
+		op(c)
+	}
+
+	return *c
```
timoreimann: How about returning a pointer since we are already applying the ops on the pointer?

provider/docker/builder_test.go
```diff
@@ -0,0 +1,69 @@
...
+			op(c.NetworkSettings.Networks[name])
+		}
+	}
+}
+
+func ipv4(ip string) func(*network.EndpointSettings) {
+	return func(s *network.EndpointSettings) {
+		s.IPAddress = ip
+	}
+}
```
timoreimann: I'm on the fence whether the builder deserves its on set if tests. It doesn't seem strictly required but would be helpful to have them.

provider/docker/docker.go
```diff
@@ -28,30 +30,29 @@ import (
...
+	Endpoint              string              `description:"Provider server endpoint. Can be a tcp or a unix socket endpoint"`
+	Domain                string              `description:"Default domain used"`
+	TLS                   *provider.ClientTLS `description:"Enable Provider TLS support"`
+	ExposedByDefault      bool                `description:"Expose containers by default"`
+	UseBindPortIP         bool                `description:"Use the ip address from the bound port, rather than from the inner network"`
+	SwarmMode             bool                `description:"Use Provider on Swarm Mode"`
 }
 
-// dockerData holds the need data to the Docker provider
+// dockerData holds the need data to the Provider p
```
vdemeester: oups 😓

provider/docker/builder_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	docker "github.com/docker/engine-api/types"
+	"github.com/docker/engine-api/types/container"
+	"github.com/docker/engine-api/types/network"
+	"github.com/docker/go-connections/nat"
+)
+
+func containerJSON(ops ...func(*docker.ContainerJSON)) docker.ContainerJSON {
+	c := &docker.ContainerJSON{
```
ldez: variable names with one letter, are you sure ?

provider/docker/builder_test.go
vdemeester: So, we always use `ContainerJSON` as not a pointer so… I prered to have just on `return *c` than a lot of `*containerJSON(…)` 😝

provider/docker/builder_test.go
vdemeester: right, so… I don't want to overdo it too. 😛

provider/docker/builder_test.go
vdemeester: not sure, but between `c`, `ctnr` or `cont`, I went lazy 👼  (didn't want to use `container` because it would override the `container` package)

provider/docker/builder_test.go
timoreimann: My point was that we pass-by-pointer when we apply the ops only to do a return-by-value eventually. But I suppose it's fine this way.

provider/provider.go
```diff
@@ -60,7 +61,7 @@ func (p *BaseProvider) getConfiguration(defaultTemplateFile string, funcMap temp
 	var defaultFuncMap = template.FuncMap{
 		"replace":   replace,
 		"tolower":   strings.ToLower,
-		"normalize": normalize,
+		"Normalize": Normalize,
```
emilevauge: I suppose that a regression might have been introduced here as no template have been modified. We should leave the first one in low caps to stay consistent `"normalize": Normalize`.

vdemeester: Oups you are right. I'll fix that  ;)

timoreimann: Shouldn't this have been caught by some test?

vdemeester: I think it is not used in any template (in the repo) but.. Yeah.. 😅

SantoDE:
LGTM ⚓️ NICE WORK @vdemeester

PR #1441, opened 2017-04-15
Do not retry unit tests.

timoreimann:
Unit tests should never be flaky; if there are, it's a bug we need to fix. Therefore, we should stop retrying them.


timoreimann:
I suppose this one requires #1443 to get fixed prior to merging.

timoreimann:
Closing in favor of #1489 which piggybacked the functionality.

PR #1444, opened 2017-04-16
Extract providers to their own packages

vdemeester:
Following #1440, this extracts the other provider to their own packages. Still refactoring to do (some stuff to un-export in k8s for ex, tests refactoring, etc..).

As it's based on #1440, only the last commit sholud be reviewed — and I'll rebase once it's merged 👼

Changes around : +840/-815


provider/kv/kv.go
```diff
@@ -12,45 +11,47 @@ import (
...
-	storeType    store.Backend
-	kvclient     store.Store
+// Provider holds common configurations of key-value providers.
+type Provider struct {
+	provider.BaseProvider `mapstructure:",squash"`
+	Endpoint              string              `description:"Comma sepparated server endpoints"`
+	Prefix                string              `description:"Prefix used for KV store"`
+	TLS                   *provider.ClientTLS `description:"Enable TLS support"`
+	StoreType             store.Backend
+	Kvclient              store.Store
```
emilevauge: I'm not sure that's a good idea to have these fields exported. They will be added automatically to the flags traefik --help (to be verified).

vdemeester: @emilevauge I don't see them in `--help` output. I guess it's only exported field that have the `description:"…"` thinguy 👼

emilevauge: OK thanks :) Is there any possibility to *not* export these fields in the future? Not really fond of this ;)

vdemeester: @emilevauge yes, it is in my plan somehow 👼

Issue #513, opened 2016-07-11
Preload image for integration tests


vdemeester:
The idea is to do more or less what we do on `docker/docker` and pre-fetch images when doing the `docker build` so that locally, integration tests are way quicker to run using `make test-integration`

/cc @containous/traefik

PR #1445, opened 2017-04-17
Pull images before running CI

vdemeester:
The idea is simple : when running the CI, the integration step should not pull anything 👼


vdemeester:
Removing `circle.yml` because we don't use it anymore.

.travis.yml
```diff
@@ -23,12 +23,11 @@ before_install:
...
   - docker version
   - pip install --user -r requirements.txt
+  - make pull
 before_script:
   - make validate
   - make binary
-script:
-  - travis_retry make test-unit
-  - travis_retry make test-integration
+script: travis_retry make test-unit && travis_retry make test-integration
```
vdemeester: This changes is simple : **if test-unit fails, DON'T run integration test** 👼.

timoreimann: Should we remove the retry on the unit test execution right away?

Makefile
```diff
@@ -82,5 +82,11 @@ lint:
 
 fmt:
 	gofmt -s -l -w $(SRCS)
+
+pull:
```
timoreimann: On second thought: can we call this target `pull-images`?

`pull` seems a bit too generic / unspecific to me.

SantoDE:
LGTM once tests are green 👼

Makefile
vdemeester: I do agree 👍

.travis.yml
vdemeester: meh, I want to fix the flakeys before 😛

timoreimann: :+1:

timoreimann:
Relates to #513.

PR #1006, opened 2016-12-30
Move some stuff in the right packages

vdemeester:
- This will help split stuff in smaller, better tested packages
- This moves some stuff like the traefik command to package `cmd`

Note that this is a start (and stays small) but the `server` package should be split up then, etc.

/cc @containous/traefik 

🐸

Signed-off-by: Vincent Demeester <vincent@sbr.pm>

PR #1449, opened 2017-04-17
Extract some code in packages

vdemeester:
Reviving #1006 👼

- This will help split stuff in smaller, better tested packages
- This moves some stuff like the traefik command to package `cmd`

It's build on top of #1444, so only the 2nd commit should be reviewed 👼 

🦁

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -67,7 +67,8 @@ $ go generate
 $ go build
 # Using gox to build multiple platform
 $ gox "linux darwin" "386 amd64 arm" \
-    -output="dist/traefik_{{.OS}}-{{.Arch}}"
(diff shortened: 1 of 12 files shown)


timoreimann:
I think you'll need to adjust the Makefile targets and `script/` helpers since the build target changed from root to `cmd`.

Docs in `.github/CONTRIBUTING.md` likely need a small update too.

timoreimann:
Hmm, given that the build ran for quite a while (even though erroneously), my comment might be false. Maybe you can double check.

vdemeester:
@timoreimann yep, I need to update docs, just want to make sure these changes are ok for @emilevauge and everybody before doing it 👼

emilevauge:
@vdemeester Looks very good to me ( love the new `server` package :heart_eyes: )

vdemeester:
Updated with the docs 👼

PR #1453, opened 2017-04-18
Fix Rancher API pagination limits

### LINKED ISSUE #1452: Traefik does not pull all environments/services/containers from the Rancher API
Traefik currently (v1.2.3 and `master`) does not pull the complete set of environments/services/containers from the Rancher API meaning that users with large Rancher deployments have missing or incomplete backend definitions.

The issue is due to the pagination enabled by default in the Rancher client ([see here](https://docs.rancher.com/rancher/v1.5/en/api/v1/#pagination)), specifically: `API responses are paginated with a limit of 100 resources per page by default.`

martinbaillie:
This fix allows the Traefik Rancher provider to obtain a complete view
of the environments, services and containers being managed by the
Rancher deployment.

Resolves #1452


provider/rancher.go
```diff
@@ -290,7 +290,10 @@ func listRancherEnvironments(client *rancher.RancherClient) []*rancher.Environme
 
 	var environmentList = []*rancher.Environment{}
 
-	environments, err := client.Environment.List(nil)
+	withoutPagination := &rancher.ListOpts{
```
timoreimann: How about setting up a "const" `ListOpts` and reuse it?

(where "const" could possibly mean: declare a var and initialize it in an `init()` function.)

timoreimann:
Could we cover the disabled pagination in one of our integration tests somehow? Not a must if it turns out to be tricky, but would be nice.

provider/rancher.go
martinbaillie: Sure @timoreimann. Like [this](https://github.com/containous/traefik/pull/1453/commits/bf331f71bcba22b44c6f70c20430ae560847f4bb)?

provider/rancher.go
```diff
@@ -44,6 +48,10 @@ type rancherData struct {
 	Health     string
 }
 
+func init() {
+	withoutPagination.Filters["limit"] = 0
```
timoreimann: You're accessing the fields of a pointer to a script. Wouldn't that cause a nil dereference error unless you initialized the pointer?

martinbaillie:
It may be tricky as it's a struct passed straight into the Rancher client but happy to take a look @timoreimann. Could you point me towards the integration tests for Rancher though? I couldn't find them 😕

timoreimann:
@martinbaillie: apologies -- I thought we had integration tests for Ranger but apparently we don't. IIRC, it's on the todo list.

Let's leave that for another PR then.

martinbaillie:
@timoreimann no problem re. integration tests. 

I've tidied up the branch commits and fixed that init mistake 🤒

martinbaillie:
@SantoDE merged in master again. cheers 👍

PR #1461, opened 2017-04-19
Update docker tests to use testify

vdemeester:
It displays diff with `go-spew` which is way better than our current failure output.

<del>It also removes t.Parallel as without it tests run in 0.018s.. and `t.Parallel` really freaks me out</del>

I intend to follow-up for each and every tests (starting with providers). For providers I'll also do the builders for each .

### DIFF
--- a/provider/docker/docker_test.go
+++ b/provider/docker/docker_test.go
@@ -1,7 +1,6 @@
 package docker
 
 import (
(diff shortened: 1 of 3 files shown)


vdemeester:
I'll close this one for now 👼

PR #1464, opened 2017-04-19
Vendor generated file

vdemeester:
The main reason is : allows compilation (and `go test -v`, …) on non-unix systems (like windows).
We already vendor dependencies so… 👼 


timoreimann:
When do developers need to re-run the auto-generation step?

The docs should probably be updated in this regard.

vdemeester:
@emilevauge yes I can 👼 I also want to add a check (like validate glide) for it (i.e. if a template changes in the commit/diff, validates that the autogen file is correct). But first I need to update `validate-gofmt` 😂

timoreimann:
Good point about the extra validation script we need. :+1:

vdemeester:
Updated with the validate files 👼

script/validate-autogen
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
timoreimann: `/usr/bin/env bash` is preferred.

script/validate-gofmt
```diff
@@ -3,7 +3,7 @@
 source "$(dirname "$BASH_SOURCE")/.validate"
 
 IFS=$'\n'
-files=( $(validate_diff --diff-filter=ACMR --name-only -- '*.go' | grep -v '^\(integration/\)\?vendor/' || true) )
+files=( $(validate_diff --diff-filter=ACMR --name-only -- '*.go' | grep -v '^\(integration/\)\?vendor/\|autogen' || true) )
```
timoreimann: Did you verify the pipe operator works (that is, it really ignores the autogen file while not breaking the vendor things)?

Last time I tried, I had quite some issues, which is why I'm asking.

timoreimann: Does this now match an `autogen` file located anywhere in our code base? Is that what we want? Do we need to be more / less specific?

.github/CONTRIBUTING.md
```diff
@@ -72,6 +72,10 @@ $ gox "linux darwin" "386 amd64 arm" \
 # run other commands like tests
 ```
 
+### Updating the templates
+
+If you happen to update the provider templates (in `/templates`), you need to run `go generate` to update the `autogen` package.
```
timoreimann: Another comment in the file mentions that the generation step is required for the web UI. If that's true, we should mention it there as well and possibly update the other comment.

script/validate-autogen
vdemeester: yep I know, all other scripts are `/bin/bash` so.. I did a mimic and was about to do a PR to switch to `/usr/bin/env bash`

script/validate-gofmt
vdemeester: yes for the first questio.
to test for the 2nd 👼

.github/CONTRIBUTING.md
vdemeester: I'll look 👼

script/validate-gofmt
vdemeester: it's ok for the second (i.e. `provider/autogen.go` for ex isn't exclude by this and this will be verified)

.github/CONTRIBUTING.md
```diff
@@ -61,7 +61,8 @@ Here's a full example:
 $ ./script/glide.sh get github.com/foo/bar
 # install another dependency, this time for the integration tests
 $ ( cd integration && ../script/glide.sh get github.com/baz/quuz )
-# generate (Only required to integrate other components such as web dashboard)
+# generate
+# (Only required to integrate other components such as web dashboard and if you change something there — web, templates)
```
timoreimann: >if you change something there [being the dashboard]

The templates aren't really related to the dashboard, are they?

How about rephrasing it like this:

```
# generate
# (required to merge non-code components into the final binary, such as the web dashboard and provider Go templates)
```

script/validate-autogen
timoreimann: I'd rather see this being updated either for `validate-autogen` or all scripts right on this PR. Otherwise, chances are the script will not work until you submitted that follow-up PR. (For instance, on my Mac `/bin/bash` points at the ancient bash 3 version that doesn't work with many features.)

script/validate-gofmt
timoreimann: 👍

vdemeester:
ping @containous/traefik 👼 (I'll rebase once I have enough LGTM…)

ldez:
@vdemeester it's time to rebase! 😃

script/validate-autogen
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+set -o errexit
+set -o pipefail
+set -o nounset
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"; export SCRIPTDIR
+source "${SCRIPTDIR}/.validate"
+
+# Iterate over all directories containing vendor folders.
```
emilevauge: s/vendor/templates ?

emilevauge:
@vdemeester, As discussing with @ldez, it seems the validation script is incomplete and lacks webui checks (in go generate: `go-bindata -pkg autogen -o autogen/gen.go ./static/... ./templates/...`).
But adding a webui check in validation will be a lot slower, building webui takes some time.
WDYT? revert?

Issue #795, opened 2016-11-02
There is no doc about how to pronounce "traefik"


alvaroaleman:
When telling others about traefik, I'd like to properly pronounce it. t would be helpful if there was a short note about that in the docs.

emilevauge:
I will add some help on this 😉
We simply say traffic, written træfɪk using the international phonetic alphabet.

michaelkrog:
Only obstacle be that not everyone have `æ` on their keyboard - like here in Scandinavia! (Muhahaha) 😄

ldez:
Fix by #1067

PR #1468, opened 2017-04-19
Mention Traefik pronunciation in docs too.

timoreimann:
Also replace Træfɪk with Træfik.

Completes #795.


Alexspayne:
This link is broken now.

PR #1469, opened 2017-04-19
Improve documentation for frontend rules.

### LINKED ISSUE #1411: Improve documentation on path matchers
We frequently get the question on Slack what the difference between the various path matchers is. We should likely take this as a signal to improve the docs.

My explanation in a nutshell:

All of them match by path.

- `Path`: match an exact path.
- `PathPrefix`: match a prefix of a path (the suffix can be arbitrary).
- `*Strip` variants: same as the other non-strip ones but remove the Path/PathPrefix after matching (do this if your application expects requests on the prefix-less route).

timoreimann:
Includes guidelines on proper usage of the more complex path matchers.

Fixes #1411.

/cc @chmod666org @bfqrst

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -73,28 +73,58 @@ And here is another example with client certificate authentication:
 
 ## Frontends
 
-A frontend is a set of rules that forwards the incoming traffic from an entrypoint to a backend.
-Frontends can be defined using the following rules:
+A frontend consists of a set of rules that determine how incoming requests are forwarded from an entrypoint to a backend.
 
-- `Headers: Content-Type, application/json`: Headers adds a matcher for request header values. It accepts a sequence of key/value pairs to be matched.
-- `HeadersRegexp: Content-Type, application/(text|json)`: Regular expressions can be used with headers as well. It accepts a sequence of key/value pairs, where the value has regex support.
-- `Host: traefik.io, www.traefik.io`: Match request host with given host list.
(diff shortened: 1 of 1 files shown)


bfqrst:
Hands down, this is way better... For some reason I always thought that if you want to strip the path from a request, you need to include the non-strip part beforehand. Like the strip-directives ONLY strip and the other directives ONLY do the matching... Does that even makes sense? :-)

Anyway, the static content serving I was struggling with is a thing of the past.

Thanks a lot guys! Much appreciated!

docs/basics.md
```diff
@@ -73,28 +73,58 @@ And here is another example with client certificate authentication:
...
+
+(Note that the variable has no special meaning; however, it is required by the gorilla/mux dependency which embeds the regular expression and defines the syntax.)
+
+#### Path Matcher Usage Guidelines
+
+This section explains when to use the various path matchers.
+
+Use `Path` if your backend listens on the exact path only. For instance, `Path: /products` would match `/products` but not `/products/shoes`.
+
+Use a `*Prefix*` matcher if your backend listens on a particular base path but also serves requests on sub-paths. For instance, `PathPrefix: /products` would match `/products` but also `/products/shoes` and `/products/shirts`. Since the path is forwarded as-is, your backend is expected to listen on `/products`.
```
7imbrook: I want to plus one this as REALLY helpful because it wasn't clear in the prior docs

PR #1470, opened 2017-04-20
Update dockerignore to ignore dist and sites

vdemeester:
As of now, it does nothing (`/dist/` doesn't filter the dist folder) and sending anything from `dist` doesn't make sense as it's mounted anyway (that's why removing `!/dist/traefik`).

To see why it's important : `make cross-binary && make validate`… look at the size of the build context sent on the 2nd command (without this PR >600M, with this PR ~60M).

> Removing the traefik binary from whitelist as the integration script
compiles the binary before running, so we don't need to send it via
the build context.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,2 @@
-/dist/
-!/dist/traefik
-/site/
+dist/
+site/


timoreimann:
Could you also explain in the issue description why you remove the Traefik binary exception?

timoreimann:
Did you find any upstream Docker bug that explains why the leading slash makes a difference (which it shouldn't according to the Dockerfile reference)?

timoreimann:
@vdemeester found the upstream bug, it's here: moby/moby#30676 (fixing PR in progress).

vdemeester:
I think it's up-to-date 👼 @timoreimann

ldez:
@vdemeester need rebase 😉

Issue #1456, opened 2017-04-18
Enable http to https redirection per frontend


thedylman:
### What version of Traefik are you using (`traefik version`)?
1.2.1

### What is your environment & configuration (arguments, toml...)?
Using docker stack deploy, traefik started with command:

```yml
command: [
      "traefik",
        "--docker",
        "--docker.swarmmode",
        "--docker.domain=sw.test.org",
        "--docker.watch",
        "--entryPoints=Name:http Address::80 Redirect.EntryPoint:https",
        "--entryPoints=Name:https Address::443 TLS:/run/secrets/domain.crt,/run/secrets/domain.key",
        "--defaultentrypoints=http,https",
        "--web",
        "--web.address=:8080"
      ]
```

### What did you do?
Is it possible to disable/enable the http to https redirection per service?

e.g:
service 1 (with http to https):
* frontend rule: Host:service1.sw.test.org
* entrypoints: http, https

service 2 (http only, is it possible to disable the http to https redirection here?)
* frontend rule: Host:service2.sw.test.org
* entrypoints: http

traefiker:
I'll close the issue due to inactivity but feel free to re-open if needed.

Yajo:
Please reopen, this is still missing.

SantoDE:
Hey @Yajo,

please have a look at #2133 as this would bring in the functionality :)

Issue #987, opened 2016-12-21
Frontend-based Redirect Rules


adamgoose:
### What version of Traefik are you using (`traefik version`)?
Latest

### What is your environment & configuration (arguments, toml...)?
dynamic file-based from confd

### What did you do?
```toml
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
      regex = "^http://(SOME|BUT|NOT|ALL|DOMAINS|HERE)/(.*)"
      replacement = "https://$1/$2"
  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]
```

### What did you expect to see?
When `https.domain.com` is included in the regex, and said frontend has `entrypoints = http & https`, and I hit `https.domain.com`, I am successfully redirected to HTTPS.

When `http.domain.com` is *not* included in the regex, and said frontend has `entrypoints = http only`, and I hit `http.domain.com`, I should *not* be redirected, and the request *should* proxy to the proper backend.

### What did you see instead?
When `http.domain.com` is *not* included in the regex, and said frontend has `entrypoints = http only`, and I hit `http.domain.com`, I am *not* redirected, but only see a white page; the request is not proxied to the backend.

---

Suggested solution: allow for redirect to/from endpoints at the frontend config level.

avarabyeu:
+1

f3l1x:
:+1:

f3l1x:
It would be nice if something like that would works.

```
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
      regex = "^http://domain1.tld/(.+)"
      replacement = "https://roodomain.tld"
    [entryPoints.http.redirect]
      regex = "^http://domain2.tld/(.+)"
      replacement = "https://roodomain.tld"
    [entryPoints.http.redirect]
      regex = "^http://domain3.tld/(.+)"
      replacement = "https://roodomain.tld"
```

Do you guys have it your plan? @emilevauge

Crempa:
yes please 👍

j0hnsmith:
Per frontend redirects would be great but I feel the config should be on a frontend rather than an entrypoint as entrypoints work at the network level. 

This way the all the flexibility already provided by frontend rules would be available to configure a redirect, eg `Host`, `HostRegexp`, `Header`, `Method`, `Path` etc etc. Maybe something like this would work

```
[frontends]
  [frontends.frontend1.routes.redirect_1]
[...]

PR #1471, opened 2017-04-20
Only enable redirection if a frontend has no backends

tcolgate:
This change allows you to selectively redirect front ends. To support
setups such as #1456 and #987

With this in place, I have a marathon template that only sets the
backend on frontends that have a label that requests redirect to https.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -598,7 +598,7 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 				}
 
 				entryPoint := globalConfiguration.EntryPoints[entryPointName]
-				if entryPoint.Redirect != nil {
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -598,7 +598,7 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 				}
 
 				entryPoint := globalConfiguration.EntryPoints[entryPointName]
-				if entryPoint.Redirect != nil {
+				if entryPoint.Redirect != nil && configuration.Backends[frontend.Backend] == nil {
```
emilevauge: This change seems to break everything.
Could you give us a working configuration to test this use case (toml)?

tcolgate:
Closing this, it was working before the introduction of negroni (I think), but looking at it now, it definitely needs a rethink

PR #1473, opened 2017-04-20
Correct typo in code comment.

mattcollier:


healthcheck/healthcheck.go
```diff
@@ -15,7 +15,7 @@ import (
 var singleton *HealthCheck
 var once sync.Once
 
-// GetHealthCheck Get HealtchCheck Singleton
+// GetHealthCheck Get HealthCheck Singleton
```
timoreimann: Let's have a proper GoDoc string and say something like

`// GetHealthCheck returns the health check which is guaranteed to be a singleton.`

mattcollier:
@timoreimann Sure, done.

Issue #261, opened 2016-03-24
Marathon label traefik.port cannot be parsed


ktf:
I am trying to use traefik to redirect to some marathon app running with `--net=host` so I try to force the port by setting the marathon label `traefik.port` to 8888. Treafik however complains it cannot parse it seems to print the message in https://github.com/containous/traefik/pull/138/files#diff-fca3fabc700de27831fb573ad618a81dR140 . If I drop the label and switch to bridge mode, it works just fine.

boymaas:
Same Issue here. Trying to run a marathoh app with --net=host and not able to explicitly define a port with traefik.port due to https://github.com/containous/traefik/blob/master/provider/marathon.go#L202

timoreimann:
This was fixed in #1394 and will be part of the upcoming 1.3 release.

PR #1474, opened 2017-04-20
Check for explicitly defined Marathon port first.

timoreimann:
Check for explicitly defined Marathon port first.
    
Previously, we did the check too late resulting in the `traefik.port` label not being effective.
    
The change comes with additional refactorings in production and tests.

Original issue at #261.
Complements PR at #1394.
Discovered along #1390.

### DIFF
--- a/provider/marathon/marathon.go
+++ b/provider/marathon/marathon.go
@@ -2,6 +2,7 @@ package marathon
 
 import (
 	"errors"
+	"fmt"
 	"math"
 	"net"
 	"net/http"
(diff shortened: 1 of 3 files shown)


timoreimann:
The initial push contains an updated test only to demonstrate the problem. Once CI goes legitimately red, the fix will be pushed.

timoreimann:
@containous/traefik we should get this into 1.3 as the `traefik.port` label override is currently broken with Marathon.

Sorry for the big delta in the tests, but the fact that they did not adequately address/test-cover the code is one of the reasons why the problem didn't surface earlier. Thus, I needed to shift things around.

timoreimann:
@vdemeester 

1. I plan to follow your good example and do some cleanup on Marathon. I need to add support for [readiness checks](https://mesosphere.github.io/marathon/docs/readiness-checks.html) in the foreseeable future, which will require doing requests against the Marathon API slightly different (it'll be one combined request per event update instead of two (apps + tasks)). This will require overhauling the tests anyway, so I'm going to take that opportunity.
2. You're definitely right, thanks for the reminder. Will do so right away.

@ttr: Sorry for the noise (again) -- I should sync my Slack and Github handles in the future. :-)

PR #1482, opened 2017-04-22
doc: enhance GitHub template.

ldez:
- add a pull request template.
- add quick guides to create issue and PR.
- enhance existing issue template.


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,32 @@
+How to write a good issue ? Read [this](https://github.com/containous/traefik/blob/master/.github/how-to-write-a-good-issue.md) and the [contributing guide](https://github.com/containous/traefik/blob/master/.github/CONTRIBUTING.md).
```
timoreimann: We should use [relative links](https://github.com/blog/1395-relative-links-in-markup-files) so that reference don't break if we ever change the repo/hoster.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,4 @@
+How to write a good pull request ? Read [this](https://github.com/containous/traefik/blob/master/.github/how-to-write-a-good-pull-request.md) and the [contributing guide](https://github.com/containous/traefik/blob/master/.github/CONTRIBUTING.md).
```
timoreimann: Relative links here too.

.github/how-to-write-a-good-issue.md
```diff
@@ -0,0 +1,25 @@
+# How to write a good issue
+
+Please keep in mind that the GitHub issue tracker is not intended as a general support forum, but for reporting bugs and feature requests.
+
+For other type of questions, consider using one of:
+
+- the [Traefik community Slack channel](https://traefik.herokuapp.com)
+- [StackOverflow](https://stackoverflow.com/questions/tagged/traefik)
```
timoreimann: I very much like this rule. What I'm wondering though is what kind of consequences are we going to draw from this? I can see two:

- Close question-style issues right away, pointing to Slack/StackOverflow.
- Just tolerate them (and maybe leave a "please use Slack/StackOverflow next time" message).

I know the Kubernetes folks do the former. However, they have to be very rigid due to the sheer amount of issues they receive.

.github/how-to-write-a-good-pull-request.md
```diff
@@ -0,0 +1,12 @@
+# How to write a good pull request
+
+- Make it small
+- Do only one thing
+- Avoid re-formatting
+- Make sure the code builds
+- Make sure all tests pass
+- Add tests
+- Write useful descriptions and titles
+- Keep your branch history clean
```
timoreimann: Not everyone may have the same understand of what this means. Can we be more specific?

timoreimann: Can we please add a point

- Address review comments in terms of additional commits; do not amend/squash existing ones unless the PR is trivial.

.github/ISSUE_TEMPLATE.md
ldez: it's not possible in an template.

.github/PULL_REQUEST_TEMPLATE.md
ldez: it's not possible in an template.

.github/ISSUE_TEMPLATE.md
ldez: This links are viewed by user who open an issue as text.
If we use relative links, the user can be lost.

timoreimann: You're right, I forgot that.

Thanks for pointing out. 👍

vdemeester: Shouldn't those comments be in html comment `<!-- …` ? Otherwise it's gonna show up in the final issue if ppl don't clean it (and I guarantee that not everybody will clean those). See https://raw.githubusercontent.com/moby/moby/master/.github/ISSUE_TEMPLATE.md 👼

.github/PULL_REQUEST_TEMPLATE.md
vdemeester: Same for PRs 👼

.github/ISSUE_TEMPLATE.md
ldez: It's a gmail filtering technique 😉

PR #1484, opened 2017-04-23
contrib: Dump keys/certs from acme.json to files

brianredbeard:
In the event that a user needs to explode their acme.json file into
a set of directories and relevant files for troubleshooting or use
with other programs this script will parse them into the components
in the following path structure:

```
certdir
├── certs
│   ├── domain-1.example.com
│   ├── domain-2.example.com
│   └── domain-n.example.com
└── private
    └── letsencrypt.key
```

### DIFF
--- a/contrib/scripts/dumpcerts.sh
+++ b/contrib/scripts/dumpcerts.sh
@@ -1,0 +1,141 @@
+#!/bin/bash
+# Copyright (c) 2017 Brian 'redbeard' Harrington <redbeard@dead-city.org>
+#
+# dumpcerts.sh - A simple utility to explode a Traefik acme.json file into a
+#                directory of certificates and a private key
+#
+# Usage - dumpcerts.sh /etc/traefik/acme.json /etc/ssl/
+#
+# Dependencies - 
+#   util-linux
+#   openssl
+#   jq
+# The MIT License (MIT)
+# 
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+# 
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+# 
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+# Exit codes:
+# 1 - A component is missing or could not be read
+# 2 - There was a problem reading acme.json
+# 4 - The destination certificate directory does not exist
+# 8 - Missing private key
+
+USAGE="dumpcerts.sh <path to acme> <destination cert directory>"
+
+# Allow us to exit on a missing jq binary
+exitJQ() {
+	echo ""
+	echo "You must have the binary 'jq' to use this."
+	echo "jq is available at: https://stedolan.github.io/jq/download/"
+	echo ""
+	echo ${USAGE}
+	exit 1
+}
+
+
+badACME() {
+	echo ""
+	echo "There was a problem parsing your acme.json file."
+	echo ""
+	echo ${USAGE}
+	exit 2
+}
+
+
+acme="${1}"
+certdir="${2}"
+
+if [ ! -r "${acme}" ]; then
+	echo ""
+	echo "There was a problem reading from '${acme}'"
+	echo "We need to read this file to explode the JSON bundle... exiting."
+	echo ""
+	echo ${USAGE}
+	exit 1
+fi	
+
+
+if [ ! -d "${certdir}" ]; then
+	echo ""
+	echo "Path ${certdir} does not seem to be a directory"
+	echo "We need a directory in which to explode the JSON bundle... exiting."
+	echo ""
+	echo ${USAGE}
+	exit 4
+fi	
+
+jq=$(which jq) || exitJQ
+
+priv=$(${jq} -r '.PrivateKey' ${acme}) || badACME
+
+if [ ! -n "${priv}" ]; then
+	echo ""
+	echo "There didn't seem to be a private key in ${acme}."
+	echo "Please ensure that there is a key in this file and try again."
+	exit 8
+fi
+
+# If they do not exist, create the needed subdirectories for our assets
+# and place each in a variable for later use, normalizing the path
+mkdir -p "${certdir}"/{certs,private}
+
+pdir=$(realpath "${certdir}/private/")
+cdir=$(realpath "${certdir}/certs/")
+
+# Save the existing umask, change the default mode to 600, then
+# after writing the private key switch it back to the default
+oldumask=$(umask)
+umask 177
+# For some reason traefik stores the private key in stripped base64 format, but
+# the certificates bundled as a base64 object without stripping headers.  This
+# normalizes the headers and formatting.
+#
+# In testing this out it was a balance between the following mechanisms:
+# gawk:
+#  echo ${priv} | awk 'BEGIN {print "-----BEGIN RSA PRIVATE KEY-----"}
+#     {gsub(/.{64}/,"&\n")}1
+#     END {print "-----END RSA PRIVATE KEY-----"}' > "${pdir}/letsencrypt.key"
+#
+# openssl:
+# echo -e "-----BEGIN RSA PRIVATE KEY-----\n${priv}\n-----END RSA PRIVATE KEY-----" \
+#   | openssl rsa -inform pem -out "${pdir}/letsencrypt.key" 
+#
+# and sed:
+# echo "-----BEGIN RSA PRIVATE KEY-----" > "${pdir}/letsencrypt.key"
+# echo ${priv} | sed 's/(.{64})/\1\n/g' >> "${pdir}/letsencrypt.key"
+# echo "-----END RSA PRIVATE KEY-----" > "${pdir}/letsencrypt.key"
+#
+# In the end, openssl was chosen because most users will need this script
+# *because* of openssl combined with the fact that it will refuse to write the
+# key if it does not parse out correctly. The other mechanisms were left as
+# comments so that the user can choose the mechanism most appropriate to them.
+echo -e "-----BEGIN RSA PRIVATE KEY-----\n${priv}\n-----END RSA PRIVATE KEY-----" \
+   | openssl rsa -inform pem -out "${pdir}/letsencrypt.key" 
+
+umask ${oldumask}
+
+# Process the certificates for each of the domains in acme.json
+for domain in $(jq -r '.DomainsCertificate.Certs[].Certificate.Domain' acme.json); do
+	# Traefik stores a cert bundle for each domain.  Within this cert 
+	# bundle there is both proper the certificate and the Let's Encrypt CA
+	echo "Extracting cert bundle for ${domain}"
+	cert=$(jq -r --arg domain "$domain" '.DomainsCertificate.Certs[].Certificate |
+         	select (.Domain == $domain )| .Certificate' ${acme}) || badACME
+	echo ${cert} | base64 --decode > "${cdir}/${domain}.pem"
+done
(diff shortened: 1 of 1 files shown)


contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
+#!/bin/bash
```
timoreimann: `/usr/bin/env bash` is more portable and thus preferred.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+# Exit codes:
+# 1 - A component is missing or could not be read
+# 2 - There was a problem reading acme.json
+# 4 - The destination certificate directory does not exist
+# 8 - Missing private key
+
+USAGE="dumpcerts.sh <path to acme> <destination cert directory>"
```
timoreimann: Can we use `$(basename "$0")` to display the script name in order to be portable should we ever rename the file?

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+# Exit codes:
+# 1 - A component is missing or could not be read
+# 2 - There was a problem reading acme.json
+# 4 - The destination certificate directory does not exist
+# 8 - Missing private key
```
timoreimann: Which exit codes one is allowed to use is a bit [disputed](http://tldp.org/LDP/abs/html/exitcodes.html); personally, I'm okay with using the "reserved" ones in the low-digit range.

Why powers of two though instead of just 1, 2, 3, and 4? We don't really need to shift numbers around, do we?

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+
+USAGE="dumpcerts.sh <path to acme> <destination cert directory>"
+
+# Allow us to exit on a missing jq binary
+exitJQ() {
+	echo ""
+	echo "You must have the binary 'jq' to use this."
+	echo "jq is available at: https://stedolan.github.io/jq/download/"
+	echo ""
```
timoreimann: Please output the message to stderr (`>&2`) using either a here document or a string stretching over multiple lines ([SO answer](http://stackoverflow.com/questions/10969953/how-to-output-a-multiline-string-in-bash)).

Same for other error outputs below.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+# Exit codes:
+# 1 - A component is missing or could not be read
+# 2 - There was a problem reading acme.json
+# 4 - The destination certificate directory does not exist
+# 8 - Missing private key
+
+USAGE="dumpcerts.sh <path to acme> <destination cert directory>"
+
+# Allow us to exit on a missing jq binary
+exitJQ() {
```
timoreimann: I'd prefer a naming convention using only lower-case and underscores (`exit_jq`), like the [Google style guide recommends](https://google.github.io/styleguide/shell.xml?showone=Function_Names#Function_Names) as well.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+	echo ""
+	echo "You must have the binary 'jq' to use this."
+	echo "jq is available at: https://stedolan.github.io/jq/download/"
+	echo ""
+	echo ${USAGE}
+	exit 1
+}
+
+
+badACME() {
```
timoreimann: My preference goes towards `bad_acme` here too.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+if [ ! -d "${certdir}" ]; then
+	echo ""
+	echo "Path ${certdir} does not seem to be a directory"
+	echo "We need a directory in which to explode the JSON bundle... exiting."
+	echo ""
+	echo ${USAGE}
+	exit 4
+fi	
+
+jq=$(which jq) || exitJQ
```
timoreimann: Please don't use `which` ([reasoning](http://stackoverflow.com/a/677212/420061)); use one of the alternatives.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+badACME() {
+	echo ""
+	echo "There was a problem parsing your acme.json file."
+	echo ""
+	echo ${USAGE}
+	exit 2
+}
+
+
+acme="${1}"
```
timoreimann: Can we try to be a bit more specific? How about `acme_file`?

contrib/scripts/dumpcerts.sh
timoreimann: Also, do we actually need to capture the result in a variable and use that one to invoke jq (as we do in line 85)? Why not invoke it directly once we know it's available?

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+# Exit codes:
+# 1 - A component is missing or could not be read
+# 2 - There was a problem reading acme.json
+# 4 - The destination certificate directory does not exist
+# 8 - Missing private key
+
```
timoreimann: Please set

```
set -o errexit
set -o pipefail
set -o nounset
```

to cover for common errors and have the script terminate prematurely.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+	echo "Path ${certdir} does not seem to be a directory"
+	echo "We need a directory in which to explode the JSON bundle... exiting."
+	echo ""
+	echo ${USAGE}
+	exit 4
+fi	
+
+jq=$(which jq) || exitJQ
+
+priv=$(${jq} -r '.PrivateKey' ${acme}) || badACME
```
timoreimann: IMHO, we don't need the `|| badACME` part here and elsewhere below _if_ we enable `errexit`: If for some reason we cannot access the file, bash should bail out and give a meaningful error message.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+mkdir -p "${certdir}"/{certs,private}
+
+pdir=$(realpath "${certdir}/private/")
+cdir=$(realpath "${certdir}/certs/")
+
+# Save the existing umask, change the default mode to 600, then
+# after writing the private key switch it back to the default
+oldumask=$(umask)
+umask 177
+# For some reason traefik stores the private key in stripped base64 format, but
```
timoreimann: It'd be nice to know the reason and phrase it here.

Maybe @emilevauge or @dtomcej know?

contrib/scripts/dumpcerts.sh
timoreimann: I'd `readonly` the variable.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+	echo ""
+	echo "There was a problem parsing your acme.json file."
+	echo ""
+	echo ${USAGE}
+	exit 2
+}
+
+
+acme="${1}"
+certdir="${2}"
```
timoreimann: I'd `readonly` the variable.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+	echo "There didn't seem to be a private key in ${acme}."
+	echo "Please ensure that there is a key in this file and try again."
+	exit 8
+fi
+
+# If they do not exist, create the needed subdirectories for our assets
+# and place each in a variable for later use, normalizing the path
+mkdir -p "${certdir}"/{certs,private}
+
+pdir=$(realpath "${certdir}/private/")
```
timoreimann: Any particular reason we need to use `realpath`? AFAIK, it's not available on all systems.

contrib/scripts/dumpcerts.sh
```diff
@@ -0,0 +1,141 @@
...
+cdir=$(realpath "${certdir}/certs/")
+
+# Save the existing umask, change the default mode to 600, then
+# after writing the private key switch it back to the default
+oldumask=$(umask)
+umask 177
```
timoreimann: Shouldn't we use a `TRAP` to revert the umask in order to cover for the case where the script breaks anywhere between this line and the point where we return to the old umask?

emilevauge:
@brianredbeard Wow great :+1: 
I think it would be even better if we could add a new command to Traefik (like `traefik exportACME /etc/ssl`). It would allow to export certs even in cluster mode (certs from the KV store). WDYT ?

contrib/scripts/dumpcerts.sh
brianredbeard: This would just be so that in the long run you could continue through a number of failure situations and then give the user a composite at the end which could be used to identify all of the things that have gone wrong before doing anything (potentially) destructive.

brianredbeard:
@emilevauge More than likely this would be a good idea, that being said this was a case of something that I was solving for myself anyways and figured could be useful for others.  I'd say making something like this available until a better answer is available (and then burning this script in a fire) is more than acceptable.

contrib/scripts/dumpcerts.sh
brianredbeard: Was only really using this to normalize paths in the event of `${certdir}` being provided with a trailing slash.  I was of two minds about this.  `realpath` is a part of `coreutils` and thus should be on the lion's share of hosts (at least where traefik would be able to run). On the other hand it *is* a better answer to just guarantee this at the time of assignment via `${certdir%/}`... Admittedly this was me making this mostly for myself and thinking of it as "contrib" status work (YMMV).  Regardless, will fix with the assignment.

contrib/scripts/dumpcerts.sh
emilevauge: @timoreimann what is your point exactly ? base64 ?

timoreimann: @emilevauge Yes. I think it would be nice to rephrase

_For some reason traefik stores the private key in stripped base64 format_

into something like

_traefik stores the private key in stripped base64 format because [reason]_

or just 

_traefik stores the private key in stripped base64 format_ (without explanation)

I don't think we should leave comments in scripts that express uncertainty about the reasoning in our own code if we can avoid it.

ldez:
@brianredbeard It's clearly acceptable, can you fix the comments made by @timoreimann ? Thanks.

@emilevauge I've opened the issue #1512 for make the command in Traefik.

emilevauge:
Ping @brianredbeard

timoreimann:
@brianredbeard I carried your PR "in-place" and addressed all of my own comments that seemed undisputed. Hope that's okay with you. Feel free to comment back on me, of course.

@emilevauge PTAL.

brianredbeard:
That's absolutely fine by me! Sorry about being a drag ass but you have my full support.

ldez:

kachkaev:
Will this dump the certs automatically in real time? I'd like to mount a volume with my certs into gitlab and a mailserver, so am looking to find a folder with `pem` somewhere.

frederik:
hi @brianredbeard. First of all, thanks for the contribution.

I have another (quick) follow up question. From how I understand let's encrypt it has two certificates. This script is extracting the domain handling key into a .key file and the certificate into the .pem file. But it doesn't extract the private key for the individual certificate (i.e. DomainsCertificate.Certs[n].Certificate.PrivateKey).

Gitlab Container Registry is asking me to provide both, so I did the base64 extraction manually to confirm that this was actually required.

I suppose extracting the PrivateKey would only be a minor duplication in the script.

**tl;dr**

Is there a reason you chose not to extract the PrivateKey? (or just didn't require it?)

kachkaev:
@frederik could you please share your current workaround? I might need to pass a private key to gitlab or another upstream service in a near future :–)

frederik:
I took the contents from DomainsCertificate.Certs[n].Certificate.PrivateKey and base64 decoded it by hand.

```
base64 -d <<< THECONTENTOF DomainsCertificate.Certs[n].Certificate.PrivateKey
-----BEGIN RSA PRIVATE KEY-----
...
-----END RSA PRIVATE KEY-----
```
(on linux -d, on Mac -D)

.. to a file and referenced the key in gitlab.

```
registry_nginx['ssl_certificate'] = "/etc/traefik/certs/my.domain.com.pem"
registry_nginx['ssl_certificate_key'] = "/etc/traefik/private/letsencrypt.key"
```

ldez:
have you tried with the new version ? https://github.com/containous/traefik/pull/2161

frederik:
I have now. This works for me. 👌

poVoq:
Getting an error in line 71 with this
```
./dumpcerts.sh /opt/traefik/acme.json /opt/traefik/certs/
./dumpcerts.sh: Zeile 71: $1 ist nicht gesetzt.
```
"ist nicht gesetzt" means "is not set"

torarnv:
Couldn’t this be a built in feature of traefik? Dumping the certs to a volume to be shared live with other software? Why require an external cron-driven export command to be issued, adding complexity for the use-case?

ldez:
@torarnv https://github.com/containous/traefik/issues/1512

torarnv:
@ldez Maybe I'm missing something, but #1512 still looks like a command that needs to be issued via `docker exec` in a running Træfik container, e.g. `traefik exportACME /etc/ssl`. Which means I need to drive it from a cron-job or similar manually, instead of just relying on Træfik to always keep a volume up to date with live certs. Isn't it just a Træfik built-in to replace the contrib-script?

PR #1485, opened 2017-04-24
Switch to new access logger and add JSON format

rjshep:
Following on from [#1408](https://github.com/containous/traefik/pull/1408) this PR adds the remaining functionality (and tests) for access logging including appenders for the current (extended common) log format and JSON.


server/server.go
```diff
@@ -175,7 +182,7 @@ func (server *Server) startHTTPServers() {
 	server.serverEntryPoints = server.buildEntryPoints(server.globalConfiguration)
 
 	for newServerEntryPointName, newServerEntryPoint := range server.serverEntryPoints {
-		serverMiddlewares := []negroni.Handler{server.accessLoggerMiddleware, server.loggerMiddleware, metrics}
+		serverMiddlewares := []negroni.Handler{server.accessLoggerMiddleware, metrics}
```
emilevauge: `server.accessLoggerMiddleware` always enabled ? And what about `saveFrontend` and `saveBackend`?

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,235 @@
+package accesslog
```
emilevauge: Just to be sure, is this code from you or did you copied it from another place (just want to avoid license issue ;) )?

server/configuration.go
```diff
@@ -40,6 +40,7 @@ type GlobalConfiguration struct {
 	Debug                     bool                    `short:"d" description:"Enable debug mode"`
 	CheckNewVersion           bool                    `description:"Periodically check if a new version has been released"`
 	AccessLogsFile            string                  `description:"Access logs file"`
```
rjshep: Done

server/configuration.go
```diff
@@ -476,6 +477,7 @@ func NewTraefikConfiguration() *TraefikConfiguration {
 		GlobalConfiguration: GlobalConfiguration{
 			GraceTimeOut:              flaeg.Duration(10 * time.Second),
 			AccessLogsFile:            "",
+			AccessLog:                 &types.AccessLog{},
```
rjshep: Agreed - removed

middlewares/accesslog/json_log_formatter.go
rjshep: It was written by one of my colleagues, rather than me personally, not copied from anywhere else.

server/server.go
rjshep: I've updated the code such that the `server.accessLoggerMiddleware` `saveFrontend` and `saveBackend` are only used if access logging is configured.

mildis:
Hi @emilevauge and @rjshep,

The idea behind #1381 was to log X-Forwarded-* request headers if present.
Theses headers (‘X-Forwarded-For: <comma-separated list of IP>’, ‘X-Forwarded-Proto: <original_proto>’ and ‘X-Forwarded-Host: <original_host>’) are set by upstream servers when traefik is behind a reverse-proxy or a load-balancer.

These are supposed to be superseded by the [Forwarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) header but I’ve never seen it in the wild.

From what I see in the new access logger, these headers could be stored in the Logdata structure if present to be used in a custom log format if wanted.

rjshep:
@mildis Yes the forwarded headers could be added into the `CoreLogData` map and then set within `access log.Logger.ServeHTTP`.  Rather than me adding more to what's already a large PR, can I suggest you wait until this PR is accepted and merged and then submit your own PR with the necessary changes?

mildis:
@rjshep Yes I'm fine with that. I'm new to Go : it will also be easier for me to get where should I change the code once integrated.

middlewares/accesslog/json_log_formatter.go
emilevauge: :+1:

middlewares/accesslog/logger_test.go
```diff
@@ -0,0 +1,435 @@
...
+		}
+		assert.False(t, scanner.Scan())
+	}
+}
+
+//-------------------------------------------------------------------------------------------------
+
+func simpleHandlerFunc(rw http.ResponseWriter, r *http.Request) {
+	if r.Body != nil {
+		io.Copy(ioutil.Discard, r.Body)
```
timoreimann: Why do we do this?

middlewares/accesslog/logger_test.go
```diff
@@ -0,0 +1,435 @@
...
+	}
+	time.Sleep(2 * time.Millisecond)
+	rw.Header().Set("Content-Length", strconv.Itoa(len(textMessage)))
+	rw.Header().Set("Content-Type", "text/html")
+	rw.Write([]byte(textMessage))
+	rw.WriteHeader(200)
+}
+
+func redirHandlerFunc(rw http.ResponseWriter, r *http.Request) {
+	time.Sleep(time.Millisecond)
```
timoreimann: Can we say `1 * time.Millisecond` for consistency reasons?

types/unitparser.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/pkg/errors"
+)
+
+// AsSI parses a string for its number. Suffixes are allowed that loosely follow SI rules: K, Ki, M, Mi.
+// 'k' and 'K' are equivalent.
+// Example: "2 KiB" returns 2048, "B", nil
+func AsSI(value string) (int64, string, error) {
+	if value == "" {
+		return 0, "", errors.New("Blank value")
```
timoreimann: Why `github.com/pkg/errors` as opposed to just `errors.New` from stdlib?

types/unitparser_test.go
```diff
@@ -0,0 +1,28 @@
+package types
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAsSI_happy(t *testing.T) {
```
timoreimann: Please name `TestAsSIHappy`.

timoreimann: How about `TestAsSIUnhappy`?

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,235 @@
...
+
+// Exiter provides a facade for os.Exit.
+type Exiter interface {
+	Exit(code int)
+}
+
+type stdExiter struct{}
+
+func (stdExiter) Exit(code int) {
+	os.Exit(code)
```
emilevauge: We should *never* call `Exit` from the code, please return error instead.

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,235 @@
...
+		ok = requireNoDuplicates(jlf.originMapping) && ok
+	}
+
+	if len(settings.DownstreamResponseHeaders) > 0 {
+		jlf.downstreamMapping = convertFieldsToMappings(settings.DownstreamResponseHeaders)
+		ok = requireNoDuplicates(jlf.downstreamMapping) && ok
+	}
+
+	if !ok {
+		exiter.Exit(1)
```
emilevauge: Wow good catch @timoreimann

rjshep:
Thanks for the comments @timoreimann.  I'll work through the necessary changes over the next day or so.

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,235 @@
...
+				strings.TrimSpace(s[:colon]),
+				strings.TrimSpace(s[colon+1:]),
+			}
+		} else {
+			ws := strings.TrimSpace(s)
+			t = tuple{ws, ws}
+		}
+
+		mapping = append(mapping, t)
+	}
```
rjshep: Couldn't get this to work in a way that used less code than was already there.

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,235 @@
...
+		var t tuple
+
+		if colon > 0 {
+			t = tuple{
+				strings.TrimSpace(s[:colon]),
+				strings.TrimSpace(s[colon+1:]),
+			}
+		} else {
+			ws := strings.TrimSpace(s)
+			t = tuple{ws, ws}
```
rjshep: Added a check.

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,235 @@
...
+
+func asSeconds(d time.Duration) float64 {
+	// half-rounded up to the nearest millisecond, then converted to seconds
+	return float64((d.Nanoseconds()+500000)/1000000) / 1000
+}
+
+func (l jsonLogFormatter) writeHeaders(buf *bytes.Buffer, mapping []tuple, hdr http.Header) {
+	for _, tup := range mapping {
+		v := hdr[tup.sourceKey]
+		if len(v) > 0 {
```
rjshep: Done

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,235 @@
...
+			l.writeField(buf, tup.translatedKey, v[0])
+			// note that multi-valued headers are not fully reported
```
rjshep: Added to docs

middlewares/accesslog/linear_writer_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLinearWriter(t *testing.T) {
+	file := logfilePath("")
+
+	w, err := os.Create(file)
+	assert.Nil(t, err, "%v", err)
```
rjshep: Added defer

middlewares/accesslog/linear_writer_test.go
```diff
@@ -0,0 +1,59 @@
...
+	counts := make([]int, n)
+	r, err := os.Open(file)
+	assert.Nil(t, err, "%v", err)
+	defer r.Close()
+	defer os.Remove(file)
+
+	scanner := bufio.NewScanner(r)
+	for i := 0; i < n; i++ {
+		assert.True(t, scanner.Scan())
+		line := scanner.Text()
```
rjshep: I think this implementation is quite simple.  I had a quick try with ReadFile but it seemed more complicated.

middlewares/accesslog/linear_writer_test.go
```diff
@@ -0,0 +1,59 @@
...
+		assert.Nil(t, err, "%v", err)
+		if 0 <= num && num < n {
+			counts[num]++
+		} else {
+			assert.Fail(t, "Wrong line number", "Got %d", num)
+		}
+	}
+
+	for i := 0; i < n; i++ {
+		assert.Equal(t, 1, counts[i], "Line %d should have count=1 but was %d", i, counts[i])
```
rjshep: If it goes inside the loop then the assertion fails.

middlewares/accesslog/log_appender.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	return nil
+}
+
+func (l *LogAppender) flushBufferAndCloseFile() error {
+	log.Debugf("Closing access log %s (%s)", l.settings.File, l.settings.Format)
+	var err error
+	if f, ok := l.buf.(Flusher); ok {
+		err = f.Flush()
+		if err != nil {
```
rjshep: Done

middlewares/accesslog/log_appender.go
```diff
@@ -0,0 +1,200 @@
...
+			return err
+		}
+	}
+	l.buf = nil
+
+	if c, ok := l.file.(io.Closer); ok {
+		err = c.Close()
+	}
+	l.file = nil
+	return err
```
rjshep: No it could still be an error from `c.Close()`

middlewares/accesslog/log_appender_test.go
```diff
@@ -0,0 +1,214 @@
...
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/containous/traefik/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLogAppenderSimpleFile(t *testing.T) {
```
rjshep: I've simplified the tests by reducing duplicated code.

middlewares/accesslog/log_appender_test.go
```diff
@@ -0,0 +1,214 @@
...
+	file := logfilePath("")
+	defer os.Remove(file)
+
+	settings := &types.AccessLog{File: file}
+	la, err := NewLogAppender(settings)
+	assert.Nil(t, err)
+
+	n := 1000
```
rjshep: This code was written by a colleague before he left the team, so I can't say for certain why 1000 was picked.

types/unitparser.go
rjshep: Changed to stdlib errors

emilevauge:
Hey @rjshep, we just tagged `v1.3.0-rc1`. As we had issues while reporting branch `v1.2` into master due to your PR #1408, we decided to revert it in #1541.
We will re-apply it in the coming days.

**In the meantime, please don't rebase this PR for now**

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,235 @@
...
+
+// exiter provides a seam for testing code containing os.Exit.
+var exiter Exiter = stdExiter{}
+
+//-------------------------------------------------------------------------------------------------
+
+func (l jsonLogFormatter) Write(w io.Writer, logDataTable *LogData) error {
+
+	buf := &bytes.Buffer{}
+	buf.WriteString("{")
```
rjshep: I've spent time looking at converting this code to use the inbuilt Go JSON routines, but as Rick discussed in the [original](https://github.com/containous/traefik/pull/1293#discussion_r109527128) PR, the reason for the current implementation approach is the desire to format the JSON value (e.g. Duration) in a particular way.  Because of this, simply adding everything to a map and then calling `json.Marshal` results in JSON without that formatting.  I guess if we had a defined struct for the log data then we could use a custom `Marshaler`, however that would add more complexity because of the need to drop/rename certain fields based on the requested output format.

I understand your concern around hand-crafting JSON.  It's not normally something I would advocate, however it does appear to be the most sensible solution in this case.  We're not creating complex structures, therefore the risk of an error being introduced is low and the code to be maintained is required to produce the output in the desired format.

timoreimann: > [...] the reason for the current implementation approach is the desire to format the JSON value (e.g. Duration) in a particular way. Because of this, simply adding everything to a map and then calling`json.Marshal` results in JSON without that formatting. I guess if we had a defined struct for the log data then we could use a custom `Marshaler`, however that would add more complexity because of the need to drop/rename certain fields based on the requested output format.

The way I see it is that we'd only need to implement custom Marshalers for those types that have custom output needs. For everything else, the default marshalling implementation already does what we need, and it's possible to mix custom and non-custom parts arbitrarily. There's also no need to construct the JSON syntax itself since Go maps and slices already provide that.

That said, talk is cheap, and I haven't implemented anything out. Though I still have my concerns, I'm okay with moving forward with the current approach if @emilevauge is too. We can always revisit our decision should the future show that this very part becomes a source of complexity.

middlewares/accesslog/log_appender.go
timoreimann: True, missed that.

middlewares/accesslog/linear_writer_test.go
timoreimann: Ok.

middlewares/accesslog/json_log_formatter.go
timoreimann: This slight variation of my previous code snippet (this time pleasing the compiler) seems to work -- at least, the tests continue to go green:

```go
fields := strings.SplitN(s, ":", 2)
t := tuple{sourceKey: strings.TrimSpace(fields[0]), translatedKey: strings.TrimSpace(fields[0])}
if len(fields) == 2 {
  t.translatedKey = strings.TrimSpace(fields[1])
}
```

Would this work, or am I missing something?

types/unitparser_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}{
+		{"1", "", 1},
+		{"1 km", "m", 1000},
+		{"1KB", "B", 1000},
+		{"2 KiB", "B", 2048},
+		{"1MJ", "J", 1000000},
+		{"2 MiB", "B", 2 * 1024 * 1024},
+	}
+
+	for _, c := range cases {
```
timoreimann: Let's use sub-tests parameterized with `in` and `unit` so that we can easily see which test case fails.

types/unitparser_test.go
```diff
@@ -0,0 +1,50 @@
...
+	cases := []struct {
+		in, unit string
+		exp      int64
+		err      bool
+	}{
+		{"", "", 0, true},
+		{"1x", "", 1, false},
+	}
+
+	for _, c := range cases {
```
timoreimann: I'd propose sub-tests here too.

middlewares/accesslog/json_log_formatter_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	assert.True(t, err == nil)
+
+	buf := &bytes.Buffer{}
+	jlf.Write(buf, fixtureLogDataTable(12345))
+	s := buf.String()
+	assert.Equal(t,
+		`{"StartUTC":"`,
+		s[:13])
+	assert.Equal(t,
```
timoreimann: What's the reason we're not testing the content in the [14, 39) interval?

middlewares/accesslog/linear_writer_test.go
```diff
@@ -0,0 +1,60 @@
...
+	wg.Wait()
+
+	err = lw.Close()
+	assert.Nil(t, err, "%v", err)
+
+	// read back and verify
+	counts := make([]int, n)
+	r, err := os.Open(file)
+	assert.Nil(t, err, "%v", err)
+	defer r.Close()
```
timoreimann: [Deferred function calls are executed in Last In First Out order after the surrounding function returns](https://blog.golang.org/defer-panic-and-recover). So I guess we should swap the order of the two defer statements?

server/server_test.go
```diff
@@ -0,0 +1,70 @@
...
+		{
+			desc: "Test with AccessLog set",
+			globalCfg: GlobalConfiguration{AccessLog: &types.AccessLog{
+				File:       "foo.log",
+				Format:     "json",
+				TimeFormat: "02/Jan/2006:15:04:05 -0700",
+				BufferSize: "4 KiB",
+				GzipLevel:  1,
+			},
+			},
```
timoreimann: Please join with the previous line.

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,210 @@
...
+	"io"
+	"math"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/types"
+	"github.com/pkg/errors"
```
rjshep: Done

middlewares/accesslog/json_log_formatter.go
```diff
@@ -0,0 +1,210 @@
...
+		ok = requireNoDuplicates(jlf.originMapping) && ok
+	}
+
+	if len(settings.DownstreamResponseHeaders) > 0 {
+		jlf.downstreamMapping = convertFieldsToMappings(settings.DownstreamResponseHeaders)
+		ok = requireNoDuplicates(jlf.downstreamMapping) && ok
+	}
+
+	if !ok {
+		return jlf, errors.New("Errors in JSON logging configuration")
```
rjshep: Done

rjshep: Agreed, I've changed this as suggested.

middlewares/accesslog/json_log_formatter_test.go
```diff
@@ -0,0 +1,135 @@
...
+			"ClientHost:ClHost",
+			"ClientPort:ClPort",
+			"ClientUsername:ClUsername",
+		},
+		RequestHeaders:            []string{"User-Agent: user_agent", "Referer: referrer"},
+		OriginResponseHeaders:     []string{"Server: upstream_http_server"},
+		DownstreamResponseHeaders: []string{"Location: sent_http_location"},
+	})
+
+	assert.True(t, err == nil)
```
rjshep: Done

middlewares/accesslog/json_log_formatter_test.go
rjshep: It's the timestamp, which is obviously different for each run.

middlewares/accesslog/json_log_formatter_test.go
```diff
@@ -0,0 +1,135 @@
...
+			`"OriginStatus":200,"DownstreamContentSize":82,"RequestCount":12345,`+
+			`"ClHost":"190.190.190.190","ClPort":"20121","ClUsername":"-",`+
+			`"user_agent":"user-agent-very-very-long-string","referrer":"http://example.com/x/y/z",`+
+			`"upstream_http_server":"foobar v1",`+
+			`"sent_http_location":"http://somewhere.else/a/b"`+"}\n",
+		s[39:])
+}
+
+func TestGzipRatioDivideByZero(t *testing.T) {
+	jlf, _ := newJSONLogFormatter(&types.AccessLog{
```
rjshep: Done

middlewares/accesslog/json_log_formatter_test.go
```diff
@@ -0,0 +1,135 @@
...
+	defer log.SetOutput(os.Stderr)
+
+	_, err := newJSONLogFormatter(&types.AccessLog{
+		CoreFields:                []string{"Duration", "StartUTC", "FrontendName:frontend", "NonExistent"},
+		RequestHeaders:            []string{"Host: http_host"},
+		OriginResponseHeaders:     []string{"Server: upstream_http_server"},
+		DownstreamResponseHeaders: []string{"Location: sent_http_location"},
+	})
+
+	assert.NotNil(t, err)
```
rjshep: Done

middlewares/accesslog/json_log_formatter_test.go
```diff
@@ -0,0 +1,135 @@
...
+	defer log.SetOutput(os.Stderr)
+
+	_, err := newJSONLogFormatter(&types.AccessLog{
+		CoreFields:                []string{" ", " : "},
+		RequestHeaders:            []string{" ", " : "},
+		OriginResponseHeaders:     []string{" ", " : "},
+		DownstreamResponseHeaders: []string{" ", " : "},
+	})
+
+	assert.NotNil(t, err)
```
rjshep: Done

middlewares/accesslog/json_log_formatter_test.go
```diff
@@ -0,0 +1,135 @@
...
+	defer log.SetOutput(os.Stderr)
+
+	_, err := newJSONLogFormatter(&types.AccessLog{
+		CoreFields:                []string{"Duration:x", "Duration:y", "FrontendName:end", "BackendName:end"},
+		RequestHeaders:            []string{"Host: http_host", "Host: http_host"},
+		OriginResponseHeaders:     []string{"Server: server", "Server: server"},
+		DownstreamResponseHeaders: []string{"Location: location", "Location: location"},
+	})
+
+	assert.NotNil(t, err)
```
rjshep: Done

middlewares/accesslog/log_appender_test.go
```diff
@@ -0,0 +1,173 @@
...
+		for i := 1; i <= n; i++ {
+			err = la.Write(fixtureLogDataTable(uint64(i)))
+			assert.Nil(t, err, "%v", err)
+		}
+
+		err = la.Close()
+		assert.Nil(t, err, "%v", err)
+
+		validateResult(t, n, file, true)
+		os.Remove(file)
```
rjshep: Done

middlewares/accesslog/log_formatter_test.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"bytes"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func init() {
```
rjshep: Removed

middlewares/accesslog/logger_test.go
```diff
@@ -0,0 +1,431 @@
...
+var (
+	logfileDir       string
+	parsedBackendURL *url.URL
+	allCoreKeySlice  []string
+)
+
+func init() {
+	logfileDir = os.TempDir()
+	var err error
+	if parsedBackendURL, err = url.Parse(testBackendURL); err != nil {
```
rjshep: Done

server/server_test.go
rjshep: Done

server/server_test.go
```diff
@@ -0,0 +1,70 @@
...
+				TimeFormat: "02/Jan/2006:15:04:05 -0700",
+				BufferSize: "4 KiB",
+				GzipLevel:  1,
+			},
+			accessLogMiddlewareSet: true,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.desc, func(t *testing.T) {
```
rjshep: Thanks! Done.

types/unitparser_test.go
rjshep: Done

types/unitparser_test.go
rjshep: Done

middlewares/accesslog/linear_writer_test.go
rjshep: Ah, didn't know that! Done

middlewares/accesslog/json_log_formatter.go
rjshep: Yes that works fine.  I've used it in place of the original.

timoreimann:
@rjshep Travis CI seems to hang -- it sometimes does and only seems to get fixed by another (hash-modifying) push.

Maybe you can take the opportunity to squash the commits.

timoreimann:
@rjshep IIRC, you should refrain from rebasing until #1571 merges. I'll make sure I review that one by end of tomorrow too.

middlewares/accesslog/json_log_formatter_test.go
timoreimann: Would it then be possible to have an assertion to verify that whatever is at the position of the timestamp matches the format we're expecting?

ldez:
We are ready, #1571 is merge! It's time to rebase! 😃

timoreimann:
@rjshep a nice trick to push a hash-changing commit without actually changing content is to do this:

`git commit --amend --no-edit`

rjshep:
Thanks @timoreimann that's what I did - Googling suggested it too!

ldez:

middlewares/accesslog/json_log_formatter_test.go
timoreimann: @rjshep ping. ^ :-)

ldez:
@rjshep Awesome job 🎉 

Could you squash your commits?

emilevauge:
@rjshep

After a long discussion with other maintainers, I'm very sorry to announce that we will finally not merge this PR.

Let me explain why:
1. It is really big: ~2000 lines added is really hard to review. We (maintainers) need to be confident in each PR we merge because we need to maintain this code later. For the years to come.
2. It is really low level. Reimplementing all the log appender/formatter things into Traefik is not what we prefer. A lot of libraries are doing this very well and we think we should not reinvent the wheel. We would also prefer not having a custom JSON marshaller and instead marshall a simple struct.
3. We don't really need the complexity of being able to configure custom fields. At least in the first iteration. Most people will probably just want to push the whole logs into a log aggregator.
4. We realized that we can achieve the same thing using Logrus but with much less code, more flexibility, and better support. We already use it for application logs. It offers a simple way to manage structured logs and already has a JSON formatter. Besides, Logrus offers several hooks that allow to export logs to a lot of useful backends (Elasticsearch, Sentry, Syslog...).

We failed.
We failed because our process failed.
We didn't took the time to discuss on a real design review between maintainers before making code reviews.
You work really hard on following our recommandations after each reviews.
As your PR was really big, we couldn't took as much time as it was required to deeply review it.
We should have told you on the first day that we wouldn't merge it because of the size.
We failed. And I'm really sorry for this.

We will not make the same mistake again in the future (for sure), and we are working on it: the process is being reviewed, we will automate a lot of time consuming tasks, this will give more time to make deep reviews, etc.

We would perfectly understand if you were angry at us as you made a really great job and we sucked...
Please accept our apologies, we are small team, and will get better at this.

emilevauge:
To be continued here #1647

PR #1147, opened 2017-02-09
Add Basic Auth per Frontend

SantoDE:
Fixes #751

Currently, only Basic Auth and the docker & rancher provider are supported.

PR #1488, opened 2017-04-24
Add basic auth to kubernetes provider

alpe:
This will add the kubernetes provider for basic authentication. Thanks to @SantoDE for #1147.

## Some constraints: 
* Basic authentication only
* Realm not configurable; only `traefik`default
* Secret is in same namespace as ingress rule
* Secret must contain only single file (*)

(*) otherwise we can have some conventions here like key name (pattern)

## Example
Based on https://docs.traefik.io/user-guide/kubernetes/
* Encode credentials
`htpasswd -c ./auth jane`
* Deploy a secret to k8s
`kubectl --namespace=kube-system create secret generic mysecret --from-file auth`
* Deploy ingress
```yaml
apiVersion: v1
kind: Service
metadata:
  name: traefik-web-ui
  namespace: kube-system
spec:
  selector:
    k8s-app: traefik-ingress-lb
  ports:
  - port: 80
    targetPort: 8080
---
kind: Ingress
metadata:
  name: traefik-web-ui
  namespace: kube-system
  annotations:
    ingress.kubernetes.io/auth-type: "basic"
    ingress.kubernetes.io/auth-secret: "mysecret"
spec:
  rules:
  - host: traefik-ui.local
    http:
      paths:
      - backend:
          serviceName: traefik-web-ui
          servicePort: 80
```
Note the two annotations to enable basic auth and link the secret

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -1,6 +1,9 @@
 package kubernetes
 
 import (
+	"bufio"
+	"bytes"
+	"errors"
 	"fmt"
 	"os"
 	"reflect"
@@ -16,6 +19,7 @@ import (
 	"github.com/containous/traefik/safe"
 	"github.com/containous/traefik/types"
 	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/pkg/apis/extensions/v1beta1"
 	"k8s.io/client-go/pkg/util/intstr"
 )
 
@@ -28,6 +32,8 @@ const (
 	ruleTypePath               = "Path"
 	ruleTypePathPrefix         = "PathPrefix"
 )
+
+const traefikDefaultRealm = "traefik"
 
 // Provider holds configurations of the provider.
 type Provider struct {
@@ -159,13 +165,21 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 				default:
 					log.Warnf("Unknown value of %s for traefik.frontend.passHostHeader, falling back to %s", passHostHeaderAnnotation, PassHostHeader)
 				}
+				if realm := i.Annotations["ingress.kubernetes.io/auth-realm"]; realm != "" && realm != traefikDefaultRealm {
+					return nil, errors.New("no realm customization supported")
+				}
 
 				if _, exists := templateObjects.Frontends[r.Host+pa.Path]; !exists {
+					basicAuthCreds, err := handleBasicAuthConfig(i, k8sClient)
+					if err != nil {
+						return nil, err
+					}
 					templateObjects.Frontends[r.Host+pa.Path] = &types.Frontend{
 						Backend:        r.Host + pa.Path,
 						PassHostHeader: PassHostHeader,
 						Routes:         make(map[string]types.Route),
 						Priority:       len(pa.Path),
+						BasicAuth:      basicAuthCreds,
 					}
 				}
 				if len(r.Host) > 0 {
@@ -276,6 +290,56 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 		}
 	}
 	return &templateObjects, nil
+}
+
+func handleBasicAuthConfig(i *v1beta1.Ingress, k8sClient Client) ([]string, error) {
+	authType, exists := i.Annotations["ingress.kubernetes.io/auth-type"]
+	if !exists {
+		return nil, nil
+	}
+	if strings.ToLower(authType) != "basic" {
+		return nil, fmt.Errorf("unsupported auth-type: %q", authType)
+	}
+	authSecret := i.Annotations["ingress.kubernetes.io/auth-secret"]
+	if authSecret == "" {
+		return nil, errors.New("auth-secret annotation must be set")
+	}
+	basicAuthCreds, err := loadAuthCredentials(i.Namespace, authSecret, k8sClient)
+	if err != nil {
+		return nil, err
+	}
+	if len(basicAuthCreds) == 0 {
+		return nil, errors.New("secret file without credentials")
+	}
+	return basicAuthCreds, nil
+}
+
+func loadAuthCredentials(namespace, secretName string, k8sClient Client) ([]string, error) {
+	secret, ok, err := k8sClient.GetSecret(namespace, secretName)
+	switch { // keep order of case conditions
+	case err != nil:
+		return nil, fmt.Errorf("failed to fetch secret %q/%q: %s", namespace, secretName, err)
+	case !ok:
+		return nil, fmt.Errorf("secret %q/%q not found", namespace, secretName)
+	case secret == nil:
+		return nil, errors.New("secret data must not be nil")
+	case len(secret.Data) != 1:
+		return nil, errors.New("secret must contain single element only")
+	default:
+	}
+	var firstSecret []byte
+	for _, v := range secret.Data {
+		firstSecret = v
+		break
+	}
+	creds := make([]string, 0)
+	scanner := bufio.NewScanner(bytes.NewReader(firstSecret))
+	for scanner.Scan() {
+		if cred := scanner.Text(); cred != "" {
+			creds = append(creds, cred)
+		}
+	}
+	return creds, nil
 }
 
 func endpointPortNumber(servicePort v1.ServicePort, endpointPorts []v1.EndpointPort) int {
--- a/provider/kubernetes/kubernetes_test.go
+++ b/provider/kubernetes/kubernetes_test.go
@@ -1470,6 +1470,35 @@ func TestIngressAnnotations(t *testing.T) {
 			ObjectMeta: v1.ObjectMeta{
 				Namespace: "testing",
 				Annotations: map[string]string{
+					"ingress.kubernetes.io/auth-type":   "basic",
+					"ingress.kubernetes.io/auth-secret": "mySecret",
+				},
+			},
+			Spec: v1beta1.IngressSpec{
+				Rules: []v1beta1.IngressRule{
+					{
+						Host: "basic",
+						IngressRuleValue: v1beta1.IngressRuleValue{
+							HTTP: &v1beta1.HTTPIngressRuleValue{
+								Paths: []v1beta1.HTTPIngressPath{
+									{
+										Path: "/auth",
+										Backend: v1beta1.IngressBackend{
+											ServiceName: "service1",
+											ServicePort: intstr.FromInt(80),
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			ObjectMeta: v1.ObjectMeta{
+				Namespace: "testing",
+				Annotations: map[string]string{
 					"kubernetes.io/ingress.class": "somethingOtherThanTraefik",
 				},
 			},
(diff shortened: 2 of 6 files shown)


provider/kubernetes/kubernetes.go
```diff
@@ -278,6 +305,35 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+func loadAuthCredentials(namespace, secretName string, k8sClient Client) ([]string, error) {
+	secret, ok, err := k8sClient.GetSecret(namespace, secretName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch secret %q/%q: %s", namespace, secretName, err)
+	}
+	if !ok {
+		return nil, fmt.Errorf("secret %q/%q not found", namespace, secretName)
+	}
+	if secret == nil {
+		return nil, errors.New("secret data must nost be nil")
```
errm: -nost- not

provider/kubernetes/kubernetes.go
```diff
@@ -159,13 +164,35 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					if authSecret == "" {
+						return nil, errors.New("auth-secret annotation must be set")
+					}
+					var err error
+					basicAuthCreds, err = loadAuthCredentials(i.Namespace, authSecret, k8sClient)
+					if err != nil {
+						return nil, err
+					}
+					if len(basicAuthCreds) == 0 {
+						return nil, errors.New("secet file without credentials")
```
timoreimann: Typo (secet -> secret)

provider/kubernetes/kubernetes.go
```diff
@@ -159,13 +164,35 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					}
+				} else if authType != "" {
+					return nil, fmt.Errorf("unsupported auth-type: %q", authType)
+				}
```
timoreimann: Could we please put the entire block in a separate function?

`loadIngresses` is already pretty big.

provider/kubernetes/kubernetes.go
```diff
@@ -159,13 +164,35 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					}
+					var err error
+					basicAuthCreds, err = loadAuthCredentials(i.Namespace, authSecret, k8sClient)
+					if err != nil {
+						return nil, err
+					}
+					if len(basicAuthCreds) == 0 {
+						return nil, errors.New("secet file without credentials")
+					}
+				} else if authType != "" {
```
timoreimann: Can we do this check before the happy path so that we don't need the `else`?

provider/kubernetes/kubernetes.go
```diff
@@ -278,6 +305,35 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+	}
+	if !ok {
+		return nil, fmt.Errorf("secret %q/%q not found", namespace, secretName)
+	}
+	if secret == nil {
+		return nil, errors.New("secret data must not be nil")
+	}
+	if len(secret.Data) != 1 {
+		return nil, errors.New("secret must contain single element only")
+	}
```
timoreimann: This block lends itself to a switch statement IMHO.

provider/kubernetes/kubernetes.go
```diff
@@ -278,6 +305,35 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+func loadAuthCredentials(namespace, secretName string, k8sClient Client) ([]string, error) {
+	secret, ok, err := k8sClient.GetSecret(namespace, secretName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch secret %q/%q: %s", namespace, secretName, err)
+	}
+	if !ok {
+		return nil, fmt.Errorf("secret %q/%q not found", namespace, secretName)
+	}
+	if secret == nil {
+		return nil, errors.New("secret data must not be nil")
```
timoreimann: Can we rephrase this without the `nil` for the user? What does a `nil` secret imply?

provider/kubernetes/kubernetes.go
```diff
@@ -278,6 +305,35 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+	}
+	if len(secret.Data) != 1 {
+		return nil, errors.New("secret must contain single element only")
```
timoreimann: Improvement suggestion: "multiple secrets are not supported".

provider/kubernetes/kubernetes.go
```diff
@@ -278,6 +305,35 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+	}
+	var firstElement []byte
+	for _, v := range secret.Data {
+		firstElement = v
+		break
+	}
```
timoreimann: Why not `firstElement := secret.Data[0]`? We have already asserted there's only a single element.

provider/kubernetes/kubernetes.go
```diff
@@ -278,6 +305,35 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+	if !ok {
+		return nil, fmt.Errorf("secret %q/%q not found", namespace, secretName)
+	}
+	if secret == nil {
+		return nil, errors.New("secret data must not be nil")
+	}
+	if len(secret.Data) != 1 {
+		return nil, errors.New("secret must contain single element only")
+	}
+	var firstElement []byte
```
timoreimann: Variable name is not ideal. Maybe `firstSecret` instead?

provider/kubernetes/kubernetes_test.go
```diff
@@ -2085,9 +2154,97 @@ func TestMissingResources(t *testing.T) {
...
+	provider := Provider{}
+	actual, err := provider.loadIngresses(client)
+	if err != nil {
+		t.Fatalf("error %+v", err)
+	}
+
+	actual = provider.loadConfig(*actual)
+	got := actual.Frontends["basic/auth"].BasicAuth
+	if !reflect.DeepEqual(got, []string{"myUser:myEncodedPW"}) {
+		t.Fatalf("unexpected creadentials: %+v", got)
```
timoreimann: Typo.

provider/kubernetes/kubernetes_test.go
```diff
@@ -2121,6 +2278,19 @@ func (c clientMock) GetService(namespace, name string) (*v1.Service, bool, error
 	return nil, false, nil
 }
 
+func (c clientMock) GetSecret(namespace, name string) (*v1.Secret, bool, error) {
+	if c.apiServiceError != nil {
```
timoreimann: This should not reuse `apiServiceError` but get its own field.

provider/kubernetes/kubernetes.go
alpe: Why can this be nil anyway when `ok` is true? I was wondering a lot when I ran into this but I did not spent time investigating.

provider/kubernetes/kubernetes.go
alpe: Data can have 0 elements, too. I think this is a temporary solution until we have some convention for a key name or pattern.

provider/kubernetes/kubernetes.go
alpe: The order of the conditions matter. The switch works top down when multiple conditions are true. So that's fine but it feels more fragile that I'll add note to keep the order.

provider/kubernetes/kubernetes.go
alpe: `Data` is a map

provider/kubernetes/kubernetes.go
alpe: The condition was `authType != "" && strings.ToLower(authType) == "basic"` for the error. The else block made sense but it's better in the extracted method.

alpe:
I have applied most of the review comments @timoreimann requested. Thanks for the feedback! The  test coverage and documentation can be better. I may be able to do a little bit more end of the week.

provider/kubernetes/kubernetes.go
timoreimann: My guess is it won't ever be `nil` unless `ok` is false or `err` is non-nil. The pointer is probably for performance  / in-place-update reasons.

provider/kubernetes/kubernetes.go
timoreimann: Ok.

provider/kubernetes/kubernetes.go
timoreimann: Ah true, missed that. 👍

timoreimann:
Thanks @alpe, looking pretty good now.

A bit of documentation (especially with regards to the limitations mentioned in the PR) and coverage of the important unhappy paths would be great to have. 👍

emilevauge:
Ping @alpe

ldez:
Documentation added

kekoav:
@alpe @timoreimann I know this is done but I was reviewing due to #1596 and saw this:
https://github.com/containous/traefik/blob/master/provider/kubernetes/client.go#L266 
```
// fireEvent checks if all controllers have synced before firing
// Used after startup or a reconnect
func (c *clientImpl) fireEvent(event interface{}, eventCh chan interface{}) {
	if !c.ingController.HasSynced() || !c.svcController.HasSynced() || !c.epController.HasSynced() {
		return
	}
	eventHandlerFunc(eventCh, event)
}
```

Was excluding the new `secController` from this check for "all controllers" intentional?

timoreimann:
Side question: What do we need the `secController` for exactly?

I'm really not too deep into that part of the code. What I know though is that we use various events as a trigger to poll the k8s APIs. Does the `secController` make sure we do that for Secrets events as well? If so, can such events ever indicate a state change relevant for Traefik?

shadycuz:
@alpe These instuctions
```
Encode credentials
htpasswd -c ./auth jane
Deploy a secret to k8s
kubectl --namespace=kube-system create secret generic mysecret --from-file auth
```
are not on the traefik website and it was very hard to eventually track down this PR and find the example for creating the secret.

alpe:
@shadycuz Apologies, this could have been better. I had very limited time to work on this when I contributed the code. I'm happy that you find it useful. I am very sure the community will appreciate it when you could help with some doc so that it is easier for others to setup the basic auth.

shadycuz:
@alpe I can do that from what I know and what I see in this PR, but I have other questions, for example can we generate the secret according to instructions for generating the web dashboard auth? 
>Passwords can be encoded in MD5, SHA1 and BCrypt: you can use htpasswd to generate those ones.

as long as its `USERNAME:ENCRYPTEDPASSWORD` ?

ldez:
https://docs.traefik.io/configuration/backends/kubernetes/#authentication

erkolson:
I think it is problematic to grant permission to `get` every secret in the kube-system namespace?  It seems like the RBAC constraints should be limited to a specific secret resource or pattern for the secret name, for example `treafik-*` ?

ldez:
@erkolson Could ask your question on the [Traefik community Slack channel](https://traefik.herokuapp.com) or [Stack Overflow](https://stackoverflow.com/questions/tagged/traefik) using the "traefik" tag

PR #1493, opened 2017-04-25
Fix Regex not working with PathStrip

kamphaus:
### This gives prefix stripping with regex capability

It reuses the regex generated for the route path matching in mux.
To get this regex I exported it in the mux package, so you need to pull the changes in that package as well.

This will close issue #928.


ldez:
Thank for the contribution, seems to be a duplicate of #1339

ldez:
See https://github.com/containous/traefik/pull/1339#issuecomment-297019893

Issue #1104, opened 2017-02-02
retry behaviour


bamarni:
_(follow-up from #1016)_

At the moment the condition for retrying a request is the following : 

``` go
func isNetworkError(status int) bool {
	return status == http.StatusBadGateway || status == http.StatusGatewayTimeout
}
```

Those response status can come either from upstream or be set by the [default oxy error handler](https://github.com/containous/oxy/blob/fcc76b52eb8568540a020b7a99e854d9d752b364/utils/handler.go#L18-L31). To compare this according to [proxy_next_upstream](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_next_upstream) semantics, that makes it like :

    error timeout http_502 http_504 non_idempotent

Personally I find it too open for a default, if upstream sends http 504 for a POST request it still might have altered a resource but didn't manage to serve the response in a timely manner. For instance with PHP-FPM and nginx I believe this is a common cause for 504 responses, and retrying those POST requests could cause duplicate data or other unexpected side effects.

Should we make this configurable? Or maybe make the pattern more strict? (to compare with nginx again, their default is only `error timeout`)

cc @emilevauge @SantoDE

jonaz:
Auch! i always assuemed traefik only retries GET. PUT and POST is really dangerous to retry...

timoreimann:
HAProxy has a similar default: Only [refused connections and timeouts](https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#retries) are eligible for retries.

I agree we should adapt a more safe default and consider making this configurable if there's a need for it.

timoreimann:
Looking at the definition of 502s and 504s more closely, the error codes are meant to be returned by servers _acting as a gateway or proxy_. In that sense, non-intermediaries should not emit such error codes AFAICS.

That's not to say we shouldn't change the behavior.

timoreimann:
I'm planning to work on this one next.

Which path should we follow? I see three options:

1. Change the existing logic to only retry on connection errors and connection timeouts (not counting established connections).
1. Change the existing logic as in 1. but allow to customize the codes on which Traefik should retry (additionally or exclusively).
1. Keep the existing logic and add an option to (additionally or exclusively) retry on connection errors/timeouts.

Personally, I'm in favor of 1. I'd rather not do 2. until/unless we get some feedback that users actually want a means to customize the codes. Otherwise, we risk maintaining code nobody wants.

[...]

PR #1495, opened 2017-04-25
Add a test of retry.go

tcolgate:
### Description

Looking at the code for retry.go, I suspect that retries of requests with Body's do not function as expected. The request.Body will may be drained by the previous attempts, the latter attempts will get an empty request body.

This test tries to demonstrate the problem.

The fix is tricky, since retrying inherently means buffering the Body, which has memory/performance implications. It may be that the behaviour should be user configurable somehow.

It's worth noting that http.ReverseProxy does not attempt to retry requests with non--nil Request.Body.

### DIFF
--- a/middlewares/retry_test.go
+++ b/middlewares/retry_test.go
@@ -1,0 +1,76 @@
+package middlewares
+
+import (
+	"bytes"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"testing"
+)
+
+func TestRetry(t *testing.T) {
+	tests := []struct {
+		rbody         *bytes.Buffer
+		rt            *Retry
+		responseCodes []int
+	}{
+		{
+			rt: &Retry{
+				attempts: 4,
+			},
+			rbody:         &bytes.Buffer{},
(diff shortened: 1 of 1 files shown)


tcolgate:
It may be that the two specific status codes used for the retry guarantee that data has not been consumed, But some tests are probably a good idea anyway.

timoreimann:
The error codes are supposed to be returned when the forwarding middleware has either failed to connect to the server or encountered a terminal issue during the connection setup phase. I suppose that the body isn't consumed in such cases.

Of course, this may not necessarily hold (though it should) when the two error codes in question are actually returned from an upstream server. #1104 will partially address this problem.

tcolgate:
closing this. I'm mostly trying to track down the multiple WriteHeader calls I'm seeing logged.

timoreimann:
@tcolgate I understood you are already certain this is an existing issue -- if so, could you please file a bug report so that we can keep track of things?

Thank you!

tcolgate:
@timoreimann I think the cause of the log noise I am seeing is in oxy https://github.com/vulcand/oxy/pull/27

PR #1497, opened 2017-04-25
Re-exclude /dist/traefik from .dockerignore.

timoreimann:
The master build is [currently broken](https://travis-ci.org/containous/traefik/jobs/225606534). Turns out we can't remove the .dockerfile exclusion for `/dist/traefik` since it's needed to build the Docker image.

This PR reverts that decision taken in #1470.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,3 @@
 dist/
+!dist/traefik
 site/


timoreimann:
Once again, the latest commit changing .travis.yml is only to show that the build should be fixed now. (We need this since `make image` is only called by `before_deploy` which seemingly isn't executed in PRs.) It will be removed once CI goes green and the review is done.

timoreimann:
CI is green, so fix is working.

PR #1501, opened 2017-04-27
doc: Add checklist to PR template

ldez:
### Description

Enhance PR template by adding a checklist.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,3 +11,15 @@ And the contributing guide: https://github.com/containous/traefik/blob/master/.g
 <!--
 Briefly describe the pull request in a few paragraphs.
 -->
+
+<!--
+ PLEASE, DON'T REMOVE THIS CHECKLIST AND COMPLETE IT.
+ -->
+- [x] Open a PR
+- [ ] Add a short title (57 chars max)
+- [ ] Write useful descriptions
+- [ ] Make sure your branch is up to date.
+- [ ] Add tests
+- [ ] Add documentation
+- [ ] Make sure the code builds
+- [ ] Make sure all tests pass


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,4 +10,16 @@ And the contributing guide: https://github.com/containous/traefik/blob/master/.g
 
 <!--
 Briefly describe the pull request in a few paragraphs.
--->
\ No newline at end of file
+-->
+
+<!--
+ PLEASE, DON'T REMOVE THIS CHECKLIST AND COMPLETE IT.
```
timoreimann: Who is going to use this check list?

If it's going to be the maintainers, I'm generally fine with the structure.

If it's going to be the PR author, then I think a list without check boxes will be good enough. Otherwise, authors will just forget to tick them off, and we'll have the extra burden to remind/ask them or tick things off ourselves, which will just be forgotten as well.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,4 +10,16 @@ And the contributing guide: https://github.com/containous/traefik/blob/master/.g
...
+ -->
+- [x] Open a PR
+- [ ] Add a short title (57 chars max)
+- [ ] Write useful descriptions
+- [ ] Make sure your branch is up to date.
```
timoreimann: IMHO we don't need this for "maintainer mode" since we have a status indicator, and you can't merge without the branch being up-to-date anyway.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,4 +10,16 @@ And the contributing guide: https://github.com/containous/traefik/blob/master/.g
...
+- [ ] Add tests
+- [ ] Add documentation
+- [ ] Make sure the code builds
```
timoreimann: IMHO not needed in "maintainer mode" either for similar reasons as above.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -10,4 +10,16 @@ And the contributing guide: https://github.com/containous/traefik/blob/master/.g
...
+- [ ] Make sure all tests pass
```
timoreimann: IMHO not needed in "maintainer mode" either for similar reasons as above.

.github/PULL_REQUEST_TEMPLATE.md
ldez: the both: maintainers and authors

timoreimann: Thinking more about it, I don't think the check boxes will serve us well:

- If the users ticks them off, it's still upon the maintainers to verify whether all points are addressed properly. Chances are they won't, at which point the regular code review process will kick in to make sure things gets addressed. So they make no sense for the user.
- If we say the check boxes are for us, then there's no guarantee the PR author (or someone else) will accidentally (un-)check them. Github does not keep record of who toggles a box (and when) either.

Consequently, the only way the points make sense to me is to dedicate them to the user and remove the check boxes. Everything else should be covered by the usual review process.

PR #1504, opened 2017-04-27
add metrics for backend_retries_total

marco-jantke:
This PR adds metrics for the backend_retries_total. On the way to implement this, I did a couple of clean up tasks and improved the testing coverage of the existing implementation.

Cleanup tasks:
- I renamed variables/methods for the request duration metrics, from latency to requestDuration. The concept was confused before in the code and in fact the request duration is tracked and the exported prometheus name was `traefik_request_duration_seconds` all the time. Therefore this is no breaking change, just clarification inside the code base.
- I remove the Handler() method from the Metrics interface as this is a concern not relevant for consumers of the interface, which are concerned about the collection of metrics. Also the method was not even used.
- I made the `ResponseRecorder` in the retry middleware private, as this is something you would not expect from the package interface to provide.

Apart of this I added unit tests for the `retry.go` and the parts I touched in the `server.go`. Especially the `server.go` method is hard to test, due to its current design, but I hope it is a first step and having the file will motivate others to increase the coverage there as well.

### DIFF
--- a/middlewares/prometheus.go
+++ b/middlewares/prometheus.go
@@ -1,57 +1,64 @@
 package middlewares
 
 import (
+	"fmt"
+
 	"github.com/containous/traefik/types"
 	"github.com/go-kit/kit/metrics"
 	"github.com/go-kit/kit/metrics/prometheus"
 	stdprometheus "github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
-	"net/http"
 )
 
 const (
-	reqsName    = "traefik_requests_total"
-	latencyName = "traefik_request_duration_seconds"
+	reqsTotalName    = "traefik_requests_total"
+	reqDurationName  = "traefik_request_duration_seconds"
+	retriesTotalName = "traefik_backend_retries_total"
 )
 
-// Prometheus is an Implementation for Metrics that exposes prometheus metrics for the latency
-// and the number of requests partitioned by status code and method.
+// Prometheus is an Implementation for Metrics that exposes the following Prometheus metrics:
+// - number of requests partitioned by status code and method
+// - request durations
+// - amount of retries happened
 type Prometheus struct {
-	reqsCounter      metrics.Counter
-	latencyHistogram metrics.Histogram
+	reqsCounter          metrics.Counter
+	reqDurationHistogram metrics.Histogram
+	retryCounter         metrics.Counter
 }
 
 func (p *Prometheus) getReqsCounter() metrics.Counter {
 	return p.reqsCounter
 }
 
-func (p *Prometheus) getLatencyHistogram() metrics.Histogram {
-	return p.latencyHistogram
+func (p *Prometheus) getReqDurationHistogram() metrics.Histogram {
+	return p.reqDurationHistogram
 }
 
-// NewPrometheus returns a new prometheus Metrics implementation.
-func NewPrometheus(name string, config *types.Prometheus) *Prometheus {
-	var m Prometheus
+func (p *Prometheus) getRetryCounter() metrics.Counter {
+	return p.retryCounter
+}
+
+// NewPrometheus returns a new Prometheus Metrics implementation.
+// With the returned collectors you have the possibility to clean up the internal Prometheus state by unsubscribing the collectors.
+// This is for example useful while testing the Prometheus implementation.
+// If any of the Prometheus Metrics can not be registered an error will be returned and the returned Metrics implementation will be nil.
+func NewPrometheus(name string, config *types.Prometheus) (*Prometheus, []stdprometheus.Collector, error) {
+	var prom Prometheus
+	var collectors []stdprometheus.Collector
 
 	cv := stdprometheus.NewCounterVec(
 		stdprometheus.CounterOpts{
-			Name:        reqsName,
+			Name:        reqsTotalName,
 			Help:        "How many HTTP requests processed, partitioned by status code and method.",
 			ConstLabels: stdprometheus.Labels{"service": name},
 		},
 		[]string{"code", "method"},
 	)
-
-	err := stdprometheus.Register(cv)
+	cv, err := registerCounterVec(cv)
 	if err != nil {
-		e, ok := err.(stdprometheus.AlreadyRegisteredError)
-		if !ok {
-			panic(err)
-		}
-		m.reqsCounter = prometheus.NewCounter(e.ExistingCollector.(*stdprometheus.CounterVec))
-	} else {
-		m.reqsCounter = prometheus.NewCounter(cv)
+		return nil, collectors, err
 	}
+	prom.reqsCounter = prometheus.NewCounter(cv)
+	collectors = append(collectors, cv)
 
 	var buckets []float64
 	if config.Buckets != nil {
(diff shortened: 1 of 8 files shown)


marco-jantke:
I will fix the build tomorrow morning, please don't hesitate to have a look at the PR already. That happens when you only execute `make test-unit` and `make test-integration` locally 🐼

ldez:
Welcome and thanks for your contribution!
Can you fix the "formatting bug" ? https://travis-ci.org/containous/traefik/jobs/226464395#L826

marco-jantke:
Thanks for the welcome and for the introduction @timoreimann. I should probably also have said a couple of words :D I am happy to be able to contribute to the project. Linting errors are fixed by the way.

marco-jantke:
Hi @vdemeester and @emilevauge. Can you have a look at this PR? @timoreimann told me that one of you has to give the approve the PR in order to get it merged :)

marco-jantke:
I just updated PR against latest master again. @vdemeester or @emilevauge can you have a look?

middlewares/prometheus.go
```diff
@@ -62,28 +63,56 @@ func NewPrometheus(name string, config *types.Prometheus) *Prometheus {
...
 
-func (p *Prometheus) handler() http.Handler {
-	return promhttp.Handler()
+func registerHistogramVec(hv *stdprometheus.HistogramVec) *stdprometheus.HistogramVec {
+	err := stdprometheus.Register(hv)
+
+	if err != nil {
+		e, ok := err.(stdprometheus.AlreadyRegisteredError)
+		if !ok {
+			panic(err)
```
emilevauge: Please, don't use panic, return an error instead. You shouldn't stop a reverse proxy because of this.

middlewares/retry.go
```diff
@@ -11,24 +11,27 @@ import (
...
 // Retry is a middleware that retries requests
 type Retry struct {
 	attempts int
 	next     http.Handler
+	metrics  RetryMetrics
 }
 
 // NewRetry returns a new Retry instance
-func NewRetry(attempts int, next http.Handler) *Retry {
+func NewRetry(attempts int, next http.Handler, metrics RetryMetrics) *Retry {
```
emilevauge: I think it would be better to add an interface here, instead of passing `RetryMetrics` directly. WDYT ?
```
type Listener interface {
	Retried(attempt int) error
}
```

server/server.go
```diff
@@ -180,7 +180,8 @@ func (server *Server) startHTTPServers() {
 		serverMiddlewares := []negroni.Handler{server.accessLoggerMiddleware, server.loggerMiddleware, metrics}
 		if server.globalConfiguration.Web != nil && server.globalConfiguration.Web.Metrics != nil {
 			if server.globalConfiguration.Web.Metrics.Prometheus != nil {
-				metricsMiddleware := middlewares.NewMetricsWrapper(middlewares.NewPrometheus(newServerEntryPointName, server.globalConfiguration.Web.Metrics.Prometheus))
+				promMetrics, _ := middlewares.NewPrometheus(newServerEntryPointName, server.globalConfiguration.Web.Metrics.Prometheus)
```
emilevauge: Shouldn't you use `newMetrics` here too ?

middlewares/prometheus.go
marco-jantke: Ok, I rebuilt it now in a way that if any of the Prometheus Metrics can not be registered, an error will be returned and the `*Prometheus` will be `nil`. This will lead to no Metrics instrumentation and an log message on Error level. Does it make sense to you this way?

server/server.go
marco-jantke: Good catch, fixed it.

middlewares/retry.go
marco-jantke: Usually I refrain from building abstractions before they are actually required. Therefore personally I would rather leave this part as is, extending it with a more generic implementation once we have an use-case for this. Agree?

marco-jantke:
Closing/Reopening to get SemaphoreCI working.

middlewares/retry.go
marco-jantke: @emilevauge WDYT? :)

emilevauge: The `retry` middleware should be loosely coupled to the `metrics` middleware. If I agree with you on not building abstractions too early, here we rather need isolation.

server/server.go
```diff
@@ -893,3 +886,37 @@ func sortedFrontendNamesForConfig(configuration *types.Configuration) []string {
 	sort.Strings(keys)
 	return keys
 }
+
+// newMetrics instantiates the proper Metrics implementation, depending on the global configuration.
+// Note that given there is no metrics instrumentation configured, it will return nil.
+func newMetrics(globalConfig GlobalConfiguration, name string) middlewares.Metrics {
+	metricsEnabled := globalConfig.Web != nil && globalConfig.Web.Metrics != nil
```
emilevauge: Any reason to not add `globalConfig.Web.Metrics.Prometheus != nil` here ?

middlewares/retry.go
marco-jantke: Ok, I see your point and I am fine with implementing the abstraction layer. Though I would like to make sure that we are on the same page with our understanding of the code. 

The Retry middleware is not coupled to the metrics middleware. It only knows about the `RetryMetrics` interface, which happens to be right now in the middleware package (IMHO this is the rather confusing part). The `Metrics` and `RetryMetrics` interfaces are in my opinion already a layer of abstraction. The metrics middleware in turn also "only" depends on the `Metrics` interface.

Though the abstraction you suggest would remove all knowledge about metrics from the retry middleware. Do I understand your intention correctly? If so and you suggest that way, I will adapt the PR accordingly.

emilevauge: >Though the abstraction you suggest would remove all knowledge about metrics from the retry middleware. Do I understand your intention correctly?

Exactly :)

marco-jantke: Ok, thanks for the feedback. I implemented it now. See 7ff5692146cc1e0f8a804a1ef54e59cfeb4e4fb1

marco-jantke:
@emilevauge can you have a look again? :)

ldez:
@marco-jantke could you resolve the conflicts?

marco-jantke:
Ok, everything should be ready now.

marco-jantke:
@emilevauge ping

server/server.go
```diff
@@ -955,3 +947,37 @@ func (*Server) configureBackends(backends map[string]*types.Backend) {
...
+			log.Errorf("Error creating Prometheus Metrics implementation: %s", err)
+			return nil
+		}
+		return metrics
+	}
+
+	return nil
+}
+
+func registerRetryMiddleware(
```
emilevauge: @marco-jantke could you remove this new function? It is called only one time, has too few lines and is then hard to read.

marco-jantke: I basically introduced to enable testing, that the middleware is constructed properly. I think this would not be possible anymore, when I inline the logic back into the `loadConfig`. So I can remove it of course if you prefer it this way, but I guess then the test also has to die.

emilevauge: Indeed, I was not thinking about this ;) It's all good then!

marco-jantke:
Thanks for the reviews and merging 🎉

PR #1517, opened 2017-04-30
Add unit tests for package safe

gottwald:
This PR adds tests to package safe.

In some cases the package is a bit awkward to test.
I also noticed something strange in the code here: https://github.com/containous/traefik/blob/e2fdc27d642c9436c9b925e956290e7df95c43bf/safe/routine.go#L110-L117

`Pool.Start()` starts the `Pool.routines`, but the only way to add a routine (not `routineCtx`) from outside of the package is by calling `Pool.Go(goroutine)`, which already starts the goroutine.

The code was added with this commit when routineCtx got introduced: https://github.com/containous/traefik/commit/bea5ad3f132bae27b6c1a83adf00154058b484b5#diff-f7272237e30bd2c1dc9ccb7f7b2ce583R89
Maybe an issue should be opened to find out if `Pool.routines` started by `Pool.Start()` are used at all.

Due to the tests I also found a bug in the `Pool.Start()` implementation, where a copy of a struct was modified instead of the struct itself.

### DIFF
--- a/safe/routine.go
+++ b/safe/routine.go
@@ -3,10 +3,11 @@ package safe
 import (
 	"context"
 	"fmt"
-	"github.com/cenk/backoff"
-	"github.com/containous/traefik/log"
 	"runtime/debug"
 	"sync"
+
+	"github.com/cenk/backoff"
+	"github.com/containous/traefik/log"
 )
 
 type routine struct {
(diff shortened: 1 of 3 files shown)


ldez:
Awesome job! The tests are gold ❤️

emilevauge:
@gottwald 
>I also noticed something strange in the code here: https://github.com/containous/traefik/blob/e2fdc27d642c9436c9b925e956290e7df95c43bf/safe/routine.go#L110-L117

>Pool.Start() starts the Pool.routines, but the only way to add a routine (not routineCtx) from outside of the package is by calling Pool.Go(goroutine), which already starts the goroutine.

The reason is to relaunch all the `routines` after a call to `Stop`. This is used in cluster context where we have to launch some specific go routines in a master instance.

Issue #1510, opened 2017-04-28
Unable to connect to DC/OS Marathon provider


willgorman:
### What version of Traefik are you using?
```
Version:      v1.2.3
Codename:     morbier
Go version:   go1.7.5
Built:        2017-04-13_07:32:59PM
OS/Arch:      darwin/amd64
```

### What is your environment & configuration (arguments, toml...)?
```json
{
 "GraceTimeOut": 10,
 "Debug": false,
 "CheckNewVersion": true,
 "AccessLogsFile": "",
 "TraefikLogsFile": "",
 "LogLevel": "DEBUG",
 "EntryPoints": {},
 "Cluster": null,
 "Constraints": [],
 "ACME": null,
 "DefaultEntryPoints": [],
 "ProvidersThrottleDuration": 2000000000,
 "MaxIdleConnsPerHost": 200,
 "InsecureSkipVerify": false,
 "Retry": null,
 "Docker": null,
 "File": null,
 "Web": null,
 "Marathon": {
  "Watch": true,
  "Filename": "",
  "Constraints": [],
  "Endpoint": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
  "Domain": "marathon.mesos",
  "ExposedByDefault": true,
  "GroupsAsSubDomains": false,
  "DCOSToken": "***REDACTED***",
  "MarathonLBCompatibility": true,
  "TLS": null,
  "DialerTimeout": 60,
  "KeepAlive": 10,
  "Basic": null
 },
 "Consul": null,
 "ConsulCatalog": null,
 "Etcd": null,
 "Zookeeper": null,
 "Boltdb": null,
 "Kubernetes": null,
 "Mesos": null,
 "Eureka": null,
 "ECS": null,
 "Rancher": null,
 "ConfigFile": "/Users/wg9059/Downloads/traefik.toml"
}
```

### What did you do?

Attempted to start traefik with a connection to marathon (DC/OS).   The version of the DC/OS cluster is 1.9 (enterprise)

### What did you expect to see?

The process start with no errors.

### What did you see instead?

```
ERRO[2017-04-28T16:03:22-05:00] Failed to register for events, 404: <html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>openresty/1.9.15.1</center>
</body>
</html>
```
[...]

PR #1278, opened 2017-03-13
Update Oxy, fix for #1199

akanto:

PR #1519, opened 2017-05-01
[Marathon] Update go-marathon

jangie:
### Description

Update `go-marathon` dependency. Addresses https://github.com/gambol99/go-marathon/issues/274 , which is problematic for Traefik users who use Traefik against the Marathon provider, where the Marathon provider is protected by basic auth. At current with that setup, if a given Marathon instance goes out of service, Traefik will not be able to reconnect to that Marathon instance and will essentially have a no-longer-updated frontend/backend configuration. (Traefik will continue to service known routes successfully, but new deployments will not be tracked.)

This dependency update aims to resolve that situation.

Related to #1510

### DIFF
--- a/script/.validate
+++ b/script/.validate
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if [ -z "${VALIDATE_UPSTREAM:-}" ]; then
 	# this is kind of an expensive check, so let's not do this twice if we
--- a/script/binary
+++ b/script/binary
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 if ! test -e autogen/gen.go; then
--- a/script/crossbinary
+++ b/script/crossbinary
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 if ! test -e autogen/gen.go; then
--- a/script/deploy-docker.sh
+++ b/script/deploy-docker.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
 if [ -n "$TRAVIS_COMMIT" ]; then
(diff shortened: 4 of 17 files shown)


timoreimann:
_Might_ be flakiness.

Retriggered the build.

timoreimann:
@jangie updating the integration deps using the proper way might be challenging for several reasons. See [this comment](https://github.com/containous/traefik/pull/1278#issuecomment-286696714) on how to cheat around things.

timoreimann:
@emilevauge @jangie I think this seems important enough to include in 1.3. Do you agree?

jangie:
@timoreimann thank you! i've been struggling against this *all day*. i am very saddened at that comment as it feels like we're fighting against the dependency management system just to get things working as they 'should'. i'll be taking a stab at doing it that way. from what i can tell, there is a breaking binary compatibility issue between a couple of the prometheus libraries, which i've attempted to work around by forcing latest on one of them that has the desired API of the other one.

it is my opinion that it would be good to push this into 1.3 once it builds happily, but i'm somewhat biased as this directly affects one of my use cases; i do not have a handle on how many people run marathon with basic auth.

timoreimann:
@jangie the state of dependency management in Go is a sad one, and the Traefik maintainers are constantly fighting their share of this battle. We plan to make things better for 1.4, there are a few routes to be taken (unify the main/integration gilde files; unpinning existing dependencies; switching to a different management tool).

I can assure you that each time I have to pull the glide trick, a part of me (and a kitten) dies a little bit.

jangie:
@timoreimann in case it wasn't clear i did not and do not mean to disparage this project or its maintainers by any means! i just was saddened that this was the state of affairs in the golang world at large, i have not yet had to do much in the means of dependency management and it was a bit of a rude awakening. having now read up a bit on some of the history, and of what is hopefully to come with `dep`, i am heartened greatly!

will be attempting to set up my home machine to take a look at following those steps now. thanks!

jangie:
ack.. made a mistake. i'll be opening up a new PR, apologies in advance for future tagging spam.

PR #1525, opened 2017-05-02
Fix systemd watchdog feature

### LINKED ISSUE #1353: Traefik shuts down when SystemD watchdog run
1.2.1

### What did you do?
Install service file from: https://github.com/containous/traefik/blob/master/contrib/systemd/traefik.service

Running service

### What did you see instead?
Crashing service every 2s, with SystemD starting it again. Adjusting watchdog timings correlate with time between crashes

guilhem:
Commit coreos/go-systemd@0c088e introduce cleaning environment.
First usage of sdnotify (for type=notify) was clearing NOTIFY_SOCKET environment variable.
sdnotify in watchdog was unable to ping back.

Fix #1353

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -244,22 +244,25 @@ func run(traefikConfiguration *server.TraefikConfiguration) {
 	svr := server.NewServer(globalConfiguration)
 	svr.Start()
 	defer svr.Close()
-	sent, err := daemon.SdNotify(true, "READY=1")
+	sent, err := daemon.SdNotify(false, "READY=1")
 	if !sent && err != nil {
 		log.Error("Fail to notify", err)
 	}
-	t, err := daemon.SdWatchdogEnabled(true)
+	t, err := daemon.SdWatchdogEnabled(false)
 	if err != nil {
 		log.Error("Problem with watchdog", err)
 	} else if t != 0 {
 		// Send a ping each half time given
 		t = t / 2
-		go func(interval time.Duration) {
-			tick := time.Tick(interval)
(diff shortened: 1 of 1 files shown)


cmd/traefik/traefik.go
```diff
@@ -244,20 +244,23 @@ func run(traefikConfiguration *server.TraefikConfiguration) {
...
 	}
-	t, err := daemon.SdWatchdogEnabled(true)
+	t, err := daemon.SdWatchdogEnabled(false)
 	if err != nil {
 		log.Error("Problem with watchdog", err)
 	} else if t != 0 {
 		// Send a ping each half time given
 		t = t / 2
+		log.Info("Watchdog activated with timer each ", t)
 		go func(interval time.Duration) {
```
emilevauge: In order to avoid panic, could you use `safe.Go` to launch the go routine here?

cmd/traefik/traefik.go
```diff
@@ -244,20 +244,23 @@ func run(traefikConfiguration *server.TraefikConfiguration) {
...
 			tick := time.Tick(interval)
 			for range tick {
-				daemon.SdNotify(true, "WATCHDOG=1")
+				if ok, _ := daemon.SdNotify(false, "WATCHDOG=1"); ok {
+					log.Debug("Tick watchdog")
```
emilevauge: Shouldn't we log on error instead ?

guilhem: fixed :)

cmd/traefik/traefik.go
guilhem: fixed

PR #1533, opened 2017-05-03
Fix error handling for docker swarm mode

### LINKED ISSUE #1100: Docker watch is terminated after first error
1.1.2

```
[docker]
endpoint = "tcp://...."
domain = "..."
watch = true
swarmmode = true
exposedbydefault = true
```

### What did you do?
Docker swarm is not available for short time

Watch will continue working after error

### What did you see instead?
Watch is terminated after first error
example error:
```
level=error msg="Failed to list services for docker, error Error response from daemon: rpc error: code = 2 desc = raft: no elected cluster leader"
```

tanyadegurechaff:
### Description

Fixes #1100

### DIFF
--- a/provider/docker/docker.go
+++ b/provider/docker/docker.go
@@ -153,15 +153,18 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 			if p.Watch {
 				ctx, cancel := context.WithCancel(ctx)
 				if p.SwarmMode {
+					errChan := make(chan error)
 					// TODO: This need to be change. Linked to Swarm events docker/docker#23827
 					ticker := time.NewTicker(SwarmDefaultWatchTime)
 					pool.Go(func(stop chan bool) {
+						defer close(errChan)
 						for {
 							select {
 							case <-ticker.C:
 								services, err := p.listServices(ctx, dockerClient)
 								if err != nil {
 									log.Errorf("Failed to list services for docker, error %s", err)
+									errChan <- err
 									return
 								}
 								configuration := p.loadDockerConfig(services)
@@ -179,6 +182,10 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 							}
 						}
 					})
+					if err, ok := <-errChan; ok {
+						return err
+					}
+					// channel closed
 
 				} else {
 					pool.Go(func(stop chan bool) {


tanyadegurechaff:
This makes backoff operation return error when an error occurs, and the operation will be retried.

tanyadegurechaff:
Could you explain in detail how a memory leak comes about?

emilevauge:
@tanyadegurechaff @beldpro-ci After reading more in details, indeed, it does not introduce any memory leak in this case. But it is still better to defer close the channel, especially because it's [inside a go routine](https://github.com/containous/traefik/pull/1533/files#diff-fa150917b5a9de47451ce96099822a9dR120). We had many leaks in the past do to not closed channels. Other than that, seems great :)

tanyadegurechaff:
Thank you for your feedbacks. @beldpro-ci @emilevauge

provider/docker/docker.go
```diff
@@ -153,15 +153,18 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 			if p.Watch {
 				ctx, cancel := context.WithCancel(ctx)
 				if p.SwarmMode {
+					errChan := make(chan error)
 					// TODO: This need to be change. Linked to Swarm events docker/docker#23827
 					ticker := time.NewTicker(SwarmDefaultWatchTime)
 					pool.Go(func(stop chan bool) {
+						defer close(errChan)
```
ldez: why you put the defer here?
I think you must move this line under the line 156.

tanyadegurechaff: The receiver at line 185 blocks the execution until the channel is closed or received an error.

emilevauge: I agree with @tanyadegurechaff on this. But it would be better to use this syntax instead:
```go
if err, ok := <-errChan; ok {
  return err
} else {
  // channel closed
  ...
}
```

tanyadegurechaff: I agree, but I got an error from golint:
```
provider/docker/docker.go:187:13: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
```

ldez: just do:
```golang
if err, ok := <-errChan; ok {
  return err
}
 // channel closed
 // ...
```

ldez:

PR #1534, opened 2017-05-03
Revert "Vendor generated file"

ldez:
Reverts containous/traefik#1464

Problems:
- WebUI missing.
- diff don't really work as expected.


timoreimann:
Yeah just that commit seems enough. We should keep the shebang changes.

ldez:
Done!

timoreimann:
A minor thing: Could we add a short description on the PR on why we needed to revert this?

This will make the decision more transparent.

Issue #1266, opened 2017-03-09
Make the CI great again!


emilevauge:
I can remember the good old times when all our tests were passing in less than 10 minutes on Travis.
Now [it's more like](https://github.com/containous/traefik/pull/1209#issuecomment-285062581) that:
> gopkg.in is flaking again. Retriggered the build.
>...
>Grrr, and now the integration tests are flaking. :(
>Retrying HARDER...

ALL THE TIME!

This global issue is here to list all the work we need to do in order to make the CI great again.

- [x] replace fixed timeouts re: docker compose readiness with `utils.Try` #1267 by @timoreimann 
- [x] test Travis alternative like CircleCI or SemaphoreCI
- [x] Improving Parallelism of tests.
- [ ] Reducing the setup time of a testing environment
- [ ] Simplify the process for a contributer to add a parallel test such as integration

dtomcej:
We should also work on:

- Improving Parallelism of tests.
- Reducing the setup time of a testing environment
- Simplify the process for a contributer to add a parallel test such as integration

vdemeester:
> Simplify the process for a contributer to add a parallel test such as integration

It's more like : simplify the process for a contributor to add tests, mainly unit-test (we should have way more unit test, and very good coverage in unit test, than integration tests)

atbore-phx:
Today I have performed a CI platform switch to circle-ci v.2 (cci2), forking the main traefik repo and the library too.
I've also translated the .travis.yml in the new .circleci/config.yml here https://github.com/atbore-phx/traefik.
The job created use the "machine" executor, cciv2 seams slower then travis see the picture and the link attached.
https://circleci.com/gh/atbore-phx/traefik/12
![cciv2_job](https://cloud.githubusercontent.com/assets/11421185/25540802/9ec2524a-2c4c-11e7-8ffd-a10eecc6b918.png)

those features could help to reduce the execution:
-  cache for the docker images
- test parallelization (we need to evaluate the impact on the actual test suite)
but the slower part is the cross compilation.

Next days I'll test SemaphoreCI.

vdemeester:
The main problem is cross compilation. It is taking more than all other steps.. We should run it parallel/separate.

timoreimann:
One idea is to only test a single architecture per operating system on each PR (e.g., Linux, Windows, and Darwin each on amd64) and postpone the other architectures to a less frequently running job. the latter could be the master branch, a daily cron, or a pre-release step.

[...]

PR #1535, opened 2017-05-03
Makefile target to enable parallel jobs

atbore-phx:
With this PR i have divided the original ./script/crossbinary file in 2 files.

- crossbinary-default
It contains windows and linux

- crossbinary-others
it contains darwin, freebsd and openbsd

Makefile has been modified using the same schema adding new targets to enable parallelism at make level.
The main goal was to reduce the make crossbinary exec time and to separate win/linux platforms  with minimum impact on actual test suite.
see https://github.com/containous/traefik/issues/1266

with parallel execution (make -j2 crossbinary-parallel) we reduce the exec time of ~10%

tested with 
make crossbinary
semaphoreci : <https://semaphoreci.com/atbore/traefik/branches/crossbinary-parallel/builds/4>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,10 @@ TRAEFIK_IMAGE := $(if $(REPONAME),$(REPONAME),"containous/traefik")
 INTEGRATION_OPTS := $(if $(MAKE_DOCKER_HOST),-e "DOCKER_HOST=$(MAKE_DOCKER_HOST)", -v "/var/run/docker.sock:/var/run/docker.sock")
 
 DOCKER_BUILD_ARGS := $(if $(DOCKER_VERSION), "--build-arg=DOCKER_VERSION=$(DOCKER_VERSION)",)
-DOCKER_RUN_TRAEFIK := docker run $(INTEGRATION_OPTS) -it $(TRAEFIK_ENVS) $(TRAEFIK_MOUNT) "$(TRAEFIK_DEV_IMAGE)"
+DOCKER_RUN_OPTS := $(TRAEFIK_ENVS) $(TRAEFIK_MOUNT) "$(TRAEFIK_DEV_IMAGE)"
+DOCKER_RUN_TRAEFIK := docker run $(INTEGRATION_OPTS) -it $(DOCKER_RUN_OPTS)
+DOCKER_RUN_TRAEFIK_NOTTY := docker run $(INTEGRATION_OPTS) -i $(DOCKER_RUN_OPTS)
+
 
 print-%: ; @echo $*=$($*)
 
@@ -35,6 +38,17 @@ binary: generate-webui build ## build the linux binary
 
 crossbinary: generate-webui build ## cross build the non-linux binaries
 	$(DOCKER_RUN_TRAEFIK) ./script/make.sh generate crossbinary
+
+crossbinary-parallel:
+	$(MAKE) generate-webui
+	$(MAKE) build
+	$(MAKE) crossbinary-default crossbinary-others
+
+crossbinary-default: generate-webui build
+	$(DOCKER_RUN_TRAEFIK_NOTTY) ./script/make.sh generate crossbinary-default
+
+crossbinary-others: generate-webui build
+	$(DOCKER_RUN_TRAEFIK_NOTTY) ./script/make.sh generate crossbinary-others
 
 test: build ## run the unit and integration tests
 	$(DOCKER_RUN_TRAEFIK) ./script/make.sh generate test-unit binary test-integration
--- a/script/crossbinary-default
+++ b/script/crossbinary-default
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -e
+
+if ! test -e autogen/gen.go; then
+	echo >&2 'error: generate must be run before crossbinary'
+	false
+fi
+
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -22,6 +22,8 @@ INTEGRATION_OPTS := $(if $(MAKE_DOCKER_HOST),-e "DOCKER_HOST=$(MAKE_DOCKER_HOST)
 
 DOCKER_BUILD_ARGS := $(if $(DOCKER_VERSION), "--build-arg=DOCKER_VERSION=$(DOCKER_VERSION)",)
 DOCKER_RUN_TRAEFIK := docker run $(INTEGRATION_OPTS) -it $(TRAEFIK_ENVS) $(TRAEFIK_MOUNT) "$(TRAEFIK_DEV_IMAGE)"
+DOCKER_RUN_TRAEFIK_NOTTY := docker run $(INTEGRATION_OPTS) -i $(TRAEFIK_ENVS) $(TRAEFIK_MOUNT) "$(TRAEFIK_DEV_IMAGE)"
```
timoreimann: Is it possible to extract all run parameters after the `-i` into a variable and use it for both TTY and non-TTY runs? That way, we won't risk changing one but not the other should they parameters change in the future.

Makefile
```diff
@@ -36,6 +38,16 @@ binary: generate-webui build ## build the linux binary
 crossbinary: generate-webui build ## cross build the non-linux binaries
 	$(DOCKER_RUN_TRAEFIK) ./script/make.sh generate crossbinary
 
+crossbinary-p:
+	$(MAKE) generate-webui build
+	$(MAKE) crossbinary-default crossbinary-others
+
+crossbinary-default:
+	$(DOCKER_RUN_TRAEFIK_NOTTY) ./script/make.sh generate crossbinary-default
```
timoreimann: Why do we actually need to run without TTY?

script/crossbinary-default
```diff
@@ -0,0 +1,43 @@
...
+
+if [ -z "$DATE" ]; then
+    DATE=$(date -u '+%Y-%m-%d_%I:%M:%S%p')
+fi
+
+# Get rid of existing binaries
+rm -f dist/traefik_*
+
+# Build 386 amd64 binaries
+OS_PLATFORM_ARG=(linux windows)
```
timoreimann: I'm in favor of including Darwin here: there are a number of Mac developers, and we should be careful not to break the build for them.

In return...

script/crossbinary-default
```diff
@@ -0,0 +1,43 @@
...
+OS_ARCH_ARG=(386 amd64)
```
timoreimann: ...I think it's acceptable to only include the amd64 architecture. IMHO, there's a higher risk to break the build for a particular OS than a particular architecture, especially given Traefik's rather high level nature.

Makefile
```diff
@@ -36,6 +38,16 @@ binary: generate-webui build ## build the linux binary
 crossbinary: generate-webui build ## cross build the non-linux binaries
 	$(DOCKER_RUN_TRAEFIK) ./script/make.sh generate crossbinary
 
+crossbinary-p:
```
timoreimann: What does the `p` stand for?

Makefile
```diff
@@ -36,6 +38,17 @@ binary: generate-webui build ## build the linux binary
 crossbinary: generate-webui build ## cross build the non-linux binaries
 	$(DOCKER_RUN_TRAEFIK) ./script/make.sh generate crossbinary
 
+crossbinary-parallel:
```
timoreimann: Can't we just make this target depend on `binary` and remove the first two make calls?

Makefile
```diff
@@ -36,6 +38,17 @@ binary: generate-webui build ## build the linux binary
...
+	$(MAKE) generate-webui
+	$(MAKE) build
+	$(MAKE) crossbinary-default crossbinary-others
+
+crossbinary-default:
```
timoreimann: Shouldn't this target (and `-others`) depend on `binary` too in case we want to run this stand-alone (i.e., without running `crossbinary-parallel` first)?

timoreimann:
One of my comments seemingly got dropped for some reason, sorry:

Is it possible to merge the two cross compilation scripts into a single one that takes a `default`, `other`, or no parameter (and, in the latter case, builds everything)? Right now, there's a lot of duplication, so we'd always need to remember to touch both if we make changes.

WDYT?

atbore-phx:
@timoreimann, Thx for the review.
All set ! 

About the default and extra params, in this PR i would just generate a minimum impact but yes i'm totally agree and i'm working to this specific subject and i'll submit it in another PR.
 
Thx for your PR i've seen your approach.

atbore-phx:
for the file duplication i know, i'll try to tune it, my issue was the make.sh. 
It shifts bundles and to parallelize make i need separate targets as dependencies (which are two different bundles)
i have tried to retouch make.sh but the impact was very huge so i need to come back to an easy solution as first shoot.
i promise to think about a solution to have an unique file, i would do that with params if possible

timoreimann:
As discussed, we'll merge as-is and deal with de-duplication in a separate PR (per issue 1539 you have created).

PR #1542, opened 2017-05-03
Working UI

maxwo:
In case you're interested, here is the corrected/working version of the UI compared to the current state on master.
Things I did:
- Add babel-loader to make the UglifyJS working with ES2015 source;
- Continue the work begun in the provider service.


webui/src/app/core/providers.resource.js
```diff
@@ -4,44 +4,41 @@ var angular = require('angular');
...
-              goodFrontend.frontendId = key;
-              return goodFrontend;
-            });
-
-        }
-      }
-
-      return rawProviders;
+    get: function() {
+      return $q((resolve, reject) => {
```
ldez: Use [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) instead of `$q`

webui/src/app/core/providers.resource.js
```diff
@@ -4,44 +4,41 @@ var angular = require('angular');
 var traefikCoreProvider = 'traefik.core.provider';
 module.exports = traefikCoreProvider;
 
-angular
-  .module(traefikCoreProvider, ['ngResource'])
-  .factory('Providers', Providers);
+angular.module(traefikCoreProvider, ['ngResource']).factory('Providers', Providers);
```
maxwo: Sure, my fault. I may have auto-format too quickly.

webui/src/app/core/providers.resource.js
maxwo: $q is the preferred way to handle promises with AngularJS (1.X) , since it wraps the promise mechanism into a $digest loop, which allows immediate view refresh. For instance, notice how it takes 2 additional seconds to display data on startup, before it effectively does the $interval call to that refreshes data.

webui/conf/browsersync.conf.js
```diff
@@ -1,11 +1,26 @@
 const conf = require('./gulp.conf');
+const proxy = require('http-proxy-middleware');
```
maxwo: This dependency allows you to run in dev mode with a real Traefik backend running on localhost:8080.
For instance, when you run :
`docker run -p 8080:8080 -p 80:80 -v /var/run/docker.sock:/var/run/docker.sock  traefik --debug --web --docker --docker.watch`
and then
`npm run serve`
On http://localhost:3000/ , the server is now able to connect to the running traefik /api and /health endpoints.
I think it could be useful for testing with a real server, and since it is in dev dependencies, it is not bundled in the final release. Let me know if you want me to remove this.

maxwo:
I just pushed the style correction, plus changed a test to a cleaner one that keeps displaying providers even if there are no backend registered to that provider.

I'd be please to contribute but migrating to an Angular, React or VueJS front-end if I have time and you are up to, with unit tests (sorry, haven't add time to add this in this PR since there wasn't any, and I just wanted to make the UI work again...). Let me know.

webui/conf/browsersync.conf.js
ldez: Please remove this part for now, and create an other PR for this. thanks.

ldez:
If you want migrate to Angular or VueJS, we will be very happy.

webui/conf/browsersync.conf.js
ldez: One PR, one subject 😉

ldez: But I love this feature.

PR #1546, opened 2017-05-04
[CI] Enhance cross-binary builds and parallelism

atbore-phx:
With this PR the make target cross binary is executed before the travis deploy step, so for each PR.
I have also added the parallelization of the same target and a new variable "GO_BUILD_CMD" for the scripts.

### DIFF
--- a/script/crossbinary
+++ b/script/crossbinary
@@ -18,6 +18,10 @@ if [ -z "$DATE" ]; then
     DATE=$(date -u '+%Y-%m-%d_%I:%M:%S%p')
 fi
 
+GIT_REPO_URL='github.com/containous/traefik/version'
+GO_BUILD_CMD="go build -ldflags"
+GO_BUILD_OPT="-s -w -X ${GIT_REPO_URL}.Version=$VERSION -X ${GIT_REPO_URL}.Codename=$CODENAME -X ${GIT_REPO_URL}.BuildDate=$DATE"
+
 # Get rid of existing binaries
 rm -f dist/traefik_*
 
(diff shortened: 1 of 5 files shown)


script/crossbinary
```diff
@@ -27,7 +30,7 @@ OS_ARCH_ARG=(386 amd64)
 for OS in ${OS_PLATFORM_ARG[@]}; do
   for ARCH in ${OS_ARCH_ARG[@]}; do
     echo "Building binary for $OS/$ARCH..."
-    GOARCH=$ARCH GOOS=$OS CGO_ENABLED=0 go build -ldflags "-s -w -X github.com/containous/traefik/version.Version=$VERSION -X github.com/containous/traefik/version.Codename=$CODENAME -X github.com/containous/traefik/version.BuildDate=$DATE" -o "dist/traefik_$OS-$ARCH" ./cmd/traefik/
+    /bin/bash -e -c "GOARCH=$ARCH GOOS=$OS CGO_ENABLED=0 $GO_BUILD_CMD -o \"dist/traefik_$OS-$ARCH\" ./cmd/traefik/"
```
timoreimann: What's the reason we need to create a sub shell here?

Besides, we cannot use `/bin/bash` since it's not portable.

ldez:
Ran for 39 min 13 sec
Total time 1 hr 17 min 57 sec

atbore-phx:
@timoreimann you absolutely right i remove the subshell.
@ldez ok let's split in default (for PR) and others (for deploy)

atbore-phx:
@ldez ~26min right now

PR #1549, opened 2017-05-04
retry only on real network errors

marco-jantke:
Now retries only happen when actual network errors occur and not only
anymore based on the HTTP status code. This is because the backend could
also send this status codes as their normal interface and in that case
we don't want to retry.

This PR should solve the issue [1104](https://github.com/containous/traefik/issues/1104)

### DIFF
--- a/middlewares/retry.go
+++ b/middlewares/retry.go
@@ -3,6 +3,7 @@ package middlewares
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"io/ioutil"
 	"net"
 	"net/http"
@@ -42,10 +43,16 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	}
 	attempts := 1
 	for {
+		netErrorOccurred := false
+		// We pass in a pointer to netErrorOccurred so that we can set it to true on network errors
+		// when proxying the HTTP requests to the backends. This happens in the custom RecordingErrorHandler.
+		newCtx := context.WithValue(r.Context(), defaultNetErrCtxKey, &netErrorOccurred)
+
 		recorder := newRetryResponseRecorder()
 		recorder.responseWriter = rw
-		retry.next.ServeHTTP(recorder, r)
-		if !isNetworkError(recorder.Code) || attempts >= retry.attempts {
+
+		retry.next.ServeHTTP(recorder, r.WithContext(newCtx))
+		if !netErrorOccurred || attempts >= retry.attempts {
 			utils.CopyHeaders(rw.Header(), recorder.Header())
 			rw.WriteHeader(recorder.Code)
 			rw.Write(recorder.Body.Bytes())
@@ -57,8 +64,29 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func isNetworkError(status int) bool {
-	return status == http.StatusBadGateway || status == http.StatusGatewayTimeout
+// netErrorCtxKey is a custom type that is used as key for the context.
+type netErrorCtxKey string
+
+// defaultNetErrCtxKey is the actual key which value is used to record network errors.
+var defaultNetErrCtxKey netErrorCtxKey = "NetErrCtxKey"
+
+// NetErrorRecorder is an interface to record net errors.
+type NetErrorRecorder interface {
+	// Record can be used to signal the retry middleware that an network error happened
+	// and therefore the request should be retried.
+	Record(ctx context.Context)
+}
+
+// DefaultNetErrorRecorder is the default NetErrorRecorder implementation.
+type DefaultNetErrorRecorder struct{}
+
+// Record is recording network errors by setting the context value for the defaultNetErrCtxKey to true.
+func (DefaultNetErrorRecorder) Record(ctx context.Context) {
+	val := ctx.Value(defaultNetErrCtxKey)
+
+	if netErrorOccurred, isBoolPointer := val.(*bool); isBoolPointer {
+		*netErrorOccurred = true
+	}
 }
 
 // RetryListener is used to inform about retry attempts.
--- a/middlewares/retry_test.go
+++ b/middlewares/retry_test.go
@@ -1,6 +1,7 @@
 package middlewares
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -41,7 +42,7 @@ func TestRetry(t *testing.T) {
 			t.Parallel()
 
 			var httpHandler http.Handler
-			httpHandler = &networkFailingHTTPHandler{failAtCalls: tc.failAtCalls}
+			httpHandler = &networkFailingHTTPHandler{failAtCalls: tc.failAtCalls, netErrorRecorder: &DefaultNetErrorRecorder{}}
 			httpHandler = NewRetry(tc.attempts, httpHandler, tc.listener)
 
 			recorder := httptest.NewRecorder()
@@ -62,10 +63,38 @@ func TestRetry(t *testing.T) {
 	}
 }
 
+func TestDefaultNetErrorRecorderSuccess(t *testing.T) {
+	boolNetErrorOccurred := false
+	recorder := DefaultNetErrorRecorder{}
+	recorder.Record(context.WithValue(context.Background(), defaultNetErrCtxKey, &boolNetErrorOccurred))
+	if !boolNetErrorOccurred {
+		t.Errorf("got %v after recording net error, wanted %v", boolNetErrorOccurred, true)
+	}
+}
+
+func TestDefaultNetErrorRecorderInvalidValueType(t *testing.T) {
+	stringNetErrorOccured := "nonsense"
+	recorder := DefaultNetErrorRecorder{}
+	recorder.Record(context.WithValue(context.Background(), defaultNetErrCtxKey, &stringNetErrorOccured))
+	if stringNetErrorOccured != "nonsense" {
+		t.Errorf("got %v after recording net error, wanted %v", stringNetErrorOccured, "nonsense")
+	}
+}
+
+func TestDefaultNetErrorRecorderNilValue(t *testing.T) {
+	nilNetErrorOccured := interface{}(nil)
+	recorder := DefaultNetErrorRecorder{}
+	recorder.Record(context.WithValue(context.Background(), defaultNetErrCtxKey, &nilNetErrorOccured))
+	if nilNetErrorOccured != interface{}(nil) {
+		t.Errorf("got %v after recording net error, wanted %v", nilNetErrorOccured, interface{}(nil))
+	}
+}
+
 // networkFailingHTTPHandler is an http.Handler implementation you can use to test retries.
 type networkFailingHTTPHandler struct {
-	failAtCalls []int
-	callNumber  int
+	netErrorRecorder NetErrorRecorder
+	failAtCalls      []int
+	callNumber       int
 }
 
 func (handler *networkFailingHTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
@@ -73,6 +102,8 @@ func (handler *networkFailingHTTPHandler) ServeHTTP(w http.ResponseWriter, r *ht
 
 	for _, failAtCall := range handler.failAtCalls {
 		if handler.callNumber == failAtCall {
+			handler.netErrorRecorder.Record(r.Context())
+
 			w.WriteHeader(http.StatusBadGateway)
 			return
 		}
--- a/server/errorhandler.go
+++ b/server/errorhandler.go
@@ -1,0 +1,40 @@
+package server
+
+import (
+	"io"
+	"net"
+	"net/http"
+
+	"github.com/containous/traefik/middlewares"
+)
+
+// RecordingErrorHandler is an error handler, implementing the vulcand/oxy
+// error handler interface, which is recording network errors by using the netErrorRecorder.
+// In addition it sets a proper HTTP status code and body, depending on the type of error occurred.
+type RecordingErrorHandler struct {
+	netErrorRecorder middlewares.NetErrorRecorder
+}
+
+// NewRecordingErrorHandler creates and returns a new instance of RecordingErrorHandler.
+func NewRecordingErrorHandler(recorder middlewares.NetErrorRecorder) *RecordingErrorHandler {
+	return &RecordingErrorHandler{recorder}
+}
+
+func (eh *RecordingErrorHandler) ServeHTTP(w http.ResponseWriter, req *http.Request, err error) {
+	statusCode := http.StatusInternalServerError
+
+	if e, ok := err.(net.Error); ok {
+		eh.netErrorRecorder.Record(req.Context())
+		if e.Timeout() {
+			statusCode = http.StatusGatewayTimeout
+		} else {
+			statusCode = http.StatusBadGateway
+		}
+	} else if err == io.EOF {
+		eh.netErrorRecorder.Record(req.Context())
+		statusCode = http.StatusBadGateway
+	}
+
+	w.WriteHeader(statusCode)
+	w.Write([]byte(http.StatusText(statusCode)))
+}
--- a/server/errorhandler_test.go
+++ b/server/errorhandler_test.go
@@ -1,0 +1,88 @@
+package server
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+type timeoutError struct{}
+
+func (e *timeoutError) Error() string   { return "i/o timeout" }
+func (e *timeoutError) Timeout() bool   { return true }
+func (e *timeoutError) Temporary() bool { return true }
+
+func TestServeHTTP(t *testing.T) {
+	tests := []struct {
+		name               string
+		err                error
+		wantHTTPStatus     int
+		wantNetErrRecorded bool
+	}{
+		{
+			name:               "net.Error",
+			err:                net.UnknownNetworkError("any network error"),
+			wantHTTPStatus:     http.StatusBadGateway,
+			wantNetErrRecorded: true,
+		},
+		{
+			name:               "net.Error with Timeout",
+			err:                &timeoutError{},
+			wantHTTPStatus:     http.StatusGatewayTimeout,
+			wantNetErrRecorded: true,
+		},
+		{
+			name:               "io.EOF",
+			err:                io.EOF,
+			wantHTTPStatus:     http.StatusBadGateway,
+			wantNetErrRecorded: true,
+		},
+		{
+			name:               "custom error",
+			err:                errors.New("any error"),
+			wantHTTPStatus:     http.StatusInternalServerError,
+			wantNetErrRecorded: false,
+		},
+		{
+			name:               "nil error",
+			err:                nil,
+			wantHTTPStatus:     http.StatusInternalServerError,
+			wantNetErrRecorded: false,
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			recorder := httptest.NewRecorder()
+
+			errorRecorder := &netErrorRecorder{}
+			req := httptest.NewRequest(http.MethodGet, "http://localhost:3000/any", nil)
+
+			recordingErrorHandler := NewRecordingErrorHandler(errorRecorder)
+			recordingErrorHandler.ServeHTTP(recorder, req, test.err)
+
+			if recorder.Code != test.wantHTTPStatus {
+				t.Errorf("got HTTP status code %v, wanted %v", recorder.Code, test.wantHTTPStatus)
+			}
+			if errorRecorder.netErrorWasRecorded != test.wantNetErrRecorded {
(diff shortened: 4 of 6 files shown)


middlewares/retry_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+		t.Run(tcName, func(t *testing.T) {
+			t.Parallel()
+
+			var httpHandler http.Handler
+			httpHandler = &networkFailingHTTPHandler{failAtCalls: tc.failAtCalls}
+			httpHandler = NewRetry(tc.retryCount, httpHandler)
+
+			recorder := httptest.NewRecorder()
+			req, err := http.NewRequest("GET", "http://localhost:3000/ok", ioutil.NopCloser(nil))
```
timoreimann: you could `s/"GET"/http.MethodGet/` here.

In some other parts of the PR too.

middlewares/retry_test.go
```diff
@@ -0,0 +1,79 @@
...
+			if err != nil {
+				panic(fmt.Sprintf("NewRequest(): %s", err))
+			}
+
+			httpHandler.ServeHTTP(recorder, req)
+
+			if tc.responseStatus != recorder.Code {
+				t.Errorf("wrong status code, expected %v: got %v", tc.responseStatus, recorder.Code)
```
timoreimann: In Go: _got_ first, _expected_/_actual_ second. E.g.,

_"got status code %v, expected %v"_

In other parts of this PR too.

middlewares/retry_test.go
```diff
@@ -0,0 +1,79 @@
...
+			}
+		})
+	}
+}
+
+// An http.Handler implementation you can use to test retries.
```
timoreimann: Should start with the type you are describing, i.e.:

_networkFailingHTTPHandler is an http.Handler implementation you can use to test retries._

server/errorhandler.go
```diff
@@ -0,0 +1,47 @@
...
+	"net/http"
+
+	"github.com/containous/traefik/middlewares"
+)
+
+// RecordingErrorHandler is an error handler, implementing the vulcand/oxy
+// error handler interface, which is recording network errors by setting the
+// request context value with the key middlewares.DefaultNetErrCtxKey to true.
+// This enables the retry middleware to only retry on actual network errors.
+type RecordingErrorHandler struct{}
```
timoreimann: Since we only care about implementing `ServeHTTP(http.ResponseWriter, *http.Request, error)`, couldn't `RecordingErrorHandler` simply be defined as a function with that signature instead of an empty struct?

server/errorhandler.go
```diff
@@ -0,0 +1,47 @@
...
+
+// NewRecordingErrorHandler creates and returns a new instance of RecordingErrorHandler.
+func NewRecordingErrorHandler() *RecordingErrorHandler {
```
timoreimann: As the implementation of `NewRecordingErrorHandler` isn't doing anything besides returning a pointer to an empty struct, couldn't we just drop it and have callers do it instead?

server/errorhandler.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	w.WriteHeader(statusCode)
+	w.Write([]byte(http.StatusText(statusCode)))
+}
+
+func recordNetError(req *http.Request) {
+	val := req.Context().Value(middlewares.DefaultNetErrCtxKey)
+
+	if netErrorOccured, ok := val.(*bool); ok {
```
timoreimann: Does this explode if `val` is `nil` (i.e., the value does not exist), or will Go handle this fine with `ok` being `false`?

middlewares/retry_test.go
```diff
@@ -0,0 +1,79 @@
...
+	failAtCalls []int
+	callNumber  int
+}
+
+func (handler *networkFailingHTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	handler.callNumber++
+
+	for _, failAtCall := range handler.failAtCalls {
+		if handler.callNumber == failAtCall {
+			val := r.Context().Value(DefaultNetErrCtxKey)
```
timoreimann: How about exporting `server.recordNetError` and calling it here?

server/errorhandler_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			recorder := httptest.NewRecorder()
+
+			netErrorOccured := false
+			req, err := http.NewRequest("GET", "http://localhost:3000/any", nil)
+			if err != nil {
+				panic(fmt.Sprintf("NewRequest(): %s", err))
```
timoreimann: A `MustNewRequest` helper function might become convenient at this point.

server/errorhandler.go
```diff
@@ -0,0 +1,47 @@
...
+	"net"
+	"net/http"
+
+	"github.com/containous/traefik/middlewares"
+)
+
+// RecordingErrorHandler is an error handler, implementing the vulcand/oxy
+// error handler interface, which is recording network errors by setting the
+// request context value with the key middlewares.DefaultNetErrCtxKey to true.
+// This enables the retry middleware to only retry on actual network errors.
```
timoreimann: IMHO we should also document the type of errors it maps to 502 and 504, respectively.

server/errorhandler_test.go
```diff
@@ -0,0 +1,81 @@
...
+			err:                io.EOF,
+			wantHTTPStatus:     http.StatusBadGateway,
+			wantNetErrRecorded: true,
+		},
+		{
+			name:               "custom error",
+			err:                errors.New("any error"),
+			wantHTTPStatus:     http.StatusInternalServerError,
+			wantNetErrRecorded: false,
+		},
```
timoreimann: Are we missing a test case for `err == nil`?

server/errorhandler.go
marco-jantke: `go` is handling this fine. The `ctx.Value()` guarantees a return value of type `interface{}`, so we can do a type assertion without danger.

middlewares/retry.go
```diff
@@ -56,9 +63,12 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func isNetworkError(status int) bool {
-	return status == http.StatusBadGateway || status == http.StatusGatewayTimeout
-}
+// NetErrorCtxKey is a custom type that is used as key for the request context.
+// Using an own type we avoid potential collisions with other keys that may have the same string name.
```
marco-jantke: Actually its even a best practice to not export the type is not necessary, to have real safety. Though I decided to have the error handler in the server package still, as for me it makes more sense there, which requires to export the type.

server/errorhandler_test.go
marco-jantke: There is a util function in `httptest.NewRequest` that will internally panic if the creation of the `NewRequest` fails. I think this would be good to use for the tests.

server/errorhandler_test.go
marco-jantke: Yes, better to also add a test for this. Doing.

middlewares/retry.go
```diff
@@ -56,9 +63,12 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+type NetErrorCtxKey string
```
gottwald: From my learnings about context I remember that the key should be a custom type as you do here, but it should also be non exported, since it's something that should be package internal.
Obviously this is not straightforward here because the server package needs to signal something to the middleware package.
But instead of exporting low level package stuff, should the retry middlewares package maybe export some functions to deal with this and keep the context key type private?
Looks like you already have some kind of function in the server package called: `recordNetError`.
This should probably be moved to the middlewares package.
Also `recordNetError` doesn't need to get passed a request, a context would be sufficient.

timoreimann: An Internal package _might_ be helpful here too.

middlewares/retry.go
```diff
@@ -42,10 +43,16 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	}
 	attempts := 1
 	for {
+		netErrorOccured := false
```
marco-jantke: Thanks, fixed.

middlewares/retry.go
```diff
@@ -42,10 +43,16 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+		// We pass in a pointer to netErrorOccured so that we can set it to true on network errors
+		// that happen when proxying the http requests to the backends.
```
marco-jantke: Ok, I extended the comment.

middlewares/retry.go
```diff
@@ -42,10 +43,16 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+		newCtx := context.WithValue(r.Context(), DefaultNetErrCtxKey, &netErrorOccured)
+
 		recorder := NewRecorder()
 		recorder.responseWriter = rw
-		retry.next.ServeHTTP(recorder, r)
-		if !isNetworkError(recorder.Code) || attempts >= retry.attempts {
+
+		retry.next.ServeHTTP(recorder, r.WithContext(newCtx))
+		if netErrorOccured == false || attempts >= retry.attempts {
```
marco-jantke: Ok, done.

middlewares/retry.go
marco-jantke: I refactored the implementation now, so that the knowledge about how to interact with the context completely to the retry middleware and therefore even can use a private type.

middlewares/retry_test.go
marco-jantke: Thanks, fixed at all places.

middlewares/retry_test.go
marco-jantke: Yes.. Fixed it. It seems to be programmed into my fingertips.

middlewares/retry_test.go
marco-jantke: Ok, done.

middlewares/retry_test.go
marco-jantke: As said before, I refactored this implementation, so the implementation for recording the network error is now in the middleware package and the tests make more sense.

server/errorhandler.go
marco-jantke: I added a general comment that it will map to the proper status codes. I fear that being to explicit in the comment will bring up the danger, that the implementation and the comment will drift apart at some time. And when people read the comment, they probably also can read the code.

server/errorhandler.go
marco-jantke: After my refactoring it has now a property for the NetErrorRecorder, so its not possible as function anymore.

server/errorhandler.go
marco-jantke: After my refactoring it has a property now and I would leave the `New..` function.

middlewares/retry.go
marco-jantke: Thanks for the hint, I refactored it now in this way. Now the logic for mutating the context is solely in the retry middleware and therefore we can use a private type 🎉

marco-jantke:
@timoreimann and @gottwald thanks for the review, I worked your feedback into the PR.

middlewares/retry.go
marco-jantke: @timoreimann I am not sure how this should look like to be honest. Also I am hesitant to change the middleware structure in this PR, as imho it would be separate concern. Anyway I think for a long term it could be beneficial to bring more structure into the middleware package.

middlewares/retry_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+			if tc.responseStatus != recorder.Code {
+				t.Errorf("got status code %v, wanted %v", recorder.Code, tc.responseStatus)
+			}
+		})
+	}
+}
+
+func TestRecordNetError(t *testing.T) {
+	boolNetErrorOccurred := false
```
marco-jantke: A small hint for reviewers, I tried to build this as a table driven test at first, but I was not able to get it working in a straightforward way, as it would require me to specify _one_ type for the initial context value I am passing in. This would have to be an `interface{}` in my case, but then the type assertion on a `*bool` in the `DefaultRecordNetError` does not work anymore.

middlewares/retry.go
```diff
@@ -56,8 +63,34 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
-	return status == http.StatusBadGateway || status == http.StatusGatewayTimeout
+// netErrorCtxKey is a custom type that is used as key for the context.
+type netErrorCtxKey string
+
+// defaultNetErrCtxKey is the actual key which value is used to record network errors.
+var defaultNetErrCtxKey netErrorCtxKey = "NetErrCtxKey"
+
+// NetErrorRecorder is an interface to record net errors.
+type NetErrorRecorder interface {
+	// RecordNetError can be used to signal the retry middleware, that an network error happened
```
timoreimann: No comma before `that`.

middlewares/retry_test.go
timoreimann: Then let's have three separate test functions so that we can easily see which test failed, can easily rerun them, etc.

middlewares/retry.go
```diff
@@ -42,10 +43,16 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	}
 	attempts := 1
 	for {
+		netErrorOccurred := false
+		// We pass in a pointer to netErrorOccurred so that we can set it to true on network errors
+		// when proxying the http requests to the backends. This happens in the custom RecordingErrorHandler.
```
timoreimann: s/http/HTTP/

server/errorhandler.go
```diff
@@ -0,0 +1,40 @@
...
+	"io"
+	"net"
+	"net/http"
+
+	"github.com/containous/traefik/middlewares"
+)
+
+// RecordingErrorHandler is an error handler, implementing the vulcand/oxy
+// error handler interface, which is recording network errors by using the netErrorRecorder.
+// In addition it sets a proper http status code and body, depending on the type of error occurred.
```
timoreimann: s/http/HTTP/

middlewares/retry.go
marco-jantke: Thanks, done.

middlewares/retry.go
marco-jantke: Thanks, done.

server/errorhandler.go
marco-jantke: Thanks, done.

middlewares/retry_test.go
marco-jantke: Ok, done.

marco-jantke:
Ok, worked in feedback again.

server/errorhandler.go
```diff
@@ -0,0 +1,40 @@
...
+// NewRecordingErrorHandler creates and returns a new instance of RecordingErrorHandler.
+func NewRecordingErrorHandler(recorder middlewares.NetErrorRecorder) *RecordingErrorHandler {
+	return &RecordingErrorHandler{recorder}
+}
+
+func (eh *RecordingErrorHandler) ServeHTTP(w http.ResponseWriter, req *http.Request, err error) {
+	statusCode := http.StatusInternalServerError
+
+	if e, ok := err.(net.Error); ok {
+		eh.netErrorRecorder.RecordNetError(req.Context())
```
gottwald: nit-pick: This stutters a bit, maybe call the method `Record` instead?

marco-jantke: Thanks for pointing this out, changed it.

marco-jantke:
@emilevauge or @vdemeester can you have a look?

ldez:

middlewares/retry.go
```diff
@@ -56,8 +63,34 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+
+// NetErrorRecorder is an interface to record net errors.
+type NetErrorRecorder interface {
+	// Record can be used to signal the retry middleware that an network error happened
+	// and therefore the request should be retried.
+	Record(ctx context.Context)
+}
+
+// NetErrorRecorderFunc is an util to be able to implement the NetErrorRecorder without the need of a struct.
+type NetErrorRecorderFunc func(ctx context.Context)
```
emilevauge: What is the need of doing this ? Why don't you want to use a `struct` here ?

middlewares/retry.go
```diff
@@ -56,8 +63,34 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
...
+// Record is there to satisfy the NetErrorRecorder interface on a NetErrorRecorderFunc.
+func (f NetErrorRecorderFunc) Record(ctx context.Context) {
+	f(ctx)
+}
+
+// DefaultRecordNetError is recording network errors by setting the context value for the defaultNetErrCtxKey to true.
+func DefaultRecordNetError(ctx context.Context) {
+	val := ctx.Value(defaultNetErrCtxKey)
+
+	if netErrorOccurred, ok := val.(*bool); ok {
```
emilevauge: Not easy to read.

server/errorhandler.go
```diff
@@ -0,0 +1,40 @@
+package server
+
+import (
+	"io"
+	"net"
+	"net/http"
+
+	"github.com/containous/traefik/middlewares"
+)
+
```
emilevauge: You will probably get issues with websockets here.
You should add 
```
var (
	_ Stateful = &RecordingErrorHandler{}
)
```

middlewares/retry.go
marco-jantke: The implementation has no need of any field. I originally had a `struct` here without fields, but got a hint in a previous review here that a function would suffice. Therefore I added this type in order to able to implement the `NetErrorRecorder` interface on a function.

marco-jantke: This said I am fine with both implementations. An empty `struct` is a bit less code, but in some way unnecessary. As you prefer :)

server/errorhandler.go
marco-jantke: Hmm it has none of the capabilities the `Stateful` interface expects. Maybe the naming is rather not perfect then? It is not an extension of `ResponseWriter` or whatsoever, its an implementation of the oxy `ErrorHandler` interface. Its called `Recording` because it calls the `NetErrorRecorder` which in our case will manipulate the request context. So I think this is unnecessary here. Any suggestion for naming improvements?

middlewares/retry.go
marco-jantke: I renamed `ok` now to `isBoolPointer`. Not sure how to improve readability in another way.

marco-jantke:
@emilevauge please have a look again, incorporated/answered your questions.

middlewares/retry.go
emilevauge: If that's OK for you, I would prefer an empty struct, which is easier to read, and with the same result :)

server/errorhandler.go
```diff
@@ -0,0 +1,40 @@
...
+	statusCode := http.StatusInternalServerError
+
+	if e, ok := err.(net.Error); ok {
+		eh.netErrorRecorder.Record(req.Context())
+		if e.Timeout() {
+			statusCode = http.StatusGatewayTimeout
+		} else {
+			statusCode = http.StatusBadGateway
+		}
+	} else if err == io.EOF {
```
emilevauge: Reading this part again, not sure this is the right way to go. We could get an EOF error *not* due to network right ?

server/errorhandler.go
emilevauge: You're right, successfully tested your code with websocket :+1:

middlewares/retry.go
marco-jantke: No problem, I changed it to a struct now.

server/errorhandler.go
marco-jantke: About this one I am not 100% sure whether its right to retry here. I am currently trying to investigate use-cases when an io.EOF can happen. For sure it can be due to a network connection problem, but also when the server closes the connection by intend. Maybe the server would do so because of too much load and we should rather try another one? Anyway I think there is no way to vary in between those cases and thats why in general I would rather also retry those request.

marco-jantke: @emilevauge WDYT?

emilevauge: @timoreimann any thoughts on this ?

timoreimann: @emilevauge @marco-jantke the net package does [distinguish between an an `ErrUnexpectedEOF` and `EOF` state](https://golang.org/pkg/io/#pkg-variables). Unfortunately, it's [not used by net/http](https://github.com/golang/go/issues/18055).

Even if we considered this to be changeworthy, IMHO we should not do so along this PR which maintains the current oxy error handler behavior except for the response code logic. We can have a separate issue to drive a potential improvement in this focused regard.

emilevauge: @marco-jantke I agree with @timoreimann on this, we should deal with this issue in another PR. The fact that it is not used in the default go http client leads to discussions :)

marco-jantke: I agree, so we handled all the questions/feedback in this PR I think :)

marco-jantke:
@timoreimann @emilevauge you are ok with me doing a rebase already to resolve the conflicts?

timoreimann:
@marco-jantke I think it's okay to do that now.

marco-jantke:
Ok, everything should be ready now.

marco-jantke:
@emilevauge ping :)

ldez:
@marco-jantke We browse all PRs regularly, you don't need to ping someone.

Issue #1077, opened 2017-01-27
Support override of HTTP 404 when service does not exist or is down


grobinson-blockchain:
### What version of Traefik are you using (`traefik version`)?

    Version:      dev
    Codename:     cheddar
    Go version:   go1.7.4
    Built:        I don't remember exactly
    OS/Arch:      darwin/amd64

### Request

Quite often sending HTTP 404 when a service is down (i.e. consul health check fails) is not the correct behavior (for both browsers or clients of APIs) and we'd much rather send HTTP 502 or HTTP 503.

I'm running a fork right now that overrides the NotFoundHandler for the default mux - but it would be nice if this becomes a configurable option, perhaps even with a go template.

```golang
    func badGatewayHandler(w http.ResponseWriter, r *http.Request) {
        w.WriteHeader(http.StatusBadGateway)
        fmt.Fprintf(w, "502 bad gateway")
    }

    func (server *Server) buildDefaultHTTPRouter() *mux.Router {
        router := mux.NewRouter()
        // router.NotFoundHandler = http.HandlerFunc(notFoundHandler)
        router.NotFoundHandler = http.HandlerFunc(badGatewayHandler)
        router.StrictSlash(true)
        router.SkipClean(true)
        return router
    }
```

dtomcej:
When a service does not exist, the correct response is 404. If you dynamically generate and configure services, it can be tempting to want to return a 50x.

Unfortunately, as far as the proxy is concerned, it is designed to be stateless, relying on providers for configuration. If your provider does not respond, or has missing services, traefik has no way to tell if that was legitimately deleted, or is missing, and therefore has to treat it as the former.

If you are having issues with your provider not passing health checks etc, you may want to look at providing a frontend/backend through a different provider (such as the toml), that can provide the persistence you are looking for.

ljagiello:
@dtomcej there is a difference between page not found and lack of service. RFC2616 (https://www.ietf.org/rfc/rfc2616.txt) describe when HTTP 502 should be used:
```
10.5.3 502 Bad Gateway

   The server, while acting as a gateway or proxy, received an invalid
   response from the upstream server it accessed in attempting to
   fulfill the request.
```
Lack of service of entire service down is not the same as page not found.

dtomcej:
@ljagiello Yes, I agree, but traefik is stateless. Unless you have provided state information, traefik cannot tell that your service is down/missing/broken or never existed.

[...]

PR #1550, opened 2017-05-04
add configuration option for not found http status code

marco-jantke:
This PR adds the ability to configure the HTTP status code that should be sent, given for an incoming request, no backend can be found. Without adding any configuration, it will behave as before, sending 404 responses and therefore keeps the compatibility for the Traefik users.

This PR should solve the issue [1077](https://github.com/containous/traefik/issues/1077).

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -844,19 +844,22 @@ func (server *Server) loadEntryPointConfig(entryPointName string, entryPoint *En
 
 func (server *Server) buildDefaultHTTPRouter(globalConfiguration GlobalConfiguration) *mux.Router {
 	router := mux.NewRouter()
-	router.NotFoundHandler = newNotFounderHandler(globalConfiguration)
+	router.NotFoundHandler = newNotFounderHandler(globalConfiguration.NoRouteResponseCode)
 	router.StrictSlash(true)
 	router.SkipClean(true)
 	return router
 }
 
-func newNotFounderHandler(globalConfiguration GlobalConfiguration) http.Handler {
-	log.Debugf("Setting up not found handler to serve %d responses", globalConfiguration.NoRouteResponseCode)
-
-	responseCode := globalConfiguration.NoRouteResponseCode
+func newNotFounderHandler(responseCode int) http.Handler {
 	if responseCode == 0 {
 		responseCode = http.StatusNotFound
 	}
+	if http.StatusText(responseCode) == "" {
+		log.Warnf("Wrong NoRouteResponseCode %d configured. Falling back to 404 (http.StatusNotFound)", responseCode)
+		responseCode = http.StatusNotFound
+	}
+
+	log.Debugf("Setting up not found handler to serve %d responses", responseCode)
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(responseCode)
--- a/server/server.go
+++ b/server/server.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -540,7 +541,7 @@ func (server *Server) prepareServer(entryPointName string, router *middlewares.H
 func (server *Server) buildEntryPoints(globalConfiguration GlobalConfiguration) map[string]*serverEntryPoint {
 	serverEntryPoints := make(map[string]*serverEntryPoint)
 	for entryPointName := range globalConfiguration.EntryPoints {
-		router := server.buildDefaultHTTPRouter()
+		router := server.buildDefaultHTTPRouter(globalConfiguration)
 		serverEntryPoints[entryPointName] = &serverEntryPoint{
 			httpRouter: middlewares.NewHandlerSwitcher(router),
 		}
@@ -841,12 +842,26 @@ func (server *Server) loadEntryPointConfig(entryPointName string, entryPoint *En
 	return rewrite, nil
 }
 
-func (server *Server) buildDefaultHTTPRouter() *mux.Router {
+func (server *Server) buildDefaultHTTPRouter(globalConfiguration GlobalConfiguration) *mux.Router {
 	router := mux.NewRouter()
-	router.NotFoundHandler = http.HandlerFunc(notFoundHandler)
+	router.NotFoundHandler = newNotFounderHandler(globalConfiguration)
 	router.StrictSlash(true)
 	router.SkipClean(true)
 	return router
+}
+
+func newNotFounderHandler(globalConfiguration GlobalConfiguration) http.Handler {
+	log.Debugf("Setting up not found handler to serve %d responses", globalConfiguration.NoRouteResponseCode)
+
+	responseCode := globalConfiguration.NoRouteResponseCode
+	if responseCode == 0 {
+		responseCode = http.StatusNotFound
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(responseCode)
+		fmt.Fprint(w, http.StatusText(responseCode))
+	})
 }
 
 func parseHealthCheckOptions(lb healthcheck.LoadBalancer, backend string, hc *types.HealthCheck, hcConfig HealthCheckConfig) *healthcheck.Options {
--- a/server/server.go
+++ b/server/server.go
@@ -541,7 +541,7 @@ func (server *Server) prepareServer(entryPointName string, router *middlewares.H
 func (server *Server) buildEntryPoints(globalConfiguration GlobalConfiguration) map[string]*serverEntryPoint {
 	serverEntryPoints := make(map[string]*serverEntryPoint)
 	for entryPointName := range globalConfiguration.EntryPoints {
-		router := server.buildDefaultHTTPRouter(globalConfiguration.NoRouteResponseCode)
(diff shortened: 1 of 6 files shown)


server/server.go
```diff
@@ -835,14 +836,28 @@ func (server *Server) loadEntryPointConfig(entryPointName string, entryPoint *En
...
+func (server *Server) buildDefaultHTTPRouter(globalConfiguration GlobalConfiguration) *mux.Router {
 	router := mux.NewRouter()
-	router.NotFoundHandler = http.HandlerFunc(notFoundHandler)
+	router.NotFoundHandler = newNotFounderHandler(globalConfiguration)
 	router.StrictSlash(true)
 	router.SkipClean(true)
 	return router
 }
 
+func newNotFounderHandler(globalConfiguration GlobalConfiguration) http.Handler {
```
timoreimann: I think we should only pass in `globalConfiguration.NoRouteResponseCode` here as the function does not care about the remaining global configuration parameters.

server/server.go
```diff
@@ -835,14 +836,28 @@ func (server *Server) loadEntryPointConfig(entryPointName string, entryPoint *En
...
+	log.Debugf("Setting up not found handler to serve %d responses", globalConfiguration.NoRouteResponseCode)
+
+	responseCode := globalConfiguration.NoRouteResponseCode
+	if responseCode == 0 {
+		responseCode = http.StatusNotFound
+	}
+
```
timoreimann: We should probably do some sanity-checking at this point, like making sure that the response code is within the valid HTTP range.

server/server_test.go
```diff
@@ -158,3 +160,52 @@ func TestServerParseHealthCheckOptions(t *testing.T) {
...
+
+	for _, test := range tests {
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			notFoundHandler := newNotFounderHandler(test.globalConfig)
+			recorder := httptest.NewRecorder()
+			req := httptest.NewRequest("GET", "http://localhost/notknown", nil)
```
timoreimann: s/"GET"/http.MethodGet/

traefik.sample.toml
```diff
@@ -46,6 +46,13 @@
 #
 # logLevel = "ERROR"
 
+# HTTP response code to send when no route to backend exists
```
timoreimann: Final punctuation missing.

server/server.go
marco-jantke: Yes it simplifies things.

server/server.go
marco-jantke: I implemented this sanity checking now, falling back to `http.StatusNotFound` and logging a warning. I realised that logging a warning leads to output when running the test directly, though it is not shown when running `make test-unit`. Therefore I guess its fine.

server/server_test.go
marco-jantke: Thanks, fixed.

traefik.sample.toml
marco-jantke: Thanks, fixed.

marco-jantke:
Thanks for the review @timoreimann, I integrated it. You can have a look again :)

I wanted to the http status code validation by the way in the `newNotFounderHandler` and even though, on wrong configuration, make the log message as explicit as possible. Therefore it kind of still carries knowledge about the configuration option in general. WDYT?

server/server.go
```diff
@@ -835,14 +836,31 @@ func (server *Server) loadEntryPointConfig(entryPointName string, entryPoint *En
...
 	router.SkipClean(true)
 	return router
 }
 
+func newNotFounderHandler(responseCode int) http.Handler {
+	if responseCode == 0 {
+		responseCode = http.StatusNotFound
+	}
+	if http.StatusText(responseCode) == "" {
+		log.Warnf("Wrong NoRouteResponseCode %d configured. Falling back to 404 (http.StatusNotFound)", responseCode)
```
timoreimann: Not sure if a misconfiguration should rather lead to an `ERROR` level log message. Can you check how Traefik usually proceeds for such cases?

timoreimann: Another point: How about just letting `newNotFounderHandler` return a `(http.Handler, error)` tuple and return a non-nil error for this case? It'd then be up to the caller to handle it accordingly (i.e., fall back to 404) and not be nested inside the handler factory.

discordianfish: Should this be checked at all? If I configure a status for which there is no StatusText, it's likely I intentionally used some special code number in which case it should be fine to have an empty body, right?

Then again if you want to ensure there is always a response text, I'd make this explicitly configurable.

marco-jantke: I checked other places and saw that usually `log.Errorf` is used. Thanks for the hint, changed it accordingly.

marco-jantke: Thank you for the comment @discordianfish. I was already wondering whether this use-case exists. Indeed I was not very interested in the body itself that is returned, but used the `http.StatusText()` function as a kind of validation that the HTTP Status Code is a valid one. I was checking now the RFC and valid is basically everything in the range of 100 - 599.   See [here](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6.1.1).

Therefore I changed now the implementation to respect this use-case.

marco-jantke: @timoreimann indeed your comment addressed it. During the course I changed implementation now a bit, pushing the configuration validation and fallback, as well as the logging part, to the `buildDefaultHTTPRouter` function. This way we don't need errors. WDYT?

ldez:

server/server_test.go
```diff
@@ -161,6 +161,49 @@ func TestServerParseHealthCheckOptions(t *testing.T) {
...
+		globalConfig           GlobalConfiguration
+		wantNotFoundStatusCode int
+	}{
+		{
+			name:                   "default",
+			globalConfig:           GlobalConfiguration{},
+			wantNotFoundStatusCode: http.StatusNotFound,
+		},
+		{
+			name:                   "sane config",
```
timoreimann: Shall we call this `valid response code`...

server/server_test.go
```diff
@@ -161,6 +161,49 @@ func TestServerParseHealthCheckOptions(t *testing.T) {
...
+			globalConfig:           GlobalConfiguration{NoRouteResponseCode: http.StatusBadGateway},
+			wantNotFoundStatusCode: http.StatusBadGateway,
+		},
+		{
+			name:                   "invalid config",
```
timoreimann: ...and this one `invalid response code` for increased test naming precision?

server/server.go
timoreimann: Personally I don't like it design-wise, but it seems to be in line with what Traefik usually does. So 👍 .

server/server_test.go
marco-jantke: Good point, implemented.

server/server_test.go
marco-jantke: Good point, implemented.

marco-jantke:
@timoreimann thanks for the review, I implemented the points. I had to do a rebase to get the semaphore CI working for this branch.

splieth:
I think this won't fix #1077 since when a service does not exists, the correct status code would be 404. Now, this would lead to the configured status code.

marco-jantke:
Understood @splieth. I am personally not sure what is the better option, I will check what is the outcome of the discussion in https://github.com/containous/traefik/issues/1077#issuecomment-301532200 and will go on with the PR, once a consensus on the desired behaviour is reached.

marco-jantke:
Before I proceed implementing more in this PR, I would like to have a clear understanding on what should be implemented. There was a lengthy discussion in the original issue with a [summary](https://github.com/containous/traefik/issues/1077#issuecomment-301532200) from @timoreimann. There was only 1 response so far, but it seems to be a desirable behaviour to be able to configure different http response codes, given there is a "frontend" for the incoming route or when no frontend exists at all.

@emilevauge and @vdemeester do you agree with this approach or do you think one configurable response code suffices?

marco-jantke:
@emilevauge can we have your opinion on this? @timoreimann provided a good summary of the options here: https://github.com/containous/traefik/issues/1077#issuecomment-301532200

ldez:
We don't want to create a stateful/cache behavior.

We (maintainers) need to discuss (again) on this issue.

emilevauge:
@ldez if I understand correctly, @timoreimann summed up 2 different needs:
>1. Return a configurable error code (e.g., something from the 50[234] range) if either a target frontend or its backends are missing.
>2. Return a configurable error code if the backends to a known frontend are missing, and a 404 if the frontend is unknown.

And this PR is supposed to fix 1. Am I right @marco-jantke ?
On the implementation, don't you think it would be possible to extend #1634 instead ?

marco-jantke:
@emilevauge yes at the moment this PR only fixes point 1. At the moment I started this PR it was not yet clear what is the desired behaviour and I believe that this addition alone is also valuable. For my use-case it would be sufficient already, as the main goal is to avoid caching 404s by our CDN provider.

Regarding #1634 I am not sure whether it fits into this one. As far as I understand the proposal you would configure an error handler on per frontend. But there is also the case that no frontend can be matched for a request. So I guess we will still need a `NotFoundHandler` for the routes and this one gets improved by this PR. WDYT?

timoreimann:
@marco-jantke I closed the originating issue and created follow-ups. As discussed, I think it'd be best to close this PR and instead work towards implementing #1688. WDYT?

marco-jantke:
Agreed, lets go for it. Closing here.

PR #1552, opened 2017-05-04
 [CI] Add SemaphoreCI 

atbore-phx:
This PR adds SemaphoreCI as new CI platform.

### DIFF
--- a/.semaphoreci/tests.sh
+++ b/.semaphoreci/tests.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+make test-unit && make test-integration
+make -j${N_MAKE_JOBS} crossbinary-default-parallel
--- a/.semaphoreci/vars
+++ b/.semaphoreci/vars
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+set -e
+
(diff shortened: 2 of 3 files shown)


.semaphoreci/vars
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+set -e
+
+export secure='btt4r13t09gQlHb6gYrvGC2yGCMMHfnp1Mz1RQedc4Mpf/FfT8aE6xmK2a2i9CCvskjrP0t/BFaS4yxIURjnFRn+ugQIEa0pLspB9UJArW/vgOSpIWM9/OQ/fg8z5XuMxN6Md4DL1/iLypMNSageA1x0TRdt89+D1N1dALpg5XRCXLFbC84TLi0gjlFuib9ibPKzEhLT+anCRJ6iZMzeupDSoaCVbAtJMoDvXw4+4AcRZ1+k4MybBLyCib5boaEOt4pTT88mz4Kk0YaMwPVJyg9Qv36VqyUcPS09Yd95LuyVQ4+tZt8Y1ccbIzULsK+sLM3hLCzxlmlpN3dQBlZJiiRtQde0mgGAKyC0P0A1XjuDTywcsa5edB+fTk1Dsewz9xZ9V0NmMz8t+UNZnaSsAPga9i86jULbXUUwMVSzVRc+Xgx02liB/8qI1xYC9FM6ilStt7rn7mF0k3KbiWhcptgeXjO6Lah9FjEKd5w4MXsdUSTi/86rQaLo+kj+XdaTrXCTulKHyRyQEUj+8V1w0oVz7pcGjePHd7y5oU9ByifVQy6sytuFBfRZvugM5bKHo+i0pcWvixrZS42DrzwxZJsspANOvqSe5ifVbvOkfUppQdCBIwptxV5N1b49XPKU3W/w34QJ8xGmKp3TFA7WwVCztriFHjPgiRpB3EG99Bg='
+
+export REPO='containous/traefik'
+
+if VERSION=$(git describe --exact-match --abbrev=0 --tags);
```
timoreimann: I'm not 100% sure this will work as expected if the git sub-command produces an error. Can you double check that the script will do the right thing then?

What are you trying to achieve here actually?

.semaphoreci/vars
```diff
@@ -0,0 +1,17 @@
...
+then
+  export VERSION
+else
+  export VERSION=''
```
timoreimann: It might be more elegant to set `VERSION` where appropriate and export it once after the if statement.

timoreimann:
Merged too soon. :-(

Please consider my comments post-mergem. :-)

.semaphoreci/tests.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+make test-unit && make test-integration
```
emilevauge: As the infra is supposed to be a lot faster, we want to try without first :)

.semaphoreci/vars
atbore-phx: hi @timoreimann, on travis you have a built-in env variable "VERSION: $TRAVIS_TAG" on semaphore seems you haven't https://semaphoreci.com/docs/available-environment-variables.html
so i would retry the git tag and export the variable version.
When the tag is not present you got an error and you export an empty string (it can be useful if we want add a conditional check on release)

PR #1554, opened 2017-05-04
feat(github): push force PR branch.

ldez:
### Description

Like cpr, rmpr but for push force the branch.

### DIFF
--- a/.github/pfpr.sh
+++ b/.github/pfpr.sh
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 #
-# git config --global alias.rmpr '!sh .github/rmpr.sh'
+# git config --global alias.pfpr '!sh .github/pfpr.sh'
 
 set -e # stop on error
 
-usage="$(basename "$0") pr -- remove a Pull Request local branch & remote"
+usage="$(basename "$0") pr -- Push force a Pull Request branch"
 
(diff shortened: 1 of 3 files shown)


.github/pfpr.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
```
timoreimann: Should be `#!/usr/bin/env bash`.

.github/pfpr.sh
```diff
@@ -0,0 +1,24 @@
...
+command -v jq >/dev/null 2>&1 || { echo "I require jq but it's not installed.  Aborting." >&2; exit 1; }
+
+set -x # echo on
+
+initial=$(git rev-parse --abbrev-ref HEAD)
+pr=$1
+remote=$(curl -s https://api.github.com/repos/containous/traefik/pulls/$pr | jq -r .head.repo.owner.login)
+branch=$(curl -s https://api.github.com/repos/containous/traefik/pulls/$pr | jq -r .head.ref)
+
+git push -f $remote $pr--$branch:$branch
```
timoreimann: `-f` should be `--force-with-lease`.

.github/pfpr.sh
ldez: I've fixed all "github" scripts.

PR #1558, opened 2017-05-05
prefere yarnpkg over yarn

Stibbons:
to avoid conflict with Hadoop Yarn cli.

I don’t know the best practice, but i do
have Apache Yarn installed on my machine, so
I get this conflict. Of course this conflict does
not arised when building within the docker.

see https://github.com/yarnpkg/yarn/issues/2337
Signed-off-by: Gaetan Semet <gaetan@xeberon.net>

### DIFF
--- a/webui/readme.md
+++ b/webui/readme.md
@@ -21,6 +21,9 @@ make generate-webui  # Generate static contents in `traefik/static/` folder.
 
 - prerequisite: [Node 4+](https://nodejs.org) [yarn](https://yarnpkg.com/)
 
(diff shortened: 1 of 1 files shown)


emilevauge:
Hi @Stibbons, thanks for your PR :)
Few questions:
1/ What is the need of modifying the Dockerfile here? 
2/ For the sake of simplicity, I would prefer mentioning the issue in the readme, without replacing `yarn` by `yarnpkg`.

Stibbons:
Indeed, modifying the dockerfile is not mandatory.
I also agree to only notify this issue in the readme. It's just I spend a few minutes understanding why `yarn` was complying with aweful java stack trace :)

PR #1565, opened 2017-05-08
Enhanced flexibility in Consul Catalog configuration

aantono:
### Description

Enhanced Consul Catalog configuration by adding the following:

- Ability to configure default FrontEndRule using Go Template
- Ability to use all tags for constraint filtering
- Ability to have empty/no prefix
- Added `getTag` and `hasTag` functions to be used in advanced custom configuration template

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -1320,6 +1320,15 @@ domain = "consul.localhost"
 # Optional
 #
 prefix = "traefik"
+
+# Default frontEnd Rule for Consul services
+# The format is a Go Template with ".ServiceName", ".Domain" and ".Attributes" available
+# "getTag(name, tags, defaultValue)", "hasTag(name, tags)" and "getAttribute(name, tags, defaultValue)" functions are available
+# "getAttribute(...)" function uses prefixed tag names based on "prefix" value
+#
+# Optional
+#
+frontEndRule = "Host:{{.ServiceName}}.{{Domain}}"
 ```
 
 This backend will create routes matching on hostname based on the service name
@@ -1334,7 +1343,7 @@ Additional settings can be defined using Consul Catalog tags:
 - `traefik.backend.loadbalancer=drr`: override the default load balancing mode
 - `traefik.backend.maxconn.amount=10`: set a maximum number of connections to the backend. Must be used in conjunction with the below label to take effect.
 - `traefik.backend.maxconn.extractorfunc=client.ip`: set the function to be used against the request to determine what to limit maximum connections to the backend by. Must be used in conjunction with the above label to take effect.
-- `traefik.frontend.rule=Host:test.traefik.io`: override the default frontend rule (Default: `Host:{containerName}.{domain}`).
+- `traefik.frontend.rule=Host:test.traefik.io`: override the default frontend rule (Default: `Host:{{.ServiceName}}.{{.Domain}}`).
 - `traefik.frontend.passHostHeader=true`: forward client `Host` header to the backend.
 - `traefik.frontend.priority=10`: override default frontend priority
 - `traefik.frontend.entryPoints=http,https`: assign this frontend to entry points `http` and `https`. Overrides `defaultEntryPoints`.
--- a/provider/consul/consul_catalog.go
+++ b/provider/consul/consul_catalog.go
@@ -1,6 +1,7 @@
 package consul
 
 import (
+	"bytes"
 	"errors"
 	"sort"
 	"strconv"
@@ -31,7 +32,9 @@ type CatalogProvider struct {
 	Endpoint              string `description:"Consul server endpoint"`
 	Domain                string `description:"Default domain used"`
 	Prefix                string `description:"Prefix used for Consul catalog tags"`
+	FrontEndRule          string `description:"Frontend rule used for Consul services"`
 	client                *api.Client
+	frontEndRuleTemplate  *template.Template
 }
 
 type serviceUpdate struct {
@@ -137,9 +140,9 @@ func (p *CatalogProvider) healthyNodes(service string) (catalogUpdate, error) {
 	}
 
 	nodes := fun.Filter(func(node *api.ServiceEntry) bool {
-		constraintTags := p.getContraintTags(node.Service.Tags)
+		constraintTags := p.getConstraintTags(node.Service.Tags)
 		ok, failingConstraint := p.MatchConstraints(constraintTags)
-		if ok == false && failingConstraint != nil {
+		if !ok && failingConstraint != nil {
 			log.Debugf("Service %v pruned by '%v' constraint", service, failingConstraint.String())
 		}
 		return ok
@@ -162,6 +165,13 @@ func (p *CatalogProvider) healthyNodes(service string) (catalogUpdate, error) {
 	}, nil
 }
 
+func (p *CatalogProvider) getPrefixedName(name string) string {
+	if len(p.Prefix) > 0 {
+		return p.Prefix + "." + name
+	}
+	return name
+}
+
 func (p *CatalogProvider) getEntryPoints(list string) []string {
 	return strings.Split(list, ",")
 }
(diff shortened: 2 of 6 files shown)


provider/consul/consul_catalog.go
```diff
@@ -202,22 +231,46 @@ func (p *CatalogProvider) getBackendName(node *api.ServiceEntry, index int) stri
 }
 
 func (p *CatalogProvider) getAttribute(name string, tags []string, defaultValue string) string {
+	//for _, tag := range tags {
```
ldez: could your remove the comment block?

provider/consul/consul_catalog.go
```diff
@@ -173,10 +184,28 @@ func (p *CatalogProvider) getBackend(node *api.ServiceEntry) string {
...
+		ServiceName: service.ServiceName,
+		Domain:      p.Domain,
+		Attributes:  service.Attributes,
 	}
-	return "Host:" + service.ServiceName + "." + p.Domain
+
+	var buffer bytes.Buffer
+	t.Execute(&buffer, templateObjects)
+
+	//return "Host:" + service.ServiceName + "." + p.Domain
```
ldez: could your remove the comment?

provider/consul/consul_catalog.go
```diff
@@ -28,11 +29,14 @@ var _ provider.Provider = (*CatalogProvider)(nil)
...
-	Prefix                string `description:"Prefix used for Consul catalog tags"`
-	client                *api.Client
+	provider.BaseProvider      `mapstructure:",squash"`
+	Endpoint                   string `description:"Consul server endpoint"`
+	Domain                     string `description:"Default domain used"`
+	Prefix                     string `description:"Prefix used for Consul catalog tags"`
+	FrontEndRule               string `description:"Frontend rule used for Consul services"`
+	AllTagsConstraintFiltering bool   `description:"Should use all tags for constraint filtering"`
+	client                     *api.Client
+	FrontEndRuleTemplate       *template.Template
```
ldez: move up this line and add the description.

aantono: @ldez The `FrontEndRuleTemplate` is the actual `Template` object that is not provided via configuration, but initialized from `FrontEndRule` value.  So I see it being similar to `api.Client` kind.  Guess it should probably be made lower-cased name `frontEndRuleTemplate` instead to clearly indicate that, correct?

ldez: yes it's the way 😃

aantono: Awesome... I've made the change in the last update to this PR along with adding content to the `traefik.sample.toml` file.  Let me know if anything else needs changing or correcting.

provider/consul/consul_catalog.go
```diff
@@ -202,22 +230,38 @@ func (p *CatalogProvider) getBackendName(node *api.ServiceEntry, index int) stri
 }
 
 func (p *CatalogProvider) getAttribute(name string, tags []string, defaultValue string) string {
+	return p.getTag(p.getPrefixedName(name), tags, defaultValue)
+}
+
+func (p *CatalogProvider) hasTag(name string, tags []string) bool {
+	tag := p.getTag(name, tags, "=!=") // Very-very unlikely that a Consul tag would ever start with '=!='
```
emilevauge: @aantono Could you explain why you need `=!=` here?

provider/consul/consul_catalog.go
```diff
@@ -28,11 +29,14 @@ var _ provider.Provider = (*CatalogProvider)(nil)
...
-	Endpoint              string `description:"Consul server endpoint"`
-	Domain                string `description:"Default domain used"`
-	Prefix                string `description:"Prefix used for Consul catalog tags"`
-	client                *api.Client
+	provider.BaseProvider      `mapstructure:",squash"`
+	Endpoint                   string `description:"Consul server endpoint"`
+	Domain                     string `description:"Default domain used"`
+	Prefix                     string `description:"Prefix used for Consul catalog tags"`
+	FrontEndRule               string `description:"Frontend rule used for Consul services"`
+	AllTagsConstraintFiltering bool   `description:"Should use all tags for constraint filtering"`
```
emilevauge: Not fan of the naming of `AllTagsConstraintFiltering`. Could you explain why you added this option in a use case?

provider/consul/consul_catalog.go
```diff
@@ -202,22 +230,38 @@ func (p *CatalogProvider) getBackendName(node *api.ServiceEntry, index int) stri
...
+	return tag != "=!="
+}
+
+func (p *CatalogProvider) getTag(name string, tags []string, defaultValue string) string {
 	for _, tag := range tags {
-		if strings.Index(strings.ToLower(tag), p.Prefix+".") == 0 {
-			if kv := strings.SplitN(tag[len(p.Prefix+"."):], "=", 2); len(kv) == 2 && strings.ToLower(kv[0]) == strings.ToLower(name) {
-				return kv[1]
+		if strings.Index(strings.ToLower(tag), strings.ToLower(name)) == 0 {
+			if kv := strings.SplitN(tag, "=", 2); strings.ToLower(kv[0]) == strings.ToLower(name) {
```
emilevauge: Could you add some comments ? This block is a bit hard to read ;)

provider/consul/consul_catalog.go
```diff
@@ -202,22 +230,38 @@ func (p *CatalogProvider) getBackendName(node *api.ServiceEntry, index int) stri
...
+		return tags
+	}
+
 	var list []string
 
 	for _, tag := range tags {
-		if strings.Index(strings.ToLower(tag), p.Prefix+".tags=") == 0 {
-			splitedTags := strings.Split(tag[len(p.Prefix+".tags="):], ",")
+		if strings.Index(strings.ToLower(tag), p.getPrefixedName("tags=")) == 0 {
+			splitedTags := strings.Split(tag[len(p.getPrefixedName("tags=")):], ",")
```
emilevauge: Could you add some comments ? This block is a bit hard to read ;)

docs/toml.md
```diff
@@ -1287,6 +1287,21 @@ domain = "consul.localhost"
...
 prefix = "traefik"
+
+# Default frontEnd Rule for Consul services
+#  - The format is a Go Template with ".ServiceName", ".Domain" and ".Attributes" available
+#  -- "getTag(name, tags, defaultValue)", "hasTag(name, tags)" and "getAttribute(name, tags, defaultValue)" functions are available
+#  --- "getAttribute(...)" function uses prefixed tag names based on "prefix" value
+#
+# Optional
+#
+frontEndRule = "{{.ServiceName}}.{{Domain}}"
```
emilevauge: Shouldn't it be `Host:{{.ServiceName}}.{{Domain}}` instead?

traefik.sample.toml
```diff
@@ -832,6 +832,21 @@
...
 # prefix = "traefik"
 
+# Default frontEnd Rule for Consul services
+#  - The format is a Go Template with ".ServiceName", ".Domain" and ".Attributes" available
+#  -- "getTag(name, tags, defaultValue)", "hasTag(name, tags)" and "getAttribute(name, tags, defaultValue)" functions are available
+#  --- "getAttribute(...)" function uses prefixed tag names based on "prefix" value
+#
+# Optional
+#
+#frontEndRule = "{{.ServiceName}}.{{Domain}}"
```
emilevauge: Same as in `traefik.sample.toml`

provider/consul/consul_catalog.go
aantono: The use-case here is to enable the ability to filter (use tag constraints) on the existing Consul tags, not only the ones that are defined in `traefik.tags`  Many places with existing Consul infrastructure already have quite an elaborate tagging setup for their service registrations, so being able to leverage those for filtering is a big plus.  For example, each Spring Boot application that uses `sping-cloud-consul` module, will automatically register the management endpoint using `management` tag.  Because it might not be desired to expose LB routes to management endpoints, one might want to create a constraint rule like `constraints = ["tag!=management"]`, and when coupled with `AllTagsConstraintFiltering=true`, it will use the "raw" `management` tag value, instead of only the values provided by `traefik.tags` tag.  Does this make things clearer?

provider/consul/consul_catalog.go
aantono: This is my internal dummy "default" value (because `getTag` function want's a default) which is unlikely to be seen in the real world (who would tag their service with `=!=`... :) )
The logic is this: if `getTag(...)` returns `=!=`, then there is no `name` tag in existence, otherwise we would get a different value.

docs/toml.md
aantono: Yes, indeed it should. Great catch!

docs/toml.md
```diff
@@ -1301,6 +1301,21 @@ domain = "consul.localhost"
 # Optional
 #
 prefix = "traefik"
+
+# Default frontEnd Rule for Consul services
+#  - The format is a Go Template with ".ServiceName", ".Domain" and ".Attributes" available
+#  -- "getTag(name, tags, defaultValue)", "hasTag(name, tags)" and "getAttribute(name, tags, defaultValue)" functions are available
+#  --- "getAttribute(...)" function uses prefixed tag names based on "prefix" value
```
emilevauge: Could explain why you added in the comments ?
```
-
--
---
```

aantono: No reason, just to help with indentation...  Removed.

script/crossbinary-default
```diff
@@ -33,12 +34,12 @@ for OS in ${OS_PLATFORM_ARG[@]}; do
...
-done
+## Build arm64 binaries
+#OS_PLATFORM_ARG=(linux)
+#OS_ARCH_ARG=(arm64)
+#for OS in ${OS_PLATFORM_ARG[@]}; do
+#  for ARCH in ${OS_ARCH_ARG[@]}; do
+#    echo "Building binary for $OS/$ARCH..."
+#    GOARCH=$ARCH GOOS=$OS CGO_ENABLED=0 $GO_BUILD_CMD "$GO_BUILD_OPT" -o "dist/traefik_$OS-$ARCH" ./cmd/traefik/
+#  done
+#done
```
emilevauge: I don't think that's what we want ^^

aantono: Oops, my bad... accidentally committed something that should not have been...  will revert ASAP

PR #1567, opened 2017-05-09
Allow overriding the port used for healthchecks

bakins:
Some services use a different port for auxiliary functions such as healthchecks and metrics.  Traefik itself does this.  This PR allows overriding the port used per backend.

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -296,8 +296,9 @@ A health check can be configured in order to remove a backend from LB rotation
 as long as it keeps returning HTTP status codes other than 200 OK to HTTP GET
 requests periodically carried out by Traefik. The check is defined by a path
 appended to the backend URL and an interval (given in a format understood by [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration)) specifying how
-often the health check should be executed (the default being 30 seconds). Each
-backend must respond to the health check within 5 seconds.
+often the health check should be executed (the default being 30 seconds).
+Each backend must respond to the health check within 5 seconds.
+By default, the port of the backend server is used, however, this may be overridden.  
 
 A recovering backend returning 200 OK responses again is being returned to the
 LB rotation pool.
@@ -309,6 +310,16 @@ For example:
     [backends.backend1.healthcheck]
       path = "/health"
       interval = "10s"
+```
+
+To use a different port for the healthcheck:
+```toml
+[backends]
+  [backends.backend1]
+    [backends.backend1.healthcheck]
+      path = "/health"
+      interval = "10s"
+      port = 8080
 ```
 
 ## Servers
@@ -346,17 +357,17 @@ Here is an example of backends and servers definition:
 
 # Configuration
 
-Træfik's configuration has two parts: 
+Træfik's configuration has two parts:
 
-- The [static Træfik configuration](/basics#static-trfk-configuration) which is loaded only at the beginning. 
+- The [static Træfik configuration](/basics#static-trfk-configuration) which is loaded only at the beginning.
 - The [dynamic Træfik configuration](/basics#dynamic-trfk-configuration) which can be hot-reloaded (no need to restart the process).
 
 
 ## Static Træfik configuration
 
-The static configuration is the global configuration which is setting up connections to configuration backends and entrypoints. 
+The static configuration is the global configuration which is setting up connections to configuration backends and entrypoints.
 
-Træfik can be configured using many configuration sources with the following precedence order. 
+Træfik can be configured using many configuration sources with the following precedence order.
 Each item takes precedence over the item below it:
 
 - [Key-value Store](/basics/#key-value-stores)
@@ -398,18 +409,18 @@ Træfik supports several Key-value stores:
 
 - [Consul](https://consul.io)
 - [etcd](https://coreos.com/etcd/)
-- [ZooKeeper](https://zookeeper.apache.org/) 
+- [ZooKeeper](https://zookeeper.apache.org/)
 - [boltdb](https://github.com/boltdb/bolt)
 
 Please refer to the [User Guide Key-value store configuration](/user-guide/kv-config/) section to get documentation on it.
 
 ## Dynamic Træfik configuration
 
-The dynamic configuration concerns : 
+The dynamic configuration concerns :
 
 - [Frontends](/basics/#frontends)
-- [Backends](/basics/#backends) 
-- [Servers](/basics/#servers) 
+- [Backends](/basics/#backends)
+- [Servers](/basics/#servers)
 
 Træfik can hot-reload those rules which could be provided by [multiple configuration backends](/toml/#configuration-backends).
 
@@ -427,7 +438,7 @@ List of Træfik available commands with description :             
 - `version` : Print version 
 - `storeconfig` : Store the static traefik configuration into a Key-value stores. Please refer to the [Store Træfik configuration](/user-guide/kv-config/#store-trfk-configuration) section to get documentation on it.
 
-Each command may have related flags. 
+Each command may have related flags.
 All those related flags will be displayed with :
 
 ```bash
@@ -439,4 +450,3 @@ Note that each command is described at the beginning of the help section:
 ```bash
 $ traefik --help
 ```
-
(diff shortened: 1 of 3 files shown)


healthcheck/healthcheck.go
```diff
@@ -127,11 +130,36 @@ func checkBackend(currentBackend *BackendHealthCheck) {
...
+	// make a copy of the url
+	u := &url.URL{}
+	*u = *serverURL
+
+	if backend.Options.Port != 0 {
+		u.Host = net.JoinHostPort(u.Hostname(), strconv.Itoa(backend.Options.Port))
+	}
+	u.Path = u.Path + backend.Path
+
+	req := &http.Request{
```
timoreimann: I'm inclined to say we should just call `http.NewRequest` to be on the safe side, especially if more fields are added over time that might be important for us but would likely be left behind.

WDYT?

healthcheck/healthcheck_test.go
```diff
@@ -136,60 +137,67 @@ func TestSetBackendsConfiguration(t *testing.T) {
...
-			defer cancel()
-			ts := newTestServer(cancel, test.healthSequence)
-			defer ts.Close()
-
-			lb := &testLoadBalancer{RWMutex: &sync.RWMutex{}}
-			backend := NewBackendHealthCheck(Options{
-				Path:     "/path",
-				Interval: healthCheckInterval,
-				LB:       lb,
+	for _, withPort := range []bool{false, true} {
```
timoreimann: I don't think we need to test all permutations here. I suggest we write a test for the newly created `newRequest` method instead where we won't need to jump through as many hoops (as we do in this test).

bakins: That makes sense. I'll just write a test for newRequest.

healthcheck/healthcheck.go
bakins: It seems inefficient to me to start with a URL, convert it to a string, and then have [NewRequest](https://golang.org/src/net/http/request.go?s=24894:24966#L748) parse it back into a URL.  In this code path, it probably doesn't matter.  I can change to NewRequest.

bakins:
@timoreimann I've made the requested changes. PTAL when you have a chance.

Thanks for the feedback!

healthcheck/healthcheck.go
```diff
@@ -127,11 +130,31 @@ func checkBackend(currentBackend *BackendHealthCheck) {
...
+}
+
 func checkHealth(serverURL *url.URL, backend *BackendHealthCheck) bool {
 	client := http.Client{
 		Timeout: backend.requestTimeout,
 	}
-	resp, err := client.Get(serverURL.String() + backend.Path)
+	req, err := backend.newRequest(serverURL)
+	if err != nil {
+		return false
```
timoreimann: Could you please add a log message here. I'm inclined to say it should be of level `ERROR` since we should apparently only run into this case if we have a logic error in our code.

healthcheck/healthcheck_test.go
```diff
@@ -193,6 +193,64 @@ func TestSetBackendsConfiguration(t *testing.T) {
...
+		{
+			host:     "backend2",
+			port:     8080,
+			path:     "/health",
+			expected: "http://backend2:8080/health",
+		},
+	}
+
+	for _, test := range tests {
+		backend := NewBackendHealthCheck(
```
timoreimann: Please use parallelized sub-tests here. See [one of the existing tests](https://github.com/containous/traefik/blob/acc7865542e0558de4de47b18c3acbe5d51d4b5c/healthcheck/healthcheck_test.go#L140-L142) for what needs to be changed for this (in a nutshell:

- `test := test`
- `t.Run(...)`
- `t.Parallel()`

)

healthcheck/healthcheck_test.go
```diff
@@ -193,6 +193,64 @@ func TestSetBackendsConfiguration(t *testing.T) {
...
+				Port: test.port,
+			})
+
+		u := &url.URL{
+			Scheme: "http",
+			Host:   test.host,
+		}
+
+		req, err := backend.newRequest(u)
+
```
timoreimann: Nit-pick: I'd prefer no blank line here.

healthcheck/healthcheck_test.go
```diff
@@ -193,6 +193,64 @@ func TestSetBackendsConfiguration(t *testing.T) {
...
+		if err != nil {
+			t.Errorf("failed to create new backend request for %s", u.String())
```
timoreimann: Once we use sub-tests, it's safe (and appropriate) to do `t.Fatalf` here.

healthcheck/healthcheck_test.go
```diff
@@ -193,6 +193,64 @@ func TestSetBackendsConfiguration(t *testing.T) {
...
+		}
+
+		actual := req.URL.String()
+		if test.expected != actual {
```
timoreimann: Nit-pick: Put actual on the left-hand and expected on the right-hand side to match the order in the error message on the following line.

healthcheck/healthcheck_test.go
timoreimann: Please include all test case-identifying input parameters and the error message in the output, e.g.:

`"URL=%s/%s port=%d: failed to create new backend request: %s", u.String(), test.path, test.port, err)`

timoreimann: Do we need the `.String()` part in `u.String()`? Is the formatter possibly "smart enough" to figure it out from `u` alone?

healthcheck/healthcheck_test.go
```diff
@@ -193,6 +193,64 @@ func TestSetBackendsConfiguration(t *testing.T) {
...
+			t.Errorf("got %s for healthcheck URL, wanted %s", actual, test.expected)
```
timoreimann: Include test-case identifiers here too.

Since we need it at two spots, it seems reasonable to me to extract them into a variable.

healthcheck/healthcheck.go
timoreimann: (I saw you made the change.)

Yeah it seems a bit cumbersome; personally though, I'd like to err on the side of safety and stability.

healthcheck/healthcheck.go
bakins: Sure.

healthcheck/healthcheck_test.go
bakins: Sure. I wasn't sure it was worth the effort, but I can make it match the others.

healthcheck/healthcheck_test.go
```diff
@@ -227,28 +232,34 @@ func TestNewRequest(t *testing.T) {
...
-			Scheme: "http",
-			Host:   test.host,
-		}
+			u := &url.URL{
+				Scheme: "http",
+				Host:   test.host,
+			}
 
-		req, err := backend.newRequest(u)
+			testCase := fmt.Sprintf("URL=%s/%s port=%d", u, test.path, test.port)
```
timoreimann: Now that you've added a description to each test case and injected that into the `t.Run()` call, we don't need the prefix anymore. :-)

Would you mind removing just that part again?

bakins:
@timoreimann changes made. Let me know when it looks okay and I can squash and repush, if you want.

bakins:
@timoreimann it's all in one commit now. Not the normal way I'd squash things, but it's done :)

timoreimann:
@bakins I actually like my commits squashed to atomic scale. ;-)

@emilevauge @vdemeester your review turn.

docs/basics.md
```diff
@@ -278,7 +278,8 @@ as long as it keeps returning HTTP status codes other than 200 OK to HTTP GET
 requests periodically carried out by Traefik. The check is defined by a path
 appended to the backend URL and an interval (given in a format understood by [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration)) specifying how
 often the health check should be executed (the default being 30 seconds). Each
-backend must respond to the health check within 5 seconds.
+backend must respond to the health check within 5 seconds. By default, the port
+of the backend sever is used, however, this may be overridden.
```
emilevauge: sed/sever/server :nerd_face:

PR #1571, opened 2017-05-09
Restore: First stage of access logging middleware.

ldez:
### Description

Restore access logging middleware.

Related to #1541, #1408, #1485


middlewares/accesslog/save_backend.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (sb *SaveFrontend) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
+	table := GetLogDataTable(r)
+	table.Core[FrontendName] = strings.TrimPrefix(sb.frontendName, "frontend-")
+
+	sb.next.ServeHTTP(rw, r)
+}
+
+//-------------------------------------------------------------------------------------------------
+// the next 3 function (SaveNegroniFrontend, NewSaveNegroniFrontend, ServeHTTP) are temporary,
```
ldez: Because the code evolve between the first merge and the revert restoration.

timoreimann: Okay, I realized now that the functions _were_ in the original PR but didn't include the `Negroni` part in the name.

Why the warning at the top and why change the names?

emilevauge: @timoreimann we worked both on resolving this conflict. we had to do this to manage different types negroni and http handlers.

timoreimann: 👍

PR #1576, opened 2017-05-09
Custom error page feature

bparli:
### Description
Since we are using Traefik as our ingress load balancer we need to display a custom error page in the case of a backend returning a 500 or empty response.  Currently Traefik displays a blank page with only the http status (for reference, see the code [here](https://github.com/containous/traefik/blob/master/vendor/github.com/vulcand/oxy/utils/handler.go#L30))

I didn't see a related issue raised or this feature being worked on yet, so have created this PR to address it:
- Add ErrorPages handler for serving custom error page (the middlewares package seemed like the best place to put this)
- Add error page handler to frontend forwarder
- Add error page handler to backend loadbalancer
- Expose optional configuration in traefik.toml
- Add necessary documentation/example
- Unit tests to ensure page is being served correctly

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -33,6 +33,13 @@
 #
 # checkNewVersion = false
 
+# Enable the use of a custom error page
+#
+# Optional
+#
+# [ErrorPages]
+# errorPage = "/usr/traefik/Custom.html"
+
 # Traefik logs file
 # If not defined, logs to stdout
 #
--- a/docs/user-guide/examples.md
+++ b/docs/user-guide/examples.md
@@ -138,3 +138,10 @@ defaultEntryPoints = ["http"]
 IdleTimeout = "360s"
 ProvidersThrottleDuration = "5s"
 ```
+
+## Return a preferred custom error page instead of the default Traefik error page.  
+
+```
+[ErrorPages]
+errorPage = "/usr/traefik/Custom.html"
+```
--- a/integration/vendor/github.com/containous/traefik/server.go
+++ b/integration/vendor/github.com/containous/traefik/server.go
@@ -1,7 +1,4 @@
-/*
-Copyright
-*/
-package main
+package server
 
 import (
 	"context"
@@ -17,18 +14,19 @@ import (
 	"reflect"
 	"regexp"
 	"sort"
+	"sync"
 	"syscall"
 	"time"
 
 	"github.com/codegangsta/negroni"
 	"github.com/containous/mux"
 	"github.com/containous/traefik/cluster"
+	"github.com/containous/traefik/healthcheck"
 	"github.com/containous/traefik/log"
 	"github.com/containous/traefik/middlewares"
 	"github.com/containous/traefik/provider"
 	"github.com/containous/traefik/safe"
 	"github.com/containous/traefik/types"
-	"github.com/mailgun/manners"
 	"github.com/streamrail/concurrent-map"
 	"github.com/vulcand/oxy/cbreaker"
 	"github.com/vulcand/oxy/connlimit"
@@ -57,13 +55,16 @@ type Server struct {
 type serverEntryPoints map[string]*serverEntryPoint
 
 type serverEntryPoint struct {
-	httpServer *manners.GracefulServer
+	httpServer *http.Server
 	httpRouter *middlewares.HandlerSwitcher
 }
 
 type serverRoute struct {
-	route         *mux.Route
-	stripPrefixes []string
+	route              *mux.Route
+	stripPrefixes      []string
+	stripPrefixesRegex []string
+	addPrefix          string
+	replacePath        string
 }
 
 // NewServer returns an initialized Server.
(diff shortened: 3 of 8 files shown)


ldez:
Great contribution! 😃 

Could you rebase and solve the conflicts?

bparli:
Sorry @ldez , I didn't notice the conflicts at first, they should be resolved now.

timoreimann:
AFAIU, this PR conflicts with #1549 (_retry only on real network errors_) where a (different) custom error handler is injected to signal whether a real network error occurred in order to restrict retries to such cases.

We could think about how to enable multiple error handlers and/or resolve the conflicting behavior between the two PRs' handlers. My thinking is though that the custom error page handler should not act before the retry handler and thus might be better implemented as a regular middleware injected at the proper place within the middleware chain.
With the designated implementation of this PR, the custom pages served right now by the error handler would probably be ignored by the retry handler, and while that hopefully shouldn't affect the overall result (either the retry handler succeeds and returns a proper result, or it does not and passes back the error page), it might hint at a chance to do better at the design stage.

@bparli @ldez @emilevauge @marco-jantke WDYT?

timoreimann:
@bparli have a look at the result of your rebase: you modified a file in `vendor/`, likely because git confused the vendored copy of server.go for the real one.

Sometimes tuning the `-X find-renames` git parameter helps.

marco-jantke:
Hi @bparli and @timoreimann, 

thanks for the notice about the conflicting approach. 

I had a look at this PR and I think we should be able to integrate this as standalone middleware. It only has to know the status code from the request. Have a look [here](https://github.com/containous/traefik/blob/master/middlewares/metrics.go#L36) on one possible way to accomplish this (using a decorated ResponseWriter). Additionally as standalone middleware it could be only enabled, given there is a custom error page configured. Do you agree?

bparli:
Thanks for bubbling this up again @marco-jantke.  I do agree with @timoreimann this is probably a larger design decision and your proposal is certainly one way to do it.  However, I'm wondering if adding a middleware to the chain solely for another errorhandler might be breaking from convention a bit.  Also, are you guys sure you want the errorhandler and retry functionalities so tightly coupled?

timoreimann:
@bparli IIRC my last discussion on the subject with @marco-jantke, one problem we were facing is that only the last middleware in the chain is able to distinguish between various network errors, so effectively we have to set the error status in the error handler.

What we could possibly do is pass a closure over multiple error handlers into the single error handler and have the implementation go over each and every, executing whatever custom error handling logic the callers want. WDYT?

marco-jantke:
I think its a reasonable approach to go this way.

Even though I still think that the error page should rather be implemented as middleware. The error handler is really specific to errors that happen when forwarding the request to the actual backend. An error page I would expect to see in case an error happens at "any" place, so it could also be an error in the middleware chain in Traefik. I don't know the codebase well enough yet, to tell whether this makes sense though.

bparli:
It seems a challenge we've run into is the error handler is somewhat ubiquitous in the codebase.  The closure approach sounds like a good one to me too @timoreimann

marco-jantke:
It also sounds good to me :) @bparli do you have a suggestion how to go on from here? I am not sure what is best for _who_ of us should implement this _when_?

timoreimann:
I think we should await some feedback on the design part from @emilevauge before proceeding in any direction.

Personally, I think we are all in agreement that the closure approach seems advisable to include multiple error handlers. What I'm still not clear on is whether the custom error page should be implemented in terms of an error handler or rather a custom middleware placed at the appropriate location within the middleware chain for reasons described earlier by @marco-jantke and myself. Depending on the decision made, we either need to generalize the error handler now (i.e., we return custom error pages from error handlers) or can postpone that effort to a point in time when the first "real" second error handler (i.e., custom error pages become a middleware) arrives.

Calling Emile for the design decision rescue. :-)

emilevauge:
@bparli Thanks a lot for your contribution :)
I know this feature is long-awaited, but I'm not sure I agree with the design.
I don't think Traefik should manage any html files (it's not a web server).
Besides, I'm pretty sure people will ask to be able to manage errors by frontend.

All in all, here is my proposal:
By frontend, we could delegate error page rendering on status code. For example, if I get a `500` status on frontend `website`, I query the frontend `error` with path `/500.html` and the error page would then be rendered by a specific service in your infrastructure with a webserver, Nginx for example.

```
 [frontends]
   [frontends.website]
   backend = "website"
   [errors]
     [error.network]
     status = "[500-600]"
     backend = "error"
     query = "/{status}.html"
   [frontends.website.routes.website]
   rule = "Host: website.mydomain.com"

 [backends]
   [backends.website]
     [backends.website.servers.website]
     url = "https://1.2.3.4"
   [backends.error]
     [backends.error.servers.error]
     url = "http://2.3.4.5"
```

WDYT?
/cc @containous/traefik

timoreimann:
@emilevauge's proposal sounds similar to how Kubernetes delegates 404 handlers to arbitrary Pods. Since we're in a similar territory here, it does look reasonable and more flexible (at the price of being a bit less convenient).

bparli:
Thanks @emilevauge. FWIW, I like your proposed design/interface better than my original.

ldez:
I love this proposal ❤️ 
@emilevauge It's time to open a new proposal issue 😃

emilevauge:
Proposal opened #1634. Closing this one. Thanks @bparli !

Issue #1569, opened 2017-05-09
ReplacePath rule executes out of order, when combined with PathPrefixStrip


aantono:
I believe I found a possible bug, or at least a not intuitive behavior when ReplacePath rule is preceded by PathPrefixStrip
Basically because PathPrefixStrip is a Matcher, but also happens to be a Modifier as well, when combined with ReplacePath, like PathPrefixStrip:/management;ReplacePath:/health, due to the order in which the Modifiers are applied in server.go, the above rule-set does not behave as expected.

While one would expect that we match the request to /management url, strip the /management prefix, followed by a replace with /health. This is NOT, however, what is happening. The matcher does the job correctly and matches the request on /management and sends it down the correct Handler chain, but that's when the issue creeps in.

Because ReplacePath Modifier gets executed first in the Handler chain, it replaces /management with /health, and when StripPrefix Modifier gets called, it is trying to match the Path with /management prefix, and obviously failing to do so, thus returning 404-Not Found.

If the Rule is written as PathPrefixStrip:/management;ReplacePath:/management/health, then everything works, but the actual invocation URL Path ends up being NOT /management/health but /health, thus making this configuration be misleading.

This issue relates to #1374

ldez:
Closed by  #1577

PR #1577, opened 2017-05-09
Fixed ReplacePath rule executing out of order, when combined with PathPrefixStrip

aantono:
Addresses issue raised in  #1569

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -191,6 +191,25 @@ backend = "backend2"
   rule = "Path:/test1,/test2"
 ```
 
+### Rules Order
+
+When combining `Modifier` rules with `Matcher` rules, it is important to remember that `Modifier` rules **ALWAYS** apply after the `Matcher` rules.  
+The following rules are both `Matchers` and `Modifiers`, so the `Matcher` portion of the rule will apply first, and the `Modifier` will apply later.
+
+- `PathStrip`
+- `PathStripRegex`
+- `PathPrefixStrip`
+- `PathPrefixStripRegex`
+
+`Modifiers` will be applied in a pre-determined order regardless of their order in the `rule` configuration section.
+
+1. `PathStrip`
+2. `PathPrefixStrip`
+3. `PathStripRegex`
+4. `PathPrefixStripRegex`
+5. `AddPrefix`
+6. `ReplacePath` 
+
 ### Priorities
 
 By default, routes will be sorted (in descending order) using rules length (to avoid path overlap):
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -2,13 +2,16 @@ package server
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
 	"reflect"
 	"testing"
 	"time"
 
 	"github.com/containous/flaeg"
+	"github.com/containous/mux"
 	"github.com/containous/traefik/healthcheck"
+	"github.com/containous/traefik/testhelpers"
 	"github.com/containous/traefik/types"
 	"github.com/vulcand/oxy/roundrobin"
 )
@@ -25,6 +28,85 @@ func (lb *testLoadBalancer) UpsertServer(u *url.URL, options ...roundrobin.Serve
 
 func (lb *testLoadBalancer) Servers() []*url.URL {
 	return []*url.URL{}
+}
+
+func TestServerMultipleFrontendRules(t *testing.T) {
+	cases := []struct {
+		expression  string
+		requestURL  string
+		expectedURL string
+	}{
+		{
+			expression:  "Host:foo.bar",
+			requestURL:  "http://foo.bar",
+			expectedURL: "http://foo.bar",
+		},
+		{
+			expression:  "PathPrefix:/management;ReplacePath:/health",
+			requestURL:  "http://foo.bar/management",
+			expectedURL: "http://foo.bar/health",
+		},
+		{
+			expression:  "Host:foo.bar;AddPrefix:/blah",
+			requestURL:  "http://foo.bar/baz",
+			expectedURL: "http://foo.bar/blah/baz",
+		},
+		{
+			expression:  "PathPrefixStripRegex:/one/{two}/{three:[0-9]+}",
+			requestURL:  "http://foo.bar/one/some/12345/four",
+			expectedURL: "http://foo.bar/four",
+		},
+		{
+			expression:  "PathPrefixStripRegex:/one/{two}/{three:[0-9]+};AddPrefix:/zero",
+			requestURL:  "http://foo.bar/one/some/12345/four",
+			expectedURL: "http://foo.bar/zero/four",
+		},
+		{
+			expression:  "AddPrefix:/blah;ReplacePath:/baz",
+			requestURL:  "http://foo.bar/hello",
+			expectedURL: "http://foo.bar/baz",
+		},
+		{
+			expression:  "PathPrefixStrip:/management;ReplacePath:/health",
+			requestURL:  "http://foo.bar/management",
+			expectedURL: "http://foo.bar/health",
+		},
+	}
+
+	for _, test := range cases {
+		test := test
+		t.Run(test.expression, func(t *testing.T) {
+			t.Parallel()
+
+			router := mux.NewRouter()
+			route := router.NewRoute()
+			serverRoute := &serverRoute{route: route}
+			rules := &Rules{route: serverRoute}
+
+			expression := test.expression
+			routeResult, err := rules.Parse(expression)
+
+			if err != nil {
+				t.Fatalf("Error while building route for %s: %+v", expression, err)
+			}
+
+			request := testhelpers.MustNewRequest(http.MethodGet, test.requestURL, nil)
+			routeMatch := routeResult.Match(request, &mux.RouteMatch{Route: routeResult})
+
+			if !routeMatch {
+				t.Fatalf("Rule %s doesn't match", expression)
+			}
+
+			server := new(Server)
+
+			server.wireFrontendBackend(serverRoute, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r.URL.String() != test.expectedURL {
+					t.Fatalf("got URL %s, expected %s", r.URL.String(), test.expectedURL)
+				}
+			}))
+			serverRoute.route.GetHandler().ServeHTTP(nil, request)
+		})
+	}
 }
 
 func TestServerLoadConfigHealthCheckOptions(t *testing.T) {
--- a/testhelpers/helpers.go
+++ b/testhelpers/helpers.go
@@ -1,6 +1,21 @@
 package testhelpers
 
+import (
+	"fmt"
+	"io"
+	"net/http"
+)
+
 // Intp returns a pointer to the given integer value.
 func Intp(i int) *int {
 	return &i
+}
+
+// MustNewRequest creates a new http get request or panics if it can't
+func MustNewRequest(method, urlStr string, body io.Reader) *http.Request {
+	request, err := http.NewRequest(method, urlStr, body)
+	if err != nil {
+		panic(fmt.Sprintf("failed to create HTTP %s Request for '%s': %s", method, urlStr, err))
+	}
+	return request
 }
--- a/server/server.go
+++ b/server/server.go
@@ -776,7 +776,17 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 }
 
 func (server *Server) wireFrontendBackend(serverRoute *serverRoute, handler http.Handler) {
-	// add prefix
+	// path replace - This needs to always be the very last on the handler chain (first in the order in this function)
+	// -- Replacing Path should happen at the very end of the Modifier chain, after all the Matcher+Modifiers ran
+	if len(serverRoute.replacePath) > 0 {
+		handler = &middlewares.ReplacePath{
+			Path:    serverRoute.replacePath,
+			Handler: handler,
+		}
+	}
+
+	// add prefix - This needs to always be right before ReplacePath on the chain (second in order in this function)
+	// -- Adding Path Prefix should happen after all *Strip Matcher+Modifiers ran, but before Replace (in case it's configured)
 	if len(serverRoute.addPrefix) > 0 {
 		handler = &middlewares.AddPrefix{
 			Prefix:  serverRoute.addPrefix,
@@ -795,14 +805,6 @@ func (server *Server) wireFrontendBackend(serverRoute *serverRoute, handler http
 	// strip prefix with regex
 	if len(serverRoute.stripPrefixesRegex) > 0 {
 		handler = middlewares.NewStripPrefixRegex(handler, serverRoute.stripPrefixesRegex)
-	}
-
-	// path replace
-	if len(serverRoute.replacePath) > 0 {
-		handler = &middlewares.ReplacePath{
-			Path:    serverRoute.replacePath,
-			Handler: handler,
-		}
 	}
 
 	serverRoute.route.Handler(handler)


middlewares/addPrefix.go
```diff
@@ -11,7 +12,9 @@ type AddPrefix struct {
 }
 
 func (s *AddPrefix) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	r.URL.Path = s.Prefix + r.URL.Path
+	newPath := s.Prefix + r.URL.Path
+	log.Debugf("Adding Prefix %s to %s = %s", s.Prefix, r.URL.Path, newPath)
```
emilevauge: Could you remove this Debug log? This is way too verbose to keep it in production code :)

middlewares/replace_path.go
```diff
@@ -15,6 +16,12 @@ const ReplacedPathHeader = "X-Replaced-Path"
 
 func (s *ReplacePath) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	r.Header.Add(ReplacedPathHeader, r.URL.Path)
+	log.Debugf("Replacing Path %s with %s", r.URL.Path, s.Path)
```
emilevauge: Could you remove this Debug log? This is way too verbose to keep it in production code :)

middlewares/stripPrefix.go
```diff
@@ -18,6 +19,7 @@ type StripPrefix struct {
 func (s *StripPrefix) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	for _, prefix := range s.Prefixes {
 		if p := strings.TrimPrefix(r.URL.Path, strings.TrimSpace(prefix)); len(p) < len(r.URL.Path) {
+			log.Debugf("Striping Path Prefix %s from %s = %s", prefix, r.URL.Path, p)
```
emilevauge: Could you remove this Debug log? This is way too verbose to keep it in production code :)

middlewares/stripPrefixRegex.go
```diff
@@ -39,7 +39,9 @@ func (s *StripPrefixRegex) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
-		r.URL.Path = r.URL.Path[len(prefix.Path):]
+		newPath := r.URL.Path[len(prefix.Path):]
+		log.Debugf("Striping Path Prefix Regex %s from %s = %s", prefix.Path, r.URL.Path, newPath)
```
emilevauge: Could you remove this Debug log? This is way too verbose to keep it in production code :)

middlewares/addPrefix.go
aantono: Hello @emilevauge, I definitely can, I just added it because I thought it would be very useful to see in the debug mode (in production) to see how the modifiers get applied.  It actually helped a lot while testing the config to make sure it works properly, and once deployed in prod, the debug mode is typically silenced, so the logs are not too verbose.  But if you think that's not necessary, I'll surely take it out.  Just let me know what your thoughts are on this.

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+			expectedURL: "http://foo.bar/baz",
+		},
+		{
+			expression:  "PathPrefixStrip:/management;ReplacePath:/health",
+			requestURL:  "http://foo.bar/management",
+			expectedURL: "http://foo.bar/health",
+		},
+	}
+
+	for _, test := range cases {
```
timoreimann: Please use sub-tests and call `t.Parallel()` inside. (Make sure you don't forget `test := test` to capture the loop variable.)

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+		router := mux.NewRouter()
+		route := router.NewRoute()
+		serverRoute := &serverRoute{route: route}
+		rules := &Rules{route: serverRoute}
+
+		expression := test.expression //"Host:foo.bar"
```
timoreimann: I don't think we need the comment.

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+		routeResult, err := rules.Parse(expression)
+
+		if err != nil {
+			t.Fatal("Error while building route for Host:foo.bar")
```
timoreimann: Please include the error.

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+		}
+
+		request, err := http.NewRequest("GET", test.requestURL, nil)
```
timoreimann: Suggestion: Use `http.MethodGet` instead of the literal method name.

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+		routeMatch := routeResult.Match(request, &mux.RouteMatch{Route: routeResult})
+
+		if routeMatch == false {
+			t.Log(err)
```
timoreimann: No need to have this. Just use `t.Fatalf` in the following line and pass the `err` along.

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+			requestURL:  "http://foo.bar/management",
+			expectedURL: "http://foo.bar/health",
+		},
+		{
+			expression:  "Host:foo.bar;AddPrefix:/blah",
+			requestURL:  "http://foo.bar/baz",
+			expectedURL: "http://foo.bar/blah/baz",
+		},
+		{
+			expression:  "PathPrefixStripRegex:/one/{two}/{three:[0-9]+}",
```
timoreimann: I have never seen `PathPrefixStripRegex`, only `PathPrefixStrip`. Is that from gorilla? Why not use `PathPrefixStrip`?

server/server_test.go
timoreimann: Could you please add a `MustNewRequest` method to `testhelpers/helpers.go` and call it from here? The implementation should panic similar to [`MustParseURL`](https://github.com/containous/traefik/pull/1485/files#diff-01f5a8447a1f7d87c65cbd26dc925dd1R14).

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+		routeResult, err := rules.Parse(expression)
+
+		if err != nil {
+			t.Fatal("Error while building route for Host:foo.bar")
+		}
+
+		request, err := http.NewRequest("GET", test.requestURL, nil)
+		routeMatch := routeResult.Match(request, &mux.RouteMatch{Route: routeResult})
+
+		if routeMatch == false {
```
timoreimann: simplify: `if !routeMatch {`

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+			t.Log(err)
+			t.Fatal("Rule Host:foo.bar don't match")
+		}
+
+		server := new(Server)
+
+		server.wireFrontendBackend(serverRoute, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if test.expectedURL != r.URL.String() {
```
timoreimann: [In Go](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures), _actual_ comes first, _expected_ second.

Adjust in the next line too please.

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+			Routes: map[string]types.Route{
+				"single-path-rule": {
+					Rule: "Path:/",
+				},
+			},
+		},
+	}
+
+	for _, lbMethod := range []string{"Wrr", "Drr"} {
+		for _, rule := range rules {
```
timoreimann: You need to have `rule := rule` between line 128 and 129.

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+		{
+			Routes: map[string]types.Route{
+				"single-path-rule": {
+					Rule: "Path:/",
+				},
+			},
+		},
+	}
+
+	for _, lbMethod := range []string{"Wrr", "Drr"} {
```
timoreimann: I'd suggest to export [`loadBalancerMethodNames`](https://github.com/containous/traefik/blob/8a551d91fd353c22a4c6092a4fcd83694e5423aa/types/types.go#L77) and iterate through that.

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+		for _, rule := range rules {
+			t.Run(fmt.Sprintf("%s/hc=%t", lbMethod, true), func(t *testing.T) {
+				globalConfig := GlobalConfiguration{
+					EntryPoints: EntryPoints{
+						"http": &EntryPoint{},
+					},
+					HealthCheck: &HealthCheckConfig{Interval: flaeg.Duration(5 * time.Second)},
```
timoreimann: We shouldn't need the health check for this test.

(By the way: In a singular, unified test, we'd just pass `nil` for "your" test cases.)

server/server_test.go
```diff
@@ -27,6 +29,145 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+			if test.expectedURL != r.URL.String() {
+				t.Fatalf("Expected URL for [%s]: %s, got %s", test.expression, test.expectedURL, r.URL.String())
+			}
+			t.Logf("Received request to: %s", r.URL.String())
+		}))
+		serverRoute.route.GetHandler().ServeHTTP(nil, request)
+	}
+}
+
+func TestServerLoadConfigMultipleFrontendRules(t *testing.T) {
```
timoreimann: Instead of duplicating lots of code from `TestServerLoadConfigHealthCheckOptions`, can we try to handle both test functions in a single test with separate test cases?

Let me know if you need help.

timoreimann:
@aantono we'd like this fix to still make it into the 1.3 release. Could you please rebase your PR onto the v1.3 branch?

Since some people tend to struggle with rebasing, here's a quick how-to if a simple `git rebase` does not work:

1. Create a branch from `upstream/v1.3` locally.
1. Put your commit on top of that branch. (One easy way to do this is to check out the new branch and do `git cherry-pick <commit from your feature branch>`; another one is to check out the existing feature branch and do `git rebase --onto <name of the new 1.3 branch> <commit from your feature branch>^`. All of this assumes you still have a single commit only.)
1. Force-push the changes to your repo.
1. Click `Edit` at the top of this Github page and change the base to v1.3.

I can also do it for you if you like. Just let me know.

server/server_test.go
aantono: Have no idea why the names are what they are, that was an existing implementation for current configuration, I just used them in the tests to verify various combinations.  Those are all configured in `rules.go:135`, I didn't make any changes to those.

server/server_test.go
aantono: I'll remove this test, as I just realized that it is actually testing the same thing as `TestServerMultipleFrontendRules` does, so no need to duplicate.

aantono:
Will rebase and force push as soon as tests finish running locally.  Will you cherry-pick the changes back into master, or do you want me to create a separate PR for that?

timoreimann:
We'll merge the v1.3 branch back into master once this PR (and one or two others currently in the making) have landed. So you don't need to worry about getting things into master, I'll handle that. ☺️

aantono:
Sounds good, just pushed the updates, please let me know if that looks as what you expect.

aantono:
Also took a stab at adding documentation section about Rules Order, so let me know if that is what you were looking for as well.

docs/basics.md
```diff
@@ -191,6 +191,10 @@ backend = "backend2"
   rule = "Path:/test1,/test2"
 ```
 
+### Rules Order
+When combining `Modifier` rules with `Matcher` rules, it is important to remember that `Modifier` rules **ALWAYS** apply after the `Matcher` rules.  Some `Path*Strip*` rules, like `PathPrefixStripRegex`, or `PathStrip` for example, are both `Matchers` and `Modifiers`, so in this case the `Matcher` portion of the rule will apply first, and the `Modifier` will apply later.
```
ldez: could you add a white line between title and text?

middlewares/addPrefix.go
ldez: @aantono we think it's not necessary and this can be a problem in some case (If you use some logs tools (like Logstash, Graylog,...) you possibly want to always use debug in production).

Could you remove? 😃

middlewares/replace_path.go
```diff
@@ -1,6 +1,7 @@
 package middlewares
 
 import (
+	"github.com/containous/traefik/log"
```
ldez: Can you add the non-stdlib imports after the stdlib ones, separated by a blank line?

middlewares/stripPrefix.go
```diff
@@ -1,6 +1,7 @@
 package middlewares
 
 import (
+	"github.com/containous/traefik/log"
```
ldez: Can you add the non-stdlib imports after the stdlib ones, separated by a blank line?

server/server_test.go
timoreimann: Wow, you are totally right -- we do offer `PathPrefixStripRegex`!

I must have missed it as it seemingly only entered the primary documentation when we released 1.3.0-rc1.

So all good. 👍

server/server_test.go
timoreimann: You forgot `test := test`. :-)

testhelpers/helpers.go
```diff
@@ -1,6 +1,20 @@
...
+	"net/http"
+)
+
 // Intp returns a pointer to the given integer value.
 func Intp(i int) *int {
 	return &i
 }
+
+// MustNewRequest creates a new http get request or panics if it can't
+func MustNewRequest(rawurl string) *http.Request {
```
timoreimann: The method should match the parameters of `http.NewRequest`, i.e., accept a method and body parameter that would then be passed through.

middlewares/replace_path.go
```diff
@@ -18,3 +18,8 @@ func (s *ReplacePath) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	r.URL.Path = s.Path
 	s.Handler.ServeHTTP(w, r)
 }
+
+// SetHandler sets handler
+func (s *ReplacePath) SetHandler(Handler http.Handler) {
```
timoreimann: This isn't used anywhere, or is it?

server/server_test.go
```diff
@@ -27,6 +30,85 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+
+			if !routeMatch {
+				t.Fatalf("Rule %s don't match", expression)
+			}
+
+			server := new(Server)
+
+			server.wireFrontendBackend(serverRoute, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r.URL.String() != test.expectedURL {
+					t.Fatalf("Expected URL for [%s]: %s, got %s", test.expression, test.expectedURL, r.URL.String())
```
timoreimann: Please revert the reporting order (got first, expected second) here too.

server/server_test.go
```diff
@@ -27,6 +30,85 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+				}
+				t.Logf("Received request to: %s", r.URL.String())
```
timoreimann: IMHO we don't need this anymore now that we use sub-tests.

middlewares/replace_path.go
aantono: To be honest, I'm not sure if this is being used or not, just saw it missing from this rule and present in `addPrefix.go` and `stripPrefix.go`, so added it here for consistency.  If you think it's not needed, I'll be glad to remove it. (love deleting code!!!)

timoreimann: It should not be in use since you've added it only but there's no call happening.

If I'm wrong, the CI will tell us. :-)

aantono: I'll go ahead and remove it.  Does that mean that it is not really needed in the other 2 places also? (`addPrefix.go` and `stripPrefix.go`)

timoreimann: That's my thinking too.

aantono: I've updated the code to remove it from this particular place, so hopefully it's good to go. ;)

server/server_test.go
```diff
@@ -27,6 +30,85 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+
+			if err != nil {
+				t.Fatalf("Error while building route for %s: %+v", expression, err)
+			}
+
+			request := testhelpers.MustNewRequest(http.MethodGet, test.requestURL, nil)
+			routeMatch := routeResult.Match(request, &mux.RouteMatch{Route: routeResult})
+
+			if !routeMatch {
+				t.Fatalf("Rule %s don't match", expression)
```
timoreimann: Grammar: don't -> doesn't.

middlewares/stripPrefixRegex.go
timoreimann: I think we can revert the entire change list block to its original state?

server/server_test.go
```diff
@@ -27,6 +30,85 @@ func (lb *testLoadBalancer) Servers() []*url.URL {
...
+			}
+
+			server := new(Server)
+
+			server.wireFrontendBackend(serverRoute, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r.URL.String() != test.expectedURL {
+					t.Fatalf("Expected URL for [%s]: got %s, expected %s", test.expression, r.URL.String(), test.expectedURL)
```
timoreimann: We don't need to mention the expression since the sub test description already includes it and adds it automatically. So let's remove everything before and including the colon.

Let's just say `got URL %s, expected %s`.

aantono:
I see this is still marked as `waiting for corrections`, is there anything else that I've missed?

aantono:
I've rebased the PR against the trunk of v1.3 branch to make it mergable...

PR #1579, opened 2017-05-10
Project-wide Editorconfig

Stibbons:
Move on top of project to hit all directories.
Go files seems to prefer tab indentation over 2-space indent.


ldez:
Thanks for contribution.

Add an EditorConfig for go files doesn't make sense to us.

The Go tooling provide an internal formatter and the compiler throw an error if the files don't use tabulation.

We decline this PR for that.

PS: I like EditorConfig and I always use EditorConfig but not with Go.

PR #1548, opened 2017-05-04
Merge v1.3 branch into master [2017-05-11]

timoreimann:
Brings #1581 into mainline.

PR #1581, opened 2017-05-10
[Kubernetes] Ignore missing pass host header annotation. [v1.3 - CHERRY-PICK]

timoreimann:
We are starting to see reports of people using 1.3-rc1 and complaining/wondering about the problem reported in #1548. Hence, I think we should include the fix in the 1.3 branch as well.


timoreimann:
@emilevauge LGTY?

PR #1582, opened 2017-05-10
doc: add labels documentation.

ldez:
### Description

Add an explanation for each label we use.

https://mensuel.framapad.org/p/EYZcWzVbgO

### DIFF
--- a/.github/MAINTAINER.md
+++ b/.github/MAINTAINER.md
@@ -1,0 +1,85 @@
+# Maintainers
+
+## Labels
+
+If we open/look an issue/PR, we must add a `king/*` and an `area/*`.
+
+### Contributor
+
+* `contributor/need-more-information`: we need more information from the contributor in order to analyze a problem.
+* `contributor/waiting-for-corrections`: we need the contributor to take actions in order to move forward with a PR. **(only for PR)**
+* `contributor/needs-rebase`: use it only when there is some conflicts (and an automatic rebase is not possible). **(only for PR)** _[bot, humans]_
+
+### Kind
+
+* `kind/enhancement`: a new or improved feature.
+* `kind/question`: It's a question. **(only for issue)**
+* `kind/proposal`: proposal PR/issues need a public debate.
+  * _Proposal issues_ are design proposal that need to be refined with multiple contributors.
+  * _Proposal PRs_ are technical prototypes that need to be refined with multiple contributors.
+
+* `kind/bug/possible`: if we need to analyze to understand if it's a bug or not. **(only for issues)** _[bot only]_
+* `kind/bug/confirmed`: we are sure, it's a bug. **(only for issues)**
+* `kind/bug/fix`: it's a bug fix. **(only for PR)**
+
+### Resolution
+
+* `resolution/duplicate`: it's a duplicate issue/PR.
+* `resolution/declined`: Rule #1 of open-source: no is temporary, yes is forever.
+* `WIP`: Work In Progress. **(only for PR)**
+
+### Platform
+
+* `platform/windows`: Windows related.
+
+### Area
+
+* `area/provider`: related to all providers.
+* `area/provider/boltdb`: Boltd DB related.
+* `area/provider/consul`: Consul related.
+* `area/provider/docker`: Docker and Swarm related.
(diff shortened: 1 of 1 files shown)


.github/MAINTAINER.adoc
```diff
@@ -0,0 +1,79 @@
...
+area/api:: Traefik API related.
+area/logs:: Traefik logs related.
+area/sticky-session:: Sticky session related.
+area/websocket:: WebSocket related.
+area/webui:: Web UI related.
+area/infrastructure:: related to CI or Traefik building scripts.
+area/documentation:: regards improving/adding documentation
+
+priority/P0:: needs hot fix. (only for issue)
+priority/P1:: fixed in next release. (only for issue)
```
timoreimann: `fixed` sounds like this is only meant for bugs. Should we say `shipping` instead?

(Same for P2.)

.github/MAINTAINER.adoc
```diff
@@ -0,0 +1,79 @@
...
+priority/P2:: fixed in the future. (only for issue)
+priority/P3:: maybe. (only for issue)
```
timoreimann: s/maybe/backlog/?

.github/MAINTAINER.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+bug/possible:: (only for issues) if we need to analyze to understand if it's a bug or not. [bot only]
+bug/confirmed:: (only for issues) we are sure, it's a bug.
+bug/fix:: (only for PR) it's a bug fix.
+
+contributor/need-more-information:: we need more information from the contributor in order to analyze a problem.
+contributor/waiting-for-corrections:: (only for PR) we need the contributor to take actions in order to move forward with a PR
+contributor/needs-rebase:: (only for PR) use it only when there is some conflicts (and an automatic rebase is not possible) [bot, humans]
+
+duplicate:: it's a duplicate issue/PR.
```
timoreimann: Let's move `duplicate` next to `declined`.

I also suggest to prefix it with with `resolution/`.

.github/MAINTAINER.adoc
```diff
@@ -0,0 +1,79 @@
...
+
+enhancement:: a new or improved feature
```
timoreimann: Can we prefix `enhancement`, `question`, and `proposal` with `kind/` (similar to how Docker and k8s do it)?

ldez:
Asciidoc have better render for everything: table, list, image, video, ...

.github/MAINTAINER.adoc
ldez: Or `need to be fixed in ...`

emilevauge:
@ldez 
>Asciidoc have better render for everything: table, list, image, video, ...

We all agree on this. But I think that contributors may be unwilling to use/learn Asciidoc instead of the simple/widespread Markdown. I think that the documentation should be the easiest possible to modify by contributors (even if Asciidoc is better and not that complex, Markdown is simpler).

.github/MAINTAINER.adoc
```diff
@@ -0,0 +1,79 @@
...
+* Proposal issues are design proposal that need to be refined with multiple contributors.
+* Proposal PRs are technical prototypes that need to be refined with multiple contributors.
+
+declined:: Rule #1 of open-source: no is temporary, yes is forever.
+
+WIP:: (only for PR) Work In Progress.
+
+platform/windows:: Windows related.
+
+area/provider/boltdb:: Boltd DB related.
```
vdemeester: We also should have a `area/provider` one (something that would affect *all* the providers for example)

ldez:
You can use Markdown syntax in a Asciidoc document.

emilevauge:
@ldez I don't think we want/need to change the documentation format right now. And we really don't want to have 2 format at the same time.

timoreimann:
Right now, the `bug/*` labels are the only ones that stand out. I think we should also group them under the `kind/` namespace, e.g., `kind/bug/possible`, `kind/bug/confirmed`, and `kind/bug/fix`. This would be consistent IMO.

ldez:
`kind/quality` it's still in discus for me.

PR #1585, opened 2017-05-10
Maintain sticky flag on LB method validation failure.

### LINKED ISSUE #1568: Enabled sticky sessions flag not effective when load-balancing method is missing or invalid.
Hello!

The documentation on stickiness on https://docs.traefik.io/basics/ says:
```
Sticky sessions are supported with both load balancers. When sticky sessions are enabled, a cookie called _TRAEFIK_BACKEND is set on the initial request. On subsequent requests, the client will be directed to the backend stored in the cookie if it is still healthy. If not, a new backend will be assigned.
```
and I guess this is not correct anymore. If I leave the default (wrr) I don't receive a loadbalancing cookie in the browser when setting sticky = true. It is only working when method = "drr".

Also, but maybe I'am wrong, I thought that there was an indicator on the web ui which was indicating that stickiness is enabled?

traefik-web-sticky

Version:      v1.2.3
Codename:     morbier
Go version:   go1.7.5
Built:        2017-04-13_07:32:59PM
OS/Arch:      linux/amd64

timoreimann:
We previously did not copy the sticky flag if the load-balancer method validation failed, causing enabled stickiness to be dropped in case of a validation error (which, technically, for us is the same as a load-balancer configuration without an explicitly set method). This change fixes that.

A few refactorings and improvements along the way:

- Move the frontend and backend configuration steps into separate   methods/functions for better testability.
- Include the invalid method name in the error value and avoid log   duplication.
- Add tests for the backend configuration part.

Fixes #1568.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -255,7 +255,7 @@ func (server *Server) defaultConfigurationValues(configuration *types.Configurat
 		return
 	}
 	server.configureFrontends(configuration.Frontends)
-	configureBackends(configuration.Backends)
+	server.configureBackends(configuration.Backends)
 }
 
 func (server *Server) listenConfigurations(stop chan bool) {
@@ -889,7 +889,7 @@ func (server *Server) configureFrontends(frontends map[string]*types.Frontend) {
 	}
 }
 
-func configureBackends(backends map[string]*types.Backend) {
+func (*Server) configureBackends(backends map[string]*types.Backend) {
 	for backendName, backend := range backends {
 		_, err := types.NewLoadBalancerMethod(backend.LoadBalancer)
 		if err != nil {
--- a/server/server.go
+++ b/server/server.go
@@ -254,19 +254,8 @@ func (server *Server) defaultConfigurationValues(configuration *types.Configurat
 	if configuration == nil || configuration.Frontends == nil {
 		return
 	}
-	for _, frontend := range configuration.Frontends {
-		// default endpoints if not defined in frontends
-		if len(frontend.EntryPoints) == 0 {
-			frontend.EntryPoints = server.globalConfiguration.DefaultEntryPoints
-		}
-	}
-	for backendName, backend := range configuration.Backends {
-		_, err := types.NewLoadBalancerMethod(backend.LoadBalancer)
-		if err != nil {
-			log.Debugf("Load balancer method '%+v' for backend %s: %v. Using default wrr.", backend.LoadBalancer, backendName, err)
-			backend.LoadBalancer = &types.LoadBalancer{Method: "wrr"}
-		}
-	}
+	server.configureFrontends(configuration.Frontends)
+	configureBackends(configuration.Backends)
 }
 
 func (server *Server) listenConfigurations(stop chan bool) {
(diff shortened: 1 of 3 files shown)


timoreimann:
@containous/traefik FYI.

/cc @ kleinsasserm.

server/server.go
```diff
@@ -887,3 +876,25 @@ func sortedFrontendNamesForConfig(configuration *types.Configuration) []string {
...
+}
+
+func configureBackends(backends map[string]*types.Backend) {
+	for backendName, backend := range backends {
+		_, err := types.NewLoadBalancerMethod(backend.LoadBalancer)
+		if err != nil {
+			log.Debugf("Validation of load balancer method for backend %s failed: %s. Using default method wrr.", backendName, err)
+			backend.LoadBalancer = &types.LoadBalancer{
+				Method: "wrr",
+				Sticky: backend.LoadBalancer.Sticky,
```
timoreimann: Effectively, this should be the only addition to the production code.

timoreimann:
Integration tests keep failing. Not exactly sure why, will have to investigate.

server/server.go
```diff
@@ -887,3 +876,25 @@ func sortedFrontendNamesForConfig(configuration *types.Configuration) []string {
...
+func (server *Server) configureFrontends(frontends map[string]*types.Frontend) {
+	for _, frontend := range frontends {
+		// default endpoints if not defined in frontends
+		if len(frontend.EntryPoints) == 0 {
+			frontend.EntryPoints = server.globalConfiguration.DefaultEntryPoints
+		}
+	}
+}
+
+func configureBackends(backends map[string]*types.Backend) {
```
ldez: why not `func (server *Server) configureBackends`?

timoreimann: Because the backend configuration requires nothing from `server` for now.

emilevauge: I also find this a bit odd to be honest ^^

timoreimann: I can turn it into a receiver method if you really want to, but it would carry the wrong message to the reader (and my IDE's "find usage" function) that `server` is somehow needed for this method to work properly, which it isn't. It also makes things like testing more complicated.

Effectively, it'd be similar to passing an unused variable to a function call.

emilevauge: @timoreimann I think there is at least 2 reasons to add `func (server *Server)` to both functions:

1. it's homogeneous (to the reader, it would be easier to understand if `configureFrontends` and `configureBackends` add the same signature)
2. we could only call `configureBackends` if we had a reference to a `Server` which is makes more sense

timoreimann: Okay, converted `configureBackends` into a method.

PTAL again.

server/server.go
```diff
@@ -887,3 +876,29 @@ func sortedFrontendNamesForConfig(configuration *types.Configuration) []string {
...
+func (server *Server) configureFrontends(frontends map[string]*types.Frontend) {
+	for _, frontend := range frontends {
+		// default endpoints if not defined in frontends
+		if len(frontend.EntryPoints) == 0 {
+			frontend.EntryPoints = server.globalConfiguration.DefaultEntryPoints
+		}
+	}
+}
+
+func (*Server) configureBackends(backends map[string]*types.Backend) {
```
emilevauge: :heart:

PR #1284, opened 2017-03-13
[WIP] Deflake integration tests.

timoreimann:
This PR refactors and improves the `integration/utils/Try*` functions and replaces a number of hard-coded `time.Sleep()` calls in our integration test fixtures by retry-based timeouts. We also stop backing off exponentially since there's little value in doing so during integration tests where we want feedback as quickly as possible.

Refs #1266.

TODO:

- [ ] Run repeatedly to see if we have caught all flaky tests.
- [ ] Look at the remaining `time.Sleep` calls and see if we can replace them by retries.

PR #1599, opened 2017-05-13
Deflake integration tests

ldez:
### Description

Base on  #1284


integration/utils/try.go
```diff
@@ -1,50 +1,209 @@
...
+	if timeout <= 0 {
+		panic("timeout must be larger than zero")
+	}
+
+	interval := time.Duration(math.Ceil(float64(timeout) / 15.0))
+	if interval > maxInterval {
+		interval = maxInterval
+	}
+
+	ci := os.Getenv("CI")
```
vdemeester: I'm really not a huge fan of that. We shouldn't have a *special case* for CI. If we need a `3x` multiplier for timeout in CI, let's just set all the timeout to `3x` their value..

integration/utils/try.go
```diff
@@ -1,50 +1,209 @@
...
+				fmt.Println("+")
+				return nil
+			}
+		}
+	}
+}
+
+// Sleep pauses the current goroutine for at least the duration d.
+// Use only when use an other Try[...] functions is not possible.
+func Sleep(d time.Duration) {
```
vdemeester: If we don't have a special case for CI, this doesn't make sense. We should be able to replace any `time.Sleep` from the tests code with a `Try` or sthg that is better than a plain old `Sleep`

integration/utils/try.go
```diff
@@ -1,50 +1,209 @@
...
+func TryGetResponse(url string, timeout time.Duration) (*http.Response, error) {
+	return doGetResponse(url, timeout, nil)
+}
+
+// TryGetResponseUntilStatusCode is like TryGetRequest, but returns the response for further
+// processing at the call site.
+// Conditions are not allowed since it would complicate signaling if the
+// response body needs to be closed or not. Callers are expected to close on
+// their own if the function returns a nil error.
+func TryGetResponseUntilStatusCode(url string, timeout time.Duration, statusCode int) (*http.Response, error) {
```
vdemeester: This one is not used elsewhere

integration/utils/try.go
```diff
@@ -1,50 +1,209 @@
...
 	return func(res *http.Response) error {
 		if res.StatusCode != status {
-			return errors.New("Bad status. Got: " + res.Status + ", expected:" + strconv.Itoa(status))
+			return fmt.Errorf("got status code %d, wanted %d", res.StatusCode, status)
 		}
 		return nil
 	}
 }
+
+func doGetResponse(url string, timeout time.Duration, condition Condition) (*http.Response, error) {
```
vdemeester: This one is not required, let's just pass the http `method` to `doResponse` and use only `doRespose`

integration/consul_catalog_test.go
```diff
@@ -98,16 +115,11 @@ func (s *ConsulCatalogSuite) TestSingleService(c *check.C) {
...
 	c.Assert(err, checker.IsNil)
 	req.Host = "test.consul.localhost"
-	resp, err := client.Do(req)
-
-	c.Assert(err, checker.IsNil)
-	c.Assert(resp.StatusCode, checker.Equals, 200)
 
-	_, err = ioutil.ReadAll(resp.Body)
+	cond := utils.ComposeCondition(utils.UntilStatusCodeIs(200), utils.BodyContains("Welcome to nginx!"))
+	err = utils.TryRequest(req, 5*time.Second, cond)
```
vdemeester: I would have make `TryRequest` taking *variadic args*, so that you could do `TryRequest(req, 5*time.Second, utils.UntilStatusCodeIs…, utils.Body…)`.

integration/utils/try.go
```diff
@@ -1,50 +1,209 @@
...
+		log.Println("Activate CI multiplier:", CITimeoutMultiplier)
+		timeout = time.Duration(float64(timeout) * CITimeoutMultiplier)
+	}
+
+	var err error
+	if err = operation(); err == nil {
+		return nil
+	}
+
+	stopTime := time.Now().Add(timeout)
```
ldez: Your code doesn't work -> infinite timeout when the operation failed.

```go
for {
	select {
	case <-time.After(timeout):
		return fmt.Errorf("try operation failed: %s", err)
	case <-time.Tick(wait):
		if err = operation(); err == nil {
			return nil
		}
	}
}
```
-> infinite timeout when the operation failed.

```go
stopTimer := time.NewTimer(timeout)
retryTick := time.NewTicker(interval)

for {
  select {
    case <-stopTimer.C:
      err = fmt.Errorf("try operation failed: %s", err)
      stopTimer.Stop()
      retryTick.Stop()
      return err
    case <-retryTick.C:
      if err = operation(); err == nil {
        stopTimer.Stop()
        retryTick.Stop()
        return err
     }
  }
}
```
-> works but not very beautiful.

integration/utils/try.go
ldez: Currently, in some case we really need to use `time.Sleep`, the `Try` logical cannot solve all case.
We need a second way with a stuff like: "Retry again for X time"

integration/utils/try.go
ldez: Currently, this doesn't work because we don't access to global env var in the integration test.

But why I have add this special case?
I don't want to wait 3 min in my local environment to known if a test fail, but in the CI some tests are more slow than a personal computer, we need to have a pragmatic approach.

integration/utils/try.go
vdemeester: > Your code doesn't work -> infinite timeout when the operation failed.

Why would it ? at some point (`timeout`) `stopTime.C` will get the message and we get out of the infinit loop.

integration/utils/try.go
vdemeester: > Currently, this doesn't work because we don't access to global env var in the integration test.

hum.. 🤔 we have access to whichever the environment variable used to execute the tests, so we just need to pass them.

> But why I have add this special case?
> I don't want to wait 3 min in my local environment to known if a test fail, but in the CI some tests are more slow than a personal computer, we need to have a pragmatic approach.

hum, I don't really buy that but *meh* I won't fight again, it's just timeout stuff

integration/utils/try.go
vdemeester: Loads of cases should/can be handled with `Try` or something that acting almost the same.

If we `Sleep` : 
-  when *waiting for the process to run*, we should use `Try` with a condition (i.e. something that checks the process is running, or some port is open, or for some output of the command to appear, or ?)
- when *waiting for a k/v value to be there*, same, we should put the get into a condition

The less we use `Sleep`, the better really 👼

integration/healthcheck_test.go
```diff
@@ -67,16 +55,40 @@ func (s *HealthCheckSuite) TestSimpleConfiguration(c *check.C) {
...
+	_, err = client.Do(healthReq1)
 	c.Assert(err, checker.IsNil)
 
-	time.Sleep(time.Second * 3)
+	// Verify frontend health : before
+	err = utils.TryRequest(req, 3*time.Second, utils.UntilStatusCodeIs(500))
+	c.Assert(err, checker.IsNil)
 
-	resp, err = client.Do(req)
+	utils.Sleep(3 * time.Second)
```
vdemeester: I'm not sure about that one, at all. What are we testing here : that the health was `500` before (it's tested above) and that in a given period of time, it will be `200`. I don't see why we re-verify `500`, wait for 3s and then wait for `200`.

integration/healthcheck_test.go
```diff
@@ -67,16 +55,40 @@ func (s *HealthCheckSuite) TestSimpleConfiguration(c *check.C) {
...
+	// Fix all whoami health to 500
+	whoamiHosts := []string{whoami1Host, whoami2Host}
+	for _, whoami := range whoamiHosts {
+		healthReq, err := http.NewRequest("POST", "http://"+whoami+"/health", bytes.NewBuffer([]byte("500")))
+		c.Assert(err, checker.IsNil)
+		_, err = client.Do(healthReq)
+		c.Assert(err, checker.IsNil)
+	}
+
+	utils.Sleep(3 * time.Second)
```
vdemeester: Almost the same here, if we want to wait 3s and then have the `TryRequest`, let's just set a bigger timeout for the `Try`.

integration/healthcheck_test.go
ldez: we need to wait the time to do the check of the "healthcheck"

integration/utils/try.go
ldez: I don't know exactly why but it produce an infinite loop.
it produce 2 loops: one stop at the timeout and one infinite.

integration/utils/try.go
ldez: https://github.com/containous/traefik/blob/master/integration/marathon_test.go#L39

This case need a `continue` approach, `try` doesn't work in this case.

PS: this test file seems to test nothing, because all request seems always give the same response, look at the comment block (it's impossible to make this work).

ldez: you can also look at https://github.com/containous/traefik/blob/master/integration/docker_test.go#L180

ldez: the first case (https://github.com/containous/traefik/blob/master/integration/marathon_test.go#L39) can handled with  a `Try`

integration/utils/try.go
atbore-phx: > Currently, this doesn't work because we don't access to global env var in the integration test.

> hum.. :thinking: we have access to whichever the environment variable used to execute the tests, so we just need to pass them.

in our repo via .semaphoreci/vars we can pass global vars to the CI, but in this case the issue is the overlay docker level we add when we want execute the tests so the the CI global vars need to be passed in the make file like that: https://github.com/atbore-phx/traefik/commit/6aef257b5f00231854ab564305796ee8db78500a

integration/utils/try.go
vdemeester: > you can also look at https://github.com/containous/traefik/blob/master/integration/docker_test.go#L180

Yes, this one could/should use `TryRequest` (with the condition on getting a `404` or sthg)

integration/utils/try.go
atbore-phx: > I'm really not a huge fan of that. We shouldn't have a special case for CI. If we need a 3x multiplier for timeout in CI, let's just set all the timeout to 3x their value..

I'm agree "in theory" with this sentence if we are speaking about "timeout" stuff. 

but i fell, I'm missing something, why do we wait 3 min more, locally ? the value is a timeout so a max "break" value. isn't it ?

atbore-phx:
when you think this PR reaches a steady state, just ping me and i could execute a series on the CI and verify if the Error Probability is decreasing like that: https://github.com/containous/traefik/issues/1600. 
IMHO: it could be a nice test to confirm if we are on the right direction before merge.
@ldez, @timoreimann  it's a very nice work and improvement BTW.

ldez:
> - Finaly, as it's a *testing* package, we could even make the `Try` function take a `t *testing.T` as first argument, and calling `t.Fatal` in case of error 👼

It's the first step, I will create the *testing* package after.

integration/https_test.go
```diff
@@ -61,7 +62,7 @@ func (s *HTTPSSuite) TestWithSNIConfigRoute(c *check.C) {
 	defer backend1.Close()
 	defer backend2.Close()
 
-	time.Sleep(2000 * time.Millisecond)
+	try.Sleep(2000 * time.Millisecond)
```
timoreimann: IIUC then the sleep here are only necessary in order to wait for the test server to be up and running.

Why not replace this by `try.Do()`s that attempt to TCP-connect to each server?

Same for the other sleeps in this file.

integration/healthcheck_test.go
```diff
@@ -67,16 +55,40 @@ func (s *HealthCheckSuite) TestSimpleConfiguration(c *check.C) {
...
+	// Fix all whoami health to 500
+	whoamiHosts := []string{whoami1Host, whoami2Host}
+	for _, whoami := range whoamiHosts {
+		healthReq, err := http.NewRequest("POST", "http://"+whoami+"/health", bytes.NewBuffer([]byte("500")))
+		c.Assert(err, checker.IsNil)
+		_, err = client.Do(healthReq)
+		c.Assert(err, checker.IsNil)
+	}
+
+	try.Sleep(2 * time.Second)
```
timoreimann: This should be replaceable by continuous GET queries `whoami`'s `/health` endpoint until it returns 500.

After that...

integration/healthcheck_test.go
```diff
@@ -67,16 +55,40 @@ func (s *HealthCheckSuite) TestSimpleConfiguration(c *check.C) {
...
+
+	// Change one whoami health to 200
+	healthReq1, err := http.NewRequest("POST", "http://"+whoami1Host+"/health", bytes.NewBuffer([]byte("200")))
+	c.Assert(err, checker.IsNil)
+	_, err = client.Do(healthReq1)
 	c.Assert(err, checker.IsNil)
 
-	time.Sleep(time.Second * 3)
+	// Verify frontend health : before
+	err = try.Request(req, 3*time.Second, try.StatusCodeIs(500))
```
timoreimann: I don't think we need this.

integration/healthcheck_test.go
```diff
@@ -67,16 +55,40 @@ func (s *HealthCheckSuite) TestSimpleConfiguration(c *check.C) {
...
+	c.Assert(err, checker.IsNil)
 
-	resp, err = client.Do(req)
+	try.Sleep(2 * time.Second)
```
timoreimann: I think we want another query check at this point rather making sure `whoami1Host`'s health check returns 200 again.

integration/healthcheck_test.go
```diff
@@ -67,16 +55,40 @@ func (s *HealthCheckSuite) TestSimpleConfiguration(c *check.C) {
...
+		healthReq, err := http.NewRequest("POST", "http://"+whoami+"/health", bytes.NewBuffer([]byte("500")))
+		c.Assert(err, checker.IsNil)
+		_, err = client.Do(healthReq)
+		c.Assert(err, checker.IsNil)
+	}
+
+	try.Sleep(2 * time.Second)
+
+	// Verify frontend health : 500
+	err = try.Request(req, 3*time.Second, try.StatusCodeIs(500))
```
timoreimann: ...this should not be needed anymore.

integration/healthcheck_test.go
```diff
@@ -67,16 +55,40 @@ func (s *HealthCheckSuite) TestSimpleConfiguration(c *check.C) {
...
 
-	time.Sleep(time.Second * 3)
+	// Verify frontend health : before
+	err = try.Request(req, 3*time.Second, try.StatusCodeIs(500))
+	c.Assert(err, checker.IsNil)
 
-	resp, err = client.Do(req)
+	try.Sleep(2 * time.Second)
+
+	// Verify frontend health : after
```
timoreimann: I think the block starting from here to line 91 (incl.) isn't needed as we just want the single, static request starting at line 93.

integration/https_test.go
```diff
@@ -25,7 +26,7 @@ func (s *HTTPSSuite) TestWithSNIConfigHandshake(c *check.C) {
 	c.Assert(err, checker.IsNil)
 	defer cmd.Process.Kill()
 
-	time.Sleep(500 * time.Millisecond)
+	try.Sleep(500 * time.Millisecond)
```
timoreimann: AFAICS, this should be the only `try.Sleep` call left after my comments regarding the other ones are addressed. What is this one waiting for? Isn't this just another "wait for Traefik to start" check that we tend to solve via a retriable request?

integration/healthcheck_test.go
```diff
@@ -42,45 +37,54 @@ func (s *HealthCheckSuite) TestSimpleConfiguration(c *check.C) {
...
+	whoamiHosts := []string{whoami1Host, whoami2Host}
+	for _, whoami := range whoamiHosts {
+		healthReq, err := http.NewRequest(http.MethodPost, "http://"+whoami+"/health", bytes.NewBuffer([]byte("500")))
+		c.Assert(err, checker.IsNil)
+		_, err = client.Do(healthReq)
+		c.Assert(err, checker.IsNil)
+	}
+
+	// Waiting for Traefik healthcheck
+	try.Sleep(2 * time.Second)
```
timoreimann: We should be able to drop this one if we just increase the timeout in line 65 a bit.

integration/healthcheck_test.go
```diff
@@ -42,45 +37,54 @@ func (s *HealthCheckSuite) TestSimpleConfiguration(c *check.C) {
...
+	healthReq1, err := http.NewRequest(http.MethodPost, "http://"+whoami1Host+"/health", bytes.NewBuffer([]byte("200")))
 	c.Assert(err, checker.IsNil)
-	_, err = client.Do(healthReq)
+	_, err = client.Do(healthReq1)
 	c.Assert(err, checker.IsNil)
 
-	time.Sleep(time.Second * 3)
-
-	resp, err = client.Do(req)
+	// Verify frontend health : before
```
timoreimann: I think the remaining block up until line 89 can be simplified as following:

1. Try-request for a 200 response code. Once we succeed, we know for sure that the first whoami host has successfully toggled its health status back to 200. Also, this way we don't have to rely on the `try.Sleep` in line 79.
1. Try-request for a 500 response code. This verifies that the second whoami host is still unhealthy.

Paraphrased, we can remove the section between line 74 and line 80.

integration/constraint_test.go
```diff
@@ -35,7 +36,16 @@ func (s *ConstraintSuite) SetUpSuite(c *check.C) {
...
+	err = try.Do(3*time.Second, func() error {
+		leader, err := consulClient.Status().Leader()
+
+		if err != nil || len(leader) == 0 {
+			return fmt.Errorf("Leader not find. %v", err)
+		}
+
+		return nil
+	})
+	c.Assert(err, checker.IsNil)
```
vdemeester: (Note that this is the reason why I think `try.Do(t, 3*time.Second, …)` would fit perfectly. But can be done in follow-ups.

integration/eureka_test.go
```diff
@@ -66,7 +63,7 @@ func (s *EurekaSuite) TestSimpleConfiguration(c *check.C) {
     }
 	}`
 
-	tmpl, err := template.New("eurekaTemlate").Parse(eurekaTemplate)
+	tmpl, err := template.New("eurekaTemplate").Parse(eurekaTemplate)
```
vdemeester: :scream:

integration/try/condition.go
```diff
@@ -0,0 +1,95 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/docker/libkv/store"
+)
+
+// ResponseCondition is a retry condition function.
+// It receives a response, and returns an error
+// if the response failed the condition.
+type ResponseCondition func(*http.Response) error
```
vdemeester: Nit: could be unexported (but `lint` or `vet` might be not happy, not sure)

integration/try/condition.go
```diff
@@ -0,0 +1,95 @@
...
+		if res.StatusCode != status {
+			return fmt.Errorf("got status code %d, wanted %d", res.StatusCode, status)
+		}
+		return nil
+	}
+}
+
+// DoCondition is a retry condition function.
+// It returns an error
+type DoCondition func() error
```
vdemeester: Nit (same as above): could be unexported (but `lint` or `vet` might be not happy, not sure)

integration/try/try.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	interval := time.Duration(math.Ceil(float64(timeout) / 15.0))
+	if interval > maxInterval {
+		interval = maxInterval
+	}
+
+	timeout = applyCIMultiplier(timeout)
+
+	var err error
```
vdemeester: yeah I don't think we need to declare it before hand.

integration/healthcheck_test.go
ldez: I deleted all sleep except this one and refactor the whole test.

integration/try/try.go
ldez: it's needed, look at lines below

PR #1601, opened 2017-05-15
Fix empty basic auth

### LINKED ISSUE #1592: Traefik crash when failing to parse Basic Authentication information
Version: `Traefik version v1.3.0-rc1 built on 2017-05-05_02:03:01PM`

### Scenario
Starting traefik in docker and trying to route to another docker container using `labels`.
If incorrect (or in the test case empty) authentication information is used in the container I try to route to - traefik outputs:
`level=fatal msg="Error creating Auth: Error parsing Authenticator user: "`
and crashes.

### Expected outcome
If traefik fails to parse authentication information it should (probably) ignore the frontend mapping and not add the routing to the container.

### What did you see instead?
Traefik crashes causing all routing to stop working.

### Steps to reproduce?
`docker-compose up`

With the following compose file:
```yml
version: '3'

services:
  webgateway:
    image: traefik:1.3.0-rc1-alpine
    command:
      - --docker
      - --web
      - --logLevel=DEBUG
    ports:
      - "8080:8080"
      - "8000:80"
    volumes:

  error:
    image: nginx:alpine
    labels:
      - traefik.port=80
      - traefik.frontend.auth.basic=
```

emilevauge:
### Description

This PR fixes traefik crash when basic auth user is empty.

Fixes #1592

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -655,7 +655,7 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 								log.Errorf("Skipping frontend %s...", frontendName)
 								continue frontend
 							}
-							hcOpts := parseHealthCheckOptions(rebalancer, frontend.Backend, configuration.Backends[frontend.Backend].HealthCheck, *globalConfiguration.HealthCheck)
+							hcOpts := parseHealthCheckOptions(rebalancer, frontend.Backend, configuration.Backends[frontend.Backend].HealthCheck, globalConfiguration.HealthCheck)
 							if hcOpts != nil {
 								log.Debugf("Setting up backend health check %s", *hcOpts)
 								backendsHealthcheck[frontend.Backend] = healthcheck.NewBackendHealthCheck(*hcOpts)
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -843,7 +844,7 @@ func (server *Server) buildDefaultHTTPRouter() *mux.Router {
 	return router
 }
 
-func parseHealthCheckOptions(lb healthcheck.LoadBalancer, backend string, hc *types.HealthCheck, hcConfig HealthCheckConfig) *healthcheck.Options {
+func parseHealthCheckOptions(lb healthcheck.LoadBalancer, backend string, hc *types.HealthCheck, hcConfig *HealthCheckConfig) *healthcheck.Options {
```
timoreimann: Do we guarantee on the call site that `hcConfig` is always != nil? Without that, we could run into a nil reference issue on line 852.

timoreimann:
Hmm, how's the health check configuration related to basic auth?

I must be missing a piece of the puzzle... 🤔

server/server.go
emilevauge: Oops, I was thinking the same and I just saw you too :) Fixed.

emilevauge:
@timoreimann You are absolutely right. I just fixed this because it failed while I was writing the new tests and I thought it would be better to remove the pointer dereferencing calling `parseHealthCheckOptions`. WDYT?

PR #1604, opened 2017-05-15
[k8s] Ignore Ingresses with empty Endpoint subsets.

### LINKED ISSUE #1462: Constant warning in Kubernetes
1.2.1

Kubernetes 1.6

### What did you do?
Deploy a service and ingress as per the documentation example.

Nothing above info level in the logs.

### What did you see instead?
Constant warning level logs of the form:
```
W  Service endpoints not found for REDACTED/REDACTED, falling back to Service ClusterIP 
W  Service endpoints not found for REDACTED/REDACTED, falling back to Service ClusterIP
W  Service endpoints not found for REDACTED/REDACTED, falling back to Service ClusterIP
W  Service endpoints not found for REDACTED/REDACTED, falling back to Service ClusterIP
W  Service endpoints not found for REDACTED/REDACTED, falling back to Service ClusterIP
```
One line roughly every second or so per service.

Seems to be caused by [this line](https://github.com/containous/traefik/blob/master/provider/kubernetes/kubernetes.go#L251). Given that `ClusterIP` is the default for a k8s Service plus it's what the documentation uses, does this need to be a warning? Perhaps lowering to debug or even removing would be appropriate.

timoreimann:
We previously fell back to using ClusterIPs. However, the approach can lead to all kinds of problems since Ingresses rely on being able to talk to Endpoints directly. For instance, it can break stickiness and retries.

I believe the issue should still go into 1.3 since the implications can be quite severe and users won't have much of a chance to spot the misconfigurations easily.

Fixes #1462.

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -252,28 +252,25 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 							}
 
 							if !exists {
-								log.Errorf("Endpoints not found for %s/%s", service.ObjectMeta.Namespace, service.ObjectMeta.Name)
-								continue
+								log.Warnf("Endpoints not found for %s/%s", service.ObjectMeta.Namespace, service.ObjectMeta.Name)
+								break
 							}
 
 							if len(endpoints.Subsets) == 0 {
-								log.Warnf("Service endpoints not found for %s/%s, falling back to Service ClusterIP", service.ObjectMeta.Namespace, service.ObjectMeta.Name)
-								templateObjects.Backends[r.Host+pa.Path].Servers[string(service.UID)] = types.Server{
-									URL:    protocol + "://" + service.Spec.ClusterIP + ":" + strconv.Itoa(int(port.Port)),
-									Weight: 1,
-								}
-							} else {
-								for _, subset := range endpoints.Subsets {
-									for _, address := range subset.Addresses {
-										url := protocol + "://" + address.IP + ":" + strconv.Itoa(endpointPortNumber(port, subset.Ports))
-										name := url
-										if address.TargetRef != nil && address.TargetRef.Name != "" {
-											name = address.TargetRef.Name
-										}
-										templateObjects.Backends[r.Host+pa.Path].Servers[name] = types.Server{
-											URL:    url,
-											Weight: 1,
-										}
+								log.Warnf("Endpoints not available for %s/%s", service.ObjectMeta.Namespace, service.ObjectMeta.Name)
+								break
+							}
+
+							for _, subset := range endpoints.Subsets {
+								for _, address := range subset.Addresses {
+									url := protocol + "://" + address.IP + ":" + strconv.Itoa(endpointPortNumber(port, subset.Ports))
+									name := url
+									if address.TargetRef != nil && address.TargetRef.Name != "" {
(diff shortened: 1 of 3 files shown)


dtomcej:
If there is an ingress, with no endpoints, shouldn't we create an empty backend to serve a 503?

timoreimann:
@dtomcej 

>If there is an ingress, with no endpoints, shouldn't we create an empty backend to serve a 503?

In that case, we [create a backend with no servers](https://github.com/timoreimann/traefik/blob/3260632c0f6781c8a53560eb5e1069ab64590a78/provider/kubernetes/kubernetes_test.go#L2097-L2103) configured. Do you think that won't be enough?

timoreimann:
Hmm, I'm having a hard time getting this to work as desired: I either get 500s or 404s, but never 502s.

Marking as WIP for the time being.

timoreimann:
@errm how did you verify the behavior you described in https://github.com/containous/traefik/issues/1077#issuecomment-281634260, especially this part:

>If the ingress and referenced service exist, but no pods exist (or they are not ready) Traefik will return a 502 Bad Gateway...

I have removed the Pods and even the Endpoints, but I never manage to get a 502; only ever a 500.

errm:
I have no idea, but I would expect that if I did get that to occur, it would have been as a result of whatever was going on with network timeouts... or perhaps I just misremembered and really I was just seeing 500 errors rather than 503...

provider/kubernetes/kubernetes.go
```diff
@@ -264,23 +264,20 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
-										}
-										templateObjects.Backends[r.Host+pa.Path].Servers[name] = types.Server{
-											URL:    url,
-											Weight: 1,
-										}
+								log.Debugf("Endpoints not available for %s/%s", service.ObjectMeta.Namespace, service.ObjectMeta.Name)
+								continue
+							}
+
+							for _, subset := range endpoints.Subsets {
```
dtomcej: This won't execute when `endpoints.Subsets` is empty, thats why there is no else?

timoreimann: As discussed on Slack, the `else` got replaced by the `continue` effectively, though the `else` block is still there.

timoreimann:
It took me a truckload of debug log messages sprinkled around the k8s provider, but I finally figured out why I would only be able to see 404s (as opposed to 50x) on missing/empty Endpoints: The k8s template relies on at least one configuration parameter to get rendered successfully in order to include the `[backends."{{$backendName}}"]` TOML table. Missing that as is the case when we don't have Endpoints, the backend will still be there in the resulting Go struct but empty, causing `server.loadConfig` to fail finding the backend associated to the frontend and finally dropping it. So the easy fix is to include the base table, which I did.

(I also changed the log level to WARNING (because the previous ERROR and DEBUG seemed too strict and loose, respectively), and used `break` instead of `continue` to not waste time iterating through the Service ports when we already had the right match.)

Now, the one thing left is the concrete 50x error code: Right now I'm only able to get a 500 when the backend is empty. This seems to be a matter out-of-scope of the k8s provider, however, and my tests with a file provider support my assumption. If we really want to make Treafik return 503s, I believe it should happen in a different PR.

@dtomcej WDYT on the current state of the PR and the 500/503 question?

PR #1628, opened 2017-05-18
Issue template: Emphasize SO and Slack for support questions.

timoreimann:
- Be more explicit in the purpose of the issue tracker.
- Move SO before Slack since it seems preferable.
- Refer to SO and Slack on first question again.

I also lean towards removing the big block of "how to write a good issue?". While I wholeheartedly agree with the content, I think it's just too much information overall and distracts from the other (IMHO more important) other information. But maybe that's to decide in another PR.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,12 +1,12 @@
 <!--
 PLEASE READ THIS MESSAGE.
 
-Please keep in mind that the GitHub issue tracker is not intended as a general support forum, but for reporting bugs and feature requests.
+=== DO *NOT* FILE ISSUES FOR GENERAL SUPPORT QUESTIONS. ===
 
-For other type of questions, consider using one of:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -21,6 +21,8 @@ HOW TO WRITE A GOOD ISSUE?
 
 ### Do you want to request a *feature* or report a *bug*?
 
+(If you intend to ask a support question: **DO NOT FILE AN ISSUE**. Use Stackoverflow or Slack instead.)
```
ldez:
```
(If you intend to ask a support question: **DO NOT FILE AN ISSUE**. Use [Stackoverflow](https://stackoverflow.com/questions/tagged/traefik) or [Slack](https://traefik.herokuapp.com) instead.)
```

timoreimann: Done.

PR #1635, opened 2017-05-19
Upgrade go-marathon to 15ea23e.

timoreimann:
Our vendored copy contains a bug that causes unavailable Marathon nodes to never be marked as available again due to a misconstruction in the URL to the Marathon health check / ping endpoint used by go-marathon internally.

A [fix](https://github.com/gambol99/go-marathon/pull/283) has been published.

Targeting 1.3 since the impact on a production deployment could be quite severe: For long-running Traefik instances, it only takes as many failed attempts to reach the Marathon API endpoint as there are Marathon nodes in the cluster for the Marathon provider to stop working properly.


ldez:
```
+ go test -cover -coverprofile=cover.out ./provider/marathon
# github.com/containous/traefik/mocks
mocks/Marathon.go:856: undefined: marathon.GetAppOpts
mocks/Marathon.go:866: undefined: marathon.Queue
mocks/Marathon.go:876: undefined: marathon.GetGroupOpts
mocks/Marathon.go:881: undefined: marathon.GetGroupOpts
FAIL	github.com/containous/traefik/provider/marathon [build failed]
```

timoreimann:
Interesting. I wonder why the mock fails now but did not do on the last go-marathon upgrade, which is ~2 commits behind and no API changes were introduced in the meantime AFAIR.

Will take a look. Thanks, @ldez!

vendor/github.com/gambol99/go-marathon/application.go
```diff
@@ -123,41 +95,6 @@ type Fetch struct {
 	Cache      bool   `json:"cache"`
 }
 
-// GetAppOpts contains a payload for Application method
-//		embed:	Embeds nested resources that match the supplied path.
-// 				You can specify this parameter multiple times with different values
-type GetAppOpts struct {
```
ldez: The problem is here

timoreimann:
Fixed, thanks for @ldez and `glide cache-clear`. :)

Issue #1123, opened 2017-02-06
Redirect path to domain.com/web is not working (Docker Swarm Mode 1.13)


pascalandy:
### What version of Traefik are you using (`traefik version`)?
traefik:v1.1.2-alpine

### What is your environment & configuration (arguments, toml...)?
https://gist.github.com/pascalandy/a28958d8794a6c142a11accb7623b817

### Intro
Hey folks!

As seen in the [concept section](https://docs.traefik.io/basics/), I want to create a path as:
> path domain.com/web will point the microservice web in your private network

*I can successfully redirect the path over `api.domain.com`. Here, I am strugling to redirect the path  over `domain.com/web`.*

### What did you do?

1) Here is how I create my **nginx** container:

```
CTN_NGX_TEST="nginxtest"
WWW_SRC="/to-mount/nginx-static/starter"
WWW_DST="/usr/share/nginx/html"

docker service create \
--name "$CTN_NGX_TEST" \
--network "$NTW_FRONT" \
--mount	type=bind,src="$WWW_SRC",target="$WWW_DST" \
--label traefik.enable="true" \
--label traefik.frontend.rule="Host:example.com;PathPrefixStrip:/nginxtest" \
--label traefik.frontend.entryPoints="http" \
--label traefik.docker.network="$NTW_FRONT" \
--label traefik.backend="$CTN_NGX_TEST" \
--label traefik.backend.loadbalancer.sticky="true" \
--label traefik.weight="10" \
--label traefik.frontend.priority="10" \
--label traefik.backend.loadbalancer.method="drr" \
--label traefik.port="80" \
nginx
```

2) Here is how I created my **Traefik** container.
http://pascalandy.com/blog/gist-2017-02-04_12h18

### The Issue
[See the screenshot](http://blog.pascalandy.com/gist-2017-02-04_12h18/). From this page, if I click on the link about, the browser points to to: `example.com/about/about-1.html`. Can you see the issue??

### What did you expect to see?
`example.com/nginxtest/about/about-1.html`

### What did you see instead?
`example.com/about/about-1.html`

**Interesting**: If I manually enter `example.com/nginxtest/index.html` in my address bar, I can surf normally within the website! 

So what is wrong ??

Thank you in advance :)

pascalandy:
My @emilevauge ,
After searching/trying for hours, I found that manu issues are related.

[...]

Issue #1106, opened 2017-02-02
Docker swarm and traefik always error 404


juanpaps03:
### What version of Traefik are you using (`traefik version`)?
latest version on docker hub at the date

### What is your environment & configuration (arguments, toml...)?
I have 2 VM running with vagrants, both machines has docker swarm, one as a manager and another one as node, and I created a network with the next command:

    docker network create --driver overlay --subnet 10.0.9.0/24 my-network

### What did you do?

I add a service with traefik and a service with my own app (a simple nginx app), here I let the commands:

    docker service create \
         --name traefik \
         --constraint=node.role==manager \
         --publish 80:80 --publish 8080:8080 \
         --mount type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock \
         --network my-network \
         traefik \
         --docker \
         --docker.swarmmode \
         --docker.domain=traefik \
         --docker.watch \
         --web

And to my app:

    docker service create --replicas 3 --name myapp --network my-network --label traefik.port=80 myappimage

And then I went to the manager vagrant IP and IP:8080

### What did you expect to see?

just with the IP I expected my app index screen (is working fine just with the container, and my app is just a simple nginx serving a html page

with IP:8080 I expected traefik dashboard

### What did you see instead?

just IP: 404 page not found
IP:8080: all right 👍 

<img width="1164" alt="screen shot 2017-02-02 at 15 42 54" src="https://cloud.githubusercontent.com/assets/1637171/22563320/dd7f507c-e95e-11e6-87a3-6174545d00cb.png">

Chabane:
Same issue

Chabane:
If we try : https://github.com/BerlinVagrant/vagrant-dns?

WTFKr0:
Traefik is reverse proxy based on Host header in http protocol.
To join your nginx service, you have to set it to, in your case, "container_name.traefik", like the rule in the traefik dashboard
```
curl -H Host:container_name.traefik http://your_ip
```
See exemple on docs : https://docs.traefik.io/

pascalandy:
Please check out [my setup](https://github.com/containous/traefik/issues/563#issuecomment-277382568). It's working perfectly over Docker Swarm Mode 1.13

Cheers!

juanpaps03:
@Chabane even if use this plugin fixed the problem, I'm using vagrant just for development, I'll need it working to production env.
@pascalandy I used your configuration and now works!, but I'm with bug #563 :(

ldez:
[...]

Issue #1081, opened 2017-01-28
Accessing WebUI through PathPrefixStrip frontend returns 404


davendesai:
### What version of Traefik are you using (`traefik version`)?

`Traefik version v1.1.2 built on 2016-12-15`

### What is your environment & configuration (arguments, toml...)?

```yml
traefik:
    container_name: traefik
    image: hypriot/rpi-traefik:latest
    command: --web --docker --docker.domain=rpi.home --logLevel=DEBUG
    restart: always
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
      - /dev/null:/etc/traefik/traefik.toml
    ports:
      - 80:80
    labels:
      - traefik.frontend.rule=PathPrefixStrip:/traefik
      - traefik.port=8080
```

### What did you do?

Try to access the WebUI via Traefik frontend.

### What did you expect to see?

The have the URL http://rpi.home/traefik take me to the WebUI via appending `/dashboard/` - i.e. http://rpi.home/traefik/dashboard/

### What did you see instead?

Instead of appending the subdirectory `/dashboard/` to the current URL, the following HTML is received:
```
<a href="/dashboard/">Found</a>.
```
This breaks the PathPrefixStrip frontend rule and instead directs to http://rpi.home/dashboard/ which results in a 404.

cyrilgdn:
+1 Same problem here.

barhun:
AFAIK, Traefik Web UI Server doesn't support relative URIs; so you must try something like the following to achieve access without hitting the root path. After Traefik launches, go to _http://yourdomain.tld/dashboard/_ and voila!

```
labels:
  - "traefik.frontend.rule=PathPrefix:/dashboard,/api,/health"
  - "traefik.port=8080"
```

timoreimann:
With #985 being merged, we could now teach the dashboard to look for the header containing the stripped prefix path and build relative URLs accordingly.

davendesai:
@barhun That's a solution that actually completely flew over my head. Might take a look into that in an upcoming weekend 😄

ldez:
Closed by #1638, in the version 1.3.0-rc3.

cyrilgdn:
@ldez Does the fix on PathPrefixStrip really close this issue? As @timoreimann said:
> we could now teach the dashboard to look for the header containing the stripped prefix path and build relative URLs accordingly.

But it's not the case ATM no?

ldez:
[...]

Issue #1054, opened 2017-01-22
Error forwarding with err: EOF in logs when using pathprefixstrip on / route in kubernetes setup


christianhuening:
### What version of Traefik are you using (`traefik version`)?
1.1.2

### What is your environment & configuration (arguments, toml...)?
Running Daemon Set Ingress in 6-node kubernetes cluster (1.5.2) hosted on-prem using Weave as pod network. Communication between nodes and pods is working quite fine (checked with cURL etc.)

Toml:
``` toml
# traefik.toml
    logLevel = "DEBUG"
    defaultEntryPoints = ["http"]
    [entryPoints]
      [entryPoints.http]
      address = ":80"
```
ingress-controller.yaml:
``` yaml
apiVersion: extensions/v1beta1
kind: DaemonSet
metadata:
  name: traefik-ingress-controller-v1
  namespace: kube-system
  labels:
    k8s-app: traefik-ingress-lb
    kubernetes.io/cluster-service: "true"
spec:
  template:
    metadata:
      labels:
        k8s-app: traefik-ingress-lb
        name: traefik-ingress-lb
    spec:
      hostNetwork: true
      terminationGracePeriodSeconds: 60
      containers:
      - image: traefik:v1.1.2
        name: traefik-ingress-lb
        imagePullPolicy: Always
        resources:
          limits:
            cpu: 200m
            memory: 30Mi
          requests:
            cpu: 100m
            memory: 20Mi
        ports:
        - name: http
          containerPort: 80
          hostPort: 80
        - name: https
          containerPort: 443
          hostPort: 443
        - name: admin
          containerPort: 8081
        volumeMounts:
          - mountPath: "/config"
            name: "config"
        args:
        - --web
        - --web.address=:8081
        - --configfile=/config/traefik.toml
        - --kubernetes
      volumes:
        - name: config
          configMap:
            name: traefik-configmap
```
ingress.yaml
``` yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    traefik.frontend.rule.type: pathprefixstrip
  name: mars-ingress
spec:
  backend:
        serviceName: webui-svc
        servicePort: http
  rules:
  - http:
      paths:
      - path: /file
        backend:
          serviceName: file-svc
          servicePort: http
      - path: /metadata
        backend:
            serviceName: metadata-svc
            servicePort: http
      - path: /result-config
        backend:
            serviceName: resultcfg-svc
            servicePort: http
      - path: /scenario-management
[...]

PR #1638, opened 2017-05-20
Fix behavior for PathPrefixStrip

seryl:
When pushing data to downstream proxies; malformed requests were being
sent.

The corrected behavior is as follows:

| Route Stripped    |     URL                |  Passed to Backend |
| ----------------- | ---------------------- | ------------------ |
| /                 |     /                  |  /                 |

| Route Stripped    |     URL                |  Passed to Backend |
| ----------------- | ---------------------- | ------------------ |
| /stat             |     /stat              |  /                 |
| /stat             |     /stat/             |  /                 |
| /stat             |     /status            |  /status           |
| /stat             |     /stat/us           |  /us               |

| Route Stripped    |     URL                |  Passed to Backend |
| ----------------- | ------------------- | ------------------ |
| /stat/            |     /stat            |  /stat       |
| /stat/            |     /stat/           |  /             |
| /stat/            |     /status         |  /status  |
| /stat/            |     /stat/us        |  /us      |

Prior, we could strip the prefixing `/`, and we'd also ignore the case
where you want to serve something like `/api` as both the index and as a
subpath.

Additionally, this should resolve a myriad of issues relating to
kubernetes ingress `PathPrefixStrip`.

### Description

Should fix the following tickets:

* #133
* #374 
* #563
* #591
* #850 
* #863 
* #920
* #924
* #943 
* #1011 
* #1037
* #1051
* #1081 
* #1106

Potentially fixes:

* #1054
* #1123

### DIFF
--- a/middlewares/stripPrefix_test.go
+++ b/middlewares/stripPrefix_test.go
@@ -1,0 +1,103 @@
+package middlewares
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStripPrefix(t *testing.T) {
+	tests := []struct {
+		desc               string
+		prefixes           []string
+		path               string
+		expectedStatusCode int
+		expectedPath       string
+	}{
+		{
+			desc:               "no prefixes configured",
+			prefixes:           []string{},
+			path:               "/noprefixes",
+			expectedStatusCode: http.StatusNotFound,
+		},
+		{
+			desc:               "wildcard (.*) requests",
+			prefixes:           []string{"/"},
+			path:               "/",
+			expectedStatusCode: http.StatusOK,
+			expectedPath:       "/",
+		},
+		{
+			desc:               "prefix and path matching",
+			prefixes:           []string{"/stat"},
+			path:               "/stat",
+			expectedStatusCode: http.StatusOK,
+			expectedPath:       "/",
+		},
+		{
+			desc:               "path prefix on exactly matching path",
+			prefixes:           []string{"/stat/"},
+			path:               "/stat/",
+			expectedStatusCode: http.StatusOK,
+			expectedPath:       "/",
+		},
+		{
+			desc:               "path prefix on matching longer path",
+			prefixes:           []string{"/stat/"},
+			path:               "/stat/us",
+			expectedStatusCode: http.StatusOK,
+			expectedPath:       "/us",
+		},
+		{
+			desc:               "path prefix on mismatching path",
+			prefixes:           []string{"/stat/"},
+			path:               "/status",
+			expectedStatusCode: http.StatusNotFound,
+		},
+		{
+			desc:               "general prefix on matching path",
+			prefixes:           []string{"/stat"},
+			path:               "/stat/",
+			expectedStatusCode: http.StatusOK,
+			expectedPath:       "/",
+		},
+		{
+			desc:               "earlier prefix matching",
+			prefixes:           []string{"/stat", "/stat/us"},
+			path:               "/stat/us",
+			expectedStatusCode: http.StatusOK,
+			expectedPath:       "/us",
+		},
+		{
+			desc:               "later prefix matching",
+			prefixes:           []string{"/mismatch", "/stat"},
+			path:               "/stat",
+			expectedStatusCode: http.StatusOK,
+			expectedPath:       "/",
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.desc, func(t *testing.T) {
+			t.Parallel()
+			var gotPath string
+			server := httptest.NewServer(&StripPrefix{
+				Prefixes: test.prefixes,
+				Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					gotPath = r.URL.Path
+				}),
+			})
+			defer server.Close()
+
+			resp, err := http.Get(server.URL + test.path)
+			require.NoError(t, err, "Failed to send GET request")
+			assert.Equal(t, test.expectedStatusCode, resp.StatusCode, "Unexpected status code")
+
+			assert.Equal(t, test.expectedPath, gotPath, "Unexpected path")
+		})
+	}
+}
--- a/middlewares/stripPrefix.go
+++ b/middlewares/stripPrefix.go
@@ -17,15 +17,27 @@ type StripPrefix struct {
 
 func (s *StripPrefix) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	for _, prefix := range s.Prefixes {
-		if p := strings.TrimPrefix(r.URL.Path, strings.TrimSpace(prefix)); len(p) < len(r.URL.Path) {
-			r.URL.Path = p
-			r.Header[forwardedPrefixHeader] = []string{prefix}
-			r.RequestURI = r.URL.RequestURI()
-			s.Handler.ServeHTTP(w, r)
+		origPrefix := strings.TrimSpace(prefix)
+		if origPrefix == r.URL.Path {
+			r.URL.Path = "/"
+			s.serveRequest(w, r, origPrefix)
+			return
+		}
+
+		prefix = strings.TrimSuffix(origPrefix, "/") + "/"
+		if p := strings.TrimPrefix(r.URL.Path, prefix); len(p) < len(r.URL.Path) {
+			r.URL.Path = "/" + strings.TrimPrefix(p, "/")
+			s.serveRequest(w, r, origPrefix)
 			return
 		}
 	}
 	http.NotFound(w, r)
+}
+
+func (s *StripPrefix) serveRequest(w http.ResponseWriter, r *http.Request, prefix string) {
+	r.Header[forwardedPrefixHeader] = []string{prefix}
+	r.RequestURI = r.URL.RequestURI()
+	s.Handler.ServeHTTP(w, r)
 }
 
 // SetHandler sets handler


seryl:
Example of malformed requests:

```
10.244.0.164 - - [19/May/2017:23:58:14 +0000] "GET app.301d8f1c250fa1437ad7.js HTTP/1.1" 400 173 "-" "-" "-"
10.244.0.164 - - [19/May/2017:23:58:15 +0000] "GET app.301d8f1c250fa1437ad7.js HTTP/1.1" 400 173 "-" "-" "-"
10.244.0.164 - - [19/May/2017:23:58:17 +0000] "GET app.301d8f1c250fa1437ad7.js HTTP/1.1" 400 173 "-" "-" "-"
10.244.0.164 - - [19/May/2017:23:58:58 +0000] "GET _/public/img/pack.png HTTP/1.1" 400 173 "-" "-" "-"
10.244.0.164 - - [19/May/2017:23:59:07 +0000] "GET favicon.ico HTTP/1.1" 400 173 "-" "-" "-"
10.244.0.164 - - [20/May/2017:00:00:35 +0000] "GET app.301d8f1c250fa1437ad7.js HTTP/1.1" 400 173 "-" "-" "-"
10.244.0.164 - - [20/May/2017:00:00:35 +0000] "GET app.301d8f1c250fa1437ad7.js HTTP/1.1" 400 173 "-" "-" "-"
10.244.0.164 - - [20/May/2017:00:00:38 +0000] "GET favicon.ico HTTP/1.1" 400 173 "-" "-" "-"
10.244.0.164 - - [20/May/2017:00:00:41 +0000] "GET favicon.ico HTTP/1.1" 400 173 "-" "-" "-"
10.244.0.164 - - [20/May/2017:00:00:44 +0000] "GET favicon.ico HTTP/1.1" 400 173 "-" "-" "-
```

**Note**: They are all missing prefixed paths.

timoreimann:
@emilevauge @ldez what do you think, should we have this fix be part of 1.3?

middlewares/stripPrefix_test.go
```diff
@@ -0,0 +1,68 @@
...
+		if resp.StatusCode != test.expectedCode {
+			t.Errorf("Received non-%d response: %d", test.expectedCode, resp.StatusCode)
+		}
+		response, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if test.expectedResponse != string(response) {
+			t.Errorf("Expected '%s' :  '%s'\n", test.expectedResponse, response)
```
seryl: I understand, I was just following the other PathPrefixStripRegex [example](https://github.com/containous/traefik/blob/master/middlewares/stripPrefixRegex_test.go#L43-L51).

I'll add these fixes shortly.

seryl:
@timoreimann Updated with your requested changes.

Let me know if there's anything else we need (rebase, etc).

middlewares/stripPrefix_test.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+	for _, stripPath := range stripList {
+		stripServers[stripPath] = httptest.NewServer(&StripPrefix{
+			Prefixes: []string{stripPath},
+			Handler:  handlerPath,
+		})
+	}
+
+	suites := []struct {
+		server *httptest.Server
```
ldez: To enhance readability, create 2 types at the below of the file:
```go
type stripPrefixTestSuite struct {
	server *httptest.Server
	desc   string
	tests  []stripPrefixTestCase
}

type stripPrefixTestCase struct {
	url                string
	expectedStatusCode int
	expectedBody       string
}
```

middlewares/stripPrefix_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	for _, suite := range suites {
+		suite := suite
+		t.Run(suite.desc, func(t *testing.T) {
+			t.Parallel()
+			defer suite.server.Close()
+
+			for _, test := range suite.tests {
+				resp, err := http.Get(suite.server.URL + test.url)
+				if err != nil {
```
ldez: You can use Testify, to enhance readability:
```go
resp, err := http.Get(suite.server.URL + test.url)
require.NoError(t, err, "Failed to send GET request")
assert.Equal(t, test.expectedStatusCode, resp.StatusCode, "Unexpected status code")

if test.expectedBody != "" {
	response, err := ioutil.ReadAll(resp.Body)
	require.NoError(t, err, "Failed to read response body")

	assert.Equal(t, test.expectedBody, string(response), "Unexpected response received")
}
```

middlewares/stripPrefix_test.go
```diff
@@ -0,0 +1,177 @@
...
+			server: stripServers["/norules"],
+			desc:   "no strip rules",
+			tests: []struct {
+				url      string
+				expected string
+				code     int
+			}{
+				{
+					url:      "/norules",
+					expected: "",
```
ldez: for the expressivity of the test, you can drop all `expected: ""`

middlewares/stripPrefix_test.go
```diff
@@ -0,0 +1,177 @@
...
+			desc:   "strip / to handle wildcard (.*) requests",
+			tests: []struct {
+				url      string
+				expected string
+				code     int
+			}{
+				{
+					url:      "/",
+					expected: "/",
+					code:     http.StatusOK,
```
ldez: can you re-odering? Put `code` before `expected`

middlewares/stripPrefix_test.go
seryl: Alright, these changes were added as well as your above suggestions.

middlewares/stripPrefix_test.go
```diff
@@ -0,0 +1,153 @@
...
+		"/nomatch",
+	}
+	for _, stripPath := range stripList {
+		stripServers[stripPath] = httptest.NewServer(&StripPrefix{
+			Prefixes: []string{stripPath},
+			Handler:  handlerPath,
+		})
+	}
+
+	type stripPrefixTestCase struct {
```
ldez: Please, put this 2 types at the below of file.

seryl:
Will rebase as soon as we get a thumbs up.

middlewares/stripPrefix_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+					assert.Equal(t, test.expectedBody, string(response), "Unexpected response received")
+				}
+			}
+		})
+	}
+}
+
+type stripPrefixTestCase struct {
+	url                string
```
timoreimann: This should probably be better called `path`.

middlewares/stripPrefix_test.go
```diff
@@ -0,0 +1,153 @@
...
+		},
+	}
+
+	for _, suite := range suites {
+		suite := suite
+		t.Run(suite.desc, func(t *testing.T) {
+			t.Parallel()
+			defer suite.server.Close()
+
+			for _, test := range suite.tests {
```
timoreimann: The given nested suite/test cases approaches comes with a few drawbacks:

- Since the description is attached to the suite, we can't see which test case exactly is failing.
- For the same reason, we cannot execute individual test cases.
- If one test in a suite fails, the entire suite is aborted.

Overall, I also find the nested approach more difficult to understand.

All of this should be solvable by going back to a single, flat hierarchy.

middlewares/stripPrefix_test.go
```diff
@@ -0,0 +1,153 @@
...
+		"/",
+		"/stat/",
+		"/stat",
+		"/nomatch",
+	}
+	for _, stripPath := range stripList {
+		stripServers[stripPath] = httptest.NewServer(&StripPrefix{
+			Prefixes: []string{stripPath},
+			Handler:  handlerPath,
+		})
```
timoreimann: Initializing outside of the parallelized context could be problematic. For instance, if I execute your test like this

`go test -count 20 -cpu 1,2,4 -run ^TestStripPrefix$ ./middlewares`

I immediately run into `too many open files` problems on my Mac.

This might not be an issue now but could possibly become one in the future.

ldez:
@timoreimann your https://github.com/timoreimann/traefik/commit/224f2a1d86d9fe070e384a10f5b07f32d95ec22d is a very clean way to test ❤️

middlewares/stripPrefix_test.go
seryl: Help me understand why you're seeing multiple files opened.

middlewares/stripPrefix_test.go
seryl: I added what I thought were the right test to begin with.

I ended up added test to validate per request status; tradeoffs are perf. I don't know if we lose anything here to be frank.

middlewares/stripPrefix_test.go
timoreimann: Linux/Unix (which Mac is, technically) represent almost everything in terms of files, including sockets. When I repeated the test 20 times and simulated multiples cores above, a higher number of sockets (and thus file handles) would be consumed concurrently by the test server.

It's perfectly okay to hit the limit at some point because my Mac machine obviously isn't optimized for highly concurrent request throughput. What's interesting though is that with the simpler approach creating one test server per more narrow-scoped test execution routine, I only see the limit when running on 32 or 64 cores.

This really isn't supposed to be a hard blocker on the given approach at all, just an indication on which implementation may be easier on the resource consumption IMHO.

middlewares/stripPrefix_test.go
timoreimann: I'm definitely okay with the (unique set of) tests you have selected. (In fact, I think you included some which I would have likely forgotten.)

It's just that I think structurally, we're not reaping the benefits of sub-tests with the current approach; notably the points described in the _Table-driven tests using subtests_ section of the [introductory blog post](https://blog.golang.org/subtests).

PR #1639, opened 2017-05-20
Fix HTTP Trailers and gRPC responses.

tcolgate:
NOTE: This PR is a direct hack to the vendored oxy. I
tried recreting the PR for containous/mux, but the
commit mentioned in glide does not seem to exist.

gRPC makes use of response trailers. oxy was not
passing on the trailers. In addition, the standard
grpc.Serve() does not pre-announce the trailers it
will send.

oxy upstream has  moved to using httputil.ReverseProxy,
which would be fine, but that is broken for unannounced
trailers too.

This PR probably needs a test.


tcolgate:
Clsoing and re-raising against the oxy fork

PR #1644, opened 2017-05-21
Adds definitions to backend kv template for health checking

zachomedia:
### Description

This PR adds definitions to the KV template to load backend health check information (path and interval).

I've only tested this change with the etcd backend.

### DIFF
--- a/templates/kv.tmpl
+++ b/templates/kv.tmpl
@@ -19,6 +19,13 @@
     sticky = {{$sticky}}
 {{end}}
 
+{{$healthCheck := Get "" . "/healthcheck/" "path"}}
+{{with $healthCheck}}
+[backends."{{Last $backend}}".healthCheck]
+    path = "{{$healthCheck}}"
+    interval = "{{ Get "30s" $backend "/healthcheck/" "interval" }}"
+{{end}}
+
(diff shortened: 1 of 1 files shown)


templates/kv.tmpl
```diff
@@ -19,6 +19,14 @@
     sticky = {{$sticky}}
 {{end}}
 
+{{$healthCheck := Get "" . "/healthcheck/" "path"}}
+{{$interval := Get "30s" . "/healthcheck/" "interval"}}
+{{with $healthCheck}}
```
emilevauge: We may move `{{$interval := Get "30s" . "/healthcheck/" "interval"}}` below `{{with $healthCheck}}` ?

zachomedia: I followed the same format as the load balancer line above (where `{{$sticky := Get "false" . "/loadbalancer/" "sticky"}}` is outside the `with`). I can move it inside the `{{with $healthCheck}}` if you want.

emilevauge: It's not mandatory, but it would be better :)

zachomedia: @emilevauge I moved the `interval` `Get` into the `with` block.

PR #1507, opened 2017-04-28
Create log folder if not present

tanyadegurechaff:
### Description

Fixes #1217

PR #1647, opened 2017-05-21
Switch access logging to logrus

rjshep:
Building on some initial work to introduce a new access logging framework, this PR uses logrus as the log handler to write access logs using information collected in the log data table.

A logrus Formatter implements Traefik Common Log Format producing logs which are backward compatible with the current access logs.

Also includes applicable changes from [#1507](https://github.com/containous/traefik/pull/1507)


server/server.go
```diff
@@ -544,7 +541,6 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 	redirectHandlers := make(map[string]negroni.Handler)
 	backends := map[string]http.Handler{}
 	backendsHealthcheck := map[string]*healthcheck.BackendHealthCheck{}
-	backend2FrontendMap := map[string]string{}
```
ldez: why you have remove this line?

server/server.go
```diff
@@ -643,7 +639,6 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 								log.Errorf("Skipping frontend %s...", frontendName)
 								continue frontend
 							}
-							backend2FrontendMap[url.String()] = frontendName
```
ldez: same here

server/server.go
```diff
@@ -670,7 +665,6 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
...
```
ldez: same here

server/server.go
```diff
@@ -762,7 +756,6 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 		}
 	}
 	healthcheck.GetHealthCheck().SetBackendsConfiguration(server.routinesPool.Ctx(), backendsHealthcheck)
-	middlewares.SetBackend2FrontendMap(&backend2FrontendMap)
```
ldez: same here

server/server.go
rjshep: Because the backend to frontend map was only used by the old logger.  The new framework, added during a previous PR, uses SaveFrontend and SaveBackend to achieve the same thing.

middlewares/accesslog/logger_formatters.go
```diff
@@ -0,0 +1,69 @@
...
+			return "-"
+		}
+		return `"` + s + `"`
+
+	case fmt.Stringer:
+		return `"` + s.String() + `"`
+	}
+
+	return "-"
+}
```
emilevauge: WDYT of this kind of refactoring?

```go
func toLogString(v interface{}) string {
	defaultValue := "-"
	if v == nil {
		return defaultValue
	}

	switch s := v.(type) {
	case string:
		return quoted(s, defaultValue)

	case fmt.Stringer:
		return quoted(s.String(), defaultValue)

	default:
		return defaultValue
	}

}

func quoted(s string, defaultValue string) string {
	if len(s) == 0 {
		return defaultValue
	}
	return `"` + s + `"`
}
```

rjshep: LGTM I've added it in.  It has meant that the log format is not fully backward compatible.  Referer and User Agent fields are now specified as a hyphen if not present, previously they were just missing.  I believe the new format is actually correct, however.

rjshep:
Hi @emilevauge I've made the requested changes.  Thanks for the design change, that is a better solution :)

integration/access_log_test.go
```diff
@@ -78,12 +78,12 @@ func (s *AccessLogSuite) TestAccessLog(c *check.C) {
...
-			c.Assert(len(tokens), checker.Equals, 13)
+			c.Assert(len(tokens), checker.Equals, 14)
 			c.Assert(regexp.MustCompile(`^\d{3}$`).MatchString(tokens[6]), checker.True)
-			c.Assert(tokens[9], checker.Equals, fmt.Sprintf("%d", i+1))
-			c.Assert(strings.HasPrefix(tokens[10], "frontend"), checker.True)
-			c.Assert(strings.HasPrefix(tokens[11], "http://127.0.0.1:808"), checker.True)
-			c.Assert(regexp.MustCompile(`^\d+ms$`).MatchString(tokens[12]), checker.True)
+			c.Assert(tokens[10], checker.Equals, fmt.Sprintf("%d", i+1))
+			c.Assert(strings.HasPrefix(tokens[11], "frontend"), checker.True)
+			c.Assert(strings.HasPrefix(tokens[12], "http://127.0.0.1:808"), checker.True)
```
timoreimann: Comprehension question: why do we have to change the indices?

middlewares/accesslog/logger.go
```diff
@@ -23,11 +29,34 @@ const (
 // Note: Current implementation collects log data but does not have the facility to
```
timoreimann: Is this comment still valid?

middlewares/accesslog/logger.go
```diff
@@ -23,11 +29,34 @@ const (
...
 // write anywhere.
 type LogHandler struct {
+	logger *logrus.Logger
+	file   *os.File
 }
 
 // NewLogHandler creates a new LogHandler
-func NewLogHandler() *LogHandler {
-	return &LogHandler{}
+func NewLogHandler(fileName string) (*LogHandler, error) {
```
timoreimann: Nit-pick: I'd probably call the variable `filePath`.

middlewares/accesslog/logger.go
```diff
@@ -23,11 +29,34 @@ const (
...
+	if len(fileName) > 0 {
+		dir := filepath.Dir(fileName)
+
+		err := os.MkdirAll(dir, 0755)
+		if err != nil {
```
timoreimann: Can be merged with previous line.

middlewares/accesslog/logger.go
```diff
@@ -23,11 +29,34 @@ const (
...
+			log.Errorf("Failed to create log path %s: %s", dir, err)
```
timoreimann: Shouldn't we return an error here?

middlewares/accesslog/logger.go
```diff
@@ -23,11 +29,34 @@ const (
...
+		}
+
+		file, err := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0664)
+		if err != nil {
+			log.Error("Error opening file", err)
```
timoreimann: Shouldn't we return an error instead?

middlewares/accesslog/logger.go
```diff
@@ -23,11 +29,34 @@ const (
...
+		logger := &logrus.Logger{
+			Out:       file,
+			Formatter: new(CommonLogFormatter),
+			Hooks:     make(logrus.LevelHooks),
+			Level:     logrus.InfoLevel,
+		}
+		return &LogHandler{logger: logger, file: file}, nil
+	}
+
+	return nil, errors.New("Empty file name specified for accessLogsFile")
```
timoreimann: Please move this return at the very top of the file, right after a check for `fileName == ""`. That way, we can save one level of indentation.

middlewares/accesslog/logger_test.go
```diff
@@ -47,16 +42,15 @@ func TestLogger(t *testing.T) {
 
 	logfilePath := filepath.Join(tmp, logfileNameSuffix)
 
-	logger = NewLogger(logfilePath)
+	logger, err = NewLogHandler(logfilePath)
+	assert.NoError(t, err)
```
timoreimann: We have added `testify.require` by now. Please use it here and wherever continuation of a test does not make sense anymore in case of a failed assertion.

integration/access_log_test.go
rjshep: The Referer and User Agent fields are now specified as a hyphen if not present, previously they were just missing. I believe the new format is actually correct, however.

middlewares/accesslog/logger.go
rjshep: Done :-)

middlewares/accesslog/logger.go
rjshep: Done

middlewares/accesslog/logger.go
rjshep: Yes

middlewares/accesslog/logger.go
rjshep: Done

middlewares/accesslog/logger.go
rjshep: Yes

middlewares/accesslog/logger_test.go
rjshep: Changed

middlewares/accesslog/logger.go
```diff
@@ -19,15 +25,36 @@ const (
...
-	return &LogHandler{}
+func NewLogHandler(filePath string) (*LogHandler, error) {
+	if len(filePath) == 0 {
+		return nil, errors.New("Empty file name specified for accessLogsFile")
+	}
+
+	dir := filepath.Dir(filePath)
+
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		log.Errorf("Failed to create log path %s: %s", dir, err)
```
timoreimann: Logging an error and returning an error normally do not go together; instead, we should return an error containing sufficient context and let the caller log.

Could you please pass the (de-capitalized) log message content to an `fmt.Errorf` call and just return that?

Same for lines 48:49.

rjshep:
@timoreimann change made and commits squashed.

middlewares/accesslog/logger.go
```diff
@@ -85,6 +110,7 @@ func (l *LogHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next h
 
 // Close closes the Logger (i.e. the file etc).
 func (l *LogHandler) Close() error {
+	l.file.Close()
```
emilevauge: I think it should be `return l.file.Close()` ;)

server/server.go
```diff
@@ -156,8 +158,9 @@ func (server *Server) Close() {
 	signal.Stop(server.signals)
 	close(server.signals)
 	close(server.stopChan)
-	server.loggerMiddleware.Close()
-	server.accessLoggerMiddleware.Close()
+	if server.accessLoggerMiddleware != nil {
+		server.accessLoggerMiddleware.Close()
```
emilevauge: You should check the return error here.

rjshep:
Thanks @emilevauge. Changes made :)

Issue #884, opened 2016-11-23
Custom or more verbose 404 page


WTFKr0:
Hi !

I'm trying the new feature of traefik constraints.
I have a front traefik, which route to n traefik instances.
Those traefik route to a set of containers.
All filtering by tags
It's not working and it's a pain to debug that when i got a 404 not found. I can't easily see which traefik send me that.

Is there a way to put some information (maybe customizable) in the default 404 page ? (like hostname or tags ?)

Thanks for reading

emilevauge:
@WTFKr0 We will add support OpenTracing #771 to easily debug this kind of issue. It should help ;)

WTFKr0:
OpenTracing should certainly help for that.  
Beyond that, I think it would be necessary for companies using traefik to provide a template for the 404 page.  
Like the ability to place a logo, a text, or some variables in that page

ryanleary:
> Like the ability to place a logo, a text, or some variables in that page

Perhaps even a template per frontend.

WTFKr0:
Hum I think of the 404 page when there is no front-end rule found 😁

salizzar:
Another good option is to redirect to a custom domain, like this thread http://stackoverflow.com/questions/9147862/nginx-return-301-redirect-when-404-error 🙂

ldez:
Closing this issue in favor of  #1634.

emilevauge:
Reopening, as it's a bit different from #1634 ;)

emilevauge:
Closing in favor of #1634 and #771

Issue #1634, opened 2017-05-19
Custom error pages


emilevauge:
### Do you want to request a *feature* or report a *bug*?

feature

### Description

This issue is a proposal following the PR https://github.com/containous/traefik/pull/1576 on custom error pages. See comment https://github.com/containous/traefik/pull/1576#issuecomment-302450897.

By frontend, we could delegate error page rendering on status code. For example, if I get a `500` status on frontend `website`, I query the frontend `error` with path `/500.html` and the error page would then be rendered by a specific service in your infrastructure with a webserver, Nginx for example.

```toml
 [frontends]
   [frontends.website]
   backend = "website"
   [errors]
     [error.network]
     status = "[500-600]"
     backend = "error"
     query = "/{status}.html"
   [frontends.website.routes.website]
   rule = "Host: website.mydomain.com"

 [backends]
   [backends.website]
     [backends.website.servers.website]
     url = "https://1.2.3.4"
   [backends.error]
     [backends.error.servers.error]
     url = "http://2.3.4.5"
```

/cc @bparli

Yajo:
Would it be possible to serve a static file too?
And to have more replacement strings in the file pattern?

Something like `/var/run/traefik/{frontend}/{status}.html` or `/etc/traefik/{hostname}/{status}.html`.

I like design from https://github.com/containous/traefik/issues/1634#issue-230001382, but it means to have a running service just to serve error pages, and that seems a bit overkill for such a simple task in some situations IMHO.

0xbaadf00d:
Any news about this feature ? currently, we can't put Traefik in production cause users won't understand why a 404 is raised in case of a general backend failure.

timoreimann:
With #1675 being merged, I think we can consider this feature to be done and to ship in the upcoming 1.4 release.

We are quite hesitant to support serving static files directly from Traefik as this will likely call for evermore features to be added in the same direction. One of Traefik's philosophies is to try to keep things as simple as possible and avoid slowly turning into the feature behemoth that other proxies are.

Serving error page containers in a reliable and safe way should be easy these days while it keeps flexibility high.

timoreimann:
@Yajo feel free to file a dedicated feature request to discuss your proposal on including more string variables.

guillaumebriday:
Hey ! First of all thank you for your work !

[...]

PR #1664, opened 2017-05-24
Custom 404 mapping

jangie:
### Custom 404 mapping

Addresses #884 , which is a pain point for us in helping our developers troubleshoot where in the stack they are experiencing issues.

### DIFF
--- a/server/adapters.go
+++ b/server/adapters.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/containous/traefik/log"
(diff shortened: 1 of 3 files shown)


emilevauge:
@jangie We just opened a proposal https://github.com/containous/traefik/issues/1634 on this. We want to avoid serving static files from Traefik. WDYT?

ldez:
Relate to  #1576

jangie:
@emilevauge Sorry! Didn't realize. After reading the proposal, I guess I'm curious about what cases the errors apply to. In this particular case, this is intended to show up if there is no matched frontend mapping rule. For the 500-level errors as listed in the proposal, would those be coming from the mapped backend?

I guess as a general piece: I'm somewhat in favor of static files being served by Traefik. There is, for example, no guarantee that the error backend is able to serve traffic, and fallback in case of that backend being down or missing a particular file is potentially more complicated to deal with. That said, I am not married to that or to this PR at all and am totally happy to close this out in favor of that eventual PR.

timoreimann:
Without having looked closer at this PR, I also get the feeling it serves a different purposes than what the proposal is trying to cover.

That said, I am not 100% sure whether exposing routing details via 404 responses is the right approach. Couldn't that potentially leak internal information to outsiders? Would the (access) logs possibly be a better place to store debugging information?

jangie:
@timoreimann We're aiming to write our custom messaging such that it doesn't specifically list that out, but at least to denote what it is that is giving back the 404. As it stands, the generic 404 message is not particularly clear at that point.

As it stands, there wouldn't be any explicit routing information given back in the envisioned usage of this as this is specifically going to be giving back a static set of content which does not have any of the request's context.

timoreimann:
@jangie to better understand your use case, could you give an example of what kind of information you'd be passing back to your users along the custom 404 page?

jangie:
A static "Welcome to Traefik, we couldn't find what you were looking for" would be akin to what we're trying to do.

timoreimann:
@jangie thanks for the example. It clarifies things for me and now it does look like #1634 would be able to solve your case too (albeit in a less static way).

jangie:
👍  closing this out then, thanks! will be watching #1634 .

PR #1667, opened 2017-05-24
Update CONTRIBUTING.md.

timoreimann:
- Go 1.8 is the current minimum requirement.
- The main binary moved to `cmd/traefik`.
- Remove obsolete gox example.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -12,7 +12,7 @@ You need to run the `binary` target. This will create binaries for Linux platfor
 $ make binary
 docker build -t "traefik-dev:no-more-godep-ever" -f build.Dockerfile .
 Sending build context to Docker daemon 295.3 MB
-Step 0 : FROM golang:1.7
+Step 0 : FROM golang:1.8
  ---> 8c6473912976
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -64,7 +64,7 @@ $ ( cd integration && ../script/glide.sh get github.com/baz/quuz )
 # generate (Only required to integrate other components such as web dashboard)
 $ go generate
 # Standard go build
-$ go build
+$ go build ./cmd/traefik
 # Using gox to build multiple platform
 $ gox "linux darwin" "386 amd64 arm" \
```
timoreimann: I'm wondering if we should kick this one out. Personally, I've never used it, but maybe it's just me...

timoreimann: I'm wondering if we should kick this one out. Personally, I've never used it, but maybe it's just me...

emilevauge: yeah, I agree too. We don't use gox anymore.

timoreimann: Alright, removed.

PR #1672, opened 2017-05-26
Allow file provider to load service config from files in a directory.

rjshep:
Currently the file provider reads service config either from the global config, or from a single file specified in global config.

In some deployments it would be nice to load from multiple files - perhaps because each service is defined in a single file and is maintained separately from the others.  This PR extends the file provider to allow this.

Also added a unit test for the file provider and fixed bug in integration test which didn't check err response of GetRequest.

### DIFF
--- a/integration/fixtures/file/directory.toml
+++ b/integration/fixtures/file/directory.toml
@@ -1,0 +1,10 @@
+defaultEntryPoints = ["http"]
+
+logLevel = "DEBUG"
+
+[entryPoints]
+  [entryPoints.http]
+  address = ":8000"
+
+[file]
+  directory = "fixtures/file/dir/"
--- a/provider/file/file.go
+++ b/provider/file/file.go
@@ -1,8 +1,9 @@
 package file
 
 import (
-	"os"
-	"path/filepath"
+	"fmt"
+	"io/ioutil"
+	"path"
 	"strings"
 
 	"github.com/BurntSushi/toml"
@@ -18,68 +19,138 @@ var _ provider.Provider = (*Provider)(nil)
 // Provider holds configurations of the provider.
 type Provider struct {
 	provider.BaseProvider `mapstructure:",squash"`
+	Directory             string `description:"Load configuration from one or more .toml files in a directory"`
 }
 
 // Provide allows the file provider to provide configurations to traefik
 // using the given configuration channel.
 func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *safe.Pool, constraints types.Constraints) error {
-	watcher, err := fsnotify.NewWatcher()
-	if err != nil {
-		log.Error("Error creating file watcher", err)
-		return err
-	}
+	configuration, err := p.loadConfig()
 
-	file, err := os.Open(p.Filename)
 	if err != nil {
-		log.Error("Error opening file", err)
 		return err
 	}
-	defer file.Close()
 
 	if p.Watch {
-		// Process events
-		pool.Go(func(stop chan bool) {
-			defer watcher.Close()
-			for {
-				select {
-				case <-stop:
-					return
-				case event := <-watcher.Events:
-					if strings.Contains(event.Name, file.Name()) {
-						log.Debug("Provider event:", event)
-						configuration := p.loadFileConfig(file.Name())
-						if configuration != nil {
-							configurationChan <- types.ConfigMessage{
-								ProviderName:  "file",
-								Configuration: configuration,
-							}
-						}
-					}
-				case error := <-watcher.Errors:
-					log.Error("Watcher event error", error)
-				}
-			}
-		})
-		err = watcher.Add(filepath.Dir(file.Name()))
-		if err != nil {
-			log.Error("Error adding file watcher", err)
+		var watchItem string
+
+		if p.Directory != "" {
+			watchItem = p.Directory
+		} else {
+			watchItem = p.Filename
+		}
+
+		if err := p.addWatcher(pool, watchItem, configurationChan, p.watcherCallback); err != nil {
 			return err
 		}
 	}
 
-	configuration := p.loadFileConfig(file.Name())
+	sendConfigToChannel(configurationChan, configuration)
+	return nil
+}
+
+func (p *Provider) addWatcher(pool *safe.Pool, directory string, configurationChan chan<- types.ConfigMessage, callback func(chan<- types.ConfigMessage, fsnotify.Event)) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return fmt.Errorf("error creating file watcher: %s", err)
+	}
+
+	// Process events
+	pool.Go(func(stop chan bool) {
+		defer watcher.Close()
+		for {
+			select {
+			case <-stop:
+				return
+			case evt := <-watcher.Events:
+				callback(configurationChan, evt)
+			case err := <-watcher.Errors:
+				log.Errorf("Watcher event error: %s", err)
+			}
+		}
+	})
+	err = watcher.Add(directory)
+	if err != nil {
+		return fmt.Errorf("error adding file watcher: %s", err)
+	}
+
+	return nil
+}
+
+func sendConfigToChannel(configurationChan chan<- types.ConfigMessage, configuration *types.Configuration) {
 	configurationChan <- types.ConfigMessage{
 		ProviderName:  "file",
 		Configuration: configuration,
 	}
-	return nil
 }
 
-func (p *Provider) loadFileConfig(filename string) *types.Configuration {
+func loadFileConfig(filename string) (*types.Configuration, error) {
 	configuration := new(types.Configuration)
 	if _, err := toml.DecodeFile(filename, configuration); err != nil {
-		log.Error("Error reading file:", err)
-		return nil
+		return nil, fmt.Errorf("error reading configuration file: %s", err)
 	}
-	return configuration
+	return configuration, nil
+}
+
+func loadFileConfigFromDirectory(directory string) (*types.Configuration, error) {
+	fileList, err := ioutil.ReadDir(directory)
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to read directory %s: %v", directory, err)
+	}
+
+	configuration := &types.Configuration{
+		Frontends: make(map[string]*types.Frontend),
+		Backends:  make(map[string]*types.Backend),
+	}
+
+	for _, file := range fileList {
+		if !strings.HasSuffix(file.Name(), ".toml") {
+			continue
+		}
+
+		var c *types.Configuration
+		c, err = loadFileConfig(path.Join(directory, file.Name()))
+
+		if err != nil {
+			return nil, err
+		}
+
+		for backendName, backend := range c.Backends {
+			if _, exists := configuration.Backends[backendName]; exists {
+				log.Warnf("Backend %s already configured, skipping", backendName)
+			} else {
+				configuration.Backends[backendName] = backend
+			}
+		}
+
+		for frontendName, frontend := range c.Frontends {
+			if _, exists := configuration.Frontends[frontendName]; exists {
+				log.Warnf("Frontend %s already configured, skipping", frontendName)
+			} else {
+				configuration.Frontends[frontendName] = frontend
+			}
+		}
+	}
+
+	return configuration, nil
+}
+
+func (p *Provider) watcherCallback(configurationChan chan<- types.ConfigMessage, event fsnotify.Event) {
+	configuration, err := p.loadConfig()
+
+	if err != nil {
+		log.Errorf("Error occurred during watcher callback: %s", err)
+		return
+	}
+
+	sendConfigToChannel(configurationChan, configuration)
+}
+
+func (p *Provider) loadConfig() (*types.Configuration, error) {
+	if p.Directory != "" {
+		return loadFileConfigFromDirectory(p.Directory)
+	}
+
+	return loadFileConfig(p.Filename)
 }
--- a/provider/file/file_test.go
+++ b/provider/file/file_test.go
@@ -1,0 +1,262 @@
+package file
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/containous/traefik/safe"
+	"github.com/containous/traefik/types"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProvideSingleFileAndWatch(t *testing.T) {
+	tempDir := createTempDir(t, "testfile")
+	defer os.RemoveAll(tempDir)
+
+	expectedNumFrontends := 2
+	expectedNumBackends := 2
+
+	tempFile := createFile(t,
+		tempDir, "simple.toml",
+		createFrontendConfiguration(expectedNumFrontends),
+		createBackendConfiguration(expectedNumBackends))
+
+	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)
+
+	provide(configurationChan, watch, withFile(tempFile))
+
+	// Wait for initial message to be tested
+	err := waitForSignal(signal, 2*time.Second, "initial config")
+	assert.NoError(t, err)
+
+	// Now test again with single frontend and backend
+	expectedNumFrontends = 1
+	expectedNumBackends = 1
+
+	tempFile = createFile(t,
+		tempDir, "simple.toml",
+		createFrontendConfiguration(expectedNumFrontends),
+		createBackendConfiguration(expectedNumBackends))
+
+	// Must fail because we don't watch the change
+	err = waitForSignal(signal, 2*time.Second, "single frontend and backend")
+	assert.NoError(t, err)
+}
+
+func TestProvideSingleFileAndNotWatch(t *testing.T) {
+	tempDir := createTempDir(t, "testfile")
+	defer os.RemoveAll(tempDir)
+
+	expectedNumFrontends := 2
+	expectedNumBackends := 2
+
+	tempFile := createFile(t,
+		tempDir, "simple.toml",
+		createFrontendConfiguration(expectedNumFrontends),
+		createBackendConfiguration(expectedNumBackends))
+
+	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)
+
+	provide(configurationChan, withFile(tempFile))
+
+	// Wait for initial message to be tested
+	err := waitForSignal(signal, 2*time.Second, "initial config")
+	assert.NoError(t, err)
+
+	// Now test again with single frontend and backend
+	expectedNumFrontends = 1
+	expectedNumBackends = 1
+
+	tempFile = createFile(t,
+		tempDir, "simple.toml",
+		createFrontendConfiguration(expectedNumFrontends),
+		createBackendConfiguration(expectedNumBackends))
+
+	// Must fail because we don't watch the changes
+	err = waitForSignal(signal, 2*time.Second, "single frontend and backend")
+	assert.Error(t, err)
+}
+
+func TestProvideDirectoryAndWatch(t *testing.T) {
+	tempDir := createTempDir(t, "testdir")
+	defer os.RemoveAll(tempDir)
+
+	expectedNumFrontends := 2
+	expectedNumBackends := 2
+
+	tempFile1 := createRandomFile(t, tempDir, createFrontendConfiguration(expectedNumFrontends))
+	tempFile2 := createRandomFile(t, tempDir, createBackendConfiguration(expectedNumBackends))
+
+	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)
+
+	provide(configurationChan, watch, withDirectory(tempDir))
+
+	// Wait for initial config message to be tested
+	err := waitForSignal(signal, 2*time.Second, "initial config")
+	assert.NoError(t, err)
+
+	// Now remove the backends file
+	expectedNumFrontends = 2
+	expectedNumBackends = 0
+	os.Remove(tempFile2.Name())
+	err = waitForSignal(signal, 2*time.Second, "remove the backends file")
+	assert.NoError(t, err)
+
+	// Now remove the frontends file
+	expectedNumFrontends = 0
+	expectedNumBackends = 0
+	os.Remove(tempFile1.Name())
+	err = waitForSignal(signal, 2*time.Second, "remove the frontends file")
+	assert.NoError(t, err)
+}
+
+func TestProvideDirectoryAndNotWatch(t *testing.T) {
+	tempDir := createTempDir(t, "testdir")
+	defer os.RemoveAll(tempDir)
+
+	expectedNumFrontends := 2
+	expectedNumBackends := 2
+
+	createRandomFile(t, tempDir, createFrontendConfiguration(expectedNumFrontends))
+	tempFile2 := createRandomFile(t, tempDir, createBackendConfiguration(expectedNumBackends))
+
+	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)
+
+	provide(configurationChan, withDirectory(tempDir))
+
+	// Wait for initial config message to be tested
+	err := waitForSignal(signal, 2*time.Second, "initial config")
+	assert.NoError(t, err)
+
+	// Now remove the backends file
+	expectedNumFrontends = 2
+	expectedNumBackends = 0
+	os.Remove(tempFile2.Name())
+
+	// Must fail because we don't watch the changes
+	err = waitForSignal(signal, 2*time.Second, "remove the backends file")
+	assert.Error(t, err)
+
+}
+
+func createConfigurationRoutine(t *testing.T, expectedNumFrontends *int, expectedNumBackends *int) (chan types.ConfigMessage, chan interface{}) {
+	configurationChan := make(chan types.ConfigMessage)
+	signal := make(chan interface{})
+
+	go func() {
+		for {
+			data := <-configurationChan
+			assert.Equal(t, "file", data.ProviderName)
+			assert.Len(t, data.Configuration.Frontends, *expectedNumFrontends)
+			assert.Len(t, data.Configuration.Backends, *expectedNumBackends)
+			signal <- nil
+		}
+	}()
+
+	return configurationChan, signal
+}
+
+func waitForSignal(signal chan interface{}, timeout time.Duration, caseName string) error {
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+
+	select {
+	case <-signal:
+
+	case <-timer.C:
+		return fmt.Errorf("Timed out waiting for assertions to be tested: %s", caseName)
+	}
+	return nil
+}
+
+func provide(configurationChan chan types.ConfigMessage, builders ...func(p *Provider)) {
+	pvd := &Provider{}
+
+	for _, builder := range builders {
+		builder(pvd)
+	}
+
+	pvd.Provide(configurationChan, safe.NewPool(context.Background()), nil)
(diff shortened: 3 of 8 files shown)


provider/file/file_test.go
```diff
@@ -0,0 +1,196 @@
...
+	"sync"
+	"testing"
+
+	"github.com/containous/traefik/safe"
+	"github.com/containous/traefik/types"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	frontends = `
```
timoreimann: Can we reduce the test definitions?

provider/file/file_test.go
```diff
@@ -0,0 +1,196 @@
...
+	c := make(chan types.ConfigMessage)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	numFrontends := 3
+	numBackends := 2
+
+	go func() {
+		for {
```
timoreimann: Seems like we don't need the for loop, do we?

provider/file/file_test.go
```diff
@@ -0,0 +1,196 @@
...
+			data := <-c
+			assert.Equal(t, "file", data.ProviderName)
+			assert.Equal(t, numFrontends, len(data.Configuration.Frontends))
+			assert.Equal(t, numBackends, len(data.Configuration.Backends))
+			wg.Done()
+		}
+
```
timoreimann: No extra blank line here please.

provider/file/file_test.go
```diff
@@ -0,0 +1,196 @@
...
+	}()
+
+	provider.Filename = tempFile.Name()
+	provider.Watch = true
+	provider.Provide(c, safe.NewPool(context.Background()), nil)
+
+	// Wait for initial message to be tested
+	wg.Wait()
+
+	// Now test again with single frontend and backend
```
timoreimann: It looks like this should be part of a different, dedicated test.

provider/file/file_test.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+
+	}()
+
+	provider.Filename = tempFile.Name()
+	provider.Watch = true
+	provider.Provide(c, safe.NewPool(context.Background()), nil)
+
+	// Wait for initial message to be tested
+	wg.Wait()
```
timoreimann: We should define an upper bound after which we stop waiting for the asserting goroutine to complete. Otherwise, our test could lock up if we make a mistake or otherwise fail to send an event down the `ConfigMessage` channel.

The usual way to achieve this is to define a `time.Timer` of, say, 500 ms or so, and have the test `t.Fail` if it fires. If you also replace the wait group by a done channel that's being closed by the asserting goroutine as soon as it finishes, you can have a simple `select` statements over the timeout and done channels.

Similar for other tests below.

provider/file/file_test.go
```diff
@@ -0,0 +1,196 @@
...
+	wg.Add(1)
+	os.Remove(tempFile1.Name())
+	wg.Wait()
+
+}
+
+func createFile(t *testing.T, file string) *os.File {
+	f, err := os.Create(file)
+	if err != nil {
+		t.Error(err)
```
timoreimann: Should likely be a `t.Fatal`.

provider/file/file_test.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	return f
+}
+
+func createTempDir(t *testing.T, dir string) string {
+	d, err := ioutil.TempDir("", dir)
+	if err != nil {
+		t.Error(err)
```
timoreimann: Should likely be a `t.Fatal`.

provider/file/file_test.go
```diff
@@ -0,0 +1,196 @@
...
+func TestProvideSingleFile(t *testing.T) {
+	tempDir := createTempDir(t, "testsingle")
+	defer os.RemoveAll(tempDir)
+
+	tempFileName := path.Join(tempDir, "temp1.toml")
+	tempFile := createFile(t, tempFileName)
+	tempFile.WriteString(frontends + backends)
+	tempFile.Close()
+
+	provider := new(Provider)
```
timoreimann: How about using the `&Provider{...}` notation and inline the field settings from lines 107-108 directly?

provider/file/file.go
```diff
@@ -18,68 +22,135 @@ var _ provider.Provider = (*Provider)(nil)
...
 func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *safe.Pool, constraints types.Constraints) error {
-	watcher, err := fsnotify.NewWatcher()
+	var configuration *types.Configuration
+	var err error
+	var watchDir string
+
+	if p.Directory != "" {
+		watchDir = p.Directory
+		configuration, err = loadFileConfigFromDirectory(p.Directory)
+	} else {
```
timoreimann: If we have both a single filename and a directory defined, the directory wins. I wonder if that's desirable or if we should have at least log a warning, if not let Traefik fail to bootstrap during a configuration sanity check. WDYT?

And whatever we decide, we should document how Traefik deals with this situation.

provider/file/file.go
```diff
@@ -18,68 +22,135 @@ var _ provider.Provider = (*Provider)(nil)
...
+	// Process events
+	pool.Go(func(stop chan bool) {
+		defer watcher.Close()
+		for {
+			select {
+			case <-stop:
+				return
+			case evt := <-watcher.Events:
+				event = &evt
+			case <-time.After(debouncePeriod):
```
rjshep: Having the timer reset each time an event comes in **is** what I wanted to achieve.  This lets things settle - for example if an external process is re-writing the files - before sending the updated configuration via the configuration channel.  What I saw during testing was that multiple events were fired, even for a single edit, and Traefik was then trying to reload config more than needed.  To be fair, the existing logic which checks if the provided config has changed kicked in and therefore nothing was actually done with it, but I felt it better just to put the delay in at the point where we know multiple events can occur in quick succession.

provider/file/file.go
rjshep: Ideally, yes, we'd notify if both a single file and directory are specified.  However, file is always specified, as it's then name of the main Traefik config file, even if the `filename` isn't set in the `[file]` config section.  I didn't see a straightforward way to address this.

rjshep:
Thanks, as always, for the comments @timoreimann.  I've made the changes requested and look forward to @emilevauge 's views too.

provider/file/file.go
timoreimann: I see your point, it's a fair one. On the other hand, that external process could be so busy updating the file that the actual event processing case could be "starving", leaving the configuration out-of-date for possibly more than two seconds.

Admittedly, I don't know how realistic that scenario actually is. My other argument would be the deviation in behavior from the other provider watches.

I can see arguments for both directions, so I'll let @emilevauge chime in on this one.

provider/file/file_test.go
```diff
@@ -0,0 +1,180 @@
...
+	signal := make(chan interface{})
+
+	numFrontends := 2
+	numBackends := 2
+
+	go func() {
+		for {
+			data := <-c
+			assert.Equal(t, "file", data.ProviderName)
+			assert.Equal(t, numFrontends, len(data.Configuration.Frontends))
```
ldez:
```golang
assert.Len(t, data.Configuration.Frontends, numFrontends)
assert.Len(t, data.Configuration.Backends, numBackends)
```

provider/file/file.go
```diff
@@ -18,68 +22,144 @@ var _ provider.Provider = (*Provider)(nil)
...
+		return nil, fmt.Errorf("error reading configuration file: %s", err)
+	}
+	return configuration, nil
+}
+
+func loadFileConfigFromDirectory(directory string) (*types.Configuration, error) {
+	var fileList []os.FileInfo
+	var err error
+
+	if fileList, err = ioutil.ReadDir(directory); err != nil {
```
ldez: you can simplify:
```golang
fileList, err := ioutil.ReadDir(directory)

if err != nil {
// ...
```

provider/file/file.go
```diff
@@ -18,68 +22,144 @@ var _ provider.Provider = (*Provider)(nil)
...
+		return nil, fmt.Errorf("unable to read directory %s: %v", directory, err)
 	}
-	return configuration
+
+	configuration := &types.Configuration{Frontends: make(map[string]*types.Frontend),
```
ldez: you can change the format to:
```golang
configuration := &types.Configuration{
	Frontends: make(map[string]*types.Frontend),
	Backends: make(map[string]*types.Backend),
}

File: provider/file/file.go [hunk 14]
```diff
@@ -18,68 +22,144 @@ var _ provider.Provider = (*Provider)(nil)
...
+		Backends: make(map[string]*types.Backend)}
+
+	for _, file := range fileList {
+		if !strings.HasSuffix(file.Name(), ".toml") {
+			continue
+		}
+
+		var c *types.Configuration
+		if c, err = loadFileConfig(path.Join(directory, file.Name())); err != nil {
```
ldez: this can be simplify:
```golang
c, err := loadFileConfig(path.Join(directory, file.Name()))
if err != nil {
// ...
```

provider/file/file.go
```diff
@@ -18,68 +22,144 @@ var _ provider.Provider = (*Provider)(nil)
...
+			}
+		}
+
+		for k, v := range c.Frontends {
+			if _, exists := configuration.Frontends[k]; exists {
+				log.Warnf("Frontend %s already configured, skipping", k)
+			} else {
+				configuration.Frontends[k] = v
+			}
+
```
ldez: could you remove blank line ?

provider/file/file.go
```diff
@@ -18,68 +22,144 @@ var _ provider.Provider = (*Provider)(nil)
...
+		if !strings.HasSuffix(file.Name(), ".toml") {
+			continue
+		}
+
+		var c *types.Configuration
+		if c, err = loadFileConfig(path.Join(directory, file.Name())); err != nil {
+			return nil, err
+		}
+
+		for k, v := range c.Backends {
```
ldez: maybe replace `k, v` by `backendName, backend`

provider/file/file.go
```diff
@@ -18,68 +22,144 @@ var _ provider.Provider = (*Provider)(nil)
...
+			if _, exists := configuration.Backends[k]; exists {
+				log.Warnf("Backend %s already configured, skipping", k)
+			} else {
+				configuration.Backends[k] = v
+			}
+		}
+
+		for k, v := range c.Frontends {
```
ldez: maybe replace `k, v` by `frontendName, frontend`

provider/file/file_test.go
```diff
@@ -0,0 +1,180 @@
...
+			assert.Equal(t, numBackends, len(data.Configuration.Backends))
```
ldez:
```golang
assert.Len(t, data.Configuration.Frontends, numFrontends)
assert.Len(t, data.Configuration.Backends, numBackends)
```

provider/file/file.go
```diff
@@ -18,68 +22,149 @@ var _ provider.Provider = (*Provider)(nil)
...
+	var err error
+	var watchDir string
+
+	if p.Directory != "" {
+		watchDir = p.Directory
+		configuration, err = loadFileConfigFromDirectory(p.Directory)
+	} else {
+		watchDir = filepath.Dir(p.Filename)
+		configuration, err = loadFileConfig(p.Filename)
+	}
```
ldez: I think you can replace this duplicate block (here and in the ` watcherCallback`) by this method:

```golang
func (p *Provider) loadConfig() (*types.Configuration, error) {
	if p.Directory != "" {
		return loadFileConfigFromDirectory(p.Directory)
	} else {
		return loadFileConfig(p.Filename)
	}
}
```

provider/file/file.go
```diff
@@ -18,68 +21,145 @@ var _ provider.Provider = (*Provider)(nil)
...
+
 	if err != nil {
-		log.Error("Error creating file watcher", err)
 		return err
 	}
 
-	file, err := os.Open(p.Filename)
-	if err != nil {
-		log.Error("Error opening file", err)
+	if err := p.addWatcher(pool, watchDir, configurationChan, p.watcherCallback); err != nil {
```
ldez: you need to use `p.Watch` condition:
```golang
if p.Watch {
	var watchDir string

	if p.Directory != "" {
		watchDir = p.Directory
	} else {
		watchDir = filepath.Dir(p.Filename)
	}

	if err := p.addWatcher(pool, watchDir, configurationChan, p.watcherCallback); err != nil {
		return err
	}
}
```

provider/file/file_test.go
ldez: I think, you can reduce test definitions like that:

<details>
<summary>file_test.go</summary>

```golang
package file

import (
	"context"
	"fmt"
	"io/ioutil"
	"os"
	"path"
	"testing"
	"time"

	"github.com/containous/traefik/provider"
	"github.com/containous/traefik/safe"
	"github.com/containous/traefik/types"
	"github.com/stretchr/testify/assert"
)

func TestProvideSingleFile(t *testing.T) {
	tempDir := createTempDir(t, "testfile")
	defer os.RemoveAll(tempDir)

	expectedNumFrontends := 2
	expectedNumBackends := 2

	tempFile := createFile(t,
		tempDir, "simple.toml",
		createFrontendConfiguration(expectedNumFrontends),
		createBackendConfiguration(expectedNumBackends))

	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)

	fileProvider := &Provider{
		BaseProvider: provider.BaseProvider{
			Filename: tempFile.Name(),
			Watch:    true,
		},
	}
	fileProvider.Provide(configurationChan, safe.NewPool(context.Background()), nil)

	// Wait for initial message to be tested
	waitForSignal(t, signal, 2*time.Second, "initial config")

	// Now test again with single frontend and backend
	expectedNumFrontends = 1
	expectedNumBackends = 1

	tempFile = createFile(t,
		tempDir, "simple.toml",
		createFrontendConfiguration(expectedNumFrontends),
		createBackendConfiguration(expectedNumBackends))

	waitForSignal(t, signal, 2*time.Second, "single frontend and backend")
}

func TestProvideDirectory(t *testing.T) {
	tempDir := createTempDir(t, "testdir")
	defer os.RemoveAll(tempDir)

	expectedNumFrontends := 2
	expectedNumBackends := 2

	tempFile1 := createRandomFile(t, tempDir, createFrontendConfiguration(expectedNumFrontends))
	tempFile2 := createRandomFile(t, tempDir, createBackendConfiguration(expectedNumBackends))

	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)

	fileProvider := &Provider{
		BaseProvider: provider.BaseProvider{
			Watch: true,
		},
		Directory: tempDir,
	}
	fileProvider.Provide(configurationChan, safe.NewPool(context.Background()), nil)

	// Wait for initial config message to be tested
	waitForSignal(t, signal, 2*time.Second, "initial config")

	// Now remove the backends file
	expectedNumFrontends = 2
	expectedNumBackends = 0
	os.Remove(tempFile2.Name())
	waitForSignal(t, signal, 2*time.Second, "remove the backends file")

	// Now remove the frontends file
	expectedNumFrontends = 0
	expectedNumBackends = 0
	os.Remove(tempFile1.Name())
	waitForSignal(t, signal, 2*time.Second, "remove the frontends file")
}

func createConfigurationRoutine(t *testing.T, expectedNumFrontends *int, expectedNumBackends *int) (chan types.ConfigMessage, chan interface{}) {
	configurationChan := make(chan types.ConfigMessage)
	signal := make(chan interface{})

	go func() {
		for {
			data := <-configurationChan
			assert.Equal(t, "file", data.ProviderName)
			assert.Len(t, data.Configuration.Frontends, *expectedNumFrontends)
			assert.Len(t, data.Configuration.Backends, *expectedNumBackends)
			signal <- nil
		}
	}()

	return configurationChan, signal
}

func waitForSignal(t *testing.T, signal chan interface{}, timeout time.Duration, caseName string) {
	timer := time.NewTimer(timeout)
	defer timer.Stop()

	select {
	case <-signal:

	case <-timer.C:
		t.Fatalf("Timed out waiting for assertions to be tested: %s", caseName)
	}
}

// createRandomFile Helper
func createRandomFile(t *testing.T, tempDir string, contents ...string) *os.File {
	return createFile(t, tempDir, fmt.Sprintf("temp%d.toml", time.Now().UnixNano()), contents...)
}

// createFile Helper
func createFile(t *testing.T, tempDir string, name string, contents ...string) *os.File {
	fileName := path.Join(tempDir, name)

	tempFile, err := os.Create(fileName)
	if err != nil {
		t.Fatal(err)
	}

	for _, content := range contents {
		_, err := tempFile.WriteString(content)
		if err != nil {
			t.Fatal(err)
		}
	}

	err = tempFile.Close()
	if err != nil {
		t.Fatal(err)
	}

	return tempFile
}

// createTempDir Helper
func createTempDir(t *testing.T, dir string) string {
	d, err := ioutil.TempDir("", dir)
	if err != nil {
		t.Fatal(err)
	}
	return d
}

// createFrontendConfiguration Helper
func createFrontendConfiguration(n int) string {
	conf := "[frontends]\n"
	for i := 1; i <= n; i++ {
		conf += fmt.Sprintf(`  [frontends.frontend%[1]d]
  backend = "backend%[1]d"
`, i)
	}
	return conf
}

// createBackendConfiguration Helper
func createBackendConfiguration(n int) string {
	conf := "[backends]\n"
	for i := 1; i <= n; i++ {
		conf += fmt.Sprintf(`  [backends.backend%[1]d]
    [backends.backend%[1]d.servers.server1]
    url = "http://172.17.0.%[1]d:80"
`, i)
	}
	return conf
}

```
</details>

provider/file/file.go
emilevauge: @rjshep I'm not sure it's really needed knowing that indeed we have a similar global mechanism in [server.go](https://github.com/containous/traefik/blob/master/server/server.go#L238). I think we should simplify this kind of filtering and keep it one place only :)

provider/file/file_test.go
ldez: After reflections, I think you can add 2 tests (when not watch) and simplify the `provide` section:

<details>
<summary>file_test.go</summary>

```go
package file

import (
	"context"
	"fmt"
	"io/ioutil"
	"os"
	"path"
	"testing"
	"time"

	"github.com/containous/traefik/safe"
	"github.com/containous/traefik/types"
	"github.com/stretchr/testify/assert"
)

func TestProvideSingleFileAndWatch(t *testing.T) {
	tempDir := createTempDir(t, "testfile")
	defer os.RemoveAll(tempDir)

	expectedNumFrontends := 2
	expectedNumBackends := 2

	tempFile := createFile(t,
		tempDir, "simple.toml",
		createFrontendConfiguration(expectedNumFrontends),
		createBackendConfiguration(expectedNumBackends))

	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)

	provide(configurationChan, watch, withFile(tempFile))

	// Wait for initial message to be tested
	err := waitForSignal(signal, 2*time.Second, "initial config")
	assert.NoError(t, err)

	// Now test again with single frontend and backend
	expectedNumFrontends = 1
	expectedNumBackends = 1

	tempFile = createFile(t,
		tempDir, "simple.toml",
		createFrontendConfiguration(expectedNumFrontends),
		createBackendConfiguration(expectedNumBackends))

	// Must fail because we don't watch the change
	err = waitForSignal(signal, 2*time.Second, "single frontend and backend")
	assert.NoError(t, err)
}

func TestProvideSingleFileAndNotWatch(t *testing.T) {
	tempDir := createTempDir(t, "testfile")
	defer os.RemoveAll(tempDir)

	expectedNumFrontends := 2
	expectedNumBackends := 2

	tempFile := createFile(t,
		tempDir, "simple.toml",
		createFrontendConfiguration(expectedNumFrontends),
		createBackendConfiguration(expectedNumBackends))

	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)

	provide(configurationChan, withFile(tempFile))

	// Wait for initial message to be tested
	err := waitForSignal(signal, 2*time.Second, "initial config")
	assert.NoError(t, err)

	// Now test again with single frontend and backend
	expectedNumFrontends = 1
	expectedNumBackends = 1

	tempFile = createFile(t,
		tempDir, "simple.toml",
		createFrontendConfiguration(expectedNumFrontends),
		createBackendConfiguration(expectedNumBackends))

	// Must fail because we don't watch the changes
	err = waitForSignal(signal, 2*time.Second, "single frontend and backend")
	assert.Error(t, err)
}

func TestProvideDirectoryAndWatch(t *testing.T) {
	tempDir := createTempDir(t, "testdir")
	defer os.RemoveAll(tempDir)

	expectedNumFrontends := 2
	expectedNumBackends := 2

	tempFile1 := createRandomFile(t, tempDir, createFrontendConfiguration(expectedNumFrontends))
	tempFile2 := createRandomFile(t, tempDir, createBackendConfiguration(expectedNumBackends))

	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)

	provide(configurationChan, watch, withDirectory(tempDir))

	// Wait for initial config message to be tested
	err := waitForSignal(signal, 2*time.Second, "initial config")
	assert.NoError(t, err)

	// Now remove the backends file
	expectedNumFrontends = 2
	expectedNumBackends = 0
	os.Remove(tempFile2.Name())
	err = waitForSignal(signal, 2*time.Second, "remove the backends file")
	assert.NoError(t, err)

	// Now remove the frontends file
	expectedNumFrontends = 0
	expectedNumBackends = 0
	os.Remove(tempFile1.Name())
	err = waitForSignal(signal, 2*time.Second, "remove the frontends file")
	assert.NoError(t, err)
}

func TestProvideDirectoryAndNotWatch(t *testing.T) {
	tempDir := createTempDir(t, "testdir")
	defer os.RemoveAll(tempDir)

	expectedNumFrontends := 2
	expectedNumBackends := 2

	createRandomFile(t, tempDir, createFrontendConfiguration(expectedNumFrontends))
	tempFile2 := createRandomFile(t, tempDir, createBackendConfiguration(expectedNumBackends))

	configurationChan, signal := createConfigurationRoutine(t, &expectedNumFrontends, &expectedNumBackends)

	provide(configurationChan, withDirectory(tempDir))

	// Wait for initial config message to be tested
	err := waitForSignal(signal, 2*time.Second, "initial config")
	assert.NoError(t, err)

	// Now remove the backends file
	expectedNumFrontends = 2
	expectedNumBackends = 0
	os.Remove(tempFile2.Name())

	// Must fail because we don't watch the changes
	err = waitForSignal(signal, 2*time.Second, "remove the backends file")
	assert.Error(t, err)

}

func createConfigurationRoutine(t *testing.T, expectedNumFrontends *int, expectedNumBackends *int) (chan types.ConfigMessage, chan interface{}) {
	configurationChan := make(chan types.ConfigMessage)
	signal := make(chan interface{})

	go func() {
		for {
			data := <-configurationChan
			assert.Equal(t, "file", data.ProviderName)
			assert.Len(t, data.Configuration.Frontends, *expectedNumFrontends)
			assert.Len(t, data.Configuration.Backends, *expectedNumBackends)
			signal <- nil
		}
	}()

	return configurationChan, signal
}

func waitForSignal(signal chan interface{}, timeout time.Duration, caseName string) error {
	timer := time.NewTimer(timeout)
	defer timer.Stop()

	select {
	case <-signal:

	case <-timer.C:
		return fmt.Errorf("Timed out waiting for assertions to be tested: %s", caseName)
	}
	return nil
}

func provide(configurationChan chan types.ConfigMessage, builders ...func(p *Provider)) {
	pvd := &Provider{}

	for _, builder := range builders {
		builder(pvd)
	}

	pvd.Provide(configurationChan, safe.NewPool(context.Background()), nil)
}

func watch(pvd *Provider) {
	pvd.Watch = true
}

func withDirectory(name string) func(*Provider) {
	return func(pvd *Provider) {
		pvd.Directory = name
	}
}

func withFile(tempFile *os.File) func(*Provider) {
	return func(p *Provider) {
		p.Filename = tempFile.Name()
	}
}

// createRandomFile Helper
func createRandomFile(t *testing.T, tempDir string, contents ...string) *os.File {
	return createFile(t, tempDir, fmt.Sprintf("temp%d.toml", time.Now().UnixNano()), contents...)
}

// createFile Helper
func createFile(t *testing.T, tempDir string, name string, contents ...string) *os.File {
	fileName := path.Join(tempDir, name)

	tempFile, err := os.Create(fileName)
	if err != nil {
		t.Fatal(err)
	}

	for _, content := range contents {
		_, err := tempFile.WriteString(content)
		if err != nil {
			t.Fatal(err)
		}
	}

	err = tempFile.Close()
	if err != nil {
		t.Fatal(err)
	}

	return tempFile
}

// createTempDir Helper
func createTempDir(t *testing.T, dir string) string {
	d, err := ioutil.TempDir("", dir)
	if err != nil {
		t.Fatal(err)
	}
	return d
}

// createFrontendConfiguration Helper
func createFrontendConfiguration(n int) string {
	conf := "[frontends]\n"
	for i := 1; i <= n; i++ {
		conf += fmt.Sprintf(`  [frontends.frontend%[1]d]
  backend = "backend%[1]d"
`, i)
	}
	return conf
}

// createBackendConfiguration Helper
func createBackendConfiguration(n int) string {
	conf := "[backends]\n"
	for i := 1; i <= n; i++ {
		conf += fmt.Sprintf(`  [backends.backend%[1]d]
    [backends.backend%[1]d.servers.server1]
    url = "http://172.17.0.%[1]d:80"
`, i)
	}
	return conf
}

```
</details>

provider/file/file.go
```diff
@@ -18,68 +20,138 @@ var _ provider.Provider = (*Provider)(nil)
...
 	}
 
-	file, err := os.Open(p.Filename)
+	if p.Watch {
+		var watchDir string
+
+		if p.Directory != "" {
+			watchDir = p.Directory
+		} else {
+			watchDir = filepath.Dir(p.Filename)
```
nmengin: By adding watcher to the directory even if you just want to wach a file, you are going to watch all the directory events.
For example, if it exists 10 toml files into your directory, you are going to reload the configuration when all of these files will be modified.

That's why I suggest you to just add a watcher for the configuration file : `watchDir = p.Filename` and, maybe rename `watchDir` into `watchFile` (dir is a file ;) ) or `watchItem`...

rjshep: If you just want to watch a file, then you'd just specify the `Filename` and not use `Directory` at all.  So that `Watch` applies only to the active setting `Filename` or `Directory` - they are mutually exclusive.

nmengin: @rjshep Of course, but when user specifies `FileName`, the watcher is added on its parent directory (`filepath.Dir(p.Filename)`). It was the mechanism before your PR.

However, before the modification the callback checked if the events received by the watcher were sent by the configuration file as described below : 

```go
case event := <-watcher.Events:
    if strings.Contains(event.Name, file.Name())
```
It would have been better to add directly the watcher on the file but, thanks to this code, the configuration was reloaded only when the `Filename` was modified.
But now, the configuration will be reloaded if one of all the toml file contained in the same directory than the `Filename`. 
There will have no problem because only the `Filename` configuration will be reloaded but, it should be better to take advantage of you PR to improve the `Filename` watching by replacing `watchDir = filepath.Dir(p.Filename)` by `watchDir = p.Filename`.

Then, the configuration will only be reloaded when the `Filename` will change.

integration/fixtures/file/directory.toml
```diff
@@ -0,0 +1,10 @@
+defaultEntryPoints = ["http"]
+
+[entryPoints]
+  [entryPoints.http]
+  address = ":8000"
+
+logLevel = "DEBUG"
```
nmengin: The `logLevel` should to be defined before the `entrypoints` otherwise the logger will not work correctly.

provider/file/file.go
rjshep: Sorry @nmengin, completely missed your original point.  Yes that is a sensible change to make.

ldez:
@rjshep could you squqh and rebase (for semaphoreCI)?

PR #1675, opened 2017-05-26
Custom Error Pages

bparli:
### Description

Follow up to [PR 1576](https://github.com/containous/traefik/pull/1576) and [issue 1634](https://github.com/containous/traefik/issues/1634).

Following the above design and discussion:
- New ErrorPage type to capture frontend error pages configurations
- New error_pages middleware containing a response recorder
- Documentation/example
- Unit tests to ensure middleware is being initialized and rendering correctly
- Basic integration tests


nmengin:
@bparli, your PR seems good for me.

But, I believe the [issue 1634](https://github.com/containous/traefik/issues/1634) forgets an Use Case : how to specify an error backend for only one error code?
Of course, it's possible to set a range of one code [404-404] but it's a bit ugly. It would be better to allow single host by using configuration as below :

```toml
 [frontends]
   [frontends.website]
   backend = "website"
   [errors]
     [error.network]
     status = "404"
     backend = "error"
     query = "/{status}.html"
   [frontends.website.routes.website]
   rule = "Host: website.mydomain.com"

 [backends]
   [backends.website]
     [backends.website.servers.website]
     url = "https://1.2.3.4"
   [backends.error]
     [backends.error.servers.error]
     url = "http://2.3.4.5"
```
Moreover, the best solution may be to specify a list of single code(s) and range (as printed page into printer windows) and configure it as below :

```toml
 [frontends]
   [frontends.website]
   backend = "website"
   [errors]
     [error.network]
     status = "401,404,[500-600]"
     backend = "error"
     query = "/{status}.html"
   [frontends.website.routes.website]
   rule = "Host: website.mydomain.com"

 [backends]
   [backends.website]
     [backends.website.servers.website]
     url = "https://1.2.3.4"
   [backends.error]
     [backends.error.servers.error]
     url = "http://2.3.4.5"
```
I guess the second suggestion is a little bit harder to develop.
It can be good to allow, at least, configuring a single code in a first time and maybe open a PR to improve the behavior with the list as described in the second solution.

@emile, @bparli, what is your mind about this?

bparli:
Thanks for taking a look @nmengin 

The concern you raise did actually cross my mind, but I stuck to the original design and made sure the one-off HTTP code was still covered (albeit not very aesthetically as you point out).  Also, I felt the one-off HTTP code may not be a common case; I've usually seen, or configured myself, for a range (or even an entire class) of codes.  Of course, you all have a good vantage point so I can work on addressing your concern if you feel its helpful or necessary.

Also, did you mean to ask @emilevauge above instead of that other person?

emilevauge:
@containous/traefik do you think we should implement the _one-off HTTP code_ on top of ranges? IMHO, ranges only may be OK at first.

nmengin:
@bparli @emilevauge , I am OK with you.

LGTM

middlewares/error_pages.go
```diff
@@ -0,0 +1,73 @@
...
+	recorder.responseWriter = w
+	next.ServeHTTP(recorder, req)
+
+	//check the recorder code against the configured http status code ranges
+	for _, block := range ep.HTTPCodeRanges {
+		if recorder.Code >= block[0] && recorder.Code <= block[1] {
+			log.Debugf("Caught HTTP Status Code %d, returning error page", recorder.Code)
+			w.WriteHeader(recorder.Code)
+			if newReq, err := http.NewRequest(http.MethodGet, ep.BackendURL, nil); err != nil {
+				w.Write([]byte(http.StatusText(recorder.Code)))
```
emilevauge: We may log `err` here

docs/basics.md
```diff
@@ -396,6 +396,30 @@ Here is an example of backends and servers definition:
...
+
+```toml
+[frontends]
+  [frontends.website]
+  backend = "website"
+  [errors]
+    [error.network]
+    status = ["500-599"]
+    backend = "error"
+    query = "/500.html"
```
emilevauge: `/{status}.html` ?

integration/error_pages_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	frontendReq, err := http.NewRequest(http.MethodGet, "http://127.0.0.1:80", nil)
+	c.Assert(err, checker.IsNil)
+	frontendReq.Host = "test.local"
+
+	err = try.Request(frontendReq, 2*time.Second, try.BodyContains("nginx"))
+	c.Assert(err, checker.IsNil)
+}
+
+func (ep *ErrorPagesSuite) TestErrorPage(c *check.C) {
```
emilevauge: It would be even better to test using a non empty query IMHO

bparli: good point

docs/basics.md
bparli: It seems I originally misunderstood this point.  I like it though and will extend to include the {status} query

middlewares/error_pages.go
```diff
@@ -0,0 +1,75 @@
...
+	recorder := newRetryResponseRecorder()
+	recorder.responseWriter = w
+	next.ServeHTTP(recorder, req)
+
+	w.WriteHeader(recorder.Code)
+	//check the recorder code against the configured http status code ranges
+	for _, block := range ep.HTTPCodeRanges {
+		if recorder.Code >= block[0] && recorder.Code <= block[1] {
+			log.Errorf("Caught HTTP Status Code %d, returning error page", recorder.Code)
+			re := regexp.MustCompile("{status}")
```
Juliens: Maybe we can allocate the compiled regex in the construction of ErrorPagesHandler ?

middlewares/error_pages.go
```diff
@@ -0,0 +1,75 @@
...
+	//Break out the http status code ranges into a low int and high int
+	//for ease of use at runtime
+	var blocks [][2]int
+	for _, block := range errorPage.Status {
+		codes := strings.Split(block, "-")
+		lowCode, err := strconv.Atoi(codes[0])
+		if err != nil {
+			return nil, err
+		}
+		highCode, err := strconv.Atoi(codes[1])
```
Juliens: If we haven't "-", it panic here with out of range
Can you handle the case when there is no "-" (with a good error)
And add a test for this

thx

middlewares/error_pages.go
ldez: I think you don't be a regew:
```golang
finalURL := strings.Replace(vli, "{status}", "400", -1)
```

bparli: Thanks @Juliens, good point

and you're right @ldez, strings package will do the trick here

middlewares/error_pages.go
bparli: I think there are actually two options here, both simple.  1) catch it and log an error or 2) catch it and assume the user means to serve the error page for just the single code.  In this case the single entry can still be used to create the correct array for checking at runtime.  So, for example if ["503"] (instead of ["503-503"]) is configured,  the array ["503", "503"] is created and we continue on.  

This was actually raised by @nmengin above, I should have just addressed it then 

WDYT @Juliens ?

Juliens: I agree that handle the "no range status code" would be better, but a log could be enought

server/server.go
```diff
@@ -772,6 +772,20 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 							backendsHealthcheck[frontend.Backend] = healthcheck.NewBackendHealthCheck(*hcOpts)
 						}
 					}
+
+					if len(frontend.Errors) > 0 {
+						for _, errorPage := range frontend.Errors {
+							if configuration.Backends[errorPage.Backend] != nil && configuration.Backends[errorPage.Backend].Servers["error"].URL != "" {
```
Juliens: Could you add a log if the backend doesn't exist and/or if the URL is empty

bparli:
Thanks everyone for the reviews and feedback.  I think its all addressed to this point

ldez:
It's time to squash and rebase 😃

docs/basics.md
```diff
@@ -396,6 +396,39 @@ Here is an example of backends and servers definition:
...
+
+Custom error pages can be returned, in lieu of the default, according to frontend-configured ranges of HTTP Status codes.
+In the example below, if a 503 status is returned from the frontend "website", the custom error page at http://2.3.4.5/503.html is returned with the actual status code set in the HTTP header.
+Note, the 503.html page itself is not hosted on traefik, but some other infrastructure.   
+
+```toml
+[frontends]
+  [frontends.website]
+  backend = "website"
+  [errors]
```
pjeby: This bit of documentation looks wrong: the test fixture that resembles this has `[frontends.frontend1.errors]`
`[frontends.frontend1.errors.networks]`, but the doc here just says `[errors]` and `[errors.networks]`.

timoreimann: Thanks for noticing! Could you please file a new issue or ideally submit a PR? This one has been closed a while ago.

Thank you.

PR #1679, opened 2017-05-28
Enhance rules tests.

ldez:
### Description

Homogenization of the tests.
The naming of cases is not perfect for `stripPrefixRegex_test` but it is more maintainable.

Use `r.Header.Add("MY-HEADER", value)` instead of  `r.Header["MY-HEADER"] = []string{value}`.

Bonus:
- we can use Git branch name with `/` like `foo/bar`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ BIND_DIR := "dist"
 TRAEFIK_MOUNT := -v "$(CURDIR)/$(BIND_DIR):/go/src/github.com/containous/traefik/$(BIND_DIR)"
 
 GIT_BRANCH := $(subst heads/,,$(shell git rev-parse --abbrev-ref HEAD 2>/dev/null))
-TRAEFIK_DEV_IMAGE := traefik-dev$(if $(GIT_BRANCH),:$(GIT_BRANCH))
+TRAEFIK_DEV_IMAGE := traefik-dev$(if $(GIT_BRANCH),:$(subst /,-,$(GIT_BRANCH)))
 REPONAME := $(shell echo $(REPO) | tr '[:upper:]' '[:lower:]')
 TRAEFIK_IMAGE := $(if $(REPONAME),$(REPONAME),"containous/traefik")
 INTEGRATION_OPTS := $(if $(MAKE_DOCKER_HOST),-e "DOCKER_HOST=$(MAKE_DOCKER_HOST)", -v "/var/run/docker.sock:/var/run/docker.sock")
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -16,7 +16,7 @@ BIND_DIR := "dist"
 TRAEFIK_MOUNT := -v "$(CURDIR)/$(BIND_DIR):/go/src/github.com/containous/traefik/$(BIND_DIR)"
 
 GIT_BRANCH := $(subst heads/,,$(shell git rev-parse --abbrev-ref HEAD 2>/dev/null))
-TRAEFIK_DEV_IMAGE := traefik-dev$(if $(GIT_BRANCH),:$(GIT_BRANCH))
+TRAEFIK_DEV_IMAGE := traefik-dev$(if $(GIT_BRANCH),:$(subst /,-,$(GIT_BRANCH)))
```
timoreimann: Nice, this one has bothered me before!

middlewares/replace_path_test.go
```diff
@@ -17,28 +17,22 @@ func TestReplacePath(t *testing.T) {
...
 
 			handler.ServeHTTP(nil, req)
-			if newPath != replacementPath {
-				t.Fatalf("new path should be '%s'", replacementPath)
-			}
-
-			if oldPath != path {
-				t.Fatalf("old path should be '%s'", path)
-			}
+			assert.Equal(t, replacementPath, expectedPath, "%s: unexpected path.", path)
```
timoreimann: [Expected must go before actual](https://godoc.org/github.com/stretchr/testify/assert#Equal) or otherwise the labeling will be wrong in `assert.Equal`'s output.

middlewares/stripPrefixRegex_test.go
```diff
@@ -1,55 +1,99 @@
...
-		}
-		response, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			t.Fatal(err)
-		}
-		if test.expectedResponse != string(response) {
-			t.Errorf("Expected '%s' :  '%s'\n", test.expectedResponse, response)
-		}
+		test := test
+		t.Run(test.desc, func(t *testing.T) {
```
timoreimann: For the test cases at hand, it would also suffice to pass `path` to `t.Run` as it's disambiguating already, and the `desc`riptions aren't overly -- descriptive. :)

ldez:
@timoreimann It's fixed!

Issue #1306, opened 2017-03-16
send traefik access logs to stdout


r0bj:
traefik: 1.1.2

I didn't find a way to send access logs `accessLogsFile` to stdout using official docker image.

It would be useful to have something like:
```
accessLogsFile = stdout
```

Or similar to `traefikLogsFile` - if not defined, send logs to stdout.

Using docker It's recommended to send all logs to stdout so that some collector could gather them, aggregate, etc.

Regner:
I make no promises but I am at least looking into what this would take. Started asking some questions on Slack. Will let you know if I make any progress.

marco-jantke:
We also need this feature in order to get Traefik production in our environment, therefore I will have a look at the issue and try to create a PR soon for it.

ldez:
A huge refactor of the access logger,  see https://github.com/containous/traefik/pull/1485

marco-jantke:
@ldez thank you for the hint. Honestly I think it makes more sense to await that PR and then to add another one adding the stdout logging capability. Otherwise it could be work for the 🗑

argeiger:
@r0bj, you can do the following:

mkfifo -m 600 /tmp/logpipe
cat <> /tmp/logpipe 1>&2 &
chown traefik:traefik /tmp/logpipe

Then in your toml file point accesslogs to /tmp/logpipe.  I know this is a little dirty but it works.  Would still like a true implementation vs. workaround.

timoreimann:
Fixed by #1683.

PR #1683, opened 2017-05-30
enable logging to stdout for access logs

marco-jantke:
This PR enables sending the access logs to the stdout. It should therefore solve #1306. Consistently to `TraefikLogsFile`, it uses stdout now when the config for `AccessLogsFile` is omitted or empty.

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -154,10 +154,15 @@ Supported filters:
 
 ## Access log definition
 
-The standard access log uses the textual Common Log Format (CLF), extended with additional fields.
-Alternatively logs can be written in JSON. 
-Using the default CLF option is simple, e.g.
+Access logs are written when `[accessLog]` is defined. 
+By default it will write to stdout and produce logs in the textual Common Log Format (CLF), extended with additional fields.
 
+To enable access logs using the default settings just add the `[accessLog]` entry.
+```toml
+[accessLog]
+```
+
+To write the logs into a logfile specify the `filePath`.
 ```toml
 [accessLog]
   filePath = "/path/to/access.log"
--- a/middlewares/accesslog/logger.go
+++ b/middlewares/accesslog/logger.go
@@ -2,7 +2,6 @@ package accesslog
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net"
 	"net/http"
@@ -38,19 +37,13 @@ type LogHandler struct {
 
 // NewLogHandler creates a new LogHandler
 func NewLogHandler(config *types.AccessLog) (*LogHandler, error) {
-	if len(config.FilePath) == 0 {
-		return nil, errors.New("Empty file path specified for accessLogsFile")
-	}
-
-	dir := filepath.Dir(config.FilePath)
-
-	if err := os.MkdirAll(dir, 0755); err != nil {
-		return nil, fmt.Errorf("failed to create log path %s: %s", dir, err)
-	}
-
-	file, err := os.OpenFile(config.FilePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0664)
-	if err != nil {
-		return nil, fmt.Errorf("error opening file: %s %s", dir, err)
+	file := os.Stdout
+	if len(config.FilePath) > 0 {
+		f, err := openAccessLogFile(config.FilePath)
+		if err != nil {
+			return nil, fmt.Errorf("error opening access log file: %s", err)
+		}
+		file = f
 	}
 
 	var formatter logrus.Formatter
@@ -71,6 +64,21 @@ func NewLogHandler(config *types.AccessLog) (*LogHandler, error) {
 		Level:     logrus.InfoLevel,
 	}
 	return &LogHandler{logger: logger, file: file}, nil
+}
+
+func openAccessLogFile(filePath string) (*os.File, error) {
+	dir := filepath.Dir(filePath)
+
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return nil, fmt.Errorf("failed to create log path %s: %s", dir, err)
+	}
+
+	file, err := os.OpenFile(filePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0664)
+	if err != nil {
+		return nil, fmt.Errorf("error opening file %s: %s", filePath, err)
+	}
+
+	return file, nil
 }
 
 // GetLogDataTable gets the request context object that contains logging data. This accretes
--- a/middlewares/accesslog/logger_test.go
+++ b/middlewares/accesslog/logger_test.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"testing"
 
 	"github.com/containous/traefik/types"
@@ -18,9 +19,8 @@ import (
 )
 
 var (
-	logger            *LogHandler
 	logFileNameSuffix       = "/traefik/logger/test.log"
-	helloWorld              = "Hello, World"
+	testContent             = "Hello, World"
 	testBackendName         = "http://127.0.0.1/testBackend"
 	testFrontendName        = "testFrontend"
 	testStatus              = 123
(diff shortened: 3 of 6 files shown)


middlewares/accesslog/logger_test.go
```diff
@@ -68,10 +69,55 @@ func TestLogger(t *testing.T) {
...
+
+func TestNewLogHandlerOutputStdout(t *testing.T) {
+	file, err := ioutil.TempFile("", "testlogger")
+	if err != nil {
+		t.Fatalf("failed to create temp file: %s", err)
+	}
+	defer file.Close()
+	// capture stdout in a file
+	stdOut := os.Stdout
+	os.Stdout = file
```
timoreimann: Ideally, it'd be nice to use a memory-mapped file. I did a quick search and couldn't find an easy solution, however.

middlewares/accesslog/logger_test.go
```diff
@@ -68,10 +69,55 @@ func TestLogger(t *testing.T) {
...
+		RemoteAddr: fmt.Sprintf("%s:%d", testHostname, testPort),
+		URL: &url.URL{
+			User: url.UserPassword(testUsername, ""),
+			Path: testPath,
+		},
+	}
+
+	logger.ServeHTTP(&logtestResponseWriter{}, req, LogWriterTestHandlerFunc)
+
+	written, err := ioutil.ReadFile(file.Name())
```
timoreimann: We should probably `require.NoError` on `err`?

middlewares/accesslog/logger_test.go
```diff
@@ -68,10 +69,55 @@ func TestLogger(t *testing.T) {
...
+	assert.NotZero(t, len(written), "expected access log message on stdout")
```
timoreimann: `require` maybe? Does `assertValidLogData` make sense if the written length is zero?

middlewares/accesslog/logger_test.go
```diff
@@ -92,7 +138,7 @@ func printLogdata(logdata []byte) string {
 	return fmt.Sprintf(
```
timoreimann: It looks like this part will be executed even in the case where all assertions succeed. Do we want that?

middlewares/accesslog/logger_test.go
```diff
@@ -82,7 +128,7 @@ func TestLogger(t *testing.T) {
 		assert.Equal(t, fmt.Sprintf("%d", len(helloWorld)), tokens[7], printLogdata(logdata))
```
timoreimann: I find the part up to and include this line somewhat hard to read. Maybe a combination of a `require`s and a switch statement could improve readability? WDYT?

middlewares/accesslog/logger.go
```diff
@@ -31,10 +30,28 @@ type LogHandler struct {
...
 // NewLogHandler creates a new LogHandler
 func NewLogHandler(filePath string) (*LogHandler, error) {
-	if len(filePath) == 0 {
-		return nil, errors.New("Empty file name specified for accessLogsFile")
+	var file *os.File
+
+	if len(filePath) > 0 {
+		f, err := openAccessLogFile(filePath)
+		if err != nil {
+			return nil, err
```
marco-jantke: Good point, I added some specific error message that adds more context to the previous one.

middlewares/accesslog/logger.go
```diff
@@ -31,10 +30,28 @@ type LogHandler struct {
...
+		}
+		file = f
+	} else {
+		file = os.Stdout
```
marco-jantke: Done.

middlewares/accesslog/logger_test.go
```diff
@@ -68,10 +69,55 @@ func TestLogger(t *testing.T) {
 
 	logger.ServeHTTP(&logtestResponseWriter{}, req, LogWriterTestHandlerFunc)
 
-	if logdata, err := ioutil.ReadFile(logfilePath); err != nil {
+	if logdata, err := ioutil.ReadFile(logfilePath); err == nil {
+		assertValidLogData(t, logdata)
+	} else {
 		fmt.Printf("%s\n%s\n", string(logdata), err.Error())
```
marco-jantke: I compared the messages a bit with and without. I dropped it then completely as it doesn't really add value. The logdata will always be empty given there is an error and err infos we get anyway using `require.NoError`.

middlewares/accesslog/logger_test.go
```diff
@@ -68,10 +69,55 @@ func TestLogger(t *testing.T) {
...
 		assert.NoError(t, err)
```
marco-jantke: Ack, done.

middlewares/accesslog/logger_test.go
```diff
@@ -68,10 +69,55 @@ func TestLogger(t *testing.T) {
...
-	} else if tokens, err := shellwords.Parse(string(logdata)); err != nil {
+	}
+}
+
+func TestNewLogHandlerOutputStdout(t *testing.T) {
+	file, err := ioutil.TempFile("", "testlogger")
+	if err != nil {
+		t.Fatalf("failed to create temp file: %s", err)
```
marco-jantke: I cleaned up the whole test case to use `testify` completely.

middlewares/accesslog/logger_test.go
marco-jantke: I also checked and saw no "native" way of doing this, whereas the current approach with `TempFile` is included. Therefore I would stick with the current solution.

middlewares/accesslog/logger_test.go
marco-jantke: Good catch, I was a bit surprised by the compiler does not complain even though we do not use err anymore after this call. The reason is clear though, its already initialised and used before.

middlewares/accesslog/logger_test.go
marco-jantke: Nope it doesn't, good catch.

middlewares/accesslog/logger_test.go
```diff
@@ -68,10 +69,55 @@ func TestLogger(t *testing.T) {
...
+	}
+
+	logger.ServeHTTP(&logtestResponseWriter{}, req, LogWriterTestHandlerFunc)
+
+	written, err := ioutil.ReadFile(file.Name())
+	assert.NotZero(t, len(written), "expected access log message on stdout")
+	assertValidLogData(t, written)
+
+	// restore original stdout
+	os.Stdout = stdOut
```
marco-jantke: I added a util function to the test `mustCaptureStdout` giving us a `restoreStdout` function that I can use in a `defer` now. Good suggestion, brought me to this small refactoring.

middlewares/accesslog/logger_test.go
marco-jantke: Changed it to use `require` and its much clearer imho.

ldez:
@marco-jantke can you resolve conflicts?

middlewares/accesslog/logger_test.go
marco-jantke: Hmm its the error message we have to provide in a "static" way to the assert/require invocations. Therefore we have to do it at least once for each test case. I improved the message a bit and inlined the formatting into the `requireValidLogData`, so that we only do it once :)

marco-jantke:
@ldez thanks for the notice, done.
@timoreimann thanks for the review. I added the feedback, though I had to do a rebase in order to resolve the conflicts.

docs/toml.md
```diff
@@ -155,18 +156,18 @@ Supported filters:
...
 
 The standard access log uses the textual Common Log Format (CLF), extended with additional fields.
-Alternatively logs can be written in JSON. 
+Alternatively logs can be written in JSON.
 Using the default CLF option is simple, e.g.
 
 ```toml
 [accessLog]
-  filePath = "/path/to/access.log"
+  filePath = "/path/to/access.log" # If empty, logs to stdout
```
ldez: could you remove this comment? Prefer add a line above.

docs/toml.md
```diff
@@ -155,18 +156,18 @@ Supported filters:
...
 ```
 
 To write JSON format logs, specify `json` as the format:
 ```toml
 [accessLog]
-  filePath   = "/path/to/access.log"
+  filePath   = "/path/to/access.log" # If empty, logs to stdout
```
ldez: could you remove this comment? Prefer add a line above.

docs/toml.md
```diff
@@ -41,6 +41,7 @@
 # traefikLogsFile = "log/traefik.log"
 
 # Access logs file
+# If not defined, logs to stdout
```
ldez: this option  is deprecated, you need to add the comment in the new option.

middlewares/accesslog/logger_test.go
```diff
@@ -165,44 +160,70 @@ func TestLoggerJSON(t *testing.T) {
...
 }
 
-func contains(values []string, value string) bool {
-	for _, v := range values {
-		if value == v {
-			return true
-		}
-	}
-	return false
+func requireValidLogData(t *testing.T, logData []byte) {
```
ldez: I prefer `assert` to `require` in this case.

docs/toml.md
marco-jantke: Yes, done.

docs/toml.md
marco-jantke: Yes, done.

docs/toml.md
marco-jantke: Ok, done.

marco-jantke:
@ldez thats exactly the behaviour we have now. Improved docs to match your description.

middlewares/accesslog/logger_test.go
marco-jantke: Having `require` in the function name makes it clear that a function call to it might abort the test evaluation. This is something the function name should communicate for me and guards against unexpected behaviour. WDYT?

ldez: I understand but:
- it's your main assertion. For me, you don't need to mimic Testify in this case.
- this method contains a mix of require and assert, you cannot guarantee it might abort.

marco-jantke: Ok, I renamed it.

docs/toml.md
```diff
@@ -155,11 +155,16 @@ Supported filters:
 ## Access log definition
 
 The standard access log uses the textual Common Log Format (CLF), extended with additional fields.
-Alternatively logs can be written in JSON. 
-Using the default CLF option is simple, e.g.
+Alternatively logs can be written in JSON.
+By default logs will be written to stdout and use the CLF format.
```
Juliens: Can you change this part in order to understand that without the `[accessLog]` section, we will not have any accesslog ?

Moreover can you add the same information in the `traefik.sample.toml`.

server/configuration.go
```diff
@@ -45,7 +45,7 @@ type GlobalConfiguration struct {
 	CheckNewVersion           bool                    `description:"Periodically check if a new version has been released"`
 	AccessLogsFile            string                  `description:"(Deprecated) Access logs file"` // Deprecated
 	AccessLog                 *types.AccessLog        `description:"Access log settings"`
-	TraefikLogsFile           string                  `description:"Traefik logs file"`
+	TraefikLogsFile           string                  `description:"Traefik logs file. Stdout is used when omitted or empty"`
```
Juliens: This information must be in AccessLog.FilePath description ?

marco-jantke: I added it here by intend, as this is actually the behaviour of the `TraefikLogsFile` and as it wasn't documented yet, I added this comment here as well. Nevertheless, I also extended the documentation in `AccessLog.FilePath`. Good point.

ldez:
@marco-jantke it's time to squash 😃

docs/toml.md
marco-jantke: I rephrased the configuration a bit to make it hopefully more clear and easier to read. Also I added and adjusted documentation in the `traefik.sample.toml` file. Good catch, didn't think about this one..

PR #1693, opened 2017-05-31
Exported getSubDomain function from Marathon provider

aantono:
When creating custom template for Marathon provider, it really helps to have a `getSubDomain` function be available to use in custom templates.  Otherwise the logic to compute the sub-domain directly inside the template gets very messy, and reusing the same logic as the provider would help with consistency.


timoreimann:
Sounds fair to me. Any concerns, @vdemeester or @emilevauge?

timoreimann:
@vdemeester can you formally approve?

timoreimann:
@aantono in the meantime, could you update the branch?

aantono:
Done

Issue #1514, opened 2017-04-30
Traefik doesn't `hot` detect newly added containers when using the consul_catalog provider


makeittotop:
### Do you want to request a *feature* or report a *bug*?
Bug. 

### What did you do?
Up-scaling the number of web-app docker containers via Nomad, traefik picks up the changes from the consul catalog only after a restart.

On the other hand, down-scaling changes are picked up immediately. 

I wonder what exactly am I doing wrong? 

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Using traefik ver1.2.3 (latest) with consul_catalog + nomad as  container orchestrator.

```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
debug = true
checkNewVersion = false
traefikLogsFile = "log/traefik.log"
accessLogsFile = "log/access.log"
logLevel = "DEBUG"
ProvidersThrottleDuration = 5
defaultEntryPoints = ["http"]

[consulCatalog]
  endpoint = "192.168.39.203:8500"
  constraints = ["tag==public"]

[entryPoints]
  [entryPoints.http]
    address = ":80"

[web]
  address = ":8080"

[web.statistics]
   RecentErrors = 10

[consul]
  endpoint = "192.168.39.203:8500"
  watch = true
  prefix = "traefik"
```

### If applicable, please paste the log output in debug mode (`--debug` switch)

```
INFO[0062] Skipping same configuration for provider consul_catalog
```

ldez:
Could you add more the full log?

makeittotop:
It took a while but here are the logs  
[traefik.txt](https://github.com/containous/traefik/files/1009249/traefik.txt)

makeittotop:
ERRO[0000] Error opening fileopen log/access.log: no such file or directory
ERRO[0000] Error opening fileopen log/traefik.log: no such file or directory
INFO[0000] Traefik version v1.2.3 built on 2017-04-13_07:21:10PM
INFO[0000] Using TOML configuration file /usr/home/apareek/dev/ruby/my-project/traefik/traefik.toml
[...]

Issue #1107, opened 2017-02-02
not reloading config on consul_catalog health check changes


tyler-johnson:
### What version of Traefik are you using (`traefik version`)?

v1.1.2

### What is your environment & configuration (arguments, toml...)?

Traefik is running as a normal process in Ubuntu 16.10. It is connected to a local Consul client agent which is connected to a Consul/Nomad cluster.

```bash
traefik -c /etc/traefik/traefik.toml
```

```ini
logLevel = "INFO"

[web]
address = ":8080"

[consul]
endpoint = "127.0.0.1:8500"
watch = true
prefix = "traefik"

[consulCatalog]
endpoint = "127.0.0.1:8500"
domain = "consul.localhost"
prefix = "traefik"
```

### What did you do?

I pushed a Nomad job with some Consul tags for Traefik. The Nomad job has a health check that is initially failing, but passes once the service is fully online. Consul registers the service correctly, with the Traefik tag and passing health checks eventually.

### What did you expect to see?

Traefik to update when the health check passes and attach some consul services to frontends.

### What did you see instead?

Sometimes it goes through, other times it does not. Traefik will still give a 404 and the web ui does not show my new service. I have also waited a decent amount of time and nothing happens. If I push a different Nomad job, Traefik will update with the missing service. When I totally disable health checks on a Consul service, Traefik will update properly.

vholovko:
Hey,

I have investigated a bit the current situation.

Basically consul_catalog provider is relying on blocking queries to `/v1/catalog/services` consul endpoint.
https://www.consul.io/docs/agent/http/catalog.html#catalog_services
The query itself might look like this one:

GET /v1/catalog/services?index=108026&wait=15000ms

where traefik config change triggers each time whenever `X-Consul-Index` is changed.

A `tcpdump` within consul container looks like this
```
# tcpdump -A -l -i eth2 | grep "?index="
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on eth2, link-type EN10MB (Ethernet), capture size 262144 bytes
.......qGET /v1/catalog/services?index=108026&wait=15000ms HTTP/1.1
........GET /v1/catalog/services?index=108026&wait=15000ms HTTP/1.1
GET /v1/catalog/services?index=108033&wait=15000ms HTTP/1.1
..~ ..~ GET /v1/catalog/services?index=108033&wait=15000ms HTTP/1.1
```

[...]

PR #1694, opened 2017-06-01
Speeding up consul catalog health change detection

vholovko:
Implemented long polling for services and for the health in the separate co-routines and react on either trigger independently by using same channel for both.

Basically two things improved:
1. Decreased Health change detection time.
2. Reduced amount of false triggering for traefik configuration rebuild by doing extra checking for actually changed services.

Detailed description: https://github.com/containous/traefik/issues/1107#issuecomment-300492280

Improvement for #1107  
Relates to #1514

### DIFF
--- a/provider/consul/consul_catalog.go
+++ b/provider/consul/consul_catalog.go
@@ -76,18 +76,25 @@ func (a nodeSorter) Less(i int, j int) bool {
 	return lentr.Service.Port < rentr.Service.Port
 }
 
-func (p *CatalogProvider) watchServices(stopCh <-chan struct{}) <-chan map[string][]string {
-	watchCh := make(chan map[string][]string)
+func getChangedKeys(currState map[string][]string, prevState map[string][]string) ([]string, []string) {
+	currKeySet := fun.Set(fun.Keys(currState).([]string)).(map[string]bool)
+	prevKeySet := fun.Set(fun.Keys(prevState).([]string)).(map[string]bool)
 
-	catalog := p.client.Catalog()
+	addedKeys := fun.Difference(currKeySet, prevKeySet).(map[string]bool)
+	removedKeys := fun.Difference(prevKeySet, currKeySet).(map[string]bool)
+
+	return fun.Keys(addedKeys).([]string), fun.Keys(removedKeys).([]string)
+}
+
+func (p *CatalogProvider) watchHealthState(stopCh <-chan struct{}, watchCh chan<- map[string][]string) {
 	health := p.client.Health()
-	var lastHealthIndex uint64
+	catalog := p.client.Catalog()
 
 	safe.Go(func() {
-		defer close(watchCh)
+		// variable to hold previous state
+		var flashback map[string][]string
 
-		catalogOptions := &api.QueryOptions{WaitTime: DefaultWatchWaitTime}
-		healthOptions := &api.QueryOptions{}
+		options := &api.QueryOptions{WaitTime: DefaultWatchWaitTime}
 
 		for {
 			select {
@@ -96,16 +103,10 @@ func (p *CatalogProvider) watchServices(stopCh <-chan struct{}) <-chan map[strin
 			default:
 			}
 
-			data, catalogMeta, err := catalog.Services(catalogOptions)
-			if err != nil {
-				log.WithError(err).Error("Failed to list services")
-				return
-			}
-
 			// Listening to changes that leads to `passing` state or degrades from it.
 			// The call is used just as a trigger for further actions
 			// (intentionally there is no interest in the received data).
-			_, healthMeta, err := health.State("passing", healthOptions)
+			_, meta, err := health.State("passing", options)
 			if err != nil {
 				log.WithError(err).Error("Failed to retrieve health checks")
 				return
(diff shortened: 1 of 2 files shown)


provider/consul/consul_catalog.go
```diff
@@ -298,17 +384,44 @@ func (p *CatalogProvider) getNodes(index map[string][]string) ([]catalogUpdate,
...
+		close(out)
+	}()
+	return out
+}
+
 func (p *CatalogProvider) watch(configurationChan chan<- types.ConfigMessage, stop chan bool) error {
 	stopCh := make(chan struct{})
-	serviceCatalog := p.watchServices(stopCh)
+
+	watchCh := merge(p.watchHealthState(stopCh), p.watchCatalogServices(stopCh))
```
keis: Rather than doing a merge make the channel a parameter of both watch* and send the same one to both.

provider/consul/consul_catalog.go
```diff
@@ -93,33 +95,117 @@ func (p *CatalogProvider) watchServices(stopCh <-chan struct{}) <-chan map[strin
...
+			if data != nil {
+				// Exclude node level checks
+				checks := fun.Filter(func(check *api.HealthCheck) bool {
+					return check.ServiceName != ""
+				}, data).([]*api.HealthCheck)
+
+				// A critical note is that the return of a blocking request is no guarantee of a change.
+				// It is possible that there was an idempotent write that does not affect the result of the query.
+				services := fun.Set(fun.Map(func(check *api.HealthCheck) string {
+					return check.ServiceName
```
keis: By only looking at the service name wouldn't that mean that health changes in 1 node in a bigger pool would not trigger a message. Or maybe I'm confused what ServiceName is here.

provider/consul/consul_catalog.go
vholovko: Thanks, good catch!
Corrected.

provider/consul/consul_catalog.go
vholovko: I was just filtering only services with real names without `consul` agents. But now I adjusted the code to work only with Catalog Services response. Should be more cleaner approach now.

provider/consul/consul_catalog_test.go
```diff
@@ -478,3 +479,146 @@ func TestConsulCatalogNodeSorter(t *testing.T) {
...
+				removedKeys: []string{"bar-service", "baz-service"},
+			},
+		},
+	}
+
+	for _, c := range cases {
+		addedKeys, removedKeys := getChangedKeys(c.input.currState, c.input.prevState)
+
+		if !reflect.DeepEqual(fun.Set(addedKeys), fun.Set(c.output.addedKeys)) {
+			t.Fatalf("expected %q, got %q", c.output.addedKeys, addedKeys)
```
vholovko: Changed the order. Thanks.

provider/consul/consul_catalog_test.go
```diff
@@ -478,3 +479,146 @@ func TestConsulCatalogNodeSorter(t *testing.T) {
...
+		}
+
+		if !reflect.DeepEqual(fun.Set(removedKeys), fun.Set(c.output.removedKeys)) {
+			t.Fatalf("expected %q, got %q", c.output.removedKeys, removedKeys)
```
vholovko: The order was changed.

provider/consul/consul_catalog.go
```diff
@@ -93,38 +100,100 @@ func (p *CatalogProvider) watchServices(stopCh <-chan struct{}) <-chan map[strin
...
 			}
-			catalogOptions.WaitIndex = catalogMeta.LastIndex
-			lastHealthIndex = healthMeta.LastIndex
+
+			options.WaitIndex = meta.LastIndex
+
+			// The response should be unified with watchCatalogServices
+			data, _, err := catalog.Services(&api.QueryOptions{})
+			if err != nil {
+				log.WithError(err).Error("Failed to list services")
```
vholovko: Done it in appropriate way.

provider/consul/consul_catalog.go
```diff
@@ -93,38 +100,100 @@ func (p *CatalogProvider) watchServices(stopCh <-chan struct{}) <-chan map[strin
...
+		for {
+			select {
+			case <-stopCh:
+				return
+			default:
+			}
+
+			data, meta, err := catalog.Services(options)
+			if err != nil {
+				log.WithError(err).Error("Failed to list services")
```
vholovko: Fixed.

provider/consul/consul_catalog.go
```diff
@@ -93,38 +100,100 @@ func (p *CatalogProvider) watchServices(stopCh <-chan struct{}) <-chan map[strin
...
+			options.WaitIndex = meta.LastIndex
+
+			// The response should be unified with watchCatalogServices
+			data, _, err := catalog.Services(&api.QueryOptions{})
+			if err != nil {
+				log.Errorf("Failed to list services: %s", err)
+				return
+			}
 
 			if data != nil {
```
emilevauge: Shouldn't we add a else after this block ?

provider/consul/consul_catalog.go
```diff
@@ -93,38 +100,100 @@ func (p *CatalogProvider) watchServices(stopCh <-chan struct{}) <-chan map[strin
...
+				return
+			}
+
+			if options.WaitIndex == meta.LastIndex {
+				continue
+			}
+
+			options.WaitIndex = meta.LastIndex
+
+			if data != nil {
```
emilevauge: Same comment, shouldn't we add a else block ?

provider/consul/consul_catalog.go
vholovko: I'm not seeing here any functionality that can be placed in the `else` block.
This branch statement was copied from original `watchServices` function and there wasn't any `else` block for it. 

From the source it is clear that there are two places when `nil` for `data` is possible
https://github.com/hashicorp/consul/blob/master/api/catalog.go#L140
https://github.com/hashicorp/consul/blob/master/api/catalog.go#L150
it is when we get an error which is handled at earlier phase in the code above.

Thus it seems that this `if` statement is just a precaution in case `json.NewDecoder` can bring `nil` without notifying us with an error (the actual JSON `null` in the consul response).
I doubt consul response body will be `null` within 200 OK GET response. 
Even in case there are not services at all the JSON response should include the empty map and this situation is handled already.

provider/consul/consul_catalog.go
vholovko: I don't think anything should be placed in the 'else' block. The explanation is given in the earlier comment.

provider/consul/consul_catalog.go
emilevauge: Thanks for you detailed explanation :+1:

ldez:
@vholovko could you squash your commits?

vholovko:
@ldez 
Sorry for inconvenience, but how do I do the squash of my commits in terms of git commands ?
What sequence and in what order should I run to not mess up things ?

ldez:
@vholovko No problem 😃 

You are in a specific case, because you have merge `master` several times in your branch.

In the following commands, I suppose your git local repository setup is:

```shell
$ git remote -v

origin	git@github.com:vholovko/traefik.git (fetch)
origin	git@github.com:vholovko/traefik.git (push)
upstream	git@github.com:containous/traefik.git (fetch)
upstream	git@github.com:containous/traefik.git (push)
```

The commands:

```shell
# delete your local branch
git branch consul_catalog_hot_detect_newly_added_containers

# Recreate a branch with the same name as the previous.
# This branch must be created on the HEAD of the containous/traefik master branch.
git checkout -b consul_catalog_hot_detect_newly_added_containers upstream/master

# Squash all your branch in this new branch
git merge --squash origin/consul_catalog_hot_detect_newly_added_containers

# Create a new commit
git commit -m "Speeding up health change detection by separating it from catalog services check."

# Push if new branch.
git push --force-with-lease origin consul_catalog_hot_detect_newly_added_containers
```

Next time, when time when you want to sync your branch with the master, use `rebase` instead `merge` 😉. In this case the squash is more simple and can be done by only one command `git rebase -i <HASH_OF_THE_COMMIT_BEFORE_YOUR_FIRST_COMMIT>`.

vholovko:
thanks @ldez , your hints were very helpful
all done as you described

PR #1630, opened 2017-05-18
[k8s] Remove rule type path list.

timoreimann:
Instead of doing sanity checks in the Kubernetes provider, we just accept any non-empty value from the annotation and rely on the server part to filter out unknown rules.

This allows us to automatically stay in sync with the currently supported Path matchers/modifiers.

Fixes #1627.

PR #1695, opened 2017-06-01
Fix capitalization of PathPrefixStrip in kubernetes doc

Miouge1:
### Description

This changes the doc and example from pathprefixstrip to PathPrefixStrip.

With v1.3 pathprefixstrip does not work, with the following error:
```
time="2017-06-01T11:46:32Z" level=error msg="Error creating route for frontend cheeses.minikube/wensleydale: Error parsing rule: Error parsing rule: 'pathprefixstrip:/wensleydale'. Unknown function: 'pathprefixstrip'" 
time="2017-06-01T11:46:32Z" level=error msg="Skipping frontend cheeses.minikube/wensleydale..." 
```

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -471,7 +471,7 @@ metadata:
   name: cheeses
   annotations:
     kubernetes.io/ingress.class: traefik
-    traefik.frontend.rule.type: pathprefixstrip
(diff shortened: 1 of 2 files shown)


timoreimann:
Note to reviewers: this was implicitly changed through #1630 where we removed the custom rules list from the k8s provider. While the change of behavior represents a bit of a breaking change, I think there's too little value in holding on to the lower case variants.

PR #1699, opened 2017-06-01
Simplify stripPrefix and stripPrefixRegex tests

ldez:
### Description

- Simplify stripPrefix and stripPrefixRegex tests.
- typo in whitelister file name.

fix flaky tests:
```
--- FAIL: TestStripPrefix (0.00s)
    --- FAIL: TestStripPrefix/general_prefix_on_matching_path (0.00s)
    	assertions.go:226:                      
	Error Trace:	stripPrefix_test.go:97
	Error:      	Received unexpected error:
	            	Get http://127.0.0.1:52739/stat/: net/http: HTTP/1.x transport connection broken: http: CloseIdleConnections called
	Messages:   	Failed to send GET request
```


middlewares/replace_path_test.go
```diff
@@ -27,8 +28,8 @@ func TestReplacePath(t *testing.T) {
 				}),
 			}
 
-			req, err := http.NewRequest("GET", "http://localhost"+path, nil)
-			assert.NoError(t, err, "%s: unexpected error.", path)
+			req, err := http.NewRequest(http.MethodGet, "http://localhost"+path, nil)
```
timoreimann: We can use `testhelpers.MustNewRequest` here and below.

timoreimann:
It is conventional in Go that all `Must*` functions / methods panic. The assumption here is that when you call them, they must only ever fail when you made a logical programming error. When then happens, you have broken an implicit invariant that you want to fix immediately like a compiler error. That's the reason for the immediate panic which will show up right in your face (as opposed to being handled somewhat gracefully when doing `t.Fatal`).

(See `regexp.MustCompile` or Prometheus' `MustRegister` for similar examples.)

Can we return to the original signature of `MustNewRequest`?

ldez:
@timoreimann or we can remove the `Must` prefix: it's a test helper, use `t` seems like normal. FYT?

healthcheck/healthcheck_test.go
```diff
@@ -182,12 +111,12 @@ func TestSetBackendsConfiguration(t *testing.T) {
 
 			lb.Lock()
 			defer lb.Unlock()
-			if lb.numRemovedServers != test.wantNumRemovedServers {
-				t.Errorf("got %d removed servers, wanted %d", lb.numRemovedServers, test.wantNumRemovedServers)
+			if test.wantNumRemovedServers != lb.numRemovedServers {
```
timoreimann: The order in Go is [_got first, want second_](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures); both in the comparison expression and the error message.

healthcheck/healthcheck_test.go
```diff
@@ -2,95 +2,24 @@ package healthcheck
...
 	"testing"
 	"time"
 
+	"github.com/containous/traefik/testhelpers"
 	"github.com/vulcand/oxy/roundrobin"
 )
 
 const healthCheckInterval = 100 * time.Millisecond
 
-type testLoadBalancer struct {
```
timoreimann: Any particular reason why you moved this to the bottom? Does any convention exist that defines the order?

middlewares/addPrefix_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+	var expectedPath string
+	handler := &AddPrefix{
+		Prefix: prefix,
+		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			expectedPath = r.URL.Path
+		}),
+	}
+
+	req := testhelpers.MustNewRequest("GET", "http://localhost"+path, nil)
```
timoreimann: One unused `http.MethodGet` slipped through here. :-)

middlewares/stripPrefixRegex_test.go
```diff
@@ -73,18 +73,16 @@ func TestStripPrefixRegex(t *testing.T) {
...
 
-			resp, err := http.Get(server.URL + test.path)
-			require.NoError(t, err, "%s: unexpected error.", test.path)
+			req := testhelpers.MustNewRequest(http.MethodGet, "http://localhost"+test.path, nil)
+			resp := &httptest.ResponseRecorder{Code: http.StatusOK}
+
+			handler.ServeHTTP(resp, req)
 
-			assert.Equal(t, test.expectedStatusCode, resp.StatusCode, "%s: unexpected status code.", test.path)
+			assert.Equal(t, test.expectedStatusCode, resp.Code, "%s: unexpected status code.", test.path)
```
timoreimann: You don't need to include `test.path` in the error description; the sub-test (which we parameterized with the path) will do that for you.

testhelpers/helpers.go
```diff
@@ -19,3 +20,12 @@ func MustNewRequest(method, urlStr string, body io.Reader) *http.Request {
 	}
 	return request
 }
+
+// MustParseURL parse an URL or panics if it can't
```
timoreimann: nit-picks:

- parse -> parses
- an URL -> a URL (the U is not a vowel sound)

middlewares/stripPrefixRegex_test.go
ldez: I need to do that because I don't see the context case (sub-test) when a test failed.

```
/home/ldez/.gvm/gos/go1.8.3/bin/go test -c -i -o /tmp/TestStripPrefixRegex_in_stripPrefixRegex_test_gogo github.com/containous/traefik/middlewares
/tmp/TestStripPrefixRegex_in_stripPrefixRegex_test_gogo -test.v -test.run ^TestStripPrefixRegex$
	Error Trace:	stripPrefixRegex_test.go:83
	Error:      	Not equal: 
	            	expected: 201
	            	received: 200
	Messages:   	/b/api2/test2: unexpected status code.
	Error Trace:	stripPrefixRegex_test.go:83
	Error:      	Not equal: 
	            	expected: 405
	            	received: 404
	Messages:   	/a/test: unexpected status code.
	Error Trace:	stripPrefixRegex_test.go:83
	Error:      	Not equal: 
	            	expected: 201
	            	received: 200
	Messages:   	/b/api/: unexpected status code.
	Error Trace:	stripPrefixRegex_test.go:83
	Error:      	Not equal: 
	            	expected: 201
	            	received: 200
	Messages:   	/c/api/123/test3: unexpected status code.
	Error Trace:	stripPrefixRegex_test.go:83
	Error:      	Not equal: 
	            	expected: 201
	            	received: 200
	Messages:   	/c/api/123/: unexpected status code.
	Error Trace:	stripPrefixRegex_test.go:83
	Error:      	Not equal: 
	            	expected: 201
	            	received: 200
	Messages:   	/a/api/test: unexpected status code.
	Error Trace:	stripPrefixRegex_test.go:83
	Error:      	Not equal: 
	            	expected: 405
	            	received: 404
	Messages:   	/c/api/abc/test4: unexpected status code.
	Error Trace:	stripPrefixRegex_test.go:83
	Error:      	Not equal: 
	            	expected: 201
	            	received: 200
	Messages:   	/b/api/test1: unexpected status code.

```

healthcheck/healthcheck_test.go
ldez: It's a conventional way: when you read test, you beginning by read the test and after you read some test helpers. To read test you don't need to known the test implementation details.

middlewares/stripPrefixRegex_test.go
ldez: It's a side effect when I run unit tests in Gogland... 😞

healthcheck/healthcheck_test.go
timoreimann: You could also argue that you need to know the implementation details in order to understand the test.

I don't mind too much either way though, was more interested in the motivation.

middlewares/stripPrefixRegex_test.go
timoreimann: Hmm, I'd rather not pollute our tests with duplicate identifiers due to one Go IDE being buggy. On all of Visual Studio Code, Atom, and vim-go, this is not a problem. Might it be acceptable to switch to the terminal and run `go test` manually for cases where tests fail on Gogland until this gets fixed?

You pasted a go test text output. Looking at [this screenshot from a Gogland issue](https://youtrack.jetbrains.com/issue/GO-2957#u=1481990264537&viewimage=snapshot210.png%3Ffile=74-370189&c=true&rw=1574&rh=432&u=1481990264537), it looks like the IDE does show the sub-test name on the left pane?

ldez: Yes, you can see sub-test names but not the log related to only one sub-test.
You can show all log (without the tests "headers") or nothing 😞 

Run tests manually it's not acceptable for me, I need a tool which run continuously my tests after each code changes.

I agree to "not pollute" our tests due to one IDE but it will be hard to me to analyze failed tests. 😞 😞 😞

healthcheck/healthcheck_test.go
ldez: > you _do_ need to know the implementation details in order to understand the test

It's a bad test for me.
A test must clear, readable and understandable by itself.

My main motivation: test readability, always the readability 😃

middlewares/stripPrefixRegex_test.go
timoreimann: AFAICS, you can still see which tests failed but cannot see the details of individual tests.

I think there are a few alternatives:

- Remove all but the failing tests if you want to see the error details.
- Switch to the terminal for failing tests.
- Switch to a different IDE/editor, either for the failing tests or permanently.

Visual Studio Code (my IDE of choice) is pretty good, though not perfect. There are a few things which cannot be done as nicely as with the Go CLI toolchain; yet, I refrain from letting those shortcomings leak into the code base.

I'll approve the PR since this is my only concern left and I don't want to block it from getting merged any longer (assuming that @emilevauge is fine with it too). Nevertheless, I'm still very much against this change, and even more when it starts to spread throughout the code base over time.

(Side note: Could you please consider filing a bug report on the Gogland issue tacker? IMHO, JetBrains should fix this problem ASAP.)

PR #1700, opened 2017-06-02
Use correct type when watching for k8s secrets

kekoav:
This was likely just a copy-paste issue, the bug should be benign because the secret is cast to the correct type later, but the additional logging is a major annoyance, and is happening even if basic auth is not in use with Kubernetes.

### DIFF
--- a/provider/kubernetes/client.go
+++ b/provider/kubernetes/client.go
@@ -221,7 +221,7 @@ func (c *clientImpl) WatchSecrets(watchCh chan<- interface{}, stopCh <-chan stru
 
 	c.secStore, c.secController = cache.NewInformer(
 		source,
-		&v1.Endpoints{},
+		&v1.Secret{},
 		resyncPeriod,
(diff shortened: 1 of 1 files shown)


timoreimann:
Good catch, @kekoav -- thanks for the fix!

@dtomcej @errm @emilevauge it seems like the problem is enough of a nuisance to justify going into a 1.3 patch release. Would you agree?

emilevauge:
@timoreimann makes sense!

timoreimann:
:+1:

@kekoav would you mind rebasing onto the v1.3 branch? We'll merge things back into master afterwards.

kekoav:
@timoreimann I cherry-picked the fix onto 1.3 branch, and changed the PR to target the v1.3 branch.

alpe:
@kekoav thanks for taking care! 💐 I was a little bit overloaded and couldn't pay enough attention to this project.

Issue #834, opened 2016-11-12
StatsD (e.g. DataDog) support


jippi:
Hi,

I'm currently evaluating traefik for our internal services, and so far I'm very impressed - good job on such an easy approachable project.

I can't find any instrumentation for statsd to get runtime information into a nice UI view like datadog.

Is this something planned, or is it there already and I totally missed it somewhere?

Would like to have insights into errors, requests, errors and so forth for frontend (and maybe backend) 

`/health` do provide some information, but not in a very granular level :)

errm:
We are planning to add prometheus metrics.

And there is also a proposal to add support for opentracing #771

jippi:
@errm okay, so no plans to add basic statsd support? both prometheus and opentracing  are fairly new technologies, where statsd is old and wildly supported and understood :)

errm:
At this stage I understand the plan is to add metrics using https://github.com/rcrowley/go-metrics.

This should make supporting a number of metrics systems quite simple, I think there is a client for statsd.

enxebre:
this is partially covered now by #1022 and #1042. We'd need to add an statsD implementation analogous to Prometheus one https://github.com/containous/traefik/blob/master/middlewares/prometheus.go by using https://github.com/go-kit/kit/tree/master/metrics/statsd

aantono:
I have a PR ready, but will probably wait till #1504 gets merged (unless there is a desire to get the changes in for 1.3 release)

timoreimann:
We have already frozen the 1.3 branch for features (only bug fixes at this stage), so waiting should be fine. 👍

hkaj:
Hey all,
Prometheus support is coming to Datadog and we're adding a check for Traefik using it (should be ready to use with the 5.14 release). That means auto discovery will work with Traefik, and also that these metrics will be official metrics in Datadog and not billed as custom metrics. Here's the PR for tracking: https://github.com/DataDog/integrations-extras/pull/64

That's just a heads up, StatsD instrumentation is still great for other solutions.

ldez:
Closed by #1701

Issue #1621, opened 2017-05-17
Support multi-port service routing for containers running on Marathon


aantono:
Similar to requirements in #444, provide Marathon support to match the following requirements:

If labels contain for example:
```
traefik.mycustomservice.port=443
traefik.mycustomservice.frontend.rule=Path:/mycustomservice
traefik.anothercustomservice.port=8080
traefik.anothercustomservice.frontend.rule=Path:/anotherservice
```
all traffic to frontend on path `/mycustomservice` is redirected to the port `443` of the container defining the label while using path `/anotherservice` will redirect to the port `8080` of the docker container defining the label

Available properties are
```
traefik.<service-name>.port
traefik.<service-name>.weight
traefik.<service-name>.protocol
traefik.<service-name>.frontend.backend
traefik.<service-name>.frontend.entryPoints
traefik.<service-name>.frontend.rule
traefik.<service-name>.frontend.passHostHeader
traefik.<service-name>.frontend.priority
traefik.<service-name>.frontend.rule
```

aantono:
Closing due to #1701 being merged to `master`

PR #1701, opened 2017-06-02
DataDog and StatsD Metrics Support

aantono:
While still waiting for #1504 to be merged and rebase on top, this is to get the review process for #834 started for DataDog and StatsD metrics support.

The solution is based on the same `go-kit/metrics` library as for Prometheus, except using the DataDog and StatsD providers.  Given that there is a possible integration from DataDog being created by https://github.com/DataDog/integrations-extras/pull/64, I've used the [matching metric names](https://github.com/DataDog/integrations-extras/pull/64/files#diff-9a41570ca95801d7da0aedd667152742R14) to push the metrics to DD agent, as the Prometheus pull is mapping.  Internally the DD agent puller is still pushing the collected data into DogStatsD listener, so it will end up flowing though the same ingestion gate, thus taking advantage of the "included integration metrics".

I've had to add the additional dependencies into Glide, so after running `./scripts/glide.sh add <>` it has changed a bunch of files in the vendor, so hopefully that's done correctly.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 6aff4c6177ddc3247530d141a93f5bb044ee72acaa63b5667ceb205828c8ad03
-updated: 2017-07-11T23:50:31.241672481+02:00
+hash: 4d24f4a986de7e07c32b63abc3c8bf365d205df0a6f65ba4a6ca3d7ac7ae2256
+updated: 2017-07-20T23:54:09.638352893+02:00
 imports:
 - name: cloud.google.com/go
   version: 2e6a95edb1071d750f6d7db777bf66cd2997af6c
@@ -82,8 +82,6 @@ imports:
   version: 9208b142303c12d8899bae836fd524ac9338b4fd
 - name: github.com/codegangsta/cli
   version: bf4a526f48af7badd25d2cb02d587e1b01be3b50
-- name: github.com/urfave/negroni
-  version: 490e6a555d47ca891a89a150d0c1ef3922dfffe9
 - name: github.com/containous/flaeg
   version: b5d2dc5878df07c2d74413348186982e7b865871
 - name: github.com/containous/mux
@@ -242,9 +240,17 @@ imports:
 - name: github.com/go-kit/kit
   version: f66b0e13579bfc5a48b9e2a94b1209c107ea1f41
   subpackages:
+  - log
   - metrics
+  - metrics/dogstatsd
   - metrics/internal/lv
+  - metrics/internal/ratemap
+  - metrics/multi
   - metrics/prometheus
+  - metrics/statsd
+  - util/conn
+- name: github.com/go-logfmt/logfmt
+  version: 390ab7935ee28ec6b286364bba9b4dd6410cb3d5
 - name: github.com/go-openapi/jsonpointer
   version: 46af16f9f7b149af66e5d1bd010e3574dc06de98
 - name: github.com/go-openapi/jsonreference
@@ -253,6 +259,8 @@ imports:
   version: 6aced65f8501fe1217321abf0749d354824ba2ff
 - name: github.com/go-openapi/swag
   version: 1d0bd113de87027671077d3c71eb3ac5d7dbba72
+- name: github.com/go-stack/stack
+  version: 54be5f394ed2c3e19dac9134a40a95ba5a017f7b
 - name: github.com/gogo/protobuf
   version: 909568be09de550ed094403c2bf8a261b5bb730a
   subpackages:
@@ -302,6 +310,8 @@ imports:
   version: 72f9bd7c4e0c2a40055ab3d0f09654f730cce982
 - name: github.com/juju/ratelimit
   version: 77ed1c8a01217656d2080ad51981f6e99adaa177
+- name: github.com/kr/logfmt
+  version: b84e30acd515aadc4b783ad4ff83aff3299bdfe0
 - name: github.com/mailgun/timetools
   version: 7e6055773c5137efbeb3bd2410d705fe10ab6bfd
 - name: github.com/mailru/easyjson
@@ -436,6 +446,8 @@ imports:
   - assert
   - mock
   - require
+- name: github.com/stvp/go-udp-testing
+  version: 06eb4f886d9f8242b0c176cf0d3ce5ec2cedda05
 - name: github.com/thoas/stats
   version: 152b5d051953fdb6e45f14b6826962aadc032324
 - name: github.com/timewasted/linode
@@ -452,6 +464,8 @@ imports:
   version: 50716a0a853771bb36bfce61a45cdefdb98c2e6e
 - name: github.com/unrolled/secure
   version: 824e85271811af89640ea25620c67f6c2eed987e
+- name: github.com/urfave/negroni
+  version: 490e6a555d47ca891a89a150d0c1ef3922dfffe9
 - name: github.com/vulcand/oxy
   version: 7da864c1d53bd58165435bb78bbf8c01f01c8f4a
   repo: https://github.com/containous/oxy.git
--- a/middlewares/datadog.go
+++ b/middlewares/datadog.go
@@ -68,7 +68,7 @@ func InitDatadogClient(config *types.Datadog) *time.Ticker {
 		}
 		pushInterval, err := time.ParseDuration(config.PushInterval)
 		if err != nil {
-			log.Warnf("Unable to parse %s into pushInterval", config.PushInterval)
+			log.Warnf("Unable to parse %s into pushInterval, using 10s as default value", config.PushInterval)
 			pushInterval = 10 * time.Second
 		}
 
--- a/middlewares/datadog.go
+++ b/middlewares/datadog.go
@@ -1,0 +1,87 @@
+package middlewares
+
+import (
+	"time"
+
+	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/types"
+	kitlog "github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/dogstatsd"
+)
+
+var _ Metrics = (Metrics)(nil)
+
+var datadogClient = dogstatsd.New("traefik.", kitlog.LoggerFunc(func(keyvals ...interface{}) error {
+	log.Info(keyvals)
+	return nil
+}))
+
+var datadogTicker *time.Ticker
+
+// Metric names consistent with https://github.com/DataDog/integrations-extras/pull/64
+const (
+	ddMetricsReqsName    = "requests.total"
+	ddMetricsLatencyName = "request.duration"
+)
+
+// Datadog is an Implementation for Metrics that exposes datadog metrics for the latency
+// and the number of requests partitioned by status code and method.
+// - number of requests partitioned by status code and method
+// - request durations
+// - amount of retries happened
+type Datadog struct {
+	reqsCounter          metrics.Counter
+	reqDurationHistogram metrics.Histogram
+	retryCounter         metrics.Counter
+}
+
+func (dd *Datadog) getReqsCounter() metrics.Counter {
+	return dd.reqsCounter
+}
+
+func (dd *Datadog) getReqDurationHistogram() metrics.Histogram {
+	return dd.reqDurationHistogram
+}
+
+func (dd *Datadog) getRetryCounter() metrics.Counter {
+	return dd.retryCounter
+}
+
+// NewDataDog creates new instance of Datadog
+func NewDataDog(name string) *Datadog {
+	var m Datadog
+
+	m.reqsCounter = datadogClient.NewCounter(ddMetricsReqsName, 1.0).With("service", name)
+	m.reqDurationHistogram = datadogClient.NewHistogram(ddMetricsLatencyName, 1.0).With("service", name)
+
+	return &m
+}
+
+// InitDatadogClient initializes metrics pusher and creates a datadogClient if not created already
+func InitDatadogClient(config *types.Datadog) *time.Ticker {
+	if datadogTicker == nil {
+		address := config.Address
+		if len(address) == 0 {
+			address = "localhost:8125"
+		}
+		pushInterval, err := time.ParseDuration(config.PushInterval)
+		if err != nil {
+			log.Warnf("Unable to parse %s into pushInterval", config.PushInterval)
+			pushInterval = 10 * time.Second
+		}
+
+		report := time.NewTicker(pushInterval)
+
+		go datadogClient.SendLoop(report.C, "udp", address)
+
+		datadogTicker = report
+	}
+	return datadogTicker
+}
+
+// Stop stops internal datadogTicker which controls the pushing of metrics to DD Agent and resets it to `nil`
+func (dd *Datadog) Stop() {
+	datadogTicker.Stop()
+	datadogTicker = nil
+}
--- a/middlewares/datadog.go
+++ b/middlewares/datadog.go
@@ -84,7 +84,7 @@ func InitDatadogClient(config *types.Datadog) *time.Ticker {
 }
 
 // Stop stops internal datadogTicker which controls the pushing of metrics to DD Agent and resets it to `nil`
-func (dd *Datadog) Stop() {
+func StopDatadogClient() {
 	if datadogTicker != nil {
 		datadogTicker.Stop()
 	}
--- a/middlewares/datadog.go
+++ b/middlewares/datadog.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/safe"
 	"github.com/containous/traefik/types"
 	kitlog "github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/metrics"
@@ -73,7 +74,9 @@ func InitDatadogClient(config *types.Datadog) *time.Ticker {
 
 		report := time.NewTicker(pushInterval)
 
-		go datadogClient.SendLoop(report.C, "udp", address)
+		safe.Go(func() {
+			datadogClient.SendLoop(report.C, "udp", address)
+		})
 
 		datadogTicker = report
 	}
--- a/middlewares/datadog.go
+++ b/middlewares/datadog.go
@@ -85,6 +85,8 @@ func InitDatadogClient(config *types.Datadog) *time.Ticker {
 
 // Stop stops internal datadogTicker which controls the pushing of metrics to DD Agent and resets it to `nil`
 func (dd *Datadog) Stop() {
-	datadogTicker.Stop()
+	if datadogTicker != nil {
+		datadogTicker.Stop()
+	}
 	datadogTicker = nil
 }
--- a/middlewares/datadog.go
+++ b/middlewares/datadog.go
@@ -83,7 +83,7 @@ func InitDatadogClient(config *types.Datadog) *time.Ticker {
 	return datadogTicker
 }
 
-// Stop stops internal datadogTicker which controls the pushing of metrics to DD Agent and resets it to `nil`
+// StopDatadogClient stops internal datadogTicker which controls the pushing of metrics to DD Agent and resets it to `nil`
 func StopDatadogClient() {
 	if datadogTicker != nil {
 		datadogTicker.Stop()
--- a/middlewares/metrics.go
+++ b/middlewares/metrics.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/multi"
 )
 
 // Metrics is an Interface that must be satisfied by any system that
@@ -20,6 +21,52 @@ type Metrics interface {
 // expose retry specific Metrics.
 type RetryMetrics interface {
 	getRetryCounter() metrics.Counter
+}
+
+// MultiMetrics is a struct that provides a wrapper container for multiple Metrics, if they are configured
+type MultiMetrics struct {
+	wrappedMetrics       *[]Metrics
+	reqsCounter          metrics.Counter
+	reqDurationHistogram metrics.Histogram
+	retryCounter         metrics.Counter
+}
+
+// NewMultiMetrics creates a new instance of MultiMetrics
+func NewMultiMetrics(manyMetrics []Metrics) *MultiMetrics {
+	counters := []metrics.Counter{}
+	histograms := []metrics.Histogram{}
+	retryCounters := []metrics.Counter{}
+
+	for _, m := range manyMetrics {
+		counters = append(counters, m.getReqsCounter())
+		histograms = append(histograms, m.getReqDurationHistogram())
+		retryCounters = append(retryCounters, m.getRetryCounter())
+	}
+
+	var mm MultiMetrics
+
+	mm.wrappedMetrics = &manyMetrics
+	mm.reqsCounter = multi.NewCounter(counters...)
+	mm.reqDurationHistogram = multi.NewHistogram(histograms...)
+	mm.retryCounter = multi.NewCounter(retryCounters...)
+
+	return &mm
+}
+
+func (mm *MultiMetrics) getReqsCounter() metrics.Counter {
+	return mm.reqsCounter
+}
+
+func (mm *MultiMetrics) getReqDurationHistogram() metrics.Histogram {
+	return mm.reqDurationHistogram
+}
+
+func (mm *MultiMetrics) getRetryCounter() metrics.Counter {
+	return mm.retryCounter
+}
+
+func (mm *MultiMetrics) getWrappedMetrics() *[]Metrics {
+	return mm.wrappedMetrics
 }
 
 // MetricsWrapper is a Negroni compatible Handler which relies on a
--- a/middlewares/statsd.go
+++ b/middlewares/statsd.go
@@ -4,6 +4,7 @@ import (
 	"time"
 
 	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/safe"
 	"github.com/containous/traefik/types"
 	kitlog "github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/metrics"
@@ -66,7 +67,9 @@ func InitStatsdClient(config *types.Statsd) *time.Ticker {
 
 		report := time.NewTicker(pushInterval)
 
-		go statsdClient.SendLoop(report.C, "udp", address)
+		safe.Go(func() {
+			statsdClient.SendLoop(report.C, "udp", address)
+		})
 
 		statsdTicker = report
 	}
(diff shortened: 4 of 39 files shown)


aantono:
Rebased on top of #1504 changes from Master.

glide.lock
```diff
@@ -194,24 +190,32 @@ imports:
...
 - name: github.com/emicklei/go-restful
-  version: 892402ba11a2e2fd5e1295dd633481f27365f14d
+  version: 89ef8af493ab468a45a42bb0d89a06fccdd2fb22
   subpackages:
   - log
   - swagger
 - name: github.com/fatih/color
   version: 9131ab34cf20d2f6d83fdc67168a5430d1c7dc23
 - name: github.com/gambol99/go-marathon
-  version: 15ea23e360abb8b25071e677aed344f31838e403
```
ldez: Could you add/update only your dependencies?

middlewares/metrics.go
```diff
@@ -22,6 +23,51 @@ type RetryMetrics interface {
...
+	counters := []metrics.Counter{}
+	histograms := []metrics.Histogram{}
+	retryCounters := []metrics.Counter{}
+
+	for _, m := range manyMetrics {
+		counters = append(counters, m.getReqsCounter())
+		histograms = append(histograms, m.getReqDurationHistogram())
+		retryCounters = append(retryCounters, m.getRetryCounter())
+	}
+	var mm MultiMetrics
```
ldez: move this line after the next empty line.

middlewares/statsd_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	expected := []string{
+		// We are only validating counts, as it is nearly impossible to validate latency, since it varies every run
+		"traefik.requests.total:2.000000|c\n",
+	}
+
+	udp.ShouldReceiveAll(t, expected, func() {
+		n.ServeHTTP(recorder, req1)
+		n.ServeHTTP(recorder, req2)
+		//		body := recorder.Body.String()
```
ldez: Dead code ?

glide.lock
aantono: Not very sure why this shows up as a diff, the `glide.yaml` has `d672c6fbb499596869d95146a26e7d0746c06c54` version for `github.com/gambol99/go-marathon` which came from a prior [commit](https://github.com/containous/traefik/commit/4e238280bceae43011b66613b94bd9f4ac1c233c) (https://github.com/containous/traefik/blame/master/glide.yaml#L96)

ldez: It's not only for `go-marathon`. It's a side affect of the `glide get`.
You must use a workaround:
1. Revert the glide manifest and lock files.
1. Use the [github.com/multiplay/glide-pin](https://github.com/multiplay/glide-pin) plugin to pin all versions in the manifest file to the ones given in the lock file.
1. Add your dependencies `glide get`
1. Run `glide update -v` to update the lock file.
1. Undo the pinning.
1. Add your dependencies in `glide.yml`
1. Use [glide-hash](https://github.com/mattfarina/glide-hash) to get the manifest file hash and put it in the lock file to please `validate-glide`.

aantono: Tried to follow those steps, only to run into errors :(

Invoked `glide pin` followed by `glide get github.com/go-kit/kit/metrics/dogstatsd` commands and things exploded:

```
[ERROR]	Error scanning github.com/docker/docker/cli/config: open /Users/aantonov/.glide/cache/src/https-github.com-docker-docker/cli/config: no such file or directory
[ERROR]	This error means the referenced package was not found.
[ERROR]	Missing file or directory errors usually occur when multiple packages
[ERROR]	share a common dependency and the first reference encountered by the scanner
[ERROR]	sets the version to one that does not contain a subpackage needed required
[ERROR]	by another package that uses the shared dependency. Try setting a
[ERROR]	version in your glide.yaml that works for all packages that share this
[ERROR]	dependency.
[ERROR]	Error scanning github.com/docker/docker/cli/config/configfile: open /Users/aantonov/.glide/cache/src/https-github.com-docker-docker/cli/config/configfile: no such file or directory
[ERROR]	This error means the referenced package was not found.
[ERROR]	Missing file or directory errors usually occur when multiple packages
[ERROR]	share a common dependency and the first reference encountered by the scanner
[ERROR]	sets the version to one that does not contain a subpackage needed required
[ERROR]	by another package that uses the shared dependency. Try setting a
[ERROR]	version in your glide.yaml that works for all packages that share this
[ERROR]	dependency.
[INFO]	--> Fetching updates for github.com/xeipuuv/gojsonpointer.
[ERROR]	Error scanning github.com/docker/docker/cli/command/image/build: open /Users/aantonov/.glide/cache/src/https-github.com-docker-docker/cli/command/image/build: no such file or directory
[ERROR]	This error means the referenced package was not found.
[ERROR]	Missing file or directory errors usually occur when multiple packages
[ERROR]	share a common dependency and the first reference encountered by the scanner
[ERROR]	sets the version to one that does not contain a subpackage needed required
[ERROR]	by another package that uses the shared dependency. Try setting a
[ERROR]	version in your glide.yaml that works for all packages that share this
[ERROR]	dependency.
[INFO]	--> Fetching updates for golang.org/x/time.
[INFO]	--> Setting version for golang.org/x/time to a4bde12657593d5e90d0533a3e4fd95e635124cb.
[INFO]	--> Fetching updates for github.com/Azure/go-ansiterm.
[INFO]	--> Fetching updates for github.com/gorilla/mux.
[INFO]	--> Setting version for github.com/gorilla/mux to e444e69cbd2e2e3e0749a2f3c717cec491552bbf.
[INFO]	--> Fetching updates for github.com/vbatts/tar-split.
[INFO]	--> Fetching updates for github.com/opencontainers/runtime-spec.
[ERROR]	Failed to retrieve a list of test dependencies: Error resolving imports
```

timoreimann: Can you try two things:

- Run `glide cc` first thing. (This will clear the glide cache which is sometimes known to cause problems.)
- use `script/glide.sh <command>` instead of `glide <command>` everywhere.

aantono: Sure thing... Should I do the `glide pin` first?

timoreimann: @aantono it shouldn't matter.

For `glide pin` and `glide hash`, you should not (and cannot) use `script/glide.sh`: Those two are not really glide commands that our scripts wraps but plugins.

aantono: Keep getting errors after cleaning the Glide cache:

```
[ERROR]	Error scanning k8s.io/client-go/kubernetes/typed/authentication/v1: open /Users/aantonov/.glide/cache/src/https-k8s.io-client-go/kubernetes/typed/authentication/v1: no such file or directory
[ERROR]	This error means the referenced package was not found.
[ERROR]	Missing file or directory errors usually occur when multiple packages
[ERROR]	share a common dependency and the first reference encountered by the scanner
[ERROR]	sets the version to one that does not contain a subpackage needed required
[ERROR]	by another package that uses the shared dependency. Try setting a
[ERROR]	version in your glide.yaml that works for all packages that share this
[ERROR]	dependency.
```
The same error is for `k8s.io/client-go/kubernetes/typed/authorization/v1`, `k8s.io/client-go/kubernetes/typed/autoscaling/v2alpha1`, `k8s.io/client-go/kubernetes/typed/certificates/v1beta1`, `k8s.io/client-go/kubernetes/typed/rbac/v1beta1`, `k8s.io/client-go/kubernetes/typed/settings/v1alpha1`, `k8s.io/client-go/kubernetes/typed/storage/v1` and `k8s.io/client-go/util/flowcontrol` dependencies. :(

timoreimann: In that case, it becomes more and more likely that we are really running into the problem the glide error describes: conflicting dependencies.

I can double check on my end. Could you tell me which additional dependencies the PR needs?

aantono: I've been trying to add:
```
github.com/go-kit/kit/metrics/dogstatsd
github.com/go-kit/kit/metrics/statsd
github.com/go-kit/kit/metrics/multi
github.com/go-kit/kit/log
github.com/go-kit/kit/util/conn
github.com/stvp/go-udp-testing
```
Would greatly appreciate if you could see if it will work on your end.

timoreimann: I'm afraid I ran into similar problem: Trying to upgrade `github.com/stvp/go-udp-testing` alone yields many conflicts with the client-go library.

I'm having a hard time understanding why we run into these conflicts since go-udp-testing looks super simple and seemingly doesn't have any direct or transitive dependencies with go-client.

Maybe we're holding it wrong at some point...

ldez: @aantono could you squash all commits and remove changes on `glide.yml`, `glide.lock` and `vendor`? I will try to resolve the problem after that.

aantono: Done

aantono:
Rebased against `master`, removing changes on `glide.yml`, `glide.lock` and `vendor`.  @ldez, this is all yours! ;)

ldez:
@aantono it's ready. 🎉

middlewares/datadog.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		pushInterval, err := time.ParseDuration(config.PushInterval)
+		if err != nil {
+			log.Warnf("Unable to parse %s into pushInterval", config.PushInterval)
+			pushInterval = 10 * time.Second
+		}
+
+		report := time.NewTicker(pushInterval)
+
+		go datadogClient.SendLoop(report.C, "udp", address)
```
ldez:
```golang
safe.Go(func() {
	datadogClient.SendLoop(report.C, "udp", address)
})
```

middlewares/statsd.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		pushInterval, err := time.ParseDuration(config.PushInterval)
+		if err != nil {
+			log.Warnf("Unable to parse %s into pushInterval", config.PushInterval)
+			pushInterval = 10 * time.Second
+		}
+
+		report := time.NewTicker(pushInterval)
+
+		go statsdClient.SendLoop(report.C, "udp", address)
```
ldez:
```golang
safe.Go(func() {
	statsdClient.SendLoop(report.C, "udp", address)
})
```

aantono:
I've updated the PR with requested changes.
@ldez Will you resolve the conflict on the `glide.lock` and `glide.yaml` from `master`?

ldez:
By mistake, you have rebase your PR on an old hash.

I will fix that.

To pull the PR you must:
```shell
git fetch origin
git pull -f
```
or
```shell
git fetch origin
git branch -D Issue834
git checkout Issue834
```

You don't need to push force.

ldez:
@aantono please stop rebase 😃 . I will do that after the final review.

aantono:
You got it.  Sorry about that, wasn't trying to cause trouble, just wanted to make sure the PR is merge-ready ;)

ldez:
ping @aantono

aantono:
I'm in the process of finishing changes for #1621, will do this right after.  Hopefully there is a bit of time to get this ready before 1.4 is released, but let me know if there is time sensitivity around this.

middlewares/datadog.go
```diff
@@ -0,0 +1,90 @@
...
+		pushInterval, err := time.ParseDuration(config.PushInterval)
+		if err != nil {
+			log.Warnf("Unable to parse %s into pushInterval", config.PushInterval)
+			pushInterval = 10 * time.Second
+		}
+
+		report := time.NewTicker(pushInterval)
+
+		safe.Go(func() {
+			datadogClient.SendLoop(report.C, "udp", address)
```
aantono: Given your next comment below, would it be better to call InitDatadogClient in the startup sequence, if the config has the configuration and just keep the client reusable all throughout the runtime of Traefik?  This way the `safe.Go` routine does not need to be managed and stopped explicitly, as it will only need to be terminated when Traefik process is being shut down.  Thoughts?

middlewares/statsd.go
```diff
@@ -0,0 +1,85 @@
...
+			statsdClient.SendLoop(report.C, "udp", address)
+		})
+
+		statsdTicker = report
+	}
+	return statsdTicker
+}
+
+// Stop stops internal statsdTicker which controls the pushing of metrics to StatsD Agent and resets it to `nil`
+func (s *Statsd) Stop() {
```
emilevauge: Is it called ?

middlewares/datadog.go
```diff
@@ -0,0 +1,92 @@
...
+		})
+
+		datadogTicker = report
+	}
+	return datadogTicker
+}
+
+// Stop stops internal datadogTicker which controls the pushing of metrics to DD Agent and resets it to `nil`
+func (dd *Datadog) Stop() {
```
emilevauge: Is it called ?

middlewares/datadog.go
emilevauge: I would prefer being able to stop gracefully this routine.
You already have a `Stop()` func that fits well, which is never called BTW.

aantono:
No worries, You bet!  Looks like the `glide.lock` is again in conflict... guess its time for @ldez to do his magic again! ;)

middlewares/datadog.go
```diff
@@ -0,0 +1,92 @@
...
+
+	m.reqsCounter = datadogClient.NewCounter(ddMetricsReqsName, 1.0).With("service", name)
+	m.reqDurationHistogram = datadogClient.NewHistogram(ddMetricsLatencyName, 1.0).With("service", name)
+
+	return &m
+}
+
+// InitDatadogClient initializes metrics pusher and creates a datadogClient if not created already
+func InitDatadogClient(config *types.Datadog) *time.Ticker {
+	if datadogTicker == nil {
```
timoreimann: I suppose we do not enter this function from multiple goroutines concurrently, and hence don't need to synchronize access?

middlewares/statsd.go
```diff
@@ -0,0 +1,85 @@
...
+	var m Statsd
+
+	m.reqsCounter = statsdClient.NewCounter(ddMetricsReqsName, 1.0).With("service", name)
+	m.reqDurationHistogram = statsdClient.NewTiming(ddMetricsLatencyName, 1.0).With("service", name)
+
+	return &m
+}
+
+// InitStatsdClient initializes metrics pusher and creates a statsdClient if not created already
+func InitStatsdClient(config *types.Statsd) *time.Ticker {
```
timoreimann: The entire file, but especially this function look _very_ similar to what is implemented in `datadog.go`. Is there any chance to factor out the duplicated code?

middlewares/statsd_test.go
```diff
@@ -0,0 +1,52 @@
+package middlewares
```
timoreimann: Tests look very similar to `datadog_test.go` too. Reuse opportunity?

server/server.go
```diff
@@ -1058,18 +1060,59 @@ func (*Server) configureBackends(backends map[string]*types.Backend) {
...
+			metric, _, err := middlewares.NewPrometheus(name, globalConfig.Web.Metrics.Prometheus)
+			if err != nil {
+				log.Errorf("Error creating Prometheus Metrics implementation: %s", err)
+			}
+			log.Debug("Configured Prometheus Metrics")
+			metrics = append(metrics, metric)
+		}
+		if globalConfig.Web.Metrics.Datadog != nil {
+			metric := middlewares.NewDataDog(name)
+			log.Debugf("Configured DataDog Metrics pushing to %s once every %s", globalConfig.Web.Metrics.Datadog.Address, globalConfig.Web.Metrics.Datadog.PushInterval)
```
timoreimann: s/Metrics/metrics/

server/server.go
```diff
@@ -1058,18 +1060,59 @@ func (*Server) configureBackends(backends map[string]*types.Backend) {
...
+			metrics = append(metrics, metric)
 		}
-		return metrics
+		if globalConfig.Web.Metrics.StatsD != nil {
+			metric := middlewares.NewStatsD(name)
+			log.Debugf("Configured StatsD Metrics pushing to %s once every %s", globalConfig.Web.Metrics.StatsD.Address, globalConfig.Web.Metrics.StatsD.PushInterval)
```
timoreimann: s/Metrics/metrics/

server/server.go
```diff
@@ -1058,18 +1060,59 @@ func (*Server) configureBackends(backends map[string]*types.Backend) {
...
+			middlewares.InitStatsdClient(globalConfig.Web.Metrics.StatsD)
+		}
+	}
+}
+
+func stopMetricsClients(globalConfig GlobalConfiguration) {
+	metricsEnabled := globalConfig.Web != nil && globalConfig.Web.Metrics != nil
+	if metricsEnabled {
+		if globalConfig.Web.Metrics.Datadog != nil {
+			middlewares.StopDatadogClient()
```
timoreimann: Do we actually need all those ifs before? It looks like `StopDatadogClient` won't try to stop anything anyway unless the ticker was previously set (i.e., we have configured Datadog for sure).

server/server.go
```diff
@@ -1058,18 +1060,59 @@ func (*Server) configureBackends(backends map[string]*types.Backend) {
...
+		}
+		if globalConfig.Web.Metrics.StatsD != nil {
+			middlewares.StopStatsdClient()
```
timoreimann: Same question about the if'ing here.

timoreimann:
I think we're also missing a docs update somewhere to inform our users about the possibility to use Datadog or Statsd for metrics collection.

middlewares/statsd.go
ldez: It's two different features, factorize don't seems a good idea.
For me, it's like [consul.go](https://github.com/containous/traefik/blob/master/provider/consul/consul.go], [etcd.go](https://github.com/containous/traefik/blob/master/provider/etcd/etcd.go) or  [boltdb.go](https://github.com/containous/traefik/blob/master/provider/boltdb/boltdb.go).

middlewares/statsd_test.go
ldez: same think here, it's two different features.

middlewares/datadog.go
aantono: Correct.  We only call `InitDatadogClient` once when the server starts.

middlewares/datadog.go
```diff
@@ -0,0 +1,92 @@
...
+		address := config.Address
+		if len(address) == 0 {
+			address = "localhost:8125"
```
aantono: The default DataDog statsD listener is running on localhost:8125, which is a well known port (just like it is 9125 for original StatsD), so it is very common for people to not define it explicitly and use the default (be that in DataDog agent config, or in the clients)  (See default dd-agent config for reference - https://github.com/DataDog/dd-agent/blob/master/datadog.conf.example#L163)

middlewares/statsd.go
aantono: While I thought about trying to generalize and factorize the code out, like @ldez mentioned, while the 2 files look very similar in pattern, they internally use different go-kit/metrics structures to initialize metrics, name them and return outbound structures, so there isn't much that can actually be shared, besides the logical steps of operation, which are similar due to the similarity in DataDog being an enhancement on top of StatsD protocol.

server/server.go
aantono: We probably don't, at least not at the current moment, I just put them in there for security measure, in case later on the internals of `StopDAtadogClient()` change in a way that would require such a check to be done, but we could forget to do it.  I would be happy to remove the `if` checks, if needed.

middlewares/statsd.go
timoreimann: @aantono The initial function is what I was mostly after. They look pretty much identical except for the ticker they reference and the receiver they trigger the loop on. That should be fairly easy to parameterize.

More importantly to me, refactoring should allow us to squash two fairly big tests into a single one.

When you say

>besides the logical steps of operation, which are similar due to the similarity in DataDog being **an enhancement on top of StatsD protocol**.
(**emphasis** by me)

to me that's really a signal for how the two share certain attributes (i.e., the function in question) that we could reflect code-wise by a central implementation.

So that's my view. Maybe @emilevauge can step in and become the tipping point. I'll be happy to follow whatever majority constitutes. :-)

timoreimann: @ldez tbh, seeing three k/v store implementations spanning 30+ LOC each that really only differ by a single store type they pass into a function is rather an *anti-example* to me. :open_mouth:

server/server.go
timoreimann: I'd argue that `StopDatadogClient` should decide if the client still needs to be stopped or not, rather than relying on static configuration parameters. The function already does this by means of using the ticker value as a flag, so overall I think we can remove the if blocks.

middlewares/statsd.go
```diff
@@ -0,0 +1,85 @@
...
+	if statsdTicker == nil {
+		address := config.Address
+		if len(address) == 0 {
+			address = "localhost:8125"
```
timoreimann: Shouldn't this use port 9125 given [what you said here](https://github.com/containous/traefik/pull/1701/files#r128391846) about the default ports?

aantono: My mistake, I confused StatsD with [statsd-exporter](https://github.com/prometheus/statsd_exporter) project, which defaults to 9125.  The default port for both StatsD and dogStatsD is indeed 8125, since the protocols are largely compatible (meaning you can send plain StatsD format data to dogStatsD agent, and it will accept it)

middlewares/statsd.go
aantono: So I looked once again at the `InitDataDogClient` and `InitStatsDClient` code and the only part they share is this:
```
                address := config.Address
		if len(address) == 0 {
			address = "localhost:8125"
		}
		pushInterval, err := time.ParseDuration(config.PushInterval)
		if err != nil {
			log.Warnf("Unable to parse %s into pushInterval", config.PushInterval)
			pushInterval = 10 * time.Second
		}
```
Given that it looks like boiler-plate code, it's really just a validation of `config` properties (but keep in mind that `config` is of different struct type: `types.Datadog` vs. `types.Statsd`), they just happen to both have `address` and `pushInterval` properties).
The rest of the function handles initializing different clients and each provider has their own ticker as well, which is a statically managed variable, so not that great for factorizing.  IMHO it would be more code and will make it harder to read and reason about the code if we to try to come up with a base config type (not even sure what to call it) and then have the 2 concrete types encapsulate it.  While I do agree that on the surface this does seem like a good possible case for boiler-plate reduction, once you look deeper, that does not turn out to be so, but curious to hear what @emilevauge thinks about this.

server/server.go
aantono: just to clarify, are you asking to replace "DataDog Metrics" part of the string in the `log.Debugf` with `DataDog metrics` lowercased `metrics`?

timoreimann: Ah sorry, I think I was being too complicated: Yes, I just meant to use lower case in the log message.

I'm only familiar with Prometheus where I'm not aware of a proper name like `Prometheus Metrics`. If you could point me at any references (e.g., docs) showing this, I'll be happy to rest my case.

middlewares/datadog.go
timoreimann: An alternative to comparing the ticker could then be to use [`sync.Once`](https://golang.org/pkg/sync/#Once). This may also help us accidentally initializing the client more than once if for whatever reason we toggle the ticker value inadvertently.

middlewares/statsd.go
timoreimann: Alright, you convinced me: I failed to see that the config structs are of different types as well. With that in mind, extracting anything doesn't seem too useful.

Appreciate the technical analysis to solve the matter. 👏

middlewares/statsd.go
timoreimann: 👍

middlewares/datadog.go
aantono: `sync.Once` sounds like a good idea.  I've tried to add it, but realized that I still need to keep track of the `ticker` reference, so that I could stop it once the shutdown occurs.  Since I have to keep the `ticker` variable around anyways, do you want to introduce another variable to store the `once`, or just let it be and use the null check instead?

timoreimann: I think it's good enough in this case. Thanks for giving it a shot. 👍

aantono:
Sounds good... Looks like this is another opportunity for @ldez to shine and show off his mad "Glide" skills! 😜

middlewares/datadog.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/safe"
+	"github.com/containous/traefik/types"
+	kitlog "github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/metrics"
+	"github.com/go-kit/kit/metrics/dogstatsd"
+)
+
+var _ Metrics = (Metrics)(nil)
```
marco-jantke: Does this statement have any effect? I saw these kind of constructs to verify a concrete implementation is implementing an interface. To have the `Metrics` on the right hand side doesn't make any sense to me. Can you explain it or clean it up?

middlewares/datadog.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (dd *Datadog) getReqsCounter() metrics.Counter {
+	return dd.reqsCounter
+}
+
+func (dd *Datadog) getReqDurationHistogram() metrics.Histogram {
+	return dd.reqDurationHistogram
+}
+
+func (dd *Datadog) getRetryCounter() metrics.Counter {
```
marco-jantke: As `retryCounter` never gets initialised won't this lead to a panic when having DataDog implementation (same for StatsD) + Retries activated? Did you try this out?

middlewares/metrics.go
```diff
@@ -22,6 +23,52 @@ type RetryMetrics interface {
...
+
+func (mm *MultiMetrics) getReqDurationHistogram() metrics.Histogram {
+	return mm.reqDurationHistogram
+}
+
+func (mm *MultiMetrics) getRetryCounter() metrics.Counter {
+	return mm.retryCounter
+}
+
+func (mm *MultiMetrics) getWrappedMetrics() *[]Metrics {
```
marco-jantke: AFAICS this method is used nowhere.

middlewares/statsd.go
```diff
@@ -0,0 +1,85 @@
...
+	log.Info(keyvals)
+	return nil
+}))
+var statsdTicker *time.Ticker
+
+// Statsd is an Implementation for Metrics that exposes statsd metrics for the latency
+// and the number of requests partitioned by status code and method.
+// - number of requests partitioned by status code and method
+// - request durations
+// - amount of retries happened
```
marco-jantke: It does not expose `amount of retries happened` AFAICS. Same for DataDog implementation.

middlewares/datadog.go
aantono: @marco-jantke Great catch... the retry metric was added after the PR was already open, so I must've lost the change during one of the numerous rebases.  The good news is that the code won't `panic`, since there is a `m.retryMetrics != nil` check before the counter is used, but the bad news is that we are missing retry metrics for DataDog & StatsD.  Thanks a lot for noticing, I'll add those to the initialization block and open another PR.

middlewares/metrics.go
aantono: Yep, you are correct... Same PR that added retryMetrics, removed the `getWrappedMetrics()` method.  It was originally present on the initial `Metrics` interface.  Will clean-up

middlewares/datadog.go
aantono: The right hand side should have been `(*DataDog)(nil)`.  Fixed.

PR #1332, opened 2017-03-22
IP Whitelists for Frontend (with Docker- & Kubernetes-Provider Support)

MaZderMind:
This PR fixes #816. It adds the ability to specify a List of CIDR Whitelists per Frontend. If Whitelists are configured and the Request-IP does not match one of them, it is aborted with a 403 Forbidden.

This obviously requires traefik to be directly mapped on public ips, if the incoming requests are nat'ed, ie through dockers publish-functionality or docker-compose ports, traefik will only see an ip of the internal network and won't be able to filter according to the source-ip.

This PR adds support for the Docker-Provider and the Kubernetes-Provider, but it should be trivial to extend the support to other Providers, I just have no test-setup for those. For both it parses the Annoatation/Label "traefik.frontend.whitelistSourceRange", for Kubernetes it also accepts the quasi-standard "ingress.kubernetes.io/whitelist-source-range". They can take a Comma-Separated List of CIDR Nets (IPv4 & v6), like this:
```yml
machine:
  image: katacoda/docker-http-server
  labels:
    - "traefik.backend=local-only-echo"
    - "traefik.frontend.whitelistSourceRange="10.42.0.0/16, 152.89.1.33/32, afed:be44::/16"
    - "traefik.frontend.rule=Host:local-only-echo.example.com"
```

or for Kubernetes:
```yml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: hello-minikube
  annotations:
    traefik.frontend.whitelistSourceRange: "10.42.0.0/16, 152.89.1.33/32, afed:be44::/16"
spec:
  rules:
  - host: echo.kube
    http:
      paths:
      - path: /
        backend:
          serviceName: hello-minikube
          servicePort: web
```

For File-Based Config, there's the `whitelistSourceRange` property, which takes an array of CIDR Net-Strings:
```toml
[frontends]
  [frontends.frontend2]
  backend = "backend1"
  passHostHeader = true
  priority = 10
  whitelistSourceRange = ["10.42.0.0/16", "152.89.1.33/32", "afed:be44::/16"]
```

The applied Whitelists are Displayed on the Dashboard:
whitelists-dashboard

For my own personal use and for easy testing, I pushed a Docker-Image based on this PR:
https://hub.docker.com/r/mazdermind/traefik/tags/

I'd be happy to receive some feedback, because Go is still fairly new.
Best Regards,
Peter

PR #1702, opened 2017-06-02
Add whitelist configuration option for entrypoints

christopherobin:
### Description

This re-uses work done on #1332 and allows users to define IP whitelisting at the entrypoint level.

```toml
[entryPoints]
  [entryPoints.http]
  address = ":1080"
  whiteListSourceRange = ["192.168.0.0/16"]
```

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -279,6 +279,12 @@ To write JSON format logs, specify `json` as the format:
 #   address = ":80"
 #   compress = true
 
+# To enable IP whitelisting at the entrypoint level:
+# [entryPoints]
+#   [entryPoints.http]
+#   address = ":80"
+#   whiteListSourceRange = ["127.0.0.1/32"]
+
 [entryPoints]
   [entryPoints.http]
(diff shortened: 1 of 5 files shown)


nmengin:
@christopherobin Thanks for this feature!

Thanks to it, users will be able to centralize all their commons white listed IP into the entrypoint section.
Moreover, it can be a workaround for providers which are not supported by the initial PR #1332( by waiting the implementation :wink: ).

LGTM

christopherobin:
@emilevauge Added support for it, though I was not sure about changing the flag description since it's already pretty long.

As for entrypoint flags maybe something like that would make them easier to read/parse:

```bash
traefik \
  --entrypoint=http \
  --entrypoint.http.address=:8000 \
  --entrypoint.http.whiteListSourceRange=192.168.0.0/16 \
  --entrypoint=http_public \
  --entrypoint.http_public.address=:8001
```

Not sure if any flag library would support this kind of stuff though

PR #1706, opened 2017-06-03
Update the documentation to use DaemonSet or Deployment

saschagrunert:


saschagrunert:
Thank you for your comment. What is the use case in using deployments in favor to daemon sets?

I think the most beneficial thing of DaemonSets is that **every** node in the cluster routes the traffic to the right destination, correct? With deployments I would need a NodePort/LoadBalancing service to achieve the same behavior.

I am not sure. I guess it should be an easy documentation with no pitfalls. 😶

timoreimann:
@saschagrunert personally, I don't think running a Daemonset with a fixed host port and a Service with a NodePort are largely different. The NodePort is more flexible while the Daemonset could be considered more "discoverable".

One key difference to me is that with a Daemonset, by default you have to buy into the one Pod per node model. This may be just the right thing for you; however, there's also a chance you're running either too many replicas (thereby wasting resources) of too few. The former can be solved by restricting on which nodes your Daemonset should run on (likely using taints and tolerations), the latter requires you to add more nodes. Both seem more heavy-weight than what a Deployment can give you where scaling is much easier (and even automatable through mechanisms like HPA).

On the plus side, Daemonsets allow you to run Pods on a dedicated set of nodes. While wasteful on the resource end, it enables better isolation. We actually run Traefik in that mode, but we consider it an intermediate solution until we're confident that running Traefik amidst regular Pods won't have any major undesirable effects.

So that's my take on the subject. :-)

timoreimann:
I forgot to mention another thing: Daemonset can be rolling upgraded just with Kubernetes 1.6. So folks still running on 1.5 or lower might have a harder time upgrading to newer versions of Traefik or updating the configuration.

saschagrunert:
This sounds all very reasonable to me. :) The performance scalability arguments is one of the keys here I guess. So, when having a deployment should be the "best practice", can we change the `kuubectl create -f ...` commands to actually target deployments? 🙂

timoreimann:
Sorry, which `kubectl create` command are you referring to exactly?

saschagrunert:
I mean the link to the yaml file comtains currently a  DaemonSet instead of a deployment: https://raw.githubusercontent.com/containous/traefik/master/examples/k8s/traefik.yaml

timoreimann:
@saschagrunert true, thanks for pointing out.

@errm I think you changed the k8s guide last. While http://docs.traefik.io/user-guide/kubernetes/#deploy-trfik-using-a-deployment-object talks about Deployments, the [example file hosted](https://github.com/containous/traefik/tree/master/examples/k8s/traefik.yaml) actually refers to a DaemonSet?

timoreimann:
I had a discussion with @dtomcej, and we both agree that both Deployment and DaemonSet should be documented as valid options.

@saschagrunert please let me know if you'd be willing to change the PR accordingly. Thanks!

timoreimann:
Other documentation "bug" is mentioned here: https://github.com/containous/traefik/pull/1729#issuecomment-307487327

saschagrunert:
Yes, I am. Actually I guess this can be done by me during the next week.

timoreimann:
@saschagrunert very cool! 👍  looking forward to it!

saschagrunert:
Closed in favor to #1735

PR #1707, opened 2017-06-03
traefik controller access to secrets

spinto:
The traefik controller shall have access to secrets for the k8s basic authentication (#1488) to work

### DIFF
--- a/examples/k8s/traefik-rbac.yaml
+++ b/examples/k8s/traefik-rbac.yaml
@@ -10,6 +10,7 @@ rules:
       - pods
       - services
       - endpoints
+      - secrets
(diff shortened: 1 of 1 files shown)


timoreimann:
I wonder if we should formulate this in a more conditional manner. Presumably, secrets access will only be required by those that actually use basic authentication?

timoreimann:
Answering myself:

>I wonder if we should formulate this in a more conditional manner. Presumably, secrets access will only be required by those that actually use basic authentication?

Sadly, Traefik does not conditionally watch for secrets as of now. See also #1784.

So this one seems okay.

Issue #1574, opened 2017-05-09
Sticky session specify cookie name


kleinsasserm:
In some scenarios it is required to specify the cookie name which is used by Traefik for session stickiness. 

For example: We would like to provide a swarm stack with an ingress Traefik loadbalancer which works without any problem but we have to provide a second Traefik that runs on a separate docker host with additional ip addresses to use port 80 and 443 (DNS entries ...). In this setup the sticky cookie setting from the first boundary loadbalancer "overwrites" the sticky cookie from the second loadbalancer because of the same name. The result is that the cookie values are garbled as you can see this in the screenshot.

![09-05-_2017_14-59-06](https://cloud.githubusercontent.com/assets/9364572/25852149/a0fe94fe-34c8-11e7-9ecf-605970274e0c.jpg)

If we use a Nginx loadbalancer as first stage, the stickiness works without any troubles. Therefore an option like stickycookie="_MYFUNNYCOKKIE" would be helpful to configure cascading Traefik setups.

ldez:
Seems to be a duplicate of #1207

kleinsasserm:
I guess no, because #1207 describes possible types stickiness (cookie, ipaddress). Here I mean, that if maybe type "cookie" is specified, it would be helpful to specify the cookie name. So this is an extension to type cookie.

timoreimann:
@ldez I agree with @kleinsasserm that the request is different from #1207. Okay to remove the duplicate status?

dtomcej:
@timoreimann I agree that those 2 issues are separate. I also think that the cookie name should be configurable, as:

1. Other proxies (like nginx or haproxy) allow for this, and:
2. Allowing the cookie name to be customized would allow applications to control it as well.

richardjq:
Another use case is if you are using Path based frontend rules on the same domain. The cookie for service A will be overwritten by the cookie for service B.

This is a very common use case in our Docker environment.

ldez:
Closed by  #1716

PR #1712, opened 2017-06-06
Set cookie name for docker backends

richardjq:
### Description

Modification to allow you to specify the cookie name for a docker container using `traefik.backend.loadbalancer.cookiename=somecookie`.

Addresses issue #1574 for docker backends.


timoreimann:
Sounds reasonable to me. @emilevauge, WDYT?

emilevauge:
Instead of using a specific Docker label, WDYT of using a cookie named `_TRAEFIK_BACKEND_{backend-name}={server-name}` ?

emilevauge:
@richardjq it should also covers the use case in #1574 :)

kleinsasserm:
Yes, this will cover my issue also. The only requirement is, that the cookie is uniqe per Traefik instance. If {server-name} reflects the dynamic Traefik docker container hostname, it will be a valid cookie naming.

richardjq:
@emilevauge - so do you want to reject this PR and I'll see if I can generate a new PR to generate a new one with `_TRAEFIK_BACKEND_{backend-name}`?

Or do you prefer to keep the option to set the cookie name and make an additional change for the default to be `_TRAEFIK_BACKEND_{backend-name}`?

ldez:
@richardjq 👍  we close this PR and you create a new PR.

richardjq:
Done. See #1716.

Thanks!

PR #1713, opened 2017-06-07
Updated kubernetes user-guide

vongohren:
Updated kubernetes userguide readme with the acctualy yamlfiles on the linked github. Also rewrote the part about deployment, when its used a daemonset instead

### Description

After running the userguide, things was not working, because the markdown did not match the github content.


ldez:
Contains some similar changes as #1706

vongohren:
Well thats perfect, then I can close this, just did as directed to on the slack: "@vongohren would you mind submitting a PR to fix this?":D

PR #1716, opened 2017-06-07
make the cookie name unique to the backend being served

richardjq:
### Description

Set cookie name so that it is unique to the backend.

Addresses issue #1574.

See PR #1712 for additional discussion.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -641,7 +641,7 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 					}
 
 					stickysession := configuration.Backends[frontend.Backend].LoadBalancer.Sticky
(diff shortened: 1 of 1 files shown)


indrekj:
This makes upgrading traefik version harder because it looks not to be backwards compatible. Imagine an ongoing sticky websocket connection. On upgrade it will reconnect but does not have this cookie any more. So the reconnected user may end up on a different server. Also gradual rollout where you have some of the instances running old version and some new may cause some issues.

This is not a problem for my case as we're also using custom header stickyness https://github.com/salemove/traefik/commit/9f8adf2b30baec6eebb58fad862dc7760c1de17c but may be a problem for others.

Issue #1722, opened 2017-06-08
Add support for Kubernetes' `ingress.kubernetes.io/rewrite-target`


mlaccetti:
Feature request:

Add support for `ingress.kubernetes.io/rewrite-target` - the ability to strip a path off the frontend URL when passing it to the backend.

I.e. if I request `foo.bar/api/` and I have the following ingress annotation `ingress.kubernetes.io/rewrite-target: /`, I'd expect Traefik to add in a rule for the frontend to replace `/api/` with `/`.

kachkaev:
Should this be closed given that the relevant PR has been merged?

timoreimann:
Yep, I think so. Thanks.

kolleroot:
Hi there, I'm currently testing the rc and I'm not quite sure, that I understand this feature. Does this mean that when requesting /api/v1/some-resource that the url will be rewritten to /v1/some-resource? The behavior, that I observe is, that all urls will be rewritten to /

timoreimann:
@kolleroot The _path_ (not the entire URL) should be rewritten to whatever you have specified in the annotation. Technically, the feature is implemented in terms of the [`ReplacePath` modifier](https://docs.traefik.io/basics/#modifiers).

If you have further questions please don't hesitate to reach out to the Traefik community either on Slack or Stackoverflow. Thank you!

Issue #563, opened 2016-07-28
Need URL rewrite to add trailing slash


javefang:
I have a simple app, which has the following file structure at root
- script.js
- style.css
- index.html (load the other two files using relative path `script.js` and `style.css`)

Since I want to access the app via URL `http://example/app`, I proxied the web app with rule `PathStripPrefix:/app`. The problem is when I try to access URL `http://example/app` (without trailing slash), it will load "index.html" fine, but not the JS and CSS file. When I look into the debugger, it tries to load:
- http://example/script.js
- http://exmaple/style.css

Instead of (the correct one):
- http://example/app/script.js
- http://example/app/style.css

It only works when I type the original URL with a trailing slash, so `http://example/app/`. This is not a big deal for me but users sometimes find it annoying since we used to use Nginx, who sends an "301 Moved Permanently" to a URL with trailing slash when it's not there. I wonder if it is possible / makes sense to implement this in Traefik?

Thank you!

godliness:
same issue, anyone have idea?

godliness:
same issue, anyone have idea?

godliness:
same issue, anyone have idea?

godliness:
same issue, anyone have idea?

godliness:
same issue, anyone have idea?

godliness:
same issue, anyone have idea?

javefang:
Think should be a feature. Although you.probably want it most of the time, and begins does it by default if you use it as reverse proxy.

Expected behaviour is explained in my original post.

royjs:
I also have the same issue

emilevauge:
Hi, this is due to the underlying muxer used in traefik http://www.gorillatoolkit.org/pkg/mux#Router.StrictSlash. We use it with `StrictSlash` enabled but there is an exception:
>Special case: when a route sets a path prefix using the PathPrefix() method, strict slash is ignored for that route because the redirect behavior can't be determined from a prefix alone. However, any subrouters created from that route inherit the original StrictSlash setting.

We need to work on this indeed...
As a workaround for now, I suggest to use `Host` rules instead of `Path` based rules if possible.
Any help on this is welcome ;)

ThinkBriK:
I was able to workaround this issue using the regex on the entrypoint:
```
defaultEntryPoints = ["http"]
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
[...]

PR #1723, opened 2017-06-08
kubernetes ingress rewrite-target implementation

mlaccetti:
Monitor the kubernetes ingress for the `ingress.kubernetes.io/rewrite-target` annotation, and if seen, configure the frontend to strip off the provided path, and replace with the one provided in the annotation.


timoreimann:
@mlaccetti do you possibly know if the `ingress.kubernetes.io/rewrite-target` annotation is somewhat of a standard, or rather something that comes from a specific controller implementation (presumably Nginx)?

/cc @dtomcej @errm; also @emilevauge who I remember was involved in a Slack discussion on this topic fairly recently.

ldez:
Related to #1722

mlaccetti:
@timoreimann The spec doesn't have any annotations listed, but I'm finding them cropping up in places, and it seemed like low-hanging fruit.

It seems that folks have started using the nginx ingress as a defacto standard, rather than actually chasing down the spec.  I'm going to chase that one down to figure out what the long term plan is.

ldez:
See https://github.com/kubernetes/ingress/blob/master/controllers/nginx/configuration.md

dtomcej:
Should we write this as a project-wide rule type?

timoreimann:
The annotations look more like de-facto standards to me. Regardless, I'm okay with supporting it.

I'd probably scope it to Kubernetes for now. If there's demand for spreading, we can have further PRs.

provider/kubernetes/kubernetes.go
```diff
@@ -211,8 +214,21 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					switch {
+					case ruleType == ruleTypeReplacePath:
+						templateObjects.Frontends[r.Host+pa.Path].Routes[pa.Path] = types.Route{
+							Rule: ruleType + ":" + rewriteTarget,
+						}
+
+					default:
+						templateObjects.Frontends[r.Host+pa.Path].Routes[pa.Path] = types.Route{
+							Rule: ruleType + ":" + pa.Path,
+						}
```
timoreimann: To me it looks like we could simplify this block. How about:

```go
rule := ruleType + ":" + pa.path

if rewriteTarget := i.Annotations[annotationKubernetesRewriteTarget]; rewriteTarget != "" {
  rule = ruleTypeReplacePath + ":" + rewriteTarget
}

templateObjects.Frontends[r.Host+pa.Path].Routes[pa.Path] = types.Route{
  Rule: rule,
}
```

provider/kubernetes/kubernetes_test.go
```diff
@@ -1555,6 +1555,34 @@ func TestIngressAnnotations(t *testing.T) {
 					},
 				},
 			},
+		}, {
+			ObjectMeta: v1.ObjectMeta{
+				Namespace: "testing",
+				Annotations: map[string]string{
+					"kubernetes.io/ingress.class":          "traefik",
```
timoreimann: We don't need the Ingress class for the test. Please remove it.

provider/kubernetes/kubernetes.go
timoreimann: Could we move the logic pertaining to setting the route type to a dedicated function to ease readability?

provider/kubernetes/kubernetes_test.go
```diff
@@ -1718,14 +1759,27 @@ func TestIngressAnnotations(t *testing.T) {
...
+			},
 		},
 	}
 
 	actualJSON, _ := json.Marshal(actual)
 	expectedJSON, _ := json.Marshal(expected)
 
 	if !reflect.DeepEqual(actual, expected) {
-		t.Fatalf("expected %+v, got %+v", string(expectedJSON), string(actualJSON))
+		t.Fatalf("expected %+v \n got %+v", string(expectedJSON), string(actualJSON))
```
timoreimann: We should stop comparing JSON representations as they can hide differences in very subtle ways. For instance, an empty slice and a `nil` one would not be `DeepEqual` but look the same in the JSON output.

Either compare on the `actual` and `expected` structs, or use `assert.Equal()`.

timoreimann:
Related discussion on the Kubernetes mailing thread I just ran into: https://groups.google.com/d/topic/kubernetes-users/jkaCCykAwXA/discussion

(**tl;dr:** The current naming isn't too well-defined.)

ldez:
@timoreimann design review before review please. 😉

timoreimann:
@ldez true. however, the only open point left seems to be the one from [@dtomcej's comment](https://github.com/containous/traefik/pull/1723#issuecomment-307455963), which should be unrelated to my comments.

dtomcej:
My concern is that every other rule we have is project wide. It would seem weird IMO to have rule types lumped in that ONLY work on kubernetes ingresses.

timoreimann:
I agree that it'd look weird on first sight. IMHO though, we might lose more than we can win by spreading the rule for a few reasons:

- We have already established a common, consistent naming scheme across all providers (mostly surrounding the notion of paths) which the new name doesn't fit in well (which obviously is because it comes from an external source). I can imagine non-Kubernetes users being irritated by this, and our only answer would be "it's because of Kubernetes".
- Even if we made it a global rule, we likely can't even make it consistent within Traefik: The `ingress.kubernetes.io/` prefix would only exist with Kubernetes while all the other providers would come without it. I'm concerned this may   lead to more confusion.
- The way I interpreted the mailing thread I posted above is that the Kubernetes maintainers themselves aren't overly happy with the naming. To me there's still room for the naming to change -- at least, it's nothing like an official standard for now. So I'd rather not have it spread to other providers until there's reasonable certainty.

Regardless of how we decide on this point, do we agree that this PR is okay to proceed? AFAICS, there's nothing in it that would affect any future decision to the extent we're concerned about. Maybe we can then postpone the bigger decision to a dedicated issue?

provider/kubernetes/kubernetes.go
mlaccetti: Simplified

provider/kubernetes/kubernetes_test.go
mlaccetti: Removed

provider/kubernetes/kubernetes_test.go
mlaccetti: Cut over to asserting on the structs.

provider/kubernetes/kubernetes.go
mlaccetti: Moved it into a function.

mlaccetti: Moved it into a function.

mlaccetti:
@timoreimann I've made the changes, though I have no idea if I made it any better. :D

timoreimann:
@mlaccetti seems better on first sight to me. :-)

@dtomcej WDYT, can we treat the should-or-should-not-be-a-project-wide-rule-type question separate from the functionality this PR provides?

dtomcej:
@timoreimann Yes, that discussion should be a separate Proposal for handling these sort of annotation updates. More will come.

provider/kubernetes/kubernetes.go
```diff
@@ -294,6 +291,25 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 	return &templateObjects, nil
 }
 
+func setPath(pa v1beta1.HTTPIngressPath, i *v1beta1.Ingress, templateObjects types.Configuration, r v1beta1.IngressRule) {
+	if len(pa.Path) > 0 {
```
timoreimann: Can we reduce the set of input parameters to the bare minimum needed for the implementation? For instance, it looks like we don't need the entire Ingress struct, just the annotations.

provider/kubernetes/kubernetes.go
```diff
@@ -294,6 +291,25 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 	return &templateObjects, nil
 }
 
+func setPath(pa v1beta1.HTTPIngressPath, i *v1beta1.Ingress, templateObjects types.Configuration, r v1beta1.IngressRule) {
```
timoreimann: Another approach to implement this function might be to just have it return the computed rule and let the caller assign it to the configuration. That way, we don't need to pass in the configurations.

provider/kubernetes/kubernetes_test.go
```diff
@@ -1718,15 +1759,23 @@ func TestIngressAnnotations(t *testing.T) {
...
 		},
 	}
 
-	actualJSON, _ := json.Marshal(actual)
-	expectedJSON, _ := json.Marshal(expected)
-
-	if !reflect.DeepEqual(actual, expected) {
-		t.Fatalf("expected %+v, got %+v", string(expectedJSON), string(actualJSON))
-	}
+	assert.Equal(t, expected, actual, "The rewritten data did not match the expected.")
```
timoreimann: I think we can drop the extra message. testify will tell us by default that the expected and actual values mismatch.

ldez:
@mlaccetti Any news?

mlaccetti:
Let me kick the code to address the last round.

provider/kubernetes/kubernetes.go
mlaccetti: I twiddled the implementation from `setPath` to `getRuleForPath`, and, assuming a rule is found, apply it.  That should address the two issues you highlighted.

provider/kubernetes/kubernetes_test.go
mlaccetti: Removed the extra message - legacy of JUnit where you don't always get a "nice" error message, and spitting out some words helped make more sense of things.

mlaccetti:
@ldez @timoreimann Hopefully that brings things up to par!

provider/kubernetes/kubernetes.go
```diff
@@ -294,6 +296,25 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 	return &templateObjects, nil
 }
 
+func getRuleForPath(pa v1beta1.HTTPIngressPath, i *v1beta1.Ingress) string {
+	if len(pa.Path) > 0 {
```
timoreimann: Can we check for `== 0` and return an empty string right away? That saves us a level of indentation and keeps the value returned in case of an empty path closer to its detection.

mlaccetti: Done and done!

mlaccetti:
@timoreimann The build is failing, but looks like a glitch with Semaphore CI - any chance you can kick it? Otherwise, I'll try to find some code to change, then revert, just to force it.

timoreimann:
@mlaccetti updating the branch should fix things. could you do that please?

timoreimann:
@mlaccetti please rebase instead of merging.

mlaccetti:
@timoreimann Sorry, hitting the `update branch` button in GitHub, seems to do a merge instead of a rebase.  (And, to be honest, I haven't learned how to do a rebase, either.)

Hopefully somebody can merge the PR, now!

timoreimann:
@mlaccetti no worries. PR looks good now!

By the way: rebasing is as simple as doing the following two git operations:

1. `git fetch origin master` (or whatever your local repository name for `containous/traefik` is).
2. (in your local PR branch:) `git rebase origin/master`

That'll unwrap all local commits from your PR branch temporarily; fast-forward the branch to the latest of `upstream/master`; and reapply your local commits again. Naturally, you may have to resolve conflicts just like with a regular merge (though in a rebase, the roles of "mine" and "theirs" are swapped).

timoreimann:
@dtomcej @errm and others: PTAL. :-)

kachkaev:
@mlaccetti could you please consider adding an example for `rewrite-target` to https://github.com/containous/traefik/blob/master/docs/configuration/backends/kubernetes.md? Do you think it could solve the trailing slash issue? https://github.com/containous/traefik/issues/563#issuecomment-317236171, https://github.com/containous/traefik/issues/563#issuecomment-322412874?

mlaccetti:
@kachkaev I'll see if I can find some time this week to add the example in - could arguably solve the problem - just have two ingresses! :)

kachkaev:
ping @mlaccetti :–)

ldez:
@kachkaev look at #2676

PR #1726, opened 2017-06-09
Update README.md

marcosnils:

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ Run it and forget it!
 
 ## Quickstart
 
-You can have a quick look at Træfik in this [Katacoda tutorial](https://www.katacoda.com/courses/traefik/deploy-load-balancer) that shows how to load balance requests between multiple Docker containers.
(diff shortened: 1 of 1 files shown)


timoreimann:
I'm okay with this, though at some point we may want to maintain a list of blog posts / external articles and just point to it from the quick start.

@emilevauge @ldez thoughts?

marcosnils:
Thx guys, it there some other tutorial that you think it would be worth to showcase for your users to play with?.    We think it's pretty cool that traefik users can learn by leveraging an interactive playground.

emilevauge:
@marcosnils indeed, we love interactive playgrounds. I'm thinking for example at play-with-k8s ;)

marcosnils:
@emilevauge got it. We're setting up training.play-with-k8s.com the same as Docker's tutorials. Do you have some content I can use to setup a K8s traefik tutorial?

emilevauge:
@marcosnils Awesome :heart_eyes: 
Here we go:
http://docs.traefik.io/user-guide/kubernetes
https://medium.com/@patrickeasters/using-traefik-with-tls-on-kubernetes-cb67fb43a948

PR #1729, opened 2017-06-09
Update documentation in kubernetes setup

matthew-marchetti:
### Description

This PR updates the docs to use a daemonset for deployments, and adds missing annotations to the kubernetes example files.

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -68,10 +68,10 @@ subjects:
 kubectl apply -f https://raw.githubusercontent.com/containous/traefik/master/examples/k8s/traefik-rbac.yaml
 ```
 
-## Deploy Træfik using a Deployment object
+## Deploy Træfik using a DaemonSet
 
 We are going to deploy Træfik with a
-[Deployment](http://kubernetes.io/docs/user-guide/deployments/), as this will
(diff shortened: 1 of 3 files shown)


ldez:
Seems a duplicate of  #1706

matthew-marchetti:
@ldez i don't agree -- His missing 2 files out of his commit that also need to be updated. Sure -- The main file is very similar, but If anything, this is a more completed version of what needs to be updated.

ldez:
@matthew-marchetti You can do a review in #1706 and suggest your changes.

ldez:
We (maintainers) are currently speaking about the better way to document k8s.
News coming soon.

matthew-marchetti:
@ldez not to chain onto this, but the https://docs.traefik.io/user-guide/kubernetes/ docs are generated directly from this correct ?? I'm noticing a discrepancy between the two for the     "traefik.frontend.rule.type: PathPrefixStrip" option. The online docs seems to be ignoring the case, which breaks it when used in the lowercase form. Any idea on what that would be ? -- The value looks correct in my PR ?

timoreimann:
@matthew-marchetti the case sensitivity used to be irrelevant for Kubernetes, but it's not anymore.

thanks for pointing out! we need to fix that too.

PR #1735, opened 2017-06-11
Update the documentation to use DaemonSet or Deployment

saschagrunert:
From the discussion of  #1706 

Waiting for your review. 👍


docs/user-guide/kubernetes.md
```diff
@@ -68,11 +68,19 @@ subjects:
...
-## Deploy Træfik using a Deployment object
+## Deploy Træfik using a Deployment or DaemonSet
 
-We are going to deploy Træfik with a
-[Deployment](http://kubernetes.io/docs/user-guide/deployments/), as this will
-allow you to easily roll out config changes or update the image.
+It is possible to use Træfik with a
+[Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/)
+or a [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/)
+object, whereas both options have their own pros and cons: The scalability os much better when
```
timoreimann: Typo: os -> is

docs/user-guide/kubernetes.md
```diff
@@ -68,11 +68,19 @@ subjects:
...
+using a Deployment, because you will have a Single-Pod-per-Node model when using
+the DeaemonSet. On the other hand the DaemonSet allows you to access any Node
+directly on Port 80 and 443, where you have to setup a 
+[Service](https://kubernetes.io/docs/concepts/services-networking/service/) object
+with a Deployment.
```
timoreimann: We could also mention that a Daemonset allows you to exclusively run a service (like Traefik) on a dedicated set of machines using taints and tolerations.

docs/user-guide/kubernetes.md
```diff
@@ -112,66 +120,147 @@ spec:
...
+apiVersion: v1
+metadata:
+  name: traefik-ingress-service
+spec:
+  selector:
+    k8s-app: traefik-ingress-lb
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 80
```
timoreimann: AFAIR,  we don't need to specify a target port when it matches the service port.

docs/user-guide/kubernetes.md
```diff
@@ -112,66 +120,147 @@ spec:
...
+    - protocol: TCP
+      port: 8081
```
timoreimann: I think we should just use 8080, which happens to be the default port for Traefik's administration interface.

docs/user-guide/kubernetes.md
```diff
@@ -112,66 +120,147 @@ spec:
...
+            cpu: 100m
+            memory: 20Mi
         ports:
-        - containerPort: 80
+        - name: http
+          containerPort: 80
           hostPort: 80
-        - containerPort: 8080
+        - name: admin
+          containerPort: 8081
```
timoreimann: 8080 would be my preference here too.

docs/user-guide/kubernetes.md
```diff
@@ -112,66 +120,147 @@ spec:
...
+        securityContext:
+          privileged: true
         args:
+        - -d
         - --web
+        - --web.address=:8081
```
timoreimann: We don't need this after switching to 8080.

docs/user-guide/kubernetes.md
```diff
@@ -112,66 +120,147 @@ spec:
...
         - --kubernetes
 ```
-[examples/k8s/traefik.yaml](https://github.com/containous/traefik/tree/master/examples/k8s/traefik.yaml)
+[examples/k8s/traefik-ds.yaml](https://github.com/containous/traefik/tree/master/examples/k8s/traefik-ds.yaml)
 
-> notice that we binding port 80 on the Træfik container to port 80 on the host.
-> With a multi node cluster we might expose Træfik with a NodePort or LoadBalancer service
-> and run more than 1 replica of Træfik for high availability.
+To deploy Træfik to your cluster start by submitting one of the yaml files to the cluster with `kubectl`:
```
timoreimann: s/yaml/YAML/

docs/user-guide/kubernetes.md
saschagrunert: Alright, I add a note about this.

saschagrunert:
Changed the files regarding your comments. :)

docs/user-guide/kubernetes.md
```diff
@@ -68,11 +68,20 @@ subjects:
...
+It is possible to use Træfik with a
+[Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/)
+or a [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/)
+object, whereas both options have their own pros and cons: The scalability is much better when
+using a Deployment, because you will have a Single-Pod-per-Node model when using
+the DeaemonSet. It is possible to exclusively run a Service on a dedicated
+set of machines using taints and tolerations with a DaemonSet. On the other hand the
+DaemonSet allows you to access any Node directly on Port 80 and 443, where you have to setup a 
+[Service](https://kubernetes.io/docs/concepts/services-networking/service/) object
+with a Deployment.
```
errm: I would add something here that deployments being suited in environments where there is some sort of external loadbalancer (Like AWS or GCP) to forward traffic to the nodeport.

I would also mention that Deployment offers the facility to upgrade the version / config in place. Whereas DaemonSet offers no ability to do this, short of manually deleting the pods on each node...

I think mostly the DaemonSet option is applicable to development environments like minikube, or bear metal, or some other usecase where you want to expose the hosts directly...

docs/user-guide/kubernetes.md
```diff
@@ -112,66 +121,143 @@ spec:
...
+
 > We expect to see a 404 response here as we haven't yet given Træfik any configuration.
 
 ## Deploy Træfik using Helm Chart
 
-Instead of installing Træfik via a Deployment object, you can also use the Træfik Helm chart.
+Instead of installing Træfik via an own object, you can also use the Træfik Helm chart. This
+allows more complex configuration via Kubernetes
+[ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configmap/) and enabled
+TLS certificates.
```
errm: Is this true, I don't think we support Kubernetes provided TSL certificates yet...

docs/user-guide/kubernetes.md
dtomcej: Daemonsets have a rolling deployment option in 1.6+ ;)

docs/user-guide/kubernetes.md
```diff
@@ -112,66 +121,143 @@ spec:
...
+            cpu: 100m
+            memory: 20Mi
         ports:
-        - containerPort: 80
+        - name: http
+          containerPort: 80
           hostPort: 80
-        - containerPort: 8080
+        - name: admin
+          containerPort: 8080
```
timoreimann: This one should probably keep its `hostPort` for symmetry reasons.

examples/k8s/traefik-ds.yaml
```diff
@@ -37,11 +37,10 @@ spec:
           containerPort: 80
           hostPort: 80
         - name: admin
-          containerPort: 8081
+          containerPort: 8080
```
timoreimann: `hostPort` here too probably.

docs/user-guide/kubernetes.md
Regner: Shouldn't this be exposed with ingress via Traefik?

docs/user-guide/kubernetes.md
saschagrunert: What do you mean? The helm chart support indeed TLS termination at the ingress.

docs/user-guide/kubernetes.md
saschagrunert: It can, but this is not necessarily needed. It makes the configuration easier.

examples/k8s/traefik-ds.yaml
timoreimann: So shall we add it?

dtomcej: Honestly, according to the kubernetes Configurations best practices documentation (https://kubernetes.io/docs/concepts/configuration/overview/) it clearly states:

`Don’t use hostPort unless it is absolutely necessary`. They recommend using `kubectl proxy` or `kubectl port-forward`. I think if we are providing an example for new users to try, we should at least follow best practices. MANY users have been bitten by hostport usage.

timoreimann: I agree that host ports are tricky in general; but haven't we accepted the deal already when we picked the DaemonSet? After all, we want to reap the benefits of easy accessibility to a known, static port (and, in particular, not add another Service configured with a NodePort).

The `kubectl`-based approaches are recommended "for debugging purposes". I don't see the admin interface as something you'd only need for rare-ish debug cases, but opinions may vary on this point. To the least, I find it inconsistent that we expose the admin port on the Deployment (through a NodePort) but do not do the same on the DaemonSet. I'm concerned this may lead to confusion.

docs/user-guide/kubernetes.md
```diff
@@ -112,66 +121,143 @@ spec:
...
+        securityContext:
+          privileged: true
```
timoreimann: Do we actually need this? It's not on the Deployment, to the least.

examples/k8s/traefik-ds.yaml
dtomcej: I agree that Hostport is convenient and easy, but many cloud providers don't allow hostport without escalating privileges. Not sure if this is something we should be "encouraging".

I would be happiest keeping the same nodeports through our examples. Nodeport is fully supported, and pretty easy for a newbie to understand.

docs/user-guide/kubernetes.md
dtomcej: Yes, Hostport on a privileged port requires this IIRC

saschagrunert:
I changed the documentation to use the nodeport service instead of host ports. Please have a look again. :)

timoreimann:
If we use Service NodePorts on DaemonSets too, what benefits do we see left that justify using/documenting them at all?

To me they are most valuable when combined with nodeSelectors / taints and tolerations so that you can define a dedicated set of pod-hosting nodes, obviously in our case for Traefik. In such scenarios, you probably wouldn't have to worry about conflicting ports and would be just fine with using host ports, especially when it improves discoverability (port 80 and 8080 are super easy to remember and stable even if you kill your NodePort-providing Service accidentally).

@dtomcej @saschagrunert WDYT?

saschagrunert:
Yes, for me the DS host port is an as valid solution as the Deployment + NodePort Service. I just reverted the last commit. But we have to come to a clear point what we want to add to the documentation for now.

timoreimann:
@dtomcej your turn again. :-)

dtomcej:
IMO, We should have the Nodeport service in front of both.

The reason is that we shouldn't require knowledge about the kubernetes cluster to determine where traffic is routed.

Example: You configure a daemonset to only run on certain nodes through labels/taints. With hostPort, you need to know which nodes the pods are running on before you can forward traffic. Furthermore, this can change on the fly by editing the daemonset spec.

Keeping the nodeport service in front of the Daemonset will alleviate these issues.

I personally think that hitting port 30080 is no harder than 80, and keeps it clear that you are hitting the SERVICE port, and not the pod port etc. New users are often confused how overlay networks etc work.

I think that we need to clearly outline the differences between the two types:

Deployment:

- Scalable up and down easily
- Can implement full pod lifecycle
- Implemented rolling deploy from K8s 1.2+
- Minimum 1 pod

Daemonset

- Automatically scales to all nodes that meet selector
- Guaranteed to fill nodes one at a time, and not "double up" on a node

Both have uses, and I personally run clusters with both.

timoreimann:
Alright @dtomcej, that does make a lot of sense. You convinced me. :-)

Just one question: By "Can implement full pod lifecycle", you probably mean everything related to rolling upgrades? AFAIK, Kubernetes 1.7 brings DaemonSets on par in this regard.

@saschagrunert could you incorporate the points Daniel made (maybe except for the lifecycle one until we've clarified)?

dtomcej:
@timoreimann Correct. Pre-1.6 Daemonsets did not have rolling restarts, rollbacks, history, image changes, etc.

I agree that my term "lifecycle" was incorrect, as yes, Daemonsets still can implement the lifecycle hooks for containers.

examples/k8s/traefik-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: traefik-ingress-service
+spec:
+  selector:
+    k8s-app: traefik-ingress-lb
+  ports:
+    - protocol: TCP
```
dtomcej: Need a name for the ports.

examples/k8s/traefik-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+        name: traefik-ingress-lb
+        args:
+        - --web
+        - --kubernetes
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: traefik-ingress-service
+spec:
```
dtomcej: Also need to be part of kube-system due to namespaced pods

ldez:
@saschagrunert Any news on this PR?

docs/user-guide/kubernetes.md
rajivm: @dtomcej Per your earlier comment, if you used a more arbitrary port `30080`, this would not need to be privileged. In most cases, since there is still an `ELB` or equivalent in front of `traefik`, this shouldn't be an issue.

docs/user-guide/kubernetes.md
errm: Yes you are correct ... I forgot since until recently I am still running a pre 1.6 cluster (for reasons)

docs/user-guide/kubernetes.md
errm: Yes but I don't think that you are currently able to provide a TLS cert as a kubernetes secret..

saschagrunert: Why not? You can create tls secrets with something like:
```
kubectl create secret tls foo-secret --key /tmp/tls.key --cert /tmp/tls.crt
```

saschagrunert:
Alright, please have a look again. 👍

docs/user-guide/kubernetes.md
```diff
@@ -105,73 +114,156 @@ spec:
...
 ```shell
-kubectl apply -f https://raw.githubusercontent.com/containous/traefik/master/examples/k8s/traefik.yaml
+$ kubectl apply -f https://raw.githubusercontent.com/containous/traefik/master/examples/k8s/traefik-ds.yaml
 ```
 
-### Check the deployment
+There are some significant differences between using Deployments and DaemonSets. The Deployment has easier
+up and down scaling possibilities. It can implement full pod lifecycle and supports rolling updates from
+Kubernetes 1.2. At least one Pod is needed to run the Deployment. The DaemonSet automatically scales to all nodes that
+meets a specific selector and guarantees to fill nodes one at a time. Rolling updates are supported from Kubernetes 1.6
```
timoreimann: 1.7 really brought DaemonSet rolling upgrades on par with Deployments. I suggest we slightly rephrase to:

>Rolling updates are **fully** supported from Kubernetes **1.7** [...]

(**Emphasis** by me to highlight the suggested delta, please **don't** include those. :-) )

saschagrunert:
Like this? :)

docs/user-guide/kubernetes.md
errm: I think we are probably talking at cross purposes here, I am thinking about the ability to specify a cert to use in the ingress annotation `ingress.kubernetes.io/auth-tls-secret`.  We don't currently support that ...

errm:
LGTM...

Just needs commits squashing down ...

ldez:
Bender say:
			The first fully automated merge is a success!
			:champagne: :tada:

PR #1740, opened 2017-06-13
Fix fallback to other nodes for Marathon

### LINKED ISSUE #659: Marathon provider doesn't seem to fall back to other addresses
I've tried to provide multiple Marathon addresses for Traefik, something like:

``` toml
[marathon]
endpoint = "http://controller01:8080,controller02:8080,controller03:8080"
watch = true
...
```

Traefik's logs confirms this config is being loaded:

```shell
time="2016-09-01T10:19:21+01:00" level=info msg="Starting provider *provider.Marathon {\
"Watch\":true,\"Filename\":\"\",\"Constraints\":null,\"Endpoint\":\"http://cont
roller01:8080,controller02:8080,controller03:8080\",...
```

"`controller01`" is down, however, and cannot be reached. Traefik seems to try this server first as it is the first in the list. The problem is that Traefik never seems to try any of the other (working) servers. It logs something like this:

```shell
time="2016-09-01T10:19:23+01:00" level=error msg="Failed to register for events, Get http://controller01:8080/v2/events: dial tcp 192.168.1.5:8080: getsockopt: connection refused"
time="2016-09-01T10:19:23+01:00" level=error msg="Marathon connection error Get http://controller01:8080/v2/events: dial tcp 192.168.1.5:8080: getsockopt: connection refused, retrying in 841.844912ms"
time="2016-09-01T10:19:24+01:00" level=error msg="Failed to register for events, Get http://controller01:8080/v2/events: dial tcp 192.168.1.5:8080: getsockopt: connection refused"
time="2016-09-01T10:19:24+01:00" level=error msg="Marathon connection error Get http://controller01:8080/v2/events: dial tcp 192.168.1.5:8080: getsockopt: connection refused, retrying in 2.353398941s"...
```

...over and over...

I tried digging in to both Traefik's and [`go-marathon`](https://github.com/gambol99/go-marathon)'s code but didn't have much luck. I'll try look again when I find the time.

marco-jantke:
This PR should fix the issue #659. The fix is actually in `github.com/gambol99/go-marathon` and this PR updates it and its sub-dependency `github.com/donovanhide/eventsource`.

I updated the dependency according to the approach @timoreimann suggested [here](https://github.com/containous/traefik/pull/1278#issuecomment-286696714). I tested the failover with Traefik against a real running Marathon cluster and took nodes down one after another. Also I tried to take all nodes down at once and see if it recovers and the failover/recovery works consistently.

Fixes #659

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 6535e5faaf0a87d89bb74841d60988f6d13ead827efa02d509fd1914aeb6e4d4
-updated: 2017-06-13T23:30:19.890844996+02:00
+hash: 34ceb7bd979d43efdbf721ccb9d983061c06db527148f90f1784db89f6d089f0
+updated: 2017-05-19T23:30:19.890844996+02:00
 imports:
 - name: cloud.google.com/go
   version: 2e6a95edb1071d750f6d7db777bf66cd2997af6c
@@ -172,7 +172,7 @@ imports:
   - store/etcd
   - store/zookeeper
 - name: github.com/donovanhide/eventsource
-  version: fd1de70867126402be23c306e1ce32828455d85b
+  version: 441a03aa37b3329bbb79f43de81914ea18724718
 - name: github.com/eapache/channels
   version: 47238d5aae8c0fefd518ef2bee46290909cf8263
 - name: github.com/eapache/queue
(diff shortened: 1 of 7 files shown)


ldez:
The `go-marathon` diff: https://github.com/gambol99/go-marathon/compare/15ea23e360abb8b25071e677aed344f31838e403...1b9c2582c26b632fb1fb295776d7ce40b68c36f2

PR #1741, opened 2017-06-13
rebuild webui when building docker image

marco-jantke:
I think its a good idea to always rebuild the web-ui when building a docker image via `make image`. Otherwise there is the risk that an image is built and used which has not the latest ui version. WDYT?

Side note: I recognised that the `make run-dev` command does not work as expected. The `main` package is now in the `server` directory. Do you think its still in use and should be cleaned up or is it obsolete?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,8 +82,14 @@ build-no-cache: dist
 shell: build ## start a shell inside the build env
 	$(DOCKER_RUN_TRAEFIK) /bin/bash
 
-image: binary ## build a docker traefik image
+image-dirty: binary ## build a docker traefik image
 	docker build -t $(TRAEFIK_IMAGE) .
+
+image: clear-static binary ## clean up static directory and build a docker traefik image
+	docker build -t $(TRAEFIK_IMAGE) .
+
+clear-static:
+	rm -rf static
 
 dist:
 	mkdir dist
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ before_deploy:
       docker version;
       pip install --user -r requirements.txt;
       make -j${N_MAKE_JOBS} crossbinary-parallel;
-      make image;
+      make image-dirty;
       mkdocs build --clean;
       tar cfz dist/traefik-${VERSION}.src.tar.gz --exclude-vcs --exclude dist .;
     fi


Makefile
```diff
@@ -82,9 +82,12 @@ build-no-cache: dist
 shell: build ## start a shell inside the build env
 	$(DOCKER_RUN_TRAEFIK) /bin/bash
 
-image: binary ## build a docker traefik image
+image: clear-static binary ## build a docker traefik image
 	docker build -t $(TRAEFIK_IMAGE) .
 
+clear-static:
+	rm -rf static
```
timoreimann: Just to be sure: When the static folder does not exist, will the make file continue nicely (and not produce any disturbing output)?

marco-jantke: Yes, `rm` exits in either case with 0.

timoreimann:
I think the change makes sense?

@ldez @emilevauge WDYT?

ldez:
`make image` is used by the deploy:

**.travis.yml**
```bash
before_deploy:
  - >
    if ! [ "$BEFORE_DEPLOY_RUN" ]; then
      export BEFORE_DEPLOY_RUN=1;
      sudo -E apt-get -yq update;
      sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install docker-engine=${DOCKER_VERSION}*;
      docker version;
      pip install --user -r requirements.txt;
      make -j${N_MAKE_JOBS} crossbinary-parallel;
      make image;
      mkdocs build --clean;
      tar cfz dist/traefik-${VERSION}.src.tar.gz --exclude-vcs --exclude dist .;
    fi
```

It's change make the deploy longer.

timoreimann:
@ldez can we measure how much time we'd lose? And it would only be for master builds, right?

If it turns out to be critical, another alternative would be to have `image-dirty` and `image[-clean]` targets where our build process uses the former and regular users take the latter.

marco-jantke:
@ldez how do you make sure that you don't deploy an old version of the webui?

We could also aim for a bit more "sophisticated" approach. `build-webui` can add a file, e.g. `LAST-REBUILD` to the static directory. Instead of the simple `if [ ! -d "static" ]` check in the Makefile we can additionally compare the modified time on the `webui` directory and only rebuild if there is something newer. Personally I am not really convinced of this approach though, as it could lead to potential bugs in the future, where the webui build artifact is influenced by things that live outside of the `webui` directory. WDYT?

ldez:
@marco-jantke this is not the deploy of the web-ui but the deploy of our releases.
Yes I'm sure we always deploy release with the latest web-ui 😉 

Create a dedicate make target (like `image-dirty`) is the best way for me :milky_way:

ldez:
@marco-jantke any news?

marco-jantke:
Sorry for the long idle time on this PR, was a lot travelling last week.

I introduced now a new target `image-dirty` that is used in the deployment scripts and have the `image` target still clean up to ensure that always the latest webui is built.

PS: github seems to sleep with one eye atm, accepting my push but not updating the PR. yet..

marco-jantke:
@timoreimann PTAL again.

PR #1257, opened 2017-03-08
Allow multiple rules from docker labels containers with traefik.<servicename>.* properties

benoitf:
It matches expectation of #444 
If labels contain for example:
   - `traefik.mycustomservice.port=443`
  -  `traefik.mycustomservice.frontend.rule=Path:/mycustomservice`
   - `traefik.anothercustomservice.port=8080`
  -  `traefik.anothercustomservice.frontend.rule=Path:/anotherservice`

all traffic to frontend on path `/mycustomservice` is redirected to the port `443` of the container defining the label while using path `/anotherservice` will redirect to the port `8080` of the docker container defining the label

Available properties are 
`traefik.<service-name>.port`
`traefik.<service-name>.weight`
`traefik.<service-name>.protocol`
`traefik.<service-name>.frontend.backend`
`traefik.<service-name>.frontend.entryPoints`
`traefik.<service-name>.frontend.rule`
`traefik.<service-name>.frontend.passHostHeader`
`traefik.<service-name>.frontend.priority`
`traefik.<service-name>.frontend.rule`

- More documentation in the docs/toml.md file
- new tests added to docker_tests.go

Change-Id: Ifaa3bb00ef0a0f38aa189e0ca1586fde8c5ed862
Signed-off-by: Florent BENOIT <fbenoit@codenvy.com>

PR #1742, opened 2017-06-13
Support multi-port service routing for containers running on Marathon

aantono:
Implementation of #1621

### DIFF
--- a/integration/marathon_test.go
+++ b/integration/marathon_test.go
@@ -75,6 +75,13 @@ func (s *MarathonSuite) extendDockerHostsFile(host, ipAddr string) error {
 	return nil
 }
 
+func deployApplication(c *check.C, client marathon.Marathon, application *marathon.Application) {
+	deploy, err := client.UpdateApplication(application, false)
+	c.Assert(err, checker.IsNil)
+	// Wait for deployment to complete.
+	c.Assert(client.WaitOnDeployment(deploy.DeploymentID, 1*time.Minute), checker.IsNil)
+}
+
 func (s *MarathonSuite) TestConfigurationUpdate(c *check.C) {
 	// Start Traefik.
 	file := s.adaptFile(c, "fixtures/marathon/simple.toml", struct {
@@ -117,13 +124,28 @@ func (s *MarathonSuite) TestConfigurationUpdate(c *check.C) {
 		Container("emilevauge/whoami")
 
 	// Deploy the test application.
-	deploy, err := client.UpdateApplication(app, false)
-	c.Assert(err, checker.IsNil)
-	// Wait for deployment to complete.
-	c.Assert(client.WaitOnDeployment(deploy.DeploymentID, 1*time.Minute), checker.IsNil)
+	deployApplication(c, client, app)
 
 	// Query application via Traefik.
 	err = try.GetRequest("http://127.0.0.1:8000/service", 30*time.Second, try.StatusCodeIs(http.StatusOK))
 	c.Assert(err, checker.IsNil)
+
+	// Create test application with services to be deployed.
+	app = marathon.NewDockerApplication().
+		Name("/whoami").
+		CPU(0.1).
+		Memory(32).
+		AddLabel(types.ServiceLabel(types.LabelFrontendRule, "app"), "PathPrefix:/app")
+	app.Container.Docker.Bridged().
+		Expose(80).
+		Container("emilevauge/whoami")
+
+	// Deploy the test application.
+	deployApplication(c, client, app)
+
+	// Query application via Traefik.
+	err = try.GetRequest("http://127.0.0.1:8000/app", 30*time.Second, try.StatusCodeIs(http.StatusOK))
+	c.Assert(err, checker.IsNil)
+
 	showTraefikLog = false
 }
--- a/provider/marathon/builder_test.go
+++ b/provider/marathon/builder_test.go
@@ -45,13 +45,14 @@ func label(key, value string) func(*marathon.Application) {
 }
 
 func labelWithService(key, value string, serviceName string) func(*marathon.Application) {
-	if len(serviceName) > 0 {
-		property := strings.TrimPrefix(key, types.LabelPrefix)
-		return func(app *marathon.Application) {
-			app.AddLabel(types.LabelPrefix+serviceName+"."+property, value)
-		}
+	if len(serviceName) == 0 {
+		panic("serviceName can not be empty")
 	}
-	return label(key, value)
+
+	property := strings.TrimPrefix(key, types.LabelPrefix)
+	return func(app *marathon.Application) {
+		app.AddLabel(types.LabelPrefix+serviceName+"."+property, value)
+	}
 }
 
 func healthChecks(checks ...*marathon.HealthCheck) func(*marathon.Application) {
--- a/provider/marathon/builder_test.go
+++ b/provider/marathon/builder_test.go
@@ -1,8 +1,10 @@
 package marathon
 
 import (
+	"strings"
 	"time"
 
+	"github.com/containous/traefik/types"
 	"github.com/gambol99/go-marathon"
 )
 
@@ -40,6 +42,16 @@ func label(key, value string) func(*marathon.Application) {
 	return func(app *marathon.Application) {
 		app.AddLabel(key, value)
 	}
+}
+
+func labelWithService(key, value string, serviceName string) func(*marathon.Application) {
+	if len(serviceName) > 0 {
+		property := strings.TrimPrefix(key, types.LabelPrefix)
+		return func(app *marathon.Application) {
+			app.AddLabel(types.LabelPrefix+serviceName+"."+property, value)
+		}
+	}
+	return label(key, value)
 }
 
 func healthChecks(checks ...*marathon.HealthCheck) func(*marathon.Application) {
--- a/provider/marathon/marathon.go
+++ b/provider/marathon/marathon.go
@@ -459,10 +459,12 @@ func extractServicesLabels(labels *map[string]string) labelServiceProperties {
 				continue
 			}
 
-			// Accordig to the regex, match index 1 is "service_name" and match index 2 is the "property_name"
+			// According to the regex, match index 1 is "service_name" and match index 2 is the "property_name"
 			serviceName := matches[1]
 			propertyName := matches[2]
-			v[serviceName] = make(servicePropertyValues)
+			if _, ok := v[serviceName]; !ok {
+				v[serviceName] = make(servicePropertyValues)
+			}
 			v[serviceName][propertyName] = value
 		}
 	}
--- a/provider/marathon/marathon.go
+++ b/provider/marathon/marathon.go
@@ -401,7 +401,7 @@ func (p *Provider) mustGetPort(task marathon.Task, application marathon.Applicat
 	if value, ok := p.getLabel(application, types.LabelPort, serviceName); ok {
 		port, err := processPorts(application, task, value, "")
 		if err != nil {
-			return "", fmt.Errorf("Unable to process port value %s for Marathon application %s and task %s: %s", value, application.ID, task.ID, err)
+			return "", fmt.Errorf("unable to process port value %s for Marathon application %s and task %s: %s", value, application.ID, task.ID, err)
 		}
 
 		return strconv.Itoa(port), nil
@@ -409,7 +409,7 @@ func (p *Provider) mustGetPort(task marathon.Task, application marathon.Applicat
 	if value, ok := p.getLabel(application, types.LabelPortIndex, serviceName); ok {
 		port, err := processPorts(application, task, "", value)
 		if err != nil {
-			return "", fmt.Errorf("Unable to process port index %s for Marathon application %s and task %s: %s", value, application.ID, task.ID, err)
+			return "", fmt.Errorf("unable to process port index %s for Marathon application %s and task %s: %s", value, application.ID, task.ID, err)
 		}
 
 		return strconv.Itoa(port), nil
@@ -417,7 +417,7 @@ func (p *Provider) mustGetPort(task marathon.Task, application marathon.Applicat
 
 	port, err := processPorts(application, task, "", "0")
 	if err != nil {
-		return "", fmt.Errorf("Unable to process ports for Marathon application %s and task %s: %s", application.ID, task.ID, err)
+		return "", fmt.Errorf("unable to process ports for Marathon application %s and task %s: %s", application.ID, task.ID, err)
 	}
 
 	return strconv.Itoa(port), nil
@@ -593,7 +593,7 @@ func (p *Provider) getBasicAuth(application marathon.Application, serviceName st
 //processPorts validates `portValue` to be valid numeric port.
 // if it is empty, proceeds to extract all ports and attempts to retrieve a port specified by the 'portIndexValue', given that it is a valid integer.
 // if all fails, returns first port (at index 0) from the list.
-func processPorts(application marathon.Application, task marathon.Task, portValue string, portIndexValue string) (int, error) {
+func processPorts(application marathon.Application, task marathon.Task, portValue, portIndexValue string) (int, error) {
 	if len(portValue) > 0 {
 		port, err := strconv.Atoi(portValue)
 		switch {
--- a/provider/marathon/marathon.go
+++ b/provider/marathon/marathon.go
@@ -211,7 +211,11 @@ func (p *Provider) loadMarathonConfig() *types.Configuration {
 	filteredApps := fun.Filter(p.applicationFilter, applications.Apps).([]marathon.Application)
 	for i, app := range filteredApps {
 		filteredApps[i].Tasks = fun.Filter(func(task *marathon.Task) bool {
-			return p.taskFilter(*task, app)
+			filtered := p.taskFilter(*task, app)
+			if filtered {
+				p.logIllegalServices(*task, app)
+			}
+			return filtered
 		}, app.Tasks).([]*marathon.Task)
 	}
 
@@ -260,21 +264,6 @@ func (p *Provider) taskFilter(task marathon.Task, application marathon.Applicati
 		return false
 	}
 
-	for _, serviceName := range p.getServiceNames(application) {
-		if _, err := p.processPorts(application, task, serviceName); err != nil {
-			log.Errorf("Filtering Marathon task %s from application %s without port", task.ID, application.ID)
-			return false
-		}
-	}
-
-	// Filter illegal port label specification.
-	_, hasPortIndexLabel := p.getAppLabel(application, types.LabelPortIndex)
-	_, hasPortLabel := p.getAppLabel(application, types.LabelPort)
-	if hasPortIndexLabel && hasPortLabel {
-		log.Debugf("Filtering Marathon task %s from application %s specifying both traefik.portIndex and traefik.port labels", task.ID, application.ID)
-		return false
-	}
-
 	// Filter task with existing, bad health check results.
 	if application.HasHealthChecks() {
 		if task.HasHealthCheckResults() {
@@ -297,6 +286,26 @@ func (p *Provider) taskFilter(task marathon.Task, application marathon.Applicati
 
 func isApplicationEnabled(application marathon.Application, exposedByDefault bool) bool {
 	return exposedByDefault && (*application.Labels)[types.LabelEnable] != "false" || (*application.Labels)[types.LabelEnable] == "true"
+}
+
+// logIllegalServices logs illegal service configurations.
+// While we cannot filter on the service level, they will eventually get
+// rejected once the server configuration is rendered.
+func (p *Provider) logIllegalServices(task marathon.Task, application marathon.Application) {
+	for _, serviceName := range p.getServiceNames(application) {
+		// Check for illegal/missing ports.
+		if _, err := p.processPorts(application, task, serviceName); err != nil {
+			log.Warnf("%s has an illegal configuration: no proper port available", identifier(application, task, serviceName))
+			continue
+		}
+
+		// Check for illegal port label combinations.
+		_, hasPortLabel := p.getLabel(application, types.LabelPort, serviceName)
+		_, hasPortIndexLabel := p.getLabel(application, types.LabelPortIndex, serviceName)
+		if hasPortLabel && hasPortIndexLabel {
+			log.Warnf("%s has both port and port index specified; port will take precedence", identifier(application, task, serviceName))
+		}
+	}
 }
 
 //servicePropertyValues is a map of services properties
@@ -391,7 +400,7 @@ func (p *Provider) getLabel(application marathon.Application, label string, serv
 func (p *Provider) getPort(task marathon.Task, application marathon.Application, serviceName string) string {
 	port, err := p.processPorts(application, task, serviceName)
 	if err != nil {
-		log.Errorf("Unable to process ports for Marathon application %s and task %s: %s", application.ID, task.ID, err)
+		log.Errorf("Unable to process ports for %s: %s", identifier(application, task, serviceName), err)
 		return ""
 	}
 
@@ -662,4 +671,12 @@ func parseIndex(index string, length int) (int, error) {
 	}
 
 	return parsed, nil
+}
+
+func identifier(app marathon.Application, task marathon.Task, serviceName string) string {
+	id := fmt.Sprintf("Marathon task %s from application %s", task.ID, app.ID)
+	if serviceName != "" {
+		id += fmt.Sprintf(" (service: %s)", serviceName)
+	}
+	return id
 }
--- a/provider/marathon/marathon.go
+++ b/provider/marathon/marathon.go
@@ -47,8 +47,8 @@ const (
 var _ provider.Provider = (*Provider)(nil)
 
 // Regexp used to extract the name of the service and the name of the property for this service
-// All properties are under the format traefik.<servicename>.frontend.*= except the port/portIndex/weight/protocol directly after traefik.<servicename>.
-var servicesPropertiesRegexp = regexp.MustCompile(`^traefik\.(?P<service_name>.+?)\.(?P<property_name>port|portIndex|weight|protocol|frontend\.(.*))$`)
+// All properties are under the format traefik.<servicename>.frontend.*= except the port/portIndex/weight/protocol/backend directly after traefik.<servicename>.
+var servicesPropertiesRegexp = regexp.MustCompile(`^traefik\.(?P<service_name>.+?)\.(?P<property_name>port|portIndex|weight|protocol|backend|auth.basic|frontend\.(.*))$`)
 
 // Provider holds configuration of the provider.
 type Provider struct {
@@ -260,9 +260,16 @@ func (p *Provider) taskFilter(task marathon.Task, application marathon.Applicati
 		return false
 	}
 
-	if _, err := processPorts(application, task, 0); err != nil {
-		log.Errorf("Filtering Marathon task %s from application %s without port: %s", task.ID, application.ID, err)
-		return false
+	for _, serviceName := range p.getServiceNames(application) {
+		foundPort := false
+		if port := p.getPort(task, application, serviceName); len(port) > 0 {
+			foundPort = true
+		}
+
+		if !foundPort {
+			log.Errorf("Filtering Marathon task %s from application %s without port", task.ID, application.ID)
+			return false
+		}
 	}
 
 	// Filter illegal port label specification.
@@ -298,18 +305,18 @@ func isApplicationEnabled(application marathon.Application, exposedByDefault boo
 }
 
 //servicePropertyValues is a map of services properties
-//example values are: weight=42
+//an example value is: weight=42
 type servicePropertyValues map[string]string
 
-//serviceProperties is a map of service properties per service, which we can get with label[serviceName][propertyName] and we got the propertyValue
+//serviceProperties is a map of service properties per service, which we can get with label[serviceName][propertyName]. It yields a property value.
 type serviceProperties map[string]servicePropertyValues
 
-// Check if for the given application, we find labels that are defining services
+//hasServices checks if there are service-defining labels for the given application
 func (p *Provider) hasServices(application marathon.Application) bool {
 	return len(extractServiceProperties(application.Labels)) > 0
 }
 
-// Extract the service labels from application labels of Application struct
+//extractServiceProperties extracts the service labels for the given application
 func extractServiceProperties(labels *map[string]string) serviceProperties {
 	v := make(serviceProperties)
 
@@ -333,14 +340,14 @@ func extractServiceProperties(labels *map[string]string) serviceProperties {
 	return v
 }
 
-//getServiceProperty returns the entry for a service label searching in all labels of the given application
+//getServiceProperty returns the property for a service label searching in all labels of the given application
 func getServiceProperty(application marathon.Application, serviceName string, property string) (string, bool) {
 	value, ok := extractServiceProperties(application.Labels)[serviceName][property]
 	return value, ok
 }
 
-//getServiceNames returns array of service names for a given application
-//An empty name "" will be added if no service specific properties exist, as an indication that there are no sub-services, but only main Application
+//getServiceNames returns a list of service names for a given application
+//An empty name "" will be added if no service specific properties exist, as an indication that there are no sub-services, but only main application
 func (p *Provider) getServiceNames(application marathon.Application) []string {
 	labelServiceProperties := extractServiceProperties(application.Labels)
 	var names []string
@@ -356,15 +363,21 @@ func (p *Provider) getServiceNames(application marathon.Application) []string {
 
 func (p *Provider) getServiceNameSuffix(serviceName string) string {
 	if len(serviceName) > 0 {
-		serviceName = provider.Replace("/", "-", serviceName)
-		serviceName = provider.Replace(".", "-", serviceName)
+		serviceName = strings.Replace(serviceName, "/", "-", -1)
+		serviceName = strings.Replace(serviceName, ".", "-", -1)
 		return "-service-" + serviceName
 	}
 	return ""
 }
 
+//getAppLabel is a convenience function to get application label, when no serviceName is available
+//it is identical in calling getLabel(application, label, "")
+func (p *Provider) getAppLabel(application marathon.Application, label string) (string, bool) {
+	return p.getLabel(application, label, "")
+}
+
 //getLabel returns a string value of a corresponding `label` argument
-//  if 'serviceName' value is present, the 'label' value gets stripped of
+//  If serviceName is non-empty, we look for a service label. If none exists or serviceName is empty, we look for an application label.
 func (p *Provider) getLabel(application marathon.Application, label string, serviceName string) (string, bool) {
 	if len(serviceName) > 0 {
 		property := strings.TrimPrefix(label, types.LabelPrefix)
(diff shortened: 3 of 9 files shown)


aantono:
Rebased against master, should be good to go.  Any other corrections that are required?

aantono:
Hello @timoreimann, thanks for the detailed feedback.  The current implementation was modeled (with some code copying) from the [existing Docker service labels support](https://github.com/containous/traefik/pull/1257).  I was trying to keep the same design, etc.  Let me know if you would like to deviate from how it is currently done in the Docker provider, would be happy to do so. (especially if we are not required to keep the template "API" backwards-compatible).

provider/marathon/marathon.go
```diff
@@ -420,6 +447,187 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
...
+			v[serviceName][result["property_name"]] = serviceProperty
+		}
+	}
+
+	return v
+}
+
+// Gets the entry for a service label searching in all labels of the given application
+func getApplicationServiceLabel(application marathon.Application, serviceName string, entry string) (string, bool) {
+	value, ok := extractServicesLabels(application.Labels)[serviceName][entry]
```
aantono: It would not panic (just tested the usecase on play.golang.org)  The `ok` would equal `false` and `value` would have default `string` value of `""`

We do check for `ok` in all the places we call `getApplicationServiceLabel`, like `getServicePriority` for example, so that covers the use-case (IMHO)

provider/marathon/marathon.go
```diff
@@ -420,6 +447,187 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
...
+			}
+			serviceName := result["service_name"]
+			if _, ok := v[serviceName]; !ok {
+				v[serviceName] = make(map[string]string)
+			}
+			v[serviceName][result["property_name"]] = serviceProperty
+		}
+	}
+
+	return v
```
aantono: The parser route is definitely possible, though I personally prefer the regex as it will be more performant and less code to maintain.  After the parser logic is all coded up, it will effectively do the same thing as Regex, scan the label text, tokenize it by dot-delimited groups, build the same `result` map with the rest of the logic to follow.

The regex logic is actually not very complex (IMHO).
`^traefik\.(?P<service_name>.*?)\.(?P<property_name>port|portIndex|weight|protocol|frontend\.(.*))$` breaks down as following:
* The label MUST start with `traefik.` (use of `^` indicates start of line)
* The next section, defined by `(?P<service_name>.*?)` group is a named & numbered capturing group (submatch) that will basically extract a piece of text that sits in between `.` separators after `traefik.` segment.
* The next section, defined by `(?P<property_name>port|portIndex|weight|protocol|frontend\.(.*))` group is another named & numbered submatch, but conditional on `property_name` text segment be only one of the listed options: `port`, `portIndex`, `weight`, `protocol` or `frontend.*`
* The rest of the label till end of line/entry, defined by `$`

The reason this regex looks complex is due to using named groups and having to use `\` in many places to escape the `.` :)

The issues with the parser are: 
* instead of 3 lines of code used in regex (regex definition, `servicesPropertiesRegexp.FindStringSubmatch(index)` and `servicesPropertiesRegexp.SubexpNames()`) we would have much more code for the parser logic 
* parser logic would be custom-written, needs to be extensively tested and add complexity that has to be maintained long term
* regex performance is already very good (Google guys have done their part. :) ), the parser might need optimizations to tune it to make sure performance is not suffering.

Thoughts?

provider/marathon/marathon.go
```diff
@@ -420,6 +447,187 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
 	return ok
 }
 
+// Map of services properties
+// we can get it with label[serviceName][propertyName] and we got the propertyValue
+type labelServiceProperties map[string]map[string]string
```
aantono: What name would you like to use instead?  (`labelServiceProperties` is already defined as a `type`, so I'm assuming you don't like the name)

provider/marathon/marathon.go
```diff
@@ -32,6 +33,10 @@ const (
 
 var _ provider.Provider = (*Provider)(nil)
 
+// Regexp used to extract the name of the service and the name of the property for this service
+// All properties are under the format traefik.<servicename>.frontent.*= except the port/portIndex/weight/protocol directly after traefik.<servicename>.
+var servicesPropertiesRegexp = regexp.MustCompile(`^traefik\.(?P<service_name>.*?)\.(?P<property_name>port|portIndex|weight|protocol|frontend\.(.*))$`)
```
aantono: Yep, good point.  TBH, I just copied the regex from https://github.com/containous/traefik/pull/1257/files#diff-0e8a9f9593585fb3fa7ef0396fd5bbecR319, so should probably change it there as well, right?

provider/marathon/marathon.go
timoreimann: 👍

provider/marathon/marathon.go
timoreimann: I agree we should. Ideally, there should be a new test (case) too.

aantono: Updated with added test label to ensure that `test..port` does not result with a service label named `` (empty string)

provider/marathon/marathon.go
timoreimann: Thanks for the explanation. My main concern is about how the regular expression is applied in the code, and how the regexp package API is used to extract the kind of information we are looking for. To be honest, I think the API isn't the most intuitive, which may stem from the fact that regular expressions aren't the most trivial topic.

Let me share my thoughts on your issues with the parser approach:

1. **code size.** Yes, we'd have somewhat more code. IMO though, I think it'd be easier to understand. When I tried to understand your regex, I basically had to jump back and forth between the regex and the algorithm in `extractServicesLabels` to convince myself that the code is really covering all cases. Personally, I tend to prefer a bit less dense code at the benefit of better comprehensibility.
1. **parser logic/testing.** For sure we'd need a custom parser. I disagree on the testing argument though: The regex and the parser approaches should have roughly the same test cases. In fact, I think the more explicit nature of the parser approach make it much clearer to see which cases need to be tested. On the contrary, it's harder to figure out the circumstances under which a regular expression matches or not (which is what took me so long to understand).
3. **performance.** Regexes require full-blown state machines whereas parsing is mostly a matter of simple string comparisons and manipulations. So I have my doubts whether the regex will outperform the parser. More importantly though, I wouldn't worry about this matter at all: It probably takes a lot of regex'ing/parsing for any performance impact to be observable, and even if we think it was, we'd need benchmarks to prove it. For now, let's focus on a solution that's easiest to read, comprehend, and test.

I'll add another point in favor of parsers: they are often simpler to extend. When the structure of what your regular expression has to match changes, chances are you'll have to rebuild the expression or start introducing hard-to-understand features like look-ahead/look-behind. That's usually where things get ugly.

FWIW, the stdlib is full of parsers, so at least the Go team thinks it's a good and fast-enough thing. :)

One implementation speaks louder than 100 discussions, so I went ahead and [implemented the parser](https://github.com/timoreimann/traefik/commit/bdab1458dd0a25ace8b8775d964fca304897407c) for comparison reasons. I admit it's a bit larger (especially because your implementation can be further simplified now that we say the service name is mandatory -- I'll address that in a separate comment). OTOH, it's super clear to see what to expect from the label pattern when being parsed, which makes writing down all test cases a rather mechanical matter.

Overall, I'm leaning towards the parser for clarity reasons, but I can see the advantage of a shorter implementation when using a regular expression. I guess I can convince myself that staying with what we have right now in the PR is acceptable, though I wouldn't want to leave this comment thread without a rather complete discussion. :-)

I'll also try to get some additional feedback from the other maintainers on this point once we've covered the remainders of the PR.

timoreimann:
>Let me know if you would like to deviate from how it is currently done in the Docker provider, would be happy to do so. (especially if we are not required to keep the template "API" backwards-compatible).

I'm currently checking back with the other maintainers as to what our understanding is towards any API guarantees. My personal opinion is that compatibility should only be maintained on the higher template rendering level (meaning that the same template input parameters should ideally yield the same output configuration across different Traefik versions) but should not apply to the specific template functions.

Once we have clarity on this point, we can decide how to proceed. I'd feel much more comfortable if we had the desirable freedom so that we don't have to add too much complexity to the Marathon provider.

Will get back to you on this one ASAP!

provider/marathon/marathon.go
```diff
@@ -420,6 +444,193 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
...
+	v := make(labelServiceProperties)
+
+	if labels != nil {
+		for label, value := range *labels {
+			matches := servicesPropertiesRegexp.FindStringSubmatch(label)
+			if matches == nil {
+				continue
+			}
+
+			result := make(map[string]string)
```
timoreimann: Here's the promised comment I announced in another comment:

I don't think we need the `result` map. Our regular expression pattern basically guarantees us that we'll have `service_name` and `property_name` sub-matches. So we can just extract the names/values at the known index positions and use them to populate the `labelServiceProperties` map.

provider/marathon/marathon.go
```diff
@@ -420,6 +444,193 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
...
+			for i, name := range servicesPropertiesRegexp.SubexpNames() {
+				if i != 0 {
+					result[name] = matches[i]
+				}
+			}
+			serviceName := result["service_name"]
+			if _, ok := v[serviceName]; !ok {
+				v[serviceName] = make(map[string]string)
```
timoreimann: `make` can now use `serviceProperties`.

provider/marathon/marathon.go
```diff
@@ -420,6 +444,193 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
 	return ok
 }
 
+// Map of services properties
+// we can get it with label[serviceName][propertyName] and we got the propertyValue
+type serviceProperties map[string]string
```
timoreimann: Since this is a map from property (names) to values, how about calling it `servicePropertyValues`?

provider/marathon/marathon.go
aantono: Those are fair points.  Do you think we can go ahead with this PR as is, and have a separate PR for creating a "generic" parser to handle the label service tokenization and extraction?  I can see how that (Parser) can be a generic/common facility used by pretty much any provider that has a "service" concept. (like various Docker derivatives [docker, kubernetes, mesos, marathon, etc.])

provider/marathon/marathon.go
```diff
@@ -420,6 +444,187 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
...
+	v := make(labelServiceProperties)
+
+	if labels != nil {
+		for label, value := range *labels {
+			matches := servicesPropertiesRegexp.FindStringSubmatch(label)
+			if matches == nil {
+				continue
+			}
+
+			// Accordig to the regex, match index 1 is "service_name" and match index 2 is the "property_name"
```
timoreimann: Typo: Accordig -> According

provider/marathon/marathon.go
```diff
@@ -420,6 +444,187 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
...
+			serviceName := matches[1]
+			propertyName := matches[2]
+			v[serviceName] = make(servicePropertyValues)
```
timoreimann: You dropped the check if the `servicePropertyValues` map exists. Aren't you now overwriting any pre-existing value with a brand new map on every loop iteration?

provider/marathon/marathon.go
timoreimann: Good idea. Okay with me. 👌

timoreimann:
@aantono could you please only push commits (as opposed to rebasing) during the review? Otherwise, it's very hard to see which changes you added. Thanks!

timoreimann:
Tests are failing. :-)

timoreimann:
@aantono looks like we're okay to consider covering the extended service-based functionality by the existing template functions, even if it means adjusting the signatures.

I haven't deeply explored the idea but think it should be feasible. Would you mind taking a stab and see how far we can get with this? Thanks!

provider/marathon/marathon.go
aantono: Good catch!

aantono:
@timoreimann, sounds good. I will try to model everything in terms of services and assign a `_default_` service when no explicit services exist.  Am I correct to understand that we are OK with removing the non-service-based methods from the Template?

timoreimann:
@aantono:

>Am I correct to understand that we are OK with removing the non-service-based methods from the Template?

Yes, I think we're free to go wild. :-)

provider/marathon/marathon.go
```diff
@@ -369,6 +449,62 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
...
 }
 
+// Map of services properties
+// we can get it with label[serviceName][propertyName] and we got the propertyValue
+type servicePropertyValues map[string]string
+
+type labelServiceProperties map[string]servicePropertyValues
+
+// Check if for the given application, we find labels that are defining services
+func (p *Provider) hasServices(application marathon.Application) bool {
```
aantono: This method is exported to the template to use that as a check if application has service labels defined for it.

provider/marathon/marathon.go
```diff
@@ -537,3 +676,12 @@ func parseIndex(index string, length int) (int, error) {
 
 	return parsed, nil
 }
+
+func extractServiceName(varg ...string) string {
+	switch {
+	case len(varg) == 1:
+		return varg[0]
+	default:
```
aantono: `extractServiceName` method is a helper to help retrieve the optional service name from the varargs.  I was trying to keep the existing template method signatures backwards compatible, and since Go does not support method overloading, the only option was to declare last argument of those methods as a vararg string.  So by convention, if there is only 1 element in the vararg, it represents a ServiceName, otherwise it's not a service name we can use, so empty string for service name is returned.

middlewares/datadog.go
```diff
@@ -11,7 +11,8 @@ import (
 	"github.com/go-kit/kit/metrics/dogstatsd"
 )
 
-var _ Metrics = (Metrics)(nil)
+var _ Metrics = (*Datadog)(nil)
+var _ RetryMetrics = (*Datadog)(nil)
```
aantono: Opps... must've been a glitch while rebasing between various branches...   should be part of #1884

provider/marathon/marathon.go
timoreimann: The current template doesn't use it though. Unless we see a concrete use case for the future, I'd rather drop the method for now.

aantono: it is not used in the default template, but definitely has use-cases in custom templates that would exist out in the wild.  If you insist, I will definitely remove it though.

provider/marathon/marathon.go
timoreimann: I think we've established before that template API stability is not a concern. As such, we can extend the existing signatures to serve our long-term maintainability needs.

provider/marathon/marathon.go
timoreimann: Do you have an example use case for me so that I can get a better picture?

provider/marathon/marathon.go
aantono: You are correct, we did have that established, I just thought it would be nice to still preserve the method signatures to save people who have custom templates the need to change them.  But can remove it if you think that would be better.

provider/marathon/marathon.go
aantono: One usecase that we have is a different front-ends (completely) depending on if the container has multiple services exposed or not.  For multi-service containers, we need to add some extra rules for adding header values, etc...

provider/marathon/marathon.go
timoreimann: I wouldn't mind preserving the method signatures if it came practically free of charge. In this case though, I believe the workaround to achieve compatibility isn't very light-weight, so I'd rather go with a breaking change.

Our template guarantees might change in the future; for now though, I think we should take advantage of the flexibility we have if it helps us.

aantono: sounds good.

provider/marathon/marathon.go
timoreimann: Hmm, I see what you mean, but it still seems rather vague to me. I vote for removing the method and ask whoever might need it in the future to file a new issue / PR.

provider/provider.go
```diff
@@ -164,24 +167,26 @@ func (clientTLS *ClientTLS) CreateTLSConfig() (*tls.Config, error) {
 	cert := tls.Certificate{}
 	_, errKeyIsFile := os.Stat(clientTLS.Key)
 
-	if _, errCertIsFile := os.Stat(clientTLS.Cert); errCertIsFile == nil {
-		if errKeyIsFile == nil {
-			cert, err = tls.LoadX509KeyPair(clientTLS.Cert, clientTLS.Key)
-			if err != nil {
-				return nil, fmt.Errorf("Failed to load TLS keypair: %v", err)
+	if len(clientTLS.Cert) > 0 && len(clientTLS.Key) > 0 {
```
aantono: This is to wrap the certificate loading only if it is indeed configured.  Otherwise things fail when you only use `clientTLS.InsecureSkipVerify=true`.  This was causing issues when trying to connect to Mesos master that has a self-signed cert.   Do you want to move that to a separate PR?

timoreimann: Yes, I'd appreciate a separate PR. Thanks!

provider/marathon/marathon.go
aantono: Since we are actually using it internally in the build from this PR, should I have another PR created just asking for this method, or can we keep it here?  It would be similar to #1693 `getSubDomain` one...

timoreimann: Apologies -- I missed you saying that you already have this very use case. I thought we were still speaking abstractly here.

In that case, by all means keep the function in this PR. My concern was it might take months or years for anyone to come by and request this feature while we had to maintain it. Since you're already around using, let's keep it. :-)

(Sorry for the misunderstanding.)

provider/marathon/marathon.go
```diff
@@ -284,7 +312,13 @@ func (p *Provider) getPort(task marathon.Task, application marathon.Application)
...
 
-func (p *Provider) getWeight(application marathon.Application) string {
+// Extract weight from labels for a given service and a given marathon application
+func (p *Provider) getWeight(application marathon.Application, varg ...string) string {
+	serviceName := extractServiceName(varg...)
+	if value, ok := getApplicationServiceLabel(application, serviceName, "weight"); ok {
+		return value
+	}
+
 	if label, ok := p.getLabel(application, types.LabelWeight); ok {
```
aantono: since we already have marathon-specific `func (p *Provider) getLabel(application marathon.Application, label string) (string, bool) {` defined, are you ok with extending it's signature to take `serviceName string` as last argument.  So we would check if `serviceName` is not empty, then try to load the service label, otherwise go for application one directly.

provider/marathon/marathon.go
aantono: No worries, all good... You have a lot of code to look over... ;)

provider/marathon/marathon.go
```diff
@@ -369,6 +449,62 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
 	return ok
 }
 
+// Map of services properties
+// we can get it with label[serviceName][propertyName] and we got the propertyValue
```
aantono: 👍

provider/marathon/marathon.go
timoreimann: Ah true, we already have a custom implementation. Sure, please go ahead and extend that one.

provider/marathon/marathon.go
```diff
@@ -369,6 +449,62 @@ func (p *Provider) hasCircuitBreakerLabels(application marathon.Application) boo
...
+
+	for k := range labelServiceProperties {
+		names = append(names, k)
+	}
+	// Adding an empty name, as an indication that there are no sub-services, but only main Application
+	if len(names) == 0 {
+		names = append(names, "")
+	}
+	return names
+}
```
aantono: 👍

aantono:
Created a separate PR #1932 for the TLS InsecureSkipVerify issue.

emilevauge:
@timoreimann any feedback on this one ?

provider/marathon/marathon.go
```diff
@@ -274,8 +357,24 @@ func (p *Provider) getLabel(application marathon.Application, label string) (str
 	return "", false
 }
 
-func (p *Provider) getPort(task marathon.Task, application marathon.Application) string {
-	port, err := processPorts(application, task)
+// Extract port from labels for a given service and a given marathon application
+func (p *Provider) getPort(task marathon.Task, application marathon.Application, serviceName string) string {
+	if value, ok := getServiceProperty(application, serviceName, "port"); ok {
```
timoreimann: The extended implementation of this method is fairly hard to reason about. I also think it's not on par with what we had before: For instance, we do not do the same level of sanity-checking on a found service port as we do on an application port.

Can we refactor both `getPort` and `processPorts` in the following regards:

- Have `getPort` look for a service port. If one exists, have it look for a service port index and pass both to `processPorts`. If no service port is found, have it look for an application port and port index and pass those to `processPorts`.
- Have `processPorts` parse the given port and port index (i.e., remove the label extraction part from it so that it becomes reusable).

provider/marathon/marathon.go
```diff
@@ -284,74 +383,86 @@ func (p *Provider) getPort(task marathon.Task, application marathon.Application)
 	return strconv.Itoa(port)
 }
 
-func (p *Provider) getWeight(application marathon.Application) string {
-	if label, ok := p.getLabel(application, types.LabelWeight); ok {
+// Extract weight from labels for a given service and a given marathon application
```
timoreimann: Please drop the GoDoc added here and for the other label-based `get*` methods that didn't have a GoDoc before unless it's not immediately clear what they do.

provider/marathon/marathon_test.go
```diff
@@ -257,6 +257,124 @@ func TestMarathonLoadConfigNonAPIErrors(t *testing.T) {
 				},
 			},
 		},
+		{
+			desc: "multiple ports",
```
timoreimann: The test cases added to this test all seem to belong to `TestMarathonGetPort`.

Also, please replace the literal type definitions by the builder functions we're using by now. See other tests for how to do that.

provider/marathon/marathon_test.go
```diff
@@ -266,6 +384,7 @@ func TestMarathonLoadConfigNonAPIErrors(t *testing.T) {
 
 			c.application.ID = "/app"
 			c.task.ID = "task"
+			c.task.AppID = c.application.ID
```
timoreimann: I suppose you can undo this once you've moved the test cases into the correct test.

provider/marathon/marathon.go
aantono: are you envisioning the signature for `processPorts` be something like: `func processPorts(application marathon.Application, task marathon.Task, portValue string, portIndex int) (int, error)`, and internally have it first see if `portValue` is a valid integer, in which case return the converted integer, otherwise attempt to extract a port at a `portIndex`?  The `""` empty `portValue` and '-1' for `portIndex` would be indicators that those don't exist...?

timoreimann: Yes -- I think that sums it up pretty well.

timoreimann: I'm also open for alternative implementations that help improve readability and completeness, however.

aantono: I'll take a stab at it...  Now, that `getLabel` always falls back to do application generic label lookup if the service label not existant, that might help collapse some of the logic, and I'll only ever need to handle 2 generic cases... port and portIndex, without having to worry about services...

provider/marathon/marathon_test.go
aantono: I added them here because they are more then just the port, though the different ports is the most obvious feature for multi-port-service , so it would be more appropriate, IMHO, to put it here, in general config loading.

timoreimann: What's the "more part"? Maybe we can keep a more minimalistic test case in this test function but move the majority over to the `TestMarathonGetPort`.

aantono: There are also tests for frontend rules, etc..  But I'll add more cases specifically to `TestMarathonGetPort` one as well...

ldez:
@aantono I think you were wrong when resolving conflicts during the rebase.
Could you fix?

```
# github.com/containous/traefik/provider/marathon
/tmp/go-build072621665/github.com/containous/traefik/provider/marathon/_test/_obj_test/marathon.go:401: undefined: provider.Replace
/tmp/go-build072621665/github.com/containous/traefik/provider/marathon/_test/_obj_test/marathon.go:402: undefined: provider.Replace
FAIL	github.com/containous/traefik/provider/marathon [build failed]
```

ldez:
@aantono You're faster than me 😄

aantono:
@ldez Trying to get it all working before the freeze... 😸 

Do you know if there is anything else missing from this PR?

ldez:
Add a simple integration test which shows that basic service label parsing works.
https://github.com/containous/traefik/blob/master/integration/marathon_test.go

aantono:
Done... hopefully that should do it. ;)

integration/marathon_test.go
```diff
@@ -75,6 +75,17 @@ func (s *MarathonSuite) extendDockerHostsFile(host, ipAddr string) error {
...
 }
 
+func deployApplication(c *check.C, client marathon.Marathon, application *marathon.Application) {
+	deploy, err := client.UpdateApplication(application, false)
+	c.Assert(err, checker.IsNil)
+	// Wait for deployment to complete.
+	c.Assert(client.WaitOnDeployment(deploy.DeploymentID, 1*time.Minute), checker.IsNil)
+}
+
+func tryAppRequest(url string) error {
```
ldez: could you unfactorize?

provider/marathon/marathon_test.go
```diff
@@ -580,14 +689,37 @@ func TestMarathonGetPort(t *testing.T) {
...
 			t.Parallel()
-			actual := provider.getPort(c.task, c.application)
-			if actual != c.expected {
+			if actual, ok := assertWithServices(provider.getServiceNames(c.application), c.expected,
+				func(serviceName string) string {
+					if value := provider.getPort(c.task, c.application, serviceName); len(value) > 0 {
+						return value
+					}
+					return ""
+				}); !ok {
```
timoreimann: The usage of `assertWithServices` looks way too convoluted and smart to me. Can't we just put the expected values into the test tables?

provider/marathon/marathon_test.go
```diff
@@ -30,6 +30,14 @@ func newFakeClient(applicationsError bool, applications marathon.Applications) *
 	return fakeClient
 }
 
+func assertWithServices(serviceNames []string, expected string, getActual func(string) string) (string, bool) {
```
timoreimann: I find the function name somewhat misleading because it is not doing any actual assertions. We should either pass a `*testing.T` to do actual assertions or rename the function.

provider/marathon/marathon_test.go
```diff
@@ -30,6 +30,14 @@ func newFakeClient(applicationsError bool, applications marathon.Applications) *
...
+	actual := ""
+	for _, serviceName := range serviceNames {
+		actual = getActual(serviceName)
+	}
```
timoreimann: I don't get the for loop: It iterates through all service names and will eventually only store the final one in `actual`. Why is that?

provider/marathon/marathon.go
```diff
@@ -265,7 +281,87 @@ func isApplicationEnabled(application marathon.Application, exposedByDefault boo
...
+	if len(serviceName) > 0 {
+		serviceName = strings.Replace(serviceName, "/", "-", -1)
+		serviceName = strings.Replace(serviceName, ".", "-", -1)
+		return "-service-" + serviceName
+	}
+	return ""
+}
+
+//getAppLabel is a convenience function to get application label, when no serviceName is available
+//it is identical in calling getLabel(application, label, "")
```
timoreimann: Hmm, is it _in_ or _to_?

provider/marathon/marathon.go
```diff
@@ -284,74 +399,83 @@ func (p *Provider) getPort(task marathon.Task, application marathon.Application)
...
 		return strings.Split(entryPoints, ",")
 	}
 	return []string{}
 }
 
 // getFrontendRule returns the frontend rule for the specified application, using
-// it's label. It returns a default one (Host) if the label is not present.
-func (p *Provider) getFrontendRule(application marathon.Application) string {
-	if label, ok := p.getLabel(application, types.LabelFrontendRule); ok {
+// it's label. If service is provided, it will look for serviceName label before generic one.
```
timoreimann: s/it's/its/

provider/marathon/marathon.go
```diff
@@ -421,33 +545,36 @@ func (p *Provider) hasHealthCheckLabels(application marathon.Application) bool {
...
 		return strings.Split(basicAuth, ",")
 	}
 
 	return []string{}
 }
 
-func processPorts(application marathon.Application, task marathon.Task) (int, error) {
-	if portLabel, ok := (*application.Labels)[types.LabelPort]; ok {
-		port, err := strconv.Atoi(portLabel)
+//processPorts validates `portValue` to be valid numeric port,
```
timoreimann: Let's end with a punctuation here and start a new sentence on the next line.

provider/marathon/marathon.go
```diff
@@ -421,33 +545,36 @@ func (p *Provider) hasHealthCheckLabels(application marathon.Application) bool {
...
+// if its empty, proceeds to extract all ports and attempt to retrieve a port specified by the 'portIndexValue', given that its a valid integer
```
timoreimann: s/if its empty/if it's empty/

timoreimann: s/attempt/attempts/

timoreimann: s/given that its a valid integer/given that it's a valid integer/

provider/marathon/marathon_test.go
aantono: The expected values are already in the test table, the reason for the function is to extract the common boiler-plate code to loop through the service names to get the right value...

provider/marathon/marathon.go
```diff
@@ -421,33 +546,36 @@ func (p *Provider) hasHealthCheckLabels(application marathon.Application) bool {
...
 	return []string{}
 }
 
-func processPorts(application marathon.Application, task marathon.Task) (int, error) {
-	if portLabel, ok := (*application.Labels)[types.LabelPort]; ok {
-		port, err := strconv.Atoi(portLabel)
+//processPorts validates `portValue` to be valid numeric port.
+// if it is empty, proceeds to extract all ports and attempts to retrieve a port specified by the 'portIndexValue', given that it is a valid integer.
+// if all fails, returns first port (at index 0) from the list.
+func processPorts(application marathon.Application, task marathon.Task, portValue string, portIndexValue string) (int, error) {
```
timoreimann: Could you please shorten the specification of the last two function parameters like `portValue, portIndexValue string`.

provider/marathon/builder_test.go
```diff
@@ -33,6 +38,16 @@ func label(key, value string) func(*marathon.Application) {
 	}
 }
 
+func labelWithService(key, value string, serviceName string) func(*marathon.Application) {
+	if len(serviceName) > 0 {
```
aantono: done

aantono:
Hello @timoreimann.  We did indeed done a LOT of refactoring (for the better, I agree), so the code has evolved and got simplified.  In essence, you are correct, the service-specific methods got removed and instead the `getLabel` method now expects a `serviceName` argument, which, if given non-empty, will retrieve the label for specified service, and if empty or no service-label exists, will fall back to look for application generic.

Not sure what exactly troubles you about the ports logic specifically, as it is no different from how we get service-specific weights, protocol, rules, etc.  I just did another look-over the `getPort` method and it seems pretty straight forward, especially since we separated the logic between it and `processPorts` fairly cleanly.  We first check if there is an explicit label (service or application level) for port, if not, proceed to check for port index label (service or app) and if that's not there either, just return the first port from the list.  In all cases, we pass the string-value to `processPorts` to ensure that the value is a valid integer and in case of an index, that it exists in the available ports.

provider/marathon/builder_test.go
```diff
@@ -33,6 +38,16 @@ func label(key, value string) func(*marathon.Application) {
...
+		property := strings.TrimPrefix(key, types.LabelPrefix)
+		return func(app *marathon.Application) {
+			app.AddLabel(types.LabelPrefix+serviceName+"."+property, value)
+		}
+	}
+	panic("serviceName can not be empty")
```
timoreimann: Minor thingie: Please implement this like

```go
if len(serviceName) == 0 {
  panic("serviceName can not be empty")
}
```

One level of indentation less to worry about; and it co-locates the root cause with the error.

aantono: done

timoreimann:
@aantono 

1. one thing that I don't find ideal is that we call `processPorts` three times from `mustGetPort`. AFAICS, we could simply collect the port and port index in `mustGetPort`, call `processPorts` once, and let it do all the hard-lifting with the general approach mimicking what we had in `processPorts` before this PR already (which is to return the port right away if it's explicitly given, and otherwise fetch the available ports and try to apply the optional index).
1. The name `mustGetPort` is a bit misleading: In Go, functions/methods prefixed by `must` conventionally panic if they hit an error (see regexp's `MustCompile` for example). So we should rename the method.
1. `getPort` is not logging any errors anymore. We should log any error returned.

So these points should get addressed. Instead of doing so individually, however, I think we can just return to `getPorts` calling `processPorts` and move the implementation currently living in `mustGetPort` into `processPorts`. This is what I meant by going back to the previous implementation, which IMO could make things simpler overall.

aantono:
While `processPorts` does get referenced 3 times, it only gets called once for each given call to `mustGetPort`.  The code is structured such a way that it would return something after the `processPorts` gets called...  For example, we only call `processPorts` if a port label is present, or if a portIndex label is present, or if no labels are present.  But there never will be 2 or 3 calls to `processPorts` as a result of a single `mustGetPort` call.    
As for error logging, I don't see how that's not the case, since `getPort` is nothing but a wrapper around `mustGetPort` call, which does log all the errors inside. (unless I'm missing something).
Do you have a suggested name for `mustGetPort`, would `getPortWithError` be ok?

Overall, I'm not entirely sure, to be honest, what you mean by "going back to the previous implementation", cause at this point it feels like we've just went full circle.  Do you mean to revert the changes to `getPort` and `processPorts` to their previous state, or you have something else in mind entirely?

timoreimann:
@aantono Code speaks louder than words, so I went ahead and made the change I had in mind.

The tests stay green on my local machine. I also changed the task filtering around the service port slightly: AFAICS, it would previously only filter out tasks if the _final_ service was missing its port (but not if, say, the first one did).

Please let me know what you think.

provider/marathon/marathon.go
```diff
@@ -233,14 +242,21 @@ func (p *Provider) taskFilter(task marathon.Task, application marathon.Applicati
...
-		log.Errorf("Filtering Marathon task %s from application %s without port: %s", task.ID, application.ID, err)
-		return false
+	for _, serviceName := range p.getServiceNames(application) {
+		foundPort := false
+		if port := p.getPort(task, application, serviceName); len(port) > 0 {
+			foundPort = true
+		}
+
+		if !foundPort {
+			log.Errorf("Filtering Marathon task %s from application %s without port", task.ID, application.ID)
```
aantono: One issue I see here is that if the application is missing an explicit port label for a service, it will fail.  The scenario where you would not have an app specific port, but say an app specific frontend.rule is valid otherwise.  So looping through all the available services to see if any of them has a port config was suppose to ensure that we find at least one port somewhere, which is enough to not filter the task.

provider/marathon/marathon_test.go
```diff
@@ -424,6 +424,16 @@ func TestMarathonTaskFilter(t *testing.T) {
...
 		},
 		{
+			desc: "single service without port",
+			task: createTask(taskPorts(80, 81)),
+			application: createApplication(
+				appPorts(80, 81),
+				labelWithService(types.LabelPort, "80", "web"),
+				labelWithService(types.LabelPort, "illegal", "admin"),
+			),
+			expected: false,
```
aantono: Why would this combination result in a task being filtered out?  Do we think that if the task has services, EVERY service MUST have a port or portIndex label?

timoreimann:
@aantono updated per our Slack discussion.

We are lacking a bit of test coverage right now by. I'll have to ponder a bit about the implications / the need to make the latest code more testable.

aantono:
Sounds good.  The latest changes you added LGTM, let me know if there is anything specific on the testing side of things you are concerned about to see if I can help.  Thanks!!!

timoreimann:
@containous/docker could someone please review the tiny improvement to the Docker regex?

timoreimann:
@aantono can I leave the resolution of the conflicts to you? I probably won't have too much time over the next couple of days.

aantono:
Sounds good.  I'll rebase as soon as I get home.

aantono:
Rebased, should be good to merge now...

timoreimann:
@aantono the build is red now. Something must have gone wrong during your rebase. Could you check please?

timoreimann:
I'm going to have a final look at the diff. Please don't merge yet.

ldez:
@timoreimann Sorry I did not see your comment 😞 
Next time add `bot/need-human-merge`

timoreimann:
@ldez no worries, will do next time.

Everything looks fine anyway, so no harm done. :)

Issue #1744, opened 2017-06-13
Ingresses for a sticky services with a single ingress path can get multiple cookies


thomas-b-jackson:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

* deploy a web app on k8s with multiple pods (3 app pods in our case)
   * app serves many endpoints on multiple paths (i.e. https://app.domain.com/path1, /path2, etc.)
* create an k8s service that selects the pods and requests stickiness through traefik.backend.loadbalancer.sticky: "true"
* deploy traefic ingress controller pods via a k8s deployment (3 traefik pods in our case)
    * configure traefik to watch a single namespace ("sets" in our case)
* create a k8s ingress resource with a single host
   * path: /
   * kubernetes.io/ingress.class: traefik
* start hitting endpoint from a single client via a web browser

### What did you expect to see?

Since stickiness is configured, and since a single, global path was specified, all requests from a given client browser session should be directed to the same app pod.

### What did you see instead?

For some web clients, all requests go to the same app pod.
   * these client have a single traefik backend cookie
![snip20170612_9](https://user-images.githubusercontent.com/7053750/27092428-27107fe8-5018-11e7-8e9e-77429fe1df1b.png)

For some other web clients, requests for path path1 always to to pod A, request for path2 got to pod B, etc.
   * clients affected always have multiple traefik backend cookies, one per path
![screen shot 2017-06-12 at 4 18 20 pm](https://user-images.githubusercontent.com/7053750/27092432-2c0c60ca-5018-11e7-986d-d4a3dd11f9d1.png)
   * as many as 4 cookies, each with a different path have been observed

### Output of `traefik version`
1.3.0

### What is your environment & configuration

traefik config
```yml
apiVersion: v1
data:
  traefik.toml: |
    # traefik.toml
    logLevel = "INFO"
    defaultEntryPoints = ["http"]
    [entryPoints]
      [entryPoints.http]
      address = ":80"
    [kubernetes]
    namespaces = ["sets"]
kind: ConfigMap
metadata:
  creationTimestamp: 2017-06-06T23:08:37Z
  labels:
    app: private-ingress
  name: private-ingress
  namespace: sets
  resourceVersion: "105250320"
  selfLink: /api/v1/namespaces/sets/configmaps/private-ingress
  uid: 12642da1-4b0d-11e7-a9df-0259bfeee5dc
```

app ingress resource
```yml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
[...]

PR #1748, opened 2017-06-14
return 503 on empty backend

### LINKED ISSUE #1688: Return 503 on missing or empty backend (with available frontend)
Traefik currently returns a 404 when the frontend is missing. When the frontend is available but the backend is either missing or empty (where the latter means that no servers are registered), we should return a 503 (Service Unavailable).

At least during my tests with Kubernetes, I found Traefik returning a 500 for such situations currently.

Note that this issue is _not_ supposed to make sure that providers configure frontends and backends accordingly in order for the distinction specified here to be reached. Its job is to make sure the right response code is returned if the configuration is adhering to the specification above.

Refs #1077.

marco-jantke:
This PR changes the response code Traefik sends to 503, given it has an empty backend. 

This should resolve the issue #1688. In the issue it is furthermore mentioned that a 503 should also be sent when a backend is completely missing. At the moment though, a frontend is not loaded in the `loadConfig` method, given there is no dedicated backend for it. See [here](https://github.com/containous/traefik/blob/2ddae2e856379f7ce2fb9b6509712e1ad523deeb/server/server.go#L689). This looks to me that this case is actually not intentioned to happen. @dtomcej and @timoreimann you had quite some discussions about this topic, WDYT about that case?

Fixes #1688

### DIFF
--- a/middlewares/empty_backend_handler_test.go
+++ b/middlewares/empty_backend_handler_test.go
@@ -1,0 +1,70 @@
+package middlewares
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/containous/traefik/testhelpers"
+	"github.com/vulcand/oxy/roundrobin"
+)
+
+func TestEmptyBackendHandler(t *testing.T) {
+	tests := []struct {
+		amountServer   int
+		wantStatusCode int
+	}{
+		{
+			amountServer:   0,
+			wantStatusCode: http.StatusServiceUnavailable,
+		},
+		{
+			amountServer:   1,
+			wantStatusCode: http.StatusOK,
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+
+		t.Run(fmt.Sprintf("amount servers %d", test.amountServer), func(t *testing.T) {
+			t.Parallel()
+
+			nextHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+			})
+			handler := NewEmptyBackendHandler(&healthCheckLoadBalancer{test.amountServer}, nextHandler)
+
+			recorder := httptest.NewRecorder()
+			req := httptest.NewRequest(http.MethodGet, "http://localhost", nil)
+
+			handler.ServeHTTP(recorder, req)
+
+			if recorder.Result().StatusCode != test.wantStatusCode {
+				t.Errorf("Received status code %d, wanted %d", recorder.Result().StatusCode, test.wantStatusCode)
+			}
+		})
+	}
+}
+
+type healthCheckLoadBalancer struct {
+	amountServer int
+}
+
+func (lb *healthCheckLoadBalancer) RemoveServer(u *url.URL) error {
+	return nil
+}
+
+func (lb *healthCheckLoadBalancer) UpsertServer(u *url.URL, options ...roundrobin.ServerOption) error {
+	return nil
+}
+
+func (lb *healthCheckLoadBalancer) Servers() []*url.URL {
+	servers := make([]*url.URL, lb.amountServer)
+	for i := 0; i < lb.amountServer; i++ {
+		servers = append(servers, testhelpers.MustParseURL("http://localhost"))
+	}
+	return servers
+}
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -3,6 +3,7 @@ package server
 import (
 	"fmt"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"reflect"
 	"testing"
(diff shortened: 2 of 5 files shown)


timoreimann:
@marco-jantke If I look at the Kubernetes provider (which we kind of implicitly used as the canonical implementation for a proper 404/503 distinction), what we do there is [start off with a non-empty backend](https://github.com/containous/traefik/blob/2ddae2e856379f7ce2fb9b6509712e1ad523deeb/provider/kubernetes/kubernetes.go#L151-L157) as soon as we have a legitimate Ingress, [remove the entire frontend](https://github.com/containous/traefik/blob/2ddae2e856379f7ce2fb9b6509712e1ad523deeb/provider/kubernetes/kubernetes.go#L227) if the Service turns out to be missing, and [populate the backend servers](https://github.com/containous/traefik/blob/2ddae2e856379f7ce2fb9b6509712e1ad523deeb/provider/kubernetes/kubernetes.go#L281-L284) if we have found proper Endpoints. With this approach, the Kubernetes provider produces `nil` backends only if the Ingress doesn't include any paths (because [this loop](https://github.com/containous/traefik/blob/2ddae2e856379f7ce2fb9b6509712e1ad523deeb/provider/kubernetes/kubernetes.go#L149) will never enter its body), and otherwise creates backends that fit the specification.

So overall, this should work. `nil` backends not causing a frontend to be built (and, in turn, leading to a 404) sounds like a reasonable assumption since it's easy to end up with an error. Reaching a 503 should require some (minimum) effort with each provider.

@dtomcej WDYT?

dtomcej:
I agree with the scenario @timoreimann proposed.

marco-jantke:
Ok, when I understand it correctly, than this PR should actually deliver the needs we have for the 503 responses. 👍 for the quick responses, thanks a lot!

server/server_test.go
```diff
@@ -516,3 +517,164 @@ type okHTTPHandler struct{}
...
+						},
+					},
+					Backends: map[string]*types.Backend{
+						"backend": {
+							Servers:      map[string]types.Server{},
+							LoadBalancer: &types.LoadBalancer{Method: "Wrr", Sticky: true},
+						},
+					},
+				}
+			},
```
timoreimann: Should we add a parameterizable helper function or (maybe even better) apply the builder pattern to construct the frontends and backends for test purposes?

We have tests like the ones in `provider/kubernetes/kubernetes_test.go` which are using inline structs so extensively, it has become fairly hard to understand and extend the tests. I'd like to make sure we're not doing the same for newish tests.

ldez: @marco-jantke see the excelent article from @vdemeester http://vincent.demeester.fr/posts/2017-01-01-go-testing-functionnal-builders/

marco-jantke: Thanks for the nice article recommendation. I applied the pattern there and I think the readability of the tests improved quite a lot. Furthermore I tried to keep the builder generic, so that they can easily be reused and extended by other tests in the future.

timoreimann: Big improvement indeed! Me ❤️ it too!

djalal:
@timoreimann what about adding logs ? 

an "empty backend" is not a nominal state, and should throw a WARNING, just like events "no route to host" or "connect timeout"

timoreimann:
@djalal the events preceeding a potential 503 should eventually be logged in each provider. The Kubernetes one does this already, others should follow. I think this should be the first priority as you likely want to have a logging notification before messages actually get routed incorrectly.

I saw you created a dedicated issue too, so I'll place the second part of my answer there. :-)

timoreimann:
@emilevauge what part of the design review is still missing from your perspective?

We seem to have discussed to topic at great length on Slack.

marco-jantke:
Is there anything I can do to bring this PR forward?

emilevauge:
@marco-jantke Should we return 503 each time `RebalancerErrorHandler` is called? Does `RebalancerErrorHandler` really fit our needs here?

marco-jantke:
@emilevauge I checked the implementation of the load balancers in detail now.

The case I am describing holds true for the RoundRobin LB as well as for the Rebalancer LB implementation. They call the error handler you pass them basically in two cases:

1. `nextServer` returns an error. This is because they can not find any valid server in the list they can route to. For that case it is correct to return the 503.
2. when sticky sessions are enabled and the server in the cookie is not yet available anymore. For that case currently we also serve a 503, but before we served a 500 because of the default error handler implementation that was used.

I think for case number 2 even the former implementation is not complete. I can't see anything that would invalidate the cookie again at some point and clients with that cookie won't be able to connect to a new Server, when the server list of the backend changes. AFAICS we should just remove the error handling in the case this happens and try to connect to a new Server in the backend list, for the following two places. When it can select a new Server it will set the cookie again, effectively overriding the old cookie and in case no Server could be found we want to still have our 503:

- [Rebalancer](https://github.com/containous/traefik/blob/master/vendor/github.com/vulcand/oxy/roundrobin/rebalancer.go#L156)
- [LoadBalancer](https://github.com/containous/traefik/blob/master/vendor/github.com/vulcand/oxy/roundrobin/rr.go#L80)

WDYT about it?

emilevauge:
@marco-jantke I'm sorry but I don't agree to use `RebalancerErrorHandler` as a workaround to know if a backend is empty or not. We should not assume why an error is set in the `errHandler` by the load balancer. It could even change in the future.
Couldn't you use the number of servers in a backend directly?

marco-jantke:
I get your point. I actually did not perceive using the LoadBalancers `ErrorHandler` as a workaround, as they are kind of the canonical place to keep track of the service place. Anyway I think we can just directly check when building the Backend whether it has no Servers and use an http.Handler that directly returns 503 without the need of even setting up LoadBalancers or the Forwarder. Does this sound like an approach you imagine? I would give it a shot, should be easy to verify as my tests should not change.

As additional point, WDYT about the "potential problem" I described above regarding sticky sessions and a change of the Server list? 

> I think for case number 2 even the former implementation is not complete. I can't see anything that would invalidate the cookie again at some point and clients with that cookie won't be able to connect to a new Server, when the server list of the Backend changes. AFAICS we should just remove the error handling in the case this happens and try to connect to a new Server in the backend list, for the following two places. When it can select a new Server it will set the cookie again, effectively overriding the old cookie and in case no Server could be found we want to still have our 503.  [Rebalancer](https://github.com/containous/traefik/blob/master/vendor/github.com/vulcand/oxy/roundrobin/rebalancer.go#L156) [LoadBalancer](https://github.com/containous/traefik/blob/master/vendor/github.com/vulcand/oxy/roundrobin/rr.go#L80)

I had no time yet to proof this point and I am not sure whether I will have time in the near future as sticky sessions are not on my list of required features, but maybe it is worth opening an issue for this to enable further investigation in the future or by others?

emilevauge:
>Anyway I think we can just directly check when building the Backend whether it has no Servers and use an http.Handler that directly returns 503 without the need of even setting up LoadBalancers or the Forwarder. Does this sound like an approach you imagine? I would give it a shot, should be easy to verify as my tests should not change.

Indeed, that's a simple solution. The only issue is that healthchecks can add/remove servers from a lb dynamically. So this solution can't be that simple :)

On the sticky session thing, couldn't we just make the cookie expire in this case?

timoreimann:
From what I can see, the sticky session implementation does reroute to a different server if the one associated with the cookie is knowingly gone.

Here's the flow how I see it:

1. The [ServeHTTP method](https://github.com/containous/traefik/blob/a89eb122a07f35b8ad2b5ede732925d0b7cc62b6/vendor/github.com/vulcand/oxy/roundrobin/rr.go#L73) is invoked.
1. [`GetBackend` is invoked](https://github.com/containous/traefik/blob/a89eb122a07f35b8ad2b5ede732925d0b7cc62b6/vendor/github.com/vulcand/oxy/roundrobin/rr.go#L78) with the list of currently known servers (which should also be the one that [the health check logic manipulates](https://github.com/containous/traefik/blob/a89eb122a07f35b8ad2b5ede732925d0b7cc62b6/healthcheck/healthcheck.go#L55-L59)).
1. `GetBackend` [checks if the if the server is still alive](https://github.com/containous/traefik/blob/a89eb122a07f35b8ad2b5ede732925d0b7cc62b6/vendor/github.com/vulcand/oxy/roundrobin/stickysessions.go#L33). This is the case if the URL from the cookie [can be found within the list of servers](https://github.com/containous/traefik/blob/a89eb122a07f35b8ad2b5ede732925d0b7cc62b6/vendor/github.com/vulcand/oxy/roundrobin/stickysessions.go#L52-L54) (called `haystack` in the code).
1. If the server cannot be found, we [try the next until one is found or we have depleted them all](https://github.com/containous/traefik/blob/a89eb122a07f35b8ad2b5ede732925d0b7cc62b6/vendor/github.com/vulcand/oxy/roundrobin/rr.go#L91-L96). _Only in the latter case_ we seem to call the error handler.
1. While we [do update the sticky cookie with the new URL](https://github.com/containous/traefik/blob/a89eb122a07f35b8ad2b5ede732925d0b7cc62b6/vendor/github.com/vulcand/oxy/roundrobin/rr.go#L99), we unfortunately [call `http.SetCookie()`](https://github.com/containous/traefik/blob/a89eb122a07f35b8ad2b5ede732925d0b7cc62b6/vendor/github.com/vulcand/oxy/roundrobin/stickysessions.go#L42). Unlike the name of the function may imply, [it only adds a cookie header](https://golang.org/src/net/http/cookie.go?s=3118:3166#L122) but does not set/overwrite an existing one.

marco-jantke:
@timoreimann thanks for the clarification on the cookie issue! I in fact misinterpreted the behaviour and apart of the already open issues. To clarify a bit more, it is right to use the `Set-Cookie` header in the Servers response. In fact there is no other way to set the cookie from the server-side AFAIK. In the issue #1744 its only the problem that multiple different paths are specified and therefore the cookie gets seemingly re-created. A cookie is considered unique and will be overwritten in case the cookie_name, domain and path parts are equal. Said this, I think further communication about the cookie issue should be moved to the dedicated issue then.

timoreimann:
@marco-jantke agreed to moving the cookie discussion to a different issue.

Final comment: [`SetCookie` does not seem to overwrite identical (per your definition) cookies](https://play.golang.org/p/w7TISJ_mnT).

emilevauge:
@marco-jantke can we also agree to change the design, and directly use the number of servers in backend (being careful to health-checks) ?

marco-jantke:
Yes, I had a look at the implementation and this sounds reasonable to me. So I would write a middleware that gets a `healthcheck.LoadBalancer` and on each request checks whether there are active Servers. If this is not the case it will directly serve a 503 response. Does this sound reasonable to you?

I will update this PR accordingly.

marco-jantke:
@emilevauge @timoreimann PTAL again. I refactored the implementation now to a middleware that knows about the `healthcheck.LoadBalancer` implementation and responds with a 503, given a Backend is empty.

middlewares/empty_backend_handler_test.go
```diff
@@ -0,0 +1,70 @@
...
+		t.Run(fmt.Sprintf("amount servers %d", test.amountServer), func(t *testing.T) {
+			t.Parallel()
+
+			nextHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.WriteHeader(http.StatusOK)
+			})
+			handler := NewEmptyBackendHandler(&healthCheckLoadBalancer{test.amountServer}, nextHandler)
+
+			recorder := httptest.NewRecorder()
+			req := httptest.NewRequest("GET", "http://localhost", nil)
```
timoreimann: `s/"GET"/http.MethodGet/`

marco-jantke: Always. So thanks for reminding me constantly Timo. At some point I will also learn it, don't give up the hope :D 🐼

emilevauge:
Design LGTM
Thanks @marco-jantke :)

PR #1755, opened 2017-06-15
add status code to request duration metric

marco-jantke:
Its best practice to monitor request durations partitioned by the status code of the response.  4xx/5xx usually have a different response time then succeeding requests. In case of 4xx they are most of the time very fast and depending on the type of server error, 5xx are also either very fast or very slow. Having no differentiation of those would distort calculations on the request durations, e.g. when you want to calculate an average.

Adding the label to the metric is not breaking existing Prometheus queries and is therefore backwards compatible.

### DIFF
--- a/middlewares/metrics.go
+++ b/middlewares/metrics.go
@@ -42,9 +42,12 @@ func (m *MetricsWrapper) ServeHTTP(rw http.ResponseWriter, r *http.Request, next
 	start := time.Now()
 	prw := &responseRecorder{rw, http.StatusOK}
 	next(prw, r)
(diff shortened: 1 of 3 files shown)


ldez:
Every time someone does not respect the issue's template, a little bunny dies.

southpark - estrella kill rabbit
southpark - estrella kill rabbit2

Please edit your issue description to respect the [issue template](https://raw.githubusercontent.com/containous/traefik/master/.github/ISSUE_TEMPLATE.md) as more as possible.

marco-jantke:
Its not an issue, do I have to first open one or create my PR in an issue like format when there is no issue for it?

ldez:
refresh, sorry mistake.

Issue #1696, opened 2017-06-01
Add support for log file rotation


rjshep:
Traefik can currently write to two log files, the traefik log and an access log.  These log files grow continuously and there is no mechanism for rotating, compressing, or archiving them.

I propose adding functionality for Traefik to handle a SIGUSR1 signal, on receipt of which any (file-backed) logs in use will be closed and opened again.  This will allow tools such as logrotate to be used to manage the logs.

juherr:
I was looking for this feature and it looks logrotate may already work with traefik files: https://github.com/mantl/mantl-packaging/pull/9

rjshep:
There's nothing there to tell Traefik to actually close and reopen the file though.  Without that, even if log rotate moves and compresses the files, Traefik will continue to use the existing file handles to write to the old files.  Unless I'm missing something?

emilevauge:
@rjshep seems like a good idea, but do you if it's possible to handle this easily with logrus ?

rjshep:
@emilevauge Yes we've already implemented this on our fork and it's working fine. Just wanted to check it's something you guys would accept a PR for - sounds like it is?

emilevauge:
Let's go then :)

ovizii:
Would someone please share their logrotate configuration which works with traefik running as a docker container?

PR #1761, opened 2017-06-16
Add log file close and reopen on receipt of SIGUSR1

rjshep:
See Issue [#1696](https://github.com/containous/traefik/issues/1696)

Traefik can currently write to two log files, the traefik log and an access log. These log files grow 
continuously and there is no mechanism for rotating, compressing, or archiving them.

This PR lets Traefik to respond to a USR1 signal by closing and reopening its log files, allowing external software, such as log rotate to be used for the above purposes.

### DIFF
--- a/integration/log_rotation_test.go
+++ b/integration/log_rotation_test.go
@@ -1,0 +1,141 @@
+// +build !windows
+
+package integration
+
+import (
+	"bufio"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"syscall"
+	"time"
+
+	"github.com/containous/traefik/integration/try"
+	"github.com/go-check/check"
+	checker "github.com/vdemeester/shakers"
+)
+
+// Log rotation integration test suite
+type LogRotationSuite struct{ BaseSuite }
+
+func (s *LogRotationSuite) TestAccessLogRotation(c *check.C) {
+	// Start Traefik
+	cmd, _ := s.cmdTraefik(withConfigFile("fixtures/access_log_config.toml"))
+	err := cmd.Start()
+	c.Assert(err, checker.IsNil)
+	defer cmd.Process.Kill()
+	defer os.Remove(traefikTestAccessLogFile)
+	defer os.Remove(traefikTestLogFile)
+
+	// Verify Traefik started ok
+	verifyEmptyErrorLog(c, "traefik.log")
+
+	// Start test servers
+	ts1 := startAccessLogServer(8081)
+	defer ts1.Close()
+
+	// Allow time to startup
+	time.Sleep(1 * time.Second)
+
+	// Make some requests
+	err = try.GetRequest("http://127.0.0.1:8000/test1", 500*time.Millisecond)
+	c.Assert(err, checker.IsNil)
+
+	// Rename access log
+	err = os.Rename(traefikTestAccessLogFile, traefikTestAccessLogFile+".rotated")
+	c.Assert(err, checker.IsNil)
+
+	// in the midst of the requests, issue SIGUSR1 signal to server process
+	err = cmd.Process.Signal(syscall.SIGUSR1)
+	c.Assert(err, checker.IsNil)
+
+	// continue issuing requests
+	_, err = http.Get("http://127.0.0.1:8000/test1")
+	c.Assert(err, checker.IsNil)
+	_, err = http.Get("http://127.0.0.1:8000/test1")
+	c.Assert(err, checker.IsNil)
+
+	// Verify access.log.rotated output as expected
+	lineCount := verifyLogLines(c, traefikTestAccessLogFile+".rotated", 0, true)
+	c.Assert(lineCount, checker.GreaterOrEqualThan, 1)
+
+	// Verify access.log output as expected
+	lineCount = verifyLogLines(c, traefikTestAccessLogFile, lineCount, true)
+	c.Assert(lineCount, checker.Equals, 3)
+
+	verifyEmptyErrorLog(c, traefikTestLogFile)
+}
+
+func (s *LogRotationSuite) TestTraefikLogRotation(c *check.C) {
+	// Start Traefik
+	cmd, _ := s.cmdTraefik(withConfigFile("fixtures/traefik_log_config.toml"))
+	err := cmd.Start()
+	c.Assert(err, checker.IsNil)
+	defer cmd.Process.Kill()
+	defer os.Remove(traefikTestAccessLogFile)
+	defer os.Remove(traefikTestLogFile)
+
+	// Ensure Traefik has started
+	err = try.GetRequest("http://127.0.0.1:8000/test1", 500*time.Millisecond)
+	c.Assert(err, checker.IsNil)
+
+	// Rename traefik log
+	err = os.Rename(traefikTestLogFile, traefikTestLogFile+".rotated")
+	c.Assert(err, checker.IsNil)
+
+	// issue SIGUSR1 signal to server process
+	err = cmd.Process.Signal(syscall.SIGUSR1)
+	c.Assert(err, checker.IsNil)
+
+	err = cmd.Process.Signal(syscall.SIGTERM)
+	c.Assert(err, checker.IsNil)
+
+	// Allow time for switch to be processed
+	err = try.Do(3*time.Second, func() error {
+		_, err = os.Stat(traefikTestLogFile)
+		return err
+	})
+	c.Assert(err, checker.IsNil)
+
+	// we have at least 6 lines in traefik.log.rotated
+	lineCount := verifyLogLines(c, traefikTestLogFile+".rotated", 0, false)
+
+	// GreaterOrEqualThan used to ensure test doesn't break
+	// If more log entries are output on startup
+	c.Assert(lineCount, checker.GreaterOrEqualThan, 6)
+
+	//Verify traefik.log output as expected
+	lineCount = verifyLogLines(c, traefikTestLogFile, lineCount, false)
+	c.Assert(lineCount, checker.GreaterOrEqualThan, 7)
+}
+
+func verifyEmptyErrorLog(c *check.C, name string) {
+	err := try.Do(5*time.Second, func() error {
+		traefikLog, e2 := ioutil.ReadFile(name)
+		if e2 != nil {
+			return e2
+		}
+		c.Assert(traefikLog, checker.HasLen, 0)
+		return nil
+	})
+	c.Assert(err, checker.IsNil)
+}
+
+func verifyLogLines(c *check.C, fileName string, countInit int, accessLog bool) int {
+	rotated, err := os.Open(fileName)
+	c.Assert(err, checker.IsNil)
+	rotatedLog := bufio.NewScanner(rotated)
+	count := countInit
+	for rotatedLog.Scan() {
+		line := rotatedLog.Text()
+		c.Log(line)
+		if accessLog {
+			if len(line) > 0 {
+				CheckAccessLogFormat(c, line, count)
+			}
+		}
+		count++
+	}
+
+	return count
+}
--- a/log/logger.go
+++ b/log/logger.go
@@ -2,14 +2,18 @@ package log
 
 import (
 	"bufio"
+	"fmt"
 	"io"
+	"os"
 	"runtime"
 
 	"github.com/Sirupsen/logrus"
 )
 
 var (
-	logger *logrus.Entry
+	logger      *logrus.Entry
+	logFilePath string
+	logFile     *os.File
 )
 
 func init() {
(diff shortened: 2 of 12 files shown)


marco-jantke:
Isn't SIGHUP usually used for signalling reloading of logs/configs files?

rjshep:
There isn't a standard from what I can tell, both are used, but SIGUSR1 seems a better fit as SIGHUP is more about process restart, which we're not doing.

marco-jantke:
Not 100% sure about this one, all programs I worked with were listening for a SIGHUP to reopen log files and the [logrotate docs](https://linux.die.net/man/8/logrotate) show it also as canonical example.

emilevauge:
@marco-jantke logrotate docs uses systemd, which indeed reloads its configuration files on SIGHUP, after restarting ;)
@rjshep's suggestion is identical to Nginx behavior https://nginx.org/en/docs/control.html.
There is no standard, but sticking with Nginx signal management seems good to me:)

marco-jantke:
Thanks for the info! LGTM :)

emilevauge:
Design: LGTM

ldez:
@rjshep Could rebase and solve conflicts?

integration/log_rotation_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/go-check/check"
+	checker "github.com/vdemeester/shakers"
+)
+
+// Log rotation integration test suite
+type LogRotationSuite struct{ BaseSuite }
+
+func (s *LogRotationSuite) TestAccessLogRotation(c *check.C) {
+	// Start Traefik
+	cmd := exec.Command(traefikBinary, "--configFile=fixtures/access_log_config.toml")
```
ldez: could you replace this line by `cmd, _ := s.cmdTraefik(withConfigFile("fixtures/access_log_config.toml"))`

integration/log_rotation_test.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+	// combined, we have 3 lines; i.e. 1 in the initial access.log
+	c.Assert(count, checker.Equals, 3)
+
+	verifyEmptyErrorLog(c, "traefik.log")
+}
+
+func (s *LogRotationSuite) TestTraefikLogRotation(c *check.C) {
+	// Start Traefik
+	cmd := exec.Command(traefikBinary, "--configFile=fixtures/traefik_log_config.toml")
```
ldez: could you replace this line by `cmd, _ := s.cmdTraefik(withConfigFile("fixtures/access_log_config.toml"))`

integration/log_rotation_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func verifyEmptyErrorLog(c *check.C, name string) {
+	err := try.Do(30*time.Second, func() error {
+		traefikLog, e2 := ioutil.ReadFile(name)
+		if e2 != nil {
+			return e2
+		}
+		if len(traefikLog) > 0 {
+			fmt.Printf("%s\n", string(traefikLog))
+			c.Assert(len(traefikLog), checker.Equals, 0)
```
ldez: could replace this line by `c.Assert(traefikLog, checker.HasLen, 0)`

log/logger.go
```diff
@@ -190,6 +194,44 @@ func Fatalln(args ...interface{}) {
...
+	}
+
+	SetOutput(logFile)
+	return nil
+}
+
+// Close closes the log and sets the Output to stdout
+func Close() error {
+	logrus.SetOutput(os.Stdout)
+	return logFile.Close()
```
ldez: `logFile` can be `nil`, maybe you need to protect the call.

log/logger.go
```diff
@@ -190,6 +194,44 @@ func Fatalln(args ...interface{}) {
 	logger.Fatalln(args...)
 }
 
+// Open opens the log file using the specified path and
+func Open(path string) error {
```
ldez: maybe rename to `OpenFile`

log/logger.go
```diff
@@ -190,6 +194,44 @@ func Fatalln(args ...interface{}) {
...
+	if logFile, err = os.OpenFile(logFilePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666); err != nil {
+		return err
+	}
+
+	SetOutput(logFile)
+	return nil
+}
+
+// Close closes the log and sets the Output to stdout
+func Close() error {
```
ldez: maybe rename to `CloseFile`

log/logger.go
```diff
@@ -190,6 +194,44 @@ func Fatalln(args ...interface{}) {
...
+	logrus.SetOutput(os.Stdout)
+	return logFile.Close()
+}
+
+// Rotate closes and reopens the log file to allow for rotation
+// by an external source.  If the log isn't backed by a file then
+// does nothing.
+func Rotate() error {
```
ldez: maybe rename to `RotateFile`

rjshep:
Hi @ldez, I've made the requested changes but I can't seem to get Semaphore to build.  It says the build has failed but without giving an error - that I can see anyway.  Any ideas?

ldez:
@rjshep I will fix that pb.

integration/log_rotation_test.go
```diff
@@ -0,0 +1,153 @@
...
+	// Start test servers
+	ts1 := startAccessLogServer(8081)
+	defer ts1.Close()
+	ts2 := startAccessLogServer(8082)
+	defer ts2.Close()
+	ts3 := startAccessLogServer(8083)
+	defer ts3.Close()
+
+	// Allow time to startup
+	time.Sleep(1 * time.Second)
```
timoreimann: Any chance we can use some of our `Try*` functions from the `integration/try` package to poll the access log servers for readiness? We're trying to avoid having static timers these days which are susceptible to flakiness.

integration/log_rotation_test.go
```diff
@@ -0,0 +1,153 @@
+package integration
```
timoreimann: Could we use a build tag here to avoid running the test on Windows (where it would always fail to my understanding)? Users may want to run the tests natively, i.e., outside of our build container.

integration/log_rotation_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// Verify Traefik started ok
+	verifyEmptyErrorLog(c, "traefik.log")
+
+	// Start test servers
+	ts1 := startAccessLogServer(8081)
+	defer ts1.Close()
+	ts2 := startAccessLogServer(8082)
+	defer ts2.Close()
+	ts3 := startAccessLogServer(8083)
```
timoreimann: Why do we need three servers instead of just one?

integration/log_rotation_test.go
```diff
@@ -0,0 +1,153 @@
...
+		line := rotatedLog.Text()
+		c.Log("rl: " + line)
+		if len(line) > 0 {
+			CheckAccessLogFormat(c, line, count)
+		}
+		count++
+	}
+	c.Assert(count, checker.Equals, 1)
+
+	// Verify access.log output as expected
```
timoreimann: How about removing the duplicated block by introducing a loop over both log file names?

integration/log_rotation_test.go
```diff
@@ -0,0 +1,153 @@
...
+	// Start Traefik
+	cmd, _ := s.cmdTraefik(withConfigFile("fixtures/traefik_log_config.toml"))
+	err := cmd.Start()
+	c.Assert(err, checker.IsNil)
+	defer cmd.Process.Kill()
+	defer os.Remove(traefikTestAccessLogFile)
+	defer os.Remove(traefikTestLogFile)
+
+	// Allow time to startup
+	time.Sleep(1 * time.Second)
```
timoreimann: Same suggestion regarding a replacement of the Sleep by a `Try* `function.

integration/log_rotation_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// issue SIGUSR1 signal to server process and move traefik log
+	err = cmd.Process.Signal(syscall.SIGUSR1)
+	c.Assert(err, checker.IsNil)
+	os.Rename(traefikTestLogFile, traefikTestLogFile+".rotated")
+
+	cmd.Process.Signal(syscall.SIGTERM)
+
+	// Allow time for switch to be processed
+	time.Sleep(3 * time.Second)
```
timoreimann: A `Try*` function? If it doesn't work here, please use `try.Sleep`.

integration/log_rotation_test.go
```diff
@@ -0,0 +1,153 @@
...
+	// we have at least 6 lines in traefik.log.rotated
+	rotated, err := os.Open(traefikTestLogFile + ".rotated")
+	c.Assert(err, checker.IsNil)
+	rotatedLog := bufio.NewScanner(rotated)
+	count := 0
+	for rotatedLog.Scan() {
+		line := rotatedLog.Text()
+		c.Log("rl " + line)
+		count++
+	}
```
timoreimann: The "open and read file"  block happens repeatedly. Can we extract it into a helper function?

integration/log_rotation_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+func verifyEmptyErrorLog(c *check.C, name string) {
+	err := try.Do(30*time.Second, func() error {
+		traefikLog, e2 := ioutil.ReadFile(name)
+		if e2 != nil {
+			return e2
+		}
+		if len(traefikLog) > 0 {
+			fmt.Printf("%s\n", string(traefikLog))
+			c.Assert(traefikLog, checker.HasLen, 0)
```
timoreimann: While it may work inside the `try.Do` block, ideally the check should happen outside. A closure could be helpful to achieve that.

log/logger.go
```diff
@@ -190,6 +194,48 @@ func Fatalln(args ...interface{}) {
 	logger.Fatalln(args...)
 }
 
+// OpenFile opens the log file using the specified path and
```
timoreimann: s/ and//.

log/logger.go
```diff
@@ -190,6 +194,48 @@ func Fatalln(args ...interface{}) {
...
+func OpenFile(path string) error {
+	logFilePath = path
+	var err error
+	if logFile, err = os.OpenFile(logFilePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666); err != nil {
```
timoreimann: Personally, I'd do

```go
logFile, err := ...
if err != nil {
    ...
}
```

to avoid having to declare `err`.

log/logger.go
```diff
@@ -190,6 +194,48 @@ func Fatalln(args ...interface{}) {
 	logger.Fatalln(args...)
 }
 
+// OpenFile opens the log file using the specified path and
+func OpenFile(path string) error {
+	logFilePath = path
```
timoreimann: Why use the extra `logFilePath` variable? Can't we pass in `path` directly?

log/logger.go
```diff
@@ -190,6 +194,48 @@ func Fatalln(args ...interface{}) {
...
+
+	if logFile != nil {
+		return logFile.Close()
+	}
+	return nil
+}
+
+// RotateFile closes and reopens the log file to allow for rotation
+// by an external source.  If the log isn't backed by a file then
+// does nothing.
```
timoreimann: "it does nothing"?

middlewares/accesslog/logger.go
```diff
@@ -139,6 +140,22 @@ func (l *LogHandler) Close() error {
 	return l.file.Close()
 }
 
+// Rotate closes and reopens the log file to allow for rotation
+// by an external source.
+func (l *LogHandler) Rotate() error {
+	var err error
```
timoreimann: Should suffice to be declared on line 150 only if we also scope `err` on line 147 to the if statement?

middlewares/accesslog/logger.go
```diff
@@ -139,6 +140,22 @@ func (l *LogHandler) Close() error {
...
+	if err = l.Close(); err != nil {
+		return fmt.Errorf("unable to close file: %s", err)
```
timoreimann: What does the error returned by `l.Close()` look like? Often, stdlib's file-related error messages already contain the file name and the problem, so we might not need the "unable to close file" part (and thereby possibly avoid stuttering).

If we need it, please mention the filename in the error message.

middlewares/accesslog/logger.go
```diff
@@ -139,6 +140,22 @@ func (l *LogHandler) Close() error {
...
+	}
+
+	l.file, err = os.OpenFile(l.filePath, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0664)
+	if err != nil {
+		return fmt.Errorf("unable to open file: %s, %s", l.filePath, err)
```
timoreimann: Same question as above if we need it.

If we do, please rephrase to

```
"unable to open file %s: %s"
```

server/server_signals_windows.go
```diff
@@ -0,0 +1,26 @@
...
+func (server *Server) configureSignals() {
+	signal.Notify(server.signals, syscall.SIGINT, syscall.SIGTERM)
+}
+
+func (server *Server) listenSignals() {
+	for {
+		sig := <-server.signals
+		switch sig {
+		default:
+			log.Infof("I have to go... %+v", sig)
```
timoreimann: Can we move the block into a central function and call it from both `listenSignals` implementations?

server/server_signals.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"os/signal"
+	"syscall"
+
+	"github.com/containous/traefik/log"
+)
+
+func (server *Server) configureSignals() {
+	signal.Notify(server.signals, syscall.SIGINT, syscall.SIGTERM, syscall.SIGUSR1)
```
timoreimann: What happens if we try to register `SIGUSR1` on Windows? Will the machine burn up in flames, or may the signal rather just get transmitted?

In case of the latter, could we maybe drop `server_signals_windows.go`?

rjshep: `SIGUSR1` isn't present in the Windows build of Go, so the code won't compile.  That was the reason for having `server_signals_windows.go` in the first place.

timoreimann: 👍

ldez:
@rjshep could fix the remaining comments of @timoreimann?

integration/log_rotation_test.go
rjshep: Yes - added

middlewares/accesslog/logger.go
rjshep: I couldn't get that to work.

server/server_signals_windows.go
rjshep: We can't move the block because `syscall.SIGUSR1` isn't defined on Windows.

middlewares/accesslog/logger.go
rjshep: You're right, changed to just return the error.

emilevauge:
@timoreimann any feedback on this one ?

PR #1768, opened 2017-06-18
Add HTTP HEAD handling to /ping endpoint

### LINKED ISSUE #1767: Traefik's /ping endpoint should answer to HTTP HEAD
It seems common for upstream load balancers/proxies, particularly in the cloud vendor space, to make `HTTP HEAD` requests when performing ping checks on their backends.

Traefik's `/ping` endpoint currently does not answer to anything other than `HTTP GET` which means that endpoint can't be used for pings in a scenario where Traefik is sitting downstream from e.g. a GCE network loadbalancer.

martinbaillie:
Resolves #1767

### DIFF
--- a/server/web.go
+++ b/server/web.go
@@ -83,7 +83,7 @@ func (provider *WebProvider) Provide(configurationChan chan<- types.ConfigMessag
 	systemRouter.Methods("GET").Path(provider.Path + "health").HandlerFunc(provider.getHealthHandler)
 
 	// ping route
-	systemRouter.Methods("GET").Path(provider.Path + "ping").HandlerFunc(provider.getPingHandler)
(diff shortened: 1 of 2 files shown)


server/web.go
```diff
@@ -84,6 +84,7 @@ func (provider *WebProvider) Provide(configurationChan chan<- types.ConfigMessag
 
 	// ping route
 	systemRouter.Methods("GET").Path(provider.Path + "ping").HandlerFunc(provider.getPingHandler)
+	systemRouter.Methods("HEAD").Path(provider.Path + "ping").HandlerFunc(provider.getPingHandler)
```
emilevauge: Could you instead use `systemRouter.Methods("GET", "HEAD").Path(provider.Path + "ping").HandlerFunc(provider.getPingHandler)` ?

martinbaillie: Sure @emilevauge. Done.

martinbaillie:
@timoreimann fixed

PR #1260, opened 2017-03-08
Added IgnoreFilter for ACME domain cert generation

CyrilPeponnet:
This Pull Request fixes the following:
- Avoid to generate ACME certs if a wildcard cert is present
- Always return the ACME cert if present and fallback to wildcard if present

And add the a new feature `ignoreFilters` to filter out domains for ACME certificate generation.

The static declaration of domains in KV/TOML will override the filters.

Test cases:

**case 1**: _I have wildcard certificate for `domain.tld` and I want acme to generate acme certs for only a subset of it._
 Use the `ignoreFilters` with a regular expression like `.*\.?[^f][^o][^o]\.domain\.tld` (negative lookahead are not supported by golang), to generate acme cert for domains except those like `*.foo.domain.tld`.

**case 2**: _I have wildcard for `domain.tld` but I want acme to take precedence over it for statically defined domains._ 
Ignore the entire domain with a filter like `.*\.domain\.tld`, and define your domains in ACME configuration (don't forget to restart traefik as this part is not reloaded dynamicaly).

**case 3**: _I have wildcard certificate for `domain.tld` and I want it to be used instead of already defined acme certs._ 
You will need to revoke and clean the acme cert from your configuration (if from KV, you will need to base64decode it, strip the cert part you want to remove, reencode, update it and restart traefik).

Let me know if I forgot something.

 Fixes #1197

PR #1772, opened 2017-06-19
Add provided certificate checking before LE certificate generation with OnHostRule option

nmengin:
During the code review of PR #1260, it appears that the provided certificates are not checked before LE certificate checking with the `OnHostRule` option.

Adding this check allows to generate a LE certificate only if domains have no provided certificates.

Thanks to it, PR #1260 will be modified to apply filter on domains to check by provided certificate in place of domains to ignore during LE certificate generation.

cc/ @CyrilPeponnet


nmengin:
@emilevauge I am agree with you for the tests.
We discussed about it with @ldez .

We believe that it can be nice to upgrade the boulder image and create ACME tests into a new PR.
I'm going to propose a new PR about this subject ASAP.

integration/acme_test.go
```diff
@@ -4,12 +4,14 @@ import (
 	"crypto/tls"
 	"net/http"
 	"os"
-	"os/exec"
 	"time"
 
 	"github.com/go-check/check"
 
 	"errors"
+
```
ldez: could you reformat imports?

acme/acme_test.go
```diff
@@ -9,6 +9,9 @@ import (
 	"testing"
 	"time"
 
+	"crypto/tls"
```
ldez: could you reformat imports?

integration/acme_test.go
```diff
@@ -3,18 +3,22 @@ package main
...
+	"github.com/containous/traefik/integration/utils"
+	"github.com/go-check/check"
+	checker "github.com/vdemeester/shakers"
 )
 
 // ACME test suites (using libcompose)
 type AcmeSuite struct {
 	BaseSuite
+	onDemand            bool
```
ldez: I think it's better to create a dedicated struct.

nmengin: The others tests use one struct for BaseSuite and all the arguments they need.
I used only one to be homogeneous, but I can changed it if it's really necessary.

ldez: they provide data shared between test (from the setup), not for tests data.

PR #1777, opened 2017-06-21
Change to actual weight

### LINKED ISSUE #1294: Backend server with weight "0"
1.2.0-rc2

Zookeeper provider

### What did you do?
Set backend server with weight "0" and another two others with weight "100"

Server with weight "0" could not receive requests.

### What did you see instead?
For each 200 requests (2 servers with weight 100) 1 is going to server with weight "0"

When I set weight "1" instead "0", I have the same behavior: 200 requests and 1 to server with weight 1. I think that setting "0" or "1", traefik load as the same value.

tanyadegurechaff:
### Description

Oxy changes weights from 0 to 1.

This PR prevents the PR that fixes #1294 from changing the default weight of backends.

### DIFF
--- a/provider/docker/docker.go
+++ b/provider/docker/docker.go
@@ -633,7 +633,7 @@ func (p *Provider) getWeight(container dockerData) string {
 	if label, err := getLabel(container, "traefik.weight"); err == nil {
 		return label
 	}
-	return "0"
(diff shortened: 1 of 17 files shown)


ldez:
I have report your commit to #1780.

PR #1780, opened 2017-06-23
Change default server weight

ldez:
### Description

Change default server weight.

Update Oxy:
- fix: Set default weight before calling Weight.

Must be merge after #1782.

Fixes: #1294

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -65,11 +65,11 @@ func (s *Server) UnmarshalTOML(data interface{}) error {
 	}
 
 	if w, ok := d["weight"]; ok {
-		v, ok := w.(int64)
+		v, ok := w.(int)
 		if !ok {
-			return errors.New("toml: server weight must be int64")
+			return errors.New("toml: server weight must be int")
 		}
-		s.Weight = int(v)
+		s.Weight = v
 	} else {
 		s.Weight = 1
 	}
--- a/types/types.go
+++ b/types/types.go
@@ -49,6 +49,34 @@ type Server struct {
 	Weight int    `json:"weight"`
 }
 
+// UnmarshalTOML define how unmarshal in TOML parsing
+func (s *Server) UnmarshalTOML(data interface{}) error {
+	d, ok := data.(map[string]interface{})
+	if !ok {
+		return errors.New("Server UnmarshalTOML want (map[string]interface{})")
+	}
+
+	if u, ok := d["url"]; ok {
+		v, ok := u.(string)
+		if !ok {
+			return errors.New("toml: server url must be string")
+		}
+		s.URL = v
+	}
+
+	if w, ok := d["weight"]; ok {
+		v, ok := w.(int64)
+		if !ok {
+			return errors.New("toml: server weight must be int64")
+		}
+		s.Weight = int(v)
+	} else {
+		s.Weight = 1
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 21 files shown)


emilevauge:
@containous/traefik this PR is contains potential breaking changes, and we need you help to certify that it is backward compatible. Thanks a lot for your help!

types/types_test.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, test := range testCases {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			server := Server{}
+			err := server.UnmarshalTOML(test.data)
+
+			if err != nil && err.Error() != test.expectedErrorMsg {
+				t.Fatalf("UnmarshalTOML(%#v): unexepected error message, got %q, want %s.", test.data, test.expectedErrorMsg, err)
```
timoreimann: The order of the last two parameters must be swapped.

timoreimann: Can we improve the message like

`unexpected error or error message: got error %q, want error message %s`

?

timoreimann: What's the benefit of using `%q` over `%s`?

ldez: I don't know: it's not my code.

vdemeester: @timoreimann it escapes stuff doesn't it (and put `"` around)

ldez:
> You also need to change the default weight in the kv template file.
It's already done.

types/types_test.go
timoreimann: @vdemeester Tbh, I can't make full sense out of it even after reading the docs and the example. Not sure how it applies here.

@tanyadegurechaff could you chime in? :)

timoreimann: Ah, only saw now that @ldez changed it already.

Anyways, please do speak out if you think we should better use `%q`. :-)

types/types.go
```diff
@@ -49,6 +49,34 @@ type Server struct {
...
+		}
+		s.URL = v
+	}
+
+	if w, ok := d["weight"]; ok {
+		v, ok := w.(int64)
+		if !ok {
+			return errors.New("toml: server weight must be int64")
+		}
+		s.Weight = int(v)
```
tanyadegurechaff: Because toml allows 64 bit range.

types/types_test.go
```diff
@@ -0,0 +1,89 @@
...
+			name: "toml server with weight 0",
+			data: map[string]interface{}{
+				"weight": int64(0),
+			},
+			expectedServer: Server{
+				Weight: 0,
+			},
+		},
+		{
+			name: "toml server with URL and weight 0",
```
timoreimann: @ldez WDYT on this one?

timoreimann:
What's the state on this one?

ldez:
@timoreimann I have a problem with the DynamoDB provider.
This supplier has no template, it is a very weird provider.

ldez:
test

jgrasl:
Hi @ldez @timoreimann what's the status on this one? It would be really nice to have this for the backend drain functionality. Can this be in the 1.4 release or it has to go to the next one? Thanks in advance.

ldez:
For now, I close this PR.
I will reopen later.

ldez:
Due to the break changes in Dynamo and Oxy, I move this feature after 1.5.

Issue #541, opened 2016-07-19
Redirect http to https on a per container (per frontend) basis


morvans:
It would be great if the ability to automatically redirect http to https could be enabled on a per-container basis using a label.

AFAICT, I don't think there's an issue and/or PR open for that.

migueleliasweb:
I was search exactly for that. Didn't find that too !

migueleliasweb:
_Context for the answer:_ 

- In the company I work in we use Marathon as the Traefik provider
- We are migrating all web apps to be "reversed proxied" by Traefik
- All our web apps use only http or https. Not both at the same time.

I had no luck finding how to make this work. (I think it's not implemented yet, at least for the Marathon provider)


This is what I tried: 
```toml
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
      entryPoint = "https"
  [entryPoints.https]
  address = ":443"
```

If I configure Traefik like this, all my http traffic gets redirected to https and that I cannot allow because behind my reverse proxy there's multiple domains and some does not have a SSL linked to it.

Some apps use only HTTPS but we could not let Traefik respond a 404 if the user tries to access via HTTP because of some legacy integrations. Currently our nginx servers redirect all HTTP requests to HTTPS.

What  I was forced to do to make it work was creating a simple redirect app on Marathon that reads and environment variable and if the requested host meets the criteria, the nginx returns a location header for the https url.

It's a workaround but until the team implements a per-frontend redirect there's no way to make this work directly from Traefik.... =/

emilevauge:
You can use Marathon labels to connect apps to a specific entrypoint, http or https.

migueleliasweb:
Hey @emilevauge I already do this but sadly when I do so the other protocol responds 404 and this breaks some older integrations that still references the old http routes...

Maybe I was not clear about what I did.

The main app (the one that runs the app server) has a "traefik.frontend.entryPoints=https" Label on Marathon, so it only recieves HTTPS requests but I have another app (the nginx redirector) that I can configure through env vars to respond a location from HTTP requests to HTTPS.

If I didn't have the redirector, my app would repond a 404 on HTTPS. That's the caveat.

emilevauge:
@morvans 
>And is there any trick like this for Docker backend ?

Yep: http://docs.traefik.io/toml/#docker-backend

emilevauge:
@migueleliasweb, I don't get what you are trying to do exactly here ^^
If you have:
```
[...]

PR #1788, opened 2017-06-27
Update basics.md to demonstrate a solution for optional redirection of HTTP to HTTPS

drakmirv:
Updated redirection section to show an example of having HTTP traffic conditionally redirect based on which entry point it is hooked into.  This was non-intuitive and numerous others were looking for a solution (see https://github.com/containous/traefik/issues/541).

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -41,6 +41,10 @@ Here is an example of entrypoints definition:
   address = ":80"
     [entryPoints.http.redirect]
     entryPoint = "https"
+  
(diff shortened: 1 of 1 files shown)


drakmirv:
This was done to show a possible solution for https://github.com/containous/traefik/issues/541

drakmirv:
This appears to not have worked.  For some reason it was working fine yesterday, but it must have been a fluke.  Today I get a bind error (which is honestly what I had been expecting before).

Issue #1465, opened 2017-04-19
Implement Auth by frontend for other providers


SantoDE:
Implement the "Auth by Frontend" Configuration into the others provider.

References: #1147 

- [x] Docker
- [x] Rancher
- [x] Marathon
- [x] Kubernetes
- [x] ECS
- [ ] Eureka
- [x] Consul catalog
- [x] Mesos
- [x] ZooKeeper

timoreimann:
Feature is available on Marathon. Updated the issue description to better see what we've got and what's yet missing.

joanayma:
Why isn't file provider listed?

ldez:
@joanayma because it's the first implementation.

> for other providers

joanayma:
@ldez Ok, thanks. I thought it. But I was trying to implement [frontend-based auth basic] and did not find any docs.

ldez:
@joanayma
```toml
[frontends]
 [frontends.frontend1]
 backend = "backend1"
   [frontends.frontend1]
    basicAuth = ["user1:pass1", "user2:pass2", "user3:pass3"]
```

joanayma:
Thanks a lot!

El dia 29 set. 2017 20:56, "Ludovic Fernandez" <notifications@github.com>
va escriure:

@joanayma <https://github.com/joanayma>

[frontends]
 [frontends.frontend1]
 backend = "backend1"
   [frontends.frontend1]
    basicAuth = ["user1:pass1", "user2:pass2", "user3:pass3"]

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/containous/traefik/issues/1465#issuecomment-333209893>,
or mute the thread

joanayma:
A more complex example for specific rule on example.org:
```toml
[frontends]
  [frontends.example1]
    rule = "Host: example.org, www.example.org"
    priority = 10
    passHostHeader = true
    basicAuth = ["user1:$basicauth", "user2:$basicauth"]
    backend = "backend1"
  [frontends.default]
    priority = 1
    passHostHeader = true
    backend = "backend1"
```

Maybe will be good to add it to documentation.

redge76:
Yes please add this to the documentation .... 
It took wayyyyyy too long to find this :-)

ldez:
Now all providers have all options 🎉

PR #1799, opened 2017-06-28
Add marathon label to configure basic auth

nikore:
### Description
Adding in support for configuring basic auth via a marathon label; there are similar functions in docker and the rancher providers.  

This addresses the marathon part of #1465

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -607,7 +607,6 @@ If you want Træfik to watch file changes automatically, just add:
 watch = true
 ```
 
-
 ## API backend
 
 Træfik can be configured using a RESTful api.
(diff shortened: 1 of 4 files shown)


provider/marathon/marathon_test.go
```diff
@@ -1801,3 +1807,33 @@ func TestParseIndex(t *testing.T) {
 		})
 	}
 }
+
+func TestMarathonGetBasicAuth(t *testing.T) {
+	provider := &Provider{}
+
+	applications := []struct {
```
timoreimann: Could you rename `applications` to `cases`?

(I realize you likely just copied this from existing tests; which unfortunately did not pick a better name either.)

provider/marathon/marathon_test.go
```diff
@@ -1801,3 +1807,33 @@ func TestParseIndex(t *testing.T) {
...
+			application: marathon.Application{
+				Labels: &map[string]string{
+					"traefik.frontend.auth.basic": "user:password",
+				},
+			},
+			expected: []string{"user:password"},
+		},
+	}
+
+	for _, a := range applications {
```
timoreimann: Could you please turn this into a group of parallelized sub-tests given proper names? See some of the other Marathon tests where we already employ those.

provider/marathon/marathon_test.go
```diff
@@ -1801,3 +1807,33 @@ func TestParseIndex(t *testing.T) {
...
+		actual := provider.getBasicAuth(a.application)
+		if !reflect.DeepEqual(actual, a.expected) {
+			t.Fatalf("expected %q, got %q", a.expected, actual)
```
timoreimann: Please revert the `expected, actual` order to match the check order one line above, and also to be in line with the [preferred order in Go](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures).

timoreimann: Once you've converted the tests into sub-tests, please don't forget to use `t.Errorf` instead of `t.Fatalf`.

docs/toml.md
```diff
@@ -1022,6 +1022,7 @@ Labels can be used on containers to override default behaviour:
 - `traefik.frontend.passHostHeader=true`: forward client `Host` header to the backend.
 - `traefik.frontend.priority=10`: override default frontend priority
 - `traefik.frontend.entryPoints=http,https`: assign this frontend to entry points `http` and `https`. Overrides `defaultEntryPoints`.
+- `traefik.frontend.auth.basic=test:$apr1$H6uskkkW$IgXLP6ewTrSuBkTrqE8wj/,test2:$apr1$d9hr9HBB$4HxwgUir3HP4EsggP/QNo0`: Sets a Basic Auth for that frontend with the users test:test and test2:test2
```
timoreimann: grammar nit-picks:

1. "Sets a Basic Auth" -> "Sets basic authentication"
1. "the users test:test and test2:test2" -> "the usernames and passwords test:test and test2:test2, respectively"

provider/marathon/marathon_test.go
```diff
@@ -99,6 +99,7 @@ func TestMarathonLoadConfig(t *testing.T) {
 				`frontend-test`: {
 					Backend:        "backend-test",
 					PassHostHeader: true,
+					BasicAuth:      []string{},
```
timoreimann: Do the unaffected tests fail if we do not add empty `BasicAuth` slices?

We should not extend them unless it's required.

nikore: Yes its required; all of the tests failed given how they just blind compare the structs.

docs/toml.md
nikore: If we want to grammar nit-pick then this line should be changed as well since I directly copy pasted it to here https://github.com/containous/traefik/blame/master/docs/toml.md#L915

provider/marathon/marathon_test.go
nikore: I would argue to keep the the style the same an in a different PR correct them all at once. Would be confusing to any one else giving a contribution why there are two styles

provider/marathon/marathon_test.go
nikore: Give me a direct example and I can try; this is literally the first goland code (besides basic hello world stuff) I have written; so this is mostly gibberish to me.

provider/marathon/marathon_test.go
nikore: Then it should also be fixed here https://github.com/containous/traefik/blob/master/provider/marathon/marathon_test.go#L1371

provider/marathon/marathon_test.go
timoreimann: Given that we run two or three different test styles in `marathon_test.go`, things must already be pretty confusing. At this point, I'm only interested in reducing the amount of alignment effort new code brings along until I get my cleanup PR through.

We'll most likely move to `cases` everywhere in the not-too-remote future, so it's okay to adjust accordingly in your PR.

provider/marathon/marathon_test.go
timoreimann: Sure thing: [Here you go](https://github.com/timoreimann/traefik/blob/bdad3d2f281126d58206fcba86c2862133850881/provider/marathon/marathon_test.go#L840-L849) (from an in-flight PR of mine).

Key points:

1. Capture the loop variable like `c := c`.
1. Pass an expressive description/test name as first parameter to `t.Run()`.
1. Call `t.Parallel()` first thing inside the body of `t.Run()`.

If you're interested in a bit of background, I suggest [this blog post](https://blog.golang.org/subtests).

Let me know if you have further questions!

provider/marathon/marathon_test.go
timoreimann: Absolutely! Would you mind taking care of it?

timoreimann:
Note to myself: Answering to review comments by another review produces duplicate comments. :/

PR #1800, opened 2017-06-28
Upgrade go-marathon to dd6cbd4.

### LINKED ISSUE #1743: Marathon provider on 1.3?
### What did you do?

Upgrade traefik from 1.2 to 1.3, when I try to start the app it says that it contains invalid responses from marathon 1.4.2. and the marathon provider do not work.

Marathon provider working

```
Version:      v1.3.0
Codename:     raclette
Go version:   go1.8.3
```

I run traefik inside docker(alpine) against a DC/OS 1.9 cluster without any authentication. Apart from marathon provider I use ```file``` provider  (it works).

```toml
endpoint = "http://172.16.8.161:8080,http://172.16.8.162:8080,http://172.16.8.163:8080"
watch = true
domain = "ikcloud.xx.es"
```

```shell
time="2017-06-13T14:38:39Z" level=error msg="Failed to create a client for marathon, error: invalid response from Marathon"
time="2017-06-13T14:38:39Z" level=debug msg="Configuration received from provider marathon: null"
```

Any hints on this?

timoreimann:
Fixes a problem with `UnreachableStrategy` being available now in two type-incompatible formats (object and string).

Fixes #1743.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -93,7 +93,7 @@ import:
 - package: k8s.io/client-go
   version: v2.0.0
 - package: github.com/gambol99/go-marathon
-  version: d672c6fbb499596869d95146a26e7d0746c06c54
+  version: dd6cbd4c2d71294a19fb89158f2a00d427f174ab
 - package: github.com/ArthurHlt/go-eureka-client
   subpackages:
   - eureka
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 6e206389bc4f381387be8b5f0cc5a41224329752654ddb3c2a805adde0333217
-updated: 2017-06-17T14:30:19.890844996+02:00
+hash: 9a62fe4058ef43ed185d96638322dd3d44ae21f7f65fae14c4c6d404876c2872
+updated: 2017-06-28T15:47:14.848940186+02:00
 imports:
 - name: cloud.google.com/go
   version: 2e6a95edb1071d750f6d7db777bf66cd2997af6c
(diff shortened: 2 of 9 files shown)


timoreimann:
Missed to update one transitive dep. Doing that now...

ldez:
https://github.com/gambol99/go-marathon/compare/15ea23e360abb8b25071e677aed344f31838e403...dd6cbd4c2d71294a19fb89158f2a00d427f174ab

https://github.com/donovanhide/eventsource/compare/fd1de70867126402be23c306e1ce32828455d85b...441a03aa37b3329bbb79f43de81914ea18724718

glide.yaml
```diff
@@ -93,7 +93,6 @@ import:
 - package: k8s.io/client-go
   version: v2.0.0
 - package: github.com/gambol99/go-marathon
-  version: d672c6fbb499596869d95146a26e7d0746c06c54
```
ldez: please add the hash instead remove it

ldez: if you don't do that you produce side effect when other dependencies are added.

timoreimann: Not sure I understand... why do we need the hash with go-marathon when we don't specify one with the vast majority of other dependencies?

Semantically, we shouldn't have to add it as we are not after a specific version: When we upgrade go-marathon, we tend to look for to latest.

ldez: It's hard to explain, sorry if I will misspoke.

Currently when you add a dependency, you don't need to pin version to yml (as we did before)

Glide do some "random" update when some version are not fixed.
I don't know exactly why, but I know Glide love the past and not the present...
Some dependencies don't update when you add a dependency but not all...

To be "safe", some dependencies must be fixed. (IMHO in the yml all dependencies must be fixed on version or hash)

I know this is not prefect but I don't want to do again a +8 hours of hash analysis.

timoreimann: Sure, understood. FWIW, a number of glide problems can be resolved by clearing the cache up-front (i.e., running `glide cc`); but I know from experience that it doesn't always help.

I updated the YML file and hash. Hope it's okay now.

PR #1810, opened 2017-06-30
remove confusing go-marathon log message

marco-jantke:
Log message produced by go-marathon was:

`time="2017-06-28T09:08:19Z" level=debug msg="listenToSSE(): failed to
handle event: failed to decode the event type, content: , error: EOF"`

The fix for this was done in the upstream project of go-marathon donovanhide/eventsource.

Background is that Marathon periodically sends a `\n` over the SSE subscription, in order to keep the connection alive. This was parsed as empty event by the eventsource and published. go-marathon in turn was not able to do something with this empty event was producing the log message above. By getting rid of publishing empty events in the downstream library, we also get rid of this log message.


ldez:
https://github.com/donovanhide/eventsource/compare/441a03aa37b3329bbb79f43de81914ea18724718...b8f31a59085e69dd2678cf51840db2ac625cb741

Issue #1559, opened 2017-05-05
Support Marathon readiness checks


timoreimann:
[Readiness checks](https://mesosphere.github.io/marathon/docs/readiness-checks.html) are a means to prevent Marathon tasks from being taken into LB rotation prior to them being ready to serve traffic. The idea is that applications define a readiness check HTTP endpoint along with a number of parameters in the Marathon specification, which will cause Marathon to only mark a task as ready during the deployment of the associated application if that endpoint returns a successful response.

Traefik already vendors a copy of the go-marathon that brings [readiness check support](https://github.com/gambol99/go-marathon/pull/273). The only thing left to do in Traefik is to filter out tasks whose readiness check result is not positive.

Refs #1185.

timoreimann:
Internally, I have already implemented the necessary logic for [Bamboo](https://github.com/QubitProducts/bamboo). It should be fairly trivial to port it to Traefik, which I'm going to do.

PR #1815, opened 2017-07-02
Use single API call to fetch Marathon resources.

### LINKED ISSUE #1846: Marathon API calls are inefficient
Traefik does the following:
Request all applications
Request all tasks
Filter applications returned
Filter tasks returned

Marathon's REST API allows all of this in one call
GET /v2/apps?embed=apps.tasks&label=traefik.tags==mumble

This also makes the marathon server workload smaller, as it doesn't have to return a large body of data to each traefik server.

timoreimann:
Change Marathon provider to make just one API call instead of two per configuration update by means of specifying embedded resources, which enable retrieving multiple response types from the API at once. Apart from the obvious savings in API calls, we primarily gain a consistent view on both applications and tasks that allows us to drop a lot of correlation logic.  Additionally, it will serve as the basis for the introduction of [readiness checks](https://github.com/containous/traefik/issues/1559) which require application/task consistency for correct leverage on the proxy end.

Additional changes:

marathon.go:
- Filter on tasks now embedded inside the applications.
- Reduce/simplify signature on multiple template functions as we do not need to check for proper application/task correlation anymore.
- Remove getFrontendBackend in favor of just getBackend.
- Move filtering on enabled/exposed applications from `taskFilter` to `applicationFilter`. (The task filter just reached out to the applications anyway, so it never made sense to locate it with the
  tasks where the filter was called once for every task even though the result would never change.)
- Remove duplicate constraints filter in tasks, where it neither made sense to keep as it operates on the application level only.
- Add context to rendering error.

marathon_test.go:
- Simplify and reduce numerous tests.
- Convert tests with high number of cases into parallelized sub-tests.
- Improve readability/structure for several tests.
- Add missing test for enabled/exposed applications.
- Simplify the mocked Marathon server.

marathon.tmpl:
- Update application/task iteration.
- Replace `getFrontendBackend` by `getBackend`.

Fixes #1846

### DIFF
--- a/provider/marathon/marathon_test.go
+++ b/provider/marathon/marathon_test.go
@@ -9,8 +9,8 @@ import (
 	"github.com/containous/traefik/provider/marathon/mocks"
 	"github.com/containous/traefik/testhelpers"
 	"github.com/containous/traefik/types"
-	"github.com/davecgh/go-spew/spew"
 	"github.com/gambol99/go-marathon"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
 )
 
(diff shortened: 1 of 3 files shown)


timoreimann:
Note that I have explicitly tried to keep the changes to tests as close to the bare minimum as possible to reflect the changes in the production code. Specifically, the Marathon tests badly need a builder pattern to reduce the amount of boilerplate.

I'll tackle that in a follow-up PR in order to keep this PR simple-ish.

emilevauge:
Great work @timoreimann, design LGTM :clap:

provider/marathon/marathon_test.go
```diff
@@ -9,107 +9,78 @@ import (
...
-					},
-				},
+			desc: "simple application",
+			application: marathon.Application{
+				Ports:  []int{80},
+				Labels: &map[string]string{},
 			},
-			tasks: &marathon.Tasks{
-				Tasks: []marathon.Task{
+			task: marathon.Task{
```
ldez: Could you reduce test definition with some helpers builder?

```golang
// ...
    task: aMarathonTask(withDefaultIPAddress),
// ...
```
```golang

func aMarathonTask(builders ...func(marathon.Task)) marathon.Task {
	task := marathon.Task{
		Host:  "localhost",
		Ports: []int{80},
	}

	for _, builder := range builders {
		builder(task)
	}

	return task
}

func withDefaultIPAddress(task marathon.Task) {
	task.IPAddresses =  []*marathon.IPAddress{
		{
			IPAddress: "127.0.0.1",
			Protocol:  "tcp",
		},
	}
}
```
I think you do the same for `marathon.Application`, `types.Backend`, `types.Frontend`.

emilevauge:
@ldez Marathon integration tests will be done https://github.com/containous/traefik/pull/1406 ;)

provider/marathon/marathon_test.go
timoreimann: [As noted](https://github.com/containous/traefik/pull/1815#issuecomment-312513938), I intentionally kept modifications to the tests as minimal as possible in order to not risk changing both production code and the our test coverage safety net at the same time.

I'd like to do the test improvements in a follow-up PR. In fact, I'm already working on that and have started building a builder very similar to what you outlined.

Okay for you?

timoreimann:
Having at least some minimal integration test would actually be nice.

I'll wipe off the dust of #1406 and see if I can put something basic together. This PR doesn't change functionality, so ideally #1406 should provide immediate benefit.

provider/marathon/marathon_test.go
ldez: I'm not really like to have 2 PRs for the same subject when 2-3 commits make the job: you can create a temporary test file, with the new tests, ensure all old tests always pass and replace old tests file by the temporary file. (with commit [and push to run CI] at each step)

But tests passed and code build.

Forget my request.

ldez:
@emilevauge I know, it's just a joke because @timoreimann and me have already speaking about Marathon integration tests. 😃

ldez:
@timoreimann Could you rebase and resolve conflicts?

timoreimann:
Rebased.

Juliens:
@timoreimann Could you resolve conflicts? Thx

timoreimann:
@Juliens Rebased.

PR #1817, opened 2017-07-03
Tell glog to log everything into STDERR.

### LINKED ISSUE #1785: Traefik in from-scratch Docker container exits with exit code 0 if RBAC permissions on Kubernetes are insufficient
Bug.

### What did you do?
Start Traefik on Kubernetes with RBAC permissions set to allow access to Ingresses, Services, and Endpoints. Specifically, I did not permit access to Secrets.

Traefik humming along nice and smooth.

### What did you see instead?
Traefik exits with error code 2 when it cannot access Secrets.

latest master.

Relevant logs:

```
E0625 21:28:59.568275       1 reflector.go:300] github.com/containous/traefik/vendor/k8s.io/client-go/tools/cache/reflector.go:94: Failed to watch *v1.Secret: the server does not allow access to the requested resource (get secrets)
log: exiting because of error: log: cannot create log: open /tmp/traefik.traefik-ingress-controller-wgt5n.unknownuser.log.ERROR.20170625-212859.1: no such file or directory
```

The last line is from glog and responsible for the actual premature program termination.

The issue is happening because of two events happening simultaneously:

- Traefik requires permissions to Secrets even when basic auth isn't used (#1784).
- client-go configures glog to log such access failures to a file but the required `/tmp` directory isn't available on a `FROM scratch` Docker container, leading to the eventual failure.

The proper fix is likely to make client-go not log to a file by default. In the meantime though, we can mitigate the problem by fixing #1784. In general, Traefik shouldn't stop running because of a permission problem with one provider.

For the Secrets case specifically, a workaround for now is to enable Secrets access on RBAC.

timoreimann:
Logging errors into a file inside a minimalistic container might not be possible, and glog bails out with an exit code > 0 if it fails.

Fixes #1785

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"errors"
+	"flag"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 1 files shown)


timoreimann:
Picked some random reviewers; @containous/traefik, feel free to review as desired. :-)

timoreimann:
@emilevauge sure thing 👍 done.

PR #1823, opened 2017-07-04
Update traefik dependencies (docker/docker and related)

vdemeester:
- Update dependencies
- Fix compilation problems
- Remove vdemeester/docker-events (in docker api now)
- Remove `integration/vendor`

hollyyyyy


emilevauge:
This PR also:

- defeats hunger in the world
- reverses climate change
- cancels wars

zeroual:
@vdemeester what are you doing 🤕 . you rewrite  the project 😂

SantoDE:
WE LOVE YOU @vdemeester

emilevauge:
Wow, what a PR @vdemeester :heart_eyes_cat: 
But I have few comments thought:
- could you split your glide.* commits from your vendor/* ones? it will be easier to review :)
- The main reason why @errm initially spitted our `glide.yaml` was to make it easier to differentiate traefik deps from tests deps. I'm concerned about this issue in your PR. Could you put all integration tests deps into `testImports` at least ?
- While creating the wonderful `script/glide.sh`, @timoreimann also added an _Interistingly weird_ © peace of code:
```bash
if ! is_integration_dir; then
  GLIDE_ARGS+=('--skip-test')
fi
```  
First, @timoreimann, why? Then, @vdemeester, you may remove this part if you add deps into `testImports` :)

ldez:
@vdemeester could you check the build with `make crossbinary-others`?

timoreimann:
@emilevauge: re: "why not stripping tests on the integration dependencies": To be honest, I can't remember the details anymore. All I know is that when I did the initial glide vendoring PR, I started off with stripping everything (nested vendor folders and tests on both main and integration dependencies) and ran into build failures. I then added pieces back in until I found a combination that worked.

Unless we drop integration dependencies completely, I think it's worthwhile checking if stripping on them works now. A lot has changed since my initial PR.

script/validate-vendor
```diff
@@ -6,32 +6,28 @@ set -o nounset
...
+    # We run glide install to and see if we have a diff afterwards
+    echo "checking ${vendor_dir} for unintentional changes..."
+    (
+    "${SCRIPTDIR}/glide.sh" install
+    )
+    # Let see if the working directory is clean
+    diffs="$(git status --porcelain -- ${vendor_dir} 2>/dev/null)"
+    if [[ "$diffs" ]]; then
+        {
+            echo "The result of 'glide install' for vendor directory '${dir}' differs"
```
timoreimann: The `dir` variable must now be called `vendor_dir`.

script/glide.sh
```diff
@@ -6,11 +6,6 @@ set -o nounset
 ####
 ### Helper script for glide[-vc] to handle specifics for the Traefik repo.
 ##
-# In particular, the 'integration/' directory contains its own set of
-# glide-managed dependencies which must not have its nested vendor folder
-# stripped. Depending on where the script is called from, it will do the Right
-# Thing.
-#
 
 CWD="$(pwd)"; readonly CWD
```
timoreimann: Given that the vendor folder location is now non-ambiguous, we could just `cd` into the right directory relative to the location of this script. That way, the script can be called from just anywhere.

ldez:
I have made some changes:
- Use `testImport`
- revert windows detected build step
- update some deps.

I have check all deps (from @vdemeester and from me) manually, one by one... Very hard work...

script/glide.sh
timoreimann: How about this one? ☺️

script/glide.sh
```diff
@@ -49,7 +36,7 @@ if ! type glide-vc > /dev/null 2>&1; then
   exit 1
 fi
 
-if [[ ! -e "${CWD}/glide.yaml" ]]; then
+if [[ ! -e "glide.yaml" ]]; then
```
timoreimann: This won't allow us to call the script from anywhere but the root directory.

I suggest to define

```bash
SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
readonly SCRIPT_DIR
readonly GLIDE_DIR="${SCRIPT_DIR}/.."
```

near the top and then use

```bash
if [[ ! -e "${GLIDE_DIR}/glide.yaml" ]]; then
```

on this line.

emilevauge: For the sake of simplicity, I would prefer to use this terrible trick only when needed ^^

vdemeester: 😂

ldez: @timoreimann I have added you "simple" trick 😉

emilevauge: @timoreimann I know bash can be a lot worse... But I find this argument quite misleading. I'm sorry but I would have preferred to keep our bash scripts as simple/small as possible. We all know that a complex shell script is not maintainable.

vdemeester:
🎉

PR #1837, opened 2017-07-06
Rename script folder.

ldez:
### Description

To  see better the separation between the code and the tools.


vdemeester:
@ldez why ? build script are part of the code 😰

errm:
I agree with vdemeester, I am not sure this is needed or sensible...

I have things setup in my shell and editor that dotfiles are invisible... this adds an un-needed level of friction to a newcomer to the project who might wan't to find out how the build works

emilevauge:
As this PR leads to debates, I suggest we close it.
The idea of @ldez was to put these build scripts somewhere else.
Using dotfiles for this may not be the best solution, but I still think it would be great to simplify the actual project tree:
```
acme
autogen
cluster
cmd
contrib
dist
docs
examples
healthcheck
integration
job
log
middlewares
provider
safe
script
server
static
templates
testhelpers
types
vendor
version
webui
```
Any idea is welcome!

timoreimann:
Instead of trying to hide/move every non-code folder, we could also just put the Go code below a single directory. That way, everything else in the top-level directory is, by definition, not directly related to the source code.

I believe that many projects keep everything below `/pkg`. I _may_ have sensed a vibe on Twitter for this being not advisable or typical for Go, though I don't have any hard facts supporting this concern.

PR #1841, opened 2017-07-08
extract lb configuration steps into method

marco-jantke:
This is a small PR that does a little code cleanup in regards to the load balancer configuration. It extracts the initial configuration into a method and therefore removes code duplication.

One thing to note is that before the healthcheck configuration was setup multiple times for the Drr LB. This was not a problem as the result was always written into the same map key, though unnecessary.

@timoreimann PTAL.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -735,24 +735,14 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 							rebalancer, _ = roundrobin.NewRebalancer(rr, roundrobin.RebalancerLogger(oxyLogger), roundrobin.RebalancerStickySession(sticky))
 						}
 						lb = rebalancer
-						for serverName, server := range configuration.Backends[frontend.Backend].Servers {
-							url, err := url.Parse(server.URL)
-							if err != nil {
-								log.Errorf("Error parsing server URL %s: %v", server.URL, err)
-								log.Errorf("Skipping frontend %s...", frontendName)
-								continue frontend
-							}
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -888,6 +868,22 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
...
 }
 
+func configureLBServers(lb healthcheck.LoadBalancer, config *types.Configuration, frontend *types.Frontend) error {
+	for serverName, server := range config.Backends[frontend.Backend].Servers {
+		url, err := url.Parse(server.URL)
+		if err != nil {
+			log.Errorf("Error parsing server URL %s: %v", server.URL, err)
+			return err
+		}
+		log.Debugf("Creating server %s at %s with weight %d", serverName, url.String(), server.Weight)
```
ldez: I know this is not your code but could you:
- rename `url` to `URL`
- remove the unnecessary `.String()` call.

marco-jantke: Sure, done.

timoreimann: >rename url to URL

This may give readers the false impression that `URL` is an exported identifier. Now you basically have to track down the scope of the variable and realize that it's defined on the stack. 🤔

marco-jantke: I think Timo has a good point here. Additionally I did not find any written conventions in the go documentation or in the formatting tools. Therefore I would also vote to keep it lowercase, to not confuse readers. WDYT @ldez?

ldez: I suggested that only to fix name conflict with a package.
You rename to `u`

marco-jantke: I renamed it now to `u`.

PR #1236, opened 2017-03-04
Create Header Middleware

dtomcej:
This is a middleware based on github.com/unrolled/secure that allows us to control, and inject request and response headers.

PR #1854, opened 2017-07-11
Allow to bypass https redirection per frontend

eliecharra:
# Description

This is an attempt to implement #541.
The goal is to allow a container to bypass http -> https redirection using a label `traefik.http.redirect`.
This is my first PR on this project (and in golang), let me know if I've made mistakes :wink: 
Dunno if I have to implement this new label for each provider or if you'd rather prefer to split the work in multiples PR's.

- [x] Fix test
- [x] Add test
- [x] Update doc
- [ ] Implement for every provider ?

## Example Usage
```
➜ docker run -it --rm --name whoami emilevauge/whoami                                      
Starting up on port 80

➜ curl -I --insecure http://whoami.localhost 
HTTP/1.1 302 Found
Location: https://whoami.localhost:443/
Date: Tue, 11 Jul 2017 14:41:29 GMT
Content-Length: 5
Content-Type: text/plain; charset=utf-8

➜ curl -I --insecure https://whoami.localhost
HTTP/1.1 200 OK
Content-Length: 256
Content-Type: text/plain; charset=utf-8
Date: Tue, 11 Jul 2017 14:41:45 GMT
```

```
➜ docker run -it --rm --name whoami --label "traefik.http.redirect=false" emilevauge/whoami                                                               
Starting up on port 80

➜ curl -I --insecure http://whoami.localhost 
HTTP/1.1 200 OK
Content-Length: 255
Content-Type: text/plain; charset=utf-8
Date: Tue, 11 Jul 2017 14:42:09 GMT

➜ curl -I --insecure https://whoami.localhost
HTTP/1.1 200 OK
Content-Length: 256
Content-Type: text/plain; charset=utf-8
Date: Tue, 11 Jul 2017 14:42:20 GMT
```

Assuming endpoint configuration is :

```toml
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    entryPoint = "https"
  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]
      CertFile = "www.crt"
      KeyFile = "www.key"
```

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -920,6 +920,7 @@ Labels can be used on containers to override default behaviour:
 - `traefik.frontend.auth.basic=test:$apr1$H6uskkkW$IgXLP6ewTrSuBkTrqE8wj/,test2:$apr1$d9hr9HBB$4HxwgUir3HP4EsggP/QNo0`: Sets basic authentication for that frontend with the usernames and passwords test:test and test2:test2, respectively
 - `traefik.frontend.whitelistSourceRange: "1.2.3.0/24, fe80::/16"`: List of IP-Ranges which are allowed to access. An unset or empty list allows all Source-IPs to access. If one of the Net-Specifications are invalid, the whole list is invalid and allows all Source-IPs to access.
 - `traefik.docker.network`: Set the docker network to use for connections to this container. If a container is linked to several networks, be sure to set the proper network name (you can check with docker inspect <container_id>) otherwise it will randomly pick one (depending on how docker is returning them). For instance when deploying docker `stack` from compose files, the compose defined networks will be prefixed with the `stack` name.
+- `traefik.http.redirect=false`: When used with endpoints setup in HTTP + HTTPS configuration, set this flag to false to prevent https redirection for a given frontend.
 
 If several ports need to be exposed from a container, the services labels can be used
 - `traefik.<service-name>.port=443`: create a service binding with frontend/backend using this port. Overrides `traefik.port`.
--- a/provider/docker/docker.go
+++ b/provider/docker/docker.go
@@ -261,6 +261,7 @@ func (p *Provider) loadDockerConfig(containersInspected []dockerData) *types.Con
 		"getProtocol":                 p.getProtocol,
 		"getPassHostHeader":           p.getPassHostHeader,
 		"getPriority":                 p.getPriority,
+		"getHTTPRedirect":             p.getHTTPRedirect,
 		"getEntryPoints":              p.getEntryPoints,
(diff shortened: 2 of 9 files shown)


dtomcej:
Currently this code was implemented in https://github.com/containous/traefik/pull/1236

That allows you to enable redirection per frontend. It has already been merged, and should be available for testing under the experimental tag!

eliecharra:
@dtomcej @ldez This is not a dupplicate for the following reasons.

Of course you can add `Location` header using new middleware, but here dont want request to reach backend application if the stream is not secure.

Using #1236 + add header `Location` allow the first request to reach application backend on an unsecure stream, and so, the response sent over unsecure stream too.
If backend application is returning a HTTP/200 response, your browser will not be able to handle redirection.

We need to upgrade http to https on frontend proxy and only secure communications with backend application. This PR is to allow some services to force http without being redirected to https using traefik default http + https redirection entrypoint config.

Header middleware canot be used to handle this for me.

Please read the following usecase : https://github.com/containous/traefik/issues/541#issuecomment-238748845

dtomcej:
@eliecharra I think you misunderstand.

That PR has 2 components:
1. The ability to add custom headers
2. It implements github.com/unrolled/secure to provide a secure middleware layer, that handles a bunch of common security functions.

in your case, the middleware SSLRedirect function does not pass the request on to the backend, it halts the middleware chain, and returns a redirect response.

Feel free to browse https://github.com/unrolled/secure, as it outlines the options that were added.

You also have the choice of returning a 301 or a 302 with your redirect, depending on if you want it to be cached or not!

ldez:
Duplicate of #1236 

https://github.com/containous/traefik/blob/master/docs/basics.md#security-headers

ssboisen:
@dtomcej We have a similar usecase as OP. I just want to clarify if it is indeed possible for a particular frontend to **SKIP** a default http->https redirect by using `SSLRedirect = false`. If that is indeed the case how does that work with kubernetes ingress?

dtomcej:
Traefik operates in 3 layers:

- Entrypoints
- Frontends
- Backends

If you force a blanket redirect on the entrypoints, traffic never reaches frontends.
To accomplish what you want, you need to _not_ put a default redirect on the entrypoint, but instead put them on the frontends.

Unfortunately, the core framework code implemented in 1236 only provides the code for the core toml configuration.

You can either open a ticket similar to #2028 but for kubernetes, or you can keep an eye out, because it is what I will be working on once 2028 is done.

Regardless, Since this ticket is closed, I will not be continuing discussion here :)

PR #1860, opened 2017-07-12
added modifications for parsing out && and || on front end rule

### LINKED ISSUE #1280: Use | , & instead of , and ; with frontend.rule
traefik version : 1.1.1
we use traefik with consul catalog as its backend. Registrator popluates data to the consul catalog. Tags are use for configuring front end rules for a service to be picked and configured by traefik. Consul allows storing multiple tags irrespective of the purpose for which they are used, we do use tags for tagging the service with specific metadata, traefik front end rules being one of them. Each of these tags are required to be separated by ',' which in turn get picked as separate tag entries on the catalog for the service.

With traefik having ',' as a separator for `||` conditions, results in the rules that acutally intened to be part of an or condition, get added a separate tag entires.

eg> `traefik.frontend.rule=Host:myservice.domain.com,myservicetwo.domain.com` 
should result in rules that perform an `||` condition on the presence of either of the two as Host header values.

since the ',' character is being seen as a separator charter to separate multiple tags, the rules get put in as separate tags rather than a single rule.

consul catalog

`myservice {
         "traefik.frontend.rule=Host:myservice.domain.com",
         "myservicetwo.domain.com"
}`
instead of
`myservice {
         "traefik.frontend.rule=Host:myservice.domain.com,myservicetwo.domain.com"
}`

it would be better if `||` is used instead of `,` and `&&` instead of `;`

sriyer:
The changes to have this implemented for the Consul Catalog are here.

Fixes #1280

### DIFF
--- a/provider/consul/consul_catalog.go
+++ b/provider/consul/consul_catalog.go
@@ -251,7 +251,7 @@ func (p *CatalogProvider) getBackend(node *api.ServiceEntry) string {
 
 func (p *CatalogProvider) getFrontendRule(service serviceUpdate) string {
 	customFrontendRule := p.getAttribute("frontend.rule", service.Attributes, "")
-	if customFrontendRule == "" || customFrontendRule == nil {
+	if customFrontendRule == "" {
 		customFrontendRule = p.FrontEndRule
 	} else {
 		customFrontendRule = translateRuleSeparators(customFrontendRule)
--- a/provider/consul/consul_catalog.go
+++ b/provider/consul/consul_catalog.go
@@ -251,8 +251,10 @@ func (p *CatalogProvider) getBackend(node *api.ServiceEntry) string {
 
 func (p *CatalogProvider) getFrontendRule(service serviceUpdate) string {
 	customFrontendRule := p.getAttribute("frontend.rule", service.Attributes, "")
-	if customFrontendRule == "" {
+	if customFrontendRule == "" || customFrontendRule == nil {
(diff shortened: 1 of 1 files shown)


ldez:
Thanks for the contribution 👍 

Could you fix [this problem](https://semaphoreci.com/containous/traefik/branches/pull-request-1860/builds/1):
```
These files are not properly gofmt'd:
 - provider/consul/consul_catalog.go
```

Please read [CONTRIBUTING.md](https://github.com/containous/traefik/blob/master/.github/CONTRIBUTING.md) and run the command `make` to ensure your code build on your computer.

sriyer:
The problem is now fixed.

emilevauge:
@sriyer thanks for your contribution. We think it would be better to both support actual separators and new separators to stay backward compatible. Besides, it would be even better to not be Consul specific and add this new separators in `rules.go` `parseRules`. I'm sorry to inform you that therefore I prefer closing this one as we can not merge it. Feel free to contact us if you want some help on this :)

sriyer:
@emilevauge i'll take a look at rules.go . btw we have this in production already for consul catalog with discovery source being registrator.

sriyer:
"We think it would be better to both support actual separators and new separators to stay backward compatible" - the change on this pull request supports the use of both. It only replaces || with , and && with ;

gtrias:
Is there any plan to merge this PR or any alternative workarround to achieve the same funcionality? 
I need this feature as well.

Issue #1866, opened 2017-07-17
Healthcheck RemoveServer does not work proper with multiple entrypoints


chrigl:
### Do you want to request a *feature* or report a *bug*?

Bug

Note: I will create two PRs with different ways to resolve this.

### What did you do?

Configured multiple entrypoints, and using a single backend with a healthcheck. There are two servers in a backend (config see below). I "broke" the healthcheck on one of the servers.

### What did you expect to see?

No matter which entrypoint I hit, I expect the "broken" server to be excluded.

### What did you see instead?

Actually. The server is only removed from the last configured entrypoint. Here `https`.  So hitting the `http`-Entrypoint always hits both servers.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.3.3
Codename:     raclette
Go version:   go1.8.3
Built:        2017-07-06_05:32:56PM
OS/Arch:      linux/amd64
```

But master is affected as well.

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
defaultEntryPoints = ["http", "https"]

[entryPoints]
  [entryPoints.http]
    address = "10.1.1.30:80"
  [entryPoints.https]
    address = ":443"
      [entryPoints.https.tls]

[frontends]
  [frontends.test1]
    backend = "test1"
    passHostHeader = true
    [frontends.test1.routes.default]
      rule = "Host:test1.example.com"

[backends]
  [backends.test1]
    [backends.test1.healthcheck]
      path = "/healthcheck"
    [backends.test1.Loadbalancer]
      method = "drr"
    [backends.test1.servers.controller1]
      url = "http://10.1.1.100:8774"
    [backends.test1.servers.controller2]
      url = "http://10.1.1.101:8774"
```

### If applicable, please paste the log output in debug mode (`--debug` switch)

```
(paste your output here)
```

ldez:
@chrigl Thanks for your interest in Traefik 😃 

We are always happy with contributions.
We are a small team and each PR take a lot of time to review.

Next time before open 2 PRs, could you just explains your solutions in this issue or come speak with us in the [Traefik community Slack channel](https://traefik.herokuapp.com) (#dev channel).

PR #1433, opened 2017-04-13
Fix too many redirect

emilevauge:
Fixes #1429 

/cc @argeiger

Signed-off-by: Emile Vauge <emile@vauge.com>

PR #1867, opened 2017-07-17
Reuse backend-LBs over entrypoints

chrigl:
### Description

Fixes: #1866
Conflicts: #1868

This approach reuses backend-LBs for multiple entrypoints. If e.g. there
are the entrypoints `http` and `https`, both use the same lb-instance.
So because of a healthcheck one server is removed from the backend, it
is removed from both entrypoints.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -663,7 +663,7 @@ func (server *Server) loadConfig(configurations configs, globalConfiguration Glo
 						}
 					}
 				}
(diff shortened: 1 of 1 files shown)


emilevauge:
@chrigl Thanks for your contribution!
This PR adds a regressions fixed by https://github.com/containous/traefik/pull/1433 few months ago.
We should instead prefer #1868.
Closing this one then :)

PR #1870, opened 2017-07-17
Server-Sent Events are broken when enabling the retry option

### LINKED ISSUE #1862: Server-Sent Events are broken when enabling the retry option
Hello,

Bug

### What did you do?
We expose [Dashing.io](http://dashing.io/) behind Traefik.
Dashing.io uses [Server-Sent Events](https://www.w3schools.com/html/html5_serversentevents.asp). 

If we let the default retry configuration (disabled), everything is working well.
If we enable the retry configuration (for example --retry.attempts=2), the event stream content type is replaced by Traefik by text/plain and the Browser throws an exception.

**Request example**
```
GET /events HTTP/1.1
Connection: keep-alive
Accept: text/event-stream
```

**Correct Response example without retry**
```
HTTP/1.1 200 OK
Content-Type: text/event-stream;charset=utf-8
Transfer-Encoding: chunked`
```
### What did you see instead?
**Incorrect Response example with --retry.attempts=2**
```
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Transfer-Encoding: chunked`
```

```
Version:      v1.3.3
Codename:     raclette
Go version:   go1.8.3
Built:        2017-07-06_05:32:56PM
OS/Arch:      linux/amd64
```

```json
{
 "GraceTimeOut": 10000000000,
 "Debug": false,
 "CheckNewVersion": true,
 "AccessLogsFile": "",
 "TraefikLogsFile": "",
 "LogLevel": "ERROR",
 "EntryPoints": {},
 "Cluster": null,
 "Constraints": [],
 "ACME": null,
 "DefaultEntryPoints": [],
 "ProvidersThrottleDuration": 2000000000,
 "MaxIdleConnsPerHost": 200,
 "IdleTimeout": 180000000000,
 "InsecureSkipVerify": false,
 "Retry": null,
 "HealthCheck": {
  "Interval": 30000000000
 },
 "Docker": null,
 "File": null,
 "Web": null,
 "Marathon": null,
 "Consul": null,
 "ConsulCatalog": null,
 "Etcd": null,
 "Zookeeper": null,
 "Boltdb": null,
 "Kubernetes": null,
 "Mesos": null,
 "Eureka": null,
 "ECS": null,
 "Rancher": null,
 "DynamoDB": null,
 "ConfigFile": ""
}
```

docker-compose example
```json
{
    "version": "3",
    "services": {
        "traefik": {
            "image": "nexus.cegedim-next.net:5001/traefik:1.3.3-alpine",
            "ports": [
                "80:80",
                "8080:8080"
            ],
            "volumes": [
                "/var/run/docker.sock:/var/run/docker.sock"
            ],
            "command": "--docker --docker.swarmmode --docker.watch --web --retry.attempts=2",
            "deploy": {
                "mode": "global"
            }
        },
        "dashing": {
            "image": "johnzan/dashing",
            "deploy": {
                "labels": {
                    "traefik.enable": "true",
                    "traefik.backend": "dashing",
                    "traefik.frontend.rule": "PathPrefixStrip:/dashing",
                    "traefik.port": "3030"
                }
            }
        }
    }
}
```
Thank you!

romlinch:
When backends streams http content and traefik is configured with retry middleware,
flush is called on response recorder of retry middleware.
Unfortunately, http headers are not copied from original response.

### Description

In retry middleware, add a boolean called stream on `ResponseRecorder`.
This boolean will be `true` when handling an http stream.

When `Flush` is called and stream is `false`, this means we didn't `copyHeaders` from original response yet.
Copy headers and set stream to true.

N.B. : stream vocable is also used in`fwd.go`.
```golang
	stream := f.streamResponse
	if !stream {
		contentType, err := utils.GetHeaderMediaType(response.Header, ContentType)
		if err == nil {
			stream = contentType == "text/event-stream"
		}
	}
```
### Reproduction of issue

Using docker-compose below

```json
{
    "version": "3",
    "services": {
        "dashing": {
            "image": "johnzan/dashing",
            "expose": ["3030"],
            "labels": {
                "traefik.enable": "true",
                "traefik.backend": "dashing",
                "traefik.frontend.rule": "PathPrefixStrip:/dashing",
                "traefik.port": "3030"
            }
        }

   }
}
```

Before patch
```
$ docker-compose up -d
$ curl -v http://localhost:8081/dashing/events
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8081 (#0)
> GET /dashing/events HTTP/1.1
> Host: localhost:8081
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Vary: Accept-Encoding
< Date: Mon, 17 Jul 2017 14:05:28 GMT
< Content-Type: text/plain; charset=utf-8
< Transfer-Encoding: chunked
< 
```
With patch, headers are copied
```
$ docker-compose up -d
$ curl -v http://localhost:8081/dashing/events
> GET /dashing/events HTTP/1.1
> Host: localhost:8081
> User-Agent: curl/7.47.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: text/event-stream;charset=utf-8
< Server: thin 1.6.1 codename Death Proof
< Vary: Accept-Encoding
< X-Accel-Buffering: no
< X-Content-Type-Options: nosniff
< Date: Mon, 17 Jul 2017 14:04:34 GMT
< Transfer-Encoding: chunked
< 
data: {"points":[{"x":620,"y":27},{"x":621,"y":22},{"x":622,"y":16},{"x":623,"y":29},{"x":624,"y":18},{"x":625,"y":41},{"x":626,"y":37},{"x":627,"y":26},{"x":628,"y":27},{"x":629,"y":15}],"id":"convergence","updatedAt":1500300270}
```

Fixes #1862

### DIFF
--- a/middlewares/retry.go
+++ b/middlewares/retry.go
@@ -42,6 +42,12 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 		recorder := NewRecorder()
 		recorder.responseWriter = rw
 		retry.next.ServeHTTP(recorder, r)
+
+		// In case of a stream, data has already been sent if code 200
+		if recorder.stream {
+			break
+		}
+
 		if !isNetworkError(recorder.Code) || attempts >= retry.attempts {
 			utils.CopyHeaders(rw.Header(), recorder.Header())
 			rw.WriteHeader(recorder.Code)
@@ -66,6 +72,7 @@ type ResponseRecorder struct {
 
 	responseWriter http.ResponseWriter
 	err            error
+	stream    bool
 }
 
 // NewRecorder returns an initialized ResponseRecorder.
@@ -74,6 +81,7 @@ func NewRecorder() *ResponseRecorder {
 		HeaderMap: make(http.Header),
 		Body:      new(bytes.Buffer),
 		Code:      200,
+		stream:    false,
 	}
 }
 
@@ -114,6 +122,10 @@ func (rw *ResponseRecorder) CloseNotify() <-chan bool {
 
 // Flush sends any buffered data to the client.
 func (rw *ResponseRecorder) Flush() {
+	if !rw.stream && !isNetworkError(rw.Code) {
+		rw.stream = true
+		utils.CopyHeaders(rw.responseWriter.Header(), rw.Header())
+	}
 	_, err := rw.responseWriter.Write(rw.Body.Bytes())
 	if err != nil {
 		log.Errorf("Error writing response in ResponseRecorder: %s", err)
--- a/middlewares/retry.go
+++ b/middlewares/retry.go
@@ -72,7 +72,7 @@ type ResponseRecorder struct {
 
 	responseWriter http.ResponseWriter
 	err            error
-	stream    bool
+	stream         bool
 }
 
 // NewRecorder returns an initialized ResponseRecorder.


romlinch:
#1862

emilevauge:
@romlinch Thanks for your contribution! Don't you think it should be fixed in our reverse-proxy lib  https://github.com/containous/oxy instead ?

emilevauge:
Ping @romlinch

emilevauge:
@romlinch 
>http://www.seloger.com/annonces/locations/appartement/lyon-7eme-69/la-guillotiere-sud/122975095.htm?p=

marco-jantke:
I had a closer look at this bug and a possible solution, as we also ran into it. 

@emilevauge regarding your suggestion that the fix might be done in the https://github.com/containous/oxy. I think this is not possible. The reverse proxy implementation works correct when it comes to SSE. I think I have to explain a bit what is going on, in order to make the problem understandable. In go you have to set the headers on a response and call `WriteHeader(statusCode)` _before_ sending the body. 

The reverse proxy does the following:
- copy headers to `ResponseWriter`
- set the status code to `ResponseWriter` 
- copy  the body to `ResponseWriter`
- _when_ it is an event stream response `Flush` the `ResponseWriter`

So this is all correct. The problem is with the response recorder implementation of the `retry` middleware. From the perspective of the reverse proxy we have replaced now the `ResponseWriter` with the `retryResponseRecorder` transparently. It doesn't know about it. Now the scenario looks like:
- copy headers to `ResponseWriter` -> will be stored only in property of `retryResponseRecorder` and not written through
- set the status code to `ResponseWriter` ->  will be stored only in property of `retryResponseRecorder` and not written through
- copy the body to `ResponseWriter` ->  will be stored only in property of `retryResponseRecorder` and not written through
- _when_ it is an event stream response `Flush` the `ResponseWriter` -> will be **actually** written through. This means we are writing the body **before** we set any headers. That's why we get the default headers and an http status code of 200. See [here](https://github.com/golang/go/blob/master/src/net/http/server.go#L116).

The _solution_ is basically as in this PR to set the headers and the http status code before calling `Write` in the `Flush` method for the first time. For the implementation I think a couple of things can be improved, but this is only relevant after the design-review.

emilevauge:
@marco-jantke In fact, I realized I was wrong after writing my comment and digging more in details into the code ;) But thanks for your great explanation anyway :clap: 

Design LGTM

marco-jantke:
@romlinch let me know whether you are willing continuing this PR. Otherwise I will pick that topic up the next days.

middlewares/retry.go
```diff
@@ -42,6 +42,12 @@ func (retry *Retry) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 		recorder := NewRecorder()
 		recorder.responseWriter = rw
 		retry.next.ServeHTTP(recorder, r)
+
+		// In case of a stream, data has already been sent if code 200
+		if recorder.stream {
```
Juliens: it could be more readable if stream become headerAlreadySent or responseAlreadySent ? WDYT

timoreimann:
Superseded and fixed by #2145.

PR #1871, opened 2017-07-17
Use test builder.

timoreimann:
This change introduces the builder pattern to the Marathon unit tests in order to simplify and reduce the amount of testing boilerplate.

Additional changes:

- Add missing unit tests.
- Make all tests look consistent.
- Use dedicated type for task states for increased type safety.
- Remove obsoleted `getApplication` function.

Note: I pushed the branch to the containous repo because I intend to submit another PR based onto this one.

@containous/marathon @containous/traefik PTAL.

### DIFF
--- a/provider/marathon/builder_test.go
+++ b/provider/marathon/builder_test.go
@@ -1,0 +1,131 @@
+package marathon
+
+import "github.com/gambol99/go-marathon"
+
+// Functions related to building applications.
+
+func createApplication(ops ...func(*marathon.Application)) marathon.Application {
+	app := marathon.Application{}
+	app.EmptyLabels()
+
+	for _, op := range ops {
+		op(&app)
+	}
+
+	return app
+}
+
+func appID(name string) func(*marathon.Application) {
+	return func(app *marathon.Application) {
+		app.Name(name)
+	}
+}
+
+func appPorts(ports ...int) func(*marathon.Application) {
+	return func(app *marathon.Application) {
+		app.Ports = append(app.Ports, ports...)
+	}
+}
+
+func label(key, value string) func(*marathon.Application) {
+	return func(app *marathon.Application) {
+		app.AddLabel(key, value)
+	}
+}
+
+func healthChecks(checks ...*marathon.HealthCheck) func(*marathon.Application) {
+	return func(app *marathon.Application) {
+		for _, check := range checks {
+			app.AddHealthCheck(*check)
+		}
+	}
+}
+
+func portDefinition(port int) func(*marathon.Application) {
+	return func(app *marathon.Application) {
+		app.AddPortDefinition(marathon.PortDefinition{
+			Port: &port,
+		})
+	}
+}
+
+func ipAddrPerTask(port int) func(*marathon.Application) {
+	return func(app *marathon.Application) {
+		p := marathon.Port{
+			Number: port,
+			Name:   "port",
+		}
+		disc := marathon.Discovery{}
+		disc.AddPort(p)
+		ipAddr := marathon.IPAddressPerTask{}
+		ipAddr.SetDiscovery(disc)
+		app.SetIPAddressPerTask(ipAddr)
+	}
+}
+
+// Functions related to building tasks.
+
+func createTask(ops ...func(*marathon.Task)) marathon.Task {
+	t := marathon.Task{
+		// The vast majority of tests expect the task state to be TASK_RUNNING.
+		State: string(taskStateRunning),
+	}
+
+	for _, op := range ops {
+		op(&t)
+	}
+
+	return t
+}
+
+func createLocalhostTask(ops ...func(*marathon.Task)) marathon.Task {
+	t := createTask(
+		host("localhost"),
+		ipAddresses("127.0.0.1"),
+	)
+
+	for _, op := range ops {
+		op(&t)
+	}
+
+	return t
+}
+
+func taskPorts(ports ...int) func(*marathon.Task) {
+	return func(t *marathon.Task) {
+		t.Ports = append(t.Ports, ports...)
+	}
+}
+
+func host(h string) func(*marathon.Task) {
+	return func(t *marathon.Task) {
+		t.Host = h
+	}
+}
+
+func ipAddresses(addresses ...string) func(*marathon.Task) {
+	return func(t *marathon.Task) {
+		for _, addr := range addresses {
+			t.IPAddresses = append(t.IPAddresses, &marathon.IPAddress{
+				IPAddress: addr,
+				Protocol:  "tcp",
+			})
+		}
+	}
+}
+
+func state(s TaskState) func(*marathon.Task) {
+	return func(t *marathon.Task) {
+		t.State = string(s)
+	}
+}
+
+func healthCheckResultLiveness(alive ...bool) func(*marathon.Task) {
+	return func(t *marathon.Task) {
+		for _, a := range alive {
+			t.HealthCheckResults = append(t.HealthCheckResults, &marathon.HealthCheckResult{
+				Alive: a,
+			})
+		}
+	}
+}
--- a/provider/marathon/marathon_test.go
+++ b/provider/marathon/marathon_test.go
@@ -3,6 +3,7 @@ package marathon
 import (
 	"errors"
 	"fmt"
+	"math"
 	"reflect"
 	"testing"
 
@@ -50,27 +51,12 @@ func TestMarathonLoadConfigNonAPIErrors(t *testing.T) {
 		expectedBackends  map[string]*types.Backend
 	}{
 		{
-			desc: "simple application",
-			application: marathon.Application{
-				Ports:  []int{80},
-				Labels: &map[string]string{},
-			},
-			task: marathon.Task{
-				Host:  "localhost",
-				Ports: []int{80},
-				IPAddresses: []*marathon.IPAddress{
-					{
-						IPAddress: "127.0.0.1",
-						Protocol:  "tcp",
-					},
-				},
-			},
+			desc:        "simple application",
+			application: createApplication(appPorts(80)),
+			task:        createLocalhostTask(taskPorts(80)),
 			expectedFrontends: map[string]*types.Frontend{
 				"frontend-app": {
-					Backend:        "backend-app",
-					PassHostHeader: true,
-					BasicAuth:      []string{},
-					EntryPoints:    []string{},
+					Backend: "backend-app",
 					Routes: map[string]types.Route{
 						"route-host-app": {
 							Rule: "Host:app.docker.localhost",
@@ -91,21 +77,15 @@ func TestMarathonLoadConfigNonAPIErrors(t *testing.T) {
 			},
 		},
 		{
-			desc: "filtered task",
-			application: marathon.Application{
-				Ports:  []int{80},
-				Labels: &map[string]string{},
-			},
-			task: marathon.Task{
-				Ports: []int{80},
-				State: "TASK_STAGING",
-			},
+			desc:        "filtered task",
+			application: createApplication(appPorts(80)),
+			task: createLocalhostTask(
+				taskPorts(80),
+				state(taskStateStaging),
+			),
 			expectedFrontends: map[string]*types.Frontend{
 				"frontend-app": {
-					Backend:        "backend-app",
-					PassHostHeader: true,
-					BasicAuth:      []string{},
-					EntryPoints:    []string{},
+					Backend: "backend-app",
 					Routes: map[string]types.Route{
 						"route-host-app": {
 							Rule: "Host:app.docker.localhost",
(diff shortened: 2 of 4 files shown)


provider/marathon/marathon_test.go
```diff
@@ -934,440 +700,565 @@ func TestMarathonGetProtocol(t *testing.T) {
...
 }
 
 func TestMarathonGetHealthCheckPath(t *testing.T) {
-	tests := []struct {
-		desc  string
-		value *string
-		want  string
+	cases := []struct {
+		desc     string
+		value    *string
```
marco-jantke: AFAICS its not necessary anymore to use `*string` as we only dereference the pointer when passing it to `AddLabel`. Or do I miss something?

provider/marathon/marathon_test.go
```diff
@@ -934,440 +700,565 @@ func TestMarathonGetProtocol(t *testing.T) {
...
 }
 
 func TestMarathonGetHealthCheckInterval(t *testing.T) {
-	tests := []struct {
-		desc  string
-		value *string
-		want  string
+	cases := []struct {
+		desc     string
+		value    *string
```
marco-jantke: Same as previous comment.

provider/marathon/marathon_test.go
```diff
@@ -275,29 +211,21 @@ func TestMarathonLoadConfigNonAPIErrors(t *testing.T) {
...
-				IPAddresses: []*marathon.IPAddress{
-					{
-						IPAddress: "127.0.0.1",
-						Protocol:  "tcp",
-					},
-				},
-			},
+				}),
+			),
+
```
timoreimann: <fist-swinging>Argh, damn you unintentional blank lines!</fist-swinging>

Removed. :-)

provider/marathon/marathon_test.go
```diff
@@ -641,184 +517,82 @@ func TestMarathonGetPort(t *testing.T) {
...
-						},
-					},
-				},
-			},
-			task: marathon.Task{
-				AppID: "app",
-				Ports: []int{},
-			},
+			desc: "port definition available",
+			application: application(
```
timoreimann: I did remove it intentionally for two reasons (should have commented this in the first place, sorry):

1. My original thinking was that the IP-per-task setup would serve as a safeguard to running into an undesirable happy path due to a bug in the code (i.e., false positive). On closer inspection, this seems highly unlikely since we can't really produce the expected port result of 443 without configuring it somewhere, which we only do for the test case at hand that we intend to cover.
1. Configuring both port definitions and IPs-per-task should actually not be allowed: [go-marathon explicitly clears the former when setting the latter](https://github.com/gambol99/go-marathon/blob/15ea23e360abb8b25071e677aed344f31838e403/application.go#L152-L157) because the Marathon API would otherwise reject our API request. In that sense, it makes even less sense to set up both in the test (though I wouldn't relax the production code logic because of that because you never know how the Marathon API may evolve over time).

provider/marathon/marathon_test.go
```diff
@@ -934,440 +700,565 @@ func TestMarathonGetProtocol(t *testing.T) {
...
-			t.Fatalf("expected %q, got %q", a.expected, actual)
-		}
+	for _, c := range cases {
+		c := c
+		t.Run(c.desc, func(t *testing.T) {
+			t.Parallel()
+			provider := &Provider{}
+			actual := provider.getPassHostHeader(c.application)
+			if actual != c.expected {
+				t.Fatalf("actual %q, expected %q", actual, c.expected)
```
timoreimann: Here and almost everywhere else -- updated.

provider/marathon/marathon_test.go
```diff
@@ -934,440 +700,565 @@ func TestMarathonGetProtocol(t *testing.T) {
...
+func TestMarathonGetCircuitBreakerExpression(t *testing.T) {
+	cases := []struct {
+		desc        string
+		application marathon.Application
+		expected    string
+	}{
+		{
+			desc:        "label missing",
+			application: application(),
+			expected:    "NetworkErrorRatio() > 1",
```
timoreimann: This part is somewhat confusing: While we do return a default value, we [only include circuit-breaker settings](https://github.com/containous/traefik/blob/c53be185f4c71d41c1b71cd80a78d1613a970b65/templates/marathon.tmpl#L22) when [the label is actually set](https://github.com/containous/traefik/blob/c53be185f4c71d41c1b71cd80a78d1613a970b65/provider/marathon/marathon.go#L369-L372).

It might make sense to change the default return value to an empty string to improve clarity. Let's consider that for a future PR.

provider/marathon/marathon_test.go
timoreimann: You're right: There's still one test left where we want to distinguish set/unset/omitted, but not this one.

Simplified.

provider/marathon/marathon_test.go
timoreimann: Simplified too.

timoreimann:
@marco-jantke all comments addressed.

provider/marathon/marathon_test.go
marco-jantke: Np, SGTM.

ldez:
@timoreimann could you fix the linting?
```
Errors from golint:
provider/marathon/marathon.go:31:6: exported type TaskState should have comment or be unexported
```

timoreimann:
@ldez should be good now.

provider/marathon/marathon_test.go
```diff
@@ -1453,11 +1336,15 @@ func TestMarathonGetBasicAuth(t *testing.T) {
...
-			if !reflect.DeepEqual(c.expected, actual) {
-				t.Errorf("expected %q, got %q", c.expected, actual)
+			if !reflect.DeepEqual(actual, c.expected) {
+				t.Errorf("actual %q, expected %q", actual, c.expected)
 			}
 		})
 	}
 }
+
+func stringp(s string) *string {
```
ldez: maybe move this into `testhelpers` like `Intp`?

provider/marathon/marathon_test.go
```diff
@@ -934,440 +699,563 @@ func TestMarathonGetProtocol(t *testing.T) {
...
+	cases := []struct {
+		desc        string
 		application marathon.Application
 		expected    string
 	}{
 		{
-			application: marathon.Application{
-				Labels: &map[string]string{}},
+			desc:        "label missing",
+			application: application(),
```
ldez: To avoid the ambiguity of names, I rather name builders like `anApplication(...)` and if necessary options like `withLabels(...)`

provider/marathon/builder_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+}
+
+func appPorts(ports ...int) func(*marathon.Application) {
+	return func(app *marathon.Application) {
+		app.Ports = append(app.Ports, ports...)
+	}
+}
+
+func labels(l map[string]string) func(*marathon.Application) {
```
ldez: I don't really show the need to use a map here.
Why not use a simple function like this:
```golang
func label(key, value string) {
	return func(app *marathon.Application) {
		app.AddLabel(k, v)
	}
}
```

provider/marathon/marathon_test.go
timoreimann: In general, I'd prefer not to use longer names in order to keep things concise since builder functions are used very frequently. IMHO, we shouldn't have extra verbs on builder operations.

`anApplication` looks quite weird to my eyes. :) How about `createApp` and `createTask` instead?

provider/marathon/marathon_test.go
timoreimann: Good idea, done.

provider/marathon/builder_test.go
timoreimann: I was leaning towards such an implementation too at some point, so I made the change. 👍

timoreimann:
Re: backend/frontend refactoring: I'd like to leave that for a separate PR as this one is already fairly big. Plus, we need to "builderize" the configuration in numerous places (server, kubernetes, etc.), so I'd rather do it in one focused wash (which will also make it easier to see how we need to design that builder).

provider/marathon/marathon_test.go
ldez: `create` prefix is not fluent syntax 😉 (builders must be fluent, I think)
@vdemeester do that like me http://vincent.demeester.fr/posts/2017-01-01-go-testing-functionnal-builders/

But in this case, I prefer you don't change anything instead use a `create` prefix.

timoreimann: Okay, now I'm confused: The blog post uses _a/an_ and _with_ prefixes. However, our Docker builder comes with just two _with_ functions and otherwise uses no prefixes. ¯\_(ツ)_/¯

@vdemeester: can you help me out here? :)

testhelpers/helpers.go
```diff
@@ -12,6 +12,12 @@ func Intp(i int) *int {
 	return &i
 }
 
+// Stringp returns a pointer to the given string value.
+func Stringp(s string) *string {
+	return &s
+}
+
```
vdemeester: one `\n` too much 😝

provider/marathon/marathon_test.go
```diff
@@ -50,27 +51,12 @@ func TestMarathonLoadConfigNonAPIErrors(t *testing.T) {
...
-				Ports: []int{80},
-				IPAddresses: []*marathon.IPAddress{
-					{
-						IPAddress: "127.0.0.1",
-						Protocol:  "tcp",
-					},
-				},
-			},
+			desc:        "simple application",
+			application: application(appPorts(80)),
```
vdemeester: You seem to use `appPorts(80)` a lot, maybe it should be a default value in the application (and setting it only when you actually care about the value in your test) 👼

provider/marathon/marathon_test.go
```diff
@@ -92,30 +78,15 @@ func TestMarathonLoadConfigNonAPIErrors(t *testing.T) {
...
-						Protocol:  "tcp",
-					},
-				},
-			},
+			application: application(
+				appPorts(80),
+				label(types.LabelBackendLoadbalancerMethod, "drr"),
+				label(types.LabelBackendCircuitbreakerExpression, "NetworkErrorRatio() > 0.5"),
+			),
+			task: localhostTask(taskPorts(80)),
```
vdemeester: Same for `taskPorts`

testhelpers/helpers.go
timoreimann: Begone!

provider/marathon/marathon_test.go
timoreimann: I should already only be setting it when I actually need it. (There are a number of `application()`-only invocations where I don't pass the `appPorts` builder function.)

I thought about setting default ports on the app and task but dropped the idea eventually: We have a lot of different cases where we sometimes need just the Marathon port and sometimes the task port; sometimes they need to be equal and sometimes they don't.

So if I added a default, I'd also need to override it explicitly sometimes. In the end, this was too intransparent for me, so I went down the slightly more explicit and but also more clear route.

provider/marathon/marathon_test.go
timoreimann: See above. :)

provider/marathon/marathon_test.go
timoreimann: @vdemeester's out-of-band feedback is that he more often than not drops the `with` prefix in his builders. :-)

So I just added the `create` prefix. Not sure if it violates fluentd builder syntaxes. I did this a lot in Java, but maybe that's a sin of the past. ;)

timoreimann:
Rebased and addressed all comments for now.

PTAL again.

provider/marathon/builder_test.go
```diff
@@ -0,0 +1,131 @@
+package marathon
+
+import "github.com/gambol99/go-marathon"
+
+// Functions related to building applications.
+
+func createApplication(ops ...func(*marathon.Application)) marathon.Application {
```
ldez: The conventional way in Java it's a `build` prefix.

- https://docs.oracle.com/javase/8/docs/api/java/util/Calendar.Builder.html
- https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.Builder.html
- https://github.com/spring-projects/spring-framework/blob/bc14c5ba83e1f211628456bbccce7b2531aac58c/spring-test/src/test/java/org/springframework/test/context/support/BootstrapTestUtilsMergedConfigTests.java
- https://github.com/spring-projects/spring-framework/blob/7018804e8451fa935e8715754de43a3ef62fa5c8/spring-webflux/src/test/java/org/springframework/web/reactive/function/server/RequestPredicatesTests.java

I prefer you simply remove the `create` prefix.

timoreimann:
I only added the prefix because I understood you asked for it. Will happily remove it again in another PR. :-)

PR #1872, opened 2017-07-18
Update docs for dnsimple env vars.

klud1:
### Description
Yesterday I followed the docs (https://docs.traefik.io/toml/#acme-lets-encrypt-configuration) to set traefik with dnsimple as dns provider in order to get Let's Encrypt certs with txt dns challenge.

Thing is traefik says I need to set some env vars for the container to use and setup dnsimple (`DNSIMPLE_EMAIL` and `DNSIMPLE_API_KEY`  in this case), I set them but wasn't able to use dnsimple (the container exited), found out in the logs it exited because of something related to an OAuth key that wasn't found ('cause it was not set), after searching for the library traefik use for this (https://github.com/xenolf/lego/blob/master/providers/dns/dnsimple/dnsimple.go#L25) I figured out I needed to set a different env var instead of `DNSIMPLE_API_KEY` (`DNSIMPLE_OAUTH_TOKEN`) so i did, now it's working without problems.

So i opened this PR in order to update the docs.

PS. Sorry, bad commit description, it's midnight here and i'm kinda sleepy.

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -374,7 +374,7 @@ entryPoint = "https"
 # and provide environment variables with access keys to enable setting it:
 #  - cloudflare: CLOUDFLARE_EMAIL, CLOUDFLARE_API_KEY
 #  - digitalocean: DO_AUTH_TOKEN
-#  - dnsimple: DNSIMPLE_EMAIL, DNSIMPLE_API_KEY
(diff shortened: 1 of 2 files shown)


ldez:
@klud1 Thanks for the contribution 👍 

Could you also change this file: https://github.com/containous/traefik/blob/master/traefik.sample.toml#L157

klud1:
@ldez sure! updated that file as well.

PR #1874, opened 2017-07-18
Added ability to override frontend priority for k8s ingress router

DiverOfDark:
### Description

Added ability to specify ```traefik.frontend.priority``` priority from kubernetes ingress annotation.
This is useful when you have following case:

Ingress A:
Host: *.mydomain

Ingress B:
Host: specific.mydomain

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -613,7 +613,7 @@ This can be done by adding annotation ```traefik.frontend.priority```, i.e.:
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
-  name: cheeses
+  name: wildcard-cheeses
   annotations:
     traefik.frontend.priority: 1
 spec:
@@ -621,14 +621,14 @@ spec:
   - host: *.minikube
     http:
       paths:
-      - path: /stilton
+      - path: /
         backend:
           serviceName: stilton
           servicePort: http
 
 kind: Ingress
 metadata:
-  name: cheeses
+  name: specific-cheeses
   annotations:
     traefik.frontend.priority: 2
 spec:
@@ -636,7 +636,7 @@ spec:
   - host: specific.minikube
     http:
       paths:
-      - path: /stilton
+      - path: /
         backend:
           serviceName: stilton
           servicePort: http
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -604,6 +604,45 @@ You should now be able to visit the websites in your browser.
 * [cheeses.minikube/cheddar](http://cheeses.minikube/cheddar/)
 * [cheeses.minikube/wensleydale](http://cheeses.minikube/wensleydale/)
 
+## Specifying priority for routing
+
+Sometimes you need to specify priority for ingress route, especially when handling wildcard routes.
+This can be done by adding annotation ```traefik.frontend.priority```, i.e.:
+
+```yaml
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: cheeses
+  annotations:
+    traefik.frontend.priority: 1
+spec:
+  rules:
+  - host: *.minikube
+    http:
+      paths:
+      - path: /stilton
+        backend:
+          serviceName: stilton
+          servicePort: http
+
+kind: Ingress
+metadata:
+  name: cheeses
+  annotations:
+    traefik.frontend.priority: 2
+spec:
+  rules:
+  - host: specific.minikube
+    http:
+      paths:
+      - path: /stilton
+        backend:
+          serviceName: stilton
+          servicePort: http
+```
+
+
 ## Forwarding to ExternalNames
 
 When specifying an [ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors),
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -607,7 +607,7 @@ You should now be able to visit the websites in your browser.
 ## Specifying priority for routing
 
 Sometimes you need to specify priority for ingress route, especially when handling wildcard routes.
-This can be done by adding annotation ```traefik.frontend.priority```, i.e.:
+This can be done by adding annotation `traefik.frontend.priority`, i.e.:
 
 ```yaml
 apiVersion: extensions/v1beta1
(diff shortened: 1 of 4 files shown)


provider/kubernetes/kubernetes.go
```diff
@@ -186,6 +186,17 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 				witelistSourceRangeAnnotation := i.Annotations[annotationKubernetesWhitelistSourceRange]
 				whitelistSourceRange := provider.SplitAndTrimString(witelistSourceRangeAnnotation)
 
+				priority := len(pa.Path)
+
+				priorityString, ok := i.Annotations[types.LabelFrontendPriority]
+				if ok {
```
timoreimann: You can merge this line with the previous one, thereby reducing the scope of `priorityString` to the necessary minimum.

provider/kubernetes/kubernetes.go
```diff
@@ -186,6 +186,17 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 				witelistSourceRangeAnnotation := i.Annotations[annotationKubernetesWhitelistSourceRange]
 				whitelistSourceRange := provider.SplitAndTrimString(witelistSourceRangeAnnotation)
 
+				priority := len(pa.Path)
+
+				priorityString, ok := i.Annotations[types.LabelFrontendPriority]
```
timoreimann: I'd suggest to rename `priorityString` to `priorityRaw` to better convey its semantics.

provider/kubernetes/kubernetes.go
```diff
@@ -186,6 +186,17 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+				if ok {
+					priorityParsed, err := strconv.Atoi(priorityString)
+
+					if err == nil {
```
timoreimann: We should not ignore errors.

When we encounter an error, we can either

- drop the associated frontend (by deleting it from the frontends map) or
- fall back to the default priority of 1.

We kind of do a mix in Traefik, so we cannot just point at any single pattern. Personally, I'm much more in favor of dropping the frontend as this makes the error explicit and way easier to identify. @dtomcej @errm thoughts?

In either case, the error should be logged.

provider/kubernetes/kubernetes.go
```diff
@@ -186,6 +186,17 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+						priority = priorityParsed;
```
timoreimann: No trailing semicolon needed here.

provider/kubernetes/kubernetes.go
```diff
@@ -186,6 +186,17 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					}
+				}
```
timoreimann: Can you put the entire priority processing logic into a dedicated function?

DiverOfDark:
[ ] Also need to update documentation

provider/kubernetes/kubernetes.go
DiverOfDark: Actually in current logic it will fall back to the default priority of len(Path), as if no priority is set.

Personally I think that dropping the frontend is ok when it has wrong Auth (because we obviously don't wanna show something secret to all users), but here probably routing with default priority is fine.

DiverOfDark:
@timoreimann , looks like I changed everything you asked, could you please check again?

provider/kubernetes/kubernetes.go
timoreimann: Yeah that does sound reasonable. 👍

provider/kubernetes/kubernetes.go
```diff
@@ -322,6 +325,23 @@ func getRuleForPath(pa v1beta1.HTTPIngressPath, i *v1beta1.Ingress) string {
...
+	priority := len(path.Path)
+
+	priorityRaw, ok := i.Annotations[types.LabelFrontendPriority]
+	if ok {
+		priorityParsed, err := strconv.Atoi(priorityRaw)
+
+		if err == nil {
+			priority = priorityParsed
+		} else {
+			log.Errorf("Error in ingress: failed to parse '%q' value '%q'.", types.LabelFrontendPriority, priorityRaw)
```
timoreimann: We don't need the extra `'` quotes if we already use `%q`, do we?

provider/kubernetes/kubernetes_test.go
```diff
@@ -1778,6 +1778,113 @@ func TestIngressAnnotations(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+
+	actualJSON, _ := json.Marshal(actual)
+	expectedJSON, _ := json.Marshal(expected)
+
+	if !reflect.DeepEqual(actual, expected) {
```
timoreimann: Please use `assert.Equal` (like [here](https://github.com/containous/traefik/blob/074b31b5e98810181e5b02ab6bad593b3438f58f/provider/kubernetes/kubernetes_test.go#L1778)). Comparing with `DeepEqual` and showing JSON can hide subtle differences, like `nil` slices vs. empty slices.

I know we use it everywhere in our Kubernetes tests, which is something we need to get rid of. :-)

timoreimann:
Sorry, I just realized something: Should the annotation actually go on the Ingress or rather the Service object?

We do keep [some annotations on Services](https://github.com/containous/traefik/blob/64ee68763ba36bee78bf4bc726f53780cf658e57/provider/kubernetes/kubernetes.go#L237). I know @dtomcej previously explained why, but I forgot...

dtomcej:
The annotation should go on the ingress in this case, because it affects how the controller handles the routing to the service. The ingress does not care how the services are configured or structured :)

docs/user-guide/kubernetes.md
```diff
@@ -510,6 +510,45 @@ You should now be able to visit the websites in your browser.
 * [cheeses.minikube/cheddar](http://cheeses.minikube/cheddar/)
 * [cheeses.minikube/wensleydale](http://cheeses.minikube/wensleydale/)
 
+## Specifying priority for routing
+
+Sometimes you need to specify priority for ingress route, especially when handling wildcard routes.
+This can be done by adding annotation ```traefik.frontend.priority```, i.e.:
```
ldez: could you replace \``` only one \` ?

DiverOfDark: sure, replaced

Issue #1174, opened 2017-02-20
Traefik is routing traffic to wrong backend.


klausenbusk:
### What version of Traefik are you using (`traefik version`)?
~~v1.1.2~~
**Edit:** also present with `v1.2.0-rc1`

### What is your environment & configuration (arguments, toml...)?
Traefik is running in a Docker container on CoreOS and pulling config from etcd and all traffic is routed through Cloudflare first (ddos protection).

etcd config: (taken from debug log `Configuration received from provider etcd:`)
```json
{
  "backends": {
    "b1": {
      "servers": {
        "10.133.102.22": {
          "url": "http://10.133.102.22",
          "weight": 0
        },
        "10.133.92.63": {
          "url": "http://10.133.92.63",
          "weight": 0
        }
      },
      "loadBalancer": {
        "method": "wrr"
      }
    },
    "b2": {
      "servers": {
        "10.133.13.121": {
          "url": "http://10.133.13.121",
          "weight": 0
        }
      },
      "loadBalancer": {
        "method": "wrr"
      }
    },
    "b3": {
      "servers": {
        "grafana": {
          "url": "http://10.133.59.44",
          "weight": 0
        }
      },
      "loadBalancer": {
        "method": "wrr"
      }
    },
    "b4": {
      "servers": {
        "10.133.21.211": {
          "url": "http://10.133.21.211",
          "weight": 0
        }
      },
      "loadBalancer": {
        "method": "wrr"
      }
    },
    "b5": {
      "servers": {
        "emq": {
          "url": "http://10.133.112.89:8090",
          "weight": 0
        }
      },
      "loadBalancer": {
        "method": "wrr"
      }
    }
  },
  "frontends": {
    "f1": {
      "entryPoints": [
        "http"
      ],
      "backend": "b1",
      "routes": {
        "r1": {
          "rule": "HostRegexp:admin.foobar.com,secure.foobar.com,{subdomain:(config|queue|status)}.barfoo.eu,barfoo.eu"
        }
      },
      "passHostHeader": true,
      "priority": 0
    },
    "f2": {
      "entryPoints": [
        "http"
      ],
      "backend": "b2",
      "routes": {
        "r1": {
          "rule": "HostRegexp:dev.foobar.com,secure.dev.foobar.com,{subdomain:(config|queue|status)}.dev.barfoo.eu,dev.barfoo.eu"
        }
      },
      "passHostHeader": true,
      "priority": 0
    },
    "f3": {
      "entryPoints": [
[...]

PR #1875, opened 2017-07-18
fix GraceTimeOut documentation

marco-jantke:
Documentation stated that GraceTimeOut describes the timeout between
hot-reloads, which is not the case. GraceTimeOut describes the timeout
Traefik uses to finish serving active requests before stopping only.

### DIFF
--- a/docs/toml.md
+++ b/docs/toml.md
@@ -9,10 +9,10 @@
 # Global configuration
 ################################################################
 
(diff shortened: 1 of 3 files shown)


timoreimann:
The doc update reflects my understanding of the actual behavior as well.

There was [some discussion on the desirable behavior](https://github.com/containous/traefik/issues/1174#issuecomment-283310193) in the past though. @emilevauge to chime in.

PR #1883, opened 2017-07-20
Add support for readiness checks.

### LINKED ISSUE #1185: Continue to serve traffic to Marathon applications during leader (re-)election phases while not serving to unready ones
#653 introduced a change to the Marathon provider that made Traefik continue to serve traffic even in the absence of health check results. The motivation was that during phases where a Marathon (re-)election would take place, Marathon would not have health-checked all applications and thus not have provided health check results, causing Traefik to lose all backend servers. Technically, however, those applications would still be available (at least with a very high probability).

#1080 undid the change again since it would cause applications to be taken into load-balancing rotation prior to their health checks going green. This could lead to traffic being served while the application was still bootstrapping.

I think this issue should come up with the answers to two questions:

1. Should #1080 be reverted prior to releasing 1.2.0?
2. Can we come up with a way to resolve the two conflicting changes such that both goals can be achieved? 

/cc @gomes @jangie @emilevauge @vdemeester @errm

timoreimann:
This change adds support for Marathon readiness checks: If the option is set on Traefik and applications are found to come with configured readiness checks, the check results (as conducted by Marathon and exposed through its API) will be taken into account at deployment times
in order to decide if a task should be taken into load-balancing rotation or not.

Note that I had to extend the event filter that controls how often the Marathon provider polls the API so that the readiness checker is able to pick up all necessary readiness state transitions.

Fixes #1185, #1559

Based on #1871 for now; will be rebased onto master once the dependent branch gets merged.

### DIFF
--- a/provider/marathon/readiness.go
+++ b/provider/marathon/readiness.go
@@ -1,0 +1,122 @@
+package marathon
+
+import (
+	"time"
+
+	"github.com/containous/traefik/log"
+	marathon "github.com/gambol99/go-marathon"
+)
+
+const (
+	// readinessCheckDefaultTimeout is the default timeout for a readiness
+	// check if no check timeout is specified on the application spec. This
+	// should really never be the case, but better be safe than sorry.
+	readinessCheckDefaultTimeout time.Duration = 10 * time.Second
+	// readinessCheckSafetyMargin is some buffer duration to account for
+	// small offsets in readiness check execution.
+	readinessCheckSafetyMargin time.Duration = 5 * time.Second
+	readinessLogHeader         string        = "Marathon readiness check: "
+)
+
+type readinessChecker struct {
+	checkDefaultTimeout time.Duration
+	checkSafetyMargin   time.Duration
+	traceLogging        bool
+}
+
+func defaultReadinessChecker(isTraceLogging bool) *readinessChecker {
+	return &readinessChecker{
+		checkDefaultTimeout: readinessCheckDefaultTimeout,
+		checkSafetyMargin:   readinessCheckSafetyMargin,
+		traceLogging:        isTraceLogging,
+	}
+}
+
+func (rc *readinessChecker) Do(task marathon.Task, app marathon.Application) bool {
+	if rc == nil {
+		// Readiness checker disabled.
+		return true
+	}
+
+	switch {
+	case len(app.Deployments) == 0:
+		// We only care about readiness during deployments; post-deployment readiness
+		// can be covered by a periodic post-deployment probe (i.e., Traefik health checks).
+		rc.logf("task %s app %s: ready = true [no deployment ongoing]", task.ID, app.ID)
+		return true
+
+	case app.ReadinessChecks == nil || len(*app.ReadinessChecks) == 0:
+		// Applications without configured readiness checks are always considered
+		// ready.
+		rc.logf("task %s app %s: ready = true [no readiness checks on app]", task.ID, app.ID)
+		return true
+	}
+
+	// Loop through all readiness check results and return the results for
+	// matching task IDs.
+	if app.ReadinessCheckResults != nil {
+		for _, readinessCheckResult := range *app.ReadinessCheckResults {
+			if readinessCheckResult.TaskID == task.ID {
+				rc.logf("task %s app %s: ready = %t [evaluating readiness check ready state]", task.ID, app.ID, readinessCheckResult.Ready)
+				return readinessCheckResult.Ready
+			}
+		}
+	}
+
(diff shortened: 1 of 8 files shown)


provider/marathon/readiness.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	// Finally, we can be certain this task is not part of the deployment (i.e.,
+	// it's an old task that's going to transition into the TASK_KILLING and/or
+	// TASK_KILLED state as new tasks' readiness checks gradually turn green.)
+	rc.logf("task %s app %s: ready = true [task with start-time %s not involved in deployment (elapsed time since task start: %s)]", task.ID, app.ID, startTime.Format(time.RFC3339), since)
+	return true
+}
+
+func (rc *readinessChecker) logf(format string, args ...interface{}) {
```
timoreimann: Note to reviewers: Ideally, I'd like to support logging in "trace level" more generally (i.e., without having to pass the trace flag into a particular function that needs this information).

I'm open to suggestions.

provider/marathon/readiness.go
```diff
@@ -0,0 +1,122 @@
...
+	// time of the task: It should take Marathon at most one readiness check timeout
+	// interval (plus some safety margin to account for the delayed nature of
+	// distributed systems) for readiness check results to be returned along the API
+	// response. Once the task turns old enough, we assume it to be part of a
+	// pre-existing deployment and mark it as ready. Note that it is okay to err
+	// on the side of caution and consider a task unready until the safety time
+	// window has elapsed because a newly created task should be readiness-checked
+	// and be given a result fairly shortly after its creation (i.e., on the scale
+	// of seconds).
+	readinessCheckTimeoutSecs := (*app.ReadinessChecks)[0].TimeoutSeconds
```
timoreimann: Note to reviewer: The readiness checker logic should be fairly clear except for this corner case (hence the lengthy comment). Internally, we went through several iterations to improve the code and iron out corner cases (though with [Bamboo](https://github.com/QubitProducts/bamboo) instead of Traefik, the logic is identical -- I basically copy & pasted the implementation). It's been running in our production system for several weeks now, so it should have gained a fair level of robustness.

Nevertheless, I'd happy for suggestions to make it simpler/better.

timoreimann:
@ldez commit 50489fd118006f94210518b4db7477a6ac918da4 removes the prefixes again. Do I understand correctly that this is how we want things now? Please let me know.

emilevauge:
@timoreimann wow, impressive job :clap: 
I'm not a huge fan of the naming `respectReadinessChecks` ;) But I can live with that, even if I tend to prefer shorter names for options (I can see the Java lover behind this ^^).
This PR is quite big&risky and it would be even better if we could have some integration tests on this (https://github.com/containous/traefik/pull/1406). I know you are waiting for https://github.com/docker/libcompose/issues/484 to complete: @vdemeester we really need your help on this one :)
LGTM on design

timoreimann:
The Javaness in me is less of a love and more of a bad, hard-to-shake-off habit. :grimacing:

I'm very open to shortening the option name. `respectReadiness`?  `filterReady[Tasks]`? `onlyReady[Tasks]`? `readyOnly`? Any suggestions?

While an integration test would be nice, I can say from experience with this feature that covering all cases will be difficult: It depends a lot on how applications/tasks behave and the timing involved. It took us two or three internal patches to get this right with Bamboo. On the plus side, this means the algorithm is battle-tested to a certain degree already. And users can turn on/off readiness check filtering on demand.

That said, if someone can help out on the upstream libcompose blocker, I'd be more than happy to contribute an integration test. :angel:

provider/marathon/readiness_test.go
```diff
@@ -0,0 +1,134 @@
...
+				readinessCheck(0),
+			),
+			rc: readinessChecker{
+				checkDefaultTimeout: 10 * time.Second,
+			},
+			expectedReady: true,
+		},
+	}
+
+	for _, c := range cases {
```
ldez: I prefer `test` or `case` instead of `c`

timoreimann: I can't use `case` because that's a reserved keyword in Go, so I renamed everything to `test`.

provider/marathon/readiness.go
timoreimann: Indeed, much better. Done.

timoreimann:
@Juliens:

> Do you know if the problem of "bad readyness check result" will be fixed in marathon, because this add a lots of useless and complicated code in traefik just in order to handle this "marathon bug".

Last time I checked I haven't seen a bug report on the matter. I'll double-check, file a report if I find nothing, and post it here.

PR #1884, opened 2017-07-21
Added RetryMetrics to DataDog and StatsD providers

aantono:
Follow-up on RFC from #1701


marco-jantke:
If it helps anything, LGTM :)

PR #1885, opened 2017-07-21
Improve Let's Encrypt documentation

nmengin:
### Description

Feature to improve the let's encrypt documentation.
A user guide has been added to show how to create a Traefik Docker service with Let's Encrypt configuration


docs/user-guide/examples.md
```diff
@@ -50,19 +50,161 @@ defaultEntryPoints = ["http", "https"]
...
 ```toml
 [entryPoints]
   [entryPoints.https]
   address = ":443"
     [entryPoints.https.tls]
-      # certs used as default certs
-      certFile = "tests/traefik.crt"
-      keyFile = "tests/traefik.key"
+      CertFile = "/etc/pki/tls/certs/mycert.crt"
```
dtomcej: Can we use a cert and key that exist in the repo already (perhaps from the integration tests), so that we don't provide fictitious paths to confuse new users?

I would either keep the same cert path ("tests/traefik.crt") or choose another from the repo

emilevauge: we do have some certs in `/examples` ;)

PR #1828, opened 2017-07-05
Undo the Secrets controller sync wait.

timoreimann:
When Secrets permissions have not been granted (which is likely to be the case for users not needing the basic auth feature), the watch on the Secrets API will never yield a response, thereby causing the controller to never sync successfully, and in turn causing the check for all controller synchronizations to fail consistently. Thus, no event will ever be handled.

Fixes #1826 
Reverts a part of #1798
Verified that #1794 continues to be fixed

Reviewers and @containous/traefik: PTAL.

PR #1798, opened 2017-06-27
Fix race on closing event channel.

timoreimann:
- Wait for secret controller to finish synchronizing.
- Improve basic auth handling.
- Continue Ingress processing on auth retrieval failure.

The last point is noteworthy: We previously returned an error when we couldn't retrieve basic auth credentials, blocking following Ingresses from being processed. Since retrieving the auth credentials is highly dependent on an external component (the Kubernetes Secret object), we should opt for dropping such failing Ingresses but move forward with the remaining ones.

Fixes #1794.

@dtomcej @errm @emilevauge @ldez and others: PTAL.

Issue #1784, opened 2017-06-26
Traefik tries to reach out to k8s Secrets even when basic auth isn't used


timoreimann:
### Do you want to request a *feature* or report a *bug*?
Bug

### What did you do?
Deploy Traefik on Kubernetes without basic authentication enabled through the `ingress.kubernetes.io/auth-type` annotation and RBAC not permitting access to Kubernetes Secrets.

### What did you expect to see?
Traefik running fine without complaints.

### What did you see instead?
Traefik continuously logging about being unable to access Secrets.

### Output of `traefik version`: (_What version of Traefik are you using?_)
latest master

### If applicable, please paste the log output in debug mode (`--debug` switch)
Here's the relevant portion from the logs:

```
E0625 21:28:59.568275       1 reflector.go:300] github.com/containous/traefik/vendor/k8s.io/client-go/tools/cache/reflector.go:94: Failed to watch *v1.Secret: the server does not allow access to the requested resource (get secrets)
```

Traefik should only start to query for Secrets when basic authentication requires it.

timoreimann:
@ldez @dtomcej @errm I wasn't sure if we want to consider the issue as a (confirmed) bug or enhancement, so I left out the relevant label.

dtomcej:
I guess as in #1785 the question is how we want to handle RBAC permissions. If we assume that traefik _should_ have full access, but might not, then we need to handle failure gracefully.

If we assume that traefik has no permissions, then we need to query the API for the permissions it does have, then operate on a limited scope based on that.

timoreimann:
I think the simplest approach would be to expect the Traefik user/administrator to know whether RBAC access to Secrets need to be granted (i.e., basic auth is going to be used) and to do so accordingly. Traefik's job would then be to get/watch for Secrets when we run into an Ingress object containing basic auth annotations _only_.

So I guess that matches your "Traefik should have full (Secrets) access" with the conditional "if we realize there's a need to access Secrets".

Note that a more sophisticated (but also more complex) approach that involves adapting the RBAC permissions dynamically is being discussed in #1626.

timoreimann:
I hope to get this fixed by Traefik 1.5.

To the least, the issue will be constrained to specific namespaces that Traefik operates in as the latest master has been switched over to a watch-per-namespace model, obsoleting the need to grant unnecessary _global_ RBAC access to secrets.

erkolson:
[...]

PR #1890, opened 2017-07-24
Add secrets resource to in-line RBAC spec.

timoreimann:
Previously only existed in the Github-hosted example file.

Fixes an oversight from #1707.

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -35,6 +35,7 @@ rules:
       - pods
       - services
       - endpoints
+      - secrets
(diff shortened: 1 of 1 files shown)


timoreimann:
@dtomcej you mean when we reverted #1798 through #1828? That's fairly unrelated as we only undid waiting for the secrets store to become populated (which would never happen in case RBAC access to Secrets wasn't set up).

We still need people to grant access to secrets to _guarantee_ basic auth works. #1784 should drive watching for secrets more selectively.

Does that answer your question?

PR #1906, opened 2017-07-31
Use default frontend priority of zero.

timoreimann:
Conventionally all providers should return a priority of zero if no override is given. The server package will then compute the priority based on the specified rules (i.e., path and host components). See also [here](https://github.com/containous/traefik/blob/94f922cd288ae5110d76be190e26d49d194fa6f9/server/server.go#L868-L870) and [here](https://github.com/containous/traefik/blob/94f922cd288ae5110d76be190e26d49d194fa6f9/server/server.go#L1021) in the server package.

The Kubernetes provider diverges and returns the path length as default value explicitly. This change removes the custom behavior.

As a side refactoring, we change the tests to stop comparing JSON values which are prone to subtle but non-visible differences (namely, nil and empty properties are not reflected). Instead, we use `assert.Equal` which also does a better job at showing a readable diff.

k8s maintainers and @containous/traefik: PTAL.

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -326,7 +326,7 @@ func getRuleForPath(pa v1beta1.HTTPIngressPath, i *v1beta1.Ingress) string {
 }
 
 func (p *Provider) getPriority(path v1beta1.HTTPIngressPath, i *v1beta1.Ingress) int {
-	priority := len(path.Path)
+	priority := 0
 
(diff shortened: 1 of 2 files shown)


timoreimann:
@dtomcej @errm as Kubernetes owners, are you ok with this PR getting merged?

ldez:
>  Waiting on code owner review from containous/kubernetes.

PR #1913, opened 2017-08-03
[ecs] Add support for several ECS backends

### LINKED ISSUE #1570: Add support for several ECS backends
Feature - I'd like traefik to support several ECS backends.

### What did you do?
Hello. I wanted to add several EC2 clusters, because I use one cluster per application. I think it's rather typical usecase.

 ### What did you see instead?
Traefik says that [ecs] section is allowed only once, so I cannot do that. Only way I see is to use one traefik per ECS backend, which isn't what I'd really want.

latest stable

mmatur:
### Description

The goal of this PR is to add support for several ECS backends.
Two new fields has been added for ECS provider:

- Clusters: used to define explicit list of clusters

- AutoDiscoverClusters: used to enable auto discover clusters functionality. If this property is set to true, Clusters property will be ignored. Træfik ECS policy need to be updated with "ecs:ListClusters", "ecs:DescribeClusters"

Fixes #1570 

To avoid breaking change ECS provider accept old field Cluster. If AutoDiscoverClusters is set Cluster field will be ignored.


provider/ecs/cluster.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+	"strings"
+)
+
+// Clusters holds ecs clusters name
+type Clusters []string
+
+//Set adds strings elem into the the parser
+//it splits str on , and ;
+func (ns *Clusters) Set(str string) error {
```
ldez: Could you add unit tests?

provider/ecs/cluster.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Clusters holds ecs clusters name
+type Clusters []string
+
+//Set adds strings elem into the the parser
```
ldez: could you add space between `//` and the comment. Same for others.

mmatur:
@ldez ECS provider use AWS API. Do you want that I mock AWS API to be able to do some integration tests ?

ldez:
Maybe you can use https://github.com/localstack/localstack?

mmatur:
I have already watch localstack. but currently they not provide ecs and ec2 endpoints. I will try to work on this new feature on localstack when i will have time

provider/ecs/ecs.go
```diff
@@ -200,103 +202,128 @@ func (p *Provider) loadECSConfig(ctx context.Context, client *awsClient) (*types
...
+		input := &ecs.ListClustersInput{}
+		result, _ := client.ecs.ListClusters(input)
+		for result.NextToken != nil {
+			clustersArn = append(clustersArn, result.ClusterArns...)
 
-		taskArns = append(taskArns, req.Data.(*ecs.ListTasksOutput).TaskArns...)
+			input.NextToken = result.NextToken
+			result, _ = client.ecs.ListClusters(input)
+		}
+		clustersArn = append(clustersArn, result.ClusterArns...)
```
nmengin: In the way to avoid repetition and possible problem if `result == nil`, what is your mind about this suggestion : 
```go
for {
    if result, _ := client.ecs.ListClusters(input); result != nil {
        clustersArn = append(clustersArn, result.ClusterArns...) 		  		 
	input.NextToken = result.NextToken
        if result.NextToken == nil {
            break
        }
    } else {
        break
    }
}

Comment (mmatur) [inline provider/ecs/ecs.go]: I think that result can never be nil. But I like the suggestion, this is simplifying the code. 
I will do the change

File: provider/ecs/ecs.go [hunk 4]
```diff
@@ -200,103 +202,131 @@ func (p *Provider) loadECSConfig(ctx context.Context, client *awsClient) (*types
...
+				break
+			}
 		}
-
-		taskArns = append(taskArns, req.Data.(*ecs.ListTasksOutput).TaskArns...)
+		for _, carns := range clustersArn {
+			clusters = append(clusters, *carns)
+		}
+	} else if p.Cluster != "" {
+		clusters = Clusters{p.Cluster}
```
emilevauge: Could you add a `log.warn` to inform the user that he uses a deprecated option? And also add a `// TODO` to remind us to remove this in the future?

provider/ecs/ecs.go
```diff
@@ -200,103 +202,131 @@ func (p *Provider) loadECSConfig(ctx context.Context, client *awsClient) (*types
...
-		if err := wrapAws(ctx, req); err != nil {
-			return nil, err
+	var instances []ecsInstance
+	var clustersArn []*string
+	var clusters Clusters
+
+	if p.AutoDiscoverClusters {
+		input := &ecs.ListClustersInput{}
+		for {
+			if result, _ := client.ecs.ListClusters(input); result != nil {
```
emilevauge: You should not discard the returned `error` here

PR #1915, opened 2017-08-03
Whitelists to work behind a proxy

joshdvir:
This PR allows the Whitelists check to check more properly what is the client IP address.

It checks 2 known headers being used by many proxy servers before using the RemoteAddr.

I implemented this since my Traefik is behind and AWS ELB LB.


dtomcej:
The premise of this one is good! I like the code implemented.

However, because X-Real-IP or X-Forwarded-For are arbitrary headers, doesn't that mean that someone from any IP can add a "safe" IP into the header, and poof, now all traffic from anywhere is allowed through?

I'm wondering if there should be an "Allow proxy" flag that will allow checking those extended headers, but not by default.

joshdvir:
@dtomcej Thanks for reviewing my PR.

'X-Real-IP' and 'X-Forwarded-For' are common headers used by most common reverse proxies out there to transfer the client IP to the applications so they weren't arbitrarily chosen.

I understand your concerns about faking the header, I think that the IP's in the whitelist should be known only inside the organization, but it's still a weak point.

Adding a flag that allows proxy can increase security in that form.

emilevauge:
@joshdvir out of curiosity, why don't you whitelist in AWS directly?
Not sure we need this in Traefik then.
Anyway, I think this should not be the default behavior, but only configured when traefik is used behind a L7 RP.

joshdvir:
Hi @emilevauge,

I'm using Traefik as a an ingress controller for K8s cluster, when the K8s cluster creates an ELB sadly I don't have control on the security group which means I have to do it manually when I create the LB and if there is a change in configuration it overwrites my changes and I need to remember to do them again.

In Traeik I set it up in the configuration files and it's applied automatically.

I think that checking the 'X-Forwarded-For' and 'X-Real-IP' headers only adds value to Traefik since it allows us to put Traefik behind another LB like ELB or even behind a CDN, when checking the IP only it limits the use cases for Traefik.

In my opinion it should be the default since it's a common practice in many applications that relay on these headers these days.

emilevauge:
I'm sorry, but I'm still not convinced. This is still a major security flaw if not behind another proxy. WDYT @containous/traefik ?

marco-jantke:
I think that this option is necessary for certain setups in which it also can be used safely, namely behind other reverse proxies that do not pass on those headers, but set them explicitly. So for me this changes make sense in some environments. This means it should be configurable, but as Emile mentioned not the default behaviour. 

The documentation should inform about the risks and maybe a typical environment where this behaviour is desired should be described.

keiths-osc:
We have been using traefik as an ingress controller for a kubernetes cluster in AWS and also have ELBs in front of our nodes.  For what it's worth, I would love to be able to use the new whitelist functionality but currently won't be able to because of this.  I would love to see this merged, but I also agree that it should be off by default.

joshdvir:
Guys i'm adding the functionality to enable the option to check headers which by default will be false.
Will update PR when done.

gempesaw:
Hullo, thanks for all your work on Traefik! :) As another use case in support, we also use Traefik behind an AWS LB. But, we can't just whitelist in AWS, because we want to configure the whitelisting on a per-ingress/per-application basis instead of using a single whitelist for all of our services.

However, I do strongly agree with everyone else that checking the headers should be disabled by default and only enabled by appropriate configs.

joe-bowman:
Any update on this @joshdvir? There appear to be a few of us in the same situation. If not I can give a go getting the change finished off. Thanks.

joshdvir:
I'm sorry everyone, my time is limited, if anyone wants to take a crack at it please do.

joe-bowman:
No problem. Happy to take a look :) thanks for your effort so far.

ldez:
Close in favor of  #2235

Issue #1575, opened 2017-05-09
Enable loadbalancer.sticky for Consul Catalog


bendenoz:
It looks like **sticky sessions** cannot be enabled through the **consul_catalog.tmpl**

What I did : set `traefik.backend.loadbalancer.sticky=true` on a consul service tag

What I expected: to see session cookie from traefik, to always use the same backend.
What I saw : no session cookie, each call using a different backend.

Thanks !

Evertvz:
I just figured out the same thing, which is very unfortunate. My go experience is limited, but I checked the docker provider since that one does support sticky sessions and found it reads the configuration from the containers here: https://github.com/containous/traefik/blob/master/provider/docker/docker.go

Now, aside from reading the flag out I don't see any other logic in there. Does this mean traefik can support sticky sessions if only the consul catalog provider would read the label and set the value for Traefik to use?

What would need to be adapted is this then:
https://github.com/containous/traefik/blob/master/provider/consul/consul_catalog.go
That doesn't sound too hard for a PR. I might give it a shot.

Can someone with some Traefik experience confirm this? @ldez ?
Do I base myself on master or v1.3?

dwarfaz:
Hi, 

in fact you can just modify https://github.com/containous/traefik/blob/master/templates/consul_catalog.tmpl

and add the sticky tag parsing in this section : 
  {{$loadBalancer := getAttribute "backend.loadbalancer" .Attributes ""}}
  {{with $loadBalancer}}
  [backends."backend-{{$service}}".loadbalancer]
    method = "{{$loadBalancer}}"
{{end}}

That's what we've done for our needs and it works like a sharm !

dwarfaz:
Here what we have done (not the best solution but it works) :
[backends."backend-{{$service}}".loadbalancer]
method = "{{getAttribute "backend.loadbalancer" .Attributes "wrr" }}"
sticky = {{getAttribute "backend.loadbalancer.sticky" .Attributes "false" }}

mmatur:
Fixed by #1917

PR #1917, opened 2017-08-04
Enable loadbalancer.sticky for Consul Catalog

nbonneval:
### Description

As commented by my colleague dwarfaz in the issue #1575, we added support for the backend.loadbalancer.sticky tag with consul_catalog by editing the provider's template.
It now always creates the backend's loadbalancer configuration section, with default values if nothing's specified, but supports changing both the load balancing method and the sticky flag.

In order to synchronize with other providers' capabilities & doc, I'd prefer the existing tag (backend.loadbalancer) to be renamed to backend.loadbalancer.method, but it would break existing configurations


nbonneval:
@nmengin Never thought about it, but why not. I did think about adding specific tests in provider/consul/consul_catalog_test.go though, but never took time to do it...

PR #1925, opened 2017-08-07
Enable loadbalancer.sticky for ECS

### LINKED ISSUE #1522: Support Session Stickiness in ECS

Feature Request

### What did you do?

Attempted to use `traefik.backend.loadbalancer.sticky=true` label on an ECS service but it was not honored.

Traefik backend would set sticky affinity and route traffic to the same container for that client.

### What did you see instead?

Traefik routed requests without honoring session stickiness.

```
Version:      v1.2.3
Codename:     morbier
Go version:   go1.7.5
Built:        2017-04-13_07:32:59PM
OS/Arch:      darwin/amd64
```

mmatur:
### Description

The goal of this PR is to add support of sticky session for ECS.
Low balancer configuration will be always present with default values:
`traefik.backend.loadbalancer.method=wrr`
`traefik.backend.loadbalancer.sticky=false`

To override these values you just to add labels 

Fixes #1522

### DIFF
--- a/provider/ecs/ecs.go
+++ b/provider/ecs/ecs.go
@@ -457,6 +457,20 @@ func (p *Provider) getFrontendRule(i ecsInstance) string {
 	return "Host:" + strings.ToLower(strings.Replace(i.Name, "_", "-", -1)) + "." + p.Domain
 }
 
+func (p *Provider) getLabelLoadBalancerSticky(i ecsInstance) string {
+	if label := i.label(types.LabelBackendLoadbalancerSticky); label != "" {
+		return label
+	}
+	return "false"
+}
+
+func (p *Provider) getLabelLoadBalancerMethod(i ecsInstance) string {
+	if label := i.label(types.LabelBackendLoadbalancerMethod); label != "" {
+		return label
+	}
+	return "wrr"
+}
+
 // Provider expects no more than 100 parameters be passed to a DescribeTask call; thus, pack
 // each string into an array capped at 100 elements
 func (p *Provider) chunkedTaskArns(tasks []*string) [][]*string {
--- a/provider/ecs/ecs.go
+++ b/provider/ecs/ecs.go
@@ -195,11 +195,11 @@ func (p *Provider) loadECSConfig(ctx context.Context, client *awsClient) (*types
 
 	services := make(map[string][]ecsInstance)
 
-	for _, i := range instances {
-		if serviceInstances, ok := services[i.Name]; ok {
-			services[i.Name] = append(serviceInstances, i)
+	for _, instance := range instances {
+		if serviceInstances, ok := services[instance.Name]; ok {
+			services[instance.Name] = append(serviceInstances, instance)
 		} else {
-			services[i.Name] = []ecsInstance{i}
+			services[instance.Name] = []ecsInstance{instance}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


ldez:
Could you fix the build?

> Errors from golint:
> provider/ecs/ecs.go:435:1: exported method Provider.LoadBalancerSticky should have comment or be unexported
> provider/ecs/ecs.go:444:1: exported method Provider.LoadBalancerMethod should have comment or be unexported
> 
> Please fix the above errors. You can test via "golint" and commit the result.

mmatur:
This PR is the implementation for feature requested on #1522

provider/ecs/ecs.go
```diff
@@ -420,6 +432,24 @@ func (p *Provider) getFrontendRule(i ecsInstance) string {
 	return "Host:" + strings.ToLower(strings.Replace(i.Name, "_", "-", -1)) + "." + p.Domain
 }
 
+func (p *Provider) getLoadBalancerSticky(instances []ecsInstance) string {
+	for instances != nil && len(instances) > 0 {
+		if label := instances[0].label(types.LabelBackendLoadbalancerSticky); label != "" {
```
ldez: could you split in two lines?

provider/ecs/ecs.go
```diff
@@ -420,6 +432,24 @@ func (p *Provider) getFrontendRule(i ecsInstance) string {
...
+			return label
+		}
+	}
+	return "false"
+}
+
+func (p *Provider) getLoadBalancerMethod(instances []ecsInstance) string {
+	for instances != nil && len(instances) > 0 {
+		if label := instances[0].label(types.LabelBackendLoadbalancerMethod); label != "" {
```
ldez: could you split in two lines?

provider/ecs/ecs.go
```diff
@@ -420,6 +432,24 @@ func (p *Provider) getFrontendRule(i ecsInstance) string {
 	return "Host:" + strings.ToLower(strings.Replace(i.Name, "_", "-", -1)) + "." + p.Domain
 }
 
+func (p *Provider) getLoadBalancerSticky(instances []ecsInstance) string {
+	for instances != nil && len(instances) > 0 {
```
ldez: `instances != nil` is unnecessary.

```go
func ExampleEmptySlice() {
	var foo []string
	fii := bar()
	fmt.Println(foo, fii)
	// output: [] []
}
func bar() []string{
	return nil
}
```

provider/ecs/ecs.go
```diff
@@ -420,6 +432,24 @@ func (p *Provider) getFrontendRule(i ecsInstance) string {
...
+		if label := instances[0].label(types.LabelBackendLoadbalancerSticky); label != "" {
+			return label
+		}
+	}
+	return "false"
+}
+
+func (p *Provider) getLoadBalancerMethod(instances []ecsInstance) string {
+	for instances != nil && len(instances) > 0 {
```
ldez: `instances != nil` is unnecessary.

```go
func ExampleEmptySlice() {
	var foo []string
	fii := bar()
	fmt.Println(foo, fii)
	// output: [] []
}
func bar() []string{
	return nil
}
```

provider/ecs/ecs.go
```diff
@@ -189,10 +191,20 @@ func (p *Provider) loadECSConfig(ctx context.Context, client *awsClient) (*types
 
 	instances = fun.Filter(p.filterInstance, instances).([]ecsInstance)
 
+	services := make(map[string][]ecsInstance)
+
+	for _, i := range instances {
```
ldez: could you rename `i` to `instance` ?

I know it's already use but I think `i` is an ambiguous name in a `for`.

provider/ecs/ecs.go
mmatur: Thanks for the go course. I will fix that asap

PR #1932, opened 2017-08-09
Added a check to ensure clientTLS configuration contains either a cert or a key

aantono:
### Description

Added a check to ensure `clientTLS` configuration contains either a cert or a key
If any of the `ClientTLS` only sets `InsecureSkipVerify` property to `true`, the `CreateTLSConfig()` method fails, because it expects a `Cert` or a `Key` property to always be present.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -157,24 +157,26 @@ func (clientTLS *ClientTLS) CreateTLSConfig() (*tls.Config, error) {
 	cert := tls.Certificate{}
 	_, errKeyIsFile := os.Stat(clientTLS.Key)
 
-	if _, errCertIsFile := os.Stat(clientTLS.Cert); errCertIsFile == nil {
-		if errKeyIsFile == nil {
-			cert, err = tls.LoadX509KeyPair(clientTLS.Cert, clientTLS.Key)
-			if err != nil {
-				return nil, fmt.Errorf("Failed to load TLS keypair: %v", err)
+	if len(clientTLS.Cert) > 0 && len(clientTLS.Key) > 0 {
+		if _, errCertIsFile := os.Stat(clientTLS.Cert); errCertIsFile == nil {
+			if errKeyIsFile == nil {
+				cert, err = tls.LoadX509KeyPair(clientTLS.Cert, clientTLS.Key)
+				if err != nil {
+					return nil, fmt.Errorf("Failed to load TLS keypair: %v", err)
+				}
+			} else {
+				return nil, fmt.Errorf("tls cert is a file, but tls key is not")
 			}
 		} else {
-			return nil, fmt.Errorf("tls cert is a file, but tls key is not")
-		}
-	} else {
-		if errKeyIsFile != nil {
-			cert, err = tls.X509KeyPair([]byte(clientTLS.Cert), []byte(clientTLS.Key))
-			if err != nil {
-				return nil, fmt.Errorf("Failed to load TLS keypair: %v", err)
+			if errKeyIsFile != nil {
+				cert, err = tls.X509KeyPair([]byte(clientTLS.Cert), []byte(clientTLS.Key))
+				if err != nil {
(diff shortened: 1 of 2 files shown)


provider/provider.go
```diff
@@ -164,24 +164,26 @@ func (clientTLS *ClientTLS) CreateTLSConfig() (*tls.Config, error) {
 	cert := tls.Certificate{}
 	_, errKeyIsFile := os.Stat(clientTLS.Key)
 
-	if _, errCertIsFile := os.Stat(clientTLS.Cert); errCertIsFile == nil {
-		if errKeyIsFile == nil {
-			cert, err = tls.LoadX509KeyPair(clientTLS.Cert, clientTLS.Key)
-			if err != nil {
-				return nil, fmt.Errorf("Failed to load TLS keypair: %v", err)
+	if len(clientTLS.Cert) > 0 && len(clientTLS.Key) > 0 {
```
nmengin: This confition will allow user to specify TLS configuration with no key r cert file even if the `InsecureSkipVerify` parameter is set to `false`.

It can be better to check the `InsecureSkipVerify` in the condition and let the current mechanism generate an error if cert file or key file are missing when `InsecureSkipVerify` is set to false.

```go
if ! clientTLS.InsecureSkipVerify {
...
}
```

aantono: The problem is that `CreateTLSConfig()` is being used from many places, where some are using this for outbound connections, for example Marathon provider uses this to create an http client to Mesos Master like:
```TLSConfig, err := p.TLS.CreateTLSConfig()
		if err != nil {
			return err
		}
		config.HTTPClient = &http.Client{
			Transport: &http.Transport{
				DialContext: (&net.Dialer{
					KeepAlive: time.Duration(p.KeepAlive),
					Timeout:   time.Duration(p.DialerTimeout),
				}).DialContext,
				TLSClientConfig: TLSConfig,
			},
		}
		client, err := marathon.NewClient(config)```
If you have Mesos Master using self-signed certificate, then all you need to configure is `InsecureSkipVerify=true`, but you don't have either a cert or a key to provide.  But there are other cases where you would have either a cert or a key, as well as `InsecureSkipVerify=true`. (but let me know if you think the above use-case is not accurate), I'd change the condition to just check for `if ! clientTLS.InsecureSkipVerify` instead.

Comment (nmengin) [inline provider/provider.go]: Thanks for your reply. 

In fact, I suggested to check the `InsecureSkipverify` parameter because, into the test you did, even if the `InsecureSkipVerify` is set to false, the test will be OK but the configuration is not correct (there is no certificate/key file to check) and I would like to avoid this kind of behavior.

The best solution may be to check the `InsecureSkipVerify` and the certificate/key files length before the test you did as described above : 

```go
if ! clientTLS.InsecureSkipVerify && (len(clientTLS.Cert) == 0 || len(clientTLS.Key) == 0) {
        return nil, fmt.Errorf("Certificate and key files must be set when TLS configuration is created") // The message must be improved ;)
}
if len(clientTLS.Cert) > 0 && len(clientTLS.Key) > 0 {
...
}
```

With this new condition, we are sure the TLSConfiguration will be created with no certificate/key file only if the `InsecureSkipVerify` is set to true otherwise, an error will be returned.

@aantono What is your mind about this solution?

provider/provider_test.go
```diff
@@ -228,6 +228,24 @@ func TestNilClientTLS(t *testing.T) {
...
+	}
+	config, err := provider.TLS.CreateTLSConfig()
+	if err != nil {
+		t.Fatal("CreateTLSConfig should assume that consumer does not want a TLS configuration if input is nil")
+	}
+	if !config.InsecureSkipVerify {
+		t.Fatal("CreateTLSConfig should support setting only InsecureSkipVerify property")
+	}
+}
+
```
nmengin: It can be great to add a test with the parameter `InsecureSkipverify` set to false and checking the error returned.

PR #1934, opened 2017-08-09
Add support for Query String filtering

driverpt:
As title says

Usage:

```
Query: foo=bar, bar=baz
```

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -109,6 +109,7 @@ Following is the list of existing matcher rules along with examples:
 - `PathPrefix: /products/, /articles/{category}/{id:[0-9]+}`: Match request prefix path. It accepts a sequence of literal and regular expression prefix paths.
 - `PathPrefixStrip: /products/`: Match request prefix path and strip off the path prefix prior to forwarding the request to the backend. It accepts a sequence of literal prefix paths. Starting with Traefik 1.3, the stripped prefix path will be available in the `X-Forwarded-Prefix` header.
 - `PathPrefixStripRegex: /articles/{category}/{id:[0-9]+}`: Match request prefix path and strip off the path prefix prior to forwarding the request to the backend. It accepts a sequence of literal and regular expression prefix paths. Starting with Traefik 1.3, the stripped prefix path will be available in the `X-Forwarded-Prefix` header.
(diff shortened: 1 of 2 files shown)


driverpt:
:shipit: !!!!

ldez:
https://github.com/containous/traefik/blob/master/.github/MAINTAINER.md#pr-review-process

docs/basics.md
```diff
@@ -109,6 +109,7 @@ Following is the list of existing matcher rules along with examples:
 - `PathPrefix: /products/, /articles/{category}/{id:[0-9]+}`: Match request prefix path. It accepts a sequence of literal and regular expression prefix paths.
 - `PathPrefixStrip: /products/`: Match request prefix path and strip off the path prefix prior to forwarding the request to the backend. It accepts a sequence of literal prefix paths. Starting with Traefik 1.3, the stripped prefix path will be available in the `X-Forwarded-Prefix` header.
 - `PathPrefixStripRegex: /articles/{category}/{id:[0-9]+}`: Match request prefix path and strip off the path prefix prior to forwarding the request to the backend. It accepts a sequence of literal and regular expression prefix paths. Starting with Traefik 1.3, the stripped prefix path will be available in the `X-Forwarded-Prefix` header.
+- `Query: foo=bar, bar=baz`: Match Query String parameters. It accepts a sequence of key=value pairs.
```
Juliens: It accepts some regexp too ? 
http://www.gorillatoolkit.org/pkg/mux#Route.Queries

driverpt: For now it doesn't. Currently it splits the = sign and analyzes each part. I can do a PR in the future supporting Regex but i need to figure out how.

ztl8702: Actually, it does. 
I just tested:  `Query: a={a:[0-1]+}` will match `http://foobar/?a=1` but not `http://foobar/?a=2` for example. It exactly as Juliens pointed out: such Regex syntax is implemented in the underlying `mux` package, by just splitting the `=` sign driverpt's commit actually silently enables this feature.  https://github.com/containous/mux/blob/06ccd3e75091eb659b1d720cda0e16bc7057954c/route.go#L387

Should update the docs I guess?

PR #1944, opened 2017-08-11
Do not run integration tests by default.

timoreimann:
A common-practice development workflow of many editors and IDEs is to execute tests every time a file is saved. For the integration tests, however, this is usually infeasible due to our heavy usage of containers and docker-compose. Another argument against immediate execution on host machines is that some tests do not work outside of our test container, with Mac OS X in particular being traditionally difficult.

This change adds a build flag `integration` to every integration test file, thereby preventing automatic test execution. Running them explicitly is as easy as adding the necessary build flag to the `go test` command:

```shell
go test -tags=integration ./integration
```

The test-integration script has been updated to always include the test.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -143,7 +143,7 @@ Unit tests can be run from the cloned directory by `$ go test ./...` which shoul
 ok      _/home/user/go/src/github/containous/traefik    0.004s
 ```
 
-Integration tests must be run from the `integration/` directory and require the `integration` build tag to be specified like this: `$ cd integration && go test -tag integration ./...`.
+Integration tests must be run from the `integration/` directory and require the `-integration` switch to be passed like this: `$ cd integration && go test -integration ./...`.
 
 ## Documentation
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -138,10 +138,12 @@ More: https://labix.org/gocheck
 
 #### Method 2: `go`
 
-- Tests can be run from the cloned directory, by `$ go test ./...` which should return `ok` similar to:
+Unit tests can be run from the cloned directory by `$ go test ./...` which should return `ok` similar to:
 ```
 ok      _/home/user/go/src/github/containous/traefik    0.004s
 ```
+
+Integration tests must be run from the `integration/` directory and require the `integration` build tag to be specified like this: `$ cd integration && go test -tag integration ./...`.
 
 ## Documentation
 
--- a/integration/access_log_test.go
+++ b/integration/access_log_test.go
@@ -1,3 +1,5 @@
+// +build integration
+
 package integration
 
 import (
--- a/integration/access_log_test.go
+++ b/integration/access_log_test.go
@@ -1,5 +1,3 @@
-// +build integration
-
 package integration
 
 import (
--- a/integration/access_log_test.go
+++ b/integration/access_log_test.go
@@ -25,10 +25,6 @@ const (
 type AccessLogSuite struct{ BaseSuite }
 
 func (s *AccessLogSuite) TestAccessLog(c *check.C) {
-	if !*integration {
-		c.Skip("skipping integration tests")
-	}
-
 	// Ensure working directory is clean
 	os.Remove(traefikTestAccessLogFile)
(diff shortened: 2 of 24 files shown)


timoreimann:
@containous/traefik let me know what you think of this PR. I personally got massively tired by having to toggle automatic test execution on/off in my IDE depending on whether I was working on unit or integration tests in Traefik. Forgetting to turn it off for the integration tests would sometimes even cause my machine to lock up, so I believe some kind of safety net would make sense. I think the trade-off in convenience is acceptable.

We may also want to extend the documentation a bit if this PR finds approval.

timoreimann:
@vdemeester IIUC, for `-short` to work we'd have to amend all tests by a 3-4 liner to ignore the test if the flag is set. IMHO, this is most useful when you have a mixture of short and non-short tests in the same file and you want to be able to distinguish on a more fine-grained level.

Since we keep all integration tests in the same file (even the same directory), however, I think for us the build tag is a better fit.

vdemeester:
> @vdemeester IIUC, for -short to work we'd have to amend all tests by a 3-4 liner to ignore the test if the flag is set. IMHO, this is most useful when you have a mixture of short and non-short tests in the same file and you want to be able to distinguish on a more fine-grained level.

@timoreimann yeah, that too 😛 the build tag seem a better idea 😉

timoreimann:
I only have a rough idea what a configuration is. Could you please elaborate on the implications to your workflow? How did you work with integration tests before? What would change through this PR?

ldez:
I know what configuration must be change.
I don't know if we want this change must be requirement for Gogland user.

timoreimann:
Playing a bit further with the build tag, I noticed one nuisance: With Visual Studio Code, it's not possible to just build the integration test code easily. While I can configure VSC to not pass the build tag to the `go test` call, the IDE insists on building the test files indirectly through a call to `go test` when the `testOnSave` option is enabled. This may be a similar problem to what @ldez [reported](https://github.com/containous/traefik/pull/1944#issuecomment-324103262).

While I could live with that by switching to the CLI for building the code, it's not ideal. So I decided to explore @vdemeester's initial suggestion to consider the `testing.Slow()` switch. Turned out it's not easily integratable since we're using _go-check_ as our testing library for the integration tests. Fortunately, go-check comes with its own approach to skipping tests by defining a simple flag and checking its value in the `SetUpSuite` method. Unlike `testing.Slow()`, this means we'll only have to add one check per test suite, eliminating the noisiness [I was referring to](https://github.com/containous/traefik/pull/1944#issuecomment-322136909) before. Although it's still a few more lines than a build tag, but I think it's acceptable.

So overall, I like the skip/flag approach more. I pushed my latest code for review.

@emilevauge @ldez @vdemeester and others, WDYT? Which variant should we go for?

ldez:
maybe you can do that:

```go
// This is the main file that sets up integration tests using go-check.
package integration

var integration = flag.Bool("integration", false, "run integration tests")

//...

func init() {
	if *integration {
		check.Suite(&AccessLogSuite{})
		check.Suite(&AcmeSuite{})
		check.Suite(&ConstraintSuite{})
		check.Suite(&ConsulCatalogSuite{})
		check.Suite(&ConsulSuite{})
		check.Suite(&DockerSuite{})
		check.Suite(&DynamoDBSuite{})
		check.Suite(&ErrorPagesSuite{})
		check.Suite(&EtcdSuite{})
		check.Suite(&EurekaSuite{})
		check.Suite(&FileSuite{})
		check.Suite(&GRPCSuite{})
		check.Suite(&HealthCheckSuite{})
		check.Suite(&HTTPSSuite{})
		check.Suite(&LogRotationSuite{})
		check.Suite(&MarathonSuite{})
		check.Suite(&MesosSuite{})
		check.Suite(&SimpleSuite{})
		check.Suite(&TimeoutSuite{})
		check.Suite(&WebsocketSuite{})	
	}
}
//...
```

timoreimann:
@ldez good catch, that seems to work! Let's see what the CI says...

I just inverted the logic so that I could log a message saying that integration tests are disabled, which will make it more transparent to the user why they didn't run.

timoreimann:
Semaphore CI seems happy running all integration tests.

@ldez PTAL again.

PR #1946, opened 2017-08-11
log X-Forwarded-For as ClientHost if present

mildis:
Log X-Forwarded-For value as ClientHost, if present.
Allow access log to display the correct caller behind a reverse proxy.


emilevauge:
@mildis Thanks for your contribution!
@rjshep as the author of this part, LGTY ?

marco-jantke:
For me this change makes sense. AFAIK this is what is usually required from applications that sit behind a proxy and Traefik could also sit behind another proxy :D

Jancis:
I don't think this checks the `forwardedHeaders.trustedIPs` and so the access log can be filled with forged addresses.

PR #1956, opened 2017-08-14
Code cleaning.

ldez:
### Description

Clean some code smells.

Analyze made with:
```shell
go get -u honnef.co/go/tools/cmd/{gosimple,staticcheck}

staticcheck $(go list ./... | grep -v '/vendor/')
gosimple $(go list ./... | grep -v '/vendor/')
```

I have not fixed some smells with chan, need more thinking 😉 

I think we must add those tools to our build chain.

### DIFF
--- a/middlewares/authenticator_test.go
+++ b/middlewares/authenticator_test.go
@@ -66,14 +66,14 @@ func TestAuthUsersFromFile(t *testing.T) {
 }
 
 func TestBasicAuthFail(t *testing.T) {
-	authMiddleware, err := NewAuthenticator(&types.Auth{
+	_, err := NewAuthenticator(&types.Auth{
 		Basic: &types.Basic{
 			Users: []string{"test"},
 		},
 	})
 	assert.Contains(t, err.Error(), "Error parsing Authenticator user", "should contains")
 
-	authMiddleware, err = NewAuthenticator(&types.Auth{
+	authMiddleware, err := NewAuthenticator(&types.Auth{
 		Basic: &types.Basic{
 			Users: []string{"test:test"},
 		},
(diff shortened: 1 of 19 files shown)


middlewares/authenticator_test.go
```diff
@@ -161,12 +161,12 @@ func TestBasicAuthUserHeader(t *testing.T) {
 		Basic: &types.Basic{
 			Users: []string{"test:$apr1$H6uskkkW$IgXLP6ewTrSuBkTrqE8wj/"},
 		},
-		HeaderField: "X-WebAuth-User",
+		HeaderField: "X-Webauth-User",
```
timoreimann: I suppose it was intentionally to lower-case here and that it won't break anything.

middlewares/error_pages_test.go
```diff
@@ -19,13 +20,15 @@ func TestErrorPage(t *testing.T) {
 	defer ts.Close()
 
 	testErrorPage := &types.ErrorPage{Backend: "error", Query: "/test", Status: []string{"500-501", "503-599"}}
+
 	testHandler, err := NewErrorPagesHandler(*testErrorPage, ts.URL)
+	require.NoError(t, err, "Should be no error")
```
timoreimann: Side note: I don't think the extra message is really needed here and with the other (changed) `NoError` calls.

middlewares/retry_test.go
```diff
@@ -41,8 +41,7 @@ func TestRetry(t *testing.T) {
 		t.Run(tcName, func(t *testing.T) {
 			t.Parallel()
 
-			var httpHandler http.Handler
-			httpHandler = &networkFailingHTTPHandler{failAtCalls: tc.failAtCalls, netErrorRecorder: &DefaultNetErrorRecorder{}}
+			var httpHandler http.Handler = &networkFailingHTTPHandler{failAtCalls: tc.failAtCalls, netErrorRecorder: &DefaultNetErrorRecorder{}}
```
timoreimann: Why not `httpHandler :=` directly?

middlewares/authenticator_test.go
ldez: I respect "canonical format" required by header.
It's just a fake header.

middlewares/retry_test.go
ldez: because `http.Handler` type is required by `NewRetry()` in the next line:
```go
httpHandler = NewRetry(tc.attempts, httpHandler, tc.listener)
```

PR #1960, opened 2017-08-16
Support X-Forwarded-Port.

### LINKED ISSUE #1220: Add X-Forwarded-Port header to incoming requests
Version:      v1.2.0-rc1
Codename:     morbier
Go version:   go1.7.5
Built:        2017-02-06_08:45:49PM
OS/Arch:      linux/amd64

File-based config on AWS Ubuntu instances.
`passHostHeader = true` on backend.

### What did you do?
When running behind a reverse-proxy, some popular web frameworks such as Symfony depend on various `X-Forwarded-*` headers to generate correct URL's, etc. Traefik correctly sets most of these headers, except for `X-Forwarded-Port`. Symfony requires that header to be set.

I expect Traefik to add an `X-Forwarded-Port` header when passing the request to a backend, in addition to the other `X-Forwarded-*` headers it already sets. The `X-Forwarded-Port` header contains the portnumber on which the incoming request was received, usually 80 or 443.

I'm now adding this header in my backend, but it should be done by the proxy since that's the only place where the actual portnumber is known.

### What did you see instead?
The following headers are already added by Traefik.
`X-Forwarded-For`
`X-Forwarded-Host`
`X-Forwarded-Proto`
`X-Forwarded-Server`
`X-Traefik-Reqid`

ldez:
### Description

Fixes #1220

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: b056b388f961ddc3509b19b4546d8dfc70fd50ebd1ca47f2fd2de67bf49ba01e
-updated: 2017-08-12T14:15:06.346751095+02:00
+hash: b970da88f72aebed9ac07a31db3f40045de0cd63190a2aae8ce18b55ec3f434a
+updated: 2017-08-21T14:15:06.346751095+02:00
 imports:
 - name: cloud.google.com/go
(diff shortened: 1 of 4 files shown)


shouze:
🎉

Issue #1958, opened 2017-08-15
Support Named/Tagged Traefiks


ryanleary:
### Feature Proposal
This is a feature request/proposal to support, in a configuration file, to support traefik instances to be named or tagged. The intended use case here is to have multiple traefik instances that support different subsets of frontends/backends, particularly when used with backends other than the file backend.

### Current Behavior
We have a set of traefiks running internally for internal load-balancing, and another set for public load-balancing. For example, when connecting to a Marathon backend, all applications with `traefik.enable=true` specified are added to both our internal and external traefik instances.

### Proposed Behavior
Propose adding a `name` or `tag` for traefik instances within the file configuration and the introduction of a new label parameter. The new label would be `traefik.$tag.enable`. This would allow applications to use either a global `traefik.enable` label or `traefik.$tag.enable`. 

If this is something that seems useful to others and would be accepted, I will work on an implementation, but wanted to first gather some feedback on the concept/idea.

timoreimann:
@ryanleary is there a particular provider you'd like the described feature to be available for?

At least for Marathon and Kubernetes, what you are describing seems already possible using a label selector and [constraints](http://docs.traefik.io/toml/#constraints), respectively. I should mention though that constraints are fairly under-documented, but I'm happy to provide details and/or extend the docs.

ryanleary:
@timoreimann thanks for the reply. I think that would likely do the trick for my use case. I'm using Marathon. I didn't see Marathon as a supported backend -- guess I can dig into the code.

ryanleary:
Digging through the code I see that this is in fact supported. Thanks for the pointer.

timoreimann:
@ryanleary I also noticed that the docs aren't up-to-date but forgot about it again. Thanks for the reminder, I'll make sure we fix that.

timoreimann:
@ldez not sure if we should pick a `resolution/*` label. I'll leave that to you.

PR #1964, opened 2017-08-16
Mark Marathon and Kubernetes as constraint-supporting.

timoreimann:
Discovered along #1958.

@containous/traefik PTAL.


ryanleary:
Suggest adding that the overriding label for containers is `traefik.tags`. I can open a separate PR for that if you think it's out of scope here.

ldez:
@ryanleary I think you can open PR 👍

PR #1850, opened 2017-07-10
extend and refactor prometheus metrics

marco-jantke:
For our internal Traefik setup I was working on improving the Prometheus metrics and wanted to upstream my work, as I guess this is potentially useful for a lot of Traefik users. It is planned to integrate more metrics in the future, but this PR already restructures the metrics implementation and provides significant benefits when monitoring Traefik with Prometheus. For more metrics  follow-up PRs will be smaller and more focused. It is important to note that in this rework, the metric naming was adapted in order to fix #1759 and to be more consistent. As the metrics implementation is in a rather young state, I had the feeling this is the right time to go for removing inconsistencies.

In general the following metrics are now available. Please check the code for details including the metric labels.
```
// server meta information
traefik_config_reloads_total
traefik_config_reload_failures_total
traefik_config_last_reload_success

// entrypoint
traefik_entrypoint_requests_total
traefik_entrypoint_request_duration_seconds

// backend level
traefik_backend_requests_total
traefik_backend_request_duration_seconds
traefik_backend_retries_total
```

In the course of extending the metrics I did some design reworks:

- I moved the implementation to an own package. This is necessary because some metrics get also collected in the server package and it makes the design more clear.
- I switched the prometheus test to a real integration test. On the one hand this made things easier on the other its a more "real world test".
- I changed the `Metrics` interface to one big interface. Even though this might not be the most idiomatic way in go, it makes it sufficiently easier for other potential future implementations to provide a "full" metrics implementation. Also the `VoidMetrics` and the general handling in the `server.go` is much easier through this.
- I integrated a `VoidMetrics` implementation to remove the necessity to check for nil on the consumer side.

PR #1968, opened 2017-08-17
Extract metrics to own package and refactor implementations

marco-jantke:
This PR should be the first step to an extension of the metrics implementations. In our Traefik fork that we are running in my project we have extended the Prometheus implementation considerably, but to get there some refactoring and alignment has to be done first. I created once already a [PR](https://github.com/containous/traefik/pull/1850) for the extension of the Prometheus implementation, but it was diverging after Datadog and StatsD implementations were added at the same time to Traefik.

Therefore this PR is only about refactoring of the structure and extracting the Metrics implementations into their own package. Once the structure is aligned I will create follow-up metrics extending the functionality. I think it is a better approach this way, as it should make reviewing considerably easier.

### Motivation

This PR is changing the design and I want to elaborate a bit in what it changes it and why:
- _extraction into a new `metrics` package_: This is important in order to be able to track metrics outside of the `middleware` package. Meaningful metrics could be for example configuration reloads. Those happen in the `server` package and so an implementation inside of `middleware` can not be used (cyclic dependency). Another positive point about the extraction is the better separation of concerns in regards and a more clear project structure.
- _initialisation of metrics implementation_: They all happen now exactly once in `NewServer` when Traefik boots up, which is the desired behaviour. Through the alignment of the initialisation the logic is now more straightforward and easier to understand.


marco-jantke:
@aantono I think this is also interesting for you, I would be glad if you can have a look too :)

metrics/datadog_test.go
```diff
@@ -0,0 +1,37 @@
+package metrics
+
+import (
+	"net/http"
```
ldez: could you reorganize the imports?

aantono:
@marco-jantke I think the name change for the label from `serviceName` to `service` should not be a big deal... but out of curiosity, why the change?

metrics/datadog_test.go
marco-jantke: Thanks for the notice, done.

marco-jantke:
@aantono my fault I mixed it up. It was `service` all the time. Sorry for the confusion :/

metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return &multiRegistry{
+		reqsCounter:          multi.NewCounter(reqsCounters...),
+		reqDurationHistogram: multi.NewHistogram(reqDurationHistograms...),
+		retriesCounter:       multi.NewCounter(retriesCounters...),
+	}
+}
+
+// multiRegistry is a Registry that wraps multiple Registries and calls each of them when a Metric is tracked.
+type multiRegistry struct {
```
ldez: Can you rename `multiRegistry` to `standardRegistry`?
This implementation is not specific to the multiple Registries.

metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+	reqsCounter          metrics.Counter
+	reqDurationHistogram metrics.Histogram
+	retriesCounter       metrics.Counter
+}
+
+func (r *multiRegistry) ReqsCounter() metrics.Counter {
+	return r.reqsCounter
+}
+
+// ReqsCounter
```
ldez: remove or add the good comment

metrics/prometheus_test.go
```diff
@@ -0,0 +1,100 @@
...
+	for _, test := range tests {
+		family := findMetricFamily(test.name, metricsFamilies)
+		if family == nil {
+			t.Errorf("gathered metrics do not contain '%s'", test.name)
+			continue
+		}
+		for _, label := range family.Metric[0].Label {
+			val, ok := test.labels[*label.Name]
+			if !ok {
+				t.Errorf("'%s' metric contains unexpected label '%s'", test.name, label)
```
ldez: could you replace `'%s'` by `%q`?

metrics/prometheus_test.go
```diff
@@ -0,0 +1,100 @@
...
+			} else if val != *label.Value {
+				t.Errorf("label '%s' in metric '%s' has wrong value '%s'", label, test.name, *label.Value)
```
ldez: could you replace `'%s'` by `%q`?

metrics/prometheus_test.go
```diff
@@ -0,0 +1,100 @@
...
+					t.Errorf("gathered metrics do not contain correct value for total retries, got '%d' expected '%d'", cv, expectedCv)
+				}
+			},
+		},
+	}
+
+	for _, test := range tests {
+		family := findMetricFamily(test.name, metricsFamilies)
+		if family == nil {
+			t.Errorf("gathered metrics do not contain '%s'", test.name)
```
ldez: could you replace `'%s'` by `%q`?

metrics/prometheus_test.go
```diff
@@ -0,0 +1,100 @@
...
+		{
+			name: retriesTotalName,
+			labels: map[string]string{
+				"service": "test",
+			},
+			assert: func(family *dto.MetricFamily) {
+				cv := uint(family.Metric[0].Counter.GetValue())
+				expectedCv := uint(1)
+				if cv != expectedCv {
+					t.Errorf("gathered metrics do not contain correct value for total retries, got '%d' expected '%d'", cv, expectedCv)
```
ldez: could you replace `'%d'` by `%q`?

metrics/metrics.go
marco-jantke: Good idea, adapted the implementation accordingly.

metrics/metrics.go
marco-jantke: Added it by accident. Removed the comment. I think they are not necessary on "Getters" and the method's struct is not exported.

metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func (r *multiRegistry) RetriesCounter() metrics.Counter {
+	return r.retriesCounter
+}
+
+// NewVoidRegistry is a noop implementation of metrics.Registry.
+// It is used to avoid nil checking in components that do metric collections.
+func NewVoidRegistry() Registry {
+	return &voidRegistry{
```
marco-jantke: Thanks, I am using the `standardRegistry` now.

metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+		c: &voidCounter{},
+		h: &voidHistogram{},
+	}
+}
+
+type voidRegistry struct {
```
marco-jantke: Thanks, I am using the `standardRegistry` now.

metrics/metrics_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func TestNewVoidRegistry(t *testing.T) {
+	registry := NewVoidRegistry()
+
+	registry.ReqsCounter().With("some", "value").Add(1)
+	registry.ReqDurationHistogram().With("some", "value").Observe(1)
+	registry.RetriesCounter().With("some", "value").Add(1)
+}
+
+func TestNewMutliRegistry(t *testing.T) {
```
marco-jantke: Thanks, fixed.

metrics/metrics_test.go
```diff
@@ -0,0 +1,100 @@
...
+	registry.ReqDurationHistogram().With("key", "durations").Observe(2)
+	registry.RetriesCounter().With("key", "retries").Add(3)
+
+	for _, collectingRegistry := range registries {
+		cReqsCounter := collectingRegistry.ReqsCounter().(*collectingCounter)
+		cReqDurationHistogram := collectingRegistry.ReqDurationHistogram().(*collectingHistogram)
+		cRetriesCounter := collectingRegistry.RetriesCounter().(*collectingCounter)
+
+		if cReqsCounter.counterValue != 1 {
+			t.Errorf("Got value %v for ReqsCounter, want %v", cReqsCounter.counterValue, 1)
```
marco-jantke: Good idea. I also extracted the `1` to a value with proper float64 type, so that I can use `%f` both times.

metrics/metrics_test.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		if cReqDurationHistogram.lastHistogramValue != 2 {
+			t.Errorf("Got last observation %v for ReqDurationHistogram, want %v", cReqDurationHistogram.lastHistogramValue, 2)
```
marco-jantke: Thanks, done.

metrics/metrics_test.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+		if cRetriesCounter.counterValue != 3 {
+			t.Errorf("Got value %v for RetriesCounter, want %v", cRetriesCounter.counterValue, 3)
```
marco-jantke: Thanks, done.

metrics/metrics_test.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		assert.Equal(t, []string{"key", "requests"}, cReqsCounter.lastLabelValues)
+		assert.Equal(t, []string{"key", "durations"}, cReqDurationHistogram.lastLabelValues)
+		assert.Equal(t, []string{"key", "retries"}, cRetriesCounter.lastLabelValues)
+	}
+}
+
+// collectingRegistry is an implementation of themetrics.Registry interface that can be used inside tests.
+type collectingRegistry struct {
```
marco-jantke: I use `standardRegistry` now.

metrics/metrics_test.go
```diff
@@ -0,0 +1,100 @@
...
+	reqsCounter          *collectingCounter
+	reqDurationHistogram *collectingHistogram
+	retriesCounter       *collectingCounter
+}
+
+func newCollectingRetryMetrics() *collectingRegistry {
+	return &collectingRegistry{
```
marco-jantke: I use `standardRegistry` now.

metrics/metrics_test.go
```diff
@@ -0,0 +1,100 @@
...
+func (c *collectingCounter) With(labelValues ...string) metrics.Counter {
+	c.lastLabelValues = labelValues
+	return c
+}
+
+func (c *collectingCounter) Add(delta float64) {
+	c.counterValue += delta
+}
+
+type collectingHistogram struct {
```
marco-jantke: Done.

metrics/prometheus.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return &prometheusRegistry{
+		reqCounter:           reqCounter,
+		reqDurationHistogram: reqDurationHistogram,
+		retryCounter:         retryCounter,
+	}
+}
+
+type prometheusRegistry struct {
+	reqCounter           metrics.Counter
```
marco-jantke: Obsolete, thanks to `standardRegistry`.

metrics/datadog.go
```diff
@@ -0,0 +1,87 @@
...
+	registry := &datadogRegistry{
+		reqsCounter:          datadogClient.NewCounter(ddMetricsReqsName, 1.0),
+		reqDurationHistogram: datadogClient.NewHistogram(ddMetricsLatencyName, 1.0),
+		retriesCounter:       datadogClient.NewCounter(ddRetriesTotalName, 1.0),
+	}
+
+	return registry
+}
+
+type datadogRegistry struct {
```
marco-jantke: Thanks, done.

metrics/prometheus.go
```diff
@@ -0,0 +1,63 @@
...
+	}, []string{"service"})
+
+	return &prometheusRegistry{
+		reqCounter:           reqCounter,
+		reqDurationHistogram: reqDurationHistogram,
+		retryCounter:         retryCounter,
+	}
+}
+
+type prometheusRegistry struct {
```
marco-jantke: I agree.

metrics/statsd.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	return &statsdRegistry{
+		reqsCounter:          statsdClient.NewCounter(ddMetricsReqsName, 1.0),
+		reqDurationHistogram: statsdClient.NewTiming(ddMetricsLatencyName, 1.0),
+		retriesCounter:       statsdClient.NewCounter(ddRetriesTotalName, 1.0),
+	}
+}
+
+type statsdRegistry struct {
```
marco-jantke: Thanks, done.

metrics/prometheus_test.go
marco-jantke: Doesn't sound right to me. When I change this it gets from:
`gathered metrics do not contain correct value for total requests, got 2 expected 23`
to
`gathered metrics do not contain correct value for total requests, got '\x02' expected '\x17'`.

Maybe though I should rather use floats for comparison and print messages with `%f`. WDYT?

metrics/prometheus_test.go
marco-jantke: Yes, done.

metrics/prometheus_test.go
marco-jantke: Good point. Adapted the error message.

metrics/prometheus_test.go
marco-jantke: Good point. Adapted and improved the error message.

metrics/metrics_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (r *collectingRegistry) ReqDurationHistogram() metrics.Histogram {
+	return r.reqDurationHistogram
+}
+
+func (r *collectingRegistry) RetriesCounter() metrics.Counter {
+	return r.retriesCounter
+}
+
+type collectingCounter struct {
```
marco-jantke: Ok makes sense. Done.

marco-jantke:
Thanks for the review @ldez. I rebased the branch to latest master version and addressed all your comments.

ldez:
@marco-jantke 
> These files are not properly gofmt'd:
>  - server/server.go

metrics/prometheus_test.go
ldez: 👍

marco-jantke: Ok, done.

server/server.go
```diff
@@ -53,6 +54,8 @@ type Server struct {
 	routinesPool                  *safe.Pool
 	leadership                    *cluster.Leadership
 	defaultForwardingRoundTripper http.RoundTripper
+	metricsRegistry               metrics.Registry
+	metricsEnabled                bool
```
emilevauge: I'm not fan of adding this flag. Can't we remove it and use `metricsRegistry` instead?

marco-jantke: `metricsRegistry` now always contains an implementation. When nothing is configured it will hold a `VoidRegistry`. This means we could write a util function like `metricsEnabled()` and it will return the `ok` of a type assertion on the metrics implementation. You would prefer this way?

Personally I am not sure about what is the more explicit option, but both is not hard to implement so you are free to choose :D

marco-jantke: I double checked the implementation and my described approach above won't work. All different registries that we create through e.g. `NewVoidRegistry`, `RegisterStatsd`, `RegisterPrometheus`.. will have the internal type `standardRegistry` and therefore I can't use a type assertion.

This means a `metricsEnabled()` utility would have to check on the global configuration once more. E.g. `if globalConfiguration.Web.Metrics.Prometheus != nil || globalConfiguration.Web.Metrics.Statsd != nil ...`. Personally I think that having this conditions only in `registerMetricClients` is the more clean approach and tend to leave it like it is. WDYT?

marco-jantke: As discussed on Slack refactored now towards an `IsEnabled()` method on the `Registry`.

metrics/metrics.go
```diff
@@ -0,0 +1,79 @@
...
+	return &standardRegistry{
+		isEnabled:            true,
+		reqsCounter:          multi.NewCounter(reqsCounters...),
+		reqDurationHistogram: multi.NewHistogram(reqDurationHistograms...),
+		retriesCounter:       multi.NewCounter(retriesCounters...),
+	}
+}
+
+type standardRegistry struct {
+	isEnabled            bool
```
ldez: could you rename `isEnabled` to `enable`?
But keep the method name `IsEnabled()`

ldez: sorry `enabled` not `enable`

marco-jantke: Sure thing, done.

Issue #1942, opened 2017-08-11
Enable graceful and asynchronous pool rotation with downstream load-balancers


timoreimann:
### Do you want to request a *feature* or report a *bug*?
feature.

### What did you do?
Do a rolling-upgrade of several traefik instances all receiving a constant stream of requests over an external load-balancer.

### What did you expect to see?
No requests are being lost/dropped.

### What did you see instead?
Requests _are_ being lost/dropped.

### Output of `traefik version`: (_What version of Traefik are you using?_)
master.

### What is your environment & configuration (arguments, toml, provider, platform, ...)?
Provider: Marathon and Kubernetes.

### Feature Request
Extend Traefik to support smooth, request-lossless pool rotation with downstream load-balancers during termination and rolling-restart phases in an asynchronous, decoupled manner. Overall, this requires two rather smallish actions to be enabled once Traefik receives a SIGTERM:

1. Toggle a pre-defined health check to artificially return unhealthy responses. This could be as simple as making an HTTP endpoint return 503s instead of 200s. Whether the endpoint should be freely choosable or rather the reuse of an existing endpoint (most likely `/ping`) is to be discussed.
1. Keep accepting requests for a configurable time (usually on the order of seconds to minutes). Once the time elapses, Traefik should proceed with the shutdown procedure as implemented today already: Invoke Go's graceful HTTP shutdown method stopping all listeners and finishing off all in-flight requests.

The two options should be configurable independently because not all setups may require both steps to be taken.

### Rationale

Traefik (and proxies in general) is often set up as a pool of instances which receive traffic from external, first-level load-balancer, often living in the cloud or in hardware (e.g., F5). Most of these LBs use some means of health checking to determine if their upstream servers are still alive and ready to receive traffic. Once such checks fail frequently enough, the affected servers are being removed from rotation until the health check goes green again.

When it comes to terminating Traefik gracefully, special care must be taken in order to avoid losing requests received from downstream at an unfortunate moment in time. Terminating hard is not an option since LBs likely won't be able to remove the server from rotation in time, leading at best to request retries (along with the induced extra latency) and at worst to requests getting dropped or rejected. Some (though by far not all) LBs provide APIs to orchestrate the shutdown procedure. Even if they do, however, a proper implementation tends to be complex and error-prone due to the tightly coupled and synchronous, state machine-like nature of things. On top of that, implementations are usually LB-specific and thus cannot be transferred over to other LB vendors.

[...]

PR #1971, opened 2017-08-18
Add request accepting grace period delaying graceful shutdown.

timoreimann:
The new grace period defines how long Traefik should continue accepting requests prior to shutting down all connection listeners. It preceeds the existing (HTTP shutdown) grace period and allows for a full-circle graceful termination by expecting downstream load-balancers to take Traefik out of rotation in time, without the need for rather complex and external request draining procedures.

Immediate beneficiaries are providers such as Kubernetes and Marathon which expect in-cluster applications to exhibit the described termination behavior.

This PR addresses one of the two necessary TODOs for #1942. However, it also stands useful on its own.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -502,6 +502,6 @@ type ForwardingTimeouts struct {
 // LifeCycle contains configurations relevant to the lifecycle (such as the
 // shutdown phase) of Traefik.
 type LifeCycle struct {
-	ReqAcceptGraceTimeOut flaeg.Duration `description:"Duration to keep accepting requests before Traefik initiates the graceful shutdown procedure"`
-	GraceTimeOut          flaeg.Duration `description:"Duration to give active requests a chance to finish before Traefik stops"`
+	RequestAcceptGraceTimeout flaeg.Duration `description:"Duration to keep accepting requests before Traefik initiates the graceful shutdown procedure"`
+	GraceTimeOut              flaeg.Duration `description:"Duration to give active requests a chance to finish before Traefik stops"`
 }
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -42,6 +42,7 @@ const (
 // GlobalConfiguration holds global configuration (with providers, etc.).
 // It's populated from the traefik configuration file passed as an argument to the binary.
 type GlobalConfiguration struct {
+	ReqAcceptGraceTimeOut     flaeg.Duration          `description:"Duration to keep accepting requests before Traefik initiates the graceful shutdown procedure"`
 	GraceTimeOut              flaeg.Duration          `short:"g" description:"Duration to give active requests a chance to finish before Traefik stops"`
 	Debug                     bool                    `short:"d" description:"Enable debug mode"`
 	CheckNewVersion           bool                    `description:"Periodically check if a new version has been released"`
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/containous/flaeg"
 	"github.com/containous/traefik/acme"
+	"github.com/containous/traefik/log"
 	"github.com/containous/traefik/provider/boltdb"
 	"github.com/containous/traefik/provider/consul"
 	"github.com/containous/traefik/provider/docker"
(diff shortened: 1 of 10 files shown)


timoreimann:
@emilevauge I remember you weren't completely convinced of this feature, so I went ahead and implemented it to help giving the full picture.

See the linked issue for some extra rationale.

timoreimann:
@emilevauge 

>Like what we did with `RespondingTimeouts`& `ForwardingTimeouts`, it may be interesting to group both `reqAcceptGraceTimeOut` & `GraceTimeOut` into a struct (something like `LifeCyleTimeouts`).

Sounds very reasonable to me. 👍  I'll work on making the necessary adjustments.

timoreimann:
Still WIP, things aren't done yet.

timoreimann:
Holding back the bot as we only want to merge this after #2114 landed in master, after we which we need to update the integration test this PR adds with the new generic show-logs-on-integration-test-failure approach.

configuration/configuration.go
```diff
@@ -37,12 +38,17 @@ const (
...
+	// DefaultGraceTimeout controls how long Traefik serves pending requests
+	// prior to shutting down.
+	DefaultGraceTimeout = 10 * time.Second
 )
 
 // GlobalConfiguration holds global configuration (with providers, etc.).
 // It's populated from the traefik configuration file passed as an argument to the binary.
 type GlobalConfiguration struct {
-	GraceTimeOut              flaeg.Duration          `short:"g" description:"Duration to give active requests a chance to finish before Traefik stops"`
+	LifeCycle                 *LifeCycle              `description:"Timeouts influencing the server life cycle"`
```
marco-jantke: Do you expect other settings than timeouts inside of the `LifeCycle` struct? If not do we want to rename it to `LifeCycleTimeouts` for consistency reasons?

timoreimann: We might, I don't know.

The likely more important point for me is that the other approach stutters: `LifeCycleTimeouts.GraceTimeOut` has one timeout too many, so it should either be removed from the struct name or the parameters IMHO. I decided for the former as that allows us to extend the struct in the future without the need for another name update.

marco-jantke: I see your point. SGTM.

docs/configuration/commons.md
```diff
@@ -261,6 +264,37 @@ The configured status code ranges are inclusive; that is, in the above example,
...
+## Life Cycle
+
+Controls the behavior of Traefik during the shutdown phase.
+
+```toml
+[lifeCycle]
+
+# Duration to keep accepting requests prior to initiating the graceful
+# termination period (as defined by the `graceTimeOut` option). This
+# option is meant to give downstream load-balancer sufficient time to
```
marco-jantke: It should be either: "a downstream load-balancer" or "downstream load-balancers".

timoreimann: Pluralized for the better.

timoreimann:
This one is ready for review again.

integration/basic_test.go
```diff
@@ -101,3 +103,67 @@ func (s *SimpleSuite) TestPrintHelp(c *check.C) {
...
+	s.createComposeProject(c, "reqacceptgrace")
+	s.composeProject.Start(c)
+
+	whoami := "http://" + s.composeProject.Container(c, "whoami").NetworkSettings.IPAddress + ":80"
+
+	file := s.adaptFile(c, "fixtures/reqacceptgrace.toml", struct {
+		Server string
+	}{whoami})
+	defer os.Remove(file)
+	cmd, output := s.cmdTraefik(withConfigFile(file))
```
Juliens: Can you use the new version of traefikCmd with defer display (https://github.com/containous/traefik/pull/2114)

configuration/configuration.go
```diff
@@ -438,3 +490,10 @@ type ForwardingTimeouts struct {
 	DialTimeout           flaeg.Duration `description:"The amount of time to wait until a connection to a backend server can be established. Defaults to 30 seconds. If zero, no timeout exists"`
 	ResponseHeaderTimeout flaeg.Duration `description:"The amount of time to wait for a server's response headers after fully writing the request (including its body, if any). If zero, no timeout exists"`
 }
+
+// LifeCycle contains configurations relevant to the lifecycle (such as the
+// shutdown phase) of Traefik.
+type LifeCycle struct {
+	ReqAcceptGraceTimeOut flaeg.Duration `description:"Duration to keep accepting requests before Traefik initiates the graceful shutdown procedure"`
```
Juliens: Why not RequestAcceptGraceTimeout ?

integration/basic_test.go
```diff
@@ -101,3 +103,67 @@ func (s *SimpleSuite) TestPrintHelp(c *check.C) {
...
+	// period has elapsed.
+	waitErr := make(chan error)
+	go func() {
+		waitErr <- cmd.Wait()
+	}()
+
+	select {
+	case err := <-waitErr:
+		c.Assert(err, checker.IsNil)
+	case <-time.After(10 * time.Second):
```
Juliens: 5s (sleep) + 10s (after) = 15s 
reqAcceptGraceTimeOut (10s) + defaultGraceTimeout(10s) = "20s"

integration/basic_test.go
timoreimann: Done.

integration/basic_test.go
timoreimann: The grace period shouldn't be used: We send the last request to the server at t =~ 5 when we are halfway into the request accept grace timeout. It should finish at t = 5 + delta, so at t = 10 no request should be left that needs to get served.

I left an extra to explain the chosen timeout.

configuration/configuration.go
timoreimann: Sure, why not -- done.

timoreimann:
@Juliens et. al: This is ready for review again. I have also adjusted for #2114.

marco-jantke:
My LGTM stays 👍

PR #1030, opened 2017-01-10
Add a forward authentication option v2

vladshub:
It seems that there is no progress in #764 so I have re-based the work that was done there and added something that I have been waiting for #764 to be merged.
Please take a look 
Now it is possible to delegate the authentication to a third party agent, like a OAuth endpoint. This authentication option extracts parameters from the original request, forward then to a authentication endpoint, if it succeed part of the response body can be added as json serialized headers.

PR #1972, opened 2017-08-19
Add forward authentication option

drampelt:
### Description
This implements forward authentication, much like in `ngx_http_auth_request_module`.

Based on the discussion in #1030 and [Proposal Doc](https://docs.google.com/document/d/1-DUcCt1k_sgyRurOxcSd4T0bgQvFRvIm4rBdDf1b5bc/edit), it's clear we needed to start with something simple and then build on it later. This implementation is a simplified version of @vladshub's implementation in #1030, removing the request and response body header/query parameter manipulation.

When forward auth is enabled, the request is first forwarded to the specified auth server. If the auth server responds with a 2XX status, the original request is performed. Otherwise, the response from the auth server is returned.

### Next steps
If everyone agrees this is a good starting point, I think the next step would be either copying or renaming response headers from the auth server as request headers sent to the backend. 

From what I can tell that seems to be the main use-case, the auth server provides authenticated user information in response headers and then the backend application uses that information to automatically authenticate the user. See [oauth2_proxy with nginx](https://github.com/bitly/oauth2_proxy#configuring-for-use-with-the-nginx-auth_request-directive) as an example of this. I believe this would bring it to feature parity with `ngx_http_auth_request_module`.

See [this commit](https://github.com/drampelt/traefik/commit/b10cc88548d699f45460da77677db7a43f0c11e3) for a potential implementation of header mapping.

Of course I'm open to discussion about how to proceed and interested in hearing other ideas.

### DIFF
--- a/auth/forward.go
+++ b/auth/forward.go
@@ -1,0 +1,47 @@
+package auth
+
+import (
+	"io/ioutil"
+	"net/http"
+
+	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/types"
+)
+
+// Forward the authentication to a external server
+func Forward(forward *types.Forward, w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
+	client := http.Client{}
+
+	forwardReq, err := http.NewRequest(http.MethodGet, forward.Address, nil)
+	if err != nil {
+		log.Debugf("Error calling %s. Cause %s", forward.Address, err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	forwardReq.Header = r.Header
+
+	forwardResponse, forwardErr := client.Do(forwardReq)
+	if forwardErr != nil {
+		log.Debugf("Error calling %s. Cause: %s", forward.Address, forwardErr)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	body, readError := ioutil.ReadAll(forwardResponse.Body)
+	if readError != nil {
+		log.Debugf("Error reading body %s. Cause: %s", forward.Address, readError)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	defer forwardResponse.Body.Close()
+
+	if forwardResponse.StatusCode < http.StatusOK || forwardResponse.StatusCode >= http.StatusMultipleChoices {
+		log.Debugf("Remote error %s. StatusCode: %d", forward.Address, forwardResponse.StatusCode)
+		w.WriteHeader(forwardResponse.StatusCode)
+		w.Write(body)
+		return
+	}
+
+	r.RequestURI = r.URL.RequestURI()
+	next(w, r)
+}
--- a/auth/forward.go
+++ b/auth/forward.go
@@ -10,7 +10,20 @@ import (
 
 // Forward the authentication to a external server
 func Forward(forward *types.Forward, w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
-	client := http.Client{}
+	httpClient := http.Client{}
+
+	if forward.TLS != nil {
+		tlsConfig, err := forward.TLS.CreateTLSConfig()
+		if err != nil {
+			log.Debugf("Impossible to configure TLS to call %s. Cause %s", forward.Address, err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		httpClient.Transport = &http.Transport{
+			TLSClientConfig: tlsConfig,
+		}
+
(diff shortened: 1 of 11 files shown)


emilevauge:
Wow, great job @drampelt !
Design LGTM :)

auth/forward.go
```diff
@@ -0,0 +1,47 @@
...
+	forwardReq.Header = r.Header
+
+	forwardResponse, forwardErr := client.Do(forwardReq)
+	if forwardErr != nil {
+		log.Debugf("Error calling %s. Cause: %s", forward.Address, forwardErr)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	defer forwardResponse.Body.Close()
```
emilevauge: I would move this line after the block:
```
if readError != nil {
...
}
defer forwardResponse.Body.Close()
```
Indeed, if `readError` is not nil, maybe the body will be empty, then the defered `Close()` would lead to a panic.

types/types.go
```diff
@@ -314,6 +315,11 @@ type Digest struct {
 	UsersFile string
 }
 
+// Forward authentication
+type Forward struct {
+	Address string
```
emilevauge: I think it would be great to add the possibility to connect in TLS (with an self-signed cert).

nmengin: I'm agree with @emilevauge for the TLS connection.
Indeed, I tried to test the PR by connecting Traefik to a [docker_auth](https://github.com/cesanta/docker_auth) server but this server allows only HTTPS connections...

emilevauge:
@drampelt The code freeze for 1.4 will happen in 2 days (August 25th). Do you think you will be able to fix your PR in the meantime ?

drampelt:
Thanks for the reviews, I have a fairly busy week but I'll see if I can get it fixed by up tomorrow.

auth/forward.go
```diff
@@ -0,0 +1,47 @@
...
+	body, readError := ioutil.ReadAll(forwardResponse.Body)
+	if readError != nil {
+		log.Debugf("Error reading body %s. Cause: %s", forward.Address, readError)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	if forwardResponse.StatusCode < 200 || forwardResponse.StatusCode >= 300 {
```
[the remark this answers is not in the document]
drampelt: I chose 300 to allow the auth server to redirect to a login page if necessary. For example if doing Google sign in, the auth server would need to redirect the user to Google by returning a 302.

types/types.go
drampelt: Yes I agree that would be useful. Unfortunately I'm not too familiar with Go yet and not really sure how to go about doing that - would anyone else be interested in implementing that?

middlewares/authenticator.go
```diff
@@ -6,7 +6,9 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/abbot/go-http-auth"
+	goauth "github.com/abbot/go-http-auth"
+
+	"github.com/containous/traefik/auth"
```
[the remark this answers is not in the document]
drampelt: Sorry, could you clarify what order they should be in? I'm not too familiar with Go yet and gofmt seems to think that is fine.

drampelt:
Fixed most of the issues I knew how to.

auth/forward.go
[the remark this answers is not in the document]
ldez: 👍

types/types.go
nmengin: No problem @drampelt.
We can do the modifications for your :wink:

RC1140:
Hi 
Any idea how to use this, I grabbed the latest RC and made the config changes as per the doc update but I never get redirected.

I used github as an example to test

defaultEntryPoints = ["http"]

[entryPoints]
    [entryPoints.http]
    address = ":80"
      [entryPoints.http.auth.forward]
      address = "http://github.com/login/oauth/authorize"

[web]
address = ":8081"

[file]

[backends]
  [backends.backend1]
       url = "http://localhost:8000"

[frontends]
  [frontends.frontend1]
      backend = "backend1"
      passHostHeader = true
      [frontends.frontend1.routes.example]
          rule = "Host:localhost"

If I try to hit localhost i simply  hit my local server on port 8000 with no auth request.

ldez:
@RC1140 Thanks for your interest in Traefik 😃 

Please discuss this in :
- [Stack Overflow](https://stackoverflow.com/questions/tagged/traefik) using the "traefik" tag
- the [Traefik community Slack channel](https://traefik.herokuapp.com)

Issue #1498, opened 2017-04-26
[Rancher] No IP in backend in host networking mode


tboyce021:
### Do you want to request a *feature* or report a *bug*?
Bug. Appears to be the same as #487 except for Rancher (I even ran into it setting up Home Assistant as well).

### What did you do?
Set up Traefik with Rancher backend for a container with net=host.

### What did you expect to see?
IP address in backend

### What did you see instead?
IP address is missing

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
1.2.3
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
# (paste your configuration here)
```

### If applicable, please paste the log output in debug mode (`--debug` switch)

```
(paste your output here)
```

ldez:
Could you provide more information? (step to reproduce, logs, configuration, ...)

dniel:
I have seen the same issue when debugging #1491, no ip registered when setting the service with [network mode to HOST](http://docs.rancher.com/rancher/v1.5/en/rancher-services/networking/#host), only port number registered.

tboyce021:
It's the exact same issue as #487 except for the Rancher backend instead of Docker.

To reproduce: Just set up Traefik with the Rancher backend and set one of the exposed containers/services to `Host` network mode. The Traefik backend will show the URL as `http://<port>`.

This is because Rancher provides a null IP address for containers in host mode. You can see this by just viewing the container in the Rancher API and you'll see `primaryIpAddress": null`. This is then used here: https://github.com/containous/traefik/blob/master/provider/rancher/rancher.go#L399.

The Docker workaround in #494 will work if the container is running on the same machine as Traefik, but otherwise it won't (same as with Docker swarm mode). The better solution would be to use the IP address of the container's host.

timoreimann:
This sounds somewhat similar to the Marathon-specific problem #1243, which got solved by auto-detecting the right mode (#1345).

dniel:
Yes this sounds very similar to #1243 .
I'm trying to find a workaround in my project, but still from what I figured out, the rancher provider does not support traefik as an external load balancer for services deployed by Rancher because of the ip of the services is always reported either as the internal rancher network address or missing.

Right now I'm trying to figure out how to customize the template for rancher provider for a custom workaround..

matq007:
Any update on how to solve this issue? I would like to use Traefik as a reverse-proxy on rancher :)

traefiker:
Closed by #2101.

PR #1977, opened 2017-08-20
Use Host IP address if PrimaryIP is empty

matq007:
### Description

When host network is selected on rancher container, traefik returns as primary IP address "". This results in proxying to http://:PORT. With this PR if the IP address is empty, the returned IP address is host's address where the container is running.

Follow up description: https://github.com/containous/traefik/issues/1498

This is a quick-fix and should be made differently in future. Probably it should start by detecting what network was chosen on a container and based on that return the proper IP address.

### DIFF
--- a/provider/rancher/rancher.go
+++ b/provider/rancher/rancher.go
@@ -231,6 +231,14 @@ func getenv(key, fallback string) string {
 	return value
 }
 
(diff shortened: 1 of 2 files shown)


matq007:
Closing because of failed build and wrong branch

PR #1978, opened 2017-08-20
Use Host IP address if PrimaryIP is empty

matq007:
### Description

Description

When host network is selected on rancher container, traefik returns as primary IP address "". This results in proxying to http://:PORT. With this PR if the IP address is empty, the returned IP address is host's address where the container is running.

Follow up description: #1498

This is a quick-fix and should be made differently in future. Probably it should start by detecting what network was chosen on a container and based on that return the proper IP address.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## [v1.3.5](https://github.com/containous/traefik/tree/v1.3.5) (2017-08-01)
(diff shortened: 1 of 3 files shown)


ldez:
Thanks for your interest in Traefik ! 👍 

You cannot manually edit files in the `vendor`.

> Rule #1 of open-source: no is temporary, yes is forever.

— Solomon Hykes https://twitter.com/solomonstre/status/715277134978113536

We don't want quick-fix and in respect of this rule I must close your PR.

We will be very happy if you open a "non quick-fix" PR on master.

Issue #1975, opened 2017-08-20
Switch to the new libkv


0xbaadf00d:
Alexandre Beslic (@abronan) one maintainer of libkv just set the project alive by merging lot of PR (etcd fixes, added Redis, ...) and forking the project to continue to upgrade this project.

It would be interresting for Traefic to switch to this new version.

URL : https://github.com/abronan/libkv

ldez:
https://github.com/containous/traefik/issues/926#issuecomment-323348372

dtomcej:
wow, there has been a LOT of activity in that repo in the last 2 weeks.

I look forward to awesome etcd3 support!

trusch:
etcd v3 is ready in this repo. bump :)

abronan:
@trusch afaik they are working on it but trying to fix a few remaining issues (as I haven't had the time to test etcd v3 support with traefik yet). This could unblock soon.

zyclonite:
looks like alexandre is in the process of (or even already done) renaming the libkv to valkeyrie (https://github.com/abronan/valkeyrie)

... and continuing to maintain only this repository

zicklag:
So does this mean that Redis is a supported KV backend now? It isn't in the [documentation](https://docs.traefik.io/user-guide/kv-config/).

Issue #926, opened 2016-12-02
letsEncrypt not working with etcd


vincentvdk:
When I try to configure Traefik in combination with 'etcd' as a backend store, i can't get LetsEncrypt working.

traefik.toml (imported with storeconfig)
```toml
[web]
address = ":8080"
defaultEntryPoints = ["https"]
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    entryPoint = "https"
  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]
[acme]
email = "me@domain.tld"
storage= "/traefik/acme/account"
entryPoint = "https"
onDemand = false
OnHostRule = true
caServer = "https://acme-staging.api.letsencrypt.org/directory"
[[acme.domains]]
   main = "domain.tld"
   sans = ["www.domain.tld", "ww2.domain.tld"]
[etcd]
endpoint = "127.0.0.1:4001"
watch = true
prefix = "/traefik"
```

Output in the logs

```shell
Dec 02 09:23:18 core-01 rkt[21448]: [131075.886503] traefik[5]: time="2016-12-02T10:23:18+01:00" level=info msg="Skipping same configuration for provider etcd"
Dec 02 09:23:22 core-01 rkt[21448]: [131079.650257] traefik[5]: time="2016-12-02T10:23:22+01:00" level=debug msg="Challenge GetCertificate www.mydomain.tld"
Dec 02 09:23:22 core-01 rkt[21448]: [131079.650881] traefik[5]: time="2016-12-02T10:23:22+01:00" level=debug msg="Loading ACME certificates [www.mydomain.tld]..."
Dec 02 09:23:22 core-01 rkt[21448]: [131079.661423] traefik[5]: time="2016-12-02T10:23:22+01:00" level=debug msg="Challenge GetCertificate www.mydomain.tld"
Dec 02 09:23:22 core-01 rkt[21448]: [131079.662657] traefik[5]: time="2016-12-02T10:23:22+01:00" level=debug msg="Loading ACME certificates [www.mydomain.tld]..."
Dec 02 09:23:22 core-01 rkt[21448]: [131080.212813] traefik[5]: time="2016-12-02T10:23:22+01:00" level=debug msg="Challenge Present www.mydomain.tld"
[...]

PR #1981, opened 2017-08-21
[vendor] use abronan/libkv instead of docker/libkv;

trusch:
docker/libkv is dead and abronan/libkv is the repository where development is happening.
this solves #1975 and therefore hopefully #926 once the etcdv2 fix is merged.
this also adds support for redis and etcdv3 storage backends.


emilevauge:
@trusch Thanks for your contribution!
Your PR will not work as we are using vendoring (and we should use aliasing in glide.yaml).
If you don't mind, I will close this one and we will do it ourselves after testing the etcdv2 fiw with Traefik.

PR #1993, opened 2017-08-22
Update doc page

jmaitrehenry:
### Description

Update doc style, add space between line, add space inside inline code block, add space between li element.

Update toml documentation page for replacing the docker labels list by a tables.

### DIFF
--- a/docs/css/traefik.css
+++ b/docs/css/traefik.css
@@ -1,8 +1,3 @@
-body {
-    font-size: 15px;
-    line-height: 24px
-}
-
 code {
     font-size: 90%;
     padding: 3px 7px;
@@ -10,11 +5,6 @@ code {
     border: solid 1px #ccc;
     color: #333;
     border-radius: 2px;
-}
-
-
-li {
-    padding-bottom: 0.7em;
 }
 
 a {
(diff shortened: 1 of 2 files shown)


tboerger:
Some Screenshot to see the difference should be cool :)

docs/css/traefik.css
```diff
@@ -1,3 +1,22 @@
+body {
+    font-size: 15px;
+    line-height: 24px
```
ldez: I don't really like size with `px`, I prefer `em` or `rem`

jmaitrehenry:
Before / After space
screen shot 2017-08-22 at 11 17 46 am
screen shot 2017-08-22 at 11 17 45 am

Before / After table
screen shot 2017-08-22 at 11 17 26 am
screen shot 2017-08-22 at 11 17 20 am

I will make the change for using em/rem and not pixel.
And I see I miss a save before commit and the space was on body and not on .main... I will include the fix with the rem/em change.

What do you think for removing the fixed width? We could have more space for long line if we use all the width available.

jmaitrehenry:
@ldez No problem, I made only one PR because it's all related to readability.

jmaitrehenry:
Done!

PR #1994, opened 2017-08-22
Update doc style

jmaitrehenry:
Increase font size, line height, and inline-code style.
Related to #1993


ldez:
Closed due to #2012

PR #2002, opened 2017-08-23
Fix command bug content.

ldez:
### Description

- Replace `toml` by `json`
- fix anonymize.


emilevauge:
@SantoDE sorry to ask but we need an `approve` and not a `comment` here, otherwise, our friend @traefiker won't accept to merge this PR :(

PR #2003, opened 2017-08-24
Add readiness endpoint failing on shutdown.

timoreimann:
This change adds a new endpoint `/ready` if a corresponding configuration option is set. It is meant to be served to downstream load-balancers as it will automatically stop serving 200 OK responses as soon as Traefik receives a termination signal.

Note that the PR contains commits from #1971. Please only review commit 2d115c2 (_Add readiness endpoint failing on shutdown_) and those following afterwards.

A design question we should answer: Currently, I made the option as simple as possible (just a boolean) and hard-coded the successful HTTP code to 200 and the failing one to 503. If we think that making this configurable in the future is realistic, we should probably go straight with a sub-struct inside the web provider configuration to leave room for further options (and not having to deprecate this very simple approach we have right now).
Another feature request often cited is draining, which we could implement with this new entry point by allowing a POST request to toggle its state (instead of a just a SIGINT/SIGTERM as of now). Allowing this option in the future would be another reason to add a generic struct.

Complements #1942.


Juliens:
Why not just make `/ping` return 503 during graceful time instead of creating a new endpoint ?
Moreover do you really think it is necessary to be able to disable it ?

timoreimann:
@Juliens:

>Why not just make /ping return 503 during graceful time instead of creating a new endpoint ?

Yes, we could extend `/ping`'s purpose as you suggested. I thought about that as well but took the conservative route. My preference is to keep ping as simple as possible and use a dedicated endpoint for all things related to readiness, just to separate concerns more cleanly.

That said, If we think that it'd be better to fold the readiness feature into the ping endpoint (and be aware of the implications it brings for future features in this regard), I can do that as well.

>Moreover do you really think it is necessary to be able to disable it ?

Not 100% sure if you're questioning the feature itself or the potential future feature to toggle the readiness endpoint off on demand.

In case of the former: I hope my rationale in the referenced issue describes why I think it's necessary.

In the case of the latter: To quote myself from the PR description:

>Another feature request often cited is draining, which we could implement with this new entry point by allowing a POST request to toggle its state (instead of a just a SIGINT/SIGTERM as of now). Allowing this option in the future would be another reason to add a generic struct.

So draining would be one desirable candidate for making things toggleable.

Juliens:
> Not 100% sure if you're questioning the feature itself or the potential future feature to toggle the readiness endpoint off on demand.

I think the `enableReadiness` option may not be necessary, we can alway enable it ?

timoreimann:
@Juliens now I got you. I only added that option so that people might be less surprised if a new endpoint suddenly pops up. But that might not be that important, so I'm okay with enabling it unconditionally.

emilevauge:
@timoreimann I tend to agree with @Juliens on this. IMHO, we already have too much endpoints:
- `/health` which should only give health info, but here returns stats
- `/metrics` which give some stats
- `/ping` used to know if traefik us up
- `/ready` would be used to know if traefik is ready

I really think `/health`, `/ping` & `/ready` are close and may be merged. **But**! We cannot change `/health` to stay backward compatible right now, and we would need to change webui. I suggest to reuse `/ping` as @Juliens said earlier :)

timoreimann:
Alright, I'll update the PR to reuse `/ping` for the purpose at hand.

@emilevauge assuming your comment approves the design, I'll move foward with the state labels (but also add a WIP one until I've finished my updates).

timoreimann:
We have recently changed our internal Traefik setup so that we are not relying on the feature this PR offers anymore.

If there's still some interest, I'd be happy to move the PR to completion. Otherwise, I'd just close it.

timoreimann:
Closing it for now. If interest resurfaces, please let me know.

timoreimann:
@ravilr we are running on GKE/GCE and switched to Google's internal Service-based LoadBalancers which natively understand pod readiness states. Thus, when the kubelet decides to SIGTERM a pod, it implicitly makes sure that the readiness state transitions to false, which in turn causes the LB to remove the pod from rotation.

If that's not an option for you, I suppose it's okay to submit another PR on your own, taking into account what has been discussed along this PR.

PR #1145, opened 2017-02-09
Adding support for the ProxyProtocol

jrb:
Tests and docs are still in the works, but I wanted to go ahead and open a PR in case I've gone in the wrong direction with this.

Adds a new config item "ProxyProtocol" to endPoints. When enabled, this causes the endPoint's listener to be wrapped with https://github.com/armon/go-proxyproto

PR #2004, opened 2017-08-24
Add proxy protocol

### LINKED ISSUE #384: Proxy Protocol Support
Is there any plans for support for [PROXY protocol](http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt)?

emilevauge:
### Description

This PR supersedes #1145 to add [Proxy Protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) support. It is based on the initial work made by [jrb](https://github.com/jrb).
it also bumps Golang to 1.9 (needed to add Proxy Protocol).

Fixes #384

### DIFF
--- a/build.Dockerfile
+++ b/build.Dockerfile
@@ -1,11 +1,8 @@
-FROM golang:1.8
+FROM golang:1.9-alpine
 
-# Install a more recent version of mercurial to avoid mismatching results
-# between glide run on a decently updated host system and the build container.
-RUN awk '$1 ~ "^deb" { $3 = $3 "-backports"; print; exit }' /etc/apt/sources.list > /etc/apt/sources.list.d/backports.list && \
-  DEBIAN_FRONTEND=noninteractive apt-get update && \
-  DEBIAN_FRONTEND=noninteractive apt-get install -t jessie-backports --yes --no-install-recommends mercurial=3.9.1-1~bpo8+1 && \
-  rm -fr /var/lib/apt/lists/
+RUN apk --update upgrade \
+&& apk --no-cache --no-progress add git mercurial bash gcc musl-dev curl tar \
+&& rm -rf /var/cache/apk/*
 
 RUN go get github.com/jteeuwen/go-bindata/... \
 && go get github.com/golang/lint/golint \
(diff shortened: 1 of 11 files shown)


build.Dockerfile
```diff
@@ -1,11 +1,8 @@
-FROM golang:1.8
+FROM golang:1.9-alpine
```
timoreimann: Do we plan to merge this even when Go 1.9 is still beta? I'd be concerned about that.

emilevauge: We think the go 1.9 final will be out way before the trafik 1.4 final is ready, and this feature is the most wanted feature ;) So all in all, we think it's a reasonable risk.

timoreimann: What gives you confidence the next Go release will be out way before Traefik 1.4? Go 1.9 is only at the first release candidate, and Traefik is going to ship its first within the next few days presumably.

emilevauge: @timoreimann the rc2 was released weeks ago, and the final release date was theoretically fixed weeks ago, and also https://github.com/golang/go/milestone/49 ;)

timoreimann: OK, missed that it's actually rc2 already.

I'm still not sure it's a good idea to merge this before Go 1.9 final is actually released. I have seen things break in subtle ways over at least the last two Go releases. I'd prefer to take the time after the code freeze to iron out potential problems.

server/configuration.go
```diff
@@ -247,12 +247,18 @@ func (ep *EntryPoints) Set(value string) error {
 		whiteListSourceRange = strings.Split(result["WhiteListSourceRange"], ",")
 	}
 
+	proxyprotocol := false
+	if len(result["ProxyProtocol"]) > 0 {
+		proxyprotocol = strings.EqualFold(result["ProxyProtocol"], "enable") || strings.EqualFold(result["ProxyProtocol"], "on")
```
dtomcej: You check to see if `proxyprotocol` is `enable` or `on`, but not `true` based on your example in the toml?

emilevauge: Yeah, I just copied what was already done for `Compress` flag. Added `true for both :)

build.Dockerfile
racastellanosm: I think a lot of people need the Proxy Protocol Support (included myself)! Any release date in the scope?

Issue #1962, opened 2017-08-16
Route sorting is not working as expected


Vad1mo:
According to the docs routes are sorted using routes length.

> By default, routes will be sorted (in descending order) using rules length (to avoid path overlap): PathPrefix:/12345 will be matched before PathPrefix:/1234 that will be matched before PathPrefix:/1.

When I run the below  compose example and curl the url ` curl http://sample.127.0.0.1.xip.io/v2/asdf`
then it routes to the  backend **web** however I was expecting to get a route to **api** because of the v2 prefix that is longer in the api then in web.

```
  version: '2'
  services:
#---#
    proxy:
      image: traefik:1.3-alpine
      command: --web --docker --docker.domain=docker.localhost --logLevel=DEBUG --docker.endpoint=unix:///var/run/docker.sock
      ports:
        - 80:80
        - 443:443
      labels:
        - traefik.enable=false
      volumes:
        - /var/run/docker.sock:/var/run/docker.sock
        - /dev/null:/traefik.toml
#---#
    service_web:
      image: vad1mo/hello-world-rest
      hostname: web
      labels:
        - traefik.backend=web
        - traefik.port=5050
        - traefik.frontend.rule=Host:sample.127.0.0.1.xip.io;PathPrefix:/,/v2/webhook,/v2/token
        - traefik.frontend.entryPoints=http
#        - traefik.priority = 5
#---#
    service_api:
      image: vad1mo/hello-world-rest
      hostname: api
      labels:
        - traefik.backend=api
        - traefik.port=5050
        - traefik.frontend.rule=Host:sample.127.0.0.1.xip.io;PathPrefix:/v2
        - traefik.frontend.entryPoints=http
       # - traefik.priority = 10
```

sirlatrom:
I'd like to reopen this issue as the title is still appropriate. The docs suggest that the priority number, if specified, takes precedence over the rule length when comparing two rules. However, the priority field rather works as a relative offset, given how it's added to the rule length in the relevant code.

Should the docs be updated (and perhaps a more suitable name for the setting, e.g. "offset"), or should the behaviour be updated to comply with the docs?

ldez:
@sirlatrom The behavior will not change because this could be breaking, but documentation update seems needed.

We will very happy if you open a PR to change the documentation.

FYI:
https://github.com/containous/traefik/blob/c0563f1a395d53edbbf1030b2387435077eeb871/server/server.go#L1338-L1347

sirlatrom:
[...]

Issue #1876, opened 2017-07-18
Frontend rules are not updated based on priorities for identical rules.


ewah:
### Do you want to request a *feature* or report a *bug*?

*bug*

### What did you do?

I am creating a series of containers that have the same frontend.rule and would like the highest frontend.priority to dictate the route.

```
sudo docker service create \
  --name hello-world-A \
  --network mynetwork \
  --publish :80 \
  --label traefik.protocol=http \
  --label traefik.port=80 \
  --label traefik.frontend.rule="Host:helloworld.com" \
  --label traefik.frontend.priority=1707180001 \
  --label traefik.docker.network=mynetwork \
  localhost:5000/hello-world-nginx:170718a

sudo docker service create \
  --name hello-world-B \
  --network mynetwork \
  --publish :80 \
  --label traefik.protocol=http \
  --label traefik.port=80 \
  --label traefik.frontend.rule="Host:helloworld.com" \
  --label traefik.frontend.priority=1707180002 \
  --label traefik.docker.network=mynetwork \
  localhost:5000/hello-world-nginx:170718b
```

When created simultaneously, Traefik routes to `hello-world-B` properly.

```
sudo docker service create \
  --name hello-world-C \
  --network mynetwork \
  --publish :80 \
  --label traefik.protocol=http \
  --label traefik.port=80 \
  --label traefik.frontend.rule="Host:helloworld.com" \
  --label traefik.frontend.priority=1707180003 \
  --label traefik.docker.network=mynetwork \
  localhost:5000/hello-world-nginx:170718c
```

Later, adding a service with the same `frontend.rule` and incremented `frontend.priority` it does not register (i.e. hello-world-B stays live).

`hello-world-C` will appear if `hello-world-A` and `hello-world-B` are removed.

(Also, a nice to have would be is to display all identical rules that are overridden via the UI/API. )

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.3.1
Codename:     raclette
Go version:   go1.8.3
Built:        2017-06-16_11:21:48AM
OS/Arch:      linux/amd64
```

dtomcej:
Discussed with @ewah on slack, and it appears that only one frontend gets created:

```
[...]

PR #2008, opened 2017-08-25
Split a joint-rule into individual rules.

### LINKED ISSUE #1663: Rule priorities are based on total rule length, not individual path prefix
### Do you want to request a feature or report a bug?

A bug. I believe Traefik uses incorrect logic to decide on routing rules priority.

### What did you do?

Consider Traefik is configured to watch for Docker containers (a very basic config that essentially boils down to `[docker] watch = true`), and observes the following configuration in their labels (as a snippet from Docker Compose file):

    web:
        ...
        labels:
            traefik.enable: "true"
            traefik.backend: "web"
            traefik.frontend.rule: "Host:one.example.org,one.example.net,old.one.example.net,old.one.example.org"

    web-ng:
        ...
        labels:
            traefik.enable: "true"
            traefik.backend: "web-ng"
            traefik.frontend.rule: "Host:one.example.org,one.example.net,PathPrefix:/api/v2/,/o/"

### What did you expect to see? What did you see instead?

I've expected a request to `one.example.org/api/v2/` to be routed to the `web-ng` container, as it has more specific `PathPrefix`.

However, requests were sent to `web` instead. This was the case until I've made the `web-ng`'s rule look like `"Host:one.example.org,one.example.net,PathPrefix:/api/,/o/,/a-long-string-to-make-this-rule-more-specific/"`.

I've expected Traefik to compare path prefixes of all matching rules rather than rules themselves. I.e. compare "" (longest PathPrefix that matches from the `web`'s rules - empty in this case) vs "/api/" (longest PathPrefix that matches from the `web-ng`'s rules). I haven't looked at the code but it seems that this isn't the case. I've experimented a little and can tell that traffic is routed to `web-ng` only when the whole rule's length is greater *or equal* to the `web`'s frontend rule's length.

    Version:      v1.3.0-rc3
    Codename:     raclette
    Go version:   go1.8.2
    Built:        2017-05-24_01:39:53PM
    OS/Arch:      linux/amd64

yue9944882:
1. Split joint rules such as `PathPrefix:/v1,/v2` into splited individual rules `PathPrefix:/v1` and `PathPrefix:/v2`)
2. Every individual rule's priority is measured by the length of its own substring
3. Take all the splited individual rules as the subrouter of `serverRoute.route`. And the priority of this parent route (`serverRoute.route`) is only assigned by frontend.Rule

### Description

It works fine and solved #1663 and its related problems. Corresponding unit test is also added.

As is stated in the commit message, it works by giving every sub-route a individual priority which is calculated by the length of args. E.g. in the expression `PathPrefix:/v1,/v2345`, it will be split into 2 sub-route which is `PathPrefix:/v1` and `PathPrefix:/v2345`, and their priority is 3 and 6.

So in the case of `PathPrefix:/,/v2/webhook,/v2/token` and `PathPrefix:/v2`, a request with `/v2` will match `/v2` becase the priority of `/v2` is 3 and `/` is `1`,`/v2/webhook` and `/v2/token` not matched. Or rather, it solves #1962 also. 

#1876 is also related.

Fixes #1663


yue9944882:
So wierd?

CI failed on provider/file/file_test.go, but my commit have no relation to this?

Any one can provide help? Thx

aantono:
Can you add an updated documentation to reflect the change in logic?

ldez:
@aantono the PR is in Design review: no changes are required in this state.

yue9944882:
@aantono I tried to add some descriptive documentation to my commit. But truly I don't know where to start with. Because this commit simply added some repeating simple codes.

The BUG referenced in #1663 and #1962 is caused by following line of code. Assigning the priority of route with the sum of priority set by frontend and the length of "rule string" itself is neither mathematical nor reasonable. And it is certainly the cause of related BUGs.
 
```
 -	newRoute.Priority(serverRoute.route.GetPriority() + len(route.Rule))
 +	newRoute.Priority(serverRoute.route.GetPriority())
```

So my PR is to transform the route assigned priority of (serverRoute.route.GetPriority() + len(route.Rule)) into a two-layered routes. The priority of parent route is given by its pre-set value from frontend rule and the priority of child routes are given by their spilt individual rules. 

As for how the rule is split into individual child rules, please refer to my commit message and description above. Thx :-)

emilevauge:
@ldez @Juliens Any thoughts on this ?

Juliens:
Thx for your PR. 

In fact, the Priority is ignored in SubRoute.
So in your modification, all the priorities you set are ignored. The order become in "declaration order"

If you try to inverse your rules declarations in your test, you will see that it will fail.

As it means a full refactor, I propose to close this PR, you can then create another one with a new implementation. WDYT ?

yue9944882:
Thx. I'll try some another implementation. 


Close this PR.

PR #2030, opened 2017-08-31
Add Custom header parsing to Docker Provider

### LINKED ISSUE #2028: Use labels to set custom HTTP headers in Docker backend

Request a feature

### What did you do?

Add a label to the Docker container like this:

traefik.frontend.customResponseHeaders=X-Custom-Header: foo; bar

Traefik should use the Header Middleware described in #1236 to serve the application with the desired header.

### What did you see instead?

Nothing happens.

```
Version:      v1.4.0-rc1
Codename:     roquefort
Go version:   go1.9
Built:        2017-08-29_08:35:24AM
OS/Arch:      linux/amd64

```

```toml
graceTimeOut = "10s"
traefikLogsFile = "/var/log/traefik/traefik.log"
accessLogsFile = "/var/log/traefik/access.log"
logLevel = "WARN"

[acme]
email = "test@test.test"
storage = "acme.json"
entryPoint = "https"
acmeLogging = true
OnHostRule = true

[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
      entryPoint = "https"
  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]

[web]
address = ":8080"

[docker]
domain = "docker.domain"
watch = true
exposedbydefault = false

```

dtomcej:
This PR provide a basis for updating the Kubernetes provider.

Fixes #2028


dtomcej:
This PR will resolve https://github.com/containous/traefik/issues/2028 and will provide a basis for updating the kubernetes provider

dtomcej:
Still working on this.

dtomcej:
Ok. Submitting this for review.

This PR allows the custom request and response headers to be set via docker labels. The labels get parsed, and invoke the headers middleware.

dtomcej:
Ping @containous/traefik

dtomcej:
If review is passed, will squash and rebase.

provider/docker/docker.go
```diff
@@ -711,6 +713,31 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
 	return []string{}
 }
 
+func (p *Provider) getRequestHeaders(container dockerData) map[string]string {
+	h := make(map[string]string)
+	if requestHeaders, err := getLabel(container, types.LabelFrontendRequestHeader); err == nil {
+		for _, v := range strings.Split(requestHeaders, ",") {
+			k := strings.Split(v, ":")
+			h[k[0]] = k[1]
```
timoreimann: We will crash if the split in the previous line yields a single line only. Should we have some extra sanity check?

provider/docker/docker.go
```diff
@@ -711,6 +713,31 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
...
+	log.Debugf("Loaded Request Headers: %v", h)
+	return h
+}
+
+func (p *Provider) getResponseHeaders(container dockerData) map[string]string {
+	h := make(map[string]string)
+	if responseHeaders, err := getLabel(container, types.LabelFrontendResponseHeader); err == nil {
+		for _, v := range strings.Split(responseHeaders, ",") {
+			k := strings.Split(v, ":")
+			h[k[0]] = k[1]
```
timoreimann: Same here.

provider/docker/docker.go
```diff
@@ -711,6 +713,31 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
...
+		}
+	}
```
timoreimann: The parse logic for the requests and responses seems identical. Might be worth extracting into a little helper function.

templates/docker.tmpl
```diff
@@ -74,6 +74,18 @@
   basicAuth = [{{range getBasicAuth $container}}
     "{{.}}",
   {{end}}]
+  {{if getRequestHeaders $container}}
```
timoreimann: Using the parse method for checking if request headers exist means we'll emit the log outputs twice: once in the check and once in the actual extraction (line 79). This might be somewhat confusing.

It might be better do introduce a little `hasRequestHeaders` template function, similar to how we do it with other functions.

templates/docker.tmpl
```diff
@@ -74,6 +74,18 @@
...
+    [frontends."frontend-{{$frontend}}".headers.customrequestheaders]
+    {{range $k, $v := getRequestHeaders $container}}
+    {{$k}} = "{{$v}}"
+    {{end}}
+  {{end}}
+  {{if getResponseHeaders $container}}
```
timoreimann: Same here.

provider/docker/docker.go
dtomcej: Will do!

provider/docker/docker.go
dtomcej: Seems legit.

templates/docker.tmpl
dtomcej: kk.

dtomcej:
@timoreimann Squashed, Rebased, Helper-ified, and Documented :)

dtomcej:
ping @containous/traefik

provider/docker/docker.go
```diff
@@ -721,6 +725,45 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
 	return []string{}
 }
 
+func (p *Provider) hasRequestHeaders(container dockerData) bool {
+	if _, err := getLabel(container, types.LabelFrontendRequestHeader); err != nil {
+		return false
+	}
+	return true
```
timoreimann: Could also be

```go
_, err := getLabel(container, types.LabelFrontendRequestHeader)
return err == nil
```

Similar with `hasResponseHeaders`.

provider/docker/docker.go
```diff
@@ -721,6 +725,45 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
...
+				log.Warnf("Could not load header %v", k)
+				return h
+			}
+			h[k[0]] = k[1]
+		}
+	}
+
+	log.Debugf("Loaded %s: %v", containerType, h)
+	return h
+
```
ldez: could you remove this line?

provider/docker/docker.go
```diff
@@ -721,6 +725,45 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
...
+	return parseCustomHeaders(container, types.LabelFrontendRequestHeader)
+}
+
+func (p *Provider) getResponseHeaders(container dockerData) map[string]string {
+	return parseCustomHeaders(container, types.LabelFrontendResponseHeader)
+}
+
+func parseCustomHeaders(container dockerData, containerType string) map[string]string {
+	h := make(map[string]string)
+	if r, err := getLabel(container, containerType); err == nil {
```
ldez: maybe you could use explicit variable name instead of one letter.

By example:
- `h` -> `headers`
- `r` -> `label`
- `v` -> `pairs`
- `k` ->  `pair`

provider/docker/docker.go
```diff
@@ -721,6 +725,45 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
 	return []string{}
 }
 
+func (p *Provider) hasRequestHeaders(container dockerData) bool {
+	if _, err := getLabel(container, types.LabelFrontendRequestHeader); err != nil {
```
ldez: Maybe you could evaluate the len?

```go
label, err := getLabel(container, types.LabelFrontendRequestHeader)
return err == nill && len(label) > 0
```

same for `hasResponseHeaders`

provider/docker/docker.go
```diff
@@ -721,6 +725,45 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
...
+}
+
+func parseCustomHeaders(container dockerData, containerType string) map[string]string {
+	h := make(map[string]string)
+	if r, err := getLabel(container, containerType); err == nil {
+		for _, v := range strings.Split(r, ",") {
+			k := strings.Split(v, ":")
+			if len(k) != 2 {
+				log.Warnf("Could not load header %v", k)
+				return h
```
ldez: If you add a `return` you could add `log.Errorf` instead of `log.Warnf`

In the case of an invalid key/value pair, you return the current map with previous values (or an empty map if it's the first pair), this seem strange:
- either you return a empty map (-> `log.Errorf`)
- or you only skip the invalid key/value pair (-> `log.Warnf`)

provider/docker/docker.go
dtomcej: 👍

dtomcej:
@ldez corrections implemented!

provider/docker/docker.go
```diff
@@ -726,6 +730,41 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
...
+}
+
+func (p *Provider) getResponseHeaders(container dockerData) map[string]string {
+	return parseCustomHeaders(container, types.LabelFrontendResponseHeader)
+}
+
+func parseCustomHeaders(container dockerData, containerType string) map[string]string {
+	customHeaders := make(map[string]string)
+	if label, err := getLabel(container, containerType); err == nil {
+		for _, headers := range strings.Split(label, ",") {
```
vito-c: `Access-Control-Allow-Methods:POST,GET,OPTIONS` wouldn't this header cause this split to fail? maybe we should use a `;` similar to how the rules are separated?

vito-c:
```
time="2017-12-01T00:03:00Z" level=warning msg="Could not load header "traefik.frontend.headers.customrequestheaders": [POST], skipping..."
time="2017-12-01T00:03:00Z" level=warning msg="Could not load header "traefik.frontend.headers.customrequestheaders": [OPTIONS], skipping..."
time="2017-12-01T00:03:00Z" level=warning msg="Could not load header "traefik.frontend.headers.customrequestheaders": [DELETE], skipping..."
time="2017-12-01T00:03:00Z" level=warning msg="Could not load header "traefik.frontend.headers.customrequestheaders": [PUT], skipping..."
```

ldez: @vito-c please open an issue instead of comment on an already merged PR.

PR #2054, opened 2017-09-04
Send traefik logs to stdout

marco-jantke:
The default setting in `logrus` is to send the logs to stderr. The documentation states already that the output should be sent to stdout.


marco-jantke:
@ldez should we also pick this for 1.4 in case design review is successful?

ldez:
@marco-jantke yes and you can do this now 😃

marco-jantke:
Ok, done.

Issue #780, opened 2016-10-26
Add option to log in json format


errm:
It is much nicer to log access and error logs as json if you want them to be ingested into some logging system e.g. ELK, Stackdriver, Papertrail whatever. 

This allows us to add extra fields to logs without worrying that we might be breaking peoples log ingestion.

It also would be good to add an option to log access and error logs to stdout and stderr, since some systems are setup to ingest that without further config.  e.g. kubernetes

SirCAS:
I have started on a proposal for how we can implement a way to print the output from the access logs as json.  I still trying to get my head around golang and the libraries Traefik uses. I'm not sure if I'm heading in the right direction with the current implementation.

The main concerns I have about the current implementation is:
- The Traefik daemon must be restarted when the config value for LogFormat is set. Is this acceptable?
- I have made two const string which contains the possible configuration values for LogFormat. The best solution would be to group the values by making a LogFormat type. My attempts on that have failed so far. Flaeg keeps complaining about a missing parser for the type and I had no luck implementing one (are custom parser only ment for collections?). 

Would anyone mind having a look at my progress so far (I have not looked at the tests or documentation yet)? Any suggestions or feedback is much appreciated (a link is below).

Also, should it be possible to configure different LogFormats for the access log file and the Traefik error log?

https://github.com/SirCAS/traefik/commit/419dc0491f03e1625369313b81a6ab74dc820d40

bakins:
I did a simple straightforward json log format option in #1114 
It's not flexible, but it works well enough as a discussion starter.

Apologies - I didn't see this issue until I'd written my code.

PR #2056, opened 2017-09-05
add json format support for traefik logs

marco-jantke:
This PR adds the possibility to use JSON format for the traefik logs. 

Also it aligns the configuration to the accessLogs and deprecates the `traefikLogsFile` option.

Sample log line:
```json
{"level":"info","msg":"Traefik version dev built on I don't remember exactly","time":"2017-09-05T02:34:52+02:00"}
```

This fixes https://github.com/containous/traefik/issues/780.

### DIFF
--- a/cmd/traefik/configuration.go
+++ b/cmd/traefik/configuration.go
@@ -153,6 +153,12 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 	var defaultEureka eureka.Provider
 	defaultEureka.Delay = "30s"
 
+	// default TraefikLog
+	defaultTraefikLog := types.TraefikLog{
+		Format:   "common",
+		FilePath: "",
+	}
+
 	// default AccessLog
 	defaultAccessLog := types.AccessLog{
 		Format:   accesslog.CommonFormat,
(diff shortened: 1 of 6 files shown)


marco-jantke:
@Juliens thanks for the review! I added one more commit in order to keep the colors when the logs are written to stdout :D :)

marco-jantke:
@nmengin it's not about access logs :-) I could, of course, add an integration test for the traefik log itself, but I am not sure on what to assert.. What I could is pick some lines and verify that they are valid JSON, without looking into the contents. Do you think this would help?

marco-jantke:
This is certainly a good point Timo. I tested with all the different constellations with CLI and toml configuration. The behaviour is consistent and as expected. Given the new configuration `traefikLog.filePath` is specified, it will always be used. Only in case it is not specified the deprecated option will be used.

I now added a warning about the usage of the deprecated and documented at the deprecated option documentation, that the new option takes precedence.

To summarise shortly:

`traefik --traefiklogsfile=./deprecated_log --traefiklog --traefiklog.filepath=./new_log --traefiklog.format="json"` will write to `./new_log` in JSON format. 

`traefik --traefiklogsfile=./deprecated_log --traefiklog --traefiklog.format="json"` will write to `./deprecated_log` in JSON format.

PR #2062, opened 2017-09-05
Enable loss less rotation of log files

marco-jantke:
There is the chance that Traefik looses log lines for requests during rotation of the log file. This is at the moment true for the traefik log file as well as the access log files. I added two unit tests to prove this. This was also responsible for one flaky behaviour in the integration test [here](https://semaphoreci.com/containous/traefik/branches/pull-request-2054/builds/2).

For the traefik logs this is accomplished by opening the new log file before closing the previous one. The traefik logs use the global `logrus` instance and it is already protected by a mutex on rotation. For the access logs such a mutex also had to be introduced.

I think there is some potential in refactoring and aligning the two logging implementations in order to avoid duplicate code when it comes to the file handling and the logrus configuration. I will do this in a follow-up PR.

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -227,8 +227,10 @@ func RotateFile() error {
 		return nil
 	}
 
-	if err := CloseFile(); err != nil {
-		return fmt.Errorf("error closing log file: %s", err)
+	if logFile != nil {
+		defer func(f *os.File) {
+			f.Close()
+		}(logFile)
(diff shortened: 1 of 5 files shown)


ldez:
Could change the base PR to 1.4 branch?

marco-jantke:
Rebased and changed the base branch of the PR.

PR #2066, opened 2017-09-05
Fix IAM policy sid.

charlieoleary:
### Description
The required IAM policy uses spaces in the `Sid` field, but this is not allowed by IAM. As a result, attempting to create this policy results in a validation error. The fix is simply to remove the spaces from the `Sid`.


ldez:
Could change your PR to point to 1.4 branch?

Ask me if you need help.

charlieoleary:
@ldez Updated!

PR #2075, opened 2017-09-06
Fix whitespaces

chulkilee:
- use LF for new line
- remove trailing whitespaces, except two spaces at EOL in markdown
- add missing new line at the end of file


chulkilee:
I think it would be better to have this one-time large commit than having individual commit with whitespaces changes on irrelevant parts.

dtomcej:
How does this interact with our linters?

Is this something that we can add to our code cleaning tools to do automatically?

docs/basics.md
```diff
@@ -132,10 +132,10 @@ For instance, `PathPrefix: /products` would match `/products` but also `/product
 Since the path is forwarded as-is, your backend is expected to listen on `/products`.
 
 Use a `*Strip` matcher if your backend listens on the root path (`/`) but should be routeable on a specific prefix.
-For instance, `PathPrefixStrip: /products` would match `/products` but also `/products/shoes` and `/products/shirts`.
```
ldez: 2 spaces have a meaning in markdown.

> When you do want to insert a <br /> break tag using Markdown, you end a line with two or more spaces, then type return.
https://daringfireball.net/projects/markdown/syntax#p

docs/basics.md
```diff
@@ -132,10 +132,10 @@ For instance, `PathPrefix: /products` would match `/products` but also `/product
 Since the path is forwarded as-is, your backend is expected to listen on `/products`.
 
 Use a `*Strip` matcher if your backend listens on the root path (`/`) but should be routeable on a specific prefix.
-For instance, `PathPrefixStrip: /products` would match `/products` but also `/products/shoes` and `/products/shirts`.  
-Since the path is stripped prior to forwarding, your backend is expected to listen on `/`.
```
ldez: same here

docs/basics.md
```diff
@@ -132,10 +132,10 @@ For instance, `PathPrefix: /products` would match `/products` but also `/product
 Since the path is forwarded as-is, your backend is expected to listen on `/products`.
 
 Use a `*Strip` matcher if your backend listens on the root path (`/`) but should be routeable on a specific prefix.
-For instance, `PathPrefixStrip: /products` would match `/products` but also `/products/shoes` and `/products/shirts`.  
-Since the path is stripped prior to forwarding, your backend is expected to listen on `/`.  
-If your backend is serving assets (e.g., images or Javascript files), chances are it must return properly constructed relative URLs.
```
ldez: same here

docs/basics.md
```diff
@@ -132,10 +132,10 @@ For instance, `PathPrefix: /products` would match `/products` but also `/product
 Since the path is forwarded as-is, your backend is expected to listen on `/products`.
 
 Use a `*Strip` matcher if your backend listens on the root path (`/`) but should be routeable on a specific prefix.
-For instance, `PathPrefixStrip: /products` would match `/products` but also `/products/shoes` and `/products/shirts`.  
-Since the path is stripped prior to forwarding, your backend is expected to listen on `/`.  
-If your backend is serving assets (e.g., images or Javascript files), chances are it must return properly constructed relative URLs.  
-Continuing on the example, the backend should return `/products/shoes/image.png` (and not `/images.png` which Traefik would likely not be able to associate with the same backend).
```
ldez: same here

docs/basics.md
```diff
@@ -208,7 +208,7 @@ Finally, you can create a rule to bind multiple domains or Path to a frontend, u
 
 #### Rules Order
 
-When combining `Modifier` rules with `Matcher` rules, it is important to remember that `Modifier` rules **ALWAYS** apply after the `Matcher` rules.
```
ldez: same here

docs/basics.md
```diff
@@ -355,7 +355,7 @@ requests periodically carried out by Traefik. The check is defined by a path
 appended to the backend URL and an interval (given in a format understood by [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration)) specifying how
 often the health check should be executed (the default being 30 seconds).
 Each backend must respond to the health check within 5 seconds.
-By default, the port of the backend server is used, however, this may be overridden.
```
ldez: same here

docs/configuration/backends/dynamodb.md
```diff
@@ -53,11 +53,11 @@ SecretAccessKey = "123"
 Endpoint = "http://localhost:8080"
 ```
 
-Items in the `dynamodb` table must have three attributes: 
+Items in the `dynamodb` table must have three attributes:
 
 - `id` (string): The id is the primary key.
 - `name`(string): The name is used as the name of the frontend or backend.
-- `frontend` or `backend` (map): This attribute's structure matches exactly the structure of a Frontend or Backend type in traefik.  
-    See `types/types.go` for details.
```
ldez: same here

docs/configuration/backends/dynamodb.md
```diff
@@ -53,11 +53,11 @@ SecretAccessKey = "123"
 Endpoint = "http://localhost:8080"
 ```
 
-Items in the `dynamodb` table must have three attributes: 
+Items in the `dynamodb` table must have three attributes:
 
 - `id` (string): The id is the primary key.
 - `name`(string): The name is used as the name of the frontend or backend.
-- `frontend` or `backend` (map): This attribute's structure matches exactly the structure of a Frontend or Backend type in traefik.
```
ldez: same here

docs/basics.md
chulkilee: Oh I didn't know that. Then here are options:

- keep the two spaces at EOL
- replace the two spaces at EOL with `<br>`
- replace the two spaces at EOL with `\n\n` to separate paragraphs
- use different formatting, such as `!!! note` or lists 

I still don't like the significant trailing spaces since it's very easy to mess up. I think using `<br>` is better if `<br>` is really needed. Otherwise, the best option is to avoid `<br>` at all - use separate paragraphs or other formatting. 

Let me know - I'll make changes in this PR accordingly.

ldez: > keep the two spaces at EOL

docs/basics.md
chulkilee: There is no following lines after this line - this is the end of the paragraph. So I think it's okay to remove two spaces here.

PR #2080, opened 2017-09-07
Flaeg struct

ldez:
### Description

Related to https://github.com/containous/flaeg/pull/35


ldez:
I close for now, due to staert/flaeg interaction.

PR #2083, opened 2017-09-07
Remove old glide elements for integration tests.

ldez:
### Description

Remove old glide elements for integration tests.


ldez:
@timoreimann it's fixed 😉

Issue #2035, opened 2017-09-01
Add documentation for Error Pages on various configurations


kcrawley:
Feature / Documentation Request:

Error pages documentation only covers usage for file based configuration, see: https://docs.traefik.io/basics/#custom-error-pages

Documentation should either specify that other configurations are incompatible with the error pages middleware or examples should be provided for aforementioned configurations. Re: docker, swarm, etcd, etc.

Related to #1675

ReidWeb:
> Documentation should either specify that other configurations are incompatible with the error pages middleware or examples should be provided for aforementioned configurations. Re: docker, swarm, etcd, etc.

Which configurations are incompatible with the error pages middleware?

kcrawley:
@ReidWeb i wouldn't know, there are no documented examples for anything other than the file provider.

0xbaadf00d:
By example: How to enable the same error pages for all available backends ?  In microservice environment  its seems """idiot""" to have to set error pages per backend rather than globally.

timoreimann:
To my knowledge, error pages do not play together directly with dynamic providers such as Docker. (@bparli, please correct me if I'm wrong from the perspective of your PR.)

A possible workaround is to [customize the template for a particular provider](https://github.com/containous/traefik/tree/d54417acfe2be53021c60576e645964dc2d3a236/templates), apply whatever Go template logic seems necessary to incorporate error handlers, and reference the customized template in the Traefik configuration. The downside is that this won't be fully dynamic.

Feel free to file additional feature request to ask for functionality expansion.

kachkaev:
I'm using k8s and i'd be interested in seeing some statement about custom error pages in the docs for my provider. After digging through the repo I conclude there are not supported yet, but newcomers could be given an easier path to figure that out.

A link to some umbrella issue with tick boxes would be ideal (one can subscribe and be notified about a PR or a long-awaited change in a new release).

timoreimann:
I think it'd be good to use this issue in order to track leaving a general remark on the current restrictions of error pages.

An alternative path to trying to control error pages from various providers via labels might be to only extend error pages so much that the information about which frontend / backend failed is forwarded to the error-handling container. That way, it'd be totally up to the handler implementation to return per-service error pages.
This needs to be discussed separately though, likely on a feature request issue.

bparli:
[...]

PR #2104, opened 2017-09-11
Document custom error page restrictions.

timoreimann:
@containous/traefik PTAL.

In particular, someone with sufficient expertise please acknowledge that custom error pages can really be used with the key/value stores we support.

Refs #2035.

### DIFF
--- a/docs/configuration/commons.md
+++ b/docs/configuration/commons.md
@@ -242,6 +242,9 @@ Instead, the query parameter can also be set to some generic error page like so:
 Now the `500s.html` error page is returned for the configured code range.
 The configured status code ranges are inclusive; that is, in the above example, the `500s.html` page will be returned for status codes `500` through, and including, `599`.
 
+Custom error pages are easiest to implement using the file provider.
+For dynamic providers, the corresponding template file needs to be customized accordingly and referenced in the Traefik configuration.
+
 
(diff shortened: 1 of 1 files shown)


docs/configuration/commons.md
```diff
@@ -242,6 +242,8 @@ Instead, the query parameter can also be set to some generic error page like so:
 Now the `500s.html` error page is returned for the configured code range.
 The configured status code ranges are inclusive; that is, in the above example, the `500s.html` page will be returned for status codes `500` through, and including, `599`.
 
+Custom error pages are easiest to implement using the file provider or one of the supported K/V stores. For other dynamic providers, the corresponding template file needs to be customized accordingly and referenced in the Traefik configuration.
```
ldez: Could you add one sentence by line?

timoreimann: Done.

Juliens:
In fact, all the k/v store use kv provider

```go
type Provider struct {
	kv.Provider `mapstructure:",squash"`
}
```
and the kv provider use template https://github.com/containous/traefik/blob/master/provider/kv/kv.go#L149

and  there is no status in the template

timoreimann:
@Juliens thanks for digging into this. I updated the documentation.

Let me know what you think.

Issue #1682, opened 2017-05-29
storeconfig doesn't store backends and frontends to Consul


zyzop:
### What did you do?
I wanted to upload my `traefik.toml` to a Consul. I have several backends and frontends configured under `[file]` section, and it all runs fine.
So I've added a `[consul]` section in the end of traefil toml and did `storeconfig`

### What did you expect to see?
whole traefik.toml uploaded to Consul

### What did you see instead?
Everything is uploaded except my backends and frontends. There's simply no frontends or backends keys in Consul.

### Output of `traefik version`: (_What version of Traefik are you using?_)
```
1.3 rc3
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
debug = true
checkNewVersion = false
logLevel = "WARN"
InsecureSkipVerify = true
defaultEntryPoints = ["http", "https"]
[web]
address = ":8080"
[entryPoints]
  [entryPoints.http]
  address = ":80"
  [entryPoints.https]
  address = ":443"
[entryPoints.https.tls]
[acme]
email = "konstantin@somehost.io"
storage = "traefik/acme/account"
storageFile = "acme.json"
entryPoint = "https"
acmeLogging = true
onDemand = true
[[acme.domains]]
main = "somehost1.io"
[[acme.domains]]
main = "elib.somehost.ru"
[[acme.domains]]
main = "somehost.io"
sans = ["www.somehost.io"]
[file]
[backends]
[backends.backend1]
[backends.backend1.servers.server1]
url = "http://somehost1:80"
[backends.backend2]
[backends.backend2.servers.server1]
url = "http://somehost2.s3-website-eu-west-1.amazonaws.com:80"
[backends.backend3]
[backends.backend3.servers.server1]
url = "http://somehost3.s3-website-eu-west-1.amazonaws.com:80"
[frontends]
[frontends.frontend1]
backend = "backend1"
passHostHeader = true
[frontends.frontend1.routes.test_1]
rule = "PathPrefix: /"
[frontends.frontend2]
backend = "backend2"
passHostHeader = true
[frontends.frontend2.routes.test_1]
rule = "Host: somehost.io, www.somehost.io"
[frontends.frontend3]
backend = "backend3"
passHostHeader = true
[frontends.frontend3.routes.test_1]
rule = "Host:elib.somehost.ru"
[consul]
prefix = "traefik"
```

zyzop:
Can it be related to this? - https://github.com/containous/traefik/issues/1039

[...]

PR #2106, opened 2017-09-11
Added support for using storeconfig command to load frontends/backends configuration from file provider

### LINKED ISSUE #2036: "storeconfig" command can't upload frontends and backends configuration to Key-value Store
`traefik storeconfig -c web.toml`
This command only set main section to KV store and can't refresh.Frontends and backends need to be set manually?

## output

```shell
root@c7 traefik]# ./traefik storeconfig -c 321.toml 
2017/09/01 11:01:52 Storing configuration: {"GraceTimeOut":10000000000,"Debug":true,"CheckNewVersion":false,"AccessLogsFile":"log/access.log","TraefikLogsFile":"/var/log/traefik.log","LogLevel":"DEBUG","EntryPoints":{"http":{"Network":"","Address":":80","TLS":null,"Redirect":null,"Auth":null,"Compress":true},"https":{"Network":"","Address":":443","TLS":null,"Redirect":null,"Auth":null,"Compress":false}},"Cluster":null,"Constraints":[],"ACME":null,"DefaultEntryPoints":[],"ProvidersThrottleDuration":2000000000,"MaxIdleConnsPerHost":200,"IdleTimeout":360000000000,"InsecureSkipVerify":false,"Retry":null,"HealthCheck":{"Interval":30000000000},"Docker":null,"File":null,"Web":{"Address":":8080","CertFile":"","KeyFile":"","ReadOnly":false,"Statistics":{"RecentErrors":10},"Metrics":null,"Path":"","Auth":null},"Marathon":null,"Consul":null,"ConsulCatalog":null,"Etcd":{"Watch":true,"Filename":"","Constraints":[],"Endpoint":"10.15.223.37:2379,10.15.223.77:2379","Prefix":"/traefik","TLS":null,"Username":"","Password":""},"Zookeeper":null,"Boltdb":null,"Kubernetes":null,"Mesos":null,"Eureka":null,"ECS":null,"Rancher":null,"DynamoDB":null}
2017/09/01 11:01:52 Error running traefik: 102: Not a file (/traefik/entrypoints/http) [235360]
```

## traefik config:
```toml
debug = true
graceTimeOut = "10s"
checkNewVersion = false
traefikLogsFile = "/var/log/traefik.log"
logLevel = "DEBUG"
IdleTimeout = "360s"
accessLogsFile = "log/access.log"
[web]
address = ":8080"
  [web.statistics]
  RecentErrors = 10
[etcd]
endpoint = "10.15.223.37:2379,10.15.223.77:2379"
watch = true
prefix = "/traefik"

[entryPoints]
  [entryPoints.http]
  address = ":80"
  compress = true
  [entryPoints.https]
  address = ":443"

[backends]
  [backends.backend1]
    [backends.backend1.circuitbreaker]
      expression = "NetworkErrorRatio() > 0.5"
    url = "http://10.15.223.36:2379"
    weight = 5
    [backends.backend1.servers.server2]
    url = "http://10.15.223.77:2379"
    weight = 5
  [backends.backend2]
    [backends.backend3.servers.server1]
    url = "10.15.223.41:80"

[frontends]
  [frontends.frontend1]
  backend = "backend2"
    [frontends.frontend1.routes.test_1]
    rule = "Host: test1.com"
    passHostHeader = true
  [frontends.frontend2]
  backend = "backend1"
    [frontends.frontend2.routes.test_2]
    rule = "Path:/test"
```

## traefik version:

```
[root@c7 traefik]# ./traefik version
Version:      v1.3.7
Codename:     raclette
Go version:   go1.8.3
Built:        2017-08-25_08:36:23PM
OS/Arch:      linux/amd64
```

aantono:
Fixes #2036, #1682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Change Log
 
+## [v1.4.3](https://github.com/containous/traefik/tree/v1.4.3) (2017-11-14)
+[All Commits](https://github.com/containous/traefik/compare/v1.4.2...v1.4.3)
+
+**Bug fixes:**
+- **[consulcatalog]** Fix Traefik reload if Consul Catalog tags change ([#2389](https://github.com/containous/traefik/pull/2389) by [mmatur](https://github.com/mmatur))
+- **[kv]** Add Traefik prefix to the KV key ([#2400](https://github.com/containous/traefik/pull/2400) by [nmengin](https://github.com/nmengin))
+- **[middleware]** Flush and Status code ([#2403](https://github.com/containous/traefik/pull/2403) by [ldez](https://github.com/ldez))
+- **[middleware]** Exclude GRPC from compress ([#2391](https://github.com/containous/traefik/pull/2391) by [ldez](https://github.com/ldez))
+- **[middleware]** Keep status when stream mode and compress ([#2380](https://github.com/containous/traefik/pull/2380) by [Juliens](https://github.com/Juliens))
+
+**Documentation:**
+- **[acme]** Fix some typos ([#2363](https://github.com/containous/traefik/pull/2363) by [tomsaleeba](https://github.com/tomsaleeba))
+- **[docker]** Minor fix for docker volume vs created directory ([#2372](https://github.com/containous/traefik/pull/2372) by [visibilityspots](https://github.com/visibilityspots))
+- **[k8s]** Link corrected ([#2385](https://github.com/containous/traefik/pull/2385) by [xlazex](https://github.com/xlazex))
+
+**Misc:**
+- **[k8s]** Add secret creation to docs for kubernetes backend ([#2374](https://github.com/containous/traefik/pull/2374) by [shadycuz](https://github.com/shadycuz))
+
 ## [v1.4.2](https://github.com/containous/traefik/tree/v1.4.2) (2017-11-02)
 [All Commits](https://github.com/containous/traefik/compare/v1.4.1...v1.4.2)
 
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -2,7 +2,7 @@
 
 Træfik can be configured to use Kubernetes Ingress as a backend configuration.
 
-See also [Kubernetes user guide](/docs/user-guide/kubernetes). 
+See also [Kubernetes user guide](/user-guide/kubernetes). 
 
 
 ## Configuration
@@ -118,10 +118,10 @@ If one of the Net-Specifications are invalid, the whole list is invalid and allo
 ### Authentication
 
 Is possible to add additional authentication annotations in the Ingress rule.
-The source of the authentication is a secret that contains usernames and passwords inside the the key auth.
+The source of the authentication is a secret that contains usernames and passwords inside the key auth.
 
 - `ingress.kubernetes.io/auth-type`: `basic`
-- `ingress.kubernetes.io/auth-secret`  
+- `ingress.kubernetes.io/auth-secret`: `mysecret`  
     Contains the usernames and passwords with access to the paths defined in the Ingress Rule.
 
 The secret must be created in the same namespace as the Ingress rule.
--- a/docs/user-guide/docker-and-lets-encrypt.md
+++ b/docs/user-guide/docker-and-lets-encrypt.md
@@ -59,8 +59,8 @@ services:
       - web
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - /srv/traefik/traefik.toml:/traefik.toml
-      - /srv/traefik/acme.json:/acme.json
+      - /opt/traefik/traefik.toml:/traefik.toml
+      - /opt/traefik/acme.json:/acme.json
     container_name: traefik
 
 networks:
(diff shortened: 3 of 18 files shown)


timoreimann:
@aantono looks like the integration tests ran into the timeout. Could you change the commit hash of your HEAD (`git commit --amend --no-edit`) and force-push again?

aantono:
I'm wondering if we might also add a command flag to indicate if we want the file provider configuration to be persisted into KV?  When a `file.filename` argument is given (which is needed to load the static file config and store it in KV), it also ends up in the KV as an entry `traefik/file/*`, which is most likely undesired, as the purpose of using `storeconfig` function is to have Traefik consume config from remote/distributed KV store instead of a local file.  

Any thoughts on adding a check of something along those lines?:

```
filename := traefikConfiguration.GlobalConfiguration.File
if filename != nil {
	traefikConfiguration.GlobalConfiguration.File = nil
}
err = kv.StoreConfig(traefikConfiguration.GlobalConfiguration)

if filename != nil {
	// Store file config structure in KV
}
```

aantono:
Looks like CI server is having issues.  Just rebased against master, and build failed.  Trying to `git commit --amend --no-edit` didn't help :(

timoreimann:
Seems like Semaphore shut down the CI machine.

Closing and reopening PR to hopefully trigger a new run.

Juliens:
@aantono That's right, I think we must add a flag to storeconfig cmd in order to choose if we want to store the `filename` or the `file content`.

To keep the previous behaviour , we must store the `filename` by default WDYT

aantono:
@Juliens Yes, I was thinking of something along the same lines as well.  I will go ahead and update the PR to add the flag.  Can it be just added into the main `TraefikConfiguration` struct, along with `ConfigFile` or should something different be done?

I was thinking of something like:

```
type TraefikConfiguration struct {
  configuration.GlobalConfiguration `mapstructure:",squash" export:"true"`
  ConfigFile                         string `short:"c" description:"Configuration file to use (TOML)." export:"true"`
  ExportFileConfigInStoreConfig      bool `description:"Enable export of File Config to K/V Store Config." export:"true"`
}
```
then set the `ExportFileConfigInStoreConfig` value to `true` by default.

Thoughts?

Juliens:
I think it would be better to have a specific struct for storeconfig configuration
```golang
type StoreConfigConfiguration struct {
  configuration.TraefikConfiguration `mapstructure:",squash" export:"true"`
  ExportFileConfigInStoreConfig      bool `description:"Enable export of File Config to K/V Store Config." export:"true"`
}
```

aantono:
Seems like a duplication of `TraefikConfiguration` struct, plus addition of another field, as we still need the `ConfigFile` property, so that will have to stay.  Is it really worth the duplication and the complexity?

Juliens:
This duplication is needed because otherwise, we will have the exportfile option when we just want to start træfik.

aantono:
@Juliens I've tried your suggestion, but `flaeg` seems to not be happy with the divergent config types.  Getting this error: `Error reading TOML config file  : Command storeconfig : Config type doesn't match with root command config type.`  Thoughts?

ldez:
The problem is due to `parseAllSources` options for `staert`: this force to use the same type for the global command and the subcommand.
We must thinking about this.

aantono:
Any chance this PR might somehow miraculously make it into 1.4 final release? 😜

emilevauge:
@aantono Is there a use case to set `ExportFileConfigInStoreConfig=false` ? I would suggest to always export file config.
Other solution, create a new sub-command `storeFileConfig`.
WDYT?

aantono:
@emilevauge I would actually argue that in most use-cases we do *NOT* want to export the file config to the K/V store, as the file config typically is just a reference to the local file that should be used to load the command, when running `storeConfig` command.  If it is set to `true`, then the K/V store will also have the configuration telling Traefik to load config from a file (which will most likely not be present on the distributed cluster, otherwise why would the K/V be used in the first place to begin with?)  Does that make sense?

emilevauge:
@aantono If you use a KV store backend, I don't see any interest of using the file provider as you should probably put your frontends/backends in the KV store instead.
The file provider should only be used to bootstrap the KV store config IMHO.
WDYT?

aantono:
@emilevauge Exactly right, that's why I was saying that if anything, we should automatically remote exporting the `file.Filename` configuration portion to the K/V, if we do not want to make that configurable.  So for `storeConfig` command we should always exclude it then (which would be equal to `ExportFileConfigInStoreConfig=false`

emilevauge:
@aantono I'm not sure we are on the same line ;)
I think your PR is great, and we should really put backends/frontends from file provider in `storeconfig`.
IMHO, we should not make it configurable, it should be the standard behavior.
It does not make sense to have a file backend provider + kv store at the same time.
So to my mind, `storeconfig` should upload backends/frontends to the KV store if defined, and disable file backend in the KV store.
Does that make sense?

aantono:
Yes, I think we are actually exactly on the same page...

aantono:
@emilevauge Done!

emilevauge:
@aantono thanks a lot :heart_eyes: 
The code looks good to me :)
I would love a tiny integration test if you can. There is already one `TestCommandStoreConfig` in `consul_test.go`, it should be straightforward.

emilevauge:
@aantono sorry I closed this pull request by mistake while I was fixing something in the test you added :( Could you push force on your remote branch and reopen a pull request ? Thanks :pray:

emilevauge:
@aantono I created another PR #2419 carrying your commits :)

PR #2110, opened 2017-09-12
Add forward auth documentation.

ldez:

### DIFF
--- a/docs/configuration/entrypoints.md
+++ b/docs/configuration/entrypoints.md
@@ -119,10 +119,22 @@ Otherwise, the response from the auth server is returned.
 ```toml
 [entryPoints]
   [entryPoints.http]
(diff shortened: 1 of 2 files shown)


docs/configuration/entrypoints.md
```diff
@@ -109,9 +109,26 @@ Users can be specified directly in the toml file, or indirectly by referencing a
...
+
+```toml
+[entryPoints]
+  [entryPoints.http]
+  address = ":80"
+    # To enable forward auth on an entrypoint
+    [entryPoints.http.auth.forward]
+    address = "http://authserver.com/auth"
+```
+
```
nmengin: Maybe can we add a example to describe how to use Forward Authentication with TLS configuration?
WDYT?

Issue #2060, opened 2017-09-05
log Basic Auth username in accesslog


bastiaanb:
### Do you want to request a *feature* or report a *bug*?

Yes ;-)
I don't know whether the behaviour is already expected to work....

### What did you do?

* configured traefik to perform Basic Auth on the default entrypoint. 
* enabled access logging.
* started traefik:
```shell
docker run --rm -i -p 10080:80 -p 18080:8080 -v $(pwd)/traefik.toml:/etc/traefik/traefik.toml traefik:1.4.0-rc1
```
I performed a authenticated request with curl:
```shell
curl -H "Host: httpbin.org" --user test:test http://localhost:10080/headers
```

### What did you expect to see?
In the ClientUsername field of the access log I expect to see the user 'test' logged. E.g.
```shell
172.17.0.1 - test [05/Sep/2017:09:14:16 +0000] "GET /headers HTTP/1.1" 200 292 - "curl/7.51.0" 1 "frontend1" "http://httpbin.org:80" 241ms
```

### What did you see instead?
I see a dash instead:
```
172.17.0.1 - - [05/Sep/2017:09:14:16 +0000] "GET /headers HTTP/1.1" 200 292 - "curl/7.51.0" 1 "frontend1" "http://httpbin.org:80" 241ms
```

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.4.0-rc1
Codename:     roquefort
Go version:   go1.9
Built:        2017-08-29_08:35:24AM
OS/Arch:      linux/amd64
```

Same results with Traefik 1.3.7

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

<details>
<summary>configuration</summary>

```json
{
 "GraceTimeOut": 10000000000,
 "Debug": false,
 "CheckNewVersion": true,
 "AccessLogsFile": "/dev/stdout",
 "AccessLog": null,
 "TraefikLogsFile": "",
 "LogLevel": "ERROR",
 "EntryPoints": {
  "http": {
   "Network": "",
   "Address": ":80",
   "TLS": null,
   "Redirect": null,
   "Auth": {
    "Basic": {
     "Users": [
      "test:$apr1$H6uskkkW$IgXLP6ewTrSuBkTrqE8wj/",
      "test2:$apr1$d9hr9HBB$4HxwgUir3HP4EsggP/QNo0"
     ],
     "UsersFile": ""
    },
    "Digest": null,
    "Forward": null,
    "HeaderField": ""
[...]

PR #2111, opened 2017-09-12
Add username in accesslog

bastiaanb:
### Description

This PR fixes issue #2060: logging of authenticated username in accesslog. Works for both Basic and Digest authentication.

### DIFF
--- a/integration/access_log_test.go
+++ b/integration/access_log_test.go
@@ -1,7 +1,10 @@
 package integration
 
 import (
+	"crypto/md5"
+	"crypto/rand"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"os"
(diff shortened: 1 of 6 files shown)


bastiaanb:
@ldez could we still get this into 1.4? We need this feature to ditch nginx as http router.....

ldez:
@bastiaanb we are ok to put this in 1.4, but you need to add tests (https://github.com/containous/traefik/blob/master/integration/access_log_test.go)

bastiaanb:
@ldez I have adjusted the access_log_test to correctly reflect how and when the req.URL.User gets set (by the authenticator code). 
Is that sufficient? My guess is that the scenario where the URL.User is already set before invocation of the logger does not happen at all.

ldez:
@bastiaanb could you remove the merge of the master branch, and change the base PR branch to 1.4 (and rebase). Ask me if you need help.

We need to support the 2 cases:
- user from Traefik auth
- user from external

Consequently we need 2 tests.

Could you revert your changes in the test and add a new test?

marco-jantke:
If possible would it be maybe better to not modify the request object itself, but to extend the access log code to understand both sources? @bastiaanb WDYT?

IMHO it's not a good practice to modify the request object in such a way as it creates an implicit dependency between our auth and logging components. Think about what happens when we would replace the auth middleware with a 3rd party one that does it for us? Then the logging would again be broken.

bastiaanb:
@ldez oh, exeternal auth is actually used? Is a usage scenario available somewhere? I presume 'external auth' typically would be TLS client certificates?
I am a go-lang novice, so it will take a bit of time to write the proper test code....

bastiaanb:
@marco-jantke what would be the proper way to relay the username from the auth component to the access logger?

marco-jantke:
I am not an expert with the HTTP authentication methods, but it seems in both cases (Digest + Basic Auth) the username is transferred in plain text and we can simply read it from the request. For the basic auth this was working anyway already before, for Digest Auth we would need some parsing in order to get the username. See https://en.wikipedia.org/wiki/Digest_access_authentication#Example_with_explanation for the digest authentication process. Can you try if this is a feasible approach?

I would also like to hear the opinion of other Maintainers and whether others see the same benefits in making the logging middleware smarter to keep coupling of the components low. @containous/traefik

bastiaanb:
Hi Marco,


By duplicating the code that identifies (not authenticates) the user in the access log code, we would violate separation of concerns a lot more than by setting the username in the req.URL.User. It would need to be aware of all authentication methods that may be supported in the future (e.g. token, cookie, oauth2, JWT). Also the identity should show up in the access log only in case of successfull authentication.

So it seems to me that the proper way would be to define what data authenticators should add to the request context, e.g. user id, othe user attributes, authentication method, identity provider, etc. That way downstream http.HandlerFuncs can add these to the backend HTTP request headers for example (instead of having the auth code do this) or implement more sophisticated authorization than just auth-required.

But I don't really know yet how the accesslog code would be able to access this. Maybe via the same pattern as the SaveBackend handler?



________________________________
Van: Marco Jantke <notifications@github.com>
Verzonden: dinsdag 19 september 2017 13:25
Aan: containous/traefik
CC: Bastiaan Bakker; Mention
Onderwerp: Re: [containous/traefik] Add username in accesslog (#2111)


I am not an expert with the HTTP authentication methods, but it seems in both cases (Digest + Basic Auth) the username is transferred in plain text and we can simply read it from the request. For the basic auth this was working anyway already before, for Digest Auth we would need some parsing in order to get the username. See https://en.wikipedia.org/wiki/Digest_access_authentication#Example_with_explanation for the digest authentication process. Can you try if this is a feasible approach?

Digest access authentication - Wikipedia<https://en.wikipedia.org/wiki/Digest_access_authentication#Example_with_explanation>
en.wikipedia.org
Digest access authentication is one of the agreed-upon methods a web server can use to negotiate credentials, such as username or password, with a user's web browser.


I would also like to hear the opinion of other Maintainers and whether others see the same benefits in making the logging middleware smarter to keep coupling of the components low. @containous/traefik<https://github.com/orgs/containous/teams/traefik>

marco-jantke:
Thank you for the reasoning @bastiaanb!

> Also the identity should show up in the access log only in case of successful authentication.

This is a very good argument to go with your approach. About duplicating the code to extract the username, I slightly disagree, because you could, of course, extract the code and reuse it at both places. Anyway I think I am convinced to go with your approach. Going into code review now :)

ldez:
@bastiaanb any news?

bastiaanb:
@ldez sorry, I haven't been able to work on this issue in the last weeks at all. I hope to pick this up in two weeks after a (laptopless) holiday.

ldez:
@bastiaanb any news?

integration/access_log_test.go
```diff
@@ -184,6 +187,57 @@ func (s *AccessLogSuite) TestAccessLogAuthEntrypoint(c *check.C) {
...
+	waitForTraefik(c, "authEntrypoint")
+
+	// Verify Traefik started OK
+	traefikLog := checkTraefikStarted(c)
+
+	// Test auth entrypoint
+	req, err := http.NewRequest(http.MethodGet, "http://127.0.0.1:8004/", nil)
+	c.Assert(err, checker.IsNil)
+	req.Host = "entrypoint.auth.docker.local"
+	req.SetBasicAuth("test", "test")
```
marco-jantke: Can we do another test request that uses digest authentication to have both cases covered? Would be fine for me to just do it in this test case.

mmatur: Done @marco-jantke

integration/access_log_test.go
```diff
@@ -184,6 +190,166 @@ func (s *AccessLogSuite) TestAccessLogAuthEntrypoint(c *check.C) {
...
+	count := checkAccessLogExactValuesOutput(err, c, expected)
+
+	c.Assert(count, checker.GreaterOrEqualThan, len(expected))
+
+	// Verify no other Traefik problems
+	checkNoOtherTraefikProblems(traefikLog, err, c)
+}
+
+func (s *AccessLogSuite) TestAccessLogDigestAuthEntrypoint(c *check.C) {
+	// Ensure working directory is clean
```
marco-jantke: This comment adds no value.

integration/access_log_test.go
```diff
@@ -184,6 +190,166 @@ func (s *AccessLogSuite) TestAccessLogAuthEntrypoint(c *check.C) {
...
+	return fmt.Sprintf("%x", digest.Sum(nil))
+}
+
+func getCnonce() string {
+	b := make([]byte, 8)
+	io.ReadFull(rand.Reader, b)
+	return fmt.Sprintf("%x", b)[:16]
+}
+
+func getDigestAuthrization(digestParts map[string]string) string {
```
marco-jantke: Typo here: missing an "o".

Issue #2090, opened 2017-09-08
Add Service Fabric provider


jjcollinge:
### Do you want to request a feature or report a bug?
Feature

### Problem
[Service Fabric](https://azure.microsoft.com/en-gb/services/service-fabric/) is an orchestration platform built by Microsoft.
The platform currently has very limited routing capabilities without using other 1st party Microsoft cloud services (API Management). By integrating Traefik into Service Fabric we can support much more advanced routing scenarios and it opens up the platform to non .NET software.

Myself and @lawrencegripper have already forked traefik and added a Service Fabric provider. We also have customers who can make use of this and so we thought it might be useful to merge back into the master trunk for others to use. We would also be happy to support future fixes.

### Ask
Before we submit a PR, we wanted to open a discussion to see whether this is something you would consider merging and if you have any reservations or issues.

emilevauge:
Hi @jjcollinge & @lawrencegripper,
Awesome news :heart_eyes: !
We are in the process of adding a plugin mechanism to Traefik, and I think this provider may be a great fit. But we still don't have any ETA on this (golang plugins are not straightforward). In the meantime, we propose you to make a pull request on traefik in order to make a code review by maintainers /cc @containous/traefik. And you are welcome to join us on our [Slack #dev channel](https://traefik.herokuapp.com/) :smile:

ldez:
Closed by #2117

PR #2117, opened 2017-09-13
Add Service Fabric Provider

lawrencegripper:
### Description

Here is our PR to add a Service Fabric (SF) Provider into Traefik. Reference issue: #2090 

We’ve documented how the provider can be used. We wanted to check whether it was best to include these docs in the PR or to have them hosted externally. We have a [basic usage intro](https://github.com/jjcollinge/traefik-on-service-fabric/blob/master/Readme.MD) and an [advanced usage intro]( https://github.com/jjcollinge/traefik-on-service-fabric/blob/master/Docs/CustomTemplates.MD) doc.  

#### Notes:

For simple routing, we use Service Fabric's Application Parameters to define which services should be exposed. The provider picks up these parameters and then builds default frontends for each.

For more complex Traefik routing, we use a .tmpl file that can be pushed to SF’s configuration system. This is then read from disk by the provider and used to customise the Traefik setup. As such, the provider checks if a .tmpl file path has been provided, if not, is uses a glob search in the SF configuration directory to find the newest configuration available. Please note we have built this primarily for Windows based Service Fabric clusters. Testing on various cluster configurations is on going. We also plan to investigate supporting Service Fabric on Linux soon too.

Me and @jjcollinge both looking forward to hearing your feedback

Here is a quick screenshot of it up and running:

screen shot 2017-09-13 at 18 08 11


Juliens:
Thanks for this PR.

Here are my firsts remarks.

The others providers handle template in a different way than you do.

There is a default template which is embed in the traefik binary, and you can override this template with another by using the [`filename` parameter](https://github.com/containous/traefik/blob/f9225c54ff07925ba0ccd835a12ffc2c21133a4d/provider/provider.go#L28) in the provider configuration.
In order to have this mechanism, you only have to call [GetConfiguration](https://github.com/containous/traefik/blob/f9225c54ff07925ba0ccd835a12ffc2c21133a4d/provider/provider.go#L54) (it handles automatically the template overload)

Next, for example, in the Docker provider, there is no predefined `Rule`, the `Rule` is set by reading the container [labels](https://github.com/containous/traefik/blob/f9225c54ff07925ba0ccd835a12ffc2c21133a4d/provider/docker/docker.go#L434)

WDYT about using the [Parameter](https://github.com/jjcollinge/traefik-on-service-fabric/blob/master/Readme.MD#simple-configuration) for settings the `Rule` Traefik will use ?

For example:
```xml
<Parameters>
    ...
    <Parameter Name="traefik.rule" Value="PathPrefix:/my-services" />
    ...
</Parameters>
```
Moreover you can see that the docker provider handles a lots of other [configuration with labels](https://docs.traefik.io/configuration/backends/docker/#labels-overriding-default-behaviour)

WDYT about creating an external repo containing your Service Fabric client ? It would be easier to manage and could be reused by other developers.

Feel free to join us on [slack](https://traefik.herokuapp.com/)

jjcollinge:
Hey @Juliens, thanks for the feedback. Myself and @lawrencegripper are currently travelling with work but it'd be good to have a chat on Slack next week to discuss the PR in a bit more depth - would that work for you?

Juliens:
Sure, don't hesitate to ping me 😉

naeemkhedarun:
This is fantastic, thanks for your hard work! I'm giving the integration a go now with our clusters 👍

lawrencegripper:
@Juliens Thanks for the feedback, we'd like to accept both of the suggestions. 

We've chatted further and agree that removing the 'findtemplate' function is best way forward. We will go back to using the filename passed into the provider, like the other providers. We can change the deployment model in SF to have roughly the same effect when updating the template. 

I like the idea of allowing parameters for defining rules. This also reduces the need for custom template files. As the parameters sit in the application manifest, with the application containing multiple services, things would get a bit messy if we used it. I'll investigate to see if we can find a way to do this at the service level rather than the application manifest, we did look at this option before but may have missed something which would allow us to do it. 

Lastly we'll look at pulling out the ServiceFabric Client code into a separate repository to keep things clean.

lawrencegripper:
Hopefully completed the initial work to move over to using labels for configuration. Borrowed heavily from the docker provider, as suggested. 

The only difference is the way in which the labels are provided. On creation of a SF service an extension can be defined in the manifest with the labels you'd like configured. Like so: 

screen shot 2017-11-01 at 17 10 04


If you which to change these at runtime you can add an override into the SF Properties Manager, basically a metadata service. We have a PR to add the ability to edit these properties from the sfctl cli tool like so:

sfctl

Would be great to get your thoughts.

templates/servicefabric.tmpl
```diff
@@ -0,0 +1,143 @@
+[backends]
+    {{$groupedServiceMap := getServicesWithLabelValueMap .Services "backend.group.name"}}
+    {{range $aggName, $aggServices := $groupedServiceMap }}
+      [backends."{{$aggName}}"]
+      {{range $service := $aggServices}}
+        {{range $partition := $service.Partitions}}
+          {{range $instance := $partition.Instances}}
+            {{if and (isHealthy $instance) (hasHTTPEndpoint $instance)}}
```
Juliens: I think this kind of test/filter should be in the provider itself

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,418 @@
...
+
+		for _, service := range services.Items {
+			item := ServiceItemExtended{
+				ServiceItem: service,
+				Application: app,
+				Labels:      make(map[string]string),
+			}
+
+			addLabelsFromServiceExtension(sfClient, service.TypeName, &app, &item)
+			addLabelsFromPropertyManager(sfClient, &item)
```
Juliens: Could be directly handled in sfClient ?

templates/servicefabric.tmpl
lawrencegripper: Sure, makes more sense, I'll move this into the provider

provider/servicefabric/servicefabric.go
lawrencegripper: These calls wrap the underlying sfsdk calls GetServiceExtension and GetProperties we've aimed to pull as much out as we can while keeping the sdk generalized not traefik specific. 

`addLabelsFromPropertyManager` deserializes into an XML type, we've moved as much of the code as we can into the sdk but the type is specific to our usage in traefik. 

`addLabelsFromPropertyManager` is a very thin wrapper which filters the properties to only those with a `traefik.` prefix. 

Happy to take advice on different ways of handling this, hopefully we've ended up at a reasonable trade off

.gitignore
```diff
@@ -12,3 +12,5 @@
 *.exe
 .DS_Store
 /example/acme/acme.json
+/cmd/traefik/debug
+debug.test
```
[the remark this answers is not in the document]
lawrencegripper: Sorry, will remove. Related to how delve debugger hooks in with vscode.

glide.lock
```diff
@@ -95,6 +95,8 @@ imports:
 - name: github.com/coreos/etcd
   version: c400d05d0aa73e21e431c16145e558d624098018
   subpackages:
+  - Godeps/_workspace/src/github.com/ugorji/go/codec
+  - Godeps/_workspace/src/golang.org/x/net/context
```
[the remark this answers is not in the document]
lawrencegripper: Sure

provider/servicefabric/provider_test.go
```diff
@@ -0,0 +1,326 @@
...
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/containous/traefik/safe"
+	"github.com/containous/traefik/types"
+	sfsdk "github.com/jjcollinge/servicefabric"
+)
+
+type clientMock struct {
```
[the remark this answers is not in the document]
lawrencegripper: Will do

provider/servicefabric/provider_test.go
```diff
@@ -0,0 +1,315 @@
...
+		Items: []sfsdk.ApplicationItem{
+			{
+				HealthState: "Ok",
+				ID:          "TestApplication",
+				Name:        "fabric:/TestApplication",
+				Parameters: []*struct {
+					Key   string `json:"Key"`
+					Value string `json:"Value"`
+				}{
+
```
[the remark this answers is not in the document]
lawrencegripper: Will do

provider/servicefabric/provider_test.go
```diff
@@ -0,0 +1,315 @@
...
+		},
+		ID: "131496928082309293",
+	}
+	isHealthy := provider.isHealthy(replica)
+	if isHealthy {
+		t.Error("Incorrectly identified replica as healthy")
+	}
+}
+
+func compareConfigurations(actual, expected types.ConfigMessage) bool {
```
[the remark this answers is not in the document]
lawrencegripper: Will do

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,376 @@
...
+						ticker.Stop()
+						return nil
+					}
+				default:
+					log.Info("Checking service fabric config")
+				}
+
+				services, err := p.getClusterServices(sfClient)
+				if err != nil {
+					log.Error(err)
```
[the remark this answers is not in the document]
lawrencegripper: Agree

lawrencegripper:
@ldez and @Juliens Thanks for all the feedback, really appreciate it. I was going through each replying but seemed simpler to say - I agree with them all and have updated the code. 

Only section unchanged is the `addLabelsFromServiceExtension`, I've put an explanation on the comment. 

Let me know what you think!

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+		webClient,
+		p.ClusterManagementURL,
+		p.APIVersion,
+		p.ClientCertFilePath,
+		p.ClientCertKeyFilePath,
+		p.InsecureSkipVerify)
+	if err != nil {
+		return err
+	}
+	// provider.Constraints = append(provider.Constraints, constraints...)
```
ldez: Could you remove this line?

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+	"github.com/containous/traefik/provider"
+	"github.com/containous/traefik/safe"
+	"github.com/containous/traefik/types"
+	sfsdk "github.com/jjcollinge/servicefabric"
+)
+
+var _ provider.Provider = (*Provider)(nil)
+
+const keyPrefix = "traefik."
+const traefikExtensionName = "Traefik"
```
ldez: Could you create a group with constants?
```go
const (
      keyPrefix = "traefik."
      traefikExtensionName = "Traefik"
)
```

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+		return ""
+	}
+	return endpoint
+}
+
+func isHealthy(instanceData sfsdk.ReplicaItemBase) bool {
+	if instanceData.ReplicaStatus == "Ready" || instanceData.HealthState != "Error" {
+		return true
+	}
+	return false
```
ldez: it can be simplified:
```go
return instanceData.ReplicaStatus == "Ready" || instanceData.HealthState != "Error"
```

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+
+func hasHTTPEndpoint(instanceData sfsdk.ReplicaItemBase) bool {
+	_, err := getDefaultEndpoint(instanceData.Address)
+	return err == nil
+}
+func decodeEndpointData(endpointData string) (map[string]string, error) {
+	var endpointsMap map[string]map[string]string
+
+	if endpointData == "" {
+		return nil, errors.New("Endpoint data is empty")
```
ldez: Could you use a lower-case first letter?

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+
+	err := json.Unmarshal([]byte(endpointData), &endpointsMap)
+	if err != nil {
+		return nil, err
+	}
+	endpoints, endpointsExist := endpointsMap["Endpoints"]
+	if !endpointsExist {
+		return nil, errors.New("Endpoint doesn't exist in endpoint data")
```
ldez: Could you use a lower-case first letter?

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+	for _, v := range endpoints {
+		isHTTP := strings.Contains(v, "http")
+		if isHTTP {
+			defaultHTTPEndpoint = v
+			defaultHTTPEndpointExists = true
+			break
+		}
+	}
+	if !defaultHTTPEndpointExists {
+		return "", errors.New("No default endpoint found")
```
ldez: Could you use a lower-case first letter?

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+}
+
+func getNamedEndpoint(endpointData string, endpointName string) (string, error) {
+	endpoints, err := decodeEndpointData(endpointData)
+	if err != nil {
+		return "", err
+	}
+	endpoint, exists := endpoints[endpointName]
+	if !exists {
+		return "", errors.New("Endpoint doesn't exist")
```
ldez: Could you use a lower-case first letter?

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+	return endpoint, nil
+}
+
+// Add labels from service manifest extensions
+func addLabelsFromServiceExtension(sfClient sfsdk.Client, serviceType string, app *sfsdk.ApplicationItem, service *ServiceItemExtended) error {
+	extensionData := ServiceExtensionLabels{}
+	err := sfClient.GetServiceExtension(app.TypeName, app.TypeVersion, serviceType, traefikExtensionName, &extensionData)
+
+	if err != nil {
+		log.Error(err)
```
ldez: Seems not needed.

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+	return false
+}
+
+func (p *Provider) doesAppParamContain(app sfsdk.ApplicationItem, key, shouldContain string) bool {
+	value := p.getApplicationParameter(app, key)
+	if strings.Contains(value, shouldContain) {
+		return true
+	}
+	return false
```
ldez: it can be simplified:
```go
return strings.Contains(value, shouldContain)
```

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+	return results
+}
+
+func (p *Provider) isPrimary(instance sfsdk.ReplicaInstance) bool {
+	_, data := instance.GetReplicaData()
+	primaryString := "Primary"
+	if data.ReplicaRole == primaryString {
+		return true
+	}
+	return false
```
ldez: it can be simplified:
```go
return data.ReplicaRole == "Primary"
```

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+							log.Error(err)
+						} else {
+							for _, instance := range instances.Items {
+								if isHealthy(*instance.ReplicaItemBase) && hasHTTPEndpoint(*instance.ReplicaItemBase) {
+									partitionExt.Instances = append(partitionExt.Instances, instance)
+								}
+							}
+							if len(partitionExt.Instances) > 0 {
+								partitionExt.HasInstances = true
+							}
```
ldez: it can be simplified:
```go
partitionExt.HasInstances = len(partitionExt.Instances) > 0
```

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,420 @@
...
+							log.Error(err)
+						} else {
+							for _, instance := range replicas.Items {
+								if isHealthy(*instance.ReplicaItemBase) && hasHTTPEndpoint(*instance.ReplicaItemBase) {
+									partitionExt.Replicas = append(partitionExt.Replicas, instance)
+								}
+							}
+							if len(partitionExt.Replicas) > 0 {
+								partitionExt.HasReplicas = true
+							}
```
ldez: it can be simplified:
```go
partitionExt.HasReplicas = len(partitionExt.Replicas) > 0
```

provider/servicefabric/types.go
```diff
@@ -0,0 +1,40 @@
...
+	Application     sfsdk.ApplicationItem
+	Partitions      []PartitionItemExtended
+	Labels          map[string]string
+}
+
+// PartitionItemExtended provides a flattened view
+// of a services partitions
+type PartitionItemExtended struct {
+	sfsdk.PartitionItem
+	HasReplicas  bool
```
ldez: Seems not needed.

provider/servicefabric/types.go
```diff
@@ -0,0 +1,40 @@
...
+	Replicas     []sfsdk.ReplicaItem
+	HasInstances bool
```
ldez: Seems not needed.

lawrencegripper:
@ldez Thanks - I've fixed those up.

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,404 @@
...
+				result[value] = append(matchingServices, service)
+			} else {
+				result[value] = []ServiceItemExtended{service}
+			}
+		}
+	}
+	return result
+}
+
+func (p *Provider) getServicesWithLabel(services []ServiceItemExtended, key string) []ServiceItemExtended {
```
ldez: seems unused

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,404 @@
...
+}
+
+func (p *Provider) hasServiceLabel(service ServiceItemExtended, key string) bool {
+	_, exists := service.Labels[key]
+	return exists
+}
+
+func (p *Provider) getServiceLabelValue(service ServiceItemExtended, key string) string {
+	value, _ := service.Labels[key]
+	return value
```
ldez:
```go
return service.Labels[key]
```

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,404 @@
...
+
+		for _, service := range services.Items {
+			item := ServiceItemExtended{
+				ServiceItem: service,
+				Application: app,
+				Labels:      make(map[string]string),
+			}
+
+			addLabelsFromServiceExtension(sfClient, service.TypeName, &app, &item)
+			addLabelsFromPropertyManager(sfClient, &item)
```
ldez: error return value not checked with the function

provider/servicefabric/servicefabric.go
```diff
@@ -0,0 +1,404 @@
...
+		p.APIVersion,
+		p.ClientCertFilePath,
+		p.ClientCertKeyFilePath,
+		p.InsecureSkipVerify)
+	if err != nil {
+		return err
+	}
+
+	p.updateConfig(configurationChan, pool, sfClient, time.Second*10)
+	return nil
```
ldez: error return value of `p.updateConfig` are not checked.

```go
return p.updateConfig(configurationChan, pool, sfClient, time.Second*10)
```

lawrencegripper:
@ldez Fixed up all of those. @Juliens we took another look and have moved the XML stuff into the SDK for now, hopefully we'll be able to do something better in the future but will keep the API so traefik code will be unaffected. The two add methods are removed, new getServiceLabels call is here https://github.com/lawrencegripper/traefik/blob/b0030068320124044dafa4dcc14ee6065c673979/provider/servicefabric/servicefabric.go#L138

Hopefully all set, let me know if there are other changes

naeemkhedarun:
@lawrencegripper I can see there is an explicit list of supported traefik configuration supported. For example if I want to use `traefik.frontend.passTLSCert` would it add it to the front-end rule if defined in my ServiceManifest? Some of our systems need end-to-end SSL and I want to avoid needed to manually edit the Traefik configuration file if it can be done in the manifest files.

ldez:
The code has been moved [here](https://github.com/containous/traefik-extra-service-fabric)

server/server.go
```diff
@@ -559,6 +559,9 @@ func (s *Server) configureProviders() {
 	if s.globalConfiguration.DynamoDB != nil {
 		s.providers = append(s.providers, s.globalConfiguration.DynamoDB)
 	}
+	if server.globalConfiguration.ServiceFabric != nil {
```
lawrencegripper: I think this should be 's' rather than 'server'. Happy for me to push a fix?

ldez: oh yes, I missed that

uditjh:
what is the networking mode to be used by the applications when load balancing them using traefik?
https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-networking-modes

ldez:
You can go to the dedicate channel (#service-fabric) in the [Traefik community Slack channel](https://traefik.herokuapp.com) to ask question if you want.

PR #2119, opened 2017-09-14
Add CLI configuration for Web backend authentication.

ldez:
### Description

Add CLI configuration for Web backend authentication.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -303,10 +303,10 @@ type Cluster struct {
 
 // Auth holds authentication configuration (BASIC, DIGEST, users)
 type Auth struct {
-	Basic       *Basic
-	Digest      *Digest
-	Forward     *Forward
-	HeaderField string
+	Basic       *Basic   `description:"Basic authentication."`
+	Digest      *Digest  `description:"Digest authentication."`
+	Forward     *Forward `description:"Forward authentication."`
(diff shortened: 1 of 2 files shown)


types/types.go
```diff
@@ -303,10 +303,10 @@ type Cluster struct {
 
 // Auth holds authentication configuration (BASIC, DIGEST, users)
 type Auth struct {
-	Basic       *Basic
-	Digest      *Digest
-	Forward     *Forward
-	HeaderField string
+	Basic       *Basic   `description:"Basic authentication."`
```
Juliens:
```
 --web.auth                                 Authentication for dashboard and API                                             (default "true")
 --web.auth.basic                           Basic authentication.                                                            (default "true")
 --web.auth.digest                          Digest authentication.                                                           (default "true")
 --web.auth.forward                         Forward authentication.                                                          (default "true")
```

`default: true` ? how can I set the users ?

ldez: `default:true`: it's a flaeg bug see #2080

ldez:
```go
type Basic struct {
	Users     `mapstructure:","`
	UsersFile string
}
```

`Users` is a slice of string, and it's used as a compostion for `Basic`.

 It's impossible to handle this in the cli without breaking the TOML.

PR #2133, opened 2017-09-18
redirect to another entryPoint per frontend

### LINKED ISSUE #1456: Enable http to https redirection per frontend

SantoDE:
### Description

This PR makes it possible, to enable a redirect to another entrypoint per frontend. Current Supported Backends are:

- Docker
- Kubernetes
- Rancher

The only thing you need to do is, set e.g. `traefik.frontend.redirect=https`.

If this redirect is enabled, it's shown on the dashboard as well.

eingefugtes_bild_18_09_17__17_00

Fixes #1456, #541


SantoDE:
Hey @emilevauge ,
could you check again please? :)

dtomcej:
@SantoDE Do you think that you could leverage the Headers middleware to accomplish this?

By setting the `SSLRedirect = true` for that frontend will invoke the middleware, and will provide redirection. That would also allow different hosts, 301 vs 302 redirects etc. 

I am currently implementing the full gamut of options in 2030. Unfortunately, I have been delayed due to school.

PR 2146 is attempting to do part of this.

@containous/traefik Thoughts?

SantoDE:
@dtomcej we would then be doing to different kind of rewrites for the "same" case. The entryPoint https redirect works the same way by utilizing the rewrite middleware. Im personally not bound to either way, but we should use the same in both cases

dtomcej:
@SantoDE I don't disagree. Unfortunately we have a bunch of different places to acoomplish the same thing. I don't want to duplicate or introduce duplicate code. All the providers have access to invoke the headers middleware, and so might be a clean way to invoke the same functions.

Although, we should decide on if that is the way we want to proceed.

dtomcej:
An example would be that you introduce an `httpsredirect` property to the frontend struct. There is already a `Header` property there, that has a `SSLRediret` property inside. That would allow you to use already-existing code to accomplish the same function.

SantoDE:
Would you then move the entryPoint redirect behavior as well? I personally don't like to utilize different rewrites from entryPoint to frontends.

@emilevauge halp! :)

kachkaev:
Hi @SantoDE @emilevauge! Given that there are still a couple of other features pending for 1.4, do you think this one could also somehow fit the same release? Configrinfg the behavior of https redirects sounds like a great option for many mixed-use environments (especially those serving legacy pre-letsencrypt stuff). Agree with @SantoDE that the PR will close #1456 (which has relatively high number of 👍).

BTW there seems to exist an intersection with #2146 if I’m not wrong.

SantoDE:
Hey @emilevauge , @dtomcej and more. 

I updated my code like you suggested. I also changed the web ui once again.

eingefugtes_bild_30_09_17__22_57

Looking forward to your feedback :)

@kachkaev yeah there is some intersection but it's not too bad :) This PR will be included in 1.5 I guess.

kachkaev:
@emilevauge would it be possible for this PR to jump into `1.4`? 🙏

ldez:
@kachkaev We put only bug fix on `1.4`, but we hope to release (final) 1.5 before the end of the year

SantoDE:
@emilevauge done :)

docs/configuration/backends/docker.md
```diff
@@ -184,6 +185,7 @@ Services labels can be used for overriding default behaviour
 | `traefik.<service-name>.frontend.passHostHeader`  | Overrides `traefik.frontend.passHostHeader`.                                                     |
 | `traefik.<service-name>.frontend.priority`        | Overrides `traefik.frontend.priority`.                                                           |
 | `traefik.<service-name>.frontend.rule`            | Overrides `traefik.frontend.rule`.                                                               |
+| `traefik.<service-name>.frontend.redirect.https`  | Overrides `traefik.frontend.redirect.https`.                                                     |
```
nmengin: The service documentation may be updated too :wink: 

s/`traefik.<service-name>.frontend.redirect.https`  | Overrides `traefik.frontend.redirect.https`./`traefik.<service-name>.frontend.redirect`  | Overrides `traefik.frontend.redirect`.

docs/configuration/backends/kubernetes.md
```diff
@@ -88,6 +88,8 @@ Annotations can be used on containers to override default behaviour for the whol
     Override the default frontend rule type. Default: `PathPrefix`.
 - `traefik.frontend.priority: 3`  
     Override the default frontend rule priority.
+- `traefik.frontend.redirect.https: true`: 
+    Enables HTTP to HTTPS Redirect.
```
nmengin: Update k8s documentation to extends the description to all entrypoints.

docs/configuration/backends/rancher.md
```diff
@@ -123,4 +123,5 @@ Labels can be used on task containers to override default behaviour:
 | `traefik.frontend.passHostHeader=true`       | forward client `Host` header to the backend.                                             |
 | `traefik.frontend.priority=10`               | override default frontend priority                                                       |
 | `traefik.frontend.entryPoints=http,https`    | assign this frontend to entry points `http` and `https`. Overrides `defaultEntryPoints`. |
-| `traefik.frontend.auth.basic=EXPR`           | Sets basic authentication for that frontend in CSV format: `User:Hash,User:Hash`.        |
+| `traefik.frontend.auth.basic=EXPR`           | Sets basic authentication for that frontend in CSV format: `User:Hash,User:Hash`.        |   
+| `traefik.frontend.redirect.https=true`       | Enables HTTP to HTTPS Redirect on that Frontend                                          |
```
nmengin: Update rancher documentation to extends the description to all entrypoints.

provider/docker/docker.go
```diff
@@ -711,6 +722,13 @@ func (p *Provider) getBasicAuth(container dockerData) []string {
 	return []string{}
 }
 
+func (p *Provider) getRedirect(container dockerData) string {
+	if httpsRedirect, err := getLabel(container, types.LabelFrontendRedirect); err == nil {
+		return httpsRedirect
```
nmengin: WDYT to rename `httpsRedirect` variable into `entryPointredirect` or other?

provider/kubernetes/kubernetes.go
```diff
@@ -50,6 +50,7 @@ type Provider struct {
 	Namespaces             Namespaces `description:"Kubernetes namespaces"`
 	LabelSelector          string     `description:"Kubernetes api label selector to use"`
 	lastConfiguration      safe.Safe
+	Redirect               string `description:"Kubernetes Redirect"`
```
dtomcej: Similar to what @nmengin suggested from the @containous/docker team, I think we should clearly identify that this should be an `entrypointRedirect`

server/server.go
```diff
@@ -935,6 +935,17 @@ func (server *Server) loadConfig(configurations types.Configurations, globalConf
 						log.Infof("Configured IP Whitelists: %s", frontend.WhitelistSourceRange)
 					}
 
+					if len(frontend.Redirect) > 0 {
+						proto := frontend.Redirect
```
dtomcej: This may be a silly question, but what happens if the entrypoint is not named for the proto its using?

For example, if I named an entrypoint `secure` for `:443` and `insecure` for `:80`?

marcwaz:
@SantoDE I have tested your forked version of traefik with the https_redirect feature. I notice that with a docker swarm it doesn't work. I used the labels below in a docker compose file.
        labels:
          - "traefik.port=80"
          - "traefik.docker.network=frontnet"
          - "traefik.backend=app"
          - "traefik.frontend.rule=Host:app.com,www.app.com"
          - "traefik.enable=true"
          - "traefik.frontend.entryPoints=http,https"
          - "traefik.frontend.redirect=https"

The result is the following:
curl -I http://www.app.com -> 302 redirect to https://www.app.com => GOOD
curl -I https://www.app.com -> 302 redirect to https://www.app.com => BAD LOOP

My toml file is the following:
debug = false
checkNewVersion = true
logLevel = "DEBUG"
defaultEntryPoints = ["https","http"]

[entryPoints]
  [entryPoints.http]
  address = ":80"
  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]

[retry]

[docker]
domain = "app.com"
swarmmode = true
watch = true
exposedbydefault = true

[acme]
email = "email@app.com"
storage = "acme.json"
entryPoint = "https"
onHostRule = true
onDemand = false
acmeLogging = true

[web]
address = ":8080"
readOnly = true

marcwaz:
@SantoDE After lots of researches, I saw that it works if I use traefik services.
 
        labels:
          - "traefik.enable=true"
          - "traefik.port=80"
          - "traefik.frontend.rule=Host:app.com,www.app.com"
          - "traefik.docker.network=frontnet"
          - "traefik.backend=mybackend"

          - "traefik.http.frontend.entryPoints=http"
          - "traefik.http.frontend.redirect=https"

          - "traefik.https.frontend.entryPoints=https"
#          - "traefik.https.frontend.redirect=http"

It's also interesting to notice that you can also have a https to http redirect if needed (of course not the two at same time) !

mariusstaicu:
can't wait for this to be merged, very nice work @SantoDE

emilevauge:
Ping @SantoDE, any news on this one ?

server/server.go
SantoDE: That's a good question. I have no idea how to fix this properly tough. Maybe we could use this as a takeway for a further refacto? I guess the use-case is pretty rare

marcwaz:
It's important to have a 301 http permanent redirection and not a 302 temporary redirection. This point need to be fixed when the merge will be done !

SantoDE:
Ping @emilevauge , @dtomcej , @nmengin 

Feedback applied :)

tiangolo:
Merged! Awesome news! :tada: Thanks @SantoDE for the work!

Traefik team: when do you think this feature would be available in the [public Docker images](https://hub.docker.com/_/traefik/)? Which tag would it be in? `1.4.4` ?

ldez:
@tiangolo We put only bug fix in the branch 1.4 (https://github.com/containous/traefik#release-cycle)

As you see the milestone is 1.5.

tourshi:
I've tried to make use of this https redirect in v1.5.0-rc3 but can't seem to get it to work. I've tried multiple scenarios, non of which result in http to https redirects.
```
        labels:
            - traefik.backend=host.com
            - traefik.frontend.rule=Host:host.com,www.host.com
            - traefik.docker.network=traefik
            - traefik.frontend.entryPoints=http
            - traefik.frontend.redirect=https
            - traefik.enable=true
```

Am I misunderstanding something here?

Thank you!

ldez:
@tourshi Due to #2570:

```yml
        labels:
            - traefik.backend=host.com
            - traefik.frontend.rule=Host:host.com,www.host.com
            - traefik.docker.network=traefik
            - traefik.frontend.entryPoints=http
            - traefik.frontend.redirect.entryPoint=https
            - traefik.enable=true
```

tourshi:
Thank you @ldez I gave the regex/replacement solution a try, but it seems to break my traefik setup. These are the labels I used: 
`            - traefik.frontend.redirect.regex="^http:\/\/host.com\/(.*)"
             - traefik.frontend.redirect.replacement="https:\/\/host.com\/$1"`

In any case, I want to thank you for your help! I did come across this label: 
`traefik.frontend.headers.SSLRedirect=true` 
which solves my basic need of redirecting all http to https

Thank you!

h-bragg:
I have a local setup that by default redirects http to https. However I would like a single container to work on http only (for legacy reasons [as always]). I couldn't figure out which combinations of labels would work to enable that.

in effect I would like: 
- http->http https->http for a single host
- http->https https->https for everything else

The setup is plain old docker running locally with the following default configuration:

```toml
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    entryPoint = "https"
  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]
```

I have tried:
```yaml
    labels:
      traefik.http.frontend.entryPoints: http
      traefik.http.frontend.redirect: http
      traefik.https.frontend.entryPoints: https
      traefik.https.frontend.redirect: http
```

```yaml
    labels:
      traefik.frontend.headers.SSLRedirect: 'false'
```

```yaml
    labels:
      traefik.frontend.entryPoints: http
      traefik.frontend.redirect.entryPoint: http
```

and various combinations of the above. Am I missing something obvious?

DMW007:
I also tried to force redirect from http to https on single containers. Background: I'm using LetsEncrypt for some public subdomains, but also have locally ones. To avoid certificate errors on the local ones, I want only use http here. 

```
    labels: 
      - "traefik.frontend.entryPoints=http"
      - "traefik.frontend.redirect.entryPoint=https"
      - "traefik.frontend.passHostHeader=true"
```

This redirects from http to https, but ended with 404 cause we've no entrypoint on https here. But when I add it using `- "traefik.frontend.entryPoints=http,https"` I ended in an infinite loop, cause it always try to redirect to https, also when already serving with https. Trying @marcwaz solution I ended in https working, but no redirect any more. 

The only workaround I found is using a dummy-container for redirection:

```
  nextcloud_http_redirector:
    image: nginx
    container_name: nextcloud_http_redirector
    mem_limit: 32MB
    restart: always
    labels: 
      - "traefik.frontend.rule=Host:mytld"
      - "traefik.frontend.redirect.entryPoint=https"

  nextcloud_nginx:
    image: nginx
    container_name: nextcloud_nginx
    mem_limit: 32MB
    restart: always
    
    labels: 
      - "traefik.port=80"
      - "traefik.frontend.rule=Host:mytld"
      - "traefik.frontend.entryPoints=https"
```

As I understand this pull-request, I should be able to use the following configuration:

```
    labels: 
      - "traefik.port=80"
      - "traefik.frontend.entryPoints=http,https"
      - traefik.frontend.redirect=https
      - "traefik.frontend.rule=Host:mytld"
```
This should server https fine and redirect http to https. But it doesn't work, the site is avaliable on http and https without any redirection. I'm using the latest docker-image `traefik:1.5-alpine`.

DMW007:
@h-bragg The following works for me:

**traefik.toml**
```
[entryPoints]
[entryPoints.http]
address = ":80"

[entryPoints.https]
address = ":443"
[entryPoints.https.tls]
```

**Docker-Compose labels**
```
    labels: 
      - "traefik.port=80"
      - "traefik.frontend.rule=Host:MyDomain"
      - "traefik.frontend.entryPoints=http"
      - "traefik.frontend.redirect.entryPoints=https"
      - "traefik.frontend.entryPoints=https"
```
In this case, I force redirection from http to https. If you have another container which should use http only, apply the following config to avoid any redirection: 

```
    labels:
      - "traefik.port=80"
      - "traefik.frontend.rule=Host:MyInsecureDomain"
      - "traefik.frontend.entryPoints=http"
```

DMW007:
@h-bragg The `SSLRedirect` header works but you've to pass it like any other label with equal sign and not colon. In addition to this, you need both http and https entrypoints. Without http we get an 404. 

Working example for redirect http to https:

```
   labels: 
      - "traefik.port=80"
      - "traefik.frontend.rule=Host:MySecureDomain"
      - "traefik.frontend.headers.SSLRedirect=true"
      - "traefik.frontend.entryPoints=http,https"
```
For your other container, simply drop the redirect-header and https entrypoint like this:

```
    labels:
      - "traefik.port=80"
      - "traefik.frontend.rule=Host:MyInsecureDomain"
      - "traefik.frontend.entryPoints=http"
```
I tested this using latest stable docker release (`traefik:1.5-alpine` at the time of writing) and it works. Please keep in mind, that browser tend to cache redirects. So I used `wget` on the command line, there you can see the plain redirect without any caching issues.

h-bragg:
@DMW007 thanks, I have it working like this for now.

Was looking for it to be ssl by default with http as the exception (each thing connecting to it might be done by a different person, so the least amount of configuration required is better).

mhumeSF:
Wanted to test this out locally. I'm in a use case where ssl termination is happening at a load balancer before hitting traefik, but I still want to redirect http requests to https. Getting this error. Is there a way to disable ssl, or using a different port to handle http requests?

```
±  curl -I -H Host:web.localhost https://127.0.0.1:443
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to 127.0.0.1:443 
```

docker-compose.yml
```
version: '2'

services:
  web:
    image: nginx:alpine
    ports:
      - "80"
    labels:
      - "traefik.enable=true"
      - "traefik.frontend.rule=Host:web.localhost"
      - "traefik.port=80"
      - "traefik.frontend.headers.SSLRedirect=true"
      - "traefik.frontend.entryPoints=http,https"

  proxy:
    image: traefik
    command: --api --docker --docker.domain=docker.localhost --logLevel=DEBUG
    ports:
      - "80:80"
      - "443:443"
      - "8080:8080"
    volumes:
      - /dev/null:/traefik.toml
```

PR #2142, opened 2017-09-19
Removing double wording from documentation

jtyr:
### Description

This PR is removing a double wording which appears in the list of features. The same typo is also present on the official web page (https://traefik.io/).


timoreimann:
I don't think your branch is up to date since the issue was seemingly fixed in #2095, in particular  [this commit](https://github.com/containous/traefik/commit/c7c9349b00a899d9618e1e8cf4997a2c93e6f630).

PR #2143, opened 2017-09-19
Fix deprecated IdleTimeout config

marco-jantke:
This PR fixes loading of the deprecated top-level option `IdleTimeout`. Also, it changes the semantics so that the deprecated option is now preferred and a warning is logged when it is used.

_All the different configuration constellations I also verified via the matching CLI parameters as well._

I tried the following configs and they deliver now the expected results:

```toml
# default config, nothing specified
```
-> default timeout of 30 seconds is used

```toml
# only deprecated option specified
idleTimeout="42s"
```
-> deprecated option of 42 seconds is used

```toml
# deprecated option with RespondingTimeouts "activated"
idleTimeout="42s"
[respondingTimeouts]
```
-> deprecated option of 42 seconds is used

```toml
# deprecated option and new option specified
idleTimeout="42s"
[respondingTimeouts]
  idleTimeout="23s"
```
-> deprecated option of 42 seconds is used

```toml
# deprecated option *other* new option specified
idleTimeout="42s"
[respondingTimeouts]
  readTimeout="23s"
```
-> deprecated option of 42 seconds is used

```toml
# only new option specified
[respondingTimeouts]
  idleTimeout="23s"
```
-> new option of 23 seconds is used

While working on this PR I realized that also the health check has "problems" with its default interval, given it is configured in the following way (imho no one would do such a thing, but still...). As this was at related code places as my change I decided to include it in this PR:

```toml
[healthcheck]

[backends]
  [backends.backend1]
    [backends.backend1.healthCheck]
    path="/health"
    url = "http://localhost:8081"
```
-> interval is initialised with 0 and will lead to `Error in Go routine: non-positive interval for NewTicker`

### DIFF
--- a/cmd/traefik/configuration.go
+++ b/cmd/traefik/configuration.go
@@ -159,6 +159,11 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 		FilePath: "",
 	}
 
+	// default HealthCheckConfig
+	healthCheck := configuration.HealthCheckConfig{
+		Interval: flaeg.Duration(configuration.DefaultHealthCheckInterval),
+	}
+
 	// default RespondingTimeouts
 	respondingTimeouts := configuration.RespondingTimeouts{
 		IdleTimeout: flaeg.Duration(configuration.DefaultIdleTimeout),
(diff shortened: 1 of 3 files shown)


marco-jantke:
@timoreimann as you were also involved in this, do you want to have a look? :)

timoreimann:
Just to be super sure, I'd also test this kind of case:

```
# deprecated option *other* new option specified
idleTimeout="42s"
[respondingTimeouts]
  readTimeout="23s"
```

This should cause an effective idle timeout of 42 seconds to be applied.

marco-jantke:
Thanks for the note Timo. I verified it successfully and added it in my PR description above.

PR #2146, opened 2017-09-19
Allow configuration of SSLRedirect and SSLProxyHeaders from ingress annotations

mattatcha:
### Description

Provides the ability to set kubernetes provider defaults for SSLRedirect and SSLProxyHeaders with per frontend overrides via annotations on ingress resources. 

Follows k8s ingress annotation convention for SSLRedirect and adds a namespaced SSLProxyHeaders annotation.

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -327,18 +327,16 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 	return &templateObjects, nil
 }
 
+// getSSLProxyHeaders
 func (p *Provider) getSSLProxyHeaders(str string) map[string]string {
+	parts := append(p.DefaultSSLProxyHeaders, strings.Split(str, ",")...)
 	headers := map[string]string{}
-	kvs := append(p.DefaultSSLProxyHeaders, strings.Split(str, ",")...)
-	for _, kvp := range kvs {
-		kv := strings.SplitN(kvp, "=", 2)
-		if kv[0] == "" {
+	for _, part := range parts {
+		key, val := splitKeyValue(part)
+		if key == "" {
 			continue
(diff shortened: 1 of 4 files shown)


davidlumley:
LGTM! 👍

timoreimann:
I'm a bit uncertain if the SSL redirect annotation name makes sense for Traefik: The PR calls it `force-ssl-redirect`, which the [Nginx documentation](https://github.com/kubernetes/ingress-nginx/blob/a18daabc51977e4e91ed25b9d83e795b8bdd5b5d/docs/annotations.md) defines as:

_Redirect non-TLS requests to TLS even when TLS is not configured._

There's also `ssl-redirect` which redirects if TLS is enabled.

Do we involve the "TLS is [not] enabled" condition somehow? If not, can/should we do it? If we cannot, should we choose a different annotation name?

I don't use TLS myself and don't have any experience with the Nginx annotation, so I cannot fully judge.

@dtomcej any thoughts?

robholland:
Is there anything we can do to help move this along?

timoreimann:
@robholland I'd be happy to hear other folks' opinion on the questions I had raised. Pinging @dtomcej again too.

dtomcej:
Thanks for pinging me on this on @timoreimann. @mattatcha I have a PR I am working on that enables the entirety of the headers suite via annotations (so redirect, htst, custom headers etc).

@mattatcha If it is ok with you, I would like to hold this open, so I can continue to refer to it, and we can continue discussion on that new PR with the extra functionality.

mmatur:
Hey @mattatcha, thanks a lot for your contribution !
We think that it is better to close this issue in favor of @dtomcej PR's
@dtomcej could you link this PR to your PR when it will be ready
Thanks again for you great work @mattatcha :clap:

PR #2167, opened 2017-09-22
Modify ETCD tests to remove docker-etcd fork

nmengin:
### Description

The Repository [docker-tecd](https://github.com/containous/docker-etcd) has been created to override the ETCD starting script in the way to know the container IP address before to really start ETCD.
Træfik tests needed its because they didn't support docker-compose api v2.

Since libkermit and libcompose dependencies have been upgraded, it's possible to define networks in a better way (not quite perfect yet but better).

This PR modify the ETCD test by defining the network IP ranges and force the containers starting order in the way to use the ETCD official image.

### DIFF
--- a/integration/etcd_test.go
+++ b/integration/etcd_test.go
@@ -25,8 +25,8 @@ const (
 	ipWhoami03 string = "172.10.1.5"
 	ipWhoami04 string = "172.10.1.6"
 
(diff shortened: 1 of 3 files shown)


nmengin:
This improvement requires a deep modification of the Tests execution with Docker (sub networks usage with dynamic attaching for the container which launches the tests).

A global PR may be opened for this subject.

Issue #751, opened 2016-10-19
HTTP basic auth per frontend


ArchiFleKs:
Hi, I think it has been asked before but how would you implement http auth basic on a per backend basis ? Currently it is only on the entrypoint.

ViViDboarder:
Do you mean per frontend? 

It was discussed on a previous PR that was merged (#547), but it's probably worth tracking Basic Auth for Fronteneds on a new issue.

ArchiFleKs:
Yes what I meant was on a per host basis for exemple so yes per frontends and not just entrypoints

SantoDE:
I hope to have some time soon again. This is one of the first features I'd contribute then :)

Starefossen:
This would be a very welcome enhancement, I have been poking around in the træfik source code to see if I can make this myself - but I lack the sufficient go knowledge to make this on my own 😢

damonpetta:
+1

j0hnsmith:
There's lots of discussion in #547 but it should be continued here. 

Can the title be edited to say "HTTP basic auth per frontend" as that's what would satisfy most of the requests for this behaviour.

scher200:
Hi there,
I am learning go, so please help out.
I found this peace of go code to do basic auth. Does anyone know where to puzzle it in?
```
func basicAuth(username, password string) string {
  auth := username + ":" + password
   return base64.StdEncoding.EncodeToString([]byte(auth))
}

func redirectPolicyFunc(req *http.Request, via []*http.Request) error{
 req.Header.Add("Authorization","Basic "+basicAuth("username1","password123"))
 return nil
}

def main() {

  client := &http.Client{
    Jar: cookieJar,
    CheckRedirect: redirectPolicyFunc,
}

req, err := http.NewRequest("GET", "http://localhost/", nil)
    req.Header.Add("Authorization","Basic "+basicAuth("username1","password123"))

    resp, err := client.Do(req)
}
```

joshuacox:
I'd like to +1 on this being a necessary feature

scher200:
@joshuacox could you maybe point me in the right direction. Where to a label influenced function like this?

joshuacox:
well where to puzzle that bit of code in I'm not so certain.  But if you merely need an http auth in front of a dev project or something else that does not have it's own auth you can do as I suggested in [referenced issue](https://github.com/rawmind0/rancher-traefik/issues/31) and place an nginx container in front of the webapp you are adding auth to, e.g.:

[...]

PR #2172, opened 2017-09-25
Collect anonymous usage

ldez:
## Collected Data

**This feature is disabled by default.**

### Why ?

In order to help us learn more about how Træfik is being used and improve it, we collect anonymous usage statistics from running instances.
Those data help us prioritize our developments and focus on what's more important (for example, which configuration backend is used and which is not used).

### What ?

Once a day, we collect:
 - the Træfik version
 - the UUID of the instance
 - an **obfuscated and anonymous version** of the static configuration (token, user name, password, URL, IP, domain, email, etc, are removed)

**We do not collect the dynamic configuration (frontends & backends).**

**We do not collect data behind the scenes to run advertising programs or to sell such data to third-party.**

#### Here is an example

- Source configuration:

```toml
[entryPoints]
    [entryPoints.http]
       address = ":80"

[web]
  address = ":8080"

[Docker]
  endpoint = "tcp://10.10.10.10:2375"
  domain = "foo.bir"
  exposedByDefault = true
  swarmMode = true

  [Docker.TLS]
    CA = "dockerCA"
    Cert = "dockerCert"
    Key = "dockerKey"
    InsecureSkipVerify = true

[ECS]
  Domain = "foo.bar"
  ExposedByDefault = true
  Clusters = ["foo-bar"]
  Region = "us-west-2"
  AccessKeyID = "AccessKeyID"
  SecretAccessKey = "SecretAccessKey"
```

- Obfuscated and anonymous configuration:

```toml
[entryPoints]
    [entryPoints.http]
       address = ":80"

[web]
  address = ":8080"

[Docker]
  Endpoint = "xxxx"
  Domain = "xxxx"
  ExposedByDefault = true
  SwarmMode = true

  [Docker.TLS]
    CA = "xxxx"
    Cert = "xxxx"
    Key = "xxxx"
    InsecureSkipVerify = false

[ECS]
  Domain = "xxxx"
  ExposedByDefault = true
  Clusters = []
  Region = "us-west-2"
  AccessKeyID = "xxxx"
  SecretAccessKey = "xxxx"
```

### Show me the code !

If you want to dig into more details, here is the source code of the collecting system: [collector.go](https://github.com/containous/traefik/blob/master/collector/collector.go)

By default we obfuscate all configuration fields, except fields tagged with `export=true`.

### How to enable this ?

You can enable the collecting system by:

- adding this line in the configuration TOML file:

```toml
# Send anonymous usage data
#
# Optional
# Default: false
#
sendAnonymousData = true
```

- adding this flag in the CLI:

```bash
./traefik --sendAnonymousData=true
```


cmd/traefik/traefik.go
```diff
@@ -366,3 +362,33 @@ func CreateKvSource(traefikConfiguration *TraefikConfiguration) (*staert.KvSourc
...
 	return kv, err
 }
+
+func collect(globalConfiguration *configuration.GlobalConfiguration) {
+	log.Info("Many thanks to contribute to Træfik improvement by allowing us to receive anonymous information from your configuration.")
+	log.Info("More information about collected data: https://docs.traefik.io/configuration/commons/#collected-data")
+	ticker := time.NewTicker(24 * time.Hour)
+	safe.Go(func() {
+		err := collector.Collect(globalConfiguration)
+		log.Debug(err)
```
emilevauge: `if err != nil` ?

cmd/traefik/traefik.go
```diff
@@ -366,3 +362,33 @@ func CreateKvSource(traefikConfiguration *TraefikConfiguration) (*staert.KvSourc
...
+		for {
+			select {
+			case <-ticker.C:
+				err := collector.Collect(globalConfiguration)
+				log.Debug(err)
```
emilevauge: if err != nil ?

collector/collector.go
```diff
@@ -0,0 +1,50 @@
...
+	Configuration string
+	UUID          string
+}
+
+// Collect anonymous and obfuscated data.
+func Collect(globalConfiguration *configuration.GlobalConfiguration) error {
+
+	obfuscatedConfig, err := Obfuscate(globalConfiguration, false)
+
+	log.Debug(obfuscatedConfig)
```
emilevauge: `Anonymous collected data+=`

docs/configuration/commons.md
```diff
@@ -437,3 +437,106 @@ Example:
...
+  ExposedByDefault = true
+  Clusters = []
+  Region = "us-west-2"
+  AccessKeyID = "xxxx"
+  SecretAccessKey = "xxxx"
+```
+
+### Show me the code !
+
+If you want to dig into more details, here is the source code of the collecting system: [collector.go](https://github.com/containous/traefik/blob/master/collector/collector.go)
```
timoreimann: Link yields 404.

emilevauge: @timoreimann yeah, this PR is still not merged ;)

timoreimann: Ah of course, my bad. :)

george-angel:
This should be rolled out with care, as some people might be using workarounds in their config for missing features, a la https://github.com/containous/traefik/issues/751#issuecomment-292169545

emilevauge:
@george-angel we don't export the dynamic configuration of Traefik (which contains frontends & backends), so we don't see any risk of data leak.
But you are right, this feature should be rolled out with care, that's why we carefully tested it, and that's why we ask people to review it here :)
If you have any doubt on this, you are welcome to drop a comment here.

george-angel:
Thanks @emilevauge 

Could you clarify regarding dynamic configuration exclusion. Does that include *all* frontends, or the ones creating via docker/kubernetes?

My example its a frontend matching a Host with a secret value defined in `traefik.toml`.

emilevauge:
@george-angel frontends & backends defined in your toml will also be excluded. Those are part of the dynamic configuration. You could have defined those in a separate file, but you can also choose to merge your static & dynamic configuration in one toml file (like you did).

In one word, frontends & backends are *not* collected (obviously :) ).

emilevauge:
@alkar, @egtann, @kachkaev: I'm curious to understand why you reacted with :-1: / :confused: to this PR.
I'm sure your understand why we need to get some usage metrics to better understand our community and improve Traefik accordingly. Could you elaborate?

kachkaev:
Thanks for offering to leave feedback @emilevauge! I'm personally a bit confused about the fact that the option is `true` by default, however I think I understand why it it so.

Given that this is a last-minute PR for 1.4, could the option be made `false` by default to avoid bugs in critical environments in its early days? Then once this functionality matures, you can maybe release 2.0 with `sendAnonymousData = true` by default and clearly explain what this means in the changelog. Quickly upgrading from 1.3 to 1.4 and later realising that you've been tracked for quite some time may frustrate a lot of users. May be risky for traefik's reputation IMO.

george-angel:
@emilevauge great, thank you.

emilevauge:
@kachkaev Thanks for your feedback !

We need more usage stats (for example, which version is still deployed, etc), to know better how Traefik is deployed. We just need anonymous data, we really really don't care about something else ;) 

**But we really don't want to upset our community.**

So let's disable this option to by default for now. But we need your help to enable this by default in the future, without loosing your trust :)

alkar:
@emilevauge although I understand that the data you're going to collect can be useful for the team, there's two points that actually make me sceptical:

- this feature is enabled by default and users could potentially be unaware when upgrading a minor version, like @kachkaev explained

- you'll be extracting information from the config which is always dangerous in a sense

To be perfectly honest, I don't see how such a feature makes sense to be a core part of a *reverse proxy* - to my knowledge it's a first.

ldez:
@alkar **The feature is disabled by default.**

alkar:
@ldez I am aware, I read Emile's comment above.

emilevauge:
Following all the feedback from the community, we think it's better to close the PR and not ship this feature in 1.4. We don't want to rush, we shouldn't have tried to add this to 1.4 (I thought it would be "ok" but it's not ^^).

I will instead open a proposal to explain why we (developers) need more stats and which data we need precisely. We really need your help to build something that match the developers and the users needs.

2015-10-21-1445410698-8403632-futurecontinued

emilevauge:
A proposal is alive here #2172 :)

Issue #2156, opened 2017-09-21
Capture all Traefik Log output in file


marco-jantke:
### Do you want to request a *feature* or report a *bug*?

**bug**

### What did you do?

Start a Traefik instance with a configuration that writes the traefik logs to a file. E.g. `traefik --traefiklog.filepath=./output.log`

### What did you expect to see?

No logs are written to stdout and all logs are written to the specified file.

### What did you see instead?

The first and some of the last messages are written to stdout.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      dev
Codename:     cheddar
Go version:   go1.9
Built:        I don't remember exactly
OS/Arch:      darwin/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
# traefik.toml
logLevel = "DEBUG"
defaultEntryPoints = ["http"]

[traefikLog]
filePath="./traefik.log"

[entryPoints.http]
  address = ":8080"

[file]
watch = true

# rules
[backends]
  [backends.backend1]
    [backends.backend1.servers.server1]
    url = "http://localhost:8081"

[frontends]
  [frontends.frontend1]
  backend = "backend1"
    [frontends.frontend1.routes.test_1]
    rule = "Path:/ok"
```

### If applicable, please paste the log output in debug mode (`--debug` switch)

```bash
➜  traefiklog traefik
INFO[0000] Using TOML configuration file /Users/mjantke/work/go/src/github.com/marco-jantke/go-playground/traefik-configs/traefiklog/traefik.toml 
^Ctime="2017-09-21T13:44:44+02:00" level=info msg="I have to go... <nil>" 
time="2017-09-21T13:44:44+02:00" level=info msg="Stopping server" 
time="2017-09-21T13:44:44+02:00" level=debug msg="Waiting 10s seconds before killing connections on entrypoint http..." 
➜  traefiklog cat traefik.log 
time="2017-09-21T13:44:42+02:00" level=info msg="Traefik version dev built on I don't remember exactly" 
[...]

PR #2176, opened 2017-09-26
Fix traefik logs to behave like configured

marco-jantke:
This PR accomplishes two things: 
1. it makes sure that all traefik logs are written in the specified format. Before the first log messages were written in text, even when you specified JSON as the format. This is because some log output was produced before configuring the logs.
2. it makes sure that all traefik logs are written to the traefik logs file. This is implemented by using `logrus.Exit()` and by closing the file in a `logrus.RegisterExitHandler`.

I also changed that errors from `http.Serve` and `http.ServeTLS` are not logged anymore, in the case they are regular `http.ErrServerClosed` errors, which is expected and the normal case. Those methods will never return a non-nil error.

This fixes https://github.com/containous/traefik/issues/2156.

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -45,19 +45,17 @@ Complete documentation is available at https://traefik.io`,
 		Config:                traefikConfiguration,
 		DefaultPointersConfig: traefikPointersConfiguration,
 		Run: func() error {
-			globalConfiguration := traefikConfiguration.GlobalConfiguration
-			if globalConfiguration.File != nil && len(globalConfiguration.File.Filename) == 0 {
-				// no filename, setting to global config file
-				if len(traefikConfiguration.ConfigFile) != 0 {
-					globalConfiguration.File.Filename = traefikConfiguration.ConfigFile
-				} else {
-					log.Errorln("Error using file configuration backend, no filename defined")
-				}
+			globalConfiguration := &traefikConfiguration.GlobalConfiguration
+			fileProviderEnabled := globalConfiguration.File != nil
+			traefikConfigFileExists := len(traefikConfiguration.ConfigFile) != 0
+
+			// Try to fallback to traefik config file in case the file provider is enabled
+			// but it has no file name configured.
+			if fileProviderEnabled && len(globalConfiguration.File.Filename) == 0 && traefikConfigFileExists {
+				globalConfiguration.File.Filename = traefikConfiguration.ConfigFile
 			}
-			if len(traefikConfiguration.ConfigFile) != 0 {
-				log.Infof("Using TOML configuration file %s", traefikConfiguration.ConfigFile)
-			}
-			run(&globalConfiguration)
+
+			run(globalConfiguration, traefikConfiguration.ConfigFile)
 			return nil
 		},
 	}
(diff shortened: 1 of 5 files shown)


timoreimann:
Do we know why simply deferring isn't enough? Is this possibly related to a missed flush or a missing close? (I don't know if flushing stdout is even possible, just throwing around wild guesses here.)

marco-jantke:
Yes, I checked that deferring is not enough. This is basically the first example I have written in the description above (`# sample of log output with closing log file`). 

Frankly speaking, I also don't understand yet why this does not work. AFAICS the `run` function is executed in the main go routine and thus when it exits, everything should exit. So my understanding of go. The output that is printed to stdout, given we close the log file in a `defer` of `run` is from other goroutines like `listenSignals`. I am currently digging deeper to find the reason for this.

marco-jantke:
@timoreimann I was able to build a solution that captures all output by using `logrus.RegisterExitHandler`. This way we can close the files properly and still catch all the output.

After having implemented this, there was surprisingly still some output written to the stdout. Namely this: `{"level":"error","msg":"Error creating server: http: Server closed","time":"2017-09-26T09:57:15+02:00"}`. I had a look into the code and this is actually confusing in my opinion. This is the error returned from the `http.Serve` or `http.ServeTLS` which always return a non-nil error. I changed the logic in 20e4471 to output only real errors in order to avoid confusion and this way, we now have no more output on the stdout 🎉 

Updating PR description right now.

ldez:
Please undo the refactoring. (We can speak about this on Slack)

marco-jantke:
@ldez and @Juliens I changed and rebased the PR now according to our discussion. This was basically to keep the `run` method agnostic of the `traefikConfiguration`. PR is ready for review.

cmd/traefik/traefik.go
```diff
@@ -45,19 +45,17 @@ Complete documentation is available at https://traefik.io`,
...
-			if len(traefikConfiguration.ConfigFile) != 0 {
-				log.Infof("Using TOML configuration file %s", traefikConfiguration.ConfigFile)
+			globalConfiguration := &traefikConfiguration.GlobalConfiguration
+			fileProviderEnabled := globalConfiguration.File != nil
+			traefikConfigFileExists := len(traefikConfiguration.ConfigFile) != 0
+
+			// Try to fallback to traefik config file in case the file provider is enabled
+			// but it has no file name configured.
+			if fileProviderEnabled && len(globalConfiguration.File.Filename) == 0 && traefikConfigFileExists {
+				globalConfiguration.File.Filename = traefikConfiguration.ConfigFile
```
Juliens: Can you move this code in the run func, and use the "new" config file param

cmd/traefik/traefik.go
```diff
@@ -317,6 +276,64 @@ func run(globalConfiguration *configuration.GlobalConfiguration) {
...
+	logrus.Exit(0)
+}
+
+func configureLogging(globalConfiguration *configuration.GlobalConfiguration) {
+	// configure default log flags
+	fmtlog.SetFlags(fmtlog.Lshortfile | fmtlog.LstdFlags)
+
+	if globalConfiguration.Debug {
+		globalConfiguration.LogLevel = "DEBUG"
+	}
```
Juliens: this code should be in setEffectiveConfiguration ? WDYT

cmd/traefik/traefik.go
```diff
@@ -317,6 +276,64 @@ func run(globalConfiguration *configuration.GlobalConfiguration) {
...
+	if err != nil {
+		log.Error("Error getting level", err)
+	}
+	log.SetLevel(level)
+
+	// configure log output file
+	logFile := globalConfiguration.TraefikLogsFile
+	if len(logFile) > 0 {
+		log.Warn("top-level traefiklogsfile has been deprecated -- please use traefiklog.filepath")
+	}
```
Juliens: this code should be in setEffectiveConfiguration and change the `globalConfiguration.TraefikLog` value ? WDYT

cmd/traefik/traefik.go
marco-jantke: I am not sure about that. The docs of `SetEffectiveConfiguration` states:

> SetEffectiveConfiguration adds missing configuration parameters derived from existing ones. It also takes care of maintaining backwards compatibility.

In my perception, this is actually not the case.  `LogLevel` is not a missing configuration parameter. It's meant to be this way.

Additionally `SetEffectiveConfiguration` is called _after_ `configureLogging`, and that is for the good reason that the logging configuration should happen as first thing in order to achieve what this PR tries to achieve. Log all messages in the desired format and write them to a file if it is configured.

cmd/traefik/traefik.go
marco-jantke: I don't think so, for the same reasons stated it in https://github.com/containous/traefik/pull/2176#discussion_r142194033.

cmd/traefik/traefik.go
marco-jantke: I went one step further and pushed it down to `SetEffectiveConfiguration` as it fits there best IMO.

marco-jantke:
@Juliens I commented/addressed all your comments. PTAL again :-)

ldez:
```
configuration/configuration_test.go:244:32: not enough arguments in call to gc.SetEffectiveConfiguration
	have ()
	want (string)
```

marco-jantke:
Ups. Will fix Wednesday morning probably. Holiday :D

marco-jantke:
@ldez @Juliens PTAL again.

cmd/traefik/traefik.go
```diff
@@ -223,8 +211,23 @@ Complete documentation is available at https://traefik.io`,
 	os.Exit(0)
 }
 
-func run(globalConfiguration *configuration.GlobalConfiguration) {
-	fmtlog.SetFlags(fmtlog.Lshortfile | fmtlog.LstdFlags)
+func run(traefikConfiguration *TraefikConfiguration) {
```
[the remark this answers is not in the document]
marco-jantke: Is this something I should address in a follow-up PR?

Also, can you explain a bit the benefit you see in this? We were discussing with @Juliens that you want to keep `run` function to not know anything about the `traefikConfiguration`, because of a future feature you were working on. Is this still aligned with this effort?

PR #2209, opened 2017-10-03
Add time between last request byte and last response byte to access log

### LINKED ISSUE #2043: Buffering client requests.
Feature

### What did you do?
We have a Traefik instance set up as ingress controller in our Kubernetes cluster. Before we used Nginx as a proxy for our services.
It was configured to buffer client' requests before sending them to upstream/destination.
Based on access metrics from Nginx and self-reported ones from microservices we set up alerts to trigger when response time passes the defined threshold. It was working fine.
However Traefik doesn't buffer clients' requests and pipes them straight to client.
In case of clients with slow user connection and e.g. file upload request between traefik and microservice/destination can take very long (tens of seconds).
It is generally fine, as we don't have any control over users connection. Problem is with metrics as we are no longer able to tell whether microservice behaves correctly based on request time.
Our Apdex alerts don't work. 

Another drawback of not buffering requests is using more resources on microservices behind Traefik which work in thread-per-request model.

We can base view on microservice healthiness on metrics we get from Traefik / self-reported by microservice.

### What did you see instead?

Metrics are highly affected by clients connection speed.

### Proposed solutions
* allow requests buffering (AFAIK it is supported by https://github.com/vulcand/oxy)
* provide more information about request processing time in logs. (e.g. last byte sent to destination to last byte received from destination)

harnash:
### Description
Fixes #2043

Adds additional metric which measures time between end of request processing (whole request was sent to the Traefik) and the response received from the backend. It is useful to measure time backend took to deliver response and does not include time needed for request to finish (slow clients).


middlewares/accesslog/logger.go
```diff
@@ -214,6 +214,13 @@ func (l *LogHandler) logTheRoundTrip(logDataTable *LogData, crr *captureRequestR
 	} else {
 		core[Overhead] = total
 	}
+	if crr != nil && crw != nil {
+		core[ResponseDuration] = crw.processingStart.Sub(crr.processingEnd)
```
marco-jantke: Won't this lead to a negative number in `ResponseDuration`? Start should be smaller than end.

marco-jantke:
@harnash thank you for this contribution! I have a couple of questions though:
1. are you sure that this works properly? AFAIK the request flow in Traefik it already allows forwarding the request to the backend even though it did not read the whole request body yet. Have you tried this in a real world scenario already? I guess also @emilevauge who will provide the design review (as marked by the PR labels) can confirm/prove my statement as wrong.
1. can you elaborate a bit more on which problem in the real world this would help to diagnose?
1. can you rephrase your PR title? We have actual metrics collection/exporting in Traefik and your PR is about access logs.

middlewares/accesslog/logger.go
harnash: It is a bit misleading (perhaps naming should be better in this case). My understanding of the flow is as following:
`crw` is `ResponseWritter` and `processingStart` indicates when Traefik started generating response. `crr` is `RequestReader` where `processingEnd` indicates when Traefik finished processing request body. The only way this will be negative when Generating response will be faster than finishing of request processing... timeouts perhaps?.

harnash:
@marco-jantke 
ad 1. We are planning to run this on our infrastructure after we get at least initial approval. I did test it only locally so far but I can deploy it on our dev/staging env to see how it behaves under moderate load (this is a bit blocked since we are not 100% ready to migrate to 1.4+ on prod yet).

ad 2. You could check description in the #2043 issue. We have alerting setup based on response time from our services (gathered from Traefik access logs). Currently they trigger "false positives" when service client has poor connection and/or is submitting large requests. We found that it would be best to have response times measured after request is transmitted in full (so we do not include time for client to send its payload). Another solution for that is to allow request buffering (vulcand/oxy does support that and I have PR almost ready with that).

ad 3. sure

marco-jantke:
Thank you for the explanations! Now it's @emilevauge do reason about the design :-)

emilevauge:
Hey @harnash, thanks a lot for your contribution :clap: 
We think that Open Tracing support added in this PR https://github.com/containous/traefik/pull/2237 also solves your issue, and is a bit more generic. We prefer to close this PR and focus on OpenTracing/Jaeger/Zipkin :)
Thanks again for you great work @harnash !

harnash:
@emilevauge nice! I'll take a look as we are planning on integrating OpenTracing in our stack too.

Issue #2203, opened 2017-10-02
Flaky Integration Test: AccessLog TestAccessLogRotation


marco-jantke:
### Do you want to request a *feature* or report a *bug*?

*bug*

### Summary

- *Which test is flaky?* -> `LogRotationSuite.TestAccessLogRotation`
- *Cause* -> Access Log file is not yet created after rotation.
- *Failed build* -> https://semaphoreci.com/containous/traefik/branches/master/builds/188

### Log excerpt

```
----------------------------------------------------------------------
FAIL: log_rotation_test.go:21: LogRotationSuite.TestAccessLogRotation

127.0.0.1 - - [26/Sep/2017:08:38:14 +0000] "GET /test1 HTTP/1.1" 200 22 - "Go-http-client/1.1" 1 "frontend1" "http://127.0.0.1:8081" 0ms
127.0.0.1 - - [26/Sep/2017:08:38:14 +0000] "GET /test1 HTTP/1.1" 200 22 - "Go-http-client/1.1" 2 "frontend1" "http://127.0.0.1:8081" 0ms
127.0.0.1 - - [26/Sep/2017:08:38:14 +0000] "GET /test1 HTTP/1.1" 200 22 - "Go-http-client/1.1" 3 "frontend1" "http://127.0.0.1:8081" 0ms
log_rotation_test.go:64:
    // Verify access.log output as expected
    lineCount = verifyLogLines(c, traefikTestAccessLogFile, lineCount, true)
log_rotation_test.go:128:
    c.Assert(err, checker.IsNil)
... value *os.PathError = &os.PathError{Op:"open", Path:"access.log", Err:0x2} ("open access.log: no such file or directory")
```

marco-jantke:
Note: for this one I already have a fix. Will open PR next days.

mmatur:
I close this issue because it is fixed by #2213

PR #2213, opened 2017-10-04
fix flakiness in log rotation test

marco-jantke:
This PR fixes https://github.com/containous/traefik/issues/2203 by waiting up to 3 seconds until the new log file is created after the signal was sent to Traefik to rotate the log files. Also, it improves logging of the log file output in case something is not correct to ease the detection of potential other problems that were hidden before.

### DIFF
--- a/integration/log_rotation_test.go
+++ b/integration/log_rotation_test.go
@@ -122,9 +122,8 @@ func (s *LogRotationSuite) TestTraefikLogRotation(c *check.C) {
 
 func logAccessLogFile(c *check.C, fileName string) {
 	output, err := ioutil.ReadFile(fileName)
-	c.Assert(err, checker.IsNil, check.Commentf("error opening access log file %s: %s", fileName, err))
-	c.Logf("Contents of file %s", fileName)
-	c.Log(string(output))
+	c.Assert(err, checker.IsNil)
+	c.Logf("Contents of file %s\n%s", fileName, string(output))
 }
 
 func verifyEmptyErrorLog(c *check.C, name string) {
--- a/integration/log_rotation_test.go
+++ b/integration/log_rotation_test.go
@@ -57,10 +57,19 @@ func (s *LogRotationSuite) TestAccessLogRotation(c *check.C) {
 	c.Assert(err, checker.IsNil)
 
 	// Verify access.log.rotated output as expected
+	logAccessLogFile(c, traefikTestAccessLogFile+".rotated")
 	lineCount := verifyLogLines(c, traefikTestAccessLogFile+".rotated", 0, true)
 	c.Assert(lineCount, checker.GreaterOrEqualThan, 1)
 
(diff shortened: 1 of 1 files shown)


integration/log_rotation_test.go
```diff
@@ -111,6 +120,13 @@ func (s *LogRotationSuite) TestTraefikLogRotation(c *check.C) {
 	c.Assert(lineCount, checker.GreaterOrEqualThan, 7)
 }
 
+func logAccessLogFile(c *check.C, fileName string) {
+	output, err := ioutil.ReadFile(fileName)
+	c.Assert(err, checker.IsNil, check.Commentf("error opening access log file %s: %s", fileName, err))
```
timoreimann: AFAIR the error from `ioutil.ReadFile` will already contain the file name and detailed error message.

integration/log_rotation_test.go
```diff
@@ -111,6 +120,13 @@ func (s *LogRotationSuite) TestTraefikLogRotation(c *check.C) {
...
+	c.Logf("Contents of file %s", fileName)
+	c.Log(string(output))
```
timoreimann: Could also be part of the previous statement, separating the message "header" and the content by a newline (`\n`).

integration/log_rotation_test.go
```diff
@@ -111,6 +120,13 @@ func (s *LogRotationSuite) TestTraefikLogRotation(c *check.C) {
 	c.Assert(lineCount, checker.GreaterOrEqualThan, 7)
 }
 
+func logAccessLogFile(c *check.C, fileName string) {
+	output, err := ioutil.ReadFile(fileName)
```
timoreimann: Maybe a case of [`t.Helper()`](https://golang.org/pkg/testing/#T.Helper)?

marco-jantke: In the case of the integration tests we can not use it, as we don't have access to `testing.T`.

marco-jantke:
@timoreimann thanks for the review, I incorporated the feedback. I am not sure though why Semaphore won't build this branch..

integration/log_rotation_test.go
marco-jantke: Good point. I removed the `Commentf`.

integration/log_rotation_test.go
timoreimann: Ah -- sad and true.

integration/log_rotation_test.go
marco-jantke: Simplified it into one line 👍

ldez:
I think 3s is too large: FYI we use a multiplier in the CI

https://github.com/containous/traefik/blob/cf508b6d481599c4ffd83c52e48beb9cf561314a/integration/try/try.go#L149-L156

ldez:
I close and reopen due to Semaphore

Issue #1889, opened 2017-07-24
Add an integration test for the Kubernetes Ingress controller client logic.


timoreimann:
### Do you want to request a *feature* or report a *bug*?
feature.

### What did you do?
Make the already very awesome Traefik Ingress controller client implementation even more awesome.

### What did you expect to see?
My change being covered by an integration test proving the next level in awesomeness achieved.

### What did you see instead?
Being on my own with just unit tests, but nothing that tests the actual client.

### Details
While we have fairly decent unit test coverage to validate our Ingress controller business logic, we are lacking integration tests for the client part (i.e., everything related to creating a client, but most importantly the functionality setting up the Informer). Unit-testing this part doesn't really make sense IMHO since the logic is highly dependent on interactions with an actual k8s API server.

I've done a bit of research how to best set up a Kubernetes system/cluster suitable for integration testing purposes (but not much more). There seem to be two viable options:

1. Use the [test apiserver](https://github.com/kubernetes/kubernetes/tree/6fdf0e4157c76eeaf7619e8411ee7231305f7d87/cmd/kube-apiserver/app/testing). The nice thing is that it runs all in-process and comes with its own etcd testing copy. Note that 1.7 doesn't come with the test apiserver yet, but something similar exists in [test/integration/framework](https://github.com/kubernetes/kubernetes/tree/release-1.7/test/integration/framework). Back-porting the test apiserver to 1.7 should also be fairly trivial.
1. Use [Mirantis's kubeadm-dind-cluster](https://github.com/Mirantis/kubeadm-dind-cluster), which brings a full cluster that only needs Docker.

AFAICS, the test apiserver would make for a more light-weight approach but might require a bit of preparation (e.g., creating a Docker image for the k8s release we're targeting). The dind approach is presumably simpler to integrate into our pre-existing docker-compose-based infrastructure but gives us things we don't really need; for instance, a multi-node setup is not strictly required to test the Informer.

timoreimann:
Kudos to @luxas and @sttts for providing consultancy on the different integration testing options. 🎉

luxas:
What about doing both the e2e (full cluster with `n` dind nodes and everything set up) that deploys a CI image with the recommended k8s YAML file + some test deployments etc. _and_ the more lightweight integration test that tests parts of the code (do we use this informer correctly here etc.)?

timoreimann:
[...]

PR #2222, opened 2017-10-07
Allow custom value for kubernetes.io/ingress.class annotation

### LINKED ISSUE #2221: Support watching for custom values of kubernetes.io/ingress.class

Feature

### What did you do?

I have multiple traefik servers installed on a kubernetes cluster using Ingress objects as backend. Since this is running on Google Cloud, there is also a default Kubernetes Ingress provider. Currently I can specify that an Ingress object should only be processed by traefik with the `kubernetes.io/ingress.class` annotation to `traefik`. This is good - however, not good enough. I'd like to be able to customize the `ingress.class` of each traefik server to be something different, so they can be more specifically targeted.

1.3.8
 
```toml
     # traefik.toml
      logLevel = "INFO"
      defaultEntryPoints = ["http"]
      [entryPoints]
        [entryPoints.http]
        address = ":80"
        compress = true
      [kubernetes]
```

yuvipanda:
### Description

Allow configuring the values of the `kubernetes.io/ingress.class` annotation
on ingress objects that traefik will process. This is useful when you have
multiple traefik servers in one kubernetes cluster, and you want tight control
over which Ingresses are processed by which servers.

Fixes #2221

### DIFF
--- a/cmd/traefik/configuration.go
+++ b/cmd/traefik/configuration.go
@@ -132,7 +132,7 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 	defaultKubernetes.Watch = true
 	defaultKubernetes.Endpoint = ""
 	defaultKubernetes.LabelSelector = ""
-	defaultKubernetes.IngressClass = "traefik"
+	defaultKubernetes.IngressClass = ""
 	defaultKubernetes.Constraints = types.Constraints{}
 
 	// default Mesos
--- a/cmd/traefik/configuration.go
+++ b/cmd/traefik/configuration.go
@@ -132,6 +132,7 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 	defaultKubernetes.Watch = true
 	defaultKubernetes.Endpoint = ""
 	defaultKubernetes.LabelSelector = ""
+	defaultKubernetes.IngressClass = "traefik"
 	defaultKubernetes.Constraints = types.Constraints{}
 
 	// default Mesos
--- a/cmd/traefik/configuration.go
+++ b/cmd/traefik/configuration.go
@@ -130,9 +130,6 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 	//default Kubernetes
 	var defaultKubernetes kubernetes.Provider
 	defaultKubernetes.Watch = true
-	defaultKubernetes.Endpoint = ""
-	defaultKubernetes.LabelSelector = ""
-	defaultKubernetes.IngressClass = ""
 	defaultKubernetes.Constraints = types.Constraints{}
 
 	// default Mesos
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -50,6 +50,15 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 #
 # labelselector = "A and not B"
 
+# Value of `kubernetes.io/ingress.class` annotation that identify Ingress objects to be processed.
+# Note that Ingress objects with an empty or nonexistent `kubernetes.io/ingress.class` value
+# will also be processed.
+#
+# Optional
+# Default: traefik (process all Ingresses)
+#
+# ingressClass = "custom-proxy"
+
 # Disable PassHost Headers.
 #
 # Optional
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -51,14 +51,15 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # labelselector = "A and not B"
 
 # Value of `kubernetes.io/ingress.class` annotation that identifies Ingress objects to be processed.
-# Note that Ingress objects with an empty or nonexistent `kubernetes.io/ingress.class` value
-# will also be processed.
-# Note that if `kubernetes.io/ingress.class` annotation  is not empty, only annotations with the prefix `traefik` will be processed.
+# If the parameter is non-empty, only Ingresses containing an annotation with the same value are processed.
+# Otherwise, Ingresses missing the annotation, having an empty value, or the value `traefik` are processed.
+#
+# Note : `ingressClass` option must begin with the "traefik" prefix.
 #
 # Optional
-# Default: empty (process all Ingresses)
+# Default: empty
 #
-# ingressClass = "traefik-custom-proxy"
+# ingressClass = "traefik-internal"
 
 # Disable PassHost Headers.
 #
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -50,7 +50,7 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 #
 # labelselector = "A and not B"
 
-# Value of `kubernetes.io/ingress.class` annotation that identify Ingress objects to be processed.
+# Value of `kubernetes.io/ingress.class` annotation that identifies Ingress objects to be processed.
 # Note that Ingress objects with an empty or nonexistent `kubernetes.io/ingress.class` value
 # will also be processed.
 #
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -53,7 +53,7 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # Value of `kubernetes.io/ingress.class` annotation that identifies Ingress objects to be processed.
 # Note that Ingress objects with an empty or nonexistent `kubernetes.io/ingress.class` value
 # will also be processed.
-# Note that only annotations with the prefix `traefik` will be processed.
+# Note that if `kubernetes.io/ingress.class` annotation  is not empty, only annotations with the prefix `traefik` will be processed.
 #
 # Optional
 # Default: empty (process all Ingresses)
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -53,11 +53,12 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # Value of `kubernetes.io/ingress.class` annotation that identifies Ingress objects to be processed.
 # Note that Ingress objects with an empty or nonexistent `kubernetes.io/ingress.class` value
 # will also be processed.
+# Note that only annotations with the prefix `traefik` will be processed.
 #
 # Optional
-# Default: traefik (process all Ingresses)
+# Default: empty (process all Ingresses)
 #
-# ingressClass = "custom-proxy"
+# ingressClass = "traefik-custom-proxy"
 
 # Disable PassHost Headers.
 #
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -46,6 +46,7 @@ type Provider struct {
 	EnablePassTLSCert      bool       `description:"Kubernetes enable Pass TLS Client Certs" export:"true"`
 	Namespaces             Namespaces `description:"Kubernetes namespaces" export:"true"`
 	LabelSelector          string     `description:"Kubernetes api label selector to use" export:"true"`
+	IngressClass           string     `description:"Value of kubernetes.io/ingress.class annotation to watch for" export:"true"`
 	lastConfiguration      safe.Safe
 }
 
@@ -135,6 +136,15 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 	return nil
 }
 
+func (p *Provider) shouldProcessIngress(ingressClass string) bool {
+	switch ingressClass {
+	case "", p.IngressClass:
+		return true
+	default:
+		return false
+	}
+}
+
 func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error) {
 	ingresses := k8sClient.GetIngresses()
 
@@ -147,7 +157,7 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 		annotationIngressClass := getAnnotationName(i.Annotations, annotationKubernetesIngressClass)
 		ingressClass := i.Annotations[annotationIngressClass]
 
-		if !shouldProcessIngress(ingressClass) {
+		if !p.shouldProcessIngress(ingressClass) {
 			continue
 		}
 
@@ -449,10 +459,6 @@ func equalPorts(servicePort v1.ServicePort, ingressPort intstr.IntOrString) bool
 		return true
 	}
 	return false
-}
-
-func shouldProcessIngress(ingressClass string) bool {
-	return ingressClass == "" || ingressClass == "traefik"
 }
 
 func getFrontendRedirect(i *v1beta1.Ingress) *types.Redirect {
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -34,7 +34,10 @@ const (
 	ruleTypeReplacePath = "ReplacePath"
 )
 
-const traefikDefaultAnnotationValue = "traefik"
+const (
+	traefikDefaultRealm        = "traefik"
+	traefikDefaultIngressClass = "traefik"
+)
 
 // Provider holds configurations of the provider.
 type Provider struct {
@@ -79,8 +82,8 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 
 	// We require that IngressClasses start with `traefik` to reduce chances of
 	// conflict with other Ingress Providers
-	if len(p.IngressClass) > 0 && !strings.HasPrefix(p.IngressClass, traefikDefaultAnnotationValue) {
-		return fmt.Errorf("kubernetes.io/ingress.class value has to be empty or start with the prefix %s, instead found %q", traefikDefaultAnnotationValue, p.IngressClass)
+	if len(p.IngressClass) > 0 && !strings.HasPrefix(p.IngressClass, traefikDefaultIngressClass) {
+		return fmt.Errorf("value for IngressClass has to be empty or start with the prefix %q, instead found %q", traefikDefaultIngressClass, p.IngressClass)
 	}
 
 	k8sClient, err := p.newK8sClient()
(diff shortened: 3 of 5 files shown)


ldez:
@containous/kubernetes WDYT? (Design review)

timoreimann:
Technically, I'd tentatively approve the design.

Whether we should accept the PR though is still subject to an agreement that the existing ways to distinguish separate Traefik deployments is really insufficient. I have [commented the related feature request issue](https://github.com/containous/traefik/issues/2221#issuecomment-334914404) for that purpose. I suggest we wait for a resolution prior to moving forward with an actual review.

timoreimann:
As [explained on the related issue](https://github.com/containous/traefik/issues/2221#issuecomment-334982248), the desirable feature is to have a _subset_ of Traefik controllers ignore certain Ingresses unless intent to process is signaled explicitly (i.e., opt-in instead of opt-out). Traefik label selectors don't fulfill this need since the default behavior of a missing selector is to grab Ingress objects.

Acknowledging the feature request, I'm uncertain if we should reuse `kubernetes.io/ingress.class` for the desired purpose. As far as I know, other Ingress implementations use it to rule out a particular _class_ of controllers only (as the name implies), not for fine-grained discriminations within a single controller. My worry is that allowing arbitrary values may collide with potential other Ingress controllers in the future, possibly becoming a source of confusion.

Maybe a better approach could be to introduce a new annotation -- I'm thinking of something including the term `sub-class`. One thing I'm not sure about though is if it would live under the `ingress.kubernetes.io/` or a custom `traefik.*/` prefix.

@aledbf any thoughts on how the feature should be best implemented from an annotation point of view? Could you maybe share some information on how new `ingress.kubernetes.io/` annotations get added? As far as I know, there's no standardization process available yet. Is it as simple as one Ingress controller implementation taking a lead and others following (with the lead usually being Nginx)?

Pulling in @dtomcej and @errm for additional feedback.

yuvipanda:
I'll note that the nginx controller *does* allow you to change this (in the `--ingress-class` commandline parameter), which is where I got this idea from.

timoreimann:
@yuvipanda thanks, I missed that! (Was only looking at the annotation documentation.)

Let's see what the folks I pinged for feedback think.

aledbf:
@timoreimann the annotation `kubernetes.io/ingress.class` is the "proper" way to filter which ingress rules should be processed or not.

> any thoughts on how the feature should be best implemented in terms of annotations?

The use of annotations is being reviewed and possible [Custom Resources](https://kubernetes.io/docs/concepts/api-extension/custom-resources/) is going to be the way to handle the ingress configuration. I am not sure when this will discussed in #sig-network

errm:
I *think* that I am fine with this... I think it can cause some confusion, but since the default will behave no differently from the current situation I think this is ok.

I would want a clear paragraph explaining the options in the documentation before we merge this this though...

yuvipanda:
w00t, thanks! Should I figure out where to add the docs and just add it now or wait for a code review ok first?

ldez:
You can write the documentation right now. This can facilitate the code review.

docs/configuration/backends/kubernetes.md
```diff
@@ -51,6 +51,15 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 #
 # labelselector = "A and not B"
 
+# Value of `kubernetes.io/ingress.class` annotation that identify Ingress objects to be processed.
+# Note that Ingress objects with an empty or nonexistent `kubernetes.io/ingress.class` value
+# will also be processed.
+#
+# Optional
+# Default: traefik (process all Ingresses)
```
timoreimann: I think we can make things easier by choosing the empty string as default. Specifically:

1. We don't specifically need to set `IngressClass` to `traefik` in tests.
1. We can simplify `shouldProcessIngress` (see below).

provider/kubernetes/kubernetes.go
```diff
@@ -135,6 +136,20 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 	return nil
 }
 
+func (p *Provider) shouldProcessIngress(ingressClass string) bool {
+	ingressClassRequired := "traefik"
```
timoreimann: We could make a constant out of this like

```go
const ingressClassRequired = "traefik"
```

to indicate that this is not a string to be mutated.

That said, I don't think we need to variable anyway though if we decide to use the empty string as default value for `p.IngressClass`.

provider/kubernetes/kubernetes.go
```diff
@@ -135,6 +136,20 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+	if p.IngressClass != "" {
+		ingressClassRequired = p.IngressClass
+	}
+
+	switch ingressClass {
```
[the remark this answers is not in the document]
timoreimann: Appreciated! I think it was me who originally made this more complicated than necessary. :/

provider/kubernetes/kubernetes.go
aledbf: @timoreimann you should not this but use a flag with a default (""). This allows the user to deploy multiple controllers with different class annotations to expose different ingress

timoreimann: Ah sorry, late night logic flaw on my end. @aledbf is right, we need the default of `"traefik"`.

@yuvipanda please ignore my comments in this regard.

timoreimann: Idea: Keep the default of `p.IngressClass` to `traefik` and still simplify the method implementation like this:

```go
func (p *Provider) shouldProcessIngress(ingressClass string) bool {
        return ingressClass == "" || ingressClass == p.IngressClass
}
```

I _think_ this _should_ work, though only tests will bring certainty...

yuvipanda:
Thanks for the review everyone! I'm caught up with other work this week, but will try to find some time within the next few days to address them!

timoreimann:
I like the prefix idea too, even though it means diverging from the Nginx equivalent option somewhat.

yuvipanda:
That looks good to me too! I'm making the required changes now.

yuvipanda:
So looks like the things I need to do still are:

1. Figure out where defaults are set, and set default for IngressClass to 'traefik'. The only place I could find is in `cmd/traefik/configuration.go`. Is that the right place? I've already modified it...
2. Figure out where validation logic for this config goes, and make sure that this value is either 'traefik' or starts with 'traefik-'
3. Write an additional test.

Does that sound right? 

I'm still sortof new to go, so any pointers for where to find (1) and (2) (or links to similar PRs / functionality in other providers) will be highly appreciated!

yuvipanda:
(I also did a quick rebase to prevent my branch from getting too stale)

timoreimann:
@yuvipanda 

1. I think you need to add a line [here](https://github.com/containous/traefik/blob/7192aa86b5264e014ceabceb6367d8e6ce9a2fe6/cmd/traefik/configuration.go#L118).
1. We need to validate the annotation for every Ingress object we're inspecting. Presumably, the check should happen inside `shouldProcessIngress` and log in case we use an illegal prefix.
1. Yes. :-)

Side note: I think we should allow the prefix `traefik`, not only `traefik-`, supporting cases where users want to have something like `traefik1`, `traefik2`, `traefik42`, etc. The dash gives extra structure, which in my opinion should be up to the user to decide for or against.

yuvipanda:
@timoreimann thank you for your quick response!

I've already done (1) early on, so I've just gotten rid of the custom check in shouldProcessIngress. I wasn't sure if (1) sets the defaults everywhere (including tests) or just when running the binary.

For (2), can you help me understand why we validating that there instead of wherever it is the config.toml file is being parsed? Wouldn't that help us error out earlier? Or do we need to do this for each ingress because traefik's configuration is so dynamic?

I'll work on tests next!

timoreimann:
@yuvipanda sorry, I misunderstood (2) that it was about filtering Ingresses for the traefik prefix.

My best bet for validating the configuration option would probably be [just before the Kubernetes client is created inside `Provide()`](https://github.com/containous/traefik/blob/7192aa86b5264e014ceabceb6367d8e6ce9a2fe6/provider/kubernetes/kubernetes.go#L78). That function is called to initialize the Kubernetes provider.

timoreimann:
@yuvipanda I _think_ the defaults are **not** set when the tests run. It might be easiest to cover that part in an integration test for Kubernetes, but that's [yet to come](https://github.com/containous/traefik/issues/1889).

I believe your tests need to specify the Ingress class accordingly.

yuvipanda:
@timoreimann great. I've pushed (2) as well.

I think this is possibly causing some tests to fail. I'll take a look asap, but might be a few days before I can get to them.

Thank you for all the help so far, and I'm excited about getting my first patch merged into traefik at some point!

ldez:
@yuvipanda Could you correct the tests?

ldez:
@yuvipanda any news?

yuvipanda:
Sorry, been a bit sick and then swamped with other stuff. I am travelling
the upcoming week too unfortunately, I'll be back to this right after.

ldez:
@yuvipanda So that your PR is integrated in the 1.5, it is imperative that it is ready by Tuesday 21 December at the latest. :rocket:

ldez:
Sorry November

yuvipanda:
I unfortunately won't have time to work on this until december :( I'll have to miss 1.5.

I'd love to keep this open until then, but am also happy to close it until I'm ready to start working on it again if it clutters up your repo. Apologies!

timoreimann:
@yuvipanda I think it's okay to keep it around and finish it whenever you're ready.

ldez:
@yuvipanda any news?

yuvipanda:
None right now unfortunately (am prepping for my kubecon talk). Hopefully before end of the year! Someone else should also feel free to take this forward if they have time :)

nmengin:
Hello @yuvipanda. I will rebase your branch on master in the way to resolve conflicts.

provider/kubernetes/kubernetes.go
```diff
@@ -515,8 +522,8 @@ func equalPorts(servicePort v1.ServicePort, ingressPort intstr.IntOrString) bool
 	return false
 }
 
-func shouldProcessIngress(ingressClass string) bool {
-	return ingressClass == "" || ingressClass == "traefik"
+func (p *Provider) shouldProcessIngress(ingressClass string) bool {
+	return ingressClass == "" || ingressClass == p.IngressClass
```
timoreimann: I'm not sure if this expresses the desirable behavior. Here is what we should do from my understanding:

- If the user did **not** specify the `p.IngressClass` parameter, we need to process an Ingress object if the Ingress object annotation is either missing, has the empty value, or the `traefik` value. (This matches the current behavior where users cannot specify `p.IngressClass`.)
- If the user **did** specify the `p.IngressClass` parameter, we should only process an Ingress object if the Ingress object annotation value matches `p.IngressClass` exactly.

The current implementation seems to always process Ingress objects which are missing the annotation or come with an empty value.

See also how Nginx does it [here](https://github.com/nginxinc/kubernetes-ingress/tree/master/examples/multiple-ingress-controllers#using-multiple-ingress-controllers) and [here](https://github.com/kubernetes/ingress-nginx#running-multiple-ingress-controllers). I think we want to mimic that behavior.

docs/configuration/backends/kubernetes.md
```diff
@@ -50,6 +50,16 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 #
 # labelselector = "A and not B"
 
+# Value of `kubernetes.io/ingress.class` annotation that identifies Ingress objects to be processed.
+# Note that Ingress objects with an empty or nonexistent `kubernetes.io/ingress.class` value
+# will also be processed.
```
timoreimann: The section might need to be updated if my assumption on how we should behave (see my other comment below) is correct.

docs/configuration/backends/kubernetes.md
```diff
@@ -50,6 +50,16 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
...
+# Note that if `kubernetes.io/ingress.class` annotation  is not empty, only annotations with the prefix `traefik` will be processed.
+#
+# Optional
+# Default: empty (process all Ingresses)
+#
+# ingressClass = "traefik-custom-proxy"
```
timoreimann: I'd use the example value `traefik-internal` -- many users actually come with this specific use case.

provider/kubernetes/kubernetes.go
timoreimann: Also sorry -- I think I explained this part incorrectly on Slack. :( The logic is somewhat tricky here...

cmd/traefik/configuration.go
```diff
@@ -132,6 +132,7 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 	defaultKubernetes.Watch = true
 	defaultKubernetes.Endpoint = ""
 	defaultKubernetes.LabelSelector = ""
+	defaultKubernetes.IngressClass = ""
```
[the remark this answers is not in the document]
ldez: good catch, we don't need to initialize with `""`

provider/kubernetes/kubernetes_test.go
```diff
@@ -797,6 +798,166 @@ func TestIngressAnnotations(t *testing.T) {
...
+		t.Run(test.desc, func(t *testing.T) {
+			t.Parallel()
+
+			actual, err := test.provider.loadIngresses(client)
+			require.NoError(t, err, "error loading ingresses")
+
+			assert.Equal(t, test.expected, actual)
+		})
+	}
+}
```
timoreimann: One general note: I'm having major troubles reading this and other Kubernetes tests for several reasons:

- We tend to have a large chunk of (Ingress) fixtures up front and then another chunk of (configuration) expectations near the bottom. I cannot really understand a single test case without either scrolling up and down several times or switching back and forth between two tabs in my browser all the time. The usual Go table-driven style combines both the test fixture and the expectation in a single test case.
- Our test cases are often too coarse and include multiple frontends/backends in the assertion, instead of focusing on a single frontend/backend pair per test case. We should prefer a higher number of relatively small test cases over a smaller number of large-ish test cases.
- We always compare complete configurations. This often requires writing a lot of boilerplate and is inefficient when the kind of functionality we want to validate touches a particular part of the expected configuration only. For instance, in `TestIngressClassAnnotation` above we shouldn't have to worry about host headers, TLS settings, and server weights. Instead, we should only care whether a frontend/backend is included, in the most cheap way conceivable.

Overall, the current structure makes it very difficult to judge if a test is good, complete, and meaningful.

I can understand if we move any major improvement into a follow-up PR; I'd personally prefer though to start having new tests properly structured already.

provider/kubernetes/kubernetes_test.go
```diff
@@ -797,6 +798,166 @@ func TestIngressAnnotations(t *testing.T) {
 	assert.Equal(t, expected, actual)
 }
 
+func TestIngressClassAnnotation(t *testing.T) {
+	ingresses := []*v1beta1.Ingress{
+		buildIngress(
+			iNamespace("testing"),
+			iAnnotation(label.TraefikFrontendPassHostHeader, "true"),
```
timoreimann: Do we need this for our particular test? How does it relate to validating the Ingress class logic?

provider/kubernetes/kubernetes_test.go
```diff
@@ -797,6 +798,166 @@ func TestIngressAnnotations(t *testing.T) {
...
+			iAnnotation(annotationKubernetesIngressClass, traefikDefaultAnnotationValue),
+			iRules(
+				iRule(
+					iHost("other"),
+					iPaths(onePath(iPath("/stuff"), iBackend("service1", intstr.FromInt(80))))),
+			),
+		),
+		buildIngress(
+			iNamespace("testing"),
+			iAnnotation(label.TraefikFrontendPassTLSCert, "true"),
```
timoreimann: Do we need this?

provider/kubernetes/kubernetes_test.go
```diff
@@ -797,6 +798,166 @@ func TestIngressAnnotations(t *testing.T) {
...
+			sName("service1"),
+			sNamespace("testing"),
+			sUID("1"),
+			sSpec(
+				clusterIP("10.0.0.1"),
+				sType("ExternalName"),
+				sExternalName("example.com"),
+				sPorts(sPort(80, "http"))),
+		),
+		buildService(
```
timoreimann: Do we need two services necessarily?

If not, can we choose the most simple service configuration that serves our testing needs?

provider/kubernetes/kubernetes.go
```diff
@@ -102,6 +103,12 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 		return err
 	}
 
+	// We require that IngressClasses start with `traefik` to reduce chances of
+	// conflict with other Ingress Providers
+	if len(p.IngressClass) > 0 && !strings.HasPrefix(p.IngressClass, traefikDefaultAnnotationValue) {
+		return fmt.Errorf("kubernetes.io/ingress.class value has to be empty or start with the prefix %s, instead found %q", traefikDefaultAnnotationValue, p.IngressClass)
```
timoreimann: Can we use `%q` instead of `%s` too?

provider/kubernetes/kubernetes.go
```diff
@@ -60,7 +60,7 @@ const (
 	annotationKubernetesIsDevelopment           = "ingress.kubernetes.io/is-development"
 )
 
-const traefikDefaultRealm = "traefik"
```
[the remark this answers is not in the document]
nmengin: done

provider/kubernetes/kubernetes_test.go
nmengin: Even if I am agree with you I think it can be done in a new PR :wink:

timoreimann: :+1:

beyond-code-github:
> I think we should also extend the Kubernetes guide at https://github.com/containous/traefik/blob/d426126a9294af8df041e7e55cbd61b8b899a4e1/docs/user-guide/kubernetes.md#between-tr%C3%A6fik-and-other-ingress-controller-implementations.

Did this ever get completed? I just read this thread with interest but it's not clear how to make use of the new feature that this PR added.

Can I just add IngressClass = "traefik-xxx" to my .toml file?

timoreimann:
@beyond-code-github I could have sworn that I pushed a follow-up PR to update the docs. Looks like I never did though!

Thanks for the hint, will fix that soon. In the meantime, the comment to the [`ingressClass` option](https://docs.traefik.io/v1.6/configuration/backends/kubernetes/) should answer your question:

```
# Value of `kubernetes.io/ingress.class` annotation that identifies Ingress objects to be processed.
# If the parameter is non-empty, only Ingresses containing an annotation with the same value are processed.
# Otherwise, Ingresses missing the annotation, having an empty value, or the value `traefik` are processed.
#
# Note : `ingressClass` option must begin with the "traefik" prefix.
#
# Optional
# Default: empty
#
# ingressClass = "traefik-internal"
```

So the answer to your question is: yes, `traefik-xxx` will work because it comes with the necessary `traefik` prefix.

beyond-code-github:
Thanks for the swift response @timoreimann !

Issue #390, opened 2016-05-20
add support for zipkin tracing


jstrachan:
[zipkin](http://zipkin.io/) is a great way to do inter-microservice invocation tracing and visualisation. It'd be awesome to add zipkin support to traefik. 

There's notes on [zipkin instrumenting here](http://zipkin.io/pages/instrumenting.html). There's an existing go implementation of zipkin tracing at https://github.com/elodina/go-zipkin

jippi:
Any plan for this? :)

martinbaillie:
I took a quick look at this the other day. I was thinking I'd implement some base OpenTracing server/client spans for v1.5.

Vlaaaaaaad:
Any updates on this?

ldez:
Closed by #2587

PR #2237, opened 2017-10-10
OpenTracing support

### LINKED ISSUE #771: Support Opentracing
OpenTracing http://opentracing.io/ is an open standard for instrumenting code for distributed tracing.  They have a library for go https://github.com/opentracing/opentracing-go.

It does not on the face of it seem too difficult, but we should probably have a think about what architecture would support this and better metric gathering in a clean way.

I have no experience with such a tool in production so it would be good to get some views from anyone who is?

tcolgate:
### Description

Support tracing initial request arrival and outbound backend calls.
This adds hooks into the middleware to inject span tracing. It uses the OpenTracing API, and this
code uses the Jaeger Tracer. If this is something people want, I'll look at adding other opentracing compliant tracer support (e.g. zipkin).

Obviously needs lots of work to make this acceptable, but I wanted to see if people were interested.

Fixes #771, #390


ldez:
Huge !

@tcolgate Could you change the base branch to master?

We only accept bug fixes in 1.4.

tcolgate:
It wasn't immediately clear to me how dependencies are being handled (the glide scripts proved a little painful. What's the correct process to introduce a dependency?

ldez:
@tcolgate could you split your commit in 2 commits:
- first the code
- second the vendor + glide

If you authorize me, I can manage the glide part.

tcolgate:
@ldez  I have split the PR, and dropped the inclusion of the actual deps.
Happy for you to manage the glide part.

I'm thinking:
- have a tracing.tracer string settting to pick which tracer to use (perhaps default to jaeger
- have tracer specific setting under tracing.jaeger.... and tracing.zipkin.... 
- move the actual tracing loging int middleware/tracer/
- Potentially hook logging and metrics up to traefik's internal logging/metrics.

In terms of more functionality, it could be possible to add tracing awareness to other middlewares

The results are already quite pleasing, you get the initial ingress trace, and if you have, e.g., retry enabled, you get a trace for each attempt. Plus trace ids get allocated so should be propagated to downstream apps.

tcolgate:
Here's an example of a jaeger trace of a request to a traefik with 3 retries enabled to a bad backend.
tracing-example

martinbaillie:
Looking good @tcolgate thanks for getting started on OpenTracing! I swapped my production gateway from Zuul to Traefik recently. Tracing is the one thing I lost in that switch and has been on my list to add to the Traefik codebase.

I'm happy to help out any way I can. Also I _do_ use Zipkin currently and think abstracting from tracer is a good idea, so let me know if I can be of assistance there.

tcolgate:
Could you provide examples of spans from zuul, or point me at their code
Idn not quite sure what tags to include right now

martinbaillie:
Here's Spring Cloud's take on OpenTracing:
https://github.com/spring-cloud/spring-cloud-sleuth
Their version of Zuul is here:
https://github.com/spring-cloud/spring-cloud-netflix/tree/master/spring-cloud-netflix-core
but
I believe it'll ultimately use Sleuth as part of client-side load balancing
through Feign/Ribbon, like (here)[
https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignClient.java
]

tcolgate:
Started work on how to config different tracers, this code is untested right now due to time, but will work on it further.
  Once jaeger+zipkin are working, I'm going to move the tracing code into the middleware directly and remove the use of the httplib middleware in favour of a custom one

tcolgate:
- Rebased on master
- Added some utilities to let other things use tracing
- Added tracing to the retry middleware.

Retry also includes an example of adding log events to a trace.

tcolgate:
@martinbaillie  the zipzkin tracing is sending stuff to a server, but for some reason, only the first span is showing up, I'm not sure what could be causing that.

tcolgate:
Main middlewars are now instrumented.
Interestingly both jaeger and zipkin see the same shape spans from the zipkin export. Either the opentracing zipkin support is broken, or I'm not setting the tracing up properly.
Jaeger export is fine and working as expected.

tcolgate:
@idez I'm struggling to track down the zipkin issue. Jaeger funcionality seems good right now. I'm not sure what the best route forward is. Maybe ship v1 without zipkin?

tcolgate:
I'm about 80% sure zipkin issues are due to me doing something wrong.
I've checked out opentracing, zipkin-opentracing-go, and thrift, at the same commits as we are vendoring in this PR. The example app show proper, nested traces. Will keep digging

tcolgate:
@martinbaillie @ldez Fixed the zipkin issue. I have a few tidyup things to do around config. But things are working. Remaining:
- Allow configuration of sampling setting for zipkin
- Some test (opentracing has a mock, so that much should be easy enough)
- Work out if config handling is inline with the projects preferences.

tcolgate:
zipkin

middlewares/tracing/tracing.go
```diff
@@ -0,0 +1,204 @@
...
+}
+
+type baseMiddleware struct {
+	opName string
+	*Tracing
+}
+
+// NewSpanMiddleware creates a new middleware wraps a span around subsequent
+// middleware
+func (t *Tracing) NewSpanMiddleware(name string) negroni.Handler {
```
mmatur: `NewSpanMiddleware` seems to not be used. What is his purpose ?

ldez:
@tcolgate So that your PR is integrated in the 1.5, it is imperative that it is ready by Tuesday 21 December at the latest. :rocket:

ldez:
Sorry November

middlewares/tracing/tracing.go
tcolgate: I intended it as the main means for adding tracing to new middlewares, however that didn't pan out. I'll remove it.

tcolgate:
I've started tidying this up, but don't have a tonne of time. I'm hoping I shokd get it done in the next couple of weeks though.

tcolgate:
This is probably going to have to wait until the new year now. Sorry about that.

errm:
No problem, thanks for your work so far @tcolgate 👍

mmatur:
@tcolgate Due to you lack of time, we have decided to carrying your PR in #2587. 

Thanks for your hard work :clap: .

PR #2243, opened 2017-10-10
Add a note about redirection rule to precise how regex/replacement work.

nmengin:
### Description

The mechanism used to redirect URLs between entrypoints does not use the fields `regex` and `replacement`, which are overriden if defined.

The PR adds precision about this point in documentation.

### DIFF
--- a/docs/configuration/entrypoints.md
+++ b/docs/configuration/entrypoints.md
@@ -34,6 +34,9 @@ To redirect an http entrypoint to an https entrypoint (with SNI support).
       KeyFile = "integration/fixtures/https/snitest.org.key"
 ```
 
+!!! note
+    Please note that `regex` and `replacement` do not have to be set in the `redirect` structure if an entrypoint is defined for the redirection (they will not be used in this case).
+
 ## Rewriting URL
 
 To redirect an entrypoint rewriting the URL.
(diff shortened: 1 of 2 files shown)


docs/configuration/entrypoints.md
```diff
@@ -34,6 +34,9 @@ To redirect an http entrypoint to an https entrypoint (with SNI support).
       KeyFile = "integration/fixtures/https/snitest.org.key"
 ```
 
+!!! note
```
dtomcej: Instead of `!!!` can we use bolding or some sort of markdown based notification ;)

docs/configuration/entrypoints.md
```diff
@@ -34,6 +34,9 @@ To redirect an http entrypoint to an https entrypoint (with SNI support).
...
+    `regex` and `replacement` have not to be set in `redirect` structure during entrypoint redirection (they are not used).
```
dtomcej: I would use the wording:

Please note that `regex` and `replacement` do not have to be set in the redirect structure if an entrypoint is defined for the redirection (they will not be used in this case)

docs/configuration/entrypoints.md
```diff
@@ -47,6 +50,9 @@ To redirect an entrypoint rewriting the URL.
     replacement = "http://mydomain/$1"
 ```
 
+!!! note
+    `regex` and `replacement` are not used during entrypoint redirection (if `entrypoint` is set in `redirect` structure).
```
dtomcej: I would use the wording:

Please note, if you have `regex` and `replacement` used to rewrite a URL, ensure that you do not have `entrypoint` defined in the redirect structure, or they will be ignored (as noted above).

docs/configuration/entrypoints.md
ldez: We use `!!!` as Markdown syntax, it's related to our new documentation theme : http://squidfunk.github.io/mkdocs-material/extensions/admonition/

dtomcej: So sorry about that, didn't know!

PR #2248, opened 2017-10-11
New entry point parser.

ldez:
### Description

Fix a problem when you write `--entryPoints='Name:http Address::8080'` (you must add a space at the end : `--entryPoints='Name:http Address::8080 '`)

Side effect:
- entrypoint configuration is now case insensitive 🎉

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


configuration/configuration.go
```diff
@@ -252,20 +248,24 @@ func (ep *EntryPoints) Set(value string) error {
...
+
+	config := make(map[string]string)
+	for _, part := range sections {
+		field := strings.SplitN(part, ":", 2)
+		name := strings.ToLower(strings.Replace(field[0], ".", "_", -1))
+		if len(field) > 1 {
+			config[name] = field[1]
+		} else {
+			if strings.EqualFold(name, "TLS") {
+				config["tls_acme"] = "TLS"
```
nmengin: **TLS** should be in lower case (re use a possible constant?)?

configuration/configuration.go
```diff
@@ -193,55 +192,52 @@ func (ep *EntryPoints) String() string {
...
 func (ep *EntryPoints) Set(value string) error {
-	result, err := parseEntryPointsConfiguration(value)
-	if err != nil {
-		return err
-	}
+	result := parseEntryPointsConfiguration(value)
 
 	var configTLS *TLS
-	if len(result["TLS"]) > 0 {
+	if len(result["tls"]) > 0 {
```
nmengin: WDYT about using constants for all fields?

configuration/configuration_test.go
```diff
@@ -16,36 +16,36 @@ func Test_parseEntryPointsConfiguration(t *testing.T) {
...
+				"address":                  ":8000",
+				"ca":                       "car",
+				"tls":                      "goo",
+				"tls_acme":                 "TLS",
+				"redirect_entrypoint":      "RedirectEntryPoint",
+				"redirect_regex":           "RedirectRegex",
+				"redirect_replacement":     "RedirectReplacement",
+				"whitelistsourcerange":     "WhiteListSourceRange",
+				"proxyprotocol_trustedips": "192.168.0.1",
+				"compress":                 "true",
```
nmengin: Re use constants? :wink:

configuration/configuration.go
ldez: in this case, this must be in Upper case

nmengin: OK

configuration/configuration.go
ldez: use constant for 2 strings seems overkill

nmengin: I'm talking about all the strings used in entrypoint definition (`tls`, `redirect_entrypoint`...) because they are used in many places, in many files.
IMHO, it's easier to modify one place.

PR #2254, opened 2017-10-11
Add proxy protocol tests

emilevauge:
### Description

This PR adds some integration tests on Proxy Protocol.
Because of an issue with the need of mounting volumes within tests (from the test container), this PR also removes the first container layer. It now runs tests directly from the CI env.


emilevauge:
As there seems to need more work to do on tests before removing the root container, I'll close this one for now.

PR #818, opened 2016-11-08
Add systemd watchdog feature

guilhem:
:warning: May be not ready to be merged

This PR add systemd behavior for watchdog

This is a proof of concept:

- [x] Discussion may happen on way to do healthcheck (currently an `HEAD` on /ping)
- [X] PR https://github.com/coreos/go-systemd/pull/209 have to be merged

PR #2283, opened 2017-10-19
Use healthcheck for systemd watchdog

guilhem:
Mutualise code for `traefik healthcheck` command and watchdog.

Link to first version of watchdog implementation #818

### DIFF
--- a/cmd/traefik/healthcheck.go
+++ b/cmd/traefik/healthcheck.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


emilevauge:
Hey @guilhem, thanks for your contribution :)
Can I ask you what is the real need to check the healthcheck in the systemd watchdog ?

guilhem:
@emilevauge of course you can :)

Currently, systemd watchdog only tests if a goroutine is working... that's not really a true healthcheck.
The server can be blocked, but watchdog ok.

With this kind of healthcheck, any problem on TCP stack (too many file open) and HTTP server is triggered

emilevauge:
Yeah, I understand.
But the healthcheck is not on the same http server than the reverse proxy, so I'm wondering if it's worth it.
WDYT ?

guilhem:
yes of course, but it's acting like any other healthcheck will do by requesting `/ping`.

In fact, I only use your work on command line `healthcheck` and generalize it to be used by systemd watchdog.

Any improvement in healthcheck function can be done :)

PR #2287, opened 2017-10-20
Fix data races.

tcolgate:
This fixes a race between dumping the config and setting up the stats.

### DIFF
--- a/api/handler.go
+++ b/api/handler.go
@@ -19,9 +19,9 @@ type Handler struct {
 	Dashboard             bool   `description:"Activate dashboard" export:"true"`
 	Debug                 bool   `export:"true"`
 	CurrentConfigurations *safe.Safe
-	Statistics            *types.Statistics `description:"Enable more detailed statistics" export:"true"`
-	Stats                 *thoas_stats.Stats
-	StatsRecorder         *middlewares.StatsRecorder
+	Statistics            *types.Statistics          `description:"Enable more detailed statistics" export:"true"`
+	Stats                 *thoas_stats.Stats         `json:"-"`
+	StatsRecorder         *middlewares.StatsRecorder `json:"-"`
 }
 
 var (
--- a/healthcheck/healthcheck.go
+++ b/healthcheck/healthcheck.go
@@ -48,6 +48,7 @@ type BackendHealthCheck struct {
 
 //HealthCheck struct
 type HealthCheck struct {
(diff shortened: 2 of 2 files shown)


tcolgate:
With these changes, test-unit passes.

tcolgate:
test-integration still fails, something to do with use of a buffer. Still investigating.

build.Dockerfile
```diff
@@ -1,8 +1,11 @@
-FROM golang:1.9-alpine
+FROM golang:1.9
```
emilevauge: Why did you change this ?

tcolgate:
The race detector requires a glibc base image. I left these commits in
incase you wish to run all tests with -race in future. It's probably a good
thing to do, but does slow down the tests.

tcolgate:
@emilevauge I've pushed what I have. I can't get the integration tests passing locally. The consul_catalogue tests pass, but whatever comes next hangs and it's not obvious from the timeout where. There may be a better approach that the Pipe for getting the output (the original approach requires locking around the buffer though, the Pipe seems cleaner).
  I'm not going to have much time to work on this. FWIW, I think the race in the integration tests appears to be purely to do with the test harness.
  If we can't see an obvious solution to the integration tests, I'd suggest at least meging the Web.Stats and healthcheck patches, since those are genuine races.

tcolgate:
I've started making a bit more progress, but there are still a lot of failing tests. staert seems to have an actual issues, ConsulSuite.TestDatastore is the main sticking point right now.

emilevauge:
@tcolgate I think it would be better to split this PR into 2: fixes on code and fixes on tests. It would help merging more quickly code races. WDYT?

tcolgate:
@emilevauge I'll leave this branch for critical fixes then, and open another PR with the test fixes.

timoreimann:
Shouldn't the PR come with a slight change to our build process, namely the activation of the race detector? At least temporarily so that we can see if race detector-enabled builds succeed now?

tcolgate:
We can enable the unit tests with these two, but not the integration tests,
which still fail.

timoreimann:
That sounds reasonable to me.

ldez:
@tcolgate Could you rebase on 1.5 ?

emilevauge:
@tcolgate we rebased your PR and kept your fixes on Stats & heathchecks.

PR #2290, opened 2017-10-20
Add ability to generate a unique X-Request-Id for each incoming request

FriggaHel:
This PR enable Traefik to generate unique X-Request-ID for incoming requests.
It allows to correlate it with every downstream requests.
It adds it to the logs too.


emilevauge:
Hey @FriggaHel, thanks for your contribution :+1: 
We understand the use case of this but we wonder if this PR is not conflicting with Open Tracing support added in https://github.com/containous/traefik/pull/2237. Don't you think we could use OpenTracing instead?
/cc @containous/traefik

emilevauge:
Ping @FriggaHel ?

FriggaHel:
Hey @emilevauge !

My usecase is: Find all logs for a specific request identified by an ID.

So, basically, we are using a micro-service architecture where each service logs parts of its treatment.

As I understood (because I never used it) OpenTracing is more for deep analysis and tracing requests time processing, so it can be used for this but seems a bit overkill to me.

In my particular use-case:
ELBv2 <==> Traefik <==> API Gateway (Zuul Proxy) <==> Micro-Services

With this feature, I can basically kick-down API Gateway since it most usefull usage is to add a uniq Request ID.

I agree this is a bit specific to my usage, and thinking forward, OpenTracing may be a better choice, if I'm the only one requiring this kind of feature.

emilevauge:
Hi @FriggaHel, we discussed internally, and after much hesitation, we are sorry to say that we prefer to focus on openTracing / Jaeger rather than implementing our own solution (even if it's simple).
Anyway, thanks for you contribution !
:heart:

hfern:
@emilevauge 

So if my understanding is correct there is no way of propagating a unique trace id using traefik presently?

jrm780:
Is there a chance this may be re-considered? `X-Request-Id` is a non-standard but very common header, and it's incredibly useful in correlating log entries. Distributed tracing aims to solve entirely different problems, so it would seem odd for OpenTracing support and adding an `X-Request-Id` header to be mutually exclusive.

aaslamin:
+💯 to this feature.

I was quite surprised to see this issue closed because of opentracing. It is very common for edge proxies to generate a unique ID for each request:

• http://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_id
• https://www.envoyproxy.io/docs/envoy/latest/configuration/http_conn_man/headers#config-http-conn-man-headers-x-request-id

One can use this to hydrate their access logs to create request scoped loggers for example as a request trickles through the service mesh. Tracing complements logging and metrics.

@emilevauge could please reconsider this?

aaslamin:
Furthermore, depending on volume, one does not create a trace for every request as per identified in [Dapper](https://ai.google/research/pubs/pub36356), whereas creating and propagating a unique ID is a cheap & effective way of having some level of insight to each and every request.

grachevko:
I also think this is a very strange decision to decline this simple solution. 
We use EFK to collect logs from many services, and we don't want use another software.
We want just have one more `ReqiestId` field in filters.

dplummer:
Is there any work around to get an `X-Request-Id` generated on each request and logged with Traefik? It would really help with troubleshooting some very infrequent errors.

Sarke:

```toml
[tracing]
  [tracing.jaeger]
    samplingParam = 0
    traceContextHeaderName = "uber-trace-id"
```

```yml
TRAEFIK_TRACING: "true"
TRAEFIK_TRACING_JAEGER_TRACECONTEXTHEADERNAME: "X-Request-ID"
```

yongzhang:
super surprised this was rejected, I think this is a basic feature as a proxy

PR #2304, opened 2017-10-23
Update github.com/xenolf/lego to 0.4.1

oldmantaiter:
- Mainly to be able to use the `AWS_HOSTED_ZONE_ID` to give the zone ID
  for a private zone
- Also fixes glide dependency problem where [github.com/pyr/egoscale](https://github.com/pyr/egoscale), a
  dependency of lego, reorganized its repo structure, moving everything from
  `src/egoscale` to the root of the repository


ldez:
thanks for the contribution 👍 

Do you authorize me to fix your PR?

ldez:
If you want to do it by yourself: you must update only the required dependencies.

oldmantaiter:
@ldez - Is this a limitation of using Glide? And I assume I must manually edit the `glide.lock` and `glide.yaml` files to only include updates for the dependencies of this package?

ldez:
Glide always update all dependencies, but we don't want that.

To explains very quickly how to do:
- make a `glide update` 
- revert each non required line.
- revert vendor
- make a `glide install`

oldmantaiter:
Ok. I figured and much and just did that and pushed, waiting on semaphore now.

oldmantaiter:
@ldez - Looks like that did the trick, I'm going to run some tests to ensure the lego functionality I am looking for works as expected.

oldmantaiter:
While testing I noticed a regression from another PR (#2050) - I'll fix that one first. There are template functions that are not defined in the function map.

glide.lock
```diff
@@ -1,5 +1,5 @@
-hash: 90d53da9a6eaba85d524d95410051ff7f12b1f76181df2ad4796b2439f3a2a37
-updated: 2017-10-24T14:08:11.364720581+02:00
+hash: 70e3ed39d2777d36b679ff55b8b0f2eb8d0b239fdfb8108f41b8b7239bc7920c
+updated: 2017-10-25T10:18:20.438135-04:00
```
ldez: could you remove extra spaces?

oldmantaiter: Fixed, thanks.

PR #2324, opened 2017-10-25
Add labels for `traefik.frontend.entryPoints` & `PassTLSCert` to Kubernetes

ryarnyah:
### Description

Add labels:
- `traefik.frontend.entryPoints`
- `traefik.frontend.passTLSCert`

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -47,6 +47,7 @@ type Provider struct {
 	Token                  string     `description:"Kubernetes bearer token (not needed for in-cluster client)"`
 	CertAuthFilePath       string     `description:"Kubernetes certificate authority file path (not needed for in-cluster client)"`
 	DisablePassHostHeaders bool       `description:"Kubernetes disable PassHost Headers" export:"true"`
+	EnablePassTLSCert      bool       `description:"Kubernetes enable Pass TLS Client Certs" export:"true"`
 	Namespaces             Namespaces `description:"Kubernetes namespaces" export:"true"`
 	LabelSelector          string     `description:"Kubernetes api label selector to use" export:"true"`
 	lastConfiguration      safe.Safe
@@ -165,24 +166,15 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 					}
 				}
 
-				PassHostHeader := p.getPassHostHeader()
-
-				passHostHeaderAnnotation, ok := i.Annotations[types.LabelFrontendPassHostHeader]
-				switch {
-				case !ok:
-					// No op.
-				case passHostHeaderAnnotation == "false":
-					PassHostHeader = false
-				case passHostHeaderAnnotation == "true":
-					PassHostHeader = true
-				default:
-					log.Warnf("Unknown value '%s' for %s, falling back to %s", passHostHeaderAnnotation, types.LabelFrontendPassHostHeader, PassHostHeader)
-				}
+				passHostHeader := getAnnotationPassHostHeader(i, p)
+				passTLSCert := getAnnotationPassTLSCert(i, p)
 				if realm := i.Annotations[annotationKubernetesAuthRealm]; realm != "" && realm != traefikDefaultRealm {
 					log.Errorf("Value for annotation %q on ingress %s/%s invalid: no realm customization supported", annotationKubernetesAuthRealm, i.ObjectMeta.Namespace, i.ObjectMeta.Name)
 					delete(templateObjects.Backends, r.Host+pa.Path)
 					continue
 				}
+
+				entryPoints := getEntrypoints(i)
 
 				whitelistSourceRangeAnnotation := i.Annotations[annotationKubernetesWhitelistSourceRange]
 				whitelistSourceRange := provider.SplitAndTrimString(whitelistSourceRangeAnnotation)
@@ -200,12 +192,14 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 
 					templateObjects.Frontends[r.Host+pa.Path] = &types.Frontend{
 						Backend:              r.Host + pa.Path,
-						PassHostHeader:       PassHostHeader,
+						PassHostHeader:       passHostHeader,
+						PassTLSCert:          passTLSCert,
 						Routes:               make(map[string]types.Route),
 						Priority:             priority,
 						BasicAuth:            basicAuthCreds,
 						WhitelistSourceRange: whitelistSourceRange,
 						Redirect:             entryPointRedirect,
+						EntryPoints:          entryPoints,
 					}
 				}
 				if len(r.Host) > 0 {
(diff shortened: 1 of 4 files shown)


emilevauge:
Ping @dtomcej or @timoreimann, WDYT ?

timoreimann:
Ah wait, this doesn't seem to be about a nee annotation, so maybe no docs needed.

Would it be worthwhile to extend the guide somehow though? @dtomcej, WDYT?

provider/kubernetes/kubernetes.go
```diff
@@ -165,25 +166,16 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
-					// No op.
-				case passHostHeaderAnnotation == "false":
-					PassHostHeader = false
-				case passHostHeaderAnnotation == "true":
-					PassHostHeader = true
-				default:
-					log.Warnf("Unknown value '%s' for %s, falling back to %s", passHostHeaderAnnotation, types.LabelFrontendPassHostHeader, PassHostHeader)
-				}
+				PassHostHeader := getRulePassHostHeader(i, p)
+				PassTLSCert := getRulePassTLSCert(i, p)
```
timoreimann: The two variables probably don't need to be exported. Can we decapitalize them?

provider/kubernetes/kubernetes.go
```diff
@@ -165,25 +166,16 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
 				if realm := i.Annotations[annotationKubernetesAuthRealm]; realm != "" && realm != traefikDefaultRealm {
 					log.Errorf("Value for annotation %q on ingress %s/%s invalid: no realm customization supported", annotationKubernetesAuthRealm, i.ObjectMeta.Namespace, i.ObjectMeta.Name)
 					delete(templateObjects.Backends, r.Host+pa.Path)
 					continue
 				}
 
+				EntryPoints := getEntrypoints(i)
```
timoreimann: Unexport this one too?

provider/kubernetes/kubernetes.go
```diff
@@ -199,10 +191,12 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 					templateObjects.Frontends[r.Host+pa.Path] = &types.Frontend{
 						Backend:              r.Host + pa.Path,
 						PassHostHeader:       PassHostHeader,
+						PassTLSCert:          PassTLSCert,
 						Routes:               make(map[string]types.Route),
 						Priority:             priority,
 						BasicAuth:            basicAuthCreds,
 						WhitelistSourceRange: whitelistSourceRange,
+						EntryPoints:          EntryPoints,
```
timoreimann: Are the entry points needed for the pass TLS functionality? If not, then I'd prefer to handle them in a separate PR.

provider/kubernetes/kubernetes.go
```diff
@@ -315,6 +309,48 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 	return &templateObjects, nil
 }
 
+func getEntrypoints(i *v1beta1.Ingress) []string {
+	entrypointsAnnotation, ok := i.Annotations[types.LabelFrontendEntryPoints]
+	if ok {
+		return strings.Split(entrypointsAnnotation, ",")
+	}
+	return nil
```
timoreimann: If the `nil` slice is handled equally to the empty slice case by Traefik, then we don't need the extra `ok` check it seems?

provider/kubernetes/kubernetes.go
```diff
@@ -315,6 +309,48 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+}
+
+func getRulePassHostHeader(i *v1beta1.Ingress, p *Provider) bool {
```
timoreimann: It's not really a _rule_ it seems to me. Can we drop that part from the function name?

provider/kubernetes/kubernetes.go
```diff
@@ -315,6 +309,48 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+		passHostHeader = false
+	case passHostHeaderAnnotation == "true":
+		passHostHeader = true
+	default:
+		log.Warnf("Unknown value '%s' for %s, falling back to %s", passHostHeaderAnnotation, types.LabelFrontendPassHostHeader, passHostHeader)
+	}
+	return passHostHeader
+}
+
+func getRulePassTLSCert(i *v1beta1.Ingress, p *Provider) bool {
```
timoreimann: Drop `Rule` from the name here too?

provider/kubernetes/kubernetes.go
```diff
@@ -315,6 +309,48 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+	case !ok:
+		// No op.
+	case passTLSCertAnnotation == "false":
+		passTLSCert = false
+	case passTLSCertAnnotation == "true":
+		passTLSCert = true
+	default:
+		log.Warnf("Unknown value '%s' for %s, falling back to %s", passTLSCertAnnotation, types.LabelFrontendPassTLSCert, passTLSCert)
+	}
+	return passTLSCert
```
timoreimann: Is the parsing logic identical to what we have in the function just above? If so, can we extract a little helper and use it from both implementation?

ldez:
So that your PR is integrated in the 1.5, it is imperative that it is ready by Tuesday 21 November at the latest. :rocket:

ldez:
@ryarnyah So that your PR is integrated in the 1.5, it is imperative that it is ready by Tuesday 21 November at the latest. :rocket:

provider/kubernetes/kubernetes.go
ryarnyah: Done.

provider/kubernetes/kubernetes.go
ryarnyah: Done.

provider/kubernetes/kubernetes.go
ryarnyah: Did i really need to make two separate PR?

provider/kubernetes/kubernetes.go
ryarnyah: `strings.Split` return a non empty slice when his first arg is empty, so if i didn't check extra `ok` , i will need to check if first of slice is empty. It seem more maintainable that way, no?

provider/kubernetes/kubernetes.go
ryarnyah: Done.

provider/kubernetes/kubernetes.go
ryarnyah: Done.

provider/kubernetes/kubernetes.go
ryarnyah: Done.

ryarnyah:
I also added some docs on these.

provider/kubernetes/kubernetes.go
timoreimann: Good point, I missed that property of `strings.Split`. Let's keep things like they are right now then.

provider/kubernetes/kubernetes.go
```diff
@@ -318,6 +312,39 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+func getEntrypoints(i *v1beta1.Ingress) []string {
+	entrypointsAnnotation, ok := i.Annotations[types.LabelFrontendEntryPoints]
+	if ok {
+		return strings.Split(entrypointsAnnotation, ",")
+	}
+	return nil
+
+}
+
+func getBoolAnnoatation(meta v1.ObjectMeta, name string, defaultValue bool) bool {
```
timoreimann: Typo: Annoatation -> Annotation.

ryarnyah: Done.

PR #2331, opened 2017-10-26
Fix IP address when Docker container network mode is container

### LINKED ISSUE #756: No IP in backend in shared network namespaces using –net=container:
I have a container that is running in shared network namespaces with another container (by using the option `--net=container:another-container`)

In that case, the new container will have the same network stack (and same IP) as the old container it is being linked to.

When traefik picks it, the backend's URL is set to http://:5050, the IP is missing

A docker inspect of the new container has no values in the Network part of the output (so i assume this is why traefik does not pick the IP). 

In that case, traefik should observe the `NetworkMode` value, which in that case is:

`"NetworkMode": "container:LINKED_CONTAINER_NAME"`,

so it can get the network settings of the new container from the network settings of the `LINKED_CONTAINER_NAME`

nmengin:
### Description

This PR allows returning an IP address for a Docker container which is in `container` network mode.

Fixes #756

### DIFF
--- a/provider/docker/docker.go
+++ b/provider/docker/docker.go
@@ -594,8 +594,23 @@ func (p *Provider) getIPAddress(container dockerData) string {
 
 	// If net==host, quick n' dirty, we return 127.0.0.1
 	// This will work locally, but will fail with swarm.
-	if "host" == container.NetworkSettings.NetworkMode {
+	if container.NetworkSettings.NetworkMode.IsHost() {
 		return "127.0.0.1"
+	}
+
+	if container.NetworkSettings.NetworkMode.IsContainer() {
+		dockerClient, err := p.createClient()
+		if err != nil {
+			log.Warnf("Unable to get IP address for container %s, error: %s", container.Name, err)
+			return ""
+		}
+		ctx := context.Background()
+		containerInspected, err := dockerClient.ContainerInspect(ctx, container.NetworkSettings.NetworkMode.ConnectedContainer())
+		if err != nil {
+			log.Warnf("Unable to get IP address for container %s : Failed to inspect container ID %s, error: %s", container.Name, container.NetworkSettings.NetworkMode.ConnectedContainer(), err)
+			return ""
+		}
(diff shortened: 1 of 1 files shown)


provider/docker/docker.go
```diff
@@ -594,10 +594,25 @@ func (p *Provider) getIPAddress(container dockerData) string {
...
 	// This will work locally, but will fail with swarm.
-	if "host" == container.NetworkSettings.NetworkMode {
+	if container.NetworkSettings.NetworkMode.IsHost() {
 		return "127.0.0.1"
 	}
 
+	if container.NetworkSettings.NetworkMode.IsContainer() {
+		dockerClient, err := p.createClient()
+		if err != nil {
+			log.Warnf("Impossible to get IP address for container %s, error: %s", container.Name, err)
```
mmatur: Could you replace `Impossible to get IP address for container...` by `Unable to get IP address for container...` :fr:

provider/docker/docker.go
```diff
@@ -594,10 +594,25 @@ func (p *Provider) getIPAddress(container dockerData) string {
...
+			return ""
+		}
+		ctx := context.Background()
+		containerInspected, err := dockerClient.ContainerInspect(ctx, container.NetworkSettings.NetworkMode.ConnectedContainer())
+		if err != nil {
+			log.Warnf("Impossible to get IP address for container %s : Failed to inspect container ID %s, error: %s", container.Name, container.NetworkSettings.NetworkMode.ConnectedContainer(), err)
```
mmatur: Could you replace `Impossible to get IP address for container...` by `Unable to get IP address for container...` :fr:

integration/docker_test.go
```diff
@@ -3,13 +3,15 @@ package integration
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
 	"time"
 
+	"io/ioutil"
```
ldez: Could sort your imports?

Issue #1689, opened 2017-05-31
Make sure each provider keeps frontends even if backends are missing/empty


timoreimann:
In order to achieve #1688, all providers must leave frontends around even if backends are missing or empty.

Kubernetes already does this correctly, Marathon presumably not; we need to go through each and fix things where needed. This could be as simple as [updating the template](https://github.com/containous/traefik/pull/1604/files#diff-d8939905996b0ac347a71f0ff77c5e66R2) to making changes to the provider implementation.

It's worth noting that for the case where a provider API isn't available temporarily, all implementations should back off and let Traefik reuse the previous configuration. Thus, there seem to be only two cases to still get a 404 when we wanted a 503 instead, namely:

1. A fresh Traefik instance is started while the provider API is currently unavailable.
2. The provider conceptually cannot yield empty frontends when backends are missing or unavailable.

List to providers that are passing the requirement already (please extend as progress is made):

- [x] Kubernetes
- [x] Marathon
- [ ] Docker

Refs #1077.

timoreimann:
Case 1 seems to be somewhat of a corner case. It could be remedied by means of exposing an "unready" state until the API has been accessed at least once successfully. An external process could then leverage the state information to mark the bootstrapping as incomplete until the ready state turns successful.

This should be considered only if we really deem it necessary, and if so be covered by a separate issue.

timoreimann:
Case 2 seems to be one that Consul is affected by, at least that's my understanding from #1077 https://github.com/containous/traefik/issues/1077#issuecomment-285503716.

@grobinson-blockchain and @bsphere, can you confirm? If so, we can continue working towards a solution. One idea that @emilevauge had was to use custom error pages as designed in #1634 and have a special backend return just the desirable error code. That'd keep Traefik free from special-casing for providers that cannot be changed according to our needs.

timoreimann:
Found to be working for Marathon as well.

beniwtv:
> "It's worth noting that for the case where a provider API isn't available temporarily, all implementations should back off and let Traefik reuse the previous configuration. Thus, there seem to be only two cases to still get a 404 when we wanted a 503 instead, ..."

There seems to be one more case of 404s. I am evaluating a 3-node cluster, and doing outage testing. Currently, if Consul has a small hiccup (say, due to a network issue) and returns error 500, Traefik currently removes all configuration, which makes the services running on it unavailable for a few seconds, returning a 404, until Consul can recover.

This is on Traefik v1.3.7.

[...]

PR #2333, opened 2017-10-26
Missing Backend key in configuration when application has no tasks

aantono:
Fixed an issue of missing Backend key in configuration when no tasks are available for an application

If Marathon application has no valid tasks in existence (either because the tasks are in bad state or they are in the process of being started or shut down), the generated configuration is completely missing the Backend key entry for the bound Frontend, resulting in a 404 response from Traefik instead of the expected 503.

### DIFF
--- a/provider/marathon/marathon_test.go
+++ b/provider/marathon/marathon_test.go
@@ -93,7 +93,9 @@ func TestMarathonLoadConfigNonAPIErrors(t *testing.T) {
 					},
 				},
 			},
-			expectedBackends: nil,
+			expectedBackends: map[string]*types.Backend{
+				"backend-app": {},
+			},
 		},
 		{
 			desc: "load balancer / circuit breaker labels",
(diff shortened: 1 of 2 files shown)


timoreimann:
Refs #1689.

timoreimann:
@dtomcej you probably want to have a look at this too.

dtomcej:
This directly addresses the cause of the issue. I am totally on board with this. 👍  This is not a band-aid approach :D

ldez:
@aantono Could you change the base branch to 1.4 and rebase? Thanks.

timoreimann:
I'll give this one a quick whirl just to be sure some time tonight.

aantono:
Should I rebase against `v1.4` branch?

timoreimann:
@aantono shouldn't be necessary: Traefiker (our bot) can handle this by now. If not, it'll complain. :)

PR #2344, opened 2017-10-29
dumpcerts.sh: Fix call to "base64" for Alpine

nknapp:
Alpine Linux' "base64" does not support the `--decode` option. Only `-d`. 
In order to run `dumpcerts.sh` inside the office "traefik:1.4-alpine" docker-image,
`base64` must be called with `-d`

### DIFF
--- a/contrib/scripts/dumpcerts.sh
+++ b/contrib/scripts/dumpcerts.sh
@@ -43,7 +43,7 @@ set -o nounset
 USAGE="$(basename "$0") <path to acme> <destination cert directory>"
 
 # Platform variations
-case `uname` in
+case $(uname) in
 	'Linux')
 		# On Linux, -d should always work. --decode does not work with Alpine's busybox-binary
 		CMD_DECODE_BASE64="base64 -d"
--- a/contrib/scripts/dumpcerts.sh
+++ b/contrib/scripts/dumpcerts.sh
@@ -42,6 +42,18 @@ set -o nounset
 
 USAGE="$(basename "$0") <path to acme> <destination cert directory>"
 
+# Platform variations
+case `uname` in
+	'Linux')
+		# On Linux, -d should always work. --decode does not work with Alpine's busybox-binary
+		CMD_DECODE_BASE64="base64 -d"
+		;;
+	*)
+		# Max OS-X supports --decode and -D, but --decode may be supported by other platforms as well.
+		CMD_DECODE_BASE64="base64 --decode"
+		;;
+esac
+
 # Allow us to exit on a missing jq binary
 exit_jq() {
 	echo "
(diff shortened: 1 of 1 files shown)


contrib/scripts/dumpcerts.sh
```diff
@@ -149,10 +149,10 @@ for domain in $(jq -r '.DomainsCertificate.Certs[].Certificate.Domain' ${acmefil
 	echo "Extracting cert bundle for ${domain}"
 	cert=$(jq -e -r --arg domain "$domain" '.DomainsCertificate.Certs[].Certificate |
          	select (.Domain == $domain )| .Certificate' ${acmefile}) || bad_acme
-	echo "${cert}" | base64 --decode > "${cdir}/${domain}.crt"
+	echo "${cert}" | base64 -d > "${cdir}/${domain}.crt"
```
timoreimann: Is `-d` supported in non-Alpine contexts as well?

If yes: cool :+1:

If no: shouldn't our script be smart enough to handle both option styles?

nknapp: The [man page](https://linux.die.net/man/1/base64) states both options. Ubuntu 10.04, Ubuntu 16.04, Centos 7 and Fedora 26 support the short-hand option.

Alpine is a bit special, because the installed binary for `base64` is actually the `busybox` binary. I guess that's the reason. I believe it is safe to use `-d` without checking for the environment.

timoreimann: The base64 binary shipped with Mac OS X seems to deviate:

```$ base64 --help
Usage:	base64 [-hvD] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -D, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
```

:(

nknapp: That's sad... I'll have to check if the is another base64 binary for alpine.

The alternative would be to check if the base64 binary is actually a symlink to busybox and then revert to the -d option. If that's an acceptable solution, I would change my PR.

timoreimann: How about just checking in the script whether [we run on Darwin](https://stackoverflow.com/questions/394230/detect-the-os-from-a-bash-script) and adjust the parameter accordingly?

contrib/scripts/dumpcerts.sh
```diff
@@ -42,6 +42,17 @@ set -o nounset
 
 USAGE="$(basename "$0") <path to acme> <destination cert directory>"
 
+case `uname` in
+	'Linux')
+		# On Linux, -d should always work. --decode does not work with Alpine's busybox-binary
+		alias decode_base64='base64 -d'
+		;;
+	*)
+		# Max OS-X supports --decode and -D, but --decode may be supported by other platforms as well.
```
timoreimann: Just to complete the picture: We could also invert the logic and switch-case on `Darwin` to set `--decode`, and use `-d` in all other cases (Linux, FreeBSD, you name it).

No idea what might be better. I'd go with your current approach when in doubt.

nknapp:
Sorry for the force-push. My last version did not work out (aliases don't work in bash-scripts, it seems). Now I'm using a variable to store the base64-decode command.

contrib/scripts/dumpcerts.sh
```diff
@@ -42,6 +42,18 @@ set -o nounset
 
 USAGE="$(basename "$0") <path to acme> <destination cert directory>"
 
+# Platform variations
+case `uname` in
```
timoreimann: Could you please use `$()` instead of backticks, just to follow [best practices](http://mywiki.wooledge.org/BashFAQ/082)?

nknapp: Done. If you want me to squash my commits, just tell me.

nknapp: Would by nicer with quotes around, I guess

timoreimann: True -- let's add the quotes for complete happiness. :)

No need to squash, our bot does that.

PR #1865, opened 2017-07-17
Add plugin support

emilevauge:
### Description
Like a phoenix, this PR supersedes #1370 and adds basic plugin support.

fenix

From https://github.com/containous/traefik/pull/1370#issuecomment-291903522:

>We can't load an external plugin if Traefik is built with CGO_ENABLED=0, and we really need this to build a statically linked golang executable to run in a Docker container.
golang/go#19569 and there are no plan on this golang/go#19569 (comment)

We will enable CGO for the Alpine based image. We then need to compile the binary in an Alpine based image with musl.

>If you compile traefik binary on your laptop, and a plugin in docker on your laptop, it does not work either: Error loading plugin: error opening plugin: plugin.Open: plugin was built with a different version of package github.com/containous/traefik/plugin

At each release, we will also now publish the Docker image used to compile Traefik binary (`build.Dockerfile`). This image will be needed to compile each plugin.

- The integration tests can't use a CGO Alpine based binary, therefore I need to generate a CGO binary during integration tests. The tests are a bit slow (1m) due to this compile. There may be a better solution.
- We need to generate 1 statically linked binary (as usual) and 1 CGO Alpine based binary. Scripts need to be updated.

Any comment/suggestion is welcome :)

Fixes #1336
Fixes #1048
Part of #30

PR #2362, opened 2017-11-06
gRPC plugin implementation

aantono:
### What does this PR do?

This PR adds support for plugins using [HashiCorp's Golang Remote Plugin](https://github.com/hashicorp/go-plugin) library.  It is based on the proposed implementation of #1865 with an assumption that the two implementations will converge.

### Motivation

The use of natively-compiled Golang binary requires a bit of complexity of having to compile the plugin together with the application and it is also limiting the use of only Go language.  The proposed PR relies on the use of `side-car` pattern where the plugin runs as a separate executable process, spawned and managed by the `go-plugin` library.  It uses Google Protocol Buffers gRPC communication mechanism, and thus allows to be integrated with a plugin implementation written in any language that supports `gRPC`. 

This capability will be great enhancement (IMHO) to allow various custom integrations into Traefik, which is especially useful for "Enterprises" that have some custom, proprietary logic/needs, without having to add more complexity into the application core itself.

### Performance Tests

I have ran some tests to see the performance aspects of this functionality and am happy to report that the plugin execution part itself takes mere `0.001` seconds, which is a very small overhead.


aantono:
For some reason `golint` fails the check on the protobuf generated code.  Found this commit in the `golint` repo that states it should not be happening - https://github.com/golang/lint/commit/a5f4a247366d8fc436941822e14fc3eac7727015

What am I doing wrong?

aantono:
Some performance testing updates:

I've ran the following benchmark: `wrk -t20 -c1000 -d60s -H "Host: test.traefik" --latency "http://traefik.localhost.com:8000/test?renderContent=true&stopChain=true"
Running 1m test @ http://traefik.localhost.com:8000/test?renderContent=true&stopChain=true` on an instance using the bare-bone plugin, which just returns empty response body and stops the chain. 

The output is the following:
```
Running 1m test @ http://traefik.localhost.com:8000/test?renderContent=true&stopChain=true
  20 threads and 1000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    73.70ms   95.86ms   1.87s    99.32%
    Req/Sec   173.61     79.47   510.00     67.89%
  Latency Distribution
     50%   68.38ms
     75%   74.85ms
     90%   81.47ms
     99%  105.39ms
  201795 requests in 1.00m, 30.60MB read
  Socket errors: connect 769, read 95, write 0, timeout 0
Requests/sec:   3357.97
Transfer/sec:    521.40KB
```

Attached DataDog metrics have specific numbers for the plugin part only, which is in milliseconds, ranging from:
`traefik.plugin.duration:0.000261|h|#plugin:plugin-test-go-grpc_darwin-amd64,error:false,order:before` and highest being
`traefik.plugin.duration:0.203950|h|#plugin:plugin-test-go-grpc_darwin-amd64,error:false,order:before`  (the number is in decimals of a second).  The majority distribution of those falls between `0.03` and `0.06` seconds.

```
1510 = cat perf.log | grep "0\.000" | wc -l
2741 = cat perf.log | grep "0\.00" | wc -l   
2741 = cat perf.log | grep "0\.00" | wc -l   
```

emilevauge:
@aantono thanks your initial analysis.
If I understand the maths well, the latency of a grpc plugin is ~ 30-60ms second vs 73ms avg for the whole reqaddeduest.
The overhead is not that small :/

aantono:
@emilevauge The numbers are very preliminary, I've just did a quick test on my dev Mac which has a bunch of "goodness" running, so while indicative, they are probably not fully accurate.  Having said that though, I actually found the numbers to be very acceptable, for what is being done under the hood.  Granted that it looks like the plugin takes the majority of the time of the request, but the number itself (30-60ms) is very small and will be acceptable for most use-cases where a custom plugin with custom logic would come into play.
I will try to setup a clean test environment and do a more "authoritative" set of tests, maybe even put a custom Traefik build in one of our lab envs to get some "real-usage" data.  Will keep you posted.

emilevauge:
@aantono 30-60ms seems acceptable at first, but compared to the 73ms of the whole request, this is not the same story. It means that 41-82% of the whole latency is due to the middleware.
Maybe we can optimize few things here and there but as far as we are on this latency percentage, I'm not sure this solution is reasonable for middleware plugins :(
WDYT @containous/traefik ?

dtomcej:
Honestly, @emilevauge I think if you are going to say that 60ms is unreasonable, it might be a good idea to come up with some framework benchmarks for middleware or plugin support _should be_ in regards to performance.

If we aim to have all middleware/plugins operate in 20ms or less, sure, we would have something to aim for. Just saying 60ms is too high without a reference to base off of is meh..

Thoughts?

aantono:
@emilevauge somehow I knew you were going to say that! 😜   Truth be told, I think that latency is usually the price to pay for capabilities.  If one wants really custom stuff, they should pay the price, and 30-60ms penalty is not a very high price to pay (in my mind, as we, for example, would be gladly accepting that increase for an ability to have a better/cleaner integration into our infrastructure and do custom stuff without having to maintain internal fork of Traefik)  I would definitely love to cooperate and see where/if/how we can bring that time down, also to maybe improve the accuracy of measurements, as what I did was very rudimentary and could be misrepresenting the true latency.

Do you have a particular number range/percentage in mind that you would feel good about?

timoreimann:
Looking at my organization for sanity-checking, there are services where an extra delay of 30-60ms would be a complete show-stopper and others where it would be no problem at all. In a real production environment, the delay may be even lower. So to me, the latency question is a case of YMMV.

The gRPC-based approach has some clear advantages: Language-agnostics and (potentially) the ability to add/remove plugins during runtime. Those are huge pluses. The Go-based approach lacks these benefits but will likely be faster (but otherwise doesn't seem to have any additional advantages?).

Ideally I'd like us to support both models as they have different trade-offs and use cases, but surely that comes with extra maintenance cost. Assuming that we want the gRPC approach for provider plugins anyway, is the remaining delta necessary to support gRPC middlewares possibly small enough to push the maintenance burden under the acceptance threshold?

aantono:
@timoreimann Those are all great points and I agree with you on the YMMV, since we also have things that are ms sensitive, so even a 30ms increase in latency would spell disaster (but many can and will tolerate the 30-60ms increase just fine)

I've tried to model my PR based on the original PR by @emilevauge #1865 , so the added pieces (IMHO) are the `remotePlugin.go` and `rpc.go`/`grpc.go` (those are very small rpc/grpc invocation stubs, which should not change after the initial implementation) + `plugin.proto` file, which doesn't really need ongoing maintenance either, unless more data fields need to be transferred to/from remote plugin.  The rest of the code is agnostic to the type of the plugin, so should be commonalities that can be leveraged.

emilevauge:
@aantono we need to think about this :) In the meantime, do you think you could do more detailed benchmarks on this (with/without plugin) ?

aantono:
@emilevauge You bet.  We are just finishing up certifying Traefik 1.4.2 for our prod deployment, so after that's done, will have some time to do a comprehensive perf test setup.  Let me know if there are more questions about the goals, functionality or the motivation that I can help answering for the group.

emilevauge:
@aantono thanks a lot for your effort on this PR. We are currently working on an alternate plugins solution for Traefik. As soon as we have something to show, we will give more details publicly :)
Besides, this PR has been inactive for quite a while so I'm closing this one for now.
Let's talk on Slack before writing any new code on this topic.
Thanks again for this great contribution :+1:

emilevauge:
Please follow #1336 for future implementations.

PR #2382, opened 2017-11-09
Fix raw path handling in strip prefix

marco-jantke:
This PR fixes forwarding of URL encoded characters to the backend servers. 

Given that Traefik receives a request path in the form: `/some/a%2Fb` it should forward the same path to the backend. This is the case when using e.g. the rule type `PathPrefix`, but as soon as the rule type is either `PathPrefixStrip` or `PathPrefixStripRegex` the path was forwarded in the decoded form like: `/some/a/b`.

The reason why this solution fixes our problem is that go uses the `String()` method of the `URL` object when it constructs the request. The docs state for this:

> // String reassembles the URL into a valid URL string.
// The general form of the result is one of:
//
//	scheme:opaque?query#fragment
//	scheme://userinfo@host/path?query#fragment
//
// If u.Opaque is non-empty, String uses the first form;
// otherwise it uses the second form.
// To obtain the path, String uses u.EscapedPath().

For us the relevant part is with `u.EscapedPath()` and there the docs state:

> // EscapedPath returns the escaped form of u.Path.
// In general there are multiple possible escaped forms of any path.
// EscapedPath returns u.RawPath when it is a valid escaping of u.Path.
// Otherwise EscapedPath ignores u.RawPath and computes an escaped
// form on its own.
// The String and RequestURI methods use EscapedPath to construct
// their results.
// In general, code should call EscapedPath instead of
// reading u.RawPath directly.

To understand the problem, note the sentence: `EscapedPath returns u.RawPath when it is a valid escaping of u.Path.` By stripping only the Path and not the RawPath we violated this statement and therefore the http package decided to use the `Path`.

### DIFF
--- a/middlewares/stripPrefix.go
+++ b/middlewares/stripPrefix.go
@@ -16,8 +16,11 @@ type StripPrefix struct {
 
 func (s *StripPrefix) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	for _, prefix := range s.Prefixes {
-		if p := strings.TrimPrefix(r.URL.Path, prefix); len(p) < len(r.URL.Path) {
-			r.URL.Path = "/" + strings.TrimPrefix(p, "/")
+		if strings.HasPrefix(r.URL.Path, prefix) {
+			r.URL.Path = stripPrefix(r.URL.Path, prefix)
+			if r.URL.RawPath != "" {
+				r.URL.RawPath = stripPrefix(r.URL.RawPath, prefix)
+			}
 			s.serveRequest(w, r, strings.TrimSpace(prefix))
 			return
 		}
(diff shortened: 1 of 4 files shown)


timoreimann:
We ran into the problem this PR fixes in our internal infrastructure and validated that the code change fixes things.

Is that good enough to classify the PR as `kind/bug/fix`, or is there more needed?

ldez:
Seems related to https://github.com/containous/oxy/pull/30

marco-jantke:
@ldez those issues are not related.

kachkaev:
@marco-jantke do you think that fixing https://github.com/containous/traefik/issues/1957 could be somewhat similar? That issue has been around for a while.

Issue #2067, opened 2017-09-06
Ability to blacklist header contents in access logs


indigoid:
### Feature request

I would like to be able to configure Traefik to not log the contents of named headers. Mostly `Authorization`, but probably others in the future.

This would help us avoid exposing and archiving header content that should be considered authentication secrets.

Related to #711.

### What did you do?

Running `experimental`-tag Traefik on Kubernetes with a mostly-defaults configuration, but with JSON logging to stdout. These logs are ingested by Fluentd and then fed to Graylog for alerting/analysis. This all works very well and has been quite reliable.

### What did you expect to see?

Header values logged to Graylog along with requests, but without contents of `Authorization` and other well-known headers that are used to convey secrets.

### What did you see instead?

All header values logged to Graylog along with requests.

### Output of `traefik version`: (_What version of Traefik are you using?_)

We are lazily tracking the `containous/traefik:experimental` public Docker repository. This is our current running pod:

```
Version:      b0ae6bc049ec4178f012e267ecfd2c72ffd9e7a9
Codename:     roquefort
Go version:   go1.9
Built:        2017-08-29_12:33:38AM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

* Kubernetes 1.7
* Traefik deployed via somewhat modified version of the [community Helm chart](https://github.com/kubernetes/charts/tree/master/stable/traefik) (to enable Prometheus metrics and JSON logging)   

Commandline args: `--configfile=/config/traefik.toml`

```toml
# traefik.toml
logLevel = "INFO"
defaultEntryPoints = ["http"]
[entryPoints]
  [entryPoints.http]
  address = ":80"
  compress = false
[kubernetes]
labelselector = "exposure = external"
[accessLog]
format = "json"
[web]
address = ":8080"
[web.metrics.prometheus]
  Buckets=[0.1,0.3,1.2,5.0]
```

### If applicable, please paste the log output in debug mode (`--debug` switch)

Not applicable.

timoreimann:
I think the expectation is missing a _not_.

indigoid:
That's what I would like, yes. I'm looking through the code trying to get my head around how Traefik logging operates. The below in `middlewares/accesslog/logger.go` (`logTheRoundTrip()`) is where the header filtering would need to be done, I think?

```go
for k := range logDataTable.Request {
  fields["request_"+k] = logDataTable.Request.Get(k)
}
```

[...]

PR #2397, opened 2017-11-12
Redact Authorization (and optionally other) headers

indigoid:
### What does this PR do?

* redact logged `Authorization` headers by default (default behaviour change! but I think it is a correct behaviour) 
* provide a mechanism to configure redaction of other headers in logs

### Motivation

My employer's production environment uses Traefik 1.4 with JSON logging feeding to both
Graylog (for short/medium term log access) and AWS S3 (long term storage). We do not
want to store authentication tokens. Previously logged an issue: #2067

### More

- [x] Added a test
- [x] Updated documentation

### Additional Notes

This will probably need rework as I am very new to Go, and am more ops than dev.

`make test` gave me a failure in the integration tests, but I reran it against `master` and it also failed there. Unit tests, including the test I added, passed.


middlewares/accesslog/logger.go
```diff
@@ -222,7 +234,11 @@ func (l *LogHandler) logTheRoundTrip(logDataTable *LogData, crr *captureRequestR
 	}
 
 	for k := range logDataTable.Request {
-		fields["request_"+k] = logDataTable.Request.Get(k)
+		if _, shouldredact := l.redact[k]; shouldredact {
+			fields["request_"+k] = "REDACTED_BY_TRAEFIK"
+		} else {
+			fields["request_"+k] = logDataTable.Request.Get(k)
+		}
```
timoreimann: We could also implement this block as following:

1. Store the redacted message in a variable.
2. If we should not redact: overwrite the variable with the value from the table.
3. Finally, set the value on the field.

This approach would enable us to mention the fields key once only.

WDYT?

timoreimann:
I got myself carried away but this really useful feature only to notice that we need an overall design review first -- sorry!

Consider my comments pending until Emile green-lights this one.

indigoid:
hey thanks for your feedback @timoreimann! FWIW I chose comma/semicolon because (a) copypasta [hence float mentions, oops!], and (b) it seemed prudent to be consistent with other options accepting lists of things, eg. when specifying Prometheus metrics buckets.

Will wait for Emile.

indigoid:
@timoreimann @ldez are you waiting on me to address the above comments? I'm happy to do so, I just deferred based on comments above and also the needs-design-review tag

ldez:
@indigoid sorry for the delay.
We will try to do a design-review asap.
Small team and lot of work  :speedboat:

indigoid:
@ldez great, thanks. Per comments on Slack — would you prefer me to try to add a delimited-list-of-strings parameter type (ie. `--option=baz,quux,sorge` becoming an array `["baz", "quux", "sorge"]` to `flaeg`? Seems like there'd be a number of applications for it, including in Traefik

ldez:
@indigoid the design-review is done, you can continue and apply the changes.

indigoid:
ace, thanks @ldez. I'll resync with current `master` and tidy things up :-)

middlewares/accesslog/logger.go
```diff
@@ -36,6 +36,7 @@ type LogHandler struct {
 	file     *os.File
 	filePath string
 	mu       sync.Mutex
+	redact   map[string]int
```
timoreimann: If this is just supposed to represent a set, then I'd use `bool` as value type. I _think_ this is also what the stdlib does in similar cases.

types/types.go
```diff
@@ -445,8 +445,33 @@ type TraefikLog struct {
 
 // AccessLog holds the configuration settings for the access logger (middlewares/accesslog).
 type AccessLog struct {
-	FilePath string `json:"file,omitempty" description:"Access log file path. Stdout is used when omitted or empty" export:"true"`
-	Format   string `json:"format,omitempty" description:"Access log format: json | common" export:"true"`
+	FilePath         string           `json:"file,omitempty" description:"Access log file path. Stdout is used when omitted or empty" export:"true"`
+	Format           string           `json:"format,omitempty" description:"Access log format: json | common" export:"true"`
+	HeaderRedactions HeaderRedactions `json:"headerredactions,omitempty" description:"List of names of sensitive headers to redact" export:"true"`
```
timoreimann: @ldez @emilevauge does `export` specify if this should be exported for usage statistics? If so, is it okay to include redacted headers?

Personally I'm not sure.

types/types.go
```diff
@@ -445,8 +445,33 @@ type TraefikLog struct {
...
+}
+
+// HeaderRedactions holds a list of HTTP header names
+type HeaderRedactions []string
+
+// Set splits string on , and adds the items to the list of headers to
```
timoreimann: No blank before the comma.

types/types.go
```diff
@@ -445,8 +445,33 @@ type TraefikLog struct {
...
+// be redacted
+func (b *HeaderRedactions) Set(str string) error {
+	headers := strings.Split(str, ",")
+	for _, header := range headers {
+		*b = append(*b, header)
```
timoreimann: What happens of someone adds blanks around header names? Shouldn't we trim each header?

timoreimann: What happens if someone uses an empty string (i.e., `Authorization,Secret,,Confidential`)? Shouldn't we skip empty values?

indigoid:
Update+resynced with `master`. Also added some unit tests in `types/`

danieladams456:
Does it look like this PR might make it into the 1.5.0 release?  I've been following this feature, and it will be very helpful!

types/types_test.go
```diff
@@ -35,3 +36,27 @@ func TestHeaders_ShouldReturnTrueWhenHasSecureHeadersDefined(t *testing.T) {
 
 	assert.True(t, headers.HasSecureHeadersDefined())
 }
+
+func TestHeaderRedactions_OneItemShouldBeOneItem(t *testing.T) {
+	redactions := HeaderRedactions{}
+	redactions.Set("Authorization")
+	assert.True(t, len(redactions) == 1)
```
timoreimann: We can use `assert.Len` here and in the other tests.

types/types_test.go
```diff
@@ -35,3 +36,27 @@ func TestHeaders_ShouldReturnTrueWhenHasSecureHeadersDefined(t *testing.T) {
...
+func TestHeaderRedactions_EmptyItemsShouldBeSkipped(t *testing.T) {
+	redactions := HeaderRedactions{}
+	redactions.Set(",Authorization,")
+	assert.True(t, len(redactions) == 1)
+}
+
+func TestHeaderRedactions_WhitespaceShouldBeStripped(t *testing.T) {
+	redactions := HeaderRedactions{}
+	redactions.Set(" Authorization  ")
+	assert.False(t, strings.Contains(redactions[0], " "))
```
timoreimann: I think I'd rather assert that the length is 1 and the sole element equals `Authorization`.

types/types_test.go
```diff
@@ -35,3 +36,27 @@ func TestHeaders_ShouldReturnTrueWhenHasSecureHeadersDefined(t *testing.T) {
...
+}
```
timoreimann: I think the first three tests could be combined into a single table-driven test using (parallelized) subtests. And if we were willing to test the content as well (which I believe would be a worthwhile addition), we could even cover all four tests.

WDYT?

timoreimann:
@danieladams456 unfortunately, the time window for features to still make it into the 1.5 release has closed several weeks time ago. This feature will be in the succeeding release.

middlewares/accesslog/logger.go
```diff
@@ -66,7 +67,18 @@ func NewLogHandler(config *types.AccessLog) (*LogHandler, error) {
 		Hooks:     make(logrus.LevelHooks),
 		Level:     logrus.InfoLevel,
 	}
-	return &LogHandler{logger: logger, file: file, filePath: config.FilePath}, nil
+
+	headerredactions := make(map[string]bool)
```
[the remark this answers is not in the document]
ldez: instead of a `map[string]bool` prefer `map[string]struct`

```go
for _, header := range config.HeaderRedactions {
	headerredactions[header] = struct{}{}
}
```

https://dave.cheney.net/2014/03/25/the-empty-struct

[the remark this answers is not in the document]
timoreimann: Personally I prefer `bool` as it seems less obscure compared to a `struct`, especially with regards to a set: an element is either part of a set or not.

timoreimann:
Re: wording: I haven't used the noun _redaction_ a lot myself; what I'm quite familiar with though is the term _redacted_ in the context of excluding certain pieces of information. So overall, it seems logical to me.

See also [the second paragraph in Wikipedia](https://en.m.wikipedia.org/wiki/Redaction). When in doubt, I'd trust @indigoid's judgement who is a native speaker to my knowledge.

middlewares/accesslog/logger.go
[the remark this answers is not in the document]
ldez: empty struct is a zero memory allocation tricks.

ldez:
@indigoid any news?

indigoid:
@ldez sorry have been really busy :-( I'll try to get back on this on the weekend

indigoid:
@timoreimann Initially I agreed with you re: `struct{}`-vs-`bool` but I read/thought about it a bit more and I agree with @ldez — this is a set of strings (header names) that request header names are checked against. The associated bool value doesn't even matter; what matters is whether it is there or not. Will fix.

Google yielded this discussion: https://groups.google.com/forum/#!topic/golang-nuts/xqqaR2N2sAI

timoreimann:
@indigoid:

>The associated bool value doesn't even matter; what matters is whether it is there or not. Will fix.

That's actually why I prefer booleans: "Whether it is there or not" essentially expresses a contains relationship, which naturally aligns with a true-or-false connotation.

I totally agree that a struct is more efficient. I'm pretty confident though that performance doesn't matter in this case. Readability should come first and only be replaced by a more performant, presumably less readable solution if benchmarks prove the benefits. That's also what the mailing list thread alludes to, and several other golang-nuts discussion emphasize more strongly and explicitly.

All that said, I'm totally okay with sticking to `struct` if that's what the majority prefers.

emilevauge:
@indigoid Thanks for this great contribution !
I know this PR has been up for a long time (sorry for that), but now that #2853 appeared, we now think it may be a good idea to implement something a bit more generic.
We had multiple discussions around this and we ended with a proposal: https://github.com/containous/traefik/issues/2875
WDYT?
_Side note: until we find a consensus, we think it would better to avoid any development on this topic :)_

Issue #1400, opened 2017-04-07
ECS Provider does not offer health check configuration


bbl232:
### What version of Traefik are you using (`traefik version`)?
1.2.0

### What is your environment & configuration (arguments, toml...)?
We're running many instances of Traefik in many ECS clusters...
```toml
accessLogsFile = "/var/log/traefik/access.log"

[web]
address = ":8080"

[ecs]

domain = "$CUSTOMER.$STACK"
Cluster = "$CUSTOMER-$STACK"
Watch = true
RefreshSeconds = 15
ExposedByDefault = true
```

### What did you do?
Searched documentation and code for how to configure the health checks

### What did you expect to see?
A label that I could apply to the containers to configure the health check endpoint and interval

### What did you see instead?
The ecs.tmpl and provider code did not show support for health checks

It seems like this wouldn't be too hard to add, but would be great to support. It would be even better if the health check failure could trigger a re-schedule of the task in ecs... or some sort of configurable hook that would execute on a failed health check.

timoreimann:
I suppose the kind of functionality that #1320 brings, just for ECS, would help you?

bbl232:
Yeah it sounds like it! Would be useful to add the concept of hooks as well, it would be nice to have some way to let ECS (or any other orchestrator) know that a container is misbehaving.

timoreimann:
@bbl232 I'm not familiar with ECS, but the orchestrators I know (Marathon, Kubernetes) allow to specify a health check within the application definition sent to the orchestrator's API so that the orchestrator knows when to reschedule a failing instance.

The kind of health check the above mentioned PR brings is more of a readiness check: It allows to take instances into / out of load rotation as they become (un-)ready; this is the case when the instance is failing but could also cover non-critical cases, like an instance losing one of its dependencies which prevent it from serving requests properly.

Usually the two concepts (when is an application instance healthy and when is it ready to serve requests) are distinguished. How would hooks fit in here? Could you explain how Traefik would potentially leverage them with regards to ECS?

Thanks.

bbl232:
@timoreimann ECS has a health check system, but only when using the AWS ELB or ALB as a load balancer. ECS will re-schedule failed/stopped tasks but that only works in the case of a crash, it would be nice to have a real health check that could actually tell wether the application is healthy or not. It would then be useful to have an unhealthy task simply killed, so that ECS re-schedules it.

[...]

PR #2421, opened 2017-11-18
Add health check label to ECS

oldmantaiter:
### What does this PR do?
- Adds support for health checks of ECS containers where you can set
  the path and interval of the check for a given backend (service)
- Partially deals with #1400

### Motivation

Even though ECS can provide status, it is (IMO) better to also check health from the load balancer to ensure the instance is up and running properly.

### More

- [x] Added/updated tests


healthcheck/healthcheck.go
```diff
@@ -148,6 +150,12 @@ func checkHealth(serverURL *url.URL, backend *BackendHealthCheck) bool {
 	client := http.Client{
 		Timeout: backend.requestTimeout,
 	}
+	if backend.Options.IgnoreCertificateErrors {
+		tr := &http.Transport{
```
Juliens: I think it should be better to use the existing transport with TLS client configuration used for backend instead of add a new option to healthcheck.
https://github.com/containous/traefik/blob/be306d651e999feb50f1f126eb68073d607324f3/server/server.go#L68

oldmantaiter: Ok. I'll pass that through

oldmantaiter:
@Juliens Updated to use the default server roundtripper, thanks for the review!

ldez:
@oldmantaiter Could you fix?

```

 M autogen/gentemplates/gen.go

Please do "go generate" to update the `autogen` package.
```

oldmantaiter:
@ldez Yep, locally the autogen validation test passes so just debugging why semaphore is not a fan

oldmantaiter:
@ldez - Fixed

oldmantaiter:
Looks like using the default  server round tripper in the health check has caused a connection leak. I'm still investigating but with health checks enabled my services start getting a ton of connections. I think this may need to be fixed to keep the client attached to the health check instead of making new requests each time.

PR #2424, opened 2017-11-18
Run Rancher tests cases in parallel.

ldez:
### What does this PR do?

Run Rancher tests cases in parallel.

### Motivation

Speed up tests.

### More

- [x] Added/updated tests

### DIFF
--- a/provider/rancher/rancher_test.go
+++ b/provider/rancher/rancher_test.go
@@ -18,10 +18,12 @@ func TestProviderServiceFilter(t *testing.T) {
 	provider.Constraints = types.Constraints{constraint}
 
 	testCases := []struct {
+		desc     string
 		service  rancherData
 		expected bool
 	}{
 		{
+			desc: "missing Port labels, don't respect constraint",
 			service: rancherData{
 				Labels: map[string]string{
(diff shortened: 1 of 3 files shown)


provider/rancher/rancher_test.go
```diff
@@ -104,16 +103,19 @@ func TestRancherServiceFilter(t *testing.T) {
...
 	}
 
-	for _, e := range services {
-		actual := provider.serviceFilter(e.service)
-		if actual != e.expected {
-			t.Fatalf("expected %t, got %t", e.expected, actual)
-		}
+	for _, test := range testCases {
+		test := test
+		t.Run("", func(t *testing.T) {
```
timoreimann: Can we use a non-empty first parameter for the test description?

(It's missing elsewhere too.)

ldez: It's hard to named tests that I don't created. 😉

timoreimann: How about using indices at least? It's far from ideal but better than nothing.

ldez: Done!

PR #2426, opened 2017-11-18
Add backend specific forwarding timeouts

oldmantaiter:
### What does this PR do?

Certain applications require a longer (or shorter) timeout than others. This
change enables use of a custom `DialTimeout` and `ResponseHeaderTimeout` in the
backend's `http.RoundTripper`.

### Motivation

This is a requirement for Traefik to replace our current proxy in our environment.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -476,6 +476,23 @@ Here is an example of backends and servers definition:
 - `backend2` will forward the traffic to two servers: `http://172.17.0.4:80"` with weight `1` and `http://172.17.0.5:80` with weight `2` using `drr` load-balancing strategy.
 - a circuit breaker is added on `backend1` using the expression `NetworkErrorRatio() > 0.5`: watch error ratio over 10 second sliding window
 
+### Forwarding Timeouts
+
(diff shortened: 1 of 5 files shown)


emilevauge:
We need to investigate a bit more on the `createHTTPTransport` function, sorry for the delay in our feedback.

emilevauge:
@oldmantaiter Sorry for the lag.
There is a major downside of recreating HTTP transports at each configuration change: this breaks keepalive connections with backends.
We agree on initial requirements, but we may find an alternative solution.
Closing this one for now, but thanks a lot for your contribution @oldmantaiter !

oldmantaiter:
@emilevauge I have no problem trying to find a better solution, I agree it is wasteful so I'll have a think about it and ask any questions here if that is OK with you. Just so I have a clear picture of what is required, you are looking for a solution where if the backend specific timeouts have *not* changed, no new transport is created.

If the backend specific timeouts have changed in a configuration reload/load, is it acceptable to assume that it is OK for the connections to be re-established with the backend hosts?

PR #2427, opened 2017-11-19
Configure the docker container with environment variables

thomasleveil:
### What does this PR do?

This PR adds a shell script in the Traefik docker image which uses it as an entrypoint. This new entrypoint looks for environment variables starting with `TRAEFIK_` and convert them to traefik command line options.

### Motivation

I deploy Traefik docker images on a few servers and while I have no issue with the current way of managing the docker container configuration (providing a .toml config with volumes, providing command line options), I'd like to use tools that orchestrate my docker containers for me. Such tools usually plays very well with environment variables.

Docker users are used to configure their containers through environment variables. This PR should make Traefik more straight forward to play with.

Also, I found an old issue (#7) requesting such a feature.

You can find a few examples on f717515 comment and a docker-compose.yml in the doc (6a3b060)

### More

- [ ] Added/updated tests

### Additional Notes

I'm not familiar with Golang integration tests. I figured out integration/docker_test.go defines a few tests around the Traefik docker container use, and my understanding of it is that those tests do not verify anything regarding passing command line options to traefik. 
I suppose then that it is fine not to add tests verifying those new environment variables do their job.

There is a drawback with this PR: the docker images size gain weight since it is now based on the Alpine docker image. Still this base image is small (2 MB) and since it is popular most users already have its layers on their systems.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
-FROM scratch
+FROM alpine
+COPY docker-entrypoint.sh /
 COPY script/ca-certificates.crt /etc/ssl/certs/
 COPY dist/traefik /
 EXPOSE 80
-ENTRYPOINT ["/traefik"]
+ENTRYPOINT ["/docker-entrypoint.sh"]
+CMD ["traefik"]
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+set -e
+
+# make traefik command line options from environment variables starting with 'TRAEFIK_'
+options=""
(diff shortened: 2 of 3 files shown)


ldez:

We made a tool to manage the different sources of configuration (CLI, TOML, KV): 
[staert](https://github.com/containous/staert)

The Docker file that you modified is not the Docker file that we used to publish, see [traefik-library-image](https://github.com/containous/traefik-library-image).

Your approach is interesting but we prefer to have support for environment variables in [staert](https://github.com/containous/staert).

We hope you understand why we must decline this change.

PR #2335, opened 2017-10-26
Split Web into API/Dashboard, ping, metric and Rest Provider

Juliens:
This PR changes the Web Provider implementation.

Now we can find all this "internal" routes separately
 - The API / Dashboard
 - The Ping
 - The Metrics
 - Rest Provider

Moreover, all this part, are configured on a entry point.
By doing it like this, we can use auth, whitelister etc.. for all internal routes.
We can attach those internal routes to the same entry point as the reverse proxy entry point

PR #2428, opened 2017-11-19
Register pprof handlers.

### LINKED ISSUE #1648: Performance profiling: Consider adding net/http/pprof endpoint
### Feature Request

Add [pprof](https://golang.org/pkg/net/http/pprof/) endpoint so we can get better detail into where and why the router's spending it's time.

[go-torch](https://github.com/uber/go-torch) should help a lot with visualization if we put this in place.

Example-torch

Likely we'll want to get this in place if we want to hit top-end limits for number of routing rules or time per routing rule.

There's some examples of getting graphviz output [here](https://blog.golang.org/profiling-go-programs).

Like this:

dotgraph of pprof output

timoreimann:
### What does this PR do?

The handlers will be hooked up under the `/debug/pprof` endpoint (enabled via the `--debug` flag).

### Motivation

This is to assist debugging performance problems.

### More

### Additional Notes

Closes #1648

### DIFF
--- a/api/debug.go
+++ b/api/debug.go
@@ -4,6 +4,7 @@ import (
 	"expvar"
 	"fmt"
 	"net/http"
+	"net/http/pprof"
 	"runtime"
 
 	"github.com/containous/mux"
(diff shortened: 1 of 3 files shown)


timoreimann:
I wasn't exactly sure why we have the debug mode documented twice. It might be that one is redundant -- let me know if I should drop one of the two.

@containous/traefik PTAL.

ldez:
It's because #2335. For now we have duplicated the doc.

PR #2432, opened 2017-11-20
Vendor generated file from template

ldez:
### What does this PR do?

Split the current assert generator in two parts:
- template
- static files

The generated file for templates is now put in the repository. 

### Motivation

#2117 

### More

### Additional Notes

Related to #1464,  #1534


timoreimann:
@ldez did we manage to resolve the problems mentioned in #1534 that caused the original attempt to vendor generated files to be reverted?

ldez:
@timoreimann I don't think so, because I vendor only one file on both.
Must be validate on a Windows system.

ldez:
@timoreimann Misreading : yes that resolve the problems.

PR #2437, opened 2017-11-20
Remote HTTP provider

aantono:
### What does this PR do?

Add support for Remote HTTP configuration provider

### Motivation

To enable custom integrations with various config management tools, the use of Remote HTTP configuration provider enables to call an HTTP endpoint either at pre-configured interval or using long-poll option to retrieve Traefik configuration in a TOML format

### More

- [x] Added/updated tests

### DIFF
--- a/cmd/traefik/configuration.go
+++ b/cmd/traefik/configuration.go
@@ -20,6 +20,7 @@ import (
 	"github.com/containous/traefik/provider/marathon"
 	"github.com/containous/traefik/provider/mesos"
 	"github.com/containous/traefik/provider/rancher"
+	"github.com/containous/traefik/provider/remote"
 	"github.com/containous/traefik/provider/rest"
 	"github.com/containous/traefik/provider/zk"
 	"github.com/containous/traefik/types"
@@ -48,6 +49,10 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 	// default Rest
 	var defaultRest rest.Provider
 	defaultRest.EntryPoint = configuration.DefaultInternalEntryPointName
+
+	// default Remote
+	var defaultRemote remote.Provider
+	defaultRemote.Watch = true
 
 	// TODO: Deprecated - Web provider, use REST provider instead
 	var defaultWeb configuration.WebCompatibility
@@ -234,6 +239,7 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 		File:               &defaultFile,
 		Web:                &defaultWeb,
 		Rest:               &defaultRest,
+		Remote:             &defaultRemote,
 		Marathon:           &defaultMarathon,
 		Consul:             &defaultConsul,
 		ConsulCatalog:      &defaultConsulCatalog,
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -22,6 +22,7 @@ import (
 	"github.com/containous/traefik/provider/marathon"
 	"github.com/containous/traefik/provider/mesos"
 	"github.com/containous/traefik/provider/rancher"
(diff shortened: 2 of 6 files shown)


timoreimann:
Hey there @aantono. Without having looked more thoroughly into the code, what's the difference between your extension and querying the `/api` endpoint (along with possibly converting the JSON to YAML)?

aantono:
@timoreimann The `/api` endpoint allows to query Traefik and do various manipulations.  The `remote` provider allows Traefik to *pull* remote HTTP endpoint for config, either once, or at interval or with long-polling.  Does that make sense?

timoreimann:
I think so. Couldn't this also implemented though by pushing configuration changes to the `/api` endpoint once or periodically?

aantono:
In theory "probably", the reality is that not all systems support push model (most are just do have http endpoints though), and also having to push config out to the entire cluster would require "knowing" where each node runs and have access to it (firewall rules typically allow egress, but not ingress).

timoreimann:
@aantono gotcha. Assuming we approve the overall usefulness of this feature, I wonder if it should be implemented by a dedicated provider or something to integrate into the web one.

But let's have Emile review the design first.

aantono:
@timoreimann I thought `web` was more about providing ways to interact with Traefik from outside, not the other way around, where Traefik calls external resources.  Am I wrong?

timoreimann:
@aantono you might be correct. Just throwing in random thoughts here. :-)

emilevauge:
@aantono 
>In theory "probably", the reality is that not all systems support push model (most are just do have http endpoints though), and also having to push config out to the entire cluster would require "knowing" where each node runs and have access to it (firewall rules typically allow egress, but not ingress).

Hmmmm, I think in this use case, it could make sense to share the configuration in a KV store when deploying multiple Traefik instances. WDYT?

(Before opening a pull request to add a new feature, I advise you to talk with the maintainer team before on Slack. This can avoid a lot of wasted time ;) )

aantono:
@emilevauge Sorry for not asking on Slack first, the coding part didn't take that long (was able to do it over a weekend), so I figured I'd just submit the code to have some substance to discuss over. ;)

On the use of KV vs HTTP pull, it's a matter of simplicity, I guess.  One could always build a bunch of bridges on their end to transform whatever their source of data is and either use Traefik API by pushing to it, or rely on KV like Consul or etcd and push there instead.  Those solutions require adding complexity in a form of a sidecar bridge that would be keeping those things in sync, where with HTTP Remote provider things become a lot simpler to maintain (IMHO).  But I could see if there is reluctance to add another provider, though the amount of code and complexity in brings are very small (IMHO).

Thoughts?

aantono:
Any thoughts on this PR, or should this be moved to Slack for discussion?

xalexslx:
Hi,

I'll give my option on this matter. Externalised configuration is useful for some environments where KV isn't setup by corporate rules but we've private network accessible by our applications. For instance, we develop in Java using the Spring Framework. For remote configuration, we'd use the spring cloud (https://github.com/spring-cloud/spring-cloud-config) for allowing remote configuration. By using spring cloud, we allow our applications to always pull the latest configuration file which is correctly versioned on private git repository. That way, we avoid having different configuration files than the versioned ones at our application servers. One addition that probably can turn this more useful is allowing the configuration url be something like "https://user:secret@myconfig.mycompany.com" allowing access to authenticated required resources.

Thanks.

emilevauge:
@aantono I'm really sorry but as nobody added any reactions (:+1:, :tada:) to this PR nor asked about this kind of feature before, we think we should close it for now. This may be a great candidate to be deployed as a plugin once we move forward on this (:pray:). Thanks again for your contribution :)

Issue #2438, opened 2017-11-20
Add tls support for k8s ingress resource


gopenguin:
### Do you want to request a *feature* or report a *bug*?
feature

### What did you do?
I've setup a k8s cluster, configured traefik, a webservice and an ingress resource with a tls cert in a secret.

### What did you expect to see?

Traefik reads the cert in the secret and uses this cert for tls termination and forwards the unencrypted requests to the backend service.

### What did you see instead?

Traefk didn't respect the tls property in the ingress and therefore didn't decrypt the https request.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.4.3
Codename:     roquefort
Go version:   go1.9.2
Built:        2017-11-14_11:14:24AM
OS/Arch:      linux/amd64

```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```
--web
--kubernetes
```

### If applicable, please paste the log output in debug mode (`--debug` switch)
N/A

gopenguin:
This issue is related to #378 which is also about tls in kubernetes but using acme and kubernetes secrets to store the aquired certs.

gopenguin:
This is resolved due to #2439 being merged.

PR #2439, opened 2017-11-20
Add support for fetching k8s Ingress TLS data from secrets

gopenguin:
### What does this PR do?

Enable Traefik to read the k8s secret defined in a k8s ingress resource via the tls property.

### Motivation

This allows a k8s user to easily setup a tls terminated endpoint

### More

- [x] Added/updated tests

### Additional Notes

This is my first try on this. I would be pleased to get your opinion on this. This is the implementation to #2438


ldez:
@gopenguin Could you fix?
```

 M autogen/gentemplates/gen.go

Please do "go generate" to update the `autogen` package.
```

gopenguin:
When I regenerate the file all file modes change from `os.FileMode(436)` to `os.FileMode(420)`. Is this supposed to happen?

ldez:
It's not the expected behavior.
Don't change nothing for now, I fix this problem in #2440.

gopenguin:
@ldez I've just pulled your changes and I think there is sill an issue with the generated file.  It isn't go fmt'ed but the version in the repo is.

gopenguin:
If you just run go generate, the file changes but after running go fmt the changes are gone.

gopenguin:
I had some time this morning and wrote some docs but I realy appreciate any help.

provider/kubernetes/kubernetes.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/containous/traefik/log"
 	"github.com/containous/traefik/provider"
 	"github.com/containous/traefik/safe"
+	traefikTls "github.com/containous/traefik/tls"
```
timoreimann: I think it's more conventional in Go to not camel case import names. I'd suggest `ttls` or `traefiktls` instead.

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,50 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 				}
 			}
 		}
+
+		for _, t := range i.Spec.TLS {
+			tlsSecret, exists, err := k8sClient.GetSecret(i.Namespace, t.SecretName)
+			if err != nil {
+				log.Errorf("Unable to retrive secret %s/%s: %s", i.Namespace, t.SecretName, err)
```
timoreimann: Typo: retrieve

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,50 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+				continue
+			}
+			if !exists {
+				log.Warnf("Unable to find secret %s/%s", i.Namespace, t.SecretName)
+				continue
+			}
+			_, tlsCrtExists := tlsSecret.Data["tls.crt"]
+			_, tlsKeyExists := tlsSecret.Data["tls.key"]
```
timoreimann: We seem to need both ignored variables later, so let's store them here for reusal below.

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,50 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+			if !tlsCrtExists || !tlsKeyExists {
```
timoreimann: Can we split this up into two dedicated errors?

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,50 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+				log.Warnf("Secret %s/%s is not of type 'kubernetes.io/tls'", i.Namespace, t.SecretName)
```
timoreimann: Is the error message correct?

provider/kubernetes/kubernetes.go
gopenguin: I thought, I'd reuse the prefix used elswhere in this project like in `acme/acme.go`, `cmd/traefik/traefik.go`, `server/server.go` and `types/types.go` but I can't remember why I introduced the prefix in the first place so I will remove it.

provider/kubernetes/kubernetes.go
gopenguin: Should there also be a specific error for both entries missing to speed up the errorhandling?

provider/kubernetes/kubernetes.go
gopenguin: Sorry, no, this message isn't correct anymore

provider/kubernetes/kubernetes.go
gopenguin: Never mind

provider/kubernetes/kubernetes.go
timoreimann: It's quite possible we used camel case elsewhere. If we do, we should fix those (by a separate PR).

provider/kubernetes/kubernetes_test.go
```diff
@@ -2262,6 +2274,169 @@ func TestBasicAuthInTemplate(t *testing.T) {
 	}
 }
 
+func TestTlsSecret(t *testing.T) {
```
timoreimann: TLS should be all caps.

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,56 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					log.Warnf("Secret %s/%s doesn't have an entry named 'tls.key'", i.Namespace, t.SecretName)
+				}
+				continue
+			}
+
+			var hosts []string
+			if len(t.Hosts) > 0 {
+				hosts = t.Hosts
+			} else {
+				// If no hosts are provided the hosts property should default ot the wildcard host. To ensure that the
```
timoreimann: Typo: `ot` instead of `to`.

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,56 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 				}
 			}
 		}
+
+		for _, t := range i.Spec.TLS {
+			tlsSecret, exists, err := k8sClient.GetSecret(i.Namespace, t.SecretName)
+			if err != nil {
+				log.Errorf("Unable to retrieve secret %s/%s: %s", i.Namespace, t.SecretName, err)
```
timoreimann: Please use _Failed_ instead of _Unable_ to be consistent with the existing wording.

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,56 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+				continue
+			}
+			if !exists {
+				log.Warnf("Unable to find secret %s/%s", i.Namespace, t.SecretName)
```
timoreimann: _find_ here and _retrieve_ above are very similar. Can we say something like _Secret %s/%s does not exist_?

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,56 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+			tlsConfig := &tls.Configuration{
+				EntryPoints: hosts,
+				Certificate: &tls.Certificate{
+					CertFile: tls.FileOrContent(tlsCrtData),
+					KeyFile:  tls.FileOrContent(tlsKeyData),
+				},
+			}
+
+			templateObjects.TLSConfiguration = append(templateObjects.TLSConfiguration, tlsConfig)
+		}
```
timoreimann: Can we put the TLS logic into a dedicated function?

provider/kubernetes/kubernetes_test.go
```diff
@@ -310,6 +311,7 @@ func TestLoadIngresses(t *testing.T) {
 				},
 			},
 		},
+		TLSConfiguration: []*tls.Configuration{},
```
timoreimann: Is it needed to specify an empty `TLSConfiguration` for tests unrelated to the TLS functionality?

If no, then I'd prefer to omit those here and further below.

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,56 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+			tlsKeyData, tlsKeyExists := tlsSecret.Data["tls.key"]
+			if !tlsCrtExists || !tlsKeyExists {
+				if !tlsCrtExists {
+					log.Warnf("Secret %s/%s doesn't have an entry named 'tls.crt'", i.Namespace, t.SecretName)
+				}
+				if !tlsKeyExists {
+					log.Warnf("Secret %s/%s doesn't have an entry named 'tls.key'", i.Namespace, t.SecretName)
+				}
+				continue
+			}
```
timoreimann: None of the four error cases above seem to have test coverage. Can we improve here?

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,56 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+				// tls config only matches rules in this ingress, add all the used hosts to simulate a wildcard like
+				// behavior.
+				hosts = []string{}
+
+				for _, r := range i.Spec.Rules {
+					if r.Host != "" {
+						hosts = append(hosts, r.Host)
+					}
+				}
+			}
```
timoreimann: My code coverage indicator tells me that the `else` branch does not have a corresponding test. We should change that.

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +310,56 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+
+			var hosts []string
+			if len(t.Hosts) > 0 {
+				hosts = t.Hosts
+			} else {
+				// If no hosts are provided the hosts property should default ot the wildcard host. To ensure that the
+				// tls config only matches rules in this ingress, add all the used hosts to simulate a wildcard like
+				// behavior.
+				hosts = []string{}
```
timoreimann: If `tls.Configuration.EntryPoints` is fine with nil slices, you can drop this line as we have already defined the nil slice in line 337, and Go can append to such slices without problems.

If `tls.Configuration.EntryPoints` requires a non-nil empty slice, then you can replace line 337 with this one (i.e., initialize an empty slice right away).

provider/kubernetes/kubernetes_test.go
gopenguin: Because I used the empty slice it was necessary (nil slice != empty slice) but I didn't know that you could append to a nil slice without any errors.

gopenguin:
Thanks for the review. I'll probably get it fixed by tomorrow evening.

provider/kubernetes/kubernetes.go
```diff
@@ -308,6 +309,13 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 				}
 			}
 		}
+
+		tlsConfigs, err := getTLSConfigurations(i, k8sClient)
+		if err != nil {
+			log.Errorf("Error configuring tls for ingress %s/%s: %s", i.Namespace, i.Name, err)
```
timoreimann: I'd like to make sure we make a conscious decision on the implications for an Ingress that has an invalid TLS configuration or whose TLS data cannot be retrieved from the corresponding secret. IIUC, with the current approach we simply do not set the TLS configuration. Does that mean it'd not be accessible via HTTPS but possibly via HTTP? IMHO, that'd be something we shouldn't tolerate and instead skip that Ingress together with a clear error log message.

The line is also missing test coverage, but let's answer the question first.

provider/kubernetes/kubernetes_test.go
```diff
@@ -2262,6 +2263,297 @@ func TestBasicAuthInTemplate(t *testing.T) {
...
+
+	tests := []struct {
+		desc      string
+		ingress   *v1beta1.Ingress
+		client    Client
+		result    []*tls.Configuration
+		errResult string
+	}{
+		{
+			desc:    "error_api_client",
```
timoreimann: It's okay to use human-readable descriptions: The Go test package automatically translates whitespaces and other non-"computer friendly" characters.

provider/kubernetes/kubernetes_test.go
```diff
@@ -2262,6 +2263,297 @@ func TestBasicAuthInTemplate(t *testing.T) {
...
+						CertFile: tls.FileOrContent("tls-crt"),
+						KeyFile:  tls.FileOrContent("tls-key"),
+					},
+					EntryPoints: []string{
+						"ep1.example.com",
+						"ep2.example.com",
+					},
+				},
+			},
+		},
```
timoreimann: Let's add another test case for the non-wildcard scenario. I know it's already covered by `TestTLSSecretLoad`, but in that test it could potentially be any other code path that leads to a valid TLS configuration we can compare against.

Including the dedicated test case in `TestGetTLSConfigurations` makes sure all code paths of the `getTLSConfigurations` function are covered by the owning test function while giving us the freedom to change `TestTLSSecretLoad`  in the future.

provider/kubernetes/kubernetes_test.go
```diff
@@ -2262,6 +2263,297 @@ func TestBasicAuthInTemplate(t *testing.T) {
...
+
+			if test.errResult == "" {
+				assert.Nil(t, err)
+			} else {
+				assert.EqualError(t, err, test.errResult)
+			}
+
+			if !reflect.DeepEqual(tlsConfigs, test.result) {
+				t.Errorf("Result should be %+v but was %+v", test.result, tlsConfigs)
+			}
```
timoreimann: I don't think we need to give any guarantees for `tlsConfigs` when we have a non-nil error. Consequently, I'd only validate the configuration during the happy path and modify the test accordingly; e.g.:

```go
if test.errResult != "" {
    assert.EqualError(t, err, test.errResult)
} else {
    assert.Nil(t, err)
    assert.Equal(t, tlsConfigs, test.result)
}
```

provider/kubernetes/kubernetes_test.go
```diff
@@ -2262,6 +2263,297 @@ func TestBasicAuthInTemplate(t *testing.T) {
...
+
+			tlsConfigs, err := getTLSConfigurations(test.ingress, test.client)
+
+			if test.errResult == "" {
+				assert.Nil(t, err)
+			} else {
+				assert.EqualError(t, err, test.errResult)
+			}
+
+			if !reflect.DeepEqual(tlsConfigs, test.result) {
```
timoreimann: Can we use `assert.Equal` instead to simplify?

docs/user-guide/kubernetes.md
```diff
@@ -330,6 +330,45 @@ echo "$(minikube ip) traefik-ui.minikube" | sudo tee -a /etc/hosts
...
+
+In addition to the modified ingress you need to provide the tls certificate via a kubernetes secret in the same namespace as the ingress. The following two commands will generate a new certificate and create a secret containing the key and cert files.
+
+```shell
+openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /tmp/tls.key -out /tmp/tls.crt -subj "/CN=traefik-ui.minikube"
+kubectl -n kube-system create secret tls traefik-ui-tls-cert --key=/tmp/tls.key --cert=/tmp/tls.crt
+```
+
+!!! note
+    The secret must have two entries named `tls.key`and `tls.crt`. See the [kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls) for more details.
```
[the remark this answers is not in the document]
gopenguin: I had no idea how to do something like this but I'm glad it is OK

provider/kubernetes/kubernetes.go
dtomcej: In my opinion, if an ingress is accessible via both http and https, and the https configuration fails, it should still be accessible via http.

However, if the TLS config fails on an https-only config, then yes, we should skip the ingress entirely.

gopenguin: Is there a way to get the endpoints to which the configuration will be attached to?

dtomcej: There are defaults, but the `traefik.frontend.entryPoints` annotation will override the defaults.

@timoreimann What do you think about adding another annotation to specify the TLS entrypoints to link to?
It feels super janky to me, since it seems to be like nailing a board with a screw. Ideally we would have access to the current entrypoint config to get the TLS-enabled entrypoints, and then use those, with the annotations overlaid if they exist.

It appears that the current k8s config does not access or reference the default entrypoints unless to override them with the annotation.

provider/kubernetes/kubernetes.go
```diff
@@ -432,6 +440,55 @@ func loadAuthCredentials(namespace, secretName string, k8sClient Client) ([]stri
...
+			// behavior.
+			for _, r := range ingress.Spec.Rules {
+				if r.Host != "" {
+					hosts = append(hosts, r.Host)
+				}
+			}
+		}
+
+		tlsConfig := &tls.Configuration{
+			EntryPoints: hosts,
```
dtomcej: Not sure this is correct. The entrypoints are defined in the static config.

templates/kubernetes.tmpl
```diff
@@ -37,3 +37,13 @@
     rule = "{{$route.Rule}}"
     {{end}}
 {{end}}
+
+{{range $tlsConfiguration := .TLSConfiguration}}
+[[tlsConfiguration]]
+  endPoints = [{{range $tlsConfiguration.EntryPoints}}
```
dtomcej: This is `entryPoints =`

provider/kubernetes/kubernetes.go
gopenguin: You are right. I've mixed up traefik's entrypoints with kubernetes ones, which are the domains on which the cert should be used. But as far as I can see there is no way to tell traefik to use a cert for only a set of domains.

gopenguin: The easiest way to addres this is probably to check if the provided cert matches the hostnames and isn't a wildcard

gopenguin:
I think, that I've missinterpreted the traefik api and there is currently no way for a provider to pass specific information about certificates like to which domain a wildcard cert is assigned to the core of traefik.

provider/kubernetes/kubernetes.go
```diff
@@ -432,6 +440,40 @@ func loadAuthCredentials(namespace, secretName string, k8sClient Client) ([]stri
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to fetch secret %s/%s: %s", ingress.Namespace, t.SecretName, err)
+		}
+		if !exists {
+			return nil, fmt.Errorf("secret %s/%s does not exist", ingress.Namespace, t.SecretName)
+		}
+
+		tlsCrtData, tlsCrtExists := tlsSecret.Data["tls.crt"]
+		tlsKeyData, tlsKeyExists := tlsSecret.Data["tls.key"]
+		if !tlsCrtExists {
```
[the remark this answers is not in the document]
gopenguin: That sounds reasonable. Would be `secret %s/%s must have two entries named 'tls.crt' and 'tls.key': both entries are missing` ok with you?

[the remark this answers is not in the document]
timoreimann: My suggestion would be `secret %s/%s is missing the following TLS data entries: %s", strings.Join(", ", missingEntries)`.

gopenguin:
@nmengin: Thanks for your review. I think your suggestion sounds reasonable.
I haven't found any way to access all the defined `EntryPoint`s from inside a provider. At least I belive, that the `GlobalConfiguration` wourd be the right struct to use but a can't get any references.
Any suggestions?

nmengin:
@gopenguin 

I just created a PR which allows managing the _by default_ behavior (#2534).
Thanks to it you have not to look for the _default TLS entryPoints_.

If the service does not define the `traefik.frontend.entryPoints`, you can set `nil`.

I hope that can help you.

gopenguin:
Thanks. Thats now much simpler

gopenguin:
I found out, that I removed the section in the template passing the certs. I'll add it back again and add a test to cover this.

timoreimann:
@nmengin have you been able to complete your tests? I'd like to make sure the entry point part is approved before I take another look at the PR.

nmengin:
@timoreimann No, I did not.

I've got few problems with my minikube instance :cry: 
I'll try to test it next week.

docs/user-guide/kubernetes.md
```diff
@@ -333,6 +333,45 @@ echo "$(minikube ip) traefik-ui.minikube" | sudo tee -a /etc/hosts
...
+In addition to the modified ingress you need to provide the TLS certificate via a kubernetes secret in the same namespace as the ingress. The following two commands will generate a new certificate and create a secret containing the key and cert files.
+
+```shell
+openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /tmp/tls.key -out /tmp/tls.crt -subj "/CN=traefik-ui.minikube"
+kubectl -n kube-system create secret tls traefik-ui-tls-cert --key=/tmp/tls.key --cert=/tmp/tls.crt
+```
+
+!!! note
+    The secret must have two entries named `tls.key`and `tls.crt`. See the [kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls) for more details.
+
```
nmengin: Can you please add a note to explain that the certificates contained into the K8s secrets will be added to all the TLS entryPoints linked to the services which use it (if the label `traefik.frontend.entryPoints` is defined) or to all the TLS entryPoints defined into `DefaultEntryPoints` by default ?

gopenguin: Would the following ok with you?

The TLS certificates will be added to all entrypoints defined by the ingress annotation `traefik.frontend.entryPoints`. If no such annotation is provided, the TLS certificates will be added     to all TLS enabled `defaultEntryPoints`.

Did you mean with 'note' the note formatting or just an addition to the documentation?

nmengin: @gopenguin 

Yes the sentence LGTM :+1: 
I mean note formatting at the end of the block instructions :wink:

docs/user-guide/kubernetes.md
```diff
@@ -333,6 +333,48 @@ echo "$(minikube ip) traefik-ui.minikube" | sudo tee -a /etc/hosts
...
+spec:
+  rules:
+  - host: traefik-ui.minikube
+    http:
+      paths:
+      - backend:
+          serviceName: traefik-web-ui
+          servicePort: 80
+  tls:
+  - hosts:
```
timoreimann: Hmm, I don't see us parsing the hosts list in our code explicitly. Am I missing something?

docs/user-guide/kubernetes.md
```diff
@@ -333,6 +333,48 @@ echo "$(minikube ip) traefik-ui.minikube" | sudo tee -a /etc/hosts
...
+    - traefik-ui.minikube
+    secretName: traefik-ui-tls-cert
+```
+
+In addition to the modified ingress you need to provide the TLS certificate via a kubernetes secret in the same namespace as the ingress. The following two commands will generate a new certificate and create a secret containing the key and cert files.
+
+```shell
+openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /tmp/tls.key -out /tmp/tls.crt -subj "/CN=traefik-ui.minikube"
+kubectl -n kube-system create secret tls traefik-ui-tls-cert --key=/tmp/tls.key --cert=/tmp/tls.crt
```
timoreimann: I'd drop the `/tmp` directory part as it may not exist on every system. Saving the key and cert in the current working directory seems good enough for me.

docs/user-guide/kubernetes.md
```diff
@@ -333,6 +333,48 @@ echo "$(minikube ip) traefik-ui.minikube" | sudo tee -a /etc/hosts
 
 We should now be able to visit [traefik-ui.minikube](http://traefik-ui.minikube) in the browser and view the Træfik Web UI.
 
+### Add a TLS Certificate to the Ingress
+
+!!! note
+    For this example to work you need a TLS entrypoint. You don't have to provide a TLS certificate at this point. For more details see [here](/configuration/entrypoints/).
+
+To setup a HTTPS protected ingress, you can leverage the TLS feature of the ingress resource.
```
timoreimann: s/a HTTPS protected/an HTTPS-protected/

docs/user-guide/kubernetes.md
```diff
@@ -333,6 +333,48 @@ echo "$(minikube ip) traefik-ui.minikube" | sudo tee -a /etc/hosts
...
+```shell
+openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout /tmp/tls.key -out /tmp/tls.crt -subj "/CN=traefik-ui.minikube"
+kubectl -n kube-system create secret tls traefik-ui-tls-cert --key=/tmp/tls.key --cert=/tmp/tls.crt
+```
+
+!!! note
+    The secret must have two entries named `tls.key`and `tls.crt`. See the [kubernetes documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls) for more details.
+
+!!! note
+    The TLS certificates will be added to all entrypoints defined by the ingress annotation `traefik.frontend.entryPoints`. If no such annotation is provided, the TLS certificates will be added to all TLS enabled `defaultEntryPoints`.
```
timoreimann: s/to all TLS enabled/to all TLS-enabled/

provider/kubernetes/builder_ingress_test.go
```diff
@@ -92,6 +92,23 @@ func iBackend(name string, port intstr.IntOrString) func(*v1beta1.HTTPIngressPat
 	}
 }
 
+func iTLSs(opts ...func(*v1beta1.IngressTLS)) func(*v1beta1.Ingress) {
```
timoreimann: Can we rename this to `iTLSes`? Makes it slightly more readable (and grammatically correct) IMHO.

provider/kubernetes/kubernetes.go
```diff
@@ -354,6 +355,13 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 				}
 			}
 		}
+
+		tlsConfigs, err := getTLSConfigurations(i, k8sClient)
+		if err != nil {
+			log.Errorf("Error configuring TLS for ingress %s/%s: %v", i.Namespace, i.Name, err)
```
timoreimann: Does this express the intended behavior as summarized by @dtomcej [here](https://github.com/containous/traefik/pull/2439#discussion_r153215631)?

provider/kubernetes/kubernetes.go
```diff
@@ -441,6 +449,48 @@ func loadAuthCredentials(namespace, secretName string, k8sClient Client) ([]stri
...
+		var missingEntries []string
+		if !tlsCrtExists {
+			missingEntries = append(missingEntries, "tls.crt")
+		}
+		if !tlsKeyExists {
+			missingEntries = append(missingEntries, "tls.key")
+		}
+		if len(missingEntries) > 0 {
+			return nil, fmt.Errorf("secret %s/%s is missing the following TLS data entries: %s", ingress.Namespace, t.SecretName, strings.Join(missingEntries, ", "))
+		}
```
timoreimann: I like this a lot 👍

provider/kubernetes/kubernetes.go
gopenguin: I couldn't figure out, if an entry point is TLS-enabled at this point as the default configuration (passing nil) is handled further down the road and override entry points are only available as strings without any context. Therefore, I can't check if an entry point is TLS-enabled or not.

But I can't imagine a way to solve this problem properly as either the internals need to know about a TLS configuration error and abort in the case of an HTTPS-only entry point or the provider needs to know about the entry points (default and override).

I'm sorry for not pointing this out earlier.

timoreimann: Thanks for doing the research.

If we can't figure out whether it's safe to accept an illegal definition or not, then my opinion is that we should err on the side of safety and skip it. Better be safe than sorry.

What do you (and @dtomcej) think?

gopenguin: Staying on the safe side should be our first choice as this is security critical. I'd also add a note to the docs regarding this and also the deviation from the kubernetes spec (host names are ignored and therefore the wildcard behaviour cannot be implemented).

timoreimann: Could you refresh my mind again: what's the reason we cannot comply with the Kubernetes spec? Is there a potential path (possibly in the future) we could improve on this point?

gopenguin: In a kubernetes ingress resource multiple TLSes per ingress are allowed. Each consists of two parts: a list of host names and a reference to a kubernetes secret. There are two cases in which a TLS configuration can be:
1. A list of host names is provided: Only the defined host names in the list are used for this TLS certificate, not the ones defined in the certificate
2. The list of host names is empty (wildcard behavior): In this case not a general wildcard is used. Instead, only the host names in the ingress are used as the wildcard. A wildcard TLS may only be used without any other TLS definitions in this ingress.

I think there is a path to implement this feature in the future as this behavior is a special case of the already implemented one in the traefik internals. (https://github.com/containous/traefik/blob/c446c291d9dc49895e69c91e51a054778c71f21a/tls/certificate.go#L150-L154 would be the default behavior, but it can be overwritten by the provider)
I have no clue what so ever how to pass this information from a provider to the internals.

dtomcej: Honestly, after doing a bit of research, there does not seem to be any implementation outside of ingress controller annotations to handle the failures of any of the components of an ingress. It appears that it is all-or-nothing, where the tls section of the spec is parsed in parallel with the rules section of the spec, leaving the annotations to handle the relationship between http/rules and tls sections.

Therefore, I would now be more inclined to just fail/skip loading the entire ingress on an error loading any portion of the tls section of the spec, since there is no way (without going into all the middlewares) to know how important the separate tls sections are.

timoreimann: @gopenguin thanks for the clarification with regards to the wildcard matter. As long as we clearly document the constraints, I'm fine with it.

In the light of @dtomcej's latest comment: could you please change the logic so that we skip illegal Ingress definitions? AFAICS, that is the last point that keeps us from merging this PR.

provider/kubernetes/kubernetes.go
```diff
@@ -174,6 +175,13 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 			continue
 		}
 
+		tlsConfigs, err := getTLSConfigurations(i, k8sClient)
+		if err != nil {
+			log.Errorf("Error configuring TLS for ingress %s/%s: %v", i.Namespace, i.Name, err)
```
timoreimann: Hmm, my IDE is telling me that this line is supposedly missing code coverage. Are we possibly missing a test to validate that we actually skip the Ingress if `getTLSConfigurations` returns an error?

gopenguin: Oh, you are right. This case isn't covered.

timoreimann:
@gopenguin our bot cannot proceed due to a merge conflict. Would you mind resolving it manually?

ldez:
@timoreimann Not due to conflicts but to the missing milestone.

gopenguin:
It was a pleasure to contribute to this awesome project. Thanks for your patience and till next time.

PR #2442, opened 2017-11-21
Fix high memory usage in retry middleware for big files

### LINKED ISSUE #1064: Traefik bad perf for routing big data files
Version: v1.1.2
Codename: camembert
Go version: go1.7.4
Built: 2016-12-15_10:21:15AM
OS/Arch: linux/amd64

```
/traefik --web --retry --retry.attempts=10 --docker --docker.watch --docker.domain=my.domain.org --logLevel=DEBUG --entryPoints=Name:http Address::80 --defaultEntryPoints=http --web.address=:8081
```

### What did you do?
I start an nginx container, with some static files in web root
I export the port, and test :
- Download a 1.3GB file on the expose port (without traefik)
- Download same file by traefik on configured frontend

Almost same downlaod time

### What did you see instead?
I download directly in 2 sec and i got 6 sec through traefik
The traefik process use 50/60% CPU for 3 sec
At the end, the traefik process take 4.626g Virtual RAM (top) instead of classical ~100M
That RAM is released few minutes later (garbage collect ?)

I can't use traefik to route those requests, i have some big files, like 20/30 GB
How traefik works for that ?
Why it have to store those data in RAM ?

Thanx for reading

iahmedov:
### What does this PR do?
Decreases memory usage in retry middleware when served big files

### Motivation
Fixes #1064 

### More

- [x] Added/updated tests

### Additional Notes
Old logic downloaded and buffered all the data from upstream server then send to downstream. This causes high memory usage when downloading >500Mb files. In order to avoid memory usage, this pull request proposes file based buffering, where each requested data will be buffered in memory until it reaches `CacheMaxCapacity` (new configuration option for retry), after that it starts buffering data into temporary file.

Small benchmarks shows that depending on load/storage on traefik hosted server, current retry logic under performs 1.3-2.5x times when retry logic not used.

### DIFF
--- a/cmd/traefik/configuration.go
+++ b/cmd/traefik/configuration.go
@@ -159,6 +159,13 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 	defaultDynamoDB.TableName = "traefik"
 	defaultDynamoDB.Watch = true
 
+	var defaultRety configuration.Retry
+	defaultRety.Attempts = 0
+	defaultRety.CacheInitialCapacity = 4 * 1024
+	defaultRety.CacheMaxCapacity = 32 * 1024
+	defaultRety.TempDir = "/tmp"
+	defaultRety.RetryInterval = 0
+
 	// default Eureka
 	var defaultEureka eureka.Provider
 	defaultEureka.Delay = "30s"
@@ -246,7 +253,7 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 		Rancher:            &defaultRancher,
 		Eureka:             &defaultEureka,
 		DynamoDB:           &defaultDynamoDB,
-		Retry:              &configuration.Retry{},
+		Retry:              &defaultRety,
 		HealthCheck:        &healthCheck,
 		RespondingTimeouts: &respondingTimeouts,
 		ForwardingTimeouts: &forwardingTimeouts,
--- a/cmd/traefik/configuration.go
+++ b/cmd/traefik/configuration.go
@@ -162,7 +162,7 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 	var defaultRety configuration.Retry
 	defaultRety.Attempts = 0
 	defaultRety.CacheInitialCapacity = 4 * 1024
-	defaultRety.CacheMaxCapacity = 32 * 1024
+	defaultRety.CacheMaxCapacity = 0
 	defaultRety.TempDir = "/tmp"
 	defaultRety.RetryInterval = 0
 
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -437,7 +437,11 @@ type Redirect struct {
 
 // Retry contains request retry config
 type Retry struct {
-	Attempts int `description:"Number of attempts" export:"true"`
+	Attempts             int    `description:"Number of attempts" export:"true"`
+	CacheInitialCapacity int    `description:"Initial capacity of cache when creating (in bytes)" export:"true"`
+	CacheMaxCapacity     int    `description:"Maximum allowed capacity for cache (in bytes) when upstream data is bigger than this value, retry starts caching data into file" export:"true"`
+	TempDir              string `description:"Place to store temporary files" export:"true"`
+	RetryInterval        int64  `description:"Retry interval in milliseconds before next try" export:"true"`
 }
 
 // HealthCheckConfig contains health check configuration parameters.
--- a/docs/configuration/commons.md
+++ b/docs/configuration/commons.md
@@ -283,12 +283,42 @@ For dynamic providers, the corresponding template file needs to be customized ac
 # Enable retry sending request if network error
 [retry]
 
+# Retry logic uses caching mechanism in order to download data fully from upstream server and supplies it to downstream when it receives all data
+
 # Number of attempts
 #
 # Optional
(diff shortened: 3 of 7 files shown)


timoreimann:
Thanks for the PR -- I think this issue has been troublesome for quite some folks.

One thing I'm wondering though if file-based buffering may lead to different kind of problems. For instance, I can imagine that this might lead to problems related to running out of disk space or quota, especially inside containers.

I wonder if there's a way to stream the response straight to the client so that buffering will happen in the network. Any thoughts on that one?

iahmedov:
> One thing I'm wondering though if file-based buffering may lead to different kind of problems. For instance, I can imagine that this might lead to problems related to running out of disk space or quota, especially inside containers.

Sure, this can happen if concurrently many big files are downloaded, each of them will be written into storage and deleted after file has been sent to downstream. 
One way of solving this is, mounting big NFS storage.
Another way is increasing retry.CacheMaxCapacity based on usage statistics. In that case most of the smaller files will be streamed from memory. But still both of them will not solve the issue.

> I wonder if there's a way to stream the response straight to the client so that buffering will happen in the network. Any thoughts on that one?

I have tried to understand logic behind retry middleware when its first written, in my understanding retry middleware sends data to downstream only if data is fully fetched from upstream. If data is streamed straight from upstream without buffering and in the mean time upstream failes, proxy will deliver half of the data to downstream and retry middleware starts another try to download data again, which maybe not same all the time and again sends this data to downstream.
There could be a way to continue downloading if there are some prerequisites like
- requests are always GET/HEAD
- data will not change while retry issued (most of the time files, Header or first bytes of Body can be analyzed)

Any improvements are welcome

timoreimann:
Traefik will only retry requests in case of lower layered connection (setup) errors. Once the TCP connection has been established, no retry should occur anymore. So to my understanding, a stream would never require resuming.

timoreimann:
I have to double check my last statement, but that's how I remember things to be.

marco-jantke:
@iahmedov thanks for the PR and your effort so far!

@timoreimann I can confirm that your memory was right :) The retry middleware only retries on net errors and when it receives an unexpected `io.EOF`, what we treat like a net error.

I also think that we should strive, at least for the beginning, for a solution that only buffers the file in memory. @iahmedov do you know if without the retry middleware file streaming would work? If this is the case I guess we have to change the mechanics [here](https://github.com/containous/traefik/blob/master/middlewares/retry.go#L65) that it won't read everything at once. WDYT?

iahmedov:
I think retry middleware build upon based something on this: https://github.com/containous/traefik/issues/1064#issuecomment-285724139, 

> I think you assume that the endpoint will always return the exact same response, which is not the case in general. Appending (part of) the result of the second response to the first part of the first response does not make much sense.

but I fully agree with the comment next to it: https://github.com/containous/traefik/issues/1064#issuecomment-285739401

> We should probably follow HAProxy's and Nginx's examples and not try to be too smart: If data is already flowing and a failure occurs, don't retry but leave it up to the client to deal with the situation.

As an additional source we can follow retry logic from Lyft's Envoy proxy (https://www.envoyproxy.io/docs/envoy/latest/configuration/http_filters/router_filter.html#x-envoy-retry-on)
Basic assumptions:
- Retry if could not connect
- Retry if 5xx error
- Retry if request is idempotent to server state (but no way to know it beforehand, thats why just use 409 error)

dtomcej:
I would echo the sentiment that the file system should only be touched if configured to. I think that by default everything should be in memory. Faster, and easier to manage. Also, docker containers with layered filesystems don't have the same i/o. I could see having buffer to file as an option, but would want to use the in-memory option as the default.

iahmedov:
@dtomcej sorry for late response

>  I think that by default everything should be in memory.

Configuration changed slightly in order to use only memory by default. When `cachemaxcapacity` set to 0, data will be cached in memory without touching file system. if set to value bigger than 0, file based caching will work

timoreimann:
@emilevauge can you help move forward from the design review to the review phase? Thanks.

timoreimann:
Does anyone know if other proxies create buffer files too for large-ish downloads?

My (maybe naive) expectation would be that all buffers live in-memory, and TCP flow control makes sure that overload situations are handled through throttling. Hopefully, Go's http package can do these things under the hood.

WDYT?

timoreimann:
To clarify a bit further: I'm okay with using files for buffering purposes if that's what we need. What I'm wondering is if there's a yet simpler approach that gives us what we need without having to resort to file I/O.

marco-jantke:
I am quite confident that I found a solution for the retry middleware that just passes the data through and doesn't need a temporary buffer like before. This should completely avoid the high RAM usage.

As this is a quite complicated part and was affected by a lot of edge cases in the part, I am currently still testing manually and will deploy this to our staging system next week and see whether it works as expected. Will update here asap.

marco-jantke:
FTR my tests look good so far but there was not a lot of traffic running over the patched Traefik version, yet. Waiting another week as we will increase traffic step-wise and I will have more insights and confidence at some time in the middle of next week.

marco-jantke:
@iahmedov I created a PR that reduces the amount of RAM usage in the retry middleware, without any caching mechanism or configuration options. Can you check it out to see whether you miss something for your setup? https://github.com/containous/traefik/pull/2740

iahmedov:
can someone please close this PR, since #2740 is already merged and it fixes the retry problem

PR #2443, opened 2017-11-21
Improve logging output for failing healthchecks

marco-jantke:
It's useful to understand why health checks failed in the case they do. Therefore I improved the logging output of the health checks, so that you can see whether it was e.g. an http/net issue or a wrong response code.

I also use now the backend name only in the logs. Before, for some logs something like `entryPointName+backendName` was used, but in most places we seem to log the backend name only.


marco-jantke:
@mmatur I rebased the changes now and resolved the conflicts. Please write a comment next time when you add this label, you don't get notified for label changes and so I wasn't seeing this now for over a week but was wondering why nothing happens with this PR.

PR #2446, opened 2017-11-22
Fix typo in examples

dahefanteng:
### What does this PR do?

Just fix typo issue,try to avoid mislead others.

### DIFF
--- a/examples/etcd-config.sh
+++ b/examples/etcd-config.sh
@@ -64,7 +64,7 @@ function insert_etcd3_data() {
 
 function show_usage() {
         echo "USAGE : etcd-config.sh ETCD_API_VERSION [ETCD_IP_ADDRESS]"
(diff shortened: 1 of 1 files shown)


examples/etcd-config.sh
```diff
@@ -64,7 +64,7 @@ function insert_etcd3_data() {
 
 function show_usage() {
         echo "USAGE : etcd-config.sh ETCD_API_VERSION [ETCD_IP_ADDRESS]"
-        echo "        ETCD_API_VERSION : Values v2 or V3 (v3 requires ETCD_IP_ADDRESS)"
+        echo "        ETCD_API_VERSION : Values V2 or V3 (v3 requires ETCD_IP_ADDRESS)"
```
mmatur: I think that you could also replace `v3 requires ETCD_IP_ADDRESS` by `V3 requires ETCD_IP_ADDRESS`

dahefanteng: Right,I have replaced that.

Issue #2369, opened 2017-11-07
Collecting stats


emilevauge:
Following https://github.com/containous/traefik/pull/2172, this is a proposal to discuss with the community on how we can get more information about deployed Traefik instances.

## Why do the developers team need more info ?
As you may know, the Traefik core development team is quite small and as with a lot of open source projects, we lack time and resources. As a consequence, we have to carefully choose which tasks and features need our attention. As a result, we usually invest our time on features needed or requested by most of the community. In order to effeciently do this, we have to know how our community uses Traefik.
So far we have been using feedback from our users on Slack and Github but we definitely need more details on usage.

To give an illustrative example, we have no way to know which configuration backend is the most used or which configuration backend is used by the least number of people. What if we discover that we maintain a configuration backend that is largely unused? Knowing this, we could have allocated our resource on something more useful, especially since we have [**a lot**](https://github.com/containous/traefik/issues) of useful things we can work on ;)

Another example is that we have no idea of release adoption/implementation. Having this knowledge  would help us to adapt our development cycle to benefit adoption. We don't need or want to release every month if users are waiting for 2 months before updating to the latest release.

_We just need to know what is used, and what is not._

## What we propose
Ideally, we would like statistics on the toml/flags Traefik configuration and Traefik versions our users are using. The toml/flags configuration would allow the development team to know what is used in Traefik and what is not.

_Only export what's needed._

We already use a mechanism to export the whole configuration when using the [`traefik bug`](http://docs.traefik.io/basics/#command-bug) command. But it only exports what's required for bug diagnostics. It only export specifically tagged configuration fields. Furthermore, all the private data (IPs, email address, etc) are not exported as they are not tagged in the code (with struct tags).

We could reuse this in this stats collection mechanism.

What's great with this solution is that exported configuration fields are hard-coded. Each time a new field will be added in the configuration, by default, it will not be exported. We will need to tag it in the code to export it. This allows us to carefully review what's being exported and what's not in future configuration changes and this can be reviewed by the community before implementation.

_Collected configuration fields are hard-coded._

## Opt-in vs. Opt-out    

Another topic we need to discuss is do we make it opt-in or opt-out?

The easiest way would be to set it opt-in: if you want to export your config, you need to enable it in your configuration.

[...]

PR #2447, opened 2017-11-22
Stats collection.

ldez:
### What does this PR do?

https://github.com/containous/traefik/issues/2369

**This feature is disabled by default.**

In the future, as discussed in the [proposal](https://github.com/containous/traefik/issues/2369), this may be enabled by default.

### Motivation

https://github.com/containous/traefik/issues/2369

### More

- [x] Added/updated tests

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -247,11 +247,9 @@ func configureLogging(globalConfiguration *configuration.GlobalConfiguration) {
 }
 
 func checkNewVersion() {
-	ticker := time.NewTicker(24 * time.Hour)
+	ticker := time.Tick(24 * time.Hour)
 	safe.Go(func() {
-		time.Sleep(10 * time.Minute)
-		version.CheckNewVersion()
-		for range ticker.C {
+		for time.Sleep(10 * time.Minute); ; <-ticker {
 			version.CheckNewVersion()
 		}
 	})
@@ -261,7 +259,7 @@ func stats(globalConfiguration *configuration.GlobalConfiguration) {
 	if globalConfiguration.SendAnonymousUsage {
 		log.Info(`
 Stats collection is enabled.
-Many thanks to contribute to Traefik improvement by allowing us to receive anonymous information from your configuration.
+Many thanks for contributing to Traefik's improvement by allowing us to receive anonymous information from your configuration.
 Help us improve Traefik by leaving this feature on :)
 More details on: https://docs.traefik.io/basic/#collected-data
 `)
@@ -276,16 +274,10 @@ More details on: https://docs.traefik.io/basic/#collected-data
 }
 
 func collect(globalConfiguration *configuration.GlobalConfiguration) {
-	ticker := time.NewTicker(24 * time.Hour)
+	ticker := time.Tick(24 * time.Hour)
 	safe.Go(func() {
-		time.Sleep(10 * time.Minute)
-		err := collector.Collect(globalConfiguration)
-		if err != nil {
-			log.Debug(err)
-		}
-		for range ticker.C {
-			err := collector.Collect(globalConfiguration)
-			if err != nil {
+		for time.Sleep(10 * time.Minute); ; <-ticker {
+			if err := collector.Collect(globalConfiguration); err != nil {
 				log.Debug(err)
 			}
 		}
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -15,6 +15,7 @@ import (
 	"github.com/containous/flaeg"
 	"github.com/containous/staert"
 	"github.com/containous/traefik/acme"
+	"github.com/containous/traefik/collector"
 	"github.com/containous/traefik/configuration"
 	"github.com/containous/traefik/job"
 	"github.com/containous/traefik/log"
@@ -149,6 +150,8 @@ func run(globalConfiguration *configuration.GlobalConfiguration, configFile stri
 		checkNewVersion()
 	}
 
+	stats(globalConfiguration)
+
 	log.Debugf("Global configuration loaded %s", string(jsonConf))
 	svr := server.NewServer(*globalConfiguration)
 	svr.Start()
@@ -248,10 +251,42 @@ func checkNewVersion() {
 	safe.Go(func() {
 		time.Sleep(10 * time.Minute)
 		version.CheckNewVersion()
-		for {
-			select {
-			case <-ticker.C:
-				version.CheckNewVersion()
+		for range ticker.C {
+			version.CheckNewVersion()
+		}
+	})
+}
+
+func stats(globalConfiguration *configuration.GlobalConfiguration) {
+	if globalConfiguration.SendAnonymousUsage {
+		log.Info(`
+Stats collection is enabled.
+Many thanks to contribute to Traefik improvement by allowing us to receive anonymous information from your configuration.
+Help us improve Traefik by leaving this feature on :)
+More details on: https://docs.traefik.io/basic/#collected-data
+`)
+		collect(globalConfiguration)
+	} else {
+		log.Info(`
+Stats collection is disabled.
+Help us improve Traefik by turning this feature on :)
+More details on: https://docs.traefik.io/basic/#collected-data
+`)
+	}
+}
+
+func collect(globalConfiguration *configuration.GlobalConfiguration) {
+	ticker := time.NewTicker(24 * time.Hour)
+	safe.Go(func() {
+		time.Sleep(10 * time.Minute)
+		err := collector.Collect(globalConfiguration)
+		if err != nil {
+			log.Debug(err)
+		}
+		for range ticker.C {
+			err := collector.Collect(globalConfiguration)
+			if err != nil {
+				log.Debug(err)
 			}
 		}
 	})
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -1,0 +1,78 @@
+package collector
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"net"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/containous/traefik/cmd/traefik/anonymize"
+	"github.com/containous/traefik/configuration"
+	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/version"
+	"github.com/mitchellh/hashstructure"
+)
+
+const containousURL = "https://collect.traefik.io/619df80498b60f985d766ce62f912b7c"
+
+// Collected data
+type data struct {
+	Version       string
+	Codename      string
+	BuildDate     string
+	Configuration string
+	Hash          string
+}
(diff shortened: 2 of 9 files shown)


cmd/traefik/traefik.go
```diff
@@ -248,3 +254,41 @@ func checkNewVersion() {
...
+Help us improve Traefik by turning this feature on :)
+More details on: https://docs.traefik.io/basic/#collected-data
+`)
+	}
+}
+
+func collect(globalConfiguration *configuration.GlobalConfiguration) {
+	ticker := time.NewTicker(24 * time.Hour)
+	safe.Go(func() {
+		time.Sleep(10 * time.Minute)
```
vdemeester: I'm not sure understand why this first call of `Collect` is required 👼 I guess you want the first one after 10min and then each and every day, right.

collector/collector.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"net"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/containous/traefik/cmd/traefik/anonymize"
```
vdemeester: 🤔 why `anonymize` is under `cmd/traefik` package ?

vdemeester: (or why `collector` is in a different parent package that `anonimize` ?)

ldez: Because before this PR it was use only by `traefik bug` command.

I can move the package `anonymize` at the root of the project.

cmd/traefik/traefik.go
ldez: exactly, we wait 10 min for eliminate some quick restart.

collector/collector.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/containous/traefik/configuration"
+	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/version"
+	"github.com/mitchellh/hashstructure"
+)
+
+const containousURL = "https://collect.traefik.io"
```
dtomcej: This is probably fine, but we might want this overridable for testing, and to allow people to post the stats to their own endpoint if they want. This would also improve public perception, as they can see EXACTLY what is being sent by posting to their own endpoint.

collector/collector.go
```diff
@@ -0,0 +1,77 @@
...
+		BuildDate:     version.BuildDate,
+		Hash:          strconv.FormatUint(hashConf, 10),
+		Configuration: base64.StdEncoding.EncodeToString([]byte(anonConfig)),
+		EmitDate:      time.Now().Format(time.RFC3339),
+	}
+
+	buf := new(bytes.Buffer)
+	json.NewEncoder(buf).Encode(data)
+
+	_, err = makeHTTPClient().Post(containousURL, "application/json; charset=utf-8", buf)
```
dtomcej: This might be a dumb question, but do we want this to retry/or log the data on failure? rather than just the error?

docs/basics.md
```diff
@@ -609,3 +609,122 @@ traefik healthcheck
...
+# Default: false
+#
+sendAnonymousUsage = true
+```
+
+- adding this flag in the CLI:
+
+```bash
+./traefik --sendAnonymousUsage=true
+```
```
dtomcej: Might want to note that in the future, that this will be enabled by default.

collector/collector.go
ldez: the error is logged here: https://github.com/containous/traefik/pull/2447/files#diff-086f37ac2cd2ef1d326af9766ce3d7b6R282

ldez: It's not critical for the application, we think the retry/log response is not needed in this case.

collector/collector.go
```diff
@@ -0,0 +1,77 @@
...
+		return err
+	}
+
+	data := &data{
+		Version:       version.Version,
+		Codename:      version.Codename,
+		BuildDate:     version.BuildDate,
+		Hash:          strconv.FormatUint(hashConf, 10),
+		Configuration: base64.StdEncoding.EncodeToString([]byte(anonConfig)),
+		EmitDate:      time.Now().Format(time.RFC3339),
```
emilevauge: Is-it needed ?

collector/collector.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	buf := new(bytes.Buffer)
+	json.NewEncoder(buf).Encode(data)
```
emilevauge: error is shadowed here

docs/basics.md
```diff
@@ -609,3 +609,122 @@ traefik healthcheck
...
+
+In order to help us learn more about how Træfik is being used and improve it, we collect anonymous usage statistics from running instances.
+Those data help us prioritize our developments and focus on what's more important (for example, which configuration backend is used and which is not used).
+
+### What ?
+
+Once a day, we collect:
+- the Træfik version
+- a hash of the configuration
+- an **obfuscated and anonymous version** of the static configuration:
```
emilevauge: s/obfuscated and anonymous version/anonymous version

docs/basics.md
```diff
@@ -609,3 +609,122 @@ traefik healthcheck
...
+  SecretAccessKey = "xxxx"
+```
+
+### Show me the code !
+
+If you want to dig into more details, here is the source code of the collecting system: [collector.go](https://github.com/containous/traefik/blob/master/collector/collector.go)
+
+By default we anonymize all configuration fields, except fields tagged with `export=true`.
+
+You can check all fields in the [godoc](https://godoc.org/github.com/containous/traefik).
```
emilevauge: Maybe we should give a this link instead https://godoc.org/github.com/containous/traefik/configuration#GlobalConfiguration

docs/basics.md
```diff
@@ -610,3 +610,122 @@ traefik healthcheck
...
+
+### What ?
+
+Once a day, we collect:
+- the Træfik version
+- a hash of the configuration
+- an **anonymous version** of the static configuration:
+    - token, user name, password, URL, IP, domain, email, etc, are removed
+
+!!! warning
```
marco-jantke: I think this is rather another note than a warning.

cmd/traefik/traefik.go
```diff
@@ -256,3 +259,41 @@ func checkNewVersion() {
...
+		err := collector.Collect(globalConfiguration)
+		if err != nil {
+			log.Debug(err)
+		}
+		for {
+			select {
```
marco-jantke: You don't really need the `select` case here, because `<-ticker.C` will block anyway.

cmd/traefik/traefik.go
```diff
@@ -256,3 +259,41 @@ func checkNewVersion() {
 		}
 	})
 }
+
+func stats(globalConfiguration *configuration.GlobalConfiguration) {
+	if globalConfiguration.SendAnonymousUsage {
+		log.Info(`
+Stats collection is enabled.
+Many thanks to contribute to Traefik improvement by allowing us to receive anonymous information from your configuration.
```
timoreimann: I think it must say (**emphasis** by me)

_Many thanks **for contributing** to **Traefik's** improvement [...]_

cmd/traefik/traefik.go
```diff
@@ -256,3 +259,41 @@ func checkNewVersion() {
...
+		log.Info(`
+Stats collection is disabled.
+Help us improve Traefik by turning this feature on :)
+More details on: https://docs.traefik.io/basic/#collected-data
+`)
+	}
+}
+
+func collect(globalConfiguration *configuration.GlobalConfiguration) {
+	ticker := time.NewTicker(24 * time.Hour)
```
timoreimann: I'd make it explicit that we want a never-ending ticker (we never call `ticker.Stop()`) by invoking `time.Tick()` here.

cmd/traefik/traefik.go
```diff
@@ -256,3 +259,41 @@ func checkNewVersion() {
...
+		}
+		for {
+			select {
+			case <-ticker.C:
+				err := collector.Collect(globalConfiguration)
+				if err != nil {
+					log.Debug(err)
+				}
+			}
+		}
```
timoreimann: It is possible to have just one call to `Collect` (slightly adopted from [here](https://www.reddit.com/r/golang/comments/5g0hor/ticker_with_instant_start/daov1qr/)):

```go
for time.Sleep(10 * time.Minute); ; <-ticker.C {
  if err := collector.Collect(globalConfiguration); err != nil {
    log.Debug(err)
  }
}
```

([Simplified Go Playground example](https://play.golang.org/p/xIhkaHd-Is).)

collector/collector.go
```diff
@@ -0,0 +1,78 @@
...
+	"time"
+
+	"github.com/containous/traefik/cmd/traefik/anonymize"
+	"github.com/containous/traefik/configuration"
+	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/version"
+	"github.com/mitchellh/hashstructure"
+)
+
+const containousURL = "https://collect.traefik.io/619df80498b60f985d766ce62f912b7c"
```
timoreimann: Can we have a more expressive name to indicate what's happening behind this URL (like `collectorURL`), possibly even a GoDoc?

collector/collector.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// Collect anonymous data.
+func Collect(globalConfiguration *configuration.GlobalConfiguration) error {
+	anonConfig, err := anonymize.Do(globalConfiguration, false)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Stats sent on %s: %s", containousURL, anonConfig)
```
timoreimann: s/on/to/

timoreimann: I'd write _**Anonymous** stats sent to [...]_.

(**Emphasis** by me.)

cmd/traefik/traefik.go
timoreimann: Can we document the delay in `basics.md` as well?

cmd/traefik/traefik.go
```diff
@@ -248,10 +251,42 @@ func checkNewVersion() {
...
+	safe.Go(func() {
+		time.Sleep(10 * time.Minute)
+		err := collector.Collect(globalConfiguration)
+		if err != nil {
+			log.Debug(err)
+		}
+		for range ticker.C {
+			err := collector.Collect(globalConfiguration)
+			if err != nil {
+				log.Debug(err)
```
timoreimann: It is possible to have just one call to `Collect` (slightly adopted from [here](https://www.reddit.com/r/golang/comments/5g0hor/ticker_with_instant_start/daov1qr/)):

```go
for time.Sleep(10 * time.Minute); ; <-ticker.C {
  if err := collector.Collect(globalConfiguration); err != nil {
    log.Debug(err)
  }
}
```

([Simplified Go Playground example](https://play.golang.org/p/xIhkaHd-Is).)

ldez:
@timoreimann I have take account of all your remarks.

PR #2452, opened 2017-11-22
add entrypoint in prometheus doc and remove web on influxdb doc

Juliens:
### What does this PR do?
Fix some mistake in the documentation after #2335 

### Motivation
Good documentation


docs/configuration/metrics.md
```diff
@@ -9,6 +9,12 @@
 
   # To enable Traefik to export internal metrics to Prometheus
   [metrics.prometheus]
+    # Name of the related entry point
```
mmatur: Could you please add a new line before ?

PR #2453, opened 2017-11-22
use consistent namespace kube-system to avoid confusing new users

lvnilesh:
use consistent namespace kube-system to avoid confusing new users


lvnilesh:
sorry, your example is correct.

PR #2461, opened 2017-11-24
Fix kubernetes path prefix rule with rewrite-target

cheungpat:
When rewrite-target is specified, any existing PathPrefix rule is
removed. Hence an ingress with rewrite-target is always matched
for the same host regardless of path prefix.


timoreimann:
I think this makes sense.

@dtomcej @emilevauge WDYT?

dtomcej:
I approved because I can't think that the code would be intended to function the way it was written :P I assume that @cheungpat's PR was what was intended.

timoreimann:
@ldez do we plan to make another 1.4 patch release, or is it okay to let the fix go into the upcoming 1.5 / master?

cheungpat:
Thanks for merging the pull request!

PR #2463, opened 2017-11-24
Change server receiver name.

ldez:
### What does this PR do?

Rename `Server` method receiver from `server` to `s`

### Motivation

Respect the convention.

Dedicate to @nmengin 😺


timoreimann:
Which convention are we referring to exactly?

The one I know is that the length of a variable should roughly match its context length. While some of the functions / methods this PR touches fall under this rule and likely justify shortening the name, we have a few others which seem way too long to come to a similar conclusion.

Overall, I'd prefer to keep the longer name where we'd otherwise lose context. To me, this holds when you can't tell what a variable name refers to when you have to scroll to find out.

emilevauge:
@timoreimann https://github.com/golang/go/wiki/CodeReviewComments#receiver-names
>The name of a method's receiver should be a reflection of its identity; often a one or two letter abbreviation of its type suffices (such as "c" or "cl" for "Client").

timoreimann:
@emilevauge hmm I think I confused receivers for method arguments. In the light of this, I can't come up with a good counter argument anymore.

Thanks for the extra pointer.

PR #2465, opened 2017-11-25
remove obsolete links in k8s docs

marco-jantke:
**Update**: as suggested by @timoreimann I removed the links completely.


timoreimann:
@marco-jantke looking at the docs page, I get the feeling that the two example links are misplaced: 
They appear in the annotations section even though neither of the manifests contain any annotations.

I'm thinking that it might be better to just drop the whole paragraph. We already have a pointer to our Kubernetes guide at the very top of the Kubernetes configuration page.

WDYT?

marco-jantke:
Makes definitely sense. Updated the PR description and title.

timoreimann:
Merge conflicts need to be resolved.

PR #2482, opened 2017-11-29
Bump Lego Version for GoDaddy DNS Provider

sjawhar:
### What does this PR do?
The _intention_ is just to bump the version of lego up to aa94fb46, to add support for the GoDaddy DNS provider. But glide and I don't seem to get along very well. It seems to just want to update ALL THE THINGS. I tried just running
```
./script/glide.sh get github.com/xenolf/lego#aa94fb4696349fd6f35168dabe3ab030ddf3e484
```
but I'm told "lego is already in glide.yaml, dummy!"

I've seen that some of you have made very small changes to glide.yaml successfully, and I'd love to know how I can do it too!

### Motivation
Now that that https://github.com/xenolf/lego/pull/416 has been merged, it would be great it the version of lego could be bumped here in traefik so I don't have to keep making custom builds :)

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,4 +1,4 @@
-hash: 4595cac0a682ce8e36b78630f12a3cfab75307fc58cb4a1f5e416017d3ae20d6
+hash: 8c5908b11f5078edd9ed93e2710ebb3a29b7e02d1259fddd679f8c46540becc9
 updated: 2017-11-29T12:05:49.613148632+01:00
 imports:
 - name: cloud.google.com/go
@@ -544,7 +544,7 @@ imports:
   - plugin/rewrite
   - router
 - name: github.com/xenolf/lego
-  version: 67c86d860a797ce2483f50d9174d4ed24984bef2
+  version: b929aa5aab5ad2e197bb3d74ef99fac61bfa47bc
   subpackages:
   - acme
   - providers/dns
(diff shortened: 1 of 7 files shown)


build.Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN mkdir -p /usr/local/bin \
 
 # Download docker
 RUN mkdir -p /usr/local/bin \
-    && curl -fL https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}-ce.tgz \
+    && curl -fL https://download.docker.com/linux/static/stable/x86_64/docker-${DOCKER_VERSION}.tgz \
```
sjawhar: When I built the `traefik-dev` image from build.Dockerfile, I noticed the `DOCKER_VERSION` environment variables was set to `17.09.0-ce`. Maybe Docker changed this recently? Anyway, the build wouldn't complete because it kept trying to download `docker-17.09.0-ce-ce.tgz`, which doesn't work so well.

glide.lock
```diff
@@ -544,7 +544,7 @@ imports:
   - plugin/rewrite
   - router
 - name: github.com/xenolf/lego
-  version: 67c86d860a797ce2483f50d9174d4ed24984bef2
+  version: aa94fb4696349fd6f35168dabe3ab030ddf3e484
```
sjawhar: This right here and the GoDaddy DNS provider on line 561 are all I really want to change. Any help accomplishing that would be much appreciate, very much a noob a gliding.

ldez:
@sjawhar Thanks for your interest in Traefik 😃 

I will fix your PR.

sjawhar:
@ldez Thanks for the assist! So I can be more helpful in the future, what's the _right_ way to update glide.yaml and .lock as minimally as you did?

Issue #572, opened 2016-07-31
Redirect domain.com -> www.domain.com


marccampbell:
Are there any plans to allow for a configuration that would be able to send a 301 redirect from a frontend to another URL? In this scenario, the other frontend may or may not be on the same traefik server, and I'd like to be able to have the client honor the redirect.

I currently host my www.domain.com web site on traefik and I have the http -> https redirect enabled. Today, visitors can visit any of the following:

| Requested Host and Protocol | Result |
| --- | --- |
| http://domain.com | 301 redirect to https://domain.com |
| https://domain.com | content |
| http://www.domain.com | 301 redirect to https://www.domain.com |
| https://www.domain.com | content |

I'm wondering if there are plans to allow: 

| Requested Host and Protocol | Result |
| --- | --- |
| http://domain.com | 301 redirect to https://www.domain.com |
| https://domain.com | 301 redirect to https://www.domain.com |
| http://www.domain.com | 301 redirect to https://www.domain.com |
| https://www.domain.com | content |

osixia:
I guess some people will also need the reverse www.domain.com -> domain.com

lxcid:
I read the docs but have no idea how to config this kind of redirection. Any help?

phoenix741:
I need this sort of feature to redirect from www.domain.com -> domain.com

thibaultlambermont:
Yes it will be perfect to have the capability to redirect www.domain.com -> domain.com
I don't find how to do it.

phoenix741:
To turn arround the problem, i use a container with nginx like this : https://github.com/Pindar/docker-nginx-redirect-301

thibaultlambermont:
Thanks ! But it will be nice to have it directly in Traefik no?

osixia:
I made a simple image that do the job too:
https://github.com/osixia/docker-www-redirect

Use it behind traefik and : 
  - If the requested host is www.example.org/whatever the visitor will be redirected to example.org/whatever

  - If the requested host is example.org/whatever the visitor will be redirected to www.example.org/whatever

But i'm agree have it directly in traefik would be great :)

NiklasA:
My "ugly" workaround was it to create a docker container in Traefik, which solves the whole thing with a single HTML page. The HTML page allows it to redirect from `http(s)://www.*` to `http(s)://*`. A integrated solution in Traefik, would be very nice!

ubershmekel:
Here's a redirecting service you can put in a `docker-compose.yml`

```
[...]

PR #2495, opened 2017-11-30
Add Feature Redirect on Frontends based on Regex and Replacement

### LINKED ISSUE #1380: Add option to force www addition or removal

Version:      v1.2.1
Codename:     morbier
Go version:   go1.7.5
Built:        2017-03-27_03:19:40PM
OS/Arch:      linux/amd64

Docker, no swarm:

        command:
            - --ACME.ACMELogging
            - --ACME.Email=mail@example.com
            - --ACME.EntryPoint=https
            - --ACME.OnHostRule
            - --ACME.Storage=/etc/traefik/acme/acme.json
            - --DefaultEntryPoints=http,https
            - --EntryPoints=Name:http Address::80 Redirect.EntryPoint:https
            - --EntryPoints=Name:https Address::443 TLS
            - --LogLevel=INFO
            - --Docker
            - --Docker.EndPoint=http://dockersocket:2375
            - --Docker.ExposedByDefault=false
            - --Docker.Watch

### What did you do?

I wanted to redirect `www.example.com` to `example.com` in a website running behind Traefik. I wanted to do the opposite (`example.org` to `www.example.org`) in another website.

I expected frontends to let you have a list of alternative host names a main one that gets all of those redirected, so I could do something like this in my `docker-compose.yaml`:

```
labels:
  traefik.frontend.forcehost: example.com
  traefik.frontend.althosts: www.example.com,www2.example.com
```

ACME should get certs for all of them.

Given this situation of `www.` removal or addition is so common (I had to do that on almost every website I ever launched in my life), maybe a shortcut flag would be useful:

```
labels:
  traefik.frontend.www: force # or: remove
```

However, first option is more flexible indeed.

### What did you see instead?

Traefik cannot redirect by frontend, just by entrypoint. This reminds me to #1346.

### Possible workarounds

2. Reconfigure Traefik itself with a spaghetti regexp on the frontends rule definition. This is quite unmaintainable because you'd be required to change it each time you boot a new microservice in a similar situation, and the regexp would grow and grow.
1. Add a container to the stack that listens on `www.example.com` and just returns a 301 redirection to `example.com`. Do the same for `example.org`. This would be more maintainable than previous workaround, but it seems a little bit absurd.
3. I guess https://github.com/containous/traefik/pull/1257 will help in v1.13 on serving both sites from same container, although not on the redirection problem.
4. Force the host name in your app. Depending on the app you're deploying, this can be easier or harder; in any case, still not a perfect solution.

aek:
### What does this PR do?
Allow Traefik to make redirect based on config fields from frontends with more liberty to make redirections from (sub)domain A to (sub)domain B and specify any custom redirection rules that will not affect all of the rest of the entrypoint frontends definitions

### Motivation
Non-Existing way to express redirection rules 
Fixes #1380

### Additional Notes

Working example of usage making the (sub)domains 
- [http://sladev.com](http://sladev.com)
- [https://sladev.com](https://sladev.com)
- [http://www.sladev.com](http://www.sladev.com)

To be redirected to:
- [https://www.sladev.com](https://www.sladev.com)

```toml
[backends]
    [backends.sladev-redirect]
    [backends.sladev]
        [backends.sladev.servers.instance]
            url = "http://127.0.0.1:9001"

[frontends]
    [frontends.sladev-redirect]
        redirectRegex = "(http://sladev.com|https://sladev.com|http://www.sladev.com)"
        redirectReplacement = "https://www.sladev.com"
        backend = "sladev-redirect"
        priority = 1
        entrypoints = ["http", "https"]
    [frontends.sladev]
        backend = "sladev"
        passHostHeader = true
        entrypoints = ["https"]
        [frontends.sladev.routes.instance]
            rule = "Host:www.sladev.com"
```


aek:
Hi @ldez 
Could you have a look at this one?. I think that there is no block for the current behavior, this just provides an enhancement of the redirect features of Traefik to avoid builds like those mentioned on 
https://github.com/containous/traefik/issues/572#issuecomment-258988609

ldez:
@aek I took the meaning of this PR and created a #2570.

I did that because the "good" way to that is a breaking change.
But the redirect by frontend have been added in 1.5 and 1.5 is in RC phase, which allow us, very exceptionally, to do this breaking change. 

I replaced the `Redirect` with a `string` type:
https://github.com/containous/traefik/blob/b6f5a66fabdfb6350d955a784844f45f2705c9f0/types/types.go#L155-L157

By the `Redirect` type use in entry point:
https://github.com/containous/traefik/blob/b6f5a66fabdfb6350d955a784844f45f2705c9f0/configuration/configuration.go#L435-L439

aek:
Thanks @ldez for your improvements and changes, already tested in v1.5.0-rc5 and it's working as expected

PR #2235, opened 2017-10-09
Check X-Forwarded-For and X-Real-Ip headers for whitelisted IPs

joe-bowman:
Taking @joshdvir's contribution in https://github.com/containous/traefik/pull/1915 and make it configurable as per @emilevauge and @marco-jantke requests.

Docs and tests to follow!

Issue #2311, opened 2017-10-24
Cant get Real-Client-IP


koljakutschera:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

Using Docker image: traefik:alpine as main reverseproxy with following config:

```toml
checkNewVersion = true
logLevel = "ERROR"
defaultEntryPoints = ["https","http"]

[web]
address = ":8080"

[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    entryPoint = "https"
  [entryPoints.https]
  address = ":443"
  [entryPoints.https.tls]

[retry]

[docker]
endpoint = "unix:///var/run/docker.sock"
domain = "docker.localhost"
watch = true
exposedbydefault = false

[acme]
email = "not@my.email"
storage = "acme.json"
entryPoint = "https"
OnHostRule = true
acmeLogging = true
```

for some containers with e.g. the following config:

```yml
      - "traefik.enable=true"
      - "traefik.example-node.backend=example-node"
      - "traefik.example-node.frontend.passHostHeader=true"
      - "traefik.example-node.port=80"
      - "traefik.example-node.docker.network=reverseproxy_default"
      - "traefik.example-node.frontend.rule=Host:not.my.domain"
```

### What did you expect to see?

Everything is working perfectly BUT when i log headers in my app-container i got:

```
,"x-forwarded-for":"172.16.0.1","x-forwarded-host":"not.my.domain","x-forwarded-proto":"https","x-forwarded-server":"xaassaxasxs"
```

172.16.0.1 is just the gateway of the reverseproxy-network.
But i expect to see the REAL-ip of the client.
I think in real-world-applications you need the real ips to e.g. rate-limit by ip...
I like traefik but this is a no-go.
I tried many different docker-network settings and `frontend.passHostHeader`... but traefik just wont expose the real-ips of the clients or did i miss something?

### What did you see instead?

IP of docker-network. Need real-client-ip

### Output of `traefik version`: (_What version of Traefik are you using?_)

1.4.0 alpine

Thank you very much.

nmengin:
Hello @koljakutschera ,

Many thanks for your interest in the project.
But are you sure `172.16.0.1 is just the gateway of the reverseproxy-network`?

Because I reproduce an environment similar to yours in a compose file and the `X-forwarded-for` header I get in my backend contains the IP address from the client.

My configuration :

```yaml
version: '2'

services:
  proxy:
[...]

PR #2502, opened 2017-11-30
Conditionally use IPs in X-Forwarded-For for whitelisting; Bypass Authentication for whitelisted IPs.

### LINKED ISSUE #2260: Make the IP "whitelisting" more flexible
### Feature: Increase the flexibility of the whitelisting

The current whitelisting acts as a strict gate to incoming traffic. The authentication is also used as a strict gate. To gain access you have to meet *both* criteria and there is no mechanism to support meeting *either*.

Whilst it may be too late to change this naming now it would appear to the current functionality is more of a `greylist` than a `whitelist` since the requests get challenged again.

The feature behaviour would be have two groups of access lists, one that does indeed whitelist the traffic and hence allow it straight through, and a second which allows the traffic through to the next level of checking.

For the simple case of no auth the two lists would be essentially the same. And if no lists are set the auth would continue to be the decider.

There is no obvious need for the `blacklist` variant since this can be better achieved using a normal firewall.

### What did you do?

Using the docker image `traefik:1.4` enable both authentication and whitelisting for entrypoints:

```toml
[entryPoints]
    [entryPoints.http]
    address = ":80"
    whiteListSourceRange = [ "x.x.x.x" ]

    [entryPoints.http.auth.basic]
    usersFile = "/etc/htpasswd"
```

Requests that failed the IP address check would trigger the basic authentication instead.

### What did you see instead?

* Requests that failed the IP address check where blocked immediately.
* Requests that met the IP address check had to meet the basic authentication criteria

```
Version:      v1.4.0-rc5
Codename:     roquefort
Go version:   go1.9.1
Built:        2017-10-10_02:04:46PM
OS/Arch:      linux/amd64
```

### Additional notes

There is a secondary part to this which is that when you are using the docker container the incoming IP address may be that of the docker host, not the original requesting host. This is certainly the case if you are working with a swarm because then the traefik container must be on an internal network. Ideally the checks would be against the originating IP address.

nemosupremo:
### What does this PR do?

This PR provides the functionality for Traefik to whitelist based on the `X-Forwarded-For` header optionally. This mirrors functionality in other reverse proxies, like [nginx](http://nginx.org/en/docs/http/ngx_http_realip_module.html) with the `set_real_ip` directive. It adds a new configuration option `whitelistTrustProxy` that accepts a list of CIDRs or IPs. If a request comes in that matches the IP of `whitelistTrustProxy`, traefik will whitelist based on the values in `X-Forwarded-For`. If there are multiple IPs in `X-Forwarded-For` (comma separated), it will use the last IP, as that is the only IP that can considered trustworthy by the [defacto standard](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html). If the remote IP is the second to last (or greater) IP, you can enter those IPs in `whitelistTrustProxy`, and Traefik will skip those IPs. For example with a configuration:

```toml
whitelistTrustProxy = ["trusted-ip1", "trusted-ip2"]
```

A request like so:

```
X-Forwarded-For: malicious-ip, client-ip, trusted-ip3, trusted-ip2, trusted-ip1
```

will return `trusted-ip3`. You can add the CIDR for the trusted-ips to get the `client-ip`. This functionality is needed for instances behind Google LB or CloudFlare.

This PR also fixes #2260, by extending this functionality to the `Auth` component. With the following config, a request from 172.17.0.1 will go through unchallenged, while any other request from any other IP will be challenged.
```toml
[entryPoints]
  [entryPoints.http]
  address = ":80"
  [entryPoints.http.auth]
    whitelistSourceRange = ["172.17.0.1"]
    whitelistTrustProxy = ["127.0.0.1/8"]
    [entryPoints.http.auth.basic]
    users = ["test:traefik:a2688e031edb4be6a3797f3882655c05 ", "test2:traefik:518845800f9e2bfb1f1f740ec24f074e"]
    usersFile = "/path/to/.htdigest"
```

This syntax is backwards compatible with Traefik deployments today (users won't suddenly see their authenticated endpoints bypassed by whitelisted IPs).

This PR provides a similar solution to the one outlined in #2235, but is stricter on the `X-Forwarded-For` policy.

### Motivation

We are using Traefik, behind Amazon ELB, to route requests from the outside world and inside our cluster to various microservices, for example a docker registry.  From inside our network, we wanted to be able use commands like `docker pull example.com/web` without a `docker login`, but also have tools like CircleCI also push images to our registry from the outside with authentication. This use case created the two problems we needed solved:

1. For an IP based authentication whitelist (so instances on our cluster could communicate unchallenged, but 3rd party services would be challenged).
2. A  trust `X-Forwarded-For` proxy so that requests that came into ELB would have their "real" ip exposed to Traefik.

### More

- [x] Added/updated tests

### Additional Notes

PR #2235 also did some work on forwarding this information to the Kubernetes frontend. Unfortunately I haven't used Kubernetes and I'm unsure if this option needs to be forwarded to the Kubernetes backend. If that is something that needs to be included, if someone can explain to me how that works I'd be glad to include it (or it can be included in another PR).

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -330,6 +330,10 @@ func (ep *EntryPoints) Set(value string) error {
 	if len(result["whitelistsourcerange"]) > 0 {
 		whiteListSourceRange = strings.Split(result["whitelistsourcerange"], ",")
 	}
+	whiteListTrustProxy := []string{}
+	if len(result["whitelisttrustproxy"]) > 0 {
+		whiteListTrustProxy = strings.Split(result["whitelisttrustproxy"], ",")
+	}
 
 	compress := toBool(result, "compress")
 
@@ -367,6 +371,7 @@ func (ep *EntryPoints) Set(value string) error {
 		Redirect:             redirect,
 		Compress:             compress,
 		WhitelistSourceRange: whiteListSourceRange,
+		WhitelistTrustProxy:  whiteListTrustProxy,
 		ProxyProtocol:        proxyProtocol,
 		ForwardedHeaders:     forwardedHeaders,
 	}
@@ -426,6 +431,7 @@ type EntryPoint struct {
 	Redirect             *Redirect   `export:"true"`
 	Auth                 *types.Auth `export:"true"`
 	WhitelistSourceRange []string
+	WhitelistTrustProxy  []string
 	Compress             bool              `export:"true"`
 	ProxyProtocol        *ProxyProtocol    `export:"true"`
 	ForwardedHeaders     *ForwardedHeaders `export:"true"`
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -330,7 +330,8 @@ func (ep *EntryPoints) Set(value string) error {
 	if len(result["whitelistsourcerange"]) > 0 {
 		whiteListSourceRange = strings.Split(result["whitelistsourcerange"], ",")
 	}
-	whiteListTrustProxy := []string{}
+
+	whiteListTrustProxy := []string(nil)
 	if len(result["whitelisttrustproxy"]) > 0 {
 		whiteListTrustProxy = strings.Split(result["whitelisttrustproxy"], ",")
 	}
--- a/docs/configuration/entrypoints.md
+++ b/docs/configuration/entrypoints.md
@@ -140,6 +140,24 @@ Users can be specified directly in the toml file, or indirectly by referencing a
   usersFile = "/path/to/.htdigest"
 ```
 
+### IP Whitelist Bypass
+
+You can optionally setup certain IP ranges to bypass Basic and Digest authentication. A request from
+a whitelisted IP will not have to authenticate.
+
+```toml
+# To enable digest auth on an entrypoint with 2 user/realm/pass: test:traefik:test and test2:traefik:test2
+[entryPoints]
+  [entryPoints.http]
+  address = ":80"
+  [entryPoints.http.auth]
+    whitelistSourceRange = ["172.17.0.1"]
+    whitelistTrustProxy = ["127.0.0.1/8"]
+    [entryPoints.http.auth.basic]
+    users = ["test:traefik:a2688e031edb4be6a3797f3882655c05 ", "test2:traefik:518845800f9e2bfb1f1f740ec24f074e"]
+    usersFile = "/path/to/.htdigest"
+```
+
 ### Forward Authentication
 
 This configuration will first forward the request to `http://authserver.com/auth`.
@@ -154,7 +172,7 @@ Otherwise, the response from the auth server is returned.
     # To enable forward auth on an entrypoint
     [entryPoints.http.auth.forward]
     address = "https://authserver.com/auth"
-    
+
     # Trust existing X-Forwarded-* headers.
     # Useful with another reverse proxy in front of Traefik.
     #
@@ -162,7 +180,7 @@ Otherwise, the response from the auth server is returned.
     # Default: false
     #
     trustForwardHeader = true
-    
+
     # Enable forward auth TLS connection.
     #
     # Optional
@@ -210,13 +228,16 @@ Responses are compressed when:
 
 ## Whitelisting
 
-To enable IP whitelisting at the entrypoint level.
+To enable IP whitelisting at the entrypoint level. You can also set the `whitelistTrustProxy` option to have whitelisting
+be done on the IP provided by the `X-Forwarded-For` header, this is useful if Traefik is behind another proxy or load
+balancer and you want to filter by the client's IP.
 
 ```toml
 [entryPoints]
   [entryPoints.http]
   address = ":80"
   whiteListSourceRange = ["127.0.0.1/32", "192.168.1.7"]
+  # whitelistTrustProxy = ["127.0.0.1/32"]
 ```
 
 ## ProxyProtocol
@@ -226,7 +247,7 @@ Only IPs in `trustedIPs` will lead to remote client address replacement: you sho
 
 !!! danger
     When queuing Træfik behind another load-balancer, be sure to carefully configure Proxy Protocol on both sides.
-    Otherwise, it could introduce a security risk in your system by forging requests. 
+    Otherwise, it could introduce a security risk in your system by forging requests.
 
 ```toml
 [entryPoints]
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 03f8c75c2523eaea36cd61abc4d335ce68115c80404de7a10d8d4f0225957fc9
-updated: 2017-11-28T15:59:09.002650953+01:00
+hash: 4595cac0a682ce8e36b78630f12a3cfab75307fc58cb4a1f5e416017d3ae20d6
+updated: 2017-11-29T12:05:49.613148632+01:00
 imports:
 - name: cloud.google.com/go
   version: 2e6a95edb1071d750f6d7db777bf66cd2997af6c
@@ -111,10 +111,6 @@ imports:
   - pkg/srv
   - pkg/types
   - version
-- name: github.com/coreos/go-semver
-  version: 8ab6407b697782a06568d4b7f1db25550ec2e4c6
-  subpackages:
-  - semver
 - name: github.com/coreos/go-oidc
   version: 5644a2f50e2d2d5ba0b474bc5bc55fea1925936d
   subpackages:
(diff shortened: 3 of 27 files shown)


beenanner:
@nemosupremo I was able to get a whitelist working today with nginx in front of traefik by using the following configs. I also tested injecting a x-forwarded-for of 8.8.8.8 to verify the setup worked with malicious ips.

Nginx configs used

```
set_real_ip_from 10.0.0.0/8;
set_real_ip_from 130.211.0.0/22;
set_real_ip_from 35.227.230.198/32;
real_ip_header X-Forwarded-For;
real_ip_recursive on;

server {
    listen 80;
    listen 8080;
    listen 443 ssl;
    server_name localhost;

    ssl_certificate /etc/ssl/certs/server.crt;
    ssl_certificate_key /etc/ssl/private/server.key;

    include /var/www-allow/whitelist.conf;

    proxy_set_header Host            $host;
    proxy_set_header X-Forwarded-For '$http_x_forwarded_for, $realip_remote_addr';
    proxy_set_header X-Real-IP $remote_addr;

    location / {
        if ($scheme = http) {
            proxy_pass http://traefik;
        }
        if ($scheme = https) {
            proxy_pass https://traefik;
        }
    }
    location ~ ^/(dashboard|api).*/ {
            proxy_pass http://traefik:8080;
    }
}
```

results in the following 

```
Hostname: f3a624b14a6d
IP: 127.0.0.1
IP: 10.0.0.2
IP: 10.0.0.3
IP: 172.18.0.3
GET / HTTP/1.1
Host: whoami.test.com
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.9
Cache-Control: no-cache
Pragma: no-cache
Upgrade-Insecure-Requests: 1
Via: 1.1 google
X-Cloud-Trace-Context: fcfaabeb8736495d6b0de68b3d394405/17312253612327182169
X-Forwarded-For: 8.8.8.8, 64.30.233.86, 35.227.230.198, 130.211.2.75, 10.0.0.9
X-Forwarded-Host: whoami.test.com
X-Forwarded-Port: 80
X-Forwarded-Proto: http
X-Forwarded-Server: 4191ca3f8689
X-Real-Ip: 64.30.233.86
```


Here are the things that matter for the GCP LB
**
set_real_ip_from 130.211.0.0/22; <-- proxy LB IPs
set_real_ip_from 35.227.230.198/32; <-- real LB IP
real_ip_header X-Forwarded-For;
real_ip_recursive on;**

**proxy_set_header X-Forwarded-For '$http_x_forwarded_for, $realip_remote_addr';
proxy_set_header X-Real-IP $remote_addr;**

nemosupremo:
Ok, I've built my feature branch and pushed it as a docker image to [nemothekid/traefik](https://hub.docker.com/r/nemothekid/traefik/).

According to your nginx config, the following traefik config should work:
```toml
[entryPoints]
  [entryPoints.http]
  address = ":80"
  whitelistSourceRange = ["127.0.0.1"]
  whitelistTrustProxy = ["35.227.230.198", "130.211.0.0/22",  "10.0.0.0/8"]
```

I'm also pretty sure in GCP, `10.0.0.0/8` is your Google private network range (or the IPs of your hosts) so for anyone reading this, you would likely need that as well even if you weren't using docker. I was also confused about the 3rd IP (which is the real LB IP), thanks for clearing that up.

beenanner:
@nemosupremo is correct you will need the 10.0.0.0/8 or whatever your private CIDR block is regardless of running docker or standalone.

@nemosupremo sweet thanks for throwing this together! I'll give your image a go and report back.

beenanner:
Looks like this is working! The only issue I see now is the X-Real-IP is still the last in the list.

```
Hostname: e3c37638a4e1
IP: 127.0.0.1
IP: 10.0.0.2
IP: 10.0.0.3
IP: 172.18.0.3
GET / HTTP/1.1
Host: whoami.acme.com
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3202.94 Safari/537.36
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.9
Cache-Control: max-age=0
Upgrade-Insecure-Requests: 1
Via: 1.1 google
X-Cloud-Trace-Context: 6a3b51bb7bf9e86f0127fb318c3df4fc/7369758590415622740
X-Forwarded-For: 170.20.11.31, 35.227.230.198, 130.211.2.173
X-Forwarded-Host: whoami.acme.com
X-Forwarded-Port: 80
X-Forwarded-Proto: http
X-Forwarded-Server: d50554d3e2ed
X-Real-Ip: 130.211.2.173
```

I see this in traefik with debug on so it seems the source-IP it's checking against is correct. Just not sure where X-Real-IP is set.

```
time="2017-12-01T15:28:02Z" level=debug msg="source-IP 170.20.11.31 matched whitelist &{[%!s(*net.IP=&[0 0 0 0 0 0 0 0 0 0 255 255 35 227 230 198])] [%!s(*net.IPNet=&{[10 0 0 0] [255 0 0 0]}) %!s(*net.IPNet=&{[170 20 11 0] [255 255 255 128]}) %!s(*net.IPNet=&{[130 211 0 0] [255 255 252 0]})] %!s(bool=false)} - passing"
```

nemosupremo:
I think the `X-Real-Ip` is coming from somewhere else (the LB?, I'm not sure). There's nothing in traefik's codebase I could fine about `X-Real-Ip`.

beenanner:
Back tracking through to https://github.com/containous/traefik/issues/2311 looks like it may be part of the vendors that were added to 1.4.1.

https://github.com/containous/traefik/blob/44b82e62311232bff6887e5a78c73312c0236712/vendor/github.com/vulcand/oxy/forward/rewrite.go#L30

beenanner:
Made @ldez could shed some light on this issue for us?

nemosupremo:
Looks like its being set by the underlying http library [oxy](http://github.com/vulcand/oxy). Fixing the `X-Real-IP` header would mean submitting a PR to that repo, and then integrating an option in traefik to passdown all the way to oxy. 

That would be a separate issue - I'm really only concerned about whitelisting.

beenanner:
@nemosupremo totally agreed as this indeeds solves the original issue of the whitelisting. The [oxy](https://github.com/vulcand/oxy) fix would be another correction that could be integrated later in a separate ticket in the event that downstream apps need to use X-Real-IP for logging or other things.

emilevauge:
Hi @nemosupremo, thanks a lot for you contribution. We want to keep the authentication section as simple as possible and we think that your use case could be addressed differently. We advise you to either use a vpn, either use the [authentication forward mechanism](http://docs.traefik.io/configuration/entrypoints/#forward-authentication) we introduced recently.
Thanks @nemosupremo

nemosupremo:
@emilevauge 

I'm revisiting this as I've currently deployed a fork of traefik with the features I needed and I want to figure out how I can get back to mainline. I'm going to contribute back the mesos changes (#2617) so they fit the new model, but this feature I'm a but puzzled as to how I would get it with authentication forwarding. Authentication forwarding seems to only work at the entrypoint level, whereas I need authentication/ip checks at the frontend level. 

Maybe it would help to explain my setup. We use Traefik as a frontend reverse proxy for all our http traffic. Some of our frontends are public (reachable by the internet), private (only reachable from a specific IP range) and authenticated (reachable by the internet with credentials OR reachable from a specific IP range). For example, we deploy Apache Spark on our cluster and we use Traefik to route to the Spark API, so we have machines in our infrastructure that will communicate with Traefik over the private secured network without authentication, but we need to also expose it publicly so that CircleCI can also push updates to the Spark frameworks working in our cluster. See how we don't want the public to run code on our machines via Spark, we use http authentication to secure that endpoint between CircleCI and Spark.

Using a VPN here is a non-starter - CircleCI and some of the other third party integrations we work with don't allow us to install VPNs on their machines. 

I looked at authentication forwarding, but it since we use Traefik for all our traffic, the authentication forwarding service build would also have to build a parallel configuration of Traefik's entire backend to make decisions on whether to allow authorization or not. Since we use container tags to register with traefik, the authentication forwarding mechanism would also have to contain logic for communicating with our cluster and parsing the traefik rules to make decisions on authentication. Authentication forwarding doesn't seem very practical. If authentication forwarding works on a frontend level, it would reduce the amount of code we would have to deploy, but still doesn't seem very attractive as it increases our security complexity.

Maybe you have a better solution for my use case that's more idiomatic? Essentially I want stricter access to a frontend based on source IP address.

PR #2512, opened 2017-12-03
Specify default backend name in Kubernetes config backend

cheungpat:
### What does this PR do?

This PR add a new option in Kubernetes config backend to specify a default backend name when Kubernetes ingress has no available endpoints. 

### Motivation

The Kubernetes Ingress controller for nginx has an option to specify a “default backend”. This backend handles all traffic when nginx doesn’t know how to handle the request. This backend is also used when there is no pod to serve the request.

I am trying to replicate the same behavior in traefik such that when there is no upstream servers serving the frontend, the request is routed to a default backend. The default backend can then present content rather than returning an error.

### More

- [x] Added/updated tests

### Additional Notes

The new option has an empty default, which preserve existing behavior (backend config is still created even if no upstream servers are available).

When a default backend name is specified, this default backend name is used in these two cases:

* when no kubernetes service is found
* when there is no available endpoints for the kubernetes ingress

The default backend config has to be configured manually. This is because the kubernetes config backend only creates a backend config when there is an ingress. If this is not desirable, I can change this PR so that a user can specify a kubernetes service as the default backend instead, but the PR is likely to grow much bigger because I need to extract the logic for creating backend config from kubernetes service.

### DIFF
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -71,6 +71,13 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # Default: <built-in template>
 #
 # filename = "kubernetes.tmpl"
+
+# Default backend name to use when ingress has no endpoint.
+#
+# Optional
+# Default: empty (no backend servers)
+#
+# defaultBackend = "default-backend"
 ```
 
 ### `endpoint`
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -70,6 +70,7 @@ type Provider struct {
 	EnablePassTLSCert      bool       `description:"Kubernetes enable Pass TLS Client Certs" export:"true"`
 	Namespaces             Namespaces `description:"Kubernetes namespaces" export:"true"`
 	LabelSelector          string     `description:"Kubernetes api label selector to use" export:"true"`
+	DefaultBackend         string     `description:"Default backend name to use when ingress has no endpoint" export:"true"`
 	lastConfiguration      safe.Safe
 }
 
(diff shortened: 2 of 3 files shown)


ldez:
WDYT @containous/kubernetes ?

timoreimann:
@cheungpat are you certain that the Nginx Ingress controller also uses the default backend when no pods are available? The [documentation](https://github.com/kubernetes/ingress-nginx#requirements) only mentions the case where requests are not mapped by any Ingress, i.e., no matching URL paths or hosts exist.

It could be noteworthy that we already have a mechanism to support [custom error pages](http://docs.traefik.io/configuration/commons/#custom-error-pages), albeit currently only supported for the file provider. I wonder if we think it's worthwhile to have a unified approach for this feature instead of something Kubernetes-specific. For me, the answer to this question also depends on how Nginx and other Ingress controllers do it, i.e., whether there's somewhat of a Kubernetes standard already.

As to the question whether the backend should be defined literally or rather through a Service: I'm leaning towards the latter due to the greater flexibility. For instance, you could swap it out during runtime without having to restart Traefik, thereby allowing dynamically changing default backends in the potential future. We probably want to extract the necessary logic anyway at some point because the Ingress spec also supports [Ingress-specific default backends](https://kubernetes.io/docs/api-reference/v1.8/#ingressspec-v1beta1-extensions).

@dtomcej @errm WDYT?

cheungpat:
@timoreimann I looked through the source code of nginx ingress controller and this is what I found

* The [annotations](https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/annotations.md#default-backend) section describe the default backend as “This service is handle the response when the service in the Ingress rule does not have endpoints.”
* Nginx ingress controller route request to default backend when there is no endpoints till version 0.9.0-beta.15.
* I just realized that nginx ingress controller no longer do that since 0.9.0-beta.16 😕.
* pull request for the change: https://github.com/kubernetes/ingress-nginx/pull/1513

So the rationale for the change is that when there is no endpoints for the ingress, the ingress controller should return a 503 rather than 404 (the 404 is returned from the default backend).

Totally understand that traefik should follow the standard kubernetes way.

timoreimann:
@cheungpat thanks for digging into the current state of affairs as far as the Nginx controller is concerned.

I'd still say that injecting a custom error handler could be useful. We already have that feature for the file provider, and a number of people have already asked to do the same with Kubernetes (and probably other providers as well). An implementation for it likely shouldn't be based on the very annotation that the Nginx controller used to have, however. Instead, I think we should try to integrate our error handling logic in a clean fashion.

cheungpat:
@timoreimann Thanks for your comments! I am going to close this pull request because it appears more design work is needed.

PR #2516, opened 2017-12-04
Simplification

ferhatelmas:
### What does this PR do?

Does some simplification.

### Motivation

Makes code easier to understand since uses less tokens.

### More

- [x] Added/updated tests (One small change touches tests)

### Additional Notes

- drop `select` with only one `case`

### DIFF
--- a/provider/docker/docker.go
+++ b/provider/docker/docker.go
@@ -190,13 +190,8 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 
 				} else {
 					pool.Go(func(stop chan bool) {
-						for {
-							select {
-							case <-stop:
-								cancel()
-								return
-							}
-						}
+						<-stop
+						cancel()
 					})
 					f := filters.NewArgs()
 					f.Add("type", "container")
--- a/provider/dynamodb/dynamodb.go
+++ b/provider/dynamodb/dynamodb.go
@@ -155,10 +155,8 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 	pool.Go(func(stop chan bool) {
 		ctx, cancel := context.WithCancel(context.Background())
 		safe.Go(func() {
-			select {
-			case <-stop:
-				cancel()
-			}
+			<-stop
+			cancel()
 		})
 
 		operation := func() error {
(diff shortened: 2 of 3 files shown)


provider/docker/docker.go
```diff
@@ -205,12 +205,8 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
 				} else {
 					pool.Go(func(stop chan bool) {
-						for {
-							select {
-							case <-stop:
-								cancel()
-								return
-							}
+						for range stop {
+							cancel()
```
timoreimann: Hmm, is this really exactly equivalent for the case where the channel gets closed? Doesn't the old version still call `cancel()` while the new one doesn't anymore?

Also, couldn't we just do

```go
<-stop
cancel()
```

as we immediately return anyway when the channel fires?

ferhatelmas: My understanding here was that we need to cancel if we take stop.  Refactoring should be keeping semantics.

Only `<-stop` will call anyway so it's wrong to me.

timoreimann: Inside the goroutine, we select on the stop channel without any other select case available. That means we can only ever escape the select when the stop channel fires (either through a message or a close). Once it does, we invoke `cancel` and return immediately, giving us zero chance to process any further events from the stop channel (which makes total sense as we want to process the stop signal exactly once).

I think my simplification suggestion does effectively the same, using just fewer keywords: wait for a channel event; call `cancel`; and then return (implicitly). Please let me know where you see the semantical difference.

Should be good though to have another pair of eyes on this one.

ldez:
@ferhatelmas could you rebase ?

provider/docker/docker.go
ferhatelmas: Since it was different from other cases bottom, I thought intention was to call only on getting something on channel. Otherwise, semantic will be kept same with just `<-stop` and actually putting `for range` won't call `cancel` if we just close channel. Updating this part and rebasing as requested by @ldez.

ferhatelmas:
@timoreimann @ldez rebased and diff got even smaller :smile:

ferhatelmas:
Updated additional notes to reflect last status.

PR #2522, opened 2017-12-04
Add app-root annotation support for kubernetes ingress

yue9944882:
### What does this PR do?
Rediret only root path elsewhere by ingress annotation. Like Nginx Ingress Controller [documented](https://github.com/kubernetes/ingress-nginx/blob/27c863085e2503e5f3249b596f24529fc2488baa/docs/examples/rewrite/README.md#deployment). Implementing this annotation will only perform `ReplacePath` on `/` path on all hosts present in the annotated ingress.

`ingress.kubernetes.io/app-root` or `traefik.ingress.kubernetes.io/app-root`

### Motivation

We catch this use case in production deployment.

### More

- [x] Added/updated tests

### Additional Notes
Seems like many new annotation support will be added to Kubernetes provider on 1.5 release, I hope this annotation can also be supported.

### DIFF
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -127,7 +127,7 @@ The following general annotations are applicable on the Ingress object:
 | `traefik.ingress.kubernetes.io/rewrite-target: /users`                          | Replaces each matched Ingress path with the specified one, and adds the old path to the `X-Replaced-Path` header.                               |
 | `traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip`                      | Override the default frontend rule type. Default: `PathPrefix`.                                                                                 |
 | `traefik.ingress.kubernetes.io/whitelist-source-range: "1.2.3.0/24, fe80::/16"` | A comma-separated list of IP ranges permitted for access. all source IPs are permitted if the list is empty or a single range is ill-formatted. |
-| `traefik.ingress.kubernetes.io/app-root: "/index.html"` | Redirects all requests for / to the defined path. Non-root paths will not be affected by this annotation and handled normally. This annotation may not be combined with the ReplacePath rule type or any other annotation leveraging that rule type. Trying to do so leads to an error and the corresponding Ingress object being ignored. |
+| `traefik.ingress.kubernetes.io/app-root: "/index.html"` | Redirects all requests for `/` to the defined path. Non-root paths will not be affected by this annotation and handled normally. This annotation may not be combined with the `ReplacePath` rule type or any other annotation leveraging that rule type. Trying to do so leads to an error and the corresponding Ingress object being ignored. |
 
 <1> `traefik.ingress.kubernetes.io/error-pages` example:
 
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -127,6 +127,7 @@ The following general annotations are applicable on the Ingress object:
 | `traefik.ingress.kubernetes.io/rewrite-target: /users`                          | Replaces each matched Ingress path with the specified one, and adds the old path to the `X-Replaced-Path` header.                               |
 | `traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip`                      | Override the default frontend rule type. Default: `PathPrefix`.                                                                                 |
 | `traefik.ingress.kubernetes.io/whitelist-source-range: "1.2.3.0/24, fe80::/16"` | A comma-separated list of IP ranges permitted for access. all source IPs are permitted if the list is empty or a single range is ill-formatted. |
+| `traefik.ingress.kubernetes.io/app-root: "/index.html"` | Redirects all requests for / to the defined path. Non-root paths will not be affected by this annotation and handled normally. This annotation may not be combined with the ReplacePath rule type or any other annotation leveraging that rule type. Trying to do so leads to an error and the corresponding Ingress object being ignored. |
 
 <1> `traefik.ingress.kubernetes.io/error-pages` example:
 
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -229,11 +229,11 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 				rule, err := getRuleForPath(pa, i)
 				if err != nil {
 					log.Errorf("Failed to get rule for ingress %s/%s: %s", i.Namespace, i.Name, err)
-					delete(templateObjects.Frontends, r.Host+pa.Path)
+					delete(templateObjects.Frontends, baseName)
 					continue
 				}
 				if rule != "" {
-					templateObjects.Frontends[r.Host+pa.Path].Routes[pa.Path] = types.Route{
+					templateObjects.Frontends[baseName].Routes[pa.Path] = types.Route{
 						Rule: rule,
 					}
 				}
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -228,7 +228,7 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 
 				rule, err := getRuleForPath(pa, i)
 				if err != nil {
-					log.Errorf("fail to get rule from ingress path: %s", err)
+					log.Errorf("Failed to get rule for ingress %s/%s: %s", i.Namespace, i.Name, err)
 					delete(templateObjects.Frontends, r.Host+pa.Path)
 					continue
 				}
@@ -328,6 +328,9 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
 	rules := []string{ruleType + ":" + pa.Path}
 
 	var pathReplaceAnnotation string
+	if ruleType == ruleTypeReplacePath {
+		pathReplaceAnnotation = annotationKubernetesRuleType
+	}
 
 	if rewriteTarget := getStringValue(i.Annotations, annotationKubernetesRewriteTarget, ""); rewriteTarget != "" {
 		if pathReplaceAnnotation != "" {
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -226,8 +226,14 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 					}
 				}
 
-				if rule := getRuleForPath(pa, i); rule != "" {
-					templateObjects.Frontends[baseName].Routes[pa.Path] = types.Route{
+				rule, err := getRuleForPath(pa, i)
+				if err != nil {
+					log.Errorf("fail to get rule from ingress path: %s", err)
+					delete(templateObjects.Frontends, r.Host+pa.Path)
+					continue
+				}
+				if rule != "" {
+					templateObjects.Frontends[r.Host+pa.Path].Routes[pa.Path] = types.Route{
 						Rule: rule,
 					}
 				}
@@ -313,19 +319,31 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
 	return configuration
 }
 
-func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.Ingress) string {
+func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.Ingress) (string, error) {
 	if len(pa.Path) == 0 {
-		return ""
+		return "", nil
 	}
 
 	ruleType := getStringValue(i.Annotations, annotationKubernetesRuleType, ruleTypePathPrefix)
 	rules := []string{ruleType + ":" + pa.Path}
 
+	var pathReplaceAnnotation string
+
 	if rewriteTarget := getStringValue(i.Annotations, annotationKubernetesRewriteTarget, ""); rewriteTarget != "" {
+		if pathReplaceAnnotation != "" {
+			return "", fmt.Errorf("rewrite-target must not be used together with annotation %q", pathReplaceAnnotation)
+		}
 		rules = append(rules, ruleTypeReplacePath+":"+rewriteTarget)
+		pathReplaceAnnotation = annotationKubernetesRewriteTarget
 	}
 
-	return strings.Join(rules, ";")
+	if rootPath := label.GetStringValue(i.Annotations, annotationKubernetesAppRoot, ""); rootPath != "" && pa.Path == "/" {
+		if pathReplaceAnnotation != "" {
+			return "", fmt.Errorf("app-root must not be used together with annotation %q", pathReplaceAnnotation)
+		}
+		rules = append(rules, ruleTypeReplacePath+":"+rootPath)
+	}
+	return strings.Join(rules, ";"), nil
 }
 
 func getRuleForHost(host string) string {
--- a/provider/kubernetes/annotations.go
+++ b/provider/kubernetes/annotations.go
@@ -29,6 +29,7 @@ const (
 	annotationKubernetesRateLimit                = "ingress.kubernetes.io/rate-limit"
 	annotationKubernetesErrorPages               = "ingress.kubernetes.io/error-pages"
 	annotationKubernetesBuffering                = "ingress.kubernetes.io/buffering"
+	annotationKubernetesAppRoot                  = "ingress.kubernetes.io/app-root"
 
 	annotationKubernetesSSLRedirect             = "ingress.kubernetes.io/ssl-redirect"
 	annotationKubernetesHSTSMaxAge              = "ingress.kubernetes.io/hsts-max-age"
--- a/provider/kubernetes/kubernetes_test.go
+++ b/provider/kubernetes/kubernetes_test.go
@@ -752,6 +752,32 @@ rateset:
 		),
 		buildIngress(
 			iNamespace("testing"),
+			iAnnotation(annotationKubernetesAppRoot, "/root2"),
+			iAnnotation(annotationKubernetesRewriteTarget, "/abc"),
+			iRules(
+				iRule(
+					iHost("root2"),
+					iPaths(
+						onePath(iPath("/"), iBackend("service2", intstr.FromInt(80))),
+					),
+				),
+			),
+		),
+		buildIngress(
+			iNamespace("testing"),
+			iAnnotation(annotationKubernetesRuleType, ruleTypeReplacePath),
+			iAnnotation(annotationKubernetesRewriteTarget, "/abc"),
+			iRules(
+				iRule(
+					iHost("root2"),
+					iPaths(
+						onePath(iPath("/"), iBackend("service2", intstr.FromInt(80))),
+					),
+				),
+			),
+		),
+		buildIngress(
+			iNamespace("testing"),
 			iAnnotation(annotationKubernetesIngressClass, "traefik"),
 			iAnnotation(annotationKubernetesCustomRequestHeaders, "Access-Control-Allow-Methods:POST,GET,OPTIONS || Content-type: application/json; charset=utf-8"),
 			iAnnotation(annotationKubernetesCustomResponseHeaders, "Access-Control-Allow-Methods:POST,GET,OPTIONS || Content-type: application/json; charset=utf-8"),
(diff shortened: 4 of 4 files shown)


ldez:
WDYT @containous/kubernetes ?

yue9944882:
@containous/kubernetes PTAL

yue9944882:
PTAL thx

errm:
I can see that this replicates the behaviour of the ngnix controller... but

It seems a bit strange that this would only work when the path is set to `/` without looking at the config the the nginx controller I would have assumed that it would be possible to set the path so that e.g. if path is `/foo` and app-root is `/bar` visiting `/foo` returns a 302 to `/foo/bar/`

I mean this does keep it consistent with how people expect this particular annotation to work so this is probably fine, and in any case I don't think implementing in this way precludes making this smarter in the future.

So um yeah I think this will be fine, but will require a note in the documentation that gets added to say this annotation only has any effect when the path is `/`

Other than that this seems reasonable in that it is only small change, follows the existing "defacto" standard and if not enabled has no way of effecting anything seriously.

So yeah with docs + tests I will be 👍

timoreimann:
What keeps us from implementing the feature such that it would also work when a path other than `/` is set (assuming that Nginx also works like that)?

errm:
Nginix only works when the path is / IDK why https://github.com/kubernetes/ingress-nginx/blob/53733483065a5619e8752c75dc2d23c628ddfc6f/rootfs/etc/nginx/template/nginx.tmpl#L636-L640

timoreimann:
@aledbf could you shed some light on why the `nginx.ingress.kubernetes.io/app-root` annotation only works when the root path (`/`) is found? Any particular reason why a path `/foo` is not replaced like `/new-root/foo`?

Thanks!

yue9944882:
In my case, this annotation helps when I have a home page serving at "a.com/foo" but I want to redirect all "a.com" root requests  to "a.com/foo", so that every time I want to browse this page I don't need to complete the full path manually which can be painful.. And at times, especially when the full path is rather long, typos will waste time.
Although this redirection is also performed by most of backend servers. But forcing it explicitly on gateway will surely be needed in some use cases.
@timoreimann @errm

yue9944882:
BTW, IDK where can I document this annotation? Should I add documentation into code/function comment?

timoreimann:
@yue9944882 Documentation of the annotation should go into `docs/configuration/backends/kubernetes.md`. If you think that a more elaborate description or example also makes sense, you can additionally extend the Kubernetes guide at docs/user-guide/kubernetes.md.

docs/configuration/backends/kubernetes.md
```diff
@@ -158,6 +158,7 @@ The following security annotations can be applied to the ingress object to add s
 | `ingress.kubernetes.io/public-key:VALUE`                 | Adds pinned HTST public key header.                                                                                                                                                                 |
 | `ingress.kubernetes.io/referrer-policy:VALUE`            | Adds referrer policy  header.                                                                                                                                                                       |
 | `ingress.kubernetes.io/is-development:false`             | This will cause the `AllowedHosts`, `SSLRedirect`, and `STSSeconds`/`STSIncludeSubdomains` options to be ignored during development.<br>When deploying to production, be sure to set this to false. |
+| `ingress.kubernetes.io/app-root:VALUE`             |  Explicitly make Ingress Controller redirects `/` requests to the defined path. |
```
timoreimann: As we're documenting the Ingress controller, can we simplify this to

_Redirects all requests for `/` to the defined path._

and also have another sentence to clearly describe what happens for non-root paths:

_Non-root paths will not be affected by this annotation and handled normally._

provider/kubernetes/kubernetes.go
```diff
@@ -385,6 +386,10 @@ func getRuleForPath(pa v1beta1.HTTPIngressPath, i *v1beta1.Ingress) string {
 		rules = append(rules, ruleTypeReplacePath+":"+rewriteTarget)
 	}
 
+	// HACK: When multiple "ReplacePath" present, only the one at the last order of rule string will work
```
timoreimann: This should be handled less implicitly: we should either log a warning when we observe two `ReplacePath` rules being used; or even log an error and ignore the Ingress in order to force the user to make a conscious decision.

I'm not sure if it's possible technically, but doesn't Traefik support applying a series of replacement rules already? In that case, we could let them resolve naturally. The downside is that it makes things more complicated to understand.

Whatever we do, we must document the behavior beyond a code comment. And the word _HACK_ should get dropped.

@dtomcej @errm thoughts?

yue9944882: [https://github.com/containous/traefik/blob/328be161d6e8cd04c1066d5884e4bf35f396b40f/server/rules.go#L97](https://github.com/containous/traefik/blob/328be161d6e8cd04c1066d5884e4bf35f396b40f/server/rules.go#L97)
[https://github.com/containous/traefik/blob/328be161d6e8cd04c1066d5884e4bf35f396b40f/server/rules.go#L184](https://github.com/containous/traefik/blob/328be161d6e8cd04c1066d5884e4bf35f396b40f/server/rules.go#L184)
I believe traefik doesn't support a series of replacement rules yet. And the last `ReplacePath` will override all the previous `ReplacePath` declaration because the rules is iterated as an array in the order of their occurrence in the rule-string. WDYT?

docs/configuration/backends/kubernetes.md
yue9944882: Make perfect sense. 👍

yue9944882:
@timoreimann  PTAL thx

yue9944882:
@ldez @timoreimann @errm BTW, this PR should be a feature-request. Seems like I should launch an issue or a discussion on Slack asking for further permission. Sorry for that.

provider/kubernetes/kubernetes.go
```diff
@@ -381,10 +382,18 @@ func getRuleForPath(pa v1beta1.HTTPIngressPath, i *v1beta1.Ingress) string {
 
 	rules := []string{ruleType + ":" + pa.Path}
 
+	rewrited := false
 	if rewriteTarget := i.Annotations[annotationKubernetesRewriteTarget]; rewriteTarget != "" {
 		rules = append(rules, ruleTypeReplacePath+":"+rewriteTarget)
+		rewrited = true
```
timoreimann: I just realized that we also have the case where the rule is explicitly set to `ReplacePath`. We can either handle that in this PR too or do a follow up.

yue9944882:
@timoreimann  As far as I'm concerned, if we just skip the offending Ingresses, it might cause some unexpected accident or loss especially in production environment, e.g. making one or more web-site managed in the same Ingress unavailable. 

How about throwing an error and ignore this annotation when at least one `ReplacePath` rule already exists?

timoreimann:
@yue9944882 my hypothesis is that failing hard and early in case the Ingress spec isn't consistent helps surfacing the problem more quickly than if we try to make the best out of it somehow. It's true that the latter could help making the Ingress work sometimes or even most of the time but may also cause it to fail in subtle, hard-to-spot corner cases. Personally I'd expect users to verify their Ingresses function properly prior to rolling them out to production systems, which seems easier to do when they don't work at all due in case some error exists.

Those are my two cents. Let's hear more voices before making a decision.

docs/configuration/backends/kubernetes.md
```diff
@@ -158,6 +158,7 @@ The following security annotations can be applied to the ingress object to add s
 | `ingress.kubernetes.io/public-key:VALUE`                 | Adds pinned HTST public key header.                                                                                                                                                                 |
 | `ingress.kubernetes.io/referrer-policy:VALUE`            | Adds referrer policy  header.                                                                                                                                                                       |
 | `ingress.kubernetes.io/is-development:false`             | This will cause the `AllowedHosts`, `SSLRedirect`, and `STSSeconds`/`STSIncludeSubdomains` options to be ignored during development.<br>When deploying to production, be sure to set this to false. |
+| `ingress.kubernetes.io/app-root:VALUE`             |  Redirects all requests for / to the defined path. |
```
timoreimann: Sorry, I meant to just drop the _on the other hand_ introduction. The remaining parts of the second sentence are good and should persist.

yue9944882:
@timoreimann I agree that failing-hard will help user to catch these error ASAP.

How about make failing-hard only impact on "/" root path? Or just prevent traefik from configuration-reloading?

I am just thinking skipping the offending Ingress and make configuration successfully reload is sort of "dangerous".

Let's hear more voice then. 👍

timoreimann:
@dtomcej any thoughts on this one, especially w.r.t. [the question](https://github.com/containous/traefik/pull/2522#pullrequestreview-82348180) of how we should treat Ingress objects that define multiple replacement rules?

yue9944882:
[https://github.com/kubernetes/ingress-nginx/issues/1813](https://github.com/kubernetes/ingress-nginx/issues/1813)

Seems related

dtomcej:
@timoreimann Unfortunately when you have users generate multiple replacement rules, someone will ultimately run into issues were the replacements do not occur the way they want to (due to the structure of the underlying code).

I think that it sucks, and we should deal with it when it comes up. If it does become an issue in the future, we can document what the order of replacements is, but until then, I don't think that making a mountain out of a molehill is required.

I think that if request replacements don't function properly, the results should be fairly obvious, and easy to debug using the debug log for backend requests. I would allow the ingress creation, and potentially making an info log note that duplicate replacement effects were detected.

yue9944882:
@dtomcej Outputting debug logs for backend requests may need a lot more changes in the code, and wouldn't this be better if we solve these rule-violation together by managing them in server/rule pkgs instead of only for kubernetes provider?  

@timoreimann  I've changed my mind and make it fail-hard when app-root violates with other `replacePath`. When app-root fails because of multi-replace-path, traefik will stop reloading configuration from ingress and throw an error log to prevent it from running into a unknown status. WDYT?

mmatur:
Could you have another look on it @containous/kubernetes please ?

provider/kubernetes/kubernetes.go
```diff
@@ -367,23 +372,35 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
...
+	pathReplaced := false
+	if ruleType == ruleTypeReplacePath {
+		pathReplaced = true
+	}
+
 	rules := []string{ruleType + ":" + pa.Path}
 
 	if rewriteTarget := i.Annotations[annotationKubernetesRewriteTarget]; rewriteTarget != "" {
 		rules = append(rules, ruleTypeReplacePath+":"+rewriteTarget)
+		pathReplaced = true
```
timoreimann: Shouldn't we return an error here already if `pathReplaced` is true at this point?

provider/kubernetes/kubernetes.go
```diff
@@ -367,23 +372,35 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
...
-		return ""
+		return "", nil
 	}
 
 	ruleType := i.Annotations[label.TraefikFrontendRuleType]
 	if ruleType == "" {
 		ruleType = ruleTypePathPrefix
 	}
 
+	pathReplaced := false
```
timoreimann: Instead of just a boolean, how about using a string denoting the annotation that we found a replace rule in previously? That way, we can clearly signal to the user which other instance of the replace rule was violating.

provider/kubernetes/kubernetes.go
```diff
@@ -367,23 +372,35 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
...
 
 	if rewriteTarget := i.Annotations[annotationKubernetesRewriteTarget]; rewriteTarget != "" {
 		rules = append(rules, ruleTypeReplacePath+":"+rewriteTarget)
+		pathReplaced = true
 	}
 
-	return strings.Join(rules, ";")
+	if rootPath := label.GetStringValue(i.Annotations, annotationKubernetesAppRoot, ""); rootPath != "" && pa.Path == "/" {
+		if pathReplaced {
+			return "", fmt.Errorf("app-root shouldn't be use together with other rules of replacepath type")
```
timoreimann: Can we phrase this more strongly by replacing _shouldn't_ with _must not_? The former gives some kind of indication that using multiple rules may still be acceptable, which it isn't.

yue9944882:
PTAL and manually squashed these commits. @timoreimann

docs/configuration/backends/kubernetes.md
```diff
@@ -161,6 +161,7 @@ The following security annotations are applicable on the Ingress object:
 | `ingress.kubernetes.io/public-key:VALUE`                 | Adds pinned HTST public key header.                                                                                                                                                                 |
 | `ingress.kubernetes.io/referrer-policy:VALUE`            | Adds referrer policy  header.                                                                                                                                                                       |
 | `ingress.kubernetes.io/is-development:false`             | This will cause the `AllowedHosts`, `SSLRedirect`, and `STSSeconds`/`STSIncludeSubdomains` options to be ignored during development.<br>When deploying to production, be sure to set this to false. |
+| `ingress.kubernetes.io/app-root:VALUE`             |  Redirects all requests for / to the defined path. Non-root paths will not be affected by this annotation and handled normally. |
```
ldez: this is not the right place to add this documentation, please move this in the section above.

provider/kubernetes/kubernetes.go
```diff
@@ -375,23 +380,38 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
...
 	if rewriteTarget := i.Annotations[annotationKubernetesRewriteTarget]; rewriteTarget != "" {
+		if pathReplaceAnnotation != "" {
+			return "", fmt.Errorf("rewrite-target conflicts with declared rule replacepath from annotation: %s", pathReplaceAnnotation)
+		}
 		rules = append(rules, ruleTypeReplacePath+":"+rewriteTarget)
+		pathReplaceAnnotation = annotationKubernetesRewriteTarget
 	}
 
-	return strings.Join(rules, ";")
+	if rootPath := label.GetStringValue(i.Annotations, annotationKubernetesAppRoot, ""); rootPath != "" && pa.Path == "/" {
```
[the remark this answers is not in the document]
timoreimann: Nevermind -- I missed there are two different variable names.

yue9944882:
@timoreimann @ldez PTAL

docs/configuration/backends/kubernetes.md
```diff
@@ -118,6 +118,7 @@ The following general annotations are applicable on the Ingress object:
     Replaces each matched Ingress path with the specified one, and adds the old path to the `X-Replaced-Path` header.
 - `ingress.kubernetes.io/whitelist-source-range: "1.2.3.0/24, fe80::/16"`
     A comma-separated list of IP ranges permitted for access. all source IPs are permitted if the list is empty or a single range is ill-formatted.
+-  `ingress.kubernetes.io/app-root: /index.html`  Redirects all requests for / to the defined path. Non-root paths will not be affected by this annotation and handled normally. This annotation may not be combined with the ReplacePath rule type any other annotation leveraging that rule type. Trying to do so leads to an error and the corresponding Ingress object being ignored.
```
timoreimann: There's an article missing: Please replace

_This annotation may not be combined with the ReplacePath rule type any other annotation leveraging that rule type._

by (**emphasis** by me):

_This annotation may not be combined with the ReplacePath rule type **or** any other annotation leveraging that rule type._

provider/kubernetes/kubernetes.go
```diff
@@ -375,23 +380,38 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
...
+			return "", fmt.Errorf("rewrite-target must not be used together with annotation %q", pathReplaceAnnotation)
+		}
 		rules = append(rules, ruleTypeReplacePath+":"+rewriteTarget)
+		pathReplaceAnnotation = annotationKubernetesRewriteTarget
 	}
 
-	return strings.Join(rules, ";")
+	if rootPath := label.GetStringValue(i.Annotations, annotationKubernetesAppRoot, ""); rootPath != "" && pa.Path == "/" {
+		if pathReplaceAnnotation != "" {
+			return "", fmt.Errorf("app-root must not be used together with annotation: %q", pathReplaceAnnotation)
```
timoreimann: Nit-pick level 11: Please remove the colon after _annotation_ to be consistent with the other error message. :-)

yue9944882:
@ldez @timoreimann  PTAL

provider/kubernetes/kubernetes.go
```diff
@@ -219,8 +219,13 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 					}
 				}
 
-				if rule := getRuleForPath(pa, i); rule != "" {
-					templateObjects.Frontends[baseName].Routes[pa.Path] = types.Route{
+				rule, err := getRuleForPath(pa, i)
+				if err != nil {
+					log.Errorf("fail to get rule from ingress path: %s", err.Error())
```
timoreimann: It suffices to use `err` instead of `err.Error()`.

provider/kubernetes/kubernetes.go
```diff
@@ -219,8 +219,13 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					return nil, err
```
timoreimann: Returning an error would cause all Ingress processing to end prematurely, leaving the old configuration active. Thus, a single misconfigured Ingress could bring everything to a halt.

What we need to do instead is log an error and skip the offending Ingress.

provider/kubernetes/kubernetes.go
```diff
@@ -219,8 +220,13 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 					}
 				}
 
-				if rule := getRuleForPath(pa, i); rule != "" {
-					templateObjects.Frontends[baseName].Routes[pa.Path] = types.Route{
+				rule, err := getRuleForPath(pa, i)
+				if err != nil {
+					log.Errorf("fail to get rule from ingress path: %s", err)
+					continue NEXT_INGRESS
```
yue9944882: How do you think about using a continue statement with label to skip the ingress?

yue9944882:
@timoreimann  PTAL

provider/kubernetes/kubernetes.go
timoreimann: Hmm, this looks correct but I'm not entirely sure. I'd rather prefer to use our existing approach of a bare `continue`.

We need to invest some time into the Kubernetes provider in order to simplify our highly nested structure. Once we do that, I believe a more elegant way to skip Ingresses not relying on continuing the right for loop will emerge naturally.

timoreimann: I think you need to invoke `delete(templateObjects.Frontends, r.Host+pa.Path)` before skipping.

yue9944882:
@timoreimann  PTAL thx

provider/kubernetes/kubernetes.go
```diff
@@ -219,8 +219,14 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 					}
 				}
 
-				if rule := getRuleForPath(pa, i); rule != "" {
-					templateObjects.Frontends[baseName].Routes[pa.Path] = types.Route{
+				rule, err := getRuleForPath(pa, i)
+				if err != nil {
+					log.Errorf("fail to get rule from ingress path: %s", err)
```
timoreimann: Please update the log message as following:

```go
log.Errorf("Failed to get rule for ingress %s/%s: %s", i.Namespace, i.Name, err)
```
(grammar/style of the first word + inclusion of the Ingress object's namespace and name)

timoreimann: My code coverage tool tells me that this line is not tested. Could you check please?

provider/kubernetes/kubernetes.go
```diff
@@ -306,19 +312,31 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
...
 	}
 
 	ruleType := getStringValue(i.Annotations, annotationKubernetesRuleType, ruleTypePathPrefix)
 	rules := []string{ruleType + ":" + pa.Path}
 
+	var pathReplaceAnnotation string
+
 	if rewriteTarget := getStringValue(i.Annotations, annotationKubernetesRewriteTarget, ""); rewriteTarget != "" {
+		if pathReplaceAnnotation != "" {
+			return "", fmt.Errorf("rewrite-target must not be used together with annotation %q", pathReplaceAnnotation)
```
timoreimann: I failed to see it in my previous review round, but this one seems to be missing code coverage too.

yue9944882: @timoreimann Oops, just realize that this line is unreachable.

yue9944882:
@timoreimann @ldez PTAL thx

PR #2534, opened 2017-12-06
Use default entryPoints when certificates are added with no entryPoints.

nmengin:
### What does this PR do?

The PR allows introducing a mechanism _by default_ for all the providers which add  TLScertificates dynamically.
If no entryPoints are given, the certificates are added to all of the `defaultEntryPoints` (from Træfik global configuration) wich have a TLS configuration.

### Motivation

This PR helps to create the certificates management from K8s secret (#2439)

### More

- [x] Added/updated tests

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -435,8 +435,12 @@ func (s *Server) loadConfiguration(configMsg types.ConfigMessage) {
 	if err == nil {
 		for newServerEntryPointName, newServerEntryPoint := range newServerEntryPoints {
 			s.serverEntryPoints[newServerEntryPointName].httpRouter.UpdateHandler(newServerEntryPoint.httpRouter.GetHandler())
-			if &newServerEntryPoint.certs != nil {
-				s.serverEntryPoints[newServerEntryPointName].certs.Set(newServerEntryPoint.certs.Get())
+			if newServerEntryPoint.certs.Get() != nil {
+				if s.globalConfiguration.EntryPoints[newServerEntryPointName].TLS == nil {
+					log.Debugf("Certificates not added to non-TLS entryPoint %s.", newServerEntryPointName)
+				} else {
+					s.serverEntryPoints[newServerEntryPointName].certs.Set(newServerEntryPoint.certs.Get())
+				}
 			}
 			log.Infof("Server configuration reloaded on %s", s.serverEntryPoints[newServerEntryPointName].httpServer.Addr)
 		}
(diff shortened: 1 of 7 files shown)


tls/tls.go
```diff
@@ -87,11 +89,20 @@ func (r *RootCAs) Type() string {
...
 		epConfiguration = make(map[string]*DomainsCertificates)
 	}
 	for _, conf := range configurations {
+		if conf.EntryPoints == nil || len(conf.EntryPoints) == 0 {
+			certName := conf.Certificate.CertFile.String()
+			// Keep only 50 characters to generate a readable log
+			if len(certName) > 50 {
+				certName = certName[0:50]
+			}
+			log.Debugf("No entryPoint is defined to add the certificate %s, it will be added to the default entryPoints: %s", certName, strings.Join(defaultEntryPoints, ","))
```
ldez:
```go
if log.GetLevel() >= logrus.DebugLevel {
	certName := conf.Certificate.CertFile.String()
	// Keep only 50 characters to generate a readable log
	if len(certName) > 50 {
		certName = certName[:50]
	}
	log.Debugf("No entryPoint is defined to add the certificate %s, it will be added to the default entryPoints: %s", certName, strings.Join(defaultEntryPoints, ","))
}
```

tls/tls.go
```diff
@@ -87,11 +90,23 @@ func (r *RootCAs) Type() string {
...
 		epConfiguration = make(map[string]*DomainsCertificates)
 	}
 	for _, conf := range configurations {
+		if conf.EntryPoints == nil || len(conf.EntryPoints) == 0 {
+			if log.GetLevel() >= logrus.DebugLevel {
+				var certName string
+				if conf.Certificate.CertFile.Path() {
+					certName = conf.Certificate.CertFile.String()
+				} else {
+					certName = strings.TrimPrefix(conf.Certificate.CertFile.String(), "-----BEGIN CERTIFICATE-----\n")[:50]
```
ldez: if the CertFile content if empty or malformed, this can panicing.

tls/certificate.go
```diff
@@ -67,6 +67,12 @@ func (f FileOrContent) String() string {
 	return string(f)
 }
 
+// Path returns true if the FileOrContent is a file path, otherwise returns false
+func (f FileOrContent) Path() bool {
```
Juliens: if it's a boolean, `isPath` or you must return a string with filepath if you have one ?

server/server.go
```diff
@@ -435,8 +435,12 @@ func (s *Server) loadConfiguration(configMsg types.ConfigMessage) {
 	if err == nil {
 		for newServerEntryPointName, newServerEntryPoint := range newServerEntryPoints {
 			s.serverEntryPoints[newServerEntryPointName].httpRouter.UpdateHandler(newServerEntryPoint.httpRouter.GetHandler())
-			if &newServerEntryPoint.certs != nil {
-				s.serverEntryPoints[newServerEntryPointName].certs.Set(newServerEntryPoint.certs.Get())
+			if newServerEntryPoint.certs.Get() != nil {
```
Juliens: certs can be nil

ldez: It's not possible because it's a struct and not a pointer

PR #2538, opened 2017-12-07
Fix broken links and improve ResponseCodeRatio() description

mvasin:
Please check if I understood `ResponseCodeRatio()` range logic right.

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -585,7 +585,7 @@ traefik [command] [--flag=flag_argument]
 List of Træfik available commands with description :
 
 - `version` : Print version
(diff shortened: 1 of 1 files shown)


mvasin:
Each correction is too minor for a PR, but all together seem to be worth it.

BTW, as soon as I occasionally found three broken links on single page, don't you think it makes sense to employ an HTML linter?  Fixing links here and there manually is boring, it's better to make it a part of the test suite.

mvasin:
Oh, I got it, this [nerdy brackets notation](https://stackoverflow.com/a/4396303/5274538).

PR #2560, opened 2017-12-12
Fix RawPath handling in addPrefix

risdenk:
This is an extension of PR #2382 for addPrefix.

This PR fixes forwarding of URL encoded characters to the backend servers.

Below is a copy straight from PR #2382 :

Given that Traefik receives a request path in the form: `/some/a%2Fb` it should forward the same path to the backend. This is the case when using e.g. the rule type `PathPrefix`, but as soon as the rule type is either `AddPrefix` the path was forwarded in the decoded form like: `/some/a/b`.

The reason why this solution fixes our problem is that go uses the `String()` method of the URL object when it constructs the request. The docs state for this:

```go
// String reassembles the URL into a valid URL string.
// The general form of the result is one of:
//
//	scheme:opaque?query#fragment
//	scheme://userinfo@host/path?query#fragment
//
// If u.Opaque is non-empty, String uses the first form;
// otherwise it uses the second form.
// To obtain the path, String uses u.EscapedPath().
```

For us the relevant part is with `u.EscapedPath()` and there the docs state:

```go
// EscapedPath returns the escaped form of u.Path.
// In general there are multiple possible escaped forms of any path.
// EscapedPath returns u.RawPath when it is a valid escaping of u.Path.
// Otherwise EscapedPath ignores u.RawPath and computes an escaped
// form on its own.
// The String and RequestURI methods use EscapedPath to construct
// their results.
// In general, code should call EscapedPath instead of
// reading u.RawPath directly.
```

To understand the problem, note the sentence: `EscapedPath` returns `u.RawPath` when it is a valid escaping of `u.Path`. By stripping only the `Path` and not the `RawPath` we violated this statement and therefore the http package decided to use the `Path`.

### DIFF
--- a/middlewares/addPrefix.go
+++ b/middlewares/addPrefix.go
@@ -12,6 +12,9 @@ type AddPrefix struct {
 
 func (s *AddPrefix) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	r.URL.Path = s.Prefix + r.URL.Path
+	if r.URL.RawPath != "" {
+		r.URL.RawPath = s.Prefix + r.URL.RawPath
+	}
 	r.RequestURI = r.URL.RequestURI()
 	s.Handler.ServeHTTP(w, r)
 }
--- a/middlewares/addPrefix_test.go
+++ b/middlewares/addPrefix_test.go
@@ -9,21 +9,56 @@ import (
 )
 
 func TestAddPrefix(t *testing.T) {
-
-	path := "/bar"
-	prefix := "/foo"
-
-	var expectedPath string
-	handler := &AddPrefix{
-		Prefix: prefix,
-		Handler: http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			expectedPath = r.URL.Path
-		}),
+	tests := []struct {
+		desc            string
+		prefix          string
+		path            string
+		expectedPath    string
+		expectedRawPath string
+	}{
+		{
+			desc:         "regular path",
+			prefix:       "/a",
+			path:         "/b",
+			expectedPath: "/a/b",
+		},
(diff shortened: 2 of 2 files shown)


risdenk:
The file `addPrefix_test.go` was refactored to match the style of `stripPrefix_test.go` for the added test case of %2F encoded paths.

risdenk:
Looks like the semaphoreci failed with the machine shutting off?

**make pull-images**
```
1fab753cf6d1: Pull complete 
Digest: sha256:28488b5daf19a9a591565fc0d1ad502c2a24b8ccc0d4579d8bf12d9796a71a27
Status: Downloaded newer image for yaman/timeout:latest
make: *** [pull-images] Error 123
```

**sudo -E apt-get -yq update**
```
Broadcast message from root@semaphore-1711

	(unknown) at 4:13 ...

The system is going down for power off NOW!

Ign http://download.rethinkdb.com trusty/main Translation-en_US
Ign http://download.rethinkdb.com trusty/main Translation-en
Ign http://download.rethinkdb.com trusty/main Translation-en_US.UTF-8
Fetched 6,572 kB in 2min 25s (45.3 kB/s)
Reading package lists...
```

ldez:
Could you fix that:
```
---> Making bundle: validate-gofmt (in .)
These files are not properly gofmt'd:
 - middlewares/addPrefix.go
 - middlewares/addPrefix_test.go

Please reformat the above files using "gofmt -s -w" and commit the result.
```

marco-jantke:
Thanks for the extension @risdenk. The change LGTM apart of the formatting errors :D

risdenk:
@ldez - sorry about the formatting. Pushed a fix.

PR #2562, opened 2017-12-12
Grammar

geraldcroes:
### What does this PR do?
Fixes minor grammar issues.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -22,7 +22,7 @@ If you want your users to access some of your microservices from the Internet, y
 - path `domain.com/web` will point the microservice `web` in your private network
 - domain `backoffice.domain.com` will point the microservices `backoffice` in your private network, load-balancing between your multiple instances
 
-But a microservices architecture is dynamic... Services are added, removed, killed or upgraded often, eventually several times a day.
+Microservices are often deployed in dynamic environments where services are added, removed, killed, upgraded or scaled many times a day.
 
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -22,7 +22,7 @@ If you want your users to access some of your microservices from the Internet, y
 - path `domain.com/web` will point the microservice `web` in your private network
 - domain `backoffice.domain.com` will point the microservices `backoffice` in your private network, load-balancing between your multiple instances
 
-But a microservices architecture is dynamic... Services are added, removed, killed or upgraded often, eventually several times a day.
+But a microservices architecture is dynamic... Services are often added, removed, killed, or upgraded, eventually several times a day.
```
errm: Not the fault of this change, but I am not sure about this sentence at all...

> But a microservices architecture is dynamic

Microservices or the architecture is not really the issue here, we are thinking about environments where instances of a service are launched dynamically.

geraldcroes: Maybe something like: 
>But microservices architecture environments are dynamic: Services are often added, 
>removed, killed, or upgraded -- eventually several times a day.

errm: How about:

> Microservices are often deployed dynamic environments where services are added, removed, killed, upgraded or scaled many times a day.

geraldcroes: I think that there's a typo: deployed /in/?

errm: yes, you are correct

PR #2567, opened 2017-12-14
extend metrics and rebuild prometheus exporting logic

### LINKED ISSUE #1759: Prometheus request metrics tracked twice

*Bug*

### What did you do?

Setup a Traefik instance that provides Prometheus metrics.

Correct recording of http requests / http request durations metrics with the `service` label set to `backend1`.

### What did you see instead?

Doubled recording of http requests / http request durations metrics. One time with the `service` label `http` (entrypoint) + `backend1`.

```
Version:      dev
Codename:     cheddar
Go version:   go1.8.3
OS/Arch:      darwin/amd64
```

```toml
# traefik.toml
logLevel = "DEBUG"
defaultEntryPoints = ["http"]

[entryPoints]
  [entryPoints.http]
  address = ":8080"

[file]

# rules
[backends]
  [backends.backend1]
     url = "http://localhost:8081"
     [backends.backend1.servers.server2]
     url = "http://localhost:8082"
     [backends.backend1.servers.server3]
     url = "http://localhost:8083"

[frontends]
  [frontends.frontend1]
  backend = "backend1"
    [frontends.frontend1.routes.test_1]
    rule = "Path:/ok"

[retry]
attempts = 3

[web]
address = ":8000"
  [web.metrics.prometheus]
    Buckets=[0.1,0.3,1.2,5.0]
```

### Problem Description

At the moment the Traefik metrics middleware is included twice in the middleware chain. One time the metrics are included in the `startHTTPServers` [here](https://github.com/containous/traefik/blob/2a578748fdb349bf92fb223c24397005b04cdc99/server/server.go#L195) and a second time when loading the config [here](https://github.com/containous/traefik/blob/2a578748fdb349bf92fb223c24397005b04cdc99/server/server.go#L786).

This leads to the following generated metrics when requesting the `/metrics` endpoint.

```
# HELP traefik_backend_retries_total How many request retries happened in total.
# TYPE traefik_backend_retries_total counter
traefik_backend_retries_total{service="backend1"} 14

# HELP traefik_request_duration_seconds How long it took to process the request.
# TYPE traefik_request_duration_seconds histogram
traefik_request_duration_seconds_bucket{service="backend1",le="0.1"} 7
traefik_request_duration_seconds_bucket{service="backend1",le="0.3"} 7
traefik_request_duration_seconds_bucket{service="backend1",le="1.2"} 7
traefik_request_duration_seconds_bucket{service="backend1",le="5"} 7
traefik_request_duration_seconds_bucket{service="backend1",le="+Inf"} 7
traefik_request_duration_seconds_sum{service="backend1"} 0.036858719
traefik_request_duration_seconds_count{service="backend1"} 7
traefik_request_duration_seconds_bucket{service="http",le="0.1"} 7
traefik_request_duration_seconds_bucket{service="http",le="0.3"} 7
traefik_request_duration_seconds_bucket{service="http",le="1.2"} 7
traefik_request_duration_seconds_bucket{service="http",le="5"} 7
traefik_request_duration_seconds_bucket{service="http",le="+Inf"} 7
traefik_request_duration_seconds_sum{service="http"} 0.037084179
traefik_request_duration_seconds_count{service="http"} 7

# HELP traefik_requests_total How many HTTP requests processed, partitioned by status code and method.
# TYPE traefik_requests_total counter
traefik_requests_total{code="200",method="GET",service="backend1"} 7
traefik_requests_total{code="200",method="GET",service="http"} 7
```

This is problematic, because effectively this means that when you aggregate the metrics in your Prometheus queries, without considering the labels, you will end up with twice as many requests as actually happen. Therefore the proper fix should be to remove the integration when starting the http servers. As one point to consider, this would mean that redirects are actually not tracked anymore. When tracking of the redirects is desired, I guess we should create a new metrics, to handle this cleanly.

Can you confirm that this is a bug for you and maybe give a recommendation on how to go for the fix?

### LINKED ISSUE #2349: Websocket metrics
Feature

### Details
In the metrics provided we have visibility to all http requests.  Which is great to see that kind of traffic coming in.  But we have no way of seeing how many websockets are open.

I have traefik in front of several servers and one of the primary data channels is websockets.  Which works great.  I just have no insight into the websockets currently being proxied through by traefik.

Traefik as a deployment in kubernetes acting as my ingress controller.

marco-jantke:
I know that this PR is large but it grew a long time in our fork now and the changes depend on each other which makes it very hard to split those up into several commits.
I am happy to assist in any way it helps you to make a review, also a e.g. a peer-review via video chat is completely fine for me.
Just let me know how I can help getting this merged.
To make it a bit better I added an extensive commit description that I quoted below to explain the different measures.
To wet the appetite why this is useful here a short list of which metrics Traefik supports now for Prometheus and it is very easy for other providers to move along!
To get the other providers on the same feature level (not required to work right now, see commit description) I shoutout to @aantono and @adityacs who integrated those to extend the implementation of those providers.

Which metrics exist now with small samples (numbers are totally random):

- `traefik_config_reloads_total`: useful to see how often config changes happen
- `traefik_config_reloads_failure_total`: useful to see failing config reloads and alert on those
- `traefik_config_last_reload_success`: useful to see when the last reload has happened
- `traefik_config_last_reload_failure`: useful to see when the last config reload has happened
- `traefik_entrypoint_requests_total`: same as before but on entrypoint level now
- `traefik_entrypoint_request_duration_seconds`: same as before but on entrypoint level now
- `traefik_entrypoint_open_connections`: useful to see how many open connections exist currently. It's split up even by protocol, so you could understand e.g. there 1k open websocket connections, 300 open SSE connections and currently 1.5k http requests in flight
- `traefik_backend_requests_total`: same as before but on backend level
- `traefik_backend_request_duration_seconds`: same as before but on backend level
- `traefik_backend_retries_total`: same as before but on backend level
- `traefik_backend_open_connections`: same as in entrypoint description but split up for each backend
- `traefik_backend_server_up`: you can see how many backend servers exist and are healthy of a specific `backend.` useful to alert when e.g. 80% of your backend servers become unhealthy.

Additionally, this PR tackles #2349 and #1759.

**Commit message with detailed explanations**

This commit extends the metrics Registry interface and extends it's usage in order to get more insightful metrics. It also renames Prometheus metrics and the service label (to either backend or entrypoint) in order to fix #1759.

In order to extend the Registry interface the MultiRegistry is used so that single metric providers do not necessarily have to implement all of the metrics, but only a subset.
At the moment only the Prometheus implementation supports all new metrics, but extension of the other providers should be straightforward and can be done afterwards.

Additionally the Prometheus exporting logic was adapted.
The problem with it before was that the Prometheus client library is stateful and samples all tracked metrics on the client.
This is a problem given the things you are tracking are dynamic.
We have a metric traefik_backend_server_up which can be used to alert given too many servers of a backend become unhealthy.
Now when Traefik learns about a new backend it will create the concrete metric with the labels for this backend.
The problem is now that given the backend is undeployed, Traefik will still export the metric of the old backend because it holds the Prometheus metric state and it will only be gone after a restart of Traefik. 
This leads to wrong results when you calculate e.g. how many of your backends are healthy / all backends and thus breaks alerting.
To fix this one has to implement a custom logic for the metric sampling.
I got a confirmation for this approach on the Prometheus users google group from Prometheus maintainers.
To fix the problem I implemented a configuration generation logic so that metrics that belong to a dynamic configuration (e.g. backends, entrypoints) get removed  again eventually given they belong to an old configuration and are not tracked recently.
https://groups.google.com/forum/#!topic/prometheus-users/EbumSJYsJyc

This PR also splits up the metric collection of requests, request  durations and retries into separate metrics for entrypoints and backends.
This will fix #1759. Before there was only one metric `traefik_requests_total` which was tracked at the entrypoint level with the label `service="$entrypointname"` and at the backend level with `service="$backendname"`.
So at each request two `traefik_requests_total` metrics were trackend and so a calculation of e.g. `sum(traefik_requests_total)` would yield twice the requests that actually happened.
This is fixed by splitting up the metric into two.
For other metric providers this problem was the same and as I did not adapt those yet, they will only count the requests now against the backends.
An extension of them should happen afterwards.
  

Fixes  #2349 #1759

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -122,7 +122,7 @@ func NewServer(globalConfiguration configuration.GlobalConfiguration) *Server {
 		server.tracingMiddleware.Setup()
 	}
 
-	server.metricsRegistry = server.registerMetricClients(globalConfiguration.Metrics)
+	server.metricsRegistry = registerMetricClients(globalConfiguration.Metrics)
 
 	if globalConfiguration.Cluster != nil {
 		// leadership creation if cluster mode
@@ -1481,7 +1481,7 @@ func configureBackends(backends map[string]*types.Backend) {
 	}
 }
 
-func (*Server) registerMetricClients(metricsConfig *types.Metrics) metrics.Registry {
+func registerMetricClients(metricsConfig *types.Metrics) metrics.Registry {
 	if metricsConfig == nil {
 		return metrics.NewVoidRegistry()
 	}
--- a/server/server.go
+++ b/server/server.go
@@ -122,10 +122,7 @@ func NewServer(globalConfiguration configuration.GlobalConfiguration) *Server {
 		server.tracingMiddleware.Setup()
 	}
 
-	server.metricsRegistry = metrics.NewVoidRegistry()
-	if globalConfiguration.Metrics != nil {
-		server.registerMetricClients(globalConfiguration.Metrics)
-	}
+	server.metricsRegistry = server.registerMetricClients(globalConfiguration.Metrics)
 
 	if globalConfiguration.Cluster != nil {
 		// leadership creation if cluster mode
@@ -304,7 +301,7 @@ func (s *Server) setupServerEntryPoint(newServerEntryPointName string, newServer
 		serverMiddlewares = append(serverMiddlewares, s.accessLoggerMiddleware)
 	}
 	if s.metricsRegistry.IsEnabled() {
-		serverMiddlewares = append(serverMiddlewares, middlewares.NewMetricsWrapper(s.metricsRegistry, newServerEntryPointName))
+		serverMiddlewares = append(serverMiddlewares, middlewares.NewEntryPointMetricsMiddleware(s.metricsRegistry, newServerEntryPointName))
 	}
 	if s.globalConfiguration.API != nil {
 		if s.globalConfiguration.API.Stats == nil {
@@ -449,8 +446,10 @@ func (s *Server) loadConfiguration(configMsg types.ConfigMessage) {
 	}
 	newConfigurations[configMsg.ProviderName] = configMsg.Configuration
 
+	s.metricsRegistry.ConfigReloadsCounter().Add(1)
 	newServerEntryPoints, err := s.loadConfig(newConfigurations, s.globalConfiguration)
 	if err == nil {
+		s.metricsRegistry.LastConfigReloadSuccessGauge().Set(float64(time.Now().Unix()))
 		for newServerEntryPointName, newServerEntryPoint := range newServerEntryPoints {
 			s.serverEntryPoints[newServerEntryPointName].httpRouter.UpdateHandler(newServerEntryPoint.httpRouter.GetHandler())
 			if s.globalConfiguration.EntryPoints[newServerEntryPointName].TLS == nil {
@@ -465,6 +464,8 @@ func (s *Server) loadConfiguration(configMsg types.ConfigMessage) {
 		s.currentConfigurations.Set(newConfigurations)
 		s.postLoadConfiguration()
 	} else {
+		s.metricsRegistry.ConfigReloadsFailureCounter().Add(1)
+		s.metricsRegistry.LastConfigReloadFailureGauge().Set(float64(time.Now().Unix()))
 		log.Error("Error loading new configuration, aborted ", err)
 	}
 }
@@ -501,6 +502,8 @@ func (s *serverEntryPoint) getCertificate(clientHello *tls.ClientHelloInfo) (*tl
 }
 
 func (s *Server) postLoadConfiguration() {
+	metrics.OnConfigurationUpdate()
+
 	if s.globalConfiguration.ACME == nil {
 		return
 	}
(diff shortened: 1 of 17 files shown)


aantono:
Great addition... I'll add the DataDog and StatsD counters as soon as this PR gets merged!

adityacs:
Will update influx metrics as well once this PR is merged.

ldez:
@marco-jantke could you rebase?

marco-jantke:
@ldez done.

marco-jantke:
Can we get this moving?

server/server.go
```diff
@@ -1470,9 +1474,12 @@ func configureBackends(backends map[string]*types.Backend) {
 	}
 }
 
-func (s *Server) registerMetricClients(metricsConfig *types.Metrics) {
-	var registries []metrics.Registry
+func (*Server) registerMetricClients(metricsConfig *types.Metrics) metrics.Registry {
```
Juliens: As you don't use receiver, you can use only a func instead of a method

marco-jantke: Good point, fixed.

ldez:
@aantono and @adityacs now it's up to you 😃

PR #2579, opened 2017-12-16
rancher.md: update default frontend rule

djtm:
Updates the rancher docs to the actual current behavior.


ldez:

The frontend rule in Rancher it's not the same as Docker:
https://github.com/containous/traefik/blob/f6520727a34dea4322cd43a0dcb424af49876c75/provider/rancher/config.go#L112-L113

Your change is invalid.

djtm:
Thanks. Better?

ldez:
I have fixed the problem in my PR #2601.

Thanks for the pointer 👍

PR #2590, opened 2017-12-18
Mesos: Use slave.PID.Host as task SlaveIP.

### LINKED ISSUE #2545: Mesos Backend Incorrectly Configures Hostname for Host IPSource

Bug.

When you setup the mesos backend, the generated config files results in
the `backend.url` field being set to something like `"http://:47675"` when you
set `IPSource` to host. 

Traefik uses mesosphere/mesos-dns@b47dc4c19f215e98da687b15b4c64e70f629bea5 to 
easily communicate with mesos, however it doesn't fully bootstrap the
configuration with something mesos-dns expects. When Traefik retrieves the 
list of running tasks from Mesos it manually sets `task.SlaveIP` to `slave.Hostname`.

https://github.com/containous/traefik/blob/c66d9de759b4bdebf8e170539eac1bcdd1295894/provider/mesos/config.go#L82-L86

The problem is `task.IP()` expects `task.SlaveIP` to be an actual IPv4 or IPv6, not
a hostname. As a result it tries to parse the hostname as an IP, but fails,
resulting in an empty string as the hostname. 

https://github.com/containous/traefik/blob/c66d9de759b4bdebf8e170539eac1bcdd1295894/provider/mesos/config.go#L207-L209

Traefik should either be using `slave.PID.Host` or a special case in `getHost` to 
properly get the hostname for the task.

### What did you do?

Set up Traefik to use the Mesos Backend

The host IP in the backend url.

### What did you see instead?

An empty string in the place of an IP address.

containous/traefik@c66d9de759b4bdebf8e170539eac1bcdd1295894

```toml
[entryPoints]
  [entryPoints.http]
  address = ":80"

[mesos]
endpoint = "zk://zoo1:2181/mesos"
watch = true
domain = "mesos"
filename = "/mnt/mesos/sandbox/mesos.tmpl"
IPSources = "host"
```

nemosupremo:
### What does this PR do?

This is the proper way to set the SlaveIP as done by the underlying `mesos-dns` helper library

### Motivation

Fixes #2545

### More

- [x] Added/updated tests

### DIFF
--- a/provider/mesos/mesos.go
+++ b/provider/mesos/mesos.go
@@ -408,7 +408,7 @@ func (p *Provider) taskRecords(sj state.State) []state.Task {
 		for _, task := range f.Tasks {
 			for _, slave := range sj.Slaves {
 				if task.SlaveID == slave.ID {
-					task.SlaveIP = slave.Hostname
+					task.SlaveIP = slave.PID.Host
 				}
 			}
 
--- a/provider/mesos/mesos_test.go
+++ b/provider/mesos/mesos_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/containous/traefik/log"
 	"github.com/containous/traefik/types"
+	"github.com/mesos/mesos-go/upid"
 	"github.com/mesosphere/mesos-dns/records/state"
 )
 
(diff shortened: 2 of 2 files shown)


timoreimann:
(Note: I don't think there's any documentation to update; the fix is more about an implementation detail to me.)

PR #2592, opened 2017-12-19
Add non regex pathPrefix

### LINKED ISSUE #2376: Excessive memory usage with k8s
I have been using the latest version and surprisingly the memory usage seems a lot higher than before. Currently 5GB on a single server which goes to 20GB eventually over two days. It really seems too much

a Possible bug

### What did you do?

Used traefik with kuberenetes with lots of ingress rules for different host names

Less memory usage

### What did you see instead?

an excessive amount of memory consumed by treafik

1.4.2

```toml
defaultEntryPoints = ["http"]
logLevel = "WARN"
MaxIdleConnsPerHost = 1000

[entryPoints]
  [entryPoints.http]
  address = ":80"
  compress = true
    [entryPoints.http.redirect]
      regex = "^http://www[.](.*)"
      replacement = "http://$1"

### PR DESCRIPTION
**Pull request by emilevauge**
### What does this PR do?

This PR adds non regex `PathPrefix` & `PathPrefixStrip` support.
When no `{` character is found in the given path, we now bypass the `gorilla/mux` regex implementation to use our lightweight implementation instead.

### Motivation

Fixes #2376

### More

- [x] Added/updated tests

### Additional Notes

The leading & trailing slash behavior has been tested to be the same as before.

### DIFF
--- a/server/rules.go
+++ b/server/rules.go
@@ -54,9 +54,30 @@ func (r *Rules) path(paths ...string) *mux.Route {
 func (r *Rules) pathPrefix(paths ...string) *mux.Route {
 	router := r.route.route.Subrouter()
 	for _, path := range paths {
-		router.PathPrefix(strings.TrimSpace(path))
+		buildPath(path, router)
 	}
 	return r.route.route
+}
+
+func buildPath(path string, router *mux.Router) {
+	cleanPath := strings.TrimSpace(path)
+	// {} are used to define a regex pattern in http://www.gorillatoolkit.org/pkg/mux.
+	// if we find a { in the path, that means we use regex, then the gorilla/mux implementation is chosen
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: server/rules.go [hunk 1]
```diff
@@ -54,11 +54,25 @@ func (r *Rules) path(paths ...string) *mux.Route {
 func (r *Rules) pathPrefix(paths ...string) *mux.Route {
 	router := r.route.route.Subrouter()
 	for _, path := range paths {
-		router.PathPrefix(strings.TrimSpace(path))
+		cleanPath := strings.TrimSpace(path)
+		if strings.Contains(cleanPath, "{") {
```
timoreimann: Can we leave a note here why `{` is important here?

emilevauge: Done :+1:

PR #2606, opened 2017-12-21
Add tests for TLS dynamic configuration in ETCD3

dahefanteng:
### What does this PR do?

When I delete all the tls cert/key pairs in my etcd(v3),the debug log and api(/api/providers) show  traefik has read configurations which has an empty tlsconfigure.But Actually it does't work.
After reading the code,I think traefik skip delete when the all the tlsconfigurations in etcd be deleted in a very short time. 

And I wrote some test for this.
### Motivation

when i write test case in my own project,I delete the tls cert/key pairs,expecting recevie the "TRAEFIK DEFAULT CERT" from traefik,but what i receive is still the deleted one after a very long wait.

### More

- [x] Added/updated tests

### DIFF
--- a/integration/etcd3_test.go
+++ b/integration/etcd3_test.go
@@ -588,9 +588,6 @@ func (s *Etcd3Suite) TestDeleteSNIDynamicTlsConfig(c *check.C) {
 	defer display(c)
 
 	// prepare to config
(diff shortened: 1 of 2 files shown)


dahefanteng:
@nmengin  I have rebase the PR and delete code in server.go

dahefanteng:
@nmengin  Really appreciate your patience,Thank you!

PR #2609, opened 2017-12-21
Send empty configuration from file provider

nmengin:
### What does this PR do?

This PR allows sending empty configuration from `file` provider which is taken in account by Træfik.

### Motivation

Træfik filters dynamic configurations which have backends, frontends and TLS configurations set to nil.
Providers have to send empty configuration to update correctly the information.

The `file` provider sent empty data if it listened to a `directory` but sent nil if it listened to a `fileName` changes.

The PR initializes correctly the dynamic configuration from a `fileName` : if the user delete all his configuration in a file, Træfik will be able to delete the backends, frontends anf TLS configurations.

### More

- [x] Added/updated tests

### DIFF
--- a/provider/file/file.go
+++ b/provider/file/file.go
@@ -127,9 +127,8 @@ func sendConfigToChannel(configurationChan chan<- types.ConfigMessage, configura
 
 func loadFileConfig(filename string) (*types.Configuration, error) {
 	configuration := &types.Configuration{
-		Frontends:        make(map[string]*types.Frontend),
-		Backends:         make(map[string]*types.Backend),
-		TLSConfiguration: make([]*tls.Configuration, 0),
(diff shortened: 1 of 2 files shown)


provider/file/file.go
```diff
@@ -126,7 +126,11 @@ func sendConfigToChannel(configurationChan chan<- types.ConfigMessage, configura
 }
 
 func loadFileConfig(filename string) (*types.Configuration, error) {
-	configuration := new(types.Configuration)
+	configuration := &types.Configuration{
+		Frontends:        make(map[string]*types.Frontend),
+		Backends:         make(map[string]*types.Backend),
+		TLSConfiguration: make([]*tls.Configuration, 0),
```
ldez: not needed because it's not a pointer on a slice.

PR #2613, opened 2017-12-21
Sticky Backend Handler

### LINKED ISSUE #1035: Sticky sessions based on ipv4/ipv6 hash
This is a feature request and not a bug.
The current and only method in traefik for sticky sessions is injecting cookies However, this is the not the best way.  The load balancer, should feature hashing based on say ipv4 or ipv6  hash as an option for sticky session. Since backends can appear, disappear, we may keep a table in ram for say 5 minutes or 10 minutes ( and setting to control session time), which  maintains   ip==backend  relationship  , if backend disappears, we assign a new backend.

### LINKED ISSUE #1207: Feature request: "sticky sessions" based on configured header

### LINKED ISSUE #2299: Encrypt Session Sticky cookie content
request a feature

### What did you do?
I enabled session sticky under docker.

The Traefik session sticky cookie content to be encrypted text.

### What did you see instead?
The cookie currently shows the internal uri with internal container IP in clear text, which is a security risk if deployed in internet.

```
1.4.0
```

```toml
```

```
(paste your output here)
```

caleblloyd:
### What does this PR do?

This PR adds a `StickyBackendHandler` middleware and rules-based or IP-based stickiness rules.  Stickiness rules are evaluated in the following order:

- Array of Rules Templates executed against the current `*http.Request`
- Client IP Address
- Cookie with optional Encryption

**Consistent Hashing**

Rules and IP sticky backends are assigned based off a Consistent Hash.  Each time the list of backends updates, sessions may be assigned to a new backend.  This allows multiple Traefik instances with the same source of backends to all route sticky requests consistently.

**Rules**:

Rules are an array of Go templates that execute against the current `*http.Request`.  They are evaluated in-order until template resolves to a non-blank value.  Templates can include [sprig functions](http://masterminds.github.io/sprig/).

```go
// Consistent Hash the value of the X-Backend header, if it exists
&types.Stickiness{
    Rules: []string{
        `{{index .Header "X-Backend" | join ""}}`,
    },
}

// Consistent Hash the /isbin/{id} path, if it exists
// These 2 URLs will produce the same consistent hash
// http://localhost/isbn/12345/author
// http://localhost/isbn/12345/description
&types.Stickiness{
    Rules: []string{
        `{{regexFind "^/isbn/\\d+" .URL.Path}}`,
    },
},
```

**Client IP**:

Performs a consistent hash of the leftmost entry in the `X-Forwarded-For` HTTP Header, or the Remote Host if `X-Forwarded-For` is not set.

```go
// Consistent Hash the Client IP address
&types.Stickiness{
    IP: true,
},

// Fall back to Client IP if no rules match
&types.Stickiness{
    IP: true,
    Rules: []string{
        `{{index .Header "X-Notfound" | join ""}}`,
    },
},
```

**Cookie**

Sticky cookie is already the default, and remains so in a backwards-compatible manner.  A `CookieEncryptKey` option has been added, and when set will use AES Encryption on the cookie so that the User Agent receives an encrypted cookie.

```go
// Cookie is still the default
&types.Stickiness{}

// Cookies can now be encrypted
&types.Stickiness{
    CookieEncryptKey: "abc123",
},

// Fall back to Cookies if no rules match
stickiness: &types.Stickiness{
    Cookie: true,
    Rules: []string{
        `{{index .Header "X-Notfound" | join ""}}`,
    },
},
```

### Motivation

Once requests can be consistently routed to the same application server, a whole new world of possibilities opens up.  Application Logic that used to require synchronization through Redis or a Database can now be done in Memory.

The `Rules` method allows for customizing sticky routing in almost any way imaginable.

For simpler use cases, `IP` and `Cookie` are easy-to-use boolean configurations.

### More

- [x] Added/updated tests

### Additional Notes

- Fixes #2299 
- Fixes #1035
- Fixes #1207


gcxRun:
Hello,
Have you considered adding this to vulcand/oxy which is managing roundrobin+stickiness ?

caleblloyd:
@gcxRun I did look at adding this to https://github.com/vulcand/oxy

Ultimately there are a lot of decisions that are made here - which Sticky methods take precedence, which template functions that rules can use, consistent hashing versus storing backends in a map, etc.

`vulcand/oxy/roundrobin` is a general purpose roundrobin algorithm.  I think that all of the configuration and logic belongs in the `traefik` repository, as other consumers of the library may want to implement stickiness in different ways.

aantono:
Can we also add Docs as well, demonstrating usage, etc?

middlewares/sticky_backend_handler.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/containous/traefik/healthcheck"
+	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/types"
+	guuid "github.com/satori/go.uuid"
+	"github.com/serialx/hashring"
+	"github.com/vulcand/oxy/forward"
+)
+
+// StickinessParsed holds the parsed representation of a Stickiness object
+type StickinessParsed struct {
```
aantono: Can we rename this to something more "pleasant", like `StickinessConfig` or something along those lines?  `StickinessParsed` just sounds very strange/weird... 😜

middlewares/sticky_backend_handler.go
```diff
@@ -0,0 +1,274 @@
...
+		UseRules:         useRules,
+	}
+}
+
+// StickyBackendHandler inspects criteria such as Client IP address and uses
+// consistent hashing to route a request to a specific backend server
+type StickyBackendHandler struct {
+	lb   healthcheck.LoadBalancer
+	next http.Handler
+	sp   *StickinessParsed
```
aantono: Rename `sp` to `config` to improve readability (relates to previous comment about `StickinessParsed` being renamed to `StickinessConfig`

middlewares/sticky_backend_handler.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		if clientIP != "" {
+			key = "clientIP: " + clientIP
+		}
+	}
+
+	if key != "" {
+		servers := h.lb.Servers()
+		if len(servers) > 0 {
+			sortedServers := make([]string, len(servers))
```
aantono: Why not pre-sort it once during initialization and store as a struct field, instead of having to sort the same list for every invocation?  Granted that the list is unlikely to be very large, but even for a few dozen entries on a high load endpoints this will cause increase in latency that is non-negligible.

caleblloyd: I was under the impression that a Load Balancer's servers came from the [Servers() method in oxy/roundrobin/rr.go](https://github.com/vulcand/oxy/blob/856ab519c4cede21de1ecb8cf3507e3d056ce2ce/roundrobin/rr.go#L188-L197)

Since the Load Balancer could take servers in/our of rotation at any time and return them in any order, a fresh list must be retrieved and sorted on each request.

aantono: Good point.  That’s certainly a possibility.  Was just trying to see if we can optimize it somehow and not have to sort it every proxy request.

aantono: Just thought about it more and realized that even though Oxy “could” do that dynamically, in practice Traefik would not, as each change to the config from the provider triggers an internal config refresh, which actually recreates the Oxy router with all the Lb and other goodness.  @ldez, @emile, @timoreiman - can you guys confirm that?

caleblloyd: I think Traefik's `healtcheck.go` calls [UpsertServer](https://github.com/containous/traefik/blob/89d90de7d8f8369a7e30e704d3e88e3ea9b5447e/healthcheck/healthcheck.go#L121) and [RemoveServer](https://github.com/containous/traefik/blob/89d90de7d8f8369a7e30e704d3e88e3ea9b5447e/healthcheck/healthcheck.go#L132), so the result of `Servers()` is still changing throughout the lifetime of a single configuration and it's result could vary request-to-request

caleblloyd:
I have added documentation.  I think the next steps are after reviewing the design would be making sure that this could work with all of the different backends.

Would integration with different backend types (docker, kubernetes, etc.) happen in this PR or a follow-on PR?

middlewares/sticky_backend_handler.go
```diff
@@ -0,0 +1,274 @@
...
+				if key != "" {
+					break
+				}
+			}
+		}
+	}
+
+	if key == "" && h.config.UseIP {
+		var clientIP string
+		if xForwardedFor := r.Header.Get(forward.XForwardedFor); xForwardedFor != "" {
```
aantono: Any reason why you only use X-Forwarded-For and not X-Real-IP?  While the X-Forwarded-For is a more common variant, the X-Real-IP is also seeing substantial usage, depending on what LB/Proxy one uses. Also, there is a "wrinkle in the fabric" in a form of usage of PROXY protocol, which, when in use, provides the real information via a different element in the originating HTTP request instead of a header (see [Proxy Protocl](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt), which is commonly used as well, especially by AWS and GCP in their load balancers.  Traefik already supports this via use of `go-proxyproto` library.  (See `EntryPoint.ProxyProtocol` config section).  So this might require adding more use-case handling.  @ldez, @emilevauge, @timoreimann - any thoughts?

ldez:
@aantono this PR is in Design review state: https://github.com/containous/traefik/blob/master/MAINTAINER.md#labels.

middlewares/sticky_backend_handler.go
caleblloyd: Good point, `X-Real-IP` needs to be here too.  It sounds like the Proxy Protocol case is handled already - is that underlying connection object swapped out before hitting the Sticky middleware?

ldez:
@caleblloyd Thanks a lot for your contribution, and sorry for the delay.
We discussed internally of this PR and we have few comments:

- We prefer to create PR on one topic only. This PR deals with ip-hash, encrypted cookie and template rules. In this case, you should have created 3 PRs.
- In the first place, we think we should focus on adding ip-hash only. Implementing this in Traefik seems to be complex as you have to generate a new Cookie. Therefore, we suggest to implement this in https://github.com/vulcand/oxy.
- Finally, we strongly encourage you to discuss with the developer team on our #dev Slack channel _before_ implementing any new feature. This avoid wrong design decision early.

@caleblloyd We suggest to close this PR for now and discuss on Slack on how we could help you on this :)

caleblloyd:
> In the first place, we think we should focus on adding ip-hash only. Implementing this in Traefik seems to be complex as you have to generate a new Cookie. Therefore, we suggest to implement this in https://github.com/vulcand/oxy.

I disagree.  Oxy is supposed to be a reusable library.  Traefik is an implementation.

When implementing this PR, I found that https://github.com/vulcand/oxy/pull/97 introduced an opinionated way of setting cookies in Oxy that makes it difficult library callers to change.  I had to put parsing in to read and re-write `Cookie` and `Set-Cookie` headers.

Instead, a caller should be able to provide Oxy with a "preferred" backend.  Oxy should then send information (via a Context) back to the caller on the "actual" backend that the request was routed to.

Oxy should not be reading `Cookie: ` headers or writing `Set-Cookie` headers, because that it the job of the _implementation_, not the _library_.

This is the same reason Oxy should not be extended to contain consistent IP hashing logic built into it.  Traefik may want to use consistent IP hashing, but some other user of Oxy may want to hold a look-up table in memory for IP hashing.

> We prefer to create PR on one topic only. This PR deals with ip-hash, encrypted cookie and template rules. In this case, you should have created 3 PRs.

This PR covers the broad topic of how to implement sticky backend middleware.  I added cookie encryption and rules out of personal interest to show how flexible this middleware could be.  It seemed easier to implement an extensible middleware for sticky backend handling.

emilevauge:
@caleblloyd I really think that all this design discussion here should have taken place on Slack in the first place. We would have found an agreement on a solution earlier.

And yes, why not creating something more generic in oxy first, I'm not against this. We maintain this library as well now.

But we only want to add one feature by PR, and I suggest to ask us on Slack before implementing each feature, otherwise you may be disappointed if we close your PR because we disagree on the design.

I really hope to see your there to continue the discussion :)

ahmedabt:
I believe this feature is highly needed, what is the final decision?

chanpl:
@emilevauge 
Have the team come to any decision on this change? I believe most of users would be happy to see Traefik supports multiple sticky session handlers.

PR #2617, opened 2017-12-23
Mesos: Integrate the Events API to get up to date information about tasks.

nemosupremo:
### What does this PR do?

Currently, the Mesos backend doesn't respond to events, and the way Traefik gets it's information about Mesos tasks is by polling. Every 30 seconds (the default) Traefik will ping Mesos about its current state and refresh it's configuration. This obviously has some drawbacks - when deploying a new application, the entire process could take less than 30 seconds, but due to the way polling works, your clients will see your application as down until Traefik updates its state.

In Mesos 1.1, the [events](http://mesos.apache.org/documentation/latest/operator-http-api/#events) API was introduced so that operators could listen to events on the Mesos state. This allows services, like Traefik, to have an up to date view on the Mesos cluster state.

This PR introduces two new options, `Subscribe` (boolean) and `SubscribeLabels`, and changes the functionality of `RefreshSeconds`. When `Subscribe` is set to true, Traefik will use the Events API to listen on Mesos state. Traefik will refresh the state when a task goes in to the `RUNNING` state or when a task is killed that Traefik is currently routing traffic to. Optionally `SubscribeLabels` can be set so that only tasks with a certain label cause a state refresh. With these new options `RefreshSeconds` can be set to 0 to disable polling. Setting `watch` to false will disable both options.

These new options are not enabled by default in order to maintain backwards compatibility.

### Motivation

With the addition of the events API (and the health check API added some time ago), this brings the Mesos backend almost up to par with other backends. I prefer using the Mesos backend as the API is more stable than Marathon's and the Mesos backend will support other frameworks (like Aurora). 

### More

- [ ] Added/updated tests

### Additional Notes

Implementation Details:

I built off the current existing code in order to maintain backwards compatibility with Mesos <1.1. However, ideally this can built by exclusively using the subscribe command and maintaining state there, which would probably be a bit more performant.

Tests:

I haven't added any tests because the new functionality depends quite a bit a mesos (since it only plugs in the Event API into the current existing architecture). It would require integration tests with e a Mesos instance which I think may be cumbersome to setup.

### DIFF
--- a/docs/configuration/backends/mesos.md
+++ b/docs/configuration/backends/mesos.md
@@ -62,12 +62,26 @@ domain = "mesos.localhost"
 #
 # ZkDetectionTimeout = 30
 
-# Polling interval (in seconds).
+# Polling interval (in seconds). If the polling interval is set to 0, polling is disabled.
 #
 # Optional
 # Default: 30
 #
 # RefreshSeconds = 30
+
+# Subscribe to events from the mesos master when tasks are added and destroyed. Requires Mesos >1.1.
+#
+# Optional
+# Default: false
+#
+# Subscribe = true
+
+# Only update the configuration in response to changes from tasks with the following labels (comma separated)
+#
+# Optional
+# Default: ""
+#
+# SubscribeLabels = "traefik.frontend.rule"
 
 # IP sources (e.g. host, docker, mesos, rkt).
 #
--- a/provider/mesos/config.go
+++ b/provider/mesos/config.go
@@ -17,7 +17,7 @@ import (
 	"github.com/mesosphere/mesos-dns/records/state"
 )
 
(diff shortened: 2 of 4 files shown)


ldez:
@nemosupremo Could you rebase?

nemosupremo:
I'll try to update this PR by the end of the week

emilevauge:
@nemosupremo any news on this ?

nemosupremo:
Sorry for the slowness on this, I will get to this by monday. The changes to mainline were more than I expected so I think I may re-implement this, and I wasn't able to merge my branches changes as I needed functinoality like #2502, as a result I had to reprioritize.

emilevauge:
@nemosupremo I'm sorry but due to inactivity, I will close this PR. Feel free to reopen a new one on this topic :) Thanks !

PR #2639, opened 2018-01-02
Add status code text to webui bar chart tooltip

wader:
### What does this PR do?

So you don't have to lookup status code numbers

### Motivation

I keept googling them

### More

- [ ] Added/updated tests

Not sure how to add test for this?

### DIFF
--- a/webui/package.json
+++ b/webui/package.json
@@ -21,6 +21,7 @@
     "angular-ui-router": "^0.3.1",
     "animate.css": "^3.4.0",
     "bootstrap": "^3.3.6",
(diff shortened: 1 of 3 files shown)


wader:
This changes the tooltip design a bit, now it's just "200 OK" before it was "[color] 200 [count]" with more padding

ldez:
@wader I think you can use https://www.npmjs.com/package/http-status-codes to simplify the code.

wader:
@ldez Ok, have switched to using http-status-codes now

wader:
Let me know if there something else to fix

Issue #2565, opened 2017-12-13
Redirect generate wrong access logs


jdel:
### Do you want to request a *feature* or report a *bug*?

Reporting a bug

### What did you do?

Run traefik with http redirected to https and examine access logs.

### What did you expect to see?

A single curl on nginx1 generates 1 line in the access log:

```
172.17.0.1 - - [13/Dec/2017:19:49:27 +0000] "GET / HTTP/1.1" 200 612 - "curl/7.54.0" 34 "Host-nginx1-test-tld-3" "http://172.17.0.3:80" 0ms
```

A single curl on nginx2 generates 1 line in the access log:

```
172.17.0.1 - - [13/Dec/2017:19:50:41 +0000] "GET / HTTP/1.1" 200 612 - "curl/7.54.0" 42 "Host-nginx2-test-tld-2" "http://172.17.0.4:80" 0ms
```

### What did you see instead?

A single curl on nginx1 generates 2 lines in the access log:

```
172.17.0.1 - - [13/Dec/2017:19:49:27 +0000] "GET / HTTP/1.1" - - - "curl/7.54.0" 33 "Host-nginx1-test-tld-3" - 0ms
172.17.0.1 - - [13/Dec/2017:19:49:27 +0000] "GET / HTTP/1.1" 200 612 - "curl/7.54.0" 34 "Host-nginx1-test-tld-3" "http://172.17.0.3:80" 0ms
```

Now a curl on nginx2 shows a erroneous output in the first line:

```
172.17.0.1 - - [13/Dec/2017:19:50:41 +0000] "GET / HTTP/1.1" - - - "curl/7.54.0" 41 "Host-nginx1-test-tld-3" - 0ms
172.17.0.1 - - [13/Dec/2017:19:50:41 +0000] "GET / HTTP/1.1" 200 612 - "curl/7.54.0" 42 "Host-nginx2-test-tld-2" "http://172.17.0.4:80" 0ms
```

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
docker run --rm traefik:v1.4 version
Version:      v1.4.5
Codename:     roquefort
Go version:   go1.9.2
Built:        2017-12-06_10:16:48AM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
defaultEntryPoints = ["http", "https"]
logLevel = "ERROR"
InsecureSkipVerify = true

[accessLog]
[...]

PR #2682, opened 2018-01-09
Fix bad access log

### LINKED ISSUE #2649: Missing quotes arround the Referer field in access logs
Both at the same time...

### What did you do?
We are sending Træfik access logs to a Logstash instance through a Docker GELF log driver.
All is working smoothly, but...

The access log format of Træfik do not quote the Referrer field when it is empty
https://github.com/containous/traefik/blob/8339139400e5dfd1fed6cb3545d611d8311ab381/middlewares/accesslog/logger_formatters.go#L66

However in the common Apache2 Combined format, this field needs to be quoted
(see https://github.com/elastic/logstash/blob/v1.4.2/patterns/grok-patterns#L91)

As a result, we are not able to parse the access log correctly.

### What did you see instead?
A typical access log line without a Referrer is like:
`x.x.112.4 - - [03/Jan/2018:09:45:13 +0000] "GET / HTTP/1.1" 200 24829 - "curl/7.55.1" 5 "HostRegexp-domain-8" "http://10.0.0.5:80" 3ms`
Note the `-` without surrounding quotes before the User-Agent.

```
Version:      v1.4.5
Codename:     roquefort
Go version:   go1.9.2
Built:        2017-12-06_10:16:48AM
OS/Arch:      linux/amd64
```

Thanks for this wonderful reverse-proxy!
Antoine

mmatur:
### What does this PR do?

This PR fix two issues:

- bad access log when Træfik throw an error
- missing quotes on referrer and user-agent fields

### Motivation

Fixes #2649 #2565 

### More

- [x] Added/updated tests


integration/fixtures/consul/simple.toml469785378
```diff
@@ -0,0 +1,16 @@
+defaultEntryPoints = ["http"]
```
Juliens: Are you sure about this file ?

integration/fixtures/traefik_log_config.toml
```diff
@@ -11,3 +15,11 @@ checkNewVersion = false
 [entryPoints]
   [entryPoints.http]
   address = ":8000"
+
+[web]
```
Juliens: [web] is deprecated

integration/fixtures/access_log_config.toml
```diff
@@ -1,48 +1,45 @@
...
+    address = ":8006"
+  [entryPoints.httpRateLimit]
+    address = ":8007"
 
 checkNewVersion = false
 
 ################################################################
 # Web configuration backend
 ################################################################
 [web]
```
Juliens: [web] is deprecated

PR #1233, opened 2017-03-03
Feature web root path

tcoupin:
## Allow to set the root path for webui and API

Adding 'path' parameter for web provider.
Adding redirect from '/' to '/path'

Path can be :
- ''
- '/'
- '/path'
- '/path/'

PR #2686, opened 2018-01-10
Fix redirect problem on dashboard + docs/tests on [web]

Juliens:
### What does this PR do?
Fix a problem with the dashboard redirect
Fix error in documentation + integration tests

### Motivation
Answer how to handle the deprecation of --web.path
Fix the doc with deprecated
related #1233

### DIFF
--- a/docs/configuration/backends/web.md
+++ b/docs/configuration/backends/web.md
@@ -35,6 +35,15 @@ address = ":8080"
 # Default: false
 #
 readOnly = true
+
+ 
+# Set the root path for webui and API
+#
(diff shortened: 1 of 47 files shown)


docs/configuration/backends/web.md
```diff
@@ -35,6 +35,14 @@ address = ":8080"
 # Default: false
 #
 readOnly = true
+
+ 
+# Set the root path for webui and API
+#
+# Optional
```
ldez: could add a "Deprecated" here.

PR #2728, opened 2018-01-19
Fix config duplication detection and throttle goroutine lifecycle.

timoreimann:
This change fixes two problems:

- Configurations were [compared against the persisted configuration](https://github.com/containous/traefik/blob/5140bbe99a79b45f98c27fbb8e9b6833194af4cb/server/server.go#L373), not the latest received; the two may deviate when throttling is active. We introduce a local configuration cache and compare against that.
- The variable managing throttling goroutines was only [scoped to the local method](https://github.com/containous/traefik/blob/5140bbe99a79b45f98c27fbb8e9b6833194af4cb/server/server.go#L365), effectively disabling goroutine cleanup and producing a leak. We properly hook up the goroutines to live as long as the server does.

We also refactor the throttling logic into a dedicated struct. While not completely separating the concerns of the server from the throttler (we still need to reference a number of server methods/values inside the throttler), it's a step forward.

Additional changes:

- Pass the stop signal from listenProviders through to the throttler eventually, avoiding the need to close the server for a proper shutdown. 
- Exit throttler if ring buffer is closed. (Could happen before stop signal is processed.)
- Test: Return published configs through channel to rule out race conditions.
- Test: Invoke stop channel function earlier to trigger goroutine shutdown.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -38,7 +38,6 @@ import (
 	traefikTls "github.com/containous/traefik/tls"
 	"github.com/containous/traefik/types"
 	"github.com/containous/traefik/whitelist"
-	"github.com/eapache/channels"
 	thoas_stats "github.com/thoas/stats"
 	"github.com/urfave/negroni"
 	"github.com/vulcand/oxy/connlimit"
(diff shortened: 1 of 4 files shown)


timoreimann:
Closing this one in favor of smaller, more digestable PRs. The first one is already out, the others will follow (they are not that critical).

PR #2733, opened 2018-01-22
Tracing statusCodeTracker need to implement CloseNotify

mmatur:
### What does this PR do?

statucCodeTracker does not implement CloseNotify

### Motivation
 Avoid noise in logs:
```console
time="2018-01-19T16:11:56Z" level=warning msg="Upstream ResponseWriter of type *tracing.statusCodeTracker does not implement http.CloseNotifier. Returning dummy channel."
```

### DIFF
--- a/middlewares/tracing/status_code.go
+++ b/middlewares/tracing/status_code.go
@@ -8,7 +8,7 @@ import (
 
 type statusCodeRecoder interface {
 	http.ResponseWriter
-	GetStatus() int
+	Status() int
 }
 
 type statusCodeWithoutCloseNotify struct {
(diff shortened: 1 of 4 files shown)


middlewares/tracing/status_code.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+// Hijack hijacks the connection
+func (s *statusCodeWithoutCloseNotify) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	return s.ResponseWriter.(http.Hijacker).Hijack()
+}
+
+// Flush sends any buffered data to the client.
+func (s *statusCodeWithoutCloseNotify) Flush() {
+	s.ResponseWriter.(http.Flusher).Flush()
```
marco-jantke: Can you do a type assertion to check whether the underlying response writer is a `Flusher`? `Flush` might actually be called in circumstances where the response writer, or better say the underlying connection, does not have this ability. Something like:

```
if flusher, ok := s.ResponseWriter.(http.Flusher); ok {
    flusher.Flush()
}
```

mmatur:
@marco-jantke I just fix your reviews :)

PR #2759, opened 2018-01-24
Fix integration tests in tracing

mmatur:
### What does this PR do?

This PR fix flaky integration tests in tracing.

Flaky was due to bug in the latest [Zipkin version 2.4.4](https://github.com/openzipkin/zipkin/releases/tag/2.4.4)

### More

- [x] Added/updated tests

### Additional Notes
screenshot from 2018-01-24 17-03-31

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -56,6 +56,7 @@ func init() {
 		check.Suite(&RateLimitSuite{})
 		check.Suite(&SimpleSuite{})
 		check.Suite(&TimeoutSuite{})
+		check.Suite(&TracingSuite{})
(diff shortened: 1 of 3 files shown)


marco-jantke:
I don't understand the PR. From the description I read it as this should fix a flaky integration test but it actually adds a new one. What am I missing?

mmatur:
@marco-jantke In the merge back fron 1.5 into master #2754. We have decided to remove `tracing_test.go` because integration tests was failing every time due to an issue in the Zipkin docker image. And we wanted to have times to understand and fix properly these tests. This this why we have re add `tracing_test.go`

PR #2795, opened 2018-02-02
Fix dnsrr endpoint mode excluded when not using swarm LB

### LINKED ISSUE #2793: Fix for DNS RR in Swarm Mode causes loss of functionality

Report an issue/bug.

### What did you do?

We use Traefik on Docker Swarm to expose hundreds of distinct services. Since Docker Swarm's overlay networks get slow with Virtual IPs on this scale we have set all the relevant services to `endpoint_mode: dns_rr`. See moby/moby#30820.

We are aware that Traefik does not (yet) support load balancing with DNS RR on Swarm. This is OK for us since our target services currently all have `scale: 1`. Up until Traefik 1.4.x this worked just fine.

Traefik 1.5 continues to discover and expose services with `endpoint_mode: dns_rr` even if it does not support load balancing or displaying the appropriate IPs in the Web UI (yet).

### What did you see instead?

Traefik 1.5 ignores all services with `endpoint_mode: dns_rr`. This is probably caused by #2490.

While this fix prevents people from accidentally relying non-functional DNS RR load-balancing it also blocks a previously working scenario.

```
Version:      v1.5.0
Codename:     cancoillotte
Go version:   go1.9.2
Built:        2018-01-23_04:42:32PM
OS/Arch:      linux/amd64
```

```toml
defaultEntrypoints = ["http"]
constraints = ["tag==public"]
logLevel = "WARN"

[accessLog]
format = "json"

[docker]
watch = true
exposedbydefault = false
swarmmode = true

[web]
readOnly = true

[entryPoints.http]
address = ":80"
```

```
Ignored endpoint-mode not supported, service name: myservice
```

mmatur:
### What does this PR do?

Fix an issue where Træfik stop exposing containers with `endpoint_mode: dnsrr` when label `traefik.backend.loadbalancer.swarm=false`

### Motivation

Fixes #2793

### More

- [x] Added/updated tests

### DIFF
--- a/provider/docker/docker.go
+++ b/provider/docker/docker.go
@@ -777,19 +777,21 @@ func listServices(ctx context.Context, dockerClient client.APIClient) ([]dockerD
 
 	for _, service := range serviceList {
 		dockerData := parseService(service, networkMap)
-		if len(dockerData.NetworkSettings.Networks) > 0 {
-			useSwarmLB, _ := strconv.ParseBool(getIsBackendLBSwarm(dockerData))
 
-			if useSwarmLB {
+		useSwarmLB, _ := strconv.ParseBool(getIsBackendLBSwarm(dockerData))
+
+		if useSwarmLB {
+			if len(dockerData.NetworkSettings.Networks) > 0 {
 				dockerDataList = append(dockerDataList, dockerData)
-			} else {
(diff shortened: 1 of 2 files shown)


mikeifomin:
Looking forward 1.5.2

PR #2800, opened 2018-02-03
Hide the pflag error when displaying help.

ldez:
### What does this PR do?

Hide the pflag error when displaying help.

### Motivation

Better help display.

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -28,6 +28,7 @@ import (
 	"github.com/containous/traefik/types"
 	"github.com/containous/traefik/version"
 	"github.com/coreos/go-systemd/daemon"
+	"github.com/ogier/pflag"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/traefik/traefik.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/containous/traefik/types"
 	"github.com/containous/traefik/version"
 	"github.com/coreos/go-systemd/daemon"
+	"github.com/ogier/pflag"
```
timoreimann: FWIW, most people use spf13/pflag instead of ogier's.

ldez: come from flaeg:

https://github.com/containous/flaeg/blob/c4797734e8b31bd78dc2cc5d88d510decdee2d7e/Gopkg.toml#L25

timoreimann: All good then, thanks. :+1:

PR #2801, opened 2018-02-03
Display file log when test fails.

ldez:
### What does this PR do?

Display file log when test fails.

Clean some test methods.

### Motivation

Have more information when tests fails.

### More

- [x] Added/updated tests

### DIFF
--- a/integration/access_log_test.go
+++ b/integration/access_log_test.go
@@ -50,19 +50,20 @@ func (s *AccessLogSuite) TestAccessLog(c *check.C) {
 	// Start Traefik
 	cmd, display := s.traefikCmd(withConfigFile("fixtures/access_log_config.toml"))
 	defer display(c)
+	defer displayTraefikLogFile(c, traefikTestLogFile)
+
 	err := cmd.Start()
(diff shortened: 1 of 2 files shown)


ldez:
@vdemeester I already tested that:
- we already use `TearDown` or `TearDownSuite` for some tests: it's not possible to have 2 methods
- this need to create a state, and that create some other problems.

integration/integration_test.go
```diff
@@ -19,8 +19,8 @@ import (
 	checker "github.com/vdemeester/shakers"
 )
 
-var integration = flag.Bool("integration", false, "run integration tests")
-var container = flag.Bool("container", false, "run container integration tests")
+var integration = flag.Bool("integration", true, "run integration tests")
```
mmatur: Could you please change to `false`

integration/integration_test.go
```diff
@@ -19,8 +19,8 @@ import (
...
+var container = flag.Bool("container", true, "run container integration tests")
```
mmatur: Could you please change to `false`

ldez: oops

PR #2817, opened 2018-02-06
Remove health check filter from Marathon tasks.

timoreimann:
### What does this PR do?

Remove health check filter from Marathon tasks.

### Motivation

When available, Traefik uses the last Marathon health check result to decide whether a task should be filtered or not. This behavior is incorrect, however, as it ignores the failure threshold denoting the
number of failed health checks it would take for Marathon to actually consider a task as failing and trigger a restart. With Traefik not taking the threshold into account, a task is removed from the
configuration already while Marathon may still deem it healthy. 

Delaying the filtering until the moment when the failure threshold is exceeded would not buy us much: At that point, Marathon will decide to restart the unhealthy task anyway and cause a task state change that will lead to a Traefik configuration update excluding the task in question.

Consequently, we can remove the filter logic.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/user-guide/marathon.md
+++ b/docs/user-guide/marathon.md
@@ -130,7 +130,6 @@ As such, there is no way to handle this situation deterministically.
 Finally, Marathon health checks are not mandatory (the default is to use the task state as reported by Mesos), so requiring them for Traefik would raise the entry barrier for Marathon users.
 
 Traefik used to use the health check results as a strict requirement but moved away from it as [users reported the dramatic consequences](https://github.com/containous/traefik/issues/653).
-If health check results are known to exist, however, they will be used to signal task availability.
 
 #### Draining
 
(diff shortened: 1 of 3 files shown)


aantono:
I think this makes good sense.  In general the delegated provider's source (i.e. Marathon, Consul, K8s, etc) should be responsible for managing health state of the entries and Traefik only responds to what they produce as "valid endpoints".  If one wants to explicitly have Traefik do health checking, that can be done via backend healthcheck config.  So in short - LGTM 👍

timoreimann:
@aantono thanks for your feedback. 👏

Juliens:
Do you think we can add an option to enable or disable this (with default to `enable` for breaking change) ?

timoreimann:
@Juliens taking our latest discussion on the subject on Slack into account, do you think it's still necessary to have a switch?

My argument is that the existing behavior is inherently broken and bears a risk to availability for Marathon users, so any setting matching this behavior by default won't do any better. Users should either adjust their Marathon health checks or employ Traefik health checks.

Let me know your thoughts.

Juliens:
After discussion, oky with not adding the switch.

PR #2839, opened 2018-02-12
Change boolean annotation values to string

hobti01:
### What does this PR do?

Update annotations documentation

### Motivation

Kubernetes annotation values must be strings. 
Booleans and numbers are not valid types.
Unfortunately no clear upstream documentation exists stating this.

### More

### Additional Notes

Kubernetes annotation values must be strings.  Booleans and numbers are not valid types.  Unfortunately no clear upstream documentation exists stating this.

[Annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
[Annotations have same requirements as labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)
[GetAnnotations accessor returns string map](https://github.com/kubernetes/kubernetes/blob/f5f6f3e715cb8dfbd9657a4229c77ec6a5eab135/staging/src/k8s.io/apimachinery/pkg/api/meta/meta.go#L583)

Formatting to add space between colon and value.


ldez:
My main problem is that the documentation on this part has been completely rewritten in master.
It's not possible to just rebase.
This must be done in 2 PRs.

PR #2849, opened 2018-02-12
Handle zipkin collector creation error

ferhatelmas:
### What does this PR do?

Handle skipped error.

### Motivation

According to current implementation, there is no need to check for error.
However, it's future proof to handle error since signature signals that err is possible.

### More

- [ ] Added/updated tests

### Additional Notes

None.


ldez:
Thanks for the contribution 👍 

We must declined this change, because this can be problem in the future.

ferhatelmas:
Sure. For my learning, I would appreciate if you can explain how it could be a problem. Thanks!

ldez:
In this case we must see `zipkin.NewHTTPCollector(c.HTTPEndpoint)` as a black box:
- this method define a contract by this return
- the behavior of the method can change in the future release of zipkin an return a real error instead of nil

ldez:
Sorry, I have misreading... could you reopen ?

PR #2853, opened 2018-02-13
Remote Address Masking

zweizeichen:
### What does this PR do?

This PR adds the ability to mask or completely redact client's IP addresses and hostnames in the access log. Partial masking is enabled by default resulting in a private by default behaviour for traefik.

### Motivation

With the [GDPR](https://en.wikipedia.org/wiki/General_Data_Protection_Regulation) becoming enforceable soon and court rulings in countries such as Germany, logging and storage of (potentially) personally identifiable information needs to be reduced in many organizations.
This allows traefik's users to configure the access logger to mask/redact their client's IP addresses and hostnames.
There are three options for `remoteIPMask`:

1. `off` Log client's full adresses and hostnames like before
2. `partial` (default) Redact hostnames, [mask the last octet of IPv4 addresses and the last 80 bytes of IPv6 addreses](https://support.google.com/analytics/answer/2763052?hl=en)
3. `full` Redact hostnames and mask the whole IP address

Changing the default behaviour to `partial` may be a breaking change for some users now, however it is expected that many (future) users will operate traefik under EU jurisdiction and neither have the need nor the legal authorization to store this data. Thus it could be beneficial in the long term to adjust traefik's default behaviour in this case.

As I am not a lawyer I can not assess the legal compliance of this PR's changes, however I think it is useful to have these tools available vs. having to disable logging entirely.

### More

- [x] Added/updated tests

### Additional Notes

- Logging of usernames and passing addresses in HTTP headers should also be evaluated

### DIFF
--- a/middlewares/accesslog/logger.go
+++ b/middlewares/accesslog/logger.go
@@ -163,7 +163,7 @@ func (l *LogHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next h
 	// Mask host's address if necessary
 	isIPv6 := strings.Contains(host, ":")
 
-	if l.maskIPv4 != nil && l.maskIPv6 != nil {
+	if len(l.maskIPv4) > 0 && len(l.maskIPv6) > 0 {
 		host = maskHost(host, l.maskIPv4, l.maskIPv6, isIPv6)
 	}
 
--- a/middlewares/accesslog/logger.go
+++ b/middlewares/accesslog/logger.go
@@ -45,8 +45,8 @@ type LogHandler struct {
 	logger   *logrus.Logger
 	file     *os.File
 	filePath string
-	maskIPv4 *net.IPMask
-	maskIPv6 *net.IPMask
+	maskIPv4 net.IPMask
+	maskIPv6 net.IPMask
 	mu       sync.Mutex
 }
 
@@ -73,8 +73,8 @@ func NewLogHandler(config *types.AccessLog) (*LogHandler, error) {
 	}
 
 	// Enable masks for remote addresses
-	var maskIPv4 *net.IPMask
-	var maskIPv6 *net.IPMask
+	var maskIPv4 net.IPMask
+	var maskIPv6 net.IPMask
 
 	// Set default value if not set
 	if len(config.RemoteIPMask) == 0 {
@@ -83,20 +83,12 @@ func NewLogHandler(config *types.AccessLog) (*LogHandler, error) {
 
 	switch config.RemoteIPMask {
 	case NoMask:
-		maskIPv4 = nil
-		maskIPv6 = nil
 	case PartialMask:
-		v4 := net.CIDRMask(24, 32)
-		v6 := net.CIDRMask(48, 128)
-
-		maskIPv4 = &v4
-		maskIPv6 = &v6
+		maskIPv4 = net.CIDRMask(24, 32)
+		maskIPv6 = net.CIDRMask(48, 128)
 	case FullMask:
-		v4 := net.CIDRMask(0, 32)
-		v6 := net.CIDRMask(0, 128)
-
-		maskIPv4 = &v4
-		maskIPv6 = &v6
+		maskIPv4 = net.CIDRMask(0, 32)
+		maskIPv6 = net.CIDRMask(0, 128)
 	default:
 		return nil, fmt.Errorf("unsupported IP address mask setting for access log: %s", config.RemoteIPMask)
 	}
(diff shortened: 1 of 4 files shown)


zweizeichen:
Just saw this might beautifully complement #2397. I should have checked this beforehand :) I hope there is no overlap.

zweizeichen:
Thanks for the review, please note that the existing documentation does not always have newlines before code blocks. I only adjusted the lines you marked, but there are other places where there is no newline preceeding a code snippet, therefore resulting in an inconsistent style.

middlewares/accesslog/logger.go
```diff
@@ -28,13 +29,24 @@ const (
...
+	FullMask = "full"
 )
 
 // LogHandler will write each request and its response to the access log.
 type LogHandler struct {
 	logger   *logrus.Logger
 	file     *os.File
 	filePath string
+	maskIPv4 *net.IPMask
+	maskIPv6 *net.IPMask
```
ldez: I think you can do (remove pointer):
```go
type LogHandler struct {
	logger   *logrus.Logger
	file     *os.File
	filePath string
	maskIPv4 net.IPMask
	maskIPv6 net.IPMask
	mu       sync.Mutex
}
```

middlewares/accesslog/logger.go
```diff
@@ -60,13 +72,42 @@ func NewLogHandler(config *types.AccessLog) (*LogHandler, error) {
...
+	if len(config.RemoteIPMask) == 0 {
+		config.RemoteIPMask = PartialMask
+	}
+
+	switch config.RemoteIPMask {
+	case NoMask:
+		maskIPv4 = nil
+		maskIPv6 = nil
+	case PartialMask:
+		v4 := net.CIDRMask(24, 32)
```
ldez: with my previous comment, you can rewrite like that:
```go

	// Set default value if not set
	if len(config.RemoteIPMask) == 0 {
		config.RemoteIPMask = PartialMask
	}

	// Enable masks for remote addresses
	var maskIPv4 net.IPMask
	var maskIPv6 net.IPMask

	switch config.RemoteIPMask {
	case NoMask:
	case PartialMask:
		maskIPv4 = net.CIDRMask(24, 32)
		maskIPv6 = net.CIDRMask(48, 128)
	case FullMask:

		maskIPv4 = net.CIDRMask(0, 32)
		maskIPv6 = net.CIDRMask(0, 128)
	default:
		return nil, fmt.Errorf("unsupported IP address mask setting for access log: %s", config.RemoteIPMask)
	}
```

middlewares/accesslog/logger.go
```diff
@@ -125,8 +166,25 @@ func (l *LogHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next h
...
 	core[RequestLine] = fmt.Sprintf("%s %s %s", req.Method, urlCopyString, req.Proto)
 
-	core[ClientAddr] = req.RemoteAddr
-	core[ClientHost], core[ClientPort] = silentSplitHostPort(req.RemoteAddr)
+	host, port := silentSplitHostPort(req.RemoteAddr)
+
+	// Mask host's address if necessary
+	isIPv6 := strings.Contains(host, ":")
+
+	if l.maskIPv4 != nil && l.maskIPv6 != nil {
```
ldez:
```go
if len(l.maskIPv4) > 0 && len(l.maskIPv6) > 0 {
```

zweizeichen:
Done

middlewares/accesslog/logger.go
zweizeichen: len(nil) == 0 
https://golang.org/pkg/builtin/#len

ldez: but `nil != []`

zweizeichen: Yes but in this case it won't make any difference? 
https://play.golang.org/p/UZUFhDfM5sF

ldez: on slice we prefer use `len`

zweizeichen: There you go

timoreimann:
Two remarks:

1. Choosing a default of `partial` seems like a breaking change. Shouldn't we pick `off` instead?
1. With both #2397 and this PR covering the functionality of fully redacting fields, what's our plan to converge to a single implementation in this regard?

zweizeichen:
1. Please check the motivation I wrote for the reason
2. I feel that the access log module is not structured that well, especially if you look at the hoops you have to jump through for testing as each test is E2E. This results in pretty convoluted testing code which is hard to get into and maintain. Maybe a large refactoring of that module would be better in the long run.

zweizeichen:
Yup thought about aggregators, too. I was just thinking that deploying traefik and enabling access logs should not carry legal risks by default.

timoreimann:
@zweizeichen I totally see your point. Given how complex the legal field is, however, I'd prefer to not try to "optimize" in this regard. In a sense, it seems safer and more robust to me to offer the technical abilities necessary and let users decide what they want to do.

timoreimann:
But let's see what the rest has to say before making any changes.

emilevauge:
@zweizeichen Thanks for this PR !
As @timoreimann said earlier, this PR and #2397 are covering somewhat the same functionality, we  think it may be a good idea to implement something a bit more generic.
We had multiple discussions around this and we ended with a proposal: #2875
WDYT?
_Side note: until we find a consensus, we think it would better to avoid any development on this topic :)_

PR #2858, opened 2018-02-14
Add warning that PUT to API only works with rest and web

gytisgreitai:
### What does this PR do?

Adds a warning to API documentation about updating providers

### Motivation

The API documentation seems to state that you can invoke PUT for updating providers for any provider, though in fact it only works for `rest` or `web`:

https://github.com/containous/traefik/blob/395b1702dee610107e3690afef0490bc0beca282/provider/rest/rest.go#L35

### More

### DIFF
--- a/docs/configuration/api.md
+++ b/docs/configuration/api.md
@@ -56,6 +56,9 @@
     For compatibility reason, when you activate the rest provider, you can use `web` or `rest` as `provider` value.
     But be careful, in the configuration for all providers the key is still `web`.
 
+!!! warning
+    Updating providers with `PUT /api/providers/{provider}` can only be used with `web` or `rest`
+
 ### Provider configurations
 
 ```shell


ldez:
The warning already exist in http://docs.traefik.io/configuration/backends/rest/#api

gytisgreitai:
Yes it does, but if you want to readup about REST Api you head to this section, and from the description you can get the impression that it will work for any provider. So:

- I've read documentation about general Config
- I've read documentation about acme
- I've read documentation about Zookeeper
- I've configured traefik with zookeeper, then spent couple of hours writing a JSON to submit via API
- Next hour was wasted trying to understand why doesn't it work:

```
$ curl -XGET http://traefik:8080/api/providers                                                                                                                                                                                                                     
{"zk":{}}%

$ curl -XPUT http://traefik:8080/api/providers/zk -d '<....>'                                                                                                                                                                                                                      
404 page not found
```

so IMHO it would be clearer to state this upfront.

ldez:
Thanks for the contribution 👍 

Closed in favor of #2863

PR #2859, opened 2018-02-14
Fix: correct conditional setting of buffering retry expression

harnash:
### What does this PR do?

This PR fixes setting buffering middleware in case of empty retry expression. `CondSetter` sets options when condition is *FALSE* (counterintuitively).

### Motivation

After testing I've noticed that buffering is not enabled even with proper settings in place.

### More

- [x] Added/updated tests


ldez:
@harnash It's my fault, this need to fixed in oxy, I will fix that now.

PR #2867, opened 2018-02-16
Refactor kubernetes client: introduce informer factory

yue9944882:
### What does this PR do?

Use `k8s.io/cilent-go/informers` to manage informers and its indexers / store.

### Motivation
Currently we are managing the stores of informers explicitly which may be unnecessary after bumping up `client-go`. Informer factory will help sort out and clearify implementation of kubernetes client.

This PR will simplify kubernetes client a lot.

### More

- [ ] Added/updated tests


ldez:
@yue9944882 like the next time, could you organize your commit to facilitate the review.
1. one commit for `Gopkg.*`
2. one commit for Træfik code.
3. one commit to the vendor.

Thanks.

yue9944882:
@rancher @marathon Sorry for pinging by mistake. ( IDK if `make-ensure` is not working correctly on my macbook but I will figure this out later )

> @yue9944882 like the last time, could you organize your commit to facilitate the review (please respect the order).


@ldez It seems that we encountered a BUG of Github displayment. This PR's commits order is inconsistent with the order in my forked repo. How about continuing on since that I didn't modify Gopkg.* directly?

screen shot 2018-02-16 at 3 53 47 am

ldez:
Github order by commit date: as you have alter your commit (`rebase -i`), the commits dates have not changed.

provider/kubernetes/client.go
```diff
@@ -186,109 +162,51 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
 	}
-
 	return result
 }
 
 // GetService returns the named service from the given namespace.
 func (c *clientImpl) GetService(namespace, name string) (*corev1.Service, bool, error) {
 	var service *corev1.Service
-	item, exists, err := c.svcStores[c.lookupNamespace(namespace)].GetByKey(namespace + "/" + name)
+	item, exists, err := c.factories[namespace].Core().V1().Services().Informer().GetStore().GetByKey(namespace + "/" + name)
```
yue9944882: Ideally this line could be `svc, err := c.factories[namespace].Core().V1().Services().Lister().Services(namespace).Get(name)` which use `Lister` to `Get` resources from indexer inside the informer. But this will need to change the return types of `Client` interface to `GetService(namespace, name string) (*corev1.Service, error)` and lots of changes in test facilities will change too.

yue9944882: Note that they are actually the same. But eliminating `namespace + "/" + name` will help make the code clearer.

yue9944882:
@ldez @timoreimann PTAL thx 😉

ldez:
@yue9944882 please stop ask `PTAL`, we add already notifying of commits, comments, ...

yue9944882:
@timoreimann Good catch, it would be no doubt a horrible BUG 😨

provider/kubernetes/client.go
```diff
@@ -150,31 +124,28 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
+			opts.LabelSelector = labelSelector
+		})
+		factory.Extensions().V1beta1().Ingresses().Informer().AddEventHandler(eventHandler)
+		factory.Core().V1().Services().Informer().AddEventHandler(eventHandler)
+		factory.Core().V1().Endpoints().Informer().AddEventHandler(eventHandler)
+		factory.Core().V1().Secrets().Informer().AddEventHandler(eventHandler)
+
+		for t, ok := range factory.WaitForCacheSync(stopCh) {
+			if !ok {
+				return nil, fmt.Errorf("timed out waiting for controller caches to sync %s in namespace %s", t.String(), ns)
```
timoreimann: What's an example output of `t.String()`?

timoreimann: Let's use `%q` for printing `ns` as it may be the empty string when no namespaces are specified.

provider/kubernetes/client.go
```diff
@@ -150,31 +124,28 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
-	for _, ns := range namespaces {
-		ns := ns
-		informManager.extend(c.WatchIngresses(ns, kubeLabelSelector, eventCh), true)
-		informManager.extend(c.WatchObjects(ns, kindServices, &corev1.Service{}, c.svcStores, eventCh), true)
-		informManager.extend(c.WatchObjects(ns, kindEndpoints, &corev1.Endpoints{}, c.epStores, eventCh), true)
-		// Do not wait for the Secrets store to get synced since we cannot rely on
-		// users having granted RBAC permissions for this object.
-		// https://github.com/containous/traefik/issues/1784 should improve the
-		// situation here in the future.
-		informManager.extend(c.WatchObjects(ns, kindSecrets, &corev1.Secret{}, c.secStores, eventCh), false)
```
timoreimann: We pass `false` as the last parameter so that we don't wait for the Secrets store to become synchronized for the described reason.

Am I missing something, or did we not carry over that behavior?

provider/kubernetes/client.go
```diff
@@ -150,31 +124,28 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
+	for _, ns := range namespaces {
+		factory := informers.NewFilteredSharedInformerFactory(c.clientset, resyncPeriod, ns, func(opts *metav1.ListOptions) {
+			opts.LabelSelector = labelSelector
+		})
+		factory.Extensions().V1beta1().Ingresses().Informer().AddEventHandler(eventHandler)
+		factory.Core().V1().Services().Informer().AddEventHandler(eventHandler)
+		factory.Core().V1().Endpoints().Informer().AddEventHandler(eventHandler)
+		factory.Core().V1().Secrets().Informer().AddEventHandler(eventHandler)
+
+		for t, ok := range factory.WaitForCacheSync(stopCh) {
```
timoreimann: The previous implementation first started all informers and then waited for the caches to sync. The new implementation does it the other way around.

Are we certain this doesn't change the observable behavior?

Gopkg.lock
```diff
@@ -1520,6 +1560,30 @@
...
+    "listers/networking/v1",
+    "listers/policy/v1beta1",
+    "listers/rbac/v1",
+    "listers/rbac/v1alpha1",
+    "listers/rbac/v1beta1",
+    "listers/scheduling/v1alpha1",
+    "listers/settings/v1alpha1",
+    "listers/storage/v1",
+    "listers/storage/v1alpha1",
+    "listers/storage/v1beta1",
```
timoreimann: Possibly more a question for @ldez: Is it normal that dep references packages which are not used? I thought dep only pulls in what is really needed.

provider/kubernetes/client.go
yue9944882: @timoreimann it will be `*v1.Secret` which is returned by `reflect.TypeOf(&corev1.Secret)`:
 [https://github.com/kubernetes/kubernetes/blob/aa02c0f5191af40cbf7a18a8beec24c56ed7b860/staging/src/k8s.io/client-go/informers/core/v1/secret.go#L84](https://github.com/kubernetes/kubernetes/blob/aa02c0f5191af40cbf7a18a8beec24c56ed7b860/staging/src/k8s.io/client-go/informers/core/v1/secret.go#L84)

Gopkg.lock
yue9944882: These are all supposed to be direct dependencies when compiling `informerFactory`. 
[https://github.com/kubernetes/kubernetes/blob/aa02c0f5191af40cbf7a18a8beec24c56ed7b860/staging/src/k8s.io/client-go/informers/factory.go#L42](https://github.com/kubernetes/kubernetes/blob/aa02c0f5191af40cbf7a18a8beec24c56ed7b860/staging/src/k8s.io/client-go/informers/factory.go#L42)
[https://github.com/kubernetes/kubernetes/blob/aa02c0f5191af40cbf7a18a8beec24c56ed7b860/staging/src/k8s.io/client-go/informers/core/v1/pod.go#L30](https://github.com/kubernetes/kubernetes/blob/aa02c0f5191af40cbf7a18a8beec24c56ed7b860/staging/src/k8s.io/client-go/informers/core/v1/pod.go#L30)

provider/kubernetes/client.go
```diff
@@ -150,31 +125,37 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
 	}
 
-	var wg sync.WaitGroup
-	for _, informer := range informManager.informers {
-		informer := informer
+	for _, ns := range namespaces {
 		safe.Go(func() {
 			wg.Add(1)
-			informer.Run(stopCh)
+			c.factories[ns].Start(stopCh)
```
timoreimann: We need to insert `ns := ns` after line 141 in order to capture the variable. Otherwise, we'd always end up with the last namespace being referenced in the anonymous goroutine (see also https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables).

provider/kubernetes/client.go
```diff
@@ -150,31 +125,37 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
 
-	if !cache.WaitForCacheSync(stopCh, informManager.syncFuncs...) {
-		return nil, fmt.Errorf("timed out waiting for controller caches to sync")
+	for _, ns := range namespaces {
+		for t, ok := range c.factories[ns].WaitForCacheSync(stopCh) {
+			// Do not wait for the Secrets store to get synced since we cannot rely on
+			// users having granted RBAC permissions for this object.
+			// https://github.com/containous/traefik/issues/1784 should improve the
+			// situation here in the future.
+			if t == reflect.TypeOf(&corev1.Secret{}) && !ok {
```
timoreimann: Nit-pick: Please swap the expression terms for better comprehensibility (and, to a neglectable degree, performance).

provider/kubernetes/client.go
```diff
@@ -186,93 +167,48 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
 func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 	var result []*extensionsv1beta1.Ingress
-
-	for _, store := range c.ingStores {
-		for _, obj := range store.List() {
-			ing := obj.(*extensionsv1beta1.Ingress)
+	for ns, factory := range c.factories {
+		ings, err := factory.Extensions().V1beta1().Ingresses().Lister().List(labels.Everything())
+		if err != nil {
+			log.Errorf("fail to list ingresses in namespace %s: %s", ns, err)
```
timoreimann: s/fail/Failed/

provider/kubernetes/client.go
timoreimann: In the previous implementation, we did not wait for the Secrets' cache to sync at all. The new implementation seems to wait for it but does not consider an error to be critical.

Assuming I got it right: Does this possibly mean that users without permitted RBAC access now see Secrets RBAC errors in their logs? I guess I'm okay with a single log occurrence, but multiple ones would be irritating IMHO.

Gopkg.lock
timoreimann: 👌

provider/kubernetes/client.go
yue9944882: I just realise that `informerFactory.Start` is an asynchronous call itself and we don't have to invoke it via `safe.Go(...)`

provider/kubernetes/client.go
```diff
@@ -150,129 +122,84 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
-	if !cache.WaitForCacheSync(stopCh, informManager.syncFuncs...) {
-		return nil, fmt.Errorf("timed out waiting for controller caches to sync")
+	for _, ns := range namespaces {
+		c.factories[ns].Start(stopCh)
 	}
 
-	safe.Go(func() {
-		<-stopCh
-		wg.Wait()
-		close(eventCh)
```
yue9944882: @timoreimann  What would happen if we don't close `eventCh` gracefully?

yue9944882:
xref of Rbac issue
[https://github.com/kubernetes/kubernetes/issues/57074#issuecomment-369170876](https://github.com/kubernetes/kubernetes/issues/57074#issuecomment-369170876)

provider/kubernetes/client.go
```diff
@@ -150,129 +122,84 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
-		object,
-		resyncPeriod,
-	)
+	// Do not wait for the Secrets store to get synced since we cannot rely on
+	// users having granted RBAC permissions for this object.
+	// https://github.com/containous/traefik/issues/1784 should improve the
+	// situation here in the future.
+	for _, ns := range namespaces {
+		c.factories[ns].Core().V1().Secrets().Informer().AddEventHandler(eventHandler)
+		c.factories[ns].Start(stopCh)
```
timoreimann: I'm not sure I get this. Aren't we already adding the event handler and starting the Secrets factory in lines 133 and line 138, respectively?

provider/kubernetes/client.go
```diff
@@ -150,129 +122,84 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, labelSelector string, stopC
...
-func loadInformer(listWatch cache.ListerWatcher, object runtime.Object, watchCh chan<- interface{}) cache.SharedInformer {
-	informer := cache.NewSharedInformer(
-		listWatch,
-		object,
-		resyncPeriod,
-	)
+	// Do not wait for the Secrets store to get synced since we cannot rely on
+	// users having granted RBAC permissions for this object.
+	// https://github.com/containous/traefik/issues/1784 should improve the
+	// situation here in the future.
```
timoreimann: Is the comment true and correctly placed here? It looks like lines 141 to 142 are including Secrets, or am I missing something?

provider/kubernetes/client.go
timoreimann: Was this addressed?

provider/kubernetes/client.go
timoreimann: From what I can tell, we should be okay not closing it: The latest refactorings seem to have made it obsolete.

I'll do some extra manual tests once the PR is complete.

provider/kubernetes/client.go
yue9944882: @timoreimann Oh it's my mistake. I should remove `c.factories[ns].Core().V1().Secrets().Informer().AddEventHandler(eventHandler)` from L133

provider/kubernetes/client.go
yue9944882: Yes and we skip the synchronization of secret by delaying the invocation of `AddEventHandler`.

timoreimann:
Apologies for the laggy response; I will try hard to get back to this PR and do some final testings within the next 2 days.

Sorry!

yue9944882:
@timoreimann Never mind. The refactoring in this PR might be a big change so we ought to spend more time to prove its correctness. I'd also like to help enhance test for this PR :-)

yue9944882:
@mmatur What does it mean by `need-human-merge`? Am I missing sth? (facepalm

mmatur:
@yue9944882, you have missed nothing. We have bots that help us to manage PR.

Explanation available [here](https://github.com/containous/traefik/blob/master/MAINTAINER.md#myrmica-lobicornis)

PR #2871, opened 2018-02-16
Smooth dashboard refresh.

ldez:
### What does this PR do?

Refresh UI only if providers data change.

### Motivation

Facilitate copy-paste.


mstenz:
Thx, this is really a cool fix. This copy & paste problem annoided me a lot of times.

Issue #2710, opened 2018-01-16
Feature Request: multiple acme entry points


raarts:
### Do you want to request a *feature* or report a *bug*?

feature request

### What did you do?

It's currently impossible to share acme certificates between entrypoints.
See: https://stackoverflow.com/questions/47277489/traefik-multiple-https-entry-points-with-acme-letsencrypt-dns

Something similar to this:

    --acme.entryPoints=[https,stream] \
    --acme.entryPoints.https.dnsProvider=digitalocean \
    --acme.entryPoints.stream.dnsProvider=vultr \

isrjo:
+1

colinmollenhour:
+1 Example, I am running both https and etcd in readonly mode on one server on different ports (443 and 2379) but want them both to use Let's Encrypt and actually share the same certificate. Currently it looks like I'd have to run two instanced of Traefik but then unless I use a KV store then I'd have two separate certificate stores.

adyromantika:
👍

jnovack:
"multiple acme entry points", to me, means that I can have multiple values in `acme/entrypoint`.  It does not imply that sites should be able to available on multiple TLS entrypoints.

```
[acme]
entryPoints = "https1,https2"

[acme.httpChallenge]
    entryPoint = "http"
```

Using this configuration, I would expect sites (`Host:xxx.example.org`) to be XOR (exclusive OR, one or the other, not both) on entrypoints.

This means to me:

* `site1.example.org` can be presented on `http` and `https1` entrypoints, and
* `site2.example.org` can be presented on `http` and `https2` entrypoints, and
* `site1.example.org` CANNOT be available on `https2` entrypoint, and
* `site2.example.org` CANNOT be available on `https1` entrypoint.

I would expect to use the same FILE of certificates on both entrypoints, but I would not expect to use a certificate on multiple entrypoints.

This interpretation seems to be different from the feature request as stated, and probably easier.  I believe the original poster is asking for:

* `site1.example.org` can be presented on `http`, `https1`, and `https2` entrypoints, and
* `site2.example.org` can be presented on `http`, `https1`, and `https2` entrypoints

**Should I make a separate Feature Request for my interpretation?**

snowzach:
I opened PR 4345

Rush:
@snowzach I don't think your PR would work for all cases. My use case is:
- One IP is providing one set of services.
- Secondary IP is providing a second set of services.

[...]

PR #2881, opened 2018-02-17
Multiple ACME-enabled entrypoints

sjawhar:
### What does this PR do?
Adds a new `entryPoints` entry in the ACME config, replacing the existing `entryPoint` entry, so you can list multiple entrypoints for ACME. I tried to keep this backward-compatible to allow for deprecation at a future date.

### Motivation
Like #2710, I want to have ACME working on multiple entrypoints. In my case, I want to use a single Traefik instance to serve both external-facing and internal-only sites. Having the internal-only sites listen on a port that isn't exposed to the outside world is simpler than maintaining an IP whitelist, and wouldn't be vulnerable to IP spoofing. But currently only a single entrypoint can use ACME. I swear there was an issue that mentioned this exact use case, but I can't find it now...

### More
- [ ] Added/updated tests

### Additional Notes
I haven't added new tests yet, but I checked that I didn't break any of the existing unit tests. Once I get the go-ahead from you guys on my approach, I'll add tests. Also I've pretty much never written anything in Go before, so just bear with me and I'll do my best to correct my foolish mistakes 😄

Some thoughts:
- How does this affect HTTP ACME challenges? I've only ever used DNS challenges.


sjawhar:
Any hints on where I should focus to get this working? I'm committed to hunting down my mistakes, if you can offer any guidance.

nmengin:
Hello @sjawhar.

Thanks for your Pull Request and your interest in Træfik.
We're sorry for the delay but we had to analyze your Pull Request deeply before to give you a feedback.

Indeed, your Pull request addresses a very interesting need and we tested it a lot in the way to know if the modifications takes care of all the Use Cases.

The analysis we done shows that the modifications you proposed can not solve entirely and correctly the need :
    
* HTTP-01 challenge is not modiffied, but you indicated it in the PR and it's not the more important,
* There is no modification to handle the storage of certificate by entrypoints.
* The PR does not allow to correctly check, per EntryPoint, the domains which need an ACME certificate. This Use Case is described in this configuration :

```toml
[entrypoints]
[entrypoints.secure1]
address=":443"
[entrypoints.secure1.tls]

[entrypoints.secure2]
address=":8443"
[entrypoints.secure2.tls]
  [[entryPoints.secure2.tls.certificates]]
  certFile = "path/to/foo.com.cert"  # self-signed certificate for foo.com
  keyFile = "path/to/foo.com.key"   # self-signed certificate for foo.com
```

In case of an `onHostRule` on `foo.com`, you must handle that ACME only creates certificate for `secure1`. But, your PR will create an ACME certificate which will be shared between both EntryPoints.
As, in order to handle this, it will need big changes which will be difficult to add to the Pull Request.

For all these reasons, we propose to close this PR for the moment. Feel free to open a new Pull Request when you think you handle everything.

Moreover, a new Pull Request is just opened a new PR (#2889) which is a refactoring of all the ACME mechanism.
The modification introduced can help to add more easily the feature you proposed.

You can join us on [slack](https://traefik.herokuapp.com/) if you want to talk more about this.

Is it OK for you?

sjawhar:
OK, I'll start over based on the ACME provider changes. My changes were small anyway.

ViViDboarder:
I was thinking of taking a crack at this.

Is it necessary to have a different certificate store per entrypoint? It's possible to share certs between entrypoints stored as standard cert files.

PR #2882, opened 2018-02-17
Check for endpoints while detecting Consul service changes

caseycs:
### What does this PR do?

Check for endpoint ip address while detecting Consul catalog changes

### Motivation

I'm experimenting with Docker Swarm, Registrator and Consul to implement service discovery in my project, and turned out, that scaling or updating an image for  a service does not trigger Traefik to refresh back-ends.  

After some investigation I figured out, that it's looking only on diff of unique node ids (that's why scaling a service on the same node never trigger update) and does not check for endpoint ip changes at all.

### More

- [x] Added/updated tests

### Additional Notes

Nothing else to mention.

### DIFF
--- a/provider/consulcatalog/consul_catalog.go
+++ b/provider/consulcatalog/consul_catalog.go
@@ -38,10 +38,11 @@ type Provider struct {
 
 // Service represent a Consul service.
 type Service struct {
-	Name  string
-	Tags  []string
-	Nodes []string
-	Ports []int
+	Name      string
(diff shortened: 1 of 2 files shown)


ldez:
@caseycs The review process: https://github.com/containous/traefik/blob/master/MAINTAINER.md#pr-review-process

Please don't merge master.

caseycs:
@ldez thanks, I missed this part.

PR #2913, opened 2018-02-22
Check all the C/N and SANs of provided certificates before to generat…

### LINKED ISSUE #473: Let's encrypt sans handling
Hi

I am using traefik on a rpi with one domain and multiple sans. 
Instead of having one certificate request with the domain and sans I have one request per sans leading to  exceed the certificate request limit.

nmengin:
### What does this PR do?

This PR fixes ACME certificates generation.
So far, only C/N of provided certificates was checked before to generate ACME certificate.

This PR adds a check on the SANs and generates ACME certificates only for domains which need it.

### Motivation

Only generate ACME certificates for domains which need it.
Fixes #473 

### More

- [x] Added/updated tests

### DIFF
--- a/acme/account.go
+++ b/acme/account.go
@@ -222,6 +222,24 @@ func (dc *DomainsCertificates) exists(domainToFind Domain) (*DomainsCertificate,
 	return nil, false
 }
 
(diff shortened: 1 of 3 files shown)


dtomcej:
This looks awesome. Would also close #473 correct?

PR #2914, opened 2018-02-22
Remove hardcoded frontend prefix in Kubernetes template

psalaberria002:
### What does this PR do?

Fixes a bug related to k8 annotations and template generation.

The hardcoded `frontend` prefix forces the generation of new frontends instead of adding the rate limiting or errors into the existing frontend.

### Motivation

I am trying to run the latest version from master in GKE, and rate limiting was not working.

### Additional Notes

Was there any specific reason for adding the `frontend` prefix @ldez ?


ldez:
Just a bad copy paste :smile:

PR #2946, opened 2018-02-28
Use responseModifier to override secure headers

### LINKED ISSUE #2618: Security headers don't overwrite upstream headers

bug

### What did you do?

Use docker security headers.

The upstream headers get replaced by the traefik version

### What did you see instead?

Duplicate headers

```
Version:      v1.5.0-rc3
Codename:     cancoillotte
Go version:   go1.9.2
Built:        2017-12-20_02:24:44PM
OS/Arch:      linux/amd64
```

My upstream server returns a XFO header, which I'm looking to override:

`Referrer-Policy: same-origin`

This is the minimal configuration I'd expect to work:

```tf
"traefik.frontend.headers.referrerPolicy" = "no-referrer"
```

I get double headers with this.

```
curl https://wiki.bb8.fun/ -I
HTTP/2 200 
referrer-policy: no-referrer
referrer-policy: same-origin
```

 Next I attempted to "clear" the upstream header:

```tf
"traefik.frontend.headers.referrerPolicy" = "no-referrer"
"traefik.frontend.headers.customResponseHeaders" = "Referrer-Policy:"
```

which works but looks unclean. My final solution is to just hand-roll the entire thing:

```tf
"traefik.frontend.headers.customResponseHeaders" = "Referrer-Policy:${var.refpolicy}"
```

At which point, I'm better off without the Traefik Docker Security header features. Most of the Security headers won't accept conflicting/duplicate values in the browser, and I think just overwriting them always makes more sense.

This is also a configuration issue, since now I have to maintain different config on each of my apps depending on whether the application already exposes a security header or not.

mmatur:
### What does this PR do?

This PR upgrade `unrolled/secure` dependency to be able to use `ModifyResponse` function to override backend headers

### Motivation

Fixes #2618 

### More

- [x] Added/updated tests

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1163,9 +1163,10 @@
   revision = "50716a0a853771bb36bfce61a45cdefdb98c2e6e"
 
 [[projects]]
+  branch = "v1"
   name = "github.com/unrolled/secure"
   packages = ["."]
-  revision = "824e85271811af89640ea25620c67f6c2eed987e"
+  revision = "7fac75837c196a7194c2533f803f3389c803472d"
 
 [[projects]]
   name = "github.com/urfave/negroni"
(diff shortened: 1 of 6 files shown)


captn3m0:
Thanks :100:

PR #2958, opened 2018-03-02
New option in secure middleware

mmatur:
### What does this PR do?

This PR add a new option in secure middleware to be able to set a custom value for header ` X-XSS-Protection`

### More

- [x] Added/updated tests


provider/docker/config_container_swarm_test.go
```diff
@@ -316,7 +318,7 @@ func TestSwarmBuildConfiguration(t *testing.T) {
 	for _, test := range testCases {
 		test := test
 		t.Run(test.desc, func(t *testing.T) {
-			t.Parallel()
+			//t.Parallel()
```
ldez: could you restore this line.

PR #2962, opened 2018-03-02
Fix Rancher Healthcheck when upgrading a service

### LINKED ISSUE #2279: Rancher healthcheck doesn't work as expected
bug

### What did you do?
Start traefik with racnher metadata. With enabling AND disabling healtcheack.

1) If healtcheck is disabled and service is updating/updated, i see service rule on the traefik web admin. And servise opend in browser.

2) if healtcheck is enabled. I want to see the same behavior at this point.

### What did you see instead?

1) If healtcheck is disabled and when service is updating i see 2 servers on service backend by 1-2 seconds. But one of this servers is no longer works. (stoped by updateding when i use "start before stoping" at racnher). At browser i see Gateway Timeout.
Perhaps this is the correct behavior?

2) if heathcheck is enabled and service updating/updated with "start before stoping" i did not see any rules of this service on traefik admin. When i confirm update in rancher, traefik show new rules 

At rancher api i see:
```
"state": "upgraded",
"healthState": "healthy",
```
But if restart traefik. I will see service rules in web admin.

```
Version:      v1.4.0
Codename:     roquefort
Go version:   go1.9.1
Built:        2017-10-16_04:52:07PM
OS/Arch:      linux/amd64
```

```toml
[rancher]
domain = "dev.rz"
Watch = true
RefreshSeconds = 15
ExposedByDefault = false
EnableServiceHealthFilter = false/true (!!!)

[rancher.metadata]
IntervalPoll = false
Prefix = "/latest"
```

jmirc:
### What does this PR do?

Here is the small change I did to fix the issue. I tested locally and it works.

### Motivation

Fixes #2279

### More

- [ ] Added/updated tests

### Additional Notes

It is my first go changes so I am sure this could be better.

### DIFF
--- a/provider/rancher/rancher.go
+++ b/provider/rancher/rancher.go
@@ -283,7 +283,7 @@ func containerFilter(name, healthState, state string) bool {
 		return false
 	}
 
-	if state != "" && state != "running" && state != "updating-running" {
+	if state != "" && state != "running" && state != "updating-running" && state != "upgraded" && state != "upgrading" {
 		log.Debugf("Filtering container %s with state of %s", name, state)
 		return false
 	}
@@ -319,7 +319,7 @@ func (p *Provider) serviceFilter(service rancherData) bool {
 			return false
 		}
 
-		if service.State != "" && service.State != "active" && service.State != "updating-active" && service.State != "upgraded" {
+		if service.State != "" && service.State != "active" && service.State != "updating-active" && service.State != "upgraded" && service.State != "upgrading" {
 			log.Debugf("Filtering service %s with state of %s", service.Name, service.State)
 			return false
 		}
--- a/provider/rancher/rancher.go
+++ b/provider/rancher/rancher.go
@@ -283,7 +283,7 @@ func containerFilter(name, healthState, state string) bool {
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


nmengin:
WDYT @containous/rancher ?

jmirc:
Hi @SantoDE ,

I am not filtering a service with the upgrading state because it doesn't mean the service can't receive load. An upgrading state is just meaning that the service is upgrading. The solution to avoid any downtimes of your service is to check the "Start before stopping" checkbox. With this option, a new container is started and once it is running the previous version is stopped. It doesn't mean the service can't receive any calls.

Actually, the service is filtering when the state is upgrading, that's mean a call to this service will return a 404 which is not what we want.

Let me know if it is not clear enough.

SantoDE:
Hi @jmirc,

okay, then it's what I expected :) 

For Services, I'm fine to pass them traffic as long as there healthstate is upgrading. However, I'm unsure about not filtering containers which state is upgrading. Imho, it would be better to not pass them traffic until they are healthy.

WDYT?

jmirc:
Hi SantoDE,

I agree with you. We should not pass traefik to the container with an upgrading state

Jerome

PR #2973, opened 2018-03-06
Add some missing quotes in templates

ldez:
### What does this PR do?

Add some missing quotes in templates

### Motivation

Be safe with Unicode characters.

### DIFF
--- a/templates/docker.tmpl
+++ b/templates/docker.tmpl
@@ -28,7 +28,7 @@
 
   {{ $healthCheck := getHealthCheck $backend }}
   {{if $healthCheck }}
-  [backends.backend-{{ $backendName }}.healthCheck]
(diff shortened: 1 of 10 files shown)


templates/docker.tmpl
```diff
@@ -223,7 +223,7 @@
     {{if $errorPages }}
     [frontends."frontend-{{ $frontendName }}".errors]
       {{range $pageName, $page := $errorPages }}
-      [frontends."frontend-{{ $frontendName }}".errors.{{ $pageName }}]
+      [frontends."frontend-{{ $frontendName }}".errors."{{ $pageName }}]
```
dtomcej: Are you missing a quote here?

ldez: Nice catch !

Issue #2722, opened 2018-01-18
PathPrefixStrip Is not working as intended


ray-matos-mpf:
Given the follow compose file

```yml
  apiv1:
    image: wouterla/whoami
    networks:
    - proxy
    environment:
      - WHOAMIserviceName=apiv1
    deploy:
      replicas: 2
      labels:
      - traefik.frontend.rule=PathPrefixStrip:/api/v1
      - traefik.priority = 1
      - traefik.docker.network=proxy
      - traefik.port=8000
      - traefik.enable=true
  apiv2:
    image: wouterla/whoami
    networks:
    - proxy
    environment:
      - WHOAMIserviceName=apiv2
    deploy:
      replicas: 2
      labels:
      - traefik.frontend.rule=PathPrefixStrip:/api/v2;/api
      - traefik.priority = 2
      - traefik.docker.network=proxy
      - traefik.port=8000
      - traefik.enable=true
```

Expected
 - /api and /api/v2 hits apiv2 backend 
 - /api/v1 hits the apiv1 backend

Actual
 - /api returns 404
 - /api/v2 hits apiv2 backend 
 - /api/v1 hits apiv1 backend

I also tried 
```traefik.frontend.rule=PathPrefixStrip:/api/v2,/api```
But this resolves all routes to apiv2

Thanks in advance

nmengin:
Hello @ray-matos-mpf .

Many thanks for your interest in the project.

Can you send us the version of Træfik you use?

Thanks in advance.

nmengin:
Hello @ray-matos-mpf ,

We analyzed deeply your issue and it appears it contains mistakes which explain why it does not work as expected.

- The labels `traefik.priority` have to be renamed into `traefik.frontend.priority`,
- The spaces in the right part of the labels `traefik.frontend.priority` have to be removed
- The `;` in the second rule has to be replaced by a `,`
- The priority are not specified correctly, you can refer to [the documentation](https://docs.traefik.io/basics/#priorities) for more details.

The configuration described below contains all the modification and works well : 

```yaml
  apiv1:
    image: wouterla/whoami
    networks:
    - proxy
    environment:
      - WHOAMIserviceName=apiv1
    deploy:
      replicas: 2
      labels:
      - traefik.frontend.rule=PathPrefixStrip:/api/v1
      - traefik.frontend.priority=5
      - traefik.docker.network=proxy
      - traefik.port=8000
      - traefik.enable=true
  apiv2:
    image: wouterla/whoami
[...]

PR #2984, opened 2018-03-08
Clarify how setting a frontend priority works

jbdoumenjou:
### What does this PR do?

Update the documentation to clarify how setting a frontend priority works

### Motivation

Some misunderstanding on the frontend priority mechanism
Related to #2722 

### More

- [x] Added/updated documentations

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -234,7 +234,7 @@ The following rules are both `Matchers` and `Modifiers`, so the `Matcher` portio
 #### Priorities
 
 By default, routes will be sorted (in descending order) using rules length (to avoid path overlap):
-`PathPrefix:/12;Host:foo.com` (length == 27) will be matched before `PathPrefixStrip:/123` (length == 20) will be matched before `PathPrefix:/12,/3` (length == 17).
+`PathPrefix:/foo;Host:foo.com` (length == 28) will be matched before `PathPrefixStrip:/foobar` (length == 23) will be matched before `PathPrefix:/foo,/bar` (length == 20).
 
 You can customize priority by frontend. The priority value is added to the rule length during sorting:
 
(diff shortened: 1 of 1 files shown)


docs/basics.md
```diff
@@ -234,7 +234,7 @@ The following rules are both `Matchers` and `Modifiers`, so the `Matcher` portio
 #### Priorities
 
 By default, routes will be sorted (in descending order) using rules length (to avoid path overlap):
-`PathPrefix:/12345` will be matched before `PathPrefix:/1234` that will be matched before `PathPrefix:/1`.
+`PathPrefix:/12;Host:foo.com` (length == 27) will be matched before `PathPrefixStrip:/123` (length == 20) will be matched before `PathPrefix:/12,/3` (length == 17).
```
ldez: could you replace numbers by letters.

ex: `PathPrefix:/12` -> `PathPrefix:/foo`

docs/basics.md
```diff
@@ -254,7 +254,7 @@ You can customize priority by frontend. The priority value is added to the rule
       rule = "PathPrefix:/toto"
 ```
 
-Here, `frontend1` will be matched before `frontend2` (`(3 + 10 == 13) > (4 + 5 == 9)`).
+Here, `frontend1` will be matched before `frontend2` (`(14 + 10 == 24) > (16 + 5 == 21)`).
```
Juliens: I think that in reality, if you specify a priority, you don't use the length and use directly the value. Can you verify this ?

timoreimann: In Traefik 1.5, a custom priority _does_ ignore the length-based factor according to my tests not too long ago.

jbdoumenjou: Yes, I made a mistake, the priority override the rule length according to https://github.com/containous/traefik/blob/master/server/server.go#L1192

Issue #2875, opened 2018-02-16
Ultimate Access log filter


mmatur:
### Do you want to request a *feature* or report a *bug*?

Feature

## Description
This issue is a proposal following PRs #2853 and #2397 on modifying access logs. 

```toml
[accessLog]
filepath = "..."
type = "..."

# filter access log line
[accessLog.filter]
# Keep on only access logs with status code in the range
statusCode = ["..."]

[accessLog.fields]
  defaultMode = "drop|keep"
  [accessLog.fields.names]
    host = "drop"
    # ...

  [accessLog.fields.headers]
   defaultMode = "drop|keep|redact"
   [accessLog.fields.headers.names]
     User-Agent = "redact"
     Authorization = "drop"
     Content-Type = "keep"
     # ...
```

## Filter

`statusCode` range format:
    
- `["200"]` :  only page with status code 200
- `["400-499"]`: only page with status code between 400 and 499
- `["200", "400-499"]`: page with status code 200 or status code between 400 and 499

#### Fields

In this example we will drop all fields except `DownstreamStatus`:
    
```toml
[accessLog.fields]
  defaultMode = "drop"
  [accessLog.fields.names]
    DownstreamStatus = "keep"
```

In this  example we will keep all fields except `DownstreamStatus`:
    
```toml
[accessLog.fields]
  defaultMode = "keep"
  [accessLog.fields.names]
    DownstreamStatus = "drop"
```

All available fields:

- `BackendAddr`
- `BackendName`
- `BackendURL`
- `ClientAddr`
- `ClientHost`
- `ClientPort`
- `ClientUsername`
- `DownstreamContentSize`
- `DownstreamStatus`
- `DownstreamStatusLine`
- `Duration`
- `FrontendName`
- `OriginContentSize`
- `OriginDuration`
- `OriginStatus`
- `OriginStatusLine`
- `Overhead`
- `RequestAddr`
- `RequestContentSize`
- `RequestCount`
- `RequestHost`
- `RequestLine`
- `RequestMethod`
- `RequestPath`
- `RequestPort`
- `RequestProtocol`
- `RetryAttempts`
- `StartLocal`
- `StartUTC`
- `level`
- `msg`
- `time`

### Headers

In this example we will drop all headers except:
- `User-Agent` that will be redacted
- `Authorization` that will be kept with original data
- `Content-Type` that will be kept with original data

```toml
[accessLog.fields.headers]
defaultMode = "drop"
  [accessLog.fields.headers.names]
[...]

Issue #2212, opened 2017-10-04
Manage deprecation


ldez:
### Do you want to request a *feature* or report a *bug*?

proposal

### What did you do?

Deprecated in the future:

- [x] `EntryPoint.Compress`: will become a struct
- [x] `ACME.OnDemand`: will be removed, use `OnHostRule` instead
- [x] By default, don't trust given `X-Forwarded-For`, unless in a whitelist
- [x] `0` server weight disables it (see https://github.com/containous/traefik/pull/1780#issuecomment-335317020)
- [x] Etcd API V2 : `etcd.useAPIV3` will be removed and Etcd v3 become the default.
- [x] suppress old basic auth, move `HeaderField` into basic/Digest auth.
- [x] don't allow dot in the segment name (label)
- [x] drop redundant log fields (ex: `OriginStatusLine` which is just a representation of `OriginContentSize`, `OriginDuration`, and `OriginStatus`)
- [x] remove `passHostHeader` in with KV  (use lower-case key instead)
- [x] replace `http` and `https` as entry point name by `web` and `websecure`

dottgonzo:
why ondemand will be deprecated? is useful for multiple traefik instances, where only the route on try to renew it's certificate, instead of every traefik instances.

nmengin:
Hello @dottgonzo.

> is useful for multiple traefik instances

I guess you want using multiple instances of Træfik in a cluster?
If you want it, the configuration is different.

indeed, you have to use a KV store which will save and share your configuration and your ACME certificates as desbribed in the [documentation](http://v1-5.archive.docs.traefik.io/user-guide/kv-config/#store-configuration-in-key-value-store) and this [user guide](http://v1-5.archive.docs.traefik.io/user-guide/cluster-docker-consul/).

Thanks to this cluster configuration, only the leader instance will create and renew the certificates. The option `onDemand` is not required in this use case.  

WDYT?

emilevauge:
@dottgonzo `ACME.OnDemand` has nothing to do with your cluster issue. I suggest we move the swarm mode + Consul + Traefik discussion from this Github issue to the #support channel on Slack. Thanks :)

nmengin:
The `acme.onDemand` option is deprecated and has to be deleted in the future because of all the problems it can induce.
When ACME has been introduced in Træfik, this option was created to allow users to create easily all their ACME certificates **once** before restarting Træfik without this option.

Indeed, keeping the `onDemand` option has many disadvantages :

[...]

PR #2988, opened 2018-03-09
Ultimate Access log filter

mmatur:
### What does this PR do?

This PR is an implementation of #2875.

### Motivation

Be able to customize accesslog like it was asked over PRs #2853 and #2397

Closes #2853, #2397, #2875

### More

- [x] Added/updated tests


types/logs_test.go
```diff
@@ -0,0 +1,403 @@
...
+)
+
+func TestStatusCodesSet(t *testing.T) {
+	tests := []struct {
+		desc     string
+		value    string
+		expected StatusCodes
+	}{
+		{
+			desc:     "One value should return StatusCodes of size 1",
```
marco-jantke: I think the "should return StatusCodes ..." stuff can be dropped from the test descriptions.

types/logs_test.go
```diff
@@ -0,0 +1,403 @@
...
+
+			var statusCodes StatusCodes
+			err := statusCodes.Set(test.value)
+			assert.Nil(t, err)
+			assert.Equal(t, test.expected, statusCodes)
+		})
+	}
+}
+
+func TestStatusCodesGet(t *testing.T) {
```
marco-jantke: Do we really need to test the Getter? :)

types/logs_test.go
ldez: I don't agree with that.

mmatur:
@marco-jantke Thanks for your review. I just push some fixes :)

types/logs_test.go
marco-jantke: I'm not super hot about this one, I just think it's good to decrease optical "noise" and redundant explanations.

types/logs_test.go
marco-jantke: Same here, this won't block my approval but adding ~30 lines of code to test one line that it is in fact super trivial and always does the same doesn't seem worth it.

types/logs.go
```diff
@@ -0,0 +1,185 @@
...
+func (f *FieldsHeadersNames) SetValue(val interface{}) {
+	*f = val.(FieldsHeadersNames)
+}
+
+// Keep check if the field need to be kept or dropped
+func (f *Fields) Keep(field string) bool {
+	defaultKeep := true
+	if f != nil {
+		defaultKeep = checkFieldValue(f.DefaultMode, defaultKeep)
+		v, ok := f.Names[field]
```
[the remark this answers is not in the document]
marco-jantke: An oversight here or do you disagree?

marco-jantke:
@mmatur one thing I just missed/forgot again. What do you think about the comment I brought up now two times to drop redundant fields? See https://github.com/containous/traefik/pull/2988#pullrequestreview-103439929

danieladams456:
@marco-jantke I am in favor. Seems like that would be follow-up PR, right?

ldez:
@marco-jantke @danieladams456  we think this be breaking for people who parse the log output.
For now, we will not change fields.

I added the need in this issue https://github.com/containous/traefik/issues/2212

PR #3007, opened 2018-03-13
Use context in Server

Juliens:
### What does this PR do?
Use `context` when the server Start and then, `cancel` the `context` when we receive signals ( SIGINT, SIGTERM ).
Close timeout is no longer `RequestAcceptGraceTimeout` and if the Close timeout, we `panic` instead of os.Exit

### Motivation
Separation of concern

### Additional Notes
Close timeout is a timeout to wait for stop of goroutines (provider and metrics). So it make no sense to configure this timeout. Moreover if we failed to stop goroutines, it's a real problem and os.Exit is not enough verbose.


server/server_signals_windows.go
```diff
@@ -1,26 +0,0 @@
...
-
-import (
-	"os/signal"
-	"syscall"
-
-	"github.com/containous/traefik/log"
-)
-
-func (s *Server) configureSignals() {
-	signal.Notify(s.signals, syscall.SIGINT, syscall.SIGTERM)
```
ldez: You cannot remove this file

server/context.go
```diff
@@ -0,0 +1,22 @@
...
+	"os"
+	"os/signal"
+	"syscall"
+)
+
+// ContextWithSignal create a context cancelled when SIGINT or SIGTERM are notified
+func ContextWithSignal(ctx context.Context) context.Context {
+	newCtx, cancel := context.WithCancel(ctx)
+	signals := make(chan os.Signal)
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
```
ldez: It will never work on Windows

server/server.go
```diff
@@ -205,6 +205,23 @@ func (s *Server) Start() {
 	go s.listenSignals()
 }
 
+// StartWithContext starts the server and Stop/Close it when context is Done
+func (s *Server) StartWithContext(ctx context.Context) {
+	go func() {
+		defer s.Close()
+		<-ctx.Done()
+		log.Infof("I have to go...")
```
mmatur: Could you please use a `log.Info` instead of `log.Infof`

cmd/traefik/traefik.go
```diff
@@ -173,7 +174,8 @@ func runCmd(globalConfiguration *configuration.GlobalConfiguration, configFile s
 		acme.Get().SetConfigListenerChan(make(chan types.Configuration))
 		svr.AddListener(acme.Get().ListenConfiguration)
 	}
-	svr.Start()
+	ctx := server.ContextWithSignal(context.Background())
```
vdemeester: should this be _just_ in the main (or in `cmd/traefik` in that case). The `server` package doesn't really have to care about signal handling.

server/server.go
```diff
@@ -235,14 +252,13 @@ func (s *Server) Stop() {
 
 // Close destroys the server
 func (s *Server) Close() {
-	ctx, cancel := context.WithTimeout(context.Background(), time.Duration(s.globalConfiguration.LifeCycle.GraceTimeOut))
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
```
vdemeester: Why the *magic* time number here instead of using `GraceTimeOut` ?

Juliens: In fact, this timeout is for the stop of goroutines in Providers and Metrics, not for `GraceTimeOut` on entry points.
Moreover it's a technical timeout, there is no reason to configure it.

timoreimann: Are we certain that those 10 seconds are always enough for providers to finish?

cmd/traefik/traefik.go
Juliens: I'm ok with this, I will move ContextWithSignal

server/server.go
Juliens: In reality, I reuse the previous DefaultValue of GraceTimeOut. So, if you don't override the GraceTimeout, it will not really change the behavior. Moreover, as I change the os.Exit to panic, I think we will have more feedback if some goroutines never end or needs a bigger timeout (without change the "real" behavior, because os.Exit or panic will only change the "logs".

timoreimann: Where is `GraceTimeOut` being used now? I see that the PR removes an existing usage but does not add it anywhere again.

Juliens: `GraceTimeOut` is used when we close entrypoints. https://github.com/containous/traefik/blob/526a04d4c8c0ad042c69c3ec7c48ab373e97627b/server/server.go#L221

PR #3009, opened 2018-03-13
Added cluster/leader endpoint

aantono:
### What does this PR do?

This PR introduces a `/cluster/leader` endpoint, which will return `200-OK` or `429-Too-Many-Requests` status, depending on if the particular node is a leader or not.  The corresponding json payload will also be returned with `{"leader":"true"}` in case the node is a leader and `{"leader":"false"}` if it is not.

### Motivation

This will enable creating various Load Balancers on top of a Traefik cluster, which will only route API requests to the current leader, instead of randomly to any instance in the cluster.  The return of `429` status code will make most load balancers (like Netscalar or ELB, etc) to keep the instances in the pool, but "temporarily" not send any requests to them.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -740,6 +740,10 @@ func (s *Server) addInternalRoutes(entryPointName string, router *mux.Router) {
 	if s.globalConfiguration.API != nil && s.globalConfiguration.API.EntryPoint == entryPointName {
 		s.globalConfiguration.API.AddRoutes(router)
 	}
+
+	if s.leadership != nil {
+		s.leadership.AddRoutes(router)
+	}
 }
 
 func (s *Server) addInternalPublicRoutes(entryPointName string, router *mux.Router) {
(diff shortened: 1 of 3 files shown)


emilevauge:
Hey @aantono, could you explain why you need to access to the API only on the leader ?

aantono:
If you add all the instances to an ELB for their api port, you get fluctuations in the data, specifically the health/stats, because each request gets randomly sent to some instance.  Having only 1 instance (the leader) of the cluster be there would make things more consistent.  All of api requests would always be routed to the leader in that setup.  If one wants to have all the instances randomly routed, they can use `/ping` as a check instead of `/cluster/leader`

emilevauge:
@aantono so you are using the cluster in fail over mode right ?

aantono:
Yep.  Though we only use it in that mode for access to API/Dashboard endpoint, but it can be used in that mode for traffic routing as well, depending on which ports the ELB is proxying.

server/server.go
```diff
@@ -724,6 +724,10 @@ func (s *Server) addInternalRoutes(entryPointName string, router *mux.Router) {
 	if s.globalConfiguration.API != nil && s.globalConfiguration.API.EntryPoint == entryPointName {
 		s.globalConfiguration.API.AddRoutes(router)
 	}
+
+	if s.leadership != nil {
```
Juliens: You can't put this route on all entrypoints.
This route must be activated only on the entrypoint where we activate `--api`. So you need to put it in the API `if`.

Moreover, I think the endpoint should be /api/cluster/leader, WDYT?

PR #3012, opened 2018-03-14
Add TLS Docs

### LINKED ISSUE #3008: Update Kubernetes docs to notify users about backend TLS connections

feature

 The docs should have a note in the ingress section notifying users that if the service ports on their ingress is 443, TLS to the pods will be enabled automatically.

We should also make a reference to the `insecureSkipVerify` setting in the same note (as most people will not want to generate signed certs).

Master

dtomcej:
### What does this PR do?

Adds a note about how to enable TLS communication in k8s to backend pods.

### Motivation

Resolves #3008

### More

### DIFF
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -94,6 +94,13 @@ A label selector can be defined to filter on specific Ingress objects only.
 
 See [label-selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) for details.
 
+### TLS communication between Traefik and backend pods
+
+Traefik automatically requests endpoint information based on the service provided in the ingress spec. Although traefik will connect directly to the endpoints (pods), it still checks the service port to see if TLS communication is required. If the service port defined in the ingress spec is 443, then the backend communication protocol is assumed to be TLS, and will connect via TLS automatically.
+
+!!! note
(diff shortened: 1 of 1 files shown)


kachkaev:
Thanks for this addition @dtomcej – It'd really help me a few months back when I was figuring out things. WDYT of adding one sentence on how to get started with obtaining trusted backend certs? A link to the details would be ideal.

dtomcej:
@kachkaev I'm unsure if that would be wise, due to the fact that there are many different ways to get trusted certs, and they are environment specific. I am hesitant to suggest one method, as it anything specific that is suggested may be unsuitable for a large portion of users.

dtomcej:
Ideally, if you are managing the cluster, you could generate your own certs, and use the k8s CA to sign them (therefore making them trusted), but we can't assume that users have access to the cluster CA keys to sign with. There are also many users that use third party cert tools to handle this.

@timoreimann thoughts?

timoreimann:
I agree that we need to be careful about what we recommend in our documentation. For instance, our Kubernetes guide is often misinterpreted as providing production-level advice despite several warnings and notes left for quite some time.

We still have our awesome Traefik wiki list though where we could add community contributions.

timoreimann:
trying to hold back the bot for some final styling improvements by @nmengin .

PR #3017, opened 2018-03-15
Update grpc.md

dioptre:
### More

- [ ] Added/updated tests


emilevauge:
Hey @dioptre, thanks for your contribution!
We would prefer to not link a user repository in our documentation as it is impossible to maintain on the long term. We think it would be better to link this in [awesome.traefik.io](https://github.com/containous/traefik/wiki/Awesome-Traefik). WDYT ?

Juliens:
Thanks for your contribution!

Can you move this in awesome.traefik.io?

Issue #2969, opened 2018-03-06
Enable graceful rotation of traefik servers behind a load balancer


ravilr:
### Do you want to request a *feature* or report a *bug*?
feature

### What did you do?
Do a rolling-upgrade of several traefik instances all receiving a constant stream of requests through an external load-balancer.

### What did you expect to see?

No requests failures to be observed.

### What did you see instead?

Observed few request failures.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
v1.5.3
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

provider : Kubernetes  with lifecycle configuration defined
```toml
[lifeCycle]
  # Duration to keep accepting requests prior to initiating the graceful
  # termination period (as defined by the `graceTimeOut` option). This
  # option is meant to give downstream load-balancers sufficient time to
  # take Traefik out of rotation.
  requestAcceptGraceTimeout = "15s"

  # Duration to give active requests a chance to finish before Traefik stops.
  graceTimeOut = "15s"
```
### Feature request

Please see https://github.com/containous/traefik/issues/1942 for the detailed explanation of the feature and rationale.  
https://github.com/containous/traefik/pull/1971 added only `requestAcceptGraceTimeout`  without implementing item 1 : 'Toggle a pre-defined health check to artificially return unhealthy responses on SIGTERM during requestAcceptGraceTimout interval'.  https://github.com/containous/traefik/pull/2003 was closed. 

This issue is to enhance /ping handler to return non-200 response upon requestAcceptGraceTimeout initiation, iff it is set. 

/cc @timoreimann

PR #3031, opened 2018-03-16
Toggle /ping to artificially return unhealthy responses on SIGTERM during requestAcceptGraceTimout interval

ravilr:
### What does this PR do?

Enhance /ping handler to return non-200 response upon requestAcceptGraceTimeout initiation, iff it is set.

### Motivation

Fixes https://github.com/containous/traefik/issues/2969


timoreimann:
During #2003, we discussed if the new behavior should be toggleable by a switch. Back then, we decided it wasn't necessary. Reflecting on that, however, I think we need to have it.

Many orchestrators (including Marathon and Kubernetes) allow to configure dedicated backend endpoints to indicate aliveness ("is the backend still up and running?") and readiness ("is the backend ready to serve requests?"). Failure on the former usually means that a backend gets killed, while failure on the latter means it gets removed from rotation. With Traefik exposing a single endpoint only to signal both aliveness and readiness, such orchestrators cannot distinguish between the two properties anymore. The consequence is that if `/ping` returned 503 during the shutdown, an orchestrator using that endpoint as aliveness indicator might kill Traefik before it's able to shutdown gracefully. That would mess up with the gracefulness of the shutdown and potentially cause request loss.

I believe we should reconsider a switch (like my PR did) so that users running Traefik within an orchestrator can leave the switch off (the default behavior to maintain backwards compatibility) and rely on the orchestrator's built-in shutdown procedure, while those users running Traefik dedicatedly (such as the PR author) can enable the switch.

The alternative is have a dedicated endpoint (like `/ready`) which always expresses the desirable behavior (503 on SIGTERM). Again, in #2003 we decided against that because we did not want to expand on the number of endpoints. We may still need that extra endpoint at some point in the future though, should our readiness check ever desire to encompass more than a simple ping (e.g., only return 200 if at least a single valid configuration has been rendered). Internally at my organization we have been thinking about that feature and would have benefited from it already as a single configuration mistake can leave Traefik unusable (or unready) while still running. That said, my proposal to stick to a single endpoint but add a toggle would allow us to expand into that feature (by allowing to configure a custom endpoint).

@ravilr @emilevauge @Juliens WDYT?

ravilr:
@timoreimann I respectfully disagree. We are also running traefik as a ingress controller in Kubernetes.
All the external traffic is ingressed to the cluster pods through external VIP/Loadbalancer->traefik pods on hostNetwork->application pods.  In this mode, the Service readiness check on traefik pod is a no-op/immaterial. because, the traefik Kubernetes Service is not even used. 

I think, for setups, which are dependent on Kuberbetes traefik Service, they don't even need the feature in this PR. the traefik requestAcceptGraceTimeout is only useful for setups involving external LB. As such, if setups dependent on internal Kubernetes Service, they need not set requestAcceptGraceTimeout (or leave the default value '0'), then they don't get this behavior of /ping returning 503 on shutdown.

Also, note that in K8s, SIGTERM is sent to a pod only during shutdown sequence, on which the Endpoint Controller removes the pod from Kubernetes Service rotation.  So, your point about liveness probe failing isn't valid, because the pod is already under shutdown mode ( pod's deletionTimeStamp is already set, so liveness probe isn't even kicked of by Kubelet).

timoreimann:
@ravilr 

>Also, note that in K8s, SIGTERM is sent to a pod only during shutdown sequence, on which the Endpoint Controller removes the pod from Kubernetes Service rotation. So, your point about liveness probe failing isn't valid, because the pod is already under shutdown mode ( pod's deletionTimeStamp is already set, so liveness probe isn't even kicked of by Kubelet).

Thanks, I wasn't aware of that. It kills my entire chain or arguments. :-)

At least for Kubernetes. I haven't had time to test it for Marathon and wouldn't know for sure for other orchestrators. The behavior that Kubernetes expresses seems to be the reasonable one though.

Therefore, I'm all good with the PR as it is!

ravilr:
@marco-jantke Updated the ping.md documentation. Please review.

ravilr:
@ldez this pr is against 1.5 branch. should i open a pr against master branch also?  Please let me know.

ldez:
@ravilr could you rebase on master.

ravilr:
@ldez @marco-jantke i screwed up my branch upon rebasing to master. Can we please abandon this pr and move to https://github.com/containous/traefik/pull/3062

Issue #41, opened 2015-10-07
Drain a backend server


andrioid:
Many load-balancers have this feature.  Example for AWS: http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/config-conn-drain.html

I would like to be able to set /backends/awesomebackend/servers/server2/drain 1

The effect would be that no new clients should be sent towards this backend. Ideally the proxy should communicate back when no existing clients remain.
## Use Case

Deployment server could launch the new environment, set the old environment to "drain=1" and that way we could deploy without disturbing existing connections.

This sounds a bit out of scope for an auto-configured reverse-proxy, but if you're going to implement #5. Then maybe consider this.

EmileVauge:
Hi! This is in the pipeline :) What will be possible is to remove a server from a backend using the API. The existing connections will then we gracefully shutted down.

psi-4ward:
anything new here?

domino14:
How could this not be implemented? There's no way to have a zero-downtime deploy of backend services?

emilevauge:
@domino14 just do it! This is how open source works...

commarla:
I think @samber can do it easily ^^

timoreimann:
A workaround currently possible (at the price of delaying requests) is to enable retrying and make sure there are enough surplus servers available to compensate for the "draining" server.

Specific to the Kubernetes provider, the canonical way to achieve draining is to have a readiness probe implemented by the backing pods. Support from Traefik won't be necessary.

Marathon also supports readiness probes (called readiness checks there), but some degree of support might be needed in Traefik.

atecey:
Would it be possible to add a --drain=true flag to a Docker swarm service which Traefik picks up and doesn't route any new requests to.  The use case would be:
I have a service running (ServiceA version 1) using the frontend rule (mydomain.com).  I then bring online ServiceA version 2 running on the same frontend rule.  With the --drain flag specified on version 1 Traefik will stop routing new requests to it but honour existing connections.  All new requests will go to version 2 allowing zero downtime

KevDBG:
Hello,

Any update here ?

Thanks,

p-baum:
Any news yet or workaround ideas? How are people doing this????  They must be doing it somehow.  How about specifying a few IPs for the service and letting the load balancer manage it when you bring up/take down services on the IPs?

blakebuthod:
Any updates? Seems like this is heavily requested  🙂

[...]

PR #3034, opened 2018-03-16
Docker Swarm: Support for real time event listening (connection drain support).

### LINKED ISSUE #3035: Use event for Docker swarm mode instead of poll

feature

### What did you do?

Use swarm mode

no poll

### What did you see instead?

https://github.com/containous/traefik/blob/526c19181e9b34192efd4505c2a1928c9d59857b/provider/docker/docker.go#L159

```
v1.5
```

kristinn:
### What does this PR do?

These changes provide a support for load balancer draining for Docker Swarm. Note, the containers and services should also support graceful shutdowns.

This change makes sure Traefik stops routing, almost instantly, traffic to containers that are not in the "running" state.

We have backwards compatibility for Docker Swarm managers that don't offer live swarm events, by polling every 15 seconds (it's the same functionality as the current "master" branch offers).

### Motivation

We require a Docker Swarm load balancer that supports connection
draining.

Related to #41
Fixes #3035 

### Additional information

These changes do not break backwards compatibility.

These changes do not affect Traefik setups that are configured to route traffic using the internal Docker Swarm load balancing (IPVS). Traefik does not use the Docker Swarm load balancing by default (does not matter if Traefik is running with swarm mode set to true or not).

#### Stress testing results

Results from some tests I did locally on my Swarm cluster, using the official Traefik Docker image from the date of the testing (15th of March 2018), versus the patched Traefik binary. The file names describe what is being tested.

https://gist.github.com/kristinn/e3c450b71aa3898f39fea20abe87bade


kristinn:
Any news on this? 😸

kristinn:
I think it's best to keep the ticker in case people are running Docker versions that don't support swarm events (it was just implemented in Docker last year).

It could be disabled on run time, as in check if the Docker that is being used has support for swarm events or not. I personally think it's also good to have as a fallback in case there is an issue with the event stream (even though it should auto reconnect), since the overhead isn't that great.

vdemeester:
@kristinn right, sounds good to me 👍

provider/docker/docker.go
```diff
@@ -155,33 +155,49 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
 			}
 			if p.Watch {
-				ctx, cancel := context.WithCancel(ctx)
 				if p.SwarmMode {
+					eventsCtx := context.Background()
+					eventsMsgChan, eventsErrorChan := dockerClient.Events(
+						eventsCtx,
+						dockertypes.EventsOptions{
+							Filters: filters.NewArgs(filters.Arg("scope", "swarm")),
+						},
```
nmengin: `swarm` events can have many types (`service`, `node`,`secret`).
WDYT about adding a filter on the `type `service`?

Thus Træfik will only updates its list when the event is about an action on a service.

kristinn: Good point. 😄

nmengin:
Hello @kristinn .

Any news about the fix? :grinning:

kristinn:
Hey @nmengin .

I have allocated time for fixing this tomorrow. Sorry about the delay. :-)

nmengin:
There is no problem @kristinn.

It was just to know if you need some help.
Thanks.

kristinn:
I guess I can do something similar to https://github.com/inteleon/traefik/blob/792c97b228e827e1a116d5ed4a72d51d9f785df5/provider/docker/docker.go#L373-L374 ?

kristinn:
This is not ready for merging. I will let you know when it's ready for final review + merge. 😺

kristinn:
@nmengin This is ready for another review. I made a lot of changes (fixed some bugs I found, added some tests, improved the logic and structure, etc.).

kristinn:
Have you had time to look at the changes? 😄

kristinn:
No worries. 😄

nmengin:
Hello @kristinn ,

I was in vacation last week.
I just began the review and... you made lot of changes!!! :wink: 

I need more time to review them correctly but I give you a feedback ASAP!

kristinn:
@nmengin no worries. I figured you were on vacation. :-D

I wasn't too happy with my previous implementation so I took the liberty to refactor the code quite a bit.

nmengin:
Hello @kristinn .

I analyzed your PR and the last modifications you made seems to complicate a lot the code!
Can you explain why did you make them please?

kristinn:
Hey @nmengin 

It was hard to write tests for the original PR. I was able to catch few bugs in the original PR by rewriting it and creating a better test coverage for it.

provider/docker/docker.go
```diff
@@ -155,47 +260,103 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+
 						defer close(errChan)
-						for {
-							select {
-							case <-ticker.C:
-								services, err := listServices(ctx, dockerClient)
+
+						callback := &EventCallback{
+							ListAndUpdateServicesFunc: func() error {
+								err := p.listAndUpdateServices(watchCtx, dockerClient, configurationChan)
```
vdemeester: could be `return p.listAndUpdateServices(…)` — the log shouldn't be there (but in the code using this one). But really, I feel `EventCallback` shouldn't be there (or that way)

provider/docker/event/listener.go
```diff
@@ -0,0 +1,130 @@
...
+		StopChan:       stopChan,
+		TickerInterval: SwarmDefaultWatchTime,
+	}
+
+	return e, nil
+}
+
+// Listener is an interface for event listeners to implement.
+type Listener interface {
+	Start()
```
vdemeester: `Start` tends to apply that it's not blocking, but here, it is blocking, not sure this is the correct name.

provider/docker/event/listener.go
```diff
@@ -0,0 +1,130 @@
...
+	Stop()
```
vdemeester: This is never called from outside the package, shouldn't be defined here (should just be handled internally by the several implementations)

kristinn:
Thanks for the pointers @vdemeester. I appreciate it.

I will fix these issues. I don't have time to do it right now, but hopefully next week.

kristinn:
Hey @vdemeester.

Now I'm finally taking some time to work on this pull request! Sorry for the long delay!

The reason for why all those interfaces are exported and why the methods are provided as arguments is because I usually put the go tests inside another package, so I'm only testing the publicly facing interface. I've come to learn that it only adds complexity and annoyance.

I will change the tests so they are in the same package as the code they are testing and then it will be much easier to implement the changes that you are suggesting.

kristinn:
@vdemeester I'm thinking about scrapping the whole event package and adding this functionality just as few functions that will be part of the Docker provider. I guess it's too abstract that makes it more complex than it should.

How does that sound?

vdemeester:
@kristinn Sounds good :angel:

nmengin:
Hello @kristinn .

Whe have to freeze the code of the next version on friday.
Do you think you can modify the PR before this date?

nmengin:
Hello @kristinn .

Did you have the time to work on the PR?
The feature seems important to us and we really would like to integrate it in the next version.

That's why, if you have no time I can carry it to integrate this feature before the code freeze deadline (tomorrow).

kristinn:
Hey @nmengin .

I'm working on it as we speak. :-) I'm aware of the deadline and the importance. You can expect a code update soon.

nmengin:
Many thanks @kristinn .

Do not hesitate to ping me on the #dev channel in the slack if you need help.

kristinn:
Will do. I hope my changes will simplify things and still provide good enough test coverage.

kristinn:
@nmengin Please take a look at my newest changes and please give me some feedback on it, if I'm on the right track, etc.

The tests succeed, but I haven't tested this in my swarm cluster. I will need to do that later today or tonight since I need to go home now. I will continue to work on this later today/tonight.

kristinn:
By the way, the reason for so many new additions is because I created a mock of the Docker APIClient using mockery. I didn't check if it existed in the code base already, if it does I will use that and remove the one I committed.

kristinn:
I did two manual tests in a local Docker Swarm cluster.

I have a simple HTTP service that has two endpoints. One that responds immediately and the other one sleeps for 5 seconds before responding.

For both tests I sent 100 requests per second for 2 minutes, total 12000 requests per test. I scaled up the HTTP service, scaled it down and force updated it. All requests succeeded, got only 200 response codes.

The HTTP service needs to support graceful shutdowns and the Docker Swarm service needs to be configured with the "order" set to "start-first" (https://docs.docker.com/compose/compose-file/#update_config)

kristinn:
Hey @nmengin .

Regarding why I call `listAndUpdateServices()` twice.

We send an empty payload on every tick (when people are running old Docker versions that don't support swarm events). The payload in question is the `eventtypes.Message` parameter to the `eventCallback` function. Therefore we check if the `msg.Actor.ID` is empty, which it is if the `msg` variable is not properly populated (which is the case when the ticker executes the callback).

So, we return very early if the ticker is executing the callback since there is nothing to be done with the payload.

Regarding the retries.

When I started running my code in production I sometimes had an issue where the Docker swarm event was processed by traefik, but still the task list was empty when fetching the list of tasks. This was either because of a race condition or because of a task restarting itself.

The retry is to make sure we don't skip updating the routes to a service in case there is some kind of a race condition. This is a real possibility since it happened for me in production.

The reason for starting it up in a go routine is because I ran into a memory leak when I had a service that had problems starting up. So traefik was retrying the callback again and again but never releasing the memory it was using between retries. Starting it up in a goroutine fixes the memory leak, but makes it harder to handle error returns from the callback, unless of course adding an extra dependency like a channel.

vdemeester:
Discussing a bit offline with @nmengin, I created an upstream issue to have `task` in the swarm-scoped events in `moby`/`docker`. That way, we would not need any `time.Sleep` and it would be way simpler to implement (i.e. service creation event => we create a frontend, task creation/ready event => we add a backend to that frontend, task deletion event => we remove that backend, …).

See https://github.com/moby/moby/issues/37403

kristinn:
Ok I understand.

We've talked on Slack, but I just want to write here as well for the record.

Will people that have Swarm with support for listening to service events still have to use the ticker or will you support both listening for service events, in case the Docker version people are using doesn't support swarm task events or will there only be support for listening to swarm task events and the ticker?

nmengin:
As discussed with @kristinn , I pushed modifications to improve the concurrency management in the Docker provider in Swarm Mode.

**Note:** there are sitll flaky behaviors when we scale down from a big number of replicas to little number (from 50 to 1 for example).
This issue may be solved thanks to the Task event management :crossed_fingers:

nmengin:
Docker task event PR: https://github.com/moby/moby/pull/37441

WTFKr0:
Hi all
thanx for this PR
Are we waiting for task event PR to go further, or the current implementation can be merged quickly ?
We really need this here :smile:

nmengin:
Hello @WTFKr0 

Unfortunately we can not merge the PR in the current state because listening only the services events can introduce regression.

For the moment, we prefer wait for Docker and use the (non perfect) current solution based on the Ticker.

WTFKr0:
Thanx for reply, do you know if a PR is open to have a custom time for the ticker ?
I think it's currently a value of 15s hard coded

nmengin:
Hello @WTFKr0 ,

No, there is no PR to make this feature.

WTFKr0:
Hi

We have to scale our traefik instance
But with this 15s poll, a replica can get a config from the docker socket, and another get the config 10 seconds later, which leads to bad user experience
Is there a way to synchronize those polls between multiple replicas ?

I tought we can back the config to etcd/consul, but it seems to only store traefik flags, and not rules from docker backend

Thanx

kristinn:
> Hi
> 
> We have to scale our traefik instance
> But with this 15s poll, a replica can get a config from the docker socket, and another get the config 10 seconds later, which leads to bad user experience
> Is there a way to synchronize those polls between multiple replicas ?
> 
> I tought we can back the config to etcd/consul, but it seems to only store traefik flags, and not rules from docker backend
> 
> Thanx

@WTFKr0 : What we ended up doing was to use the Docker Swarm networking for routing.

Try enabling `traefik.backend.loadbalancer.swarm` for the Traefik container and see if it will solve your issue. This will change how traefik sees your services. Each service will be available via one overlay network IP instead of each task having it's own separate IP that Traefik routes to.

This basically means Traefik will route all traefik to each service to a single IP, which is then load balanced by the Docker Swarm mesh routing.

https://docs.traefik.io/configuration/backends/docker/#on-containers for more info.

WTFKr0:
Thanx for that @kristinn 
But I think my problem is elsewhere

My problem is a layer up, between my Load Balancer and the 3 traefik instances
I got a F5 VIP, load balancing on 3 swarm manager nodes, each running a traefik instance (exposing on 80/443) and watch local docker socket

Take the following scenario : 
- We planned to release a new app myapp.mydomain.com
- Before that, we create our DNS alias to the cluster VIP, so when we curl that, we got a 404 from traefik
- When we deploy the service (1 replica), the 3 traefik instance are going to discover it
- 1 sec after deploy, the 1st trafik instance poll the socket and discover the new service, and update rules

If now we curl our app, with a round robin load balancer in front, we got a 404 on 2 requests / 3, and the app 1/3
We could get this situation for a max of 15s, when all 3 instances have polled the docker socket and update configuration

If I can reduce this time of 15s to 3s, i think it's acceptable, but i don't know what it means about performance !

kristinn:
I understand the problem you are experiencing @WTFKr0. In fact, it sounds exactly like the problem we were experiencing.

Test running Traefik in swarm mode and also enabling the swarm load balancing as I posted earlier today. Give it a try and see if it solves the problem for you. 😸

WTFKr0:
Well, don't understand why the `traefik.backend.loadbalancer.swarm` flag can solve my problem, but i'll try soon
Thanx

nmengin:
https://github.com/moby/moby/pull/38214#issuecomment-449365822

ldez:
First, thank you for your works @kristinn .

It's been a long time since this PR is open, we have no concrete news from Docker.

So we will close this PR, if Docker advance on the subject, the PR can be reopened.

PR #3036, opened 2018-03-16
Fix goroutine leak in consulcatalog when consul is down

yutopp:
### What does this PR do?

Fix goroutine leaks in consulcatalog provider. Use `sync.Once` to avoid blocking when sending errors to the channel which already used and an owner function of it is already exited.

Below is a number of goroutines by metrics data.
```
# Before patch
> date && curl -s http://localhost:10080/metrics | grep go_goroutines             
Fri Mar 16 19:32:28 JST 2018
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 28

> sleep 120 && date && curl -s http://localhost:10080/metrics | grep go_goroutines
Fri Mar 16 19:34:29 JST 2018
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 54
```

```
# After patch
> date && curl -s http://localhost:10080/metrics | grep go_goroutines             
Fri Mar 16 18:49:38 JST 2018
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 21

> sleep 120 && date && curl -s http://localhost:10080/metrics | grep go_goroutines
Fri Mar 16 18:51:39 JST 2018
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 21
```
(traefik command is `./dist/traefik --consulcatalog --entryPoints=Name:http Address::18080 --web.address=:10080 --web.metrics --web.metrics.prometheus --web.metrics.prometheus.entrypoint=traefik`, and `consul` is not started)

### Motivation

Assume there are 2 goroutines are spawned and they have the same error channel. If one goroutine sends an error to the channel, the caller returns by getting an error from the channel. Then the other goroutine will stuck when also trying to send an error to the channel because an owner of the error channel is already exited.

### DIFF
--- a/provider/consul/consul_catalog.go
+++ b/provider/consul/consul_catalog.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"text/template"
 	"time"
 
@@ -135,7 +136,7 @@ func getChangedIntKeys(currState []int, prevState []int) ([]int, []int) {
 	return fun.Keys(addedKeys).([]int), fun.Keys(removedKeys).([]int)
 }
 
-func (p *CatalogProvider) watchHealthState(stopCh <-chan struct{}, watchCh chan<- map[string][]string, errorCh chan<- error) {
+func (p *CatalogProvider) watchHealthState(stopCh <-chan struct{}, watchCh chan<- map[string][]string, notifyError func(error)) {
 	health := p.client.Health()
 	catalog := p.client.Catalog()
 
@@ -156,7 +157,7 @@ func (p *CatalogProvider) watchHealthState(stopCh <-chan struct{}, watchCh chan<
 			healthyState, meta, err := health.State("passing", options)
 			if err != nil {
 				log.WithError(err).Error("Failed to retrieve health checks")
-				errorCh <- err
+				notifyError(err)
 				return
 			}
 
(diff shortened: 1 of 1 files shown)


provider/consulcatalog/consul_catalog.go
```diff
@@ -138,8 +139,15 @@ func (p *Provider) watch(configurationChan chan<- types.ConfigMessage, stop chan
 	watchCh := make(chan map[string][]string)
 	errorCh := make(chan error)
```
Juliens: Why not add 1 buffer on this chan instead of the sync.Once?

yutopp: Adding buffers will reduce maintainability because a number of capacity is depended on the number of gorountines which spawned at other functions. By using `sync.Once`, it will no longer need to worry about thinking the number of the buffer and blocking of the channel.

Juliens: ok for me.

We need this bug fix in 1.5, can you do this PR on branch v1.5 ?

yutopp: Changed base branch to v1.5 and patched the fix to it.

PR #3037, opened 2018-03-16
Set INFO log level in Kubernetes guide and examples.

timoreimann:
### What does this PR do?

Lower the log level for the Kubernetes manifests from DEBUG to INFO.

### Motivation

Less logging noise.

### Additional Notes

We previously enabled DEBUG log level implicitly by means of the `-d` parameter. Apart from being fairly intransparent to the regular user that debug mode hard-wires the DEBUG log level, it also produces an overwhelming amount of logs.

INFO seems much more sensible while still allowing to observe that Traefik has started, even in the absence of warnings and errors.

Finally, we also align the parameter usage across all examples we offer.

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -184,9 +184,9 @@ spec:
         securityContext:
           privileged: true
         args:
-        - -d
         - --api
         - --kubernetes
+        - --logLevel=INFO
 ---
 kind: Service
 apiVersion: v1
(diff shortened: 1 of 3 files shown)


timoreimann:
Note: I believe the `-d` switch sneaked in along #1735 rather unintentionally.

dtomcej:
Silly question, but isn't `INFO` the implicit default?

timoreimann:
@dtomcej no; `ERROR` is the default.

PR #3040, opened 2018-03-17
Added initial support for request mirroring

### LINKED ISSUE #2989: Feature Request: mirror traffic to two sets of endpoints
Please add a feature to mirror/shadow incoming traffic to two sets of endpoints and be able to configure which amount goes where...e.g., something like this:

https://dzone.com/articles/traffic-shadowing-with-istio-reducing-the-risk-of

aantono:
### What does this PR do?

This PR adds initial support for request mirroring, similar to the capabilities offered by some other proxies, like Istio (https://dzone.com/articles/traffic-shadowing-with-istio-reducing-the-risk-of)

### Motivation

Provide ability to do request shadowing, which is very useful in various testing scenarios

### More

- [x] Added/updated tests

### Additional Notes
Fixes #2989

### DIFF
--- a/middlewares/mirror.go
+++ b/middlewares/mirror.go
@@ -16,9 +16,9 @@ import (
 
 //Mirror is a middleware that provides the request mirroring capabilities
 type Mirror struct {
-	BackendURL     *url.URL
-	RequestHeaders map[string]string
-	SampleRate     uint64
+	backendURL     *url.URL
+	requestHeaders map[string]string
+	sampleRate     uint64
 	mirror         *forward.Forwarder
 	counter        uint64
 }
@@ -37,21 +37,21 @@ func NewMirrorMiddleware(frontend *types.Frontend, backend *types.Backend) (*Mir
 	}
 
 	forwarder, err := forward.New(
-	//forward.Stream(true),
-	//forward.PassHostHeader(frontend.PassHostHeader),
+		forward.Stream(true),
+		forward.PassHostHeader(frontend.PassHostHeader),
 	)
 	if err != nil {
 		return nil, err
 	}
-	return &Mirror{BackendURL: mirrorURL,
-			RequestHeaders: mirror.RequestHeaders,
-			SampleRate:     sampleRate,
+	return &Mirror{backendURL: mirrorURL,
+			requestHeaders: mirror.RequestHeaders,
+			sampleRate:     sampleRate,
 			mirror:         forwarder},
 		nil
 }
 
 func (m *Mirror) ServeHTTP(w http.ResponseWriter, req *http.Request, next http.HandlerFunc) {
-	if m.counter%m.SampleRate == 0 {
+	if m.counter%m.sampleRate == 0 {
 		atomic.AddUint64(&m.counter, 1)
 		// Copy the request and forward to 'next' and to 'mirror'
 		mReq, err := m.copyRequest(req)
@@ -71,6 +71,7 @@ func (m *Mirror) ServeHTTP(w http.ResponseWriter, req *http.Request, next http.H
 		next.ServeHTTP(w, req)
 	} else {
 		// no mirroring
+		atomic.AddUint64(&m.counter, 1)
 		next.ServeHTTP(w, req)
 	}
 }
@@ -107,7 +108,7 @@ func (m *Mirror) copyRequest(req *http.Request) (*http.Request, error) {
 		o.Body = ioutil.NopCloser(bytes.NewReader(body))
 	}
 
-	o.URL = utils.CopyURL(m.BackendURL)
+	o.URL = utils.CopyURL(m.backendURL)
 	o.Header = make(http.Header)
 	utils.CopyHeaders(o.Header, req.Header)
 
--- a/middlewares/mirror.go
+++ b/middlewares/mirror.go
@@ -1,0 +1,149 @@
+package middlewares
+
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"sync/atomic"
+
+	"github.com/containous/traefik/log"
+	"github.com/containous/traefik/safe"
+	"github.com/containous/traefik/types"
+	"github.com/vulcand/oxy/forward"
+	"github.com/vulcand/oxy/utils"
+	"net/url"
+)
+
+//Mirror is a middleware that provides the request mirroring capabilities
+type Mirror struct {
+	BackendURL     *url.URL
+	RequestHeaders map[string]string
+	SampleRate     uint64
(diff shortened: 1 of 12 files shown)


aantono:
given that the PR submission for 1.6 is March 20th and I'll be out of reach for the next week, I figured I'd get this PR in to start the review process, will make necessary fixes the week of March 26th, once I'm back to civilization. :)

middlewares/mirror.go
```diff
@@ -0,0 +1,149 @@
...
+		safe.GoWithRecover(func() {
+			mRw := createDummyResponseWriter()
+			m.mirror.ServeHTTP(mRw, mReq)
+		}, func(err interface{}) {
+			log.Errorf("Error in request mirroring routine: %s", err)
+		})
+
+		next.ServeHTTP(w, req)
+	} else {
+		// no mirroring
```
marco-jantke: We also have to increase the `m.counter` here.

middlewares/mirror.go
```diff
@@ -0,0 +1,149 @@
...
+	if mirror.SampleRate > 0 {
+		sampleRate = uint64(mirror.SampleRate)
+	}
+	mirrorURL, err := url.Parse(backend.Servers["mirror"].URL)
+	if err != nil {
+		return nil, err
+	}
+
+	forwarder, err := forward.New(
+	//forward.Stream(true),
```
marco-jantke: Did you have problems with this options?

middlewares/mirror.go
```diff
@@ -0,0 +1,149 @@
...
+		next.ServeHTTP(w, req)
+	}
+}
+
+func (m *Mirror) copyRequest(req *http.Request) (*http.Request, error) {
+	var body []byte
+	o := *req
+
+	if req.Body != nil {
+		b, err := ioutil.ReadAll(req.Body)
```
marco-jantke: I can see a couple of problems with consuming the request body here:
- won't this "drain" the request body and therefore make it appear as empty or at the end for other consumers of the request body like the original request forwarding?
- this is still in the synchronous path of the original request execution and therefore it would block until a request is forwarded until the whole request body is consumed. This can delay proxying significantly for slow clients, like mobile phones with low bandwidth.

middlewares/mirror.go
```diff
@@ -0,0 +1,149 @@
...
+		})
+
+		next.ServeHTTP(w, req)
+	} else {
+		// no mirroring
+		next.ServeHTTP(w, req)
+	}
+}
+
+func (m *Mirror) copyRequest(req *http.Request) (*http.Request, error) {
```
marco-jantke: I checked how the forwarding middleware does copying and it is only using:

```
outReq := new(http.Request)
*outReq = *inReq // includes shallow copies of maps, but we handle this in Director
```

Would this also be enough in our case here?

middlewares/mirror.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/containous/traefik/safe"
+	"github.com/containous/traefik/types"
+	"github.com/vulcand/oxy/forward"
+	"github.com/vulcand/oxy/utils"
+	"net/url"
+)
+
+//Mirror is a middleware that provides the request mirroring capabilities
+type Mirror struct {
+	BackendURL     *url.URL
```
marco-jantke: Can we make those fields also private? I think it's best practice to only export when it's really required.

Juliens:
Thank you for this PR!

For now, your PR only covers the file provider.
As you may know, in 1.6, @ldez made a big effort to standardize all the providers. 
That's why we really need an implementation of each new features on all the providers.

Do you think you can handle your feature with labels on docker, annotations on kubernetes, etc...?
Be careful, @ldez is still modifying those part, don't forget to synchronize your changes with him.

aantono:
@Juliens, that’s the plan.  That’s why I made this as the initial implementation for the feature.  Once we agree on the functionality it was my plan to add support for tags/labels/annotations etc to all the providers that support such configuration options.  I plan on getting those done next week, once I’m back in town. (assuming this gets reviewed and approved).

middlewares/mirror.go
aantono: This shouldn't drain the request body as the `Body` is being read into a `body` byte array, which later gets assigned to the `Body` of the original request and the clone individually.  So by the time `next.ServeHTTP` is being called, each request has a proper body.

The choice of doing request mirroring assumes that the developers are willing to "pay" for the capability with the latency introduced by the need to buffer up (read in) the request data in order to be able to fork it to both destinations.  Do you have any proposed techniques that might improve this?

middlewares/mirror.go
aantono: The forwarding middleware does not need to actually preserve teh `inReq` object, so they don't care about things like `Body`, etc.  Also there is additional code in `modifyRequest` which does some URL massaging magic.  The `copyRequest` part was borrowed from the `oxy/buffer` package, which seems to better match the current needs, IMHO.

middlewares/mirror.go
aantono: 👍

aantono:
Added backend code to the supported providers for labels, will submit the template changes in a separate commit.

middlewares/mirror.go
aantono: Done

emilevauge:
@aantono thanks for your contribution (one more :) )
We need discuss a bit more on this one with the team as it might be a good idea to generalize this feature to cover what's done in https://github.com/containous/traefik/pull/3072.
We will get back to you in the coming days.

emilevauge:
Hey @aantono, we discussed multiple times internally on this feature and we ended up in a similar solution, except that we think it would be a lot better to attach the mirroring option to the backend instead of the frontend. It makes a bit more sense.
Beyond that, this feature requires quite a refactor of the glorious `Server.loadConfig()` function. We need to separate backends and frontends creation. We plan to do this refactor in the next 1.7 release (probably done by @ldez and/or @Juliens), this implies you will need to wait a bit before continuing.
Sorry about this setback, we love this idea, but we prefer not to rush on key feature ;)
I propose to close the PR for now, to avoid too much divergence, and to reopen on new one once the refactor is done. WDYT @aantono ?

aantono:
@emilevauge Thanks a lot for the update.  Glad to hear you like the feature and it sounds like you've arrived to a similar conclusion as myself, matching what we've discussed in Slack about having separate Frontend and Backend creation.  Definitely let's close this PR and get the refactoring in place first.  Please let me know if I can help in any way in the refactoring efforts!

yangm97:
Is this refactoring being tracked somewhere?

PR #3041, opened 2018-03-17
Miss-leading Docker backend documentation

Rakhmanov:
1. Two Different ways of running inside of Docker-Compose syntax
Problem encountered, when switched to "new" syntax, none of the options were applied, when using docker-compose up, docker-compose up --scale... etc.. Since docker giving warning that deploy labels will be ignored.

2. External network problem:
When stack has multiple networks setting label to the external name only works. All attempts to link to current stack's networks and have services accessible failed. Since it selects the random network. 

There is a need to further clarify and explain two different ways of running traefik with `docker bundle`, and just regular `docker-compose` with service definition.  It affects how docker networks are being named and how compose file are read.

The issue is silent for networks until there is scenario where multiple are used.

 ```
Docker Image: Traefik 1.5.4
Version: 17.12.0-ce-win47 (15139)
Channel: stable
```

### What does this PR do?

Documentation change for Docker backend

### Motivation

Ran in to the issue when deploying.

### More

### CODE REVIEW

**Comment by revolunet** [comment]
is using `traefik.docker.network` the same as using networks ?

```
services:
  whoami:
    networks:
      - web
networks:
  web:
    external:
      name: traefik_webgateway
```

mmatur:
@revolunet You can use `traefik.docker.network=web` or `traefik.docker.network=traefik_webgateway` both work

Issue #3020, opened 2018-03-15
Keep AccessLog entries based on retry attempts


marco-jantke:
### Do you want to request a *feature* or report a *bug*?

*feature*

### Description

We want to keep all access logs that either responded with a 5xx or where at least one retry attempt happened. While the former is perfectly possible now with the access log filters for status codes, we can't achieve the second, yet.

marco-jantke:
This is implemented and available in v1.6. Closing the issue :tada:

PR #3042, opened 2018-03-18
add access log filter for retry attempts

marco-jantke:
This PR implements https://github.com/containous/traefik/issues/3020.

### DIFF
--- a/docs/configuration/logs.md
+++ b/docs/configuration/logs.md
@@ -17,6 +17,7 @@ logLevel = "INFO"
 
   [accessLog.filters]
     statusCodes = ["200", "300-302"]
+    retryAttempts = true
 
   [accessLog.fields]
     defaultMode = "keep"
@@ -44,6 +45,7 @@ For more information about the CLI, see the documentation about [Traefik command
 --accessLog.filePath="/path/to/access.log"
 --accessLog.format="json"
 --accessLog.filters.statusCodes="200,300-302"
+--accessLog.filters.retryAttempts="true"
 --accessLog.fields.defaultMode="keep"
 --accessLog.fields.names="Username=drop Hostname=drop"
 --accessLog.fields.headers.defaultMode="keep"
@@ -122,7 +124,7 @@ filePath = "/path/to/access.log"
 format = "json"
 ```
 
-To filter logs by status code:
+To filter logs:
 ```toml
 [accessLog]
 filePath = "/path/to/access.log"
@@ -130,12 +132,19 @@ format = "json"
 
   [accessLog.filters]
 
-  # statusCodes keep only access logs with status codes in the specified range
+  # statusCodes keep access logs with status codes in the specified range
   #
   # Optional
   # Default: []
   #
   statusCodes = ["200", "300-302"]
+
+  # retryAttempts keep access logs when at least one retry happened
+  #
+  # Optional
+  # Default: false
+  #
+  retryAttempts = true
 ```
 
 To customize logs format:
--- a/docs/configuration/logs.md
+++ b/docs/configuration/logs.md
@@ -124,7 +124,7 @@ filePath = "/path/to/access.log"
 format = "json"
 ```
 
-To filter logs:
+To filter logs you can specify a set of filters which are logically "OR-connected". Thus, specifying multiple filters will keep more access logs than specifying only one:
 ```toml
 [accessLog]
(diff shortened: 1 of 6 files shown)


middlewares/accesslog/logger_test.go
```diff
@@ -357,6 +357,17 @@ func TestNewLogHandlerOutputStdout(t *testing.T) {
 			},
 			expectedLog: `TestHost - TestUser [13/Apr/2016:07:14:19 -0700] "POST testpath HTTP/0.0" 123 12 "testReferer" "testUserAgent" 23 "testFrontend" "http://127.0.0.1/testBackend" 1ms`,
 		},
+		{
```
mmatur: If I add this test case 

```go
{
	desc: "Default config",
	config: &types.AccessLog{
		FilePath: "",
		Format:   CommonFormat,
		Filters:  &types.AccessLogFilters{},
	},
	expectedLog: `TestHost - TestUser [13/Apr/2016:07:14:19 -0700] "POST testpath HTTP/0.0" 123 12 "testReferer" "testUserAgent" 23 "testFrontend" "http://127.0.0.1/testBackend" 1ms`,
},
```

The default response has changed

types/types_test.go
```diff
@@ -93,3 +94,43 @@ func TestNewHTTPCodeRanges(t *testing.T) {
...
+			statusCode: 400,
+			contains:   false,
+		},
+	}
+
+	for _, test := range testCases {
+		test := test
+		testName := fmt.Sprintf("%q contains %d", test.strBlocks, test.statusCode)
+		t.Run(testName, func(t *testing.T) {
+			httpCodeRanges, err := NewHTTPCodeRanges(test.strBlocks)
```
mmatur: Could you please run these test cases in parallel

middlewares/accesslog/logger_test.go
marco-jantke: That is right. Now when you specify `Filters` but nothing inside it will remove everything. Isn't that the kind of logical behavior? The filters are OR-connected and when you specify multiple (status codes and retry attempts) than the result is more than when you specify only one of those. WDYT?

types/types_test.go
marco-jantke: Good point, fixed.

middlewares/accesslog/logger_test.go
mmatur: Thanks for your response @marco-jantke 
When we specify nothing in `Filters`, we need to keep all access logs. And when something is specified in `Filters` we need to keep only access logs matching at least one filter

marco-jantke: As said before I think that's not logical. And "optimizing" for a use-case where someone would specify an empty filters config doesn't make too much sense. However, I implemented it now as you requested in https://github.com/containous/traefik/pull/3042/commits/6ca142b96ee9a998e5a261e23e97a034e1c0209b. PTAL and let me know what you think. I didn't see a nicer option for detecting an empty configuration except validating each config value one by one to check whether it's the zero value. Note that with this approach we have to extend this method each time a new option gets added. I'm open to improvement ideas.

mmatur: Thanks for your correction @marco-jantke 

I am pretty agree with you but I think that it is not the aim of this PR to change the default behavior. To me you have choose the best option to detect an empty configuration.

We should maybe do an other PR to improve this method to avoid to extend it every time we add a new option

docs/configuration/logs.md
```diff
@@ -122,20 +124,27 @@ filePath = "/path/to/access.log"
...
 
-  # statusCodes keep only access logs with status codes in the specified range
+  # statusCodes keep access logs with status codes in the specified range
   #
   # Optional
   # Default: []
   #
   statusCodes = ["200", "300-302"]
+
+  # retryAttempts keep access logs when at least one retry happened
```
nmengin: Can you add a specify that this option will keep `keep access logs when at least one retry happened` **even if the status code is not in the `statusCodes` ranges**.
Thus users can understand that filters are _OR-connected_ :wink:

marco-jantke: @nmengin I added now a sentence at the beginning of the segment, so that we don't have to repeat that sentence on every filter we have :D 

Can you have a look and let me know whether it would have helped you?

PR #3047, opened 2018-03-19
add http headers to healthcheck

### LINKED ISSUE #3043: Add support to additional httpheaders in healthchecks

feature

### What did you do?

I am trying to create healthcheck which contains custom httpheaders. If I understand correctly traefik does not support Hostname in backend healthcheck?

I except that I should have possibility to configure hostname and other additional httpheaders in healthchecks. In my opinion in kubernetes have pretty good custom httpheaders example https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-http-request If traefik could have something like similar it would be awesome. Then we could configure hostname(and others custom headers) in healthchecks 

### What did you see instead?

I do not see possibility to configure these headers

```
(paste your output here)
```

```toml
```

```
(paste your output here)
```

zetaab:
### What does this PR do?

Tries to fix issue #3043 

### Motivation

We need this feature, describe in issue

### More

- [x] Added/updated tests

### Additional Notes

Fixes  #3043


zetaab:
This is still WIP

I am adding healthchecks 
```
        "healthcheck": {
            "path": "/",
            "interval": "5s",
            "headers": {
            	"Host": "foobar.com",
            	"header1": "test",
            	"headeR2": "test2"
            }
        },
```

I can see in logs DEBU[2018-03-19T14:32:53+02:00] Setting up backend health check [Headers: map[Host:foobar.com header1:test headeR2:test2] Path: / Port: 0 Interval: 5s]

However, still I am not able to see Host header in client.

```
ERROR:root:Host: 192.168.1.43:8000
User-Agent: Go-http-client/1.1
Header1: test
Header2: test2
```

I am thinking might this cause this issue still https://github.com/golang/go/issues/7682

zetaab:
```

**Comment by zetaab** [comment]
Got it working, its little bit tricky in Host header. Is these tests and docs enough or do we need more?

**Comment by Juliens** [comment]
Thanks for your contribution!

I understand your use case with the `hostname`, but I'm not sure about changing `headers`.

Do you think you can just add "hostname option" and remove other headers option?

**Comment by zetaab** [comment]
@Juliens IMO headers are useful for instance if the healthcheck needs authentication. So I do not see why we could not add those in same PR

**Comment by Juliens** [comment]
@zetaab After discussion with the team, we agree to add both `hostname` and `headers` on healthcheck.

But we think we need to separate hostname from the headers.
If I take your proposal, it becomes like this:
```
        "healthcheck": {
            "path": "/",
            "interval": "5s",
            "hostname": "foobar.com",
            "headers": {
            	"header1": "test",
            	"headeR2": "test2"
            }
        },
```

Moreover, we need the functionnality to be added on every providers.

If you need help, you can share with us on our [Slack workspace](https://traefik.herokuapp.com/)

ldez:
@zetaab Could you also update the doc for each provider, thanks.

PR #3050, opened 2018-03-19
Allow overriding the log level in debug mode.

### LINKED ISSUE #2968: Support overriding the log level in debug mode

Feature.

### What did you do?

Start Traefik with the switches `--debug` and `--logLevel=WARN`.

Expect Traefik to run at the WARN log level.

### What did you see instead?

Traefik is running at the DEBUG level.

master

### Additional info

Debug mode does a few extra things besides setting the log level to DEBUG. Namely, it also enables expvars and sets up pprof handlers which can both be helpful in identifying performance issues. However, a log level of DEBUG often makes it impossible to isolate performance problems in high-volume scenarios. For instance, vulcand/oxy produces (simplistic) access logs for each request on that log level, which provable hampers benchmarking.

When the user has enabled both the debug mode *and* provided an explicit log level (other than DEBUG), that specified log level should take precedence over the default DEBUG one.

timoreimann:
### What does this PR do?

Allow overriding the log level in debug mode.

### Motivation

Enabling the debug mode caused the log level to be hard-wired to DEBUG. We change the behavior so that an explicitly defined log level always overrides.

One primary motivation is to allow pprof-based benchmarking without excessive DEBUG logging distorting the results.

### Additional Notes

We also remove a piece of code from `SetEffectiveConfiguration` which was never effective because the log level would already be set previously.

Fixes #2968.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -62,7 +62,7 @@ Add more configuration information here.
 -->
 
 
-### If applicable, please paste the log output in debug mode (`--debug` switch)
+### If applicable, please paste the log output at DEBUG level (`--logLevel=DEBUG` switch)
 
 ```
 (paste your output here)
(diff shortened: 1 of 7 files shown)


timoreimann:
@nmengin good catch -- updated.

PR #3072, opened 2018-03-23
Adding priority to backend server properties

raehalme:
### What does this PR do?

The PR adds a new field `priority` to backend servers. Only servers with the highest priority are considered when selecting a server while all servers are available for serving existing users. Weight can still be used to balance requests across servers (of the same priority). 

### Motivation

Iimplementing zero downtime upgrades on applications making use of HTTP sessions is a bit more complex than simply replacing the servers with the new version. Replicating sessions across servers is not necessarily an option (due to, for example, cross-version incompatibility or performance) and sticky sessions are often used to avoid the need for replication.

Using weight has been offered as a solution to implement session draining. The problem with weight, however, is that you need to update existing configuration. With providers such as AWS ECS this may be impossible without recreating the containers (which defeats the purpose).

This PR makes it possible to properly implement session draining: existing users keep using the current version while new users are directed to the latest version. For example, assume you have version 1.0 deployed with backend servers having priority 1. When version 2.0 is deployed, priority 2 can be used on the new backend servers. If the version 1.0 still has active users, they may keep using the old servers (with priority 1). New users, however, are sent to the 2.0 version due to its higher priority on the new backend servers.

The priority is an integer and can be anything from build number to timestamp.  

### More

- [x] Docker
- [ ]  Rancher
- [ ]  Marathon
- [ ]  k8s
- [x]  Mesos
- [x]  ECS
- [ ]  Consul Catalog
- [ ]  KV (bold, consul, etcd, zk)
- [ ] Updated documentation

### DIFF
--- a/autogen/gentemplates/gen.go
+++ b/autogen/gentemplates/gen.go
@@ -504,6 +504,7 @@ var _templatesDockerTmpl = []byte(`{{$backendServers := .Servers}}
   {{range $serverName, $server := getServers $servers }}
   [backends."backend-{{ $backendName }}".servers."{{ $serverName }}"]
     url = "{{ $server.URL }}"
+    priority = {{ $server.Priority }}
     weight = {{ $server.Weight }}
   {{end}}
 
--- a/autogen/gentemplates/gen.go
+++ b/autogen/gentemplates/gen.go
@@ -1505,6 +1505,7 @@ var _templatesMesosTmpl = []byte(`[backends]
   {{range $serverName, $server := getServers $tasks }}
   [backends."backend-{{ $backendName }}".servers."{{ $serverName }}"]
     url = "{{ $server.URL }}"
+    priority = {{ $server.Priority }}
     weight = {{ $server.Weight }}
   {{end}}
 {{end}}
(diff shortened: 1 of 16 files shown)


Toub:
Great. If accepted, the documentation should also be updated.

emilevauge:
@raehalme thanks for your contribution !
We need discuss a bit more on this one with the team as it might be a good idea to generalize this feature to cover what's done in https://github.com/containous/traefik/pull/3040.
We will get back to you in the coming days.

raehalme:
No prob! I'm happy to work with this further.
I'm not really sure, however, if I understand how this relates to #3040? Do you mean that the prioritization should be implemented with some middleware component?

Juliens:
Thanks for your contribution!

After some discussion, we agree that we can accept this functionality, but we think the option should be called `version` instead of `priority`.

Moreover, I’m not sure that we must/can modify the oxy implementation, I think you can handle it directly in the `loadConfig` in `server.go`.

raehalme:
Yes, I'm happy to make the changes. Just haven't had the time to work on this for a while.

emilevauge:
@raehalme Sorry to ask ping you again, any chance to work on this soon?
Don't want to rush ;)

emilevauge:
@raehalme Thanks a lot for starting this contribution :)
I think we may get back to it in the future but for now, due to inactivity, I prefer to close it.
Don't hesitate to reopen a new one!
Thanks again

PR #3080, opened 2018-03-26
Add wildcard match to acme domains

oldmantaiter:
### What does this PR do?

- Allows matching of hosts like `x.traefik.wtf` to ACME wildcard
  certs `*.traefik.wtf`

### Motivation

- Tested wildcard acme v2 and noticed that it was not properly selecting my wildcard domains

### More

- [x] Added/updated tests

### DIFF
--- a/acme/account.go
+++ b/acme/account.go
@@ -8,6 +8,7 @@ import (
 	"crypto/x509"
 	"fmt"
 	"reflect"
+	"regexp"
 	"sort"
 	"strings"
 	"sync"
@@ -219,6 +220,12 @@ func (dc *DomainsCertificates) getCertificateForDomain(domainToFind string) (*Do
 
 	for _, domainsCertificate := range dc.Certs {
 		for _, domain := range domainsCertificate.Domains.ToStrArray() {
+			if strings.HasPrefix(domain, "*.") {
+				selector := "^" + strings.Replace(domain, "*.", "[^\\.]*\\.", -1) + "$"
+				if domainCheck, _ := regexp.MatchString(selector, domainToFind); domainCheck {
+					return domainsCertificate, true
+				}
+			}
 			if domain == domainToFind {
 				return domainsCertificate, true
 			}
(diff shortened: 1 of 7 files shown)


emilevauge:
@oldmantaiter could you explain more in details why you added those lines in `getCertificateForDomain()` ?
```
			if strings.HasPrefix(domain, "*.") {
				selector := "^" + strings.Replace(domain, "*.", "[^\\.]*\\.", -1) + "$"
				if domainCheck, _ := regexp.MatchString(selector, domainToFind); domainCheck {
					return domainsCertificate, true
				}
			}
```

oldmantaiter:
@emilevauge for sure.

I was testing the new ACMEv2 merge and created a certificate for (as an example) `*.foo.bar` and have backends that route based on `Host:x.foo.bar`  or `Host:y.foo.bar`. These were not properly getting the stored ACME wildcard cert because this function only looked for and exact string match

dtomcej:
@emilevauge This code snippet has been repeated in the acme package 3x already, and in `server.go`, do you think that we should refactor this to be a more unified function, to check for matches better? rather than re-duplicating code?

oldmantaiter:
@dtomcej I noticed the same, I can combine if that is easier

nmengin:
I am agree with @dtomcej .

I guess we can transform the list of `DomainCertificate` into a `map[string]*tls.Certificate` instead of a `[]string` in the [loop instruction](https://github.com/containous/traefik/pull/3080/files#diff-1cbe282e78372e8b10a131fbec146506R222) (by using [the metod Domain.ToStrArray](searchProvidedCertificateForDomains) as keys and [the DomainCertificate.Certificate](https://github.com/containous/traefik/blob/master/acme/account.go#L265) as values).

Then, we can use [acme.searchProvidedCertificateForDomains method](https://github.com/containous/traefik/pull/3080/files#diff-1cbe282e78372e8b10a131fbec146506R222) instead of copy/paste the regexp matching.

@oldmantaiter Do you think you can do it? Or do you need some help?

nmengin:
@oldmantaiter 

Many thanks for your PR.
Because this fix is very important, I did the _pattern matching_ refacto in the way to generate quickly a new Release Candidate.

oldmantaiter:
@nmengin Thanks. There is one more place I saw this pattern matching code (`provider/acme/provider.go`) so I was going to put the function inside `provider/acme` and reference it from the `acme` package.

nmengin:
@oldmantaiter I saw it too and I am currently analyzing a solution to integrate all of them.

I think it will be in an other PR.

Many thanks for your feedback :+1:

PR #3093, opened 2018-03-28
Update kubernetes.md

rdrgporto:
### What does this PR do?

In the section **Add a TLS Certificate to the Ingress** ( [Link](https://github.com/containous/traefik/blob/master/docs/user-guide/kubernetes.md) ) :

```yml
kind: Ingress
metadata:
  name: traefik-web-ui
  namespace: kube-system
  annotations:
spec:
  rules:
  - host: traefik-ui.minikube
    http:
      paths:
      - backend:
          serviceName: traefik-web-ui
          servicePort: 80
  tls:
    secretName: traefik-ui-tls-cert
```

If we apply this manifest, we receive an error:

```shell
error: error validating "traefik-ingress.yaml": error validating data: expected type array, for field spec.tls, got map; if you choose to ignore these errors, turn validation off with --validate=false
```

In order to solve the issue, it is necessary to add a **hyphen**:

```yml
 - secretName: traefik-ui-tls-cert
```

### More

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -371,7 +371,7 @@ spec:
           serviceName: traefik-web-ui
           servicePort: 80
   tls:
-    secretName: traefik-ui-tls-cert
+   - secretName: traefik-ui-tls-cert
 ```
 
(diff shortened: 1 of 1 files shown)


rdrgporto:
You are welcome :+1:

Issue #2785, opened 2018-01-30
Homogenization of the providers


ldez:
### Do you want to request a *feature* or report a *bug*?

feature

### What did you do?

Have the same possibility for each provider.

### Details

1. [x] split build configuration and provide
   - Docker, Eureka, k8s, Marathon, Rancher, Mesos, ECS: #2518
   - KV (BoldDB, Consul, Etcd, ZK): #2616
1. [x] have all options available for all providers
   - k8s: #2612
   - KV: #2652
   - Consul Catalog: #2646
   - Mesos: #2687
   - Docker: #2584, #2604
   - ECS: #2605
   - Marathon: #2602
   - Rancher: #2601
1. [x] have "shared" template fragments (Homogenization of templates)
   - Rancher: #2687
   - Consul Catalog: #2668
   - Marathon: #2665
   - ECS: #2663
   - Rancher: #2662
   - KV: #2661
   - Docker: #2659
1. [x] have a common way to manage "service" labels. (rename "service" to "segment")
   - Docker: #3055
   - Marathon: #3073
   - Rancher: #3073
   - ECS #3817
   - Mesos  #3383
1. [x] have a shared way to manage labels.
   - Docker: #3055
   - Marathon: #3073
   - Rancher: #3073
   - Consul Catalog: #3099
   - ECS: #3099, #3159 
   - Mesos: #3099
 
1. [ ] add generic methods on all providers to allow easy customization.
1. [ ] add documentation about template customization. (issue #1590)
1. [ ] _under consideration_: remove template (keep the use of template only for custom needs)

| Provider       | Labels | Segment | Annotations |
|----------------|--------|---------|-------------|
| Consul Catalog | x      | -       | -           |
| Docker         | x      | x       | -           |
| ECS            | x      | x       | -           |
| k8s            | -      | -       | x           |
| Marathon       | x      | x       | -           |
| Mesos          | x      | x       | -           |
| Rancher        | x      | x       | -           |
| Eureka         | ~      | -       | -           |
| DynamoDB       | -      | -       | -           |

ldez:
@tcolgate we will not remove template (see the 2 last points)

PR #3099, opened 2018-03-28
Factorize labels managements.

ldez:
### What does this PR do?

Factorize labels managements for:
- ECS
- Consul Catalog
- Mesos

### Motivation

Homogenization of the providers [part5]: have a shared way to manage labels.

### More

- [x] Added/updated tests

### Additional Notes

Related to #2785


provider/mesos/deprecated_config_test.go
```diff
@@ -0,0 +1,431 @@
...
+				},
+			},
+		},
+	}
+
+	for _, test := range testCases {
+		test := test
+		t.Run(test.desc, func(t *testing.T) {
+
+			actualConfig := p.buildConfigurationV1(test.tasks)
```
mmatur: Is there a reason do don't run these test cases in parallel ?

PR #3100, opened 2018-03-28
Improved cluster api to include the current leader node

aantono:
Added information to display the current cluster leader node to the API

### DIFF
--- a/cluster/leadership.go
+++ b/cluster/leadership.go
@@ -14,10 +14,11 @@ import (
 	"github.com/unrolled/render"
 )
 
+const clusterLeaderKeySuffix = "/leader"
+
 var templatesRenderer = render.New(render.Options{
 	Directory: "nowhere",
 })
-var clusterLeaderKeySuffix = "/leader"
 
 // Leadership allows leadership election using a KV store
 type Leadership struct {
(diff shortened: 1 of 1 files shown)


emilevauge:
@aantono hmmm not sure not having a GUID is a good thing.
How about adding a `Hostname` field that would be always set ?

aantono:
@emilevauge Not having a GUID is already possible today, as one can use the `--cluster.node=anything` as a config, and in that case the `Node` identifier will not be a GUID, but whatever the supplied value is `anything`.  So this is just a convenience instead of having to externally derive the hostname and pass it as a value.  But I can do that today already via doing `--cluster.node=${HOSTNAME}` and it works. :)

emilevauge:
I got it, but we prefer to set this value in a specific `cluster.hostname` field then, and not reuse `cluster.name` :) Ok for you ?

aantono:
@emilevauge As per our discussion in Slack, I've removed the part about `UseHostname`, so the `cluster.node` will have to be set externally or via config file if desired.

mmatur:
Hi @aantono, could you please explain us why you have rebase your PR on branch v1.6 ?

aantono:
That was done per @ldez request to be merged into 1.6 release.  Is that no longer needed?

ldez:
@aantono Are you sure? where do you ask for this? Are you sure you do not confuse with another PR?

PR #3101, opened 2018-03-29
Moved /api/cluster/leadership handler under public routes (requires no authentication)

aantono:
As it was brought up in Slack, the `/cluster/leadership` URL is mostly being called from various load balancers and other sources that don't typically do authentication.  Having `/cluster` handler be under API endpoint, which otherwise might need to be guarded with authentication, prevents access to `/cluster/leadership` URL from load balancers.  Moving it under `Ping` endpoint, will solve this problem, and, as it seems from the intent of `/cluster/leadership` URL, it actually belongs under `Ping` to begin with, as it serves similar purpose.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -744,15 +744,16 @@ func (s *Server) addInternalRoutes(entryPointName string, router *mux.Router) {
 
 	if s.globalConfiguration.API != nil && s.globalConfiguration.API.EntryPoint == entryPointName {
 		s.globalConfiguration.API.AddRoutes(router)
-		if s.leadership != nil {
-			s.leadership.AddRoutes(router)
-		}
 	}
 }
 
 func (s *Server) addInternalPublicRoutes(entryPointName string, router *mux.Router) {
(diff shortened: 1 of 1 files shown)


CumpsD:
Thanks @aantono!

Juliens:
Thanks for your contribution!

WDYT about keeping `/api/cluster/leader`, enabling it with `api` option and just disabling auth on this endpoint ?

CumpsD:
that would serve the use case too

aantono:
@Juliens do you mean to disable auth on just `/api/cluster/leader` without disabling auth on the rest of the `/api/*` URLs?

Juliens:
@aantono yes, we need to keep auth on /api/* but no auth on /api/cluster/leader

As this endpoint was introduce in 1.6.0, can you put it in 1.6.0 to fix the auth bug ?

aantono:
You bet.  Can you point me to the right place to tweak that config?  Looked for it myself, but can’t find the right nobs to turn ;)

Juliens:
You need two things:
- rebase your branch on top of v1.6 `git rebase --onto upstream/v1.6 HEAD^`
- Change the base branch ( `edit` next to the PR Title and then you can change it)

PR #3102, opened 2018-03-29
Fixed documentation urls on README.md

emir:
I am not following existing PR template because it's just very little update for fixing broken urls on README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,18 +70,18 @@ _(But if you'd rather configure some of your routes manually, Træfik supports t
 
 ## Supported Backends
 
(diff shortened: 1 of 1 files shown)


geraldcroes:
Thx for spotting this one.

Actually, we think that pointing to the documentation is the solution.

Could you please update your PR and point to URLs from the documentation, e.g.

`https://docs.traefik.io/configuration/backends/docker/`

instead of 

`docs/configuration/backends/docker.md`

Thanks again for your help.

emir:
Hello Gérald, I've updated urls👌

geraldcroes:
Thank you :smile:

PR #3112, opened 2018-04-01
feat(kubernetes): specify backend servers' weight via annotation

### LINKED ISSUE #2729: Support specifying weights on Kubernetes

Feature request.

### What did you do?

Try to assign different weights to pods/backends. One use case is the ability to do canary releases where pods associated with a new deployment get assigned increasingly higher fractions of traffic over time.

An annotation allowing to tweak weights.

### What did you see instead?

No annotation available. 😢 

### Additional information

This feature has been requested a number of times on the Slack channel. It seems like a good addition to Traefik.

yue9944882:
### What does this PR do?

Fixes #2729. Also previous discussions.

Provides a new ingress annotation `ingress.kubernetes.io/backend-weights` of which the value is a `map[string]int`-typed json object. With this annotation, we can do canary release by dynamically adjust the weight of ingress backends.

Since that currently the weight of `types.Server` is integer, so I created a simple allocator to make the weight of the server as average as possible.

### More

- [X] Added/updated tests


yue9944882:
@timoreimann In previous discussion, you proposed a `percentage-based` value. Do you mean a string in the form like "12%" or a integer value ranged from 0 to 100? As for the latter one, do we need to constraint that the weights under the same `ingress.Spec.Rule.Path` should sum up to 100 to make  sure it is "percentaged"?

btw what if we want a smaller granularity like 1/1000-ish?

trondhindenes:
@yue9944882 I think it should match Traefik's "regular" weight logic, where weight is just an arbitrary number and doesn't have to sum up to 100 in order to be valid.

yue9944882:
@trondhindenes +1 with that. But the problem is how to spread the weight without breaking the proportion of the weights from annotation when they are integer.

My currently logic tries to calculate a remainder and re-assign it to the servers by `1` each time. Otherwise we can scale the integer 1000 times aka taking the weight as milli-value which is similar to  what `resource.Quantity` does in kubernetes to help to do division on calculation of `total weight / instance number`.

provider/kubernetes/weight.go
```diff
@@ -0,0 +1,51 @@
...
+func (a *defaultWeightAllocator) GetWeight() int {
+	return defaultServerWeight
+}
+
+// weightPoolAllocator performs a integer-based instance weight allocation.
+// It takes *totalWeight* and *allocationCount* as input and try to allocate the
+// total weight as average as possible by integer.
+type weightPoolAllocator struct {
+	flooredWeight  int
+	weightRemainer int
```
yue9944882: typo. will fix it after @timoreimann reviews.

timoreimann:
@yue9944882 I suppose the answer to your question depends on what we consider to be the most useful UI to specify weights. My personal opinion is that I'd rarely want to specify them literally -- I'd rather prefer to note down percentage values and rely on the implementation to translate those into integer-based, relative weights while taking the number of endpoints into account accordingly. While I could do the math myself, it's not trivial and seems to require knowledge about how many endpoints I have.

Let's look at a service that fronts 20 endpoints. If I assign a weight of 10 to that service, it means that only half of those endpoints will actually receive traffic, right? That'd be quite wasteful. With only literal weights to specify, I'd have to know the weight-to-endpoints relationship a priori and scale my weights accordingly so that all endpoints receive some traffic, all while aiming for evenly distributed traffic overall. IMHO, the algorithm should take care of this which it should if we just specify percentages.

Does that make sense?

yue9944882:
@timoreimann PTAL again
> To clarify my argument, let's look at a service that fronts 20 endpoints. If I assign a weight of 10 to that service, it means that only half of those endpoints will actually receive traffic, right?
I just changed the implementation. Now we use regard all weight as a milli value at the first place e.g. 10 -> 10  * 1000, then we can do the division with maximum loss of 10^-3 precision. WDYT?

timoreimann:
@yue9944882 thanks, I think that should address the scaling problem.

Before discussing about implementation details, I'd like to raise the question again if pure integer weights represent the right interface as opposed to percentages. My guess is that most users would think in terms of percentage values anyway and translate them into integer values because Traefik would require them at the current stage. If that's the case, shouldn't we decide to accept percentages exclusively? Or is there a legitimate use case to specify integers?

timoreimann:
@trondhindenes appreciate any feedback you may have on the question I raised above.

yue9944882:
@timoreimann Umm..🧐I'd +1 if we could use percentage weight. This may help if other providers is doing similar weight-specifying things. And my little concern is that whether percentage weight would be cause some overhead or potential "float-precision-overflowing" problem.

timoreimann:
Had a chat with @marco-jantke, he came up with another useful feature for the percentage interface: It'd be convenient if you could omit the specification for the last fraction and let Traefik compute it. For instance, if I were to specify 5% on one of two services, then Traefik should conclude that 95% need to go the other service.

This would allow users to promote canary release stages easily just by updating the percentage of the new deployment/service gradually (e.g., 1%, 5%, 25%, 50%, 100%), as opposed to having to update both deployments/services (e.g., 1%/99%, 5%/95%, etc.).

trondhindenes:
@timoreimann that makes sense. Still, I guess I'd prefer consistency with how Traefik works "outside" of Kubernetes, where the weight is just an arbitrary number (30 + 30 + 100 is completely okay). 

That said: I'm fine either way - as long as things are documented I don't think it matters too much. I totally see the point in allowing auto-calculation of the "remainder weight" if using percentages - that's functionality that (afaik) doesn't exist in Traefik today.

yue9944882:
> It'd be convenient if you could omit the specification for the last fraction and let Traefik compute it.

@timoreimann Agree that it would be more convenient. But does it mean that we could only specify two service for each ingress rule in  this feature?  IDK if there are use-cases where we need to specify 3+. E.g. we may do "canary analysis" in which we need to split the traffic 98%(main traffic), 1 %(canary traffic), 1%(baseline traffic)  and do comparative analysis over them.

timoreimann:
@yue9944882 the idea is that you can only omit the _last_ fraction. So in your example, you'd likely want to specify 1% for the canary, 1% for the baseline, and omit the 98% for the main traffic as a convenience. If you had two canaries each with baselines running plus the main traffic, you could specify 4 out of 5 fractions, etc.

Does that make sense?

yue9944882:
> @containous/traefik any thought on whether we should expose an integer-based interface or not? Is there a valid use case for it?

@timoreimann I'm happy to update this PR. Once your comment above is addressed.

timoreimann:
@yue9944882  I asked the other maintainers internally: The percentage-based approach seems preferred.

I believe we are good to go.

ldez:
Could you use YML instead of JSON like for error page or buffering.
https://github.com/containous/traefik/blob/de3aeb97323951ec5de3ab7b4ed2217bea18b8ab/provider/kubernetes/kubernetes.go#L640-L653

yue9944882:
@timoreimann @ldez Updated , could u shed some light on this PR plz?

btw CI seems not working recently, what's happening :)

ldez:
```

---> Making bundle: validate-golint (in .)
Errors from golint:
provider/kubernetes/percentage.go:26:1: exported function PercentageValueFromFloat64 should have comment or be unexported

Please fix the above errors. You can test via "golint" and commit the result.
```

yue9944882:
@ldez Fixed. But previously I saw error messages like "docker.sock is not responding.." (but now it's different). I took it as errors from upstream CI 🤕. Should be my mistake and sry about that.

provider/kubernetes/percentage.go
```diff
@@ -0,0 +1,44 @@
...
+
+// PercentageValue is int64 form of percentage value with 10^-3 precision.
+type PercentageValue int64
+
+// PercentageValueFromString tries to read percentage value from string like "1.1%", "6%".
+// It will lose the extra precision if there are more than 3 digit after decimal point.
+func PercentageValueFromString(s string) (PercentageValue, error) {
+	idx := strings.Index(s, "%")
+	if idx < 0 {
+		return 0, fmt.Errorf("missing %% for parsing percentage value: %q", s)
```
dtomcej: I don't know if the `%` value is required.....is it really adding anything? You are still using `ParseFloat` on the string anyways....why not just strip it off if it is there, then continue on as usual. If you were parsing percentages differently than raw values, sure, but that doesn't appear to be the case here.

provider/kubernetes/percentage.go
```diff
@@ -0,0 +1,44 @@
...
+	"strconv"
+	"strings"
+)
+
+// PercentageValue is int64 form of percentage value with 10^-3 precision.
+type PercentageValue int64
+
+// PercentageValueFromString tries to read percentage value from string like "1.1%", "6%".
+// It will lose the extra precision if there are more than 3 digit after decimal point.
+func PercentageValueFromString(s string) (PercentageValue, error) {
```
dtomcej: Could you not use a variadic function for precision, so that it could be configured to allow for more digits if required?  `(s string, precision ...int)` ?

provider/kubernetes/percentage.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// Float64 returns its decimal float64 value.
+func (v PercentageValue) Float64() float64 {
+	return float64(v) / (1000 * 100)
+}
+
+// String returns its string form of percentage value.
+func (v PercentageValue) String() string {
+	return fmt.Sprintf("%.3f%%", v.Float64()*100)
```
dtomcej: again, 3 seems a bit arbitrary, see above comment.

provider/kubernetes/percentage_test.go
```diff
@@ -0,0 +1,39 @@
...
+		{"99.999%", false, 99999, "99.999%", 0.99999},
+		{"99.9999%", false, 99999, "99.999%", 0.99999},
+		{"-99.999%", false, -99999, "-99.999%", -0.99999},
+		{"-99.9999%", false, -99999, "-99.999%", -0.99999},
+		{"0%", false, 0, "0.000%", 0},
+		{"%", true, 0, "", 0},
+		{"foo", true, 0, "", 0},
+		{"1.11", true, 0, "", 0},
+		{"", true, 0, "", 0},
+	}
```
dtomcej: what about the case of `s = "0.5"`? Should that error?

provider/kubernetes/kubernetes.go
```diff
@@ -178,12 +178,34 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 		}
 		templateObjects.TLS = append(templateObjects.TLS, tlsSection...)
 
+		backendPercentageWeightMap, err := getPercentageWeightMap(i)
```
ldez:
```go
servicesPercentageWeights, err := getServicesPercentageWeights(i)
```

provider/kubernetes/kubernetes.go
```diff
@@ -178,12 +178,34 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+			continue
+		}
+
 		for _, r := range i.Spec.Rules {
 			if r.HTTP == nil {
 				log.Warn("Error in ingress: HTTP is nil")
 				continue
 			}
 
+			var leftFractionPercentage float64 = 1
```
ldez:
```go
leftFractionPercentage, leftFractionInstanceCount := getLeftFraction(k8sClient, r, servicesPercentageWeights, i)
```

```go
func getLeftFraction(k8sClient Client, rule extensionsv1beta1.IngressRule, servicesPercentageWeights map[string]*PercentageValue, i *extensionsv1beta1.Ingress) (float64, int) {
	var leftFractionPercentage float64 = 1
	leftFractionInstanceCount := 0

	for _, pa := range rule.HTTP.Paths {
		percentageWeight, found := servicesPercentageWeights[pa.Backend.ServiceName]
		if found {
			leftFractionPercentage -= percentageWeight.Float64()
			continue
		}

		endpoints, exist, err := k8sClient.GetEndpoints(i.Namespace, pa.Backend.ServiceName)
		if err == nil && exist {
			for _, subset := range endpoints.Subsets {
				leftFractionInstanceCount += len(subset.Addresses)
			}
		}
	}

	return leftFractionPercentage, leftFractionInstanceCount
}
```

morever, I think `leftFractionPercentage` must be a `PercentageValue` instead of `float64`.

provider/kubernetes/percentage.go
```diff
@@ -0,0 +1,62 @@
...
+// PercentageValueFromFloat64 reads percentage value from float64
+func PercentageValueFromFloat64(f float64) *PercentageValue {
+	return &PercentageValue{
+		value:     int64(f * 100 * 1000),
+		precision: defaultPercentageValuePrecision,
+	}
+}
+
+// RawValue returns its internal raw int64 form of percentage value.
+func (v *PercentageValue) RawValue() int64 {
```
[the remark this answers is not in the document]
ldez: For me the `rawValue` is a `float64` not a `int64` (`value:  int64(f * 100 * 1000)` is not really raw)

`Float64` and  `RawValue` must be rethink.

provider/kubernetes/kubernetes.go
```diff
@@ -299,6 +322,16 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
 							}
 
+							if percentageWeight, found := backendPercentageWeightMap[pa.Backend.ServiceName]; found {
+								instanceCount := 0
+								for _, subset := range endpoints.Subsets {
+									instanceCount = instanceCount + len(subset.Addresses)
+								}
+								serverWeight = int(percentageWeight.RawValue()) / instanceCount
+							} else if leftFractionPercentage < 1 {
+								serverWeight = int(PercentageValueFromFloat64(leftFractionPercentage).RawValue() / int64(leftFractionInstanceCount))
```
ldez: instead of:
```go
	serverWeight = int(percentageWeight.RawValue()) / instanceCount
} else if leftFractionPercentage < 1 {
	serverWeight = int(PercentageValueFromFloat64(leftFractionPercentage).RawValue() / int64(leftFractionInstanceCount))
```
I think you can add a method `computeWeight/getWeight` on the struct `PercentageValue`:

```go
func (p *PercentageValue) computeWeight(count int64) int { ... }
```

yue9944882:
> package golang.org/x/lint: unrecognized import path "golang.org/x/lint" (parse https://golang.org/x/lint?go-get=1: no go-import meta tags ())


@ldez What should I do to resolve this error plz?

ldez:
@yue9944882 it's a google/golang issue https://github.com/golang/lint/issues/397

yue9944882:
@ldez Shrug😿

provider/kubernetes/percentage.go
```diff
@@ -0,0 +1,90 @@
...
+// PercentageValueFromFloat64 reads percentage value from float64
+func percentageValueFromFloat64(f float64) *percentageValue {
+	return &percentageValue{
+		value: int64(f * 100 * 1000),
+	}
+}
+
+// rawValue returns its internal raw int64 form of percentage value.
+func (v *percentageValue) rawValue() float64 {
+	return float64(v.value)
```
ldez: This is not what I have suggested.

provider/kubernetes/kubernetes.go
```diff
@@ -677,3 +701,65 @@ func getRateLimit(i *extensionsv1beta1.Ingress) *types.RateLimit {
...
+		if err != nil {
+			return nil, fmt.Errorf("invalid percentage value %q in ingress %s: %v", percentageStr, i.Name, err)
+		}
+
+		servicesPercentageWeights[serviceName] = percentageValue
+	}
+	return servicesPercentageWeights, nil
+}
+
+func getLeftFraction(k8sClient Client, namespace string, paths []extensionsv1beta1.HTTPIngressPath, servicesPercentageWeights map[string]*percentageValue) (func(string) (*percentageValue, int), error) {
```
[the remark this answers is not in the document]
ldez: the closure was an example of the complexity, not a point to fix.
But, for now, keep the code as this.

provider/kubernetes/kubernetes.go
```diff
@@ -178,12 +178,24 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 		}
 		templateObjects.TLS = append(templateObjects.TLS, tlsSection...)
 
+		servicesPercentageWeightMap, err := getServicesPercentageWeights(i)
+		if err != nil {
+			log.Errorf("Invalid yaml format for backend weight annotation of ingress %s/%s: %v", i.Namespace, i.Name, err)
```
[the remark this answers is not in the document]
yue9944882: > return fmt.Sprintf("yaml: unmarshal errors:\n  %s", strings.Join(e.Errors, "\n  "))

Just print the error should be enough.

provider/kubernetes/kubernetes.go
```diff
@@ -677,3 +702,61 @@ func getRateLimit(i *extensionsv1beta1.Ingress) *types.RateLimit {
...
 	return rateLimit
 }
+
+func getServicesPercentageWeights(i *extensionsv1beta1.Ingress) (map[string]*percentageValue, error) {
+	rawPercentageWeights := getStringValue(i.Annotations, annotationKubernetesPercentageWeights, "")
+
+	percentageWeight := make(map[string]string)
+
+	if err := yaml.Unmarshal([]byte(rawPercentageWeights), percentageWeight); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
yue9944882: > return fmt.Sprintf("yaml: unmarshal errors:\n  %s", strings.Join(e.Errors, "\n  "))

Just print the error should be enough.

provider/kubernetes/annotations.go
```diff
@@ -31,6 +31,7 @@ const (
 	annotationKubernetesErrorPages                = "ingress.kubernetes.io/error-pages"
 	annotationKubernetesBuffering                 = "ingress.kubernetes.io/buffering"
 	annotationKubernetesAppRoot                   = "ingress.kubernetes.io/app-root"
+	annotationKubernetesPercentageWeights         = "ingress.kubernetes.io/percentage-weights"
```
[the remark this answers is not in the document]
yue9944882: I am not really sure about this name either. Is there a more suitable annotation name for this?

provider/kubernetes/kubernetes.go
```diff
@@ -677,3 +706,64 @@ func getRateLimit(i *extensionsv1beta1.Ingress) *types.RateLimit {
...
+	leftFractionInstanceCounts := make(map[string]int)
+	oneHundredPercentageValue := newPercentageValueFromFloat64(1)
+
+	for _, pa := range paths {
+		leftFractionPercentages[pa.Path] = oneHundredPercentageValue
+		leftFractionInstanceCounts[pa.Path] = 0
+	}
+
+	for _, pa := range paths {
+		endpoints, exist, err := k8sClient.GetEndpoints(namespace, pa.Backend.ServiceName)
```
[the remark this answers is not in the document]
yue9944882: Yes, I agree with you. This function is not really to simplify the abstraction but to make the loop logic over ingresses looks short as far as I think.
 
> Is there a chance we can inject the number of subset addresses

What do you mean by "inject"? To retrieve the endpoint at the very begin of iterating paths of ingress? Since that we are using informers to do the actual "get" or "list". The impact of performance should be neglectable.

provider/kubernetes/kubernetes.go
```diff
@@ -677,3 +706,64 @@ func getRateLimit(i *extensionsv1beta1.Ingress) *types.RateLimit {
...
+		if err != nil || !exist {
+			log.Warnf("fail to get endpoints %s/%s", namespace, pa.Backend.ServiceName)
+			continue
+		}
+
+		for _, subset := range endpoints.Subsets {
+			leftFractionInstanceCounts[pa.Path] += len(subset.Addresses)
```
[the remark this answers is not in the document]
yue9944882: IMHO, this annotation will work fine with multi-port service. Since that we are doing calculations based on numbers of instances for each endpoints, it's doesn't matter which port the ingress is actually referencing.
[https://github.com/yue9944882/traefik/blob/f88a7cdd60e67070135f02655f33d16a1a220dfa/provider/kubernetes/kubernetes_test.go#L2022](https://github.com/yue9944882/traefik/blob/f88a7cdd60e67070135f02655f33d16a1a220dfa/provider/kubernetes/kubernetes_test.go#L2022) This test could be an example for multi-port service.

timoreimann:
>>Couldn't we have maintained the integer-based approach and just put a small layer on top of it that translates percentages to integers (say, 20% plus 80% turn into 20 plus 80 = 100 integer weights, multiplied by a large enough factor to account for proper distribution among the backends)?
>
>Nice idea. That would be better both for readability. Small question for this solution, will this integer-based approach exported to the users as another annotation? I don't think we would need both integer-based and percentage-based annotations at the same time.

I believe we kind of agreed during the issue and PR discussion that we couldn't see a true benefit of the integer-based interface. Therefore, I think we should stick to a percentage-based one for now.

If you agree, then I'd suggest to adjust the PR accordingly before we continue with the review. Okay for you?

yue9944882:
@timoreimann Sure. Definitely :)

ldez:
@yue9944882 any news?

yue9944882:
@ldez SRY for delay🤭
I was busy working on another project for my job. It was a mess THO.. But I think I will update this PR within this week

yue9944882:
@timoreimann @ldez Ping for review. 
I created an abstraction of `WeightAllocator` which is used to assign integer weights for backends, meanwhile the fractional percentage weight computing logic is wrapped under `WeightAllocator`. How do you think about it?

provider/kubernetes/weight_allocator.go
```diff
@@ -0,0 +1,163 @@
...
+		for pa, svcs := range fractionalPathServices {
+			totalFractionalInstanceCount := 0
+
+			for _, svc := range svcs {
+				totalFractionalInstanceCount += serviceInstanceCounts[ingressService{
+					host:    rule.Host,
+					path:    pa,
+					service: svc,
+				}]
+			}
```
ldez: `totalFractionalInstanceCount` is used, seems like something is missing.

provider/kubernetes/kubernetes.go
```diff
@@ -188,6 +188,17 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 		}
 		templateObjects.TLS = append(templateObjects.TLS, tlsSection...)
 
+		var weightAllocator weightAllocator = &defaultWeightAllocator{}
+		annotationPercentageWeights := getAnnotationName(i.Annotations, annotationKubernetesPercentageWeights)
+		if _, ok := i.Annotations[annotationPercentageWeights]; ok {
+			fractionalAllocator, err := newFractionalWeightAllocator(i, k8sClient)
+			if err != nil {
+				log.Errorf("Failed to read backend percentage weights for ingress %s/%s: %v", i.Namespace, i.Name, err)
+				continue
```
ldez: Maybe use the `defaultWeightAllocator` in this case instead of skip.

provider/kubernetes/weight_allocator.go
yue9944882: @ldez Done

yue9944882:
@ldez Oops.. Seems like I made a bad mistake. 😵I forgot to rebase your previous `review: minor change` commit, now it's removed by my force push.  Could you please find that commit in your local `reflog` then rebase and push it again please? I am so sorry 😭

docs/configuration/backends/kubernetes.md
```diff
@@ -206,6 +208,15 @@ Non-root paths will not be affected by this annotation and handled normally.
 This annotation may not be combined with the `ReplacePath` rule type or any other annotation leveraging that rule type.
 Trying to do so leads to an error and the corresponding Ingress object being ignored.
 
+<5> `traefik.ingress.kubernetes.io/percentage-weights` example:
+
+```yaml
+service_backend1: 1% # Note that the field names must match service names referenced in the Ingress object.
+service_backend2: 33.33%
+service_backend3: 0.9
```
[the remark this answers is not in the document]
ldez: @yue9944882 after some discuss with @timoreimann and @dtomcej, we think that float value must interpret as a percentage value: `0.9` => `0.9%`.

By reformulating: the percentage sign must be optional.

marco-jantke:
Is there anything I can do to help bring this PR forward? My team is waiting eagerly for this feature.

yue9944882:
@timoreimann PTAL

ldez:
https://github.com/containous/traefik/pull/3112#discussion_r191408953

> @yue9944882 after some discuss with @timoreimann and @dtomcej, we think that float value must interpret as a percentage value: `0.9` => `0.9%`.
>
> By reformulating: the percentage sign must be optional.

yue9944882:
Rebase & resolved a conflict at:

[https://github.com/containous/traefik/blob/599b699ac95fa055a6c09ce6fc10a0484cfb4ad0/provider/kubernetes/kubernetes.go#L290](https://github.com/containous/traefik/blob/599b699ac95fa055a6c09ce6fc10a0484cfb4ad0/provider/kubernetes/kubernetes.go#L290)

yue9944882:
Rebase & resolving conflict at:

[https://github.com/containous/traefik/blob/599b699ac95fa055a6c09ce6fc10a0484cfb4ad0/provider/kubernetes/kubernetes.go#L290](https://github.com/containous/traefik/blob/599b699ac95fa055a6c09ce6fc10a0484cfb4ad0/provider/kubernetes/kubernetes.go#L290)

provider/kubernetes/weight_allocator.go
```diff
@@ -0,0 +1,163 @@
...
+
+	for _, rule := range ingress.Spec.Rules {
+		for _, pa := range rule.HTTP.Paths {
+			count := 0
+			endpoints, exists, err := client.GetEndpoints(ingress.Namespace, pa.Backend.ServiceName)
+			if err != nil {
+				return nil, fmt.Errorf("fail to get endpoints %s/%s: %v", ingress.Namespace, pa.Backend.ServiceName, err)
+			}
+			if !exists {
+				return nil, fmt.Errorf("fail to get endpoints %s/%s: non-existent endpoint", ingress.Namespace, pa.Backend.ServiceName)
```
timoreimann: Can we rephrase the text to `Endpoints not found for %s/%s` to be consistent with our [existing message](https://github.com/containous/traefik/blob/bddb4cc33c87108d6bc9ce1dc456024031dfe63b/provider/kubernetes/kubernetes.go#L308)?

provider/kubernetes/kubernetes.go
```diff
@@ -190,6 +190,18 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 		}
 		templateObjects.TLS = append(templateObjects.TLS, tlsSection...)
 
+		var weightAllocator weightAllocator = &defaultWeightAllocator{}
+		annotationPercentageWeights := getAnnotationName(i.Annotations, annotationKubernetesServiceWeights)
+		if _, ok := i.Annotations[annotationPercentageWeights]; ok {
+			fractionalAllocator, err := newFractionalWeightAllocator(i, k8sClient)
+			if err != nil {
+				log.Warnf("Failed to read backend percentage weights for ingress %s/%s: %v", i.Namespace, i.Name, err)
+				weightAllocator = &defaultWeightAllocator{}
```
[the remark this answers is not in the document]
dtomcej: I guess the argument is that a user editing the weights could completely disable the ingress with an error in the annotation, whereas the fallback would allow the ingress to continue with the default weights. Preference IMO.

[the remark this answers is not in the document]
dtomcej: I would be ok with falling back to default weights if the annotation is invalid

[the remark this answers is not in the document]
timoreimann: The dreadful scenario I'm thinking of is that I intend to serve 1% of traffic to my canary release, but due to a syntax bug in the annotation Traefik suddenly serves 100%, thereby overloading the application. A 404 served from Traefik because the frontend isn't even registered seems easier to detect/debug.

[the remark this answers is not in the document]
dtomcej: I can agree with that logic

yue9944882:
@timoreimann Thanks. Previous reviews addressed. Could u plz take a further look?

provider/kubernetes/kubernetes.go
```diff
@@ -190,6 +190,16 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 		}
 		templateObjects.TLS = append(templateObjects.TLS, tlsSection...)
 
+		var weightAllocator weightAllocator = &defaultWeightAllocator{}
+		annotationPercentageWeights := getAnnotationName(i.Annotations, annotationKubernetesServiceWeights)
+		if _, ok := i.Annotations[annotationPercentageWeights]; ok {
+			fractionalAllocator, err := newFractionalWeightAllocator(i, k8sClient)
+			if err != nil {
+				return nil, fmt.Errorf("failed to create fractional weight allocator for ingress %s/%s: %v", i.Namespace, i.Name, err)
```
timoreimann: Sorry, I probably should have been more clear in my original comment: what I meant by "failing fast" was that we skip the affected Ingress. Returning an error would have the negative effect that a single "bad" Ingress would cause processing for all other "good" Ingresses to stop as well, which is likely not what we want.

yue9944882: @timoreimann NP. Updated :)

provider/kubernetes/weight_allocator.go
```diff
@@ -0,0 +1,163 @@
...
+	return label.DefaultWeight
+}
+
+type ingressService struct {
+	host    string
+	path    string
+	service string
+}
+
+type fractionalWeightAllocator struct {
```
timoreimann: Couldn't `fractionalWeightAllocator` be typed as `map[ingressService]int` directly?

yue9944882:
@timoreimann Thanks a lot for detailed testing.

Plz see the commit `0da0056`. Now the constructor of fractionalWeightAllocator will fail in the following cases:

1. the weights in the same path exceeding  100%
2. the sum of weights in the same path less than 100% and no "fractional service" left for "auto-completing".

yue9944882:
1. Plz see my previous comment.

> I realize though that integer-to-float conversions / rounding errors may make this more complicated. WDYT? 

Yes. Supposing we have 2 serviceA, serviceB in a ingress rule and service and the annotation is:

```yaml
serviceA: 0.0001%
serviceB: 99.9999%
```
"0.0001%" will be taken as "0.000%" due to integer-to-float conversions and "99.9999%" will be "99.999%", so their sum is still not yielding 100% and this ingress will be skipped with error. Should we document it clearly to prevent users from abusing precision?

2. Ditto

3. The error msg is updated like: "percentage value %s must not exceed 100%"

4. 
> omitting a service now leads to 100% for that omitted service.

Umm. I think the users might also try to omit a service in order to split all its traffic to the "fractional" service. It makes sense at some point, isn't it?

5. 
> I'm okay with either doing it in the scope of this PR or a follow-up one (either by you or someone else, say, me).

Yes. It does confuse some users, especially if they are not quite familiar with YAML format. As the size of this PR grows, IMO maybe it would be better to place these changes into another.

trondhindenes:
This was the whole reason I was hoping weight would be based on arbitrary numbers and not a percentage.

dtomcej:
```
As the size of this PR grows, IMO maybe it would be better to place these changes into another.
```

I agree. I think we need to get this passing tests, then we can create a follow up PR to fix the specific details instead of holding up other PRs :)

Thoughts?

timoreimann:
@yue9944882:

1. Agreed to documenting the boundaries. IMHO, having 4 decimal places should be good enough for the vast majority of cases. Bonus points if we can make the error message more clear, but we can also omit this point if it helps finishing the PR earlier.
2. Ditto.
3. 👍 
4. Not exactly sure I know what you mean. What's your expected, preferred behavior for the case described (i.e., a single service has no percentage assigned)?
5. 🆗 

@dtomcej 

>I agree. I think we need to get this passing tests, then we can create a follow up PR to fix the specific details instead of holding up other PRs :)

Agreed in general.

Except for point 4 (omitting the percentage for a single service), I think everything else is easy to add within this PR (e.g., quick doc fixes) or already done. I'm okay with skipping point 4 too but would like to know where we want to head and make sure the state of the PR allows us to go there (e.g., not introduce a breaking behavioral change just a few PRs from now).

yue9944882:
@timoreimann 

> However, omitting a service now leads to 100% for that omitted service

Supposing we have two service backends `A (1 server explicitly 25% traffic)` and `B(1 server implicitly 75% traffic)` for `foo.com/bar`, I think following is what'll happen when we do the deletion:

- Deleting `A` will make `B` serves 100% traffic because `B` is the only backend left and it's not annotated. E.g. users might just want this when reverting a 25% canary from `A`. In terms of integer weights, before the deletion the server of `A`'s priority would be 25000 and `B` would be 75000. After deleting `A`, `B` would be 100000. It looks good to me for now, WDYT?

- Deleting `B` will make the reload fail and throw an error because `B` is explicitly specified to serve 25% traffic but no other backend can handle the left 75%. It violates with:

> 2. the sum of weights in the same path less than 100% and no "fractional service" left for "auto-completing".

yue9944882:
> I'm okay with skipping point 4 too but would like to know where we want to head and make sure the state of the PR allows us to go there.

I'm not really sure about this. It might depend on the feedback IMHO?

timoreimann:
@yue9944882 I tried to understand / verify what the current state of the PR does. You wrote that

>Now the constructor of _fractionalWeightAllocator_ will fail [if] the sum of weights in the same path less than 100% and no "fractional service" left for "auto-completing".

I read this as: if there is one service left for "auto-completing", the constructor should *not* fail. How do I let the implementation auto-complete?

I tried both

```yaml
traefik.ingress.kubernetes.io/service-weights: |
  service1: 75%
```

and

```yaml
traefik.ingress.kubernetes.io/service-weights: |
  service1: 75%
  service2:
```

for a two service backends Ingress, but to no avail: I always see the error message

```
ERRO[2018-06-22T18:57:43+02:00] failed to create fractional weight allocator for ingress default/service1: the sum of weights(75.000%) in the path //my/service1 must be 100% when no omitted fractional service left
```

Am I holding it wrong?

yue9944882:
@timoreimann Yeah, you've got my point. For a two backends ingress:

> traefik.ingress.kubernetes.io/service-weights: |
  service1: 75%

This one should not fail. I wrote another simple test according to your description but it works fine. Could you put some more detail about your test? 
```go
func TestSimpleServiceAnnotation(t *testing.T) {
	ingresses := []*extensionsv1beta1.Ingress{
		buildIngress(
			iAnnotation(annotationKubernetesServiceWeights, `
service1: 10%
`),
			iNamespace("testing"),
			iRules(
				iRule(
					iHost("host1"),
					iPaths(
						onePath(iPath("/foo"), iBackend("service1", intstr.FromString("8080"))),
						onePath(iPath("/foo"), iBackend("service2", intstr.FromString("8080"))),
					)),
			),
		),
	}
	services := []*corev1.Service{
		buildService(
			sName("service1"),
			sNamespace("testing"),
			sUID("1"),
			sSpec(
				clusterIP("10.0.0.1"),
				sPorts(sPort(8080, "")),
			),
		),
		buildService(
			sName("service2"),
			sNamespace("testing"),
			sUID("1"),
			sSpec(
				clusterIP("10.0.0.1"),
				sPorts(sPort(8080, "")),
			),
		),
	}

	endpoints := []*corev1.Endpoints{
		buildEndpoint(
			eNamespace("testing"),
			eName("service1"),
			eUID("1"),
			subset(
				eAddresses(
					eAddress("10.10.0.1"),
					eAddress("10.10.0.2"),
				),
				ePorts(ePort(8080, "")),
			),
		),
		buildEndpoint(
			eNamespace("testing"),
			eName("service2"),
			eUID("1"),
			subset(
				eAddresses(
					eAddress("10.10.0.3"),
					eAddress("10.10.0.4"),
				),
				ePorts(ePort(8080, "")),
			),
		),
	}

	watchChan := make(chan interface{})
	client := clientMock{
		ingresses: ingresses,
		services:  services,
		endpoints: endpoints,
		watchChan: watchChan,
	}
	provider := Provider{}

	actual, err := provider.loadIngresses(client)
	require.NoError(t, err, "error loading ingresses")

	expected := buildConfiguration(
		backends(
			backend("host1/foo",
				servers(
					server("http://10.10.0.1:8080", weight(int(newPercentageValueFromFloat64(0.05)))),
					server("http://10.10.0.2:8080", weight(int(newPercentageValueFromFloat64(0.05)))),
					server("http://10.10.0.3:8080", weight(int(newPercentageValueFromFloat64(0.45)))),
					server("http://10.10.0.4:8080", weight(int(newPercentageValueFromFloat64(0.45)))),
				),
				lbMethod("wrr"),
			),
		),
		frontends(
			frontend("host1/foo",
				passHostHeader(),
				routes(
					route("/foo", "PathPrefix:/foo"),
					route("host1", "Host:host1")),
			),
		),
	)

	assert.Equal(t, expected, actual, "error loading percentage weight annotation")
}

```
But as for the following one, I saw a failure due to invalid yaml format. And I add another test case for  it.

>  traefik.ingress.kubernetes.io/service-weights: |
  service1: 75%
  service2:

timoreimann:
@yue9944882 this is the YAML I have been tested with:

```yaml
kind: Ingress
metadata:
  annotations:
    traefik.frontend.rule.type: PathPrefixStrip
    traefik.ingress.kubernetes.io/service-weights: |
      my-test1: 75%
  creationTimestamp: 2018-06-13T14:55:44Z
  generation: 2
  labels:
    k8s-app: my-test1
  name: my-test1
  namespace: default
  resourceVersion: "26903202"
  selfLink: /apis/extensions/v1beta1/namespaces/default/ingresses/my-test1
  uid: d9537967-6f19-11e8-8a4c-42010a8400db
spec:
  rules:
  - http:
      paths:
      - backend:
          serviceName: my-test1
          servicePort: 80
        path: /my/test1
      - backend:
          serviceName: my-test2
          servicePort: 80
        path: /my/test2
```

yue9944882:
> @yue9944882 this is the YAML I have been testing with:

As for your testing YAML,  `/my/test1` has only 1 backend `my-test1` and it is already specified 75% weight but there is no other backend to handle the left 25% weight so the ingress fails. Isn't it just the behavior we want?🤔

emilevauge:
Hi!
The code freeze for the next 1.7 release will happen at the end of next week, on Friday 6th July 2018.
This PR should then be mreged **before** this deadline :)

timoreimann:
@yue9944882 

>As for your testing YAML, `/my/test1` has only 1 backend `my-test1` and it is already specified 75% weight but there is no other backend to handle the left 25% weight so the ingress fails. Isn't it just the behavior we want?🤔

So what would be the correct syntax to let Traefik auto-complete `my-test2` at 25%? I can't seem to make it happen.

yue9944882:
> So what would be the correct syntax to let Traefik auto-complete the my-test2 backend at 25%? I can't seem to make it happen.

Umm..IMHO, even though `"/my/test1"->my-test1` and `"/my/test2"->my-test2` are defined in the same ingress, they are actually irrelevant rules. It's safer that the changes on `"/my/test1"->my-test1` doesn't influence the other.🤔btw what's the use case of your test case? I suppose what you are trying to do is sort of like backend immigrating between paths, am I right? Current implementation is basically path-oriented i.e. the weight will be calculated and auto-completed within a single path. The classic use case is that we want to bump backends of path "/my/test1" from "my-test1" to "my-test2", which is a process of canary release for "/my/test1".

Could you introduce a use-case for your test YAML please? @timoreimann Sadly I'm afraid I really didn't consider this usage before.

trondhindenes:
I urge you to reconsider the usage of percentages here.
1: It doesn't align with existing functionality in Traefik (arbitrary weight)
2. It's (as @timoreimann shows above) _extremely_ dangerous. 

I've been against this from the start. We're starting to consider other ingress controllers, since we need this functionality but the implementation in this PR just seems completely wrong.

I'm just a user here (not even a support-paying one) but please consider the above points.

yue9944882:
>1: It doesn't align with existing functionality in Traefik (arbitrary weight)

I don't think it a problem. Or it could be supported in the future without confliction. Previously as @timoreimann suggests, we are justing putting a tiny layer of weight allocator above Traefik's arbitrary weight functionality and the percentage-based allocation is just one of the possible implementation. I believe you might overwrite the allocation logic via specifying another `weightAllocator` e.g. via an annotation like: `"weight-alloc.traefik.io": "percentage" or "integer" `. 

IMO, percentage-based allocation is more user-friendly to manually operations and might be more suitable for kubernetes operation engineers. And as for your feature request, it's more flexible to fit Traefik into automation systems. It depends on the use case.

>2. It's (as @timoreimann shows above) extremely dangerous.

I'm not really sure about this. 😲

ldez:
@yue9944882 @trondhindenes @timoreimann I propose to you to have a discussion on Slack in order to move more quickly on the subject.
Channel  [#dev](https://traefik.slack.com/messages/C2S653DAM/)

marco-jantke:
@trondhindenes right now Traefik alread picks up multiple services with the same path in an ingress and will merge them into one backend. That means if you control the weight of your backend servers carefully on your own you would be already able to do what you want. Agree? Please correct me if I miss something important.

timoreimann:
>>1: It doesn't align with existing functionality in Traefik (arbitrary weight)

>IMO, percentage-based allocation is more user-friendly for manually operations and might be more helpful to kubernetes operation engineers. And as for your feature request, it's more flexible to fit Traefik into automation systems. It depends on the use case.

That is my key argument as well. I find percentages more appropriate to express traffic volume relationships as you normally want to do in canary systems, so they are worth the extra effort. Weights seem too low an abstraction to me in general. But as @yue9944882 said, nothing speaks against extending the feature in the future given a use case arises.

>It's (as @timoreimann shows above) extremely dangerous.

I'm not exactly sure which part of my comments you may be referring to. If it is about the various boundaries (not less than 100%, not more than 100%, etc.): That's certainly an extra code burden, but I consider it to be an opportunity for us to sanity-check user input. With pure weights, it'd be up to the user alone to make sure that the relationships are appropriate.

>I've been against this from the start. We're starting to consider other ingress controllers, since we need this functionality but the implementation in this PR just seems completely wrong.

Which part(s) of the implementation do you consider wrong?
By the way, [previously](https://github.com/containous/traefik/pull/3112#issuecomment-381396515) you seemed to have expressed approval or at least indifference to the approach we agreed on.

timoreimann:
@yue9944882 

>Umm..IMHO, even though "/my/test1"->my-test1 and "/my/test2"->my-test2 are defined in the same ingress, they are actually irrelevant rules.

Argh, DAMNIT -- all this time, I missed to see that my paths for services `my-test1` and `my-test2` did **not** match. I had copy-pasted blindly; as soon as I replaced `path: /my/test2` by `path: /my/test1`, it worked as intended. Sorry, my bad!

With that, my last open point seems resolved. I'll have another final look over the PR.

Additionally, I'd like to conclude the discussion with @trondhindenes to understand if there's a use case we could not cover with the given implementation now or through an extension in the future.

timoreimann:
Telling the bot to back off from merging just a bit longer in order to finish any last discussions.

dtomcej:
This PR makes me happy. It provides the basis for some really cool features.

trondhindenes:
Happy to put this thru its paces in dev as-is - I might be over-reacting to a percentage-based implementation since I don't agree with it. Anyways, we'll deploy this to dev clusters and spend some time testing it out + getting feedback from devs that it meets their needs. Happy to discuss in Slack after we've put some mileage on it.

timoreimann:
@trondhindenes thanks, I appreciate it a lot. You've been part of the discussion of this feature from day one, so I'm definitely keen on hearing your feedback once you've had a chance to test-drive things.

PR #3130, opened 2018-04-04
Server weight zero

### LINKED ISSUE #1294: Backend server with weight "0"

ldez:
## What does this PR do?

Currently, a server with a  weight of `0` continue to receive calls: it's because if a server have a weight of `0` the real weight is `1`.

These PR change this behavior:
- a server with a weight of `0` will have a real weight of `0`.
- if a server have a weight of `0`, the server will be removed from the available servers.

### Motivation

Fixes #1294

### More

- [x] Added/updated tests

### Additional Notes

Related to https://github.com/vulcand/oxy/pull/121, https://github.com/containous/oxy/pull/18, #1780

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -263,8 +263,8 @@
 [[projects]]
   name = "github.com/containous/traefik-extra-service-fabric"
   packages = ["."]
-  revision = "29a6d70ad0f15175efbaa5fd93d8afdd8b373b93"
-  version = "v1.1.1"
+  revision = "503022efdc178146d598911092af75690510a80c"
+  version = "v1.1.3"
 
 [[projects]]
   name = "github.com/coreos/bbolt"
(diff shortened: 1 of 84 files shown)


yue9944882:
@ldez @timoreimann  if 0 weight is semantically equal with 1 weight, how could we isolate or make a server unreachable 🧐A potential 0-weight use case is for canary release in which sometimes we want to do a 0% canary release or rollback.

yue9944882:
Looking forward to this one too!

kachkaev:
`AllowMinWeightZero` in toml might need to be renamed to `allowMinWeightZero` for consistency with other params (https://github.com/containous/traefik/pull/3132)

PR #3132, opened 2018-04-04
Normalize parameter names in configs

### LINKED ISSUE #3005: Naming consistency in configuration: paramName vs ParamName
Some parameters [in `traefik.toml`](https://docs.traefik.io/configuration/commons/) such as `graceTimeOut`, `checkNewVersion`, `debug`, etc begin with a lowercase character, while others start with a capital letter (`ProvidersThrottleDuration`, `MaxIdleConnsPerHost`, `RootCAs`, etc). A quick glance through the traefik's repo and the PRs suggests that a decision to make each parameter styled in its particular way is justified by how it is named in the source code. However, this logic is out of reach for an end user and so the params in `traefik.toml` simply look inconsistently named without any clear reason.

I was going to add `InsecureSkipVerify` option [to traefik's official helm chart](https://github.com/kubernetes/charts/tree/master/stable/traefik) and somewhat stumbled, because I could not figure out how to name it. If I kept the first uppercase letter, I'd get inconsistency with all other chart parameters, while if I named the option as `insecureSkipVerify`, the inconsistency would shift, but not disappear. At the moment, there seems to be no good naming solution that would not raise end users' questions and thus their waste of time.

It looks like both `InsecureSkipVerify` and `insecureSkipVerify` work equally well in `traefik.toml`, which means that the existing naming inconsistency can be resolved by just updating the docs and this will not introduce any breaking changes. Given the size of the traefik's community, I find it worth investing a bit of time in making config naming more coherent and thus avoid time-draining questions like _should this parameter start with a cap? and this?_

WDYT of going through the docs and renaming all configuration parameters that begin with a capital letter (e.g. `RootCAs` → `rootCAs`)? What can this break?

kachkaev:
### What does this PR do?

This PR is a response to https://github.com/containous/traefik/issues/3005, where it was agreed to improve consistency in config parameter names. It:
* renames parameter names to lowerCamelCase in tomls throughout the docs (e.g. `InsecureSkipVerify` becomes `insecureSkipVerify`)
* ~~~standardizes the use of term `entrypoint` (renames occurrences of `entry point` and `entryPoint`)~~~

### Motivation

Close #3005

### More

### Additional Notes

<s>Renaming `entryPoint` to `entrypoint` may not sound like an obvious improvement in the first place, but let me explain the motivation behind this decision. Initially, I simply noticed that all three possible variants (`entryPoint`, `entrypoint` and `entry point`) are used in the docs, which required some cleanup. I started renaming the occurrences of `entrypoint` to `entryPoint` in tomls and `entry point` in text, however, in the middle of the process I noticed that a few things were getting worse. This made me I change my mind about the naming of this term and go into the opposite direction. Here are the problems that getting rid of `entrypoint` was creating:

* I had to rename a header in `basics.html`, which would break at [least one anchor link](https://docs.traefik.io/basics/#entrypoints) as well as the search:
     screen shot 2018-04-04 at 22 27 59

* `entryPoint` was becoming different to other common terms like `frontend` or `filename` and this looked pretty bad when such terms stood next to each other.
* Because docker compose uses [`entrypoint`](https://docs.docker.com/compose/compose-file/#entrypoint), an inconsistency was occurring.

I have not renamed `EntryPoint` to `Entrypoint` in any of the Go code, but I'm wondering if this should be done in future as well.

FYI: Renaming `entrypoint` to `entryPoint`/`entry point` would not make the PR smaller, both flavours of this term were very common in the docs.</s>

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -645,18 +645,18 @@ Once a day (the first call begins 10 minutes after the start of Træfik), we col
   swarmMode = true
 
   [Docker.TLS]
-    CA = "dockerCA"
-    Cert = "dockerCert"
-    Key = "dockerKey"
-    InsecureSkipVerify = true
+    ca = "dockerCA"
+    cert = "dockerCert"
+    key = "dockerKey"
+    insecureSkipVerify = true
 
 [ECS]
-  Domain = "foo.bar"
-  ExposedByDefault = true
-  Clusters = ["foo-bar"]
-  Region = "us-west-2"
-  AccessKeyID = "AccessKeyID"
-  SecretAccessKey = "SecretAccessKey"
+  domain = "foo.bar"
+  exposedByDefault = true
+  clusters = ["foo-bar"]
+  region = "us-west-2"
+  accessKeyID = "AccessKeyID"
+  secretAccessKey = "SecretAccessKey"
 ```
 
 - Obfuscated and anonymous configuration:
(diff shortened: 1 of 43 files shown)


kachkaev:
If tests pass for 049bf87870d3c4bcbb9e8b9a9575e28e3bc95275, I think I'm ready for another review.

ldez:
@kachkaev yes you can "revert/keep" 6f1fbb378db870e765416cf09018d0d92dcc8aab

ldez:
Could you squash your commits.

kachkaev:
@ldez squash done. Left the [original unsquashed commits](https://github.com/containous/traefik/compare/03ce6a1cc4947f138ac3f7893322befd67556dc3...kachkaev:normalize-parameter-names-before-squash) in branch `normalize-parameter-names-before-squash` of my fork, just in case.

kachkaev:
Many thanks for your reviews @ldez, @mmatur and @Juliens! I'm exited to add my 50 cent to traefik despite zero knowledge in Go! 😅  What you're doing is awesome and really useful to the world!

PR #3148, opened 2018-04-07
update variable names in dumpcert to reflex changes in acme.json

butlerx:
### What does this PR do?

Fix script for dumping certs.

### Motivation

I went to export certs and i got errors telling me `acme.json` couldn't be parsed.

### More

- [ ] Added/updated tests

### DIFF
--- a/contrib/scripts/dumpcerts.sh
+++ b/contrib/scripts/dumpcerts.sh
@@ -92,7 +92,6 @@ ${USAGE}" >&2
 	exit 2
 fi
 
-
 if [ ! -d "${certdir}" ]; then
 	echo "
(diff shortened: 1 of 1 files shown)


ldez:
As @nmengin  said in the issue 3115#issuecomment-378550214, the script has been changed to allow extracting files in new format.

That's why the modifications you made about it cannot be merged.

Use https://github.com/containous/traefik/blob/v1.5/contrib/scripts/dumpcerts.sh instead.

butlerx:
ah hadn't noticed that sorry
There was a couple other things i noticed in the handling of bad_acme i noticed but i can open a separate pr for that

PR #3116, opened 2018-04-01
Minor updates to dumpcerts.sh

mathuin:
Closes #3115.

There was an extraneous $1 on line 69 which threw an error.  The locations of keys and certificates in the JSON file have changed.

### What does this PR do?

This pull request modifies `dumpcerts.sh` to run without errors and to access keys and certs in the JSON file correctly.

### Motivation

I now use Traefik to manage my LetsEncrypt certs for my mail server, so I needed to learn how to extract the certs and keys as files.

### More

- [ ] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

I'm very new to Traefik.  All I can say about this script is "it works for me".  If it does not work for the general case, I would love to get some feedback to make a proper solution that works for both me and everyone else.  Thank you in advance for your patience!

PR #3149, opened 2018-04-07
pass arges to bad_acme so it can print acme path in error

butlerx:
### What does this PR do?
passes args to functions so path can be printed 

### Motivation

was using the wrong version of the script and the error messages were not being printed correctly 

### More

- [ ] Added/updated tests


nmengin:
Hello @butlerx,

Many thanks for your interest in the project.
Your PR seems to be a duplicated of #3116 

This PR fix the probelm by deleting the `$1` option which is not needed.

I close your PR.

PR #3170, opened 2018-04-10
Fix: receiving unknown flag --api, switched to --web flag to end web …

justinlarose:
…backend

### What does this PR do?

Fixes the example for quickstart. When running docker-compose received this error:
```
Error parsing command: unknown flag: --api
Error here : unknown flag: --api
```
Changed to `--web` flag instead

### More

- [ ] Added/updated tests

### DIFF
--- a/examples/quickstart/README.md
+++ b/examples/quickstart/README.md
@@ -14,10 +14,10 @@ version: '3'
 services:
   reverse-proxy:
     image: traefik #The official Traefik docker image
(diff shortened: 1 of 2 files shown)


ldez:

I think you are using an old version of Traefik (<1.4).

`--web` is deprecated sine 1.4 when `--api` has been added.

Try to pull a new version: `docker pull traefik:1.5`

PR #3187, opened 2018-04-13
fix: redirect to HTTPS first before basic auth if header redirect (secure) is set

### LINKED ISSUE #3134: Do SSLRedirect before auth.basic

Bug

### What did you do?

Deploy a docker service with the following labels:

```yaml
        - traefik.frontend.rule=Host:private.example.com
        - traefik.frontend.entryPoints=http,https
        - traefik.frontend.auth.basic=username:password
        - traefik.frontend.headers.SSLRedirect=true
```

Opening `http://private.example.com` redirects to `https://private.example.com` before prompting for authorisation.

### What did you see instead?

`http://private.example.com` requires authorisation before redirecting. This means user has to enter authorisation data twice, and the first authorisation data is sent over an insecure connection.

```
1.5.4
```

SantoDE:
### What does this PR do?

Fixes #3134 by moving the order of middleware invoking

### Motivation

Don't transfer basic auth credentials over an insecure connection if security headers are set.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1210,7 +1210,7 @@ func (s *Server) loadConfig(configurations types.Configurations, globalConfigura
 						} else {
 							lb = bufferedLb
 						}
-					}	
+					}
 
 					if config.Backends[frontend.Backend].CircuitBreaker != nil {
 						log.Debugf("Creating circuit breaker %s", config.Backends[frontend.Backend].CircuitBreaker.Expression)
--- a/server/server.go
+++ b/server/server.go
@@ -1174,6 +1174,16 @@ func (s *Server) loadConfig(configurations types.Configurations, globalConfigura
 						}
 					}
 
+					if headerMiddleware != nil {
+						log.Debugf("Adding header middleware for frontend %s", frontendName)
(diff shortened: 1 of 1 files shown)


server/server.go
```diff
@@ -1192,26 +1202,6 @@ func (s *Server) loadConfig(configurations types.Configurations, globalConfigura
...
-						log.Debugf("Adding header middleware for frontend %s", frontendName)
-						n.Use(s.tracingMiddleware.NewNegroniHandlerWrapper("Header", headerMiddleware, false))
-					}
-
-					if secureMiddleware != nil {
-						log.Debugf("Adding secure middleware for frontend %s", frontendName)
-						n.UseFunc(secureMiddleware.HandlerFuncWithNextForRequestOnly)
-					}
-
-					if config.Backends[frontend.Backend].Buffering != nil {
```
mmatur: Bye bye Buffering :wave:

dtomcej: Buffering is for the weak.

SantoDE: Nothing happened @dtomcej @mmatur :D 😇

Issue #3003, opened 2018-03-12
Bug: Command "storeconfig" resets certificates 


davidhiendl:
When using the `storeconfig` command to store or update the configuration in a key-value store all previously requested ACME certificates are deleted. 
This is potentially problematic when updating the configuration is required on a cluster with many certificates as rate limits would prevent the re-acquisition of all those certificates.

### What did you do?
- Create a consul cluster
- Store initial configuration with one-off job `traefik storeconfig --configfile=/etc/traefik/traefik.toml` 
- Start traefik load balancer instances that use consul: `traefik --consul --consul.endpoint=consul:8500`
- Open a few of the configured hosts to trigger ACME certificate requests
- Update the configuration again by running the `traefik storeconfig --configfile=/etc/traefik/traefik.toml` 

### What did you expect to see?
Configuration should be updated but certificates should remain.

### What did you see instead?
The entire configuration is updated and certificates are reset.

### Output of `traefik version`:
Docker: traefik:1.5
```
Version:      v1.5.3
Codename:     cancoillotte
Go version:   go1.9.4
Built:        2018-02-27_02:47:04PM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
    debug = false
    checkNewVersion = false

    # traefikLogsFile = "log/traefik.log"
    # accessLogsFile = "log/access.log"
    logLevel = "INFO"

    InsecureSkipVerify = false
    defaultEntryPoints = ["http", "https"]

    [entryPoints]
      
      [entryPoints.http]
      address = ":80"
      #[entryPoints.http.redirect]
      #entryPoint = "https"
      
      [entryPoints.https]
      address = ":443"
        [entryPoints.https.tls]
        minVersion = "VersionTLS12"
        MozillaRecommendedConfiguration = "modern"

    
    [acme]
    email = "***@*******.tld"
    entryPoint = "https"
    acmeLogging = true
    # onDemand = false
    onHostRule = true
    caServer = "https://acme-staging.api.letsencrypt.org/directory"
    storage = "traefik/acme/account"
    [acme.httpChallenge]
      entryPoint = "http"

    [web]
    address = ":8080"

    [kubernetes]

    [consul]
    endpoint = "consul:8500"
    watch = true
    prefix = "traefik"
```

thomasjpfan:
[...]

PR #3202, opened 2018-04-18
Enable to override certificates in key-value store when using storeconfig

thomasjpfan:
### What does this PR do?

Adds `--acme.keepcerts` option to traefik. When this option is set to true, certificates already in key-value store will not be overwritten when using `storeconfig`.

### Motivation

Closes https://github.com/containous/traefik/issues/3003. 

### More

- [ ] Added/updated tests.


emilevauge:
@thomasjpfan thanks a lot for your contribution :)
We had a discussion on this topic, and we think the initial issue is a bug, we shouldn't erase existing certificates from KV store when calling `storeconfig`. Instead, we think we should keep existing certificates in KV store, and erase them while calling `storeconfig` only if providing an option `--acme.overrideCertificates`.
WDYT ?

thomasjpfan:
That is a great idea. I will update the PR to reflect your suggestion.

PR #3225, opened 2018-04-23
Remove acme provider dependency in server

Juliens:
### What does this PR do?
Remove acme provider dependency in server.

### Motivation
Better separation of concern

### More
- [x] Added/updated tests


server/server.go
```diff
@@ -79,10 +78,18 @@ type Server struct {
 	entryPoints                   map[string]EntryPoint
 }
 
+// CertificateStore store for dynamic and static certificates
+type CertificateStore struct {
```
mmatur: Could you try to move this struct in `types` package please

Issue #1816, opened 2017-07-02
Kubernetes ExternalNames cannot be mapped to custom ports via the Service port


timoreimann:
### Do you want to request a *feature* or report a *bug*?
Bug.

### What did you do?
Create an Ingress and a Service object in Kubernetes where the latter specifies an ExternalName and a Service port/target port mapping.

### What did you expect to see?
The ExternalName URL is being requested on the mapped target port.

### What did you see instead?
The target port is always 80.

### Output of `traefik version`: (_What version of Traefik are you using?_)
master.

### Details

Right now, the Service port is only required to determine if the protocol should be HTTPS or HTTP (depending on the port being 443 or not). It'd be convenient if the target port was used when forwarding requests to the external name.

FWIW, the Nginx controller [does the same](https://github.com/kubernetes/ingress/issues/621).

timoreimann:
I discovered the status quo while formulating a [response on a different issue](https://github.com/containous/traefik/issues/1277#issuecomment-312121758).

@dtomcej have been discussing options. I went ahead a filed a feature request for a direction that looked legitimate to me. Happy to discuss.

trnl:
To me it looks like it is fixed in https://github.com/kubernetes/ingress/pull/629/files

timoreimann:
@trnl you are right that ExternalNames don't provide endpoints -- their sole purpose is to redirect to some DNS name.

What this issue is about though is to extend Traefik such that any Service port mapping will be incorporated into the forwarded request. Right now, all ExternalName requests may only hit port 80. We want to make it possible that requests may go to an arbitrary port as defined by the Service object.

wojciech12:
@timoreimann  You can achieve  using Service and Enpoint objects, see https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service :

```
---
kind: Service
apiVersion: v1
metadata:
  name: my-service
spec:
  ports:
  - protocol: TCP
    port: 80
    targetPort: 9376
---
kind: Endpoints
apiVersion: v1
metadata:
  name: my-service
subsets:
  - addresses:
      - ip: 1.2.3.4
    ports:
      - port: 9376
```

timoreimann:
@wojciech12 that doesn't seem to allow accessing the service through a DNS name but only an IP address, does it?

wojciech12:
@timoreimann  True. It let you to define the target port. Traefik does not respect the target port, when the service is external.

timoreimann:
[...]

PR #3231, opened 2018-04-24
 Mapping ExternalNames to custom ports

gildas:
### What does this PR do?

This fixes issue #1816 

Basically, if the `servicePort.Port` is not 443 or 80, we add it to the url (not the name).
And if the `servicePort.name` is "https" we also use "https" as the protocol. That allows to proxy https traffic to any port.

### Motivation

In my deployment, I need træfik to proxy to external servers on custom ports. I do not control these servers.

### More

- [x] Added/updated tests

### DIFF
--- a/provider/kubernetes/builder_service_test.go
+++ b/provider/kubernetes/builder_service_test.go
@@ -120,6 +120,23 @@ func TestBuildService(t *testing.T) {
 	)
 
 	assert.EqualValues(t, sampleService2(), actual2)
+
+	actual3 := buildService(
+		sName("service3"),
+		sNamespace("testing"),
+		sUID("3"),
+		sSpec(
+			clusterIP("10.0.0.3"),
+			sType("ExternalName"),
+			sExternalName("example.com"),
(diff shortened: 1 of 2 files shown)


emilevauge:
WDYT @containous/kubernetes ?

dtomcej:
I can get behind checking to see if the port name has an `https` prefix. I can't think of any reason this should _not_ be allowed.

👍

gildas:
It would also be very easy to change the PR from:
```go
						if port.Port == 443 || port.Name == "https" 
```
to:

```go
						if port.Port == 443 || strings.HasPrefix(port.Name, "https") 
```

dtomcej:
ping @timoreimann

billimek:
This is merged to master but still doesn't appear to be in any of the traefik releases.  What is the typical path forward to get something like this included in a release?

timoreimann:
@billimek the milestone indicates that the PR will be part of Traefik 1.7. Thus, the next release will include the feature.

Issue #3224, opened 2018-04-23
Add a matcher for partitioning services 


lawrencegripper:
### Do you want to request a *feature* or report a *bug*?

Feature (I'll write it)

### What did you do?

Service Fabric uses partitioning of services to improve scalability. I would like to add a matching rule which allows requests to be partitioned. A frontend would be created per partition and the matching rule would ensure requests are matched to the correct frontend based on the value of a hash function – allowing you to evenly distribute across n number of partitions. This would be useful to other providers, for example allowing requests to be partitioned across multiple container instances or services in Kubernetes. 

Additional discussion: https://github.com/jjcollinge/traefik-on-service-fabric/issues/45

### Proposal

Add an additional matching rule to Traefik which enables a hashed range match for example `HashedRange: type:header value:x-partitionheader match:0-100 range:0-300 `. It would take an input and use a hashing algorithm to convert this to an `int` with even distribution in a range. In this case the full range would be 0-300 and this rule would match if the hashed result of the header `x-partitionheader` fell in the range 0-100. 

This could be used to create 3 partitions with a KeyMin=0 and KeyMax=300 for example and distribute load between them:

- Frontend for Parition 1 with matcher `HashedRange: type:header value:x-partitionheader match:0-100 range 0-300`
- Frontend for Parition 2 with matcher `HashedRange: type:header value:x-partitionheader match:100-200 range 0-300`
- Frontend for Parition 2 with matcher `HashedRange: type:header value:x-partitionheader match:200-300 range 0-300`

In addition to the `type:header` option I would also look to add `url-regex` which would match a section of the url to hash,

I can think of more `types` but I think these two cover most use cases.

### Example of url-regex type 

`URL: http://example.com/bob/?customerid=jamesnesbit`
`HashedRange: type:url-regex value:[=].* match:0-100 range:0-300`

This would hash `jamesnesbit` and match if it the result was in range 0-100

### What did you expect to see?

The service fabric provider would query stateful services and create a frontend for each partition with the appropriate `hashedrange` matcher. Requests would then match the correct partition based on the value of their `header` or `url-regex`. 

### What did you see instead?

I don't believe it's currently possible to achieve this behavior in Traefik

CC: @jjcollinge

geraldcroes:
Hi, I'm quite new to the Service Fabric world so excuse my candor here.

Why can't you use the existing matchers in your use case? Like, a simple regex-based matcher?
Why do you need to hash the value?

[...]

PR #3239, opened 2018-04-26
Add a matcher for partitioning services

lawrencegripper:
### What does this PR do?

This adds a new matcher for partitioned services. For detail see [proposal](https://github.com/containous/traefik/issues/3224)

### Motivation

Allow services to partition/shard load easily. For detail see [proposal](https://github.com/containous/traefik/issues/3224)

### More

- [x] Added/updated tests
- [ ] Added/updated documentation (Waiting on review of approach before adding)

### Additional Notes

I'm not happy with the code which pulls out the 'range', 'type' etc from the rule. Currently it sits in the `hashedRange` function but I think it may be better suited to the `parseRule` function. Currently if the rule is missing required details it returns matcher which is always `false`. I think in `parseRule` I could do something more sensible. Likewise we could change the format of the rule to be less verbose which could simplify the code.

### DIFF
--- a/rules/rules.go
+++ b/rules/rules.go
@@ -126,10 +126,7 @@ func (r *Rules) hashedRange(rule ...string) *mux.Route {
 		hash := hash(value)
 
 		factor := float64(rangeHigh-rangeLow) / float64(maxUint)
-		log.Error(factor)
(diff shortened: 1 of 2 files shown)


ldez:
https://github.com/containous/traefik/issues/3224#issuecomment-393907738

PR #3244, opened 2018-04-27
Allow log rotation in Windows

fcatae:
### What does this PR do?

This implementation contains a proposal for Log Rotation in Windows. 

### Motivation

The current work done in #1761 allows the user to signal the process to rotate the log. However, it is well known that Windows has no support for unix signals.

Here we implemented the tradicional approach of watching WinObjects. We create a global Event object, and keep watching the signaling.

### More

- [ ] Pending: rotate access log
- [ ] Pending: tests
- [ ] Pending: documentation
- [ ] Pending: client example

### Additional Notes

Unfortunately, the current RotateFile function fails with "file already in use" in Windows. Because of that, we had to adopt another approach for log rotation. In our case, we chose generating new filenames. 
I explored other options available in Windows such as MoveFileEx, ReplaceFile and also using Hard Links. However, all options have problems for not being atomic operations. There is a long discussion around this theme in Internet, so I opted for the simplest one.
- https://github.com/golang/go/issues/8914
- https://blogs.msdn.microsoft.com/adioltean/2005/12/28/how-to-do-atomic-writes-in-a-file/
- https://msdn.microsoft.com/en-us/library/hh802690(v=vs.85).aspx

//cc: @eladiw @jjcollinge @lawrencegripper

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -218,6 +218,61 @@ func CloseFile() error {
 	return nil
 }
 
+// ReOpenFile re-opens the log file using the specified path
+// It looks similar to Close/Open, except that there is no temporary
+// redirection to StdOutput
+func ReOpenFile(path string) error {
+	var oldLogFile = logFile
+	newLogFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
+
+	if err != nil {
+		return err
+	}
+
+	logrus.SetOutput(newLogFile)
+
+	if oldLogFile != nil {
+		return oldLogFile.Close()
+	}
+
+	logFilePath = path
+	logFile = newLogFile
+
+	return nil
+}
+
+var originalFilePath = ""
+
+func generateNewFileName(filename string, rotationNumber int) string {
+	return fmt.Sprintf("%s.%d", filename, rotationNumber)
+}
+
+// RotateNewFile opens a new file
+func RotateNewFile(rotationNumber int) error {
+	var oldFilePath = logFilePath
+
+	if rotationNumber == 1 {
+		originalFilePath = logFilePath
+	}
+
+	var newFilePath = generateNewFileName(originalFilePath, rotationNumber+1)
+	ReOpenFile(newFilePath)
+
+	// special case for first file: rename the original filename
+	if rotationNumber == 1 {
+		filePath01 := generateNewFileName(logFilePath, 1)
+		os.Rename(oldFilePath, filePath01)
+
+		Infoln("Rename:", oldFilePath, "->", filePath01)
+
+		oldFilePath = filePath01
+	}
+
+	Infoln("RotateNewFile:", oldFilePath, "->", newFilePath)
+
+	return nil
+}
+
 // RotateFile closes and reopens the log file to allow for rotation
 // by an external source.  If the log isn't backed by a file then
 // it does nothing.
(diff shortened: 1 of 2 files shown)


ldez:
@fcatae Thanks for your work.

Your PR seems to be in WIP, could give us the state for this PR?

fcatae:
Hello @ldez, 

I believe the PR is feature complete, but I decided to wait for someone to review the design decisions before proceeding to finish the work. This PR addresses 2 important issues in Windows:
1. Log rotation - In Windows, you can't create a file before you close it. 
2. No Unix Signals or similar - In Windows, there is nothing similar to Unix signals. 

It works!

We provided the following solutions:
1. File renaming to prevent this issue. It starts the server using <error>.log as it is the current behavior. When the recycle starts, it renames the <error>.log to <error>.1, and the following files become <error>.2, <error>.3, ... This seemed to be the most robust mechanism in case of unexpected server restarts. Other approaches were: close/open the file (but it suffers from data loss), create the first file as 001 (instead of log) and then follow the sequence (it could break existing software that already points to the current extension .log)
2. We implemented kernel object signaling. Other possible approaches were using Windows message (however, we need to create a message loop), using RPC (too much), using timers (user has no control when the file is recycled). We decided to use Kernel Object signaling, which resembles SIGHUB.

Finally, there is no KILL <signal> <pid> in Windows. We also need to provide a client to trigger the recycle or integrate with the Traefik command line (--recyclelog). We could provide a file size limit to automatically recycle the log (that is a common approach in Windows, but then we need a place to host this configuration and maybe ensure it works in Linux too). I would go with option 1, and implement an executable traefik-recycle-log.exe that would be specific for Windows.

If you feel ok with the proposal, I can go ahead and finish 1) the port to Access Log, 2) create the tests, 3) add documentation, 4) create a standalone client app (for Windows) and show examples. Feel free to disagree with the ideas, and plan the solution.

ldez:
Hello @fcatae,

We've discussed about the topic, and our conclusions are:
- we want to keep the current behavior on Unix platform (file naming).
- you can add an option for the file size limit for all platform.
- the Windows behavior ("reverse" file naming) must be implemented in a dedicated go file and documented.

fcatae:
Agree. I will continue the work on this PR, and then I will ask for another review. Thanks!

dtomcej:
We have had a feature request for similar log rotation capabilities in linux/container environments.

https://github.com/containous/traefik/issues/3489

Perhaps we could get two birds with one stone.

ldez:
@fcatae The code freeze for the next 1.7 release will happen at the end of next week, on Friday 6th July 2018.

emilevauge:
@fcatae any news? Do you still want to work on this one?

fcatae:
@emilevauge @ldez Sorry for the delay. After being warned about #3489, I decided not to move forward before the log rotation solution. This PR includes both the "SIGHUB emulation" in Windows platform AND the log rotation. I am not really sure how to proceed here - I will probably close this PR, and resubmit when things get clear. What are your thoughts?

dduportal:
Hello @fcatae ! 

We think that your assumption is right: could you please close this PR, and open a new one only focused onto the "SIGHUB emulation" please?

About the log rotation, let's wait for #3489 to have a direction, before jumping into this topic for Windows' environments.

As a side note for your information, are you aware that the first Docker image for Windows based on `nanoserver-sac2016` is on the ramp (cf. #3603) ? It might help (or might not :) )

fcatae:
@dduportal thank your for the guidance! I am closing this PR.

PR #3261, opened 2018-05-03
Changed "is could" to should

blakethepatton:
### What does this PR do?
Changes `is could` to `should` in the documentation

### Motivation
It bothered me, I had to read the sentence twice because it was phrased weird.


docs/index.md
```diff
@@ -12,7 +12,7 @@
 
 Træfik is a modern HTTP reverse proxy and load balancer that makes deploying microservices easy.
 Træfik integrates with your existing infrastructure components ([Docker](https://www.docker.com/), [Swarm mode](https://docs.docker.com/engine/swarm/), [Kubernetes](https://kubernetes.io), [Marathon](https://mesosphere.github.io/marathon/), [Consul](https://www.consul.io/), [Etcd](https://coreos.com/etcd/), [Rancher](https://rancher.com), [Amazon ECS](https://aws.amazon.com/ecs), ...) and configures itself automatically and dynamically.
-Telling Træfik where your orchestrator is could be the _only_ configuration step you need to do.
+Telling Træfik where your orchestrator should be the _only_ configuration step you need to do.
```
dtomcej: Not sure I agree with this change. How about:
`Pointing Træfik at your orchestrator should be the _only_ configuration step you need.`?

geraldcroes: We'll also have to update the README.md file accordingly.

PR #3274, opened 2018-05-06
Add basicAuth example for KV

MichaelErmer:
### What does this PR do?

added basicAuth to the example

### DIFF
--- a/docs/user-guide/kv-config.md
+++ b/docs/user-guide/kv-config.md
@@ -329,14 +329,15 @@ And there, the same dynamic configuration in a KV Store (using `prefix = "traefi
 
 - frontend 2
 
-| Key                                                | Value              |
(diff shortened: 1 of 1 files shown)


emilevauge:
Hey @MichaelErmer, thank you for your contribution !
We think that our user guide should stay as simple as possible and thus, it appears that basic authentication may not appear in that section. I will close this PR then. Thanks again :)

MichaelErmer:
@emilevauge But its the only place where the K/V Storage is documented at all, all other pages are linking to that page, eg.: `Please refer to the Key Value storage structure section to get documentation on Traefik KV structure.`https://docs.traefik.io/configuration/backends/consul/

emilevauge:
Indeed @MichaelErmer, you are right. So for being a bit too fast on this one...

PR #3286, opened 2018-05-08
sane default and configurable Marathon request timeouts

marco-jantke:
### What does this PR do?

Set sane timeouts for requests to Marathon and make them configurable.

### Motivation

Before the timeouts were either overly excessive or not existent at all. For requests done by go-marathon there was no timeout, which means when there was temporarily a network problem which doesn't immediately show a failure (e.g. when the packets are dropped by a firewall) Traefik will never be able to create a connection to Marathon, as this one request will be running basically forever. I used the same timeouts as go-marathon does by default. We have to do this in Traefik as we pass in our own http.Client.

### More

- [x]  Added/updated tests: -> no adaption of tests required for this PR

### DIFF
--- a/provider/marathon/marathon.go
+++ b/provider/marathon/marathon.go
@@ -55,7 +55,9 @@ type Provider struct {
 	MarathonLBCompatibility   bool             `description:"Add compatibility with marathon-lb labels" export:"true"`
 	FilterMarathonConstraints bool             `description:"Enable use of Marathon constraints in constraint filtering" export:"true"`
 	TLS                       *types.ClientTLS `description:"Enable TLS support" export:"true"`
-	DialerTimeout             flaeg.Duration   `description:"Set a non-default connection timeout for Marathon" export:"true"`
+	DialerTimeout             flaeg.Duration   `description:"Set a non-default dialer timeout for Marathon" export:"true"`
+	ResponseHeaderTimeout     flaeg.Duration   `description:"Set a non-default response header timeout for Marathon" export:"true"`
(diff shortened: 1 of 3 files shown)


marco-jantke:
1. I think that in a large cluster it could actually happen that Marathon will take longer than 10 seconds to send the first bytes in it's response. Especially, when the server is under a high load at this point in time. In order to be able to serve all users of Traefik I think we should make the timeouts configurable and aim for a high default value like ~ 60 seconds for the `ResponseHeaderTimeout`. For the `TLSHandshake` I would stick with 5 seoncds as default and also make it configurable. Sounds good to you?
1. I also think we should stick to master.

timoreimann:
Sounds reasonable on both fronts. 👍

marco-jantke:
I updated the PR accordingly. PTAL.

provider/marathon/marathon.go
```diff
@@ -55,7 +55,9 @@ type Provider struct {
 	MarathonLBCompatibility   bool             `description:"Add compatibility with marathon-lb labels" export:"true"`
 	FilterMarathonConstraints bool             `description:"Enable use of Marathon constraints in constraint filtering" export:"true"`
 	TLS                       *types.ClientTLS `description:"Enable TLS support" export:"true"`
-	DialerTimeout             flaeg.Duration   `description:"Set a non-default connection timeout for Marathon" export:"true"`
+	DialerTimeout             flaeg.Duration   `description:"Set a non-default dialer timeout for Marathon" export:"true"`
```
dtomcej: Do we need to use the phrase "non-default"? Isn't it kind of assumed that defaults will be used unless specified?

marco-jantke: Good point, thanks! I removed the "non-default".

PR #3287, opened 2018-05-08
Improve Prometheus metrics removal

marco-jantke:
### What does this PR do?

Improve Prometheus metrics removal.

### Motivation

Formerly metric removal was done by tracking configuration generations. This was an easy first solution but is actually not correct and for setups where Traefik does not receive a lot of configuration updates, old metrics can hang around for quite some time.

The new approach is to keep track of Traefik's current configuration and to remove metrics that "belong" to an outdated configuration, once they were scraped. This way we can be sure that we don't keep old metrics too long and we can ensure that metrics for short-lived configuration still appear when the /metrics is scraped, as we only remove them after the scrape happened.

### More

- [x] Added/updated tests

### DIFF
--- a/metrics/prometheus.go
+++ b/metrics/prometheus.go
@@ -248,6 +248,7 @@ func (ps *prometheusState) Collect(ch chan<- stdprometheus.Metric) {
 	}
 
 	for _, key := range outdatedKeys {
+		ps.state[key].delete()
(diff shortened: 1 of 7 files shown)


marco-jantke:
FTR: I missed a local optimization and will update the PR soon again.

PR #3297, opened 2018-05-09
webui: display health durations with a 3 digits precision

mvertes:
### What does this PR do?

webui: display health durations with a 3 digits precision

### Motivation
So instead of `10h50m42.15673436371s`, display `10h50m42.156s`
It is applied to "total response time" and "average response time".

### More

- [ ] Added/updated tests


ldez:
I'm working on a PR about WebUI, and I have already fixed that point and lot more things.

emilevauge:
@mvertes won :yum:

Issue #3282, opened 2018-05-07
Custom path not usable for Dashboard


dpvDuncan:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I wanted to have the dashboard exposed on my default https entrypoint, with a custom path, as explained at "https://docs.traefik.io/configuration/api/", si I created a new .toml file as following:
```toml
[backends]
  [backends.traefik]
    [backends.traefik.servers.traefik]
      url = "http://localhost:8080"

[frontends]
  [frontends.traefikAPI]
    backend = "traefik"
    [frontends.traefikAPI.routes.traefikAPI]
      rule = "PathPrefixStrip:/traefik"
```
### What did you expect to see?

With the previous WebUI, I had no problems (up to 1.6.0-RC6), and I did not expect any by upgrading to 1.6.0.
I could see my dashboard, with the proviedres at https://host.mydomain.com/traefik/dashboard/

### What did you see instead?

With 1.6.0, https://host.mydomain.com/traefik/dashboard/ shows "No providers found.", and spawns http 404 errors because it requests '/api/providers' which is routed to another project.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.6.0
Codename:     tetedemoine
Go version:   go1.10.1
Built:        2018-04-30_09:28:44PM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

<details>
<summary>config</summary>

```json
{
 "LifeCycle": {
  "RequestAcceptGraceTimeout": 0,
  "GraceTimeOut": 0
 },
 "GraceTimeOut": 0,
 "Debug": false,
 "CheckNewVersion": true,
 "SendAnonymousUsage": false,
 "AccessLogsFile": "/var/log/traefik/access.log",
 "AccessLog": null,
 "TraefikLogsFile": "/var/log/traefik/traefik.log",
 "TraefikLog": null,
 "Tracing": null,
 "LogLevel": "INFO",
 "EntryPoints": {
  "http": {
   "Address": "xxxx",
   "TLS": null,
   "Redirect": {
    "entryPoint": "xxxx"
   },
   "Auth": null,
   "WhitelistSourceRange": null,
   "WhiteList": null,
   "Compress": false,
   "ProxyProtocol": null,
   "ForwardedHeaders": null
  },
  "https": {
   "Address": "xxxx",
   "TLS": {
    "MinVersion": "",
    "CipherSuites": null,
[...]

PR #3299, opened 2018-05-09
Fix webui

### LINKED ISSUE #3258: UX issues in the new dashboard

bug

### What did you do?

While we're [waiting for the official 1.6 docker image](https://github.com/docker-library/official-images/pull/4304), I downloaded a shiny new traefik 1.6 from `containous/traefik@sha256:74c7e144161a850b648e4a690ad6940c1ef152c7797787e56673c6352846e51c` (https://github.com/containous/traefik-library-image/issues/29#issuecomment-385984440) and started playing with the fresh dashboard. Here's a CSS glitch that I'm observing in Firefox:

screen shot 2018-05-02 at 18 41 54

Context:

screen shot 2018-05-02 at 18 42 51

The same divs in Chrome look a bit different, but styling is still glitchy:

kapture 2018-05-02 at 18 48 58

ldez:
### What does this PR do?

- [x] Health page
  - [x] reduce CPU usage
  - [x] Average Response Time (graph)
    - [x] speedup loading
    - [x] remove some animation
    - [x] better resizing support
  - [x] Total Status Code Count (graph)
    - [x] speedup loading
    - [x] add colors
  - [x] Total Response Time (value): compact display
  - [x] Average Response Time (value): compact display
  - [x] Recent errors
    - [x] remove request link
    - [x] smooth refresh
    - [x] human readable date
- [x] Global
  - [x] navbar on small screen
  - [x] API and custom path
  - [x] replace px by em or rem
  - [x] clean sass files
  - [x] relative fonts path
- [x] Providers
  - [x] limit frontends and backends
  - [x] search filter
    - [x] clean filter
    - [x] consistent behavior
  - [x] frontend name & backend name
  - [x] backend name in frontend
  - [x] table overlap
  - [x] frontend without backend -> turn to red
  - [x] backend without servers -> turn to red

### Motivation

Fixes #3258, #3282

### Additional Notes

<details>
<summary>Provider page</summary>

dashboard-providers-fix

</details>

<details>
<summary>Health page</summary>

dashboard-health-fix

</details>


kachkaev:
Many thanks for putting this together @ldez! 🎉 I'd be keen to try it out, but I'm not sure if I can quickly spin up traefik without Docker. Is an image automatically ‘backed’ by CI by a chance?

Two questions:

* In addition to removing time precision as suggested in https://github.com/containous/traefik/pull/3297, how about making time in the error log a bit more humane? Or is it not worth it?

* I see that two ticks in https://github.com/containous/traefik/issues/3258 are still off. They say:
    > * [ ] Going to ‘Health’ tab (e.g. http://localhost:8080/dashboard/status) and then refreshing the page strips `status` from the URI and brings back to the main dashboard page.

    > * [ ] I'd also use `white-space: nowrap` css in _Status_ and even possibly _Time_ column. That's so that their rather short values did not look torn on narrow screens. URLs could either have `overflow:hidden` or use more than one line if they don't fit the rest of the table width.

   Are you planning to leave these things as is for now or are the checkboxes simply out of sync?

The screenshots look awesome! I especially like your choice of colours for HTTP status codes! 💯

ldez:
@kachkaev I will answer you into the issue.

ldez:
@kachkaev I will create a Docker image for you.

ldez:
@kachkaev you can test with `ldez/traefik:webui` Docker image

kachkaev:
Many thanks for sharing the image @ldez!

Looks like something is broken because the UI is unable to fetch any data from the API:

screen shot 2018-05-09 at 18 03 32

Requests like `http://localhost:8080/dashboard/api/providers` return 404. Image is `sha256:2c1e71d965063f13ac332f40cbf165c5396e354f26073f52e8cdb97df18955a6`, which should be the one you've shared (it's up-to-date when I `docker pull`).

Because of this glitch I could only check the menu and horizontal padding of the health blocks – that looks good! The only minor issue I could find is that the third row in the burger menu is clickable and takes a user to `http://localhost:8080/dashboard/null`:

screen shot 2018-05-09 at 18 08 31

---
There was also one unrelated issue that I had initially, it is to do with `--metrics.statistics.recenterrors` that I set after migrating to 1.6.0. This flag caused `Error parsing command: unknown flag: --metrics.statistics.recenterrors` and I had to check out the new docs and change it to `--api.statistics.recenterrors`.

ldez:
> The only minor issue I could find is that the third row in the burger menu is clickable and takes a user to ...

Normally, this menu item display the Traefik version, this problem is related to the issue with the API call. 

Could you retry to pull the image (I updated the tag)?

Ch1ch1:
> Could you retry to pull the image (I updated the tag)?

Now working like a charm with ldez/traefik:webui

kachkaev:
Many thanks for updating the image @ldez! The UX is smoother now 🚀  Here are a few observations, which may improve the dashboard in future, but all except the first one are not critical IMO:

* [ ] URI protocol in error log shows up as `unsafe::`, which  generates links like `unsafe:://localhost/favicon.ico`:

    screen shot 2018-05-09 at 19 26 09

   I can share my launch options if that matters.

* [ ] Selecting text in the error log does not reset immedeately, however, hitting a new error page does break the current text selection. That would not be an issue for me with 0.5 requests per day 😅, however, more advanced users may find this side-effect inconvenient. I guess that Angular is not reusing list items and simply flushes the entire list on an update. Does there exist some Node ID concept in Angular that could help reuse DOM nodes? This would be solvable with `<Item key={item.id || `${item.status}-${item.uri}-${item.timestamp}`} data={item} />` in React.

* [ ] When the health summary is shown on a mobile screen, vertical separators remain visible and make a bit of noise:

    screen shot 2018-05-09 at 19 05 53

* [ ] In _Uptime Since_ the format is `05/09/2018 18:56:41`, where month precedes day. I know this is a common in States, but would not it be more helpful to see `2018-05-09` (i.e. `YYYY-MM-DD`)? This is more international and thus reduces a change of interpreting, say, Jan 2 as Feb 1. Also, which time zone is this? UTC? Server time? My local time? The UI does not give an answer.

* [ ] It'd be great if the burger was a bit less greedy and did not hide the entire menu at ≈1080px of page width. This creates quite a lot of unused empty space near the logo where at least `providers` and `health` could fit. This would save laptop and tablet users a mouse move and a click millions of times.

* [ ] This one is purely a matter of taste: the boxes near the number of frontends or backends might look better aligned if moved to the right a bit. This is especially desired on a small screen where the ‘islands’ stick to the page border:

    screen shot 2018-05-09 at 19 07 38  

    ↓  

    screen shot 2018-05-09 at 19 09 00

* [ ] Very related to previous: what is the rationale for aligning frontend and backend headers to the right? This creates a gab between an icon and a text, which weakens their semantic relationship and also makes it a bit harder for a human eye to scan through the names (this is a typical issue for an alphabetically ordered lists that's alight to right):

   screen shot 2018-05-09 at 19 09 50

    How about using `align: left` in in card headers?

* [ ]  The response time chart still running a quite funny animation after switching from `health` to `dashboard` and back (not a big deal).

* [ ] Time and status code in error log still uses quite a lot of space. From https://github.com/containous/traefik/issues/3258#issuecomment-387792432:

    > It can be a human-readable time in UTC or a local time zone (say, with a minute precision) with a more technically looking time coordinate shown when the value is hovered. A simple and quite a common way of implementing this is just assigning `title="TTTT"` to a span. Could this trick be applied to status codes too? I.e. the first column would say `404` or `500` (coloured the same way as the chart above) and hovering over a box with the number would reveal `title="Not Found"`. That should leave more space for the URL while not being a problem for an average user since the dashboard is targeting sysadmins.

rroblik:
Yeah nice fix ! Working now also for me on `https://docker.mydomain.com/proxy` with `yml` file like that (if anybody have the same need)

```yaml
version: "3.5"

services:
  traefik:
    #image: traefik:1.6.0
    image: ldez/traefik:webui
    command:
      --docker
      --docker.swarmmode
      --docker.watch
      --logLevel=DEBUG
    ports:
      - target: 80
        published: 80
        protocol: tcp
        mode: host
      - target: 443
        published: 443
        protocol: tcp
        mode: host
    volumes:
      - ./configs/traefik.toml:/traefik.toml
      - ./rules/:/etc/traefik/rules/
      - ./_certs/:/ssl/
    networks:
      - webgateway
    deploy:
      labels:
        traefik.enable: "true"
        traefik.docker.network: "webgateway"
        traefik.backend: "proxy"
        traefik.port: "8080"
        traefik.frontend.rule: "Host:docker.mydomain.com;PathPrefixStrip:/proxy;PathPrefix:/proxy"
        traefik.weight: 100
      replicas: 1
      placement:
        constraints: 
          - node.role == manager
      #replicas: 3
      restart_policy:
        condition: any
    
networks:
  webgateway:
    external: true
```

Any idea for a release date in "official" traefik image ?

PR #3312, opened 2018-05-11
Fix backend reuse

### LINKED ISSUE #2323: Some frontend configuration is ignored when re-using same backend
Bug

### What did you do?
When two frontends are using the same backend, only the first processed frontend is able to set settings like basicAuth or whitelistSourceRange.

I would like to use multiple frontends to apply different behaviour based on the Path of the URI.

A simple test scenario:
Two frontends with a different PathPrefix but using the same backend

**frontend website-a:**
- priority: 10
- backend: app
- rule: "PathPrefix: /secure"
- basicAuth: ["aaa:aaa"]

**frontend website-b:**
- priority: 5
- backend: app
- rule: "PathPrefix: /hello"

**backend app:**
 - url: "http://10.10.10.10:32779/"

A short investigation in Traefik's code reveals the following:

  server/server.go:
```go
    if backends[entryPointName+frontend.Backend] == nil {
      ...
      Configure IP Whitelist
      Configure Basic Auth
      etc.. etc..
      ...
    } else {
      log.Debugf("Reusing backend %s", frontend.Backend)
    }
```
It seems frontend specific settings are skipped because the backend is already configured.

When I duplicate the backend configuration to app2 and configure app2 as a backend on website-b the configuration works as expected.

In case of basic authentication defined on two frontends using the same backend:

```console
$ http localhost/secure
HTTP/1.1 401 Unauthorized
 ```
```console
$ http localhost/hello
HTTP/1.1 200 OK
```

### What did you see instead?
In case of basic authentication defined on two frontends using the same backend:

```console
$ http localhost/secure
HTTP/1.1 401 Unauthorized
```
```console
$ http localhost/hello
HTTP/1.1 401 Unauthorized
```

Traefik 1.4.0

```toml
    [file]
    [frontends]
      [frontends.website-a]
      priority = 10
      backend = "app"
      basicAuth = ["aaa:aaa"]
    
        [frontends.website-a.routes.a]
          rule = "PathPrefix: /secure"

      [frontends.website-b]
      priority = 5
      backend = "app"
    
      [frontends.website-b.routes.b]
        rule = "PathPrefix: /hello"
    
    [backends]
      [backends.app]
        [backends.app.servers.app]
        url = "http://10.10.10.10:32779/"
```

arnested:
### What does this PR do?

This fixes the way backends are currently cached leading t some frontend configuration being ignored when re-using same backend (reported in #2323).

### Motivation

Not being able to reuse backends in different frontends is unexpected and the workaround results in defining additional, identical backends.

In a current project of mine I could potentially end up with 400 identical backends to work around this.

Fixes  #2323

### More

- [x] Added/updated tests
- [x] Added/updated documentation: No, this fixes a bug in internal behaviour

### Additional Notes

Currently backends are cached under a key generated by `entryPointName+providerName+frontend.Backend`. The `frontend.Backend` doesn't respect the differences present in frontends.

I add `Hash()` method to the `Frontend` struct and use that instead of just the backend name.

The Hash method uses [github.com/mitchellh/hashstructure](https://github.com/mitchellh/hashstructure) which is also used in [collector/collector.go](/containous/traefik/blob/master/collector/collector.go) (a nice feature of it is being able to ignore some fields when computing the hash).

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -950,7 +950,7 @@ func (s *Server) loadConfig(configurations types.Configurations, globalConfigura
 						redirectHandlers[entryPointName] = handlerToUse
 					}
 				}
-				if backends[entryPointName+providerName+frontend.Backend] == nil {
+				if backends[entryPointName+providerName+frontend.Hash()] == nil {
 					log.Debugf("Creating backend %s", frontend.Backend)
 
 					roundTripper, err := s.getRoundTripper(entryPointName, globalConfiguration, frontend.PassTLSCert, entryPoint.TLS)
@@ -1208,14 +1208,14 @@ func (s *Server) loadConfig(configurations types.Configurations, globalConfigura
 					} else {
 						n.UseHandler(lb)
 					}
-					backends[entryPointName+providerName+frontend.Backend] = n
+					backends[entryPointName+providerName+frontend.Hash()] = n
 				} else {
 					log.Debugf("Reusing backend %s", frontend.Backend)
 				}
 				if frontend.Priority > 0 {
 					newServerRoute.Route.Priority(frontend.Priority)
 				}
-				s.wireFrontendBackend(newServerRoute, backends[entryPointName+providerName+frontend.Backend])
+				s.wireFrontendBackend(newServerRoute, backends[entryPointName+providerName+frontend.Hash()])
 
 				err := newServerRoute.Route.GetError()
 				if err != nil {
--- a/server/server.go
+++ b/server/server.go
@@ -950,7 +950,14 @@ func (s *Server) loadConfig(configurations types.Configurations, globalConfigura
 						redirectHandlers[entryPointName] = handlerToUse
 					}
 				}
-				backendCacheKey := entryPointName + providerName + frontend.Hash()
+
+				frontendHash, err := frontend.Hash()
+				if err != nil {
+					log.Errorf("Error calculating hash value for frontend %s: %v", frontendName, err)
+					log.Errorf("Skipping frontend %s...", frontendName)
+					continue frontend
+				}
+				backendCacheKey := entryPointName + providerName + frontendHash
 				if backends[backendCacheKey] == nil {
 					log.Debugf("Creating backend %s", frontend.Backend)
 
@@ -1218,7 +1225,7 @@ func (s *Server) loadConfig(configurations types.Configurations, globalConfigura
 				}
 				s.wireFrontendBackend(newServerRoute, backends[backendCacheKey])
 
-				err := newServerRoute.Route.GetError()
+				err = newServerRoute.Route.GetError()
 				if err != nil {
 					log.Errorf("Error building route: %s", err)
 				}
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -925,7 +925,7 @@ func (s *Server) loadConfig(configurations types.Configurations, globalConfigura
 						redirectHandlers[entryPointName] = handlerToUse
 					}
 				}
-				if backends[entryPointName+providerName+frontend.Backend] == nil {
+				if backends[entryPointName+providerName+frontend.Hash()] == nil {
```
dtomcej: Could this string be stored, and referenced, instead of being recalculated multiple times?

types/types.go
```diff
@@ -188,6 +189,13 @@ type Frontend struct {
 	Redirect             *Redirect             `json:"redirect,omitempty"`
 }
 
+// Hash returns the hash value of a Frontend struct.
+func (f *Frontend) Hash() string {
+	hash, _ := hashstructure.Hash(f, nil)
```
dtomcej: Should handle the `error` here. If the hash goes sideways, you are going to want to log or potentially return an error.

server/server.go
arnested: Yes, that should be doable. I'll just need to make sure the frontend isn't changed between usages of a store string.

types/types.go
arnested: Agree. I think I took a shortcut because from te definition of the `Hash()` it is not able to produce an error with the current definition of the `Frontend` struct. But that's not very kind on the people doing changes in the future.

I'll create fixes for this (today or tomorrow probably).

server/server.go
dtomcej: I was just thinking inside the loop. Instead of calculating `.Hash()` three times.
`entryPointName+providerName+frontend.Hash()` in your PR.

arnested: I hadn't checked if shady things where done to the frontend in the loop (although that would also have been a bug if it did).

Implemented in b716cb2b3c892a678d9cac77954dea11ae9439bc.

types/types.go
arnested: Implemented in d44fde676f86f2d039b2b8e4798f6b25a5d98e3a.

arnested:
Thank you for the approve, @dtomcej.

GitHub has detected a conflict in `server/server_test.go` because the tests have been refactored into using testhelpers.

I have refactored my test method to use the new test helpers as well.

Fun story. The refactored test helpers have a bug in [`WithFrontend()`](https://github.com/containous/traefik/blob/master/testhelpers/config.go#L89): it will cache the frontends under the backend name (the fun part being this is the same kind og bug as I'm trying to fix in this pull request :smile:).

I had to rebase on master to not make everything completely unreadable. I introduced a new commit (cc69be54446b021ed3bdbbf11d770497d7fa70ae) that fixes the bug in the test helper. Next commit (fb9ed685b943d918e6b1e78659061b0ca88282d2) is my original test case refactored to use the new test helpers. The rest of the commits are as you reviewed them.

PR #3324, opened 2018-05-14
Enable Ingress Status updates

### LINKED ISSUE #2173: Feature Request: Update Kubernetes Ingress Status
Currently (as of `v1.4.0-rc3`) traefik does not update the ingress [`status.loadBalancer`](https://kubernetes.io/docs/api-reference/v1.7/#loadbalancerstatus-v1-core) field to anything. This makes it difficult to auto-assign DNS (such as  [kubernetes external-dns](https://github.com/kubernetes-incubator/external-dns)). Other Kubernetes ingress controllers update this field (see the `update-status` flag on the [nginx controller](https://github.com/kubernetes/ingress/tree/master/controllers/nginx#command-line-arguments))

The Kuberentes [`LoadBalancerIngress`](https://kubernetes.io/docs/api-reference/v1.7/#loadbalanceringress-v1-core) displays either a `hostname` or an `ip` field consisting of a single string value. It would be fantastic to have a Configuration option to specify the value to update managed ingress rules with. A user-configurable flag like `--kubernetes.ingressEndpoint.ip`  or `--kubernetes.ingressEndpoint.hostname`.

**Feature**

### What did you do?

**my-ingress.yaml**
```yaml
kind: Ingress
metadata:
  name: my-ingress
  namespace: default
spec:
  rules:
  - host: my-service.example.com
    http:
      paths:
      - backend:
          serviceName: my-service
          servicePort: 3000
        path: /
```

```bash
$ kubectl apply -f my-ingress.yaml
```

```
$ kubectl get -f my-ingress.yaml -o yaml
```
```yaml
kind: Ingress
metadata:
  name: my-ingress
  namespace: default
spec:
  rules:
  - host: my-service.example.com
    http:
      paths:
      - backend:
          serviceName: my-service
          servicePort: 3000
        path: /
status:
  loadBalancer:
    ingress:
    - hostname: my-elb-hostname.us-east-1.elb.amazonaws.com
```

### What did you see instead?

```
$ kubectl get -f my-ingress.yaml -o yaml
```
```yaml
kind: Ingress
metadata:
  name: my-ingress
  namespace: default
spec:
  rules:
  - host: my-service.example.com
    http:
      paths:
      - backend:
          serviceName: my-service
          servicePort: 3000
        path: /
status:
  loadBalancer: {}
```

```
$ docker run -it --rm traefik:v1.4.0-rc3 version
Version:      v1.4.0-rc3
Codename:     roquefort
Go version:   go1.9
Built:        2017-09-18_04:38:27PM
OS/Arch:      linux/amd64
```

Deployed in Kubernetes via daemonset
```toml
# No config file used
```

```yaml
kind: DaemonSet
metadata:
  name: traefik-858bd5f3
  namespace: kube-system
spec:
  selector:
    matchLabels:
      app: traefik-ingress-controller
  template:
    metadata:
      labels:
        app: traefik-ingress-controller
    spec:
      containers:
      - args:
        - --web
        - --kubernetes
        image: traefik:latest
        name: traefik-ingress-controller
        ports:
        - containerPort: 80
          protocol: TCP
        - containerPort: 8080
          protocol: TCP
      nodeSelector:
        node-role.kubernetes.io/worker: ""
```

```
# N/A
```

dtomcej:
### What does this PR do?

Allows Traefik to update the Status section of Ingress objects either via static IP/Hostname configuration, or via extracting the status section of a fronting service.

### Motivation

Fixes #2173

Users requested that the status field be set so that DNS and other discovery tools would work properly.

### More

- [x] Added/updated tests

### Additional Notes

This configuration is entirely optional, and will only apply if one of the 3 configuration settings are enabled.

### DIFF
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -90,11 +90,11 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # [kubernetes.ingressEndpoint]
 #
 # One must be configured.
-# Publishedservice will override the hostname and ip settings if configured.
+# `publishedservice` will override the `hostname` and `ip` settings if configured.
 #
 # hostname = "localhost"
 # ip = "127.0.0.1"
-# publishedService = "default/cheddar"
+# publishedService = "namespace/servicename"
 ```
 
 ### `endpoint`
@@ -121,7 +121,7 @@ See [label-selectors](https://kubernetes.io/docs/concepts/overview/working-with-
 
 ### `ingressEndpoint`
 
-You can configure a static hostname or IP address that traefik will add to the status section of ingress objects that it manages.
+You can configure a static hostname or IP address that Traefik will add to the status section of Ingress objects that it manages.
 If you prefer, you can provide a service, which traefik will copy the status spec from.
 This will give more flexibility in cloud/dynamic environments.
 
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -81,6 +81,20 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # Default: <built-in template>
 #
 # filename = "kubernetes.tmpl"
+
+# Enable IngressEndpoint configuration.
+# This will allow Traefik to update the status section of ingress objects, if desired.
+#
+# Optional
+#
+# [kubernetes.ingressEndpoint]
+#
+# One must be configured.
+# Publishedservice will override the hostname and ip settings if configured.
+#
+# hostname = "localhost"
+# ip = "127.0.0.1"
+# publishedService = "default/cheddar"
 ```
 
 ### `endpoint`
@@ -105,23 +119,7 @@ A label selector can be defined to filter on specific Ingress objects only.
 
 See [label-selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) for details.
 
-```toml
-# Enable IngressEndpoint configuration.
-# This will allow Traefik to update the status section of ingress objects, if desired.
-#
-# Optional
-#
-[kubernetes.ingressEndpoint]
-
-# One must be configured.
-# Publishedservice will override the hostname and ip settings if configured.
-
-# hostname = localhost
-
-# ip = 127.0.0.1
-
-# publishedservice = default/cheddar
-```
+### `ingressEndpoint`
 
 You can configure a static hostname or IP address that traefik will add to the status section of ingress objects that it manages.
 If you prefer, you can provide a service, which traefik will copy the status spec from.
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -105,6 +105,28 @@ A label selector can be defined to filter on specific Ingress objects only.
 
 See [label-selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) for details.
 
+```toml
+# Enable IngressEndpoint configuration.
+# This will allow Traefik to update the status section of ingress objects, if desired.
+#
+# Optional
+#
+[kubernetes.ingressEndpoint]
+
+# One must be configured.
+# Publishedservice will override the hostname and ip settings if configured.
+
+# hostname = localhost
+
+# ip = 127.0.0.1
+
+# publishedservice = default/cheddar
+```
+
+You can configure a static hostname or IP address that traefik will add to the status section of ingress objects that it manages.
+If you prefer, you can provide a service, which traefik will copy the status spec from.
+This will give more flexibility in cloud/dynamic environments.
+
 ### TLS communication between Traefik and backend pods
 
 Traefik automatically requests endpoint information based on the service provided in the ingress spec.
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -89,7 +89,7 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 #
 # [kubernetes.ingressEndpoint]
 #
-# One must be configured.
+# At least one must be configured.
 # `publishedservice` will override the `hostname` and `ip` settings if configured.
 #
 # hostname = "localhost"
--- a/provider/kubernetes/client.go
+++ b/provider/kubernetes/client.go
@@ -169,12 +169,14 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 
 // UpdateIngressStatus updates an Ingress with a provided status.
 func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) error {
-	item, exists, err := c.factories[c.lookupNamespace(namespace)].Extensions().V1beta1().Ingresses().Informer().GetStore().GetByKey(namespace + "/" + name)
+	keyName := namespace + "/" + name
+
+	item, exists, err := c.factories[c.lookupNamespace(namespace)].Extensions().V1beta1().Ingresses().Informer().GetStore().GetByKey(keyName)
 	if err != nil {
-		return fmt.Errorf("failed to get ingress %s/%s with error: %v", namespace, name, err)
+		return fmt.Errorf("failed to get ingress %s with error: %v", keyName, err)
 	}
 	if !exists {
-		return fmt.Errorf("failed to update ingress %s/%s because it does not exist", namespace, name)
+		return fmt.Errorf("failed to update ingress %s because it does not exist", keyName)
 	}
 
 	ing := item.(*extensionsv1beta1.Ingress)
@@ -187,9 +189,9 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
 
 		_, err := c.clientset.ExtensionsV1beta1().Ingresses(ingCopy.Namespace).UpdateStatus(ingCopy)
 		if err != nil {
-			return fmt.Errorf("failed to update ingress status %s/%s with error: %v", namespace, name, err)
+			return fmt.Errorf("failed to update ingress status %s with error: %v", keyName, err)
 		}
-		log.Infof("Updated status on ingress %s/%s", namespace, name)
+		log.Infof("Updated status on ingress %s", keyName)
 	}
 	return nil
 }
--- a/provider/kubernetes/client.go
+++ b/provider/kubernetes/client.go
@@ -44,6 +44,7 @@ type Client interface {
 	GetService(namespace, name string) (*corev1.Service, bool, error)
 	GetSecret(namespace, name string) (*corev1.Secret, bool, error)
 	GetEndpoints(namespace, name string) (*corev1.Endpoints, bool, error)
+	UpdateIngressStatus(namespace, name, ip, hostname string) error
 }
 
 type clientImpl struct {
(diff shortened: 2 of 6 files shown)


provider/kubernetes/client.go
```diff
@@ -166,6 +167,22 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 	return result
 }
 
+//UpdateIngressStatus updates an Ingress with a provided status.
```
ldez: Could you add a space between `//` and the message

provider/kubernetes/client.go
```diff
@@ -166,6 +167,22 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
...
+func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) error {
+	item, _, _ := c.factories[c.lookupNamespace(namespace)].Extensions().V1beta1().Ingresses().Informer().GetStore().GetByKey(namespace + "/" + name)
```
ldez: You need to manage the error.

provider/kubernetes/kubernetes.go
```diff
@@ -129,6 +137,9 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 						if err != nil {
 							return err
 						}
+						if err != nil {
```
ldez: Seems not needed

provider/kubernetes/kubernetes.go
```diff
@@ -155,6 +167,50 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 	return nil
 }
 
+func (p *Provider) updateIngressStatuses(k8sClient Client) error {
+	//Only run these ingress updates if config is provided. This will reduce overhead
```
ldez: Could you add a space between `//` and the message

provider/kubernetes/kubernetes.go
```diff
@@ -155,6 +167,50 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+	if p.IngressEndpoint.Hostname != "" || p.IngressEndpoint.IP != "" || p.IngressEndpoint.PublishedService != "" {
```
ldez: I think you need to check if  `p.IngressEndpoint` is not nil

provider/kubernetes/client.go
```diff
@@ -166,6 +167,22 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
...
+	ing := item.(*extensionsv1beta1.Ingress)
+	if ing.Status.LoadBalancer.Ingress[0].Hostname == hostname && ing.Status.LoadBalancer.Ingress[0].IP == ip {
+		//If status is already set, skip update
```
ldez: Could you add a space between `//` and the message

provider/kubernetes/kubernetes.go
```diff
@@ -138,6 +149,7 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 								Configuration: p.loadConfig(*templateObjects),
 							}
 						}
+						err = p.updateIngressStatuses(k8sClient)
```
ldez: you need to manage the error

provider/kubernetes/kubernetes.go
```diff
@@ -155,6 +167,50 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+					continue
+				}
+				if exists {
+					err := k8sClient.UpdateIngressStatus(i.Namespace, i.Name, svc.Status.LoadBalancer.Ingress[0].IP, svc.Status.LoadBalancer.Ingress[0].Hostname)
+					if err != nil {
+						log.Errorf("Cannot update Ingress %s/%s due to error: %s", i.Namespace, i.Name, err)
+						return err
+					}
+				} else {
+					log.Errorf("Cannot update Ingress %s/%s due to missing service: %s", i.Namespace, i.Name, p.IngressEndpoint.PublishedService)
```
ldez: Maybe return an error in this case?

provider/kubernetes/kubernetes.go
```diff
@@ -155,6 +167,50 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+				//Set the status to the status of the fronting LB service
+				if exists && svc.Status.LoadBalancer.Ingress == nil {
+					//service exists, but has no LoadBalancer status
+					log.Debugf("Skipping updating Ingress %s/%s due to service %s having no status set", i.Namespace, i.Name, p.IngressEndpoint.PublishedService)
+					continue
+				}
+				if exists {
+					err := k8sClient.UpdateIngressStatus(i.Namespace, i.Name, svc.Status.LoadBalancer.Ingress[0].IP, svc.Status.LoadBalancer.Ingress[0].Hostname)
+					if err != nil {
+						log.Errorf("Cannot update Ingress %s/%s due to error: %s", i.Namespace, i.Name, err)
```
ldez: I think the log can be transform to an error.

provider/kubernetes/kubernetes.go
```diff
@@ -155,6 +167,50 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+						return err
+					}
+				} else {
+					log.Errorf("Cannot update Ingress %s/%s due to missing service: %s", i.Namespace, i.Name, p.IngressEndpoint.PublishedService)
+				}
+			} else {
+				err := k8sClient.UpdateIngressStatus(i.Namespace, i.Name, p.IngressEndpoint.IP, p.IngressEndpoint.Hostname)
+				if err != nil {
+					log.Errorf("Cannot update Ingress %s/%s due to error: %s", i.Namespace, i.Name, err)
```
ldez: I think the log can be transform to an error.

docs/configuration/backends/kubernetes.md
```diff
@@ -105,14 +105,35 @@ A label selector can be defined to filter on specific Ingress objects only.
...
+#
+# Optional
+# ip = 127.0.0.1
+#
+# Optional
+# publishedservice = default/cheddar
+
+```
+
+You can configure a static hostname or IP address that traefik will add to the status section of ingress objects that it manages. If you prefer, you can provide a service, which traefik will copy the status spec from. This will give more flexibility in cloud/dynamic environments.
```
ldez: One sentence, one line

docs/configuration/backends/kubernetes.md
```diff
@@ -105,14 +105,35 @@ A label selector can be defined to filter on specific Ingress objects only.
...
+[kubernetes.ingressendpoint]
+#
+# Optional
+# hostname = localhost
+#
+# Optional
+# ip = 127.0.0.1
+#
+# Optional
+# publishedservice = default/cheddar
```
ldez: all these values don't seem optional.

provider/kubernetes/kubernetes.go
```diff
@@ -155,6 +167,50 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+
+			if !p.shouldProcessIngress(ingressClass) {
+				continue
+			}
+
+			if p.IngressEndpoint.PublishedService != "" {
+				s := strings.Split(p.IngressEndpoint.PublishedService, "/")
+				serviceNamespace, serviceName := s[0], s[1]
+				svc, exists, _ := k8sClient.GetService(serviceNamespace, serviceName)
+				//Set the status to the status of the fronting LB service
```
ldez: Could you add a space between `//` and the message

provider/kubernetes/kubernetes.go
```diff
@@ -155,6 +167,50 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+				if exists && svc.Status.LoadBalancer.Ingress == nil {
+					//service exists, but has no LoadBalancer status
```
ldez: Could you add a space between `//` and the message

provider/kubernetes/client.go
dtomcej: Fun police ;)

provider/kubernetes/kubernetes.go
dtomcej: Missed on move

docs/configuration/backends/kubernetes.md
dtomcej: All three are optional!

provider/kubernetes/kubernetes.go
dtomcej: If someone submits a non-compliant ingress, we just want to log an error, and continue processing. We don't want to interrupt the processing of other ingresses.

docs/configuration/backends/kubernetes.md
```diff
@@ -105,14 +105,37 @@ A label selector can be defined to filter on specific Ingress objects only.
 
 See [label-selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) for details.
 
+```toml
+# Enable IngressEndpoint configuration. This will allow Traefik to update the status section of ingress objects.
+#
+# Optional
+# Default: false
```
ldez: You could remove this line, because `kubernetes.ingressendpoint` is not a boolean.
I know that related to CLI option.

provider/kubernetes/kubernetes.go
```diff
@@ -323,10 +333,46 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 				}
 			}
 		}
+
+		err = p.updateIngressStatus(i, k8sClient)
+		if err != nil {
+			log.Printf("cannot update Ingress %s/%s due to error: %s", i.Namespace, i.Name, err)
```
dtomcej: should be errorf

ldez: Not here, because we don't want to stop to browse all other ingress

docs/configuration/backends/kubernetes.md
```diff
@@ -81,6 +81,20 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
...
 # filename = "kubernetes.tmpl"
+
+# Enable IngressEndpoint configuration.
+# This will allow Traefik to update the status section of ingress objects, if desired.
+#
+# Optional
+#
+# [kubernetes.ingressEndpoint]
+#
+# One must be configured.
```
marco-jantke: I think we have to emphasize more that there are two different configuration possibilities. So either: hostname + IP OR the service name. Or do I understand this wrong?

provider/kubernetes/kubernetes.go
```diff
@@ -36,17 +36,25 @@ const (
 	traefikDefaultIngressClass = "traefik"
 )
 
+// IngressEndpoint holds the endpoint information for the Kubernetes provider
+type IngressEndpoint struct {
+	IP               string `description:"IP used for Kubernetes Ingress endpoints"`
+	Hostname         string `description:"Hostname used for Kubernetes Ingress endpoints"`
+	PublishedService string `description:"Published Kubernetes Service to copy status from"`
```
marco-jantke: I don't understand why it's a "published service". Can you explain a bit why you chose the name?

provider/kubernetes/kubernetes.go
marco-jantke: But why would `log.Errorf` stop processing? If it's an error we should mark it that way in the logs, shouldn't we?

marco-jantke: You mix `%s` and `%v` for errors. I don't really mind as there is no logical difference in the output, but consistence is a thing and `%s` should be faster as it doesn't have to do type assertions and therefore would stick with it.

provider/kubernetes/kubernetes.go
```diff
@@ -323,10 +333,46 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
 }
 
+func (p *Provider) updateIngressStatus(i *extensionsv1beta1.Ingress, k8sClient Client) error {
+	// Only process if an IngressEndpoint has been configured
+	if p.IngressEndpoint == nil {
+		return nil
+	}
+
+	if p.IngressEndpoint.PublishedService == "" {
+		return k8sClient.UpdateIngressStatus(i.Namespace, i.Name, p.IngressEndpoint.IP, p.IngressEndpoint.Hostname)
```
marco-jantke: Should we maybe emit a warning when one of the fields is an empty string? Or is that expected?

docs/configuration/backends/kubernetes.md
```diff
@@ -81,6 +81,20 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
...
+# Publishedservice will override the hostname and ip settings if configured.
+#
+# hostname = "localhost"
+# ip = "127.0.0.1"
+# publishedService = "default/cheddar"
```
marco-jantke: Can we document that it's namespace/servicename? Or do you think this is clear to Kubernetes users? I didn't get it in the first moments. Maybe simply a different example would help like `publishedService = namespaceName/serviceName`

provider/kubernetes/kubernetes.go
marco-jantke: Logs should start with upper case.

ldez: oh yes sorry I wa thinking about `fmt.Errorf`

provider/kubernetes/kubernetes.go
errm: I think the name is copied from the Nginx ingress controller, It makes sense to have the same name.

I think the idea is that this is the service that publishes the ingress controller to the outside world.

provider/kubernetes/kubernetes.go
errm: I think is is expected that there would only be one of IP or Hostname ... perhaps we should check that they are not both empty?

docs/configuration/backends/kubernetes.md
ldez: I think the explanation is on the next line.

provider/kubernetes/kubernetes.go
ldez: I added a check if the both are empty.

provider/kubernetes/kubernetes.go
marco-jantke: Thanks!

docs/configuration/backends/kubernetes.md
marco-jantke: Can we change it to `At least one has to be configured`? This + the next sentence describes everything well then.

ldez:
It's already merged.

anandsinghkunwar:
Will this be cherry-picked for the older versions?

timoreimann:
@anandsinghkunwar we cherry-pick bug fixes only to minimize the risk of affecting otherwise stable releases adversely. Especially in this case, the code delta is quite big.

Traefik 1.7 is on the horizon though, it shouldn't take much longer anymore until we ship.

PR #3333, opened 2018-05-16
Route rules overlaps in UI

ldez:
### What does this PR do?

Route rules overlaps in UI.


marco-jantke:
You are sure that the app.sass should be deleted?

ldez:
Yes because it unused, we use the app.sass at the root of the project.

Issue #3355, opened 2018-05-20
Weird ACME errors out of a blue


sokoow:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

Tried to create a new ingress on kube with ACME enabled, suddenly getting cert errors

### What did you expect to see?

Traefik requesting new cert for new subdomain just fine

### What did you see instead?

```
time="2018-05-20T11:53:04Z" level=debug msg="Creating frontend git.puczat.pl/"
time="2018-05-20T11:53:04Z" level=debug msg="Wiring frontend git.puczat.pl/ to entryPoint http"
time="2018-05-20T11:53:04Z" level=debug msg="Creating route git.puczat.pl Host:git.puczat.pl"
time="2018-05-20T11:53:04Z" level=debug msg="Creating backend git.puczat.pl/"
time="2018-05-20T11:53:04Z" level=debug msg="Wiring frontend git.puczat.pl/ to entryPoint https"
time="2018-05-20T11:53:04Z" level=debug msg="Creating route git.puczat.pl Host:git.puczat.pl"
time="2018-05-20T11:53:04Z" level=debug msg="Creating backend git.puczat.pl/"
time="2018-05-20T11:53:04Z" level=debug msg="Try to challenge certificate for domain [git.puczat.pl] founded in Host rule"
time="2018-05-20T11:53:04Z" level=debug msg="Looking for provided certificate(s) to validate [\"git.puczat.pl\"]..."
time="2018-05-20T11:53:04Z" level=debug msg="Domains [\"git.puczat.pl\"] need ACME certificates generation for domains \"git.puczat.pl\"."
time="2018-05-20T11:53:04Z" level=debug msg="Loading ACME certificates [git.puczat.pl]..."
time="2018-05-20T11:53:04Z" level=error msg="Unable to obtain ACME certificate for domains \"git.puczat.pl\" detected thanks to rule \"PathPrefix:/\" : cannot get ACME client directory missing new registration URL"
time="2018-05-20T11:53:43Z" level=debug msg="No certificate provided dynamically can check the domain \"git.puczat.pl\", a per default certificate will be used."
```

### Output of `traefik version`: (_What version of Traefik are you using?_)
 1.6.1

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```yml
apiVersion: v1
data:
  traefik.toml: |
    # traefik.toml
    defaultEntryPoints = ["http","https"]

    [entryPoints]
      [entryPoints.http]
[...]

PR #3358, opened 2018-05-20
Force to use ACME v02 endpoint.

ldez:
### What does this PR do?

Force to use ACME v02 endpoint: rewrite API URL if needed.

### Motivation

Have a simple configuration.

Related to #3355

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -363,7 +363,12 @@ func (gc *GlobalConfiguration) initACMEProvider() {
 			gc.ACME.CAServer = caServer
 		}
 
-		// TODO: to remove in the futurs
+		if gc.ACME.DNSChallenge != nil && gc.ACME.HTTPChallenge != nil {
+			log.Warn("Unable to use DNS challenge and HTTP challenge at the same time. Fallback to DNS challenge.")
+			gc.ACME.HTTPChallenge = nil
+		}
+
+		// TODO: to remove in the future
 		if len(gc.ACME.StorageFile) > 0 && len(gc.ACME.Storage) == 0 {
 			log.Warn("ACME.StorageFile is deprecated, use ACME.Storage instead")
 			gc.ACME.Storage = gc.ACME.StorageFile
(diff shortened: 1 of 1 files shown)


configuration/configuration.go
```diff
@@ -356,6 +359,16 @@ func (gc *GlobalConfiguration) initTracing() {
 
 func (gc *GlobalConfiguration) initACMEProvider() {
 	if gc.ACME != nil {
+		if len(gc.ACME.CAServer) == 0 {
+			gc.ACME.CAServer = DefaultAcmeCAServer
+		}
+
+		if strings.Contains(gc.ACME.CAServer, "v01") {
+			caServer := strings.Replace(gc.ACME.CAServer, "v01", "v02", 1)
+			log.Warn("the CA server %q refer to a v01 endpoint of the ACME API, please change to %q.", gc.ACME.CAServer, caServer)
```
dtomcej: The log entry doesn't mention that you swapped v01 for v02. it is possible that for some reason, someone may want to use a CA server that contains `v01` in the URL. I would test for a more exact match, or just warn in the logs.

configuration/configuration.go
```diff
@@ -391,6 +401,28 @@ func (gc *GlobalConfiguration) initACMEProvider() {
...
 }
 
+func getSafeACMECAServer(caServerSrc string) string {
+	if len(caServerSrc) == 0 {
+		return DefaultAcmeCAServer
+	}
+
+	if strings.HasPrefix(caServerSrc, "https://acme-v01.api.letsencrypt.org") {
+		caServer := strings.Replace(caServerSrc, "v01", "v02", 1)
+		log.Warnf("The CA server %q refer to a v01 endpoint of the ACME API, please change to %q.", caServerSrc, caServer)
```
nmengin: s/refer/refer**s**
No sure users will understand that a modification will be automatically done.
WDYT to use a more explicit message?

configuration/configuration.go
```diff
@@ -391,6 +401,28 @@ func (gc *GlobalConfiguration) initACMEProvider() {
...
+		log.Warnf("ACME CA server fallback to %q.", caServer)
```
nmengin: Why are you using 2 logs instructrions?

Issue #3081, opened 2018-03-26
Backend server healthcheck handles status codes incorrectly


SniperCZE:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

1) There is healthcheck configured on backend server pool
2) Backend server returns 301 on healthcheck request

### What did you expect to see?

All status codes lower than 500 (strictly spoken lower than 400) should be consider as valid answer.

### What did you see instead?

Any code other than 200 results in disabled backend server. No info in documentation about this behavior.

### Output of `traefik version`: (_What version of Traefik are you using?_)
```
1.5.3
```

kuruwan87:
I'm falling in the same kind of situation here, too bad traefik handle only 200, this part should be customizable imho.
For instance I try to LB influxdb servers and the /ping request return a 204 which is quite accurate.

PR #3364, opened 2018-05-22
support all non-error (1xx, 2xx, 3xx) HTTP status codes for health check

SniperCZE:
### What does this PR do?
Allows all non-error HTTP status codes (based on RFC) to pass healthcheck.

### Motivation

PR #3362 adds only support for 2xx HTTP status codes, but according to RFC, 1xx and 3xx are valid answers too and should not make server pushed out of pool, so this improve fix of #3081.

### More

- [X] Added/updated tests


SniperCZE:
Hi @dtomcej I mean RFC 7231. There is section 6 about status codes and only 4xx and 5xx are labeled as error status. 1xx are informational - used ie. for upgrade http -> websockets. On the other hand, 1xx status codes are quite rare (in terms of loadbalancer healthcheck URLs) so probably 200 - 399 will be OK.

PR #3368, opened 2018-05-22
Update Kubernetes Port Documentation

### LINKED ISSUE #3348: Kubernetes Documentation Examples Port Configuration

Documentation Update

### What needs updating?

The kubernetes examples (https://docs.traefik.io/user-guide/kubernetes/) and (https://github.com/containous/traefik/tree/master/examples/k8s/traefik-ds.yaml) Have a nodeport service fronting a daemonset with hostports. This creates duplicate methods for accessing traefik, and is not clear for new users.

We should clean up the examples to keep the deployment/daemonset examples as close to each other as possible, and update the port configurations to be the same.

We should also make a note about hostports, and how they are not supported in production, and possibly a link to the k8s documentation.

dtomcej:
### What does this PR do?

Updates the documentation for kubernetes, in particular the port examples, and adds some notes about best practices

### Motivation

Fixes #3348 

### More

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -81,9 +81,11 @@ For namespaced restrictions, one RoleBinding is required per watched namespace a
 It is possible to use Træfik with a [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) or a [DaemonSet](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) object,
  whereas both options have their own pros and cons:
 
-- The scalability is much better when using a Deployment, because you will have a Single-Pod-per-Node model when using the DaemonSet.
-- It is possible to exclusively run a Service on a dedicated set of machines using taints and tolerations with a DaemonSet.
-- On the other hand the DaemonSet allows you to access any Node directly on Port 80 and 443, where you have to setup a [Service](https://kubernetes.io/docs/concepts/services-networking/service/) object with a Deployment.
+- The scalability can be much better when using a Deployment, because you will have a Single-Pod-per-Node model when using a DaemonSet, whereas you may need less replicas based on your environment when using a Deployment.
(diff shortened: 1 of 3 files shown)


dtomcej:
ping @containous/kubernetes

docs/user-guide/kubernetes.md
```diff
@@ -81,9 +81,11 @@ For namespaced restrictions, one RoleBinding is required per watched namespace a
...
+- The scalability can be much better when using a Deployment, because you will have a Single-Pod-per-Node model when using a DaemonSet, whereas you may need less replicas based on your environment when using a Deployment.
+- DaemonSets automatically scale to new nodes, when the nodes join the cluster, whereas Deployment pods are only scheduled on new nodes if required.
+- DaemonSets ensure that only one replica of pods run on any single node. Deployments require affinity settings if you want to ensure that two pods don't end up on the same node.
+- DaemonSets can be run with the `NET_BIND_SERVICE` capability, which will allow it to bind to port 80/443/etc on each host. This will allow bypassing the kube-proxy, and reduce traffic hops. Note that this is against the Kubernetes Best Practices [Guidelines](https://kubernetes.io/docs/concepts/configuration/overview/#services), and raises the potential for scheduling/scaling issues. Despite potential issues, this remains the choice for most ingress controllers.
+- If you are unsure which to choose, start with the Daemonset.
```
errm: If you want to run more than one Ingress config it is much more convenient to use a Deployment...
e.g. I have run setups with one controller for admin/intranet services and another for external traffic ...

You could run multiple demon sets with all different ports configured but it gets messy quickly...

Also we should mention that running as DeamonSet is the best for bare metal or any other environment where there is no supported load balancer ...

Also we should point out here that DaemonSets don't have to run on all nodes they can be targeted to a labeled subset if required https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/#running-pods-on-only-some-nodes

One strategy can be to have a dedicated pool of ingress enabled nodes that are registered (perhaps manually) with an upstream L3 LB

PR #3371, opened 2018-05-23
Implement h2c with backend

### LINKED ISSUE #2139: h2c (HTTP/2 without TLS) backend support
Would be nice to add support for HTTP/2 backend without TLS, reducing overhead and avoiding the use of selfsigned certificate and backend configuration.

We just need a way to indicate to Traefik that the backend should be contacted using HTTP2 directly (could be a h2c:// scheme in the url maybe?)

Maybe related to https://github.com/golang/go/issues/14141

Juliens:
### What does this PR do?
Add h2c support with backend.

### Motivation
Simplify the use of http2 with backend.
Fixes #2139 

### More

- [X] Added/updated tests

### Additional Notes
This PR add a new "protocol" in the backend server.
`h2c://{backend}`


docs/user-guide/grpc.md
```diff
@@ -76,40 +60,25 @@ rootCAs = [ "./backend.cert" ]
 ```
 
 !!! warning
-    With some backends, the server URLs use the IP, so you may need to configure `insecureSkipVerify` instead of the `rootCAS` to activate HTTPS without hostname verification.
+    For provider with label, you will have to specify the `frontend.protocol=h2c`
```
mmatur: `frontend.protocol` is a new label? It should be `traefik.protocol` instead?

PR #3382, opened 2018-05-24
[WIP] Namespace Discovery for using Kubernetes Labels

### LINKED ISSUE #3284: Proposal: Dynamic Namespace discovery for Kubernetes
## Proposal

This proposal is to discuss the addition of dynamic Namespace discovery for Traefik in Kubernetes.

### Background

Currently, it is possible to configure the Namespaces watched by Traefik via the static `kubernetes.namespaces` property. This property is used at Treafik startup to open watches on all relevant Kubernetes resources in these Namespaces. 

In our use case we're running on a Kubernetes cluster without cluster-admin rights. We have access to all the Namespaces we want to watch with Traefiik, but we don't have access to the `kube-system` Namespace. So we're able to deploy our Ingress resources and watch them with Traefik, if we configure these Namespaces statically in the Traefik configuration.

We're creating new Namespaces for every feature branch of our repository, so we have to deal with new Namespaces all the time. To get this to work with Traefik, we would have to update the `kubernetes.namespaces` property whenever a Namespace is added (or deleted) and restart Traefik.

### Suggestion

I propose to add Namespace discovery to Traefik. This would require a watch on Namespaces and some subsequent filtering, so that not every Namespace has to be watched. The filtering can be done in one of the following ways:
* whitelist/ blacklist with regexp
* labelSelector on Namespaces (similar to the [namespaceSelector of NetworkPolicies](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#networkpolicypeer-v1beta1-extensions))

Whenever changes in the set of Namespaces which should be watched are discovered, the watches have to be updated accordingly.

dtomcej:
### What does this PR do?

Enables namespace discovery using kubernetes label selectors

### Motivation

Fixes #3284 

### More

- [x] Added/updated tests

*WIP* Currently working on getting some integration/better tests to be able to catch issues with this.

Concurrency issues, and server interaction need to be testable.

### DIFF
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -43,6 +43,13 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 #
 # namespaces = ["default", "production"]
 
+# Namespace label selector to filter namespaces that should be watched.
+#
+# Optional
+# Default: empty (watch all namespaces).
+#
+# namespaceLabelSelector = "label = value"
+
 # Ingress label selector to filter Ingress objects that should be processed.
 #
 # Optional
@@ -122,6 +129,11 @@ See [label-selectors](https://kubernetes.io/docs/concepts/overview/working-with-
 You can configure a static hostname or IP address that Traefik will add to the status section of Ingress objects that it manages.
 If you prefer, you can provide a service, which traefik will copy the status spec from.
 This will give more flexibility in cloud/dynamic environments.
+
+### `namespaceLabelSelector`
+
+You can configure Traefik to watch for namespaces that have a particular label and value set.
+This will allow you to dynamically manage which namespaces Traefik watches without having to statically define them, or restart Traefik.
 
 ### TLS communication between Traefik and backend pods
 
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -44,6 +44,7 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # namespaces = ["default", "production"]
 
 # Namespace label selector to filter namespaces that should be watched.
+# This label selector is ignored if a static list of namespaces is provided in the "namespaces" option above
 #
 # Optional
 # Default: empty (watch all namespaces).
@@ -134,6 +135,7 @@ This will give more flexibility in cloud/dynamic environments.
 
 You can configure Traefik to watch for namespaces that have a particular label and value set.
 This will allow you to dynamically manage which namespaces Traefik watches without having to statically define them, or restart Traefik.
+This setting will be ignored if you provide a static list of namespaces to watch.
 
 ### TLS communication between Traefik and backend pods
 
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -44,7 +44,7 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # namespaces = ["default", "production"]
 
 # Namespace label selector to filter namespaces that should be watched.
-# This label selector is ignored if a static list of namespaces is provided in the "namespaces" option above
+# This label selector is ignored if a static list of namespaces is provided in the `namespaces` option above.
 #
 # Optional
 # Default: empty (watch all namespaces).
@@ -135,7 +135,7 @@ This will give more flexibility in cloud/dynamic environments.
 
 You can configure Traefik to watch for namespaces that have a particular label and value set.
 This will allow you to dynamically manage which namespaces Traefik watches without having to statically define them, or restart Traefik.
-This setting will be ignored if you provide a static list of namespaces to watch.
+This setting will be ignored if you provide a static list of namespaces to watch (via the `namespaces` label selector).
 
 ### TLS communication between Traefik and backend pods
 
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -135,7 +135,12 @@ This will give more flexibility in cloud/dynamic environments.
 
 You can configure Traefik to watch for namespaces that have a particular label and value set.
 This will allow you to dynamically manage which namespaces Traefik watches without having to statically define them, or restart Traefik.
-This setting will be ignored if you provide a static list of namespaces to watch (via the `namespaces` label selector).
+This setting will be ignored if you provide a static list of namespaces to watch (via the `namespaces` option).
+
+!!! note
+    Please note that by enabling the namespace label selector, Traefik may take longer to process ingresses when watching a new namespace with a large amount of objects, or a large number of new namespaces.
+    This is due to the fact that Traefik has to create new watchers for objects in the new namespaces, to allow for dynamic updates.
+    If you are concerned about performance in a large environment, it is advisable to provide a static list of namespaces to watch (via the `namespaces` option).
 
 ### TLS communication between Traefik and backend pods
 
--- a/provider/kubernetes/client.go
+++ b/provider/kubernetes/client.go
@@ -47,7 +47,6 @@ type Client interface {
 	GetSecret(namespace, name string) (*corev1.Secret, bool, error)
 	GetEndpoints(namespace, name string) (*corev1.Endpoints, bool, error)
 	UpdateIngressStatus(namespace, name, ip, hostname string) error
-	GetNamespaces() (*corev1.NamespaceList, error)
 }
 
 type clientImpl struct {
@@ -131,6 +130,7 @@ func (c *clientImpl) WatchNamespaces(namespaces Namespaces, stopCh <-chan struct
 		} else {
 			// namespacelabels are being used: watch all namespaces for events and use them to get namespaces to watch.
 			c.namespaceFactory.Start(stopCh)
+			c.namespaceFactory.WaitForCacheSync(stopCh)
 		}
 	}
 	return nil
@@ -141,13 +141,13 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}, eve
 	eventHandler := c.newResourceEventHandler(eventsChan)
 
 	var namespacesToWatch []string
-	namespaceList, err := c.GetNamespaces()
+	namespaceList, err := c.getNamespaces()
 	if err != nil {
 		return fmt.Errorf("could not list namespaces: %v", err)
 	}
 
 	for _, item := range namespaceList.Items {
-		log.Debugf("Adding found namespace: %q to namespace list", item.ObjectMeta.Name)
+		log.Debugf("Adding found namespace %q to namespace list", item.ObjectMeta.Name)
 		namespacesToWatch = append(namespacesToWatch, item.ObjectMeta.Name)
 	}
 
@@ -185,7 +185,7 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}, eve
 
 // GetIngresses returns all Ingresses for observed namespaces in the cluster.
 func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
-	namespaceList, err := c.GetNamespaces()
+	namespaceList, err := c.getNamespaces()
 	if err != nil {
 		log.Errorf("could not list namespaces: %v", err)
 		return nil
@@ -233,7 +233,7 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
 }
 
 // GetNamespaces returns namespaces with the configured labelselector.
-func (c *clientImpl) GetNamespaces() (*corev1.NamespaceList, error) {
+func (c *clientImpl) getNamespaces() (*corev1.NamespaceList, error) {
 	return c.clientset.CoreV1().Namespaces().List(metav1.ListOptions{LabelSelector: c.namespaceLabelSelector.String()})
 }
 
--- a/provider/kubernetes/client.go
+++ b/provider/kubernetes/client.go
@@ -140,28 +140,18 @@ func (c *clientImpl) WatchNamespaces(namespaces Namespaces, stopCh <-chan struct
 func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}, eventsChan chan<- interface{}) error {
 	eventHandler := c.newResourceEventHandler(eventsChan)
 
-	// If no namespaces are specified
-	if len(namespaces) == 0 {
-		// If no namespacelabels are specified
-		if c.namespaceLabelSelector.Empty() {
-			namespaces = Namespaces{metav1.NamespaceAll}
-		} else {
-			// namespacelabels are being used, get all namespaces being watched
-			namespaceList, err := c.GetNamespaces()
-			if err != nil {
-				return fmt.Errorf("could not list namespaces: %v", err)
-			}
-
-			log.Debugf("Current Namespace List: %+v", namespaceList)
-
-			for _, item := range namespaceList.Items {
-				log.Debugf("Adding found namespace: %q to namespace list", item.ObjectMeta.Name)
-				namespaces = append(namespaces, item.ObjectMeta.Name)
-			}
-		}
+	var namespacesToWatch []string
+	namespaceList, err := c.GetNamespaces()
+	if err != nil {
+		return fmt.Errorf("could not list namespaces: %v", err)
 	}
 
-	for _, ns := range namespaces {
+	for _, item := range namespaceList.Items {
+		log.Debugf("Adding found namespace: %q to namespace list", item.ObjectMeta.Name)
+		namespacesToWatch = append(namespacesToWatch, item.ObjectMeta.Name)
+	}
+
+	for _, ns := range namespacesToWatch {
 		factory := informers.NewFilteredSharedInformerFactory(c.clientset, resyncPeriod, ns, nil)
 		factory.Extensions().V1beta1().Ingresses().Informer().AddEventHandler(eventHandler)
 		factory.Core().V1().Services().Informer().AddEventHandler(eventHandler)
@@ -169,11 +159,11 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}, eve
 		c.factories[ns] = factory
 	}
 
-	for _, ns := range namespaces {
+	for _, ns := range namespacesToWatch {
 		c.factories[ns].Start(stopCh)
 	}
 
-	for _, ns := range namespaces {
+	for _, ns := range namespacesToWatch {
 		for t, ok := range c.factories[ns].WaitForCacheSync(stopCh) {
 			if !ok {
 				return fmt.Errorf("timed out waiting for controller caches to sync %s in namespace %q", t.String(), ns)
@@ -185,7 +175,7 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}, eve
 	// users having granted RBAC permissions for this object.
 	// https://github.com/containous/traefik/issues/1784 should improve the
 	// situation here in the future.
-	for _, ns := range namespaces {
+	for _, ns := range namespacesToWatch {
 		c.factories[ns].Core().V1().Secrets().Informer().AddEventHandler(eventHandler)
 		c.factories[ns].Start(stopCh)
 	}
--- a/provider/kubernetes/client.go
+++ b/provider/kubernetes/client.go
@@ -71,7 +71,7 @@ func newClientImpl(clientset *kubernetes.Clientset) *clientImpl {
 func newInClusterClient(endpoint string) (*clientImpl, error) {
 	config, err := rest.InClusterConfig()
 	if err != nil {
-		return nil, fmt.Errorf("failed to create in-cluster configuration: %s", err)
+		return nil, fmt.Errorf("failed to create in-cluster configuration: %v", err)
 	}
 
 	if endpoint != "" {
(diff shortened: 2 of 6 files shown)


provider/kubernetes/client.go
```diff
@@ -196,6 +226,16 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
 	return nil
 }
 
+// GetNamespaces returns namespaces with the given labelselector.
+func (c *clientImpl) GetNamespaces() (*corev1.NamespaceList, error) {
+	var namespaces *corev1.NamespaceList
```
```go
func (c *clientImpl) GetNamespaces() (*corev1.NamespaceList, error) {
	item, err := c.clientset.CoreV1().Namespaces().List(metav1.ListOptions{LabelSelector: c.namespaceLabelSelector.String()})
	if err != nil {
		return nil, err
	}
	return item, nil
}
```

provider/kubernetes/client.go
```diff
@@ -155,8 +177,16 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}) (<-
...
 func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 	var result []*extensionsv1beta1.Ingress
-	for ns, factory := range c.factories {
-		ings, err := factory.Extensions().V1beta1().Ingresses().Lister().List(c.ingressLabelSelector)
+	namespaceList, err := c.GetNamespaces()
+	if err != nil {
+		log.Errorf("Could not list namespaces: %s", err)
+	}
+	namespaces := Namespaces{}
+	for i := 0; i < len(namespaceList.Items); i++ {
```
```go
func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
	var result []*extensionsv1beta1.Ingress
	namespaceList, err := c.GetNamespaces()
	if err != nil {
		log.Errorf("could not list namespaces: %v", err)
		return result
	}

	var namespaces Namespaces
	for _, item := range namespaceList.Items {
		namespaces = append(namespaces, item.ObjectMeta.Name)
	}

	for _, ns := range namespaces {
		ings, err := c.factories[ns].Extensions().V1beta1().Ingresses().Lister().List(c.ingressLabelSelector)
		if err != nil {
			log.Errorf("Failed to list ingresses in namespace %s: %s", ns, err)
			continue
		}
		for _, ing := range ings {
			result = append(result, ing)
		}
	}
	return result
}
```

provider/kubernetes/client.go
```diff
@@ -113,13 +116,32 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+			namespaces = Namespaces{metav1.NamespaceAll}
+			c.isNamespaceAll = true
+		} else {
+			// namespacelabels are being used, watch all namespaces for events, use them to get namespaces to watch..
+			c.namespaceFactory.Start(stopCh)
+			namespaceList, err := c.clientset.CoreV1().Namespaces().List(metav1.ListOptions{LabelSelector: c.namespaceLabelSelector.String()})
+			if err != nil {
+				return eventCh, fmt.Errorf("Could not list namespaces with error: %s", err)
+			}
+			namespaces = Namespaces{}
```
```go
var namespaces Namespaces
for _, item := range namespaceList.Items {
	namespaces = append(namespaces, item.ObjectMeta.Name)
}
```

provider/kubernetes/client.go
```diff
@@ -113,13 +116,32 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+		// If no namespacelabels are specified
+		if c.namespaceLabelSelector.String() == "" {
+			namespaces = Namespaces{metav1.NamespaceAll}
+			c.isNamespaceAll = true
+		} else {
+			// namespacelabels are being used, watch all namespaces for events, use them to get namespaces to watch..
+			c.namespaceFactory.Start(stopCh)
+			namespaceList, err := c.clientset.CoreV1().Namespaces().List(metav1.ListOptions{LabelSelector: c.namespaceLabelSelector.String()})
+			if err != nil {
+				return eventCh, fmt.Errorf("Could not list namespaces with error: %s", err)
```
ldez: `Could` -> `could`

provider/kubernetes/client.go
```diff
@@ -155,8 +177,16 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}) (<-
 // GetIngresses returns all Ingresses for observed namespaces in the cluster.
 func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 	var result []*extensionsv1beta1.Ingress
-	for ns, factory := range c.factories {
-		ings, err := factory.Extensions().V1beta1().Ingresses().Lister().List(c.ingressLabelSelector)
+	namespaceList, err := c.GetNamespaces()
+	if err != nil {
+		log.Errorf("Could not list namespaces: %s", err)
```
```go
log.Errorf("could not list namespaces: %v", err)
```

provider/kubernetes/kubernetes.go
```diff
@@ -53,17 +53,22 @@ type Provider struct {
...
+func (p *Provider) newK8sClient(ingressLabelSelector, namespaceLabelSelector string) (Client, error) {
 	ingLabelSel, err := labels.Parse(ingressLabelSelector)
 	if err != nil {
 		return nil, fmt.Errorf("invalid ingress label selector: %q", ingressLabelSelector)
 	}
-	log.Infof("ingress label selector is: %q", ingLabelSel)
+
+	nsLabelSel, err := labels.Parse(namespaceLabelSelector)
+	if err != nil {
+		return nil, fmt.Errorf("invalid namespace label selector: %q", namespaceLabelSelector)
```
ldez:
```go
return nil, fmt.Errorf("invalid namespace label selector %q: %v", namespaceLabelSelector, err)
```

provider/kubernetes/kubernetes.go
```diff
@@ -105,7 +111,8 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 	}
 
 	log.Debugf("Using Ingress label selector: %q", p.LabelSelector)
-	k8sClient, err := p.newK8sClient(p.LabelSelector)
+	log.Debugf("Using Namespace label selector: %q", p.NamespaceLabelSelector)
```
ldez:
```go
log.Debugf("Using Ingress label selector %q and Namespace label selector %q", p.LabelSelector,  p.NamespaceLabelSelector)
```

provider/kubernetes/client.go
```diff
@@ -155,8 +177,18 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}) (<-
 // GetIngresses returns all Ingresses for observed namespaces in the cluster.
 func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 	var result []*extensionsv1beta1.Ingress
-	for ns, factory := range c.factories {
-		ings, err := factory.Extensions().V1beta1().Ingresses().Lister().List(c.ingressLabelSelector)
+	namespaceList, err := c.GetNamespaces()
+	if err != nil {
+		log.Errorf("could not list namespaces: %v", err)
```
```go
log.Errorf("could not list namespaces: %v", err)
return result
```

provider/kubernetes/client.go
```diff
@@ -155,8 +177,18 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}) (<-
...
+	}
+
+	var namespaces Namespaces
+	for _, item := range namespaceList.Items {
+		namespaces = append(namespaces, item.ObjectMeta.Name)
+	}
+
+	for _, ns := range namespaces {
+		ings, err := c.factories[ns].Extensions().V1beta1().Ingresses().Lister().List(c.ingressLabelSelector)
```
```go
for _, ns := range namespaces {
	ings, err := c.factories[ns].Extensions().V1beta1().Ingresses().Lister().List(c.ingressLabelSelector)
	if err != nil {
		log.Errorf("Failed to list ingresses in namespace %s: %s", ns, err)
		continue
	}
	for _, ing := range ings {
		result = append(result, ing)
	}
}
```

provider/kubernetes/client.go
```diff
@@ -113,13 +116,32 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
 // WatchAll starts namespace-specific controllers for all relevant kinds.
 func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}) (<-chan interface{}, error) {
 	eventCh := make(chan interface{}, 1)
+	eventHandler := c.newResourceEventHandler(eventCh)
+
+	nsf := informers.NewSharedInformerFactory(c.clientset, resyncPeriod)
+	nsf.Core().V1().Namespaces().Informer().AddEventHandler(eventHandler)
+	c.namespaceFactory = nsf
```
vdemeester: is there any reason for doing that over the following ?

```go
c.namespaceFactory = informers.NewSharedInformerFactory(c.clientset, resyncPeriod)
c.namespaceFactory.Core().V1().Namespaces().Informer().AddEventHandler(eventHandler)
```

dtomcej: `non-name c.namespaceFactory on left side of :=` ;)

dtomcej: oooo cleaner!

dtomcej:
Testing:
- Start traefik with `--kubernetes.namespacelabelselector="environment=test"`
- Create 2 namespaces (`one` and `two`)
- traefik should register that 2 namespaces were created, but will ignore them
- create an ingress in namespace `one` and an ingress in namespace `two`
- label the namespace with `kubectl label namespace one environment=test`
- traefik should now pick up the ingress in the `one` namespace
- remove the label from namespace `one` with `kubectl label namespace one environment-`
- traefik should remove the ingress from its configuration
- label namespace two with `kubectl label namespace twi environment=test`
- traefik should pick up the ingress from the `two` namespace

dtomcej:
Note that this PR currently has a bunch of Debug logging that will be removed when the functionality is confirmed.

docs/configuration/backends/kubernetes.md
```diff
@@ -43,6 +43,13 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 #
 # namespaces = ["default", "production"]
 
+# Namespace label selector to filter namespaces that should be watched.
+#
+# Optional
+# Default: empty (watch all namespaces).
+#
+# namespaceLabelSelector = "label = value"
```
timoreimann: We should document the behavior if both this new option and `namespaces` are defined.

provider/kubernetes/client.go
```diff
@@ -110,16 +114,53 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+
+	c.namespaceFactory = informers.NewSharedInformerFactory(c.clientset, resyncPeriod)
+	c.namespaceFactory.Core().V1().Namespaces().Informer().AddEventHandler(eventHandler)
 
+	// If no namespaces are specified
 	if len(namespaces) == 0 {
-		namespaces = Namespaces{metav1.NamespaceAll}
-		c.isNamespaceAll = true
+		// If no namespacelabels are specified
+		if c.namespaceLabelSelector.String() == "" {
```
timoreimann: `c.namespaceLabelSelector` should have an `Empty()` method, which would be more idiomatic to use.

provider/kubernetes/client.go
```diff
@@ -110,16 +114,53 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+			namespaces = Namespaces{metav1.NamespaceAll}
+			c.isNamespaceAll = true
+		} else {
+			// namespacelabels are being used, watch all namespaces for events, use them to get namespaces to watch..
```
timoreimann: Extra trailing dot.

timoreimann: `ETOOMANYCOMMAS` :) How about

`// namespacelabels are being used: watch all namespaces for events and use them to get namespaces to watch.`

provider/kubernetes/client.go
```diff
@@ -110,16 +114,53 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}) (<-chan interface{}, error) {
+	eventCh := make(chan interface{}, 1)
 	eventHandler := c.newResourceEventHandler(eventCh)
+
+	// If no namespaces are specified
+	if len(namespaces) == 0 {
+		// If no namespacelabels are specified
+		if c.namespaceLabelSelector.String() == "" {
+			namespaces = Namespaces{metav1.NamespaceAll}
+			c.isNamespaceAll = true
```
timoreimann: We've already set that variable above. Is it necessary to do it again?

provider/kubernetes/client.go
```diff
@@ -110,16 +114,53 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+		} else {
+			// namespacelabels are being used, get all namespaces being watched
+			namespaceList, err := c.clientset.CoreV1().Namespaces().List(metav1.ListOptions{LabelSelector: c.namespaceLabelSelector.String()})
+			log.Debugf("Current Namespace List: %+v", namespaceList)
+			if err != nil {
+				return eventCh, fmt.Errorf("could not list namespaces with error: %s", err)
```
timoreimann: Drop `with error`.

provider/kubernetes/client.go
```diff
@@ -110,16 +114,53 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+			}
+			for _, item := range namespaceList.Items {
+				log.Debugf("Adding found namespace: %s to namespace list", item.ObjectMeta.Name)
```
timoreimann: I'd omit the colon and use `%q` instead of `%s`.

provider/kubernetes/client.go
```diff
@@ -110,16 +114,53 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+				namespaces = append(namespaces, item.ObjectMeta.Name)
+			}
+		}
+	}
+	log.Debugf("Starting watches on %v namespaces", namespaces)
```
timoreimann: How about:

`"Starting watches on namespaces: %v", namespaces`

provider/kubernetes/client.go
```diff
@@ -155,10 +196,23 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}) (<-
...
+	if err != nil {
+		log.Errorf("could not list namespaces: %v", err)
+		return result
+	}
+
+	var namespaces Namespaces
+	for _, item := range namespaceList.Items {
+		namespaces = append(namespaces, item.ObjectMeta.Name)
+	}
+	for _, ns := range namespaces {
```
timoreimann: You can just nest this for loop above (i.e., move it to line 207 and skip the extra `namespaces` slice).

provider/kubernetes/client.go
```diff
@@ -155,10 +196,23 @@ func (c *clientImpl) WatchAll(namespaces Namespaces, stopCh <-chan struct{}) (<-
...
+		log.Debugf("ns: %v, sel: %v", ns, c.ingressLabelSelector)
+		log.Debugf("factories: %+v", c.factories)
```
timoreimann: We probably want to drop this debug message (and possibly a few others as well) before merging.

provider/kubernetes/client.go
```diff
@@ -196,6 +250,15 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
 	return nil
 }
 
+// GetNamespaces returns namespaces with the given labelselector.
```
timoreimann: s/given/configured/

provider/kubernetes/kubernetes.go
```diff
@@ -116,45 +122,20 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
-				eventsChan, err := k8sClient.WatchAll(p.Namespaces, stopWatch)
+				err := p.watch(k8sClient, configurationChan)
 				if err != nil {
 					log.Errorf("Error watching kubernetes events: %v", err)
 					timer := time.NewTimer(1 * time.Second)
+					// Don't retry on a stop message, return nil.
 					select {
 					case <-timer.C:
 						return err
 					case <-stop:
```
ZloyDyadka: There is a problem with exiting the function if we want to exit by the stop signal, not an error

provider/kubernetes/kubernetes.go
```diff
@@ -159,10 +140,66 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+	}
+
+	for {
+		//	Start watching the namespaced content
+		err := p.watchNamespaceContent(k8sClient, configurationChan, namespaceChan)
+		if err != nil {
+			return fmt.Errorf("error watching namespaced content: %v", err)
+		}
+	}
+}
```
timoreimann: I'm concerned about performance/scalability here: AFAIU, a change in one of the watched namespaces (addition, removal, etc.) causes all informers for all watched namespaces to be recreated. We have to wait for each informer to be synched (i.e., populated), and we do so sequentially, which means that no Ingresses can be processed during this period.

The delays incurred could become unacceptable for installations with a larger number of namespaces, or those with a moderate number of namespaces that contain a large number of objects. Aggravatingly, the resync period would cause a reset to happen even when no watched namespaces change. (That issue could be fixed rather easily though by disabling the resync period just for the namespace watcher.)

Given my understanding is correct, I believe that in order for the approach to become bearable, we would have to change the implementation such that we identify delta changes happening on the namespaces watch (i.e., distinguish between namespaces getting added, removed, or modified) and respond to those only. That is, we start watching new namespaces and remove watches on disappearing namespaces, all while keeping unchanged namespaces untouched. But even then, we might still have to run these optimized changes concurrently so that Ingress processing for untouched namespaces does not slow down.

I wonder if the added complexity we'd have to buy here still justifies the feature value. Before considering that path I'd like to hear what you have to say first -- my hope is that I get things wrong and everything is actually much better than I see it right now. :-)

provider/kubernetes/kubernetes.go
timoreimann: I think @ZloyDyadka is right: The new method `watch()` isn't meant to return anymore, which means that we normally have no chance to process this line to terminate Ingress processing gracefully.

provider/kubernetes/kubernetes.go
dtomcej: I think that you are correct about performance and scalability. The delay could potentially scale out of the trivial range, but in environments where scalability/performance is an issue, the dynamic namespace addition may not be a feasible solution, and they may need to stick to the static selection.

The namespace watcher is on a completely separate channel, and should not trigger on anything other than a namespace event.

You are correct about having to implement a delta system for watches for large numbers of objects, but of course, due to synchronicity issues and caching, there is a possibility that the namespace may not meet the requirements before it is removed, and may cause issues, even with the statically defined namespace filter that we previously used.

The point of this PR was to change as little of the operational code as possible (merely change the flow to extend functionality), and then potentially optimize in the future if performance issues arise (as described above, there should be no performance change for statically defined namespaces). This way, the feature could be released in 1.7, and then improved in future patch releases, as needed, without holding up the feature for smaller environments where performance may not be an issue.

timoreimann: >This way, the feature could be released in 1.7, and then improved in future patch releases, as needed, without holding up the feature for smaller environments where performance may not be an issue.

I'm not too sure about this. These days, Traefik seems to be used in many environments where scalability is needed right from the start, with thousands of Ingresses involved sometimes. The current Ingress processing implementation is very fast and responds to updates rapidly, which I think users appreciate and possibly even expect at this point. My feeling is that if we don't build in scalability right from the start for new features, we risk losing more users than we can win by a functional but under-performing functionality.

I think it'd help a lot if we knew the approximate point at which the implementation would start degrading measurably. I'd be much more relaxed if we knew that namespace changes for 500 Ingresses can be handled just fine, but not so much if it were 50. Maybe we can try to quantify the impact?

>The namespace watcher is on a completely separate channel, and should not trigger on anything other than a namespace event.

Except for the resync period which causes "fake" events to be emitted periodically. At the current state, we [reuse the existing resync period](https://github.com/dtomcej/traefik/blob/837f2548c07ca0dab3325127bafbaadc23818e81/provider/kubernetes/client.go#L122) so informer factories are recreated at least every 10 minutes. This could be a lightweight operation or induce seconds-scale latencies, depending on the number of namespaces and objects.

provider/kubernetes/client.go
```diff
@@ -111,32 +114,60 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+	c.namespaceFactory = informers.NewSharedInformerFactory(c.clientset, resyncPeriod)
+	c.namespaceFactory.Core().V1().Namespaces().Informer().AddEventHandler(eventHandler)
 
+	// If no namespaces are specified
 	if len(namespaces) == 0 {
-		namespaces = Namespaces{metav1.NamespaceAll}
-		c.isNamespaceAll = true
+		// If no namespacelabels are specified
+		if c.namespaceLabelSelector.Empty() {
+			namespaces = Namespaces{metav1.NamespaceAll}
```
timoreimann: The linter tells me: _This value of namespaces is never used._

provider/kubernetes/client.go
```diff
@@ -111,32 +114,60 @@ func createClientFromConfig(c *rest.Config) (*clientImpl, error) {
...
+			c.isNamespaceAll = true
```
timoreimann: Is this variable obsolete now?

dtomcej:
After discussing this with the team, I am going to be closing this PR, but keeping the branch for future development, for the following reasons:
1. This PR introduces a really cool feature, but it completely refactors the kubernetes provider. It needs it, but it interferes with the next major Traefik release, as many server refactors have already been done.
2. To be able to properly test the effect of this PR, we need to improve our k8s tests, and I don't want to keep this PR open while that happens.
3. Once the major server refactors have taken place, I will reopen a new PR with this feature implemented, but based _after_ the server refactor, instead of trying to carry this forward.

laurivosandi:
What is current status on this? Traefik v2.x seems to only have `--providers.kubernetescrd.namespaces`

ddtmachado:
Hi @laurivosandi, currently there is no plan to move forward with this feature.

PR #3391, opened 2018-05-25
Metrics: Add support for InfluxDB Database / RetentionPolicy and HTTP client

drewkerrigan:
### What does this PR do?

* Adds fields to `types.InfluxDB`
    * `Protocol`: Defaults to `"udp"`, can also be `"http"`
    * `Database`, `RetentionPolicy`: Allows specification of database and RP when protocol is `"http"` (UDP client does not support queries)
* Adds a `net/http/httptest` / channel based test similar to the existing udp based test in metrics/influxdb_test.go
* Adds new fields to documentation

### Motivation

The InfluxDB UDP endpoint does not support choosing a database, so environments that have many datasources for InfluxDB generally need the ability to separate influx data into multiple databases.

### More

- [x] Added/updated tests

### Additional Notes

One of the more bigger changes here is the fact that it attempts to create a database if an error is encountered containing the message `"database not found"`. This is a nice to have, (but not required) feature present in other similar applications (like telegraf -> https://github.com/influxdata/telegraf/blob/master/plugins/outputs/influxdb/influxdb.go#L209-L215 for example).


metrics/influxdb.go
```diff
@@ -67,22 +67,31 @@ func RegisterInfluxDB(config *types.InfluxDB) Registry {
...
-			log.Errorf("Unable to parse influxdb address: %s", err)
+			log.Errorf("Unable to parse influxdb address: %v, defaulting to udp.", err)
+			config.Protocol = "udp"
+			config.Database = ""
+			config.RetentionPolicy = ""
 		}
-	} else {
-		log.Warnf("Unsupported protocol: %s, defaulting to udp")
+	default:
+		log.Warnf("Unsupported protocol: %s, defaulting to udp.", config.Address)
```
drewkerrigan: Should be

```
log.Warnf("Unsupported protocol: %s, defaulting to udp.", config.Protocol)
```

PR #3395, opened 2018-05-26
Fix the Consulcatalog provider not refreshing a service with multiple health checks

rycus86:
### What does this PR do?

Fixes an issue with the Consulcatalog provider when there are multiple health checks defined for a service, and only part of them starts failing or comes back. Because only Service IDs were previously collected, the set of them didn't change. Check IDs should be unique, so we'll notice when any of them goes away or comes back.

### Motivation

I was setting up some demos with Traefik and Consul, and it was bothering me why it doesn't always work. :)
See an example here: https://github.com/rycus86/podlike/tree/master/examples/service-mesh

### More

- [x] Added/updated tests

### Additional Notes

Thank you!

### DIFF
--- a/integration/consul_catalog_test.go
+++ b/integration/consul_catalog_test.go
@@ -1,12 +1,15 @@
 package integration
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"time"
 
 	"github.com/containous/traefik/integration/try"
 	"github.com/containous/traefik/provider/label"
+	"github.com/docker/docker/api/types"
+	docker "github.com/docker/docker/client"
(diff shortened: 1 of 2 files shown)


mmatur:
Hi @rycus86,

Thanks for your interest in Træfik. 

As it is explain [here](https://www.consul.io/api/agent/check.html#id). The `CheckID` is unique for one node.
So if there are multiple consul nodes in the cluster with the same CheckID your fix will not handle this case

By the way there is already a PR that fix this issue #3390. So I will close this one in favor of #3390.

If you have questions or comments please fell free to ask.

PR #3400, opened 2018-05-28
LocalAddr and LocalIP matchers

### LINKED ISSUE #3399: Add ability to match on local listening address

Feature

Especially in larger, dynamic environments it might be useful to not only define a Host or HostRegexp matcher, but also match on the local IP address or IP:port combination that is handling the incoming request.
This allows pinning specific IPs to specific backends.

speijnik:
### What does this PR do?

Introduce LocalAddr and LocalIP matchers, which can be used for matching a specific local IP or IP:port combination to a specific backend.
Fixes #3399.

### Motivation

I am working on larger, automated Traefik deployments which need the ability to pin specific IP addresses to specific backends.

### More

- [x] Added/updated tests


dtomcej:
Closed via #3399

PR #3403, opened 2018-05-28
CNAME flattening

### LINKED ISSUE #141: CNAME flattening
I'd like Traefik to support CNAME flattening so that when I have a domain with a CNAME that resolves to the proxy the proxy should:

1) Do a normal backend match to see if anything matches the given host name, do the normal thing if matched
2) When 1 fails to find a match, resolve DNS on the host name to check for cname records. 
3) If a CNAME record exists, retry #1
4) 404 as normal

We should probably also keep a capped in-memory cache that respects DNS TTL.

Happy to make a PR, just wanted to establish some consensus on approach.

gamalan:
### What does this PR do?
Allow handling host rule with different names than the HOST header that comes in.

### Motivation

I am working in a project that allow user masking their domain/subdomain to our domain. Previously, only able to serve request that have same host rule and host header that came.

Fixes #141

### More

- [x] Added dependency(https://github.com/patrickmn/go-cache)
- [x] Add new set of config
- [x] Added/updated unit test
- [x] Add new integration test
- [x] Add new documentation

### Additional Notes

- To allow generate/serving the request SSL, must use httpChallenge with active onDemand rule, dnsChallenge is only possible if the request host dns is also managed by ours own.
- ACME onDemand config, should not be deprecated, if it being deprecated and removed, then this feature will lost the ability to generate the requested SSL cert. Resulting to a breaking response because SSL protocol error, Invalid CA/CN, etc.
- Current additional config example 
```toml
[hostResolver] #currently using resolver label
cnameFlattening = true #default is false, if this not set true, following config is useless
resolvConfig = "/etc/resolv.conf" # default is /etc/resolv.conf, if changed, the following file must follow resolv.conf standard
resolvDepth = 5 # maximum CNAME lookup depth before finishing
```


nmengin:
Hello @gamalan .

Many thanks for this PR. :+1: 

We talked about it with the team and, even if we understand the need, we are not sure that it's a best practice to check the `CNAME flattening` by default.
Indeed, only Cloudflare implements this (useful) feature and, I guess, only a part of Clouflare users use it.

That's why, I guess it should be great if the `CNAME flattening` was enable thanks to a flag. Thus only users who really need it will activate it. 

WDYT?

gregorybrzeski:
AFAIK Openstack Swift is using CNAME flattening, as a precaution they have a limit (lookup_depth) of dns resolves which are allowed for each request.

https://docs.openstack.org/newton/config-reference/object-storage/features.html

"CNAME lookup"

https://github.com/openstack/swift/blob/master/swift/common/middleware/cname_lookup.py

gamalan:
@nmengin agreed, not all people need this. My take is setup an additional set of global config, inside entrypoint, called entryPoints.resolver. With set of value for : 
1. CNAMEFlattening = true/false if false, any other option is useless, default is false
2. resolverDepth = 0-XX, how deep it should lookup a dns record, 0 means unlimited. default is 0
3. cache = 0-XX, cache time, in a minutes format, default is 30 minutes.

An example : 
```toml
[entryPoints.resolver]
CNAMEFlattening = true
resolverDepth = 10
cache = 30
```
Happy to accept some opinion.

gregorybrzeski:
@gamalan if you set `resolverDepth` by default to 0 (unlimited), then what would happen if the CNAME resolve chain has a circular reference ? Maybe setting it by default to 3 or 5 would be a better option ?

Would you also consider a case where dns servers for CNAMEFlattening resolving can be specified in config file ? The `/etc/resolv.conf` can be pointing to internal dns servers which might not be (in some cases) best choice to use as a resolving dns servers for CNAMEFlattening process.

gamalan:
@gregorybrzeski ah, forgot about that case, you are right, I think 5 should suffice for default. Also for the resolver config, good idea, but it should follow the resolv.conf standard(not sure gonna work in windows though).

gamalan:
Already make a new set of config, haven't been pushed yet. Currently, the CNAME flattening process is separated as a new package, the unit test has been added, the integration test and documentation is WIP. Not sure where to put for the documentation.

The additional config is set to global, as example
```toml
[resolver] #currently using resolver
cnameFlattening = true #default is false, if this not set true, following config is useless
resolvConfig = "/etc/resolv.conf" # default is /etc/resolv.conf, if changed, the following file must follow resolv.conf standard
resolvDepth = 5 # maximum CNAME lookup depth before finishing
cacheDuration = "30m" # default cache expiration 30 minute
```

nmengin:
Hello @gamalan ,

Many thanks for the modifications you made.

We think the `resolver` block is a good idea but we prefer keep the `resolvdepth` and the `cacheduration` hardcoded.

The flag should look like this : 
```toml
[resolver] #currently using resolver label
cnameFlattening = true #default is false, if this not set true, following config is useless
resolvConfig = "/etc/resolv.conf" # default is /etc/resolv.conf, if changed, the following file must follow resolv.conf standard
```

Moreover, we would like to give the possibility to the users to enable `resolver` and customize`resolver.resolvconfig` even if the `resolver.cnameflattening` is set to false.
It can be useful for others Use Cases.

WDYT?

gamalan:
@nmengin well, actually, resolvConfig still can be set if cnameFlattening is not true, but it will not be used for the process. But I like to be able to set resolvDepth and cacheDuration, because some people might have their own setting preference. The default is hardcoded and some might not need to change it anyway. Also, I change the block from `resolver` to `hostResolver`, to align it with the added package.

nmengin:
Hello @gamalan .

We are agree with you about the customization of `resolvDepth` but we are not sure for the `cacheDuration`.

Indeed, users can need more or less depth levels and if we do not allow the `resolvDepth` customization, these users can have a real problem.
But the cache duration is a just a timeout, it can be nice to customize it but users can manage correctly their own Træfik without modifying it. 

Even if, it's just an optional tag, we try to keep Træfik configuration as simpler as we can, that's why we prefer to not introduce  options which seem to not be very useful.
WDYT?

I have another question about your Pull Request description:

> must use httpChallenge with active onDemand rule, 

Can explain why only `onDemand` works and not `onHostRule` and `acme.domains` please?

Thanks in advance

gamalan:
Hello @nmengin 
The `cacheDuration` exists to limit CNAME flattening process. This is not for the DNS resolving timeout, it's for how long the result stays in cache. If the cache stays longer, then it will decrease the DNS resolving process count. It can be removed, but then the user won't be able to change it and stuck with the default. Another scenario is removing the cache altogether, but then it would cost a lot of processes.

Regarding this part
> must use httpChallenge with active onDemand rule
Although it can use `onHostRule` or `[acme.domains]`, it means the CNAME record(host) must be added in the firsthand, which defeat the purpose. Also change in the config, when running as docker container, is not refreshed unless restarted(CMIIW). So, to generate the request SSL cert it must use `onDemand` process, despite its slow process and Let's Encrypt rate limit. This is a tradeoff that one must understand when using this feature, especially when serving https request.

Thank you.

gregorybrzeski:
I second the ability to allow users to set their own `resolvDepth` however I believe that it should be set to a low number by default. At the same time I have a mixed feelings if there should be a hardcoded maximum value or allow unlimited (0). Numerous application which I have seen have a hardcoded maximum value to prevent circular references, on the other hand I like the ability for a knowledgable user to have the flexibility to set it to any number which particular `use case` requires. 

Regarding `cacheDuration` what would you say to set it to `TTL` value ? There is no additional overhead as the dns responses should already contain this information. This approach is simple (doesn't require additional configuration settings) and at the same time allows user flexibility in managing the cache duration.

nmengin:
@gamalan 

> Also change in the config, when running as docker container, is not refreshed unless restarted(CMIIW)

Are you talking about restarting the container (it's true) or restarting Træfik (it's false) ?

I understand your need about the `onDemand` option and, even if I guess the use case can be solved by using the DNS challenge (if the NS entry exists before starting Træfik), we will take this use case in account for the `onDemand` option.

Design LGTM.

gamalan:
@nmengin

The container one. I think using DNS Challenge is not possible, well not
entirely. If the incoming request DNS also internally managed, then it
would be possible. But, if it not, then it would be impossible, because we
can't change the DNS record.

So far, haven't got any idea, unless using HTTP Challenge.

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+	result = append(result, host)
+	if hr.Cache == nil {
+		hr.Cache = cache.New(cache.DefaultExpiration, 5*time.Minute)
+	}
+	rst, found := hr.Cache.Get(host)
+	if found {
+		result = strings.Split(rst.(string), ",")
+	} else {
+		config, _ := dns.ClientConfigFromFile(hr.ResolvConfig)
+		c := new(dns.Client)
```
nmengin: In the way to be homogeneouse with all the code, can you replace this line and the following one by `c := dns.Client{Timeout : 30 * time.Second}` please ?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+		c.Timeout = 30 * time.Second
+		m := new(dns.Msg)
```
nmengin: In the way to be homogeneouse with all the code, can you replace this line and the following one by `m := dns.Msg{}.SetQuestion(dns.Fqdn(host), dns.TypeCNAME)` please ?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+	var result []string
+	result = append(result, host)
+	if hr.Cache == nil {
+		hr.Cache = cache.New(cache.DefaultExpiration, 5*time.Minute)
+	}
+	rst, found := hr.Cache.Get(host)
+	if found {
+		result = strings.Split(rst.(string), ",")
+	} else {
+		config, _ := dns.ClientConfigFromFile(hr.ResolvConfig)
```
nmengin: Can you manage the error please?

integration/integration_test.go
```diff
@@ -60,6 +60,7 @@ func init() {
 		check.Suite(&TimeoutSuite{})
 		check.Suite(&TracingSuite{})
 		check.Suite(&WebsocketSuite{})
+		check.Suite(&HostResolverSuite{})
```
nmengin: Can you insert your test suite after `HealthCheckSuite` (alphabetical order)?

hostresolver/hostresolver.go
nmengin: Can you manage the error please?

hostresolver/hostresolver.go
nmengin: In the way to be homogeneouse with the rest of the code, can you replace this line and the following by `c := dns.Client{Timeout: 30 * time.Second}` please?

hostresolver/hostresolver.go
nmengin: In the way to be homogeneouse with the rest of the code, can you replace this line and the following by `m := dns.Msg{}.SetQuestion(dns.Fqdn(host), dns.TypeCNAME)` please?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+		c := new(dns.Client)
+		c.Timeout = 30 * time.Second
+		m := new(dns.Msg)
+		m.SetQuestion(dns.Fqdn(host), dns.TypeCNAME)
+		var cacheDuration = 0 * time.Second
+		for i := 0; i < hr.ResolvDepth; i++ {
+			r, _, err := c.Exchange(m, net.JoinHostPort(config.Servers[0], config.Port))
+			if err != nil {
+				i--
```
nmengin: If the `c.Exchange` method always returns an error, we will be blocked here.
Moreover, can you manage the error (log it or return it) please?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+		result = strings.Split(rst.(string), ",")
+	} else {
+		config, _ := dns.ClientConfigFromFile(hr.ResolvConfig)
+		c := new(dns.Client)
+		c.Timeout = 30 * time.Second
+		m := new(dns.Msg)
+		m.SetQuestion(dns.Fqdn(host), dns.TypeCNAME)
+		var cacheDuration = 0 * time.Second
+		for i := 0; i < hr.ResolvDepth; i++ {
+			r, _, err := c.Exchange(m, net.JoinHostPort(config.Servers[0], config.Port))
```
nmengin: Not sure to understand, do you want to always test the exchange with the first server set in the resolv file?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+			if err != nil {
+				i--
+				continue
+			}
+			if r != nil && len(r.Answer) > 0 {
+				temp := strings.Fields(r.Answer[0].String())
+				str := strings.TrimSuffix(strings.TrimSpace(temp[len(temp)-1]), ".")
+				result = append(result, str)
+				if i == 0 {
```
nmengin: Can you explain why do we override the TTL only if the answer comes from the first level please?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+					ttl, err := strconv.Atoi(temp[1])
+					if err != nil {
+						cacheDuration = time.Duration(ttl) * time.Second
```
nmengin: Can you log the error please?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+					}
+				}
+				m.SetQuestion(dns.Fqdn(str), dns.TypeCNAME)
```
nmengin: Cannot we `break` the loop after this instruction?

hostresolver/hostresolver_test.go
```diff
@@ -0,0 +1,23 @@
+package hostresolver
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestRules_CNAMEFlatten(t *testing.T) {
+	hostResolver := NewDefaultHostResolver()
```
nmengin: Can you change the test to make it similat to [these tests](https://github.com/containous/traefik/blob/master/provider/acme/provider_test.go)?

We are currently doing a refactoring of our tests in this way.

rules/rules.go
```diff
@@ -26,9 +28,20 @@ func (r *Rules) host(hosts ...string) *mux.Route {
...
 		}
-		for _, host := range hosts {
-			if types.CanonicalDomain(reqHost) == types.CanonicalDomain(host) {
-				return true
+		if r.HostResolver != nil && r.HostResolver.Enabled {
+			reqH, flatH := r.HostResolver.CNAMEFlatten(types.CanonicalDomain(reqHost))
+			for _, host := range hosts {
+				if types.CanonicalDomain(reqH) == types.CanonicalDomain(host) {
+					return true
+				} else if types.CanonicalDomain(flatH) == types.CanonicalDomain(host) {
```
nmengin: Can you remove the `else` statement which is not mandatory after a `return` please?
Only use another `if` statement.

server/server.go
```diff
@@ -1559,3 +1561,11 @@ func buildModifyResponse(secure *secure.Secure, header *middlewares.HeaderStruct
 		return nil
 	}
 }
+
+// buildHostResolver build host resolver, if no config, create default
+func buildHostResolver(globalConfig configuration.GlobalConfiguration) *hostresolver.HostResolver {
+	if globalConfig.HostResolver != nil {
+		return hostresolver.NewHostResolver(globalConfig.HostResolver.CnameFlattening, globalConfig.HostResolver.ResolvConfig, globalConfig.HostResolver.ResolvDepth)
+	}
+	return hostresolver.NewDefaultHostResolver()
```
nmengin: Can you return `nil` please?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+// HostResolver used for host resolver
+type HostResolver struct {
+	Enabled      bool
+	Cache        *cache.Cache
+	ResolvConfig string
+	ResolvDepth  int
+}
+
+// NewHostResolver init host resolver
+func NewHostResolver(enabled bool, resConfig string, resDepth int) *HostResolver {
```
nmengin: In the way to be homogeneouse ith the rest of the code, can you delete the `NewHostResolver` and initialize the structure directly please?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,82 @@
...
+	return &HostResolver{
+		Enabled:      enabled,
+		ResolvConfig: resConfig,
+		ResolvDepth:  resDepth,
+		Cache:        cache.New(cache.DefaultExpiration, 5*time.Minute),
+	}
+}
+
+// NewDefaultHostResolver init default host resolver
+func NewDefaultHostResolver() *HostResolver {
```
nmengin: Can you delete this function please?
Refer [here](https://github.com/containous/traefik/pull/3403/files#diff-91bbeda7eb98a7adc57b9e47e2cf5c2bR1570) for more details.

nmengin:
@gamalan I just foud this [Pull Request](https://github.com/xenolf/lego/pull/506).

Do you think it can help for the CNAME flattening?

gamalan:
@nmengin thanks for the feedback, I'll check it out and update it next week after Eid Al-fitr day. Currently, I'm unable to fully check it.

hostresolver/hostresolver.go
gamalan: Assuming the record is like 
```bind
sub.userA.com TTL IN CNAME host.domain.com
host.domain.com TTL IN CNAME server.domainB.com
```
Whilst the one we serve is server.domainB.com. The CNAMEFlatten will result sub.userA.com mapped to server.domainB.com, and because it will be cached, then we use the first record highest TTL. The TTL of host.domain.com will be ignored, because if it changed(not pointing to server.domainB.com), then we won't need to serve sub.userA.com anymore.

gamalan:
@nmengin looking at the [pull request](https://github.com/xenolf/lego/pull/506). I don't think it's solve the problem. Because traefik only able to change the dns that it's managed right? CMIIW.

docs/configuration/commons.md
```diff
@@ -415,6 +415,35 @@ If no units are provided, the value is parsed assuming seconds.
...
+
+# resolvDepth is the maximum CNAME recursive lookup
+#
+# Optional
+# Default : 5
+resolvDepth = 5
+
+```
+
+- To allow serving secure https request and generate the SSL using ACME while `cnameFlattening` is set to `true`. The `acme` configuration
```
nmengin: Please only write one sentence by line.

docs/configuration/commons.md
```diff
@@ -415,6 +415,35 @@ If no units are provided, the value is parsed assuming seconds.
...
+# resolvDepth is the maximum CNAME recursive lookup
+#
+# Optional
+# Default : 5
+resolvDepth = 5
+
+```
+
+- To allow serving secure https request and generate the SSL using ACME while `cnameFlattening` is set to `true`. The `acme` configuration 
+for `HTTP-01` challenge and `onDemand` is mandatory. Refer to [ACME configuration](/configuration/acme) for more information.
```
nmengin: Please only write one sentence by line.

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,89 @@
...
+	ResolvDepth  int
+}
+
+// CNAMEResolv used for storing CNAME result
+type CNAMEResolv struct {
+	TTL    int
+	Record string
+}
+
+// CNAMEFlatten check if CNAME records is exists, flatten if possible
```
nmengin: > CNAMEFlatten check if CNAME records is exists,

Please modify in `CNAMEFlatten check if CNAME records exists`.

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,89 @@
...
+			} else {
+				break
+			}
+		}
+		hr.Cache.Add(host, strings.Join(result, ","), cacheDuration)
+	}
+	return result[0], result[len(result)-1]
+}
+
+// CNAMEResolve resolve CNAME if exists, and return with the highest TTL
```
nmengin: Can you correct the comment please: CNAMEResolve resolve**s** CNAME if exists, and return**s** with the highest TTL ?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,89 @@
...
+func (hr *HostResolver) CNAMEResolve(host string) *CNAMEResolv {
+	config, _ := dns.ClientConfigFromFile(hr.ResolvConfig)
```
nmengin: Can you manage the error please?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,89 @@
...
+	c := dns.Client{Timeout: 30 * time.Second}
+	c.Timeout = 30 * time.Second
+	m := &dns.Msg{}
+	m.SetQuestion(dns.Fqdn(host), dns.TypeCNAME)
+	var result []*CNAMEResolv
+	for i := 0; i < len(config.Servers); i++ {
+		r, _, err := c.Exchange(m, net.JoinHostPort(config.Servers[i], config.Port))
+		if err != nil {
+			log.Errorf("Failed to resolve host %s with server %s", host, config.Servers[i])
+		}
```
nmengin: Do you want to contin ue to precces the statement in the loop if there is an error?
WDYT to add a instruction`continue` if `err != nil`?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,89 @@
...
+				break
+			}
+		}
+		hr.Cache.Add(host, strings.Join(result, ","), cacheDuration)
+	}
+	return result[0], result[len(result)-1]
+}
+
+// CNAMEResolve resolve CNAME if exists, and return with the highest TTL
+func (hr *HostResolver) CNAMEResolve(host string) *CNAMEResolv {
```
nmengin: The method seems to be used only on this file.
WDYT to make it private?

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,94 @@
...
+
+// HostResolver used for host resolver
+type HostResolver struct {
+	Enabled      bool
+	Cache        *cache.Cache
+	ResolvConfig string
+	ResolvDepth  int
+}
+
+// CNAMEResolv used for storing CNAME result
```
nmengin: Please replace `for storing` by `to store`

hostresolver/hostresolver.go
```diff
@@ -0,0 +1,94 @@
...
+type CNAMEResolv struct {
+	TTL    int
+	Record string
+}
+
+// CNAMEFlatten check if CNAME records exists, flatten if possible
```
nmengin: Do you mean `records exist` or `record exists`?

integration/hostresolver_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	err := cmd.Start()
+	c.Assert(err, checker.IsNil)
+	defer cmd.Process.Kill()
+
+	req, err := http.NewRequest(http.MethodGet, "http://127.0.0.1:8000/", nil)
+	c.Assert(err, checker.IsNil)
+	req.Host = "frontend1.docker.local"
+
+	err = try.Request(req, 500*time.Millisecond, try.StatusCodeIs(http.StatusOK), try.HasBody())
```
nmengin: Not sure, but this test only checks if the usual behavior is the same if the `cnameflattening` is enabled but it does not test if this one really works, it isn't?

Is it possible to add a test to check if CNAME flattening works or is it hard to check?

rules/rules.go
```diff
@@ -26,9 +28,18 @@ func (r *Rules) host(hosts ...string) *mux.Route {
...
-		for _, host := range hosts {
-			if types.CanonicalDomain(reqHost) == types.CanonicalDomain(host) {
-				return true
+		if r.HostResolver != nil && r.HostResolver.Enabled {
+			reqH, flatH := r.HostResolver.CNAMEFlatten(types.CanonicalDomain(reqHost))
+			for _, host := range hosts {
+				if types.CanonicalDomain(reqH) == types.CanonicalDomain(host) || types.CanonicalDomain(flatH) == types.CanonicalDomain(host) {
+					return true
+				}
+			}
```
nmengin: WDYT to add a `DEBUG` log to indicate to user that the `CNAMEFlattening` did not allow resolving his domain?

gamalan:
@nmengin there is a failed build on Semaphore, but it got weird output. This [build](https://semaphoreci.com/containous/traefik/branches/pull-request-3403/builds/16) the integration test is successful but the unit test is failed when generating cross-binary. But, in this [build](https://semaphoreci.com/containous/traefik/branches/pull-request-3403/builds/17) the unit test is successful, yet the integration is failed. Both builds are using same commit points.

PR #3404, opened 2018-05-28
Create Global Backend Ingress

### LINKED ISSUE #1051: Support for default route for k8s backend
Traefik current doesn't recognize default backend of kubernetes ingress, showing 404 instead:

```yaml
kind: Ingress
metadata:
  namespace: deis
  name: deis
spec:
  backend:
    serviceName: deis-router
    servicePort: 80
```

Is it reasonable to configure catch-all in traefik if following ingress is present? I'd say so.

dtomcej:
### What does this PR do?

Allows for an ingress with a Global Parameter Backend to create a default backend route.

### Motivation

Fixes #1051 , and also allows for traefik to support web frameworks that use global defaults.

### More

- [x] Added/updated tests

### Additional Notes

Priority is set based on the ingress priority. Enabling this without setting priority on other ingresses may run into issues with overlapping rules...May want to discuss moving the default priority for kubernetes ingresses to 10, to allow for the default route to be 5, and still give some flexibility.


provider/kubernetes/kubernetes.go
```diff
@@ -188,6 +188,103 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					templateObjects.Backends[defaultBackendName] = &types.Backend{
+						Servers: make(map[string]types.Server),
+						LoadBalancer: &types.LoadBalancer{
+							Method: "wrr",
+						},
+					}
+
+					service, exists, err := k8sClient.GetService(i.Namespace, i.Spec.Backend.ServiceName)
+					if err != nil {
+						log.Errorf("Error while retrieving service information from k8s API %s/%s: %v", i.Namespace, i.Spec.Backend.ServiceName, err)
```
nmengin: Because the error seems to be logged in a higher level function, WDYT to replace the `log.Errorf` by a `fmt.Errorf` and return this new error without logging it?

provider/kubernetes/kubernetes.go
```diff
@@ -188,6 +188,103 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					templateObjects.Backends[defaultBackendName].Buffering = getBuffering(service)
+
+					endpoints, exists, err := k8sClient.GetEndpoints(service.Namespace, service.Name)
+					if err != nil {
+						log.Errorf("Error retrieving endpoints %s/%s: %v", service.Namespace, service.Name, err)
+						return nil, err
+					}
+					if !exists {
+						log.Warnf("Endpoints not found for %s/%s", service.Namespace, service.Name)
+						break
```
nmengin: If the service doest not exist, an `error` is logged and the `continue` instruction is called.
Can you explain to me why is there a `Warn` log and a `break` instruction when the entrypoint does not exist please?

provider/kubernetes/kubernetes.go
```diff
@@ -188,6 +188,103 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					}
+					if len(endpoints.Subsets) == 0 {
+						log.Warnf("Endpoints not available for %s/%s", service.Namespace, service.Name)
+						break
```
nmengin: If the service doest not exist, an `error` is logged and the `continue` instruction is called.
Can you explain to me why is there a `Warn` log and a `break` instruction when the `entrypoint.subset` does not exist please?

provider/kubernetes/kubernetes.go
```diff
@@ -188,6 +188,103 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					}
+
+					for _, subset := range endpoints.Subsets {
+						endpointPort := i.Spec.Backend.ServicePort
```
nmengin: Are we sure the `i.Spec.Backend.ServicePort` is never nil or empty?

provider/kubernetes/kubernetes.go
dtomcej: Service not existing is usually due to someone mistyping something into an ingress hence the `Error` being logged, and the `continue` skips processing the rest of that ingress, and continues processing the next ingress.

The warn and break are due to endpoints not being found. This is a _very_ common occurrence, as k8s pods are created and killed. These are not user-error prone, and the rest of the ingress can still be processed if there are no endpoints found.

provider/kubernetes/kubernetes.go
dtomcej: yes

dtomcej: Will test.

dtomcej: yes

provider/kubernetes/kubernetes.go
dtomcej: Yes. On second thought, that makes sense

docs/configuration/backends/kubernetes.md
```diff
@@ -283,3 +283,22 @@ More information are available in the  [User Guide](/user-guide/kubernetes/#add-
...
+spec:
+  backend:
+    serviceName: stilton
+    servicePort: 80
+```
+This ingress follows the [Global Default Backend](https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource) property of ingresses.
+This will allow users to create a "default backend" that will match all unmatched requests.
+
+!!! note
+    Due to traefik's use of priorities, you may have to set this ingress priority lower than other ingresses in your environment, to avoid this global ingress from satisfying requests that _could_ match other ingresses.
```
nmengin: s/traefik/Træfik please :wink:

docs/configuration/backends/kubernetes.md
```diff
@@ -283,3 +283,22 @@ More information are available in the  [User Guide](/user-guide/kubernetes/#add-
 !!! note
     Only TLS certificates provided by users can be stored in Kubernetes Secrets.
     [Let's Encrypt](https://letsencrypt.org) certificates cannot be managed in Kubernets Secrets yet.
+
+### Global Default Backend Ingresses
+
+Ingresses can be created that look like the following:
+```yaml
```
nmengin: Can you add an empty line before and after the `yaml` example please?

provider/kubernetes/kubernetes.go
```diff
@@ -190,6 +190,105 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					templateObjects.Backends[defaultBackendName] = &types.Backend{
+						Servers: make(map[string]types.Server),
+						LoadBalancer: &types.LoadBalancer{
+							Method: "wrr",
+						},
+					}
+
+					service, exists, err := k8sClient.GetService(i.Namespace, i.Spec.Backend.ServiceName)
+					if err != nil {
+						return nil, fmt.Errorf("Error while retrieving service information from k8s API %s/%s: %v", i.Namespace, i.Spec.Backend.ServiceName, err)
```
nmengin: Please downcase the `Error` word in the error message.

provider/kubernetes/kubernetes.go
```diff
@@ -190,6 +190,105 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+					}
+
+					templateObjects.Backends[defaultBackendName].CircuitBreaker = getCircuitBreaker(service)
+					templateObjects.Backends[defaultBackendName].LoadBalancer = getLoadBalancer(service)
+					templateObjects.Backends[defaultBackendName].MaxConn = getMaxConn(service)
+					templateObjects.Backends[defaultBackendName].Buffering = getBuffering(service)
+
+					endpoints, exists, err := k8sClient.GetEndpoints(service.Namespace, service.Name)
+					if err != nil {
+						return nil, fmt.Errorf("Error retrieving endpoint information from k8s API %s/%s: %v", service.Namespace, service.Name, err)
```
nmengin: Please downcase the `Error` word in the error message.

provider/kubernetes/kubernetes.go
```diff
@@ -264,8 +363,7 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 
 				service, exists, err := k8sClient.GetService(i.Namespace, pa.Backend.ServiceName)
 				if err != nil {
-					log.Errorf("Error while retrieving service information from k8s API %s/%s: %v", i.Namespace, pa.Backend.ServiceName, err)
-					return nil, err
+					return nil, fmt.Errorf("Error while retrieving service information from k8s API %s/%s: %v", i.Namespace, pa.Backend.ServiceName, err)
```
nmengin: Please downcase the `Error` word in the error message.

provider/kubernetes/kubernetes.go
```diff
@@ -300,8 +398,7 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 						} else {
 							endpoints, exists, err := k8sClient.GetEndpoints(service.Namespace, service.Name)
 							if err != nil {
-								log.Errorf("Error retrieving endpoints %s/%s: %v", service.Namespace, service.Name, err)
-								return nil, err
+								return nil, fmt.Errorf("Error retrieving endpoint information from k8s API %s/%s: %v", service.Namespace, service.Name, err)
```
nmengin: Please downcase the `Error` word in the error message.

provider/kubernetes/kubernetes_test.go
```diff
@@ -1449,8 +1513,20 @@ func TestKubeAPIErrors(t *testing.T) {
 
 			provider := Provider{}
 
-			if _, err := provider.loadIngresses(client); err != apiErr {
-				t.Errorf("Got error %v, wanted error %v", err, apiErr)
+			if _, err := provider.loadIngresses(client); err != nil {
+				switch {
+				case client.apiServiceError != nil:
+					expected := errors.New("Error while retrieving service information from k8s API testing/service1: failed kube api call")
```
nmengin: Please downcase the `Error` word in the error message.

provider/kubernetes/kubernetes_test.go
```diff
@@ -1449,8 +1513,20 @@ func TestKubeAPIErrors(t *testing.T) {
...
+					if err.Error() != expected.Error() {
+						t.Errorf("Got error: %+v, wanted error: %+v, provided: %v", err, expected, client.apiServiceError)
+					}
+
+				case client.apiEndpointsError != nil:
+					expected := errors.New("Error retrieving endpoint information from k8s API testing/service1: failed kube api call")
```
nmengin: Please downcase the `Error` word in the error message.

docs/configuration/backends/kubernetes.md
dtomcej: We should add this as a linter, so they don't get missed.

provider/kubernetes/kubernetes.go
```diff
@@ -190,161 +192,214 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+		}
+		if !exists {
+			log.Warnf("Endpoints not found for %s/%s", service.Namespace, service.Name)
+		}
+		if len(endpoints.Subsets) == 0 {
+			log.Warnf("Endpoints not available for %s/%s", service.Namespace, service.Name)
+		}
+
+		for _, subset := range endpoints.Subsets {
+			endpointPort := endpointPortNumber(corev1.ServicePort{Protocol: "TCP", Port: int32(i.Spec.Backend.ServicePort.IntValue())}, subset.Ports)
```
marco-jantke: It seems to be still valid to specify the service port as string, referencing the actual service port. Therefore, I think we have to write a second function that does that logic for the ingress service port. WDYT?

provider/kubernetes/kubernetes.go
```diff
@@ -190,161 +192,214 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
-				if _, exists := templateObjects.Backends[baseName]; !exists {
-					templateObjects.Backends[baseName] = &types.Backend{
-						Servers: make(map[string]types.Server),
-						LoadBalancer: &types.LoadBalancer{
-							Method: "wrr",
-						},
-					}
+			protocol := "http"
+			for _, address := range subset.Addresses {
+				if endpointPort == 443 {
```
marco-jantke: Do we also have to extend this check for the `https` prefix in the port name? Goes together with the comment above.

provider/kubernetes/kubernetes.go
```diff
@@ -736,3 +791,58 @@ func getRateLimit(i *extensionsv1beta1.Ingress) *types.RateLimit {
 
 	return rateLimit
 }
+
+func createEmptyBackend(name string, templateObjects types.Configuration) (types.Configuration, error) {
+	if _, exists := templateObjects.Backends[name]; !exists {
```
marco-jantke: If you turn around the logic we could save one level of indent :) The comment below would also become obsolete as the error message documents the if then. The same holds true for the functions below.

marco-jantke:
That was a tough one for me to review, but it's a nice feature and a nice PR as well :-) I left a couple of comments.

Regarding the priority: it's definitely a tricky area. I'm not sure what is the desired way at the moment. I have to get a clearer picture before I can say something qualified :D

provider/kubernetes/kubernetes.go
dtomcej: You are correct, but its even simpler for this spec! 😄

docs/configuration/backends/kubernetes.md
```diff
@@ -283,3 +283,24 @@ More information are available in the  [User Guide](/user-guide/kubernetes/#add-
...
+  backend:
+    serviceName: stilton
+    servicePort: 80
+```
+
+This ingress follows the [Global Default Backend](https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource) property of ingresses.
+This will allow users to create a "default backend" that will match all unmatched requests.
+
+!!! note
+    Due to Træfik's use of priorities, you may have to set this ingress priority lower than other ingresses in your environment, to avoid this global ingress from satisfying requests that _could_ match other ingresses.
```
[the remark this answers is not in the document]
dtomcej: Done!

provider/kubernetes/kubernetes.go
dtomcej: Tests added!

provider/kubernetes/kubernetes.go
dtomcej: Done!

docs/configuration/backends/kubernetes.md
```diff
@@ -283,3 +283,24 @@ More information are available in the  [User Guide](/user-guide/kubernetes/#add-
...
+  backend:
+    serviceName: stilton
+    servicePort: 80
+```
+
+This ingress follows the [Global Default Backend](https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource) property of ingresses.
+This will allow users to create a "default backend" that will match all unmatched requests.
+
+!!! note
+    Due to Træfik's use of priorities, you may have to set this ingress priority lower than other ingresses in your environment, to avoid this global ingress from satisfying requests that _could_ match other ingresses. To do this, use the `traefik.ingress.kubernetes.io/priority` annotation (as seen in the General Annotations table above) on your ingresses accordingly.
```
nmengin: One sentence per line please.

provider/kubernetes/kubernetes.go
```diff
@@ -270,8 +280,7 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 
 				service, exists, err := k8sClient.GetService(i.Namespace, pa.Backend.ServiceName)
 				if err != nil {
-					log.Errorf("Error while retrieving service information from k8s API %s/%s: %v", i.Namespace, pa.Backend.ServiceName, err)
-					return nil, err
+					return nil, fmt.Errorf("error while retrieving service information from k8s API %s/%s: %v", i.Namespace, pa.Backend.ServiceName, err)
```
timoreimann: Is this change related to what the PR essentially tries to achieve?

I don't mind a good drive-by improvement. What I'm wondering here though is if we had a reason to log the error first. From what I can tell, [this line](https://github.com/dtomcej/traefik/blob/2248a79a27e8c71c8605e609d9b3ad2b84d9ea2f/provider/kubernetes/kubernetes.go#L159) will eventually log the error. I'm not sure if this reads well though as it would show up as

```
Cannot connect to Provider: error while retrieving service information from k8s API [...]
```

We are beyond the connection phase at this point already, aren't we?
How about annotating the [error returned here](https://github.com/dtomcej/traefik/blob/2248a79a27e8c71c8605e609d9b3ad2b84d9ea2f/provider/kubernetes/kubernetes.go#L137) like `return fmt.Errorf("failed to load Ingress: %s", err)` and log the error from line 159 as-in?

Similar for line 318.

provider/kubernetes/kubernetes.go
```diff
@@ -401,6 +409,100 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
 	return configuration
 }
 
+func (p *Provider) addGlobalBackend(k8sClient Client, i *extensionsv1beta1.Ingress, templateObjects *types.Configuration) error {
```
timoreimann: I'd rename `k8sClient` to just `cl`.

provider/kubernetes/kubernetes.go
```diff
@@ -401,6 +409,100 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
...
+	// Ensure that we are not duplicating the frontend
+	if _, exists := templateObjects.Frontends[defaultFrontendName]; exists {
+		return errors.New("duplicate frontend: global-default-backend")
```
timoreimann: You probably mean `global-default-frontend`. Why not use the constant here?

provider/kubernetes/kubernetes.go
```diff
@@ -401,6 +409,100 @@ func (p *Provider) loadConfig(templateObjects types.Configuration) *types.Config
...
+	}
+
+	// Ensure we are not duplicating the backend
+	if _, exists := templateObjects.Backends[defaultBackendName]; exists {
+		return errors.New("duplicate backend: global-default-backend")
```
timoreimann: Constant usage here too?

provider/kubernetes/kubernetes.go
timoreimann: The method name seems incorrect. To the least, the method also populates the frontend, right?

What is the scope of the method, what's it supposed to do? (Its large size _may_ hint at a smell.)

provider/kubernetes/kubernetes.go
dtomcej: Looking at it in that perspective, I like the original logging, then return pattern.

provider/kubernetes/kubernetes.go
dtomcej: The reason I chose this name is in reference to the `backend` section of the spec, which k8s named `global default backend`. If the name bothers us, we can change it in an upcoming refactor :)

timoreimann: Ah okay, missed that -- makes sense to me now, so feel free to keep it. :)

provider/kubernetes/kubernetes.go
```diff
@@ -306,8 +316,7 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 						} else {
 							endpoints, exists, err := k8sClient.GetEndpoints(service.Namespace, service.Name)
 							if err != nil {
-								log.Errorf("Error retrieving endpoints %s/%s: %v", service.Namespace, service.Name, err)
-								return nil, err
+								return nil, fmt.Errorf("error retrieving endpoint information from k8s API %s/%s: %v", service.Namespace, service.Name, err)
```
timoreimann: Do you want to undo this one as well?

PR #3446, opened 2018-06-06
Note that the docker backend does not work with Docker for Mac

Timmmm:
Trust me it doesn't. I wasted so much time on this and then tried the exact same thing on Ubuntu and it worked fine.

Hopefully this will stop other people wasting hours too.

### DIFF
--- a/docs/configuration/backends/docker.md
+++ b/docs/configuration/backends/docker.md
@@ -1,7 +1,7 @@
 
 # Docker Provider
 
-Træfik can be configured to use Docker as a provider.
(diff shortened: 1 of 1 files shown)


Timmmm:
I don't use docker compose, but try this:

1. Create a `traefik.toml` containing:

```
logLevel = "DEBUG"
[entryPoints]
	[entryPoints.http]
		address = ":80"

[docker]
watch = true
```

2. Run a simple docker web server.

    docker run -l traefik.backend=simplehttp -l traefik.enable=true -l traefik.port=8080 -l traefik.frontend.rule=Host:localhost trinitronx/python-simplehttpserver

3. Run traefik.

    traefik --configFile=traefik.toml

4. Connect to http://localhost/

On Mac I get `Bad Gateway`. I'm pretty sure this works fine on Linux (I can't test now unfortunately).

PR #3456, opened 2018-06-07
Remove unnecessary loop

ZloyDyadka:
The cycle here is useless, the new iteration is unattainable

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -115,40 +115,36 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 
 	pool.Go(func(stop chan bool) {
 		operation := func() error {
-			for {
-				stopWatch := make(chan struct{}, 1)
-				defer close(stopWatch)
-				eventsChan, err := k8sClient.WatchAll(p.Namespaces, stopWatch)
-				if err != nil {
-					log.Errorf("Error watching kubernetes events: %v", err)
-					timer := time.NewTimer(1 * time.Second)
-					select {
-					case <-timer.C:
-						return err
-					case <-stop:
-						return nil
-					}
+			stopWatch := make(chan struct{}, 1)
+			defer close(stopWatch)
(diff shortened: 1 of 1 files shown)


provider/kubernetes/kubernetes.go
```diff
@@ -115,40 +115,36 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 
 	pool.Go(func(stop chan bool) {
 		operation := func() error {
+			stopWatch := make(chan struct{}, 1)
+			defer close(stopWatch)
+			eventsChan, err := k8sClient.WatchAll(p.Namespaces, stopWatch)
+			if err != nil {
+				log.Errorf("Error watching kubernetes events: %v", err)
+				timer := time.NewTimer(1 * time.Second)
```
marco-jantke: Do you understand why this is required? I mean to populate the error after a second and not directly?

ZloyDyadka: @marco-jantke Thank you for review! But after error we exit from function, new iteration will not happen

marco-jantke: Yes, I agree with that. I was rather asking out of curiosity. That's why I approved right away even though I added the comment.

Juliens: @marco-jantke In fact, this is because, we don't want to send an error if the `pool.Go` was stopped. If we do that, the `backoff.RetryNotify` will continue to retry, and will not stop the `pool.Go`.
So we put a 1s timer that will not trigger in the case if the stoppedChan is populate.

dtomcej:
@ZloyDyadka  I'm pretty sure I worked this loop out when I refactored the k8s provider in https://github.com/containous/traefik/pull/3382. Can you take a look there, and see if the main provider looks ok?

ZloyDyadka:
@dtomcej I looked at your PR, it solves this problem

ldez:
Closed in favor of #3456

ZloyDyadka:
Actually, because https://github.com/containous/traefik/pull/3382 is not accepted.

PR #3470, opened 2018-06-11
Substitude hardcoded "<namespace>/<name>" with k8s ListerGetter

yue9944882:
### What does this PR do?

Small refactor for kubernetes provider.

Remove those hardcoded `<namespace>/<name>` index and use common functions of `ListerGetter` provided by kubernetes informers.  Or developers might get confused by previous code. 🤔

Note that according to current implementations of kubernetes informers on master branch which haven't been changed for a long time, the error returned by `ListerGetter` can only be "Not Found" error. So, with this PR, the return value by `GetEndpoints`,`GetService`,`GetSecret` will only be:

- Endpoints/Service/Secret, true, nil
- nil, false, "Not Found Error"

### Motivation

For better human readability

### More

- [x] Added/updated tests

### DIFF
--- a/provider/kubernetes/client.go
+++ b/provider/kubernetes/client.go
@@ -198,29 +198,22 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
 // GetService returns the named service from the given namespace.
 func (c *clientImpl) GetService(namespace, name string) (*corev1.Service, bool, error) {
 	service, err := c.factories[c.lookupNamespace(namespace)].Core().V1().Services().Lister().Services(namespace).Get(name)
-	if err != nil {
-		return nil, !kubeerror.IsNotFound(err), err
-	}
-	return service, true, nil
+	exist, err := translateNotFoundError(err)
+	return service, exist, err
 }
 
 // GetEndpoints returns the named endpoints from the given namespace.
 func (c *clientImpl) GetEndpoints(namespace, name string) (*corev1.Endpoints, bool, error) {
 	endpoint, err := c.factories[c.lookupNamespace(namespace)].Core().V1().Endpoints().Lister().Endpoints(namespace).Get(name)
-	if err != nil {
-		return nil, !kubeerror.IsNotFound(err), err
-	}
-	return endpoint, true, nil
+	exist, err := translateNotFoundError(err)
+	return endpoint, exist, err
 }
 
 // GetSecret returns the named secret from the given namespace.
 func (c *clientImpl) GetSecret(namespace, name string) (*corev1.Secret, bool, error) {
-	var secret *corev1.Secret
 	secret, err := c.factories[c.lookupNamespace(namespace)].Core().V1().Secrets().Lister().Secrets(namespace).Get(name)
-	if err != nil {
-		return nil, !kubeerror.IsNotFound(err), err
-	}
-	return secret, true, nil
+	exist, err := translateNotFoundError(err)
+	return secret, exist, err
 }
 
 // lookupNamespace returns the lookup namespace key for the given namespace.
@@ -258,4 +251,16 @@ func eventHandlerFunc(events chan<- interface{}, obj interface{}) {
 	case events <- obj:
 	default:
 	}
+}
+
+// translateNotFoundError will translate a "not found" error to a boolean return
+// value which indicates if the resource exists and a nil error.
+func translateNotFoundError(err error) (bool, error) {
+	if err != nil {
+		if kubeerror.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
 }
--- a/provider/kubernetes/client.go
+++ b/provider/kubernetes/client.go
@@ -170,11 +170,9 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 
 // UpdateIngressStatus updates an Ingress with a provided status.
 func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) error {
-	keyName := namespace + "/" + name
-
 	ing, err := c.factories[c.lookupNamespace(namespace)].Extensions().V1beta1().Ingresses().Lister().Ingresses(namespace).Get(name)
 	if err != nil {
-		return fmt.Errorf("failed to get ingress %s with error: %v", keyName, err)
+		return fmt.Errorf("failed to get ingress %s: %v", namespace+"/"+name, err)
 	}
 
 	if len(ing.Status.LoadBalancer.Ingress) > 0 {
@@ -189,9 +187,9 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
 
 	_, err = c.clientset.ExtensionsV1beta1().Ingresses(ingCopy.Namespace).UpdateStatus(ingCopy)
 	if err != nil {
-		return fmt.Errorf("failed to update ingress status %s with error: %v", keyName, err)
+		return fmt.Errorf("failed to update ingress status %s: %v", namespace+"/"+name, err)
 	}
-	log.Infof("Updated status on ingress %s", keyName)
+	log.Infof("Updated status on ingress %s", namespace+"/"+name)
 	return nil
 }
 
@@ -256,11 +254,8 @@ func eventHandlerFunc(events chan<- interface{}, obj interface{}) {
 // translateNotFoundError will translate a "not found" error to a boolean return
 // value which indicates if the resource exists and a nil error.
 func translateNotFoundError(err error) (bool, error) {
-	if err != nil {
-		if kubeerror.IsNotFound(err) {
-			return false, nil
-		}
-		return false, err
+	if kubeerror.IsNotFound(err) {
+		return false, nil
 	}
-	return true, nil
+	return err == nil, err
 }
--- a/provider/kubernetes/client.go
+++ b/provider/kubernetes/client.go
@@ -9,6 +9,7 @@ import (
 	"github.com/containous/traefik/log"
 	corev1 "k8s.io/api/core/v1"
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	kubeerror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/informers"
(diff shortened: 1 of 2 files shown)


provider/kubernetes/client.go
```diff
@@ -198,32 +195,30 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
...
 
 // GetEndpoints returns the named endpoints from the given namespace.
 func (c *clientImpl) GetEndpoints(namespace, name string) (*corev1.Endpoints, bool, error) {
-	var endpoint *corev1.Endpoints
-	item, exists, err := c.factories[c.lookupNamespace(namespace)].Core().V1().Endpoints().Informer().GetStore().GetByKey(namespace + "/" + name)
-	if item != nil {
-		endpoint = item.(*corev1.Endpoints)
+	endpoint, err := c.factories[c.lookupNamespace(namespace)].Core().V1().Endpoints().Lister().Endpoints(namespace).Get(name)
+	if err != nil {
+		return nil, !kubeerror.IsNotFound(err), err
```
timoreimann: IIUC, we will now return a non-nil error even if this is just about a missing element. However, our [call site](https://github.com/yue9944882/traefik/blob/402ef53199819f6eee1d03fc00430f684dde9e2e/provider/kubernetes/kubernetes.go#L302) still considers non-nil errors as triggers to end Ingress parsing prematurely and back off as there is some presumed issue with the Kubernetes API connection.

I think we still need to translate _not found_ errors into a `nil, false, nil` triple (or move the distinction to the call site).

Similar for other changes you have made.

timoreimann:
Even though upstream returns _not found_ errors only right now, I'd still implement processing of the return value such that any other, additional error values introduced in the future are handled properly.

provider/kubernetes/client.go
yue9944882: > (or move the distinction to the call site)

In order to narrow the scope of changes, I'd prefer the former solution. 😀

yue9944882:
> Even though upstream returns not found errors only right now, I'd still implement processing of the return value such that any other, additional error values introduced in the future are handled properly.

Agree with that. IMHO, handling error values would be better to be done in the call site. Or just returning double (object, error) would be event better to me🧐

timoreimann:
>>Even though upstream returns not found errors only right now, I'd still implement processing of the return value such that any other, additional error values introduced in the future are handled properly.
>Agree with that. IMHO, handling error values would be better to be done in the call site. Or just returning double (object, error) would be even better to me.

Agree to differentiating at the call site as long as we are still able to tell the "not found" and "other error" cases apart where we need to (which is at the point where we decide to either log a warning and continue processing, or log an error and back of).

provider/kubernetes/client.go
```diff
@@ -200,32 +197,33 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
...
-	var endpoint *corev1.Endpoints
-	item, exists, err := c.factories[c.lookupNamespace(namespace)].Core().V1().Endpoints().Informer().GetStore().GetByKey(namespace + "/" + name)
-	if item != nil {
-		endpoint = item.(*corev1.Endpoints)
+	endpoint, err := c.factories[c.lookupNamespace(namespace)].Core().V1().Endpoints().Lister().Endpoints(namespace).Get(name)
+	if err != nil {
+		if kubeerror.IsNotFound(err) {
+			return nil, false, nil
+		}
+		return nil, false, err
```
timoreimann: You only did the distinction for the endpoints now, but we need it for services and all other objects as well where we eventually treat absence and general errors differently with regards to Ingress processing.

provider/kubernetes/client.go
```diff
@@ -171,15 +172,11 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) error {
 	keyName := namespace + "/" + name
 
-	item, exists, err := c.factories[c.lookupNamespace(namespace)].Extensions().V1beta1().Ingresses().Informer().GetStore().GetByKey(keyName)
+	ing, err := c.factories[c.lookupNamespace(namespace)].Extensions().V1beta1().Ingresses().Lister().Ingresses(namespace).Get(name)
 	if err != nil {
 		return fmt.Errorf("failed to get ingress %s with error: %v", keyName, err)
```
timoreimann: I think we should omit the `with error` part now. It used to be necessary to distinguish between the "missing" and "true error" cases, but now the upstream error will include this information.

provider/kubernetes/client.go
```diff
@@ -200,32 +197,23 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
...
 func (c *clientImpl) GetService(namespace, name string) (*corev1.Service, bool, error) {
-	var service *corev1.Service
-	item, exists, err := c.factories[c.lookupNamespace(namespace)].Core().V1().Services().Informer().GetStore().GetByKey(namespace + "/" + name)
-	if item != nil {
-		service = item.(*corev1.Service)
-	}
-	return service, exists, err
+	service, err := c.factories[c.lookupNamespace(namespace)].Core().V1().Services().Lister().Services(namespace).Get(name)
+	exist, err := translateNotFoundError(err)
+	return service, exist, err
```
timoreimann: What do you think about just doing

```go
return service, translateNotFoundError(err)
```

here and elsewhere too?

provider/kubernetes/client.go
```diff
@@ -264,3 +252,15 @@ func eventHandlerFunc(events chan<- interface{}, obj interface{}) {
...
+// value which indicates if the resource exists and a nil error.
+func translateNotFoundError(err error) (bool, error) {
+	if err != nil {
+		if kubeerror.IsNotFound(err) {
+			return false, nil
+		}
+		return false, err
+	}
+	return true, nil
+}
```
timoreimann: I wonder if this is more or less readable:

```go
func translateNotFoundError(err error) (bool, error) {
  if kubeerror.IsNotFound(err) {
    return false, nil
  }

  return err == nil, err
}
```

WDYT?

timoreimann: I think we should add a quick unit test for this function... it's easy to get some boolean wrong while the test should be easy to write.

provider/kubernetes/client.go
yue9944882: @timoreimann I'm afraid my IDE is not allowing me to do so. Is this syntax supported by golang?

ldez: it's not supported

provider/kubernetes/client_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	kubeerror "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"testing"
+)
+
+func TestTranslateNotFoundError(t *testing.T) {
+	testCases := []struct {
+		err            error
```
ldez: could you add  `desc` and put a description of each case.

provider/kubernetes/client_test.go
```diff
@@ -0,0 +1,38 @@
...
+			expectedError:  nil,
+		},
+		{
+			err:            fmt.Errorf("bar error"),
+			expectedExists: false,
+			expectedError:  fmt.Errorf("bar error"),
+		},
+	}
+	for _, testCase := range testCases {
+		exists, err := translateNotFoundError(testCase.err)
```
ldez:
``` go

		test := test
			t.Parallel()

			exists, err := translateNotFoundError(test.err)
			assert.Equal(t, test.expectedExists, exists)
			assert.Equal(t, test.expectedError, err)
		})
	}
```

provider/kubernetes/client.go
```diff
@@ -192,40 +187,31 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
 
 	_, err = c.clientset.ExtensionsV1beta1().Ingresses(ingCopy.Namespace).UpdateStatus(ingCopy)
 	if err != nil {
-		return fmt.Errorf("failed to update ingress status %s with error: %v", keyName, err)
+		return fmt.Errorf("failed to update ingress status %s: %v", namespace+"/"+name, err)
 	}
-	log.Infof("Updated status on ingress %s", keyName)
+	log.Infof("Updated status on ingress %s", namespace+"/"+name)
```
```go
log.Infof("Updated status on ingress %s/%s", namespace,name)
```

provider/kubernetes/client.go
```diff
@@ -192,40 +187,31 @@ func (c *clientImpl) UpdateIngressStatus(namespace, name, ip, hostname string) e
 
 	_, err = c.clientset.ExtensionsV1beta1().Ingresses(ingCopy.Namespace).UpdateStatus(ingCopy)
 	if err != nil {
-		return fmt.Errorf("failed to update ingress status %s with error: %v", keyName, err)
+		return fmt.Errorf("failed to update ingress status %s: %v", namespace+"/"+name, err)
```
```go
return fmt.Errorf("failed to update ingress status %s/%s: %v", namespace, name, err)
```

provider/kubernetes/client.go
```diff
@@ -169,17 +170,11 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
...
-	keyName := namespace + "/" + name
-
-	item, exists, err := c.factories[c.lookupNamespace(namespace)].Extensions().V1beta1().Ingresses().Informer().GetStore().GetByKey(keyName)
+	ing, err := c.factories[c.lookupNamespace(namespace)].Extensions().V1beta1().Ingresses().Lister().Ingresses(namespace).Get(name)
 	if err != nil {
-		return fmt.Errorf("failed to get ingress %s with error: %v", keyName, err)
-	}
-	if !exists {
-		return fmt.Errorf("failed to update ingress %s because it does not exist", keyName)
+		return fmt.Errorf("failed to get ingress %s: %v", namespace+"/"+name, err)
```
```go
return fmt.Errorf("failed to get ingress %s/%s: %v", namespace, name, err)
```

provider/kubernetes/client.go
timoreimann: true, my bad.

provider/kubernetes/client_test.go
```diff
@@ -0,0 +1,38 @@
+package kubernetes
+
+import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	kubeerror "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"testing"
```
ldez:
```go
import (
	"fmt"
	"testing"

	"github.com/stretchr/testify/assert"
	kubeerror "k8s.io/apimachinery/pkg/api/errors"
	"k8s.io/apimachinery/pkg/runtime/schema"
)
```

Issue #3449, opened 2018-06-06
Configuring default value for traefik.docker.network


JoyceBabu:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?
I was told from the Slack channel that it is not possible to define a default value for `traefik.docker.network` via toml configuration file.

I request you to add a new configuration parameter in the toml file for setting the default value for the docker network.

nmengin:
Seems to be solved by #3471.

I close the issue.

PR #3471, opened 2018-06-11
Add a default value for the docker.network configuration

jbdoumenjou:
### What does this PR do?

Add a default value for the traefik.docker.network configuration.

### Motivation

Implement the #3449 

### More

- [x] Added/updated tests

### Additional Notes

I can't test it due to the limitation of the docker test library (we can't manipulate the networks).


docs/configuration/backends/docker.md
```diff
@@ -71,6 +71,12 @@ usebindportip = true
 #
 swarmMode = false
 
+# Define a default docker network to use for connections to all containers.
```
emilevauge: Maybe we should add the fact that it's possible to override this with the label `traefik `traefik.docker.network`

provider/docker/docker.go
```diff
@@ -45,6 +45,7 @@ type Provider struct {
 	ExposedByDefault      bool             `description:"Expose containers by default" export:"true"`
 	UseBindPortIP         bool             `description:"Use the ip address from the bound port, rather than from the inner network" export:"true"`
 	SwarmMode             bool             `description:"Use Docker on Swarm Mode" export:"true"`
+	Network               string           `description:"Default Docker network used" export:"true"`
```
emilevauge: Shouldn't we call it `DefaultNetwork` instead?

ldez: To be homogeneous with `domain`, we need to keep `Network`.

provider/docker/docker.go
```diff
@@ -45,6 +45,7 @@ type Provider struct {
 	ExposedByDefault      bool             `description:"Expose containers by default" export:"true"`
 	UseBindPortIP         bool             `description:"Use the ip address from the bound port, rather than from the inner network" export:"true"`
 	SwarmMode             bool             `description:"Use Docker on Swarm Mode" export:"true"`
+	DefaultNetwork        string           `description:"Default Docker network used" export:"true"`
```
ldez: To be homogeneous with `domain`, we need to keep `Network`.

emilevauge: :grin:

jbdoumenjou: Are you both ok with Network ? :)

docs/configuration/backends/docker.md
jbdoumenjou: done

Issue #2985, opened 2018-03-08
Incorrect frontend modifiers in Kubernetes


calummoore:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?
Create an Ingress for Kubernetes and add a path prefix to the request using `ReplacePathRegex` rule - `"ReplacePathRegex: ^/(.*) /api/v1/$1"`.

I also tried using AddPrefix - `"AddPrefix: /api/v1`.

```yml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: server-api
  namespace: graphlog
  annotations:
    kubernetes.io/ingress.class: traefik
    traefik.frontend.rule.type: "ReplacePathRegex: ^/(.*) /api/v1/$1"
spec:
  rules:
  - host: api.graphlog.co
    http:
      paths:
      - path: /
        backend:
          serviceName: server-core
          servicePort: http
```

### What did you expect to see?

Forward all requests from `https://api.graphlog.co` to server-core with prefixed path `/api/v1`

`https://api.graphlog.co/events` -> `/api/v1/events`

### What did you see instead?

**`ReplacePathRegex: ^/(.*) /api/v1/$1`**
Additional `:/` is added to the end of the forwarded path:
`https://api.graphlog.co/events` -> `/api/v1/events:/`

**`AddPrefix: /api/v1`**
Additional `:/` is added between the prefix and the path
`/api/v1://events`

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.5.3
Codename:     cancoillotte
Go version:   go1.9.4
Built:        2018-02-27_02:47:04PM
OS/Arch:      linux/amd64
```

dtomcej:
Hello @calummoore, sorry for the delayed response on this one.

You are correct, this is a bug due to the way our rules logic checks annotations.

Our original ruleset did not have any replacement or addition rules, so the code in this case was not designed to handle them.

A PR will be coming shortly to address this.

ldez:
`traefik.frontend.rule.type` can be use to define the type of a rule: `PathPrefix|PathPrefixStrip|...` but not the rule it-self

calummoore:
The rule seems to do something though? 

Also, how do you set the rule, if `traefik.frontend.rule.type` is only for defining the type of rule?

alexquintero:
[...]

PR #3473, opened 2018-06-11
Allow Modifier type to be changed via annotation in a Kubernetes Ingress

alexquintero:
### What does this PR do?

This allows you to specify ReplacePathRegex rewrites using a Kubernetes Ingress backend. It works by creating an annotation (`traefik.ingress.kubernetes.io/modifier-type: ReplacePathRegex`) that overrides the default of ‘ReplacePath’.

### Motivation

There was no way to specify these kinds of rewrite rules using Kubernetes Ingress backends. We needed this to preserve some compatibility between old clients and new APIs hosted in Kubernetes.

### More

- [x] Added/updated tests

Potentially solves [Issue 2985](https://github.com/containous/traefik/issues/2985).


dtomcej:
Hello @alexquintero,

Thanks for your contribution!

Unfortunately, rewrite-target is not an annotation that can have its behavior switched, as it will cause other issues with compatability.

However, the need for matchers is indeed important, and I have created a PR (https://github.com/containous/traefik/pull/3481) that will enable _all_ the matchers.

Once that PR is merged, we will close this one. Please take a look, and give feedback on that PR if you wish.

ldez:
Closed in favor of  #3481

Issue #3230, opened 2018-04-24
Traefik with Kubernetes Ingress is unable to prepend uri path


trondhindenes:
*bug*

Given this config:
```yml
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: util-testproxy-v0-prod-main-httpsecure
  namespace: volum
  annotations:
     traefik.frontend.rule.type: PathPrefix
     traefik.frontend.passHostHeader: "true"
     ingress.kubernetes.io/rewrite-target: /integration.test/2/
  labels:
    expose: external-httpsecure
spec:
  rules:
  - host: test-api.api.domain.no
    http:
      paths:
      - path: /
        backend:
          serviceName: util-testproxy-v0-prod-main
          servicePort: http

```
### What did you expect to see?
I would expect Traefik to to rewrite urls as such:
original url: `https://test-api.api.domain.no/things` --> url sent to backend: `http://<backend-ip>/integration.test/2/things`

### What did you see instead?
Instead, the original uri stem is simply stripped away, so the actual request sent to the backend becomes:
`http://<backend-ip>/integration.test/2/`

This isn't problematic in itself, but it seems that using Traefik with Kubernetes-ingress there's no way to prepend a "base path" to the request, while retaining (appending) the original uri.

Current functionality aligns well with the nginx ingress `nginx.ingress.kubernetes.io/rewrite-target`. However, nginx has the `nginx.ingress.kubernetes.io/add-base-url` annotation for prepending urls. It seems that Traefik is lacking this functionality.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
1.5.2
```

giggio:
I can confirm the problem with `1.6.3`.

qct:
any way to achieve this goal before the issue being fixed ?

qct:
I found the solution described here: [gh-863](https://github.com/containous/traefik/issues/863):
```
annotations:
kubernetes.io/ingress.class: traefik
traefik.frontend.rule.type: ReplacePathRegex

 - backend:
     serviceName: aServer
     servicePort: 80
   path: /prefix/aServer/(.*) /aServer/$1
 - backend:
     serviceName: aSecondServer
     servicePort: 80
   path: /prefix/aSecondServer/(.*) /aSecondServer/$1
```

trondhindenes:
awesome, that's a good workaround

dtomcej:
[...]

PR #3481, opened 2018-06-12
Add annotation to allow modifiers to be used properly in kubernetes

### LINKED ISSUE #2985: Incorrect frontend modifiers in Kubernetes

dtomcej:
### What does this PR do?

Adds a `request-modifier` annotation to the kubernetes provider, that will allow the modifiers to be used (`AddPrefix, ReplacePath, ReplacePathRegex`) properly, and without inferfering with the matcher rules.

### Motivation

Fixes #2985, #3230 

### More

- [x] Added/updated tests

### Additional Notes

This PR also introduces a warning when users use `ReplacePath` as a rule-type, as it is a modifier, not a matcher rule.


docs/configuration/backends/kubernetes.md
```diff
@@ -156,7 +156,8 @@ The following general annotations are applicable on the Ingress object:
...
 | `traefik.ingress.kubernetes.io/redirect-replacement: http://mydomain/$1`        | Redirect to another URL for that frontend. Must be set with `traefik.ingress.kubernetes.io/redirect-regex`.                                     |
 | `traefik.ingress.kubernetes.io/rewrite-target: /users`                          | Replaces each matched Ingress path with the specified one, and adds the old path to the `X-Replaced-Path` header.                               |
-| `traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip`                      | Override the default frontend rule type. Default: `PathPrefix`.                                                                                 |
+| `traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip`                      | Override the default frontend rule type. Only path related matchers should be used (`Path`, `PathPrefix`, `PathStrip`, `PathPrefixStrip`). Default: `PathPrefix`.                                                                                 |
```
timoreimann: _should_ or rather _can_?

provider/kubernetes/kubernetes.go
```diff
@@ -400,7 +405,24 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
...
+		fallthrough
+	case ruleTypePathPrefix:
+		fallthrough
+	case ruleTypePathStrip:
+		fallthrough
+	case ruleTypePathPrefixStrip:
+		rules = append(rules, ruleType+":"+pa.Path)
+	case ruleTypeReplacePath:
+		rules = append(rules, ruleType+":"+pa.Path)
+		log.Warnf("Using %s as rule-type will be deprecated in the future. Please use request-modifier annotation instead", ruleType)
```
timoreimann: Can we name the full new annotation name?

provider/kubernetes/kubernetes.go
```diff
@@ -421,6 +443,27 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
 		}
 		rules = append(rules, ruleTypeReplacePath+":"+rootPath)
 	}
+
+	if requestModifier := getStringValue(i.Annotations, annotationKubernetesRequestModifier, ""); requestModifier != "" {
```
timoreimann: What happens if the annotation consists of all whitespaces only?

docs/configuration/backends/kubernetes.md
dtomcej: There is crazy legacy code in there that allows for "ReplacePath" to be used as the rule-type. But we want to discourage it (hence the log.warn, and it not being listed in the "should" list), but it can technically be used to maintain backwards compatibility to people already using it.

provider/kubernetes/kubernetes.go
dtomcej: Rude

dtomcej: Also, added tests to include a bunch of other oddities

provider/kubernetes/kubernetes.go
```diff
@@ -421,9 +435,45 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
...
 
+func parseRequestModifier(requestModifier, ruleType string) (string, error) {
+	if strings.Trim(requestModifier, " :") == "" {
+		return "", fmt.Errorf("could not get rule due to incomplete rule: %q", requestModifier)
+	}
+
+	// Split annotation to determine modifier type
+	modifierParts := strings.Split(strings.Trim(requestModifier, " :"), ":")
+	if len(modifierParts) < 2 {
+		return "", fmt.Errorf("could not get rule due to incomplete rule: %q", requestModifier)
```
timoreimann: Can we make this error message different from the one in line 452 to allow distinguishing?

provider/kubernetes/kubernetes.go
```diff
@@ -421,9 +435,45 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
...
+	}
+
+	modifier := strings.TrimSpace(modifierParts[0])
+	value := strings.TrimSpace(modifierParts[1])
+
+	switch modifier {
+	case ruleTypeAddPrefix, ruleTypeReplacePath, ruleTypeReplacePathRegex:
+		if ruleType == ruleTypeReplacePath {
+			return "", fmt.Errorf("cannot use '%s: %s' and '%s: %s'",
```
timoreimann: Why not? Can we include a reason?

docs/configuration/backends/kubernetes.md
```diff
@@ -156,7 +156,8 @@ The following general annotations are applicable on the Ingress object:
...
 | `traefik.ingress.kubernetes.io/redirect-replacement: http://mydomain/$1`        | Redirect to another URL for that frontend. Must be set with `traefik.ingress.kubernetes.io/redirect-regex`.                                     |
 | `traefik.ingress.kubernetes.io/rewrite-target: /users`                          | Replaces each matched Ingress path with the specified one, and adds the old path to the `X-Replaced-Path` header.                               |
-| `traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip`                      | Override the default frontend rule type. Default: `PathPrefix`.                                                                                 |
+| `traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip`                      | Override the default frontend rule type. Only path related matchers can be used (`Path`, `PathPrefix`, `PathStrip`, `PathPrefixStrip`). Default: `PathPrefix`.                                                                                 |
```
timoreimann: How about linking to the path matchers documentation here?

docs/configuration/backends/kubernetes.md
```diff
@@ -156,7 +156,8 @@ The following general annotations are applicable on the Ingress object:
...
+| `traefik.ingress.kubernetes.io/request-modifier: AddPath: /users`               | Add a request modifier to the backend request.                                                                                                  |
```
timoreimann: Link to request modifiers docs from here?

docs/configuration/backends/kubernetes.md
```diff
@@ -156,7 +156,8 @@ The following general annotations are applicable on the Ingress object:
 | `traefik.ingress.kubernetes.io/redirect-regex: ^http://localhost/(.*)`          | Redirect to another URL for that frontend. Must be set with `traefik.ingress.kubernetes.io/redirect-replacement`.                               |
 | `traefik.ingress.kubernetes.io/redirect-replacement: http://mydomain/$1`        | Redirect to another URL for that frontend. Must be set with `traefik.ingress.kubernetes.io/redirect-regex`.                                     |
 | `traefik.ingress.kubernetes.io/rewrite-target: /users`                          | Replaces each matched Ingress path with the specified one, and adds the old path to the `X-Replaced-Path` header.                               |
-| `traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip`                      | Override the default frontend rule type. Default: `PathPrefix`.                                                                                 |
```
timoreimann: Should we leave a note that `ReplacePath` is deprecated?

provider/kubernetes/kubernetes.go
```diff
@@ -400,6 +405,15 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
 	}
 
 	ruleType := getStringValue(i.Annotations, annotationKubernetesRuleType, ruleTypePathPrefix)
+
+	switch ruleType {
+	case ruleTypePath, ruleTypePathPrefix, ruleTypePathStrip, ruleTypePathPrefixStrip:
+	case ruleTypeReplacePath:
+		log.Warnf("Using %s as %s will be deprecated in the future. Please use the %s annotation instead", ruleType, annotationKubernetesRuleType, annotationKubernetesRequestModifier)
+	default:
+		return "", fmt.Errorf("cannot use nonexistent, or non-matcher rule: %q", ruleType)
```
timoreimann: How about adding a `case "":` and log the non-existent case separately? This would be slightly clearer as we don't have to log an empty string.

provider/kubernetes/kubernetes.go
```diff
@@ -421,9 +435,45 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
...
+			return "", err
+		}
+
+		rules = append(rules, rule)
+	}
 	return strings.Join(rules, ";"), nil
 }
 
+func parseRequestModifier(requestModifier, ruleType string) (string, error) {
+	if strings.Trim(requestModifier, " :") == "" {
```
timoreimann: Extracting the `Trim` call into a variable allows reusing in line 456.

Additionally, we don't log any blanks-only value, thus allowing for a clearer error message.

provider/kubernetes/kubernetes.go
```diff
@@ -421,9 +435,45 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
...
+		return "", fmt.Errorf("could not get rule due to empty rule: %q", requestModifier)
```
timoreimann: I'd simplify to `"rule %q is empty"` as the [calling site](https://github.com/dtomcej/traefik/blob/844d02f11372b98dfb1943e312f37675dd81fee6/provider/kubernetes/kubernetes.go#L260) already provides context that the error is about rule retrieval.

provider/kubernetes/kubernetes.go
```diff
@@ -421,9 +435,45 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
...
+	}
+
+	// Split annotation to determine modifier type
+	modifierParts := strings.Split(strings.Trim(requestModifier, " :"), ":")
+	if len(modifierParts) < 2 {
+		return "", fmt.Errorf("could not get rule due to incomplete rule (missing rule type or value): %q", requestModifier)
```
timoreimann: Suggestion to simplify to `"rule %q is missing type or value"`.

provider/kubernetes/kubernetes.go
timoreimann: `TrimRight` should suffice, shouldn't it?

provider/kubernetes/kubernetes.go
```diff
@@ -421,9 +435,45 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
...
+	}
+
+	modifier := strings.TrimSpace(modifierParts[0])
+	value := strings.TrimSpace(modifierParts[1])
+
+	switch modifier {
+	case ruleTypeAddPrefix, ruleTypeReplacePath, ruleTypeReplacePathRegex:
+		if ruleType == ruleTypeReplacePath {
+			return "", fmt.Errorf("cannot use '%s: %s' and '%s: %s, as this leads to rule duplication, and unintended behavior'",
```
timoreimann: Missing closing `'`.

provider/kubernetes/kubernetes.go
```diff
@@ -421,9 +435,45 @@ func getRuleForPath(pa extensionsv1beta1.HTTPIngressPath, i *extensionsv1beta1.I
...
+				annotationKubernetesRuleType, ruleTypeReplacePath, annotationKubernetesRequestModifier, modifier)
+		}
+	default:
+		return "", fmt.Errorf("cannot use nonexistent, or non-modifier rule: %q", modifier)
```
timoreimann: Suggestion for separate empty string case here too.

provider/kubernetes/annotations.go
```diff
@@ -31,6 +31,7 @@ const (
 	annotationKubernetesErrorPages                = "ingress.kubernetes.io/error-pages"
 	annotationKubernetesBuffering                 = "ingress.kubernetes.io/buffering"
 	annotationKubernetesAppRoot                   = "ingress.kubernetes.io/app-root"
+	annotationKubernetesRequestModifier           = "ingress.kubernetes.io/request-modifier"
```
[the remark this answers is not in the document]
dtomcej: Not really.

docs/configuration/backends/kubernetes.md
dtomcej: Not really. If it is "deprecated" and we support it, whats the point of saying its deprecated...

timoreimann: You could argue that we should phrase the log message even stronger: that the old way _will_ not be deprecated but _is_ from now on effectively. This means that we will still support it but plan to remove support in a future release. At least, that's what I remember we (in particular @ldez) did in the past to migrate options.

ldez: I have completely remove annotations with the previous "label" style from the documentation.
I think can do the same here.

provider/kubernetes/annotations.go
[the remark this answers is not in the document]
dtomcej: ^^ Not really as in: I checked, and nginx does not provide that kind of functionality through an annotation. We are providing "new" functionality with this one.

docs/configuration/backends/kubernetes.md
```diff
@@ -154,7 +154,8 @@ The following general annotations are applicable on the Ingress object:
...
 | `traefik.ingress.kubernetes.io/rewrite-target: /users`                          | Replaces each matched Ingress path with the specified one, and adds the old path to the `X-Replaced-Path` header. |
-| `traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip`                      | Override the default frontend rule type. Default: `PathPrefix`.                                                   |
+| `traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip`                      | Override the default frontend rule type. Only path related matchers can be used [(`Path`, `PathPrefix`, `PathStrip`, `PathPrefixStrip`)](/basics/#path-matcher-usage-guidelines). Note: ReplacePath is deprecated in this annotation, use the `traefik.ingress.kubernetes.io/request-modifier` annotation instead. Default: `PathPrefix`.                                                                                 |
+| `traefik.ingress.kubernetes.io/request-modifier: AddPath: /users`               | Add a [request modifier](/basics/#modifiers) to the backend request.                                                                                                  |
```
tompson: It looks like there is a type - should'n it be `AddPrefix` instead of `AddPath`?

ferrarimarco: @tompson thanks. After changing to `AddPrefix` my ingress started working :)

tompson: @ferrarimarco can you maybe share your ingress configuration? I still cannot get mine working

ferrarimarco:
```
kind: Ingress
metadata:
  name: a12
  labels:
    app: a1
    chart: a1
    release: a1
    heritage: Tiller
  annotations:
    traefik.ingress.kubernetes.io/preserve-host: "true"
    traefik.ingress.kubernetes.io/app-root: "/index.html"
    traefik.ingress.kubernetes.io/request-modifier: "AddPrefix: /a1"
spec:
  rules:
    - host: host1.tld
      http:
        paths:
          - path: /test/aaaa
            backend:
              serviceName: service-name
              servicePort: http
```

tompson: Thank you very much! 

Unfortunately I am still not able to get my desired configuration to work.

I got a backend that is available under `/app` and I want a traefik configuration that allows user to call it with `/app` or with `/oldapp` - both ways should work exactly the same.

dtomcej:
```
kind: Ingress
metadata:
  name: default
  annotations:
    traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip
    traefik.ingress.kubernetes.io/request-modifier: "AddPrefix: /app"
spec:
  rules:
  - host: httpecho.kube.troii.net
    http: 
      paths:
      - path: /oldapp
        backend:
          serviceName: echo-server
          servicePort: 80
      - path: /app
        backend:
          serviceName: echo-server
          servicePort: 80
---```

tompson: thanks @dtomcej ! should this already work with v1.6.5 ?

tompson: I looked into the commit history and it seems like this will be in included in 1.7.0 but is not part of 1.6.5 - so I will have to wait to try it

PR #3491, opened 2018-06-13
Add more k8s tests

dtomcej:
### What does this PR do?

Add and fix existing k8s tests

### Motivation

More people using k8s, better tests to assist with refactoring

### More

- [x] Added/updated tests


provider/kubernetes/kubernetes_test.go
```diff
@@ -2163,3 +2222,28 @@ func TestProviderUpdateIngressStatus(t *testing.T) {
 	}
 
 }
+
+func TestProviderNewK8sInClusterClient(t *testing.T) {
+	p := Provider{}
+	os.Setenv("KUBERNETES_SERVICE_HOST", "localhost")
+	os.Setenv("KUBERNETES_SERVICE_PORT", "443")
+	_, err := p.newK8sClient("")
+	os.Clearenv()
```
timoreimann: How about `defer`ing the cleanup right after we have set up the custom environment (i.e., right after line 2229)?

Similar below.

Issue #3293, opened 2018-05-09
Overlapping TLS Certs and Ingress TLS Certs


dtomcej:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

When users have a static (or ACME) wildcard certificate loaded on an entrypoint, and submit an ingress with a TLS Certificate with an exact subdomain match, Traefik will "randomly" serve requests using either certificate.

### What did you expect to see?

The specific certificate provided being used for the host request specified in the ingress spec

### What did you see instead?

Another (albeit matching certificate being used intermittently). This can cause issues with clients that have trusted certificates, or that are not able to handle wildcard certificates.

Issue #3095, opened 2018-03-28
Allow configurable response for non-SNI requests


tomtastic:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you do?

Traefik is responding with its certificate 'TRAEFIK DEFAULT CERT' for non-SNI aware clients,
please provide a configurable option per frontend to set a domainname
certificate to return.

eg.
When a non-SNI aware client requests TLS secured traffic from www.domain.name,
consult something like the below imagined option to respond with the certificate for www.domain.name, rather than the 'TRAEFIK DEFAULT CERT' :

`traefik.frontend.xyz.nonSNICertDomain=www.domain.name`

lucwillems:
hi all ,
i also have currently a issues with a customer which uses tools which doesn't seem to support SNI.
There where complaining about bad "TRAEFIK" certificates which was not allowed by there corporate security policy. The fact that we present a self signed certificate is maybe not so good response.

i would like to see that clients which do not support a valid SNI connection, we log a error about this client (ip address) and drop the connection completely. This way we see this kind of connections error quickly en inform our customer that they need to to use more "modern" tooling /libraries .

my point of view is that if you don't support SNI , which is critical to have many site hosted using single ip , you beter upgrade your tooling or just leave and go fishing. SNI is so old (introduced in RFC3546 , in 2003 ) . if the client doesn't support this , he's better of to not go on the internet , because how well will his stack be protected against all issues currently known .

so something like "drop non SNI client " feature would greatly help this.
something simular like apache has (http://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslstrictsnivhostcheck)

Issue #2874, opened 2018-02-16
traefik occassionally delivers wrong (default) certificate - remote error: tls: unknown certificate authority


dac388:
### Do you want to request a *feature* or report a *bug*?

BUG

### What did you do?

Open frontend url in browser.

### What did you expect to see?

secure webpage

### What did you see instead?

Sporadically, and rarely, the page shows insecure, traefik delivers traefik default cert

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.5.2
Codename:     cancoillotte
Go version:   go1.9.4
Built:        2018-02-12_10:56:31AM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```json
{
 "LifeCycle": null,
 "GraceTimeOut": 0,
 "Debug": true,
 "CheckNewVersion": true,
 "SendAnonymousUsage": false,
 "AccessLogsFile": "",
 "AccessLog": {
  "file": "/var/run/traefik-access.log",
  "format": "common"
 },
 "TraefikLogsFile": "",
 "TraefikLog": {
  "format": "xxxx"
 },
 "LogLevel": "INFO",
 "EntryPoints": {
  "http": {
   "Network": "",
   "Address": "xxxx",
   "TLS": null,
   "Redirect": null,
   "Auth": null,
   "WhitelistSourceRange": null,
   "Compress": false,
   "ProxyProtocol": {
    "Insecure": false,
    "TrustedIPs": []
   },
   "ForwardedHeaders": null
  },
  "https": {
   "Network": "",
   "Address": "xxxx",
   "TLS": {
    "MinVersion": "",
    "CipherSuites": null,
    "Certificates": null,
    "ClientCAFiles": null,
    "ClientCA": {
     "Files": null,
     "Optional": false
    }
   },
   "Redirect": null,
   "Auth": null,
   "WhitelistSourceRange": null,
   "Compress": false,
   "ProxyProtocol": {
    "Insecure": false,
    "TrustedIPs": []
   },
   "ForwardedHeaders": null
  }
 },
 "Cluster": null,
 "Constraints": [],
 "ACME": null,
 "DefaultEntryPoints": [
  "https",
  "http"
 ],
 "ProvidersThrottleDuration": 2000000000,
 "MaxIdleConnsPerHost": 200,
 "IdleTimeout": 0,
 "InsecureSkipVerify": false,
 "RootCAs": null,
 "Retry": null,
[...]

Issue #2118, opened 2017-09-13
Feature Request: Strict-SNI Support


fritz-fritz:
Looking to refuse the connection if SNI does not match a configured frontend.

In haproxy this can be configured by appending `strict-sni` to the bind line as follows:

```
frontend WebTraffic
	bind			127.0.0.1:8443 name 127.0.0.1:8443 ssl ciphers AES128+EECDH:AES128+EDH force-tlsv12 no-sslv3 crt /var/etc/haproxy/WebTraffic.pem crt /var/etc/haproxy/WebTraffic strict-sni
```

this would be desirable instead of serving a 404 with an invalid *.local cert as currently configured. The invalid cert just really bugs me.

ldez:
Thanks for your interest in Traefik 😃 

Duplicate of #1814

fritz-fritz:
I don't believe this is a duplicate. This is not about overriding the default cert it is about dropping the connection before serving it.

lphuberdeau:
There is an other related issue about the SNI implementation.

From RFC6066:

     A server that receives a client hello containing the "server_name"
     extension MAY use the information contained in the extension to guide
     its selection of an appropriate certificate to return to the client,
     and/or other aspects of security policy.  In this event, the server
     SHALL include an extension of type "server_name" in the (extended)
     server hello.  The "extension_data" field of this extension SHALL be
     empty.

Currently, Traefik does not include the server_name in the TLS response. Without this required acknowledgement, some clients skip the SNI extension on follow-up connections.

PR #3512, opened 2018-06-18
Improve TLS Handshake

### LINKED ISSUE #3472: Manage entirely the TLS handshake

Feature

So far, for each HTTPS requests, Træfik checks internaly if a certificate provided dynamically or throught ACME exist.
If there is no certificate, Træfik let the GO HTTPServer implementation find a static certificates.

This behavior has some limitations and few issues are opened in relation with it.

I propose to check all the certificates internaly to manage entirely the behavior and the response.

This issue can solve #3293 #2874 and it can probably fix the flaky tests detected in ACME.

dtomcej:
### What does this PR do?

Improves the TLS handshake to take control of which certificate is returned in response to a clientHello

### Motivation

Fixes #3472, #3293, #2874, #3095, #2118, #1814 

### More

- [x] Enable SNI Strict Support
- [x] Added/updated tests
- [x] Enable Default Certificate Support
- [x] Added/updated tests
- [x] Return best certificate, not first match
- [x] Added/updated tests
- [x] Enable Certificate lookup cache
- [x] Added/updated tests


server/server.go
```diff
@@ -318,6 +345,17 @@ func (s *Server) createTLSConfig(entryPointName string, tlsOption *traefiktls.TL
 		return nil, err
 	}
 
+	if tlsOption.DefaultCertificate != nil {
```
nmengin: Before this code, the method `certificate.CreateTLSConfig` is called [here](https://github.com/containous/traefik/blob/master/server/server.go#L316).

It generates a `by default` self signed certificate if necessayr and add all the static certificates (or the generated one) to the `tls.Config.Certificates` list.
IMHO, the check on the default certificate can be done before to generate a sefl-signed certificate and,in addition, the `tls.Config.Certificates` list has not to be set because you code allows Træfik to manage entirely the `getCertificate` mechanism and the `tls.Config.Certificates` will not be used.

WDYT?

integration/https_test.go
```diff
@@ -21,6 +21,8 @@ import (
 // HTTPSSuite
 type HTTPSSuite struct{ BaseSuite }
 
+const httpTimeout = 1500
```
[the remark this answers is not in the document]
ldez: Personally I do not agree with creating this constant.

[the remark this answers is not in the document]
dtomcej: This was to ensure all tests had enough time to pass in my environment. Easily reverted :)

server/server.go
dtomcej: The problem is that `tlsOption.Certificates.CreateTLSConfig(entryPointName)` is called on the `tls.Config.Certificates` slice. It does not allow for `tls.Config.DefaultCertificate` to be added in if present.

Yes. You are correct. This was from before I handled static certs. This is now irrelevant, and will be removed.

dtomcej: Re-instating this, because I have found out that `getCertificate` does not always get called ;)

tls/certificate_store_test.go
```diff
@@ -0,0 +1,116 @@
...
+				)
+				dynamicMap[test.dynamicCert] = &dynamicCert
+			}
+			store := CertificateStore{
+				DynamicCerts: safe.New(dynamicMap),
+				StaticCerts:  safe.New(staticMap),
+			}
+			var expected *tls.Certificate
+			if test.expectedCert != "" {
+				cert, _ := tls.LoadX509KeyPair(
```
ldez:
```go
func loadTestCert(certName string) (*tls.Certificate, error) {
	staticCert, err := tls.LoadX509KeyPair(
		fmt.Sprintf("../integration/fixtures/https/%s.cert", strings.Replace(certName, "*", "wildcard", -1)),
		fmt.Sprintf("../integration/fixtures/https/%s.key", strings.Replace(certName, "*", "wildcard", -1)),
	)
	if err != nil {
		return nil, err
	}

	return &staticCert, nil
}
```

server/server_configuration.go
```diff
@@ -554,10 +555,31 @@ func (s *Server) buildServerEntryPoints() map[string]*serverEntryPoint {
 			httpRouter:       middlewares.NewHandlerSwitcher(s.buildDefaultHTTPRouter()),
 			onDemandListener: entryPoint.OnDemandListener,
 		}
+
+		serverEntryPoints[entryPointName].certs = &traefiktls.CertificateStore{}
+		serverEntryPoints[entryPointName].certs.StaticCerts = &safe.Safe{}
+		serverEntryPoints[entryPointName].certs.DynamicCerts = &safe.Safe{}
```
ldez:
```go
serverEntryPoints[entryPointName].certs = &traefiktls.CertificateStore{
	StaticCerts : &safe.Safe{},
	DynamicCerts: &safe.Safe{},
}
```

docs/configuration/entrypoints.md
```diff
@@ -341,6 +341,39 @@ To specify an https entry point with a minimum TLS version, and specifying an ar
...
+    [entryPoints.https.tls]
+    [entryPoints.https.tls.defaultCertificate]
+      certFile = "integration/fixtures/https/snitest.com.cert"
+      keyFile = "integration/fixtures/https/snitest.com.key"
+```
+
+!!! note
+    There can only be one `defaultCertificate` set per entrypoint.
+    Use a single set of square brackets `[ ]`, instead of the two needed for normal certificates.
+
```
nmengin: WDYT to add a note to explain that a default self signed certificate is created if there is no static certificate and no defautl certifcates providen?

tls/certificate_store.go
```diff
@@ -31,3 +35,82 @@ func (c CertificateStore) GetAllDomains() []string {
...
+	}
+	if len(matchedCerts) > 0 {
+		//sort map by keys
+		keys := make([]string, 0, len(matchedCerts))
+		for k := range matchedCerts {
+			keys = append(keys, k)
+		}
+		sort.Strings(keys)
+		//cache best match
+		c.CertCache[domainToCheck] = matchedCerts[keys[len(keys)-1]]
```
nmengin: Cannot we check other information?

For example:
- Check if a certificate is self-signed and do not provide it if another signed by a recognized CA
- Check if a certificate is not still valid and do not provide it if another certifcate is still valid.

docs/configuration/entrypoints.md
dtomcej: Sure.

tls/certificate_store.go
dtomcej: We can, but did we want to include that in this PR? or did we want to add more checks etc in a separate PR?

tls/certificate_store_test.go
```diff
@@ -84,33 +84,36 @@ func TestGetBestCertificate(t *testing.T) {
 			dynamicMap := map[string]*tls.Certificate{}
 
 			if test.staticCert != "" {
-				staticCert, _ := tls.LoadX509KeyPair(
-					fmt.Sprintf("../integration/fixtures/https/%s.cert", strings.Replace(test.staticCert, "*", "wildcard", -1)),
-					fmt.Sprintf("../integration/fixtures/https/%s.key", strings.Replace(test.staticCert, "*", "wildcard", -1)),
-				)
-				staticMap[test.staticCert] = &staticCert
+				staticCert, _ := loadTestCert(test.staticCert)
```
ldez:
```go
cert, err := loadTestCert(test.staticCert)
require.NoError(t, err)
```

tls/certificate_store_test.go
```diff
@@ -84,33 +84,36 @@ func TestGetBestCertificate(t *testing.T) {
...
+				staticMap[test.staticCert] = staticCert
 			}
 			if test.dynamicCert != "" {
-				dynamicCert, _ := tls.LoadX509KeyPair(
-					fmt.Sprintf("../integration/fixtures/https/%s.cert", strings.Replace(test.dynamicCert, "*", "wildcard", -1)),
-					fmt.Sprintf("../integration/fixtures/https/%s.key", strings.Replace(test.dynamicCert, "*", "wildcard", -1)),
-				)
-				dynamicMap[test.dynamicCert] = &dynamicCert
+				dynamicCert, _ := loadTestCert(test.dynamicCert)
```
ldez:
```go
cert, err := loadTestCert(test.dynamicCert)
require.NoError(t, err)
```

tls/certificate_store_test.go
```diff
@@ -84,33 +84,36 @@ func TestGetBestCertificate(t *testing.T) {
...
 				StaticCerts:  safe.New(staticMap),
 			}
 			var expected *tls.Certificate
 			if test.expectedCert != "" {
-				cert, _ := tls.LoadX509KeyPair(
-					fmt.Sprintf("../integration/fixtures/https/%s.cert", strings.Replace(test.expectedCert, "*", "wildcard", -1)),
-					fmt.Sprintf("../integration/fixtures/https/%s.key", strings.Replace(test.expectedCert, "*", "wildcard", -1)),
-				)
-				expected = &cert
+				cert, _ := loadTestCert(test.expectedCert)
```
ldez:
```go
cert, err := loadTestCert(test.expectedCert)
require.NoError(t, err)
```

tls/certificate_store.go
dtomcej: Because if we are inspecting the certs, we probably want to do a bit more checks such as the RSA/EC key checks etc...

nmengin: It is OK is we do that in a next PR :+1:

tls/certificate_store.go
```diff
@@ -2,14 +2,19 @@ package tls
...
 )
 
 // CertificateStore store for dynamic and static certificates
 type CertificateStore struct {
-	DynamicCerts *safe.Safe
-	StaticCerts  *safe.Safe
+	DynamicCerts       *safe.Safe
+	StaticCerts        *safe.Safe
+	DefaultCertificate *tls.Certificate
+	CertCache          map[string]*tls.Certificate
```
ldez: use a `map` as don't seem to be a good idea:
- what's happen when you update a dynamic cert?
- what's happen when an ACME cert is renew?

PR #3516, opened 2018-06-19
Allow any kubernetes ingressClass value

rtreffer:
Migrating from one ingress controller to another can require an intentional collision of ingressClass values.

### What does this PR do?

The ingressClass in kubernetes can be used to run multiple ingress controllers or to otherwise split the defined ingresses into multiple sets.
The values that Traefik would allow for an ingress class must start with `traefik`.

This patch removes that limitation as
1. The name is user defined anyway
2. It makes migrations from other ingress controllers harder

### Motivation

I am trying to move a larger cluster over to traefik and our ingress classes were split into internal and external. Traefik can't handle these names which makes a migration harder than necessary.

### More

- [x] Added/updated tests

### Additional Notes

The key change is the removal of a few lines in `provider/kubernetes/kubernetes.go`

### DIFF
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -54,8 +54,6 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # If the parameter is non-empty, only Ingresses containing an annotation with the same value are processed.
 # Otherwise, Ingresses missing the annotation, having an empty value, or the value `traefik` are processed.
 #
-# Note : `ingressClass` option must begin with the "traefik" prefix.
-#
 # Optional
 # Default: empty
 #
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -864,8 +864,8 @@ It is also possible to set the `ingressClass` option in Træfik to a particular 
 For instance, setting the option to `traefik-internal` causes Træfik to process Ingress objects with the same `kubernetes.io/ingress.class` annotation value, ignoring all other objects (including those with a `traefik` value, empty value, and missing annotation).
 
 !!! note
-    Letting multiple ingress controllers handle the same ingress objects can lead to uninteded behavior.
-    It is recommended to prefix all ingressClass values with traefik to avoid unintended collisions with other ingress implementations.
+    Letting multiple ingress controllers handle the same ingress objects can lead to unintended behavior.
+    It is recommended to prefix all ingressClass values with `traefik` to avoid unintended collisions with other ingress implementations.
 
 ### Between multiple Træfik Deployments
 
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -859,10 +859,8 @@ Sometimes Træfik runs along other Ingress controller implementations. One such 
 The `kubernetes.io/ingress.class` annotation can be attached to any Ingress object in order to control whether Træfik should handle it.
 
 If the annotation is missing, contains an empty value, or the value `traefik`, then the Træfik controller will take responsibility and process the associated Ingress object.
-If the annotation contains any other value (usually the name of a different Ingress controller), Træfik will ignore the object.
 
-It is also possible to set the `ingressClass` option in Træfik to a particular value.
-If that's the case and the value contains a `traefik` prefix, then only those Ingress objects matching the same value will be processed.
+It is also possible to set the `ingressClass` option in Træfik to a particular value. Træfik will only process matching Ingress objects.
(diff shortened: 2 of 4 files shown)


dtomcej:
In 1.7, we introduce dynamic namespacing, and ingress status updates.

Being that we are now modifying the ingress objects, allowing users to set "nginx" as the ingress class _could_ lead to problems, if there are multiple controllers trying to satisfy the ingress...however, the user is ultimately the cause.

I guess the other issue was at the time, we didn't support many of the annotations, and we wanted to ensure that users knew that they would have to modify ingresses that were written for other controllers. Now we support a _much_ larger annotation set, so 🤷‍♂️ 

I'm willing to allow it to be more flexible, and let users "figure it out"...

docs/user-guide/kubernetes.md
```diff
@@ -859,10 +859,8 @@ Sometimes Træfik runs along other Ingress controller implementations. One such
 The `kubernetes.io/ingress.class` annotation can be attached to any Ingress object in order to control whether Træfik should handle it.
 
 If the annotation is missing, contains an empty value, or the value `traefik`, then the Træfik controller will take responsibility and process the associated Ingress object.
-If the annotation contains any other value (usually the name of a different Ingress controller), Træfik will ignore the object.
 
-It is also possible to set the `ingressClass` option in Træfik to a particular value.
-If that's the case and the value contains a `traefik` prefix, then only those Ingress objects matching the same value will be processed.
+It is also possible to set the `ingressClass` option in Træfik to a particular value. Træfik will only process matching Ingress objects.
```
dtomcej: Should add a note that having multiple controllers satisfying an ingress can lead to unintended behavior

provider/kubernetes/kubernetes_test.go
```diff
@@ -1109,6 +1109,15 @@ func TestIngressClassAnnotation(t *testing.T) {
 					iPaths(onePath(iPath("/derp"), iBackend("service1", intstr.FromInt(80))))),
 			),
 		),
+		buildIngress(
+			iNamespace("testing"),
+			iAnnotation(annotationKubernetesIngressClass, "custom"),
+			iRules(
+				iRule(
+					iHost("herp"),
```
dtomcej: Could you replace the `herp/derp` with `foo/bar` or `foo/path` to match the format of other tests?

provider/kubernetes/kubernetes_test.go
```diff
@@ -1204,6 +1213,28 @@ func TestIngressClassAnnotation(t *testing.T) {
...
 			),
 		},
+		{
+			desc:     "Provided IngressClass annotation",
+			provider: Provider{IngressClass: "custom"},
+			expected: buildConfiguration(
+				backends(
+					backend("herp/derp",
+						servers(
+							server("http://example.com", weight(1)),
```
dtomcej: could you update this to be an IP/Port combo instead of a URL?
Currently that is only implemented as a URL when its an externalName service IIRC

docs/configuration/backends/kubernetes.md
```diff
@@ -54,8 +54,6 @@ See also [Kubernetes user guide](/user-guide/kubernetes).
 # If the parameter is non-empty, only Ingresses containing an annotation with the same value are processed.
 # Otherwise, Ingresses missing the annotation, having an empty value, or the value `traefik` are processed.
 #
-# Note : `ingressClass` option must begin with the "traefik" prefix.
-#
```
rtreffer: @timoreimann dropped.

docs/user-guide/kubernetes.md
rtreffer: Added a `!!! note` underneath with a _recommendation_ to prefix ingressClasses.

docs/user-guide/kubernetes.md
```diff
@@ -863,6 +863,10 @@ If the annotation is missing, contains an empty value, or the value `traefik`, t
 It is also possible to set the `ingressClass` option in Træfik to a particular value. Træfik will only process matching Ingress objects.
 For instance, setting the option to `traefik-internal` causes Træfik to process Ingress objects with the same `kubernetes.io/ingress.class` annotation value, ignoring all other objects (including those with a `traefik` value, empty value, and missing annotation).
 
+!!! note
+    Letting multiple ingress controllers handle the same ingress objects can lead to uninteded behavior.
```
timoreimann: Typo: _ uninteded_ -> _ unintended_

docs/user-guide/kubernetes.md
```diff
@@ -863,6 +863,10 @@ If the annotation is missing, contains an empty value, or the value `traefik`, t
...
+    It is recommended to prefix all ingressClass values with traefik to avoid unintended collisions with other ingress implementations.
```
timoreimann: Could you backtick _traefik_ so that it renders as `traefik`?

PR #3523, opened 2018-06-21
Add support for stale reads from Consul catalog

### LINKED ISSUE #2332: [consul] allow using "stale" data
The consul Go SDK allow using `stale` data - [meaning it can talk to any consul client to get information from the catalog instead of the elected leader](https://github.com/hashicorp/consul/blob/master/api/api.go#L72)

Due to how raft work, stale data usually mean 50ms delay in updates, and everything otherwise works like normal (blocked queries etc)

This setting is crucial in scaling out usage of consul, and would be a nice improvement to the consul catalog backend, and should probably always be default on

I can submit a PR with this change if needed

marenzo:
### What does this PR do?

Allow to read stale data from Consul, thus spreading the read load across Consul agents rather than fetching from leader.

https://www.consul.io/api/index.html#consistency-modes

For most of the use-cases, stale read is preferred, when you're willing to take the consistency trade-off and enjoy faster reads and distributed load. 

Fixes #2332 

### Motivation

Started using Traefik, and Consul catalog is the main provider.
As we're having multiple deployments of Traefik, and another systems that are actively fetching catalog
from the leaders, we see slowness of updates after some periods of time, especially when there's noise in Consul. Allowing to read stale data will reduce the load on the leaders, and serve better Traefik users.

### More

### Additional Notes

Currently, I set the default as false, only for backward compatibility purposes.
Looking forward, this option should be on by default, as vast majority of the cases it's the right way to work against Consul.

### DIFF
--- a/cmd/configuration.go
+++ b/cmd/configuration.go
@@ -108,6 +108,7 @@ func NewTraefikDefaultPointersConfiguration() *TraefikConfiguration {
 	defaultConsulCatalog.Constraints = types.Constraints{}
 	defaultConsulCatalog.Prefix = "traefik"
 	defaultConsulCatalog.FrontEndRule = "Host:{{.ServiceName}}.{{.Domain}}"
+	defaultConsulCatalog.Stale = false
 
 	// default Etcd
 	var defaultEtcd etcd.Provider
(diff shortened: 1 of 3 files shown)


marenzo:
hey guys, any chance you checked the PR?

Issue #2854, opened 2018-02-13
Need a way to refer to container port instead of host port


OlivierCazade:
Hi,

### Do you want to request a *feature* or report a *bug*?

Request a feature

### What did you do?

We are using traefik with ecs backend.

Our services are using dynamic host port (host port is dynamically set by ecs). As long as services are only exporting one port, Traefik is doing great.

When using multiple ports, we can not tell traefik which port to use since using traefik.port label is about host port.

### What did you expect to see?

A way to refer to internal port.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
$docker run traefik:v1.5 version
Version:      v1.5.1
Codename:     cancoillotte
Go version:   go1.9.3
Built:        2018-01-29_02:14:02PM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
[entryPoints]
    [entryPoints.http]
    address = ":80"

[ecs]
clusters = ["default"]
watch = true
domain = "docker.test"
autoDiscoverClusters = false
refreshSeconds = 15
exposedByDefault = false
```

Here is how we set container ports in Task definition files
```json
"portMappings": [
  {
    "containerPort": 80
  },
  {
    "containerPort": 8080
  }
]
```

mmatur:
Fixed by #3533

Issue #2896, opened 2018-02-20
ECS backend task definition examples


eduherraiz:
### Do you want to request a *feature* or report a *bug*?

feature

### What did you do?

I'm trying to configure the ECS backend with the last version of traefik, my personal decision is configure the traefik service inside the same ECS cluster where the applications are. 
I'm in trouble to get a working definition task for traefik in ECS and it is not well documented.
The ALB balancer can't healthcheck traefik's enpoint and I can't see the dashboard.

### What did you expect to see?

I expect to see a definition_task example for ECS in:
https://github.com/containous/traefik/tree/master/examples
Or see a way to use in docs.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.5.1
Codename:     cancoillotte
Go version:   go1.9.3
Built:        2018-01-29_02:14:02PM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

My task and not working definition now lookslike:

```json
[
    {
        "name": "proxy",
        "image": "traefik:${traefik_version}",
        "command": [
          "--api",
          "--ecs.clusters=${environment}",
          "--ecs.exposedbydefault=false",
          "--loglevel=DEBUG"
        ],
        "essential": true,
        "environment": [
            {
              "name": "AWS_ACCESS_KEY_ID",
              "value": "${aws_access_key_id}"
            },
            {
              "name": "AWS_REGION",
              "value": "${aws_region}"
            },
            {
              "name": "AWS_SECRET_ACCESS_KEY",
              "value": "${aws_secret_access_key}"
            }
        ],
        "dockerLabels": {
          "traefik.frontend.rule": "Host:traefik.${environment}.${domain}",
          "traefik.enabled": "true",
          "traefik.port": "8080"
        },
        "portMappings": [
            {
                "protocol": "tcp",
                "containerPort": 80,
                "hostPort": 0
            },
            {
                "protocol": "tcp",
                "containerPort": 8080,
                "hostPort": 0
            }
        ],
        "cpu": 20,
        "memoryReservation": 90
    }
]
```

eduherraiz:
Working on a new task definition I get a task_definition where the healthcheck and default http entrypoint is working, but I can't get work the dashboard:

```
[
    {
        "name": "proxy",
        "image": "traefik:${traefik_version}",
        "command": [
          "--api",
          "--ping",
          "--ping.entrypoint=http",
[...]

PR #3533, opened 2018-06-26
Allow binding ECS container port

andrewstucki:
### What does this PR do?

Implements https://github.com/containous/traefik/issues/2854. Makes Traefik look for an ECS label called `containerPort` to allow for dynamic mapping.

### Motivation

See https://github.com/containous/traefik/issues/2854 for a longer description, but basically there's no current way to specify routing behavior for ECS containers with more than one dynamically exposed port. This means without this feature:

1. In order to make Traefik route correctly you must either:
a. expose only a single port
b. statically bind the container port to a well-known host port
2. 1b. breaks the ability to schedule multiple containers of the same type on a single EC2 host.

### More

- [x] Added/updated tests

### Additional Notes

In the final solution, we use only the `traefik.port` label

| label                  | Docker Host      | Container | Value used       |
|------------------------|------------------|-----------|------------------|
| `traefik.port` not set | 8080             | 8080      | 8080             |
| `traefik.port` not set | DYNAMIC_PORT | 8080      | DYNAMIC_PORT |
| `traefik.port=8080`    | 8080             | 8080      | 8080             |
| `traefik.port=8080`    | DYNAMIC_PORT | 8080      | DYNAMIC_PORT |
| `traefik.port=80`      | DYNAMIC_PORT | 8080      | 80               |


andrewstucki:
Also, for clarity, as per #2896 the only way to do mapping prior to this is something along these lines in a task definition:

```
	"portMappings": [
            {
		"containerPort": 80
            },
            {
		"hostPort": 8080,
		"protocol": "tcp",
		"containerPort": 8080
            }
	],
	"dockerLabels" :           {
            "traefik.enable" : "true",
            "traefik.frontend.rule" : "Host:XXXX",
            "traefik.port" : "8080"
        }
```

The static port binding clearly means you can only have one of these tasks running on a given host. The PR allows for something like:

```
	"portMappings": [
            {
		"containerPort": 80
            },
            {
		"containerPort": 8080
            }
	],
	"dockerLabels" :           {
            "traefik.enable" : "true",
            "traefik.frontend.rule" : "Host:XXXX",
            "traefik.containerPort" : "8080"
        }
```

Which allows you to do bin-packing and removes the necessity for static port binding.

mmatur:
Hi @andrewstucki,

Thanks for your interest in Træfik.

Why do you need a new label `containerPort`?
I think that you can reuse `traefik.port` label and adapt `getPort` function

WDYT ?

andrewstucki:
@mmatur well, the current `traefik.port` label (which the ECS provider supports) just sets up the routes so that Traefik assumes something is listening on that port on the host (it assumes you know what port the container is going to bind to, which in the case of dynamic port binding, you don't).

I wouldn't mind changing the functionality of the label so that it resolves dynamic port binding instead and treats the port specified as the container's port, but figured that making a breaking change to existing behavior was probably something we didn't want to do, hence the implementation of a new label. Off the top of my head I can't think of a way of implementing the new behavior while keeping the old behavior without a new label.

Let me know if you have any ideas.

andrewstucki:
So, hopefully to illustrate the point better, here's how things currently work.

Assuming static port binding a setup can look like this:

| label                                | Docker Host  | Container      |
|------------------------| ------------- | ------------- |
| `traefik.port`                 | **8081**       | 8080              |
| `traefik.containerPort` | 8081              | **8080**       |

Where `traefik.port` is a reference to the host-level port. The new label references the container-level port. With static binding either would work.

However, when you have dynamic binding it looks like more this:

| label                                | Docker Host  | Container      |
|------------------------| ------------- | ------------- |
| `traefik.port`                 | **?**              | 8080              |
| `traefik.containerPort` | ?                     | **8080**       |

In the dynamic binding scenario we have no way of using the behavior `traefik.port` implements because we don't know which host-level port will get bound to. The only way you can choose the right port is to reference the port on the container side of things, in other words, the new `traefik.containerPort` label.

So, in most scenarios this probably won't be a huge deal since many containers only expose a single port. As such, Traefik will just choose the first port that it sees and route to it. However, In the case where you have multiple dynamically bound ports, then this becomes an issue because we currently have no way of specifying which of the exposed ports is the one we care about.

My particular use case is co-locating a bunch of services that expose two ports dynamically, a port that's internally scraped by Prometheus, and a port that a web-server serving up and API listens on. Since I want to co-locate all of the services, I can't dynamically bind either the Prometheus or the web-server ports, but need a way to tell Traefik to only route to the port that the web-server is bound on, meaning I need to tell Traefik to only route to the port exposed as port 8080 in my container and to ignore whatever translates to port 9000. What that maps to at the host level, I have no way of knowing due to the dynamic nature of the port binding.

mmatur:
@andrewstucki Thanks for you feedback.

I was thinking about a solution like that 

```go
func getPort(i ecsInstance) string {
	if value := label.GetStringValue(i.TraefikLabels, label.TraefikPort, ""); len(value) > 0 {
		port, err := strconv.ParseInt(value, 10, 64)
		for _, mapping := range i.machine.ports {
			if err == nil {
				if port == mapping.hostPort || port == mapping.containerPort {
					return strconv.FormatInt(mapping.hostPort, 10)
				}
			}
		}
		return value
	}
	return strconv.FormatInt(i.machine.ports[0].hostPort, 10)
}
```

If `traefik.port` match the `hostPort` we return the `hostPort` value
If `traefik.port` match the `containerPort` we return the `hostPort` value
If `traefik.port` does not match the `hostPort` or the containerPort we return the `traefik.port` value
if `traefik.port` match `hostPort` and `containerPort` we return `hostPort` value

With an example
| label                         | Docker Host         | Container | Value used             |
|----------------------------|-------------------------|---------------|--------------------------|
| `traefik.port` not set | 8080                      | 8080        | 8080                       |
| `traefik.port` not set | DYNAMIC_PORT | 8080         | DYNAMIC_PORT  |
| `traefik.port=8080`   | 8080                     | 8080         | 8080                       |
| `traefik.port=8080`   | DYNAMIC_PORT | 8080         | DYNAMIC_PORT  |
| `traefik.port=80`       | DYNAMIC_PORT | 8080         | 80                           |

WDYT ?

mmatur:
Hi @andrewstucki,

The code freeze for the next 1.7 release will happen at the end of this week, on Friday 6th July 2018.
So I will do the change on your PR because we want to be able to merge your PR into the next release

shanthakumarf22:
is this documented ? does `traefik.containerPort` exist in `master`?

PR #3542, opened 2018-06-28
Change code block of traefik-web-ui to match file

drewgwallace:
@dtomcej Docs still show corrections necessary against Master
https://github.com/containous/traefik/blob/master/examples/k8s/ui.yaml

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -323,6 +323,7 @@ For more information, check out [the documentation](https://github.com/kubernete
 Lets start by creating a Service and an Ingress that will expose the [Træfik Web UI](https://github.com/containous/traefik#web-ui).
 
 ```yaml
+---
 apiVersion: v1
 kind: Service
 metadata:
@@ -332,7 +333,8 @@ spec:
   selector:
     k8s-app: traefik-ingress-lb
   ports:
-  - port: 80
+  - name: web
+    port: 80
     targetPort: 8080
 ---
 apiVersion: extensions/v1beta1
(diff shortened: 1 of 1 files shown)


docs/user-guide/kubernetes.md
```diff
@@ -323,6 +323,7 @@ For more information, check out [the documentation](https://github.com/kubernete
 Lets start by creating a Service and an Ingress that will expose the [Træfik Web UI](https://github.com/containous/traefik#web-ui).
 
 ```yaml
+---
```
dtomcej: This line is not needed.

docs/user-guide/kubernetes.md
```diff
@@ -332,24 +333,24 @@ spec:
...
   rules:
   - host: traefik-ui.minikube
     http:
       paths:
-      - backend:
+      - path: /
+        backend:
           serviceName: traefik-web-ui
-          servicePort: 80
+servicePort: web
```
dtomcej: Indenting

docs/user-guide/kubernetes.md
dtomcej: Can probably remove that line from the actual yaml too

drewgwallace:
I updated with those changes. It appears this document has the leading '---' on any code block with multiple components.

docs/user-guide/kubernetes.md
```diff
@@ -332,24 +332,24 @@ spec:
...
     targetPort: 8080
 ---
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
   name: traefik-web-ui
   namespace: kube-system
-  annotations:
-    kubernetes.io/ingress.class: traefik
 spec:
```
timoreimann: The annotation makes sense to have in general. I'm thinking that we should rather add it to the YAML file than remove it from the inline spec.

@dtomcej WDYT?

dtomcej: The argument is that we don't specify an ingress class to watch on any of our examples. If we keep it, we _should_ be adding the class to our deployment/daemonset...

timoreimann: Gotcha. Even if we decided to do that, it should be done by a separate PR. No reason to block this one.

PR #3556, opened 2018-07-02
Add a k8s guide section on traffic splitting via service weights.

timoreimann:
### What does this PR do?

Explain service weights in the Kubernetes guide.

### Motivation

To ease adoption of the feature for users.

### Additional Notes

Complements #3112.

@dtomcej PTAL

/cc @yue9944882

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -876,6 +876,89 @@ For such cases, it is advisable to classify Ingress objects through a label and 
 To stick with the internal/external example above, all Ingress objects meant for internal traffic could receive a `traffic-type: internal` label while objects designated for external traffic receive a `traffic-type: external` label.
 The label selectors on the Træfik Deployments would then be `traffic-type=internal` and `traffic-type=external`, respectively.
 
+## Traffic Splitting
+
+It is possible to split Ingress traffic in a fine-grained manner between multiple deployments using _service weights_.
+
+One canonical use case is canary releases where a deployment representing a newer release is to receive an initially small but ever-increasing fraction of the requests over time.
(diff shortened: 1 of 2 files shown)


docs/user-guide/kubernetes.md
```diff
@@ -876,6 +876,89 @@ For such cases, it is advisable to classify Ingress objects through a label and
...
+
+Take note of the `traefik.ingress.kubernetes.io/service-weights` annotation: It specifies the distribution of requests among the referenced backend services, `my-app` and `my-app-canary`.
+With this definition, Træfik will route 99% of the requests to the pods backed by the `my-app` deployment, and 1% to those backed by `my-app-canary`.
+Over time, the ratio may slowly shift towards the canary deployment until it is deemed to replace the previous main application, in steps such as 5%/95%, 10%/90%, 50%/50%, and finally 100%/0%.
+
+A few conditions must hold for service weights to be applied correctly:
+
+- The associated service backends must share the same path and host.
+- The total percentage shared across all service backends must yield 100% (see the section on [omitting the final service](#omitting-the-final-service), however).
+- The percentage values are interpreted as floating point numbers to a supported precision as defined in the [configuration details](/configuration/backends/kubernetes).
```
yue9944882: Could we add an anchor of this document? It's a long document and  I don't think users might find the detail of this annotation easily.

> /configuration/backends/kubernetes#specify-weight

timoreimann: A deep link seems tricky to do unless we resort to using custom HTML in markdown. For now, I decided to point to the annotations section. Let me know what you think.

timoreimann:
Pushed another commit that removes the canary use case example from the backend docs as we now have that in the user guide. I replaced it with a link.

PR #3560, opened 2018-07-04
Split the error handling from Consul Catalog (deadlock)

ortz:
### The issue
There's a deadlock in Consul Catalog's watch method: the notifyError function writes to the same channel where the select loop is trying to read from.

### Fix
Separating the watch channel from the stop/kill channels and implement it using a different goroutine 
so that read & write from the error channel won't block each other.

### Steps to reproduce
1. Launch traefik using a Consul Catalog against a Consul agent.
2. Restart Consul agent (causes Consul to return an error).
3. Change/deploy a new service to Consul and see that the backends/frontends don't get updated.

### DIFF
--- a/provider/consulcatalog/consul_catalog.go
+++ b/provider/consulcatalog/consul_catalog.go
@@ -1,7 +1,6 @@
 package consulcatalog
 
 import (
-	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -155,14 +154,8 @@ func (p *Provider) watch(configurationChan chan<- types.ConfigMessage, stop chan
 	defer close(stopCh)
 	defer close(watchCh)
 
-	for {
-		select {
-		case <-stop:
-			return nil
-		case index, ok := <-watchCh:
-			if !ok {
-				return errors.New("consul service list nil")
-			}
+	go func() {
+		for index := range watchCh {
 			log.Debug("List of services changed")
 			nodes, err := p.getNodes(index)
 			if err != nil {
@@ -173,6 +166,13 @@ func (p *Provider) watch(configurationChan chan<- types.ConfigMessage, stop chan
 				ProviderName:  "consul_catalog",
 				Configuration: configuration,
 			}
+		}
+	}()
+
+	for {
+		select {
+		case <-stop:
+			return nil
 		case err := <-errorCh:
 			return err
 		}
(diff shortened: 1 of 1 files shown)


provider/consulcatalog/consul_catalog.go
```diff
@@ -156,14 +155,8 @@ func (p *Provider) watch(configurationChan chan<- types.ConfigMessage, stop chan
 	defer close(stopCh)
 	defer close(watchCh)
 
-	for {
-		select {
-		case <-stop:
-			return nil
-		case index, ok := <-watchCh:
-			if !ok {
-				return errors.New("consul service list nil")
```
Juliens: In fact, as we return an error here, the caller (https://github.com/ortz/traefik/blob/257d278191f547db00c5cd1fccd008324e5b8a60/provider/consulcatalog/consul_catalog.go#L109) will automatically retry because we use backoff.RetryNotify (https://github.com/ortz/traefik/blob/257d278191f547db00c5cd1fccd008324e5b8a60/provider/consulcatalog/consul_catalog.go#L111)

provider/consulcatalog/consul_catalog.go
```diff
@@ -156,14 +155,8 @@ func (p *Provider) watch(configurationChan chan<- types.ConfigMessage, stop chan
...
-			}
+	go func() {
```
Juliens: You can't just launch new goroutine without handling the stop. Because we need to be able to stop all providers goroutines when we want. This is the purpose of the `safe.Pool` we receive in the `Provide` method, and this is the purpose of the stop chan here.

ozkatz: Not really familiar with Traefik's design, but the stop channel will cause the watch method to return, causing `defer close(watchCh)` [here](https://github.com/containous/traefik/pull/3560/files/257d278191f547db00c5cd1fccd008324e5b8a60#diff-cd8329c0a91a00757f75bd879fa2cf7cL157)  to execute, which will in turn end the iteration in the goroutine - so stop is handled implicitly. Or am I misunderstanding?

Juliens: You're right, it will be closed, but I think we should use a `safe.Go` (to recover eventual panic)

ldez:
@ortz could you rebase you pull request on the branch `v1.6`. Thanks.

ortz:
@ldez done, I've also changed the base branch to v1.6

provider/consulcatalog/consul_catalog.go
```diff
@@ -155,14 +154,8 @@ func (p *Provider) watch(configurationChan chan<- types.ConfigMessage, stop chan
...
```
ldez: @ortz could you use a `safe.Go`

ortz: @ldez done

PR #3563, opened 2018-07-05
Performance enhancements for the rules matchers.

ShaneSaww:
This commit works towards calling `types.CannonicalDomain` less and reducing the amount of calls that are in the hotpath for a request.

The big change here is moving the cannon'd request Host header into a middleware so it only has be cannonized once per request.

### What does this PR do?

This PR creates a middleware that gets hit early on in the request lifecycle and puts then `CanonicalDomain` request host into a context to be read during the route Matcher funcs. 

Along with that this re-works the matcherFuncs to call `CanonicalDomain` less by using the context, and calling `CanonicalDomain` for the hosts at the initialization of the matcherFunc not inside the matcher func where it's expensive.  

### Motivation

Performance tuning in our production environment showed that `CanonicalDomain` was using a good chunk of CPU time. 

See the include flameGraph photos of the `matcherFunc` section. 

Before:
matcherfuncbefore

After:
matcherfuncafter

### More

- [x] Added/updated tests


middlewares/reqHost.go
```diff
@@ -0,0 +1,41 @@
...
+		}
+	}
+	reqHost = types.CanonicalDomain(reqHost)
+	if next != nil {
+		next.ServeHTTP(rw, r.WithContext(context.WithValue(r.Context(), reqHostKey, reqHost)))
+	}
+}
+
+// GetCannonHost plucks the cannonized host key from the requst of a context that was put through the middleware
+func GetCannonHost(ctx context.Context) (string, bool) {
```
mmatur: Should maybe rename into `GetCanonHost`

middlewares/reqHost.go
```diff
@@ -0,0 +1,41 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/containous/traefik/types"
+)
+
+var reqHostKey struct{}
+
+// ReqHostMiddleware is the struct for the middleware that adds the CanonicalDomain of the requst Host into a context for later use.
+type ReqHostMiddleware struct{}
```
mmatur: Could you rename `ReqHostMiddleware` by `ReqHost` please

middlewares/reqHost.go
```diff
@@ -0,0 +1,41 @@
...
+
+// NewReqHostMiddleware is a middleware that adds the CanonicalDomain of the requst Host into a context for later use.
+func NewReqHostMiddleware() *ReqHostMiddleware {
```
mmatur: This function is not necessary, you could directly use `&ReqHostMiddleware{}` instead of creating a function for that. Like it is done for `Compress` or `AddPrefix` middlewares

middlewares/reqHost.go
```diff
@@ -0,0 +1,41 @@
...
+	return &ReqHostMiddleware{}
+}
+
+func (rhm *ReqHostMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
```
mmatur: You could maybe rename `rhm` with `rh`

middlewares/reqHost.go
```diff
@@ -0,0 +1,41 @@
...
+	reqHost := r.Host
```
mmatur: You could maybe simplify like that 
```go
func (rh *ReqHostMiddleware) ServeHTTP(rw http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
	reqHost := types.CanonicalDomain(parseHost(r.Host))
	if next != nil {
		next.ServeHTTP(rw, r.WithContext(context.WithValue(r.Context(), reqHostKey, reqHost)))
	}
}

func parseHost(addr string) string {
	if !strings.Contains(addr, ":") {
		return addr
	}

	host, _, err := net.SplitHostPort(addr)
	if err != nil {
		return addr
	}
	return host
}
```

WDYT?

And you should be able to test `parseHost` function

mmatur:
The code freeze happen today, so I will do the change to be able to merge your PR today before the code freeze

PR #3564, opened 2018-07-05
Call functions to enable block/mutex pprof profiles.

timoreimann:
### What does this PR do?

Call `runtime.SetBlockProfileRate` and `runtime.SetMutexProfileFraction` when debug mode is enabled.

### Motivation

According to the [documentation](https://golang.org/pkg/net/http/pprof/), these functions must be called for block and mutex pprof profiles to be generated actually.


timoreimann:
Kudos to @marco-jantke for clarifying the need.

Issue #3597, opened 2018-07-11
Support multiple frontends per service for consul catalog provider


hsmade:
### Do you want to request a *feature* or report a *bug*?
Feature

### What did you expect to see?

In some cases, a single frontend per backend isn't enough. For example when having the need for multiple routes that can't be combined into a single rule. In the File and the Consul provider, we can then use multiple frontends to solve that, but the consul catalog provider does not support this.

hsmade:
Implemented by #3796

PR #3598, opened 2018-07-11
Wf multiple frontends consulcatalog

hsmade:
### What does this PR do?

This adds support for multiple frontends in the consul catalog provider.

### Motivation

To implement this feature request:
https://github.com/containous/traefik/issues/3597

### Additional Notes

I'm wondering if it's preferred to drop the single frontend support (which will make it possible to clean up the code, and remove the 'hacks' for backwards compatibility), and maybe make this a V3?


ldez:
I see several design issues:
- the signature of the method `getFrontendRule` has been changed: it's breaking
- the duplication of the code in the template and the construction logic placed in the template goes against the work that has been done on [the homogenization of templates](https://github.com/containous/traefik/issues/2785).
- using the syntax used by the segment labels, without having real support, does not suit me. (And I do not think it's a good idea to add support for segments)

hsmade:
After looking at the docker provider (I'm completely new to traefik and it's code), I start to see why this doesn't really fit what is there already. I will refactor my PR to work somewhat the same way (have a `Frontends` var in `templateObjects` instead of trying to put that logic into the template) and thus also get rid of the extra funcs I added to the template.

ldez:
Could you fix:
```

 M autogen/gentemplates/gen.go
```

hsmade:
@ldez Done

hsmade:
This PR has been replaced by https://github.com/containous/traefik/pull/3781 which implements this in a much simpler way and closer to the original source

PR #3604, opened 2018-07-11
Update Namecheap status

stoinov:
### What does this PR do?

I updated the status of Namecheap for the DNS challenge 

### Motivation

I tested and was able to use the DNS challenge for a wildcard certificates with Namecheap

### More


dtomcej:
Changing to 1.7 base

PR #3648, opened 2018-07-19
Add possibility to set a protocol

### LINKED ISSUE #3641: Backend protocol selection for Kubernetes provider

Feature

With 1.7.0 we have the new h2c backend protocol. To support h2c backends with the Kubernetes provider, we need some way to specify which backend protocol to use in the Kubernetes ingress configuration. Perhaps a new Kubernetes annotation can be added to allow the backend protocol to be explicitly selected?

SantoDE:
### What does this PR do?

Adds posibility to set a custom protocol for an ingress with the K8s provider

### Motivation

Fixes #3641

### More

- [X] Added/updated tests

### Additional Notes

I'm official dumb and screwed the other pr :)

### DIFF
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -159,7 +159,8 @@ The following general annotations are applicable on the Ingress object:
 | `traefik.ingress.kubernetes.io/whitelist-source-range: "1.2.3.0/24, fe80::/16"` | A comma-separated list of IP ranges permitted for access (6).                                                     |
 | `ingress.kubernetes.io/whitelist-x-forwarded-for: "true"`                       | Use `X-Forwarded-For` header as valid source of IP for the white list.                                            |
 | `traefik.ingress.kubernetes.io/app-root: "/index.html"`                         | Redirects all requests for `/` to the defined path. (4)                                                           |
-| `traefik.ingress.kubernetes.io/service-weights: <YML>`                          | Set ingress backend weights specified as percentage or decimal numbers in YAML. (5)                               |
+| `traefik.ingress.kubernetes.io/service-weights: <YML>`                          | Set ingress backend weights specified as percentage or decimal numbers in YAML. (5)            
+| `ingress.kubernetes.io/protocol: <NAME>`                                        | Set the protocol Traefik will use to communicate with pods 
 
 
 <1> `traefik.ingress.kubernetes.io/error-pages` example:
@@ -256,7 +257,7 @@ The following annotations are applicable on the Service object associated with a
 | `traefik.ingress.kubernetes.io/load-balancer-method: drr`                | Override the default `wrr` load balancer algorithm.                                                                                                                                   |
 | `traefik.ingress.kubernetes.io/max-conn-amount: 10`                      | Set a maximum number of connections to the backend.<br>Must be used in conjunction with the below label to take effect.                                                               |
 | `traefik.ingress.kubernetes.io/max-conn-extractor-func: client.ip`       | Set the function to be used against the request to determine what to limit maximum connections to the backend by.<br>Must be used in conjunction with the above label to take effect. |
-| `traefik.ingress.kubernetes.io/session-cookie-name: <NAME>`              | Manually set the cookie name for sticky sessions.                                                                                                                                     |
+| `traefik.ingress.kubernetes.io/session-cookie-name: <NAME>`              | Manually set the cookie name for sticky sessions.                                                                                                                                     | 
 
 !!! note
     `traefik.ingress.kubernetes.io/` and `ingress.kubernetes.io/` are supported prefixes.
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -160,7 +160,7 @@ The following general annotations are applicable on the Ingress object:
 | `ingress.kubernetes.io/whitelist-x-forwarded-for: "true"`                       | Use `X-Forwarded-For` header as valid source of IP for the white list.                                            |
 | `traefik.ingress.kubernetes.io/app-root: "/index.html"`                         | Redirects all requests for `/` to the defined path. (4)                                                           |
 | `traefik.ingress.kubernetes.io/service-weights: <YML>`                          | Set ingress backend weights specified as percentage or decimal numbers in YAML. (5)            
-| `ingress.kubernetes.io/protocol: <NAME>`                                        | Set the protocol Traefik will use to communicate with pods 
+| `ingress.kubernetes.io/protocol: <NAME>`                                        | Set the protocol Traefik will use to communicate with pods.
 
 
 <1> `traefik.ingress.kubernetes.io/error-pages` example:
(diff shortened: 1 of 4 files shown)


docs/configuration/backends/kubernetes.md
```diff
@@ -159,7 +159,8 @@ The following general annotations are applicable on the Ingress object:
 | `traefik.ingress.kubernetes.io/whitelist-source-range: "1.2.3.0/24, fe80::/16"` | A comma-separated list of IP ranges permitted for access (6).                                                     |
 | `ingress.kubernetes.io/whitelist-x-forwarded-for: "true"`                       | Use `X-Forwarded-For` header as valid source of IP for the white list.                                            |
 | `traefik.ingress.kubernetes.io/app-root: "/index.html"`                         | Redirects all requests for `/` to the defined path. (4)                                                           |
-| `traefik.ingress.kubernetes.io/service-weights: <YML>`                          | Set ingress backend weights specified as percentage or decimal numbers in YAML. (5)                               |
+| `traefik.ingress.kubernetes.io/service-weights: <YML>`                          | Set ingress backend weights specified as percentage or decimal numbers in YAML. (5)            
+| `ingress.kubernetes.io/protocol: <NAME>`                                        | Set the protocol Traefik will use to communicate with pods
```
dtomcej: Needs a `.` after `pods`

docs/configuration/backends/kubernetes.md
```diff
@@ -256,7 +257,7 @@ The following annotations are applicable on the Service object associated with a
...
 | `traefik.ingress.kubernetes.io/max-conn-amount: 10`                      | Set a maximum number of connections to the backend.<br>Must be used in conjunction with the below label to take effect.                                                               |
 | `traefik.ingress.kubernetes.io/max-conn-extractor-func: client.ip`       | Set the function to be used against the request to determine what to limit maximum connections to the backend by.<br>Must be used in conjunction with the above label to take effect. |
-| `traefik.ingress.kubernetes.io/session-cookie-name: <NAME>`              | Manually set the cookie name for sticky sessions.                                                                                                                                     |
+| `traefik.ingress.kubernetes.io/session-cookie-name: <NAME>`              | Manually set the cookie name for sticky sessions.                                                                                                                                     |
```
dtomcej: Does this need to be here?

SantoDE:
Morning @dtomcej,

requested changes applied :)

docs/configuration/backends/kubernetes.md
```diff
@@ -159,7 +159,8 @@ The following general annotations are applicable on the Ingress object:
 | `traefik.ingress.kubernetes.io/whitelist-source-range: "1.2.3.0/24, fe80::/16"` | A comma-separated list of IP ranges permitted for access (6).                                                     |
 | `ingress.kubernetes.io/whitelist-x-forwarded-for: "true"`                       | Use `X-Forwarded-For` header as valid source of IP for the white list.                                            |
 | `traefik.ingress.kubernetes.io/app-root: "/index.html"`                         | Redirects all requests for `/` to the defined path. (4)                                                           |
-| `traefik.ingress.kubernetes.io/service-weights: <YML>`                          | Set ingress backend weights specified as percentage or decimal numbers in YAML. (5)                               |
+| `traefik.ingress.kubernetes.io/service-weights: <YML>`                          | Set ingress backend weights specified as percentage or decimal numbers in YAML. (5)            
+| `ingress.kubernetes.io/protocol: <NAME>`                                        | Set the protocol Traefik will use to communicate with pods.
```
yue9944882: i thought ingress model is only for http/https protocol. this annotation allows users to set their own arbitrary protocol and once someone abuse it might cause some potential SQL-injection-like danger?🧐

dtomcej: @yue9944882 The issue here is that h2c runs over http, but uses its own protocol. All of the other providers have the `traefik.protol` label, but the k8s provider does not. This allows users to specify h2c as a backend protocol so h2c will work as is does with other providers. Although your point about potential abuse is valid.

@SantoDE what do you think about instead of blindly assigning the value to `protocol`, running it through a select/switch with acceptable protocols? `http`/`https`/`h2c` so that someone can't add `http://myinjectionsite.com/endpoint.php http` as the protocol ;)

SantoDE:
Hey @yue9944882 , @dtomcej 

thanks for your concerns. I added a switch to only accept wanted protocols. :-)

provider/kubernetes/kubernetes.go
```diff
@@ -1029,3 +1040,8 @@ func getRateLimit(i *extensionsv1beta1.Ingress) *types.RateLimit {
 
 	return rateLimit
 }
+
+func getProtocol(i *extensionsv1beta1.Ingress) string {
```
timoreimann: This function seems unused. I suppose we can just remove it?

provider/kubernetes/kubernetes.go
```diff
@@ -312,6 +314,15 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 							protocol = "https"
 						}
 
+						switch getStringValue(i.Annotations, annotationKubernetesProtocol, protocol) {
+						case allowedProtocolHTTPS:
+							protocol = allowedProtocolHTTPS
+						case allowedProtocolH2C:
+							protocol = allowedProtocolH2C
+						default:
```
timoreimann: I'm not too much of a fan of falling back to some default value for the case where the user entered invalid data. How about indicating an error and skipping the offending Ingress if the given annotation value is unsupported?

That's also how we seem to handle things in most other cases of the Kubernetes provider.

provider/kubernetes/kubernetes.go
```diff
@@ -312,6 +314,15 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
...
+							protocol = label.DefaultProtocol
```
timoreimann: Assuming we hit line 314 and have protocol be `https`: Aren't we overwriting the value to `http` again if the annotation is missing or empty (as the default of `http` will be returned to `getStringValue`, thereby causing our _default_ switch case to apply)?

provider/kubernetes/kubernetes_test.go
```diff
@@ -1224,6 +1224,24 @@ rateset:
...
 		),
+		buildIngress(
+			iNamespace("testing"),
+			iAnnotation(annotationKubernetesProtocol, "h2c"),
+			iRules(
+				iRule(
+					iHost("protocol"),
+					iPaths(onePath(iPath("/valid"), iBackend("service1", intstr.FromInt(80))))),
+			),
+		),
```
timoreimann: How about adding the test case where we don't use port 443 (i.e., protocol is set to `http`) but explicitly specify the `https` protocol through the annotation?

SantoDE: I added some more tests:)

SantoDE:
Hey @dtomcej  and @timoreimann ,

I addressed your feedback :)

provider/kubernetes/kubernetes.go
```diff
@@ -312,6 +314,18 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 							protocol = "https"
 						}
 
+						switch getStringValue(i.Annotations, annotationKubernetesProtocol, protocol) {
+						case allowedProtocolHTTPS:
+							protocol = allowedProtocolHTTPS
+						case allowedProtocolH2C:
+							protocol = allowedProtocolH2C
+						case label.DefaultProtocol:
+							protocol = label.DefaultProtocol
```
timoreimann: The switch statement can be simplified to:

```go
protocol = getStringValue(i.Annotations, annotationKubernetesProtocol, protocol)
switch protocol {
case allowedProtocolHTTPS:
case allowedProtocolH2C:
case label.DefaultProtocol:
default:
    log.Errorf(...)
}

PR #3655, opened 2018-07-20
Added default configuration for DataDog APM Tracer

aantono:
Fixed DataDog APM initialization error due to missing default configs.

### DIFF
--- a/configuration/configuration.go
+++ b/configuration/configuration.go
@@ -11,6 +11,7 @@ import (
 	"github.com/containous/traefik/api"
 	"github.com/containous/traefik/log"
 	"github.com/containous/traefik/middlewares/tracing"
+	"github.com/containous/traefik/middlewares/tracing/datadog"
 	"github.com/containous/traefik/middlewares/tracing/jaeger"
 	"github.com/containous/traefik/middlewares/tracing/zipkin"
 	"github.com/containous/traefik/ping"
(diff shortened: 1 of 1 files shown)


configuration/configuration.go
```diff
@@ -345,6 +346,18 @@ func (gc *GlobalConfiguration) initTracing() {
...
 			}
+		case datadog.Name:
+			if gc.Tracing.DataDog == nil {
+				gc.Tracing.DataDog = &datadog.Config{
+					LocalAgentHostPort: "localhost:8126",
+					GlobalTag:          "",
+					Debug:              false,
+				}
+			}
+			if gc.Tracing.DataDog != nil {
```
mmatur: Should be 
```go
if gc.Tracing.Zipkin != nil {
	log.Warn("Zipkin configuration will be ignored")
	gc.Tracing.Zipkin = nil
}
if gc.Tracing.Jaeger != nil {
	log.Warn("Jaeger configuration will be ignored")
	gc.Tracing.Jaeger = nil
}
```

or 

```go
log.Warn("Zipkin & Jaeger configuration will be ignored")
gc.Tracing.Jaeger = nil
gc.Tracing.Zipkin = nil
```

According to the chosen solution, you also need to update the `Zipkin` and `Jaeger` case

mmatur:
I am not very a fan to have multiple conditional check for each tracer. I would propose different solution.

Solution 1
```go
case jaeger.Name:
	if gc.Tracing.Jaeger == nil {
		gc.Tracing.Jaeger = &jaeger.Config{
			SamplingServerURL:  "http://localhost:5778/sampling",
			SamplingType:       "const",
			SamplingParam:      1.0,
			LocalAgentHostPort: "127.0.0.1:6831",
		}
	}
	if gc.Tracing.Zipkin != nil || gc.Tracing.DataDog != nil {
		log.Warn("Zipkin & DataDog configuration will be ignored")
		gc.Tracing.Zipkin = nil
                gc.Tracing.DataDog = nil
	}
```

Solution 2
```go
	log.Info("Using Jaeger configuration other tracing configurations will be ignored")
	if gc.Tracing.Jaeger == nil {
		gc.Tracing.Jaeger = &jaeger.Config{
			SamplingServerURL:  "http://localhost:5778/sampling",
			SamplingType:       "const",
			SamplingParam:      1.0,
			LocalAgentHostPort: "127.0.0.1:6831",
		}
	}
	gc.Tracing.Zipkin = nil
	gc.Tracing.DataDog = nil
```

Personally I prefer the second solution.

WDYT?

Issue #3650, opened 2018-07-20
acme provider does not reuse registered account


zyclonite:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

* starting traefik
* creates acme user account and requests certificates
* configure a new cert or have one installed which expires soon (or expired)
* restart traefik
* creates acme account **again** and requests new certificates

### What did you expect to see?

reuse of the already registered account

### What did you see instead?

```
DEBU[2018-07-20T15:52:34+02:00] Building ACME client...                      
DEBU[2018-07-20T15:52:34+02:00] https://acme-staging-v02.api.letsencrypt.org/directory 
INFO[2018-07-20T15:52:34+02:00] Register...                                  
INFO[2018-07-20T15:52:34+02:00] legolog: [INFO] acme: Registering account for test@mydomain.com 
```

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      d8f69700e6821465d818b769816710aa5f20908c
Codename:     cheddar
Go version:   go1.10.3
Built:        2018-07-20_02:18:47PM
OS/Arch:      linux/amd64
```
and the same behavior in the latest released version **1.6.5**

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
[acme]
  email = "test@mydomain.com"
  storage = "/etc/traefik/acme.json"
  entryPoint = "https"
  acmeLogging = true
  onDemand = false
  caServer = "https://acme-staging-v02.api.letsencrypt.org/directory"
```

### Possible root cause

https://github.com/containous/traefik/blob/d8f69700e6821465d818b769816710aa5f20908c/provider/acme/provider.go#L133

here the CAServer is compared against the account uri, which will never match on prefix...
```
https://acme-staging-v02.api.letsencrypt.org/directory != https://acme-staging-v02.api.letsencrypt.org/acme/acct/1234567
```
maybe parsing the URI and matching the hostname would be fine

dduportal:
Hi @zyclonite !

Thanks for your interest in this project.

I was not able to reproduce your issue with the informations you've put here: The toml extract is not complete. I infered some missing attributes, but was not able to reproduce your scenario (Creating a certificate, it create my account, I revoke the cert, restart traefik: the account is not re-created).

I'm going to ask for help about the potential code issue you are mentioning (ping @mmatur : do you know this part?), but it looks like there is something else. Could you add more reproduction information please?

[...]

PR #3664, opened 2018-07-24
Fix acme account deletion without provider change

zyclonite:
### What does this PR do?

prevent traefik from deleting the acme.account without provider change
fixes issue #3650

### DIFF
--- a/acme/acme.go
+++ b/acme/acme.go
@@ -232,13 +232,13 @@ func (a *ACME) leadershipListener(elected bool) error {
 	return nil
 }
 
-func isAccountMatchingCaServer(accountUri string, serverUri string) bool {
-	aru, err := url.Parse(accountUri)
+func isAccountMatchingCaServer(accountURI string, serverURI string) bool {
+	aru, err := url.Parse(accountURI)
 	if err != nil {
 		log.Infof("Unable to parse account.Registration URL : %v", err)
 		return false
 	}
-	cau, err := url.Parse(serverUri)
+	cau, err := url.Parse(serverURI)
 	if err != nil {
 		log.Infof("Unable to parse CAServer URL : %v", err)
 		return false
--- a/acme/acme.go
+++ b/acme/acme.go
@@ -184,7 +184,7 @@ func (a *ACME) leadershipListener(elected bool) error {
 		account := object.(*Account)
 		account.Init()
 		// Reset Account values if caServer changed, thus registration URI can be updated
-		if account != nil && account.Registration != nil && isAccountMatchingCaServer(account.Registration.URI, a.CAServer) {
+		if account != nil && account.Registration != nil && !isAccountMatchingCaServer(account.Registration.URI, a.CAServer) {
 			log.Info("Account URI does not match the current CAServer. The account will be reset")
 			account.reset()
 		}
--- a/acme/acme.go
+++ b/acme/acme.go
@@ -187,14 +187,14 @@ func (a *ACME) leadershipListener(elected bool) error {
 		if account != nil && account.Registration != nil {
 			aru, err := url.Parse(account.Registration.URI)
 			if err != nil {
-				log.Error("Unable to parse account.Registration URL")
+				return fmt.Errorf("Unable to parse account.Registration URL : %v", err)
 			}
 			cau, err := url.Parse(a.CAServer)
 			if err != nil {
-				log.Error("Unable to parse CAServer URL")
+				return fmt.Errorf("Unable to parse CAServer URL : %v", err)
 			}
 			if strings.Compare(cau.Hostname(), aru.Hostname()) != 0 {
-				log.Error("Account does not match the current CAServer")
+				log.Info("Account URI does not match the current CAServer. The account will be reset")
 				account.reset()
 			}
 		}
--- a/acme/acme.go
+++ b/acme/acme.go
@@ -184,19 +184,9 @@ func (a *ACME) leadershipListener(elected bool) error {
 		account := object.(*Account)
 		account.Init()
 		// Reset Account values if caServer changed, thus registration URI can be updated
-		if account != nil && account.Registration != nil {
-			aru, err := url.Parse(account.Registration.URI)
-			if err != nil {
-				return fmt.Errorf("Unable to parse account.Registration URL : %v", err)
-			}
-			cau, err := url.Parse(a.CAServer)
-			if err != nil {
-				return fmt.Errorf("Unable to parse CAServer URL : %v", err)
-			}
-			if strings.Compare(cau.Hostname(), aru.Hostname()) != 0 {
-				log.Info("Account URI does not match the current CAServer. The account will be reset")
-				account.reset()
-			}
+		if account != nil && account.Registration != nil && isAccountMatchingCaServer(account.Registration.URI, a.CAServer) {
+			log.Info("Account URI does not match the current CAServer. The account will be reset")
+			account.reset()
 		}
 
 		var needRegister bool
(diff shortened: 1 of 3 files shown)


provider/acme/provider.go
```diff
@@ -130,8 +131,19 @@ func (p *Provider) Init(_ types.Constraints) error {
 	}
 
 	// Reset Account if caServer changed, thus registration URI can be updated
-	if p.account != nil && p.account.Registration != nil && !strings.HasPrefix(p.account.Registration.URI, p.CAServer) {
-		p.account = nil
+	if p.account != nil && p.account.Registration != nil {
+		aru, err := url.Parse(p.account.Registration.URI)
+		if err != nil {
+			log.Error("Unable to parse account.Registration URL")
```
mmatur: You need to return an error

provider/acme/provider.go
```diff
@@ -130,8 +131,19 @@ func (p *Provider) Init(_ types.Constraints) error {
...
+		}
+		cau, err := url.Parse(p.CAServer)
+		if err != nil {
+			log.Error("Unable to parse CAServer URL")
```
mmatur: You need to return an error

acme/acme.go
```diff
@@ -183,8 +184,19 @@ func (a *ACME) leadershipListener(elected bool) error {
 		account := object.(*Account)
 		account.Init()
 		// Reset Account values if caServer changed, thus registration URI can be updated
-		if account != nil && account.Registration != nil && !strings.HasPrefix(account.Registration.URI, a.CAServer) {
-			account.reset()
+		if account != nil && account.Registration != nil {
+			aru, err := url.Parse(account.Registration.URI)
+			if err != nil {
+				log.Error("Unable to parse account.Registration URL")
```
mmatur: You need to return an error

acme/acme.go
```diff
@@ -183,8 +184,19 @@ func (a *ACME) leadershipListener(elected bool) error {
...
+			}
+			cau, err := url.Parse(a.CAServer)
+			if err != nil {
+				log.Error("Unable to parse CAServer URL")
```
mmatur: You need to return an error

provider/acme/provider.go
```diff
@@ -130,8 +131,19 @@ func (p *Provider) Init(_ types.Constraints) error {
...
+		}
+		if strings.Compare(cau.Hostname(), aru.Hostname()) != 0 {
+			log.Error("Account does not match the current CAServer")
```
mmatur: Could you please change the level of this log and improve the message with something like that 
```go
log.Info("Account URI does not match the current CAServer. The account will be reset")
```

mmatur: Could you please change the level of this log and improve the message with something like that 
```go
log.Info("Account URI does not match the current CAServer. The account will be reset")
```

acme/acme.go
```diff
@@ -183,8 +184,19 @@ func (a *ACME) leadershipListener(elected bool) error {
...
+			}
+			if strings.Compare(cau.Hostname(), aru.Hostname()) != 0 {
+				log.Error("Account does not match the current CAServer")
```
mmatur: Could you please change the level of this log and improve the message with something like that 
```go
log.Info("Account URI does not match the current CAServer. The account will be reset")
```

mmatur: Could you please change the level of this log and improve the message with something like that 
```go
log.Info("Account URI does not match the current CAServer. The account will be reset")
```

provider/acme/provider.go
```diff
@@ -130,8 +131,19 @@ func (p *Provider) Init(_ types.Constraints) error {
...
+	if p.account != nil && p.account.Registration != nil {
+		aru, err := url.Parse(p.account.Registration.URI)
+		if err != nil {
+			log.Error("Unable to parse account.Registration URL")
+		}
+		cau, err := url.Parse(p.CAServer)
+		if err != nil {
+			log.Error("Unable to parse CAServer URL")
+		}
+		if strings.Compare(cau.Hostname(), aru.Hostname()) != 0 {
```
mmatur: As you don't return error before - `cau` and `aru` can be nil

acme/acme.go
```diff
@@ -183,8 +184,19 @@ func (a *ACME) leadershipListener(elected bool) error {
...
+		if account != nil && account.Registration != nil {
+			aru, err := url.Parse(account.Registration.URI)
+			if err != nil {
+				log.Error("Unable to parse account.Registration URL")
+			}
+			cau, err := url.Parse(a.CAServer)
+			if err != nil {
+				log.Error("Unable to parse CAServer URL")
+			}
+			if strings.Compare(cau.Hostname(), aru.Hostname()) != 0 {
```
mmatur: As you don't return error before - `cau` and `aru` can be nil

zyclonite:
i have no proper go dev environment, it's a bit trail and error with simply a text editor...

as i do not know how to properly write tests in go, it would be nice if someone could support me here

acme/acme.go
```diff
@@ -230,6 +232,20 @@ func (a *ACME) leadershipListener(elected bool) error {
...
+	if err != nil {
+		log.Infof("Unable to parse account.Registration URL : %v", err)
+		return false
+	}
+	cau, err := url.Parse(serverURI)
+	if err != nil {
+		log.Infof("Unable to parse CAServer URL : %v", err)
+		return false
+	}
+	return strings.Compare(cau.Hostname(), aru.Hostname()) != 0
```
mmatur: You should replace by 
```go
return cau.Hostname() == aru.Hostname()
```

provider/acme/provider.go
```diff
@@ -142,6 +144,20 @@ func (p *Provider) Init(_ types.Constraints) error {
...
+	if err != nil {
+		log.Infof("Unable to parse account.Registration URL : %v", err)
+		return false
+	}
+	cau, err := url.Parse(serverURI)
+	if err != nil {
+		log.Infof("Unable to parse CAServer URL : %v", err)
+		return false
+	}
+	return strings.Compare(cau.Hostname(), aru.Hostname()) != 0
```
mmatur: You should replace by 
```go
return cau.Hostname() == aru.Hostname()
```

acme/acme.go
```diff
@@ -183,7 +184,8 @@ func (a *ACME) leadershipListener(elected bool) error {
 		account := object.(*Account)
 		account.Init()
 		// Reset Account values if caServer changed, thus registration URI can be updated
-		if account != nil && account.Registration != nil && !strings.HasPrefix(account.Registration.URI, a.CAServer) {
+		if account != nil && account.Registration != nil && isAccountMatchingCaServer(account.Registration.URI, a.CAServer) {
```
mmatur: It's should be `!isAccountMatchingCaServer(account.Registration.URI, a.CAServer)` instead of `isAccountMatchingCaServer(account.Registration.URI, a.CAServer)` no ?

provider/acme/provider.go
```diff
@@ -130,7 +131,8 @@ func (p *Provider) Init(_ types.Constraints) error {
 	}
 
 	// Reset Account if caServer changed, thus registration URI can be updated
-	if p.account != nil && p.account.Registration != nil && !strings.HasPrefix(p.account.Registration.URI, p.CAServer) {
+	if p.account != nil && p.account.Registration != nil && isAccountMatchingCaServer(p.account.Registration.URI, p.CAServer) {
```
mmatur: It's should be `!isAccountMatchingCaServer(account.Registration.URI, a.CAServer)` instead of `isAccountMatchingCaServer(account.Registration.URI, a.CAServer)` no ?

provider/acme/provider_test.go
```diff
@@ -429,3 +429,72 @@ func TestDeleteUnnecessaryDomains(t *testing.T) {
...
+			expected:   false,
+		},
+		{
+			desc:       "malformed account url",
+			accountURI: "/test.example.com/acme/acct/1234567",
+			serverURI:  "https://test.example.com/acme/directory",
+			expected:   false,
+		},
+		{
+			desc:       "malformed server url",
```
mmatur: Could you please add a test with 2 malformed url
Like that: 
```go
{
    desc:       "malformed account & server url",
    accountURI: "/test.example.com/acme/acct/1234567",
    serverURI:  "/test.example.com/acme/acct/1234567",
    expected:   false,
},
```

provider/acme/provider.go
zyclonite: true, my mistake

provider/acme/provider_test.go
zyclonite: problem is, it's not really malformed for the parser.. i'll find a different approach

PR #3670, opened 2018-07-24
Adding segments to ECS

terrycain:
## What does this PR do?
This PR is currently a work in progress. I'm mainly a Python developer so there is probably a fair bit that could be done better. The intention is to add in "segments" for ECS like there is with the docker backend. So therefore Traefik can make use of multiple ports exposed by an ECS container.

Here is a snippet of the docker labels I have in ECS on my test container
```
"traefik.enable": "true",
"traefik.service1.domain": "test.traefik",
"traefik.service1.port": "8888",
"traefik.service1.backend": "backend1",
"traefik.service1.frontend.rule": "Host:test.traefik",
"traefik.service2.domain": "test.traefik"
"traefik.service2.port": "4444",
"traefik.service2.backend": "backend2",
"traefik.service2.frontend.rule": "Host:test.traefik;Path:/test",
```
Attached is some sample cURL output of it working :D, a sample ecs task definition and the current output of /api/providers
[curl_output.txt](https://github.com/containous/traefik/files/2225546/curl_output.txt)
[ecs_taskdef.txt](https://github.com/containous/traefik/files/2225547/ecs_taskdef.txt)
[traefik_api_providers.txt](https://github.com/containous/traefik/files/2225548/traefik_api_providers.txt)

I could do with someone looking over this and ensuring I'm going about this in the right way and also any ideas on what else is needed would be helpful too :)

### More
- [ ] Add more tests
- [ ] Update documentation

### DIFF
--- a/provider/ecs/config_segment_test.go
+++ b/provider/ecs/config_segment_test.go
@@ -1,12 +1,12 @@
 package ecs
 
 import (
-	"testing"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/containous/traefik/provider/label"
 	"github.com/containous/traefik/types"
 	"github.com/stretchr/testify/assert"
-	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/containous/traefik/provider/label"
-	"github.com/aws/aws-sdk-go/service/ecs"
+	"testing"
 )
 
 func TestSegmentBuildConfiguration(t *testing.T) {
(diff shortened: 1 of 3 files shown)


provider/ecs/config_segment_test.go
```diff
@@ -0,0 +1,99 @@
+package ecs
+
+import (
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ecs"
+	"github.com/containous/traefik/provider/label"
+	"github.com/containous/traefik/types"
+	"github.com/stretchr/testify/assert"
+	"testing"
```
mmatur: Could you please sort your import

traefiker:
Closed by #3817.

PR #3674, opened 2018-07-25
Docs verify failed only on 4XX

mmatur:
### What does this PR do?

This PR allow Docs verify to failed only on 4XX


dduportal:
LGTM

PR #3711, opened 2018-08-01
Improve Connection Limit Kubernetes Documentation

dtomcej:
### What does this PR do?

Fixes a mistake in documentation.

### Motivation

Example documentation does not have an annotation value quoted. Since annotations must be strings, this annotation just gets skipped by the kubernetes API server, as it is parsed as an integer when unquoted. Quoting the number solves the issue.

### More

- [x] Added/updated tests - none needed, documentation update


errm:
Perhaps we should add a note before the list of annotations to this effect ... it's useful information for debugging because its a simple typo to make whenever the value is a number ....

PR #3742, opened 2018-08-07
Extend https redirection tests, and fix incorrect behavior

### LINKED ISSUE #3741: Wrong behaviour of HTTP -> HTTPS redirect with request-modifier (ReplacePathRegex)

Bug

### What did you do?

There's been a long-standing issue with preserving paths when redirecting from HTTP to HTTPS (https://github.com/containous/traefik/issues/1957), but it was finally fixed in https://github.com/containous/traefik/pull/3631 by @dtomcej. This is great stuff and I can confirm that `PathPrefixStrip` works properly in `1.7.3-rc3` 🎉 

However, it looks like the problem is still out there when we use `ReplacePathRegex`.

Here's an ingress annotation that works correctly:

```yaml
kind: Ingress
metadata:
  name: example-ingress
  annotations:
    traefik.ingress.kubernetes.io/rule-type: PathPrefixStrip
spec:
  rules:
  - host: example.com
    http:
      paths:
      - path: /backend-path
        backend:
          serviceName: example-backend
          servicePort: http
```

```bash
http http://example.com/backend-path/abc
# HTTP/1.1 302 Found
# Content-Length: 5
# Content-Type: text/plain; charset=utf-8
# Date: Mon, 06 Aug 2018 20:55:37 GMT
# Location: https://example.com/backend-path/abc
# 
# Found
```

And here's the same rule idea, but implemented via `request-modifier` / `ReplacePathRegex`:

```yaml
kind: Ingress
metadata:
  name: example-ingress
  annotations:
    traefik.ingress.kubernetes.io/request-modifier: "ReplacePathRegex: ^/backend-path-2/(.*) /$1"
spec:
  rules:
  - host: example.com
    http:
      paths:
      - path: /backend-path-2
        backend:
          serviceName: example-backend
          servicePort: http
```

```bash
http http://example.com/backend-path-2/abc
# HTTP/1.1 302 Found
# Content-Length: 5
# Content-Type: text/plain; charset=utf-8
# Date: Mon, 06 Aug 2018 20:55:40 GMT
# Location: https://example.com/abc
# 
# Found
```

I expected `"ReplacePathRegex: ^/backend-path-2/(.*) /$1"` work just like `PathPrefixStrip`, but it acted as before https://github.com/containous/traefik/pull/3631.

```
official helm chart, traefik image tag: `1.7.0-rc3-alpine`
```

---
**PS**: I switched to `ReplacePathRegexp` in an attempt to configure both path prefix stripping and adding trailing slash in one ingress rule (see [my stackoveflow question](https://stackoverflow.com/questions/51697574/adding-trailing-slash-to-microservice-root-path-in-traefik-ingress)). Not sure if it's the best solution though, regardless of this issue with HTTP → HTTPS redirects.

dtomcej:
### What does this PR do?

Rebuilds https redirection paths after being modified by middlewares

### Motivation

Fixes #3741 

### More

- [x] Add `replacePath` to context
- [x] Add tests
- [x] Add `replacePathRegex` to context
- [x] Add tests
- [x] Fix `stripPathPrefix`
- [x] Added/updated tests
- [x] Added/updated documentation - None needed, corrects intended behavior


kachkaev:
Many thanks for putting this together @dtomcej! Does this PR fix a double-trailing slash when the ingress path ends with a slash? https://github.com/containous/traefik/issues/3741#issuecomment-410958581

I looked through your changes, but could not find a new test case for this, so am curious. Sorry in advance if that's a lame question – I don't know Go, so asking is the only option I have 😆 

In any case, I'll be keen to test the result once the PR is merged! Is `containous/traefik:experimental` the right image to pull if I can't wait for `1.7.0` / `1.7.0-rc4`?

kachkaev:
I have just tried `dtomcej/traefik:issue-3741` and can confirm that I no longer reproduce  https://github.com/containous/traefik/issues/3741 and https://github.com/containous/traefik/issues/3741#issuecomment-410958581 🎉 

Many thanks for your work on this @dtomcej 🙌 💯

integration/https_test.go
```diff
@@ -751,114 +752,92 @@ func (s *HTTPSSuite) TestEntrypointHttpsRedirectAndPathModification(c *check.C)
 
 	testCases := []struct {
 		desc        string
-		host        string
+		host        []string
```
mmatur: Could you please rename `host` into `hosts`

integration/https_test.go
```diff
@@ -751,114 +752,92 @@ func (s *HTTPSSuite) TestEntrypointHttpsRedirectAndPathModification(c *check.C)
...
-			host:        "foo.com",
-			sourceURL:   "http://127.0.0.1:8888/api/bacon/",
-			expectedURL: "https://foo.com:8443/api/bacon/",
+			expectedURL: "https://<HOST>:8443/wtf/foo",
 		},
 	}
 
 	for _, test := range testCases {
-		test := test
+		for _, subtest := range test.host {
```
mmatur: Could you please rename `subtest` into `host`

integration/https_test.go
```diff
@@ -751,114 +752,92 @@ func (s *HTTPSSuite) TestEntrypointHttpsRedirectAndPathModification(c *check.C)
...
-		req.Host = test.host
+			resp, err := client.Do(req)
+			c.Assert(err, checker.IsNil)
+			defer resp.Body.Close()
 
-		resp, err := client.Do(req)
-		c.Assert(err, checker.IsNil)
-		defer resp.Body.Close()
+			location := resp.Header.Get("Location")
+			expected := strings.Replace(test.expectedURL, "<HOST>", subtest, 1)
```
mmatur: You should maybe replace by 
```go
expected := fmt.Sprintf(test.expectedURL, host)
```

And replace all `<HOST>` occurrences in `expectedURL` by `%s`

integration/https_test.go
```diff
@@ -740,7 +741,7 @@ func (s *HTTPSSuite) TestEntrypointHttpsRedirectAndPathModification(c *check.C)
 	defer cmd.Process.Kill()
 
 	// wait for Traefik
-	err = try.GetRequest("http://127.0.0.1:8080/api/providers", 500*time.Millisecond, try.BodyContains("Host: example.com"))
+	err = try.GetRequest("http://127.0.0.1:8080/api/providers", 1000*time.Millisecond, try.BodyContains("Host: example.com"))
```
Juliens: `1*time.Second`?

PR #3773, opened 2018-08-16
Wf test old template

hsmade:
This is a modification of https://github.com/containous/traefik/pull/3598


hsmade:
This PR has been replaced by https://github.com/containous/traefik/pull/3781

Issue #3754, opened 2018-08-09
ACME key seems to not persist


atombender:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I've started with *fresh config*. I don't have an `acme.json` and don't use the migration path. I start out with an empty Etcd.

I'm running into a problem where the ACME key specified in `storage = ...` is not persisted in Etcd. It seems Traefik gets certs every time it starts. I have:

```
storage = "traefik/acme/account"
```

but this key does not ever exist:

```
$ etcdctl get --prefix / --keys-only | grep acme | grep -v '^$'
/traefik/acme/
/traefik/acme/acmelogging
/traefik/acme/caserver
/traefik/acme/delaydontcheckdns
/traefik/acme/dnsprovider
/traefik/acme/domains/0/main
/traefik/acme/email
/traefik/acme/entrypoint
/traefik/acme/httpchallenge/
/traefik/acme/httpchallenge/entrypoint
/traefik/acme/keytype
/traefik/acme/ondemand
/traefik/acme/onhostrule
/traefik/acme/overridecertificates
/traefik/acme/storage
```

I use `storeconfig` to update the config in Etcd. This seems to work fine. (Though the storage key isn't created at that point, nor did I expect it to be.)

There's nothing interesting in the logs except things like `Datastore sync error: object lock value: expected 05254333-722a-4557-a320-0a12bda7645c, got 686712f1-fe57-4bb0-9438-0f06f3a54db9, retrying in 743.158272ms`.

I start Traefik without a config file, and specify the bootstrapping params (`--etcd.endpoint=...` and so on) on the command line. I don't use a config file once I've bootstrapped Etcd. I only use `storeconfig` to update it.

When Traefik starts up, it requests certs from Let's Encrypt. However, the key for the ACME state is never saved. The next time Traefik starts up, it has no state, and so requests everything from LE again.

### What did you expect to see?

I expected Traefik to persist the ACME account data.

### What did you see instead?

No ACME account data persisted to Etcd.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
$ /traefik version
Version:      v1.7.0-rc3
Codename:     maroilles
Go version:   go1.10.3
Built:        2018-08-01_01:37:51PM
OS/Arch:      linux/amd64
```

#### etcd (v3 only enabled, v2 disabled):

```
$ etcdctl version
etcdctl version: 3.3.9
API version: 3.3
```

#### Config file (used with `traefik storeconfig`)

[...]

Issue #1512, opened 2017-04-29
Add an export ACME command


ldez:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you want?

Add a new command to Traefik (like `traefik exportACME /etc/ssl`).
It would allow to export certs even in cluster mode (certs from the KV store).

Keridos:
Would really love to see that, better even if traefik could automatically dump certain certificates in docker labels whenever they get updated.

dtomcej:
@ldez  This was merged with #1484.

Can we close this ticket?

alwinmarkcf:
@dtomcej unfortunately not. Its fixing only half of this feature request, not this part:
> It would allow to export certs even in cluster mode (certs from the KV store).

Also your tip at slack with the encryption was not right or I'm too dumb to extract those values. When I do a:

`etcdctl get traefik/acme/account --prefix`
Im just getting random chars in lines. It would be at least helpful to find out how these values are stored. Are they some kind of serialized?

alwinmarkcf:
@dtomcej so this evolves to a show stopper for us. We now went two times in a row offline (luckily but bad enough only internal tools yet) because we had issues with the certs.

I even can't look into this deeper as it seems that the value of `traefik/acme/account/object` is just a blob and most properly serialized go structs (as far as I can tell).

While beeing able to export those certs, I could inspect whats going wrong and I would have a fallback/backup meachnism. So exporting the certs value as acme.json would be fine.


Also I'm wondering if https://github.com/containous/traefik/issues/3754 relates to this. As I don't know whether the client key is stored in that blob for me it's possible that this could be the cause of that issue.

If I can help with anything I'm willing to, unfortunately I'm not very familiar with go code, but in the meantime I try my best to figure out whats happening there.

alwinmarkcf:
Ok I guess I found the issue

Currently the acme.json is stored as key/values:
```
/traefik/acme/storage
traefik/acme/account/lock
traefik/acme/account/object
```

while the account struct looks like [this](https://github.com/containous/traefik/blob/master/cluster/datastore.go#L21-L23):
```golang
type Metadata struct {
	object Object
	Object []byte
	Lock   string
}
```
_Note that etcd does not care about capital letters._

the serialization works like this:
```golang
func (m *Metadata) Marshall() error {
	var err error
	m.Object, err = json.Marshal(m.object)
	return err
}

func (m *Metadata) unmarshall() error {
	if len(m.Object) == 0 {
[...]

Issue #1325, opened 2017-03-22
kv value too large when using consul backend and acme


Paxxi:
### What version of Traefik are you using (`traefik version`)?
v.1.1.2

### What is your environment & configuration (arguments, toml...)?
docker swarm 1.13

<details>

```yml
Containers: 5                                                   
 Running: 2                                                     
 Paused: 0                                                      
 Stopped: 3                                                     
Images: 33                                                      
Server Version: 1.13.0                                          
Storage Driver: overlay                                         
 Backing Filesystem: extfs                                      
 Supports d_type: true                                          
Logging Driver: json-file                                       
Cgroup Driver: cgroupfs                                         
Plugins:                                                        
 Volume: glusterfs local                                        
 Network: bridge host macvlan null overlay                      
Swarm: active                                                   
 NodeID: c8rqvh23p2kqqpbyj9ydf5xxy                              
 Is Manager: true                                               
 ClusterID: n1td1rbeiwcokhp7gglg6z4gy                           
 Managers: 3                                                    
 Nodes: 3                                                       
 Orchestration:                                                 
  Task History Retention Limit: 5                               
 Raft:                                                          
  Snapshot Interval: 10000                                      
  Number of Old Snapshots to Retain: 0                          
  Heartbeat Tick: 1                                             
  Election Tick: 3                                              
 Dispatcher:                                                    
  Heartbeat Period: 5 seconds                                   
 CA Configuration:                                              
  Expiry Duration: 3 months                                     
 Node Address: 10.135.25.157                                    
 Manager Addresses:                                             
  10.135.25.157:2377                                            
  10.135.36.196:2377                                            
  10.135.8.136:2377                                             
Runtimes: runc                                                  
Default Runtime: runc                                           
Init Binary: docker-init                                        
containerd version: 03e5862ec0d8d3b3f750e19fca3ee367e13c090e    
runc version: 2f7393a47307a16f8cee44a37b262e8b81021e3e          
init version: 949e6fa                                           
Security Options:                                               
 apparmor                                                       
 seccomp                                                        
  Profile: default                                              
Kernel Version: 4.4.0-59-generic                                
Operating System: Ubuntu 16.04.1 LTS                            
OSType: linux                                                   
Architecture: x86_64                                            
CPUs: 2                                                         
Total Memory: 3.859 GiB                                         
Name: ohmy-rancher1                                             
[...]

Issue #3760, opened 2018-08-15
Health status for certbot


alwinmarkcf:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

The most fragile part I experienced with Traefik is the Certbot, so I now always have mixed feelings not knowing if everything is working correctly and e.g certs get updated properly. Parsing logs is timeconsuming and not quite clear.

In order to know the status of the certbot and the certs traefik is managing
As an administrator
I would like to have a a section in the dashboard and health endpoint showing me:

* If a certbot client has been created, which is useable or if client creation rate limit is active
* A list of certs with valid to date and if exist, rate limitations (either on the health page or directly on the frontend view)

PR #3776, opened 2018-08-17
Store serialized JSON ACME Account as String

alwinmarkcf:
### What does this PR do?

Before it was stored as []byte which make it hard to backup ACME
Certificates and debug the behaviour of getting new Certificates. So this PR changes this behaviour and stores the serialized JSON ACME Account as String

### Motivation

Basically the motivation were these two issues, as for a productive usage of traefik it's a must have to backup certs. #1512

Also I have several problems during an upgrade process with kubernetes I can't really debug without knowing what happens there: #3754

### More

- [ ] Added/updated tests

### Additional Notes

Actually I don't know why this Meta object is used for the Account. Guess using [staert](https://github.com/containous/staert) with the Account struct itself could remove the limitation: 
> !!! note It's possible to store up to approximately 100 ACME certificates in Consul.

Otherwise it would make more sense to have a dedicated storeMethod/Transferobject for the Account or certificate.

Described here: https://github.com/nmengin/traefik/blob/357d8762a16405a881d72907c39f61d78372a26f/docs/configuration/acme.md
Issued here: https://github.com/containous/traefik/issues/1325

Also this may bring up again the topic about: https://github.com/containous/traefik/issues/1325 but where #1325 makes it hard having multiple certs, the current behaviour makes it impossible to backup certs which prevents the usage of traefik in production

### DIFF
--- a/cluster/datastore.go
+++ b/cluster/datastore.go
@@ -19,7 +19,7 @@ import (
 // Metadata stores Object plus metadata
 type Metadata struct {
 	object Object
-	Object []byte
+	Object string
 	Lock   string
 }
 
@@ -30,8 +30,10 @@ func NewMetadata(object Object) *Metadata {
 
 // Marshall marshalls object
 func (m *Metadata) Marshall() error {
+	var byteArray []byte
 	var err error
-	m.Object, err = json.Marshal(m.object)
+	byteArray, err = json.Marshal(m.object)
+	m.Object = string(byteArray)
 	return err
 }
 
@@ -39,7 +41,7 @@ func (m *Metadata) unmarshall() error {
 	if len(m.Object) == 0 {
 		return nil
 	}
-	return json.Unmarshal(m.Object, m.object)
+	return json.Unmarshal([]byte(m.Object), m.object)
 }
 
 // Listener is called when Object has been changed in KV store


alwinmarkcf:
I don't know how the process is for merging into multiple branches.

atombender:
Doesn't this need to have a fallback to the old byte format so that it can read the existing data that people currently have stored in their clusters?

emilevauge:
Hey @alwinmarkcf 
Thanks for your contribution!

There are 2 major regressions brought by this fix:
- this is not backward compatible with the current implementation
- more importantly, this removes the compression we added https://github.com/containous/traefik/pull/2814, which allows to store up to approximately 100 ACME certificates

The second one is probably a deal breaker here, as it will not be possible to get the best of both worlds (store many certificates AND human readable format).

In order to help users debug things, we could instead add a new command to Traefik, something like `traefik debugConfig`, that would read the kv store, and print the content on std out. WDYT?

atombender:
Could there be (or is there) a flag to toggle compression? Etcd has no such limit on the size of the value, so the binary format is of no use there.

atombender:
Also, why not just gzip the JSON? The curren object appears to be a binary structure. If it were gzipped JSON, you could just do `etcdctl get --print-value-only /traefik/acme/account/object | gzip -dc` to read it.

alwinmarkcf:
@emilevauge so in general not specific to this PR I have following usecases, which are in my opinion needed to make Traefik really "cloudnative" and production ready.

Unfortunately I kind of love this project and the idea behind it and we as crowdfox put some effort into it, so it would be awesome if we could help bringing traefik into this direction.

_Actually I don't know where to post this without loosing the overview as its more on an epic level then an issue. Maybe you can move it to the right place._

## Helm install

```gherkin
Given a proper values.yaml for HA mode
When I do a "helm install"
Then  traefik should run in HA mode
```
This one works since we got: https://github.com/helm/charts/pull/6813 merged

## Upgrade

### simple upgrade

```gherkin
Given a proper values.yaml for HA mode
And traefik is installed via helm in HA mode
When I change some parameter in the values.yaml
Then I these changes should take effect
```

For this one, I'd love to see what's discussed here: https://github.com/containous/traefik/issues/3754
Tl;dr:
* Keep configs in TOML and only store states other than configs in KV Store (Like ACME Certs). 
* make Traefik start with empty kv set and generate them on the fly

### Idempotent upgrade

```gherkin
Given a proper values.yaml for HA mode
And traefik is installed via helm in HA mode
When I change some parameter in the values.yaml
And I install on a different cluster traefik with the same values.yaml from scratch
Then both traefik server should behave the same
```

Would be fixed when only the config file would be considered. The Problem here currently is that storeconfig does not clean Keys, so you have legacy key/values which you only can clean manually. Thats a huge Problem. Eg: `traefik/acme/domains`

## Backup

### create backup

```gherkin
Given traefik is installed 
When I run a backup command
Then I want to be able to backup all states which are not given by:
* Providers
* Configuration config.tom
(means especially ACME certificates)
```
This would work merging this Ticket, even though it would strongly reduce max number of certs for consul. But then you could backup your config running:
```bash
kubectl run  --rm -it traefik-etcd-test --image=quay.io/coreos/etcd:v3.1.8 --restart=Never --env="ETCDCTL_API=3" -l app=traefik -- etcdctl --endpoints=http://traefik-etcd-cluster-client:2379 get traefik/acme/domains --prefix --print-value-only | jq .
```

### import backup

```gherkin
Given traefik is installed
When I run an import backup command (e.g. storeconfig)
Then all states given by this backup should be restored (especially ACME certs)
```
This is some kind of working now, using the storeconfig job, https://github.com/helm/charts/pull/7173, and paramter: `--set traefik.kvprovider.importAcme=true`

## ACME Troubleshooting

### Refetch certificate

Given an invalid/lost cert
When I run update cert for domain "example.dev"
Then traefik should try to get a new cert for this domain

While having the ability to export the config and import it back you could do this also manually.

### Healthcheck

https://github.com/containous/traefik/issues/3760

This would be really nice to have a check running if certs are too old and can't be refetched because of ratelimits or a broken letsencrypt client (key)

nmengin:
Hello @atombender,

> Also, why not just gzip the JSON? 

Træfik has used the library `staert` to insert data into the KV stores.
`staert` stores the strings in the KV stores without modifying the data, it only can compress the `[]byte` and modifying this behavior can be hard thats why the JSON is compressed as a `[]byte` and not a `string`.

nmengin:
Hello @alwinmarkcf,

Many thanks for the PR and all the propositions you suggested to improve Træfik.
I guess, it should be better to open an issue per propositions, thus we can analyse and prioritize them.

Concerning the PR, as @emilevauge said previously, the modifications you propose require more work to be backward compatible (like supporting a third format for the ACME Account Key).

That's why I just opened a new issue #3847 to create a subcommand to export the data you need without changing nothing in Træfik.
A PR will come to implement it.

Once again, thanks for your work and all the propositions you have made in this PR and the others issues.

Closed in favor of #3847.

Issue #2712, opened 2018-01-16
Acme process cannot found TLS key on Etcd3


yboyer:
### Do you want to request a *feature* or report a *bug*?
bug

### What did you do?
I created a docker-compose with Træfik and Etcd3. Since Træfik v1.5.0-rc1, I got a `Key not found in store` when I enable Acme.
It works on acme disabled and with a storage file as a store instead of etcd too.

_PS: It worked on Træfik v1.4 and Etcd2_

### What did you expect to see?
Træfik to start

### What did you see instead?
```
etcd_1     | 2018-01-16 18:48:49.672552 I | etcdmain: etcd Version: 3.2.13
etcd_1     | 2018-01-16 18:48:49.672696 I | etcdmain: Git SHA: 95a726a
traefik_1  | time="2018-01-16T18:48:51Z" level=info msg="Using TOML configuration file /etc/traefik/traefik.toml"
etcd_1     | 2018-01-16 18:48:49.672728 I | etcdmain: Go Version: go1.8.5
etcd_1     | 2018-01-16 18:48:49.672754 I | etcdmain: Go OS/Arch: linux/amd64
etcd_1     | 2018-01-16 18:48:49.672769 I | etcdmain: setting maximum number of CPUs to 4, total number of available CPUs is 4
etcd_1     | 2018-01-16 18:48:49.672928 I | embed: listening for peers on http://0.0.0.0:2380
traefik_1  | time="2018-01-16T18:48:51Z" level=warning msg="web provider configuration is deprecated, you should use these options : api, rest provider, ping and metrics"
traefik_1  | time="2018-01-16T18:48:51Z" level=info msg="Traefik version v1.5.0-rc5 built on 2018-01-15_03:59:03PM"
etcd_1     | 2018-01-16 18:48:49.673032 I | embed: listening for client requests on 0.0.0.0:2379
etcd_1     | 2018-01-16 18:48:49.675664 I | pkg/netutil: resolving etcd:2380 to 172.22.0.2:2380
traefik_1  | time="2018-01-16T18:48:51Z" level=info msg="
traefik_1  | Stats collection is disabled.
traefik_1  | Help us improve Traefik by turning this feature on :)
traefik_1  | More details on: https://docs.traefik.io/basic/#collected-data
traefik_1  | "
[...]

PR #3779, opened 2018-08-17
Include missing key in error message for KV store

camelpunch:
### What does this PR do?

The generic missing-key error message that comes from valkeyrie doesn't point the operator at the key that's missing. Several folk, including myself, have spent time figuring out what that missing key is. When setting up TLS there's a further potential confusion between KV store keys and TLS keys!

See https://github.com/containous/traefik/issues/2712 for an example issue that might have gone smoother with this more detailed logging.

### Motivation

I spent several hours debugging this, and that time could've been saved by knowing what the missing key was.

### More

No tests / docs updated. I don't believe this warrants a test.

I began running the tests but had to go do something else before they finished!

### Additional Notes

Worth checking that the format string is safe with all possible return values.

### DIFF
--- a/cluster/datastore.go
+++ b/cluster/datastore.go
@@ -78,7 +78,7 @@ func (d *Datastore) watchChanges() error {
 	stopCh := make(chan struct{})
 	kvCh, err := d.kv.Watch(d.lockKey, stopCh, nil)
 	if err != nil {
-		return err
+		return fmt.Errorf("%v: %s", err, d.lockKey)
 	}
 	safe.Go(func() {
 		ctx, cancel := context.WithCancel(d.ctx)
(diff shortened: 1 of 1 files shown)


cluster/datastore.go
```diff
@@ -78,7 +78,7 @@ func (d *Datastore) watchChanges() error {
 	stopCh := make(chan struct{})
 	kvCh, err := d.kv.Watch(d.lockKey, stopCh, nil)
 	if err != nil {
-		return err
+		return fmt.Errorf("%v: %s", err, d.lockKey)
```
Juliens: Can you add a sentence here like `error while watching key %s: %v`

ProPheT777:
Why not simply bootstrap etcd with lock and object when the etcd entry is present ? It's so much complicated to "pre bootstrap" etcd by playing one time script ... Don't understand this choice to throw an error that say "you need to bootstrap" rather than auto bootstrap

PR #3781, opened 2018-08-17
Add segments to the consulcatalog provider

hsmade:
### What does this PR do?

This adds segments to the consulcatalog provider. It does this in the same way as the docker provider does.

### Motivation

https://github.com/containous/traefik/issues/3597
We are facing the issue that our consul services have routes for multiple domains and share the fqdn between services. This means we would need more than one `rule` to express the routes for each service. This PR enables us to do so.

### More

- [X] Added/updated tests

### Additional Notes

I implemented the segments using the existing `label.ExtractTraefikLabels`. I envision that the frontends which are the result of one single service have both the original service name and the segment name in their name: traefik.rule1.frontend.rule:*** on service `myservice` leads to a frontend named frontend-myservice-rule1. All the frontends which are generated out of a service, share the same backend, to minimise resources. For this I needed to adjust `getServiceBackendName` to use the original service name as a name, which I store in `serviceUpdate.ParentServiceName`.
I added a test that expands a single service into 2 frontends, by having 2 segments.


hsmade:
After discussion with @ldez , this PR is is closed in favor of https://github.com/containous/traefik/pull/3796

PR #3787, opened 2018-08-19
Add a note about overrideCertificates to acme docs

camelpunch:
### What does this PR do?

Add an extra note to the docs indicating that `overrideCertificates` will create some required keys when using a KV store to store TLS certs.

There's an [existing comment in the config file example](https://github.com/containous/traefik/commit/4a3b1f384750606c3fe77de3b1e4905aaa81b3ea#diff-9891ec9a4af35901191c75a98627b9e7R66), but it doesn't mention the specific keys that might be missing when a user runs into the log message in [my other PR](https://github.com/containous/traefik/pull/3779).

Although it introduces a coupling between docs and code, I think that the key names should be present in the docs so that the solution appears in search / browser find results.

### Motivation

Some sunk time figuring out how to do HA ACME with etcd and idempotent config. Ended up learning about this setting via a [Slack conversation](https://traefik.slack.com/archives/C34HHN6JZ/p1534513726000100?thread_ts=1534460459.000100&cid=C34HHN6JZ).

### More

### DIFF
--- a/docs/configuration/acme.md
+++ b/docs/configuration/acme.md
@@ -421,6 +421,8 @@ storage = "traefik/acme/account"
 
 Because KV stores (like Consul) have limited entry size the certificates list is compressed before it is saved as KV store entry.
 
+If using [storeconfig](/user-guide/kv-config/#store-configuration-in-key-value-store), consider using `acme.overrideCertificates = true` to create the `acme/account` and `acme/lock` keys required for certificate storage.
+
 !!! note
     It is possible to store up to approximately 100 ACME certificates in Consul.
 


alwinmarkcf:
This is really a problem currently with the 1.7 release and maybe the most frequent question at the moment in the kubernetes slack channel.

camelpunch:
I have a suspicion that `overrideCertificates` will cause further support headaches, since it always overwrites certs, meaning you need two separate `storeconfig`s:

1. With `overrideCertificates`, to set the keys that cause havoc when unset
2. Without `overrideCertificates`, to avoid wiping away previously obtained certs and account data.

I think the code should create or ignore missing keys instead. That way operators won't have to perform manual steps or write conditional code.

alwinmarkcf:
So just taking Kubernetes Helm Chart usecases:
1. new installation (Revision == 0) => it should "override" the certificates
2. "upgrade" traefik => it should NOT override certificate
3. import certificates from backup => it should override the certificates using the storeconfigfile

I guess this is pretty clear and sufficient. Splitting the parameter would in my opinion just lead to further on top complexity for usecases like:
* Splitting SAN Certificates
* Merging SAN Certificates

camelpunch:
Could (1) be achieved by just ignoring the missing keys and letting ACME happen? That way you'd only need to 'override' when you were actually overriding existing certs.

That said, this is a complex piece of software with a lot of moving parts. I defer to someone with more context!

geraldcroes:
Hi, and once again thank you for the pull request.

After having discussed it, we deem that fixing the current behaviour is a better alternative (see #3844) and will avoid people struggling with the first initialization of the object.

That's why this PR will be closed in favor of #3844.

Thank you for taking the time to point an inconsistent behaviour, it is now fixed!

camelpunch:
Even better - thanks!

PR #3809, opened 2018-08-21
Enable CORS configuration

### LINKED ISSUE #3686: set Access-Control-Allow-Origin header according to origin header
Feature

It seems that traefik only support set Access-Control-Allow-Origin to a fixed value such as "Access-Control-Allow-Origin: *", "Access-Control-Allow-Origin: https://google.com" via traefik.frontend.headers.customResponseHeaders.

It would be better if we can set Access-Control-Allow-Origin to origin if origin satisfy requirements.
for example:

origin: http://domain1.google.com  -> Access-Control-Allow-Origin: http://domain1.google.com 
origin: http://domain2.google.com  -> Access-Control-Allow-Origin: http://domain2.google.com

dtomcej:
### What does this PR do?

Enables dynamic CORS configuration, based on (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) and (https://www.w3.org/TR/cors/#access-control-allow-origin-response-header)

### Motivation

Security is good. CORs are good. CORs security headers are better.

Also:
Fixes #3686 

### More

- [x] Added/updated tests
  - [x] Preflight Tests
  - [x] Request Tests
  - [x] Response Tests

### Additional Notes

Due to the complicated nature of CORs, I need some more eyes on this PR to see if I have handled the logic correctly...


dtomcej:
This PR is still a WIP, but due to some large configuration PRs that are going to be merged, I will close this, and Re-Open in the future.

shane-axiom:
I do hope you pick this back up when the timing is right. Looks promising!

dtomcej:
Now that the config PRs are merged, reopening this PR!

dtomcej:
Yay! Tests Passing!

dtomcej:
Needs `Vary: Origin` implemented.

middlewares/headers/headers.go
```diff
@@ -102,33 +108,117 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
-	s.next.ServeHTTP(rw, req)
+func (s *Header) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	reqAcMethod := req.Header.Get("Access-Control-Request-Method")
+	reqAcHeaders := req.Header.Get("Access-Control-Request-Headers")
+	s.originHeader = req.Header.Get("Origin")
+
+	if reqAcMethod != "" && reqAcHeaders != "" && s.originHeader != "" && req.Method == http.MethodOptions {
+		// Preflight request, build response
+		if s.headers.AccessControlAllowCredentials {
+			rw.Header().Add("Access-Control-Allow-Credentials", "true")
```
ldez:
```suggestion
			rw.Header().Set("Access-Control-Allow-Headers", allowHeaders)
```

middlewares/headers/headers.go
```diff
@@ -102,33 +108,117 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		}
+
+		allowHeaders := strings.Join(s.headers.AccessControlAllowHeaders, ",")
+		if allowHeaders != "" {
+			rw.Header().Add("Access-Control-Allow-Headers", allowHeaders)
```
ldez:
```suggestion
			rw.Header().Set("Access-Control-Allow-Headers", allowHeaders)
```

middlewares/headers/headers.go
```diff
@@ -102,33 +108,117 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		}
+
+		allowMethods := strings.Join(s.headers.AccessControlAllowMethods, ",")
+		if allowMethods != "" {
+			rw.Header().Add("Access-Control-Allow-Methods", allowMethods)
```
ldez:
```suggestion
			rw.Header().Set("Access-Control-Allow-Methods", allowMethods)
```

middlewares/headers/headers.go
```diff
@@ -102,33 +108,117 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		}
+
+		allowOrigin := s.getAllowOrigin()
+
+		if allowOrigin != "" {
+			rw.Header().Add("Access-Control-Allow-Origin", allowOrigin)
```
ldez:
```suggestion
			rw.Header().Set("Access-Control-Allow-Origin", allowOrigin)
```

middlewares/headers/headers.go
```diff
@@ -102,33 +108,117 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
 
-func (s *header) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	s.modifyRequestHeaders(req)
-	s.next.ServeHTTP(rw, req)
+func (s *Header) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	reqAcMethod := req.Header.Get("Access-Control-Request-Method")
+	reqAcHeaders := req.Header.Get("Access-Control-Request-Headers")
+	s.originHeader = req.Header.Get("Origin")
+
+	if reqAcMethod != "" && reqAcHeaders != "" && s.originHeader != "" && req.Method == http.MethodOptions {
```
ldez: Could you add a comment to explain this `if`

middlewares/headers/headers.go
```diff
@@ -102,33 +108,117 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+		}
+
+		allowOrigin := s.getAllowOrigin()
+
+		if allowOrigin != "" {
+			rw.Header().Add("Access-Control-Allow-Origin", allowOrigin)
+		}
+
+		rw.Header().Add("Access-Control-Max-Age", strconv.Itoa(int(s.headers.AccessControlMaxAge)))
+	} else {
```
ldez: could you add a `return` instead of `else`

middlewares/headers/headers.go
```diff
@@ -102,33 +108,117 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	switch s.headers.AccessControlAllowOrigin {
+	case "origin-list-or-null":
+		if s.originHeader == "" {
+			return "null"
+		}
+		return s.originHeader
+	case "*":
+		return "*"
+	}
+	return ""
```
ldez:
```go
func (s *Header) getAllowOrigin() string {
	switch s.headers.AccessControlAllowOrigin {
	case "origin-list-or-null":
		if s.originHeader == "" {
			return "null"
		}
		return s.originHeader
	case "*":
		return "*"
	default:
		return ""
	}
}

```

middlewares/headers/headers_test.go
```diff
@@ -4,6 +4,7 @@ package headers
 
 import (
 	"context"
+	"github.com/containous/traefik/tracing"
```
ldez: could you sort imports.

responsemodifiers/headers.go
```diff
@@ -4,46 +4,43 @@ import (
 	"net/http"
 
 	"github.com/containous/traefik/config"
+	"github.com/containous/traefik/middlewares/headers"
 	"github.com/unrolled/secure"
 )
 
-func buildHeaders(headers *config.Headers) func(*http.Response) error {
+func buildHeaders(h *config.Headers) func(*http.Response) error {
```
ldez: could you restore the variables name `headers` because it can be confused with a receiver name.

middlewares/headers/headers.go
```diff
@@ -102,33 +108,117 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+
+	if allowOrigin != "" {
+		res.Header.Set("Access-Control-Allow-Origin", allowOrigin)
+
+		if s.headers.AddVaryHeader {
+			varyHeader := res.Header.Get("Vary")
+			if varyHeader != "" {
+				varyHeader += ",Origin"
+			} else {
+				varyHeader = "Origin"
```
ldez:
```go
		if s.headers.AddVaryHeader {
			varyHeader := res.Header.Get("Vary")
			if varyHeader != "" {
				varyHeader += ","
			}
			varyHeader += "Origin"
			
			res.Header.Set("Vary", varyHeader)
		}
```

responsemodifiers/headers.go
dtomcej: No, because if we do, there is a naming conflict between `config.Headers` and `middlewares.headers`, resulting in the error:

`headers.NewHeader undefined (type *config.Headers has no field or method NewHeader)`

dtomcej: Could we name it something else so there is no conflict? like `hdr`?

ldez: ok

thomas15v:
Any change that this can be backported to V1.7?

jeff1985:
@dtomcej 
can you fix the confilicting docs/basics.md ?

pkg/middlewares/headers/headers.go
```diff
@@ -102,33 +108,121 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	// Loop through Custom response headers
+	for header, value := range s.headers.CustomResponseHeaders {
+		if value == "" {
+			res.Header.Del(header)
+		} else {
+			res.Header.Set(header, value)
+		}
+	}
+
+	allowOrigin := s.getAllowOrigin()
```
dtomcej: This is tied to a different instance.

https://github.com/containous/traefik/pull/3809/files#diff-8ab61f315c59ef6e9b44f0bebf69ee71R128

This is not set on this instance.

pkg/middlewares/headers/headers.go
```diff
@@ -102,33 +108,121 @@ func (s secureHeader) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+	return nil
+}
+
+func (s *Header) getAllowOrigin() string {
+	switch s.headers.AccessControlAllowOrigin {
+	case "origin-list-or-null":
+		if s.originHeader == "" {
+			return "null"
+		}
+		return s.originHeader
```
dtomcej: This is set on request, not on response.

docs/content/middlewares/headers.md
```diff
@@ -82,7 +82,35 @@ This functionality allows for some easy security features to quickly be set.
...
+
+### CORS Headers
+
+CORS (Cross-Origin Resource Sharing) headers can be added and configured per frontend in a similar manner to the custom headers above.
+This functionality allows for more advanced security features to quickly be set.
+
+??? example "File"
+
+    ```toml    
+[http.middlewares]
```
geraldcroes: incorrect indentation (won't render properly)

Wirone:
@dtomcej how allowed origins can be handled while using `origin-list-or-null`? I would like to add `ACCESS-CONTROL-ALLOW-ORIGIN` for specified origins only, for other do not allow CORS. I didn't find example for it, is it possible?

jeff1985:
@dtomcej I also dont understand how the origin-list is supposed to work? Because there is no config option for a whitelist, the effect of "origin-list-or-null" will be the same as "*", as any domain will be allowed to send requests. Am I overlooking something?

PR #3851, opened 2018-09-04
Avoid goroutine leak in server

nmengin:
### What does this PR do?

The PR allows the goroutine `ListenSignals` to listen to the server context.
Thus, this goroutine will be stopped when the server is stopped.

### Motivation

Improve the quality of the code by making it more robust.

### Additional Notes

The PR contains a test on the configMessage in the ThrottleChan to avoid a possible `panic` when the cast is done on a nil referenced object.


jbdoumenjou:
LGTM

PR #3854, opened 2018-09-05
Guarantee that only one traceId in http headers

joway:
### What does this PR do?

- Guarantee that only one tracing Id in HTTP headers.

### Motivation

When I using traefik as my reverse proxy backend and tracing request by jaeger. I found that client cannot get the correct tracing context, and all requests' headers have two values on one key `uber-trace-id`. So client may just get the old traceId and not drop the one trafik inject.

### More

- Bugfix


geraldcroes:
Hi and thank you for your efforts in improving Traefik 👍 

In your PR, it seems that you remove the header only if the backend name is jaeger ... which seems more like a workaround than an actual fix.

Could you be more specific about the use case so we can understand better what you're trying to achieve?

joway:
@geraldcroes  Hi, when I use traefik as my reverse proxy,

My first request like:
```
Headers:
uber-trace-id: xxx

POST http://traefik.com/xxx
```
When traefik got the request, and proxy to upstream , it will inject tracing headers on request object and the result like:

```
Headers:
uber-trace-id: xxx
uber-trace-id: bbb

POST http://upstream/xxx
```

Because the inject method does not ensure that there is only one key in headers .
So , I want delete the header `uber-trace-id` before inject .

joway:
I have reviewed the jaeger go client codes and I think the trace id header should be removed in outside caller.

mmatur:
Hi @joway,

Thanks for your feedback. After a lot of investigation. We have found the issue.

It is an issue in [opentracing/opentracing-go](https://github.com/opentracing/opentracing-go) ant it has been fixed by https://github.com/opentracing/opentracing-go/pull/191 but for now we are not able to upgrade this dependency. 

You fixed was just for jaeger and this issue is present on all tracer. This is why I preferred to work on more generic fix and I submitted a new PR #3878.

This fix will be included in the 1.7 Træfik version

Thanks

joway:
@mmatur LGTM! Thanks!

traefiker:
Closed by #3878.

PR #3857, opened 2018-09-05
Multiple kubernetes ingress definitions may cause flapping frontend

rtreffer:
### What does this PR do?

If multiple ingress objects define the same frontend (host + path) and one of the ingresses is broken (e.g. referenced service does not exist) then the frontend will be flapping (appear, disappear, ...).

### Motivation

We found this in production and I've extracted the problem into a testcase...

Given 2 ingresses:
```
- apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
    namespace: testing
    name: web-a
  spec:
    rules:
    - host: host-a
      http:
        paths:
        - backend:
            serviceName: service1
            servicePort: 80
- apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
    namespace: testing
    name: web-b
  spec:
    rules:
    - host: host-a
      http:
        paths:
        - backend:
            serviceName: missing
            servicePort: 80
```

The `missing` service will cause the whole `host-a` frontend to disappear if the ingress objects are processed in order `[ingress/web-a, ingress/web-b]`. We have seen in production that the ordering can be random, thus causing the frontend to become unavailable.

### More

- [x] Added/updated tests

### Additional Notes

Instead if writing to the `Frontends` map directly this patch checks out the current `frontend` and writes it to the map only if basic validations pass.

There was a second test-failure after applying this patch. The `root2` testcase has 2 ingress definitions, one of those is broken.
The behavior now changed to accept the valid one and keep the frontend.

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -242,7 +242,10 @@ func (p *Provider) loadIngresses(k8sClient Client) (*types.Configuration, error)
 					continue
 				}
 
-				if _, exists := templateObjects.Frontends[baseName]; !exists {
+				var frontend *types.Frontend
+				if fe, exists := templateObjects.Frontends[baseName]; exists {
+					frontend = fe
+				} else {
 					auth, err := getAuthConfig(i, k8sClient)
 					if err != nil {
 						log.Errorf("Failed to retrieve auth configuration for ingress %s/%s: %s", i.Namespace, i.Name, err)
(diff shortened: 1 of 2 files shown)


timoreimann:
Definitely looks like a bug to me we should fix.

@containous/traefik @dtomcej  okay if we mark this as a bug (fix) and assign a proper priority?

nmengin:
Hello @rtreffer ,

As your PR is a bugfix, can you rebase it on the `v1.6` branch please?

rtreffer:
@nmengin sure!

rtreffer:
@nmengin / @timoreimann there are unfortunately some changes needed to put it onto v1.6 and I've thus opened a new PR: https://github.com/containous/traefik/pull/3862

It also did not require the second commit from this branch: https://github.com/containous/traefik/pull/3857/commits/5b5e8dc61f277b17dfddd8e63a6ad764d6d2dab0

rtreffer:
I'm closing this one (just for reference of the testcase fix that might be needed)

PR #3862, opened 2018-09-06
Avoid flapping of multiple Ingress definitions

rtreffer:
### What does this PR do?

If multiple ingress objects define the same frontend (host + path) and one of the ingresses is broken (e.g. referenced service does not exist) then the frontend will be flapping (appear, disappear, ...).

### Motivation

We found this in production and I've extracted the problem into a testcase...

Given 2 ingresses:
```yml
- apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
    namespace: testing
    name: web-a
  spec:
    rules:
    - host: host-a
      http:
        paths:
        - backend:
            serviceName: service1
            servicePort: 80
- apiVersion: extensions/v1beta1
  kind: Ingress
  metadata:
    namespace: testing
    name: web-b
  spec:
    rules:
    - host: host-a
      http:
        paths:
        - backend:
            serviceName: missing
            servicePort: 80
```

The `missing` service will cause the whole `host-a` frontend to disappear if the ingress objects are processed in order `[ingress/web-a, ingress/web-b]`. We have seen in production that the ordering can be random, thus causing the frontend to become unavailable.

### More

- [x] Added/updated tests

### Additional Notes

Instead if writing to the `Frontends` map directly this patch checks out the current `frontend` and writes it to the map only if basic validations pass.

There was a second test-failure after applying this patch. The `root2` testcase has 2 ingress definitions, one of those is broken.
The behavior now changed to accept the valid one and keep the frontend.

This is a rebase of https://github.com/containous/traefik/pull/3857 on v1.6
This version does not require the following testcase patch: https://github.com/containous/traefik/pull/3857/commits/5b5e8dc61f277b17dfddd8e63a6ad764d6d2dab0

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -772,7 +772,7 @@ spec:
 Træfik will now look for cheddar service endpoints (ports on healthy pods) in both the cheese and the default namespace. Deploying cheddar into the cheese namespace and afterwards shutting down cheddar in the default namespace is enough to migrate the traffic.
 
 !!! note
-   The kubernetes documentation does not specify this merging behavior. The reference nginx implementation has undefined behavior: If two ingress objects define the same host/port then one of them will randomly win on reload.
+   The kubernetes documentation does not specify this merging behavior.
 
 !!! note
    Merging ingress definitions can cause problems if the annotations differ or if the services handle requests differently. Be careful and extra cautious when running multiple overlapping ingress definitions.
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -746,9 +746,11 @@ You should now be able to visit the websites in your browser.
 
 Træfik will merge multiple Ingress definitions for the same host/path pair into one definition.
 
-Let's say the number of cheese services is growing. It is now time to move the cheese services to a dedicated cheese namespace to simplify the managements of cheese and non-cheese services.
+Let's say the number of cheese services is growing.
+It is now time to move the cheese services to a dedicated cheese namespace to simplify the managements of cheese and non-cheese services.
 
 Simply deploy a new Ingress Object with the same host an path into the cheese namespace:
+
 ```yaml
 apiVersion: extensions/v1beta1
 kind: Ingress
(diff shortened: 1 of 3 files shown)


rtreffer:
Applying this PR on newer branches may require https://github.com/containous/traefik/pull/3857/commits/5b5e8dc61f277b17dfddd8e63a6ad764d6d2dab0

dtomcej:
@timoreimann I guess the only question I have is: If you have two identical ingresses witht he same host/path combo, but different services, are we supposed to merge the backends if they are valid?

If so, then this resolution is fine. Otherwise, we should define what _should_ happen.

I agree having the frontend appear/disappear is not ideal, but that doesn't mean that merging all overlaps is the only other option.

Thoughts?

timoreimann:
IMHO, the host/path tuple is a unique identifier for an Ingress'ed resource. As such, I feel it'd be natural to merge -- it seems like there should be no difference between an Ingress spec holding all backends and two Ingress specs holding all backends together.

The only alternative I could see is to reject Ingresses that refer an already existing host/path pair, or make things configurable. I don't really have a use case for that, other than to protect Ingress objects from each other. If people want that, however, they can control access to Ingress resources more tightly (which is what my org does) or run separate Traefik instances per namespace.

@dtomcej WDYT?

dtomcej:
Sorry for the delayed response.

I agree @timoreimann that the tuple _should_ be unique. In the case at hand, the annotations from the ingress resource first loaded would be applied. I do think that this case may be noteworthy as far as documentation goes, since it could lead to non-deterministic configuration.

rtreffer:
Sure. Merging is not new (it is already happening on 2 service backed ingress definitions) but I couldn't find it in the docs. I'll add a paragraph about it in the kubernetes user-guide.

docs/user-guide/kubernetes.md
```diff
@@ -742,6 +742,41 @@ You should now be able to visit the websites in your browser.
...
+      - path: /cheddar
+        backend:
+          serviceName: cheddar
+          servicePort: http
+```
+
+Træfik will now look for cheddar service endpoints (ports on healthy pods) in both the cheese and the default namespace. Deploying cheddar into the cheese namespace and afterwards shutting down cheddar in the default namespace is enough to migrate the traffic.
+
+!!! note
+   The kubernetes documentation does not specify this merging behavior. The reference nginx implementation has undefined behavior: If two ingress objects define the same host/port then one of them will randomly win on reload.
```
timoreimann: I'd omit the description pertaining to Nginx's behavior (second sentence to the end of the paragraph) as it may change in the future, leaving our docs outdated without us being able to notice the fact easily.

rtreffer: Removed, added it to explain the motivation, but that's not helpful for someone just reading the docs :man_shrugging:

PR #3878, opened 2018-09-11
Fix tracing duplicated headers

mmatur:
### What does this PR do?

This PR fix duplicated header in Tracing.

### Motivation

When an incoming request arrives with a tracing header, Træfik should used this header to create the new tracing headers and keep only the last one.

### Additional Notes

Closes #3854 

It is an issue in [opentracing/opentracing-go](https://github.com/opentracing/opentracing-go) ant it has been fixed by https://github.com/opentracing/opentracing-go/pull/191 but for now we are not able to upgrade this dependency

### DIFF
--- a/middlewares/tracing/carrier.go
+++ b/middlewares/tracing/carrier.go
@@ -1,0 +1,25 @@
+package tracing
+
+import "net/http"
+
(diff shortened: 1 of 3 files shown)


jbdoumenjou:
LGTM

PR #3885, opened 2018-09-13
Add missing quotes around backendName in kv template

NatMarchand:
### What does this PR do?

This PR fixes the KV template with missing quotes around $backendName.

### Motivation

The generated templates are wrong and therefore backends containing dot in the name doesn't work well.
Here is an excerpt of the difference in a generated config (note the quotes for loadbalancer section whereas no quotes for healthCheck) :
```
[backends.\"backend.consul\".loadBalancer]
      method = \"drr\"
      sticky = false
[backends.backend.consul.healthCheck]
    path = \"/\"
    port = 0
    interval = \"10s\"
```


NatMarchand:
Thx for your quick approval ! It was my pleasure to contribute :)

PR #3891, opened 2018-09-14
Added two example / Using labels in docker-compose.yml

pascalandy:
### What does this PR do?

Doc: Added two example / Using labels in docker-compose.yml

### Motivation

https://github.com/containous/traefik/issues/563#issuecomment-421479401

### More

null

### Additional Notes

https://github.com/containous/traefik/issues/563#issuecomment-421479401

### DIFF
--- a/docs/user-guide/examples.md
+++ b/docs/user-guide/examples.md
@@ -371,7 +371,9 @@ deploy:
 
 Something more tricky using `regex`.
 
-In this case a slash is added to `siteexample.io/portainer` and redirect to `siteexample.io/portainer/`. For more details: https://github.com/containous/traefik/issues/563 
+In this case a slash is added to `siteexample.io/portainer` and redirect to `siteexample.io/portainer/`. For more details: https://github.com/containous/traefik/issues/563
+
+The double sign `$$` are variables managed by the docker compose file ([documentation](https://docs.docker.com/compose/compose-file/#variable-substitution)). 
 
 ```
 portainer:
--- a/docs/user-guide/examples.md
+++ b/docs/user-guide/examples.md
@@ -329,3 +329,85 @@ providersThrottleDuration = "5s"
 [respondingTimeouts]
 idleTimeout = "360s"
 ```
+
+## Using labels in docker-compose.yml
+
+Pay attention to the **labels** section:
+
+```
+home:
+image: abiosoft/caddy:0.10.14
+networks:
+  - ntw_front
+volumes:
+  - ./www/home/srv/:/srv/
+deploy:
+  mode: replicated
+  replicas: 2
+  #placement:
+  #  constraints: [node.role==manager]
+  restart_policy:
+    condition: on-failure
+    max_attempts: 5
+  resources:
+    limits:
+      cpus: '0.20'
+      memory: 9M
+    reservations:
+      cpus: '0.05'
+      memory: 9M
+  labels:
+  - "traefik.frontend.rule=PathPrefixStrip:/"
+  - "traefik.backend=home"
+  - "traefik.port=2015"
+  - "traefik.weight=10"
+  - "traefik.enable=true"
+  - "traefik.passHostHeader=true"
+  - "traefik.docker.network=ntw_front"
+  - "traefik.frontend.entryPoints=http"
+  - "traefik.backend.loadbalancer.swarm=true"
+  - "traefik.backend.loadbalancer.method=drr"
+```
+
+Something more tricky using `regex`.
+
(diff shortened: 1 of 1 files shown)


MindTooth:
Maybe add a small notice regarding `$$`.  So that there is an explicit mention.

Nice PR otherwise.  Been missing such section a long time. 👍

pascalandy:
I don't know what to say regarding the `$$`. Please enlighted me :)

MindTooth:
I'll look into maybe writing some text.  I think the official documentation have some great stuff though.

Source: https://docs.docker.com/compose/compose-file/#variable-substitution

pascalandy:
Thanks, I'll add a note about it. Need to run this morning!

ldez:
@pascalandy could you rebase on v1.7 and change the base branch of the PR to v1.7 (edit button on the top of the PR)

pascalandy:
Add more details and switched to branch 1.7
Hope everything is good now :)

docs/user-guide/examples.md
```diff
@@ -329,3 +329,87 @@ providersThrottleDuration = "5s"
 [respondingTimeouts]
 idleTimeout = "360s"
 ```
+
+## Using labels in docker-compose.yml
```
dduportal: `Using Labels with Docker-Compose`

docs/user-guide/examples.md
```diff
@@ -329,3 +329,87 @@ providersThrottleDuration = "5s"
...
+  - "traefik.passHostHeader=true"
+  - "traefik.docker.network=ntw_front"
+  - "traefik.frontend.entryPoints=http"
+  - "traefik.backend.loadbalancer.swarm=true"
+  - "traefik.backend.loadbalancer.method=drr"
+```
+
+Something more tricky using `regex`.
+
+In this case a slash is added to `siteexample.io/portainer` and redirect to `siteexample.io/portainer/`. For more details: https://github.com/containous/traefik/issues/563
```
dduportal: * `In this case, a slash ...` => Missing comma
* `... and redirects...` => missing `s` for the verb

docs/user-guide/examples.md
```diff
@@ -329,3 +329,87 @@ providersThrottleDuration = "5s"
...
+
+The double sign `$$` are variables managed by the docker compose file ([documentation](https://docs.docker.com/compose/compose-file/#variable-substitution)).
```
dduportal: The double dollar sign `$$` allows escaping: it means a litteral value of `$` to prevent interpolation of the variable by docker-compose: `$MY_VAR` is interpolated, while `$$MY_VAR` is not.

docs/user-guide/examples.md
pascalandy: Looks good!

docs/user-guide/examples.md
pascalandy: Thanks!

docs/user-guide/examples.md
pascalandy: You're the boss!

pascalandy:
I agree!

> would it be possible to create a new User Guide's page named "Using Labels with Docker-Compose"?

Let's start be merging this first :)

dduportal:
Hello @pascalandy , after a discussion with other maintainers, it sounds like that instead of a very specific example in the user guide section (which is already "complicated"),
updating the section https://docs.traefik.io/configuration/backends/docker/#using-docker-compose with a note specifying the `$$` trick could be a simple but valuable information.

We fear the maintenance effort of a full `docker-compose.yml` as you provided (there are already too much in the User Guides ;) ) could be a pain.


Is there another intent in this user guide you wanted to underline (a use case)? If, could you details the intent?

dduportal:
Hello @pascalandy , my apologies for the delay of response.

Thanks for this contribution. It is LGTM.

We might want to rework example section later, but right now it is ok.

docs/user-guide/examples.md
```diff
@@ -371,7 +371,9 @@ deploy:
 
 Something more tricky using `regex`.
 
-In this case a slash is added to `siteexample.io/portainer` and redirect to `siteexample.io/portainer/`. For more details: https://github.com/containous/traefik/issues/563 
+In this case a slash is added to `siteexample.io/portainer` and redirect to `siteexample.io/portainer/`. For more details: https://github.com/containous/traefik/issues/563
```
MindTooth: Should this not be `redirected` instead?

pascalandy: correct redirected

PR #3896, opened 2018-09-17
Add tags label to Docker provider documentation

artheus:
### What does this PR do?

Adds documentation for the `traefik.tags` in Docker provider documentation

### Motivation

I had to go through the source code of Træfik for understanding how to use the tags label. I did not find any documentation explaining this label.

### Additional Notes

This documentation should be added to the other Providers as-well.
Critical for me was only the Docker provider.


artheus:
An alternative to my change here, would be to add a more detailed documentation of the `traefik.tags` label in the `common` documentation page. This would then allow users to read about how to actually use tags and constraints with any of the supported providers listed.

PR #3900, opened 2018-09-18
Remove hop-by-hop headers from forward auth response

stffabi:
### What does this PR do?

Hop-By-Hop headers from a ForwardAuth response get removed.

### Motivation

Hop-By-Hop headers from a ForwardAuth response (e.g. Redirect) should never be passed to the receiver. Failing to do so results in problems with Safari, which shows a "NSPOSIXErrorDomain:100" error if a ForwardAuth on a HTTP-1.1 server sends a redirect.

### More

- [x] Added/updated tests


stffabi:
Thanks all of you for this really fast merge...

PR #3928, opened 2018-09-25
ReplacePathRegex: Support adding query parameters in replacement path

holygeek:
### What does this PR do?

Support adding/replacing query parameters for `ReplacePathRegex` modifier.

### Motivation

https://stackoverflow.com/questions/48545586/traefik-how-to-append-parameters-to-path/52489528

> There is no way to do what you want. (@ldez)

So naturally I had no choice but to do it ;)

### More

- [x] Added/updated tests

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -88,7 +88,7 @@ Following is the list of existing modifier rules:
 
 - `AddPrefix: /products`: Add path prefix to the existing request path prior to forwarding the request to the backend.
 - `ReplacePath: /serverless-path`: Replaces the path and adds the old path to the `X-Replaced-Path` header. Useful for mapping to AWS Lambda or Google Cloud Functions.
-- `ReplacePathRegex: ^/api/v2/(.*) /api/$1`: Replaces the path with a regular expression and adds the old path to the `X-Replaced-Path` header. Separate the regular expression and the replacement by a space.
+- `ReplacePathRegex: ^/api/v3/(.*) /api/$1`: Replaces the path with a regular expression and adds the old path to the `X-Replaced-Path` header. Separate the regular expression and the replacement by a space. Query parameters can also be replaced, for example `/api/$1?foo=bar`. The original query parameters can be retained using `%{QUERY_STRING}`, for example `/api/$1?%{QUERY_STRING}&foo=bar`.
 
 #### Matchers
 
--- a/middlewares/replace_path_regex.go
+++ b/middlewares/replace_path_regex.go
@@ -33,7 +33,24 @@ func (s *ReplacePathRegex) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if s.Regexp != nil && len(s.Replacement) > 0 && s.Regexp.MatchString(r.URL.Path) {
 		r = r.WithContext(context.WithValue(r.Context(), ReplacePathKey, r.URL.Path))
 		r.Header.Add(ReplacedPathHeader, r.URL.Path)
-		r.URL.Path = s.Regexp.ReplaceAllString(r.URL.Path, s.Replacement)
+		if strings.Contains(s.Replacement, "?") {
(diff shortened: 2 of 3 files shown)


dtomcej:
My first concern with this is that Query Parameters are not part of Paths as per RFC 3986: (https://tools.ietf.org/html/rfc3986#section-3.4). 

I would prefer to see a new modifier that modifies only Query params like `ReplaceQueryRegex`

The second concern is that we don't want to use apache style replacements, but that is a post-design review discussion.

cention-nazri:
Thank you for taking the time to look into this pull request.

Some of the rewrite rules for the apache proxy that I'm attempting to replace with traefik look like the following:

    RewriteRule e/([0-9]+)/a/([0-9]+)    http://server:4242/e/a1?e=$1&a=$2 [P]
    RewriteRule e/([0-9]+)/w/a/([0-9]+)  http://server:4242/e/a2?e=$1&a=$2 [P]
    RewriteRule w/v/a/([0-9]+)           http://server:4242/e/a3?v=$1      [P]
    RewriteRule w/u/a/([0-9]+)           http://server:4242/e/a4?a=$1      [P]
    RewriteRule n/([0-9]+)/a/([0-9]+)    http://server:4242/e/a5?n=$1&a=$2 [P]

Here the original path is changed into a combination of path and query parameters. At the moment this is impossible to do with traefik because `ReplacePathRegex` modifies only the path.

I don't think a new modifier `ReplaceQueryRegex` that modifies only query params would be able to solve this too. The goal here is to transform the original path into a new path with query parameters whose names and values were components of the original path.

We'd have to use both `ReplacePathRegex` and `ReplaceQueryRegex` to do this. Let's try and see how it would look like if we already had `ReplaceQueryRegex`:

    [frontends.foo]
    backend = "bar"
    [frontends.foo.routes.match]
    rule = "Path: /e/{id1:[0-9]+}/a/{id2:[0-9]+}"
    [frontends.foo.routes.modifypath]
    rule = "ReplacePathRegex:.* /e/a1"
    [frontends.foo.routes.modifyquery]
    rule = "ReplaceQueryRegex: /e/([0-9]+)/a/([0-9]+) e=$1&a=$2"

If we want to preserve the existing query, here's how it could look like:

    rule = "ReplaceQueryRegex: /e/([0-9]+)/a/([0-9]+)\\?(.*) e=$1&a=$2&$3"

Some issues with this implementation are:

1. Does traefik currently support multiple modifiers in a single frontend?  I don't think it does, because I tried it before with multiple `ReplacePathRegex` in one frontend and it didn't seem to work so I had to break it down into multiple frontends.
2. The `ReplaceQueryRegex` has to know about path which don't constraint it to strictly on query parameters anymore, similar to how `ReplacePathRegex` goes into the query parameter territory in this pull request.

For comparison here's how the config would look like with what is implemented in this pull request:

    [frontends.foo]
    backend = "bar"
    [frontends.foo.routes.match]
    rule = "Path: /e/{id1:[0-9]+}/a/{id2:[0-9]+}"
    [frontends.foo.routes.modifypathandquery]
    rule = "ReplacePathRegex: /e/([0-9]+)/a/([0-9]+) /e/a1?e=$1&a=$2"

which looks simple, with a simpler implementation too, isn't it?

ldez:
Træfik support multiple modifiers in a single frontend.

I agree with @dtomcej: create a `ReplaceQueryRegex` modifier seems better.

cention-nazri:
> Træfik support multiple modifiers in a single frontend.

This is true only if the modifiers are distinct. That is if you have multiple `ReplacePathRegex` then it won't work (only one of the `ReplacePathRegex` would work, and the others seem to be ignored).

cention-nazri:
I have made a new pull request for adding a new `ReplaceQueryRegex`: https://github.com/containous/traefik/pulls

dduportal:
Closed in favor of #4000

Issue #3930, opened 2018-09-25
HostRegexp erroneously rejects hosts with capital letters


bgandon:
### Do you want to request a *feature* or report a *bug*?

This is a bug report for both `traefik` and the `mux` library.

### What did you do?

When using Træfik in front of Cloud foundry, I use a catch-all frontend rule, using a `dummy` variable:
```
            [frontends.cf-http-frontend.routes.catch-all-http]
              # Workaround for a catch-all rule:
              rule = "HostRegexp: {dummy:[-.0-9A-Z_a-z]+}"
```

When performing a request on the CF smoke-tests app (a basic Ruby web app), it works when the host is lowercased:
```
# curl -k https://smoke-tests.apps.easyfoundry.prototyp.it
Healthy
...
```

But when the host is capitalized, then it fails:
```
# curl -k http://SMOKE-TESTS.apps.easyfoundry.prototyp.it
404 page not found
```

And Træfik logs say `backend not found`:
```
"GET / HTTP/1.1" 404 19 "-" "curl/7.35.0" 10 "backend not found" "/" 0ms
```

### What did you expect to see?

I expect the request with capital letters to succeed, as it does when querying CF without Træfik in front of it:

```
# curl -k http://SMOKE-TESTS.apps.easyfoundry.prototyp.it
Healthy
...
```

### What did you see instead?

When the host is capitalized, the request fails:
```
# curl -k http://SMOKE-TESTS.apps.easyfoundry.prototyp.it
404 page not found
```

And Træfik logs say `backend not found`:
```
"GET / HTTP/1.1" 404 19 "-" "curl/7.35.0" 10 "backend not found" "/" 0ms
```

### Output of `traefik version`: (_What version of Traefik are you using?_)

I use version 1.6.5, as deployed by the [traefik-boshrelease](https://github.com/gstackio/traefik-boshrelease), but the incriminated piece of code hasn't changed in master since then, see below.

```
1.6.5
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

The bosh release deploys and configures Træfik v1.6.5 on Ubuntu, and currently I'm deploying on Ubuntu 14.04, but this is not relevant here.

### If applicable, please paste the log output in DEBUG level (`--logLevel=DEBUG` switch)

Not applicable.

### Code Analysis

[...]

PR #3931, opened 2018-09-26
Case insensitive host rule 

bgandon:
### What does this PR do?

This PR introduces a failing test for issue #3930.

Here we also rename parameters and variables in the `hostRegexp()` function to better reflect what is actually handled. Indeed, this function is written as if it was manipulating plain hostnames, whereas it actually handles host patterns that might include regular expressions.

### Motivation

I wanted to demonstrate the issue #3930, and also address a misnaming that was pointed out while analyzing that issue.

### More

- [x] Added/updated tests

### Additional Notes

Relates to #3930.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -291,7 +291,7 @@
   branch = "master"
   name = "github.com/containous/mux"
   packages = ["."]
-  revision = "06ccd3e75091eb659b1d720cda0e16bc7057954c"
+  revision = "c33f32e268983f989290677351b871b65da75ba5"
 
 [[projects]]
   name = "github.com/containous/staert"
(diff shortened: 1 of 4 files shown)


bgandon:
It's normal that tests are failing because the fix is in containous/mux#5.

juliens:
Thanks you for your PR.

I made a PR on https://github.com/containous/mux/pull/6.
With this PR we can remove the `strings.ToLower` https://github.com/containous/traefik/blob/6a55772cda1684546a6a5456b6847e0f9b3df44d/rules/rules.go#L59 and your tests should pass.

Do you think you can add a test to verify that the rule `HostRegex:CASE.test.com` match `case.test.com`, and that `HostRegex:case.test.com` match `CASE.test.com`?

Moreover, this a bug fix so I think we can merge it in the next 1.7. Can you make this PR on the 1.7 branch?

bgandon:
Thank you guy for merging!
Shall we close the related issue #3930?

PR #3938, opened 2018-09-27
Do not Errorf during file watcher verification test loop.

timoreimann:
### What does this PR do?

Change the logic in `file_test.go`'s `TestProvideWithWatch` verification loop to `t.Error` / `t.Fail` only if the timeout hits (that is, we fail to collect the desired configuration updates in time).

### Motivation

The loop in `TestProvideWithWatch` is meant to be executed multiple times until either all configuration elements are found or the timeout expires. By calling `assert.*` (such as `assert.Len`), however, an assertion failure in a non-terminal loop run will cause `t.Error` to be invoked, thereby marking the whole test as failed even though subsequent loop runs could still yield a positive test result overall.

### Additional Notes

This test has failed repeatedly (albeit not consistently) in our (HolidayCheck's) local CI. I couldn't exactly pinpoint the reason, but I suspect different concurrency behavior and/or CI machine specs. Still, the logic as it is implemented now seems incorrect for the reasons I outlined above (`t.Error` should only be called on terminal test errors), and thus the implementation change should be okay IMHO.


dtomcej:
@timoreimann Can you rebase on 1.7? Better tests in stable is a good thing ;)

timoreimann:
@dtomcej done!

PR #3942, opened 2018-09-27
Document the default accessLog format

dfredell:
### What does this PR do?

Add documentation for the default accessLog format on the logs doc. 

### Motivation

I wanted to know what the access log format was so I could parse and label the fields correctly in my centralized logger. Ex:
```
111.111.133.18 - - [27/Sep/2018:18:25:09 +0000] "GET / HTTP/1.1" 200 1022 "-" "Scalyr-monitor; https://www.scalyr.com/help/monitors" 1234 "frontend" "http://192.168.10.10:8080" 2ms
```

I had to read the code to see what the different values were. https://github.com/containous/traefik/blob/5acd43efaf076ae2204285e750e41976c29069f2/middlewares/accesslog/logger_formatters.go#L34

### DIFF
--- a/docs/configuration/logs.md
+++ b/docs/configuration/logs.md
@@ -60,12 +60,14 @@ For more information about the CLI, see the documentation about [Traefik command
 By default the Traefik log is written to stdout in text format.
 
 To write the logs into a log file specify the `filePath`:
+
 ```toml
 [traefikLog]
   filePath = "/path/to/traefik.log"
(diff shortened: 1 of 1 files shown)


dfredell:
@ldez, Are you not a fan of showing what values are quoted, like `"<Traefik frontend name>"`? I'm good either way, but it is just helpful when creating a parser to know that up front.

PR #3958, opened 2018-09-30
Correctly initialize kv store if storage key missing

jfrabaute:
### What does this PR do?

It looks like this commit might have introduced a bug
(I'm new to the code, so I might be wrong):
e4bb506#diff-89592b0a96bd24ccd41291548f9a07a3

On line 88, keyExists will return an error ErrKeyNotFound if the key does not exists,
because keyExists seems to not check if a key exists, but if a key is empty.

### More

- [ ] Added/updated tests


jfrabaute:
Great, thanks for the quick review and merge!

What is the usual timeframe between path review (like 1.7.2 which will have the fix)?

Going to use the containous/traefik image in the meantime to continue my tests.

PR #4015, opened 2018-10-10
acme: prevent some malformed errors.

### LINKED ISSUE #4011: Malformed ACME identifiers with trailing period

Bug

### What did you do?

N/A - I'm an employee of Let's Encrypt and noticed in our server-side logs that Traefik (versions 1.6.4 ... 1.7.2.) are responsible for a large portion of the `"urn:ietf:params:acme:error:malformed"` type error responses to newOrder requests. The majority of these have the detail message `"Error creating new order :: DNS name ends in a period"` indicating there is probably a common mis-configuration at work Traefik might be able to address.

The Traefik user-agent not being a dominant source of malformed errors in our logs :-)

### What did you see instead?

Many malformed errors being returned to Traefik users indicating that a newOrder request was received containing identifiers with a trailing period character, not a valid domain name.

I expect this is a result of browsers/HTTPS user-agents sending a request to a domain with a trailing period backed by a Traefik instance that doesn't strip the trailing period before creating an order with Lego/Let's Encrypt. It probably makes sense for Traefik to be responsible for stripping that trailing period instead of handling that corner-case in Lego.

N/A, in the last 3 days of logs I see versions 1.6.4 through 1.7.2 represented.

N/A

### If applicable, please paste the log output in DEBUG level (`--logLevel=DEBUG` switch)

N/A

ldez:
### What does this PR do?

Prevent some malformed errors.

### Motivation

Not being a dominant source of malformed errors in Let's Encrypt logs.

Fixes #4011

### More

- [-] Added/updated tests

### DIFF
--- a/provider/acme/provider.go
+++ b/provider/acme/provider.go
@@ -770,6 +770,7 @@ func (p *Provider) getValidDomains(domain types.Domain, wildcardAllowed bool) ([
 		}
 		cleanDomains = append(cleanDomains, cleanDomain)
 	}
+
 	return cleanDomains, nil
 }
 
--- a/provider/acme/provider.go
+++ b/provider/acme/provider.go
@@ -12,7 +12,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/BurntSushi/ty/fun"
 	"github.com/cenk/backoff"
 	"github.com/containous/flaeg"
 	"github.com/containous/traefik/log"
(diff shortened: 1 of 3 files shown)


provider/acme/provider.go
```diff
@@ -762,8 +761,16 @@ func (p *Provider) getValidDomains(domain types.Domain, wildcardAllowed bool) ([
...
+	var cleanDomains []string
+	for _, domain := range domains {
+		canonicalDomain := types.CanonicalDomain(domain)
+		cleanDomain := acme.UnFqdn(canonicalDomain)
+		if canonicalDomain != cleanDomain {
+			log.Warnf("FQDN detected, please remove the trailing dot: %s", canonicalDomain)
+		}
+		cleanDomains = append(cleanDomains, cleanDomain)
+	}
+	return cleanDomains, nil
```
mmatur: Could you please add a new line before

PR #4016, opened 2018-10-11
Add user guide for setting up a Docker Swarm mode cluster

tiangolo:
### What does this PR do?

Add user guide for setting up a Docker Swarm mode cluster with a Traefik main load balancer, handling HTTPS certificates listening on a filtering tag, exposing the web UI through HTTPS using a HTTP Basic Auth password.

### Motivation

I'm using a Docker Swarm mode cluster to deploy my applications. And Traefik is a crucial component in it, and inside the application stacks too.

I created this guide (as part of my project generator for full stack projects): https://github.com/tiangolo/full-stack/blob/master/docker-swarm-cluster-deploy.md

and I use that guide every time I'm setting up a new cluster to create the main Traefik proxy / load balancer. It has some tricks that were hard for me to discover, like exposing the web UI from Traefik through itself, using HTTPS and HTTP Basic Auth. So I thought it was worth sharing as part of the official docs.

I reduced it to cover the minimum as an atomic piece of content. And I haven't added links to it from the main pages because I would like to see your feedback first. 

I didn't add an example `docker-stack.yml` this setup because I don't know if it would be too much content for this guide. Or if it should be in another guide. I think it would be worth adding it somewhere, but please first let me know what you think of the current state of the guide. And how would you like me to modify or complement it.

### More

- [n/a] Added/updated tests

### Additional Notes

Some :cake: and Traefik :heart:

### DIFF
--- a/docs/user-guide/swarm-mode-https-web-ui.md
+++ b/docs/user-guide/swarm-mode-https-web-ui.md
@@ -1,0 +1,263 @@
+# Cluster setup with Docker Swarm Mode and Traefik
+
+This guide shows you how to create a cluster of Linux servers managed with Docker Swarm mode to deploy your projects.
+
+It also shows how to set up an integrated main Traefik load balancer / proxy to receive incoming connections, re-transmit communication to Docker containers based on the domains, generate TLS / SSL certificates (for HTTPS) with Let's Encrypt and handle HTTPS.
+
+And it shows how to expose the internal web UI through the same Traefik itself, with a HTTPS certificate and HTTP Basic Auth.
+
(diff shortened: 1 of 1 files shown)


dduportal:
Hell @tiangolo !

👏 Thanks for this awesome content and User Guide. This is really a neat piece of work and could help a lot of persons.

However we have an issue with such kind of documentation: it is maintenance effort. Given the frequency and amount of changes in areas such as Docker or Kubernetes, we cannot be sure that this content will be still up-to-date in 3, or 6 month, even tough the quality here is very good.

Our proposal is the following: Could you update the page [Awesome Træfik](https://github.com/containous/traefik/wiki/Awesome-Traefik) in the project's Wiki, to point out to your own resources? So it would still be very useful for end users, without puting the maintenance effort at risk. Bonus: your contribution would still be there! 

WDYT?

tiangolo:
Sure! Thanks for the feedback. 

I'll publish it as a blog post and then I'll come back with the link :smile:

PR #4020, opened 2018-10-11
Allow multiple ingress classes

rtreffer:
### What does this PR do?

Allow more than one ingress class to be specified. Why?
1. The default ingress matching for traefik is `["", "traefik"]` which is currently not configurable
2. Advanced setups may have traefik instances that handle a superset of another ingress (e.g. in our case an ingress controller that handles the `internal` and `public` ingress classes)

### Motivation

This is the last custom patch we are running in production. It is a rather small change that might be useful to others. I am also a fan of defaults that can be expressed through configuration.

Our use-case is a setup where we run 2 ingress pools for our applications: a public pool for external traffic and an internal one for internal traffic (often tools for humans).
However the internal one does route the external names, too. This is needed as some services go through the internal ingress infrastructure as a shortcut.

### More

- [x] Updated tests
- [ ] Added tests

### Additional Notes

I expect this PR to need more tests and a docs update but I wanted to get the idea out (for discussion).


dtomcej:
Out of curiosity, is this not something that the label selector could handle for you?

traefiker:
Thanks to your contribution.

As there is no feedback, and this PR is old, I will close it.
Feel free to reopen an issue or a pull request.

Issue #3272, opened 2018-05-06
traefik doesn't detect/load changed certificates


RonnyMaas:
### What did you do?
created a file backend with self-signed certificates

### What did you expect to see?
That traefik detects and reloads new/changed certificates. (after touch of traefik.toml) 

### What did you see instead?

New/Changed certificates are not detect and are not getting loaded.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
1.5.4 and 1.6.0

Version:      v1.6.0
Codename:     tetedemoine
Go version:   go1.10.1
Built:        2018-04-30_09:28:44PM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
[file]
watch = true
logLevel = "DEBUG"
accessLogsFile = "/dev/stdout"

[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    entryPoint = "https"

  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]

[backends]
  [backends.backend]
    [backends.backend1]
    url = "http://127.0.0.1:8080"
    weight = 2

[frontends]
  [frontends.frontend1]
  backend = "backend1"
  entrypoints = ["http", "https"]
    [frontends.frontend1.routes.test]

[[tls]]
  entryPoints = ["http", "https"]
  [tls.certificate]
    certFile = "/etc/traefik/frontend1.crt"
    keyFile = "/etc/traefik/frontend1.key"
```

### If applicable, please paste the log output at DEBUG level (`--logLevel=DEBUG` switch)

```
openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout ./frontend1.key -out ./frontend1.crt -subj '/C=US/ST=Oregon/L=Portland/CN=CERT2'

In logs:
level=info msg="Skipping same configuration for provider file"
```

```
openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout ./frontend1.key -out ./frontend1.crt -subj '/C=US/ST=Oregon/L=Portland/CN=CERT1'

touch traefik.toml

[...]

Issue #3083, opened 2018-03-27
Detect TLS cert changes automatically


spawnia:
### Do you want to request a *feature* or report a *bug*?

This could be seen as both, dynamically updating certificates on change seems like it would surely make Traefik better.

### What did you do?

I am dynamically providing certificates, which i generate by myself and put onto the server. Those should be hot-reloaded on change.

### What did you expect to see?

When switching out a certificate on the disk, i would expect traefik to pick up on this change and start delivering the new certificate.

### What did you see instead?

The new certificate only got applied after either restarting traefik or making a change in the dynamic config file, which causes traefik to update.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.5.4
Codename:     cancoillotte
Go version:   go1.9.4
Built:        2018-03-15_01:35:21PM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

CLI
````
docker run
  -v=/var/run/docker.sock:/var/run/docker.sock
  -v=/home/core/traefik-config:/traefik-config
  -p=80:80
  -p=443:443
  -p=8080:8080
  traefik
  --web
  --docker
  --docker.domain=XXX
  --docker.exposedbydefault=false
  --file.directory=traefik-config
  --file.watch=true
  --defaultEntryPoints='http,https'
  --entryPoints='Name:http Address::80'
  --entryPoints='Name:https Address::443 TLS'
````

Contents of the mounted folder:
````
````

```toml
# (paste your configuration here)
```

### If applicable, please paste the log output at DEBUG level (`--logLevel=DEBUG` switch)

```
(paste your output here)
```

nmengin:
Hello @spawnia .

If I understand correctly your use case, I presume it's not a bug but the expected behavior.
Indeed, Træfik is only looking for modifications in all the TOML files contained in`file.directory`.
That's why modifications in the other files are not automatically detected.

Do you want to propose a feature to listen to the certificate files modifications?

spawnia:
Surely that would be a great feature to have. I can not think of a use case where you would update the certificate files but would like to hold off on publishing those changes.

Reflecting the change immediately would make sense and go nicely with traefik's killer feature, the way it can dynamically update its configuration.

dtomcej:
One concern here is that if you update the certificate using a new key, the keypair will be mismatched for a period of time, which will cause a TLS failure. I'm wondering if there should be a delay after a triggered update to prevent this.

[...]

PR #4022, opened 2018-10-11
Check for dynamic tls updates on configuration preload

ffilippopoulos:
### What does this PR do?
Changes to look for tls certificate changes during `preLoadConfiguration` function. The aim is to solve issue: https://github.com/containous/traefik/issues/3272 and make reload configuration because of cert changes when `touch` toml files.

### Motivation

We are hitting this issue because we run an HA traefik setup in a kubernetes cluster and we use `cert-manager` to manage/update certificates. Each time a certificate is updated we have to go and restart traefik instances atm.

### More

- [x] Added/updated tests
added one test `TestListenProvidersPublishWhenTLSCertChange` under server/server_tests.go that loads a conf that includes `TLS`, then provides the same conf 2 times and expect no changes and finally rewrite new certs and provide the same conf that should trigger a reload.

### DIFF
--- a/provider/file/file.go
+++ b/provider/file/file.go
@@ -175,17 +175,36 @@ func (p *Provider) loadFileConfig(filename string, parseTemplate bool) (*types.C
 	} else {
 		configuration, err = p.DecodeConfiguration(fileContent)
 	}
-
 	if err != nil {
 		return nil, err
 	}
+
+	var tlsConfigs []*tls.Configuration
+	for _, conf := range configuration.TLS {
+		bytes, err := conf.Certificate.CertFile.Read()
+		if err != nil {
+			log.Error(err)
+			continue
+		}
+		conf.Certificate.CertFile = tls.FileOrContent(string(bytes))
+
+		bytes, err = conf.Certificate.KeyFile.Read()
+		if err != nil {
+			log.Error(err)
+			continue
+		}
+		conf.Certificate.KeyFile = tls.FileOrContent(string(bytes))
+		tlsConfigs = append(tlsConfigs, conf)
+	}
+	configuration.TLS = tlsConfigs
+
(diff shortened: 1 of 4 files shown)


ffilippopoulos:
semaphoreci looks like is having issues with docker:
```
docker build  "--build-arg=DOCKER_VERSION=18.06.1" -t "traefik-dev:master" -f build.Dockerfile .
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

dbachrach:
We are running into a similar issue where cert-manager updates the secrets but the Traefik pods need to be explicitly killed. I had a question on this part though:

> The aim is to solve issue: #3272 and make reload configuration because of cert changes when touch toml files.

Should you have to touch the toml file at all? Should a modification to the TLS files be all that's needed? If not, what aspect in the cert-manager update process will touch the toml file to trigger a reload in traefik?

ffilippopoulos:
@dbachrach that is proposed/discussed here: https://github.com/containous/traefik/issues/3083 and is more of an enhancement. If you read through the issue there is a small explanation on why this mechanism (touching the toml file) was picked. This PR is just trying to solve the bug that is linked in the description.

Arguably this is still not going to be the ideal approach in our (and your) case as you still have to run a sidecar to watch the certs (i am using `inotify` for that) and `touch` the toml file in case of change.

ffilippopoulos:
Is there anything we can do to help this one progress? We are happy to review code if there are any concerns or anything similar. It is just holding us to migrate from tls termination on elbs as we would have to go and manually restart traefiks every couple of months which can be disturbing.

juliens:
Thank you for your PR.

I think that this is better to replace file path with the content directly in the provider file. With this, it will fix the bug because the `loadConfiguration` will see the content modification.

I did the modification in order to merge it in the next bug fix release.

PR #4083, opened 2018-10-19
Add missing tmp directory to scratch image

### LINKED ISSUE #4064: Buffering enabled on k8s service throws errors about /tmp not available in non-alpine docker image

Bug

### What did you do?

Enabled buffering via traefik.ingress.kubernetes.io/buffering annotation on kubernetes services

```yaml
    traefik.ingress.kubernetes.io/buffering: |
      maxrequestbodybytes: 10485760
      memrequestbodybytes: 2097153
      maxresponsebodybytes: 10485761
      memresponsebodybytes: 2097152
      retryExpression: IsNetworkError() && Attempts() <= 2
```

Running traefik:1.7.3 due to http handler panic in 1.7.2 with same configuration.

Browse to site (GET), get served the content normally available on the URI as long as request body is not too large.

### What did you see instead?

No content, just a HTTP 200 instead of expected content, evne with no request body on a GET request.

In traefik's log, this message is spammed en masse:
```
time="2018-10-17T07:51:09Z" level=error msg="open /tmp/temp-multibuf-691319615: no such file or directory"
```
It would appear /tmp is not writable in the non-alpine container. Switching to 1.7.3-alpine allows the configuration to work, and content is received as expected.

```
(paste your output here)
```

Same environment/configuration as in #4005, but with 1.7.3 versions of docker hub image.

### If applicable, please paste the log output in DEBUG level (`--logLevel=DEBUG` switch)

```
(paste your output here)
```

dtomcej:
### What does this PR do?

Fixes buffering by re-adding the missing `/tmp` directory.

### Motivation

Closes #4064

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM scratch
 COPY script/ca-certificates.crt /etc/ssl/certs/
 COPY dist/traefik /
+RUN mdir -p /tmp
 EXPOSE 80
 ENTRYPOINT ["/traefik"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM scratch
 COPY script/ca-certificates.crt /etc/ssl/certs/
 COPY dist/traefik /
-RUN mdir -p /tmp
+ WORKDIR /tmp
+ WORKDIR /
 EXPOSE 80
 ENTRYPOINT ["/traefik"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM scratch
+WORKDIR /tmp
+WORKDIR /
 COPY script/ca-certificates.crt /etc/ssl/certs/
 COPY dist/traefik /
- WORKDIR /tmp
- WORKDIR /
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM scratch
 COPY script/ca-certificates.crt /etc/ssl/certs/
 COPY dist/traefik /
+RUN mdir -p /tmp
```
mmatur: `mdir` is not a valid command.

```suggestion
 WORKDIR /tmp
 WORKDIR /
```

dtomcej: lol that was supposed to be `mkdir`

Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM scratch
 COPY script/ca-certificates.crt /etc/ssl/certs/
 COPY dist/traefik /
+ WORKDIR /tmp
```
mmatur: These two `WORKDIR` should be before the first `COPY` we will be able to use cache ;)

dduportal:
LGTM 👍

PR #4112, opened 2018-10-25
Add flush interval option on backend

### LINKED ISSUE #4007: GRPC bidirectional streaming:  small messages (< 4kB) are buffered/delayed

Bug

### What did you do?

Using Traefik (as a replacement for Envoy) for grpc bidirectional streaming API when there are client and server sending very small protobuf messages (e.g. few bytes) and "normal size" messages ( > 4kB). 

Both client and server should receive both message sizes without delay.

### What did you see instead?

Larger messages are received immediately, small messages got buffered for approx. 100ms and then send. 

```
Version:      v1.7.2
Codename:     maroilles
Go version:   go1.11.1
Built:        2018-10-04_01:44:36PM
OS/Arch:      linux/amd64
```

```
Docker 18.06, Ubuntu 18.04
```

We have the same API implemented using Websocket (no problems here both with Traefik and Envoy) and also have the same API running behind Envoy without problems. Does not matter if using h2c or https with self signed cert + insecureSkipVerify. From what I tested it works as long as the message has more than 4096 bytes. Any smaller message (3kB ...) gets buffered and send with delay (consistently around 100ms - avg. RTT is 20ms in my setup). Ignoring Traefik and doing direct communication (grpc client - server) does not have this issue.

juliens:
### What does this PR do?
Add an option to specify the response forwarding flush interval.

### Motivation
Fixes #4007

### More

- [x] Added/updated tests

### Additional Notes
Before this pull request it was not possible to configure the flush interval and it was set to 100ms.


docs/configuration/backends/file.md
```diff
@@ -23,6 +23,9 @@ Traefik can be configured with a file.
 
     [backends.backend1.circuitBreaker]
       expression = "NetworkErrorRatio() > 0.5"
+      
+    [backends.backend1.responseforwarding]
+      flushinterval = "10ms"
```
mmatur:
```suggestion
      flushInterval = "10ms"
```

docs/configuration/backends/file.md
```diff
@@ -23,6 +23,9 @@ Traefik can be configured with a file.
 
     [backends.backend1.circuitBreaker]
       expression = "NetworkErrorRatio() > 0.5"
+      
+    [backends.backend1.responseforwarding]
```
mmatur:
```suggestion
    [backends.backend1.responseForwarding]
```

integration/fixtures/grpc/config_with_flush.toml
```diff
@@ -0,0 +1,31 @@
...
+     keyFile  = """{{ .KeyContent }}"""
+
+
+[api]
+
+[file]
+
+[backends]
+  [backends.backend1]
+  [backends.backend1.responseforwarding]
```
mmatur:
```suggestion
  [backends.backend1.responseForwarding]
```

Issue #4040, opened 2018-10-14
Support for custom frontend redirect codes


danielrs:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Currently, frontend redirects only support the `permanent` option which maps to either 301 or 302. However, I think we need to be able to specify which code to use for some cases e.g. POSTing to a redirected URL only works with 307 redirect.

dm17:
Now: how to use them? I'm trying to do a 307 redirect.

ldez:
Fixed by #4116

PR #4116, opened 2018-10-26
Redirection status codes for methods different than GET

r--w:
For HTTP methods different than GET on redirection status codes should be:
 * 307 temporary (https://tools.ietf.org/html/rfc7231#section-6.4.7)
 * 308 permanent (https://tools.ietf.org/html/rfc7538)
in order to keep requested HTTP method.

#4040 #hacktoberfest #greywizard

### DIFF
--- a/docs/configuration/ping.md
+++ b/docs/configuration/ping.md
@@ -27,7 +27,7 @@ The `/ping` health-check URL is enabled with the command-line `--ping` or config
 Thus, if you have a regular path for `/foo` and an entrypoint on `:80`, you would access them as follows:
 
 * Regular path: `http://hostname:80/foo`
-* Dashboard: `http://hostname:8080/`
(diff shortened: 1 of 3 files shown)


docs/configuration/ping.md
```diff
@@ -27,7 +27,7 @@ The `/ping` health-check URL is enabled with the command-line `--ping` or config
 Thus, if you have a regular path for `/foo` and an entrypoint on `:80`, you would access them as follows:
 
 * Regular path: `http://hostname:80/foo`
-* Admin panel: `http://hostname:8080/`
```
dduportal: Could you remove this change from the PR, since it is out the scope.

r--w: Thank you for the review. Commit has been reverted.

PR #4121, opened 2018-10-27
Fix mistake in the documentation of several backends

whalehub:
### What does this PR do?

This PR changes the documentation for the following three labels to say that they add a HPKP header instead of a pinned HTST public key header.

`<prefix>.frontend.headers.publicKey=VALUE` (Consul Catalog)
`traefik.frontend.headers.publicKey=VALUE` (Azure Service Fabric, Docker, ECS, Marathon, Mesos, Rancher)
`ingress.kubernetes.io/public-key: VALUE` (Kubernetes Ingress)

### Motivation

The documentation for these labels currently says that they add a pinned HTST public key header. The problem with this is that there is no such thing as a pinned HTST (?) or even HSTS (HTTP Strict Transport Security) public key header. What these labels actually do is add a HPKP (HTTP Public Key Pinning) header.

### More

- [ ] Added/updated tests

### Additional Notes

N/A


mmatur:
Hi @whalehub, 

Could you please rebase your PR on v1.7 branch.

whalehub:
Hi @mmatur 

I opened a new PR based on the v1.7 branch: #4133

Issue #4032, opened 2018-10-12
Add provider specific cnameFlattening Trigger


gamalan:
### Do you want to request a *feature* or report a *bug*?

Feature (as per @nmengin suggestion, moved from PR #3843)

### What did you expect to see?

Add provider specific cnameFlattening Trigger. 
```
traefik.frontend.cnameFlattening
traefik.ingress.kubernetes.io/cname-flattening
```

To enable cnameFlattening only for a specific provider. But keep global if needed. Theoretically, it should only be enabled for a specific service.

This addressing PR #3403, by default when cnameFlattening is enabled, it will check for all route. After some thinking/testing again, although the record is being cached, it should not necessarily check all service. 

If approved, then traefik-extra-service-fabric need to be updated as well.

PR #4134, opened 2018-10-30
Add provider specific cnameFlattening Trigger

gamalan:
### What does this PR do?

Add provider specific cnameFlattening Trigger. 
```
traefik.frontend.cnameFlattening
traefik.ingress.kubernetes.io/cname-flattening
```

### Motivation

To enable cnameFlattening only for a specific provider. But keep global if needed. Theoretically, it should only be enabled for a specific service.

This addressing PR #3403, by default when cnameFlattening is enabled, it will check for all route. After some thinking again, although the record is being cached, it should not necessarily check all service. 

### More

- [x] Added/updated tests

### Additional Notes

If approved, then traefik-extra-service-fabric need to be updated as well.
Re add from PR #3843


geraldcroes:
Hi @gamalan!

Thank you for your contribution to the project, we're very grateful ... but at the same time we're embarrassed because we're not ready to merge your work. As we said in the [latests release note](https://blog.containo.us/traefik-1-7-yet-another-slice-of-awesomeness-2a9c99737889) , we're re-thinking some core aspects of Traefik. As a consequence, the code base has drastically changed (you can see for yourself the #4168 PR ... quite a refactoring).

Even though we cannot accept the PR right now, we would like to keep it in our radar.

I propose we close the current PR but keep #4032 open to ping you when the time comes.

WDYT?

gamalan:
@geraldcroes ah, I see. Okay then, I probably check other issues to solve.

Issue #3974, opened 2018-10-03
Traefik (1.6.5 and 1.7.1) stops watching file after it detects Marathon is down


daltonmatos:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I'm trying to use the `file.watch` feature. The idea is to be able to reload SSL certificates without restaring traefik.

I have the TLS config on a separated file, and here is my main config file (the part that matters):

```
[file]
watch = true
fileName = "/etc/ssl.toml"
```

I have two copies of "/etc/ssl.toml", each one with a different certificate, like this:

`ssl1.toml`:

```
[[tls]]
  entryPoints = ["https"]
  [tls.certificate]
      certFile = "/etc/cert1/fullchain.pem"
      keyFile = "/etc/cert1/privkey.pem"

```

and `ssl2.toml`:
```
[[tls]]
  entryPoints = ["https"]
  [tls.certificate]
      certFile = "/etc/cert2/fullchain.pem"
      keyFile = "/etc/cert2/privkey.pem"
```

And then what I do is just a simple `cp`, like this:

```
cd /etc
cp ssl2.toml ssl.toml
cp ssl1.toml ssl.toml
```

And check the certificates were reloaded with a simple curl to localhost.

### What did you expect to see?

I expected to see traefik change the certificates **every time** I did the `cp`.

### What did you see instead?

Traefik changes the certificates correctly, until this message appears in the debug log:

```
ERRO[2018-10-03T09:45:23-03:00] Failed to retrieve Marathon applications: all the Marathon hosts are presently down
```

As soon as this message appears, traefik **does not** detect the file change anymore and then the certificates stop being reloaded.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.6.5
Codename:     tetedemoine
Go version:   go1.10.3
Built:        2018-07-10_03:54:03PM
OS/Arch:      linux/amd64
```
The version here is `1.6.5` but this same behavior happens with `1.7.1`.

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```json
{
 "LifeCycle": {
  "RequestAcceptGraceTimeout": 0,
  "GraceTimeOut": 0
 },
 "GraceTimeOut": 0,
 "Debug": false,
 "CheckNewVersion": true,
 "SendAnonymousUsage": false,
 "AccessLogsFile": "",
 "AccessLog": null,
 "TraefikLogsFile": "",
 "TraefikLog": null,
[...]

PR #4146, opened 2018-11-01
Fix Marathon config load on connection error

diegooliveira:
### What does this PR do?

Add a logic to keeping trying to load the initial configuration if Marathon is inaccessible the first time.

### Motivation

There is an open issue, #3974, listing the problem.


ldez:
Could you use the branch 1.7 instead of 1.5?

diegooliveira:
@ldez yes, I'll make a pull request for that versions tonight. We are using the 1.5 version, so it would be nice to have the fix in all three versios, 1.5, 1.6 and 1.7.

diegooliveira:
@ldez are you going to release a new version of the 1.5 branch? We are using this in production and this  is a current problem in this version.

diegooliveira:
@ldez I just submited two PRs #4150 for v1.6 and #4151 for v1.7

ldez:
We create release only for version 1.7 according to:

> Each version is supported until the next one is released (e.g. 1.1.x will be supported until 1.2.0 is out)

https://github.com/containous/traefik#release-cycle

PR #4170, opened 2018-11-07
Include an explicit list of kubernetes protocol annotations in docs.

shanna:
### What does this PR do?

Includes an explicit list of valid protocols in kubernetes backend documentation.

### Motivation

The idea of using port 443 or the special ingress spec port name seemed like something that will trip up other developers on my team. I saw the `ingress.kubernetes.io/protocol` annotation in the documentation but couldn't find the valid values. Both `http2` and `h2` as someone suggested on slack both produced errors in the log so I was forced to search the source code.

### More

- [ ] Added/updated tests

### Additional Notes

Can the special port `443` and `https` prefix logic be deprecated and eventually removed now there is an explicit way to override the default protocol in the kubernetes back end?


docs/configuration/backends/kubernetes.md
```diff
@@ -166,7 +167,7 @@ The following general annotations are applicable on the Ingress object:
...
 | `traefik.ingress.kubernetes.io/whitelist-source-range: "1.2.3.0/24, fe80::/16"` | A comma-separated list of IP ranges permitted for access (7).                                                                                                                              |
 | `ingress.kubernetes.io/whitelist-x-forwarded-for: "true"`                       | Use `X-Forwarded-For` header as valid source of IP for the white list.                                                                                                                     |
-| `ingress.kubernetes.io/protocol: <NAME>`                                        | Set the protocol Traefik will use to communicate with pods.                                                                                                                                |
+| `ingress.kubernetes.io/protocol: http,https,h2c`                                | Set the protocol Traefik will use to communicate with pods.                                                                                                                                |
```
dtomcej:
```suggestion
| `ingress.kubernetes.io/protocol:<NAME>`                                | Set the protocol Traefik will use to communicate with pods. Acceptable protocols: http,https,h2c                                                                                                                        |
```

dtomcej: We should put options and such in the description to avoid appearing that the annotation can accept a list 😄

PR #4177, opened 2018-11-08
Cache exising task definitions to avoid rate limiting

hwhelan-CB:
### What does this PR do?

Stores existing task definitions to avoid rate limiting of the AWS describeTask API.  If there is an existing definition, use instead of making another API call.

### Motivation

Deployments taking hours instead of minutes. Due to using multiple instances of Traefik, each making frequent desribeTask API calls, and quickly hitting AWS's rate limit.

### More

- [ ] Added/updated tests

### DIFF
--- a/provider/ecs/ecs.go
+++ b/provider/ecs/ecs.go
@@ -23,6 +23,7 @@ import (
 
 var _ provider.Provider = (*Provider)(nil)
 var existingTaskDef = make(map[string]*ecs.TaskDefinition)
+
 // Provider holds configurations of the provider.
 type Provider struct {
 	provider.BaseProvider `mapstructure:",squash" export:"true"`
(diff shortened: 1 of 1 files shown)


blackey-mitek:
We have grown our usage of Traefik within AWS and it appears that we have hit this limitation as well.\ with several environments in an account each with several traefik instances we are hitting the API rate limiting constantly. The logic in the original implementation abuses the AWS describe task api at any reasonable scale. This request addresses much of it by reducing duplication. Might also want a delay/back off when receiving 429 for describe task.

provider/ecs/ecs.go
```diff
@@ -22,6 +22,7 @@ import (
 )
 
 var _ provider.Provider = (*Provider)(nil)
+var existingTaskDef = make(map[string]*ecs.TaskDefinition)
```
ldez: Could you use a cache that support expiration.

We already use github.com/patrickmn/go-cache inside Traefik.

emtammaru: I made this requested change in a separate PR. Please review, thanks!

Issue #4179, opened 2018-11-09
only one backend when loadbalancer.swarm=true


siyu6974:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

Running 2+ backends (Docker swarm 18.09) with the label traefik.backend.loadbalancer.swarm=true

### What did you expect to see?

Get 1 frontend with 2+ backends

### What did you see instead?

1 frontend with only 1 backend.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.7.4
Codename:     maroilles
Go version:   go1.11.1
Built:        2018-10-30_10
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
debug = false

logLevel = "ERROR"
defaultEntryPoints = ["https","http"]

[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    entryPoint = "https"
  [entryPoints.https]
  address = ":443"
  [entryPoints.https.tls]

[retry]

[docker]
endpoint = "tcp://127.0.0.1:2376"
domain = "my.domain"
watch = true
exposedByDefault = false
swarmMode = true

[acme]
email = "email@test.com"
storage = "acme.json"
entryPoint = "https"
onHostRule = true
[acme.httpChallenge]
entryPoint = "http"
```

### If applicable, please paste the log output in DEBUG level (`--logLevel=DEBUG` switch)
This is the exact same issue as #2138, and yes I shamelessly copied "What did you expect to see" and "What did you see instead", however using the latest version, the issue is still there.

Everything is working when traefik.backend.loadbalancer.swarm is set to false.
When loadbalancer.swarm=true, the LB is actually working, when I make requests to the endpoint, by inspecting the log, I can see backends taking their turn logging the response.

ldez:
Could you provide your docker-compose file?

Could you check that the labels are in the section `deploy`:
https://docs.traefik.io/master/configuration/backends/docker/#using-docker-with-swarm-mode

jlevesy:
Hey @siyu6974, thanks for opening this issue.

In order to achieve load balancing, docker swarm allocates a virtual IP per service, and this VIP is load balanced between all known backends.

With the `traefik.backend.loadbalancer.swarm=true`,  traefik will use this VIP instead of the containers IPs on the backend. From the traefik point of view, we're only seeing only one IP, so having one backend is the correct behaviour.

We're going to clarify this on the documentation, thanks again for reporting.

Issue #2138, opened 2017-09-18
Issue when traefik.backend.loadbalancer.swarm=true


0xbaadf00d:
### What did you do?

Running 2+ backends (Docker swarm 17.06) with the label `traefik.backend.loadbalancer.swarm=true`

### What did you expect to see?

Get 1 frontend with 2+ backends

### What did you see instead?

1 frontend with only 1 backend.

![image](https://user-images.githubusercontent.com/1005086/30556422-1a95c8a6-9cab-11e7-920c-c46deae84749.png)

PS : I have try with whoiam image and scale = 3, the LB don't working too.

### Output of `traefik version`:

```
Version: v1.4.0-rc2
Codename: roquefort
Go version: go1.9
Built: 2017-09-08_07:31:37PM
OS/Arch: linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
###############################################################################
## Entry points
# ~~~~~
defaultEntryPoints = ["http", "https"]
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    entryPoint = "https"
  [entryPoints.https]
  address = ":443"
  [[entryPoints.https.tls.certificates]]
    CertFile = "/etc/ssl/xxxx.io.crt"
    KeyFile = "/etc/ssl/xxxx.io.key"
  [[entryPoints.https.tls.certificates]]
    CertFile = "/etc/ssl/yyyy.ski.crt"
    KeyFile = "/etc/ssl/yyyy.ski.key"


[web]

# Web administration port
address = ":8080"

# Set REST API to read-only mode
ReadOnly = true

# Authentication
# https://websistent.com/tools/htdigest-generator-tool/
[web.auth.digest]
users = [
  "xxxx:traefik:xxxxx"
]


[docker]

# Docker server endpoint. Can be a tcp or a unix socket endpoint.
endpoint = "unix:///var/run/docker.sock"

# Default domain used.
# Can be overridden by setting the "traefik.domain" label on a container.
domain = "xxxx.io"

# Enable watch docker changes
watch = true

# Expose containers by default in traefik
# If set to false, containers that don't have `traefik.enable=true` will be ignored
exposedbydefault = false

# Use Swarm Mode services as data provider
swarmmode = true


[healthcheck]

# Set the default health check interval. Will only be effective if health check
# paths are defined. Given provider-specific support, the value may be
# overridden on a per-backend basis.
[...]

PR #4194, opened 2018-11-13
Clarify swarm loadbalancer documentation

jlevesy:
### What does this PR do?

Add a note which clarifies Traefik's behaviour when the `traefik.backend.loadbalancer.swarm=true` label is set on a backend.

### Motivation

Traefik's behaviour in that case caused some misunderstanding (see #4179  and #2138).
Here's a proposal to improve the documentation.

WDYT folks ?


docs/configuration/backends/docker.md
```diff
@@ -292,11 +292,16 @@ If a container is linked to several networks, be sure to set the proper network
 For instance when deploying docker `stack` from compose files, the compose defined networks will be prefixed with the `stack` name.
 Or if your service references external network use it's name instead.
 
-[2] `traefik.frontend.auth.basic.users=EXPR `:
```
ldez: could you restore the 2 spaces at the end of line.

jlevesy: Oops, sorry. I thought it was whitespaces.

kachkaev:
Thank you for adding this note @jlevesy! It finally explains why I could not configure sticky sessions for one of my projects that uses docker swarm :-D

PR #4132, opened 2018-10-29
Implement Case-insensitive SNI matching

dtomcej:
### What does this PR do?

Implements case-insensitive SNI matching for certificate selection

### Motivation

If your certificates did not match with case sensitivity, the default certificate would be served (which may be the incorrect one).

Since the `types.CanonicalDomain(clientHello.ServerName)` always returns lowercase, we should have all our certs keyed in lowercase.

Addresses issues in #4081 

### More

- [x] Added/updated tests
- [x] Added/updated documentation - None needed, Bugfix

PR #4204, opened 2018-11-16
Lower casing the certificate domain name

vineetvermait:
Update for lower casing the certificate domain name in case the same is not set to lower case.

Example Scenario:

certificate generated for LocalNet.com
domain to be checked LocalNet.com

while checking the domain to be checked is lower cased, but the certificate domain name is not set to lower case.

updated the code for the same

### More

- [ ] Added/updated tests


ldez:
duplicate of #4132

PR #4168, opened 2018-11-07
Dynamic Configuration Refactoring

ldez:
**!!! Breaking changes !!!**

### What does this PR do

This PR introduces changes on the dynamic configuration.
- frontend becomes router
- backend becomes service
- new middleware approach (order customization, reusable configuration, chains)
- rules are now striclty matching the request without tweaking it (no more modifiers)

Tooling
- introduces a context based logger

### Motivation

A more flexible configuration file that gives additional possibilities to the end user
Logs that provide context to the end user

### Additional Notes

The only provider that has been migrated is `file`, as a proof of concept. **All other providers are broken**.

This Pull Request is the result of the work of many, and more specifically (in alphabetical order) @geraldcroes, @jbdoumenjou, @juliens, and @ldez

PR #4205, opened 2018-11-17
Update  for Documentation dead links and lower casing the certificate domain name

vineetvermait:
### What does this PR do?

Update for lower casing the certificate domain name in case the same is not set to lower case.

Example Scenario:

certificate generated for LocalNet.com
domain to be checked LocalNet.com

while checking the domain to be checked is lower cased, but the certificate domain name is not set to lower case.

updated the code for the same

### More

- [ ] Added/updated tests


ldez:
Its' already fixed in v1.7.

As we said in the [latests release note](https://blog.containo.us/traefik-1-7-yet-another-slice-of-awesomeness-2a9c99737889) , we're re-thinking some core aspects of Traefik. As a consequence, the code base has drastically changed in master (you can see for yourself the #4168 PR ... quite a refactoring).

For now I recommend to not use `master`.

mmatur:
Closed by #4206

PR #4208, opened 2018-11-19
Update swarm mode endpoint

siyu6974:
### What does this PR do?

According to docker's doc, https://docs.docker.com/engine/swarm/swarm-tutorial/#the-ip-address-of-the-manager-machine, port 2375 is replaced by 2377 for cluster management communications

### Motivation

Use port 2375 simply does not work with the new swarm the doc needs an update

### More

- [ ] Added/updated tests

### Additional Notes

`endpoint = "tcp://127.0.0.1:2377"` is only usable when launching traefik directly on the host. In a dockerized environment, `endpoint = "unix:///var/run/docker.sock"` should be used with a mounted volume `/var/run/docker.sock:/var/run/docker.sock`

### DIFF
--- a/docs/configuration/backends/docker.md
+++ b/docs/configuration/backends/docker.md
@@ -112,7 +112,7 @@ To enable constraints see [provider-specific constraints section](/configuration
 #
 # swarm classic (1.12-)
 # endpoint = "tcp://127.0.0.1:2375"
-# swarm intergated (1.12+)
+# docker swarm mode (1.12+)
 endpoint = "tcp://127.0.0.1:2377"
 
 # Default base domain used for the frontend rules.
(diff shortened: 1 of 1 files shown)


mmatur:
Closed by #4206

docs/configuration/backends/docker.md
```diff
@@ -110,7 +110,10 @@ To enable constraints see [provider-specific constraints section](/configuration
 # Required
 # Default: "unix:///var/run/docker.sock"
 #
-endpoint = "tcp://127.0.0.1:2375"
+# swarm classic (1.12-)
+# endpoint = "tcp://127.0.0.1:2375"
+# swarm intergated (1.12+)
```
nmengin: WDYT to replace `swarm integrated` by `swarm mode` which is the real name?

siyu6974: good idea!

PR #4224, opened 2018-11-23
Expose TLS setup in /api/tls

athoune:
### What does this PR do?

This PR add an entrypoint in the api for exposing tels configs.

### Motivation

It's hard to understand troubles with https. I want to see what is done, not what I asked.

### More

- [ ] Added/updated tests

### DIFF
--- a/api/handler.go
+++ b/api/handler.go
@@ -46,6 +46,16 @@ type ServiceRepresentation struct {
 	ID string `json:"id"`
 }
 
+type CertificateRepresentation struct {
+	CertFile string `json:"cert_file`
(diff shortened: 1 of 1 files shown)


athoune:
I'm fighting with tls configuration. I need some explicit certificates, some star certificates, and an ACME fallback. `[file]` works well for adding certificates.

But now, I'm fighting with some client side crash.
`curl: (35) error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error`

I'm seeking some visibility.

PR #4252, opened 2018-11-29
Remove Deprecated StorageFile

juliens:
### What does this PR do?
Remove Deprecated StorageFile.

### Motivation
No more deprecated.


djmaze:
Well, does this mean it is no longer possible to migrate ACME certificates into the KV store? That'd be a pity.

(Btw, the documentation still [says otherwise](https://docs.traefik.io/user-guide/kv-config/#store-configuration-in-key-value-store)).

Issue #4249, opened 2018-11-28
ECS service discovery recognizes non-RUNNING tasks as available


hwhelan-CB:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

During a deployment of an ECS service, all but one task started successfully.  The last task reached a PENDING state, then eventually STOPPED.  AWS continued to try to start the final desired task repeatedly (every 15 min after backoff) until it was eventually able to succeed.

### What did you expect to see?

Expected Traefik to only route requests to tasks that are in RUNNING state

### What did you see instead?

Traefik routed requests to tasks in the PENDING state, resulting in request timeouts.  As a result, every 15 minutes (AWS attempted to start the task every 15 min), we saw a high volume of requests to this service timeout.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
v1.7.3
```

mmatur:
Hi @hwhelan-CB,

Thanks for your interest in the project.

Has you can see in the code:
https://github.com/containous/traefik/blob/9d26839e54de26e0fd94a4cb9f5ed980b42e62a4/provider/ecs/ecs.go#L230

We fetch only Running task.

So could you give us more information.

hwhelan-CB:
Hi @mmatur This line filters tasks by the DesiredStatus, not the LastStatus.  Tasks can have a desired status of RUNNING, while the LastStatus is PENDING or STOPPED.  In this case, ECS is trying to bring the task to RUNNING, but that does not accurately reflect the current status of the task.  In our case, unhealthy ECS instances caused our new tasks (tasks that are assigned a DesiredStatus of RUNNING) to get stuck in the PENDING state, then eventually move to STOPPED.  I assume tasks usually reach RUNNING status very quickly so this is not an issue.

hwhelan-CB:
See PR

PR #4255, opened 2018-11-29
Filter ECS tasks by LastStatus before adding to list of service tasks

hwhelan-CB:
### What does this PR do?

Filters ECS tasks by their LastStatus, only adding them to the list of tasks if they are RUNNING

### Motivation

Fixes https://github.com/containous/traefik/issues/4249

### More

- [ ] Added/updated tests


ldez:
@hwhelan-CB could you use the branch v1.7 instead of master. Thanks.

PR #4262, opened 2018-12-02
Docs: Avoid ambivalent meaning of protocol in the docs docker.labels

EugenMayer:
### What does this PR do?

When reading the docs, one might not understand what 

`traefik.protocol=https | Overrides the default http protocol`

And assume, this means, what the frontend should attach to (http or https endpoint). In fact this defines the protocol which is used to talk to the backend.

Maybe stating it more explicit makes sense to avoid a possible confusion

### Motivation

Make the docs a little more clear for this particular  (important) switch.
"Set the protocol" could just be both, for the frontend or backend.

### More

- [ ] Added/updated tests


EugenMayer:
@dtomcej thanks for the review. I am kind of distracted/confused by the diff right now, 40 commits?
Did something go wrong, maybe i forked from the wrong branch? Or did something else change?

jbdoumenjou:
Hi @EugenMayer, Could you rebase?

ldez:
Since the PR has not been modified, and the remote deposit has disappeared, we are obliged to close this PR.

Do not hesitate to reopen one if necessary.

I hope you understand.

PR #4277, opened 2018-12-06
Refactor/update lego

### LINKED ISSUE #2879: Resolve domain names to Punycode before requesting a TLS certificate with acme

depends on perspective.

### What did you do?

i configured a Docker container with a label `traefik.frontend.rule=Host:<hostname>` where the hostname includes a non-ascii character.

traefik is configured to automatically request tls-certificates from Let's Encrypt:

```toml
[acme]
email = "…"
storage = "acme.json"
entryPoint = "https"
onHostRule = true

  [acme.httpChallenge]
  entryPoint = "http"
```

the error message is similar to the one reported in #2703, but also preceeded by this line:

    level=error msg="Near line 28 (last key parsed 'backends.backend-gitea.servers.server-gitea_gitea_1'): expected '.' or ']' to end table name, but got 'ä' instead" 

no error messages, but an automatic conversion of the hostname to Punycode as discussed in #2703.

ldez:
### What does this PR do?

- Support non-ascii domain name (punnycode)
- Check all challenges in a predictable order
- Disable pending authorizations
- use pointer for RemoteError/ProblemDetails
- Poll authz URL instead of challenge URL
- The ability for a DNS provider to solve the challenge sequentially
- Check all nameservers in a predictable order
- Logs every iteration of waiting for the propagation
- update DNSimple client
- update github.com/miekg/dns
- azure: Do not overwrite existing TXT records

### Motivation

Fixes bugs.

Fixes #2879

### More

- [x] Added/updated tests


EugenMayer:
- Seeing the adjustment on the Nameserver / DNS resolving part, does this change the way LEGO does resolve the SOA record for a domain to be able to "check" for the TXT record on the primary authoritative server?

- Maybe in addition, does this solve the issue that the order from /etc/resolve.conf was inverted or sometimes even random ( seems so from the description )?

Thanks

ldez:
Take a look to:
- https://github.com/xenolf/lego#usage
- https://github.com/xenolf/lego/pull/700
- https://github.com/containous/traefik/pull/4101
- https://docs.traefik.io/configuration/acme/#resolvers

The nameservers are now used in their initial order.

EugenMayer:
Thank you. Especially https://docs.traefik.io/v1.7/configuration/acme/#resolvers is extremly useful - i entirely missed that one. Guess that was 1.7.4 then

Very helpful in many ways, thank you

PR #4298, opened 2018-12-17
Add Pass TLS Cert Issuer and Domain Component

### LINKED ISSUE #4199: Properly uniquely identify TLS client certificate subjects and issuers

Feature

That Traefik would communicate all the information necessary for frontends to compare certificate Distinguished Names ("subjects") to those it has stored for access control.

Traefik does not presently support all the [subject fields required](https://tools.ietf.org/html/rfc3739#section-3.1.2) and [any issuer information whatsoever](https://tools.ietf.org/html/rfc3739#section-3.1.1) to [uniquely identify](https://tools.ietf.org/html/rfc3739#section-2.4) a certificate.

That is to say, you do not properly reconstruct certificate Distinguished Names and they are only unique for a given issuer. If you were to use a bundle to verify certificates (e.g., on Debian, `/etc/ssl/certs/ca-certificates.crt`) you do not provide the information necessary for the backend to double-check the issuer.

As a practical matter, I will have to pass the whole PEM and check its subject on the frontend. If I'm doing that, I might as well be verifying it myself.

Here's my Subject DN:
```
Subject: DC=org, DC=cilogon, C=US, O=LIGO, CN=Thomas Downes thomas.downes@ligo.org
```
Here's my Issuer DN:
```
Issuer: DC=org, DC=cilogon, C=US, O=CILogon, CN=CILogon Basic CA 1
```
In particular, Traefik doesn't know about `DC`, only `O`. The RFC must specify how to uniquely re-construct these DNs (UPDATE: well, it doesn't, but it does have specified rules about comparison which Traefik can't implement. Though, I believe OpenSSL has a "canonical" approach to constructing the DN) and you should really be offering that as an option. Something like:
```
traefik.frontend.passTLSClientCert.subject=true
traefik.frontend.passTLSClientCert.issuer=true
```
that would mimic this:
```
       Satisfy any
       SSLRequireSSL
       SSLVerifyClient require
       SSLVerifyDepth  10
       SSLOptions +ExportCertData +StrictRequire +LegacyDNStringFormat
       SSLOptions +StdEnvVars

       # Set SSL_* headers which are used by Django code
       RequestHeader set SSL_CLIENT_S_DN "%{SSL_CLIENT_S_DN}s"
       RequestHeader set SSL_CLIENT_I_DN "%{SSL_CLIENT_I_DN}s"
```
It's worth looking at the full range of [Apache support](https://httpd.apache.org/docs/current/mod/mod_ssl.html#envvars) for an understanding of what it does. You can see that they support both `SSL_CLIENT_S_DN` and `SSL_CLIENT_S_DN_$FIELD` headers.

jbdoumenjou:
### What does this PR do?

* Add the cert issuer data
* Add the domain component to the subject and issuer part

### Motivation

Fixes #4199 

### More

- [x] Added/updated tests


docs/configuration/commons.md
```diff
@@ -528,3 +528,40 @@ Example:
...
+                organization = true
+                commonName = true
+                serialNumber = true
+```
+
+Pass TLS Client Cert `pem` defines if the escaped pem is added to a `X-Forwarded-Ssl-Client-Cert` header.  
+Pass TLS Client Cert `infos` configuration defines how the certificate data are added to a `X-Forwarded-Ssl-Client-Cert-Infos` header.  
+
+The following example shows an unescaped result that uses all the available fields:
+If there is more than one certificates, their are separated by a `;`
```
geraldcroes:
```suggestion
If there are more than one certificate, they are separated by a `;`
```

docs/configuration/commons.md
```diff
@@ -528,3 +528,40 @@ Example:
...
+                domainComponent = true
+                province = true
+                locality = true
+                organization = true
+                commonName = true
+                serialNumber = true
+```
+
+Pass TLS Client Cert `pem` defines if the escaped pem is added to a `X-Forwarded-Ssl-Client-Cert` header.  
+Pass TLS Client Cert `infos` configuration defines how the certificate data are added to a `X-Forwarded-Ssl-Client-Cert-Infos` header.
```
geraldcroes:
```suggestion
Pass TLS Client Cert `infos` defines how the certificate data are added to a `X-Forwarded-Ssl-Client-Cert-Infos` header.  
```

PR #4302, opened 2018-12-17
Harden Traefik systemd service

jacksgt:
Since Traefik is running as root on the system, it makes sense to
apply various lock down measures to keep the system as safe as possible.

Includes mounting most of the directories as read-only or even making them
inaccessible, restricting kernel modifications and limiting the number of
processes the unit may spawn.

Also add checks at service startup to ensure all required files are present.

### DIFF
--- a/contrib/systemd/traefik.service
+++ b/contrib/systemd/traefik.service
@@ -1,11 +1,41 @@
 [Unit]
 Description=Traefik
+Documentation=https://docs.traefik.io
+After=network-online.target
+AssertFileIsExecutable=/usr/bin/traefik
+AssertPathExists=/etc/traefik/traefik.toml
 
 [Service]
+# Run traefik as its own user (create new user with: useradd -r -s /bin/false -U -M traefik)
+#User=traefik
+#AmbientCapabilities=CAP_NET_BIND_SERVICE
+
+# configure service behavior
 Type=notify
-ExecStart=/usr/bin/traefik --configFile=/etc/traefik.toml
+ExecStart=/usr/bin/traefik --configFile=/etc/traefik/traefik.toml
 Restart=always
 WatchdogSec=1s
+
+# lock down system access
+# prohibit any operating system and configuration modification
+ProtectSystem=strict
+# create separate, new (and empty) /tmp and /var/tmp filesystems
+PrivateTmp=true
+# make /home directories inaccessible
+ProtectHome=true
+# turns off access to physical devices (/dev/...)
+PrivateDevices=true
+# make kernel settings (procfs and sysfs) read-only
+ProtectKernelTunables=true
+# make cgroups /sys/fs/cgroup read-only
+ProtectControlGroups=true
+
+# allow writing of acme.json
+ReadWritePaths=/etc/traefik/acme.json
+# depending on log and entrypoint configuration, you may need to allow writing to other paths, too
+
+# limit number of processes in this unit
+LimitNPROC=1
 
 [Install]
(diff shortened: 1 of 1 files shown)


jacksgt:
No idea why the build failed, it definitely isn't related to my change.
```
Step 3/9 : RUN go get github.com/containous/go-bindata/... && go get golang.org/x/lint/golint && go get github.com/kisielk/errcheck && go get github.com/client9/misspell/cmd/misspell
 ---> Running in 17a424b3a78e
package github.com/containous/go-bindata/...: github.com/containous/go-bindata/...: invalid import path: malformed import path "github.com/containous/go-bindata/...": double dot
The command '/bin/sh -c go get github.com/containous/go-bindata/... && go get golang.org/x/lint/golint && go get github.com/kisielk/errcheck && go get github.com/client9/misspell/cmd/misspell' returned a non-zero code: 1
make: *** [build] Error 1
make validate failed, attempt 3/3
```
Can someone please trigger a re-build?

alemairebe:
it could also be run by a standard user with these 2 under [Service]
User=traefik
AmbientCapabilities=CAP_NET_BIND_SERVICE

jacksgt:
@alemairebe I thought about adding that too, but then again it requires setting up a user for traefik and assigning appropriate permissions to `/etc/traefik.toml` and `/etc/acme.json`.
Nothing complicated, but at least its not documented anywhere.

Maybe we can add that as a comment to the service file?
```
[Service]
# Run traefik as its own user (create new user with: useradd -r -s /bin/false -U -M traefik)
#User=traefik
#AmbientCapabilities=CAP_NET_BIND_SERVICE
ExecStart=...
```

jacksgt:
@mmatur Could you please review this PR?

contrib/systemd/traefik.service
```diff
@@ -1,11 +1,34 @@
...
+# lock down system access
+ProtectSystem=strict
+PrivateTmp=true
+ProtectHome=true
+PrivateDevices=true
+ProtectKernelTunables=true
+ProtectControlGroups=true
+
+# allow writing of acme.json
+ReadWritePaths=/etc/acme.json
```
mmatur: You allow only writing `acme.json` but it is not the only one file that traefik can write or read:
- [Logs](https://docs.traefik.io/configuration/logs/)
- [Entrypoints](https://docs.traefik.io/configuration/entrypoints/)

contrib/systemd/traefik.service
```diff
@@ -1,11 +1,34 @@
 [Unit]
 Description=Traefik
+Documentation=https://docs.traefik.io
+After=network-online.target
+AssertFileIsExecutable=/usr/bin/traefik
+AssertPathExists=/etc/traefik.toml
```
mmatur: Please refer to the documentation https://docs.traefik.io/basics/#configuration-file

`/etc/traefik.toml` is not the default path.

> By default, Traefik will try to find a traefik.toml in the following places:
> - /etc/traefik/
> - $HOME/.traefik/
> - `.` the working directory

contrib/systemd/traefik.service
jacksgt: The file is only supposed to serve as a general baseline, but I added a comment to the file about the log / entrypoint config.

contrib/systemd/traefik.service
jacksgt: Fixed.

jacksgt:
I added brief comment for all new parameters (for more in-depth information the systemd man page should be consulted).

Issue #4187, opened 2018-11-12
Adding wildcard ingress results in server config reloaded perpetually every 5 seconds


rosskevin:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

Added a wildcard ingress in `namespace: spinnaker` at `2018-11-12T16:48:19Z`.  Traefik is installed and running in `namespace: prod`.
 
`kubectl apply -f spinnaker-api.yml`
```yaml
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: traefik
    traefik.ingress.kubernetes.io/preserve-host: "true"
    traefik.ingress.kubernetes.io/priority: "100"
  labels:
    app: spin-gate
    ksonnet.io/component: spinnaker.api
  name: spin-gate
  namespace: spinnaker
spec:
  rules:
  - host: spinnaker-api.mydomain.com
    http:
      paths:
      - backend:
          serviceName: spin-gate
          servicePort: 8084
  tls:
  - hosts:
    - spinnaker-api.mydomain.com
    secretName: cert-wildcard-mydomain.com
```

### What did you expect to see?
A working ingress, calm logs, and a traefik that responds timely to service/ingress changes.

### What did you see instead?
Repeated logging every 5 seconds and eventually traefik stops responding to service/config changes (but continues serving).  In all cases, the traefik dashboard and whatever ingress I add works (even if I have to delete traefik to make it see the new service).

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
1.7.4
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

- GKE v1.11.2-gke.15
- traefik in the `prod` namespace
- cert-manager for acme wildcard certificates
- spinnaker-api ingress in the `spinnaker` namespace 

```toml

logLevel = "INFO"
sendAnonymousUsage = true
defaultEntryPoints = ["http","https"]
[entryPoints.http]
    address = ":8080"
    compress = true

[entryPoints.http.redirect]
    regex = "^http://(.*)"
    replacement = "https://$1"
    permanent = true

[entryPoints.https]
    address = ":8443"
    compress = true

[entryPoints.https.tls]
    minVersion = "VersionTLS11"
    cipherSuites = [
      "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
      "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
[...]

PR #4305, opened 2018-12-17
Fix for #4245

### LINKED ISSUE #4245: Traefik killed by OOM killer in Kubernetes

Bug

### What did you do?

- Running Traefik as Deployment (ingress controller) in K8s
- Having bunch of Services targeting no Pods (thus having no related Endpoints object)
- Our setup is specific in way that we downscale tens of deployments to 0 on some schedule to not waste resources of K8s cluster.
 
- Traefik to work and reply with error when tried to access such a service

### What did you see instead?

- Traefik leaking memory and eventually being killed by OOM killer

1.7.4

```toml
    [entryPoints]
      [entryPoints.http]
      address = ":80"
      compress = true
        [entryPoints.http.proxyProtocol]
        trustedIPs = ["xxxx"]
      [entryPoints.https]
      address = ":443"
      compress = true
        [entryPoints.https.proxyProtocol]
        trustedIPs = ["xxxx"]
        [entryPoints.https.tls]
        minVersion = "VersionTLS11"
        cipherSuites = [
          "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
          "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305",
          "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA",
          "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA",
          "TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA",
          "TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA"
        ]
    [respondingTimeouts]
      idleTimeout = "900s"
    [lifeCycle]
      requestAcceptGraceTimeout = "10s"
      graceTimeOut = "10s"
    [api]
    [rest]
    [kubernetes]
    [kubernetes.ingressEndpoint]
      hostname = "xxxx.xxxx"
    [metrics]
      [metrics.prometheus]
    [accessLog]
```

### Logs
```
time="2018-11-28T11:44:33Z" level=warning msg="Endpoints not available for xxx"
time="2018-11-28T11:44:33Z" level=warning msg="Endpoints not available for xxx"
time="2018-11-28T11:44:33Z" level=warning msg="Endpoints not available for xxx"
```
Tons of those ^

I suspect it being a memory leak in watcher watching the Services/Endpoints as this memory growth is specific just in this case, on our other clusters where we don't downscale Deployments like this Traefik works just fine.

coufalja:
### What does this PR do?

Fixes #4245

### Motivation

In our Kubernetes cluster our Traefik pods were frequently killed by OOM killer.

### Additional Notes

Issue with leaking memory was actually not about leaking memory, but rather about GC not being able to keep up with the amount of garbage created. That was because DeepEquals of Ingresses loaded was almost always returning false due to Ingresses loaded in unspecified order. The problem was caused by TLS field as our services has wildcard certificate defined and if there are multiple Ingresses with the same TLS section (same wildcard certificate in the same secret) the latter ones are skipped. Unspecified order of processing leads to different results from the struct POV while functionally the same but returning false for mentioned DeepEquals function.


provider/kubernetes/client.go
```diff
@@ -18,7 +19,10 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
-const resyncPeriod = 10 * time.Minute
+const (
+	resyncPeriod   = 10 * time.Minute
+	ingressPattern = "%s/%s"
```
```suggestion
```

provider/kubernetes/client.go
```diff
@@ -165,6 +169,13 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 			result = append(result, ing)
 		}
 	}
+
+	sort.Slice(result, func(i, j int) bool {
+		i1 := result[i]
+		i2 := result[j]
+		return fmt.Sprintf(ingressPattern, i1.Namespace, i1.Name) < fmt.Sprintf(ingressPattern, i2.Namespace, i2.Name)
```
```suggestion
		return result[i].Namespace < result[j].Namespace ||
			result[i].Namespace == result[j].Namespace && result[i].Name < result[j].Name
```

provider/kubernetes/client.go
```diff
@@ -165,6 +169,13 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
 			result = append(result, ing)
 		}
 	}
+
+	sort.Slice(result, func(i, j int) bool {
+		i1 := result[i]
```
```suggestion
```

provider/kubernetes/client.go
```diff
@@ -165,6 +169,13 @@ func (c *clientImpl) GetIngresses() []*extensionsv1beta1.Ingress {
...
+		i2 := result[j]
```
```suggestion
```

coufalja: True but the maps are fine it is only the TLS which is a slice (which has an order) that is offending.

coufalja:
True but the maps are fine it is only the TLS which is a slice (which has an order) that is offending.

ReillyProcentive:
I have a hunch that this might also fix #4187

zarqman:
Possibly related, I just submitted PR #4307 which de-dups the TLS section of the config which has the possibility of significantly reducing memory if there are a large number of Ingresses referencing the same set of TLS Secrets. That PR doesn't affect the list of frontends or backends at all though, if those data structures are still part of the concern.

coufalja:
@ldez Today I have tested #4307 and it does not fix #4245 the configuration is still reloaded each time in our case and still running out of memory. I am trying to debug why.

ldez:
@coufalja  I fixed this problem :wink:

PR #4313, opened 2018-12-18
Exclude websocket connections from Average Response Time

### LINKED ISSUE #3620: Exclude websocket connections from Average Response Time (µs)
I'm running some websockets thru Traefik, this causes the Average Response Time (µs) to spike to huge values.

This would be solved if there was some way to exclude websocket connections.

siyu6974:
### What does this PR do?

Update dependency to fix #3620

### Motivation

The average response time on my dashboard was really bad because of it counting websocket connect.

### More

not relevant 
~~- [ ] Added/updated tests~~
~~- [ ] Added/updated documentation~~

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -371,7 +371,6 @@
   version = "v3.2.0"
 
 [[projects]]
-  branch = "master"
   name = "github.com/dimchansky/utfbom"
   packages = ["."]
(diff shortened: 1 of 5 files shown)


ldez:
@siyu6974 could you don't update your branch.

siyu6974:
Ok!

Issue #4284, opened 2018-12-10
Backend naming not being set as specified


amitsaha:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

Consider the following `docker-compose` file:

```
version: '3'

services:
  reverse-proxy:
    image: traefik # The official Traefik docker image
    command: --api --docker # Enables the web UI and tells Traefik to listen to docker
    ports:
      - "80:80"     # The HTTP port
      - "8080:8080" # The Web UI (enabled by --api)
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock # So that Traefik can listen to the Docker events
  whoami:
    image: containous/whoami # A container that exposes an API to show its IP address
    labels:
      - "traefik.frontend.rule=Host:whoami.docker.localhost"
      - "traefik.backend=backend1"

  whoami-again:
    image: containous/whoami # A container that exposes an API to show its IP address
    labels:
      - "traefik.frontend.rule=Host:whoami.docker.localhost"
      - "traefik.site1.backend=backendsite1"
```

Run `docker-compose up`

Go to `http://localhost:8080/api/providers/docker/backends`, we will get the three backends that's configured:

```
{
  "backend-backend1": {
    "servers": {
      "server-traefik-demo-whoami-1-86460ec963c2-f0078ecb386e282a8fc546f06636ff94": {
        "url": "http:\/\/172.18.0.3:80",
        "weight": 1
      }
    },
    "loadBalancer": {
      "method": "wrr"
    }
  },
  "backend-reverse-proxy-traefik-demo": {
    "servers": {
      "server-traefik-demo-reverse-proxy-1-807284c2bf53-a4b9e1129a86189ee88fc1a031f0c65d": {
        "url": "http:\/\/172.18.0.4:80",
        "weight": 1
      }
    },
    "loadBalancer": {
      "method": "wrr"
    }
  },
  "backend-whoami-again-traefik-demo-backendsite1": {
    "servers": {
      "server-traefik-demo-whoami-again-1-36490b790acf-780f230448df16d66397c0c29cebc062": {
        "url": "http:\/\/172.18.0.2:80",
        "weight": 1
      }
    },
    "loadBalancer": {
      "method": "wrr"
    }
  }
}
```

### What did you expect to see?

I expected to see the backend for `whoami` service to be `backend1` and the backend for `whoami-again` to be `backendsite1`.

### What did you see instead?

[...]

Issue #4024, opened 2018-10-11
Question: How to remove backend prefix to docker containers?


JemiloII:
When I create the label:
`traefik.backend: aoh` in my docker-compose file, traefik adds `backend_aoh` instead of `aoh`, how can I remove that? If I manually set the rules in the toml file, I can get the backend to be named `aoh` by using `[backends.aoh]`

dduportal:
Hello @JemiloII , thanks for your interest in the project!

This naming is internal, and depends on dynamic generation from Docker.
You don't need to change this since it is not related to any configuration.

If you have any question, do not hesitate to use the slack channel: https://slack.traefik.io/ .

PR #4317, opened 2018-12-19
Rephrase the `traefik.backend` definition in documentation

dduportal:
Signed-off-by: Damien DUPORTAL <damien.duportal@gmail.com>

### What does this PR do?

This PR rephrases the definition of the label `traefik.backend` in the documentation.

### Motivation

As underlined in #4284 and #4024, the actual definition of the label `traefik.backend`
leads to confusion: the provided value is expected to be the full backend name.
But it is an internal naming, which depends on dynamic name generation,
and the provided value is used as a suffix, to discriminate backends from each others.

So here is this PR, trying to rephrase to make it more close to reality, and to avoid false expectations when using the API.

### More

- [ ] Added/updated tests

### Additional Notes

Here is a nice gopher image, because gophers are cute:

image


dduportal:
@ldez : I've taken your feedback and explanations in account.

The code from `providers/*/config.go` and `templates/*.tmpl` had been used for the 5 providers, to get the right "object" name:

- Docker: `container`
- ECS and Rancher: `service`
- Mesos: `task`
- Marathon: `application`

docs/configuration/backends/docker.md
```diff
@@ -269,80 +269,80 @@ services:
...
+| `traefik.protocol=https`                                             | Overrides the default `http` protocol                                                                                                                                                                                            |
+| `traefik.weight=10`                                                  | Assigns this weight to the container                                                                                                                                                                                             |
+| `traefik.backend=foo`                                                | Overrides the container name by `foo` in the generated name of the backend.                                                                                                                                                      |
```
amitsaha: This is accurate for backends which are not segments. However, when we use segments, as explained in https://github.com/containous/traefik/issues/4284, the backend name is being set as `backend-<container-name>-backendsite1` where `traefik.backend=backendsite1` has been specified by
the user.

PR #4318, opened 2018-12-19
Implement SkipVerify for entrypoint ClientCA

ti-mo:
### What does this PR do?

Added an option to skip client certificate verification on a TLS entrypoint.

### Motivation

When passing through client certificates over HTTP headers, it is assumed that some form of client certificate validation happens on the back-end. In some organizations, these back-ends might belong to multiple groups/teams that manage their own PKI infrastructure, and don't necessarily have permission to manage trusted roots on the load balancer.

This patch allows the administrator to skip verification of the client certificate and fully delegate that task to the back-end. Installing a 'dummy' root certificate is still necessary with this patch, something I would like to be able to omit. For example, only specifying the following configs:

- `entryPoints.https.tls.ClientCA.optional = true`
- `entryPoints.https.tls.ClientCA.skipVerify = true`

In my ideal scenario, this would pass an empty cert pool, request the client to provide a certificate, skip verification, and pass the urlized PEM to the designated back-end. Not having this is fine, however.

### More

- [ ] Added/updated tests

### Additional Notes

Before committing to writing tests, I would like some input about whether or not this is an acceptable solution, and whether or not everything in this change set is necessary. I've touched both `struct types.ClientTLS` and `struct tls.ClientCA`, and I'm not sure yet which one is redundant. I have a feeling `types` is used for the Auth back-end, and my change there doesn't make sense.

Also, looking at master, I see all of this code changed significantly, but I'd be interested in contributing the same feature in 2.0 as well.

### DIFF
--- a/configuration/entrypoints.go
+++ b/configuration/entrypoints.go
@@ -237,9 +237,11 @@ func makeEntryPointTLS(result map[string]string) (*tls.TLS, error) {
 			files := tls.FilesOrContents{}
 			files.Set(result["ca"])
 			optional := toBool(result, "ca_optional")
+			skipVerify := toBool(result, "ca_skipverify")
 			configTLS.ClientCA = tls.ClientCA{
-				Files:    files,
-				Optional: optional,
+				Files:      files,
+				Optional:   optional,
+				SkipVerify: skipVerify,
 			}
 		}
 
--- a/server/server.go
+++ b/server/server.go
@@ -17,7 +17,7 @@ import (
 	"sync"
 	"time"
 
-	"github.com/armon/go-proxyproto"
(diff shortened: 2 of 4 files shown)


emilevauge:
Hi @ti-mo 
Thanks a lot for this great contribution :)
We have one concern though, we will not be able to merge this one in 1.7 branch, as it's a new feature.
May I suggest you to rebase your work on master branch?
⚠️we are in the process of refactoring the master branch, things might have changed a lot/might change in the future. Do no hesitate to contact us on Slack (#dev channel) if you need some help.

ti-mo:
@emilevauge Great, gladly! Will drop by Slack for the questions I have, as they don't really belong here.

PR #4340, opened 2018-12-27
Skip TLS section with no secret in Kubernetes ingress

### LINKED ISSUE #4280: [Kubernetes] If a "tls" section is specified in an Ingress, but a secret is not specified, Traefik will fail to load the Ingress

Bug

### What did you do?

We deployed Traefik with a "default certificate" specified, so that we didn't have to specify a TLS cert on every Ingress resource.  (We have a wildcard TLS cert so we'd prefer to just specify in one place - at the Traefik level.)  Our Ingress `spec` sections look like the following:

```yaml
spec:
  rules:
  - host: foo.bar.com
    http:
      paths:
        <paths...>
  tls:
  - hosts:
    - foo.bar.com
```
^ Notice the "tls" section, which simply has the host listed but not any secret info. This is because we are relying on the default certificate.

Traefik would load the Ingress rule, and would present its default certificate since a different one wasn't specified in the Ingress.  For reference, this works as expected with the Nginx Ingress controller.

### What did you see instead?

Traefik error logs:
```
{"level":"error","msg":"Error configuring TLS for ingress mynamespace/myservice: secret mynamespace/ does not exist","time":"2018-12-07T19:38:38Z"}
```
^Notice the empty secret name.

```
Version:      v1.7.5
Codename:     maroilles
Go version:   go1.11.2
Built:        2018-12-03_11:01:00AM
OS/Arch:      linux/amd64
```
(Note, also repros on v1.7.4)

```toml
# traefik.toml
logLevel = "INFO"
[entryPoints]
  [entryPoints.http]
  address = ":80"
  compress = true
    [entryPoints.http.redirect]
      regex = "^http://(.*)"
      replacement = "https://$1"
      permanent = true
  [entryPoints.https]
  address = ":443"
  compress = true
    [entryPoints.https.tls]
      minVersion = "VersionTLS12"
      cipherSuites = [
        "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256",
        "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384",
        "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256",
        "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384",
        "TLS_RSA_WITH_AES_128_GCM_SHA256",
        "TLS_RSA_WITH_AES_256_GCM_SHA384"
      ]
    [entryPoints.https.tls.defaultCertificate]
      CertFile = "/ssl/tls.crt"
      KeyFile = "/ssl/tls.key"
  [entryPoints.traefik]
  address = ":8080"
[kubernetes]
[traefikLog]
  format = "json"
[api]
  entryPoint = "traefik"
  dashboard = true
```

dtomcej:
### What does this PR do?

Checks to see if a secret is defined in the `TLS` block of an ingress before attempting to load the secret. Logs a debug message if TLS section is skipped.

### Motivation

Fixes #4280.

### More

- [x] Added/updated tests
- [x] Added/updated documentation - None needed, intended behavior

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -644,6 +644,11 @@ func getTLS(ingress *extensionsv1beta1.Ingress, k8sClient Client, tlsConfigs map
 	for _, t := range ingress.Spec.TLS {
 		newEntryPoints := getSliceStringValue(ingress.Annotations, annotationKubernetesFrontendEntryPoints)
 
+		if t.SecretName == "" {
+			log.Debugf("Skipping TLS sub-section for ingress %s/%s: No secret name provided", ingress.Namespace, ingress.Name)
+			continue
+		}
+
 		configKey := ingress.Namespace + "/" + t.SecretName
 		if tlsConfig, tlsExists := tlsConfigs[configKey]; tlsExists {
 			for _, entryPoint := range newEntryPoints {
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -642,12 +642,12 @@ func getRuleForHost(host string) string {
 
 func getTLS(ingress *extensionsv1beta1.Ingress, k8sClient Client, tlsConfigs map[string]*tls.Configuration) error {
 	for _, t := range ingress.Spec.TLS {
-		newEntryPoints := getSliceStringValue(ingress.Annotations, annotationKubernetesFrontendEntryPoints)
-
 		if t.SecretName == "" {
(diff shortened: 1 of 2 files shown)


provider/kubernetes/kubernetes.go
```diff
@@ -644,6 +644,11 @@ func getTLS(ingress *extensionsv1beta1.Ingress, k8sClient Client, tlsConfigs map
 	for _, t := range ingress.Spec.TLS {
 		newEntryPoints := getSliceStringValue(ingress.Annotations, annotationKubernetesFrontendEntryPoints)
 
+		if t.SecretName == "" {
```
ldez: you can move this to the start of the loop (L645) :wink:

PR #4347, opened 2018-12-30
Test and exit for jq error before domain loop

muhlemmer:
### What does this PR do?

In `contrib/scripts/dumpcerts.sh`: Test and exit for `jq` error before domain loop

### Motivation

In an automated (inotify) environment, there are cases where this script gets executed before Traefik actually finishes all writes to `acme.json`. This can result in `jq` errors when it's called in the the `domain` loop. The script did not check the `jq` return code and subsequently exited with return code `0`.

Original jq error:
```
jq: error (at /traefik/acme.json:18): Cannot iterate over null (null)
```

### More

Test and docs are not applicable in this case.

### Additional Notes

Nope

### DIFF
--- a/contrib/scripts/dumpcerts.sh
+++ b/contrib/scripts/dumpcerts.sh
@@ -155,7 +155,7 @@ echo -e "-----BEGIN RSA PRIVATE KEY-----\n${priv}\n-----END RSA PRIVATE KEY-----
    | openssl rsa -inform pem -out "${pdir}/letsencrypt.key"
 
 # Process the certificates for each of the domains in acme.json
-domains=$(jq -r '.Certificates[].Domain.Main' ${acmefile}) || bad_adme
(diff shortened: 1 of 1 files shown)


contrib/scripts/dumpcerts.sh
```diff
@@ -155,7 +155,8 @@ echo -e "-----BEGIN RSA PRIVATE KEY-----\n${priv}\n-----END RSA PRIVATE KEY-----
    | openssl rsa -inform pem -out "${pdir}/letsencrypt.key"
 
 # Process the certificates for each of the domains in acme.json
-for domain in $(jq -r '.Certificates[].Domain.Main' ${acmefile}); do
+domains=$(jq -r '.Certificates[].Domain.Main' ${acmefile}) || bad_adme
```
mmatur:
```suggestion
domains=$(jq -r '.Certificates[].Domain.Main' ${acmefile}) || bad_acme
```

PR #4348, opened 2018-12-31
Allow multiple acme entrypoints

snowzach:
### What does this PR do?
This adds the setting under entryPoint.name.tls `useAcme = true` to use the acme obtained certificates for that entrypoint. It re-assigns the entryPoints certificate store to the same one used for acme. The specified entrypoint should be use to obtain the cert and any additional entrypoints tagged with this setting can use the certs also.

### Motivation
I use acme to obtain a wildcard certificate for my domain. I expose a small handful of services to the internet for which I want valid certs. I also use another entrypoint to expose many other services on another entrypoint internally. I want to be able to use the same wildcard certificate on the internal entrpoint. 

### More

- [X] Added/updated tests

This addresses issue https://github.com/containous/traefik/issues/2710


Morl99:
@ldez is there anything that we can do to get this merged? I am really missing this in order to allow Traefik to work for both internal and external services (running on different ports, one which is exposed to the internet, and the other is only available in the internal network).

Let me know if I can be of any help.

jnovack:
> is there anything that we can do to get this merged?

@snowzach @Morl99 A rebase at this point is pretty much required.  Lots of conflicts.  That will at least make the reviewer less daunting.

jnovack:
Unfortunately, this only helps use-cases where you use wildcard certs, where DNS integration is required.

I (I'm a nobody, I have no say here) would much rather see a solution that permits the second (et. al.) TLS endpoint to ALSO use ACME via HTTP-01 Challenges on the designated http endpoint.

bohmandan:
> Unfortunately, this only helps use-cases where you use wildcard certs, where DNS integration is required.
> 
> I (I'm a nobody, I have no say here) would much rather see a solution that permits the second (et. al.) TLS endpoint to ALSO use ACME via HTTP-01 Challenges on the designated http endpoint.

I must say that this would be very valuable, so not "unfortunately" for me 🙂 
Its perfectly fine that it requires DNS challenge.

juliens:
Thank you for this PR.

Sorry for the late answer.

V2.0 uses the routers to configure tls and certificates, and as routers can be linked to multiple port, it will be possible to have the same let’s encrypt certificate on multiple ports.

lweberk:
What is the release schedule on v2.0?

Issue #4368, opened 2019-01-08
Consul KV provider becomes extremely slow with growing number of frontends/backends


untoldwind:
Bug

### What did you do?

About 100 frontends and backends configured via consul KV.
Consul is running in a 3 node cluster that traefik queries directly (with watch=true).

### What did you expect to see?

"Nearly" immediate configuration updates whenever there is a change in the KV store.

### What did you see instead?

Configuration updates may take up to a minute or longer.

### Output of `traefik version`: 

```
Version:      v1.7.6
Codename:     maroilles
Go version:   go1.11.3
Built:        2018-12-14_06:43:37AM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
[consul]
endpoint = "__CONSUL_ENDPOINTS__"
watch = true
prefix = "traefik"
```

### Analysis

The cause of this problem appears to be the way the KV provider works under the hood.

* There is a "WaitTree" operation, each time a tree update is registered
* ... a new configuration template is generate by querying the relevant keys one by one (or nearly one by one)
* ... this results on about 20 (or so) requests to consul per frontend/backend (i.e. in total A LOT)
* Additionally each of these requests is done if "consistent=true", which generates a lot of overhead inside the consul cluster (i.e. it puts some load on the cluster and is slow by comparison)

### Suggested fix

Especially in case of consul all the additional request are completely unnecessary since the result of WatchTree already contains all KVPairs of the tree. So instead of re-quering everything from the KV store that configuration template might es well use the output of WatchTree directly.

iameli:
Is this still likely to be an issue? We're running into something that looks very similar where changes take 60+ seconds to propagate when we have 500-1000 different routes.

JakeCooper:
Noticed this running against the Redis provider

Sometimes it takes upwards of 3 minutes propagate (~5k services and ~10k routers)

Definitely would love a workaround/ack on this cause we're at the point where I've tried a variety of things and the only thing left to try is ripout Traefik entirely

sdelicata:
Hello!
I close this issue as it targets an unsupported version of Traefik.
Feel free to re-open it or join our [Community Forum](https://community.traefik.io/).

PR #4369, opened 2019-01-08
Optionally use KV pairs from WatchTree

untoldwind:
As described in #4368 this change uses the output from WatchTree instead of re-querying the KV store.

In case of consul this works very well with a large number of frontends/backends. The feature itself is optional and has to be activate by adding a "cacheKeyValues = true" to the provider configuration.

### DIFF
--- a/provider/kv/kv.go
+++ b/provider/kv/kv.go
@@ -24,8 +24,10 @@ type Provider struct {
 	TLS                   *types.ClientTLS `description:"Enable TLS support" export:"true"`
 	Username              string           `description:"KV Username"`
 	Password              string           `description:"KV Password"`
+	CacheKeyValues        bool             `description:"Cache KVs while applying the template"`
 	storeType             store.Backend
 	kvClient              store.Store
+	kvCache               map[string]*store.KVPair
 }
 
 // CreateStore create the K/V store
@@ -71,15 +73,21 @@ func (p *Provider) watchKv(configurationChan chan<- types.ConfigMessage, prefix 
 			select {
 			case <-stop:
 				return nil
-			case _, ok := <-events:
+			case tree, ok := <-events:
 				if !ok {
 					return errors.New("watchtree channel closed")
+				}
+				if p.CacheKeyValues {
+					kvCache := make(map[string]*store.KVPair, len(tree))
+					for _, kv := range tree {
+						kvCache[kv.Key] = kv
+					}
+					p.kvCache = kvCache
 				}
 				configuration, errC := p.buildConfiguration()
 				if errC != nil {
 					return errC
 				}
-
 				if configuration != nil {
 					configurationChan <- types.ConfigMessage{
 						ProviderName:  string(p.storeType),
@@ -114,11 +122,21 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 				}
 			})
 		}
+		if p.CacheKeyValues {
+			tree, err := p.kvClient.List(p.Prefix, nil)
+			if err != nil {
+				return fmt.Errorf("failed to list KV store: %v", err)
+			}
+			kvCache := make(map[string]*store.KVPair, len(tree))
+			for _, kv := range tree {
+				kvCache[kv.Key] = kv
+			}
+			p.kvCache = kvCache
+		}
 		configuration, err := p.buildConfiguration()
 		if err != nil {
 			return err
 		}
-
 		configurationChan <- types.ConfigMessage{
 			ProviderName:  string(p.storeType),
 			Configuration: configuration,
--- a/provider/kv/kv_config.go
+++ b/provider/kv/kv_config.go
@@ -575,7 +575,11 @@ func (p *Provider) listServers(backend string) []string {
 
 func (p *Provider) serverFilter(serverName string) bool {
 	key := fmt.Sprint(serverName, pathBackendServerURL)
-	if _, err := p.kvClient.Get(key, nil); err != nil {
+	if p.CacheKeyValues && p.kvCache != nil {
+		if _, ok := p.kvCache[key]; !ok {
+			return false
+		}
+	} else if _, err := p.kvClient.Get(key, nil); err != nil {
 		log.Errorf("Failed to retrieve value for key %s: %s", key, err)
 		checkError(err)
 
@@ -586,9 +590,16 @@ func (p *Provider) serverFilter(serverName string) bool {
 
 func (p *Provider) checkConstraints(keys ...string) bool {
 	joinedKeys := strings.Join(keys, "")
-	keyPair, err := p.kvClient.Get(joinedKeys, nil)
-	if err != nil {
-		checkError(err)
+	var err error
+	var keyPair *store.KVPair
+
+	if p.CacheKeyValues && p.kvCache != nil {
+		keyPair = p.kvCache[joinedKeys]
+	} else {
+		keyPair, err = p.kvClient.Get(joinedKeys, nil)
+		if err != nil {
+			checkError(err)
+		}
 	}
 
 	value := ""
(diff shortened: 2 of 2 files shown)


GeorgianaElena:
Hi! Is there any progress with this PR? I'm experiencing the same issue with etcd and I can confirm this patch works with this provider also.
@untoldwind many thanks for this awesome feature.

mriehl:
@GeorgianaElena we use the consul provider and the watchTree output when a change happens actually returns the entire tree, not just what changed. This might be different with etcd.

With consul it was getting the entire tree and then running amok re-querying the entire subtrees. Which is the reason for this change.
If it's an option I would suggest moving to consul (we had lots of other problems with etcd) and using this PR, which definitely works wonderfully.

bomgar:
It would be great if someone reviews this. It is a much needed patch. The pull request is so old that there are conflicts already. I tried to rebase it but it is currently untested. Will do that as soon as I have the time https://github.com/21re/traefik/tree/cache-kv

mriehl:
It's probably a good idea to have the config option be less general though...
From https://github.com/abronan/valkeyrie#limitations :

> Calls like WatchTree may return different events (or number of events) depending on the backend (for now, Etcd and Consul will likely return more events than Zookeeper that you should triage properly). Although you should be able to use it successfully to watch on events in an interchangeable way (see the docker/leadership repository or the pkg/discovery/kv package in docker/docker).

So it's quite possible that this is an optimization that only makes sense with consul (but it still makes a lot of sense)

GeorgianaElena:
@mriehl, yes is not very intuitive to have such differences.

From what I understand, the configuration is built based on [kv.tmpl](https://github.com/containous/traefik/blob/v1.7/templates/kv.tmpl#L2).

The first thing the template does is to list everything under *"/backends/"*. In etcd's case, `list` returns all the keys that match but also their **values** . Is it the same for consul?

I'm asking this because I'm wondering if it wouldn't be possible to pass the dict returned by `List .Prefix "/backends/"` to the subsequent functions  (e.g. `getCircuitBreaker`, `getResponseForwarding`...) and search the keys in this dict instead of making other requests to the store.

mriehl:
> The first thing the template does is to list everything under _"/backends/"_. In etcd's case, `list` returns all the keys that match but also their **values** . Is it the same for consul?
Yes I think that's a standard for valkeyrie. Since the return signature is ` ([]*store.KVPair, error)`.
Sounds like another useful optimization, because the requests to the store usually request a fully monotonic read which is very slow...

bomgar:
I updated my branch again. https://github.com/21re/traefik/tree/cache-kv

Would be really nice if someone could review this. This is a critical fix for us. Without it traefik is unusable with consul in an environment with lots of backends.

dduportal:
Hi @bomgar and @untoldwind , this PR is having file conflicts. A quick reminder that the source branch is `v1.7` on [21re's repository](https://github.com/21re/traefik/tree/v1.7): can you make sure that everything is alright please?

We understand that this issue is causing trouble, but as we are working hard on Traefik v2.0, (https://github.com/containous/traefik/projects/3), the time required to review external PRs is short for us. Please note that the priority of the related issue is `P2` (ref. <https://docs.traefik.io/v2.0/contributing/maintainers/#status-workflow>).

So actually, this PR is waiting for one of the core maintainers to do a design review.

If the design is "LGTM-ed", then the PR's label will be updated to "status-2", and the maintainers will be expected to do the code review.
If not, the PR will be commented or refused based on the feedback.

bomgar:
I rebased everything. The pull request is now linked with the right commit. I'm just saying it would be great if someone reviews this before new merge conflicts show up.

wmolyneux:
We are suffering from the same issue that this change is looking to solve. Is there a way we can download an alpha/beta build of this version to validate it solves our issue before we look for an alternative work around to this performance issue?

bomgar:
@wmolyneux just checkout the branch and build the binary. That's the way we are currently using it.

provider/kv/kv.go
```diff
@@ -114,11 +122,21 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
...
+			tree, err := p.kvClient.List(p.Prefix, nil)
+			if err != nil {
+				return fmt.Errorf("failed to list KV store: %v", err)
+			}
+			kvCache := make(map[string]*store.KVPair, len(tree))
+			for _, kv := range tree {
+				kvCache[kv.Key] = kv
+			}
+			p.kvCache = kvCache
+		}
```
mpl: Why do we need this L125-L135 block in addition to what's already done in watchKv?

provider/kv/kv_config.go
```diff
@@ -575,7 +575,11 @@ func (p *Provider) listServers(backend string) []string {
 
 func (p *Provider) serverFilter(serverName string) bool {
 	key := fmt.Sprint(serverName, pathBackendServerURL)
-	if _, err := p.kvClient.Get(key, nil); err != nil {
+	if p.CacheKeyValues && p.kvCache != nil {
```
mpl: there are two things that bother me at every place you do a cache read in this patch:

1) at every cache miss, you consider that the value does not actually exist at all. But the way a cache works is that usually you do

```
val, ok := cache.get(key)
if !ok {
  // cache miss
  val, ok = upstreamSource.get(key)
  if !ok {
    return errors.New("not found")
  }
}
```

So I assume that you took this approach because you think the whole world is in cache, so when we do a get on the cache, it's equivalent to doing a get on the upstream source. But are we sure that is true?

2) I think it would be more readable if we had a Getter on the provider, that is in charge of abstracting away the cache read and the upstream source read, i.e. something like:

```
func (p *Provider) Get(key string) (string, err) {
  if !p.CacheKeyValues || p.kvCache == nil {
    return p.kvClient.Get(key, nil)
  }
  val, ok := p.kvCache[key]
  if !ok {
    return p.kvClient.Get(key, nil)
  }
}
```

I know the way we treat errors or values not found is a bit different at each place in the code, but maybe there's still a way to abstract away some of the repetition?

bomgar:
The cache is always complete because the consul watch returns the complete tree everytime. That's why this is so much faster.

I don't want to add more things to the code while the design review is pending. Please tell me that this is something you consider merging before asking me to invest more time. We opened this 8 months ago.

mpl:
> The cache is always complete because the consul watch returns the complete tree everytime. That's why this is so much faster.
> 
> I don't want to add more things to the code while the design review is pending. Please tell me that this is something you consider merging before asking me to invest more time. We opened this 8 months ago.

Hi,

As far as I can see, the design review label was removed right before my comments were published. So yes, we are currently trying to see if/how this PR could get in.

mriehl:
@mpl right so one big outstanding design problem IMHO is that this entire feature doesn't make any sense for some providers, for instance on ETCD backends I think the watch tree will only return what changed.
So I don't think this PR is in a state where "this line is duplicated" and "we should abstract this so that it is more clear" are useful directions to head into.

I don't think it's good practice to have configuration options that will either boost performance by 1000% or completely break traefik depending on the backend you're using.
Do you have any existing solutions for backend specific optimizations? Unfortunately, the library you're using (valkeyrie) follows a YOLO approach to it:

>Calls like WatchTree may return different events (or number of events) depending on the backend (for now, Etcd and Consul will likely return more events than Zookeeper that you should triage properly). Although you should be able to use it successfully to watch on events in an interchangeable way (see the docker/leadership repository or the pkg/discovery/kv package in docker/docker).
(From https://github.com/abronan/valkeyrie#limitations )

I think for valkeyrie this is a fine decision if you just provide a "notify me if this tree changes" function, but for traefik the performance impact is so brutal that it would make sense to always do this on a consul backend, not just if some obscure configuration flag is enabled.

So in short I think this should be somehow baked into the consul backend, which requires revisiting the kv abstraction entirely (since IMHO using watch tree just to get notified about changes when some backends actually return the entire tree is a sign that the abstraction is not working)

mpl:
> @mpl right so one big outstanding design problem IMHO is that this entire feature doesn't make any sense for some providers, for instance on ETCD backends I think the watch tree will only return what changed. If you enable this flag there, you're in for a hell of a ride.

Ah good to know, thanks.

bomgar:
So what now? Is #4368 something that is going to be fixed in 2.0 anyway? Refactoring the whole thing seems like a waste of time if it is going to be replaces by 2.0 anyway.

mpl:
@bomgar At the moment Consul is not available in V2.0. But yes, when it is reintroduced, it is likely that #4368 will be taken into account in doing so.

PR #4378, opened 2019-01-10
handle errors when working with rancher

apsifly:
This PR adds error handling for requests to rancher api

Before: only first http request was used to check if the server is alive
After: all http requests are checked, if any of them fails, traefik configuration is not updated.

Motivation: sometimes http requests can fail randomly, and this results in inconsistent traefik configuration


SantoDE:
Thanks @apsifly ,

looks better now :) 👍

PR #4388, opened 2019-01-12
Fix Cloudflare Environment Variables

ruffy91:
### What does this PR do?

Fix the Environment Variable names for the cloudflare acme provider

### Motivation

I stumbled over these wrong values when setting up traefik

### More

- [ ] Added/updated tests


ruffy91:
Sorry still wrong, will ad a new PR

PR #4390, opened 2019-01-13
Note about quotes for entrypoint definition with docker-compose

Dragnucs:
### More

- [ ] Added/updated tests


Dragnucs:
@dahefanteng as I see @ldez was way faster than me. Thank you.

PR #4403, opened 2019-01-17
cache exising task definitions to avoid aws rate limiting

emtammaru:
This is copied from https://github.com/containous/traefik/pull/4177 with the change requested by @ldez  -- not sure if @hwhelan-CB is still active, so we're picking up where he left off:

> ### What does this PR do?
> 
> Stores existing task definitions to avoid rate limiting of the AWS describeTask API. If there is an existing definition, use instead of making another API call.
> 
> ### Motivation
> 
> Deployments taking hours instead of minutes. Due to using multiple instances of Traefik, each making frequent desribeTask API calls, and quickly hitting AWS's rate limit.
> 
> ### More
> 
> - [ ] Added/updated tests
> - [ ] Added/updated documentation
>

### DIFF
--- a/provider/ecs/ecs.go
+++ b/provider/ecs/ecs.go
@@ -19,9 +19,11 @@ import (
 	"github.com/containous/traefik/provider"
 	"github.com/containous/traefik/safe"
 	"github.com/containous/traefik/types"
+	"github.com/patrickmn/go-cache"
 )
 
 var _ provider.Provider = (*Provider)(nil)
+var existingTaskDefCache = cache.New(30*time.Minute, 5*time.Minute)
 
 // Provider holds configurations of the provider.
 type Provider struct {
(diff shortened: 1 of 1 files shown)


ldez:
Hello,
We understand that you really seem to want this fix.

You must follow some fair conventions:

- Wait some time (some days) after that changes are been requested.

- Before carrying the code of a PR, open a PR on [the contributor's repository](https://github.com/carbonblack/traefik/tree/cache_ecs_task_definitions).
If the contributor does not respond, let us know and we will put your commits in the main PR.

- The code cannot be copied like that, you have to respect some ownership rules: you have to keep the commits.

Could you close this PR and open a PR on https://github.com/carbonblack/traefik/tree/cache_ecs_task_definitions.
After you can put a link to your PR in #4177

emtammaru:
Ok let’s wait a few more days (months already) for a simple fix 😕

PR #4418, opened 2019-01-22
Update kubernetes.md

cntsw:
### More

- [ ] Added/updated tests


jlevesy:
Hello @cntsw .

Thanks for your pull request.

However we're not going to accept it, because this sentence is actually correct as-is.
What we want to say here is that you have a much better control on scalability using a deployment (you can choose the amount of instances you want) than a daemon set (fixed one instance per node).

PR #4444, opened 2019-01-30
Disable zoom on the web ui

adrienpsn:
Update index.html of the web ui for more compatibility with web browsers.

### What does this PR do?

Thanks to this modification, it is no longer possible to zoom in on the web interface when viewed on a small screen.

### Motivation

Fix bug on web ui.


ldez:
Could you explain why you disable the zoom?

adrienpsn:
> Could you explain why you disable the zoom?

Because the zoom on the responsive side of the interface. I think it's better to disable it!

ldez:
We will follow @Slashgear recommendation.
Accessibility is important for us.

Thanks for your contribution.

Issue #4026, opened 2018-10-11
Support IP blacklisting


negz:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?
Traefik currently supports IP whitelisting at the entrypoint level. I'd like to be able to support blacklisting. It seems like such a feature would be fairly straightforward; I'd be happy to contribute it if there were interest.

For further context; I'm considering how a GeoIP database might be used to block certain requests by IP address at Traefik. These databases tend to have particular file formats, and are occasionally updated. One approach would be to build a specific GeoIP middleware that could read these databases (and ideally reload them without reloading Traefik), but it seems like a simple blacklist functionality might be a good first step in this direction.

RRAlex:
A way to reload the configuration live, with specific blacklisting for a given k8s ingress (based on some logs or a trigger from fail2ban, etc) would be very useful to many I'm sure!

HitkoDev:
I currently use HAProxy with confd in front of Traefik, to fetch blacklisted IPs from etcd. That obviously ads a lot of overhead compared to proposed solution, especially since Traefik instances already connect to the same etcd instance.

aido93:
Guys, what's about geoip? Is this feature in plans or in progress? If someone has already begun this work I can help.

hb3b:
GeoIP/Maxmind support would be helpful to us.

joeldeteves:
GeoIP blocking would be very useful

H2so4:
I have a quick write-up on how to use the ForwardAuth middleware to block a list of IPs https://scaleup.us/2020/06/21/how-to-block-ips-in-your-traefik-proxy-server

I really hope this is built in soon. Traefik is a great modern tool.

This can easily be modified to do blocking based on GeoIP.

notsureifkevin:
Greetings! I wrote a Traefik Plugin for this use case specifically -- feel free to try it out!

https://pilot.traefik.io/plugins/272416229286740480/deny-ip-plugin

PR #4454, opened 2019-02-01
Add IP Blacklisting

mattdavis90:
### What does this PR do?

Added IP Blacklist support, in line with the current IP Whitelisting. Influenced by https://github.com/containous/traefik/issues/4026

### Motivation

Issue https://github.com/containous/traefik/issues/4026

### More

- [ ] Added/updated tests

### DIFF
--- a/autogen/gentemplates/gen.go
+++ b/autogen/gentemplates/gen.go
@@ -309,6 +309,15 @@ var _templatesConsul_catalogTmpl = []byte(`[backends]
       useXForwardedFor = {{ $whitelist.UseXForwardedFor }}
     {{end}}
 
+    {{ $blacklist := getBlackList $service.TraefikLabels }}
+    {{if $blacklist }}
+    [frontends."frontend-{{ $service.ServiceName }}".blackList]
+      sourceRange = [{{range $blacklist.SourceRange }}
+        "{{.}}",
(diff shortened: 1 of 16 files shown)


nmengin:
Hello, @mattdavis90 

Thanks for you contribution :+1: 

We accept only bug fixes on 1.7 https://github.com/containous/traefik#release-cycle

Your PR seems interesting, could you rebase on master?
Thanks.

ldez:
Thanks for your contribution :+1: 

We want to keep the IP filtering section as simple as possible and we think that your use case could be addressed differently.

We think that a blacklisting task can be better achieved using a firewall.

So, for now, and I insist on the "for now", we will decline your pull request.

anantsaraswat:
@ldez I agree it's easy to use firewall for the same, But my ony issue is how we get client IP address? 

Let's say I am running tomcat behind traefik and a hacker run DDOS attack against the same, how will I get the client IP address from traefik? So that I can block that.

helmut72:
> We think that a blacklisting task can be better achieved using a firewall.

A firewall rule, pushed from Fail2ban, will block all services behind Traefik, if someone try too many wrong logins to a specific service.

> We want to keep the IP filtering section as simple as possible

Feels to simple ;)

PR #4459, opened 2019-02-04
Add Tracing Header Context Name option for Jaeger

gadoor:
`jaeger-client-go <= 2.9.0` only allows the context propagation header to be 'uber-trace-id', however, newer versions allow custom header, hence, adding tracingContextHeaderName as optional configuration parameter will enable `Traefik` users to set custom header for jaeger context propagation.

- Kept the default for `TraceContextHeaderName` as `uber-trace-id` for backward compatibility. 
- Changed `jaeger-client-go` version from `2.9.0` to `2.15.0`.

This change also comes to comply with https://github.com/opentracing/specification/blob/master/rfc/trace_identifiers.md


ldez:
@gadoor you don't need to update your PR, we have a bot :wink:

PR #4483, opened 2019-02-11
Introduce support for correlation ids

jan-martens:
### What does this PR do?

This PR adds the possibility to add [correlation id](https://blog.rapid7.com/2016/12/23/the-value-of-correlation-ids/) headers to a frontent. If enabled, every incoming request will have a `X-Correlation-ID` header added containing a random UUID. If the header is already present, it will not get changed. The header name and content are also configurable.

### Motivation

In our environment we make heavy use of correlation ids to track requests throughout our mircoservice infrastructure. We recently moved from Nginx to Traefik and it brought us a lot of benefits, but the possibility to add custom headers is somewhat limited compared to Nginx. We need to add a new random UUID to every requests that comes in from outside our infrastructure to properly track it's path through the enviroment. This could have been done with the help of custom request headers, but those have some disadvantages:

If the header is already present, it will get overwritten. Also you cannot generate a new random UUID for every request. All you could do is generate a random UUID for every frontend each time Traefik reloads the configuration ( with the help of the spring template library ) but then you can have a lot of requests with the same correlation id. This is why we wrote a [http middleware](https://gitlab.com/JanMa/correlation) that adds these headers to every incoming request and would like to add it here.

### More

- [x] Added/updated tests


ldez:
Hello, thanks for you contribution +1

We already had a PR on this subject (https://github.com/containous/traefik/pull/2290).

FYI, we accept only bug fixes on 1.7 https://github.com/containous/traefik#release-cycle

juliens:
Hi and thank you for your contribution and hard work on improving Traefik.

Unfortunately, we cannot accept this pull request “as is” since, as announced previously (and stated by ldez before), the 1.7 branch doesn’t accept new features any more (only master does).

That being said, since it will require quite some work, we encourage you to write and submit a proposal _before_ rebasing this on master because even if it works and meets some  use cases, it’s not guaranteed that every maintainer will agree on the feature being included in the project. We’ve already had related PRs that were rejected (#2290 ), so it’s best to discuss this beforehand.

Thanks again! Looking forward to discussing this further with you!

Sarke:
@JanMa  

```toml
[tracing]
  [tracing.jaeger]
    samplingParam = 0
    traceContextHeaderName = "X-Correlation-ID"
```

```yml
TRAEFIK_TRACING: "true"
TRAEFIK_TRACING_JAEGER_TRACECONTEXTHEADERNAME: "X-Correlation-ID"
```

PR #4506, opened 2019-02-19
NoIP support

nanoak:
I use NoIP dynamic DNS service. I think NoIP support would be great.


dduportal:
Hello @nanoak , could you please open an issue to request such a feature (and use the issue template) please? 

This is a Pull Request to contribute the code, which does not seem what you wanted.

PR #4512, opened 2019-02-21
Enable override of certificate domains

asannes:
### What does this PR do?

This makes it possible to specify which domains are used with a TLS certificate.

```
[[tls]]
  [tls.certificate]
   domains = ["one.example.com", "two.example.com"]
   certFile = "/certs/wc.example.com.chain.crt.pem"
   keyFile = "/certs/wc.example.com.key.pem"
```

### Motivation

Had a case where I had two certificates overlapping, and I wanted to be explicit about which one should be used for what domain.

### More

- [ ] Added/updated tests

### Additional Notes

I'd like some feedback if this is the right approach.


juliens:
Thanks for your contribution.

I think the problem your want to resolve should be resolve by this proposal #4756 because you will be able to assign a specific certificate to a specific store and then use this store in your router.

That's why I will close this PR, if you think you have a use case not resolve by this proposal, don't hesitate to give feedback on the proposal.

asannes:
@juliens Ah, yes. #4756 would cover it. Although in my case it would be more configuration (I would end up with a lot of stores), but the proposal is much more flexible and re-usable in other cases.

:thinking: Only feedback I could give about the proposal is that it might be tricky to choose the certificate when the router has other rules than sni host pointing to different certificates stores... so that has to be well understood.

PR #4515, opened 2019-02-22
Forward all header values from forward auth response

ctas582:
### What does this PR do?

Forwards all header values from the forward authentication response to the destination.

### Motivation

I'm using Traefik as an [authenticating proxy](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#authenticating-proxy) in front of the Kubernetes API server. Traefik is configured to use forward authentication, which returns multiple "X-Remote-Group" headers. Currently, however, only the first of these headers is forwarded on to the destination. The Kubernetes API server requires a separate header to be sent for each group.

### More

- [ ] Added/updated tests


ldez:
Thank you for this PR.

Sorry for the late answer.

Could you rebase this pull request on v2.0 branch?

PR #4532, opened 2019-02-27
Web UI code refactor on health page

imcotton:
### What does this PR do?

- Giving type structure of API calling result to endpoint `/health`, gains static typing.
- Reduce `subscription` in favor of `async` pipe.
- Having field property initialized earlier.
- Having `private readonly` modifier when it's possible.
- Setting change detection strategy as `OnPush`.
- Window resize event hooks on debouncing, and call `unsubscribe` to avoid memory leaking.
-  Passing unit tests.

### Motivation

Best practice of Angular, TypeScript and RxJS.

### Additional Notes

Any concerns, feedbacks are welcome.

### DIFF
--- a/webui/src/app/charts/bar-chart/bar-chart.component.ts
+++ b/webui/src/app/charts/bar-chart/bar-chart.component.ts
@@ -2,22 +2,39 @@ import {
   Component,
   ElementRef,
   Input,
-  OnChanges,
   OnInit,
-  SimpleChanges,
-  OnDestroy
+  OnDestroy,
+  ChangeDetectorRef
 } from '@angular/core';
 import { axisBottom, axisLeft, max, scaleBand, scaleLinear, select } from 'd3';
 import { format } from 'd3-format';
 import * as _ from 'lodash';
 import { WindowService } from '../../services/window.service';
 
+interface DataModel {
+  code: string;
+  count: number;
+}
+
 @Component({
   selector: 'app-bar-chart',
   templateUrl: './bar-chart.component.html'
 })
-export class BarChartComponent implements OnInit, OnChanges, OnDestroy {
-  @Input() value: any;
+export class BarChartComponent implements OnInit, OnDestroy {
+  @Input() set value(data: DataModel[] | null) {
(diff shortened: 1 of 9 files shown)


webui/src/app/charts/bar-chart/bar-chart.component.ts
```diff
@@ -2,21 +2,39 @@ import {
...
+}
+
 @Component({
   selector: 'app-bar-chart',
   templateUrl: './bar-chart.component.html'
 })
-export class BarChartComponent implements OnInit, OnChanges {
-  @Input() value: any;
+export class BarChartComponent implements OnInit, OnDestroy {
+  @Input() set value(data: DataModel[] | null) {
```
Slashgear:
```suggestion
  @Input() set value(data?: DataModel[]) {
```

webui/src/app/charts/line-chart/line-chart.component.ts
```diff
@@ -52,9 +54,22 @@ export class LineChartComponent implements OnChanges, OnInit {
   margin = { top: 40, right: 40, bottom: 60, left: 60 };
   loading = true;
 
+  private readonly resize$$ = this.windowService.resizeDebounce$.subscribe(
+    w => {
```
Slashgear:
```suggestion
    () => {
```

webui/src/app/charts/bar-chart/bar-chart.component.ts
imcotton: This is case before the changing, however in setter case with `async` input, `null` is exact the first value been passed in.

webui/src/app/charts/line-chart/line-chart.component.ts
imcotton: Good catch, will remove this unused parameter.

ldez:
Thanks for your contribution :+1: 

As you can see the WebUI has been drop temporary, we are working on a new WebUI.

I will close your PR but I hope you will help us to improve the future WebUI.

Thank you again.

PR #4555, opened 2019-03-05
Added support for Haystack tracing

aantono:
Adding support for Haystack tracing system - https://expediadotcom.github.io/haystack/

It is an OpenTracing compatible tracing and logging system which supports high volume trace data and distributed backend storage system.


ldez:
Thank you for this PR.

Sorry for the late answer.

Could you rebase this pull request on v2.0 branch?

ldez:
@aantono any news?

aantono:
Sorry, forgot to rebase... Hopefully it's all good now!

aantono:
Looks like `validate-lint` is failing, but can't figure out why :( 
Any ideas what broke during the dependency upgrade?

ldez:
The problem come because you are dep 0.4, we migrated to dep 0.5.

Could you rebase on branch v2.0?

Gopkg.toml
```diff
@@ -282,3 +290,7 @@ required = [
 [[constraint]]
   name = "github.com/instana/go-sensor"
   version = "1.4.12"
+
+[[constraint]]
+  name = "github.com/ExpediaDotCom/haystack-client-go"
+  revision = "3ee3ba12ce07ae900855f7cf3e3f543418e865bf"
```
ldez: why are you using a revision instead of a version? (https://github.com/ExpediaDotCom/haystack-client-go/releases/tag/0.2.3)

Gopkg.toml
```diff
@@ -198,7 +198,15 @@ required = [
 
 [[constraint]]
   name = "google.golang.org/grpc"
-  version = "1.5.2"
+  version = "1.13.0"
+
+[[override]]
+  name = "golang.org/x/sys"
+  revision = "1c9583448a9c3aa0f9a6a5241bf73c0bd8aafded"
```
ldez: could you explain why you need this override?

Gopkg.toml
aantono: Fixed... the commit I needed was available before the 0.2.3 was released at the time of this PR.

Gopkg.toml
aantono: Needed to upgrade the version as it was conflicting with the transitive protobuf dependency for auto-generated code bits (was complaining of some missing methods)

PR #4551, opened 2019-03-04
Add support proxyprotocol v2

c0va23:
### What does this PR do?

This PR adds support for proxyprotocol v2 headers.

The package provides support for proxy protocol v1 (github.com/armon/go-proxyproto) is replaced by the package (github.com/c0va23/go-proxyprotocol) supporting the analysis of both proxy protocol versions at the same time.

### Motivation

Support for proxy protocol v2 need that traefik can be run using AWS NLB.

### More

- [ ] Added/updated tests
- [x] Updated documentation

### Additional Notes

Traefik build with its commits tested with HA Proxy (v1 and v2) and with AWS NLB.

PR #4561, opened 2019-03-06
Add support proxyprotocol v2

c0va23:
### What does this PR do?

This PR adds support for proxyprotocol v2 headers.

The package provides support for proxy protocol v1 (https://github.com/armon/go-proxyproto) is replaced by the package (https://github.com/c0va23/go-proxyprotocol) supporting the analysis of both proxy protocol versions at the same time.

### Motivation

Support for proxy protocol v2 need that traefik can be run using AWS NLB.

### More

- [x] Added/updated tests
- [x] Updated documentation

### Additional Notes

Traefik build with its commits tested with HA Proxy (v1 and v2) and with AWS NLB.

It PR #4551 with fixed integration tests


juliens:
Thanks for your contribution.

First of all, adding ProxyProtocol v2 is more an enhancement than a bug fix, that's why we will not be able to merge this PR on 1.7. You will need to do it on the v2.0 branch.

In addition, the ProxyProtocol library you use is quiet young and has only one star and one contributor. So to be able to trust this library, we will need to do a review of the library before merging this PR.

c0va23:
@juliens, i create PR #4755 with base branch v2.0.

Traefik with patches from its PR we use on production from beginning of march.

PR #4564, opened 2019-03-06
Support external name service on global default backend

kippandrew:
### What does this PR do?

This PR adds support for global backends pointing to an ExternalName services. This is accomplished by checking the service type when adding the global default backend. If the service type is an ExternalName service, the backend is mapped to the ExternalName service's url. This is basically a copy paste of what is done with normal backends. 

### Motivation

It would be useful in my environment (and others I'm sure) for the Kubernetes provider to support a global default backend that's points to an ExternalName service. This is already supported for regular backends so I imagine not supporting it here was an oversight.

### More

- [X] Added/updated tests

### Additional Notes

None


kippandrew:
Bump! Any feedback on this PR? Thanks for your attention.

dtomcej:
I approve the design of this PR. @kippandrew is correct this is intended behavior.

kippandrew:
Bump. Any possibility someone can take a look at this? Thanks!

PR #4595, opened 2019-03-14
Remove everything templates related

mpl:
Also remove examples and contrib directories.

### What does this PR do?

It removes the templates, examples, and contrib directories.
It also removes or disables the parts relying on the templates, and it removes the parts that were generating the templates.

### Motivation

These directories are supposed to be deprecated and should not be used anymore.

### More

I had to disable (with build tags) some of the unit tests of the old providers.


dtomcej:
Screen Shot 2019-03-14 at 9 39 41 AM

Oh ya baby!
:shipit:

PR #4620, opened 2019-03-18
doc: kubernetes CRD provider

mpl:
### What does this PR do?

This PR documents the kubernetes provider with Custom Resource Definition.

### Motivation

Because docs were non-existent, and hence needed.

### More

~~- [ ] Added/updated tests~~

### Additional Notes

Co-authored-by: Jean-Baptiste Doumenjou <jb.doumenjou@gmail.com>


docs/content/middlewares/addprefix.md
```diff
@@ -26,6 +26,20 @@ The AddPrefix middleware updates the URL Path of the request before forwarding i
...
 
+??? example "Kubernetes -- Prefixing with /bar"
+
+    ```yaml
+    apiVersion: traefik.containo.us/v1alpha1
+    kind: Middleware
+    metadata:
+      name: addprefix
+      namespace: default
+
```
dtomcej: Empty line not needed

docs/content/middlewares/addprefix.md
```diff
@@ -26,6 +26,20 @@ The AddPrefix middleware updates the URL Path of the request before forwarding i
...
     ```
 
+??? example "Kubernetes -- Prefixing with /bar"
+
+    ```yaml
+    apiVersion: traefik.containo.us/v1alpha1
+    kind: Middleware
+    metadata:
+      name: addprefix
+      namespace: default
```
dtomcej: Don't need to define a namespace if it is the default

mpl: Sure. but do you think it harms readability or clarity to keep it? And if not, do you suggest removing it in all other places as well?

dtomcej: mmmm Not sure. I'm ok keeping it, and revisiting at a later date :)

mpl: Ok, since they were all using "default", I've removed all instances of namespaces, for the sake of clarity. I guess we can re-add some when we have specific examples of dedicated namespace usage.

tlvenn:
Hi, maybe a little out of scope but reading the doc, it's not clear to me if the old way (using labels) is still supported.

ldez:
Hi @tlvenn, Traefik has 2 modes in v2:
- ingress strict (only ingress spec without annotations support)
- CRD - IngressRoute (the subject of this PR), in this case you don't need annotations (and they are not supported) because the spec allow to define all options :tada: .

For more information, join us on Slack https://slack.traefik.io

PR #4621, opened 2019-03-18
Add support for tcp labels in docker provider

juliens:
### What does this PR do?
Handle labels for tcp routing in the docker provider

### Motivation
Be able to do tcp with the docker provider

### More
- [x] Added/updated tests

### DIFF
--- a/docs/content/providers/docker.md
+++ b/docs/content/providers/docker.md
@@ -247,8 +247,8 @@ You can declare TCP Routers and/or Services using labels.
 
 !!! warning "TCP and HTTP"
 
-    If you declare TCP Routers/Services, we will not create any HTTP Routers/Services automatically.
-    If you declare both TCP Routers/Services and HTTP Routers/Services we will create both.
+    If you declare a TCP Router/Service, it will prevent Traefik from automatically create an HTTP Router/Service (like it does by default if no TCP Router/Service is defined).
+    You can declare both a TCP Router/Service and an HTTP Router/Service for the same container (but you have to do so manually).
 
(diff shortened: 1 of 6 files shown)


ldez:
Could you add some documentation.

docs/content/providers/docker.md
```diff
@@ -223,11 +223,33 @@ You can declare pieces of middleware using labels starting with `traefik.http.mi
...
+           # ...
+           labels:
+             - traefik.tcp.routers.my-router.rule="HostSNI(`my-host.com`)"
+             - traefik.tcp.routers.my-router.rule.tls="true"
+             - traefik.tcp.services.my-service.loadbalancer.server.port="4123"
+    ```
+
+!!! warning "TCP and HTTP"
+
+    If you declare TCP Routers/Services, we will not create any HTTP Routers/Services automatically.
```
geraldcroes:
```suggestion
    If you declare a TCP Router/Service, it will prevent Traefik from automatically create an HTTP Router/Service (like it does by default if no TCP Router/Service is defined).
```

docs/content/providers/docker.md
```diff
@@ -223,11 +223,33 @@ You can declare pieces of middleware using labels starting with `traefik.http.mi
...
+    If you declare both TCP Routers/Services and HTTP Routers/Services we will create both.
```
geraldcroes:
```suggestion
    You can declare both a TCP Router/Service and an HTTP Router/Service for the same container (but you have to do so manually).
```

PR #4625, opened 2019-03-19
Fix problem in aggregator provider

juliens:
### What does this PR do?
Fix a concurrent bug in provider aggregator

### Motivation
Fix a bug


pkg/provider/aggregator/aggregator.go
```diff
@@ -82,6 +82,7 @@ func (p ProviderAggregator) Provide(configurationChan chan<- config.Message, poo
 	}
 
 	for _, prd := range p.providers {
+		prd := prd
```
emilevauge: 😭

PR #4629, opened 2019-03-19
Fix log msgs about label selector

mpl:
### What does this PR do?

In the kubernetes CRD provider, this PR changes the ingressLabelSelector var name into labelSelector, and it removes references to the word "ingress" in the related log messages.

### Motivation

This is the kubernetes CRD code, so it has nothing to do with the Kubernetes Ingress, which is why references to Ingress should be removed.

### More

~~- [ ] Added/updated tests~~
~~- [ ] Added/updated documentation~~


pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -95,7 +95,9 @@ func (p *Provider) Provide(configurationChan chan<- config.Message, pool *safe.P
 		return err
 	}
 
-	logger.Debugf("Using Ingress label selector: %q", p.LabelSelector)
+	if p.LabelSelector != "" {
```
ldez: could you remove this `if`

mpl: done

Issue #4626, opened 2019-03-19
dnsChallenge: LEGO does not wait for keyboard input -> acme: error presenting token: EOF


ntaranov:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I'm trying to set up Let's Encrypt authentication using dnsChallenge. I've done this with certbot before.

I set up the challenge using traefik command line arguments in docker-compose file. Then I deployed the configuraiton using `docker stack deploy`. I read the logs (noted that the container didn't wait for input on the initial occasion. 

I added the TXT records required. Then I restared the container like this:
```
docker stop <id>
docker start -ia <id>
```

When I restart the container, the TXT records required are the same all the `lego: ` messages have new current timestamps. 

This might be a traefik issue or LEGO issue or I might do something wrong. 

### What did you expect to see?

Either traefik container/LEGO waiting for input ("press ENTER") on initial deploy or successful check completion on restart. 

### What did you see instead?

```
lego: Please create the following TXT record in your kindrobots.com. zone:
_acme-challenge.finance.kindrobots.com. 120 IN TXT "1NYC_MP1EmQVWk9xqY9Yr8CCU-n-twblDBd68S82WKM"
lego: Press 'Enter' when you are done
time="2019-03-19T10:01:29Z" level=info msg="legolog: [INFO] [eng.kindrobots.com] acme: Preparing to solve DNS-01"
lego: Please create the following TXT record in your kindrobots.com. zone:
_acme-challenge.eng.kindrobots.com. 120 IN TXT "EWAwsyme0fzsQ_fNrkAeJeSLW_kt-l05Y_GmOfPou-g"
lego: Press 'Enter' when you are done
time="2019-03-19T10:01:29Z" level=info msg="legolog: [INFO] [eng-api.kindrobots.com] acme: Preparing to solve DNS-01"
lego: Please create the following TXT record in your kindrobots.com. zone:
_acme-challenge.eng-api.kindrobots.com. 120 IN TXT "8iuV5EP9H8OPrG24iJWmOds_uFM-2lIbNnnhomh0d54"
lego: Press 'Enter' when you are done
time="2019-03-19T10:01:30Z" level=info msg="legolog: [INFO] [auth.kindrobots.com] acme: Preparing to solve DNS-01"
lego: Please create the following TXT record in your kindrobots.com. zone:
[...]

PR #4636, opened 2019-03-19
Clarify that manual dnsChallenge provider works only when run as docker run -it

ntaranov:
### What does this PR do?

Updates documentation as discussed with @dduportal 
Makes it more clear that manual dnsChallenge provider requires image to be run with docker run -it and that it isn't going to work with just docker stack deploy

### Motivation

Save others from stumbling upon this
https://github.com/containous/traefik/issues/4626

### More

- [ ] Added/updated tests
- [v] Added/updated documentation

### DIFF
--- a/docs/content/https-tls/acme.md
+++ b/docs/content/https-tls/acme.md
@@ -145,7 +145,7 @@ Do not hesitate to complete it.
 | [Lightsail](https://aws.amazon.com/lightsail/)              | `lightsail`    | `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `DNS_ZONE`                                                                                    | Not tested yet                 |
 | [Linode](https://www.linode.com)                            | `linode`       | `LINODE_API_KEY`                                                                                                                            | Not tested yet                 |
 | [Linode v4](https://www.linode.com)                         | `linodev4`     | `LINODE_TOKEN`                                                                                                                              | Not tested yet                 |
-| manual                                                      | -              | none, but you need to run Traefik interactively, turn on `acmeLogging` to see instructions and press <kbd>Enter</kbd>.                      | YES                            |
+| manual                                                      | -              | none, but you need to run Traefik interactively (that is, `docker run -it`, [^4]), turn on `acmeLogging` to see instructions and press <kbd>Enter</kbd>.                      | YES                            |
 | [MyDNS.jp](https://www.mydns.jp/)                           | `mydnsjp`      | `MYDNSJP_MASTER_ID`, `MYDNSJP_PASSWORD`                                                                                                     | YES                            |
 | [Namecheap](https://www.namecheap.com)                      | `namecheap`    | `NAMECHEAP_API_USER`, `NAMECHEAP_API_KEY`                                                                                                   | YES                            |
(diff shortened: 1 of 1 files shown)


docs/content/https-tls/acme.md
```diff
@@ -169,9 +169,10 @@ Do not hesitate to complete it.
...
-[^2]: [providing_credentials_to_your_application](https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application)
-[^3]: [google/default.go](https://github.com/golang/oauth2/blob/36a7019397c4c86cf59eeab3bc0d188bac444277/google/default.go#L61-L76)
+[^1]: more information about the HTTP message format can be found [here](https://go-acme.github.io/lego/dns/httpreq/)  
+[^2]: [providing_credentials_to_your_application](https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application)  
+[^3]: [google/default.go](https://github.com/golang/oauth2/blob/36a7019397c4c86cf59eeab3bc0d188bac444277/google/default.go#L61-L76)  
+[^4]: as there is no way to support terminal attached to container when deploying with `docker stack`, you might resort to first running the container interactively (that is with `docker run -it`) to generate certificates and then reuse the certificates inside your docker stack container. You can mount a named volume with the same name containing the acme storage file to both containers to achieve this.
```
ldez:
```suggestion
[^4]: when deploying with `docker stack`, as there is no way to support terminal attached to container.
```

docs/content/https-tls/acme.md
```diff
@@ -145,7 +145,7 @@ Do not hesitate to complete it.
...
 | [Linode](https://www.linode.com)                            | `linode`       | `LINODE_API_KEY`                                                                                                                            | Not tested yet                 |
 | [Linode v4](https://www.linode.com)                         | `linodev4`     | `LINODE_TOKEN`                                                                                                                              | Not tested yet                 |
-| manual                                                      | -              | none, but you need to run Traefik interactively, turn on `acmeLogging` to see instructions and press <kbd>Enter</kbd>.                      | YES                            |
+| manual                                                      | -              | none, but you need to run Traefik interactively (that is, `docker run -it`, [^4]), turn on `acmeLogging` to see instructions and press <kbd>Enter</kbd>.                      | YES                            |
```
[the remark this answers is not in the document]
ntaranov: Makes sense

docs/content/https-tls/acme.md
ntaranov: I would leave the `docker run -it` part in the footnote to avoid confusion or put something equally clarifying. Firstly docker is already mentioned here. Secondly, the whole point is that "interactively" word IMO is to fuzzy to map to `docker run -it ...`  intuitively, and there are other commands that might appear "interactive" like `docker attach`  or `docker start -ai ...`, but they wouldn't work. I basically was going round in circles before @dduportal hinted that just attaching system streams might not be quite the same as doing `docker run -it ...`. After all, the fact is that LEGO is in fact launched "interactively" inside the container anyway, it's hard to point out when it's "interactive" enough, on the other hand it's very easy to figure out when it's `docker run -it` enough ;)

ldez: you can add the `docker run -it` in the footnote

ntaranov:
Ready for push if the text is alright.

ntaranov:
@ldez ?

PR #4643, opened 2019-03-21
Add _FILE Environment Variable Documentation

### LINKED ISSUE #3853: Provide a file option for all sensitive

Feature

treafik should support defining a file as a source of sensitive informations like API Tokens, passwords, ssh keys and so on, like is already possible for SSL Certificates and such.

For example, [Docker Swarm secrets](https://docs.docker.com/engine/swarm/secrets/#how-docker-manages-secrets) store the secrets on `/run/secrets/<secret_name>`, and could be used to store `CLOUDFLARE_API_KEY` [DNS provider configuration](https://docs.traefik.io/configuration/acme/#provider) for ACME (Let's Encrypt) Configuration

Read: https://diogomonica.com/2017/03/27/why-you-shouldnt-use-env-variables-for-secret-data/

#### ENV variables 

The suffix `_FILE` could be added to every ENV variable that already expects a secret maybe is a good starting point, and it's being used in official docker images [like this](url
https://github.com/docker-library/mysql/blob/master/8.0/docker-entrypoint.sh#L21-L41).

Ex.: 

```bash
export CLOUDFLARE_EMAIL=me@example.com
export CLOUDFLARE_API_KEY_FILE="/run/secrets/cloudflare_api_key"
```

#### treafik TOML and CLI parameters

To keep consistency, in the same way a suffix is suggested for ENV variables, a suffix `_file` and `File` could be used for TOML and CLI parameters respectively.

* TOML

```toml
[etcd]
endpoint = "127.0.0.1:2379"
watch = true
prefix = "/traefik"
username = "myuser"
password_file = "/run/secret/etcd_password"
```

* CLI

`traefik ... --consul.httpTokenFile` see #3602 

### Considerations

The parameters `some_secret` an `some_secret_file` are exclusives, and should give an error if both are defined.

dargmuesli:
### What does this PR do?
Closes #3853.

### Motivation
A pointer to the existence of Docker secret support within Traefik's documentation was missing.

### More

### Additional Notes
None.

### DIFF
--- a/docs/configuration/acme.md
+++ b/docs/configuration/acme.md
@@ -273,6 +273,8 @@ Useful if internal networks block external DNS queries.
 
 Here is a list of supported `provider`s, that can automate the DNS verification, along with the required environment variables and their [wildcard & root domain support](/configuration/acme/#wildcard-domains) for each.
 Do not hesitate to complete it.
+Every lego environment variable can be overriden by their respective `_FILE` counterpart, which should have a filepath to a file that contains the secret as its value.
+For example, `CF_API_EMAIL_FILE=/run/secrets/traefik_cf-api-email` could be used to provide a Cloudflare API email address as a Docker secret named `traefik_cf-api-email`.
 
 | Provider Name                                               | Provider Code  | Environment Variables                                                                                                                       | Wildcard & Root Domain Support |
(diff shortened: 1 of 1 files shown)


dargmuesli:
The typo is fixed now :)

PR #4652, opened 2019-03-22
Full ACME+CRD example

mpl:
### What does this PR do?

This PR adds a documentation page for a working configuration example of Traefik with Kubernetes through the IngressRoute CRD, and TLS with Let's Encrypt.

### Motivation

It will help people get started with Traefik and Kubernetes. Also we wanted to be absolutely sure that TLS with Let's Encrypt works with the kubernetes CRD provider.

### More

~~- [ ] Added/updated tests~~

### Additional Notes

Co-authored-by: Jean-Baptiste Doumenjou <jb.doumenjou@gmail.com>


docs/content/getting-started/crd-acme/index.md
```diff
@@ -0,0 +1,73 @@
...
+--8<-- "content/getting-started/crd-acme/02-services.yml"
+```
+
+Next, the deployments, i.e. the actual pods behind the services. Again, one pod for Traefik, and one for the whoami app.
+
+```yaml
+--8<-- "content/getting-started/crd-acme/03-deployments.yml"
+```
+
+Now, as an exception to what we said above, please note that you should not let the ingressRoute resources below be applied automatically to your cluster. The reason is, as soon as the acme provider of Traefik detects we have TLS routers, it will try to generate the certificates for the corresponding domains. And this will not work, because as it is, our Traefik pod is not reachable from the outside, which will make the ACME TLS challenge fail. Therefore, for the whole thing to work, we must delay applying the ingressRoute resources until we have port-forwarding set up properly, which is the next step.
```
SantoDE: Maybe we should add a note somewhere here, that upfront, you should have set up the chosen hostname to go against the public ip of the machine you're executing this? Otherwise TLS challenge will fail, right?

mpl: Done.

docs/content/user-guides/crd-acme/04-ingressroutes.yml
```diff
@@ -0,0 +1,30 @@
...
+  entrypoints:
+    - websecure
+  routes:
+  - match: Host(`your.domain.name`) && PathPrefix(`/tls`)
+    kind: Rule
+    services:
+    - name: whoami
+      port: 80
+  tls:
+    secretName: ""
```
jbdoumenjou:
```suggestion
    secretName: ""
    
```

integration/fixtures/k8s/03-ingressroute.yml
```diff
@@ -14,5 +14,4 @@ spec:
     priority: 12
     services:
     - name: whoami
-      port: 80
-
+      port: 80
```
jbdoumenjou:
```suggestion
      port: 80
      
```

docs/content/user-guides/crd-acme/04-ingressroutes.yml
mpl: done

integration/fixtures/k8s/03-ingressroute.yml
mpl: done

docs/content/user-guides/crd-acme/01-crd.yml
```diff
@@ -0,0 +1,89 @@
...
+metadata:
+  name: traefik-ingress-controller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: traefik-ingress-controller
+subjects:
+  - kind: ServiceAccount
+    name: traefik-ingress-controller
+    namespace: default
```
dtomcej: Needs a newline at the end of the file

docs/content/user-guides/crd-acme/01-crd.yml
```diff
@@ -0,0 +1,89 @@
...
+    plural: middlewares
+    singular: middleware
+  scope: Namespaced
+
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: traefik-ingress-controller
+rules:
```
dtomcej: The other yaml portions of the file have a space separating the metadata from the following blocks.
Please remove them, or add one here for consistency 😄

docs/content/user-guides/crd-acme/03-deployments.yml
```diff
@@ -0,0 +1,78 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: traefik-ingress-controller
+
```
dtomcej: Since you referenced a namespace in the CRB, you need to add the namespace here to confirm

docs/content/user-guides/crd-acme/03-deployments.yml
```diff
@@ -0,0 +1,78 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: traefik
+  template:
+    metadata:
+      labels:
+        app: traefik
+    spec:
+      serviceAccountName: traefik-ingress-controller
```
dtomcej: Similarly, since the SA is namespaced, we have to match, and add the namespace to this object too

docs/content/user-guides/crd-acme/03-deployments.yml
```diff
@@ -0,0 +1,78 @@
...
+  replicas: 2
+  selector:
+    matchLabels:
+      app: whoami
+  template:
+    metadata:
+      labels:
+        app: whoami
+    spec:
+      serviceAccountName: traefik-ingress-controller
```
dtomcej: I woudn't want to run a whoami with the SA of the ingress controller ;)

docs/content/user-guides/crd-acme/04-ingressroutes.yml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  entrypoints:
+    - websecure
+  routes:
+  - match: Host(`your.domain.name`) && PathPrefix(`/tls`)
+    kind: Rule
+    services:
+    - name: whoami
+      port: 80
+  tls:
```
dtomcej: Is this required? It seems wrong that an "empty" tls section is required.

docs/content/user-guides/crd-acme/03-deployments.yml
mpl: Well, things as they are right now seem to be working. Whereas when I tried your suggestion, they started to break, so I probably misunderstood what you meant, and it's making my heard hurt right now. I'll have another pass at it later.

docs/content/user-guides/crd-acme/01-crd.yml
mpl: done

docs/content/user-guides/crd-acme/01-crd.yml
mpl: done

docs/content/user-guides/crd-acme/03-deployments.yml
mpl: ok, removed.

docs/content/user-guides/crd-acme/04-ingressroutes.yml
mpl: you already found out in the meantime, but as a reference for others: see https://github.com/containous/traefik/pull/4654

docs/content/user-guides/crd-acme/03-deployments.yml
mpl: done.

docs/content/user-guides/crd-acme/03-deployments.yml
mpl: ok, I think I've addressed your comment while not breaking anything. PTAL.

dtomcej: Usually I would put the namespace on the deployment object, not on the pod spec. If the deployment is created in another namespace, it could get weird.

mpl: I figured it out. The error I was getting, when adding a namespace on the deployment object i.e.: "failed to process /var/lib/rancher/k3s/server/manifests/03-deployments.yml: got extensions/v1beta1, Kind=Deployment objects that are both namespaced and not namespaced"
was because then we have to add a namespace on the _other_ deployment object (the whoami one) as well.

PR #4654, opened 2019-03-22
Document the TLS with ACME case

mpl:
### What does this PR do?

When we did the kubernetes crd implementation, we forgot to test the
case for when one wants TLS, but handled with Let's Encrypt, i.e.
without having to provide a Kubernetes Secret. We assumed it would be
enough to provide (in YAML) a tls object with no field set, which would
get us a non-nil IngressRouteSpec.TLS, allowing us to use
IngressRouteSpec.TLS as the sentinel for whether TLS should be enabled.

However, as IngressRouteSpec.TLS is a pointer, a tls object with no
fields set will actually result in IngressRouteSpec.TLS being nil. This
means at least one of the fields of the YAML tls object must exist.
Therefore, we now use the secretName field value (for now, as it is
the only field anyway) as the sentinel for whether Let's Encrypt should
be used.

This PR documents the above behaviour, and adds a unit test for it.

In addition, this PR fixes a related bug in the ACME provider: when a
router is not configured with TLS enabled, the ACME provider does not
try anymore to generate a (useless) certificate for the corresponding
domain.

### Motivation

As described above, documenting this common case was needed.

### More

- [X] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/provider/acme/provider.go
+++ b/pkg/provider/acme/provider.go
@@ -383,6 +383,9 @@ func (p *Provider) watchNewDomains(ctx context.Context) {
 			case config := <-p.configFromListenerChan:
 				if config.TCP != nil {
 					for routerName, route := range config.TCP.Routers {
+						if route.TLS == nil {
+							continue
+						}
 						ctxRouter := log.With(ctx, log.Str(log.RouterName, routerName), log.Str(log.Rule, route.Rule))
 
 						domains, err := rules.ParseHostSNI(route.Rule)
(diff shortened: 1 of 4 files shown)


dtomcej:
Oof this feels bad, perhaps a new boolean instead? Idk. This will work short term, but I think we should come up with a better way to do this before GA 😅

mpl:
> Oof this feels bad, perhaps a new boolean instead? Idk. This will work short term, but I think we should come up with a better way to do this before GA 😅

yeah, it's far from ideal.

raelga:
Maybe is just a visualisation issue, but the raw dashboard api seems to expect `entryPoints` in snakeCase in the IngressRoute manifest, otherwise the api shows "null"as endpoint:

❯ curl -sSL http://traefik-v2:8080/api/rawdata | jq .

```json
{
  "ACME": {
    "HTTP": {},
    "TCP": {},
    "TLSOptions": null,
    "TLSStores": null
  },
  "kubernetescrd": {
    "HTTP": {
      "routers": {
        "default/cats-fe7c78d104ccea9b6f28": {
          "entryPoints": [
            "web"
          ],
          "middlewares": [
            "default/rmpath"
          ],
          "service": "default/cats-fe7c78d104ccea9b6f28",
          "rule": "Path(`/cache`)"
        },
...
```

PR #4716, opened 2019-04-03
Throttle Kubernetes config refresh

benweissmann:
### What does this PR do?

This is an in-progress change to throttle how frequently we refresh the Kubernetes ingress configuration.

This isn't ready to merge; I wanted to start a discussion about the approach and share my results testing this change in my production environment. 

### Motivation

My team is in the process of moving from the Nginx ingress controller to Traefik. Overall, it's been going well but we saw concerningly high CPU usage from Traefik. We're not handling very much traffic, but each of our 19 Traefik pods was using around 200 millicores, with spikes up to 700-800 millicores. For reference, our nginx pods were using around 40-50 millicores for a heaver traffic load.

Our cluster is roughly 25 nodes, but we're running many more independent services than is typical for a Kubernetes cluster -- we have about 2000 ingresses and 700 services. 

I took a pprof of the traefik process:

profile-pre

I noticed that most of the time was spent loading the configuration from Kubernetes and diffing it with the current configuration. If I understand correctly, Traefik throttles these updates *after* they've been sent from the Kubernetes provider to [server_configuration.go](https://github.com/containous/traefik/blob/v1.7/server/server_configuration.go#L440) -- so when we get a lot of updates from Kubernetes, Traefik spends a lot of time generating and diffing configurations that are then immediately discarded by the throttler.

The PR is a first stab at *also* throttling in the Kubernetes provider. We've deployed this branch to our production environment, with a 10-second throttle, and observed that our CPU usage has declined to roughly 25 millicores (almost a 10x improvement). Here's the pprof after applying this PR:

profile001

Even throttled that heavily, you can see that we're still spending most of our time refreshing the Kubernetes config, but it's definitely a big improvement. I think there's a lot of additional performance that could be squeezed out of the code that refreshes the config, but starting with this throttling is a good first step.

### Open Questions

- Is this approach generally reasonable?
- How should we configure the throttling? I think the ideal thing would be to use the same throttle configuration as `server_configuration.go`, but I don't think that config value is passed down to the Kubernetes provider
- Are there other low-hanging performance improvements we could make to improve performance in clusters with a large number of ingresses, services, and endpoints? 
- I've been working on top of the `v1.7` branch because I wanted to test this in our production environment -- is this something that can make it into the `v1.7` branch or should I work against the `master` branch?

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -140,12 +140,13 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 				}
 			}
 
-			eventsChanBuffered := eventsChan
+			eventsChanToRead := eventsChan
 			throttleDuration := time.Duration(p.ThrottleDuration)
 			if throttleDuration > 0 {
 				// Create a buffered channel to hold the pending event (if we're
 				// delaying processing the event due to throttling)
 				eventsChanBuffered := make(chan interface{}, 1)
+				eventsChanToRead = eventsChanBuffered
 
 				// Run a goroutine that reads events from eventChan and does a
 				// non-blocking write to pendingEvent. This guarantees that writing to
@@ -176,7 +177,7 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 				select {
 				case <-stop:
 					return nil
-				case event := <-eventsChanBuffered:
+				case event := <-eventsChanToRead:
 					// Note that event is the *first* event that came in during this
 					// throttling interval -- if we're hitting our throttle, we may have
 					// dropped events. This is fine, because we don't treat different
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -23,6 +23,7 @@ import (
 	"github.com/containous/traefik/safe"
 	"github.com/containous/traefik/tls"
 	"github.com/containous/traefik/types"
+	"github.com/eapache/channels"
 	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
@@ -137,16 +138,42 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 					return nil
 				}
 			}
+
+			throttledEvents := channels.NewRingChannel(1)
+			go func() {
+				for {
+					select {
+					case <-stop:
+						return
+					case nextEvent := <-eventsChan:
+						throttledEvents.In() <- nextEvent
+					}
+				}
+			}()
+
+			// TODO: should use Traefik config
+			throttlePeriod := 5 * time.Second
+			throttleEnv := os.Getenv("KUBE_EVENT_THROTTLE_SECONDS")
+			if throttleEnv != "" {
+				throttleEnvParsed, err := strconv.Atoi(throttleEnv)
+				if err != nil {
+					log.Errorf("Invalid KUBE_EVENT_THROTTLE_SECONDS: %s", throttleEnv)
+				} else {
+					throttlePeriod = time.Duration(throttleEnvParsed) * time.Second
+				}
+			}
+
 			for {
 				select {
 				case <-stop:
 					return nil
-				case event := <-eventsChan:
+				case event := <-throttledEvents.Out():
 					log.Debugf("Received Kubernetes event kind %T", event)
 					templateObjects, err := p.loadIngresses(k8sClient)
 					if err != nil {
 						return err
 					}
+
 					if reflect.DeepEqual(p.lastConfiguration.Get(), templateObjects) {
 						log.Debugf("Skipping Kubernetes event kind %T", event)
 					} else {
@@ -156,6 +183,8 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 							Configuration: p.loadConfig(*templateObjects),
 						}
 					}
+
+					time.Sleep(throttlePeriod)
 				}
 			}
 		}
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -140,38 +140,8 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 				}
 			}
 
-			eventsChanToRead := eventsChan
 			throttleDuration := time.Duration(p.ThrottleDuration)
-			if throttleDuration > 0 {
-				// Create a buffered channel to hold the pending event (if we're
-				// delaying processing the event due to throttling)
-				eventsChanBuffered := make(chan interface{}, 1)
-				eventsChanToRead = eventsChanBuffered
-
-				// Run a goroutine that reads events from eventChan and does a
-				// non-blocking write to pendingEvent. This guarantees that writing to
-				// eventChan will never block, and that pendingEvent will have
-				// something in it if there's been an event since we read from that
-				// channel
-				go func() {
-					for {
-						select {
-						case <-stop:
-							return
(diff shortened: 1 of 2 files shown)


dtomcej:
Hello @benweissmann, thanks for submitting this as a conversation topic. Hopefully my answers to your questions will give more insight.

>Is this approach generally reasonable?

Reasonable? Sure. However, there are tradeoffs with this approach.
 
>How should we configure the throttling? I think the ideal thing would be to use the same throttle configuration as server_configuration.go, but I don't think that config value is passed down to the Kubernetes provider

In 1.x, each of the providers collects data from their sources, and generates a templated configuration, which is passed back to the server core. The server core combines the configurations and then applies the config.

Most providers are able to give a state snapshot, which is transactional, and allows for complete state to be attained.

The kubernetes provider, on the other hand, streams events and updates asynchronously, and therefore causes issues with that process. Each event would trigger a provider update, and each provider update would trigger a server reload. This lead very quickly to the traefik instance almost constantly reloading its configuration.

The solution decided upon was to store a copy of the _last_ kubernetes provider config, and compare it to the new provider config, to see if there were any changes (so naming changes etc would not be a change). If not, the config would be discarded, and the server core would not refresh.

The main issue with throttling/caching/combing event updates is that they lead to delays in implementation, disconnected state, and introduce troubleshooting complications.

A simple way to do this would be to cache 10 events, and only flush the cache and update the config when the 10 events are full. It would reduce load by 90% for heavy users, but could lead to configurations not being applied in small clusters.

> Are there other low-hanging performance improvements we could make to improve performance in clusters with a large number of ingresses, services, and endpoints?

Yes, anything that deals with string modification/etc is heavy on CPU load (annotations etc), could possibly be improved.

>I've been working on top of the v1.7 branch because I wanted to test this in our production environment -- is this something that can make it into the v1.7 branch or should I work against the master branch?

The master branch, and the v2 branch are for the new Traefik 2.0, and have a completely rewritten provider. There are no templates in the new version, and performance optimization will require different solutions.

The 1.7 branch is fine for optimizations/bugfixes, but be aware that these optimizations will not be carried forward to 2.0

benweissmann:
Thanks for the feedback, @dtomcej !

> The main issue with throttling/caching/combing event updates is that they lead to delays in implementation, disconnected state, and introduce troubleshooting complications.
> 
> A simple way to do this would be to cache 10 events, and only flush the cache and update the config when the 10 events are full. It would reduce load by 90% for heavy users, but could lead to configurations not being applied in small clusters.

Right, and I don't think that's a good tradeoff -- anything that would lead to config not being eventually applied seems like a bug.

What do you think about the implementation in this PR? I based it off the throttling in [server_configuration.go](https://github.com/containous/traefik/blob/v1.7/server/server_configuration.go#L440) -- so we'll always update the config, we just might delay the update a bit / fold multiple events into one if a lot of events come all at once. So in small clusters (where there's less than 1 event per throttling interval), the behavior will be exactly the same -- we'll refresh the config as soon as the event comes in. With a lot of events, we'll end up refreshing a maximum of once per throttling interval, but we'll always end up doing a refresh after the last event we receive -- so the end behavior should be the same.

Moreover, Traefik *already* does this throttling/collapsing in `server_configuration.go` -- this PR is just trying to push that optimization upstream so we can avoid doing work in the Kubernetes provider that will just get thrown away by the throttler in `server_configuration.go`.

> 
> The 1.7 branch is fine for optimizations/bugfixes, but be aware that these optimizations will not be carried forward to 2.0

OK -- I'll continue working on this PR against 1.7, and then once it's landed on that branch I'll open a new PR again 2.0 with the same optimizations. I think this throttling optimization is relatively straightforward in both branches, but I'll defer any lower-level work on optimizing the actual refresh process to the 2.0 branch to avoid duplicating work.

benweissmann:
Hi @dtomcej, just wanted to check in about this issue. We've been running my branch on our production infrastructure for about two weeks now, and it's been correctly picking up configuration changes, etc. We've seen a roughly 10x decrease in CPU usage from my branch vs. the stable branch.

Is this PR (or something like it) something you'd be willing to merge, and if so, what can I do to get it to a mergeable state? We can't really go back to using Traefik without this fix, and we'd obviously prefer to be on mainline and be able to contribute this performance improvement back to the community. I'm also happy to rebase on top of v2.0 and work on getting this merged in to that branch if that's where the focus is now.

dtomcej:
Hey @benweissmann, This seems like something that would be immensely useful.

I would love it if we could implement it in our v2 providers also. However, there are lots of changes regarding k8s in v2, so I think for now, the best thing is to get this PR happy, then we can review getting the functionality in v2.

Some things that I would want to see:

1. Throttling disabled by default
2. Enabled by the kubernetes traefik config
3. I'm not sure that package used is the best option for implementation, it is a bit old...Is there no more current package that we can use? or at least use the native go channels to accomplish? Just not sure using a 5 year old package is the right option here.

Thanks!

benweissmann:
Great, I'll work on getting it ready to land on 1.7.

For configuration, I can think of a couple ways to handle it:
1) Create a new configuration option in the `kubernetes` section, `kubernetes.throttleIngressRefresh`. It would default to 0, which would mean no throttling. Users can set it to a duration (like `10s`) to enable throttling.
2) Use the existing `providersThrottleDuration` from the common configuration, with the same `2s` default, so we match the upstream throttling behavior that gets applied in `server_configuration.go`. I *think* this will mean that the overall behavior is unchanged -- we'll end up just applying the throttle sooner to get the performance increase, and never hit the `server_configuration.go` throttle. This seems like the best option to me because it reduces the cognitive burden on end-users who are configuring Traefik, but I'm not sure how to get access to the common configuration section from within the Kubernetes provider -- it seems like it only gets the kubernetes section of the configuration. If we want to go down this route, i'd love some advice on how to access the common confiugration.
3) Like 2), but with an addition `kubernetes.throttleIngressRefresh` boolean option that default to false (meaning "don't apply the throttle", can be set to true apply the common `providersThrottleDuration`), to guarantee that we don't have any behavior changes unless users opt-in (and we could maybe remove this or default it to true in 2.0 if we don't get any issues reported about the behavior). Would still need to figure out how to get the `providersThrottleDuration` into the kubernetes provider.

For the ring channel package, I was cribbing from the code in [`server_configuration.go`](https://github.com/containous/traefik/blob/v1.7/server/server_configuration.go#L441), which uses the same package and roughly the same throttling implementation. Do you still think we should diverge from that? Looking at the [source code](https://github.com/eapache/channels/blob/master/ring_channel.go), it looks like the ring channel is doing some non-obvious performance optimizations, be we could also just copy [this simpler implementation](https://content.pivotal.io/blog/a-channel-based-ring-buffer-in-go) into the code and use that instead.

benweissmann:
@dtomcej  I've updated this PR to reflect your feedback -- a few notes:

- I've gone with approach #1 from my previous comment, which I think is the least risky. I still think we should consider approach #2 in the long run to reduce configuration burden on users, but maybe that's a better change for the 2.0 branch?
- I swapped out the use of the `channels` package for a simpler implementing using a buffered channel and a non-blocking write, because we don't care about getting the *latest* event; it's fine for us to get *any* event from the throttle interval.

provider/kubernetes/kubernetes.go
```diff
@@ -137,16 +139,56 @@ func (p *Provider) Provide(configurationChan chan<- types.ConfigMessage, pool *s
 					return nil
 				}
 			}
+
+			eventsChanToRead := eventsChan
+			throttleDuration := time.Duration(p.ThrottleDuration)
+			if throttleDuration > 0 {
```
mpl: I think you could have all of 145-174 in a

func throttleEvents(throttleDuration time.Duration) chan interface{} {
  if throttleDuration <= 0 {
    return nil
  }
  ....
  return eventsChanBuffered
}

then at 145 you could do:

throttledChan := throttleEvents()
if throttledChan != nil {
  eventsChanToRead = throttledChan
}

for less indented code and better readability.

Issue #3660, opened 2018-07-23
Forward authentication response headers


plaisted:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

It would be useful to allow headers returned from the forward authentication HTTP endpoint to be added to the final HTTP response. This would allow scenarios where the forward authentication endpoint can refresh authentication credentials and add the new credentials as a header/cookie.

This can be accomplished with a new configuration for http forward auth of "addHeadersToResponse" or similar.

Example:
```
addHeadersToResponse = ["Set-Cookie"]
```
This would result in any cookies added from HTTP forward auth response to be returned with the final response to the requesting client.

End-to-End example.
1. Traefik receives request to entrypoint with forward auth.
2. Traefik forward request to auth HTTP URI
3. JWT is used as a cookie for authentication. Forward auth server validates JWT, but expiration of JWT is soon.
4. Forward auth server requests new JWT for user and uses "Set-Cookie" header to overwrite previous JWT. Returns 20* response telling Traefik to continue routing.
5. Request is routed to correct frontend/backend and response obtained.
6. Forward auth middleware adds "Set-Cookie" header to outgoing response with new JWT.
7. New JWT cookie is used for subsequent requests.

ldez:
Duplicate of #3515

Fixed by #3521

plaisted:
@ldez I don't think this is a duplicate. That solution allows headers to be added to the request that is sent to the backend. The use case above requires the headers to be added to the outgoing response.

cptaffe:
I could really use this.

We use a Rails server for authentication, and use forward authentication to convert the cookie to a JWT for other services. This works great, but doesn't allow the cookie to get updated.

My Kubernetes Ingress configuration looks like this

```yml
kind: Ingress
apiVersion: extensions/v1beta1
metadata:
  name: example
  annotations:
    kubernetes.io/ingress.class: traefik
    ingress.kubernetes.io/auth-type: forward
    ingress.kubernetes.io/auth-url: http://example-rails.default/authenticate
    ingress.kubernetes.io/auth-response-headers: Authorization
    ingress.kubernetes.io/rewrite-target: /
spec:
  rules:
    - host: example.com
      http:
        paths:
          - path: /api
            backend:
              serviceName: example
              servicePort: 80
```

[...]

PR #4730, opened 2019-04-05
Add addHeadersToResponse option in forward auth middleware

ashutoshgngwr:
### What does this PR do?

Adds a config option in forward auth middleware to allow adding headers from auth server to request's response.

### Motivation

Traefik allows copying response headers from auth server to upstream request in forward authentication. This makes it difficult to perform tasks such as updating authentication headers in response.

#### Inspiration
_I was fooling around and all unit tests passed_, and also #3660 

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/middlewares.go
+++ b/pkg/config/middlewares.go
@@ -113,10 +113,11 @@ type ErrorPage struct {
 
 // ForwardAuth holds the http forward authentication configuration.
 type ForwardAuth struct {
-	Address             string     `description:"Authentication server address" json:"address,omitempty"`
-	TLS                 *ClientTLS `description:"Enable TLS support" json:"tls,omitempty" export:"true"`
-	TrustForwardHeader  bool       `description:"Trust X-Forwarded-* headers" json:"trustForwardHeader,omitempty" export:"true"`
-	AuthResponseHeaders []string   `description:"Headers to be forwarded from auth response" json:"authResponseHeaders,omitempty"`
+	Address              string     `description:"Authentication server address" json:"address,omitempty"`
+	TLS                  *ClientTLS `description:"Enable TLS support" json:"tls,omitempty" export:"true"`
+	TrustForwardHeader   bool       `description:"Trust X-Forwarded-* headers" json:"trustForwardHeader,omitempty" export:"true"`
+	AuthResponseHeaders  []string   `description:"Headers to be forwarded from auth response" json:"authResponseHeaders,omitempty"`
+	AddHeadersToResponse []string   `description:"Headers to be added to response from auth response" json:"addHeadersToResponse,omitempty"`
 }
 
 // +k8s:deepcopy-gen=true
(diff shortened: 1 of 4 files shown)


jwheeler-gs:
Can there be an option added to this to allow appending to headers instead of just overwriting like Envoy proxy supports? If the motivation is to be able to update authentication headers, this could cause problems if a lower layer is also trying to update headers (ie. via set-cookie) on the same response.

ashutoshgngwr:
@dduportal I was wondering about a few things myself as @jwheeler-gs pointed out. In my additions, I am directly setting header values (overwriting any previously set values, I am not sure if there are?). Should this be desired behaviour or should I consider appending header values. Also, once auth middleware sets a value for a header, will it be overwritten if any other (next-in-queue) middleware or the upstream request to backend tries to set it?

iamolegga:
@ashutoshgngwr hey!

> Edit: I just ran unit tests to check if response headers are overwritten by upstream request. They do and I need help because I think this is unacceptable behaviour (for headers such as Set-Cookie)!

Did you push this test to github, because I don't see any checks related to this problem. I would like to continue your work if you don't mind 🙂

ashutoshgngwr:
@iamolegga It was a long time ago and I was merely familiar with Golang syntax at the time. If you really want to work on this feature, I'd suggest a start from scratch. At the time, the issue was that in this implementation, the response headers from the actual request's response would get overwritten by those from the response of the forward auth. This only happened for headers that were present in both the responses. And no, I didn't push the unit tests.

PS, I haven't read through it in-depth but #3660 seems to propose this very feature.

iamolegga:
Ok, thanks!

iamolegga:
I've just dived to the problem with overwriting headers. So, what I think is we have 3 cases:

- `Set-Cookie` header is allowed to be set multiple times for different cookie names. [rfc6265](https://tools.ietf.org/html/rfc6265#page-7). So there should not be problem with that.
- `Set-Cookie` sent from `forwardAuth` service should be overwritten by target service in case they have same cookie names.
- For all the other headers values must be concatenated

Is this this the right behaviour?

ashutoshgngwr:
@iamolegga I am not sure about the correct behaviour but nothing should be overwritten with anything by default. We can leave the behavior to the user if we use `Header().Add` instead and move the adding-response-headers bit to after the next middleware is called.

https://github.com/ashutoshgngwr/traefik/blob/b9363180ac162c975e8a7f606e926e6421680274/pkg/middlewares/auth/forward.go#L155-L160

Try moving lines 155-157 to after 160. And using `Add` in-place of `Set` in 156.

iamolegga:
@ashutoshgngwr I thought the same, but in that case if both send `Set-Cookie` with same cookie name, value from auth header will rewrite value from target header, and I think this is wrong behaviour. I think that target service headers should be prioritised over auth service headers

ashutoshgngwr:
@iamolegga here are my thoughts on this.

Since user is specifying exactly which headers to copy, we should try to copy them after next middlewares have responded. We should also check if we will be overwriting anything. If we are in such a situation, we can print a warning/error to the user saying that we will not copy such header from auth response since it is already present in request's response.

Issue #4723, opened 2019-04-04
Make it possible to use static segment backend names with the Docker provider


mback2k:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

I expected the Docker container/service label "traefik.{segment-name}.backend" to behave the same way as "traefik.backend". Its raw value should have been used as the backend name instead of it being prefixed with the service name.

I understand that the default is to include the service name to avoid naming conflicts due to duplicates. But wouldn't this already be achieved by including the segment name inside the backend name?

The function used for "traefik.backend" is [getDefaultBackendName](https://github.com/containous/traefik/blob/ba44619828f9be7815d00a7d58e390bf4efc59db/provider/docker/config.go#L287) which either uses just the provided value or the service name.

The function used for "traefik.{segment-name}.backend" is [getSegmentBackendName](https://github.com/containous/traefik/blob/ba44619828f9be7815d00a7d58e390bf4efc59db/provider/docker/config.go#L278) which either uses the provided value prefixed with the service name or the segment name prefixed with the service name. So the segment name is already, but only used in case no value was given.

I think the behaviour of getSegmentBackendName is not consistent and causes trouble with random container names showing up, e.g. in Prometheus metrics. I would like to see it aligned with getDefaultBackendName by using the provided value or the service name, each suffixed with the segment name instead of always including the service name.

Basically this would mean to use the segment instead of the service name with the provided value, and make the service name the default value.

Examples of label to backend name conversion **at the moment**:

* traefik.backend: "my-single-backend" --> "my-single-backend"
* traefik.backend: unset --> "my-container/service-name"
* **traefik.my-segment.backend**: "my-single-backend" --> "my-container/service-name-my-single-backend"
* traefik.my-segment.backend: unset --> "my-container/service-name-my-segment"

Examples of label to backend name conversion **with this feature request implemented**:

* traefik.backend: "my-single-backend" --> "my-single-backend"
* traefik.backend: unset --> "my-container/service-name"
* **traefik.my-segment.backend**: "my-single-backend" --> "my-single-backend-my-segment"
* traefik.my-segment.backend: unset --> "my-container/service-name-my-segment"

This would make it possible to have static backend names by setting the .backend label. What do you think?

At the moment I am unable to really monitor the metrics produced by traefik from my docker services with prometheus and grafana due to random metric names.

ldez:
To understand the current behavior you need to read #3485, #3482, #3317, #3351

[...]

PR #4731, opened 2019-04-05
Make it possible to have consistent and constant backend names

mback2k:
### What does this PR do?

This PR is a first attempt to implement a possible solution/fix for #4723. This is a Work-in-Progress PR, tests still need to be adapted.

### Motivation

See #4723 for more information. Basically the idea is to make it possible to have consistent and constant backend names, especially in case segments are used.

At the moment the backend name for a segment would look like this for Docker and ECS:

`[service-name]-[segment-value-or-fallback-to-segment-name]`

Only the Racher provider currently uses the following layout:

`[service-name]-[segment-value]`
/
`[service-name]-[default-value-or-fallback-to-service-name]-[segment-name]`

This PR aligns the 3 providers and changes the layout to the following:

`[segment-value-or-fallback-to-service-name]-[segment-name]`

Therefore the segment name is always kept at the end of the backend name, and the segment-based value replaces the default value of the service name instead of the segment name.

An alternative solution could be to use a 3-part names for segment-based backends, like the Rancher provider did by default before aligning it within the first commit of this PR, e.g.:

`[service-name]-[default-value-or-fallback-to-service-name]-[segment-name]`

Since it does not make sense to include the service name twice, I would then instead propose the following:

`[default-value-or-fallback-to-service-name]-[segment-value-or-fallback-to-segment-name]`

### More

To be done:
- [ ] Added/updated tests

### Additional Notes

Please do not hesitate to ask me any questions about this. I am looking forward to feedback.

### DIFF
--- a/provider/docker/config.go
+++ b/provider/docker/config.go
@@ -278,7 +278,7 @@ func getBackendName(container dockerData) string {
 func getSegmentBackendName(container dockerData) string {
 	serviceName := getServiceName(container)
 	if value := label.GetStringValue(container.SegmentLabels, label.TraefikBackend, ""); len(value) > 0 {
-		return provider.Normalize(serviceName + "-" + value)
+		return provider.Normalize(value + "-" + container.SegmentName)
 	}
 
 	return provider.Normalize(serviceName + "-" + container.SegmentName)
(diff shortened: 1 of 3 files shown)


mback2k:
The alternative solution presented inside this PR is implemented with #4732. Please let me know which PR I should continue to work on.

ldez:
Thanks for your contribution.

I will close this PR to continue the discussion in #4723.

PR #4732, opened 2019-04-05
Use default backend name instead of service name in backend name

mback2k:
### What does this PR do?

This PR is an alternative attempt to #4731 in order to implement a possible solution/fix for #4723. This is a Work-in-Progress PR, tests still need to be adapted.

### Motivation

Please see original/primary PR #4731 and issue #4723. This PR implements the alternative solutions already explained in the original PR:

> An alternative solution could be to use a 3-part names for segment-based backends, like the Rancher provider did by default before aligning it within the first commit of this PR, e.g.:
> 
> `[service-name]-[default-value-or-fallback-to-service-name]-[segment-name]`
> 
> Since it does not make sense to include the service name twice, I would then instead propose the following:
> 
> `[default-value-or-fallback-to-service-name]-[segment-value-or-fallback-to-segment-name]`

### More

To be done:
- [ ] Added/updated tests

### Additional Notes

Please let me know which PR I should continue to work on.

### DIFF
--- a/provider/docker/config.go
+++ b/provider/docker/config.go
@@ -278,10 +278,10 @@ func getBackendName(container dockerData) string {
 func getSegmentBackendName(container dockerData) string {
 	serviceName := getServiceName(container)
 	if value := label.GetStringValue(container.SegmentLabels, label.TraefikBackend, ""); len(value) > 0 {
-		return provider.Normalize(serviceName + "-" + value)
+		return provider.Normalize(getDefaultBackendName(container) + "-" + value)
 	}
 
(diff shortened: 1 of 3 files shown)


ldez:
Thanks for your contribution.

I will close this PR to continue the discussion in #4723.

tuwaertuwa:
#4732

tuwaertuwa:
#4731

tuwaertuwa:
default-value-or-fallback-to-service-name]-[segment-value-or-fallback-to-segment-name]

Issue #4085, opened 2018-10-19
PathPrefixStrip rule does not work with SSL redirection


sagikazarmark:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I specified an ingress for my service in Kubernetes:

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: traefik
    traefik.frontend.rule.type: PathPrefixStrip
    traefik.ingress.kubernetes.io/ssl-redirect: "true"
  name: myingress
  namespace: default
spec:
  rules:
  - host: myhost
    http:
      paths:
      - backend:
          serviceName: monitoring-grafana
          servicePort: 80
        path: /mypath
```

Inspecting the service on the dashboard shows it gets created properly with the correct rule.

### What did you expect to see?

I expected `http://myhost/mypath` to be redirected to `https://myhost/mypath`

### What did you see instead?

Traefik redirected the request to `https://myhost`

Removing the `PathPrefixStrip` properly redirects the request.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
1.7.2-alpine
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
    logLevel = "INFO"
    InsecureSkipVerify = true
    defaultEntryPoints = ["http","https"]
    [entryPoints]
      [entryPoints.http]
      address = ":80"
      compress = true
      [entryPoints.https]
      address = ":443"
      compress = true
        [entryPoints.https.tls]
          [[entryPoints.https.tls.certificates]]
          CertFile = "/ssl/tls.crt"
          KeyFile = "/ssl/tls.key"
      [entryPoints.prometheus]
      address = ":9120"
      compress = true
      [entryPoints.ping]
      address = ":81"
      [entryPoints.dashboard]
      address = ":8080"
      compress = true
    [ping]
    entryPoint = "ping"
    [kubernetes]
    [kubernetes.ingressEndpoint]
      publishedService = "kube-system/ingress-traefik"
    [acme]
    email = "my@email.com"
    storage = "/acme/acme.json"
    entryPoint = "https"
    onHostRule = true
    acmeLogging = true
      [acme.dnsChallenge]
      provider = "route53"
    [api]
    dashboard = true
    entryPoint = "dashboard"
    # Metrics definition
    [metrics]
      [metrics.prometheus]
        entryPoint = "prometheus"
```

### Additional context

[...]

PR #4748, opened 2019-04-08
Use context-aware secure middleware functions

dtomcej:
### What does this PR do?

Adds a context-aware function to the secure middleware so that redirections after a path has been modified will correctly redirect

### Motivation

#4085

### More

- [x] Added/updated tests
- [x] Added/updated documentation - None needed, correcting behavior

### Additional Notes

Contexts suck

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1564,11 +1564,11 @@
 
 [[projects]]
   branch = "v1"
-  digest = "1:60888cead16f066c948c078258b27f2885dce91cb6aadacf545b62a1ae1d08cb"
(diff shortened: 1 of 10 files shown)


juliens:
I think, we don't need this to handle the problem (in v2), because the only thing we have to do is to change the order of the middlewares. 

If the `headers` middleware with `sslRedirect=true` is put before the middleware that change the path, you should not have any problem.

WDYT @dtomcej ?

dtomcej:
@juliens, Correct.

The only issue with that is described in https://github.com/containous/traefik/issues/5538, which is under active development.

Therefore, will be closing this PR.

Issue #4777, opened 2019-04-16
dumpcerts.sh compatibility


mlan:
The dumpcerts.sh script uses OpenSSL to verify that the private key is valid. 
In doing so, the key is piped to the command openssl, without limiting the line lengths. 
This appears not to work for all implementations of SSL.

To exemplify:
Within a Docker image based on alpine:3.8 the key line length need to be less than 80 characters.

```
# cat test.pem | fold -w 79 | openssl rsa -inform pem -out key.pem
writing RSA key
```

```
# cat test.pem | fold -w 80 | openssl rsa -inform pem -out key.pem
unable to load Private Key
140046610254796:error:09FFF064:PEM routines:CRYPTO_internal:bad base64 decode:pem/pem_lib.c:800:
```

This limitation does not appear on other systems. The reason for this compatibility issue might be that 
Alpine Linux use an openssl implementation based on LibreSSL:

```
# openssl version
LibreSSL 2.7.4
```

ldez:
Hello,

We want to remove this script from the Traefik repository.
You can find more explanation and alternatives here: #4783

PR #4778, opened 2019-04-16
dumpcerts.sh compatibility

mlan:
The dumpcerts.sh script uses OpenSSL to verify that the private key is valid.

In doing so, the key is piped to the command openssl, without limiting the line lengths.
This appears not to work for all implementations of SSL.
Within a Docker image based on alpine:3.8 the key line length need to be less than 80 charaters.

See issue:   #4777


ldez:
Thank you for your contribution :+1: 

We want to remove this script from the Traefik repository (see #4783).

Sorry, but I will close your PR.

Issue #4812, opened 2019-04-24
Path restrictions


sylr:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

It would be nice to be able to enforce global path restrictions which would applied to all frontends.

The use case for this is to protect internal endpoints such as `/metrics` (prometheus scrape endpoint), `/ping` & `/ready` (kubernetes probes endpoints) ... etc so that they wouldn't be publicly available. 

The must would be the possibility to choose what kind of error traefik would return (404, 503 or else) and also be able to whitelist ip ranges that could access those paths.

treezio:
really interested on this !

ahoulgrave:
FYI it's better to expose one external and one internal port on your services,  so you expose for example `:8081/metrics` and then `:8080/public-uri`. Your ingress or rule should target the public port. So i'm not really sure this feature should be implemented.

treezio:
In some cases its not that easy to expose those metrics in a different port 
(ex: NodeJS express server prometheus metrics will be exposed in the default port and its quite tricky to changes that port.)

sylr:
Agree with @treezio , you can't rely that the soft you want to proxy through Traefik use a specific port for internal endpoints (most of the time they don't).

ahoulgrave:
Ok, sure, it's true this a needed feature.
Still, I think you should expose two different ports whenever it's possible

treezio:
Although you are definitely right @ahoulgrave , I think this is a quite cool feature.

koalalorenzo:
I would love this feature!

ericjee:
Hi guys, any updates?

jhaumont:
Same like @koalalorenzo, I want to drop nginx ! Any news for a schedule of this feature ?

SantoDE:
You could come close with v2 here I guess? How about defining a middleware with your desired actions and attach that to all routers needed?

scwall:
Hello I am very interested in this feature too.

fwkz:
@juliens Is there any update on this?

canidam:
Would love to see this feature in traefik. I need it for a different use case, to block some paths on my app. Really need this feature

notsureifkevin:
[...]

PR #4814, opened 2019-04-24
Create BlockPathRegex Middleware

### LINKED ISSUE #4813: Create BlockPathRegex Middleware

Feature

Since golang and goregexp does not support lookaheads and readbehinds, issues come into play when we want to expose all except a specific path. For example, `PathPrefix: /`, but NOT `/bacon`.

There should be configurable response codes for the block match.

By having this coded in middlewares, it could be re-used, and shared between providers.

dtomcej:
### What does this PR do?

Creates a regex middleware that blocks requests, and returns a configured response code and message.

### Motivation

Fixes #4813, #4812 

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/middlewares.go
+++ b/pkg/config/middlewares.go
@@ -327,6 +327,15 @@ type ReplacePathRegex struct {
 
 // +k8s:deepcopy-gen=true
 
+// BlockPathRegex holds the BlockPathRegex configuration.
+type BlockPathRegex struct {
+	Regex        string `json:"regex,omitempty"`
+	ResponseCode int    `json:"responseCode,omitempty"`
+	Message      string `json:"message,omitempty"`
+}
+
+// +k8s:deepcopy-gen=true
+
 // Retry holds the retry configuration.
 type Retry struct {
 	Attempts int `description:"Number of attempts" export:"true"`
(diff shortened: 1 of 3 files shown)


pkg/config/middlewares.go
```diff
@@ -327,6 +327,15 @@ type ReplacePathRegex struct {
 
 // +k8s:deepcopy-gen=true
 
+// BlockPathRegex holds the BlockPathRegex configuration.
+type BlockPathRegex struct {
+	Regex        string `json:"regex,omitempty"`
+	ResponseCode int    `json:"responseCode,omitempty"`
+	Message      string `json:"message,omitempty"`
+}
```
sylr: Shouldn't `BlockPathRegex` be added to the `Middleware` struct ?

sylr:
@dtomcej are you still working on this ?

ldez:
This PR in in **status/1-needs-design-review**.

https://docs.traefik.io/contributing/maintainers/#pr-review-process

sylr:
@ldez it has been in this state for 7 months now. Can someone do the design review ?

ldez:
We consider that this middleware is a routing constraint, and we think that this kind must be able by the routing itself and not a middleware.

So I close for now.

PR #4000, opened 2018-10-10
Add new modifier: ReplaceQueryRegex

cention-nazri:
### What does this PR do?

Allows manipulation of query parameters using regular expression

A new modifier `ReplaceQueryRegex` matches against the path and query parameters ("origin-form" in rfc7230 section 5.3.1 (`net/http.Request.RequestURI`)). The result of the regular expression replacement executed against the incoming `net/http.Request.RequestURI` replaces the existing query parameters  in the request.

### Motivation

It was not possible to modify the query parameters (or add new ones) to backends.

### More

- [x] Added/updated tests
- [x] Added/updated documentation

### Additional Notes

This pull request obsoletes https://github.com/containous/traefik/pull/3928. The changes here are in response to the review comments there.

PR #4820, opened 2019-04-25
Add Replace Query Regex Middlewares

dtomcej:
### What does this PR do?

Adds 2 new middlewares:
- ReplaceQueryRegex: To modify query strings
- ReplacePathQueryRegex: To modify paths and query strings

### Motivation

This PR carries on the work started with #4000

### More

- [x] Added/updated tests

### Additional notes

Co-authored by @holygeek / @cention-nazri


ldez:
I am troubled with `ReplacePathQueryRegex` because we also have `ReplacePathRegex`.

I wonder if we could merge it with `ReplacePathRegex` :thinking:

jbdoumenjou:
I agree with @ldez, a merge between these middlewares seems to be more appropriate.

dtomcej:
>I wonder if we could merge it with ReplacePathRegex 🤔

Yes. I will make this happen. It makes sense to me too.

Will aim for:
- `ReplacePathRegex` => Path and query modifier
and
- `ReplaceQueryRegex` => Query only modifier

dtomcej:
@ldez @jbdoumenjou I looked into updating this PR, but I ran into the design reason for the difference between `ReplacePathRegex` and `ReplacePathQueryRegex`:

`ReplacePathRegex` modifies the path of the request, but `ReplacePathQueryRegex` only allows for the regex to match and use parts of the path to modify queries. `ReplacePathQueryRegex` does not allow for the path to be modified in this manner.

`ReplacePathRegex` operates on `*http.Request.URL.Path`, and `ReplacePathQueryRegex` operates on `http.Request.RequestURI`, and I'm not sure there is a good way to merge the two without incorporating breaking/design changes.

Thoughts?

jbdoumenjou:
Sorry for the delay @dtomcej .
So, as we don't use the same data to apply those middlewares, I agree that it may be confusing to merge them.  Perhaps we need to define two better names to avoid any confusion. WDYT ?

dtomcej:
>Perhaps we need to define two better names to avoid any confusion

Yessir. I think this would help out immensely.

jbdoumenjou:
We can have only one middleware with an option like "src/target/criterion/..." to define the request part used. But it seems to be no compliant with the "spirit" of the middleware responsibility.
So, we can factorize the code between this two middlewares and try to find a better name (but I didn't managed to do it). It is just some non blocking comment, so I will accept the PR.

freddydk:
Was the ability of changing the Query part of the URL as part of the modifiers ever implemented?
Thanks

ldez:
Some notes:
- `ReplacePathQueryRegex`: use the path and the query to modify **the query string only**, there is no modification of the path.
- I continue to think that `replacePathRegex`, `replaceQueryRegex`, `replacePathQueryRegex` can be a single middleware.

I will continue to rewrite the PR, maybe I will open an another PR.

ldez:
After discussion with the team, we think that the approach is too naive which will produce a high complexity of use.

So we will decline this PR but we keep the idea and for that I created an issue on this subject (#6276).

If you want to participate in the reflection on the subject or explain your use cases, I invite you to express yourself in this issue (#6276).

Thank you for your understanding.

PR #4844, opened 2019-05-03
Fix `make validate` on MacOS outside container

dduportal:
### What does this PR do?

Fix a shell script which fails on macOS and BSD, 
allowing the command `PRE_TARGET= make validate` to work fine on macOS.

### Motivation

Because some of us have damned souls using macOS :trollface: 

### More

- ~[ ] Added/updated tests~

### Additional Notes

This fix is from @mpl :)

### DIFF
--- a/script/test-integration
+++ b/script/test-integration
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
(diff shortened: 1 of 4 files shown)


mpl:
For the record, I only found the issue, @dduportal is the one who found the fix.

PR #4862, opened 2019-05-09
Suggest to choose Deployment over DaemonSets as first choice

Federkun:
### What does this PR do?

I'm assuming the above was a typo - k8s's Deployment should be the suggested choice when you are not sure.

### More

- [ ] Added/updated tests


dtomcej:
@Federkun We _used_ to suggest deployments over daemonsets, but changed the recommendation for the following reasons:

1. It is difficult to preserve source IPs through a deployment, due to the need for exposing via a service
2. Ingress controllers are a special application that rarely can take advantage of the flexibility deployments really offer
3. Other ingress controllers (such as nginx and haproxy) install as daemonsets by default, and sticking to the empirical norm was less confusing to our users.

For these reasons, we switched our recommendation from deployments to daemonsets by default. I agree that for most applications, deployments should be the easy choice. Our more advanced users can use deployments if they can take advantage of its scalability and other features, but in this case, for beginner users, we found users had less issues with daemonsets.

Federkun:
Cool, thanks for explaining that to me! Could I suggest to swap the order of `The Deployment objects looks like this:` section with `The DaemonSet objects looks not much different:`? Having `DaemonSet` as the first example would encourage me to use it instead of the alternative

dtomcej:
Hey @Federkun,

Just checking in on this PR. Did you want to update it? or did you want to close this for the time being?

dduportal:
I would suggest that we do NOT suggest any of these 2 patterns. Instead, we should propose a "Pros & Cons" section, to help user to decide which one.

There are 2 reasons to prefer Deployments (with 1 replica): 

- To be able to use Let's Encrypt/ACME
- To benefit from orchestrator rescheduling capability when a node is crashing or a drain operation is running

SantoDE:
Hey @Federkun,

what do you think about @dduportal suggestion? :)

ldez:
Due to age, I will close this PR, please open an issue to speak about the topic If you want to continue the discussion.

PR #4877, opened 2019-05-15
Enhance KV logs.

ldez:
### What does this PR do?

Give more context when an error occurs.

### Motivation

Enhance KV logs.

### More

- [ ] ~Added/updated tests~


salzig:
I would say this is not only an enhancement, but also a bug fix (Line 804)

ldez:
The change to L804 is related to a L642 and it was not needed before this change.

ldez:
The change to L804 is related to the change L642, it was not needed before.

PR #4923, opened 2019-05-30
Specify that Rancher provider is for 1.x only

bradjones1:
### What does this PR do?

Updates the Rancher documentation to specify it is specific to Rancher 1.x.

### Motivation

Rancher 2.x is basically a super-duper reverse proxy and control plane for Kubernetes. 

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/providers/rancher.md
+++ b/docs/content/providers/rancher.md
@@ -8,7 +8,9 @@ A Story of Labels, Services & Containers
 Attach labels to your services and let Traefik do the rest!
 
 !!! important
-    This provider is specific to Rancher 1.x. Rancher 2.x requires Kubernetes and does not have a metadata endpoint of its own for Traefik to query. As such, Rancher 2.x users should utilize the [Kubernetes provider](./kubernetes-crd.md) directly.
+    This provider is specific to Rancher 1.x.
+    Rancher 2.x requires Kubernetes and does not have a metadata endpoint of its own for Traefik to query.
+    As such, Rancher 2.x users should utilize the [Kubernetes provider](./kubernetes-crd.md) directly.
 
 ## Configuration Examples
 
--- a/docs/content/providers/rancher.md
+++ b/docs/content/providers/rancher.md
@@ -1,4 +1,4 @@
-# Traefik & Rancher 1.x
+# Traefik & Rancher
 
 A Story of Labels, Services & Containers
 {: .subtitle }
--- a/docs/content/providers/rancher.md
+++ b/docs/content/providers/rancher.md
@@ -7,7 +7,7 @@ A Story of Labels, Services & Containers
 
 Attach labels to your services and let Traefik do the rest!
 
-Note: This provider is specific to Rancher 1.x. Rancher 2.x requires Kubernetes and does not have a metadata endpoint of its own for Traefik to query. As such, Rancher 2.x users should utilize the [Kubernetes provider](../kubernetes-crd/index.md) directly.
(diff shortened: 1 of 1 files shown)


dduportal:
Hi @bradjones1 , thanks for this contribution.

Could you fix the issues raised by the CI (Travis) related to your change please?

```text
=== Checking HTML content...
- /app/site/providers/rancher/index.html
  *  internally linking to ../kubernetes-crd/index.md, which does not exist (line 1577)
     <a href="../kubernetes-crd/index.md">Kubernetes provider</a>
```

(you can reproduce it locally before pushing your change by following instructions from the page [Contributing -> Documentation](https://docs.traefik.io/v2.0/contributing/documentation/#check-the-documentation), with the target `make -C ./docs/ docs-verify`.

docs/content/providers/rancher.md
```diff
@@ -1,12 +1,14 @@
-# Traefik & Rancher
+# Traefik & Rancher 1.x
```
dduportal: Could you let the title as it was? As Rancher 2.x is mentioned below thanks to your change, the reader should not be "filtered" at the title, but at the note instead.

docs/content/providers/rancher.md
```diff
@@ -1,12 +1,14 @@
...
 
-A Story of Labels, Services & Container
+A Story of Labels, Services & Containers
 {: .subtitle }
 
 ![Rancher](../assets/img/providers/rancher.png)
 
 Attach labels to your services and let Traefik do the rest!
 
+Note: This provider is specific to Rancher 1.x. Rancher 2.x requires Kubernetes and does not have a metadata endpoint of its own for Traefik to query. As such, Rancher 2.x users should utilize the [Kubernetes provider](../kubernetes-crd/index.md) directly.
```
dduportal:
```suggestion
Note: This provider is specific to Rancher 1.x. Rancher 2.x requires Kubernetes and does not have a metadata endpoint of its own for Traefik to query. As such, Rancher 2.x users should utilize the [Kubernetes provider](./kubernetes-crd.md) directly.
```

dduportal: Reason: the links shall follow the file structure

```shell
$ pwd
${GOPATH}/src/github.com/containous/traefik/docs/content/providers
$ ls -l ./kubernetes-crd.md 
-rw-r--r--  1 user  user  6856 May 31 10:55 ./kubernetes-crd.md
```

docs/content/providers/rancher.md
bradjones1: I don't feel strongly either way but I think actually filtering out at the heading level makes some sense; If you are on Rancher 2.x you won't need to read through the detail below to know it's not the right place to be.

docs/content/providers/rancher.md
```diff
@@ -1,12 +1,14 @@
...
 
-A Story of Labels, Services & Container
+A Story of Labels, Services & Containers
 {: .subtitle }
 
 ![Rancher](../assets/img/providers/rancher.png)
 
 Attach labels to your services and let Traefik do the rest!
 
+Note: This provider is specific to Rancher 1.x. Rancher 2.x requires Kubernetes and does not have a metadata endpoint of its own for Traefik to query. As such, Rancher 2.x users should utilize the [Kubernetes provider](./kubernetes-crd.md) directly.
```
ldez:
```suggestion
!!! important
    This provider is specific to Rancher 1.x. Rancher 2.x requires Kubernetes and does not have a metadata endpoint of its own for Traefik to query. As such, Rancher 2.x users should utilize the [Kubernetes provider](./kubernetes-crd.md) directly.
```

docs/content/providers/rancher.md
ldez: I agree with @dduportal

bradjones1: 2 to 1 - resolved. :-)

Issue #4926, opened 2019-06-01
Traefik not closing tcp connections, server getting unresponsive after a bit of time


kolaente:
### Do you want to request a *feature* or report a *bug*?

Bug

### Did you try using a 1.7.x configuration for the version 2.0?

- [ ] Yes
- [x] No

### What did you do?

I deployed my (rather simple) go application using docker and docker-compose with some monitoring (using prometheus, not related to traefik) and traefik (also in docker) by pretty much following [this](https://docs.traefik.io/user-guide/docker-and-lets-encrypt/) guide. All in all not much, 7 containers in total (with traefik). I used [the cheapest Hetzner cloud vps](https://www.hetzner.de/cloud) which however should have more than enough ressources for my use case.

I can provide the compose files if needed.

### What did you expect to see?

Application working normally, no issues.

### What did you see instead?

After a while, the server gets totally unresponsive. Not even ssh did work anymore. When I restarted it via Hetzners web console, it started working again, but after a few days I had the same issue again. After some digging, I found this to be an error with "too many orphaned tcp connections" and thought my application was to blame, found an issue with that, fixed it, redeployed it and thought it was all good. But then, the same problem started appearing again. Being a bit short on time, I just raised all limits which got it somehow working again.

Then I finally did some more looking for the error and found this:
```
~# ss -sa
Total: 99822 (kernel 99836)
TCP:   99753 (estab 1, closed 99748, orphaned 1, synrecv 0, timewait 225/0), ports 0

Transport Total     IP        IPv6
*	  99836     -         -        
RAW	  0         0         0        
UDP	  1         1         0        
TCP	  5         2         3        
INET	  6         3         3        
FRAG	  0         0         0 
```
Notice the _99748_ closed tcp connections which don't show up in the further stats. These didn't show up in `netstat` either, which I thought is a bit weird. Then I started shutting down parts of my application, one after another, then the monitoring, checking each time if the number of closed connections had gone down (spoiler: it didn't much) until there was only traefik left. When I shut it down, I noticed this:

```
/opt/compose/traefik# docker-compose down
Stopping traefik ... done
Removing traefik ... done
Network web is external, skipping
/opt/compose/traefik# ss -sa
Total: 127 (kernel 156)
TCP:   3904 (estab 1, closed 3901, orphaned 3667, synrecv 0, timewait 234/0), ports 0

Transport Total     IP        IPv6
*	  156       -         -        
RAW	  0         0         0        
UDP	  1         1         0        
[...]

PR #4927, opened 2019-06-02
Use the latest stable version of traefik in the docs

kolaente:
### What does this PR do?

This pr updates version of the traefik docker container in the example docker-compose file in the user guide for using traefik with docker and LE.

### Motivation

 I've had some issues caused by me using an outdated version of traefik (see #4926) which was mostly caused by just copying the docker compose file. From [docker hub](https://hub.docker.com/_/traefik) it looks like `latest` ist the latest stable image, so I propose to just use that in the examples.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/user-guide/cluster-docker-consul.md
+++ b/docs/user-guide/cluster-docker-consul.md
@@ -91,7 +91,7 @@ To watch docker events, add `--docker.watch`.
 version: "3"
 services:
   traefik:
-    image: traefik:1.5
+    image: traefik:<stable version from https://hub.docker.com/_/traefik>
     command:
       - "--api"
       - "--entrypoints=Name:http Address::80 Redirect.EntryPoint:https"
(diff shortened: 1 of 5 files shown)


dduportal:
Hi @kolaente , you underline a good usage element here: "Code snippet from the documentation should be up-to-date".

However, they are a few drawbacks here:

1. Using the tag `: latest` for Docker images: `Friends don't let friends use :latest`. Using this tag would be seen as the recommended way, which is not. We strongly advise such usage, even in the context of a documentation. To solve this challenge, some "release process plumbing" should be done to propagate the latest version across all the documentation, which might be studied after Traefik `2.0` release. WDYT about setting the image/tag to `traefik:<stable version from https://hub.docker.com/_/traefik>` ? So end user would be warned, solving your initial intent, without putting users at risk?

2. Such a change should be done on all exemples of the section `User Guide` to avoid more configuration drift between pages, WDYT?

kolaente:
Sounds like a reasonable idea, I'll update as suggested!

kolaente:
Thanks!

PR #4933, opened 2019-06-04
rework loadbalancer support

juliens:
### What does this PR do?

Remove wrr and drr LoadBalancers to prepare configuration to accept new load balancers.

### Motivation

We plan on adding more load balancer strategies later on. In order to do that we'll need some sort of generic method that may turn out to be incompatible with the current way wrr and drr are implemented. Therefore, if we do not want to break configuration later on when we add these new strategies, we have to remove wrr and drr now.

### More

- [x] Added/updated tests


integration/fixtures/healthcheck/simple.toml
```diff
@@ -27,7 +27,5 @@ level = "DEBUG"
       timeout = "0.9s"
     [[http.services.service1.loadbalancer.servers]]
       url = "http://{{.Server1}}:80"
-      weight = 1
-    [[http.services.service1.loadbalancer.servers]]
+          [[http.services.service1.loadbalancer.servers]]
```
jbdoumenjou:
```suggestion
    [[http.services.service1.loadbalancer.servers]]
```

integration/fixtures/multiprovider.toml
```diff
@@ -22,7 +22,6 @@ level = "DEBUG"
 
       [[http.services.service.LoadBalancer.Servers]]
         URL = "{{.Server}}"
-        Weight = 1
-[http.middlewares]
+        [http.middlewares]
```
jbdoumenjou:
```suggestion
[http.middlewares]
```

integration/fixtures/tcp/mixed.toml
```diff
@@ -30,8 +30,7 @@ level = "DEBUG"
           [http.services.whoami.loadbalancer]
             [[http.services.whoami.loadbalancer.servers]]
               url = "http://localhost:8085"
-              weight=1
-[tcp]
+    [tcp]
```
jbdoumenjou:
```suggestion
[tcp]
```

Issue #4740, opened 2019-04-07
sendAnonymousUsage has inverse meaning in v2.0.0-alpha3


powerman:
### Do you want to request a *feature* or report a *bug*?

Bug

### Did you try using a 1.7.x configuration for the version 2.0?

- [ ] Yes
- [x] No

### What did you do?

`--global.sendAnonymousUsage=false`

### What did you expect to see?

```
level=info msg="\nStats collection is disabled.\nHelp us improve Traefik by turning this feature on :)\nMore details on: https://docs.traefik.io/basics/#collected-data\n"
```

### What did you see instead?

```
level=error msg="\nYou haven't specify the sendAnonymousUsage option, it will be enable by default.\n"
level=info msg="\nStats collection is enabled.\nMany thanks for contributing to Traefik's improvement by allowing us to receive anonymous information from your configuration.\nHelp us improve Traefik by leaving this feature on :)\nMore details on: https://docs.traefik.io/basics/#collected-data\n"
```

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      2.0.0-alpha3
Codename:     faisselle
Go version:   go1.12.1
Built:        2019-03-29T17:53:25Z
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

No config file, only arguments:
```yaml
    traefik:
        image: traefik:2.0
        command: >
            --api
            --entrypoints='Name:web Address::8000'
            --metrics
            --metrics.prometheus
            --providers
            --providers.docker
            --providers.docker.exposedbydefault=false
            --providers.docker.watch
            --global.sendAnonymousUsage=false
            --global.debug
            --log.loglevel=debug
            --accesslog
```

### Workaround:

With `--global.sendAnonymousUsage=true` I get in logs:
```
level=info msg="\nStats collection is disabled.\nHelp us improve Traefik by turning this feature on :)\nMore details on: https://docs.traefik.io/basics/#collected-data\n"
```
But I'm not sure is actual stats enabled or disabled - something is inversed, but is it only log message or both log message and sending stats?

jbdoumenjou:
Hi @powerman, thank you for pointing out this issue.
This behavior is due to the way the command line is parsed.
We will fix this behavior as soon as possible.
Until this fix, a workaround is to use the toml file configuration which works as expected.

traefiker:
Closed by #4935.

Issue #4653, opened 2019-03-22
CLI help default value wrong sometimes


SantoDE:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

traefik --help

### What did you expect to see?

Correct default values

### What did you see instead?

providers.file is shown with a default value of true, but in fact, it's false

### Output of `traefik version`: (_What version of Traefik are you using?_)

At least v2. Not sure about v1

traefiker:
Closed by #4935.

Issue #4645, opened 2019-03-21
Configuration files with dash are ignored


tkzv:
If configuration file is set in the command line with "-c" option and the file name contains a dash (hyphen, minus sign) "-", Traefik will use the default configuration. It fails silently, with no messages that the file was ignored; it's only noticeable by its behaviour.

For example,
```
./traefic -c config.toml 
```
correctly loads the configuration in config.toml, but 
```
./traefic -c config-2.toml 
```
uses default settings and ignores the configuration file.

Observed in the latest 2.0.0 alpha in Windows and Linux:
```
$ ./traefik version
Version:      2.0.0-alpha1
Codename:     faisselle
Go version:   go1.12.1
Built:        2019-03-18T14:26:35Z
OS/Arch:      linux/amd64

>traefik.exe version
2019/03/21 13:51:55 Using high precision timer
Version:      2.0.0-alpha1
Codename:     faisselle
Go version:   go1.12.1
Built:        2019-03-18T14:40:53Z
OS/Arch:      windows/amd64
```

Example configuration:
```toml
[global]
sendanonymoususage = false
```
If the configuration file is called "config-2.toml", there will be a message "You haven't specify the sendAnonymousUsage option, it will be enable by default."

Workaround:
```
./traefik --configfile=config-2.toml
```
works correctly

ldez:
hello, when you have a value that contains a hyphen, you have to add quotes around the value.

```
./traefic -c "config-2.toml"
```

>  It fails silently, 

related to https://github.com/containous/traefik/issues/2927

PS: could you follow the issue template.

tkzv:
> PS: could you follow the issue template.
Do you want me to rewrite the bug with the template?

ldez:
Sorry it's:
```
./traefic --config="config-2.toml"
```

tkzv:
I wrote above that --config= works even without quotes. The bug is that -c does not work with dashes. Regardless of the quotes.

ldez:
yes short flag doesn't works with dash, it's a standard behavior.

tkzv:
Why? Never seen anything like that.

traefiker:
Closed by #4935.

Issue #3985, opened 2018-10-06
Allowing space in CLI start options


tcoupin:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I trying to define an entrypoint with a custom realm "Hello World".
With a traefik.toml file, it works:
```
[entryPoints.http]
address = ":8080"
[entryPoints.http.Auth.Basic]
realm = "Hello World"
users = ["test:$apr1$H6uskkkW$IgXLP6ewTrSuBkTrqE8wj/"]
```

But with CLI options, it does not work:

`./dist/traefik -d --entryPoints='Name:http Address::8080 Auth.Basic.Users:test:$apr1$H6uskkkW$IgXLP6ewTrSuBkTrqE8wj/ Auth.Basic.Realm:"Hello World"'`

I try using double quotes to define realm since space is the option separator (https://docs.traefik.io/master/configuration/entrypoints/#cli)

### What did you expect to see?

```
curl -v 127.0.0.1:8080
* Rebuilt URL to: 127.0.0.1:8080/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 401 Unauthorized
< Content-Type: text/plain
< Www-Authenticate: Basic realm="Hello World"
< Date: Sat, 06 Oct 2018 17:50:47 GMT
< Content-Length: 17
< 
401 Unauthorized
* Connection #0 to host 127.0.0.1 left intact
```

### What did you see instead?

```
curl -v 127.0.0.1:8080
* Rebuilt URL to: 127.0.0.1:8080/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to 127.0.0.1 (127.0.0.1) port 8080 (#0)
> GET / HTTP/1.1
> Host: 127.0.0.1:8080
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 401 Unauthorized
< Content-Type: text/plain
< Www-Authenticate: Basic realm=""Hello"
< Date: Sat, 06 Oct 2018 17:50:47 GMT
< Content-Length: 17
< 
401 Unauthorized
* Connection #0 to host 127.0.0.1 left intact
```

### Output of `traefik version`: (_What version of Traefik are you using?_)

Branch master

```
[...]

PR #4935, opened 2019-06-05
New static configuration loading system.

### LINKED ISSUE #7: Configuration through environment
Any plans to do configuration through environment? I'd love the ability to configure with something like:

``` shell
docker run \
  -e DOCKER_ENDPOINT="unix:///tmp/docker.sock" \
  -e DOCKER_DOMAIN="docker.localhost" \
  -e DOCKER_WATCH=true \
  -v /var/run/docker.sock:/tmp/docker.sock \
  emilevauge/traefik
```

ldez:
### What does this PR do?

New static configuration loading system that allows:

- commands and sub-commands
- flags (map and slice of struct)
- files
- environment variables

Removes Flaeg and Staert :tada: 

### Motivation

Being able to handle sub-commands as well as maps and slices of structs, and environment variables.

Since Flaeg and Staert didn't allow implementing easily the above, and given it was hard to maintain, we took the opportunity to fully replace it. :tada: 

Closes #7, #4740, #4653, #3985, #4645

### More

- [x] Added/updated tests

### Co-authors


ldez:
For the help, with the configuration of the v2 only display the top level is an issue:
- some options (like `acme` or `providers.docker`) cannot be enable with the root element
- to display sub-element you need to have some sub flags for flags or create a complex system to understand what element to display.

We plan to create a system to "automatically" create and update the documentation (the md files) for flags and env vars.

ldez:
There are 2 slices of struct: `domains` and `constraints`.

For the `domains` (acme), they will be remove from the static configuration, look at #4872

For the `constraints`, they will be replace by a slice of string like in the v1.

emilevauge:
OK for the slices topic.
About the help, I understand your points, but I really think we have to do something on this. The current implementation has become unreadable.
I was thinking of :
```
$ traefik --help
--accesslog  (Default: "false")
    Access log settings.
--acme.*
    ACME configuration
--api  (Default: "false")
    Enable api/dashboard.
--hostresolver  (Default: "false")
    Enable CNAME Flattening.

$ traefik --accesslog --help
--accesslog  (Default: "false")
    Access log settings.

--accesslog.bufferingsize  (Default: "0")
    Number of access log lines to process in a buffered way.

--accesslog.fields.defaultmode  (Default: "keep")
    Default mode for fields: keep | drop

--accesslog.fields.headers.defaultmode  (Default: "keep")
    Default mode for fields: keep | drop | redact
```

The `--acme` option is an exception indeed, but we can find a workaround.

mpl:
> OK for the slices topic.
> About the help, I understand your points, but I really think we have to do something on this. The current implementation has become unreadable.

I do not understand why you say it has _become_ unreadable. The number of items displayed is not larger than it was before. So do you mean that:
1) it was already unreadable before this PR, because there was/are too many options ? (fair enough)
2) or you find the new format not as readable as the old format? if yes, why?

> I was thinking of :
> 
> ```
> $ traefik --help
> --accesslog  (Default: "false")
>     Access log settings.
> --acme.*
>     ACME configuration
> --api  (Default: "false")
>     Enable api/dashboard.
> --hostresolver  (Default: "false")
>     Enable CNAME Flattening.
> 
> $ traefik --accesslog --help
> --accesslog  (Default: "false")
>     Access log settings.
> 
> --accesslog.bufferingsize  (Default: "0")
>     Number of access log lines to process in a buffered way.
> 
> --accesslog.fields.defaultmode  (Default: "keep")
>     Default mode for fields: keep | drop
> 
> --accesslog.fields.headers.defaultmode  (Default: "keep")
>     Default mode for fields: keep | drop | redact
> ```
> 
> The `--acme` option is an exception indeed, but we can find a workaround.

emilevauge:
@mpl 
>I do not understand why you say it has become unreadable. The number of items displayed is not larger than it was before. So do you mean that:
>1. it was already unreadable before this PR, because there was/are too many options ? (fair enough)
>2. or you find the new format not as readable as the old format? if yes, why?

3 years ago, it was readable, because Traefik had far less options at that time. But not anymore. So basically, it's 1/ :)

emilevauge:
After discussing, we will work on the help issue in the near future.

pkg/provider/rest/rest.go
```diff
@@ -19,7 +19,13 @@ var _ provider.Provider = (*Provider)(nil)
...
 type Provider struct {
 	configurationChan chan<- config.Message
-	EntryPoint        string `description:"EntryPoint" export:"true"`
+	EntryPoint        string `description:"EntryPoint." export:"true"`
+}
+
+// SetDefaults sets the default values.
+func (p *Provider) SetDefaults() {
+	p.EntryPoint = "traefik"
+	// FIXME p.EntryPoint = static.DefaultInternalEntryPointName
```
jbdoumenjou: As it seems to be difficult to avoid the import cycle, is this FIXME usefull ? Same question for in types/metrics.go

Issue #3870, opened 2018-09-08
Traefik log can't switch the timezone


maifusha:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

In my case, I wonder if I can switch the time zone to use the local time, and the traefik log and access log will make more sense and easier to understand

But I search over the  traefik official document or github issues, i cant find any options to achieve it

jpalczewski:
For me:
```
                volumes:
                        - /var/run/docker.sock:/var/run/docker.sock
                        - /var/run/log/traefik/access.log:/tmp/access.log
                        - "/etc/timezone:/etc/timezone:ro"
                        - "/etc/localtime:/etc/localtime:ro"
```
did the trick - `StartLocal` in json-formatted access log became true `StartLocal`.

v7lin:
for me:

docker-compose.yml
````
  traefik:
    container_name: traefik
    image: traefik:1.7.5-alpine
    restart: always
    hostname: traefik
    ports:
      - 80:80
      - 443:443
#      - 8080:8080
    volumes:
      - ../dev-ops-repo/traefik/acme:/etc/traefik/acme
      - /var/run/docker.sock:/var/run/docker.sock
      - ./traefik/entrypoint-fixed.sh:/entrypoint-fixed.sh
    environment:
      - TZ=${TIME_ZONE}
    command:
      - "--loglevel=WARN"
      - "--api"
      - "--defaultentrypoints=http,https"
      - "--entrypoints=Name:http Address::80"
      - "--entrypoints=Name:https Address::443 TLS"
      - "--acme"
      - "--acme.storage=/etc/traefik/acme/acme.json"
      - "--acme.entryPoint=https"
      - "--acme.httpChallenge.entryPoint=http"
      - "--acme.onHostRule=true"
      - "--acme.onDemand=false"
      - "--acme.email=${TRAEFIK_ACME_EMAIL}"
      - "--docker"
    entrypoint:
      - /entrypoint-fixed.sh
````

traefik/entrypoint-fixed.sh
````
#!/bin/sh
apk update && apk add curl bash tree tzdata
./entrypoint.sh
````

````
/ # ls
bin                  entrypoint.sh        lib                  proc                 sbin                 tmp
dev                  etc                  media                root                 srv                  usr
entrypoint-fixed.sh  home                 mnt                  run                  sys                  var
/ # date -R
Thu, 13 Dec 2018 10:12:45 +0800
````

ycharron:
[...]

PR #4953, opened 2019-06-13
Update documentation to clarify the default format for logs

### LINKED ISSUE #4869: Log format config ignored
Ver: 1.7.11 - traefik_linux-arm

```
################################################################
# Global configuration
################################################################
                  
# Enable debug mode      
#
# Optional                                                      
# Default: false            
#                                                               
debug = true
              
# Log level                           
#
# Optional
# Default: "ERROR"
#           
# logLevel = "DEBUG"
logLevel = "INFO"                                                              
                                                    
# Entrypoints to be used by frontends that do not specify any entrypoint.
# Each frontend can specify its own entrypoints.
#                   
# Optional
# Default: ["http"]           
#                       
[traefikLog]
                    
# Sets the filepath for the traefik log. If not specified, stdout will be used.
# Intermediate directories are created if necessary.
#                                                                        
# Optional                                      
# Default: os.Stdout
#         
# filePath = "log/traefik.log"
filePath = "traefik.log"
                                      
# Format is either "json" or "common".
#                                                               
# Optional                 
# Default: "common"                                             
#
# Is this broken (1.7.11) ?
format = "common"
```

With the above cfg, content of traefik.log is in json format. It should be in 'common' format.
I'm sure this was working correctly in 1.7.4 - just upgraded.

dduportal:
Signed-off-by: dduportal <1522731+dduportal@users.noreply.github.com>

### What does this PR do?

This PR clarifies the documentation page about logs:

- Adding comment to explicit that the default value is `common` and refers to the default CLF format
- Clarifies that customizing the fields requires to switch to the JSON format

### Motivation

This PR follows up #3942 and is inspired by #3870 where the user was not aware about the mandatory format switch, which could have avoided wasting time searching for a non existent configuration.

Closes #4869.

### More

- ~[ ] Added/updated tests~

### Additional Notes

sheldon-nitpick

### DIFF
--- a/docs/configuration/logs.md
+++ b/docs/configuration/logs.md
@@ -13,7 +13,7 @@ logLevel = "INFO"
 
 [accessLog]
   filePath = "/path/to/access.log"
-  format = "json" # Default: "common"
+  format = "json"
 
   [accessLog.filters]
     statusCodes = ["200", "300-302"]
@@ -44,7 +44,7 @@ For more information about the CLI, see the documentation about [Traefik command
 --traefikLog.filePath="/path/to/traefik.log"
 --traefikLog.format="json"
 --accessLog.filePath="/path/to/access.log"
---accessLog.format="json"  # Default: "common"
+--accessLog.format="json"
 --accessLog.filters.statusCodes="200,300-302"
 --accessLog.filters.retryAttempts="true"
 --accessLog.filters.minDuration="10ms"
--- a/docs/configuration/logs.md
+++ b/docs/configuration/logs.md
@@ -9,11 +9,11 @@ logLevel = "INFO"
 
 [traefikLog]
   filePath = "/path/to/traefik.log"
-  format   = "json"
+  format   = "json" # Default: "common"
 
 [accessLog]
   filePath = "/path/to/access.log"
-  format = "json"
+  format = "json" # Default: "common"
 
   [accessLog.filters]
     statusCodes = ["200", "300-302"]
@@ -39,12 +39,12 @@ logLevel = "INFO"
 
 For more information about the CLI, see the documentation about [Traefik command](/basics/#traefik).
 
-```shell
+```bash
 --logLevel="DEBUG"
 --traefikLog.filePath="/path/to/traefik.log"
---traefikLog.format="json"
+--traefikLog.format="json"  # Default: "common"
 --accessLog.filePath="/path/to/access.log"
---accessLog.format="json"
+--accessLog.format="json"  # Default: "common"
 --accessLog.filters.statusCodes="200,300-302"
 --accessLog.filters.retryAttempts="true"
 --accessLog.filters.minDuration="10ms"
@@ -53,7 +53,6 @@ For more information about the CLI, see the documentation about [Traefik command
 --accessLog.fields.headers.defaultMode="keep"
 --accessLog.fields.headers.names="User-Agent=redact Authorization=drop Content-Type=keep"
 ```
-
 
 ## Traefik Logs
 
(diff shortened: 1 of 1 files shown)


dduportal:
=> To be done: we need to check, there might be an issue classified as "bug/possible" around setting the format to the value `common` which does not work...

dduportal:
@ldez it looks like that the value `common` is the right one (by default):

- https://github.com/containous/traefik/blob/v1.7/cmd/configuration.go#L195
- https://github.com/containous/traefik/blob/v1.7/middlewares/accesslog/logger.go#L29


Also, I tried to set `accesslog.format` (with CLI below, but also tried with `traefik.toml`) to `common`, and no error and the behavior is the same as only enabling `accesslog` alone

```yml
version: '3'

services:
  lb:
    image: traefik:v1.7.12
    restart: unless-stopped
    command: --api --docker --accesslog.format=common
    ports:
      - 80:80
      - 8080:8080
    volumes:
    labels:
      - traefik.enable=false

  webapp:
    image: containous/whoami
    labels:
      - "traefik.frontend.rule=Host:localhost"
```

I'm removing the `bot/no-merge` and this PR is ready to review (unless I missed something of course!)

docs/configuration/logs.md
```diff
@@ -9,11 +9,11 @@ logLevel = "INFO"
 
 [traefikLog]
   filePath = "/path/to/traefik.log"
-  format   = "json"
+  format   = "json" # Default: "common"
```
ldez:
```suggestion
  format   = "json"
```

docs/configuration/logs.md
```diff
@@ -9,11 +9,11 @@ logLevel = "INFO"
...
 
 [accessLog]
   filePath = "/path/to/access.log"
-  format = "json"
+  format = "json" # Default: "common"
```
ldez:
```suggestion
  format = "json"
```

docs/configuration/logs.md
```diff
@@ -39,12 +39,12 @@ logLevel = "INFO"
 
 For more information about the CLI, see the documentation about [Traefik command](/basics/#traefik).
 
-```shell
+```bash
 --logLevel="DEBUG"
 --traefikLog.filePath="/path/to/traefik.log"
---traefikLog.format="json"
+--traefikLog.format="json"  # Default: "common"
```
ldez:
```suggestion
--traefikLog.format="json"
```

docs/configuration/logs.md
```diff
@@ -39,12 +39,12 @@ logLevel = "INFO"
...
 --accessLog.filePath="/path/to/access.log"
---accessLog.format="json"
+--accessLog.format="json"  # Default: "common"
```
ldez:
```suggestion
--accessLog.format="json"
```

docs/configuration/logs.md
```diff
@@ -66,15 +65,14 @@ To write the logs into a log file specify the `filePath`:
   filePath = "/path/to/traefik.log"
 ```
 
-To write JSON format logs, specify `json` as the format:
+To switch to JSON format instead of [Common Log Format (CLF)](#clf-common-log-format), specify `json` as the format:
```
ldez:
```suggestion
To switch to JSON format instead of standard format (`common`), specify `json` as the format:
```

docs/configuration/logs.md
```diff
@@ -66,15 +65,14 @@ To write the logs into a log file specify the `filePath`:
...
 
 ```toml
 [traefikLog]
   filePath = "/path/to/traefik.log"
-  format   = "json"
+  format   = "json" # Default: "common"
```
ldez:
```suggestion
  format   = "json"
```

docs/configuration/logs.md
dduportal: Isn't this true for the `accessLog` ?

docs/configuration/logs.md
dduportal: Isn't this true for the `accessLog` ?

PR #4954, opened 2019-06-13
Allows logs to use local time zone instead of UTC

### LINKED ISSUE #3870: Traefik log can't switch the timezone

dduportal:
### What does this PR do?

This PR introduces a fix to allow local timezone for the field `timestamp` on the log.

Kudos to @ldez for providing this code fix without breaking the current behavior!

### Motivation

We want user to be able to print the time in their local,
without requiring switching to JSON,
and without introducing a breaking change.

closes #3870

### More

- [x] Added/updated tests

### Additional Notes

gif antman

Co-authored-by: Ludovic Fernandez <ldez@users.noreply.github.com>

### DIFF
--- a/docs/configuration/logs.md
+++ b/docs/configuration/logs.md
@@ -292,3 +292,34 @@ This allows the logs to be rotated and processed by an external program, such as
 
 !!! note
     This does not work on Windows due to the lack of USR signals.
+
+## Time Zones
+
+The timestamp of each log line is in UTC time by default.
+
+If you want to use local timezone, you need to ensure the 3 following elements:
+
+1. Provide the timezone data into /usr/share/zoneinfo
+2. Set the environement variable TZ to the timezone to be used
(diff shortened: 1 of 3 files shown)


docs/configuration/logs.md
```diff
@@ -292,3 +292,34 @@ This allows the logs to be rotated and processed by an external program, such as
...
+  traefik:
+    image: containous/traefik:[latest stable version]
+    ports:
+      - "80:80"
+    environment:
+      - "TZ=US/Alaska"
+    command:
+      - --docker
+      - --accesslog
+      - --accesslog.fields.names="StartLocal=keep StartUTC=drop"
```
mbleichner:
```suggestion
      - --accesslog.fields.names="StartUTC=drop"
```

is actually enough to get the desired behavior.

PR #4960, opened 2019-06-15
feat: service weights for ingressroute

moolen:
### What does this PR do?

Services defined in an IngressRoute CRD do not support the `weight` attribute. This PR adds that functionality.

### Motivation
I was trying to figure out how to do a canary deployment w/ weighted services - i.e. having `whoami-stable` and `whoami-canary` with a 9:1 traffic ratio.

### More

- [x] Added/updated tests

### Additional Notes

I just added a example to the dynamic configuration doc. Do you think that the spec of IngressRoute and Middleware CRDs is kinda final? I'd like to add it to the docs.

I used this IR for testing:
```yaml
kind: IngressRoute
metadata:
  name: whoami
spec:
  entryPoints:
    - web
  routes:
  - match: HostRegexp(`{host:whoami.*}`)
    kind: Rule
    services:
    - name: whoami-stable
      weight: 1
      port: 80
    - name: whoami-canary
      weight: 99
      port: 80

```

I did some trivial analysis by sending n requests to traefik and looking which upstream is used. Here are the results:

| ratio | num reqs | result | actual ratio |
| --- | --- | --- | --- |
| 1:99 | 17k | 1 : 16999 | 0.006% |
| 2:98 | 1k | 6 : 994 | 0.6% |
| 3:97 | 1k | 3 : 997 | 0.3% |
| 5:95 | 1k | 63 : 964 | 6.5% |
| 10:90  | 1k | 87 : 913 | 9.5% |
| 30:70  | 1k | 285 : 715 | 39.9% |

I just did a single run. I think that ratios below 5% are not stable. I'll do further analysis and add it to the docs once i find the time.

### DIFF
--- a/docs/content/reference/dynamic-configuration/kubernetes-crd.yml
+++ b/docs/content/reference/dynamic-configuration/kubernetes-crd.yml
@@ -45,6 +45,8 @@ spec:
       services:
         - name: s1
           port: 80
+          # specify how much will be routed to this service
+          weight: 90
           healthCheck:
             path: /health
             host: baz.com
@@ -55,6 +57,7 @@ spec:
           strategy: RoundRobin
         - name: s2
           port: 433
+          weight: 10
           healthCheck:
             path: /health
             host: baz.com
--- a/pkg/provider/kubernetes/crd/fixtures/with_two_services_weighted.yml
+++ b/pkg/provider/kubernetes/crd/fixtures/with_two_services_weighted.yml
@@ -1,0 +1,21 @@
+apiVersion: traefik.containo.us/v1alpha1
+kind: IngressRoute
+metadata:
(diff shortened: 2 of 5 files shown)


ldez:
Thanks.

FYI the current dev branch is `v2.0` (the branch `master` not up to date for now)

We have removed `weight` for all providers https://github.com/containous/traefik/pull/4933

> We plan on adding more load balancer strategies later on. In order to do that we'll need some sort of generic method that may turn out to be incompatible with the current way wrr and drr are implemented. Therefore, if we do not want to break configuration later on when we add these new strategies, we have to remove wrr and drr now.

Then we will have to decline your PR.

Thanks for your time and we hope you will continue to contribute.

Issue #1656, opened 2017-05-23
Support all tags from Providers for constraint filtering


aantono:
The use-case here is to enable the ability to filter (use tag constraints) on the existing provider tags (Consul tags, Docker labels, etc), not only the ones that are defined in `traefik.tags`. 

Many places with existing SD infrastructure already have quite an elaborate tagging setup for their services, so being able to leverage those for filtering is a big plus.

 For example, each Spring Boot application that uses `sping-cloud-consul` module, will automatically register the management endpoint using `management` tag. 

Because it might not be desired to expose LB routes to management endpoints, one might want to create a constraint rule like constraints = `["tag!=management"]`, and when coupled with `AllTagsConstraintFiltering=true`, it will use the "raw" `management` tag value, instead of only the values provided by `traefik.tags` tag.

traefiker:
Closed by #4965.

PR #4965, opened 2019-06-17
New constraints management.

### LINKED ISSUE #895: More generic traefik constraints "tags" for docker provider
Hi,

I think it would be great if the `--constraints` parameter become more generic.
Some of labels are automatically placed by docker or compose (and maybe by other orchestrator), and traefik could use them to filter

Today we place a constraint (`--constraints='tag==eu'`)
And then a label `traefik.tag=eu` is placed on container

For example, if my container already have a label like `"com.docker.compose.project": "proj1"`, I would like to tell traefik instance to manage only this project

Maybe the `--contraints` parameter can be more generic by letting user decide which label to check, like : 
`--constraints='com.docker.compose.project=="proj1"'`
So the old way can be done with : 
`--constraints='traefik.tags=="eu"'`

WDYT ?

ldez:
### What does this PR do?

New constraints management:

- allows to use any labels as a constraint.
- removes `traefik.tags`
- allows to use AND, OR, NOT, ...

```toml
[providers.docker]
constraints = "Label(`a.label.name`, `valueA`) && (Label(`another.label.name`, `valueB`) || LabelRegexp(`yet.another.label.name`, `a.+`))"
```

```toml
[providers.marathon]
constraints = "MarathonConstraint(`A:B:C`) && Label(`a.label.name`, `value`)"
```

Fixes #895, #1656

### Motivation

Make constraint configuration simpler and more powerful.

### More

- [x] Added/updated tests

### Additional Notes


docs/content/providers/docker.md
```diff
@@ -215,6 +215,48 @@ _Optional, Default=15_
 
 Defines the polling interval (in seconds) in Swarm Mode.
 
+### `constraints`
```
jbdoumenjou: It may be easier to include one external file with all the common examples between the provider that use constraints. It could avoid some duplication. WDYT ?

ldez: it's not possible because the comments are different for each provider.

jbdoumenjou: ok

smithy9901:
@ldez Tell me please, now there is no support constraints for the consul?

ldez:
- https://doc.traefik.io/traefik/v2.3/providers/consul-catalog/#constraints
- https://doc.traefik.io/traefik/v2.3/providers/consul/#rootkey

Issue #3237, opened 2018-04-26
Sporadic 502 response only when running through traefik


koliyo:
### Do you want to request a *feature* or report a *bug*?
Bug

### What did you do?

I have a graphql API running on NodeJS using Apollo and Express, with traefik in front.

When proxying through traefik I get sporadic 502 responses that I have not been able to resolve.

I does _never_ happen when I bypass the proxy and connect directly to the backend node server.

I am running all tests locally on my dev machine.

My first attempt to force the error was load testing with the [locust](locust.io) framework. However, even when sending large amounts of request through the proxy I was unable to replicate the error. It only happens when I use the frontend application in the browser.

After reading [this](https://github.com/vulcand/oxy/issues/112) oxy issue I started suspecting cancelled connections.

I added a custom HTTP header with a UUID to be able to trace all requests, which I print on the backend.
```js
  app.use((req, res, next) => {
    const id = req.headers['x-request-id'];
    if (id) {
      console.log(`Request id: ${id}`);
    }
    next();
  });

```

Then I also added the following event listener to the express server to [detect cancelled requests](https://stackoverflow.com/questions/35198208/handling-cancelled-request-with-express-node-js-and-angular)
```js
  app.use((req, res, next) => {
    req.connection.on('close', () => {
      const id = req.headers['x-request-id'];
      console.log(`Cancelled request ${id}`);
    });

    next();
  });
```

What I can see is that I _do_ get cancelled requests when running the application in the browser, and at some point i get a `502` response from traefik. And in the traefik log this is 

```
DEBU[2018-04-26T13:43:51+02:00] vulcand/oxy/forward/http: Round trip: http://localhost:6543, code: 502, Length: 11, duration: 66.352475ms 
```

And the nodejs backend log looks something like this:

```
...
Request id: 7455804b-490a-4361-98e5-43d12bf4aca8
Request id: 737f8d9d-3300-461b-858b-07006582a937
POST /graphql 200 83.542 ms - 310
POST /graphql 200 16.441 ms - 682
Request id: 096e0e39-90e6-475c-b8ad-0aa2dfd2e345
POST /graphql 200 5.338 ms - 163
Request id: 69f17cb2-cdf1-4db5-a9f5-08e46d795892
Request id: 50d3aec6-5cda-4e8b-ac0e-a30a57fa94c9
POST /graphql 200 58.596 ms - 310
POST /graphql 200 15.526 ms - 682
[...]

Issue #727, opened 2016-10-10
Support for configurable Keep-Alive for both frontend and backend 


bgv:
Hi,
I want to propose adding new configuration options prompt by #718.

While certainly its nice to have Keep-Allive from end to end sometimes we don't have full control over the applications we deploy.

In some situations it's good to have more granular control over the connection, so for example if we use the application just to print few bits of information we don't need the connection staying open for another 30 seconds.

For example when creating new service with docker swarm we can just add new label for this particular frontend: `--label traefik.frontend.rule=Client-KeepAlive:5,Backend-KeepAlive:0` which will tell traefik to keep the connection to client  browser open for 5 sec but close connection to backend after the response is finished.

So the full service command becomes:

``` bash
docker service create --name whoami --replicas 3 --label traefik.port=80 --label traefik.frontend.rule=Client-KeepAlive:5,Backend-KeepAlive:0 --network frontend emilevauge/whoami
```

I don't think this is urgent feature but certainly good to have.
Cheers.

jsravn:
I think this feature is pretty urgent. It means traefik will generate errors for non-idempotent requests - go http client only retries closed idle connections for idempotent requests. Only way to prevent this is to set the forwarder's transport idle timeout to less than our server timeouts. In golang 1.6 this isn't straightforward, while 1.7 has added the idle timeout that can be used for this. I'm not sure how easy it will be to do this, it probably requires a change to the oxy library that traefik uses.

27Bslash6:
Adding my reasoning for this request: optimal configuration for Akamai <> origin is keepalive of 301 seconds (https://community.akamai.com/customers/s/question/0D50f00005RtqgwCAB/default-pconn-timeout?language=en_US) but we don't want Traefik <> backend service connections persisting this long

stealthybox:
+1
It may be difficult/impossible to disable keep-alive support on a backend server
(particularly if the servers are part of a third party application).

#### Common use-case:
It's difficult to do zero-downtime rolling deploys of NGINX pods if keep-alive is enabled.
NGINX can't drain keep-alive connections during a graceful shutdown so it drops in-flight requests:
- https://github.com/kubernetes/contrib/issues/1123
- https://trac.nginx.org/nginx/ticket/1022

Say we're using Traefik to load-balance some NGINX servers;
It would be convenient to work around this issue by configuring the Traefik to not keep-alive those backend connections.

pmontepagano:
[...]

PR #4983, opened 2019-06-19
Make HTTP Keep-Alive timeout configurable for backend connections

mszabo-wikia:
### What does this PR do?
Make the HTTP Keep-Alive timeout for backend connections configurable (right now it is hardcoded to 90 seconds). 

An integration test was added to verify that the connections are closed in the specified time. 
### Motivation
There are a couple of open issues indicating that the ability to configure this would be welcome (#727, #3237). Furthermore, testing in our environment where traefik is used as a k8s ingress proxy indicated that the ability to tweak this setting has a measurable impact on the stability and performance of our workload.

### More

- [x] Added/updated tests


mszabo-wikia:
Done in d558bfdd

PR #4985, opened 2019-06-20
Filter env vars configuration

ldez:
### What does this PR do?

Filter env vars configuration.

### Motivation

Kubernetes automatically creates a bunch of variables prefixed with `TRAEFIK_` (probably based on the image name), and since our config loader was checking for the presence of the above prefix, this situation was obviously creating a conflict.

https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables

### More

- [x] Added/updated tests


dtomcej:
Not sure, but you may want to check these too, as services add environment variables by default too:

https://kubernetes.io/docs/concepts/services-networking/service/#environment-variables

PR #4990, opened 2019-06-20
Use name@provider instead of provider@name.

ldez:
### What does this PR do?

Uses `name@provider` instead of `provider@name`.

### Motivation

Enhance readability.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/middlewares/overview.md
+++ b/docs/content/middlewares/overview.md
@@ -121,8 +121,8 @@ If you use multiple `providers` and wish to reference a middleware declared in a
         image: your-docker-image
 
         labels:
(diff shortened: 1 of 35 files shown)


dtomcej:
I love this change. Especially with the move from `.` -> `@`

polarathene:
Is the `@` syntax meant to be working atm? I know it's [listed in the docs](https://docs.traefik.io/v2.0/middlewares/overview/#advanced-configuration) but when I tried it, [I didn't have any luck](https://github.com/containous/traefik/issues/4995)(see "Referencing cross-provider").

I had to use the `.` syntax instead.

PR #4992, opened 2019-06-20
Update default keytype in acme

ad-m:
According @Kindratte comment ( https://github.com/containous/traefik/issues/2673#issuecomment-472348299 ) ECDSA cryptography is more efficiently optimized in Go than RSA cryptography.

The following dokumnt explain benefits and limitations of the changes: https://csrc.nist.gov/csrc/media/events/workshop-on-elliptic-curve-cryptography-standards/documents/presentations/session2-andrews-rick.pdf 
Mozilla recommends to use ECDSA in "modern" configuration of web server ( https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility ). 
It would be optimal to implement dual-stack support and support both at the same time. My Go skills does not allow it.

Issue #2673, opened 2018-01-08
High CPU usage


tipruzs:
### Do you want to request a *feature* or report a *bug*?

bug

### What did you do?

When using traefik in front of dynamic web applications (e.g. nextcloud, speedtest) i see very high cpu usage up to 100%, when data is transferred (e.g. large downloads or speedtests).

Note: traefik and web apps run both in docker swarm

**Update 1:** disable compression on entrypoint. cpu usage still at 50-60% -  is this expected?

### What did you expect to see?

no cpu spikes (e.g. like general load balancers / reverse proxies)

### What did you see instead?

cpu rises up to 100% cpu load

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.4.5
Codename:     roquefort
Go version:   go1.9.2
Built:        2017-12-06_10:16:48AM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```yml
services:
  traefik:
    image: ${IMAGE}:${RELEASE}
    environment:
      ACME_EMAIL: ${ACME_EMAIL}
      ACME_DEFAULT_HOST: ${ACME_DEFAULT_HOST}
    command:
      - --configfile=/run/secrets/traefik_admin
      - --loglevel=WARN
      - --checknewversion=false
      - --insecureskipverify=true
      - --defaultentrypoints=http,https
      - "--entrypoints=Name:http Address::80 Compress:true Redirect.EntryPoint:https"
      - "--entryPoints=Name:https Address::443 TLS Compress:true"
      - --acme
      - --acme.email=${ACME_EMAIL}
      - --acme.domains=${ACME_DEFAULT_HOST}
      - --acme.onhostrule
      - --acme.entrypoint=https
      - --acme.storage=/data/acme.json
      - --docker
      - --docker.watch
      - --docker.exposedbydefault=false
      - --docker.swarmmode
      # - --debug
    networks:
     - traefik
    ports:
      - target: 80
        published: 80
        protocol: tcp
        mode: host
      - target: 443
        published: 443
        protocol: tcp
        mode: host
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock:ro
      - traefik:/data
    healthcheck:
      test: ["CMD-SHELL", "traefik healthcheck --web"]
      interval: 30s
      timeout: 5s
      retries: 2
      start_period: 30s
    deploy:
      restart_policy:
        condition: any
        delay: 2s
      update_config:
[...]

Issue #4872, opened 2019-05-14
Certificates Provider/Resolver


ldez:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

- Dynamic wildcard
- Challenge by domain(s) / multiple challenges
- DNS provider by domain(s) / multiple DNS providers
- Multiple certificates providers (acme, vault, ...)

---
### Static

```toml
[certificatesResolvers.default.acme]
  tlsStore = "test"
  storage = "acme.json"
  caServer = "https://acme-staging-v02.api.letsencrypt.org/directory"
  email = "test@test.com"
  keyType = "RSA4096"
  acmeLogging = true
#  resolvers = ["1.1.1.1:53", "8.8.8.8:53"]

  [certificatesResolvers.default.acme.TLSchallenge]
```

### Dynamic

```toml
[http.routers.router1]
  service = "service2"
  rule = "Host(`goo.snitest.com`)"
  [http.routers.router1.tls]
#    store = "test"
    certResolver = "default"
```

## Case: wildcard

Note: If a section `domain` is defined => it doesn't use the Host rules.

### Static

```toml
[certificatesResolvers.default.acme]
  tlsStore = "test"
  storage = "acme.json"
  caServer = "https://acme-staging-v02.api.letsencrypt.org/directory"
  email = "test@test.com"
  keyType = "RSA4096"
  acmeLogging = true

  [certificatesResolvers.default.acme.DNSchallenge]
```

### Dynamic

```toml
[http.routers.router1]
  service = "service2"
  rule = "Host(`goo.snitest.com`)"
  [http.routers.router1.tls]
#    store = "test"
    certResolver = "default"
    [[http.routers.router1.tls.domain]]
    main = "*.snitest.com"
#    SANs = ["foo.com", "fii.com", "fuu.com"]
#    certResolver = "default"
```

### Related use cases

- Host + wildcard
- HostRegex + wildcard
- no Host + wildcard
- Host + domain
- no Host + domain
- HostRegex + domain

## Case: wildcards

- Hosts + wildcards

### Static

```toml
[certificatesResolvers.default.acme]
  tlsStore = "test"
  storage = "acme.json"
  caServer = "https://acme-staging-v02.api.letsencrypt.org/directory"
  email = "test@test.com"
  keyType = "RSA4096"
  acmeLogging = true

  [certificatesResolvers.default.acme.DNSChallenge]
```

### Dynamic

```toml
[http.routers.router1]
  service = "service2"
[...]

PR #4993, opened 2019-06-20
Update default keytype in acme

ad-m:
According @Kindratte comment ( https://github.com/containous/traefik/issues/2673#issuecomment-472348299 ) ECDSA cryptography is more efficiently optimized in Go than RSA cryptography.

The following dokumnt explain benefits and limitations of the changes: https://csrc.nist.gov/csrc/media/events/workshop-on-elliptic-curve-cryptography-standards/documents/presentations/session2-andrews-rick.pdf 
Mozilla recommends to use ECDSA in "modern" configuration of web server ( https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility ). 
It would be optimal to implement dual-stack support and support both at the same time. My Go skills does not allow it.

Me personally motivates that this change is beneficial when the client and server are Go app.

This PR is resubmit of #4992 after mistake.

### DIFF
--- a/pkg/provider/acme/provider.go
+++ b/pkg/provider/acme/provider.go
@@ -55,7 +55,7 @@ type Configuration struct {
 func (a *Configuration) SetDefaults() {
 	a.CAServer = lego.LEDirectoryProduction
 	a.Storage = "acme.json"
-	a.KeyType = "RSA4096"
+	a.KeyType = "EC256"
 }
 
 // Certificate is a struct which contains all data needed from an ACME certificate


ldez:
I re-check the LE recommendations about this subject:

https://letsencrypt.org/docs/integration-guide/#supported-key-algorithms

> Our recommendation is to serve a dual-cert config, offering an RSA certificate by default, and a (much smaller) ECDSA certificate to those clients that indicate support.

As we don't support dual-cert config, I think we have to stay on RSA for now.

juliens:
Thanks for your contribution!

As previously said by ldez, we need to keep serving RSA by default, and we should add the ability to serve a dual-cert for ECDSA.

However as we will modify lots of things related to ACME soon ( cf. #4872 ), I think the best solution is to wait for all these important changes before handling dual-cert.

PR #5012, opened 2019-06-25
Remove X-Forwarded-(Uri, Method, Tls-Client-Cert and Tls-Client-Cert-Info) from untrusted IP

stffabi:
### What does this PR do?

The changes of the PR #4036 didn't make their way to v2.0

### More

- [x] Added/updated tests


jbdoumenjou:
Hi @stffabi,

Could you please add the `X-Forwarded-Tls-Client-Cert` and `X-Forwarded-Tls-Client-Cert-Info` ? 
I would be great! :smile:

stffabi:
I've added the commit to remove the passTLSClientCert Headers.

PR #5026, opened 2019-06-26
Ensure WaitGroup.Done() is always called

bsdelf:
### What does this PR do?

This PR makes `Pool.Stop()` and `Pool.Cleanup()` finish their jobs even after panicking goroutines encountered.

### Motivation

Though `safe.Pool` supports recoverable goroutine, `Pool.Stop()` and `Pool.Cleanup()` does not working correctly with panicking goroutines due to an unreliable `WaitGroup` counter usage.

### More

- [x] Added/updated tests

### Additional Notes

A new method `Pool.AddGo()` was added corresponding to existing `Pool.AddGoCtx()` to make a compact unit test possible.


pkg/safe/routine_test.go
```diff
@@ -173,6 +173,73 @@ func TestPoolStartWithStopChan(t *testing.T) {
...
+			p := NewPool(context.Background())
+
+			timer := time.NewTimer(500 * time.Millisecond)
+			defer timer.Stop()
+
+			test.fn(p)
+
+			testDone := make(chan bool, 1)
+			go func() {
+				p.Stop()
```
mpl: why not p.Cleanup() like in the other tests?

bsdelf: patch updated.

pkg/safe/routine.go
```diff
@@ -59,12 +59,23 @@ func (p *Pool) GoCtx(goroutine routineCtx) {
...
 	Go(func() {
+		defer p.waitGroup.Done()
 		goroutine(p.ctx)
-		p.waitGroup.Done()
 	})
 	p.lock.Unlock()
 }
 
+// AddGo adds a recoverable goroutine, and can be stopped with stop chan
+func (p *Pool) AddGo(goroutine func(stop chan bool)) {
```
mpl: If AddGo is not used in any other place than in the tests, let's make it non-exported please.

PR #5037, opened 2019-06-27
Allow fsnotify to reload config files on k8s (or symlinks)

### LINKED ISSUE #1188: Ability to use k8s configmap for traefik configuration (with live reload)
Based from [this](https://github.com/containous/traefik/issues/7#issuecomment-281486988) post, this issue is a feature request for the ability to utilize [ConfigMap](https://kubernetes.io/docs/user-guide/configmap/) in k8s for Traefik conifguration with the added ability to support live reloading on changes.

This ask is to help not require users to painfully interact with the toml file but instead use a k8s native construct in which they can merely write the ConfigMap with the needed options, and deploy it as part of they're already existent manifests (or in a [helm](https://github.com/kubernetes/helm) chart etc.)

dtomcej:
### What does this PR do?

Watches for delete events on the watched config file, and re-register the watch on the same file.

This allows configmaps to be updated in k8s, and have traefik follow the new updated code.

See:

https://martensson.io/go-fsnotify-and-kubernetes-configmaps/
http://thrawn01.org/posts/2016/03/22/howto-write-configmap-enabled-golang-microservices/
https://medium.com/@xcoulon/kubernetes-configmap-hot-reload-in-action-with-viper-d413128a1c9a

### Motivation

Fixes the original issue in #1188

### DIFF
--- a/pkg/provider/file/file.go
+++ b/pkg/provider/file/file.go
@@ -111,7 +111,10 @@ func (p *Provider) addWatcher(pool *safe.Pool, directory string, configurationCh
 				return
 			case evt := <-watcher.Events:
 				if evt.Op == fsnotify.Remove {
-					watcher.Remove(evt.Name)
+					err = watcher.Remove(evt.Name)
+					if err != nil {
+						log.WithoutContext().WithField(log.ProviderName, providerName).Errorf("Could not remove watcher: %s", err)
+					}
 					err = watcher.Add(directory)
 					if err != nil {
 						log.WithoutContext().WithField(log.ProviderName, providerName).Errorf("Could not re-add watcher: %s", err)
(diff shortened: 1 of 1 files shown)


shane-axiom:
Just a note, this doesn't fix all of the use cases identified in #1188 (for instance, file watches not working on NFS shares). The title of that issue has evolved to "Provide a hook to have Traefik live-reload its configuration." This PR is great but I'd hate for that issue to get closed out with only a partial fix.

pkg/provider/file/file.go
```diff
@@ -110,6 +110,17 @@ func (p *Provider) addWatcher(pool *safe.Pool, directory string, configurationCh
...
 				return
 			case evt := <-watcher.Events:
+				if evt.Op == fsnotify.Remove {
+					err = watcher.Remove(evt.Name)
+					if err != nil {
+						log.WithoutContext().WithField(log.ProviderName, providerName).Errorf("Could not remove watcher: %s", err)
+					}
+					err = watcher.Add(directory)
+					if err != nil {
+						log.WithoutContext().WithField(log.ProviderName, providerName).Errorf("Could not re-add watcher: %s", err)
```
mpl: Do we want to add the directory name to the error message ("Could not re-add watcher for %s: %s"), or does it already show up within err itself?

Issue #4951, opened 2019-06-12
[forward auth] add option to reuse the original request's Host header


costela:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

When forwarding a request to the auth server, the `Host` header carries the domain of the auth server itself, unlike other headers from the original request. This is fine for a lot of use-cases, but it would also be nice to be able to optionally reuse the `Host` header from the original request.
This would allow multiple different frontends to more easily use the same forward auth server. Or, conversely, allow the auth server to more easily differentiate between the entrypoints it's serving.

My specific use case is a k8s cluster using [keycloak](/keycloak/keycloak) and with a single [keycloak-gatekeeper](/keycloak/keycloak-gatekeeper) instance as forward auth server. Keycloak-gatekeeper is not used in its "normal" way (i.e.: as a transparent authenticating proxy), instead only proxying to a server that always returns `200 OK`.
Unauthenticated requests are forwarded to the `keycloak` instance, whereas authenticated requests always succeed.

Is there maybe some other easier way I might be overlooking? 

If there are no objections on principle, I could try my hand at a PR.

costela:
FTR: in the course of preparing my PR I've noticed my original idea probably won't work as I thought: when keycloak-gatekeeper receives the forwarded request for auth, it first performs a redirect-to-self to start the OIDC flow. If we use the original request's Host header when contacting the forward-auth server, we end up with a redirect loop.

I'll still submit my PR as-is for consideration, in case anyone else is using a non-OIDC auth server which happens to profit from this use-case.

PeerD:
Wouldn't it be more conveniant to send the original host and request url as a separate header to the forward auth server? Then you don't need a catch all hosts there (thus allowing you to run it on the same traefik instance as the site itself) and still have all information available.

The original host is neccessary to set the proper cookie domain during authentication while the original request url is needed to limit access to certain parts of the site. Also to "capture" an url like /logout to remove the session cookies again. I know this would probably be possible with reverse proxies for those urls, but just forwarding the request url to the forward auth server would be far easier imo.

PR #5041, opened 2019-06-28
add PassHostHeader option to forward auth

costela:
### What does this PR do?

Add a PassHostHeader option to forward auth, so that requests to the auth server use the incoming `Host` header, as opposed to inferring the `Host` header from the auth server's address.

This PR is against the 1.7 branch, but I would port it to 2.0 if accepted.

### Motivation

This should enable a few new use-cases for auth servers interacting with incoming requests from multiple separate frontends (e.g.: easier redirecting unauthed reqs to the right domain; doing different auth flows for different vhosts; etc).

My original use-case in #4951 turned out to not quite work as expected, but I still believe this might open up a few use-cases.

### More

- [x] Added/updated tests

### Additional Notes

The PR currently only supports docker and k8s providers, because these are the systems I could test on. If it's accepted on principle, I can work on adding support to the other providers.

### DIFF
--- a/autogen/gentemplates/gen.go
+++ b/autogen/gentemplates/gen.go
@@ -756,6 +756,7 @@ var _templatesDockerTmpl = []byte(`{{$backendServers := .Servers}}
       [frontends."frontend-{{ $frontendName }}".auth.forward]
         address = "{{ $auth.Forward.Address }}"
         trustForwardHeader = {{ $auth.Forward.TrustForwardHeader }}
+        passHostHeader = {{ $auth.Forward.PassHostHeader }}
         {{if $auth.Forward.AuthResponseHeaders }}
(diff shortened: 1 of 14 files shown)


ldez:
Hello,

your PR is interesting, thank you for your works :+1: 

But we will very happy :tada: if you can "rebase" your PR on the branch `v2.0`.

Thanks.

ldez:
@costela any news?

costela:
@ldez sorry, didn't manage to get around to it. It's still on my TODO, though.
Did you want to include it in the betas for 2.0? If yes, I can try to prioritize it a bit.
Not sure I'd be able to add support for all providers though. :confused:

ldez:
it would be nice to be able to integrate that before the RC phase

ldez:
I have to close this one because of its lack of activity :disappointed:

Feel free to re-open it.

PR #5099, opened 2019-07-15
Update example for letsencrypt

Letme:
I tried to run the example, but the container would not stay up (on RPI3, same image as above). `kubectl logs` was reporting:

```
command traefik error: failed to decode configuration from flags: field not found, node: trace
```

So I have decided to remove the line and the container started. I have no idea of the possible implication  of this removal.


ldez:
Thanks but it's already fixed in #5096

PR #5106, opened 2019-07-15
Add acme certificate import/export scripts

nanoscopic:
Created perl scripts to import/export basic certificates to/from acme.json

This enables users to more easily migrate to Traefik from other solutions, as well as to extract plain certificate files from acme.json for whatever reason that may be needed.

### What does this PR do?

It adds scripts that can be used to import or export Let's Encrypt certs to/from the internal acme.json file of Traefik.

### Motivation

I needed to import a certificate into Traefik to handle DNS migration when moving from Nginx to Traefik. Changing DNS at the same time as changing to a different box cannot be done reliably as DNS may change for clients before it changes for Let's Encrypt, resulting in no certificate being issuable on Traefik with TLS challenge.

### More

- [ ] Added/updated tests
- [ X] Added/updated documentation

### Additional Notes

Removing the previous script to export certs wasn't exactly helpful. I can understand that the shellscript was potentially hard to read and fragile. Hopefully this is cleaner and more reliably and is a more maintainable solution long term.

### DIFF
--- a/contrib/letsencrypt_acme/README.md
+++ b/contrib/letsencrypt_acme/README.md
@@ -1,0 +1,44 @@
+## Let's Encrypt Acme Import/Export
+
+The two scripts provided here are designed to import / export basic certificates issued by Let's Encrypt.
+
+Two scenarios are addressed by this:
+
+ - Using a pre-existing Let's Encrypt certificate with Traefik ( such as when migrating during a DNS change to Traefik when previously using Nginx )
+ - Migrating away from Traefik to Nginx
+
+### Import - Migrating to Traefik
+
+#### View usage
+
+    ./import.pl --help
+
+#### Import My Site
+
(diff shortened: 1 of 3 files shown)


ldez:
Thank you for your time.

We removed the previous export script (#4783) from the repository is favor of an external tool: https://github.com/ldez/traefik-certs-dumper/

Sorry but we will decline your PR.

nanoscopic:
The external tool only dumps, it does not import. Importing is the primary use case for this. Also; I already stated I was aware you removed the previous script. Sorry but you haven't given a good reason for declining it.

ldez:
Sorry, I may have sounded a little dry.
It was not a "go away", I was not my goal, I'm sorry.

I invite you to contribute to the dumper to add an import system.

FYI The main problem with the previous script was that we did not have time to handle this part properly, the fact that it was in bash was not really a problem.

PR #5119, opened 2019-07-18
API: remove configuration of Entrypoint and Middlewares

mpl:
### What does this PR do?

This PR removes the Middlewares and Entrypoint fields of the API handler configuration. As we're probably going to redo the routing and TLS handling (among other things) of other handlers (such as Ping, Rest provider, Prometheus metrics, etc) in the same way we're going to do it for the API handler, this PR also removes similarly extraneous fields for these as well.

### Motivation

This change comes in preparation of a a rework of the API handler configuration, which will define how traffic is routed to it, how we handle TLS with it, etc.
We envision that the Entrypoint and Middlewares are not going to be part of its configuration anymore when we redo it, so remove them as of now, so that we don't have to break the configuration later on.

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/cmd/healthcheck/healthcheck.go
+++ b/cmd/healthcheck/healthcheck.go
@@ -51,7 +51,7 @@ func Do(staticConfiguration static.Configuration) (*http.Response, error) {
 		return nil, errors.New("please enable `ping` to use health check")
 	}
 
-	pingEntryPoint, ok := staticConfiguration.EntryPoints[staticConfiguration.Ping.EntryPoint]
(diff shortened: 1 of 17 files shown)


cyberlooper:
Traefik:latest from docker hub totally breaks my config.
(command traefik error: field not found, node: address) this is what I get right now, I have removed the container and its image so that it starts brand new but no change.

I forced V1.7.16 with traefik:v1.7.16 and my installation is back online again.

Please advise so I can return to the latest builds.

PR #5121, opened 2019-07-18
Add a docker-compose & let's encrypt user-guide

pbenefice:
### What does this PR do?

Addind a User Guide to the existing documentation giving an example with docker-compose and let's encrypt ~~behind a basic auth~~

### Motivation

- I personally would have liked to find something like this in the doc
- it respond to some topics in the forum 
    - https://community.containo.us/t/thats-it-time-to-learn-traefik/896
    - ~~https://community.containo.us/t/a-global-http-https-redirection/864~~

### More

- [ ] ~~Added/updated tests~~

### Additional Notes

I wanted to know if you could be interested in this ?
If yes, feel free to give me feedback =) (I did it quite quickly)

Thanks.

### DIFF
--- a/docs/content/user-guides/docker-compose/acme-dns/docker-compose.yml
+++ b/docs/content/user-guides/docker-compose/acme-dns/docker-compose.yml
@@ -1,0 +1,41 @@
+version: "3.3"
+
+services:
+
+  traefik:
+    image: "traefik:v2.0.0-beta1"
+    container_name: "traefik"
+    command:
+      #- "--log.level=DEBUG"
+      - "--api=true"
+      - "--global.sendanonymoususage=false"
+      - "--providers.docker=true"
+      - "--providers.docker.exposedbydefault=false"
+      - "--entrypoints.web.address=:80"
+      - "--entrypoints.websecure.address=:443"
+      - "--certificatesresolvers.mydnschallenge.acme.dnschallenge=true"
+      - "--certificatesresolvers.mydnschallenge.acme.dnschallenge.provider=ovh"
+      #- "--certificatesresolvers.mydnschallenge.acme.caserver=https://acme-staging-v02.api.letsencrypt.org/directory"
+      - "--certificatesresolvers.mydnschallenge.acme.email=postmaster@mydomain.com"
+      - "--certificatesresolvers.mydnschallenge.acme.storage=/letsencrypt/acme.json"
+    ports:
+      - "0.0.0.0:80:80"
+      - "0.0.0.0:443:443"
+      - "127.0.0.1:8080:8080"
+    environment:
+      - "OVH_ENDPOINT=xxx"
+      - "OVH_APPLICATION_KEY=xxx"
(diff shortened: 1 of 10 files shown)


ldez:
Could you rebase on `v2.0` branch?

pbenefice:
I think I did something wrong xD, I tried :
```
git remote add upstream https://github.com/containous/traefik.git
git fetch upstream
git pull --rebase upstream v2.0
git push origin user_guide-docker-compose -f
```
at least all the commits from v2.0 are here :joy: 

I'll try to fix that

pbenefice:
I changed base to `containous:v2.0`, is it okay @ldez ?

ldez:
It's ok :+1: 

FYI:
```bash
git remote add upstream https://github.com/containous/traefik.git
git fetch upstream
git rebase --onto upstream/v2.0 HEAD^
git push origin user_guide-docker-compose -f
```

pbenefice:
Thanks for the tip :+1:

pbenefice:
Thanks for your review, I'll try to revamp this during the weekend.

lifeofguenter:
@ldez is that actually working? I think there are some users (including me) who can not get it to work as documented: https://community.containo.us/t/acme-config-broken-in-beta1/958

I am getting the following error:

```
time="2019-07-21T08:54:09Z" level=error msg="the router jenkins_https uses a non-existent resolver: basic"
```
However the resolver exists in the `traefik.toml`:

```
[certificatesResolvers.basic.acme]
  email = "my@email.com"
  storage = "/etc/traefik/acme/acme.json"

  [certificatesResolvers.basic.acme.httpChallenge]
    entryPoint = "web"
```

pbenefice:
@ldez Hi again =)
I rewrote the guide focusing more on let's encrypt. What do you think ?
cc @SantoDE 

Thanks.

pbenefice:
Hi @SantoDE, thanks for your feedback.

IMHO the DNS challenge is important. It is the only challenge allowing for wildcard cert generation and it does not require to be directly exposed to the internet.
And, it's my opinion, but I don't feel it would be too difficult to adapt to a different provider : you just need to change the variables.

I can add a section detailing the TLS challenge if you wish, but my opinion is that the DNS one as real advantages and should be introduced (which doesn't force users to adopt it).
But it's only my opinion. What do you think ?

SantoDE:
Hey @pbenefice,

I agree. That would be a nice idea :) And yeah, maybe we make it clear that the DNS stuff changes per provider (values).

I also think detailing TLS might be a good idea :)

pbenefice:
Well, I did something wrong with git, and lost all the work ^^

pbenefice:
Thanks `reflog`, I managed to get my work back. I'm definitly not familiar with upstreams ^^ 
Anyway, I added a basic example with docker-compose that I use as a starting point to detail each 3 let's encrypt challenges in their own page.

@SantoDE WDTY ?

PR #5131, opened 2019-07-19
Add example for CLI

alvarezbruned:
There are few examples for CLI, and I see necessary CLI example to have everything in the docker-compose manifest; 
in a single file instead of in a manifest and in a TOML file

### More

- [ ] Added/updated tests


dduportal:
Hi @SantoDE , let's focus this PR on `v1.7`, as `v2.0`already have the CLI example now :)

PR #5147, opened 2019-07-23
Do not follow redirects for the health check URLs

coder-hugo:
### What does this PR do?

This PR avoids that the health checker follows HTTP redirects.

### Motivation

The default behavior of the go HTTP client is to follow redirects and stop after 10 consecutive requests ([see](https://golang.org/src/net/http/client.go?s=1928:3975#L46)). I wouldn't expect this from a health checker. A redirect could point to a completely different host which is broken. The result of this would be that traefik removes a healthy backend.

### More

- [x] Added/updated tests


jbdoumenjou:
Hello @coder-hugo,

We think that not following the redirect for the health check makes sense but we can't integrate it like that without risking breaking existing users relying on the current behavior.

WDYT about integrating it as an health check option?

pkg/healthcheck/healthcheck.go
```diff
@@ -234,11 +235,17 @@ func checkHealth(serverURL *url.URL, backend *BackendConfig) error {
 
 	req = backend.addHeadersAndHost(req)
 
-	client := http.Client{
+	client := &http.Client{
```
mpl: what is the rationale for that change?

PR #5167, opened 2019-07-30
Correct Kubernetes Ingress and IngressRoute port heuristic for choosing HTTPS

### LINKED ISSUE #5164: "kubernetesIngress" provider does not detect protocol for backend pods correctly

Bug

- [ ] Yes
- [x] No

### What did you do?

Per the discussion in the Discourse "Traefik v2" topic [_Kubernetes Ingress annotation doesn't force HTTPS to backend pods_](https://community.containo.us/t/1056), [the three ways supported by earlier versions of Traefik](https://docs.traefik.io/configuration/backends/kubernetes/#tls-communication-between-traefik-and-backend-pods) to indicate that a Kubernetes _Ingress_ object requires HTTPS communication don't work properly in Traefik version 2 (as of version 2.0.0-beta1). The only method that works as advertised is detecting that a Kubernetes _Service_ port has a name and the name starts with "https."

Suppose you have a Kubernetes _Ingress_ object pointing at a _Service_ for which the backing server pods speak HTTPS. You want Traefik to use HTTPS for its proxy connections to these upstream servers. If your _Service_ port either lacks a name or has a name that doesn't start with "https," you want to fall back to one of the other two methods that used to work in earlier versions of Traefik.

First, the _Ingress_ annotation "ingress.kubernetes.io/protocol" with a value of "https" would allow you to request use of HTTPS regardless of your _Service_ ports; neither the name nor the port value should matter. Traefik no longer honors this annotation.

However, if your _Service_ port—not the target port on the backing pod container, but rather the port on which the _Service_ listens to receive connections—is 443, and if that port is selected by the _Ingress_ rule, then Traefik should use HTTPS. Traefik _tries_ here, but gets the implementation wrong.

An _Ingress_ object annotated with a key of "ingress.kubernetes.io/protocol" and a value of "https" pointing at a _Service_ port that lacks a name—or has a name that doesn't start with "https"—should use HTTPS.

Alternately, an _Ingress_ object pointing at a _Service_ port that's not named but has a value of 443—regardless of the target port used by the selected pods—should use HTTPS.

### What did you see instead?

Traefik both ignores the aforementioned annotation and looks at the wrong port to decide whether it sees 443; it looks at the _Endpoints_ object's ports, which are the target pod's ports, and not the ports used by the _Service_ sitting in front of those pods.

```
v2 Beta 1
```

The "kubernetesIngress" provider is active.

### Detailed Problem Analysis

Let's review the first _Ingress_ inspection predicate, as written for the previous version of Traefik:
> If the service port defined in the ingress spec is 443 (note that you can still use `targetPort` to use a different port on your pod)

At present, [the documentation for Traefik version 2's "kubernetesIngress" provider](https://docs.traefik.io/v2.0/providers/kubernetes-ingress/) makes no such promise, but the code is adapted from the earlier version of Traefik, and [is clearly trying to implement this promise](https://github.com/containous/traefik/blob/v2.0/pkg/provider/kubernetes/ingress/kubernetes.go#L216-L228).

In file _pkg/provider/kubernetes/ingress/kubernetes.go_, rather than inspecting [the `EndpointPort.Port` field](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L4133-L4134) at [line 228](https://github.com/containous/traefik/blob/v2.0/pkg/provider/kubernetes/ingress/kubernetes.go#L228), we should instead inspect [the `ServicePort.Port` field](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3928-L3929), available in function `loadService` as the expression `portSpec.Port`. In other words, line 228 could read as follows:
```go
if portSpec.Port == 443 || strings.HasPrefix(portName, "https") {
```
That's still not addressing taking the "ingress.kubernetes.io/protocol" annotation into account, but it's an improvement.

The test fixtures don't capture this case. The one that exercises this port 443 detection uses [a Kubernetes _Service_ whose service port is equivalent to its target port](https://github.com/containous/traefik/blob/v2.0/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-a-basic-rule-on-one-path-with-https-(port-%3D%3D-443)_service.yml#L9).

Note that this same problem afflicts the "kubernetesCRD" provider in file _pkg/provider/kubernetes/crd/kubernetes.go_, in the similarly named `loadServers` function [here](https://github.com/containous/traefik/blob/v2.0/pkg/provider/kubernetes/crd/kubernetes.go#L285-L287).

seh:
### What does this PR do?

When applying the port value heuristic to decide whether to contact a Kubernetes _Service_'s _Endpoints_ using HTTP or HTTPS, inspect the _Service_'s front-end port instead of the upstream servers' port values. That is, for example, when Traefik services an _Ingress_ or _IngressRoute_ object pointing at a _Service_ that forwards from port 443 to servers listening on port 8443, it should use HTTPS.

### Motivation

Fixes #5164 (though not yet honoring the "ingress.kubernetes.io/protocol" annotation yet [like in Traefik version 1.7](https://github.com/containous/traefik/blob/v1.7/provider/kubernetes/kubernetes.go#L355-L363))

### More

- [x] Added/updated tests

### Additional Notes

Should we include the "ingress.kubernetes.io/protocol" annotation here? At present, this patch addresses only the _Service_ port value.

### DIFF
--- a/pkg/provider/kubernetes/crd/fixtures/services.yml
+++ b/pkg/provider/kubernetes/crd/fixtures/services.yml
@@ -68,6 +68,7 @@ spec:
   ports:
     - name: web-secure
       port: 443
+      targetPort: 8443
   selector:
     app: containous
     task: whoami2
@@ -85,7 +86,7 @@ subsets:
       - ip: 10.10.0.6
     ports:
       - name: web-secure
-        port: 443
+        port: 8443
 
 ---
 apiVersion: v1
--- a/pkg/provider/kubernetes/crd/kubernetes.go
+++ b/pkg/provider/kubernetes/crd/kubernetes.go
@@ -273,7 +273,7 @@ func loadServers(client Client, namespace string, svc v1alpha1.Service) ([]dynam
 			case "http", "https", "h2c":
 				protocol = svc.Scheme
 			case "":
-				if port == 443 || strings.HasPrefix(portSpec.Name, "https") {
+				if portSpec.Port == 443 || strings.HasPrefix(portSpec.Name, "https") {
 					protocol = "https"
 				}
 			default:
--- a/pkg/provider/kubernetes/crd/kubernetes_test.go
+++ b/pkg/provider/kubernetes/crd/kubernetes_test.go
@@ -1324,10 +1324,10 @@ func TestLoadIngressRoutes(t *testing.T) {
 							LoadBalancer: &dynamic.LoadBalancerService{
 								Servers: []dynamic.Server{
 									{
-										URL: "https://10.10.0.5:443",
+										URL: "https://10.10.0.5:8443",
 									},
 									{
-										URL: "https://10.10.0.6:443",
+										URL: "https://10.10.0.6:8443",
 									},
 								},
 								PassHostHeader: true,
--- a/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-a-basic-rule-on-one-path-with-https-(port-==-443)_endpoint.yml
+++ b/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-a-basic-rule-on-one-path-with-https-(port-==-443)_endpoint.yml
@@ -8,8 +8,8 @@ subsets:
 - addresses:
   - ip: 10.10.0.1
   ports:
-  - port: 443
+  - port: 8443
 - addresses:
   - ip: 10.21.0.1
   ports:
-  - port: 443
+  - port: 8443
--- a/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-a-basic-rule-on-one-path-with-https-(port-==-443)_service.yml
+++ b/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-a-basic-rule-on-one-path-with-https-(port-==-443)_service.yml
@@ -7,4 +7,5 @@ metadata:
 spec:
   ports:
   - port: 443
+    targetPort: 8443
   clusterIp: 10.0.0.1
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -216,7 +216,7 @@ func loadService(client Client, namespace string, backend v1beta1.IngressBackend
 			}
 
 			protocol := "http"
-			if port == 443 || strings.HasPrefix(portName, "https") {
+			if portSpec.Port == 443 || strings.HasPrefix(portName, "https") {
 				protocol = "https"
 			}
 
--- a/pkg/provider/kubernetes/ingress/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes_test.go
@@ -732,10 +732,10 @@ func TestLoadConfigurationFromIngresses(t *testing.T) {
 								PassHostHeader: true,
 								Servers: []dynamic.Server{
 									{
-										URL: "https://10.10.0.1:443",
+										URL: "https://10.10.0.1:8443",
 									},
 									{
-										URL: "https://10.21.0.1:443",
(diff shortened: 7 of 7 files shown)


dtomcej:
This PR exposes how the new providers do not support multiport services.

😢

seh:
I hadn’t thought of that, coming at it with a narrow focus on repair. Should we confront that gap here?

dtomcej:
Going to discuss with the team tomorrow. May require a refactor.

seh:
While we do [collect only one port per reference to a _Service_](https://github.com/containous/traefik/blob/cf2b190efeae53f906edda30bd23c1b5f78c1224/pkg/provider/kubernetes/crd/kubernetes.go#L392-L403), an [an _IngressRoute_ can define multiple _Routes_](https://github.com/containous/traefik/blob/cf2b190efeae53f906edda30bd23c1b5f78c1224/pkg/provider/kubernetes/crd/traefik/v1alpha1/ingressroute.go#L8-L9), each of which can [point at any number of _Services_](https://github.com/containous/traefik/blob/cf2b190efeae53f906edda30bd23c1b5f78c1224/pkg/provider/kubernetes/crd/traefik/v1alpha1/ingressroute.go#L8-L9). Even though each reference includes only one port of the _Service_, it looks like it's still possible, though, to include multiple references to the same _Service_, each referring to a different port.

Studying this code again, it's surprising that we watch all of these object kinds, but if any one of them changes, we reload the universe by making requests of the Kubernetes API server. If instead we used Informers, we'd have an in-memory cache of these objects, and could react more specifically to what changed. For instance, if an _IngressRoute_ changes, you don't have to reload every other _IngressRoute_ too. If an _Endpoints_ changes, you have to work backwards through the dependency graph to figure out whether any _Services_ of the same name exist and are referenced by any existing _IngressRoutes_, and, if so, only update those.

dtomcej:
@seh,

Discussed with the team, and we feel that it is acceptable to proceed with this PR as it is now.

We will work on the multi-port issue in a future PR.

seh:
Thank you. I'd like to hear more detail about the multi-port concern, as my cursory reading [per my comment above](https://github.com/containous/traefik/pull/5167#issuecomment-516646492) saw that we _do_ accommodate _Services_ with more than one port. Perhaps you're referring to a different meaning of "service" here.

Is it worth filing an issue suggesting those improvements to the controller reconciliation procedure, watching and reacting to more fine-grained changes in the object graph?

dtomcej:
You can definitely file an issue, but currently providers are stateless. To maintain an object dependency graph, the state would need to be passed between iterations of the provider cycle, and potentially shared with the core server to persist with provider restarts.

May be worth examining, but Traefik does not operate as a controller, but as a watcher. The informers are registered to trigger a configuration reload, but each reload is atomic, so there is no inconsistencies to manage, and no "broken states".

seh:
Is there something else to be done here? We’d like to take advantage of this fix in the next container image release.

ldez:
@seh FYI our review process https://docs.traefik.io/v2.0/contributing/maintainers/#pr-review-process

seh:
I read that document several times, and the most sense I can make of it is that we need two more maintainers to approve this patch, given that @dtomcej already gave it the nod. If there are daily meetings to discuss open requests, can you tell me whether this one has come up for discussion?

dduportal:
Hi @seh , the daily meeting is for triaging. The review work on the PRs heavily depends on the maintainer availability. With the context here (assumption's week, summer's week), it's not realistic to commit on a deadline.

We hear your concern and we try our best, but right now there is nothing else outside waiting, as rushing a review is never a good idea.

Thanks for this contribution and for the help proposal, it's always really appreciated!

ldez:
FYI We don't know who is @reddy-s  but it's not a maintainer so his approval does not count.

ldez:
The error is not related to the PR, I will that.

PR #5171, opened 2019-08-02
Fix Traefik CLI command example

Livven:
### What does this PR do?

Fixes the documentation: `--entryPoints='Name:http Address::80'` did not work for me, but removing the quotes `--entryPoints=Name:http Address::80` as shown [here](https://docs.traefik.io/user-guide/kubernetes/#add-a-tls-certificate-to-the-ingress) did.

Maybe this is also specific to using it in a Kubernetes `.yaml` file, I'm not sure.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/configuration/entrypoints.md
+++ b/docs/configuration/entrypoints.md
@@ -85,8 +85,8 @@ defaultEntryPoints = ["http", "https"]
 For more information about the CLI, see the documentation about [Traefik command](/basics/#traefik).
 
 ```shell
---entryPoints='Name:http Address::80'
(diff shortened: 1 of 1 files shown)


dduportal:
Hi @Livven , the snippet here is for shell, so the quotes are required, or the shell interpreter will split based on the whitespace.

Just behind this example, there is an example adapted for `docker-compose.yml`.

Thanks a lot for this proposal, but we are refusing it.

If you feel like that there is still a problem, could you please open an issue describing the problem first please, so we can reproduce the problem?

PR #5191, opened 2019-08-11
Migration to go modules

ldez:
### What does this PR do?

- remove vendor and dep configuration
- add go modules files
- update scripts (Makefile, Dockerfile, Travis, scripts, ...)
- update contributing guide
- update k3s version (0.5.0 -> 0.8.0)

### Motivation

Be able to use `v2+` go module.

### More

- [x] Added/updated tests


r7l:
In order to add Traefik 2.x to Gentoo Linux, i was just looking around for the reason of the vendor directory to have vanished from Traefik 2.x unlike what was shipped with up to Traefik 1.7. Up until now, i was and still am maintaining the Traefik package for the 1.x Versions and would have loved to move on with Traefik 2.x. But it's unlikely to happen in the state it is now.

The issue now is that any provided release of Traefik 2.x can be considered broken unless you compile the thing yourself. Maintaining packages for distributions won't be thing unless the distribution in question ships their packages pre build or does not bother about random network access during build phase.

Let me explain:

Gentoo for example expects all the dependencies to exist in system or at least downloaded as files prior of starting the build phase of the actual package. To write a package (called Ebuild) for Gentoo, i would have to go through the list of dependencies listed in `go.mod` and copy paste every of these links to my browser and get the actual commit ID out of their repo and put both into a list inside the Ebuild. This might not be an issue for 2 or 3 dependencies but Traefik now comes with 100+. That's out of the question!

While not a Go developer myself, i am aware of this: https://tip.golang.org/cmd/go/#hdr-Modules_and_vendoring and it's just a bad practice when shipping production software and it's not supported by Gentoo and it will not be in future.

That means Traefik is on a dead end for Gentoo Linux. But on a side note, it seems that not only Gentoo is stuck now. FreeBSD for example seems also to be stuck on 1.7: https://www.freshports.org/net/traefik 

Please stop shipping software releases to production that is missing half the code!

ldez:
Hello,

It's not a real good place to speak about that but I will try to provide a quick anwser.

> Please stop shipping software releases to production that is missing half the code!

It's the go module way: no vendor and the use of [proxies](https://proxy.golang.org) to get, securely, dependencies files.

Go modules uses mainly the same approach than npm, maven, ... : a file descriptor to define direct dependencies and override of indirect dependencies.

For me, there is the same problem with other languages.

Maybe I don't understand the gentoo packaging, but seems like a problem for every Go (and node, java, ...) software.

As I understand, gentoo still use the "GOPATH' way to build Go application, it's a problem because the `GOPATH` will die.

I can see some Go software with Go modules on FreeBSD: https://www.freshports.org/security/lego

I propose to continue the discussion in a dedicated issue or a topic on the [forum](https://community.containo.us/)

r7l:
Thanks for your reply. Thought about your suggestion to discuss this in forums but i don't really see any progress coming out of it for my end as i don't really see you adding the vendors back. I am also not in the position to change Gentoo policy regarding network access while installing a package.

The main reason, i've replied here in the first place was that this is the source of the vanishing vendor directory.

What could help would be something like this https://crates.io/crates/cargo-ebuild which is a tool for Rust to solve the same issue. It basically translates their file containing a download list into a list usable for an Ebuild in Gentoo. But there is nothing like that for Go right now and i am not a Go developer. So i am not going to write something.

Regarding your statement about the future of `GOPATH`, if that's ever going to be removed, it might end up like it did for Composer (similar tool for PHP) which is now shipped with a patch that allows a slightly different behavior for Gentoo. But the patch actually comes from Fedora. Gentoo just picked it up from there.

The general idea about the inclusion of vendors into the package is that a package provided by any Linux distribution always comes with the goal to have to installed as expected from the maintainer when requested. For that reason, every of those dependencies will be put on the Distribution mirrors in order to not depend on every of those developers actually keeping their stuff online. Just imaging the fact, you release a new version and one of those dependencies in your list just decides to move their project or it gets taken over by someone with bad intentions. I've seen both situations already. Not with Go but with NodeJS. That's not entirely happening with the Ebuild i am maintaining as it's not included in the main tree. But that's where the policy comes from.

So thanks again for your reply.

dtomcej:
@r7l does `go mod vendor` not properly build the vendor directory as you are expecting?

It should create a vendor directory, and then populate it with all the dependencies that are required for the go build.

ldez:
@r7l maybe this can be useful https://archives.gentoo.org/gentoo-dev/message/f2a7abcc8506ae3e56a0ebb0ea0cadc8

But once again I don't know Gentoo packaging.

We will not restore the `vendor` in the git repo, but we can help you to write a tool or you can help us to write a tool, we can think about other solutions.

So I recommend to open a new dedicated issue in GitHub to speak more about that.

PR #5214, opened 2019-08-15
Updated proxy protocol dependency

Wagum:
### What does this PR do?

Update proxy protocol dependency 

### Motivation
Ability to expose proxy protocol destination address

### More
I wrote a custom header middleware to expose proxy protocol destination address in http header. This was not possible due to an old version lock on armon's proxy protocol repository. Im currently using this updated dependency in traefik version 1.7.13 and did not recognize any side effects.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -262,11 +262,11 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:35d20140aeb30f9ae04b481f20020fd06c9742339a64a55c1a5a5bfddd740564"
+  digest = "1:661a29880e546f59fa5c17211369379eb6431d382021431959adc13d66c89461"
   name = "github.com/armon/go-proxyproto"
   packages = ["."]
   pruneopts = "NUT"
-  revision = "48572f11356f1843b694f21a290d4f1006bc5e47"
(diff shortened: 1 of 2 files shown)


ldez:
Currently the development is on the `v2` branch.

As you can see, https://github.com/armon/go-proxyproto is already updated.

https://github.com/containous/traefik/blob/v2.0/go.mod#L19


Thanks for time.

ldez:
After reflection I think you tried to change the version `v1` (the development is happening on the branch` v1`).

The v1 is in maintenance mode, we only accept bug fixes.

We need to think about how to qualify your change: enhancement or bug-fix

Wagum:
Im sorry for requesting the PR on the master branch. This was built on 1.7. and i did not notice that the dependency is already updated on v2 development branch. 

Currently i am not using v2 but i will update asap. 
Without this change, i was not able to access the proxy protocol destination address. So this seems to be an enhancement for 1.7 users only. 

Thanks this information. I will think about contributing to the dev branch in the future :)

ldez:
After reflection with the team, we consider this PR as an enhancement.

As you are building a custom middleware we suppose you have a fork of Traefik, so we recommend to apply this change on your fork.

We will decline the PR.

PR #5229, opened 2019-08-19
Add constant tags for the DatadogStatsd metrics provider.

cbrintnall:
### What does this PR do?

Add constant tags for the DatadogStatsd metrics provider.

### Motivation

Needed it for personal use - but, also helps differentiate traefik metrics in a multi-tenant cluster where there is more than one traefik ingress.

### More

- [x] Added/updated tests


ldez:
Hello,

Thank you for your works.

As your PR is an enhancement we will decline the PR.

cbrintnall:
@ldez No worries - Is this feature currently implemented in v2? I can move the work over there if it isn't.

Issue #5226, opened 2019-08-19
healthcheck does not recognize ping configuration


jansauer:
## Do you want to request a *feature* or report a *bug*?

Bug

### Did you try using a 1.7.x configuration for the version 2.0?

- [ ] Yes
- [X] No

### What did you do?

I tried adding a health check to my traefik docker setup. After learning that curl is not included in the alpine image my goal was to use the `healthcheck` cli command. Unfortunately the command always fails with the error message the i'm allegedly don't have the ping enabled.
My configuration can be reproduce/dumped down using...  

```
$ docker run -d \
  --name traefik \
  --publish 80:80 \
  --publish 8080:8080 \
  --restart unless-stopped \
  traefik:v1.7.14-alpine \
    --ping \
    --api \
    --loglevel=DEBUG
$ docker exec -it traefik traefik healthcheck
Error calling healthcheck: please enable `ping` to use health check
```

Based on the debug log output and the fact that i can call `/ping` via curl i'm sure that ping is enabled.

```
$ curl http://localhost:8080/ping
OK
```

### What did you expect to see?

A 0 exit code.

### What did you see instead?

`Error calling healthcheck: please enable `ping` to use health check` with exit code 1.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
traefik:v1.7.14-alpine
Version:      v1.7.14
Codename:     maroilles
Go version:   go1.12.8
Built:        2019-08-14_09:46:58AM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

All configuration is provided via `docker run` command arguments.

### If applicable, please paste the log output in DEBUG level (`--log.level=DEBUG` switch)

```
$ docker logs traefik
time="2019-08-19T08:47:23Z" level=info msg="Traefik version v1.7.14 built on 2019-08-14_09:46:58AM"
[...]

PR #5233, opened 2019-08-20
Replace ambiguous cli help message wording

jansauer:
The term `API` could be mistaken with the Traefik web interface
exposing the configuration of router, services and middlewares
(As in https://docs.traefik.io/v2.0/operations/api/).

This pr replaces this term with the word `endpoint`, which is also
used in the documentation for this feature
(https://docs.traefik.io/v2.0/operations/ping/).

Fixes: #5226

### DIFF
--- a/cmd/healthcheck/healthcheck.go
+++ b/cmd/healthcheck/healthcheck.go
@@ -15,7 +15,7 @@ import (
 func NewCmd(traefikConfiguration *static.Configuration, loaders []cli.ResourceLoader) *cli.Command {
 	return &cli.Command{
 		Name:          "healthcheck",
-		Description:   `Calls Traefik /ping to check the health of Traefik (the endpoint must be enabled).`,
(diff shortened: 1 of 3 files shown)


cmd/healthcheck/healthcheck.go
```diff
@@ -15,7 +15,7 @@ import (
 func NewCmd(traefikConfiguration *static.Configuration, loaders []cli.ResourceLoader) *cli.Command {
 	return &cli.Command{
 		Name:          "healthcheck",
-		Description:   `Calls Traefik /ping to check the health of Traefik (the API must be enabled).`,
+		Description:   `Calls Traefik /ping to check the health of Traefik (the endpoint must be enabled).`,
```
mpl:
```suggestion
		Description:   `Calls Traefik /ping endpoint (disabled by default) to check the health of Traefik.`,
```

PR #5234, opened 2019-08-20
Remove the arguments from the usage help message

jansauer:
Traefik does not have or use any arguments.


ldez:
hello,

We plan to generate this documentation so your changes will be reverted at this time, so we wlll decline your PR.

thanks for you.

jansauer:
@ldez thx for the feedback

PR #5241, opened 2019-08-22
Update deprecated function call in k8s providers 

Wagum:
### What does this PR do?
Remove the usage of deprecated SharedInformerFactory in kubernetes providers. 

### Motivation
The kubernetes providers should use NewSharedInformerFactoryWithOptions(), due to the fact that NewFilteredSharedInformerFactory() is deprecated and may get removed in future releases.

### DIFF
--- a/pkg/provider/kubernetes/crd/client.go
+++ b/pkg/provider/kubernetes/crd/client.go
@@ -164,7 +164,7 @@ func (c *clientWrapper) WatchAll(namespaces []string, stopCh <-chan struct{}) (<
 		factoryCrd.Traefik().V1alpha1().TLSOptions().Informer().AddEventHandler(eventHandler)
 		factoryCrd.Traefik().V1alpha1().TraefikServices().Informer().AddEventHandler(eventHandler)
 
-		factoryKube := informers.NewSharedInformerFactoryWithOptions(c.clientset, resyncPeriod, informers.WithNamespace(ns))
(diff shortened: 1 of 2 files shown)


Wagum:
I see that the crd client is also using this function call. Should this be also refactored?

dtomcej:
Hello @Wagum,

Thanks for this!

You are correct, we should also investigate using `informers.WithTweakListOptions` to use labelselectors instead of filtering later on.

We should also look at updating the CRD provider to use `NewSharedInformerFactoryWithOptions` too. Ideally this would be in this PR for consistency.

Wagum:
Hey  @dtomcej , 

im going to add the call on `NewSharedInformerFactoryWithOptions` to the k8s CRD provider as well. 

Im concerned about adding `informers.WithTweakListOptions` to the InformerFactory, because the ListOption from TweakListOptionsFunc will apply to all objects, that informerFactory provides. 
(i could not identify Ingress objects inside of the TweakListOptionsFunc).

This implies, that the user needs to add the ingressLabelSelector to the corresponding secrets, endpoints and services inside the k8s cluster as well. 

An other option would be adding a factory for ingresses including TweakListOption, and one dedicated factory for secret, endpoints and services. This dedicated factory should not use the TweakListOption. I think you did something similar in the CRD provider for Kubernetes.  

Im going to think about this again and propose a possible solution to you asap. 
Please let me know, what you think about this after investigating with your team :)

PR #5267, opened 2019-08-30
Improve API for the web UI

ldez:
### What does this PR do?

- add search filters (only in names for now).
- add type field for:
  - HTTP services
  - HTTP middlewares
  - TCP services
- HTTP errors with JSON responses. 
- add `Using` for entry points on routers.
- fix(tcp): Add errors about unknown entryPoint in runtime api 

### Motivation

Required for the web UI.

### More

- [x] Added/updated tests


pkg/config/runtime/runtime_tcp.go
```diff
@@ -28,21 +34,31 @@ func (c *Configuration) GetTCPRoutersByEntryPoints(ctx context.Context, entryPoi
 				entryPointsRouters[entryPointName] = make(map[string]*TCPRouterInfo)
 			}
 
+			entryPointsCount++
+			fmt.Println("dklskmlqsdkmk")
```
mpl:
```suggestion
```

PR #5291, opened 2019-09-03
fix: stripPrefix and stripPrefixRegex.

ldez:
### What does this PR do?

Don't return a 404 when the path is not stripped.

### Motivation

To be able to use  stripPrefix and stripPrefixRegex.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/stripprefixregex/strip_prefix_regex.go
+++ b/pkg/middlewares/stripprefixregex/strip_prefix_regex.go
@@ -53,6 +53,9 @@ func (s *stripPrefixRegex) ServeHTTP(rw http.ResponseWriter, req *http.Request) 
 		parts := exp.FindStringSubmatch(req.URL.Path)
 		if len(parts) > 0 && len(parts[0]) > 0 {
 			prefix := parts[0]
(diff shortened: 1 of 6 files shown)


pkg/middlewares/stripprefixregex/strip_prefix_regex.go
```diff
@@ -46,32 +49,25 @@ func (s *stripPrefixRegex) GetTracingInformation() (string, ext.SpanKindEnum) {
...
-	if s.router.Match(req, &match) {
-		params := make([]string, 0, len(match.Vars)*2)
-		for key, val := range match.Vars {
-			params = append(params, key)
-			params = append(params, val)
-		}
+	for _, exp := range s.expressions {
+		parts := exp.FindStringSubmatch(req.URL.Path)
+		if len(parts) > 0 && len(parts[0]) > 0 {
+			prefix := parts[0]
```
mpl:
```suggestion
			prefix := parts[0]
			if !strings.HasPrefix(req.URL.Path, prefix) {
				continue
			}
```

Issue #4239, opened 2018-11-27
Swarm Mode Cluster tutorial not working


StefanTod:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?
Followed [this tutorial](https://docs.traefik.io/user-guide/swarm-mode/) line-by-line from the website.

### What did you expect to see?
The results shown on the tutorial. Basic connectivity between Treafik and Docker Swarm.

### What did you see instead?
`Failed to connect to 192... port 80: Connection refused`

### Output of `traefik version`: (_What version of Traefik are you using?_)
v1.7.4

```
(paste your output here)
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
None.
```

### If applicable, please paste the log output in DEBUG level (`--logLevel=DEBUG` switch)

```
(paste your output here)
```

ameesme:
I'm experiencing the same issue. Running `curl -H Host:whoami0.traefik http://$(docker-machine ip manager)` just returns a `Failed to connect`.

WokoLiu:
Got same error. I followed [this tutorial](https://docs.traefik.io/user-guide/swarm-mode/) line-by-line from the website. 

At last, when I run `curl -H Host:whoami0.traefik http://$(docker-machine ip manager)` it returned `curl: (7) Failed to connect to 192.168.99.xxx port 80: Connection refused`.

I'm using macOS Mojave 10.14.1, docker version 18.09.0, build 4d60db4, traefik:latest(v1.7.5)

ameesme:
@ldez Is there a place where we can file issues with the documentation? The issue is still present.

WokoLiu:
Still not sure why, but it finally worked, maybe this [Error in driver during machine creation](https://github.com/machine-drivers/docker-machine-driver-xhyve/issues/211)

So I chose to use [xhyve](https://github.com/machine-drivers/docker-machine-driver-xhyve) instead of virturlbox, and [v18.06.1-ce](https://github.com/boot2docker/boot2docker/releases/tag/v18.06.1-ce) instead of v18.09.0.

It works on macOS Mojave 10.14.1
```
$ brew install docker-machine-driver-xhyve
$ wget -O /tmp/boot2docker.iso http://github.com/boot2docker/boot2docker/releases/download/v18.06.1-ce/boot2docker.iso
$ docker-machine create -d xhyve --xhyve-boot2docker-url=/tmp/boot2docker.iso --engine-storage-driver overlay2 manager
```

[...]

PR #5292, opened 2019-09-03
Make documented traefik image version explicit

mootari:
### What does this PR do?

Sets an explicit traefik image tag in the ["Deploy Traefik" section](https://docs.traefik.io/user-guide/swarm-mode/#deploy-traefik) of the swarm-mode user guide.

### Motivation

The current specifier `traefik:<stable version from https://hub.docker.com/_/traefik>` has led to friction (see #4239) as people may accidently reference the incompatible 2.0 tag.

Since this is the 1.7 branch it should be safe to document the 1.7 tag.

### DIFF
--- a/docs/user-guide/swarm-mode.md
+++ b/docs/user-guide/swarm-mode.md
@@ -85,7 +85,7 @@ docker-machine ssh manager "docker service create \
 	--publish 80:80 --publish 8080:8080 \
 	--mount	type=bind,source=/var/run/docker.sock,target=/var/run/docker.sock \
 	--network traefik-net \
-	traefik:<stable version from https://hub.docker.com/_/traefik> \
+	traefik:1.7 \
 	--docker \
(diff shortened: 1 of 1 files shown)


ldez:
The change was not made because of v2 but due the use of old v1 version: https://github.com/containous/traefik/issues/4926#issuecomment-498024932

The change made by the PR #4927 have a main advantage:
- `v1.7` is like if you are using `latest` it's possible to be sure to have the real latest version of the v1 of Traefik.

So I prefer to keep the placeholder instead of `v1.7`

mootari:
@ldez Should I pivot this MR towards dynamically replacing the placeholder `<stable version from https://hub.docker.com/_/traefik>` with the respective branch name when `gh-pages` gets deployed?

ldez:
Thanks but I will declined, we want to keep it simple.

PR #5310, opened 2019-09-05
Improve rate limiter tests

mpl:
### What does this PR do?

This PR rewrites the rate limiter tests in order to hopefully make them more resilient to CPU load, so they don't exhibit as much of a flaky behaviour.
While we were at it, we rewrote them so the input would be more similar to other load testing tools (incoming load + duration), and we added a few more cases.

### Motivation

To improve the tests and make them less flaky.

### More

- [x] Added/updated tests

### Additional Notes


mpl:
one of the tests is still flaky, hence the no-merge for now.

PR #5311, opened 2019-09-06
New API security

juliens:
### What does this PR do?

This PR changes the default behavior of the API. When you just enable the API, it will create a new internal service named `api@internal`, and you can configure a secure router to be able to use the API. 
If you just want to try the API in a development environment, you can enable the insecure API (`api.insecure=true`).

### Motivation
The API contains sensitive data, that's why we want more security on this part.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/static/static_config.go
+++ b/pkg/config/static/static_config.go
@@ -85,6 +85,7 @@ type ServersTransport struct {
 
 // API holds the API configuration
 type API struct {
+	Insecure  bool `description:"Activate API on an insecure entryPoints named traefik." json:"insecure,omitempty" toml:"insecure,omitempty" yaml:"insecure,omitempty" export:"true"`
 	Dashboard bool `description:"Activate dashboard." json:"dashboard,omitempty" toml:"dashboard,omitempty" yaml:"dashboard,omitempty" export:"true"`
 	Debug     bool `description:"Enable additional endpoints for debugging and profiling." json:"debug,omitempty" toml:"debug,omitempty" yaml:"debug,omitempty" export:"true"`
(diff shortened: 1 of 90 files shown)


pkg/config/static/static_config.go
```diff
@@ -85,6 +85,7 @@ type ServersTransport struct {
 
 // API holds the API configuration
 type API struct {
+	Insecure  bool `description:"Activate API on an insecure entryPoints named traefik." json:"insecure,omitempty" toml:"insecure,omitempty" yaml:"insecure,omitempty" export:"true"`
```
mpl:
```suggestion
	Insecure  bool `description:"Activate API directly on the entryPoint named traefik." json:"insecure,omitempty" toml:"insecure,omitempty" yaml:"insecure,omitempty" export:"true"`
```

pkg/provider/rest/rest.go
```diff
@@ -18,6 +18,7 @@ var _ provider.Provider = (*Provider)(nil)
 
 // Provider is a provider.Provider implementation that provides a Rest API.
 type Provider struct {
+	Insecure          bool `description:"Activate REST Provider on an insecure entryPoints named traefik." json:"insecure,omitempty" toml:"insecure,omitempty" yaml:"insecure,omitempty" export:"true"`
```
mpl:
```suggestion
	Insecure          bool `description:"Activate REST Provider directly on the entryPoint named traefik." json:"insecure,omitempty" toml:"insecure,omitempty" yaml:"insecure,omitempty" export:"true"`
```

PR #5314, opened 2019-09-06
Default to CLF when accesslog format is unsupported

mpl:
### What does this PR do?

This PR makes the accesslog middleware default to Common Log Format when the provided format is unsupported (e.g. user made a config mistake). Previously we used to return an error in that case, now we log instead and provide a working middleware with CLF.

### Motivation

To have a sane fallback in case of PEBKAC.

### More

- ~[ ] Added/updated tests~

### DIFF
--- a/pkg/middlewares/accesslog/logger.go
+++ b/pkg/middlewares/accesslog/logger.go
@@ -77,7 +77,7 @@ func NewHandler(config *types.AccessLog) (*Handler, error) {
 	case JSONFormat:
 		formatter = new(logrus.JSONFormatter)
 	default:
-		log.WithoutContext().Warnf("unsupported access log format: %q, defaulting to common format instead.", config.Format)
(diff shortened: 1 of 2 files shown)


pkg/middlewares/accesslog/logger.go
```diff
@@ -77,7 +77,8 @@ func NewHandler(config *types.AccessLog) (*Handler, error) {
 	case JSONFormat:
 		formatter = new(logrus.JSONFormatter)
 	default:
-		return nil, fmt.Errorf("unsupported access log format: %s", config.Format)
+		log.WithoutContext().Warnf("unsupported access log format: %q, defaulting to common format instead.", config.Format)
```
ldez:
```suggestion
		log.WithoutContext().Errorf("unsupported access log format: %q, defaulting to common format instead.", config.Format)
```

PR #5328, opened 2019-09-09
Configurable prefix for statsd metrics

schulterklopfer:
We installed multiple instances of traefik for request seperation. To monitor our setup in a better way, we needed configurable statsd prefixes, so the metrics won't end up in one big bucket :)


ldez:
Hello,

Thank you for your works.

As your PR is an enhancement, we will decline the PR.

PR #5353, opened 2019-09-11
Consider default cert domain in certificate store

nrwiersma:
### What does this PR do?

This allows for the default certificate to be considered when getting all domains from a TLS certificate store.

### Motivation

When getting all domains for a store, the default cert domain should be included in the domains list. This is applicable when generating ACME domains, it should not generate a certificate for a domain that falls under the default domain.

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: Nicolas Mengin <nmengin.pro@gmail.com>

### DIFF
--- a/tls/certificate_store.go
+++ b/tls/certificate_store.go
@@ -56,28 +56,6 @@ func (c CertificateStore) GetAllDomains() []string {
 	return allCerts
 }
 
-func getCertificateDomains(cert *tls.Certificate) []string {
-	var names []string
-
-	if cert == nil {
-		return names
-	}
-
-	x509Cert, err := x509.ParseCertificate(cert.Certificate[0])
-	if err != nil {
-		return nil
-	}
-
-	if len(x509Cert.Subject.CommonName) > 0 {
-		names = append(names, x509Cert.Subject.CommonName)
-	}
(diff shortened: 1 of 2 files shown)


tls/certificate_store.go
```diff
@@ -47,9 +48,36 @@ func (c CertificateStore) GetAllDomains() []string {
...
 	}
+
+	// Get Default certificate
+	if c.DefaultCertificate != nil {
+		allCerts = append(allCerts, c.getCertificateDomains(c.DefaultCertificate)...)
+	}
 	return allCerts
 }
 
+func (c CertificateStore) getCertificateDomains(cert *tls.Certificate) []string {
```
ldez: can be a function instead of a method:
```go
func getCertificateDomains(cert *tls.Certificate) []string {
	if cert == nil {
		return nil
	}

	x509Cert, err := x509.ParseCertificate(cert.Certificate[0])
	if err != nil {
		return nil
	}

	var names []string

	if len(x509Cert.Subject.CommonName) > 0 {
		names = append(names, x509Cert.Subject.CommonName)
	}

	for _, san := range x509Cert.DNSNames {
		names = append(names, san)
	}

	return names
}
```

nrwiersma:
Fixed

Issue #4493, opened 2019-02-13
passTLSCert middleware not modifying headers on kubernetes


rochaporto:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

Followed issue #4048 but still couldn't see the headers being modified and forwarded.

On a test setup with kubernetes + ingress, i tried:

1. curl with a custom CA
2. backend is a golang server that dumps the HTTP request (https://gitlab.cern.ch/rbritoda/go-headers)

### What did you expect to see?

X-Forwarded-Ssl-Client-Cert and X-Forwarded-Tls-Client-Cert-info showing up in my go test implementation with the client cert information.

### What did you see instead?

No such headers appeared.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
kubectl -n kube-system exec ingress-traefik traefik version
traefik version
Version:      v1.7.8
Codename:     maroilles
Go version:   go1.11.5
Built:        2019-01-29_04:33:36PM
OS/Arch:      linux/amd64

kubectl -n kube-system exec ingress-traefik ps auxw
PID   USER     TIME  COMMAND
    1 root      0:08 traefik traefik --api --traefikLog=true --kubernetes --kubernetes.enablepasstlscert=true --logLevel=DEBUG --entrypoints=Name:http Address::80 --entrypoints=Name:https Address::443 TLS
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

Certificate creation:
```
openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout tls.key -out tls.crt -subj "/CN=passtlstest001.cern.ch"
kubectl  create secret tls go-headers-secret --key=tls.key --cert=tls.crt
```

Sample kubernetes deployment + ingress:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: go-headers-deployment
spec:
  selector:
    matchLabels:
      app: go-headers
  replicas: 1
  template:
    metadata:
      labels:
        app: go-headers
    spec:
      containers:
      - name: go-headers
        image: go-headers
        ports:
        - containerPort: 8000
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: go-headers
  annotations:
    kubernetes.io/ingress.class: traefik
    traefik.ingress.kubernetes.io/frontend-entry-points: https
    traefik.ingress.kubernetes.io/pass-client-tls-cert: |
      pem: true
      infos:
        notafter: true
        notbefore: true
[...]

PR #5362, opened 2019-09-12
Add note clarifying client certificate header

bradjones1:
### What does this PR do?

Adds note clarifying client certificate header is not passed through unless it is from a CA specified on the EntryPoint.

### Motivation

This section suggests the client certificate sent is added to the header if the configuration is set on the Ingress; this alone is not sufficient (it must also pass validation at the EntryPoint, which requires separate config.)

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/configuration/backends/kubernetes.md
+++ b/docs/configuration/backends/kubernetes.md
@@ -221,7 +221,10 @@ infos:
 If `pem` is set, it will add a `X-Forwarded-Tls-Client-Cert` header that contains the escaped pem as value.
 If at least one flag of the `infos` part is set, it will add a `X-Forwarded-Tls-Client-Cert-Infos` header that contains an escaped string composed of the client certificate data selected by the infos flags.
 This infos part is composed like the following example (not escaped):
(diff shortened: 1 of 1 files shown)


bradjones1:
Refs https://github.com/containous/traefik/issues/4493

docs/configuration/backends/kubernetes.md
```diff
@@ -218,7 +218,7 @@ infos:
     serialnumber: true
 ```
 
-If `pem` is set, it will add a `X-Forwarded-Tls-Client-Cert` header that contains the escaped pem as value.
+If `pem` is set, it will add a `X-Forwarded-Tls-Client-Cert` header that contains the escaped pem as value. (Note this works only with certificates issued by CAs included in the applicable [EntryPoint ClientCA section](/configuration/entrypoints/#tls-mutual-authentication); certificates from other CAs are not passed through as-is.)
```
jbdoumenjou: As this mechanism applies to the both options, this note could be added after the next sentence and start with something like "These two options applies only...". WDYT ?

bradjones1: Good catch. I've updated the branch accordingly.

Issue #5426, opened 2019-09-19
2.0 is released, is cluster HA  ready ?


Jamlee:
2.0 is released, is cluster HA  ready ?

bradjones1:
My understanding is that while it hasn't been officially summarized all in one place, clustering is a major selling point for EE and may not be supported in CE 2.x, though that would need confirmed itself.

I recently received a note from a Containous sales rep putting it this way:

> TraefikEE is a clustered and highly available version of Traefik CE.
> You are right there were some issues with the HA mode in Traefik 1.X, that's why we deprecated it and decided to work on it on the EE edition.

There are indeed a number of not-obvious bugs in the 1.x HA implementation, especially with respect to ACME/Let's Encrypt: https://github.com/containous/traefik/issues/4851 https://github.com/containous/traefik/issues/3487 https://github.com/containous/traefik/issues/5047 https://github.com/containous/traefik/issues/3833 and likely others. While it's entirely possible to run Traefik 1.x in production with HA, you do need to navigate these not-insignificant issues.

Per [this post](https://community.containo.us/t/please-explain-distributed-lets-encrypt-feature-of-traefikee/872/3) in the discussion forum, I've found the most definitive statements regarding clustering in 2.x, which is that it will be an EE feature.

To be clear, I am not disputing Containous' right to decide what is and is not in their CE vs EE product, but I think it's worth summarizing the information available and perhaps for a note about the loss of HA support in 2.x to be made more prominent in the documentation.

bradjones1:
By way of example where the docs could be updated, there is a note:

> For concurrency reason, this file cannot be shared across multiple instances of Traefik. Use a key value store entry instead.

on https://docs.traefik.io/https/acme/#storage - if this is not an option, it should probably be removed. I opened a PR for this.

SantoDE:
Thanks a lot for your question.

As already mentioned on the Community Forums, (https://community.containo.us/t/please-explain-distributed-lets-encrypt-feature-of-traefikee/872/2), our intent with Traefik 2.0 has been to make Traefik Community Edition a good old stateless data plane again.

We want Traefik to be the best at what it's designed for, which is providing a simple solution to get your applications reachable from the outside world.

[...]

PR #5395, opened 2019-09-16
Add Consul Catalog provider 

negasus:
### What does this PR do?

Add new provider 'Consul Catalog'

### Motivation

Popular feature of Traefik v1
Close #4675

### More

- [x] Added/updated tests
- [x] Added/updated documentation

Issue #5169, opened 2019-07-31
Add a provider for Key Value (KV) Store in Traefik v2


SantoDE:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Traefik v1 supports dynamic config generation for and from a Key Value Store, i.e. it has a kv provider.

Now that the main changes for v2 are out, we can re-add previously supported providers, such as the one for Key Value Stores.

The existing providers, can be used for inspiration.

lokeshpeta:
when can we expect this feature?

bradjones1:
See https://github.com/containous/traefik/issues/5426

lokeshpeta:
Oh good. Thanks very much.

dduportal:
Hi @bradjones1 @lokeshpeta , for information, Traefik v2.0 only drops the "HA" part as explained in #5426 
But the provider "KV", which allows dynamic configuration of Traefik to be stored in KV systems as etcd, consul, etc. is still an upcoming feature.

PR #5433, opened 2019-09-19
Update ACME storage docs to remove reference to KV store in CE

bradjones1:
### What does this PR do?

Remove reference to KV storage for Traefik CE. Per https://github.com/containous/traefik/issues/5426

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -390,7 +390,7 @@ docker run -v "/my/host/acme:/etc/traefik/acme" traefik
 ```
 
 !!! warning
-    For concurrency reason, this file cannot be shared across multiple instances of Traefik. Use a key value store entry instead.
+    For concurrency reason, this file cannot be shared across multiple instances of Traefik. Clustered operation is available in Traefik EE.
 
 ## Fallback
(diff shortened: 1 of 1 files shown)


imyller:
Does this mean that even if Consul store gets implemented for v2 (#5395) the ACME will be prevented from using this KV store in v2 CE?

imyller:
Also, quoting discussion about KV+ACME in v2 CE: https://blog.containo.us/back-to-traefik-2-0-the-online-meet-up-4e4b6411fbfa#4b39

> But hey! You can still use multiple Traefik instances with your favorite KV store, we only removed the unstable and experimental part :)

This seems to imply that there will be nothing preventing from v2 ACME using KV store in CE version.

geraldcroes:
@imyller for now, let's focus on this PR that fixes the documentation. Nothing has yet been decided whether we'll allow ACME to store certificates into KV stores ... if we allow it, we have to make sure that users are aware of the limitations. I think we need a pros and cons of each scenario to make a sound decision. Shall we use #5169 to discuss this?

PR #5434, opened 2019-09-19
Add a warning note regarding optional TLS mutual auth

bradjones1:
### What does this PR do?

Adds a warning to the TLS mutual authentication section regarding non-universal client support for optional client certificates during handshaking.

### Motivation

We got burned by certain clients which do not treat this option as truly optional; see [this Chromium ticket](https://bugs.chromium.org/p/chromium/issues/detail?id=1002638) for an example of this incompatibility in the wild.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/configuration/entrypoints.md
+++ b/docs/configuration/entrypoints.md
@@ -240,7 +240,8 @@ TLS Mutual Authentication can be `optional` or not.
 * If `optional = false`, Traefik will only accept clients that present a certificate signed by a specified Certificate Authority (CA).
 
 !!! warning
-  While the TLS [1.1](https://tools.ietf.org/html/rfc4346#section-7.4.6) and [1.2](https://tools.ietf.org/html/rfc5246#section-7.4.6) RFCs specify that clients should proceed with handshaking by sending an empty list should they have no certs for the CAs specified by the server, not all do so in practice. Use this feature with caution should you require maximum compatibility with a wide variety of client user agents which may not strictly implement these specs.
+    While the TLS [1.1](https://tools.ietf.org/html/rfc4346#section-7.4.6) and [1.2](https://tools.ietf.org/html/rfc5246#section-7.4.6) RFCs specify that clients should proceed with handshaking by sending an empty list should they have no certs for the CAs specified by the server, not all do so in practice.
+    Use this feature with caution should you require maximum compatibility with a wide variety of client user agents which may not strictly implement these specs.
 
 `ClientCAFiles` can be configured with multiple `CA:s` in the same file or use multiple files containing one or several `CA:s`.
 The `CA:s` has to be in PEM format.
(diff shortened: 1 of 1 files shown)


jbdoumenjou:
Hi @bradjones1, 

Could you give more details on your difficulties and use cases to be sure to understand everything?

bradjones1:
@jbdoumenjou thanks for the response. The issue linked in the original post sums up a usual case pretty well; not all browsers properly implement the TLS 1.1 and 1.2 standards regarding client certificate CAs expressed as supported by the server. Per the specs, the client either `SHOULD` (1.1) or `MUST` return an empty set of client certificates (since they have none that match) and the server can choose to continue the handshaking ("optional" mode in Traefik.)

Not all browsers - in our experience older browsers on mobile, and even some newer ones - properly implement this flow. In the case of Chrome on Android, because Android < 10 does not reveal client certificate storage info to the app, the user is always presented with a scary-looking certificate request. If you cancel it out, the handshaking proceeds, but this is bad UX particularly for customer-facing sites.

In our experience, hosting many client web sites that serve the broad public, this UA incompatibility on many devices means we had to back out of using client certificate validation, as it meant a measurable decrease in traffic from affected devices/UAs.

Does that help?

jbdoumenjou:
Thank you for this clarification.
Have you tried v2.0? Mutual authentication is now [configurable](https://docs.traefik.io/v2.0/https/tls/#client-authentication-mtls)

bradjones1:
We are evaluating 2.0 but no I have not yet had a chance to and we are on 1.x for the time being. I think this is helpful to add to the documentation as it caused quite a surprise for us and I don't see much elsewhere about the risk of implementing this with respect to these mobile browsers.

docs/configuration/entrypoints.md
```diff
@@ -239,11 +239,13 @@ TLS Mutual Authentication can be `optional` or not.
 * If `optional = true`, if a certificate is provided, verifies if it is signed by a specified Certificate Authority (CA). Otherwise proceeds without any certificate.
 * If `optional = false`, Traefik will only accept clients that present a certificate signed by a specified Certificate Authority (CA).
 
+!!! warning
```
mpl: You messed up the formatting, so the output does not look like a note (check it out in https://deploy-preview-5434--traefik-doc.netlify.com/configuration/entrypoints/).
I think you need a blank line after L242, and then you need to indent the whole paragraph. Have a look at the other warning note at the end of that file.

mpl: You messed up the formatting, so the output does not look like a note (check it out in https://deploy-preview-5434--traefik-doc.netlify.com/configuration/entrypoints/).
I think you need a blank line after L242, and then you need to indent the whole paragraph. Have a look at the other warning note at the end of that file.

PR #5465, opened 2019-09-23
Add tls option for Elliptic Curve Preferences

FlDutchman:
### What does this PR do?

This PR adds the a tls option for setting Curve Preferences of the server, which is then passed on to the go crypto library, see: `CurvePreferences []CurveID` at [https://godoc.org/crypto/tls#Config](https://godoc.org/crypto/tls#Config).

### Motivation

At default go crypto uses all available elliptic curves in the following order `x25519, secp256r1, secp384r1, secp521r1`. Some user may want to change this. I'm no expert myself but according to SSLLabs.com it can increase the key exchange security. See the attached pictures

Before PR:
curves_default

After PR with following settings:
```
tls:
  options:
    test:
      minVersion: VersionTLS12
      curvePreferences:
        - secp521r1
        - secp384r1
      cipherSuites:
        - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
        - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
```
curves_custom

### More

- [ ] Added/updated tests

### Additional Notes

The CurvePreferences option is equivalent to the CipherSuites option. Since there was no test for that, I did not see the need for new tests now. 

This is the first time of me contributing to this project. Please correct me if I'm overseeing something :)

### DIFF
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -226,6 +226,47 @@ spec:
     With TLS 1.3, the cipher suites are not configurable (all supported cipher suites are safe in this case).
     <https://golang.org/doc/go1.12#tls_1_3>
 
+### Curve Preferences
+
+See [CurveID](https://godoc.org/crypto/tls#CurveID) for more information. With this option you 
+can set the prefered elliptic curves. This option is order specific. You can use the names of 
+the curves defined by crypto (e.g. CurveP521) and the RFC defined names (e. g. secp521r1).
+
+```toml tab="File (TOML)"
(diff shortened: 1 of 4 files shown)


ldez:
Hello,

Thanks for your contribution.

In the branch `v2.0`, as it's a stable release, we only accept bug fixes and documentation improvements.

So could you rebase (not merge) on the branch `master`, thanks.

FlDutchman:
Reopened, see #5466

PR #5466, opened 2019-09-23
Add tls option for Elliptic Curve Preferences

ksarink:
### What does this PR do?

This PR adds the a tls option for setting Curve Preferences of the server, which is then passed on to the go crypto library, see: `CurvePreferences []CurveID` at [https://godoc.org/crypto/tls#Config](https://godoc.org/crypto/tls#Config).

### Motivation

At default go crypto uses all available elliptic curves in the following order `x25519, secp256r1, secp384r1, secp521r1`. Some user may want to change this. I'm no expert myself but according to SSLLabs.com it can increase the key exchange security. See the attached pictures

Before PR:
curves_default

After PR with following settings:
```
tls:
  options:
    test:
      minVersion: VersionTLS12
      curvePreferences:
        - secp521r1
        - secp384r1
      cipherSuites:
        - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
        - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
```
curves_custom

### More

- [ ] Added/updated tests

### Additional Notes

The CurvePreferences option is equivalent to the CipherSuites option. Since there was no test for that, I did not see the need for new tests now. 

(Almost) same as #5465 but now PR into master

This is the first time of me contributing to this project. Please correct me if I'm overseeing something :)


pkg/tls/tlsmanager.go
```diff
@@ -231,6 +231,20 @@ func buildTLSConfig(tlsOption Options) (*tls.Config, error) {
 		}
 	}
 
+	// Set the list of CurvePreferences/CurveIDs if set in the config TOML
+	if tlsOption.CurvePreferences != nil {
+		conf.CurvePreferences = make([]tls.CurveID, 0)
+		// if our list of CipherSuites is defined in the entryPoint config, we can re-initialize the suites list as empty
```
mpl: That comment needs fixing

FlDutchman: Must have missed that one. Fixed it

ldez:
Design LGTM even if
```
    // CurvePreferences contains the elliptic curves that will be used in
    // an ECDHE handshake, in preference order. If empty, the default will
    // be used. The client will use the first preference as the type for
    // its key share in TLS 1.3. This may change in the future.
    CurvePreferences []CurveID
```
> This may change in the future.

I do not know what level of confidence we can have on this option.

PR #5490, opened 2019-09-25
fix: update linter.

ldez:
### What does this PR do?

fix: update linter.

### Motivation

Allow to use some interesting new linters

### More

- [x] Added/updated tests


ldez:
I will revert the linter update because of an excessive memory consumption (https://github.com/golangci/golangci-lint/issues/337) but I will keep the linted files in the PR.

PR #5516, opened 2019-09-27
fix: default passHostHeader for file provider.

ldez:
### What does this PR do?

default passHostHeader for file provider.

### Motivation

Have homogeneous behavior between all providers.

### More

- [x] Added/updated tests


sirlatrom:
While I appreciate the added consistency, this fix broke our setup today because we had not explicitly specified this setting and did not expect changes in default values in a patch release.

PR #5524, opened 2019-09-27
Mention both ACME and LE in documentation 

dduportal:
### What does this PR do?

This PR update the title of the documentation page related to ACME/Let's Encrypt configuration.

### Motivation

There are still a lot of newcomer users who don't know ACME and LEt's encrypt.
We want the documentation to be clear enough for them.

### More

- ~[ ] Added/updated tests~

### Additional Notes

image

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -1,4 +1,4 @@
-# Let's Encrypt
+# ACME / Let's Encrypt
 
 Automatic HTTPS
 {: .subtitle }
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -93,7 +93,7 @@ nav:
   - 'HTTPS & TLS':
       - 'Overview': 'https/overview.md'
       - 'TLS': 'https/tls.md'
-      - 'Let''s Encrypt': 'https/acme.md'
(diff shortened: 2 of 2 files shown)


dduportal:
Right, but then we should adapt to stay coherent:

- The page URL is `https://docs.traefik.io/v2.0/https/acme/` => It says "ACME".
- The title of the page is currently "Let's Encrypt" (in the page and on the table of content" => It says "Let's Encrypt".
- The page starts with `You can configure Traefik to use an ACME provider (like Let's Encrypt) for automatic certificate generation.` => It clearly explains that Let's Encrypt is using the ACME protocol.
- The first configuration example says "enable" ACME.

So it is not clear if the page is about ACME or Let's Encrypt. 

By setting the title to "ACME /" first, the goal is to make it clear that it's about ACME,
considering that Let's Encrypt is an ACME implementation.

Alternatively, we could ensure that this page is about ACME protocol only, and add another page which demonstrates how to use Traefik and Let's Encrypt together as an applied use case.

dduportal:
As discussed together: we are closing this issue, as this subject will be worked on as soon as the next ACME provider other than Let's Encrypt will be tested and documented in Traefik.

PR #5534, opened 2019-09-28
 doc: add example to specify certResolver

contrun:
### What does this PR do?

This PR add instructions on how to specify certResolver on http.router.

### Motivation

I was walking in the dark for quite some time. This PR could help anyone who is also wondering how to make acme works in v2.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/migration/v1-to-v2.md
+++ b/docs/content/migration/v1-to-v2.md
@@ -226,6 +226,10 @@ Then, a [router's TLS field](../routing/routers/index.md#tls) can refer to one o
         # will terminate the TLS request
         [http.routers.Router-1.tls]
           options = "myTLSOptions"
+          certResolver = "myCertResolver"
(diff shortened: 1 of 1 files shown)


docs/content/migration/v1-to-v2.md
```diff
@@ -226,6 +226,10 @@ Then, a [router's TLS field](../routing/routers/index.md#tls) can refer to one o
         # will terminate the TLS request
         [http.routers.Router-1.tls]
           options = "myTLSOptions"
+          certResolver = "myCertResolver"
```
ldez: This example is for illustrate how to use external certificate, so it's not the good place for your example.

contrun: Thanks for your links. When I saw `options = "myTLSOptions"`, I thought it would be appropriate to add also `certResolver = "myCertResolver"`.

PR #5539, opened 2019-09-28
Add overview to API documentation

lnxbil:
### What does this PR do?

Add the missing `/api/overview` to the list with a description what I think it does.

### Motivation

Ran into a problem where I wanted just the API calls of traefik beeing routed through the `/api` ingress and found out, that the entry was missing from the documentation.

### More

### DIFF
--- a/docs/content/operations/api.md
+++ b/docs/content/operations/api.md
@@ -159,25 +159,25 @@ api:
 
 All the following endpoints must be accessed with a `GET` HTTP request.
 
-| Path                           | Description                                                                               |
-|--------------------------------|-------------------------------------------------------------------------------------------|
-| `/api/http/routers`            | Lists all the HTTP routers information.                                                   |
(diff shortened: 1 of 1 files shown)


docs/content/operations/api.md
```diff
@@ -173,6 +173,7 @@ All the following endpoints must be accessed with a `GET` HTTP request.
 | `/api/tcp/services/{name}`     | Returns the information of the TCP service specified by `name`.                           |
 | `/api/entrypoints`             | Lists all the entry points information.                                                   |
 | `/api/entrypoints/{name}`      | Returns the information of the entry point specified by `name`.                           |
+| `/api/overview`                | Returns statistic information about http and tcp aswell as enabled features and providers |
```
jbdoumenjou:
```suggestion
| `/api/overview`                | Returns statistic information about http and tcp as well as enabled features and providers |
```

PR #5549, opened 2019-09-30
fix: return an error instead of panic.

ldez:
### What does this PR do?

fix: return an error instead of panic.

### Motivation

Don't panic.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/file/raw_node.go
+++ b/pkg/config/file/raw_node.go
@@ -83,7 +83,7 @@ func decodeRaw(node *parser.Node, vData reflect.Value, filters ...string) error 
 						values = append(values, val)
 					}
 				default:
-					return fmt.Errorf("field %s use unsupported slice type: %s", child.Name, item.Kind().String())
+					return fmt.Errorf("field %s uses unsupported slice type: %s", child.Name, item.Kind().String())
 				}
 			}
 
@@ -94,7 +94,7 @@ func decodeRaw(node *parser.Node, vData reflect.Value, filters ...string) error 
 				return err
 			}
 		default:
-			return fmt.Errorf("field %s use unsupported type: %s", child.Name, value.Kind().String())
+			return fmt.Errorf("field %s uses unsupported type: %s", child.Name, value.Kind().String())
 		}
 
 		node.Children = append(node.Children, child)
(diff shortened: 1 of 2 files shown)


pkg/config/file/raw_node.go
```diff
@@ -63,40 +71,52 @@ func decodeRaw(node *parser.Node, vData reflect.Value, filters ...string) {
...
-						values = append(values, getSimpleValue(sValue))
+						val, err := getSimpleValue(sValue)
+						if err != nil {
+							return err
+						}
+						values = append(values, val)
 					}
 				default:
-					panic("Unsupported slice type: " + item.Kind().String())
+					return fmt.Errorf("field %s use unsupported slice type: %s", child.Name, item.Kind().String())
```
mpl:
```suggestion
					return fmt.Errorf("field %s uses unsupported slice type: %s", child.Name, item.Kind().String())
```

pkg/config/file/raw_node.go
```diff
@@ -63,40 +71,52 @@ func decodeRaw(node *parser.Node, vData reflect.Value, filters ...string) {
...
 			child.Value = strings.Join(values, ",")
 		case reflect.Map:
-			decodeRaw(child, value)
+			err := decodeRaw(child, value)
+			if err != nil {
+				return err
+			}
 		default:
-			panic("Unsupported type: " + value.Kind().String())
+			return fmt.Errorf("field %s use unsupported type: %s", child.Name, value.Kind().String())
```
mpl:
```suggestion
			return fmt.Errorf("field %s uses unsupported type: %s", child.Name, value.Kind().String())
```

pkg/config/file/raw_node.go
```diff
@@ -63,40 +71,52 @@ func decodeRaw(node *parser.Node, vData reflect.Value, filters ...string) {
...
+		return strconv.FormatUint(item.Uint(), 10), nil
 	case reflect.Float32, reflect.Float64:
-		return strings.TrimSuffix(strconv.FormatFloat(item.Float(), 'f', 6, 64), ".000000")
+		return strings.TrimSuffix(strconv.FormatFloat(item.Float(), 'f', 6, 64), ".000000"), nil
 	case reflect.Bool:
-		return strconv.FormatBool(item.Bool())
+		return strconv.FormatBool(item.Bool()), nil
 	default:
-		panic("Unsupported Simple value type: " + item.Kind().String())
+		return "", fmt.Errorf("unsupported Simple value type: %s", item.Kind().String())
```
mpl:
```suggestion
		return "", fmt.Errorf("unsupported simple value type: %s", item.Kind().String())
```

PR #5574, opened 2019-10-03
Fix Location response header http to https when SSL

elielgoncalves:
When SSL is enabled the Location response header is not replaced to HTTPS, causing an unnecessary redirect.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -1909,11 +1909,11 @@
 
 [[projects]]
   branch = "v1"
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

The vendor cannot be edited by hand, you have to fix this on the library repo https://github.com/unrolled/secure

Thanks.

elielgoncalves:
Hello,

This bug has already been fixed in vendor, can you updated in traefik?

Best

PR #5581, opened 2019-10-04
Add Agent Cache support to consulCatalog Provider

megalan247:
## What does this PR do?

This PR adds support for consul request caching on a consul agent, allowing a local agent to offer a cache of all the services in the catalog while still validating the cache is up to date. This is particularly useful on large deployments to avoid overloading consul servers. We found that when over 100 instances of traefik were deployed this reduced the RPC requests to consul by orders of magnitude. 

### Motivation

After deploying this feature internally, and testing it with thousands of consul services, I would like to submit it upstream! 

### More

- [ ] Added/updated tests

### Additional Notes

I know that v2 is out now but at the time of writing it doesn't have consul catalog support so I hope this will be OK to add as a new feature.


dtomcej:
Hello @megalan247,

This PR looks promising, especially since we just merged ConsulCatalog support into master! (https://github.com/containous/traefik/pull/5581)

Do you think that you could rebase this branch onto master, and apply it to the new consul catalog provider?

Thanks!

megalan247:
Hey! Unfortunately I don't have the time at the moment to try and integrate this feature into v2, could this feature still be merged into v1.7?

Issue #5562, opened 2019-10-01
Matcher: PathPrefixStrip in 2.0 


ptr1120:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Documentation on how the matcher PathPrefixStrip [Documentaion v1.7](https://docs.traefik.io/v1.7/basics/#matchers) can be used in traefik version 2.

ldez:
Hello,

`PathPrefixStrip` has been replaced by the rule [`PathPrefix`](https://docs.traefik.io/rv2.0/outing/routers/#rule) and the middleware [`stripPrefix`](https://docs.traefik.io/v2.0/middlewares/stripprefix/)

Issue #5408, opened 2019-09-18
How to migrate v1 rule to a v2 routing version?


gerbil:
All my services got simple v1 working frontend rule:

_"traefik.frontend.rule=PathPrefix:/test;ReplacePathRegex: ^/test/(.*) /$$1"_

Is there any example doc describing how to migrate this rule to a v2 routing version?

I've tried _PathPrefix(`/test`)_, _PathPrefix(`/test/`)_, _PathPrefix(`/test/*`_) ... etc but all internal calls like for /test/folder/folder/script.js are not working!

traefiker:
Hi! I'm Træfiker :robot: the bot in charge of communication regulation.

Thanks for your interest in Traefik!

We dedicate the issue tracker to bug reports and feature requests only. My advanced AI has spotted that your issue might be a configuration problem or relates to something that doesn't look like a bug.

To confirm this, please join our [Community Forum](https://community.containo.us/) and reach out to us on the [Traefik section](https://community.containo.us/c/traefik).

In case I'm wrong (well, that would be embarrassing :sweat_smile:), my developers will re-open the issue and fix me!

In the meantime, you can double check [Traefik's documentation](https://docs.traefik.io/).

PR #5600, opened 2019-10-08
Migration guide: pathprefixstrip migration

dduportal:
### What does this PR do?

This PR introduces the following changes on the migration guide of documentation:

- [chore] Remove empty whitespaces
- [chore] Fix title casings to fit a "Chicago" US casing (ref. https://capitalizemytitle.com/)
- [enhancement] Add a section about transforming URL path prefix, to educate v1 users to split responsibilities (1 router catch the request with the rule, and then middleware transform request)

### Motivation

#5562 #5408 

We want to improve the user experience of users migrating from v1

### More

- ~[ ] Added/updated tests~

### Additional Notes

Nope


dduportal:
Update: whitespace are NOT trailed when inside code block. Thanks @ldez for pointing this .

PR #5605, opened 2019-10-09
Update apiVersion in documentation descriptor

pyaillet:
### What does this PR do?

Update apiVersion in documentation descriptor

### Motivation

The old extensions/v1beta1 Deployment API is [deprecated](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/). 

### More

- [X] Updated documentation

### DIFF
--- a/docs/content/user-guides/crd-acme/03-deployments.yml
+++ b/docs/content/user-guides/crd-acme/03-deployments.yml
@@ -6,7 +6,7 @@ metadata:
 
 ---
 kind: Deployment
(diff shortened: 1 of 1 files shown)


pyaillet:
@dduportal I greped through the directory and didn't find other references to extensions/v1beta1 for Deployments.

However there are many references to extensions/v1beta1 Ingress which should be updated for Kubernetes 1.20, but as the networking.k8s.io/v1beta1 Ingress is only served since 1.14 I think this could be done later.

Issue #5622, opened 2019-10-10
Using several times a cert floods logs


sylr:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

We use a wildcard certificates that is duplicated in each namespaces it is needed for in our kubernetes cluster (v1.13.6).

In the logs traefik writes a lot of warning saying that the certificates already exists for the default entryPoint (shouldn't it be certificate store instead of entryPoint ?).

```
{"level":"warning","msg":"Skipping addition of certificate for domain(s) \"*.domain.com,domain.com\", to EntryPoint default, as it already exists for this Entrypoint.","time":"2019-10-10T09:44:32Z"}
{"level":"warning","msg":"Skipping addition of certificate for domain(s) \"*.domain.com,domain.com\", to EntryPoint default, as it already exists for this Entrypoint.","time":"2019-10-10T09:44:32Z"}
{"level":"warning","msg":"Skipping addition of certificate for domain(s) \"*.domain.com,domain.com\", to EntryPoint default, as it already exists for this Entrypoint.","time":"2019-10-10T09:44:32Z"}
{"level":"warning","msg":"Skipping addition of certificate for domain(s) \"*.domain.com,domain.com\", to EntryPoint default, as it already exists for this Entrypoint.","time":"2019-10-10T09:44:32Z"}
{"level":"warning","msg":"Skipping addition of certificate for domain(s) \"*.domain.com,domain.com\", to EntryPoint default, as it already exists for this Entrypoint.","time":"2019-10-10T09:44:32Z"}
{"level":"warning","msg":"Skipping addition of certificate for domain(s) \"*.domain.com,domain.com\", to EntryPoint default, as it already exists for this Entrypoint.","time":"2019-10-10T09:44:32Z"}
{"level":"warning","msg":"Skipping addition of certificate for domain(s) \"*.domain.com,domain.com\", to EntryPoint default, as it already exists for this Entrypoint.","time":"2019-10-10T09:44:32Z"}
{"level":"warning","msg":"Skipping addition of certificate for domain(s) \"*.domain.com,domain.com\", to EntryPoint default, as it already exists for this Entrypoint.","time":"2019-10-10T09:44:32Z"}
{"level":"warning","msg":"Skipping addition of certificate for domain(s) \"*.domain.com,domain.com\", to EntryPoint default, as it already exists for this Entrypoint.","time":"2019-10-10T09:44:32Z"}
[...]

PR #5641, opened 2019-10-11
Use debug for log about skipping addition of cert

sylr:
### What does this PR do?

This entry can flood the logs if a cert is duplicated like wildcard certs can be.

### Motivation

Using debug for this can lower the amount of logs in INFO mode. (https://github.com/containous/traefik/issues/5622)

### More

- [ ] Added/updated tests


sylr:
Hi guys 👋 

Could we merge this please ? I'd appreciate it.

Thanks.

PR #5644, opened 2019-10-13
chore(node/webui): update node to 12.11 version

Slashgear:
### What does this PR do?

- Update Node to a up to date version
- Update readme of the webui

### Motivation

Keep træfik up to date ❤️ 

### More

- [ ] Added/updated tests

### Additional Notes

- The webui is not concerned by the breaking changes of NodeJS between 8 and 12.
- This will normally boost the build time for the webui
- *Node version 8 will be a dead version by the end of the year https://github.com/nodejs/Release*

### DIFF
--- a/exp.Dockerfile
+++ b/exp.Dockerfile
@@ -1,5 +1,5 @@
 # WEBUI
-FROM node:8.15.0 as webui
+FROM node:12.11 as webui
 
 ENV WEBUI_DIR /src/webui
 RUN mkdir -p $WEBUI_DIR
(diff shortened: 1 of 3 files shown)


exp.Dockerfile
```diff
@@ -1,13 +1,13 @@
...
+FROM node:12.11 as webui
 
 ENV WEBUI_DIR /src/webui
 RUN mkdir -p $WEBUI_DIR
 
 COPY ./webui/ $WEBUI_DIR/
 
 WORKDIR $WEBUI_DIR
-RUN yarn install
+RUN npm install
```
SuperSandro2000: Why the switch from yarn to npm?

ldez: because the new webui uses npm.
This file was a previous webui relicate.

Issue #5649, opened 2019-10-14
Support for MaxVersion parameter in tls.Options


kmeekva:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Would like to be able to set MaxVersion in the tls.Options section of the traefik.toml similarly to how you can set MinVersion.  The MaxVersion options is not supported.

golang tls library has an option **MaxVersion** in the tls.Options config. https://golang.org/pkg/crypto/tls/#Config

**Use Case:**
While testing traefik v2.0.1 using client certificate authentication, we found some chrome clients had difficulty negotiating tls connections using the same tls options (ciphers) as we used successfully with traefik v.1.7.  The problem appears to be related to the attempted negotiation of   TLS13 downgrading  to TLS12  during negotiations, and chrome closing the connection.

To test, I created simple golang tls http server and verified that when I set MaxVersion to VersionTLS12  it connects successfully ( not attempting the TLS13 in the negotiation).  But if MaxVersion is set to TLS13 it tries TLS13 and then fails to downgrade to a TLS12 connection.  

I would like to be able to set the MaxVersion tls.option in traefik to work around this problem.

I have a pull request I will submit.

ldez:
Closed by #5650

PR #5650, opened 2019-10-14
Add support for MaxVersion in tls.Options

kmeekva:
### What does this PR do?
Adds support for maxVersion to the tls.Options 

### Motivation
Some Chrome clients still have issues with negotiations associated with TLS Version 13.  
Having the ability to set Maximum can avoid this problem.

### More

- [ ] Added/updated tests

### Additional Notes
This is my first contribution to traefik, so let me know if I've done this correctly. 
see #5649


pkg/tls/tlsmanager.go
```diff
@@ -217,6 +217,12 @@ func buildTLSConfig(tlsOption Options) (*tls.Config, error) {
 		conf.MinVersion = minConst
 	}
 
+	// Set the maximum TLS version if set in the config TOML
+	if maxConst, exists := MaxVersion[tlsOption.MaxVersion]; exists {
+		conf.PreferServerCipherSuites = true
```
ldez: Could you explain this line?

kmeekva: I was just trying to replicate the way the minVersion option is handled.  

I used same code as block above that checks if tlsOptions.MinVersion is set in config and adds to the conf structure. 

That code set the PreferServerCipherSuites to true so I kept that as well.

That should probably be separately configurable I would think.

The PreferServerCipherSuites setting can be removed and still achieve my objective.

ldez: thank you, you can keep this line.

SuperSandro2000:
I think a note which discourages this setting should be added to not disable TLS1.3. The right approach would be to encourage the clients to update.

PR #5660, opened 2019-10-15
Adding support for HTTPs backends with Kubernetes ExternalName services

kpeiruza:
### What does this PR do?

This PR adds support for HTTPS with Kubernetes ExternalName, for both CRD and Ingress Class.

### Motivation

Trying to configure a backend with HTTPS, traefik hardcodded "http://" in front of the URL

### Additional Notes

I've already tested it with docker image kpeiruza/traefik:2.0 and is working. Dunno why the build has failed with travis-ci, it said something about documentation!!!!


kpeiruza:
Please check the code, it's my first Golang attempt. I've used a new variable to determine if the protocol should be http or https. In the original code, there was only support for http.

I've already  checked it and it's working as expected: docker -> kpeiruza/traefik:2.0

PR #5681, opened 2019-10-17
Update apiVersion

ialidzhikov:
Ref https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#deprecations

### More

- [ ] Added/updated tests


ldez:
Hello,

For now, `networking.k8s.io/v1beta1` is not really supported, so we don't want to change the doc.

PR #5706, opened 2019-10-22
Fix Security Headers Doc

FlorianPerrot:
Headers configuration not valid with `SSLRedirect` or `FrameDeny`  should be lowercase.


dtomcej:
@FlorianPerrot,

Good catch! Can you update your PR to ensure that they are camel cased? (not lowercase) as in the reference documentation:

https://docs.traefik.io/reference/dynamic-configuration/file/

PR #5709, opened 2019-10-23
Feature/grpc middleware

negasus:
#### What does this PR do?

Append middleware 'grpc'
Allow send Request to remote client over grpc and process Actions from client response

Actions:
- set request headers
- remove request headers
- set response headers
- cancel request with set response code and body

### Motivation

For example, issue https://github.com/containous/traefik/issues/1336

### More

- [ ] Added/updated tests

### Additional Notes

Example client handler:
```
func (client *TraefikGrpcClient) Handle(ctx context.Context, req *api.Request) (*api.Response, error) {
	response := &api.Response{
		ClientID: "clientID",
		Actions: &api.Actions{
			SetRequestHeaders: map[string]string{
				"X-Request-Header": "foo",
			},
			RemoveRequestHeaders: []string{
				"Authorization",
			},
			SetResponseHeaders: map[string]string{
				"X-Response-Header": "bar",
			},
			CancelRequest: &api.CancelRequest{
				Code: 403,
				Body: "Forbidden",
			},
		},
	}

	return response, nil
}
```

**Propose to discussion**


negasus:
Questions:

1. Multiple client addresses with load balancing
2. Obtain client address from service discovery, for example - consul
3. ???

ldez:
Hello,

I will give the same answer on the previous PR on the same topic: https://github.com/containous/traefik/pull/2362#issuecomment-356886024

thanks anyway

PR #5718, opened 2019-10-25
change logs to Debug

### LINKED ISSUE #5693: make "Skipping same configuration for provider docker" level DEBUG instead of INFO
feature / change of logging level of a specific message

instead of the current behaviour, which floods my INFO-logging with:
time="..." level=info msg="Skipping same configuration for provider docker" providerName=docker

i would hope that this becomes a DEBUG-level to stop spamming an INFO-log with non-information every 15 seconds...

jairovsky:
### What does this PR do?
Change two calls from `logger.Infof` to `logger.Debugf`.
Fixes #5693.

### Motivation
Less pollution in everyday logs.

### DIFF
--- a/pkg/server/server_configuration.go
+++ b/pkg/server/server_configuration.go
@@ -212,12 +212,12 @@ func (s *Server) preLoadConfiguration(configMsg dynamic.Message) {
 	}
 
 	if isEmptyConfiguration(configMsg.Configuration) {
-		logger.Infof("Skipping empty Configuration for provider %s", configMsg.ProviderName)
+		logger.Debugf("Skipping empty configuration for provider %s", configMsg.ProviderName)
 		return
 	}
 
 	if reflect.DeepEqual(currentConfigurations[configMsg.ProviderName], configMsg.Configuration) {
-		logger.Infof("Skipping same configuration for provider %s", configMsg.ProviderName)
+		logger.Debugf("Skipping same configuration for provider %s", configMsg.ProviderName)
 		return
 	}
 


ldez:
Hello,

I recommend to use the level `WARN` (the default level) instead of `INFO` in your configuration.

jnovack:
> I recommend to use the level `WARN` (the default level) instead of `INFO` in your configuration.

Sure, that's possible, but I believe the argument on the table is the message in question is a `DEBUG` message, rather than an `INFO` message.

RFC 5424 does not define what a level is, just says the following on page 10.

```
              6       Informational: informational messages
              7       Debug: debug-level messages
```

Other charts and references around the web detail further (see references below), and there's no clear-cut line.  If we aggregate all the other definitions, we can generally agree that `INFO` means "something happened" or "something changed", and `DEBUG` and `TRACE` become catch-alls for documenting program flow.

In this instance, I would argue that nothing changed, the configuration is the same, there is no need to tell me everything is the same, it's what I *EXPECT*.  If I'm expecting it, it's not informational.

`WARN`, `ERROR` and `FATAL` are easy to define, but the argument is always about `INFO` or `DEBUG`.  To me, the difference between `INFO` and `DEBUG` are "things that users care about" and "things that developers care about".  When viewed in that light, `INFO` and `DEBUG` become clear.

The question on the table would then be, *"Are 'skipping' messages, `INFO` or `DEBUG`?"*

I'm not a member, I'm not a contributor (yet), I can't answer the above question, but I can tell you, looking through my logs, that the messages provide me no useful information.

-------
**References**

Level | Description
-- | --
DEBUG | Designates fine-grained informational events that are most useful to debug an application.
INFO | Designates informational messages that highlight the progress of the application at coarse-grained level.

----------

https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels

* Debug - Information that is diagnostically helpful to people more than just developers (IT, sysadmins, etc.).
* Info - Generally useful information to log (service start/stop, configuration assumptions, etc). Info I want to always have available but usually don't care about under normal circumstances. This is my out-of-the-box config level.

-------

*INFO*
[[source](https://reflectoring.io/logging-levels/)]

The *INFO* level should be used to document state changes in the application or some entity within the application.  

Concrete examples for using the *INFO* level are:

* the application has started with configuration parameter x having the value y
* a new entity (e.g. a user) has been created or changed its state
* the state of a certain business process (e.g. an order) has changed from “open” to “processed”
* a regularly scheduled batch job has finished and processed z items.

-------

Apache Log Levels
[[source](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/Level.html)

* The DEBUG Level designates fine-grained informational events that are most useful to debug an application.
* The INFO level designates informational messages that highlight the progress of the application at coarse-grained level.

ldez:
Hello,

First, thank you @jnovack for the references.

The message `Skipping empty configuration for provider` is more important for a user than a dev.

The log levels in Traefik are:
- `ERROR`
- `WARN`
- `INFO`
- `DEBUG`

There are no `TRACE` level.

It's not comparable to Syslog, HTTPD, or SLF4J.

So I think `INFO` is the good level for this log.

ldez:
Related to my previous comment, I will decline.

Thanks anyway.

PR #5726, opened 2019-10-25
Use debug for log about skipping addition of cert

sylr:
### What does this PR do?

This entry can flood the logs if a cert is duplicated like wildcard certs can be.

### Motivation

Backport https://github.com/containous/traefik/pull/5641 in v1.7 tree.

### More

- [ ] Added/updated tests

### DIFF
--- a/tls/certificate.go
+++ b/tls/certificate.go
@@ -192,7 +192,7 @@ func (c *Certificate) AppendCertificate(certs map[string]map[string]*tls.Certifi
 		}
 	}
 	if certExists {
-		log.Warnf("Skipping addition of certificate for domain(s) %q, to EntryPoint %s, as it already exists for this Entrypoint.", certKey, ep)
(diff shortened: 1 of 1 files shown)


sylr:
FYI, this is what happened when we deployed the v2.0 patched version:

Search___Splunk_7_0_11_1

ldez:
Hello,

I recommend to use the level `WARN` instead of `INFO`.

sylr:
Hi

> I recommend to use the level WARN (the default level) instead of INFO in your configuration.

But if I do this I would lose access logs which use the INFO level, wouldn't I ?

ldez:
The log level is for [Traefik log](https://docs.traefik.io/v1.7/configuration/logs/#traefik-logs).

The [access logs](https://docs.traefik.io/v1.7/configuration/logs/#access-logs) don't have a level.

ldez:
As much as this change made sense in the v2 because of the management of the TLS on the routers, as much on the v1 that does not seem to me relevant.

So we will declined.

Thanks anyway.

PR #5737, opened 2019-10-28
Fix rate limiting and SSE

sylr:
### Motivation

While using long lasting SSE connections the mutex is not released
until the connection is closed. This caused rate limiter to throttle
any SSE endpoint to 1 service connection.

### What does this PR do?

Replacing the defer call that releases the mutex by a direct call
at the appropriate location removes the contention.

### More

- [ ] Added/updated tests


sylr:
After taking a deeper look I believe that the `bucketsMu` is unnecessary as `ttlmap.TtlMap` implements is own locking.

sylr:
Thanks @juliens!

PR #5738, opened 2019-10-28
Let metrics libs handle the atomicity

sylr:
### What does this PR do?

Let metrics libs handle the atomicity.

### Motivation

Removes an atomic lock that the libs already implement.

Removes a defer call, the less the better until go 1.14.

### More

- [ ] Added/updated tests


pkg/middlewares/metrics/metrics.go
```diff
@@ -81,17 +80,14 @@ func (m *metricsMiddleware) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
-		openConns := atomic.AddInt64(&m.openConns, -1)
-		m.openConnsGauge.With(labelValues...).Set(float64(openConns))
-	}(labels)
+	m.openConnsGauge.With(labels...).Add(1)
 
 	start := time.Now()
 	recorder := &responseRecorder{rw, http.StatusOK}
 	m.next.ServeHTTP(recorder, req)
 
+	m.openConnsGauge.With(labels...).Add(-1)
```
sylr: I've got a strange behaviour with request being rate limited by traefik.

It's like they are not passing here so the gauge is not decremented.

pkg/middlewares/metrics/metrics.go
```diff
@@ -78,23 +77,22 @@ func WrapServiceHandler(ctx context.Context, registry metrics.Registry, serviceN
...
 	labels = append(labels, m.baseLabels...)
 
-	openConns := atomic.AddInt64(&m.openConns, 1)
-	m.openConnsGauge.With(labels...).Set(float64(openConns))
-	defer func(labelValues []string) {
-		openConns := atomic.AddInt64(&m.openConns, -1)
-		m.openConnsGauge.With(labelValues...).Set(float64(openConns))
-	}(labels)
+	m.openConnsGauge.With(labels...).Add(1)
+	defer m.openConnsGauge.With(labels...).Add(-1)
```
sylr: I restored the `defer` call instead of moving `m.openConnsGauge.With(labels...).Add(-1)` after `m.next.ServeHTTP(recorder, req)` because I've a scenario where, when doing a load testing of an SSE endpoint with rate limiting activated, we never go past `m.next.ServeHTTP(recorder, req)` and the Gauge is never decremented.

@mpl if you have an idea why I would be really interested.

mpl: well, that suggests m.next.ServeHTTP never returns, either because of a deadlock or a blocked channel, does it not?

pkg/middlewares/metrics/metrics.go
```diff
@@ -78,23 +74,22 @@ func WrapServiceHandler(ctx context.Context, registry metrics.Registry, serviceN
 }
 
 func (m *metricsMiddleware) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	labels := []string{"method", getMethod(req), "protocol", getRequestProtocol(req)}
+	// Initialize labels slice with correct size
+	labels := make([]string, 0, 6+len(m.baseLabels))
```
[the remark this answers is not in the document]
sylr: Because line 79 add 4 entries and line 90 adds 2 more.

I'll add comments about it.

pkg/middlewares/metrics/metrics.go
```diff
@@ -78,23 +74,25 @@ func WrapServiceHandler(ctx context.Context, registry metrics.Registry, serviceN
 }
 
 func (m *metricsMiddleware) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	labels := []string{"method", getMethod(req), "protocol", getRequestProtocol(req)}
+	// Initialize labels slice with correct size -> 6 + len(m.baseLabels)
+	labels := make([]string, 0, 6+len(m.baseLabels))
```
mpl:
```suggestion
	labels := make([]string, 0, 2*len([]string{"method", "protocol", "code"})+len(m.baseLabels))
```
then it is self-documented, and we'll never make the mistake of adding new labels while forgetting to update that hardcoded 6.

ratelimit.toml
```diff
@@ -0,0 +1,47 @@
+[global]
```
mpl: Did you add that file on purpose to the PR?

pkg/middlewares/metrics/metrics.go
sylr: 👍

ratelimit.toml
sylr: Nope

sylr:
@ldez why did you remove the attempt to create the label slice with the correct size ?

PR #5760, opened 2019-10-31
Add missing "-" in traefik configuration example file

graille:
### What does this PR do?

A "-" was missing to match with the configuration example at https://docs.traefik.io/routing/entrypoints/

### Motivation

When you follow the tutorial on the website, the entrypoint "web-secure" is used, in this file, the entrypoint is "websecure", which does not match with website examples.

### DIFF
--- a/traefik.sample.toml
+++ b/traefik.sample.toml
@@ -25,7 +25,7 @@
   [entryPoints.web]
     address = ":80"
 
-  [entryPoints.websecure]
+  [entryPoints.web-secure]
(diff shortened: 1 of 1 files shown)


dduportal:
Hi @graille, after a discussion with the rest of the team, we want to solve this `'websecure` naming challenge from the root: we're going to open a subsequent PR that will homogenise the naming everywhere.

Thanks a lot for this proposal! We're closing the issue with this in mind :)

Issue #5506, opened 2019-09-26
Traefik cannot handle HTTP & HTTPS in one route


SeaLife:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

In Traefik 2.0 i need a hell of docker labels on my containers to achieve what was simple in v1.7.

In my case im thinking of the routers and how they handle SSL and HTTP (no SSL) now.

What i expected to happen when i configured my service like this:

```
  [...]
    labels:
      - "traefik.http.services.traefik.loadbalancer.server.port=8080"
      - "traefik.http.routers.traefik.rule=Host(`traefik.r3ktm8.de`)"
      - "traefik.http.routers.traefik.service=traefik"
      - "traefik.http.routers.traefik.entrypoints=https,http"
      - "traefik.http.routers.traefik.middlewares=ssl"
      - "traefik.http.routers.traefik.tls.certresolver=le"

      - "traefik.http.middlewares.ssl.redirectscheme.scheme=https"
```

is traefik listening to my route on :80 and :443 and accepting a TLS connection on 443 and a normal HTTP connection on :80.

Since the TLS option was moved to a "certificateResolver" traefik does not know what entrypoint is a TLS entrypoint...

But now i need atleast 3 more labels to achieve the redirect scheme to get active on HTTP connection:

```
      - "traefik.http.services.traefik.loadbalancer.server.port=8080"
      - "traefik.http.routers.traefik.rule=Host(`traefik.r3ktm8.de`)"
      - "traefik.http.routers.traefik.service=traefik"
      - "traefik.http.routers.traefik.entrypoints=https"
      - "traefik.http.routers.traefik.middlewares=ssl"
      - "traefik.http.routers.traefik.tls.certresolver=le"

      - "traefik.http.routers.traefik_http.middlewares=ssl"
      - "traefik.http.routers.traefik_http.entrypoints=http"
      - "traefik.http.routers.traefik_http.rule=Host(`traefik.r3ktm8.de`)"

      - "traefik.http.middlewares.ssl.redirectscheme.scheme=https"
```

I need 9 Labels to get a simple routing with SSL redirection.. (+1 label for the middleware, but this can be re-used, so - no issue there)

In 1.7 i only had to add those lables:

```
#!/bin/bash
docker run [...] \
        -l "traefik.admin.frontend.rule=Host:traefik.r3ktm8.de" \
        -l "traefik.admin.port=8080" \
[...]

PR #5766, opened 2019-11-03
Update router entrypoint example

woto:
Fixes problem described in [comment 545077143](https://github.com/containous/traefik/issues/5506#issuecomment-545077143) and [comment 549100716](https://github.com/containous/traefik/issues/5506#issuecomment-549100716) which can be mislead users.

### What does this PR do?

Updates docker.md

### Motivation

[comment 545077143](https://github.com/containous/traefik/issues/5506#issuecomment-545077143)
[comment 549100716](https://github.com/containous/traefik/issues/5506#issuecomment-549100716)

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/routing/providers/docker.md
+++ b/docs/content/routing/providers/docker.md
@@ -165,7 +165,7 @@ For example, to change the rule, you could add the label ```traefik.http.routers
     See [entry points](../routers/index.md#entrypoints) for more information.
 
     ```yaml
-    - "traefik.http.routers.myrouter.entrypoints=web,websecure"
+    - "traefik.http.routers.myrouter.entrypoints=web,another-web"
     ```
 
 ??? info "`traefik.http.routers.<router_name>.middlewares`"
--- a/docs/content/routing/providers/docker.md
+++ b/docs/content/routing/providers/docker.md
@@ -165,7 +165,7 @@ For example, to change the rule, you could add the label ```traefik.http.routers
     See [entry points](../routers/index.md#entrypoints) for more information.
 
     ```yaml
-    - "traefik.http.routers.myrouter.entrypoints=web,another-web"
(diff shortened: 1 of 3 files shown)


woto:
Hi, thank you for reply. Difficult to add something new to the comments above  :(
As i said previously documentation may mislead user as he (also as me) may think that one route may work both with http and https.

dduportal:
The difficult point of the request is that this section of the documentation is aimed to be a reference documentation, not a tutorial. The expectation are different for the reader:

- Reference documentation should provide always up-to-date reference for configuration items, without context.
- Tutorial / User Guide should give a context and show a working example using items from the reference.

My understanding is that the issue to be solved is having a good tutorial/get started/user guide to explain that Traefik v2.0 makes you defining one router for HTTP and another for HTTPS, each on being attached to a single entrypoint.

Based on this explanation (I tried to explain my thought as much I can, please feel free to ask for better explaantion if not clear) here is a proposal (to be discussed), which focus on the definition of the label instead of the meaning of the example:

```text
??? info "`traefik.http.routers.<router_name>.entrypoints`"
    
    Specify the [entry points](../routers/index.md#entrypoints) to attach the [router](../routers/index.md) named <router_name> to. Please note that one or many entry points can be specified, using a comma (`,`) delimited list.

    See [entry points](../routers/index.md#entrypoints) for more information.

    ```yaml tab="Attach router `myrouter` to a single entry point"
    # Attache to the entry point named "web"
    "traefik.http.routers.myrouter.entrypoints=web"
    ```

    ```yaml tab="Attach router `myrouter` to multiple entry points"
    # Attache to the entry points "web-internal" and "web-external"
    "traefik.http.routers.myrouter.entrypoints=web-internal,web-external"
    ```
```

WDYT?

PR #5806, opened 2019-11-09
fix: stripPrefix middleware with empty resulting path.

ldez:
### What does this PR do?

Add an option to allow to bypass the buggy behavior of the `stripPrefix` middleware.

### Motivation

 when the resulting striped path is empty, the path must be empty and not  `/`

### More

- [x] Added/updated tests


pkg/middlewares/stripprefix/strip_prefix.go
```diff
@@ -20,18 +20,20 @@ const (
 
 // stripPrefix is a middleware used to strip prefix from an URL request.
 type stripPrefix struct {
-	next     http.Handler
-	prefixes []string
-	name     string
+	next       http.Handler
+	prefixes   []string
+	forceSlash bool // TODO Must be removed (breaking), the default behavior must be forceSlash=false
```
jbdoumenjou: Why don't we use the deprecated flag ?

pkg/middlewares/addprefix/add_prefix.go
```diff
@@ -41,23 +41,35 @@ func New(ctx context.Context, next http.Handler, config dynamic.AddPrefix, name
...
 	req.RequestURI = req.URL.RequestURI()
 
-	ap.next.ServeHTTP(rw, req)
+	a.next.ServeHTTP(rw, req)
+}
+
+func ensureLeadingSlash(str string) string {
+	if str == "" {
+		return str
+	}
```
mpl: It bothers me that this case is a lie with respect to the function name.

pkg/middlewares/stripprefix/strip_prefix.go
```diff
@@ -59,10 +61,25 @@ func (s *stripPrefix) serveRequest(rw http.ResponseWriter, req *http.Request, pr
 	s.next.ServeHTTP(rw, req)
 }
 
-func getPrefixStripped(s, prefix string) string {
-	return ensureLeadingSlash(strings.TrimPrefix(s, prefix))
+func (s *stripPrefix) getPrefixStripped(urlPath, prefix string) string {
+	if s.forceSlash {
+		// Only for compatibility reason with the previous behavior,
+		// but the previous behavior is wrong.
+		// This need to be remove in the next breaking version.
```
mpl:
```suggestion
		// This needs to be removed in the next breaking version.
```

docs/content/middlewares/stripprefix.md
```diff
@@ -85,3 +85,84 @@ If your backend is serving assets (e.g., images or Javascript files), chances ar
...
+```
+
+```yaml tab="Kubernetes"
+apiVersion: traefik.containo.us/v1alpha1
+kind: Middleware
+metadata:
+  name: example
+spec:
+  stripPrefix:
+    prefixes:
```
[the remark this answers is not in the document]
ldez: fixed

pkg/middlewares/addprefix/add_prefix.go
ldez: `ensureLeadingSlashIfThePathIsNotEmpty` `ensureLeadingSlashIfNeeded` :smile:

docs/content/middlewares/stripprefix.md
```diff
@@ -85,3 +85,85 @@ If your backend is serving assets (e.g., images or Javascript files), chances ar
...
+        prefixes:
+          - "/foobar"
+        forceSlash: false
+```
+
+The `forceSlash` option makes sure that the resulting stripped path is not the empty string, by replacing it with `/` when necessary.
+
+This option was added to keep the initial (non-intuitive) behavior of this middleware, in order to avoid introducing a breaking change.
+
+It's recommended to explicitly set `forceSlach` to `false`.
```
dtomcej:
```suggestion
It's recommended to explicitly set `forceSlash` to `false`.
```

dtomcej:
Should each middleware have its own `func ensureLeadingSlash(str string) string`?

Is that something that could be provided by a base middleware utility/tools package? just so that we don't end up with one middleware behaving differently?

mpl:
> Should each middleware have its own `func ensureLeadingSlash(str string) string`?
> 
> Is that something that could be provided by a base middleware utility/tools package? just so that we don't end up with one middleware behaving differently?

If we do that, we should have a func (name) that does not mislead about what it actually does :-)

pkg/middlewares/addprefix/add_prefix.go
mpl: the second one SGTM

ldez: The both was a joke :smile: 

The method is not `addLeadingSlash` but `ensureLeadingSlash`, so for me it's already explicit.

ldez:
for now I will not factorize.

PR #5817, opened 2019-11-12
Wrong endpoint on the TLS secret example

yacinelazaar:
This snippet provided in that section uses 'web' as an endpoint instead of 'websecure'.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/routing/providers/kubernetes-crd.md
+++ b/docs/content/routing/providers/kubernetes-crd.md
@@ -95,7 +95,7 @@ metadata:
 
 spec:
   entryPoints:
-    - web
+    - websecure
   routes:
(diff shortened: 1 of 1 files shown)


dduportal:
Hi @yacinelazaar , if my understanding of https://traefik.slack.com/archives/C0CDT22PJ/p1573560589045800 is correct: the snippet you changed is referenced in https://docs.traefik.io/v2.0/routing/providers/kubernetes-crd/#middleware, and does not enable TLS: so the entry point has to be `web` (it is the port 80).

If you meant https://docs.traefik.io/routing/providers/kubernetes-crd/#tls, can you add `websecure` to the entry points list (and not replace the existing one) please?

dduportal:
Also, could you rebase this PR to the branch `v2.0` please?

PR #5873, opened 2019-11-20
Fix finding proper provided certificate when ACME is enabled

yazd:
### What does this PR do?

Fixes domain to certificate matching when ACME is enabled and non-acme certificates are provided via configuration.

Currently the domain to certificate matching algorithm [here](https://github.com/containous/traefik/blob/f68b6294693b01fd18fcd1f1dd5818be6d0c2c85/acme/acme.go#L611) fails and the default certificate is served. The result is that if two or more non-ACME certificates are provided, only the first one is used.
HTTPS requests to domains not covered by the default certificate fail.

### More

- [ ] Added/updated tests

### Additional Notes

I am not aware of the project's architecture, or whether the place where I added the call to `BuildNameToCertificate` is the best place for it.

### DIFF
--- a/tls/certificate.go
+++ b/tls/certificate.go
@@ -120,6 +120,7 @@ func (c *Certificates) CreateTLSConfig(entryPointName string) (*tls.Config, erro
 			}
 		}
 	}
+	config.BuildNameToCertificate()
 	return config, nil
 }
 
--- a/tls/certificate_test.go
+++ b/tls/certificate_test.go
@@ -1,0 +1,103 @@
+package tls
+
+import (
+	"crypto/rand"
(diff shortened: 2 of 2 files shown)


yazd:
To ease review, `a.TLSConfig.NameToCertificate` is empty [here](https://github.com/containous/traefik/blob/f68b6294693b01fd18fcd1f1dd5818be6d0c2c85/acme/acme.go#L601) if `BuildNameToCertificate` ([docs](https://golang.org/pkg/crypto/tls/#Config.BuildNameToCertificate)) is not called, which means matching will never occur, only the default certificate will be provided.

It is a one-line change.
Can someone review please?

Issue #4881, opened 2019-05-16
http+unix URL support for backends


sloonz:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Nginx allows reverse proxying to HTTP servers that do not listen on a TCP socket, but an Unix socket (`upstream backend { server unix:/tmp/backend.sock; }`).

Listening on Unix sockets is extremely useful in non-containerized environments (like the file provider), since it doesn’t require port allocation management, they honor filesystem ACLs, and they don’t risk raw port exposure on the internet in case of firewall/listening interface misconfiguration (in architectures when access control is the responsibility of the reverse proxy). 

This has already been mentionned #1396 but the issue has been closed due to a misunderstanding : @dtomcej and @ldez thought it was just an Unix variant of raw TCP reverse proxying (#10), but it is not : it is still HTTP reverse proxying, just on a different transport.

For reference, there is an `http.RoundTripper` implementation for doing HTTP roundtrip on Unix transport : https://github.com/tv42/httpunix/

dtomcej:
@sloonz You are correct! When we closed #1396, most of the feature requests were centred around implementing layer 4 functionality. Now that Traefik v2 supports L4 and L7, this feature becomes even more interesting and powerful. Thank you for pointing this out!

hostmit:
I vote up this feature! I've been waiting for it quite a while.

fsdrw08:
Any update of this request?

fsdrw08:
I would like to expose podman unix socket by traefik, by default, podman socket only serve as a unix socket.  
I understand that podman socket is able to serve as a background service to listen system tcp connection, but I just want to keep podman as lean as possible.

shoenig:
Another use case - the imminent Nomad 1.5 release contains a [Task API](https://github.com/hashicorp/nomad/blob/v1.5.0-rc.1/website/content/api-docs/task-api.mdx#task-api) feature - whereby any task configured with a new `identity` block is given a unique workload idenitty and a Unix domain socket with which to communicate with the Nomad API.

We had begun looking into replacing this [NGINX reverse proxy](https://developer.hashicorp.com/nomad/tutorials/manage-clusters/reverse-proxy-ui) guide with one based on Traefik leveraging the new feature, but this issue is a blocker.

rtribotte:
Hello @shoenig,

We understand that you want to expose Nomad's WebUI with Traefik, using this new Task API.
[...]

PR #5888, opened 2019-11-24
Unix socket support for services

sloonz:
### What does this PR do?

This adds Unix domain socket support for services, both on HTTP/1 and HTTP/2.

### Motivation

Copy-paste from https://github.com/containous/traefik/issues/4881 requesting the feature : 

Listening on Unix sockets is extremely useful in non-containerized environments (like the file provider), since it doesn’t require port allocation management, they honor filesystem ACLs, and they don’t risk raw port exposure on the internet in case of firewall/listening interface misconfiguration (in architectures when access control is the responsibility of the reverse proxy).

### More

- [x] Added/updated tests

### Additional Notes

What is missing is TLS (for HTTP 1 and HTTP 2) over Unix sockets. Given that unix sockets are necessarily on the same machine, I don’t see any possible use-case for it.


hostmit:
Have this been tested already? Merge soon?

PackElend:
hi,
will it be merged into main?

use case is here
https://meta.discourse.org/t/difference-between-socket-and-port-based-connection-to-outer-nginx/60071
UNIX socket is preferred over TCP socket

juliens:
Sorry for the delay.

Thank you for your interest in the project and the time you took to complete this PR!

I have talked with other maintainers about this feature, and to be honest, we are not really "fans" of it because we think it doesn't fit well in a cluster environment.

Although we understand that sometimes "the UNIX socket is preferred to the TCP socket", we believe that Traefik should not have direct access to this socket.

I will decline this PR for now, but if other use cases appear in the linked issue, we will be happy to consider reopening it.

Razikus:
Use case:
1. Container needs access to host network, and exposes api
2. api can't be grabbed by traefik, cause it's bridged
3. can't make it working without workarounds, REALLY easy to make with unix sockets

maxried:
TCP sockets are extremely slow compared with UNIX sockets as the add unnecessary overhead.

rytsh:
There is another way to do that, write custom plugin and call unix socket with http.

PR #5902, opened 2019-11-27
Document LE caveats with Kubernetes on v2

### LINKED ISSUE #5792: v2 & Let's Encrypt in an HA setup
Feature

By design, Traefik 2.0 is stateless and HA: Users can deploy multiple instances to ensure that at least one Traefik instance can route requests at any time.

Though, when using Let's Encrypt for automatic certificate generation, the certificate negotiation cannot be consistently achieved because there is no guarantee that the initiator of the negotiation gets the subsequent calls. 

Traefik 1.X used to leverage K/V stores & elect leaders that would be responsible for the certificate negotiations. Still, many issues were left unsolved ( #4851 #3487 #5047 #3833) because of a flawed design. In 2.0, we chose to remove the feature to keep Traefik a simple and elegant data plane (#5426).

Thanks to third-party software (like CertManager), using Let's Encrypt in an HA setup is still possible, and we need to document how.

dtomcej:
### What does this PR do?

This PR improves the documentation for LetsEncrypt use by making note of the HA caveats, and providing workarounds for known design choices.

### Motivation

Documentation was not clear on the removal of distributed LetsEncrypt, and how it would affect kubernetes users.

### More

- [x] Added/updated tests - Not needed, documentation update

Fixes #5792

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -84,6 +84,13 @@ If there are less than 30 days remaining before the certificate expires, Traefik
 !!! info ""
     Certificates that are no longer used may still be renewed, as Traefik does not currently check if the certificate is being used before renewing.
 
+## Using LetsEncrypt with Kubernetes
+
+When using LetsEncrypt with kubernetes, there are some known caveats with both the [ingress](../providers/kubernetes-ingress.md) and [crd](../providers/kubernetes-crd.md) providers.
+
+!!! info ""
+    If you intend to run multiple instances of Traefik with LetsEncrypt, please ensure you read the sections on those provider pages.
+
 ## The Different ACME Challenges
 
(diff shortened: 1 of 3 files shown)


Coksnuss:
> Unfortunately, it is not possible to run multiple instances of Traefik 2.0 with LetsEncrypt enabled, because there is no way to ensure that the correct instance of Traefik will receive the challenge request, and subsequent responses.

Couldn't one use a dedicated DNS entry that resolves to a dedicated traefik instance within a HA setup (e.g. traefik-master.myzone.tld)? Furthermore, this instance could be configured differently from the other instances. IMO the problem is not with routing the challenge requests to the correct Traefik instance. It's that traefik does (currently) not allow to share the received certificates with other instances (in read-only mode)

Tolsto:
@dtomcej You mention that there is currently work being done on integrating cert-manager with Traefik CRDs. Can you share what the current status is there?

PR #5910, opened 2019-11-28
Web UI: Take off logic from generic table component

matthieuh:
### What does this PR do?

Take off logic from generic table component

### Motivation

Some display logics are in the generic MainTable vue component (like list of columns based on a type prop).


ldez:
Waiting for the merge back of v2.1 in master

PR #5919, opened 2019-12-02
Fix documented separator between client certificates

seiffert:
### What does this PR do?

This PR updates the documentation of the PassTLSClientCert middleware to document the correct separator used in the presence of multiple client certificates.

### Motivation

We've implemented parsing of the "X-Forwarded-Tls-Client-Cert" header based on the documentation and realized that multiple certificates weren't correctly split.

### More

- [ ] Added/updated tests

### Additional Notes

See
https://github.com/containous/traefik/blob/cf1ace3a73d971e0bd5a0b52ab4040e60a4e0ddb/pkg/middlewares/passtlsclientcert/pass_tls_client_cert.go#L272

### DIFF
--- a/docs/content/middlewares/passtlsclientcert.md
+++ b/docs/content/middlewares/passtlsclientcert.md
@@ -380,7 +380,7 @@ In the example, it is the part between `-----BEGIN CERTIFICATE-----` and `-----E
 !!! info "Extracted data"
     
     The delimiters and `\n` will be removed.  
-    If there are more than one certificate, they are separated by a "`;`".
+    If there are more than one certificate, they are separated by a "`,`".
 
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

Thank you :+1: 

We think that is more a bug with the separator because the 2 headers don't use the same separator.

Take a look to #5921

seiffert:
Hi @ldez,

Thank you for your feedback! I think that makes sense. If you merge #5921, please note that [this line](https://github.com/containous/traefik/pull/5919/files#diff-83587bc227c44b30705bb92802e049d6R403) in the documentation will also need to be adjusted. While we will not be affected by this because we don't currently use the info header, I want to note that this change will break compatibility for applications that parse the info header.

traefiker:
Closed by #5921.

PR #5946, opened 2019-12-05
Fix target branch for Traefik v2

hartwork:
### What does this PR do?

Update the pull request template to ask people to target branch `v2.1` for Traefik v2 changes, instead of `v2.0`.

### Motivation

I was presented the template mentioning `v2.0` and it seemed out of date and I found it confusing and worth a bump (or adding more text to say why not `v2.1` should be targeted). Happy to adjust as needed.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,11 +3,11 @@ PLEASE READ THIS MESSAGE.
 
 Documentation fixes or enhancements:
 - for Traefik v1: use branch v1.7
-- for Traefik v2: use branch v2.0
+- for Traefik v2: use branch v2.1
 
 Bug fixes:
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

The current stable version is v2.0.x, so the template is correct:
- bug fixes (code) must be applied on the `v2.0` branch.
- documentation (fix or enhancement) must be applied on the `v2.0` branch.
- new features (enhancement) must be done on the `master` branch

Thanks anyway

hartwork:
Would you accept a pull request appending ` (latest stable, not v2.1!)`?

ldez:
We want to use explicit branch names not just "latest stable".

The v2.1 is in RC phase, so if someone want to fix a specific v2.1, he can use the branch `v2.1`.

The template goal's is to drive the majority of the "bug-fixer" to the good branch.

PR #5952, opened 2019-12-06
Don't throw away valid configuration updates

zaphod42:
### What does this PR do?

The configuration reloading code takes various actions in order to limit the number of reloads that have to be performed.
One of those actions is to check if the new configuration is the same as the current configuration. 
If the new configuration is the same, then there is no need to apply it and it can be safely ignored.

However, the check was performed at the wrong location.
The location assumed that the `c.currentConfiguration`s variable is up to date.
This assumption does not hold given the throttling and asynchronous handling of configuration updates.
The effect of this incorrect assumption is that a configuration update can be thrown away and leave a prior configuration in place.

This resolves the problem by moving the check to the `loadMessage` method.
This is the point where the `c.currentConfigurations` variable is modified and so there is no chance that the new configuration will be compared against an incorrect version.

### Motivation

Fixes #5901 

### More

- [x] Added/updated tests
- [ ] Added/updated documentation

PR #5954, opened 2019-12-06
Don't throw away valid configuration updates

zaphod42:
### What does this PR do?

The configuration reloading code takes various actions in order to limit
the number of reloads that have to be performed. One of those actions is
to check if the new configuration is the same as the current
configuration. If the new configuration is the same, then there is no
need to apply it and it can be safely ignored.

However, the check was performed at the wrong location. The location
assumed that the c.currentConfigurations variable is up to date. This
assumption does not hold given the throttling and asynchronous handling
of configuration updates. The effect of this incorrect assumption is
that a configuration update can be thrown away and leave a prior
configuration in place.

This resolves the problem by moving the check to the loadMessage method.
This is the point where the c.currentConfigurations variable is modified
and so there is no chance that the new configuration will be compared
against an incorrect version.

Note: the test was deleted in this branch as the behaviour has been
moved causing the existing test to break. The refactor of this code on
the master branch made it amenable to a more comprehensive unit test.
However, this branch is not.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/server/server_configuration.go
+++ b/pkg/server/server_configuration.go
@@ -32,6 +32,12 @@ import (
 func (s *Server) loadConfiguration(configMsg dynamic.Message) {
 	currentConfigurations := s.currentConfigurations.Get().(dynamic.Configurations)
 
+	if reflect.DeepEqual(currentConfigurations[configMsg.ProviderName], configMsg.Configuration) {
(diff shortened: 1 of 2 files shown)


ldez:
Closed in favor of #5952

PR #5957, opened 2019-12-06
Proxy Protocol support on TCP Receiver side

mohansella:
### What does this PR do?

This Pull Request contains a simple implementation of [Proxy Protocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) on the receiver side of TCP Proxy.

### Motivation

While i was switching our LBs, from Nginx to Traefik, I found that the current version (v2.0.6) doesn't support proxy protocol but only on the LB frontend. For security reasons, we need the client's IP as well as port information in our TCP implementation (which can sense proxy protocol). So I have implemented a basic work version as per my need with public usage in concern. 

Limitations:

 - Proxy Protocol Chaining is not implemented.
 - Version 2 (binary format) is not implemented. 

To Reviewers:
  I don't have prior experience in Golang. So, there may be changes needed due to language performance or roadmap.
 
Usage:
  Enable the flag `proxyProtocol` under the required service in dynamic conf `tcp.services`. Sample below. 
```
tcp:
  services:
    myservicename:
      proxyProtocol: true
      loadBalancer:
        servers:
        - address: "MyAppServer:8080"
```


juliens:
Thank you @mohansella for your contribution, this is really appreciated.

But, I spoke with the team, and we think that to be consistent with the http part, ProxyProtocol should take place on the router as middleware, and not on the service with a new field.

So I will close this PR and invite you to create a proposal if you want to talk a little more. If after discussion of the proposal, we decide that your implementation seems correct, we can always reopen this PR.

PR #5958, opened 2019-12-06
Changed case in redirectscheme

cdemi:
It looks like redirectscheme is case sensitive and needs to be with a small s


ldez:
Hello,

TOML and YAML are case sensitive, the valid syntax is `redirectScheme` with a upper `S`.

https://github.com/containous/traefik/blob/829649e905700ba6f80313e466362c0dd65f32a4/pkg/config/dynamic/middlewares.go#L28

cdemi:
For me, the only use case that worked was with a small S for some reason
________________________________

Issue #4598, opened 2019-03-14
Support docker ssh daemon endpoint


frejonb:
### Do you want to request a *feature* or report a *bug*?
Feature

### Suggestion

Docker supports connecting to a daemon via ssh from 18.09: https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-socket-option 

Implementing this would solve the security issues of mounting the Unix socket, as well as being able to run on worker.

### What did you expect to see?
Support for ssh protocol

### What did you see instead?
Only TCP or Unix

cakiwi:
While the documentation suggests that a user can run the docker provider with a providers.docker.endpoint using ssh the reality is the container is missing the ssh binary or a go ssh connector.

https://community.traefik.io/t/failing-to-specify-an-ssh-endpoint-for-docker-provider/9738

nodje:
+1

yvess:
I solved it like that, mounting root ssh folder and create a custom `entrypoint.sh`:

```yaml
# docker-compose.yml
services:
  traefik:
    image: traefik:v2.9
   ...
    volumes:
     ...
      - ./ssh:/root/.ssh/
      - ./entrypoint.sh:/entrypoint.sh
```

```bash
# entrypoint.sh
#!/bin/sh
set -e

# add openssh for ssh docker connection
apk add openssh

# first arg is `-f` or `--some-option`
if [ "${1#-}" != "$1" ]; then
    set -- traefik "$@"
fi

# if our command is a valid Traefik subcommand, let's invoke it through Traefik instead
# (this allows for "docker run traefik version", etc)
if traefik "$1" --help >/dev/null 2>&1
then
    set -- traefik "$@"
else
    echo "= '$1' is not a Traefik command: assuming shell execution." 1>&2
fi

exec "$@"
```

emilevauge:
closed by #5969

PR #5969, opened 2019-12-09
Support SSH connection to Docker

sh7dm:
### What does this PR do?
Add support for connecting to Docker daemon via SSH.
Some code was copied from Docker CLI: https://github.com/docker/cli/blob/master/cli/context/docker/load.go#L112-L118

### Motivation
#4598

### More

- [ ] Added/updated tests

### Additional Notes
Not sure how to test this. Probably a complex integration test is required (container with Docker-dind + SSH). I tested it locally and it works.


sh7dm:
What's this check?
image

pkg/provider/docker/docker.go
```diff
@@ -109,58 +109,69 @@ type networkData struct {
...
+		client.WithVersion(apiVersion),
+		client.WithHTTPClient(httpClient),
+		client.WithHTTPHeaders(httpHeaders),
+	)
+}
+
+func (p *Provider) getHTTPClient() (*http.Client, string, error) {
+	helper, err := connhelper.GetConnectionHelper(p.Endpoint)
+	if err != nil {
+		return nil, "", err
```
sh7dm: I think if we experience SSH URL parsing error we should continue, shouldn't we?

sh7dm: @ldez looks like it can't return error if we've got correct URL. https://github.com/docker/cli/blob/8547dfcff72fd65426ef71a5bf949ac310ee3419/cli/connhelper/connhelper.go#L24

sh7dm: If scheme isn't `ssh`, the only error that can be returned is error in last line, that's not defined: https://github.com/docker/cli/blob/8547dfcff72fd65426ef71a5bf949ac310ee3419/cli/connhelper/connhelper.go#L44
[url.Parse](https://godoc.org/net/url#URL.Parse) shouldn't produce errors if URL is HTTP(S) or any other. Sorry for mis-interpreting your code.

docs/content/providers/docker.md
```diff
@@ -145,6 +145,7 @@ You can specify which Docker API Endpoint to use with the directive [`endpoint`]
         - Accounting at container level, by exposing the socket on a another container than Traefik's.
           With Swarm mode, it allows scheduling of Traefik on worker nodes, with only the "socket exposer" container on the manager nodes.
         - Accounting at kernel level, by enforcing kernel calls with mechanisms like [SELinux](https://en.wikipedia.org/wiki/Security-Enhanced_Linux), to only allows an identified set of actions for Traefik's process (or the "socket exposer" process).
+        - SSH public key authentication (SSH is supported with Docker > 18.09)
```
dtomcej: Can we give an example of what this looks like?

is it `ssh://user@example.com:22`?

We provide an example for TCP, but we should for SSH too!

sh7dm: Okay, will do, just a sec.

sh7dm: @dtomcej  Where should that be demonstrated? also, you're right about URL, `ssh://u@h` is absolutely right.

dtomcej: https://github.com/containous/traefik/blob/master/docs/content/providers/docker.md#endpoint

sh7dm: @dtomcej as a comment?
```yaml tab="File (YAML)"
providers:
  docker:
    # endpoint: "ssh://user@example.com"
```

sh7dm: Ahh, okay
image

sh7dm: @dtomcej Done!

pkg/provider/docker/docker.go
```diff
@@ -108,46 +109,75 @@ type networkData struct {
...
+			Transport: &http.Transport{
+				DialContext: helper.Dialer,
+			},
+		}
+
+		return []client.Opt{
+			client.WithHTTPClient(httpClient),
+			client.WithHost(helper.Host), // To avoid 400 Bad Request: malformed Host header daemon error
+			client.WithDialContext(helper.Dialer),
+		}, nil
```
juliens: as on `client.WithHTTPClient` you only change the `DialContext`, you can use `client.WithDialContext` only.

sh7dm: Really? I thought `WithHTTPClient` says that Docker should connect via given HTTP

antonedvard:
This feature is not available inside the Traefik Docker Image as it doesn't find SSH in $PATH...

kobalteule:
any news on this? I am trying to get traefik with ssh as endpoint running instead of using docker.sock

this is the error message I get:

level=error msg="Provider connection error error during connect: Get \"http://docker/v1.24/version\": exec: \"ssh\": executable file not found in $PATH, retrying in 4.459624189s" providerName=docker

this is the config snippet from traefik.yml:

providers:
  docker:
    endpoint: "ssh://user@localhost"
    watch: true
    exposedByDefault: false
    network: "traefik"

ldez:
>  exec: \"ssh\": executable file not found in $PATH

The error means that the commad `ssh` is not found in the directories referenced by the `PATH` env var.
So you have to put the `ssh` command into a directory referenced by your  `PATH`.

xosofox:
> > exec: "ssh": executable file not found in $PATH
> 
> The error means that the commad `ssh` is not found in the directories referenced by the `PATH` env var.
> So you have to put the `ssh` command into a directory referenced by your `PATH`.

The issue seems to be that the traefik image (using v2.3) does not containe ssh in its PATH. As surely traefik somehow includes ssh per se, wouldn't there be a way to include access to the executable? It seems cumbersome to bind/mount/copy ssh into the container

PR #5984, opened 2019-12-10
Use valid condition in the service details panel UI

jbdoumenjou:
### What does this PR do?

Use valid condition for the termination delay section in the service details panel.

### Motivation

Have a working service detail panel.

### More

- [ ] ~Added/updated tests~

### DIFF
--- a/webui/src/components/_commons/PanelServiceDetails.vue
+++ b/webui/src/components/_commons/PanelServiceDetails.vue
@@ -54,7 +54,7 @@
         </div>
       </q-card-section>
 
-      <q-card-section v-if="data.loadBalancer.terminationDelay">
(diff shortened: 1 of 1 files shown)


webui/src/components/_commons/PanelServiceDetails.vue
```diff
@@ -54,7 +54,7 @@
         </div>
       </q-card-section>
 
-      <q-card-section v-if="data.loadBalancer.terminationDelay">
+      <q-card-section v-if="data.loadBalancer && data.loadBalancer.terminationDelay">
```
mpl: data itself is never nil? i.e., is there no need for:

"data && data.loadBalancer && data.loadBalancer.terminationDelay" 

?

PR #5985, opened 2019-12-10
detect CloseNotify capability in accesslog and metrics

mpl:
### What does this PR do?

This PR makes sure that the ResponseWriters in accesslog and metrics only implement CloseNotify if their underlying ResponseWriter does so.

### Motivation

Fixes https://github.com/containous/traefik/issues/5953#issuecomment-563933225

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/accesslog/capture_response_writer.go
+++ b/pkg/middlewares/accesslog/capture_response_writer.go
@@ -13,12 +13,36 @@ var (
 	_ middlewares.Stateful = &captureResponseWriter{}
 )
 
+type capturer interface {
+	http.ResponseWriter
+	Size() int64
+	Status() int
+}
+
+func newCaptureResponseWriter(rw http.ResponseWriter) capturer {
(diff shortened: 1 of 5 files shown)


pkg/middlewares/accesslog/capture_response_writer.go
```diff
@@ -13,6 +13,20 @@ var (
 	_ middlewares.Stateful = &captureResponseWriter{}
 )
 
+type capturer interface {
+	http.ResponseWriter
+	Size() int64
```
dtomcej: Is this implemented?

mpl: Yes, line 66 (in old) or 90 (in new)?
Or are you talking about something else?

dtomcej: Ah, I didn't see it that far down.

mpl: happens even to the best of us ;)

PR #5996, opened 2019-12-10
Rework access control origin configuration

### LINKED ISSUE #5567: Application sent CORS headers are either not overwritten, or blocked completely

Bug

`accessControlAllowOrigin` in headers middleware doesn't return a valid value if the application already sends the `Access-Control-Allow-Origin` header. 

Final response contains duplicated headers (invalid CORS response) or zero `Access-Control-Allow-Origin` headers.

- [ ] Yes
- [x] No

### What did you do?

Application behind Traefik emits 'Access-Control-Allow-Origin: *' header.

Used the headers middleware with the following configuration:
```
kind: Middleware
metadata:
  name: cors-all
  namespace: default
spec:
  headers:
    accessControlAllowMethods:
      - "GET"
      - "OPTIONS"
      - "PUT"
      - "POST"
    accessControlAllowOrigin: "origin-list-or-null"
    accessControlMaxAge: 100
    accessControlAllowHeaders:
      - "Content-Type"
    addVaryHeader: "true"
```

Queried with curl:
```
curl localhost:80/graphql -H 'Host: my.domain' -X GET -v -H 'Origin: https://my.domain' -H'Access-Control-Request-Method: GET' -H 'Access-Control-Request-Headers: content-type'
```

```
Access-Control-Allow-Origin: https://my.domain
```

### What did you see instead?
```
Access-Control-Allow-Origin: https://my.domain
Access-Control-Allow-Origin: *
```
(the header inserted by Traefik does not replace the header sent by the application, this results in an invalid CORS configuration).

If `*` is used as the value of  `accessControlAllowOrigin`, the response contains two headers with the same value, which is also rejected by the browser.
```
Access-Control-Allow-Origin: *
Access-Control-Allow-Origin: *
```

if `null` is used, nothing happens (the documentation suggests null is valid, but this has been removed from the code - presumably for security reasons)

2.0.1

KubernetesCRD provider
Simple IngressRoute+Middleware:
```
kind: IngressRoute
metadata:
  name: ingressroute-graphql
  namespace: default
spec:
  entryPoints:
    - http
  routes:
    - match: Host(`my.domain`) && PathPrefix(`/`)
      kind: Rule
      services:
        - name: graphql-http
          port: 80
          priority: 12
      middlewares:
      - name: cors-all
---
kind: Middleware
metadata:
  name: cors-all
  namespace: default
spec:
  headers:
    accessControlAllowMethods:
      - "GET"
      - "OPTIONS"
      - "PUT"
      - "POST"
    accessControlAllowOrigin: "origin-list-or-null"
    accessControlMaxAge: 100
    accessControlAllowHeaders:
      - "Content-Type"
    addVaryHeader: "true"
```

### LINKED ISSUE #5994: Allow Configurable Access-Control-Origin list

Feature

Currently we allow 3 possible options for cors access control origin:

- `origin-list-or-null`
- `*`
- `null`

As found in the w3 documentation:
https://www.w3.org/TR/cors/#access-control-allow-origin-response-header

However, since implementation, the documentation has updated to recommend that the server keep a list of allowable origins, and only return the single origin if it is allowed, and matches the request origin:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin
https://www.w3.org/TR/cors/#access-control-allow-origin-response-header
https://tools.ietf.org/html/rfc6454#section-7.1

Also, we should remove null, as it is no longer recommended as a return value:
https://w3c.github.io/webappsec-cors-for-developers/#avoid-returning-access-control-allow-origin-null

This changes the configuration to a simpler config by allowing a list of origins:
- If a wildcard origin `"*"` is configured, always use it, and ignore all other list items
- if a wildcard is not listed, check the request origin, if it matches, return it, if not, then do something else.

We can also cement the behavior that we overwrite all upstream headers that cause issue in #5567. 

This behavior combined with documentation updates should resolve a lot of user confusion around configuring the list.

Related issue #5567

dtomcej:
### What does this PR do?

This PR:

- Updates the behavior of the header middleware in regards to access-control-origin.
- Updates the documentation to provide more configuration examples and provides links to our source material

### Motivation

Fixes  #5994 
Fixes #5567 

### More

- [x] Added/updated tests


helloIAmPau:
Hello,
any chance to know when this will be merged?

Thanks

docs/content/middlewares/headers.md
```diff
@@ -246,7 +248,7 @@ labels:
 [http.middlewares]
   [http.middlewares.testHeader.headers]
     accessControlAllowMethods= ["GET", "OPTIONS", "PUT"]
-    accessControlAllowOrigin = "origin-list-or-null"
+    accessControlAllowOrigin = ["https://foo.bar.org","https://example.org"]
```
juliens: It's breaking here, and we need to find a non-breaking solution, do you have any idea?

dtomcej: Is the issue we can't change the field type without a CRD version bump?

Can I add another field?
So can I do:

```yaml
accessControlAllowOrigin = "origin-list-or-null"
accessControlAllowOriginList = ["https://foo.bar.org","https://example.org"]
```

?

almassapargali:
Any ETA on this?

docs/content/middlewares/headers.md
ldez: For now, CRD don't have scheme so it's a problem.

nadilas:
Can we merge this? 👍

meabed:
Thank you! would this PR merged anytime soon?

PR #6005, opened 2019-12-11
Add edge case for root path with rewrite-target

### LINKED ISSUE #5990: rewrite-target invalid for kubernetes ingress on root path
Bug

### What did you do?

Creating an Ingress resource on a subdomain having only one rule on `/` that point to a backend handling requests on `/app`.

```yaml
kind: Ingress
metadata:
  name: my-ingress
  annotations:
    traefik.ingress.kubernetes.io/rewrite-target: /app
spec:
  rules:
  - host: app.my-domain.fr
    http:
      paths:
      - path: /
        backend:
          serviceName: my-app
          servicePort: http
```

Traefik to send all requests with URLs on domain with `app.my-domain.fr` to my backend on path:
* `/` -> `/app/`
* `/foo/` -> `/app/foo/`

Same as `nginx.ingress.kubernetes.io/rewrite-target` behavior.

### What did you see instead?

```
curl -s https://app.my-domain.fr/ | grep GET
<pre>Cannot GET /app</pre>
```
with a 404 status code.

But, request on `https://app.my-domain.fr//` (notice the additional `/` at end) give a 200 with expected response.

### traefik version

In Kubernetes, installed with Helm chart stable/traefik v1.82.3

```
traefik-helm: traefik:1.7.19
```

Route rule (on dashboard):
```
PathPrefix:/;ReplacePathRegex: ^/(.*) /app$1
Host:app.my-domain.fr
```

```toml
# traefik.toml
logLevel = "info"
[entryPoints]
  [entryPoints.http]
  address = ":80"
  compress = true
  [entryPoints.https]
  address = ":443"
  compress = true
    [entryPoints.https.tls]
      CertFile = "/ssl/tls.crt"
      KeyFile = "/ssl/tls.key"
  [entryPoints.traefik]
  address = ":8080"
[ping]
entryPoint = "http"
[kubernetes]
ingressClass = "traefik"
[traefikLog]
  format = "json"
[accessLog]
  format = "json"
[accessLog.fields]
  defaultMode = "keep"
[accessLog.fields.names]
[accessLog.fields.headers]
  defaultMode = "keep"
[api]
  entryPoint = "traefik"
  dashboard = true
```

dtomcej:
### What does this PR do?

This PR:

- Fixes behavior of an edge case with rewrite-target

### Motivation

Fixes #5990 

### More

- [x] Added/updated tests
- [x] Added/updated documentation - None needed, bugfix

### Additional Notes

This edge case was where we trim the trailing slashes on the rewrite-target before appending the contents of the matched path. This behavior was required becuase it ensures that we don't duplicate slashes.

However, it introduced an edge case where if the trailing slash _was_ the leading slash, it could get lost.

This is that edge case. The solution here is to capture the leading slash as part of the path capture group, and append it to the rewrite target.

### DIFF
--- a/provider/kubernetes/fixtures/ingressAnnotations_ingresses.yml
+++ b/provider/kubernetes/fixtures/ingressAnnotations_ingresses.yml
@@ -162,6 +162,23 @@ spec:
         path: /api
 
 ---
+  apiVersion: extensions/v1beta1
+  kind: Ingress
+  metadata:
+    annotations:
+      ingress.kubernetes.io/rewrite-target: /app
+    namespace: testing
+  spec:
+    rules:
+    - host: rewritetargetrootpath
+      http:
+        paths:
+        - backend:
+            serviceName: service1
+            servicePort: 80
+          path: /
+
+---
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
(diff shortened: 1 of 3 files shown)


seboudry:
Tested on own docker image builded with `make image`.
Both URLs mentioned on https://github.com/containous/traefik/issues/5990#issuecomment-564716850 respond correctly.

PR #6009, opened 2019-12-12
Add tracing headers to forwardAuth middleware requests

### LINKED ISSUE #5774: Add tracing headers to forwardAuth middleware

Feature

Add tracing headers to forwardAuth middleware. It will be nice to see traces and logs from my authorization service in full request trace.

aliksend:
### What does this PR do?

Added tracing headers to forwardAuth middleware requests
Fixes #5774

### Motivation

Motivation is to see traces and logs from authorization service in full request trace

### More

- [ ] Added/updated tests

### Additional Notes

It looks like a bug. The line `tracing.InjectRequestHeaders(forwardReq)` already present in `middlewares/auth/forward.go`, but `InjectRequestHeaders` injects headers for context that present in provided request. But in case of forwardAuth request is new and have no context. 

I tried to save context, provided in `tracing.FromContext`, in `tracing.Tracing` struct and use it in `StartSpanf`. But if I correctly understands scope of use of `tracing.Tracing`, it uses as middleware in `server.NewServer`, it independent of request, uses to "enrich" requests with tracing data and should not contain any request-dependent state. 

So I added new method (`StartSpanfWithContext`) that allows to start new span for request using  context provided from other request as parent span. 
Maybe this is not the best solution, but it is working. If you know how to improve it - write about it.

### DIFF
--- a/pkg/middlewares/auth/forward.go
+++ b/pkg/middlewares/auth/forward.go
@@ -88,6 +88,14 @@ func (fa *forwardAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	tr, err := tracing.FromContext(req.Context())
+	if err == nil {
+		opParts := []string{fa.name}
+		_, forwardReqWithCtx, finish := tr.StartSpanfWithContext(forwardReq, req.Context(), ext.SpanKindRPCClientEnum, "forward-auth", opParts, "/")
+		forwardReq = forwardReqWithCtx
+		defer finish()
+	}
+
 	writeHeader(req, forwardReq, fa.trustForwardHeader)
 
 	tracing.InjectRequestHeaders(forwardReq)
--- a/pkg/middlewares/auth/forward.go
+++ b/pkg/middlewares/auth/forward.go
@@ -91,7 +91,7 @@ func (fa *forwardAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	tr, err := tracing.FromContext(req.Context())
 	if err == nil {
 		opParts := []string{fa.name}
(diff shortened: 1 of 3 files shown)


pkg/tracing/tracing.go
```diff
@@ -148,9 +155,9 @@ func LogEventf(r *http.Request, format string, args ...interface{}) {
 	}
 }
 
-// StartSpan starts a new span from the one in the request context
-func StartSpan(r *http.Request, operationName string, spanKind ext.SpanKindEnum, opts ...opentracing.StartSpanOption) (opentracing.Span, *http.Request, func()) {
-	span, ctx := opentracing.StartSpanFromContext(r.Context(), operationName, opts...)
+// StartSpan starts a new span from the span in provided context
```
jcchavezs: Maybe better to use `StartSpan starts a new span from the one in provided context and the request.`

pkg/middlewares/auth/forward.go
```diff
@@ -88,6 +88,14 @@ func (fa *forwardAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	tr, err := tracing.FromContext(req.Context())
```
jcchavezs: I think if there is not tracing in the context we should not start a new trace. We should only honor those traces the come from traefik itself.

pkg/tracing/tracing.go
aliksend: `StartSpan` now starts span from provided context and connects it to provided request. Updated message

pkg/middlewares/auth/forward.go
aliksend: If there is no tracing in context, then error will be returned  (https://github.com/containous/traefik/blob/54bc758463db39bd227046d6584c49ff57da24d5/pkg/tracing/tracing.go#L36) and `forward-auth` span will not be started

ldez:
close in favor of #6072

PR #6011, opened 2019-12-12
Support Brotli compression format

### LINKED ISSUE #4202: Brotli Support
When will Traefik support the Brotli compression algorithm? Are there any plans in implementing this?

sh7dm:
### What does this PR do?
Add support of modern Brotli compression format to Compress middleware.

### Motivation
Fixes #4202

### More

- [x] Added/updated tests

### Additional Notes
The package used in this update is a drop-in replacement for gzip middleware which supports Brotli.

### DIFF
--- a/docs/content/middlewares/compress.md
+++ b/docs/content/middlewares/compress.md
@@ -5,18 +5,18 @@ Compressing the Response before Sending it to the Client
 
 ![Compress](../assets/img/middleware/compress.png)
 
-The Compress middleware enables the gzip compression. 
+The Compress middleware enables the Brotli or gzip compression. 
 
 ## Configuration Examples
 
 ```yaml tab="Docker"
-# Enable gzip compression
+# Enable compression
 labels:
   - "traefik.http.middlewares.test-compress.compress=true"
 ```
 
 ```yaml tab="Kubernetes"
-# Enable gzip compression
+# Enable compression
 apiVersion: traefik.containo.us/v1alpha1
 kind: Middleware
 metadata:
@@ -26,7 +26,7 @@ spec:
 ```
 
 ```yaml tab="Consul Catalog"
-# Enable gzip compression
+# Enable compression
 - "traefik.http.middlewares.test-compress.compress=true"
 ```
 
@@ -37,19 +37,19 @@ spec:
 ```
 
 ```yaml tab="Rancher"
-# Enable gzip compression
+# Enable compression
 labels:
   - "traefik.http.middlewares.test-compress.compress=true"
 ```
 
 ```toml tab="File (TOML)"
-# Enable gzip compression
+# Enable compression
 [http.middlewares]
   [http.middlewares.test-compress.compress]
 ```
 
 ```yaml tab="File (YAML)"
-# Enable gzip compression
+# Enable compression
 http:
   middlewares:
     test-compress:
@@ -61,8 +61,8 @@ http:
     Responses are compressed when:
     
     * The response body is larger than `1400` bytes.
-    * The `Accept-Encoding` request header contains `gzip`.
-    * The response is not already compressed, i.e. the `Content-Encoding` response header is not already set.
+    * The `Accept-Encoding` request header contains `gzip` or `br`.
+    * The response is not already compressed, i.e. the `Content-Encoding` response header is not already set to `gzip`, `br` or `deflate`.
 
 ## Configuration Options
 
--- a/dynamic.toml
+++ b/dynamic.toml
@@ -1,0 +1,16 @@
+[http]
+  [http.routers]
+    [http.routers.router0]
+      entryPoints = ["web"]
+      middlewares = ["my-compression"]
+      service = "service-foo"
+      rule = "Path(`/`)"
+
+[http.middlewares]
+  [http.middlewares.my-compression.compress]
+
+[http.services]
+  [http.services.service-foo]
+    [http.services.service-foo.loadBalancer]
+      [[http.services.service-foo.loadBalancer.servers]]
+        url = "http://127.0.0.1:4000/"
--- a/go.mod
+++ b/go.mod
@@ -75,6 +75,7 @@ require (
 	github.com/prometheus/client_golang v1.1.0
 	github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4
 	github.com/rancher/go-rancher-metadata v0.0.0-00010101000000-000000000000
+	github.com/sh7dm/brotlihandler v0.0.1
 	github.com/sirupsen/logrus v1.4.2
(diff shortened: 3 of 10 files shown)


ldez:
:vertical_traffic_light: Some remarks, to integrate a dependency, we have to define our level of trust for this dependency:
- :exclamation: your package don't contains any tests and don't use a CI
-  :warning: as it's a new package, no one use it: the issues detection is limited
- :exclamation: the name of your package seems ambiguous: is it only for brotli?
- :question: what's the performance of https://github.com/andybalholm/brotli?
- :question: what's the performance of your handler versus the previous gzip handler?
- :question: do yo have some documentation reference/spec about the header value `br`?

sh7dm:
Docs about `br`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

sh7dm:
@ldez Sorry, but I couldn't find any packages for Brotli & Go's HTTP server. I can change the name if you want. Perf is a question to lib's author.

sh7dm:
Testing is an issue, right. I can try to create tests but I'm not sure if I can really create them using my Go skills.

sh7dm:
@ldez I added tests & Travis CI to project: https://travis-ci.org/sh7dm/brotlihandler/builds/624162426. The only thing I'm unsure about is Push, I don't even know how to test it manually with no TLS.

mpl:
Hello @sh7dm ,

we feel that this subject would require a non-trivial amount of work to bring it to a level of confidence that we are happy about.
Since this feature request seems to be pretty popular, we agree that we should put some effort in moving it forward. To make progress, we propose either one of these approaches:
1) We create a new PR from scratch that directly adds most of the code to the Traefik project so we (and you) can iterate easily on it, and we'll find a way to give you some attribution. (the solution we would favor).
2) Or we collaborate with you directly on github.com/sh7dm/brotlihandler with PRs and commits, and to make things smoother it would greatly help if you could add us as collaborators to the repo.

What do you think?

sh7dm:
I'd like the first one! We can integrate this into Traefik, I won't mind that. I can just give this code to Traefik, it also will be available for other uses as `import github.com/containous/traefik/path/to/brotlihandler`.

chris13524:
I'm wondering if there has been any progress on this feature?

J3m5:
Any news on this feature?

sh7dm:
@ldez what should I do? I selected the variant my code is migrated to Traefik repo. Should I just push it in this PR?

jayjun:
More relevant to [github.com/sh7dm/brotlihandler](url) but gzip and Brotli compression levels should not be conflated into a single `level` setting.

[Akamai](https://blogs.akamai.com/2016/02/understanding-brotlis-potential.html) and [CloudFlare](https://blog.cloudflare.com/results-experimenting-brotli/) experiments (dated but commonly cited) show Brotli compression speed is always _worse_ compared to the same setting on gzip. Because Brotli is given precedence, users always get lousier performance.

Sadly, there is no formula to find the equivalent Brotli setting. So the compression handler must take different settings for Brotli and gzip. For sensible defaults, Akamai found Brotli (4) beats the standard default of gzip (6).

SantoDE:
Hello @sh7dm,

thanks for your interest in the project.

As you already asked @ldez, we would like to have your code implemented in Traefik directly rather than in an "unknown" dependeny.

Could you please reimplement that functionality from scratch directly in Traefik?

sh7dm:
Hello @SantoDE,
I'll try to do so. I think it'll be better to add that module into Traefik repo to make it usable for others. How do you think?

jayjun:
akamai

Sorry if I wasn’t clear. Let me point at some numbers.

1. [Akamai](https://blogs.akamai.com/2016/02/understanding-brotlis-potential.html) found Brotli (6) is **56% slower** than gzip (6), the standard default.
2. But Brotli (4) is **faster and smaller** than gzip (6).
3. If you configure the same compression level for both Brotli and gzip, you always get worse performance with Brotli. [CloudFlare](https://blog.cloudflare.com/results-experimenting-brotli/) experiments show the same.
4. Moreover, Brotli levels range 1-11 while gzip is 1-9. They are not comparable and should not be treated as the same number.

Compression speed is very important because we are dynamically compressing content on-the-fly. With the current implementation, Brotli-capable clients will experience increased latency compared to sticking with plain gzip.

J3m5:
You can still serve precompressed files.

jayjun:
@J3m5 I agree but that’s off-topic. This pull request patches the Compress middleware, which compresses HTTP responses on-the-fly.

It uses [gziphandler](https://github.com/nytimes/gziphandler#gzip-handler),
> a tiny Go package which wraps HTTP handlers to transparently gzip the response body ...

In fact, Traefik cannot pre-compress anything, it’s an edge router. You use an asset bundler like webpack to pre-compress content.

J3m5:
Ok, so does that we can already serve files pre compressed with brotli, with nginx for example ?

sh7dm:
You need some extra modules for that, but with a module you can: https://blobfolio.com/2017/06/nginx-brotli-on-debian-stretch-without-any-source-monkeying/ .

SantoDE:
@sh7dm not sure what you're asking for, to be honest. We'd need your code reimplemented in Traeifk.

sh7dm:
@SantoDE I have commited it already, see commits at this PR.

SantoDE:
@sh7dm given the code comments in `pkg/brotlihandler/brotli.go`, it looks like, we might have a licensing issue. 

Are you sure, you reimplemented that functionality?

sh7dm:
You can compare: original https://github.com/gorilla/handlers/blob/master/compress.go to my https://github.com/sh7dm/brotlihandler/blob/master/brotli.go . It's based on gorilla/handlers code.

notsureifkevin:
@SantoDE @sh7dm can this be rectified by including the full [LICENSE](https://github.com/gorilla/handlers/blob/master/LICENSE) from the Gorilla Handlers repo in the header of `pkg/brotlihandler/brotli.go` , instead of just incorrectly referencing our MIT license?

The BSD 2-Clause is permissive, we just need to include it with any source code that is based on their code (which is the case for https://github.com/sh7dm/brotlihandler/blob/master/brotli.go)

sh7dm:
I agree!

SantoDE:
Hello @sh7dm,

as already mentioned, you've based of your work on the gorilla work which comes with its own license. Copy pasting it into our headers, makes it "invisible" of our build process on how to figure out which licenses we use in our dependencies.

As your work was based on Gorilla and must include their license, we are choosing to close this PR due to the limitation mentioned above.

You're still free to open a new PR with a complete re-implementation, which might make it possible to merge it.

Thanks again for your work and I'm sorry, that we have to close that PR.

PR #6013, opened 2019-12-12
Don't nest middleware spans

aliksend:
### What does this PR do?

Converts this
before
to this
after

In other words, my PR makes middleware span to display only time when middleware is processing, not all nested middleware and forward spans.

### Motivation

To display correct span times in tracing UI

### More

- [ ] Added/updated tests


aliksend:
I think I can improve it a little bit more

aliksend:
Now it looks like this
(I've added a "long" auth middleware)

after 2

PR #6030, opened 2019-12-13
Don't set user-agent to Go-http-client/1.1

### LINKED ISSUE #5621: why traefik change user_agent(null)?

Bug

- [ ] Yes
- [√] No

### What did you do?
I access nginx through traefik.
UA="curl/7.29.0" is GOOD.

but change UA=null.i found something wrong with UA.    
UA="Go-http-client/1.1"

I want to see the original message.
I  think traefik is wrong to modify ua.
```
curl -i local.test.com
{ "@timestamp": "2019-10-10T10:53:17","status": "200", "http_request": "http://local.test.com/", "method": "GET", "user_agent":"curl/7.29.0" }
curl -i local.test.com -A ""
{ "@timestamp": "2019-10-10T10:57:55","status": "200", "http_request": "http://local.test.com/", "method": "GET", "user_agent":"Go-http-client/1.1" }
```
I want to see the original message.
If the user requests null,Please pass UA=null to bankends

traefik: V1.7.17
k8s:   v1.13.8

sh7dm:
### What does this PR do?
Force `User-Agent` header to `""` if it's not set to avoid it being replaced with `Go-http-client/1.1"`.

https://github.com/golang/go/blob/7d30af8e17d62932f8a458ad96f483b9afec6171/src/net/http/request.go#L614-L619

### Motivation
Fixes #5621, but only for v2.1.

### More

- [ ] Added/updated tests

### Additional Notes
Probably we should add a backport for this to v1.7, since the issue was reported to occur on that version. If the backport is needed, I can create it.

### DIFF
--- a/pkg/server/service/proxy.go
+++ b/pkg/server/service/proxy.go
@@ -52,6 +52,10 @@ func buildProxy(passHostHeader *bool, responseForwarding *dynamic.ResponseForwar
 			outReq.ProtoMajor = 1
 			outReq.ProtoMinor = 1
 
+			if _, ok := outReq.Header["User-Agent"]; !ok {
+				outReq.Header.Set("User-Agent", "")
+			}
+
 			// Do not pass client Host header unless optsetter PassHostHeader is set.
 			if passHostHeader != nil && !*passHostHeader {
 				outReq.Host = outReq.URL.Host


mpl:
@sh7dm According to the code you pointed at https://github.com/golang/go/blob/7d30af8e17d62932f8a458ad96f483b9afec6171/src/net/http/request.go#L614-L619
the intent is apparently to set a default _only_ if the header does not exist. i.e. if it exists but it is blank (""), then the stdlib does not touch it.
To that extent, I'm wondering if curl is not the one at fault here, and that maybe it should voluntarily set the header to an empty string when -A "" is used.
WDYT?

sh7dm:
If I have time I'll try `wget` instead of `curl`

mpl:
@sh7dm Nevermind, I think you are right. With respect to the client, Traefik is a server, and it should not be blindly using the default behaviour that is set when creating a client (the code you pointed out). 
And indeed, what you're proposing is what the director of the stdlib reverseproxy does too: https://github.com/golang/go/blob/master/src/net/http/httputil/reverseproxy.go#L119

PR #6048, opened 2019-12-17
Rename the non-exposed field "count" to "size"

sylr:
### What does this PR do?

This PR renames a variable which name led to believe it contains a count whereas it contains a size.

### Motivation

I've lost quite a bit of time while debuging thinking that `crr.count` was wrongly incremented. 

### More

- [ ] Added/updated tests


sylr:
I really don't like `captureRequestReader.count`, I think a variable name should describe what it contains, not how it is populated.

If you don't want to name it `size` what about `bytesRead` ?

PR #6054, opened 2019-12-17
fix statsd scale for duration based metrics

### LINKED ISSUE #5944: Statsd incorrect unit
### What did you do?

When I configure statsd provider for tracing and I trace an HTTP request which is known to last approximately 1 second.

On my statsd receiver:
```
traefik.service.request.duration:1055.487|ms
traefik.entrypoint.request.duration:1058.461|ms
```
See statsd spec https://github.com/b/statsd_spec#timers explains that the timers are milliseconds.

### What did you see instead?
```
traefik.service.request.duration:1.055487|ms
traefik.entrypoint.request.duration:1.058461|ms
```
```
Version:      2.0.1
Codename:     montdor
Go version:   go1.13.1
Built:        2019-09-26T16:20:36Z
OS/Arch:      windows/amd64
```

### Additional infos
It seems that there was issues with go-kit/kit which is used to handle metrics, see https://github.com/go-kit/kit/issues/608 and https://github.com/go-kit/kit/pull/610

ddtmachado:
### What does this PR do?

Fixes #5944  

### Motivation

StatsD assumes that duration based metric values are in Milliseconds.

Currently Traefik always calls the `Observe` function converting the `time.Duration` to seconds. With this change we can set the scale in the metric provider implementation, on statsD case Milliseconds, while keeping the current value as the default for those that don't specify it.

### More

- [X] Added/updated tests

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,6 +1,8 @@
 package metrics
 
 import (
+	"time"
+
 	"github.com/go-kit/kit/metrics"
 	"github.com/go-kit/kit/metrics/multi"
 )
@@ -20,12 +22,12 @@ type Registry interface {
 
 	// entry point metrics
 	EntryPointReqsCounter() metrics.Counter
-	EntryPointReqDurationHistogram() metrics.Histogram
+	EntryPointReqDurationHistogram() ScalableHistogram
 	EntryPointOpenConnsGauge() metrics.Gauge
 
 	// service metrics
 	ServiceReqsCounter() metrics.Counter
-	ServiceReqDurationHistogram() metrics.Histogram
+	ServiceReqDurationHistogram() ScalableHistogram
 	ServiceOpenConnsGauge() metrics.Gauge
 	ServiceRetriesCounter() metrics.Counter
 	ServiceServerUpGauge() metrics.Gauge
@@ -46,10 +48,10 @@ func NewMultiRegistry(registries []Registry) Registry {
 	var lastConfigReloadSuccessGauge []metrics.Gauge
 	var lastConfigReloadFailureGauge []metrics.Gauge
 	var entryPointReqsCounter []metrics.Counter
-	var entryPointReqDurationHistogram []metrics.Histogram
+	var entryPointReqDurationHistogram []ScalableHistogram
 	var entryPointOpenConnsGauge []metrics.Gauge
 	var serviceReqsCounter []metrics.Counter
-	var serviceReqDurationHistogram []metrics.Histogram
+	var serviceReqDurationHistogram []ScalableHistogram
 	var serviceOpenConnsGauge []metrics.Gauge
 	var serviceRetriesCounter []metrics.Counter
 	var serviceServerUpGauge []metrics.Gauge
@@ -101,10 +103,10 @@ func NewMultiRegistry(registries []Registry) Registry {
 		lastConfigReloadSuccessGauge:   multi.NewGauge(lastConfigReloadSuccessGauge...),
 		lastConfigReloadFailureGauge:   multi.NewGauge(lastConfigReloadFailureGauge...),
 		entryPointReqsCounter:          multi.NewCounter(entryPointReqsCounter...),
-		entryPointReqDurationHistogram: multi.NewHistogram(entryPointReqDurationHistogram...),
+		entryPointReqDurationHistogram: NewMultiHistogram(entryPointReqDurationHistogram...),
 		entryPointOpenConnsGauge:       multi.NewGauge(entryPointOpenConnsGauge...),
 		serviceReqsCounter:             multi.NewCounter(serviceReqsCounter...),
-		serviceReqDurationHistogram:    multi.NewHistogram(serviceReqDurationHistogram...),
+		serviceReqDurationHistogram:    NewMultiHistogram(serviceReqDurationHistogram...),
 		serviceOpenConnsGauge:          multi.NewGauge(serviceOpenConnsGauge...),
 		serviceRetriesCounter:          multi.NewCounter(serviceRetriesCounter...),
 		serviceServerUpGauge:           multi.NewGauge(serviceServerUpGauge...),
@@ -119,10 +121,10 @@ type standardRegistry struct {
 	lastConfigReloadSuccessGauge   metrics.Gauge
 	lastConfigReloadFailureGauge   metrics.Gauge
 	entryPointReqsCounter          metrics.Counter
-	entryPointReqDurationHistogram metrics.Histogram
+	entryPointReqDurationHistogram ScalableHistogram
 	entryPointOpenConnsGauge       metrics.Gauge
 	serviceReqsCounter             metrics.Counter
-	serviceReqDurationHistogram    metrics.Histogram
+	serviceReqDurationHistogram    ScalableHistogram
 	serviceOpenConnsGauge          metrics.Gauge
 	serviceRetriesCounter          metrics.Counter
 	serviceServerUpGauge           metrics.Gauge
@@ -156,7 +158,7 @@ func (r *standardRegistry) EntryPointReqsCounter() metrics.Counter {
 	return r.entryPointReqsCounter
 }
 
-func (r *standardRegistry) EntryPointReqDurationHistogram() metrics.Histogram {
+func (r *standardRegistry) EntryPointReqDurationHistogram() ScalableHistogram {
 	return r.entryPointReqDurationHistogram
 }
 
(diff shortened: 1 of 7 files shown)


pkg/middlewares/metrics/metrics.go
```diff
@@ -94,7 +96,26 @@ func (m *metricsMiddleware) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 
 	labels = append(labels, "code", strconv.Itoa(recorder.getCode()))
 	m.reqsCounter.With(labels...).Add(1)
-	m.reqDurationHistogram.With(labels...).Observe(time.Since(start).Seconds())
+	m.reqDurationHistogram.With(labels...).Observe(m.getDurationValue(time.Since(start)))
+}
+
+func (m *metricsMiddleware) getDurationValue(d time.Duration) float64 {
+	switch m.timeScale {
```
mpl: This whole switch does not work if someone creates a standardRegistry with e.g. `timeScale = 500 * time.Millisecond`, right?
Shouldn't we do something about that?
At the very least, I think the default case L116 should be used to catch the above case, and return an error or generate a warning.

ddtmachado: yes, Initially I thought it would be better to just fallback to Seconds (current behavior) if any other value was passed rather than introducing a new error, but that can lead to some tricky situations in the future. I'll add the error return.

ddtmachado: I just noted that there is another problem in this PR and it won't work as expected on `metrics.NewMultiRegistry`. Fixing this before I update again.

pkg/metrics/metrics.go
```diff
@@ -183,3 +185,90 @@ func (r *standardRegistry) ServiceRetriesCounter() metrics.Counter {
...
+
+// ScalableHistogram is a Histogram with a predefined time unit used when
+// producing observations allowing the call to Start() and ObserveDuration()
+// without explicitly setting the observed value.
+type ScalableHistogram interface {
+	With(labelValues ...string) ScalableHistogram
+	Start()
+	Observe(v float64)
+	ObserveDuration()
+}
```
mpl: Why is it useful to have this interface? as far as I can see, there's only one implementation for it (aside from histogramMock), which is HistogramWithScale, right? So why don't we simply have HistogramWithScale everywhere?

ddtmachado: Initially I thought about tests and trying to keep it interchangeable with Timer, but it didn't work out :)
So yes we can have HistogramWithScale directly if you prefer

mpl: yes, I don't think the interface brings anything to the table, and it looks like it makes the whole change a little bit more complicated than it needs to be.

mpl:
By the way, @ddtmachado have you noticed that the tests do not pass on the CI (https://semaphoreci.com/containous/traefik/branches/pull-request-6054/builds/4) ?
This seems to be the culprit:

```
FAIL: simple_test.go:256: SimpleSuite.TestMetricsPrometheusDefaultEntryPoint

simple_test.go:268:
    c.Assert(err, checker.IsNil)
... value *errors.errorString = &errors.errorString{s:"try operation failed: failed to read response body: unexpected EOF"} ("try operation failed: failed to read response body: unexpected EOF")
```

pkg/metrics/metrics.go
ddtmachado: actually the interface is used to hide `MultiHistogram` (slice of Histograms) as well. In true all metric objects on `standardRegistry` are slices. I think it will be better to keep the same pattern in this case. Do we keep it or proceed to expose a concrete type only for histograms?

mpl: ah ok. If both the type and the slice of type are indeed publicly needed, then yes, I suppose it's fair enough to keep them behind the interface, if that makes the code simpler.

pkg/metrics/metrics.go
```diff
@@ -183,3 +185,90 @@ func (r *standardRegistry) ServiceRetriesCounter() metrics.Counter {
...
+func (s *HistogramWithScale) With(labelValues ...string) ScalableHistogram {
+	s.histogram = s.histogram.With(labelValues...)
+	return s
+}
+
+// Start implements ScalableHistogram.
+func (s *HistogramWithScale) Start() {
+	s.timer = metrics.NewTimer(s.histogram)
+	s.timer.Unit(s.unit)
+}
```
mpl: I'm not sure we need any sort of pseudo-atomicity for the multi-histogram, but if we do, 
we could have a private method, such as:

func (s *HistogramWithScale) start(t time.Time)

which instantiates a Timer with that specific time.
then Start would call s.start(time.Now())
and the multi histogram would do:
t := time.Now()
and call s.start(t) for each of the histograms in the range.

ddtmachado: :+1: done

pkg/metrics/metrics.go
```diff
@@ -186,12 +186,12 @@ func (r *standardRegistry) ServiceServerUpGauge() metrics.Gauge {
...
 
-// ScalableHistogram is a Histogram with a predefined time unit used when
-// producing observations allowing the call to Start() and ObserveDuration()
-// without explicitly setting the observed value.
+// ScalableHistogram is a Histogram with a predefined time unit, used when
+// producing observations without explicitly setting the observed value.
 type ScalableHistogram interface {
 	With(labelValues ...string) ScalableHistogram
 	Start()
+	StartAt(t time.Time)
```
mpl: I'm not sure it's interesting to provide this exported StartAt. It is only a very small convenience over Start, right? there is nothing that StartAt allows that could not also be done just with Start, isn't it?
If you want to get rid of Start, and keep only StartAt, that's fine imo, but I think keeping both is overkill.

pkg/metrics/metrics.go
```diff
@@ -243,13 +249,20 @@ func NewMultiHistogram(h ...ScalableHistogram) MultiHistogram {
 	return MultiHistogram(h)
 }
 
-// Start implements ScalableHistogram.
+// Start implements ScalableHistogram. Not atomic in time start for each histogram.
 func (h MultiHistogram) Start() {
 	for _, histogram := range h {
 		histogram.Start()
 	}
 }
```
mpl: I thought the whole point of the last change was to fix that method? why not:

```
now := time.Now()
for _, histogram := range h {
	histogram.StartAt(now)
}
```
?

pkg/metrics/metrics.go
```diff
@@ -211,15 +211,21 @@ func (s *HistogramWithScale) With(labelValues ...string) ScalableHistogram {
...
+// StartAt implements ScalableHistogram.
+func (s *HistogramWithScale) StartAt(t time.Time) {
+	s.start = t
 }
 
 // ObserveDuration implements ScalableHistogram.
 func (s *HistogramWithScale) ObserveDuration() {
-	if s.timer != nil {
-		s.timer.ObserveDuration()
+	d := float64(time.Since(s.start).Nanoseconds()) / float64(s.unit)
```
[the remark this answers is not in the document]
ddtmachado: goo catch, wdyt of just returning if the unit is <= 0 to avoid generating invalid observations?

pkg/metrics/metrics.go
ddtmachado: agreed, I'll keep just StartAt

pkg/metrics/metrics.go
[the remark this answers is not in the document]
mpl: Yes, we need at least that as a last line of defence. However, since we are already making clear to the users of the API that they should use the provided NewHistogramWithScale constructor anyway, I would also make NewHistogramWithScale return nil, errors.New("provided unit must be > 0")

pkg/metrics/metrics.go
```diff
@@ -183,3 +185,94 @@ func (r *standardRegistry) ServiceRetriesCounter() metrics.Counter {
...
+	s.histogram.Observe(v)
+}
+
+// NewHistogramWithScale returns a ScalableHistogram
+func NewHistogramWithScale(h metrics.Histogram, u time.Duration) ScalableHistogram {
+	return &HistogramWithScale{
+		histogram: h,
+		unit:      u,
+	}
+}
```
mpl: Please see my answer in ObserveDuration.
Also, let's make the argument name more self-explanatory (s/u/unit/).

pkg/metrics/metrics.go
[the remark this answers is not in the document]
ddtmachado: I'm reluctant to add an error throwing in there as it would have to be sent down the stack, where they never expect failures while allocating a metric type. If you are ok with it I would leave only the defensive code to avoid panic and the responsibility for correct unit value passing to the user of this Type.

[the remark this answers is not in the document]
mpl: Are you referring to the call of NewHistogramWithScale in RegisterStatsd of statsd.go? If yes, I fail to see the problem: we are in control of the call there, and we know it is called with a correct value (time.Millisecond), so it would be fine to ignore the error returned by NewHistogramWithScale there, no?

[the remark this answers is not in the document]
mpl: Hey. We are in a hurry to get that PR merged, so that it is included in the next release. So jsyk, I'm going to try to do the error return for NewHistogramWithScale in a commit within this PR.

PR #6055, opened 2019-12-18
Add period for rate limiter middleware

### LINKED ISSUE #5918: Specify period for rate limiting on v2

Feature

1.7 supported to specify periods for rate-limiting, e.g. 10 requests per 60 seconds (https://docs.traefik.io/v1.7/configuration/commons/#rate-limiting). I don't see this functionality with v2 and the rate limit middleware (https://docs.traefik.io/middlewares/ratelimit/). 

Was this removed intentionally? Is the smallest possible rate limit with v2 therefore 1 per second?

mpl:
### What does this PR do?

We're getting some user feedback that they would like to be able to specify a rate limit lower than 1 request per second.
One possibility would have been to change the "average" field to a float (e.g. 6 requests / minute would have been average=0.1).
However, that would break the configuration (among other things) as TOML will not decode a value that looks like an int (no decimal point specified) into a Go float64.

Therefore, this PR (re-)introduces the notion of a period, that allows one to (optionally) specify the duration for which the given average applies.
The default value for the period is 1 second, which means if no period is specified, the meaning of the average does not change compared with before this change (it is still in requests/s).

### Motivation

Fixes #5918 

### More

- [x] Added/updated tests

### DIFF
--- a/script/test-unit
+++ b/script/test-unit
@@ -11,7 +11,7 @@ GREEN=$'\033[32m'
 TEXTRESET=$'\033[0m' # reset the foreground colour
 
 # -failfast -timeout=5m
-TESTFLAGS=(-cover "-coverprofile=cover.out" "${TESTFLAGS}")
+TESTFLAGS=(-test.short=true -cover "-coverprofile=cover.out" "${TESTFLAGS}")
 
 if [ -n "$VERBOSE" ]; then
     TESTFLAGS+=(-v)
--- a/script/test-unit
+++ b/script/test-unit
@@ -11,7 +11,7 @@ GREEN=$'\033[32m'
 TEXTRESET=$'\033[0m' # reset the foreground colour
 
 # -failfast -timeout=5m
-TESTFLAGS=(-test.short=true -cover "-coverprofile=cover.out" "${TESTFLAGS}")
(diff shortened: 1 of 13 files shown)


script/test-unit
```diff
@@ -11,7 +11,7 @@ GREEN=$'\033[32m'
 TEXTRESET=$'\033[0m' # reset the foreground colour
 
 # -failfast -timeout=5m
-TESTFLAGS=(-cover "-coverprofile=cover.out" "${TESTFLAGS}")
+TESTFLAGS=(-test.short=true -cover "-coverprofile=cover.out" "${TESTFLAGS}")
```
dtomcej: Is this change intended to be committed?

mpl: We need to think about what behaviour we want where, but ultimately, I think yes, we want some switch around test.short=true , so one can easily skip the couple of long-ass tests I've added in this PR.
Question is, where/when do we want to disable them? on Semaphore? on localhost? always except when specified otherwise? etc.

mpl:
yay I deleted my branch while doing some cleanup

script/test-unit
mpl: As discussed with others, we'll come back to that question later if we feel like we need to. I just wanted to raise attention to the fact that some of the added tests were long-lived, which is now done, so I'll remove that part for now.

Issue #4678, opened 2019-03-27
Migrate integration tests to docker-compose v3


jbdoumenjou:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

There's a number features in docker-compose version 3 (in particular around networking) that Traefik could benefit from. Unfortunately, the Go package for docker-compose that Traefik is currently using (github.com/libkermit/compose) does not support docker-compose version 3.

Therefore, this issue is about finding a suitable replacement for libkermit/compose, compatible with docker-compose version 3, and using it in Traefik. It could be as simple as directly using calls to the docker-compose command-line tool.

PR #6061, opened 2019-12-18
Fix integration tests on MacOS

sylr:
### What does this PR do?

This PR improves the ability to run integration tests on MacOS.

To do so, we make the defaults traefik entryPoints listen on `localhost:<port>` instead of `:<port>` so that macos does not complain an unsigned application is trying to receive incoming connection.

Also fix all test fixture configs to use `localhost:<port>`.

### Motivation

I'm trying to run integrations test on my dev machine to validate my work.

### More

- [ ] Added/updated tests

### DIFF
--- a/integration/proxy_protocol_test.go
+++ b/integration/proxy_protocol_test.go
@@ -3,6 +3,7 @@ package integration
 import (
 	"net/http"
 	"os"
+	"runtime"
 	"time"
 
 	"github.com/containous/traefik/v2/integration/try"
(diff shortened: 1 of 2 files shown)


dtomcej:
Hey @sylr, it was my understanding that the container based integration tests worked fine, while the host based tests did not.

Is this something that can be explored to decrease the scope of the tests that are affected by this PR?

sylr:
Hi @dtomcej,

You are right, the fixtures/ changes might be useless. I'll rollback them.

Thanks.

dtomcej:
Instead of https://github.com/containous/traefik/pull/6061/files#diff-8449e3d93a890b1153a65283a93cf08bR22 Where tests are disabled because the IP isn't routable,

How about modifying the tests so that the request doesn't come from the host, but from another container.

It would require creating a tools container, and then docker exec or such to run the commands from that container. That would allow the tests to run, but without running from the host.

sylr:
Are you suggesting we move the `ProxyProtocolSuite` from `host` to `container` ?

ldez:
The root cause, of the majority of the problem with our integration tests on MacOS, is the docker-compose version that we use. We know that since a long time.

This require to change the tooling we use to run those kind of tests. https://github.com/containous/traefik/issues/4678

Changing Traefik code only to run integration tests on MacOS is not a good approach and having conditional tests only for MacOS does not seem a good approach either.

sylr:
Skipping the ProxyProtocolSuite allowed me to run the whole test integrations suite successfully on my machine (with the changes made to the Makefile that are in https://github.com/containous/traefik/pull/6064).

ldez:
Related to my previous post, we will decline this PR.

Thanks anyway.

PR #6064, opened 2019-12-18
Improve Makefile

sylr:
### What does this PR do?

This is an attempt to improve and rationalize the build and crossbuild systems.

### Motivation

Currently when building on MacOS, docker build export an artifact that is for linux/amd64 instead of darwin/amd64.

During my investigations I found it hard to understand the Makefile so at a point I decided to try to simplify it.

This is the result, I hope you will like it.

### More

- [ ] Added/updated tests


ldez:
The goal is interesting but there are too many changes, and a lot of this changes are very opinionated.

Because of the amount of changes, it's possible to review it and discuss on that PR.

So I recommend a "small steps" approach, with small PRs, instead of this huge PR.

sylr:
Yes there is quite a bit of changes but it's mostly located in the Makefile since I've almost entirely rewritten all the targets to clearly distinguish what is done locally and what is done inside docker.

I don't know how I could break it down in several PRs.

PR #6072, opened 2019-12-19
fix(tracing): makes sure tracing headers are being propagated when using forwardAuth

### LINKED ISSUE #5774: Add tracing headers to forwardAuth middleware

jcchavezs:
### What does this PR do?

This PR **fixes** the propagation of tracing headers when using `forwardAuth`

### Motivation

fixes #5774 
close #6009

### More

- [x] Added/updated tests

### Additional Notes

This PR fixes the problems described in https://github.com/containous/traefik/pull/6009.

**Important Note:** While working on this I realized that tracing propagation will only work if tracing middleware happens before the forwardauth middleware, is this something we can guarantee?

Ping @ldez @aliksend

### DIFF
--- a/pkg/middlewares/auth/forward.go
+++ b/pkg/middlewares/auth/forward.go
@@ -88,9 +88,11 @@ func (fa *forwardAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	writeHeader(req, forwardReq, fa.trustForwardHeader)
+	// Ensure tracing headers are in the request before we copy the headers to the
+	// forwardReq.
+	tracing.InjectRequestHeaders(req)
 
-	tracing.InjectRequestHeaders(forwardReq)
+	writeHeader(req, forwardReq, fa.trustForwardHeader)
 
 	forwardResponse, forwardErr := httpClient.Do(forwardReq)
 	if forwardErr != nil {
--- a/pkg/middlewares/auth/forward_test.go
+++ b/pkg/middlewares/auth/forward_test.go
@@ -3,13 +3,18 @@ package auth
 import (
 	"context"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"testing"
 
 	"github.com/containous/traefik/v2/pkg/config/dynamic"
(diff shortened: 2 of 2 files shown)


pkg/middlewares/auth/forward.go
```diff
@@ -88,9 +88,11 @@ func (fa *forwardAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	writeHeader(req, forwardReq, fa.trustForwardHeader)
+	// Ensure tracing headers are in the request before we copy the headers to the
+	// forwardReq.
+	tracing.InjectRequestHeaders(req)
 
-	tracing.InjectRequestHeaders(forwardReq)
+	writeHeader(req, forwardReq, fa.trustForwardHeader)
```
jcchavezs: If we don't inject request headers before we copy all the headers in `writeHeader` then the tracing headers get lost (remember, `forwardReq` and `req` don't share the context.

ldez:
Hello, could you rebase on the branch v2.1. Thank you.

jcchavezs:
@ldez done

aliksend:
@ldez This PR it duplicate of my (#6009). Can you pls review it too because "There can be only one"

jcchavezs:
@aliksend yeah, so based on what you said 

> It looks like a bug. The line tracing.InjectRequestHeaders(forwardReq) already present in middlewares/auth/forward.go, but InjectRequestHeaders injects headers for context that present in provided request. But in case of forwardAuth request is new and have no context.

I think this is more like a bug because no unit test was added for checking that the context was propagated. This PR fixes that bug keeping the code **mostly** as it is now. I prefer this approach as the side effects are zero.

jcchavezs:
Any input on this @ldez ?

ldez:
The approach with zero side effects seems better to me, so I will "choose" this PR.

ldez:
@jcchavezs you don't need to "merge master", we have a bot for that :wink:

jcchavezs:
Awesome I did not know @ldez! Thanks.

PR #6083, opened 2019-12-23
Implement dual certificate support (RSA/EC on same domain)

### LINKED ISSUE #3483: Can't have RSA and EC certificates for same domain(s)

Bug

### What did you do?
I wanted to add 2 certificates for the same domain name (one RSA and one EC cert) in order to mimic HAProxy (http://cbonte.github.io/haproxy-dconv/1.8/configuration.html#5.1-crt)

I expected Traefik to serve the EC cert to clients advertising ECDSA in their cipher suite and RSA to the others

### What did you see instead?

During startup of `traefik` the following WARN is printed:
```
WARN[2018-06-12T18:06:09+02:00] Into EntryPoint https, try to add certificate for domains which already have this certificate (,*.xxx.tld,xxx.tld). The new certificate will not be append to the EntryPoint. 
```

It is generated by https://github.com/containous/traefik/blob/master/tls/certificate.go#L178

```
Version:      dev
Codename:     cheddar
Go version:   go1.10.3
OS/Arch:      linux/amd64
```

```toml
debug = false
checkNewVersion = false
logLevel = "INFO"
defaultEntryPoints = ["https","http"]

[entryPoints]
  [entryPoints.traefik]
    address = ":6969"
  [entryPoints.http]
    address = ":80"
  [entryPoints.https]
    address = ":443"
    [entryPoints.https.tls]
      minVersion = "VersionTLS12"
      certFile = "cert.rsa.pem"
      keyFile = "cert.rsa.key"
      certFile = "cert.ecdsa.pem"
      keyFile = "cert.ecdsa.key"

[...]
```

If I understand correctly, only the certificate name is checked in order to it append (or not) to the map.

Maybe the check should be cert name + PublicKeyAlgorithm ? But then there might be other implications in other places of the code (I'm not very familiar with traefik's source code)

icedream:
### What does this PR do?

This allows loading multiple certificates for the same domains as long as they aren't of the same type (RSA vs. EC, with EC including ECDSA and Ed25519) and also enables generation of default certificates for RSA and ECDSA.

Fixes #3483.

### Motivation

Allowing both RSA and EC certificates for a domain allows using a wider range of cipher suites, since some only work based on an RSA certificate and some only work with an EC certificate. Newer clients could then be served the EC certificate and use newer cipher suites for better and faster encryption while older clients can fall back to RSA.

### More

- [x] Added/updated tests

### Additional Notes

This merge request does not implement nor enable any capabilities for certificate providers to request certificates for both RSA and EC, instead it currently relies on the configuration either being set up with usage of default certificates or existing certificates (for example by providing them from another ACME client). I see this is as a separate follow-up issue. See #4993 for a previous PR that attempted to change the default key type for the built-in ACME client.

This PR tries to implement support as transparently as possible, but it includes a deprecation of the `defaultCertificate` TLS store configuration key and replaces it with the `defaultCertificates` array while still keeping support. This should probably be discussed on before merging this.


icedream:
etcd-related integration testing is failing for reasons that are unknown to me, the results also seem inconsistent as the testing fails with different configuration entries missing on my local run compared to the run on Semaphore CI.

Local run:

```
----------------------------------------------------------------------
FAIL: etcd_test.go:59: EtcdSuite.TestSimpleConfiguration

etcd_test.go:157:
    c.Error(text)
... Error: --- Expected
+++ Got
@@ -27,7 +27,6 @@
       "tls": {
         "domains": [
           {
-            "main": "aaa.localhost",
             "sans": [
               "aaa.aaa.localhost",
               "bbb.aaa.localhost"
@@ -36,8 +35,7 @@
           {
             "main": "bbb.localhost",
             "sans": [
-              "aaa.bbb.localhost",
-              "bbb.bbb.localhost"
+              "aaa.bbb.localhost"
             ]
           }
         ]
@@ -110,8 +108,7 @@
     "striper@etcd": {
       "stripPrefix": {
         "prefixes": [
-          "foo",
-          "bar"
+          "foo"
         ],
         "forceSlash": true
       },
@@ -154,8 +151,7 @@
         "service": "simplesvc",
         "mirrors": [
           {
-            "name": "srvcA",
-            "percent": 42
+            "name": "srvcA"
           },
           {
             "name": "srvcB",

```

starcraft66:
Would be really happy to see this added!

dtomcej:
Hello @icedream,

We have upgraded Traefik to be using golang 1.14, and with that brought some TLS changes that affect this PR.

1. Do you think that this PR is still needed with the changes to Golang 1.14 TLS certificate handling?
2. Do you think that you could use the 1.14 features to implement this feature while reducing the complexity?

Please let us know!

icedream:
Hello,

while I didn't check through the changes in Go 1.14 in detail for this PR yet, here are a few points from the [changelog](https://golang.org/doc/go1.14) that will become highly relevant for this PR:

> When multiple certificate chains are provided through the Config.Certificates field, the first one compatible with the peer is now automatically selected. This allows for example providing an ECDSA and an RSA certificate, and letting the package automatically select the best one. Note that the performance of this selection is going to be poor unless the Certificate.Leaf field is set.

This means if we fully want to off-load certificate selection to Golang itself, we would need to completely rewrite how Traefik manages certificates internally. Right now Traefik handles that logic itself and I don't know how much time I would need to pour into this PR to make the code work around what Go now offers. I am a bit limited in how much I can work at this PR at the moment but creating a second PR to replace this one seems like the thing to do here.

>  The new (*ClientHelloInfo).SupportsCertificate and (*CertificateRequestInfo).SupportsCertificate methods expose whether a peer supports a certain certificate. 

This would be the thing to use if we are going to keep how Traefik currently implements TLS logic.

If anyone else checked through the changes and has concrete suggestions for reworking this PR, I will definitely take these suggestions into account!

notsureifkevin:
> This means if we fully want to off-load certificate selection to Golang itself, we would need to completely rewrite how Traefik manages certificates internally.

The team agrees with your assessment, but are limited in the ability to scope or commit to building a proposal for this change.

> creating a second PR to replace this one seems like the thing to do here

Given the capabilities introduced in 1.14 this makes the most sense. For now, I propose closing this PR and updating the original issue (https://github.com/containous/traefik/issues/3483) with a summary of what may be required to exploit the capabilities of most recent updates to the tls package.

PR #6085, opened 2019-12-23
Add namespace attribute on IngressRouteTCP service

jbdoumenjou:
### What does this PR do?

Add a namespace attribute on `IngressRouteTCP` service.

### Motivation

Allow to specify a namespace to reference a Kubernetes service from `IngressRouteTCP`.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/crd/fixtures/tcp/with_different_services_ns.yml
+++ b/pkg/provider/kubernetes/crd/fixtures/tcp/with_different_services_ns.yml
@@ -11,11 +11,11 @@ spec:
   routes:
   - match: HostSNI(`foo.com`)
     services:
-#     without namespace
+    # without namespace
     - name: whoamitcp
       port: 8000
       weight: 2
-#     with default namespace
+    # with default namespace
     - name: whoamitcp2
       namespace: default
       port: 8080
--- a/pkg/provider/kubernetes/crd/fixtures/tcp/with_different_services_ns.yml
+++ b/pkg/provider/kubernetes/crd/fixtures/tcp/with_different_services_ns.yml
@@ -1,0 +1,31 @@
+apiVersion: traefik.containo.us/v1alpha1
+kind: IngressRouteTCP
+metadata:
+  name: test.route
+  namespace: default
+
+spec:
+  entryPoints:
+    - foo
+
+  routes:
+  - match: HostSNI(`foo.com`)
+    services:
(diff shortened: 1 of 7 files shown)


dtomcej:
Cross namespace referencing is discouraged due to fishing, and security/accountability reasons.

Is there a use case where an IngressRouteTCP could not be created in the target namespace?

Also, this does not take into account that the target namespace may not be watched/allowed via RBAC, and the informers may not be up to date with the endpoints...

dduportal:
There is a long dicussion in Kubernetes tracker about this if you want to deep dive the subject: https://github.com/kubernetes/kubernetes/issues/17088 .

Nginx usage for solving this is interesting though: they have the concept of "merged Ingress": https://github.com/nginxinc/kubernetes-ingress/tree/master/examples-of-custom-resources/cross-namespace-configuration .

dtomcej:
I think that if this was enabled by a feature flag, it would be fine.

`--provider.kubernetescrd.crossnamespacereferences=true` with a default of false would be enough.

It would keep the current behavior, and allow the defaults to be secure, and allow users to allow references if required.

This could be implemented to the ingressRoute and ingressRouteTCP crds IMO if that feature flag were implemented.

ldez:
The namespace is already available for the services in IngressRoute, so we cannot introduce the option with the expected behavior without breaking.

pkg/provider/kubernetes/crd/fixtures/tcp/with_different_services_ns.yml
```diff
@@ -0,0 +1,31 @@
...
+  namespace: default
+
+spec:
+  entryPoints:
+    - foo
+
+  routes:
+  - match: HostSNI(`foo.com`)
+    services:
+#     without namespace
```
mpl:
```suggestion
    #  without namespace
```

pkg/provider/kubernetes/crd/fixtures/tcp/with_different_services_ns.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: whoamitcp
+      port: 8000
+      weight: 2
+#     with default namespace
```
mpl:
```suggestion
    # with default namespace
```

mpl:
LGTM after formatting fix.

PR #6087, opened 2019-12-24
Log the ignored namespace only when needed

### LINKED ISSUE #6033: Warning for ignored namespace in cross-provider context is too noisy

Bug

### What did you do?

Define an `IngressRoute` with KubernetesCRD provider as follows:

```
kind: IngressRoute
metadata:
  name: traefik-dashboard
  namespace: ingress
spec:
  entryPoints:
  - traefik
  routes:
  - kind: Rule
    match: PathPrefix(`/dashboard`) || PathPrefix(`/api`)
    services:
    - kind: TraefikService
      name: api@internal
```

Relatively quiet logs with level for standard operation with `--log.level=WARN`

### What did you see instead?

A constant stream of warnings about the namespace "ingress" being ignored.

```
Version:      2.1.1
Codename:     cantal
Go version:   go1.13.5
Built:        2019-12-12T19:01:37Z
OS/Arch:      linux/amd64
pod "traefik" deleted
```

```
    - --entryPoints.traefik.address=:9000
    - --entryPoints.web.address=:8000
    - --entryPoints.websecure.address=:8443
    - --api.dashboard=true
    - --ping=true
    - --providers.kubernetescrd
    - --log.level=WARN
    - --accesslog=true
    - --global.checknewversion=true
    - --global.sendanonymoususage=true
```

There is a section in the code somewhere that suppresses the namespace if it is `default`.   A current workaround is to explicitly specify a namespace of default for cross-provider references.

IMO this is incorrect behavior. It should suppress the messages if a namespace is not provided, or possibly if the namespace matches that of where the resource is defined.

Another thought to consider is reducing the level, warning is pretty high for something benign

jbdoumenjou:
### What does this PR do?

In the Kubernetes CRD provider, log the ignored namespace only when a service namespace is explicitly defined with a cross provider reference.

### Motivation

Fixes #6033

### More

- [ ] ~Added/updated tests~

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes_http.go
+++ b/pkg/provider/kubernetes/crd/kubernetes_http.go
@@ -380,11 +380,11 @@ func (c configBuilder) nameAndService(ctx context.Context, namespaceService stri
 			return "", nil, err
 		}
 
-		fullName := fullServiceName(svcCtx, namespace, service.Name, service.Port)
+		fullName := fullServiceName(svcCtx, namespace, service, service.Port)
 
 		return fullName, serversLB, nil
 	case service.Kind == "TraefikService":
-		return fullServiceName(svcCtx, namespace, service.Name, 0), nil, nil
+		return fullServiceName(svcCtx, namespace, service, 0), nil, nil
 	default:
 		return "", nil, fmt.Errorf("unsupported service kind %s", service.Kind)
 	}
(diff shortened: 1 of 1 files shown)


pkg/provider/kubernetes/crd/kubernetes_http.go
```diff
@@ -380,11 +380,11 @@ func (c configBuilder) nameAndService(ctx context.Context, namespaceService stri
 			return "", nil, err
 		}
 
-		fullName := fullServiceName(svcCtx, namespace, service.Name, service.Port)
+		fullName := fullServiceName(svcCtx, namespace, service, service.Port)
```
dtomcej: If we are passing in the service object, does it make sense to also pass in `service.Port` separately?

ldez: No, because we use this function in another case:
https://github.com/containous/traefik/pull/6087/files#diff-8a53ad85b7fce7f8a74e0d7269162e3bR387

PR #6121, opened 2020-01-02
k8s Ingress: fix crash on rules with nil http

grimmy:
### What does this PR do?

Fixes a crash when using the KubernetesIngress Provider if an Ingress has a rule without a `nil` `http` element.

This is a common case if you're using Let's Encrypt `http01` verification to create a certificate for a non-HTTP based service.  Or in my case, the ingress just uses an annotation to HTTP redirect with ingress-nginx (which i'm migrating away from).

Basically, just adds a `nil` check.  Without this `nil` check a single ingress rule that has a `nil` `http` element will make the entire provider not work.

I also added a unit test to verify the change and to avoid it from happening again in the future.

### Motivation

I'm trying to migrate to traefik from ingress-nginx and after hours of searching for why my secrets weren't being found, I finally dug into the nil pointer deference I was also seeing and it turned out that it was the culprit all along.

### More

- [x] Added/updated tests

### Additional Notes

This is targeting the master branch as during my testing it appeared that 028683666 may have had some impact on the secret resolution, but ultimately I don't think that really matters

### DIFF
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -319,6 +319,10 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 				continue
 			}
 
+			if rule.HTTP == nil {
+				continue
+			}
+
 			for _, p := range rule.HTTP.Paths {
 				service, err := loadService(client, ingress.Namespace, p.Backend)
 				if err != nil {
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -313,57 +313,57 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 				conf.HTTP.Services["default-backend"] = service
 			}
 		}
+
 		for _, rule := range ingress.Spec.Rules {
 			if err := checkStringQuoteValidity(rule.Host); err != nil {
 				log.FromContext(ctx).Errorf("Invalid syntax for host: %s", rule.Host)
 				continue
 			}
 
-			if rule.HTTP == nil {
-				continue
-			}
+			if rule.HTTP != nil {
+				for _, p := range rule.HTTP.Paths {
+					service, err := loadService(client, ingress.Namespace, p.Backend)
+					if err != nil {
+						log.FromContext(ctx).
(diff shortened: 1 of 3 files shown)


pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -313,53 +313,57 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
...
-						WithField("servicePort", p.Backend.ServicePort.String()).
-						Errorf("Cannot create service: %v", err)
-					continue
-				}
-
-				if err = checkStringQuoteValidity(p.Path); err != nil {
-					log.FromContext(ctx).Errorf("Invalid syntax for path: %s", p.Path)
-					continue
-				}
+			if rule.HTTP != nil {
```
mpl: why do we need to get in p.updateIngressStatus if rule.HTTP is nil?

grimmy: The problem is the code right now assumes every ingress has a non-nil `rule.HTTP` which causes it to stack trace.  This just avoid that's for scenarios either other ingresses like `ingress-nginx` where you can have an ingress with an annotation for a redirect but still have a host to make it work.

I can provide said stack trace if you're interested in it.  That said, The `Ingress` I added as a test fixture is basically what I have setup minus the `ingress-nginx`  annotation.

mpl: Oh I understand your PR and the motivation for it. I'm just asking @ldez why he changed your more readable "continue early", to a less readable inverted check. The only reason I see for that, is that we would still want to get into p.updateIngressStatus even when rule.HTTP is nil but I don't know why we would want that.

grimmy: Ah gotcha, I was curious on that as well, but not being very familiar with the style of the project just let it go :)

dtomcej: >why do we need to get in p.updateIngressStatus if rule.HTTP is nil?

Because the systems that rely on ingress status often do not require or interact with `rule.HTTP`, and may continue to function as normal.

mpl: ok, but I was hoping we knew exactly what would break if we just did

    if rule.HTTP == nil {
      continue
    }

instead. but it's not essential, so I'll figure it out myself later.

Issue #5922, opened 2019-12-02
Panic in accesslog middleware


sylr:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I have a kubernetes traefik instances which panics in the accessLog middleware

### Output of `traefik version`: (_What version of Traefik are you using?_)

2af8589a

```
v2.0.5-18-g2af8589a
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
    # traefik.toml
    [log]
      level = "INFO"
      format = "json"
    [accessLog]
      format = "json"
      [accessLog.fields]
        defaultMode = "keep"
    [entryPoints.http]
      address = ":80"
    [entryPoints.http-internal]
      address = ":81"
    [entryPoints.https]
      address = ":443"
    [entryPoints.traefik]
      address = ":8080"
    [entryPoints.traefik-ext]
      address = ":8081"
    [providers.kubernetesIngress]
       ingressClass = "traefik-v2"
    [providers.kubernetesCRD]
    [providers.file]
      directory  = "/config/static"
      watch = true
    [api]
      insecure = true
      dashboard = true
      debug = true
    [metrics.prometheus]
      entryPoint = "traefik"
      buckets = [0.1,0.2,0.3,0.4,0.5,0.6,0.7,0.8,0.9,1.0,1.5,2.0]
```

### If applicable, please paste the log output in DEBUG level (`--log.level=DEBUG` switch)

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x58768f]
goroutine 18618 [running]:
bufio.(*Writer).Available(...)
	/usr/local/go/src/bufio/bufio.go:610
bufio.(*Writer).WriteString(0x0, 0x24d8bdd, 0x19, 0x40c306, 0x7f3ef6644460, 0x28c4885)
	/usr/local/go/src/bufio/bufio.go:691 +0x7f
net/http.(*expectContinueReader).Read(0xc0026fffe0, 0xc00252e000, 0x8000, 0x8000, 0x7f3ef6644460, 0x0, 0x10)
	/usr/local/go/src/net/http/server.go:888 +0x13b
github.com/containous/traefik/v2/pkg/middlewares/accesslog.(*captureRequestReader).Read(0xc00286a060, 0xc00252e000, 0x8000, 0x8000, 0x1f895c0, 0x1, 0xc00252e000)
[...]

PR #6126, opened 2020-01-03
Fix panic in accesslog with recover()

sylr:
### What does this PR do?

To the best of my current knowledge there is no way to not be racing for `Request.Body` with `net/http.(*conn)`.

So the only way to not panic if you happen to try to read a request body after the client connection has been closed (rogue client) is to use `recover()`.

The addition of the mutex is to fix another potential data race I discovered with running traefik built with `-race`.

### Motivation

Fix https://github.com/containous/traefik/issues/5922

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/middlewares/accesslog/capture_request_reader.go
+++ b/pkg/middlewares/accesslog/capture_request_reader.go
@@ -1,18 +1,72 @@
 package accesslog
 
-import "io"
+import (
+	"errors"
+	"net/http"
+	"sync"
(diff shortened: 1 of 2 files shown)


ldez:
The root problem is a nil pointer dereference, so I don't think that a recover is the solution.

It is rather necessary to fix the root of the problem.

So we will declined this PR.

Thanks anyway.

PR #6129, opened 2020-01-04
Add HTTP2 push middleware

### LINKED ISSUE #906: HTTP2 Server Push
Hi :) 

I'm using Traefik with a docker-compose setup. I migrated my express webserver to a library that supports HTTP2 Push : https://github.com/indutny/node-spdy

Everything is working fine when I access directly the server with a proxy : 
image

When i'm passing through Traefik, the site is loading but I don't see the benefits of HTTP2 Server pushing in the debugger

image

Is it something possible ? Is there a particuliar configuration ?

Thank you :)

pipe01:
### What does this PR do?

Add HTTP2 push capability through the use of a middleware (H2Push) that reads the service's returned headers and looks for "Link" headers, then pushes the requested files to the response.

### Motivation

Fixes #906

### More

- [x] Added/updated tests

### Additional Notes

This isn't yet ready for merging, but I wanted to get thoughts from maintainers to see what to improve before polishing it.


brandond:
Looking forward to this so that I can finally drop the extra nginx hop that's doing nothing but push bundling!

byoungb:
> Looking forward to this so that I can finally drop the extra nginx hop that's doing nothing but push bundling!

I could not get the nginx push to work with traefik. @brandond how were you able to do it? I got nginx to work i think but it wouldn't pass through traefik.

SantoDE:
Hey @pipe01,

sorry for the no answer for so long. We were quite busy with keeping up with the PR's. 

After looking at that PR, we think it doesn't really fit with Traefik. But where it will be fitting, is by adding it as a [plugin](https://github.com/containous/plugindemo).

WDYT?

pipe01:
I do agree it doesn't exactly fit into a middleware, however I also think this functionality is common enough to be included into Traefik, so I'm unsure what the best course of action is.

SantoDE:
After further investigation, we think the best place to have it is to transform it to a plugin as mentioned above. Therefore, I'll close that PR.

Thanks for your work and I hope, you'll be fine with a plugin.

Neirda24:
Hi. I'm just a concerned user and wondering why it is best as a plugin when this is a standard in HTTP specification that will come more and more of use every day ?

SantoDE:
For now, we're concerned about potential resource consumption and similiar issues. For now, out of our view, Traefik should focus on being a reverse proxy. Doing HTTP2 push is something, that we more count towards a "webserver-ish" feature and thus, we think it should be in a plugin first.

notsureifkevin:
> 
> 
> Hi. I'm just a concerned user and wondering why it is best as a plugin when this is a standard in HTTP specification that will come more and more of use every day ?

One of the advantages of the plugin system is that we can get an indication of a particular middlewares popularity and usage before accepting it upstream and supporting it officially. Ultimately, this allows the internal development team to stay focused on making Traefik more stable, extendable, and flexible. Thank you for your contribution, and we look forward to seeing this in the new plugin marketplace!

Neirda24:
ok I understand the indicator by the use of plugins. Though I don't understand the part `just a reverse proxy`. Without this functionnality, traefik breaks any PUSH server from upstream. So in the end I must have something like `NginX (doing HTTP/2 features) -> Traefik (reverse proxy) -> NginX (My webservice)` it is kind of weird. I am not the contributor but me too can't wait to see it in the plugin Marketplace.

pipe01:
I will be creating a plugin for this shortly, we'll see if it becomes popular enough.

stafyniaksacha:
Hey @pipe01 ! I'll be interested by your plugin ! Can I help with something ?

pipe01:
I'm currently waiting on #7303, once that's resolved the plugin should be pretty much done

PR #6131, opened 2020-01-04
k8s: make sure to return an error when no healthy endpoints

sylr:
### What does this PR do?

This PR makes sure kubernetes endpoints discovery does not return en empty list of servers but rather an error.

### Motivation

In the case that all pods referenced by a k8s service are unhealthy `kubernetes endpoints discovery` will return an empty list of service which will ultimately end up by a `Service Unavailable` for the client.

By returning an error here I hope that traefik will continue matching the request against other routes so that we can do a fallback.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes_http.go
+++ b/pkg/provider/kubernetes/crd/kubernetes_http.go
@@ -356,6 +356,10 @@ func (c configBuilder) loadServers(fallbackNamespace string, svc v1alpha1.LoadBa
 		}
 	}
 
+	if len(servers) == 0 {
+		return nil, fmt.Errorf("no healthy endpoints found for %s/%s", namespace, sanitizedName)
+	}
+
 	return servers, nil
 }
 
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -255,6 +255,10 @@ func loadService(client Client, namespace string, backend v1beta1.IngressBackend
 		}
 	}
 
+	if len(servers) == 0 {
+		return nil, errors.New("no healthy endpoints found")
+	}
+
 	return &dynamic.Service{
 		LoadBalancer: &dynamic.ServersLoadBalancer{
 			Servers:        servers,


sylr:
FYI I was able to test this patch on v2.1 this morning and it did what I was hoping for.

Given the following ingress routes:

```
kind: IngressRoute
metadata:
  name: go-lev-tls
  namespace: default
spec:
  entryPoints:
  - https
  routes:
  - kind: Rule
    match: Host(`go-lev.company.com`)
    services:
    - name: go-lev
      passHostHeader: true
      port: 80
  tls:
    secretName: company-com-tls
---
kind: IngressRoute
metadata:
  name: go-lev-maintenance-tls
  namespace: default
spec:
  entryPoints:
  - https
  routes:
  - kind: Rule
    match: Host(`go-lev.company.com`)
    priority: 2
    services:
    - name: maintenance
      namespace: maintenance
      passHostHeader: true
      port: 80
  tls:
    secretName: company-com-tls
```

If the kubernetes service `go-lev` were to reference only unhealthy pods then the `go-lev-tls` would be ignored and traefik would consider `go-lev-maintenance-tls`.

Presently if `go-lev` service were to reference only unhealthy pods, traefik would return a 503.

dtomcej:
Hello @sylr,

The current behavior is to return a 503 when there are no healthy endpoints available to serve the request. This behavior acknowledges that Traefik should accept the request, and that there is a valid backend, but it is currently unhealthy/unavailable.

Kubernetes services inform Traefik of state, allowing us to infer that there _should_ be valid endpoints. This inference is what dictates this behavior.

This PR inverts this logic, and I don't think that is correct behavior.

Therefore, I vote to decline this PR.

sylr:
Hi @dtomcej,

Thank you for your reply.

This is indeed a change of bahaviour but it would match the behaviour observed when the kuberbetes service reference no pods at all.

| | No Pods | No Healthy Pods |
|---|---|---|
| Today | Next Rule | 503 |
|With Patch | Next Rule | Next Rule |

dtomcej:
Hello @sylr,

> No pods or no healthy pods should behave alike.

Completely agree.


> No Pods: Next Rule

Are you sure this is the current behavior? If there is a service referenced in an ingress[Route] it should create a Router, regardless of if there are valid endpoints.

It is possible that the No-pods behavior is incorrect...

sylr:
I did not test it, I make sense of it when I saw this:

https://github.com/containous/traefik/blob/49a5af151ea56c91ef3f2916ac36d630865073b9/pkg/provider/kubernetes/crd/kubernetes_http.go#L322-L324

To me it seems that we would hit this code when a service has no pods.

sylr:
So I did the test and if a service has no pods associated the rule is ignored (-> Next Rule).

This behaviour is really handy as it allows to do a fallback.

jbdoumenjou:
We believe that this proposal fits into a more complex problem and we want to think about it more carefully in a dedicated issue. Could you open the issue accordingly with the details of your use case ?

PR #6140, opened 2020-01-07
fix memleak in safe.Pool

### LINKED ISSUE #6125: Using the service mirror causes OOM
### **Do you want to request a feature or report a bug?**
bug

### **What did you do?**
enable service mirror and HTTP performance test

### **What did you expect to see?**
Traefik not error

### **What did you see instead?**
image

image

Then I made a million HTTP requests and stopped, and the memory wasn't reclaimed
image

### **Output of traefik version:**
Traefik viersion:  v2.1

### **What is your environment & configuration (arguments, toml, provider, platform, ...)?**
OS: CentOS 7.6  
Kernel: 3.10.0-957.21.3.el7
Static Config
```
entryPoints:
  web:
    address: ":80"

  websecure:
    address: ":443"

providers:
# Enable the file provider to define routers/middlewares/services in file
  file:
    directory: /etc/traefik/conf
```

Provider Config:
```
http:
  routers:
    to-whoami-https:
      rule: "Host(`aaa.com`) && PathPrefix(`/`)"
      entryPoints:
      - "websecure"
      service: mirrored-api
      tls: {}
    to-whoami-http:
      rule: "Host(`aaa.com`) && PathPrefix(`/`)"
      entryPoints:
      - "web"
      service: mirrored-api

  services:
    mirrored-api:
      mirroring:
        service: appv1
        mirrors:
        - name: appv2
          percent: 100

    appv1:
      loadBalancer:
        servers:
        - url: http://xxxxa/

    appv2:
      loadBalancer:
        servers:
        - url: http://xxxxxb/
tls:
  certificates:
    - certFile: /etc/traefik/ssl/test.pem
      keyFile: /etc/traefik/ssl/test.key
      stores:
        - default
```

mpl:
### What does this PR do?

safe.Pool is not a pool in the usual sense as it is not bounded and
it keeps on appending new goroutines, instead of reusing a limited
number of goroutines. Moreover, even when a goroutine is done, there's
still a small footprint of it, as there is a reference to it that was
appended in a slice, and that never gets removed.

This PR fixes the pool by using maps instead of slices, and by making sure the entry in the map corresponding to a goroutine is removed from the map when the goroutine terminates.

### Motivation

Fixes #6125

### More

- ~[ ] Added/updated tests~

### Additional Notes

### DIFF
--- a/pkg/safe/routine.go
+++ b/pkg/safe/routine.go
@@ -2,7 +2,6 @@ package safe
 
 import (
 	"context"
-	"crypto/rand"
 	"fmt"
 	"runtime/debug"
 	"sync"
@@ -20,14 +19,13 @@ type routineCtx func(ctx context.Context)
 
 // Pool is a pool of go routines
 type Pool struct {
-	routines    map[string]*routine   // keyed by a randomly generated id
-	routinesCtx map[string]routineCtx // keyed by a randomly generated id
-	waitGroup   sync.WaitGroup
-	lock        sync.Mutex
-	baseCtx     context.Context
-	baseCancel  context.CancelFunc
-	ctx         context.Context
-	cancel      context.CancelFunc
+	routines   []routine
+	waitGroup  sync.WaitGroup
+	lock       sync.Mutex
+	baseCtx    context.Context
+	baseCancel context.CancelFunc
+	ctx        context.Context
+	cancel     context.CancelFunc
 }
 
 // NewPool creates a Pool
@@ -35,12 +33,10 @@ func NewPool(parentCtx context.Context) *Pool {
 	baseCtx, baseCancel := context.WithCancel(parentCtx)
 	ctx, cancel := context.WithCancel(baseCtx)
 	return &Pool{
-		baseCtx:     baseCtx,
-		baseCancel:  baseCancel,
-		ctx:         ctx,
-		cancel:      cancel,
-		routinesCtx: make(map[string]routineCtx),
-		routines:    make(map[string]*routine),
+		baseCtx:    baseCtx,
+		baseCancel: baseCancel,
+		ctx:        ctx,
+		cancel:     cancel,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


pkg/safe/routine.go
```diff
@@ -46,19 +48,55 @@ func (p *Pool) Ctx() context.Context {
 	return p.baseCtx
 }
 
+func genID() string {
```
dtomcej: shouldn't we use a UUID if this is supposed to be random?

mpl: I (honestly) don't know. why? what's the problem with the current approach that UUID would solve?

dtomcej: the larger number of connections you have, the larger the risk of a collision.

That would result in a random thread being removed no?

Using a UUID would prevent this because it is almost assuredly unique

mpl: > the larger number of connections you have, the larger the risk of a collision.
> 
> That would result in a random thread being removed no?

in theory yes, in practice no. :-)
I think I've made sure that we never reuse an ID that already exists in the Pool.
But I'll look at UUIDs anyway, I wonder what the overhead is.

> Using a UUID would prevent this because it is almost assuredly unique

mpl: @dtomcej looking at the implementation here: https://github.com/google/uuid/blob/master/version4.go#L29 , it looks like what we're doing in genID is pretty similar, and at least as strong wrt to collisions (maybe even stronger, since we have 160 bits instead of 122), so I'm not worried about leaving it as is. WDYT?

dtomcej: LGTM

pkg/safe/routine.go
```diff
@@ -20,7 +21,7 @@ type routineCtx func(ctx context.Context)
 // Pool is a pool of go routines
 type Pool struct {
 	routines    []routine
```
juliens: You should do the same with this slice?

mpl: Yes we should. But maybe in next PR, since it is not needed to fix #6125 ?

mpl: Done

pkg/safe/routine.go
freundallein: Sorry for breaking in. I accidentally saw your conversation. What if we use ints instead of strings, as map keys? Then we can use unix timestamp and atomic counter, it will allocate only 64 bit per key (for example, twitter snowflakes unique ids)
Something like this:

```
func (p *Pool) genID() uint64 {
    var id uint64
    epoch := uint64(time.Now().UTC().UnixNano()) / 10000
    id = epoch << (64 - 48)                      // 48 bits for epoch
    id |= uint64(p.counter % (1<<16 - 1))        // 16 bits for counter
    atomic.AddUint64(&p.counter, 1)
    return id
```
Pool's counter can be uint64, for example. It will take much time to overload it.
And we can hold 65536 calls per 0.01 ms.
What do you think?

mpl: thanks, but unless necessary, I'd rather not make genID more complex than it needs to be. especially now that I've seen that the uuid package does something pretty similar.

PR #6148, opened 2020-01-09
systematically call updateIngressStatus

mpl:
### What does this PR do?

This PR moves the call to updateIngressStatus so that it is always called, whereas previously it would not have been called in .e.g some error cases. While we were at it, also added a small refactor to unindent some code.

### Motivation

This should have done (or at least discussed) in https://github.com/containous/traefik/pull/6121, but IMO the review got rushed.

### More

- ~[ ] Added/updated tests~

### DIFF
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -242,33 +242,34 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 				continue
 			}
 
-			if rule.HTTP != nil {
-				for _, pa := range rule.HTTP.Paths {
-					if err = checkStringQuoteValidity(pa.Path); err != nil {
-						log.FromContext(ctx).Errorf("Invalid syntax for path: %s", pa.Path)
-						continue
-					}
-
-					service, err := loadService(client, ingress.Namespace, pa.Backend)
-					if err != nil {
-						log.FromContext(ctx).
-							WithField("serviceName", pa.Backend.ServiceName).
-							WithField("servicePort", pa.Backend.ServicePort.String()).
-							Errorf("Cannot create service: %v", err)
-						continue
-					}
-
-					serviceName := provider.Normalize(ingress.Namespace + "-" + pa.Backend.ServiceName + "-" + pa.Backend.ServicePort.String())
-					conf.HTTP.Services[serviceName] = service
-
-					routerKey := strings.TrimPrefix(provider.Normalize(rule.Host+pa.Path), "-")
-					conf.HTTP.Routers[routerKey] = loadRouter(ingress, rule, pa, rtConfig, serviceName)
-				}
+			if err := p.updateIngressStatus(ingress, client); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -320,52 +320,56 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
...
-					service, err := loadService(client, ingress.Namespace, p.Backend)
-					if err != nil {
-						log.FromContext(ctx).
-							WithField("serviceName", p.Backend.ServiceName).
-							WithField("servicePort", p.Backend.ServicePort.String()).
-							Errorf("Cannot create service: %v", err)
-						continue
-					}
+			if rule.HTTP == nil {
+				if err := p.updateIngressStatus(ingress, client); err != nil {
```
dtomcej: Could we not move this to line 322, and remove from 372 since we update this status whether HTTP is nil or not?

mpl: @dtomcej If we want to keep the same logic as before (and I'm not saying we should, as I am not sure), then no we cannot. Because if you look inside the for loop L330, there are error cases in there where we continue, which means in these cases we currently do not call p.updateIngressStatus. Then again, maybe it's ok to call it anyway in such a case, I'm still not sure (which is also why I'm opening this PR, as I want us to be clear on that matter). WDYT?

dtomcej: I _personally_ think that if we are processing the ingress, we should update the status, regardless of what sort of configuration we end up building from it.

The updated status makes it clear that Traefik processed it, and induces users to look back to traefik for configuration debugging if something is not routing correctly.

I think that we should _always_ update the status, unless we cannot, or there is a severe error that prevents us from doing so.

dtomcej: @timoreimann What do you think about this?

mpl: @dtomcej fine by me, if no one else thinks that it could be cause for problems. It sure would make the code even simpler. (it actually was @juliens initial suggestion too).

timoreimann: Does @dtomcej's suggestion imply that we could enable routing to endpoints even for cases where the routing might knowingly be broken, whereas today we'd keep those endpoints from being taken into rotation (and allowing Traefik to continue serving traffic correctly in an old but otherwise working configuration)?

If yes, then I'd be concerned we start introducing broken forwarding rules that we previously protected users from (which is what all these error checks are for). If no, then I think I'm good with Daniel's proposal.

dtomcej: Nope, This is only about adding the loadbalancer information into the status section of the ingress, nothing more. Right now we don't copy that information in all cases.

It does not change any of the routing rules or configurations.

mpl: @dtomcej Sounds good to us, thanks.

PR #6151, opened 2020-01-10
fix useXForwardedFor ip white list match error

gefire:
### What does this PR do?

fix useXForwardedFor=true ip white list match error in some env

### Motivation

if use F5 or other LB  and  LB's address in ip whitelist , then users can always access the hosts.

### More

- [ ] Added/updated tests

### DIFF
--- a/whitelist/ip.go
+++ b/whitelist/ip.go
@@ -60,6 +60,7 @@ func (ip *IP) IsAuthorized(req *http.Request) error {
 	if ip.useXForwardedFor {
 		xFFs := req.Header[XForwardedFor]
 		if len(xFFs) > 0 {
+			match := false
 			for _, xFF := range xFFs {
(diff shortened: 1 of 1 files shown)


mpl:
@gefire Sorry but we do not understand at all what is the problem you are trying to fix, or what you are trying to do, so we cannot review that PR until then. Can you please open an issue detailing the problem you are facing?

gefire:
thks for view, issue #6158 .

ldez:
The current behavior is the expected behavior, the following test prove that:

https://github.com/containous/traefik/blob/d24f78222c6fb54e2f8ad2052f261145e8eacbba/whitelist/ip_test.go#L46-L53

`useXForwardedFor` don't means "use only `X-Forwarded-For` Header"

PR #6160, opened 2020-01-13
Add Ingress annotations support

### LINKED ISSUE #6074: Improvements of the Ingress Support in Traefik v2

Feature

In Traefik, we used to support Kubernetes routing via the Kubernetes Ingress Object (which is quite limited in itself) and a set of annotations (which are intrinsically non-standardized).
We then switched to supporting Kubernetes routing mainly through a Custom Resource Definition, as explained with more detail in [this announcement](https://containo.us/blog/back-to-traefik-2-0-2f9aa17be305/#kubernetes-crd).
We also kept a basic - no annotations - support of the Kubernetes Ingress Object.

However, a fair amount of users must stick with Ingress for a while because they cannot (for various reasons) switch easily to using CRDs.
As a result, they end up facing limitations with what they can do with respect to routing capabilities, because they lack quite a number of features that are now available only through CRDs.

We have heard their concerns. Therefore, we are going to try and remedy the situation by allowing the Ingress provider to configure some of the Traefik router options through annotations.
Based on feedback, we already have a good grasp of what needs to be done, and we're pretty much set on the number of features and router options to include: TLS, entryPoints, and references to middleware definitions.

If you think we may have forgotten some critical aspects, feel free to speak up on this issue.

ldez:
### What does this PR do?

Add Ingress annotations support.

### Motivation

Fixes #6074

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: jbdoumenjou <jb.doumenjou@gmail.com>

### DIFF
--- a/pkg/provider/kubernetes/ingress/annotations.go
+++ b/pkg/provider/kubernetes/ingress/annotations.go
@@ -1,0 +1,108 @@
+package ingress
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/containous/traefik/v2/pkg/config/dynamic"
+	"github.com/containous/traefik/v2/pkg/config/label"
+)
+
+const (
+	// https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
+	annotationsPrefix = "traefik.ingress.kubernetes.io/"
+)
+
+// RouterConfig is the router's root configuration from annotations.
+type RouterConfig struct {
+	Router *RouterIng `json:"router,omitempty"`
+}
+
+// RouterIng is the router's configuration from annotations.
+type RouterIng struct {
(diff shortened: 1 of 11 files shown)


pkg/provider/kubernetes/ingress/annotations.go
```diff
@@ -0,0 +1,108 @@
...
+
+// ServiceIng is the service's configuration from annotations.
+type ServiceIng struct {
+	ServersScheme  string          `json:"serversScheme,omitempty"`
+	PassHostHeader *bool           `json:"passHostHeader"`
+	Sticky         *dynamic.Sticky `json:"sticky,omitempty" label:"allowEmpty"`
+}
+
+// SetDefaults sets the default values.
+func (r *ServiceIng) SetDefaults() {
```
mmatur: Could you rename `r`

PR #6180, opened 2020-01-15
Don't nest middleware spans

aliksend:
### What does this PR do?

Converts this
before
to this
(I've also added a "long" auth middleware to be more revealing)
after

In other words, my PR makes middleware span to display only time when middleware is processing, not all nested middleware and forward spans.

### Motivation

To display correct span times in tracing UI

### More

- [ ] ~Added/updated tests~

### DIFF
--- a/pkg/middlewares/tracing/wrapper.go
+++ b/pkg/middlewares/tracing/wrapper.go
@@ -61,7 +61,8 @@ type TraceStarter struct {
 }
 
 func (ts *TraceStarter) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	_, err := tracing.FromContext(req.Context())
+	parentContext := req.Context()
+	_, err := tracing.FromContext(parentContext)
 	if err != nil {
 		if ts.middleware != nil {
 			ts.middleware.ServeHTTP(rw, req)
@@ -74,6 +75,7 @@ func (ts *TraceStarter) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 	if ts.finisher != nil {
 		ts.finisher.finish = ts.finish
+		ts.finisher.ctx = parentContext
 	}
 
 	if ts.middleware != nil {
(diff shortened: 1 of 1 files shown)


aliksend:
That was my fail. It is a duplicate of #6013 

I accidentally deleted the fork after closing other my PR

pkg/middlewares/tracing/wrapper.go
```diff
@@ -21,48 +21,99 @@ func Wrap(ctx context.Context, constructor alice.Constructor) alice.Constructor
...
 }
 
-// NewWrapper returns a http.Handler struct
-func NewWrapper(next http.Handler, name string, spanKind ext.SpanKindEnum) http.Handler {
-	return &Wrapper{
-		next:     next,
-		name:     name,
-		spanKind: spanKind,
+// NewTraceStarter returns a *TracingStarted struct
+func NewTraceStarter(middleware http.Handler, name string, spanKind ext.SpanKindEnum, finisher *TraceFinisher) *TraceStarter {
```
mpl: Neither this constructor, nor any of the new types in this file, need to be exported, do they?

pkg/middlewares/tracing/wrapper.go
```diff
@@ -21,48 +21,99 @@ func Wrap(ctx context.Context, constructor alice.Constructor) alice.Constructor
...
+		return
+	}
+
+	_, req, ts.finishFn = tracing.StartSpan(req, ts.name, ts.spanKind)
+	defer ts.finish()
+
+	if ts.finisher != nil {
+		ts.finisher.finish = ts.finish
+		ts.finisher.ctx = parentContext
+	}
```
mpl: instead of having the extra ts.finish method, and the extra ts.finishFn and ts.finished fields, couldn't we simply handle all the logic with:

```
_, req, finishFn = tracing.StartSpan(req, ts.name, ts.spanKind)
if ts.finisher != nil {
    ts.finisher.finish = finishFn
    ts.finisher.ctx = parentContext
} else {
    defer finishFn()
}
````
?

sylr:
I personally wouldn't do that because middlewares are nested within each other and can do as much after the request has been forwarded as beforehand (like accesslog or metrics middlewares).

juliens:
@aliksend Thank you for your contribution!

We talked about it in the team and we think the fact that the middlewares are nested is normal. The middlewares are really nested because, you can have code before the next middleware, and after the next middleware, so it's not okay to "stop" a middleware span when the next middleware is called.

For me, it's like code profiling, the duration of a function is the sum of its own time + the entire duration of the functions it calls.

So I will decline your PR, but as always in open source, “No is temporary, yes is forever.” , so if you think we are really wrong, don't hesitate to open a new proposal to explain more what you think.

PR #6193, opened 2020-01-17
more priority rule example

hchoong:
### What does this PR do?

Add more detail and example documentation on ambigious priority 

### Motivation

Have documentation matches code behavior, https://github.com/containous/traefik/blob/master/pkg/rules/rules.go#L56-L58

### More

- [ ] Added/updated tests


ldez:
Hello,

> If the length of a rule with `0` priority greater than another rule's priority value, the fore rule precedes the latter.

Your sentence is confusing for non native speaker.

and the example break the chaining between examples: the example in `Set priorities` is the answer to "To solve this issue, the priority must be set."

So, I will declined your PR.

Thanks anyway.


https://docs.traefik.io/v2.1/routing/routers/#priority

PR #6194, opened 2020-01-18
Adding an explanation how to use `htpasswd` for k8s secret

jamct:
# What does this PR do?

The tip to use htpasswd was not very clear. You have to encode the result to base64. Adding an example command.

### Motivation

Make it easier to use basic auth and generate a base64 encoded string.

### More


ldez:
Hello,

You don't need to encode the result to base64.

I think you are using a docker-compose file, so you need to escape `$` as explained [here](https://docs.traefik.io/v2.1/middlewares/basicauth/#configuration-examples)

You can try it with this simple example:

```yml
version: "3.7"

services:

  traefik:
    image: traefik:v2.1.2
    command:
      - --entrypoints.web.address=:80
      - --entrypoints.websecure.address=:443
      - --api
      - --providers.docker.exposedbydefault=false
    ports:
      - 80:80
      - 443:443
    labels:
      traefik.enable: true

      traefik.http.routers.traefik.rule: Host(`traefik.localhost`)
      traefik.http.routers.traefik.service: api@internal
      traefik.http.routers.traefik.entrypoints: web

      # echo $(htpasswd -bnBC 10 admin secret) | sed -e s/\\$/\\$\\$/g
      # curl -u admin:secret http://whoami.localhost
      traefik.http.middlewares.auth.basicauth.users: admin:$$2y$$10$$zdGj/1e2U9K1S1kSyI3dqu.XLOkMwvU7RFuR.54.Wwy36iSrB13te

      # htpasswd -bnBC 10 admin admin | openssl base64
      # curl -u admin:admin http://whoami.localhost
      # Doesn't work: level=error msg="error parsing BasicUser: YWRtaW46JDJ5JDEwJEZHYVN0YnR2Q0JjZWswdGczUWdMYU8yWWViMHNLSlpZT3o4STgub2w1OFNQTmNQRjd5WXpPCgo="
      # traefik.http.middlewares.auth.basicauth.users: YWRtaW46JDJ5JDEwJEZHYVN0YnR2Q0JjZWswdGczUWdMYU8yWWViMHNLSlpZT3o4STgub2w1OFNQTmNQRjd5WXpPCgo=

    volumes:

  whoami:
    image: containous/whoami:v1.4.0
    labels:
      traefik.enable: true

      traefik.http.routers.whoami.rule: Host(`whoami.localhost`)
      traefik.http.routers.whoami.entrypoints: web
      traefik.http.routers.whoami.middlewares: auth
```

jamct:
@ldez: I tried to deploy on Kubernetes and got an error saying "invalid base64..."
It worked when I encoded it to base64.

ldez:
So the base64 is not required for all providers: only for K8s secret.

The base64 encoding is a requirement of `Secret`, as you can see in the IngressRoute example.

So, you can add a comment in the IngressRoute example to explain that, but that cannot be a global tip.

Could you edit your PR, thank you.

jamct:
Thanks @ldez for clarification and quick approval.

docs/content/middlewares/basicauth.md
```diff
@@ -107,6 +107,10 @@ spec:
     secret: authsecret
 
 ---
+# Note: in a kubernetes secret the string (e.g. generated by htpasswd) must be base64-encoded first.
+# To create user:password pair, it's possible to use this command:
```
mpl:
```suggestion
# To create an encoded user:password pair, the following command can be used:
```

PR #6206, opened 2020-01-20
Use EDF schedule algorithm for WeightedRoundRobin

pkumza:
### What does this PR do?

Use EDF schedule algorithm for WeightedRoundRobin
(https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling)

Each pick from the schedule has the earliest deadline entry selected. Entries have deadlines set at current time + 1 / weight, providing weighted round robin behavior with floating point weights and an O(log n) pick time.

Picking up complexity of EDF algorithm is stable, but current simple algorithm sometimes takes O(n) pick time.

[Envoy](https://github.com/envoyproxy/envoy) uses the same EDF algorithm for WRR.

### Motivation

Current WRR algorithm has significant bias problem for unbalanced weight endpoints.

e.g.

Endpoints: [{"name":"A", "weight": 11}, {"name":"B", "weight": 3}]

| picking up algorithm | sequence |
| --- | --- |
|current  | A A A A A A A A A B A B A B|
|EDF | A A A B A A A A B A A A B A|

### More

- [x] Added/updated tests


pkumza:
@ldez rebased

pkumza:
`continuous-integration/travis-ci/pr` got a **429(Too many requests)** Error for link [https://containo.us/blog/traefik-and-docker-a-discussion-with-docker-captain-bret-fisher-7f0b9a54ff88/](https://containo.us/blog/traefik-and-docker-a-discussion-with-docker-captain-bret-fisher-7f0b9a54ff88/) in checking external links. 😓

Definitely, that's not my fault.
I didn't modify anything in docs directory.
This PR would pass the checks if it could be run again.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -151,5 +149,15 @@ func (b *Balancer) AddService(name string, handler http.Handler, weight *int) {
 	if weight != nil {
 		w = *weight
 	}
-	b.handlers = append(b.handlers, &namedHandler{Handler: handler, name: name, weight: w})
+	if w <= 0 { // non-positive weight is meaningless
+		return
+	}
+	h := &namedHandler{Handler: handler, name: name, weight: float64(w)}
+	h.deadline = b.curDeadline + 1/h.weight
```
mpl: b.curDeadline is concurrently mutated in nextServer, so shouln't there be a read lock to protect it here?

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -151,5 +149,15 @@ func (b *Balancer) AddService(name string, handler http.Handler, weight *int) {
...
+	if h.deadline < b.baseDeadline {
+		h.deadline = math.Ceil((b.baseDeadline-h.deadline)*h.weight) / h.weight
+	}
+	b.curIndex++
```
mpl: similarly,  I think this write on b.curIndex needs to be locked, no?
also, I don't actually understand why we want to increment curIndex here. Shouldn't it be only updated whenever a handler is selected, just like with curDeadline?

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -37,79 +44,70 @@ func New(sticky *dynamic.Sticky) *Balancer {
...
+	old := *n
+	h := old[len(old)-1]
+	*n = old[0 : len(old)-1]
+	return h
+}
+
+// Balancer is a WeightedRoundRobin load balancer based on Earliest Deadline First (EDF).
+// (https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling)
+// Each pick from the schedule has the earliest deadline entry selected. Entries have deadlines set
+// at current time + 1 / weight, providing weighted round robin behavior with floating point
```
mpl: "current time" does not seem like an accurate description. the algorithm does not actually use time anywhere afaics.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -151,5 +149,15 @@ func (b *Balancer) AddService(name string, handler http.Handler, weight *int) {
...
 	}
-	b.handlers = append(b.handlers, &namedHandler{Handler: handler, name: name, weight: w})
+	if w <= 0 { // non-positive weight is meaningless
+		return
+	}
+	h := &namedHandler{Handler: handler, name: name, weight: float64(w)}
+	h.deadline = b.curDeadline + 1/h.weight
+	if h.deadline < b.baseDeadline {
+		h.deadline = math.Ceil((b.baseDeadline-h.deadline)*h.weight) / h.weight
+	}
```
mpl: Could you explain this part please? I understand neither its rationale, nor its intended effect.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -37,79 +44,70 @@ func New(sticky *dynamic.Sticky) *Balancer {
...
+	// Pick handler with closest deadline.
+	handler := heap.Pop(b.handlers).(*namedHandler)
+	// curDeadline should be handler's deadline so that new added entry would have a fair
+	// competition environment with the old ones.
+	b.curDeadline = handler.deadline
+	handler.deadline += 1 / handler.weight
+	b.curIndex++
+	handler.index = b.curIndex
+	heap.Push(b.handlers, handler)
+	log.WithoutContext().Debugf("Service Select: %s", handler.name)
```
mpl: let's update this log message, which wasn't great to begin with. how about:
("Service selected by WRR: %s", handler.name)
?

pkg/server/service/loadbalancer/wrr/wrr.go
pkumza: I'd like to remove this.

---

It has effects.
Suppose we have 1000 hosts.
900 of them have weight 100, as a1, a2, a3, ..., a900
100 of them have weight 1, as b1, b2, b3,...,b100
so, in the picking loop, we always pick ax first.
We need to pick a about 90000 times before we pick the first b.
Sometimes, when ServeHTTP is called 50000 times, endpoints refreshed, so we pick a1, a2, a3 again, as b is never picked.

the Ceil is used for a random pick in the cycle. it makes b has the opportunity to be picked first.

pkg/server/service/loadbalancer/wrr/wrr.go
pkumza: fixed

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -37,79 +44,70 @@ func New(sticky *dynamic.Sticky) *Balancer {
...
-	for _, s := range b.handlers {
-		if s.weight > max {
-			max = s.weight
-		}
+// Len implements heap.Interface/sort.Interface
+func (n namedHandlers) Len() int { return len(n) }
+
+// Less implements heap.Interface/sort.Interface
+func (n namedHandlers) Less(i, j int) bool {
+	// We want Pop to give us the highest, not lowest, priority so we use greater than here.
```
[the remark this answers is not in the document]
pkumza: fixed

pkg/server/service/loadbalancer/wrr/wrr.go
pkumza: fixed

pkg/server/service/loadbalancer/wrr/wrr.go
pkumza: Removed

pkg/server/service/loadbalancer/wrr/wrr.go
pkumza: Thanks, good suggestion.
fixed.

pkg/server/service/loadbalancer/wrr/wrr.go
pkumza: index is removed.

pkumza:
Hi @mpl,
 
I met a CI problem but I didn't change anything about `simple_test.go`.

I met this problem since merge the *namedHandler problem, so I'd believe it's something wrong the master branch.

```
----------------------------------------------------------------------
FAIL: simple_test.go:608: SimpleSuite.TestUDPServiceConfigErrors

simple_test.go:620:
    c.Assert(err, checker.IsNil)
... value *errors.errorString = &errors.errorString{s:"try operation failed: could not find '[\"the udp service \\\"service1@file\\\" does not have any type defined\"]' in body '[{\"status\":\"enabled\",\"usedBy\":[\"router4@file\"],\"name\":\"service1@file\",\"provider\":\"file\"},{\"loadBalancer\":{\"servers\":[{\"address\":\"127.0.0.1:9010\"}]},\"status\":\"enabled\",\"usedBy\":[\"router5@file\"],\"name\":\"service2@file\",\"provider\":\"file\",\"type\":\"loadbalancer\"}]\n'"} ("try operation failed: could not find '[\"the udp service \\\"service1@file\\\" does not have any type defined\"]' in body '[{\"status\":\"enabled\",\"usedBy\":[\"router4@file\"],\"name\":\"service1@file\",\"provider\":\"file\"},{\"loadBalancer\":{\"servers\":[{\"address\":\"127.0.0.1:9010\"}]},\"status\":\"enabled\",\"usedBy\":[\"router5@file\"],\"name\":\"service2@file\",\"provider\":\"file\",\"type\":\"loadbalancer\"}]\n'")
```

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -37,79 +42,69 @@ func New(sticky *dynamic.Sticky) *Balancer {
...
+// (https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling)
+// Each pick from the schedule has the earliest deadline entry selected. Entries have deadlines set
+// at currentDeadline + 1 / weight, providing weighted round robin behavior with floating point
+// weights and an O(log n) pick time.
+type Balancer struct {
+	handlers     *namedHandlers
+	mutex        *sync.Mutex
+	curDeadline  float64
+	curLock      sync.RWMutex
+	baseDeadline float64 // for fixing starting bias
```
mpl: I think you're not using baseDeadline anymore, so you can remove it.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -37,79 +42,69 @@ func New(sticky *dynamic.Sticky) *Balancer {
...
+// Balancer is a WeightedRoundRobin load balancer based on Earliest Deadline First (EDF).
+// (https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling)
+// Each pick from the schedule has the earliest deadline entry selected. Entries have deadlines set
+// at currentDeadline + 1 / weight, providing weighted round robin behavior with floating point
+// weights and an O(log n) pick time.
+type Balancer struct {
+	handlers     *namedHandlers
+	mutex        *sync.Mutex
+	curDeadline  float64
+	curLock      sync.RWMutex
```
mpl: why did you add a new mutex?
I think you can simply change mutex L84 to a RWMutex, and use it to guard all the fields in Balancer that require it, no? and then you can remove curLock.

pkumza:
> Ok, I will look into why the CI fails and I will let you know.
> 
> Regarding the part about baseDeadline (sorry, github is so shitty that it does not let me reply to you in the right place, even though I unresolved the conversation): why do you want to remove it if you think it is useful? (I need to think about your explanation some more myself anyway).

---

The best practice of the "Ceil" is going along with rand.Seed(time.UnixNano()).
That could makes cases unstable.
and I am wondering if add a `rand.Seed(time.UnixNano())` in NewScheduler is suitable.

What's more, it cannot fix the biasing for non-weighted situation.
It could be better to do several random picking action before normal picking for actually serve HTTP. Which need `rand.Seed(time.UnixNano())` too.

To make the Pull Request pure, I'd like ignore these features unless traefik meets many bad cases in the future that really need to modify the algorithm.

pkg/server/service/loadbalancer/wrr/wrr.go
pkumza: Sorry for my careless, fix this soon.

mpl:
> > Ok, I will look into why the CI fails and I will let you know.
> > Regarding the part about baseDeadline (sorry, github is so shitty that it does not let me reply to you in the right place, even though I unresolved the conversation): why do you want to remove it if you think it is useful? (I need to think about your explanation some more myself anyway).
> 
> The best practice of the "Ceil" is going along with rand.Seed(time.UnixNano()).
> That could makes cases unstable.
> and I am wondering if adding a `rand.Seed(time.UnixNano())` in NewScheduler is suitable.
> 
> What's more, it cannot fix the biasing for non-weighted situation.
> It could be better to do several random picking action before normal picking for actually serve HTTP. Which need `rand.Seed(time.UnixNano())` too.
> 
> To make the Pull Request pure, I'd like ignore these features unless traefik meets many bad cases in the future that really need to modify the algorithm.

Ok, that sounds reasonable. However, just to be clear, regarding the example you gave in the Motivation part of the PR message, i.e.:
A A A A A A A A A B A B A B becoming A A A B A A A A B A A A B A thanks to the use of EDF; does that still apply without the baseDeadline random initialisation trick? In other words, with the PR as it is right now, does it still fix the main problem/example that you were describing in the motivation?

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -23,9 +25,10 @@ type stickyCookie struct {
 
 // New creates a new load balancer.
 func New(sticky *dynamic.Sticky) *Balancer {
+	handlers := make(namedHandlers, 0)
 	balancer := &Balancer{
-		mutex: &sync.Mutex{},
-		index: -1,
+		handlers: &handlers,
+		mutex:    &sync.RWMutex{},
```
mpl: This line is probably not needed.
But see my comment L82 first.

pkumza:
Yes, it still works without the baseDeadline random initialisation trick.
@mpl

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -37,79 +40,65 @@ func New(sticky *dynamic.Sticky) *Balancer {
...
+}
+
+// Balancer is a WeightedRoundRobin load balancer based on Earliest Deadline First (EDF).
+// (https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling)
+// Each pick from the schedule has the earliest deadline entry selected. Entries have deadlines set
+// at currentDeadline + 1 / weight, providing weighted round robin behavior with floating point
+// weights and an O(log n) pick time.
+type Balancer struct {
+	handlers     *namedHandlers
+	mutex        *sync.RWMutex
```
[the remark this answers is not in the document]
pkumza: fixed~

pkumza:
> Ok, I will look into why the CI fails and I will let you know.

It seems like pushing another new commit may cause the previous CI failed. 
Maybe there's no problem for continuous-integration/travis-ci/pr when checking `simple_test.go:608`

pkumza:
> > Ok, I will look into why the CI fails and I will let you know.
> 
> It seems like pushing another new commit may cause the previous CI failed.
> Maybe there's no problem for continuous-integration/travis-ci/pr when checking `simple_test.go:608`

I take my words back...
CI surely has problem.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -151,5 +140,13 @@ func (b *Balancer) AddService(name string, handler http.Handler, weight *int) {
 	if weight != nil {
 		w = *weight
 	}
-	b.handlers = append(b.handlers, &namedHandler{Handler: handler, name: name, weight: w})
+	if w <= 0 { // non-positive weight is meaningless
```
mpl: I understand why you're doing that, but I am not sure we can afford it.
There was nothing preventing users adding a service w <=0 before, and I do not think it was ignored. In fact, if i understand correctly the old nextServer algorithm, even servers with a weight of 0 were taken into account when picking a server. We only had an error if absolutely all the weights were equal to 0 at first glance.
Usually the least we could do is document that AddService will ignore a service if its weight is <=0, but I don't think that would be enough. if we don't want to break our existing users, we need to make the whole algorithm valid within the same parameters as the old algorithm.

But maybe I am wrong, and maybe the old nextServer algorithm actually ignored any service with a weight <=0, even thought it was added to the list of services. We need to make sure of that. I will double-check and ask around.

mpl: OK, nevermind that last remark. Even if the old algorithm was not clear on the matter, the intent was indeed to ignore any service with a weight <=0. So let's document AddService in consequence. I will add a suggestion in a minute.

pkumza: Added

pkumza:
@mpl problem solved and all checks passed. 
Please review again~

pkg/server/service/loadbalancer/wrr/wrr_test.go
```diff
@@ -112,3 +112,24 @@ func TestSticky(t *testing.T) {
...
+		rw.WriteHeader(http.StatusOK)
+	}), Int(3))
+
+	recorder := &responseRecorder{ResponseRecorder: httptest.NewRecorder(), save: map[string]int{}}
+	for i := 0; i < 4; i++ {
+		balancer.ServeHTTP(recorder, httptest.NewRequest(http.MethodGet, "/", nil))
+	}
+	assert.Less(t, recorder.save["first"], 4)
+	assert.Less(t, 0, recorder.save["second"])
+}
```
mpl: I found that your example in the PR message was a good illustration of the problem, so I propose we change TestBalancerBias so that it represents perfectly that illustration. That way, people reading the git history can actually even better see what the problem was (and also we can avoid regressions). So, I propose:

```
// TestBalancerBias makes sure that the WRR algorithm spreads elements evenly
// right from the start, and that it does not "over-favor" the high-weighted ones
// with a biased start-up regime.
func TestBalancerBias(t *testing.T) {
	balancer := New(nil)

	balancer.AddService("first", http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
		rw.Header().Set("server", "A")
		rw.WriteHeader(http.StatusOK)
	}), Int(11))

	balancer.AddService("second", http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
		rw.Header().Set("server", "B")
		rw.WriteHeader(http.StatusOK)
	}), Int(3))

	recorder := &responseRecorder{ResponseRecorder: httptest.NewRecorder(), save: map[string]int{}}
	for i := 0; i < 14; i++ {
		balancer.ServeHTTP(recorder, httptest.NewRequest(http.MethodGet, "/", nil))
	}
	wantSequence := []string{"A", "A", "A", "B", "A", "A", "A", "A", "B", "A", "A", "A", "B", "A"}
	assert.Equal(t, wantSequence, recorder.sequence)
}
```

with the responseRecorder changed accordingly:

```
func (r *responseRecorder) WriteHeader(statusCode int) {
	r.save[r.Header().Get("server")]++
	r.sequence = append(r.sequence, r.Header().Get("server"))
	r.ResponseRecorder.WriteHeader(statusCode)
}
```

WDYT?

pkumza: Yeah, I agree with you.

Once I thought of adding such a test, but after the `ceil` trick, it changes every time, so the idea was abort. 
Since we have removed  the random trick from this LB algorithm, it would be perfect to add this. 

Fixing in minutes...

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -37,79 +39,66 @@ func New(sticky *dynamic.Sticky) *Balancer {
...
+// Balancer is a WeightedRoundRobin load balancer based on Earliest Deadline First (EDF).
+// (https://en.wikipedia.org/wiki/Earliest_deadline_first_scheduling)
+// Each pick from the schedule has the earliest deadline entry selected. Entries have deadlines set
+// at currentDeadline + 1 / weight, providing weighted round robin behavior with floating point
+// weights and an O(log n) pick time.
+type Balancer struct {
+	stickyCookie *stickyCookie
+
+	mutex       sync.RWMutex
+	handlers    *namedHandlers
```
mpl: I think this field does not have to be a pointer, and everywhere needed we could call the methods on &b.handlers instead of b.handlers, but I don't have any strong argument in favor of that for now, so I'll defer to the other reviewers.

mpl: After consulting with co-workers, here is a reason why we should try to do better: if we move the implementation of Pop (and others) one level up, i.e. to the Balancer instead of directly on the handlers, then we might be able to remove the extra indirection (handlers would become a slice again instead of a pointer), which would remove a bit of complexity. Can you please try to see if it can be done?

pkumza: Solved.

mpl: Perfect, it indeed all looks much better now, thanks.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -37,79 +36,63 @@ func New(sticky *dynamic.Sticky) *Balancer {
...
+	b.handlers = append(b.handlers, h)
 }
 
-func gcd(a, b int) int {
-	for b != 0 {
-		a, b = b, a%b
-	}
-	return a
+// Pop implements heap.Interface for poping an item from the heap
+func (b *Balancer) Pop() interface{} {
```
mpl: Now that all the methods are on an exported type, it opens the door to panics due to misuses, such as calling b.Pop() on a freshly initialised Balancer. However, I don't think it is our job to make the methods robust to such misuses, since the callers have b.Len() that allows them to call b.Pop() (and others) responsibly.
However, let's at least document document that fact, by adding a line above, such as:

```
// It panics if b.Len() < 1.
```

I don't think we need to do it for Less and Swap, as the panics from them would be really obvious imho.

PR #6226, opened 2020-01-23
Add http request scheme to logger

### LINKED ISSUE #6130: Access logs should contain the request scheme
Feature

I'm using `json` access logs, and I realized there is no way to figure out if a request is sent over http or https from the logs. It would be nice to have the request scheme logged.

valtlfelipe:
### What does this PR do?

Added to the logger the http request scheme HTTP/HTTPS.

### Motivation

Fixes #6130 

### More

- [x] Added/updated tests

### Additional Notes

Got solution from https://github.com/golang/go/issues/28940#issuecomment-441749380.
Need to add test for the https scheme, but could not figure it out. If someone could give me some directions please.

### DIFF
--- a/pkg/middlewares/accesslog/logger.go
+++ b/pkg/middlewares/accesslog/logger.go
@@ -193,6 +193,11 @@ func (h *Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next http
 	core[RequestPath] = urlCopyString
 	core[RequestProtocol] = req.Proto
 
+	core[RequestScheme] = "http"
+	if req.TLS != nil {
+		core[RequestScheme] = "https"
+	}
+
 	core[ClientAddr] = req.RemoteAddr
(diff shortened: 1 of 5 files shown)


pkg/middlewares/accesslog/logger.go
```diff
@@ -193,6 +193,11 @@ func (h *Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next http
 	core[RequestPath] = urlCopyString
 	core[RequestProtocol] = req.Proto
 
+	core[RequestScheme] = "http"
+	if req.TLS != nil {
+		core[RequestScheme] = "https"
+	}
```
mpl: what about ws and wss? do we want to take them into account? do we want to look at the headers to see the connection upgrade maybe?

mpl: nevermind, it's probably irrelevant.

PR #6240, opened 2020-01-26
Fix dnspod update.

iineva:
### What does this PR do?

Fix dnspod certificate update.

### Motivation

I have encountered this problem. The certificate cannot be update correctly.

### Issue log

```
acme: error cleaning up: API call failed: json: cannot unmarshal number into Go struct field DomainInfo.share_total of type string
```

### Additional Notes

Dnspod updated API struct. `DomainInfo.ShareTotal` from `string` to `int`.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -477,12 +477,11 @@
   version = "v1.1.0"
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


ldez:
Hello,

you cannot change the files in the `vendor` folder by hand.

PR #6255, opened 2020-01-30
Add metrics about TLS

sylr:
### What does this PR do?

Add metrics about TLS version and TLS cipher used

### Motivation

TLS configuration can have a major effect on CPU usage so this allows to see what TLS versions and ciphers are used.

### More

- [x] Added/updated tests


sylr:
Fullscreen_03_02_2020_11_56

pkg/middlewares/metrics/metrics.go
```diff
@@ -131,6 +144,88 @@ func getMethod(r *http.Request) string {
 	return r.Method
 }
 
+func getRequestTLSVersion(req *http.Request) string {
+	var tlsVersion string
+
+	switch req.TLS.Version {
+	case tls.VersionTLS10:
+		tlsVersion = "tls-1.0"
```
dtomcej: This really isn't in a format that is commonly used, could these labels be updated to something more standard?

dtomcej:
```suggestion
		tlsVersion = "TLS 1.0"
```

pkg/middlewares/metrics/metrics.go
```diff
@@ -131,6 +144,88 @@ func getMethod(r *http.Request) string {
...
+	case tls.VersionTLS11:
+		tlsVersion = "tls-1.1"
+	case tls.VersionTLS12:
+		tlsVersion = "tls-1.2"
+	case tls.VersionTLS13:
+		tlsVersion = "tls-1.3"
+	case tls.VersionSSL30: //nolint
```
dtomcej: why nolint here?

pkg/middlewares/metrics/metrics.go
```diff
@@ -131,6 +144,88 @@ func getMethod(r *http.Request) string {
...
+
+	return tlsVersion
+}
+
+func getRequestTLSCipher(req *http.Request) string {
+	var tlsCipher string
+
+	switch req.TLS.CipherSuite {
+	case tls.TLS_RSA_WITH_RC4_128_SHA:
+		tlsCipher = "TLS_RSA_WITH_RC4_128_SHA"
```
dtomcej:
```suggestion
		return "TLS_RSA_WITH_RC4_128_SHA"
```

pkg/middlewares/metrics/metrics.go
sylr: Because VersionSSL30 will be removed in Go 1.14 so the linter throws an error about that.

pkg/middlewares/metrics/metrics.go
sylr: At first I only put the version number as value and then when I noticed there still was SSL v3 in go I added the protocol name.

I don't really care, I just don't really like spaces in label values.

pkg/middlewares/metrics/metrics.go
sylr: Addressed.

pkg/middlewares/metrics/metrics.go
sylr: Addressed.

pkg/middlewares/metrics/metrics.go
```diff
@@ -131,6 +144,84 @@ func getMethod(r *http.Request) string {
...
+	default:
+		tlsVersion = "unknown"
+	}
+
+	return tlsVersion
+}
+
+func getRequestTLSCipher(req *http.Request) string {
+	switch req.TLS.CipherSuite {
+	case tls.TLS_RSA_WITH_RC4_128_SHA:
```
dtomcej: on second thought here, we already have all of the supported suites and their string equivalents stored in https://github.com/containous/traefik/blob/master/pkg/tls/certificate.go#L35, don't you think you could find the value in the map and return the associated string key? That would prevent having to add/update this cipher list if the certificate.go suite list is updated.

This would also apply to the version above.

Does that seem simpler and more extensible?

sylr: I'm not comfortable to loop over `pkg/tls.CipherSuites` at each request to find the string that correspond to its uint16 code.

So I made a reverse map and a test that ensure that both maps are always in sync.

What do you think ?

pkg/middlewares/metrics/metrics.go
SuperSandro2000: Shouldn't SSL (not TLS) be dead dead?

pkg/middlewares/metrics/metrics.go
```diff
@@ -74,13 +79,22 @@ func WrapServiceHandler(ctx context.Context, registry metrics.Registry, serviceN
 }
 
 func (m *metricsMiddleware) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	var labels []string
+	labels := make([]string, 0, len(m.baseLabels)+6)
```
ldez: could you remove the `make`.

pkg/middlewares/metrics/metrics.go
```diff
@@ -74,13 +79,22 @@ func WrapServiceHandler(ctx context.Context, registry metrics.Registry, serviceN
...
 	labels = append(labels, m.baseLabels...)
 	labels = append(labels, "method", getMethod(req), "protocol", getRequestProtocol(req))
 
 	m.openConnsGauge.With(labels...).Add(1)
 	defer m.openConnsGauge.With(labels...).Add(-1)
 
+	// TLS metrics
+	if req.TLS != nil {
+		tlsLabels := make([]string, 0, len(m.baseLabels)+4)
```
ldez: could you remove the `make`.

pkg/tls/cipher_test.go
```diff
@@ -0,0 +1,32 @@
...
+func TestCiphersMapsSync(t *testing.T) {
+	for k, v := range CipherSuites {
+		// Following names are legacy aliases. We do not test for their presence
+		// in CipherSuitesReversed
+		switch k {
+		case "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305":
+			fallthrough
+		case "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305":
+			continue
+		}
```
ldez:
```suggestion
		switch k {
		case "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305", "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305":
			continue
		}
```

pkg/tls/cipher_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+		if rv, ok := CipherSuitesReversed[v]; !ok {
+			t.Errorf("Maps not in sync: `%v` key is missing in tls.CipherSuitesReversed", v)
```
ldez:
```suggestion
			t.Errorf("Maps not in sync: `%d` key is missing in tls.CipherSuitesReversed", v)
```

pkg/tls/cipher_test.go
```diff
@@ -0,0 +1,32 @@
...
+		} else if k != rv {
+			t.Errorf("Maps not in sync: tls.CipherSuites[%v] = `%v` AND tls.CipherSuitesReversed[`%v`] = `%v`", k, v, v, rv)
```
ldez:
```suggestion
			t.Errorf("Maps not in sync: tls.CipherSuites[%s] = `%d` AND tls.CipherSuitesReversed[`%d`] = `%v`", k, v, v, rv)
```

pkg/tls/cipher_test.go
```diff
@@ -0,0 +1,32 @@
...
+		}
+	}
+
+	for k, v := range CipherSuitesReversed {
+		if rv, ok := CipherSuites[v]; !ok {
+			t.Errorf("Maps not in sync: `%v` key is missing in tls.CipherSuites", v)
```
ldez:
```suggestion
			t.Errorf("Maps not in sync: `%s` key is missing in tls.CipherSuites", v)
```

pkg/tls/cipher_test.go
```diff
@@ -0,0 +1,32 @@
...
+		} else if k != rv {
+			t.Errorf("Maps not in sync: tls.CipherSuitesReversed[`%v`] = `%v` AND tls.CipherSuites[`%v`] = `%v`", k, v, v, rv)
```
ldez:
```suggestion
			t.Errorf("Maps not in sync: tls.CipherSuitesReversed[`%d`] = `%s` AND tls.CipherSuites[`%s`] = `%d`", k, v, v, rv)
```

pkg/tls/cipher_test.go
```diff
@@ -0,0 +1,32 @@
+package tls
+
+import (
+	"testing"
+)
+
+func TestCiphersMapsSync(t *testing.T) {
+	for k, v := range CipherSuites {
+		// Following names are legacy aliases. We do not test for their presence
+		// in CipherSuitesReversed
```
ldez:
```suggestion
		// Following names are legacy aliases. We do not test for their presence in CipherSuitesReversed
```

pkg/middlewares/metrics/metrics.go
```diff
@@ -131,6 +145,33 @@ func getMethod(r *http.Request) string {
...
+		tlsVersion = "1.1"
+	case tls.VersionTLS12:
+		tlsVersion = "1.2"
+	case tls.VersionTLS13:
+		tlsVersion = "1.3"
+	default:
+		tlsVersion = "unknown"
+	}
+
+	return tlsVersion
```
ldez:
```suggestion
	switch req.TLS.Version {
	case tls.VersionTLS10:
		return "1.0"
	case tls.VersionTLS11:
		return "1.1"
	case tls.VersionTLS12:
		return "1.2"
	case tls.VersionTLS13:
		return "1.3"
	default:
		return "unknown"
	}
```

pkg/middlewares/metrics/metrics.go
sylr: I would prefer to keep it. We know the cap of the slice so it is optimized to initialize it with the correct cap so that following `append()` calls do not need to grow it.

ldez: I would prefer to remove it

sylr:
It should be good now.

pkg/middlewares/metrics/metrics.go
ldez: These changes are required to merge the PR.

ldez: If these changes are not made quickly, the PR will not be part of the next release.

vearutop: @ldez out of curiosity, what is your preference based on?

ldez: - avoid magic number
- optimization need to be proved

rules-optimization

vearutop: ok, no meaningful reason, thanks and have a nice day!

ldez: There are no meaningful reason do use the `make` without knowing if this has an impact.

sylr:
```go
package main

import "testing"

func BenchmarkNoCap(b *testing.B) {
	for i := 0; i < b.N; i++ {
		var slice []string
		slice = append(slice, []string{"a", "q", "w"}...)
		slice = append(slice, []string{"z", "s", "x"}...)
	}
}

func BenchmarkCap(b *testing.B) {
	for i := 0; i < b.N; i++ {
		slice := make([]string, 0, 6)
		slice = append(slice, []string{"a", "q", "w"}...)
		slice = append(slice, []string{"z", "s", "x"}...)
	}
}
```

```
$ go test -bench . 
goos: darwin
goarch: amd64
BenchmarkNoCap-12    	 7565632	       146 ns/op
BenchmarkCap-12      	57868803	        21.0 ns/op
PASS
ok  	_/Users/s.rabot/tmp/go-cert	3.505s
```

ldez: micro benchmarck prove nothing on real and concrete impact in the context of this code..

PR #6266, opened 2020-02-03
Add consul catalog filtering to v2

chuckyz:
### What does this PR do?

Adds https://www.consul.io/api/features/filtering.html to Traefik

### Motivation

At present, Traefik retrieves the entire catalog and then sends two requests (health, and catalog) per service.  This is a way to ensure the traffic sent/received is as small as possible.

### More

### Additional Notes

Getting the entire catalog is not bad, however getting all services and health of all services may be an issue.  With this change, it allows some endpoints (e.g.: v1/catalog/service/foo) to return `[]` if the filter is not met.  We have some particularly large services which return hundreds of responses for both service/ and health/ endpoints, and simply returning `[]` would be nicer.

Unfortunately the actual catalog itself does not have a filter endpoint outside some node-level filtering.

I would like to propose a followup of filtering the actual catalog/services endpoint within the code itself as well.  This would not be too difficult, as the entire response body of it is a json blob with keys of service names, and tags inside of them.  I believe this would be best done inside of the `fetchServices` function, with another new config flag (I propose "catalogFilter").

### DIFF
--- a/docs/content/providers/consul-catalog.md
+++ b/docs/content/providers/consul-catalog.md
@@ -158,6 +158,32 @@ providers:
 
 Use local agent caching for catalog reads.
 
+### `filter`
+
+_Optional_
+
+```toml tab="File (TOML)"
+[providers.consulCatalog]
+  filter = "traefik.enable=true in ServiceTags"
+  # ...
+```
+
+```yaml tab="File (YAML)
+providers:
+  consulCatalog:
+    filter: "traefik.enable=true in ServiceTags"
+    # ...
+```
+
+```bash tab="CLI"
+--providers.consulcatalog.filter="traefik.enable=true in ServiceTags"
+# ...
+```
+
+Defines the Consul filter to send to the server.  See more details at https://www.consul.io/api/features/filtering.html.
+
+Suggested for deployments with large catalogs.
+
 ### `endpoint`
 
 Defines the Consul server endpoint.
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

The filtering on `traefik.enable=true` is already done by the tag filter on `fetchService`.
You have to use [exposedbydefault](https://docs.traefik.io/v2.1/providers/consul-catalog/#exposedbydefault) and put a tag (`traefik.enable=true`) on service.

From the doc, inside this PR, I think that was the goal of this PR

ldez:
I created a PR #6307, that use our standard filtering system ([constraints](https://docs.traefik.io/v2.1/providers/consul-catalog/#constraints), [exposedbydefault](https://docs.traefik.io/v2.1/providers/consul-catalog/#exposedbydefault), [restrict-the-scope-of-service-discovery](https://docs.traefik.io/v2.1/providers/overview/#restrict-the-scope-of-service-discovery).

I think it's better to filter catalog services when we retrieve them, this will reduce the number of requests much more.

For now, the [filtering](https://www.consul.io/api/features/filtering.html) from the catalog, as it don't works with [`/catalog/services`](https://www.consul.io/api/catalog.html#list-services), and it don't have much more benefit than our filtering system, seems to not be the best solution.

So we will decline this PR, but thank you anyways for putting forward the problem of the number of requests and filtering.

traefiker:
Closed by #6307.

chuckyz:
Thank you @ldez !  I did something very similar in my 1.7 branch, and I feel much better about our upgrade to 2.1 now. :)

PR #6270, opened 2020-02-03
Add TLSStores to Kubernetes CRD

### LINKED ISSUE #5468: ability to specify default TLS certificate for Kubernetes CRD

Feature

Currently it is possible to configure default TLS certificate via TOML/YAML using:
```toml
[tls.stores]
  [tls.stores.default]
    [tls.stores.default.defaultCertificate]
      certFile = "path/to/cert.crt"
      keyFile  = "path/to/cert.key"
```

with KubernetesCRD, however, it is not possible, because the only things we can configure are those via using CRDs (`tlsoption`, `ingressroute`, `middleware`, `ingressroutetcp`) and CLI, or by mounting `configmap`/`secret` containing the tls.crt and tls.key to Traefik container.

As far as I understand it, the latest is not possible because Traefik generates a self-signed TLS cert at runtime and stores it in memory.

CLI does not have such an option.

dtomcej:
### What does this PR do?

This PR Adds the TLS Stores functionality to the CRD, allowing kubernetes users the ability to provide default certificates without using config files and mounting certificates.

### Motivation

Fixes #5468 

### More

- [x] Added/updated tests - more may be needed

### Additional Notes

For some reason `make validate` seemed to update the kv-ref doc, but idk why 🤷‍♂


dtomcej:
Blocked on TLS Store naming question

pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -509,6 +510,49 @@ func buildTLSOptions(ctx context.Context, client Client) map[string]tls.Options
...
+
+	for _, tlsStore := range tlsStoreCRD {
+		namespace := tlsStore.Namespace
+		logger := log.FromContext(log.With(ctx, log.Str("tlsStore", tlsStore.Name), log.Str("namespace", namespace)))
+		secretName := tlsStore.Spec.DefaultCertificate.SecretName
+
+		secret, exists, err := client.GetSecret(namespace, secretName)
+		if err != nil {
+			logger.Errorf("Failed to fetch secret %s/%s: %v", namespace, secretName, err)
+			return tlsStores
```
mpl: I'm not so sure about this (and the others below) return.
If we are in "failure mode", shouldn't we rather return nil, instead of a half-processed tlsStores? otoh, if we don't consider this error as critical, shouldn't we keep going through the stores? (with a continue statement)

pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -509,6 +510,49 @@ func buildTLSOptions(ctx context.Context, client Client) map[string]tls.Options
...
+
+		if _, ok := tlsStores["default"]; !ok {
+			// Since only the default TLS Store is used, we just need to set it if it doesn't exist.
+			tlsStores["default"] = tls.Store{
+				DefaultCertificate: &tls.Certificate{
+					CertFile: tls.FileOrContent(cert),
+					KeyFile:  tls.FileOrContent(key),
+				},
+			}
+		}
```
mpl: So, just to be clear:
we allow for the CRDs to contain several TLS stores, but we only take into account the first valid one we find while iterating through them?
If yes, why do we go through the trouble of iterating through all of them then?
Why do not simply change line 543 to:

    if _, ok := tlsStores["default"]; ok {
        break
    }

?

pkg/provider/kubernetes/crd/kubernetes.go
dtomcej: I think that the continue would be better, as it would be non-blocking, and would match the failure process of TLS Options.

pkg/provider/kubernetes/crd/kubernetes.go
dtomcej: Why not?

Because I'm not as awesome as you are.

<3

pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -507,6 +508,52 @@ func buildTLSOptions(ctx context.Context, client Client) map[string]tls.Options
...
+			break
+		}
+
+		// Since only the default TLS Store is used, we just need to set it if it doesn't exist.
+		tlsStores["default"] = tls.Store{
+			DefaultCertificate: &tls.Certificate{
+				CertFile: tls.FileOrContent(cert),
+				KeyFile:  tls.FileOrContent(key),
+			},
+		}
```
mpl: So, my last remark was under the current algorithm assumption, i.e. that we only wanted to keep whatever first store we find and call it the "default" one. But after chatting with the team, I don't think that's what we want to do.

We need to make sure that the day we start supporting stores other than the default one, we do not break the pre-existing behaviour for a given config/user-provided CRD. I think that means it's not the role of the config builder here to make the call about whether or not a TLS store is a useful one (aka the default one). In other words, each of them should be kept, with:

    tlsStores[tlsStore.Name] = blabla

and then it's up to the route handlers to pick up the tlsStore they need/want in that map later on. even though we know that for now they'll only ever pick the tlsStores["default"] on).

If in doubt, I think you can look at how things behave in pkg/provider/file/file.go on that matter.

dtomcej: The issue here is: What do we take the store name to be?

Do we take it to be the name of the kubernetes CRD object? (note that it will more than likely not be "default", or there could be a "default" in every namespace).

mpl: well yeah, I don't see what other choice we have than the object name. Is there any other possibility/suggestion?

dtomcej: Not really without making any changes.

Ideally we use the same `makeID` as the TLSOptions, but as soon as you add the namespace into the mix, it will never make a "default" key.

I think this is the best option for now.

docs/content/routing/providers/kubernetes-crd.md
```diff
@@ -890,6 +890,75 @@ or referencing TLS options in the [`IngressRoute`](#kind-ingressroute) / [`Ingre
 	just as in the [middleware case](../../middlewares/overview.md#provider-namespace).
 	Specifying a namespace attribute in this case would not make any sense, and will be ignored.
 
+### Kind: `TLSStore`
+
+`TLSStore` is the CRD implementation of a [Traefik "TLS Store"](../../https/tls.md#certificates-stores).
+
+Register the `TLSStore` kind in the Kubernetes cluster before creating `TLSStore` objects
+or referencing TLS stores in the [`IngressRoute`](#kind-ingressroute) / [`IngressRouteTCP`](#kind-ingressroutetcp) objects.
+
```
mpl: maybe we add a warning/info here that only a TLSStore named "default" will be actually used by Traefik?

pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -507,6 +508,48 @@ func buildTLSOptions(ctx context.Context, client Client) map[string]tls.Options
...
+	tlsStores = make(map[string]tls.Store)
+
+	for _, tlsStore := range tlsStoreCRD {
+		namespace := tlsStore.Namespace
+		logger := log.FromContext(log.With(ctx, log.Str("tlsStore", tlsStore.Name), log.Str("namespace", namespace)))
+		secretName := tlsStore.Spec.DefaultCertificate.SecretName
+
+		secret, exists, err := client.GetSecret(namespace, secretName)
+		if err != nil {
+			logger.Errorf("Failed to fetch secret %s/%s: %v", namespace, secretName, err)
```
mpl: maybe we add somme context:

Failed to fetch secret %s/%s for TLS store %s: %v, namespace, secretName, tlsStore.Name, err

pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -507,6 +508,48 @@ func buildTLSOptions(ctx context.Context, client Client) map[string]tls.Options
...
+			continue
+		}
+		if !exists {
+			logger.Errorf("Secret %s/%s does not exist", namespace, secretName)
+			continue
+		}
+
+		cert, key, err := getCertificateBlocks(secret, namespace, secretName)
+		if err != nil {
+			logger.Errorf("Could not get certificate blocks: %v", err)
```
mpl: is the secretName going to be mentioned in the error message? if not, we should probably add it ourselves for extra context?

btw, I've just realized; couldn't we add the secretName to the logger L522, so it's always in all error messages?

docs/content/routing/providers/kubernetes-crd.md
dtomcej: Yes, this is probably a good idea, since it could be confusing behavior if not documented.

pkg/provider/kubernetes/crd/kubernetes.go
dtomcej: The secret name is inserted into the error on L633, but its not a bad idea to ensure that it is always logged.

PR #6271, opened 2020-02-04
Consul Catalog to read the token from the env

greut:
### What does this PR do?

The Consul Catalog token will be read from the environment using the usual name `CONSUL_HTTP_TOKEN`. It's usually not a big secret but it feels nicer to have it separately from the rest of the configuration (if needed)

### Motivation

I didn't like having a secret in my config.

### More

- [ ] Added/updated tests

### Additional Notes

Any recommendations on how to set up a test for this?


mpl:
Hello, and thank you for your contribution.
We understand your problem but we want to think further about it, because we think we might rather come up with a broader scope solution. And we don't want to commit to supporting your solution right now, if it turns out it does not fit well with the broader one that might emerge.
Notably, this kind of problem might be included in the scope of this proposal: https://github.com/containous/traefik/issues/3853

greut:
@mpl thanks, I'll follow its resolution closely then. Cheers,

PR #6277, opened 2020-02-04
Improve rate-limiting doc

### LINKED ISSUE #4549: Unclear rate limiting explanation

Bug

### What did you see?

The documentation for Rate Limiting currently reads:

> In the above example, frontend1 is configured to limit requests by the client's ip address.
> An average of 5 requests every 3 seconds is allowed and an average of 100 requests every 10 seconds.
> These can "burst" up to 10 and 200 in each period respectively.

The second statement, 

> These can "burst" up to 10 and 200 in each period respectively.

is confusing. Does this mean that in any 1s period that the request limit is the "burst" limit? Better explanation of the "period", "average", and "burst" fields would be helpful.

```
1.7.4
```

mpl:
### What does this PR do?

Expand a bit on the meaning of the configuration parameters for rate-limiting.

### Motivation

Fixes #4549 

### More

- ~[ ] Added/updated tests~

### DIFF
--- a/docs/configuration/commons.md
+++ b/docs/configuration/commons.md
@@ -256,7 +256,7 @@ Another way to describe the above parameters, is to use the [leaky bucket](https
 for rateset1, the size of the bucket is 200 drops, and it is leaking at a rate of 10 drop/s.
 If the incoming rate of drops falling into the bucket gets high enough that the bucket gets filled,
 any subsequent drop overflows out of the bucket (i.e. the request is discarded).
-That situation holds until the incoming rate gets low enough again for the water level in the bucket to go down.
+This situation holds until the incoming rate gets low enough again, and remains that way, for the water level in the bucket to go down.
 
 Valid values for `extractorfunc` are:
   * `client.ip`
--- a/docs/configuration/commons.md
+++ b/docs/configuration/commons.md
@@ -249,8 +249,14 @@ Multiple sets of rates can be added to each frontend, but the time periods must 
 ```
 
 In the above example, frontend1 is configured to limit requests by the client's ip address.  
-An average of 100 requests every 10 seconds is allowed and an average of 5 requests every 3 seconds.  
-These can "burst" up to 200 and 10 in each period respectively. 
+A sustained rate of 100 requests every 10 seconds (10 req/s) is allowed for rateset1, and 5 requests every 3 seconds (~1.67 req/s) for rateset2.  
+In addition, these can "burst" up to 200 and 10 in each period respectively.
+
+Another way to see the above parameters, is along the [leaky bucket](https://en.wikipedia.org/wiki/Leaky_bucket) analogy:
+for rateset1, the size of the bucket is 200 drops, and it is leaking at a rate of 10 drop/s.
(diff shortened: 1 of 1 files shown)


docs/configuration/commons.md
```diff
@@ -249,8 +249,14 @@ Multiple sets of rates can be added to each frontend, but the time periods must
 ```
 
 In the above example, frontend1 is configured to limit requests by the client's ip address.  
-An average of 100 requests every 10 seconds is allowed and an average of 5 requests every 3 seconds.  
-These can "burst" up to 200 and 10 in each period respectively. 
+A sustained rate of 100 requests every 10 seconds (10 req/s) is allowed for rateset1, and 5 requests every 3 seconds (~1.67 req/s) for rateset2.  
+In addition, these can "burst" up to 200 and 10 in each period respectively.
+
+Another way to see the above parameters, is along the [leaky bucket](https://en.wikipedia.org/wiki/Leaky_bucket) analogy:
```
dtomcej:
```suggestion
Another way to describe the above parameters, is to use the [leaky bucket](https://en.wikipedia.org/wiki/Leaky_bucket) analogy:
```

docs/configuration/commons.md
```diff
@@ -249,8 +249,14 @@ Multiple sets of rates can be added to each frontend, but the time periods must
...
+for rateset1, the size of the bucket is 200 drops, and it is leaking at a rate of 10 drop/s.
+If the incoming rate of drops falling into the bucket gets high enough that the bucket gets filled,
+any subsequent drop overflows out of the bucket (i.e. discarded request).
```
dtomcej:
```suggestion
any subsequent drop overflows out of the bucket (i.e. the request is discarded).
```

docs/configuration/commons.md
```diff
@@ -249,8 +249,14 @@ Multiple sets of rates can be added to each frontend, but the time periods must
...
+That situation holds until the incoming rate gets low enough again for the water level in the bucket to go down.
```
dtomcej:
```suggestion
This situation holds until the incoming rate remains low enough for the water level in the bucket to go down.
```

mpl: hmm I don't think "remains" works here. we're in the state where the incoming rate is too high (requests are getting discarded), and this sentence describes how we're getting back to a state where the incoming rate is lower than what we allow in.

dtomcej: The idea being that just not sending requests for a millisencond will not set the rate to 0, and allow all requests.

The incoming rate has to remain low enough for long enough that the bucket stops overflowing.

mpl: how about the new wording I edited your suggestion into?

dtomcej: Looks good to me!

mpl: @dtomcej cool. suggestion committed.

george-angel:
Hello - this was really useful to find, what happened to this explanation, my git foo is weak to find where it was removed and why. Thank you.

mpl:
@george-angel yes

PR #6306, opened 2020-02-12
Docs: Clarifying format of ingress endpoint service name

BretFisher:
### What does this PR do?

Updates a doc page to clarify the format of the value for `providers.kubernetesingress.ingressendpoint.publishedservice`

### Motivation

I used the key=value wrong based on the example. Once I saw the traefik log output saying I needed to enter the value in the format of namespace/service my simple setup worked. I'm assuming for this doc change that it always needs to be in that format for any use of `publishedservice` across TOML/YAML/arg.

### DIFF
--- a/docs/content/providers/kubernetes-ingress.md
+++ b/docs/content/providers/kubernetes-ingress.md
@@ -326,7 +326,8 @@ providers:
 --providers.kubernetesingress.ingressendpoint.publishedservice=namespace/foo-service
 ```
 
-Published Kubernetes Service to copy status from. In the format of `namespace/service`.
+Published Kubernetes Service to copy status from.
+Format: `namespace/servicename`.
 
 ### `throttleDuration`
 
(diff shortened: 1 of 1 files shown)


docs/content/providers/kubernetes-ingress.md
```diff
@@ -298,23 +298,23 @@ _Optional, Default: empty_
...
     # ...
 ```
 
 ```bash tab="CLI"
---providers.kubernetesingress.ingressendpoint.publishedservice=foo-service
+--providers.kubernetesingress.ingressendpoint.publishedservice=namespace/foo-service
 ```
 
-Published Kubernetes Service to copy status from.
+Published Kubernetes Service to copy status from. In the format of `namespace/service`.
```
dtomcej:
```suggestion
Published Kubernetes Service to copy status from.
Format: `namespace/servicename`.
```

PR #6325, opened 2020-02-16
Improvement of the unique name of the router for Ingress.

### LINKED ISSUE #6320: HTTP and HTTPS router to same domain and path from Kubernetes Ingress Provider
This is almost a repost of #6317 , as I believe the suggested fix :
- Does not fix the issue
- Is beside the point of the bug report, as one might want to answer both HTTP and HTTPS on the same entrypoint.

However, I still applied the suggested fix in this bug report

Bug

### What did you do?

Using a build from branch master (built Feb 13th, 2020 around 18:00 UTC) of Traefik 2.x, I generated a Middleware redirecting to HTTPS, an Ingress listening on / at example.org on HTTP referencing the Middleware, and an Ingress listening on / at example.org on HTTPS with a certificate, associated to a simple whoami service

- `curl -vkL http://example.org` gets the HTTP router, gets Redirected to https://example.org, and shows the page provided by the whoami service
-  The Traefik Dashboard should show two routers with the same rules, two different names, one with TLS, and the other without TLS

### What did you see instead?

- the curl on http either 
  - gets redirected to the https which answers a 404 not found with the default (incorrect) certificate
  - answers directly a 404, in which case a `curl https://example.org/` answers the correct page with the correct certificate
 Both possibilities change at random over time
- Traefik Dashboard shows only one Router matching the rule `Host("example.org") && PathPrefix("/")` sometimes with TLS support, sometimes without

```
Version:      7c430e5c9de5305b2f58dce77cdf50eeba944a25
Codename:     cheddar
Go version:   go1.13.7
Built:        2020-02-13_04:06:02PM
OS/Arch:      linux/amd64
```

```yaml
kind: Middleware
metadata:
  name: force-https
  namespace: default
spec:
  redirectScheme:
    scheme: https
    permanent: false
---
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: echo-http-notls
  namespace: default
  annotations:
    traefik.ingress.kubernetes.io/router.middlewares: "default-force-https@kubernetescrd"
    traefik.ingress.kubernetes.io/router.entrypoints: "web"
spec:
  rules:
  - host: "example.org"
    http:
      paths:
      - path: /
        backend:
          serviceName: echo-http
          servicePort: 8080
---
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: echo-http
  namespace: default
  annotations:
    traefik.ingress.kubernetes.io/router.entrypoints: "websecure"
    cert-manager.io/issuer: "acme-staging"
spec:
  tls:
  - hosts:
    - example.org
    secretName: test-ingress-crt
  rules:
  - host: "example.org"
    http:
      paths:
      - path: /
        backend:
          serviceName: echo-http
          servicePort: 8080
```
No relevant log output found in log in DEBUG level

ldez:
### What does this PR do?

Use the namespace and the name of the Ingress to create a more unique router name.

### Motivation

Fixes #6320

### More

- [x] Added/updated tests
- ~~[ ] Added/updated documentation~~

### DIFF
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -267,7 +267,7 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 				conf.HTTP.Services[serviceName] = service
 				conf.HTTP.Services[serviceName] = service
 
-				routerKey := strings.TrimPrefix(provider.Normalize(rule.Host+pa.Path), "-")
+				routerKey := strings.TrimPrefix(provider.Normalize(ingress.Namespace+"-"+ingress.Name+"-"+rule.Host+pa.Path), "-")
 				conf.HTTP.Routers[routerKey] = loadRouter(ingress, rule, pa, rtConfig, serviceName)
 			}
 		}
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -267,7 +267,7 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 				conf.HTTP.Services[serviceName] = service
 				conf.HTTP.Services[serviceName] = service
 
-				routerKey := strings.TrimPrefix(provider.Normalize(ingress.Namespace+"-"+ingress.Name+"-"+rule.Host+pa.Path), "-")
+				routerKey := strings.TrimPrefix(provider.Normalize(ingress.Name+"-"+ingress.Namespace+"-"+rule.Host+pa.Path), "-")
 				conf.HTTP.Routers[routerKey] = loadRouter(ingress, rule, pa, rtConfig, serviceName)
 			}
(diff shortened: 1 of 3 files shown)


pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -267,7 +267,7 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 				conf.HTTP.Services[serviceName] = service
 				conf.HTTP.Services[serviceName] = service
 
-				routerKey := strings.TrimPrefix(provider.Normalize(rule.Host+pa.Path), "-")
+				routerKey := strings.TrimPrefix(provider.Normalize(ingress.Namespace+"-"+ingress.Name+"-"+rule.Host+pa.Path), "-")
```
dtomcej: Usually the order is name -> namespace for k8s things. Do you think we could use this here?

ldez: fixed

PR #6327, opened 2020-02-17
Add UDP in providers with labels

juliens:
### What does this PR do?
This PR handles UDP on all providers with labels .

### Motivation
Being able to define UDP load balancing using labels in rancher, consulcatalog, docker and marathon.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/configuration.go
+++ b/pkg/provider/configuration.go
@@ -46,6 +46,12 @@ func Merge(ctx context.Context, configurations map[string]*dynamic.Configuration
 	routersTCPToDelete := map[string]struct{}{}
 	routersTCP := map[string][]string{}
 
+	servicesUDPToDelete := map[string]struct{}{}
+	servicesUDP := map[string][]string{}
+
+	routersUDPToDelete := map[string]struct{}{}
+	routersUDP := map[string][]string{}
+
 	middlewaresToDelete := map[string]struct{}{}
 	middlewares := map[string][]string{}
 
@@ -85,6 +91,20 @@ func Merge(ctx context.Context, configurations map[string]*dynamic.Configuration
 			}
 		}
 
+		for serviceName, service := range conf.UDP.Services {
+			servicesUDP[serviceName] = append(servicesUDP[serviceName], root)
+			if !AddServiceUDP(configuration.UDP, serviceName, service) {
+				servicesUDPToDelete[serviceName] = struct{}{}
+			}
+		}
+
+		for routerName, router := range conf.UDP.Routers {
+			routersUDP[routerName] = append(routersUDP[routerName], root)
+			if !AddRouterUDP(configuration.UDP, routerName, router) {
+				routersUDPToDelete[routerName] = struct{}{}
+			}
+		}
+
 		for middlewareName, middleware := range conf.HTTP.Middlewares {
 			middlewares[middlewareName] = append(middlewares[middlewareName], root)
 			if !AddMiddleware(configuration.HTTP, middlewareName, middleware) {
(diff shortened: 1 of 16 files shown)


pkg/provider/configuration.go
```diff
@@ -223,6 +280,26 @@ func BuildTCPRouterConfiguration(ctx context.Context, configuration *dynamic.TCP
...
+			if len(configuration.Services) > 1 {
+				delete(configuration.Routers, routerName)
+				loggerRouter.
+					Error("Could not define the service name for the router: too many services")
+				continue
+			}
+
+			for serviceName := range configuration.Services {
+				router.Service = serviceName
+			}
```
mpl: I know it is not needed at the moment, because we already check the number of services above, but just for clarity, and to make sure no regression ever happens later:

```suggestion
				break
			}
```

pkg/provider/consulcatalog/config.go
```diff
@@ -42,6 +43,22 @@ func (p *Provider) buildConfiguration(ctx context.Context, items []itemData) *dy
...
+				logger.Error(err)
+				continue
+			}
+
+			provider.BuildUDPRouterConfiguration(ctxSvc, confFromLabel.UDP)
+
+			tcpOrUDP = true
+		}
+
+		if tcpOrUDP {
```
mpl: afaics, it can be contracted in just one "if" block:

```
if tcpOrUDP && if len(confFromLabel.HTTP.Routers) == 0 &&
```

pkg/provider/consulcatalog/config.go
```diff
@@ -42,6 +43,22 @@ func (p *Provider) buildConfiguration(ctx context.Context, items []itemData) *dy
...
 			provider.BuildTCPRouterConfiguration(ctxSvc, confFromLabel.TCP)
 
+			tcpOrUDP = true
+		}
+
+		if len(confFromLabel.UDP.Routers) > 0 || len(confFromLabel.UDP.Services) > 0 {
+			err := p.buildUDPServiceConfiguration(ctxSvc, item, confFromLabel.UDP)
+			if err != nil {
+				logger.Error(err)
+				continue
```
mpl: Are we sure that we do not want to try building the other parts of the config (HTTP, if it exists) if the UDP part fails?

(and same question for the TCP case L41 of course).

pkg/provider/marathon/config.go
```diff
@@ -175,6 +190,47 @@ func (p *Provider) buildTCPServiceConfiguration(ctx context.Context, app maratho
...
+		var servers []dynamic.UDPServer
+
+		defaultServer := dynamic.UDPServer{}
+
+		if len(service.LoadBalancer.Servers) > 0 {
+			defaultServer = service.LoadBalancer.Servers[0]
+		}
+
+		for _, task := range app.Tasks {
+			if p.taskFilter(ctx, *task, app) {
```
mpl: this test can be inverted to continue early.

pkg/provider/rancher/config.go
```diff
@@ -91,6 +106,28 @@ func (p *Provider) buildTCPServiceConfiguration(ctx context.Context, service ran
...
+		configuration.Services = make(map[string]*dynamic.UDPService)
+		lb := &dynamic.UDPServersLoadBalancer{}
+
+		configuration.Services[serviceName] = &dynamic.UDPService{
+			LoadBalancer: lb,
+		}
+	}
+
+	for _, confService := range configuration.Services {
+		err := p.addServerUDP(ctx, service, confService.LoadBalancer)
```
mpl: this can be contracted to the 

```
if err := blabla; err != nil {
```

form

PR #6338, opened 2020-02-18
Add config directory loader

ddtmachado:
### What does this PR do?

Adds a DirLoader strategy to the existing config loaders that can take precedence on the FileLoader to allow a user to use a directory containing multiple files with partial config contents.

### Motivation

This can be really useful while tweaking and testing a Traefik config, and also for those that have to quickly setup very different environments for QA or Tests while modifying sections of a base config.

As an example, one could have a base config named `base.yaml` like:

```yaml
global:
  checkNewVersion: true
  sendAnonymousUsage: false

entryPoints:
  web:
    address: ":80"
  websecure:
    address: ":443"
```

Then a separate file, named `k8s.yaml` to setup a kubernetes provider:

```yaml
providers:
  kubernetescrd: {}
```

And one more, named `swarm.yaml` to add a docker provider in swarm mode:

```yaml
providers:
  docker:
    swarmMode: true
    exposedByDefault: false
    endpoint: http://docker-api:4444
```

At this point if you want to disable the swarm provider, leaving only k8s configured, you can move `swarm.yaml` outside the config dir or even rename it to an unsupported config extension like `swarm.txt` making the loader ignore it.

Usage, when all the configs are placed in `/tmp/traefik/config.d/`:

```bash
$ traefik --configFile /tmp/traefik/config.d/ 
```

### More

- [X] Added/updated tests


ldez:
Hello,

This will a have a lot of important side effects and it will be more complicated to understand for users, so for now we will declined.

Thanks anyway.

PR #6340, opened 2020-02-19
Modular Whitelists: Configure Whitelists To Append Peers

jayheinlein:
### What does this PR do?

This PR allows whitelist middleware to reference a list of other whitelists to append. This allows for modular whitelists to simplify administration of multiple whitelists and multiple services.

### Motivation

I have a Kubernetes environment where I am managing multiple client whitelists and multiple services. Each service may have any combination of clients that are allowed to access it. In this case, we previously had to update the whitelist for each assigned service with its own custom combination of client whitelists whenever any client adds or removed whitelist IPs.

This PR allows us to centrally manage all client whitelists without fear of services having a whitelist that is out-of-date. Instead, all service whitelists reference their assigned clients' whitelists by name and Traefik updates their configuration when the client whitelists change.

### More

- [x] Added/updated tests


mpl:
Sorry, but unless there's an extremely demanded feature, or say a critical bug to fix, we want to keep the ipwhitelist middleware as simple as possible and therefore not touch it.
Consequently we're going to decline that PR, especially since your issue seems to be fixable outside of Traefik with a little more automation.
Please feel free to open an issue/proposal if you think we misunderstood the situation.

PR #6345, opened 2020-02-20
fix: consul-catalog use port from label instead of item port.

ldez:
### What does this PR do?

fix: consul-catalog use port from label instead of item port.

### Motivation

The label `traefik.http.services.xxx.loadbalancer.server.port` had no effect.

https://community.containo.us/t/override-consul-port-via-tags/4521

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/consulcatalog/config.go
+++ b/pkg/provider/consulcatalog/config.go
@@ -149,21 +149,25 @@ func (p *Provider) addServerTCP(ctx context.Context, item itemData, loadBalancer
 		return errors.New("load-balancer is not defined")
 	}
 
+	var port string
+	if len(loadBalancer.Servers) > 0 {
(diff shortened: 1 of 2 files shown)


pkg/provider/consulcatalog/config.go
```diff
@@ -149,21 +149,25 @@ func (p *Provider) addServerTCP(ctx context.Context, item itemData, loadBalancer
...
+		port = loadBalancer.Servers[0].Port
+	}
+
 	if len(loadBalancer.Servers) == 0 {
 		loadBalancer.Servers = []dynamic.TCPServer{{}}
 	}
 
-	var port string
 	if item.Port != "" {
 		port = item.Port
```
juliens: this will override the port from label?

PR #6352, opened 2020-02-21
Fix docs and code to match in haystack tracing.

evanlurvey:
### What does this PR do?

This PR addresses documentation errors and a default value change. The docs stated that defaults were one way but the code was defined otherwise. Sensible defaults were also chosen in the docs over "sample".

### Motivation

This PR was motivated by consuming the docs when integrating with haystack and getting an unexpected outcome.

### More

- [ ] Added/updated tests

### Additional Notes

N/A


pkg/tracing/haystack/haystack.go
```diff
@@ -26,7 +26,7 @@ type Config struct {
 
 // SetDefaults sets the default values.
 func (c *Config) SetDefaults() {
-	c.LocalAgentHost = "LocalAgentHost"
+	c.LocalAgentHost = "127.0.0.1"
```
mpl: Can you please explain as if we did not know anything about haystack (which is the case for most of us), what is the rationale behind that change?

evanlurvey: Sure, Haystack is a tracing tool designed by Expedia Group. To forward metrics to haystack you need a [haystack-agent](https://github.com/ExpediaDotCom/haystack-agent). The most common way to run this agent is as a docker side car to your application. When configured this way the agent will be listing on port 35000 by default and be associated with the same loopback address as your main service (traefik in this case). Also in the existing documentation for traefik it states that the default is 127.0.0.1 however I think it was accidentally set to LocalAgentHost.

ldez: ping @aantono

ldez:
@evanlurvey  thanks for your contribution.

You don't need to "merge v2.1", we have a bot for that :wink:

Issue #4864, opened 2019-05-09
traefik memory usage keeps growing


ravilr:
### Do you want to request a *feature* or report a *bug*?

Bug

### Did you try using a 1.7.x configuration for the version 2.0?

- [ ] Yes
- [* ] No

### What did you do?

we are notificing RSS memory of traefik process running within Kubernetes pod container keeps growing, eventually reaching the configured memory limits and getting OOMKilled.  

heap dump pprof profile shows most of the memory accumulated at ' crypto/tls.(*block).reserve /usr/local/go/src/crypto/tls/conn.go'
```
(pprof) top20
Showing nodes accounting for 982.10MB, 96.97% of 1012.75MB total
Dropped 227 nodes (cum <= 5.06MB)
Showing top 20 nodes out of 97
      flat  flat%   sum%        cum   cum%
  740.49MB 73.12% 73.12%   740.49MB 73.12%  crypto/tls.(*block).reserve /usr/local/go/src/crypto/tls/conn.go
   73.56MB  7.26% 80.38%    73.56MB  7.26%  crypto/tls.Server /usr/local/go/src/crypto/tls/tls.go
   47.01MB  4.64% 85.02%    47.01MB  4.64%  crypto/aes.(*aesCipherGCM).NewGCM /usr/local/go/src/crypto/aes/aes_gcm.go
   30.51MB  3.01% 88.04%    30.51MB  3.01%  crypto/aes.newCipher /usr/local/go/src/crypto/aes/cipher_asm.go
   15.50MB  1.53% 89.57%    15.50MB  1.53%  crypto/tls.(*halfConn).newBlock /usr/local/go/src/crypto/tls/conn.go
   13.50MB  1.33% 90.90%    13.50MB  1.33%  bytes.makeSlice /usr/local/go/src/bytes/buffer.go
      13MB  1.28% 92.18%       13MB  1.28%  syscall.anyToSockaddr /usr/local/go/src/syscall/syscall_linux.go
      12MB  1.19% 93.37%       12MB  1.19%  net.newFD /usr/local/go/src/net/fd_unix.go
      11MB  1.09% 94.45%   126.60MB 12.50%  crypto/tls.(*Conn).readHandshake /usr/local/go/src/crypto/tls/conn.go
       8MB  0.79% 95.24%        8MB  0.79%  crypto/tls.ekmFromMasterSecret /usr/local/go/src/crypto/tls/prf.go
[...]

PR #6357, opened 2020-02-23
Clear closed hijacked h2c connections

omerkay:
### What does this PR do?

- Fix memory leak caused by addition of hijacked h2c connections to managed list, but never evicting them.
- Fix redundant wait period during shutdown when h2c connections have been established.
- Remove redundant error logs during shutdown when h2c connections have been established.

### Motivation

Memory leak noticed In a scenario where grpc connections are established via Traefik repeatedly.
By using pprof heap analysis, it is evident that `hijackConnectionTracker` memory consumption keeps increasing as long as Traefik stays up.
See image for example:
Increased memory consumption

Also, since h2c hijacked connections are closed properly by Traefik after the connection is served, errors are logged when `hijackConnectionTracker` attempts to close these connections again during shutdown (after waiting for the full grace period):
```
Error while closing Hijacked conn: close tcp 127.0.0.1:8080->127.0.0.1:61623: use of closed network connection
Error while closing Hijacked conn: close tcp 127.0.0.1:8080->127.0.0.1:61629: use of closed network connection
Error while closing Hijacked conn: close tcp 127.0.0.1:8080->127.0.0.1:61618: use of closed network connection
....
```

### Additional Notes

similar issue with websockets (fixed): https://github.com/containous/traefik/issues/4864

### DIFF
--- a/h2c/h2c.go
+++ b/h2c/h2c.go
@@ -57,20 +57,20 @@ func (s Server) Serve(l net.Listener) error {
 			if http2VerboseLogs {
 				log.Debugf("Attempting h2c with prior knowledge.")
 			}
-			conn, err := initH2CWithPriorKnowledge(w)
+			conn, err := initH2CWithPriorKnowledge(w, s.closeHijackedConnQuietly)
 			if err != nil {
 				if http2VerboseLogs {
 					log.Debugf("Error h2c with prior knowledge: %v", err)
 				}
 				return
 			}
-			defer conn.Close()
+			defer s.closeHijackedConnQuietly(conn)
 			h2cSrv := &http2.Server{}
 			h2cSrv.ServeConn(conn, &http2.ServeConnOpts{Handler: originalHandler})
 			return
 		}
-		if conn, err := h2cUpgrade(w, r); err == nil {
-			defer conn.Close()
+		if conn, err := h2cUpgrade(w, r, s.closeHijackedConnQuietly); err == nil {
+			defer s.closeHijackedConnQuietly(conn)
 			h2cSrv := &http2.Server{}
 			h2cSrv.ServeConn(conn, &http2.ServeConnOpts{Handler: originalHandler})
 			return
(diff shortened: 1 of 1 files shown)


omerkay:
thanks @juliens.

as a side note, I suspect the following memory consumption will also be resolved by this fix (taken from :
Screen Shot 2020-02-25 at 15 34 14

since these are the actual `net.Conn` allocations.

omerkay:
@ldez @juliens 
Thank you for the quick review.
Do you have an estimate on when this will be available in an official version?
I noticed that I just missed the last release, but it is an important fix for grpc use cases.

h2c/h2c.go
```diff
@@ -57,20 +57,20 @@ func (s Server) Serve(l net.Listener) error {
...
-			defer conn.Close()
+			defer s.closeHijackedConnQuietly(conn)
 			h2cSrv := &http2.Server{}
 			h2cSrv.ServeConn(conn, &http2.ServeConnOpts{Handler: originalHandler})
 			return
 		}
-		if conn, err := h2cUpgrade(w, r); err == nil {
-			defer conn.Close()
+		if conn, err := h2cUpgrade(w, r, s.closeHijackedConnQuietly); err == nil {
+			defer s.closeHijackedConnQuietly(conn)
```
mpl: Couldn't we be smarter within closeHijackedConnQuietly, so that here we could systematically call s.closeHijackedConnQuietly regardless of the value of err, and it would be a NOOP when there was an error that occurred before the connection was hijacked?
Then we would not need to call closeHijackedConnQuietly L183, and there would be no need to pass it as an argument.

omerkay: The main reason for passing it as argument is for handling connection closing in cases where the connection already hijacked, but is not returned (both `initH2CWithPriorKnowledge` and `h2cUpgrade` have such cases).

Another option would be to return the hijacked connection anyway even if an error occurred, but that would mean passing responsibility of cleanup to the caller.

PR #6384, opened 2020-02-26
Remove unnecessary brackets from ipv6 literals

42wim:
### What does this PR do?

Removes brackets from IPv6 addresses, eg set by broken proxies.

### Motivation

ipwhitelist middleware failed to work in our setup (netscaler before traefik)
Netscaler (incorrectly) injects for example `[::ffff:a27:f81d]`  in the `X-Forwarded-For` header. 

This breaks ipwhitelist. 

Removing the brackets fixes the issue. (brackets should only be used in URLs or when specifying a port)

### More

- [x] tests: added some ipv6 addresses with brackets between the existing tests.

### DIFF
--- a/pkg/ip/checker.go
+++ b/pkg/ip/checker.go
@@ -90,6 +90,9 @@ func (ip *Checker) ContainsIP(addr net.IP) bool {
 }
 
 func parseIP(addr string) (net.IP, error) {
+	addr = strings.ReplaceAll(addr, "[", "")
+	addr = strings.ReplaceAll(addr, "]", "")
+
 	userIP := net.ParseIP(addr)
 	if userIP == nil {
 		return nil, fmt.Errorf("can't parse IP from address %s", addr)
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

The "brackets syntax" is not valid as IPv6, and `net.ParseIP` know how to parse IP6.

We will not fix errors made by Netscaler.

Thanks anyway.

ldez:
After checking the RFC, the "brackets syntaxt" seems valid.

https://tools.ietf.org/html/rfc3986#section-3.2.2

I need to think about that :thinking:

42wim:
It's not a good case for my PR, but the brackets are invalid in the `X-Forwarded-For`.
They are only valid in a host of an URI (as in RFC 3986) (not as a hostname an sich) 
:(

ldez:
We reflect on the problem, it's an invalid header value, the problem seems specific to Netscaler.

We cannot accept a fix for an invalid behavior from another product, if we accept that, we will have to accept every specific fix.

So the best I can offer you is to open an issue instead of a PR, and maybe contact Citrix to fix the problem at the root.

Thanks anyway.

bv-kul:
X-Forwarded-For is a non-standard HTTP header, so the syntax may vary across implementations. The Forwarded header (RFC 7239) on the other hand does enclose IPv6 addresses in square brackets.

In my opinion there is nothing wrong with way Citrix has implemented this header.

PR #6413, opened 2020-03-02
Added support for replacement containing escaped characters

### LINKED ISSUE #6355: Unexpected URL-encoding of path in ReplacePathRegex middleware

Bug

### What did you do?

I tried to URL-encode a `/` in a path using the ReplacePathRegex middleware. Traefik took the `%` sign of `%2F` and encoded it again, resulting in `%252F` instead.
Is this related to this middleware handling the `Path` rather than the `RawPath` property, similar to the problem that was fixed in https://github.com/containous/traefik/pull/2382/files?

e.g.
```
/entities/projects/some-namespace/some-project 
--> 
/entities/projects/some-namespace%2Fsome-project
```

### What did you see instead?

```
/entities/projects/some-namespace/some-project 
--> 
/entities/projects/some-namespace%252Fsome-project
```

```
Version:      2.1.4
Codename:     cantal
Go version:   go1.13.7
Built:        2020-02-06T17:10:06Z
OS/Arch:      linux/amd64
```

```toml
[http.middlewares.urlEncodeForwardSlash.ReplacePathRegex]
  regex = "/entities/([^/]*)/([^/]*)/(.*)"
  replacement = "/entities/${1}/${2}%2F${3}"
```

```
time="2020-02-22T14:58:15Z" level=debug msg="vulcand/oxy/roundrobin/rr: completed ServeHttp on request" Request="{\"Method\":\"GET\",\"URL\":{\"Scheme\":\"\",\"Opaque\":\"\",\"User\":null,\"Host\":\"\",\"Path\":\"/entities/projects/andreas%2Fmy-project\",\"RawPath\":\"\",\"ForceQuery\":false,\"RawQuery\":\"\",\"Fragment\":\"\"},\"Proto\":\"HTTP/1.1\",\"ProtoMajor\":1,\"ProtoMinor\":1,\"Header\":{\"Accept\":[\"application/json\"],\"Accept-Encoding\":[\"gzip, deflate\"],\"Cache-Control\":[\"no-cache\"],\"Cookie\":[\"session=9b14534af647ce28_5e4ef2c3.RaUgKduhfJFR-vq-sjl-4-8Wsyo; sails.sid=s%3ASKKV5IEbjSWTk_g_se8t3HjhqSwFhCxx.j0kFSx%2Fr8ogYMR5T8GyroHEzNbPGmO70mv29rbVj1Dc\"],\"User-Agent\":[\"PostmanRuntime/7.20.1\"],\"X-Forwarded-Host\":[\"andreas.dev.renku.ch\"],\"X-Forwarded-Port\":[\"80\"],\"X-Forwarded-Proto\":[\"http\"],\"X-Forwarded-Server\":[\"andreas-renku-gateway-7c96dc49c9-nfgrz\"],\"X-Original-Uri\":[\"/entities/projects/andreas/my-project\"],\"X-Real-Ip\":[\"10.42.3.61\"],\"X-Replaced-Path\":[\"/entities/projects/andreas/my-project\"],\"X-Request-Id\":[\"be0c6c922b80b103dd14f49871bd9717\"],\"X-Requested-With\":[\"XMLHttpRequest\"],\"X-Scheme\":[\"https\"]},\"ContentLength\":0,\"TransferEncoding\":null,\"Host\":\"andreas.dev.renku.ch\",\"Form\":null,\"PostForm\":null,\"MultipartForm\":null,\"Trailer\":null,\"RemoteAddr\":\"10.42.3.61:33898\",\"RequestURI\":\"/entities/projects/andreas%252Fmy-project\",\"TLS\":null}"
```

rtribotte:
### What does this PR do?

Added support for replacement containing escaped characters

### Motivation

Fixes #6355

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: Ludovic Fernandez <ldez@users.noreply.github.com>

### DIFF
--- a/pkg/middlewares/replacepath/replace_path.go
+++ b/pkg/middlewares/replacepath/replace_path.go
@@ -3,6 +3,7 @@ package replacepath
 import (
 	"context"
 	"net/http"
+	"net/url"
 
 	"github.com/containous/traefik/v2/pkg/config/dynamic"
 	"github.com/containous/traefik/v2/pkg/log"
@@ -40,8 +41,22 @@ func (r *replacePath) GetTracingInformation() (string, ext.SpanKindEnum) {
 }
 
 func (r *replacePath) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
-	req.Header.Add(ReplacedPathHeader, req.URL.Path)
-	req.URL.Path = r.path
+	if req.URL.RawPath == "" {
+		req.Header.Add(ReplacedPathHeader, req.URL.Path)
+	} else {
+		req.Header.Add(ReplacedPathHeader, req.URL.RawPath)
+	}
+
+	req.URL.RawPath = r.path
+
+	var err error
+	req.URL.Path, err = url.PathUnescape(req.URL.RawPath)
+	if err != nil {
+		log.FromContext(middlewares.GetLoggerCtx(context.Background(), r.name, typeName)).Error(err)
+		http.Error(rw, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
 	req.RequestURI = req.URL.RequestURI()
 
 	r.next.ServeHTTP(rw, req)
--- a/pkg/middlewares/replacepath/replace_path_test.go
+++ b/pkg/middlewares/replacepath/replace_path_test.go
@@ -3,43 +3,93 @@ package replacepath
 import (
 	"context"
 	"net/http"
+	"net/http/httptest"
 	"testing"
 
 	"github.com/containous/traefik/v2/pkg/config/dynamic"
-	"github.com/containous/traefik/v2/pkg/testhelpers"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func TestReplacePath(t *testing.T) {
-	var replacementConfig = dynamic.ReplacePath{
-		Path: "/replacement-path",
+	testCases := []struct {
+		desc            string
+		path            string
+		config          dynamic.ReplacePath
+		expectedPath    string
+		expectedRawPath string
+		expectedHeader  string
+	}{
+		{
+			desc: "simple path",
+			path: "/example",
+			config: dynamic.ReplacePath{
+				Path: "/replacement-path",
+			},
+			expectedPath:    "/replacement-path",
+			expectedRawPath: "",
+			expectedHeader:  "/example",
+		},
+		{
+			desc: "long path",
+			path: "/some/really/long/path",
+			config: dynamic.ReplacePath{
+				Path: "/replacement-path",
+			},
+			expectedPath:    "/replacement-path",
+			expectedRawPath: "",
+			expectedHeader:  "/some/really/long/path",
+		},
+		{
+			desc: "path with escaped value",
+			path: "/foo%2Fbar",
+			config: dynamic.ReplacePath{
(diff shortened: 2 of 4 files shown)


pkg/middlewares/replacepath/replace_path.go
```diff
@@ -40,8 +41,22 @@ func (r *replacePath) GetTracingInformation() (string, ext.SpanKindEnum) {
...
+	if req.URL.RawPath == "" {
+		req.Header.Add(ReplacedPathHeader, req.URL.Path)
+	} else {
+		req.Header.Add(ReplacedPathHeader, req.URL.RawPath)
+	}
+
+	req.URL.RawPath = r.path
+
+	var err error
+	req.URL.Path, err = url.PathUnescape(req.URL.RawPath)
```
dtomcej: This PR modifies `req.URL.Path` and `req.URL.RawPath` without returning them to the original values. Is that something we want?

ldez: I don't understand your question, could you elaborate?

dtomcej: Look at this test case: https://github.com/containous/traefik/pull/6413/files#diff-d3677828257e31aeca5933a90a133acbR54

In that case, the _original_ request `req.URL.Path` is `/path`, but now is `/foo/bar`, which is neither the requested path, nor the replaced path...

I think this is a behaviour change?

pkg/middlewares/replacepath/replace_path_test.go
```diff
@@ -3,43 +3,93 @@ package replacepath
...
+		{
+			desc: "replacement with escaped value",
+			path: "/path",
+			config: dynamic.ReplacePath{
+				Path: "/foo%2Fbar",
+			},
+			expectedPath:    "/foo/bar",
+			expectedRawPath: "/foo%2Fbar",
+			expectedHeader:  "/path",
+		},
```
mpl: maybe add a test case that is the combination of the above two, i.e.: "path with escaped value, and replacement with escaped value" ?

pkg/middlewares/replacepath/replace_path.go
mpl: What is your concern @dtomcej ? Before the PR, this middleware was mutating req.URL.Path but not req.URL.RawPath. After this PR, it's going to possibly mutate both of them. Are you worried that one of the next handlers in the chain was relying on req.URL.RawPath not being mutated?
imho, since we already meddle with the request and modify req.URL.Path, I think we already conceded that all bets are off for the next handlers. So going one step further and mutating req.URL.RawPath as well is not such a big deal.

dtomcej: >Before the PR, this middleware was mutating req.URL.Path but not req.URL.RawPath

Before this pr, the middleware was mutating `req.URL.Path` according to the user-supplied regex replacement, so it ended up what the user expected.

yes, I was thinking for the next step.

ldez: If something was using `req.URL.RawPath` to know the "ancestor path" it was an error.
And I don't think we do that.

pkg/middlewares/replacepath/replace_path_test.go
ldez: I don't think that is a special case, the 2 other cases test all the possibilities.

PR #6422, opened 2020-03-04
Add some missing doc.

ldez:
### What does this PR do?

Add some missing doc.

### Motivation

Have a better documentation.

### More

- ~~[ ] Added/updated tests~~


docs/content/reference/static-configuration/cli-ref.md
```diff
@@ -346,7 +346,7 @@ TLS key
 Use the ip address from the bound port, rather than from the inner network. (Default: ```false```)
 
 `--providers.docker.watch`:  
-Watch provider. (Default: ```true```)
+Watch Docker Swarm events. (Default: ```true```)
```
jbdoumenjou: Why not just "Docker events" ?

ldez: Because it's only related to Swarm

PR #6428, opened 2020-03-05
Improve kubernetes external name service support

### LINKED ISSUE #6390: IngressRouteTCP object not created when the route is a Service of type ExternalName

Bug

### What did you do?

I've created an object of `kind: IngressRouteTCP` and configured in such way:

```yaml
kind: IngressRouteTCP
metadata:
  name: external-tcp-service
spec:
  entryPoints:
    - tcp
  routes:
    - match: HostSNI(`my-external-service.domain.com`)
      services:
        - name: my-external-service
          port: 6379
  tls: {}
---
apiVersion: v1
kind: Service
metadata:
  name: external-tcp-service
spec:
  externalName: internaltcpsrv.domain.local
  type: ExternalName

```

I would expect that the object `external-tcp-service` would be created.

### What did you see instead?

```sh
level=error msg="Cannot create service: service port not found" 
```

```
2.1.3
```

Kubernetes 1.15.9

I believe that the reason is of a bug here https://github.com/containous/traefik/blob/master/pkg/provider/kubernetes/crd/kubernetes_tcp.go#L165

There is an immediate lookup of the port in the `Service` and only after there is a check of the type to verify if the `Service` is `type: ExternalName`.

I believe that the logic must be rearranged a bit. I can wire a PR if is needed.

rtribotte:
### What does this PR do?

Improve kubernetes external name service support by allowing three ways to define port for backends

### Motivation
fixes #6390 

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: jbdoumenjou <jb.doumenjou@gmail.com>

### DIFF
--- a/docs/content/routing/providers/kubernetes-crd.md
+++ b/docs/content/routing/providers/kubernetes-crd.md
@@ -472,7 +472,7 @@ Register the `IngressRoute` [kind](../../reference/dynamic-configuration/kuberne
 !!! important "Using Kubernetes ExternalName Service"
 
     Traefik backends creation needs a port to be set, however Kubernetes [ExternalName Service](https://kubernetes.io/fr/docs/concepts/services-networking/service/#externalname) could be defined without any port.
-    Accordingly, Traefik supports defining a port in three ways:
+    Accordingly, Traefik supports defining a port in two ways:
     
     - only on `IngressRoute` service
     - on both sides, you'll be warned if the ports don't match, and the `IngressRoute` service port is used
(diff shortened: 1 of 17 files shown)


docs/content/routing/providers/kubernetes-crd.md
```diff
@@ -467,6 +467,112 @@ Register the `IngressRoute` [kind](../../reference/dynamic-configuration/kuberne
     - Setting the kubernetes service port to use port 443 (https)
 
     If you do not configure the above, Traefik will assume an http connection.
+    
+
+!!! important "Using Kubernetes ExternalName Service"
+
+    Traefik backends creation needs a port to be set, however Kubernetes [ExternalName Service](https://kubernetes.io/fr/docs/concepts/services-networking/service/#externalname) could be defined without any port.
+    Accordingly, Traefik supports defining a port in three ways:
```
dtomcej: But you have only two in the list?

docs/content/routing/providers/kubernetes-crd.md
```diff
@@ -928,6 +1034,111 @@ Register the `IngressRouteTCP` [kind](../../reference/dynamic-configuration/kube
       tls.key: LS0tLS1CRUdJTiBQUklWQVRFIEtFWS0tLS0tCi0tLS0tRU5EIFBSSVZBVEUgS0VZLS0tLS0=
     ```
 
+!!! important "Using Kubernetes ExternalName Service"
+
+    Traefik backends creation needs a port to be set, however Kubernetes [ExternalName Service](https://kubernetes.io/fr/docs/concepts/services-networking/service/#externalname) could be defined without any port.
+    Accordingly, Traefik supports defining a port in three ways:
```
dtomcej: same as above

PR #6443, opened 2020-03-06
Add Anonymous Authentication middleware

Amygos:
### What does this PR do?
 The Anonymous Authentication middleware (anonAuth) is inspired by the Apache Module [`mod_authn_anon`](https://httpd.apache.org/docs/2.4/mod/mod_authn_anon.html).

### Motivation
It can be used to force the authentication, filter the allowed usernames (by a list of regexps) or pass the provided username to the downstream middlewares and put it in the access logs.

### More

- [x] Added/updated tests

### Additional Notes
The current implementation supports only the HTTP Basic authentication.


ldez:
Hello,

for now this feature seems a edge-case.

Could you open a proposal, to allow us to see if the feature is popular or not.

So for now, we will declined.

Thanks anyway.

PR #6460, opened 2020-03-10
Fix example values for swarmModeRefreshSeconds

skjnldsv:
## What does this PR do?

Fix typo in documentation
The default doesn't have the `s` unit, the example should probably not have it either :)
https://github.com/containous/traefik/blob/09c07f45ee999fad6dc97463b4b825fb08c7c34b/pkg/provider/docker/docker.go#L68

## Motivation

Was reading the documentation, find it confusing :grin: 

## More

- [ ] Added/updated tests

## Additional Notes

You're all incredible people and this is an amazing software! :tada:

### DIFF
--- a/docs/content/providers/docker.md
+++ b/docs/content/providers/docker.md
@@ -473,19 +473,19 @@ _Optional, Default=15_
 
 ```toml tab="File (TOML)"
 [providers.docker]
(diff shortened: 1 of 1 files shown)


skjnldsv:
> FYI `30`, `"30s"` and "30" are allowed.

Good to know :)
Thanks!

PR #6466, opened 2020-03-11
Update traefik install documentation

mmatur:
### What does this PR do?

Since the helm chart is production ready, we need to update the documentation

### Motivation

Be up-to-date

### More

- ~[ ] Added/updated tests~


docs/content/getting-started/install-traefik.md
```diff
@@ -29,46 +29,50 @@ For more details, go to the [Docker provider documentation](../providers/docker.
 
 !!! warning "Experimental Helm Chart"
```
dtomcej: Do we want to change this to "Helm Chart"?

PR #6471, opened 2020-03-11
doc: Use neutral domains.

### LINKED ISSUE #6470: Unregistered domain used in documentation

Bug

### What did you do?

Read the documentation

Example domains or domains registered by the developers

### What did you see instead?

The current domain `my-awesome-app.org` is available for registration.
If a bad actor decides to register the domain and host malicious/offensive content it will reflect badly on this project.
Example domains should either be example.com / example.org or registered by the developers.

N/A

N/A

N/A

ldez:
### What does this PR do?

Use only the reserved example second level domain names in the documentation.

https://tools.ietf.org/html/rfc2606#section-3

### Motivation

Have a better documentation.

Fixes  #6470

### More

- ~~[ ] Added/updated tests~~

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -79,7 +79,7 @@ Please check the [configuration examples below](#configuration-examples) for mor
         address = ":443"
     
     [certificatesResolvers.myresolver.acme]
-      email = "your-email@your-domain.org"
+      email = "your-email@example.com"
       storage = "acme.json"
       [certificatesResolvers.myresolver.acme.httpChallenge]
         # used during the challenge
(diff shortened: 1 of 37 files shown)


jsoref:
Fwiw, this change corrupted a number of the examples.

The spirit of the change is absolutely desirable, but the examples had more problems (including `scheese.com` in places) which meant that a naive search and replace didn't achieve the desired results (using the RFC `example.com` domain).

But, ignoring the fact that a search and replace of bad content results in bad content, a search and replace of cert content that doesn't fix up the encoded form is problematic.

PR #6499, opened 2020-03-16
Fix error watching lock key, when key doesn't exist

kepka42:
### What does this PR do?

Before watching create empty lock key in storage if key doesn't exists.

### Motivation

When using etcd and acme together, an error occurs. The /traefik/acme/lock key is not set. Corrected the fact that you need to manually add a null value in the key.

### More

- [ ] Added/updated tests

### DIFF
--- a/cluster/datastore.go
+++ b/cluster/datastore.go
@@ -76,6 +76,19 @@ func NewDataStore(ctx context.Context, kvSource staert.KvSource, object Object, 
 
 func (d *Datastore) watchChanges() error {
 	stopCh := make(chan struct{})
+
+	existsKey, err := d.kv.Exists(d.lockKey, nil)
+	if err != nil {
+		return fmt.Errorf("error check exists key %s: %v", d.lockKey, err)
+	}
+
+	if !existsKey {
+		err := d.kv.Put(d.lockKey, []byte{}, nil)
+		if err != nil {
+			return fmt.Errorf("error store key %s: %v", d.lockKey, err)
+		}
+	}
+
 	kvCh, err := d.kv.Watch(d.lockKey, stopCh, nil)
 	if err != nil {
 		return fmt.Errorf("error while watching key %s: %v", d.lockKey, err)


kepka42:
Apparently, integration tests are broken? They didn 't work before my commit

till:
Can someone have a look at this? This seems to address a ton of issues for people on 1.7 which use a k/v store?

kepka42:
@ldez can you help me?

SantoDE:
Hello @kepka42,

thanks for your interest in the project.

While we would like to review that PR, we're wondering if you have some configurations or similiar, so we can easily reproduce that your solution is working?

You're fix looks really "simple" so we want to make sure that it really works out.

Thanks.

kepka42:
Hello @SantoDE,

For reproduce my solution you need to start Etcd server. For Traefik you can use this traefik.toml:

```
logLevel = "DEBUG"

[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    entryPoint = "https"
  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]

[api]
  entryPoint = "traefik"
  dashboard = true

[log]
  filePath = "/var/log/traefik.log"
  
[etcd]
  watch = true
  prefix = "/traefik"
  useAPIV3 = true

[acme]
#acmeLogging = true
email = "test@test.com"
storage = "/traefik/acme"
caServer = "https://acme-staging-v02.api.letsencrypt.org/directory"
entryPoint = "https"
  [acme.httpChallenge]
  entryPoint = "http"

  [acme.dnsChallenge]
  provider = "cloudflare"

[[acme.domains]]
  main = "test.localhost"
[[acme.domains]]
  main = "test.localhost"
```

ddtmachado:
Hello @kepka42 , we're [not supporting the v1.x](https://doc.traefik.io/traefik/deprecation/releases/) branch anymore thus we cannot review and merge this PR as there won't be any future releases for it. 

If you confirm the problem still exists on v2 please don't hesitate to open a new issue / PR, also if you still have any blockers for migrating to v2 let us know so we can consider that use case and improve on what's missing.

PR #6504, opened 2020-03-17
fix: Ingress TLS support

### LINKED ISSUE #6501: [2.2.0-rc2] kubernetes/ingress: bare host failed match on HTTP
Bug

### What did you do?

Configure HTTP and HTTPS endpoints with middleware chain to redirect http to https.

Redirect HTTP to HTTPS in all situations, including no or bare "/" path.

### What did you see instead?

Without any path "" or maybe "/", the route "GET http://sub.domain" would fail to redirect to "https://sub.domain". It would just stop with a 404.

```
Version:      2.2.0-rc2
Codename:     chevrotin
Go version:   go1.14
Built:        2020-03-11T17:42:16Z
OS/Arch:      linux/amd64
```

Static configuration through environment for endpoints and Kubernets Ingress and Kubernetes CRD setup. Kubernetes configuration for dynamic configuration for routes.

Kubernetes Ingress and KubernetesCRD, configured outside of the cluster.

Nothing applicable in DEBUG level log.

---

I'm using the newly (re-)introduced Kubernetes Ingress (not IngressRoute) for my main configuration for various reasons. I'm also using a Middleware chain that is redirecting http to https and non-www to www sub-domains (plus header addition, etc). Appears to work well.

When testing certain bare HTTP sub-domains, I found traefik would return a 404 *before* it even encountered the middleware. Testing was complicated due to HSTS header. But for example I found: `GET http://qa.mydomain.com` -> 404 (middleware not encountered) while `http://qa.mydomain/anything` -> 200 using middleware. I was able to use the jaeger integration to determine if the middle was was encountered at all.

I manually resolved this by creating Kubernetes IngressRoutes that bound directly on the HTTP endpoint only, used the same middleware, and only matched on the host, no Path or PathPrefix. This corrected the behavior to the desired behavior.

I suspect this has to do with how a Kubernetes Ingress object is internally represented + how HTTP (non-TLS) traffic paths are handled. Traefik creates to rules for Kubernetes Ingress objects: Host and PathPrefix. I suspect that internally the Request.URL.Path is either "/" or "" (empty) for this request either incoming or after processing. Then the created PathPrefix(`/`) rule doesn't match. This only appears to happen on HTTP, not HTTPS.

ldez:
### What does this PR do?

Now, the TLS section in a Ingress don't control the activation of the TLS.

To enable TLS:
- use the TLS configuration on the entry point: https://docs.traefik.io/v2.2/routing/entrypoints/#tls
- use annotation `traefik.ingress.kubernetes.io/router.tls: "true"`: https://docs.traefik.io/v2.2/routing/providers/kubernetes-ingress/#on-ingress

### Motivation

Fixes #6501,#6506

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: Julien Salleyron <julien@containo.us>

### DIFF
--- a/docs/content/migration/v2.md
+++ b/docs/content/migration/v2.md
@@ -172,3 +172,133 @@ rules:
 ```
 
 After having both resources applied, Traefik will work properly.
+
+### Kubernetes Ingress
+
+The TLS section in the Ingress is now not enough to have HTTPS.
+
+#### Expose an Ingress on 80 and 443
+
+Define the default TLS configuration on the HTTPS entry point. 
+
+```yaml tab="Ingress"
+kind: Ingress
+apiVersion: networking.k8s.io/v1beta1
+metadata:
+  name: example
+
+spec:
+  tls:
+  - secretName: myTlsSecret
+
+  rules:
+  - host: example.com
+    http:
+      paths:
+      - path: "/foo"
+        backend:
+          serviceName: example-com
+          servicePort: 80
+```
+
+Entry points definition and enable Ingress provider:
+
+```yaml tab="File (YAML)"
+# Static configuration
+
+entryPoints:
+  web:
+    address: :80
+  websecure:
+    address: :443
+    http:
+      tls: {}
+
+providers:
+  kubernetesIngress: {}
+```
+
+```toml tab="File (TOML)"
+# Static configuration
+
+[entryPoints.web]
+  address = ":80"
+
+[entryPoints.websecure]
+  address = ":443"
+  [entryPoints.websecure.http]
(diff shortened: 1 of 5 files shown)


docs/content/migration/v2.md
```diff
@@ -172,3 +172,133 @@ rules:
 ```
 
 After having both resources applied, Traefik will work properly.
+
+### Kubernetes Ingress
+
+The TLS section in the Ingress is now not enough to have HTTPS.
```
mpl:
```suggestion
To enable HTTPS, it is not sufficient anymore to only rely on a TLS section in the Ingress.
```

PR #6518, opened 2020-03-18
Add decryption option for keyfiles

### LINKED ISSUE #1262: Feature request: support encrypted private keys for ssl
It would be great if traefik's ssl configuration supported a private key using a passphrase. Something like:

```toml
...
[[entrypoints.https.tls.certificates]]
certFile = "/etc/traefik/ssl/cert.pem"
keyFile = "/etc/traefik/ssl/key.pem"
keyFilePassphrase = "foobar"
...
```

This would be especially useful when combined with #7. The keyfile password wouldn't even need to be written in the config file.

v1.1.2

```toml
[entryPoints]
  [entryPoints.http]
  address = ":80"
    [entryPoints.http.redirect]
    entryPoint = "https"
  [entryPoints.https]
  address = ":443"
    [entryPoints.https.tls]
      certFile = "/etc/traefik/ssl/cert.pem"
      keyFile = "/etc/traefik/ssl/key.pem"
```

in my case, I would like to have a passphrase for `/etc/traefik/ssl/key.pem`

### What did you do?
I tried to use the above configuration in Traefik, where the keyFile was encrypted with a passphrase.

A configuration option for the keyFile's passphrase

### What did you see instead?
I saw the following log output before Traefik exited with an error:
```
time="2017-03-09T10:24:32+11:00" level=error msg="Error creating TLS config tls: failed to parse private key" 
time="2017-03-09T10:24:32+11:00" level=fatal msg="Error preparing server: tls: failed to parse private key"
```

dtomcej:
### What does this PR do?

This PR provides a passphrase with which to decrypt encrypted private keys

### Motivation

Many users have expressed the need to utilize encrypted keys. Manually decrypting and storing these copies is not always acceptable.

Fixes #1262 

### More

Updates:
- File provider
- K8s CRD provider
- KV provider (nothing needed after file provider handled)

Completed:
- [x] Added/updated tests

### DIFF
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -42,6 +42,40 @@ tls:
     It is the only available method to configure the certificates (as well as the options and the stores).
     However, in [Kubernetes](../providers/kubernetes-crd.md), the certificates can and must be provided by [secrets](https://kubernetes.io/docs/concepts/configuration/secret/). 
 
+### Encrypted private keys
+
+To use TLS certificates with encrypted private keys, the passphrase must be provided with which to decrypt the key.
+
+```toml tab="File (TOML)"
+# Dynamic configuration
+
+[[tls.certificates]]
+  certFile = "/path/to/domain-encrypted.cert"
+  keyFile = "/path/to/domain-encrypted.key"
+  passphrase = "mysuperpassphrase"
+```
+
+```yaml tab="File (YAML)"
+# Dynamic configuration
+
+tls:
+  certificates:
+    - certFile: /path/to/domain-encrypted.cert
+      keyFile: /path/to/domain-encrypted.key
+      passphrase: mysuperpassphrase
+```
+
+!!! important "Incorrect passphrase"
+
+    An incorrect/missing passphrase when using an encrypted private key will result in the certificate being unusable.
+    There will also be corresponding decryption failures in the logs.
+    These errors may generate a lot of log volume, so exercise care to ensure the passphrase is correct.
+
+!!! important "Restriction"
+
+    In the above example, we've used the [file provider](../providers/file.md) to handle these definitions.
+    However, in [Kubernetes](../providers/kubernetes-crd.md), the certificates can and must be provided by [secrets](https://kubernetes.io/docs/concepts/configuration/secret/), and the passphrase must also be provided in a secret.
+
 ## Certificates Stores
 
 In Traefik, certificates are grouped together in certificates stores, which are defined as such:
--- a/pkg/tls/certificate.go
+++ b/pkg/tls/certificate.go
@@ -3,6 +3,7 @@ package tls
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -48,8 +49,9 @@ var (
 // Certificate holds a SSL cert/key pair
 // Certs and Key could be either a file path, or the file content itself
 type Certificate struct {
-	CertFile FileOrContent `json:"certFile,omitempty" toml:"certFile,omitempty" yaml:"certFile,omitempty"`
-	KeyFile  FileOrContent `json:"keyFile,omitempty" toml:"keyFile,omitempty" yaml:"keyFile,omitempty"`
+	CertFile   FileOrContent `json:"certFile,omitempty" toml:"certFile,omitempty" yaml:"certFile,omitempty"`
+	KeyFile    FileOrContent `json:"keyFile,omitempty" toml:"keyFile,omitempty" yaml:"keyFile,omitempty"`
+	Passphrase string        `json:"passphrase,omitempty" toml:"passphrase,omitempty" yaml:"passphrase,omitempty"`
 }
 
 // Certificates defines traefik certificates type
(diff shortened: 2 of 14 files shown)


pkg/tls/certificate.go
```diff
@@ -237,3 +245,30 @@ func (c *Certificates) Set(value string) error {
...
+		if err != nil {
+			return nil, fmt.Errorf("unable to decrypt TLS certificate key : %w", err)
+		}
+
+		keyDecoded = pem.EncodeToMemory(&pem.Block{
+			Type:  keyPemBlock.Type,
+			Bytes: keyDecoded,
+		})
+	} else {
+		keyDecoded = pem.EncodeToMemory(keyPemBlock)
```
rtribotte, line 270: Hi @dtomcej,
Returning directly the key content must work as well (at least it's not breaking unit tests), that would make the call to EncodeToMemory not necessary, what do you think ?

docs/content/https/tls.md
```diff
@@ -42,6 +42,40 @@ tls:
...
+      passphrase: mysuperpassphrase
+```
+
+!!! important "Incorrect passphrase"
+
+    An incorrect/missing passphrase when using an encrypted private key will result in the certificate being unusable.
+    There will also be corresponding decryption failures in the logs.
+    These errors may generate a lot of log volume, so exercise care to ensure the passphrase is correct.
+
+!!! important "Restriction"
```
rtribotte, line 74: Here, regarding the discussions in the issue, and aside kubernetes, i think the documentation can promote the usage of environnement variable or flag. WDYT ?

pkg/tls/certificate.go
dtomcej: I'm not sure what you mean.

If the key block is not encoded, it is returned after being encoded from a `pem.block` to `[]byte`.

`keyPemBlock` is not a byte array.

rtribotte: I was thinking of something like that :
```go
if len(passphrase) == 0 {
    // No passphrase provided, return provided key.
    return keyContent, nil
}

keyPemBlock, _ := pem.Decode(keyContent)
if x509.IsEncryptedPEMBlock(keyPemBlock) {
    keyDecoded, err := x509.DecryptPEMBlock(keyPemBlock, []byte(passphrase))
    if err != nil {
        return nil, fmt.Errorf("unable to decrypt TLS certificate key : %w", err)
    }

    return pem.EncodeToMemory(&pem.Block{
        Type:  keyPemBlock.Type,
        Bytes: keyDecoded,
    }), nil
} 

return keyContent, nil
```

WDYT ?

docs/content/https/tls.md
```diff
@@ -42,6 +42,40 @@ tls:
...
+
+    In the above example, we've used the [file provider](../providers/file.md) to handle these definitions.
+    However, in [Kubernetes](../providers/kubernetes-crd.md), the certificates can and must be provided by [secrets](https://kubernetes.io/docs/concepts/configuration/secret/), and the passphrase must also be provided in a secret.
```
rtribotte, line 77:
```suggestion
    However, providing the passphrase this way, imply loosing the benefit of an encrypted private certificate, and could be considered as convenience.
    Besides, in [Kubernetes](../providers/kubernetes-crd.md), the certificates can and must be provided by [secrets](https://kubernetes.io/docs/concepts/configuration/secret/), and the passphrase must also be provided in a secret.
```

WDYT ?

rtribotte:
Hi @dtomcej,

We believe that this decryption feature is an edge case and also, providing this convenience, to use a certificate protected by a passphrase, do not improve the security.

In the file provider, giving the passphrase in plain text do not provide more security than a certificate not protected by a passphrase.
For the IngressRoute and IngressRouteTCP, because the Kubernetes secrets are trusted to store the passphrase, using a certificate protected by a passphrase, both stored in secrets, seems superfluous.

After some discussions, unfortunately we want to decline this PR.

Please, let us know if we are missing something.
Thank you anyway and we hope you are well!

Steve973:
Sometimes, people have to use corporate PKI and we do not have the luxury of deciding whether or not to use a passphrase for keys that we use.  Allowing users to supply a `keyFilePassphrase` or a `keyFilePassphraseFile` allows for *much* more flexibility in the case where a developer or a devops engineer does not have full control over corporate PKI that we are required to utilize.  The latter option addresses your "superfluous security" concern, since a podman or docker secret can be mounted.  Those that share your concern about the security can omit the passphrase, if they wish.  Now, this inflexibility likely means that I will be unable to use Traefik as a TCP reverse proxy for my application stack.

bluet:
It's really sad to see this PR been declined.

In government systems team, we need to use gov issued certs, and they are password (passphrase) protected in case someone leaked the file by accident.

In the process of replacing old nginx reverse proxy with traefik, everything's fine and smooth, until the last step... There's no way to supply password file from docker secret to traefik...

dmitry-xypro:
@rtribotte  I believe you guys should reconsider your decision from 2020. Traefik supports windows deployment without containers. In this scenario the passphrase is absolutely necessary. Further, as others have mentioned, using corporate certificates often doesn't allow us (or our customers) to choose whether to have or not have a passphrase since corporate policy mandates them. These days security standards are being tightened to new levels due to the uptick in ransomware attacks. Having additional security features is better than calling them "edge cases". 

Can you please consider reopening this one and the associated ticket #1262 to address these security concerns of 2023+? 

- TYIA

rtribotte:
Hello @dmitry-xypro,

Thanks for sharing this.
This PR is not an adequate place to discuss this, could you please open a new issue to explain your use case and why we should reconsider supporting certificate decryption?

PR #6519, opened 2020-03-18
Doc: fix wrong name of config format

Nek-:
### What does this PR do?

Fix a typo.

### Motivation

No contribution is totally useless. (again!)

### More

- [x] Added/updated documentation

PR #6543, opened 2020-03-21
Fix tab name

mavimo:
### What does this PR do?

Fix tab name in sample

### Motivation

In the current docs for the page

https://docs.traefik.io/https/acme/

expanding examples:

image

there is a mismatch on the tab `File (TOML)`

### More

- [ ] Added/updated tests


ldez:
Related to #6519

PR #6558, opened 2020-03-23
Add HTTP Provider

### LINKED ISSUE #1053: Feature request: HTTP backend
Hey,

This feature is called [Service Discovery](https://tyk.io/docs/tyk-api-gateway-v-2-0/api-management/service-discovery/) in [Tyk](https://tyk.io). The basic idea is that there's external API I can setup that Traefik uses to "resolve" API endpoints. For example I could setup an application and store all routes in database, then expose an endpoint that would return these routes, so Traefik can query and use them to route incoming requests. The flow is as follows:

1. Incoming request arrives to Traefik for some domain/port/path(optionally)
2. Traefik asks HTTP backend to resolve incoming query
3. Traefik uses received information to properly route incoming request
4. Next incoming request arrives for the same domain/port/path(optionally)
5. Traefik uses cached information in 2. to properly route this request
6. Traefik purges cache after 10+ seconds and will query HTTP backend next time such request arrives

What do you think?

dtomcej:
### What does this PR do?

This PR introduces a new provider (http) that polls an endpoint for a published configuration, and provides it to the server for use.

### Motivation

Fixes #1053 

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/providers/overview.md
+++ b/docs/content/providers/overview.md
@@ -35,9 +35,10 @@ Below is the list of the currently supported providers in Traefik.
 | [Rancher](./rancher.md)               | Orchestrator | Label                      |
 | [File](./file.md)                     | Manual       | TOML/YAML format           |
 | [Consul](./consul.md)                 | KV           | KV                         |
-| [etcd](./etcd.md)                     | KV           | KV                         |
+| [Etcd](./etcd.md)                     | KV           | KV                         |
 | [Redis](./redis.md)                   | KV           | KV                         |
 | [ZooKeeper](./zookeeper.md)           | KV           | KV                         |
+| [Http](./http.md)                     | Manual       | JSON format                |
 
 !!! info "More Providers"
 
--- a/docs/content/providers/http.md
+++ b/docs/content/providers/http.md
@@ -4,7 +4,7 @@ Provide your configuration via an http endpoint and let Traefik do the rest!
 
 ## Routing Configuration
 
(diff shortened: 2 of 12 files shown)


docs/content/providers/overview.md
```diff
@@ -35,9 +35,10 @@ Below is the list of the currently supported providers in Traefik.
 | [Rancher](./rancher.md)               | Orchestrator | Label                      |
 | [File](./file.md)                     | Manual       | TOML/YAML format           |
 | [Consul](./consul.md)                 | KV           | KV                         |
-| [etcd](./etcd.md)                     | KV           | KV                         |
+| [Etcd](./etcd.md)                     | KV           | KV                         |
```
SantoDE: Why that change?

dtomcej: The lack of capitalization in that one case bugged me.

dtomcej:
I am going to close this PR for 3 reasons:

1. This feature has been declined in the past, and we don't feel that the circumstances have changed much in the meantime. As the saying goes, "No is temporary, yes is forever".

2. Although this feature does fit in line with the other providers, and their goals, at this time, there is a large amount of overlap between this provider and other providers, and there is not a clear public use case to support its inclusion.

3. Due to how Traefik is written, it can be used as a library, and this provider added into the implementation. This allows  providers to be implemented in separate projects without requiring a fork to be maintained.

Hopefully this comment helps other users in the future if they come across this PR.

PR #6571, opened 2020-03-24
Doc improvement on entrypoints whitelist for v1.7

### LINKED ISSUE #5816: Documentation: Whitelisting not accurate enough

Documentation Bug

This concerns 1.x (1.7 specifically being used here) but may also apply to 2.x, and concerns the documentation more than the software.

### What did you do?

We added our AWS VPC CIDR to our whitelisting.

Secure access to backends from selected IP ranges and from other resources within our VPC deployment.

### What did you see instead?

In addition, we spotted we were now able to access protected resources from anywhere.

At this point, allow me to elaborate.

We have a VPC with an _internal_ ALB. Traffic from the outside world is routed in via a Global Accelerator. We use ECS a lot, so applied Docker Labels to containers to advertise themselves and apply source range whitelisting with X-Forwarded-For headers.

Initially, this worked fine to allow private access from say our office IP. But then we realised we have "sibling" resources within the VPC also needing to make connections so added our private IP CIDR range and asked the clients to reach us via the ALB. This, too, worked.

But, and here is the crux of this ticket: it seems by switching on x-forwarded-for, we are not asking Traefik to *only* examine x-forwarded-for, but add the x-forwarded-for headers to the source addresses of the traffic being inspected. If only the x-forwarded-for headers where used, traffic from foreign parties would never match, but because it's additive the ALB's own source IP addresses match too. Whoops.

Perhaps explain in the documentation that source range is always the source of the TCP transmission, and that switching on x-forwarded-for ADDS to this address, and does not REPLACE them.

```
1.7.18
```

AWS with Global Accelerator linked to an _internal_ Application Load Balancer forwarding to various Traefik instances hosted by ECS.

Docker labels (CloudFormation in use):
```
traefik.frontend.whiteList.useXForwardedFor: "true"
traefik.frontend.whiteList.sourceRange: !Join [", ", [!Ref SourceIPRangesStaff,  "172.20.0.0/16"]] # 172.20/16 is our CIDR
```

By allowing our CIDR, it allows all traffic passing through the ALB in, too.

ddtmachado:
### What does this PR do?

Improve doc on v1.7 regarding entrypoints whitelisting by adding a brief description of the behavior when enabling `useXForwardedFor` and a warning to users behind another load balancer.

### Motivation

To fix #5816

### More

- [ ] ~~Added/updated tests~~

### DIFF
--- a/docs/configuration/entrypoints.md
+++ b/docs/configuration/entrypoints.md
@@ -485,6 +485,11 @@ To enable IP white listing at the entry point level.
       # useXForwardedFor = true
 ```
 
+By setting the `useXForwardedFor` option, the `sourceRange` addresses will be matched against the request header `X-Forwarded-For` address list, from left to right.
+
+!!! danger
+    When using Traefik behind another load-balancer, it's own internal address will be appended in the `X-Forwarded-For` header. Be sure to carefully configure the `sourceRange` as adding the internal network CIDR, or the load-balancer address directly, will cause all requests coming from it to pass through.
+
 ## ProxyProtocol
 
 To enable [ProxyProtocol](https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) support.
--- a/docs/configuration/entrypoints.md
+++ b/docs/configuration/entrypoints.md
@@ -488,7 +488,7 @@ To enable IP white listing at the entry point level.
 By setting the `useXForwardedFor` option, the `sourceRange` addresses will be matched against the request header `X-Forwarded-For` address list, from left to right.
 
 !!! danger
-    When using Traefik behind another load-balancer, it's own internal address will be appended in the `X-Forwarded-For` header.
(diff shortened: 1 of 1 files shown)


docs/configuration/entrypoints.md
```diff
@@ -485,6 +485,13 @@ To enable IP white listing at the entry point level.
       # useXForwardedFor = true
 ```
 
+By setting the `useXForwardedFor` option, the `sourceRange` addresses will be matched against the request header `X-Forwarded-For` address list, from left to right.
+
+!!! danger
+    When using Traefik behind another load-balancer, it's own internal address will be appended in the `X-Forwarded-For` header.
```
mpl, line 491:
```suggestion
    When using Traefik behind another load-balancer, its own internal address will be appended in the `X-Forwarded-For` header.
```

PR #6581, opened 2020-03-25
spnegoclient middleware

bothejjms:
### What does this PR do?

Adds an spnegoclient middleware which can be used to add SPNEGO headers to the outgoing http requests

The client can be turned on in a loadBalancer with the spnegoClient option.

### Motivation

It allows us to add traefik as a sidecar in the containers and handing the outgoing spnego authentication for the application in the container.

### More

- [-] Added/updated tests - No. The logic is really in the added go kerberos library
- [-] Added/updated documentation - Not yet. I will add the docs later.


mpl:
Sorry, we do not fully understand what you are trying to achieve, so we do not see the need for this PR.
(And we suspect there might be solutions to your need that do not require changing traefik.)
Can you please open an issue or a proposal that explains precisely what you're trying to achieve?

Issue #5853, opened 2019-11-16
Allow to add path preifx for both api and dashboard endpoints


huangjunwen:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

As #5374 has been fixed, now dashboard assets can be served under any path prefix, but it continues to fetch api data from `/api`, which seems to be determined at build time or hard coded (?)

I would suggest to add something like `--api.pathprefix=/path/prefix` then these endpoints will be served under `/path/prefix/api` and `/path/prefix/dashboard`. Which will make them easier to integrate into existing auth system: specific cookie under some path prefix.

traefiker:
Hi! I'm Træfiker :robot: the bot in charge of communication regulation.

Thanks for your interest in Traefik!

We dedicate the issue tracker to bug reports and feature requests only. My advanced AI has spotted that your issue might be a configuration problem or relates to something that doesn't look like a bug.

To confirm this, please join our [Community Forum](https://community.containo.us/) and reach out to us on the [Traefik section](https://community.containo.us/c/traefik).

In case I'm wrong (well, that would be embarrassing :sweat_smile:), my developers will re-open the issue and fix me!

In the meantime, you can double check [Traefik's documentation](https://docs.traefik.io/).

MichaelSchmidle:
Following the [conversation in the Community Forum](https://community.containo.us/t/how-to-add-path-prefix-to-api-endpoints/2974):

From my point of view, this is more a bug report than a feature request. Here's why:

The dashboard should work with relative URLs **in general**. #5374 was fixed regarding URLs to dashboard assets only. What's missing now is a fix regarding URLs to the API endpoint: These should be relative as well, effectively allowing to put the entire Traefik dashboard/API behind a path prefix.

andrewsav-datacom:
The goal is to be able to host the dashboard at http://mysite/my/crazy/random/prefix. I suggest this as a test case that the implementation (bugfix) has worked. if you can navigate to that url and have the dashboard loaded and behave, then chances are the fix has worked.

firecow:
Have this feature/bug been fixed, I can't find any documentation as to how the dashboard n' api endpoints is supposed to be changed?

kdankert:
Are there any updates concerning this issue. I would love to host bot my dashboard and my api under the PathPrefix /traefik, cause I want to use the path /api myself.

nitishpt:
[...]

PR #6609, opened 2020-03-31
Fix api path for prefixed api/dashboards path

dschniepp:
The current implementation does not allow to use the dashboard in a prefixed setup. In our setup the api  and the dashboard are exposed on `/random-path`. The dashboard therefore will be delivered at `/random-path/dashboard/#/` but expects the api at `/api` but actually needs to search in `/random-path/api`.

Therefore the fix will change the path to a relative path, in order to search the api next to the dashboard path. This PR will affect the build process of the static files.

This issue is also described here #5853.

### DIFF
--- a/webui/quasar.conf.js
+++ b/webui/quasar.conf.js
@@ -118,11 +118,11 @@ module.exports = function (ctx) {
       env: process.env.APP_ENV === 'development'
         ? { // staging:
           APP_ENV: JSON.stringify(process.env.APP_ENV),
-          APP_API: JSON.stringify(process.env.APP_API || '/api')
+          APP_API: JSON.stringify(process.env.APP_API || '../api')
         }
         : { // production:
           APP_ENV: JSON.stringify(process.env.APP_ENV),
-          APP_API: JSON.stringify(process.env.APP_API || '/api')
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

this fix doesn't work because the api path cannot be relative.

dschniepp:
Thanks for the feedback. What do you mean by the simple case? In case it is not prefixed?

ebuildy:
Hello @ldez , do you have any clue about how to deal with this? I check on Quasar documentation about "runtime configuration" but see nothing about :/ 

Thanks you,

ebuildy:
Also @dschniepp , why `..` ? the easiest way is to remove the ``/`` so ``api`` will be relative to full URL.

dschniepp:
@ebuildy as dashboard is exposed with an pending `/` and therefore it is necessary to go one directory up.

ebuildy:
Did you try with only ``api``? many web sites and SPA do that.

mpl:
FYI (and because there was some noise about it from other channels), the conversation on this PR was closed because, as the PR itself is controversial, it is better to keep all discussion about it in the same place, i.e. on the issue: #5853 .

PR #6629, opened 2020-04-04
Dont check for configuration equality

### LINKED ISSUE #6626: Traefik v1.7 misses events from k8s, does not serve services

Bug

### What did you do?

Just running traefik as an ingress controller on Kubernetes.

Expect to route traffic to all ready backends/pods.

### What did you see instead?

Some ready pods were not treated by Traefik, and thus, was not routed traffic to. Meanwhile observed `Skipping same configuration for provider` messages.

```
Version:      v1.7.24
Codename:     maroilles
Go version:   go1.14.1
Built:        2020-03-25_04:34:11PM
OS/Arch:      linux/amd64
```

Kubernetes deployment.

### Hints

I found that, the mentioned message comes from https://github.com/containous/traefik/blob/v1.7/server/server_configuration.go#L324. If a pod goes to a NotReady state, it generates an event for traefik. It starts processing it, but if recovery is too fast (i.e. the new event comes in while still processing the first one) the code pointed finds that existing and new configuration are equal, thus skipping. Later, processing the first event finishes, resulting in the service brought and remaining in down state from Traefik's perspective.

Perhaps even during the --providersthrottleduration interval such a recovery event will be dropped.

On my environment, reloading the configuraion (the function at https://github.com/containous/traefik/blob/v1.7/server/server_configuration.go#L34) sometimes takes a few seconds, so it is very easy to reproduce the issue. I have a pod with 1 sec readinessProbe, and just need to flap its' readiness down and up, thus Traefik receives two events, with the latter being dropped.

rkojedzinszky:
Fixes #6626

### What does this PR do?

Fixes #6626

### Motivation

Fix the bug.

### More

- [ ] Added/updated tests

### DIFF
--- a/server/server_configuration.go
+++ b/server/server_configuration.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"reflect"
 	"sort"
 	"strings"
 	"time"
@@ -309,7 +308,6 @@ func buildServerRoute(serverEntryPoint *serverEntryPoint, frontendName string, f
 func (s *Server) preLoadConfiguration(configMsg types.ConfigMessage) {
 	providersThrottleDuration := time.Duration(s.globalConfiguration.ProvidersThrottleDuration)
 	s.defaultConfigurationValues(configMsg.Configuration)
-	currentConfigurations := s.currentConfigurations.Get().(types.Configurations)
 
 	if log.GetLevel() == logrus.DebugLevel {
 		jsonConf, _ := json.Marshal(configMsg.Configuration)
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

sorry but this change is not a valid solution for #6626: there are a lot of side effect and that will produce a lot of bugs.

thanks anyway.

rkojedzinszky:
@ldez As I understood the code, that piece of code was responsible to make some optimization, to not reload configuration that matches with existing one. Unfortunately, that causes my issues, that reloading configuration takes longer than time difference between two different configuration events.

PR #6633, opened 2020-04-04
Refactor configuration reload/throttling

### LINKED ISSUE #6632: Ensure that configurationValidatedChan ever receives the latest different configuration from each provider

Feature / Enhancement

I was debugging issue #6626 and meanwhile noticed that configurationValidatedChan channel is buffered, which means, that if during configuration reload, even due to throttling and same configuration checks present, it is still possible to enqueue a flapping state.

Imagine as:

```
| Reload time
+----------------------------------------------------------------------------------------+
+-- throttling --+-- throtrling --+-- throttling --+-- throttling --+-- throttling --+
c1               c2               c1               c2               c1               c2
```

So, if a reload takes multiple throttling time intervals, then you can see, that different c1 and c2 configurations will be enqueued, and with long relad times, will cause services flapping long after
they became stable in real.

I've already refactored the relevant parts, but still needs some cleanup. Is it worth to create a PR?

rkojedzinszky:
### What does this PR do?

Refactor configuration reload throttling/processing to remove configuration flapping, and to improve responsiveness under high input from (a) provider(s).

### Motivation

Some scenarios around the handling and reloading of configurations were not behaving well. Notably,
1) when receiving an alternation of recurring configurations, it is useless and costly to load the intermediate configuration states (configuration flapping).
2) when receiving lots of configurations from one (or several) provider(s) for an extended duration, the incurred stress on the configuration watcher was such that none, or at least not enough, configurations were actually reloaded.

In addition, in the past the providersThrottleDuration option (either user-set or through its default value) was applied to all the providers, including some internal ones, such as the TLS ALPN acme challenge one. This was problematic because when one wanted to throttle high input providers (e.g. Kubernetes), one would also implicitly throttle the acme one, which is not desirable if one wants a responsive certificates renewal system.

Therefore, this PR, through a refactoring of the whole throttling and reloading of configuration addresses the problems described above. And notably, it introduces a new abstraction which allows the throttling to be applied independently to each provider. For example, in the current state, there never is any throttling applied to the TLS ALPN acme one.

Fixes #6632

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: Romain rtribotte@users.noreply.github.com

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -181,7 +181,7 @@ func setupServer(staticConfiguration *static.Configuration) (*server.Server, err
 	httpChallengeProvider := acme.NewChallengeHTTP()
 
 	// we need to wait at least 2 times the ProvidersThrottleDuration to be sure to handle the challenge.
-	tlsChallengeProvider := acme.NewChallengeTLSALPN(time.Duration(staticConfiguration.Providers.ProvidersThrottleDuration) * 2)
+	tlsChallengeProvider := acme.NewChallengeTLSALPN(5 * time.Second)
 	err = providerAggregator.AddProvider(tlsChallengeProvider)
 	if err != nil {
 		return nil, err
@@ -265,7 +265,6 @@ func setupServer(staticConfiguration *static.Configuration) (*server.Server, err
 	watcher := server.NewConfigurationWatcher(
 		routinesPool,
 		providerAggregator,
-		time.Duration(staticConfiguration.Providers.ProvidersThrottleDuration),
 		getDefaultsEntrypoints(staticConfiguration),
 		"internal",
 	)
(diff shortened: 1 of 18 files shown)


juliens:
Thx for your PR!

Could you provide us a unit test to show us the use case it should handle?

rkojedzinszky:
> Thx for your PR!
> 
> Could you provide us a unit test to show us the use case it should handle?

I will need time for that, but please check https://github.com/containous/traefik/issues/6632#issue-593783690

rtribotte:
Hello @rkojedzinszky,

We took the liberty of experimenting and iterating on the PR.
We hope that's ok.

rkojedzinszky:
> Hello @rkojedzinszky,
> 
> We took the liberty of experimenting and iterating on the PR.
> We hope that's ok.
> 
> WDYT?

Sure. Happy to see that the PR is interesting.

mpl:
Hello @rkojedzinszky ,

Could you please let us know if this new version of the PR as it is right now fixes the issue for you? In particular, if it works for the flapping state problem?

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks 👍

mpl (approved):

PR #6664, opened 2020-04-12
Allows multi-level KV prefixes

niki-timofe:
### What does this PR do?
Allows slashes in root key of KV-storages.
 

### Motivation
We've got single KV storage shared across many of teams. Each team have its own directory in it. My team soon to have some separated Traefiks. So it would be cool, if we can separate configurations as:

```
team-a/traefik-1
team-a/traefik-2
```

There is no information about inadmissibility of slashes neither in documentation nor in logs, only message: "invalid root label". I see two possibilities of resolving this situation: add this constraint to documentation or add this feature.
As it's technically not too complex to implement, I chose 2nd variant.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/kv/kv_node.go
+++ b/pkg/config/kv/kv_node.go
@@ -20,13 +20,12 @@ func DecodeToNode(pairs []*store.KVPair, rootName string, filters ...string) (*p
 	var node *parser.Node
 
 	for i, pair := range sortedPairs {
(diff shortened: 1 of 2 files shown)


jstdips:
It is possible to review and merge this PR? Will be helpful for us using Zookeeper as backend config store

dacameron:
I can't see this in the latest release: v2.2.11. Any chance of seeing it soon? Thanks!

Issue #6068, opened 2019-12-19
Add a way to filter the body of a request for forward authentication


llacroix:
Feature
======

The new option would let traefik send a request to an endpoint without passing the full request to the auth endpoint. As it is right now, the whole request would be sent to traefik then to the forwardAuth then the same request body would be sent to the actual endpoint.

One of the reason would be to prevent the authServer from receiving too much traffic and making sure the authServer has only what it needs to validate a request. 

So if a request contains multiple megabytes of data, it won't try to stream the data to the auth server if it doesn't have to. It's okay when data is received as GET but when request are made with POST method or possibly any other method type, it may make the forwardAuth endpoint more or less complicated to implement.

For example, you're using a framework that has a particular way to handle requests and automatically parse the body depending on the headers. But the forwardAuth route is defined with a basic type of data, as the request is received, it may end up in the body parser which would trigger a body parsing error as the body parsed doesn't match the body that is supposed to be parsed by the route.

Stripping the body of data makes it easy to keep the request small and reduce the amount of time/errors an endpoint would take to parse the body if it was present. On a security standpoint it could also prevent a person from sending a malformed body to the auth endpoint. For traefik, it means the request can be sent as soon as the header is loaded and while the auth request is being sent, it can spend time buffering the request until the auth response is received. Once the auth response is received it can transmit the buffer to the actual endpoint. In theory it should make the forwardAuth middleware more responsive.

jamct:
@llacroix At the moment the body of the request is not forwarded to the auth server. You can see it [here in the code (line 80)](https://github.com/containous/traefik/blob/master/pkg/middlewares/auth/forward.go):
`forwardReq, err := http.NewRequest(http.MethodGet, fa.address, nil)`
The third parameter is `nil`. I would like to forward the body at my auth server (to evaluate rules in Open Policy Agent) so I would like to prepare a PR. Maybe it is a good idea to add a new setting to activate forwarding of the body.

llacroix:
Ah yes that's correct.. thought I do remember having issues, but that could have been caused by this. My endpoint was probably expecting JSON and received an empty body. Having a setting to enable the feature would be the very least to do while keeping default behaviour.

daudimeck:
is there a way to allow the passing of the body to the auth endpoint at the moment?

jamct:
I just added PR #6669. It adds a new parameter forwardBody (boolean) to activate forwarding of the request body.

jbdoumenjou:
Hello,

[...]

PR #6669, opened 2020-04-12
feature: forward body

jamct:
### What does this PR do?

This PR adds the ability to forward the request body to the auth server when using ForwardAuth Middleware.
There is a new option for this middleware called `ForwardBody`. If true, the body will be forwarded.

### Motivation

I want to use Traefik in combination with a gatekeeper using Open Policy Agent to make decisions. Decisions should be made based on request bodies.

This topic was discussed in Issue #6068  

### More

- [ ] Added/updated tests

### Additional Notes

If merged I can add a PR for documentation as well.

### DIFF
--- a/pkg/config/dynamic/middlewares.go
+++ b/pkg/config/dynamic/middlewares.go
@@ -142,6 +142,7 @@ type ForwardAuth struct {
 	TLS                 *ClientTLS `json:"tls,omitempty" toml:"tls,omitempty" yaml:"tls,omitempty"`
 	TrustForwardHeader  bool       `json:"trustForwardHeader,omitempty" toml:"trustForwardHeader,omitempty" yaml:"trustForwardHeader,omitempty" export:"true"`
 	AuthResponseHeaders []string   `json:"authResponseHeaders,omitempty" toml:"authResponseHeaders,omitempty" yaml:"authResponseHeaders,omitempty"`
+	ForwardBody         bool       `json:"forwardBody,omitempty" toml:"forwardBody,omitempty" yaml:"forwardBody,omitempty" export:"true"`
 }
 
 // +k8s:deepcopy-gen=true
--- a/pkg/middlewares/auth/forward.go
+++ b/pkg/middlewares/auth/forward.go
@@ -1,6 +1,7 @@
 package auth
 
 import (
+	"bytes"
 	"context"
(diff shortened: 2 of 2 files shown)


notsureifkevin:
Thank you for your contribution, as noted in the PR template, new features / capabilities should include documentation / tests. This PR will be reviewed for design once those requirements have been met.

Some concerns which should be addressed in the test include:

- Performance implications of a large amount of requests w/ large content body
- Validating that FA tests pass with this feature enabled

Shemeikka:
Hi,

This feature would be extremely useful for an authentication services that are required to inspect the request body. I would really like to have this feature.

Documentation could be as simple as
```
Set the forwardBody option to true to forward request body. This can affect performance if the request has a large body.
```

I think that everyone realizes that there are performance implications when forwarding body, especially if it's a large one. I personally think that it would be enough to mention this on the documentation and if someone creates an issue about poor performance, then some optimizations could be done in the future.

@notsureifkevin What kind of tests are required? Are there any previous tests that could be used as a base for this?

@jamct Can you add tests and documentation if we can agree on the text and required tests.

notsureifkevin:
@Shemeikka 

Thank you for your contribution to the conversation.

> What kind of tests are required? Are there any previous tests that could be used as a base for this?

I suspect there are tests covering the base functionality of ForwardAuth, those would likely require some refactoring in order to support testing the existing functionality while this feature was enabled. We'd also need tests convering the specific use cases that would cover this test being used (is it working as intended).

totemcaf:
To handle the performance problem an additional value could be added to limit the size of the passed body.

Something like: maxBodySize.

rtribotte:
Hello @jamct,

After discussion, we have concluded that we do not want to introduce this feature as a core functionality.
Our concern is about memory consumption, as the ForwardAuth middleware would read the body for each request.

Unfortunately, I'm going to close this pull request, but it would be great if you can add this feature as a plugin.
Plugins are introduced since the v2.3RC, and we encourage you to check out the [demo repo plugin](https://github.com/containous/plugindemo) where you will find how to create a plugin.
One of the advantages of the plugin system is that we can get an indication of a middleware popularity and usage, which could lead to reconsider including it as a core feature.

Nevertheless, thank you for your time and interest in Traefik!

PR #6689, opened 2020-04-16
Add Access log chapter for migration v1->v2

MartinKoerner:
### What does this PR do?

Add a hint to the migration documentation, that the default configuration of the access logs in v2 won't log the same details as v1 anymore.

### Motivation

Stumbled across the need of reconfiguration later in my migration path and would loved to get a hint right at the beginning of my configuration adaptions.

### More

- [ ] Added/updated tests

### Additional Notes

-

### DIFF
--- a/docs/content/migration/v1-to-v2.md
+++ b/docs/content/migration/v1-to-v2.md
@@ -801,7 +801,10 @@ There is no more log configuration at the root level.
 
 ## Access Logs
 
-Access Logs are configured in the same way as before. But all request headers are now filtered out by default in Traefik v2. So during migration, you might want to consider enabling some needed fields (see [access log configuration](../observability/access-logs.md)).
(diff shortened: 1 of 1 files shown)


docs/content/migration/v1-to-v2.md
```diff
@@ -799,6 +799,10 @@ There is no more log configuration at the root level.
     --log.format=json
     ```
 
+## Access Logs
+
+Access Logs are configured in the same way as before. But v2 filters request headers by default. So during migration, you might want to consider enabling some needed fields (see [access log configuration](../observability/access-logs.md)).
```
mpl, line 804:
```suggestion
Access Logs are configured in the same way as before. But all request headers are now filtered out by default in Traefik v2. So during migration, you might want to consider enabling some needed fields (see [access log configuration](../observability/access-logs.md)).
```

Issue #6688, opened 2020-04-16
Ping resulting in 503 http code during graceful shutdown


L3o-pold:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

Request shutdown of traefik and wait for a livenessProbe check.

It's well documented in the [1.7](https://docs.traefik.io/v1.7/configuration/ping/#using-ping-for-external-load-balancer-rotation-health-check) doc but not in the [2.x](https://docs.traefik.io/operations/ping/).

When used in a livenessProbe in k8s an 503 http code is seen as [an error](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request).

### What did you expect to see?

200 on ping request

### What did you see instead?

503 on ping request

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
2.1.3
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
[ping]
  entryPoint = "ping"
[api]
  dashboard = true

[entryPoints]
  [entryPoints.ping]
    address = ":8081"
    [entryPoints.ping.transport]
      [entryPoints.ping.transport.lifeCycle]
        requestAcceptGraceTimeout = "60s"
        graceTimeOut = "2h"
```

Kubernetes example:

```yaml
livenessProbe:
  httpGet:
    path: /ping
    port: 8081
    scheme: HTTP
  initialDelaySeconds: 10
  timeoutSeconds: 2 
```

### If applicable, please paste the log output in DEBUG level (`--log.level=DEBUG` switch)

```
time="2020-04-16T14:10:56Z" level=error msg="close tcp [::]:80: use of closed network connection" entryPointName=http
time="2020-04-16T14:10:56Z" level=error msg="close tcp [::]:8080: use of closed network connection" entryPointName=traefik
time="2020-04-16T14:10:56Z" level=error msg="accept tcp [::]:8080: use of closed network connection" entryPointName=traefik
time="2020-04-16T14:10:56Z" level=error msg="accept tcp [::]:80: use of closed network connection" entryPointName=http
[16/Apr/2020:14:11:00 +0000] "GET /ping HTTP/1.1" 503 19 "-" "-" 18 "ping@internal" - 0ms
[16/Apr/2020:14:11:14 +0000] "GET /ping HTTP/1.1" 503 19 "-" "-" 21 "ping@internal" - 0ms
```
[...]

PR #6696, opened 2020-04-17
Add custom ping http code when Traefik is terminating

L3o-pold:
### What does this PR do?

Allow custom http code in ping route when Traefik is terminating (graceful shutdown)

### Motivation

Issue https://github.com/containous/traefik/issues/6688 
Better handling of graceful shutdown with Kubernetes livenessprobe.

### More

- [x] Added/updated tests

### DIFF
--- a/integration/simple_test.go
+++ b/integration/simple_test.go
@@ -161,6 +161,37 @@ func (s *SimpleSuite) TestRequestAcceptGraceTimeout(c *check.C) {
 	}
 }
 
+func (s *SimpleSuite) TestCustomPingTerminationStatusCode(c *check.C) {
+	file := s.adaptFile(c, "fixtures/custom_ping_termination_status_code.toml", struct{}{})
+	defer os.Remove(file)
+	cmd, display := s.traefikCmd(withConfigFile(file))
+	defer display(c)
+
+	err := cmd.Start()
+	c.Assert(err, checker.IsNil)
+	defer cmd.Process.Kill()
+
+	// Wait for Traefik to turn ready.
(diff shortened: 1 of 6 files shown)


pkg/ping/ping.go
```diff
@@ -29,7 +30,11 @@ func (h *Handler) WithContext(ctx context.Context) {
 func (h *Handler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
 	statusCode := http.StatusOK
 	if h.terminating {
-		statusCode = http.StatusServiceUnavailable
+		if h.TerminatingStatusCode > 0 {
+			statusCode = h.TerminatingStatusCode
+		} else {
+			statusCode = http.StatusServiceUnavailable
+		}
```
dtomcej, line 37:
```suggestion
		statusCode = http.StatusServiceUnavailable

		if h.TerminatingStatusCode > 0 {
			statusCode = h.TerminatingStatusCode
		}
```

integration/simple_test.go
```diff
@@ -161,6 +161,57 @@ func (s *SimpleSuite) TestRequestAcceptGraceTimeout(c *check.C) {
 	}
 }
 
+func (s *SimpleSuite) TestCustomPingTerminationStatusCode(c *check.C) {
+	s.createComposeProject(c, "reqacceptgrace")
```
dtomcej, line 165: Instead of using compose for this, since you are never actually checking the response, could you use a non-docker test server like here:

https://github.com/containous/traefik/blob/master/integration/headers_test.go#L38

This would make the test faster, since it does not require docker, and would make the code cleaner.

Issue #5063, opened 2019-07-05
X-Request-Start header


scalp42:
Hi folks,

Looking at the headers documentation (https://docs.traefik.io/v2.0/middlewares/headers/), we can't figure out if Traefik supports setting a dynamic header based on micro/milliseconds since epoch.

Example with Nginx:

```
proxy_set_header X-Request-Start "t=${msec}";
```

or Apache:

```
RequestHeader set X-Request-Start "%t"
```

By setting the header, it allows visibility into requeust queuing to the app from farther upstream (Traefik in our scenario).

It's also needed for APM vendors (New Relic, Scout, Instana, etc):

- https://docs.newrelic.com/docs/apm/applications-menu/features/request-queue-server-configuration-examples
- https://docs.scoutapm.com/#http-header

Let us know if it's possible.

Thanks in advance ✌️

ldez:
Hello,

We think that your request is covered by the proposal #5036

Issue #5036, opened 2019-06-27
Custom Variables in Proxy Headers


discostur:
**Feature**

In Nginx or Haproxy you can use a variety of server variables like 

* destination IP
* source IP
* etc.

All of that variables you can use to build your own proxy headers, for example:

```
http-request set-header X-dst-ip "%[dst]"
http-request set-header X-cl-ip "%[req.hdr(X-Forwarded-For,-1)]"  if  found_http_xff
```

In Traefik you can set your own custom headers but only with static content - would it be possible to use some variables in the future? Or even use conditions like if / else statements?

Thanks

geraldcroes:
related to #5030

scalp42:
Related to #5063 we would also need to have "dynamic" headers (like `msec` one in Nginx) not only static ones.

axwell:
Any news about this ? 
I need proxy headers v2 (as described in the protocol http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt)
to be added to the TCP service in order for my rabbitmq cluster to show the real IP of the connected clients, instead of the docker cluster IP.

https://www.rabbitmq.com/networking.html#proxy-protocol

Reapor-Yurnero:
My use case in #7279 can be better resolved with this feature. A stronger and more flexible way to manipulate the headers as a middleware is very necessary in my opinion.

mathiasAichinger:
Are plugins the way to go, to solve such challenges?
@scalp42 Today, I have started to work on a plugin which should solve your issue. https://github.com/EasySolutionsIO/traefikxrequeststart

ibigbug:
looks like it's be quiet for a while - but would be great to see this happening

tfny:
This issue can be found in [#8897](https://github.com/traefik/traefik/issues/8897) combined with [#5030](https://github.com/traefik/traefik/issues/5030).  

We need help solving this problem, so join the [Traefik Proxy 3.0 Hackaethon](https://info.traefik.io/traefik-proxy-3-0-hackaethon?cta=pop) to help us solve it and win some awesome swag (and maybe a prize or two).

dberardo-com:
i do see a lot of benefits from dynamically added headers, especially in combination with the ForwardAuth middleware, which currently cannot pass the payload of the HTTP request, but only the headers themselves, and so it limits heavily the possibilities to make RBAC from incoming requests.

fabiendv:
[...]

Issue #4640, opened 2019-03-20
Add support for Correlation IDs


jan-martens:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

A few weeks ago, i created PR #4483 which adds support for Correlation IDs to Traefik which was closed. @juliens suggested that this might be worth discussing in a separate issue again, so here it is. 

To recap my reasoning for this feature request:

> In our environment we make heavy use of Correlation IDs to track requests throughout our mircoservice infrastructure. We recently moved from Nginx to Traefik and it brought us a lot of benefits, but the possibility to add custom headers is somewhat limited compared to Nginx. We need to add a new random UUID to every requests that comes in from outside our infrastructure to properly track it's path through the environment. This could have been done with the help of custom request headers, but those have some disadvantages:
> 
> If the header is already present, it will get overwritten. Also you cannot generate a new random UUID for every request. All you could do is generate a random UUID for every frontend each time Traefik reloads the configuration ( with the help of the spring template library ) but then you can have a lot of requests with the same correlation id. This is why we wrote a [http middleware](https://gitlab.com/JanMa/correlation) that adds these headers to every incoming request and would like to add it here.

There has already been PR #2290 that addressed a similar issue but it was closed at the time in favor of supporting OpenTracing. Seeing that support for OpenTracing is now present in Traefik I would like to re-open the discussion of adding Correlation ID support to Traefik. 

> In my opinion the use case for a correlation id differs quite a lot from using a tracing engine. With correlation ids it is very easy to track an external request through all service calls in your environment. You "just" have to make sure your applications log the corresponding header name in their access logs and can easily gather all relevant log entries in a tool like Kibana. Having this information in your logging stack is incredibly useful for debugging, especially in a microservice environment. Using a tracing engine on the other hand is useful when you want to gather low level insights of your inter-service communication and see for example how long single remote procedure call takes. Also implementing full tracing support in all your services takes a lot more effort.

I think having both features available in Traefik would be very beneficial. Please let me know what your thoughts are about this :-)

Kind regards,
Jan

jan-martens:
Any ideas/feedback for my proposal?

Kind regards,
Jan

Sarke:
@JanMa, here's a workaround I've shared this previously that you can use.

You can enable the Jaeger tracer and set the sampling rate to 0, but it will still send the header.  Some quick benchmarks show that it doesn't add any significant overhead.

```toml
[tracing]
  [tracing.jaeger]
[...]

Issue #4519, opened 2019-02-23
Allow traefik variables in ingress.kubernetes.io/custom-response-headers


sokoow:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

It'd be really cool if I could do an annotation like this:
```
ingress.kubernetes.io/custom-response-headers: X-forwarded-for: $ClientHost
```

and that'd point to a value of ClientHost that traefik log entry is spitting out. Possible?

ldez:
Close in favor of  #5036 - Feel free to subscribe there for updates.

PR #6707, opened 2020-04-21
custom header UUID

Sarke:
### What does this PR do?

Adds ability to create a custom header with a UUIDv4 (aka "Correlation ID" or "X-Request-ID").

### Motivation

This is something several users want (quick search #4640, #1333, #4483, #2290, #2377), including myself.  An **easy** way to match Traefik request logs with other service logs, without having to create an entire tracing ecosystem or additional services.  

It is a common proxy feature.

### More

- [x] Added/updated tests

### Additional Notes

@JanMa [put it well](https://github.com/containous/traefik/pull/4483#issuecomment-462833870):

> In my opinion the use case for a correlation differs quite a lot from using a tracing engine. With correlation ids it is very easy to track an external request through all service calls in your environment. You "just" have to make sure your applications log the corresponding header name in their access logs and can easily gather all relevant log entries in a tool like Kibana. Having this information in your logging stack is incredibly useful for debugging, especially in a microservice environment. Using a tracing engine on the other hand is useful when you want to gather low level insights of your inter-service communication and see for example how long single remote procedure call takes. Also implementing full tracing support in all your services takes a lot more effort.

@jrm780 [noted](https://github.com/containous/traefik/pull/2290#issuecomment-422573460):

> `X-Request-Id` is a non-standard but **very common header**, and it's incredibly useful in correlating log entries. **Distributed tracing aims to solve entirely different problems**, so it would seem odd for OpenTracing support and adding an X-Request-Id header to be mutually exclusive.

@aaslamin [noted that this is a feature included in several other Traefik alternatives](https://github.com/containous/traefik/pull/2290#issuecomment-426464339):

> +100 to this feature.
> 
> I was quite surprised to see this issue closed because of opentracing. It is very common for edge > proxies to generate a unique ID for each request:
> 
> • http://nginx.org/en/docs/http/ngx_http_core_module.html#var_request_id
> • https://www.envoyproxy.io/docs/envoy/latest/configuration/http_conn_man/headers#config-http-conn-man-headers-x-request-id
> • https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#4-unique-id-format
> 
> One can use this to hydrate their access logs to create request scoped loggers for example as a request trickles through the service mesh. 

Finally, tracing (more expensive while processing each request, requires additional services, usually only done on a small percentage of requests during production) and simply adding a request id (inexpensive, no additional services required, can be enabled on all requests in production) are **not the same thing** and does not have the same use-case.  The reply that OpenTracing should be used instead is not a valid argument for not including this feature.

Lastly, I have kept this **as simple as possible**, so maintenance and adverse effects should be minimal.  It expands on the existing Custom Header functionality, and can possibly be expanded on in the future to include other dynamic values (as suggested: #5036, #5063, #4519).

### Simple implementation example

```toml
[http.middlewares]
  [http.middlewares.requestIdHeader.headers]
    [http.middlewares.requestIdHeader.headers.customRequestHeaders]
        X-Request-ID = "{uuid}"

[accessLog]
  filePath = "/path/to/access.log"
  format = "json"
  [accessLog.fields]
    [accessLog.fields.headers]
      defaultMode = "keep"
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,6 +42,7 @@ require (
 	github.com/gogo/protobuf v1.3.0 // indirect
 	github.com/golang/protobuf v1.3.4
 	github.com/google/go-github/v28 v28.1.1
+	github.com/google/uuid v1.1.1
 	github.com/googleapis/gnostic v0.1.0 // indirect
 	github.com/gorilla/mux v1.7.3
 	github.com/gorilla/websocket v1.4.1
--- a/pkg/middlewares/headers/headers.go
+++ b/pkg/middlewares/headers/headers.go
@@ -173,7 +173,7 @@ func (s *Header) modifyCustomRequestHeaders(req *http.Request) {
 		case strings.EqualFold(header, "Host"):
 			req.Host = value
 
-		case strings.EqualFold(value, "{{uuid}}"):
+		case strings.EqualFold(value, "{uuid}"):
 			if v4, err := uuid.NewRandom(); err == nil {
 				req.Header.Set(header, v4.String())
 			}
--- a/pkg/middlewares/headers/headers.go
+++ b/pkg/middlewares/headers/headers.go
@@ -12,6 +12,7 @@ import (
 	"github.com/containous/traefik/v2/pkg/log"
 	"github.com/containous/traefik/v2/pkg/middlewares"
 	"github.com/containous/traefik/v2/pkg/tracing"
+	"github.com/google/uuid"
 	"github.com/opentracing/opentracing-go/ext"
 	"github.com/unrolled/secure"
 )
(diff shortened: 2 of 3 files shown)


ldez:
Hello,

We will close for the same reasons as the previous PR on this topic.

https://github.com/containous/traefik/pull/2290#issuecomment-350268589

Thank you anyway.

Sarke:
It is not the same thing!

rafipiccolo:
I was searching for this functionality too. Which i had in apache.
This PR looks clean and easy to integrate. Why can't we have this feature merged ? code debt ? not enough people interested ? (i doubt)

Can we have an alternative ? You suggest Opentracing ?
i think it would be very expensive to trace all request just to have request ids for all services.
Am I wrong ?

Sarke:
@rafipiccolo 

```toml
[tracing]
  [tracing.jaeger]
    samplingParam = 0
    traceContextHeaderName = "uber-trace-id"
```

```yml
TRAEFIK_TRACING: "true"
TRAEFIK_TRACING_JAEGER_TRACECONTEXTHEADERNAME: "X-Request-ID"
```

PR #6713, opened 2020-04-21
ratelimit: do not default to ipstrategy too early

### LINKED ISSUE #6687: RateLimit middleware always uses ipStrategy sourceCriterion

Bug

### What did you do?
Configure the RateLimit middleware for an endpoint using the `requestHeaderName` sourceCriterion.

Requests fall into different buckets for different header values of the header key defined in `sourceCriterion.requestHeaderName`

### What did you see instead?
The `sourceCriterion.requestHeaderName` is ignored, and the defined header has no effect. Even though the middleware appears with the appropriate configuration on the Traefik dashboard, the logs confirm it is using the default sourceCriterion (`ipStrategy`) instead.

```
Traefik version 2.2.0 built on 2020-03-25T17:32:57Z
```

```
Docker labels
- "traefik.enable=true"
- "traefik.http.routers.whoami.rule=Host(`localhost`)"
- "traefik.http.routers.whoami.entrypoints=web"
- "traefik.http.middlewares.test-ratelimit.ratelimit.average=2"
- "traefik.http.middlewares.test-ratelimit.ratelimit.burst=5"
- "traefik.http.middlewares.test-ratelimit.ratelimit.period=1m"
- "traefik.http.routers.whoami.middlewares=test-ratelimit@docker"
- "traefik.http.middlewares.test-ratelimit.ratelimit.sourcecriterion.requestheadername=username"
```

```
traefik    | time="2020-04-16T12:35:05Z" level=debug msg="Using IPStrategy" middlewareType=RateLimiterType middlewareName=test-ratelimit@docker entryPointName=web routerName=whoami@docker
```

The condition of the if clause in [extractor.go](https://github.com/containous/traefik/blob/v2.2/pkg/middlewares/extractor.go#L26) is always evaluated as `true` due to the [defaults](https://github.com/containous/traefik/blob/v2.2/pkg/config/dynamic/middlewares.go#L341) of this middleware.

At first I was going to suggest changing the order of the `if`'s in [extractor.go](https://github.com/containous/traefik/blob/v2.2/pkg/middlewares/extractor.go#L26) to place `if sourceMatcher.IPStrategy != nil` in the last position, since that is the default, but I think this would break the `InFlightReq` middleware, which has `requestHost` as its default.

mpl:
### What does this PR do?

This PR removes ipstrategy being systematically set as the default source
criterion, which was causing any other source criterion provided in the config
to be ignored.

We already take care of setting ipstrategy as the default in the rate limiter
constructor anyway.

### Motivation

Fixes #6687

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/dynamic/middlewares.go
+++ b/pkg/config/dynamic/middlewares.go
@@ -337,9 +337,6 @@ type RateLimit struct {
 func (r *RateLimit) SetDefaults() {
 	r.Burst = 1
 	r.Period = types.Duration(time.Second)
-	r.SourceCriterion = &SourceCriterion{
-		IPStrategy: &IPStrategy{},
-	}
 }
 
 // +k8s:deepcopy-gen=true
--- a/pkg/config/dynamic/middlewares.go
+++ b/pkg/config/dynamic/middlewares.go
@@ -286,13 +286,6 @@ type InFlightReq struct {
 	SourceCriterion *SourceCriterion `json:"sourceCriterion,omitempty" toml:"sourceCriterion,omitempty" yaml:"sourceCriterion,omitempty"`
 }
 
-// SetDefaults Default values for a InFlightReq.
-func (i *InFlightReq) SetDefaults() {
-	i.SourceCriterion = &SourceCriterion{
-		RequestHost: true,
-	}
-}
-
 // +k8s:deepcopy-gen=true
 
 // PassTLSClientCert holds the TLS client cert headers configuration.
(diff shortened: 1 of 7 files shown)


mpl:
adding no merge because current test does not actually prevent regression for now.

mpl:
> adding no merge because current test does not actually prevent regression for now.

after consideration, I don't think the regression should be prevented in the unit tests of the ratelimiter package, since SetDefaults is applied out of this package.

However, I'm wondering if we should enforce that the fields of SourceCriterion are all mutually exclusive (they already actually are), and in which case we could add a test reflecting that property.

PR #6714, opened 2020-04-21
Hash sticky cookie value in secure mode

sylr:
### What does this PR do?

Hash sticky cookie value in secure mode.

### Motivation

Service URL in cookie value is not a proper way to do things.

### More

- [x] Added/updated tests


sylr:
I'm closing this because it does not work, at all ...

PR #6719, opened 2020-04-22
Add sentence about the resource namespace and middleware

### LINKED ISSUE #6318: [Documentation] Adding precisely how to reference Kubernetes-CRD Middleware from other providers

Feature

A Kubernetes-crd Middleware lives in kubernetes, and as such, has a kubernetes name (for instance `fooName`) and a kubernetes namespace (for instance `barNS`))
If one needs to configure traefik via another provider a routing using that middleware (for instance kubernetes Ingress middleware with the current master branch (commit 7c430e5c9de5305b2f58dce77cdf50eeba944a25 in my case) ), one needs to give the correct traefik name of the namespace. After some testing and a view of the dashboard, it appears the correct name is `barNS-fooName@kubernetescrd`

As this might be unclear for people getting familiar, I suggest the rule ( `<K8SNamespace>-<K8SName>@kubernetescrd`) be written in the documentation.

Furthermore, and I'll probably create some bug reports on that later, this naming rule can generate collision, for instance if one has a namespace `foo-bar` containing a middleware `baz` and a namespace `foo` containing a middleware `bar-baz`. This might require to change to traefik naming rule for kubernetes-crd, possibly before getting that rule written down in the documentation. I suggest using a `.` as separator, as this character is forbidden in kubernetes objet names.

SantoDE:
### What does this PR do?

Adds a sentence about the fact, that the namespace of a middleware is internally added automatically, as it's important when you want to reference that middleware from another provider.

### Motivation

Enhance our Docs

Fix #6318

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/routing/providers/kubernetes-crd.md
+++ b/docs/content/routing/providers/kubernetes-crd.md
@@ -623,7 +623,7 @@ Register the `Middleware` [kind](../../reference/dynamic-configuration/kubernete
 	when the definition of the middleware comes from another provider.
 	In this context, specifying a namespace when referring to the resource does not make any sense, and will be ignored.
 	Additionally, when you want to reference a Middleware from the CRD Provider, you have to append the namespace of the
-    Ressource in the ressource-name as Traefik appends the namespace internally automatically.
+    resource in the ressource-name as Traefik appends the namespace internally automatically.
 
 More information about available middlewares in the dedicated [middlewares section](../../middlewares/overview.md).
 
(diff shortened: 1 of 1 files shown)


docs/content/routing/providers/kubernetes-crd.md
```diff
@@ -622,6 +622,8 @@ Register the `Middleware` [kind](../../reference/dynamic-configuration/kubernete
 	(in the reference to the middleware) with the [provider namespace](../../middlewares/overview.md#provider-namespace),
 	when the definition of the middleware comes from another provider.
 	In this context, specifying a namespace when referring to the resource does not make any sense, and will be ignored.
+	Additionally, when you want to reference a Middleware from the CRD Provider, you have to append the namespace of the
+	Ressource in the ressource-name as Traefik appends the namespace internally automatically.
```
mpl, line 626:
```suggestion
	Resource in the resource-name as Traefik appends the namespace internally automatically.
```

SantoDE:
Should be done @notsureifkevin

Issue #5332, opened 2019-09-09
Traefik answer 404 error even with a catch all rule when backend is missing/unhealthy


felicienfrancois:
### Do you want to request a *feature* or report a *bug*?

Bug

### Did you try using a 1.7.x configuration for the version 2.0?

- [ ] Yes
- [ X] No

### What did you do?

With Docker provider, I have a catch-all rule defined as follow on a static nginx service:
```
        - traefik.http.routers.errors.priority=1
        - traefik.http.routers.errors.rule=HostRegexp(`{catchall:.*}`)
```

I successfully get it serve any url not catched by other services (ex: not-catched.mydomain.com).

BUT, when restarting other services (or in any case where no container is in healthy state): *I get the default 404 error page from traefik*.

### What did you expect to see?

I expect to get a 502 or 503 error indicating that the backend is not reachable. I would also like to be able to customize this error page (ex: through middleware)
BUT according to #1689, the current status is that frontend is not kept when backend is down. If true, I should have the catch-all rule working (so have my custom error page).

### What did you see instead?
The traefik default 404 page
```
404 page not found
```

It means that the catch-all workaround to have a custom error page does not work for down backends.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      2.0.0-rc2
Codename:     montdor
Go version:   go1.13rc2
Built:        2019-09-03T20:10:11Z
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
```

Provider: docker

### If applicable, please paste the log output in DEBUG level (`--log.level=DEBUG` switch)

```
```

PR #6739, opened 2020-04-28
do not throw away route if service behind it is misconfigured/down

breunigs:
### What does this PR do?

Previously Traefik would ignore all rules where there were some configuration issues with the 
service it referenced, e.g. scaled to 0, service missing or mistyped, incorrect port configuration, all pods down, etc. This patch changes the behaviour to log an error instead, but keep the route otherwise. In practice, this means Traefik will serve a 503 instead of using the next lower prioritized route. 

### Motivation

If a service is down and the pre-patch behaviour of "route fallback" is in place, the following issues
appear:
- pods suddenly get requests they're not expecting. Depending on their scaling, this might lead to cascading failures.
- rendering e.g. the startpage instead of the expected page (or an error) is confusing to humans. Search engines might delist pages if they receive a 404 instead of a 503.
- using a high priority route to block traffic to sensitive paths (e.g. `/admin` on an external ingress) depends on the error pages service being up (defense in depth approach, these routes should not be configured in the first place of course)

### More

- [x] Added/updated tests

### Additional Notes

The same bug was fixed for the Kubernetes provider, but reintroduced/not fixed from the get-go in the Kubernetes CRD one.

Related tickets:
https://github.com/containous/traefik/issues/1689
https://github.com/containous/traefik/issues/5332

Since the integration CI is not passing for me locally on master, I only ran a subset of it. I will look at the CI failures and fix them, if needed.

I kept the behaviour of Traefik accepting a route with an empty list of services, but added a log for it and also added some tests to ensure this is not inadvertently changed. It's easy to make this a "configuration error" by simply adding a `continue` and adjusting the test there. If that's the desired approach, using OpenAPI schema validation on the CRDs would prevent this edge case from appearing in the first place, though. I use the redirect middleware for the tests here so I can get a different status code. The benefit is that it keeps the test focused on Traefik. Otherwise the `--disable-agent` flag on `k3s` would have to be removed to get networking support, so that a `whoami` pod would actually be reachable… but that increases the covered area a lot, without improving the test.

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes_http.go
+++ b/pkg/provider/kubernetes/crd/kubernetes_http.go
@@ -91,19 +91,11 @@ func (p *Provider) loadIngressRouteConfiguration(ctx context.Context, client Cli
 			normalized := provider.Normalize(makeID(ingressRoute.Namespace, serviceKey))
 			serviceName := normalized
 
-			if len(route.Services) > 1 {
-				spec := v1alpha1.ServiceSpec{
-					Weighted: &v1alpha1.WeightedRoundRobin{
-						Services: route.Services,
-					},
-				}
-
-				errBuild := cb.buildServicesLB(ctx, ingressRoute.Namespace, spec, serviceName, conf.Services)
-				if errBuild != nil {
-					logger.Error(errBuild)
-					continue
-				}
-			} else if len(route.Services) == 1 {
+			switch len(route.Services) {
+			case 0:
+				logger.Error("Empty service list")
+				continue
+			case 1:
 				fullName, serversLB, err := cb.nameAndService(ctx, ingressRoute.Namespace, route.Services[0].LoadBalancerSpec)
 				if err != nil {
 					logger.Error(err)
@@ -115,8 +107,18 @@ func (p *Provider) loadIngressRouteConfiguration(ctx context.Context, client Cli
 				} else {
 					serviceName = fullName
 				}
-			} else {
-				logger.Info("Empty service list")
+			default:
+				spec := v1alpha1.ServiceSpec{
+					Weighted: &v1alpha1.WeightedRoundRobin{
+						Services: route.Services,
+					},
+				}
+
+				errBuild := cb.buildServicesLB(ctx, ingressRoute.Namespace, spec, serviceName, conf.Services)
+				if errBuild != nil {
+					logger.Error(errBuild)
+					continue
+				}
 			}
 
 			conf.Routers[normalized] = &dynamic.Router{
--- a/pkg/provider/kubernetes/crd/kubernetes_http.go
+++ b/pkg/provider/kubernetes/crd/kubernetes_http.go
@@ -115,6 +115,8 @@ func (p *Provider) loadIngressRouteConfiguration(ctx context.Context, client Cli
 				} else {
 					serviceName = fullName
 				}
+			} else {
+				logger.Info("Empty service list")
 			}
 
 			conf.Routers[normalized] = &dynamic.Router{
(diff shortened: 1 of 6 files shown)


sylr:
I personally would really like this to stay as it is.

The fact that traefik continues to match ingressroutes if the first one that match has no backend makes it possible to have a catch all service which can display custom error pages.

breunigs:
Displaying custom error pages is definitely a needed feature, which Traefik implements through this middleware: https://docs.traefik.io/middlewares/errorpages/

As is, the error middleware doesn’t kick in, or at least not with the right status code, since you lose the information that there was another backend at all. Showing a 503 to a search engine will make it revisit your page later, a 404 will get it delisted. Alternatively, if a backend grabs /something and it is down, it will now render your startpage, which presumably grabs /. This is confusing for humans as well.

I really think the error middleware is the better choice here instead of relying on this bug.

breunigs:
Looking into further edge cases, I found that a mistyped (= non existent) service name will make Traefik ignore the route, too. Somewhat arbitrarily, if a route has no services defined (i.e. `services: []`) Traefik will not report this as an error and accept the route. I'm not quite sure how Traefik deals with this case; something is weird in my local setup here.

In any case, I don't think this PR makes sense to merge as is, and I have a WIP patch that covers the additional cases. Still need to figure out the "no service defined" case though.

breunigs:
okay, I figured out the issues I mentioned in the previous post. During the tests `k3s` is started with `--disable-agent`, so it doesn't setup networking. I use the redirect middleware to get a different status code to assert the behaviour, which doesn't depend on the backend to be up.

pkg/provider/kubernetes/crd/kubernetes_http.go
```diff
@@ -332,12 +340,12 @@ func (c configBuilder) loadServers(fallbackNamespace string, svc v1alpha1.LoadBa
 		}
 
 		if port == 0 {
-			return nil, fmt.Errorf("cannot define a port for %s/%s", namespace, sanitizedName)
+			return nil, false, fmt.Errorf("cannot define a port for %s/%s", namespace, sanitizedName)
 		}
 
 		protocol, err := parseServiceProtocol(svc.Scheme, svcPort.Name, svcPort.Port)
```
hakkibagci, line 346: I think, in the first place, this should not have been inside this for loop at all. Not your fault of course. Should we move it outside of the for loop? Maybe you also want to update your PR to the upstream.

rdxmb:
> Somewhat arbitrarily, if a route has no services defined (i.e. services: []) Traefik will not report this as an error and accept the route. 

I have a definition like

```
kind: IngressRoute
metadata:
  name: http
spec:
  entryPoints:
    - internal-http
  routes:
  - kind: Rule
    match: "Host(`test.example.com`)" 
    middlewares:
      - name: redirect-to-https-perm
        namespace: traefik-2
    services: []
```

and the dashboard says

`ERRORS: the service "MY_NAMESPACE-http-994bcc99bd0352120a65@kubernetescrd" does not exist `

Also, the redirect does not work with `services: []`

jbdoumenjou:
Hi @breunings,

It is very difficult to infer what would be the correct configuration and various reasons may appear in the different providers.

The actual mechanism is oriented to be strongly dynamic and stateless, and the configuration is rebuilt at each update.

Currently, you can achieve your use case by adding a fallback router to serve an error page (or whatever you need) through a custom service.
If you want to have more details on how to configure Traefik to do so, please go to the [community forum](https://community.containo.us/). 

So we will close the PR accordingly.
Anyway, Thank you for your interest in Traefik.

Issue #4674, opened 2019-03-27
Add a provider for ECS in Traefik v2


jbdoumenjou:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Traefik v1 supports dynamic config generation for and from a ECS cluster, i.e. it has an ECS provider.

Now that the main changes for v2 are out, we can re-add previously supported providers, such as the one for ECS.

The existing providers, such as docker, as well as the code for the ECS provider in v1.7, can be used for inspiration.

m13t:
It’s a real shame that ECS Support was dropped as part of V2, but some were brought forward. We have quite a lot of V1 instances running which we were looking forward to bringing up to V2.

SantoDE:
It's not really "dropped", just not done yet. If you like, take a shot ;)

m13t:
Okay that’s good to know. I thought given v2 is now released and it didn’t make the cut, it must have been dropped.

JavierPandu:
How about using a "resolver" like in nginx?

ysz:
@SantoDE can I take a shot? 🎄

ldez:
@ysz Traefik is an open source project, so feel free to work on this.

andrewelkins:
Has anyone made progress on creating support for ECS in v2?

a6patch:
> It's not really "dropped", just not done yet. If you like, take a shot ;)

Isn't this sort of a lie?  ECS is supported in the 2.x EE version, which means it HAS been done already.  While I think that adding features to an open source project and bundling them as paid premium tier is perfectly ok, REMOVING a feature that already existed as you go from version 1.7 to 2.x is a violation of the spirit of open source and the people that freely contributed to getting the feature into V1.7.  

Will ECS ever be added to the Community edition of 2.x?

SantoDE:
not sure what you talking about @a6patch or where you get that information from. EE 2.X *does NOT* have ECS as a provider. If its listed somewhere, we'll need to correct it as its untrue.

For ECS in Traefik 2.X, there is the PR #6749 which I'm currently testing on ECS :-)

a6patch:
Sorry if I misunderstood.  After waiting a number of months after 2.0 for ECS to get added, I had a conversation with a salesperson that very specifically said that that was an EE feature and wasn't available in the non-EE 2.x Traefik. She must have been mistaken.  

[...]

PR #6749, opened 2020-04-29
Add AWS ECS provider

alekitto:
### What does this PR do?

Added ECS dynamic provider in v2 (#4674)

### More

- [x] Added/updated tests

### Additional Notes

I used the v1 version of ECS provider as a base and adapted it based on v2 docker provider.


docs/content/providers/ecs.md
```diff
@@ -0,0 +1,193 @@
...
+```
+
+```bash tab="CLI"
+--providers.ecs.region="us-east-1"
+--providers.ecs.accessKeyID="abc"
+--providers.ecs.secretAccessKey="123"
+# ...
+```
+
+If accessKeyID/secretAccessKey is not given credentials will be resolved in the following order:
```
SantoDE, line 189: given = provided maybe?

rtribotte: ok 👍

alekitto: Ok, I'll change it ASAP

alekitto:
One small thing I noticed yesterday: the AutoDiscoverClusters setting is true by default. I think that should be false by default if the `clusters` list is not empty.

What do you think about it?

PR #6762, opened 2020-05-03
Improve acme CLI options in Let's Encrypt documentation 

netoax:
### What does this PR do?

When reading the current documentation I noticed the Let's Encrypt section shows the acme CLI flags in camelCase format instead of lowercase, which is different from other sections of documentation such as the CLI configuration [reference](https://docs.traefik.io/reference/static-configuration/cli/). Because of that, this PR updates the flags to keep homogeneous documentation. Example:

From: `--certificatesResolvers.<name>.acme.caServer`
To: `--certificatesresolvers.<name>.acme.caserver`

### Motivation

My main motivation is to improve the documentation in a way that avoids newcomers get confused.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -105,13 +105,13 @@ Please check the [configuration examples below](#configuration-examples) for mor
     ```
     
     ```bash tab="CLI"
-    --entryPoints.web.address=:80
-    --entryPoints.websecure.address=:443
+    --entrypoints.web.address=:80
+    --entrypoints.websecure.address=:443
     # ...
-    --certificatesResolvers.myresolver.acme.email=your-email@example.com
-    --certificatesResolvers.myresolver.acme.storage=acme.json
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

Thank you,

The CLI flags are case insensitive, so camelCase or lowercase produce the same result.

The trailing spaces are not errors but used to handle section.

netoax:
Hello @ldez, thanks for the feedback. I didn't know about the trailing spaces being used to control section, sorry.

Is this still valuable changing the flags to keep standardization across the docs?

ldez:
Have an homogeneous because is still a valuable change.

Could you restore the trailing space, thank you.

netoax:
> Have an homogeneous documentation is still a valuable change.
> 
> Could you restore the trailing space, thank you.
> 
> Could you also edit the PR description.

Done.

PR #6763, opened 2020-05-03
Use specified network for "container" network mode

bjeanes:
### What does this PR do?

When defining a service in the docker provider which is pointing to a container with a networking mode of `host:other-container`, the `traefik.docker.network` was ignored and an arbitrary IP of the `other-container` was used as the upstream.

If that IP is not routable from the Traefik instance, this could cause Gateway Timeouts.

This change adds steps to parse configuration from labels on the connected container from which the service container inherits its network.

My strawman fix prioritises the `traefik.docker.network` label on the network-owning container, but falls back to the network specified on the service container. Perhaps this should be flipped, but the network-owning container is the authoritative source of which networks it is connected to, so this makes sense to me.

### Motivation

I have a configuration where several containers with a `network_mode` of `container:<x>` are services served by a Traefik instance. The container `<x>` has exactly 1 shared network with Traefik, but Traefik is not reliably selecting that interface.

<details>
<summary>More detail and replication case with as <code>docker-compose.yml</code> here</summary>

Without this change, the `docker-compose.yml` below creates a situation where (sometimes, may not be deterministic) Traefik will use the IP address from the `private` network in its service definition.

<details><summary><code>docker-compose.yml</code></summary>

``` yml
version: '3.5'
networks:
  public:
  private:
  traefik:
    internal: true

services:
  traefik:
    image: traefik:v2.2
    container_name: traefik
    # bug present even with --providers.docker.network=traefik_traefik
    command: --api.insecure=true --providers.docker=true --providers.docker.exposedByDefault=false --providers.docker.network=traefik_traefik
    volumes:
      - "/var/run/docker.sock:/var/run/docker.sock"
    ports:
      - "80:80"
      - "8080:8080"
    networks:
      - public
      - traefik

  out:
    image: "wardsco/sleep"
    networks:
      - private
      - traefik
    labels:
      - "traefik.docker.network=traefik_traefik"

  one:
    image: xperimental/goecho:v1.7
    container_name: one
    command: -addr :8003
    network_mode: service:out
    labels:
      - "traefik.enable=true"
      - "traefik.http.services.one.loadbalancer.server.port=8001"

  two:
    image: xperimental/goecho:v1.7
    container_name: two
    command: -addr :8002
    network_mode: service:out
    labels:
      - "traefik.enable=true"
      - "traefik.http.services.two.loadbalancer.server.port=8002"
```

</details>

It will instead print:

```
traefik    | time="2020-05-03T05:51:27Z" level=warning msg="Could not find network named 'traefik_traefik' for container '/one'! Maybe you're missing the project's prefix in the label? Defaulting to first available network." providerName=docker container=one-traefik-cd4e924fe8fc508351a3a548d95ec180429a2a746a05143e11c1f9c2cc9ac6a1 serviceName=one
traefik    | time="2020-05-03T05:51:27Z" level=warning msg="Could not find network named 'traefik_traefik' for container '/two'! Maybe you're missing the project's prefix in the label? Defaulting to first available network." serviceName=two providerName=docker container=two-traefik-97929de8aeb9fdb4a535036267ff57e7006476d9d285811660849d3011c45ccd
```

As can be seen from the Traefik dashboard listing of services and the `docker inspect` dump, it is using the `private` network (`172.30.0.2`) instead of the `traefik` network (`172.31.0.2`)

image

``` jsonc
// $ docker inspect traefik_out_1
// (many fields elided for brevity)
[
    {
        "Id": "04966273dcd0f0bc0f508b03f7b9d88d8c083a4509a636cb36c9a9487c94d1a8",
        "Name": "/traefik_out_1",
        "HostConfig": {
           "NetworkMode": "traefik_private",
        },
        "Mounts": [],
        "Config": {
            "Hostname": "04966273dcd0",
            "Labels": {
                "traefik.docker.network": "traefik_traefik"
            }
        },
        "NetworkSettings": {
            "Bridge": "",
            "Networks": {
                "traefik_private": {
                    "IPAddress": "172.30.0.2",
                },
                "traefik_traefik": {
                    "IPAddress": "172.31.0.2",
                }
            }
        }
    }
]
```

</details>

### More

- [ ] Added/updated tests

### Additional Notes

NOTE: This bug is also present when `--providers.docker.network` is set, which was also surprising to me. It is unclear to me at this time whether this patch also fixes that (but I don't think it does), as I don't understand how this functions yet.

<details>
<summary>~There are two pre-existing failing unit tests~ #6768</summary>

These _appear_ to have been caused by 337273b5df198125a1a73fd7d26c1051eb5a42cb but as far as I can see Semaphore does not run the unit tests, or perhaps this only fails when run on devices which have a resolveable `host.docker.internal`.

```
--- FAIL: TestDockerGetIPAddress (0.00s)
    --- FAIL: TestDockerGetIPAddress/no_network,_no_network_label,_mode_host (0.00s)
        config_test.go:3256:
                Error Trace:    config_test.go:3256
                Error:          Not equal:
                                expected: "127.0.0.1"
                                actual  : "10.144.231.49"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -127.0.0.1
                                +10.144.231.49
                Test:           TestDockerGetIPAddress/no_network,_no_network_label,_mode_host
    --- FAIL: TestDockerGetIPAddress/two_networks,_no_network_label,_mode_host (0.00s)
        config_test.go:3256:
                Error Trace:    config_test.go:3256
                Error:          Not equal:
                                expected: "127.0.0.1"
                                actual  : "10.144.231.49"

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1 +1 @@
                                -127.0.0.1
                                +10.144.231.49
                Test:           TestDockerGetIPAddress/two_networks,_no_network_label,_mode_host
```
</details>

I do not yet know the best way to test this. I haven't looked deeply at integration tests yet, but it appears unit tests will be too difficult due to the interaction with the `dockerClient`. Guidance welcome.

cc @ldez

### DIFF
--- a/pkg/provider/docker/config.go
+++ b/pkg/provider/docker/config.go
@@ -349,7 +349,13 @@ func (p Provider) getIPAddress(ctx context.Context, container dockerData) string
 		// Check connected container for traefik.docker.network, falling back to
 		// the network specified on the current container.
 		containerParsed := parseContainer(containerInspected)
-		extraConf, _ := p.getConfiguration(containerParsed)
+		extraConf, err := p.getConfiguration(containerParsed)
+
+		if err != nil {
+			logger.Warnf("Unable to get IP address for container %s : failed to get extra configuration for container %s: %s", container.Name, containerInspected.Name, err)
+			return ""
+		}
+
 		if extraConf.Docker.Network == "" {
 			extraConf.Docker.Network = container.ExtraConf.Docker.Network
 		}
--- a/pkg/provider/docker/config.go
+++ b/pkg/provider/docker/config.go
@@ -345,7 +345,17 @@ func (p Provider) getIPAddress(ctx context.Context, container dockerData) string
 			logger.Warnf("Unable to get IP address for container %s : Failed to inspect container ID %s, error: %s", container.Name, connectedContainer, err)
 			return ""
 		}
-		return p.getIPAddress(ctx, parseContainer(containerInspected))
+
+		// Check connected container for traefik.docker.network, falling back to
+		// the network specified on the current container.
+		containerParsed := parseContainer(containerInspected)
+		extraConf, _ := p.getConfiguration(containerParsed)
+		if extraConf.Docker.Network == "" {
+			extraConf.Docker.Network = container.ExtraConf.Docker.Network
+		}
+
+		containerParsed.ExtraConf = extraConf
+		return p.getIPAddress(ctx, containerParsed)
 	}
 
 	for _, network := range container.NetworkSettings.Networks {


pkg/provider/docker/config.go
```diff
@@ -345,7 +345,17 @@ func (p Provider) getIPAddress(ctx context.Context, container dockerData) string
 			logger.Warnf("Unable to get IP address for container %s : Failed to inspect container ID %s, error: %s", container.Name, connectedContainer, err)
 			return ""
 		}
-		return p.getIPAddress(ctx, parseContainer(containerInspected))
+
+		// Check connected container for traefik.docker.network, falling back to
+		// the network specified on the current container.
+		containerParsed := parseContainer(containerInspected)
+		extraConf, _ := p.getConfiguration(containerParsed)
```
bjeanes, line 352: I chose not to handle the error here, because this check previously didn't happen at all, but if there's a log line worth outputting here, let me know how you'd like that handled.

mpl:
> There are two pre-existing failing unit tests
> These _appear_ to have been caused by [337273b](https://github.com/containous/traefik/commit/337273b5df198125a1a73fd7d26c1051eb5a42cb) but as far as I can see Semaphore does not run the unit tests, or perhaps this only fails when run on devices which have a resolveable `host.docker.internal`.

Yes, that is exactly the case. See https://github.com/containous/traefik/pull/6768

> ```
> --- FAIL: TestDockerGetIPAddress (0.00s)
>     --- FAIL: TestDockerGetIPAddress/no_network,_no_network_label,_mode_host (0.00s)
>         config_test.go:3256:
>                 Error Trace:    config_test.go:3256
>                 Error:          Not equal:
>                                 expected: "127.0.0.1"
>                                 actual  : "10.144.231.49"
> 
>                                 Diff:
>                                 --- Expected
>                                 +++ Actual
>                                 @@ -1 +1 @@
>                                 -127.0.0.1
>                                 +10.144.231.49
>                 Test:           TestDockerGetIPAddress/no_network,_no_network_label,_mode_host
>     --- FAIL: TestDockerGetIPAddress/two_networks,_no_network_label,_mode_host (0.00s)
>         config_test.go:3256:
>                 Error Trace:    config_test.go:3256
>                 Error:          Not equal:
>                                 expected: "127.0.0.1"
>                                 actual  : "10.144.231.49"
> 
>                                 Diff:
>                                 --- Expected
>                                 +++ Actual
>                                 @@ -1 +1 @@
>                                 -127.0.0.1
>                                 +10.144.231.49
>                 Test:           TestDockerGetIPAddress/two_networks,_no_network_label,_mode_host
> ```

pkg/provider/docker/config.go
mpl, line 352: I'm not sure yet that you/we have got the right idea here (still thinking), but in any case, I think that line can instead just be:
```
extraConf := configuration{
	Docker: specificConfiguration{
		Network: p.Network,
	},
}
```
right?

pkg/provider/docker/config.go
```diff
@@ -345,7 +345,17 @@ func (p Provider) getIPAddress(ctx context.Context, container dockerData) string
...
+		if extraConf.Docker.Network == "" {
+			extraConf.Docker.Network = container.ExtraConf.Docker.Network
+		}
```
mpl, line 355: I don't think this does anything useful. Whatever container you're looking at, I think its ExtraConf.Docker.Network always comes from p.Network anyway, so it's the same value everywhere.
Or am I missing anything?

mpl: ah nevermind, my bad, I see now that it can be overridden by the specific label on the container.

pkg/provider/docker/config.go
mpl: nevermind that comment too, we indeed need the full method in case a label overrides the value from p.Network.

mpl: yeah, let's handle it like in the rest of the code please:

```
		if err != nil {
			logger.Warnf("Unable to get IP address for container %s : failed to get extra configuration for container %s: %s", container.Name, inspected.Name, err)
			return ""
		}
```

pkg/provider/docker/config.go
bjeanes: Yeah I'm not sure that's really what we want. I initially had labelled the service container expecting it to work, but in the course of doing this PR, it seemed to make more sense to label the container which owned the network stack.

What do you think? Is my approach of checking both reasonable enough?

pkg/provider/docker/config.go
bjeanes: @mpl Hmm... It's different to the warning on L345, where we have failed to get any information about that connected container:

https://github.com/bjeanes/traefik/blob/container-networking-use-specified-container/pkg/provider/docker/config.go#L343-L347

At the site where I skipped error handling, the connected container has been fetched and identified, but parsing the extra configuration might have failed. Essentially, we can see the container, but we may not be able to parse any `traefik.docker.*` labels on it.

Is an early return the right direction? It seems extreme to me, because we may still get an IP address from that container, it just might not be the one corresponding to the specified network. There is already a warning that will be raised in that case though. Essentially, it would be the existing behaviour prior to this patch.

A warning here on L352 definitely seems pragmatic, but the code snippet you've added in your comment doesn't seem like a home run to me.

I could be missing something here, as I'm less familiar with this code, of course... 

Let me know if I've got this wrong, and I'll make the change!

bjeanes:
@mpl @ldez I'm worried the `contributor/waiting-for-corrections` tag means my clarification about the corrections is being missed. Can you please take a look [at my question](https://github.com/containous/traefik/pull/6763#discussion_r421139276)

pkg/provider/docker/config.go
mpl: What would be the point of getting an IP address which has little to do with how the user intended to configure things? it seems to me that it might give them a fake sense of comfort, and might make them miss the fact that they mis-configured something. If we fail hard and return no IP address, there's a better chance the misconfiguration will get noticed.

That is, unless you see a case where p.getConfiguration would fail for another reason than misconfiguration, and in which case, it might make sense to try and cope with it.

bjeanes: @mpl it's a breaking change, is all.

bjeanes: As far as I can tell, though, this is the behaviour in Traefik for a more conventional setup. When the specified network can't be found, it assigns a non-deterministic IP from one of the networks on the container.

https://github.com/containous/traefik/blob/6b1158235e145635d79d251ac1d8a3865258e329/pkg/provider/docker/config.go#L313-L323

In that case, if an IP is found from that network, it returns early with the IP. If it is _not_ found, it warns and continues.

This PR is just delegating that same logic to the container which owns the network stack. This section of my PR simply makes sure the referenced container struct has the necessary metadata so that the existing logic can do the right thing.

If we want to fail when the specified network is non-existent, shouldn't it be done in the code I just referenced, so that it is consistent across all setups?

@mpl If you still feel, given the above, that this should be treated differently and fail early, then just confirm again and I will happily do as advised. Cheers.

bjeanes: Bump!

mpl: I understand what you're saying, and I understand the argument of consistency (i.e. if at one point we fail to find the network, then we should always leave it to the final loop to pick an address at random in one of our networks).

However, if we wanted to be really consistent everywhere, then I see no reason why we should keep the returns at L339 and L346. These errors too could be considered as non-critical, and we would also go YOLO on them and let the final for loop pick an address in that case. And same for L332 too actually. Right?

And I think the crux of it really is that when we're in IsHost or IsContainer mode, we try to respect the user's intent when they configured that. We assume we knew what they wanted and they just made a mistake while doing it, so we're not sneakily picking something else at random against their intent.

And so that's why it might be ok end up in the final loop at L361 only if we are in the error case of L321, even though it does not look consistent at first sight. I think we should either go that way, or go "full consistent" everywhere else and change all the other returns as well (L332, L339, L346) but I'm very wary of that last proposition, as it would probably break a number of things.

bjeanes: Thanks for the response and clarity on your position. I'll address the feedback ASAP

bjeanes: Sorry, I re-read that and realised it _could_ come off as passive aggressive. I hear you, and that logic makes total sense so I am very onboard. I'll try to fix this up today.

bjeanes:
@mpl what's your preference with regards to the branch being out of date now? Do you want me to rebase, merge master, or let you handle conflict upon (a hopeful) merge?

pkg/provider/docker/config.go
mpl: Np, you were the right amount of insistent as far as I am concerned. Enough to guilt me into coming back to the PR regularly, but not enough to piss me off. ;)

mpl:
> @mpl what's your preference with regards to the branch being out of date now? Do you want me to rebase, merge master, or let you handle conflict upon (a hopeful) merge?

It's always better if you can rebase yourself, as it saves us time. If not, we (either the bot if it does not get confused, or the reviewers) will do it for you.

bjeanes:
No worries I'll rebase it.

Issue #6773, opened 2020-05-06
Cannot correctly escape in IngressRoute 


fefe982:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I'll trying to add an IngressRoute using `HeadersRegexp` to match a literal `.` (dot) in an IP address. Withe the following config:
```
apiVersion: traefik.containo.us/v1alpha1
kind: IngressRoute
metadata:
  name: bug
  namespace: traefik
spec:
  entryPoints:
    - web
  routes:
  - match: Host(`dummy.host`) && HeadersRegexp(`X-Real-Ip`, `172\..*`)
    kind: Rule
    services:
    - name: dummy-svc
      namespace: dummy-test
      port: 80
```

### What did you expect to see?

It should be a correct rule and can match `X-Real-Ip: 172.18.3.1`.

### What did you see instead?

1. This rule would not show up in the UI. If there is an error, it should show up in the dashboard. Disappearing from the dashboard made me confused and hard to track what is going on.
2. It doesn't take effect
3. The logs of traefik has the following line:
```
time="2020-05-06T15:12:41Z" level=error msg="Invalid syntax for match rule: Host(`dummy.host`) && HeadersRegexp(`X-Real-Ip`, `172\\..*`)" ingress=bug namespace=traefik providerName=kubernetescrd
```

By the way, I also tried with double backslash ``HeadersRegexp(`X-Real-Ip`, `172\\..*`)``. It is recognized as a correct rule, but it cannot match an IP of `172.18.3.0`.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      2.2.0
Codename:     chevrotin
Go version:   go1.14.1
Built:        2020-03-25T17:32:57Z
OS/Arch:      linux/amd64

```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

Installed using helm chart `traefik/traefix` with helm3 from "https://containous.github.io/traefik-helm-chart" into a kubernetes cluster.

```toml
# (paste your configuration here)
```

### If applicable, please paste the log output in DEBUG level (`--log.level=DEBUG` switch)

```
(paste your output here)
```

fefe982:
[...]

PR #6775, opened 2020-05-07
Remove checkStringQuoteValidity in loadIngressRouteConf

fefe982:
### What does this PR do?

This PR is to show a simple fix to the issue "Cannot correctly escape in IngressRoute" containous/traefik#6773 . It may not ready to be merged in production code.

With this patch the configuration can work as expected.

However, without knowing what `checkStringQuoteValidity` is intended to check, I cannot provide a replacement.

### More

- [ ] Added/updated tests


dtomcej:
Hello @fefe982,

This function should be removable in its entirety.

Can you update this PR to remove the function and all its references?

fefe982:
There are two copies of the `checkStringQuoteValidity` function, one in  "[pkg/provider/kubernetes/crd/kubernetes.go](https://github.com/containous/traefik/blob/f12c27aa7cf0be864fba7788f0891602eda65dff/pkg/provider/kubernetes/crd/kubernetes.go#L612)", the other in "[pkg/provider/kubernetes/ingress/kubernetes.go](https://github.com/containous/traefik/blob/f12c27aa7cf0be864fba7788f0891602eda65dff/pkg/provider/kubernetes/ingress/kubernetes.go#L562)".

Do they both need to be removed?

PR #6780, opened 2020-05-08
Add CustomHeaderStrategy

### LINKED ISSUE #5713: ipWhiteList middleware different header than X-Forwarded-For
Feature

When using the middleware `ipWhiteList`, the middleware will look inside of the `X-Forwarded-For` header for the IP matching.
It would be good to be able to specify which header to use instead of the `X-Forwarded-For`.
I'm currently using Cloudflare to serve the domain of my kubernetes cluster, and when the request hits the Traefik pod, the `X-Forwarded-For` header only contains the IP from Cloudflare, not the client IP. The client IP can be found in the `CF-CONNECTING-IP` instead, so if I could use this header in the middleware `ipWhiteList`, that would easily solve this problem.

NerdsvilleCEO:
Default behavior was to rely on X-Forwarded-For header in order to get the true client IP.
In some cases, especially ones where a CDN is in use, this behavior won't suffice.
This PR resolves that with a new strategy for IP checking.

### What does this PR do?

This PR adds a new IP strategy to allow true client IP resolution method to be changed.

### Motivation

A friend of mine was struggling with using CloudFlare with Traefik, so I decided to implement a potential fix. (Please thoroughly review as I am new to this project :))

fixes #5713

### More

- [X] Added/updated tests
- [-PARTIAL-] Added/updated documentation

### Additional Notes

Please let me know where to update the documentation for this change, I updated comments but not any other docs.

### DIFF
--- a/pkg/config/dynamic/middlewares.go
+++ b/pkg/config/dynamic/middlewares.go
@@ -237,14 +237,17 @@ func (h *Headers) HasSecureHeadersDefined() bool {
 
 // IPStrategy holds the ip strategy configuration.
 type IPStrategy struct {
-	Depth       int      `json:"depth,omitempty" toml:"depth,omitempty" yaml:"depth,omitempty" export:"true"`
-	ExcludedIPs []string `json:"excludedIPs,omitempty" toml:"excludedIPs,omitempty" yaml:"excludedIPs,omitempty"`
+	Depth              int      `json:"depth,omitempty" toml:"depth,omitempty" yaml:"depth,omitempty" export:"true"`
+	ExcludedIPs        []string `json:"excludedIPs,omitempty" toml:"excludedIPs,omitempty" yaml:"excludedIPs,omitempty"`
+	TrueClientIPHeader string   `json:"trueClientIPHeader" toml:"trueClientIPHeader" yaml:"trueClientIPHeader"`
 	// TODO(mpl): I think we should make RemoteAddr an explicit field. For one thing, it would yield better documentation.
 }
 
 // Get an IP selection strategy.
 // If nil return the RemoteAddr strategy
-// else return a strategy base on the configuration using the X-Forwarded-For Header.
+// else return a strategy based on the configuration
+// using either the X-Forwarded-For Header, or custom
+// header when using CustomHeaderStrategy.
 // Depth override the ExcludedIPs
 func (s *IPStrategy) Get() (ip.Strategy, error) {
 	if s == nil {
@@ -254,6 +257,12 @@ func (s *IPStrategy) Get() (ip.Strategy, error) {
 	if s.Depth > 0 {
 		return &ip.DepthStrategy{
 			Depth: s.Depth,
+		}, nil
+	}
+
+	if s.TrueClientIPHeader != "" {
+		return &ip.CustomHeaderStrategy{
+			Header: s.TrueClientIPHeader,
 		}, nil
 	}
 
--- a/pkg/ip/strategy.go
+++ b/pkg/ip/strategy.go
@@ -43,6 +43,18 @@ func (s *DepthStrategy) GetIP(req *http.Request) string {
 	return strings.TrimSpace(xffs[len(xffs)-s.Depth])
 }
 
+// CustomHeaderStrategy is a strategy which allows
+// to check the trusted IP list against a header
+// other than X-Forwarded-For
+type CustomHeaderStrategy struct {
+	Header string
+}
+
+// GetIP returns the IP from the custom header
+func (s *CustomHeaderStrategy) GetIP(req *http.Request) string {
+	return strings.TrimSpace(req.Header.Get(s.Header))
+}
+
 // CheckerStrategy a strategy based on an IP Checker
 // allows to check that addresses are in a trusted IPs
 type CheckerStrategy struct {
--- a/pkg/ip/strategy_test.go
+++ b/pkg/ip/strategy_test.go
@@ -33,6 +33,35 @@ func TestRemoteAddrStrategy_GetIP(t *testing.T) {
 	}
 }
 
+func TestCustomHeaderStrategy_GetIP(t *testing.T) {
+	testCases := []struct {
+		desc           string
+		expected       string
+		cfConnectingIP string
+	}{
+		{
+			desc:           "Use CustomHeader",
+			expected:       "192.0.2.1",
+			cfConnectingIP: "192.0.2.1",
+		},
+	}
+
+	for _, test := range testCases {
+		test := test
+		t.Run(test.desc, func(t *testing.T) {
+			t.Parallel()
+
+			strategy := CustomHeaderStrategy{
+				Header: "CF-Connecting-IP",
+			}
+			req := httptest.NewRequest(http.MethodGet, "http://127.0.0.1", nil)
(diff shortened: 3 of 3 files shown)


NerdsvilleCEO:
I'll add a comment to the exported method I missed for commenting

NerdsvilleCEO:
@rouing I'll fix the merge conflicts that occurred since this was submitted til now

pkg/config/dynamic/middlewares.go
```diff
@@ -237,14 +237,17 @@ func (h *Headers) HasSecureHeadersDefined() bool {
 
 // IPStrategy holds the ip strategy configuration.
 type IPStrategy struct {
-	Depth       int      `json:"depth,omitempty" toml:"depth,omitempty" yaml:"depth,omitempty" export:"true"`
-	ExcludedIPs []string `json:"excludedIPs,omitempty" toml:"excludedIPs,omitempty" yaml:"excludedIPs,omitempty"`
+	Depth              int      `json:"depth,omitempty" toml:"depth,omitempty" yaml:"depth,omitempty" export:"true"`
+	ExcludedIPs        []string `json:"excludedIPs,omitempty" toml:"excludedIPs,omitempty" yaml:"excludedIPs,omitempty"`
+	TrueClientIPHeader string   `json:"trueClientIPHeader" toml:"trueClientIPHeader" yaml:"trueClientIPHeader"`
```
notsureifkevin, line 242: using `true` here seems contrary to the strategy deployed, which is `CustomHeaderStrategy`, in addition to considering the assumptions one makes with regards to booleans. Should we consider renaming this to `CustomClientIPHeader`?

NerdsvilleCEO: @notsureifkevin: I'm fine with changing the naming, however the reason I named it this is true client IP is the terminology for getting the actual IP address for the requesting client. I'd be open to getting more feedback on this because I'm fine either way as long as it makes sense :)

notsureifkevin:
> Please let me know where to update the documentation for this change

This is deployed in multiple middleware configurations, once written though should be able to be applied universally to the following pages:

https://github.com/containous/traefik/blob/master/docs/content/middlewares/inflightreq.md
https://github.com/containous/traefik/blob/master/docs/content/middlewares/ipwhitelist.md
https://github.com/containous/traefik/blob/master/docs/content/middlewares/ratelimit.md

pkg/config/dynamic/middlewares.go
NerdsvilleCEO: @notsureifkevin doesn't look like we are getting much other feedback on this, at least so far... do you think I should go ahead and make that change?

NerdsvilleCEO:
> > Please let me know where to update the documentation for this change
> 
> This is deployed in multiple middleware configurations, once written though should be able to be applied universally to the following pages:
> 
> https://github.com/containous/traefik/blob/master/docs/content/middlewares/inflightreq.md
> https://github.com/containous/traefik/blob/master/docs/content/middlewares/ipwhitelist.md
> https://github.com/containous/traefik/blob/master/docs/content/middlewares/ratelimit.md

Understood, I will update the documentation in these places

mpl:
@NerdsvilleCEO I don't get it. According to https://support.cloudflare.com/hc/en-us/articles/200170986-How-does-Cloudflare-handle-HTTP-Request-headers- , even if there were other proxies on the route, the original client IP should still be in the comma separated list of values for X-Forwarded-For. I should be the first one actually. Why does your friend say the list only contains the value of the CloudFlare proxy?

NerdsvilleCEO:
@mpl thanks for your reply. In that exact link it also says not to rely on X-Forwarded-For to retrieve true client IP. I'm also basing this on my own professional CDN experience. Making the header configurable allows for more flexibility in general in the cases such as cloudflare where there is a specific single address header.

mpl:
> @mpl thanks for your reply. In that exact link it also says not to rely on X-Forwarded-For to retrieve true client IP.

I do not think that's accurate. I understand they advise to use CF-Connecting-IP, because as it is single-valued, it is easier to deal with than X-Forwarded-For, which might be multi-valued. They do not claim X-Forwarded-For should not be used.

> I'm also basing this on my own professional CDN experience that X-Forwarded-For tends to be used to prevent CDN routing loops since it stores the path the requests take. Making the header configurable allows for more flexibility in general in the cases such as cloudflare where there is a specific single address header.

Yes, but at the (unnecessary imho) cost of added complexity to the traefik config overall, so we have to weight one against the other.

NerdsvilleCEO:
@mpl, what I am suggesting is `X-Forwarded-For` is not a standard (see also the MDN reference https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) and while it is a de-facto common way of retrieving true client IP, it is not guaranteed to work as expected. There are many variables at hand that make it very difficult to determine why we are seeing the behavior that X-Forwarded-For is not accurate through cloudflare for this user. I tested as well on his proxy and was seeing the same behavior he was describing, until he enabled true client IP support in cloudflare and then it existed under the `CF-Connecting-IP` header.

Feel free to close this PR, but my one suggestion would be, if there isn't one already a mechanism to create custom middleware would be super helpful.

Thanks.

mpl:
> @mpl, what I am suggesting is `X-Forwarded-For` is not a standard (see also the MDN reference https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) and while it is a de-facto common way of retrieving true client IP, it is not guaranteed to work as expected. There are many variables at hand that make it very difficult to determine why we are seeing the behavior that X-Forwarded-For is not accurate through cloudflare for this user. I tested as well on his proxy and was seeing the same behavior he was describing, until he enabled true client IP support in cloudflare and then it existed under the `CF-Connecting-IP` header.
> 
> Feel free to close this PR, but my one suggestion would be, if there isn't one already a mechanism to create custom middleware would be super helpful.

Not closing yet, still need to think about it, but in any case:
https://github.com/containous/traefik/issues/1336 

> Thanks.

mpl:
@NerdsvilleCEO After some additional thinking we're still reluctant to add this change, sorry.

For once, it seems to us that if Cloudflare is indeed behaving poorly wrt to X-Forwarded-For (we would need to double-check, but we can take your word for it for the sake of this discussion), it would be unfair that Traefik (or any other reverse-proxy for that matter) has to be the one compensating for their mistakes. Why should we shoulder the burden of additional code complexity when they're the ones messing up? So we think you should ask them to fix the problem on their end.

In addition, there's additional trust and security issues to be considered, that haven't been implemented in this PR. Namely, if you want your whole trust chain to keep on working as it would if you used X-Forwarded-For, you would need to trust whatever other header you want to replace it with (in that case, `CF-Connecting-IP`). Which means among other things that it should be added to the list of headers that you see at `pkg/middlewares/forwardedheaders/forwarded_header.go`, which cannot be done dynamically (and for good reason).

halradaideh:
@mpl 

we are in April 2021, and this still an issue, I totally understand your point, but this would simply move people away from Traefik

alexdelprete:
In case someone needs this for CloudFlare: https://pilot.traefik.io/plugins/610e1a6c2d6e04b9c3de1a27/rea-l-ip-from-cloudflare

Without that plugin, I couldn't obtain the client's real IP, with my domain behind CF proxy.

PR #6785, opened 2020-05-10
Fix cors bug when migrate from v2.0 to v2.2 and accessControlAllowOrigin is "origin-list-or-null"

letian0805:
### What does this PR do?

Return `origin` instead of `item` in function `isOriginAllowed` in `pkg/middlewares/headers/headers.go`when item is `origin-list-or-null`.

### Motivation

Chrome reports error when migrate traefik from v2.0 to v2.2 and  set `accessControlAllowOrigin` to `"origin-list-or-null"`.
Return `origin` instead of `item` in function `isOriginAllowed` in `pkg/middlewares/headers/headers.go` will fix the bug.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/middlewares/headers/headers.go
+++ b/pkg/middlewares/headers/headers.go
@@ -272,6 +272,8 @@ func (s *Header) isOriginAllowed(origin string) (bool, string) {
 	for _, item := range s.headers.AccessControlAllowOriginList {
 		if item == "*" || item == origin {
 			return true, item
+		} else if item == "origin-list-or-null" {
+			return true, origin
 		}
 	}
 


dtomcej:
According to (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin), the wildcard response is valid...

Can you elaborate on the issue/user case that this is attempting to solve?

letian0805:
> According to (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin), the wildcard response is valid...
> 
> Can you elaborate on the issue/user case that this is attempting to solve?
Chrome reports error when set accessControlAllowOriginList to ["*"]

image

letian0805:
> According to (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin), the wildcard response is valid...
> 
> Can you elaborate on the issue/user case that this is attempting to solve?

@dtomcej You are right. I have updated this pull request to fix bug when migrate from v2.0 and `Access-Control-Allow-Origin` is `origin-list-or-null`

rtribotte:
Hi @letian0805,

While we dropped the support of returning null values (as stated in the documentation: https://docs.traefik.io/middlewares/headers/#accesscontrolalloworiginlist), we dropped the support of `origin-list-or-null`.
Because the option `origin-list-or-null` implies to return a null value if the origin is not matching.

For this reason, i don't think we want to re-introduce the `origin-list-or-null` option.
From what i understand, you wanted the ability to return only the origin value.
But as @dtomcej mentionned it, the `*` value is valid, and furthermore, when the `accessControlAllowOrigin` is configured with a subset of origins, only the matching origin value is returned.

Also, when migrating from v2.0 to v2.2 the `accessControlAllowOrigin` value previously settled in you configuration will not be used, because it has been replaced by the new `accessControlAllowOriginList` option.

WDYT ?

letian0805:
> Hi @letian0805,
> 
> While we dropped the support of returning null values (as stated in the documentation: https://docs.traefik.io/middlewares/headers/#accesscontrolalloworiginlist), we dropped the support of `origin-list-or-null`.
> Because the option `origin-list-or-null` implies to return a null value if the origin is not matching.
> 
> For this reason, i don't think we want to re-introduce the `origin-list-or-null` option.
> From what i understand, you wanted the ability to return only the origin value.
> But as @dtomcej mentionned it, the `*` value is valid, and furthermore, when the `accessControlAllowOrigin` is configured with a subset of origins, only the matching origin value is returned.
> 
> Also, when migrating from v2.0 to v2.2 the `accessControlAllowOrigin` value previously settled in you configuration will not be used, because it has been replaced by the new `accessControlAllowOriginList` option.
> 
> WDYT ?

Hi @rtribotte ,

I know `accessControlAllowOrigin` and `origin-list-or-null` will be dropped. But some users are using `origin-list-or-null` in v2.0 and v2.1.  Browser reports CORS error when upgrading traefik to v2.2 and set accessControlAllowOrigin value to `*` and framework is React or VUE.

image

rtribotte:
Hi @letian0805,

To better understand your use case, i wonder why you can't use an "allow list" of origins in your configuration ?
Also, for the sake of clarification, when you refer to `accessControlAllowOrigin`, do you speak of the new option `accessControlAllowOriginList` ?

If so, again, it seems, after some [reading](https://stackoverflow.com/a/45723981), that the `withCredentials` angular option should be either set to `false`, or be used with a set of allowed origins. In that case, i would say that the current configuration (v2.2) feet with this use case.

WDYT ?

rtribotte:
Hi @letian0805,

I'm sorry, we don't see a valid use case here, or at least after discussion we don't get it.
Since removing the support of `origin-list-or-null` has been a security choice, we don't want to re-introduce it.

So i close the PR.

Nevertheless, thank you for your time and for your contribution.
Feel free to discuss about your problem by opening a post on our [community forum](https://community.containo.us/c/traefik/traefik-v2/10) or an issue, and see if it can be reconsidered as possible enhancement.

letian0805:
> Hi @letian0805,
> 
> To better understand your use case, i wonder why you can't use an "allow list" of origins in your configuration ?
> Also, for the sake of clarification, when you refer to `accessControlAllowOrigin`, do you speak of the new option `accessControlAllowOriginList` ?
> 
> If so, again, it seems, after some [reading](https://stackoverflow.com/a/45723981), that the `withCredentials` angular option should be either set to `false`, or be used with a set of allowed origins. In that case, i would say that the current configuration (v2.2) fits with this use case.
> 
> WDYT ?

Hi @rtribotte 

Thanks reply me!

image

According to above code in traefik v2.1,  we can't set `AccessControlAllowOrigin` to a exact host. Otherwise it will return `""`instead of `origin`. 

We can only set it to `origin-list-or-null` or `*`. But if we want to set it to `*`, we must change our front-end code to disable  `withCredentials`. So we use `origin-list-or-null` in traefik v2.1.

Our front-end javascript codes may be cached by browser or  mobile application. If we change our front-end code, and set  `AccessControlAllowOrigin` to `*`in traefik,  may cause an error in the user's request. This is what we don't want to see.

So, in order to smoothly upgrade to traefik v2.2,  we should change traefik's code to compatible with `origin-list-or-null` and upgrade to traefik v2.2. Then we can set `AccessControlAllowOrigin` to a exact host. We don't need to change the front-end code.

In order to  smoothly upgrade to traefik v2.2, we can fix the bug of traefik v2.1, or compatible with `origin-list-or-null`in traefik v2.2.

As many users are upgrading traefik to v2.2, i suggest to traefik v2.2 compatible with `origin-list-or-null`.

WDYT?

rtribotte:
Hi @letian0805,

As a best practice, I encourage you to open an issue or a thread on the [community forum](https://community.containo.us/), because closed PR are not the right place.
Nevertheless, you are right, the v2.2 introduces a breaking change, and it will potentially break things when migrating without updating the front end code of one application, especially in your case, without the list of the allowed origins set in the configuration.
Thus, like i said, we dropped this option because of security concerns, and reintroducing the option to be backward compatible is not convincing enough, because you can either:
- update your frontend code prior migrating to v2.2 in order to solve cache issues
- configuring the set of allowed origin
- even handling cors on your backends

Feel free to come and discuss about your migration strategy on our [community forum](https://community.containo.us/) !

Issue #6761, opened 2020-05-02
Memory leak


FL3SH:
### Do you want to request a *feature* or report a *bug*?
Bug
### What did you do?
Configure traefik with kubernetsCRD using IngressRoute, IngressRouteTCP, and IngressRouteUDP.

I think it might be correlated with my IngressRouteTCP.
### What did you expect to see?
Memory usage shouldn't constantly increase until pod is killed by Kubernetes.

### What did you see instead?
Constantly increasing memory usage.

<img width="1602" alt="Screenshot 2020-05-02 at 14 54 42" src="https://user-images.githubusercontent.com/9084725/80869070-84dcb680-8c9e-11ea-8164-1050f04bf6ef.png">

 If necessary I can post more screenshots from Prometheus.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
image: traefik:2.2.1 (same behavior for 2.2.0)
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?
Kubernetes CRD
```yaml
apiVersion: traefik.containo.us/v1alpha1
kind: IngressRouteTCP
metadata:
  name: mqtt
  namespace: default
spec:
  entryPoints:
  - mqtt
  routes:
  - match: HostSNI(`*`)
    services:
    - name: mosquitto
      port: 8883
---
apiVersion: traefik.containo.us/v1alpha1
kind: IngressRouteUDP
metadata:
  name: radius1-udp
  namespace: default
spec:
  entryPoints:
  - radius1
  routes:
  - services:
    - name: myapp-udp
      port: 1812
---
apiVersion: traefik.containo.us/v1alpha1
kind: IngressRouteUDP
metadata:
  name: radius2-udp
  namespace: default
spec:
  entryPoints:
  - radius2
  routes:
  - services:
    - name: myapp-udp
      port: 1813
```

IngressRoutes are quite long, let me know if I need to post them too.

### If applicable, please paste the log output in DEBUG level (`--log.level=DEBUG` switch)
I think only applicable log that I am seeing:
```
msg="Error during connection: readfrom tcp 10.128.2.20:47420->10.128.4.15:8883: read tcp 10.128.2.20:8883->202.130.98.126:29192: read: connection reset by peer"
```

FL3SH:
After some testing, I determined that problem is in handing UDP.
After removing IngressRouteUDP memory leak disappear.
You can easily reproduce this issue by sending UDP packest ex. using bash:
```bash
[...]

PR #6815, opened 2020-05-16
Fix mem leak on UDP connections

ddtmachado:
### What does this PR do?

Fix a memory / go routine leak on udp connections, specifically in allocating a new `time.Ticker` under the `Conn` struct.

**No expert here**, but after taking a heap profile in a stress test which created lots of new udp connections it showed that we were leaking memory, and probably go routines (runtime malg), as we allocated a time.Ticker in `udp.(*Listener)newConn`. This is also consistent with the profile sent on #6761.

traefik_2 1 1_leak

It could be that the `Ticker`, which was referenced by the `Conn`, was not being collected by the GC, justifying the dead go routines and memory leak.

Changing the implementation to allocate and stop the ticker in the main `readLoop` seems to solve the memory and go routine problems as demonstrated in the new profile taken after running the same stress test:

traefik_2 1 1_leak_fix

### Motivation

Issue #6761

### More

- [ ] ~~Added/updated tests~~

### DIFF
--- a/pkg/udp/conn.go
+++ b/pkg/udp/conn.go
@@ -177,7 +177,7 @@ func (l *Listener) newConn(rAddr net.Addr) *Conn {
 		readCh:    make(chan []byte),
 		sizeCh:    make(chan int),
 		doneCh:    make(chan struct{}),
-		ticker:    time.NewTicker(timeoutTicker),
+		timeout:   timeoutTicker,
 	}
 }
 
@@ -194,7 +194,7 @@ type Conn struct {
 	muActivity   sync.RWMutex
 	lastActivity time.Time // the last time the session saw either read or write activity
 
-	ticker   *time.Ticker // for timeouts
+	timeout  time.Duration // for timeouts
 	doneOnce sync.Once
 	doneCh   chan struct{}
 }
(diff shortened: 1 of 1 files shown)


pkg/udp/conn.go
```diff
@@ -204,12 +204,15 @@ type Conn struct {
 // that is to say it waits on readCh to receive the slice of bytes that the Read operation wants to read onto.
 // The Read operation receives the signal that the data has been written to the slice of bytes through the sizeCh.
 func (c *Conn) readLoop() {
+	ticker := time.NewTicker(c.timeout)
+	defer ticker.Stop()
```
mpl, line 208: I don't understand at all why this would fix anything for now.
If we suppose (even though I don't know why or how) that the problem that somehow ticker.Stop was not called, then I understand why L208 (VS L284) might make a difference.

But then why is L207 needed at all? Why can't L208 just be:

```
    defer c.ticker.Stop()
```

?

mpl: Also, if we keep on assuming that the problem comes from ticker.Stop not being called, that probably means that c.Close was not always called, which is still a big problem in itself. Because, among other things, that would mean c.listener.conns is not being cleaned up either (which would also be a memleak).

This is all very puzzling.

ddtmachado: I confirmed c.Close was called (actually multiple times once the timeout was reached), so c.ticker.Stop() was actually being called as well but for some reason the routine and objects were not GCed. I'm puzzled with this as well, my best guess is that the conn, stored in c.listener.conns is not being cleaned up as well but I confirmed it gets removed from the map.

So you are right that we might have more to it than just the ticker, and actually I don't see memory getting back to the starting point level after experiencing a lot of new connections, maybe its conn that is not being GCed for some reason.

Issue #2957, opened 2018-03-02
Not working with intermediate certs from InCommon


dimm0:
### Do you want to request a *feature* or report a *bug*?

bug

### What did you do?
Deployed traefik as kubernetes ingress with SSL certificate from InCommon. The certificate if multi-domain wildcart (domain.com, *.domain.com)

### What did you expect to see?
Work properly with chain of certificates.

### What did you see instead?
It works fine if I only use the certificate itself, but fails if I use certificate with intermediate certs, in any order.

I'm getting:
```time="2018-03-02T00:44:26Z" level=error msg="Unable to add a certificate to the entryPoint "https" : unable to generate TLS certificate : tls: private key does not match public key"```

### Output of `traefik version`: (_What version of Traefik are you using?_)

1.5.3

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

kubeadm kubernetes 1.9.3

dimm0:
I'm looking at traefik code at https://github.com/containous/traefik/blob/a2db3e0499ce8319f3346d0f123d9d62a73d997c/tls/certificate.go#L144

Seems like this https://golang.org/pkg/crypto/tls/#LoadX509KeyPair function supports intermediate certs, while the one used in traefik https://golang.org/pkg/crypto/tls/#X509KeyPair does not

ldez:
I think it's just a missing comment because the difference between `LoadX509KeyPair` and `X509KeyPair` is only `ioutil.ReadFile`

https://github.com/golang/go/blob/1023b016d51ee0e6b7245482b9d8715170d785ee/src/crypto/tls/tls.go#L173-L188

dimm0:
Ok, agree... I guess I'll submit the bug to go directly then, after making a minimal test example.

dimm0:
I figured it. Needed to manually combine the provided certs to have the order right.

themaxx32000:
@dimm0 what's the correct order? I'm experiencing a similar issue...

PR #6820, opened 2020-05-18
feat: document how to use intermediate certificates

Trois-Six:
### What does this PR do?

Just add a sentence on how we support intermediate certificates in Traefik

### Motivation

Looking at the documentation, I asked myself if Traefik supports intermediate certificates, and it wasn't obvious. I am not the only one to ask the question: https://github.com/containous/traefik/issues/2957.

### More

- [ ] Added/updated tests

### Additional Notes

I am not sure about the location of that new paragraph ?

### DIFF
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -130,6 +130,15 @@ tls:
 
 If no default certificate is provided, Traefik generates and uses a self-signed certificate.
 
+### Intermediate Certificate
+
+If you need to chain certificates to establish the chain of trust, you
+can concatenate your intermediate certificate after the certificate:
+
+```bash tab="CLI"
+cat foo.crt intermediate.crt > cert.crt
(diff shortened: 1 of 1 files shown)


mpl:
I'm not convinced this belongs to the traefik doc itself, so I'll let someone else approve it for now.

PCM2:
Can anyone confirm that the proposed solution works? I don't presently have a development environment that would allow me to do this.

notsureifkevin:
> 
> 
> I'm not convinced this belongs to the traefik doc itself, so I'll let someone else approve it for now.

I second this, this is more of a client / consumer concern, and we're sort of just a messenger.

SantoDE:
Hello @Trois-Six,

after checking, we will turn down this PR for now and see, if it will have a place later on.

Thanks!

PR #6827, opened 2020-05-20
Provide username in log data on auth failure

### LINKED ISSUE #6722: Basic auth does not log username on failure
Hi, 
Is this behavior intended?

### Bug report

Basic auth middleware does not log provided username when authentication failed.

### What did you do?

1. Existing user, bad password: username IS NOT present in the log
```
$ curl http://test:testdsds@traefik.localhost.dv/user-exists-badpass
> 172.26.0.1 - - [20/Apr/2020:14:09:03 +0000] "GET /user-exists-badpass HTTP/1.1" 401 17 "-" "-" 5 "whoami@docker" "-" 1ms
```
`172.26.0.1 - -` **please note no username is logged**

2. Existing user, good password: username IS present in the log
```
$ curl http://test:test@traefik.localhost.dv/user-exists-goodpass
> 172.26.0.1 - test [20/Apr/2020:14:07:08 +0000] "GET /user-exists-goodpass HTTP/1.1" 200 396 "-" "-" 3 "whoami@docker" "http://172.26.0.2:80" 9ms
```
3. Unknown user: username IS NOT present in the log
```
$ curl http://noexist:badpass@traefik.localhost.dv/no-exists
> 172.26.0.1 - - [20/Apr/2020:14:14:49 +0000] "GET /no-exists HTTP/1.1" 401 17 "-" "-" 1 "whoami@docker" "-" 0ms
```

According https://github.com/fail2ban/fail2ban/issues/2693#issuecomment-618257204, the username should be logged when the authentication failed.

Excepted log:
```
$ curl http://test:test@traefik.localhost.dv/user-exists-goodpass
> 172.26.0.1 - test [20/Apr/2020:14:14:49 +0000] "GET /no-exists HTTP/1.1" 401 17 "-" "-" 1 "whoami@docker" "-" 0ms
```

### What did you see instead?

```
$ curl http://test:test@traefik.localhost.dv/user-exists-goodpass
> 172.26.0.1 - - [20/Apr/2020:14:14:49 +0000] "GET /no-exists HTTP/1.1" 401 17 "-" "-" 1 "whoami@docker" "-" 0ms
```

Tested with traefik 1.7 and traefik 2.2: same behavior observed.

Configuration used for tests is available at https://github.com/fail2ban/fail2ban/issues/2693#issuecomment-616583684

rtribotte:
### What does this PR do?

This PR adds the username in the log data, if present, on auth failure for middlewares basic auth and digest auth.

### Motivation

fixes #6722
Providing the user on auth failure in the access log is an expected behavior for some tools, at least fail2ban.

### More

- [x] Added/updated tests

### DIFF
--- a/integration/access_log_test.go
+++ b/integration/access_log_test.go
@@ -111,6 +111,20 @@ func (s *AccessLogSuite) TestAccessLogAuthFrontend(c *check.C) {
 			routerName: "rt-authFrontend",
 			serviceURL: "-",
 		},
+		{
+			formatOnly: false,
+			code:       "401",
+			user:       "test",
+			routerName: "rt-authFrontend",
+			serviceURL: "-",
+		},
+		{
+			formatOnly: false,
+			code:       "200",
+			user:       "test",
+			routerName: "rt-authFrontend",
+			serviceURL: "http://172.17.0",
+		},
 	}
 
 	// Start Traefik
@@ -130,12 +144,22 @@ func (s *AccessLogSuite) TestAccessLogAuthFrontend(c *check.C) {
 	// Verify Traefik started OK
 	checkTraefikStarted(c)
 
-	// Test auth frontend
+	// Test auth entrypoint
 	req, err := http.NewRequest(http.MethodGet, "http://127.0.0.1:8006/", nil)
 	c.Assert(err, checker.IsNil)
 	req.Host = "frontend.auth.docker.local"
 
 	err = try.Request(req, 500*time.Millisecond, try.StatusCodeIs(http.StatusUnauthorized), try.HasBody())
+	c.Assert(err, checker.IsNil)
+
+	req.SetBasicAuth("test", "")
+
+	err = try.Request(req, 500*time.Millisecond, try.StatusCodeIs(http.StatusUnauthorized), try.HasBody())
+	c.Assert(err, checker.IsNil)
+
+	req.SetBasicAuth("test", "test")
+
+	err = try.Request(req, 500*time.Millisecond, try.StatusCodeIs(http.StatusOK), try.HasBody())
 	c.Assert(err, checker.IsNil)
 
 	// Verify access.log output as expected
@@ -155,6 +179,13 @@ func (s *AccessLogSuite) TestAccessLogDigestAuthMiddleware(c *check.C) {
 			formatOnly: false,
 			code:       "401",
 			user:       "-",
+			routerName: "rt-digestAuthMiddleware",
+			serviceURL: "-",
+		},
+		{
+			formatOnly: false,
+			code:       "401",
+			user:       "test",
 			routerName: "rt-digestAuthMiddleware",
 			serviceURL: "-",
 		},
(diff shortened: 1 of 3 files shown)


integration/access_log_test.go
```diff
@@ -458,54 +496,6 @@ func (s *AccessLogSuite) TestAccessLogFrontendWhitelist(c *check.C) {
 	checkNoOtherTraefikProblems(c)
 }
 
-func (s *AccessLogSuite) TestAccessLogAuthFrontendSuccess(c *check.C) {
```
mpl, line 461: why is this test removed?

pkg/middlewares/auth/basic_auth.go
```diff
@@ -65,6 +65,14 @@ func (b *basicAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	if username := b.auth.CheckAuth(req); username == "" {
 		logger.Debug("Authentication failed")
 		tracing.SetErrorWithEvent(req, "Authentication failed")
+
+		if username, _, ok := req.BasicAuth(); ok {
```
mpl, line 69: It is a bit silly and wasteful that we call req.BasicAuth() another time, whereas we have already computed the result in CheckAuth. Maybe L65 we should stop using CheckAuth, so we don't lose the result of req.BasicAuth (i.e. the username). Or maybe we could change the signature of CheckAuth so that it returns the username even when it does not match.
WDYT?

pkg/middlewares/auth/basic_auth.go
```diff
@@ -65,6 +65,14 @@ func (b *basicAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+			logData := accesslog.GetLogData(req)
+			if logData != nil {
+				logData.Core[accesslog.ClientUsername] = username
+			}
+		}
+
 		b.auth.RequireAuth(rw, req)
```
mpl, line 76: we could return after this line, and get rid of the else block afterwards.

pkg/middlewares/auth/digest_auth.go
```diff
@@ -63,6 +63,19 @@ func (d *digestAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 	username, authinfo := d.auth.CheckAuth(req)
 	if username == "" {
+		headerField := d.headerField
+		if d.headerField == "" {
+			headerField = "Authorization"
+		}
+
+		auth := goauth.DigestAuthParams(req.Header.Get(headerField))
```
mpl, line 71: similarly, it is a bit silly that we are redoing the computation for the username here, but d.auth.CheckAuth is more complicated in the digest auth case than in the basic auth case, so it does not bother me as much here as in the basic auth case.

integration/access_log_test.go
rtribotte: Because of redundancy with the test L122, but i would agree to reintroduce it for clarity.

pkg/middlewares/auth/basic_auth.go
rtribotte: I agree, it was bothering me also.
Changing the signature is breaking some other usages, if it also seems fair to you to call BasicAuth and check the secret in place of L65, i would go for that.

pkg/middlewares/auth/basic_auth.go
rtribotte: Ok 👌

pkg/middlewares/auth/digest_auth.go
rtribotte: Let say that if you agree to don’t change CheckAuth signature, this one would remain as it is ?

integration/access_log_test.go
mpl: if you are sure we are not losing anything in coverage, no need, thanks.

pkg/middlewares/auth/basic_auth.go
mpl: yes, let's try that and see if it does not look too ugly.

pkg/middlewares/auth/digest_auth.go
mpl: I mean, maybe changing CheckAuth signature is no big deal, if we can easily change the other calls everywhere else. Or we could also write another method that does the same thing as CheckAuth but returns the username. (and CheckAuth would just be a wrapper of it, that drops the username before returning.). (And same thing for BasicAuth).
But yes, if it does not bother you, feel free to ignore that comment.

crazy-max:
Nice thanks!

PR #6828, opened 2020-05-20
Middlewares: add wildcard pattern support to forwardauth.authResponse…

### LINKED ISSUE #6823: ForwardAuth authResponseHeaders accept regexp or wildcard

Feature for `ForwardAuth` middleware to accept in `authResponseHeaders` regular expressions or wildcard

Real world use-case is adding new headers by some pattern from auth-service without redeploying traefik. For example we could have `X-User-Id` header, but then add `X-User-Role` header. Now we should change current config, but with this feature we could provide `X-User-.*` or `X-User-*` pattern instead at first time.

iamolegga:
…Headers

### What does this PR do?

fixes #6823 

### Motivation

described in #6823 

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/middlewares/forwardauth.md
+++ b/docs/content/middlewares/forwardauth.md
@@ -165,10 +165,11 @@ http:
 ### `authResponseHeaders`
 
 The `authResponseHeaders` option is the list of the headers to copy from the authentication server to the request.
+It could contain wildcard patterns as well.
 
 ```yaml tab="Docker"
 labels:
-  - "traefik.http.middlewares.test-auth.forwardauth.authResponseHeaders=X-Auth-User, X-Secret"
+  - "traefik.http.middlewares.test-auth.forwardauth.authResponseHeaders=X-Auth-User, X-Secret-*"
 ```
 
 ```yaml tab="Kubernetes"
@@ -181,29 +182,29 @@ spec:
     address: https://example.com/auth
     authResponseHeaders:
       - X-Auth-User
-      - X-Secret
+      - X-Secret-*
 ```
 
 ```yaml tab="Consul Catalog"
-- "traefik.http.middlewares.test-auth.forwardauth.authResponseHeaders=X-Auth-User, X-Secret"
+- "traefik.http.middlewares.test-auth.forwardauth.authResponseHeaders=X-Auth-User, X-Secret-*"
 ```
 
 ```json tab="Marathon"
 "labels": {
-  "traefik.http.middlewares.test-auth.forwardauth.authResponseHeaders": "X-Auth-User,X-Secret"
+  "traefik.http.middlewares.test-auth.forwardauth.authResponseHeaders": "X-Auth-User,X-Secret-*"
 }
 ```
 
 ```yaml tab="Rancher"
 labels:
-  - "traefik.http.middlewares.test-auth.forwardauth.authResponseHeaders=X-Auth-User, X-Secret"
+  - "traefik.http.middlewares.test-auth.forwardauth.authResponseHeaders=X-Auth-User, X-Secret-*"
 ```
 
 ```toml tab="File (TOML)"
 [http.middlewares]
   [http.middlewares.test-auth.forwardAuth]
     address = "https://example.com/auth"
-    authResponseHeaders = ["X-Auth-User", "X-Secret"]
+    authResponseHeaders = ["X-Auth-User", "X-Secret-*"]
 ```
 
 ```yaml tab="File (YAML)"
@@ -214,7 +215,7 @@ http:
         address: "https://example.com/auth"
         authResponseHeaders:
           - "X-Auth-User"
-          - "X-Secret"
+          - "X-Secret-*"
 ```
 
 ### `tls`
--- a/pkg/middlewares/auth/forward.go
+++ b/pkg/middlewares/auth/forward.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


iamolegga:
One addition to current implementation:

For dealing with headers I think supporting only `?` and `*` symbols could be enough. If so I could suggest [my package](https://github.com/iamolegga/wildcard), that has limitations to supported symbols but it's `Match` function works 3.37 times faster than `filepath`'s one.

If maintainers are OK with that, I could add it to current PR.

mpl:
@iamolegga Hi.

We think your use-case and the feature you're asking for are justified, and something like that would be nice indeed.

Now about the implementation: your title and comment says wildcard, but AFAICS, the PR would allow support for full-featured globbing as supported by the stdlib, right? so it is not just for suffixes, but patterns like "X-*-Foo" would work too, correct?
I'm asking because we wouldn't want something that only half-implements globbing, as it may support your use-case, but would be useless for other users expecting full-featured globbing.

If yes, then there's still the question of whether globbing is not conflicting with headers contents. Have you thoroughly checked that none of the globbing characters (*, ?, ^, [, ], ... am I forgetting something?) are allowed in a header key?

iamolegga:
> Now about the implementation: your title and comment says wildcard, but AFAICS, the PR would allow support for full-featured globbing as supported by the stdlib, right? so it is not just for suffixes, but patterns like "X-*-Foo" would work too, correct?

Yes

> I'm asking because we wouldn't want something that only half-implements globbing, as it may support your use-case, but would be useless for other users expecting full-featured globbing.

Sure, that's why I preferred to implement this PR at start with built-in [`filepath.Match`](https://golang.org/pkg/path/filepath/#Match) that is support all glob symbols.

> If yes, then there's still the question of whether globbing is not conflicting with headers contents. Have you thoroughly checked that none of the globbing characters (*, ?, ^, [, ], ... am I forgetting something?) are allowed in a header key?

As I understand from [[1]](https://tools.ietf.org/html/rfc7230#section-3.2) [[2]](https://tools.ietf.org/html/rfc7230#section-3.2.6) header keys can have special symbols `^` and `*` (please correct me, if I'm wrong). But I think it should not conflict with those names, anyway, just a fast check shows next result: https://play.golang.org/p/EjgFM8PJcz8, so glob patterns are correct when checking those special characters, and there should be no problems

mpl:
> > Now about the implementation: your title and comment says wildcard, but AFAICS, the PR would allow support for full-featured globbing as supported by the stdlib, right? so it is not just for suffixes, but patterns like "X-*-Foo" would work too, correct?
> 
> Yes
> 
> > I'm asking because we wouldn't want something that only half-implements globbing, as it may support your use-case, but would be useless for other users expecting full-featured globbing.
> 
> Sure, that's why I preferred to implement this PR at start with built-in [`filepath.Match`](https://golang.org/pkg/path/filepath/#Match) that is support all glob symbols.
> 
> > If yes, then there's still the question of whether globbing is not conflicting with headers contents. Have you thoroughly checked that none of the globbing characters (*, ?, ^, [, ], ... am I forgetting something?) are allowed in a header key?
> 
> As I understand from [[1]](https://tools.ietf.org/html/rfc7230#section-3.2) [[2]](https://tools.ietf.org/html/rfc7230#section-3.2.6) header keys can have special symbols `^` and `*` (please correct me, if I'm wrong). But I think it should not conflict with those names, anyway, just a fast check shows next result: https://play.golang.org/p/EjgFM8PJcz8, so glob patterns are correct when checking those special characters, and there should be no problems

Ok, we are still on the fence but it sounds reasonable to me in theory. Our main concern now is there might be some edge cases where the pattern one writes matches some headers that one did not intend to match. Maybe we would feel more at ease with it if we had many more examples of different cases, to help make sure we're not forgetting about a weird case. Do you think you could spend more time writing such test cases?

iamolegga:
> Do you think you could spend more time writing such test cases?

@mpl Sure, I could add more tests, but I hardly understand what cases you want there to be added 🙂 Because from my point of view current test case is already checks that glob pattern is applied successfully with expected result and I think it could be enough. More examples will not add any value because they will just check `filepath.Match` for some pattern-value pairs, and that is not our goal.

But anyway, if I'm wrong and you could add more explanation on that I'm totally fine to add such tests. Thanks! 🙂

mpl:
> > Do you think you could spend more time writing such test cases?
> 
> @mpl Sure, I could add more tests, but I hardly understand what cases you want there to be added 🙂 Because from my point of view current test case is already checks that glob pattern is applied successfully with expected result and I think it could be enough. More examples will not add any value because they will just check `filepath.Match` for some pattern-value pairs, and that is not our goal.

Yes, the goal is not to add some tests just for the sake of it. I was suggesting that because if I wanted to find out weird cases that demonstrate that globbing is a bad idea when the input is a header key, I think I would write more examples and tests to give me ideas and inspiration.

> But anyway, if I'm wrong and you could add more explanation on that I'm totally fine to add such tests. Thanks! 🙂

If I already knew what kind of cases might be problematic I wouldn't ask you to look for them ;)
It is just a suspicion some of us in the team are having that globbing might not be suitable for header keys, and we think we should think a bit harder about it before going further with the implementation.

iamolegga:
> If I already knew what kind of cases might be problematic I wouldn't ask you to look for them ;)

ok, got you now, but unfortunately I don't see any problem with globbing here 🤷‍♂️

> It is just a suspicion some of us in the team are having that globbing might not be suitable for header keys

But I described the the case in [issue](https://github.com/containous/traefik/issues/6823) already. So, do I understand correctly that I should just wait now for the final decision of the team? 🙄

mpl:
> > If I already knew what kind of cases might be problematic I wouldn't ask you to look for them ;)
> 
> ok, I got you now, but unfortunately I don't see any problem with globbing here 🤷‍♂️
> 
> > It is just a suspicion some of us in the team are having that globbing might not be suitable for header keys
> 
> But I described the the case in [issue](https://github.com/containous/traefik/issues/6823) already. So, do I understand correctly that I should just wait now for the final decision of the team? 🙄

yes, we need a bit more time to think about it.

iamolegga:
Thanks @mpl, if it needs some more improvements just let me know

mpl:
To keep you updated, there are several things to consider about this PR, before we can move forward:

- first of all, as it is it might be breaking, because if some user was already matching a header specifically containing one of the special characters (e.g. "*"), then the meaning of their config will suddenly change, and it might start matching things they didn't previously intend to match.
	- one of the solutions for that particular problem would be to add/use another field, instead of changing the meaning of authResponseHeaders

- secondly, even if globbing addresses your particular problem, it is a limited response. And if another user comes along later and needs something more powerfull, such as regexes, again we'll be stuck because we won't be able to change the syntax from globbing to regexes without breaking existing users (such as you).
	- again that could be addressed if we used other specific fields for each syntax, but it could get cumbersome to support both globbing and regexes for the same feature, so we'd rather avoid that.

- Finally, there's also the argument of existing practices and familiarity in the config. We already make use of regexes in several places, and even sometimes of a specific syntax, but nowhere do we use globbing. So we are worried it could be a confusing change in that respect.

To sum up, we need to consider and decide 1) if we want to use another field instead of directly authResponseHeaders 2) if we want to use another matching system instead of globbing. It could be regexp, or it could be another made up system entirely (probably something simple like a boolean matching of HasPrefix, HasSuffix, Contains).

iamolegga:
@mpl Thanks you for such detailed response. My thoughts:

if traefik is using regexp in some config fields already and we want to keep consistency I could change PR to use regexp, that's not a problem for me.
Adding new property for that is also good from my point of view.
So if your final decision is using regexp and adding new property, please just say how to name it, and I'll update this PR.

mpl:
@iamolegga yes, we're leaning towards regexp + a new field, as it seems the safest. So if you want to implement that solution, feel free to go about it. We haven't thought about the name of the new field yet, but we can discuss/change that later if need be.

iamolegga:
@mpl cool, thanks! I think I'll start with just adding `Regex` to the end of `authResponseHeaders` so it will be `authResponseHeadersRegex` just as `RedirectRegex` is named for example. I'll try to update this PR soon

jbdoumenjou:
Hi @iamolegga ,

Do you think you will have the time to work on this?

iamolegga:
Hi, sorry for waiting so long. Yes, I am definitely going to update this in two weeks 🙏

jbdoumenjou:
Great, thanks!

iamolegga:
Hi, sorry for long waiting, I've just created new pr #7449 , closing this in favour of new one

PR #6835, opened 2020-05-22
Support X-Auth-Request-Redirect

### LINKED ISSUE #6814: Support X-Auth-Request-Redirect header in ForwardAuth

Feature

Adding support for 'X-Auth-Request-Redirect' header to ForwardAuth middleware so that oauth2_proxy (https://github.com/oauth2-proxy/oauth2-proxy) can be easily used as an auth server.

pieter-lazzaro:
### What does this PR do?

Fixes #6814 

### Motivation

See #6814 

### More

- [x] Added/updated tests


rtribotte:
Hi @pieter-lazzaro,

Thank you for your interest in the project and your contribution.

At first read, your pull request seems to solve one specific use case, the one from the issue you linked, to deal with the `X-Auth-Request-Redirect` header and support oauth2_proxy.
One concern we have is that we probably don't want to indicate that the support of the `X-Auth-Request-Redirect` header is only for oauth2_proxy.
Another concern is that we wonder if this enhancement could be even more generic to cover usage of other headers.

What do you think ?

rtribotte:
Hello @pieter-lazzaro,

As i already mentioned it, we think that this feature would be too specific, and without a more generic approach that something we don't want to introduce as a core feature and I'm going to close this pull request, unfortunately.

Nevertheless, it would be great if you can build this feature as a plugin.
The plugin system has been introduced since the v2.3RC, and you can check the [demo repo plugin](https://github.com/containous/plugindemo) to learn how to build your own plugin.
One of the advantages of the plugin system is that we can get an indication of a middleware popularity and usage, which could lead to reconsider including it as a core feature.

Thank you for your time and interest in Traefik!

PR #6847, opened 2020-05-25
Fix healthcheck.interval in docs

OndrejIT:
### What does this PR do?

Fixing "time: missing unit in duration 10"
Because this feature used time.ParseDuration library .

### Motivation

:-)

### More

- [ ] Added/updated tests


mpl:
As a doc fix, I think this should be merged against the latest branch (v2.2).

PR #6857, opened 2020-05-27
Add HTTP Provider

### LINKED ISSUE #1053: Feature request: HTTP backend

dtomcej:
### What does this PR do?

This PR introduces a new provider (http) that polls an endpoint for a published configuration, and provides it to the server for use.

### Motivation

Fixes #1053

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/providers/overview.md
+++ b/docs/content/providers/overview.md
@@ -35,9 +35,10 @@ Below is the list of the currently supported providers in Traefik.
 | [Rancher](./rancher.md)               | Orchestrator | Label                      |
 | [File](./file.md)                     | Manual       | TOML/YAML format           |
 | [Consul](./consul.md)                 | KV           | KV                         |
-| [etcd](./etcd.md)                     | KV           | KV                         |
+| [Etcd](./etcd.md)                     | KV           | KV                         |
 | [Redis](./redis.md)                   | KV           | KV                         |
 | [ZooKeeper](./zookeeper.md)           | KV           | KV                         |
+| [Http](./http.md)                     | Manual       | JSON format                |
 
 !!! info "More Providers"
 
--- a/docs/content/providers/overview.md
+++ b/docs/content/providers/overview.md
@@ -38,7 +38,7 @@ Below is the list of the currently supported providers in Traefik.
 | [Etcd](./etcd.md)                     | KV           | KV                         |
 | [Redis](./redis.md)                   | KV           | KV                         |
 | [ZooKeeper](./zookeeper.md)           | KV           | KV                         |
-| [Http](./http.md)                     | Manual       | JSON format                |
+| [HTTP](./http.md)                     | Manual       | JSON format                |
 
 !!! info "More Providers"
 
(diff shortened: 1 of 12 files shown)


docs/content/providers/overview.md
```diff
@@ -35,9 +35,10 @@ Below is the list of the currently supported providers in Traefik.
 | [Rancher](./rancher.md)               | Orchestrator | Label                      |
 | [File](./file.md)                     | Manual       | TOML/YAML format           |
 | [Consul](./consul.md)                 | KV           | KV                         |
-| [etcd](./etcd.md)                     | KV           | KV                         |
+| [Etcd](./etcd.md)                     | KV           | KV                         |
 | [Redis](./redis.md)                   | KV           | KV                         |
 | [ZooKeeper](./zookeeper.md)           | KV           | KV                         |
+| [Http](./http.md)                     | Manual       | JSON format                |
```
notsureifkevin, line 41: https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol

NIT: I noticed we use `Http` and `HTTP` interchangeably throughout the documentation and I recommend the naming standard as noted in the RFC should be used for consistency.

docs/content/providers/overview.md
```diff
@@ -35,9 +35,10 @@ Below is the list of the currently supported providers in Traefik.
 | [Rancher](./rancher.md)               | Orchestrator | Label                      |
 | [File](./file.md)                     | Manual       | TOML/YAML format           |
 | [Consul](./consul.md)                 | KV           | KV                         |
-| [etcd](./etcd.md)                     | KV           | KV                         |
+| [Etcd](./etcd.md)                     | KV           | KV                         |
```
notsureifkevin, line 38: while `etcd` has no RFC, it is commonly referred to as `etcd` by the publishers of the project. You can also find examples all over the web where the `e` is capitalized, however, I noticed we use `Etcd` consistently throughout the documentation so this makes sense.

notsureifkevin:
If an operator wants to utilize this with a `https` endpoint that doesn't have a valid cert, for any number of reasons, I'm assuming this would fail.

Should we introduce a flag which allows bypassing a cert check?

docs/content/providers/overview.md
dtomcej: Also when used as a proper noun, usually capitalized.

dtomcej:
Closing in favour of https://github.com/containous/traefik/pull/6976

PR #6858, opened 2020-05-27
Document how to use multiple rules in docker label 'rule'

TeNNoX:
### What does this PR do?

Document how to use multiple rules in docker label 'rule' .

### Motivation
It took me a while and only stackoverflow could help me in the end:
https://stackoverflow.com/questions/44232354/define-host-and-path-frontend-rule-for-traefik

### More

### Additional Notes

I'm not sure if the only way to do this is in the `Host:domain.com` notation, but ``Host(`..`) &&`` did not work.

Also, I don't know how to do OR rules in this notation.

### DIFF
--- a/docs/content/routing/providers/docker.md
+++ b/docs/content/routing/providers/docker.md
@@ -159,6 +159,11 @@ For example, to change the rule, you could add the label ```traefik.http.routers
     ```yaml
     - "traefik.http.routers.myrouter.rule=Host(`example.com`)"
     ```
+    
+    Separate multiple rules with a semicolon:
+    ```yaml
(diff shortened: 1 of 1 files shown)


dtomcej:
Hello @TeNNoX,

The v1 rules are documented here:
https://docs.traefik.io/v1.7/basics/#combining-multiple-rules

The v2 rules are documented here:
https://docs.traefik.io/v2.2/routing/routers/#rule

Unfortunately this PR is adding v1 documentation that is no longer accurate. For this reason, we are going to decline this PR.

TeNNoX:
Whoops. I'm using traefik 1 and wasn't aware of it. Sorry :facepalm:

PR #6863, opened 2020-05-29
refactor X-Forwarded-Proto

jcgruenhage:
## What does this PR do?

This code makes the X-Forwarded-Proto code path a tiny bit faster, more robust, and IMO more readable.

### Motivation

I looked at the code and it took me a while to understand it, so I tried making it easier to understand

### Additional Notes

The unified diff is pretty useless here, look at a split diff instead.


juliens:
Thanks for your PR!

I agree with the refactoring of this part of the code, but one needs to understand that it changes the current behavior.

If you have another reverse-proxy in front of Traefik that configures X-Forwarded-Proto with "http" or "https" even if it's a websocket request, today, Traefik will change the X-Forwarded-Proto to "wss" or "ws", but with your change, the X-Forwarded-Proto would stay unchanged.

WDYT @SantoDE @jbdoumenjou @mpl ?

jcgruenhage:
Yes, that's true. I didn't think of that when writing the description for the PR, definitely worth mentioning.

In my opinion, a proxy that has another proxy in front of it should *not* overwrite the X-Forwarded-* headers written by the first proxy. For all the other X-Forwarded-* headers, this is already done, and for the X-Forwarded-Proto header, it is also done, as long as it's not a websocket request. Special casing websocket requests here seems wrong to me.

SantoDE:
I agree with your argument @jcgruenhage. I think, if there is another proxy in front which might to termination and sets the headers, the second proxy should not overwrite.

So I agree.

jbdoumenjou:
As long as the headers are put from a trusted source (or that we don't care), it seems logical to me not to overwrite them. So I also agree.
And I also think that if we apply it, we should clearly inform users of this "new" behavior.

jcgruenhage:
@jbdoumenjou Anything I need to do for that? Include something for the changelog in my PR maybe?

jbdoumenjou:
@jcgruenhage , I am not sure, if we consider it as a bug fix, we have nothing to do.
@SantoDE @juliens do we considered this PR as a bugfix? If we do so, we have to rebase it to the v2.2 branch ;)

SantoDE:
Oh, I'm not really sure on that. I could live with both to be honest. Wdyt @juliens?

PR #6875, opened 2020-06-02
feat: use parser to load dynamic config from file.

ldez:
### What does this PR do?

Uses parser to load dynamic config from file.

### Motivation

Be able to use `SetDefault` method.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/file/file.go
+++ b/pkg/config/file/file.go
@@ -26,13 +26,13 @@ func Decode(filePath string, element interface{}) error {
 		return err
 	}
 
-	metaOpts := parser.MetadataOpts{TagName: parser.TagFile, AllowSliceAsStruct: true}
+	metaOpts := parser.MetadataOpts{TagName: parser.TagFile, AllowSliceAsStruct: false}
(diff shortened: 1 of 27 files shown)


pkg/config/file/file.go
```diff
@@ -30,3 +34,49 @@ func Decode(filePath string, element interface{}) error {
...
+	node, err := decodeRawToNode(data, parser.DefaultRootName, filters...)
+	if err != nil {
+		return err
+	}
+
+	if len(node.Children) == 0 {
+		return nil
+	}
+
+	metaOpts := parser.MetadataOpts{TagName: "noop", AllowSliceAsStruct: false}
```
jlevesy, line 75:
```suggestion
	metaOpts := parser.MetadataOpts{TagName: parser.TagLabel, AllowSliceAsStruct: false}
```

Does it makes sense to do something like this ?

ldez: nope it's the expected tag name.

Issue #6047, opened 2019-12-17
Traefik header transformation


llacroix:
### Feature:

Header transformation rules using a predefined set of operators

### Syntax:

Something like this:

`traefik.http.middlewares.<middleware>.headers.request_transform = op1, op2, op3`
`traefik.http.middlewares.<middleware>.headers.response_transform = op1, op2, op3`

Or possibly:

`traefik.http.middlewares.<middleware>.headers.transform.operations = op1, op2, op3`
`traefik.http.middlewares.<middleware>.headers.transform.type = 'request' or 'response'`

### Goal:

Add a toolset that lets traefik transform headers. Currently in traefik it's possible to do basic operations of headers like adding,removing headers from a request or response but the support for manipulation of headers is quite simplistic. It doesn't really let the users to modify/transform header by changing the content or the name of the headers.

#### Rename(header, new_header) 

Rename a header by matching an exact `header` to a `new_header`. 

Example:

    Rename(`MY_HEADER`, `NEW_HEADER`)

Would change:

    MY_HEADER: super
    OTHER_HEADER: 123

To:

    NEW_HEADER: super
    OTHER_HEADER: 123

#### HeaderRewriteRule(regex, replace)

 Rewrite header name using a regex with a match/capture. 

Example:

    HeaderRewriteRule(`MY_(.*)`, `NEW_$1`)

Would change the following requests from:

    MY_HEADER: super
    MY_OTHER_HEADER: 123

To:

    NEW_HEADER: super
    NEW_OTHER_HEADER: 123

#### ValueRewriteRule(header, regex, replace)

Same as the header rewrite but for their content, header is a parameter to match an header or a set of headers to apply the rewrite rule.

#### Set(header, value)

Set a value to a header name

#### Remove(header)

Remove an header

#### Join(header, sep, value/header)

Possibility to join multiple headers inside one.
Example:

    Set(`PROXY_PATH`, `129.2.3.4`), Join(`PROXY_PATH`, `,`, `127.0.0.1`)

Would result with this header:

    PROXY_PATH: 129.2.3.4,127.0.01

As you can see, the order of the operators change the context of evaluation so op1 can change the context in a way that op2 will be able to transform headers that weren't present in the original response or request.  I guess that having some generic ways to manipulate headers would render some of the existing header configuration in the middleware useless. Or at least they could become simple shortcuts. 

[...]

PR #6880, opened 2020-06-03
Headers middleware - prefix cookie path

sparkoo:
### What does this PR do?

Adds support for add prefix to cookie path on response `Set-Cookie` header.

Configuration in current state of the PR looks like this:
```
middlewares:
  cookiePrefixMiddleware:
    headers:
      prefixCookiePath: "/prefix"
```

I would like to discuss here:
 - is there some possible solution to workaround this issue ?
 - are you open to accept this feature ?
 - is it ok as a part of `headers` middleware or should it be in separate middleware? (AFAIK that's the only one that can modify response headers at the moment)
 - is configuration ok in this very straightforward way, or should it be something more general?

### Motivation

We need this feature because we're using `stripPrefix` and then cookies with explicitly set path does not work as server does not know full path. Browser then can't use this cookie as it's set for different path. I've wrote about it on your discussion forum  https://community.containo.us/t/how-to-rewrite-set-cookie-path/6277

### More

- [x] Added/updated tests

### Additional Notes

I know about there 2 issues that would potentially fix our problem.
https://github.com/containous/traefik/issues/1336
https://github.com/containous/traefik/issues/6047
However, considering their complexity and activity there, I don't see it coming anytime soon (next few releases). Also my proposal is more direct and straightforward, so it might be good to have it separate from these.

Thank you in advance for any feedback!

### DIFF
--- a/pkg/config/dynamic/middlewares.go
+++ b/pkg/config/dynamic/middlewares.go
@@ -150,6 +150,7 @@ type ForwardAuth struct {
 type Headers struct {
 	CustomRequestHeaders  map[string]string `json:"customRequestHeaders,omitempty" toml:"customRequestHeaders,omitempty" yaml:"customRequestHeaders,omitempty"`
 	CustomResponseHeaders map[string]string `json:"customResponseHeaders,omitempty" toml:"customResponseHeaders,omitempty" yaml:"customResponseHeaders,omitempty"`
+	PrefixCookiePath      string            `json:"prefixCookiePath,omitempty" toml:"prefixCookiePath,omitempty" yaml:"prefixCookiePath,omitempty"`
 
 	// AccessControlAllowCredentials is only valid if true. false is ignored.
 	AccessControlAllowCredentials bool `json:"accessControlAllowCredentials,omitempty" toml:"accessControlAllowCredentials,omitempty" yaml:"accessControlAllowCredentials,omitempty"`
@@ -195,6 +196,10 @@ type Headers struct {
 func (h *Headers) HasCustomHeadersDefined() bool {
 	return h != nil && (len(h.CustomResponseHeaders) != 0 ||
 		len(h.CustomRequestHeaders) != 0)
+}
+
+func (h *Headers) HasCookiePrefix() bool {
+	return h.PrefixCookiePath != ""
 }
 
 // HasCorsHeadersDefined checks to see if any of the cors header elements have been set.
--- a/pkg/middlewares/headers/headers.go
+++ b/pkg/middlewares/headers/headers.go
@@ -45,8 +45,9 @@ func New(ctx context.Context, next http.Handler, cfg dynamic.Headers, name strin
 	hasSecureHeaders := cfg.HasSecureHeadersDefined()
 	hasCustomHeaders := cfg.HasCustomHeadersDefined()
 	hasCorsHeaders := cfg.HasCorsHeadersDefined()
+	hasCookiePrefix := cfg.HasCookiePrefix()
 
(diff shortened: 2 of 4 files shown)


sparkoo:
@mpl any eta when you get to this? I would like to hear any feedback to questions in the PR description.

l0rd:
@emilevauge @ldez can we have someone to review this PR? We are considering to include Traefik as a component of Eclipse Che and Red Hat CodeReady Workspaces. But we need cookie path rewrite.

ldez:
Hey @l0rd, good to see you here!

Our process is described here: https://docs.traefik.io/v2.2/contributing/maintainers/#pr-review-process

The next step will be a design review by the maintainer assign to this PR.

We will trying to review as soon as possible.

mpl:
@sparkoo Hi, we are not convinced that this PR is necessary for your use-case, since the information about the path that was stripped is not actually lost: when the stripprefix middleware does its thing, it also sets the X-Forwarded-Prefix header. So your backend server only has to get the value from this header to reconstruct the correct full path that should be set in the cookie.

sparkoo:
@mpl that's not satisfying solution for us. It would make our code aware of deployment, and we have deployment types without reverse-proxy as well. Ideally, we would like this to be invisible to the backend as much as possible. Also as Eclipse Che is cloud IDE, users are running their custom applications on top of it, so we would need to force them to reflect this as well. This is quite heavy limitation for our customers.

mpl:
@sparkoo Sorry, but we're still not convinced. Similarly to how a backend should properly take into account X-Forwarded-For to know where a request comes from, we think a backend should also properly take into account X-Forwarded-Prefix to know how the prefix path was handled. Especially given how easy it is to do the job properly for cookies with X-Forwarded-Prefix.

That said, if you feel strongly about being able to modify headers in general, as you have already noticed there already is https://github.com/containous/traefik/issues/6047 , which we think is the best avenue for making something like that happen.

cdauth:
I think it is a pity that this pull request was not accepted. In my opinion this is a basic feature of a reverse proxy, and there is a good reason why other reverse proxies have a directive for this that appears in every example configuration (for example [proxy_cookie_path](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_cookie_path) in nginx and [ProxyPathReverseCookiePath](https://httpd.apache.org/docs/2.4/mod/mod_proxy.html#proxypassreversecookiepath) in Apache).

Most web applications are not written with being run behind a reverse proxy in mind. And those that are will usually not have support for the `X-Forwarded-Prefix` header built in. For example, PHP applications usually use `session_start()` to set a session ID cookie, which sets the path to `/`, unless manually configured to use another path. It is unpractical to ask the maintainers of PHP itself and/or or the many PHP applications to add support for running behind a traefik proxy.

sparkoo:
@cdauth Yes, I think that proxy should have tools to do this. However, I think your best bet now is to use Traefik plugin https://pilot.traefik.io/plugins. Maybe there is already one that can do that. Or you may write your own, it's not very complicated.

cdauth:
@sparkoo But for that I need the enterprise version of traefik, right? At least according to the [pricing page](https://traefik.io/pricing/) it is needed for using plugins. That's not really an option for a non-commercial setup...

ldez:
Plugins are available for free and the plugins are free and open-source https://doc.traefik.io/traefik/v2.6/plugins/

sparkoo:
yes, its free. with private plugins you dont even need any registration, you just put the code to filesystem and configure its path in traefik.

cdauth:
That's good to know. Either I misunderstood the pricing page or it's outdated.

Issue #6851, opened 2020-05-26
Subdomain wildcard CORS


davedray:
### Do you want to request a *feature* or report a *bug*?
Feature

Feature

### What did you expect to see?

I would like to be able to define a wildcard subdomain for CORS:

```
apiVersion: traefik.containo.us/v1alpha1
kind: Middleware
metadata:
  name: testHeader
spec:
  headers:
    accessControlAllowMethods:
      - "GET"
      - "OPTIONS"
      - "PUT"
    accessControlAllowOriginList:
      - "https://*.bar.org"
    accessControlMaxAge: 100
    addVaryHeader: "true"
```

and then have it allow `foo.bar.org` as well as `baz.bar.org` for example.

We have a scenario where we are allowing vanity urls for certain clients, and they all hit the same API.  We either have to move CORS back into our api and out of traefik, do something hacky with forward auth, or (ideally) solve it in Traefik, but I don't want to work on the PR if it's unlikely to get merged in for whatever reason.

If it's a feature that (assuming the code was fine, of course) would be merged in I'm happy to do the work.

dtomcej:
I think that this would be a welcome feature for people that have corporate domains that may serve hundreds of subdomains.

Manually keeping an up-to date list could be challenging.

Please feel free to open a PR.

barthez:
I was considering similar report, but I would rather handle regular expressions than wildcards. There is a question how should the configuration look like, is it going to be a list of strings that headers middleware would treat as wildcards/regexps or there should be a way to provide a type of allowed origin matcher:

```
accessControlAllowMethods:
  - "http://plain.string.origin.com"
  - type: wildcard
    value: "http://*.wildcard.origin.com"
  - type: regexp
    value: "\Ahttps?://(subdomains|list)\.domain\.net\z"
```

jodosha:
@davedray @dtomcej I followed @barthez suggestion and created a PR to allow regular expressions to be used here: https://github.com/containous/traefik/pull/6881 Please let me know what you think there.

almereyda:
Is it possible that #6881 validated this ticket, which could be closed?

There is documentation about the feature, dubbed `accessControlAllowOriginListRegex`, in:

- https://doc.traefik.io/traefik/middlewares/http/headers/#accesscontrolalloworiginlistregex

rtribotte:
Hello @almereyda,

Thanks, you are right, we missed closing it when we merged #6881.

Closed by #6881

PR #6881, opened 2020-06-03
Allow to use regular expressions for `AccessControlAllowOriginList`

jodosha:
## What does this PR do?

This PR allows to use regular expressions to be used as `AccessControlAllowOriginList` value.

### Motivation

Regular expressions brings more flexibility to the configuration. We're working on a SOA environment that requires temporary hosts to be spun for development and demo purposes. Having a fixed list of hosts requires to frequently update Traefik configuration.

See also https://github.com/containous/traefik/issues/6851

### More

- [x] Added/updated tests

### Additional Notes

Because `AccessControlAllowOriginList` is of type `[]string`, I leveraged `regexp.MatchString(pattern string, s string)` (the utility function from the `regexp` package) instead of `Regexp.MatchString(s string)` (the exported function from `Regexp` type).

In this way the configuration can define regular expressions as strings. This works with the current code type structure. It also avoids to compile regexps, but at the same time it handle malformed ones (see the tests).

However, there is a limitation regarding `regexp.MatchString`. According to its [documentation](https://golang.org/pkg/regexp/#MatchString):

> More complicated queries need to use Compile and the full Regexp interface.

But I believe that this is a problem that we won't hit.

### DIFF
--- a/pkg/middlewares/headers/header.go
+++ b/pkg/middlewares/headers/header.go
@@ -3,6 +3,7 @@ package headers
 import (
 	"context"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -163,6 +164,10 @@ func (s *Header) isOriginAllowed(origin string) (bool, string) {
 	for _, item := range s.headers.AccessControlAllowOriginList {
 		if item == "*" || item == origin {
 			return true, item
+		}
+
+		if match, _ := regexp.MatchString(item, origin); match {
+			return true, origin
 		}
 	}
 
--- a/pkg/middlewares/headers/header.go
+++ b/pkg/middlewares/headers/header.go
@@ -2,6 +2,7 @@ package headers
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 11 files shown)


jodosha:
On a second thought (credit goes to @barthez), the current proposed solution can be a potential breaking change / security problem.

Imagine an **existing** company that uses a fixed host as a value for `AccessControlAllowOriginList`, for instance: `https://foo.bar.co`, if an incoming request that sets `Origin` to `https://foo.bar.company` the existing match fails because the two strings are different, but because of this PR, it will try the regular expression fallback, which it will pass, even if it shouldn't.

You can have a demonstration, by applying this patch to the PR:

```diff
diff --git a/pkg/middlewares/headers/headers_test.go b/pkg/middlewares/headers/headers_test.go
index 0b751f3e..1ed87a9a 100644
--- a/pkg/middlewares/headers/headers_test.go
+++ b/pkg/middlewares/headers/headers_test.go
@@ -412,13 +412,13 @@ func TestCORSResponses(t *testing.T) {
 		{
 			desc: "Test Simple Request",
 			header: NewHeader(emptyHandler, dynamic.Headers{
-				AccessControlAllowOriginList: []string{"https://foo.bar.org"},
+				AccessControlAllowOriginList: []string{"https://foo.bar.co"},
 			}),
 			requestHeaders: map[string][]string{
-				"Origin": {"https://foo.bar.org"},
+				"Origin": {"https://foo.bar.company"},
 			},
 			expected: map[string][]string{
-				"Access-Control-Allow-Origin": {"https://foo.bar.org"},
+				"Access-Control-Allow-Origin": {"https://foo.bar.company"},
 			},
 		},
 		{
```

---

I believe that the PR cannot be merged as it is, but I'm looking for maintainers advice on how to proceed. Thanks in advance.

jodosha:
👋  Do you have any news? Thanks in advance 😄

jbdoumenjou:
Hello @jodosha,

Thanks for your contribution!
We prefer to have a new option such as "AccessControlAllowOriginListRegex", to avoid having a breaking change and imposing a potential performance loss for users who do not have this need.

What do you think?

ToGo101:
Just want to leave a +1 here for this feature.

jbdoumenjou:
:wave: @jodosha, do you think you will have some time to work on it ? 
Thanks

pkg/middlewares/headers/header.go
```diff
@@ -166,5 +167,16 @@ func (s *Header) isOriginAllowed(origin string) (bool, string) {
 		}
 	}
 
+	for _, item := range s.headers.AccessControlAllowOriginListRegex {
+		matched, err := regexp.MatchString(item, origin)
+		if err != nil {
+			log.WithoutContext().Debugf("an error occurred during origin parsing: %v", err)
```
kevinpollet, line 173: Maybe it's better to log an `Error` here, WDYT?

```suggestion
			log.WithoutContext().Errof("Unable to match origin: %v", err)
```

pkg/middlewares/headers/header_test.go
```diff
@@ -267,6 +267,30 @@ func TestNewHeader_CORSResponses(t *testing.T) {
 				"Access-Control-Allow-Origin": {"*"},
 			},
 		},
+		{
+			desc: "Regexp Origin Request",
```
kevinpollet, line 271: Maybe here we could add a test for a non-matching origin, wdyt?

kevinpollet (changes requested):
Just minor comments 👍

jbdoumenjou (approved):
LGTM

kevinpollet (approved):
LGTM, Nice work 👍

rtribotte (approved):
LGTM 👍

jodosha:
Thank you all. ❤️

codefriar:
@jodosha - I don't suppose you have an example of this working in a toml file? I've tried this:

```
    cors-headers:
      headers:
        accessControlAllowMethods:
          - GET
          - HEAD
          - POST
        accessControlMaxAge: 100
        addVaryHeader: true
        accessControlAllowOriginListRegex:
          - https://.*\.foo\.com$
 ```

Yet this doesn't seem to be working :(

jbdoumenjou:
Hello @codefriar,

For help on your configuration, please join our [Community Forum](https://community.traefik.io/) and reach out to us on the [Traefik section](https://community.traefik.io/c/traefik). :smiley:

PR #6900, opened 2020-06-08
fix: access logs header names filtering is case insensitive

### LINKED ISSUE #6837: Access logs header filtering is case sensitive, headers are case insensitive per RFC

Bug

### What did you do?

Specifying a header retention, skip or redaction spec in the Access Logs filtering section of [the config](https://docs.traefik.io/observability/access-logs/#limiting-the-fields) is case-sensitive to the header as received by Traefik. Per [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2), header names are case insensitive and many tools now prefer lowercase header name specification to reflect this. Regardless, Traefik should normalize the header name to account for a mismatch between the header as listed in the config and the header name received over HTTP.

Specifying a filter for `x-forwarded-for` should match against `X-Forwarded-For`, and other case-mismatched variants.

### What did you see instead?

The header name must match the casing of the received header to be included in the filter set.

Traefik 2.2.1

```toml
[accessLog]
  format = "json"
  [accessLog.filters]
    ...
    [accessLog.fields.headers]
      defaultMode = "drop"
        "User-Agent" = "keep"
        "Content-Type" = "keep"
        "Cf-Connecting-Ip" = "keep"
        "x-forwarded-for" = "keep"
```

mjeanroy:
### What does this PR do?

This PR fix the bug described in issue #6837 (when filtering access logs, header name check is now case insensitive).

Fixes #6837

### Motivation

To be honest, I was looking for a "good first issue" since I wanted to take the time to learn how Traefik works and contributing to it seems a good way to learn ;) 

The issue looked like a good one to start with, please let me know.
If my fix is not appropriate (or the bug should not have been fixed at all), feel free to close it.
If you want me to update it according to your guidelines, please let me know and I will be more than happy to fix it!

### More

- [X] Added/updated tests

### Additional Notes

Traefik is awesome ❤️


emilevauge:
👋 @mjeanroy 
Nice to see you around :)

pkg/types/logs.go
```diff
@@ -95,10 +99,21 @@ func (f *AccessLogFields) KeepHeader(header string) string {
...
 
+		// Keep fast path in case of exact match.
 		if v, ok := f.Headers.Names[header]; ok {
 			return checkFieldHeaderValue(v, defaultValue)
 		}
+
+		// Check for case insensitive header name.
+		caseInsensitiveHeader := strings.ToLower(header)
+
+		for h, v := range f.Headers.Names {
```
dtomcej, line 110: instead of looping over all header names, could you not do a lookup like above?

```
		if v, ok := f.Headers.Names[caseInsensitiveHeader]; ok {
			return checkFieldHeaderValue(v, defaultValue
		}
```
?

mjeanroy: That was my first idea, using a map for `Headers.Names` with lowercase keys, so a simple lookup in the map would be enough. It probably needs a lot more changes.

I thought that a simple solution would be enough since I was not sure that using a map would be really faster: I don't know how many headers an http request may contain (on average, or in the worst case), but I guess it's should not be too much (maybe 20 our 30 at most, so looping over 30 elements should be fast).

Would you like me to implement it using a map instead?

SantoDE: That would be a good point I guess, yes :)

dtomcej: @mjeanroy I totally understand that simple is better, however, this method (Keepheader()) gets called for each header in the request...So if you have 30 headers, each one is looped each time, meaning that there are 900 comparisons instead ;)

I know its a pain in the butt to think about this sort of thing, but its far easier to review and update these things early, than  trying to debug logging performance issues down the road 😄

mjeanroy: Hi @dtomcej, that's a very good point!
Thanks for the review, I'll do the changes asap ;)

mjeanroy:
@rtribotte Sorry, I did not have the time yet to implement the changes discussed with @dtomcej, so thanks for fixing it ;) 
Anyway, feel free to ping me if you want me to do something else, I'll try to find the time to do it ;)

PR #6904, opened 2020-06-08
Fix v1-> v2 migration: unify domain name in documentation example

sinacek:
What does this PR do?

Unify domain name in migration v1->v2 documentation.
Motivation

Have compact and clear examples.

### DIFF
--- a/docs/content/migration/v1-to-v2.md
+++ b/docs/content/migration/v1-to-v2.md
@@ -97,7 +97,7 @@ Then any router can refer to an instance of the wanted middleware.
 
     ```yaml tab="Docker"
     labels:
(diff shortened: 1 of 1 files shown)


jbdoumenjou:
Hello @sinacek ,

the domain "example.com" has been added by #6471  to follow good documentation practices.
The correction would be to change "test.localhost" to "example.com".
Would you agree to do it?

sinacek:
Hello @jbdoumenjou,
I don't care which one will be there. I just want to unify it to the same value. I send PR with `example.com` -> `test.localhost`, because `test.localhost` is more often use within this Example 1.

Screenshot 2020-06-09 at 10 30 53
Screenshot 2020-06-09 at 10 35 22
Screenshot 2020-06-09 at 10 35 28
Screenshot 2020-06-09 at 10 35 12

PR #6911, opened 2020-06-10
Update Let's Encrypt documentation about resolvers for certificateresolvers

tomwiggers:
### What does this PR do?

Update documentation about certificatesresolvers to point out that domain names are not supported as resolvers.

### Motivation

I had to find out the hard way that domain names are not supported. I migrated my domain from Cloudflare toDdigitalOcean and thus had to update my certificatesresolvers with the new config and new API auth token, etc etc. After a lot of time struggling and failing to get it to work. I came up with the idea to try an IP address instead of a domain name, and it actually worked.

### Additional Notes

Feel free to change the text I put. I just wanted to point this out. Please excuse me if this is already documented elsewhere and I looked over it.

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -371,7 +371,7 @@ For example, `CF_API_EMAIL_FILE=/run/secrets/traefik_cf-api-email` could be used
 
 #### `resolvers`
 
-Use custom DNS servers to resolve the FQDN authority.
+Use custom DNS servers to resolve the FQDN authority. Only IP addresses are supported as resolvers, not domain names.
 
 ```toml tab="File (TOML)"
 [certificatesResolvers.myresolver.acme]


jbdoumenjou:
Hi @tomwiggers,
you're assertion is not true.
But you need to use a domain name that your system is able to resolve. (cf https://github.com/go-acme/lego/blob/fb72e1a292f256c85fade0f4ae146f5f810efbdf/challenge/dns01/nameserver.go#L24-L27)

tomwiggers:
I am sure my system is able to resolve ns1.digitalocean.com, I just verified this using nslookup.

notsureifkevin:
> 
> 
> I am sure my system is able to resolve ns1.digitalocean.com, I just verified this using nslookup.

It's possible that within the Traefik container it's unable to resolve domain names. Traefik does not come with network tools, or even a shell, so you may want to try and spin up busybox and check to make sure that your containers can do DNS lookups.

tomwiggers:
The traefik container is also able to resolve domain names. The `nslookup` command is available inside the container and it successfully resolves. Bash is not available, no, but `sh` is.

Anyway if the container did not come with networking tools, then why would the default be DNS-based nameservers? As @jbdoumenjou pointed out.

notsureifkevin:
> The traefik container is also able to resolve domain names. The nslookup command is available inside the container and it successfully resolves. Bash is not available, no, but sh is.

I just double checked and this is accurate. I was referring to command line networking tools, it appears the code @jbdoumenjou referenced leverages the go library https://github.com/miekg/dns

In any case, it seems peculiar that the default lookups use FQDN, while you're unable to.

notsureifkevin:
https://github.com/containous/traefik/tree/v2.2.1/integration/fixtures/acme

I didn't see where custom DNS servers being used in these tests, and it's possible this case isn't covered - but I spent ~5 minutes looking, so this may be something worth investigating further.

zhangbaoheng:
test

PR #6922, opened 2020-06-15
Add custom fallback to circuit breaker

sylr:
### What does this PR do?

This is a modest attempt to add the possibility to define a service as fallback to the circuit breaker middleware.

Any pointers to improve it are welcomed.

### More

- [ ] Added/updated tests


go.mod
```diff
@@ -107,3 +107,5 @@ replace (
 	github.com/mailgun/minheap => github.com/containous/minheap v0.0.0-20190809180810-6e71eb837595
 	github.com/mailgun/multibuf => github.com/containous/multibuf v0.0.0-20190809014333-8b6c9a7e6bba
 )
+
+replace github.com/vulcand/oxy => github.com/sylr/oxy v1.1.1-0.20200616103808-a0437a24ad17
```
sylr, line 111: This is only temporary.

sylr:
Could you tell me if you would consider merging this so that I know whether or not I should go ahead and start updating documentation.

go.mod
sylr: Depending on https://github.com/vulcand/oxy/pull/201

SantoDE:
Hello @sylr,

Thanks for contributing that PR.

Prior to a code-review, we are wondering about the actual use-case that would be needed for? Can you give us some thoughts on that? How should / would we handle situations, where the fallback service is down / or not properly configured? Should Traefik then only ignore the circuit-breaker or the whole service configuration?

sylr:
In my case we have implemented a service which serves error pages, computes stats and sends requests in kafka.

This services is reached when no route is matched and we wanted the circuit breaker to be able to reach this service as well when it is tripped.

Regarding the behavior if the fallback services is down, I'm not sure right know what should be the best thing to do.

SantoDE:
Hello @sylr,

thanks for your feedback. In that case, as we're also not sure about those questions we will close that PR. 

If you're still interested in that topic, please raise an issue first so we can discuss all the details prior to a second attempt of implementation.

Issue #5614, opened 2019-10-09
Metric for expiration of certificates


sylr:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

nginx ingress controller has a metric (`nginx_ingress_controller_ssl_expire_time_seconds`) for each certificates with the time as unix timestamp which tells when a certificate will expire. This is quite useful to setup alerts to not forget to renew certificates.

PR #6924, opened 2020-06-16
Add TLS certs expiration metric

sylr:
### What does this PR do?

Adds a metrics with the expiration timestamp of the certificates.

This is not my most elegant work, if you have pointers to make this in a proper manner, they are very welcome.

### Motivation

Replicate `nginx-ingress-controller` which does that and therefor allow alerting on certificates about to expire. Fixes https://github.com/containous/traefik/issues/5614

### More

- [ ] Added/updated tests


sylr:
Could I have some feedback about this ?

pkg/tls/tlsmanager.go
```diff
@@ -113,12 +124,16 @@ func (m *Manager) Get(storeName string, configName string) (*tls.Config, error)
...
+				m.updateMetrics(cert)
 				return cert, nil
 			}
 		}
 
 		bestCertificate := store.GetBestCertificate(clientHello)
+
 		if bestCertificate != nil {
+			m.updateMetrics(bestCertificate)
+
```
rtribotte, line 136: Hello @sylr,

Here, i think the updateMetrics calls must not be done on the critical path more appropriately each time the certificate_store stores a certificate (L84), since the metric should change only when the certs are updated.
It also has to be call when configuring the default certificate if it is not generated.

WDYT ?

rtribotte (changes requested):

pkg/tls/tlsmanager.go
sylr: Hi @rtribotte, sorry for the late reply byt I'm currently on vacation.

I'll review your proposition when I get back.

Thanks.

rtribotte:
Hello @sylr,

After a second look, i think that would be better to not pass the gauge to the tls manager and update the metric directly into the listener func:
https://github.com/traefik/traefik/blob/d38d11f02e47a1e81edfecc6fcf4a1c42390a81e/cmd/traefik/traefik.go#L272-L275

Also, I wonder why don't you have implemented the metric for other providers than prometheus ?

sylr:
I took a look and although I agree it would be better looking it implies replicating a lot of the logic implemented in the `UpdateConfigs()` method and I don't think it is advised.

rtribotte:
Hello @sylr,

Thanks for your feedback, i think you are right, even if i'm not perfectly fine with that.
Having said that, i now think that it would be better to pass the metricsRegistry to the tlsManager during the construction, instead of setting the gauge.

sylr:
I've tried to pass the metricRegistry to `traefiktls.NewManager` but unfortunately it generates a circular dependency.

```
package github.com/traefik/traefik/v2/cmd/traefik
	imports github.com/traefik/traefik/v2/cmd
	imports github.com/traefik/traefik/v2/pkg/config/static
	imports github.com/traefik/traefik/v2/pkg/provider/acme
	imports github.com/traefik/traefik/v2/pkg/config/dynamic
	imports github.com/traefik/traefik/v2/pkg/tls
	imports github.com/traefik/traefik/v2/pkg/metrics
	imports github.com/traefik/traefik/v2/pkg/config/dynamic: import cycle not allowed
```

To circumvent that problem I had to move the metrics.Registry interface in its own package with no references to `github.com/traefik/traefik/v2/pkg/config/dynamic`.

rtribotte:
Hello @sylr,

I agree with you, i didn't see that coming !
Thus, i would prefer to redefine the Registry interface inside the tlsManager, instead of moving it in its own package:
```go
type Registry interface {
	TLSCertsNotAfterTimestampGauge() metrics.Gauge
}
```

WDYT ?

pkg/metrics/prometheus.go
```diff
@@ -28,6 +28,10 @@ const (
 	configLastReloadSuccessName    = metricConfigPrefix + "last_reload_success"
 	configLastReloadFailureName    = metricConfigPrefix + "last_reload_failure"
 
+	// TLS.
+	metricsTLSPrefix          = MetricNamePrefix + "tls_"
+	tlsCertsNotAfterTimestamp = metricsTLSPrefix + "cert_not_after"
```
rtribotte, line 33: I think the naming should be either aligned with other providers or we should use the singular everywhere.
It looks like we used the singular in other gauges.

```suggestion
	tlsCertsNotAfterTimestamp = metricsTLSPrefix + "certs_not_after"
```

rtribotte (changes requested):
Hello @sylr,

Thank you for the changes and also adding other providers implementations.
The last thing that remains is the update of the unit tests for every providers.
Are you willing to add this tests ? / WDYT ?

I have also a small comment on the form below.

sylr:
Hi @rtribotte, if you can take over the unit testing part I would appreciate.

Otherwise I'll try to find some time.

rtribotte (approved):
LGTM 👍

ldez (approved):
Thanks @rtribotte :+1:

kevinpollet (approved):
LGTM 🚀

bschoenmaeckers:
When will this be released? Its not in v.2.4.2.

rtribotte:
@bschoenmaeckers It will come with the future v2.5 release

Issue #6909, opened 2020-06-10
Traefik should make sure it uses the most recent valid wildcard certificate


sylr:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

We have multiple traefik ingressroutes in multiple kubernetes namespaces which uses the `tls.secretName` property to pass a TLS certificate.

All the secrets contains the **same wildcard certificate** (I know, this is an heritage from the time we used nginx ingress controller).

While doing certificates renewal when we start updating the kubernetes secrets with the new wildcard certificate, upon the first update, all services starts to randomly use the current and new versions of the certificates.

```
$ while [ true ]; do curl -s -v https://service.company.com 2>&1 | grep "expire date"; done
*  expire date: Jun 10 11:49:56 2021 GMT
*  expire date: Jun 21 12:22:01 2020 GMT
*  expire date: Jun 10 11:49:56 2021 GMT
*  expire date: Jun 10 11:49:56 2021 GMT
*  expire date: Jun 21 12:22:01 2020 GMT
*  expire date: Jun 10 11:49:56 2021 GMT
*  expire date: Jun 10 11:49:56 2021 GMT
*  expire date: Jun 21 12:22:01 2020 GMT
```

### What did you expect to see?

At first I expected services to use and only use the version of the certificate stored in the secret referenced in each ingressroute.

But, if I'm not mistaken, it seems that those secrets are used to populate a TLS store which can be used by all routes in the instance.

Knowing that I would expect that the TLS store gives, in the case they are multiple certs with the same SANs, one which is valid (in terms of `not before` and `not after` properties) and the most recent (the one with the greater `not before` property).

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
2.2
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
# (paste your configuration here)
```

### If applicable, please paste the log output in DEBUG level (`--log.level=DEBUG` switch)

```
(paste your output here)
```

nmengin:
Hello there,

This proposal has been open for a while with minimal traction. We won't be adding it to our roadmap at this time and prefer to close it for now.

If circumstances change or interest grows, we can always reopen this later.
Thank you, @sylr , for the suggestion!

PR #6926, opened 2020-06-16
Replace certificate when given a "better" one

sylr:
### What does this PR do?

Taking a shot at https://github.com/containous/traefik/issues/6909

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/tls/certificate.go
+++ b/pkg/tls/certificate.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/containous/traefik/v2/pkg/log"
 	"github.com/containous/traefik/v2/pkg/tls/generate"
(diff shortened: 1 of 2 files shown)


sylr:
Could I have some feedback about this ?

rtribotte:
Hello @sylr,

After more thoughts on your proposal, we think that deciding which certificate is better can be done on other considerations than only comparing the boundaries. For instance, regarding key usage extensions, we can wonder if computing the "better" certificate criteria only on boundaries would lead to choose the right certificate in all cases.
In consequence, we think that the actual implementation could lead to choose the wrong certificate in some unexpected scenarios.

Unfortunately, for this reason, i will close this PR.

Anyway, thank you for your time and your interest in Traefik

sylr:
I don't understand the reasoning.

You mention other considerations than time boundaries but at the moment none are applied. Currently you could randomly end up with a invalid with wrong extensions certificate.

At least with this patch you ensure that a valid cert will be used if one is provided.

PR #6931, opened 2020-06-17
Do not log error when kubernetes endpoints are empty

sylr:
### What does this PR do?

Remove the error log when a kubernetes service endpoint subset is empty.

### Motivation

There are valid use cases where we have no subset because all resources have been voluntary scaled down but this generates a lot of log lines in Traefik.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/crd/errors.go
+++ b/pkg/provider/kubernetes/crd/errors.go
@@ -1,0 +1,29 @@
+package crd
+
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrResourceNotFound ...
+	ErrResourceNotFound = fmt.Errorf("resource not found")
+)
+
+// Error is an error wrapper with a Log property which indicates
+// whether the error should be logged or not.
+type Error struct {
+	err error
+	// Log tells wether or not this error should be logged.
+	Log bool
+}
+
+// Error implements error interface.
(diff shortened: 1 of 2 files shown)


jbdoumenjou:
@sylr ,

With this error the router creation will be avoided, the user need this error log to detect and understand why.   

Unfortunately, from the Traefik point of view, your PR seems to answer a specific need of log filtering. 

For this reason, we are going to decline this PR.

Thanks for your time.

sylr:
@jbdoumenjou For information on our production environment, with 3 traefik instances, this represent around 2000 log lines per minute:

Search___Splunk_7_2_9

So I don't believe it's a log filtering issue but rather a log spamming issue.

notsureifkevin:
> log spamming issue

are you seeing this issue when replicas are scaled down to 1 or 0?

sylr:
0

notsureifkevin:
@jbdoumenjou perhaps traefik is miscategorizing this as an error, as it's possible for kubernetes operators to scale replicasets down to zero for various legitimate reasons including, but not limited to, maintenance, cost reduction, or testing.

MohanSha:
@jbdoumenjou as @kevtainer mentioned the there are legitimate reasons for scaling down to 0, which causes a lot of log spam from traefik. Is there any plan to address this and resolve this?

PR #6941, opened 2020-06-19
Configure retry methods

nkonev:
### What does this PR do?

This PR adds new config parameter, purpose of it is setting which of HTTP methods traefik should to retry, same as in `methods` in [Spring Cloud Gateway](https://cloud.spring.io/spring-cloud-gateway/reference/html/#the-retry-gatewayfilter-factory).

### Motivation

I want to have ability to enable retries for several methods, not for all, only for GET for example, but skip it for POST ones. 

Let's consider we have game application. It has in-play payment system. I don't want to retry any money charges(POST, PUT for example), but I still want to retry all non-modifying GET queries.

### More

I also added new kind of tests: TestRetryWithRequestBody()

### DIFF
--- a/pkg/config/dynamic/middlewares.go
+++ b/pkg/config/dynamic/middlewares.go
@@ -370,6 +370,7 @@ type ReplacePathRegex struct {
 // Retry holds the retry configuration.
 type Retry struct {
 	Attempts int `json:"attempts,omitempty" toml:"attempts,omitempty" yaml:"attempts,omitempty" export:"true"`
+	Methods []string `json:"methods,omitempty" toml:"methods,omitempty" yaml:"methods,omitempty" export:"true"`
 }
 
 // +k8s:deepcopy-gen=true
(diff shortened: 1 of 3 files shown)


notsureifkevin:
@nkonev I think this is a great idea. However, with only `method` implemented, there are several other factors that need consideration when retrying requests other than timeouts.

The spring cloud example you referenced includes configuration for a number of other parameters which would be required for this to be considered, including:

- `retries` (number of retries)
- `statuses`, `series` (ex: 503, 401 || 5XX)
- `backoff` (and subsequent settings to implement exponential backoff)

 Please consider opening an issue with your feature request.

nkonev:
@notsureifkevin sorry, but at this moment I would like to confine the scope only with `method` parameter.

I gonna open the issue, thank you!

mpl:
@nkonev this PR should not be needed, as the retry mechanism does _not_ apply as soon as traefik manages to connect to the backend. So even if only 1 byte of your POST request makes it to the backend, the request won't be retried.
So unless you've actually experienced problems on that matter, we think we can close this PR. Let us know.

nkonev:
@mpl Thank you. I will read code, which calls retry middleware, scrupulously.

PR #6951, opened 2020-06-23
feat: update more than one LoadBalancer ip

### LINKED ISSUE #6950: Traefik only copies a single IP from a Kubernetes Service to the Ingress

Bug

### What did you do?

I'm running Traefik on a k3s cluster using the v2 Helm Chart. I have this specified, so as to populate all of the Ingress statuses from the IPs in the LoadBalancer service:

```
--providers.kubernetesIngress.ingressEndpoint.publishedService=default/base-ingress-traefik
```

The Service in question has three IPs:

```yaml
apiVersion: v1
kind: Service
metadata
  [ ... etc  ... ]
  name: base-ingress-traefik
  namespace: default
spec:
  [ ... etc .. ]
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - ip: 194.61.20.16
    - ip: 194.61.20.198
    - ip: 194.61.20.61
```

However, Traefik is only copying one of the three IPs from the LoadBalancer to the Ingress status:

```yaml
kind: Ingress
metadata:
  name: staging-livepeer-api
  namespace: default
spec: [redacted]
status:
  loadBalancer:
    ingress:
    - ip: 194.61.20.16
```

This isn't great for load balancing, I'd like all three to be there! Am I doing something wrong?

```
Version:      2.2.1
Codename:     chevrotin
Go version:   go1.14.2
Built:        2020-04-29T18:02:09Z
OS/Arch:      linux/amd64
```

Here's the Helm values:

```yaml
  deployment:
    replicas: 3
  service:
    type: LoadBalancer
  affinity: {}
  additionalArguments:
    - "--accesslog=true"
    - "--entrypoints.websecure.http.tls=true"
    - "--providers.kubernetesIngress.ingressEndpoint.publishedService=default/base-ingress-traefik"
  rbac:
    enabled: true
  service:
    annotations:
      # livepeer.live/external-dns: "true"
      external-dns.alpha.kubernetes.io/ttl: "120"
```

iameli:
### What does this PR do?

Traefik now updates Kubernetes Ingresses with all of the load balancer IPs from its `providers.kubernetesIngress.ingressEndpoint.publishedService`

Fixes #6950

I haven't yet updated any tests, but it appears to work properly.

### Motivation

We're running `external-dns` on our ingresses to keep DNS records synced up nicely. Before this change, we'd only get one of our IP records into DNS. Bad for load balancing.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/ingress/client_mock_test.go
+++ b/pkg/provider/kubernetes/ingress/client_mock_test.go
@@ -125,6 +125,6 @@ func (c clientMock) WatchAll(namespaces []string, stopCh <-chan struct{}) (<-cha
 	return c.watchChan, nil
 }
 
-func (c clientMock) UpdateIngressStatus(_ *networkingv1beta1.Ingress, _, _ string) error {
+func (c clientMock) UpdateIngressStatus(_ *networkingv1beta1.Ingress, _ []corev1.LoadBalancerIngress) error {
 	return c.apiIngressStatusError
 }
--- a/pkg/provider/kubernetes/ingress/client.go
+++ b/pkg/provider/kubernetes/ingress/client.go
@@ -54,7 +54,7 @@ type Client interface {
 	GetService(namespace, name string) (*corev1.Service, bool, error)
 	GetSecret(namespace, name string) (*corev1.Secret, bool, error)
 	GetEndpoints(namespace, name string) (*corev1.Endpoints, bool, error)
-	UpdateIngressStatus(ing *networkingv1beta1.Ingress, ip, hostname string) error
+	UpdateIngressStatus(ing *networkingv1beta1.Ingress, ingresses []corev1.LoadBalancerIngress) error
 	GetServerVersion() (*version.Version, error)
 }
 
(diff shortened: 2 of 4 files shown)


iameli:
Remaining work on this:

* Fix tests
* I suppose if the ingress status is set correctly but the IPs are in a different order, we shouldn't update. I don't know if that's deterministic or not... probably not. Should do a set comparison rather than a list comparison.

ldez:
hello, could you take a look at https://github.com/livepeer/traefik/pull/1, thank you.

iameli:
Rebased and resolved conflicts, thanks @kevinpollet!

pkg/provider/kubernetes/ingress/client_mock_test.go
```diff
@@ -125,6 +125,6 @@ func (c clientMock) WatchAll(namespaces []string, stopCh <-chan struct{}) (<-cha
 	return c.watchChan, nil
 }
 
-func (c clientMock) UpdateIngressStatus(_ *networkingv1beta1.Ingress, _, _ string) error {
+func (c clientMock) UpdateIngressStatus(_ *v1beta1.Ingress, _ []corev1.LoadBalancerIngress) error {
```
ldez, line 128:
```suggestion
func (c clientMock) UpdateIngressStatus(_ *networkingv1beta1.Ingress, _ []corev1.LoadBalancerIngress) error {
```

ldez (changes requested):

ldez (approved):
Thanks :+1:

rtribotte (approved):
LGTM 👍

jbdoumenjou (approved):
LGTM

Issue #5565, opened 2019-10-02
Custom TLS Settings silently ignored when cipher suite not available


sebthom:
### Do you want to request a *feature* or report a *bug*?

Bug

### Did you try using a 1.7.x configuration for the version 2.0?

- [ ] Yes
- [X] No

### What did you do?

I created a TLS option having the restricted cipher suites configured that are listed [here](https://square.github.io/okhttp/tls_configuration_history/#restricted_tls-cipher-suites).

```yaml
tls:
  options:
    my_tls_settings:
      minVersion: VersionTLS12
      clientAuth:
        clientAuthType: RequireAnyClientCert
      sniStrict: true
      cipherSuites:
        - TLS_AES_128_GCM_SHA256
        - TLS_AES_256_GCM_SHA384
        - TLS_CHACHA20_POLY1305_SHA256
        - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
        - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
        - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
        - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
        - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256
        - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256

```

In my router I am referencing the TLS option like so:
`traefik.http.routers.myservice.tls.options=my_tls_settings@file`

### What did you expect to see?

I expected being prompted with a client certificate request and being required to use TLS 1.2 or higher and one of the specified cipher suites.

### What did you see instead?

The TLS settings are ignored and default options are applied instead. Checking the settings using https://www.ssllabs.com/ssltest/analyze.html confirm that neither TLS version, nor cipher suites are restricted as specified.

When enabling DEBUG level in traefik, I see this log statement:
``` 
time="2019-10-02T11:02:47+02:00" level=debug msg="invalid CipherSuite: TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256" entryPointName=https routerName=myservice@docker
```

This leads me to the conclusion that if one of the listed cipher suites is not available, Traefik silently ignores the complete TLS option instead of just ignoring the unavailable cipher suite or raising an error.

Especially the behaviour of falling back to default TLS settings feels like a security flaw to me.

### Output of `traefik version`: (_What version of Traefik are you using?_)

[...]

PR #6970, opened 2020-06-27
Security: Don't silently ignore bad TLS config

Jukie:
### What does this PR do?

Returns an error if errors are found in TLS configuration

### Motivation
#5565
Currently if errors are encountered in the TLS config they are ignored and the default is used instead which poses a security threat. The only way to know it failed would be to enable debug logs or to experience the issue on a live deployment.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/tls/tlsmanager.go
+++ b/pkg/tls/tlsmanager.go
@@ -91,7 +91,7 @@ func (m *Manager) Get(storeName string, configName string) (*tls.Config, error) 
 	config, ok := m.configs[configName]
 	if !ok {
 		err = fmt.Errorf("unknown TLS options: %s", configName)
-		tlsConfig = &tls.Config{}
+		return nil, err
 	}
 
 	store := m.getStore(storeName)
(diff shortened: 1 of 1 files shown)


mpl:
This change is too naive, as it will break the whole configuration reload in case of a bad TLS config update.

SuperSandro2000:
So what else do you suggest to fix this?

Jukie:
@mpl isn't that the entire point? Currently the code silently continues on (unless debug is enabled) but users should be notified if the tls config is failing.

SantoDE:
Maybe, we could open up an issue to discuss things first. Then we can give some more ideas about how a solution could be functional. Then, do the implementation.

Jukie:
Please take a look at #5565 then, I'm happy to discuss. This issue was opened last October but hadn't received any other feedback or ideas.
I'll admit I didn't expect this to be immediately merged without comment but the issue wasn't getting any traction and this seemed like a viable resolution.

Issue #6859, opened 2020-05-28
Make timeout for UDP connections configurable


tazle:
We are planning to push DTLS-protected UDP traffic through Traefik on its way into our system. When trying this, we noticed that Traefik forgets the UDP connections after 3 seconds (https://github.com/containous/traefik/blob/master/pkg/udp/conn.go#L17). The reporting interval for our data sources is at least 10 seconds, often 60 seconds, and sometimes longer, so they will form a new Traefik "connection" for each report.

The problem with forming a new "connection" for reach report is that DTLS does not include any kind of session id in the DTLS record, and instead relies on external identifiers (in this case UDP 4-tuple) to tie incoming packets to the correct session. With a new "connection" for each report, the source port changes for each report, which first causes a timeout and then causes a new DTLS session to be established. We would like to avoid this timeout and the cost of establishing a new DTLS session.

To reach this goal, we need to make the timeout for UDP connections configurable.

Lindenk:
+1

Honestly a timeout of 3 seconds is a pretty useless default in it's own right. Most routers keep UDP connections alive for at least 30s by default. Even if the timeout is hardcoded in traefik, 3s is unreasonably short

kevinpollet:
Closed by #6982

PR #6982, opened 2020-06-30
Feature: add udp timeout configuration

Lindenk:
### What does this PR do?

Fixes issue #6859 

### Motivation

The UDP endpoint feature is rather useless if traefik closes the session after 3 seconds. My particular use case is to route mumble voip packets through traefik and having the server break after 3 seconds of inactivity is not useful.

### More

- [x] Added/updated tests

### Additional Notes

If there is a recommended alternative for creating a dummy ticker, that might be a better solution for the implementation of timeout values of 0.


juliens (approved):
LGTM

ldez (approved):
LGTM

jspdown (approved):
LGTM :+1:

PR #7047, opened 2020-07-14
Get Entrypoints Port Address without protocol for redirect

SantoDE:
### What does this PR do?

Takes the entrypoints address without the protocol for the entrypoint based redirection.

### Motivation

Fix https://github.com/containous/traefik-helm-chart/issues/208

### More

- [X] Added/updated tests
- ~~[ ] Added/updated documentation~~

### DIFF
--- a/pkg/provider/traefik/internal_test.go
+++ b/pkg/provider/traefik/internal_test.go
@@ -205,7 +205,7 @@ func Test_createConfiguration(t *testing.T) {
 						},
 					},
 					"websecure": {
-						Address: ":443/tcp",
+						Address: ":443",
 					},
 				},
 			},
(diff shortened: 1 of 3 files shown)


pkg/provider/traefik/internal_test.go
```diff
@@ -205,7 +205,7 @@ func Test_createConfiguration(t *testing.T) {
 						},
 					},
 					"websecure": {
-						Address: ":443",
```
rtribotte, line 208: @SantoDE Adding a specific test case for the trailing protocol in the address could be a more explicit. WDYT ?

SantoDE: What ever you prefer to be honest. I can live with both.

rtribotte: Ok 👍i'll do it

PR #7094, opened 2020-07-23
Support configuring a HTTP client timeout in the Docker provider

### LINKED ISSUE #7093: Support configuring a HTTP client timeout in the Docker provider

Feature

I'm debugging an issue where I've configured the Docker provider to connect to a remote Docker engine through a TLS bridging load balancer. I'm not sure why this makes a difference, but it appears that most connections opened are not closed, and Traefik's Docker client keeps sending ACKs long after the response has been received. However, the connections do not appear to be reused.

It would be great to be able to configure the HTTP client timeout, which is exposed through the `client.WithTimeout()` function on the Docker client.

Note that this is specifically for the _provider's_ client accessing the Docker API, not for the data plane clients of Traefik and how they access Docker discovered services.

sirlatrom:
### What does this PR do?

This PR adds a new configuration option on the Docker provider setting the HTTP client timeout when a HTTP endpoint is used. It also sets a default of 1 hour, which should probably be debated as there is currently _no_ timeout by default.

### Motivation

Closes #7093.

### More

- [x] Added/updated tests

### Additional Notes

Let's debate whether it's a good or a bad idea to set a default timeout of e.g. 1 hour.

### DIFF
--- a/pkg/provider/docker/docker.go
+++ b/pkg/provider/docker/docker.go
@@ -57,6 +57,7 @@ type Provider struct {
 	SwarmMode               bool             `description:"Use Docker on Swarm Mode." json:"swarmMode,omitempty" toml:"swarmMode,omitempty" yaml:"swarmMode,omitempty" export:"true"`
 	Network                 string           `description:"Default Docker network used." json:"network,omitempty" toml:"network,omitempty" yaml:"network,omitempty" export:"true"`
 	SwarmModeRefreshSeconds ptypes.Duration  `description:"Polling interval for swarm mode." json:"swarmModeRefreshSeconds,omitempty" toml:"swarmModeRefreshSeconds,omitempty" yaml:"swarmModeRefreshSeconds,omitempty" export:"true"`
+	HTTPClientTimeout       ptypes.Duration  `description:"Client timeout for HTTP connections." json:"httpClientTimeout,omitempty" toml:"httpClientTimeout,omitempty" yaml:"httpClientTimeout,omitempty" export:"true"`
 	defaultRuleTpl          *template.Template
 }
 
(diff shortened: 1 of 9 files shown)


sirlatrom:
I don't understand what to do about the test failures of `Test_decodeFileToNode_Yaml` and `Test_decodeFileToNode_Toml` as I've placed the new option in the same order in all sample files.

sirlatrom:
I appreciate the fixes added to the PR. What makes 32 a good default?

ldez:
It's the same value a the dialer https://github.com/moby/moby/blob/516d9719d554ab983d4b53d19fa88ff21a98e091/vendor/github.com/docker/go-connections/sockets/sockets.go#L12

---

Please don't update your branch by merging master, we have a bot that manage automatically the rebase of the PR.

sirlatrom:
> Please don't update your branch manually (unless there are conflicts to be managed by hand), we have a bot that manage automatically the rebase of the PR.

Point taken, GitHub simply offered it with a big inviting button.

pkg/provider/docker/docker.go
```diff
@@ -176,7 +180,7 @@ func (p *Provider) getClientOpts() ([]client.Opt, error) {
 			return nil, err
 		}
 
-		opts = append(opts, client.WithHTTPClient(&http.Client{Transport: tr}))
+		opts = append(opts, client.WithHTTPClient(&http.Client{Transport: tr, Timeout: time.Duration(p.HTTPClientTimeout)}))
```
rtribotte, line 183: It seems that the timeout is not needed here, since it will be set to the client anyway because of L159 ?

```suggestion
		opts = append(opts, client.WithHTTPClient(&http.Client{Transport: tr}))
```

ldez: the timeout must be defined here because of the order in which the options are applied.

rtribotte: Ok i agree, i made a wrong assumption

SantoDE (approved):
LGTM 👍

SantoDE (changes requested):
needs rebase

SantoDE (approved):
Now it's good :)

sirlatrom:
Any chance of this making it into the next 2.2 release? It was already on the "next" milestone before the last 2.2 release.

ldez:
the current next is v2.4

PR #7109, opened 2020-07-27
fix: clean X-Forwarded-Prefix header for the dashboard.

ldez:
### What does this PR do?

Clean X-Forwarded-Prefix header for the dashboard.

### Motivation

Be safe.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/api/dashboard_test.go
+++ b/pkg/api/dashboard_test.go
@@ -22,7 +22,7 @@ func Test_safePrefix(t *testing.T) {
 		{
 			desc:     "host with path",
 			value:    "https://example.com/foo/bar?test",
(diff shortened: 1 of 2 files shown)


pkg/api/dashboard_test.go
```diff
@@ -0,0 +1,49 @@
...
+			value:    "https://example.com",
+			expected: "",
+		},
+		{
+			desc:     "path",
+			value:    "/foo/bar",
+			expected: "/foo/bar",
+		},
+		{
+			desc:     "path without trailing slash",
```
dtomcej, line 28:
```suggestion
			desc:     "path without leading slash",
```

PR #7170, opened 2020-08-13
Include enabling traefik in example

JanDevDE:
You need to enable traefik on the default traefik service for the router rules to apply. (I'm too lazy to add this to every example, feel free to do that yourself.)

### What does this PR do?

Adds a label to the example which is required to make the example work in the first place.

### Motivation

Two hours of struggle.

### More

- [ ] Added/updated tests

### Additional Notes

I only updated one example.

### DIFF
--- a/docs/content/operations/include-dashboard-examples.md
+++ b/docs/content/operations/include-dashboard-examples.md
@@ -1,6 +1,7 @@
 ```yaml tab="Docker"
 # Dynamic Configuration
 labels:
+  - "traefik.enable=true"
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

I fact you don't need to use `traefik.enable=true`, you need to use this label only if override the default behavior by defining `--providers.docker.exposedbydefault=false`

https://docs.traefik.io/v2.2/providers/docker/#exposedbydefault

So we will decline this PR, thanks anyway.

JanDevDE:
@ldez Ah, okay. Yes, I was following https://docs.traefik.io/v2.2/user-guides/docker-compose/acme-tls/ when installing traefik. So I wasn't aware of the fact that I set that.

PR #7171, opened 2020-08-14
Fix invalid string(int) instead of fmt.Sprint(int)

sylr:
### What does this PR do?

Fix a go vet reported issue.

### More

- [ ] Added/updated tests


ldez:
Already fix in #7169

Issue #6608, opened 2020-03-31
Traefik tries to use HTTP/2 with NTLM


tfenster:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I have a service behind Traefik that uses Windows Auth. As stated here https://docs.microsoft.com/en-us/iis/get-started/whats-new-in-iis-10/http2-on-iis#when-is-http2-not-supported, Windows Auth is not supported with HTTP/2. Unfortunately it seems to me like Traefik always uses HTTP/2 when the service scheme is https. This leads to the following error in the Traefik debug log:

```
traefik    | time="2020-03-31T16:31:51Z" level=debug msg="'500 Internal Server Error' caused by: stream error: stream ID 3; HTTP_1_1_REQUIRED"
```

If I run the backend service with http instead of https, then Windows Auth succeeds. 

### What did you expect to see?
Windows Authentication goes through, even when the backend service is running https. It would be good to have an option to e.g. force http/1.1 between Traefik and backend when needed

### What did you see instead?
Windows Authentication fails

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      2.2.0
Codename:     chevrotin
Go version:   go1.14.1
Built:        2020-03-25T17:17:27Z
OS/Arch:      windows/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

My docker-compose.yml looks like this when it works. Note that I disable SSL for the backend service and then use scheme http and port 80 on the service loadBalancer
```yml
  bc:
    image: mcr.microsoft.com/businesscentral/onprem:1810-ltsc2019
    container_name: bc
    hostname: traeftest
    environment:
      - accept_eula=y
      - clickonce=y
      - webserverinstance=bc
      - publicdnsname=traeftest.westeurope.cloudapp.azure.com
      - user=VM-Administrator
      - password=...
      - auth=Windows
#      - usessl=y
    labels:
      - traefik.enable=true
      - traefik.http.routers.bc.rule=Host(`traeftest.westeurope.cloudapp.azure.com`) && PathPrefix(`/bc`)
      - traefik.http.routers.bc.entrypoints=websecure
      - traefik.http.routers.bc.tls.certresolver=myresolver
      - traefik.http.routers.bc.service=bc@docker
      - traefik.http.services.bc.loadBalancer.server.scheme=http
      - traefik.http.services.bc.loadBalancer.server.port=80
```

[...]

PR #7179, opened 2020-08-14
introduce an option to disable http/2

tfenster:
### What does this PR do?

This PR adds a config option to disable http/2. This is needed in cases like described in #6608 where the backend doesn't work with http/2 but the request isn't automatically switched to http/1.1

### Motivation

I wanted a fix for #6608

### More

- [ ] Added/updated tests --> I didn't find the right place where I could put this. If you could point me in the right direction, I'd be happy to try and add a test as well

### Additional Notes

This is my first traefik PR and actually my first PR in Go, so sorry in advance for any stupid mistakes and thanks for feedback where I went wrong

### DIFF
--- a/docs/content/providers/docker.md
+++ b/docs/content/providers/docker.md
@@ -199,7 +199,7 @@ Therefore you **must** specify the port to use for communication by using the la
 Docker Swarm Mode follows the same rules as Docker [API Access](#docker-api-access).
 
 As the Swarm API is only exposed on the [manager nodes](https://docs.docker.com/engine/swarm/how-swarm-mode-works/nodes/#manager-nodes), you should schedule Traefik on the Swarm manager nodes by default,
-by deploying Traefik with a [constraint](https://success.docker.com/article/using-contraints-and-labels-to-control-the-placement-of-containers) on the node's "role":
+by deploying Traefik with a [constraint](https://docs.docker.com/engine/reference/commandline/service_create/#specify-service-constraints---constraint) on the node's "role":
 
 ```shell tab="With Docker CLI"
 docker service create \
--- a/docs/content/routing/overview.md
+++ b/docs/content/routing/overview.md
@@ -400,3 +400,26 @@ serversTransport:
 ## Static configuration
 --serversTransport.forwardingTimeouts.idleConnTimeout=1s
 ```
+
+
+### `disableHTTP2`
+
+_Optional, Default=false_
+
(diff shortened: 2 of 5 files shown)


rtribotte:
Hello @tfenster,

Thanks for your time and your interest in Traefik !

The issue that you want to solve seems to come from the backend side.
At least, with IIS, the negotiation goes wrong and the fallback to HTTP1.1 is not working, which is not the case for all http servers.

As HTTP2 has become a standard, we are not willing to allow it deactivation as a global behavior.
We believe that it would lead to unwanted side effects.

Thus, an upcoming feature will be to set up the ServersTransport options by service thanks to #7203 
We may reconsider your proposal, if you wait for it to be merged and take advantage of this new feature, to limit the scope of the deactivation of HTTP2 for only one service, and not as a global configuration.

WDYT ?

tfenster:
@rtribotte That 100% makes sense for me. I was actually looking for something more limited and only did it globally because I couldn't find a good place for it. But doing it by service of course is a lot better. (And I agree that the actual problem probably is the backend but as the chances are between 0 and non-existent that I can get a fix in IIS, I would love a workaround in Traefik...)

Would you have a rough idea when #7203 will be merged? I understand that you can't make promises, just a noncommittal best guess

rtribotte:
Hello @tfenster,

Sorry for the delay, even a noncommittal best guess was something i couldn’t give you at the time of your answer.
But just to let you know, the #7203 PR has finally been merged !
Do you think you can still make the changes ?

Thanks again for your time and interest in Traefik !

tfenster:
Hello @rtribotte,

No problem, thanks for letting me know. I'll try to make my changes again and get back

jcuzzi:
@tfenster I'm looking forward to this change.. It will help me out a lot.

tfenster:
@jcuzzi same here ;) I was stuck with other stuff, hope to get it done this or next weekend

kevinpollet:
Hello @tfenster,

Maybe you have already seen that a [pull request](https://github.com/traefik/traefik/pull/7645) has been opened to disable `HTTP/2` by using the `ServersTransports` feature introduced in #7203.

Will you have the time to update your PR or we can close it in favor of #7203?

tfenster:
I was hoping to get this done over the holidays, but if the other PR covers the issue, then I am happy to close this one and find something else to do :)

kevinpollet:
@tfenster thanks again for your time and your contribution!

Closed in favor of #7645

PR #7192, opened 2020-08-18
added mention of target address for prometheus

smeierzubiesen:
there's no trace of a target address for the prometheus observable

not sure of that is by design by here's my proposed change of the documentation

### What does this PR do?

This PR fixes a "missing-documentation" occurence under the Prometheus Observability Tool
There is no mention of the target prometheus server (or port)

### Motivation

Missing/incorrect documentation should always be adressed

### More

### Additional Notes

If there's something I'm missing in my PR I'm more than happy to fix it

### DIFF
--- a/docs/content/observability/metrics/prometheus.md
+++ b/docs/content/observability/metrics/prometheus.md
@@ -16,6 +16,28 @@ metrics:
 --metrics.prometheus=true
 ```
 
+#### `address`
+
+_Required, Default="localhost:9090"_
+
+Address instructs exporter to send metrics to Prometheus at this address.
+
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

Prometheus uses a pull system, so there no `address` field.

The exposed part of the Prometheus implementation in Traefik is managed by `EntryPoint` or `manualRouting`

- https://docs.traefik.io/v2.2/observability/metrics/prometheus/#entrypoint
- https://docs.traefik.io/v2.2/observability/metrics/prometheus/#manualrouting

So we have to decline your PR, thanks anyway.

PR #7237, opened 2020-09-02
doc: specify HostSNI rule removal only for HTTP routers

### LINKED ISSUE #7229: Clarification on HostSNI removal

Bug

In the migration docs [here](https://docs.traefik.io/v2.2/migration/v2/#v21-to-v22) :

```
HostSNI rule matcher removal¶
In v2.2.2 we introduced a new rule matcher (HostSNI) which was allowing to match the Server Name Indication at the router level. Since v2.2.5 this rule has been removed, and you should not use it anymore.
```

In creating a TCP router, using `Host('*')` results in:

```
time="2020-08-30T16:31:34Z" level=error msg="unknown rule Host(`pihole.cluster.local`)" entryPointName=dns routerName=default-pihole-dns-a2be5f
1a3e4fc38e3759@kubernetescrd
```

The error disappears if i use `HostSNI` but the docs say that it's removed.

`HostSNI` is also referenced in the v2.3 docs [here](https://docs.traefik.io/v2.3/routing/providers/kubernetes-crd/#kind-ingressroutetcp)

An error using `HostSNI` and no error using `Host`

### What did you see instead?

An error using `Host` and no error using `HostSNI`

± kubectl exec -it traefik-85d7c864fb-hw64j traefik version
Version:      2.3.0-rc4
Codename:     picodon
Go version:   go1.15
Built:        2020-08-19T16:02:03Z
OS/Arch:      linux/arm64

no relevant

rtribotte:
### What does this PR do?

This PR fix the documentation to avoid misleading assumptions when reading the HostSNI rule removal section in the migration guide, by specifying the scope of the removal of the HostSNI rule to HTTP routers only.

### Motivation

closes #7229 

### More

- [ ] ~Added/updated tests~

### DIFF
--- a/docs/content/migration/v2.md
+++ b/docs/content/migration/v2.md
@@ -312,5 +312,5 @@ Since `v2.2.5` this global option has been removed, and you should not use it an
 
 ### HostSNI rule matcher removal
 
-In `v2.2.2` we introduced a new rule matcher (`HostSNI`) which was allowing to match the Server Name Indication at the router level.
-Since `v2.2.5` this rule has been removed, and you should not use it anymore.
+In `v2.2.2` we introduced a new rule matcher (`HostSNI`) for HTTP routers which was allowing to match the Server Name Indication at the router level.
+Since `v2.2.5` this rule has been removed for HTTP routers, and you should not use it anymore.


ldez (approved):
LGTM

PR #7269, opened 2020-09-09
Improve documentation for usage of Kubernetes Ingress

grieshaber:
### What does this PR do?

Provide an example for the usage of a Middleware with Kubernetes Ingress, include the requirements for the annotations in these scenarios, and improve the documentation for the setup of default TLSOptions in k8s.

### Motivation

We had a hard time figuring out, how to setup Middlewares / TLSOptions with the "old" style Kubernetes Ingress after upgrading to v2.x.
As we do have a lot of projects in our clusters, we didn't want to force an update to CRD to function correctly. Therefore we needed the correct annotation-format and a default TLSOption working for all ingresses.  

### More

### DIFF
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -138,8 +138,9 @@ The TLS options allow one to configure some parameters of the TLS connection.
 
     When using the TLSOptions-CRD in Kubernetes, one might setup a default set of options that,
     if not explicitly overwritten, should apply to all ingresses. To achieve that, you'll have to
-    create a TLSOptions CR with the name `default` in the namespace `default`.
-    To explicitly use a different TLSOptions CR (and using the Kubernetes Ingress resources) you'll 
+    create a TLSOptions CR with the name `default`. There may exist only one TLSOption with the 
+    name `default` (across all namespaces) - otherwise they will be dropped.  
+    To explicitly use a different TLSOption (and using the Kubernetes Ingress resources) you'll 
     have to add an annotation to the Ingress in the following form:
     `traefik.ingress.kubernetes.io/router.tls.options: <resource-namespace>-<resource-name>@kubernetescrd`
 
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -134,6 +134,15 @@ If no default certificate is provided, Traefik generates and uses a self-signed 
 
 The TLS options allow one to configure some parameters of the TLS connection.
 
+!!! important "TLSOptions in Kubernetes"
+
+    When using the TLSOptions-CRD in Kubernetes, one might setup a default set of options that,
+    if not explicitly overwritten, should apply to all ingresses. To achieve that, you'll have to
+    create a TLSOptions CR with the name `default` in the namespace `default`.
+    To explicitly use a different TLSOptions CR (and using the Kubernetes Ingress resources) you'll 
+    have to add an annotation to the Ingress in the following form:
+    "traefik.ingress.kubernetes.io/router.tls.options: <namespace-of-CR>-<name-of-CR>@kubernetescrd"
+
 ### Minimum TLS Version
 
 ```toml tab="File (TOML)"
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -141,7 +141,7 @@ The TLS options allow one to configure some parameters of the TLS connection.
     create a TLSOptions CR with the name `default` in the namespace `default`.
     To explicitly use a different TLSOptions CR (and using the Kubernetes Ingress resources) you'll 
     have to add an annotation to the Ingress in the following form:
-    "traefik.ingress.kubernetes.io/router.tls.options: <namespace-of-CR>-<name-of-CR>@kubernetescrd"
+    `traefik.ingress.kubernetes.io/router.tls.options: <resource-namespace>-<resource-name>@kubernetescrd`
 
 ### Minimum TLS Version
 
--- a/docs/content/middlewares/overview.md
+++ b/docs/content/middlewares/overview.md
@@ -25,7 +25,7 @@ whoami:
     - "traefik.http.routers.router1.middlewares=foo-add-prefix@docker"
 ```
 
-```yaml tab="Kubernetes"
+```yaml tab="Kubernetes IngressRoute"
 # As a Kubernetes Traefik IngressRoute
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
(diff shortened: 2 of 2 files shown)


docs/content/https/tls.md
```diff
@@ -134,6 +134,15 @@ If no default certificate is provided, Traefik generates and uses a self-signed
 
 The TLS options allow one to configure some parameters of the TLS connection.
 
+!!! important "TLSOptions in Kubernetes"
+
+    When using the TLSOptions-CRD in Kubernetes, one might setup a default set of options that,
+    if not explicitly overwritten, should apply to all ingresses. To achieve that, you'll have to
+    create a TLSOptions CR with the name `default` in the namespace `default`.
```
rtribotte, line 141:
```suggestion
    create a TLSOptions CR with the name `default`.
```

rtribotte:
Hello @grieshaber,

I made several suggestions that fix the syntax and also removing that the namespace `default` is needed, because it is not to be recognized as the "default" option.
I'll need them to be applied to approve this PR, do you think that you can apply them ?

Thanks for your time and your interest in Traefik !

grieshaber:
Hi @rtribotte,

thanks for your review.
Of course, i'll apply the changes shortly. Sorry for the syntax-stuff 🙄 

Regarding the default TLSOptions: i'll give it a quick try in one of our clusters. Doesn't the namespace care at all? I thought, i'd tried other settings without success.

docs/content/middlewares/overview.md
```diff
@@ -63,6 +63,45 @@ spec:
       - name: stripprefix
 ```
 
+```yaml tab="Kubernetes Ingress
```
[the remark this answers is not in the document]
grieshaber: Fixed with commit https://github.com/containous/traefik/pull/7269/commits/6d13a6067f738608aac0daa925d0e7f40b9db8e2.

docs/content/middlewares/overview.md
```diff
@@ -148,6 +187,9 @@ then you'll have to append to the middleware name, the `@` separator, followed b
     In this case, since the definition of the middleware is not in kubernetes,
     specifying a "kubernetes namespace" when referring to the resource does not make any sense,
     and therefore this specification would be ignored even if present.
+    On the other hand, if you declare the middleware as a CRD in Kubernetes and use the non-crd Ingress 
+    objects, you'll have to add the kubernetes namespace, the middleware lives in, to the annotation like
+    this "<middleware-namespace>-<middleware-name>@kubernetescrd".
```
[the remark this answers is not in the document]
grieshaber: Fixed with commit https://github.com/containous/traefik/pull/7269/commits/6d13a6067f738608aac0daa925d0e7f40b9db8e2.

docs/content/https/tls.md
```diff
@@ -134,6 +134,15 @@ If no default certificate is provided, Traefik generates and uses a self-signed
...
+    To explicitly use a different TLSOptions CR (and using the Kubernetes Ingress resources) you'll 
+    have to add an annotation to the Ingress in the following form:
+    "traefik.ingress.kubernetes.io/router.tls.options: <namespace-of-CR>-<name-of-CR>@kubernetescrd"
```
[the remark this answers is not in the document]
grieshaber: Fixed with commit https://github.com/containous/traefik/pull/7269/commits/6d13a6067f738608aac0daa925d0e7f40b9db8e2.

rtribotte:
Hi @grieshaber,

For the default tlsOption the only constraint is to declare only one resource with the name `default`, otherwise all tlsOptions with the name `default` will be dropped, and then the fallback is done on the "internal default tls option".

docs/content/https/tls.md
grieshaber: Clarified with commit https://github.com/containous/traefik/pull/7269/commits/7b31d6c6cfa5af39e210d545faf02a94cfa8218e.

grieshaber:
Thanks for clarification. I tested the described behaviour successfully. Therefore i adapted the description to include the speciality you mentioned.

docs/content/middlewares/overview.md
```diff
@@ -63,6 +63,45 @@ spec:
...
+      - /stripit
+
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: ingress
+  namespace: appspace
+  annotations:
+    # annotation the form <middleware-k8s-namespace>-<middleware-name>@kubernetescrd
```
rtribotte, line 99: No worries, thanks for the changes :)

After a second thought, i think that the cross-provider example you added should be placed in the next examples section entitled "Referencing a Middleware from Another Provider".
It's more appropriate since the examples section in which you added it, is only showing how to declare and reference a middleware from one provider perspective.

I would also change the comment for the annotation to be more specific:
```suggestion
    # referencing a middleware from Kubernetes CRD provider: <middleware-namespace>-<middleware-name>@kubernetescrd
```

Do you think you can apply those changes too ?

docs/content/middlewares/overview.md
```diff
@@ -148,6 +187,9 @@ then you'll have to append to the middleware name, the `@` separator, followed b
     In this case, since the definition of the middleware is not in kubernetes,
     specifying a "kubernetes namespace" when referring to the resource does not make any sense,
     and therefore this specification would be ignored even if present.
+    On the other hand, if you declare the middleware as a CRD in Kubernetes and use the non-crd Ingress
```
SantoDE, line 190: Can we rephrase to something like

`On the other hand, if you declare the middleware as a Custom Ressource in Kubernetes...`? Because the CRD is the definition of the resource, not the ressource itself

docs/content/middlewares/overview.md
```diff
@@ -148,6 +187,9 @@ then you'll have to append to the middleware name, the `@` separator, followed b
     In this case, since the definition of the middleware is not in kubernetes,
     specifying a "kubernetes namespace" when referring to the resource does not make any sense,
     and therefore this specification would be ignored even if present.
+    On the other hand, if you declare the middleware as a CRD in Kubernetes and use the non-crd Ingress 
+    objects, you'll have to add the kubernetes namespace, the middleware lives in, to the annotation like
```
SantoDE, line 191: you`ll have to add the kubernetes namespace of the middleware to the annotation... ?

SantoDE (changes requested):
👋  Thanks for your work.

Too minor suggestions :)

docs/content/middlewares/overview.md
grieshaber: I fully agree that it would fit best in the section you mentioned, although i liked it in the other section, as it would provide a more complete overview - but, nonetheless, i moved the example and improved the comment with https://github.com/containous/traefik/pull/7269/commits/4143d12d3479b1f7d83b2834fe2234958f994b19.

docs/content/middlewares/overview.md
grieshaber: Yes - you're right. CR is the correct term here.. fixed with https://github.com/containous/traefik/pull/7269/commits/4143d12d3479b1f7d83b2834fe2234958f994b19

docs/content/middlewares/overview.md
grieshaber: Changed with https://github.com/containous/traefik/pull/7269/commits/4143d12d3479b1f7d83b2834fe2234958f994b19

rtribotte (approved):
Thanks 👍

SantoDE (approved):
Lgtm - thanks :)

jbdoumenjou (approved):
LGTM

PR #7271, opened 2020-09-10
Fix typo when comparing exported data

RealOrangeOne:
### What does this PR do?

Fix a typo in the documentation for what data is requested.

The diff for this PR isn't the best, but the default screen for this page (https://docs.traefik.io/contributing/data-collection/#example-of-collected-data) shows `insecureSkipVerify = true` in the original code, but `false` for the obfuscated. Looking at `anonymize.go`, I can't see anything which would result in a boolean flip, so I'm assuming this is a typo.

### Motivation

Typos = bad


ldez (approved):
Thanks :+1:

PR #7272, opened 2020-09-10
Avoid Traefik Pilot iframe code in Traefik webui regarding notifications

matthieuh:
### What does this PR do?

Avoid Traefik Pilot code in Traefik webui regarding notifications.

### Motivation

- For the notification displayed when current Traefik instance is not registered in Traefik Pilot, use the NavBar iframe in the header instead of having this logic and rendering done in the Traefik webui.
- Some changes have been applied to make it possible like adding the ability to change iframe url path from itself and avoiding to have the `/instances` path statically defined.
- Makes Pilot notification less prominent having it only in the header.

### More

- [ ] Added/updated tests

### Additional Notes

Since the iframe in the nav bar is a bit larger this PR also include a fix displaying only 7 first digits of the version when this one is the commit sha.

Screen Shot 2020-09-10 at 15 47 11


ldez (approved):
LGTM

jbdoumenjou (approved):
LGTM

rtribotte (approved):
LGTM 👍

Issue #6560, opened 2020-03-24
traefik v2.x prometheus metrics


DeoLeung:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

at the moment I can see `v2.x` expose following metrics to prometheus

* traefik_service_requests_total
* traefik_config_last_reload_failure
* traefik_entrypoint_request_duration_seconds_count
* traefik_service_request_duration_seconds_sum
* traefik_config_reloads_total
* traefik_entrypoint_request_duration_seconds_sum
* traefik_service_open_connections
* traefik_entrypoint_requests_total
* traefik_service_request_duration_seconds_count
* traefik_config_last_reload_success
* traefik_config_reloads_failure_total
* traefik_entrypoint_request_duration_seconds_bucket
* traefik_entrypoint_open_connections
* traefik_service_request_duration_seconds_bucket

I would like to request some features
 *  metrics like `server_up` in `v1.x` so I can count the active servers
 * flag to enable adding `url` label to `traefik_service_request_duration_seconds` so we can do a bit more detail observation on that
 * with `docker` provider, since traefik knows the address of the service servers, would be great if traefik support exposing it to `prometheus`

zolech:
+1 for `server_up` metric

I would like to see that as well. Right now I had to remove this from my dashboards after migrating from Marathon to k8s provider. I see this information in traefik dashboard but there is no such thing in metrics.

thewilli:
having the amount of days until a used tls certificate is about to expire would be useful as well

Kiina:
Adding to the tls certificate, how about seeing handshake informaton like cipherSuites and tls protocol versions for connections? It would be kinda nice to know how many connections aren't made due to the client only supporting low tls versions or weak ciphers.

rcomanne:
Yes, +1 for the `server_up` metrics, we use this for monitoring our services, losing this metric means waiting with upgrading.
Any idea if this will be added anytime soon?

mgerasimchuk:
Recently I moved from ingress-nginx controller to traefik, and only now I realised, that traefik unfortunately doesn't have any way to collect metrics by URL path. For example in ingress-nginx controller this behavior works by default: https://github.com/kubernetes/ingress-nginx/blob/96b6228a6b65a85e421b8a348a149e99181664d1/deploy/grafana/dashboards/request-handling-performance.json#L314

[...]

Issue #3196, opened 2018-04-17
Prometheus: More detailed metrics


danielh1989:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?
I like to have a more detailed metric with prometeus.
For example:

- Requests per Route
- Requests per Backend
- Average Time per Backend
- Average Time per Route

mmatur:
Hi @danielh1989,

Thanks for your interest in Træfik projet.

For information, the current available metrics are :

- config_reloads_total
- traefik_config_reloads_failure_total
- traefik_config_last_reload_success
- traefik_config_last_reload_failure
- traefik_entrypoint_requests_total
- traefik_entrypoint_request_duration_seconds
- traefik_entrypoint_open_connections
- traefik_backend_requests_total
- traefik_backend_request_duration_seconds
- traefik_backend_open_connections
- traefik_backend_retries_total
- traefik_backend_server_up

marco-jantke:
What do you mean exactly when you tell "per route"? Prometheus doesn't like "unbound" label values and so putting an arbitrary path into it is not an option if you don't want to kill your Prometheus server.

Place1:
as an example, i'd really like to know which API (based on a path) is slower than others when using the traefik_backend_requests_total metric. my services don't have an unbounded number of APIs so perhaps Traefik could only fill the label for backend requests that are not a 404 to avoid client spam.

mgerasimchuk:
Recently I moved from ingress-nginx controller to traefik, and only now I realised, that traefik unfortunately doesn't have any way to collect metrics by URL path. For example in ingress-nginx controller this behavior works by default: https://github.com/kubernetes/ingress-nginx/blob/96b6228a6b65a85e421b8a348a149e99181664d1/deploy/grafana/dashboards/request-handling-performance.json#L314

Traefik has a lot of pros like SSL certificates management and very comfortable routing, but observability of the system also is one of the most important thing, and now I can't understand which route of my API works slow....
Hope that the `path` label will be added in nearest future..

rayjanoka:
I found the metric labels were more usable if I used an intermediate `TraefikService` in the path to my actual k8s service.

hick:
[...]

PR #7278, opened 2020-09-11
Add URL path to metrics labels

frejonb:
### What does this PR do?

It adds the request URL path to the metrics middleware.

### Motivation

I'm using a single traefik service that exposes an API Gateway. With this configuration, I can't get metrics for each internal service (roughly a path corresponds to an internal service). This PR addresses partially https://github.com/containous/traefik/issues/3196 and https://github.com/containous/traefik/issues/6560. 

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/metrics/prometheus.go
+++ b/pkg/metrics/prometheus.go
@@ -137,17 +137,17 @@ func initStandardRegistry(config *types.Prometheus) Registry {
 	if config.AddEntryPointsLabels {
 		entryPointReqs := newCounterFrom(promState.collectors, stdprometheus.CounterOpts{
 			Name: entryPointReqsTotalName,
-			Help: "How many HTTP requests processed on an entrypoint, partitioned by status code, protocol, and method.",
-		}, []string{"code", "method", "protocol", "entrypoint"})
+			Help: "How many HTTP requests processed on an entrypoint, partitioned by status code, protocol, method, and path.",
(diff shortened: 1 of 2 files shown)


sylr:
This change could increase exponentially the cardinality of the metric which is not a good thing.

I understand the need but I think adding the path as label should be configurable/optional.

frejonb:
@sylr indeed I have noticed that. Internally, I have added some rules on the possible paths to reduce the cardinality. But I haven’t really thought of a good configuration interface for it. I’m open to ideas

rtribotte:
Hello @frejonb,

I also think that the cardinality increase for those four metrics is an issue.
Making it optional could be a solution, but metrics are not homogenous, and the option would impact only a subset of services and entrypoints metrics (essentially those four metrics you modified in your PR).
But what about the TLS requests counter ? I think that the path label is also interesting for this metric.
I tend to think that introducing a new set of metrics ("path" metrics) would be the best option.

Thus, having said that, after discussing it with the team, i can say that we believe this is not a common use case nor a common practice to aggregate metrics on a "path" label. One can easily imagine how risky it would be to create a new metric entry for each wrong path incoming request.

I will decline this PR accordingly, but i invite you to defend your proposal in a dedicated issue because we maybe miss something here. Also i don't think the issues #3196 and #6560 are related to the introduction of a path label.

PR #7281, opened 2020-09-13
Remove allocation with the use of the buffer pool

diegobernardes:
More info: https://staticcheck.io/docs/checks#SA6002

### What does this PR do?
The current buffer pool is doing an unnecessary allocation. More info here: https://staticcheck.io/docs/checks#SA6002

### Motivation
Free performance improvement.

### DIFF
--- a/pkg/server/service/bufferpool.go
+++ b/pkg/server/service/bufferpool.go
@@ -23,5 +23,5 @@ func (b *bufferPool) Get() []byte {
 }
 
 func (b *bufferPool) Put(bytes []byte) {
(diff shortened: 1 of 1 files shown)


ldez:
Hello, in fact, in this case, it's a false-positive report of staticcheck, because we have to use a copy.

FYI this change produces a panic:
```
panic: interface conversion: interface {} is *[]uint8, not []uint8 [recovered]
	panic: interface conversion: interface {} is *[]uint8, not []uint8
```

Thanks anyway.

Issue #4429, opened 2019-01-25
Add exceptions for entryPoints.http.authBasic


dkossako:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

It would be useful to add whitelisting for entryPoints.http.auth.basic. Like:
```
[entryPoints.http.auth.basic]
usersFile = "/etc/traefik/.htpasswd"
whiteList = [ "192.168.0.0/16" ]
```
Such feature would be used in projects to allow local network users view entrypoints without being authenticated.

wodzik:
+1

alesnav:
This would be a nice feature, but it is interesting also for `traefik.http.middlewares.xxx.forwardauth`, because of the same reason.

GildedHonour:
Is help needed with this?

maksadbek:
Should this feature get client's IP from `request.RemoteAddr` or from HTTP header value like `X-Real-IP` or `X-Forwarded-For` ?

dkossako:
imo remote addr, but maybe it would be fine to add some kind of boolean switch to change this behavior and allow to take from header?

maksadbek:
Okay, that sounds reasonable. In this case the config should look like this:
```
[entryPoints.http.auth.basic]
  usersFile = "/etc/traefik/.htpasswd"
  [entryPoints.http.auth.basic.ipAllowList]
     allowedIPs = [ "192.168.0.0/16" ]
     clientIPFromHeader = ["X-Real-IP", "X-Forwarded-For"]  # if any of them is valid and in the given subnet use them. Otherwise, fallback to `request.RemoteAddr`
```

If the `clientIPFromHeader` is empty, then fallback to `request.RemoteAddr`

dkossako:
Looks good to me

ldez:
For those who don't know: since one year, the v1 of Traefik is in maintenance mode, which means there are more new features added and only fixes can be added.

Otherwise, this issue can be linked to #6007

rtribotte:
Hello,

Based on the reading of the PR of @maksadbek, we think that a better approach would be to rely on the routing, in place of trying to "enhance" current middlewares with IP allow-listing.
The idea is to provide a specific routing based on IPs, that would allow to have routes with and without an auth middleware (for instance).

WDYT ?

maksadbek:
Hello @rtribotte 

[...]

Issue #6007, opened 2019-12-12
OR-Chained middlewares


kolaente:
### Do you want to request a *feature* or report a *bug*?

Feature

Feature

### What did you expect to see?

It would be cool if traefik could support chaining middlewares with or instead of executing them one after one. Currently, there is the [chain middleware](https://docs.traefik.io/middlewares/chain/), which executes middlewares one after the other. The thing I have in mind would be to execute the first one in the list, if that fails execute the second one etc. A problem here might be a middleware does not "fail" or "succeed", it either calls the next one or does not iirc.

A bit more context on the use case:
We have one site which we want to secure using ip whitelist and basic auth, either one should work. If a user accesses the site not on the ip whitelist, it should need to complete the basic auth.

I'm not sure if a general way of doing this is a good idea or if the use case is too specific and thus, a custom thing would be better suited.

I've discussed this a bit in https://github.com/justinas/alice/issues/50

alexandru-enciu:
Hi,

I had a simillar situation, needing to access the site via basic auth only if the client's IP is not in a whitelist. You can accomplish this with a trick:
1. Create a dummy route in your application that will always ask for basic auth.
2. Create another dummy route, that will return an HTML page, that will do a redirect from Javascript to the path at point 1.
3. Add a Traefik errors middleware, that captures 403 errors and the service is the path created at point 2.
4. Add a Traefik route for this path with a rule that matches the path and with a chained middleware: Basic auth and redirect to site's home path
5. Add a Traefik route with a check if the request has "Authorization" header. Attach the basic auth middleware to this route.
6. Add your usual Traefik route. Attach to this route a chained middleware, with the errors middleware at point 3 and the IP whitelist middleware. Attention, the order is important: the first middleware in the sequence must be the errors one, so it will capture 403 error codes from the IP whitelist middleware.

Hope this helps.

jayheinlein:
Adding another use-case. Sadly, I don't think the above workaround covers me.

We have multiple clients, each having access to a subset of our services. We want to or-chain multiple whitelists depending on which client(s) have access to which service. This would allow both easy readability and single source-of-truth for each whitelist instead of updating all service whitelists any time we need to change a client whitelist.

For instance -
- Client A can access service 1 and 2
- Client B can only access service 2
- Service 1 has a route that uses the Client A whitelist
- Service 2 has a route that uses Client A whitelist or Client B whitelist.

[...]

PR #7292, opened 2020-09-15
Add exceptions to basic auth and forward auth middlewares 

maksadbek:
### What does this PR do?
This PR adds enhancement for basic and forward authentication middlewares to set list of IPs and subnets in the configurations  to add exceptional logic for certain IPs or for IPs from subnets.

In other words, if the user sets the subnet in the allow list, users from this subnet can bypass the basic or forward authentication. 

The source of client's IP address can be a value from the header of the request. This works on if the `clientIPSourceHeader` is populated with some value. Otherwise, the host from `request.RemoteAddr` is used.

This PR is intended to close the following issue: https://github.com/containous/traefik/issues/4429  

Currently, I am interested in to get a feedback about the implementation and feature. When everything is ok, I will update the documentation.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/dynamic/fixtures/sample.toml
+++ b/pkg/config/dynamic/fixtures/sample.toml
@@ -406,6 +406,12 @@
         status = ["foobar", "foobar"]
         service = "foobar"
         query = "foobar"
+    [http.middlewares.Middleware22]
+      [http.middlewares.Middleware22.forwardAuth]
+        address = "https://foo.bar"
+        trustForwardHeader = true
+        authResponseHeaders = ["foo", "bar"]
+        allowList = ["192.168.0.0/24"]
   [http.services]
     [http.services.Service0]
       [http.services.Service0.loadBalancer]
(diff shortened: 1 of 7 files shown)


ldez:
Hello,

As I said in #4429, I think that a more global approach (#6007) seems a better way to handle the need.

We still need to think about it.

maksadbek:
@ldez 

Do you any suggestions how to implement the global approach ? Should I close this PR in this case ?

dkossako:
> Hello,
> 
> As I said in #4429, I think that a more global approach (#6007) seems a better way to handle the need.
> 
> We still need to think about it.

There are no moves in OR-chained feature request since 2019. Uploaded change would solve a lot of our troubles related to VPNs, internal networks, etc.

rtribotte:
Hello @maksadbek,

Indeed, we prefer to close this PR and continue to discuss about a better approach in the related issue #4429.

Thanks anyway for your time and contribution, looking forward talking with you about this in the issue.

jr0dd:
This is also related to https://github.com/traefik/traefik/issues/6007.

PR #7308, opened 2020-09-18
Add metrics response recorders that implement http.Pusher

### LINKED ISSUE #7303: http.Pusher interface not implemented by plugin middleware's response writer

Bug

### What did you do?

I'm writing a [middleware plugin](https://github.com/pipe01/plugin-h2push) for HTTP2 pushing, [Checking](https://github.com/pipe01/plugin-h2push/blob/d29f3fab0b384a42e41fb898be4d40276ba8346d/h2push.go#L68) if the response writer supports pushing through the `http.Pusher` interface always returns false, even when the downstream service communicates using HTTP2.

`p, ok := rw.(http.Pusher); ok == true` in https://github.com/pipe01/plugin-h2push/blob/d29f3fab0b384a42e41fb898be4d40276ba8346d/h2push.go#L68

```
Version:      2.3.0-rc6
Codename:     picodon
Go version:   go1.15.2
Built:        2020-09-16T15:20:51Z
OS/Arch:      linux/amd64
```

Static:
```yaml
entryPoints:
    web:
        address: :80
    websec:
        address: :443

providers:
    file:
        filename: traefik.dyn.yaml

pilot:
    token: "****"

experimental:
    devPlugin:
        goPath: C:\Users\piper\go
        moduleName: github.com/pipe01/plugin-h2push
```

Dynamic:
```yaml
http:
  services:
    webserver:
      loadBalancer:
        servers:
        - url: 'http://localhost:5000'

  routers:
    my-router:
      entrypoints:
        - web
        - websec
      rule: "Host(`localhost`)"
      service: webserver
      tls: true
      middlewares:
        - push

  middlewares:
    push:
      plugin:
        dev:
          files:
          - url: url1
            match: match1
```

pipe01:
### What does this PR do?

This PR makes the metrics response recorder `http.ResponseWriter` implementation also implement the `http.Pusher` interface whenever the inner response writer supports it.

### Motivation

I'm writing a [plugin](https://github.com/pipe01/plugin-h2push) for HTTP2 pushing, and found that the response writer passed to the middleware didn't implement the `http.Pusher` interface even when pushing *was* supported since it is of type `*metrics.responseRecorderWithCloseNotify`. Related to #7303 and #906.

### More

- [ ] Added/updated tests


pkg/middlewares/metrics/recorder.go
```diff
@@ -34,12 +44,27 @@ type responseRecorderWithCloseNotify struct {
...
 }
 
+type responseRecorderWithH2push struct {
+	*responseRecorder
+}
+
+type responseRecorderWithCloseNotifyAndH2push struct {
+	*responseRecorder
+	*responseRecorderWithCloseNotify
+	*responseRecorderWithH2push
```
pipe01, line 54: I did it this way in order to not duplicate the methods below, however I don't know if this is the best approach.

rtribotte:
Hello @pipe01,

Thank you for your contribution and sorry for the late answer.

In order to be able to comply with HTTP2 push, the support of the `http.Pusher` interface should be done in all needed places.
We saw that you addressed it for the metrics middleware, but there are much more places to make that change.

It seems that important work remains to make Traefik HTTP2 push capable.

We also have in mind that is probable that the Chromium team will ([remove the support for HTTP2 push](https://groups.google.com/a/chromium.org/g/blink-dev/c/K3rYLvmQUBY/m/vOWBKZGoAQAJ?pli=1)), and we are looking forward to seeing their conclusion.

Unfortunately, because it's been quite some time you have opened this PR, I prefer to decline it for now and reopen it once we'll be convinced that we have to add this feature to Traefik.

pipe01:
Given the likely discontinuation of the push feature by the groups I understand it may not be worth it to implement it within Traefik at this point, so I agree with your decision.

PR #7309, opened 2020-09-18
fix: pilot metrics unit for req duration.

ldez:
### What does this PR do?

pilot metrics unit for req duration.

### Motivation

use `ms` instead of `s`

### More

- [ ] Added/updated tests


mmatur (approved):
LGTM

PR #7332, opened 2020-09-23
Fix consul catalog router tag example

### LINKED ISSUE #7330: seems wrong document about consul catalog provider usage

bug

### What did you do?

https://docs.traefik.io/providers/consul-catalog/

the config exampl say after enable consul catalog then i shoud tag service with below tag

```
- traefik.http.services.my-service.rule=Host(`example.com`)
```

image

but it does not work until i changed to 

```
- traefik.http.routers.my-service.rule=Host(`example.com`)
```

and i double checked https://docs.traefik.io/v2.3/reference/dynamic-configuration/consul-catalog/
it does not have labels like `traefik.http.services.my-service.rule` but only `traefik.http.routers.my-service.rule`

then i realized this may be a document issue. 

please confirm if i am right since i waste hours find the reason.
thank you.

rtribotte:
### What does this PR do?

Fixes consul catalog router tag example.

### Motivation

Fixes #7330 

### More

- [ ] ~Added/updated tests~

### DIFF
--- a/docs/content/providers/consul-catalog.md
+++ b/docs/content/providers/consul-catalog.md
@@ -29,7 +29,7 @@ Attach tags to your services and let Traefik do the rest!
     Attaching tags to services
 
     ```yaml
-    - traefik.http.services.my-service.rule=Host(`example.com`)
+    - traefik.http.routers.my-router.rule=Host(`example.com`)
     ```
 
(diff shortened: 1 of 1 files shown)


ldez (approved):
LGTM

PR #7336, opened 2020-09-23
Add support for ECS Task metadata in ECS provider

yacchi:
### What does this PR do?

This request adds for ECS Task metadata support in ECS provider.

### Motivation

I used the docker provider with AWS ECS on EC2 and migrate to AWS ECS on Fargate now.
However, the ECS provider required additional IAM policy and cluster configuration.

This option makes it easier to execute a single task workload.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/providers/ecs.md
+++ b/docs/content/providers/ecs.md
@@ -81,6 +81,35 @@ Search for services in clusters list.
 - If set to `true` the configured clusters will be ignored and the clusters will be discovered.
 - If set to `false` the services will be discovered only in configured clusters.
 
+### `useTaskMetadata`
+
+_Optional, Default=false_
+
+```toml tab="File (TOML)"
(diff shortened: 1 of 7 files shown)


grigorov (approved):

kevinpollet:
Hello @yacchi,

We have looked at your PR, and we are not sure to understand, why is it better to use the Task metadata instead of defining the proper IAM role with the AWS client?   

As a side note, we are a bit worried about maintaining the HTTP client on our side and not using something provided by AWS.

kevinpollet:
Hello @yacchi,

After looking deeper and having a discussion on this feature, we are not willing to introduce a specific behaviour for single task setups. This would be too specific and introducing a new configuration to handle this case could be confusing.

Unfortunately, we will close this PR accordingly.

PR #7337, opened 2020-09-23
Upgrade Instana tracer and make process profiling configurable

cedricziel:
### What does this PR do?

Upgrade the instana go-sensor dependency and make the continous process
profiler configurable.

### Motivation

The profiling should be configurable.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/observability/tracing/instana.md
+++ b/docs/content/observability/tracing/instana.md
@@ -1,6 +1,6 @@
 # Instana
 
-To enable the Instana:
+To enable the Instana tracer:
 
 ```toml tab="File (TOML)"
 [tracing]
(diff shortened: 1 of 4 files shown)


kevinpollet:
Hi @cedricziel,

After reviewing your PR the [auto profiling](https://www.instana.com/docs/ecosystem/go/#autoprofile) option enabled with `EnableAutoProfiling` is in beta. As this feature is in beta we will not merge your PR for now because in case of removal this will introduce a breaking change.

Do you have more information on that?

kevinpollet:
Hi @cedricziel,

As said in the previous answer, the `auto profiling` feature is in beta, and we're not going to accept this pull request until it reaches a stable state. 

We will close this pull request accordingly but don't hesitate to reopen it when this feature reaches a stable state.

Thanks anyway.

PR #7345, opened 2020-09-24
Fix default value of docker client timeout

kevinpollet:
### What does this PR do?

This PR fixes the default value of the docker client timeout. By default, timeout is now set to `0` which means no timeout. This is also the value defined in the docker client [here](https://github.com/moby/moby/blob/ecdb0b22393bb669325099320d26d18687425e5f/client/client.go#L158).

### Motivation

Fixes the wrong default value which leads to `ERROR` in the logs:

```
ERRO[2020-09-24T15:32:21+02:00] Provider connection error context deadline exceeded (Client.Timeout or context cancellation while reading body), retrying in 552.330144ms  providerName=docker
```

### More

- ~[ ] Added/updated tests~

### DIFF
--- a/docs/content/providers/docker.md
+++ b/docs/content/providers/docker.md
@@ -496,7 +496,7 @@ Defines the polling interval (in seconds) in Swarm Mode.
 
 ### `httpClientTimeout`
 
-_Optional, Default=32_
+_Optional, Default=0_
 
 ```toml tab="File (TOML)"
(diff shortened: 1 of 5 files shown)


dtomcej (approved):
LGTM
:shipit:

ldez (approved):
LGTM

rtribotte (approved):
LGTM ok

PR #7355, opened 2020-09-25
acme: Fix race condition in LocalStore during saving.

walkline:
### What does this PR do?

Fixes race condition in the `acme` `LocalStore` during saving.

### Motivation

Fix the bug and make a contribution :)

### More

- [x] Added/updated tests

### Additional Notes

I'm not sure about tests with real files.

### DIFF
--- a/pkg/provider/acme/local_store.go
+++ b/pkg/provider/acme/local_store.go
@@ -31,10 +31,16 @@ func NewLocalStore(filename string) *LocalStore {
 
 func (s *LocalStore) save(resolverName string, storedData *StoredData) {
 	s.lock.Lock()
-	defer s.lock.Unlock()
-
 	s.storedData[resolverName] = storedData
-	s.saveDataChan <- s.storedData
+	s.lock.Unlock()
+
+	s.lock.RLock()
+	storedDataCopy := s.unSafeCopyOfStoredData(s.storedData)
+	s.lock.RUnlock()
+
+	// we cannot pass s.storedData directly, map is reference type and as result
+	// we can face with race condition, so we need to work with objects copy
+	s.saveDataChan <- storedDataCopy
 }
 
 func (s *LocalStore) get(resolverName string) (*StoredData, error) {
@@ -81,7 +87,10 @@ func (s *LocalStore) get(resolverName string) (*StoredData, error) {
 				}
 				if len(certificates) < len(storedData.Certificates) {
 					storedData.Certificates = certificates
-					s.saveDataChan <- s.storedData
+
+					// we cannot pass s.storedData directly, map is reference type and as result
+					// we can face with race condition, so we need to work with objects copy
+					s.saveDataChan <- s.unSafeCopyOfStoredData(s.storedData)
 				}
 			}
 		}
@@ -109,6 +118,15 @@ func (s *LocalStore) listenSaveAction() {
 			}
 		}
 	})
+}
+
+// unSafeCopyOfStoredData creates maps copy of source. Is not thread safe, you should use `s.lock`.
+func (s *LocalStore) unSafeCopyOfStoredData(source map[string]*StoredData) map[string]*StoredData {
+	result := map[string]*StoredData{}
+	for k, v := range source {
+		result[k] = v
+	}
+	return result
 }
 
 // GetAccount returns ACME Account.
--- a/pkg/provider/acme/local_store.go
+++ b/pkg/provider/acme/local_store.go
@@ -31,16 +31,13 @@ func NewLocalStore(filename string) *LocalStore {
 
 func (s *LocalStore) save(resolverName string, storedData *StoredData) {
 	s.lock.Lock()
-	s.storedData[resolverName] = storedData
-	s.lock.Unlock()
+	defer s.lock.Unlock()
 
-	s.lock.RLock()
-	storedDataCopy := s.unSafeCopyOfStoredData(s.storedData)
-	s.lock.RUnlock()
+	s.storedData[resolverName] = storedData
 
 	// we cannot pass s.storedData directly, map is reference type and as result
 	// we can face with race condition, so we need to work with objects copy
-	s.saveDataChan <- storedDataCopy
+	s.saveDataChan <- s.unSafeCopyOfStoredData()
 }
 
 func (s *LocalStore) get(resolverName string) (*StoredData, error) {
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

Could you provide more information about the condition of the race?
Could you provide a minimal reproducible example to get the bug?

walkline:
Hey @ldez. Sure.

So the problem is in the next code:

```
func (s *LocalStore) save(resolverName string, storedData *StoredData) {
// ...

	// so here we making map modification (there is a lock)
	s.storedData[resolverName] = storedData

	// here we passing map s.storedData
	s.saveDataChan <- s.storedData
}

//...
func (s *LocalStore) listenSaveAction() {
	safe.Go(func() {
		logger := log.WithoutContext().WithField(log.ProviderName, "acme")
		for object := range s.saveDataChan {

			// here we iterating that map (we don't have lock here)
			data, err := json.MarshalIndent(object, "", "  ")
			if err != nil {
				logger.Error(err)
			}
//...
```

We are writing to the map in one goroutine and reading from another.
In that example, during writing, we have mutex lock, but during reading/iteration we haven't.
Yeah, we used channel to receive that map. But the map is a reference type object, so we actually working with the same object.

To reproduce this issue you can use the next test:
```
func TestRaceCondition(t *testing.T) {
	s := NewLocalStore(localStorageFileName)

	go func() {
		for {
			s.SaveAccount("traefik.wtf", &Account{
				Email: "email1@email.com",
			})
		}
	}()
	go func() {
		for {
			s.SaveAccount("traefik.wtf", &Account{
				Email: "email2@email.com",
			})
		}
	}()

	time.Sleep(time.Second*20)
}
```
In my case, I have the next output: `fatal error: concurrent map iteration and map write`.

pkg/provider/acme/local_store.go
```diff
@@ -111,6 +120,15 @@ func (s *LocalStore) listenSaveAction() {
...
 
+// unSafeCopyOfStoredData creates maps copy of source. Is not thread safe, you should use `s.lock`.
+func (s *LocalStore) unSafeCopyOfStoredData(source map[string]*StoredData) map[string]*StoredData {
+	result := map[string]*StoredData{}
+	for k, v := range source {
+		result[k] = v
+	}
+	return result
+}
+
```
rtribotte, line 131: Hello @walkline,

Thanks for your contribution and interest in Traefik!

I would take advantage of your method to make it thread safe.
It would be clearer i think.

```suggestion
// copyOfStoredData creates maps copy of source.
func (s *LocalStore) copyOfStoredData() map[string]*StoredData {
	s.lock.RLock()
	defer s.lock.RUnLock()
	
	result := map[string]*StoredData{}
	for k, v := range s.storedData {
		result[k] = v
	}

	return result
}

```

pkg/provider/acme/local_store.go
```diff
@@ -81,7 +87,10 @@ func (s *LocalStore) get(resolverName string) (*StoredData, error) {
 				}
 				if len(certificates) < len(storedData.Certificates) {
 					storedData.Certificates = certificates
-					s.saveDataChan <- s.storedData
+
+					// we cannot pass s.storedData directly, map is reference type and as result
+					// we can face with race condition, so we need to work with objects copy
+					s.saveDataChan <- s.unSafeCopyOfStoredData(s.storedData)
 				}
```
rtribotte, line 94: To follow up my first comment.

```suggestion

					// we cannot pass s.storedData directly, map is reference type and as result
					// we can face with race condition, so we need to work with objects copy
					s.saveDataChan <- s.copyOfStoredData()
				}
```

pkg/provider/acme/local_store.go
```diff
@@ -31,10 +31,16 @@ func NewLocalStore(filename string) *LocalStore {
...
-	s.saveDataChan <- s.storedData
+	s.lock.Unlock()
+
+	s.lock.RLock()
+	storedDataCopy := s.unSafeCopyOfStoredData(s.storedData)
+	s.lock.RUnlock()
+
+	// we cannot pass s.storedData directly, map is reference type and as result
+	// we can face with race condition, so we need to work with objects copy
+	s.saveDataChan <- storedDataCopy
```
rtribotte, line 43: To follow up my first comment.

```suggestion

	// we cannot pass s.storedData directly, map is reference type and as result
	// we can face with race condition, so we need to work with objects copy
	s.saveDataChan <- s.copyOfStoredData()
```

pkg/provider/acme/local_store.go
```diff
@@ -31,10 +31,16 @@ func NewLocalStore(filename string) *LocalStore {
 
 func (s *LocalStore) save(resolverName string, storedData *StoredData) {
 	s.lock.Lock()
```
rtribotte, line 33: To follow up my first comment.

```suggestion
	s.lock.Lock()
	defer s.lock.Unlock()
```

rtribotte (changes requested):

pkg/provider/acme/local_store.go
walkline: @rtribotte In this case, I think, we will have a deadlock because at line 41/47 we already made locking.

pkg/provider/acme/local_store.go
walkline: Hey @rtribotte. Yeah, with your suggestion, it looks clearer, but it becomes a little bit tricky. Before calling `copyOfStoredData()`, you need to make sure that our mutex unlocked. In our case, it locked, so we will, probably, have a deadlock.

rtribotte, line 131: Re Hello @walkline,

You are right about the deadlock.
Still, i would use the receiver since i do not see side effect to that, and it's more consistent for a method.

```suggestion
// unSafeCopyOfStoredData creates maps copy of source. Is not thread safe, you should use `s.lock`.
func (s *LocalStore) unSafeCopyOfStoredData() map[string]*StoredData {
	result := map[string]*StoredData{}
	for k, v := range s.storedData {
		result[k] = v
	}
	return result
}
```

pkg/provider/acme/local_store.go
```diff
@@ -31,10 +31,16 @@ func NewLocalStore(filename string) *LocalStore {
...
-	defer s.lock.Unlock()
-
 	s.storedData[resolverName] = storedData
-	s.saveDataChan <- s.storedData
+	s.lock.Unlock()
+
+	s.lock.RLock()
+	storedDataCopy := s.unSafeCopyOfStoredData(s.storedData)
+	s.lock.RUnlock()
```
rtribotte, line 39: Here it seems unnecessary to differentiate the lock usage, so reintroduce the `defer the s.lock.Unlock` seems better to me, in place of this code.

rtribotte (changes requested):

walkline:
@rtribotte thanks for feedback! Agreed with your latest suggestions.

ldez (approved):
Thanks :+1:

rtribotte (approved):
LGTM 👍

jbdoumenjou (approved):
LGTM

PR #7359, opened 2020-09-26
Extend marathon port discovery to allow port names as identifier

basert:
Every marathon port has a name attached. Currently the "index:" prefix is supported to select a marathon port by it's index.
Extend this lookup to search for the port name if the label prefix is "name:".

### What does this PR do?

Extend marathon port discovery to allow the "name:" prefix to select a port.

### Motivation

We at [ESL](https://about.eslgaming.com/) use marathon, mesos-dns and nginx to discover our microservices via DNS. mesos-dns uses the port names in it's DNS payload. We are currently evaluating traefik to replace this setup.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/user-guides/marathon.md
+++ b/docs/content/user-guides/marathon.md
@@ -17,10 +17,19 @@ Traefik tries to detect the configured mode and route traffic to the right IP ad
 Traefik also attempts to determine the right port (which is a [non-trivial matter in Marathon](https://mesosphere.github.io/marathon/docs/ports.html)).
 Following is the order by which Traefik tries to identify the port (the first one that yields a positive result will be used):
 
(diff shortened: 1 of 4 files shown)


basert:
Is there an update when this change can land in traefik?

basert:
Hey guys, it's been 2 months, can we get this merged? We want to start using traefik in our infrastructure, would be cool if we dont need to run out own fork.

rtribotte (approved):
Thanks 👍

ldez (approved):
Thanks :+1:

SantoDE (approved):
LGTM. Thanks 👼

kevinpollet (approved):
LGTM 👍

PR #7364, opened 2020-09-28
Fix blank webui on some browsers

matthieuh:
### What does this PR do?

Avoid to use `semver-regex` which fail on some browsers.

### Motivation

Fix https://github.com/traefik/traefik/issues/7357

### More

- [ ] Added/updated tests


obeyler:
@matthieuh may there is another pb as a get is done on Pilot url whereas the pilot is not enabled in the config

rtribotte (approved):
LGTM 👌

jbdoumenjou (approved):
LGTM :+1:

ldez (approved):
LGTM

PR #7368, opened 2020-09-29
Add Option to PassTLSClientCert to send only the first certificate

cnwaldron:
### What does this PR do?

Change the default behavior for PassTLSClientCert to only send the first client certificate and not join the additional certificates (if available). 
Added the option ```chain``` to the PassTLSClientCert middleware to provide the option to still have the client certificate chain to be sent if wanted.

### Motivation
Tried to use Keycloak's x509 authentication behind Traefik. The "," separator would cause Keycloak to error out while reading the escaped PEM from the header. This change is more inline with the behavior of other ingresses.

### More

- [x] Added/updated tests


ldez:
Hello,

Thank you for your contribution, but you cannot change the default behavior.
Could you change the default value of `chain` to `true`, thanks.

docs/content/middlewares/passtlsclientcert.md
```diff
@@ -403,17 +417,21 @@ In the example, it is the part between `-----BEGIN CERTIFICATE-----` and `-----E
     ML9n/4zmm1PMhzZHcEA72ZAq0tKCxpz10djg5v2qL5V+Oaz8TtTOZbPsxpiKMQ==
     -----END CERTIFICATE-----
     ```
-    
+
```
SuperSandro2000, line 420:
```suggestion

```

cnwaldron:
Changed 'chain' to 'ignoreChain' so the default is false.

cnwaldron:
@ldez What are the next steps for this PR? Is there anything else I need to do on my side? This issue is still hindering progress on my project.  
Thanks!

jbdoumenjou:
Hello @cnwaldron,

Thanks for your interest in Traefik.

Your PR has been labeled as status/2-needs-review: 
This label means there does not seem to be any obvious design issue with the PR, and the next step is careful code and documentation review by the maintainers.

Please read the [PR Guidelines](https://github.com/traefik/contributors-guide/blob/master/pr_guidelines.md) and
[PR status workflow](https://doc.traefik.io/traefik/v2.3/contributing/maintainers/#status-workflow) for more details.

jspdown:
Hello @cnwaldron :wave: 

We are currently in the process of validating the use case of this new option. To better understand what you tried to achieve, we will need more information. Could you tell us more about what you tried to achieve and how you configured keycloak, like, which provider is used and what is its configuration?

As there is no standard way of sending certificate information though headers and as every proxy uses a different format, there's some places where our format might not be understood. Keycloak already provides specific `x509cert-lookup` providers for `haproxy` and `apache` (https://www.keycloak.org/docs/latest/server_admin/#client-certificate-lookup) but nothing `traefik` specific.

cnwaldron:
Hi @jspdown 
We currently used Keycloak's x509 lookup behind NGINX. When we wanted to switch to using Traefik, the only change we made to the configuration was changing the "sslClientCert" value to "x-forwarded-tls-client-cert (https://www.keycloak.org/docs/latest/server_admin/#nginx-certificate-lookup-provider). The issue I ran into with this configuration was Keycloak was expecting just the client certificate and not the whole chain of certificates. The comma between certificates is considered an invalid character and would throw errors. I have Keycloak configured with a truststore that rebuilds the chain and validates the incoming certificate. This pull request was to allow the option to better fit with existing architecture with minimal changes. 

Please let me know if you need any more information or clarification. 
Thanks!

jspdown:
Nginx and Traefik format for passing client certs are really close indeed, the only exception being that Nginx can only pass a single cert. We understand your goal but we are not sure introducing an option for doing an implicit format conversion is the best approach. We think it's not Traefik's responsability to convert formats, either Traefik should use a different format or either it's an integration issue which should be addressed directly on Keycloak. 

If you think it worth reconsidering a different format for Traefik we would be happy to discuss it further with the community in a `proposal`. But right know, the best approach we can suggest will be to write a custom provider for Keycloak, as described in https://www.keycloak.org/docs/latest/server_admin/#other-reverse-proxy-implementations and https://www.keycloak.org/docs/latest/server_development/#_providers.

As a workaround, it's still possible to use Traefik Plugins to [rewrite the header](https://github.com/mrdrelar/traefik-plugin-rewriteheader) and keep only the first certificate in the chain. 

We really appreciate the time you've spent on this PR and we hope you understand our point of view.

Issue #7354, opened 2020-09-25
TCP File Provider is not updating ip's resolved from DNS


Enquier:
### What did you do?

We setup (in docker swarm) a akka based system that has to cluster using independently defined seeds and hosts. These systems have a webserver and a TCP application level connection. When one of those nodes restarts, Traefik's TCP router appears to be caching the IP of that node (not resolving the DNS name on the fly) and is still pointing to the static IP. Since this is a swarm system there is no way for us to know the IP name a-priori

### What did you expect to see?

Traefik addressing the node by DNS (or polling periodically) and not caching the IP

### What did you see instead?

Traefik caches the IP and doesn't resolve or poll the dns name 

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      2.2.10
Codename:     chevrotin
Go version:   go1.14.8
Built:        2020-09-04T15:53:27Z
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml

 
dynamic-service.yml
http:                                                                       
  routers:                                                                 
    https:                                                     
      entryPoints: ["https"]                                                
      service: https@file                                      
      tls: {}                                                              
      rule: Host(`hostname.com`) && PathPrefix(`/api`)
  services:
    https:
       loadBalancer:
          servers:
          - url: https://seed-1:8111/
          - url: https://seed-2:8111/
          - url: https://node-1:8111/
tcp:
  routers:
    tcp:
      entryPoints: ["tcp_endpoint"]
      service: tcp@file
      rule: HostSNI(`*`)
      tls:
        passthrough: true
  services:
    tcp:
      loadBalancer:
        servers:
        - address: seed-1:10002
        - address: seed-2:10002
        - address: node-1:10002
```

### If applicable, please paste the log output in DEBUG level (`--log.level=DEBUG` switch)

```
Traefik logs:
time="2020-09-02T15:58:19Z" level=error msg="Error while connection to backend: dial tcp 10.0.3.245:10002: connect: no route to host",
time="2020-09-02T16:00:48Z" level=error msg="Error while connection to backend: dial tcp 10.0.3.249:10002: connect: no route to host",
time="2020-09-02T16:01:47Z" level=error msg="Error while connection to backend: dial tcp 10.0.3.251:10002: connect: no route to host",
time="2020-09-02T16:05:41Z" level=error msg="Error while connection to backend: dial tcp 10.0.3.245:10002: connect: no route to host",
[...]

PR #7370, opened 2020-09-29
Improve service name lookup on TCP routers

### LINKED ISSUE #7167: "no available server" when connect to a container using definition of IngressRouteTCP. Restart helps. 2.2.8

Bug

### What did you do?
Traefik 2.2.8. I create IngressRouteTCP with tls passthrough enabled.  In a few minutes I try to connect to the container and my app says "context deadline exceeded" and in traefik logs "no available server". On the traefik dashboard everything is ok.  Problem is solved once I restart traefik pod. 
Works fine on 2.0.1
```yml
- apiVersion: traefik.containo.us/v1alpha1
  kind: IngressRouteTCP
  metadata:
    creationTimestamp: "2020-08-05T19:55:41Z"
    generation: 1
    labels:
      node: a2
      resourceName: org1peers
    name: peerset-org1peers-a4-org1peers
    namespace: default
    ownerReferences:
    - apiVersion: peerset.x/v1beta
      controller: true
      kind: PeerSet
      name: org1peers
      uid: 3f91c80c-ebd6-4a1c-9d64-b0de29894853
    resourceVersion: "55267"
    selfLink: /apis/traefik.containo.us/v1alpha1/namespaces/default/ingressroutetcps/peerset-org1peers-a2-org1peers
    uid: d6145fa0-4560-476a-8e74-98d7bb90f1bd
  spec:
    entryPoints:
    - websecure
    routes:
    - match: HostSNI(`peerset-org1peers-a4-org1peers.proxy.bank1.x.com`)
      services:
      - name: peerset-org1peers-a4-org1peers
        port: 9111
    tls:
      passthrough: true

---

apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2020-08-05T19:55:34Z"
  labels:
    app: proxy
    node: a1
    resourceName: org1peers
  name: peerset-org1peers-a4-org1peers
  namespace: default
  ownerReferences:
  - apiVersion: peerset.x/v1beta
    controller: true
    kind: PeerSet
    name: org1peers
    uid: 3f91c80c-ebd6-4a1c-9d64-b0de29894853
  resourceVersion: "55194"
  selfLink: /api/v1/namespaces/default/services/peerset-org1peers-a1-org1peers
  uid: fc68509b-e867-46d3-be3a-eb262507b2d6
spec:
  externalName: peerset-org1peers-a4.org1peers.default.svc.cluster.local #<------ a pod of some StatefulSet like controller (have crd with similar way of handling dns for a container using obj.ServiceName and pod.Host)
  ports:
  - name: peerset-org1peers-a4
    port: 9111
    protocol: TCP
    targetPort: 9111
  sessionAffinity: None
  type: ExternalName
status:
  loadBalancer: {}
```

I expected to be connected to the pod in a few seconds after IngressRouteTCP is created with no restart of the traefik pod.

### What did you see instead?

context deadline exceeded 

```
2.2.8
on 2.0.1 works fine 
```

```
Kubernetes GKE 1.16.13-gke.1
Default installation of traefik using helm `helm install traefik2 traefik/traefik --namespace kube-system`
```

After creation of IngressRouteTCP there is an error 
```
In service \"default-peerset-org1peers-a4-org1peers-425aaf203a1844d04c5f@kubernetescrd\" server \"peerset-org1peers-a4.org1peers.default.svc.cluster.local:9111\": lookup peerset-org1peers-a4.org1peers.default.svc.cluster.local on 10.52.0.10:53: no such host
```
It's because of service defined in ingress has `ExternalName` type and it points to a pod of a StatefulSet which could being creating at the moment. In a few seconds `nslookup peerset-org1peers-a4.org1peers.default.svc.cluster.local` returns host inside of the traefik container. 
image

When I connect -> 
`" level=error msg="no available server"`
```
(paste your output here)
```

ddtmachado:
### What does this PR do?

Fix the service name resolution to IP addresses for TCP services referenced by hostname/DNS.

Fixes #7167 #7354 

### Motivation

Currently, for TCP routers the service name resolution is done during the router setup (Proxy object) and never updated again, in this case, any changes to the actual addresses behind that hostname will have no effect unless the Traefik instance is restarted.

Moving the resolution to the connection handling allows for a fast reaction upon address changes, every connection to the server will trigger a lookup of the hostname. 

Pros of this approach:

* Fast reaction to changes in addresses (per connection)
* Not dependent on pooling (either by the provider config or independent check)
* Should work for all providers (tested with File and Kubernetes)

Cons:

* ~~Adds an address lookup per connection~~

### More

- [X] Added/updated tests

### Additional Notes

I didn't notice any meaningful impact in latency while testing (basic checks with ncat), compared with v2.2, but a more intensive performance check with thousands of short-lived connections could be done if needed.


kopaygorodsky:
in #7167 traefik updates IP when config is reloaded, it just ignores some of k8s Endpoints events. It explained [here](https://github.com/traefik/traefik/issues/7167#issuecomment-688900376). The problem will disappear once reloading of the config in k8s provider is fixed.

ddtmachado:
@kopaygorodsky yes, but a config reload is not a solution when working with a DNS name external service, as kubernetes will never trigger a config update since the object never changes (service definition without endpoints). In this case you would have to rely on restarting Traefik or deleting and creating the router again to update the actual addresses.

kopaygorodsky:
I thought we can watch for Endpoints resource, can't we?

ddtmachado:
@kopaygorodsky an [external name](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) service generates no endpoints in kubernetes. You can create them manually but it would be ignored by Traefik in this case anyway.

For external name services Traefik resolves the `externalName` key to an actual TCP address that will be stored at the router service servers list. That was done only during the router setup and this PR changes it to the router's proxy server inside Traefik responsible for handling requests, thus resolving the name for each new request.

TCP based services will usually connect once and keep the connection until the work is done, like databases for example, and for that reason I guess it won't have a negative impact. Even for multiple distinct requests I did not see a noticeable increase in latency, but I would like validation from more community members or senior maintainers.

kopaygorodsky:
oh, I got where I was wrong. Thanks a lot for answering :)

pkg/tcp/proxy.go
```diff
@@ -10,28 +10,29 @@ import (
...
-
-	return &Proxy{target: tcpAddr, terminationDelay: terminationDelay}, nil
+	return &Proxy{address: address, terminationDelay: terminationDelay}, nil
 }
 
 // ServeTCP forwards the connection to a service.
 func (p *Proxy) ServeTCP(conn WriteCloser) {
 	log.Debugf("Handling connection from %s", conn.RemoteAddr())
 
+	tcpAddr, err := net.ResolveTCPAddr("tcp", p.address)
```
dtomcej, line 26: Could we store the` target   *net.TCPAddr` in the `Proxy` struct, with a timeout time? if the timeout is reached, then re-resolve the TCP address?

That way we could "cache" the address for a short-ish period of time, reduce DNS lookups, and would probably reduce the overhead of multiple resolutions.

I was thinking maybe a 5 second timeout? just something that if 1000 requests come in at once, we still only resolve once, but if 1000 come in over the course of a minute, we are still updating the address 20 times, so it will respond to dynamic changes fairly quickly.  We may want this to be configurable, idk.

Thoughts?

ddtmachado: Yes, this could work well If we make it configurable. We could even allow a zero value for "never update" that would be like keeping the current behavior.

ddtmachado: @dtomcej I added this but instead of never updating I only left an always updating alternative with a negative value, I think it makes more sense.

If this is reasonable I'll add the missing docs, examples and everything else

dtomcej: Looks good to me.

jspdown:
@ddtmachado Thanks for your PR, It looks promising!

I wonder if it wouldn't be a more understandable behavior to not have a default value for `addrLookupCache`. If the user don't specify the `addrLookupCache` option, it would just resolve the address for every connection. If this option gets configured, the users chose its value and is aware of the side effects. As it's an optimization, it doesn't have to be enabled by default.

Also, what do you think about changing the `addrLookupCache` into a duration? If there's no default value it doesn't have to be a pointer to an int and we don't need to allow negative values.

jspdown:
@ddtmachado After discussing with the team, we think having a cache adds more complexity than it helps. Right now the current implementation has some concurrency issues around the `p.target` and `p.resolvedAt`. Fixing concurrency while still avoiding simultaneous resolves and connection bottle neck is doable but would make the code even more complex. The only cases where you want to use this resolution is when using the file provider or with ExternalName services on Kubernetes. As it is an edge case, not caching resolutions is less of a problem.

What do you think about doing the resolution only if the given `address` is a service name? The most common use case won't be impacted. If users want to avoid stressing too much there DNS server it is still possible to rely on a local DNS cache using Kubernetes [nodelocaldns](https://kubernetes.io/docs/tasks/administer-cluster/nodelocaldns/) or more generally with [dnsmasq](http://www.thekelleys.org.uk/dnsmasq/doc.html).

ddtmachado:
@jspdown yes, it is not synced but it does not cause any race condition as well, in the worst case scenario it will trigger more lookups than necessary. Being so I agree that the complexity to add synchronicity is not worth in this case but I wouldn't mind keeping it like that.

> What do you think about doing the resolution only if the given address is a service name?

I wouldn't mind doing this instead, but then I would prefer to do it [outside the Proxy](https://github.com/traefik/traefik/blob/d38d11f02e47a1e81edfecc6fcf4a1c42390a81e/pkg/server/service/tcp/service.go#L61) to properly resolve the name into multiple ip addresses and avoid relying on the DNS round robin to load balance, wdyt?

ddtmachado:
@jspdown Actually forget my suggestion on doing it outside as we would have to spawn another routine to keep it updated through time and it would be more complex than fixing concurrency 😅

jspdown:
@ddtmachado You are right there's no race condition. It's just that this concurrency issue shows us that implementing a DNS caching mechanism is not that trivial. Instead of committing right now on how we want to configure the cache we would like more user inputs. The workaround (installing a DNS cache) currently provides a more capable cache than ours, so we should let users open issues describing their use-case and iterate on it.

jspdown (commented):
Sorry for the delay. I just finished testing the PR. It works great. One last suggestion:

pkg/tcp/proxy.go
```diff
@@ -31,6 +30,15 @@ func (p *Proxy) ServeTCP(conn WriteCloser) {
 	// needed because of e.g. server.trackedConnection
 	defer conn.Close()
 
+	if p.target == nil || p.targetIsHostname() {
```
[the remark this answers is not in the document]
ddtmachado: I did it while keeping the full source address and just setting a bool to define if we should refresh the TCP lookup or not.

pkg/tcp/proxy_test.go
```diff
@@ -79,3 +82,83 @@ func TestCloseWrite(t *testing.T) {
 	require.Equal(t, int64(4), n)
 	require.Equal(t, "PONG", buffer.String())
 }
+
+func mustHaveExternalNetwork(t *testing.T) {
+	if runtime.GOOS == "js" {
```
jbdoumenjou, line 87: What is it for?

ddtmachado: To make sure we don't try to run this test in an OS that doesn't support external networks

jspdown (approved):
LGTM :+1:

jbdoumenjou (approved):
LGTM :+1:

pkg/tcp/proxy_test.go
```diff
@@ -79,3 +81,72 @@ func TestCloseWrite(t *testing.T) {
...
 }
+
+func TestLookupAddress(t *testing.T) {
+	testCases := []struct {
+		desc       string
+		address    string
+		expectSame bool
+	}{
+		{
+			desc:       "IP don't need refresh",
```
kevinpollet, line 92:
```suggestion
			desc:       "IP doesn't need refresh",
```

kevinpollet (approved):
LGTM

kopaygorodsky:
woohoo! Thx for your effort!

Issue #7197, opened 2020-08-19
[ECS Provider] Scrape configs from multiple AWS regions in ECS provider


leshik:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Would be great if `--providers.ecs.region` could be specified multiple times and Traefik could scrape ECS services from different regions. I have multiple clusters in 3 regions and trying to unify URLs by utilizing DNS with geolocation policy and Traefik instance in every region.

kevinpollet:
Hi @leshik,

Thanks for your interest in Traefik!

Each ECS cluster is in a different region, and routing is done through a DNS which routes the client to the right region in function of some criteria. Each cluster in a region has a Traefik instance used to discover and route the traffic to the right ECS service in the cluster.

Why do you need Traefik to scrape config from another region?
What do you mean by unifying URL?

Btw, as you might guess the `--providers.ecs.region` parameter is directly bound to the region of the AWS client which accepts only one region. This change will imply to have a list of AWS client in the provider, and I'm not sure that it is something we want.

leshik:
@kevinpollet consider that you have *.example.com, and it points to traefik instances in each region using geolocation DNS. Then actual tenant might be in any region, but not all. In case the request comes from the tenant's region, all is fine. But if it's not, traefik should be able to proxy the request to the corresponding tenant.

kevinpollet:
Hi @leshik,

In a typical AWS multi-region architecture, the infrastructure is replicated between regions, and the routing is done with DNS and geolocation. In your use-case, the routing criteria might be the tenant ID which will always route the client to the region containing the requested tenant.

The use case you are describing seems to be an edge-case, and that's something we do not plan to support for the moment.

Therefore, we will decline your proposal and close the issue accordingly.

PR #7376, opened 2020-10-02
ECS: Support for multiple AWS regions

leshik:
This is to address (closed) #7197.

@kevinpollet I understand that you might not want to merge this, but this change is quite lightweight and it breaks nothing. I created this PR in case someone else needs this functionality.


jbdoumenjou:
Hello,

Kevin's answer was the result of a team consultation.
The reasons given remain valid: the use case is a borderline case that is not recommended.
Moreover, this change is breakable at the configuration level.

Anyway, thank you for your work.

PR #7395, opened 2020-10-07
ECS: Add an option to set service names from `traefik.ecs.service` task tags

leshik:
### What does this PR do?

This adds an option to set predictable service names in Traefik using ECS task tags `traefik.ecs.service`.

### Motivation

We use `CloudFormation` to deploy AWS resources, therefore AWS service names are generated automatically, and they can change in case the service needs to be recreated. We depend on the ECS provider for service discovery, but we also use the Redis provider at the same time to provision routers and rules. Thus we need more predictable service names. AWS ECS allows propagating tags from the service (or task definition) down to individual tasks. This PR uses this option to set service names in Traefik from `traefik.ecs.service` task tags.

### DIFF
--- a/pkg/provider/ecs/ecs.go
+++ b/pkg/provider/ecs/ecs.go
@@ -72,6 +72,7 @@ const DefaultTemplateRule = "Host(`{{ normalize .Name }}`)"
 var (
 	_                    provider.Provider = (*Provider)(nil)
 	existingTaskDefCache                   = cache.New(30*time.Minute, 5*time.Minute)
+	existingTagsCache                      = cache.New(30*time.Minute, 5*time.Minute)
 )
 
 // SetDefaults sets the default values.
@@ -283,9 +284,12 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
 			return nil, err
 		}
 
+		taskTags := p.lookupTaskTags(ctx, client, tasks)
+
 		for key, task := range tasks {
 			containerInstance := ec2Instances[aws.StringValue(task.ContainerInstanceArn)]
 			taskDef := taskDefinitions[key]
+			tags := taskTags[key]
 
 			for _, container := range task.Containers {
 				var containerDefinition *ecs.ContainerDefinition
@@ -346,8 +350,17 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
 					}
 				}
 
+				name := fmt.Sprintf("%s-%s", strings.Replace(aws.StringValue(task.Group), ":", "-", 1), *container.Name)
+
+				// Use service name from the task's tag if it exists
+				for _, t := range tags {
+					if aws.StringValue(t.Key) == "traefik.ecs.service" {
+						name = aws.StringValue(t.Value) + "-" + *container.Name
+					}
+				}
+
 				instance := ecsInstance{
-					Name:                fmt.Sprintf("%s-%s", strings.Replace(aws.StringValue(task.Group), ":", "-", 1), *container.Name),
+					Name:                name,
 					ID:                  key[len(key)-12:],
 					containerDefinition: containerDefinition,
 					machine:             mach,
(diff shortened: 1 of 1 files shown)


rtribotte:
Hello @leshik,

Sorry for the delay and thanks for your contribution!

I have a question on how relying on such tag value is more predictable than the current naming strategy.

Today the Traefik service name is the combination of the normalized task group and the container name.
If the task group is not set, by default, the task group is the family name of the task definition (accordingly to this [documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement-constraints.html#task-groups)).
Either way, you'll always get something predictable for the name of your services.

Did I miss something, WDYT?

leshik:
@rtribotte have you ever used CloudFormation?
Here is a typical service name resulting after applying templates: `DS-ApiProxy-66KRHO4XK9BF-EcsService-ET26CPGWBOCM`. In case you make a change in your templates that results in the recreation of service, you'll get completely different alphanumeric strings added to the CloudFormation resource names.

rtribotte:
@leshik thanks for your quick feedback.

I never used CloudFormation, this is the reason for my questions. But it was clear in the first place that it generates service names with some random alphanumeric string, as you mentioned in the PR description.
Still, something is not clear for me, the service name here is the name of the ECS service, right?

I should have phrased my question this way: isn’t the container name (in the task definition) also containing a random alphanumeric string?

Because the service name would now be the combination of the tag value (if set) and the container name, I understand that the container name is more predictable than the task group in the task definition when using CloudFormation, is that right?

Nonetheless, if and when the PR would go in review status, I think there is a need for new tests that will yield documentation in addition to testing the tag logic.

leshik:
@rtribotte container name is set in the task definition, e.g. `nginx` etc. It doesn't contain random strings and is used to link containers in multi container tasks, as well as to set the container to route traffic to in ELB definition.

rtribotte:
@leshik Thanks for your feedback.
Can you share a typical CloudFormation task definition you are using?

leshik:
@rtribotte it can be as simple as this:

```yaml
  EcsTask:
    Type: AWS::ECS::TaskDefinition
    Properties:
      ExecutionRoleArn: !Ref EcsTaskExecutionRole
      Family: myapp-task
      RequiresCompatibilities: [EC2]
      ContainerDefinitions:
        - Name: nginx
          Image: nginx:stable
          Cpu: 256
          Memory: 256
          PortMappings:
            - ContainerPort: 80
              Protocol: tcp

  AppEcsService:
    Type: AWS::ECS::Service
    Properties:
      Cluster: !Ref Cluster
      TaskDefinition: !Ref EcsTask
      LaunchType: EC2
      DesiredCount: 4
      Tags:
        - Key: traefik.ecs.service
          Value: myapp
```

rtribotte:
Hello @leshik,

Thanks for your last feedback.

Apologies for the late answer, but i was finally able to test your pull request, and give it more thoughts with the team.
As you said, with CloudFormation, the task group is automatically set to the generated service name value, regardless of the family name value defined for the task.

Unfortunately, after discussion and while we definitely see the goal you wanted to achieve here, we will decline your pull request because of these reasons:

- Currently, naming a service is achieve with Docker labels, and we don't want to change this logic.
- Using tags as a new source of configuration for Traefik services can be confusing.
- Fetching the task tags would add traffic to the ECS API, and can possibly lead to reach the rate limit more frequently than before in some setups.

Hence, with Docker labels you have the possibility to name the services discovered by the ECS provider.
For instance, by only configuring the scheme: `traefik.http.services.{MyService}.loadbalancer.server.scheme: "http"`.

Also, while i can't confirm this, it might be possible to configure the task group using a custom AWS::Events::Rule with the `EcsParameters`.

rafaljanicki:
@rtribotte sorry for coming back here, but I do have another use case for using tags instead of labels

Currently, AWS allows only for specifying Docker labels when creating a task definition i.e. you can't modify those when running a task. Which is a problem in my usecase where I have a single task definition which I start for every user. To make it simple, imagine a Wordpress image which is the same for everyone. However, when I start the task (not a service), I give it appropriate tags to differentiate from the other tasks.

rtribotte:
Hello @rafaljanicki,

While I agree with you, I think this is not relevant regarding the concern here.
Using tags is a totally valid use case (I cannot argue against that), but when it comes to using it for Traefik autodiscovery configuration for service naming, it's not a straightforward answer.
Regarding the use case of this PR, CloudFormation would allow to easily redefine the task definition Docker labels value with custom variables.

PR #7396, opened 2020-10-07
Filter out k8s objects that belong to Helm

### LINKED ISSUE #7384: Cluster deployments Helm v2 -> v3 migration made traefik crash with OOMKill

Bug

### What did you do?

We have migrated all the deployments of a Kubernetes cluster from helm v2.x to v3.x (traefik itself is **not** deployed with Helm).

Helm v2 used to store Helm "releases" using **ConfigMaps**. In Helm v3 the "releases" are stored using **Secrets**.

The contents of theses new Secrets (we had around 700 of them) are quite large binaries (at least hundreds of KiB, but lesser than 1MiB which is kubernetes/etcd limitation) and it seems that Traefik downloads them all from Kubernetes and keeps them in memory (or at least it does not have time to discard them before it is killed).

FYI before we made the migration the memory limit for Traefik pods was 512Mi and we never had a problem. We had to increase it to 2Gi to make it work until we understood where the memory consumption was coming from.

For the record Helm v2 made the `nginx-ingress-controller` behave quite similarly (OOMkilled because it reached the limit) because it was downloading all the clusters ConfigMaps including all the ones storing Helm v2 releases. The following patch was merged to avoid downloading Helm related ConfigMaps any longer https://github.com/kubernetes/ingress-nginx/pull/5702. Note that this would not work as it for Traefik and Helm v3 Secrets as it seems that Helm v3 Secret labeling use `owner=helm` instead of `OWNER=TILLER`, i don't know if the filtering is case sensitive. 

I would have expected that traefik does not fetch all the secrets from the Kubernetes api-server, only the ones it needs.

### What did you see instead?

Traefik being killed because it consumes too much memory.

[v2.2.1+sylr.1](https://github.com/sylr/traefik/commits/v2.2.1%2Bsylr.4)

```toml
    [log]
      level = "INFO"
      format = "json"
    #[tracing]
    #  [tracing.instana]
    #    localAgentHost = "instana-agent.instana.svc.cluster.local."
    #    localAgentPort = 42699
    #    logLevel = "info"
    [accessLog]
      format = "json"
      [accessLog.fields]
        defaultMode = "keep"
      [accessLog.fields.names]
        ClientAddr = "drop"
        RequestAddr = "drop"
      [accessLog.fields.headers]
          "User-Agent" = "keep"
    [entryPoints.http]
      address = ":80"
    [entryPoints.http-internal]
      address = ":81"
    [entryPoints.https]
      address = ":443"
    [entryPoints.traefik]
      address = ":8080"
    [entryPoints.traefik-ext]
      address = ":8081"
    [providers.kubernetesIngress]
      ingressClass = "traefik-v2"
    [providers.kubernetesCRD]
    [providers.file]
      directory  = "/config/static"
      watch = true
    [api]
      insecure = true
      dashboard = true
      debug = true
    [metrics.prometheus]
      entryPoint = "traefik"
      buckets = [0.1,0.25,0.5,0.75,1.0,1.5,2.0]
```

```
(paste your output here)
```

sylr:
### What does this PR do?

Filter out kubernetes secret generated by Helm v3 which contains large binary data.

### Motivation

Fix #7384, #7406

### More

- [ ] Added/updated tests

### Additional Notes

Banners_and_Alerts_and_Prometheus_Time_Series_Collection_and_Processing_Server

### DIFF
--- a/pkg/provider/kubernetes/crd/client.go
+++ b/pkg/provider/kubernetes/crd/client.go
@@ -13,6 +13,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	kubeerror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
@@ -64,8 +65,11 @@ type clientWrapper struct {
 	csCrd  *versioned.Clientset
 	csKube *kubernetes.Clientset
 
-	factoriesCrd  map[string]externalversions.SharedInformerFactory
-	factoriesKube map[string]informers.SharedInformerFactory
+	factoriesCrd              map[string]externalversions.SharedInformerFactory
+	factoriesKube             map[string]informers.SharedInformerFactory
+	factoriesKubeIngress      map[string]informers.SharedInformerFactory
+	factoriesKubeSecretTLS    map[string]informers.SharedInformerFactory
+	factoriesKubeSecretOpaque map[string]informers.SharedInformerFactory
 
 	labelSelector labels.Selector
 
@@ -89,10 +93,13 @@ func createClientFromConfig(c *rest.Config) (*clientWrapper, error) {
 
 func newClientImpl(csKube *kubernetes.Clientset, csCrd *versioned.Clientset) *clientWrapper {
 	return &clientWrapper{
-		csCrd:         csCrd,
-		csKube:        csKube,
-		factoriesCrd:  make(map[string]externalversions.SharedInformerFactory),
-		factoriesKube: make(map[string]informers.SharedInformerFactory),
+		csCrd:                     csCrd,
+		csKube:                    csKube,
+		factoriesCrd:              make(map[string]externalversions.SharedInformerFactory),
+		factoriesKube:             make(map[string]informers.SharedInformerFactory),
+		factoriesKubeIngress:      make(map[string]informers.SharedInformerFactory),
+		factoriesKubeSecretTLS:    make(map[string]informers.SharedInformerFactory),
+		factoriesKubeSecretOpaque: make(map[string]informers.SharedInformerFactory),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


rtribotte:
Hello @sylr,

Thanks for your contribution !

As a first lookup, i can say that a fixed labelSelector excluding for all informers the helm resources is not something that we will accept, because it's one use case.

Also, i wonder why you don't use the configurable label selector from the client to filter out the resources ?

rtribotte:
Re @sylr 

I forgot, but can you please rebase your PR to the v2.3 ?

sylr:
> As a first lookup, i can say that a fixed labelSelector excluding for all informers the helm resources is not something that we will accept, because it's one use case.

OK, I can understand that, I try another approach.

> Also, i wonder why you don't use the configurable label selector from the client to filter out the resources ?

Because the labelSelectors in traefik's kubernetes/kubernetesCDR providers configuration are only applied once traefik has received the objects from the api-server, it's too late as it is already in memory.

The next patch iteration will fix that.

> I forgot, but can you please rebase your PR to the v2.3 ?

I believe the problem is severe enough to be fixed at least on the v2.2 line. Once something is accepted in v2.2 i'll merge it into v2.3.

ldez:
We only maintain one minor version, so we will not create a fix release on v2.2.

https://github.com/traefik/traefik#release-cycle

So could you rebase on v2.3, thank you.

jbdoumenjou:
Hi @sylr,

We think that using the labelSelector as tweakListOptions for K8s Ingress and CRD Provider is a good idea, therefore we created a specific PR #7521.
In the same way, we think that it will be simpler to add only one hardcoded labelSelector with "owner!=helm" (from Helm [doc](https://helm.sh/docs/topics/advanced/#configmap-storage-backend)) as tweakListOptions on the `Secret` informer both for the Ingress and CRD providers. With that solution, we only filter Secrets created by Helm Chart. WDYT?

sylr:
I think it would have been more considerate to ask me first before doing a parallel PR which will ultimately make this one obsolete and consequently make the time I spent working on it worthless ...

Anyhow, I'm all for applying label selection to all objects but this is a breaking change so not mergeable on the current 2.3 branch. This patch however is transparent in term of discovery.

jbdoumenjou:
Hi @sylr, thanks for your feedback.

The other PR applies the filter only to the Traefik CRD Objects and Ingresses.
Moreover, the other PR is based on master so you don't have to rebase your branch on it.

Concerning this fix, as we already mentioned, having one specific filter as tweakListOptions with a labelselector "owner=helm" only on the secrets informer will avoid excessive memory usage.

sylr:
> The other PR applies the filter only to the Traefik CRD Objects and Ingresses.

That is a breaking change that can not be applied on current v2.3 branch.

> Moreover, the other PR is based on master so you don't have to rebase your branch on it.

I've already been asked to rebase from 2.2 to 2.3 so it is not really a problem.

> Concerning this fix, as we already mentioned, having one specific filter as tweakListOptions with a labelselector "owner=helm" only on the secrets informer will avoid excessive memory usage.

And that's what I did the with the first iterations of this patch and then I thought about it and decided to filter secrets using their type rather than labels because Traefik has no business retrieving secrets with types it has no use for. This will avoid more troubles if the labels of helm change again or if another component start using kubernetes Secrets has data source.

jbdoumenjou:
We think that the fix on v2.3 with a labelselector is enough as a first step, if helm changes the label, or if another issue comes, we will update the code accordingly.

Can you update the PR in that way?

Thanks

sylr:
> Can you update the PR in that way?

No, I'm good, I stand by my fix.

PR #7416, opened 2020-10-13
Add Kubernetes Gateway Provider

### LINKED ISSUE #7258: Implement Gateway API v1Alpha1

Feature

There will be an upcoming release of the Gateway API in v1Alpha1. We should support that new specification for fetching dynamic configuration. For the beginning, we should support a limited subset of `GateWay`, `GateWayClass` and `HTTPRoute`.

```yaml
kind: GatewayClass
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: traefik-class #Just a name
spec:
  controller: "traefik.io/gateway-controller"
----
kind: Gateway
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: gateway
  namespace: default
spec:
  class: traefik-class
  listeners:
    - port: 80
      protocol: HTTPS
      tls:
      options: {}
      certificateRefs:
        - resource: secrets
        group: core
        name: app
      routes:
        namespaceSelector: {}
        routeSelector:
          matchLabels:
            app: myapp
```

```yaml
kind: HTTPRoute
apiVersion: networking.x-k8s.io/v1alpha1
metadata:
  name: http-app
  namespace: default
  labels:
    controller: traefik
spec:
  hosts:
  - hostname: "foo.com"
    rules:
    - match:
        path: /
        headers: 'my-awesome-header'
      action:
        forwardTo:
          targetRef:
            name: my-service
            group: core
            resource: services
          targetPort: 8080 #optional, if the service endpoint only has 1 port
```

**Criteria**:
* HTTPRoutes should translate into discovered routes, given the routes visible
* Listeners on a Gateway must match Traefiks Entrypoints
* The matching string in the controller spec of a `GatewayClass` should be `traefik.io/gateway-controller`

rtribotte:
### What does this PR do?

This PR brings a new provider for the Kubernetes Gateway API.
The current implementation provides a first step to support the new Kubernetes Gateway API, and it's currently focused on supporting HTTPRoute only.

At the time of the opening of this PR:
- The code uses the [Service APIs](https://github.com/kubernetes-sigs/service-apis/) SIGs project [v0.1.0](https://github.com/kubernetes-sigs/service-apis/releases/tag/v0.1.0) tag.
- Only basic `HTTP` and `HTTPS` scenarios are supported
  - with `Host`, `Header` and `Path` rules
  - that refers to a Kubernetes `Service`
  - with TLS certificate defined in a Kubernetes `Secret`

### Motivation

fixes #7258
Propose a first implementation of Kubernetes Gateway API specification.

### More

 - [x] Added/updated tests

### Additional Notes

This PR will remain a work in progress until the first stable version of Kubernetes Gateway API.
Therefore, more features will be added in the meantime.


kevinpollet (changes requested):

docs/content/reference/dynamic-configuration/kubernetes-gateway.md
```diff
@@ -0,0 +1,24 @@
+# Kubernetes Configuration Reference
+
+Dynamic configuration with Kubernetes Gateway provider
```
kevinpollet, line 3:
```suggestion
Dynamic configuration with Kubernetes Gateway provider.
```

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -0,0 +1,867 @@
...
+				})
+
+				logger.Errorf(msg)
+				listenerInError = true
+				continue
+			}
+
+			for _, httpRoute := range httpRoutes {
+				// Should never happen
+				if httpRoute == nil {
```
kevinpollet, line 386: Are you sure that this check is really necessary? 

I think the Kubernetes provider will not return an array with `nil` inside.

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -0,0 +1,867 @@
...
+			}
+			conf.TLS.Certificates = append(conf.TLS.Certificates, getTLSConfig(tlsConfigs)...)
+		}
+	}
+
+	return conf
+}
+
+func hostRule(httpRouteSpec v1alpha1.HTTPRouteSpec) string {
+	if len(httpRouteSpec.Hostnames) > 0 {
```
kevinpollet, line 538: I think this check is not necessary because the `hostRule` will be empty as we will not iterate on the `httpRouteSpec.Hostnames`.

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -0,0 +1,867 @@
...
+}
+
+func extractRule(routeRule v1alpha1.HTTPRouteRule, hostRule string) string {
+	var rule string
+	var matchesRules []string
+
+	for _, match := range routeRule.Matches {
+		var matchRules []string
+		// TODO handle other path types
+		if match.Path.Type == v1alpha1.PathMatchExact {
```
kevinpollet, line 562: Maybe it's just a matter of taste but I would have used a `switch` here, and add a log if the user has used an unsupported path `Type` matching, wdyt?

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -0,0 +1,867 @@
...
+				log.WithoutContext().Errorf("A multiple ports Kubernetes Service cannot be used if unspecified forwardTo.Port")
+				continue
+			}
+
+			var portName string
+			var portSpec corev1.ServicePort
+			var match bool
+
+			for _, p := range service.Spec.Ports {
+				if forwardTo.Port == 0 || p.TargetPort.IntVal == int32(forwardTo.Port) {
```
kevinpollet, line 768: I think we should also check that the `TargetPort` is of `int` type, because the `TargetPort` of a service can reference a container port by name.

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -0,0 +1,867 @@
...
+						// TODO update the RouteStatus condition / deduplicate conditions on listener
+						continue
+					}
+
+					if routeRule.ForwardTo != nil {
+						wrrService, subServices, err := loadServices(client, gateway.Namespace, routeRule.ForwardTo)
+						if err != nil {
+							msg := fmt.Sprintf("Cannot load service from HTTPRoute %s/%s : %v", gateway.Namespace, httpRoute.Name, err)
+							// update "ResolvedRefs" status true with "DroppedRoutes" reason
+							listenerStatuses[i].Conditions = append(listenerStatuses[i].Conditions, metav1.Condition{
```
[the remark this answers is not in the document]
jbdoumenjou: They omit to update this comment accordingly to the update of the Gateway Status.

pkg/provider/kubernetes/gateway/client.go
```diff
@@ -0,0 +1,425 @@
...
+	}
+
+	return httpRoutes, nil
+}
+
+func (c *clientWrapper) GetGateways() ([]*v1alpha1.Gateway, error) {
+	var result []*v1alpha1.Gateway
+
+	for ns, factory := range c.factoriesGateway {
+		// TODO: add real label selector
```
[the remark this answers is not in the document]
jbdoumenjou: As it is the first step, we will not apply a specific filter.

pkg/provider/kubernetes/gateway/client.go
```diff
@@ -0,0 +1,425 @@
...
+			continue
+		}
+	}
+
+	if index != -1 {
+		gc.Status.Conditions[index] = condition
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		_, err := c.csGateway.NetworkingV1alpha1().GatewayClasses().UpdateStatus(ctx, gc, metav1.UpdateOptions{})
```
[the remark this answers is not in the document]
rtribotte: We don't reuse the GatewayClass after updating it, so for now, we'll keep it as is

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -0,0 +1,867 @@
...
+	gatewayClasses, err := client.GetGatewayClasses()
+	if err != nil {
+		logger.Errorf("Cannot found GatewayClasses : %v", err)
+		return conf
+	}
+
+	for _, gatewayClass := range gatewayClasses {
+		if gatewayClass.Spec.Controller == "traefik.io/gateway-controller" {
+			allowedGateway[gatewayClass.Name] = struct{}{}
+
```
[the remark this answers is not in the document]
rtribotte: Should be ok by now

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -0,0 +1,866 @@
...
+				Type:               string(v1alpha1.GatewayConditionReady),
+				Status:             metav1.ConditionFalse,
+				LastTransitionTime: metav1.Now(),
+				Reason:             string(v1alpha1.GatewayReasonListenersNotValid),
+				Message:            "All Listeners must be valid",
+			})
+
+			err := client.UpdateGatewayStatus(gateway, gatewayStatus)
+			if err != nil {
+				logger.Debugf("An error occurred while updating gateway status: %v", err)
```
kevinpollet, line 475: I think it's better to log an error here, wdyt?

```suggestion
				logger.Errorf("An error occurred while updating gateway status: %v", err)
```

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -0,0 +1,866 @@
...
+		// update "Ready" status with "ListenersValid" reason
+		gatewayStatus.Conditions = append(gatewayStatus.Conditions, metav1.Condition{
+			Type:               string(v1alpha1.GatewayConditionReady),
+			Status:             metav1.ConditionTrue,
+			LastTransitionTime: metav1.Now(),
+		})
+
+		err := client.UpdateGatewayStatus(gateway, gatewayStatus)
+		if err != nil {
+			logger.Debugf("An error occurred while updating gateway status: %v", err)
```
kevinpollet, line 497: Same here.

```suggestion
			logger.Errorf("An error occurred while updating gateway status: %v", err)
```

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -0,0 +1,866 @@
...
+
+				if port == 0 {
+					return nil, nil, errors.New("cannot define a port")
+				}
+
+				protocol := getProtocol(portSpec, portName)
+
+				for _, addr := range subset.Addresses {
+					svc.LoadBalancer.Servers = append(svc.LoadBalancer.Servers, dynamic.Server{
+						URL: fmt.Sprintf("%s://%s:%d", protocol, addr.IP, port),
```
jspdown, line 809: Could you concat the IP and Port using `net.JoinHostPort`. Otherwise the URL will be invalid for IPv6 as it's expected to be `http://[2001::]:80`.

pkg/provider/kubernetes/gateway/client.go
```diff
@@ -0,0 +1,444 @@
...
+}
+
+func (c *clientWrapper) UpdateGatewayStatus(gateway *v1alpha1.Gateway, gatewayStatus v1alpha1.GatewayStatus) error {
+	if !c.isWatchedNamespace(gateway.Namespace) {
+		return fmt.Errorf("cannot update Gateway status %s/%s: namespace is not within watched namespaces", gateway.Namespace, gateway.Name)
+	}
+
+	g := gateway.DeepCopy()
+
+	if statusEquals(g.Status, gatewayStatus) {
```
kevinpollet, line 302: I'm wondering here if it's appropriate to return an `error` if the statuses are equal?
Maybe we just have to return?

kevinpollet (commented):
Just minor comments 😄

kevinpollet (approved):
LGTM 👍

ldez (approved):
LGTM

jbdoumenjou (approved):
LGTM

PR #7433, opened 2020-10-19
Forward Proxy-Authorization header to authentication server

### LINKED ISSUE #7374: HTTP Header Proxy-Authenticate can be conditionally forwarded.

Feature

When using the forwardAuth middleware and a request using the `Proxy-Authenticate` header, I expect the header to be forwarded to the authentication server.

### Justification

Let me start by saying I am aware that Proxy-Authenticate is a hop by hop header and should normally be stripped according to [RFC 2616](https://tools.ietf.org/html/rfc2616#page-92). However, [RFC 7235](https://tools.ietf.org/html/rfc7235#section-4.4) extends on authentication headers. In section 4.4 it states:
```
   Unlike Authorization, the Proxy-Authorization header field applies
   only to the next inbound proxy that demanded authentication using the
   Proxy-Authenticate field.  When multiple proxies are used in a chain,
   the Proxy-Authorization header field is consumed by the first inbound
   proxy that was expecting to receive credentials.  A proxy MAY relay
   the credentials from the client request to the next proxy if that is
   the mechanism by which the proxies cooperatively authenticate a given
   request.
```

Since the forward auth middleware is delegating authentication for the proxy, the `Proxy-Authenticate` header can be sent along freeing up the `Authorization` header for applications behind the proxy that also want authentication. This would be Traefik's way of consuming the header. By optionally allowing the header to be passed, it would also allow for proxy chaining with proxy authentication and would also not not hog the application `Authorization` header.

Scapal:
### What does this PR do?

This PR fixes issue #7374 and related to the ForwardAuth middleware header filtering behaviour.

It allows the `Proxy-Authorization` header to be forwarded to the Authentication service.
It also fixes a potential issue with the `authRequestHeaders` option, skipping header filtering if an explicit header list is given.

### Motivation

I am using Traefik 2 with Authelia using ForwardAuth.
Authelia expects the authentication header `Proxy-Authorization` for proxy authentication but it is currently removed when the request if sent to the authentication service as it is considered a "hop" header. 

Without modification we are unable to use this setup to protect APIs.

fixes #7374 

### More

- [x] Added/updated tests


Scapal:
@kevinpollet @jbdoumenjou Anything blocking or is it just the lack of time ?

jbdoumenjou:
Hello @Scapal,

sorry for a long time no answer.

As you can see in our [contribution guide](https://github.com/traefik/contributors-guide/blob/master/review_guidelines.md),
`needs-design-review` means that we have to take a closer look at the scope of impact for that PR to see, how it would interact with the other parts of Traefik.
For that, it's just a lack of time by now.

jbdoumenjou (approved):
LGTM :+1:

kevinpollet (approved):
Thanks 👍

ldez (approved):
LGTM

PR #7446, opened 2020-10-21
Add missing root certificates to ca-certificates.txt

miff2000:
Add missing root certificates to ca-certificates.txt

Add missing root certs for The USERTRUST Network & for AAA Certificate Services by Comodo CA Limited

These were missing in the latest 2.3 build, so my Sectigo (was Comodo) issued certificates were seen as untrusted.
Comodo had one of their intermediate CAs expire in May 2020, so this adds the replacements and resolves that issue.

Ref: <https://support.sectigo.com/articles/Knowledge/Sectigo-AddTrust-External-CA-Root-Expiring-May-30-2020>

### What does this PR do?

Adds missing CA certificates which are needed to enable trust to Sectigo-published SSL certificates

### Motivation

The root certificates file which is included in the containers doesn't include the latest CAs from Comodo/Sectigo

### More

- [ ] Added/updated tests


miff2000:
I'm not sure what's failed here, but I don't think it will be related to my changes?

rtribotte:
Hello @miff2000,

Thanks for your contribution !

The file you have modified is only used for the run-dev goal of the makefile, or at least it's not used to build our docker images.
For official docker images, we rely on the package ca-certificates for certificate authorities.

I recommend to investigate this way, since we are always using a recent version of alpine (at the moment v3.11).
From a quick look at the package content, you can see that there is only one comodo certificate authority in v3.11 & v3.12:
https://pkgs.alpinelinux.org/contents?file=*Comodo*&path=&name=ca-certificates&branch=v3.11&repo=main&arch=x86
https://pkgs.alpinelinux.org/contents?file=*Comodo*&path=&name=ca-certificates&branch=v3.12&repo=main&arch=x86

I will close this PR because i think we cannot address the issue directly.

rtribotte:
Hello again,

I missed to point you out the right project regarding our official images : https://github.com/traefik/traefik-library-image

PR #7468, opened 2020-10-27
Adding MA support

ghatwala:
- Adding power arch to travis
- ensuring go-bindata generates via go get

### What does this PR do?

Attempts to add support for power arch

### Motivation
Making this work for non-intel arch

### More

- [ ] Added/updated tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
-sudo: required
-dist: trusty
+arch:
+  - amd64
+  - ppc64le
 
 git:
   depth: false
--- a/exp.Dockerfile
+++ b/exp.Dockerfile
@@ -19,9 +19,7 @@ RUN apk --update upgrade \
     && update-ca-certificates \
     && rm -rf /var/cache/apk/*
 
-RUN mkdir -p /usr/local/bin \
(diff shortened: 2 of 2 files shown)


ldez:
Hello,

Travis is not used to running tests or to build binaries.

We only used Travis to build the documentation.

Thanks anyway.

Duplicate of https://github.com/traefik/traefik/pull/7212#issuecomment-680060568

ghatwala:
@ldez - thanks for the info about travis,yml update , but what about `exp.Dockerfile` file changes in this PR , this change was needed to create traefik docker image on non-intel platforms as this binary - https://github.com/containous/go-bindata/releases is not supported on non-intel.  wdyt ?

ldez:
The docker images are not created from these Dockerfile.
This Dockerfile is used only to create some experimental images.

ghatwala:
Guess this is the official `traefik` image's dockerfile repo - https://github.com/containous/traefik-library-image - ? and there could be an option of adding power support here - https://github.com/containous/traefik-library-image/blob/master/alpine/Dockerfile ? is that correct , just confirming my understanding - @ldez  ?

ldez:
The Dockerfiles used to create the official image are is this repo https://github.com/containous/traefik-library-image
There are several files: `tmplv1.Dockerfile`, `tmplv2.Dockerfile`

PR #7475, opened 2020-10-27
Ability To Set Cookie Domain On Sticky Sessions

Distortedlogic:
### What does this PR do?

Simply added domain to cookie configurations for sticky sessions

Now a Dev can apply the label:
`- traefik.http.services.backend.loadbalancer.sticky.cookie.domain=my.desired.cookie.domain.com`
or comparable config in toml/yml/json, to specify what cookie domain is set for the sticky session cookie

### Motivation

traefik sits between my frontend and backend. So the cookie domain set by traefik does not match the domain of the frontend

ie fontend domain - "memehub.lol". and backend domain - "backend.memehub.lol"
Traefik sets the cookie domain to "backend.memehub.lol" by default itself since that is the domain it sits on.

When the frontend is refreshed, if the cookie domain is not matching the current domain/subdomain then the cookie is lost on the refresh.

After traefik sets the initial sticky cookie, if I manually change the domain of the cookie to the desired cookie domain ".memehub.lol" then the cookie persists through refreshes without issue, BUT, if the cookie domain is left as traefik self sets it "backend.memehub.lol" then the cookie is lost on refresh.

When the cookie is lost, then when replicas for the backend service is more than one, the requests dont get routed to the same server each time as desired, but get defaulted to round robin

Before this PR, Traefik had no way set to the cookie domain. 
The package used by traefik for round robin accepts a cookie domain.
Just needed to add a couple lines to add domain to the traefik config and have traefik pass it on through.

### Additional Notes

I am a fairly noob coder, less than one year under my belt.
So if I am missing something about this PR process, go lax on me.
I built my pr fork with the exp.Dockerfile and tested it was functional on my site.

### DIFF
--- a/pkg/config/dynamic/http_config.go
+++ b/pkg/config/dynamic/http_config.go
@@ -118,6 +118,7 @@ type Cookie struct {
 	Secure   bool   `json:"secure,omitempty" toml:"secure,omitempty" yaml:"secure,omitempty"`
 	HTTPOnly bool   `json:"httpOnly,omitempty" toml:"httpOnly,omitempty" yaml:"httpOnly,omitempty"`
 	SameSite string `json:"sameSite,omitempty" toml:"sameSite,omitempty" yaml:"sameSite,omitempty"`
+	Domain string `json:"domain,omitempty" toml:"domain,omitempty" yaml:"domain,omitempty"`
 }
 
 // +k8s:deepcopy-gen=true
--- a/pkg/server/service/loadbalancer/wrr/wrr.go
+++ b/pkg/server/service/loadbalancer/wrr/wrr.go
@@ -126,7 +126,7 @@ func (b *Balancer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 
 	if b.stickyCookie != nil {
-		cookie := &http.Cookie{Name: b.stickyCookie.name, Value: server.name, Path: "/", HttpOnly: b.stickyCookie.httpOnly, Secure: b.stickyCookie.secure}
+		cookie := &http.Cookie{Name: b.stickyCookie.name, Value: server.name, Path: "/", HttpOnly: b.stickyCookie.httpOnly, Secure: b.stickyCookie.secure, Domain: b.stickyCookie.domain}
 		http.SetCookie(w, cookie)
 	}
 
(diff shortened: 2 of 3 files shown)


Distortedlogic:
idk why make docs failed when I didnt change the docs?

kevinpollet:
Hello @Distortedlogic,

You've done nothing wrong we had an issue with our CI and we have restarted your builds.

kevinpollet:
Hello @Distortedlogic,

Thanks for your pull request and your interest in Traefik!

We tried to reproduce the issue you are facing but without any success. When the frontend page is refreshed the sticky cookie is not removed by the browser and sent on refresh to the backend through the AJAX request.

Allowing the configuration of the cookie `Domain` could lead to misconfigurations, e.g. if the Traefik service with the sticky cookie is used by different routers. We are also thinking that sharing the sticky cookie between subdomains does not really make sense.

According to that, we will close this PR but https://github.com/traefik/traefik/issues/7463 will remain open if you want to provide us with a reproducible case.

Distortedlogic:
Did you have traefik on a different domain than the frontend? The issue occurred for me both on chrome and brave and I provided screen shots on the issue I opened for this prior to the pr. Regardless, why not just state the caveat for any worries in docs like anything else n give the option to use or not to the devs instead of choosing for them?

mdklapwijk:
We are using multiple identical instances of traefik:
- running the same synced config
- hosting multiple routers, 1 for each subdomain,
- routers grouped; cookie name for each router in a group set the same e.g. `routerid.md5(groupname)`,
- routers ungrouped; cookie name set to e.g. `routerid.md5(fqdn)`,
- subdomains referencing/linking/redirecting to each other.
- running over multiple uplinks using multiple a-records+rfc2136

We'd like to share the cookies between subdomains, to be able to make sure a user stays with the same real server as much as possible, no matter which subdomain they are using. 

@kevinpollet, I do not see any real issues concerning misconfiguration as one should set the name to something unique for each router/group. So it would be nice if the setting of 'domain' on the cookie was made possible.

kevinpollet:
Hello @mdklapwijk and thanks for your interest in Traefik,

Please, could you open a feature request describing the use case?

Distortedlogic:
Years later, I still cant believe such a small straightforward useful feature with the code already PR'ed with only 3 lines of code that literally only passes through one additional setting that affects nothing else, and another dev laying out their need requires them to create a feature request? What a clown show 🤡 . Glad I dropped Traefik from my tech stack.

Issue #7421, opened 2020-10-15
Add logging field for TLS client version


timwsuqld:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?
https://doc.traefik.io/traefik/observability/access-logs/#limiting-the-fieldsincluding-headers

A field for logging the client TLS versions. This is provided in Nginx and AWS Load balancers, and is useful to get a metric of how many outdated clients you have before you disable an older TLS version.

PR #7478, opened 2020-10-28
Add TLS version and cipher to the accessLog

na4ma4:
### What does this PR do?

Adds the TLS Version and Cipher to the logs when using JSON.

### Motivation

Knowing the TLS Version and Cipher of requests helps indicate when old ciphers and versions can be deprecated.

### More

- [x] Added/updated tests

### Additional Notes

relates to issue #7421


sylr:
I believe this is related to https://github.com/traefik/traefik/pull/6255 so maybe worth moving and making `getRequestTLSVersion` and `getRequestTLSCipher` public so that metrics and logs get coherent values.

pkg/middlewares/accesslog/logger_test.go
```diff
@@ -355,6 +355,8 @@ func TestLoggerJSON(t *testing.T) {
 				Duration:                  assertFloat64NotZero(),
 				Overhead:                  assertFloat64NotZero(),
 				RetryAttempts:             assertFloat64(float64(testRetryAttempts)),
+				"TLSVersion":              assertString("TLS13"),
+				"TLSCipher":               assertString("TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"),
```
jspdown, line 359: Could you directly use `TLSVersion` and `TLSCipher` constants here?

na4ma4:
Would it be better to move the `getRequestTLSVersion` and `getRequestTLSCipher` functions into [pkg/tls/tls.go](https://github.com/traefik/traefik/blob/master/pkg/tls/tls.go) as exported functions, rather than copy/pasting the code ?

Since it uses the data from there anyway.

kevinpollet (approved):
Thanks 👍

sylr (approved):

dtomcej (approved):
LGTM
:shipit:

jspdown (approved):
LGTM :+1:

PR #7480, opened 2020-10-28
fix: translate configured server port into correct mapped host port

### LINKED ISSUE #7470: ECS load balancer port should refer to container port, not to host port

Bug

### What did you do?

I'm using traefik 2.3.2 on ECS. My services are using dynamic allocated ports with bridge networking mode.
Two of my services expose two ports that need to be routed with different rules.  
I've added two `traefik.http.services.<service_name>.LoadBalancer.server.port=<container_exposed_port>` labels on the container exposing the two ports, expecting that the port would be converted to the dynamic allocated port on the EC2 host, but it didn't work.

Example:

```
EC2 private address: 172.16.0.133

container port: 80 -> host port: 32134
container port: 81 -> host port: 32135
```

When specifing the port on service load balancer, I expected to see the service address to be `http://<private_ec2_address>:<host_port>` where `host_port` is the port mapped to the container port written on the container label.

On the previous example, I expected to see two services with addresses:

- `http://172.16.0.133:32134`
- `http://172.16.0.133:32135`

### What did you see instead?

The port is used on service address without checking the port mapping.  
This results in the service to be created with address `http://<private_ec2_address>:<container_port>` which is terribly wrong.

On the previous example, I see:

- `http://172.16.0.133:80`
- `http://172.16.0.133:81`

```
Version:      2.3.2
Codename:     picodon
Go version:   go1.15.3
Built:        2020-10-19T18:36:22Z
OS/Arch:      linux/amd64
```

alekitto:
### What does this PR do?

Fixes #7470: ecs `getPort` function now checks the exposed/mapped ports for services to correctly handle dynamic allocated ports.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/ecs/config.go
+++ b/pkg/provider/ecs/config.go
@@ -307,6 +307,13 @@ func (p Provider) getIPAddress(instance ecsInstance) string {
 
 func getPort(instance ecsInstance, serverPort string) string {
 	if len(serverPort) > 0 {
+		for _, port := range instance.machine.ports {
+			containerPort := strconv.FormatInt(port.containerPort, 10)
+			if serverPort == containerPort {
+				return strconv.FormatInt(port.hostPort, 10)
+			}
+		}
+
 		return serverPort
 	}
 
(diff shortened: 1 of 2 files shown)


kevinpollet (approved):
Thanks 👍

rtribotte (approved):
LGTM 👍

ldez (approved):
Thank you :+1:

PR #7487, opened 2020-10-28
Produce ConfigReloadsFailureCounter and LastConfigReloadFailureGauge metrics in case of config load error

pbarsotti-glovo:
### What does this PR do?

The metrics `ConfigReloadsFailureCounter` and `LastConfigReloadFailureGauge` are currently not being published in case of config load errors. This PR includes this missing behaviour. More information in this [issue](https://github.com/traefik/traefik/issues/7471).

### Motivation

A suggestion by @kevinpollet to collaborate fixing this issue.

### More

- [X] Added/updated tests

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -290,6 +290,11 @@ func setupServer(staticConfiguration *static.Configuration) (*server.Server, err
 		}
 	})
 
+	watcher.AddConfigLoadErrorListener(func() {
+		metricsRegistry.ConfigReloadsFailureCounter().Add(1)
+		metricsRegistry.LastConfigReloadFailureGauge().Set(float64(time.Now().Unix()))
+	})
+
 	return server.NewServer(routinesPool, serverEntryPointsTCP, serverEntryPointsUDP, watcher, chainBuilder, accessLog), nil
 }
 
--- a/pkg/config/dynamic/config.go
+++ b/pkg/config/dynamic/config.go
@@ -8,8 +8,9 @@ import (
 
 // Message holds configuration information exchanged between parts of traefik.
 type Message struct {
-	ProviderName  string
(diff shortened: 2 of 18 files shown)


pbarsotti-glovo:
Hey @kevinpollet, @ldez, @jbdoumenjou any news regarding the status revision of this PR?
I don't know the formalities of the process to pass the validation of PRs, but if I can help somehow, please let me know.
It would be very useful for us to have this metric under the radar to spot possible configuration issues, since many people has access to edit services and routes.

jbdoumenjou:
Hello @Scapal,

sorry for a long time no answer.
As you can see in our [contribution guide](https://github.com/traefik/contributors-guide/blob/master/review_guidelines.md),
`needs-design-review` means that we have to take a closer look at the scope of impact for that PR to see, how it would interact with the other parts of Traefik.

For that, it's just a lack of time by now.

jbdoumenjou:
Hello @pbarsotti-glovo,

Thanks for your contribution!

I wonder if we want to update those metrics during the initialization of the provider knowing that the provider will not be launch at all.
So, in other words, do you think we have to consider the errors preventing a provider to start as loading configuration errors?

rtribotte:
Hello @pbarsotti-glovo,

Sorry for the late feedback,

We think that reusing the configuration channel would lead to drop "error" configuration messages because of the throttling mechanism, and this would make the metric not reliable.
Modifying the provider API would be a better solution but it raises other concerns (especially regarding [provider plugins](https://github.com/traefik/pluginproviderdemo)) and we would prefer to address them later in a major version.

For that reason, and unfortunately, we are going to decline this PR.

As a side note, we already removed the documentation for these metrics as they were never implemented: https://github.com/traefik/traefik/pull/8752/files.

Issue #7088, opened 2020-07-22
Support for private plugins


negasus:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Support for private plugins. Connecting plugins from private repositories.

Maybe: this plugin must be placed locally, and we should to define path to plugin in the configuration file. 

That plugin should not be traking by Pilot

emilevauge:
Hi @negasus 👋 
Plugins are super fresh in Traefik, and for now, we want to focus on making those stable. 
Even though there are specific use cases where you need private plugins, right now, we do want to focus on the most requested feature: open source plugins.
But we would love to hear some use cases where private plugins would be required.
Thanks :)

negasus:
I'm not suggesting doing this right now)

For example:
In my case, we need to authorize users by the auth header. And I want to make plugin for obtain some data from Database, cache it, and use this data fot check users. This case very specific for out application. And may contains some business logic.

koalado:
Would love to see support for plugins in private repositories as well.
Following...

lexfrei:
> But we would love to hear some use cases where private plugins would be required.

@emilevauge, in some cases, I have to use traefik in intranet, so there is no option to use pilot or else, but I can place plagins in private repos in local git or place them directly to the docker-image.

dio-trinny:
I would also love this. there's some logic that we need to run specific to us, and I'd rather not expose the logic or create a non-reusable plugin

xavivars:
The other part of the current plugin implementation is that it requires the routing to be connected to Traefik pilot. 

I don't have any problem open-sourcing the plugins I need, but I'd love to be able to create plugins that, for example, enable more middlewares, but being able to do that without having to connect to an arbitrary server at runtime (or startup time) to download arbitrary code (even if signed by Traefik). Instead, I'd like to build "my own" Traefik version, with any plugins I need built into the image.

Other reverse proxies like Skipper allow any developer to do that easily, and I would hope to be able to do the same with Traefik.

xavivars:
@SantoDE : you mentioned in #7515 that private plugins goes against the spirit of the "open-source" catalog

(quoting)

> The challenge I see with private plugins is that it directly contradicts many of the positive outcomes we're trying to achieve with the plugin feature.

and also asked to continue the conversation here. 

Could we continue the conversation here? 

[...]

PR #7515, opened 2020-11-05
add private plugins support

fedorovvl:
### What does this PR do?

Adds support for private plugin repo (#7088)
Two new flags
--pilot.private bool
--pilot.repo string

### Motivation

In my case i have private network without internet and i totally can't use plugins. And all i need for using plugins is self hosted repository

### Additional Notes

i remove check phase if private repo enabled.. its not nessesary i think
private repo tree looks like (with --pilot.repo=http://example.com/traeffik_plugins/)
http://example.com/traeffik_plugins/my-cool-plugin/1.1.0
1.1.0 is zip archive without extension

### DIFF
--- a/cmd/traefik/plugins.go
+++ b/cmd/traefik/plugins.go
@@ -8,26 +8,33 @@ import (
 const outputDir = "./plugins-storage/"
 
 func initPlugins(staticCfg *static.Configuration) (*plugins.Client, map[string]plugins.Descriptor, *plugins.DevPlugin, error) {
-	if !isPilotEnabled(staticCfg) || !hasPlugins(staticCfg) {
-		return nil, map[string]plugins.Descriptor{}, nil, nil
+
+	if (!staticCfg.Pilot.Private && !(isPilotEnabled(staticCfg) || hasPlugins(staticCfg))) {
+	    return nil, map[string]plugins.Descriptor{}, nil, nil
 	}
+	
 
 	opts := plugins.ClientOptions{
 		Output: outputDir,
 		Token:  staticCfg.Pilot.Token,
+		RepoUrl:  staticCfg.Pilot.RepoUrl,
+		Private: staticCfg.Pilot.Private,
 	}
 
 	client, err := plugins.NewClient(opts)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


SantoDE:
Greetings @fedorovvl, thank you for taking the time to prepare this PR and for your contribution to Traefik. Before I share my opinion on the issue, full disclosure, I work at Traefik Labs and have discussed your PR at great length with some other maintainers before answering. I apologize for the delay in answer.

I understand your use case and the need for private plugins. When we introduced the plugin system, our goal was to provide an "open-source" plugin catalog, and there are a number of reasons behind this:

- Most importantly, it allows everyone to benefit from the plugin system. Contributors are encouraged to share their work, and have a centralized repository for distribution. We want to see peoples ideas and needs, and in turn this helps us understand what users are looking to achieve with Traefik.
- It allows the community as a whole to see the most popular plugins and as a result we can better determine what features we should be accepting upstream into the project.
- It's also a easy first step for new contributors to get involved with the project. This encourages collaboration and allows us to nurture and onboard new contributors and maintainers.

The challenge I see with private plugins is that it directly contradicts many of the positive outcomes we're trying to achieve with the plugin feature.

Still, I cannot deny the popularity of this request, and we are investigating ways to implement the support for private plugins, while preserving the benefits the open source community has been afforded since the implementation of this feature. Here are some challenges we want to ensure don't become a detriment to the project:

- Yaegi, the open source engine for the plugins, is still in beta. Feedback from the community through the implementation of public plugins is a significant source of bug reports in that project.
- By implementing the support for private plugins, the ability to support end-users who use this functionality would be made more difficult for maintainers and community members. We receive dozens of questions every day across multiple channels (GitHub, Forums, Reddit, StackOverflow), and introducing this capability may harm the user experience both from an end-user perspective and from a community support perspective.

While we appreciate your contribution, I don't feel it is the approach I'd like to take to ensure the best user experience while at the same time encouraging open collaboration. We are open to your ideas and feedback on how we could overcome some of these challenges. In the meantime,  to keep the conversation going, I propose we switch discussions to the relevant issue (#7088).

PR #7519, opened 2020-11-06
Refactor Exponential Backoff

danieladams456:
### What does this PR do?

This is a followup refactor from #7460.

### Motivation

The previous PR focused on adding the exponential backoff while retaining the same logic flow in the handler method.  This presents an alternative style that might be easier to read.

### More

- [ ] Added/updated tests
  - Internal refactor only
  - Internal refactor only

### Additional Notes
Previously we were just using `NextBackOff()` directly and doing our own orchestration, so the `nexter` interface only had one method.  Now that I am trying to pass that object to another function in the `backoff` package that expects a full backoff object, a function needed to be added to the interface.  I'm not sure if the changed role of the interface necessitates a rename.

In the call to `backoff.RetryNotify()`, right now I am currently throwing away the returned error since the rest of the response writing logic handles it.  This is equivalent to what happens today.  Should I leave it as is or add a special log statement for the terminal failure?


rtribotte (approved):
Thanks 👍

kevinpollet (approved):
Thanks 👍

mpl (approved):

PR #7521, opened 2020-11-06
Apply labelSelector as a TweakListOptions for Kubernetes informers

rtribotte:
### What does this PR do?

This PR improves both the Kubernetes Ingress provider and the Kubernetes CRD provider.
It modifies how the `labelSelector` is applied.
If the labelSelector is configured on providers:

- Before this PR: The resources objects were loaded in the informer's cache then filtered with the `labelSelector`
- With this PR: The resources objects are loaded in the informer's cache only if they match the `labelSelector`

### Motivation

Provide a better usage of the Kubernetes client for the Kubernetes Ingress provider and the Kubernetes CRD provider.

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: Jean-Baptiste Doumenjou 925513+jbdoumenjou@users.noreply.github.com


docs/content/providers/kubernetes-crd.md
```diff
@@ -177,26 +177,32 @@ _Optional,Default: empty (process all resources)_
 
 ```toml tab="File (TOML)"
 [providers.kubernetesCRD]
-  labelselector = "A and not B"
+  labelselector = "foo"
```
dtomcej, line 180: I don't think this is a valid labelselector, and is not indicative of what would be used.

I would recommend:

```
labelSelector = "app=traefik
```

As this is what is used in our example code below

rtribotte: Hey @dtomcej 
While i agree that is not indicative, it is a valid labelSelector which only do label key matching.
Suggestion applied !

kevinpollet (approved):
LGTM 👍

docs/content/providers/kubernetes-crd.md
dtomcej: I didn't know you could do bare label key matching! TIL!

dtomcej (approved):
LGTM
:shipit:

jspdown (approved):
LGTM :+1:

ldez (approved):
LGTM

PR #7542, opened 2020-11-12
Fixed 500 errors in head requests - forwardauth

volker-schukai:
### What does this PR do?

Possible fix for https://github.com/traefik/traefik/issues/6124

A Head-Request has no body and therefore this request runs on an error. 
With the check if the request is a head request you can easily avoid this.

### Motivation

Bugfix:docker push does not work with auth module

### More

- [ ] Added/updated tests

### Additional Notes

https://github.com/thomseddon/traefik-forward-auth/issues/156


kevinpollet:
Hello @volker-schukai,

Thanks for your contribution and interest on Traefik!

I've tried to reproduce the issue you are facing but without any success. In the Traefik code, we are only issuing `GET` requests to the configured `fowardauth` address so nothing could be related to a potential `HEAD` request bug on our side. 

I think the issue is in the [traefik-forward-auth](https://github.com/thomseddon/traefik-forward-auth) project you are using. As described in https://github.com/thomseddon/traefik-forward-auth/issues/156 the log cause (`Cause: context deadline exceeded (Client.Timeout or context cancellation while reading body)`) seems to indicate that Traefik is waiting for a response body which never arrives.

As we think that it's not a Traefik issue, we will close this PR accordingly.

PR #7548, opened 2020-11-13
/bin/bash replaced with /usr/bin/env bash to match other scripts

kunchev:
### What does this PR do?
Change shebang from '/bin/bash' to '/usr/bin/env bash' to match other scripts.

### Motivation
All shell scripts in 'scripts/' using bash shell to have the '/usr/bin/env bash' shebang.

### DIFF
--- a/script/crossbinary-default
+++ b/script/crossbinary-default
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -e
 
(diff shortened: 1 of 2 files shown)


dtomcej (approved):
LGTM
:shipit:

ldez (approved):
Thank you :+1:

kevinpollet (approved):
Thanks 👍

Issue #7466, opened 2020-10-27
k8s Provider: use networking.k8s.io/v1 API


phin1x:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

With Kubernetes 1.19 the Ingress API and Ingress Class API  ist now stable (graduated to networking.k8s.io/v1) and will be removed from the networking.k8s.io/v1beta1 and extensions/v1beta1 API Group with Version 1.22.

Traefik still receive ingress objects via the extensions/v1beta1 informer:
https://github.com/traefik/traefik/blob/015f24a90166163160819c0fb3dcfc50c44aeaba/pkg/provider/kubernetes/ingress/client.go#L150
Traefik should check the cluster version and use the stable networking.k8s.io/v1 ingress and ingress class api if supported (1.19+) or the older extensions/v1beta1 and networking.k8s.io/v1beta1 API if the cluster version is blow 1.19 and convert them to networking.k8s.io/v1 objects.
Since this is already implemented (version check and conversion)  it should not take much effort.

SantoDE:
Hello @phin1x,

thanks for your interest.

Why do you think that change is required? With our current mechanism we are fully backward compatible and can fetch all ingresses, no matter the version.

What is the issue with that approach? Do you see a problematic edge case?

phin1x:
Hi @SantoDE ,

From the kubernetes 1.16 release notes:

Ingress resources will no longer be served from extensions/v1beta1 in v1.20. Migrate use to the networking.k8s.io/v1beta1 API, available since v1.14. Existing persisted data can be retrieved via the networking.k8s.io/v1beta1 API.

So traefik will not work with the next version of kubernetes. This will prevent us from upgrading kubernetes to the next minor release in january 2021.
The extension/v1beta1 api has been deprecated for over 1.5 years and should no longer be used.

phin1x:
The release notes seams to be inconsistent, the removal of the extensions/v1beta1 is scheduled for kubernetes 1.22 which will be released in about one year. With this release, only networking.k8s.io/v1 will be servered, so a change is required anyway.

The timeline can be found here: https://github.com/kubernetes/kubernetes/issues/43214

ldez:
https://github.com/traefik/traefik/blob/015f24a90166163160819c0fb3dcfc50c44aeaba/pkg/provider/kubernetes/ingress/client.go#L211

[...]

PR #7549, opened 2020-11-13
Upgrade Ingress Handling to work with networkingv1/Ingress

SantoDE:
### What does this PR do?

Upgrade the K8s Ingress Provider to use the networkingv1/Ingress type, rather than the old extension type. Also, just start needed informers when needed.

### Motivation

Be ready for the future.

### More

- [x] Added/updated tests

### Additional Notes

We have to upgrade the IngressClass Ressource once as well. We need to decide whether to do that in a seperate PR or just here.

Re #7466

### DIFF
--- a/pkg/provider/kubernetes/ingress/client.go
+++ b/pkg/provider/kubernetes/ingress/client.go
@@ -272,15 +272,6 @@ func (c *clientWrapper) GetIngresses() []*networkingv1.Ingress {
 				results = append(results, n)
 			}
 		}
-
-		if supportsNetworkingV1Ingress(serverVersion) {
-			// networking
-			listNew, err := factory.Networking().V1().Ingresses().Lister().List(labels.Everything())
-			if err != nil {
-				log.Errorf("Failed to list ingresses in namespace %s: %v", ns, err)
-			}
-			results = append(results, listNew...)
-		}
 	}
 	return results
 }
--- a/pkg/provider/kubernetes/ingress/client.go
+++ b/pkg/provider/kubernetes/ingress/client.go
@@ -351,7 +351,7 @@ func (c *clientWrapper) UpdateIngressStatus(src *networkingv1.Ingress, ingStatus
 		return c.updateIngressStatusOld(src, ingStatus)
 	}
 
-	ing, err := c.factoriesIngress[c.lookupNamespace(src.Namespace)].Networking().V1beta1().Ingresses().Lister().Ingresses(src.Namespace).Get(src.Name)
+	ing, err := c.factoriesIngress[c.lookupNamespace(src.Namespace)].Networking().V1().Ingresses().Lister().Ingresses(src.Namespace).Get(src.Name)
 	if err != nil {
 		return fmt.Errorf("failed to get ingress %s/%s: %w", src.Namespace, src.Name, err)
 	}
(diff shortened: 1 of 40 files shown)


pkg/provider/kubernetes/ingress/client.go
```diff
@@ -275,8 +297,53 @@ func extensionsToNetworking(ing marshaler) (*networkingv1beta1.Ingress, error) {
...
+		port := networkingv1.ServiceBackendPort{}
+		if old.Spec.Backend.ServicePort.Type == intstr.Int {
+			port.Number = old.Spec.Backend.ServicePort.IntVal
+		} else {
+			port.Name = old.Spec.Backend.ServicePort.StrVal
+		}
+
+		ing.Spec.DefaultBackend = &networkingv1.IngressBackend{
+			Service: &networkingv1.IngressServiceBackend{
+				Name: old.Spec.Backend.ServiceName,
```
rtribotte, line 311: Service name is also optional, we may not have to create the IngressBackend when it's an empty string

pkg/provider/kubernetes/ingress/client.go
```diff
@@ -275,8 +297,53 @@ func extensionsToNetworking(ing marshaler) (*networkingv1beta1.Ingress, error) {
 	return ni, nil
 }
 
+func addServiceFromV1Beta1(ing *networkingv1.Ingress, old networkingv1beta1.Ingress) {
+	if old.Spec.Backend != nil {
+		port := networkingv1.ServiceBackendPort{}
+		if old.Spec.Backend.ServicePort.Type == intstr.Int {
```
rtribotte, line 303: ServicePort is optional, we should test if it has the default value

pkg/provider/kubernetes/ingress/client.go
```diff
@@ -275,8 +297,53 @@ func extensionsToNetworking(ing marshaler) (*networkingv1beta1.Ingress, error) {
...
+	}
+
+	for rc, rules := range ing.Spec.Rules {
+		if rules.HTTP == nil || len(rules.HTTP.Paths) == 0 {
+			return
+		}
+		for pc, path := range rules.HTTP.Paths {
+			if path.Backend.Service == nil {
+				port := networkingv1.ServiceBackendPort{}
+				if old.Spec.Rules[rc].HTTP.Paths[pc].Backend.ServicePort.Type == intstr.Int {
```
rtribotte, line 328: Are we assured that the length of Rules of the old ingress is indentical as the length of the Rules of the networking v1 one ?

rtribotte: The same applies to the Paths slice

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -221,17 +221,17 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 			log.FromContext(ctx).Errorf("Error configuring TLS: %v", err)
 		}
 
-		if len(ingress.Spec.Rules) == 0 && ingress.Spec.Backend != nil {
+		if len(ingress.Spec.Rules) == 0 && (ingress.Spec.DefaultBackend != &networkingv1.IngressBackend{}) {
```
rtribotte, line 224: We should keep the nil test here, since it's optional, we don't want to proceed further if the default backend is nil

pkg/provider/kubernetes/ingress/client.go
SantoDE: Yes we are :)

pkg/provider/kubernetes/ingress/kubernetes.go
SantoDE: done :)

pkg/provider/kubernetes/ingress/client.go
SantoDE: done :)

pkg/provider/kubernetes/ingress/client.go
SantoDE: done :)

rtribotte (commented):
A few typos in test fixtures and a duplicated code portion.

pkg/provider/kubernetes/ingress/client.go
```diff
@@ -231,42 +237,62 @@ func (c *clientWrapper) WatchAll(namespaces []string, stopCh <-chan struct{}) (<
...
-		if err != nil {
-			log.Errorf("Failed to list ingresses in namespace %s: %v", ns, err)
+		if supportsNetworkingV1Ingress(serverVersion) {
+			// networking
+			listNew, err := factory.Networking().V1().Ingresses().Lister().List(labels.Everything())
+			if err != nil {
+				log.Errorf("Failed to list ingresses in namespace %s: %v", ns, err)
+			}
+			results = append(results, listNew...)
 		}
```
rtribotte, line 284:
```suggestion
```

jbdoumenjou (approved):
LGTM :+1:

rtribotte (approved):
LGTM 👌

ldez (approved):

PR #7553, opened 2020-11-14
Add unicode support in Host rule using Middleware

sidharthv96:
### What does this PR do?

Use `net/idna` to convert punycode in req.host back to unicode to match unicode hosts.

### Motivation

Fixes https://github.com/traefik/traefik/issues/7494

### More

- [x] Added/updated tests

#### Change hostname to unicode for every request. (Implemented in this PR)
Pros:
- Will work with Host and HostRegexp

Cons:
- Punycode to Unicode conversion has to be done per request.

#### Change hostname to punycode when rules are loaded. (Implemented in https://github.com/traefik/traefik/pull/7554)
Pros:
- Only One time conversion is required
 
Cons:
- Will not work with HostRegexp as matching is done inside mux.

### DIFF
--- a/pkg/middlewares/requestdecorator/request_decorator.go
+++ b/pkg/middlewares/requestdecorator/request_decorator.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/containous/alice"
 	"github.com/traefik/traefik/v2/pkg/types"
+	"golang.org/x/net/idna"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hello @sidharthv96,

Thanks for your pull request and your interest in Traefik!

We cannot accept your pull request, because converting the `Host` on every request introduces a breaking change. For example, if someone as a punycoded `Host` in a router rule this will not match anymore. This change could also introduce performance issues as the conversion is done on every request.

Therefore, we will close this pull request accordingly.

PR #7554, opened 2020-11-14
Add Unicode support to Host rule.

sidharthv96:
### What does this PR do?

Use `net/idna` to convert URL to punycode while registering Host rules to match punycode hosts.

### Motivation

Fixes https://github.com/traefik/traefik/issues/7494

### More

- [x] Added/updated tests

#### Change hostname to unicode for every request. (Implemented in https://github.com/traefik/traefik/pull/7553)
Pros:
- Will work with Host and HostRegexp

Cons:
- Punycode to Unicode conversion has to be done per request.

#### Change hostname to punycode when rules are loaded. (Implemented in this PR)
Pros:
- Only One time conversion is required
 
Cons:
- Will not work with HostRegexp as matching is done inside mux.

### DIFF
--- a/pkg/rules/rules.go
+++ b/pkg/rules/rules.go
@@ -9,6 +9,7 @@ import (
 	"github.com/traefik/traefik/v2/pkg/log"
 	"github.com/traefik/traefik/v2/pkg/middlewares/requestdecorator"
 	"github.com/vulcand/predicate"
+	"golang.org/x/net/idna"
 )
 
 var funcs = map[string]func(*mux.Route, ...string) error{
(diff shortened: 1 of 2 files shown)


rtribotte:
Hello @sidharthv96,

Thank you very much for both of your contributions.
We are sorry for the late review, but after re-thinking this feature, we are not willing to introduce it.

Providing an automatic punycoded domain name for a router with `Host` rule would definitely simplify the configuration for the user.
However, because it cannot be done for the `HostRegexp` rule matcher, we believe that this inconsistency would add confusion.

We tend to believe that using the punycoded version of the domain name is expected.

Moreover, the current mechanism of TLSOptions cannot handle non-ASCII domains in rules. Today, a given TLSOption is linked to a router thanks to the domains that can be found in the router's rule.
Supporting non-ASCII domain names in the router's rule matcher `Host` will lead to not be able to find the TLSOption associated with the router and fallback to the default one, because the incoming request server name (punycoded) would not match.

Since those routers would be (with your enhancement) enabled to actually route something (no more 404), and since then the served TLSOption would always be the default one, it will introduce a security issue.

Unfortunately, for now, and for these reasons, I'll decline your pull request.

To improve things until we provide better support for non-ASCII domain names, I've opened a pull request (#7986) that improves the documentation by clarifying the current limitations, and also raises an error when a router is configured with a non-ASCII domain name in its rule.

PR #7555, opened 2020-11-14
ipwhitelist requires router reference

ismailyenigul:
The following lines are not enough to make ipwhitelisting middleware work.

```
  - traefik.http.middlewares.wl3.ipwhitelist.sourceRange=8.8.8.8
  - traefik.http.middlewares.wl3.ipwhitelist.ipStrategy.depth=3
```
You must define the name of the ipwhitelist middleware in router configuration.
`- traefik.http.routers.rt3.middlewares=wl3`
see  
https://github.com/traefik/traefik/blob/master/integration/resources/compose/whitelist.yml

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/middlewares/ipwhitelist.md
+++ b/docs/content/middlewares/ipwhitelist.md
@@ -76,6 +76,10 @@ The `ipStrategy` option defines two parameters that sets how Traefik will determ
 
 The `depth` option tells Traefik to use the `X-Forwarded-For` header and take the IP located at the `depth` position (starting from the right).
 
+To make IP whitelist work, you must register the ipwhitelist middleware in the router definition.
+like ` traefik.http.routers.myroute.middlewares=testIPwhitelist`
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

The page https://doc.traefik.io/traefik/v2.3/middlewares/ipwhitelist/, like the other pages about middleware, describes the way to define the middleware.

The way to apply a middleware definition on a router is documented in the following pages:
- https://doc.traefik.io/traefik/v2.3/middlewares/overview/
- https://doc.traefik.io/traefik/v2.3/routing/providers/docker/#middleware

A middleware can be set to any container and used on any other container. There is no correlation between the container of the definition and the container that uses the middleware.

The change that you propose can let's think that the middleware definition must be set at the same place as the router.

ismailyenigul:
I see!
It seems mentioning this in a howto is better place.
I already created one at https://ismailyenigul.medium.com/limit-clients-to-specific-ips-with-treaefik-v2-ipwhitelist-middlewares-24357cf8f6ba

PR #7593, opened 2020-11-25
Add k8s provider option to create services without endpoints

Lucaber:
### What does this PR do?

Adds a new kubernetes provider option to force the creation of services even if there are no endpoints available.
This results in `503` http responses instead of `404`.

### Motivation

If a Kubernetes Deployment is temporarily not available (scaled to 0), for example due to maintenance, the corresponding service has no endpoints.
The creation of the traefik service failed with `Cannot create service: subset not found` and got removed completely.
Browsing the configured url resulted in an error 404, because the route is no longer configured in traefik.
Furthermore because the route creation is completely canceled, configured middlewares on the ingress (via annotation) like an error middleware are ignored too. 

The route now returns an error code 503 and error middlewares are executed.

The old behaviour can be activated by setting `ignoreEmptyServices=true` 

### More

- [x] Added/updated tests

### Additional Notes

#1077


bradjones1:
Going to be "that guy" and +1 this with a comment, but to add: This makes sense for providers who may spin down targeted pods to zero, but wish to keep the ingress and service intact; the additional log spam of "no matching subset" is a bit misleading if this is your intent.

dtomcej:
I'm wondering if this should instead of behind a flag, be default behavior.

If the state provided by k8s is that there is an expected service with endpoints, and there are no endpoints, we should return a 503.

This could also assist in debugging services, as it would assist in debugging misconfigured services/selectors.

Lucaber:
@dtomcej Yes, i agree but just didn't want to break installations where this is considered a "feature".
We could inverse the flag to something like `ignoreEmptyServices=false` and enable this behaviour by default.


I added a new commit with the new flag and rebased the branch.

icefed:
Thank you, there are too many `Cannot create service: subset not found` logs.

Is that a global option? so I doesn't need to modify Ingress config.

thueske:
Are there any news? What is missing for the pull request to be merged? The change would be very important for us.

thueske:
Are there any news? What is missing for the pull request to be merged? The change would be very important for us.

Maybe @rtribotte? :)

jbdoumenjou:
Hello,
 
Thanks for your contribution!

thueske:
> Hello,
> 
> 
> This PR is labeled `needs-design-review`,
> as you can see in our [contribution guide](https://github.com/traefik/contributors-guide/blob/master/review_guidelines.md), this means that we have to take a closer look at the scope of impact for that PR to see, how it would interact with the other parts of Traefik.
> 
> We will come back to you **soon** when we did the first design review iteration.

Is there an estimation how "soon" we can get a feedback? The pull request is quite small and clear and the contribution is meanwhile 5 months old... A merge would be really important for us.

jbdoumenjou:
Hello,

we started the first review of this PR. :wink: 
If everything goes well, it will be shipped with the v2.5 in the coming weeks.

jbdoumenjou:
We worked on it. 
We reversed the flag because, as you said, it could be breaking for some users.
We fixed some tests and we updated the WebUI to be compliant with a service without any servers.
WDYT ?

dtomcej (approved):
LGTM
:shipit:

jbdoumenjou (approved):
LGTM

rtribotte (approved):
LGTM

Lucaber:
LGTM

holmesb:
Any idea when this will be released?  Wasn't in yesterday's v2.4.9 release.

PR #7599, opened 2020-11-28
acme: add external account binding support.

### LINKED ISSUE #7118: Add EAB support for ACME client
## Background

Many universities leverage InCommon for their TLS certs, who has support for the ACME protocol. However, they require usage of the EAB (ExternalAccountBinding) mechanism. As an example using certbot:

```
certbot -d <domain.com> --server <ACME_Server_URL> --eab-kid <EAB_KEY_ID> --eab-hmac-key <EAB_HMAC_KEY>  --manual --preferred-challenges dns certonly
```

I would like to contribute the work for Traefik to support this use case, but wanted to pass it by the team first to make sure 1) it's something you're open to and 2) that I'm making a PR that is likely to actually be merged (my decisions/design is appropriate).

## Proposal

My proposal is to add configuration to the ACME provider (TOML version)...

```toml
[certificatesResolvers.myresolver.acme]
  [certificatesResolvers.myresolver.acme.eab]
    keyId = "the key id"
    hmacFile = "/file/path/to/the/hmac/secret"
```

That looks like an update to the `Configuration` struct in the `pkg/provider/acme/provider.go` file. Seems fairly straight forward.

At that point, it's up to register the cert the right way. Based on what I can figure out (I'm still fairly new to go), it looks like the lego library has a separate `RegisterWithExternalAccountBinding` method that would be used instead of the `Register` method on the line below. Seems like a fairly straight forward update to make.

https://github.com/containous/traefik/blob/73ca7ad0c195cb752147a36fba7c715aaa09e2e8/pkg/provider/acme/provider.go#L233. 

Then of course there are the docs, which are pretty straight forward too.

## Questions

- I'd like this to work in both the Traefik 1 and 2 versions. What's the proper way to make a PR to support that? 
- What sorts of tests should we have here? 

Feedback/thoughts welcome!

ldez:
### What does this PR do?

Add external account binding (EAB) support.

### Motivation

Fixes #7118

### More

- ~~[ ] Added/updated tests~~

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -408,7 +408,7 @@ certificatesResolvers:
 [ACME V2](https://community.letsencrypt.org/t/acme-v2-and-wildcard-certificate-support-is-live/55579) supports wildcard certificates.
 As described in [Let's Encrypt's post](https://community.letsencrypt.org/t/staging-endpoint-for-acme-v2/49605) wildcard certificates can only be generated through a [`DNS-01` challenge](#dnschallenge).
 
-## External account binding.
+## External Account Binding
 
 - `kid`: Key identifier from External CA
 - `hmacEncoded`: HMAC key from External CA, should be in Base64 URL Encoding without padding format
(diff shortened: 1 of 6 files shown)


kevinpollet (approved):
LGTM 👍

jbdoumenjou (approved):
LGTM

rtribotte (approved):
LGTM 👌

PR #7609, opened 2020-12-02
Fix TLS options fallback when domain and options are the same 

### LINKED ISSUE #7385: multiple rules for same domain in ingressroute ignores given tls options

Bug

### What did you do?
* created a TLSOptions Resource
* created a Certificate which is valid for two domains host1.example.com & host2.example.com
* created a IngressRoute which contains three rules for the two domains
* i reused a http2 connection established on host2.example.com for host1.example.com

Response from host2.example.com as the tls options are the same

### What did you see instead?
421 Misdirected Request as traefik falls back to default tls options  

when removing the rule with the PathPrefix it works as expected

```
Version:      2.3.1
Codename:     picodon
Go version:   go1.15.2
Built:        2020-09-29T15:49:06Z
OS/Arch:      linux/amd64
```

Provider: Kubernetes

```yaml
kind: IngressRoute
metadata:
  name: example-route
  namespace: example
spec:
  entryPoints:
    - secure
  routes:
    - kind: Rule
      match: Host(`host1.example.com`)
      middlewares:
        - name: headers-ssl
          namespace: traefik2
        - name: headers
          namespace: traefik2
        - name: host1-feature-policy
          namespace: example
        - name: browser-security
          namespace: traefik2
      services:
        - kind: Service
          name: service1
          namespace: example
          port: 3000
    - kind: Rule
      match: Host(`host2.example.com`) && PathPrefix(`/some-path`)
      middlewares:
        - name: headers-ssl
          namespace: traefik2
        - name: headers
          namespace: traefik2
        - name: strip-prefix-example
          namespace: example
      services:
        - kind: Service
          name: service2
          namespace: example
          port: 4445
    - kind: Rule
      match: Host(`host2.example.com`)
      middlewares:
        - name: headers-ssl
          namespace: traefik2
        - name: headers
          namespace: traefik2
      services:
        - kind: Service
          name: service2
          namespace: example
          port: 4444
  tls:
    options:
      name: example-tls-option
      namespace: traefik2
    secretName: example-cert

```

I may also provide more resources but I think looking at https://github.com/traefik/traefik/blob/f2e53a356946454e7fc366fce35378973ce43fd1/pkg/server/router/tcp/router.go#L154 makes clear where the problem is.
Maybe it also should check if the tls options are diffrent?

```
[traefik2-ingress-controller-r84l6] time="2020-10-06T09:38:01Z" level=debug msg="TLS options difference: SNI=default, Header:traefik2-example-tls-option" host=host1.example.com req.Host=host1.example.com req.TLS.ServerName=host2.example.com
```

jspdown:
### What does this PR do?

This PR fixes a bug on the TCP Router when multiple HTTP routers specifies the same domains and TLS options. Fallback on the `default` TLS options should be done only when multiple routers use the same domain with different TLS options.

### Motivation

Fixes #7385

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/server/router/tcp/router_test.go
+++ b/pkg/server/router/tcp/router_test.go
@@ -314,7 +314,7 @@ func TestDomainFronting(t *testing.T) {
 		{
 			desc: "Request is misdirected when TLS options are different",
 			routers: map[string]*runtime.RouterInfo{
-				"router-1": {
+				"router-1@file": {
 					Router: &dynamic.Router{
 						EntryPoints: []string{"web"},
 						Rule:        "Host(`host1.local`)",
@@ -323,7 +323,7 @@ func TestDomainFronting(t *testing.T) {
 						},
 					},
 				},
-				"router-2": {
+				"router-2@file": {
 					Router: &dynamic.Router{
 						EntryPoints: []string{"web"},
 						Rule:        "Host(`host2.local`)",
@@ -336,7 +336,7 @@ func TestDomainFronting(t *testing.T) {
 		{
 			desc: "Request is OK when TLS options are the same",
 			routers: map[string]*runtime.RouterInfo{
-				"router-1": {
+				"router-1@file": {
 					Router: &dynamic.Router{
 						EntryPoints: []string{"web"},
 						Rule:        "Host(`host1.local`)",
@@ -345,7 +345,7 @@ func TestDomainFronting(t *testing.T) {
 						},
 					},
 				},
-				"router-2": {
+				"router-2@file": {
 					Router: &dynamic.Router{
 						EntryPoints: []string{"web"},
 						Rule:        "Host(`host2.local`)",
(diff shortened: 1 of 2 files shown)


dtomcej (approved):
LGTM
:shipit:

pkg/server/router/tcp/router_test.go
```diff
@@ -237,7 +290,186 @@ func TestRuntimeConfiguration(t *testing.T) {
...
+							Options: "host1",
+						},
+					},
+				},
+				"router-2": {
+					Router: &dynamic.Router{
+						EntryPoints: []string{"web"},
+						Rule:        "Host(`host2.local`)",
+						TLS: &dynamic.RouterTLSConfig{
+							Options: "default",
```
rtribotte, line 331:
```suggestion
```

jbdoumenjou (approved):
LGTM

rtribotte (approved):
LGTM 👌

juliens (approved):
LGTM

Issue #7605, opened 2020-12-01
Missing some documentation on how to make work multiple domains, different ports on the same docker node


williamdes:
This is a documentation request, here is the working example I did ([Source](https://gist.github.com/ismailyenigul/867132f298664156068b6e85847c0dfe#gistcomment-3545998))

### `docker-compose.yml`

```yml
version: "3"
services:
   traefik:
    image: traefik:v2.2
   # More config ...
  gophish:
    image: gophish/gophish
    volumes:
      - ${GOPHISH_DB_DIR}:/app/database
      - ${GOPHISH_STATIC_DIR}:/app/static/endpoint
      - ${GOPHISH_STATIC_DIR}:/opt/gophish/static/endpoint
    restart: on-failure:5
    dns_search: example.fr
    hostname: www.example.fr
    domainname: example.fr
    container_name: gophish
    networks:
      internal-server:
        aliases:
          - example.fr
          - www.example.fr
    environment:
      - ADMIN_LISTEN_URL=0.0.0.0:3333
      - ADMIN_USE_TLS=false
      - PHISH_LISTEN_URL=0.0.0.0:888
      - PHISH_USE_TLS=false
      - GOPHISH_INITIAL_ADMIN_PASSWORD=public
      - DB_FILE_PATH=/app/database/gophish.db
    labels:
      traefik.enable: true
      traefik.docker.network: internal-server

      # The HTTPS config
      traefik.http.routers.www-router.rule: host(`www.example.fr`)
      traefik.http.routers.www-router.entrypoints: websecure
      traefik.http.routers.www-router.tls: "true"
      traefik.http.routers.www-router.tls.certresolver: "acmeresolver"
      traefik.http.routers.www-router.service: www

      # The HTTP config
      traefik.http.routers.www-router-http.rule: host(`www.example.fr`)
      traefik.http.routers.www-router-http.entrypoints: web
      traefik.http.routers.www-router-http.service: www

      # The port config for the 2 routers above config
      traefik.http.services.www.loadbalancer.server.port: 888

      # The HTTPS config
      traefik.http.routers.admin-router.rule: host(`gophish-admin.example.fr`)
      traefik.http.routers.admin-router.entrypoints: websecure
      traefik.http.routers.admin-router.tls: "true"
      traefik.http.routers.admin-router.tls.certresolver: "acmeresolver"
      traefik.http.routers.admin-router.service: administration

      # The HTTP config
[...]

PR #7610, opened 2020-12-02
Add example for multiple service per container

notsureifkevin:
### What does this PR do?

Adds a configuration example explaining how to configure more than one service port and router on a single docker container using the docker provider.

### Motivation

While the documentation isn't misleading or explicitly obfuscates this information, it also doesn't give an example of a real-world use case for utilizing the `service` parameter on an HTTP router in the context of the Docker provider. 

In addition, this PR addresses #7605

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/routing/providers/docker.md
+++ b/docs/content/routing/providers/docker.md
@@ -58,7 +58,7 @@ Attach labels to your containers and let Traefik do the rest!
         Setting the label `traefik.http.services.xxx.loadbalancer.server.port`
         overrides that behavior.
 
-??? example "Specifying more than one router per container"
+??? example "Specifying more than one router and service per container"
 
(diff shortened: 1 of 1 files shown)


docs/content/routing/providers/docker.md
```diff
@@ -58,6 +58,26 @@ Attach labels to your containers and let Traefik do the rest!
         Setting the label `traefik.http.services.xxx.loadbalancer.server.port`
         overrides that behavior.
 
+??? example "Specifying more than one router per container"
```
dtomcej, line 61:
```suggestion
??? example "Specifying more than one router and service per container"
```

dtomcej (approved):
LGTM
:shipit:

kevinpollet (approved):
LGTM 👍

rtribotte (approved):
LGTM 👌

PR #7616, opened 2020-12-03
Documentation: Add spacing to sidebars so the last item is always visible

paulocfjunior:
### What does this PR do?

This PR adds some spacing at the bottom of the sidebar menus on the documentation, which is sufficient to make the last item visible when there are many items and we scroll all the way down.

### Motivation

The documentation as we have today, depending on which menus are opened, the last item of the sidebar menu in the documentation (both left or right), might not be visible, because it's covered by the footer bar.
So the next time we export the doc, it will be correct.

### More

- [ ] Added/updated tests

#### Preview
Screen Recording 2020-12-03 at 11 10 48

### DIFF
--- a/docs/content/assets/styles/menu.css
+++ b/docs/content/assets/styles/menu.css
@@ -35,6 +35,10 @@
   padding: 0;
 }
 
+.md-sidebar__scrollwrap {
+  padding-bottom: 50px;
+}
+
 .md-sidebar--secondary .md-sidebar__scrollwrap {
   border-radius: 8px;
   background-color: var(--light-blue) !important;
--- a/docs/content/assets/styles/menu.css
+++ b/docs/content/assets/styles/menu.css
@@ -36,7 +36,7 @@
 }
 
 .md-sidebar__scrollwrap {
-  height: calc(100% - 50px);
+  max-height: calc(100% - 50px);
 }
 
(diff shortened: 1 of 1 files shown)


ldez (approved):
LGTM

kevinpollet (changes requested):
I tested the change on Netlify with Firefox and it doesn't work for me but on Chrome and Safari it works fine.

Maybe something is wrong with my browser cache?

Capture d’écran 2020-12-03 à 16 36 38

paulocfjunior:
I checked here, it doesn't seem to be cache, I'll work on a solution for this that works on other browsers too.

paulocfjunior:
@kevinpollet I switched to another solution that also works for Chrome and all other modern browsers. Would you mind taking a look on it again?

kevinpollet:
Hey @paulocfjunior,

I tested your solution with `Netlify,` the left menu is fixed, but the right menu takes all the page even if there no item.
Is it the expected behaviour?

Capture d’écran 2020-12-05 à 11 15 26

paulocfjunior:
@kevinpollet actually it looks better if it doesn't increase its height unnecessarily, I committed a change that does that.

kevinpollet (approved):
LGTM 👍

jbdoumenjou (approved):
LGTM

PR #7624, opened 2020-12-04
Add ability to set IngressClass.controller attribute different than traefik.io/ingress-controller for using mutli traefik controllers at a time

### LINKED ISSUE #7611: Multiple Treafik Ingress controller in single cluster
We have a requirement for a Multiple Treafik Ingress controller in a single cluster.

My understanding is when we have to customize kubernetes.io/ingress.class. Else restrict the access for the treafik specific controller. Do we have a feature for this? Could you please help me with this?

abdellaui:
### What does this PR do?

This PR adds ability to set the value for the controller attribute of an IngressClass to watch for.

### Motivation

We are using multiple traefik controllers at the same time, one is attached with an internal loadbalancer and one with a public loadbalancer. Now, both traefik controllers are watching for all ingress resources while they are referencing to different IngressClass-Resources.  The reason is, that the IngressClasses have the same value for the controller, which is "traefik.io/ingress-controller".

Fixes #7611

### More

- [x] Added/updated tests


travisghansen:
This. My setup gets very broken with multiple controllers and `IngressClass` resources enabled.

kevinpollet:
Hello @abdellaui,

Have you tried to use the `labelSelector` option?

As described in our [documentation](https://doc.traefik.io/traefik/providers/kubernetes-ingress/#labelselector) the `labelSelector` option can be leveraged to filter `Ingress` objects. Therefore, you could have a different label selector for your `internal` and `public` Traefik instances.

Our understanding of the `IngressClass` specification is that the controller field is not meant to be configurable. So in our case, this value should remain `traefik.io/ingress-controller`.

abdellaui:
Hey @kevinpollet,

according to [kubernetes docs](https://v1-18.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#ingressclassspec-v1beta1-networking-k8s-io) the controller name is immutable, but there is no restriction about configurability less than a recommendation to use a domain-prefixed path for matching with the ingress-controllers implementations.

Of course, we can use the `labelSelector` or the way with annotations without specifying an `spec.ingressClassName`, but this will hold us back to use the new `IngressClass`-Ressource of Kubernetes 1.18

travisghansen:
The field being immutable simply means it can’t be updated after creation, you must delete the class and recreate if you want to ‘update’ the value (this mimics storage class behavior which heavily influenced ingress class structure).

IMO after thinking about this a bit more though, I’m not entirely sure it’s necessary. The controller should be ignoring anything that doesn’t have a matching class anyway right? Meaning, we already have options to tell the traefik instance which class it is, and that value is implicit in creating an IngressClass...it seems we’re simply not using the info we already have.

abdellaui:
Yes, we are able to tell the traefik instance which ingress class it watch for via `providers.kubernetesingress.ingressclass`. But for that, we have to annotate all Ingress-Ressources with a matching `kubernetes.io/ingress.class` value, which is not more necessary / recommended according to [here](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation). Since Kubernetes v1.18 we can specify the ingress-controler over `spec.ingressClassName`, which is referencing to an `IngressClass` that indicates the controller via `spec.controller`. While the value of the controller inside the `IngressClass` is remaining to a constant like `traefik.io/ingress-controller`, we are not able to deploy two traefik instances, because both of them will watch for all `Ingress`-ressources including a reference to a `IngressClass`. It doesn't matter that the referenced IngressClasses are different, since the `spec.controller` value is the same. That is why we need to be able to set the IngressClass-Controller value.

travisghansen:
@abdellaui I'm guess I'm not understanding why we're treating the annotation and the proper `spec.ingressClassName` differently.

If traefik has been told it's class is `X` via `providers.kubernetesingress.ingressclass` then why are we not using/adhering to that in both case of the annotation and the proper `spec.ingressClassName`? What are we trying to make them distinct options?

To be clear, I'm not against allowing configuration of the controller at all. I feel like it's totally reasonable to be able to do so. I'm just not sure it's necessary to achieve the original goal.

I'll also make note (again, understanding that the design of ingress classes is heavily influenced by storage classes) that the idea is that a controller instance can have multiple classes feed into it (this is where the parameters come into play)...I think it would be a grave mistake to facilitate multiple controllers using a single class (which indeed is the issue we're trying to address here). With that, there are a couple different possible approaches:

- continue telling the controller which ingress class(es) it should be responsible for via cli args and effectively ignore the spec.controller value
- tell the controller which spec.controller value it should be looking for, and have it only handle the associated class names and *not* tell the controller explicitly which classes it watches explicitly via command line args (essentially, auto-discovery of class names via the k8s api instead of using explicit launch time args)

From a backwards compatibility standpoint, the first seems the appropriate route. If we go the second path (I think what you're mostly vouching for), then it's definitely a requirement to be able to 'bind' the controller instances to classes via `spec.controller` in some shape or form which implies `spec.controller` should be configurable (at least a creation time).

In *either* case however, the controller should be watching ingresses with the proper annotation **and** ingresses with the `spec.ingressClassName` with appropriate values (however that list of classes is being determined doesn't really matter) IMO.

kevinpollet:
Hi @abdellaui and @travisghansen,

Sorry for the delayed response...

> Of course, we can use the labelSelector or the way with annotations without specifying a spec.ingressClassName, but this will hold us back to use the new IngressClass-Ressource of Kubernetes 1.18

Just to be accurate it is possible to use the `labelSelector` in conjunction with the `spec.ingressClassName` (which sounds a bit redundant we must admit). The only limitation we currently have in our implementation is that we are retrieving only one `IngressClass` in the cluster, the PR #7624 has been opened to fix that.

> Yes, we are able to tell the traefik instance which ingress class it watch for via providers.kubernetesingress.ingressclass. But for that, we have to annotate all Ingress-Ressources with a matching kubernetes.io/ingress.class value, which is not more necessary / recommended according to here

You are right the [`ingressClass`](https://doc.traefik.io/traefik/providers/kubernetes-ingress/#ingressclass) option has no effect with the new `IngressClass` resource. After some discussions, we think that allowing the configuration of the matching `controller` field value is not the way to go.

We think that the best option is to support the provider option `ingressClass` with the new `IngressClass` resource. This means that when the `ingressClass` option is defined we will create the configuration only for the ingresses which reference an `IngressClass` resource with the configured name and the controller value of `traefik.io/ingress-controller`.

Do you think that you can make the changes in the following days? If not, we will take over on this subject and fix it in the coming days.

Thanks again for your time.

travisghansen:
So the desired end result here is:

- legacy ingresses with the annotation will continue to be driven by the `providers.kubernetesingress.ingressclass` option
- Ingresses with the `ingressClassName` will also be driven by the `providers.kubernetesingress.ingressclass` option
- All ingress classes (resource definitions) will be set with the hard-coded controller value
- Will properly support multiple deployments/ingress classes for ingresses with the new `ingressClassName`

kevinpollet:
Exactly 👍

abdellaui:
I'm sorry, I dont think that i will make the changes in the upcoming days.

kevinpollet:
No worries we will work on it.
Thanks again for your work and your time! 

I will close this PR accordingly.

travisghansen:
@kevinpollet can you link to the new PR when it's created so we can follow along?

kevinpollet:
Hello all and sorry for the late answer,

For your interest, here is the link to the pull request: #7915

Issue #7266, opened 2020-09-08
Add HostSNIRegexp


varac:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

I'd like to have a `HostSNIRegexp` so that I can route all subdomains to a certain service, without referencing them all at once.
I.e. similar to the `HostRegexp` rule:

    HostSNIRegexp(`example.com`, `{subdomain:[a-z]+}.example.com`, ...)

That would save time with static configs and a lot of subdomains.

ldez:
related to  #6003

FuNK3Y:
Any news about this one? On the related PR it was mentioned that a discussion should take place about the replacement of mux. Is it the right place?

varac:
I also don't understand if #6003 fixes/closes this.. ?

varac:
Ok, it seems this is still not implemented. #6003 was closed and *not* merged.

nmasse-itix:
Hi ! 

I'm interested in this feature (using it and eventually contributing to). 

How can we make this issue progress one step further ?
Has the new regex syntax mentioned in #7635 been implemented ?

> As explained in # 6003, we think that the mux syntax for regex is not perfect. For a new regex matcher we think it's not a good idea to use the mux syntax, but if we introduce a new syntax, we need to handle it in both TCP and HTTP regex matches.

Blackclaws:
This is also relevant to me. I have a downstream cluster that handles all requests for a certain subdomain and I cannot know in advance what the domains will be. TLS Termination in traefik is not an option unfortunately.

JeanTekk:
I would be also glad about such a feature.
I am have some docker services on a different network. The docker host runs also a traefik instance. The traefik on the docker host routes all containers on common ports and does the detection of the services and ssl via labels. 
The docker host forwads the traefik ports for web and websecure(ssl) over an ssh tunnel to another publicly accessible  gatway.
From there i would like to route all requests matching (e.g. "*.docker.myhost.com") to the tunnels.

Soffid:
Hello all.

I've faced the same problem, and managed to patch Traefik 2.5.4 to allow wildcards in TCP  rules. Now, I can write a rule like this one:

- match: HostSNI(`iam3-gw.soffid.com`)||HostSNI(`*.iam3-gw.soffid.com`) 
 
This is the first time I code in GO and I am not a Traefik expert, but as long as we have tested, it has worked fine.

Hope it helps you and can be merged in future versions.

[...]

PR #7635, opened 2020-12-08
Allow HostSNI values to match on a wildcard *. subdomain

richbayliss:
### What does this PR do?

Allow HostSNI values to match on a wildcard *. subdomain

### Motivation

This was needed to allow any subdomains of a given domain to
be passed through to a single service.

### More

- [x] Added/updated tests

### DIFF
--- a/integration/fixtures/tcp/mixed.toml
+++ b/integration/fixtures/tcp/mixed.toml
@@ -56,6 +56,13 @@
       entryPoints = [ "tcp" ]
       [tcp.routers.to-whoami-no-cert.tls]
 
+    [tcp.routers.to-whoami-wildcard]
+      rule = "HostSNI(`*.whoami-a.test`)"
+      service = "whoami-a"
+      entryPoints = [ "tcp" ]
+      [tcp.routers.to-whoami-wildcard.tls]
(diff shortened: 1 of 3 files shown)


richbayliss:
For some reason, after following the docs, I couldn't get the tests to pass on my dev machine; I am not sure why but the Docker tests would fail, so my added test code hasn't been checked locally. If I have understood the test framework correctly then I expect they should pass though, and I have manually run this build and checked the feature works as expected/intended.

kevinpollet:
Hello @richbayliss,

Thanks for your interest in Traefik and your pull request!

The implementation you have made does not conform to what we have implemented for the `HeadersRegexp` and `HostRegexp` rules. To be consistent a new `rule` should be implemented which supports more complex use cases. 

As explained in https://github.com/traefik/traefik/pull/6003#issuecomment-640494203, we think that the `mux` syntax for `regex` is not perfect. For a new `regex` matcher we think it's not a good idea to use the mux syntax, but if we introduce a new syntax, we need to handle it in both TCP and HTTP regex matches. 

That's why we think it's better to talk about this before in #7266.

Therefore, we will decline your pull request accordingly.

PR #7654, opened 2020-12-12
Add setting 'AppendURIToAddress' to ForwardAuth Middleware

maduma:
ForwardAuth middleware pass original request URI via X-Forwarded-Uri
header. Having an option to append the original URI on the
authentication server adress is very helpfull. This help using
'out-of-the-box' auth server expecting the path of the protected
resource in the url (e.g Forgerock AM)

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/middlewares/forwardauth.md
+++ b/docs/content/middlewares/forwardauth.md
@@ -174,6 +174,57 @@ http:
         trustForwardHeader: true
 ```
 
+### `appendURIToAddress`
+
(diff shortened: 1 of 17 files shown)


rtribotte:
Hello @maduma,

Thanks for your time and interest in Traefik!

After discussions, we prefer to evaluate the opportunity and the different use cases in a dedicated issue before introducing this new option.
In other words, we would like to have feedback on this to evaluate what is the best approach.

We would definitely like to see your pull request reopening after the way to do this enhancement has been discussed.
For now, I will decline this pull request, but as I already mentioned, I please you to open a new issue to discuss further the topic.

Issue #7617, opened 2020-12-03
Allow configuration of SameSite settings in Service Fabric provider


nokjuh:
Feature

Support for SameSite and Secure configuration in Service Fabric provider sticky sessions, similar to other providers. The support to most other providers was added in 1.7.23, but Service Fabric was missing.

Service Fabric provider is not yet supported in Traefik 2.x. Hopefully it will be at some point, but this seems like a small fix that could be included in 1.7.x.

It should be possible to add at least SameSite=None and Secure=true to sticky sessions in container labels.

For example same configuration as with other providers:
| `traefik.backend.loadbalancer.stickiness.secure=true`                   | Sets secure cookie option for sticky sessions.                                                                                                                                                                                |
| `traefik.backend.loadbalancer.stickiness.httpOnly=true`                 | Sets http only cookie option for sticky sessions.                                                                                                                                                                             |
| `traefik.backend.loadbalancer.stickiness.sameSite=none`                 | Sets same site cookie option for sticky sessions. (`none`, `lax`, `strict`)

Related: #6261

nokjuh:
This is working in Traefik v1.7.28, thank you for the fix.

PR #7667, opened 2020-12-14
Fix stickiness support in Service Fabric.

### LINKED ISSUE #6261: Allow configuration of SameSite settings for sticky session cookies
Feature
With Chrome 80 going stable on Feb 4, 3rd party cookies without the `SameSite=None` setting will be blocked. https://blog.chromium.org/2019/10/developers-get-ready-for-new.html. In our testing, this breaks sticky sessions for services that are iframed into another page.

Relates to #2653, #4738

tpayne84:
In fix #6261 that adds the ability to configure [ `sameSite`, `httpOnly`, `secure` ] cookie options... the template for service fabric did not receive updates and still does not support these options.

### What does this PR do?

This change adds the support for the options noted above by adding entries for them into the service fabric template file.

Documentation has also been updated to include support for the added config values.

### Motivation

As noted in issue #7617, the functionality was added to most other providers in the fix for #6261... However the service fabric template is not found in the same location as the others and was not updated as part of the fix.

### More

- [ ] Added/updated tests

### Depends on 
Template change [PR in containous/traefik-extra-service-fabric](https://github.com/containous/traefik-extra-service-fabric/pull/60)


ldez (approved):
Thanks :+1:

Issue #10, opened 2015-09-23
TCP support


chendo:
Would be awesome if this also supported TCP backends (ideally with optional TLS support)!

EmileVauge:
It needs some (a lot of ?) work... But it would be awsome :)

xiaods:
TCP + TLS ? need more details on this proposal

pragone:
+1. 
I love the concept of traefik for a microservices environment as it facilitates service discovery and balancing. But not all services are HTTP based. So, from what I understand, it can't be used for load-balancing MySQL or similar services.
If I have a microservices environment where I need another LB technology for any non-http service (for example HAProxy + Marathon template) I won't adopt traefik just to keep the number of moving pieces low. I think therefore that this is an important feature.
My 2 cents

Ozzadar:
I highly support the addition of this feature =)

faddat:
+1

rvm2015:
+1 on this, I've been pushing a similar project called Fabio that has already solved this with a consul backend to add Mesos/Marathon as an option but I'd really like to see this here.

emilevauge:
Hi @rvm2015, as far as I know, fabio is only supporting HTTP/S right now. How are you doing this with Fabio?

rvm2015:
You are correct, I am mistaken. I've been shopping around for a TCP load balancer and I misread their feature set.

altvnk:
+1 for TCP services

faddat:
+1 indeed.  We currently use an ugly hack to do it.  I feel that traefik
was a lot more elegant, not to mention stable.

Jacob Gadikian
E-mail: faddat@gmail.com
SKYPE: faddat
Phone/SMS: +84 167 789 6421

tymofii-polekhin:
+1 for TCP

klausenbusk:
+1

hd-deman:
@faddat how did you hacked this?

Ozzadar:
@hd-deman It wasn't with traefik. It was a combination of adding Consul to our marathon stack, using a marathon-consul container to report changes in applications to consul then using a haproxy-consul bridge to port forward from there.

I'm not going to link here but if you'd like links to the different components feel free to send me a message and I'll forward you to them. This stack isn't extremely suitable for production though as you'd need to write a watcher daemon to make sure that all the components don't explode on you. (as they tend to do)

faddat:
see @ozzadar's answer-- he's the one who actually implemented the hack.
[...]

PR #7675, opened 2020-12-16
Add missing quotes in errorpages k8s example yaml

icelynjennings:
### What does this PR do?

Fix docs

### Motivation

Adding quotations prevents the error:

```
v1alpha1.ErrorPage.Status: []string: ReadString: expects " or n, but found 4, error found in #10 byte of ...|status":[401]}}},{"a|..., bigger context ...|ice":{"name":"XXXX","port":80},"status":[401]}}},{"apiVersion":"traefik.containo.us/v1alpha1|...
```

### DIFF
--- a/docs/content/middlewares/errorpages.md
+++ b/docs/content/middlewares/errorpages.md
@@ -28,7 +28,7 @@ metadata:
 spec:
   errors:
     status:
-      - 500-599
+      - "500-599"
     query: /{status}.html
     service:
       name: whoami


rtribotte (approved):
LGTM 👍

kevinpollet (approved):
Thanks 👍

jbdoumenjou (approved):
LGTM :+1:

ldez (approved):
Thanks :+1:

feily-os:
hi！ 
I have the same problem in v2.4
So, what should I do?



`Failed to watch *v1alpha1.Middleware: failed to list *v1alpha1.Middleware: v1alpha1.MiddlewareList.Items: []v1alpha1.Middleware: v1alpha1.Middleware.Spec: v1alpha1.MiddlewareSpec.Errors: v1alpha1.ErrorPage.Status: []string: ReadString: expects " or n, but found 4, error found in #10 byte of ...|status":[403]}}},{"a|..., bigger context ...|service":{"name":"xxx","port":8080},"status":[403]}}},{"apiVersion":"traefik.containo.us/v1alpha1|...`

icelynjennings:
Hi @feily-os, I'm unsure you'll have had the same issue I. When this appeared for me was when I copied a Kubernetes example from https://doc.traefik.io/traefik/master/middlewares/errorpages/. If you followed similar steps, defining the middleware object as following worked for me:

```
kind: Middleware
metadata:
  name: test-errorpage
spec:
  errors:
    status:
      - "500-599"
    query: /{status}.html
    service:
      name: whoami
      port: 80
```

Notice above that I had added quotation marks around the status code range, as follows: `"500-599"` 

The error no longer appeared afterwards.

feily-os:
Hi @icelynjennings, I do the same. 
I change the status configuration, And the documentation states that integers can be used. 

`You can define either a status code as a number (500) or ranges by separating two codes with a dash (500-599).`

I tried to use `"400-403"` but it was not redirected to the specified static page. 
I tried to verify this middleware simply, but it failed at present

PR #7711, opened 2020-12-28
Fix wildcard hostname issue

### LINKED ISSUE #7497: Kubernetes Ingress PathPrefix service name with dash duplicate
### What did you do?

Declare a kubernetes Ingress resource with 2 rules with different names but almost the sames :

```yml
kind: Ingress
metadata:
  annotations:
    traefik.ingress.kubernetes.io/router.middlewares: strip@file # middleware stripPrefixRegex with regex ["/"]
  name: whoami
  namespace: traefik
spec:
  rules:
  - host: domain.com
    http:
      paths:
      - backend:
          serviceName: whoami
          servicePort: web
        path: /whoami/test
  - host: domain.com
    http:
      paths:
      - backend:
          serviceName: whoami
          servicePort: web
        path: /whoami-test
  tls:
    - secretName: traefik-default-cert
```

I Can access both path `/whoami-test` and `/whoami/test`

### What did you see instead?

Only `/whoami-test` is working
If I change the order of the rules in the ingress configuration the last one is working (/whoami/test) but not the first one.
I think it's an issue with the generated name of the traefik router :
Both ingress rules have the same router name : `websecure-whoami-traefik-traefik-domain-com-whoami-test`

```
image: traefik:2.2.7
```

```toml
[providers]
  [providers.kubernetesIngress]
  [providers.kubernetesIngress.ingressEndpoint]
      publishedService = "traefik/traefik-ingress-service"
  [providers.file]
    directory = "/config"
    watch = true
```

### LINKED ISSUE #7691: Wildcard(*) in ingress.spec.host is not honored

Feature

Request  to  `*.somthing.svc.foo.bar/api` goes to to correct svc ie `k8s-service-app`

### What did you see instead?
Request  to `somthing.svc.foo.bar/api` goes to **correct** svc.
Request  to  `*.somthing.svc.foo.bar/api` goes to to **incorrect**(default) svc.  

```
Version:      2.3.3
Codename:     picodon
Go version:   go1.15.5
Built:        2020-11-19T17:43:21Z
OS/Arch:      linux/amd64
```

```yaml
#traefik.yaml
serversTransport:
  insecureSkipVerify: true
entryPoints:
  http:
    address: :8080
    proxyProtocol:
      trustedIPs:
      - 10.0.0.0/8
  https:
    address: :8443
    http:
      tls: {}
    proxyProtocol:
      trustedIPs:
      - 10.0.0.0/8
  traefik:
    address: :8888
  ping:
    address: :9000
providers:
  providersThrottleDuration: 2s
  kubernetesIngress:
    ingressEndpoint:
      publishedService: traefik/traefik
  kubernetesCRD: {}
tls:
  options:
    default:
      minVersion: "VersionTLS11"
api:
  insecure: true
ping: 
  entryPoint: "ping"
metrics:
  prometheus: {}
log:
  level: DEBUG
  format: json
accessLog:
  format: json
  fields:
    defaultMode: keep
    names:
      ServiceURL: keep
      ServiceName: keep
      RouterName: kep
      ClientAddr: keep
      Duration: keep
      OriginDuration: keep
      OriginStatus: keep
      RequestMethod: keep
      RequestPath: keep
      RequestProtocol: keep
      RequestUserAgentHeader: keep
      RetryAttempts: keep
      headers: keep
    headers:
      defaultMode: drop
      names:
        CF-Connecting-IP: keep
        User-Agent: keep
        X-Forwarded-For: keep

```
```yaml
kind: Ingress
metadata:
  annotations:
  name: k8s-service
  namespace: my-path
spec:
  rules:
  - host: '*.somthing.svc.foo.bar'
    http:
      paths:
      - backend:
          serviceName: k8s-service-app
          servicePort: 8000
        path: /my-path
      - backend:
          serviceName: k8s-service-app
          servicePort: 8000
        path: /api
  - host: '*.somthing.api.foo.bar'
    http:
      paths:
      - backend:
          serviceName: k8s-service-app
          servicePort: 8000
        path: /my-path
      - backend:
          serviceName: k8s-service-app
          servicePort: 8000
        path: /api
  - host: somthing.svc.foo.bar
    http:
      paths:
      - backend:
          serviceName: k8s-service-app
          servicePort: 8000
        path: /my-path
      - backend:
          serviceName: k8s-service-app
          servicePort: 8000
        path: /api
  - host: somthing.api.foo.bar
    http:
      paths:
      - backend:
          serviceName: k8s-service-app
          servicePort: 8000
        path: /my-path
      - backend:
          serviceName: k8s-service-app
          servicePort: 8000
        path: /api
```
There is **no** `*` in dashboard rule. 
no-star

```
(paste your output here)
```

avdhoot:
### What does this PR do?
Fix #7691

### Motivation
This affecting our migration from 1.x to 2.x. Same ingress is working in 1.x well. 

### More

- [X] Added/updated tests

### Additional Notes

Please check by changing router key if anythings breaks. We running this in our sandbox cluster for an weeks without issue with build `avdhoot/traefik:v2.3.6-host-fix`

### DIFF
--- a/pkg/provider/configuration.go
+++ b/pkg/provider/configuration.go
@@ -349,7 +349,7 @@ func BuildRouterConfiguration(ctx context.Context, configuration *dynamic.HTTPCo
 // Normalize Replace all special chars with `-`.
 func Normalize(name string) string {
 	fargs := func(c rune) bool {
-		return !unicode.IsLetter(c) && !unicode.IsNumber(c) && c != '*'
+		return !unicode.IsLetter(c) && !unicode.IsNumber(c)
 	}
 	// get function
 	return strings.Join(strings.FieldsFunc(name, fargs), "-")
--- a/pkg/provider/configuration.go
+++ b/pkg/provider/configuration.go
@@ -349,7 +349,7 @@ func BuildRouterConfiguration(ctx context.Context, configuration *dynamic.HTTPCo
 // Normalize Replace all special chars with `-`.
 func Normalize(name string) string {
 	fargs := func(c rune) bool {
-		return !unicode.IsLetter(c) && !unicode.IsNumber(c)
+		return !unicode.IsLetter(c) && !unicode.IsNumber(c) && c != '*'
 	}
 	// get function
 	return strings.Join(strings.FieldsFunc(name, fargs), "-")
--- a/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-conflicting-routers-on-host_endpoint.yml
+++ b/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-conflicting-routers-on-host_endpoint.yml
@@ -1,0 +1,15 @@
+kind: Endpoints
+apiVersion: v1
+metadata:
+  name: service1
+  namespace: testing
+
+subsets:
+- addresses:
+  - ip: 10.10.0.1
+  ports:
+  - port: 8080
+- addresses:
+  - ip: 10.21.0.1
+  ports:
+  - port: 8080
--- a/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-conflicting-routers-on-host_ingress.yml
+++ b/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-conflicting-routers-on-host_ingress.yml
@@ -6,14 +6,7 @@ metadata:
 
 spec:
   rules:
-  - host: '*.tarefik.tchouk'
-    http:
-      paths:
-      - path: /bar
-        backend:
-          serviceName: service1
-          servicePort: 80
-  - host: tarefik.tchouk
+  - host: "*.bar"
     http:
       paths:
       - path: /bar
@@ -21,3 +14,10 @@ spec:
           serviceName: service1
           servicePort: 80
 
+  - host: "bar"
+    http:
+      paths:
+      - path: /bar
+        backend:
+          serviceName: service1
+          servicePort: 80
--- a/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-conflicting-routers-on-host_service.yml
+++ b/pkg/provider/kubernetes/ingress/fixtures/Ingress-with-conflicting-routers-on-host_service.yml
@@ -1,10 +1,10 @@
----
 kind: Service
 apiVersion: v1
 metadata:
   name: service1
   namespace: testing
+
 spec:
   ports:
   - port: 80
-  clusterIp: 10.0.0.100
+  clusterIp: 10.0.0.1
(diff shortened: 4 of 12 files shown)


rtribotte:
Hello @avdhoot,

Thank you for your interest in Traefik and your time on this.

The fix your are providing comes with two possible issues (that remain the same):

1. Same host rules without path will produce different routers
2. Same path rules without host will produce different routers

Those routers would have the same matching configuration but possibly refer to different services.
In both cases, the previous behavior was to keep only the last rule's router (and was originating the issue you are addressing).

This PR will not fix #7497 when the paths that would be normalized in the same form (e.g. `/path/to` and `/path-to`) are declared in the same rule, because you are using the rule index and not the path index to create the router's key.
As the title of your pull request state it explicitly, it would be better to address only the wildcard domain issue in this PR.

Since this key has a meaning only in the Traefik point of view (no matter which index was the rule or the path in the ingress object), I would suggest adding the `*` rune to the `Normalize` function (to address only the wildcard domains router's key generation issue).

avdhoot:
Thanks @rtribotte for all explanation. I assumed that there might be reason(s) why special stings are replaced with `-`. Hence did count thing. 
 
 Is this right code change to keep `*` in router name.  [ref](https://github.com/avdhoot/traefik/blob/c21597c593fdee87018bba2655766ff3e492913d/pkg/provider/configuration.go#L382) 
```go
fargs := func(c rune) bool {
		return !unicode.IsLetter(c) && !unicode.IsNumber(c)  && string(c) != "*"
	}

**Comment by rtribotte** [comment]
@avdhoot This is the right place, but I would use:
```go
... && c != '*'
```

avdhoot:
@rtribotte Done & thanks for all help. Not sure why `DockerSuite.TestDockerContainersWithTCPLabels` failing on local + plus here.

avdhoot:
FYI @rtribotte After deployed 2nd patch(keeping `*` ) Middleware & dashboard/api are not working.  Docker tag `avdhoot/traefik:v2.3.6-host-fix-after-reiew`

rtribotte:
Hello @avdhoot,

Thanks for your feedback.
Thus I wanted to give you the actual state of the thoughts of the team regarding the fix.

First, I probably misled you, as my first comment was to give you my feedback on your first implementation since the pull request is in the "design-review" state. This means that we are evaluating the opportunity and the manner that a given pull request is providing a fix or an enhancement.
When I suggested how to solve it, by allowing the rune `*` in the normalize function, it was more to open the discussion and it's actually a bad idea as the function usage is not only scoped to the Kubernetes Ingress provider and we can expect side effects (as you just experienced).

Second, after discussion, we think that we must solve the collision on the router's key for both host and path cases.

We want to provide the following fix:

1. Detect router's keys collision for one ingress
2. Append at the end of the router's key a hash based on router's rule, for colliding routers only
3. Notify with a debug log that a collision has been detected and solved by adding this hash

It's a backward compatible approach that implies change only for collided routers.

Nonetheless, we thank you very much for your time on this PR.
As it derived from your PR and for the sake of simplicity, we'll take it over and address that fix in a new commit.

Please don't hesitate to give us your feedback on this.

rtribotte (approved):
LGTM

kevinpollet (approved):
LGTM 👍

jbdoumenjou (approved):
LGTM :+1:

PR #7725, opened 2021-01-06
Support for private plugins

rocdove:
1. support offline
2. support private plugins


ldez:
duplicate https://github.com/traefik/traefik/pull/7515#issuecomment-726133992

PR #7752, opened 2021-01-13
Increase sleeps and grace period in shutdown tests

sylr:
### What does this PR do?

Increasing sleeps and grace period.

### Motivation

Test execution on CI hosts with limited performances induce test flackiness.

### More

- [x] Added/updated tests


rtribotte:
Hello @sylr,

I see that you are trying to fix some flaky tests.
Thanks for that, it's annoying and it will be very useful to fix them.
But trying to fix it by pushing changes is also annoying as the CI has limited resources.
We will take over this because we have access to the CI machine, and it will be easier to re-run the tests.

sylr:
@rtribotte I think this should already be plenty to greatly improve current situation.

kevinpollet:
Hello @sylr,

As we have access to the CI, it will be easier for us to find the root cause of the failing tests in order to fix them without increasing the timeouts (which is maybe not related to the underlying problem).

Therefore, we will take over on this subject and close this pull request accordingly.

Thanks again for your time and efforts.

PR #7782, opened 2021-01-20
Fix plugin type on middleware endpoint response

### LINKED ISSUE #7778: Custom middleware misses type in dashboard

Bug

### What did you do?
Added a custom plugin using custom middleware and opened the Traefik dashboard to view the custom middleware.

A custom middleware type defined by the plugin developer or at least something like 'Custom'

### What did you see instead?
Empty middleware type

image

```
Version:      2.3.7
Codename:     picodon
Go version:   go1.15.6
Built:        2021-01-11T18:03:02Z
OS/Arch:      linux/amd64
```

jspdown:
### What does this PR do?

This PR fixes a bug on the `/api/http/middlewares/{middlewareID}` endpoint when the requested middleware is a plugin. In this case the response didn't return the plugin type but instead an empty string.

### Motivation

Fixes #7778

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/api/handler.go
+++ b/pkg/api/handler.go
@@ -8,6 +8,7 @@ import (
 
 	assetfs "github.com/elazarl/go-bindata-assetfs"
 	"github.com/gorilla/mux"
+	"github.com/traefik/traefik/v2/pkg/config/dynamic"
 	"github.com/traefik/traefik/v2/pkg/config/runtime"
 	"github.com/traefik/traefik/v2/pkg/config/static"
 	"github.com/traefik/traefik/v2/pkg/log"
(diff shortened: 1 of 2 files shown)


pkg/api/handler.go
```diff
@@ -157,6 +158,14 @@ func extractType(element interface{}) string {
 	v := reflect.ValueOf(element).Elem()
 	for i := 0; i < v.NumField(); i++ {
 		field := v.Field(i)
+
+		if field.Kind() == reflect.Map && field.Type().Elem() == reflect.TypeOf(dynamic.PluginConf{}) {
+			keys := field.MapKeys()
+			if len(keys) == 1 {
```
dtomcej, line 164: can `keys` only have 0 or 1 entries? is it possible to have 2?

what about:
```
if keys := field.MapKeys(); len(keys) > 0 {
  return keys[0].String()
}
```

jspdown: If misconfigured it can indeed be 0 or greater than 1. But if there's more than one plugin in the `PluginConf` it's going to be considered invalid. So, I think we shouldn't return any type for this middleware.

dtomcej:
```suggestion
if keys := field.MapKeys(); len(keys) == 1 {
  return keys[0].String()
}```

File: pkg/api/handler.go [hunk 2]
```diff
@@ -157,6 +158,14 @@ func extractType(element interface{}) string {
...
+				return keys[0].String()
+			}
```
dtomcej, line 166:
```suggestion
			if keys := field.MapKeys(); len(keys) == 1 {
				return keys[0].String()
			}
```

dtomcej (approved):
LGTM
:shipit:

rtribotte (approved):
LGTM 👍

ldez (approved):
LGTM

PR #7789, opened 2021-01-22
Improve CA certificate loading from kubernetes secret

### LINKED ISSUE #7788: [k8s ServersTransport] Remove strict check on file numbers in certificate/rootCAs secrets
Instead of checking the number of files in a secret when referencing them in a `ServersTransport` under either `spec.rootCAsSecrets` and `spec.certificatesSecrets`, check for the presence of specific file names.

Right now when using it as a `rootCAsSecrets` entry it expects a single field to be there and it simply uses that field. When using a secret as a `certificatesSecrets` entry it checks if there are exactly 2 files and pulls out the two files called `tls.crt` and `tls.key`.

When using cert-manager at least 3 different issuers add the root CAs certificate if possible as `ca.crt`. As far as I know it's the `SelfSigned`, `CA` and `Vault` issuers. It could also be the case that people want to put a lot of different secret files in the same secret for whatever reason. 

Currently it prints an error in the access logs:
```
time="2021-01-22T16:22:46Z" level=error msg="Error while loading certificates secret-tls: found 3 elements for secret 'some-namespace/secret-tls', must be two elements exactly" providerName=kubernetescrd serversTransport=secret-mtls
```
### What  did you expect to see?

I expected it to simply pull out a file called `ca.crt` when we reference a secret under `rootCAsSecrets` and when referenced under `certificatsSecrets` pull out the files `tls.crt` and `tls.key` as it is doing now but do not enforce the not more than 2 files rule.

rio:
Instead of relying on an exact number of entries in a secret for use as a root ca or mTLS certificate set, we check for the presence of specific names.
For the rootCA's field we keep backwards compatibility with the previous behavior of just picking the first entry but only after an explicit check for `ca.crt` and `tls.ca` fails.

fixes #7788

### DIFF
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -447,7 +447,7 @@ metadata:
 
 spec:
   clientAuth:
-    # the CA certificate is extracted from key `tls.ca` of the given secrets.
+    # the CA certificate is extracted from key `tls.ca` or `ca.crt` of the given secrets.
     secretNames:
       - secretCA
     clientAuthType: RequireAndVerifyClientCert
--- a/docs/content/middlewares/http/forwardauth.md
+++ b/docs/content/middlewares/http/forwardauth.md
@@ -373,7 +373,8 @@ metadata:
   namespace: default
 
 data:
-  ca: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0=
+  # Must contain a certificate under either a `tls.ca` or a `ca.crt` key. 
+  tls.ca: LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCi0tLS0tRU5EIENFUlRJRklDQVRFLS0tLS0=
 ```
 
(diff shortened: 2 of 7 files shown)


Rio:
I'm not sure if this could be considered for 2.4.1. If it is I'd very much like it to go out soon. I'll rebase on the v2.4 branch if needed.

jbdoumenjou:
Hello @Rio,
 
Thanks for your contribution!

As you can see in our [contribution guide](https://github.com/traefik/contributors-guide/blob/master/review_guidelines.md), that means that we have to take a closer look at the scope of impact for that PR to see, how it would interact with the other parts of Traefik.

Rio:
Thanks @jbdoumenjou, very much appreciated. I'm not sure what is happening but the integration tests seem to be failing constantly for be, including the main branch without any changes. Could be me doing something wrong of course.

jbdoumenjou:
Don't worry, we are facing some issues with the CI :wink:

jbdoumenjou (approved):
LGTM

dtomcej (approved):
LGTM
:shipit:

rtribotte (approved):
LGTM

PR #7792, opened 2021-01-24
Documentation: Add examples

Fonata:
Feedback to this documentation change suggestions are welcome.

### What does this PR do?

Only suggest minor documentation improvements.

### Motivation
I suggest small documentation changes that would have saved me a lot of trial and error.

### More

- [x] Markdown lint is green

### DIFF
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -14,7 +14,7 @@ See the [Let's Encrypt](./acme.md) page.
 To add / remove TLS certificates, even when Traefik is already running, their definition can be added to the [dynamic configuration](../getting-started/configuration-overview.md), in the `[[tls.certificates]]` section:
 
 ```toml tab="File (TOML)"
-# Dynamic configuration
+# Dynamic configuration only; don't put this in the static traefik.toml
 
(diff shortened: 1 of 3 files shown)


ldez:
> Dynamic configuration only; don't put this in the static traefik.toml
> ...
> Dynamic configuration only

The [dynamic configuration](https://docs.traefik.io/v2.4/getting-started/configuration-overview/#the-dynamic-configuration) and the [static configuration](https://docs.traefik.io/v2.4/getting-started/configuration-overview/#the-static-configuration) have completely different meaning, all is explained here: https://doc.traefik.io/traefik/v2.4/getting-started/configuration-overview/
So it's useless to say that is only for dynamic configuration.

> For HTTPS endpoints one needs to specifiy the `loadbalancer.server.scheme=https` 

This is wrong:
- the label `loadbalancer.server.scheme` is only to define the scheme used by the applications called by Traefik
- the page `docs/content/routing/routers/index.md` is the router's configuration only and this page only uses the file provider.
- this label needs to be used only to force a scheme, in 99% of the case, you don't need to define it.

So, thanks for your time but we will have to decline your PR.

PR #7833, opened 2021-01-30
tls Manager: do not build a default certificate for ACME challenges store

### LINKED ISSUE #7831: Traefik v2.4 on Kubernetes generates new certificate on each config reload

Bug

### What did you do?

Changed some ingress configuration.

Expected traefik to reload its configuration with its default certificate loaded, without generating a new one internally.

### What did you see instead?

I am observing the log message whenever a configuration change happens:
```
time="2021-01-30T20:17:53Z" level=debug msg="No default certificate, generating one"
```

```
Version:      2.4.0
Codename:     livarot
Go version:   go1.15.6
Built:        2021-01-19T17:26:51Z
OS/Arch:      linux/amd64
```

In Kubernetes, command line arguments:
```
        - --providers.kubernetescrd
        - --entrypoints.http.address=:8080
        - --entrypoints.https.address=:8443
        - --entrypoints.https.http.tls=true
        - --entrypoints.traefik.address=:8000
        - --ping
        - --serverstransport.forwardingtimeouts.dialtimeout=1
        - --api
        - --api.debug
        - --api.insecure
        - --log.level=DEBUG
```

Also, my traefik installation has a default store defined:

```
$ kubectl -n ingress-traefik get tlsstores.traefik.containo.us -o wide
NAME      AGE
default   13h
```

It seems that https://github.com/traefik/traefik/blob/2461e36ed466b4be236df9d43bafefcd667e29b5/pkg/tls/tlsmanager.go#L97 is requesting a store which is not the default, and I even dont know how to define one for it in K8S.

So, in an environment, where configuration change happens frequently, traefik consumes much cpu on generating certificates. This was not the case with traefik v1.7 with the following command line arguments:

```
        - --api
        - --kubernetes
        - --checknewversion=false
        - --entrypoints=Name:http Address::8000
        - --entrypoints=Name:https Address::8443 TLS:/certs/server.crt,/certs/server.key TLS.MinVersion:VersionTLS12
        - --forwardingtimeouts.dialtimeout=5s
```

rkojedzinszky:
### What does this PR do?

The changes in this PR are two-fold:

1) It makes sure that when initializing the TLS store for the ACME challenges certificates, it does not generate a (useless) default certificate (as opposed to with the other TLS stores). This solves the initial problem reported, as it won't uselessly burn CPU anymore for that on configuration reload.

2) While we were at it, we cleaned up the mutex lockings and the related writes/initialization on the Manager struct. The end result is that UpdateConfig method is now the only one actually mutating to the struct, while the Get and GetStore methods are now read-only (as their names suggested).

### Motivation

On slower hardware generating certificates can cause noticeable delays when configuration changes.

### More

~- [ ] Added/updated tests~

Fixes #7831

### DIFF
--- a/pkg/tls/tlsmanager.go
+++ b/pkg/tls/tlsmanager.go
@@ -153,6 +153,8 @@ func (m *Manager) GetCertificates() []*x509.Certificate {
 	return certificates
 }
 
+// getStore returns the store found for storeName. If not found, a new one is
+// created, populated with a default certificate, and written to m.stores.
 func (m *Manager) getStore(storeName string) *CertificateStore {
 	_, ok := m.stores[storeName]
 	if !ok {
--- a/pkg/tls/tlsmanager.go
+++ b/pkg/tls/tlsmanager.go
@@ -93,9 +93,6 @@ func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 		tlsConfig = &tls.Config{}
 	}
 
-	store := m.getStore(storeName)
-	acmeTLSStore := m.getStore(tlsalpn01.ACMETLS1Protocol)
-
 	if err == nil {
 		tlsConfig, err = buildTLSConfig(config)
 		if err != nil {
@@ -107,6 +104,7 @@ func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 		domainToCheck := types.CanonicalDomain(clientHello.ServerName)
 
 		if isACMETLS(clientHello) {
+			acmeTLSStore := m.getStore(tlsalpn01.ACMETLS1Protocol)
 			certificate := acmeTLSStore.GetBestCertificate(clientHello)
 			if certificate == nil {
 				return nil, fmt.Errorf("no certificate for TLSALPN challenge: %s", domainToCheck)
@@ -115,6 +113,7 @@ func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 			return certificate, nil
 		}
 
+		store := m.getStore(storeName)
 		bestCertificate := store.GetBestCertificate(clientHello)
 		if bestCertificate != nil {
 			return bestCertificate, nil
--- a/pkg/tls/tlsmanager.go
+++ b/pkg/tls/tlsmanager.go
@@ -20,11 +20,11 @@ var DefaultTLSOptions = Options{}
 
 // Manager is the TLS option/store/configuration factory.
 type Manager struct {
+	lock         sync.RWMutex
 	storesConfig map[string]Store
 	stores       map[string]*CertificateStore
 	configs      map[string]Options
 	certs        []*CertAndStores
-	lock         sync.RWMutex
 }
 
 // NewManager creates a new Manager.
@@ -49,7 +49,7 @@ func (m *Manager) UpdateConfigs(ctx context.Context, stores map[string]Store, co
 	m.stores = make(map[string]*CertificateStore)
 	for storeName, storeConfig := range m.storesConfig {
 		ctxStore := log.With(ctx, log.Str(log.TLSStoreName, storeName))
-		store, err := buildCertificateStore(ctxStore, storeConfig)
+		store, err := buildCertificateStore(ctxStore, storeConfig, storeName)
 		if err != nil {
 			log.FromContext(ctxStore).Errorf("Error while creating certificate store: %v", err)
 			continue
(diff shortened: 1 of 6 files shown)


pkg/tls/tlsmanager.go
```diff
@@ -107,6 +104,7 @@ func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 		domainToCheck := types.CanonicalDomain(clientHello.ServerName)
 
 		if isACMETLS(clientHello) {
+			acmeTLSStore := m.getStore(tlsalpn01.ACMETLS1Protocol)
```
mpl, line 107: two things worry me, related to the fact that there's a mutex guarding m for the whole duration of Get.

1) if you're moving m.getStore into the GetCertificate closure, you're taking it out of the mutex protection, therefore potentially introducing a race.

2) (unrelated to your change) since the very first call to getStore potentially modifies m.stores, it might have been a mistake to use a Read lock at the beginning of Get, and it might be necessary to use a RW lock instead.

mpl (commented):
Even if this change was correct (not sure yet that it is), I don't think it would be enough to obtain the full desired effect to fix the issue. In the setup I used to test it, it indeed prevented a systematic generation of the default certificate for the acme-tls/1 store, but it did not prevent the generation of the default certificate for the default store. Because m.getStore("default") is also systematically called  somewhere else outside of m.Get().
So this change would only have half (maybe less?) of the intended effect.

pkg/tls/tlsmanager.go
mpl: And in that respect, it is also strange that the access to `m.configs[configName].SniStrict` L122 isn't guarded either.

rkojedzinszky: Sorry for that, I did not check the code carefully, so it might be wrong regarding mutexes. Also, you are right, only in case of acme store does it provide a solution. The default store's issue can be solved externally, providing a default certificate, but acme's cannot. At least, I did not find a way for it.

rkojedzinszky: Would it be enough to lock the mutex inside the closure function?

rkojedzinszky: According to your comments and with my changes, an RLock() may be enough at the beginning of Manager.Get(), and in the  tlsConfig.GetCertificate function an RW Lock should be obtained. Am I right?

mpl: Adding an RWLock within GetCertificate sounds scary and a recipe for deadlocks at first impression, but maybe. But hold on please, I want to understand first 1) why there needs to be a default certificate in the acme store, 2) whether, just like for the default store, we could simply let the configuration provide the default cert for the acme store, and 3) why we only have an RLock (instead of an RWLock) at the beginning of Get.

mpl:
@rkojedzinszky I pushed an alternative version to your PR. WDYT?

rkojedzinszky:
> @rkojedzinszky I pushed an alternative version to your PR. WDYT?

Sorry, I did not have time to review changes, and I am also not familiar with this layer. If it avoids regenerating certificates each time on configuration change, then it is great!

mpl:
> > @rkojedzinszky I pushed an alternative version to your PR. WDYT?
> 
> Sorry, I did not have time to review changes, and I am also not familiar with this layer. If it avoids regenerating certificates each time on configuration change, then it is great!

Well ok, but then could you please try that last version of the PR out, and let us know if it actually fixes your CPU usage issue?

rkojedzinszky:
Is this PR still relevant? Shall I close it and let us find a correct solution for the issue?

rkojedzinszky:
> > > @rkojedzinszky I pushed an alternative version to your PR. WDYT?
> > 
> > 
> > Sorry, I did not have time to review changes, and I am also not familiar with this layer. If it avoids regenerating certificates each time on configuration change, then it is great!
> 
> Well ok, but then could you please try that last version of the PR out, and let us know if it actually fixes your CPU usage issue?

Do I still need to confirm the changes, or has it been done by others?

mpl:
> > > > @rkojedzinszky I pushed an alternative version to your PR. WDYT?
> > > 
> > > 
> > > Sorry, I did not have time to review changes, and I am also not familiar with this layer. If it avoids regenerating certificates each time on configuration change, then it is great!
> > 
> > 
> > Well ok, but then could you please try that last version of the PR out, and let us know if it actually fixes your CPU usage issue?
> 
> Do I still need to confirm the changes, or has it been done by others?

Yes please, it would be great if you could tell us if the PR in its actual state solves the problem for you.

rkojedzinszky:
> > > > > @rkojedzinszky I pushed an alternative version to your PR. WDYT?
> > > > 
> > > > 
> > > > Sorry, I did not have time to review changes, and I am also not familiar with this layer. If it avoids regenerating certificates each time on configuration change, then it is great!
> > > 
> > > 
> > > Well ok, but then could you please try that last version of the PR out, and let us know if it actually fixes your CPU usage issue?
> > 
> > 
> > Do I still need to confirm the changes, or has it been done by others?
> 
> Yes please, it would be great if you could tell us if the PR in its actual state solves the problem for you.

After a quick check, it seems that cpu usage is at the same level as with v2.3, so I think the optimization works.

mpl (approved):

dtomcej (approved):
LGTM
:shipit:

rtribotte (approved):
LGTM

PR #7860, opened 2021-02-04
Fix the static reference documentation for the internal redirection router.

jbdoumenjou:
### What does this PR do?

Update the default priority value of the internal redirection router in static reference documentation.

### Motivation

Having documentation up to date. :nerd_face: 

### More

- [ ] ~~Added/updated tests~~


ldez (approved):
LGTM

Issue #3097, opened 2018-03-28
Feature Request: A way to restore client's real ip from a specified header when behind a proxy


basro:
### Do you want to request a *feature* or report a *bug*?

feature

### What did you do?

I wanted to use traefik behind cloudflare's proxy but found no way to restore the client's real ip in a secure way.

### What did you expect to see?

I expected some form of middleware similar to [nginx's real ip module](http://nginx.org/en/docs/http/ngx_http_realip_module.html)

The feature would allow you to select a header from which to read the real ip and a set of ip ranges that are trusted.

It should affect middleware like the rate limiting, so that rate limits by client ip use the ip obtained from the header.

softyn:
Does rate limiting functionality use real ip (from X- header) or proxy ip in that case?

zeph:
I follow... I was hunting for such all through the documentation until I decided to checkout the code

then landed here

zeph:
this is where it gets copied over... but there is no code to copy the left-most IP in the x-forwarded-for
```c
func (x *XForwarded) rewrite(outreq *http.Request) {
	if clientIP, _, err := net.SplitHostPort(outreq.RemoteAddr); err == nil {
		clientIP = removeIPv6Zone(clientIP)

		if unsafeHeader(outreq.Header).Get(xRealIP) == "" {
			unsafeHeader(outreq.Header).Set(xRealIP, clientIP)
		}
	}
```
...there's the logic from the ipFiltering, with the excludeIp thing... that might be leveraged

zeph:
```c
	err := wl.whiteLister.IsAuthorized(wl.strategy.GetIP(req))
```
the GetIP() could be leveraged, but it requires the middleware logic & configuration... so not an option

either we copy the whole logic to be used in the static configuration in association with an entryPoint
...else I wouldn't know how to... I mean, they already have the trustIP list... that can be used as an exclusion list 
as it also is already related/enabling the trusting of the x-forwarded-for headers

zeph:
funny thing is that the complete logic around the "client IP" seems inconsistent... for instance, for logging, they take either the req.RemoteAddr as they do for the ipWhitelisting if u do not provide either a depth or a excludeIPs parameter... or, they fully trust the forwarded for ... with no pruning of known IPs... 
```c
	core[ClientHost], core[ClientPort] = silentSplitHostPort(req.RemoteAddr)

	if forwardedFor := req.Header.Get("X-Forwarded-For"); forwardedFor != "" {
[...]

PR #7875, opened 2021-02-08
Use X-Forwarded-For header to define X-Real-IP header

zeph:
### What does this PR do?

It's implementing a mechanism similar to nginx to determine the x-real-ip by excluding proxy IPs

### Motivation

I stumbled on this issue while reviewing my freshly installed traefik v2, 
noticing that the x-real-ip header was not properly populated... #3097 
(I referred the feature request opened by @basro 2 years ago)

### More

I added a couple of test, the 1st to validate the feature
...the 2nd to avoid to break old behaviour

### Additional Notes

Also the logging shall be policed. Right not the `ClientAddr` is 
taken from the `X-Forwarded-For` just because it is present
...the proposed `X-Real-IP` would be a valid substitute

### DIFF
--- a/docs/content/routing/entrypoints.md
+++ b/docs/content/routing/entrypoints.md
@@ -293,6 +293,8 @@ You can configure Traefik to trust the forwarded headers information (`X-Forward
     --entryPoints.web.forwardedHeaders.trustedIPs=127.0.0.1/32,192.168.1.7
     ```
 
+!!! important "`forwardedHeaders.trustedIPs` will also enable the parsing of the header `X-Forwarded-For` (if present) to determine the `X-Real-IP` by the same (reading from right to left) logic applied for `ipStrategy.excludedIps` within the Middlewares"
+
 ??? info "`forwardedHeaders.insecure`"
     
     Insecure Mode (Always Trusting Forwarded Headers).
--- a/pkg/ip/strategy.go
+++ b/pkg/ip/strategy.go
@@ -20,8 +20,6 @@ type RemoteAddrStrategy struct{}
 
 // GetIP returns the selected IP.
 func (s *RemoteAddrStrategy) GetIP(req *http.Request) string {
-	// FIXME: I would use req.Header.Get(xRealIP) instead
-	// as related to #3097 I would properly try to populate it
 	ip, _, err := net.SplitHostPort(req.RemoteAddr)
 	if err != nil {
 		return req.RemoteAddr
--- a/pkg/ip/strategy.go
+++ b/pkg/ip/strategy.go
@@ -20,6 +20,8 @@ type RemoteAddrStrategy struct{}
 
 // GetIP returns the selected IP.
 func (s *RemoteAddrStrategy) GetIP(req *http.Request) string {
+	// FIXME: I would use req.Header.Get(xRealIP) instead
+	// as related to #3097 I would properly try to populate it
 	ip, _, err := net.SplitHostPort(req.RemoteAddr)
 	if err != nil {
 		return req.RemoteAddr
--- a/pkg/middlewares/forwardedheaders/forwarded_header.go
+++ b/pkg/middlewares/forwardedheaders/forwarded_header.go
@@ -137,7 +137,11 @@ func (x *XForwarded) rewrite(outreq *http.Request) {
 
 		if x.ipChecker != nil {
 			strategy := ip.CheckerStrategy{Checker: x.ipChecker}
-			clientIP = strategy.GetIP(outreq)
+			xClientIP := strategy.GetIP(outreq)
+			if xClientIP != "" {
+				// X-Forwarded-For wasn't empty
+				clientIP = xClientIP
+			}
 		}
 
 		if unsafeHeader(outreq.Header).Get(xRealIP) == "" {
--- a/pkg/middlewares/forwardedheaders/forwarded_header.go
+++ b/pkg/middlewares/forwardedheaders/forwarded_header.go
@@ -135,6 +135,11 @@ func (x *XForwarded) rewrite(outreq *http.Request) {
 	if clientIP, _, err := net.SplitHostPort(outreq.RemoteAddr); err == nil {
 		clientIP = removeIPv6Zone(clientIP)
 
+		if x.ipChecker != nil {
+			strategy := ip.CheckerStrategy{Checker: x.ipChecker}
+			clientIP = strategy.GetIP(outreq)
+		}
+
 		if unsafeHeader(outreq.Header).Get(xRealIP) == "" {
 			unsafeHeader(outreq.Header).Set(xRealIP, clientIP)
 		}
--- a/pkg/middlewares/forwardedheaders/forwarded_header_test.go
+++ b/pkg/middlewares/forwardedheaders/forwarded_header_test.go
@@ -178,6 +178,15 @@ func TestServeHTTP(t *testing.T) {
 			},
 		},
 		{
+			desc:       "xRealIP populated from remote address, but not tricked by Trusted Ips",
+			insecure:   false,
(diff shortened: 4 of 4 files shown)


zeph:
p.s. it was already failing out of the master I branched off... happy to rebase as soon as u give me a heads up

zeph:
fails on origin/v2.4
```golang
FAIL: grpc_test.go:365: GRPCSuite.TestGRPCBufferWithFlushInterval

grpc_test.go:423:
    c.Assert(err, check.IsNil)
... value *fmt.wrapError = &fmt.wrapError{msg:"try operation failed: failed to receive stream data", err:(*errors.errorString)(0xc000e6e980)} ("try operation failed: failed to receive stream data")
```

zeph:
failure on `master`

```golang
FAIL: docker_test.go:184: DockerSuite.TestDockerContainersWithLabels

docker_test.go:221:
    c.Assert(err, checker.IsNil)
... value *fmt.wrapError = &fmt.wrapError{msg:"try operation failed: got status code 404, wanted 200", err:(*errors.errorString)(0xc000951140)} ("try operation failed: got status code 404, wanted 200")
```

ldez:
I guess that you are trying to run the integration tests locally.
To do that:
- first: `make pull-images` (only once)
- then `CI=1 make test-integration`

zeph:
@ldez I'm indeed... since I wanna see my PR passing xD
...trying with an empty image set, lets see... `CI=1` this time

zeph:
ok, all green even on my workstation... 
thanks for the heads-up @ldez & for kicking the CI process again :)

jbdoumenjou:
Hello,
 
Thanks for your contribution!

zeph:
@jbdoumenjou I stepped on this issue this morning https://github.com/traefik/traefik/issues/7884
...eventually this `X-Real-IP` properly populated could be the new reference for the Middleware `IpWhitelist`
and make https://doc.traefik.io/traefik/middlewares/ipwhitelist/#ipstrategyexcludedips obsolete

zeph:
@jbdoumenjou I'll wait ur heads-up before doing a rebase, seems that the merge is not possible automatically anymore

made2591:
Any update on this?

jbdoumenjou:
Hello @zeph,

We had a first look and want to clarify the purpose of the PR.
It seems that the addressed case here is when Traefik trusts the previous proxy that provides a `X-Forwarded-For` header but no `X-Real-Ip` header.
Currently, when the `X-Real-Ip` header is not set, Traefik creates it using the remote IP address.

This PR uses the `X-Forwarded-For` as a source to populate the `X-Real-Ip`, as it should contain the chain of previous client IPs.
Looking at the implementation, it uses the `forwardedHeader.trustedIPs` list to ignore all trusted IPs in the `X-Forwarded-For` and apply the first "non trusted" IP as `X-Real-Ip` header value.
This `trustedIPs` option is currently used to define the IPs (the request remote address) from which the forwarded headers will be trusted, so not deleted and used.

Why using this option, which is not expected, to filter out the IPs from the `X-Forwarded-For` to find the `X-Real-Ip` value? It does not seem to be appropriate.

Moreover, if the previous proxy is trusted, the `X-Forwarded-For` header should be trusted too and the `X-Real-Ip` value should be the first IP of the `X-Forwarded-For`.

WDYT?

zhiguangwang:
@jbdoumenjou This PR is useful when you put traefik behind an AWS ALB (Application Load Balancer).

In this case, the request flows like this:

```
Client => ALB => Traefik => Application
```

Because [ALB appends client IP to X-Forwarded-For](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/x-forwarded-headers.html#x-forwarded-for), when the request reaches the application, the HTTP headers becomes:

```
X-Forwarded-For: <client_ip>, <alb_ip>
X-Real-Ip: <alb_ip>
```

Further more, the client can forge `X-Forwarded-For`, so the HTTP headers becomes:

```
X-Forwarded-For: <fake_client_ip> <client_ip>, <alb_ip>
X-Real-Ip: <alb_ip>
```

But what we are really expecting is:

```
X-Forwarded-For: <client_ip>, <alb_ip>
X-Real-Ip: <client_ip>
```

jbdoumenjou:
Hello,

Thanks to this PR, we realized that we have several concerns about the X-Real-Ip expected behavior.
Moreover, this implementation will not be backward compatible as it will change the current behavior.

So we will close this PR and continue the discussion in the original issue #3097

zeph:
@jbdoumenjou can you make some examples of configurations that are going to break if introducing this "new behaviour", please?

zeph:
@zhiguangwang with my patch the `<alb_ip>` is going to be removed from the list if it has been passed in the "trusted proxies" list
-> it is not going to be presented as "<real ip>" as in your example

zeph:
@zhiguangwang to clarify... ok, in my infrastructure I have 2 reverse proxies in front of my traefik ... I remove from the game the 1st set with the `trustIPs` as main cluster configuration... and the 2nd instead is addressed when doing real IP filtering with this middleware 

```
$ cat templates/ipwhitelist.yml 
kind: Middleware
metadata:
  name: geo
spec:
  ipWhiteList:
    sourceRange:
 [ ... trusted 3rd parties ... ]

    ipStrategy:
      # https://doc.traefik.io/traefik/middlewares/ipwhitelist/#ipstrategyexcludedips
      excludedIPs:
        - 172.31.7.2/32 # 217.115.64.70,.71,.72 SNAT
        # https://www.cloudflare.com/ips-v4
        - 103.21.244.0/22
        - 103.22.200.0/22
        - 103.31.4.0/22
        - 104.16.0.0/12
        - 108.162.192.0/18
        - 131.0.72.0/22
        - 141.101.64.0/18
        - 162.158.0.0/15
        - 172.64.0.0/13
        - 173.245.48.0/20
        - 188.114.96.0/20
        - 190.93.240.0/20
        - 197.234.240.0/22
        - 198.41.128.0/17
        # https://www.cloudflare.com/ips-v6
        - 2400:cb00::/32
        - 2606:4700::/32
        - 2803:f800::/32
        - 2405:b500::/32
        - 2405:8100::/32
        - 2c0f:f248::/32
        - 2a06:98c0::/29
```

in your case... you need to exclude the ip ranges declared by AWS as the subnets for your `<alb ip>`

zeph:
@zhiguangwang I pass the trustedIPs with this... in this case the IP of the 1st reverse proxy
```
    # https://community.traefik.io/t/how-to-properly-use-x-forwarded-for-in-v2/3834
    - "--entrypoints.websecure.forwardedHeaders.trustedIPs=x.x.x.x/32"
    - "--entrypoints.web.forwardedHeaders.trustedIPs=x.x.x.x/32"

```

jbdoumenjou:
@Zeph,

The [trustedIPs](https://doc.traefik.io/traefik/routing/entrypoints/#forwarded-headers) is a way to trust the forwarded headers sent by a Client by checking its address. Therefore, It could be defined to trust only specific proxy/servers.

Using the trustedIPs option to set the X-Real-IP is a workaround and could impact the other forwarded headers as Traefik could trust unwanted servers. For example, as soon as one trusted proxy reaches Traefik and does not provide the X-Real-IP header, the X-Real-IP value would be then different from the proxy IP. We understand that this is the behavior you were expecting, thus it is a breaking change as the X-Real-IP value will change without modifying the existing configuration.

As said in the original issue (#3097), the current implementation is naive/simple but there is no real specification to oppose to it that would define the expected behavior.

In other words, we want to keep the contract for X-Real-IP header identical as for forwarded headers, the chain of trust is enforced by trusting or not the client which is reaching Traefik, and as for now we think that more specific/custom behaviors/capabilities should be handled with plugins and not by Traefik itself.

PR #7890, opened 2021-02-12
Allow overriding the service name in consul.

apollo13:
### What does this PR do?

Consul allows to register multiple instances behind a single service
name with different tags. For instance it is possible to register the
service  `backend-api` twice with the following tags:

 * ["traefik.enable=true", "dev"]
 * ["traefik.enable=true", "prod"]

Without this patch, traefik would generate a single service and consider
all those instances the same. After this patch it is possible to
reconfigure the services like this:

 * ["traefik.enable=true", "traefik.service=backend-api-dev", "dev"]
 * ["traefik.enable=true", "prod"]

And traefik would be able to distinguish those two services & route
accordingly.

### More

- [x] Added/updated tests
- [x] Added/updated documentation -- not 100% sure where to add more documentation for the "traefik.service" label or if it should get renamed to something else.

### DIFF
--- a/docs/content/reference/dynamic-configuration/consul-catalog.yml
+++ b/docs/content/reference/dynamic-configuration/consul-catalog.yml
@@ -1,1 +1,2 @@
 - "traefik.enable=true"
+- "traefik.service=service-name"
--- a/integration/consul_catalog_test.go
+++ b/integration/consul_catalog_test.go
@@ -150,6 +150,70 @@ func (s *ConsulCatalogSuite) TestWithNotExposedByDefaultAndDefaultsSettings(c *c
 	c.Assert(err, checker.IsNil)
 }
 
+func (s *ConsulCatalogSuite) TestWithNotExposedByDefaultAndMultipleServicesBehindTags(c *check.C) {
+	reg1 := &api.AgentServiceRegistration{
+		ID:      "whoami1",
+		Name:    "whoami",
+		Tags:    []string{"whoami1", "traefik.enable=true"},
+		Port:    80,
+		Address: s.composeProject.Container(c, "whoami1").NetworkSettings.IPAddress,
+	}
+	err := s.registerService(reg1, false)
+	c.Assert(err, checker.IsNil)
+
+	reg2 := &api.AgentServiceRegistration{
+		ID:      "whoami2",
+		Name:    "whoami",
+		Tags:    []string{"whoami2", "traefik.enable=true", "traefik.service=someone-else"},
+		Port:    80,
+		Address: s.composeProject.Container(c, "whoami2").NetworkSettings.IPAddress,
+	}
+	err = s.registerService(reg2, false)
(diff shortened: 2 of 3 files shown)


jbdoumenjou:
Hello,
 
Thanks for your contribution!

rtribotte:
Hello @apollo13,

Thanks for your contribution and your time.

After giving it more thought with the team, unfortunately, we will decline your pull request.

Currently, naming services is already possible for all label providers. For instance, by configuring the scheme: 
`traefik.http.services.{MyService}.loadbalancer.server.scheme: "http"`

Also, while we are not willing to introduce a new behavior to name services only for the Consul Catalog provider, a strong reason to not allow the naming through a specific label is that it will not be possible to declare two services for the same container/service and be able to name them consistently (when using such label).

(We also declined this service naming feature for the ECS provider as you can see in #7395)

apollo13:
Hi @rtribotte, can we talk about this a little bit:

> Currently, naming services is already possible for all label providers. For instance, by configuring the scheme:
traefik.http.services.{MyService}.loadbalancer.server.scheme: "http"

This is true for HTTP, what would you recommend to override for UDP & TCP? I cannot override the port for UDP since I generally do not know it (that is the whole point of using service discover)

> Also, while we are not willing to introduce a new behavior to name services only for the Consul Catalog provider

Fair, it would make more sense to provide a global mechanism to override this.

> a strong reason to not allow the naming through a specific label is that it will not be possible to declare two services for the same container/service and be able to name them consistently (when using such label).

I am not sure I follow. Doesn't the PR exactly do that?

rtribotte:
> This is true for HTTP, what would you recommend to override for UDP & TCP? I cannot override the port for UDP since I generally do not know it is (that is the whole point of using service discover) -- and that seems to be the only option.

Well, redefining the `terminationdelay` using the default value (100) should do the trick for TCP services, but for UDP services I have to admit that there is no workaround other than defining the port.

> I am not sure I follow. Doesn't the PR exactly do that?

The point is that you cannot define multiple services with a label like `traefik.service: "myService"'.
Today you can create two service and be consistent :
```
traefik.http.services.service01.loadbalancer.server.scheme: "http"
traefik.http.services.service02.loadbalancer.server.scheme: "http"
```
With your proposal, it's not possible to name two services, and we would have to converge consistently with configuration such as:
```
traefik.service: "service01"
traefik.http.services.service02.loadbalancer.server.scheme: "http"
```

I hope that it will better explain the current limitations we are foreseeing regarding the introduction of a such label.

Thus, if you feel that it needs to be debated, please, feel free to open an issue that would be the place to discuss your proposal.

apollo13:
@rtribotte Thank you for explaining that. It is more understandable now. I also fully understand that you dislike `traefik.service` -- it was simply the best I could come up with to demonstrate the issue. I'll open a new issue with and see if we can come up with a better idea :)

Issue #7897, opened 2021-02-15
improve basic auth performance by caching expensive password hashing computations 


wbolster:
### tl;dr:

using http basic auth with high bcrypt complexity results in high cpu load and slow response times; caching may alleviate this.

### Do you want to request a *feature* or report a *bug*?

a feature (or a performance bug, depending on how you look at it)

### What were you trying to do?

i want to use basic auth (over tls) with strong bcrypt complexity as a simple but sensible security setup for things like the traefik dashboard.

### What did you expect to see?

i expected it to work, and it does…

### What did you actually see?

…but it slows down _every request_. merely having the dashboard open in a browser results in periodic polls to the `/api` endpoints, and each request is slow and eats CPU resources at the server.

### What do you think can be done about it?

this slowness seems unnecessary. similar things have been reported before (e.g. #4791) but it seems it did not get the proper explanation/attention.

each request will have a valid user/pass in normal situations, since that is how basic auth operates. in practice, the basic auth middleware invokes `goauth.CheckSecret` _for every request_ ([source code](https://github.com/traefik/traefik/blob/master/pkg/middlewares/auth/basic_auth.go#L68)).

the main problem here is that the code performs a (potentially very slow) bcrypt operation too often: it is called for _every request_ by the same user with the _exact same inputs_, giving the _exact same results_ (`bcrypt` is a pure operation). this is simply wasteful, resulting in slow responses and high cpu usage.

my suggestion would be to use a small in-memory (!) caching mechanism to track recent _valid_ authentication attempts. for the expected use cases such as traefik dashboard + basic auth + strong bcrypt + small amount of users, this will likely avoid 99%+ of all slow operations.

the cache would contain a mapping from a key `(user, pass, realm)` to any value. presence of a key in the cache means it was validated correctly, absence means a real bcrypt check has to be done (which would result in `.Add()` when successful). the thread-safe `Cache` implementation in the `golang-lru` package ([source](https://github.com/hashicorp/golang-lru), [docs](https://pkg.go.dev/github.com/hashicorp/golang-lru)) may be helpful here.

note that there seem no additional security implications here: secrets are only ever kept in memory, and they are already in memory to begin with. but in practice it may be advisable to use a cheap(er) hash algorithm (e.g. sha512) to hash the user/pass/realm combination, and use that as the map key for the cache.

wbolster:
(copying relevant part of the #7900 description to keep the discussion centralized)

### What does this PR do?

[...]

PR #7900, opened 2021-02-16
Cache password hashing results in basicAuth middleware

wbolster:
### What does this PR do?

See #7897.

This introduces a small in-memory caching mechanism to the BasicAuth
middleware to improve response times and reduce CPU usage.

### Motivation

While HTTP Basic Authentication is not widely used for externally facing
services in the real world (for multiple reasons), it is a reasonably
secure and simple way to setup access control for things like
internal-facing dashboard applications, such as the Traefik
API/dashboard. In practice, dashboard tools (e.g. Traefik's own
dashboard) continuously perform HTTP requests to update themselves. When
using a strong(er) hashing algorithm, merely having a dashboard open in
a web browser can cause significant server CPU load and high latency… 😿

HTTP Basic Authentication requires embedding a user/password combination
in every request. The BasicAuth middleware verifies every request, which
involves password hashing, and this can be quite slow (see below). The
good news is that in the common case, the password hashing function is
repeatedly called with identical inputs. Since this is a pure
function (no side effects, no randomness involved), the results can be
cached, which helps tremendously for common use cases such as repeated
requests from the same logged in user (each time with the same user/password
combination in the HTTP request).

### Implementation details

- Use a small in-memory LRU Cache from golang-lru, which was already an
  indirect dependency.

- Cache the result of goauth.CheckSecret() calls. Note that this caches
  both positive and negative results, which covers both the ‘everything
  fine’ and the ‘misconfigured cron job continuously trying to log in
  with the wrong credentials’ cases.

- Use a SHA512 hash (unrelated to password hashing operation) over the
  realm/user/password combination to produce a cache key. This
  guarantees that plaintext secrets (such as the password in the request
  headers) are not kept in memory longer than necessary. (This depends
  also on garbage collection, reused memory, etc.)

- Note that the cache size is hardcoded. It uses a small amount of
  memory, while at the same time it is very effective for the ‘handful
  of users accessing an internal dashboard’ use case. Exposing the cache
  size as a configuration value would be technically possible, but the
  usefulness is questionable (and has implications for docs,
  maintenance, upgrades, etc.)

- Adapt the tests to perform multiple requests to hit both the hit/miss
  scenarios.

### More

As an extra, some more performance numbers: as an extreme example,
bcrypt with a very high complexity can take multiple seconds to compute.
Some example timings (in seconds) from a modern Intel i7 laptop CPU:

    $ for c in $(seq 4 17); do \
        time=$(which time); \
        t=$($time htpasswd -nBb -C$c test test \
          2>&1 >/dev/null \
          | awk '{print $1}' \
          | head -n1); \
        printf "complexity %s: %s\n" $c $t; \
      done
    complexity 4: 0.00user
    complexity 5: 0.00user
    complexity 6: 0.00user
    complexity 7: 0.00user
    complexity 8: 0.01user
    complexity 9: 0.02user
    complexity 10: 0.04user
    complexity 11: 0.08user
    complexity 12: 0.17user
    complexity 13: 0.35user
    complexity 14: 0.68user
    complexity 15: 1.37user
    complexity 16: 2.71user
    complexity 17: 5.42user

Substituting a value like this one in the unit tests (instead of a
simpler hashing mechanism) highlights the problem quite visibly:

    $ htpasswd -nBb -C17 test test
    test:$2y$17$DvxbL0Tk9IyfIgWyLF84qOILEEDmNt40HgZWIr/YzTSdnBN0A9sNK

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -40,6 +40,7 @@ require (
 	github.com/hashicorp/consul/api v1.3.0
 	github.com/hashicorp/go-multierror v1.0.0
 	github.com/hashicorp/go-version v1.2.0
+	github.com/hashicorp/golang-lru v0.5.4
 	github.com/influxdata/influxdb1-client v0.0.0-20191209144304-8bf82d3c094d
 	github.com/instana/go-sensor v1.5.1
 	github.com/libkermit/compose v0.0.0-20171122111507-c04e39c026ad
--- a/pkg/middlewares/auth/basic_auth.go
+++ b/pkg/middlewares/auth/basic_auth.go
@@ -2,12 +2,14 @@ package auth
 
 import (
 	"context"
+	"crypto/sha512"
 	"fmt"
 	"net/http"
 	"net/url"
 	"strings"
 
 	goauth "github.com/abbot/go-http-auth"
+	lru "github.com/hashicorp/golang-lru"
 	"github.com/opentracing/opentracing-go/ext"
 	"github.com/traefik/traefik/v2/pkg/config/dynamic"
 	"github.com/traefik/traefik/v2/pkg/log"
@@ -27,6 +29,7 @@ type basicAuth struct {
 	headerField  string
 	removeHeader bool
 	name         string
+	cache        *lru.Cache
 }
 
 // NewBasic creates a basicAuth middleware.
@@ -37,12 +40,18 @@ func NewBasic(ctx context.Context, next http.Handler, authConfig dynamic.BasicAu
 		return nil, err
 	}
 
+	cache, err := lru.New(128)
+	if err != nil {
+		return nil, err
+	}
+
 	ba := &basicAuth{
 		next:         next,
 		users:        users,
 		headerField:  authConfig.HeaderField,
 		removeHeader: authConfig.RemoveHeader,
 		name:         name,
+		cache:        cache,
 	}
 
 	realm := defaultRealm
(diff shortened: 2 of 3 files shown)


wbolster:
cc @SantoDE b/c assigned to #7897

SantoDE:
Hello @wbolster,

thanks for your contribution.

Please note, we have marked that PR as `needs-design-review`. As stated in our [contribution guide](https://github.com/traefik/contributors-guide/blob/master/review_guidelines.md), that indicates that we will take a close and extensive look on that PR now.

This will most likely require some time, so we wanted to inform you beforehand.

We are not yet sure about potential impact to Traefik and what it might mean from a security point of view, so please bear with us for a bit :)

wbolster:
🤗 no worries, take your time! none of this is urgent.

SantoDE:
Hello @wbolster,

once again, thanks for your interest in PR and the PR.

We have had a lengthy discussion internally and came to the conclusion, that for now, we want to decline that PR. We have a lot of (general) questions unsolved, which we should discuss maybe in the proposal before.

- What would be the strategy to revoke users? (the cache lib you used is basically a fifo stack configured with 128 slots)
- A fifo stack won’t be beneficial in any way if it’s always full of unsuccessful attempts because somebody is trying to brut force the auth
- What is the impact on performances, making the key to store in cache (sha512) versus computing bcrypt each times
- What are the most common scenario of complexity used with bcrypt encoded passwords?
- In what cases does that PR improve performance?
- What is the benefit for high frequency configuration reload context

We are also unsure, if that would not make Traefik sensitive to time pad attacks, which is something we want to avoid off course.

For all that reasons, we want to decline the PR for now, with a strong emphasis on for now. We can discuss the points in the proposal and once we reach an agreement, we are fine in accepting a PR. 

We are just not confident right now.

wbolster:
i guess it's more practical to continue the discussion in the open ticket #7897 instead of in this closed pull request, so i'll respond there!

wbolster:
fyi: i have copied most relevant pieces of info to #7897

PR #7904, opened 2021-02-17
Only allow iframes to be loaded from our domain

SantoDE:
### What does this PR do?

Only allows iframes to be loaded from our domain

### Motivation

Be more secure :)

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/api/dashboard.go
+++ b/pkg/api/dashboard.go
@@ -35,6 +35,7 @@ func (g DashboardHandler) Append(router *mux.Router) {
 
 func (g DashboardHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// allow iframes from our domains only
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src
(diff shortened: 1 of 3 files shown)


pkg/api/dashboard.go
```diff
@@ -33,6 +33,12 @@ func (g DashboardHandler) Append(router *mux.Router) {
 		Handler(http.StripPrefix("/dashboard/", http.FileServer(g.Assets)))
 }
 
+func (g DashboardHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// allow iframes from our domains only
+	w.Header().Set("Content-Security-Policy", "frame-ancestors 'self' traefik.io *.traefik.io;")
```
dtomcej, line 38: Is this something that we could configure with a default? I'm just thinking worst case scenario, if something happened to those records, users could temporarily override with a CLI arg, rather than having to use a rebuilt binary/image.

rtribotte (approved):
LGTM 👌

mpl (approved):

ldez (approved):
LGTM

Issue #7592, opened 2020-11-25
Exclude ips for rate limit


myclau:
I am working with rate limit for blocking the robotic access.

traefik version is 2.2.8

I can easily setup the rate limit by using annotation `traefik.ingress.kubernetes.io/rate-limit`

Our endpoint can be access with anyone and any device, but we also have users like shopping mall will use it.
As these users they will provide  public wifi to the customer to access our end point and we cannot block them. so it is possible to mis-block them as many people will share the same ip from wifi to access our end point.

But there is no annotation or setting I can exclude some known ips (these users will give us their public ips) for the rate-limit checking.

Something similar is `traefik.ingress.kubernetes.io/whitelist-source-range` but this annotation only allow the whitelist ip to access which is not I need. I need to block robotic access and allow all normal access or known users access (public wifi).

How can I using the rate limit and can exclude some known ip?

SantoDE:
Related to #7138

sebschlue:
I would like to vote for this issue.

I also need to exclude some IPs from being rate limited.

Issue #7138, opened 2020-08-03
Origin IP as Router rule matcher


stevenkaras:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

I'd like to be able to route requests based on source IP.
To some extent, this is already done via middlewares such as the IPWhitelist, which when used in combination with the Errors middleware can be used to guide requests somewhat, however this can be somewhat subpar in UX.

Ideally, this would support the following use cases:

1. Requiring basic auth when accessing a path from outside an IP range (I'm aware of two workarounds to support this already, but this makes it more explicit and less surprising for new users)
2. Routing requests from non-residential IPs towards a subset of servers (to avoid a misconfigured bot from denying service to actual human users)
3. Routing request from external IPs through extra logging middleware.

I want to propose two different syntaxes which are not mutually exclusive. The first one is preferable, but would require much more work to get working as it adds yet another top level element to configurations:

#### list-aware syntax

This syntax would be aware of IP lists (ideally the ipwhitelist middleware would be improved to also use this).
This allows there to be a single source of truth reused across similar components:

```yaml
sourceRanges:
  internalIPs:
    - 192.168.0.0/16
    - 172.16.0.0/12
    - 10.0.0.0/8
  anotherExample:
    - 192.0.2.0/24

...

http:
  middlewares:
    internal-restricted:
      ipWhiteList:
        sourceRangeList: anotherExample

  routers:
    internal-only:
      rule: "OriginList(internalIPs) ..."
```

#### CIDR-syntax

This syntax would simplify the current workaround used by providing a syntax that is CIDR based rather than regexp based:

```yaml
http:
  routers:
    internal-only:
      rule: "OriginCIDR(192.168.0.0/16) || OriginCIDR(10.0.0.0/8) ..."
```

jspdown:
@stevenkaras Thanks for your proposal. I can see clear needs for this kind of criteria. The use cases I'm thinking of are:
- Having different middleware chains following the requester. Such as higher rate limits for internal IPs or different authentication methods.
- Forwarding requests to different services.
- Having different router rules for different requesters. e.g. ACL

In some cases we can leverage the `errors` middleware to forward the unauthorized request to a different service. But it works only if you want to split requesters in two groups and send them to different services.

[...]

PR #7911, opened 2021-02-18
Add IPs exclusion for middleware rate limiting

mathcantin:
### What does this PR do?

Add in the middleware rate limiting a IPs whitelist to exclude.

The yaml syntaxe for that:

```
http:
  middlewares:
    test-ratelimit:
      rateLimit:
        average: 100
        exclusion:
          sourceRange:
            - 127.0.0.1/32
            - 192.168.1.7
```

### Motivation

Feature request : #7592 

I think this is normal when we add middleware to want some exceptions. There are a way to do that in nginx and haproxy.

### More

- [X] Added/updated tests

### Additional Notes

The `exclusion.ipStrategy` config isn't inherited from the `sourceCriterion.ipStrategy` because they may have a use case to use the exclusion with the `sourceCriterion.requestHost` config and probably to set two different strategies.

### DIFF
--- a/docs/content/middlewares/ratelimit.md
+++ b/docs/content/middlewares/ratelimit.md
@@ -493,3 +493,74 @@ http:
         sourceCriterion:
           requestHost: true
 ```
+
+### `exclusion`
+
+The `exclusion` option defines a whitelist of IPs to exclude from the rate limiting.
+
+#### `exclusion.sourceRange`
+
+The `sourceRange` option sets the whitelisted IPs (or ranges of whitelisted IPs by using CIDR notation).
+
+```yaml tab="Docker"
+labels:
+  - "traefik.http.middlewares.test-ratelimit.ratelimit.average=100"
+  - "traefik.http.middlewares.test-ratelimit.ratelimit.exclusion.sourceRange=127.0.0.1/32, 192.168.1.7"
+```
+
+```yaml tab="Kubernetes"
+apiVersion: traefik.containo.us/v1alpha1
+kind: Middleware
+metadata:
+  name: test-ratelimit
+spec:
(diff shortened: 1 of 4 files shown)


mathcantin:
It's possible to relaunch semaphoreci build? I don't think that is related to me.

rtribotte:
Hello @mathcantin,

Thanks for your contribution and interest in Traefik!

Unfortunately, I'll decline your PR and I apologize for the misleading open proposal you have linked in your PR.

The proposal label is to declare an issue as a potential enhancement that needs to be discussed first, to gather thoughts and opinions, decide whether we want it or not, and ultimately decide how to do it.

While we definitely see use cases for which the IP exclusion would be needed, we think that is not the best way to handle those cases.

One reason is that the same need could be raised for other middlewares, and then would mean that the IP exclusion should be added to every middleware that will need to have this feature.
But this also can drive unhandy if somehow one needs to exclude requests in multiple middlewares in a chain base on the origin IP, having to configure the exclusion on every middleware.

Moreover, the origin IP is only one criterion among valid others to bypass a middleware, so why not adding others, like headers.

This led us to think that the origin IP is more a routing criterion (#7138) and bypassing a middleware for a set of IPs should be done by setting up an appropriate router and middleware chain.

Regarding the existing IpWhitelist middleware, the difference between routing and aborting a request if not allowed (the origin IP) is the error. No routes for one IP would end by a 404 when an unauthorized IP blocked the IPWihitelist middleware would end by a 403.

As a side note, there is a potential dirty workaround if the rate limit middleware is at the end of the middleware chain.
By using the Errors middleware followed by the IPWhitelist middleware, one could fallback to a given service if the IP is not allowed.

PR #7918, opened 2021-02-20
Checkversion 7878 master

sbathe:
### What does this PR do?

Make sure that "checkNewVersion" respects what is set in the configuration file and is not overridden in the global configurations

### Motivation

To make sure the container does not make unexpected outbound connections. Plus this is a bug, if we give a say a parameter does this, it should do this.

### Additional Notes

General validations pass on my setup.


sbathe:
I need to base it against the correct branch. I definitely did not change so many files :)

Issue #7878, opened 2021-02-09
unwanted check new version 


obeyler:
### bug

--global.checknewversion arg was not set and so on regarding the documentation it should be define to false
 but I can see on my log :
```
time="2021-02-09T08:45:13Z" level=warning msg="Error checking new version: Get \"https://update.traefik.io/repos/traefik/traefik/releases\": dial tcp 104.26.2.101:443: i/o timeout"
```

args use on our deployment in K8S
```
              args:
              - --api=true
              - --api.dashboard=true
              - --log.level=INFO
              - --accesslog=true
              - --global.sendanonymoususage=false
              - --providers.kubernetescrd=true
              - --providers.kubernetescrd.allowcrossnamespace=false
              - --entrypoints.web=true
              - --entryPoints.web.address=:8000
              - --entrypoints.websecure=true
              - --entryPoints.websecure.address=:8443
              - --metrics.prometheus=true
              - --entrypoints.postgresql=true
              - --entryPoints.postgresql.address=:5432
              - --serverstransport.insecureskipverify=true

```
### Output of `traefik version`: 
```
traefik:v2.4.0
```

obeyler:
may the reason is : https://github.com/traefik/traefik/blob/326be29568426e4ac8e2bf65fdae3906e8dd58fd/cmd/configuration.go#L22

sbathe:
Seeing the same in latest v1.7 as well. I see even in tests, we do not set CheckNewVersion: false anywhere. So that case is not tested anywhere as well.

traefiker:
Closed by #7933.

PR #7919, opened 2021-02-20
make sure checkNewVersion configurationg flag is honored

sbathe:
### What does this PR do?

- Changes to configuration.go to not override checkNewVersion
- Add some logging to checkNewVersion function, similar to func stats

#7878

### Motivation

Make sure we maintain the documented behaviour and honour the configuration flags.


ldez:
Hello,

as you can see on issue #7878, we put the label `area/documentation` which means that the problem is not the code but the documentation.

So we will decline your PR.

PR #7926, opened 2021-02-24
Add option to Deny IPs on middleware Whitelist

mathcantin:
### What does this PR do?

Add an option on the middleware whitelist to transform it to denylist.

### Motivation

It's very useful to deny access to an IP range from traefik. The middleware whitelist is suitable for this because it always defines which IP address is allowed, but in negative mode.

### More

- [X] Added/updated tests

### DIFF
--- a/docs/content/middlewares/ipwhitelist.md
+++ b/docs/content/middlewares/ipwhitelist.md
@@ -225,3 +225,67 @@ http:
             - "127.0.0.1/32"
             - "192.168.1.7"
 ```
+
+### `deny`
+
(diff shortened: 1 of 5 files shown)


rtribotte:
Hello @mathcantin,

Thank you for your work and your interest in Traefik.

Unfortunately, this feature has already been discussed and declined, as you can see here #4454 and here #4026.
For the same reasons, I'll decline your pull request, though, there's currently at least two [plugins](https://pilot.traefik.io/plugins) that are providing this feature: 

- https://pilot.traefik.io/plugins/276812076107694611/deny-ip-plugin
- https://pilot.traefik.io/plugins/280093067746214409/fail2-ban

Issue #7951, opened 2021-03-04
Add OrganizationalUnitName (OU) to passtlscert middleware


fernferret:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

With the `passtlscert` middleware I cannot get the OU from a certificate passed in the `info` section. My workaround is to pass the `pem` cert but as noted in the traefik docs this can cause [issues with large header](https://doc.traefik.io/traefik/middlewares/passtlsclientcert/) sizes. It also forces a second PEM decode later, when all I really need is the OU.

From the [tests](https://github.com/traefik/traefik/blob/master/pkg/middlewares/passtlsclientcert/pass_tls_client_cert_test.go#L354), I'd like to see the following:

`Subject="DC=org,DC=cheese,C=FR,C=US,ST=Cheese org state,ST=Cheese com state,L=TOULOUSE,L=LYON,O=Cheese,O=Cheese 2,2,CN=*.cheese.com"`

become

`Subject="DC=org,DC=cheese,C=FR,C=US,ST=Cheese org state,ST=Cheese com state,L=TOULOUSE,L=LYON,O=Cheese,O=Cheese 2,OU=Simple Signing Section,OU=Simple Signing Section 2,CN=*.cheese.com"`

I've already written up the diff (for both Subject and Issuer) and can open a PR but wanted to make sure there wasn't a good reason not to add these fields, or that they were omitted intentionally. I noticed the DNInfo struct [references rfc3739](https://tools.ietf.org/html/rfc3739) which does include `organizationalUnitName`.

It passes the tests and doesn't seem to cause any additional shenanigans as it's added just like `Organization` so it's optional.

jbdoumenjou:
Hello @FernFerret,

Thanks for your interest in Traefik.

Indeed, the RFC defines that the `organizationalUnitName` is only available in the [subject](https://tools.ietf.org/html/rfc3739#section-3.1.2) and not in the [issuer](https://tools.ietf.org/html/rfc3739#section-3.1.1) information.

The fields are added accordingly to users' needs to avoid having too many options.
Feel free to open a PR by following the [guidelines](https://github.com/traefik/contributors-guide/blob/master/pr_guidelines.md). :smiley:

FernFerret:
>  is only available in the subject and not in the issuer information.

[...]

PR #7958, opened 2021-03-06
Add organizationalUnit to passtlscert middleware

FernFerret:
### What does this PR do?

Adds the Organizational Unit Name; aka OU (traefik annotation `organizationalUnit` to follow `organization` vs `Organization Name`) to the `passtlscert` middleware. This field is added to both the `subject` and the `issuer`. Initial discussion took place in #7951.

While the OU on the "issuer" not officially part of the RFC
(https://tools.ietf.org/html/rfc3739#section-3.1.1) I've included it
~~because traefik includes the "province" in the same fashion~~. In
addition, it would be extra logic to remove it when it might actually be
useful to have.

### Motivation

I'd like to be able to access the OU passed without having to pass the entire `pem` block. Doing so requires a second pem decode (when traefik is already doing one).

### More

- [x] Added/updated tests
  * I added one additional test that we can get rid of if you'd like which tests to make sure it's not present if not requested, to ensure legacy requests aren't affected.

### Additional Notes

I ended up changing more than I originally liked about the Vue UI so that Organization and Organizational Unit could be "next" to each other. I'm totally cool with reverting this portion and tacking OU onto the end to minimize diffs. Your call.

Here are previews of what the UI looks like with this PR:

Screen Shot 2021-03-06 at 12 06 34 PM

The second screenshot shows what it would look like for existing users. For example if this was added in `2.5` and they didn't add the `organizationalUnit`, they'd get a red switch next to it:

Screen Shot 2021-03-06 at 12 07 04 PM


FernFerret:
Well, now that it's plastered everywhere (in my commit message and initial PR) I realized my mistake in the commit comment. The province is in fact included in `stateOrProvinceName`, so I guess there would be justification for removing this field from the subject.

Again, open to suggestions/ideas.

jbdoumenjou:
Hello @FernFerret ,
 
Thanks for your contribution!

jbdoumenjou:
Hello @FernFerret,

Thanks to your PR, we realized that it would be cleaner to separate the distinguished name of the issuer and the subject.
In that way, we can avoid providing non defined attribute in the [rfc](https://datatracker.ietf.org/doc/html/rfc3739).

WDYT?

FernFerret:
@jbdoumenjou I'm fine with that!

I think you're saying we separate the `DistinguishedNameOptions` into 2 structs (one for issuer and one for subject). `IssuerDistinguishedNameOptions` vs `SubjectDistinguishedNameOptions`?

jbdoumenjou:
Yes, to have structures containing only the fields defined by the RFC.
And split `getDNInfo` in something like `getSubjectDNInfo` and `getSubjectDNInfo`.
It will duplicate some code but it will be clearer and provide better support of the RFC.

However, to be backward compatible, we must support all the existing fields in the two structures and manage the deprecation (comment + doc).

We can create a PR to handle the refactor/doc so you can rebase on it and manage only the new field.

WDYT?

FernFerret:
Sounds good to me I'll get the changes made soon!

FernFerret:
Hi @jbdoumenjou I pushed another commit today that I believe makes the requested changes. I split the struct into one for issuer vs subject and I do like it better, even at the cost of the duplicated fields in the struct. It maps a lot nicer to the RFCs this way.

I wasn't exactly sure what you meant when you said:

> However, to be backward compatible, we must support all the existing fields in the two structures and manage the deprecation (comment + doc).
> 
> We can create a PR to handle the refactor/doc so you can rebase on it and manage only the new field.

I ended up leaving `CommonName` in the `Issuer` despite it not being part of the RFC. I believe this is what you meant, but I just wanted clarification. Is there some action I'm expected to take for the deprecation guide? Is there any information you have that I could read up on that?

Thanks again and sorry for the delay!

jbdoumenjou:
Hello @FernFerret,

I was wrong for the deprecation.
In fact, the RFC says:
> Additional attributes MAY be present, but they SHOULD NOT be
   necessary to identify the issuing organization. 

So, I think it is ok to keep the field as-is.

rtribotte (approved):
LGTM

jbdoumenjou (approved):
LGTM

mpl:
WIP on improving the doc.

mpl (approved):

PR #7962, opened 2021-03-07
Use Docker dependency directly without replace directive

dom3k:
### What does this PR do?

This pull request remove usage of replace directive for Docker dependency.

### Motivation

According to this (source: https://github.com/testcontainers/testcontainers-go/issues/127#issuecomment-571616048):

> Adding a replace works if you want to use a version that matches a tagged release of docker (for which tags can be found in the docker/engine repository), but if that's the reason, it's probably most most logical to pick a tag (but pin by commit, because Go doesn't accept the YY.MM tags)

> Unfortunately, the replace won't help for consumers of this project though (only for the project itself); the replace is not inherited by those projects, so everyone using this package would need to add the same replace rule to their go.mod.

And this (source: https://github.com/testcontainers/testcontainers-go/issues/127#issuecomment-651114052):

> FWIW, all tags from the docker/engine repository have been added to the upstream docker/docker (moby/moby) repository, so it should no longer be needed to use the docker/engine repository

Now it is possible to usage Docker dependency directly by `go get github.com/docker/docker@v19.03.6`

### Additional Notes

https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive

> As described in the 'go.mod' concepts section above, replace directives provide additional control in the top-level go.mod for what is actually used to satisfy a dependency found in the Go source or go.mod files, while replace directives in modules other than the main module are ignored when building the main module.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,7 @@ module github.com/traefik/traefik/v2
 
 go 1.16
 
+// github.com/docker/docker v17.12.0-ce-rc1.0.20200204220554-5f6d6f3f2203+incompatible => v19.03.6
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/ExpediaDotCom/haystack-client-go v0.0.0-20190315171017-e7edbdf53a61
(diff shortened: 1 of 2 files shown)


ldez (approved):
Thanks :+1:

dom3k:
Thank you, too! :+1:

PR #7963, opened 2021-03-07
Add wildcard hostname rule to kubernetes gateway

### LINKED ISSUE #7874: Service API hostname wildcards
The service api allows wildcards in the httproute hostname field. This is hard to see because their documentation site appears to be mis-rendering the `*` as markdown markup rather than a literal `*` so the text is garbled https://kubernetes-sigs.github.io/service-apis/spec/#networking.x-k8s.io/v1alpha1.HTTPRouteSpec . That said it can be gleaned from the PR that introduced it https://github.com/kubernetes-sigs/service-apis/pull/416/files .

Yes this is bad, and I'm going to bring up the rendering issue with them soon. Still, without this I cannot figure out how to do wildcard hostnames using traefik as a gateway. It seems that you could parse the hostname [here](https://github.com/traefik/traefik/blob/d211437d6ca4b26bd5e8f7901adb916bdf44f61b/pkg/provider/kubernetes/gateway/kubernetes.go#L574) and generate a subdomain HostRegexp from it instead. If I was more confident in go I'd offer to do it myself, but you were so quick on #7731 that I thought I'd ask this here first.

jberger:
closes #7874

### What does this PR do?

Allows a kubernetes gateway api httproute definition to include wildcard hostnames (ie. `*.foo.com`). As mentioned in #7874 per [the spec](https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.Hostname) this is an allowed usage but was originally badly rendered by their documentation renderer (which has now been fixed) so as to properly render the `*` character, which had previously been interpreted as markdown markup). 

### Motivation

I've been enjoying using the gateway api via traefik but this became a blocker for rolling it out in what should otherwise be a simple deployment.

### More

- [x] Added/updated tests

I did not add documentation but as it is documented in the gateway-api spec and the traefik documentation refers to that, I think it isn't incorrect to omit it.

### Additional Notes

I'm not a go developer (I'm a Perl dev with other dynamic languages under my fingers) and a first time contributor (of code) to this project, so if this isn't idiomatic code or if it isn't even the proper generation of rules, please let me know and I'll happily update it (or feel free to correct it for me).

### DIFF
--- a/pkg/provider/kubernetes/gateway/kubernetes.go
+++ b/pkg/provider/kubernetes/gateway/kubernetes.go
@@ -588,22 +588,25 @@ func hostRule(httpRouteSpec v1alpha1.HTTPRouteSpec) (string, error) {
 
 	for _, hostname := range httpRouteSpec.Hostnames {
 		host := string(hostname)
-		if host == "" {
-			continue
+		// When unspecified, "", or *, all hostnames are matched.
+		// This field can be omitted for protocols that don't require hostname based matching.
+		// TODO Refactor this when building support for TLS options.
+		if host == "*" || host == "" {
+			return "", nil
 		}
 
 		wildcard := strings.Count(host, "*")
-		if wildcard >= 1 {
-			// https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.Hostname
-			if !strings.HasPrefix(host, "*.") || wildcard > 1 {
-				return "", fmt.Errorf("invalid rule: %q", host)
-			}
-
-			hostRegexNames = append(hostRegexNames, strings.Replace(host, "*.", "{subdomain:[a-zA-Z0-9-]+}.", 1))
+		if wildcard == 0 {
+			hostNames = append(hostNames, host)
 			continue
 		}
 
-		hostNames = append(hostNames, host)
+		// https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.Hostname
+		if !strings.HasPrefix(host, "*.") || wildcard > 1 {
+			return "", fmt.Errorf("invalid rule: %q", host)
+		}
+
+		hostRegexNames = append(hostRegexNames, strings.Replace(host, "*.", "{subdomain:[a-zA-Z0-9-]+}.", 1))
 	}
 
 	var res string
--- a/pkg/provider/kubernetes/gateway/kubernetes.go
+++ b/pkg/provider/kubernetes/gateway/kubernetes.go
@@ -423,7 +423,17 @@ func (p *Provider) fillGatewayConf(client Client, gateway *v1alpha1.Gateway, con
 				continue
 			}
 
-			hostRule := hostRule(httpRoute.Spec)
+			hostRule, err := hostRule(httpRoute.Spec)
+			if err != nil {
+				listenerStatuses[i].Conditions = append(listenerStatuses[i].Conditions, metav1.Condition{
+					Type:               string(v1alpha1.ListenerConditionResolvedRefs),
+					Status:             metav1.ConditionFalse,
+					LastTransitionTime: metav1.Now(),
+					Reason:             string(v1alpha1.ListenerReasonDegradedRoutes),
+					Message:            fmt.Sprintf("Skipping HTTPRoute %s: invalid hostname: %v", httpRoute.Name, err),
+				})
+				continue
+			}
 
 			for _, routeRule := range httpRoute.Spec.Rules {
 				rule, err := extractRule(routeRule, hostRule)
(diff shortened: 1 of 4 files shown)


jberger:
I don't understand why travis failed. How can it be restarted?

jberger:
I was able to force it to restart with a trivial rebase and force-push but it still fails. I have no idea from that output what the error might be. Rather frustrating.

jberger:
Looks like it was just some hiccup in travis and/or the docker registry.

jberger:
Travis is finally passing and now semaphore stops? What is going on?! I'm not changing anything!

jberger:
And now travis failed again. This is ridiculous. someone else can deal with broken CI systems, I'm done.

jbdoumenjou:
Hello @jberger,
 
Thanks for your contribution!

Don't worry about the CI, we are actively working on CI fixes. :smiley:

jberger:
Thanks for getting back to me. Honestly one of the biggest reasons I was frustrated with the CI (apologies!) is that in many larger projects, a failing CI report means you don't even get looked at and I was concerned that I wouldn't even hear from you.

For the future reviewer, please note that I took the core of the new business logic from the [k8s ingress](https://github.com/traefik/traefik/blob/606b43dc512fbeceb79d756a6ddcf796676dc884/pkg/provider/kubernetes/ingress/kubernetes.go#L375-L377) provider, so I hope that it is at least that correct. My biggest design concern is whether a mixed set of wildcard and non-wildcare hosts should all be grouped together in one HostRegexp rule or should they be two rules `||`-ed together. The documentation suggests that you can have non-regexp entries in a HostRegexp rule but I don't know if there's a performance concern. That's up to you.

Thanks again for getting back to me.

jbdoumenjou:
Hi @jberger,

We made some modifications, please see our commit.
We separated the `Host` and `HostRegexp` rules, manage the error in the listener statuses and add some tests.
Does it seem good for you?

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -571,27 +581,46 @@ func (p *Provider) makeGatewayStatus(listenerStatuses []v1alpha1.ListenerStatus)
...
 			continue
 		}
-		if strings.HasPrefix(host, "*.") {
-			ruleType = "HostRegexp"
-			host = strings.Replace(host, "*.", "{subdomain:[a-zA-Z0-9-]+}.", 1)
+
+		wildcard := strings.Count(host, "*")
+		if wildcard >= 1 {
+			// https://gateway-api.sigs.k8s.io/spec/#networking.x-k8s.io/v1alpha1.Hostname
+			if !strings.HasPrefix(host, "*.") || wildcard > 1 {
```
dtomcej, line 597: I'm pretty sure that this will work incorrectly if there are 2 wildcards no? the OR will always pass?

could this be reworked as a fast fail instead of nested ifs?

```
if !strings.HasPrefix(host, "*.") && wildcard > 0 {
// fail if the wildcard is not at the start
return error
}

if wildcard > 1 {
// use hostregex
} else {
//use host
}

Comment (dtomcej) [reply pkg/provider/kubernetes/gateway/kubernetes.go:597]: If the fail on 2 wildcards is intended, then this would do this:

```
if !strings.HasPrefix(host, "*.") && wildcard > 0 {
// fail if the wildcard is not at the start
return error
}

switch wildcard {
case wildcard > 1 :
// fail that there are too many wildcards
break
case wildcard = 1:
// use hostregex
break
default:
//use host
break
}

jberger: Multiple wildcards are not allowed and the wildcard must be at the front of the domain `*.foo.com`. It basically is the fail early that you mention, I'm not sure there's much practical difference in the change you propose.

pkg/provider/kubernetes/gateway/kubernetes_test.go
```diff
@@ -945,14 +946,78 @@ func TestHostRule(t *testing.T) {
...
+				},
+			},
+			expectedRule: "HostRegexp(`{subdomain:[a-zA-Z0-9-]+}.bar.foo`)",
+		},
+		{
+			desc: "Alone wildcard",
+			routeSpec: v1alpha1.HTTPRouteSpec{
+				Hostnames: []v1alpha1.Hostname{
+					"foo.foo",
+					"*",
```
jberger, line 974: I think technically this is allowed too, though I care much less about supporting it as you could just not specify a host rule at all. And technically an empty string is `*` too.

jberger:
It did take me a couple of reads on the error checking logic before I grokked it but yes it seems fine to me. I made one comment that technically supporting `*` is specced but not very relevant (for me at least).

jbdoumenjou:
Thanks to your comments, the last modifications propose a better implementation following the specification on "" and "*" hostname.
WDYT?

jberger:
Much clearer! My only question, and this is probably in the weeds and/or pedantic: if someone had hostname `foo.com` and `*`, should that just ignore all hostname requirements? I think in this code it would make the hostname effectively require `foo.com` would it not?

Other than that, LGTM!

jbdoumenjou:
@jberger, 
I'm glad it is clearer :smiley: 

TBH we were wondering how to handle the case where there is a wildcard and, at least, one other hostname.
From a logical perspective, as soon as we use a wildcard, it is like we don't care about the domain name anymore as we will match all of them.
It could be an issue with the way we determine the TLS option to use with the request but as we don't support that feature now, for the sake of simplicity, we have chosen to ignore all other hostnames.

That is why we add the comment:
> // TODO Refactor this when building support for TLS options.

If you see another issue with that or something we missed, please let us know :pray:

jbdoumenjou (approved):
LGTM

SantoDE (approved):
LGTM 👍

jberger:
I see some approvals, is there something I can do to push this over the finish line?

dtomcej (approved):
LGTM
:shipit:

jberger:
Cheers! Thanks! I look forward to using this!

PR #7971, opened 2021-03-09
Improve host name resolution for TCP proxy

H-M-H:
### What does this PR do?

Previously `ResolveTCPAddr` has always been used. But as stated in go's
docs this is not recommended:
> The address parameter can use a host name, but this is not
> recommended, because it will return at most one of the host name's
> IP addresses.

For host names `ResolveTCPAddr` in combination with `DialTCP` has now
been replaced with a single call to `Dial`.

The added advantage is that `Dial` tries all IPs the host name resolves
to:
> When using TCP, and the host resolves to multiple IP addresses, Dial
> will try each IP address in order until one succeeds.

### Motivation
The reason why I stumbled over this is that one of the IPs a host name resolves to is invalid and I tried to figure out how to get traefik to use the valid one.

### More

- [x] Added/updated tests
  - Removed `TestLookupAddress` as it feels like it is only testing implementation details for `Proxy.target`, which is now unused if a host name has been given as address.

### DIFF
--- a/pkg/tcp/proxy.go
+++ b/pkg/tcp/proxy.go
@@ -31,7 +31,7 @@ func NewProxy(address string, terminationDelay time.Duration, proxyProtocol *dyn
 		return nil, fmt.Errorf("unknown proxyProtocol version: %d", proxyProtocol.Version)
 	}
 
-	// enable the refresh of the target only if the address in an IP
+	// enable the refresh of the target only if the address in not an IP
 	refreshTarget := false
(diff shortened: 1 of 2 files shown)


H-M-H:
Looks like I am running into the same issue with travis like #7963.

rtribotte (approved):
Thank you 👍

jbdoumenjou (approved):
LGTM

ldez (approved):
Thanks :+1:

PR #7973, opened 2021-03-10
Make RouterName and ServiceName uniquely identifiable in k8s Ingress provider

semekh:
### What does this PR do?

Since namespaces and service names are allowed to contain `-` in them, current RouterName and ServiceName do not uniquely identify the underlying resource. This PR tries to address that by using `.` instead of `-` as a separator, resulting in a unique mapping.

### Motivation

Currently, there is no clean way to identify the actual Ingress/Service an access log line belongs to.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -278,10 +278,10 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 					continue
 				}
 
-				serviceName := provider.Normalize(ingress.Namespace + "-" + pa.Backend.ServiceName + "-" + pa.Backend.ServicePort.String())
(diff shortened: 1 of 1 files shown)


SeMeKh:
@ldez Friendly ping. Do you think it's possible to have this merged?

dtomcej:
Hello @SeMeKh  ,

Thanks for your contribution!

as you can see in our contribution guide, this means that we have to take a closer look at the scope of impact for that PR to see, how it would interact with the other parts of Traefik.

We will come back to you soon when we did the first design review iteration. We understand that this PR has been in review for a while, but it has not been forgotten. If you would like to update this PR to resolve merge conflicts and to fix CI failures, that would assist in moving this PR along.

Thanks again!

rtribotte:
Hello @SeMeKh,

Thank you very much for your contribution.

We took the time to review your pull request.
There are two main issues with the change you are proposing that make it breaking for existing configurations:
- it would be breaking for people who are today relying on the current format to parse the access logs
- it would be breaking for configurations that use cross provider references (services)

Unfortunately, for those reasons, I'll decline your PR.

As for now, the next step would be to discuss of that enhancement in a dedicated proposal issue.

PR #8027, opened 2021-03-31
Add option to keep only healthy ECS tasks

Michampt:
### What does this PR do?

~Allow specification of ECS Services to monitor for tasks to add
when it's unnecessary or undesirable to scan the entire cluster.~

Add checking the health state of a task and require it to have a Healthy status before configuring a route for it

### Motivation

~I deploy Traefik in ECS to handle requests for other services in ECS, however, we don't want Traefik scanning every task in the cluster(s) so I needed a way to tell each Traefik service to only handle the tasks of specific services, yet still maintain the ability to scan every task if none specified.~

Sometimes adding a task to the route before it's healthy according to the ECS health checks is undesirable as there is a possibility that the task is not ready to handle connections and is still having traffic routed to it.

### More

- [ ] Added/updated tests

### Additional Notes

~None~

After more consideration, the AWS ECS Service filter feature has been removed from this PR (see https://github.com/traefik/traefik/pull/8027#issuecomment-1252319120)

### DIFF
--- a/pkg/provider/ecs/ecs.go
+++ b/pkg/provider/ecs/ecs.go
@@ -2,13 +2,13 @@ package ecs
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"text/template"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
@@ -265,20 +265,15 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
 				}
 				taskArray, err := p.listTasks(ctx, client, input)
 				if err != nil {
+					var snfe *ecs.ServiceNotFoundException
+					var cnfe *ecs.ClusterNotFoundException
 					// Don't want to stop operations if a cluster or service was misstyped or went down so we just log it and move on
-					if aerr, ok := err.(awserr.Error); ok {
-						switch aerr.Code() {
-						case ecs.ErrCodeServiceNotFoundException:
-							logger.Errorf("Service not found: %s", service)
-							break
-						case ecs.ErrCodeClusterNotFoundException:
-							logger.Errorf("Cluster not found: %s", cluster)
-							break
-						default:
-							return nil, err
-						}
-					} else {
-						// If it's not an AWS Error, it's probably something worse so hand it up
+					switch {
+					case errors.As(err, &snfe):
+						logger.Errorf("Service not found: %s", service)
+					case errors.As(err, &cnfe):
+						logger.Errorf("Cluster not found: %s", cluster)
+					default:
 						return nil, err
 					}
 				}
@@ -293,15 +288,12 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
 			}
 			taskArray, err := p.listTasks(ctx, client, input)
 			if err != nil {
-				if aerr, ok := err.(awserr.Error); ok {
-					switch aerr.Code() {
-					case ecs.ErrCodeClusterNotFoundException:
-						logger.Errorf("Cluster not found: %s", cluster)
-						break
-					default:
-						return nil, err
-					}
-				} else {
+				var cnfe *ecs.ClusterNotFoundException
+				// Don't want to stop operations if a cluster or service was misstyped or went down so we just log it and move on
+				switch {
+				case errors.As(err, &cnfe):
+					logger.Errorf("Cluster not found: %s", cluster)
+				default:
 					return nil, err
 				}
 			}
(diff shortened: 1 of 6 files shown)


Michampt:
@jbdoumenjou Is there anything further you'd like me to do for this PR? 

Looking at the semaphoreci failure it doesn't seem to be anything from this PR but I can try to work on getting that to pass if need be.

pkg/provider/ecs/ecs.go
```diff
@@ -31,8 +32,10 @@ type Provider struct {
 	DefaultRule      string `description:"Default rule." json:"defaultRule,omitempty" toml:"defaultRule,omitempty" yaml:"defaultRule,omitempty"`
 
 	// Provider lookup parameters.
-	Clusters             []string `description:"ECS Clusters name" json:"clusters,omitempty" toml:"clusters,omitempty" yaml:"clusters,omitempty" export:"true"`
+	Clusters             []string `description:"ECS Cluster names" json:"clusters,omitempty" toml:"clusters,omitempty" yaml:"clusters,omitempty" export:"true"`
+	Services             []string `description:"ECS Service names" json:"services,omitempty" toml:"services,omitempty" yaml:"services,omitempty" export:"true"`
 	AutoDiscoverClusters bool     `description:"Auto discover cluster" json:"autoDiscoverClusters,omitempty" toml:"autoDiscoverClusters,omitempty" yaml:"autoDiscoverClusters,omitempty" export:"true"`
+	RequireHealthyTask   bool     `description:"Require a task to be in the healthy state" json:"requireHealthyTask,omitempty" toml:"requireHealthyTask,omitempty" yaml:"requireHealthyTask,omitempty" export:"true"`
```
ddtmachado, line 38: maybe `RequireRunningTask ` sounds better and is more coherent with ECS task status

pkg/provider/ecs/ecs.go
```diff
@@ -230,41 +237,67 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
...
-				resp, err := client.ecs.DescribeTasksWithContext(ctx, &ecs.DescribeTasksInput{
-					Tasks:   page.TaskArns,
-					Cluster: &c,
-				})
+		if len(services) > 0 {
+			for _, s := range services {
+				service := s
+				input = &ecs.ListTasksInput{
+					Cluster:       &cluster,
+					DesiredStatus: aws.String(ecs.DesiredStatusRunning),
```
ddtmachado, line 261: I don't see the point on setting it here without checking the flag, especially since it seems to be then ignored by listTasks

pkg/provider/ecs/ecs.go
```diff
@@ -230,41 +237,67 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
...
 					}
 				}
+				for _, t := range taskArray {
+					tasks[aws.StringValue(t.TaskArn)] = t
+				}
+			}
+		} else {
+			input = &ecs.ListTasksInput{
+				Cluster:       &cluster,
+				DesiredStatus: aws.String(ecs.DesiredStatusRunning),
```
ddtmachado, line 285: same here

ddtmachado (changes requested):
Hi @Michampt, I'll leave proper design / code review for my colleagues but still I have a few notes from trying this out

pkg/provider/ecs/ecs.go
Michampt: I did this so that more specifically the task health status needs to be set to "Healthy" not just running. prior to this change the task would get added to Traefik's list, but it hadn't passed healthchecks yet so it was undesirable to have traffic forwarded to that task.

pkg/provider/ecs/ecs.go
Michampt: Good point, not really being used. I'll make a new commit without this. Thanks!

pkg/provider/ecs/ecs.go
Michampt: Same as above, Thank you!

kennon:
The added functionality here looks great - without the container healthcheck there are lots of ECS use cases where the routing happens before the services are actually fully initialized, and there doesn't appear to be any way to prevent this without this extension. Bump :-)

rtribotte:
Hello @Michampt,

We are sorry for getting back so late, but we are now ready to review the PR actively!
Is it still a subject that you are interested in, and available to answer any potential questions?

Michampt:
Hey @rtribotte , no worries! I actually no longer work at the company that utilized this fix, and actually they stopped using Traefik altogether. I'm not opposed to getting this PR to the finish line though, as I think it'd be useful for future users.

rtribotte:
Hello @Michampt,

Thanks for the feedback!

As of now, the PR adds only the possibility to filter out services only if the cluster auto-discovery is disabled, meaning also that it is mandatory to filter out services to also filter out clusters, and so, configure a list of clusters as well as a list of services.

It could be arguable to filter only on services no matter what the cluster is.
Another possible use case would be to filter out services per cluster.

To be clear, we don't think we should do something as complex as the latter suggestion.
Nonetheless, we are wondering what you think is the most common use case and also what was, regarding that specific matter, the use case you wanted to solve.

Michampt:
So, racking my brain here from over a year ago, I think basically:

I had a Traefik deployment per specific ECS Service, so maintenance on one didn't affect others, plus separation of traffic and all that. So I found it useful to specify a specific service in a cluster to have managed by a specific Traefik deployment.

The other issue I solved with this PR is when tasks in an ECS service would come up, the app that would get traffic routed to it took quite a bit to finish health checks and be ready for traffic, so it was getting traffic well before it was actually healthy and caused a lot of unnecessary panic.

kevinpollet:
Hello @Michampt,

Thanks for your feedback,

> I had a Traefik deployment per specific ECS Service, so maintenance on one didn't affect others, plus separation of traffic and all that. So I found it useful to specify a specific service in a cluster to have managed by a specific Traefik deployment.

To make Traefik discover only a subset of ECS services, the constraints option could be leveraged as described in the [documentation](https://doc.traefik.io/traefik/providers/ecs/#constraints) (we fixed the missing documentation in https://github.com/traefik/traefik/pull/9354).

Because of that reason, we will remove the service filtering feature and keep only the filtering of healthy task.

rtribotte (approved):
Thanks 👍

kevinpollet (approved):
Thanks 👍

ldez (approved):
LGTM

Michampt:
@kevinpollet Sounds good. Thanks!

PR #8029, opened 2021-04-01
Add the ability to require a healthy task

Michampt:
### What does this PR do?

Add checking the health state of a task and require it to have a Healthy status before configuring a route for it

### Motivation

Sometimes adding a task to the route before it's healthy according to the ECS health checks is undesirable as there is a possibility that the task is not ready to handle connections and is still having traffic routed to it.

### More

- [ ] Added/updated tests

### Additional Notes

None

### DIFF
--- a/docs/content/reference/static-configuration/cli-ref.md
+++ b/docs/content/reference/static-configuration/cli-ref.md
@@ -486,6 +486,9 @@ Polling interval (in seconds) (Default: ```15```)
 `--providers.ecs.region`:  
 The AWS region to use for requests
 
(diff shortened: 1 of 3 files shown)


Michampt:
I added #8027 yesterday, let me know if I should combine these into 1 since they're both pretty small code changes.

Also I'm not sure if "Added/Updated documentation" counts as the two md files being updated or not so I'm leaving it unchecked for now

Michampt:
Closing this as I don't want to cause conflicts with #8027 , Added a separate commit to that PR with these changes

PR #8033, opened 2021-04-05
Error span on 5xx only

kevtainer:
### What does this PR do?

Generates error spans on `5xx` errors rather than `>= 400` errors.

### Motivation
A client generating hundreds of 401s every few minutes.

Generating error spans for client-side errors creates a lot of noise on high-volume systems when the problem is not addressable by the server operator.

Most tracers implement similar logic, only generating an error when the response code is 500 or greater. Examples:

https://github.com/opentracing-contrib/go-stdlib/blob/v1.0.0/nethttp/client.go#L191-L193
https://github.com/opentracing-contrib/nginx-opentracing/blob/v0.14.0/opentracing/src/request_tracing.cpp#L64-L67

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/migration/v2.md
+++ b/docs/content/migration/v2.md
@@ -348,7 +348,7 @@ It is therefore necessary to update [RBAC](../reference/dynamic-configuration/ku
 
 ### Non-ASCII Domain Names
 
-In `v2.4.8` we introduced a new check on domain names used in HTTP router rule `Host` and `HostRegexp` expressions,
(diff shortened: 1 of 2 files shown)


dtomcej (approved):
LGTM
:shipit:

SantoDE (approved):
LGTM 👼

jbdoumenjou (approved):
LGTM :+1:

Issue #5228, opened 2019-08-19
v2: support `NOT` in `rules` definition


Kosta-Github:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

I would like to see support for `NOT` (or `!`) in addition to `AND` and `OR` for defining routing `rules`, just to enable support for proper `boolean` operations.

E.g.:
```text
rule = "Host(`traefik.io`) || (Host(`containo.us`) && NOT Path(`/no-traefik`)"
```
or (note the `!` in front of `Path()`):
```text
rule = "Host(`traefik.io`) || (Host(`containo.us`) && !Path(`/no-traefik`)"
```

denysvitali:
I was about to open an identical issue, but found yours:
here is my original issue:  
   
---
### What did you expect to see?

When writing a router rule, it would be nice to be able to negate parts of the expression, e.g: 
```
[http.routers.https-redirect]
   rule = "HostRegexp(`{any:.*}`) && !Host(`localhost`)"
   middlewares = ["https-redirect"]
   service = "dummy"
```

Currently this feature doesn't seem to be mentioned in the [documentation](https://docs.traefik.io/v2.0/routing/routers/#rule), nor it seems to exists a set of allowed operators (only `&&` and `||` are mentioned in a green "admonition").

---

I think that using the `!` operator will result in a more clean code (and will be more consistent with the `&&` / `||`) operators -- either we go with the Python style operators (`and`, `or`, `not`) or the C style operators as they're being used now (`&&`, `||`)

glebovdev:
NOT (or !) operator support feature is still pretty much needed. 
My use case, for example: show basic auth only for those who do not provide a specific header in request. 
But the following code of course doesn't work as expected:
<pre><code>- "traefik.http.middlewares.auth.basicauth.users=user:hashed_password"
- "traefik.http.routers.traefik-auth.rule=!Headers(`PassBasicAuth`, `true`)"
- "traefik.http.routers.traefik-auth.middlewares=auth"</code></pre>

everflux:
I would like this as well - especially in the context of a TCP router. I would like to express "if no SNI host is present" (plain TLS handshake without SNI).
Use case: forward these connections to a special handler while providing https on the same port for browsers

CourtneySherman:
I will add another plus for this feature request and another use case. For services that allow api calls, the not operator would make it much easier to allow the api calls to bypass middleware based authentication (basic auth, authelia, oauth).

[...]

PR #8036, opened 2021-04-06
Add PathExcept as a new Router Rule

oxzi:
### What does this PR do?

This pull request adds another Routing Rule named `PathExcept` to deny certain paths, as addressed in #4812.

### Motivation

_The following text is a copy of the commit message:_

Sometimes it is necessary to limit the allowed paths. Currently,
however, only a positive logic is provided to allow paths. The
possibility to exclude paths is missing so far and was requested several
times, e.g., in #4812.

One approach was #4814 which created a middleware for this purpose.
However, this pull request got stalled and the decision was made that
this should be part of the routing, not the task of a middleware.

Thus, this is another approach which adds a one-time negative logic to
the routing rules. The PathExcept rule denies the paths defined in it,
which may be specified as regular expressions.

Since this rule alone is pointless, it must be used together with
another path rule (Path or PathPrefix). This other rule allows all
subpaths, while the PathExcept rule defines exclusions for this.

### More

- [x] Added/updated tests

### Additional Notes

As far as I have seen, both the demand for such a feature as well as the necessity are debatable. The changes in this pull request are a suggestion.

### DIFF
--- a/docs/content/routing/routers/index.md
+++ b/docs/content/routing/routers/index.md
@@ -228,17 +228,18 @@ If the rule is verified, the router becomes active, calls middlewares, and then 
 
 The table below lists all the available matchers:
 
-| Rule                                                                   | Description                                                                                                    |
-|------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------|
-| ```Headers(`key`, `value`)```                                          | Check if there is a key `key`defined in the headers, with the value `value`                                    |
-| ```HeadersRegexp(`key`, `regexp`)```                                   | Check if there is a key `key`defined in the headers, with a value that matches the regular expression `regexp` |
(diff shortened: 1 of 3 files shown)


brunocascio:
Hi folks, 

Do you have an ETA?

rtribotte:
Hello @oxzi,

Thanks a lot for your contribution.

We took a look at your PR and the origin issue.
Unfortunately, after discussion between the maintainers, we prefer to decline your PR and the origin issue profit of negation syntax for rules (#5228).
The negation would allow achieving the same goal, but will not introduce new rules matcher.

oxzi:
Thanks for getting back to this, @rtribotte.

A general negation is of course more beautiful and I am glad that it is being worked on.

Please keep up the good work!

PR #8050, opened 2021-04-11
Add docker buildx Dockerfile to build multi-arch image

sylr:
### What does this PR do?

It adds an all-in-one buildx Dockerfile which generates a multi-arch image.

### Motivation

Being able to generate multi-arch image locally.

### More

- [ ] Added/updated tests


ldez:
Hello,

Currently, we provide multi-arch images via the official Docker images.
We don't handle that in our CI because the images are built by Docker itself.

We will decline for PR.

PR #8057, opened 2021-04-13
healthcheck: add support at the load-balancers of services level

### LINKED ISSUE #7693: Weighted round robin with healthcheck
### Do you want to request a feature or report a bug?

Feature

### What did you do?

Configure two servers with weighted round robin and health checks.

If one of the two servers is offline then the whole traffic should be routed to the other server.

### What did you see instead?
Traefik response with 503 Service unavailable if the request ist dispatched to the offline server.
It seems to me that the health check did not remove the unhealthy server from the wrr service list.

```
Version:      2.3.1
Codename:     picodon
Go version:   go1.15.2
Built:        2020-09-29T15:49:06Z
OS/Arch:      windows/amd64

```

```yaml
# traefik.yaml
entryPoints:
  wsp-1:
    address: ":8080"
  traefik:
    address: ":9090"
  debug: true
  
api:
  dashboard: true
  insecure: true
  debug: true

log:
  filePath: "d:/projects/golang/traefik/traefik.log"
  level: INFO

providers:
  file:
    filename: d:\projects\golang\traefik\dynamic.simple.yaml
    watch: true
```

```yaml
# dynamic.simple.yaml
http:
   routers:
      wsp-router-1:
         rule: PathPrefix(`/`)
         entryPoints:
         - wsp-1
         service: wsp-service
   services:
      wsp-service:
         weighted:
            services:
            -  name: wsp1
               weight: 3
            -  name: wsp2
               weight: 1
      wsp1:
         loadBalancer:
            healthCheck:
               path: /status
               interval: 10s
               timeout: 3s
            servers:
            -  url: http://localhost:8081
      wsp2:
         loadBalancer:
            healthCheck:
               path: /status
               interval: 10s
               timeout: 3s
            servers:
            -  url: http://localhost:8082
```

```
time="2020-12-18T17:39:21+01:00" level=info msg="Traefik version 2.3.1 built on 2020-09-29T15:49:06Z"
time="2020-12-18T17:39:21+01:00" level=info msg="\nStats collection is disabled.\nHelp us improve Traefik by turning this feature on :)\nMore details on: https://doc.traefik.io/traefik/contributing/data-collection/\n"
time="2020-12-18T17:39:21+01:00" level=info msg="Starting provider aggregator.ProviderAggregator {}"
time="2020-12-18T17:39:21+01:00" level=info msg="Starting provider *file.Provider {\"watch\":true,\"filename\":\"d:\\\\projects\\\\golang\\\\traefik\\\\dynamic.simple.yaml\"}"
time="2020-12-18T17:39:21+01:00" level=info msg="Starting provider *traefik.Provider {}"
time="2020-12-18T17:40:03+01:00" level=warning msg="Health check failed, removing from server list. Backend: \"wsp1@file\" URL: \"http://localhost:8081\" Weight: 1 Reason: HTTP request failed: Get \"http://localhost:8081/status\": dial tcp [::1]:8081: connectex: No connection could be made because the target machine actively refused it."
time="2020-12-18T17:40:13+01:00" level=warning msg="Health check still failing. Backend: \"wsp1@file\" URL: \"http://localhost:8081\" Reason: HTTP request failed: Get \"http://localhost:8081/status\": dial tcp [::1]:8081: connectex: No connection could be made because the target machine actively refused it."

```

mpl:
### What does this PR do?

This change adds support for automatic self-healthcheck for the
WeightedRoundRobin type of service, in order to make the whole tree of
load-balancing aware of status changes at the "leaf" level. That is, let the
load-balancing algorithm adjust globally when a server status changes (gets down
or up).

### Motivation

So far healthcheck was supported only at the "leaf" level, i.e. a load-balancer of
servers (at the bottom of the load-balancing tree) was able to do active health
checks on his own servers to adjust its load-balancing algorithm. However when
e.g. all of its servers would go down, there was no propagation upwards of this
status change, which means requests would still arrive to this load-balancer
even though it would in effect be down, and should have been ignored by its
parent(s).

Therefore this change adds a mechanism so that all status changes can be
propagated upwards to let all parents know (and by extension, the whole tree)
when a service is in effect down.

The corresponding configuration change is the introduction of the HealthCheck
option to the WeightedRoundRobin element. When the HealthCheck option is present
in a WeightedRoundRobin, automatic propagation of a status change of its
children is enabled. None of the fields of HealthCheck are relevant in this
context, so they are ignored if present.

### More

- [x] Added/updated tests

### Additional Notes

Fixes #7693 

Co-authored-by: Dmitry Sharshakov <d3dx12.xx@gmail.com>


ddtmachado:
Thanks, I was really eager to try this one ;) 
So far I only found one anomaly that I think it's worth further investigation and maybe a unit test.

Here is the base static config used:

```yaml
http:
  services:
    whoami-public:
      weighted:
        #healthCheck: {}
        services:
        - name: whoami-dc1
          weight: 1
        - name: whoami-dc2
          weight: 1

    whoami-dc1:
      loadBalancer:
        servers:
          - url: "http://127.0.0.1:8090"
          - url: "http://127.0.0.1:8091"
        healthCheck:
          scheme: http
          interval: 2s
          timeout: 1s
          path: /ping

    whoami-dc2:
      loadBalancer:
        servers:
          - url: "http://127.0.0.1:8092"
          - url: "http://127.0.0.1:8093"
        healthCheck:
          scheme: http
          interval: 2s
          timeout: 1s
          path: /ping

  routers:
    dashboard:
      rule: Host(`traefik.docker.localhost`)
      service: api@internal

    whoami-router:
      rule: Host(`whoami.docker.localhost`)
      service: whoami-public
```

Step / Status:

1. ✔️  Setup a WRR service without the `healthcheck` enabled (whoami-public)
2. ✔️ Allow one of the underlying services on it to fail all servers HC (whoami-dc2)
3. ✔️ Enable the `healthcheck` on the root WRR (whoami-public) and let Traefik reload the config
4. ❌ Check the failed service (whoami-dc2) is not getting requests

Even after enabling the `healthcheck` the failed service was still considered during round robin from the root WRR, resulting in `Service Unavailable` errors when sending requests. 

I'm also attaching [logs](https://github.com/traefik/traefik/files/6306425/new-wrr-anomaly.log) starting from the config reload as behavior confirmation.

Other than that, everything work as expected when I start it with `healthcheck` already enabled!

PS.: `whoami` is not actually [whoami](https://github.com/traefik/whoami) here, but a custom backend that allows me to control the `/ping` endpoint

mpl:
> Thanks, I was really eager to try this one ;)
> So far I only found one anomaly that I think it's worth further investigation and maybe a unit test.
> 
> Here is the base static config used:
> 
> ```yaml
> http:
>   services:
>     whoami-public:
>       weighted:
>         #healthCheck: {}
>         services:
>         - name: whoami-dc1
>           weight: 1
>         - name: whoami-dc2
>           weight: 1
> 
>     whoami-dc1:
>       loadBalancer:
>         servers:
>           - url: "http://127.0.0.1:8090"
>           - url: "http://127.0.0.1:8091"
>         healthCheck:
>           scheme: http
>           interval: 2s
>           timeout: 1s
>           path: /ping
> 
>     whoami-dc2:
>       loadBalancer:
>         servers:
>           - url: "http://127.0.0.1:8092"
>           - url: "http://127.0.0.1:8093"
>         healthCheck:
>           scheme: http
>           interval: 2s
>           timeout: 1s
>           path: /ping
> 
>   routers:
>     dashboard:
>       rule: Host(`traefik.docker.localhost`)
>       service: api@internal
> 
>     whoami-router:
>       rule: Host(`whoami.docker.localhost`)
>       service: whoami-public
> ```
> 
> Step / Status:
> 
> 1. ✔️  Setup a WRR service without the `healthcheck` enabled (whoami-public)
> 2. ✔️ Allow one of the underlying services on it to fail all servers HC (whoami-dc2)
> 3. ✔️ Enable the `healthcheck` on the root WRR (whoami-public) and let Traefik reload the config
> 4. ❌ Check the failed service (whoami-dc2) is not getting requests
> 
> Even after enabling the `healthcheck` the failed service was still considered during round robin from the root WRR, resulting in `Service Unavailable` errors when sending requests.
> 
> I'm also attaching [logs](https://github.com/traefik/traefik/files/6306425/new-wrr-anomaly.log) starting from the config reload as behavior confirmation.
> 
> Other than that, everything work as expected when I start it with `healthcheck` already enabled!
> 
> PS.: `whoami` is not actually [whoami](https://github.com/traefik/whoami) here, but a custom backend that allows me to control the `/ping` endpoint

Thanks @ddtmachado , I will investigate.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -81,22 +95,105 @@ func (b *Balancer) Pop() interface{} {
...
+	}
+	if !up {
+		return nil, errNoAvailableServer
+	}
 
-	// Pick handler with closest deadline.
-	handler := heap.Pop(b).(*namedHandler)
+	var handler *namedHandler
+	up = false
+	for !up {
```
mpl, line 184: TODO(mpl): up is useless here

mpl:
> Thanks @ddtmachado , I will investigate.
> Ah I see what you mean, you enabled the healthcheck on the top service after things had already started to fail below. I indeed didn't think of that case.

@ddtmachado , I've updated the PR with a new, simpler, design. From what I've tested it seems it should also fix the problem you reported. But could you please test it out and confirm?

pkg/server/service/service.go
```diff
@@ -161,9 +161,24 @@ func (m *Manager) getWRRServiceHandler(ctx context.Context, serviceName string,
...
+		}
+		childName := service.Name
+		updater, ok := serviceHandler.(healthcheck.StatusUpdater)
+		if !ok {
+			log.FromContext(ctx).Debugf("Child service %v will not propagate to its parent %v any status change", service.Name, serviceName)
+			continue
+		}
+		log.FromContext(ctx).Debugf("Child service %v will update parent %v on status change", service.Name, serviceName)
+		// TODO(mpl): do we make RegisterStatusUpdater return an error (e.g. in case the
+		// updater actually does not want healthcheck enabled, according to its own config)
```
mpl, line 176: As of now (but still TBD), we probably want to go that way, and straight up return here the value returned by RegisterStatusUpdater. This should make the service creation fail, which should be visible e.g. in the dashboard.

ddtmachado:
@mpl just tested again and I confirm it does work for that edge case as well, nice work!

pkg/healthcheck/healthcheck.go
```diff
@@ -277,26 +284,69 @@ func NewLBStatusUpdater(bh BalancerHandler, info *runtime.ServiceInfo) *LbStatus
...
+		return err
+	}
+	if lb.serviceInfo != nil {
 		lb.serviceInfo.UpdateServerStatus(u.String(), serverDown)
 	}
-	return err
+
+	if !upBefore {
+		// we were already down, and we still are, no need to propagate.
+		log.WithoutContext().Debugf("child %s now DOWN, but we were already DOWN, so no need to propagate.", u.String())
```
ddtmachado, line 311: maybe we could add context for the log messages?
IMO it's easier to follow when there are other log messages in between

mpl: yes, I also already had the same dilemma with SetStatus (in wrr.go), where the only reason I pass balancerName as an argument is for better logging. Julien and I were considering passing a context instead (with the balancerName inside it), but it didn't seem much better.
But yeah, we probably should. Adding a TODO for now.
Ah, and in the case of RemoveServer (and UpsertServer) maybe we can't even do that, because the signatures have to match the ones from oxy. I have to double-check.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -81,22 +91,81 @@ func (b *Balancer) Pop() interface{} {
...
+		}
+		return
+	}
+
+	if len(b.status) > 0 {
+		// we were up before and we still are, no need to propagate
+		log.WithoutContext().Debugf("No need to propagate that %s is still UP", balancerName)
+		return
+	}
+	// propagate upwards that we are now DOWN
```
tomMoulard, line 128:
```suggestion
	}

	// propagate upwards that we are now DOWN
```

pkg/server/service/service.go
```diff
@@ -161,9 +161,24 @@ func (m *Manager) getWRRServiceHandler(ctx context.Context, serviceName string,
 		if err != nil {
 			return nil, err
 		}
-
 		balancer.AddService(service.Name, serviceHandler, service.Weight)
```
tomMoulard, line 164:
```suggestion

		balancer.AddService(service.Name, serviceHandler, service.Weight)
```

pkg/server/service/service.go
```diff
@@ -161,9 +161,24 @@ func (m *Manager) getWRRServiceHandler(ctx context.Context, serviceName string,
...
+		if config.HealthCheck == nil {
+			continue
+		}
+		childName := service.Name
```
tomMoulard, line 168:
```suggestion
		}

```

pkg/server/service/service.go
```diff
@@ -161,9 +161,24 @@ func (m *Manager) getWRRServiceHandler(ctx context.Context, serviceName string,
...
+		updater, ok := serviceHandler.(healthcheck.StatusUpdater)
+		if !ok {
+			log.FromContext(ctx).Debugf("Child service %v will not propagate to its parent %v any status change", service.Name, serviceName)
+			continue
+		}
+		log.FromContext(ctx).Debugf("Child service %v will update parent %v on status change", service.Name, serviceName)
+		// TODO(mpl): do we make RegisterStatusUpdater return an error (e.g. in case the
+		// updater actually does not want healthcheck enabled, according to its own config)
+		updater.RegisterStatusUpdater(func(up bool) {
+			balancer.SetStatus(serviceName, childName, up)
```
tomMoulard, line 178:
```suggestion
			balancer.SetStatus(serviceName, service.Name, up)
```

mpl: nope, it's very important to not change that part. I will let you think about it for a while ;)

pkg/healthcheck/healthcheck.go
mpl: yeah, as I was suspecting, at the moment we're kindof blocked because of the fixed signatures of RemoveServer and UpsertServer, because they are coming from oxy. It might be doable to work-around that, but I didn't want to delay the PR further with that. Especially since we hope to rework the part depending on oxy soon.
So for now I half-way addressed your suggestion: the code within the method already uses a (useless for now) context, but we'll pass around as argument the actual context later.

rtribotte (approved):
LGTM

ddtmachado (approved):

PR #8072, opened 2021-04-15
doc: clarify usage for ratelimit's excludedIPs

### LINKED ISSUE #7884: no strategy fallback in case of X-Forwarded-For's absence
ok, I'm on ur latest stable `v2.4`... seems that for traffic originated from the same subnet (in my case I have kubernetes nodes with an F5 in front) ...the traffic which is going from the cluster out, and back from the F5 into the traefik setup is not getting the `X-Forwarded-For` header... therefore, cause I have an `excludedIPs` list (with the external CDNs IPs in it) I'm into the ipWhitelisting middleware... with the `CheckerStrategy` ... there is no fallback to the `RemoteAddrStrategy`

I wrote a unit test for it... I wonder if I shall make a patch or simply got for a Middleware to check the existence of the header first

```golang
+++ b/pkg/ip/strategy_test.go
@@ -78,6 +78,7 @@ func TestExcludedIPsStrategy_GetIP(t *testing.T) {
        testCases := []struct {
                desc          string
                excludedIPs   []string
+               remoteAddr    string
                xForwardedFor string
                expected      string
        }{
@@ -93,6 +94,12 @@ func TestExcludedIPsStrategy_GetIP(t *testing.T) {
                        xForwardedFor: "10.0.0.4,10.0.0.3,10.0.0.2,10.0.0.1",
                        expected:      "10.0.0.3",
                },
+               {
+                       desc:          "Use excluded IPs, but X-Forwarded-For is not present",
+                       excludedIPs:   []string{"10.0.0.2", "10.0.0.1"},
+                       remoteAddr:    "10.2.3.1:123",
+                       expected:      "10.2.3.1",
+               },
                {
                        desc:          "Use excluded IPs CIDR",
                        excludedIPs:   []string{"10.0.0.1/24"},
@@ -117,7 +124,9 @@ func TestExcludedIPsStrategy_GetIP(t *testing.T) {
 
                        strategy := CheckerStrategy{Checker: checker}
                        req := httptest.NewRequest(http.MethodGet, "http://127.0.0.1", nil)
-                       req.Header.Set(xForwardedFor, test.xForwardedFor)
+                       if test.xForwardedFor != "" {
+                               req.Header.Set(xForwardedFor, test.xForwardedFor)
+                       }
                        actual := strategy.GetIP(req)
                        assert.Equal(t, test.expected, actual)
                })
```
https://github.com/zeph/traefik/commit/fb002c2106e32ceae2e415e6ad300686823e1316

### LINKED ISSUE #8052: Rate limiting with excluded IPs doesn't use request.RemoteAddr

Bug

### What did you do?

Tried to set a ratelimit that uses the IPStrategy and ues the remote IP address, since there is no `X-Forwarded-For` header in our setup (Traefik is listening to requests directly).

Exhausing the rate limit from one IP should not affect the limit of any other IP, and `123.123.123.123` should be excluded from the limit entirely.

### What did you see instead?

When the `ratelimit.sourcecriterion.ipstrategy.excludedips` is set, Traefik only considers the value of `X-Forwarded-For`
It took a bit of debugging and adding log statements all over the place, but it turns out that the `source` at https://github.com/traefik/traefik/blob/v2.4.7/pkg/middlewares/ratelimiter/rate_limiter.go#L108 is just an empty string as soon as you set the excluded IPs. That means every request shares the same bucket.

The cause for this is most likely found at these places:

https://github.com/traefik/traefik/blob/v2.4.7/pkg/config/dynamic/middlewares.go#L262
https://github.com/traefik/traefik/blob/v2.4.7/pkg/ip/strategy.go#L58

There are two tests for the excluded IP behavior at https://github.com/traefik/traefik/blob/v2.4.7/pkg/ip/strategy_test.go but they both only test by setting the header.

```
Version:      2.4.7
Codename:     cheddar
Go version:   go1.16.2
OS/Arch:      linux/amd64
```

Following configs were set via Consul:

```
traefik.enable=true
traefik.http.routers.testing.rule=Host(`example.com`)
traefik.http.routers.testing.entrypoints=https
traefik.http.routers.testing.tls=true
traefik.http.routers.testing.middlewares=testing-ratelimit@consulcatalog
traefik.http.middlewares.testing-ratelimit.ratelimit.average=100
traefik.http.middlewares.testing-ratelimit.ratelimit.burst=250
traefik.http.middlewares.testing-ratelimit.ratelimit.period=1m
traefik.http.middlewares.testing-ratelimit.ratelimit.sourcecriterion.ipstrategy.excludedips=123.123.123.123
```

mpl:
### What does this PR do?

This change clarifies the documentation of the excludedIPs option for the ratelimit middleware.

### Motivation

Inspired by discussion from  #8052 

### More

- [x] Added/updated tests

### Additional Notes

Fixes #8052


jbdoumenjou (approved):
LGTM :+1:

pkg/ip/strategy_test.go
```diff
@@ -74,49 +74,81 @@ func TestDepthStrategy_GetIP(t *testing.T) {
...
-			checker, err := NewChecker(test.excludedIPs)
+			checker, err := NewChecker(test.trustedIPs)
 			require.NoError(t, err)
 
-			strategy := CheckerStrategy{Checker: checker}
+			strategy := PoolStrategy{Checker: checker, UseRemote: test.useRemote}
 			req := httptest.NewRequest(http.MethodGet, "http://127.0.0.1", nil)
+			if test.useRemote {
+				req.RemoteAddr = "127.0.0.1"
+			}
```
rtribotte, line 151:
```suggestion
			req.RemoteAddr = "127.0.0.1"
```

I think it's better to always have the remoteAddr on the request, whatever the config.

mpl:
Moving to no merge for now, because we're considering changing our minds and just make this PR about better doc.

jbdoumenjou (approved):
LGTM :+1:

rtribotte (approved):
LGTM 👌

dtomcej (approved):
LGTM
:shipit:

PR #8082, opened 2021-04-19
Adds support for the inclusion of arbitrary tags in influxDB metrics

jimnelson2:
### What does this PR do?

This change adds support for the inclusion of arbitrary tags in influxDB metrics.

### Motivation

As described in https://github.com/traefik/traefik/issues/6301, the inclusion of additional tags
allows users to classify their metrics into useful categories for analysis such as environment or
region. This is critical in larger deployments across multiple regions where no other mechanism
exists to segment metrics into such categories.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/observability/metrics/influxdb.md
+++ b/docs/content/observability/metrics/influxdb.md
@@ -213,3 +213,29 @@ metrics:
 ```bash tab="CLI"
 --metrics.influxdb.pushInterval=10s
 ```
+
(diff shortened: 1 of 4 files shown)


kevinpollet:
Hello @jimnelson2 and thanks for your interest in Traefik,

We just merged https://github.com/traefik/traefik/pull/8308 which does what you want.

PR #8107, opened 2021-04-29
Upgrade Node version to LTS on webui folder

leonardxfce:
### What does this PR do?

Bump node version to the latest LTS version based on the docker-hub tag

### Motivation

The v12 release of Node entered a maintenance status. As a good practice, the base image should be moved to the next active LTS version
https://nodejs.org/en/about/releases/

### DIFF
--- a/webui/Dockerfile
+++ b/webui/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.18
+FROM node:lts
 
 ENV WEBUI_DIR /src/webui
 ARG ARG_PLATFORM_URL=https://pilot.traefik.io
--- a/webui/Dockerfile
+++ b/webui/Dockerfile
@@ -1,4 +1,5 @@
-FROM node:lts
+FROM node:14.16
+# Current Active LTS release according to (https://nodejs.org/en/about/releases/)
 
 ENV WEBUI_DIR /src/webui
 ARG ARG_PLATFORM_URL=https://pilot.traefik.io


dtomcej:
Hello @leonardxfce,

Thanks for raising this issue. According to the link you referred to, Production systems *should* be using either active or maintenance LTS releases (of which v12 is currently in maintenance).

Do you think that it would be more appropriate to jump from v12 to v16 in the new year (Jan/Feb 2022) to prevent having to leapfrog smaller versions?

Also, we try to avoid rotating tagged images, like `:latest`, as we have no control over when they get changed. Do you think that it would be better to have something like this instead?

```dockerfile
FROM node:14.16
# Current Active LTS release according to (https://nodejs.org/en/about/releases/)
```

leonardxfce:
Sound good to me. I can do the change to a more specific image. Give me some mins.

leonardxfce:
Pushed. Let me know if looks good for you. If not we can delay this kind of PR to next LTS release

dtomcej:
Hello @leonardxfce,

Is it possible that you may have reverted a commit when you rebased? It appears that your change did not get pushed.

Thanks!

dtomcej (approved):
LGTM
:shipit:

rtribotte (approved):
LGTM 👌

jbdoumenjou (approved):
LGTM :+1:

PR #8110, opened 2021-05-02
Fix tracing documentation for elastic serverUrl

L3o-pold:
### What does this PR do?

Fix documentation for tracing option with elastic provider.

### Motivation

when using quotes, Traefik will fail to parse serverUrl.

```bash
time="2021-05-02T21:02:44Z" level=warning msg="Unable to create tracer: parse \"\\\"http://apm:8200\\\"\": first path segment in URL cannot contain colon"
```

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/observability/tracing/elastic.md
+++ b/docs/content/observability/tracing/elastic.md
@@ -35,7 +35,7 @@ tracing:
 ```
 
 ```bash tab="CLI"
---tracing.elastic.serverurl="http://apm:8200"
+--tracing.elastic.serverurl=http://apm:8200
 ```
 
 #### `secretToken`


ldez:
Hello,

The documentation is right: the quotes are ignored when you use the binary into a shell, they are even required in some cases.

L3o-pold:
@ldez using docker compose for this and it fails. But if you estimate that the documentation is good let's close this.

ldez:
In some specific context like a Docker compose the quotes produce an error, but it's related to the constraints of Docker Compose itself.
If you are in a shell the quotes can be required, but if you are in a Docker Compose file they are a problem :shrug: 
So there is no good way to document that.

PR #8111, opened 2021-05-04
fix: use defaultEntryPoints when no entryPoint is defined in a TCPRouter

LandryBe:
### What does this PR do?

This PR updates `mergeConfiguration` to using `defaultEntrypoints` when there is no entryPoint set on a TCPRouter.

### Motivation

When no entryPoint is defined for a TCP router, this router uses all the entryPoints defined, including the `Traefik` one. As a result, the API is broken.

### More

- [x] Added/updated tests


jbdoumenjou (approved):
LGTM :+1:

rtribotte (approved):
LGTM 👍

ldez (approved):
LGTM

PR #8120, opened 2021-05-06
Fix ingressRouteTCP external name service examples in documentation

### LINKED ISSUE #8112: [Documentation] IngressRoute with ExternalName service shows example without mandatory 'port' field
I think this is an error in the documentation, even though I wish it would work as shown there :)

Please have a look at this screenshot of the current documentation for the IngressRoute resource found at https://doc.traefik.io/traefik/routing/providers/kubernetes-crd/#kind-ingressroute

grafik

This example shows the declaration of an IngressRoute to a service of type "ExternalName" with no port number:
```
kind: IngressRoute
metadata:
  name: test.route
  namespace: default

spec:
  entryPoints:
    - foo

  routes:
  - match: Host(`example.net`)
    kind: Rule
    services:
    - name: external-svc
```

This seems to be an error, because the field `spec.routes[].services[].port` seems to be mandatory, even if the declared `Service` already defines its port. The text above the example is correct, but the example itself seems to be wrong.

rtribotte:
### What does this PR do?

This PR removes an unsupported example in the `IngressRouteTCP` documentation.

### Motivation

Fixes #8112

### More

- [ ] ~Added/updated tests~

### DIFF
--- a/docs/content/routing/providers/kubernetes-crd.md
+++ b/docs/content/routing/providers/kubernetes-crd.md
@@ -1202,7 +1202,7 @@ Register the `IngressRouteTCP` [kind](../../reference/dynamic-configuration/kube
     
     ??? example "Examples"
         
-        ```yaml tab="Only on the IngressRouteTCP"
+        ```yaml tab="Only on IngressRouteTCP"
         ---
         apiVersion: traefik.containo.us/v1alpha1
         kind: IngressRouteTCP
--- a/docs/content/routing/providers/kubernetes-crd.md
+++ b/docs/content/routing/providers/kubernetes-crd.md
@@ -1202,7 +1202,7 @@ Register the `IngressRouteTCP` [kind](../../reference/dynamic-configuration/kube
     
     ??? example "Examples"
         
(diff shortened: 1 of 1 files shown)


docs/content/routing/providers/kubernetes-crd.md
```diff
@@ -1202,7 +1202,7 @@ Register the `IngressRouteTCP` [kind](../../reference/dynamic-configuration/kube
     
     ??? example "Examples"
         
-        ```yaml tab="IngressRouteTCP"
+        ```yaml tab="Only on the IngressRouteTCP"
```
jbdoumenjou, line 1205:
```suggestion
        ```yaml tab="Only on IngressRouteTCP"
```

jbdoumenjou (approved):
LGTM

ChristianCiach:
This should also be fixed for the `IngressRoute` CRD documentation, not just the `IngressRouteTCP` CRD. In fact, in my issue #8112 i actually had the `IngressRoute` documentation in mind.

Issue #8123, opened 2021-05-06
Allow assuming default host if no SNI indication is given (for Let's Encrypt)


akx:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Following up on https://github.com/traefik/traefik/issues/2829, I have a situation where I'd like to proxy Redis (TCP, non-HTTP) connections over TLS with automatic Let's Encrypt certificates. (I am aware that Redis 6 has built-in TLS, but it doesn't have built-in Let's Encrypt 😄)

Redis clients in general do not send SNI indications, so one needs a HostSNI(`*`) on the router for it to basically do any routing. However, with that enabled, Traefik always serves the built-in default certificate since there's no other indication on the connection. `openssl s_client` connects nicely, and I can talk Redis to the underlying server, but the certificate is `CN = TRAEFIK DEFAULT CERT`, of course.

### Takeaway

It would be handy to be able to tell Traefik to assume a given fallback SNI name if none was sent by the peer, e.g. (c.f. the below example configuration)

```toml
[tcp.routers.redis-tls-router.tls]
fallbackSNI = "some-example-domain.example.com"
```

### Testing configuration

My very simple file-based testing configuration is (with DNS names censored)

#### static configuration

```toml
[log]
level = "DEBUG"

[entryPoints]
web = { address = ":80" }
redis-tls = { address = ":55555" }

[certificatesResolvers.le.acme]
email = "example@example.com"
storage = "acme.json"
caServer = "https://acme-staging-v02.api.letsencrypt.org/directory"

[certificatesResolvers.le.acme.httpChallenge]
entryPoint = "web"

[providers.file]
filename = "/root/traefik.dyn.toml"
```

#### traefik.dyn.toml
```toml
[tcp.routers.redis-tls-router]
entryPoints = [ "redis-tls" ]
rule = "HostSNI(`*`)"
service = "redis"

[tcp.routers.redis-tls-router.tls]
certResolver = "le"

[[tcp.routers.redis-tls-router.tls.domains]]
main = "some-example-domain.example.com"

[tcp.services.redis.loadBalancer]
servers = [ { address = "127.0.0.1:6379" } ]
```

adyanth:
Refer https://community.traefik.io/t/tls-options-for-snistrict-false-in-tcp-router/4626

adyanth:
The current behaviour is very bad, I had to enable strictSNI to see this error and understand it was considering the domain as empty.

```
[...]

PR #8124, opened 2021-05-06
Implement SNI Fallback in TLS settings

akx:
### What does this PR do?

This implements an optional `sniFallback` field in TLS options, akin to what I requested in #8123.

It can be used in conjunction with SNI-less requests and Let's Encrypt certificates, to blindly assume a request that has no other matching certificate to use a named one.

### Motivation

Please see #8123.

### More

- [ ] Added/updated tests
  - Not yet! Not sure how to best test this, I'm not an experienced gopher.

### DIFF
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -366,6 +366,40 @@ spec:
   sniStrict: true
 ```
 
+### SNI Fallback
+
+SNI Fallback allows TLS connections where the client does not send an SNI name indication,
+or if the SNI name indicated matched no known certificate, to use a fallback certificate,
+as if the client had originally indicated that server name to be used.
+
+```toml tab="File (TOML)"
+# Dynamic configuration
+
+[tls.options]
+  [tls.options.default]
+    sniFallback = "my-service.example.com"
+```
+
+```yaml tab="File (YAML)"
+# Dynamic configuration
+
+tls:
+  options:
+    default:
+      sniFallback: "my-service.example.com"
+```
+
+```yaml tab="Kubernetes"
+apiVersion: traefik.containo.us/v1alpha1
+kind: TLSOption
+metadata:
+  name: default
+  namespace: default
+
+spec:
+  sniFallback: "my-service.example.com"
+```
+
 ### Prefer Server Cipher Suites
 
 This option allows the server to choose its most preferred cipher suite instead of the client's.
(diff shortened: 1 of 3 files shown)


akx:
The Travis build failure seems to be a fluke:

```
3.13: Pulling from library/alpine
age configuration: unexpected EOF
make[1]: *** [docs-pull-images] Error 123
make[1]: Leaving directory `/home/travis/build/traefik/traefik/docs'
```

dtomcej:
Hello @akx,

How does this differ from using the defaultCertificate for non-matching requests? (https://doc.traefik.io/traefik/https/tls/#default-certificate)

akx:
> How does this differ from using the defaultCertificate for non-matching requests? (https://doc.traefik.io/traefik/https/tls/#default-certificate)

Hi @dtomcej – as discussed in https://github.com/traefik/traefik/issues/2829#issuecomment-364450184 `defaultCertificate`s need to be on the file system. This works for certs issued by LE within Traefik.

akx:
Could I get e.g. @SantoDE to take a look at this? (I see you'd assigned yourself for #8123)

stacab:
Hi,
I need this feature so much in Traefik. Any chance to have it reviewed soon at least?

akx:
Rebased on `master`.

akx:
Rebased on `master` once again.

@jbdoumenjou Is there anything I can do to move this forward?

jbdoumenjou:
Hello @akx,
 
Thanks for your contribution!

akx:
@jbdoumenjou Hi there! Sorry to keep bothering you – how's the design review process coming along? I don't think this is a very large change architecturally, and it's really a last-chance fallback thing...

akx:
@jbdoumenjou Any news? 🍂  I rebased this branch for your convenience.

adyanth:
Would love to see this pulled in too. Here is another request on community https://community.traefik.io/t/tls-options-for-snistrict-false-in-tcp-router/4626

I have a need for this to implement DNS over TLS since not even modern smartphones send SNI when doing DoT requests.

rtribotte:
Hello @akx,

While we definitely see the benefit of your changes to handle SNI-less requests, we were hesitant about the solution and we finally decided to decline the PR, and here's why:

We acknowledge that what we already have (i.e. to let the default cert being selected when no cert matches the SNI) feels a bit uneasy, as it seems to go against the intent of the SNI concept. Which is exactly why we have the sniStrict option by the way. In that regard, we do not like the idea of going further in that "wrong" direction, i.e. to widen the pool of possible certs when SNI does not match. Which is why we'd rather go in a subtly different direction (to achieve a similar goal): we won't change the fact that we get the default cert when there is no SNI match, but we'll expand on the mechanisms that can (dynamically) generate the (unique) default cert.

To that end, we've already started investigating a way to generate the default cert through Let's Encrypt, and we'll keep you posted on how that goes.

adyanth:
@rtribotte 
>we won't change the fact that we get the default cert when there is no SNI match, but we'll expand on the mechanisms that can (dynamically) generate the (unique) default cert.

Does this mean that if an HTTPS request comes in without SNI, a Let's Encrypt cert can be served?

Also, is there a place we can track these changes?

Issue #8130, opened 2021-05-09
Make HTTP3 port configurable


valerauko:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

It would be nice if the HTTP3 port could be configured.

#### Reason

When Traefik 2 is installed with [the helm chart](https://github.com/traefik/traefik-helm-chart), the address of entryPoints is _on the container_ -- eg web runs on 8000 and websecure on 8443, and are exposed as 80 and 443 by Kubernetes. If I enable HTTP3 on websecure then, the alt-svc header advertises HTTP3 on port 8443 instead of the actually publicly exposed 443.

This can be worked around by running Traefik privileged and using ports 80 and 443 directly, but I think making the HTTP3 port configurable is better (I'd prefer to run stuff with as little privileges as possible). Maybe as `entrypoints.<name>.http3.address`?

starcraft66:
Definitely looking forward to this.

starcraft66:
This issue should be renamed to "Make advertised HTTP/3 port configurable".

To clarify, without this being configurable, it is annoying to run traefik as a non-root user or without additional capabilities allowing it to bind to ports under 1024. The default configuration in the Traefik Helm chart makes it run as a non-root user and sets the default entrypoints to 8000 for http and 8443 for https, those ports can then be remapped to 80 and 443 using a kubernetes Service.

The problem with this is that enabling http/3 will cause the `alt-svc` header to announce the 8443 port instead of the 443 port remapped using a service. I don't think that changing the bind port of the http/3 service is useful. Instead the value advertised in the `alt-svc` header should be configurable.

PR #8131, opened 2021-05-10
Add HTTP3Config

valerauko:
### What does this PR do?

* moves an EntryPoint's enableHTTP3 flag into a HTTP3Config struct under the http3 key
* adds option to configure what to advertise as the http3 authority

### Motivation

#8130 

### More

- [ ] Added/updated tests

### Additional Notes

I saw that the Docker integration tests are failing, but I can't figure out what is causing the issue. If someone could provide some feedback I'll fix it...

- [x] make pull-images
- [x] make validate
- [x] make test-unit
- [ ] make test-integration

### DIFF
--- a/docs/content/migration/v2.md
+++ b/docs/content/migration/v2.md
@@ -415,3 +415,12 @@ For more advanced use cases, you can use either the [RedirectScheme middleware](
 Following up on the deprecation started [previously](#x509-commonname-deprecation),
 as the `x509ignoreCN=0` value for the `GODEBUG` is [deprecated in Go 1.17](https://tip.golang.org/doc/go1.17#crypto/x509),
 the legacy behavior related to the CommonName field can not be enabled at all anymore.
+
+## v2.5 to v2.6
+
+### HTTP3
+
+Traefik v2.6 introduces the `AdvertisedPort` option,
+which allows advertising, in the `Alt-Svc` header, a UDP port different from the one on which Traefik is actually listening (the EntryPoint's port).
+By doing so, it introduces a new configuration structure `http3`, which replaces the `enableHTTP3` option (which therefore doesn't exist anymore).
+To enable HTTP3 on an EntryPoint, please check out the [HTTP3 configuration](../routing/entrypoints.md#http3) documentation.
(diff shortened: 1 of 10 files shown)


rtribotte:
Hello @valerauko,

I see that the address specified with the new option `advertiseAs` is used to set up the `http.Server`.
Thus, I don't think we want to effectively change the port Traefik will be listening on, but just advertise a different port in the `Alt-Svc` header when the port is not public.

Also since the only need is to configure an alternate port value, I think that the option should be a port value.

WDYT?

rtribotte:
@valerauko  So if I get you, it's not advisable to advertise a different port. But then, why not solve this issue by choosing a port for Traefik `EntryPoint` that matches the public port?
Then we wouldn't need this option.
From what I understand, having such an option should be used indeed to advertise a different port.

valerauko:
In the case of the traefik helm chart, the container port (by default 8443 on websecure) is different from the exposed port (443 by default). Without this option, traefik will advertise the incorrect 8443 instead of the exposed 443 it should.

rtribotte:
Hello @valerauko,

I reworked the PR to only allow the configuration of the port to advertise in the `Alt-Svc` header.

To test it with k3d, here's an example repo: https://github.com/rtribotte/traefik-k8s-http3-example.

valerauko:
I see you made it way stricter with only being able to advertise the port. If you think that's more desirable, I have no objections.

sylr:
If memory serves me well I had troubles with kubernetes because services backed by AWS ELB do not support both UDP & TCP at the same time.

This means you if you want both HTTP and HTTP3 you shall have 2 different loadbalancers with 2 different IP adresses, one listening on :443/tcp and the other on :443/udp.

This means if you want to server H3 on the same port as HTTP there is a need to advertise H3 on a different IP address.

rtribotte:
@sylr We have to confirm that, but yes that would be a limitation.
Thus, I think that this feature would be a good first step, as this limitation is another use case that could be discussed/handled by another issue.

With this PR, if the constraint is having two AWS ELB which cannot be configured on the same port with different protocols,
a solution would be to create two EntryPoints with HTTP3 enabled.
The first EntryPoint would handle HTTP/2 connections and advertise the port of the second EntryPoint for HTTP/3 connections, while the second EntryPoint would handle HTTP/3 connections.

juliens (approved):
LGTM

ldez (approved):
Thank you

jbdoumenjou (approved):
LGTM :+1:

PR #8136, opened 2021-05-14
Fix plugin unzip call on windows

ddtmachado:
### What does this PR do?

Fix the unzip call for plugins when running on windows, or any other platform that does not use a *nix path separator.

### Motivation

It seems that the zip.FileHeader does not care about os specific file separator and instead always uses forward slashes. So to run on windows or other platforms with different path separators we cannot rely on os.PathSeparator when using its file name reference.

### More

- [ ] ~~Added/updated tests~~

### DIFF
--- a/pkg/plugins/client.go
+++ b/pkg/plugins/client.go
@@ -281,7 +281,7 @@ func unzipFile(f *zipa.File, dest string) error {
 
 	defer func() { _ = rc.Close() }()
 
-	pathParts := strings.SplitN(f.Name, string(os.PathSeparator), 2)
+	pathParts := strings.SplitN(f.Name, "/", 2)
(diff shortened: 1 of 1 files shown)


dtomcej:
Interesting, and that there is not real option to configure a path separator in the FileHeader:

From (https://golang.org/pkg/archive/zip/#FileHeader):

```
    // It must be a relative path, not start with a drive letter (such as "C:"),
    // and must use forward slashes instead of back slashes. A trailing slash
    // indicates that this file is a directory and should have no data.
```

I guess there isn't any harm in hardcoding it if that is what is expected!

dtomcej (approved):
LGTM
:shipit:

ldez (approved):
LGTM

rtribotte (approved):
LGTM

Issue #5472, opened 2019-09-23
Provide dnsChallenge credentials per resolver


bstuff:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

multiple resolvers is 🔥
but what if I have multiple accounts

something like this:
```
[certificatesResolvers.sample1.acme]
  email = "info@example.com"
  storage = "acme/acme.json"
  [certificatesResolvers.sample1.acme.dnsChallenge]
    provider = "cloudflare"
    <API_KEY> = <.......>
[certificatesResolvers.sample2.acme]
  email = "info2@example.com"
  storage = "acme/acme.json"
  [certificatesResolvers.sample2.acme.dnsChallenge]
    provider = "cloudflare"
    <API_KEY> = <.......>
```

use case: Traefik running multiple domains that belongs to different accounts on Cloudflare

dduportal:
Hi @bstuff , thanks for reporting this use case. This issue is labelled as "proposal" as there should be a discussion on the topics below:

It looks like that implementing this would require the following challenge to be solved:

- Disparity of behavior between DNS providers: using token, API key, instance profiles or any other way for using account credentials depends on the provider. This could cause a lot of complexity.
- Some providers make it really hard for this. Example with AWS where the client has to manage credentials between API keys, instance profiles and IAM roles. Profiles should be used in this case, but only in this case.

Finally, please note that this proposal would lead to some changes in Lego (the library used for ACME usage in Traefik), not only on Traefik.

orblazer:
Hello, something is new for that proposal ?

I ask because i want setup traefik with multiple domains but i have multiple credentials :(
For now i am oblige to manage certificates with certbot externally to traefik.

Thanks.

er1z:
May be a solution is to run the `lego` as a separated process? It would be really nice to have separated credentials per domain…

debilausaure:
I have a specific use case for this, which I hope will bring interest to this feature request.

I share a server with someone, and we have two domains, one each. We use Traefik to redirect users to our respective containerized services and provide the respective certificates, which are both **wildcard** certs. 

It'd be a great value to add this feature to Traefik, since I think it matches its philosophy quite well.
Being able to configure secrets per provider would also be quite nice.

#5632 seems related

debilausaure:
Here is an idealized `Dockerfile` that mimics the actual [DNS challenge user guide](https://docs.traefik.io/v2.0/user-guides/docker-compose/acme-dns/).

```
[...]

PR #8140, opened 2021-05-17
[WIP] Configure DNSChallenge provider using environment variables passed through config.

IlyaGulya:
### What does this PR do?

Adds the ability to apply multiple configurations to DNS challenge providers.
It allows passing specific environment variables during the creation of the DNS challenge provider.

Those variables are placed under `certificatesResolvers.<name>.acme.dnsChallenge.environment`

Example:
```
[certificatesResolvers.CertificateResolver1.acme.dnsChallenge.environment]
  foobar = "foobar"
```
Environment variable `foobar` with the value `foobar` will be passed to the DNS challenge provider during its creation.

This is done by overriding environment variables during provider creation and restoring the previous environment status after it.

### Motivation

I wanted the ability to pass different configurations to different providers because I own several domains and my DNS provider generates different API keys for different domains.
Also, there's an issue that is not resolved yet: https://github.com/traefik/traefik/issues/5472

### More

- [x] Added tests for dns challenge provider configuration
- [x] Updated documentation

### Additional Notes

I'm completely aware that this implementation is quite hacky.
The right solution will require huge PR to https://github.com/go-acme/lego for it to support configuration through code.

### DIFF
--- a/docs/content/reference/static-configuration/cli-ref.md
+++ b/docs/content/reference/static-configuration/cli-ref.md
@@ -63,6 +63,9 @@ Assume DNS propagates after a delay in seconds rather than finding and querying 
 `--certificatesresolvers.<name>.acme.dnschallenge.disablepropagationcheck`:  
 Disable the DNS propagation checks before notifying ACME that the DNS challenge is ready. [not recommended] (Default: ```false```)
 
+`--certificatesresolvers.<name>.acme.dnschallenge.environment.<name>`:  
+Optional environment variables to override before running lego for ACME challenge.
+
 `--certificatesresolvers.<name>.acme.dnschallenge.provider`:  
 Use a DNS-01 based challenge provider rather than HTTPS.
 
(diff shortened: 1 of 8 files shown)


ldez:
Hello,

As explained here https://github.com/traefik/traefik/issues/5472#issuecomment-534908109, this approach will not really work.

And there is a concurrency issue.

I already declined PRs with this kind of approach in lego.

So we will decline.

IlyaGulya:
> Hello,
> 
> As explained here [#5472 (comment)](https://github.com/traefik/traefik/issues/5472#issuecomment-534908109), this approach will not really work.
> 
> And there is a concurrency issue.
> 
> I already declined PRs with this kind of approach in lego.
> 
> So we will decline.

According to my small research, every single DNS provider in `lego` currently uses environment variables to configure itself.
So, this approach does really work. And it works for each of supported providers in `lego`.

Until there no other ways to configure providers in `lego`, I think this is not that bad way to fix this issue.

ldez:
I'm the main maintainer of lego, according to my knowledge, every single DNS provider doesn't use env vars, sorry.

IlyaGulya:
> I'm the main maintainer of lego, according to my knowledge which is exhaustive, every single DNS provider doesn't use env vars, sorry.

Ok.
I've looked over every provider before writing previous message and haven't seen any which can't be configured using env vars :smile: 
Maybe I've missed one or two, my fault. (you could have pointed on one of them for this conversation to be more productive, btw)

Nevertheless, my approach adds a configuration entry called `environment`.
I think it's quite obvious for consumers that this feature allows to configure only those DNS providers which can be configured using environment variables.

What's wrong with ability to solve several users' issue?
This change does not seem to impact on any other traefik subsystems.
It's very easy and clear.
It would be easy to remove once proper solution will be implemented.

Current situation involves a lot of unneeded complexity for cases which does not require it.

PR #8143, opened 2021-05-17
require that secret is a valid PEM

### LINKED ISSUE #8142: Kubernetes provider fails when a secret is referenced which contains DER encoded TLS data.

Bug

### What did you do?
Created a secret containing a key and certificate encoded in DER format. Then reference this secret in an ingress resource to provide TLS configuration. 

As traefik only seems to support PEM formatted x509 certificates and keys; I would expect to see the ingress not be configured properly. Perhaps, an error message and a warning about the ingress not being properly configured for TLS. 

### What did you see instead?
Once the DER encoded TLS secret is referenced by the bad ingress, the provider will fail to reconcile all state changes in the cluster with the runtime configuration. This manifest first with a parser error:

```
{"level":"error","msg":"Near line 31937 (last key parsed 'tls.certificate.certFile'): unexpected EOF","time":"2021-04-30T19:08:09Z"}
```
This is causes `nil` to be passed from loadConfig here: 

https://github.com/traefik/traefik/blob/a0d8ee5a0252205e378a9a80d7ddc54483a5f03c/provider/kubernetes/kubernetes.go#L488-L495

The failure occurs here:

https://github.com/traefik/traefik/blob/v1.7/provider/kubernetes/kubernetes.go#L823-L831

Where the binary data is encoded as a string and then injected into the toml document. Subsequent calls to load the configuration from the template data fail. 

This condition will persist until the faulty tls data is removed or the service is restarted. However, all reconciliation of ingress, services, endpoints, and secrets will continue to fail.

```
Version:      v1.7.24
Codename:     maroilles
Go version:   go1.14.1
Built:        2020-03-25_04:34:11PM
OS/Arch:      linux/amd64
```

```toml
```

```
(paste your output here)
```

jr0d:
### What does this PR do?

If a user creates a secret with binary encoded key and certificate data (DER), it will cause `p.loadConfig` to fail to parse the template configuration generated from `p.loadIngresses`.

### Motivation

I have not verified if the k8s provider in traefik-2 is susceptible to this behavior, but will be happy to submit a new PR for 2.x if the issue exist there as well. Another option would be to support DER format, and either encode it properly; or convert to PEM format. 

Closes #8142

### More

- [x] Added/updated tests

### DIFF
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -836,7 +836,7 @@ func getCertificateBlocks(secret *corev1.Secret, namespace, secretName string) (
 			namespace, secretName, strings.Join(missingEntries, ", "))
 	}
 
-	if !isPem(tlsCrtData){
+	if !isPem(tlsCrtData) {
 		missingEntries = append(missingEntries, "tls.crt")
 	}
 
@@ -1285,6 +1285,15 @@ func templateSafeString(value string) error {
 }
 
 func isPem(data []byte) bool {
-	block, _ := pem.Decode(data)
-	return block != nil
+	for {
+		block, rest := pem.Decode(data)
+		if block == nil {
+			return false
+		}
+		if len(rest) == 0 {
+			break
+		}
+		data = rest
+	}
+	return true
 }
--- a/provider/kubernetes/kubernetes.go
+++ b/provider/kubernetes/kubernetes.go
@@ -3,6 +3,7 @@ package kubernetes
 import (
 	"bufio"
 	"bytes"
+	"encoding/pem"
 	"errors"
 	"flag"
 	"fmt"
@@ -835,6 +836,19 @@ func getCertificateBlocks(secret *corev1.Secret, namespace, secretName string) (
 			namespace, secretName, strings.Join(missingEntries, ", "))
 	}
 
+	if !isPem(tlsCrtData){
+		missingEntries = append(missingEntries, "tls.crt")
+	}
+
+	if !isPem(tlsKeyData) {
+		missingEntries = append(missingEntries, "tls.key")
+	}
+
+	if len(missingEntries) > 0 {
+		return "", "", fmt.Errorf("secret %s/%s does not contain PEM formatted TLS data entries: %s",
+			namespace, secretName, strings.Join(missingEntries, ","))
+	}
+
 	return cert, key, nil
 }
 
@@ -1268,4 +1282,9 @@ func getPassTLSClientCert(i *extensionsv1beta1.Ingress) *types.TLSClientHeaders 
 func templateSafeString(value string) error {
 	_, err := strconv.Unquote(`"` + value + `"`)
 	return err
+}
+
+func isPem(data []byte) bool {
+	block, _ := pem.Decode(data)
+	return block != nil
 }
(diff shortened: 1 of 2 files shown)


provider/kubernetes/kubernetes.go
```diff
@@ -1269,3 +1283,8 @@ func templateSafeString(value string) error {
 	_, err := strconv.Unquote(`"` + value + `"`)
 	return err
 }
+
+func isPem(data []byte) bool {
+	block, _ := pem.Decode(data)
```
dtomcej, line 1288: since according to RFC 7468  PEM blocks must begin with `-----BEGIN`, could you not first do a comparison to the first bytes in the slice to see if they contain a valid header, and fail fast without having to decode the whole dataset first?

dtomcej: This would also allow a clean message to be logged : `Missing or malformed PEM header` sort of thing...

jr0d: @dtomcej Thanks for taking a look at this.

I think this should be sufficient in most cases, however, if the PEM is malformed (or malicious) it could still break configuration reconciliation. If you do not think this is worth the performance hit, I could modify the check to:

1. check the header/footer
2. validate the data is valid base64

I actually just realized that I should be looping over the  rest of the data, in the case of a multi document PEM file. So, if I do go that approach, I'll need to re-implement a good chunk of pem.Decode() anyway. 

wdyt?

joejulian: I think both would be reasonable. Fail fast on a malformed PEM header, then pem.Decode if the header is valid. A regex header validation would also work to break apart a multi-document PEM file for a looping decode.

dtomcej:
So a `Block` in `encoding/pem` is:

```
type Block struct {
    Type    string            // The type, taken from the preamble (i.e. "RSA PRIVATE KEY").
    Headers map[string]string // Optional headers.
    Bytes   []byte            // The decoded bytes of the contents. Typically a DER encoded ASN.1 structure.
}
```
from (https://golang.org/pkg/encoding/pem/#Block).

If we have DER encoding, we can pretty much just jam them into the Bytes....however, we may not be able to infer the type directly, since we may not know the encoding method. OpenSSL uses many different preambles that are not all covered in the RFCs: (https://github.com/openssl/openssl/blob/master/include/openssl/pem.h#L32).

Since we use different certificate and key files, I wonder if we manually took the key DER, jammed it in a block with a "PRIVATE KEY" preamble, and then did a `pem.Decode(pem.Encode(block))` to allow golang to verify/propagate the contents of the struct. We could then do the same for the certificate.

It wouldn't solve all cases, but it would allow DER support for non-encrypted certificates and keys, which would be the goal here.

Thoughts?

dtomcej:
Hello @jr0d,

The Traefik team discussed this PR, and here is where we are at:

- Since a "malformed" certificate can break configuration parsing, we view that as a bug, and will accept a fix into the 1.7 branch. We have marked the base issue as a bug, because only accept bugfixes and security updates to 1.7 as it is an older release. Therefore, in this PR, we should focus on mitigating the impact that the certificates have, such as adding checks, and failing/continuing to parse.
- We think that DER support would be fine to add in v2, and we would invite you to create a new issue to track that enhancement, as it would be classified separately from this issue.

Thanks!

jr0d:
@dtomcej Thank you for the update! I'll work on merging this PR and creating a new issue/feature/PR for traefik-2. I'll circle back to this at the beginning of next week most likely. Thanks again.

rtribotte (approved):
Thanks 👍

tomMoulard (approved):
LGTM :ok_hand:

mpl (approved):

rtribotte:
Closed in benefit of https://github.com/traefik/traefik/pull/8624.

PR #8156, opened 2021-05-26
Fix incorrect behaviour with multi-port endpoint subsets

coufalja:
### What does this PR do?

Fixes an incorrect behaviour when the Endpoints generated for Service pods contain two distinct subsets for two different ports.

### Motivation

The bug had broken down one of our production services for a while and we had to revert to Traefik 1.7.
The bug caused some (third in our case) of HTTP requests to be forwarded to pods that hadn't had the port exposed.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -529,8 +529,8 @@ func loadService(client Client, namespace string, backend networkingv1beta1.Ingr
 		return nil, errors.New("subset not found")
 	}
 
-	var port int32
 	for _, subset := range endpoints.Subsets {
+		var port int32
 		for _, p := range subset.Ports {
 			if portName == p.Name {
 				port = p.Port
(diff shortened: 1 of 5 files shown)


pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -529,8 +529,8 @@ func loadService(client Client, namespace string, backend networkingv1beta1.Ingr
 		return nil, errors.New("subset not found")
 	}
 
-	var port int32
 	for _, subset := range endpoints.Subsets {
+		var port int32
```
coufalja, line 533: I guess that this was a mere oversight, by declaring the variable before the cycle its value remained unchanged in subsequent cycles thus (incorrectly) falling through the `if port == 0` guard.

jbdoumenjou (approved):
LGTM :+1:

rtribotte (approved):
Thanks 👍

ldez (approved):
LGTM

PR #8169, opened 2021-06-02
Add routing IP rule matcher

### LINKED ISSUE #7138: Origin IP as Router rule matcher

tomMoulard:
### What does this PR do?

This PR adds IP routing based on the client IP.

You can now write your rules: ```ClientIP(`10.0.0.0`, `192.168.0.0/24`, `80fe::1`, `::1/64`)```

It supports both IPv4 and IPv6 with and without CIDR.

### Motivation

Fixes #7138.

### More

- [X] Added/updated tests

### Additional Notes

`ClientIP` matcher only tries to match the request client IP and does not rely on the `X-Forwarded-For` header.

Could be nice to add a rule example in an appropriate section of the documentation.

### DIFF
--- a/docs/content/routing/routers/index.md
+++ b/docs/content/routing/routers/index.md
@@ -239,6 +239,7 @@ The table below lists all the available matchers:
 | ```Path(`/path`, `/articles/{cat:[a-z]+}/{id:[0-9]+}`, ...)```         | Match exact request path. It accepts a sequence of literal and regular expression paths.                       |
 | ```PathPrefix(`/products/`, `/articles/{cat:[a-z]+}/{id:[0-9]+}`)```   | Match request prefix path. It accepts a sequence of literal and regular expression prefix paths.               |
 | ```Query(`foo=bar`, `bar=baz`)```                                      | Match Query String parameters. It accepts a sequence of key=value pairs.                                       |
+| ```ClientIP(`10.0.0.0/16`, `::1`)```                                   | Check if the request client IP matches one of the given IP/CIDR. It accepts IPv4, IPv6 and CIDR format.        |
 
 !!! important "Non-ASCII Domain Names"
 
@@ -271,6 +272,10 @@ The table below lists all the available matchers:
     Use a `*Prefix*` matcher if your service listens on a particular base path but also serves requests on sub-paths.
     For instance, `PathPrefix: /products` would match `/products` but also `/products/shoes` and `/products/shirts`.
     Since the path is forwarded as-is, your service is expected to listen on `/products`.
+
+!!! info "ClientIP matcher"
+
+    `ClientIP` matcher only try to match the request client IP and does not rely on the `X-Forwarded-For` header.
 
 ### Priority
 
--- a/docs/content/routing/routers/index.md
+++ b/docs/content/routing/routers/index.md
@@ -239,7 +239,7 @@ The table below lists all the available matchers:
 | ```Path(`/path`, `/articles/{cat:[a-z]+}/{id:[0-9]+}`, ...)```         | Match exact request path. It accepts a sequence of literal and regular expression paths.                       |
 | ```PathPrefix(`/products/`, `/articles/{cat:[a-z]+}/{id:[0-9]+}`)```   | Match request prefix path. It accepts a sequence of literal and regular expression prefix paths.               |
 | ```Query(`foo=bar`, `bar=baz`)```                                      | Match Query String parameters. It accepts a sequence of key=value pairs.                                       |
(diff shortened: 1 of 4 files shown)


dtomcej:
This only implements the matcher for HTTP routers correct? Did we want to add this matcher for TCP/UDP routers too? Or did we want to add that in a future PR?

tomMoulard:
Indeed, this pr only adds the matcher for HTTP routers.

As the matching could use more than just the `ClientIP` for TCP/UDP, it should be addressed in another pr.

docs/content/routing/routers/index.md
```diff
@@ -239,6 +239,7 @@ The table below lists all the available matchers:
 | ```Path(`/path`, `/articles/{cat:[a-z]+}/{id:[0-9]+}`, ...)```         | Match exact request path. It accepts a sequence of literal and regular expression paths.                       |
 | ```PathPrefix(`/products/`, `/articles/{cat:[a-z]+}/{id:[0-9]+}`)```   | Match request prefix path. It accepts a sequence of literal and regular expression prefix paths.               |
 | ```Query(`foo=bar`, `bar=baz`)```                                      | Match Query String parameters. It accepts a sequence of key=value pairs.                                       |
+| ```ClientIP(`10.0.0.0/16`, `::1`)```                                   | Check if the request client IP matches one of the given IP/CIDR. It accepts IPv4, IPv6 and CIDR format.        |
```
dtomcej, line 242:
```suggestion
| ```ClientIP(`10.0.0.0/16`, `::1`)```                                   | Match if the request client IP is one of the given IP/CIDR. It accepts IPv4, IPv6 and CIDR formats.        |
```

docs/content/routing/routers/index.md
```diff
@@ -272,6 +273,10 @@ The table below lists all the available matchers:
     For instance, `PathPrefix: /products` would match `/products` but also `/products/shoes` and `/products/shirts`.
     Since the path is forwarded as-is, your service is expected to listen on `/products`.
 
+!!! info "ClientIP matcher"
+
+    `ClientIP` matcher only try to match the request client IP and does not rely on the `X-Forwarded-For` header.
```
dtomcej, line 278:
```suggestion
    The `ClientIP` matcher will only match the request client IP and does not use the `X-Forwarded-For` header for matching.
```

pkg/rules/rules.go
```diff
@@ -155,6 +157,47 @@ func host(route *mux.Route, hosts ...string) error {
...
+	var authorizedIPsNet []*net.IPNet
+
+	for _, ip := range clientIPs {
+		if ipAddr := net.ParseIP(ip); ipAddr == nil {
+			_, ipNet, err := net.ParseCIDR(ip)
+			if err != nil {
+				return fmt.Errorf("invalid value %q for \"ClientIP\" matcher: %w", ip, err)
+			}
+			authorizedIPsNet = append(authorizedIPsNet, ipNet)
+		} else {
```
dtomcej, line 171: Could this be done with a `continue` instead of an else?

juliens (approved):
LGTM

rtribotte (approved):
LGTM 👍

dtomcej (approved):
LGTM
:shipit:

fkollmann:
Would it be possible to get this for TCP routers, too?

PR #8174, opened 2021-06-04
Support for Consul namespaces in Consul Catalog

Lucretius:
### What does this PR do?

This pull request adds support for Consul namespaces in the Consul Catalog provider (issue https://github.com/traefik/traefik/issues/8171)

### Motivation

I have a Consul cluster which makes use of Namespaces.  I would like to have Consul Catalog be able to search across any specified Consul namespaces to find services.

### More

- [ ] Added/updated tests

### Additional Notes

I ran `go generate`, which updated some `.md` files that do not appear to relate to any of my changes.  I also upgraded the Consul client, as the old one did not have namespace support

Consul Namespaces are a feature of Consul Enterprise, so I don't think it's possible to write tests in the existing suite which cover this new feature - only that the new feature does not break existing tests.  The way I have been testing this is by building a custom docker image and using it against my production enterprise cluster.

I have confirmed that:
1) If `namespaces` is missing, only the default namespace is searched (for back-compatibility)
2) If `namespaces` is set to `['*']`, all namespaces are searched (if namespacing is fairly dynamic this will make for easier configuration)
3) If `namespaces` is set to a specific set `['default', 'namespace1']`, those specified namespaces are searched.

One last thing - in the UI I was hoping to be able to differentiate two services which may have the same name, but exist in different namespaces.  I tried to make the service name something like `service@namespace` or `namespace/service` (if namespace is `default` then just use `service` as the name), but I think all the special characters are forbidden in the service name.  Is there any suggestion for a naming convention?

### DIFF
--- a/docs/content/reference/dynamic-configuration/kv-ref.md
+++ b/docs/content/reference/dynamic-configuration/kv-ref.md
@@ -168,6 +168,7 @@
 | `traefik/http/serversTransports/ServersTransport0/certificates/0/keyFile` | `foobar` |
 | `traefik/http/serversTransports/ServersTransport0/certificates/1/certFile` | `foobar` |
 | `traefik/http/serversTransports/ServersTransport0/certificates/1/keyFile` | `foobar` |
+| `traefik/http/serversTransports/ServersTransport0/disableHTTP2` | `true` |
 | `traefik/http/serversTransports/ServersTransport0/forwardingTimeouts/dialTimeout` | `42s` |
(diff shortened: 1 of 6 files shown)


jbdoumenjou:
Hello @Lucretius,
 
Thanks for your contribution!

kevinpollet:
Hello @Lucretius,

As explained in #8592, it is not currently possible to support multiple Consul Enterprise Namespaces in the Consul Catalog provider (more details in the pull request comment).

As we opened a new pull request on this subject, would you mind being added as a co-author on #8592?

According to this, we will decline and close this pull request.

Issue #6997, opened 2020-07-03
Add host tag on InfluxDB metrics


acouvreur:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

The actual metrics collection for InfluxDB has 5 tags : code, method, protocol and service/entrypoint.

When working with multiple Traefik instance that store the metrics on the same InfluxDB instance it would be nice to have a way to distinguish who emit which data. 

I suggest adding a tag named `host`. That defaults to the machine hostname. Like many collectors (e.g. telegraf) they include the hostname as a tag (and can be overridden in the config file).

philippreston:
Hey 

We run multiple Traefiks around our environment for different purposes.  Without host tags the native metrics here are un-usable.

Thanks

Phil

kevinpollet:
Closed by #8308

PR #8178, opened 2021-06-08
Add configurable tags to influxdb metrics

Tetha:
### What does this PR do?

This allows users to set additional influxdb tags on their traefik instances so they can distinguish different traefik instances in the influxdb. For users with existing influxdb changes, this PR changes nothing during an update and the old configuration functions identically, except that they can improve it afterwards.

### Motivation

We have a certain fleet of traefik instances across different production environments and with different specializations. When I enabled the influxdb metric collection, it was an easy experience to do so, but it resulted in a bit of a mess on the influxdb side. With this pull request, I can add different tags to my traefik instances, so I can distinguish them.

There also is an issue about this, #6997. However, I don't think introducing default tags is a good strategy. Quite a few tools try to do so and just end up making more work because the default tags might not apply to all situations. For example, a default host tag would just be tag-spam for our container-based instances. Just pushing a map into the influxdb client allows all users to adapt the  tags to their specific setup. 

### More

- [x] Added/updated tests - I have modified one unit test in `influxdb_metrics`.  I've also done some manual testing in case something weird explodes. :)
- [x] Added/updated documentation - I have modified the Observability chapter.

### Additional Notes

As mentioned in the commit message, overall, this was pretty simple. I practically just moved the empty list from a hardcoded call in the influx-client constructor to the configuration type and that was all I had to do.

However, there was one snag when changing the tests. There are two tests in place, one for udp and one for http. I figured - lets run one test with additional tags, and one without. That way, I'd know that both setting additional tags works, and not setting additional tags works. 

However, I discovered that only the first call to `RegisterInfluxDB` across both tests had an effect. This is because on master, `RegisterInfluxDB` sets the global `influxDBClient` if and only if it is nil. 

I figured it was not entirely clean that these tests are coupled in this way and figured there are two ways to decouple this - either `RegisterInfluxDB` has to always set the `influxDBClient`, or `StopInfluxDB` has to clear the `influxDBClient` after stopping the ticker. After some deliberation, I figured that clearing the `influxDBClient` in the `StopInfluxDB` method has a higher chance of complications - what if the ticker runs right now, or what if the ticker takes some time to stop, what if the client has some metrics buffered, ...?

Thus, I changed the `RegisterInfluxDB` method to always overwrite the `influxDBClient` when called. This should be fine for now, because as far as I can see the code, the `RegisterInfluxDB` method is called once per traefik instance start, and in some test cases. So at the moment, as far as I can see, the if-statement I removed was only exercised in tests.  This might become a problem with some kind of config reload for the metrics, but that wouldn't have worked either way.

### DIFF
--- a/docs/content/observability/metrics/influxdb.md
+++ b/docs/content/observability/metrics/influxdb.md
@@ -240,7 +240,7 @@ metrics:
 
 _Optional, Default={}_
 
-Additional labels (influxdb tags) to send to influxdb for all metrics.
+Additional labels (influxdb tags) on all metrics.
 
 ```toml tab="File (TOML)"
 [metrics]
--- a/docs/content/observability/metrics/influxdb.md
+++ b/docs/content/observability/metrics/influxdb.md
@@ -236,16 +236,16 @@ metrics:
 --metrics.influxdb.pushInterval=10s
 ```
 
-#### `additionalTags`
+#### `additionalLabels`
 
 _Optional, Default={}_
 
-Additional tags to send to influxdb for all metrics.
+Additional labels (influxdb tags) to send to influxdb for all metrics.
 
 ```toml tab="File (TOML)"
 [metrics]
   [metrics.influxDB]
-    [metrics.influxDB.additionalTags]
+    [metrics.influxDB.additionalLabels]
       host = "traefik01.example.com"
       environment = "production"
 ```
@@ -253,11 +253,11 @@ Additional tags to send to influxdb for all metrics.
 ```yaml tab="File (YAML)"
 metrics:
   influxDB:
-    additionalTags:
+    additionalLabels:
       host: traefik01.example.com
       environment: production
 ```
 
 ```bash tab="CLI"
---metrics.influxdb.additionaltags.host=traefik01.example.com --metrics.influxdb.additionaltags.environment=production
+--metrics.influxdb.additionallabels.host=traefik01.example.com --metrics.influxdb.additionallabels.environment=production
 ```
--- a/docs/content/observability/metrics/influxdb.md
+++ b/docs/content/observability/metrics/influxdb.md
@@ -235,3 +235,29 @@ metrics:
 ```bash tab="CLI"
 --metrics.influxdb.pushInterval=10s
 ```
+
+#### `additionalTags`
+
+_Optional, Default={}_
+
+Additional tags to send to influxdb for all metrics.
+
+```toml tab="File (TOML)"
+[metrics]
+  [metrics.influxDB]
+    [metrics.influxDB.additionalTags]
+      host = "traefik01.example.com"
+      environment = "production"
+```
+
+```yaml tab="File (YAML)"
+metrics:
+  influxDB:
+    additionalTags:
+      host: traefik01.example.com
+      environment: production
+```
+
+```bash tab="CLI"
+--metrics.influxdb.additionaltags.host=traefik01.example.com --metrics.influxdb.additionaltags.environment=production
+```
--- a/docs/content/reference/static-configuration/cli-ref.md
+++ b/docs/content/reference/static-configuration/cli-ref.md
@@ -246,6 +246,9 @@ InfluxDB metrics exporter type. (Default: ```false```)
 `--metrics.influxdb.addentrypointslabels`:  
 Enable metrics on entry points. (Default: ```true```)
 
+`--metrics.influxdb.additionallabels.<name>`:  
+Additional labels (influxdb tags) on all metrics
+
 `--metrics.influxdb.address`:  
 InfluxDB address. (Default: ```localhost:8089```)
 
--- a/docs/content/reference/static-configuration/env-ref.md
+++ b/docs/content/reference/static-configuration/env-ref.md
@@ -246,6 +246,9 @@ InfluxDB metrics exporter type. (Default: ```false```)
 `TRAEFIK_METRICS_INFLUXDB_ADDENTRYPOINTSLABELS`:  
 Enable metrics on entry points. (Default: ```true```)
 
+`TRAEFIK_METRICS_INFLUXDB_ADDITIONALLABELS_<NAME>`:  
+Additional labels (influxdb tags) on all metrics
+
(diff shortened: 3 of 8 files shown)


Tetha:
Hm, the tests seem to be in a bad mood today. Traefik had one green build and two timeouts, and semaphore had a valid failure about the formatting, but now the integration tests fail in some unrelated middleware (but the validation succeeds).

rtribotte:
Hello @Tetha 

Thanks for your contribution!

rtribotte:
Hello @Tetha,

I took a look at your PR and overall it looks great.

I have a first question regarding the naming of the option. In Traefik documentation we always talk about `labels` for metrics.
As well, the existing option to activate metrics on `EntryPoint`, `Routers` and `Services` are named with the `labels` suffix.
So I believe it would be better to keep that concept and replace the `AdditionalTags` option with something like `BaseLabels`, `RootLabels` or `AdditionalLabels`.
This way, if we want to introduce the same option for other metrics providers, it would allow reusing the same naming.
 
I also tested your change on the initialization of the metrics client:
https://github.com/traefik/traefik/blob/0482b52df73bdc72a462bb989a9619743e08281c/pkg/metrics/influxdb.go#L59-L65
When I revert this change, the unit tests do not fail on my machine, so I'm unsure what issue you did encounter.
Could you check that again? For now, I would revert those changes.

The documentation references should also be updated.

Of course, I can address all those changes for you if you want.

Tetha:
Hi,

>  This way, if we want to introduce the same option for other metrics providers, it would allow reusing the same naming.

I've modified the option to be called "additionalLabels", and mentioned in the documentation that these are mapped to influxdb tags. That should be a good middle ground between traefiks naming and clarity how it works in influx.

>  When I revert this change, the unit tests do not fail on my machine, so I'm unsure what issue you did encounter.
Could you check that again? For now, I would revert those changes.

Hmm, this is strange. I just reverted my changes to the `RegisterInfluxdb` - conditionally initialize the InfluxDBClient like this:

```go
	if influxDBClient == nil {
		influxDBClient = initInfluxDBClient(ctx, config)
	}
```

When I run `make test`, which runs both influxdb unit tests, the latter one fails like this, because the influxdb metrics includes the unexpected tags. The `tag1=val` is not expected in the second test, because that second test does not configure `AdditionalLabels`. 

```
--- FAIL: TestInfluxDBHTTP (5.00s)
    influxdb_test.go:157: Got "traefik.config.reload.total.failure,tag1=val1 count=1 1624779091187176947\ntraefik.config.reload.total,tag1=val1 count=1 1624779091187176947\ntraefik.config.reload.lastSuccessTimestamp,tag1=val1 value=1 1624779091187176947\ntraefik.config.reload.lastFailureTimestamp,tag1=val1 value=1 1624779091187176947\n" [], want "(traefik\\.config\\.reload\\.total count=1) [\\d]{19}"
    influxdb_test.go:157: Got "traefik.config.reload.total.failure,tag1=val1 count=1 1624779091187176947\ntraefik.config.reload.total,tag1=val1 count=1 1624779091187176947\ntraefik.config.reload.lastSuccessTimestamp,tag1=val1 value=1 1624779091187176947\ntraefik.config.reload.lastFailureTimestamp,tag1=val1 value=1 1624779091187176947\n" [], want "(traefik\\.config\\.reload\\.total\\.failure count=1) [\\d]{19}"
    influxdb_test.go:157: Got "traefik.config.reload.total.failure,tag1=val1 count=1 1624779091187176947\ntraefik.config.reload.total,tag1=val1 count=1 1624779091187176947\ntraefik.config.reload.lastSuccessTimestamp,tag1=val1 value=1 1624779091187176947\ntraefik.config.reload.lastFailureTimestamp,tag1=val1 value=1 1624779091187176947\n" [], want "(traefik\\.config\\.reload\\.lastSuccessTimestamp value=1) [\\d]{19}"
    influxdb_test.go:157: Got "traefik.config.reload.total.failure,tag1=val1 count=1 1624779091187176947\ntraefik.config.reload.total,tag1=val1 count=1 1624779091187176947\ntraefik.config.reload.lastSuccessTimestamp,tag1=val1 value=1 1624779091187176947\ntraefik.config.reload.lastFailureTimestamp,tag1=val1 value=1 1624779091187176947\n" [], want "(traefik\\.config\\.reload\\.lastFailureTimestamp value=1) [\\d]{19}"
...
```

rtribotte:
Hello @Tetha,

I see that you have merged the review commits, but they were supposed to be on top of your commits.
Also, the merge commits are not cleaned, maybe the rebase didn't go well.

Are you willing to let the maintainers of Traefik writing on your branch? It could be simpler, as i could rewrite the history for you.

WDYT?

Tetha:
Yes, sorry. I got confused there with the PR on the PR.

I'm not that well versed in github permissions - who should I grant write access on that fork to?

rtribotte:
Hey sorry for that, I would have initially pushed my commits to your branch, because it's simpler.
But because you probably don't check the box to allow maintainers to edit your branch, I've opened a PR directly on your fork, since a lot of changes are difficult to work out with GitHub UI only.

To allow us (Traefik maintainers) to push to your branch, you would simply check that box ([documentation](https://docs.github.com/en/github/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)).
Is that what you were looking for?

Tetha:
Moin @rtribotte - you should have access to the repository now.

rtribotte:
Hello @Tetha,

Sorry for the late answer, I have rebased your PR locally and I have tried to push a clean commit history, but unfortunately, it seems that I still do not have the right to push.

Tetha:
Moin, 

in the repo, you are still marked as "invite pending". Did you accept that invite?

rtribotte:
Hey @Tetha,

Sorry for that, I indeed had an invitation that I didn't accept in time. It is now expired.

Nonetheless, having the code on a protected organization branch can also be an issue later on for the bot that is responsible for the merge of the PR.
The easiest way would be to submit the PR from a branch of your own fork.

WDYT?

Tetha:
Closing in favor of #8308

PR #8184, opened 2021-06-14
Fixed BIND_DIR quoting

apollo13:
Really small PR that fixes the bind mount to actually have `dist` as bind_dir and not `"dist"`


jbdoumenjou:
Hello @apollo13,

Could you explain what it solves?

apollo13:
It probably fixes nothing. It just seems that make does not need quoting and would otherwise create folders named "dist" (sic) when using that variable on the makefile.

apollo13:
http://web.mit.edu/gnu/doc/html/make_6.html also (at least on a quick glance) does not suggest that quotes would be needed there

jbdoumenjou (approved):
LGTM

PR #8185, opened 2021-06-14
Don't remove ingress config on API call failure

### LINKED ISSUE #8066: When Kubernetes apiserver down , Traefik 2 starts returning 404

Bug

### What did you do?

To verify traefik ingress controller works correctly  when all kubernetes apiserver is down. 
Traefik work proper a period of time, then return 404.
I see it's fixed in  treafik v1 https://github.com/traefik/traefik/pull/1295

No matter what circumstances return to 200.

### What did you see instead?
Return 200  a period of time and return to 404.

```
Version:      2.4.8
Codename:     livarot
Go version:   go1.16.2
Built:        2021-03-23T15:48:39Z
OS/Arch:      linux/amd64
```

```
spec:
  containers:
  - args:
    - --global.checknewversion
    - --global.sendanonymoususage
    - --entryPoints.traefik.address=:9000/tcp
    - --entryPoints.web.address=:8000/tcp
    - --entryPoints.websecure.address=:8443/tcp
    - --api.dashboard=true
    - --ping=true
    - --providers.kubernetescrd
    - --api.insecure=true
```

```
E0414 04:40:47.679821       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1alpha1.ServersTransport: failed to list *v1alpha1.ServersTransport: Get "https://10.248.252.1:443/apis/traefik.containo.us/v1alpha1/serverstransports?limit=500&resourceVersion=0": dial tcp 10.248.252.1:443: connect: connection refused
E0414 04:41:18.969402       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1alpha1.ServersTransport: failed to list *v1alpha1.ServersTransport: Get "https://10.248.252.1:443/apis/traefik.containo.us/v1alpha1/serverstransports?limit=500&resourceVersion=0": dial tcp 10.248.252.1:443: connect: connection refused
E0414 04:41:58.696359       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1alpha1.ServersTransport: failed to list *v1alpha1.ServersTransport: Get "https://10.248.252.1:443/apis/traefik.containo.us/v1alpha1/serverstransports?limit=500&resourceVersion=0": dial tcp 10.248.252.1:443: connect: connection refused
E0414 04:42:43.657032       1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1alpha1.ServersTransport: failed to list *v1alpha1.ServersTransport: Get "https://10.248.252.1:443/apis/traefik.containo.us/v1alpha1/serverstransports?limit=500&resourceVersion=0": dial tcp 10.248.252.1:443: connect: connection refused
```

dtomcej:
### What does this PR do?

This PR modifies the kubernetes ingress provider to allow API failures without removing all routers and content.

The `GetServerVersion()` method does not use a lister, and will fail when connectivity is lost.

### Motivation

Fixes #8066 

### More

- ~~[ ] Added/updated tests~~ Bugfix, manual testing
- ~~[ ] Added/updated documentation~~ Bugfix

### Additional Notes

This PR was tested by:

- Running a Kubernetes cluster (1.19.7) and deploying Traefik with the kubernetes ingress provider accessing the kubernetes cluster via `kubectl proxy`
- Deploying a whoami pod and exposing it using Ingress
- Verifying that the http router is created
- Kill the `kubectl proxy` command to make the Kubernetes API unreachable
- Wait until the resync period elapses (~10')
- Confirm that the router is still present in the dashboard/api

This PR currently:
- Adds an error to the `loadConfigurationFromIngresses` method, and if a `serverVersion` error is returned, break out of the config hash case to prevent dropping the "old but good" configuration. This should prevent mis-categorizing ingress objects if the class cannot be determined.

When Testing:
- If you change the `rsyncPeriod` in `client.go` from 10m to 10s, you can reduce the rsync period accordingly. This will greatly speed up testing.

Co-authored-by: rtribotte@users.noreply.github.com

### DIFF
--- a/pkg/provider/kubernetes/ingress/client.go
+++ b/pkg/provider/kubernetes/ingress/client.go
@@ -60,7 +60,7 @@ type Client interface {
 	GetSecret(namespace, name string) (*corev1.Secret, bool, error)
 	GetEndpoints(namespace, name string) (*corev1.Endpoints, bool, error)
 	UpdateIngressStatus(ing *networkingv1beta1.Ingress, ingStatus []corev1.LoadBalancerIngress) error
-	GetServerVersion() (*version.Version, error)
+	GetServerVersion() *version.Version
 }
 
 type clientWrapper struct {
@@ -209,15 +209,18 @@ func (c *clientWrapper) WatchAll(namespaces []string, stopCh <-chan struct{}) (<
 		}
 	}
 
-	// Reset the stored server version to recheck on reconnection.
-	c.serverVersion = nil
 	// Get and store the serverVersion for future use.
-	serverVersion, err := c.GetServerVersion()
+	serverVersionInfo, err := c.clientset.Discovery().ServerVersion()
 	if err != nil {
-		log.WithoutContext().Errorf("Failed to get server version: %v", err)
-		return eventCh, nil
+		return eventCh, fmt.Errorf("could not retrieve server version: %w", err)
 	}
 
+	serverVersion, err := version.NewVersion(serverVersionInfo.GitVersion)
+	if err != nil {
+		return eventCh, fmt.Errorf("could not parse server version: %w", err)
+	}
+
+	c.serverVersion = serverVersion
 	if supportsIngressClass(serverVersion) {
 		c.clusterFactory = informers.NewSharedInformerFactoryWithOptions(c.clientset, resyncPeriod)
 		c.clusterFactory.Networking().V1beta1().IngressClasses().Informer().AddEventHandler(eventHandler)
@@ -430,23 +433,8 @@ func (c *clientWrapper) lookupNamespace(ns string) string {
 }
 
 // GetServerVersion returns the cluster server version, or an error.
-func (c *clientWrapper) GetServerVersion() (*version.Version, error) {
-	if c.serverVersion != nil {
-		return c.serverVersion, nil
-	}
-
-	serverVersionInfo, err := c.clientset.Discovery().ServerVersion()
-	if err != nil {
-		return nil, fmt.Errorf("could not retrieve server version: %w", err)
-	}
-
-	serverVersion, err := version.NewVersion(serverVersionInfo.GitVersion)
-	if err != nil {
-		return nil, fmt.Errorf("could not parse server version: %w", err)
-	}
-
-	c.serverVersion = serverVersion
-	return c.serverVersion, nil
+func (c *clientWrapper) GetServerVersion() *version.Version {
+	return c.serverVersion
 }
 
 // eventHandlerFunc will pass the obj on to the events channel or drop it.
--- a/pkg/provider/kubernetes/ingress/client.go
+++ b/pkg/provider/kubernetes/ingress/client.go
@@ -72,6 +72,7 @@ type clientWrapper struct {
 	ingressLabelSelector string
 	isNamespaceAll       bool
 	watchedNamespaces    []string
+	serverVersion        *version.Version
 }
 
 // newInClusterClient returns a new Provider client that is expected to run
@@ -208,6 +209,9 @@ func (c *clientWrapper) WatchAll(namespaces []string, stopCh <-chan struct{}) (<
 		}
 	}
 
+	// Reset the stored server version to recheck on reconnection.
+	c.serverVersion = nil
+	// Get and store the serverVersion for future use.
 	serverVersion, err := c.GetServerVersion()
 	if err != nil {
 		log.WithoutContext().Errorf("Failed to get server version: %v", err)
@@ -427,12 +431,22 @@ func (c *clientWrapper) lookupNamespace(ns string) string {
 
 // GetServerVersion returns the cluster server version, or an error.
 func (c *clientWrapper) GetServerVersion() (*version.Version, error) {
-	serverVersion, err := c.clientset.Discovery().ServerVersion()
+	if c.serverVersion != nil {
+		return c.serverVersion, nil
+	}
+
+	serverVersionInfo, err := c.clientset.Discovery().ServerVersion()
 	if err != nil {
 		return nil, fmt.Errorf("could not retrieve server version: %w", err)
 	}
 
-	return version.NewVersion(serverVersion.GitVersion)
+	serverVersion, err := version.NewVersion(serverVersionInfo.GitVersion)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse server version: %w", err)
+	}
+
+	c.serverVersion = serverVersion
+	return c.serverVersion, nil
 }
 
 // eventHandlerFunc will pass the obj on to the events channel or drop it.
--- a/pkg/provider/kubernetes/ingress/client.go
+++ b/pkg/provider/kubernetes/ingress/client.go
@@ -471,6 +471,10 @@ func (c *clientWrapper) isWatchedNamespace(ns string) bool {
 // IngressClass objects are supported since Kubernetes v1.18.
 // See https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class
 func supportsIngressClass(serverVersion *version.Version) bool {
+	if serverVersion == nil {
+		return false
+	}
+
 	ingressClassVersion := version.Must(version.NewVersion("1.18"))
 
 	return ingressClassVersion.LessThanOrEqual(serverVersion)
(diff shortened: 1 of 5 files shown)


pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -187,12 +187,11 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 	serverVersion, err := client.GetServerVersion()
 	if err != nil {
 		log.FromContext(ctx).Errorf("Failed to get server version: %v", err)
-		return conf
 	}
 
 	var ingressClasses []*networkingv1beta1.IngressClass
 
-	if supportsIngressClass(serverVersion) {
+	if serverVersion != nil && supportsIngressClass(serverVersion) {
```
ldez, line 194: I think you can move this into the `supportsIngressClass()` method

dtomcej: Good call.

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -187,7 +187,6 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 	serverVersion, err := client.GetServerVersion()
 	if err != nil {
 		log.FromContext(ctx).Errorf("Failed to get server version: %v", err)
-		return conf
```
ravilr, line 190: won't this ignoring of error here could result in improper processing of IngressClass resource semantics. Assume an installation with IngressClass resource defined and an Ingress resource referring to that.  Under scenarios where kube-api is unavailable, this would cause the `shouldProcessIngress()` method to skip such an ingress in the provider sync loop.

ravilr:
Also noticed that this PR won't fix the issue in master branch, as there are more places where `GetServerVersion()` is being used in master branch.

pkg/provider/kubernetes/ingress/kubernetes.go
dtomcej: @ravilr that is a great scenario that I hadn't considered.

dtomcej: I have taken your suggestion, and reworked this to kick the error back to the provider loop.

ravilr:
LGTM for v2.4 branch. master branch requires further changes to fix this issue.

dtomcej:
@ravilr, The master branch does not contain any more instances than 2.4.

This fix should be able to merge forward from 2.4 -> master without issue.

ravilr:
> AFAICT the master branch does not contain any more instances than 2.4.

there are more places in master branch where GetServerVersion() is called :

1. GetIngresses() returns empty list of Ingresses on GetServerVersion() error here:
 https://github.com/traefik/traefik/blob/0a3e40332a86cc061bf16d8895b10defeb999171/pkg/provider/kubernetes/ingress/client.go#L233

2. GetIngressClasses() returns err from GetServerVersion() error here:
 https://github.com/traefik/traefik/blob/0a3e40332a86cc061bf16d8895b10defeb999171/pkg/provider/kubernetes/ingress/client.go#L475 , but the error is ignored.

dtomcej:
Hello @ravilr,

The `GetIngresses()` method is used here: https://github.com/traefik/traefik/blob/e1e1fd640c3104c953f240128a6cfca1c4ce7dd2/pkg/provider/kubernetes/ingress/kubernetes.go#L208

and the `GetIngressClasses()` method is used here:

https://github.com/traefik/traefik/blob/0a3e40332a86cc061bf16d8895b10defeb999171/pkg/provider/kubernetes/ingress/kubernetes.go#L196

both of which should have already returned due to the failed call on https://github.com/traefik/traefik/blob/e1e1fd640c3104c953f240128a6cfca1c4ce7dd2/pkg/provider/kubernetes/ingress/kubernetes.go#L187.

Unless you are thinking that the API could fail between those lines, resulting in similar behavior. The only way I can see that would prevent this is to move the `getServerVersion()` method to the client init, where it could be stored as a client variable, and would not require API access for this.

Thoughts?

ravilr:
> Unless you are thinking that the API could fail between those lines, resulting in similar behavior. 

yes. Kubernetes API could become unavailable at any time, Traefik's behavior shouldn't be dependent on the timing of it.

- Either the errors are propagated up the call stack to provider sync loop, or
- To keep the logic simple and correct,  it is better to pass the `serverVersion` obtained at https://github.com/traefik/traefik/blob/0a3e40332a86cc061bf16d8895b10defeb999171/pkg/provider/kubernetes/ingress/kubernetes.go#L187 as method argument to all other calls.  Once Traefik decides to drop support for very old Kubernetes API versions, the argument can be cleaned up from all those methods.

dtomcej:
@ravilr,

Gotcha. Totally agree with you on this one. Will work on an update in this PR.

dtomcej:
@ravilr What do you think of this?

In this case, the serverVersion is stored in the original `WatchAll()` after factory setup. This means that if the API connection is lost, the lister won't update, but neither will the cached serverVersion. The version is refreshed when the watch all returns.

By utilizing the interface and the client to cache the result, we avoid having to recode every instance, or modify methods.

pkg/provider/kubernetes/ingress/client.go
```diff
@@ -208,6 +209,9 @@ func (c *clientWrapper) WatchAll(namespaces []string, stopCh <-chan struct{}) (<
 		}
 	}
 
+	// Reset the stored server version to recheck on reconnection.
+	c.serverVersion = nil
```
ravilr, line 213: I'm not sure if this ever gets reset during the lifetime of Traefik once started.

ravilr:
This looks fine, but it is not clear to me whether WatchAll() gets invoked again after the initial Traefik startup. 

Also, unrelated to this change, while reviewing this change, noticed that the done signal passed here https://github.com/traefik/traefik/blob/0a3e40332a86cc061bf16d8895b10defeb999171/pkg/provider/kubernetes/ingress/kubernetes.go#L112 seems to be the traefik's main() background context, which means that the operation() func could be retried without stopping the previously setup Informer watches. Don't we want the Informer watches to be stopped before creating new ones, when the operation() func is retried ?

dtomcej:
> This looks fine, but it is not clear to me whether WatchAll() gets invoked again after the initial Traefik startup, which could mean that the Traefik may not recognize the server version change across Kubernetes API rolling upgrades, unless Traefik is also restarted ?
> 
> Also, unrelated to this change, while reviewing this change, noticed that the done signal passed here
> 
> https://github.com/traefik/traefik/blob/0a3e40332a86cc061bf16d8895b10defeb999171/pkg/provider/kubernetes/ingress/kubernetes.go#L112
> 
> seems to be the traefik's main() background context, which means that the operation() func could be retried without stopping the previously setup Informer watches. Don't we want the Informer watches to be stopped before creating new ones, when the operation() func is retried ?

@ldez what do you think about this? Should we be creating a new context to pass into this method?

My understanding was that the whole `watchAll()` would be retried if the operation returned with an error. Am I wrong?

Should we modify this code?

pkg/provider/kubernetes/ingress/client.go
rtribotte: @dtomcej Moreover, I think we can put the logic of retrieving the server version in the `WatchAll` method, instead of the `GetServerVersion`, and then change `GetServerVersion` method to be only a "getter" method.
So we can remove the error from the signature.

pkg/provider/kubernetes/ingress/client.go
```diff
@@ -427,12 +431,22 @@ func (c *clientWrapper) lookupNamespace(ns string) string {
...
+		return c.serverVersion, nil
+	}
+
+	serverVersionInfo, err := c.clientset.Discovery().ServerVersion()
 	if err != nil {
 		return nil, fmt.Errorf("could not retrieve server version: %w", err)
 	}
 
-	return version.NewVersion(serverVersion.GitVersion)
+	serverVersion, err := version.NewVersion(serverVersionInfo.GitVersion)
```
dtomcej, line 443: move this to the watchall

phin1x:
can we get this merged as soon possible. would it also be possible to include this patch in the 2.5 release?

SantoDE:
We plan to have it in the next 2.4 release and thus, also in 2.5

rtribotte (approved):
LGTM 👍

jbdoumenjou (approved):
LGTM

mpl (approved):

PR #8223, opened 2021-06-24
Improve empty configuration check

rtribotte:
### What does this PR do?

This PR improves the empty configuration check in the configuration watcher.
Currently, if a provider sends an empty configuration with maps initialized in the HTTP/TCP/UDP section, it is considered as a nonempty configuration, which would then be applied.

### Motivation

Have a valid check of the configuration to evaluate it's empty or not.

### More

- [x] Added/updated tests

### Additional Notes

This PR is an enhancement as it doesn't address any known issue regarding the empty configuration check.
Thus, as some providers (e.g Ingress provider) would send an empty configuration when an error is encountered, this will ensure that this empty configuration is not applied over the existing configuration.

### DIFF
--- a/pkg/server/configurationwatcher.go
+++ b/pkg/server/configurationwatcher.go
@@ -251,10 +251,10 @@ func isEmptyConfiguration(conf *dynamic.Configuration) bool {
 		conf.UDP = &dynamic.UDPConfiguration{}
 	}
 
-	httpEmpty := conf.HTTP.Routers == nil && conf.HTTP.Services == nil && conf.HTTP.Middlewares == nil
-	tlsEmpty := conf.TLS == nil || conf.TLS.Certificates == nil && conf.TLS.Stores == nil && conf.TLS.Options == nil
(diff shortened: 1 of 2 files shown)


rtribotte:
This PR breaks the reset of the configuration for providers who were producing empty configurations with initialized structures.
Still, we should consider a clarification on what is an empty configuration, and moreover when a configuration should be applied or not.
We might have to consider that a nil configuration is not applicable, which would mean that there was an error during the build of the configuration. And therefore consider not nil configurations as applicable since it's a way to reset the configuration.
By working on this PR, we identified that currently, any KV provider returns a non-nil configuration with no substructures initialized when there is no configuration from pairs, therefore it's not possible to reset totally a previously applied configuration.
Regarding the HTTP provider, we identified that resetting the configuration would only work by providing a configuration defining at least one of the HTTP/TCP/UDP/TLS sections, and not by just providing and empty JSON `{}`.

PR #8212, opened 2021-06-22
docs: fix invalid subdomain

WLun001:
### What does this PR do?
Update middleware headers docs to comply with [RFC 1123 subdomain](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names)

It will throw the following error if directly copy from [docs](https://doc.traefik.io/traefik/middlewares/headers/) and apply to Kubernetes
```bash
The Middleware "testHeader" is invalid: metadata.name: Invalid value: "testHeader": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

PR #8226, opened 2021-06-24
Fix: malformed Kubernetes resource names and references in tests

rtribotte:
### What does this PR do?

This PR fixes malformed Kubernetes resource names in test fixtures.

### Motivation

Be consistent with [Kubernetes validation constraints](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names).
It follows PR #8212 which was focused on documentation.

### More

- [x] Added/updated tests

### Additional Notes

This PR does not affect the tests in practice, since the fixtures impacted are used in unit tests and are loaded through the mock client which is not making validation against resources, and so tolerate any string value as a name inside a resource.
Therefore, the goal of this PR is only to sanitize fixtures, so it will hopefully prevent using malformed names later on.

### DIFF
--- a/pkg/provider/kubernetes/crd/fixtures/tcp/with_bad_tls_options.yml
+++ b/pkg/provider/kubernetes/crd/fixtures/tcp/with_bad_tls_options.yml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: secretCA1
+  name: secret-ca1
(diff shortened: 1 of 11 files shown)


jbdoumenjou (approved):
LGTM

ldez (approved):
LGTM

juliens (approved):
LGTM

PR #8237, opened 2021-06-29
kv: support /traefik/alias prefix from v1 again

fho:
### What does this PR do?

This commit adds support for atomic configuration changes in the KV
stores by allowing to change the key-prefix from that the configuration
is retrieved.

If the traefik/alias entry exist in the KV store, it's value is used as
prefix instead of the one from p.RootKey.

The implementation is basically the same then the one for Traefik v1:
- https://github.com/traefik/traefik/pull/203
- https://doc.traefik.io/traefik/v1.7/user-guide/kv-config/#atomic-configuration-changes

### Motivation

Support atomic configuration changes when using the consul kv store.

### More

- [ ] Added/updated tests

### Additional Notes

I tested it locally and it works.
Please let me know if you are willing to accept the PR and if there are any reasons that make this solution not viable for traefik v2.

I can extend the PR with docs and tests if wanted.

This addresses issue https://github.com/traefik/traefik/issues/7419.

### DIFF
--- a/pkg/provider/kv/kv.go
+++ b/pkg/provider/kv/kv.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/abronan/valkeyrie"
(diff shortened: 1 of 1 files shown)


pkg/provider/kv/kv.go
```diff
@@ -143,14 +144,27 @@ func (p *Provider) watchKv(ctx context.Context, configurationChan chan<- dynamic
 	return nil
 }
 
+func (p *Provider) configKeyPrefix() string {
+	aliasKey := p.RootKey + "/alias"
+
+	res, err := p.kvClient.Get(aliasKey, nil)
+	if err != nil {
```
fho, line 151: Instead of always ignoring the error, we could only ignore errors that indicate that the entry does not exist (if the kvClient has an error type for that), all others could be returned from buildConfiguration().

PR #8260, opened 2021-07-09
Disable Cross-Namespace by default for IngressRoute provider

dtomcej:
### What does this PR do?

Disables the `allowCrossNamespace` by default for IngressRoute provider

### Motivation

Security

### More

- [x] Added/updated tests
- [x] Added/updated documentation

PR #8261, opened 2021-07-09
Disable ExternalName Services by default on Kubernetes providers

dtomcej:
### What does this PR do?

This PR:

- Disables ExternalName services by default in the KubernetesCRD provider
- Disables ExternalName services by default in the KubernetesIngress provider
- Exposes a flag to allow ExternalName services on both providers
- Adds documentation to explain the flag
- Does not create a router for invalid services (aka ExternalName services when not allowed) to maintain current behavior, as this is a bugfix, and not a PR to change core server behavior.

### Motivation

Security

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes.go
+++ b/pkg/provider/kubernetes/crd/kubernetes.go
@@ -38,15 +38,22 @@ const (
 
 // Provider holds configurations of the provider.
 type Provider struct {
-	Endpoint            string          `description:"Kubernetes server endpoint (required for external cluster client)." json:"endpoint,omitempty" toml:"endpoint,omitempty" yaml:"endpoint,omitempty"`
-	Token               string          `description:"Kubernetes bearer token (not needed for in-cluster client)." json:"token,omitempty" toml:"token,omitempty" yaml:"token,omitempty"`
-	CertAuthFilePath    string          `description:"Kubernetes certificate authority file path (not needed for in-cluster client)." json:"certAuthFilePath,omitempty" toml:"certAuthFilePath,omitempty" yaml:"certAuthFilePath,omitempty"`
-	Namespaces          []string        `description:"Kubernetes namespaces." json:"namespaces,omitempty" toml:"namespaces,omitempty" yaml:"namespaces,omitempty" export:"true"`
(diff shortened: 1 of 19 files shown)


dtomcej:
Should have this PR create the router to allow the user to see that the service creation encountered an error.

pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -38,20 +38,22 @@ const (
...
+	LabelSelector             string          `description:"Kubernetes label selector to use." json:"labelSelector,omitempty" toml:"labelSelector,omitempty" yaml:"labelSelector,omitempty" export:"true"`
+	IngressClass              string          `description:"Value of kubernetes.io/ingress.class annotation to watch for." json:"ingressClass,omitempty" toml:"ingressClass,omitempty" yaml:"ingressClass,omitempty" export:"true"`
+	ThrottleDuration          ptypes.Duration `description:"Ingress refresh throttle duration" json:"throttleDuration,omitempty" toml:"throttleDuration,omitempty" yaml:"throttleDuration,omitempty" export:"true"`
+	lastConfiguration         safe.Safe
 }
 
 // SetDefaults sets the default values.
 func (p *Provider) SetDefaults() {
-	p.AllowCrossNamespace = func(b bool) *bool { return &b }(true)
+	p.AllowCrossNamespace = true
```
dtomcej, line 55: can remove this method once #8260  is merged, and this value switches to false

rtribotte (approved):
LGTM

ldez (approved):
LGTM

lawrencedudley:
@dtomcej I really feel like this should at the very least have been in a point release :(

It's categorically a breaking change

ldez:
This change has been made because of a Kubernetes CVE: https://github.com/kubernetes/kubernetes/issues/103675

lawrencedudley:
Fair enough, I do mostly understand the abundance of caution there

dtomcej:
@lawrencedudley, as mentioned by @ldez, due to the fact that this was a security patch due to the CVE, even if we were to apply it to the new minor release (2.5), our policy would be to backport it to the latest stable release anyways.

We did discuss this together as a team, but decided that ultimately its importance for security outweighed the inconvenience that users may encounter with updating.

Thanks for your understanding!

tomMoulard:
Hello @wodka,

Thanks for your interest in Traefik !

> if something like this pops up a note on the website would be great

Actually, there is : https://doc.traefik.io/traefik/v2.5/migration/v2/#k8s-externalname-service

PR #8286, opened 2021-07-20
Get Kubernetes server version early

rtribotte:
### What does this PR do?

This PR refactor changes from PR #8185.
It allows failing early if the server version cannot be fetched.

It also changes the return of the `WatchAll` method in case of error when fetching the server version, in which case, the event chan returned would be nil.
As per this modification, a change in the unit tests of the client allows setting a valid version in the fake client to avoid this error case.

### Motivation

Re-introduce fixes that were lost during a rebase in PR #8185.

### More

- [x] Added/updated tests

### Additional Notes

It is also more compliant with further changes on the Ingress provider on branch v2.5.


ldez (approved):
LGTM

mpl (approved):

PR #8290, opened 2021-07-20
feat: Add new CI system

mmatur:
### What does this PR do?

This PR removes SemaphoreCI v1 and travis and switch to Semaphoreci v2 and Github action

### Motivation

To have a a better CI

### More

- ~[ ] Added/updated tests~


.semaphore/semaphore.yml
```diff
@@ -0,0 +1,94 @@
+version: v1.0
+name: Traefik
+agent:
+  machine:
+    type: e1-standard-4
+    os_image: ubuntu1804
```
dtomcej, line 6:
```suggestion
    os_image: ubuntu2004
```

Since the rest use 2004?

dtomcej: 2004 only works with e1-standard-2 machines: https://docs.semaphoreci.com/ci-cd-environment/ubuntu-20.04-image/

Disregard.

ldez: it's related to the integration test and groups v2

.github/workflows/validate.yaml
```diff
@@ -0,0 +1,115 @@
...
+      - 'docs/'
+
+jobs:
+  validate:
+    runs-on: ubuntu-20.04
+    env:
+      GO_VERSION: 1.16
+      GOLANGCI_LINT_VERSION: v1.41.1
+      MISSSPELL_VERSION: v0.3.4
+      PRE_TARGET: ""
```
dtomcej, line 24: Should these be set on the task/block instead of on the job?

dtomcej, line 24: Should these be set for all jobs instead of redeclaring them in each job?

dtomcej (approved):
LGTM
:shipit:

ldez (approved):

jbdoumenjou (approved):
LGTM

PR #8291, opened 2021-07-20
If applied, this commit will Upgrade indirect dependency gogo/protobuf

tspearconquest:
Signed-off-by: Thomas Spear <tspear@conquestcyber.com>

### What does this PR do?

Upgrades the indirect dependency gogo/protobuf

### Motivation

The vulnerablity is causing our CI/CD pipeline to fail upon attempting to deploy the v2.4.11 release of Traefik's Docker image.
While protobuf is not directly used by Traefik, it is present in the binary and causes the app to fail vulnerability scans by Aquasecurity Trivy open source vulnerability scanner.

### More

- [ ] Added/updated tests


ldez:
Hello,
As I answered in https://github.com/traefik/traefik/issues/8283#issuecomment-883396726 we don't need to do that.

We will declined your PR.

PR #8299, opened 2021-07-23
gateway api: support RouteNamespaces

### LINKED ISSUE #8246: Support Gateway Namespaces for Kubernetes GatewayAPI
### Welcome!

We are looking to use Traefik as an implementation of the K8s GatewayAPI, we wanted to use a single Gateway managed by one team to provide the ability for teams using different namespaces to use. 

We have found that the implementation only looks for HTTPRoutes in the same namespace as the Gateway as per: https://github.com/traefik/traefik/blob/master/pkg/provider/kubernetes/gateway/kubernetes.go#L488

We currently would need to create a Gateway per-namespace for Traefik to pickup HTTPRoutes.

We have a listner configured on a GW with Namespaces: All
```yaml
  Listeners:
    Port:      8000
    Protocol:  HTTP
    Routes:
      Group:  networking.x-k8s.io
      Kind:   HTTPRoute
      Namespaces:
        From:  All
      Selector:
        Match Labels:
          App:  traefik
```

but it only looks in the Default NS (where the Gateway is deployed)

tomMoulard:
### What does this PR do?

This PR adds support for the `namespace` directive in a {HTTP,TCP,TLS}Route.
It allows routes [filtering](https://gateway-api.sigs.k8s.io/concepts/security-model/#route-namespaces) based on namespaces.

### Motivation

Close #8246

### More

- [X] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/provider/kubernetes/gateway/kubernetes.go
+++ b/pkg/provider/kubernetes/gateway/kubernetes.go
@@ -492,8 +492,23 @@ func (p *Provider) fillGatewayConf(ctx context.Context, client Client, gateway *
 }
 
 func gatewayHTTPRouteToHTTPConf(ctx context.Context, ep string, listener v1alpha1.Listener, gateway *v1alpha1.Gateway, client Client, conf *dynamic.Configuration) []metav1.Condition {
-	selector := getRouteBindingSelectorSelector(listener.Routes.Selector)
-	namespace, err := getRouteBindingSelectorNamespace(gateway.Namespace, listener.Routes.Namespaces)
+	selector := labels.Everything()
+
+	if listener.Routes.Selector != nil {
+		var err error
+		selector, err = metav1.LabelSelectorAsSelector(listener.Routes.Selector)
+		if err != nil {
+			return []metav1.Condition{{
+				Type:               string(v1alpha1.ListenerConditionResolvedRefs),
+				Status:             metav1.ConditionFalse,
+				LastTransitionTime: metav1.Now(),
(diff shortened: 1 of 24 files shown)


pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -492,21 +492,28 @@ func (p *Provider) fillGatewayConf(ctx context.Context, client Client, gateway *
...
 	}
 
-	httpRoutes, err := client.GetHTTPRoutes(gateway.Namespace, selector)
+	namespace := gateway.Namespace
+	if listener.Routes.Namespaces != nil && listener.Routes.Namespaces.From != nil {
+		// TODO: Implement v1alpha1.RouteSelectSelector
+		if *listener.Routes.Namespaces.From == v1alpha1.RouteSelectAll {
+			namespace = metav1.NamespaceAll
+		}
+	}
```
dtomcej, line 506: Should this be in a method since it is re-used? This would also allow `v1alpha1.RouteSelectSelector` to be implemented in one place, rather than replicated.

tomMoulard: Solved with https://github.com/traefik/traefik/pull/8299/commits/fa127abb8730d2d17b751a188c55c9fb9c9f4a90 :)

jbdoumenjou (approved):
LGTM

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks 👍

PR #8308, opened 2021-07-26
Add configurable tags to influxdb metrics

Tetha:
### What does this PR do?

This allows users to set additional influxdb tags on their traefik instances so they can distinguish different traefik instances in the influxdb. For users with existing influxdb changes, this PR changes nothing during an update and the old configuration functions identically, except that they can improve it afterwards.

### Motivation

We have a certain fleet of traefik instances across different production environments and with different specializations. When I enabled the influxdb metric collection, it was an easy experience to do so, but it resulted in a bit of a mess on the influxdb side. With this pull request, I can add different tags to my traefik instances, so I can distinguish them.

There also is an issue about this, #6997. However, I don't think introducing default tags is a good strategy. Quite a few tools try to do so and just end up making more work because the default tags might not apply to all situations. For example, a default host tag would just be tag-spam for our container-based instances. Just pushing a map into the influxdb client allows all users to adapt the  tags to their specific setup. 

### More

- [x] Added/updated tests - I have modified one unit test in `influxdb_metrics`.  I've also done some manual testing in case something weird explodes. :)
- [x] Added/updated documentation - I have modified the Observability chapter.

### Additional Notes

As mentioned in the commit message, overall, this was pretty simple. I practically just moved the empty list from a hardcoded call in the influx-client constructor to the configuration type and that was all I had to do.

However, there was one snag when changing the tests. There are two tests in place, one for udp and one for http. I figured - lets run one test with additional tags, and one without. That way, I'd know that both setting additional tags works, and not setting additional tags works. 

However, I discovered that only the first call to `RegisterInfluxDB` across both tests had an effect. This is because on master, `RegisterInfluxDB` sets the global `influxDBClient` if and only if it is nil. 

I figured it was not entirely clean that these tests are coupled in this way and figured there are two ways to decouple this - either `RegisterInfluxDB` has to always set the `influxDBClient`, or `StopInfluxDB` has to clear the `influxDBClient` after stopping the ticker. After some deliberation, I figured that clearing the `influxDBClient` in the `StopInfluxDB` method has a higher chance of complications - what if the ticker runs right now, or what if the ticker takes some time to stop, what if the client has some metrics buffered, ...?

Thus, I changed the `RegisterInfluxDB` method to always overwrite the `influxDBClient` when called. This should be fine for now, because as far as I can see the code, the `RegisterInfluxDB` method is called once per traefik instance start, and in some test cases. So at the moment, as far as I can see, the if-statement I removed was only exercised in tests.  This might become a problem with some kind of config reload for the metrics, but that wouldn't have worked either way.

### DIFF
--- a/docs/content/observability/metrics/influxdb.md
+++ b/docs/content/observability/metrics/influxdb.md
@@ -246,7 +246,7 @@ Additional labels (influxdb tags) on all metrics.
 [metrics]
   [metrics.influxDB]
     [metrics.influxDB.additionalLabels]
-      host = "traefik01.example.com"
+      host = "example.com"
       environment = "production"
 ```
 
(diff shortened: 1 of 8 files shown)


rtribotte:
Hello @Tetha,

Glad to see this new PR opened, thank you!

rtribotte:
@Tetha Can you edit your PR to have a clear description of what the PR is doing?

Tetha:
I have included my original description.

rtribotte (approved):
LGTM

dtomcej (approved):
LGTM
:shipit:

kevinpollet (approved):
Thanks 👍

PR #8310, opened 2021-07-27
Add `HeadersDistrib` rule

wizawu:
### What does this PR do?

This PR adds another routing rule named `HeadersDistrib` to route requests to fixed backend (without cookie's help) based on its header (value). Meanwhile, each backend can have different weight to accept requests. For example,

```
router-hello:
  rule: HeadersDistrib(`X-User-Id`, `0`, `0.75`)
  service: service-hello

router-world:
  rule: HeadersDistrib(`X-User-Id`, `0.75`, `1`)
  service: service-world
```

### Motivation

We were looking for a router for A/B testing (and we have been used traefik since v2.2 released). The backends are different machine learning model services. Each request can be marked with the ID of a front-end user (and we decided to put it in HTTP header).

The problem is, when we do A/B testing, we would like to route a certain percentage (like 75%) of requests to the old service backend and the rest (like 25%) to the new one, and make sure that all the requests with the same user ID go to the same backend. So that we can get the complete profile of a user from either old services or new services.

The solution is simple here. I just hash the specified header value (e.g. `X-User-Id`) and distribute the hash result to [0, 1). Then each router can be assigned a range to match the distributed result.

### More

- [x] Added/updated tests

### Additional Notes

* I am not a native English speaker, so the name `HeadersDistrib` may not be accurate for its usage. Please advise me another proper rule name **if** this routing feature is useful.
* The documentation is not updated in this PR, since I wonder this feature can be widely used and merged :)

### DIFF
--- a/pkg/rules/rules.go
+++ b/pkg/rules/rules.go
@@ -1,8 +1,11 @@
 package rules
 
 import (
+	"errors"
 	"fmt"
+	"hash/fnv"
 	"net/http"
+	"strconv"
 	"strings"
 	"unicode/utf8"
 
@@ -14,16 +17,17 @@ import (
 )
 
 var funcs = map[string]func(*mux.Route, ...string) error{
-	"Host":          host,
-	"HostHeader":    host,
-	"HostRegexp":    hostRegexp,
-	"ClientIP":      clientIP,
-	"Path":          path,
-	"PathPrefix":    pathPrefix,
-	"Method":        methods,
-	"Headers":       headers,
-	"HeadersRegexp": headersRegexp,
-	"Query":         query,
+	"Host":           host,
+	"HostHeader":     host,
+	"HostRegexp":     hostRegexp,
+	"ClientIP":       clientIP,
+	"Path":           path,
+	"PathPrefix":     pathPrefix,
+	"Method":         methods,
+	"Headers":        headers,
+	"HeadersRegexp":  headersRegexp,
+	"HeadersDistrib": headersDistrib,
+	"Query":          query,
 }
 
 // Router handle routing with rules.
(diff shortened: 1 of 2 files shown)


rtribotte:
Hello @wizawu,

Thanks for your contribution!

After a first look, we are wondering why the weighted loadBalancer combined with sticky cookies doesn't achieve the same goal?

Here is an example of configuration:

```yaml
## Dynamic configuration
http:
  services:
    wrr1:
      weighted:
        sticky:
          cookie:
            name: lvl1
        services:
          - name: service-hello
            weight: 3
          - name: service-world
            weight: 1

    service-world:
      loadBalancer:
        servers:
          - url: http://127.0.0.1:8082

    service-hello:
      loadBalancer:
        servers:
          - url: http://127.0.0.1:8084
```

wizawu:
To be more clear, let's say the router (traefik)  is placed between 3 Kubernetes deployments and each has **4** pods.

```
deployement A -> traefik -- 75% --> deployment B
                        \-- 25% --> deployment C
```

There're two key points in our implementation.
- Each user (identified by a unique ID, no matter we put it in header, cookie, query parameter, or hash it to traefik service name) must be routed to the same deployment (B or C).
- The ratio of the amount of **users** routed to B and to C should be 75% : 25% (or we say 3 : 1).

Now back to the sticky cookie solution, which we do notice before. There are a few pitfalls when apply it to our scenario.
- The original request is produced by a backend service. We need extra work to preserve the cookie (with no browser). Or we need to pre-generete cookie (e.g. `Cookie: lvl1=service-hello`) in the upstream deployment A.
- As you notice, we have 4 pod in deployment A. We have to make sure that all these pods share the same cookie for a single user.
- There is a slight difference between `weight` and distribution ratio. For example, if each pod in deployment A send a request representing user X for the first time, since none of those HTTP clients has cookies, 3 requests may go to B and 1 to C. On the other hand, when routed by `HeaderDistrib`, requests with the same header value will be always "distributed" to the same backend deployment.
- NIT: It's quite common to have tracking ID in headers rather than cookie, such as [X-Request-ID](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#generate-request-id) in ingress.

After all, we do admit that there are few scenarios to apply the `HeaderDistrib` match rule. Please feel free to close the PR. And surely we can just continue to use a forked version :)

Regards.

ddtmachado:
Hi @wizawu , if I understood your use case the only real blocker for using the example referenced by @rtribotte would be the fact that you want to determine (hash) the source criteria to apply stickiness from a header instead of a cookie right? 

Other than that the weighted round robin seems to fit well for distribution if used in conjunction with multiple routers, something that would look like this:

```yaml
http:
  services:  
    app-svc:
      weighted:
        sticky:
          #New sticky option
          header:
            name: X-User-Id

        services:
        - name: appv1
          weight: 3
        - name: appv2
          weight: 1

    appv1:
      loadBalancer:
        servers:
          - url: "http://127.0.0.1:8080"
        healthCheck:
          scheme: http
          path: /health

    appv2:
      loadBalancer:
        servers:
          - url: "http://127.0.0.1:8180"
        healthCheck:
          scheme: http
          path: /health

    #Redirect to traefik itself for prerouting transformations
    traefik-svc:
      loadbalancer:
        servers:
          - url: http://traefik.local.address

  routers:
    dashboard:
      rule: Host(`traefik.docker.localhost`)
      service: api@internal

    app-without-header:
      rule: Host(`app.docker.localhost`)
      service: traefik-svc
      middlewares:
        # Custom plugin that would hash the source req based on custom logic
        # and then inject the Header "X-User-Id" with either appv1 or appv2
        - my-custom-plugin 

    app-with-header:
      rule: Host(`app.docker.localhost`) && Header(`X-User-Id`)
      service: app-svc
```

I prefer this approach as it would not mix the router matching rule with the service load balancing distribution in the config.

Would something like that work for you?

PR #8313, opened 2021-07-28
Kubernetes: detect changes for resources other than endpoints

rtribotte:
### What does this PR do?

This PR fixes the resources changes detection inside the Kubernetes event handler.

### Motivation

Be able to detect changes for all resources and generate a new configuration accordingly.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/k8s/event_handler.go
+++ b/pkg/provider/kubernetes/k8s/event_handler.go
@@ -48,12 +48,10 @@ func objChanged(oldObj, newObj interface{}) bool {
 	}
 
 	if _, ok := oldObj.(*corev1.Endpoints); ok {
-		if endpointsChanged(oldObj.(*corev1.Endpoints), newObj.(*corev1.Endpoints)) {
-			return true
-		}
+		return endpointsChanged(oldObj.(*corev1.Endpoints), newObj.(*corev1.Endpoints))
 	}
 
-	return false
+	return true
 }
 
 func endpointsChanged(a, b *corev1.Endpoints) bool {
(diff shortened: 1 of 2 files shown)


jbdoumenjou (approved):
LGTM

mpl (approved):

pkg/provider/kubernetes/k8s/event_handler.go
```diff
@@ -48,12 +48,10 @@ func objChanged(oldObj, newObj interface{}) bool {
...
 
 	if _, ok := oldObj.(*corev1.Endpoints); ok {
-		if endpointsChanged(oldObj.(*corev1.Endpoints), newObj.(*corev1.Endpoints)) {
-			return true
-		}
+		return endpointsChanged(oldObj.(*corev1.Endpoints), newObj.(*corev1.Endpoints))
 	}
 
-	return false
+	return true
```
dtomcej, line 54: Should we keep this false, and invert the condition in line 46 instead?
```suggestion
	return false
```

dtomcej:
```
if oldObj.(metav1.Object).GetResourceVersion() != newObj.(metav1.Object).GetResourceVersion() {
	return true
}
```

jbdoumenjou: If we do that, it will return true as soon as the version will be different and the endpointsChanged will be useless.

dtomcej: Ah yep.  Definitely overlooked that.

dtomcej (commented):
LGTM
:shipit:

dtomcej (approved):

PR #8319, opened 2021-07-29
fix: remove hop-by-hop headers define in connection header beore some middleware

ldez:
### What does this PR do?

Clean connection header added by the header and the forward auth middleware.

### Motivation

Don't allow to drop a header by spoofing the `Connection` header.

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/middlewares/auth/forward.go
+++ b/pkg/middlewares/auth/forward.go
@@ -15,6 +15,7 @@ import (
 	"github.com/traefik/traefik/v2/pkg/config/dynamic"
 	"github.com/traefik/traefik/v2/pkg/log"
 	"github.com/traefik/traefik/v2/pkg/middlewares"
+	"github.com/traefik/traefik/v2/pkg/middlewares/connectionheader"
 	"github.com/traefik/traefik/v2/pkg/tracing"
 	"github.com/vulcand/oxy/forward"
 	"github.com/vulcand/oxy/utils"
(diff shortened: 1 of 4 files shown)


juliens (approved):
LGTM

jbdoumenjou (approved):
LGTM

pkg/middlewares/auth/forward.go
```diff
@@ -89,7 +90,7 @@ func NewForward(ctx context.Context, next http.Handler, config dynamic.ForwardAu
 		fa.authResponseHeadersRegex = re
 	}
 
-	return fa, nil
+	return connectionheader.Remove(fa), nil
```
mpl, line 93: This call does not remove anything, so I think it is misleading. I'd call the method something like "Remover" instead, as it returns something that is in charge of removing.

pkg/middlewares/connectionheader/connectionheader.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+	}
+}
+
+func upgradeType(h http.Header) string {
+	if !httpguts.HeaderValuesContainsToken(h["Connection"], "Upgrade") {
+		return ""
+	}
+	return h.Get("Upgrade")
+}
```
mpl, line 44: is there any particular need for this func? it is only 4 lines and called in only one location...

mpl (approved):

PR #8368, opened 2021-08-19
Replace x/sys for go 1.17

roopakv:
### What does this PR do?

Updates x/sys to support go 1.17 and unblock brew. More details here: https://github.com/Homebrew/homebrew-core/pull/83413

### Motivation

brew is blocked

### Additional Notes

Please cut a release after merging this

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -109,4 +109,5 @@ replace (
 	github.com/gorilla/mux => github.com/containous/mux v0.0.0-20181024131434-c33f32e26898
 	github.com/mailgun/minheap => github.com/containous/minheap v0.0.0-20190809180810-6e71eb837595
 	github.com/mailgun/multibuf => github.com/containous/multibuf v0.0.0-20190809014333-8b6c9a7e6bba
+	golang.org/x/sys => golang.org/x/sys v0.0.0-20210817190340-bfb29a6856f2
 )
--- a/go.sum
+++ b/go.sum
@@ -1439,6 +1439,8 @@ golang.org/x/sys v0.0.0-20210225134936-a50acf3fe073/go.mod h1:h1NjWce9XRLGQEsW7w
 golang.org/x/sys v0.0.0-20210303074136-134d130e1a04/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20210330210617-4fbd30eecc44 h1:Bli41pIlzTzf3KEY06n+xnzK/BESIg2ze4Pgfh/aI8c=
 golang.org/x/sys v0.0.0-20210330210617-4fbd30eecc44/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20210817190340-bfb29a6856f2 h1:c8PlLMqBbOHoqtjteWm5/kbe6rNY2pbRfbIMVnepueo=
+golang.org/x/sys v0.0.0-20210817190340-bfb29a6856f2/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/term v0.0.0-20201117132131-f5c789dd3221/go.mod h1:Nr5EML6q2oocZ2LXRh80K7BxOlk5/8JxuGnuhpl+muw=
 golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1/go.mod h1:bj7SfCRtBDWHUb9snDiAeCFNEtKQo2Wmx5Cou7ajbmo=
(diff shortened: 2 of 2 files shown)


go.mod
```diff
@@ -109,4 +109,5 @@ replace (
 	github.com/gorilla/mux => github.com/containous/mux v0.0.0-20181024131434-c33f32e26898
 	github.com/mailgun/minheap => github.com/containous/minheap v0.0.0-20190809180810-6e71eb837595
 	github.com/mailgun/multibuf => github.com/containous/multibuf v0.0.0-20190809014333-8b6c9a7e6bba
+	golang.org/x/sys => golang.org/x/sys v0.0.0-20210817190340-bfb29a6856f2
```
thaJeztah, line 112: This will pin it to this exact version. Wondering, does it work if you add golang.org/x/sys to "require" with an `// indirect`, to force a more current version?

ldez: @thaJeztah you are right: the replace directive must not be used to upgrade/pin a version, the right way is to use `indirect`

mpl:
@roopakv 
Hello.
Looking at https://github.com/Homebrew/homebrew-core/pull/83413 , I'm a bit confused by your diagnostic, especially since this issue is lacking details about what is failing. Notably:

1) it seems like when you were discussing this issue, and if I'm looking at the history on https://github.com/Homebrew/homebrew-core/commits/c1700179ceb4878d580dac84fa996fd41e898914/Formula/traefik.rb , when you were running these tests, you were not on traefik 2.5, but on traefik 2.4 (which does not support Go 1.17). Am I wrong, can you clarify that?

2) As of now, if I'm running brew test (on Catalina) with:

```
mbp:traefik mpl$ brew info traefik
traefik: stable 2.5.0 (bottled), HEAD
Modern reverse proxy
https://traefik.io/
/usr/local/Cellar/traefik/2.5.0 (8 files, 99.4MB) *
  Poured from bottle on 2021-08-19 at 15:00:43
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/traefik.rb
```

I get

```
mbp:traefik mpl$ brew test --retry --verbose traefik
/usr/local/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.3_2/bin/bundle install
Fetching gem metadata from https://rubygems.org/.........
...
Installing rubocop-performance 1.11.5
Bundle complete! 33 Gemfile dependencies, 72 gems now installed.
Bundled gems are installed into `../../../../../../usr/local/Homebrew/Library/Homebrew/vendor/bundle`
Removing rubocop-performance (1.11.4)
==> Testing traefik
/usr/bin/sandbox-exec -f /private/tmp/homebrew20210819-12914-1pyb6c8.sb ruby -W1 -- /usr/local/Homebrew/Library/Homebrew/test.rb /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core/Formula/traefik.rb --verbose --retry
INFO[0000] Configuration loaded from file: /private/tmp/traefik-test-20210819-12915-g8tcsm/traefik.toml
==> curl -sIm3 -XGET http://127.0.0.1:52417/
==> curl -sIm3 -XGET http://127.0.0.1:52416/dashboard/
==> /usr/local/Cellar/traefik/2.5.0/bin/traefik version 2>&1
```

which seems to be working.

3) If the problems are only limited to ARM (not sure, as the linked PR is confusing on that aspect), which would explain why the test is working on my non-ARM laptop, then can you tell us what exactly is failing on ARM? it is true that we do not run any tests for ARM builds, but at least we provide binaries for all architectures, so we do know that at least traefik 2.5 is building on ARM.

roopakv:
@mpl 

Take a look at this: https://github.com/golang/go/issues/45702

Basically with go 1.17 traefik will not build unless x/sys is upgraded

I can write up something a bit more detailed in a bit :)

ldez:
@roopakv currently traefik (v2.5.0) is built with go1.17

roopakv:
@ldez interesting. Let me see why the arm bottle failed again. I was sure that all failures I collected were the x/sys issue.

mpl:
> @ldez interesting. Let me see why the arm bottle failed again. I was sure that all failures I collected were the x/sys issue.

Yeah, in my point 3. above, it was implicit but I meant we are building traefik, for ARM, with go 1.17.

thaJeztah:
If someone knows which commit in `x/sys` fixes the issue, then people could check if the version they use is has that commit; do we know which specific commit fixes the issue?

roopakv:
@mpl from the logs on the golang brew upgrade: 
image


this is definitely the x/sys issue. Upgrading x/sys as mentioned here: https://github.com/golang/go/issues/46763
and https://github.com/golang/go/issues/45702 will unblock building traefick with go 1.17 and unblock homebrew as well.

ldez:
Can you provide the hash or the tag of traefik that you are trying to build?

roopakv:
`531a8ff24843bcd903c195dcf418dbb0eb5c8e0d` with go 1.17

ldez:
As suggested by @thaJeztah, could you point the exact commit in `x/sys` that fixes the problem?

roopakv:
@ldez this is the commit that fixes the problem in x/sys https://github.com/golang/sys/commit/a76c4d0a0096537dc565908b53073460d96c8539

PR #8393, opened 2021-08-23
OCSP stapling

### LINKED ISSUE #212: OCSP stapling support
Let's Encrypt for example also provides a lightweight chain file for OCSP stapling 
(nginx e.g. supports this with option stapling_verify).

So the user should be able to provide a chain file (like nginx ssl_trusted_certificate option) for traefik.

Thank you and 
with best regards

alekitto:
### What does this PR do?

Add OCSP stapling support

### Motivation

I needed OCSP stapling support as I experienced very slow connection performance on iOS devices [this article explains why there's a delay with letsencrypt certs on iOS](https://www.programmersought.com/article/44346829822/), so I wrote a prototype for supporting OCSP stapling in treafik (also checking how caddyserver does it) to make it work in my environment.
Now it's working flawlessly since two weeks ago, but I think I need some guidance on how to structure this more consistently with the rest of the codebase and some suggestions to make this configurable.

### More

- [x] Added/updated tests

### Additional Notes

Fixes #212

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -83,6 +83,7 @@ require (
 	github.com/vulcand/predicate v1.1.0
 	go.elastic.co/apm v1.11.0
 	go.elastic.co/apm/module/apmot v1.11.0
+	golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97
 	golang.org/x/mod v0.4.2
 	golang.org/x/net v0.0.0-20210410081132-afb366fc7cd1
 	golang.org/x/time v0.0.0-20210220033141-f8bda1e9f3ba
--- a/go.sum
+++ b/go.sum
@@ -1266,8 +1266,9 @@ golang.org/x/crypto v0.0.0-20201002170205-7f63de1d35b0/go.mod h1:LzIPMQfyMNhhGPh
 golang.org/x/crypto v0.0.0-20201016220609-9e8e0b390897/go.mod h1:LzIPMQfyMNhhGPhUkYOs5KpL4U8rLKemX1yGLhDgUto=
 golang.org/x/crypto v0.0.0-20201221181555-eec23a3978ad/go.mod h1:jdWPYTVW3xRLrWPugEBEK3UY2ZEsg3UU495nc5E+M+I=
 golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83/go.mod h1:jdWPYTVW3xRLrWPugEBEK3UY2ZEsg3UU495nc5E+M+I=
-golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a h1:kr2P4QFmQr29mSLA43kwrOcgcReGTfbE9N577tCTuBc=
 golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a/go.mod h1:P+XmwS30IXTQdn5tA2iutPOUgjI07+tq3H3K9MVA1s8=
+golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97 h1:/UOmuWzQfxxo9UtlXMwuQU8CMgg1eZXqTRwkSQJWKOI=
+golang.org/x/crypto v0.0.0-20210711020723-a769d52b0f97/go.mod h1:GvvjBRRGRdwPK5ydBHafDWAxML/pGHZbMvKqRZ5+Abc=
 golang.org/x/exp v0.0.0-20190121172915-509febef88a4/go.mod h1:CJ0aWSM057203Lf6IL+f9T1iT9GByDxfZKAQTCR3kQA=
 golang.org/x/exp v0.0.0-20190306152737-a1d7652674e8/go.mod h1:CJ0aWSM057203Lf6IL+f9T1iT9GByDxfZKAQTCR3kQA=
 golang.org/x/exp v0.0.0-20190510132918-efd6b22b2522/go.mod h1:ZjyILWgesfNpC6sMxTJOJm9Kp84zZh5NQWvqDGG3Qr8=
@@ -1437,8 +1438,9 @@ golang.org/x/sys v0.0.0-20201231184435-2d18734c6014/go.mod h1:h1NjWce9XRLGQEsW7w
 golang.org/x/sys v0.0.0-20210119212857-b64e53b001e4/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20210225134936-a50acf3fe073/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/sys v0.0.0-20210303074136-134d130e1a04/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
-golang.org/x/sys v0.0.0-20210330210617-4fbd30eecc44 h1:Bli41pIlzTzf3KEY06n+xnzK/BESIg2ze4Pgfh/aI8c=
 golang.org/x/sys v0.0.0-20210330210617-4fbd30eecc44/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
+golang.org/x/sys v0.0.0-20210615035016-665e8c7367d1 h1:SrN+KX8Art/Sf4HNj6Zcz06G7VEz+7w9tdXTPOZ7+l4=
+golang.org/x/sys v0.0.0-20210615035016-665e8c7367d1/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/term v0.0.0-20201117132131-f5c789dd3221/go.mod h1:Nr5EML6q2oocZ2LXRh80K7BxOlk5/8JxuGnuhpl+muw=
 golang.org/x/term v0.0.0-20201126162022-7de9c90e9dd1/go.mod h1:bj7SfCRtBDWHUb9snDiAeCFNEtKQo2Wmx5Cou7ajbmo=
 golang.org/x/term v0.0.0-20210220032956-6a3ed077a48d h1:SZxvLBoTP5yHO3Frd4z4vrF+DBX9vMVanchswa69toE=
--- a/pkg/anonymize/testdata/anonymized-dynamic-config.json
+++ b/pkg/anonymize/testdata/anonymized-dynamic-config.json
@@ -348,7 +348,8 @@
         "certificates": [
           {
             "certFile": "xxxx",
-            "keyFile": "xxxx"
+            "keyFile": "xxxx",
+            "ocsp": {}
           }
         ],
         "maxIdleConnsPerHost": 42,
@@ -445,6 +446,7 @@
       {
         "certFile": "xxxx",
         "keyFile": "xxxx",
+        "ocsp": {},
         "stores": [
           "foo"
         ]
@@ -469,7 +471,8 @@
       "foo": {
         "defaultCertificate": {
           "certFile": "xxxx",
-          "keyFile": "xxxx"
+          "keyFile": "xxxx",
+          "ocsp": {}
         }
       }
     }
--- a/pkg/provider/acme/provider.go
+++ b/pkg/provider/acme/provider.go
@@ -628,6 +628,9 @@ func (p *Provider) refreshCertificates() {
 			Certificate: traefiktls.Certificate{
 				CertFile: traefiktls.FileOrContent(cert.Certificate.Certificate),
 				KeyFile:  traefiktls.FileOrContent(cert.Key),
+				OCSPConfig: traefiktls.OCSPConfig{
+					DisableStapling: true,
+				},
 			},
 			Stores: []string{cert.Store},
 		}
--- a/pkg/provider/consulcatalog/connect_tls.go
+++ b/pkg/provider/consulcatalog/connect_tls.go
@@ -28,6 +28,9 @@ func (c *connectCert) getLeaf() traefiktls.Certificate {
 	return traefiktls.Certificate{
 		CertFile: traefiktls.FileOrContent(c.leaf.cert),
 		KeyFile:  traefiktls.FileOrContent(c.leaf.key),
+		OCSPConfig: traefiktls.OCSPConfig{
+			DisableStapling: false,
+		},
 	}
 }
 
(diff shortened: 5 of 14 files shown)


Keridos:
Could any member please take a look at it?

rtribotte:
Hello @alekitto,

Just to let you know, we are working on OCSP support for Let's Encrypt certificate revocation.
While your PR is addressing OCSP stapling, and hence another feature, we want to address OCSP revocation first and come back later to OCSP stapling.

Nonetheless, thanks for your contribution!

tfny:
Hello everyone!  This concept is a candidate for a bounty during the Traefik Proxy 3.0 Hackaethon!  The goal for the version on the hackathon is just a little bit different, so I thought I would link you to the [blog post that describes it](https://traefik.io/blog#:~:text=Hack%20Into%20the,Read%20more) so that you can check that out and decide if you want to implement any changes. 

There will also be a more in-depth update on the issues associated with it coming soon.  

This means that if the design and POC or the fully fleshed-out PR for this issue is accepted in the Hackaethon, the person or team behind doing it will receive $250 as well as a chance at the Grand Prize of $1000! 

All accepted PRs and designs will receive a tshirt and stickers.  

To learn more about the Traefik Proxy 3.0 Hackaethon, check out our blog post that covers the goals, rules, scoring criteria, and dates in detail. 

To join us, [sign-up here](https://info.traefik.io/traefik-proxy-3-0-hackaethon?cta=pop)!

ddtmachado:
Hey @pkern , I'll try to elaborate a little bit more on the previous answer as it feels indeed very high level, but please @rtribotte  or any other maintainer really jump in to correct me if I say something stupid out of my knowledge on OCSP and OCSP stapling .

Supporting certificate revokation with OCSP means Traefik would get in touch with an OSCP server, usually referenced in the certificate itself if the CA supported it, and receive a response that states if the certificate is valid, revoked or unknown. Based on that it could behave in different ways depending on the design details, for example it could fail all the routers using that certificate right away or it could just remove the certificate from its store, leaving the routers serving the clients with the default certificate fallback, whatever is configured. 

That is the simplest approach and it would already bring a lot of questions on right behavior and implementation details. On the other hand OCSP stapling is an extension that basically says: "to avoid doing a call do the OCSP server every time you need to check a certificate what about just caching a certificate status and include it on the client TLS handshake response?", and while that makes complete sense when you think about client -> server relationship, Traefik is not a simple client, it sits between the real client and the server, so what would be the advantages in doing that? 
Should we just adhere to the protocol by forwarding the response to the client and leave it to validate that? This is a valid use case for apps that already handle OCSP stapling validation but what about the original goal of the extension, to save resources and reduce overhead? One could arguee that to achieve this in the context of a proxy it should synchronize and centralize its calls to the OCSP server, serving the stamped certificate status to its peers to enable the quick termination of compromised certificates in all its pairs while avoiding multiple OCSP calls. Those are just examples of the discussions we must have for the design and implementation.

I'm extending myself here but I felt it was necessary to clarify the kind of questions we're asking ourselves when looking at the design for these features, and why we wanted to tackle the simplest use case possible first with OSCP revokation. Finally OCSP stapling would still be a nice to have feature once we figure out how to fit it in Traefik, design wise, and this PR is a starting point at least.

I hope it helped :)

mholt:
Out of interest I've been watching this thread for a while now, and wanted to contribute from one perspective with my limited experience as the author of the Caddy web server and its OCSP implementation. Disclaimer: I am not intimately familiar with Traefik, but we hear about it a lot from our users who compare (apples to oranges, perhaps).

> Supporting certificate revokation with OCSP means Traefik would get in touch with an OSCP server, usually referenced in the certificate itself if the CA supported it, and receive a response that states if the certificate is valid, revoked or unknown. Based on that it could behave in different ways depending on the design details, for example it could fail all the routers using that certificate right away or it could just remove the certificate from its store, leaving the routers serving the clients with the default certificate fallback, whatever is configured.

This is something Caddy does, but we implemented stapling years before this.

> That is the simplest approach and it would already bring a lot of questions on right behavior and implementation details.

This is actually trickier than OCSP stapling. You can get some ideas though for it in this document, written by a collaboration of industry experts: https://github.com/https-dev/docs/blob/master/acme-ops.md

(I also have a chapter in my Expert Caddy content specializing in certificate automation logic, including OCSP/revocation: https://matt.life/expert-caddy/establishing-trust-pki-and-tls -- although this is normally just for Caddy sponsors, let me know if the Traefik team would like private access and I'll be happy to share.)

> On the other hand OCSP stapling is an extension that basically says: "to avoid doing a call do the OCSP server every time you need to check a certificate what about just caching a certificate status and include it on the client TLS handshake response?", and while that makes complete sense when you think about client -> server relationship, Traefik is not a simple client, it sits between the real client and the server, so what would be the advantages in doing that?

Forgive my relative ignorance, but isn't Traefik a proxy server? If Traefik is terminating TLS for clients -- whether the certificates are manually provided or automatically managed -- it should be stapling OCSP. If you're just proxying bytes across a transport layer (i.e. shuttling TLS without terminating it) then you don't have to do OCSP stapling.

> Should we just adhere to the protocol by forwarding the response to the client and leave it to validate that? This is a valid use case for apps that already handle OCSP stapling validation but what about the original goal of the extension, to save resources and reduce overhead? One could arguee that to achieve this in the context of a proxy it should synchronize and centralize its calls to the OCSP server, serving the stamped certificate status to its peers to enable the quick termination of compromised certificates in all its pairs while avoiding multiple OCSP calls. Those are just examples of the discussions we must have for the design and implementation.

If I understand you correctly, then yes: multiple instances of a TLS server should coordinate, cache and share OCSP staples so that the OCSP request load isn't multiplied times over.

Here's a classic writeup with some practical tips for server developers: https://gist.github.com/sleevi/5efe9ef98961ecfb4da8

(Caddy implements all those recommendations, and we collectively recommend other servers do the same.)

Anyway, IMO: stapling should be the first priority for OCSP support, then you can build in the cherry on top which is handling revoked status appropriately.

Hope that is helpful.

ddtmachado:
@mholt the questions I presented were given just as an example to make it clear there is more behind it than what it looks.

Thanks for the references, and for stating your opinion on what should be tackled first, I'm sure they'll be relevant for those interested in contributing code on the topic.

Having said that I would kindly ask to you to keep polite and focused on constructive feedback, mainly let's quit sarcasm as we're all grown up. I don't know much about Caddy myself, or have ever used it, but I can defintely say I hear a lot about it on social media and github as it seems you collectively enjoy a provocative marketing strategy and will not miss a chance to compare it where its beneficial. Fine seeing that on social media but this is definitely not the place to promote it.

Thanks for the comprehension!

tfny:
Hey Everyone, 

Thank you for your input, we are grateful to have it.  This has been a big topic of conversation with the maintainers for the last year and I would like to walk you through our thoughts.  

**The TL;DR** 
We will be doing stapling but we will be doing it after we have completed TLS revocation. The reason for this is that if we do not complete revocation, users will be receiving information that can become outdated; putting them at risk. We are going to continue the conversation on [this issue](https://github.com/traefik/traefik/issues/212). 

**The Details:**
Part of why it has taken us so much time to address this PR is that it is a draft and we do not review drafts. We only review PRs that indicate they are ready for review and instead focus on responding to direct questions that the writer(s)/community may have as they are developing the PR when they are in draft. That said, this PR has driven a lot of passionate internal discussion.

This discussion is the reason it has taken us so long to come back and see if the author was ready to review or to move forward on this PR on our own – we were trying to align on which parts of the OCSP standard makes sense for Traefik Proxy and our user base. We looked at the value and use case for OCSP stapling in Traefik Proxy and struggled because we felt that the cost outweighs the value for most users. We were particularly concerned about the impact on performance as we thought that it would require Traefik to parse the certificate and check it, stamping the status on the TLS handshake response for every connection. This would mean an added overhead for every served connection, while the benefit will certainly not apply to every client since the adoption of the standard is [still in progress](https://en.wikipedia.org/wiki/OCSP_stapling#Deployment), as it has been for a few years. 

However, we have realized that the performance impact is less significant than we initially thought.

That said, OCSP Stapling is not orthogonal to [OCSP support for certificate revocation](https://github.com/traefik/traefik/issues/9615). As explained in this new proposal we first want to make sure Traefik is only serving valid certificates by adding a background check to the OCSP server of the CA, and then we can move by stapling all responses with the certificate status information. 


If @[alekitto](https://github.com/alekitto) is interested in finishing any work they were waiting on while it is in draft form we would be very interested to see it, but as it is, we think we can take it over as co-authors and make some small tweaks before merging it.   

Thanks @alekitto for your hard work and everyone else here for the excellent points.

alekitto:
Hi everyone!
I have room to work on this in these weeks as well as discuss/help to code the revocation part.

To be clear: this PR is still in draft because its design has been not reviewed and thus tests and documentation have not been updated.

What is needed here (after a rebase to update its code to a more recent codebase) is a general review if the design or simply an evaluation on how the code has been organised/written, then I will write tests and update the docs and mark it as ready for a final review.

tfny:
So you would like a design review from the team to move forward?  We can totally do that.  Go ahead and rebase the code and the team will review the design.

pkern:
I think it'd be good if the stapling implementation would a) do OCSP refreshes detached from TLS handshakes - i.e. in a background Go routine with randomized exponential backoff and b) persist OCSP responses across restarts to avoid a startup dependency. I'm assuming that both will come with some corner-cases where you'd still want to block on startup for the initial OCSP query if you don't have an OCSP response on file for the certificate and MustStaple is set. The tradeoff here would be to hang upon connect to get a response on demand. There is currently no locking in the code, so receiving a ton of connection attempts would cause the same amount of pending OCSP requests.

PeterDraex:
@rtribotte do you still have plans to implement OCSP stapling yourself?

techsolo12:
Can anyone tell me the actual status? As i can see the traefik team should review the design.

alekitto:
Three years have passed since I've opened this PR and I think this should be moved forward and completed in some way.
I had some time to work on this, so I've updated the documentation, added some unit tests and finally marked this PR as ready to review.

I've separated OCSP handling (request, parsing, etc) from the certificate data, thus it could be reused for revocation in case it is needed, I've added a lock in order to issue only one request to the OCSP responder and I've also left a "todo" in the `Staple` method that should be replaced with the logic on what to do when OCSP status is "revoked".

I did not implement a storage to persist the responses: I think this should be done in a separate PR.

techsolo12:
Thank you for your PR and you hard at this. Hopefully @ldez find some time to review this one. :)

mholt:
FWIW, Let's Encrypt is [ending their OCSP service](https://letsencrypt.org/2024/07/23/replacing-ocsp-with-crls), so, unfortunately, OCSP stapling will be less deployed (which IMO will hurt privacy on the Web for clients who do use OCSP, and it will ultimately break Must-Staple -- I [opened a discussion on this](https://community.letsencrypt.org/t/what-will-happen-to-must-staple/222397)). Many popular, especially ACME CAs, such as Google and ZeroSSL/Sectigo, tend to follow Let's Encrypt's lead, so it's likely that OCSP will be declining at this point anyway.

alekitto:
> FWIW, Let's Encrypt is [ending their OCSP service](https://letsencrypt.org/2024/07/23/replacing-ocsp-with-crls), so, unfortunately, OCSP stapling will be less deployed (which IMO will hurt privacy on the Web for clients who do use OCSP, and it will ultimately break Must-Staple -- I [opened a discussion on this](https://community.letsencrypt.org/t/what-will-happen-to-must-staple/222397)). Many popular, especially ACME, CAs such as Google and ZeroSSL/Sectigo, tend to follow Let's Encrypt's lead, so it's likely that OCSP will be declining at this point anyway.

Too bad (don't want to go OT, but is very bad for privacy and I hope in a deeper discussion).
Anyway, an intent is not a final decision and the OCSP is not removed from the standard, so it can be useful anyway to provide this functionality.

rtribotte:
Hi @alekitto,

Sorry for delay.

We are finally ready to start reviewing the changes.
We were wondering if you would mind if we pushed review commits during the review (if necessary)?

alekitto:
Hi @rtribotte, no problem for me if you push review commits.

Should I do a rebase?

rtribotte:
> Should I do a rebase?

Sure, it is always better to keep the changes up to date with the target branch, thanks!

alekitto:
> Sure, it is always better to keep the changes up to date with the target branch, thanks!

Ok, rebased

rtribotte:
Hi @alekitto,

We have rebased the PR and added a review commit.
We changed the approach to have an opt-in support for OCSP stapling with an install configuration option.

As mentioned in the doc, the behavior is as follows:
```
When OCSP is enabled, Traefik checks the status of every certificate in the store,
including the default certificate, and staples the OCSP response to the TLS handshake.
The OCSP check is performed when the certificate is loaded,
and once every hour until it is successful at the halfway point before the update date.

Traefik caches the OCSP response as long as the associated certificate is provided by the configuration.
When a certificate is no longer provided,
the OCSP response has a 24 hour TTL waiting to be provided again or eventually removed.
The OCSP response is cached in memory and is not persisted between Traefik restarts.
```

Also, Traefik does not check the status of the certificate, it just passes through the information to the client.

WDYT?

pkg/tls/ocsp.go
```diff
@@ -0,0 +1,200 @@
...
+		issuer:     issuer,
+		responders: responders,
+	}, cache.NoExpiration)
+
+	return nil
+}
+
+// ResetTTL resets the expiration time for all items that has no expiration.
+// This allows to set a TTL to cache that do not exist in the dynamic configuration anymore.
+// For those existing, the TTL will be set to zero when the Upsert method will be called during
```
lbenguigui, line 115: I don't get this comment, `Upsert` will set a ttl to -1, not 0.

pkg/tls/ocsp.go
```diff
@@ -0,0 +1,200 @@
...
+// the UpdateConfigs method of the TLS manager.
+func (o *ocspStapler) ResetTTL() {
```
lbenguigui, line 117: I'm wondering if setting a TTL for unused certificates is really worth it.
We'll be making OCSP calls every hour or on each `UpdateConfigs` run for certs that aren't even in use, and it adds extra complexity to the code.

lbenguigui (changes requested):

pkg/tls/ocsp.go
```diff
@@ -0,0 +1,207 @@
...
+		if err := o.updateStaple(ctx, entry); err != nil {
+			log.Error().Err(err).Msgf("Unable to retieve OCSP staple for: %s", entry.leaf.Subject.CommonName)
+			continue
+		}
+	}
+}
+
+// obtainStaple obtains the OCSP stable for the given leaf certificate.
+func (o *ocspStapler) updateStaple(ctx context.Context, entry *ocspEntry) error {
+	// TODO: check FIPS compliance for SHA1 used as default hash, if set the hash options.
```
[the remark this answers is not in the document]
kevinpollet: After discussing it, there is no issue with FIPS.
We removed the TODO.

juliens (approved):
LGTM

traefiker:

traefiker:

logopk:
Hi, I have tried to get this to work on 3.5.0-rc2, but besides tcpdump network traffic to the responder I don't see any OCSP response.

Apparently there is no DBG output. Would that be a good addition to troubleshoot this? @alekitto

Thank you.

PR #8400, opened 2021-08-25
Add the ability to configure DataDog metrics using environment variables

mostafahussein:
### What does this PR do?

This PR will add the ability to use the standard DataDog's variables when defining the metrics host and port which will fallback to localhost if not defined.

### More

- [ ] Added/updated tests


kevinpollet:
Hello @mostafahussein and thanks for your contribution to Traefik,

Since `v2.5.0`, it's possible to define the metrics endpoint configuration with the `DD_AGENT_HOST` and `DD_DOGSTATSD_PORT` environment variables (introduced by https://github.com/traefik/traefik/pull/7968). After double-checking everything is working as expected.

According to this, we will decline and close this pull request.

PR #8404, opened 2021-08-26
added SSL-CLIENT-FINGERPRINT header

Powerhamster:
### What does this PR do?

This request adds the option `fingerprint` to `PassTLSClientCert` middleware. When set to `true` the option will provide the SHA1 hash of the certificate as header `SSL-CLIENT-FINGERPRINT`

### Motivation

Nginx provides the variable `$ssl_client_fingerprint` that can be passed as header or fastcgi parameter. This is a fast way to check if the provided certificate is matching an expected one, without passing the complete certificate.

### More

- [x] Added/updated tests


kevinpollet:
Hello @Powerhamster and thanks for your contribution,

Currently, when the [`PassTLSClientCert`](https://doc.traefik.io/traefik/middlewares/http/passtlsclientcert/#general) middleware is added to a router the `X-Forwarded-Tls-Client-Cert` header is added to the request and contains the escaped PEM of the client certificate. Therefore, it should be straightforward to compute the fingerprint of the forwarded client cert.

That being said, if we are not missing something, we are not sure to understand why it is faster to check if the provided certificate is matching an expected one, and the use case?

Why not let Traefik enforce the client cert validity with [mTLS](https://doc.traefik.io/traefik/https/tls/#client-authentication-mtls)?

 Can you give us more insights?

Powerhamster:
You are right. I just tried to avoid checking for another non standardized header inside the application code.

PR #8408, opened 2021-08-30
Fix alpine docker image to version 3.14

rtribotte:
### What does this PR do?

This PR pins the alpine docker image version to v3.14.

### Motivation

Having all released v3.14 bug fixes.

### More

- [ ] ~Added/updated tests~


mmatur (approved):

mmatur:
We will test this with the experimental images and once it is ok we will apply this change for the official image. A [PR](https://github.com/traefik/traefik-library-image/pull/11) is already created for this change

Issue #5999, opened 2019-12-11
Support RateLimit and InFlightReq for tcp proxy


zyf0330:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?
I find RateLimit and InFlightReq middlewares seem to support only http service. Hope to support tcp service.

dduportal:
Hi, thanks for your proposal! 

This feature request requires #5680 to be implemented first.

EndermanAPM:
Why is #5680 closed if is not implemented yet?

kanishkatn:
Could we please get this feature?

zyf0330:
Hope more people support this feature to implement.

grouvig:
For information, "[Feature Request] middlewares for tcp routers" #6684 replaces #5680.

PR #8428, opened 2021-09-07
feat: add rate limit middleware

tomMoulard:
### What does this PR do?

This PR introduces a new TCP Middleware called RateLimit.

This middleware does the same as the HTTP counterpart (i.e. `average`, `period` and `burst`), except that you can't have `sourceCriterion` since you have less metadata about a request in TCP than in HTTP.

### Motivation

Solve half of #5999

### More

- [X] Added/updated tests


mmoya:
This one was rejected while #8429 (same author, org member, other half of #5999) got merged. Is this just delayed/deprioritized?

tomMoulard:
Hello @mmoya,

This TCP rate limit middleware was a re-implementation of the HTTP rate limit middleware.
In TCP, Traefik handles connections, not requests, therefore the implementation was not good enough.

> Is this just delayed/deprioritized?

This middleware is not on the roadmap anymore.

PR #8431, opened 2021-09-09
Fix ServersTransport reference from IngressRoute service definition

rtribotte:
### What does this PR do?

This PR fixes the ServersTransport reference mechanism from an IngressRoute service definition.

It ensures that ServersTransport resources can't override each other when collecting them, by adding the namespace in the key.

It also introduces logic to handle cross-namespacing limitations and also adds the service namespace in the key, when building the service load balancer.

### Motivation

Make it impossible to reference a wrong ServersTransport.

### More

- [x] Added/updated tests

### Additional Notes


jbdoumenjou (approved):
LGTM

kevinpollet (changes requested):

kevinpollet (approved):

tomMoulard (approved):
LGTM :ok_hand:
Thank you :smiley:

PR #8432, opened 2021-09-09
Add traffic size metrics

OmarElawady:
### What does this PR do?

Adds `bytes_sent` and `bytes_received` metrics for all metric backends.

### Motivation

We make a gateway [solution](https://github.com/threefoldtech/zos/issues/1367) based on traefik, and this metric is needed to know how much traffic was used by each service. And it may be useful for other use cases as well, also it would be great to have your feedback whether this is the correct way to calculate it.

### More

- [x] Added/updated tests

### Additional Notes

- The relevant tests passes but the integration tests seems to fail at random unrelated tests each time.
- The pilot backend seems to be missing the router metrics. By accident?

### DIFF
--- a/docs/content/observability/metrics/overview.md
+++ b/docs/content/observability/metrics/overview.md
@@ -118,6 +118,8 @@ traefik_config_last_reload_failure
 |-----------------------------------------------------------|---------|----------|------------|--------|
 | [HTTP Requests Count](#http-requests-count)               | ✓       | ✓        | ✓          | ✓      |
 | [HTTPS Requests Count](#https-requests-count)             |         |          | ✓          |        |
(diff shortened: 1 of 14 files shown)


kmskpraveen:
Any update on this? This is a very crucial metrics that will be helpful especially with load testing.

rtribotte:
Hello @kmskpraveen,
 
Thanks for your interest in Traefik and thanks @OmarElawady for your contribution!

tomMoulard:
Hello @OmarElawady,

Thanks for your PR, I've worked a bit on this topic, and opened a new PR based on your work.

I'm closing this PR in favor of #9208.

PR #8433, opened 2021-09-10
middleware: add errors middleware host header config

### LINKED ISSUE #8267: Customize Host header for Errors middleware
### Welcome!

The Errors middleware currently does not allow to customize the Host header that is sent to the error service. This means that some service setups are currently not supported as an error page backend (if the server requires a specific Host header).

In our concrete case, we have Kubernetes with Traefik CRDs, and want to redirect error pages to an Kubernetes "ExternalName" service, which is a DNS redirect to Azure Storage (similar to Amazon S3). This does not work because the Errors middleware does not use the correct hostname in the Host header.

Example setup:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: external-errorpage
spec:
  type: ExternalName
  externalName: "customsubdomain.blob.core.windows.net"

---
kind: Middleware
metadata:
  name: errorpage
spec:
  errors:
    status:
      - "503"
    query: "/error.html"
    service:
      name: external-errorpage
      port: 443
```

Incidentally, a normal IngressRoute that points to the external service works correctly. Apparently Traefik rewrites the Host header to the new DNS name in that case. Example:
```yaml
kind: IngressRoute
metadata:
  name: exampleroute
spec:
  routes:
    - kind: Rule
      match: "PathPrefix(`/`)"
      priority: 1
      services:
        - kind: Service
          name: external-errorpage
          namespace: default
          passHostHeader: false
          port: 443
```

So I would propose to add a config option to Errors middleware that defines what to do with the Host header: don't touch, automatically determine (as for IngressRoute), or use a custom fixed value.

There was already a GitHub issue for this, but it was ignored: https://github.com/traefik/traefik/issues/7275

aaronraff:
### What does this PR do?

Adds a new configuration called `Host` to the errors middleware. The value set for this configuration is used as the host header when proxying requests to the error service.

If no config value is set, or if it's set to an empty string, then the original request's host header will be used.

### Motivation

fixes #8267 

### More

- [x] Added/updated tests

### Additional Notes

I'm curious to know if any other changes are necessary when updating the Kubernetes CRDs. It seems to me like the config is dynamically read in, so there's no need to bump the CRD version here, but I could be wrong!

### DIFF
--- a/docs/content/middlewares/http/errorpages.md
+++ b/docs/content/middlewares/http/errorpages.md
@@ -115,3 +115,7 @@ The service that will serve the new requested error page.
 ### `query`
 
 The URL for the error page (hosted by `service`). You can use the `{status}` variable in the `query` option in order to insert the status code in the URL.
+
+### `host`
+
+The Host header of the request made to the error service. By default the host of the original request is used.
--- a/pkg/anonymize/anonymize_config_test.go
+++ b/pkg/anonymize/anonymize_config_test.go
@@ -233,6 +233,7 @@ func TestDo_dynamicConfiguration(t *testing.T) {
 					Status:  []string{"foo"},
 					Service: "foo",
 					Query:   "foo",
+					Host:    "foo",
 				},
 				RateLimit: &dynamic.RateLimit{
 					Average: 42,
--- a/pkg/anonymize/testdata/anonymized-dynamic-config.json
+++ b/pkg/anonymize/testdata/anonymized-dynamic-config.json
@@ -186,7 +186,8 @@
             "foo"
           ],
           "service": "foo",
-          "query": "foo"
+          "query": "foo",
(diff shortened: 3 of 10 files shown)


kevinpollet:
Hello @aaronraff,

To fix the issue https://github.com/traefik/traefik/issues/8267 a new option is not necessary as the [`PassHostHeader`](https://doc.traefik.io/traefik/routing/services/#pass-host-header) option could be leveraged to forward the configured error service URL Host instead of `0.0.0.0`.

After looking deeper at the Errors middleware code, we also thought that it is better to forward the client request Host header to the configured error service instead of `0.0.0.0` and we opened a PR for that https://github.com/traefik/traefik/pull/8460.

As this PR introduces a new option and according to what we said before we will close it.

PR #8448, opened 2021-09-15
Ensure disableHTTP2 works with k8s crd

ssboisen:
### What does this PR do?

This PR ensures that disabling http2 works when using the kubernetes `ServersTransport` CRD. I have added a test verifying that it works now. Previously setting the `disableHTTP2` field in the CRD did not change that configuration option.

### Motivation

We host traefik as an ingress loadbalancer for a number of application servers running on k8s that do not support http2 - when trying to disable http2 we kept seeing errors on our application servers indicating that http2 had not been disabled.

### More

- [x] Added/updated tests

### Additional Notes

From my review of the code when implementing this change it looks like the `PeerCertURI` option in the CRD is also not being translated into a proper configuration of traefik. However I have not functionally verified this as we do not utilize this feature.


jbdoumenjou (approved):
LGTM

rtribotte (approved):
LGTM

kevinpollet (approved):

PR #8469, opened 2021-09-26
Add pattern log formatter and docs

JasonWangA:
### What does this PR do?

Enable custom logger formatter for access log.

### Motivation

In some situation.  We desire to define custom log fields to display the log content we want.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/observability/access-logs.md
+++ b/docs/content/observability/access-logs.md
@@ -44,6 +44,8 @@ accessLog:
 
 By default, logs are written using the Common Log Format (CLF).
 To write logs in JSON, use `json` in the `format` option.
+To write logs in Pattern, use `pattern` in the `format` option, then configure
+the custom log format in 'pattern' option.
 If the given format is unsupported, the default (CLF) is used instead.
 
 !!! info "Common Log Format"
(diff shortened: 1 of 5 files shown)


rtribotte:
Hello @JasonWangA,

Thanks for your contribution!

The PR is in `design-review` state, which means that we want to evaluate the design of the feature and if we would like to have such a feature.

As for now, we are wondering what are the exact limitations you are facing with the current formats and filter options. Can you elaborate on your use case?

As a side note, for enhancements, the target branch is master, so if the PR passes to `review` state, which would mean that we want to merge your PR at some point, it will be necessary to rebase it on the master branch.

JasonWangA:
> Hello @JasonWangA,
> 
> 
> The PR is in `design-review` state, which means that we want to evaluate the design of the feature and if we would like to have such a feature.
> 
> As for now, we are wondering what are the exact limitations you are facing with the current formats and filter options. Can you elaborate on your use case?
> 
> As a side note, for enhancements, the target branch is master, so if the PR passes to `review` state, which would mean that we want to merge your PR at some point, it will be necessary to rebase it on the master branch.

we want to define the log fields should display just like the way or style can be desing for nginx.

mpl:
> > As for now, we are wondering what are the exact limitations you are facing with the current formats and filter options. Can you elaborate on your use case?
> 
> we want to define the log fields should display just like the format or style can be set for nginx.

There's already third-party logging tools to do that. Especially when traefik runs within Kubernetes or other orchestrators. Why do you think we should add that functionality to traefik when it can already be done by other tools?

JasonWangA:
> > > As for now, we are wondering what are the exact limitations you are facing with the current formats and filter options. Can you elaborate on your use case?
> > 
> > 
> > we want to define the log fields should display just like the format or style can be set for nginx.
> 
> There's already third-party logging tools to do that. Especially when traefik runs within Kubernetes or other orchestrators. Why do you think we should add that functionality to traefik when it can already be done by other tools?

we want to see the log which the format designed flexibly by user from traefik but depends on other thirdparty tools.

kevinpollet:
Thanks for your feedback,

As for now, after discussing this feature, we think that the log format transformation should be handled by a third-party tool.

As we lack of a clear use case, for which we would need to add this responsibility to Traefik, we will close this pull request.

PR #8545, opened 2021-10-28
fix: use host's root CA set if ClientTLS ca is not defined

kevinpollet:
### What does this PR do?

This PR fixes the built `tls.Config` by the `ClientTLS.CreateTLSConfig` method to use the host's system root CA set when the `tls.ca` option is not defined. As described in the `tls.Config` documentation:

> RootCAs defines the set of root certificate authorities that clients use when verifying server certificates. If RootCAs is nil, TLS uses the host's root CA set. 

### Motivation

Being able to use the system root CA set in the `ForwardAuth` middleware and in the `Consul`, `Consul Catalog`, `HTTP`, `Marathon`, `Redis`, `ZooKeeper`, `Docker` and `Etcd` providers

### More

- [X] Added/updated tests

### Additional Notes

### DIFF
--- a/docs/content/middlewares/http/forwardauth.md
+++ b/docs/content/middlewares/http/forwardauth.md
@@ -354,7 +354,7 @@ The `tls` option is the TLS configuration from Traefik to the authentication ser
 #### `tls.ca`
 
 Certificate Authority used for the secured connection to the authentication server,
-defaults to the system bundle if not specified.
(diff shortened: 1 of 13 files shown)


tomMoulard (approved):
LGTM :ok_hand:

mpl (approved):

PR #8560, opened 2021-11-08
fix: increase UDP read buffer length to max datagram size

kevinpollet:
### What does this PR do?

This PR increase the buffer length used to read a UDP datagram received by Traefik and by the proxied server. When reading from a `SOCK_DGRAM` if the length of the buffer is lower than the datagram data size, the excess bytes are discarded.

As it is not possible to know the buffer length before reading the datagram data (without using `RAW` sockets), the buffer length is now initialized to the maximum datagram size (65535 bytes).

### Motivation

Fixes https://github.com/traefik/traefik/issues/8547

### More

- [X] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/udp/conn.go
+++ b/pkg/udp/conn.go
@@ -9,8 +9,6 @@ import (
 )
 
 // maxDatagramSize is the maximum size of a UDP datagram.
-// The UDP datagram length is coded on 2 bytes (16 bits),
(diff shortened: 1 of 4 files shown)


rtribotte (approved):
Thanks 👍

mpl (approved):

tomMoulard (approved):
LGTM :ok_hand:

PR #8575, opened 2021-11-12
docs: health check use readiness probe in k8s

Vampouille:
### What does this PR do?

Change link to the kubernetes documentation regarding Health check behaviour.

### Motivation

The kubernetes liveness probe status should not be used by load balancer. Traffic should be redirected to 'ready' Pods, not 'live' Pods.

From the k8s doc: 
`The kubelet uses readiness probes to know when a container is ready to start accepting traffic.`

https://github.com/kubernetes/website/blob/main/content/en/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes.md?plain=1#L17-L18

### DIFF
--- a/docs/content/routing/services/index.md
+++ b/docs/content/routing/services/index.md
@@ -340,7 +340,7 @@ Below are the available options for the health check mechanism:
 
     The Traefik health check is not available for `kubernetesCRD` and `kubernetesIngress` providers because Kubernetes
     already has a health check mechanism.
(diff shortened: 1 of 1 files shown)


rtribotte (approved):
Thanks 👍

kevinpollet (approved):
Thanks 👍

tomMoulard (approved):
LGTM :ok_hand:

PR #8577, opened 2021-11-15
Add support for tracing to ETW in Windows

dariopb:
### What does this PR do?

This commit adds support for tracing to go to ETW on windows systems, this is done via the hook functionality available in logrus and it is transparent to the current logging targets.
There is a new flag that enables the functionality under log, named **EnableETW**. When enabled, the logs can be retrieved subscribing to the provider id  **{63fae199-b614-503b-8fb6-c0be6dbe3fe5}**

For example:

```
...
log:
  level: DEBUG
  enableETW: true
...
```

One way to create a tracer by hand:

```
logman create trace traefik -p {63fae199-b614-503b-8fb6-c0be6dbe3fe5} -o c:\temp\traefik
logman start traefik
...
logman stop traefik
netsh trace convert c:\temp\traefik_000003.etl

```

### Motivation

When Traefik is run on Windows systems, most likely the infra log collection is setup to be using ETW. This is very common in Azure deployments on Windows for example. Having Traefik tracing directly to the native event tracers allows immediate collection of the logs on those deployments.

### More

- [ ] Added/updated tests

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -532,6 +532,11 @@ func configureLogging(staticConfiguration *static.Configuration) {
 			log.WithoutContext().Errorf("Error while opening log file %s: %v", logFile, err)
 		}
 	}
+
+	// configure the hook for native logging (etw in windows)
+	if staticConfiguration.Log != nil && staticConfiguration.Log.EnableETW {
+		log.InitNativeTracer()
+	}
 }
 
 func checkNewVersion() {
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/ExpediaDotCom/haystack-client-go v0.0.0-20190315171017-e7edbdf53a61
 	github.com/Masterminds/sprig/v3 v3.2.2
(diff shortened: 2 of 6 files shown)


mpl:
Hello @dariopb ,

Thank you for the contribution.

Ultimately our stance is that Traefik should not be in charge of accommodating all the various logging mechanisms, and it's up to the system englobing Traefik, such as an orchestrator, to take care of that piece of infrastructure.

As an analogy, if you look at how it's done in Docker, the application does not have to bother with that. Each container can be configured with a specific logging driver, and ETW is actually one of the supported drivers. See:
https://docs.docker.com/config/containers/logging/configure/#supported-logging-drivers
and
https://docs.docker.com/config/containers/logging/etwlogs/

For this reason, we're going to decline this PR.

dariopb:
"_Ultimately our stance is that Traefik should not be in charge of accommodating all the various logging mechanisms, and it's up to the system englobing Traefik, such as an orchestrator, to take care of that piece of infrastructure_.", this is totally fine when Traefik is run inside a container, but it doesn't help much when Traefik is running directly on the vm/bm, that is the case this PR was trying to fix. How do you suggest natively tracing in that case?

PR #8582, opened 2021-11-15
Add support HTTP/2 Server Push

janatjak:
### What does this PR do?

This MR add support for HTTP/2 Server Push

### Motivation

Access log and metrics wrap `http.ResponseWriter` and then it's not possible to use build-in Push feature in local middleware plugin (example https://github.com/janatjak/traefik-plugins/blob/master/http2push/http2push.go#L76)

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/middlewares/accesslog/capture_response_writer.go
+++ b/pkg/middlewares/accesslog/capture_response_writer.go
@@ -81,3 +81,11 @@ func (crw *captureResponseWriter) Status() int {
 func (crw *captureResponseWriter) Size() int64 {
 	return crw.size
 }
+
+// Push HTTP2-Push support
+func (crw *captureResponseWriter) Push(target string, opts *http.PushOptions) error {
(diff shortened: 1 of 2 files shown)


valy23:
Hi,
I would like to support the merge of this code to Traefik.
ServerPush technology delivers measurable performance gains and is part of the HTTP/2 standard. It is supported by all browsers.

jakubhonisek:
Great, thanks for fixing this issue. We also encountered this problem.

kbogi:
When is the release scheduled?

rtribotte:
Hello @janatjak,

As mentioned in the closed PR #7308:

> In order to be able to comply with HTTP2 push, the support of the http.Pusher interface should be done in all needed places.
...
We also have in mind that is probable that the Chromium team will (remove the support for HTTP2 push), and we are looking forward to seeing their conclusion.

Unfortunately, for those same reasons, we are still not convinced that we have to add this feature to Traefik.

I will close this PR accordingly.

PR #8596, opened 2021-11-26
fix: process all X-Forwarded-For headers in the request

### LINKED ISSUE #8586: Traefik only processes the first X-Forwarded-For entry in the request headers
### Welcome!

### What did you do?

A HTTP request can contain multiple times the same Header. For some Headers, only one value is expected (eg. `Content-Type`), for others, multiple values are OK (eg. `X-Forwarded-For`).

In the second case, multiple values can be set in a single "value" by comma-separating them: `X-Forwarded-For: 1.2.3.4, 5.6.7.8`.

However, according to documentation, you can set multiple values by repeating the headers, and you can combine. In this case, the order should be respected (when relevant):

```
X-Forwarded-For: 1.2.3.4, 5.6.7.8
X-Forwarded-For: 4.3.2.1
X-Forwarded-For: 10.11.12.13, 14.15.16.17
```

is equivalent to

```
X-Forwarded-For: 1.2.3.4, 5.6.7.8, 4.3.2.1, 10.11.12.13, 14.15.16.17
```

We use a network loadbalancing appliance which will append the client IP it sees in a new header-entry (and **not** append it to an existing entry)

### What did you see instead?

Traefik only takes the first occurrence of this header, by [using the `Get` function here](https://github.com/traefik/traefik/blob/master/pkg/ip/strategy.go#L37):

```golang
xff := req.Header.Get(xForwardedFor)
```

This will only return the first matching header, and thus in the above example, only `1.2.3.4, 5.6.7.8` will be parsed.

A quick solution, would be to use this instead:

```golang
xff := strings.Join(req.Header.Values(xForwardedFor), ",")
```

`2.4.14`

N/A

N/A

kevinpollet:
### What does this PR do?

This PR fixes the way we are processing the list of Forwarded IPs.

Per https://www.rfc-editor.org/rfc/rfc2616#section-4.2, the computed list must combine the Forwarded IPs received in all X-Forwarded-For header(s) values.

### Motivation

Fixes #8586

### More

- [X] Added/updated tests

### DIFF
--- a/pkg/middlewares/forwardedheaders/forwarded_header.go
+++ b/pkg/middlewares/forwardedheaders/forwarded_header.go
@@ -165,6 +165,12 @@ func (x *XForwarded) rewrite(outreq *http.Request) {
 		unsafeHeader(outreq.Header).Set(xForwardedHost, outreq.Host)
 	}
 
+	// Per https://www.rfc-editor.org/rfc/rfc2616#section-4.2, the Forwarded IPs list is in
+	// the same order as the values in the X-Forwarded-For header(s).
+	if xffs := unsafeHeader(outreq.Header).Values(xForwardedFor); len(xffs) > 0 {
+		unsafeHeader(outreq.Header).Set(xForwardedFor, strings.Join(xffs, ", "))
+	}
+
 	if x.hostname != "" {
 		unsafeHeader(outreq.Header).Set(xForwardedServer, x.hostname)
 	}
@@ -196,6 +202,10 @@ func (h unsafeHeader) Get(key string) string {
 		return ""
 	}
 	return h[key][0]
+}
+
+func (h unsafeHeader) Values(key string) []string {
+	return h[key]
 }
 
 func (h unsafeHeader) Del(key string) {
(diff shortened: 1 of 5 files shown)


pkg/middlewares/forwardedheaders/forwarded_header.go
```diff
@@ -165,6 +165,12 @@ func (x *XForwarded) rewrite(outreq *http.Request) {
 		unsafeHeader(outreq.Header).Set(xForwardedHost, outreq.Host)
 	}
 
+	// Per https://www.rfc-editor.org/rfc/rfc2616#section-4.2, the Forwarded IPs list is in
+	// the same order as the values in the X-Forwarded-For header(s).
+	if xffs := unsafeHeader(outreq.Header).Values(xForwardedFor); len(xffs) > 0 {
```
tomMoulard, line 170:
```suggestion
	if xffs, exists := unsafeHeader(outreq.Header[xForwardedFor]; exists {
```

rtribotte (approved):
Thanks 👍

tomMoulard (approved):
LGTM :ok_hand:

mpl (approved):

PR #8620, opened 2021-12-08
plugins: start the go routine before calling Provide

ldez:
### What does this PR do?

- fix: start the goroutine before calling Provide
- fix: catch Stdout Stderr

### Motivation

Prevent "deadlock" when sending the configuration during the `Provide` call.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/plugins/builder.go
+++ b/pkg/plugins/builder.go
@@ -6,6 +6,8 @@ import (
 	"net/http"
 	"os"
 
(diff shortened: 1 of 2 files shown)


jbdoumenjou (approved):
Thanks

tomMoulard (approved):
LGTM :ok_hand:

rtribotte (approved):
LGTM

Issue #3376, opened 2018-05-23
Do not attempt to renew certificate no longer used


Yajo:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

I booted one service in one server, behind Traefik, configured with docker backend and correctly labeled.

Everything worked fine:

- Traefik contacted LE
- Traefik obtained cert
- Traefik renewed cert after awhile, when needed
- Traefik :guitar: :heart: 

But, after some time, i moved the service to another equally-configured server.

### What did you expect to see?

Traefik in old server stop renewing cert if no container is using it actively.

### What did you see instead?

These logs:

```
INFO[2018-05-23T16:30:05Z] Renewing certificate from LE : {Main:example.com SANs:[]} 
legolog: 2018/05/23 16:30:05 [INFO][odoo.turbointernacional.com] acme: Trying renewal with -1447 hours remaining
legolog: 2018/05/23 16:30:05 [INFO][odoo.turbointernacional.com] acme: Obtaining bundled SAN certificate
legolog: 2018/05/23 16:30:05 [INFO][odoo.turbointernacional.com] AuthURL: https://acme-v01.api.letsencrypt.org/acme/authz/******
legolog: 2018/05/23 16:30:05 [INFO][odoo.turbointernacional.com] acme: Could not find solver for: dns-01
legolog: 2018/05/23 16:30:05 [INFO][odoo.turbointernacional.com] acme: Trying to solve HTTP-01
ERRO[2018-05-23T16:31:35Z] Error renewing certificate from LE: acme: Error 403 - urn:acme:error:unauthorized - Invalid response from http://example.com/.well-known/acme-challenge/****** [83.48.28.165]: 404
Error Detail:
        Validation for example.com:80
        Resolved to:
                *.*.*.*
        Used: *.*.*.*
```

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.5.3
Codename:     cancoillotte
Go version:   go1.9.4
Built:        2018-02-27_02:47:04PM
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

Docker backend, CLI flags configuration... but I don't think is relevant for this issue, where everything is working as expected.

minhdanh:
I'm seeing this behavior on Traefik 1.7-rc2 as well

kumy:
Is there any known workaround for this?

MaxWinterstein:
[...]

PR #8647, opened 2021-12-20
acme: don't store outdated certificates

xaurx:
### What does this PR do?

Traefik keeps and tries to update outdated certificates forever even if they are no longer in use or/and can't be renewed anymore. This PR removes outdated certificates from the acme store to avoid polluting it.

### Motivation

When one provisions white label app and manages domains on his behalf he inevitable removes some domains from time to time. But traefik keeps outdated certificates and moreover tries to update them with no reason.

### DIFF
--- a/pkg/provider/acme/provider.go
+++ b/pkg/provider/acme/provider.go
@@ -102,6 +102,7 @@ type Provider struct {
 	account                *Account
 	client                 *lego.Client
 	certsChan              chan *CertAndStore
+	certsDelChan           chan *CertAndStore
 	configurationChan      chan<- dynamic.Message
 	tlsManager             *traefiktls.Manager
 	clientMutex            sync.Mutex
@@ -527,6 +528,10 @@ func (p *Provider) addCertificateForDomain(domain types.Domain, certificate, key
 	p.certsChan <- &CertAndStore{Certificate: Certificate{Certificate: certificate, Key: key, Domain: domain}, Store: tlsStore}
 }
 
+func (p *Provider) delCertificateForDomain(domain types.Domain, tlsStore string) {
+	p.certsDelChan <- &CertAndStore{Certificate: Certificate{Domain: domain}, Store: tlsStore}
+}
+
 // getCertificateRenewDurations returns renew durations calculated from the given certificatesDuration in hours.
 // The first (RenewPeriod) is the period before the end of the certificate duration, during which the certificate should be renewed.
 // The second (RenewInterval) is the interval between renew attempts.
@@ -604,6 +609,7 @@ func deleteUnnecessaryDomains(ctx context.Context, domains []types.Domain) []typ
 
 func (p *Provider) watchCertificate(ctx context.Context) {
 	p.certsChan = make(chan *CertAndStore)
+	p.certsDelChan = make(chan *CertAndStore)
 
 	p.pool.GoCtx(func(ctxPool context.Context) {
 		for {
@@ -619,6 +625,18 @@ func (p *Provider) watchCertificate(ctx context.Context) {
 				}
 				if !certUpdated {
 					p.certificates = append(p.certificates, cert)
+				}
+
+				err := p.saveCertificates()
+				if err != nil {
+					log.FromContext(ctx).Error(err)
+				}
+			case cert := <-p.certsDelChan:
+				for i, domainsCertificate := range p.certificates {
+					if reflect.DeepEqual(cert.Domain, domainsCertificate.Certificate.Domain) {
+						p.certificates = append(p.certificates[:i], p.certificates[i+1:]...)
+						break
+					}
 				}
 
 				err := p.saveCertificates()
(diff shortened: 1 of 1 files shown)


psavva:
+1 for this, please can we get this merged into Traefik

Blackstareye:
Is there an update, regarding this issue ? 
I think it would help many sys admins, if the certs would be removed automatically :)

ldez:
This PR is very interesting but from the point of the design, we will try to create something with a larger scope (revocation, ...).
We will work on that soon, stay tuned.

Blackstareye:
Thanks, for the quick reply! :) 

Okay, so what is your recommended 'work around' for this problem, regarding this issue ?
something discussed in this issue  #3376  or something else? 

I have this issue right now  in the company I work, and it would be nice, having a solution, that includes the larger scope of this project, you mentioned.

flopana:
It has been 1 1/2 years. It would be really nice if Traefik would remove outdated certificates or not try to renew them. It's not remotely practical to manually remove unused certificates and restart Traefik every time a domain is no longer in use.

For example:
A client configures their own domain in my application. Traefik receives this configuration and obtains a certificate from LetsEncrypt.
This client moves on and stops using my service, changes the DNS records or lets the domain expire. This will cause unnecessary errors when Traefik tries to renew the certificate and will mask any real domain problems I might have by writing error logs every day.

msnisha:
Any idea on when this issue going to be resolved?

emilevauge:
Hey here, on behalf of the maintainer team, I would like to apologize for this radio silence. This is not great. This was put on hold because we wanted to provide a more extensive solution, but at the end of the day, we never found the time to finalize this. If we had to do it again, we would instead move quickly with this PR, as a first step.
We will review it in the next few weeks and move forward as fast as possible.

nmengin:
First and foremost, please apologize again for the long time with no action on this PR. As Emile explained, we had planned to improve the TLS certificate management globally, so we haven’t moved forward on this PR.

Such a modification hasn’t been brought yet and it’s now time to review it.

Let’s talk about the PR 🙂

I speak on my behalf, and even if the use case makes sense, I don’t think this PR is the best way to tackle this topic.

My main concern is the condition checked to delete the certificates from the renewal list: this PR proposes to delete the certificates [that are outdated for more than 7 days](https://github.com/traefik/traefik/pull/8647/files#diff-4f93ba21a09614847c01e2f916eba5dcc78e000cb9c076900dc033aa5d2bfc58R700).

IMO, using the expiry date is not the right approach: if you delete a certificate that is used by a router, the situation will be worse: instead of failing during the certificate renewal once a day, Traefik will try to create this certificate during each dynamic configuration reload.

In my PoV, it would make more sense to check if a certificate is used by at least one router before renewing it. Thus Traefik could delete the unnecessary certificates, and serving an outdated certificate would indicate an error during the challenge resolution.

For this reason, I propose the following actions:

- Closing this PR (it is pretty outdated and modifying it would be complicated now)
- Opening a proposal describing another approach based on the router's domain check.

Before moving forward on it, I’d like to have your feedback. I may miss something.

Many thanks for considering my request.

flopana:
> In my PoV, it would make more sense to check if a certificate is used by at least one router before renewing it. Thus Traefik could delete the unnecessary certificates, and serving an outdated certificate would indicate an error during the challenge resolution.

Hello @nmengin 

This was also my first intuition to solving the problem. Just don't renew certificates that are no longer used by a router.
But I had a discussion with @ldez about this via email.
The concern he raised as to why renewing unused certificates is necessary is as follows.

If you have thousands of certificates that are used on and off, the problem is that when you add routers that use these certificates back in, Traefik may need to renew more certificates than the limits of Let's Encrypt allow.

However, if Traefik renews these certificates gradually over a period of time, you can stay under the limits.
This means that removing unused certificates would break the functionality of Traefik for some customers and it would take weeks or even months to get the certificates back.

The solution that we both agreed would work for everyone is to introduce a time limit (or failed attempts, which is basically the same thing).

Let's Encrypt certificates are valid for 90 days.
After 60 days, Traefik will try to renew them once a day. If the renewal fails, Traefik will try again the next day.
If you could configure a limit to the number of days (or attempts) that traefik will try to renew before removing the certificates, we wouldn't break traefik for some customers and outdated certificates would actually be removed.

Quote from Idez:

> Example:
>
>- Expiration date: 2024/06/30
>- First attempt at renewal: 2024/05/31
>- Time limit: 2 days
>
>-> The certificate will be removed the 2024/06/02.

Idez please correct me if I'm wrong, I wrote this partly from memory and didn't read our whole conversation again.

pfaffman:
That's an interesting point that people might need certs for domains they aren't using.

The opposite problem that can occur is if you are renewing certs for a former customer who no longer points their DNS to your traefik server then renewing that certificate will fail, so you really want not to try to renew that cert that isn't routed.

flopana:
> The opposite problem that can occur is if you are renewing certs for a former customer who no longer points their DNS to your traefik server then renewing that certificate will fail, so you really want not to try to renew that cert that isn't routed.

Yes this is exactly the scenario I am facing in my company and was the reason why I spoke to Idez. The solution I mentioned earlier would still produce failed renewals but these would stop after n days. I personally would be ok with this.

The only solution I've come up with that does not break Traefik in some cases and still does not spam my logs with "failed" renewals is implementing a different storage option for certificates other than a json file.

If the certificates could be stored in for example [HashiCorp Vault](https://www.vaultproject.io/) I could just remove the certificate from the store programmatically after my customer has deleted their domain from my service.

I know I could also edit the json file programmatically and restart traefik, but I don't want to do that every time a customer decides to delete a domain on my service.

If anyone has any further input on this, I would love to hear your thoughts.

ldez:
This PR #10782 should resolve the problem.

nmengin:
Hey @xaurx,

According to the previous comments, I close this PR in favor of #10782.

PR #8660, opened 2021-12-25
Implement ChaCha cipher suite prioritization.

### LINKED ISSUE #8659: Prioritize ChaCha20 for devices without AES-NI support
### Welcome!

Many mobile devices (e. g. Android devices) experience performance loss when connecting to TLS servers that prefer AES due to them missing AES-NI (hardware accelerated AES). An attempt at optimizing for those would mean that a server has to configure the CPU-optimized ChaCha20 ciphers as a preference over AES which currently leads to ChaCha20 being preferred for *all* devices though. A third choice would be just leaning on the client preferences completely. None of these choices are optimal.

A client without AES-NI signals to the server that it prefers ChaCha20 ciphers by listing these ciphers at the top of its own preferred ciphers list. Traefik could then adjust its preferences to list ChaCha20 first as well if configured/wanted by the admin, leading to ChaCha20 being preferred for those clients only, and all other ciphers for other clients instead.

[OpenSSL has the `PrioritizeChaCha` option for this](https://www.openssl.org/docs/man1.1.1/man3/SSL_CONF_cmd_value_type.html), and as such this feature is already available in other HTTP servers using it such as [nginx](https://trac.nginx.org/nginx/ticket/1445).

Qualy's SSL Labs test checks for this prioritization and marks ChaCha20 ciphers with a <sup>P</sup> if prioritization works correctly.

icedream:
### What does this PR do?

This implements logic to allow encryption to be done using a ChaCha-based cipher suite if the client notifies the server that it prefers ChaCha-based cipher suites itself. This is usually the case if the client has no hardware acceleration for AES-based cryptography (mobile phones and similar embedded devices).

Fixes #8659.

### Motivation

It's a feature that surprisingly rarely has been requested but would absolutely work to optimize mobile performance of many HTTPS websites on Traefik.

Besides the usefulness, I honestly had this code around since Traefik 2.1 but didn't have time to properly test it, create an issue ticket and pull request for it yet, so I'm doing it now. Since I'm intending to hack on the code again for a bit anyways, I thought why not prepare this first.

### More

- [x] Added/updated tests

### Additional Notes

**Currently work in progress.**

TODO:

- This code seemingly doesn't affect ciphers in TLS 1.3 if I can trust Qualy's SSL Labs tests here. Needs to be fixed up.
- Should this logic be configurable (as in one can disable it if not wanted)?
- Documentation

The code respects the configured order of ciphers: If ChaCha20 is explicitly listed *at the top* of preferred cipher lists in the configuration, it will be enforced, if ChaCha20 is just listed *lower* in the configured cipher list it will only be preferred with devices that want it.

This code was running successfully on a company's production Traefik instance as a patch for a good while now.

### DIFF
--- a/pkg/tls/tlsmanager.go
+++ b/pkg/tls/tlsmanager.go
@@ -108,6 +108,12 @@ func (m *Manager) UpdateConfigs(ctx context.Context, stores map[string]Store, co
 	}
 }
 
+func isChaChaCipherSuite(cipherSuite uint16) bool {
+	return cipherSuite == tls.TLS_CHACHA20_POLY1305_SHA256 ||
+		cipherSuite == tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305 ||
+		cipherSuite == tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305
+}
+
 // Get gets the TLS configuration to use for a given store / configuration.
 func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 	m.lock.RLock()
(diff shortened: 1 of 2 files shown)


icedream:
Closing this issue as I overlooked a change in Go 1.16 that makes ChaCha20 preferences work by default.

For TLS 1.3 this behavior can not be changed.

PR #8674, opened 2022-01-03
Allow to use several TLS certificates for the same hostname

rbeuque74:
### What does this PR do?

Remove the limitation that restrict traefik to have only one certificate per hostname, and select the most compatible certificate for the client.

### Motivation

Prior to this commit, traefik forbid having two certificates using the
same hostname (SAN included). This was preventing having both a RSA and
EC certificates, improving performance for modern clients, while keeping
compatibility with older clients.

### More

- [x] Added/updated tests

### Additional Notes

First PR here, feel free to comment if I need to change stuff !
Relates to #3483.


rbeuque74:
Integration tests are failing, because the certificates emitted by pebble are not containing anything in SAN, and only contains the domain name as CN.

Cf error message:

> certificate is not valid for requested server name: x509: certificate relies on legacy Common Name field, use SANs instead

This might be a regression since I introduced `tls.ClientHelloInfo.SupportsCertificate` which calls `x509.Certificate.VerifyHostname`. Since https://github.com/golang/go/commit/02ce4118219dc51a14680a0c5fa24cf6e73deeed, golang no longer matches certificates using the CN, and is only checking with the extension SAN.
Don't know if that's something we still want to cover, as the majority of the certificate issuers includes SAN in their certificate.

I will check to make them pass tomorrow.

Alkorin (approved):

rbeuque74:
Hello guys,
Any thoughts?
Could be nice to have this feature introduced in v2.6.0 😃

Thanks
Romain

ddtmachado:
Hello @rbeuque74 , not here to comment on the PR itself but just to clarify that even if this is merged before 2.6 GA it won't be included in the release as once we publish the RC it means there will be no added features until GA, just bug fixes :)

rbeuque74:
Hello guys🙂,
Any thoughts on my pull-request?

Thanks,
Romain

rbeuque74:
Hello,
I just rebased the branch on master.

Romain

rbeuque74:
Hello all :wave:,
Do you have any visibility on when we could do the pull request design review ?
Two majors releases of traefik have passed, while the code in the PR is ready and working.
Any blockers here ?

Thanks,
Romain

mpl:
@rbeuque74 Hello,

Sorry, we still have no ETA to give you.
Basically, at first glance our concern is whether it is worth it, i.e. whether the use case is important enough VS the added maintenance cost, and runtime cost. Which is also why we haven't prioritized that PR and spent much time on it.

Regards,
Mathieu

rbeuque74:
Hello @mpl and @rtribotte,

Our background story behind this pull-request is that we are using Traefik as Ingress Controller on Kubernetes, and this ingress is used globally within our company, this include services hosted inside Kubernetes, but also legacy services outside of the Kubernetes cluster, that have not the same cryptographic capabilities than our microservices.
We would fancy to have the dual TLS certificates, this would allow us to:
- have a faster TLS handshake and lower the CPU impact of the TLS handshake for modern clients
- keep the compatibility with legacy components that doesn't handle ECDSA handshakes

As we have a high volume requests (~300 millions requests per day), this kind of optimization is critical for us.

I hope this can help you to consider that this usecase is important enough regarding the added maintenance cost (that seems negligible as this is just a modification of the existing code).

PS: PR is rebased over `origin/master`

Thanks again,
Romain

rbeuque74:
Thanks for change of status.
I will look at the failing tests tomorrow, probably something changed while the rebase

rtribotte:
Hello @rbeuque74,

> I will look at the failing tests tomorrow, probably something changed while the rebase

Thanks, but it is not necessary, we are going to handle test fixes if needed while we are reviewing.

Just to let you know, we are evaluating all the possible impacts, we are especially concerned by the impact during the handshake, in balance with the performance improvement for established connections.

mpl:
Hello @rbeuque74 ,

After careful consideration, the major problem we have with the PR is the (unavoidable) use of the clientHello.SupportsCertificate method. Used as is, it would be breaking for traefik users, since it would be stricter about which certificates it considers as matching wrt to the server name. That is, because it enforces https://go.dev/doc/go1.15#commonname , which so far traefik (as a server) did not. Which means it would change the behaviour of traefik wrt to the selection of appropriate certificates.

We considered some hacks to avoid that, e.g. setting the clientHello.ServerName to "" before calling the method, but it is not only gross and with potential side-effects, it would be pretty unmaintainable as well (as we have to control over the implementation of clientHello.SupportsCertificate).

However, we're not against the PR overall, and we think it's probably a good idea to eventually be stricter and enforce all that clientHello.SupportsCertificate does. So we're going to wait for the next major release, where we can be breaking, to revisit the subject, and this PR included.

ddtmachado:
Hey @rbeuque74 ,

Thank you so much for all the time and effort you put into this!

After a new round of discussions, we decided to not proceed with this feature. The value added was enough to justify the added complexity at this point and I must say this was a hard decision to take as there is a clear use case and we dig the technical aspect of it, but still it's a really niche use case that would only benefit a small part of our user base while potentially changing behaviour for everyone else.

PR #8686, opened 2022-01-10
Improve regexp matcher documentation

Hades32:
This PR should make the use of regular expressions in Traefik clearer. I wasted multiple hours (and found multiple issues and posts from others who had struggled the same) because I expected that the parameter of `HostRegexp(...)` was, well ... a regex... 🤷‍♂️ 

I hope the explicit statement, that values are interpreted verbatim and not as regex outside of curly braces will help others


rtribotte (approved):
Thanks👌

tomMoulard (approved):
LGTM :ok_hand:

kevinpollet (approved):
Thanks 👍

mpl (approved):

ngbrown:
If this is true: "Named regexps, of the form \`{name:regexp}\`, are the only expressions considered for regexp matching." Does that mean the example in the priorities section on the page are incorrect?

Examples like: ``HostRegexp(`.*\.traefik\.com`)`` and

kevinpollet:
Hello @ngbrown,

> Does that mean the example in the priorities section on the page are incorrect?

Right, thanks for pointing that out, we opened https://github.com/traefik/traefik/pull/8817 to fix the `HostRegexp` examples.

PR #8690, opened 2022-01-10
Don't ignore labels from unhealthy containers in docker provider

jvasseur:
### What does this PR do?

This PR will stops the docker provider from ignoring unhealthy containers but instead will still pull configuration from them but without adding them as servers to the generated services.

### Motivation

This will make Traefik return a 503 status code for unhealthy services instead of the current 404.

This should fix the docker provider of https://github.com/traefik/traefik/issues/1689

### More

- [x] Added/updated tests

### Additional Notes

I've made the PR for master because I wasn't sure if this should be considered an enhancements or a bugfix. I will be happy to rebase it on another branch if needed.

### DIFF
--- a/pkg/provider/docker/config.go
+++ b/pkg/provider/docker/config.go
@@ -89,6 +89,8 @@ func (p *Provider) buildConfiguration(ctx context.Context, containersInspected [
 }
 
 func (p *Provider) buildTCPServiceConfiguration(ctx context.Context, container dockerData, configuration *dynamic.TCPConfiguration) error {
+	logger := log.FromContext(ctx)
+
 	serviceName := getServiceName(container)
 
 	if len(configuration.Services) == 0 {
@@ -98,6 +100,11 @@ func (p *Provider) buildTCPServiceConfiguration(ctx context.Context, container d
 		configuration.Services[serviceName] = &dynamic.TCPService{
 			LoadBalancer: lb,
 		}
+	}
+
+	if container.Health != "" && container.Health != "healthy" {
+		logger.Debug("Filtering unhealthy or starting container")
+		return nil
 	}
 
 	for name, service := range configuration.Services {
@@ -112,6 +119,8 @@ func (p *Provider) buildTCPServiceConfiguration(ctx context.Context, container d
 }
 
 func (p *Provider) buildUDPServiceConfiguration(ctx context.Context, container dockerData, configuration *dynamic.UDPConfiguration) error {
+	logger := log.FromContext(ctx)
+
 	serviceName := getServiceName(container)
 
 	if len(configuration.Services) == 0 {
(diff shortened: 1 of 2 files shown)


jvasseur:
I have rebased the PR on the last version of master. Can I do something to help move this PR forward ?

kevinpollet:
Hello @jvasseur,
 
jvasseur:
Ok no problems, it's just that since the PR was opened quite some time ago I wanted to make sure it wasn't stuck waiting for something on my side but I understand that reviewing PRs takes time :slightly_smiling_face:

ddtmachado:
Hello @jvasseur , just to give you a quick update on this, it should definitely be an enhancement and not a bug since the behavior was intended and it works like that across different providers. Also it is a breaking change because it modifies existing behavior without backward compatibility.

Having said that I believe now is a good time to discuss that change since work for a new major is ramping up, although the team is still focused on other priorities right now so we still don't have an ETA to give.

kohenkatz:
> we are wondering about the specific use case this PR is addressing, can you please elaborate on this?

I don't know if this is the original intent, but for me the use-case is for things like GitLab, which can take several minutes to start up after doing an update. Currently, Traefik returns a 404 page while GitLab is booting, Ave several times I've had users contact me to complain that "GitLab has disappeared". These users know enough that if the error message was a 503 they would understand that it was temporarily down and would be back in a few minutes.

docs/content/providers/docker.md
```diff
@@ -714,3 +714,27 @@ providers:
...
 ```
+
+### `allowEmptyServices`
+
+_Optional, Default=false_
+
+If the parameter is set to `true`,
+it allows the creation of an empty [servers load balancer](../routing/services/index.md#servers-load-balancer)
+if the corresponding docker containers are not in a [healthy state](https://docs.docker.com/engine/reference/builder/#healthcheck).
+This results in `503` HTTP responses instead of `404` ones.
```
mpl, line 725:
```suggestion
If the parameter is set to `true`, any [servers load balancer](../routing/services/index.md#servers-load-balancer)
defined for docker containers stays alive and responsive when it becomes empty, i.e. when all its children containers
become [unhealthy](https://docs.docker.com/engine/reference/builder/#healthcheck).
This results in `503` HTTP responses instead of `404` ones.
```

mpl, line 725:
```suggestion
If the parameter is set to `true`, any [servers load
balancer](../routing/services/index.md#servers-load-balancer) defined for docker
containers is created regardless of the
[healthiness](https://docs.docker.com/engine/reference/builder/#healthcheck) of
the corresponding containers. It also then stays alive and responsive even at
times when it becomes empty, i.e. when all its children containers become
unhealthy.
This results in `503` HTTP responses instead of `404` ones, in the above cases.
```

tomMoulard (approved):
LGTM :ok_hand:

PR #8699, opened 2022-01-13
Redact credentials before logging

ibrahimalihc:
### What does this PR do?

During startup, Traefik logs the configuration for each configuration discovery provider. A provider's configuration can contain security sensitive values such as tokens, passwords, or keys which will appear in the emitted logs. To prevent this, this PR extends the mechanism already used to redact privacy sensitive values, so that it can also be used for security sensitive values. The mechanism is then applied before logging the config.

### Motivation

Fixes https://github.com/traefik/traefik/issues/8698

### More

- [x] Added/updated tests

### Additional Notes


mpl:
@ibrahimalihc Hello. We realized the issue required more work than just changing the log levels, so we starting working on another approach. Which is why we pushed new commits to the PR.

tomMoulard (approved):
LGTM :ok_hand:

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
LGTM

kaznovac:
this feature should be optional:
the config files are logged only in `debug` log level (as i'm aware of), and then it's expected to see the resolved variables for debugging purposes

ibrahimalihc:
Thanks for getting this in!

Issue #8044, opened 2021-04-09
Possible memory leak, regexp instructions consuming a lot of memory


kungfukennyg:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?
We run a single Traefik node in a Docker container (started via Nomad) on an AWS t3.medium EC2 instance hooked up to a `Consul` network, and our workflow involves creating many thousands of Nomad jobs that are made accessible to the wider internet via Traefik. Traefik works fine for this, but exhibits leaky behavior as it stays up longer, until it eventually kills the EC2 instance or gets restarted by Nomad.
![Screen Shot 2021-04-06 at 15 49 41](https://user-images.githubusercontent.com/6355483/114197918-d535ad00-9920-11eb-962b-4defc674b868.png)

We initially noticed this on Traefik v2.2.5, and upgraded to v.2.3.7 to see if that fixed the issue, but sadly it did not. 

### What did you expect to see?
Memory falling after Nomad jobs are completed and the applications are no longer accessible.

### What did you see instead?
Memory growing unbounded. 

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
/ # traefik version
Version:      2.3.7
Codename:     picodon
Go version:   go1.15.6
Built:        2021-01-11T18:03:02Z
OS/Arch:      linux/amd64
```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

Environment variables:
```console
# env | grep TRAEFIK
TRAEFIK_METRICS_STATSD_ADDRESS=127.0.0.1:8125
TRAEFIK_PROVIDERS_CONSULCATALOG_EXPOSEDBYDEFAULT=false
TRAEFIK_LOG_LEVEL=INFO
TRAEFIK_ACCESSLOG=true
TRAEFIK_LOG_FORMAT=json
TRAEFIK_ACCESSLOG_FORMAT=json
TRAEFIK_ENTRYPOINTS_tcp_ADDRESS=:5672
TRAEFIK_METRICS_STATSD=true
TRAEFIK_PROVIDERS_CONSULCATALOG_ENDPOINT_SCHEME=http
TRAEFIK_ENTRYPOINTS_http_ADDRESS=:8080
TRAEFIK_PROVIDERS_CONSULCATALOG_ENDPOINT_ADDRESS=127.0.0.1:8500
TRAEFIK_API_DEBUG=true
TRAEFIK_API_INSECURE=true
TRAEFIK_API_DASHBOARD=true
[...]

Issue #7533, opened 2020-11-09
Memory leaks lead to OOM on VM


alex-roman:
### Do you want to request a *feature* or report a *bug*?
Bug

### What did you do?
Thanks for the awesome app, folks! Here's a description of the issue we're grappling with in our prod env.
Two m5.large ubuntu 20 VMs have traefik 2.3.2 up and running. Both instances experience intermittent OOMs. Traefik is the only prod service running on the boxes.

### What did you expect to see?
Traefik consuming a reasonable amount of memory and not being OOM killed

### What did you see instead?
![image](https://user-images.githubusercontent.com/5751312/98546652-083a1e80-2265-11eb-9edd-12369effb929.png)

How often?
```
# dmesg | grep "Out of memory"
[28224.365945] Out of memory: Killed process 1280 (traefik) total-vm:12695484kB, anon-rss:7520220kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:15796kB oom_score_adj:0
[28375.784145] Out of memory: Killed process 5460 (traefik) total-vm:15401052kB, anon-rss:7528288kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:15060kB oom_score_adj:0
[28399.713629] Out of memory: Killed process 5488 (traefik) total-vm:22978332kB, anon-rss:7528016kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:15156kB oom_score_adj:0
[94915.294507] Out of memory: Killed process 5511 (traefik) total-vm:15942780kB, anon-rss:7517724kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:15852kB oom_score_adj:0
[94931.163620] Out of memory: Killed process 18783 (traefik) total-vm:16483484kB, anon-rss:7525320kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:15072kB oom_score_adj:0
[95110.762503] Out of memory: Killed process 18816 (traefik) total-vm:37592188kB, anon-rss:7520028kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:15280kB oom_score_adj:0
[96053.725702] Out of memory: Killed process 18860 (traefik) total-vm:19189820kB, anon-rss:7514768kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:15828kB oom_score_adj:0
[...]

Issue #7191, opened 2020-08-18
traefik memory usage keeps growing


paylm:
### Do you want to request a *feature* or report a *bug*?

Bug

### What did you do?

```

we are notificing RSS memory of traefik process running within Kubernetes pod container keeps growing, eventually reaching the configured memory limits and getting OOMKilled .

I tried to close all log output, but the memory usage is still keeps growing without release after a few days .


```
![Image_20200818145555](https://user-images.githubusercontent.com/16893729/90480799-5f978980-e163-11ea-9b21-c34838f258dd.png)
![Image_20200818145603](https://user-images.githubusercontent.com/16893729/90480814-62927a00-e163-11ea-8a2f-4b7b6111622b.png)

![Image_20200818145613](https://user-images.githubusercontent.com/16893729/90480848-70e09600-e163-11ea-8958-b64617ccf079.png)
![Image_20200818145622](https://user-images.githubusercontent.com/16893729/90480849-7211c300-e163-11ea-8274-1576d1971207.png)

### What did you expect to see?

No unlimited memory growth/leak . 
When the business is low, memory can be recycled .

### What did you see instead?

```memory leak```

### Output of `traefik version`: (_What version of Traefik are you using?_)

```

traefik:v1.7.12

```

### What is your environment & configuration (arguments, toml, provider, platform, ...)?

```toml
--api
--kubernetes
--logLevel=FATAL
```

```
    insecureSkipVerify = true
    defaultEntryPoints = ["http", "https"]
    [entryPoints]
      [entryPoints.http]
        address = ":80"
        #[entryPoints.http.redirect]
        #  entryPoint = "https"
        #[entryPoints.http.forwardedHeaders]
        #  trustedIPs = ["10.1.0.0/16", "172.20.0.0/16", "192.168.1.3"]
      [entryPoints.https]
        address = ":443"
        [entryPoints.https.tls]
          [[entryPoints.https.tls.certificates]]
            CertFile = "/ssl/tls.crt"
            KeyFile = "/ssl/tls.key"
    [accessLog]
      #accesslog -> stdout
      #filePath = "/logs/traefik.access.log"
      #bufferingSize = 500
    [metrics]
      [metrics.prometheus]
        entryPoint = "traefik"
        buckets = [0.1,0.3,1.2,5.0]
```

k8s limit config
```
args:
        - '--configfile=/config/traefik.toml'
        - '--api'
        - '--kubernetes'
        - '--logLevel=FATAL'
      ports:
[...]

PR #8706, opened 2022-01-17
Mitigate memory leak

mpl:
### What does this PR do?

This PR bumps the mux dependency version, because it contains a work-around to mitigate a memory leak.

### Motivation

To fix a memory leak, which might be the one observed at #7964 for example.

### More

~- [ ] Added/updated tests~

### Additional Notes

#8044, #7533

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -94,7 +94,7 @@ require (
 replace (
 	github.com/abbot/go-http-auth => github.com/containous/go-http-auth v0.4.1-0.20200324110947-a37a7636d23e
 	github.com/go-check/check => github.com/containous/check v0.0.0-20170915194414-ca0bf163426a
-	github.com/gorilla/mux => github.com/containous/mux v0.0.0-20181024131434-c33f32e26898
+	github.com/gorilla/mux => github.com/containous/mux v0.0.0-20220113180107-8ffa4f6d063c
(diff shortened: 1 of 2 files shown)


rtribotte (approved):
LGTM 👌

kevinpollet (approved):
Thanks 👍

tomMoulard (approved):

darkl0rd:
Is this also affecting Traefik 1.7? And if so, will this be backported? (#7191)

llussy:
Hi, is there a fix for version 1.7? I didn't find a newer version.

tfny:
We are looking into it, we will get back to you soon. :)

ddtmachado:
@darkl0rd @llussy Just to clarify version 1.7 is not supported anymore and we are preparing a new doc section that would make it easy to find what versions we are actively maintaining and what are the rules for that, including exceptions if any.

The PR is still under review, and feedback is welcomed as always.

Having said that, yes this problem affects 1.7 and we might look at backporting it at some point if really needed. Really needed in this context means a customer is affected and couldn't finish the migration to 2.x in time or hit some blocker in doing so, then we might consider a priority to avoid production issues until the migration is finished.

Of course if someone in the community is willing to work on it and backport the fix into 1.7 I see no reason why we would not review and merge it as soon as possible.

Issue #5107, opened 2019-07-16
Frequent Skipping Configuration


timeandtimeago:
Hi, 
I was wondering if someone could tell me what it means when traefik outputs a line like this:
```level=info msg="Skipping same configuration for provider docker"```

When nothing is happening on my server - it will sit there outputting this same line over and over.

I just want to make sure I don't have something going on. If this is expected - apologies for the unnecessary issue.

fopina:
Second this. Search this repository for the string does not show anything...

dduportal:
Hi @timeandtimeago, this message is normal: when the Docker API emits an event for any change,
Traefik is notified and react by retrieving the configuration from the API, and compare it to its last know version.

If nothing changed from last known version (meaning no item related to Traefik has changed), then this message is printed.

fopina:
Any chance this would make sense as debug message instead of info?
For log sanity :)

traefiker:
Hi! I'm Træfiker :robot: the bot in charge of communication regulation.

Thanks for your interest in Traefik!

Issue templates help us help you by providing all necessary information.

Please edit your issue and use the available templates:

- [Feature request](https://raw.githubusercontent.com/containous/traefik/master/.github/ISSUE_TEMPLATE/Feature_request.md)
- [Bug report](https://raw.githubusercontent.com/containous/traefik/master/.github/ISSUE_TEMPLATE/Bug_report.md)

And remember: each time someone ignores the template, a cute little bunny dies.
<details>
<summary>#SaveTheCuteBunny :heart: :rabbit: :heart:</summary>

![southpark - estrella kill rabbit](https://user-images.githubusercontent.com/5674651/27143914-3cea7896-5130-11e7-81f8-02d8ac2da7a7.gif)

![southpark - estrella kill rabbit2](https://user-images.githubusercontent.com/5674651/27143924-49104c72-5130-11e7-82ae-ccedbd5dfa1c.gif)

</details>

CybotTM:
seems already be declined: https://github.com/containous/traefik/pull/5718

dduportal:
Thanks @CybotTM for pointing this. We are closing this issue as the answer to the initial question is "yes it is expected behavior". More details on the PR you pointed.

Follow up on #5693.

PR #8709, opened 2022-01-18
refactor: debug log level

rhtenhove:
### What does this PR do?

It changes the log level of the message `"Skipping same configuration"` from `info` to `debug`

### Motivation

I wanted to see what was happening with a dnsChallenge request, which had quite a long timeout. All the while the logs were filled with a `"Skipping same configuration"` message every 15 seconds. At some point my display is full with messages telling me nothing is happening and I have no idea if I might have missed a relevant message.

While researching this I found #5107, #5693 and #5718.

The decision in #5718 seemed to be based on another message, `"Skipping empty configuration for provider"`. Showing this message makes sense as this message indicates no configuration is found for a provider, and that is usually not the case and indicative of problems (or a redundant provider).

The other message though, `"Skipping same configuration"` is normal, meaning a configuration was found, and it's kept as is. It makes no sense to tell me as a user that "Things are operational, and I will not change anything.". The case where this info would be useful is the case where a change is expected, but not seen in operation.
Even then as a user I would expect a message "I've changed something!" each time a change happens, and the problem becoming apparent of that message not displaying.

Either way, this change will make the logs much more readable and useful, and hopefully the argument above will make for a different decision than in 2019.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2016-2020 Containous SAS; 2020-2021 Traefik Labs
+Copyright (c) 2016-2020 Containous SAS; 2020-2022 Traefik Labs
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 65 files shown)


tomMoulard:
Closing this because @rhtenhove used his master branch, so maintainer cannot rebase branch

mpl:
Sorry @rhtenhove , we messed something up while trying to rebase your branch, so we had to create a new PR instead. You'll be a co-author on the new one of course.

PR #8715, opened 2022-01-19
Fix HTTP provider endpoint config example 

kevinpollet:
### What does this PR do?

This pull request fixes the endpoint config example for the HTTP provider. The config example is working, but using an array is misleading as one could think that the HTTP provider supports multiple endpoints (see https://community.traefik.io/t/multiple-http-providers/13102).

### Motivation

Fix the misleading HTTP provider endpoint config example.

### More

- [ ] Added/updated tests


ldez (approved):
LGTM

PR #8742, opened 2022-01-27
Add simpler and faster debug Makefile target

mpl:
### What does this PR do?

This PR adds new targets in order to build faster (especially on Mac) docker images, for easier iteration while debugging on e.g. Kubernetes. The generated image is also alpine-based, with tools, instead of being from scratch.
It also cleans up a few things to ease maintainability/readability.

### Motivation

To ease the debugging workflow on Kubernetes.

### More

~- [ ] Added/updated tests~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,6 @@ SHA := $(shell git rev-parse HEAD)
 VERSION_GIT := $(if $(TAG_NAME),$(TAG_NAME),$(SHA))
 VERSION := $(if $(VERSION),$(VERSION),$(VERSION_GIT))
 
-BIND_DIR := dist
-
 GIT_BRANCH := $(subst heads/,,$(shell git rev-parse --abbrev-ref HEAD 2>/dev/null))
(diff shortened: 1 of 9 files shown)


mpl:
Blocking it for now because I suspect it broke the "optimization" for running the integration tests faster.

rtribotte (approved):
LGTM 👌

Makefile
```diff
@@ -125,6 +127,10 @@ build-image: clean-webui binary
 build-image-dirty: binary
 	docker build -t $(TRAEFIK_IMAGE) .
 
+## Locally build traefik for linux, then shove it an ubuntu image, with basic tools.
```
rtribotte, line 130:
```suggestion
## Locally build traefik for linux, then shove it an alpine image, with basic tools.
```

mpl:
blocking the merging for now as we need to figure out the apk upgrade question (and be consistent in all Dockerfiles)

tomMoulard (approved):

PR #8747, opened 2022-01-31
Add simpler and faster debug Makefile target - Semaphore

mmatur:
### What does this PR do?

This PR adds new targets in order to build faster (especially on Mac) docker images, for easier iteration while debugging on e.g. Kubernetes. The generated image is also ubuntu-based, with tools, instead of being from scratch.
It also cleans up a few things to ease maintainability/readability.

### Motivation

To ease the debugging workflow on Kubernetes.

### More

~- [ ] Added/updated tests~

### Additional Notes

Test PR to debug Semaphore trigger workflow

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,7 +8,7 @@ on:
 env:
   GO_VERSION: 1.17
   CGO_ENABLED: 0
-  PRE_TARGET: ""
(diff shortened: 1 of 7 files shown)


mmatur:
SempahoreCI has deployed a fix so we can close this test PR

PR #8182, opened 2021-06-11
Add muxer for TCP Routers

dtomcej:
### What does this PR do?

This PR adds muxxing to TCP routers, allowing for multiple and extensible matchers to be used to match and find routes.
Thanks to the above, this PR then also adds ClientIP matcher for TCP. Which means both `ClientIP` and `HostSNI` matchers are now available for TCP.

### Motivation

TCP Routing can be improved.

### More

- [x] Added/updated tests
- [x] Added/updated documentation

### Additional Notes

- This PR adds the muxing into `github.com/traefik/traefik/v2/pkg/tcp.Router`. I'm not 100% sure that is the correct place to do this.
- This PR adds clientIP routing into router builder

PR #8748, opened 2022-01-31
Add HostSNI wildcards

Links2004:
### What does this PR do?

this change allow to use wildcards with HostSNI and tcp routers and tls passthrough.

### Motivation

updating the traefik config for each new subdomin a service needs is complicated and unnecessary ;) 

### Additional Notes

based on the patch from here:
https://github.com/traefik/traefik/issues/7266#issuecomment-989204620


tomMoulard:
Hello @Links2004,

We are currently working on #8182 which implements a muxer for TCP. It would allow introducing a new `HostSNIRegexp` matcher. This new matcher will support more complex use cases.

Therefore, we will decline your PR in favor of #8182.

PR #8751, opened 2022-02-02
add support to reverse-proxy postgres connections

maigl:
PostgreSQL performs a StartTLS handshake before sending any TLS header.
Since traefik doesn't perform this handshake no TLS header is sent and
so traefik cannot extract the target server name (SNI). For that
reason SNI-routing is not possible for postgres connections.

This PR adds support to perform the StartTLS handshake (if configured)
in the following steps:
- traefik simulates a PostgreSQl server and performs StartTLS handshake with client
- client now sends TLS header
- traefik extracts SNI
- traefik performs StartTLS handshake with server
- traefik routes connection from client to target PostgreSQL instance

### Motivation

I'm part of a team that provides managed databases inside our company. 
We use traefik enhanced with this feature for 

### More

- [ ] Added/updated tests


maigl:
#8751

maigl:
.. I just want to point out that my PR most certainly does not fix https://github.com/traefik/traefik/issues/7366. 
This was added by somebody.

There are many different StartTLS flavors and in this PR I focused on **postgres StartTLS**. But the changes in the PR could be a good starting point to also implement SMTP StartTLS.

tfny:
Thanks for your PR!  We popped it into the design review process, we have scheduled it for review the week of Feb 14th.  

There is another PR [8182](https://github.com/traefik/traefik/pull/8182) that touches the same topic that we wish to integrate before so that we can make sure they play nicely together.

tfny:
Just letting you know that there has been a delay with [PR 8182](https://github.com/traefik/traefik/pull/8182).  We still need that merged before we can review this one.  Once it is merged one of the reviewers on PR 8182 (@mpl, @tomMoulard, @rtribotte @juliens) will be reviewing this PR.

maigl:
Currently there is a merge conflict. Due to the router proxy changes this is not a trial merge. I will merge it in the next couple of days .. But it would also be awesome to get feedback to this feature even though it's not fully merged yet. 
(The overall feature won't change much)

Is this feature something you would consider to merge? So if you have time to review it, I would really appreciate any ideas on this.

rtribotte:
Hello @maigl,

Thanks for your interest in Traefik and for your contribution,

I already handled the rebase locally to take a look at the code in the context of the refactoring implied by the introduction of the TCP muxer.

The PR is labeled as `needs-design-review`, as you can see in our [contribution guide](https://github.com/traefik/contributors-guide/blob/master/review_guidelines.md), this means that we have to take a longer look to evaluate how it would interact with the other parts of Traefik, as well as the opportunity of introducing this new feature.

As of now, I can say that we understand the changes, and I tested it successfully to proxy TLS connections (well, TCP STARTTLS connections) to PostgreSQL.
We are currently concerned by the interaction with non-STARTTLS routing, as well as the impact on performances.
Moreover, we are also hesitating on adding/supporting a specific implementation for each STARTTLS flavor. Plus, we identified that unfortunately server-first protocols like SMTP, are not going to work, and hence it would not be sufficient to add a new hook. This is a bit of a downside to that PR since most traction we have to support STARTTLS is for SMTP if I'm not mistaken.

As a side concern, ideally, we would have liked that supporting STARTTLS doesn't add more complexity to the actual TCP routing. We have no alternative yet, but we want to take the time to think about it.

In short, thank you for the work done here, and we are going to handle the rebase for you if we decide to validate the design (but I can definitely push the rebased branch right away if you prefer).

maigl:
@rtribotte Thanks for your feedback, I understand all your points and concerns very well .. we also thought about the downsides (like SMTP problem) and we have no good solution either .. 
It would be awesome if you could handle the rebase.
Also, I'm happy to take part in further discussions, maybe we find a better way.

rtribotte:
Hello @maigl,

I tried to push to your branch the rebased changes, unfortunately, this is not allowed for branches belonging to organization forks.
Can you re-open a PR from a branch on your personal fork?

maigl:
> Hello @maigl,
> 
> I tried to push to your branch the rebased changes, unfortunately, this is not allowed for branches belonging to organization forks. Can you re-open a PR from a branch on your personal fork?

.. sure I can do this .. but it would be a new PR and we would loose the current thread right ? Alternatively, you can also send me a patch and I push it ? What do you prefer ?

PR #8766, opened 2022-02-10
Replace ACME provider default RSA4096 with EC256

lnsp:
### What does this PR do?

Replace the default ACME provider RSA4096 key configuration with EC256.

### Motivation

TLS handshakes with RSA4096 keys are pretty slow in Go. [I found out about it yesterday](https://blog.espe.tech/posts/letsencrypt-may-slow-you-down/) while digging through performance issues in my own reverse-proxy project, but thought that traefik could profit from these changes as well.  Go's TLS implementation for the P-256 curve has been [hand optimized for performance](https://groups.google.com/g/golang-nuts/c/IOIPoyWvN2k) and basically all systems should support ECDSA today anyway.

### DIFF
--- a/pkg/provider/acme/provider.go
+++ b/pkg/provider/acme/provider.go
@@ -50,7 +50,7 @@ type Configuration struct {
 func (a *Configuration) SetDefaults() {
 	a.CAServer = lego.LEDirectoryProduction
 	a.Storage = "acme.json"
-	a.KeyType = "RSA4096"
+	a.KeyType = "EC256"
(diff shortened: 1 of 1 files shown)


kevinpollet:
Hello @lnsp,

Thanks for your contribution.

Changing the default value for the `KeyType` option from `RSA4096` to `EC256` is a breaking change and this cannot be done in a minor or patch release. That being said, feel free to open a proposal on this subject which might be considered in a new major release.

According to that, we will decline this pull request.

PR #8783, opened 2022-02-17
Replace npm with yarn to install/run the webui

tomMoulard:
### What does this PR do?

This PR changes the way that Web UI dependencies are installed through docker.
It now uses [`yarn install`](https://www.npmjs.com/package/yarn) instead of `npm install`.

This PR also replaces the usage of `npm run` by `yarn` to avoid using two utilities when only one is required.

### Motivation

This PR helps to build the Web UI locally by speeding up the process.

### More

- ~[ ] Added/updated tests~

### Additional Notes

For my tests, I saw a decrease of building time, going from 201.5s to 32.2s.

I also tried to use [`pnpm`](https://pnpm.io), but it requires more changes on the building process.

This feature was already introduced in #846, but was revert in #5249.


rtribotte (approved):
LGTM 👌

0xflotus (changes requested):
You should consider to copy the resulting yarn.lock also to the docker image. 
https://github.com/traefik/traefik/blob/11ff7415121ad4231e7b95d984ab2670f672c371/webui/Dockerfile#L9

Now you copy the package-lock.json. If you use yarn, yarn will create a yarn.lock as equivalent.

webui/Dockerfile
```diff
@@ -7,10 +7,11 @@ ENV PLATFORM_URL=${ARG_PLATFORM_URL}
 RUN mkdir -p $WEBUI_DIR
 
 COPY package.json $WEBUI_DIR/
-COPY package-lock.json $WEBUI_DIR/
+COPY yarn.json $WEBUI_DIR/
```
0xflotus, line 10: The file is called `yarn.lock`

0xflotus (approved):

kevinpollet (changes requested):
If we are switching to `yarn`, I would prefer to replace every `npm` command by the `yarn` equivalent to avoid using two different tools for the webui.

kevinpollet (approved):
LGTM 🚀

PR #8784, opened 2022-02-17
Fix empty WebUI static assets directory

rtribotte:
### What does this PR do?

This PR makes the WebUI static assets directory always contains a file. So that the go embed directive to load those assets never fails.

### Motivation

This PR fixes the empty WebUI static assets directory issue when using the Traefik v2 module which uses go embed as a dependency.

### More

- [ ] Added/updated tests

### Additional Notes

A rework of the `build-image` makefile goal is also brought by this PR. It should ensure that each execution produces the same behavior.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -23,7 +23,7 @@ jobs:
 
       - name: Build webui
         run: |
-          make generate-webui
+          make clean-webui generate-webui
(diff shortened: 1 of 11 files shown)


tomMoulard (approved):
LGTM :ok_hand:

mmatur (approved):

PR #8823, opened 2022-03-07
doc: fix, docker uses Label(), not Tag()

mpl:
### What does this PR do?

This PR fixes the doc about constraints for the docker provider.

### Motivation

The doc was incorrectly stating that the constraints are using the "Tag" syntax instead of the "Label" syntax.

### More

~- [ ] Added/updated tests~


rtribotte (approved):
LGTM 👌

PR #8834, opened 2022-03-11
Move consulcatalog provider to only use health apis

chuckyz:
### What does this PR do?

This PR drops the usage of the Catalog API within the Consul Catalog provider and switches to only using the Health API(s).  This is nicely compatible with the Connect endpoints as they live under the same API and output exactly the same types.

### Motivation

The Health API with the cache query parameter set to true uses the streaming health check system.  This means that if streaming is enabled in a Consul cluster, Traefik is now effectively 'free' to run, going so far as an operator can set the refresh interval down to 1 second and likely not notice anything in their Consul cluster.

Even if streaming is disabled, this still cuts the request rate in half, meaning ~1/2 bandwidth usage on top of that which is helpful in really large clusters.

### More

- [x] Checked tests

### DIFF
--- a/pkg/provider/consulcatalog/consul_catalog.go
+++ b/pkg/provider/consulcatalog/consul_catalog.go
@@ -265,31 +265,31 @@ func (p *Provider) getConsulServicesData(ctx context.Context) ([]itemData, error
 		}
 
 		for _, consulService := range consulServices {
-			address := consulService.ServiceAddress
+			address := consulService.Service.Address
 			if address == "" {
-				address = consulService.Address
+				address = consulService.Node.Address
 			}
 
-			namespace := consulService.Namespace
+			namespace := consulService.Service.Namespace
 			if namespace == "" {
 				namespace = "default"
 			}
 
-			status, exists := statuses[consulService.ID+consulService.ServiceID]
+			status, exists := statuses[consulService.Node.ID+consulService.Service.ID]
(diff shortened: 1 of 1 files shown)


chuckyz:
@ldez @kevinpollet is the semophore failure a thing I could've caused or a flakey test?  it's hard to determine from the output (for me).

chuckyz:
@kevinpollet for some internal research I compared the following structs:

The catalog API uses https://pkg.go.dev/github.com/hashicorp/consul/api#CatalogService this type.

For comparison both the health service and connect apis use: https://pkg.go.dev/github.com/hashicorp/consul/api#ServiceEntry

ServiceEntry has https://pkg.go.dev/github.com/hashicorp/consul/api#Node and https://pkg.go.dev/github.com/hashicorp/consul/api#AgentService types within it. 

When looking at the CatalogService type, The "Service(...)" fields are all contained within AgentService, and the non-prefixed fields are identical to the Node fields.

kevinpollet:
Hello @chuckyz,

As your pull request is made from an organization repository, it is not possible for us to rebase it, and our automation will not work properly.

Can you re-open this pull request from your personal fork?

kevinpollet:
@chuckyz we opened a new pull request (#9140) and added you as a co-author, thanks again for your contribution.

Superseded by #9140

PR #8847, opened 2022-03-18
docs/content/https/acme.md: "certificates resolvers"

NReilingh:
### What does this PR do?

Applies consistent usage of "certificates resolvers" to the Let's Encrypt documentation.

### Motivation

Confusing `certificatesResolvers` with `certificateResolvers` or `certificateResolver` seems to be a common issue. Cite:

* https://github.com/traefik/traefik/issues/7414
* https://community.traefik.io/t/a-solution-to-the-incredibly-unhelpful-the-router-uses-a-non-existent-resolver-letsencrypt-message/3859

I completely did not notice the double-plural when I was testing my initial configuration, and figuring out the issue was quite frustrating. I believe if the documentation used "certificates resolvers" consistently I might have noticed my error earlier.

It might also be worthwhile to create a highlight panel in the docs to reinforce the double-pluralization.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -11,19 +11,19 @@ You can configure Traefik to use an ACME provider (like Let's Encrypt) for autom
     Use Let's Encrypt staging server with the [`caServer`](#caserver) configuration option
     when experimenting to avoid hitting this limit too fast.
 
-## Certificate Resolvers
+## Certificates Resolvers
 
-Traefik requires you to define "Certificate Resolvers" in the [static configuration](../getting-started/configuration-overview.md#the-static-configuration),
+Traefik requires you to define "Certificates Resolvers" in the [static configuration](../getting-started/configuration-overview.md#the-static-configuration),
 which are responsible for retrieving certificates from an ACME server.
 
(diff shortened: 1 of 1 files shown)


mpl:
Sorry, we indeed messed up and we probably should never have pluralised "certificate" in any of our variable names (i.e. it should be certificateResolver and certificateResolvers, but never certificatesResolver(s)). However, just because we made a mistake there, we should not mirror the same mistake in plain english in the documentation.

NReilingh:
@mpl I disagree that it's a grammatical error if that's what the configuration object is actually called. Either way, the documentation is currently inconsistent looking at the plain english alone, so that should be corrected, no?

And any thoughts to making an explicit callout/highlight/warning panel about this?

I'd be happy to add another commit to this branch making the opposite change for consistency and suggesting a Warning panel addition if that makes sense to you.

mpl:
> Either way, the documentation is currently inconsistent looking at the plain english alone, so that should be corrected, no?

How so? Have you found any instance where it is not "certificate resolver" for the singular, or where it is not "certificate resolvers" for the plural?

NReilingh:
@mpl Yes. Opened a new PR #8859

PR #8857, opened 2022-03-20
Adds a new Branching middleware

ddtmachado:
### What does this PR do?

Adds a new middleware to allow conditional activation of another middleware chain based on request data.

### Motivation

A personal project :)

### More

- [X] Added/updated tests

### Additional Notes

Potentially solves many use cases where a middleware or middleware chain should only be activated conditionally.


zikphil:
Do we have an idea of when this will be merged?

tfny:
Hey @zikphil, we are currently reviewing it, but we hope it will be ready for 2.9, our next minor release. :)

ldez:
This PR is invalid: middleware must not be used for routing.

ddtmachado:
it does not change routing from my perspective @ldez ;) 
In the sense it doesn't touch servers endpoints, transport or matching rules but rather just conditionally triggers another chain of middleware

ldez:
> just conditionally triggers another chain of middleware

Trigger based on request -> routing

ldez:
My design review: this middleware is doing routing (this is not an opinion, this is the reality), and the routing must not be handled by middleware but by routers.

This is broken by design.
This is an invalid solution to a misinterpreted and overrated problem.

As a Traefik maintainer, my opinion is to decline this PR.

Rule n°1 of open-source: no is temporary, yes is forever.

ddtmachado:
@ldez fair point, but I would argue there is no true reality here as everything is just an opinionated design choice in the end. I'm fine if Traefik's design as a product is to never touch routing from a middleware, and in that case I would look into improving the router mechanism instead.

Would you be satisfied if the solution was moved to the router? To something like this maybe:

```
http:
  routers:
    test:
      entryPoints:
      - web
      rule: Host(`test.docker.localhost`)
      service: test
      middlewares:
        - name: mycustomheaders 
          condition: Header(`foo`,  `bar`)

        - name: mystripprefix
```

That would be a breaking change of course

ldez:
>  I would argue there is no true reality here as everything is just an opinionated design choice in the end

The reality exists, not everything is a matter of opinion.

> Would you be satisfied if the solution was moved to the router?

I will not discuss that in this PR. Understanding the real needs and finding a proper solution takes more than 5 min.

ddtmachado:
I'm not trying to find the proper solution here but rather to take a decision about this PR specifically.
Considering I got mixed feedback other maintainers and your hard no I'm closing it in favor of opening a design discussion for v3.

PR #8864, opened 2022-03-22
Update entrypoint.md to add consistent CLI syntax

nstankov-bg:
### What does this PR do?
- Makes sure CLI syntax is consistent. (one per row)

| Before | After 
|--------|--------|
|    | 

### Motivation

I was being lazy and could not easily copy-paste for my compose file.

### More

### DIFF
--- a/docs/content/routing/entrypoints.md
+++ b/docs/content/routing/entrypoints.md
@@ -258,7 +258,6 @@ In most scenarios, this entryPoint is the same as the one used for TLS traffic.
     ```
     
     ```bash tab="CLI"
-    ## Static configuration
     --experimental.http3=true 
     --entrypoints.name.http3
     ```
@@ -290,7 +289,6 @@ It can be used to override the authority in the `alt-svc` header, for example if
     ```
     
     ```bash tab="CLI"
-    ## Static configuration
     --experimental.http3=true 
     --entrypoints.name.http3.advertisedport=443
     ```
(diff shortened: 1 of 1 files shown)


rtribotte:
Hello @nstankov-bg,

Thanks for your contribution!

We took a look at the PR and we think that the current syntax/form of the CLI examples in the documentation is preferable.
We think it is simpler to break it down into multiple lines than inlining the arguments.

As a side note, if we were inclined to change it, for the sake of consistency, we would have to change every example in the documentation.

Unfortunately, for those reasons, we are declining your PR.

docs/content/routing/entrypoints.md
```diff
@@ -288,7 +290,9 @@ It can be used to override the authority in the `alt-svc` header, for example if
     ```
     
     ```bash tab="CLI"
-    --experimental.http3=true --entrypoints.name.http3.advertisedport=443
+    ## Static configuration
```
rtribotte, line 293:
```suggestion
```

docs/content/routing/entrypoints.md
```diff
@@ -258,7 +258,9 @@ In most scenarios, this entryPoint is the same as the one used for TLS traffic.
     ```
     
     ```bash tab="CLI"
-    --experimental.http3=true --entrypoints.name.http3
+    ## Static configuration
```
rtribotte, line 261:
```suggestion
```

rtribotte (changes requested):

rtribotte (approved):
Thanks 👍

nstankov-bg:
Thank you!

PR #8893, opened 2022-03-28
Return TLS unrecognized_name error when no certificate is available

### LINKED ISSUE #8888: Handle TLS for unknown names correctly
### Welcome!

### What did you do?

Set up traefik on a host with multiple DNS names. One of them is available for traefik in a cert and has routers, the other one is not supposed to be available on https. When connecting with that name that's not supposed to be available via https to port 443, traefik should give back a TLS error 112 (unrecognized_name).

### What did you see instead?

When connecting with that name that's not supposed to be available via https to port 443, traefik gives back a TLS error 80 (internal_error), but should give back a TLS error 112 (unrecognized_name).

Version:      2.5.2
Codename:     livarot
Go version:   go1.17
Built:        2021-09-02T15:07:43Z
OS/Arch:      linux/amd64

Not applicable, as this is not config dependent.

_No response_

rtribotte:
### What does this PR do?

This PR removes the returned errors from the `GetCertificate` func in TLSConfig on the server side and replaces them with a log.

### Motivation

This will allow the private `getCertificate` from `crypto/tls` package to return the correct `errNoCertificates` error (see crypto/tls [common.go](https://cs.opensource.google/go/go/+/dev.boringcrypto.go1.17:src/crypto/tls/common.go;l=1058))
when no certificate can be found.

Fixes #8888

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/tls/tlsmanager.go
+++ b/pkg/tls/tlsmanager.go
@@ -143,7 +143,10 @@ func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 		if isACMETLS(clientHello) {
 			certificate := acmeTLSStore.GetBestCertificate(clientHello)
 			if certificate == nil {
-				return nil, fmt.Errorf("no certificate for TLSALPN challenge: %s", domainToCheck)
+				log.WithoutContext().Debugf("TLS: no certificate for TLSALPN challenge: %s", domainToCheck)
+				// Let crypto/tls determine the appropriate error and the TLS alert.
+				// https://cs.opensource.google/go/go/+/dev.boringcrypto.go1.17:src/crypto/tls/common.go;l=1058
+				return nil, nil
 			}
 
 			return certificate, nil
@@ -155,7 +158,10 @@ func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 		}
 
 		if sniStrict {
-			return nil, fmt.Errorf("strict SNI enabled - No certificate found for domain: %q, closing connection", domainToCheck)
+			log.WithoutContext().Debugf("TLS: strict SNI enabled - No certificate found for domain: %q, closing connection", domainToCheck)
+			// Let crypto/tls determine the appropriate error and the TLS alert.
+			// https://cs.opensource.google/go/go/+/dev.boringcrypto.go1.17:src/crypto/tls/common.go;l=1058
+			return nil, nil
 		}
 
 		log.WithoutContext().Debugf("Serving default certificate for request: %q", domainToCheck)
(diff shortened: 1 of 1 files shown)


pkg/tls/tlsmanager.go
```diff
@@ -143,7 +143,10 @@ func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 		if isACMETLS(clientHello) {
 			certificate := acmeTLSStore.GetBestCertificate(clientHello)
 			if certificate == nil {
-				return nil, fmt.Errorf("no certificate for TLSALPN challenge: %s", domainToCheck)
+				log.WithoutContext().Debugf("TLS: no certificate for TLSALPN challenge: %s", domainToCheck)
+				// Let crypto/tls determine the appropriate error and the TLS alert.
```
mpl, line 147:
```suggestion
				// We want the user to eventually get the (alertUnrecognizedName) "unrecognized
				// name" error.
				// Unfortunately, if we returned an error here, since we can't use
				// the unexported error (errNoCertificates) that our caller (config.getCertificate
				// in crypto/tls) uses as a sentinel, it would report an (alertInternalError)
				// "internal error" instead of an alertUnrecognizedName.
				// Which is why we return no error, and we let the caller detect that there's
				// actually no certificate, and fall back into the flow that will report
				// the desired error.
```

pkg/tls/tlsmanager.go
```diff
@@ -155,7 +158,10 @@ func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 		}
 
 		if sniStrict {
-			return nil, fmt.Errorf("strict SNI enabled - No certificate found for domain: %q, closing connection", domainToCheck)
+			log.WithoutContext().Debugf("TLS: strict SNI enabled - No certificate found for domain: %q, closing connection", domainToCheck)
+			// Let crypto/tls determine the appropriate error and the TLS alert.
+			// https://cs.opensource.google/go/go/+/dev.boringcrypto.go1.17:src/crypto/tls/common.go;l=1058
```
mpl, line 163:
```suggestion
			// Same comment as above, as in the isACMETLS case.
```

kevinpollet (approved):
LGTM

tomMoulard (approved):

jcgruenhage:
Thanks, that was quicker than expected. I had already applied a similar patch locally and spent quite some time setting up a test env to get this going (haven't compiled traefik from source in quite some time) and can confirm that with these changes it behaves as expected. Thanks for the amazing reaction time!

mpl (approved):

PR #8905, opened 2022-03-31
Update logger_formatters_test.go

wangfeng22:
type *testing.T has no method Setenv

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/middlewares/accesslog/logger_formatters_test.go
+++ b/pkg/middlewares/accesslog/logger_formatters_test.go
@@ -2,6 +2,7 @@ package accesslog
 
 import (
 	"net/http"
+	"os"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


kevinpollet:
Hello @wangfeng22,

Thanks for your contribution,

The `testing.T` type has a `Setenv` method as described here https://pkg.go.dev/testing@go1.17.8#T.Setenv.
Maybe this is because the Go version you are using to compile Traefik locally is not the right one, see https://doc.traefik.io/traefik/contributing/building-testing/

Thus, we will decline this pull request.

wangfeng22:
> Hello @wangfeng22,
> 
> Thanks for your contribution,
> 
> The `testing.T` type has a `Setenv` method as described here https://pkg.go.dev/testing@go1.17.8#T.Setenv. Maybe this is because the Go version you are using to compile Traefik locally is not the right one, see https://doc.traefik.io/traefik/contributing/building-testing/
> 
> Thus, we will decline this pull request.

Yes, thank you

Issue #8820, opened 2022-03-04
Narrow down debug level for ECS rate exceeded error


applike-ss:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.containo.us) and didn't find any.

### What did you expect to see?

I just updated our traefik version.
While doing so, i saw that in the logs we've got stuff like:
```
time="2022-03-04T14:39:10Z" level=error msg="Unable to describe task definition: ThrottlingException: Rate exceeded\n\tstatus code: 400, request id: foo" providerName=ecs
time="2022-03-04T14:39:10Z" level=error msg="Provider connection error failed to get ECS configuration: ThrottlingException: Rate exceeded\n\tstatus code: 400, request id: foo, retrying in 344.333774ms" providerName=ecs
```
This is happening because traefik is running into an AWS request rate limit while discovering aws ecs clusters and services and task definitions used.

I would assume a rate limit to be not an error, therefor i would like to suggest to make it a warning.
Furthermore a gracefull exponential or linear backoff with a small note upon success or an error upon eventual failure would be great!

applike-ss:
Sounds perfect :-)

applike-ss:
Ok thanks :)

PR #8916, opened 2022-04-04
Remove duplicate error logs

rtribotte:
### What does this PR do?

This PR removes several duplicated error logs that could be produced by the ECS provider.

### Motivation

To narrow down the verbosity.
To partially address feedback from #8820.

### More

- [ ] Added/updated tests

### Additional note

The first intention was also to narrow down the log level for the specific `ThrottlingException: Rate exceeded` error, as mentioned in proposal #8820.
But, after a second thought, this will make it non-trivial to debug this error in production if the level moves to DEBUG.
Hence, and as of now, the PR only addresses the removal of duplicate error logs.

### DIFF
--- a/pkg/provider/ecs/ecs.go
+++ b/pkg/provider/ecs/ecs.go
@@ -263,8 +263,7 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
 			return !lastPage
 		})
 		if err != nil {
-			logger.Error("Unable to list tasks")
-			return nil, err
+			return nil, fmt.Errorf("list tasks: %w", err)
 		}
 
 		// Skip to the next cluster if there are no tasks found on
@@ -370,7 +369,6 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
 }
 
 func (p *Provider) lookupEc2Instances(ctx context.Context, client *awsClient, clusterName *string, ecsDatas map[string]*ecs.Task) (map[string]*ec2.Instance, error) {
-	logger := log.FromContext(ctx)
 	instanceIds := make(map[string]string)
 	ec2Instances := make(map[string]*ec2.Instance)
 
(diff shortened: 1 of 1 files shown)


pkg/provider/ecs/ecs.go
```diff
@@ -263,8 +263,7 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
 			return !lastPage
 		})
 		if err != nil {
-			logger.Error("Unable to list tasks")
-			return nil, err
+			return nil, fmt.Errorf("list tasks: %w", err)
```
mpl, line 266:
```suggestion
			return nil, fmt.Errorf("listing tasks: %w", err)
```

pkg/provider/ecs/ecs.go
```diff
@@ -389,8 +387,7 @@ func (p *Provider) lookupEc2Instances(ctx context.Context, client *awsClient, cl
 			Cluster:            clusterName,
 		})
 		if err != nil {
-			logger.Errorf("Unable to describe container instances: %v", err)
-			return nil, err
+			return nil, fmt.Errorf("describe container instances: %w", err)
```
mpl, line 390:
```suggestion
			return nil, fmt.Errorf("describing container instances: %w", err)
```

pkg/provider/ecs/ecs.go
```diff
@@ -418,8 +415,7 @@ func (p *Provider) lookupEc2Instances(ctx context.Context, client *awsClient, cl
 				return !lastPage
 			})
 			if err != nil {
-				logger.Errorf("Unable to describe instances: %v", err)
-				return nil, err
+				return nil, fmt.Errorf("describe instances: %w", err)
```
mpl, line 418:
```suggestion
				return nil, fmt.Errorf("describing instances: %w", err)
```

pkg/provider/ecs/ecs.go
```diff
@@ -440,8 +436,7 @@ func (p *Provider) lookupTaskDefinitions(ctx context.Context, client *awsClient,
 				TaskDefinition: task.TaskDefinitionArn,
 			})
 			if err != nil {
-				logger.Errorf("Unable to describe task definition: %v", err)
-				return nil, err
+				return nil, fmt.Errorf("describe task definition: %w", err)
```
mpl, line 439:
```suggestion
				return nil, fmt.Errorf("describing task definition: %w", err)
```

kevinpollet (approved):
Thanks 👍

tomMoulard (approved):
LGTM :ok_hand:

Issue #7366, opened 2020-09-29
SMTP-STARTTLS proxying


schildbach:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Support SMTP-STARTTLS proxying. Traefik would accept connections on port 25, handle encryption (via SMTP-STARTTLS) and proxy the protocol as plain non-encrypted SMTP to a mail server (MTA). Note that unlike TCP proxying, this would involve rewriting parts of the SMTP protocol (at least the "welcome phase").

Note that sadly [there is no SMTPS](https://en.wikipedia.org/wiki/SMTPS#Difference_between_SMTPS_and_smtps) (SMTP via TLS). It has been revoked by the IANA in favour of SMTP-STARTTLS, and SMTPS is almost never tried by other MTAs. So Traefik's TLS support can't be used for SMTP.

SMTP is the only widely used protocol that currently isn't handled by Traefik. This means if you're running a mix of services (e.g. mail server and web server) SSL certificates are handled by different entities (e.g. certbot for mail, traefik for http). It would be nice if Traefik could uniformly proxy all common services, greatly simplifying such setups.

Note also that nginx [supports this special case](https://docs.nginx.com/nginx/admin-guide/mail-proxy/mail-proxy/#setting-up-ssl-tls-for-a-mail-proxy).

mschneider82:
@frederikbosch  if you use TCP, you can already handle smtp to a mta, this issue is about implementing smtp protocol on top of tcp, so traefik acts as an smtpd server and handle tls.

If you use TCP to forward SMTP to a MTA, the client IP is currently lost, so you may have an open relay or cannot use spam ip blacklists. For this i have opened a pull req for supporting proxy protocol to the upstream mta in traefik to keep the client ip.

frederikbosch:
@mschneider82 Thank you for the clarification, and good work on the PR, hopefully it will be merged anytime soon!

nounoursPand:
I'm very interesting for solution for PostgreSQL, when will it arrive ?

khashashin:
@nounoursPand  same here

wbolster:
:pray:  please refrain from making non-constructive ‘when will it be done?’ and ‘me too’ comments. those are actually detrimental to open-source projects and especially disrespectful to their maintainers.

(speaking for myself only but i am confident this sentiment is widely shared among open source maintainers)

designermonkey:
[...]

PR #8935, opened 2022-04-13
add postgres StartTLS proxying

maigl:
### What does this PR do?

This adds postgres starttls support which allows to reverse proxy postgres connections through Traefik (via SNI).

Related to #7366

### Motivation

I'm part of a team that provides managed databases inside our company.
We needed to reverse-proxy many databases and to do this we use Traefik enhanced with this feature successfully for almost 2 years now.

#### Context:

We have a number of PostgreSQL databases running inside our kubernetes cluster.
All database instances are accessed through Traefik via a single IP address. In order to
access the right PostgreSQL instances from the clients we want to use different subdomains.
One subdomain should be the target address of a specific PostgreSQL instance and we want to
route to it via the SNI (server name indication in TLS header).
This approach seems to be the only possible solution in our setup. We don't have any other information
to identify the target PostgreSQL instance from a client request as it is a tcp connection and not http-based.

#### Problem:

PostgreSQL performs a StartTLS handshake before sending any TLS header. Since Traefik doesn't perform
this handshake no TLS header is sent and hence Traefik cannot extract the target server name.

#### Idea:

We propose the following steps:

- Traefik simulates a PostgreSQl server and performs StartTLS handshake with client
client now sends TLS header
- Traefik extracts SNI
- Traefik performs StartTLS handshake with server
- Traefik routes connection from client to target PostgreSQL instance

### More

- [x] Added/updated tests

### Additional Notes

Other starttls flavors might be added in future. 

<sub>Michael Kuhnt <michael.kuhnt@daimler.com> Mercedes-Benz Tech Innovation GmbH ([ProviderInformation](https://github.com/mercedes-benz/daimler-foss/blob/master/PROVIDER_INFORMATION.md))
</sub>


maigl:
@ldez and @rtribotte I created the pr from a fresh fork directy from traefik/traefik, I only changed the title. The checkmark 'allow edits for maintainers .. ' is set. I hope I did it right this time.

tfny:
Just a quick update, we are still reviewing.

dani:
Do I understand it correctly that this PR would only allow passthrough connexion to postgres ? Or would traefik also be able to do the TLS termination and use either plain TCP or separated TLS to the postgres server ?

rtribotte:
Hello @dani,

We are working on an alternative implementation, that should supersede this PR soon.
The support we are targeting will only be done for Postgres STARTTLS, and it will be possible to do both TLS termination and passthrough.

rtribotte:
Since PR https://github.com/traefik/traefik/pull/9377 has been opened, we are closing this PR.

PR #8951, opened 2022-04-20
Change default TLS options for more security

### LINKED ISSUE #6756: Please disable TLS 1.0 and 1.1 by default
Using TLS 1.0 or 1.1 is considered insecure these days. It should be used only by those who really know what they're doing. For that reason, I suggest disabling TLS 1.0 and 1.1 by default.

I'm on Traefik 2.2.1.

ddtmachado:
### What does this PR do?

Change the default TLS Options for more security while maintaining compatibility with most clients

### Motivation

https://github.com/golang/go/issues/45428

Fixes #6756 

Go will remove support for TLS 1.0 and 1.1 soon, so I think it's better if we prepare in advance by changing the default values and then adding a notice on the deprecations page.

**Plus** lots of complaints about Traefik not being secure enough out of the box, from the TLS perspective.

### More

- [X] Added/updated tests

### Additional Notes

On this PR I trust that Go will keep an up to date list of secure ciphers during its lifecycle, then the tests with static values will ensure we notice when things change so we don't break compatibility without notice.


rtribotte (approved):
LGTM 👌

ldez (approved):
LGTM

juliens (approved):
LGTM

PR #8970, opened 2022-04-26
Added ACME provider configuration to enable/disable certificate auto renewal

smasset-orange:
### What does this PR do?

Adds an ACME provider configuration item to enable/disable automatic certificate renewal

### Motivation

Should help with concurrency issues when running multiple instances of Traefik (where an external component could handle generation and renewals)

### More

- [ ] Added/updated tests

### Additional Notes

Couldn't find an elegant way to unit/integration test this feature. Happy to implement it under guidance though.

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -153,6 +153,8 @@ and starts to renew certificates 30 days before their expiry.
 When using a certificate resolver that issues certificates with custom durations,
 one can configure the certificates' duration with the [`certificatesDuration`](#certificatesduration) option.
 
+One can also disable automatic renewal of certificates with the [`autoRenewCertificates`](#autorenewcertificates) option.
+
 !!! info ""
     Certificates that are no longer used may still be renewed, as Traefik does not currently check if the certificate is being used before renewing.
 
@@ -593,6 +595,35 @@ certificatesResolvers:
 | >= 7 days            | 1 day             | 1 hour                  |
 | >= 24 hours          | 6 hours           | 10 min                  |
 | < 24 hours           | 20 min            | 1 min                   |
+
+### `autoRenewCertificates`
+
+_Optional, Default=true_
+
+The `autoRenewCertificates` option controls whether certificates should be automatically renewed.
+It defaults to `true`.
+
+```yaml tab="File (YAML)"
+certificatesResolvers:
+  myresolver:
+    acme:
+      # ...
+      autoRenewCertificates: false
+      # ...
+```
+
+```toml tab="File (TOML)"
+[certificatesResolvers.myresolver.acme]
+  # ...
+  autoRenewCertificates=false
+  # ...
+```
+
+```bash tab="CLI"
+# ...
+--certificatesresolvers.myresolver.acme.autorenewcertificates=false
+# ...
+```
 
 ### `preferredChain`
 
--- a/docs/content/https/ref-acme.toml
+++ b/docs/content/https/ref-acme.toml
@@ -30,6 +30,14 @@
   #
   # certificatesDuration=2160
 
+  # Automatically renew certificates.
+  # It defaults to true.
+  #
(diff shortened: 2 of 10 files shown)


ddtmachado:
Hi @smasset-orange, thanks for your PR

Could you clarify the use cases requiring that option?

From a first glance, it would seem like a mitigation and not a fix, for a use case that's not even supported anymore on the project: running the LE/ACME integration in scalable HA mode. Just to be clear, Traefik 1.x had support to use a K/V store for certificates which would enable HA, but that was dropped since 2.x because of the added complexity and issues.

So IMHO, allowing the user to partially disable a feature to mitigate that use case only doesn't seem reasonable to me. And if that is indeed the only case, I would prefer to try and figure out a better solution for supporting HA with LE/ACME integration in the long run.

smasset-orange:
Hi and thanks for the quick feedback.

FYI, I had already read other discussions regarding K/V certificate store and the reasons which lead to the decision to drop it in v2. I'm also aware that Traefik EE has native support for HA.

My (current) personal use case for this feature is to have multiple instances of CE Traefik v2 only taking care of proxying requests while still retrieving certificates from ACME provider.

If disabling certificate renewal in the current implementation of local ACME provider is not an option, another solution would be to have another implementation (completely without the renewal code that I surrounded with an if statement in this PR) that could be selected from static configuration. This would bring no change for most users while still allowing others to select a different provider implementation.

IMHO the configuration item added in this PR brings the same flexibility with far less code change, but the alternative could be a first step in a more complete HA solution at the cost of later availability.

ldez:
Disabling certificate renewal will not fix the real problem: if you start several Traefik instances at the same time you will still have problems with multiple calls to the ACME server and concurrency access to the `acme.json` file.

> Should help with concurrency issues when running multiple instances of Traefik (where an external component could handle generation and renewals)

If you use an external component that could handle generation and renewals, you must not use the Traefik ACME implementation but the [classic certificates feature](https://doc.traefik.io/traefik/v2.6/https/tls/).

smasset-orange:
Here's a diagram of what I had in mind:
Traefik

I believe this solves most (if not all) race conditions apart from renewals which might be requested from proxy instances (I'm currently relying on shorter `certificatesDuration` setting on ACME instance to have it renew certificates before the proxy instances get a chance to).

FYI, the config exporter is a small Python API running as a Docker service aggregating outputs of different Traefik API calls to the ACME instance and formatting it so that it can be accepted by proxy instances using the HTTP provider. It currently sits outside of the ACME instance but I might open a dedicated PR against Traefik APIs to remove it.

One other potential remaining race condition could be that proxy instances retrieve new entrypoint/service before the ACME challenge is resolved (unless it can be filtered out in API calls based on its status) and each request a certificate too. This would eventually stabilize but might hit Let's Encrypt failure quotas in the process.

kevinpollet:
Hello @smasset-orange,

Thanks for the explanation and the nice diagram. Unfortunately, as said in https://github.com/traefik/traefik/pull/8970#issuecomment-1110208164, disabling the certificate renewal in order to do HA does not seem reasonable. As you stated, this solution does not resolve all race conditions, so it is not satisfactory.

According to this, we will decline this pull request.

smasset-orange:
> As you stated, this solution does not resolve all race conditions, so it is not satisfactory.

@kevinpollet, to be fair, I never stated it didn't resolve all race conditions, just mentioned one *could* remain.

smasset-orange:
Can someone comment on whether Treafik /routers/ API returns routers before their challenge has been resolved or not (when requested to only return routers with status `enabled`) ?

ddtmachado:
@smasset-orange no, there is no check made. It will return all routers regardless (actually there is no real connection between a router and a certificate request or challenge, Traefik just uses the router rule as a trigger to get one and add it to the cert store)

PR #8976, opened 2022-04-29
Support certificates configuration in TLSStore CRD

### LINKED ISSUE #7880: Support multiple certificates for TLSStore defined via K8S

Feature

The goal of this proposal is to be able to have multiple certificate in a single TLSStore defined via K8S.

### Issue
I have two wildcard certificates for two domain zones.
Both zones have a DNS entry for my k8s pod. Therefore I defined an IngressRoute listening on both domains.
Issue #4756 wont solve my issue, as I want to use both domains in a single IngressRoute definition using `tls: {}`
The following IngressRoute works when using the `file provider` to define the TLSStore but it is not possible to define the same TLSStore within k8s.
```
kind: IngressRoute
  name: configs
  namespace: configs
spec:
  entryPoints:
  - websecure
  routes:
  - kind: Rule
    match: Host(`config.cluster-name.some.example`)
    services:
    - name: configs
      port: 80
  - kind: Rule
    match: Host(`test.some.example`)
    services:
    - name: configs
      port: 80
  tls: {}
```

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ traefik_namespace }}-v2-dynamic-file
  namespace: {{ traefik_namespace }}-v2
data:
  tls.yaml: |
    tls:
      certificates:
        - certFile: /cert/example/tls.crt
          keyFile: /cert/example/tls.key
          stores:
            - default
        - certFile: /cert/default/tls.crt
          keyFile: /cert/default/tls.key
          stores:
            - default
      stores:
        default:
          defaultCertificate:
            certFile: /cert/example/tls.crt
            keyFile: /cert/example/tls.key
``` 

### Example of a TLSStore definition
Beside the `defaultCertificate` key add a `certificates` key where a list of certs can be parsed.
```
kind: TLSStore
metadata:
  name: default
  namespace: traefik-v2
spec:
  certificates:
    - secretName: some-cert
    - secretName: k8s-cluster-name-wildcard
  defaultCertificate:
    secretName:  k8s-cluster-name-wildcard
```

### Side note
The issue is duplicate to #7210 but it was closed with duplicate of #4756 even so it is not duplicate in my opinion.
I also opened this issue as non of the forum entry's about this issue were answered in any way.
Issue #4756 won't solve my issue as I would need to define an IngressRoute for each Domain I want to use. This would result in an doubled amount of definitions which are unneeded.

kevinpollet:
### What does this PR do?

This pull request allows adding certificates to a TLSStore (as of now, only to the default one) by using the [TLStore CRD](https://doc.traefik.io/traefik/routing/providers/kubernetes-crd/#kind-tlsstore).

### Motivation

Fixes #7880

### More

- [X] Added/updated tests

### Additional Notes


docs/content/https/tls.md
```diff
@@ -364,8 +364,8 @@ spec:
 
 ### Strict SNI Checking
 
-With strict SNI checking enabled, Traefik won't allow connections from clients
-that do not specify a server_name extension or don't match any certificate configured on the tlsOption.
+With strict SNI checking enabled, Traefik won't allow connections from clients that do not specify a server_name extension 
+or don't match any configured certificate (which does not include the DefaultCertificate).
```
mpl, line 368:
```suggestion
or don't match any of the configured certificates. The DefaultCertificate is irrelevant on that matter.
```

docs/content/routing/providers/kubernetes-crd.md
```diff
@@ -1630,13 +1630,17 @@ or referencing TLS stores in the [`IngressRoute`](#kind-ingressroute) / [`Ingres
...
+      defaultCertificate:                      # [2]
+        secretName: secret                     
     ```
 
-| Ref | Attribute    | Purpose                                                                                                                                                     |
-|-----|--------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| [1] | `secretName` | The name of the referenced Kubernetes [Secret](https://kubernetes.io/docs/concepts/configuration/secret/) that holds the default certificate for the store. |
+| Ref | Attribute            | Purpose                                                                                                                                    |
+|-----|----------------------|--------------------------------------------------------------------------------------------------------------------------------------------|
+| [1] | `certificates`       | List of Kubernetes [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) that hold certificate to add to the store.         |
```
mpl, line 1642:
```suggestion
| [1] | `certificates`       | List of Kubernetes [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) , each of them holding a key/certificate pair to add to the store.         |
```

docs/content/routing/providers/kubernetes-crd.md
```diff
@@ -1630,13 +1630,17 @@ or referencing TLS stores in the [`IngressRoute`](#kind-ingressroute) / [`Ingres
...
+| [2] | `defaultCertificate` | Name of a Kubernetes [Secret](https://kubernetes.io/docs/concepts/configuration/secret/) that holds the default certificate for the store. |
```
mpl, line 1643:
```suggestion
| [2] | `defaultCertificate` | Name of a Kubernetes [Secret](https://kubernetes.io/docs/concepts/configuration/secret/) that holds the default key/certificate pair for the store. |
```

pkg/provider/kubernetes/crd/traefik/v1alpha1/tlsstore.go
```diff
@@ -20,13 +20,14 @@ type TLSStore struct {
 
 // TLSStoreSpec configures a TLSStore resource.
 type TLSStoreSpec struct {
-	DefaultCertificate DefaultCertificate `json:"defaultCertificate"`
+	DefaultCertificate *Certificate  `json:"defaultCertificate,omitempty"`
```
mpl, line 23: TODO: add doc on fields so that generated description in the yamls is correct.

docs/content/reference/dynamic-configuration/traefik.containo.us_tlsstores.yaml
```diff
@@ -36,9 +36,20 @@ spec:
           spec:
             description: TLSStoreSpec configures a TLSStore resource.
             properties:
+              certificates:
+                items:
+                  description: Certificate holds a secret name for the TLSOption resource.
```
mpl, line 41: this is probably wrong: it's a list, and it has nothing to do with the TLSOption (rather the TLSStore?).
But it should be fixed directly as the godocs on the struct, right?

integration/fixtures/k8s/01-traefik-crd.yml
```diff
@@ -1356,9 +1356,20 @@ spec:
...
+                        Secret to specify the certificate details.
+                      type: string
+                  required:
+                  - secretName
+                  type: object
+                type: array
               defaultCertificate:
-                description: DefaultCertificate holds a secret name for the TLSOption
-                  resource.
+                description: Certificate holds a secret name for the TLSOption resource.
```
mpl, line 1372: Since you made the choice to have the certificates directly within a tlsstore definition (which is the "opposite" of what we do everywhere else), this would set a precedent. I think we need to take the time to consider whether in the future (traefik3 ?) we want to abandon the notion that there could be multiple tls stores. And maybe we want to be consistent everywhere else, and also have the certificates directly in the tlsstore.

mpl: nvm, we discussed it internally.

mpl (approved):

rtribotte (approved):
LGTM 👌

tomMoulard (approved):
LGTM :ok_hand:

pkarrot:
is it possible to make a key (secret) per domain?

kevinpollet:
Hello @pkarrot,

> is it possible to make a key (secret) per domain?

Yes, with the new `certificates` property, it is possible to define multiple Kubernetes secrets containing a TLS certificate which will be added to the default TLS store.

bangonkali:
Hi @pkarrot, how do i typically know which https://hub.docker.com/_/traefik?tab=tags docker tag will the be merged to ? I just tried with the following **Helm Values**:

```yaml

image:
  name: traefik
  pullPolicy: Always
  tag: v2.8

pilot:
  enabled: false

```

I still get the following:

```
 error validating data: ValidationError(TLSStore.spec): unknown field "certificates" in us.containo.traefik.v1alpha1.TLSStore.spec; if you choose to ignore these errors, turn validation off with --validate=false
```

Btw, I did a helm `helm uninstall --namespace=traefik traefik` and then:
```
helm install --namespace=traefik \
  --values=values/traefik-values.yaml \
  traefik \
  traefik/traefik
``` 
to upgrade from an older version of **Traefik**.

Any advise?

bangonkali:
Ok so it may just be an issue with my Helm Install. I tried the following and the new CRD's seem to work just fine. 

```bash
kubectl apply -f https://raw.githubusercontent.com/traefik/traefik/v2.8/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
kubectl apply -f https://raw.githubusercontent.com/traefik/traefik/v2.8/docs/content/reference/dynamic-configuration/kubernetes-crd-rbac.yml
```

Issue #8990, opened 2022-05-05
Retry middleware should only retry idempotent HTTP methods (not POST and PATCH)


hoh:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

The retry middleware should by default only retry idempotent HTTP methods (`OPTION`, `GET`, `HEAD`, `PUT`, and `DELETE`). Retrying `POST` and `PATCH` can lead to issues on the backend.

## Context
While _"As soon as the server answers, the middleware stops retrying"_, in HTTP the server does not have to start answering before it starts to process the request. This means that, for non idempotent methods, the server may change state in a way that should not be called multiple times, and this before starting to answer.

This can lead to issues such as double-spending, multiple object creation, ... and should be avoided by default.

An option to allow the middleware to retry on non idempotent methods would be a nice to have.

Duplicate of the frozen-due-to-age: https://github.com/traefik/traefik/issues/6942

lplazas:
@kevinpollet is this something you will be working on? If not I'd love to have a go at to get kickstarted in the codebase.

github-sys-admin:
go ahead

kevinpollet:
Hello @lplazas,

Thanks for asking, we discussed it with the team and someone is already working on the subject. Feel free to take another subject, help is always welcome.

lplazas:
Thank you for your reply @kevinpollet. Seems like last Friday I was trolled by someone who posted that I could go ahead with this task, but now that comment is deleted. 
I worked on this over the weekend so I'll anyways open a PR, feel free to check it if you think it's worth it or use it as a base for the final change: https://github.com/traefik/traefik/pull/8998

kevinpollet:
Hello @hoh,

Thanks for your interest in Traefik,

As stated in https://github.com/traefik/traefik/pull/8998#issuecomment-1123705126, the middleware will not retry a request if one byte of data is sent to the backend. We had wrongly classified this issue as a legitimate feature.

According to this, we will decline this feature request.

PR #8998, opened 2022-05-09
Add flag to disable retry middleware on non-idempotent methods

lplazas:
### What does this PR do?

Adds a new flag to disable retries on POST and PATCH methods, as retries on these kinds of methods
may lead to unexpected behaviour in the backend.
All integration tests have the flag set to true to allow for pre-existing functionality to work as expected
Closes: https://github.com/traefik/traefik/issues/8990

### More

- [X] Added/updated tests


build.Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN mkdir -p /usr/local/bin \
 RUN curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | bash -s -- -b $GOPATH/bin v1.45.0
 
 # Download misspell binary to bin folder in $GOPATH
-RUN curl -sfL https://raw.githubusercontent.com/client9/misspell/master/install-misspell.sh | bash -s -- -b $GOPATH/bin v0.3.4
+RUN go install github.com/client9/misspell/cmd/misspell@v0.3.4
```
lplazas, line 19: This never went through on my machine, this change fixed it. 

I had some troubles running integration tests such as: `(runtime: failed to create new OS thread (have 2 already; errno=22))` as I use a arm64 machine.

kevinpollet:
Hello @lplazas,

Thanks for your contribution,

As stated in the following comment https://github.com/traefik/traefik/pull/6941#issuecomment-647595361, the middleware will not retry a request if one byte of data is sent to the backend. We had wrongly classified #8990 as a legitimate feature since it is actually safe to use this middleware with non-idempotent methods.

According to this, we will close this pull request.

PR #9002, opened 2022-05-10
docs/routers: add a note about "hanged" tcp connections

romantomjak:
### What does this PR do?

Clarifies why some connections may appear to have "hanged".

### Motivation

I was struggling to understand why am I not seeing the SMTP greeting when connecting to an email server behind Traefik. It appeared to work once I hit "enter", but obviously that's not how the protocol works and from the documentation it wasn't immediately clear that Traefik was waiting for bytes to arrive before deciding how to handle the request.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/routing/routers/index.md
+++ b/docs/content/routing/routers/index.md
@@ -672,6 +672,16 @@ If no matching route is found for the TCP routers, then the HTTP routers will ta
 If not specified, TCP routers will accept requests from all defined entry points.
 If you want to limit the router scope to a set of entry points, set the entry points option.
 
+!!! important "Entry points and routing"
+
+    Routers by default accept requests from all defined entry points, meaning that, incoming
+    requests can be a mix of TLS and non-TLS ones. In order to correctly handle the request,
+    Traefik needs to wait for the first few bytes to arrive. For protocols that do not send
+    first, such as SMTP, this may cause a situation where both sides are waiting for data and
+    the connection appears to have "hanged". To avoid this, please ensure that there is only
+    one router assigned to the entry point, in which case Traefik will forward the request
+    immediately.
+
 ??? example "Listens to Every Entry Point"
 
     **Dynamic Configuration**


docs/content/routing/routers/index.md
```diff
@@ -672,6 +672,16 @@ If no matching route is found for the TCP routers, then the HTTP routers will ta
...
 
+!!! important "Entry points and routing"
+
+    Routers by default accept requests from all defined entry points, meaning that, incoming
+    requests can be a mix of TLS and non-TLS ones. In order to correctly handle the request,
+    Traefik needs to wait for the first few bytes to arrive. For protocols that do not send
+    first, such as SMTP, this may cause a situation where both sides are waiting for data and
+    the connection appears to have "hanged". To avoid this, please ensure that there is only
+    one router assigned to the entry point, in which case Traefik will forward the request
+    immediately.
```
rtribotte, line 683: Hello @romantomjak,

Thanks for your contribution!

While I definitely see the situation you want to clarify here, I fear that this statement is not true.
On one EntryPoint, to bypass the reading of the first bytes of the ClientHello, and then forward directly to the backend, there should be no TLS routers (TCP and HTTP) attached to that EntryPoint, and at least one Non-TLS TCP router.

Would you mind rewording to include those conditions?

romantomjak: Heya 👋  Yes, absolutely. I've tried to further clarify the requirements, but it's tricky to describe it simply 😅  Let me know if it needs some more thought.

mpl:
Hello @romantomjak , 
this PR raised some questions about how we want to document some aspects related to what you're saying.
We need to think a bit more about it, and we might have to commit some things ourselves on the PR if that's alright with you.

romantomjak:
Yeah of course! No worries at all

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks 👍

Issue #8988, opened 2022-05-05
Router priority is not shown in dashboard


aleksvujic:
Let's say that you define a router in `traefik.yml` like this:

```yml
http:
  routers:
    catch-all:
      rule: "HostRegexp(`{host:(blog|support)(-(dev|test))?\\.company\\.(local|com)}`)"
      priority: 1
      tls: {}
      service: "noop@internal"
```

Once you open the dashboard, priority isn't shown anywhere in router details:

![image](https://user-images.githubusercontent.com/17112542/160404637-3d1ae312-fed9-4357-97d0-08ae73118792.png)

Can priority info be added to router details?

tomMoulard:
Fixed by #9004

PR #5981, opened 2019-12-10
Display priority in web UI

sh7dm:
### What does this PR do?
Add priority to main routing table

### Motivation
Adresses #5667, but only partially.

### More

- [ ] Added/updated tests
- [ ] Added/updated documentation

PR #9004, opened 2022-05-10
Added router priority to webui's list and detail page

bendre90:
### What does this PR do?

This PR addresses the feature request in #8988 

### Motivation

I want to contribute to open source :)

###  #More

- [ ] Added/updated tests

### Additional Notes

None

### Screenshots 

screen_list
screen_details
e

### DIFF
--- a/webui/src/_mixins/GetTableProps.js
+++ b/webui/src/_mixins/GetTableProps.js
@@ -87,6 +87,14 @@ const allColumns = [
     label: 'Provider',
     fieldToProps: row => ({ name: row.provider }),
     component: ProviderIcon
+  },
+  {
+    name: 'priority',
+    align: 'left',
+    label: 'Priority',
+    component: QChip,
+    fieldToProps: () => ({ class: 'app-chip app-chip-accent', dense: true }),
+    content: row => row.priority
   }
 ]
 
@@ -98,7 +106,8 @@ const columnsByResource = {
     'name',
     'service',
     'tls',
-    'provider'
+    'provider',
+    'priority'
   ],
   udpRouters: ['status', 'entryPoints', 'name', 'service', 'provider'],
   services: ['status', 'name', 'type', 'servers', 'provider'],
--- a/webui/src/components/_commons/PanelRouterDetails.vue
+++ b/webui/src/components/_commons/PanelRouterDetails.vue
@@ -73,6 +73,18 @@
           </div>
         </div>
       </q-card-section>
+      <q-card-section v-if="data.priority">
+        <div class="row items-start no-wrap">
+          <div class="col">
+            <div class="text-subtitle2">PRIORITY</div>
+            <q-chip
+              dense
+              class="app-chip app-chip-entry-points">
+              {{ data.priority }}
+            </q-chip>
+          </div>
+        </div>
+      </q-card-section>
       <q-card-section v-if="data.error">
         <div class="row items-start no-wrap">
           <div class="col">


ldez:
Related to #5981

bendre90:
> Related to #5981

im sorry, looks like i didnt dig deep enough in the first place.
Looks like this issue was addressed before and got abadoned.

So we close this PR aswell as the issue related?

tfny:
Hey @bendre90, we are not going to close it, we want to review it. We agree with the idea, but need to review the approach. However, it may take some time.

tfny:
@bendre90, hey, so after reviewing this, we really like the idea.  Unfortunately, the UI needs some work.  

The problem we see is that the way priority is represented will make it impossible to quickly understand what is happening once you get 5+ routers attached.  The numbers are too long to be quickly identified and processed by humans in this format.  

If you can address that problem and make the priority more human-friendly, we think it would be a great addition to Traefik Proxy.

bendre90:
@tfny Thanks for your response! I appreciate your considerations about this contribution.

Unfortunately, I am unable to come up with an idea to convert the integers to a human friendly format.
In my opinion, numbers make a good scale for priorities, but I can understand, that many routers with high digit count can make them hard to understand quickly.

May I ask for clarification of what kind of format transformation you thought of?

So far I just computed missing priorities as documented [here](https://doc.traefik.io/traefik/routing/routers/#priority) with 
cc6e182 . This way, at least the missing values are filled.
Maybe a sortable table would do the trick now?
Another idea would be to order the priority integers and just display their index on the list? This would alter the real value.

Thanks for your time and input!

mpl:
> @tfny Thanks for your response! I appreciate your considerations about this contribution.
> 
> Unfortunately, I am unable to come up with an idea to convert the integers to a human friendly format. In my opinion, numbers make a good scale for priorities, but I can understand, that many routers with high digit count priorities can make them hard to understand quickly.
> 
> May I ask for clarification of what kind of format transformation you thought of?
> 
> So far I just computed missing priorities as documented [here](https://doc.traefik.io/traefik/routing/routers/#priority) with [cc6e182](https://github.com/traefik/traefik/commit/cc6e182c65aadcd2db1bf58cd20c8ffefea97e5c) . This way, at least the missing values are filled. Maybe a sortable table would do the trick now? Another idea would be to order the priority integers and just display their index on the list? This would alter the real value.
> 
> Thanks for your time and input!

Hi @bendre90 ,

imho it's not that big of a deal to have such large numbers in the column, and it indeed becomes even less of a problem if the routers can be sorted by priority (because then you don't actually have to eyeball the numbers themselves most of the time).
And on the flipside, I think it's somewhat important to keep track of the actual value, in case the user starts debugging in the code and printing the values, so he can relate them to what's showing up in the web UI. So if I were to display them in another "format" in the web UI I would try to do it so that it's also very easy to get the actual value.
For example, wouldn't it be possible to display only the 3 or 4 first digits (and maybe followed by an ellipsis, like so: `498...`), and when you hover on it, you get a tooltip with the actual full value?

bendre90:
> Hi @bendre90 ,
> 
> imho it's not that big of a deal to have such large numbers in the column, and it indeed becomes even less of a problem if the routers can be sorted by priority (because then you don't actually have to eyeball the numbers themselves most of the time).
> 
> And on the flipside, I think it's somewhat important to keep track of the actual value, in case the user starts debugging in the code and printing the values, so he can relate them to what's showing up in the web UI. So if I were to display them in another "format" in the web UI I would try to do it so that it's also very easy to get the actual value. For example, wouldn't it be possible to display only the 3 or 4 first digits (and maybe followed by an ellipsis, like so: `498...`), and when you hover on it, you get a tooltip with the actual full value?

Hi @mpl 
thanks for your great input, it really helped me out.

So I did as you suggested: the MainTable is now sortable, and also the hover feature you described was added.
I feel like, the implementation is a bit messy since the table on the listing components reloads every 5000ms which makes it hard to sort that array. So I copied the data array into a scoped array to finally sort this. Is this the right way to do this?
I'd appreciate criticism on the added code, since vuejs is quite new to me.

There was also an error on the infinitescroll component, which was already there before I started editing the MainTable.vue file and i felt fixing this is outside this PR's scope.
That's why I commented out the `this.onLoadMore({ page: index })` section in the `handleMore` method to silence the errors.

There is code in the backend to sort the json output for the tables. For example, in [`pkg/api/handler_http.go:75`](https://github.com/traefik/traefik/blob/master/pkg/api/handler_http.go#L75). Should we remove that?

Thanks for your time

mpl:
> > Hi @bendre90 ,
> > imho it's not that big of a deal to have such large numbers in the column, and it indeed becomes even less of a problem if the routers can be sorted by priority (because then you don't actually have to eyeball the numbers themselves most of the time).
> > And on the flipside, I think it's somewhat important to keep track of the actual value, in case the user starts debugging in the code and printing the values, so he can relate them to what's showing up in the web UI. So if I were to display them in another "format" in the web UI I would try to do it so that it's also very easy to get the actual value. For example, wouldn't it be possible to display only the 3 or 4 first digits (and maybe followed by an ellipsis, like so: `498...`), and when you hover on it, you get a tooltip with the actual full value?
> 
> Hi @mpl thanks for your great input, it really helped me out.
> 
> So I did as you suggested: the MainTable is now sortable, and also the hover feature you described was added. I feel like, the implementation is a bit messy since the table on the listing components reloads every 5000ms which makes it hard to sort that array. So I copied the data array into a scoped array to finally sort this. Is this the right way to do this? I'd appreciate criticism on the added code, since vuejs is quite new to me.

Thanks a lot for your effort.
We're trying to see if our coworkers, who are more competent than we are in frontend stuff, can have a look at your work.
But if that fails, we'll come back to it ourselves and see if we can help.

paulocfjunior:
Hello, I'd like to contribute with a few suggestions on this regarding UX, please feel free to comment and give your opinion on them:

### Display vs omit the Priority Number
I understand it can be a large number, and the user may also define a custom value for it, so I'd propose to keep it visible in the UI and allow users to sort by just clicking the column header will make use of learned behavior, so it should feel natural for the users.

An alternative that would allow us to hide the column, would be to have a button next to the search box saying "Sort by Priority", where we can choose any column or Priority, which would not be a visible column.

### The number can be huge
IMO truncate the number with ellipsis would not help much because it would prevent us from knowing the number dimension. Imagine, for example, a table where all the priorities have more than 6 digits, no matter the sorting, we would not know which one is larger.

To workaround this problem, I'd propose using the scientific notation (e.g. `1.23e+20` instead of `123456789101112131415`), which is shorter but at the same time gives the user the notion of how big the number is, for example between `1.23e+15` and `1.23e+20`, without adding more characters. 
This can be easily calculated in Javascript using:
```js
    Number.parseFloat(theBigNumber).toExponential(howManyDecimals)
```
> ⚠️ In any case we need to to allow the user to see the exact number, we may display a tooltip with it when hovering over the number.

### Sort the items in the backend
As we have a table that is loaded page by page, and the data is refreshed in an interval, we must sort the data before filtering it to get the page slice, for this pattern we usually have some query params we send in the request to inform the backend how the data should be sorted and which portion of the data we are willing to get, for example:
```
/api/data/list?sortBy=priority&direction=asc&limit=10&offset=10
```
Given this query, we know that the backend should return:
- `limit=10&offset=10`: a list of 10 items at maximum, skipping the first 10
- `sortBy=priority`: ordered by priority, could be any column
- `direction=asc`: ordered in ascending direction, could be `desc` for descending

> ⚠️ Important: If we _don't_ sort it in the backend, we might have a situation where after the user sort the table the top priority rule is not in the screen yet, because it was never fetched.

This approach requires the Priority to be calculated in the backend, as we will use it for sorting.

### How to calculate the priority
I'd recommend calculating the priority in the backend using the same methods and parsers as when traefik does it for the routing. I'm wondering if calculating it in the front end by parsing the length of the rule could cause differences with the priority calculated when the rule will be applied to the requests.

mpl:
Hey @bendre90 , 

you can see above the advice of our resident front-end expert ;)
Would you be interested in iterating on the PR some more to implement his recommendations?

bendre90:
Hey @mpl ,

thanks for your effort to let paulocfjunior invest his valueable time on this PR. I really appreciate a lot!
Of course I'm still interested on iterating this PR.

I will code on this, this weekend !

Thanks a lot!

bendre90:
Hey,

I refactored this PR to make use of backend sorting instead of sorting results in the frontend.
Thanks again for the detailed review. It helped me to realize the issues with my implementation. :)

Additional questions:
- Should we "activate" the scientific notation behavior only for higher numbers?
I'd suggest a threshold of five digits - `>=10000`. (Maybe [this](https://en.wikipedia.org/wiki/Subitizing) can be helpful to determine a threshold?)
- The backend uses a different `sort.Slice()` function for each possible combination of asc/desc and column name.
Is there a better way to do this?
- Should I add tests for sorting results?
- Some columns can have the same value. To prevent items jumping around in the table, I used the name as a second property without the user's knowledge. Could this lead to confusion?
(We could make the problematic columns, which are not that easily comparable, unsortable. But I thought there should be as many sortable columns as possible.)
- The priority doesn't have always a value. I couldn't figure out, where I could add the missing value in the backend. I'd either need to ask for help here or we render an empty string to the table. (like the "Type" on the "HTTP Service Table")

Thank you for your time and opinions.

rtribotte:
Hello @bendre90,

Thanks for all the work here!

> Should we "activate" the scientific notation behavior only for higher numbers?
>I'd suggest a threshold of five digits - >=10000. (Maybe [this](https://en.wikipedia.org/wiki/Subitizing) can be helpful to determine a threshold?)

This is only my opinion but I would say that the threshold should be the `math.MaxInt32` value.
It is only 9-digit and I think that besides the 6-digit of exponential, it does not take much place, at least no more than what is affordable in this column:
image

Also `math.MaxInt32` is the default value for several internal routers, and basically a value that should never be lower than any other priority that is computed on any router (except when the priority on a router has been set with a greater value to purposely "override" an internal router).

> The backend uses a different sort.Slice() function for each possible combination of asc/desc and column name.
> Is there a better way to do this?

Well, in this particular case, I don't think so.

> Should I add tests for sorting results?

Sure, please.

> Some columns can have the same value. To prevent items jumping around in the table, I used the name as a second property without the user's knowledge. Could this lead to confusion?
>(We could make the problematic columns, which are not that easily comparable, unsortable. But I thought there should be as many sortable columns as possible.)

This should make the sorting deterministic in any case, which I think is good, even if the user does not have knowledge about this. Choosing an arbitrary second criterion to sort will not be less accurate than randomness.

> The priority doesn't have always a value. I couldn't figure out, where I could add the missing value in the backend. I'd either need to ask for help here or we render an empty string to the table. (like the "Type" on the "HTTP Service Table")

On the backend side, if not user-defined, the priority is computed to the character length of the router's rule.
Thus, it is done during the application of the dynamic configuration (when Traefik is actually building routers). So in the runtime configuration that the API is manipulating and serving, the priority remains the value received from the providers, and the priority can be empty.

The WebUI could compute the priority in the same way, but then you'll still miss the priority when sorting.
As of now, I feel this should be discussed with other maintainers, so we'll let you know what we think is the best approach.

mpl:
@bendre90 
Hi, just to clarify, @rtribotte was thinking out loud, but we are not yet in full agreement among us about what we want to do for the notation. So please hold your horses on that matter, as we don't want to make you do work that we might change our minds about (again) later.
Sorry about that!

mpl:
> Hey,

Hey,
We've discussed again, so we can come back to you in agreement.
Let's assume that whatever @rtribotte already answered, and that isn't addressed again below as an addendum, can be taken as our answer on the matter. ;)

> Additional questions:
> 
> * Should we "activate" the scientific notation behavior only for higher numbers?
>   I'd suggest a threshold of five digits - `>=10000`. (Maybe [this](https://en.wikipedia.org/wiki/Subitizing) can be helpful to determine a threshold?)

Let's go with something simple for now. Let's go with an arbitrary 10 digits (which should be able to hold any int32) as the cutoff.
And let's drop the scientific notation. Let's just add an ellipsis when the number is over 10 digits in size.
And I suppose that for it to look good we should keep the number of digits before the ellipsis so that:
size of (truncated number + ellipsis) ~= 10 digits
But maybe I'm naive and it won't look good with any font, I dunno.

> * The priority doesn't have always a value. I couldn't figure out, where I could add the missing value in the backend. I'd either need to ask for help here or we render an empty string to the table. (like the "Type" on the "HTTP Service Table")

yeah, as @rtribotte say, let's just compute again the priority wherever needed in the API handler. We'll to try to do better/optimize in a subsequent PR. You can assume that the priority is simply the length of the rule, just as it is done in pkg/muxer/tcp/mux.go.

> Thank you for your time and opinions.

Thank you for your patience.

bendre90:
Hi @mpl 

sorry for my late response.
Do i understand correctly, the priority number should be displayed as rtribotte suggested like this: `[0-9]{2}\.\.\.[0-9]{3}` when the number has 10 or more digits?

I'm going to propose another iteration on the weekend, please have some more patience with me.
Thanks a lot!

paulocfjunior:
Hello folks, IMHO as long as we sort the data in the backend and show in the table that the column is getting sorted, the presentation of the numbers is not going to be a big deal, so I don't have a strong opinion on that matter.

pkg/muxer/tcp/mux.go
```diff
@@ -13,6 +13,12 @@ import (
 	"github.com/vulcand/predicate"
 )
 
+var catchAllParser predicate.Parser
+
+func init() {
+	catchAllParser, _ = rules.NewParser([]string{"HostSNI"})
+}
+
```
ldez, line 21: I'm not sure about that.

ldez (changes requested):

pkg/api/sort.go
```diff
@@ -0,0 +1,388 @@
...
+		return
+	}
+
+	// Descending
+	sort.Slice(results, func(i, j int) bool {
+		if fn(i) == fn(j) {
+			return results[i].name() > results[j].name()
+		}
+		return fn(i) > fn(j)
+	})
```
tomMoulard, line 387:
```suggestion
	sortByName(direction, results)

	// Ascending
	if direction == ascendantSorting {
		sort.Slice(results, func(i, j int) bool {
			return fn(i) < fn(j)
		})

		return
	}

	// Descending
	sort.Slice(results, func(i, j int) bool {
		return fn(i) > fn(j)
	})
```

This could be clearer, but at the expense of some perfs

pkg/config/runtime/runtime_http.go
```diff
@@ -71,6 +71,8 @@ func unique(src []string) []string {
 // RouterInfo holds information about a currently running HTTP router.
 type RouterInfo struct {
 	*dynamic.Router // dynamic configuration
+	// EffectivePriority is the user set priority if defined, or the rule's computed priority otherwise.
+	EffectivePriority int `json:"effectivePriority,omitempty"`
```
tomMoulard, line 75: I don't see any advantage to have `EffectivePriority` and `Priority` altogether, I'm fine to override `Priority` with the `effectivePriority`. Therefore, no need to change integration tests rawdata assets

rtribotte (approved):
Thanks 👍

tomMoulard (approved):
LGTM :ok_hand:

PR #9012, opened 2022-05-12
Add support for deriving distinct traefik `Services` entries from consul tag

breathe:
This PR adds functionality needed to address this: https://github.com/traefik/traefik/issues/8987

### What does this PR do?

Adds ability to specify a suffix in consul which is appended to the end of the traefik service name to form a service that is distinct in traefik despite having the same consul service name.  This mechanism is needed to allow a canary and non-canary service (as these entities exist in nomad) to be routed to independently via traefik.

### Motivation

Wanted to be able to use traefik with consul and nomad and take advantage of the recommended nomad canary deployment strategy.

Fixes https://github.com/traefik/traefik/issues/8987

### More

- [X] Added/updated tests

### DIFF
--- a/pkg/provider/consulcatalog/connect_tls.go
+++ b/pkg/provider/consulcatalog/connect_tls.go
@@ -2,7 +2,6 @@ package consulcatalog
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/hashicorp/consul/agent/connect"
 	"github.com/traefik/traefik/v2/pkg/config/dynamic"
@@ -55,7 +54,7 @@ func (c *connectCert) equals(other *connectCert) bool {
 func (c *connectCert) serversTransport(item itemData) *dynamic.ServersTransport {
 	itemName := item.Name
 	if item.ExtraConf.ConsulNameSuffix != "" {
-		itemName = strings.TrimSuffix(item.Name, "-"+item.ExtraConf.ConsulNameSuffix)
+		itemName = item.ConsulServiceName
 	}
 
 	spiffeIDService := connect.SpiffeIDService{
--- a/pkg/provider/consulcatalog/connect_tls.go
+++ b/pkg/provider/consulcatalog/connect_tls.go
@@ -2,6 +2,7 @@ package consulcatalog
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/consul/agent/connect"
 	"github.com/traefik/traefik/v2/pkg/config/dynamic"
@@ -52,15 +53,20 @@ func (c *connectCert) equals(other *connectCert) bool {
 }
 
 func (c *connectCert) serversTransport(item itemData) *dynamic.ServersTransport {
+	itemName := item.Name
+	if item.ExtraConf.ConsulNameSuffix != "" {
+		itemName = strings.TrimSuffix(item.Name, "-"+item.ExtraConf.ConsulNameSuffix)
+	}
+
 	spiffeIDService := connect.SpiffeIDService{
 		Namespace:  item.Namespace,
 		Datacenter: item.Datacenter,
-		Service:    item.Name,
+		Service:    itemName,
 	}
 
 	return &dynamic.ServersTransport{
 		// This ensures that the config changes whenever the verifier function changes
-		ServerName: fmt.Sprintf("%s-%s-%s", item.Namespace, item.Datacenter, item.Name),
+		ServerName: fmt.Sprintf("%s-%s-%s", item.Namespace, item.Datacenter, itemName),
 		// InsecureSkipVerify is needed because Go wants to verify a hostname otherwise
 		InsecureSkipVerify: true,
 		RootCAs:            c.getRoot(),
--- a/pkg/provider/consulcatalog/config.go
+++ b/pkg/provider/consulcatalog/config.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"strings"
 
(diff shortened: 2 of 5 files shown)


kevinpollet:
Hello @breathe,
 
Thanks for your contribution,

This PR is labeled as `needs-design-review`, as you can see in our [contribution guide](https://github.com/traefik/contributors-guide/blob/master/review_guidelines.md), this means that we have to take a longer look to evaluate how it would interact with the other parts of Traefik.

breathe:
Thanks kevinpollet!  Please let me know how things go -- and I would be happy to make any changes/improvements needed to help get something like this into traefik!

For some additional information about this PR ... I tested this out in one of our development environments and it works in the way that I was expecting it to for at least half the problem I am trying to solve ...  What works is that it _does_ create a unique traefik service for consul service registrations with the `traefik.consulnamesuffix` consul tag defined.  Instances of the service with the tag can be routed to independently of instances of the same service that are registered into traefik without that tag.  So that part works great!

What I still haven't gotten to work is zero-downtime 'promotion' of the canary.  Say I have a service with 10 instances and wish to use 2 canary instances.  On deployment of a new version nomad spins up the canary instances which are registered into traefik as a distinct service and independently routable.  I do whatever testing I wish against the canary and then when satisfied I promote the canary in nomad.  After promotion nomad kicks of a 'rolling update' with max_parallel equal to 1 -- which as I understand it is supposed to bring down 1 instance of the service, 'replace it' with one of the canary instances if any exist, validate health of the canary after replacement, and then proceed to spin down and replace other instances 1 at a time repeating this process (when a canary instance doesn't exist then a new instance or the service is spun up instead).

So far I haven't been able to get zero-downtime deployments to work following this process.  To be clear -- I've also been unable to get zero-downtime deployments to work using traefik <-> consul <-> nomad without canary deployment either -- so its possible I'm just missing some required configuration somewhere ... a timeout or a parameter related to the health check configuration in traefik maybe ...

apollo13:
I'd also love to see something like this, but my approach got shut down :/ https://github.com/traefik/traefik/pull/7890

> I was able to get zero-downtime deploys to work but it requires co-operation from the service. The service must begin failing health-check when given the shutdown signal and stay alive long enough for traefik to remove the instance from its pool 

I wonder if this is something we should talk about with the nomad/consul teams. Ie we need a way to mark a service instance as "going down".

FWIW if you are on traefik 2.7 you can use consul watch queries so traefik doesn't have to poll every 15 seconds and will get informed of container shutdowns earlier.

pkg/provider/consulcatalog/connect_tls.go
```diff
@@ -52,15 +53,20 @@ func (c *connectCert) equals(other *connectCert) bool {
 }
 
 func (c *connectCert) serversTransport(item itemData) *dynamic.ServersTransport {
+	itemName := item.Name
+	if item.ExtraConf.ConsulNameSuffix != "" {
+		itemName = strings.TrimSuffix(item.Name, "-"+item.ExtraConf.ConsulNameSuffix)
```
apollo13, line 58: This smells like a hack :D I think it would make more sense to add a `ConsulServiceName` to `itemData` so you always have access to the original unmodified service name instead of trimming here.

breathe:
> > I was able to get zero-downtime deploys to work but it requires co-operation from the service. The service must begin failing health-check when given the shutdown signal and stay alive long enough for traefik to remove the instance from its pool
> 
> I wonder if this is something we should talk about with the nomad/consul teams. Ie we need a way to mark a service instance as "going down".
> 

I put in a feature request related to this to the nomad folks -- with an idea for how nomad could support easier zero-downtime deploy of consul services -- it looks like it might get picked up! 

https://github.com/hashicorp/nomad/issues/13032#issuecomment-1147482150

pkg/provider/consulcatalog/connect_tls.go
breathe: @apollo13 I updated the PR to store the original ConsulServiceName on itemData and use that where the unmodified consul service name is needed

kevinpollet:
Hello @breathe,

We reviewed this pull request and, after some discussion, we decided to decline it.

Adding a new tag to add a suffix to the generated service name only for consul-catalog is not something we would consider and we would prefer a global approach. We also declined the following pull requests https://github.com/traefik/traefik/pull/7890 and https://github.com/traefik/traefik/pull/7395 because of that reason.

Therefore, we will open a new feature request because we think that allowing the naming of the services for the label providers by using a dedicated label is something needed.

In the meantime, it is possible to name the service by defining a [service option](https://doc.traefik.io/traefik/routing/providers/consul-catalog/#services) in the `canary_tags`. For example, one could use the following tag to define the server scheme `traefik.http.services.my-canary-service-name.loadbalancer.server.scheme=http`.

kevinpollet:
Hello all,

After thinking a bit more about this subject, we thought that introducing a `traefik.consulcatalog.canary` and `traefik.nomad.canary` would solve this issue without having to introduce a global label to name the service.

We opened https://github.com/traefik/traefik/pull/9216 for that, feel free to test it and give us some feedback.

apollo13:
Hi @kevinpollet, thank you this looks great. So if I understand this correctly one would simply set `traefik.consulcatalog.canary=true` which would result in a new service name being generated by adding a hash of the tags?

kevinpollet:
> So if I understand this correctly one would simply set traefik.consulcatalog.canary=true which would result in a new service name being generated by adding a hash of the tags?

Yes 👍

PR #9013, opened 2022-05-13
Fix: Do not ignore the error in the user configuration

fungiboletus:
### What does this PR do?

This PR return an error instead of creating a default SSL configuration when the TLS configuration is invalid.

### Motivation

See issue  #5565

### More

- [ ] Added/updated tests

### Additional Notes

Feel free to close this pull-request, but it looked like a very simple fix.

### DIFF
--- a/pkg/tls/tlsmanager.go
+++ b/pkg/tls/tlsmanager.go
@@ -125,7 +125,7 @@ func (m *Manager) Get(storeName, configName string) (*tls.Config, error) {
 		err = fmt.Errorf("unknown TLS options: %s", configName)
 	}
 	if err != nil {
-		tlsConfig = &tls.Config{}
+		return nil, err
 	}
 
 	store := m.getStore(storeName)


tfny:
Hey @fungiboletus, Thanks for this PR!  We agree that we should return an error and as we are beginning to plan for Traefik 3.0, this is the perfect time for a breaking change.  

However, before we go into the review, we need a few things on this change.  We are a little worried that the answer might be simplistic. Would you help us out by doing an impact study to look for any unintended impacts?  We also noticed that it has not been tested.  What tests did you run, can you share them and the results?

fungiboletus:
Well, to be honest, I didn't even compile traefik with my change. So I didn't test it. It did sound like a very straightforward fix to me.

I'm not an expert on the Traefik codebase and its uses, but I don't see any unintended impacts. Invalid configurations will now crash Traefik at start, forcing people to fix their configurations instead of default settings. Already valid configurations shouldn't have any change. A major breaking change release sounds like a good time for this indeed.

If my pull-request doesn't fit your requirements, feel free to close it.

tfny:
Great!  Then we will close this one, but keep the idea and work on it when we get a chance to make sure it does not have any unintended consequences.

PR #9018, opened 2022-05-16
Implement Traefik provider for Nomad orchestrator

shoenig:
### What does this PR do?

Implements Traefik provider for Nomad service discovery 

### Motivation

https://www.hashicorp.com/blog/nomad-1-3-adds-native-service-discovery-and-edge-workload-support
Nomad and Traefik already go great together. Now that Nomad has it's own service discovery without the use of Consul, it'd be great if Traefik could work with Nomad directly without Consul, too. 

### More

- [X] Added/updated tests

### Additional Notes

Nomad is a task orchestrator by Hashicorp. Starting in the recently
released Nomad v1.3.0 version, Nomad supports built-in service discovery
by default, independent from Consul.

This PR implements a Traefik provider to make use of Nomad's built-in
service discovery for routing ingress traffic via Traefik. The implementation
is largely inspired by the consulcatalog provider and supports the same
configuration - minus the Consul Connect and Consul Watch pieces which Nomad
does not support.

The integration tests for Nomad are a bit hacky. Nomad effectively cannot be
used inside a container (as the existing integration tests do), nor can it
run without elevated privileges. To make things work, the Nomad integration test
downloads Nomad into /tmp and runs it from there. The test must be run with sudo.
To avoid mangling the existing test flow, it is gated behind a new '-root` flag.


ddtmachado:
Hi @shoenig, thanks for the PR! I'm personally excited by this new built in service discovery in Nomad :)

With the team we took a quick look at the PR and it seems all the required pieces are in place already, so we will come back to you soon, once the design review iteration is done.

mpl:
@shoenig Hello. just to let you know, as we're extremely constrained with deadlines, we won't have time to do the usual feedback loop of review, changes, etc, with you. So we're probably going to add the changes we'd like ourselves to the PR.
Please do let us know quickly if that is not acceptable to you.

shoenig:
@mpl that's perfectly acceptable, let us know if there's anything we can do to help!

kevinpollet:
Hi @shoenig, could you provide us an svg Nomad icon to be able to display it in the webui when the provider is enabled? (something like https://github.com/traefik/traefik/blob/master/webui/src/statics/providers/consulcatalog.svg)

rtribotte (approved):
Thanks 👍

mpl (approved):

kevinpollet (approved):
Thanks 👍

DTTerastar:
This is missing the filtering based on health.  (https://github.com/traefik/traefik/blob/4f6c15cc14cdedc34484c697994134959fdff493/pkg/provider/nomad/config.go#L166)
Is nomad able to support that yet?  I saw this: https://github.com/hashicorp/nomad/issues/18361 which seems to be related to this.  Is there anything I can do to help move this along?

PR #9021, opened 2022-05-17
Fix initial tcp lookup when address is not available

### LINKED ISSUE #8347: Traefik not repeating lookup for services not present at initialisation in Docker environment
### Welcome!

### What did you do?

Set up Traefik as a proxy to another group of services using a file provider in a docker swarm environment. Expected Traefik to fail on DNS lookups for those  prior to the initialisation, but then succeed when those downstream services initialised.

### What did you see instead?

Traefik failed the initial DNS looks up on initialisation, and then continued to fail the DNS lookups for those services even once they had instantiated. The services are available by their name from within the Traefik container. Furthermore, restarting the Traefik service/container caused it to start succeeding on the DNS lookups.

behaviour present with v2.3.2, v2.3.7 v.2.4.13

```yaml
#dynamic

tcp:

  routers:
    <router_name>:
      entrypoints:
        - "<entrypoint_name>"
      rule: "HostSNI(`*`)"
      service: <service_name>

  services:
    <service_name>:
      loadBalancer:
        servers:
          - address: "<service_name_1>:<port>"
          - address: "<service_name_2>:<port>"
          - address: "<service_name_n>:<port>"
          
#static

entryPoints:
  <entrypoint_name>:
    address: ":<port>"


providers:
  file:
    filename: "/etc/traefik/traefik-dynamic.yml"
    watch: true

serversTransport:
  maxIdleConnsPerHost: -1
```

`time="<timestamp" level=error msg="no available server"` - this line repeated ad infinitum, mixed in with other items about other services which are not relevant to this bug.

ddtmachado:
### What does this PR do?

Fix #8347
since we probably missed that use case in #7370

### Motivation

When initializing a TCP proxy the code was failing fast if the initial address was not resolvable, meaning it would never get refreshed or attempt a new lookup until the dynamic configuration is reloaded to rebuild the router.

This is especially problematic when you're starting both the proxy and the target service at the same time without any guarantee that the target service name will be ready when Traefik builds the router/proxy.

### More

- [X] Added/updated tests

### DIFF
--- a/pkg/tcp/proxy.go
+++ b/pkg/tcp/proxy.go
@@ -25,20 +25,15 @@ func NewProxy(address string, terminationDelay time.Duration, proxyProtocol *dyn
 		return nil, fmt.Errorf("unknown proxyProtocol version: %d", proxyProtocol.Version)
 	}
 
-	// enable the refresh of the target only if the address is not an IP
-	refreshTarget := false
-	if host, _, err := net.SplitHostPort(address); err == nil && net.ParseIP(host) == nil {
-		refreshTarget = true
-	}
-
-	// for IP based addresses there is no need to resolve the name on every new connection
 	var target *net.TCPAddr
-	if !refreshTarget {
-		tcpAddr, err := net.ResolveTCPAddr("tcp", address)
+	// Creates the target only for IP based addresses,
+	// because there is no need to resolve the name on every new connection,
+	// and building it should happen once.
+	if host, _, err := net.SplitHostPort(address); err == nil && net.ParseIP(host) != nil {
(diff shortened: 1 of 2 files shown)


ykuksenko:
Out of curiosity what happens if the DNS record is updated with a new IP a few days after traefik is brought up?
I do not know go but it did not seem like that situation was covered.

rtribotte:
Hello @ykuksenko,

Indeed, the resolution is done each time a connection starts for hostname-based addresses.

kevinpollet (approved):
LGTM

rtribotte (approved):
LGTM 👌

PR #9024, opened 2022-05-18
Fix TCP-TLS/HTTPS routing precedence

rtribotte:
### What does this PR do?

This PR fixes the TCP-TLS/HTTPS routing precedence, following the introduction of a new TCP muxer (PR  https://github.com/traefik/traefik/pull/8182, and actually reverts it to the v2.6 behavior.

With this PR, the routing precedence is roughly as follows:
- TCP over HTTP
- HTTPS over TCP-TLS

Discrepancies for server sending first bytes support:
- On v2.6, it is possible as long as you have one and only one TCP Non-TLS ```HostSNI(`*`)``` router (so-called CatchAllNoTLS) defined.
- On v2.7, is is possible as long as you have zero TLS/HTTPS router defined.

Discrepancies in precedence between TCP and HTTP routers:
- TCP ```HostSNI(`*`)``` and HTTP ```Host(`foobar`)```
	- On v2.6 and v2.7, the TCP one takes precedence.
- TCP ```ClientIP(`[::]`)``` and HTTP ```Host(`foobar`)```
	- On v2.6, ClientIP matcher doesn't exist.
	- On v2.7, the TCP one takes precedence.

No discrepancies between TCP-TLS and HTTPS routers (considering a request/connection with the server name "foobar"):
- TCP-TLS ```HostSNI(`*`)``` and HTTPS ```Host(`foobar`)```
	- On v2.6 and v2.7, the HTTPS one takes precedence.
- TCP-TLS ```HostSNI(`foobar`)``` and HTTPS ```Host(`foobar`)```
	- On v2.6 and v2.7, the HTTPS one takes precedence (overriding the TCP-TLS one in v2.6).
- TCP-TLS ```HostSNI(`*`)``` and HTTPS ```PathPrefix(`/`)```
	- On v2.6 and v2.7, the HTTPS one takes precedence (overriding the TCP-TLS one in v2.6).
- TCP-TLS ```HostSNI(`foobar`)``` and HTTPS ```PathPrefix(`/`)```
	- On v2.6 and v2.7, the TCP-TLS one takes precedence.

### Motivation

Do not change the contract of the TCP-TLS/HTTPS routing precedence in v2.7, in comparison with v2.6.

### More

- [x] Added/updated tests

### Additional Notes

The PR adds a regression unit test to settle the TCP-TLS/HTTPS routing precedence.

### DIFF
--- a/pkg/muxer/tcp/mux.go
+++ b/pkg/muxer/tcp/mux.go
@@ -96,14 +96,14 @@ func NewMuxer() (*Muxer, error) {
 }
 
 // Match returns the handler of the first route matching the connection metadata.
-func (m Muxer) Match(meta ConnData) tcp.Handler {
+func (m Muxer) Match(meta ConnData) (tcp.Handler, bool) {
 	for _, route := range m.routes {
 		if route.matchers.match(meta) {
-			return route.handler
+			return route.handler, route.catchAll
 		}
 	}
 
-	return nil
+	return nil, false
 }
 
 // AddRoute adds a new route, associated to the given handler, at the given
@@ -139,8 +139,9 @@ func (m *Muxer) AddRoute(rule string, priority int, handler tcp.Handler) error {
 
 	newRoute := &route{
 		handler:  handler,
-		priority: priority,
 		matchers: matchers,
+		catchAll: rule == "HostSNI(`*`)",
+		priority: priority,
 	}
 	m.routes = append(m.routes, newRoute)
 
@@ -207,8 +208,9 @@ type route struct {
 	matchers matchersTree
 	// handler responsible for handling the route.
 	handler tcp.Handler
-
-	// Used to disambiguate between two (or more) rules that would both match for a
+	// catchAll indicates whether the rule is the catchAll rule (HostSNI(`*`)).
+	catchAll bool
+	// priority is used to disambiguate between two (or more) rules that would both match for a
 	// given request.
 	// Computed from the matching rule length, if not user-set.
 	priority int
--- a/pkg/server/router/tcp/router.go
+++ b/pkg/server/router/tcp/router.go
@@ -93,7 +93,7 @@ func (r *Router) ServeTCP(conn tcp.WriteCloser) {
 			return
 		}
 
-		handler := r.muxerTCP.Match(connData)
+		handler, _ := r.muxerTCP.Match(connData)
 		// If there is a handler matching the connection metadata,
 		// we let it handle the connection.
 		if handler != nil {
@@ -133,7 +133,7 @@ func (r *Router) ServeTCP(conn tcp.WriteCloser) {
 	}
 
 	if !tls {
-		handler := r.muxerTCP.Match(connData)
+		handler, _ := r.muxerTCP.Match(connData)
 		switch {
 		case handler != nil:
 			handler.ServeTCP(r.GetConn(conn, peeked))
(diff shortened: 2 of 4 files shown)


pkg/muxer/tcp/mux.go
```diff
@@ -96,14 +96,14 @@ func NewMuxer() (*Muxer, error) {
 }
 
 // Match returns the handler of the first route matching the connection metadata.
```
mpl, line 98:
```suggestion
// Match returns the handler of the first route matching the connection metadata,
// and whether the match is of kind HostSNI(*).
```

pkg/server/router/tcp/router.go
```diff
@@ -145,20 +145,33 @@ func (r *Router) ServeTCP(conn tcp.WriteCloser) {
...
 	// it is the httpsForwarder that is used for all HTTPS connections that match
 	// (which is also incidentally the same used in the last block below for 404s).
 	// The added value from doing Match, is to find and use the specific TLS config
 	// requested for the given HostSNI.
-	handler = r.muxerHTTPS.Match(connData)
-	if handler != nil {
-		handler.ServeTCP(r.GetConn(conn, peeked))
+	handlerHTTPS, catchAllHTTPS := r.muxerHTTPS.Match(connData)
+	if handlerHTTPS != nil && !catchAllHTTPS {
+		handlerHTTPS.ServeTCP(r.GetConn(conn, peeked))
```
mpl, line 155:
```suggestion
		// In order not to depart from the behaviour in 2.6, we only allow an HTTPS router
		// to take precedence over a TCP-TLS router if it is _not_ an HostSNI(*) router (so
		// basically any router that has a specific HostSNI based rule).
		handlerHTTPS.ServeTCP(r.GetConn(conn, peeked))
```

pkg/server/router/tcp/router.go
```diff
@@ -145,20 +145,33 @@ func (r *Router) ServeTCP(conn tcp.WriteCloser) {
...
+	}
+
+	// Contains also TCP TLS passthrough routes.
+	handlerTCPTLS, catchAllTCPTLS := r.muxerTCPTLS.Match(connData)
+	if handlerTCPTLS != nil && !catchAllTCPTLS {
+		handlerTCPTLS.ServeTCP(r.GetConn(conn, peeked))
+		return
+	}
+
+	// Fallback on HTTPS catchAll.
```
mpl, line 166:
```suggestion
	// Fallback on HTTPS catchAll.
	// We end up here for e.g. an HTTPS router that only has a PathPrefix rule,
	// which under the scenes is counted as an HostSNI(*) rule.
```

pkg/server/router/tcp/router.go
```diff
@@ -145,20 +145,38 @@ func (r *Router) ServeTCP(conn tcp.WriteCloser) {
...
 	// it is the httpsForwarder that is used for all HTTPS connections that match
 	// (which is also incidentally the same used in the last block below for 404s).
 	// The added value from doing Match, is to find and use the specific TLS config
 	// requested for the given HostSNI.
-	handler = r.muxerHTTPS.Match(connData)
-	if handler != nil {
-		handler.ServeTCP(r.GetConn(conn, peeked))
+	handlerHTTPS, catchAllHTTPS := r.muxerHTTPS.Match(connData)
+	if handlerHTTPS != nil && !catchAllHTTPS {
+		// In order not to depart from the behaviour in 2.6, we only allow an HTTPS router
```
tomMoulard, line 155:
```suggestion
		// In order not to depart from the behavior in 2.6, we only allow an HTTPS router
```

mpl (approved):

kevinpollet (approved):
LGTM

PR #9029, opened 2022-05-19
Update proxy.go

wangfeng22:
I think that this code can be deleted.

### More

- [ ] Added/updated tests


kevinpollet:
Please open an issue or at least fill up the motivation to explain the goal of this PR.

wangfeng22:
> Please open an issue or at least fill up the motivation to explain the goal of this PR.

ok thanks. how about this?

kevinpollet:
Why do you think that those lines are redundant?

Maybe you missed that the websocket headers case are not the same (e.g. Sec-Web**S**ocket-Key is not the same as Sec-Web**s**ocket-Key).

wangfeng22:
> Why do you think that those lines are redundant?
> 
> Maybe you missed that the websocket headers case are not the same (e.g. Sec-Web**S**ocket-Key is not the same as Sec-Web**s**ocket-Key).

ohh, yes, thank you for your reminder

PR #9030, opened 2022-05-19
IPWhiteList middleware: add the ability to configure custom body

mcorbin:
### What does this PR do?

This commit adds a new "ErrorBody" option to the IPWhiteList middleware
which can be used to return a custom message when a request is unallowed.
The default behavior is unchanged.

### Motivation

The IPWhiteList middleware currently always returns the "Forbidden"
string when a request is unallowed.
I have some use cases where I would like to customize this error message.

### More

- [x] Added/updated tests

I also tested the change locally (built traefik and tried the option).

## Additional Notes

I cannot find how to generate the related Kubernetes CRD in order to add to it the new option.

### DIFF
--- a/docs/content/middlewares/http/ipwhitelist.md
+++ b/docs/content/middlewares/http/ipwhitelist.md
@@ -73,6 +73,10 @@ http:
 
 The `sourceRange` option sets the allowed IPs (or ranges of allowed IPs by using CIDR notation).
 
+### `errorBody`
+
+The `errorBody` option allows you to configure the message to return when an IP is unallowed (default to `Forbidden`).
+
 ### `ipStrategy`
 
 The `ipStrategy` option defines two parameters that set how Traefik determines the client IP: `depth`, and `excludedIPs`.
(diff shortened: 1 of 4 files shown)


rtribotte:
Hello @mcorbin,

Thanks for the contribution!

We discussed the feature your PR is introducing and, unfortunately, we will decline it.
We don't think we should add more configuration/responsibility to the `IPWhiteList` middleware.

To serve custom responses for specific status codes, one should use the [ErrorPages](https://doc.traefik.io/traefik/middlewares/http/errorpages/) middleware. It requires to have a service that will return the appropriate response.

Just so you know, we already declined the https://github.com/traefik/traefik/pull/1576/files that was introducing the ability to load a local static file as the body of the response from the `ErrorPages` middleware.

mcorbin:
Hello @rtribotte,

Thank you for your answer.
I'm aware of the `ErrorPages` middleware but it's a complex setup to only return a different error message. We also would like to only change the error message if the `IPWhiteList` returns an error, and not the service behind traefik (if we use `ErrorPages` we will if I'm not wrong not be able to make this distinction).

rtribotte:
One way could be to have two `ErrorPages` middleware, one before and one after the `IPWhiteList` middleware.
Thus, depending on your setup (the status codes you need to handle), you may face a current limitation when stacking `ErrorPages` middleware that is described in https://github.com/traefik/traefik/issues/8866.

PR #9037, opened 2022-05-23
feat: use rand index for the first time picking in tcp/udp wrr

qmloong:
### What does this PR do?

Fix the following issure:

TCP/UDP WRR will always pick the first server when setting up.
This may cause load unbalance when 
we use traefik as forward tcp proxy and the upper application use tcp keep-alive connection.

This is a classic LB issue, the issue is that the first index servers will receive larger traffic during startup.

For example, I have, 2 clients (C), 3 traefik (T) and 3 upstream server (U).

Clients(1,2) → Traefik(1,2,3) → Upstream Severs)

1: C1 → T1 → U1

2: C2 → T2 → U1

3: C3 → T3 → U1

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/tcp/wrr_load_balancer.go
+++ b/pkg/tcp/wrr_load_balancer.go
@@ -3,8 +3,10 @@ package tcp
 import (
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/traefik/traefik/v2/pkg/log"
+	"k8s.io/apimachinery/pkg/util/rand"
 )
 
 type server struct {
@@ -22,8 +24,9 @@ type WRRLoadBalancer struct {
 
 // NewWRRLoadBalancer creates a new WRRLoadBalancer.
 func NewWRRLoadBalancer() *WRRLoadBalancer {
+	rand.Seed(time.Now().UnixNano())
 	return &WRRLoadBalancer{
-		index: -1,
+		index: rand.Int() - 1,
 	}
 }
 
--- a/pkg/udp/wrr_load_balancer.go
+++ b/pkg/udp/wrr_load_balancer.go
@@ -2,7 +2,9 @@ package udp
 
 import (
 	"fmt"
+	"math/rand"
 	"sync"
+	"time"
 
 	"github.com/traefik/traefik/v2/pkg/log"
 )
@@ -22,8 +24,9 @@ type WRRLoadBalancer struct {
 
 // NewWRRLoadBalancer creates a new WRRLoadBalancer.
 func NewWRRLoadBalancer() *WRRLoadBalancer {
+	rand.Seed(time.Now().UnixNano())
 	return &WRRLoadBalancer{
-		index: -1,
+		index: rand.Int() - 1,
 	}
 }
 


rtribotte:
Hello @qmloong,

Thanks for your contribution!

At first glance, we think that the index shouldn't be leveraged to introduce entropy in server selection.
We would prefer to insert a server in the list at a random position when it is added, instead of appending it to the list.

Do you want to take care of it, or shall we do it?

qmloong:
> We would prefer to insert a server in the list at a random position when it is added, instead of appending it to the list.

In this way, the user experience is not very good for a large number of traefik to do forward proxy clusters...Especially automated configuration or deployment for traefik.

> we think that the index shouldn't be leveraged to introduce entropy in server selection.
BTW, what's wrong with doing this? 

Refer to:
https://github.com/grpc/grpc-go/blob/30b9d59a766858a4b51148e47edb3af2766ab617/balancer/roundrobin/roundrobin.go#L63

rtribotte:
> In this way, the user experience is not very good for a large number of traefik to do forward proxy clusters...Especially automated configuration or deployment for traefik.

The server list is the result of applying the dynamic configuration, so it doesn't affect the configuration itself. What do you mean by "not very good user experience" in that case?

> BTW, what's wrong with doing this?

The problem lies in those lines:
https://github.com/traefik/traefik/blob/d5ff301d90709c0625363a5c95e0298ed8e96e47/pkg/tcp/wrr_load_balancer.go#L111-L117

The index value `-1` is "used" to set the value of the current weight on the first iteration. Changing the index value brings additional complexity compared to randomly changing the order of servers in the list.

qmloong:
```
 What do you mean by "not very good user experience" in that case?
```
For instance, we use traefik as the forward proxy for kubelet (1 : 1) as local proxy to access to 3 kube-apiserver instance A, B, C
If my cluster has 5k nodes(in other words, we have 5k traefik), then I need make 5k dynamic configurations of traefik in different order.. It's not so good right? If traefik picks server in random, so we can deploy traefik in the same dynamic configuration in ansible playbook or any othor tools.


```
The index value -1 is "used" to set the value of the current weight on the first iteration. Changing the index value will break that condition on the first iteration when the random value is different from -1, and allow the selection of any server of any weight.
```
I had known this issure.. I just ignored the first iteration. But I can fix it if the above scenario is reasonable.

rtribotte:
> For instance, we use traefik as the forward proxy for kubelet (1 : 1) as local proxy to access to 3 kube-apiserver instance A, B, C
If my cluster has 5k nodes(in other words, we have 5k traefik), then I need make 5k dynamic configurations of traefik in different order.. It's not so good right? If traefik picks server in random, so we can deploy traefik in the same dynamic configuration in ansible playbook or any othor tools.

Computing the random index will also happen on every Traefik instance independently, so the discrepancy is expected, as well as having the list of servers in a different order on each instance.
Thus, if you are raising concerns regarding performances, the impact will not be on the "critical path" (during the routing), and then, adding the server to the list at a random index shouldn't be a concern.

> I had known this issure.. I just ignored the first iteration. But I can fix it if the above scenario is reasonable.

At this point, we're not sure we're making ourselves clear. So maybe we should add the commit ourselves to the PR so you can understand/see better? Is that ok for you?

rtribotte (approved):
Thanks 👍

integration/retry_test.go
```diff
@@ -83,11 +76,12 @@ func (s *RetrySuite) TestRetryWebsocket(c *check.C) {
 	err = try.GetRequest("http://127.0.0.1:8080/api/rawdata", 60*time.Second, try.BodyContains("PathPrefix(`/`)"))
 	c.Assert(err, checker.IsNil)
 
-	// This simulates a DialTimeout when connecting to the backend server.
+	// The test only verifies that the retry middleware allows finally to reach the working service.
 	_, response, err := websocket.DefaultDialer.Dial("ws://127.0.0.1:8000/echo", nil)
 	c.Assert(err, checker.IsNil)
 	c.Assert(response.StatusCode, checker.Equals, http.StatusSwitchingProtocols)
 
+	// The test only verifies that the retry middleware allows finally to reach the working service (a second time).
```
mpl, line 84:
```suggestion
	// The test verifies a second time that the working service is eventually reached.
```

mpl (changes requested):

mpl (approved):

kevinpollet (approved):
LGTM

PR #9039, opened 2022-05-23
Update Yaegi to v0.12.0

mpl:
### What does this PR do?

Update Yaegi to v0.12.0

### Motivation

https://github.com/traefik/yaegi/compare/v0.11.0...v0.12.0

### More

~- [ ] Added/updated tests~

### Additional Notes

There's an API change, but it is a non-breaking one (new `FileSet` and `Globals` methods on the interpreter type). It is only for debugging purposes, and should not affect traefik plugins themselves.


rtribotte (approved):
LGTM 👌

PR #9051, opened 2022-05-30
Fix typo in stripPrefix middleware docs

rbarbey:
### What does this PR do?

This PR fixes a small typo in the docs for the `stripPrefix` middleware (`images.png` instead of `image.png`).

### Motivation

While reading the docs I was a wee bit confused for a second or two by the typo which let me re-read the above section to see whether I've overlooked something in the example config for the middleware. To prevent this from happening to others as well I created this tiny MR.


rtribotte (approved):
Thanks 👍

PR #9081, opened 2022-06-09
Adds a new example to the documentation

jare:
### What does this PR do?

Adds an example to the documentation when the installation
is in a dedicated namespace

### Motivation

Existing example doesn't work when using a dedicated
namespace (line 76)

### More

- [ ] Added/updated tests


jare:
bad branch

PR #9088, opened 2022-06-10
Filter which action to run based on a path inclusion/exclusion

tomMoulard:
### What does this PR do?

This PR filter which GitHub Action to run based on which file was modified, on a PR.
For example, if a file in the `docs` folder was modified, it is only important to run the [`Check Documentation`](https://github.com/traefik/traefik/actions/workflows/check_doc.yml) action.

### Motivation

It is not useful to run actions that are not necessary for a given PR

### More

- ~[ ] Added/updated tests~

### Additional comments

Feel free to move this PR's targeted branch.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,6 +4,10 @@ on:
   pull_request:
     branches:
       - '*'
+    paths-ignore:
+      - 'docs/**'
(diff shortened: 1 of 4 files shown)


ldez:
This doesn't work as expected.

PR #9100, opened 2022-06-14
Add back defaultEntryPoints setting (for HTTP/TCP)

### LINKED ISSUE #6984: defaultEntryPoints in v2
Feature

An ability to set default entry points for the services which does not specify it. This is mainly a security feature which gives an administrator a way to reliably divide services into types (public, private, etc.). So that private service will not be published to the world accidentally without specifying entry point. Or event stricter - service will not be published (available) at all if no entry point is specified for it and `defaultEntryPoints` is empty. Of course saving current behavior like `defaultEntryPoints = *`. This could also be divided into default http/tcp/udp entry points if this make sense. 

This was suggested earlier in https://github.com/containous/traefik/issues/5672 but the reason of the decision is quite confusing, since it is closed now, writing here:
what is the problem in filtering by entry point across all possible variants (current) compared to filtering across a part of them? Incorrect configuration of default/service entry points might lead to unexpected results I guess, but shouldn't it be a problem of an administrator (correct and wide documentation) rather than a software restriction? The same time this would be nice to have a security foolproof since there are environments where static (main) configuration and services are controlled by different people/departments.

Thanks.

jilleJr:
### What does this PR do?

- Add defaultEntryPoints setting, like the setting Traefik v1 had. The defaultEntryPoints has no effect on UDP routers. To keep it simple, I only implemented the HTTP/TCP part, and leaving the UDP to a future enhancement when it will eventually be requested.

### Motivation

Closes #6984

We also needed this at my current company, where we want to change the default behavior when using Traefik to use the Kubernetes Ingress as routers. For security reasons, we want it to by default not expose the apps the Ingress objects target to all entry points, in case someone forgets to add the correct annotation.

This enables some entrypoints to be opt-in instead of opt-out.

### More

- [x] Added/updated tests

### Additional Notes

@dduportal originally wrote this concern:

> The `defaultEntryPoints` directive had a meaning in v1 when TLS options where set per entrypoint.
> With v2, TLS being defined at router level + ability to route at TCP level, the default entrypoint does not make sense as it could break the TCP router, or accidentally mux HTTP/TCP.
> 
> This is why we're going to close this issue as declined.
>
> https://github.com/traefik/traefik/issues/5672#issuecomment-542775022

Does this concern still apply? Specifically with this implementation?

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -372,6 +372,12 @@ func getDefaultsEntrypoints(staticConfiguration *static.Configuration) []string 
 			log.WithoutContext().Errorf("Invalid protocol: %v", err)
 		}
 
+		if len(staticConfiguration.DefaultEntryPoints) > 0 {
+			if !containsStr(staticConfiguration.DefaultEntryPoints, name) {
+				continue
+			}
+		}
+
 		if protocol != "udp" && name != static.DefaultInternalEntryPointName {
 			defaultEntryPoints = append(defaultEntryPoints, name)
 		}
@@ -379,6 +385,15 @@ func getDefaultsEntrypoints(staticConfiguration *static.Configuration) []string 
 
 	sort.Strings(defaultEntryPoints)
 	return defaultEntryPoints
+}
+
+func containsStr(slice []string, str string) bool {
+	for _, v := range slice {
+		if v == str {
+			return true
+		}
+	}
+	return false
 }
 
 func switchRouter(routerFactory *server.RouterFactory, serverEntryPointsTCP server.TCPEntryPoints, serverEntryPointsUDP server.UDPEntryPoints, aviator *pilot.Pilot) func(conf dynamic.Configuration) {
--- a/docs/content/reference/static-configuration/cli-ref.md
+++ b/docs/content/reference/static-configuration/cli-ref.md
@@ -99,6 +99,9 @@ Storage to use. (Default: ```acme.json```)
 `--certificatesresolvers.<name>.acme.tlschallenge`:  
 Activate TLS-ALPN-01 Challenge. (Default: ```true```)
 
+`--defaultentrypoints`:  
+Default HTTP/TCP entry point names used when none are specified in a router. (Does not apply to UDP entry points)
+
 `--entrypoints.<name>`:  
 Entry points definition. (Default: ```false```)
 
--- a/docs/content/reference/static-configuration/env-ref.md
+++ b/docs/content/reference/static-configuration/env-ref.md
@@ -99,6 +99,9 @@ Storage to use. (Default: ```acme.json```)
 `TRAEFIK_CERTIFICATESRESOLVERS_<NAME>_ACME_TLSCHALLENGE`:  
 Activate TLS-ALPN-01 Challenge. (Default: ```true```)
 
+`TRAEFIK_DEFAULTENTRYPOINTS`:  
+Default HTTP/TCP entry point names used when none are specified in a router. (Does not apply to UDP entry points)
+
 `TRAEFIK_ENTRYPOINTS_<NAME>`:  
 Entry points definition. (Default: ```false```)
 
--- a/docs/content/routing/entrypoints.md
+++ b/docs/content/routing/entrypoints.md
@@ -102,6 +102,7 @@ They can be defined by using a file (YAML or TOML) or CLI arguments.
 
     ```yaml tab="File (YAML)"
     ## Static configuration
+    defaultEntryPoints = [name1, name2, name3]
     entryPoints:
       name:
         address: ":8888" # same as ":8888/tcp"
@@ -131,6 +132,7 @@ They can be defined by using a file (YAML or TOML) or CLI arguments.
 
     ```toml tab="File (TOML)"
     ## Static configuration
+    defaultEntryPoints = ["name1", "name2", "name3"]
     [entryPoints]
       [entryPoints.name]
         address = ":8888" # same as ":8888/tcp"
@@ -156,6 +158,7 @@ They can be defined by using a file (YAML or TOML) or CLI arguments.
 
     ```bash tab="CLI"
     ## Static configuration
+    --defaultEntryPoints=name1,name2,name3
     --entryPoints.name.address=:8888 # same as :8888/tcp
     --entryPoints.name.http2.maxConcurrentStreams=42
     --entryPoints.name.http3.advertisedport=8888
@@ -169,6 +172,80 @@ They can be defined by using a file (YAML or TOML) or CLI arguments.
     --entryPoints.name.forwardedHeaders.insecure=true
     --entryPoints.name.forwardedHeaders.trustedIPs=127.0.0.1,192.168.0.1
     ```
+
+### Default EntryPoints
+
+Each Traefik [route](./routers/index.md) has a list of entry points associated
+with it. When this list is empty or not defined, Traefik will use the default
+entry points.
+
+By default, all entry points are used as the default entry points, with the
+exception of some special entry points of names: `traefik`,
+[`traefikhub-api`, and `traefikhub-tunl`](../traefik-hub/index.md#entrypoints).
+Specifying any of these special entry point names in the configuration has no
+effect, they will always be excluded from the default entry point behavior.
+
+The list of default entry points can be overridden for TCP & HTTP entry points
+via the static configuration.
+
+??? example "Overriding default entry points"
+
+    ```yaml tab="File (YAML)"
+    ## Static configuration
+    defaultEntryPoints = [name1, name3]
+    # by default, defaultEntryPoints defaults to all entry points, e.g [name1, name2, name3]
+    
+    entryPoints:
+      name1:
+        address: ":8081"
+      name2:
+        address: ":8082"
+      name3:
+        address: ":8083"
+    ```
+
+    ```toml tab="File (TOML)"
+    ## Static configuration
+    defaultEntryPoints = ["name1", "name3"]
+    # by default, defaultEntryPoints defaults to all entry points, e.g ["name1", "name2", "name3"]
+    
+    [entryPoints]
+      [entryPoints.name1]
+        address = ":8081"
+      [entryPoints.name2]
+        address = ":8082"
+      [entryPoints.name3]
+        address = ":8083"
+    ```
+
+    ```bash tab="CLI"
+    ## Static configuration
+    --defaultEntryPoints=name1,name3
+    # by default, defaultEntryPoints defaults to all entry points, e.g name1,name2,name3
+    
+    --entryPoints.name1.address=:8081
+    --entryPoints.name2.address=:8082
+    --entryPoints.name3.address=:8083
+    ```
+
+You cannot disable "default entry points". If you specify no entry points as
+default entry points via the configuration, then the fallback behavior of using
+all entry points will be used instead.
+
+!!! warning "Does not apply for UDP entry points"
+
+    The `defaultEntryPoints` config only applies to HTTP/TCP entry points and not
+    to UDP entry points (yet).
+
+    Routes will always use all UDP entry points when no entry points are
+    explicitly specified.
+
+!!! info "The `defaultEntryPoints` feature was re-introduced in Traefik v2.8.0"
+
+    Traefik v1 had support for specifying default entry points since v1.0.0, but
+    after Traefik v2's overhaul of the code base, this feature has been
+    postponed. All versions from Traefik v2.0.0 up until Traefik v2.8.0 are
+    missing this config feature.
 
 ### Address
 
--- a/docs/content/routing/routers/index.md
+++ b/docs/content/routing/routers/index.md
@@ -94,7 +94,7 @@ or act before forwarding the request to the service.
 
 ### EntryPoints
 
-If not specified, HTTP routers will accept requests from all defined entry points.
+If not specified, HTTP routers will accept requests from the defined default entry points.
 If you want to limit the router scope to a set of entry points, set the `entryPoints` option.
 
 ??? example "Listens to Every EntryPoint"
@@ -106,7 +106,7 @@ If you want to limit the router scope to a set of entry points, set the `entryPo
     http:
       routers:
         Router-1:
-          # By default, routers listen to every entry points
+          # By default, routers listen to static config of default entry points
           rule: "Host(`example.com`)"
           service: "service-1"
     ```
@@ -115,7 +115,7 @@ If you want to limit the router scope to a set of entry points, set the `entryPo
     ## Dynamic configuration
     [http.routers]
       [http.routers.Router-1]
-        # By default, routers listen to every entry points
+        # By default, routers listen to static config of default entry points
         rule = "Host(`example.com`)"
         service = "service-1"
     ```
@@ -124,6 +124,7 @@ If you want to limit the router scope to a set of entry points, set the `entryPo
 
     ```yaml tab="File (YAML)"
     ## Static configuration
+    #defaultEntryPoints: [web, websecure, other] # By default, every entry point is used
     entryPoints:
       web:
         address: ":80"
@@ -135,6 +136,7 @@ If you want to limit the router scope to a set of entry points, set the `entryPo
 
     ```toml tab="File (TOML)"
     ## Static configuration
+    #defaultEntryPoints: ["web", "websecure", "other"] # By default, every entry point is used
     [entryPoints]
       [entryPoints.web]
         address = ":80"
@@ -146,6 +148,7 @@ If you want to limit the router scope to a set of entry points, set the `entryPo
 
     ```bash tab="CLI"
     ## Static configuration
+    #--defaultentrypoints=web,websecure,other # By default, every entry point is used
     --entrypoints.web.address=:80
     --entrypoints.websecure.address=:443
     --entrypoints.other.address=:9090
@@ -182,6 +185,7 @@ If you want to limit the router scope to a set of entry points, set the `entryPo
 
     ```yaml tab="File (YAML)"
     ## Static configuration
+    #defaultEntryPoints = [web, websecure, other] # Unused when entry points are specified for a route
     entryPoints:
       web:
         address: ":80"
@@ -193,6 +197,7 @@ If you want to limit the router scope to a set of entry points, set the `entryPo
 
     ```toml tab="File (TOML)"
     ## Static configuration
+    #defaultEntryPoints = ["web", "websecure", "other"] # Unused when entry points are specified for a route
     [entryPoints]
       [entryPoints.web]
         address = ":80"
@@ -204,6 +209,7 @@ If you want to limit the router scope to a set of entry points, set the `entryPo
 
     ```bash tab="CLI"
     ## Static configuration
+    #--defaultentrypoints=web,websecure,other # Unused when entry points are specified for a route
     --entrypoints.web.address=:80
     --entrypoints.websecure.address=:443
     --entrypoints.other.address=:9090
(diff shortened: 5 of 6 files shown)


rtribotte:
Hello @jilleJr,

Thanks for your contribution,

> Does this concern still apply? Specifically with this implementation?

I'm not sure it still applies. With the [entrypoints HTTP options](https://doc.traefik.io/traefik/routing/entrypoints/#http-options) the TLS options can be defined at the EntryPoints level.

I passed your pull request in `needs-review` state and I pushed a commit to move the option to the EntryPoint resource level, which is, after internal discussions, the approach we preferred.

cmd/traefik/traefik.go
```diff
@@ -360,7 +360,23 @@ func getHTTPChallengeHandler(acmeProviders []*acme.Provider, httpChallengeProvid
...
+			break
+		}
+	}
+
 	for name, cfg := range staticConfiguration.EntryPoints {
+		// Ignores EntryPoints that are marked as not part of default set,
+		// if at least one is configured to be in the default set.
+		if userDefinedDefaults && !cfg.DefaultSet {
+			continue
+		}
```
jilleJr, line 378: Do you maybe want to invert this, so it by default is included in the default set?

I think the additional logic of "if none are defined as default, then all are" is a weird semantic.

Just a suggestion, but I think the config could be changed to e.g:

- `--entrypoints.<name>.skipAsDefault`
- `--entrypoints.<name>.skipDefaultSet`
- `--entrypoints.<name>.noDefaultSet`
- `--entrypoints.<name>.optOutAsDefault`

or something like that. Then have the setting default to `false`, as now, but then you don't need this extra logic

jilleJr: > Indeed we choose to use an opt-in fashion because we thought it was better once you started to decide which EntryPoint should be in the default set, since it prevents having an EntryPoint in the default set only because you forgot to opt-out.

Yea that makes a lot more sense, especially since this feature was requested from a security perspective to begin with.

> We were worried that a `DefaultEntrypoints` option could be seen as an option to choose what EntryPoint should be set up automatically.

Brilliant foresight!

And yeah having the config inside the EntryPoint model is probably a lot more future-proof. Good call

rtribotte (approved):
Thanks 👍

cmd/traefik/traefik.go
```diff
@@ -360,7 +360,23 @@ func getHTTPChallengeHandler(acmeProviders []*acme.Provider, httpChallengeProvid
 
 func getDefaultsEntrypoints(staticConfiguration *static.Configuration) []string {
 	var defaultEntryPoints []string
+
+	// Determines if at least one EntryPoint is configured to be in the default set.
+	var userDefinedDefaults bool
```
ddtmachado, line 365:
```suggestion
	var hasDefinedDefaults bool
```

cmd/traefik/traefik.go
```diff
@@ -360,7 +360,23 @@ func getHTTPChallengeHandler(acmeProviders []*acme.Provider, httpChallengeProvid
...
+	for _, ep := range staticConfiguration.EntryPoints {
+		if ep.DefaultSet {
+			userDefinedDefaults = true
```
ddtmachado, line 368:
```suggestion
			hasDefinedDefaults = true
```

cmd/traefik/traefik.go
```diff
@@ -360,7 +360,23 @@ func getHTTPChallengeHandler(acmeProviders []*acme.Provider, httpChallengeProvid
...
+			break
+		}
+	}
+
 	for name, cfg := range staticConfiguration.EntryPoints {
+		// Ignores EntryPoints that are marked as not part of default set,
+		// if at least one is configured to be in the default set.
+		if userDefinedDefaults && !cfg.DefaultSet {
```
ddtmachado, line 376:
```suggestion
		if hasDefinedDefaults && !cfg.DefaultSet {
```

cmd/traefik/traefik.go
```diff
@@ -360,7 +360,23 @@ func getHTTPChallengeHandler(acmeProviders []*acme.Provider, httpChallengeProvid
...
+	for _, ep := range staticConfiguration.EntryPoints {
+		if ep.DefaultSet {
+			userDefinedDefaults = true
+			break
+		}
+	}
+
 	for name, cfg := range staticConfiguration.EntryPoints {
+		// Ignores EntryPoints that are marked as not part of default set,
+		// if at least one is configured to be in the default set.
```
ddtmachado, line 375:
```suggestion
		// By default, if no entrypoints exist on the default set, all entrypoints are considered.
		// If at least one is included on the default set, then all entrypoints are ignored unless flagged.
```
not sure on the actual wording here but I'm trying to make it clearer, as the original comment wasn't for me :)

docs/content/reference/static-configuration/cli-ref.md
```diff
@@ -105,6 +105,9 @@ Entry points definition. (Default: ```false```)
 `--entrypoints.<name>.address`:  
 Entry point address.
 
+`--entrypoints.<name>.defaultset`:  
+Defines whether to add this EntryPoint to the default set. (Default: ```false```)
```
ddtmachado, line 109:
```suggestion
Flags this EntryPoint to be used by default when placing a router configuration that does not specify one. (Default: ```false```)
```

docs/content/reference/static-configuration/env-ref.md
```diff
@@ -105,6 +105,9 @@ Entry points definition. (Default: ```false```)
 `TRAEFIK_ENTRYPOINTS_<NAME>_ADDRESS`:  
 Entry point address.
 
+`TRAEFIK_ENTRYPOINTS_<NAME>_DEFAULTSET`:  
+Defines whether to add this EntryPoint to the default set. (Default: ```false```)
```
ddtmachado, line 109:
```suggestion
Flags this EntryPoint to be used by default when placing a router configuration that does not specify one. (Default: ```false```)
```

docs/content/routing/entrypoints.md
```diff
@@ -233,6 +233,55 @@ If both TCP and UDP are wanted for the same port, two entryPoints definitions ar
 
     Full details for how to specify `address` can be found in [net.Listen](https://golang.org/pkg/net/#Listen) (and [net.Dial](https://golang.org/pkg/net/#Dial)) of the doc for go.
 
+### DefaultSet
+
+_Optional, Default=false_
+
+The `DefaultSet` option defines whether the EntryPoint should be included in the EntryPoints default set.
```
ddtmachado, line 240:
```suggestion
The `DefaultSet` option flags the EntryPoint to be used by default when configuring HTTP and TCP routers that do not define the [EntryPoints option](./routers/index.md#entrypoints).
```

docs/content/routing/entrypoints.md
```diff
@@ -233,6 +233,55 @@ If both TCP and UDP are wanted for the same port, two entryPoints definitions ar
...
+
+The EntryPoints default set is used to configure HTTP and TCP routers which are not defining the [EntryPoints option](./routers/index.md#entrypoints).
```
ddtmachado, line 242:
```suggestion
```

docs/content/routing/entrypoints.md
```diff
@@ -233,6 +233,55 @@ If both TCP and UDP are wanted for the same port, two entryPoints definitions ar
...
+
+!!! info "How the EntryPoints default set is built"
+
+    If no EntryPoint is explicitly included in the EntryPoints default set, 
+    meaning that all EntryPoints have the `DefaultSet` option set to `false`,
+    the EntryPoints default set includes all HTTP/TCP EntryPoints.
```
ddtmachado, line 248:
```suggestion
    then it includes all HTTP/TCP EntryPoints.
```

docs/content/routing/entrypoints.md
```diff
@@ -233,6 +233,55 @@ If both TCP and UDP are wanted for the same port, two entryPoints definitions ar
...
+
+    If at least one EntryPoint is explicitly included in the EntryPoints default set,
+    meaning that at least one EntryPoint have the `DefaultSet` option set to `true`,
+    the EntryPoints default set includes only EntryPoints that have the `DefaultSet` option set to `true`.
```
ddtmachado, line 252:
```suggestion
    then only EntryPoints that have the `DefaultSet` option set to `true` are included.
```

docs/content/routing/routers/index.md
```diff
@@ -94,8 +94,8 @@ or act before forwarding the request to the service.
 
 ### EntryPoints
 
-If not specified, HTTP routers will accept requests from all defined entry points.
-If you want to limit the router scope to a set of entry points, set the `entryPoints` option.
+If not specified, HTTP routers will accept requests from every EntryPoints in the default set.
+If you want to limit the router scope to a set of EntryPoints, set the `entryPoints` option.
```
ddtmachado, line 98:
```suggestion
If you want to limit the router scope to specific EntryPoints, set the `entryPoints` option.
```

ddtmachado (commented):
Looks great, some nitpicks on docs and comments

cmd/traefik/traefik.go
jilleJr: What about?

```suggestion
		// By default all entrypoints are considered.
		// If at least one is flagged, then only flagged entrypoints are included.
```

ddtmachado (approved):

jilleJr:
The tests passes locally, but fails in the CI pipeline.

The failing test should definitely not be affected by this PR:

```
--- FAIL: TestHTTP3AdvertisedPort (0.00s)
    server_entrypoint_tcp_http3_test.go:126: 
        	Error Trace:	server_entrypoint_tcp_http3_test.go:126
        	Error:      	"" does not contain ":8080"
        	Test:       	TestHTTP3AdvertisedPort
2022/07/08 07:11:41 failed to sufficiently increase receive buffer size (was: 208 kiB, wanted: 2048 kiB, got: 416 kiB). See https://github.com/lucas-clemente/quic-go/wiki/UDP-Receive-Buffer-Size for details.
time="2022-07-08T07:11:41Z" level=error msg="accept tcp 127.0.0.1:34437: use of closed network connection"
time="2022-07-08T07:11:41Z" level=error msg="Error while starting server: accept tcp 127.0.0.1:34437: use of closed network connection"
time="2022-07-08T07:11:42Z" level=error msg="accept tcp 127.0.0.1:42871: use of closed network connection"
time="2022-07-08T07:11:42Z" level=error msg="Error while starting server: accept tcp 127.0.0.1:42871: use of closed network connection"
time="2022-07-08T07:11:42Z" level=error msg="accept tcp 127.0.0.1:36971: use of closed network connection"
time="2022-07-08T07:11:42Z" level=error msg="Error while starting server: accept tcp 127.0.0.1:36971: use of closed network connection"
time="2022-07-08T07:11:44Z" level=error msg="Error while Peeking first byte: read tcp [::1]:43767->[::1]:[410](https://github.com/traefik/traefik/runs/7247077064?check_suite_focus=true#step:6:411)62: i/o timeout"
FAIL
coverage: 63.2% of statements
FAIL	github.com/traefik/traefik/v2/pkg/server	10.262s
```

(from https://github.com/traefik/traefik/runs/7247077064?check_suite_focus=true)

ddtmachado:
@jilleJr we had an internal discussion about the flag name because `defaultset` might feel too technical. We didn't had any good alternatives so far but it occurred to me that `autoAttach` or even `autoAttachRouters` could work, wdyt?
Let us know if you have any other suggestion in mind

jilleJr:
> we had an internal discussion about the flag name because `defaultset` might feel too technical. We didn't had any good alternatives so far but it occurred to me that `autoAttach` or even `autoAttachRouters` could work, wdyt? Let us know if you have any other suggestion in mind

Hello @ddtmachado!

I've a bunch of naming ideas, such as:

- `--entrypoints.<name>.includeAsFallback`

I'm still more keen in inverting the flag. Now the `defaultset` flag has this weird condition where it behaves like the default value is true if all `defaultset` is set to false, otherwise the default value is false. If we invert the semantics of the flag then we could get rid of this additional condition.

That also works better conceptually with the entrypoints that are hardcoded to be excluded (`traefik` and `traefikhub-api`/`-tunl`), as the semantics then correctly suggest that the user can't override them to manually enable those entrypoints to be used by default. We could also then add to the documentation that `traefik`, `traefikhub-api`/`-tunl` has this flag set to `true`, to tie it all together.

To build on your idea of "auto attach", we could then have for example:

- `--entrypoints.<name>.noAutoAttach`
- `--entrypoints.<name>.skipAutoAttach`

Some other inverted flag name suggestions:

- `--entrypoints.<name>.hideByDefault`
- `--entrypoints.<name>.skipAsDefault`
- `--entrypoints.<name>.unusedByDefault`
- `--entrypoints.<name>.disableAsFallback`
- `--entrypoints.<name>.excludeAsFallback`
- `--entrypoints.<name>.explicitUseOnly`
- `--entrypoints.<name>.requireExplicitUse`

---

My personal favorite is:

- `--entrypoints.<name>.explicitUseOnly`

jilleJr:
Could also take some inspiration from the existing configs.

- `--accesslog.fields.defaultmode`:

  - `--entrypoints.<name>.defaultMode`: keep | drop (default: `keep`)

- `--providers.<type>.exposedByDefault`:

  - `--entrypoints.<name>.exposedByDefault`: true | false (default: `true`)

jilleJr:
I've now changed to `--entrypoints.<name>.explicitUseOnly`. If you disagree then it's just a matter of reverting 035cd6f

jilleJr:
@ddtmachado Hello! I understand this might be low prio, but is there something I can do to speed up this process of getting this merged and released? Does this have to be re-reviewed again?

ddtmachado:
@jilleJr after internal discussions we decided to postpone merging this PR until we release the next minor version, targeted for the end of this month. In the meantime we'll have time to decide on the naming **and** default behavior as the way you proposed on the last update is a breaking change that would not be possible in a minor release. We'll keep you posted if the situation changes.

ddtmachado:
hello again @jilleJr , with the 2.9 RC out we're ready to get back to this topic. As said in my previous message we took some time to think about this and we really want to avoid breaking changes when possible, which means we would like to stick with the previous behaviour:

* by default every entrypoint is considered to be included (ignore no config)
* once one entrypoint is explicitly flagged to then all entrypoints are excluded unless they explicity set the flag to true

We believe that this approach maintains backward compatibility while still allowing for stricter configuration.

As for the team thinks a simple `asDefault` would work. Please let us know your opinion and if you're willing to proceed with those changes.

jilleJr:
@ddtmachado Heyo glad to hear it was not forgotten. I started to worry there for a second :)

Personally I'm voting for the `explicitUseOnly`-styled config, but my voice is also one new contributor vs a team of experienced Traefik maintainers, so I'll just change it to `asDefault` as you propose.

rtribotte (approved):
Thanks 👍

pkg/config/static/entrypoints.go
```diff
@@ -12,6 +12,7 @@ import (
 // EntryPoint holds the entry point configuration.
 type EntryPoint struct {
 	Address          string                `description:"Entry point address." json:"address,omitempty" toml:"address,omitempty" yaml:"address,omitempty"`
+	AsDefault        bool                  `description:"Defines whether to attach routers, that don't have any Entrypoint defined, to this Entrypoint." json:"asDefault,omitempty" toml:"asDefault,omitempty" yaml:"asDefault,omitempty"`
```
tomMoulard, line 15:
```suggestion
	AsDefault        bool                  `description:"Flag Entrypoint to be use by default on routers that don't have any Entrypoint defined." json:"asDefault,omitempty" toml:"asDefault,omitempty" yaml:"asDefault,omitempty"`
```

jilleJr: Sure!

docs/content/reference/static-configuration/cli-ref.md
```diff
@@ -105,6 +105,9 @@ Entry points definition. (Default: ```false```)
 `--entrypoints.<name>.address`:  
 Entry point address.
 
+`--entrypoints.<name>.asdefault`:  
+Flag Entrypoint to be use by default on routers that don't have any Entrypoint defined. (Default: ```false```)
```
mpl, line 109:
```suggestion
Asdefault adds the entrypoint to the list of entrypoints that are used by default on routers that don't have any Entrypoint defined. (Default: ```false```)
```

docs/content/routing/entrypoints.md
```diff
@@ -233,6 +233,54 @@ If both TCP and UDP are wanted for the same port, two entryPoints definitions ar
 
     Full details for how to specify `address` can be found in [net.Listen](https://golang.org/pkg/net/#Listen) (and [net.Dial](https://golang.org/pkg/net/#Dial)) of the doc for go.
 
+### AsDefault
+
+_Optional, Default=false_
+
+The `AsDefault` option flags the EntryPoint to be in the list of default EntryPoints.
```
mpl, line 240:
```suggestion
The `AsDefault` option marks the EntryPoint to be in the list of default EntryPoints.
```

docs/content/routing/entrypoints.md
```diff
@@ -233,6 +233,54 @@ If both TCP and UDP are wanted for the same port, two entryPoints definitions ar
...
+EntryPoints in this list are used (by default) on HTTP and TCP routers that do not define the [EntryPoints option](./routers/index.md#entrypoints).
```
mpl, line 241:
```suggestion
EntryPoints in this list are used (by default) on HTTP and TCP routers that do not define their own [EntryPoints option](./routers/index.md#entrypoints).
```

docs/content/routing/entrypoints.md
```diff
@@ -233,6 +233,54 @@ If both TCP and UDP are wanted for the same port, two entryPoints definitions ar
...
+
+!!! info "List of default EntryPoints"
+
+    If there is no EntryPoint with the `AsDefault` option set to `true`, 
+    then the list of default EntryPoints includes all HTTP/TCP EntryPoints.
+
+    If at least one EntryPoint have the `AsDefault` option set to `true`,
```
mpl, line 248:
```suggestion
    If at least one EntryPoint has the `AsDefault` option set to `true`,
```

docs/content/routing/entrypoints.md
```diff
@@ -233,6 +233,54 @@ If both TCP and UDP are wanted for the same port, two entryPoints definitions ar
...
+    then the list of default EntryPoints includes only EntryPoints that have the `AsDefault` option set to `true`.
+
+    Some built-in EntryPoints are always excluded from the list, namely: `traefik`, `traefikhub-api`, and `traefikhub-tunl`.
+
+!!! warning "Only TCP and HTTP"
+
+    The `AsDefault` option has no effect on UDP EntryPoints.
+    When UDP routers are not defining the [EntryPoints option](./routers/index.md#entrypoints_2),
+    they are attached to all available UDP EntryPoints.
```
mpl, line 257:
```suggestion
    When a UDP router does not define the [EntryPoints option](./routers/index.md#entrypoints_2),
    it is attached to all available UDP EntryPoints.
```

mpl (approved):

brandond:
I see that this is still only on master, is this going to be pulled into 2.9 at any point?

ldez:
We follow semver so that when a minor is created, no new features are added.
Features merge on master stay on master and they will be a part of the next major/minor release of Traefik.

https://github.com/traefik/traefik/tree/master#release-cycle

brandond:
Are there images built from the master branch available anywhere?

ldez:
The tag `experimental-master` is updated after each merge into master (`traefik/traefik:experimental-master`)

https://hub.docker.com/r/traefik/traefik/tags

brandond:
Fantastic, i didn't know about that repo - I was just looking at library/traefik.

ekle:
Is there any stable release which contains this feature?

applejag:
@ekle sadly no. I'm still waiting for v3 to drop.

It has been available since [v3.0.0-beta1](https://github.com/traefik/traefik/releases/tag/v3.0.0-beta1), released Dec 5, 2022.

Don't know how far away v3.0.0 is from being released. The [3.0 milestone](https://github.com/traefik/traefik/milestone/20) is currently at 86% complete 21 issues still open, but I don't think it's possible to extract some kind of ETA from that.

PR #9104, opened 2022-06-16
Add log when missing path in health check

moutoum:
### What does this PR do?

This PR adds a log to help users understand their health check configuration is missing the `path` argument.

### Motivation

When the health check does not have a path, the configuration is omitted and not effective. There is no indication and everything looks like working properly.

### More

~~- [ ] Added/updated tests~~


kevinpollet (approved):
Thanks 👍

rtribotte (approved):
LGTM 👌

Issue #9110, opened 2022-06-18
HTTP file provider does not poll at start up


fcwoknhenuxdfiyv:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

With a configuration similar to this:

```
[providers.http]
  endpoint = "https://api.thingy/traefik"
  pollInterval = "300s"
```

I have to wait 5 minutes before Traefik becomes usable. It should fire at start up.

### What did you see instead?

I have to wait `pollInterval` seconds for Traefik to serve TLS sites/services.

### What version of Traefik are you using?

```
Version:      2.7.1
Codename:     epoisses
Go version:   go1.17.11
Built:        2022-06-13T13:35:10Z
OS/Arch:      linux/amd64
```

### What is your environment & configuration?

```yaml
# (paste your configuration here)
```

Add more configuration information here.

traefiker:
Hi! I'm Træfiker :robot: the bot in charge of communication regulation.

Thanks for your interest in Traefik!

We dedicate the issue tracker to bug reports and feature requests only. My advanced AI has spotted that your issue might be a configuration problem or relates to something that doesn't look like a bug.

To confirm this, please join our [Community Forum](https://community.containo.us/) and reach out to us on the [Traefik section](https://community.containo.us/c/traefik).

In case I'm wrong (well, that would be embarrassing :sweat_smile:), my developers will re-open the issue and fix me!

In the meantime, you can double check [Traefik's documentation](https://docs.traefik.io/).

PR #9117, opened 2022-06-21
Poll http provider at start up

fcwoknhenuxdfiyv:
### What does this PR do?

This fixes issue #9110 by polling the http proivider at start up.

### Motivation

We do not need to poll our configuarion API very frequently as our configuration is reasonably static. However, the configuration is not applied at start up which means our services need 5 minutes to apply their certificates.

### More

- [x] N/A Added/updated tests
- [x] N/A Added/updated documentation

### DIFF
--- a/pkg/provider/http/http.go
+++ b/pkg/provider/http/http.go
@@ -66,6 +66,40 @@ func (p *Provider) Init() error {
 	return nil
 }
 
+// run performs call to configuration endpoint.
+func (p *Provider) run(configurationChan chan<- dynamic.Message) error {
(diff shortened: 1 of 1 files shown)


rtribotte:
Hello @fcwoknhenuxdfiyv,

Thanks for your contribution!

That is unfortunate, but we opened a PR to fix the issue almost at the same time: https://github.com/traefik/traefik/pull/9116
We can close yours and add you as co-author to the one we opened, WDYT?

fcwoknhenuxdfiyv:
Sounds good to me :-)

PR #9131, opened 2022-06-27
Fix muxer

juliens:
### What does this PR do?
This PR fixes an issue with the `HostRegexp` matcher and the `Query` matcher.

Before the fix,`HostRegexp` used an empty host if the url path was http://@/ whereas the `Host` matcher used the host header, `HostRegexp` now uses the host header in case the url contains an empty host.

Before the fix `Query` considered semi-colons as invalid and so ignored all the query params, Query now handles `;` as separator and sanitizes those `;` with `&` when the request is forwarded to the backend.

### Motivation
Fix some weird behaviour with negative matchers.

### More

- [x] Added/updated tests


ldez (approved):
LGTM

PR #9140, opened 2022-06-29
Move consulcatalog provider to only use health apis

kevinpollet:
### What does this PR do?

This PR drops the usage of the Catalog API within the Consul Catalog provider and switches to only using the Health API(s).  This is nicely compatible with the Connect endpoints as they live under the same API and output exactly the same types.

### Motivation

The Health API, with the cache query parameter set to `true`, uses the streaming health check system. This means that if streaming is enabled in a Consul cluster, Traefik is now effectively 'free' to run, going so far as an operator can set the refresh interval down to 1 second and likely not notice anything in their Consul cluster.

Even if streaming is disabled, this still cuts the request rate in half, meaning $\approx\frac{1}{2}$ bandwidth usage on top of that which is helpful in really large clusters.

Supersedes #8834

### More

- [ ] Added/updated tests

### Additional Notes

Co-authored-by: Charles Zaffery <czaffery@roblox.com>

### DIFF
--- a/pkg/provider/consulcatalog/consul_catalog.go
+++ b/pkg/provider/consulcatalog/consul_catalog.go
@@ -316,31 +316,31 @@ func (p *Provider) getConsulServicesData(ctx context.Context) ([]itemData, error
 		}
 
 		for _, consulService := range consulServices {
-			address := consulService.ServiceAddress
+			address := consulService.Service.Address
 			if address == "" {
-				address = consulService.Address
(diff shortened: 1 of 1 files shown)


kevinpollet (commented):
Thanks @chuckyz 👍

tomMoulard (approved):
LGTM :ok_hand:

ldez (approved):
LGTM

PR #9151, opened 2022-06-30
integration: use VPN for integration tests (for Mac)

### LINKED ISSUE #8962: To improve test suites
### Welcome!

There currently are several problems with the test suites, and in particular with the integration tests.

- some of them are intrinsically flaky. We should investigate which ones and mitigate as much as possible.

- it used to be possible to run at least some of them without docker. Now, even though we can do the preparation steps (e.g. building traefik) out of docker, the (integration) tests themselves all run in docker. Consequently:
	- there is a non-negligible performance hit, especially on Mac. This is notably a problem when iterating on a test or a bugfix, when frequent reruns would be helpful, but are greatly hindered by the setup time of the test suite(s) in docker. 
	- it adds an additional barrier to the debugging flow. For example, if one wanted to add breakpoints to one of the tests (arguably an unlikely scenario), one could not. Similarly, if one wanted to replay "by hand" the steps of the test, one would have first to jump on the container in order to do that.

As a result, the main improvement seems to be hingeing on being able to (conditionally) pull out as much as possible from the tests out of docker.

One can see a test as having 3 potential different kinds of actors: the client(s) (i.e. the tests cases themselves), traefik, and the backends (can be any number of exotic things, such as redis, postgres, etc). Which means, one can potentially envision 3 incremental degrees of improvement, where we first pull out the tests, then traefik itself, and finally the backends. Even though the last step does not seem interesting, as one of the compelling points of docker is to avoid having a plethora of third party actors installed on the local dev machine.

In theory, at least for Linux, the two first degrees of improvement should be doable in most cases, aside from some name resolution issues that arose when we upgraded a while ago the libcompose we depend on.

On Mac it is a more complicated story, as the VM from Docker Desktop is an additional barrier of communication between the processes running on the host and the ones running within docker.

mpl:
### What does this PR do?

This change adds a way to setup a VPN (namely Tailscale)
that connects the host to the containers, which allows leaving the third-parties in
containers, and running Traefik and the test client code directly on the host.

### Motivation

Our integration tests run everything by default within docker.

One reason for that (among others) is if the host is a Mac, a program running on
the host can't reach a service running in a container (roughly, because there's
a Linux VM in between the two). Which means we can't ideally have e.g. Traefik
running on the host, the go tests (typically some client code) running on the
host, and the necessary third-parties (e.g. some specific web servers) living in
containers. Which is why both Traefik and the test client code also have to live
in docker by default.

This is not only pretty cumbersome (when iterating on traefik, a docker image
has to be rebuilt everytime one wants to rerun the tests), but it also makes
running the test suites extremely slow.

To address this problem, this change adds a way to setup a VPN (namely Tailscale)
that connects the host to the containers, which allows leaving the third-parties in
containers, and running Traefik and the test client code directly on the host.

### More

- [X] Added/updated tests

### Additional Notes

Fixes #8962 

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -4,8 +4,11 @@ package integration
 import (
 	"bytes"
 	"context"
+	"errors"
 	"flag"
 	"fmt"
+	"io/fs"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -40,8 +43,20 @@ func Test(t *testing.T) {
 		return
 	}
 
+	// TODO(mpl): very niche optimization: do not start tailscale if none of the
+	// wanted tests actually need it (e.g. KeepAliveSuite does not).
+	var vpn *tailscaleNotSuite
+	if os.Getenv("IN_DOCKER") != "true" {
+		vpn = setupVPN(nil, "tailscale.secret")
+		if vpn != nil {
+			defer vpn.TearDownSuite(nil)
+		}
+	}
+
 	check.Suite(&AccessLogSuite{})
-	check.Suite(&AcmeSuite{})
+	if vpn == nil {
+		check.Suite(&AcmeSuite{})
+	}
 	check.Suite(&ConsulCatalogSuite{})
 	check.Suite(&ConsulSuite{})
 	check.Suite(&DockerComposeSuite{})
@@ -55,12 +70,16 @@ func Test(t *testing.T) {
 	check.Suite(&HostResolverSuite{})
 	check.Suite(&HTTPSSuite{})
 	check.Suite(&HTTPSuite{})
-	check.Suite(&K8sSuite{})
+	if vpn == nil {
+		check.Suite(&K8sSuite{})
+	}
 	check.Suite(&KeepAliveSuite{})
 	check.Suite(&LogRotationSuite{})
-	check.Suite(&MarathonSuite15{})
 	check.Suite(&MarathonSuite{})
-	check.Suite(&ProxyProtocolSuite{})
+	check.Suite(&MarathonSuite15{})
+	if vpn == nil {
+		check.Suite(&ProxyProtocolSuite{})
+	}
 	check.Suite(&RateLimitSuite{})
 	check.Suite(&RedisSuite{})
 	check.Suite(&RestSuite{})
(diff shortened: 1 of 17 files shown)


integration/integration_test.go
```diff
@@ -40,8 +43,24 @@ func Test(t *testing.T) {
...
 
+	// TODO(mpl): very niche optimization: do not start tailscale if none of the
+	// wanted tests actually need it (e.g. KeepAliveSuite does not).
+	var (
+		vpn    *tailscaleNotSuite
+		useVPN bool
+	)
+	if os.Getenv("IN_DOCKER") != "true" {
+		vpn = setupVPN(nil, "tailscale.secret")
+		if vpn != nil {
```
kevinpollet, line 54:
```suggestion
		if vpn = setupVPN(nil, "tailscale.secret"); vpn != nil {
```

kevinpollet (approved):

integration/integration_test.go
```diff
@@ -285,3 +325,44 @@ func (s *BaseSuite) getContainerIP(c *check.C, name string) string {
...
+		}
+		return nil
+	}
+	authKey := strings.TrimSpace(string(data))
+	// TODO: copy and create versions that don't need a check.C?
+	vpn := &tailscaleNotSuite{}
+	vpn.createComposeProject(c, "tailscale")
+	vpn.composeUp(c)
+	time.Sleep(5 * time.Second)
+	// TODO(mpl): make sure this docker subnet stays the same as the one we setup in Makefile.
```
mpl, line 365:
```suggestion
	// If we ever change the docker subnet in the Makefile, we need to change this one below correspondingly.
```

integration/integration_test.go
```diff
@@ -285,3 +325,45 @@ func (s *BaseSuite) getContainerIP(c *check.C, name string) string {
...
+
+// tailscaleNotSuite includes a BaseSuite out of convenience, so we can benefit
+// from composeUp et co., but it is not meant to function as a TestSuite per se.
+type tailscaleNotSuite struct{ BaseSuite }
+
+// setupVPN starts tailscale on the corresponding container, and makes it a subnet
+// router, for all the other containers (whoamis, etc) subsequently started for the
+// integration tests.
+// It only does so if the file provided as argument exists, and contains a
+// tailscale auth key (an ephemeral, but reusable, one is recommended).
```
tomMoulard, line 337:
```suggestion
// setupVPN starts Tailscale on the corresponding container, and makes it a subnet
// router, for all the other containers (whoamis, etc) subsequently started for the
// integration tests.
// It only does so if the file(provided as an argument) exists, and contains a
// Tailscale auth key (an ephemeral, but reusable, one is recommended).
```

tomMoulard (approved):
LGTM :ok_hand:

PR #9180, opened 2022-07-12
Refactor certificate domains matching func

rtribotte:
### What does this PR do?

This PR refactors the `MatchDomain` function used to match certificate domains against the server name from the TLS client hello.
It also adds comments regarding the trailing dots for FQDNs and the TLS SNI extension.

### Motivation

While passing by this code portion it appears that the function could be unexported.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/tls/certificate_store.go
+++ b/pkg/tls/certificate_store.go
@@ -73,17 +73,17 @@ func (c *CertificateStore) GetBestCertificate(clientHello *tls.ClientHelloInfo) 
 	if c == nil {
 		return nil
 	}
-	domainToCheck := strings.ToLower(strings.TrimSpace(clientHello.ServerName))
-	if len(domainToCheck) == 0 {
+	serverName := strings.ToLower(strings.TrimSpace(clientHello.ServerName))
+	if len(serverName) == 0 {
 		// If no ServerName is provided, Check for local IP address matches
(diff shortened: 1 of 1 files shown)


kevinpollet (approved):
LGTM

pkg/tls/certificate_store.go
```diff
@@ -121,17 +121,20 @@ func (c CertificateStore) ResetCache() {
 	}
 }
 
-// MatchDomain return true if a domain match the cert domain.
-func MatchDomain(domain, certDomain string) bool {
-	if domain == certDomain {
+// matchDomain returns true if the server name matches the cert domain.
```
mpl, line 124:
```suggestion
// matchDomain returns whether the server name matches the cert domain.
```

pkg/tls/certificate_store.go
```diff
@@ -121,17 +121,20 @@ func (c CertificateStore) ResetCache() {
...
+// The server name, from TLS SNI, can't have trailing dots (https://datatracker.ietf.org/doc/html/rfc6066#section-3).
```
mpl, line 125:
```suggestion
// The server name, from TLS SNI, must not have trailing dots (https://datatracker.ietf.org/doc/html/rfc6066#section-3).
```

mpl (approved):

PR #9186, opened 2022-07-13
Add example of dynamic configuration

pionl:
### What does this PR do?

Helps to understand exactly what dynamic configuration means when working with enabling API / Dashboard.

This cost me few hours to find out, my bad :) Event tutorials are not not "precise". 

Thank you :)

### DIFF
--- a/docs/content/operations/dashboard.md
+++ b/docs/content/operations/dashboard.md
@@ -77,22 +77,6 @@ to allow defining:
 - A [router rule](#dashboard-router-rule) for accessing the dashboard,
   through Traefik itself (sometimes referred as "Traefik-ception").
 
-#### Example
-
-```
-$ cat /etc/traefik/traefik.yml
-api: { }
-
-providers:
-  file:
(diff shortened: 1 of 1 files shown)


mpl:
There are several mentions of the concept on this very page already:

1) the containing section (https://doc.traefik.io/traefik/operations/dashboard/#secure-mode) already mentions the difference between static conf and dynamic conf
2) the "the [router rule](https://doc.traefik.io/traefik/routing/routers/#rule)" part leads to the section about routers, which is part of the dynamic conf
3) the "Dashboard Dynamic Configuration Examples" have the commentary "# Dynamic Configuration" 

So with respect to what already exists, we do not think that this PR is actually necessary. Did you miss these parts, or do you think they were not clear enough, or was there another problem altogether ?

pionl:
Hi, np. 

For me it was not clear enough with the correct meaning of dynamic configuration etc. It is more advance. I think its is better to show "more" real example than just the part name of the property. With that you can later on understand it more clearly.

Have a great day!

Issue #5730, opened 2019-10-26
podman support


anibal-aguila:
Traefik 2.0 Now GA! and want to know if have some plan to support [podman](https://github.com/containers/libpod)

Thanks in advance,

SuperSandro2000:
It probably support podman already but no podman service discovery.

kittydoor:
Is there any interest in supporting podman for service discovery?

phozzy:
I agree with @SuperSandro2000 it is quite easy to run traefik using podman, the question is more about service discovery.
And here comes two variants:
* traefik and a container that traefik will forward requests to run inside one pod. So they share namespaces... is there any opportunity for serice discovery?
* traefik and containers that traefik will forward requests to run in separate containers. What can be used here for service discovery? systemd? I mean probably each container will have appropriate systemd service.

NoX1De:
@kittydoor @phozzy Podman being **fully** supported as a backend is very desirable (including service discovery) in that case I believe both variants mentioned by @phozzy would be desirable.  Supporting podman properly with it being daemonless and also support considering that podman can run in rootless mode would all be desirable.  FWIW I believe podman is going to eventually supersede Docker in desirability for containerized development work as podman really eases the transition for organizations to eventually migrate/import things to orchestration platforms (being that it supports the concept of pods whereas Docker does not).

bkircher:
Another reason why this would be nice to have is that current Docker CE on EL7 is broken somewhat because Red Hat decided to go with nftables in EL8 but docker does not really play nicely with that change (yet).

neingeist:
podman's API: https://www.projectatomic.io/blog/2018/05/podman-varlink/

tomkukral:
Varlink api will be deprecated and docker compatible API is going to land soon. See https://podman.io/blogs/2020/01/17/podman-new-api.html

XakV:
Willing to help test, just following the issue.

DnR-iData:
:+1: To get Podman fully supported as a backend.

Thanks a lot

llabusch93:
Once Traefik will have full service discovery with podman, I say good bye to Docker. Highly interested in it as well, so +1!

dtomcej:
Hello All,

We have discussed this a few times as a team, and have come to the following conclusions:

- Podman is not designed to provide any HA features itself, and it was built to be a drop in replacement for docker in Kubernetes environments

[...]

PR #9190, opened 2022-07-14
Add support for reaching containers using host networking on Podman

freundTech:
### What does this PR do?

This PR adds support for reaching containers using host networking on systems using podman by allowing traefik to correctly detect the podman gateway.

On docker traefik detects the gateway by looking up `host.docker.internal`. Podman doesn't support the `host.docker.internal` domain, but added support for `host.containers.internal` in https://github.com/containers/podman/pull/9972 instead.
Without this PR podman users would have to manually set `host.docker.internal` using add-host/extra_hosts.

### Motivation

I'm using traefik on a server running podman and ran into this issue.

Related to #5730

### More

- [ ] Added/updated tests

### Additional Notes

It would also be possible so solve this using a loop, but because there are only two domains I think that it is more readable without one. Should more domains be needed in the future I would recommend using a loop here.

### DIFF
--- a/docs/content/providers/docker.md
+++ b/docs/content/providers/docker.md
@@ -133,7 +133,7 @@ the IP address of the host is resolved as follows:
 
 <!-- TODO: verify and document the swarm mode case with container.Node.IPAddress coming from the API -->
 - try a lookup of `host.docker.internal`
-- if the lookup was unsuccessfull, try a lookup of `host.containers.internal`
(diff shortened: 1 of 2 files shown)


kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks 👍

mpl (approved):

Issue #6861, opened 2020-05-29
Filtering out ping access logs


MikaelElkiaer:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?
I would like for the option of filtering out calls to /ping in the access log.
My readiness checks in Kubernetes are called very frequently, so having all these calls to /ping in the access log clutters the view.
A possible solution is adding additional filtering of access logs, for example based on RequestPath (e.g.: /ping), RouterName (e.g.: ping@internal), or RequestPort (e.g.: 8080).

I cannot find any issue mentioning this, and only a dead community post: https://community.containo.us/t/is-there-a-way-to-filter-out-ping-requests-in-access-logs-output/4356

notsureifkevin:
Hi Mikael, thank you for the feature request.

One possible solution to this would be to add a regex filter to the `access.log` configuration provider.

(idea stolen from https://docs.nginx.com/nginx/admin-guide/monitoring/logging/#enabling-conditional-logging)

MikaelElkiaer:
> Hi Mikael, thank you for the feature request.
> 
> One possible solution to this would be to add a regex filter to the `accesslog` configuration provider.
> 
> (inspired by: https://docs.nginx.com/nginx/admin-guide/monitoring/logging/#enabling-conditional-logging)

That would do the trick. :)

BulatSaif:
I have the same problem but with prometheus /metrics endpoint, will be nice to filter it out.

yongzhang:
the same request

dhirschfeld:
This would be a very valuable feature. On top of hiding useful information in a flood of irrelevant logs it also dramatically increases the *cost* of monitoring `traefik`. On my AKS cluster, monitoring the `traefik` logs with Azure Monitor is by far my biggest monitoring cost to the point where I will probably have to turn off access logs entirely.

![image](https://user-images.githubusercontent.com/881019/107311678-7e575700-6ada-11eb-9747-5fbb1035ffa7.png)

dhirschfeld:
***xref:*** #6059

fredsig:
[...]

PR #9205, opened 2022-07-23
Add access log filter for excluded URLs

hakman:
### What does this PR do?

This change allows users to add access log filters based on URLs.

### Motivation

This would allow to:
1. not log access to sensitive resources, when privacy is a requirement
2. not log health checks and other annoyances

### More

- [x] Added/updated tests

### Additional Notes

Fixes: #6861

### DIFF
--- a/docs/content/observability/access-logs.md
+++ b/docs/content/observability/access-logs.md
@@ -85,11 +85,14 @@ accessLog:
 
 ### Filtering
 
-To filter logs, you can specify a set of filters which are logically "OR-connected".
-Thus, specifying multiple filters will keep more access logs than specifying only one.
+To filter logs, you can specify a set of filters which are logically "OR-connected",
+except for the `excludedURLsRegex` filter which is logically "AND-connected".
+Thus, specifying multiple filters will keep more access logs than specifying only one,
+given they don't match any of the excluded URLs.
 
 The available filters are:
 
+- `excludedURLsRegex`, to exclude the access logs to requests with a URL matching any of the [regular expressions](https://golang.org/pkg/regexp/) in the specified list
 - `statusCodes`, to limit the access logs to requests with a status codes in the specified range
 - `retryAttempts`, to keep the access logs when at least one retry has happened
 - `minDuration`, to keep access logs when requests take longer than the specified duration (provided in seconds or as a valid duration format, see [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration))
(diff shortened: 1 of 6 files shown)


kevinpollet:
Hello @hakman,
 
hakman:
Thanks for the feedback @kevinpollet. Looking forward to the review. 😄

kevinpollet:
Hello @hakman,

Thanks for your contribution,

We reviewed your pull request, and unfortunately, we are declining it. Adding an `excludedURLsRegex` option will negatively impact the performances, as this `regex` needs to be evaluated for each HTTP request.

We think that a better approach would be to always disable the access logs for the ping router or leverage the log aggregation tool to filter out unwanted entries/URLs.

Dirrk:
@kevinpollet would a static string match also not be acceptable? ie: 
```
excludedPaths:
      - /ping
      - /metrics
```

kevinpollet:
Hi @Dirrk,

This would have an impact on performances (certainly less) but this check will have to be done for each request, and for sure one would want regex at some point. So, we still think that a better approach would be to always disable the access logs for the ping router or leverage the log aggregation tool to filter out unwanted entries/URLs.

obourdon:
Hi @kevinpollet could you please give some hints on how to achieve disabling of access logs for the ping router. I was unable to find anything easily in the docs for that matter. Many thanks

obourdon:
IMHO I am more for letting people have access to this kind of feature with indeed a big warning on performances. I can think of several other products/technologies which had to implement similar requests with very similar impacts but almost all of them agreed to let those in.

obourdon:
@kevinpollet (or anyone else) any comments on what I wrote above please ?

mpl:
> Hi @kevinpollet could you please give some hints on how to achieve disabling of access logs for the ping router? I was unable to find anything easily in the docs for that matter. Many thanks

Just to clarify as of now this cannot be done just with configuration. Kevin was talking about how we could/should do it to actually fix the problem with code in Traefik.

ngbrown:
I see that this was declined because of regex. If it was exact match instead of regex, would that be acceptable? I see this configurable approach may be preferable over disabling on the ping router because an application may need to be regularly pinged through the traefik gateway.

changhyuni:
If Regex performance is an issue, why not implement substring?

zawadaa:
IMO for 99% of users enough will be one flag:
log or not *@internal

PR #9210, opened 2022-07-25
Add proxy overhead metrics

wtzhang23:
### What does this PR do?

Add metric for determining the overhead induced by the proxy per request. This is done by passing a timestamp datastructure via the request [context](https://pkg.go.dev/context) through the middleware chain and populating the timestamps within the entrypoint metrics middleware and the service metrics middleware.

### Motivation

Currently, Traefik provides metrics with respect to the entrypoint and the service. Although a rough estimate of the mean is technically able to be computed via a difference, we may want to know finer level details such as the distribution of proxy overhead.

### More
- [ ] Added/updated tests

### Additional Notes

* Not sure if this is something Traefik wants to support. 
* I need to add tests; if you have advice on where to add them please let me know.

### DIFF
--- a/docs/content/reference/static-configuration/cli-ref.md
+++ b/docs/content/reference/static-configuration/cli-ref.md
@@ -234,6 +234,9 @@ Datadog metrics exporter type. (Default: ```false```)
 `--metrics.datadog.addentrypointslabels`:  
 Enable metrics on entry points. (Default: ```true```)
 
+`--metrics.datadog.addproxylabels`:  
+Enable metrics on the proxy as a whole. (Default: ```false```)
+
 `--metrics.datadog.address`:  
 Datadog's address. (Default: ```localhost:8125```)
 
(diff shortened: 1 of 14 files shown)


rtribotte:
Hello @wtzhang23,

Thanks for this contribution!
It has been quite a long time since you opened this PR, sorry for the late "get in touch".

I'd like to know about the opportunity of having such a metric in Traefik, what the reasoning is, and how much (and why) this is a must-have for you.

I guess that the average overhead time added by Traefik for a request could be deduced by subtraction of the service request duration from the entry point request duration, and also, it is possible to get that information with tracing, this is why I am wondering about introducing this new metric.

Thanks!

rtribotte:
For completion, I just stumbled onto https://github.com/traefik/traefik/issues/9261, and my message is quite a duplicate of [@tomMoulard one](https://github.com/traefik/traefik/issues/9261#issuecomment-1366004999).

Considering the issue status, I will close this PR, but feel free to reopen it.

PR #9216, opened 2022-07-28
Support Nomad canary deployment

### LINKED ISSUE #8987: Need a way to define or alter the traefik `service name` based on consul tags
### Welcome!

This is a new description of an older problem originally discussed here: https://github.com/traefik/traefik/issues/3882

That issue was originally opened in such a way that it appeared related to Traefik v1 -- however the issue affects Traefik v2 as well and @ddtmachado requested a new issue be created to describe the problem.

## Problem

When using the nomad <-> consul <-> traefik integration, it is not possible to use the canary deployment workflow that is supported by nomad.

An introduction to the nomad canary roll-out strategy can be found [here](https://learn.hashicorp.com/tutorials/nomad/job-blue-green-and-canary-deployments).  The crux of the problem is that nomad will register the non-canary version of the service and the canary version of the service with the same consul service-name such that traefik routing policy cannot distinguish between the canary and non-canary instances of the service.  Both service types are presented in traefik as one 'service' representing a pool of instances of both instance types.  This means you can't use traefik routing policy to direct traffic to only one or the other instance type independently.

## Background

Nomad supports a service deployment lifecycle that includes a canary deployment process.  When a new version of a service that uses the canary deployment feature is deployed, nomad will spin up `n` instances of the canary service 'alongside' the existing service instances and allow the canary instances to be health-checked and verified as desired before they are 'promoted'.  After promotion, the normal deployment process is followed and the running older instances are replaced with the new version.

Nomad supports defining consul parameters specific for the canary instances via the [canary_tags](https://www.nomadproject.io/docs/job-specification/service#canary_tags) and [canary_meta](https://www.nomadproject.io/docs/job-specification/service#canary_meta) nomad parameters.  These parameters parallel the [tags](https://www.nomadproject.io/docs/job-specification/service#tags) and [meta](https://www.nomadproject.io/docs/job-specification/service#meta) parameters that are used to populate consul for regular instances of the service.  When nomad spins up canary instances it populates consul based on canary_tags/canary_meta and when nomad spins up regular instances it populates consul based on tags/meta.

For both canary and non-canary instances, nomad registers the service with the same consul service-name.  Nomad only allows canary/non-canary instances to differ in their presentation to consul via the canary_tags and canary_meta parameters.

Traefik on the hand, considers all consul registrations with the same name to be the 'same service'.  Its not currently possible to define a traefik routing policy that will route traffic to canary only or non-canary only nomad instances.  As far as traefik is concerned the canary and non-canary instances are the same pool of instances and these instances can't be independently addressed..

If we could customize the presentation of the service in traefik via the nomad `canary_tags` or `canary_meta` properties in such a way as to persuade traefik to consider these instance types as two distinct services, then we would be able to take advantage of the full flexibility of traefik routing behaviors to decide how we would like to route to these now distinct-from-the-perspective-of-traefik services.

## Possible Solution

I suggest allowing to define a consul parameter that allows deriving an alternate name by combining with a suffix with the base consul service name.

For example -- suppose support is added for a consul tag named `traefik.consulnamesuffix` and suppose a nomad service declaration like this:

```
service {
  name = "health-check-tool"
  port = "http"

  tags = [
    "traefik.enable=true",
    "traefik.http.middlewares.strip_canary_header.headers.customrequestheaders.Canary=",
    "traefik.http.routers.deploy_automation_tester.rule=Host(`${var.domain}`)",
    "traefik.http.routers.deploy_automation_tester.tls=true",
    "traefik.http.routers.deploy_automation_tester.entrypoints=https",
    "traefik.http.routers.deploy_automation_tester.middlewares=strip_canary_header@consulcatalog"
  ]

  canary_tags = [
    "traefik.consulnamesuffix=`canary`",
    "traefik.enable=true",
    "traefik.http.routers.deploy_automation_tester_canary.rule=Host(`${var.domain}`) && Headers(`canary`, `true`)",
    "traefik.http.routers.deploy_automation_tester_canary.tls=true",
    "traefik.http.routers.deploy_automation_tester_canary.entrypoints=https",
  ]

  connect {
    sidecar_service {}
  }

  check {
    name     = "alive"
    type     = "http"
    path     = "/ping"
    interval = "10s"
    timeout  = "2s"
  }
}
```

While an active nomad deployment is in progress we would see two services in `traefik` -- a service named `health_check_tool` which would represent the 'old' version of the service and a service named `health_check_tool_canary` which would represent the 'new' desired version of the service that is being evaluated.  In this example, there would also be two traefik routers -- one named `deploy_automation_tester`  would route traffic for `${var.domain}` to `health_check_tool` and one named `deploy_automation_tester_canary` which would route traffic for `${var.domain}` that contains the `canary: true` header to the `health_check_tool_canary` service.

A mechanism like this would integrate well with the nomad deployment lifecycle and would allow using the normal traefik routing policy mechanisms to send traffic to the canary/non-canary instances of the service in whatever manner is desired during deployment of a new service version.

rtribotte:
### What does this PR do?

This PR adds support for [Nomad canary deployment](https://learn.hashicorp.com/tutorials/nomad/job-blue-green-and-canary-deployments#deploy-with-canaries).

### Motivation

When Nomad orchestrator is a provider (of service registration) for Traefik,
one might have the need to distinguish within Traefik between a [Canary](https://learn.hashicorp.com/tutorials/nomad/job-blue-green-and-canary-deployments#deploy-with-canaries) instance of a service, or a production one.
For example if one does not want them to be part of the same load-balancer.

Therefore, this option, which is meant to be provided as one of the values of the `canary_tags` field in the Nomad [service stanza](https://www.nomadproject.io/docs/job-specification/service#canary_tags), allows Traefik to identify that the associated instance is a canary one.

Fixes #8987
Supersedes #9012

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/docs/content/routing/providers/consul-catalog.md
+++ b/docs/content/routing/providers/consul-catalog.md
@@ -470,6 +470,20 @@ You can tell Traefik to consider (or not) the service as a Connect capable one b
 
 This option overrides the value of `connectByDefault`.
 
+#### `traefik.consulcatalog.canary`
+
+```yaml
+traefik.consulcatalog.canary=true
+```
+
+When ConsulCatalog, in the context of a Nomad orchestrator,
+is a provider (of service registration) for Traefik,
+one might have the need to distinguish within Traefik between a [Canary](https://learn.hashicorp.com/tutorials/nomad/job-blue-green-and-canary-deployments#deploy-with-canaries) instance of a service, or a production one.
+For example if one does not want them to be part of the same load-balancer.
+
+Therefore, this option, which is meant to be provided as one of the values of the `canary_tags` field in the Nomad [service stanza](https://www.nomadproject.io/docs/job-specification/service#canary_tags),
+allows Traefik to identify that the associated instance is a canary one.
+
 #### Port Lookup
 
 Traefik is capable of detecting the port to use, by following the default consul Catalog flow.
--- a/docs/content/routing/providers/nomad.md
+++ b/docs/content/routing/providers/nomad.md
@@ -460,6 +460,19 @@ You can tell Traefik to consider (or not) the service by setting `traefik.enable
 
 This option overrides the value of `exposedByDefault`.
 
+#### `traefik.nomad.canary`
+
+```yaml
+traefik.nomad.canary=true
+```
+
+When Nomad orchestrator is a provider (of service registration) for Traefik,
+one might have the need to distinguish within Traefik between a [Canary](https://learn.hashicorp.com/tutorials/nomad/job-blue-green-and-canary-deployments#deploy-with-canaries) instance of a service, or a production one.
+For example if one does not want them to be part of the same load-balancer.
+
+Therefore, this option, which is meant to be provided as one of the values of the `canary_tags` field in the Nomad [service stanza](https://www.nomadproject.io/docs/job-specification/service#canary_tags),
+allows Traefik to identify that the associated instance is a canary one.
+
 #### Port Lookup
 
 Traefik is capable of detecting the port to use, by following the default Nomad Service Discovery flow.
--- a/pkg/provider/consulcatalog/config.go
+++ b/pkg/provider/consulcatalog/config.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"net"
+	"strings"
 
 	"github.com/hashicorp/consul/api"
 	"github.com/traefik/traefik/v2/pkg/config/dynamic"
@@ -37,8 +39,7 @@ func (p *Provider) buildConfiguration(ctx context.Context, items []itemData, cer
 		if len(confFromLabel.TCP.Routers) > 0 || len(confFromLabel.TCP.Services) > 0 {
 			tcpOrUDP = true
 
-			err := p.buildTCPServiceConfiguration(ctxSvc, item, confFromLabel.TCP)
-			if err != nil {
+			if err := p.buildTCPServiceConfiguration(item, confFromLabel.TCP); err != nil {
 				logger.Error(err)
 				continue
 			}
@@ -49,8 +50,7 @@ func (p *Provider) buildConfiguration(ctx context.Context, items []itemData, cer
 		if len(confFromLabel.UDP.Routers) > 0 || len(confFromLabel.UDP.Services) > 0 {
 			tcpOrUDP = true
 
-			err := p.buildUDPServiceConfiguration(ctxSvc, item, confFromLabel.UDP)
-			if err != nil {
+			if err := p.buildUDPServiceConfiguration(item, confFromLabel.UDP); err != nil {
 				logger.Error(err)
 				continue
 			}
@@ -75,8 +75,7 @@ func (p *Provider) buildConfiguration(ctx context.Context, items []itemData, cer
 			}
 		}
 
-		err = p.buildServiceConfiguration(ctxSvc, item, confFromLabel.HTTP)
-		if err != nil {
+		if err = p.buildServiceConfiguration(item, confFromLabel.HTTP); err != nil {
 			logger.Error(err)
 			continue
 		}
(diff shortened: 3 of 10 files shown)


apollo13:
I have tested it and it works great :+1:

pkg/provider/consulcatalog/config.go
```diff
@@ -300,3 +293,13 @@ func (p *Provider) addServer(ctx context.Context, item itemData, loadBalancer *d
...
 	return provider.Normalize("tls-" + item.Namespace + "-" + item.Datacenter + "-" + item.Name)
 }
+
+func getName(i itemData) string {
+	if !i.ExtraConf.ConsulCatalog.Canary {
+		return provider.Normalize(i.Name)
+	}
+
+	hasher := fnv.New64()
+	hasher.Write([]byte(strings.Join(i.Tags, ",")))
```
apollo13, line 303: I am going out on a limb here and wonder if it would make sense to sort the tags before joining. I am not sure if the tag order is guranteed.

rtribotte: Thanks for the feedback,

The hash has no other purpose than to produce a unique name temporarily, so we didn't sort the tags on purpose because it doesn’t really matter here (if we didn’t miss something).

apollo13: I guess it all depends on whether the tags returned by the consul catalog are guaranteed to keep order. Otherwise you will probably see a config change every X seconds (X = refreshinterval) even if it basically results in a noop. In that sense using sort is just a means of better stability against external changes -- I agree that it might not matter, but given that the tag list is probably small enough it won't hurt that much either. YMMV

docs/content/routing/providers/consul-catalog.md
```diff
@@ -470,6 +470,20 @@ You can tell Traefik to consider (or not) the service as a Connect capable one b
...
+#### `traefik.consulcatalog.canary`
+
+```yaml
+traefik.consulcatalog.canary=true
+```
+
+When ConsulCatalog, in the context of a Nomad orchestrator,
+is a provider (of service registration) for Traefik,
+one might have the need to distinguish within Traefik between a [Canary](https://learn.hashicorp.com/tutorials/nomad/job-blue-green-and-canary-deployments#deploy-with-canaries) instance of a service, or a production one.
+For example if one does not want them to be part of the same load-balancer.
```
kevinpollet, line 482:
```suggestion
For example, if one does not want them to be part of the same load-balancer.
```

docs/content/routing/providers/nomad.md
```diff
@@ -460,6 +460,19 @@ You can tell Traefik to consider (or not) the service by setting `traefik.enable
...
 
+#### `traefik.nomad.canary`
+
+```yaml
+traefik.nomad.canary=true
+```
+
+When Nomad orchestrator is a provider (of service registration) for Traefik,
+one might have the need to distinguish within Traefik between a [Canary](https://learn.hashicorp.com/tutorials/nomad/job-blue-green-and-canary-deployments#deploy-with-canaries) instance of a service, or a production one.
+For example if one does not want them to be part of the same load-balancer.
```
kevinpollet, line 471:
```suggestion
For example, if one does not want them to be part of the same load-balancer.
```

kevinpollet (approved):
LGTM 🚀

ldez (approved):
LGTM but I don't agree with oneliners `if`

AmirL:
Hi @rtribotte 

Could you help with ideas how to use it with Nomad provider, without Consul?
```

      service {
        name = "backend-api"
        port = "back"

        provider = "nomad"
        tags = [
          "traefik.enable=true",
          "traefik.http.routers.backend-app.rule=Host(`backend.internal`)",
        ]

        canary_tags = [
          "traefik.enable=true",
          "traefik.http.routers.backend-canary.rule=Host(`backend-canary.internal`)",
          // "traefik.http.routers.backend-canary.service=backend-canary",
          "trafik.nomad.canary=true"
        ]
```
backend.internal and  backend-canary.internal actually use the same Nomad service. If I uncomment different service name, it will throw use not existing service. 

Can you explain how to use "trafik.nomad.canary=true"?

the-sun-will-rise-tomorrow:
@AmirL You have a typo, should be `traefik` not `trafik`.

But I agree that more complete documentation (including a full Nomad job example + Traefik configuration) would be very helpful, as it stands the documentation for this option is quite minimal.

PR #9241, opened 2022-08-10
Control allocation and copy of labelNamesValues type

rtribotte:
### What does this PR do?

This PR makes the `labelNamesValues` type method `With' not mutate the original slice, by doing an explicit copy.

### Motivation

Avoid potential metrics overlapping.

### More

- [ ] Added/updated tests

### Additional Notes

This change needs to be benchmarked.


chuckyz:
Just an off-hand comment but in my local testing where I confirmed the initial Prometheus metrics performance was fixed, this fix is actually showing an _increase_ in perf vs. 2.8.1.  I suspect there is some performance hit; but that it's largely offset by Golang 1.19. :)

kevinpollet (approved):
LGTM

mpl (approved):

ldez (approved):
LGTM

PR #9278, opened 2022-08-26
Update access log add filecache using bufio

dragonflylee:
### What does this PR do?

Add memery cache for access log write to file

### Motivation

Reduce disk io on access log，help performances on some case

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/observability/access-logs.md
+++ b/docs/content/observability/access-logs.md
@@ -83,6 +83,32 @@ accessLog:
 --accesslog.bufferingsize=100
 ```
 
+### `fileCacheSize`
+
+This option represents the bytes of log size Traefik will keep in memory before writing them to the selected output.
+In some cases, this option can reduce IOPS on writing log.
+
+```yaml tab="File (YAML)"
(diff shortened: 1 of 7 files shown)


mpl:
Hello @dragonflylee ,

we have several concerns about this PR.

1) We have no doubt that this is a valuable trade-off for you, but adding an option has maintenance costs. So we would have liked to see a proposal with concrete figures on the improvements this PR would bring. This would also have helped us gauge the interest of the community for this feature.

2) We think you might be able to mitigate your problem, if you let an external tool be in charge of the actual writes to disk. Which means traefik would write to stdout, and the external tool would get the stdout stream as input.

3) it might be problematic to only specify a cache in bytes. Because the cache size would rarely ever correspond with the end of a log message. So we'd end up writing partial log lines (it should fix by itself with the subsequent write, but still).

4) Lastly, your PR made us rethink again the validity/usefulness of the existing BufferingSize option. And it made us maybe redo it in a way that would perhaps also cover your use case. So we want to take the time to think about it some more and come back to it someday.

Issue #7729, opened 2021-01-07
Ingressclasses not available on namespace scope level


bschoenbach:
Hi traefik community,

After upgrading of k8s cluster to  **v1.18+**  we get the following error during startup of traefik **v2.3+**

`E0107` 06:45:40.895430       1 reflector.go:178] pkg/mod/k8s.io/client-go@v0.18.2/tools/cache/reflector.go:125: Failed to list *v1beta1.IngressClass: ingressclasses.networking.k8s.io is forbidden: User "system:serviceaccount:traefik:traefik" cannot list resource "ingressclasses" in API group "networking.k8s.io" at the cluster scope`

**The problem:**
Within k8s v1.18+ the "IngressClass" was introduced to avoid bad annotations on "Ingress" resource objects themselves (see [here)](https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/) . However,  the idea of that concept is pretty cool but lacks in implementation because it requires access on cluster scope level. Hence, developers or teams which can only work on namespace scope cannot use traefik ingress > 2.3.x on gke > 1.18.x.

**Possible solutions could be:**

- Introduce a configuration switch to disable the listing/ usage of ingressclasses during startup of traefik  v2.3+
OR
- Raise this issue to k8s community so that they can work on the cluster scope issue of "ingressclasses" on k8s v1.18+. In my opinion it must be possible to define ingressclasses also on namespace scope.

**Temporary Solution** 

Currently we downgraded our traefik to **v2.2.11** where lisitng of "ingressclasses" is not implemented and treafik starts up correctly also on gke 1.18+

nustiueudinastea:
Having the same issue when trying to run the latest Helm chart using namespaced deployment.

joeldeteves:
Just ran into the same issue. Downgrading for now.

joeldeteves:
Hello, any update on this issue?

We are unable to patch Traefik for security releases until this is resolved.

Thank you,

alexangas:
Also experiencing this issue on Kubernetes 1.19.7 with Traefik 2.3.3.

Possibly related issue posted [on Stack Overflow](https://stackoverflow.com/q/63109422/6651) but the suggested fix was already in the helm chart.

rtribotte:
Hello @bschoenbach @nustiueudinastea @joeldeteves @alexangas,

Thanks for your interest in Traefik!

[...]

PR #9281, opened 2022-08-26
Add flag to the k8s ingress provider to disable ingressclass lookup

jandillenkofer:
### What does this PR do?

This PR introduces a configuration switch to disable the listing/ usage of ingressclasses during startup of traefik v2.3+.

### Motivation

I was inspired to this PR by [issue #7729](https://github.com/traefik/traefik/issues/7729).
We run Traefik on a k8s production environment where we do not have the permission to list ingressclasses.
A downgrade to v2.2.11 of Traefik was also not possible, because our k8s cluster (RKE version v1.22.10+rke2r2) does no longer support the ingress class resource in version networking.k8s.io/v1beta1, which this version of Traefik still tries to list.

### More

- [X ] Added/updated tests
- [X ] Added/updated documentation

### DIFF
--- a/pkg/provider/kubernetes/ingress/client.go
+++ b/pkg/provider/kubernetes/ingress/client.go
@@ -39,7 +39,7 @@ type marshaler interface {
 // WatchAll starts the watch of the Provider resources and updates the stores.
 // The stores can then be accessed via the Get* functions.
 type Client interface {
-	WatchAll(namespaces []string, stopCh <-chan struct{}) (<-chan interface{}, error)
+	WatchAll(disableIngressClassLookup bool, namespaces []string, stopCh <-chan struct{}) (<-chan interface{}, error)
 	GetIngresses() []*networkingv1.Ingress
 	GetIngressClasses() ([]*networkingv1.IngressClass, error)
 	GetService(namespace, name string) (*corev1.Service, bool, error)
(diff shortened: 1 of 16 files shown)


Patanouk:
Hello,
Would you by any chance have an idea about the timeline for this change to be released?

To be clear, this is **not** a request to get this released earlier. I know this is open-source, so it is perfectly fine if maintainers have other priorities

mpl (approved):

docs/content/providers/kubernetes-ingress.md
```diff
@@ -344,6 +344,34 @@ providers:
...
+### `disableIngressClassLookup`
+
+_Optional, Default: false_
+
+If the parameter is set to `true`,
+Traefik will not discover IngressClasses in the cluster.
+By doing so, it alleviates the requirement of giving Traefik the rights to look IngressClasses up.
+Furthermore, when this option is set to `true`,
+Traefik is not able to handle Ingresses with IngressClass references,
+therefore such Ingresses will be ignored.
```
mpl, line 356:
```suggestion
therefore such Ingresses will be ignored.
Please note that annotations are not affected by this option.
```

nmengin:
Hello @Patanouk,

We've planned to bring it in the next Traefik major version.

tomMoulard (approved):
LGTM :ok_hand:

PR #9285, opened 2022-08-29
Log tls client subject

xmessi:
### What does this PR do?

This PR adds support for logging the TLS client subject, when it is used.

### Motivation

In my use case, the TLS client subject contains a username; TLS client certificates are used to authenticate users. Logging this improves traceability of actions performed, similarly to the `ClientUsername` field which is filled in case of HTTP Basic Auth. 

### More

- [X] Added/updated tests

### Additional Notes

Thanks for considering this PR for inclusion


rtribotte (approved):
Thanks 👍

nmengin (approved):

ldez (approved):
LGTM

PR #9322, opened 2022-09-11
Add a note on case insensitive regex matching

NEwa-05:
### What does this PR do?

This pull request adds one example to use with the Path or PathPrefix rule for a case-insensitive backend.

Fix https://github.com/traefik/traefik/issues/4180

### More


rtribotte (approved):
LGTM

Issue #9307, opened 2022-09-05
Feature: Support AllNamespacesNamespace when using the Nomad Provider


0teh:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Using the AllNamespacesNamespace as the nomad provider namespace (traefik CLI option `--providers.nomad.namespace=*`) results in no services being found.

In my particular usecase, I have multiple workloads running across multiple namespaces, and would like to provide a single entrypoint for them all.  I think this would also help match user expectations when comparing the traefik configuration with querying nomad directly (`nomad service list -namespace=*`).

I am filing this as a feature request instead of a bug report because I think there are some design considerations: if nomad has two services with the same name but in different namespaces, with this feature, traefik would see one service with two servers (based on experiments with local changes to traefik).  This is fine for my own usecase but might not be for others.

I am willing to prepare a PR, especially if the design considerations above can be disregarded for now.

Thanks!

rtribotte:
Hello @0teh,

Thanks for your interest in Traefik!

From the nomad documentation, it seems that there is no `*` value that would allow matching any namespace.
Considering that, and if we didn’t miss something, the Nomad provider only lack of the support of multiple namespaces, like Consul and ConsulCatalog providers do (https://github.com/traefik/traefik/issues/8896). WDYT?

0teh:
Hello,

The nomad documentation is quite sparse, but the `*` namespace can be found on the page for [`nomad service list`](https://www.nomadproject.io/docs/commands/service/list) as well as in the sample requests for the [equivalent HTTP request](https://www.nomadproject.io/api-docs/services#sample-request).  This works for listing nomad services, but fetching the details of a specific service require the exact namespace, which is part of the list results.

Supporting a namespaces array like consul would be very nice, and seems to me like an additional feature alongside `*`.  I think it makes sense that the namespace is then appended to the traefik service name.

rtribotte:
Thanks for your feedback!

You are right. I overlooked that the `*` namespace is available for the `nomad service list` endpoint.

And we have indeed a limitation induced by how the namespace option is used right now. It is configured on the client and then used globally, so the `*` namespace cannot be used because it would then be used to read the service details.

We think that supporting the `*` namespace is a valid proposal.
[...]

PR #8979, opened 2022-05-03
Support multiple namespaces for Consul and ConsulCatalog providers

rtribotte:
### What does this PR do?

This PR adds the support for multiple namespaces configuration for the Consul and ConsulCatalog providers.

### Motivation

Fixes #8896

Be more compliant with the [Namespaces](https://www.consul.io/docs/enterprise/namespaces) feature of [Consul Enterprise](https://www.consul.io/docs/enterprise).

### More

- [ ] Added/updated tests
- [x] Added/updated documentation

### Additional Notes

Co-authored-by: Kevin Pollet <pollet.kevin@gmail.com>

PR #9332, opened 2022-09-14
Support multiple namespaces in the Nomad Provider

0teh:
### What does this PR do?

Deprecates the `namespace` option in the Nomad provider and adds a `namespaces` option to discover services in multiple namespaces, similar to the Consul and Consul Catalog providers.

### Motivation

As an operator of a small Nomad cluster, I would like to provide a Traefik instance to workloads across multiple namespaces.

### More

- [x] Added/updated tests

### Additional Notes

These changes are basically the same as changes in #8979.

See #9307 for previous discussion.


kevinpollet (commented):
Thanks 👏

pkg/config/static/static_config.go
```diff
@@ -333,6 +333,10 @@ func (c *Configuration) ValidateConfiguration() error {
 		return fmt.Errorf("consul provider cannot have both namespace and namespaces options configured")
 	}
 
+	if c.Providers.Nomad != nil && c.Providers.Nomad.Namespace != "" && len(c.Providers.Nomad.Namespaces) > 0 {
+		return fmt.Errorf("nomad provider cannot have both namespace and namespaces options configured")
```
mpl, line 337:
```suggestion
		return fmt.Errorf("Nomad provider cannot have both namespace and namespaces options configured")
```

mpl (approved):

PR #9335, opened 2022-09-14
Support access log timestamp with milliseconds

### LINKED ISSUE #9317: Support access log timestamp with milliseconds
### Welcome!

Traefik access log timestamps with at least millisecond precision, or a configuration option to enable it. A more complex solution could allow an operator to provide a Go `time.Format()` string in the static configuration to use.

By default, the `logrus.JSONFormatter` uses RFC3339 which does not include sub-second precision. However, the struct does take a `TimestampFormat` string that could modify this behavior with a limited amount of work on the Traefik side: https://pkg.go.dev/github.com/sirupsen/logrus#JSONFormatter

A very naïve implementation on the JSON logger in `pkg/middlewares/accesslog/logger.go`:
```diff
@@ -91,7 +91,11 @@ func NewHandler(config *types.AccessLog) (*Handler, error) {
 	case CommonFormat:
 		formatter = new(CommonLogFormatter)
 	case JSONFormat:
-		formatter = new(logrus.JSONFormatter)
+		if len(config.TimestampFormat) > 0 {
+			formatter = &logrus.JSONFormatter{TimestampFormat: config.TimestampFormat}
+		} else {
+			formatter = new(logrus.JSONFormatter)
+		}
```

(FWIW, I found [this question](https://community.traefik.io/t/timestamp-in-access-logs-has-no-milliseconds/11909) on the community forums that unfortunately wasn't answered.)

V-R-Dighe:
### Motivation

Fixes #9317

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/middlewares/accesslog/logger.go
+++ b/pkg/middlewares/accesslog/logger.go
@@ -92,7 +92,11 @@ func NewHandler(config *types.AccessLog) (*Handler, error) {
 	case CommonFormat:
 		formatter = new(CommonLogFormatter)
 	case JSONFormat:
-		formatter = new(logrus.JSONFormatter)
+		if len(config.TimestampFormat) > 0 {
+			formatter = &logrus.JSONFormatter{TimestampFormat: config.TimestampFormat}
+		} else {
+			formatter = new(logrus.JSONFormatter)
+		}
 	default:
 		log.WithoutContext().Errorf("unsupported access log format: %q, defaulting to common format instead.", config.Format)
 		formatter = new(CommonLogFormatter)
--- a/pkg/types/logs.go
+++ b/pkg/types/logs.go
@@ -34,11 +34,12 @@ func (l *TraefikLog) SetDefaults() {
 
 // AccessLog holds the configuration settings for the access logger (middlewares/accesslog).
 type AccessLog struct {
-	FilePath      string            `description:"Access log file path. Stdout is used when omitted or empty." json:"filePath,omitempty" toml:"filePath,omitempty" yaml:"filePath,omitempty"`
-	Format        string            `description:"Access log format: json | common" json:"format,omitempty" toml:"format,omitempty" yaml:"format,omitempty" export:"true"`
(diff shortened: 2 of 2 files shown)


thaJeztah:
In case it's useful; if you want to use a default with nanosecond precision,  in containerd and docker, we define a `RFC3339NanoFixed` const, to use nanosecond-precision, but to make sure timestamps always have the same length (to prevent trailing `0` to be stripped); 

https://github.com/containerd/containerd/blob/3d7a93fed8f3153783ba1490f08b9368f8cf9de0/cmd/containerd/command/main.go#L354-L375

```go
func setLogFormat(config *srvconfig.Config) error {
	f := config.Debug.Format
	if f == "" {
		f = log.TextFormat
	}

	switch f {
	case log.TextFormat:
		logrus.SetFormatter(&logrus.TextFormatter{
			TimestampFormat: log.RFC3339NanoFixed,
			FullTimestamp:   true,
		})
	case log.JSONFormat:
		logrus.SetFormatter(&logrus.JSONFormatter{
			TimestampFormat: log.RFC3339NanoFixed,
		})
	default:
		return fmt.Errorf("unknown log format: %s", f)
	}

	return nil
}
```


const itself is here; https://github.com/containerd/containerd/blob/22beecb7d9bd06e743be9da7e519976417755466/log/context.go#L41-L43

```go
// RFC3339NanoFixed is time.RFC3339Nano with nanoseconds padded using zeros to
// ensure the formatted time is always the same number of characters.
RFC3339NanoFixed = "2006-01-02T15:04:05.000000000Z07:00"
```

tfny:
Hey @V-R-Dighe and @thaJeztah, thanks so much for this PR.  In the end, we have decided to revamp the whole logging system as we are switching from [Logrus](https://github.com/sirupsen/logrus) to [zerolog](https://github.com/rs/zerolog).  

To learn more about this, you can check out this [issue](https://github.com/traefik/traefik/issues/9453) and associated [PR](https://github.com/traefik/traefik/pull/9515).

PR #9357, opened 2022-09-20
fix: UDP loadbalancer tags not being used with Consul Catalog

t3hchipmunk:
### What does this PR do?

Fixes an issue of loadbalancer tags being used in Nomad jobs for Consul Catalog not picking up the port in the Consul Catalog addServerUDP. I used a lot of the addServerTCP code to realize the difference and correct for those differences.

### Motivation

We recently ran into an issue of Traefik not picking up the port for our service written in Nomad using Consul Catalog (https://community.traefik.io/t/traefik-showing-port-0-for-service/15796). The port that Traefik would establish for the service was always 0. I was able to get it working by using a file configuration but that was insufficient for our use case as Nomad was in play.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/consulcatalog/config.go
+++ b/pkg/provider/consulcatalog/config.go
@@ -206,10 +206,11 @@ func (p *Provider) addServerTCP(item itemData, loadBalancer *dynamic.TCPServersL
 		return errors.New("address is missing")
 	}
 
-	port := item.Port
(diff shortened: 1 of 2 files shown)


rtribotte (approved):
LGTM 👍

ldez (approved):
LGTM

mmatur (approved):

Issue #9368, opened 2022-09-23
Character "= " into a query parameter. How to escape it in my route ?


r0m1CH:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

Trying to catch an URL with a query parameter APIKey which ends with an =

### What did you see instead?

An error: mux: number of parameters must be multiple of 2

### What version of Traefik are you using?

2

### What is your environment & configuration?

"traefik.http.routers.tools-api-rtr.rule=Host(`tools.$DOMAINNAME`) && PathPrefix(`/api`) && Query(`apikey`, `$tools_API_KEY`)"
Where apikey="sdfgsdfDSDsdç232="

traefiker:
Hi! I'm Træfiker :robot: the bot in charge of communication regulation.

Thanks for your interest in Traefik!

We dedicate the issue tracker to bug reports and feature requests only. My advanced AI has spotted that your issue might be a configuration problem or relates to something that doesn't look like a bug.

To confirm this, please join our [Community Forum](https://community.containo.us/) and reach out to us on the [Traefik section](https://community.containo.us/c/traefik).

In case I'm wrong (well, that would be embarrassing :sweat_smile:), my developers will re-open the issue and fix me!

In the meantime, you can double check [Traefik's documentation](https://docs.traefik.io/).

ldez:
Hello,

the right syntax for the query parameter is ```Query(`api=aaa`)```

https://doc.traefik.io/traefik/routing/routers/#rule

But there is a bug if you try to do  ```Query(`api=aa=a`)```, I fixed the bug in https://github.com/traefik/traefik/pull/9369

It will be a part of v2.8.6.

https://community.traefik.io/t/character-into-a-query-parameter-how-to-escape-it/15878/2

r0m1CH:
Thanks. I will wait for this version.

PR #9369, opened 2022-09-23
fix: query parameter matching with equal

ldez:
### What does this PR do?

query parameter with equals

### Motivation

Related to #9368

### More

- [x] Added/updated tests


kevinpollet (approved):
LGTM

PR #9408, opened 2022-10-02
add support for HTTPRouteFilter's in k8s Gateway API

romantomjak:
### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/gateway/kubernetes.go
+++ b/pkg/provider/kubernetes/gateway/kubernetes.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/http"
 	"os"
 	"sort"
 	"strconv"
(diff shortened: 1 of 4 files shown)


pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -1655,6 +1680,42 @@ func loadTCPServices(client Client, namespace string, backendRefs []v1alpha2.Bac
...
+				RedirectScheme: &dynamic.RedirectScheme{
+					Scheme:    *filter.RequestRedirect.Scheme,
+					Permanent: *filter.RequestRedirect.StatusCode == http.StatusMovedPermanently,
+				},
+			}
+		case v1alpha2.HTTPRouteFilterRequestHeaderModifier:
+			headers := make(map[string]string)
+			for _, h := range filter.RequestHeaderModifier.Add {
+				headers[http.CanonicalHeaderKey(string(h.Name))] = h.Value
+			}
```
rtribotte, line 1701: If I did not miss something the Headers middleware cannot "add" headers value, it would either set a header or delete it.

rtribotte (commented):
Thanks for this contribution!

pkg/provider/kubernetes/gateway/kubernetes.go
romantomjak: I got that from scanning through the Gateway API spec. The filters prop on [HTTPRouteRule](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteRule) object is a slice of [HTTPRouteFilter](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPRouteFilter). The filter object has a prop requestHeaderModifier of type [HTTPHeaderFilter](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.HTTPHeaderFilter). And the header filter has set, add and remove props. From the looks of it they are there since v1beta1

rtribotte: Sure, I was only noticing this, and warning that the current implementation of the `Headers` middleware will not fit all the expectations of the `HTTPHeaderFilter` as it does not support adding header values.
This needs more discussion but our current stance would be to drop the `HTTPHeaderFilter` support completely until the `Headers` middleware can support it.

romantomjak: Ohhh, I see! Yes, that makes sense, I'll drop it from the PR.

romantomjak: I dropped the `HTTPHeaderFilter` and added some extra validation for the redirect middleware 👍 

Are there any special considerations required for adding the http header merging to the header middleware or can I just whip up a PR for that?

rtribotte: Sorry for the late reply.
> Are there any special considerations required for adding the http header merging to the header middleware or can I just whip up a PR for that?

Well, the only configuration available today achieves both the set and delete operations.
The add operation can be a new configuration option, but, in that case, what about the precedence between options?
There are for sure other alternatives, like breaking the current configuration options.
In a nutshell, we have several options. It could be great if you can open a proposal issue for that!

romantomjak: Great idea, I'll do that 👍

rtribotte (approved):
Thanks 👍

tomMoulard (approved):
LGTM :ok_hand:

nmengin (approved):
LGTM 👏

PR #9412, opened 2022-10-04
Remove raw cert escape in PassTLSClientCert middleware

rtribotte:
### What does this PR do?

This PR removes the escaping of the raw client cert as a value for the `X-Forwarded-Tls-Client-Cert` header.

### Motivation

As discussed in https://github.com/traefik/traefik/issues/9379, the escape form of the raw client cert is not always supported/expected on backends, as with PhenixID.

Fixes https://github.com/traefik/traefik/issues/9379

### More

- [x] Added/updated tests


ldez (approved):
LGTM

juliens (approved):
LGTM

Issue #9423, opened 2022-10-07
AllowEmptyService flag is not taken into account on traefikService in kubernetes crd


jeromeguiard:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I was adding an empty service to a TraefikService

This parameter was setted
```
--providers.kubernetescrd.allowemptyservices=true
```

### What did you see instead?

Building a TraefikService object with empty service as source led to this error:

```
Error while building TraefikService: subset not found for web/my-app-canary","providerName":"kubernetescrd","serviceName":"my-app","time":"2022-08-05T12:34:56Z"}
```

### What version of Traefik are you using?

v2.9.1

### What is your environment & configuration?

```yaml
apiVersion: traefik.containo.us/v1alpha1
kind: TraefikService
metadata:
  name: my-app
  namespace: web
spec:
  weighted:
    services:
      - name: my-app-canary
        namespace: web
        port: 80
        weight: 0
      - name: my-app-stable
        namespace: web
        port: 80
        weight: 100

---

apiVersion: v1
kind: Service
metadata:
  name: my-app-stable
  namespace: web
spec:
  selector:
    app: my-app-stable
  type: ClusterIP
  ports:
    - name: main
      port: 80
      targetPort: 80
      protocol: TCP

---

apiVersion: v1
kind: Service
metadata:
  name: my-app-canary
  namespace: web-ui
spec:
  selector:
    app: my-app-canary
  type: ClusterIP
  ports:
    - name: main
      port: 80
      targetPort: 80
      protocol: TCP

```

jeromeguiard:
It appears that the parameter isn't passed into the config builder. I am trying to build a fix https://github.com/traefik/traefik/commit/99552f53ce324629f2586ec758e7328a6c59c68d

PR #9424, opened 2022-10-07
fix: allowemptyservice issue#9423

jeromeguiard:
### What does this PR do?

This PR fix issue #9423 related to flag not taken into account for traefik service

### Motivation

It is a bug in the configBuilder object that doesn't represent expected code behavior

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/crd/fixtures/with_empty_services.yml
+++ b/pkg/provider/kubernetes/crd/fixtures/with_empty_services.yml
@@ -15,3 +15,19 @@ spec:
       services:
         - name: whoami-without-endpoints-subsets
           port: 80
+        - name: tr-svc-es
+          kind: TraefikService
+
+---
+apiVersion: traefik.containo.us/v1alpha1
+kind: TraefikService
+metadata:
+  name: tr-svc-es
+  namespace: default
+
+spec:
+  weighted:
+    services:
+      - name: whoami-without-endpoints-subsets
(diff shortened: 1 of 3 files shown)


ldez:
Hello,

can you explain the Draft state of this PR?

jeromeguiard:
Hello,

The draft status was because I wanted to make sure about the tests. I added one instead of changing an existing one.

It is done now. 

Thanks

rtribotte:
Hello @jeromeguiard,

We thought about the changes and about the use case, and here are our takes:

- We do agree that the propagation of the `allowEmptyServices` option all along TraefikServices is missing. So we are moving the PR to the review state.

- Regarding Canary or Blue/Green deployment scenarios, we think that propagating the `allowEmptyServices` option would help if weights of the WRR are correctly managed. Indeed, propagating the `allowEmptyServices` makes it possible to build a WRR with some services that have servers and some others that don't. This would make a fraction of requests (depending on the weight) end up with a 503 status code.

- Nonetheless, because the `allowEmptyServices` option is a provider option, it would impact all services discovered by the IngressRouteCRD provider. Having to leverage it to support properly Blue/Green deployments may not fit in every situation. Hence, we may lack an alternative, for instance, ignoring zero-weight WRR services (that will, anyway, never handle the traffic).

mpl (approved):

PR #9427, opened 2022-10-08
Updated rbac to use version v1 instead of v1beta1

BarnabyDT:
ClusterRole and ClusterRoleBinding don't exist in rbac.authorization.k8s.io/v1beta1

### What does this PR do?

Updates the example configuration to use endpoint rbac.authorization.k8s.io/v1 instead of rbac.authorization.k8s.io/v1beta1, because ClusterRole and ClusterRoleBinding don't exist in v1beta1 anymore.

### Motivation

Was running v1.7 traefik to test something in a local cluster and noticed the documentation wasn't correct.

### DIFF
--- a/docs/user-guide/kubernetes.md
+++ b/docs/user-guide/kubernetes.md
@@ -35,7 +35,7 @@ For the sake of simplicity, this guide will use a ClusterRoleBinding:
 ```yaml
 ---
 kind: ClusterRole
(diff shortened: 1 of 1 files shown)


BarnabyDT:
Actually, scrap this. If you're using a later version of k8s, probably shouldn't be using 1.7 anyway.

PR #9431, opened 2022-10-11
Rework servers load-balancer to use the WRR

juliens:
### What does this PR do?
This PR reworks the servers load-balancer to use the WRR load balancer instead of the oxy load balancer.
The healthcheck was rewrote to fit the WRR healthcheck mechanism.

### Motivation
To simplify the code base.

### More

- [x] Added/updated tests

### Additional Notes

As we are reusing the sticky cookie mechanism we implemented in WRR, we dropped the support of old cookie value ( before 2.5 ) in the sticky feature. 
This means that if a Traefik is upgraded from <2.5 to this version, the active user's sticky session will be discarded.

The services are now reused, this means that if two routers use the same service, load balancing will be shared by both routers.


ddtmachado:
Linking the motivation of the motivation :) 
https://github.com/traefik/traefik/issues/9122 and https://github.com/traefik/traefik/issues/1035#issuecomment-1262168925

ldez (approved):
LGTM

kevinpollet (approved):
LGTM ✨

rtribotte (approved):
LGTM

PR #9460, opened 2022-10-21
Remove side effect on default transport tests

sdelicata:
This PR removes side effects on default transport, that doesn't have to be altered.


ldez (approved):
LGTM

PR #9494, opened 2022-11-04
Add snap packaging

ed10vi:
### What does this PR do?

Add snapcraft.yaml so it can be distributed as snap

### Motivation

It is useful for installing in edge devices running Ubuntu Core

### More

- [ ] Added/updated tests

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,82 @@
+name: traefik
+adopt-info: traefik
+summary: The Cloud Native Application Proxy
+description: |
+  Traefik (pronounced traffic) is a modern HTTP reverse proxy
+  and load balancer that makes deploying microservices easy.
+  Traefik integrates with your existing infrastructure components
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

Traefik is a cloud-native proxy, snap is not a cloud technology.

Snap is only supported by Ubuntu.

In the world of the cross-distribution packaging system, Snap is not the winner (I will not open the debate about the reasons), the winner is Flatpack.

If at some point, we want to support Snap we will use the [goreleaser system](https://goreleaser.com/customization/snapcraft/)

Also, we don't want to support all packaging systems inside the Traefik repository.

In this context, we choose to only maintain Docker as a packaging system.

But there is a [community GitHub organization](https://github.com/traefik-contrib) that can be a home for this kind of element.

If you agree I will invite you to this organization, and create a repo for the snap.

ed10vi:
Hello.

Thank you for your response. I agree flatpak is the winer in open source desktop app distribution (not for edge devices or closed source apps).

Not to support all packaging systems seems a reasonanle decision

I'll take a look to traefik-contrib.

Thank you again.

PR #9500, opened 2022-11-09
proxy: context log with error

Torwang1:
Bug fixes:
- for Traefik v2: use branch master

### What does this PR do?

print the error log with:
- context log;
- error level;

### Motivation

When the proxy encounters a problem, print the error log

### More

Problems I'm having:

- In my plugin. I create new http.Request object, using the wrong URL.Schema. So, httputil.ReverseProxy returns an error message;
- In my environment, logs use the **info** level;

So, missing critical error messages, got me stuck in troubleshooting.


ldez:
Hello,

The level is "debug" because this can be spammy.

We will decline this PR.

PR #9502, opened 2022-11-10
Add a safe.Atomic wrapper for handler switchers

sylr:
I believe the safe wrapper around the HTTP/TCP/UDP Handlers is here to protect them against data races during configuration changes.

I don't think we need to ensure synchronization for them so using the atomic primitives instead of sync ones could improve performances.

Any thoughts ?


rtribotte:
Hello @sylr,

Thanks for your contribution!

We took the time to review the changes and found that we will not accept them.

We are not aware of any performance issue on the handlerSwitcher mechanism, and this PR does not clarify what the issue is.
Therefore, since there is no justification for this PR, and since it is not trivial, we are closing it.

PR #9506, opened 2022-11-14
docs: update helm repository

charlie-haley:
### What does this PR do?
This PR updates the Helm Repository URL in the documentation to point at the new repository. The old URL will remain live for compatibility reasons.

### Motivation
The Helm Repository has changed so we need to ensure new users are using the correct URL.

### More

- [ ] Added/updated tests


nmengin (approved):
LGTM 📖

PR #9519, opened 2022-11-21
doc: clarify PathPrefix greediness

### LINKED ISSUE #9514: Include caveat in documentation that `PathPrefix` matches loosely
### Welcome!

It seems that `PathPrefix` matches exact prefixes, i.e. without taking into consideration `/` as special separator characters.

Example: ```PathPrefix(`/products`)``` matches all of the following paths:

- ✅ `/products`
- ✅ `/products/shoes`
- ✅ `/products/shirts`
- ⚠️ `/productsforsale`
- ⚠️ `/productsforsale/shoes`

The last two I found surprising, regarding the examples given in the documentation. I could not find anything to say that this was intended behavior but I would suggest to include this as explicit caveat in the docs:

> Use a `*Prefix*` matcher if your service listens on a particular base path but also serves requests on sub-paths. For instance, `PathPrefix: /products` would match `/products` but also `/products/shoes` and `/products/shirts`, **as well as `/productsforsale`**. Since the path is forwarded as-is, your service is expected to listen on `/products`.

Traefik version: `v2.9.4`

mpl:
### What does this PR do?

This PR clarifies that the PathPrefix matcher is very greedy and does not care about / as a separator.

### Motivation

Fixes #9514 

### More

~- [ ] Added/updated tests~

### DIFF
--- a/docs/content/routing/routers/index.md
+++ b/docs/content/routing/routers/index.md
@@ -276,10 +276,11 @@ The table below lists all the available matchers:
 
 !!! info "Path Vs PathPrefix"
 
-    Use `Path` if your service listens on the exact path only. For instance, `Path: /products` would match `/products` but not `/products/shoes`.
+    Use `Path` if your service listens on the exact path only. For instance, ```Path(`/products`)``` would match `/products` but not `/products/shoes`.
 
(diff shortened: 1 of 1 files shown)


rtribotte (approved):
LGTM

PR #9531, opened 2022-11-22
Update routing syntax

### LINKED ISSUE #9520: Unify router rule matchers
### Welcome!

The goal of this proposal is to clarify router rule matchers in Traefik.

We see four issues with the current rule syntax:

1) Regexps are defined using the [gorilla/mux](https://github.com/containous/mux) syntax, which arbitrary enforces naming matching groups.
This adds complexity, and should be removed, as it's not even used by Traefik.
2) It is sometimes unclear which matcher uses regexps (e.g., there are `Host`, and `HostRegexp` matchers, but not `Path`, and `PathRegexp` matchers).
There should be an explicit regexp matcher when appropriate (i.e., `PathRegexp`, `PathPrefixRegexp`, and `QueryRegexp`).
3) Matchers that accept multiple values are ORing them when it makes sense (e.g., `Host`, `Path`, etc.), or ANDing them when it doesn't (e.g., `Query`, `Headers`, etc.).
Even thought it's not wrong, this can cause confusion where one would expect those matchers to deal with their values similarly (i.e., ORing them all).
Making every matcher accept only one value, and combining them with logical operators is more verbose, but eliminates this issue.
4) `Query` and `Headers` matchers both expect a key and value as arguments.
These two matchers should get their argument using the same syntax (e.g., using `Headers` matcher one's).

These changes would transform:
```yml
rule: Host(`example.com`, `example.org`) && Path(`/products/{id:[a-z]+}`) || Query(`mobile=true`)
```

into:
```yml
rule: (Host(`example.com`) || Host(`example.org`)) && PathRegexp(`/products/[a-z]+`) || Query(`mobile`, `true`)
```

skwair:
### What does this PR do?

- Regexp matcher now use Go's regexp syntax.
- Regexp matchers are now explicitly named SomeMatcherRegexp.
- All matchers now take a single value and should be explicitly combined using logical operators to mimic previous behavior.
- Remove deprecated HostHeader matcher.
- Rename Headers* matchers to Header*

### Motivation

Fixes #9520.

### More

- [x] Added/updated tests

### Additional Notes


ldez (approved):
LGTM

juliens (approved):
LGTM

tomMoulard (approved):
LGTM :ok_hand:

PR #9534, opened 2022-11-23
Fix error when setting ServerUp metric labels

kevinpollet:
### What does this PR do?

This pull request fixes an error raised when setting the ServerUp gauge metric value.

### Motivation

To fix an error.

### More

- [ ] Added/updated tests


ldez (approved):
LGTM

PR #9555, opened 2022-11-28
Skip service if an error occurs

### LINKED ISSUE #9547: Traefik returns 404 when service with 0 pods and service with ExternalName are combined
### Welcome!

### What did you do?

Loadbalancing is applied as expected when having an IngressRoute with references to a regular service with selector and a service with ExternalName.

IngressRoute with two services is show below. The weight is set very high for the regular service and thereby most request will stay within cluster unless no pods matches the regular service.

```yaml
kind: IngressRoute
metadata:
  name: hello-world-with-failover
spec:
  entryPoints:
    - websecure
  routes:
    - kind: Rule
      match: Host(`hello-world.apps.foo.bar`)
      services:
        - kind: Service
          name: regular-service
          namespace: demo
          port: 80
          weight: 10000000
        - kind: Service
          name: externalname-service
          namespace: demo
          port: 443
          scheme: https
          weight: 1
  tls:
    secretName: hello-world-tls
```

### What did you see instead?

When the regular service does not match any pods (e.g. scaled to 0) then I expected all requests to be handled by the ExternalName service.
Instead Traefik returns 404 Not Found in that case.

It could be related to the change in [pull-request 8057](https://github.com/traefik/traefik/pull/8057).

Version: 2.9.4
Kubernetes: 1.23
Traefik Helm Chart: 20.2.0

```yaml
```

Enabled TRACE log level.

When we remove the regular-service and only have an ExternalName Service then the server is added. Here is the log from that scenario:
```
{"entryPointName":"websecure","level":"debug","middlewareName":"metrics-service","middlewareType":"Metrics","msg":"Creating middleware","routerName":"verification-hello-world-with-failover-1234@kubernetescrd","serviceName":"verification-hello-world-with-failover-1234","time":"2022-11-28T09:13:11Z"}
{"entryPointName":"websecure","level":"debug","msg":"Creating load-balancer","routerName":"verification-hello-world-with-failover-1234@kubernetescrd","serviceName":"verification-hello-world-with-failover-1234","time":"2022-11-28T09:13:11Z"}
{"entryPointName":"websecure","level":"debug","msg":"Creating server 0 https://hello-world.site2.foo.bar.dk:443","routerName":"verification-hello-world-with-failover-1234@kubernetescrd","serverName":0,"serviceName":"verification-hello-world-with-failover-1234","time":"2022-11-28T09:13:11Z"}
{"level":"debug","msg":"child https://hello-world.site2.foo.bar.dk:443 now UP","time":"2022-11-28T09:13:11Z"}
{"level":"debug","msg":"Propagating new UP status","time":"2022-11-28T09:13:11Z"}
```
And the route is added:
```
{"entryPointName":"websecure","level":"debug","msg":"Adding route for hello-world.apps.foo.bar with TLS options default","time":"2022-11-28T09:13:11Z"}
```

However, then the regular-service and the ExternalName service are added the non of these kinds of log statements are shown.

ErikLundJensen:
### What does this PR do?

Fixes #9547 

### Motivation

A similar solution exists for kubernetes_tcp.go

### More

We should probably add some logging of errors, however, the logger is not available inside this method.
I'll let the maintainers decide how they want the logging to be implemented.

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes_http.go
+++ b/pkg/provider/kubernetes/crd/kubernetes_http.go
@@ -223,7 +223,7 @@ func (c configBuilder) buildServicesLB(ctx context.Context, namespace string, tS
 	for _, service := range tService.Weighted.Services {
 		fullName, k8sService, err := c.nameAndService(ctx, namespace, service.LoadBalancerSpec)
 		if err != nil {
-			// logger.Error().Msg(err)
+			log.Ctx(ctx).Error().Err(err).Msgf("Skip service: %s/%s", service.Namespace, service.Name)
 			continue
 		}
 
--- a/pkg/provider/kubernetes/crd/kubernetes_http.go
+++ b/pkg/provider/kubernetes/crd/kubernetes_http.go
@@ -223,7 +223,8 @@ func (c configBuilder) buildServicesLB(ctx context.Context, namespace string, tS
 	for _, service := range tService.Weighted.Services {
 		fullName, k8sService, err := c.nameAndService(ctx, namespace, service.LoadBalancerSpec)
 		if err != nil {
-			return err
+			// logger.Error().Msg(err)
+			continue
 		}
 
 		if k8sService != nil {


ErikLundJensen:
I see that the integration test is failing as no error is returned as expected. The IngressRouteTCP does not return an errror, the question is therefore if the test case should be changed so that IngresRoute and IngresRouteTCP has the same error semantics.
Changing the error handling could have impact on running installations of Traefik. Do we need a flag to indicate that the new error handling is used to stay backwards compatible?

ddtmachado:
Hi @ErikLundJensen, as this fix actually changes the behavior of the provider and because it's fixing an unconfirmed bug (the use case was not validated yet) I'm tempted to view this more like a proposal that Traefik should skip failed services instead of breaking all the chain.

We're currently working on the next major version, 3.0, which means we have "some" freedom to introduce breaking changes, so its the right time to discuss this kind of behavior changes. In this sense would you be comfortable in opening a new issue, in a proposal format, and then closing [the bug possible](https://github.com/traefik/traefik/issues/9547)?

With that we would be able to link the proposal to [the epic tracking CRD improvements](https://github.com/traefik/traefik/issues/9125) and have discussions on the approach such as using a flag or not and what are the benefits and impacts of having it.

ErikLundJensen:
Okay, I'll create a new proposal and close the this pull-request and related bug.

PR #9595, opened 2022-12-08
make file provider more resilient wrt first configuration

mpl:
### What does this PR do?

This PR lets the file provider recover, when the initial configuration is broken, but a watcher has been setup. Because the file provider can still potentially be operational, and work with subsequent configurations sent by the watcher.

### Motivation

We noticed that the file provider dies if the initial configuration is broken, which seemed surprising given that providers are supposed to be somewhat resilient.

### More

~- [ ] Added/updated tests~

### Additional Notes


rtribotte (approved):
LGTM

juliens (approved):
LGTM

ldez (approved):
LGTM

PR #9598, opened 2022-12-09
fix: sanitize X-Forwarded-Proto header is RedirectScheme middleware

### LINKED ISSUE #9597: RedirectScheme middleware returns bad gateway for joined X-Forwarded-Proto header
### Welcome!

### What did you do?

If Traefik is used behind a proxy server it might receive requests containing a `X-Forwarded-Proto` header. If there are multiple proxies involved or one of them is misbehaving it might happen that there are multiple X-Forwarded-Proto headers.
According to the [HTTP/1.1 RFC](https://www.rfc-editor.org/rfc/rfc2616#page-31)
>     Multiple message-header fields with the same field-name MAY be
>     present in a message if and only if the entire field-value for that
>     header field is defined as a comma-separated list [i.e., #(values)].
>     It MUST be possible to combine the multiple header fields into one
>     "field-name: field-value" pair, without changing the semantics of the
>     message, by appending each subsequent field-value to the first, each
>     separated by a comma. The order in which header fields with the same
>     field-name are received is therefore significant to the
>     interpretation of the combined field value, and thus a proxy MUST NOT
>     change the order of these field values when a message is forwarded.
So in that case the values of the X-Forwarded-Proto might be joined together by a proxy in front of Traefik which will result in the header `X-Forwarded-Proto: https, https`.

In #9121 the RedirectScheme middleware got the ability to interpret that header to decide whether it should return a redirect. In those cases it doesn't behave correctly, though.

For requests **with** the space between the header values Traefik always returns status code 502 with this error in the debug logs:
```
'502 Bad Gateway' caused by: parse \"https://https, https://<router>/\": invalid character \" \" in host name
```

Note this is only possible with insecure forwarded headers enabled.

Relevant part in the code: https://github.com/traefik/traefik/blob/e54ee89330a800d509da7b11b46a6ecbb331e791/pkg/middlewares/redirect/redirect_scheme.go

### What did you see instead?

Traefik returns a response with status code 502

2.9.5, problem was introduced in 2.7.2 by #9121

relevant CLI args
```
--entrypoints.http.address=:8000/tcp
--entrypoints.http.forwardedheaders.insecure=true
```

middleware configuration:
```
redirectScheme:
  permanent: true
  scheme: https
```

ingressroute (but also works with ingress or other kinds of configuration):
```
kind: IngressRoute
metadata:
  name: httpbin-with-middleware
spec:
  entryPoints:
  - http
  - https
  routes:
  - kind: Rule
    match: Host(`<router>`) && PathPrefix(`/`)
    services:
    - port: 8080
      name: httpbin
    middlewares:
    - name: ingress-controller-redirect-to-https@kubernetescrd
```

_No response_

ldez:
### What does this PR do?

Sanitize X-Forwarded-Proto header is RedirectScheme middleware

### Motivation

Fixes #9597

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/middlewares/redirect/redirect_scheme.go
+++ b/pkg/middlewares/redirect/redirect_scheme.go
@@ -71,12 +71,10 @@ func clientRequestURL(req *http.Request) string {
 		// Given that we're in a middleware that is only used in the context of HTTP(s) requests,
 		// the only possible valid schemes are one of "http" or "https", so we convert back to them.
 		switch {
-		case strings.EqualFold(xProto, "ws"):
+		case strings.EqualFold(xProto, "ws"), strings.EqualFold(xProto, schemeHTTP):
 			scheme = schemeHTTP
-		case strings.EqualFold(xProto, "wss"):
+		case strings.EqualFold(xProto, "wss"), strings.EqualFold(xProto, schemeHTTPS):
 			scheme = schemeHTTPS
-		default:
-			scheme = xProto
 		}
 	}
 
--- a/pkg/middlewares/redirect/redirect_scheme_test.go
+++ b/pkg/middlewares/redirect/redirect_scheme_test.go
@@ -72,7 +72,7 @@ func TestRedirectSchemeHandler(t *testing.T) {
 			headers: map[string]string{
 				"X-Forwarded-Proto": "bar",
 			},
-			expectedURL:    "https://bar://foo",
+			expectedURL:    "https://foo",
 			expectedStatus: http.StatusFound,
 		},
 		{


juliens (approved):
LGTM

sylr:
There is a typo in the commit message title `is` -> `in`.

Thor77:
Is it intended that this would also cause a redirect for an invalid X-Forwarded-Proto header (like a combination of two of them, `https, https`, which some software creates)?

ldez:
Yes, it's intended because by convention this header is mono-valued.
Even inside the [RFC7239](https://www.rfc-editor.org/rfc/rfc7239#section-5.4) about `Forwarded` it's a mono-valued header.

pkg/middlewares/redirect/redirect_scheme.go
```diff
@@ -14,10 +14,16 @@ import (
 
 const (
 	typeSchemeName  = "RedirectScheme"
-	uriPattern      = `^(https?:\/\/)?(\[[\w:.]+\]|[\w\._-]+)?(:\d+)?(.*)$`
+	uriPattern      = `^(https?:\/\/)(\[[\w:.]+\]|[\w\._-]+)?(:\d+)?(.*)$`
```
rtribotte, line 17:
```suggestion
	uriPattern      = `^((?:wss?|https?):\/\/)(\[[\w:.]+\]|[\w\._-]+)?(:\d+)?(.*)$`
```

We should probably also support `ws` and `wss` scripts.

ldez: no because we replace it by http/https

rtribotte (approved):
LGTM

tomMoulard (approved):

Issue #9108, opened 2022-06-17
Config option to remove Hub button from dashboard


magicrobotmonkey:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

In traefik 2.8, there is a big yellow button labelled "Go to Hub Dashboard" on the traefik dashboard. It would be great if there was a config option to remove this button. This issue was previously raised in #8912, but it appears that was closed because hub integration wasn't included in 2.7.

ldez:
Hello,

could you explain why do you want to remove this button?

magicrobotmonkey:
We don't have a hub account, so it's confusing to users. When users click it, it just takes them to a login page they don't have credentials for.

magicrobotmonkey:
Yes but even once it's launched, I may not use it.

ddtmachado:
@magicrobotmonkey Traefik Hub Beta is now launched so anyone clicking on the button will land on the correct page, so would you agree the original problem is not there anymore?

Also it's literally just a link, no data collection, no stop the world advertising and it's a useful link to jump to the Hub Dashboard for those that will use it. Would you still be bothered by it?

Personally, I feel that having a new config option on the static configuration would be worst than just ignoring a link/button on an UI :)

magicrobotmonkey:
I don't disagree that more config options may not be the way to go, but considering that there appears to already be configuration around activating/configuring hub integration, maybe this button should be gated behind that?

schodemeiss:
> I don't disagree that more config options may not be the way to go, but considering that there appears to already be configuration around activating/configuring hub integration, maybe this button could be gated behind that?

^This^. I've been trying to disable it by disabling everything I can about the hub. It's an annoying distraction of a big yellow button which none of m'colleagues will be clicking on or using anytime soon. Please consider letting us remove it.

tfny:
Hey @magicrobotmonkey and @schodemeiss, 

Thanks for the feedback.  We don't have any plans to remove the button because we want to make sure that people using the product know that they can use Traefik Hub easily and for free.  However, when it comes time to reconsider in a few months, we will take your feedback into consideration.

[...]

PR #9608, opened 2022-12-14
Only show Hub button in navbar if enabled in config

SuNNjek:
### What does this PR do?

This PR removes the big "Go to Hub Dashboard" button from the navigation bar at the top if the feature is not enabled in the configuration.

### Motivation

The button is *very* prominently placed in the navigation bar for something that people aren't necessarily going to use. In fact, lots of people (myself included) find this button pretty annoying and too "in your face" (as evidenced by the number of comments and :+1:  reactions on #9108). That's why I decided to create this PR, which only shows the button when people actually want to use this feature.

### More

- [ ] Added/updated tests

### DIFF
--- a/webui/src/components/_commons/NavBar.vue
+++ b/webui/src/components/_commons/NavBar.vue
@@ -30,7 +30,7 @@
           </q-tabs>
           <div class="right-menu">
             <q-tabs>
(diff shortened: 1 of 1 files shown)


ldez:
As explained here: https://github.com/traefik/traefik/issues/9108#issuecomment-1170519502

We will decline your PR.

SuNNjek:
Figured this would happen, guess the "time to reconsider" hasn't come yet :smile:

PR #9609, opened 2022-12-15
Update submitting pull requests to include language about drafts

tfny:
### What does this PR do?

I updated the language to include our draft policy.  While we have always had a policy of not reviewing drafts, it may not have been clear to everyone.

### Motivation

I think it is good to be clear with our community so they know what to expect and so we do not have conflicting expectations.

### More

- [ ] Added/updated tests
- [ x] Added/updated documentation


rtribotte (approved):
LGTM

PR #9621, opened 2022-12-27
fix no rate limiting if average is 0

witalisoft:
### What does this PR do?

[Docs](https://doc.traefik.io/traefik/middlewares/http/ratelimit/#average) about `ratelimit` says that you can omit rate limiting by specifying `average` as `0`.  In traefik version `< 2.7.2` it was true version `2.7.2` introduce a bug and you cannot disable rate limiting in that way. The bug was probably related to bumping package [golang.org/x/time](https://github.com/traefik/traefik/compare/v2.7.1...v2.7.2#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R70). This change treats zero limits [differently](https://cs.opensource.google/go/x/time/+/f0f3c7e86c11ca8f8b99d970e28445dda0b41195:rate/rate.go;dlc=1f47c861a9ac5a6e7645609f91b895398ff31642) and now `rate.Inf` should be set when `average == 0`. 

### Motivation

To restore the option with disabling rate limiting.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/ratelimiter/rate_limiter.go
+++ b/pkg/middlewares/ratelimiter/rate_limiter.go
@@ -85,6 +85,10 @@ func New(ctx context.Context, next http.Handler, config dynamic.RateLimit, name 
 	// the value of maxDelay does not matter since the reservation will (buggily) give us a delay of 0 anyway.
 	var maxDelay time.Duration
 	var rtl float64
+
+	// rate.Inf for no rate limiting if config.Average == 0
+	rtl = float64(rate.Inf)
+
 	if config.Average > 0 {
 		rtl = float64(config.Average*int64(time.Second)) / float64(period)
 		// maxDelay does not scale well for rates below 1,


mpl:
We had already noted that we were relying on a buggy behaviour of x/time, and we had open an issue for that:
https://github.com/golang/go/issues/47221
Are you saying that they finally fixed the issue above in x/time, and that this PR adjusts things consequently?

witalisoft:
I wasn't aware of this issue was already submitted and I don't have any idea if this was fixed. My PR findings were discovered because of the docs vs code difference and the PR fix an issue with the following [scenario](https://github.com/traefik/traefik/blob/master/pkg/middlewares/ratelimiter/rate_limiter_test.go#L224-L232)

mpl (approved):

rtribotte (approved):
Thanks 👍

tomMoulard (approved):
LGTM :ok_hand:

PR #9633, opened 2023-01-02
Remove observability for internal resources

### LINKED ISSUE #6861: Filtering out ping access logs

### LINKED ISSUE #9170: Stop Tracing Internal Traffic (ping@internal, prometheus@internal, etc.)
### Welcome!

When tracing is enabled, Traefik traces those internal traffic such as ping@internal, prometheus@internal. 

It is the best practices to stop tracing health checks and monitoring type of traffic because the resulting traces provide less values while adding burdens to the tracing infra. These traffic are quite chatty!

Can Traefik stop tracing them? Or at least provides an option to turn off tracing on those?

Screen Shot 2022-07-06 at 6 38 24 PM

rtribotte:
### What does this PR do?

This PR disables access logs, metrics, and tracing, for internal routers and services.

### Motivation

Fixes #9170
Fixes #6861

### More

- [x] Added/updated tests


juliens (approved):
LGTM

netsandbox:
I'm not sure if unconditionally dropping observability for all **@internal** resources is the best way, because maybe there are use cases where you want or need this information.

For example if you secure the Prometheus metrics with BasicAuth and want to see the remote IP for the HTTP 403 errors, you need the information for the `prometheus@internal` service in the **access log**.
For the same use case you may also need the **Prometheus metrics** for `prometheus@internal` to monitor the HTTP 4xx errors.

dhirschfeld:
> I'm not sure if unconditionally dropping observability for all **`@internal`** resources is the best way, because maybe there are use cases where you want or need this information.

I'm just piping up as a user who'd really like the ability to turn off these logs as they just add noise and make debugging harder in the common case.

If you're trying to debug an issue with internal services then I guess not having any logs is equally as painful.

Perhaps the logging for internal services could be configurable so you could set it to e.g. only log errors rather than every successful ping request.

rtribotte:
Hello @netsandbox @dhirschfeld,

Thanks for your feedback!

We did think about adding an option to get the observability back for internal resources when opening the PR.
But we also thought at that time that we wanted to gather first feedback to see the traction around it before yet adding another option.

This makes total sense and we are going to rework the PR to add the option.

rtribotte:
Hello @netsandbox @dhirschfeld,

We changed our mind about introducing the opt-in option to get back internal observability in this PR.
We want to address it in another PR and to have first the opportunity to gather more feedback on the need, to determine what would be the best approach for this option.
Would you be inclined to open a new issue to discuss it?

zetaab:
@rtribotte @juliens what is status of this PR could it be merged? As I see it, it would be nice to ignore these internal services

ldez:
to be merged a PR needs to be up to date, CI ok, and 3 approves

rtribotte:
Hello,

We decided to move the target of this PR to the next milestone, as we changed our minds about making it optional on another iteration.
The next step is to introduce an option to control the observability of the internal resources.

mmatur (approved):

nmengin (approved):
LGTM 🎉

comminutus:
What version will this be included in next and how can one activate / use it?

PR #9640, opened 2023-01-04
customerrors middleware: allow preserving status code and method

vmax:
### What does this PR do?

Extend `customerrors` middleware to support new behaviours:

* `preserveStatusCode`: allow using status code returned by service serving error page.
* `preserveMethod`: query the service serving the error page using the same HTTP method as the one that caused the original error.

### Motivation

We've switched from NGINX Ingress controller to Traefik and that changed how `OPTIONS` requests are treated in the error (e.g. 503) cases. On the initial `OPTIONS` call, even if backend is unavailable, we want to serve `Access-Control-Allow-Origin: *` with `204 No Content` status code.

For more context see kubernetes/ingress-nginx#2140

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/middlewares/http/errorpages.md
+++ b/docs/content/middlewares/http/errorpages.md
@@ -135,3 +135,19 @@ The table below lists all the available variables and their associated values.
 |------------|--------------------------------------------------------------------|
 | `{status}` | The response status code.                                          |
 | `{url}`    | The [escaped](https://pkg.go.dev/net/url#QueryEscape) request URL. |
+
+
+### `preserveStatusCode`
+
+Determines whether HTTP status code returned by the error service should be respected.
+By default, code of the error (from `status` range) is used.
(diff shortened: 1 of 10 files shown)


tfny:
Hey!  Thanks for this PR. We have reviewed it and we think that adding this behavior to the errorpage middleware will dilute the purpose of the errorpage middleware. 

We think it would work best as a plugin. If you decide to create it as a plugin, we would be glad to add it to our plugin catalog.

RubenNL:
Hi! Would a pull request like this, but where the method is moved to the a url parameter, something like `{method}` be accepted?

I would love to have the `preserveStatusCode` config option for nearly the same reason as @vmax, but for me it's sending a 200 status code on a 503, to prevent default browser error pages to show on some old browsers.

Can this be re-evaluated?

PR #9642, opened 2023-01-05
Add middleware via annotation made clear

KR-Ravindra:
For some reason this missed out in the documentation ,... Please add this here.

### More

- [ ] Added/updated tests


kevinpollet:
Hello @KR-Ravindra and thanks for your interest in Traefik,

Unfortunately, we are declining this pull request as the above section (see https://doc.traefik.io/traefik/routing/providers/kubernetes-ingress/#annotations) is already explaining how to reference resources in annotations.

Issue #9101, opened 2022-06-14
Allow multiple providers of the same type


ddtmachado:
### What did you expect to see?

This feature proposal should unify other isolated requests on specific providers like HTTP, Docker and Consul.

The goal is to add support for specifying and instantiating multiple providers of the same type. To achieve that the providers section must be modified so it uses named resources similar to a middleware. Example:

```yaml
providers:
  mycrd:
    kubernetescrd: {}
  myingress:
    kubernetesIngress: {}      
```

As this is a breaking change (configuration wise) it is only possible in a new major version.

ldez:
After more thinking about this subject, I am not sure that a generic approach around multiple providers is the right approach.

From previous issues and the general design, I think that generic approach will lead to needed complexity (define a name even if 90% of the users use only one provider) and bad practices (merge dev, staging, and prod environment into the same Traefik).

I think that a better approach is just to split the Docker provider into 2 providers ("standard" Docker and Docker Swarm) and evaluate case by case how providers can handle multiple sources.

EdgarSedov:
Not like i have any credibility here but imho splitting at the first glance looks like a best option to me, because currently in docker.go there're condition checks almost for everything on any step.

veelasky:
+1 on this one.

on my case, I just want to to configure multiple `file` definition with different directories. since traefik not detecting changes in subdirectory (https://github.com/traefik/traefik/issues/3986)

alfianabdi:
Upvote this one. 
In my case, I want to use 1 traefik as gateway to multiple kubernetes clusters, all clusters are belong to one environment - not merging dev,stg, prod in one traefik instance.

soupdiver:
I'm having a setup of 2 "datacenters".
home and remote.
Each is running its own Consul setup.
I want traefik that runs in "remote" to use Consul of both "home" _and_ "remote" and dataproviders and expose services

ddtmachado:
Following the arguments presented on this thread plus some internal discussions the maintainers agreed we should not attempt implementing a generic multi provider support but rather tackle use cases individually on each provider. 

For that reason I'm closing this as declined. Thanks everyone that helped with use cases and proposing alternative solutions, and if you have an use case that you think would require the declaration of multiple providers please open a proposal with the specific use case around that provider only.

PR #9652, opened 2023-01-10
Split Docker provider

### LINKED ISSUE #4790: Move docker.swarm to standalone Swarm provider

Feature

I think that swarm option should be moved to standalone Swarm provider, because it's not possible to use proxy with services deployed with docker stack command while you can still run containers on the same host (because of missing priviledge mode in swarm)

ldez:
### What does this PR do?

the provider Docker has been split into 2 providers:

- Docker provider (without Swarm support)
- Swarm provider  (Swarm support only)

### Motivation

Fixes #4790
Related to #9101

### More

- [x] Added/updated tests


docs/content/providers/swarm.md
```diff
@@ -0,0 +1,697 @@
...
+
+    Please consider the security implications by reading the [Security Note](#security-note).
+
+    A good example can be found on [Bret Fisher's repository](https://github.com/BretFisher/dogvscat/blob/master/stack-proxy-global.yml#L124).
+
+### `endpoint`
+
+_Required, Default="unix:///var/run/docker.sock"_
+
+See section [Docker Swarm API Access](#docker-api-access) for more information.
```
rtribotte, line 201:
```suggestion
See the [Docker Swarm API Access](#docker-api-access) section for more information.
```

docs/content/providers/docker.md
```diff
@@ -12,8 +12,7 @@ A Story of Labels & Containers
 
 Attach labels to your containers and let Traefik do the rest!
 
-Traefik works with both [Docker (standalone) Engine](https://docs.docker.com/engine/)
-and [Docker Swarm Mode](https://docs.docker.com/engine/swarm/).
+Traefik works with [Docker (standalone) Engine](https://docs.docker.com/engine/).
```
rtribotte, line 15: I don't know exactly how to reword this if so, but it seems that previously the intention was just to yield about the two modes, so do we need it anymore?
Maybe keeping the previous intention by adding a link to the swarm provider documentation page could make sense.

rtribotte (approved):
LGTM 👌

michaelkrieger:
Once split, are there plans to support multiple docker providers in the standalone docker provider (ie: ssh to 1-2 hosts and the and local docker.socket)?

Individual services can be run on a docker host that is a part of a swarm, but using only the mesh networking connecting multiple hosts.  This allows two or three servers to run services, each communicating with each other over the mesh network, while attributed to the individual host vs a stack.

Currently this is possible with the file provider manually setting up the services on alternate hosts. Adding configuration discovery in traefik would be a huge win for small configs that don't want to deploy swarm services but do have more than one machine.

rtribotte:
Hello @michaelkrieger,

As per https://github.com/traefik/traefik/issues/9101#issuecomment-1168428421 and https://github.com/traefik/traefik/issues/9101#issuecomment-1316970977, we are not planning to provide multiple docker provider support.

Nonetheless, as you noticed, the split would allow having the swarm and docker provider at the same time.

michaelkrieger:
Hi @rtribotte 
It was my understanding that those comments refer to "**generic** multi provider support" of all providers but may consider addressing "use cases individually on **each provider**".

As it relates to the Docker standalone provider (not generic multi-provider), adding the ability to poll more than one socket and consolidate the results would allow two or three systems to be provided by Traefik (using the common network created by simply doing a `swarm init` and `swarm join`, without actually using swarm deployments).  This is useful for small environments which may want to service requests on multiple systems, but not deploy more than one of any service.

> we are not planning to provide multiple docker provider support.
I understand. It is disappointing.  It appears it would not require too much to add and provides a great use case.

Thanks for the time.

juliens (approved):
LGTM

mmatur (approved):
LGTM

PR #9687, opened 2023-01-31
fix: differentiate UDP stream and TCP connection in logs

rtribotte:
### More

- [ ] Added/updated tests


ldez (approved):
LGTM

PR #9701, opened 2023-02-08
doc: add note about remoteaddr strategy

### LINKED ISSUE #9696: Traefik uses a different set of XFF header values for its whitelisting middleware than what it passes on
### Welcome!

### What did you do?

First off, I'm the person who created this post on the community forum: https://community.traefik.io/t/setting-up-both-forwardedheaders-and-ipwhitelist-or-how-to-trust-ip-and-whitelist-the-proxy-not-the-end-user/17423 (just to let you know that I did indeed search the forum).

So requests are being made like this:

    me ---> Cloudflare ---> Traefik

Basically, when trusting forwarded headers (with the `forwardedHeaders.trustedIPs` config in the `entryPoint`, Traefik adds on the "last requester"'s IP just like many other reverse proxies do.

So if we were to track the XFF values across the request, it would look something like this:

    N/A ---> $my_ip ---> $my_ip, $cloudflare_ip

However, I noticed that the XFF that is being _passed on_ to the applications behind it (i.e. `$my_ip, $cloudflare_ip` in this case) is NOT the XFF that Traefik works off of when doing IP Whitelisting!

In other words, when Traefik looks up the IP using `ipStrategy.depth = 1`, I'd *expect* to Traefik to look at the rightmost value in the XFF header `$my_ip, $cloudflare_ip` (i.e. the XFF value it passes on to `whoami`).

### What did you see instead?

Instead, Traefik looks at the rightmost value in the XFF header `$my_ip` (i.e. the value that it receives from Cloudflare, NOT the value it passes onto the services sitting behind it).

I've scoured through the docs, and the difference between "XFF value we _received_" vs. "XFF value we're _passing on_" is _never_ documented anywhere, especially not in the ip whitelisting middleware section (where it talks about reading values from the XFF header).

So although it may be "working as intended", I'm filing this under the "bug" category because it is certainly not what is _expected_, at least based on the documentation. Thus, I'd like to see the documentation clarify the fact that it's working off of the _received_ XFF, not the XFF that Traefik _adds on_ to.

Secondly, I would greatly appreciate it if Traefik were to add an option somewhere to allow, well, basically anything reading off of the XFF header to use _the value that's being passed on_, rather than the value that's being _received_.

My use case is that I'd like to block all traffic hitting Traefik that is NOT being proxied by Cloudflare... but as of now, it is not possible to do this in Traefik due to the fact that the XFF header it works off of (remember, there are two distinct values here) _doesn't_ include Cloudflare's IP (even though the XFF being passed on _does_).

So if an option was added to change "which" XFF value Traefik looks at for things like the IP Whitelist middleware - the one it receives, vs. the one it passes on - I would be able to get the IP Whitelist middleware to work off of the latter, not the former, and IP Whitelist based on Cloudflare's IP, not the "real" IP.

Thank you.

2.9.6

Please see the community post above for more details, but here is the configuration for the whitelist middleware:

```yml
cloudflare-whitelist:
  ipWhiteList:
    sourceRange: $list_of_cloudflare_ips
    ipStrategy:
      depth: 1
```

And I also have a `forwardedHeaders.trustedIPs` setup so that Traefik "trusts" and passes on the XFF values (this is confirmed working).

_No response_

mpl:
### What does this PR do?

this PR adds a couple of notes about the default ipstrategy, and about the lifecycle of X-Forwarded-For.

### Motivation

Users were not aware of the default remoteaddr ipstrategy.

### More

~- [ ] Added/updated tests~

### Additional Notes

Fixes #9696

### DIFF
--- a/docs/content/middlewares/http/ipwhitelist.md
+++ b/docs/content/middlewares/http/ipwhitelist.md
@@ -75,7 +75,10 @@ The `sourceRange` option sets the allowed IPs (or ranges of allowed IPs by using
 
 ### `ipStrategy`
 
-The `ipStrategy` option defines two parameters that set how Traefik determines the client IP: `depth`, and `excludedIPs`.
+The `ipStrategy` option defines two parameters that set how Traefik determines the client IP: `depth`, and `excludedIPs`.  
+If no strategy is set, the default behaviour is to match sourceRange against the Remote address found in the request.
+
+!!! important "As a middleware, whitelisting happens before the actual proxying to the backend takes place. In addition, the previous network hop only gets appended to `X-Forwarded-For` during the last stages of proxying, i.e. after it has already passed through whitelisting. Therefore, during whitelisting, as the previous network hop is not yet present in `X-Forwarded-For`, it cannot be matched against sourceRange."
 
 #### `ipStrategy.depth`
 


docs/content/middlewares/http/ipwhitelist.md
```diff
@@ -75,7 +75,10 @@ The `sourceRange` option sets the allowed IPs (or ranges of allowed IPs by using
 
 ### `ipStrategy`
 
-The `ipStrategy` option defines two parameters that set how Traefik determines the client IP: `depth`, and `excludedIPs`.
+The `ipStrategy` option defines two parameters that set how Traefik determines the client IP: `depth`, and `excludedIPs`.  
+If no strategy is set, the default behaviour is to match sourceRange against the Remote address found in the request.
```
ldez, line 79:
```suggestion
If no strategy is set, the default behavior is to match `sourceRange` against the Remote address found in the request.
```

docs/content/middlewares/http/ipwhitelist.md
```diff
@@ -75,7 +75,10 @@ The `sourceRange` option sets the allowed IPs (or ranges of allowed IPs by using
...
+
+!!! important "As a middleware, whitelisting happens before the actual proxying to the backend takes place. In addition, the previous network hop only gets appended to `X-Forwarded-For` during the last stages of proxying, i.e. after it has already passed through whitelisting. Therefore, during whitelisting, as the previous network hop is not yet present in `X-Forwarded-For`, it cannot be matched against sourceRange."
```
ldez, line 81:
```suggestion
!!! important "As a middleware, whitelisting happens before the actual proxying to the backend takes place. In addition, the previous network hop only gets appended to `X-Forwarded-For` during the last stages of proxying, i.e. after it has already passed through whitelisting. Therefore, during whitelisting, as the previous network hop is not yet present in `X-Forwarded-For`, it cannot be matched against `sourceRange`."
```

ldez (changes requested):

ldez (approved):
LGTM

JaneJeon:
Hello - just wanted to let you know that you might also want to make the same note in InFlightReq and RateLimit middlewares (where the same ipStrategy + XFF confusion can arise)

kevinpollet (approved):
LGTM

nmengin (approved):
LGTM 👍 📚

PR #9704, opened 2023-02-09
fix(sec): upgrade mkdocs to 1.3.0

whiteChen233:
### What happened？
There are 1 security vulnerabilities found in mkdocs 1.2.2
- [MPS-2022-6935](https://www.oscs1024.com/hd/MPS-2022-6935)

### What did I do？
Upgrade mkdocs from 1.2.2 to  for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-mkdocs==1.2.2
+mkdocs==1.3.0
 mkdocs-traefiklabs>=100.0.7
 
 appdirs==1.4.4


ldez:
Hello,

There is no CVE related to that.

The OSCS site is only in Chinese and Google is not able to translate it (https://www-oscs1024-com.translate.goog/hd/MPS-2022-6935?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=fr&_x_tr_pto=wapp)

The references on the page are useless because they are not related to relevant information.

I don't see how this can be exploited in our workflow.

Issue #5502, opened 2019-09-26
Traefik 1.7 Allow use of socket for Datadog metrics


morsok:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

Currently we have to specify an url:port for the datadog metrics. On kubernetes, you can't provide 127.0.0.1 but instead provide the ip of the node on which the pod is running, which can be done using the downward api, and when using the helm chart, you use cli arguments to set the address.

However Traefik when in HA mode with consul for example, only use consul configuration and ignores cli arguments and so tries to use the default value of 127.0.0.1.

One easy way to fix that issue when in HA would be to allow using the datadog unix socket that you can mount into the pod (see https://docs.datadoghq.com/developers/dogstatsd/unix_socket/), when doing that, every traefik pod will talk to the locally mounted socket from the agent running on the same node without any issues.

dduportal:
Hi @morsok , thanks a lot for reporting this feature request as we released Traefik v2.0, the v1.7 is still supported but does not accept new features: only bug fixes and security fixes.

As Traefik v2.0 is not supporting yet Consul as KV provider, would you mind describing the problem on our Community forum so a solution to help you might be found and shared.

traefiker:
Hi! I'm Træfiker :robot: the bot in charge of communication regulation.

Thanks for your interest in Traefik!

We dedicate the issue tracker to bug reports and feature requests only. My advanced AI has spotted that your issue might be a configuration problem or relates to something that doesn't look like a bug.

To confirm this, please join our [Community Forum](https://community.containo.us/) and reach out to us on the [Traefik section](https://community.containo.us/c/traefik).

In case I'm wrong (well, that would be embarrassing :sweat_smile:), my developers will re-open the issue and fix me!

In the meantime, you can double check [Traefik's documentation](https://docs.traefik.io/).

morsok:
@dduportal Ok thank you for your swift answer, that's what i was afraid of.

I openned https://github.com/helm/charts/pull/17466#issuecomment-535564512 to compensate.

PR #9714, opened 2023-02-14
Add support to send DataDog traces via Unix Socket

der-eismann:
### What does this PR do?

This PR adds the possibility to send DataDog traces to a local UNIX socket instead of an HTTP host. There was #5502 for Traefik v1.7 in the past, but the issue is still valid. 

### Motivation

When using the DataDog agent on Kubernetes, you usually have one agent per host. But on Kubernetes you can't address them via `localhost` unless you are using the host network. That's why they added the possibility to send traces via a local socket that gets mounted by the agent and the service that should send traces.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/observability/tracing/datadog.md
+++ b/docs/content/observability/tracing/datadog.md
@@ -23,24 +23,46 @@ tracing:
 
 #### `localAgentHostPort`
 
-_Required, Default="127.0.0.1:8126"_
+_Optional, Default="localhost:8126"_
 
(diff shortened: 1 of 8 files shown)


der-eismann:
@rtribotte Since a month is over since you added the labels - is there anything I can do to speed up the review? This feature is very important for us.

ldez (approved):
LGTM

PR #9740, opened 2023-02-23
Native Kubernetes service load-balancing

### LINKED ISSUE #9655: Allow to bypass Traefik load-balacing to use native Kubernetes load-balacing
### Welcome!

Today, Traefik is only forwarding incoming traffic to [service pods](https://community.traefik.io/t/communication-between-traefik-and-pods).

In some scenarios, it is needed to bypass Traefik load-balancing to use native Kubernetes load-balancing (through k8s services).

For example, when using service meshes to handle the cluster traffic from the Ingress to the pods, it is needed to forward traffic to meshed Kubernetes services (see https://community.traefik.io/t/communication-between-traefik-and-pods). 

This behavior could be configured at the `Service` level with a k8s annotation, or with a property when configuring an `IngressRoute` service.

rtribotte:
### What does this PR do?

This PR introduces a new option for Ingresses (service annotation) and IngressRoutes to control, when creating the load-balancer, whether the LB's children are directly the pods IPs or if the only child is the Kubernetes Service clusterIP.
It allows relying on the Kubernetes Service itself to load-balance to the pods.

### Motivation

Fixes #9655

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
+++ b/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
@@ -104,6 +104,13 @@ spec:
                             description: Namespace defines the namespace of the referenced
                               Kubernetes Service or TraefikService.
                             type: string
+                          nativeLB:
+                            description: NativeLB controls, when creating the load-balancer,
+                              whether the LB's children are directly the pods IPs
+                              or if the only child is the Kubernetes Service clusterIP.
+                              The Kubernetes Service itself does load-balance to the
+                              pods. By default, NativeLB is false.
+                            type: boolean
                           passHostHeader:
                             description: PassHostHeader defines whether the client
                               Host header is forwarded to the upstream Kubernetes
@@ -357,6 +364,13 @@ spec:
                             description: Namespace defines the namespace of the referenced
                               Kubernetes Service.
                             type: string
+                          nativeLB:
+                            description: NativeLB controls, when creating the load-balancer,
+                              whether the LB's children are directly the pods IPs
+                              or if the only child is the Kubernetes Service clusterIP.
+                              The Kubernetes Service itself does load-balance to the
+                              pods. By default, NativeLB is false.
+                            type: boolean
                           port:
                             anyOf:
                             - type: integer
@@ -542,6 +556,13 @@ spec:
                             description: Namespace defines the namespace of the referenced
                               Kubernetes Service.
                             type: string
+                          nativeLB:
+                            description: NativeLB controls, when creating the load-balancer,
+                              whether the LB's children are directly the pods IPs
+                              or if the only child is the Kubernetes Service clusterIP.
+                              The Kubernetes Service itself does load-balance to the
+                              pods. By default, NativeLB is false.
+                            type: boolean
                           port:
                             anyOf:
                             - type: integer
@@ -827,6 +848,13 @@ spec:
                         description: Namespace defines the namespace of the referenced
                           Kubernetes Service or TraefikService.
                         type: string
+                      nativeLB:
+                        description: NativeLB controls, when creating the load-balancer,
+                          whether the LB's children are directly the pods IPs or if
+                          the only child is the Kubernetes Service clusterIP. The
+                          Kubernetes Service itself does load-balance to the pods.
+                          By default, NativeLB is false.
+                        type: boolean
                       passHostHeader:
                         description: PassHostHeader defines whether the client Host
                           header is forwarded to the upstream Kubernetes Service.
@@ -1981,6 +2009,13 @@ spec:
                           description: Namespace defines the namespace of the referenced
                             Kubernetes Service or TraefikService.
                           type: string
+                        nativeLB:
+                          description: NativeLB controls, when creating the load-balancer,
+                            whether the LB's children are directly the pods IPs or
+                            if the only child is the Kubernetes Service clusterIP.
+                            The Kubernetes Service itself does load-balance to the
+                            pods. By default, NativeLB is false.
+                          type: boolean
                         passHostHeader:
                           description: PassHostHeader defines whether the client Host
                             header is forwarded to the upstream Kubernetes Service.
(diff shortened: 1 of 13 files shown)


pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -424,6 +426,18 @@ func getServicePort(svc *corev1.Service, port intstr.IntOrString) (*corev1.Servi
...
+func getNativeServiceAddress(service corev1.Service, svcPort corev1.ServicePort) (string, error) {
+	if service.Spec.ClusterIP == "None" {
+		return "", fmt.Errorf("no clusterIP on headless service: %s/%s", service.Namespace, service.Name)
+	}
+
+	if service.Spec.ClusterIP == "" {
+		return "", fmt.Errorf("no clusterIP found for service: %s/%s", service.Namespace, service.Name)
+	}
+
+	return net.JoinHostPort(service.Spec.ClusterIP, strconv.Itoa(int(svcPort.Port))), nil
```
mpl, line 438: there should at least be a comment explaining why this strconv.Itoa is guaranteed to give us a valid value (so we don't wonder about it next time), or we should check the error returned.

pkg/provider/kubernetes/crd/kubernetes_tcp.go
```diff
@@ -226,6 +226,15 @@ func (p *Provider) loadTCPServers(client Client, namespace string, svc v1alpha1.
 		return nil, err
 	}
 
+	if svc.NativeLB {
+		address, err := getNativeServiceAddress(*service, *svcPort)
+		if err != nil {
+			return nil, fmt.Errorf("native Kubernetes Service load-balancing: %w", err)
```
mpl, line 232:
```suggestion
			return nil, fmt.Errorf("getting native Kubernetes Service address: %w", err)
```

pkg/provider/kubernetes/crd/kubernetes_udp.go
```diff
@@ -120,6 +120,15 @@ func (p *Provider) loadUDPServers(client Client, namespace string, svc v1alpha1.
 		return nil, err
 	}
 
+	if svc.NativeLB {
+		address, err := getNativeServiceAddress(*service, *svcPort)
+		if err != nil {
+			return nil, fmt.Errorf("native Kubernetes Service load-balancing: %w", err)
```
mpl, line 126:
```suggestion
			return nil, fmt.Errorf("getting native Kubernetes Service address: %w", err)
```

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -538,6 +538,21 @@ func (p *Provider) loadService(client Client, namespace string, backend networki
 		if svcConfig.Service.ServersTransport != "" {
 			svc.LoadBalancer.ServersTransport = svcConfig.Service.ServersTransport
 		}
+
+		if svcConfig.Service.NativeLB {
+			address, err := getNativeServiceAddress(*service, portSpec)
+			if err != nil {
+				return nil, fmt.Errorf("native Kubernetes Service load-balancing: %w", err)
```
mpl, line 545:
```suggestion
				return nil, fmt.Errorf("getting native Kubernetes Service address: %w", err)
```

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -587,6 +602,18 @@ func (p *Provider) loadService(client Client, namespace string, backend networki
...
 
+func getNativeServiceAddress(service corev1.Service, svcPort corev1.ServicePort) (string, error) {
+	if service.Spec.ClusterIP == "None" {
+		return "", fmt.Errorf("no clusterIP on headless service: %s/%s", service.Namespace, service.Name)
+	}
+
+	if service.Spec.ClusterIP == "" {
+		return "", fmt.Errorf("no clusterIP found for service: %s/%s", service.Namespace, service.Name)
+	}
+
```
mpl, line 613: Same remark as in the other provider

juliens (approved):
LGTM

docs/content/routing/providers/kubernetes-ingress.md
```diff
@@ -299,6 +299,17 @@ which in turn will create the resulting routers, services, handlers, etc.
 
 #### On Service
 
+??? info "`traefik.ingress.kubernetes.io/service.nativelb`"
+
+    Controls, when creating the load-balancer, whether the LB's children are directly the pods IPs or if the only child is the Kubernetes Service clusterIP.
+    The Kubernetes Service itself does load-balance to the pods.
+    Please note that the connection reuse with services (connection pool) can avoid to load-balance requests to pods.
```
nmengin, line 306:
```suggestion
    Please note that, by default, Traefik reuses the established connections to the backends for performance purposes. This can prevent the requests load balancing between the replicas from behaving as one would expect when the option is set.
```

docs/content/routing/providers/kubernetes-crd.md
```diff
@@ -639,6 +641,47 @@ More information in the dedicated server [load balancing](../services/index.md#l
         task: app2
     ```
 
+!!! important "Kubernetes Service Native Load-Balancing"
+
+    To avoid creating the server load-balancer with the pods IPs and use Kubernetes Service clusterIP directly,
+    one should set the service `NativeLB` option to true.
+    Please note that the connection reuse with services (connection pool) can avoid to load-balance requests to pods.
```
nmengin, line 648:
```suggestion
    Please note that, by default, Traefik reuses the established connections to the backends for performance purposes. This can prevent the requests load balancing between the replicas from behaving as one would expect when the option is set.
```

nmengin (changes requested):

nmengin (approved):
LGTM 👏 👏

mpl (approved):

prajithp13:
Can we have the option to set this default for all the ingress routes?  something like providers.kubernetescrd.useNativeLoadBalancer=true

prajithp13:
Can we have the option to set this default for all the ingressroutes?  something like 
```
providers.kubernetescrd.useNativeLoadBalancer=true
```

tfny:
Hey @prajithp13, We are interested in your idea to set this as default for all ingressroutes. Could you please make a separate issue for that so we can track it?

PR #9758, opened 2023-03-03
Fix default configuration settings for Nomad Provider

aofei:
### What does this PR do?

Follow the docs below to fix the default configuration settings for Nomad Provider.

https://github.com/traefik/traefik/blob/20e47d91026fa3cdd8277841af75e6b2ca817974/pkg/provider/nomad/nomad.go#L109-L115

### Motivation

There is no code anywhere in the `github.com/traefik/traefik/v3/pkg/provider/nomad` package that calls [`github.com/hashicorp/nomad/api.DefaultConfig`](https://pkg.go.dev/github.com/hashicorp/nomad/api#DefaultConfig), and the [`github.com/hashicorp/nomad/api.NewClient`](https://pkg.go.dev/github.com/hashicorp/nomad/api#NewClient) doesn't implicitly do that for us, so we have to set those defaults ourselves .

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/reference/static-configuration/cli-ref.md
+++ b/docs/content/reference/static-configuration/cli-ref.md
@@ -775,7 +775,7 @@ Constraints is an expression that Traefik matches against the Nomad service's ta
 Default rule. (Default: ```Host(`{{ normalize .Name }}`)```)
 
 `--providers.nomad.endpoint.address`:  
-The address of the Nomad server, including scheme and port.
+The address of the Nomad server, including scheme and port. (Default: ```http://127.0.0.1:4646```)
 
 `--providers.nomad.endpoint.endpointwaittime`:  
 WaitTime limits how long a Watch will block. If not provided, the agent default values will be used (Default: ```0```)
--- a/docs/content/reference/static-configuration/env-ref.md
+++ b/docs/content/reference/static-configuration/env-ref.md
@@ -775,7 +775,7 @@ Constraints is an expression that Traefik matches against the Nomad service's ta
 Default rule. (Default: ```Host(`{{ normalize .Name }}`)```)
 
 `TRAEFIK_PROVIDERS_NOMAD_ENDPOINT_ADDRESS`:  
-The address of the Nomad server, including scheme and port.
+The address of the Nomad server, including scheme and port. (Default: ```http://127.0.0.1:4646```)
 
 `TRAEFIK_PROVIDERS_NOMAD_ENDPOINT_ENDPOINTWAITTIME`:  
 WaitTime limits how long a Watch will block. If not provided, the agent default values will be used (Default: ```0```)
--- a/pkg/provider/nomad/nomad.go
+++ b/pkg/provider/nomad/nomad.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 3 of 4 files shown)


aofei:
The failed job https://github.com/traefik/traefik/actions/runs/4321085853/jobs/7541952997 doesn't seem to be related. And maybe it can be passed by just re-run.

rtribotte:
Hello @aofei,
 
The defaulting is handled directly when creating the client: https://github.com/hashicorp/nomad/blob/fc40c491cacec3d8ec3f2f98cd82b9068a50797c/api/api.go#L461-L463.

So it is not necessary to handle it as your changes suggested.

Accordingly, we are declining the PR.

aofei:
Hi @rtribotte. That's where the problem is as I said. As you can see, `NewClient` actually only uses `defConfig.Address`, which means it ignores `NOMAD_TOKEN` and the like. The reason I opened this PR is because Traefik failed to read my `NOMAD_TOKEN`.

https://github.com/hashicorp/nomad/blob/fc40c491cacec3d8ec3f2f98cd82b9068a50797c/api/api.go#L461-L484

aofei:
@rtribotte

And FYI, the entire `github.com/hashicorp/nomad/api` package only calls the `DefaultConfig` func once, which is in the `NewClient` you mentioned. (See https://github.com/search?q=repo%3Ahashicorp%2Fnomad+path%3Aapi%2F+NOT+path%3A*_test.go+DefaultConfig&type=code)

Nomad provides `DefaultConfig` for users to call explicitly. `NewClient` does not use all of its values implicitly, as for using the `defConfig.Address` I believe it is only for backward compatibility. So, Traefik should either call `DefaultConfig` explicitly, or set those defaults manually like this PR does.

rtribotte:
Hello @aofei,

You are right. The call to `DefaultConfig` is only done to get the address.

We are reopening the PR, thus it seems indeed more appropriate to call `DefaultConfig` rather than reading the env variables.

aofei:
Hi @rtribotte,

This PR is ready to be reviewed. But before that I have one more question, should we take other defaults besides `NOMAD_ADDR`, `NOMAD_REGION`, `NOMAD_TOKEN`? Given that `api.DefaultConfig` also has some defaults related to TLS configuration that can be used for `nomad.EndpointConfig.TLS`.

https://github.com/hashicorp/nomad/blob/fc40c491cacec3d8ec3f2f98cd82b9068a50797c/api/api.go#L319-L339

rtribotte:
Hello @aofei,

Thanks for making the changes!
> But before that I have one more question, should we take other defaults besides NOMAD_ADDR, NOMAD_REGION, NOMAD_TOKEN? Given that api.DefaultConfig also has some defaults related to TLS configuration that can be used for nomad.EndpointConfig.TLS.

Yes, for the current four TLS options supported (https://github.com/traefik/traefik/blob/v3.0/pkg/provider/nomad/nomad.go#L220-L225), it would make sense 👍.

rtribotte:
Hello again @aofei,

I forgot to ask previously but can you please also rebase the PR to target the branch v2.9?

Thanks!

aofei:
Hi @rtribotte. This PR is ready to be reviewed.

Also, these changes are for the `v3.0` branch. I don't think I should rebase my branch directly to `v2.9`, given that `v2.9` and `v3.0` have a big difference in nomad provider codebase (https://github.com/traefik/traefik/compare/c4cc30c..044dc6a#diff-176aae5b7e87cad7507ec6ed0429b6bcbafb613993b9e1cd617dfedde3371d95). Rebasing it will cause conflicts.

But I can open another PR if you think `v2.9` also needs to be fixed.

ldez:
What @rtribotte asked you, is to "recreate" (rebase) your PR on v2.9.

If your PR target v2.9, after the merge we will manage the backport to v3.0.

aofei:
Hi @ldez. I just recreated the commit based on the latest v2.9.

ldez (approved):
LGTM

rtribotte (approved):
Thanks 👍

PR #9763, opened 2023-03-07
Bump hashicorp consul

dlorenc:
…SA-m69r-9g56-7mv8.

This is the output of:
go get go.mongodb.org/mongo-driver@v1.11.2
go get github.com/Azure/go-autorest/autorest/adal@v0.9.22 go get github.com/hashicorp/go-discover@v0.0.0-20230125175744-70b45b3eb8b7 go get github.com/hashicorp/consul@v1.15.0
go mod tidy

### More

- [ ] Added/updated tests


ldez:
Hello,

we are not affected by those CVEs because they affect pieces of code that we don't use.
The Consul CVE is about the Consul server but we are only using the client.

Also, the problem related to updating a dependency because of a false positive is the impact of transitive dependencies. For example, an update of Consul can produce an update of gRPC, but gRPC is known to break things between patch/minor versions.
Any update has side effects.

So I will decline your PR.

PR #9789, opened 2023-03-20
feat: use env variable in github actions

mmatur:
### Description

Use env variable in Github Actions instead of secrets


ldez (approved):
LGTM

PR #9795, opened 2023-03-21
fix nomad client defaults

rtribotte:
### What does this PR do?

This PR fixes the Nomad client defaults changes introduced by #9758 in respect of the TLS configuration.

### More

- [x] Added/updated tests


ldez (approved):
LGTM

PR #9810, opened 2023-03-29
Update headers.md

yingshaoxo:
add `accessControlAllowHeaders: "*"` as an example for accept all kinds of headers.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/middlewares/http/headers.md
+++ b/docs/content/middlewares/http/headers.md
@@ -212,6 +212,7 @@ instead the response will be generated and sent back to the client directly.
 ```yaml tab="Docker"
 labels:
   - "traefik.http.middlewares.testheader.headers.accesscontrolallowmethods=GET,OPTIONS,PUT"
+  - "traefik.http.middlewares.testheader.headers.accesscontrolallowheaders=*"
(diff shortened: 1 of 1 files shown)


rtribotte:
Hello @yingshaoxo,

Thanks for your contribution!

After taking a quick look at the changes, we think that it would be even clearer to add documentation for the wildcard in the documentation section of the `accessControlAllowHeaders` option itself (https://doc.traefik.io/traefik/v2.9/middlewares/http/headers/#accesscontrolallowheaders). 

Could you please address those changes?

yingshaoxo:
> Hello @yingshaoxo,
> 
> 
> After taking a quick look at the changes, we think that it would be even clearer to add documentation for the wildcard in the documentation section of the `accessControlAllowHeaders` option itself (https://doc.traefik.io/traefik/v2.9/middlewares/http/headers/#accesscontrolallowheaders).
> 
> Could you please address those changes?

No, it's not.

Because, if I do not add `accessControlAllowHeaders` to the topest example, I wouldn't even know there has an option called `accessControlAllowHeaders`.

So even if I set all options to * according to the top example, it won't successfully allow CROS, which would be useless.

mpl (approved):

ldez (approved):

PR #9822, opened 2023-04-02
Display period setting of the RateLimit middleware in the webui

smatyas:
This PR adds a minor enhancement to the webui:

It makes the "period" setting of the RateLimit middleware to be displayed on the middleware view page. It is as important as the other settings of the middleware, so it is equally important to display.

The data was already there in the API, so it is just a vue file modification.

----

Before:

Screenshot_20230402_145317

After:

Screenshot_20230402_150935

Test config:

```
[http.middlewares]
  [http.middlewares.test-ratelimit.rateLimit]
    average = 1000
    burst = 200
    period = 330
```

### DIFF
--- a/webui/src/components/_commons/PanelMiddlewares.vue
+++ b/webui/src/components/_commons/PanelMiddlewares.vue
@@ -723,7 +723,7 @@
             </div>
           </q-card-section>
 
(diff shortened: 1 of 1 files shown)


ldez (approved):
LGTM

PR #9832, opened 2023-04-09
Allow short healthcheck interval with long timeout

### LINKED ISSUE #9812: Feature: allow healthcheck interval shorter than timeout
### Welcome!

According to the documentation, a service healthcheck timeout must be less than the healthcheck interval. In Traefik 2.x, a shorter interval [generates a warning](https://github.com/traefik/traefik/blob/v2.9.9/pkg/server/service/service.go#L354-L356), but is permitted. However in 3.x, the restriction [is enforced](https://github.com/traefik/traefik/blob/master/pkg/healthcheck/healthcheck.go#L74).

I have an application where I'd like to use a frequent healthcheck (1s), but also allow for the case where that healthcheck may _occasionally_ take longer than 1s to respond. I'd rather not decrease the check frequency, but I also don't want a slow healthcheck to be treated as failing. I think this is a valid approach.

The behaviour is working well on 2.x, but it goes against what the documentation (and warning) states. And as mentioned, it looks like this will stop working in 3.x.

Could we remove this restriction?

kevinmcconnell:
### What does this PR do?

Removes the restriction on healthchecks that requires the check interval to be longer than its timeout. Fixes #9812.

### Motivation

I'd like to configure frequent healthchecks so that changes to health status are discovered quickly. However this currently requires the timeout to be short too. If a check endpoint is occassionally slow to respond, this can lead to false negatives. Ideally a longer timeout is allowed even when the interval is short.

### More

- I've updated the documentation to remove mention of the old restriction.
- As this is a change to the previously documented behaviour, I've also added some basic test coverage for how these timeout & interval limits are handled.

### DIFF
--- a/docs/content/routing/services/index.md
+++ b/docs/content/routing/services/index.md
@@ -338,7 +338,6 @@ Below are the available options for the health check mechanism:
 !!! info "Interval & Timeout Format"
 
     Interval and timeout are to be given in a format understood by [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration).
-    The interval must be greater than the timeout. If configuration doesn't reflect this, the interval will be set to timeout + 1 second.
 
 !!! info "Recovering Servers"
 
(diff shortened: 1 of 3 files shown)


rtribotte (approved):
Thanks!

ldez (approved):
LGTM

mmatur (approved):
LGTM

Issue #9884, opened 2023-05-03
HTTPRouteStatus resource is not being updated


romantomjak:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I'm trying to make it so that [ExternalDNS](https://github.com/kubernetes-sigs/external-dns) updates DNS records from the [HTTPRoute resource](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPRoute), but all I get in ExternalDNS logs is:
```
No endpoints could be generated from HTTPRoute monitoring/prometheus
```

I suspect it's because there is no address assigned to the Gateway and it cannot figure out what IP address to use for the DNS record:

```sh
$ kubectl -n gateway-ingress get gateway public-gw
NAME        CLASS     ADDRESS   PROGRAMMED   AGE
public-gw   traefik                          107m
```

### What did you see instead?

When Traefik updates the gateway status, it [nullifies the address field](https://github.com/traefik/traefik/blob/619045eb4bbe842be550fd16d67dc7c8eb61ebe8/pkg/provider/kubernetes/gateway/kubernetes.go#L513).

Ideally, Traefik would grab the assigned IP/Hostname from the exposed `type: LoadBalancer` service and then update the gateway with those addresses, but that may be a substantial amount of work.

Alternatively, Traefik could perhaps just copy the addresses from the Gateway spec (if they are set)?
```yaml
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: Gateway
metadata:
  name: public-gw
  namespace: gateway-ingress
spec:
  gatewayClassName: traefik
  addresses:
    - type: IPAddress
      value: a.b.c.d 
  listeners:
    ...
```

At minimum, it should at least stop nullifying it for now so that you can at least `kubectl edit` it and set it manually.

### What version of Traefik are you using?

2.10.1

### What is your environment & configuration?

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: traefik
  namespace: gateway-ingress
spec:
  selector:
    matchLabels:
      app: traefik-lb
  template:
    metadata:
      labels:
        app: traefik-lb
    spec:
      serviceAccountName: traefik-controller
      containers:
        - name: traefik
          image: traefik:v2.10.1
          args:
[...]

PR #9892, opened 2023-05-14
Update HTTP/TLS/TCPRoute status

romantomjak:
### What does this PR do?

This PR implements updating status resource for HTTP/TLS/TCPRoute objects, so that they appear attached to the gateway from the API perspective.

### Motivation

For background see #9884

### More

- [x] Added/updated tests

### Additional Notes

n/a


pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -1823,3 +1879,109 @@ func makeListenerKey(l gatev1alpha2.Listener) string {
...
+
+		routeParentStatus.Conditions = append(routeParentStatus.Conditions, metav1.Condition{
+			Type:               string(gatev1alpha2.ConditionRouteAccepted),
+			Status:             metav1.ConditionTrue,
+			ObservedGeneration: routeMeta.Generation,
+			LastTransitionTime: metav1.NewTime(timeNow()),
+			Reason:             string(gatev1alpha2.ConditionRouteAccepted),
+			Message:            "The route was attached to the Gateway",
+		})
+	}
```
rtribotte, line 1984: In the [`RouteParentStatus` documentation](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.RouteParentStatus), and for the `conditions` field, it is said:
```
If the Route’s ParentRef specifies an existing Gateway that supports Routes of this kind AND that Gateway’s controller has sufficient access, then that Gateway’s controller MUST set the “Accepted” condition on the Route, to indicate whether the route has been accepted or rejected by the Gateway, and why.
```

According to this, I think the condition `ConditionRouteAccepted` should not always be true, it should also inform if the attachment failed and probably give a reason.
But it also says that there are cases where it is not necessary:
```
There are a number of cases where the “Accepted” condition may not be set due to lack of controller visibility, that includes when:

- The Route refers to a non-existent parent.
- The Route is of a type that the controller does not support.
- The Route is in a namespace the controller does not have access to.
```

Here, if I didn’t overlook something, it seems that we would never update the RouteStatus with a status false.
Also, while we should prevent overriding another Gateway/Controller `RouteParentStatus`, I'm not sure we do need to keep a "state" and update it.

rtribotte (commented):
Hey @romantomjak,

Thanks for the recent changes!
At a glance, I have a few questions regarding the current implementation and the spec.

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -1823,3 +1879,109 @@ func makeListenerKey(l gatev1alpha2.Listener) string {
...
+		kind := gatev1alpha2.Kind(kindGateway)
+		namespace := gatev1alpha2.Namespace(gateway.Namespace)
+
+		// Only set the section name if the route requests it
+		var sectionName *gatev1alpha2.SectionName
+		for _, p := range routeSpec.ParentRefs {
+			if p.SectionName != nil && *p.SectionName == listener.Name {
+				sectionName = p.SectionName
+			}
+		}
```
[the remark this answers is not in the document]
romantomjak: > I think the current approach would not allow specifying an empty string to yield that the Route has been attached to all listeners of a Gateway.

If i understood the spec correctly then I think it's still possible to specify an empty string. The motivation here was to set the section name explicitly on the HTTPRouteStatus object if the HTTProute spec requests a particular gateway listener object.

If the HTTPRoute does not care about which listener it gets attached to, the section name is left empty in the manifest file which means that the route can be attached to any listener and so the route parent status object would have an empty string there too if that makes sense?

Here's an example of what I'm thinking:
```yaml
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: Gateway
metadata:
  name: public-gw
  namespace: gateway-ingress
spec:
  gatewayClassName: traefik
  listeners:
    - name: http
      protocol: HTTP
      port: 80
      ...
    - name: https # <-- this is the section name
      protocol: HTTPS
      port: 443
      ...
---
apiVersion: gateway.networking.k8s.io/v1alpha2
kind: HTTPRoute
metadata:
  name: prometheus
  namespace: monitoring
spec:
  parentRefs:
    - name: public-gw
      namespace: gateway-ingress
      sectionName: https # <-- only attach the route to HTTPs listener. if left empty it'll get
                         # attached to both - port 80 and port 443 and the corresponding
                         # status object will have an empty section name too
  rules: ...
```

> the case of a Gateway having 3 Listeners, if the route is attached only to two listeners, the empty string value doesn’t seem appropriate but a single sectionName is not good as well.

I think that's how the spec wants it. If you want a particular listener you must specify that in the HTTPRoute object (the sectioname). Otherwise it'll just get attached to whatever was matched.

pkg/provider/kubernetes/gateway/kubernetes.go
romantomjak: Good point! I'll have a look if I can add this with minimal blast radius 😅

romantomjak:
Sorry for radio silence. I became a father recently, so haven't found time to get back to this 🐣

pkg/provider/kubernetes/gateway/kubernetes.go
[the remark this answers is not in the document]
rtribotte: Sure, I think we are on the same line regarding the attachment.
My question was more about the `RouteStatus` update.
As of today, the place where it is built doesn't allow taking into account the multiple listener attachment, and hence, specifying an empty string.
If I'm not overlooking this, to do that, it is needed to build the `RouteStatus` after every listener has been processed.

Also, it is still unclear to me if multiple [RouteParentStatus](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.RouteParentStatus) can be added for the same Gateway in a RouteStatus.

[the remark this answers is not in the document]
romantomjak: I've read the spec more times than I'd like to admit, but I think the current implementation is accurate. The status object is from the perspective of the route. If the route spec has an empty `sectionName` it means it can be attached to *any* listener when deciding whether to allow attaching the route to the gateway. I don't think we need to update the status object with the listener that accepted it. It should be set to whatever route requested because that is the object it is referencing.

The same goes for when the route spec has set a `sectionName` and/or `port`. Both values must match in order for the route to be considered successfully attached to the gateway in which case the status object will also contain the `sectionName` and/or `port` that the *route* requested.

pkg/provider/kubernetes/gateway/kubernetes.go
romantomjak: I'm collecting route status conditions now and passing them through to the status maker function. I haven't yet finished the work on that yet, but perhaps you can take a look to see if the approach generally makes sense so far?

romantomjak: > not sure we do need to keep a "state" and update it.

Sorry, I completely missed this ☝️  Do you mean that we shouldn't search for existing route parent object when updating the status? I've done it because [RouteStatus](https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io%2fv1beta1.RouteStatus) docs have this paragraph:
> When this route attaches to a parent, the controller that manages the parent must add an entry to this list when the controller first sees the route and should update the entry as appropriate when the route or gateway is modified.

rtribotte:
Hello @romantomjak,

Thanks for the recent work on this.

I discussed the PR with other maintainers, and we have some feedback.

Regarding the ParentRef sectionName discussion: we agree with your stand, we should specify the sectionName in the status only if the route requires it.

Concerning the route status conditions update, by updating the route status for each listener independently it will still remain as creating a sort of "state".
For instance, if a listener is dropped from a gateway definition, it won't be removed from the status.
There is also an induced complexity due to the update operation per listener, versus a complete rebuild of the status related to the controller with one single update operation that would replace any previous status for that controller.

We have also some concerns about the bump of gateway-api, because `gateway.networking.k8s.io/v1beta1` is released a year now ([v0.5.0](https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.5.0)), and we have in mind to provide the support of it.
Because it may have impacts on the current provider design, we think that it would be better to address the status route update on top of the bump to the latest version ([v0.7.1](https://github.com/kubernetes-sigs/gateway-api/releases/tag/v0.7.1)) of the gateway API spec and the support of `gateway.networking.k8s.io/v1beta1`.
This would give the opportunity to take into account any clarification in the spec related to status updates, if there are, or to build status updates on top of any rework of the provider if the bump requires it.

As a side concern, it would have been better for the review process to address the listener status update independently from the route status update, or even to address each route type status update.

To sum up, we think that it would be better to bump gateway-api before addressing the route status update, and hence wait for it before moving on this PR.

Is there any chance that you'll be willing to address that bump?

ps: For both the bump and the future of this PR, the changes should target the v3 branch.

romantomjak:
Ah, perfect! Thanks for the feedback - all suggestions seem reasonable and I'll gladly pick up the version bump. I couldn't find an issue for the version bump, shall I open one to chat through any details before starting the work?

What do we do with this PR? I reckon I'll close it for now and open a new one after bumping the API version work is finished?

rtribotte:
Hello @romantomjak,

> Ah, perfect! Thanks for the feedback - all suggestions seem reasonable and I'll gladly pick up the version bump. I couldn't find an issue for the version bump, shall I open one to chat through any details before starting the work?

Good news!
I have found this one: https://github.com/traefik/traefik/issues/9476, which is a proposal for the bump to APIVerison `v1beta1`.

I think there is no need to discuss it further, as long as the scope is limited. I would say that bumping the dependency version and moving from APIVersion `v1alpha1` to `v1beta1`, without widening the current feature support is enough for one PR.

> What do we do with this PR? I reckon I'll close it for now and open a new one after bumping the API version work is finished?

Yes, closing it is probably better, and it would also give it a "fresh start" on its reopening. 

Thanks!

PR #9910, opened 2023-05-25
feat: add request header to show url passing to service

VladislavKravitski4888:
### What does this PR do?

In the replace path and replace path regex middlewares we defined a new header that contains the X-Service-Path. 

### Motivation

For debugging purpose is really helpful to see if your expected path to service is correct.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/replacepath/replace_path.go
+++ b/pkg/middlewares/replacepath/replace_path.go
@@ -14,7 +14,9 @@ import (
 const (
 	// ReplacedPathHeader is the default header to set the old path to.
 	ReplacedPathHeader = "X-Replaced-Path"
-	typeName           = "ReplacePath"
+	// ServicePathHeader is the destination path set by service.
+	ServicePathHeader = "X-Service-Path"
+	typeName          = "ReplacePath"
 )
 
 // ReplacePath is a middleware used to replace the path of a URL request.
@@ -56,6 +58,7 @@ func (r *replacePath) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	req.Header.Add(ServicePathHeader, req.URL.Path)
 	req.RequestURI = req.URL.RequestURI()
 
 	r.next.ServeHTTP(rw, req)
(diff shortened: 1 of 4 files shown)


ottowagner1337 (commented):
LGTM

nmengin:
Hello @VladislavKravitski4888,

We'd like to better understand your needs and motivation.
Could you associate it with an issue to help us evaluate the use case, please?
This will also allow other users to upvote and expand on their needs too.

Thanks in advance.

VladislavKravitski4888:
Hello @nmengin,

e.g. we tried to use the replacePathRegex middleware like regex: “^/whoami/(.*)“, replacement: "/whoami/bli/bla/blub/$1“. As we could see, there was a header named X-Replaced-Path containing original request url but no request header to show what url would go to a service. For consistent reasons it’s for our aim better to see which both urls coming at a service were. It were fine for us to see both information in logging but it is not the case at the moment. In the log files you can only see the original request path but no url going to a service. Like … "RequestMethod":"GET","RequestPath":"/whoami/bla","RequestPort":"8443","RequestProtocol":"HTTP/1.1","RequestScheme":"http","RetryAttempts":0,"RouterName":"my-router@file","ServiceAddr":"apigateway_demo:8080","ServiceName":"service-whoami@file","ServiceURL":{"Scheme":"http","Opaque":"","User":null,"Host":"apigateway_demo:8080","Path":"","RawPath":"","OmitHost":false,"ForceQuery":false,"RawQuery":"","Fragment":"","RawFragment“:““},…

rtribotte:
Hello @VladislavKravitski4888,

Thanks for the contribution!

We discussed the opportunity of the changes proposed in this PR, and we don't think that Traefik should add a header for debugging purposes only.
For this reason, we are declining this PR. Nonetheless, if another need for this is raised in the future, we are open to reconsidering it.

As @nmengin mentioned in his previous comment, we could also consider adding a DEBUG log for that purpose.

Additionally, we think that your use case fits with the plugin mechanism, which you could leverage to build your own plugin to add the header containing the service path to the request.

PR #9914, opened 2023-05-26
Fix multiple subsets endpoint

joaosilva15:
### What does this PR do?

Currently, in traefik ingress routes, if an endpoint has more than one
subset and the first one does not contain the port target in the ingress
route traefik will throw an error. This PR searches all subsets before
giving up

PR is equal to https://github.com/traefik/traefik/pull/7997/files but
for a different provider

### Motivation

Traefik was unable to find a port for a service with multiple ports defined

### More

- [x] Added/updated tests


kevinpollet (approved):
Thanks 👍

nmengin (approved):
LGTM

PR #9921, opened 2023-05-29
Embed and plugin in code support, upgrade opentelemetry

coyzeng:
### What does this PR do?

1、Support TCP/HTTP Middleware programing in code and embed deploy. 
2、Support TCP/TLS client hello intercept for tcp handshake enhancement.
3、Support net.Dialer intercept for connection create enhancement.
4、Support  LoadBalancer programing in code and embed deploy. 
5、Support AccessLog writer override .
6、Support Transport proxy customized instead of use global environment variable.
7、Upgrade opentelemetry to latest stable version.

Middleware can be plugged like
```go
import (
	"context"
	"fmt"
	"github.com/traefik/traefik/v3/pkg/server/middleware"
	"net/http"
	"strings"
	"time"
)

func init() {
	var _ http.Handler = new(barrier)
	var _ prsim.Listener = barriers
	middleware.Provide(barriers)
}

var barriers = &barrierMiddleware{routes: map[string]*types.Route{}}

type barrier struct {
	name     string
	next     http.Handler
        disabled bool
}

func (that *barrier) ServeHTTP(writer http.ResponseWriter, request *http.Request) {
	if !that.disabled {
                that.next.ServeHTTP(writer, request)
                return
        }
        writer.WriteHeader(http.StatusForbidden)
	_, err := writer.Write([]byte(cause.NetDisable.Message))
	if nil != err {
		log.Error(err)
	}
}

type barrierMiddleware struct {
}

func (that *barrierMiddleware) Name() string {
	return fmt.Sprintf("barrier@%s", ProviderName)
}

func (that *barrierMiddleware) Priority() int {
	return 0
}

func (that *barrierMiddleware) Scope() int {
	return 1
}

func (that *barrierMiddleware) New(ctx context.Context, next http.Handler, name string) (http.Handler, error) {
	return &barrier{next: next, name: name, disabled: true}, nil
}
```

LoadBalancer can be plugged like
```go
import (
	"github.com/traefik/traefik/v3/pkg/server/service/loadbalancer/wrr"
	"net/http"
)

func init() {
	wrr.Provide(new(SubsetLBStrategy))
}

type SubsetLBStrategy struct {
}

func (that *SubsetLBStrategy) Name() string {
	return "subset@mesh"
}

func (that *SubsetLBStrategy) Priority() int {
	return 1
}

func (that *SubsetLBStrategy) Next(w http.ResponseWriter, req *http.Request, servers []wrr.Server) []wrr.Server {
	subset := prsim.MeshSubset.GetHeader(req.Header)
	if "" == subset {
		return that.skipServerNoSubset(servers)
	}
	var ss []wrr.Server
	for _, server := range servers {
		if nil == server.URL() {
			continue
		}
		if types.FromURL(server.URL()).GetS() == subset {
			ss = append(ss, server)
		}
	}
	if len(ss) > 0 {
		return ss
	}
	return servers
}

func (that *SubsetLBStrategy) skipServerNoSubset(servers []wrr.Server) []wrr.Server {
	var ss []wrr.Server
	for _, server := range servers {
		if nil == server.URL() {
			continue
		}
		var s = types.FromURL(server.URL()).GetS()
		if len(s) == 0 {
			ss = append(ss, server)
		}
	}
	if len(ss) == 0 {
		return servers
	}
	return ss
}

```

### More

- [ ] Added/updated tests


ldez:
Hello,

There are too many topics in this PR.
You will have to create several PRs.

PR #9961, opened 2023-06-15
updated quick-start-with-kubernetes.md

cherrymu:
### What does this PR do?

This will keep the doc page updated

### Motivation

Made changes to align with hitting localhost at the end of the document

### More

- [ ] Added/updated tests


nmengin:
Hello @cherrymu,

Thanks for submitting this PR.

But we actually need to expose Traefik using `Loadbalancer` services to reach the entrypoints from the outside of the cluster (which is not possible with a `ClusterIP` service).

At the end of the Getting Started, we're using the `localhost` address because we are using a local instance of a Kubernetes cluster.

For that reason, we are declining the changes.

PR #10008, opened 2023-07-04
Fixes the Hub popup content

mdeliatf:
## What does this PR do?

Fixes the description of Traefik Hub in the popup and introduce an option to hide the button.

### Motivation

Show at a glance what Traefik Hub is about. The previous description didn't provide relevant information.
Allow the users to hide the button if they want it.


ldez:
This is not a fix but the addition of an ad inside the UI.
The addition of the option to disable the ad is close to the request to add an option to disable the Hub button but it's not the target of this PR.

Please use a more accurate title.

As a maintainer not employed by Traefik Labs, this commercial modification is something I disagree with.

geraldcroes:
I agree with you; we're on the edge when updating the content of this button in a bugfix, but we believe the description was inaccurate and wanted to fix it.

For the button itself, I totally agree again: It helps Traefik Labs promote its commercial offering (Traefik Hub) to an audience that uses and loves traefik. We want to ensure our supporters know about our commercial products if they ever need them.

This, by itself, should have been the scope of the pull request.

Still, we've heard people complaining about pushing the ad. Our goal is that our users know we provide great commercial products on top of offering great open-source software. These commercial products help us sustain efforts and enhance traefik, which we greatly invested in and want to support even further.

Our initial stance was that we wanted to keep that button on the dashboard with no alternative. But in the end, we realized it defeated our purpose: If people know about our other products, then it's enough, and we don't want our users to be bothered by them (quite the contrary).

Semver encourages us to wait for the next minor version to add the option, but we saw no reason to wait to fix our stance on this button that's not in any critical path of traefik. (We already made such moves in the past with the now unsupported 1.7.x.)

If you still believe we should split the PR and wait for the next minor to add the option, we can, but I see no benefit in waiting.

Let's discuss and settle this issue with other maintainers, finding the best solution for everyone.

ldez:
I didn't suggest splitting the PR because it doesn't make sense.
I suggested using a more accurate title.

ldez:
I disagree with the fact to add an external ad inside the UI.

The title is still not right: it doesn't "fix" the Hub popup content, because the PR converts a static button to a dynamic popup. It's not a fix.

Also the sources of `webui/src/statics/traefiklabs-hub-button-app/main-v1.js` are not available.
https://github.com/traefik/traefik/pull/10008/files#diff-12c8377a8cde2131d8a8c8e0c59c474bc9b2357996446c078a45ba87bcd44344R119

geraldcroes:
> I disagree with the fact to add an externally controlled ad inside the UI.
> The title is still not right: it doesn't "fix" the Hub popup content, because the PR converts a static button to a dynamic popup. It's not a fix.

So I understand you're ok with the ad itself, but you disagree with the fact it's controlled by a dynamic content. Please keep me honest if this is not the case.

If so, to me, it's implementation detail. Our goal was to update the content, and we chose to change the way the content was built to avoid unnecessary releases in the future to do so.

Still, I want to make sure we're on the same page. I encourage you to join the next triage session on discord so we can discuss and find proper solution that will satisfy everyone. If you can't make it there, let's just have an async conversation over there (on the maintainer discord) until we find an agreement.

ldez:
> So I understand you're ok with the ad itself, but you disagree with the fact it's controlled by a dynamic content.

Not really.

> I encourage you to join the next triage session on discord so we can discuss and find proper solution that will satisfy everyone.

PR reviews are not a part of the triage.

I just asked for a more accurate title, I don't understand what is the difficulty here.
A title example: Replace Hub button with a popup ad.

ldez:
I insist on title and sources because: 
- a button with a static link = no security issue
- the execution of remote javascript = security issues

It is essential to have auditable sources and a precise title in order to have an accurate changelog, to allow users to be informed and thus make the right decision.

geraldcroes:
> It is essential to have auditable sources and a precise title in order to have an accurate changelog, to allow users to be informed and thus make the right decision.

I agree with you.

In the initial PR we made the sourcemap available so that anyone could inspect the non minified version of the javascript. Still, there was a risk that if/when we update, this sourcemap would be out of sync.

Now ([2271b7e](https://github.com/traefik/traefik/pull/10008/commits/2271b7ea05d5ac456423fc299ea15b2abca93173)) we've updated to publish the sourcemap of the online file, and we added the path directly in the js so that browsers can auto-discover it and present it to people who want to audit.

I've also updated again the title of the PR so that it's clear we now use a web component for the popup.

ldez:
> In the initial PR we made the sourcemap available so that anyone could inspect the non minified version of the javascript. Still, there was a risk that if/when we update, this sourcemap would be out of sync.

Those kind of files are not sources. So it's not possible to audit the real sources without render then inside a browser.
https://github.com/traefik/traefik/pull/10008/files#diff-12c8377a8cde2131d8a8c8e0c59c474bc9b2357996446c078a45ba87bcd44344R119

I think it's easier to open source the repository of this code instead of adding unused files inside Traefik.

> I've also updated again the title of the PR so that it's clear we now use a web component for the popup.

Before the PR:
webui-before

After the PR:
webui-after

Converting a simple tooltip into an advertisement for Hub can be see as an update of a content but it's on the edge.
Any way, the title is more accurate than before, so we can keep the new title.

mdeliatf:
We open sourced the repository of [traefiklabs-hub-button-app](https://github.com/traefik/traefiklabs-hub-button-app) and added the MIT license!

mdeliatf:
Hi! I added the comment!
Regarding the .js and .map files, we can't remove them because one is the local fallback when the remote counterpart is not available, and since we would like to keep it, in my opinion it makes sense to keep also its source map!

Unrelated, I can see a failed check due to a 500 from nodejs.org! Unfortunately I cannot restart it when it happens!

juliens (approved):
LGTM

mmatur (approved):

geraldcroes (approved):

Issue #10022, opened 2023-07-18
Missing support for Kubernetes Ingress backend resource/apiGroup/kind


GiuseppeChiesa-TomTom:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I'm trying to use the following pattern:

1. Deploy Kubernetes Ingress CRD (with annotations it offers a flawless integration with both external-dns and cert-manager)
2. Set as backend for the ingress rule a `resource` instead typical service. As per [Kubernetes Ingress CRD documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#resource-backend) this should be possible. 

Example:
```yaml
---
apiVersion: traefik.io/v1alpha1                                                                                                                                                               
kind: TraefikService                                                                                                                                                                          
metadata:                                                                                                                                                                                     
  name: echo-server-my-application                                                                                                                                                            
  namespace: default                                                                                                                                                                          
spec:                                                                                                                                                                                         
  weighted:                                                                                                                                                                                   
    services:                                                                                                                                                                                 
    - name: echo-server-my-application-active # existing K8S Service                                                                                                                                             
      port: 80                                                                                                                                                                                
    - name: echo-server-my-application-preview # existing K8S Service                                                                                                                                                
      port: 80                                                                                                                                                                                
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt
    kubernetes.io/ingress.allow-http: 'false'
    traefik.ingress.kubernetes.io/router.entrypoints: websecure
    traefik.ingress.kubernetes.io/router.tls: 'true'
    traefik.ingress.kubernetes.io/router.tls.options: default
  name: echo-server-my-application
  namespace: default
spec:
  ingressClassName: traefik
  rules:
    - host: echo-test.example.com
      http:
        paths:
          - backend:
              resource:
                apiGroup: traefik.io
                kind: TraefikService
                name: echo-server-my-application
            path: /
            pathType: ImplementationSpecific
  tls:
    - hosts:
        - echo-test.example.com
      secretName: sample-app-certificate                  
```

[...]

PR #10023, opened 2023-07-18
fix: avoid panic on resource backends

ldez:
### What does this PR do?

Returns an error instead of panic on the non-supported resource backends.

### Motivation

Avoid panic on resource backends with Ingress.

### More

- [ ] Added/updated tests

### Additional Notes

Related to #10022


kevinpollet (approved):
LGTM

juliens (approved):
LGTM

PR #10025, opened 2023-07-18
fix: integration test with go v1.20.6

mmatur:
### What does this PR do?

This PR fixes the integration test following the patch introduced in go v1.20.6

### More

- [x] Added/updated tests


ldez (approved):
LGTM

Issue #10009, opened 2023-07-04
K8s ingress samesite broken


davhdavh:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

1. set `"traefik.ingress.kubernetes.io/service.sticky.cookie.sameSite"  = "strict"` on service
2. delete cookie in browser
3. refresh page
4. get cookie with no samesite setting and browser warning that samesite setting is missing
Test with "none", or "lax" or "whyyyyyyy?" does nothing, and no error.

### What did you see instead?

![image](https://github.com/traefik/traefik/assets/1201165/6ba53483-0888-4197-834e-08f37edb293d)
![image](https://github.com/traefik/traefik/assets/1201165/087ee154-56ce-414f-bb1a-aa69aa25cdf8)

### What version of Traefik are you using?

3.0.0-beta3

### What is your environment & configuration?

```yaml
apiVersion: v1
kind: Service
metadata:
  annotations:
    traefik.ingress.kubernetes.io/service.sticky.cookie: "true"
    traefik.ingress.kubernetes.io/service.sticky.cookie.httponly: "true"
    traefik.ingress.kubernetes.io/service.sticky.cookie.samesite: strict
    traefik.ingress.kubernetes.io/service.sticky.cookie.secure: "true"
  name: whoami
  namespace: whoami
...
```

nmengin:
Hey @davhdavh,

Thanks for the bug report.
Could you provide a minimal reproducible case (for instance, full Kubernetes manifest to reproduce the issue)?

In the remaining time, if any community member can help us find verified steps to reproduce, we would love the help.

davhdavh:
Any service with ingress with do (I didnt check the traafik crd, since we use external-dns and cert-manager, and they dont play well together)
See the cookie sent using this config
```
    traefik.ingress.kubernetes.io/service.sticky.cookie: "true"
    traefik.ingress.kubernetes.io/service.sticky.cookie.samesite: strict
```
vs this config
```
    traefik.ingress.kubernetes.io/service.sticky.cookie: "true"
    traefik.ingress.kubernetes.io/service.sticky.cookie.samesite: none
```
vs this config
```
    traefik.ingress.kubernetes.io/service.sticky.cookie: "true"
```
[...]

PR #10066, opened 2023-08-07
Fix: set sameSite field for wrr load balancer sticky cookie

sunyakun:
### What does this PR do?

This PR fix issue: #10009 

### Motivation

The load-balancer implementation was changed to traefik/wrr in PR: #9431  but the wrr disregards the sameSite field ([SourceCode](https://github.com/traefik/traefik/blob/f4dc298406cc3874abbb0e9a239ac46659ffbbac/pkg/server/service/loadbalancer/wrr/wrr.go#L59)).

### More

- [ ] Set 'sameSite' field for the sticky cookie
- [ ] Add unit test for sticky cookie

### DIFF
--- a/pkg/server/service/loadbalancer/wrr/wrr.go
+++ b/pkg/server/service/loadbalancer/wrr/wrr.go
@@ -22,6 +22,20 @@ type stickyCookie struct {
 	name     string
 	secure   bool
 	httpOnly bool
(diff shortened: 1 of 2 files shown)


ldez (approved):
LGTM

rtribotte (approved):
Thanks!

PR #10070, opened 2023-08-09
fix(sec): upgrade github.com/docker/docker to 23.0.3

ChengDaqi2023:
### What happened？
There are 1 security vulnerabilities found in github.com/docker/docker v20.10.21+incompatible
- [CVE-2023-28840](https://www.oscs1024.com/hd/CVE-2023-28840)

### What did I do？
Upgrade github.com/docker/docker from v20.10.21+incompatible to 23.0.3 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### How can we automate the detection of these types of issues?
By using the [GitHub Actions](https://github.com/murphysecurity/actions) configurations provided by murphysec, we can conduct automatic code security checks in our CI pipeline.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/docker/cli v20.10.11+incompatible
 	github.com/docker/compose/v2 v2.0.1
-	github.com/docker/docker v20.10.21+incompatible
+	github.com/docker/docker v23.0.3+incompatible
 	github.com/docker/go-connections v0.4.0
 	github.com/fatih/structs v1.1.0
 	github.com/go-acme/lego/v4 v4.12.0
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

We are not impacted by this CVE.

We can update but there is no impact on Traefik.

soumiksamanta:
> Hello,
> 
> We are not impacted by this CVE.
> 
> We can update but there is no impact on Traefik.

Can we at the lease update to 20.10.24 which will be a minor update for Traefik?  Is there any doc reference which specifies that Traefik is not impacted by this CVE

ldez:
> Is there any doc reference which specifies that Traefik is not impacted by this CVE

No, but, as the maintainer who did the analysis, the docker/docker module is large and we just use a very small part (a part of the API client), the CVE is not about something we use.

> Can we at the least update to 20.10.24 which will be a minor update for Traefik?

The updates of Docker are never simple because the [module](https://github.com/moby/moby) doesn't have a go.mod. So every dependency should be checked by hand.
I still hope that Docker will replace their custom approach of Go module with something that really works with the Go module system.
You can understand "minor" update doesn't really exist with this module.

madhavajay:
@ldez I get that this sucks, scanners have become a bit of a scourge of open source projects especially with legitimate false positives. 

Do you have a suggestion on what to do here? We can try to report these as false positives but there are a tonne of scanners and its getting a bit out of hand. Users don't care they just see RED warnings and think that artefacts are potentially dangerous.

I have seen several other `go` projects decline to update their insecure dependencies claiming they are not effected but how can anyone know for sure without doing a full audit.

ldez:
The real solution is to improve vulnerability databases.

The core of the problem is that vulnerability scanning tools don't share the same knowledge.
The best solution is to have a shared, free, and open-source security database with a way to report false positives.
Without a global place to report false positives, no tool that only does dependency analysis can be guaranteed without false positives.

PR #10079, opened 2023-08-14
doc: Subdomain could also contain digits.

ReeseWang:
### What does this PR do?

Updated the HostRegexp examples in router documentation to allow digits in domain names.


ldez:
Hello,

it's just a regex sample, not a regex to handle all subdomains.
domains can contain a lot more characters than a-z and 0-9.

We will decline your PR.

ReeseWang:
okay

PR #10089, opened 2023-08-24
Disconnect Postgres clients attempting plaintext connection

sjmiller609:
### What does this PR do?

TCP disconnect when a Postgres client is attempting a plaintext connection, instead of just hanging the connection.

### Motivation

https://github.com/traefik/traefik/issues/9929

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/server/router/tcp/postgres.go
+++ b/pkg/server/router/tcp/postgres.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"sync"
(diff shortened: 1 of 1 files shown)


sjmiller609:
Having an error in tests like
```
fatal error: all goroutines are asleep - deadlock!
```

I expect that's because I'm trying to read at least 5 bytes before doing anything else, and that's invalid for short messages. The check at the top of the loop isn't doing what I thought, so maybe we have to return false, nil in the case of a short message, however then it seems that a short message that is an exact match for the initial bytes of PostgresStartTLS would have an issue?

ldez:
can you rebase your PR onto the v3.0 branch?

sjmiller609:
@ldez 

Ok done! Thank you.

rtribotte:
Hello @sjmiller609,

We reviewed the changes and have some feedback on the drawbacks of it, as you already have identified.

The `isPostgres` function was previously returning as soon as one byte would not match one of the 8 bytes of the TLS startup message. This was done not to block protocols that could use less than 8 bytes.
With the changes, Traefik would now block on peek operation for protocols using less than 5 bytes.

While we don't have in mind any protocol for which the client initiates a first message lower than 5 bytes, we do not want to introduce a breaking behavior.

We lack a non-blocking way to check the first bytes sent by a client, and the only other alternative would be to make that behavior optional. Still, we are not comfortable with adding yet another very specific option to the Traefik static configuration.

For these reasons, we are declining this PR.

As a workaround, and as long as the entryPoint would not be used to handle non-TLS HTTP nor another non-TLS TCP service, one can attach a TCP catchAll router routing to an empty TCP server load balancer.
This way, psql connections that would not use the correct sslmode would be closed.

Example with the file provider:
```yaml
tcp:
  routers:
    catchAll:
      entryPoints:
        - "[PostGRES dedicated entryPoint]"
      rule: "HostSNI(`*`)"
      service: empty
  
  services:
    empty:
      loadBalancer:
        servers: {}
```

sjmiller609:
It works! 🎉 
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: postgres-catch-all
  namespace: traefik
data:
  postgres-catch-all.yaml: |
    tcp:
      routers:
        catchAll:
          entryPoints:
            - "postgresql"
          rule: "HostSNI(`*`)"
          service: empty
      services:
        empty:
          loadBalancer:
            servers: {}
```

related helm chart values
```
          volumes:
          - name: 'postgres-catch-all'
            mountPath: "/config"
            type: configMap
          additionalArguments:
            - "--entryPoints.postgresql.address=:5432/tcp"
...
            - "--providers.file.filename=/config/postgres-catch-all.yaml"
```

PR #10096, opened 2023-08-28
Adjust forward auth to avoid connection leak

wdhongtw:
### What does this PR do?

This MR provides a minor bug fix for forward auth middleware implementation.

During the sub-request of forward auth process, If error happens on `io.ReadAll(forwardResponse.Body)`, current implement won't close the body, the underline http connection can not be reuse for next request, and results in connection leak in HTTP 1.1 environment.

The only change in this MR is to register the defer statement earlier, covering all possible execution path.

Although it’s unlikely for a working connection to return error on body read, it’s still a good practice to `defer body.Close()` immediately after `client.Do` if it does not return error.

See also:

- `net/http` document: <https://pkg.go.dev/net/http>
- `persistConn` and `bodyEOFSignal` implementation: <https://github.com/golang/go/blob/master/src/net/http/transport.go>

### Motivation

It's not a serious bug, just nitpicking some implementation details. :D

### More

- [ ] Added/updated tests

### Additional Notes

I grep the code base with `client.Do` usage, another issue I found is in `integration/try/try.go`.
Since that package is only used in test, it’s should be ok to keep it untouched. :D

### DIFF
--- a/pkg/middlewares/auth/forward.go
+++ b/pkg/middlewares/auth/forward.go
@@ -126,6 +126,7 @@ func (fa *forwardAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		rw.WriteHeader(http.StatusInternalServerError)
 		return
 	}
+	defer forwardResponse.Body.Close()
 
 	body, readError := io.ReadAll(forwardResponse.Body)
 	if readError != nil {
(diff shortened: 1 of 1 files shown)


rtribotte:
Hello @wdhongtw,

Thanks for this contribution!

Can you please rebase the PR on the v2.10 branch?
While we think it is worth doing this change as a better pattern, we are unsure that the problem could really happen.
Nonetheless, this is more of a bug fix and hence it should be merged on the current v2.10.

nmengin (approved):
LGTM thanks 👍

ldez (approved):

PR #10109, opened 2023-09-04
Fix Trivy Scan Vulnerability

bhoodevsharma14:
### What does this PR do?
I was trying to use the traefik latest image for a project before deployment we generally Scan the image using trivy scan for the vulnerability there I found some High vulnerability , to fix those vulnerability issues I have updated the go modules to the fixed version it will be good for all if the project is least vulnerable


ldez:
As already said 100 times, we are not impacted by those CVEs.

PR #10128, opened 2023-09-15
Improve CNAME flattening to avoid unnecessary error logging

niallnsec:
### What does this PR do?

This PR improves the Request Decorator middleware to avoid logging unnecessary error messages and performing invalid DNS lookups.

### Motivation

The motivation for this pull request was from seeing a large number of error messages on a production system like the following:

```
{"level":"error","msg":"Item example.com already exists","time":"2023-09-15T08:42:54Z"}
{"level":"error","msg":"Failed to resolve host 127.0.0.1: empty answer for server 127.0.0.11","time":"2023-09-15T08:46:02Z"}
{"level":"error","msg":"Failed to resolve host staging.example.com: empty answer for server 127.0.0.11","time":"2023-09-15T08:47:56Z"}
```

The first message is the result of calling the Add method on the resolver cache instead of Set. According to the documentation for [Add](https://pkg.go.dev/github.com/patrickmn/go-cache?utm_source=godoc#Cache.Add) it returns an error if an entry already exists. I suspect that the original author used Add because earlier in the code path we confirm the item is not present in the cache, however, there is a TOCTOU issue as this code is inherently race-y with no locking. Rather than introduce locking over the get and set operation which would introduce unnecessary contention on the cache and potentially cause significant lags for some requests, switching to the Set method removes the error and ensures an up to date record is stored in the cache. This also eliminates the ambiguous error from the snippet above.

There is also an error logged any time that Traefik tries to resolve the CNAME for an IP address. Since IP addresses are not domains, we should check to see if the host is an IP before we attempt to resolve it. Doing so removes the second error message.

The third error message is output whenever a host has no CNAME record. I a real world system, it is quite likely that you will have a mix of domains that require flattening and domains that are directly mapped. If the DNS lookup returns no CNAME records for a host, this should not be treated as an error since it only implies that the given host is not subject to flattening.

Rather than remove the error completely, I opted to change its level so that it is only reported at the Debug level. The reasoning behind this is that the error could still be useful for someone debugging a CNAME flattening failure in the case that they have not created the necessary DNS records or that propagation is still pending.

### DIFF
--- a/pkg/middlewares/requestdecorator/hostresolver.go
+++ b/pkg/middlewares/requestdecorator/hostresolver.go
@@ -2,6 +2,7 @@ package requestdecorator
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"sort"
@@ -65,9 +66,7 @@ func (hr *Resolver) CNAMEFlatten(ctx context.Context, host string) string {
 		request = resolv.Record
 	}
 
-	if err := hr.cache.Add(host, result, cacheDuration); err != nil {
-		logger.Error().Err(err).Send()
-	}
+	hr.cache.Set(host, result, cacheDuration)
 
 	return result
 }
@@ -86,8 +85,16 @@ func cnameResolve(ctx context.Context, host, resolvPath string) (*cnameResolv, e
 
 	var result []*cnameResolv
 	for _, server := range config.Servers {
+		if net.ParseIP(server) != nil {
+			continue
+		}
+
 		tempRecord, err := getRecord(client, m, server, config.Port)
 		if err != nil {
+			if errors.Is(err, errNoCNAMERecord) {
+				log.Ctx(ctx).Debug().Err(err).Msgf("Failed to resolve host %s", host)
+				continue
+			}
 			log.Ctx(ctx).Error().Err(err).Msgf("Failed to resolve host %s", host)
 			continue
 		}
@@ -102,6 +109,8 @@ func cnameResolve(ctx context.Context, host, resolvPath string) (*cnameResolv, e
 	return result[0], nil
 }
 
+var errNoCNAMERecord = errors.New("no CNAME record for host")
+
 func getRecord(client *dns.Client, msg *dns.Msg, server, port string) (*cnameResolv, error) {
 	resp, _, err := client.Exchange(msg, net.JoinHostPort(server, port))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


rtribotte:
Hello @niallnsec,

We took a look at the changes and we think that the change of the log level and the usage of the `Set` method for the cache, could be in a fix PR that targets the v2 branch.

The use of `net.ParseIP` to avoid the CNAME flattening for IPs is more on the edge, but as it can be seen as a fix, let's keep those three changes in the same PR.

Could you please rebase the PR on the v2.10 branch?

niallnsec:
Thanks! Sure, no problem.

niallnsec:
Hoping that I got that right @rtribotte since I've not tried changing the base of a PR before instead of opening a new one. Looks good as far as I can tell though and tests are still passing for me locally.

pkg/middlewares/requestdecorator/hostresolver.go
```diff
@@ -88,7 +91,11 @@ func cnameResolve(ctx context.Context, host, resolvPath string) (*cnameResolv, e
 	for _, server := range config.Servers {
 		tempRecord, err := getRecord(client, m, server, config.Port)
 		if err != nil {
-			log.FromContext(ctx).Errorf("Failed to resolve host %s: %v", host, err)
+			if errors.Is(err, errNoCNAMERecord) {
+				log.FromContext(ctx).Errorf("Failed to resolve host %s: %s", host, err)
```
rtribotte, line 95:
```suggestion
				log.FromContext(ctx).Errorf("CNAME lookup for hostname %q: %s", host, err)
```

pkg/middlewares/requestdecorator/hostresolver.go
```diff
@@ -88,7 +91,11 @@ func cnameResolve(ctx context.Context, host, resolvPath string) (*cnameResolv, e
...
+				continue
+			}
+			log.FromContext(ctx).Errorf("Failed to resolve host %s: %s", host, err)
```
rtribotte, line 98:
```suggestion
			log.FromContext(ctx).Errorf("CNAME lookup for hostname %q: %s", host, err)
```

rtribotte (changes requested):
It seems more appropriate to be factual on the operation, more than warn about the failure that will be anyway explained by the error.

niallnsec:
Yeah that makes sense, I have applied your suggested changes

niallnsec:
I just noticed when I rebased I inadvertently reverted the Debug level error message back to Error level. The last update fixed that.

rtribotte (approved):
Thanks 👍

kevinpollet (approved):
Thanks 👍

ldez (approved):
LGTM

PR #10131, opened 2023-09-17
Instruct people to send enhancements to the v3 branch

jfly:
Per
https://github.com/traefik/traefik/pull/10130#issuecomment-1722540432:

> It's more complex: currently, the v3 is in beta so we continue
> integrating enhancements and breaking changes. It was not the initial
> plan but it's the current status.


ldez (approved):
LGTM

PR #10132, opened 2023-09-18
Fix X-Forwarded-For delete operation

### LINKED ISSUE #9979: Ability to remove X-Forwarded-* headers
### Welcome!

Hello Traefik Team,

I'm opening this feature request to address a situation with the handling of `X-Forwarded-*` headers in Traefik. Currently, these headers are added after the middlewares have been processed, which makes it impossible for **custom middlewares** (plugins) to remove or alter these headers. While I understand that these headers are vital for proper functioning of many architectures, giving the users an option to control these headers can provide additional flexibility and can be crucial in certain scenarios.

### Use Cases:

1. In some scenarios, we might want to fully control the headers passed on to the downstream services, and these automatic headers may conflict with the control.

2. It might be a security concern for some users. While `X-Forwarded-*` headers are generally harmless and useful, there might be cases where users don't want to expose certain information about the request's path.

3. It provides more flexibility in terms of header management, which could be useful in complex architectures and during debugging.

### Proposed Solution:

Introduce a configuration option that enables users to choose whether to include these headers or not. This option can be true by default to maintain the current behavior, which is adding these headers.

Please note that this feature request is not about disabling these headers completely, but about giving users a choice. If a user doesn't specify anything, the default behavior should remain intact, that is, the headers should be added as they are now.

P.S. ChatGPT wrote this, did way better job than I would be able to. :pray:

rtribotte:
### What does this PR do?

This PR adds support for the deletion of the `X-Forwarded-For` header.
It changes the way the `headers` middleware proceeds for this header deletion operation, to take advantage of the changes of https://github.com/golang/go/commit/ecdbffd4ec68b509998792f120868fec319de59b.

Previously the entry was removed from the request headers map, which would let the reverse proxy add back the header.

Now the entry is kept in the request header map but the value is set to nil, preventing the reverse proxy from adding the header later.

### Motivation

Fixes #9979

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: landrybe <lbenguigui@gmail.com>

### DIFF
--- a/integration/fixtures/headers/remove_reverseproxy_headers.toml
+++ b/integration/fixtures/headers/remove_reverseproxy_headers.toml
@@ -1,0 +1,31 @@
+[global]
+  checkNewVersion = false
+  sendAnonymousUsage = false
+
+[log]
+  level = "DEBUG"
+
+[entryPoints]
+  [entryPoints.web]
+    address = ":8000"
+
+[providers.file]
+  filename = "{{ .SelfFilename }}"
+
+## dynamic configuration ##
+
+[http.routers]
+  [http.routers.router1]
+    rule = "Host(`test.localhost`)"
+    middlewares = ["remove"]
+    service = "service1"
+
+[http.middlewares]
+  [http.middlewares.remove.headers.customRequestHeaders]
+    X-Forwarded-For = ""
+    Foo = ""
+
+[http.services]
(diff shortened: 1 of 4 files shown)


tomMoulard (approved):
LGTM :ok_hand:

kevinpollet (approved):
LGTM

Issue #7113, opened 2020-07-28
Improve circuit breaker to accept fallback service


sylr:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

The circuit breaker middleware is quite useful but limited to a single response (a HTTP 503 error).

It would be quite useful to allow this middleware to forward request to another service when it is tripped.

See #6922.

ldez:
Could you provide more information, in the issue description, about your use-cases, etc.

Related to https://github.com/containous/traefik/pull/6922#issuecomment-648879208

sylr:
At my company we have developed a service which aims to catch all requests that have not been treated, either because there is no service registered for the route (404) or because of a service unavailability.

This service makes stats and stores requests in kafka in order to be replayed in needed.

We would like to forward requests intercepted by the circuit breaker.

dsikes:
We also could find this feature useful. Being able to configure the options (including a failover URL) would make the circuit breaker significantly more useful than it currently is. I'm happy to work on a PR if it would be welcomed.

sylr:
@dsikes I already made one that I'm using at my company https://github.com/traefik/traefik/pull/6922

dsikes:
@sylr Looks like it won't get merged? For our use case, we won't build the binaries from source. We'll need to use the official ones. So unless the team is willing to help us out, the feature we are requesting will remain lacking.

SantoDE:
Hello @dsikes,

last time, we declined that PR for the reasons given in the PR.

However, maybe it could be worth trying to implement that functionality as a plugin? https://github.com/traefik/plugindemo

dsikes:
@SantoDE - building it as a plugin is certainly an option. 
I do have a couple questions around this. 

1. Is there **any** performance impact when using a plugin? 
    From what I understand, the plugins are loaded in and executed via Yaegi.
    I just want to be sure that there wouldn't be a cost associated with performance.

2. Is there any requirement on Traefik Pilot for using plugins?
     Or can we manually specify our plugin in the static config and not have a dependency on Pilot?

rkr-kununu:
I see that this topic has stalled, which is a pity 'cause I see a few use cases for it:
1. customizing the public facing http-503 with a body that matches the theme of the company/project.
2. Add a [Retry-After](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) Header to _only_ the http-503's (to tell the crawlers to come again after a period, I define).
[...]

PR #10147, opened 2023-10-01
Circuitbreaker/responsecode

fahhem:
### What does this PR do?

This allows CircuitBreaker to be configured to respond with a response code other than 503.
Closes https://github.com/traefik/traefik/issues/7113

### Motivation

This is useful for when clients of an API handle certain response codes differently, especially external clients, and you want to return another code instead. This could be expanded to add custom headers.

### More

- [ ] Added/updated tests

### Additional Notes

This seems like it could instead be done through the Errors middleware, so I will be trying that in a moment and will reply to the issue and possibly close both if that works.

### DIFF
--- a/pkg/config/dynamic/middlewares.go
+++ b/pkg/config/dynamic/middlewares.go
@@ -139,6 +139,8 @@ type CircuitBreaker struct {
 	FallbackDuration ptypes.Duration `json:"fallbackDuration,omitempty" toml:"fallbackDuration,omitempty" yaml:"fallbackDuration,omitempty" export:"true"`
 	// RecoveryDuration is the duration for which the circuit breaker will try to recover (as soon as it is in recovering state).
 	RecoveryDuration ptypes.Duration `json:"recoveryDuration,omitempty" toml:"recoveryDuration,omitempty" yaml:"recoveryDuration,omitempty" export:"true"`
+	// ResponseCode is the code that the circuit breaker will return while it is in the tripped state.
+	ResponseCode int `json:"responseCode,omitempty" toml:"responseCode,omitempty" yaml:"responseCode,omitempty" export:"true"`
 }
 
 // SetDefaults sets the default values on a RateLimit.
@@ -146,6 +148,7 @@ func (c *CircuitBreaker) SetDefaults() {
 	c.CheckPeriod = ptypes.Duration(100 * time.Millisecond)
 	c.FallbackDuration = ptypes.Duration(10 * time.Second)
 	c.RecoveryDuration = ptypes.Duration(10 * time.Second)
+	c.ResponseCode = 503
 }
 
 // +k8s:deepcopy-gen=true
--- a/pkg/middlewares/circuitbreaker/circuit_breaker.go
+++ b/pkg/middlewares/circuitbreaker/circuit_breaker.go
@@ -30,12 +30,14 @@ func New(ctx context.Context, next http.Handler, confCircuitBreaker dynamic.Circ
 	logger.Debug().Msg("Creating middleware")
 	logger.Debug().Msgf("Setting up with expression: %s", expression)
 
+	responseCode := confCircuitBreaker.ResponseCode
+
 	cbOpts := []cbreaker.Option{
 		cbreaker.Fallback(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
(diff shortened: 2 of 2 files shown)


fahhem:
Tried the Errors middleware and it's not as good, went with a plugin for now

geraldcroes:
Hi @fahhem and thanks for this PR.

I don't think this PR would close #7113 that mostly deals with having this circuit breaker forward to a different service.

At first I like the idea of being able to configure a different status code, but after giving it more thoughts, I can't find something more accurate than 503 in this context. Could you elaborate a bit more?

fahhem:
My client wanted to be able to return 204, as that's what the API was documented as returning in the case that the API has nothing new to return. When we return a 503, clients consider that downtime, but in our use-case the circuit breaker is used to avoid real downtime (where we give invalid results).

At the moment, we're doing this through a custom plugin, but it would be nice to replace it all with this much smaller PR

geraldcroes:
Thank you.

I understand the use case, but I still think it can't find its solution in the circuit breaker middleware.

I'll try to explain my point.

In the use case of the circuit breaker (by itself), it triggers open when the service is unhealthy and unable to provide proper response. Hence, out of the box, 503 (service unavailable) seems to be the right answer -> it is open when the service has reached a behavior that traefik deem unhealthy, based on the configuration.

If I wanted to configure a 204 (no response) instead, the rest of the chain wouldn't know that the underlying service is unhealthy when it's open ... and could believe the service is fine and tells it has nothing to say (the exact purpose of 204 in your example).

If, for whatever reason, I'd like _then_ to tweak that so that the caller gets a 204, it's to me a separate concern, hence I would chain it with the error middleware, converting the status code to another code.

You should get the behavior you're asking for, but the subtle difference is that when you chain the CircuitBreaker middleware with others (embedded or plugins), the rest of the chain would know the proper code and act accordingly. Then, eventually, at the end of the chain, you could add your HTTP Status Code "translator" (in the form of the error middleware) to have this 503 -> 204 translation.

Could you give it a try and tell me if it works for you? If it does not, I'm sure we can still find a different solution based on this idea.

geraldcroes:
We discussed this during triage today, and we came to the conclusion that adding this option had its perks compared to mixing and matching with other middleware.

One argument in favor of making it configurable was being able to put appart use cases when the circuit breaker was open and cases when the actual service was sending a 503.

We also discussed the idea of creating a dedicated middleware capable of translating http-status-codes that people could chain, which can for now be accomplished using a plugin.

All in all, we've decided to proceed with the review.

geraldcroes (changes requested):
Thx again for this PR.

Could you add the documentation of this new option?

fahhem (commented):
Thanks @geraldcroes ! Sorry for the delay, but I added the docs you requested

nmengin:
Hello @fahhem,

It seems that you have conflicts, on top of that the CI steps are not all successfully executed.
Could you take a look? Do you need help on it?

fahhem:
@nmengin when I looked at this, the check failure was due to the year changing and the upstream branch didn't have the update in yet so there was no point. Looks like that chore was done the next day

juliens (approved):
LGTM

pkg/config/dynamic/middlewares.go
```diff
@@ -141,13 +141,16 @@ type CircuitBreaker struct {
...
+	// ResponseCode is the status code that the circuit breaker will return while it is in the open state.
+	ResponseCode int `json:"responseCode,omitempty" toml:"responseCode,omitempty" yaml:"responseCode,omitempty" export:"true"`
 }
 
 // SetDefaults sets the default values on a RateLimit.
 func (c *CircuitBreaker) SetDefaults() {
 	c.CheckPeriod = ptypes.Duration(100 * time.Millisecond)
 	c.FallbackDuration = ptypes.Duration(10 * time.Second)
 	c.RecoveryDuration = ptypes.Duration(10 * time.Second)
+	c.ResponseCode = 503
```
rtribotte, line 153:
```suggestion
	c.ResponseCode = http.StatusServiceUnavailable
```

nmengin (approved):
LGTM 👏

nmengin:
Hey @fahhem,

It seems that there is a CI issue with your PR.
Could you please rebase your PR in the v3.0 branch?

Thanks in advance.

geraldcroes (approved):
LGTM

PR #10150, opened 2023-10-04
Add Coraza web application firewall (WAF) support

zetaab:
### What does this PR do?

This PR adds [Coraza](https://github.com/corazawaf/coraza) web application firewall (WAF) support to Traefik. It includes basic OWASP rules from https://github.com/coreruleset/coreruleset/

### Motivation

I am impressed how Traefik has been improving during years. There are lots of improvements everywhere and nowadays I would say Traefik is top3 solution to ingress controllers in Kubernetes. However, it is missing key component which is WAF. The choice of the technology was not difficult as Modsecurity is deprecated. Coraza supports same syntax than Modsecurity.

### More

- [x] Added/updated tests

### Additional Notes

I am not sure how this configuration format can be supported in other platforms like docker & swarm. But please leave a suggestion to improve the config format.


jptosso:
Cc @emilevauge @jcchavezs

docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
```diff
@@ -788,6 +788,19 @@ spec:
                   type detected from the response content, when it is not set by the
                   backend.
                 type: object
+              coraza:
+                description: Coraza holds the configuration of the coraza middleware.
+                  This middleware enables you to define coraza web application firewall
```
jcchavezs, line 793:
```suggestion
                  This middleware enables you to define the directives for coraza web application firewall
```

pkg/provider/kubernetes/crd/traefikio/v1alpha1/middleware.go
```diff
@@ -49,6 +49,7 @@ type MiddlewareSpec struct {
 	Retry             *Retry                     `json:"retry,omitempty"`
 	ContentType       *dynamic.ContentType       `json:"contentType,omitempty"`
 	GrpcWeb           *dynamic.GrpcWeb           `json:"grpcWeb,omitempty"`
+	Coraza            *Coraza                    `json:"coraza,omitempty"`
```
jcchavezs, line 52: I'd call it CorazaWAF or WAF. Coraza by itself is not too intuitive.

zetaab: First I was thinking to make coraza under waf category, but it might be that there are not many waf implementations really? Perhaps the best is to use `CorazaWAF`, because it might be a problem to use just `WAF` if someone wants to add second waf middleware in future

jptosso:
It would be nice to get an opinion from the Traefik team. This is a major feature.
Unfortunately it is impossible to bring coraza as a patch or as a plugin. This is the only way

keskad:
Wow, this approach could simplify WAF on Kubernetes. Is it enabled/disabled per ingress or for the whole cluster? Would it work outside Kubernetes - if yes then it would be nice to have a config example.

jcchavezs:
> Wow, this approach could simplify WAF on Kubernetes. Is it enabled/disabled per ingress or for the whole cluster? Would it work outside Kubernetes - if yes then it would be nice to have a config example.

@keskad it should work outside kubernetes as this is only another middleware for traefik.

docs/content/middlewares/http/corazawaf.md
```diff
@@ -0,0 +1,82 @@
...
+As an example we might want to block /admin path in production.
+
+```yaml tab="Kubernetes"
+apiVersion: traefik.io/v1alpha1
+kind: Middleware
+metadata:
+  name: waf
+spec:
+  corazaWAF:
+    directives: |
```
jcchavezs, line 41: QQ on this: does traefik support passing the config as a JSON? If so, we should probably make this a list of string, otherwise this looks alright.

docs/content/middlewares/http/corazawaf.md
```diff
@@ -0,0 +1,82 @@
...
+      SecRuleEngine On
+      SecRule REQUEST_URI "@streq /admin" "id:101,phase:1,t:lowercase,log,deny"
+```
+
+### `crsEnabled`
```
jcchavezs, line 46: For the sake of consistency I'd strongly recommend call it `loadOWASPCRS` as in https://github.com/corazawaf/coraza-caddy/blob/main/coraza.go#L34C31-L34C45

zetaab:
this is how I think it should work (kubernetes, yaml and toml):

```yaml tab="Kubernetes"
# Denying /admin path
kind: Middleware
metadata:
  name: waf
spec:
  corazaWAF:
    directives: |
      SecRuleEngine On
      SecRule REQUEST_URI "@streq /admin" "id:101,phase:1,t:lowercase,log,deny"
```

```yaml tab="File (YAML)"
# Redirect to https
http:
  middlewares:
    waf:
      corazaWAF:
        directives: |
          SecRuleEngine On
          SecRule REQUEST_URI "@streq /admin" "id:101,phase:1,t:lowercase,log,deny"
```

```toml tab="File (TOML)"
# Redirect to https
[http.middlewares]
  [http.middlewares.waf.corazaWAF]
    directives = [
      "SecRuleEngine On",
      "SecRule REQUEST_URI \"@streq /admin\" \"id:101,phase:1,t:lowercase,log,deny\"",
    ]
```

However, I have no idea how these 2 (Consul Catalog and Docker&Swarm) configuration models could work. Both models takes variables "as labels", its not planned to insert long string configs using labels.

nmengin:
Hey @zetaab,

Thank you so much for this PR and contributing to Traefik.
The idea is appealing, but because it impacts the philosophy of the project, we want to give the community time to express their opinion on the WAF topic as a whole.

We’ve just opened a new [proposal](https://github.com/traefik/traefik/issues/10162) and I propose we move the conversation there before deciding the next step with this PR.

geraldcroes:
> > it impacts the philosophy of the project
> 
> Could you please explain what do you mean? You do not want that Traefik and services behind it to be as secure as possible? As I see it: all modern reverse proxies should have WAF capability. Of course we cannot force everyone to use WAF, but even make it possible for the users.

Hi! Jumping in the conversation.

On my end, of course I want traefik and services behind it to be as secure as possible, I do not think that this is what Nicolas meant by philosophy. To me, philosophy is more about being lightweight, simplicity in usage and configuration: how do we bring the feature, how it should be configured, and is embedding a given technology the best choice for the user base. I'm aware of many use cases were people are perfectly happy (operating a secured system) with traefik not embedding a WAF.

Don't get me wrong, I value at the highest level this contribution (and any contribution that tries to improve traefik). I also have an opinion on whether or not we should proceed with your PR. Still, I also value opening the conversation before any commitment: If we see, then decide that this is the best way, we'll fully support the feature for years to come. So to me at least, this is very important that we open a proposal, and discuss the proposal first.

It adds a bit of delay, but this is for the best interest of everyone :-)

On a side note, one thing for sure: with this PR you caught full attention of many maintainers.

jcchavezs:
Thanks @geraldcroes for the thoughtful feedback.

I think specifically in a reverse-proxy/load balancer the WAF functionality is a first class concern because it is the first line of defence so the question would be "if not embedded what is the alternative?" I see only two alternatives: putting yet another proxy in front or moving onto another ingress which supports WAF.

[As of 2025 PCI DSS 4.0 compliance will require to be secured by a WAF](https://www.securitymetrics.com/blog/updates-and-changes-in-pci-dss-40) so I guess this kind of feature request will be more often but given the intensive labour of a WAF in a proxy, embedding it is optimal performance wise: request headers+body is copied into WAF memory to do analysis and resolve, same for response headers a body, any non embedded will require copies or worst case network traffic which can be a pain in the hotpath.

I would really love to hear more about your opinions on this specific PR so maybe we can get to a good point.

dune73:
OWASP CRS co-lead hear with 18 years of WAF experience.

I am biased apparently, but welcome any WAF functionality and if it eases WAF on Kubernetes / Ingress, all the better.

Still, I think the users should have a choice. Enable WAF in blocking mode by default for traefik installation would be very difficult for users (in case this is on the table - not sure I get the conversation correctly). CRS is the best WAF rule set around in terms of detection, but CRS is a demanding companion and false positives can spoil the fun. So configurability and documentation is key.

If you have more questions with regards to CRS, please shout, I'll try to follow this conversation here.

geraldcroes:
> I think specifically in a reverse-proxy/load balancer the WAF functionality is a first class concern because it is the first line of defence [...] 

I must admit that I am personally biased in favor of embedding a WAF in traefik. If done well, as you pointed out, it would provide a nice first line of defense. 

I even thought about advocating for adding the feature as experimental and see what'd be the feedback and usage. Still, I think we first need to consider the bigger picture, hence the proposal (and nice conversation there).

As @juliens mentioned there (in the proposal), our biggest problem right now is that I'd very much welcome the feature in any plugin system, but we don't provide a stable enough system for that kind of feature at this very moment.

So my opinion about 
- WAF and Traefik: I think we need to find a way to embed such a feature. It'd be a plus for many users.
- Coraza and Traefik: even if we eventually find that embedding it in the core is not the best idea and we want to go for a different approach, we need to provide a clean integration system for Coraza users.

emilevauge:
My 2 cents:

- I also think we should move forward with this WAF integration
- It looks like WASM based plugins with http-wasm [is really promising](https://github.com/traefik/traefik/issues/9552#issuecomment-1771731496) and will allow a fast integration between Traefik & WASM for middlewares
- I would kill two birds with one stone & create a first PR adding http-wasm plugins as an experimental feature, and then create another PR adding Coraza as a WASM plugin

WDYT?

zetaab:
@emilevauge sounds enough good for me. However, if Traefik wants to use current plugin system I think the http-wasm plugin system needs to be implemented by Traefik people (because you know the codebase the best, and you know how you want to it to be). After that it would be nice if Traefik could also provide small "hello world" http-wasm example. After that like this Coraza PR can be converted to http-wasm plugin.

One interesting thing is that at least I am interested about performance with native Coraza middleware (this PR) and http-wasm Coraza (coming soon). But I can measure that somehow later.

evacchi:
Hey, my 2c. I am in the dev team of the [wazero](https://wazero.io) Wasm runtime (used in https://github.com/http-wasm/http-wasm-host-go). FWIW, I would also advocate for considering a native Go Coraza WAF implementation because in its current state the Wasm version requires serializing data to the linear memory. Adopting the Go version allows to pass around cheap references, avoiding the serialization step which could be a pain when dealing with evaluation of big request/response payloads in Coraza.

jcchavezs:
Shall we proceed with this PR?

geraldcroes:
Hi everyone, and especially @zetaab who initiated this work.

As described in the WAF Middleware issue (#10162) we decided not to proceed with this PR in favor of (as a first step) embedding it as a WASM plugin (thanks once again to @zetaab's work on #10189 that we'll start reviewing / integrating this week).

Thank you again for this PR that triggered so many interesting ideas.

dune73:
This decision sounds reasonable from my point of view. Thank you for making your plans transparent.

Issue #10174, opened 2023-10-18
Support non-standard nginx status code 444


springhack:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Nginx use non-standart status code 444 for close current connection, while there is no way close connection in traefik, I think it's a better way to close connection while upstream server return status 444.

nmengin:
Hello @springhack,

Thanks for reaching out.

As you've mentioned in the description, the status code 444 is not standard where we want Traefik to fit the RFC, so we prefer to close your issue and the related PR.

However you're right, there is no way to close connections in Traefik, and we've opened a [proposal](https://github.com/traefik/traefik/issues/10191) to work on such a mechanism.

PR #10175, opened 2023-10-18
Support non-standard nginx status code 444

springhack:
### What does this PR do?

Close connection while upstream server return status 444

### Motivation

Support non-standard nginx status code 444 for close connection #10174 

### More

- [ ] Added/updated tests


nmengin:
Hello @springhack,

Thanks for the PR.

As said in the related [issue](https://github.com/traefik/traefik/issues/10174), you've mentioned in the description that the status code 444 is not standard where we want Traefik to fit the RFC, so we prefer to close the PR.

However you're right, there is no way to close connections in Traefik, and we've opened a https://github.com/traefik/traefik/issues/10191 to work on such a mechanism.

PR #10178, opened 2023-10-19
don't compress requests if not Accept-Encoding header is present

robin-moser:
### What does this PR do?

Fixes https://github.com/traefik/traefik/issues/9734

To ensure compatibility, it is recommended not to compress a request when there is no Accept-Encoding header present, even though [RFC9110 (accept-encoding)](https://httpwg.org/specs/rfc9110.html#field.accept-encoding) would allow any compression.

### Motivation

Stops traefik from breaking multiple applications that don't send a Accept-Encoding header and at the same time don't have any support for encoded http responses. Two breaking examples are `yum/dnf` and `Zenarmor on Opnsense`.

### More

- [x] Added/updated tests


rjsocha:
The wording in RFC 2616, specifically in https://datatracker.ietf.org/doc/html/rfc2616#section-14.3 is similar to this

" If no Accept-Encoding field is present in a request, the server MAY
   assume that the client will accept any content coding. In this case,
   if "identity" is one of the available content-codings, then the
   server SHOULD use the "identity" content-coding, unless it has
   additional information that a different content-coding is meaningful
   to the client.
   "
   
  In any case, I believe this option should be configurable. For example:
```
  label:
  ...
  - traefik.http.middlewares.test4compress.compress.acceptEncodingMissing=no-compress
```
It should, by default, be disabled, following the semantics of RFC 9110.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,7 +11,7 @@ Bug fixes:
 
 Enhancements:
 - for Traefik v2: we only accept bug fixes
-- for Traefik v3: use branch master
+- for Traefik v3: use branch v3.0
```
rtribotte, line 14: The template should not be modified in this PR, also iirc, the templates in use are those of the master branch.

```suggestion
- for Traefik v3: use branch master
```

rtribotte:
> The wording in RFC 2616, specifically in https://datatracker.ietf.org/doc/html/rfc2616#section-14.3 is similar to this
> 
> " If no Accept-Encoding field is present in a request, the server MAY assume that the client will accept any content coding. In this case, if "identity" is one of the available content-codings, then the server SHOULD use the "identity" content-coding, unless it has additional information that a different content-coding is meaningful to the client. "
> 
> In any case, I believe this option should be configurable. For example:
> 
> ```
>   label:
>   ...
>   - traefik.http.middlewares.test4compress.compress.disableCompressionOnMissingAcceptEncoding=true
>   - traefik.http.middlewares.test4compress.compress.disableOnMissingAcceptEncoding=true
>   - traefik.http.middlewares.test4compress.compress.noAcceptEncodingCompression=true
> ```
> 
> It should, by default, be disabled, following the semantics of RFC 9110.

Hello @rjsocha,

Thanks for this feedback. It seems to me that the need to introduce an option to control this behavior should be discussed in a dedicated issue first.

rtribotte:
Hello @robin-moser,

I have enabled the checks, and as you can see, the `TestShouldCompressWhenNoAcceptEncodingHeader` test should be adapted as well.

robin-moser:
Sorry, I missed that one!
I updated the unit tests, locally they are passing now.

rtribotte (approved):
Thanks 👍

pkg/middlewares/compress/compress.go
```diff
@@ -93,11 +93,11 @@ func (c *compress) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	// Client allows us to do whatever we want, so we br compress.
-	// See https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.3
+	// Client doesn't specify preferred encoding, for compatibility dont encode the request
```
lbenguigui, line 96:
```suggestion
	// Client doesn't specify a preferred encoding, for compatibility don't encode the request
```

lbenguigui (changes requested):

kevinpollet (changes requested):

nmengin:
Hey @robin-moser ,

We'd like to bring your feature in Traefik v3.0.
Do you need some help on it?

robin-moser:
Hi @nmengin,
sorry for the late reaction. I just committed the suggested documentation changes.

kevinpollet (approved):
Thanks 👍

lbenguigui (approved):
LGTM

PR #10179, opened 2023-10-19
Disable br compression when no Accept-Encoding is present

### LINKED ISSUE #9734: compress midddleware: dont encode content if Accept-Encoding header is missing
### Welcome!

### What did you do?

I enabled content encoding with the new (3.0.0-beta2) comress middleware. When including a request header, for example `Accept-Encoding: br, gzip`, the content gets encoded properly by the new broli compression.

[RFC9110 (accept-encoding)](https://httpwg.org/specs/rfc9110.html#field.accept-encoding) states, that "If no Accept-Encoding header field is in the request, any content coding is considered acceptable by the user agent." So technically, the new implementation from traefik is correct. But in practice, this is really uncommon beahviour and could break many applications out there. 

The problem with this behaviour is, that tools like curl often don't send any header by default but can't deal with compression by themself. Given a tool with a webui and an api, the use wants the webui to compress the content, but some scripts or cli tools may use the api and fails because of the compression. So usually, a site doesn't compress the content if no header is present to ensure compatibility.

Examples: (all three sites support brotli compression)

```bash
# with the header

curl --silent --output /dev/null --dump-header /dev/stdout  -H 'Accept-Encoding: gzip,br' https://www.google.com | grep encoding
content-encoding: gzip
curl --silent --output /dev/null --dump-header /dev/stdout  -H 'Accept-Encoding: gzip,br' https://www.facebook.com | grep encoding
content-encoding: br
curl --silent --output /dev/null --dump-header /dev/stdout -H 'Accept-Encoding: gzip,br' https://www.docker.com | grep encoding
content-encoding: br
curl --silent --output /dev/null --dump-header /dev/stdout -H 'Accept-Encoding: gzip,br' https://<traefik-v3-with-compress> | grep encoding
content-encoding: br

#now without the header

curl --silent --output /dev/null --dump-header /dev/stdout https://www.google.com | grep encoding
curl --silent --output /dev/null --dump-header /dev/stdout https://www.facebook.com | grep encoding
curl --silent --output /dev/null --dump-header /dev/stdout https://www.docker.com | grep encoding
curl --silent --output /dev/null --dump-header /dev/stdout https://<traefik-v3-with-compress> | grep encoding
content-encoding: br

# this is the stated problem

curl https://<traefik-v3-with-compress>
Warning: Binary output can mess up your terminal. Use "--output -" to tell
Warning: curl to output it to your terminal anyway, or consider "--output
Warning: <FILE>" to save to a file.
```

I would suggest to omit compression if no `Accept-Encoding` header is present.

### What did you see instead?

When omitting the header, the content unexpectedly gets encoded by brotli, too. 

v3.0.0 beta2

middlewares.yml
```yaml
http:
  middlewares:
    compression:
      compress: {}
```

traefik.yml
```yaml
entrypoints:
  websecure:
    address: :443
    http:
      middlewares:
        - compression@file
```

_No response_

dani:
As this behavior, while compliant with
https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.3, breaks several clients (like curl or yum/dnf)

Should fix #9734

### What does this PR do?

Disable compression when no Accept-Encoding is present in the request

### Motivation

While using any encoding is valid when no Accept-Encoding (RFC wise), in real word it breaks several clients (at least curl, yum and dnf). Disabling compression when no Accept-Encoding is present is also a valid option, and will cause less issue.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/middlewares/http/compress.md
+++ b/docs/content/middlewares/http/compress.md
@@ -55,7 +55,7 @@ http:
     Responses are compressed when the following criteria are all met:
 
     * The `Accept-Encoding` request header contains `gzip`, `*`, and/or `br` with or without [quality values](https://developer.mozilla.org/en-US/docs/Glossary/Quality_values).
-    If the `Accept-Encoding` request header is absent, it is meant as br compression is requested.
+    If the `Accept-Encoding` request header is absent, then compression is disabled.
     If it is present, but its value is the empty string, then compression is disabled.
     * The response is not already compressed, i.e. the `Content-Encoding` response header is not already set.
     * The response`Content-Type` header is not one among the [excludedContentTypes options](#excludedcontenttypes).
--- a/pkg/middlewares/compress/compress.go
+++ b/pkg/middlewares/compress/compress.go
@@ -93,11 +93,14 @@ func (c *compress) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	// Client allows us to do whatever we want, so we br compress.
+	// In theory, if no Accept-Encoding is present in the request, the client
+	// should accept any Encoding.
 	// See https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.3
+	// In practice, using brotli in such a case breaks several clients
+	// (like curl or yum/dnf)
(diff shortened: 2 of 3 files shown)


ldez:
Duplicate of #10178

PR #10189, opened 2023-10-25
Add http-wasm plugin support for Traefik

### LINKED ISSUE #9552: It would be great if Traefik supported wasm plugins
### Welcome!

Envoy WASM plugins：
https://github.com/envoyproxy/envoy/tree/master/examples/wasm-cc

zetaab:
### What does this PR do?

This PR will add http-wasm plugin support for Traefik. 

fixes #9552 

### Motivation

http-wasm plugin ecosystem is going to be THE thing in next years. When http-wasm plugin system is implemented to Traefik, it will be possible to implement plugins also using another languages than Golang. When other ingress controllers will implement http-wasm support, it will be possible to share plugins between solutions as well.

### More

- [ ] Added/updated tests

### Additional Notes

I could not verify the remote plugin side. If someone could help with that. Release is available in https://github.com/zetaab/traefik-wasm-demo

### Performance tests

Setup: arm mac m1 pro, 32GB ram, ventura 13.6

Running Traefik with command `go run ./cmd/traefik/ --configFile=static.yaml`.
Using traefik whoami as backend `docker run -tid -p0.0.0.0:8081:80 traefik/whoami`

#### static.yaml

```
entryPoints:
  web:
    address: :8000

log:
  level: debug

api:
  dashboard: true
  insecure: true

providers:
  file:
    filename: ./dynamic.yaml

metrics:
  prometheus: {}

experimental:
  plugins:
    example:
      moduleName: github.com/traefik/plugindemo
      version: v0.2.2
  localPlugins:
    wasmLocalExample:
      moduleName: github.com/zetaab/traefik-wasm-demo
```

#### dynamic.yaml

```
http:
  routers:
    customer1:
      rule: Host(`powpow.demo.traefiklabs.tech`)
      service: customer1
      middlewares:
        - localWasm
    customer2:
      rule: Host(`powpow2.demo.traefiklabs.tech`)
      service: customer1
    customer3:
      rule: Host(`powpow3.demo.traefiklabs.tech`)
      service: customer1
      middlewares:
        - traefikdemo
      
  services:
    customer1:
      loadbalancer:
        servers:
          - url: "http://127.0.0.1:8081"

  middlewares:
    traefikdemo:
      plugin:
        example:
          headers:
            traefik: demo
    localWasm:
      plugin:
        wasmLocalExample:
          headers:
            local: foo
```

So what this setup means? It means that we do have 3 possible routes:
- one with wasm plugin (does quite same than Traefik demo)
- one without any plugins
- one with Traefik Yaegi demo plugin

#### Results:

(raw results https://gist.github.com/zetaab/0520d93b4754d86526d67a304c65dc4e)

Tests were executed using `ab -n 10000 -H 'Host: [host]' -c 10 http://127.0.0.1:8000/` command.

| route | req/s | time/req | 
| ----- | ----- | --------- |
| customer1 (with http-yasm) | 762 | 13.1 |
| customer2 (without any middleware) | 759 | 13.1 |
| customer3 (with Traefik demo, Yaegi) | 708 | 14.1 |

So this test looks like http-wasm performance is pretty much same than without any middleware. Yaegi is slightly worse.

However, the interesting stats are coming from Connection times:

| route | Connect Mean (+/-sd) | Processing Mean (+/-sd) |  Waiting Mean (+/-sd) | Total Mean (+/-sd) |
| ----- | ----- | --------- | --------- | --------- | 
| customer1 (with http-yasm) | 4 (1.9) | 9 (2.9) | 8 (2.7) | 12 (3.5) |
| customer2 (without any middleware) | 4 (1.7) | 8 (2.8) | 8 (2.5) | 12 (3.4) | 
| customer3 (with Traefik demo, Yaegi) | 5 (24) | 8 (6.6) | 8 (6.3) | 13 (24.9) |

What we can see here? We can see that http-wasm is little bit slower processing than without any middleware. However, if we compare statistics from Traefik demo to any other routes we can clearly see the difference. The standard deviation between http-wasm & without middleware and Yaegi is huge. This means that Yaegi will not perform that well than these other 2 routes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,6 +33,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.4
 	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/nomad/api v0.0.0-20220506174431-b5665129cd1f
+	github.com/http-wasm/http-wasm-host-go v0.5.1
 	github.com/influxdata/influxdb-client-go/v2 v2.7.0
 	github.com/influxdata/influxdb1-client v0.0.0-20191209144304-8bf82d3c094d
 	github.com/instana/go-sensor v1.38.3
@@ -351,6 +352,7 @@ require (
 	github.com/stretchr/objx v0.5.0 // indirect
 	github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/common v1.0.490 // indirect
 	github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/dnspod v1.0.490 // indirect
+	github.com/tetratelabs/wazero v1.2.0 // indirect
 	github.com/theupdateframework/notary v0.6.1 // indirect
 	github.com/tinylib/msgp v1.1.6 // indirect
 	github.com/tonistiigi/fsutil v0.0.0-20201103201449-0834f99b7b85 // indirect
--- a/pkg/plugins/builder.go
+++ b/pkg/plugins/builder.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"path/filepath"
 
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
@@ -38,31 +39,14 @@ func NewBuilder(client *Client, plugins map[string]Descriptor, localPlugins map[
 
 		logger := log.With().Str("plugin", "plugin-"+pName).Str("module", desc.ModuleName).Logger()
 
-		i := interp.New(interp.Options{
-			GoPath: client.GoPath(),
-			Env:    os.Environ(),
-			Stdout: logs.NoLevel(logger, zerolog.DebugLevel),
-			Stderr: logs.NoLevel(logger, zerolog.ErrorLevel),
-		})
-
-		err = i.Use(stdlib.Symbols)
+		i, wasmPath, err := initMiddlewareBuilder(logger, client.GoPath(), desc.ModuleName, manifest)
 		if err != nil {
-			return nil, fmt.Errorf("%s: failed to load symbols: %w", desc.ModuleName, err)
-		}
-
-		err = i.Use(ppSymbols())
-		if err != nil {
-			return nil, fmt.Errorf("%s: failed to load provider symbols: %w", desc.ModuleName, err)
-		}
-
-		_, err = i.Eval(fmt.Sprintf(`import "%s"`, manifest.Import))
-		if err != nil {
-			return nil, fmt.Errorf("%s: failed to import plugin code %q: %w", desc.ModuleName, manifest.Import, err)
+			return nil, fmt.Errorf("%s: %w", desc.ModuleName, err)
 		}
 
 		switch manifest.Type {
 		case "middleware":
-			middleware, err := newMiddlewareBuilder(i, manifest.BasePkg, manifest.Import)
+			middleware, err := newMiddlewareBuilder(i, manifest.BasePkg, manifest.Import, wasmPath)
 			if err != nil {
 				return nil, err
 			}
@@ -87,31 +71,14 @@ func NewBuilder(client *Client, plugins map[string]Descriptor, localPlugins map[
 
 		logger := log.With().Str("plugin", "plugin-"+pName).Str("module", desc.ModuleName).Logger()
 
-		i := interp.New(interp.Options{
-			GoPath: localGoPath,
-			Env:    os.Environ(),
-			Stdout: logs.NoLevel(logger, zerolog.DebugLevel),
-			Stderr: logs.NoLevel(logger, zerolog.ErrorLevel),
-		})
-
-		err = i.Use(stdlib.Symbols)
+		i, wasmPath, err := initMiddlewareBuilder(logger, localGoPath, desc.ModuleName, manifest)
 		if err != nil {
-			return nil, fmt.Errorf("%s: failed to load symbols: %w", desc.ModuleName, err)
-		}
-
-		err = i.Use(ppSymbols())
-		if err != nil {
-			return nil, fmt.Errorf("%s: failed to load provider symbols: %w", desc.ModuleName, err)
-		}
-
-		_, err = i.Eval(fmt.Sprintf(`import "%s"`, manifest.Import))
-		if err != nil {
-			return nil, fmt.Errorf("%s: failed to import plugin code %q: %w", desc.ModuleName, manifest.Import, err)
+			return nil, fmt.Errorf("%s: %w", desc.ModuleName, err)
 		}
 
 		switch manifest.Type {
 		case "middleware":
-			middleware, err := newMiddlewareBuilder(i, manifest.BasePkg, manifest.Import)
+			middleware, err := newMiddlewareBuilder(i, manifest.BasePkg, manifest.Import, wasmPath)
 			if err != nil {
 				return nil, err
 			}
@@ -129,4 +96,37 @@ func NewBuilder(client *Client, plugins map[string]Descriptor, localPlugins map[
 	}
 
 	return pb, nil
+}
+
+func initMiddlewareBuilder(logger zerolog.Logger, goPath string, moduleName string, manifest *Manifest) (*interp.Interpreter, string, error) {
+	if !manifest.IsYaegiPlugin() {
+		return nil, filepath.Join(goPath, "src", moduleName, manifest.WasmPath), nil
+	}
+	i, err := initInterp(logger, goPath, manifest.Import)
+	return i, "", err
+}
+
+func initInterp(logger zerolog.Logger, goPath string, manifestImport string) (*interp.Interpreter, error) {
+	i := interp.New(interp.Options{
+		GoPath: goPath,
+		Env:    os.Environ(),
+		Stdout: logs.NoLevel(logger, zerolog.DebugLevel),
+		Stderr: logs.NoLevel(logger, zerolog.ErrorLevel),
+	})
+
+	err := i.Use(stdlib.Symbols)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load symbols: %w", err)
+	}
+
+	err = i.Use(ppSymbols())
+	if err != nil {
+		return nil, fmt.Errorf("failed to load provider symbols: %w", err)
+	}
+
+	_, err = i.Eval(fmt.Sprintf(`import "%s"`, manifestImport))
+	if err != nil {
+		return nil, fmt.Errorf("failed to import plugin code %q: %w", manifestImport, err)
+	}
+	return i, nil
 }
--- a/pkg/plugins/logger.go
+++ b/pkg/plugins/logger.go
@@ -1,0 +1,32 @@
+package plugins
+
+import (
+	"context"
+
+	"github.com/http-wasm/http-wasm-host-go/api"
+	"github.com/rs/zerolog"
+)
+
+func initWasmLogger(logger *zerolog.Logger) *WasmLogger {
+	return &WasmLogger{
+		logger: logger,
+	}
+}
+
+// compile-time check to ensure ConsoleLogger implements api.Logger.
+var _ api.Logger = WasmLogger{}
+
+// WasmLogger is a convenience which writes anything above LogLevelInfo to os.Stdout.
+type WasmLogger struct {
+	logger *zerolog.Logger
+}
+
+// IsEnabled implements the same method as documented on api.Logger.
+func (w WasmLogger) IsEnabled(level api.LogLevel) bool {
+	return true
+}
+
+// Log implements the same method as documented on api.Logger.
+func (w WasmLogger) Log(_ context.Context, level api.LogLevel, message string) {
+	w.logger.WithLevel(zerolog.Level(level + 1)).Msg(message)
+}
--- a/pkg/plugins/middlewares.go
+++ b/pkg/plugins/middlewares.go
@@ -2,13 +2,18 @@ package plugins
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"net/http"
+	"os"
 	"path"
 	"reflect"
 	"strings"
 
+	"github.com/http-wasm/http-wasm-host-go/handler"
+	wasm "github.com/http-wasm/http-wasm-host-go/handler/nethttp"
 	"github.com/mitchellh/mapstructure"
+	"github.com/traefik/traefik/v3/pkg/middlewares"
 	"github.com/traefik/yaegi/interp"
 )
 
@@ -34,30 +39,73 @@ func (b Builder) Build(pName string, config map[string]interface{}, middlewareNa
 type middlewareBuilder struct {
 	fnNew          reflect.Value
 	fnCreateConfig reflect.Value
+	pluginType     string
+	wasmPath       string
 }
 
-func newMiddlewareBuilder(i *interp.Interpreter, basePkg, imp string) (*middlewareBuilder, error) {
+func findPluginType(wasmPath string) string {
+	if wasmPath != "" {
+		return PluginTypeWASM
+	}
+	return PluginTypeYaegi
+}
+
+func newMiddlewareBuilder(i *interp.Interpreter, basePkg, imp, wasmPath string) (*middlewareBuilder, error) {
+	builder := &middlewareBuilder{
+		wasmPath:   wasmPath,
+		pluginType: findPluginType(wasmPath),
+	}
+
+	if builder.pluginType == PluginTypeWASM {
+		return builder, nil
+	}
+
 	if basePkg == "" {
 		basePkg = strings.ReplaceAll(path.Base(imp), "-", "_")
 	}
+	var err error
 
-	fnNew, err := i.Eval(basePkg + `.New`)
+	builder.fnNew, err = i.Eval(basePkg + `.New`)
 	if err != nil {
 		return nil, fmt.Errorf("failed to eval New: %w", err)
 	}
 
-	fnCreateConfig, err := i.Eval(basePkg + `.CreateConfig`)
+	builder.fnCreateConfig, err = i.Eval(basePkg + `.CreateConfig`)
 	if err != nil {
 		return nil, fmt.Errorf("failed to eval CreateConfig: %w", err)
 	}
-
-	return &middlewareBuilder{
-		fnNew:          fnNew,
-		fnCreateConfig: fnCreateConfig,
-	}, nil
+	return builder, nil
 }
 
 func (p middlewareBuilder) newHandler(ctx context.Context, next http.Handler, cfg reflect.Value, middlewareName string) (http.Handler, error) {
+	if p.pluginType == PluginTypeWASM {
+		code, err := os.ReadFile(p.wasmPath)
+		if err != nil {
+			return nil, err
+		}
+		logger := middlewares.GetLogger(ctx, middlewareName, PluginTypeWASM)
+		opts := []handler.Option{
+			handler.Logger(initWasmLogger(logger)),
+		}
+		i := cfg.Interface()
+		if i != nil {
+			config, ok := i.(map[string]interface{})
+			if !ok {
+				return nil, fmt.Errorf("could not type assert config: %T", i)
+			}
+			b, err := json.Marshal(config)
+			if err != nil {
+				return nil, err
+			}
+			opts = append(opts, handler.GuestConfig(b))
+		}
+
+		mw, err := wasm.NewMiddleware(context.Background(), code, opts...)
+		if err != nil {
+			return nil, err
+		}
+		return mw.NewHandler(ctx, next), nil
+	}
 	args := []reflect.Value{reflect.ValueOf(ctx), reflect.ValueOf(next), cfg, reflect.ValueOf(middlewareName)}
 	results := p.fnNew.Call(args)
 
@@ -78,6 +126,9 @@ func (p middlewareBuilder) newHandler(ctx context.Context, next http.Handler, cf
 }
 
 func (p middlewareBuilder) createConfig(config map[string]interface{}) (reflect.Value, error) {
+	if p.pluginType == PluginTypeWASM {
+		return reflect.ValueOf(config), nil
+	}
 	results := p.fnCreateConfig.Call(nil)
 	if len(results) != 1 {
 		return reflect.Value{}, fmt.Errorf("invalid number of return for the CreateConfig function: %d", len(results))
--- a/pkg/plugins/plugins.go
+++ b/pkg/plugins/plugins.go
@@ -10,7 +10,11 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-const localGoPath = "./plugins-local/"
+const (
+	localGoPath     = "./plugins-local/"
+	PluginTypeWASM  = "wasm"
+	PluginTypeYaegi = "yaegi"
+)
 
 // SetupRemotePlugins setup remote plugins environment.
 func SetupRemotePlugins(client *Client, plugins map[string]Descriptor) error {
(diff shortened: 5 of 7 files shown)


zetaab:
cc @jcchavezs could you review this as well.

pkg/plugins/logger.go
```diff
@@ -0,0 +1,32 @@
+package plugins
+
+import (
+	"context"
+
+	"github.com/http-wasm/http-wasm-host-go/api"
+	"github.com/rs/zerolog"
+)
+
+func initWasmLogger(logger *zerolog.Logger) *WasmLogger {
```
zetaab, line 10: looks like it does not make big difference if we use zerolog here. The performance is still good.

pkg/plugins/logger.go
```diff
@@ -0,0 +1,32 @@
...
+}
+
+// IsEnabled implements the same method as documented on api.Logger.
+func (w WasmLogger) IsEnabled(level api.LogLevel) bool {
+	return true
+}
+
+// Log implements the same method as documented on api.Logger.
+func (w WasmLogger) Log(_ context.Context, level api.LogLevel, message string) {
+	w.logger.WithLevel(zerolog.Level(level + 1)).Msg(message)
```
jcchavezs, line 31: Shall we do `w.logger.WithContext(ctx).WithLevel(zerolog.Level(level + 1)).Msg(message)` here?

pkg/plugins/middlewares.go
```diff
@@ -34,30 +39,73 @@ func (b Builder) Build(pName string, config map[string]interface{}, middlewareNa
 type middlewareBuilder struct {
 	fnNew          reflect.Value
 	fnCreateConfig reflect.Value
+	pluginType     string
```
jcchavezs, line 42: Shall we use a enum-like here?

```
type pluginType int8 

const (
    PluginTypeYaegi pluginType = 0
    PluginTypeWasm pluginType = 1
)

File: pkg/plugins/plugins.go [hunk 4]
```diff
@@ -10,7 +10,11 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-const localGoPath = "./plugins-local/"
+const (
+	localGoPath     = "./plugins-local/"
+	PluginTypeWASM  = "wasm"
```
jcchavezs, line 15: See my comment in https://github.com/traefik/traefik/pull/10189/files#r1371295003

pkg/plugins/middlewares.go
```diff
@@ -34,30 +39,73 @@ func (b Builder) Build(pName string, config map[string]interface{}, middlewareNa
...
-
-	return &middlewareBuilder{
-		fnNew:          fnNew,
-		fnCreateConfig: fnCreateConfig,
-	}, nil
+	return builder, nil
 }
 
 func (p middlewareBuilder) newHandler(ctx context.Context, next http.Handler, cfg reflect.Value, middlewareName string) (http.Handler, error) {
+	if p.pluginType == PluginTypeWASM {
```
jcchavezs, line 81: I suggest to move this code inside a function.

pkg/plugins/logger.go
zetaab: Screenshot 2023-10-25 at 10 43 39

zerolog.UpdateContext is perhaps the correct one instead. However, not sure should this be really implemented

pkg/plugins/middlewares.go
```diff
@@ -34,30 +39,73 @@ func (b Builder) Build(pName string, config map[string]interface{}, middlewareNa
...
+		}
+		logger := middlewares.GetLogger(ctx, middlewareName, PluginTypeWASM)
+		opts := []handler.Option{
+			handler.Logger(initWasmLogger(logger)),
+		}
+		i := cfg.Interface()
+		if i != nil {
+			config, ok := i.(map[string]interface{})
+			if !ok {
+				return nil, fmt.Errorf("could not type assert config: %T", i)
```
jcchavezs: I wonder if this type assertion is required. I think the marshaling will do also some sort of type assertion,

pkg/plugins/plugins.go
```diff
@@ -144,12 +148,16 @@ func checkLocalPluginManifest(descriptor LocalDescriptor) error {
...
+		if m.Import == "" {
+			errs = multierror.Append(errs, fmt.Errorf("%s: missing import", descriptor.ModuleName))
+		}
 
-	if !strings.HasPrefix(m.Import, descriptor.ModuleName) {
-		errs = multierror.Append(errs, fmt.Errorf("the import %q must be related to the module name %q", m.Import, descriptor.ModuleName))
+		if !strings.HasPrefix(m.Import, descriptor.ModuleName) {
+			errs = multierror.Append(errs, fmt.Errorf("the import %q must be related to the module name %q", m.Import, descriptor.ModuleName))
+		}
+	} else if m.WasmPath == "" {
```
jcchavezs, line 161: Looks redudant, if `m.IsYaegiPlugin()` returns false it means that m.WasmPath is not empty.

zetaab: actually, if wasmpath is empty it will always think that the plugin is yaegi plugin. So basically this else is not executed at all. I will remove this

pkg/plugins/logger.go
jcchavezs: My bad. I meant `logger.Ctx(ctx)`

pkg/plugins/middlewares.go
```diff
@@ -34,30 +39,77 @@ func (b Builder) Build(pName string, config map[string]interface{}, middlewareNa
...
 		return nil, fmt.Errorf("failed to eval CreateConfig: %w", err)
 	}
+	return builder, nil
+}
 
-	return &middlewareBuilder{
-		fnNew:          fnNew,
-		fnCreateConfig: fnCreateConfig,
-	}, nil
+func newWasmHandler(ctx context.Context, next http.Handler, cfg reflect.Value, middlewareName, wasmPath string) (http.Handler, error) {
```
jcchavezs, line 80: I think this could be tested but probably in a next PR.

pkg/plugins/middlewares.go
```diff
@@ -78,6 +130,9 @@ func (p middlewareBuilder) newHandler(ctx context.Context, next http.Handler, cf
 }
 
 func (p middlewareBuilder) createConfig(config map[string]interface{}) (reflect.Value, error) {
+	if p.pluginType == PluginTypeWasm {
```
jcchavezs, line 133: I suggest a comment in here on why this happens.

pkg/plugins/middlewares.go
```diff
@@ -34,30 +39,77 @@ func (b Builder) Build(pName string, config map[string]interface{}, middlewareNa
...
+		return nil, err
+	}
+	return mw.NewHandler(ctx, next), nil
 }
 
 func (p middlewareBuilder) newHandler(ctx context.Context, next http.Handler, cfg reflect.Value, middlewareName string) (http.Handler, error) {
+	if p.pluginType == PluginTypeWasm {
+		return newWasmHandler(ctx, next, cfg, middlewareName, p.wasmPath)
+	}
 	args := []reflect.Value{reflect.ValueOf(ctx), reflect.ValueOf(next), cfg, reflect.ValueOf(middlewareName)}
```
jcchavezs, line 113: I suggest from here on you encapsulate it into newDefaultHandler or newYaegiHandler.

pkg/plugins/logger.go
zetaab: it does not have that either :)

pkg/plugins/middlewares.go
zetaab: not sure, but it does not cost much to double check

juliens:
Thank you for this PR, it's really great.

I'm curious about the possibility of introducing a new category of plugins called "wasm-middleware" instead of relying on the wasmPath. This approach has the potential to streamline the code and enhance its clarity. WDYT?

zetaab:
@juliens just to clarify: so you mean like we could have new category here https://github.com/traefik/traefik/blob/master/pkg/plugins/builder.go#L64 ?

juliens:
@zetaab Yep exactly, I'm not sure it's the right solution, but I'm not really fan of the "wasmPath switch".

zetaab:
Well I can try to modify it, lets see how it will look like

ldez:
I think it's better to add a new field like "category" instead of reusing the field "type"

juliens:
>  hmm or should it scan the folder structure and just take first *.wasm?
I don't like it either

I was thinking about the hardcoded way, but I was not convinced that's why I didn't talk about it. Let's keep the `wasmPath` for now.

zetaab:
added runtime field to manifest. It is now required value for wasm plugins. However, yaegi will fallback to empty value (which is currently the case in all plugins). Change to plugin https://github.com/zetaab/traefik-wasm-demo/commit/ea45bf5c3b45b5c4906c43d8b279e5d422d83a15

pkg/plugins/middlewares.go
```diff
@@ -34,30 +39,69 @@ func (b Builder) Build(pName string, config map[string]interface{}, middlewareNa
...
 		return nil, fmt.Errorf("failed to eval CreateConfig: %w", err)
 	}
+	return builder, nil
+}
 
-	return &middlewareBuilder{
-		fnNew:          fnNew,
-		fnCreateConfig: fnCreateConfig,
-	}, nil
+func (p middlewareBuilder) newWasmHandler(ctx context.Context, next http.Handler, cfg reflect.Value, middlewareName string) (http.Handler, error) {
```
juliens, line 75: Instead of having one `middlewareBuilder` and two methods named `newWasmHandler` and `newYaegiHandler`, you should have an interface and two implementations ( one for wasm and one for yaegi )

zetaab: its bigger change :)

pkg/plugins/middlewares.go
```diff
@@ -2,41 +2,53 @@ package plugins
...
 // Build builds a middleware plugin.
 func (b Builder) Build(pName string, config map[string]interface{}, middlewareName string) (Constructor, error) {
-	if b.middlewareBuilders == nil {
-		return nil, fmt.Errorf("no plugin definition in the static configuration: %s", pName)
+	if b.yaegiMiddlewareBuilders == nil || b.wasmMiddlewareBuilders == nil {
+		return nil, fmt.Errorf("no plugin definitions in the static configuration: %s", pName)
 	}
 
-	descriptor, ok := b.middlewareBuilders[pName]
+	// plugin (pName) can be located in yaegi or wasm middleware builders
```
zetaab, line 26: @juliens the logic is now divided to yaegiMiddlewareBuilders and wasmMiddlewareBuilders. All other functionality is in separate functions and classes but this Build function will build middlewares for wasm and yaegi. The main problem in this is that if someone has similar named yaegi and wasm plugin, it will always use yaegi. However, I do not see this quite big issue?

zetaab: hmm actually, that is not perhaps possible because the configurations are coming from yaml. It does not allow duplicate keys. It will load only one of them anyway

juliens: And I think that the problem already exists with two yaegi plugin.

pkg/plugins/plugins.go
jcchavezs: Not sure this was resolved

jcchavezs:
Amazing work @zetaab! what is missing here?

zetaab:
@juliens @geraldcroes is there something missing?

juliens:
@zetaab Can you take a look at https://github.com/zetaab/traefik/pull/2
Next week, I will be at KubeCon so I don't know if I will be able to answer but I will try to do my best.

pkg/plugins/middlewares.go
```diff
@@ -107,27 +116,87 @@ func (p middlewareBuilder) createConfig(config map[string]interface{}) (reflect.
...
+func newWasmMiddlewareBuilder(wasmPath string) *wasmMiddlewareBuilder {
+	return &wasmMiddlewareBuilder{
+		wasmPath: wasmPath,
+	}
+}
+
+func (p wasmMiddlewareBuilder) newWasmHandler(ctx context.Context, next http.Handler, cfg reflect.Value, middlewareName string) (http.Handler, error) {
+	code, err := os.ReadFile(p.wasmPath)
+	if err != nil {
+		return nil, err
```
jcchavezs, line 157:
```suggestion
		return nil, fmt.Errorf("loading Wasm binary: %w", err)
```

pkg/plugins/middlewares.go
```diff
@@ -107,27 +116,87 @@ func (p middlewareBuilder) createConfig(config map[string]interface{}) (reflect.
...
+	}
+	i := cfg.Interface()
+	if i != nil {
+		config, ok := i.(map[string]interface{})
+		if !ok {
+			return nil, fmt.Errorf("could not type assert config: %T", i)
+		}
+		b, err := json.Marshal(config)
+		if err != nil {
+			return nil, err
```
jcchavezs, line 172:
```suggestion
			return nil, fmt.Errorf("marshaling config: %w", err)
```

jcchavezs:
Can we merge this now?

geraldcroes:
> Can we merge this now?

The PR is being reviewed. As written in the project guidelines, it will require 3 lgtms to be merged.

We discussed it during triage, and we prioritized it's review. Judging from my perception of the quality it should be quick now.

jcchavezs:
Thanks @geraldcroes!

juliens (approved):
LGTM
I put the label no-merge, because I would like to modify the plugin catalog before merging that.
I worked on it yesterday, I hope we will be able to merge it soon.

jcchavezs:
Not sure I get the comment @ldez. What do you mean exactly?

zetaab:
I think he means that this current PR is using .wasm files from the repository files itself. However, the production version could use released artifacts from github release system. That needs some changes if people want it work like that.

jcchavezs:
> I think he means that this current PR is using .wasm files from the repository files itself. However, the production version could use released artifacts from github release system. That needs some changes if people want it work like that.

I might be missing something but I can't see any .wasm file in the PR. Could you point me to the code which uses wasm files in the repo?

image

zetaab:
@jcchavezs he means the plugins itself, not this repo. Check for instance my traefik wasm demo repository. It needs to have wasm file in repository. This PR defines that logic

jcchavezs:
I see. I think this was a temporary thing to show how it works but then it would change.

Basically we discussed this in the past (see https://github.com/dapr/components-contrib/issues/2700) and there are a couple of things to consider here:

1. I agree we should not ask the plugins to have the wasm file check out, they mainly won't. Instead we publish the artifacts in the realase page (see for example https://github.com/corazawaf/coraza-proxy-wasm/releases/tag/0.3.0).
2. One important thing to notice is that some wasm files might require a filesystem to be mounted along with the binary and hence we somehow convected to deliver the artifact as a tar.gz where the main binary comes and we also include ancillary files. If too complicated we can not support ancillary files for now and only grab the wasm file.

Also relevant: https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases

juliens:
I agree, we should use the github release.

Initially, I thought it might be too complicated for users to handle, but upon further reflection, while the automation could be complex, a user's task of simply 'uploading' the file on the release isn't that complicated. The plugin catalog could then filter releases based on whether the wasm file has been uploaded.

As I understand it, the only modification required in this pull request is establishing a convention for the plugin filename. The remaining changes would manifest in the plugin catalog code, shifting from relying on the GitHub repo to fetching plugins from the release.

> One important thing to notice is that some wasm files might require a filesystem to be mounted along with the binary and hence we somehow convected to deliver the artifact as a tar.gz where the main binary comes and we also include ancillary files. If too complicated we can not support ancillary files for now and only grab the wasm file.

@jcchavezs, I'm uncertain about this aspect. Are you suggesting that we might want to include other files in the release archive?"

zetaab:
@juliens compiling wasm artifact with github actions and uploading that to release is not difficult. Perhaps worth of having that in example repo.

It could work like:
1) user goes manually to github releases page
2) creates new release using existing tag / creates new tag
3) github actions pipeline is triggered
4) github actions runner will build the artifact and upload to the github release

jcchavezs:
> @jcchavezs, I'm uncertain about this aspect. Are you suggesting that we might want to include other files in the release archive?"

I am saying some plugins will require additional files to work. In general I would suggest when it comes to loading the plugin we don't do it too restrictive to traefik plugin catalog, http-wasm will embrace plugins outside that plugin catalog to be used so I suggest we stick with the simpler: accept a tar.gz and use the .wasm that is inside.

juliens:
>  it is now as file in github repo. Instead it should check the github releases and download the artifact using that.

@zetaab, I'm a bit uncertain about this. Currently in Traefik, the assumption is that it downloads an archive from the plugin catalog without really distinguishing whether it's from a GitHub repository or another source. That's why, from my perspective, removing the wasmPath seems like a simple change, relying solely on the naming convention. In the case of local plugins, I would assume that the corresponding directory contains the correct wasm file, without concerning ourselves with whether the source is present or not.

@jcchavezs, understood. The requirement is now clear. I agree that we should allow for an archive with files other than the wasm file, rather than limiting it to only a wasm file.

However, my suggestion is to start without the wasmPath. Introducing a wasmPath referencing an archive from anywhere requires more careful consideration.

ldez:
I'm not sure about the requirement of `wasmPath`.
Also if this option is required it should not be able to download external files for security reasons.

tomMoulard:
As a side project, I've implemented a GitHub action that build fail2ban plugin's WASM file and put it as an asset in the release. You can find my **WIP testing** scenario [here](https://github.com/tomMoulard/fail2ban/blob/72e149173f651f50e378ae0ef169fdbfeaadd178/.github/workflows/release.yml). It should be easy to add it to any plugin without much changes.
But why not fetching and building each plugin on Traefik's startup (with caching), thus providing a simple interface for plugin's authors. This would increase startup time (the first time a new plugin is loaded), but would not rely on a plugin release asset. One of the caveats would be to include everything that is needed to build such WASM asset "on the fly".

jcchavezs:
> But why not fetching and building each plugin on Traefik's startup (with caching)

I would avoid this for different reasons, first, you need tinygo to build the wasm binary and tinygo versions can defer and also means traefik plugin source must be aware of that. Also some compilations could include code generation which also make it go dependent. In general I think this would lead us to non deterministic builds.

I would see the way to make it work fetching it from releases page.

immanuelfodor:
Just an idea folks, could maybe the [plugin catalog](https://plugins.traefik.io/plugins) do the fetching and provide a unified interface for Proxy to download the necessary files? It already does the plugin YAML meta fetching, could it maybe do the release fetching, too?

ldez:
> Just an idea folks, could maybe the [plugin catalog](https://plugins.traefik.io/plugins) do the fetching and provide a unified interface for Proxy to download the necessary files? It already does the plugin YAML meta fetching, could it maybe do the release fetching, too?

It's already how it works.

juliens:
In summary, my proposal is as follows:
We should include a traefik-plugin.zip file in each release. This zip file must contain, at a minimum, the plugin.wasm file and the .traefik.yml file. Additional files can also be included in this zip file.

In Traefik, as is currently practiced, we will retrieve a zip file from the plugin catalog and extract its contents into the appropriate directory. Within this directory, we will locate the .traefik.yml file, the plugin.wasm file, and any other files present in the traefik-plugin.zip asset.

If everyone is in agreement, the only modification needed for this implementation is to eliminate the wasmPath and instead directly read the hardcoded plugin.wasm file.

What are your thoughts on this?

zetaab:
@juliens that is then hardcoded to have one plugin per repository. However, I think it was like that already before?

jcchavezs:
I think I understand now. I agree with this proposal. I would only dispute
the naming convention, I think plugin is a traefik concept and I would vote
for using the *first* wasm file being found and let the others to be
mounted. This way we provide more flexibility.

juliens:
> @juliens that is then hardcoded to have one plugin per repository. However, I think it was like that already before?
@zetaab yes, it's already like this for Yaegi plugin so I think this is ok.

zetaab:
so @juliens are you going to need needed changes to this PR or how this pr can be merged? Basically you can push commits to this PR or open new one after this is merged, to modify the behaviour of wasmPath

go.mod
```diff
@@ -33,6 +33,7 @@ require (
 	github.com/hashicorp/go-retryablehttp v0.7.4
 	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/nomad/api v0.0.0-20220506174431-b5665129cd1f
+	github.com/http-wasm/http-wasm-host-go v0.5.2-0.20231031144646-6e96a183660d
```
mmatur, line 36: @zetaab do you if it is planned to have a release v0.5.2  version

pkg/plugins/builder.go
```diff
@@ -36,97 +41,123 @@ func NewBuilder(client *Client, plugins map[string]Descriptor, localPlugins map[
...
+	}
+
+	// plugin (pName) can be located in yaegi or wasm middleware builders.
+	if descriptor, ok := b.middlewareBuilders[pName]; ok {
+		m, err := descriptor.newMiddleware(config, middlewareName)
 		if err != nil {
-			return nil, fmt.Errorf("%s: failed to read manifest: %w", desc.ModuleName, err)
+			return nil, err
 		}
+		return m.NewHandler, nil
```
mmatur, line 122:
```suggestion
		}
		
		return m.NewHandler, nil
```

mmatur (changes requested):

pkg/plugins/middlewarewasm.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/traefik/traefik/v3/pkg/middlewares"
+)
+
+type wasmMiddlewareBuilder struct {
+	wasmPath string
+}
+
+func newWasmMiddlewareBuilder(goPath string, moduleName string) *wasmMiddlewareBuilder {
+	return &wasmMiddlewareBuilder{
+		wasmPath: filepath.Join(goPath, "src", moduleName, "plugin.wasm"),
```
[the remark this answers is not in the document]
jcchavezs: Let's stick with this now but I truly hope we don't end up requiring projects to use `plugin.wasm` convention or them ending up having two releases because `plugin.wasm` isn't meaningful in other ecosystems (e.g. envoy)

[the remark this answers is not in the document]
ldez: After a discussion with @juliens we will follow this:

- if there is no explicit filename (or path) in the manifest, use `plugin.wasm`.
- if there is an explicit filename (or path) defined in the manifest, use it.

go.mod
jcchavezs: It is out now https://github.com/http-wasm/http-wasm-host-go/releases/tag/v0.5.2

pkg/plugins/middlewarewasm.go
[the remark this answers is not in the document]
jcchavezs: sounds reasonable.

go.mod
jcchavezs: cc @zetaab @ldez

jcchavezs:
@mmatur dependency is upgraded to v0.5.2, could you approve?

ieugen:
this is interesting.
What wasm runtime will run the wasm binaries?
Would it support WASM garbage collection and Tail calls features? 
Both features where standardized and are showing up in browsers https://webassembly.org/roadmap/ . 

I am asking because I would like to experiment using guile-hoot to write some middleware for Traefik 
https://spritely.institute/files/docs/guile-hoot/latest/Introduction.html

jcchavezs:
The runtime is wazero which is a Go runtime and already provides a host
implementation for http wasm. As for GC and Tail calls, AFAIK no other
runtimes besides V8 support it but @evacchi can give better details on the
plan for wazero. Still wasm extensions have existed and performed at prod
level for a while without this so I would not consider this anyhow a
blocker.

ldez (approved):
LGTM.

We have tested the PR with the new demo plugin: https://github.com/traefik/plugindemowasm

The 2 systems (Wasm, Yaegi) work as expected.

PR #10231, opened 2023-11-10
chore: update github.com/docker/docker

kevinpollet:
### What does this PR do?

This pull request updates the github.com/docker/docker dependency to https://github.com/moby/moby/compare/v20.10.21...v23.0.7.

### Motivation

Because of the number of issues related to CVEs reported by scanning tools, we are updating the docker dependency.

Please note that Traefik is not affected by those CVEs. They affect the server code of the library, and Traefik is only using the client code.

### More

- [ ] Added/updated tests


ldez (changes requested):
This update is incomplete: docker doesn't have a go.mod and transitive dependencies should be checked and updated individually by hand.

kevinpollet:
Closed in favor of https://github.com/traefik/traefik/pull/10239

PR #10233, opened 2023-11-12
Fixed datadog json format issue

### LINKED ISSUE #10208: Datadog tracer logs are not json formatted
### Welcome!

### What did you do?

Hello, I've been investigating an issue we initially spotted in our Fluentd log collector and traced it down to usage of the datadog tracer in Traefik.

We have Traefik configured in our kubernetes as the ingress controller, and in this cluster we are also running datadog.

In order to get traces from Traefik to Datadog, we enabled the traces in Traefik by adding the following to the traefik container `args:`

```
      - --tracing.datadog=true
      - --tracing.datadog.localAgentHostPort=$(DD_AGENT_HOST):8126
      - --tracing.datadog.prioritySampling=true
```

And we set the `DD_AGENT_HOST` environment variable:

```
      env:
      - name: DD_ENTITY_ID
        valueFrom:
          fieldRef:
            apiVersion: v1
            fieldPath: metadata.uid
      - name: DD_AGENT_HOST
        value: datadog.datadog.svc.cluster.local
```

In Traefik, we use json logging format:

```
      - --log.format=json
```

However, we're noticing some log parsing errors in Fluentd due to logs coming from Traefik which aren't json-formatted.

On investigation, we found that traefik is outputting messages which are initially created by the datadog tracer. So I opened an issue with their team to ask for a fix and from what I understood in the [response](https://github.com/DataDog/dd-trace-go/issues/2152#issuecomment-1790586933), Traefik simply needs to provide its json logger to the tracer in order to make the logs json formatted.

Could this be implemented wherever is appropriate in order to ensure that logs output from external tooling are json formatted?

### What did you see instead?

Log messages which are not json formatted

2.10.4

```yaml
```

_No response_

sssash18:
### What does this PR do?

This PR tries to fix #10208 , in which datadog logs were not logged in json format.

### Motivation

This PR now passes the logger in datadog to allow json format.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/tracing/datadog/logger.go
+++ b/pkg/tracing/datadog/logger.go
@@ -1,0 +1,15 @@
+package datadog
+
+import "github.com/traefik/traefik/v2/pkg/log"
+
+type DatadogLogger struct {
+	logger log.Logger
+}
+
+func newDatadogLogger(logger log.Logger) *DatadogLogger {
+	return &DatadogLogger{logger: logger}
+}
+
+func (d DatadogLogger) Log(msg string) {
+	d.logger.Debug(msg)
+}
--- a/pkg/tracing/datadog/datadog.go
+++ b/pkg/tracing/datadog/datadog.go
@@ -1,6 +1,7 @@
 package datadog
 
 import (
+	"context"
 	"io"
 	"net"
 	"os"
@@ -47,6 +48,8 @@ func (c *Config) SetDefaults() {
 
 // Setup sets up the tracer.
 func (c *Config) Setup(serviceName string) (opentracing.Tracer, io.Closer, error) {
+	ctx := log.With(context.Background(), log.Str(log.MetricsProviderName, "datadog"))
+
 	opts := []datadog.StartOption{
 		datadog.WithService(serviceName),
 		datadog.WithDebugMode(c.Debug),
@@ -56,6 +59,7 @@ func (c *Config) Setup(serviceName string) (opentracing.Tracer, io.Closer, error
 			PriorityHeader: c.SamplingPriorityHeaderName,
 			BaggagePrefix:  c.BagagePrefixHeaderName,
 		})),
+		datadog.WithLogger(newDatadogLogger(log.FromContext(ctx))),
 	}
 
 	if c.LocalAgentSocket != "" {


pkg/tracing/datadog/logger.go
```diff
@@ -0,0 +1,15 @@
...
+	logger log.Logger
+}
+
+func newDatadogLogger(logger log.Logger) *DatadogLogger {
+	return &DatadogLogger{logger: logger}
+}
+
+func (d DatadogLogger) Log(msg string) {
+	d.logger.Debug(msg)
+}
```
mloiseleur, line 15: Why force `Debug` level on Log() ?
It seems to be explicitly [supported by zerolog](https://github.com/rs/zerolog#logging-without-level-or-message)

sssash18: @mloiseleur yes I agree with you, only took it as it's implemented the same way in branch v3.0, so took it for consistency. Should I remove it ?

sssash18: @mloiseleur Actually in v2.10 the logger used is not zerolog , its "github.com/traefik/traefik/v2/pkg/log" and does not have any such general log function.

mloiseleur: ah, that was introduced in v3, that's right !

mmatur (changes requested):
Thanks @sssash18 for your contribution.

WDYT to move the `logger.go` file into the `log` package and rename the file to `datadog.go`

mmatur (approved):
LGTM

rtribotte (approved):
LGTM

pkg/log/datadog.go
```diff
@@ -0,0 +1,13 @@
...
+type DatadogLogger struct {
+	logger Logger
+}
+
+func NewDatadogLogger(logger Logger) *DatadogLogger {
+	return &DatadogLogger{logger: logger}
+}
+
+func (d DatadogLogger) Log(msg string) {
+	d.logger.Error(msg)
```
rtribotte, line 12:
```suggestion
	d.logger.Debug(msg)
```

rtribotte (changes requested):
Hello @sssash18,

Upon further consideration, I propose utilizing the DEBUG level for the logs in question. It seems that the dd-trace library not only generates ERROR level logs but also others, which might be overly assertive if treated as errors in Traefik logs.

Given this dilemma, it appears that setting all log levels of the dd-trace library to DEBUG in Traefik would be a prudent approach to maintain a balanced logging output.

pkg/log/datadog.go
sssash18: Made the changes @rtribotte . Please review it again.

ldez (approved):

rtribotte (approved):
Thanks

PR #10237, opened 2023-11-14
fix: improve confusing suggested value on `tracing.openTelemetry.grpc`

mloiseleur:
### What does this PR do?

Clarify a confusing suggestion on YAML configuration for `tracing.openTelemetry.grpc`

### Motivation

Fixes traefik/traefik-helm-chart#964

### More

### DIFF
--- a/docs/content/observability/tracing/opentelemetry.md
+++ b/docs/content/observability/tracing/opentelemetry.md
@@ -233,7 +233,7 @@ This instructs the reporter to send spans to the OpenTelemetry Collector using g
 ```yaml tab="File (YAML)"
 tracing:
   openTelemetry:
-    grpc: {}
+    grpc: true
 ```
 
 ```toml tab="File (TOML)"
--- a/docs/content/observability/tracing/opentelemetry.md
+++ b/docs/content/observability/tracing/opentelemetry.md
@@ -233,7 +233,7 @@ This instructs the reporter to send spans to the OpenTelemetry Collector using g
 ```yaml tab="File (YAML)"
 tracing:
   openTelemetry:
-    grpc: true
+    grpc: {}
 ```
 
 ```toml tab="File (TOML)"
@@ -242,5 +242,5 @@ tracing:
 ```
 
 ```bash tab="CLI"
---tracing.openTelemetry.grpc=true
+--tracing.openTelemetry.grpc
 ```


docs/content/observability/tracing/opentelemetry.md
```diff
@@ -242,5 +242,5 @@ tracing:
 ```
 
 ```bash tab="CLI"
---tracing.openTelemetry.grpc=true
+--tracing.openTelemetry.grpc
```
ldez, line 245:
```suggestion
--tracing.openTelemetry.grpc=true
```

ldez (changes requested):

docs/content/observability/tracing/opentelemetry.md
mloiseleur: It would be nicer to be as consistent as possible.

@ldez Can you detail why it should be a boolean in CLI and an empty struct in YAML ?

ldez: because an empty struct is a boolean inside the CLI. The CLI is not JSON or YAML.

`--tracing.openTelemetry.grpc` is equal to `--tracing.openTelemetry.grpc=true` => a boolean value.

mloiseleur: I understand from your explanation that both are working. That's also what I observed in my tests.

A real fix for this confusion could be to align CLI struct with JSON and YAML. FTM, I don't know how to make this kind of fix.

As a workaround, since they are equal and one of them, `--tracing.openTelemetry.grpc=true`, can make uninformed user believe a boolean is expected, the other is not, then why not helping users to be less confused ?

ldez: CLI cannot be aligned with JSON and YAML because CLI uses simple values and not objects.

An empty struct is not `struct{}` (it's just a specific case), it's a struct that has fields but the fields are not filled.
For example, you need to enable Docker but using the default configuration.

Can you accept my suggestion?

mloiseleur: I would be happy to accept your suggestion, you often have interesting point of view.

I understand that aligning CLI conf with TOML and YAML is subtle and delicate.

Since `--tracing.openTelemetry.grpc` is less confusing and works, I do not understand why you are suggesting to keep the more confusing `--tracing.openTelemetry.grpc=true` in the doc ?

ldez: We always document the boolean option (real boolean or empty struct) with an explicit value to be homogenous and avoid confusion for real CLI users.

mloiseleur: :thinking: What is a real CLI user ? Is there false CLI user ? I do not understand what that means.

I agree with you that there is value to be homogenous. 

Since the boolean option in CLI can represent a real boolean or an empty struct, wdyt about:

* documenting all _real boolean_ conf on CLI with `--xx.yy=true`
* documenting all _empty struct_ conf on CLI with `--xx.yy`

nmengin:
Hello @mloiseleur,

Thank you for the proposition.

After discussing your Pull Request with other maintainers we prefer to close it.
Indeed, we would prefer discussing the modification for the whole documentation rather than only for one field.
If you agree, could you please open such a proposal?

Moreover, note that the field `tracing.openTelemetry.grpc` will be deleted once the new [OTel PR](https://github.com/traefik/traefik/pull/10223) is merged.

PR #10242, opened 2023-11-15
Refuse recursive requests (CVE-2023-47633)

rtribotte:
### What does this PR do?

This PR introduces an internal middleware that is automatically appended to the default rule's routers, to stop request recursion on Traefik.

### Motivation

The exposure of the Traefik container combined with the default rule mechanism can lead to create a router targeting itself in a loop.
In this case, to prevent an infinite loop, Traefik adds an internal middleware to refuse the request if already been passed by the same router.

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: Michael <michael.matur@gmail.com>

### DIFF
--- a/pkg/config/dynamic/http_config.go
+++ b/pkg/config/dynamic/http_config.go
@@ -42,12 +42,12 @@ type Service struct {
 
 // Router holds the router configuration.
 type Router struct {
-	EntryPoints   []string         `json:"entryPoints,omitempty" toml:"entryPoints,omitempty" yaml:"entryPoints,omitempty" export:"true"`
-	Middlewares   []string         `json:"middlewares,omitempty" toml:"middlewares,omitempty" yaml:"middlewares,omitempty" export:"true"`
-	Service       string           `json:"service,omitempty" toml:"service,omitempty" yaml:"service,omitempty" export:"true"`
-	Rule          string           `json:"rule,omitempty" toml:"rule,omitempty" yaml:"rule,omitempty"`
-	Priority      int              `json:"priority,omitempty" toml:"priority,omitempty,omitzero" yaml:"priority,omitempty" export:"true"`
-	TLS           *RouterTLSConfig `json:"tls,omitempty" toml:"tls,omitempty" yaml:"tls,omitempty" label:"allowEmpty" file:"allowEmpty" kv:"allowEmpty" export:"true"`
+	EntryPoints []string         `json:"entryPoints,omitempty" toml:"entryPoints,omitempty" yaml:"entryPoints,omitempty" export:"true"`
+	Middlewares []string         `json:"middlewares,omitempty" toml:"middlewares,omitempty" yaml:"middlewares,omitempty" export:"true"`
+	Service     string           `json:"service,omitempty" toml:"service,omitempty" yaml:"service,omitempty" export:"true"`
+	Rule        string           `json:"rule,omitempty" toml:"rule,omitempty" yaml:"rule,omitempty"`
+	Priority    int              `json:"priority,omitempty" toml:"priority,omitempty,omitzero" yaml:"priority,omitempty" export:"true"`
+	TLS         *RouterTLSConfig `json:"tls,omitempty" toml:"tls,omitempty" yaml:"tls,omitempty" label:"allowEmpty" file:"allowEmpty" kv:"allowEmpty" export:"true"`
 	DefaultRule bool             `json:"-" toml:"-" yaml:"-" label:"-" file:"-"`
 }
 
--- a/pkg/provider/configuration.go
+++ b/pkg/provider/configuration.go
@@ -415,7 +415,7 @@ func BuildRouterConfiguration(ctx context.Context, configuration *dynamic.HTTPCo
 			}
 
 			// Flag default rule routers to add the denyRouterRecursion middleware.
-			router.IsDefaultRule = true
+			router.DefaultRule = true
 		}
 
 		if router.Service == "" {
--- a/pkg/provider/consulcatalog/config_test.go
+++ b/pkg/provider/consulcatalog/config_test.go
@@ -49,9 +49,9 @@ func TestDefaultRule(t *testing.T) {
 				HTTP: &dynamic.HTTPConfiguration{
 					Routers: map[string]*dynamic.Router{
 						"Test": {
-							Service:       "Test",
-							Rule:          "Host(`foo.bar`)",
-							IsDefaultRule: true,
+							Service:     "Test",
+							Rule:        "Host(`foo.bar`)",
+							DefaultRule: true,
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{},
@@ -100,9 +100,9 @@ func TestDefaultRule(t *testing.T) {
 				HTTP: &dynamic.HTTPConfiguration{
 					Routers: map[string]*dynamic.Router{
 						"Test": {
-							Service:       "Test",
-							Rule:          `Host("Test.foo.bar")`,
-							IsDefaultRule: true,
+							Service:     "Test",
+							Rule:        `Host("Test.foo.bar")`,
+							DefaultRule: true,
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{},
@@ -235,9 +235,9 @@ func TestDefaultRule(t *testing.T) {
 				HTTP: &dynamic.HTTPConfiguration{
 					Routers: map[string]*dynamic.Router{
 						"Test": {
-							Service:       "Test",
-							Rule:          "Host(`Test`)",
-							IsDefaultRule: true,
+							Service:     "Test",
+							Rule:        "Host(`Test`)",
+							DefaultRule: true,
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{},
@@ -321,9 +321,9 @@ func Test_buildConfiguration(t *testing.T) {
 				HTTP: &dynamic.HTTPConfiguration{
 					Routers: map[string]*dynamic.Router{
 						"dev-Test": {
-							Service:       "dev-Test",
-							Rule:          "Host(`dev-Test.traefik.wtf`)",
-							IsDefaultRule: true,
+							Service:     "dev-Test",
+							Rule:        "Host(`dev-Test.traefik.wtf`)",
+							DefaultRule: true,
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{},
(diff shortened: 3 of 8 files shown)


pkg/config/dynamic/http_config.go
```diff
@@ -48,6 +48,7 @@ type Router struct {
 	Rule        string           `json:"rule,omitempty" toml:"rule,omitempty" yaml:"rule,omitempty"`
 	Priority    int              `json:"priority,omitempty" toml:"priority,omitempty,omitzero" yaml:"priority,omitempty" export:"true"`
 	TLS         *RouterTLSConfig `json:"tls,omitempty" toml:"tls,omitempty" yaml:"tls,omitempty" label:"allowEmpty" file:"allowEmpty" kv:"allowEmpty" export:"true"`
+	LoopStop    bool             `json:"-" toml:"-" yaml:"-" label:"-" file:"-"`
```
ldez, line 51: This option is not useful for now and this option can be introduced in another form (like a int instead of bool)
This option should be removed.

ldez (changes requested):

pkg/config/dynamic/http_config.go
mmatur: Why this option is not useful? I don't have the full context. 

This option is private, no one can configure it. I don't see the benefit of having an `int`.

You suggest removing this option but you also suggest using an `int` instead of a  bool. I'm a bit lost to be honest with what you are looking for.

ldez: My fault I read it too quickly.

juliens, line 51: I wonder if something like DefaultRouterRule would be a better name? WDYT?

pkg/middlewares/loopstop/loop_stop.go
```diff
@@ -0,0 +1,63 @@
...
+		routerName:     routerName,
+		routerNameHash: hash,
+		next:           next,
+	}, nil
+}
+
+// ServeHTTP implements http.Handler.
+func (l *LoopStop) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if r.Header.Get(xTraefikRouter) == l.routerNameHash {
+		log.WithoutContext().Debugf("Rejecting request in provenance of the same default rule's router (%q) to stop potential infinite loop.", l.routerName)
```
lbenguigui, line 45:
```suggestion
		log.WithoutContext().Debugf("Rejecting request in provenance of the same router (%q) to stop potential infinite loop.", l.routerName)
```
You have no guarantee that you come from a default rule from here.

pkg/middlewares/loopstop/loop_stop.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"crypto/sha1"
+	"fmt"
+	"net/http"
+
+	"github.com/containous/alice"
+	"github.com/traefik/traefik/v2/pkg/log"
+)
+
+const xTraefikRouter = "X-Traefik-Router"
```
kevinpollet, line 12:
```suggestion
const headerXTraefikRouter = "X-Traefik-Router"
```

pkg/middlewares/loopstop/loop_stop.go
```diff
@@ -0,0 +1,63 @@
...
+		return
+	}
+
+	r.Header.Set(xTraefikRouter, l.routerNameHash)
+
+	l.next.ServeHTTP(w, r)
+}
+
+func makeHash(routerName string) (string, error) {
+	h := sha1.New()
```
kevinpollet, line 57: Why not using a fast hash, fnv? Do we need a cryptographic hash?

pkg/middlewares/loopstop/loop_stop.go
```diff
@@ -0,0 +1,63 @@
...
+	if _, err := h.Write([]byte(routerName)); err != nil {
```
kevinpollet, line 58: As explained in the godoc, this never returns an error so we can ignore it (see https://pkg.go.dev/hash#Hash).

pkg/middlewares/loopstop/loop_stop.go
```diff
@@ -0,0 +1,63 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/containous/alice"
+	"github.com/traefik/traefik/v2/pkg/log"
+)
+
+const xTraefikRouter = "X-Traefik-Router"
+
+type LoopStop struct {
```
kevinpollet, line 14: WDYT of naming it `DenyRecursion`, `DenyRouterRecursion`?

pkg/middlewares/loopstop/loop_stop_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestServeHTTP(t *testing.T) {
+	req, err := http.NewRequest(http.MethodGet, "", nil)
+	require.NoError(t, err)
+
+	m, err := NewLoopStop("myRouter", http.HandlerFunc(func(_ http.ResponseWriter, _ *http.Request) {}))
```
kevinpollet, line 16: I think it could be great to assert that the next router is called and not called when we have a recursion.

pkg/middlewares/loopstop/loop_stop_test.go
```diff
@@ -0,0 +1,30 @@
...
+	require.NoError(t, err)
+
+	recorder := httptest.NewRecorder()
+	m.ServeHTTP(recorder, req)
+
+	assert.Equal(t, http.StatusOK, recorder.Code)
+
+	require.Equal(t, m.routerNameHash, req.Header.Get(xTraefikRouter))
```
kevinpollet, line 24:
```suggestion
	assert.Equal(t, m.routerNameHash, req.Header.Get(xTraefikRouter))
```

lbenguigui, line 24: Don't you think you should test that `xTraefikRouter` is the hash of `myRouter`?
If for some reason `m.routerNameHash` is empty, the test will pass but the middleware is broken.

pkg/middlewares/loopstop/loop_stop.go
```diff
@@ -0,0 +1,63 @@
...
+
+	return &LoopStop{
+		routerName:     routerName,
+		routerNameHash: hash,
+		next:           next,
+	}, nil
+}
+
+// ServeHTTP implements http.Handler.
+func (l *LoopStop) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
kevinpollet, line 43:
```suggestion
func (l *LoopStop) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
```

kevinpollet (changes requested):

pkg/middlewares/loopstop/loop_stop.go
rtribotte: We have used the same naming convention as in the Forwarded middleware.

juliens (approved):
LGTM

pkg/provider/configuration.go
```diff
@@ -413,9 +413,12 @@ func BuildRouterConfiguration(ctx context.Context, configuration *dynamic.HTTPCo
 				delete(configuration.Routers, routerName)
 				continue
 			}
+
+			// Flag default rule routers to add the loopStop middleware.
```
lbenguigui, line 417:
```suggestion
			// Flag default rule routers to add the denyRouterRecursion middleware.
```

lbenguigui (approved):
LGTM

kevinpollet (approved):
LGTM

pkg/middlewares/denyrouterrecursion/deny_router_recursion_test.go
```diff
@@ -0,0 +1,42 @@
...
+	m, err := New("myRouter", http.HandlerFunc(func(_ http.ResponseWriter, _ *http.Request) {
+		next++
+	}))
+	require.NoError(t, err)
+
+	recorder := httptest.NewRecorder()
+	m.ServeHTTP(recorder, req)
+
+	assert.Equal(t, http.StatusOK, recorder.Code)
+
```
ldez, line 29:
```suggestion
```

pkg/middlewares/denyrouterrecursion/deny_router_recursion_test.go
```diff
@@ -0,0 +1,42 @@
...
+	assert.Equal(t, "995d26092d19a224", m.routerNameHash)
+
```
ldez, line 31:
```suggestion
```

pkg/middlewares/denyrouterrecursion/deny_router_recursion_test.go
```diff
@@ -0,0 +1,42 @@
...
+	assert.Equal(t, m.routerNameHash, req.Header.Get(xTraefikRouter))
+
```
ldez, line 33:
```suggestion
```

pkg/middlewares/denyrouterrecursion/deny_router_recursion_test.go
```diff
@@ -0,0 +1,42 @@
...
+	assert.Equal(t, 1, next)
+
+	recorder = httptest.NewRecorder()
+	m.ServeHTTP(recorder, req)
+
+	assert.Equal(t, 1, next)
+
```
ldez, line 40:
```suggestion
```

pkg/config/dynamic/http_config.go
```diff
@@ -42,12 +42,13 @@ type Service struct {
...
-	Priority    int              `json:"priority,omitempty" toml:"priority,omitempty,omitzero" yaml:"priority,omitempty" export:"true"`
-	TLS         *RouterTLSConfig `json:"tls,omitempty" toml:"tls,omitempty" yaml:"tls,omitempty" label:"allowEmpty" file:"allowEmpty" kv:"allowEmpty" export:"true"`
+	EntryPoints   []string         `json:"entryPoints,omitempty" toml:"entryPoints,omitempty" yaml:"entryPoints,omitempty" export:"true"`
+	Middlewares   []string         `json:"middlewares,omitempty" toml:"middlewares,omitempty" yaml:"middlewares,omitempty" export:"true"`
+	Service       string           `json:"service,omitempty" toml:"service,omitempty" yaml:"service,omitempty" export:"true"`
+	Rule          string           `json:"rule,omitempty" toml:"rule,omitempty" yaml:"rule,omitempty"`
+	Priority      int              `json:"priority,omitempty" toml:"priority,omitempty,omitzero" yaml:"priority,omitempty" export:"true"`
+	TLS           *RouterTLSConfig `json:"tls,omitempty" toml:"tls,omitempty" yaml:"tls,omitempty" label:"allowEmpty" file:"allowEmpty" kv:"allowEmpty" export:"true"`
+	IsDefaultRule bool             `json:"-" toml:"-" yaml:"-" label:"-" file:"-"`
```
ldez, line 51:
```suggestion
	DefaultRule bool             `json:"-" toml:"-" yaml:"-" label:"-" file:"-"`
```

ldez (changes requested):

pkg/server/router/router.go
```diff
@@ -202,6 +203,10 @@ func (m *Manager) buildHTTPHandler(ctx context.Context, router *runtime.RouterIn
 		chain = chain.Append(metricsMiddle.WrapRouterHandler(ctx, m.metricsRegistry, routerName, provider.GetQualifiedName(ctx, router.Service)))
 	}
 
+	if router.IsDefaultRule {
```
rtribotte, line 206:
```suggestion
	if router.DefaultRule {
```

ldez (approved):

PR #10247, opened 2023-11-16
Adds KeepAliveMax* features

juliens:
### What does this PR do?

This PR adds 2 options to control how we want to close keep alive connection

### Motivation

Sometimes, LB in front of Traefik are "stick" to a traefik instance because of keep alive, this feature allow to mitigate this problem by forcing to renew connections.

### More

- [x] Added/updated tests

### Additional Notes

This PR adds a new var in expvar to be able to follow what happens on connections.

### DIFF
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -3,6 +3,7 @@ package server
 import (
 	"context"
 	"errors"
+	"expvar"
 	"fmt"
 	stdlog "log"
 	"net"
@@ -36,11 +37,16 @@ var httpServerLogger = stdlog.New(log.WithoutContext().WriterLevel(logrus.DebugL
 
 type key string
 
-const connState key = "connState"
+const connStateKey key = "connState"
 
-type ConnState struct {
-	start     time.Time
-	idleCycle int
+var connStates = map[string]*connState{}
+var connStatesMu = sync.RWMutex{}
+
+type connState struct {
+	State          string
+	KeepAliveState string
+	Start          time.Time
+	IdleCycle      int
 }
 
 type httpForwarder struct {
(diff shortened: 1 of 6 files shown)


ldez:
there are some unrelated changes, I think it's because you are using the latest version of golangci-lint.
I think it's better to do that in another PR.
I will create a PR with only golangci-lint changes.

pkg/server/server_entrypoint_tcp.go
```diff
@@ -548,13 +570,57 @@ func createHTTPServer(ctx context.Context, ln net.Listener, configuration *stati
 		})
 	}
 
+	var lastHandler http.Handler
+	if debug || (configuration.Transport != nil && (configuration.Transport.KeepAliveMaxTime > 0 || configuration.Transport.KeepAliveMaxRequests > 0)) {
+		lastHandler = http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
```
kevinpollet, line 575: Here I would do like above and have a dedicated unexposed function, to ease the readability (the function name is just an example) and have something like the following at the end:

```go
if configuration.Transport.KeepAliveMaxTime > 0 || configuration.Transport.KeepAliveMaxRequests > 0 {
handler = closeKeepAliveConnections(...)
}
```

kevinpollet (changes requested):

kevinpollet (approved):
LGTM 🚀

lbenguigui (approved):
LGTM

rtribotte (approved):
LGTM 👌

hejingyong123456:
If data is being transmitted when the connection is disconnected, it may cause the data to not be transmitted properly. How should we solve this situation

hejingyong123456:
If data is being transmitted when the connection is disconnected, it may cause the data to not be transmitted properly. How should we solve this situation

PR #10249, opened 2023-11-17
Deprecate IPWhiteList middleware in favor of IPAllowList

lbenguigui:
### What does this PR do?

This PR deprecates IPWhiteList middleware in favor of the IPAllowList middleware.

### More

- [x] Added/updated tests

### Additional Notes


peter-englmaier:
This middleware could also allow setting a redirection URL in case the IP is not on the AllowList instead of 403 forbidden. If that were the case, it would be useful to redirect users to a maintenance page while still letting devops login and test the application.

jspdown:
@peter-englmaier Thanks for the feedback. 

This PR only deprecates the old IPWhiteList middleware in favor of the IPAllowList middleware and doesn't add any new behavior/feature. 

Could you open a dedicated issue for tracking this enhancement proposal? This way we might be able to gather more feedback from the community to see if there's a need for it and if so how it could be tackled.

rtribotte (approved):
LGTM

juliens (approved):
LGTM

peter-englmaier:
> Could you open a dedicated issue for tracking this enhancement proposal? This way we might be able to gather more feedback from the community to see if there's a need for it and if so how it could be tackled.

Sure, I can do that. I think it would be useful for many.

jspdown (approved):

PR #10252, opened 2023-11-19
fix: use matching or only service for routers

joshka:
### What does this PR do?

When the router configuration does not specify a serviceName, and a
service with the same name as the router is found, use that service.
If there is only a single service, use that service.

This allows the configuration to omit configuring the serviceName for
routers that have an identical name to the service.

Fixes: https://github.com/traefik/traefik/issues/4807
Also: https://community.traefik.io/search?q=%22too%20many%20services%22

### Motivation

The "too many services" error message is ambiguous and does not match the documentation that suggests this behavior should just work.
See
- https://doc.traefik.io/traefik/routing/routers/#service
- https://doc.traefik.io/traefik/routing/providers/docker/#service-definition

"If a label defines a router (e.g. through a router Rule) and a label defines a service (e.g. implicitly through a loadbalancer server port value), but the router does not specify any service, then that service is automatically assigned to the router."

### More

- [ ] Added/updated tests
  - No tests for configuration.go
  - documentation already states this behavior

### Additional Notes

I've made this PR on the master branch - I'm not sure if it should be put on the v2 branch or v3 (or both?).

I've also never writtn anything in golang before - so apologies if this is not idiomatic - feel free to roast / suggest fixes.

### DIFF
--- a/pkg/provider/configuration.go
+++ b/pkg/provider/configuration.go
@@ -388,18 +388,28 @@ func BuildTCPRouterConfiguration(ctx context.Context, configuration *dynamic.TCP
 			continue
 		}
 
-		if len(router.Service) == 0 {
-			if len(configuration.Services) > 1 {
-				delete(configuration.Routers, routerName)
-				loggerRouter.Error().
-					Msg("Could not define the service name for the router: too many services")
-				continue
-			}
+		if len(router.Service) > 0 {
+			continue
+		}
 
+		// When the router does not have a service, we try to find one in the following order:
+		// - a service with the same name as the router
+		// - the only service defined in the configuration
+		if _, isServiceDefined := configuration.Services[routerName]; isServiceDefined {
+			router.Service = routerName
+			continue
+		}
+		if len(configuration.Services) == 1 {
 			for serviceName := range configuration.Services {
 				router.Service = serviceName
 			}
+			loggerRouter.Info().
+				Msg("No service defined matching the router name, using the only one defined in the configuration")
+			continue
 		}
+		delete(configuration.Routers, routerName)
+		loggerRouter.Error().
+			Msg("Could not define the service name for the router. More than 1 service and none match the router name. Please specify the service name in the router configuration.")
 	}
 }
 
@@ -412,17 +422,24 @@ func BuildUDPRouterConfiguration(ctx context.Context, configuration *dynamic.UDP
 			continue
 		}
 
-		if len(configuration.Services) > 1 {
-			delete(configuration.Routers, routerName)
-			loggerRouter.
-				Error().Msg("Could not define the service name for the router: too many services")
+		// When the router does not have a service, we try to find one in the following order:
+		// - a service with the same name as the router
+		// - the only service defined in the configuration
+		if _, isServiceDefined := configuration.Services[routerName]; isServiceDefined {
+			router.Service = routerName
 			continue
 		}
-
-		for serviceName := range configuration.Services {
-			router.Service = serviceName
-			break
+		if len(configuration.Services) == 1 {
+			for serviceName := range configuration.Services {
+				router.Service = serviceName
+			}
+			loggerRouter.Info().
+				Msg("No service defined matching the router name, using the only one defined in the configuration")
+			continue
 		}
+		delete(configuration.Routers, routerName)
+		loggerRouter.Error().
+			Msg("Could not define the service name for the router. More than 1 service and none match the router name. Please specify the service name in the router configuration.")
 	}
 }
 
@@ -456,18 +473,28 @@ func BuildRouterConfiguration(ctx context.Context, configuration *dynamic.HTTPCo
 			}
 		}
 
-		if len(router.Service) == 0 {
-			if len(configuration.Services) > 1 {
-				delete(configuration.Routers, routerName)
-				loggerRouter.Error().
-					Msg("Could not define the service name for the router: too many services")
-				continue
-			}
+		if len(router.Service) > 0 {
+			continue
+		}
 
+		// When the router does not have a service, we try to find one in the following order:
+		// - a service with the same name as the router
+		// - the only service defined in the configuration
+		if _, isServiceDefined := configuration.Services[routerName]; isServiceDefined {
+			router.Service = routerName
+			continue
+		}
+		if len(configuration.Services) == 1 {
 			for serviceName := range configuration.Services {
 				router.Service = serviceName
 			}
+			loggerRouter.Info().
+				Msg("No service defined matching the router name, using the only one defined in the configuration")
+			continue
 		}
+		delete(configuration.Routers, routerName)
+		loggerRouter.Error().
+			Msg("Could not define the service name for the router. More than 1 service and none match the router name. Please specify the service name in the router configuration.")
 	}
 }
 


ldez:
Hello,

> When the router configuration does not specify a serviceName, and a
service with the same name as the router is found, use that service.
If there is only a single service, use that service.

This is not an expected behavior:
- a service with an identical name as the router is not necessarily the right service
- if a router doesn't have a service (it can be created automatically) but it should not default to a "random" service.

The behavior should be deterministic.

joshka:
> if a router doesn't have a service (it can be created automatically) it should not default to a "random" service.

The current behavior of the code is that if there is a single service then that service is used for the router. This PR keeps that behavior. It's not a "Random service", it's the only service that is defined.

Can you explain the rationale behind this:

> a service with an identical name as the router is not necessarily the right service

As a user I expected the following to work (lines commented were the necessary addition):
```yaml
      - traefik.enable=true
      - traefik.http.services.calibre.loadbalancer.server.port=8080
      - traefik.http.services.calibre-webserver.loadbalancer.server.port=8081
      # - traefik.http.routers.calibre.service=calibre
      - traefik.http.routers.calibre.entrypoints=websecure
      - traefik.http.routers.calibre.rule=Host(`calibre.home.joshka.net`)
      - traefik.http.routers.calibre.tls.certresolver=letsencrypt
      # - traefik.http.routers.calibre-webserver.service=calibre-webserver
      - traefik.http.routers.calibre-webserver.entrypoints=websecure
      - traefik.http.routers.calibre-webserver.rule=Host(`calibre-webserver.home.joshka.net`)
      - traefik.http.routers.calibre-webserver.tls.certresolver=letsencrypt
```

Why? Most docker containers that I setup have a single relevant port that is hooked up to single router and the `.services` line is unnecessary. So when adding a second service to a container, my mental model changes from the router and the service share the same name so they just work to. I have to manually hook up the router and the service even though they have the same name. This seems unnecessary.

>The error message "Could not define the service name for the router: too many services" is not really related to this sentence of the documentation.

For me, it's directly related. I read that documentation and understood that if I have an implicitly defined service then that will be associated with a router of the same name - because that's what happens in the single service case.

> The issue linked to this PR is not a bug but a question related to an invalid configuration.

When several people interact with software and experience an error based on a faulty understanding of how the software works, it's worth considering the software is faulty rather than the understanding.

This seems like a good default, what are the downsides I'm missing?

ldez:
> The current behavior of the code is that if there is a single service then that service is used for the router. This PR keeps that behavior. 

Not really, your PR starts by trying to find and use a service with a name identical to the router instead of trying to check if there is only one service.
It's clearly not the same behavior.

> It's not a "Random service", it's the only service that is defined.

It takes a service just based on a name so it's "random" because the name can be defined by the user.

> When several people interact with software and experience an error based on a faulty understanding of how the software works, it's worth considering the software is faulty rather than the understanding.

A lot of people have 404, so it's the fault of Traefik?
This argument is not really strong: if several users have the same errors it's just because configuration errors are possible, we are all humans.

joshka:
>Not really, your PR starts by trying to find and use a service with a name identical to the router instead of trying to check > if there is only one service.
> It's clearly not the same behavior.

If there is only one service then the behavior prior to this PR is to use that service.
The behavior after this PR  is to use that service also? It doesn't matter whether the check for the same name happens first as there is only one service.

> It takes a service just based on a name so it's "random" because the name can be defined by the user and this kind of magic convention is unexpected.

Unexpected by whom? At least 3 counterpoints to 'expected' here (issue, forum posts, my experience).

How is there is just one service and just one router and so we hook one to the other any less magical than there are two services and two routers and we hook the ones with the same names to each other?

> This argument is not really relevant: if several users have the same errors it's just because configuration errors are possible, we are all humans.

And good defaults exist for the same reason - we're human

joshka:
>I can understand your usage, but it's just one of the possible usages.

What fails due to this change?

joshka:
> A lot of people have 404, so it's the fault of Traefik?

Yes actually. If the standard of configuration that people fall into encourages people to fall into error cases rather than be successful in setting up a service then the Traefik is wrong, not the people.

ldez:
> Yes actually. If the standard of configuration that people fall into encourages people to fall into error cases rather than be successful in setting up a service then the Traefik is wrong, not the people.

But this is not the case, the standard of configuration doesn't encourage that.
You can think that errors are always because Traefik is wrong but it's not true, and I am sure that don't think that.

This PR is not a bug fix, and it is not related to Docker only but to the core of the configuration handling for all the providers. Traefik handles several providers with different constraints and different ways to build routing. The internal convention based on the service number is fully controlled and stable because it's internal and cannot be confused by a user mistake.

joshka:
Strong disagreement with everything you wrote here, but it seems I cannot convince you. C'est la vie.

ldez:
Maybe I missed something but I don't see how it's possible to disagree with everything I wrote here, but anyway, I just tried my best to explain the problems I saw.

I'm not the only maintainer, the discussion can continue, and maybe someone will see what I haven't seen.

joshka:
Specifically:

> But this is not the case, the standard of configuration doesn't encourage that.

The current standard of configuration is that traefik accepts when a single router and a single service is defined by the user to mean that the single service and router correspond to each other regardless of the service / router name. The user can configure the router and service without having to link them explicitly.

When a user transitions from a situation were there is a single service to now having two services, when the second service is added, the configuration that previously worked fine for the first service now fails. This is a failure caused by something peripherally related (a service with an entirely different name) rather than a failure due to something directly related. Put another way, in order to add a second service I have to reconfigure both the existing router and the new router. This is counter-intuitive as shown by the intuition of every person who has previously configured this "incorrectly".

- https://github.com/traefik/traefik/issues/4807
- https://community.traefik.io/t/https-h2c-could-not-define-the-service-name-for-the-router-too-many-services/7092/2
- https://community.traefik.io/t/could-not-define-the-service-name-for-the-router-too-many-services/5514
- https://community.traefik.io/t/too-many-services-with-more-subdomains/15865
- https://community.traefik.io/t/service-router-wont-get-created/6769
- https://community.traefik.io/t/https-middleware-not-firing-for-certain-stack/3284
- https://community.traefik.io/t/bad-gateway-on-portainer/16384/3
- https://community.traefik.io/t/traefik-with-lots-of-routes-in-one-docker-service/9270
- https://community.traefik.io/t/2-services-in-one-docker-one-hides-the-other/6651

In each of the above cases someone has added a service and a router with the same name with 2 or more services and expected that their configuration should work. I'd bet there are many more that hit the same error, and googled the issue and found one of those posts.

The impact of this PR is *ONLY* to change a broken but reasonable configuration into a working one. It has zero impact on any working configuration. The only possible issue is that it adds extra meaning to the name when there are 2 or more services. This is a good thing though. Where does that break things?

mloiseleur:
@joshka If I understand you correctly, what you are saying is that:

With one service, it works without specifying the service, like that:

```yaml
      - traefik.enable=true
      - traefik.http.services.calibre.loadbalancer.server.port=8080
      - traefik.http.services.calibre-webserver.loadbalancer.server.port=8081
      - traefik.http.routers.calibre.entrypoints=websecure
      - traefik.http.routers.calibre.rule=Host(`calibre.home.joshka.net`)
      - traefik.http.routers.calibre.tls.certresolver=letsencrypt
```

When adding another service, it fails with `too many services error` until user figures out he needs to explicitly link router with the service:

```yaml
      - traefik.http.routers.calibre.service=calibre
```

And the same for the other service. 
Your proposal is to ease user life by autolinking router and service when user choose same name for both.
This step would stay mandatory when service and router names are different. 

Is that correct ?

joshka:
>Your proposal is to ease user life by autolinking router and service when user choose same name for both.
This step would stay mandatory when service and router names are different.

>Is that correct ?

Yes

mloiseleur:
Thanks for this interesting proposal. 
I opened [an issue](https://github.com/traefik/traefik/issues/10258) to study what can really solve this issue. This proposal is not the only way, we can take the time to see what would work best.

ldez:
FYI @mloiseleur is **NOT** a Traefik maintainers.

mloiseleur:
Indeed, I'm a random contributor trying to help.

joshka:
Ah got it :)

joshka:
Closing as there seems to be no interest from the devs in making this progress.

emilevauge:
I forgot an important aspect in my previous message: thank you @mloiseleur for being involved on this. Even though this proposal won't be implemented, your contribution was very helpful to identify weak points 🙏

PR #10255, opened 2023-11-20
Fixed stripPrefix middleware is not applied to retried attempts

niki-timofe:
### What does this PR do?

Before the fix, when using `stripPrefix` and `retry` middleware, retries were sent with the original, not the stripped URL.

### Motivation

We faced with this issue in production.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/middlewares/retry/retry.go
+++ b/pkg/middlewares/retry/retry.go
@@ -96,8 +96,12 @@ func (r *retry) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			},
 		}
 		newCtx := httptrace.WithClientTrace(req.Context(), trace)
+		newReq := req.WithContext(newCtx)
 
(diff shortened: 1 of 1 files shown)


mloiseleur:
\o Thanks for this PR @niki-timofe !

Do you think you can add a test to ensure this won't happen again ?

kevinpollet:
Hello @niki-timofe, and thanks for this pull request.

After reviewing it, we agreed that there is an issue when using the StripPrefix middleware with the Retry one.

However, a deep copy (https://pkg.go.dev/net/http#Request.Clone) of the request is needed, because the StripPrefix middleware also mutates the request headers.

Could you make the changes?

niki-timofe:
@kevinpollet, done!

lbenguigui (approved):
LGTM

kevinpollet (approved):
Thanks 👍

mmatur (approved):
LGTM

noptwork:
Thanks.

PR #10277, opened 2023-12-02
Add TCP loadBalancer sourceIPs

mzealey:
### What does this PR do?

In haproxy and other loadbalancers, a `source-ip` can be specified for each connection. Linux by default only uses the primary IP of a given interface, but to get > 60k simultaneous connections to a single remote IP, multiple source IPs are required. This patch should allow an optional list of `sourceIPs` to be specified for outbound TCP connections and will use one at random if so.

I have not fully tested this patch yet, but it compiles and parses correctly and seems relatively harmless.

### Motivation

We want to use traefik to handle a large number of long-lived TCP connections.

### More

- [ ] Added/updated tests
- [ ] Added/updated documentation (partially - not sure about toml)

### Additional Notes

I'm guessing this may need to be implemented elsewhere, however TCP streaming was sufficient for our use-case.

### DIFF
--- a/docs/content/routing/services/index.md
+++ b/docs/content/routing/services/index.md
@@ -1363,6 +1363,10 @@ Currently, the two available kinds are `LoadBalancer`, and `Weighted`.
 
 The servers load balancer is in charge of balancing the requests between the servers of the same service.
 
+An additional option of `sourceIPs` is provided which should be a list of TCP
+source addresses belonging to your server to randomly use for the outbound
+connections if you have multiple IP addresses on an interface.
+
 ??? example "Declaring a Service with Two Servers -- Using the [File Provider](../../providers/file.md)"
 
     ```yaml tab="YAML"
@@ -1374,6 +1378,9 @@ The servers load balancer is in charge of balancing the requests between the ser
             servers:
             - address: "xx.xx.xx.xx:xx"
             - address: "xx.xx.xx.xx:xx"
+            sourceIPs:
+            - 192.168.1.1
+            - 192.168.1.2
     ```
 
     ```toml tab="TOML"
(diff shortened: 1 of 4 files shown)


mzealey:
Unfortunately there does not seem to be a simple way in Linux to achieve this, other than specifying the source IP address, much less something that is cross-platform portable. All large-scale proxy servers that I have seen have this sort of option for precisely this reason. I'm not convinced that 'get all IPs of the specified interface' or of the server is a sensible approach as there may be complex routing setups.

juliens:
Since we haven't found a "simple" Linux solution, we've decided to accept this pull request for review in the upcoming minor version (v3.1). Would you mind shifting your PR to the master branch and performing a rebase?

docs/content/routing/services/index.md
```diff
@@ -1363,6 +1363,10 @@ Currently, the two available kinds are `LoadBalancer`, and `Weighted`.
 
 The servers load balancer is in charge of balancing the requests between the servers of the same service.
 
+An additional option of `sourceIPs` is provided which should be a list of TCP
+source addresses belonging to your server to randomly use for the outbound
+connections if you have multiple IP addresses on an interface.
```
jspdown, line 1368:
```suggestion
Moreover, the `sourceIPs` option enables the configuration of a list of IPs, 
from which the source address for outbound connections will be selected. 
Each new connection will randomly choose a source address from the specified list.
```

pkg/tcp/proxy.go
```diff
@@ -90,15 +93,21 @@ func (p *Proxy) ServeTCP(conn WriteCloser) {
 }
 
 func (p Proxy) dialBackend() (*net.TCPConn, error) {
```
jspdown, line 95: Could you add a test, similar to `TestLookupAddress`, to make sure a source IP is obtained from a configured list?

jspdown (changes requested):

nmengin:
Hello @mzealey,

Thanks for your contribution.

Have you planned to move forward on this PR?

PR #10279, opened 2023-12-04
[In-progress PoC] dumping connection logs

mzealey:
I'd appreciate a review of this code to let me know if I'm on a track that you'd be happy to eventually upstream... I have tried to implement as separate wrapper functions so that they can be toggled based on whether this config parameter exists or not with no change to other code.

Unfortunately, the only way I could make it so that the TLS and TCP connections dumped in the same place was to change the signature of `ServeTCP` to include a context object similar to HTTP's `req.Context()` as I did not want to extend `tcp.WriterCloser` to include context. I feel like there should be a better way to do this, but I'm not a golang expert.

### What does this PR do?

Creates a 'connectionlog' similar to 'accesslog' but on a per-TCP connection level (later can be extended UDP/QUIC)

### Motivation

We wish to account for bandwidth more accurately than just the http payload. In order to do this we currently account:
- plain data (ie post-TLS decryption)
- raw data (ie pre-TLS decryption)
- TCP overhead and other stats via the kernel where possible.

Currently it's just dumping JSON to the standard log, however we will extend it by copying a fair bit of the `accesslog` code to make it relatively compatible with this.

The idea is that with appropriate connection open/close timestamps and source ip/port this could then be correlated with the accesslog to create a fuller view of the connection.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/middlewares/tcp/inflightconn/inflight_conn.go
+++ b/pkg/middlewares/tcp/inflightconn/inflight_conn.go
@@ -38,9 +38,9 @@ func New(ctx context.Context, next tcp.Handler, config dynamic.TCPInFlightConn, 
 }
 
 // ServeTCP serves the given TCP connection.
-func (i *inFlightConn) ServeTCP(conn tcp.WriteCloser) {
-	ctx := middlewares.GetLoggerCtx(context.Background(), i.name, typeName)
-	logger := log.FromContext(ctx)
+func (i *inFlightConn) ServeTCP(ctx context.Context, conn tcp.WriteCloser) {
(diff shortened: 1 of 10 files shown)


emilevauge:
Hey @mzealey, do you still plan to finalize this PR?

mzealey:
Unsure if we will progress with this & other patchsets, but feel free to continue them if you would like, or close off and i'll re-open if we progress.

nmengin:
Hello @mzealey,

Thank you for your feedback.
I close the PR accordingly, feel free to open a new one if you can move forward on this topic.

PR #10281, opened 2023-12-05
Allow user-configurable compression level

mzealey:
### What does this PR do?

Allow user to specify level of compression

### Motivation

We want to use the Best level rather than the default

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/config/dynamic/middlewares.go
+++ b/pkg/config/dynamic/middlewares.go
@@ -154,6 +154,9 @@ type Compress struct {
 	// MinResponseBodyBytes defines the minimum amount of bytes a response body must have to be compressed.
 	// Default: 1024.
 	MinResponseBodyBytes int `json:"minResponseBodyBytes,omitempty" toml:"minResponseBodyBytes,omitempty" yaml:"minResponseBodyBytes,omitempty" export:"true"`
+	// CompressionLevel defines the compression level (-1 = default, 0-9).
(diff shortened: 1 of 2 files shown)


mloiseleur:
Yes, it can help on some use cases. 
Thanks for this contribution, @mzealey.

Do you think you can add test and documentation on this new parameter ?

kevinpollet:
Hello @mzealey, and thank you for your contribution.

Since configuring the compression level is considered as an enhancement, could you please rebase this pull request on the v3.0 branch?"

kevinpollet:
Hey @mzealey,

Do you want us to move forward on it?

By the way, as we released the version RC version of Traefik v3, this pull request has to be rebased on the master branch.

kevinpollet:
Hello @mzealey,

As this pull request is an enhancement we wanted to integrate it in the next minor version of Traefik v3, but we introduced the support of the Brotli compression algorithm in v3 which brings some questions.

To keep things simple we were thinking that introducing a unique option for gzip and brotli could be a good fit. This option could have the following values: `BestSpeed`, `BestCompression`, `DefaultCompression` and maybe `NoCompression`. Another option could be to have a dedicated option for every compression algorithm.

We prefer to discuss the different options in a proposal, and to see the traction on the subject before going further. Could you please open a proposal for this?

Therefore, I am declining this pull request.

PR #10283, opened 2023-12-06
docs: fix default value for anonymous usage statistics references

ariyonaty:
### What does this PR do?

This PR updates misleading documentation regarding the default value (`false`) of the anonymous usage option.

### Motivation

When looking through the CLI reference documentation, noticed room for clarfication with the `--global.sendanonymoususage` option, where the default value was said to be `false`, however the description said, "If the option is not specified, it will be enabled by default.". 

This PR aims to help clear up ambiguity for both the CLI Ref and the ENV ref docs regarding the option's default value.

### More

- [ ] Added/updated tests

### Additional Notes

This will be my first PR to Traefik. Hopefully can make some more impactful contributions in the future.

### DIFF
--- a/docs/content/reference/static-configuration/cli-ref.md
+++ b/docs/content/reference/static-configuration/cli-ref.md
@@ -211,7 +211,7 @@ plugin's version.
 Periodically check if a new version has been released. (Default: ```true```)
 
 `--global.sendanonymoususage`:  
-Periodically send anonymous usage statistics. If the option is not specified, it will be enabled by default. (Default: ```false```)
+Periodically send anonymous usage statistics. If the option is not specified, it will be enabled by default. (Default: ```true```)
 
 `--hostresolver`:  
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

the default value of `sendanonymoususage` send anonymous usage is `false`, the problem is related to the description, not the value.

kevinpollet:
Hello @ariyonaty, and thanks for opening this pull request,

@ldez is right, the default value of the `global.sendAnonymousUsage` option is `false` but the description should be: 

```
Periodically send anonymous usage statistics. If the option is not specified, it will be disabled by default. (Default: ```false```)
```

To fix the command description, you have to edit it here: https://github.com/traefik/traefik/blob/master/pkg/config/static/static_config.go#L99, and to generate the reference documentation with `go generate`.

Could you please open another pull request? 

In the meantime, I will close this pull request.

PR #10284, opened 2023-12-06
Allow TLS sessionTicketKeys config

### LINKED ISSUE #4550: Add option to configure TLS session ticket keys

Feature

Traefik uses TLS session tickets, those are TLS parameter, encrypted by a server-side key, that are send to the server in order to do session resumption and this way speed up the initial handshake.

The problem: As soon as you run more than one instance you need to synchronize those keys, otherwise the tickets become useless. So I was looking for a way to set them. The documentation, doesn't mention it, and searching for the [golang-function to set them](https://golang.org/pkg/crypto/tls/#Config.SetSessionTicketKeys), brings up empty results for Traefik.

I expected some kind of config parameter in the TLS config to configure one (or even better multiple) of those keys, to run TLS with full steam.

mzealey:
### What does this PR do?

Allow specification of custom TLS session ticket keys.

Fixes #4550 

### Motivation

From the added docs:

This option allows you to specify a list of TLS Session Ticket Keys which store 
the TLS connection state on the client and allow the TLS session resumption 
with less overhead than re-establishing it from scratch. This is used in both 
TLSv1.2 Session Tickets and TLSv1.3 PSK algorithms. 
 
The use-case for this is 3-fold: 
1. If you run a cluster of Traefik servers you can put the same Session Ticket 
   Keys on all servers and then a client can use it for lightweight session 
   re-establishment to any of the servers in the cluster. By default each 
   server will have its own random keys so a ticket valid for one server would 
   not be valid for the others. 
2. Restart traefik without loosing TLS ticket validity. As the default ticket 
   keys are stored in memory, when Traefik restarts they will be lost. By 
   writing them to the config file they are persisted across restarts 
3. Custom cycling and retention periods: The default daily rotation and 7-day 
   retention periods are not customizable. You may wish to write a cronjob to 
   update the Traefik config file more or less frequently, with longer or 
   shorter retention periods. 
 
If you use this option you are responsible for key rotation, which should occur 
at least on a daily basis or Perfect Forward Secrecy is compromised. These keys 
should be stored securely or your TLS sessions may be intercepted and 
decrypted. 
 
If not specified, the default golang crypto/tls algorithm is used which is to 
keep 7 days of keys and rotate on a daily basis.  

### More

- [ ] Added/updated tests - Unsure how to write automated tests for this so basic procedure documented below

### Additional Notes

Test like:

```
$ openssl s_client -no_tls1_3 -connect server:443 -sess_out sess < /dev/null 2>&1 |egrep 'SSL handsh|Reused|subject='
subject=CN = ssl-test4.systest.moya.app
SSL handshake has read 4171 bytes and written 316 bytes
$ openssl s_client -no_tls1_3 -connect server:443 -sess_in sess < /dev/null 2>&1 |egrep 'SSL handsh|Reused|subject='
subject=CN = ssl-test4.systest.moya.app
SSL handshake has read 287 bytes and written 427 bytes
Reused, TLSv1.3, Cipher is ECDHE-ECDSA-AES128-GCM-SHA256
```

Restarting the server and running the second command again will currently show the first output (ie no `Reused` entry), but with the patch will show a `Reused` line.

Tested with and without `-no_tls1_3` to test 1.2 and 1.3 as they use different methods.

### DIFF
--- a/docs/content/https/tls.md
+++ b/docs/content/https/tls.md
@@ -505,6 +505,76 @@ spec:
     - h2
 ```
 
+### TLS Session Ticket Keys
+
+_Optional_
+
+This option allows you to specify a list of TLS Session Ticket Keys which store
+the TLS connection state on the client and allow the TLS session resumption
+with less overhead than re-establishing it from scratch. This is used in both
+TLSv1.2 Session Tickets and TLSv1.3 PSK algorithms.
+
+The use-case for this is 3-fold:
+1. If you run a cluster of Traefik servers you can put the same Session Ticket
+   Keys on all servers and then a client can use it for lightweight session
+   re-establishment to any of the servers in the cluster. By default each
+   server will have its own random keys so a ticket valid for one server would
+   not be valid for the others.
+2. Restart traefik without loosing TLS ticket validity. As the default ticket
+   keys are stored in memory, when Traefik restarts they will be lost. By
+   writing them to the config file they are persisted across restarts
+3. Custom cycling and retention periods: The default daily rotation and 7-day
+   retention periods are not customizable. You may wish to write a cronjob to
+   update the Traefik config file more or less frequently, with longer or
+   shorter retention periods.
+
+If you use this option you are responsible for key rotation, which should occur
+at least on a daily basis or Perfect Forward Secrecy is compromised. These keys
+should be stored securely or your TLS sessions may be intercepted and
+decrypted.
+
+If not specified, the default golang crypto/tls algorithm is used which is to
+keep 7 days of keys and rotate on a daily basis. 
+
+These should be a list of 32-byte random strings which have been
+base64-encoded. You could generate valid key strings via a command like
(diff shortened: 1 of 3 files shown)


kevinpollet:
Hello @mzealey,

Thanks for your contribution and sorry for the delay,

Please could you rebase this pull request on branch master? 
We can handle it if you want!

sdelicata (changes requested):

kevinpollet:
Hello @mzealey,

Are you still working on this pull request?

kevinpollet:
Hello @mzealey,

Do not hesitate to ping us if anything changes on your end.

xjewer:
@kevinpollet are you still keen to merge such feature request if PR to be resurrected/recreated? or is anyone else working on it?

kevinpollet:
Hi @xjewer,

> are you still keen to merge such feature request if PR to be resurrected/recreated? or is anyone else working on it?

As far as I know, no one is working on this, so feel free to open a new pull request.

PR #10286, opened 2023-12-06
chore: update to newer dependencies in package.json and regen yarn.lock

### LINKED ISSUE #9932: Upgrade UI development dependencies
### Welcome!

nickboldt:
### What does this PR do?

this is specifically to get latest browserify-sign 4.2.2, but other updates are good too

I can offer a similar change for 2.10 and 3.0 branches, but I need this against 2.9 as it's the last stream that uses golang 1.20. 

### Motivation

Fixes #9932

### More

- [ ] Added/updated tests


ldez (changes requested):
Hello,

the v2.9 is deprecated, we don't accept PRs on this branch, you should target at least v2.10.

https://doc.traefik.io/traefik/deprecation/releases/

https://raw.githubusercontent.com/traefik/traefik/master/.github/PULL_REQUEST_TEMPLATE.md

nickboldt:
> you should target at least v2.10.

Can I target 2.10.5? I need something that uses golang 1.20, not 1.21. 

https://github.com/traefik/traefik/blob/v2.10/go.mod#L3

I see your main (er., master)  branch is still on golang 1.20. https://github.com/traefik/traefik/blob/master/go.mod 

Could 2.10 be similarly rolled back to build with 1.20?

ldez:
> Can I target [2.10.4](https://github.com/traefik/traefik/blob/v2.10.4/go.mod)?

you should target the branch [v2.10](https://github.com/traefik/traefik/tree/v2.10)

> Could 2.10 be similarly rolled back to build with 1.20?

no, because we want to use go1.21.

> Why is master on 1.20 but 3.0 is on 1.21?

Because have to do a mergeback of the branch v3.0 to the branch master.
We will handle that soon.

nickboldt:
> no, because we want to use go1.21.

OK, I guess I have no choice but to fork your repo and fix this myself, at least until go 1.21 exists as an RPM in RHEL8 or 9. 

Closing the PR.

PR #10298, opened 2023-12-14
Update ratelimit.md

rmburton:
A better configuration example has the burst set to a higher value than the average. A lower value burst will either have no effect or effectively reduce the average to the burst value.

### What does this PR do?

Update the documentation for RateLimit to provide a better example.

### Motivation

I was confused by the example on the RateLimit documentation since it implied that you should use a smaller value for burst than for average. I had to find a different source (https://traefik.io/blog/the-ultimate-guide-to-managing-application-traffic/#implementing-rate-limiting) to confirm that burst should be set to a higher value than average.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/middlewares/http/ratelimit.md
+++ b/docs/content/middlewares/http/ratelimit.md
@@ -16,15 +16,15 @@ It is based on a [token bucket](https://en.wikipedia.org/wiki/Token_bucket) impl
 
 ```yaml tab="Docker"
 # Here, an average of 100 requests per second is allowed.
-# In addition, a burst of 50 requests is allowed.
+# In addition, a burst of 200 requests is allowed.
 labels:
   - "traefik.http.middlewares.test-ratelimit.ratelimit.average=100"
-  - "traefik.http.middlewares.test-ratelimit.ratelimit.burst=50"
+  - "traefik.http.middlewares.test-ratelimit.ratelimit.burst=200"
 ```
 
 ```yaml tab="Kubernetes"
 # Here, an average of 100 requests per second is allowed.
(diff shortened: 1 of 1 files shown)


jspdown:
The `ratelimit` middleware implements the [Token Bucket](https://en.wikipedia.org/wiki/Token_bucket) algorithm, which can be pictured as a bucket full of water, each time a request comes in, it removes one unit of water from the bucket, if the bucket is empty, the request is delayed until the bucket have enough water. The average along with the period is used to compute the rate at which the buckets fills up.

Therefore, your updated example implies:
- You cannot receive more than 200 req simultaneously.
- If you are rate-limited, you will have to wait a 10ms (rate of 100/s) before doing another call.
- If you are rate-limited, you would have to wait for 2s before being able to consume another 200 req simultaneously.

Before your PR, the only thing different is the amount of requests you were allowed to make simultaneously.

The burst as 2 effects:
- It limits the number of requests that can be made simultaneously
- The lower it is the smoother your traffic will be (an empty bucket implies that requests will be delayed)

The original example and yours aren't wrong, they just present different use cases.

nmengin (approved):
LGTM

kevinpollet (approved):
Thanks 👍

jspdown:
@rmburton Thanks for your PR. As explained in my first comment. Your changes don't actually fix the example but instead present a different use case. We are not against this change and we therefore agreed to merge it. 

You are very welcome to open more PR and issue if you find things that can be improved/fixed. Thanks.

jspdown (approved):

PR #10302, opened 2023-12-15
Optimize memory usage on the TLS manager/store

ddtmachado:
### What does this PR do?

Optimize the memory footprint of TLS certificate handling on the TLS Manager and Store, mainly:

* Avoid parsing the certificate multiple times
* Avoid copying certificates back and forth multiple times
* Do not wipe the store on every refresh (dynamic config update)

I included a small bench test for the manager in the PR, it's goal is to switch TLS certificates in batches of 10 from a pool of 100. With it I observed a large decrease from +10k allocs/op average to 1.9k allocs/op average as follows:

```
#Before
122390966 ns/op	 3410744 B/op	   10843 allocs/op

#After
704874 ns/op	  187272 B/op	    1981 allocs/op
```

### Motivation

Since v2 Traefik received many reports with concerns and problems caused by the increased memory usage, specially while handling TLS certs.

I had some free time to investigate and found the code on the manager and store overly complex and probably trading memory for compatibility without strong justification, other than "I won't risk touching this code" maybe? :D

### More

- [ ] Added/updated tests

### Additional Notes

I did not had time to properly test it in a real environment with real data. Needs way more testing.

### DIFF
--- a/integration/conf_throttling_test.go
+++ b/integration/conf_throttling_test.go
@@ -101,7 +101,7 @@ func (s *ThrottlingSuite) TestThrottleConfReload() {
 	// The test tries to trigger a config reload with the REST API every 200ms,
 	// 10 times (so for 2s in total).
 	// Therefore the throttling (set at 400ms for this test) should only let
-	// (2s / 400 ms =) 5 config reloads happen in theory.
-	// In addition, we have to take into account the extra config reload from the internal provider (5 + 1).
-	assert.LessOrEqual(s.T(), reloads, 6)
+	// (1st reload + 2s / 400 ms =) 6 config reloads happen in theory.
+	// In addition, we have to take into account the extra config reload from the internal provider (6 + 1).
+	assert.LessOrEqual(s.T(), reloads, 7)
 }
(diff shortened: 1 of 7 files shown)


nmengin:
Hello @ddtmachado,

Thank you for this PR, we are very interested in it but won't be able to test/review/merge it before releasing Traefik v3.0.
That's why I moved it to the `next` milestone.

On your side, could you rebase it on the `master` branch?

rtribotte (changes requested):
Hey @ddtmachado!

Nice to see you around and thanks for this contribution!
I took a look at the changes and I have a few comments.
Regarding the next steps of the review process, do you want to work and iterate on the changes, or you don't mind if we push some review commits?

rtribotte:
Hello @ddtmachado,

I pushed review comments and also rebased the changes.
I have marked the PR as need-review, I think I could LGTM it right away, but I prefer to sleep on it to give it maybe more thought :-)

rtribotte (approved):
Thanks!

juliens:
Hello @ddtmachado

Thank you for your PR.

After discussing with other maintainers, I’m closing this PR for now as its current implementation has unintended impacts on the management of certificate. Currently, certificates are switched atomically, in sync with the entire dynamic configuration. However, the changes in this PR lead to incremental updates, which could conflict with the rest of the dynamic configuration.

That said, for v4, we plan to explore the possibility of improving the global atomic switch to make it more granular. This effort will also include considerations for memory optimization, particularly around TLS management.

Thank you for your understanding, and I look forward to revisiting these ideas in the future.

PR #10323, opened 2024-01-01
Native Kubernetes service load-balancing at the provider level

Prajithp:
### What does this PR do?
This pull request introduces a new provider option to use the native Kubernetes load balancer instead of the one provided by Traefik at the global level. Currently, the same can be achieved at the IngressRoute level by setting nativeLB, but it requires modifying the existing IngressRoute, and there's no option to use the feature globally.

### Motivation
When using service meshes to manage cluster traffic from the Ingress to the pods, it is necessary to forward the traffic to meshed Kubernetes services. When using nativeLB, we are currently required to set this at the IngressRoute level, with no option available to configure it at the global level.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes.go
+++ b/pkg/provider/kubernetes/crd/kubernetes.go
@@ -58,8 +58,8 @@ type Provider struct {
 	IngressClass              string          `description:"Value of kubernetes.io/ingress.class annotation to watch for." json:"ingressClass,omitempty" toml:"ingressClass,omitempty" yaml:"ingressClass,omitempty" export:"true"`
 	ThrottleDuration          ptypes.Duration `description:"Ingress refresh throttle duration" json:"throttleDuration,omitempty" toml:"throttleDuration,omitempty" yaml:"throttleDuration,omitempty" export:"true"`
 	AllowEmptyServices        bool            `description:"Allow the creation of services without endpoints." json:"allowEmptyServices,omitempty" toml:"allowEmptyServices,omitempty" yaml:"allowEmptyServices,omitempty" export:"true"`
-
-	lastConfiguration safe.Safe
+	UseNativeLoadBalancer     bool            `description:"Use native Kubernetes Loadbalancing instead of traefik provided" json:"useNativeLoadBalancer,omitempty" toml:"useNativeLoadBalancer,omitempty" yaml:"useNativeLoadBalancer,omitempty" export:"true"`
+	lastConfiguration         safe.Safe
 
 	routerTransform k8s.RouterTransform
 }
--- a/pkg/provider/kubernetes/crd/kubernetes_http.go
+++ b/pkg/provider/kubernetes/crd/kubernetes_http.go
@@ -55,6 +55,7 @@ func (p *Provider) loadIngressRouteConfiguration(ctx context.Context, client Cli
 			allowCrossNamespace:       p.AllowCrossNamespace,
 			allowExternalNameServices: p.AllowExternalNameServices,
 			allowEmptyServices:        p.AllowEmptyServices,
+			useNativeLoadBalancer:     p.UseNativeLoadBalancer,
 		}
 
 		for _, route := range ingressRoute.Spec.Routes {
@@ -201,6 +202,7 @@ type configBuilder struct {
 	allowCrossNamespace       bool
 	allowExternalNameServices bool
 	allowEmptyServices        bool
+	useNativeLoadBalancer     bool
 }
 
 // buildTraefikService creates the configuration for the traefik service defined in tService,
@@ -376,20 +378,6 @@ func (c configBuilder) loadServers(parentNamespace string, svc traefikv1alpha1.L
 		return nil, err
 	}
 
-	if svc.NativeLB {
-		address, err := getNativeServiceAddress(*service, *svcPort)
-		if err != nil {
-			return nil, fmt.Errorf("getting native Kubernetes Service address: %w", err)
-		}
-
-		protocol, err := parseServiceProtocol(svc.Scheme, svcPort.Name, svcPort.Port)
-		if err != nil {
-			return nil, err
-		}
-
-		return []dynamic.Server{{URL: fmt.Sprintf("%s://%s", protocol, address)}}, nil
-	}
-
 	var servers []dynamic.Server
 	if service.Spec.Type == corev1.ServiceTypeExternalName {
 		if !c.allowExternalNameServices {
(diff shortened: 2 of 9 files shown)


nmengin:
Hello @Prajithp,

Thanks for your contribution.
After discussing it with other maintainers, we'd prefer you open an issue to discuss your use case before moving forward on your PR review.

Indeed we are unsure about the use case and we'd like to have the community's opinion on it.

Thanks in advance.

Prajithp:
Can this be considered? @nmengin

nmengin:
Hello @prajithp13,

I want to wait a bit more to check if we have more feedback on the PR.
After that, I will share the proposition with the other maintainers to get their opinion.

I'll make sure to update the proposal and the PR accordingly.

pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -58,8 +58,8 @@ type Provider struct {
 	IngressClass              string          `description:"Value of kubernetes.io/ingress.class annotation to watch for." json:"ingressClass,omitempty" toml:"ingressClass,omitempty" yaml:"ingressClass,omitempty" export:"true"`
 	ThrottleDuration          ptypes.Duration `description:"Ingress refresh throttle duration" json:"throttleDuration,omitempty" toml:"throttleDuration,omitempty" yaml:"throttleDuration,omitempty" export:"true"`
 	AllowEmptyServices        bool            `description:"Allow the creation of services without endpoints." json:"allowEmptyServices,omitempty" toml:"allowEmptyServices,omitempty" yaml:"allowEmptyServices,omitempty" export:"true"`
-
-	lastConfiguration safe.Safe
+	UseNativeLoadBalancer     bool            `description:"Use native Kubernetes Loadbalancing instead of traefik provided" json:"useNativeLoadBalancer,omitempty" toml:"useNativeLoadBalancer,omitempty" yaml:"useNativeLoadBalancer,omitempty" export:"true"`
```
nmengin, line 61: WDYT to rename the option to `NativeLBByDefault`?

pkg/provider/kubernetes/crd/kubernetes_tcp.go
```diff
@@ -232,21 +232,22 @@ func (p *Provider) loadTCPServers(client Client, namespace string, svc traefikv1
...
+		// External services will not have cluster IPs, so this should be handled after externalName checks.
+		if svc.NativeLB || p.UseNativeLoadBalancer {
+			address, err := getNativeServiceAddress(*service, *svcPort)
+			if err != nil {
+				return nil, fmt.Errorf("getting native Kubernetes Service address: %w", err)
+			}
+
+			return []dynamic.TCPServer{{Address: address}}, nil
+		}
+
```
nmengin, line 250: Same suggestion as above.

pkg/provider/kubernetes/crd/kubernetes_udp.go
```diff
@@ -121,21 +121,22 @@ func (p *Provider) loadUDPServers(client Client, namespace string, svc traefikv1
...
+		// External services will not have cluster IPs, so this should be handled after externalName checks.
+		if svc.NativeLB || p.UseNativeLoadBalancer {
+			address, err := getNativeServiceAddress(*service, *svcPort)
+			if err != nil {
+				return nil, fmt.Errorf("getting native Kubernetes Service address: %w", err)
+			}
+
+			return []dynamic.UDPServer{{Address: address}}, nil
+		}
+
```
nmengin, line 139: Same as above

nmengin (changes requested):
Hello @Prajithp,

Thank you for your PR.
My suggestions after a first review.

pkg/provider/kubernetes/crd/kubernetes.go
Prajithp: I have renamed this to `UseNativeLB`. I believe this makes more sense, but please let me know if you have any preferences, and I can make the change accordingly.

pkg/provider/kubernetes/crd/kubernetes_http.go
```diff
@@ -408,6 +396,21 @@ func (c configBuilder) loadServers(parentNamespace string, svc traefikv1alpha1.L
...
+		}
+
+		protocol, err := parseServiceProtocol(svc.Scheme, svcPort.Name, svcPort.Port)
+		if err != nil {
+			return nil, err
+		}
+
+		return []dynamic.Server{{URL: fmt.Sprintf("%s://%s", protocol, address)}}, nil
+	}
+
```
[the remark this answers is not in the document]
Prajithp: > Could you tell me why you have moved this statement after the other one?

An ExternalName Service does not have the Service's 'clusterIP' and 'port.'  Therefore, if useNativeLB is set at the provider level or NativeLB is set at the ingressroute level, it will result in a 502 bad gateway error. This needs to be handled after checking for ServiceTypeExternalName.

> To do so, you can replace the current NativeLB type (bool) with a *bool to distinguish when a user sets false or sets no value.

Done

Prajithp:
@nmengin can you please review this?

Prajithp:
@mmatur Have already added a comment for that.

pkg/provider/kubernetes/crd/kubernetes.go
nmengin: With other maintainers, we have thought that NativeLBByDefault was a good enough name because it's homogeneous with other existing options that allow the same kind of behavior (general option in the static configuration that can be overridden in the dynamic configuration), like [connectByDefault](https://doc.traefik.io/traefik/providers/consul-catalog/#connectbydefault) in Consul Catalog.

nmengin:
Hey @Prajithp,

It looks like my answer to your comment was not applied 😅 
I re-answered

pkg/provider/kubernetes/crd/kubernetes.go
Prajithp: Updated.

Prajithp:
@nmengin have updated the PR.

nmengin:
Hello @Prajithp,

Thank you for the modification 👍 

After a deeper look, I've seen you've just modified the Kubernetes CRD provider.
Could you apply them to both the [Ingress provider](https://github.com/traefik/traefik/blob/master/pkg/provider/kubernetes/ingress/kubernetes.go) too?

Once the modifications are done, could you describe the new option in their documentation too ([CRD](https://doc.traefik.io/traefik/providers/kubernetes-crd/), [Ingress](https://doc.traefik.io/traefik/providers/kubernetes-ingress/))?

Of course, if you need any help or assistance to go further, just let me know and we'll do our best to help you move forward.

Prajithp:
@nmengin Done. Please check.

rtribotte (approved):
Thanks!

nmengin (approved):
Hey @Prajithp,

Thanks a lot for this useful feature 👏 🚀 
LGTM

PR #10345, opened 2024-01-11
update kubernetes example in middleware/http

hevinhsu:
### What does this PR do?

I updated the apiVersion of the kubernetes example in the middleware/http section to the current version.

### Motivation

To have a usable example

### More

- [ ] Added/updated tests

### Additional Notes

this PR replace all apiVersion in middleware/http doc from ```traefik.io/v1alpha1``` to ```traefik.containo.us/v1alpha1``` in kubernetes example


nmengin:
Hello @hevinhsu,

Thanks for your contribution.
I have to close it because the modification you propose is wrong: we have introduced the APIGroup `traefik.io` in v3 that's why this APIGroup is in the documentation instead of the group `traefik.containo.us`.

PR #10361, opened 2024-01-19
fix: tailscale is required for Docker Desktop users

mmatur:
### What does this PR do?

Tailscale is required ofr Docker Desktop users

cf: https://doc.traefik.io/traefik/v2.11/contributing/building-testing/#testing

### Motivation

Block Docker Desktop users to run integration tests if `tailscale.secret` is not present


rtribotte (approved):
LGTM

PR #10367, opened 2024-01-22
misc(Makefile): add `help` target to display the help msg

Halimao:
### What does this PR do?

This pr added a new target `help` to project `Makefile` for displaying which commonly used commands that the project `Makefile` supported with their descriptions.

image

### Motivation

This is useful when a new developer goes to find a helpful tool for testing or building etc.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,19 +22,22 @@ LINT_EXECUTABLES = misspell shellcheck
 DOCKER_BUILD_PLATFORMS ?= linux/amd64,linux/arm64
 
 .PHONY: default
+#? default: Run `make generate` and `make binary`
 default: generate binary
 
-## Create the "dist" directory
+#? dist: Create the "dist" directory
 dist:
 	mkdir -p dist
 
 ## Build WebUI Docker image
 .PHONY: build-webui-image
+#? build-webui-image: Build WebUI Docker image
 build-webui-image:
 	docker build -t traefik-webui -f webui/Dockerfile webui
 
 ## Clean WebUI static generated assets
 .PHONY: clean-webui
+#? clean-webui: Clean WebUI static generated assets
 clean-webui:
 	rm -r webui/static
 	mkdir -p webui/static
@@ -47,15 +50,18 @@ webui/static/index.html:
 	docker run --rm -v "$(PWD)/webui/static":'/src/webui/static' traefik-webui chown -R $(shell id -u):$(shell id -g) ./static
 
 .PHONY: generate-webui
+#? generate-webui: Generate WebUI
 generate-webui: webui/static/index.html
 
 ## Generate code
 .PHONY: generate
+#? generate: Generate code
 generate:
 	go generate
 
 ## Build the binary
 .PHONY: binary
+#? binary: Build the binary
 binary: generate-webui dist
 	@echo SHA: $(VERSION) $(CODENAME) $(DATE)
 	CGO_ENABLED=0 GOGC=off GOOS=${GOOS} GOARCH=${GOARCH} go build ${FLAGS[*]} -ldflags "-s -w \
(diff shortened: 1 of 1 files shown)


mmatur:
Hi @Halimao ,

Thanks for your PR, could you please rebase your PR on v2.11 branch.

Halimao:
> Hi @Halimao ,
> 
> Thanks for your PR, could you please rebase your PR on v2.11 branch.

Is it ok now?

Makefile
```diff
@@ -47,15 +50,18 @@ webui/static/index.html:
 	docker run --rm -v "$(PWD)/webui/static":'/src/webui/static' traefik-webui chown -R $(shell id -u):$(shell id -g) ./static
 
 .PHONY: generate-webui
+#? generate-webui: Generate WebUI
 generate-webui: webui/static/index.html
 
 ## Generate code
 .PHONY: generate
+#? generate: Generate code
```
rtribotte, line 58:
```suggestion
#? generate: Generate code (Dynamic and Static configuration documentation reference files)
```

Makefile
```diff
@@ -138,49 +152,65 @@ multi-arch-image-%: binary-linux-amd64 binary-linux-arm64
...
 ## Format the Code
 .PHONY: fmt
+#? fmt: Format the Code
 fmt:
 	gofmt -s -l -w $(SRCS)
+
+#? help: Get more info on make commands
+help: Makefile
+	@echo " Choose a command run in traefik:"
+	@sed -n 's/^#?//p' $< | column -t -s ':' |  sort | sed -e 's/^/ /'
```
mmatur, line 214:
```suggestion
	@sed -n 's/^##//p' $< | column -t -s ':' |  sort | sed -e 's/^/ /'
```

Why not using the comment already in place?

mmatur (changes requested):
Thanks for your contribution @Halimao 

Few comments

Makefile
```diff
@@ -138,49 +152,65 @@ multi-arch-image-%: binary-linux-amd64 binary-linux-arm64
...
 
 ## Generate code from dynamic configuration https://github.com/traefik/genconf
 .PHONY: generate-genconf
+#? generate-genconf: Generate code from dynamic configuration github.com/traefik/genconf
 generate-genconf:
 	go run ./cmd/internal/gen/
 
 ## Create packages for the release
 .PHONY: release-packages
+#? release-packages: Format the Code
```
mmatur, line 201:
```suggestion
#? release-packages: Create packages for the release
```

mmatur (changes requested):

Makefile
Halimao: In my opinion, `#?` is more particular than `##`, and won't be effected by `###+`

mmatur: What is the meaning of `###+` in a Makefile?

As I can read [here](https://www.gnu.org/software/make/manual/make.html#Makefile-Contents):

> ‘#’ in a line of a makefile starts a comment. It and the rest of the line are ignored, except that a trailing backslash not escaped by another backslash will continue the comment across multiple lines. A line containing just a comment (with perhaps spaces before it) is effectively blank, and is ignored. If you want a literal #, escape it with a backslash (e.g., \#). Comments may appear on any line in the makefile, although they are treated specially in certain situations.

To me the maintain of the Makefile will be more difficult if we introduce a new format just for the help target

Halimao: Sorry about this. 

I used `###+` as a regex pattern, which means that there may be more than two `#` in the Makefile.

The reason why I use `#?` is to avoid someone using `###` for a comment or something to describe a target group (see the following screenshot), or something else.
image

mmatur: No worries, thanks for the explanation.

To me we should get rid of `##` comment and only kept `#?`

mmatur (approved):
LGTM

juliens (approved):
LGTM

ldez (approved):

PR #10375, opened 2024-01-22
Nomad provider to allow empty services

chrispruitt:
### What does this PR do?

This enables the nomad provider the ability to allow empty services. For instance, when a service is scaled down to 0 the route will persist. This would be an opt-in feature (false by default).
Fixes https://github.com/traefik/traefik/issues/10374

### Motivation

I currently have services in Nomad that scaled down to zero and use a port-knocking solution that scales the service up when the route receives traffic. Also, the Kubernetes provider already has [this option](https://github.com/traefik/traefik/pull/8802), so I felt like it was missing from the nomad provider. 

### More

- [x] Added/updated tests

### Additional Notes

I attempted to implement this just like the Kubernetes provider for consistency.
Also, to make this possible, I had to utilize Nomad's job API methods, rather than the service API methods to retrieve the list of existing services. This is because Nomad [does not currently list nonregistered services](https://github.com/hashicorp/nomad/issues/19731#issuecomment-1899046059) (or services scaled to 0) with service API methods.


pkg/provider/nomad/nomad_test.go
```diff
@@ -161,70 +300,144 @@ func Test_getNomadServiceData(t *testing.T) {
...
+	// make the query for services
+	items, err := p.getNomadServiceData(context.TODO())
+	require.NoError(t, err)
+
+	// Should not be listed as job is stopped
+	require.Empty(t, items)
+}
+
+func setup() error {
+	responsesDir := "./nomad_responses"
```
mmatur, line 428:
```suggestion
	responsesDir := "./fixtures"
```

Could you please rename `nomad_responses` directory to `fixtures`?

pkg/provider/nomad/nomad.go
```diff
@@ -235,55 +237,90 @@ func (p *Provider) loadConfiguration(ctx context.Context, configurationC chan<-
 }
 
 func (p *Provider) getNomadServiceData(ctx context.Context) ([]item, error) {
-	// first, get list of service stubs
-	opts := &api.QueryOptions{AllowStale: p.Stale}
-	opts = opts.WithContext(ctx)
+	jobsOpts := &api.QueryOptions{}
```
mmatur, line 240:
```suggestion
	jobsOpts := &api.QueryOptions{AllowStale: p.Stale}
```

mmatur: WDYT of keeping the `getNomadServiceData` method at it was and creating a new method `getNomadServiceDataWithEmptyServices` and updating the `loadConfiguration` method liake this:

```go
func (p *Provider) loadConfiguration(ctx context.Context, configurationC chan<- dynamic.Message) error {
	var items []item
	var err error
	if p.AllowEmptyServices {
		items, err = p.getNomadServiceDataWithEmptyServices(ctx)
		if err != nil {
			return err
		}
	} else {
		items, err = p.getNomadServiceData(ctx)
		if err != nil {
			return err
		}
	}
	
	configurationC <- dynamic.Message{
		ProviderName:  p.name,
		Configuration: p.buildConfig(ctx, items),
	}

	return nil
}
```
It will avoid to current users that don't need empty services to be affected by the change of fetching Jobs instead of Services.
For large cluster it will impact memory consumption and if we fetching jobs instead of services.

mmatur (changes requested):
Thanks a lot for this contribution @chrispruitt 

Few comments below

pkg/provider/nomad/nomad.go
chrispruitt: I agree, that would be a better approach. I'll make the suggested changes. Thanks!

pkg/provider/nomad/nomad_test.go
chrispruitt: changes have been made

pkg/provider/nomad/nomad.go
chrispruitt: I implemented the suggested changes. Thanks!

mmatur:
Hi @chrispruitt 

Thanks for your changes. Seems good to me regarding the feature.

I am one more question what do you think to use a mock instead of creating these fixtures files for test?

chrispruitt:
@mmatur I just followed the same process this provider package already had in place. The only change I made was moving the JSON responses to files. I don't have a strong opinion either way. They do end up with the same result in my opinion and the fixtures approach is easy to track mocked responses IMO.

On another note, do I need to rebase this branch? Or would you do that before merging? Thanks!

mmatur (approved):
Thanks for your feedback @chrispruitt .

Let's move forward with that and see what other maintainers think about it

I gave it a try and it seems good to me 

LGTM

nmengin (approved):
Thank you @chrispruitt for this contribution!

LGTM 👏

chrispruitt:
Are we able to get one more approval and merge this guy???

Thanks!

rtribotte (approved):
Thanks 👌

PR #10387, opened 2024-01-25
chore: Upgrade UI development dependencies

### LINKED ISSUE #9932: Upgrade UI development dependencies

andsarr:
### What does this PR do?

This PR upgrades the UI dependencies and remove some of the deprecated dependencies. It also upgrades the base Docker image to Node JS 20.11

### Motivation

Fixes #9932

### More

- [ ] Added/updated tests


andsarr:
I just realized that the project is using Vue 2.x which has reached EOL. I'll take the opportunity to upgrade to Vue 3.x, but this may take a while.

andsarr:
Hi @mmatur I totally agree with you. It is better to do this in stages instead of all at once. I'll open another PR for just upgrading Node.js!

andsarr:
Closing this in favor of [10416](https://github.com/traefik/traefik/pull/10416)

PR #10388, opened 2024-01-25
Improvement: ConsulCatalog StrictChecks

djenriquez:
### What does this PR do?

This PR allows users to define a `StrictChecks` config for the `consulcatalog` provider. `StrictChecks` lets a user define which healthcheck statuses should be allowed when determining which upstreams to include in the upstream targets.

`StrictChecks` defaults to `"passing", "warning"` to preserve the existing behavior, allowing this config to be backwards compatible.

### Motivation

Our organization relies on "warning" health statuses to determine upstreams that are "starting up", which is a different state than "critical", but not yet ready for traffic. The existing build sends traffic immediately to "warning" upstreams, which causes 5XX errors for our use case. 

With this config, users can choose what health status to allow based on the available statuses that Consul supports as their organization requires.

### More

- [X] Added/updated tests

### Additional Notes

Question: will I need to add additional logic to properly handle the different configuration input methods or will the configuration handler automatically unmarshal the values into the provider?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ plugins-local/
 traefik_changelog.md
 integration/tailscale.secret
 integration/conformance-reports/
+go.work*
--- a/pkg/provider/consulcatalog/config.go
+++ b/pkg/provider/consulcatalog/config.go
@@ -9,7 +9,6 @@ import (
 	"sort"
 	"strings"
 
-	"github.com/hashicorp/consul/api"
 	"github.com/rs/zerolog/log"
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
 	"github.com/traefik/traefik/v3/pkg/config/label"
@@ -132,7 +131,7 @@ func (p *Provider) keepContainer(ctx context.Context, item itemData) bool {
 		return false
 	}
 
-	if item.Status != api.HealthPassing && item.Status != api.HealthWarning {
+	if !p.includesHealthStatus(item.Status) {
 		logger.Debug().Msg("Filtering unhealthy or starting item")
 		return false
 	}
--- a/pkg/provider/consulcatalog/config_test.go
+++ b/pkg/provider/consulcatalog/config_test.go
@@ -3206,3 +3206,448 @@ func extractNSFromProvider(providers []*Provider) []string {
 	}
 	return res
 }
+
+func TestFilterHealthStatuses(t *testing.T) {
+	testCases := []struct {
+		desc         string
+		items        []itemData
+		defaultRule  string
+		expected     *dynamic.Configuration
+		strictChecks []string
+	}{
+		{
+			// No value passed in here, we assume the default of ["passing"]
+			desc: "test default strict checks",
+			items: []itemData{
+				{
+					ID:      "id",
+					Node:    "Node1",
+					Name:    "Test1",
+					Address: "127.0.0.1",
+					Port:    "80",
+					Labels:  nil,
+					Status:  api.HealthPassing,
+				},
+				{
+					ID:      "id",
+					Node:    "Node2",
+					Name:    "Test2",
+					Address: "127.0.0.1",
+					Port:    "81",
+					Labels:  nil,
+					Status:  api.HealthWarning,
+				},
+			},
+			defaultRule: "Host(`foo.bar`)",
+			expected: &dynamic.Configuration{
+				TCP: &dynamic.TCPConfiguration{
+					Routers:           map[string]*dynamic.TCPRouter{},
+					Middlewares:       map[string]*dynamic.TCPMiddleware{},
+					Services:          map[string]*dynamic.TCPService{},
+					ServersTransports: map[string]*dynamic.TCPServersTransport{},
+				},
+				UDP: &dynamic.UDPConfiguration{
+					Routers:  map[string]*dynamic.UDPRouter{},
+					Services: map[string]*dynamic.UDPService{},
+				},
+				HTTP: &dynamic.HTTPConfiguration{
+					Routers: map[string]*dynamic.Router{
+						"Test1": {
+							Service:     "Test1",
+							Rule:        "Host(`foo.bar`)",
+							DefaultRule: true,
+						},
+						"Test2": {
+							Service:     "Test2",
+							Rule:        "Host(`foo.bar`)",
+							DefaultRule: true,
+						},
+					},
+					Middlewares: map[string]*dynamic.Middleware{},
+					Services: map[string]*dynamic.Service{
+						"Test1": {
+							LoadBalancer: &dynamic.ServersLoadBalancer{
+								Servers: []dynamic.Server{
+									{
+										URL: "http://127.0.0.1:80",
+									},
+								},
+								PassHostHeader: Bool(true),
+								ResponseForwarding: &dynamic.ResponseForwarding{
+									FlushInterval: ptypes.Duration(100 * time.Millisecond),
+								},
+							},
+						},
+						"Test2": {
+							LoadBalancer: &dynamic.ServersLoadBalancer{
+								Servers: []dynamic.Server{
+									{
+										URL: "http://127.0.0.1:81",
+									},
+								},
+								PassHostHeader: Bool(true),
+								ResponseForwarding: &dynamic.ResponseForwarding{
+									FlushInterval: ptypes.Duration(100 * time.Millisecond),
+								},
+							},
+						},
+					},
+					ServersTransports: map[string]*dynamic.ServersTransport{},
+				},
+			},
+		},
+		{
+			// The item's health status is not included in the default checks, do not expect any containers
+			desc: "test status not included",
+			items: []itemData{
+				{
+					ID:      "id",
+					Node:    "Node1",
+					Name:    "Test",
+					Address: "127.0.0.1",
+					Port:    "80",
+					Labels:  nil,
+					Status:  api.HealthCritical,
+				},
+			},
+			defaultRule: "Host(`foo.bar`)",
+			expected: &dynamic.Configuration{
+				TCP: &dynamic.TCPConfiguration{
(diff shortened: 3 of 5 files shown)


djenriquez:
Reference Issue: https://github.com/traefik/traefik/issues/10074

pkg/provider/consulcatalog/config.go
```diff
@@ -132,8 +131,8 @@ func (p *Provider) keepContainer(ctx context.Context, item itemData) bool {
 		return false
 	}
 
-	if item.Status != api.HealthPassing && item.Status != api.HealthWarning {
-		logger.Debug().Msg("Filtering unhealthy or starting item")
+	if !p.includesHealthStatus(item.Status) {
+		logger.Debug().Msg(fmt.Sprintf("Status \"%s\" is not included in the configured strictChecks of \"%s\"", item.Status, strings.Join(p.StrictChecks, ",")))
```
jspdown, line 135:
```suggestion
		logger.Debug().Msgf("Status %q is not included in the configured strictChecks of %q", item.Status, strings.Join(p.StrictChecks, ", "))
```

pkg/provider/consulcatalog/config_test.go
```diff
@@ -3206,3 +3206,448 @@ func extractNSFromProvider(providers []*Provider) []string {
...
 	return res
 }
+
+func TestFilterHealthStatuses(t *testing.T) {
+	testCases := []struct {
+		desc         string
+		items        []itemData
+		defaultRule  string
+		expected     *dynamic.Configuration
+		strictChecks []string
```
jspdown, line 3216: nit: could you move `strictChecks` bellow `items` so we can more easily see the configuration of the test?

pkg/provider/consulcatalog/config_test.go
```diff
@@ -3206,3 +3206,448 @@ func extractNSFromProvider(providers []*Provider) []string {
...
+	}{
+		{
+			// No value passed in here, we assume the default of ["passing"]
```
jspdown, line 3219:
```suggestion
			// No value passed in here, we assume the default of ["passing", "warning"]
```

pkg/provider/consulcatalog/consul_catalog.go
```diff
@@ -136,6 +139,11 @@ func (p *Provider) Init() error {
 		return fmt.Errorf("error while parsing default rule: %w", err)
 	}
 
+	// Set the default health check to keep container to "passing"
+	if len(p.StrictChecks) == 0 {
+		p.StrictChecks = []string{api.HealthPassing, api.HealthWarning}
```
jspdown, line 144: You got it right in `SetDefaults`, but I don't think we should set the default values here. 
If `StrictChecks` is empty at this stage, it's deliberately the choice of the user. I don't see why someone would want to do it, but from a user perspective, the behavior is easier to understand.

pkg/provider/consulcatalog/config_test.go
```diff
@@ -3206,3 +3206,448 @@ func extractNSFromProvider(providers []*Provider) []string {
 	}
 	return res
 }
+
+func TestFilterHealthStatuses(t *testing.T) {
+	testCases := []struct {
+		desc         string
+		items        []itemData
+		defaultRule  string
```
jspdown, line 3214: WDYT about removing the `defaultRule` field from this test? Since we are not going to change it for testing health statuses based filtering, I think we could simply hardcode it in the config.

jspdown (changes requested):
Thanks a lot for opening this PR :smiley: 

I have some feedback, nothing critical.

> will I need to add additional logic to properly handle the different configuration input methods or will the configuration handler automatically unmarshal the values into the provider?

The configuration will be automatically unmarshaled,  there's nothing else to add in the code. But for updating the reference part of the documentation you will have to run `make generate`.

pkg/provider/consulcatalog/consul_catalog.go
djenriquez: I wasn't seeing `SetDefaults` called anywhere, and noticed that tests were failing, which told me the defaults weren't actually being set, so, I added it here. Are we missing a call to `SetDefaults` somewhere?

djenriquez: Oh i see, defaults arent set during tests, but are set in the main flow, so just need to make sure we include that in the tests.

djenriquez:
The only thing I'm unsure of is whether the default values are applied when a user does not define `strictChecks` in their config, I might be missing a mapping somewhere for this new property. This will be critical to backwards compat, any ideas of the best way to test that or verify that it should work?

pkg/provider/consulcatalog/consul_catalog.go
jspdown: Yes indeed, `SetDefaults` is not explicitly called in the code. When building the static configuration we recursively call this method via reflect. For your test, it would be best to call this method manually on the test config.

pkg/provider/consulcatalog/config_test.go
```diff
@@ -3206,3 +3208,444 @@ func extractNSFromProvider(providers []*Provider) []string {
...
+	for _, test := range testCases {
+		test := test
+		t.Run(test.desc, func(t *testing.T) {
+			t.Parallel()
+
+			p := Provider{
+				Configuration: Configuration{
+					ExposedByDefault: true,
+					DefaultRule:      "Host(`foo.bar`)",
+					StrictChecks:     test.strictChecks,
```
jspdown, line 3633: We might want to change the initialization of the provider a little bit to prevent diverging in the tests from the `SetDefaults` content. WDYT about constructing it that way:

```go
var config Configuration

config.SetDefaults()
config.DefaultTule = "Host(`foo.bar`)"

if test.strictChecks != nil {
  config.StrictChecks = test.strictChecks
}
``` 

This would let us test the default `strictChecks` without specifying them in the tests. Which also permit us to hard code those in the `SetDefaults` method.

jspdown (commented):
One last feedback

pkg/provider/consulcatalog/config_test.go
djenriquez: I think that makes perfect sense, thank you for that guidance. Will make the change now.

djenriquez:
Hi all, checking in if there are other items or actions needed? Thanks!

jspdown:
@djenriquez No, sorry for the delay. I couldn't allocate time to review your changes recently. I will have a look at it later today.

pkg/provider/consulcatalog/consul_catalog.go
```diff
@@ -88,6 +89,7 @@ type Configuration struct {
 	ConnectByDefault  bool            `description:"Consider every service as Connect capable by default." json:"connectByDefault,omitempty" toml:"connectByDefault,omitempty" yaml:"connectByDefault,omitempty" export:"true"`
 	ServiceName       string          `description:"Name of the Traefik service in Consul Catalog (needs to be registered via the orchestrator or manually)." json:"serviceName,omitempty" toml:"serviceName,omitempty" yaml:"serviceName,omitempty" export:"true"`
 	Watch             bool            `description:"Watch Consul API events." json:"watch,omitempty" toml:"watch,omitempty" yaml:"watch,omitempty" export:"true"`
+	StrictChecks      []string        `description:"A list of service health statuses to allow taking traffic. Defaults to [\"passing\", \"warning\"]." json:"strictChecks,omitempty" toml:"strictChecks,omitempty" yaml:"strictChecks,omitempty" export:"true"`
```
youkoulayley, line 92:
```suggestion
	StrictChecks      []string        `description:"A list of service health statuses to allow taking traffic." json:"strictChecks,omitempty" toml:"strictChecks,omitempty" yaml:"strictChecks,omitempty" export:"true"`
```

The defaults are automatically generated with the `make generate` command. Like you can see in the `cli-ref.md` for example: 
```
A list of service health statuses to allow taking traffic. Defaults to ["passing", "warning"]. (Default: ```passing, warning```)
```

So I think you could remove the one you wrote and regenerate the documentation so that it is only written once.

.gitignore
```diff
@@ -20,3 +20,4 @@ plugins-local/
 traefik_changelog.md
 integration/tailscale.secret
 integration/conformance-reports/
+go.work*
```
youkoulayley, line 23:
```suggestion
```

pkg/provider/consulcatalog/config.go
```diff
@@ -132,8 +132,8 @@ func (p *Provider) keepContainer(ctx context.Context, item itemData) bool {
 		return false
 	}
 
-	if item.Status != api.HealthPassing && item.Status != api.HealthWarning {
-		logger.Debug().Msg("Filtering unhealthy or starting item")
+	if !p.includesHealthStatus(item.Status) {
+		logger.Debug().Msgf("Status \"%s\" is not included in the configured strictChecks of \"%s\"", item.Status, strings.Join(p.StrictChecks, ","))
```
youkoulayley, line 136:
```suggestion
		logger.Debug().Msgf("Status %q is not included in the configured strictChecks of %q", item.Status, strings.Join(p.StrictChecks, ","))
```

youkoulayley (commented):
Tested and works perfectly, just a few minor comments

djenriquez:
> Tested and works perfectly, just a few minor comments

Thanks for the review! Feedback applied 👍 .

nmengin (approved):
Thank you for your contribution 👍 

LGTM

mmatur (approved):
LGTM

jspdown (approved):
Thanks a lot for this PR, LGTM :ship:

mmatur:
@djenriquez could you please rebase your PR.

Our bot is not able to do it because the PR has been created from an GitHub Organisation fork.

djenriquez:
> @djenriquez could you please rebase your PR.
> 
> Our bot is not able to do it because the PR has been created from an GitHub Organisation fork.

Definitely, sorry for the delay. Thanks!

PR #10398, opened 2024-01-31
Fix the keepAlive options for the CLI examples

immanuelfodor:
### What does this PR do?

The KeepAliveMaxTime and KeepAliveMaxRequests options were introduced in https://github.com/traefik/traefik/pull/10247 but there is an error in the provided examples.

Although Max is used everywhere, the CLI options are without Max:

In the CLI reference, it's clear that both should use the word max:

### Motivation

Fix a configuration example with the proper static config key.

### More

- [ ] Added/updated tests


nmengin (approved):
LGTM 📖 👍

mmatur (approved):

PR #10405, opened 2024-02-02
Fix NTLM and Kerberos

juliens:
### What does this PR do?

IIS needs to have the same connection for the all negotiating process for kerberos and NTLM authentication. So we need to "stick" to the TCP connection with the backend for a single client TCP Connection, otherwise, authentication fail.

This PR fixes this issues with kerberos ad NTLM authentication.

### Motivation

Fix failing kerberos and NTLM authentication

### More

- [ ] Added/updated tests


nmengin (approved):
LGTM

lbenguigui (approved):
LGTM

ldez (approved):

PR #10408, opened 2024-02-03
Add http.real_ip span attribute for entrypoint

prutheus:
### What does this PR do?

As requested in https://github.com/traefik/traefik/issues/7123, this PR adds a attribute to the entry point tracing span for the client IP if present from headers.

### Motivation

https://github.com/traefik/traefik/issues/7123
Be able to trace specific user IP requests.

### More

- [X] Added/updated tests

### DIFF
--- a/pkg/middlewares/tracing/entrypoint.go
+++ b/pkg/middlewares/tracing/entrypoint.go
@@ -44,6 +44,13 @@ func (e *entryPointTracing) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	tracingCtx, span := e.tracer.Start(tracingCtx, "EntryPoint", trace.WithSpanKind(trace.SpanKindServer))
 	defer span.End()
 
+	if req.Header.Get("X-Real-Ip") != "" {
+		span.SetAttributes(attribute.String("http.real_ip", req.Header.Get("X-Real-Ip")))
+	}
+	if req.Header.Get("X-Forwarded-For") != "" {
+		span.SetAttributes(attribute.String("http.real_ip", req.Header.Get("X-Forwarded-For")))
+	}
+
 	req = req.WithContext(tracingCtx)
 
 	span.SetAttributes(attribute.String("entry_point", e.entryPoint))
--- a/pkg/middlewares/tracing/entrypoint_test.go
+++ b/pkg/middlewares/tracing/entrypoint_test.go
@@ -27,6 +27,7 @@ func TestEntryPointMiddleware(t *testing.T) {
 			expected: expected{
 				name: "EntryPoint",
 				attributes: []attribute.KeyValue{
+					attribute.String("http.real_ip", "123.456.789.123"),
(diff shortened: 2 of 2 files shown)


pkg/middlewares/tracing/entrypoint.go
```diff
@@ -44,6 +44,13 @@ func (e *entryPointTracing) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	tracingCtx, span := e.tracer.Start(tracingCtx, "EntryPoint", trace.WithSpanKind(trace.SpanKindServer))
 	defer span.End()
 
+	if req.Header.Get("X-Real-Ip") != "" {
+		span.SetAttributes(attribute.String("http.real_ip", req.Header.Get("X-Real-Ip")))
+	}
+	if req.Header.Get("X-Forwarded-For") != "" {
+		span.SetAttributes(attribute.String("http.real_ip", req.Header.Get("X-Forwarded-For")))
+	}
```
mloiseleur, line 52:
```suggestion
	if realIP := req.Header.Get("X-Real-Ip") : realIP != "" {
		span.SetAttributes(attribute.String("http.real_ip", realIP))
	}
	if forwardedFor := req.Header.Get("X-Forwarded-For") ; forwardedFor != "" {
		span.SetAttributes(attribute.String("http.real_ip", forwardedFor))
	}
```

rtribotte:
Hello @prutheus,

The attribute `http.real_ip` is not part of [semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/attributes-registry/http.md), and we are following them.

In the HTTP server attributes set, the `client.address` is described as:

> The IP address of the original client behind all proxies, if known (e.g. from [Forwarded#for](https://developer.mozilla.org/docs/Web/HTTP/Headers/Forwarded#for), [X-Forwarded-For](https://developer.mozilla.org/docs/Web/HTTP/Headers/X-Forwarded-For), or a similar header). Otherwise, the immediate client peer address.

In that regard, we may have to rework how we hydrate this attribute, and it should match your need, WDYT?

As well, the `client.socket.address` attribute should be reworked as it should represent the "immediate client peer address - unix domain socket name, IPv4 or IPv6 address".

prutheus:
@rtribotte thanks for the details. You are right, but with v3 I see tag `client.address` available and this is fine for me. Therefore, we can close this PR for now.

rtribotte:
Hello @prutheus,

Sure, nonetheless, as I mentioned the `client.address` attribute implementation in v3 doesn't look correct regarding what is expected in the semantic conventions.

> Therefore, we can close this PR for now.

Do you mean you will address this is another PR?

prutheus:
@rtribotte sorry, my requirement was to get the client IP for a tracing span. This is possible with v3, therefore I can not contribute further here due to lack of time.

rtribotte:
Hello @prutheus,

Sure, no worries, I'm closing this PR accordingly.

Thanks anyway!

PR #10416, opened 2024-02-07
Feature/vue3 migration

### LINKED ISSUE #9932: Upgrade UI development dependencies

andsarr:
### What does this PR do?

Vue.js 2.x has reached EOL and it should not be used anymore. This PR migrates to Quasar Framework 2.x and Vue.js 3.x.

### Motivation

Fixes #9932

### More

- [ ] Added/updated tests


andsarr:
@mmatur the current changes have migrated the WebUI from Vue.js v2.x to 3.x, from Quasar 1.x to v2.x and from Webpack to Vite. 

Should I stop at this point or should I also migrate from Vuex to Pinia?

mmatur:
@andsarr, to me you can stop it.

Thanks for your job on it.

mmatur (changes requested):
Previous version:
image

New version
image

@andsarr thanks for the contribution.

As you can see on previous screenshot, it seems some colors and components are not the same before and after the migration. Could you please have a look?

andsarr:
The color difference is fixed as well as the chart difference. 

 There is one problem that I haven't been able to fixed and that kind of happens randomly. When navigating to the to the Service Details Page or the Middleware Details Page, an exception is thrown:

Screenshot_20240213_230859

This is because the `on-load-more` property is not define for the `MainTable` used in those components. But I get the exception even after setting a default value to the property or defining it explicitly. I have no idea why.

andsarr:
The error in the `MainTable` component is fixed. @mmatur, @rtribotte  should I do something else to get this merged?

andsarr:
By the way, after the review, it is probably good to squash the commits before merging.

ldez:
You don't need to squash, our merge-bot always squash the commits.

webui/quasar.conf.js
```diff
@@ -114,6 +120,10 @@ module.exports = function (ctx) {
     supportIE: false,
 
     build: {
```
gndz07, line 122: Hello @andsarr, thank you for your job on this.
I just have a small suggestion to add `compilerOptions.isCustomElement` to get rid of this warning:

image

```suggestion
    build: {
     viteVuePluginOptions: {
        template: {
          compilerOptions: {
            isCustomElement: (tag) => tag.startsWith('hub-')
          }
        }
      },
```

gndz07 (approved):
LGTM

rtribotte (approved):
Thanks 🎉

nmengin (approved):

nmengin:
Hey @andsarr,

Thank you for the contribution 🎉

PR #10423, opened 2024-02-11
Add X-Amzn-Trace-Id span attribute for entrypoint 

changhyuni:
### What does this PR do?
Add X-Amzn-Trace-Id, the http trace header used by aws, in the span attribute
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-request-tracing.html

as requests come into the load balancer, Amazon will look for the X-Amzn-Trace-Id header. If it doesn't exist, Amazon will inject the tracing ID (and log it). And, if the HTTP header does exist, Amazon will modify it with the tracing ID 

### Motivation
When using traefik as a kubernetes ingress in an aws environment, you can trace between AWS ELB or API Gateway and Traefik it will help us correlate client-side requests with the traffic that goes through our load balancers.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/tracing/entrypoint.go
+++ b/pkg/middlewares/tracing/entrypoint.go
@@ -44,6 +44,10 @@ func (e *entryPointTracing) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	tracingCtx, span := e.tracer.Start(tracingCtx, "EntryPoint", trace.WithSpanKind(trace.SpanKindServer))
 	defer span.End()
 
+	if amznTraceId  := req.Header.Get("X-Amzn-Trace-Id") : amznTraceId != "" {
+		span.SetAttributes(attribute.String("http.amzn_trace_id", amznTraceId))
+	}
+
 	req = req.WithContext(tracingCtx)
 
 	span.SetAttributes(attribute.String("entry_point", e.entryPoint))
--- a/pkg/middlewares/tracing/entrypoint.go
+++ b/pkg/middlewares/tracing/entrypoint.go
@@ -44,8 +44,8 @@ func (e *entryPointTracing) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	tracingCtx, span := e.tracer.Start(tracingCtx, "EntryPoint", trace.WithSpanKind(trace.SpanKindServer))
 	defer span.End()
 
-	if amznTraceId  := req.Header.Get("X-Amzn-Trace-Id") : amznTraceId != "" {
-		span.SetAttributes(attribute.String("http.amzn_trace_id", amznTraceId))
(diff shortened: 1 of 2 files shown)


mmatur (changes requested):
Hello @changhyuni,

Thanks for your contribution.

Traefik uses OpenTelemetry for the tracing and we are following the [Semantic Conventions v1.23.1](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.1/docs/http/http-spans.md)

Span attribute `http.amzn_trace_id` is not listed as a valid attribute in the convention.

Could you please use the `http.response.header.<key>` and `http.request.header.<key>` span attribute instead of creating a custom attribute.

changhyuni:
@mmatur 
i fixed it.
Redefined attribute names to match conventions
`http.request.header.x-amzn_trace_id`

mmatur:
@changhyuni 

Thanks for adapting your PR, I will a bit picky but the [Semantic Convention v1.23.1](https://github.com/open-telemetry/semantic-conventions/blob/v1.23.1/docs/http/http-spans.md)  says:

> [4]: Instrumentations SHOULD require an explicit configuration of which headers are to be captured. Including all response headers can be a security risk - explicit configuration helps avoid leaking sensitive information. Users MAY explicitly configure instrumentations to capture them even though it is not recommended. The attribute value MUST consist of either multiple header values as an array of strings or a single-item array containing a possibly comma-concatenated string, depending on the way the HTTP library provides access to headers.

Could you adapt your PR regarding that?

If you don't have the time, please let us know and we will be able to handle that for you if needed.

changhyuni:
@mmatur 
thanks you for your detail i fixed it.
I followed the convention as closely as I could

According to the OpenTelemetry instructions, explicitly configure which headers to capture and
I'm following the instructions above by extracting a specific header (`X-Amzn-Trace-Id`) from the main code and putting the values from that header into an array

mmatur:
Hi @changhyuni,

Thanks for the change. 

I was thinking about adding an attribute in the static configuration to allow user to define which headers they want to add as a tracing attribute.

Referencing this sentence in the semantic convention:

> Instrumentations SHOULD require an explicit configuration of which headers are to be captured

Could you adapt in that regard please? If you don't have the time please let us know and we will be able to handle that for you on your PR

changhyuni:
@mmatur 
> I was thinking about adding an attribute in the static configuration to allow user to define which headers they want to add as a tracing attribute.
Thanks you it's good idea. 
it would be very helpful for tracing if users could add any header they want

I have a few ideas but
we need this feature, so it would be better if the PR team could do it as soon as possible on this PR.
Thanks you.

rtribotte:
Hello @changhyuni,

I have opened PR #10457 which supersedes yours as the changes were much more important and a bit outside the initial scope of this PR.

I'm closing this one accordingly.

PR #10428, opened 2024-02-13
Don't allow routers higher than internal ones

ldez:
### What does this PR do?

Use a dynamic approach to get the maxInt.

on 64 bits: MaxInt64
on 32 bits: MaxInt32

### Motivation

Don't allow routers higher than internal ones.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/static/entrypoints.go
+++ b/pkg/config/static/entrypoints.go
@@ -2,14 +2,12 @@ package static
 
 import (
 	"fmt"
+	"math"
 	"strings"
 
 	ptypes "github.com/traefik/paerser/types"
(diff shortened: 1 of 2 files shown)


mmatur (approved):
LGTM

rtribotte (approved):
LGTM

hhromic:
This comment is not a review, but I'm wondering why not using `math.MaxInt` ?
I like to learn from experienced engineers and this one got me thinking.

Ref: <https://pkg.go.dev/math#pkg-constants>

hhromic:
Thanks for the context and explanation, makes sense!

kevinpollet (approved):
LGTM

PR #10435, opened 2024-02-14
Add toggle to enable alpha APIs

SantoDE:
### Motivation
The [standard channel of Gateway API](https://gateway-api.sigs.k8s.io/guides/?h=install#install-standard-channel) does not include Alpha API Versions anymore. If a user does the standard channel installation and then starts Traefik, Traefik won't be able to start properly.

With adding that knob, users can opt-in into the usage of the alpha APIs in case they need / want to.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/gateway/client.go
+++ b/pkg/provider/kubernetes/gateway/client.go
@@ -46,7 +46,7 @@ func (reh *resourceEventHandler) OnDelete(obj interface{}) {
 // WatchAll starts the watch of the Provider resources and updates the stores.
 // The stores can then be accessed via the Get* functions.
 type Client interface {
-	WatchAll(namespaces []string, stopCh <-chan struct{}) (<-chan interface{}, error)
+	WatchAll(namespaces []string, enableAlphaAPis bool, stopCh <-chan struct{}) (<-chan interface{}, error)
 	GetGatewayClasses() ([]*gatev1.GatewayClass, error)
 	UpdateGatewayStatus(gateway *gatev1.Gateway, gatewayStatus gatev1.GatewayStatus) error
 	UpdateGatewayClassStatus(gatewayClass *gatev1.GatewayClass, condition metav1.Condition) error
@@ -154,7 +154,7 @@ func newExternalClusterClient(endpoint, caFilePath string, token types.FileOrCon
 }
 
 // WatchAll starts namespace-specific controllers for all relevant kinds.
-func (c *clientWrapper) WatchAll(namespaces []string, stopCh <-chan struct{}) (<-chan interface{}, error) {
+func (c *clientWrapper) WatchAll(namespaces []string, enableAlpha bool, stopCh <-chan struct{}) (<-chan interface{}, error) {
 	eventCh := make(chan interface{}, 1)
 	eventHandler := &resourceEventHandler{ev: eventCh}
 
@@ -195,13 +195,17 @@ func (c *clientWrapper) WatchAll(namespaces []string, stopCh <-chan struct{}) (<
 		if err != nil {
 			return nil, err
 		}
-		_, err = factoryGateway.Gateway().V1alpha2().TCPRoutes().Informer().AddEventHandler(eventHandler)
-		if err != nil {
-			return nil, err
-		}
-		_, err = factoryGateway.Gateway().V1alpha2().TLSRoutes().Informer().AddEventHandler(eventHandler)
-		if err != nil {
-			return nil, err
+
+		if enableAlpha {
+			_, err = factoryGateway.Gateway().V1alpha2().TCPRoutes().Informer().AddEventHandler(eventHandler)
+			if err != nil {
+				return nil, err
+			}
+			_, err = factoryGateway.Gateway().V1alpha2().TLSRoutes().Informer().AddEventHandler(eventHandler)
+			if err != nil {
+				return nil, err
+			}
+
 		}
 		_, err = factoryGateway.Gateway().V1beta1().ReferenceGrants().Informer().AddEventHandler(eventHandler)
 		if err != nil {
--- a/pkg/provider/kubernetes/gateway/kubernetes.go
+++ b/pkg/provider/kubernetes/gateway/kubernetes.go
@@ -58,6 +58,7 @@ type Provider struct {
 	LabelSelector    string                `description:"Kubernetes label selector to select specific GatewayClasses." json:"labelSelector,omitempty" toml:"labelSelector,omitempty" yaml:"labelSelector,omitempty" export:"true"`
 	ThrottleDuration ptypes.Duration       `description:"Kubernetes refresh throttle duration" json:"throttleDuration,omitempty" toml:"throttleDuration,omitempty" yaml:"throttleDuration,omitempty" export:"true"`
 	EntryPoints      map[string]Entrypoint `json:"-" toml:"-" yaml:"-" label:"-" file:"-"`
+	EnableAlphaAPIs  bool                  `description:"Toggles whether or not Alpha APIs should be used" json:"enableAlphaAPIs,omitempty" toml:"enableAlphaAPIs,omitempty" yaml:"enableAlphaAPIs,omitempty" export:"true"`
 
 	lastConfiguration safe.Safe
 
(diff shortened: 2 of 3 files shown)


xEBIx:
Thank you for filing this pull request.
I have some questions, are you just making v1alpha2 API Versions switchable, so I have the choice to use this or no one? To make it work we must use the API Version "V1" as cert-manager and of course Gateway API are already using it. 
Also I see that you are making TCP and TLS routes switchable, Dont we need all of them, like HTTProute, Gateway and Gatewayclass is also using these versions?
Or dont I understand something right?

rtribotte:
Hello @SantoDE,

Nice to see you around, and thanks for this contribution!

I started to review the PR, and first I wanted to know if toggling the informers' setup would be enough. After some testing, there are no errors raised, even if Traefik RBAC does not give rights for resources of the experimental channel.

Nonetheless, as the provider will go through resource discovery, toggling only the informers' setup, will not prevent the experimental channel resources (for now `TCPRoutes` and `TLSRoutes`) code path from being executed, and by doing so, the user won't get much information on why this part of the discovery fails. Moreover, debug logs would only mention that no resources were found, which is misleading.
I think the option could be used to raise an error if the gateway is using route kinds not in the standard channel if the option is not set and prevent further operation on resources that are not "allowed".

I noticed that Kubernetes CRDs are distributed distinctively for each channel, and an annotation (`gateway.networking.k8s.io/channel`) is present to indicate the channel in the definition. Maybe, we could also leverage this information somehow to check consistency with Traefik configuration. In that regard, it could answer @xEBIx question, Traefik would expect to be in a cluster with definitions matching its configuration, being to standard channel or experimental channel.

Regarding the option naming, I would find it more accurate to reuse the channel vocabulary, standard channel versus experimental channel, more than alpha versus beta.

xEBIx:
I dont get it, I am sorry. 
The Gateway API has moved forward to release the API V1, v1alpha1 is basicly deprecated. Making v1alpha1 switchable is only making sense to me, if the user has the old software in his cluster and doesnt want to update because of some reason. 
The Experimental or Standard question is not about API Versions, just the question if you need TLS and TCP Routes, that is correct, as far as I understood. For now the most urgend change would be to use the V1 API Version, both in the helm Chart and in Traefik itself.

SantoDE:
Thank you all for the comments. Let me try to elaborate :) 

@xEBIx:
That is not my understanding, as you can see in the public APIs. `v1` only has a subset of resources included, such as `HTTPRoute`, `Gateway` and so on.

`v1alpha2` is as of today, still perfectly in use for resources that have not graduated yet, such as e.g. `TLSRoute`. You can see that [here](https://github.com/kubernetes-sigs/gateway-api/tree/main/apis/v1alpha2).

Does that make more sense?

Last but not least, yes, bumping the CRDs on the other side is done already. [Helm Chart PR](https://github.com/traefik/traefik-helm-chart/pull/1001
)
@rtribotte:
Re-using or better, also checking on `gateway.networking.k8s.io/channel` could indeed make sense. I didn't thought about that before, but it could definitely make sense. Let me noodle on that for a bit :)

About the other code paths being executed:  I was under the impression, since we don't add the eventHandler, it wouldn't see changes. I was wrong :) Hence, I agree with you - I think we should check at the actual route kind when building config, whether or not we should process it. WDYT?

rtribotte:
Hello @SantoDE,

Sure, indeed I think checking if the route kind is supported seems to way to go.
Operations for one gateway should stop as soon as it is identified that it references "unauthorized" route kinds, and update its status accordingly.

xEBIx:
I am afraid, no it does not make sense to me. 
Please have a look into their v1.0.0 release CRD files (https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.0.0). 
There in the standard Version, you only have the referencegrants as v1alpha2, the rest of these CRDs is already V1. And furthermore the referencegrants is marked as deprecated in v1alpha2. The Experimental release has more CRDs still in v1alpha2, like tlsroute, that is true, but not the problem here. 

It wouldnt be that kind of a deal, if it would work in the end. But it does NOT! I cant install traefik with helm, because of the CRDs Traefik wants to have in v1alpha2 and Gateway has it in v1, and also cert-manager is using it in v1. So, in the process of creating certificates, certmanager is creating resources in v1 and Traefik searches them in v1alpha2, and of course cant find them, because they have been created in v1 by cert-manager. Please see my other issue here. Or the one at Gateway API: https://github.com/kubernetes-sigs/gateway-api/issues/2771

Furthermore there is another issue on there end, I created, because their admissions server did not support your created resources in there recent release, they reanabled those in the admission server again, more or less for traefik to work.

SantoDE:
Hello @xEBIx, 

I think there is a bit of confusion here. The `V1` release (i.e. 1.0.0) is distributed via two different channels. The standard installation and the experimental installation. The standard installation features all APIs which have graduated by now, which are  `Gateway`, `HTTPRoute` and so on. As you have seen that by yourself :)

However, not graduated resources yet do (as of now) still have the `v1alpha2` API Version. There is for example no `TCPRoute` with the `v1` API Version as of now.

However, your claim is correct. Prior to the Traefik version `3.0.0-rc1`, Traefik is looking for `v1alpha1` versions despite the fact that new versions are available. If you want to use Gateway API with its `V1`, I suggest you to test it with Traefik version `3.0.0-rc1`. This version is looking for `v1` resources where `v1` versions exists. 

Im also on the CNCF slack or the Kubernetes slack if you wanna discuss more.

pkg/provider/kubernetes/gateway/client.go
```diff
@@ -46,7 +46,7 @@ func (reh *resourceEventHandler) OnDelete(obj interface{}) {
 // WatchAll starts the watch of the Provider resources and updates the stores.
 // The stores can then be accessed via the Get* functions.
 type Client interface {
-	WatchAll(namespaces []string, stopCh <-chan struct{}) (<-chan interface{}, error)
+	WatchAll(namespaces []string, experimentalChannel bool, stopCh <-chan struct{}) (<-chan interface{}, error)
```
ldez, line 49: I think this can be moved the structure (and constructor)

rtribotte: @ldez I addressed your comment in the last review commit.

rtribotte (approved):
LGTM

youkoulayley (approved):
LGTM 👍

docs/content/providers/kubernetes-gateway.md
```diff
@@ -212,6 +212,29 @@ providers:
 --providers.kubernetesgateway.namespaces=default,production
 ```
 
+### `experimentalChannel`
+
+_Optional, Default: false_
+
+Toggles support for the Experimental Channel resources ([Gateway API release channels documentation](https://gateway-api.sigs.k8s.io/concepts/versioning/#release-channels)).
+This option currently enables support for `TCPRoutes` and `TLSRoutes`.
```
kevinpollet, line 220: I would use the name of the resource to not confuse the user, wdyt?

```suggestion
This option currently enables support for `TCPRoute` and `TLSRoute`.
```

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -51,13 +51,15 @@ const (
...
+	CertAuthFilePath    string              `description:"Kubernetes certificate authority file path (not needed for in-cluster client)." json:"certAuthFilePath,omitempty" toml:"certAuthFilePath,omitempty" yaml:"certAuthFilePath,omitempty"`
+	Namespaces          []string            `description:"Kubernetes namespaces." json:"namespaces,omitempty" toml:"namespaces,omitempty" yaml:"namespaces,omitempty" export:"true"`
+	LabelSelector       string              `description:"Kubernetes label selector to select specific GatewayClasses." json:"labelSelector,omitempty" toml:"labelSelector,omitempty" yaml:"labelSelector,omitempty" export:"true"`
+	ThrottleDuration    ptypes.Duration     `description:"Kubernetes refresh throttle duration" json:"throttleDuration,omitempty" toml:"throttleDuration,omitempty" yaml:"throttleDuration,omitempty" export:"true"`
+	ExperimentalChannel bool                `description:"Toggles Experimental Channel resources support (TCPRoutes, TLSRoutes...)." json:"experimentalChannel,omitempty" toml:"experimentalChannel,omitempty" yaml:"experimentalChannel,omitempty" export:"true"`
```
kevinpollet, line 60: I would do the same here, but do not miss to generate again the documentation.

```suggestion
	ExperimentalChannel bool                `description:"Toggles Experimental Channel resources support (TCPRoute, TLSRoute...)." json:"experimentalChannel,omitempty" toml:"experimentalChannel,omitempty" yaml:"experimentalChannel,omitempty" export:"true"`
```

PR #10445, opened 2024-02-16
fix: regression on flags using spaces between key and value

ldez:
### What does this PR do?

Fixes the regression on flags using spaces between key and value.

### Motivation

Fixes 10442

### More

- [ ] Added/updated tests

### Additional Notes

Slices are basically pointers on an array, so a function can modify a slice passed as an argument.

Demo:
- https://go.dev/play/p/IkNsUAo4ozj
- https://go.dev/play/p/Xoi_D1q4myn

### DIFF
--- a/pkg/cli/deprecation.go
+++ b/pkg/cli/deprecation.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"os"
 	"reflect"
+	"slices"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


rtribotte (approved):
LGTM

kevinpollet (approved):
LGTM

lbenguigui (approved):
LGTM

PR #8663, opened 2021-12-27
Add k8-crd external-service healthcheck

StephanSalas:
### What does this PR do?

This PR Enables users of the kubernetesCRD provider to create healthchecks for externalNamedServices within HttpIngressRoutes Kubernetes Custom Resources and leverage them within mirrored, weighted, and serverloadbalanced scenarios.

Importantly to the reviewer, these changes are not backward-compatible as they are adding a new healthcheck field to several CRDs. This might be the most important design consideration for this PR.

### Motivation

See Enhancement Proposal: https://github.com/traefik/traefik/issues/8541

### More
**Still need TODO**: I don't suspect this will be a long-running task.
- [ ] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

This is still in DRAFT status, placed in case any maintainers have initial suggestions or feedback. I will do the work to bring it up to merge-able status.

That said, I have tested this PR manually on my local machine, and things seem to work. I will certainly write new unit-tests/enhanced docs in the not-too-distant future.

PR #10467, opened 2024-02-23
Add LoadBalancer type in TraefikService CRD

### LINKED ISSUE #8144: Support for registering traefik to consul, AWS ELB, etc.

Feature

marcmognol:
### What does this PR do?
This PR adds the possibility to configure a standard service LoadBalancer type from a TraefikService CRD.

### Motivation
In some cases, we need to be able to configure a Load Balancer from a CRD instead of configuring it from file configuration only.
Service Load Balancer type adds some nice features like HealthCheck on external endpoint (URL)

Closes #8144 ?

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes_http.go
+++ b/pkg/provider/kubernetes/crd/kubernetes_http.go
@@ -213,9 +213,22 @@ func (c configBuilder) buildTraefikService(ctx context.Context, tService *traefi
 		return c.buildServicesLB(ctx, tService.Namespace, tService.Spec, id, conf)
 	} else if tService.Spec.Mirroring != nil {
 		return c.buildMirroring(ctx, tService, id, conf)
+	}	} else if tService.Spec.LoadBalancer != nil {
+		return c.buildLoadBalancer(ctx, tService.Spec, id, conf)
 	}
 
 	return errors.New("unspecified service type")
+}
+
+// buildLoadBalancer creates the configuration for the load-balancer of services named id, and defined in tService.
+// It adds it to the given conf map.
+func (c configBuilder) buildLoadBalancer(ctx context.Context, tService traefikv1alpha1.TraefikServiceSpec, id string, conf map[string]*dynamic.Service) error {
+
+	conf[id] = &dynamic.Service{
+		LoadBalancer: tService.LoadBalancer,
+	}
+
+	return nil
 }
 
 // buildServicesLB creates the configuration for the load-balancer of services named id, and defined in tService.
--- a/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
+++ b/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
@@ -2492,6 +2492,126 @@ spec:
                 required:
                 - name
                 type: object
+              loadBalancer:
+                description: Weighted defines the Weighted Round Robin configuration.
+                properties:
+                  sticky:
+                    description: |-
+                      Sticky defines the sticky sessions configuration.
+                      More info: https://doc.traefik.io/traefik/v3.0/routing/services/#sticky-sessions
+                    properties:
+                      cookie:
+                        description: Cookie defines the sticky cookie configuration.
+                        properties:
+                          httpOnly:
+                            description: HTTPOnly defines whether the cookie
+                              can be accessed by client-side APIs, such as JavaScript.
+                            type: boolean
+                          maxAge:
+                            description: |-
+                              MaxAge indicates the number of seconds until the cookie expires.
+                              When set to a negative number, the cookie expires immediately.
+                              When set to zero, the cookie never expires.
+                            type: integer
+                          name:
+                            description: Name defines the Cookie name.
+                            type: string
+                          sameSite:
+                            description: |-
+                              SameSite defines the same site policy.
+                              More info: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
+                            type: string
+                          secure:
+                            description: Secure defines whether the cookie can
+                              only be transmitted over an encrypted connection
+                              (i.e. HTTPS).
+                            type: boolean
+                        type: object
+                    type: object
+                  servers:
+                    description: cocou
+                    items:
+                      description: coucou
+                      properties:
+                        url:
+                          description: need a correct description
+                          type: string
+                        weight:
+                          description: need a correct description
+                          type: integer
+                        scheme:
+                          description: need a correct description
+                          type: string
+                        port:
+                          description: need a correct description
+                          type: string
+                      type: object
+                    type: array
+                  healthCheck:
+                    description: need a correct description
+                    properties:
+                      scheme:
+                        description: need a correct description
+                        type: string
+                      mode:
+                        description: need a correct description
+                        type: string
+                      path:
+                        description: need a correct description
+                        type: string
+                      method:
+                        description: need a correct description
+                        type: string
+                      status:
+                        description: need a correct description
+                        type: integer
+                      port:
+                        description: need a correct description
+                        type: integer
+                      interval:
+                        description: need a correct description
+                        type: string
+                      timeout:
+                        description: need a correct description
+                        type: string
+                      hostname:
+                        description: need a correct description
+                        type: string
+                      followRedirects:
+                        description: need a correct description
+                        type: boolean
+                      headers:
+                        description: need a correct description
+                        type: string
+                    required:
+                    - hostname
+                    type: object
+                  responseForwarding:
+                    description: ResponseForwarding defines how Traefik forwards
+                      the response from the upstream Kubernetes Service to the
+                      client.
+                    properties:
+                      flushInterval:
+                        description: |-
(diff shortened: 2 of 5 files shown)


marcmognol:
Hi @emilevauge,

This PR is related to issue 8147 for which you requested help from the community.
I would like, if you may, tell me if I am on the right direction with this draft PR ?

Then, if it is ok, I'll add unit tests and documentation.

I tested these modifications and it answers my need.

Thanks for your help.

Marc

marcmognol:
Hello,

Do I have to rebase it on v2.11 or  can I keep it on master ?

Thanks.

nmengin:
Hey @marcmognol,

No need to rebase the PR (it's an enhancement).
We mark it as `design-review` to allow the team to check your PR design before moving forward on the review.

kevinpollet:
Hello @marcmognol and thanks for this contribution,

Instead of allowing the user to define the server URLs in a TraefikService, we would prefer to leverage K8s `ExternalName` services and provide a property allowing the user to enable healtchecks for `ExternalName` services.

As explained in https://github.com/traefik/traefik/issues/8541#issuecomment-1081014912, I think you can get some inspiration from #8663. If that is ok for you, could you please also give some credits to @StephanSalas in the pull request description?

marcmognol:
Hello @kevinpollet 

Thank you for your reply.

I choose to do this way, because I thought it is easier to configure with the CRD only.

So, I followed your advice, and thanks to the PR of @StephanSalas I was able to add some changes to this PR.

Could you please tell me if I am in the right direction before going further ? Editing the title of this PR and add reference to the previous PR of @StephanSalas.

So, with this method, I can define Health Check on the service level.
My concerns is that when I define 2 services for a single Ingress Route, Traefik will create 3 services, 2 of type Load Balancer and one of type Weighted.

Is there any performance downside compared to a single service of type Load Balancer?
What is the difference between this configuration and a single Load Balancer?

Thanks for your answers.

Marc

marcmognol:
Hello @kevinpollet

I can achieve having single LoadBalancer with CRD when I use static Endpoints object for a Service. But I can only use IP address, I am not able to use FQDN of external services on the Endpoint.

That's why I think addind CRD LoadBalancer type can be a great idea.

Please let me know if we could also implement CRD for LoadBalancer?

Thanks

kevinpollet:
Hello @marcmognol and sorry for the late answer,

> Is there any performance downside compared to a single service of type Load Balancer?
> What is the difference between this configuration and a single Load Balancer?

There is not a real performance downside. This is because, in the IngressRoute configuration, it is possible to configure a weight for the service, which implies using a WRR. This is not true anymore since https://github.com/traefik/traefik/pull/10372, but this should not be changed in this PR.

> I can achieve having single LoadBalancer with CRD when I use static Endpoints object for a Service. But I can only use IP address, I am not able to use FQDN of external services on the Endpoint.

Here I am not sure to follow can you elaborate? 

Regarding the code, here https://github.com/traefik/traefik/pull/10467/files#diff-7d5c919304c65587ab6f978e73008d6d268e9bbb6db829eadf8d72f4f492e28fR306 the health check has to be set up only for external name services. For other services, we want to rely only on K8s healtcheck, did I miss something?

marcmognol:
Hello @kevinpollet,

Thanks for your reply.
It is good to know there is no performance downside regarding wrr and it seeams that it is something I could use in my situation.
I also need to pay attention that each request going through this ingressroute is transferred to the same service, so I need to use sticky cookie, but AFAIK the sticky cookie is set on the WRR level not the loabalancer itself.

In fact I would need to convert the following file configuration into CRDs:

```
  services:
    uat-service:
      loadBalancer:
        serversTransport: uat-transport
        servers:
          - url: https://localserver-1.asf.pri:443
          - url: https://localserver-2.asf.pri:443
        healthCheck:
          hostname: uat.local.pri
          path: /landing-page/
          interval: 15s
        sticky:
          cookie:
            secure: true
            httpOnly: true
```

Where keeping sticky cookie and health checks are mandatory.
That's why this PR implements the creation of CRD for LoadBalancer which is for my point of view the best solution.
But, feel free to tell what would be the best solution.

Thanks for your help on this.

kevinpollet:
Hello @marcmognol,

I think there is a misunderstanding, https://github.com/traefik/traefik/issues/8147 and https://github.com/traefik/traefik/issues/8541 was about adding the health check feature for [ExternalName](https://kubernetes.io/docs/concepts/services-networking/service/#externalname) services. For other K8s service types, we want to use the orchestrator health check which adds or removes endpoints dynamically to the corresponding services.

If I am not missing something by using ExternalName services it is already possible to do what you want, we are just missing the healtcheck feature (please note that the following [option](https://doc.traefik.io/traefik/providers/kubernetes-ingress/#allowexternalnameservices) has to be enabled).

Something like:

```yaml
---
kind: IngressRoute
metadata:
  name: foo
spec:
  routes:
    - match: Host(`foo.com`)
      kind: Rule
      services:
        - name: wrr
          kind: TraefikService

---
kind: TraefikService
metadata:
  name: wrr
spec:
  weighted:
    services:
       - name: external-svc1
         scheme: https
         port: 443
       - name: external-svc2
         scheme: https
         port: 443

---
apiVersion: v1
kind: Service
metadata:
  name: external-svc1
spec:
  externalName: localserver-1.asf.pri
  type: ExternalName

---
apiVersion: v1
kind: Service
metadata:
  name: external-svc2
spec:
  externalName: localserver-2.asf.pri
  type: ExternalName
```

Does it make sense?

marcmognol:
Hello @kevinpollet 

Yes it makes sense thank you. What about the sticky cookie which is required for me to ensure each request is routed to right service ?

If it is possible, then I'll update actual PR to add HealthCheck feature.

Marc

kevinpollet:
Hello @marcmognol,

> Yes it makes sense thank you. What about the sticky cookie which is required for me to ensure each request is routed to right service ?

I missed to add the sticky configuration in the example, as explained in https://doc.traefik.io/traefik/routing/providers/kubernetes-crd/#stickiness-and-load-balancing it is possible to configure the stickiness at the WRR level.

marcmognol:
Hello @kevinpollet
Thanks for everything. I've updated the PR. Could you please have a look ?

Thanks.

kevinpollet:
Hello @marcmognol, did I miss something, the code is still adding the health check configuration for all services? (https://github.com/traefik/traefik/pull/10467/files#diff-7d5c919304c65587ab6f978e73008d6d268e9bbb6db829eadf8d72f4f492e28fR306)

marcmognol:
Hello @kevinpollet 
You're right, I've added a check (+ unit test) when service type is not ExternalName and HealthCheck is defined.
Am I in the right direction ?

Thanks

marcmognol:
@kevinpollet I rebased my branch and now tests are failing. Could you please tell me if I am in the right direction before fixing my code to fit new unit tests ?

Thanks

pkg/provider/kubernetes/crd/kubernetes_http.go
```diff
@@ -392,6 +393,12 @@ func (c configBuilder) loadServers(parentNamespace string, svc traefikv1alpha1.L
 	}
 
 	var servers []dynamic.Server
+	if service.Spec.Type != corev1.ServiceTypeExternalName {
+		if svc.HealthCheck != nil {
+			return nil, fmt.Errorf("HealthCheck not allowed for externalName services: %s/%s", namespace, sanitizedName)
+		}
+	}
```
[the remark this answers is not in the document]
kevinpollet: Here I would prefer to do the check in the buildServersLB function, wdyt?

kevinpollet (changes requested):

kevinpollet:
Hello @marcmognol,

Sorry for the late answer, you are in the right direction, I added some suggestions.
Thanks for the work.

marcmognol:
Hi @kevinpollet 

Thanks for your feedback.
Feel free to tell me if I can do more for this PR.

Marc

kevinpollet:
Hi @marcmognol,

Last thing, could you add some documentation in the following section https://doc.traefik.io/traefik/routing/providers/kubernetes-crd/#kind-ingressroute?

Could you also rebase the pull request to fix the CI?

marcmognol:
@kevinpollet 

Thanks for your comment.
I've rebased the PR, but during this process I've made some mistakes, so I had to force push... Sorry.
I also added some documentation as requested.

Marc

kevinpollet (approved):
Thanks for the effort @marcmognol, and sorry again for the delay 👍

pkg/provider/kubernetes/crd/fixtures/with_one_external-service-and-health-check.yml
sdelicata, line 1: Could you rename your three fixture files using snake case, please?

sdelicata (changes requested):

youkoulayley (approved):
LGTM 👍

pkg/provider/kubernetes/crd/fixtures/with_one_external-service-and-health-check.yml
marcmognol: Yes, It is done. Marc

sdelicata (approved):
Thank you!

PR #10488, opened 2024-03-01
feat(accesslogs): allow filter accesslogs by service names whitelist

ttys3:
### What does this PR do?

Enhancements

allow filter accesslogs by service names whitelist

like this:

```yaml
# Configuring Multiple Filters
accessLog:
  filePath: "/path/to/access.log"
  format: json
  filters:
    serviceNames:
      - "service1@file"
      - "service2@consulcatalog"
      - "service3@consulcatalog"
  ```

### Motivation

we have traefik which proxy api requests to all kinds of service, its a very heavy server, we do not want traefik write all the logs,  only a few kind of important service we want access logs.

currently there's no way to filter like this in traefik.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/observability/access-logs.md
+++ b/docs/content/observability/access-logs.md
@@ -113,6 +113,7 @@ The available filters are:
 - `statusCodes`, to limit the access logs to requests with a status codes in the specified range
 - `retryAttempts`, to keep the access logs when at least one retry has happened
 - `minDuration`, to keep access logs when requests take longer than the specified duration (provided in seconds or as a valid duration format, see [time.ParseDuration](https://golang.org/pkg/time/#ParseDuration))
+- `serviceNames` to keep access logs with service names in the whitelist
 
 ```yaml tab="File (YAML)"
 # Configuring Multiple Filters
@@ -125,6 +126,10 @@ accessLog:
       - "300-302"
     retryAttempts: true
     minDuration: "10ms"
+    serviceNames:
+      - "service1@file"
+      - "service2@consulcatalog"
+      - "service3@consulcatalog"
 ```
 
 ```toml tab="File (TOML)"
--- a/pkg/middlewares/accesslog/filter_test.go
+++ b/pkg/middlewares/accesslog/filter_test.go
@@ -1,0 +1,1 @@
+package accesslog
--- a/pkg/middlewares/accesslog/logger.go
+++ b/pkg/middlewares/accesslog/logger.go
@@ -57,13 +57,14 @@ type handlerParams struct {
 
 // Handler will write each request and its response to the access log.
 type Handler struct {
-	config         *types.AccessLog
-	logger         *logrus.Logger
-	file           io.WriteCloser
-	mu             sync.Mutex
-	httpCodeRanges types.HTTPCodeRanges
-	logHandlerChan chan handlerParams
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by rtribotte** [comment]
Hello @ttys3,

Thanks for this contribution!

As this would be an enhancement, the changes should target the master branch for the next v3 minor version.

Should the filtering of access logs be configurable statically, at least only, since services are dynamic resources?
We are unsure if this is the best approach, WDYT?

**Comment by ttys3** [comment]
> Hello @ttys3,
> 
> Thanks for this contribution!
> 
> As this would be an enhancement, the changes should target the master branch for the next v3 minor version. Could you please rebase it?
> 
> Should the filtering of access logs be configurable statically, at least only, since services are dynamic resources? We are unsure if this is the best approach, WDYT?

rebased and change target to master

Yes, I agree that static array filters would be a little not extenable.

I'd glad to change it from `ServiceNames  []string` to something like `ServiceNamePatterns  []string`

so people could use regex to match service names, but compared to static array filters, this will need a little more cpu.

do you think it is better to change it to regex ?

**Comment by rtribotte** [comment]
Hello @ttys3,

I was implying that the current option is static, meaning the configuration won't change at runtime and that this could be a limitation because one could not often dress the list of service names before they exist. Also, with some providers, the service name is determined automatically, so it would be difficult to list them beforehand (even with regular expression support).
I suggest controlling whether to enable access-logs for a given service based on its configuration, WDYT?

**Comment by ttys3** [comment]
> Hello @ttys3,
> 
> I was implying that the current option is static, meaning the configuration won't change at runtime and that this could be a limitation because one could not often dress the list of service names before they exist. Also, with some providers, the service name is determined automatically, so it would be difficult to list them beforehand (even with regular expression support). I suggest controlling whether to enable access-logs for a given service based on its configuration, WDYT?

do you mean use access-logs as dynamic middleware ?

**Comment by rtribotte** [comment]
Hello @ttys3,

No, it shouldn't be an access-log middleware, my suggestion is to be able to opt-out inside the service configuration, with something like:
```yaml
services:
    service1:
        accesslogs: false
```

I apologize for the misunderstanding, I should have been more specific in my proposal.

ttys3:
the problem is `service1`, in file provider, we always write static service name, it is no problem here.

but in dynamic provider, like consul,  

according to the doc https://doc.traefik.io/traefik/v3.0/routing/providers/consul-catalog/#services

for example, we must config a tag(label) like this:

```
traefik.http.services.{name-of-your-choice}.accesslogs=false
```

for example

```
traefik.http.services.service1.accesslogs=false
```

the problem is, is this OK to write the `service1`? Because the service name is basically auto calculated.

rtribotte:
Hello @ttys3,

Actually, the file provider is no different from any other and produces a dynamic configuration that can be updated at runtime.

In fact, in order to provide a configuration for a service, it needs to be "named", which in the case of "orchestrator" providers, somewhat defeats the purpose of automatic service discovery and hence naming.
But it is no different for any service configuration options, it would not create precedence.

Perhaps it's a hint that this configuration should belong to the router, not the service. I'm not sure, both make sense in my opinion.

WDYT?

ttys3:
close due to https://github.com/traefik/traefik/pull/11308 already meet the need

Issue #9757, opened 2023-03-02
Incorrect X-FORWARDED-PROTO, X-FORWARDED-PORT for external TLS termination and proxy protocol


peterkappelt:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I think I've discovered an edge case (or myself not completely understanding the docs) that relates to setting the `X-FORWARDED-PROTO` and `-PORT` when proxy protocol is enabled.

I was able to reproduce the bug with this minimal setup:
* my browser @ `80.1.2.3`
  * connects to the public ip of the external load balancer
* an external load balancer service, does TLS termination, `10.11.0.200`
  * wireshark shows that it correctly sets `X-Forwarded-Proto https`, `X-Forwarded-Port 443`
  * proxies to Traefik on port 8181, proxy protocol enabled
* Traefik
  * `--entrypoints.web.address=:8181/tcp --entrypoints.web.forwardedHeaders.trustedIPs=10.11.0.200 --entrypoints.web.proxyProtocol.trustedIPs=10.11.0.200` 
  * service that connects to the sample app
* Sample App, simply dumps all request headers in the response

I'd expect the sample app to receive those headers:
```
x-forwarded-for 80.1.2.3, 10.11.0.200
x-forwarded-host hostname.example
x-forwarded-port 443
x-forwarded-proto https
x-forwarded-server traefik-node-name
x-real-ip 80.1.2.3
```

### What did you see instead?

Traefik forwards the headers, except for `X-FORWARDED-PROTO` and `-PORT`. The sample app receives:
```
x-forwarded-for 80.1.2.3 
x-forwarded-host hostname.example
x-forwarded-port 80
x-forwarded-proto http
x-forwarded-server traefik-node-name
x-real-ip 80.1.2.3
```

I can get the expected `x-forwarded-proto` and `x-forwarded-port` by:
* disabling proxy protocol on both Traefik and the external load balancer
* setting  `forwardedHeaders.trustedIPs=0.0.0.0/0` (or `forwardedHeaders.insecure=true`)
* setting `forwardedHeaders.trustedIPs=80.1.2.3` (my client IP)

None of this is really suitable for production. 

[...]

PR #10492, opened 2024-03-03
Trust Forwarded headers from the peer socket for PROXY Protocol connection

andsarr:
### What does this PR do?
 When the PROXY Protocol is enabled, and the PROXY Header is sent, the [XForwarded ](https://github.com/traefik/traefik/blob/c1ef7429771104e79f2e87b236b21495cb5765f0/pkg/middlewares/forwardedheaders/forwarded_header.go#L44) only sees the remote address contained on the PROXY Header since the `go-proxyproto.Listener` has already modified the connection (technically it just wraps it) in the TCP layer. 
 This means, that the  `XForwarded` cannot determined if the remote proxy (the server connection to Traefik) is allowed to forward headers. 

This PR makes the `XForwarded` aware of the PROXY Protocol. When a new HTTP connection is established, the IP of the proxy is added to the request context. 

### Motivation

Fixes [9757](https://github.com/traefik/traefik/issues/9757)

### More

- [ ] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -366,15 +366,25 @@ func (e *TCPEntryPoint) SwitchRouter(rt *tcprouter.Router) {
 	}
 }
 
-// writeCloserWrapper wraps together a connection, and the concrete underlying
+// proxyProtoConn wraps together a connection, and the concrete underlying
 // connection type that was found to satisfy WriteCloser.
-type writeCloserWrapper struct {
+type proxyProtoConn struct {
 	net.Conn
 	writeCloser tcp.WriteCloser
+	remoteAddr  proxyProtoAddr
 }
 
-func (c *writeCloserWrapper) CloseWrite() error {
-	return c.writeCloser.CloseWrite()
+type proxyProtoAddr struct {
+	net.Addr
+	peerSocketAddr string
+}
+
+func (p *proxyProtoConn) CloseWrite() error {
+	return p.writeCloser.CloseWrite()
+}
+
+func (p *proxyProtoConn) RemoteAddr() net.Addr {
+	return p.remoteAddr
 }
 
 // writeCloser returns the given connection, augmented with the WriteCloser
@@ -386,7 +396,15 @@ func writeCloser(conn net.Conn) (tcp.WriteCloser, error) {
 		if !ok {
 			return nil, errors.New("underlying connection is not a tcp connection")
 		}
-		return &writeCloserWrapper{writeCloser: underlying, Conn: typedConn}, nil
+
+		return &proxyProtoConn{
+			Conn:        typedConn,
+			writeCloser: underlying,
+			remoteAddr: proxyProtoAddr{
+				Addr:           typedConn.RemoteAddr(),
+				peerSocketAddr: typedConn.Raw().RemoteAddr().String(),
+			},
+		}, nil
 	case *net.TCPConn:
 		return typedConn, nil
 	default:
@@ -561,11 +579,9 @@ func createHTTPServer(ctx context.Context, ln net.Listener, configuration *stati
 		return nil, err
 	}
 
-	proxyProtocolEnabled := configuration.ProxyProtocol != nil
 	var handler http.Handler
 	handler, err = forwardedheaders.NewXForwarded(
 		configuration.ForwardedHeaders.Insecure,
-		proxyProtocolEnabled,
 		configuration.ForwardedHeaders.TrustedIPs,
 		configuration.ForwardedHeaders.Connection,
 		next)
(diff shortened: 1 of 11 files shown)


pkg/server/server_entrypoint_tcp.go
```diff
@@ -616,6 +618,21 @@ func createHTTPServer(ctx context.Context, ln net.Listener, configuration *stati
 
 	prevConnContext := serverHTTP.ConnContext
 	serverHTTP.ConnContext = func(ctx context.Context, c net.Conn) context.Context {
+		// This adds the remote address of the server making the connection if the PROXY Protocol is enabled
```
andsarr, line 621: I am wondering if it is better to use this code instead:

```go
	// This adds the remote address of the server making the connection if the PROXY Protocol is enabled
		getProxyIP := func(c net.Conn) string {
			for {
				switch conn := c.(type) {
				case *tcprouter.Conn:
					c = conn.WriteCloser
				case *trackedConnection:
					c = conn.WriteCloser
				case *writeCloserWrapper:
					c = conn.writeCloser
				case *net.TCPConn:
					return c.RemoteAddr().String()
				default:
					return ""
				}
			}
		}

		if proxyProtocolEnabled {
			proxyIP := getProxyIP(c)
			ctx = context.WithValue(ctx, forwardedheaders.ProxyAddrKey, proxyIP)
		}
```

It looks less brittle since it does not depends on the exact order of the "wrapping".  What do you think @juliens, @mmatur ?

andsarr: I am going with this variant!

andsarr:
Hi guys! 

 Can we get this merged?

kevinpollet:
Hello @andsarr and sorry for the delay,

We reviewed this pull request and we think that another approach could work, avoiding the type unwrapping in an infinite for-loop.

The idea is to change the [`writeCloserWrapper`](https://github.com/traefik/traefik/blob/v2.11/pkg/server/server_entrypoint_tcp.go#L368) implementation to override the `RemoteAddr` method. The overridden `RemoteAddr` method could then return a custom struct allowing to retrieve the previous hop IP.

Thus in the `ConnContext` function calling the `RemoteAddr` method on the `net.Conn` would return the custom `Addr` implementation, allowing to assert that we are in a proxyproto context and to fill the request context with the right IP.

As a side note, using `net.SplitHostPort` is necessary to remove the port in returned Addr.

WDYT?
Do you want to address those changes, or do you prefer us to make them?

andsarr:
Hi @kevinpollet 

 Please do the changes!

kevinpollet (dismissed):
Thanks 👍

rtribotte:
@andsarr @peterkappelt @yongzhang 
We have reworked the PR and added adaptive behavior for the `X-Forwarded-For` header on the reverse proxy side.
With the changes, the IP that matches the `trustedIPs` configuration is the IP added to the `X-Forwarded-For` header, whether it is the "original" IP provided by the PROXY protocol or the peer socket IP.
So it is up to the `trustedIPs` configuration to control whether to use the peer socket IP or the IP provided by the PROXY protocol.

We would like to get your feedback on this.
We are still unsure if this solution won't break something and want to give it some time to think about.

andsarr:
Looks good to me!

pkg/server/service/proxy.go
```diff
@@ -37,45 +38,71 @@ func buildProxy(passHostHeader *bool, responseForwarding *dynamic.ResponseForwar
...
+			// In case of a ProxyProtocol connection the http.Request#RemoteAddr is the original one,
+			// and if the Forwarded headers were trusted from the peer socket address (see trustedIPs) and not from the original IP,
+			// the peer socket address should be used instead.
+			// The ForwardedHeaders middleware provides the address to be used to populate the X-Forwarded-For header.
+			if xForwardedForAddr, ok := req.In.Context().Value(forwardedheaders.XForwardedForAddr).(string); ok {
+				remoteAddr = xForwardedForAddr
+			}
+
+			// Adapted from httputil.ReverseProxy.
+			if clientIP, _, err := net.SplitHostPort(remoteAddr); err == nil {
```
lbenguigui, line 77: I think that when the port is provided in the Proxy Protocol, we should populate the ```X-Forwarded-Port``` header with that value, rather than ignoring it.

lbenguigui (changes requested):

juliens:
Hi,

To move forward with this PR, I’d like to invite everyone to share their thoughts on the linked issue #9757 . Your input there would be really helpful in determining the best path forward.

Thanks.

juliens:
As explained on the related [issue](https://github.com/traefik/traefik/issues/9757#issuecomment-2886530808)).

I close this PR. 

Thank you all for helping on this.

PR #10499, opened 2024-03-05
Fix log when default TLSStore and TLSOptions are defined multiple times

rtribotte:
### What does this PR do?

This PR fixes the error log when default TLSStore and TLSOptions are defined multiple times.

### Motivation

To clarify the error logs.

### More

- [ ] Added/updated tests


ldez (approved):
LGTM

kevinpollet (approved):

PR #10525, opened 2024-03-15
docs: improve middleware example

mloiseleur:
### What does this PR do?

Remove optional `namespace` option of middleware.

### Motivation

Shows that `Middleware` in the same namespace that `IngressRoute` can be used without specifying it.

### More

### Additional Notes

 Just above, the other example show how to specify namespace on middleware when `IngressRoute` is using a different namespace:
image


emilevauge:
Hey @mloiseleur, thanks for your contribution :)
The team do agree it should be more explicit that this option is optionnal, but deleting it from the example may not be the best. We would prefer instead completing the `# [7]` comment to specify that `namespace ` is optionnal.

mloiseleur:
Thanks @emilevauge. 
Following an unexpected behavior with a git command, I'm not able to pursue on this PR.
We can continue on #10532.

PR #10532, opened 2024-03-18
docs: improve middleware example

mloiseleur:
### What does this PR do?

Specify that `namespace` option of middleware is optional.

### Motivation

Shows that `Middleware` in the same namespace that `IngressRoute` can be used without specifying it.

### More

### Additional Notes

 Just above, the other example show how to specify namespace on middleware when `IngressRoute` is using a different namespace:
image


emilevauge:
Thanks @mloiseleur, I would be even more verbose on the description, why not explain your use case directly?

> Shows that Middleware in the same namespace that IngressRoute can be used without specifying it.

rtribotte (approved):
LGTM

kevinpollet (approved):
Thanks 👍

lbenguigui (approved):
LGTM

PR #10550, opened 2024-03-27
Adjust ECS network interface detection logic

### LINKED ISSUE #10548: ECS provider misdetecs tasks with attached volumes as missing network interfaces when not in `awsvpc` network mode
### Welcome!

### What did you do?

The ECS provider should register tasks that have no network interfaces when the network type is not `awsvpc`. The [current](https://github.com/traefik/traefik/blob/c1ef7429771104e79f2e87b236b21495cb5765f0/pkg/provider/ecs/ecs.go#L322) logic rejects tasks that have attachments with no network interfaces, but this is for example a valid configuration when using `bridge` networking and having EBS volumes attached.

Here is an excerpt from what the API returns when describing the task in question - this task has functional networking and doesn't have any interfaces attached as it uses bridge networking.
```
{
    "tasks": [
        {
            "attachments": [
                {
                    "id": "xxx",
                    "type": "AmazonElasticBlockStorage",
                    "status": "ATTACHED",
                    "details": [
                        {
                            "name": "deleteOnTermination",
                            "value": "true"
                        },
                        {
                            "name": "roleArn",
                            "value": "arn:aws:iam::123456789012:role/task-role"
                        },
                        {
                            "name": "volumeName",
                            "value": "ebs-volume"
                        },
                        {
                            "name": "volumeId",
                            "value": "vol-abc123xyz890"
                        }
                    ]
                }
            ],
...
            "containers": [
                {
                    "networkInterfaces": [],
...
```

Excerpt from task definition:
```
{
...
        "networkMode": "bridge",
...
```

### What did you see instead?

Traefik's ECS provider misdetects ECS tasks with an attached volume as missing a network interface and refuses to register them:

```
{"level":"error","msg":"Skip container FOO: no network interfaces","providerName":"ecs","time":"2024-03-27T09:32:44Z"}
```

[2.11.0](https://hub.docker.com/layers/library/traefik/v2.11.0/images/sha256-903f89ddacc3059657565d3e5b3b8e8dcba4cc3b8af08183dad6fdc415459111)

```
Version:      2.11.0
Codename:     cheddar
Go version:   go1.22.0
Built:        2024-02-12T15:26:45Z
OS/Arch:      linux/amd64
```

Relevant portion:

```
            --providers.ecs.autodiscoverclusters=true \
            --providers.ecs.exposedbydefault=false
```

_No response_

amaxine:
### What does this PR do?

Fixes #10548.

ECS task attachments aren't uniquely related to network interfaces, and they can also be returned by the API when e.g. using attached EBS storage. As such this condition should only be checked when using `awsvpc` networking mode, where a network interface should be present, and not any others.

### Motivation

Affected by above mentioned bug.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/ecs/ecs.go
+++ b/pkg/provider/ecs/ecs.go
@@ -314,7 +314,7 @@ func (p *Provider) listInstances(ctx context.Context, client *awsClient) ([]ecsI
 				}
 
 				var mach *machine
-				if len(task.Attachments) != 0 {
+				if aws.StringValue(taskDef.NetworkMode) == "awsvpc" && len(task.Attachments) != 0 {
 					if len(container.NetworkInterfaces) == 0 {
 						logger.Errorf("Skip container %s: no network interfaces", aws.StringValue(container.Name))
 						continue


mmatur (approved):
LGTM
Thanks for this fix

rtribotte (approved):
Thanks!

jspdown (approved):

PR #10645, opened 2024-04-21
first draft that allows non-running containers to be retrieved using allowEmptyServices

acouvreur:
…allowEmptyServices

### What does this PR do?

This PR is a draft attempting to change the `allowEmptyServices` capabilities to the Docker provider by allowing non running containers to be discovered.

### Motivation

Non-running containers not returning 404, but 503 instead and having the full middleware chain built and executed on requests.

https://github.com/traefik/traefik/issues/9907

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/docker/config.go
+++ b/pkg/provider/docker/config.go
@@ -276,6 +276,9 @@ func (p *DynConfBuilder) addServer(ctx context.Context, container dockerData, lo
 	}
 
 	if port == "" {
+		if p.AllowEmptyServices {
+			return nil
+		}
 		return errors.New("port is missing")
 	}
 
@@ -310,7 +313,7 @@ func (p *DynConfBuilder) getIPPort(ctx context.Context, container dockerData, se
 		port = getPort(container, serverPort)
 	}
 
-	if len(ip) == 0 {
+	if len(ip) == 0 && !p.AllowEmptyServices {
 		return "", "", fmt.Errorf("unable to find the IP address for the container %q: the server is ignored", container.Name)
 	}
 
--- a/pkg/provider/docker/pdocker.go
+++ b/pkg/provider/docker/pdocker.go
@@ -165,7 +165,9 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
 }
 
 func (p *Provider) listContainers(ctx context.Context, dockerClient client.ContainerAPIClient) ([]dockerData, error) {
-	containerList, err := dockerClient.ContainerList(ctx, dockertypes.ContainerListOptions{})
+	containerList, err := dockerClient.ContainerList(ctx, dockertypes.ContainerListOptions{
+		All: p.AllowEmptyServices,
+	})
 	if err != nil {
 		return nil, err
 	}
@@ -173,7 +175,7 @@ func (p *Provider) listContainers(ctx context.Context, dockerClient client.Conta
 	var inspectedContainers []dockerData
 	// get inspect containers
 	for _, container := range containerList {
-		dData := inspectContainers(ctx, dockerClient, container.ID)
+		dData := inspectContainers(ctx, dockerClient, container.ID, p.AllowEmptyServices)
 		if len(dData.Name) == 0 {
 			continue
 		}
--- a/pkg/provider/docker/shared.go
+++ b/pkg/provider/docker/shared.go
@@ -35,7 +35,7 @@ type Shared struct {
 	defaultRuleTpl *template.Template
 }
 
-func inspectContainers(ctx context.Context, dockerClient client.ContainerAPIClient, containerID string) dockerData {
+func inspectContainers(ctx context.Context, dockerClient client.ContainerAPIClient, containerID string, allowEmptyServices bool) dockerData {
 	containerInspected, err := dockerClient.ContainerInspect(ctx, containerID)
 	if err != nil {
 		log.Ctx(ctx).Warn().Err(err).Msgf("Failed to inspect container %s", containerID)
@@ -43,8 +43,8 @@ func inspectContainers(ctx context.Context, dockerClient client.ContainerAPIClie
 	}
 
 	// This condition is here to avoid to have empty IP https://github.com/traefik/traefik/issues/2459
-	// We register only container which are running
-	if containerInspected.ContainerJSONBase != nil && containerInspected.ContainerJSONBase.State != nil && containerInspected.ContainerJSONBase.State.Running {
+	// We register only container which are running unless we explicitly want all containers
+	if allowEmptyServices || containerInspected.ContainerJSONBase != nil && containerInspected.ContainerJSONBase.State != nil && containerInspected.ContainerJSONBase.State.Running {
 		return parseContainer(containerInspected)
 	}
 
@@ -201,6 +201,10 @@ func getPort(container dockerData, serverPort string) string {
 }
 
 func getServiceName(container dockerData) string {
+	if condition {
+		
+	}
+
 	serviceName := container.ServiceName
 
 	if values, err := getStringMultipleStrict(container.Labels, labelDockerComposeProject, labelDockerComposeService); err == nil {


rtribotte:
Hello @acouvreur,

The extension of the Docker provider `allowEmptyServices` option to include stopped containers could have a side effect that does not exist with other providers, or at least not the same meaning.

As the service definition with a docker container isn't a really "thing" (at most a docker-compose service), that Traefik can read and rely on, multiple stopped containers providing different dynamic configurations for the same resources (routers, middleware, etc...) can co-exist, which will result as being in error and dropped because of conflicting configurations found.

With docker-compose, this is mitigated if the service name never changes, as the stopped containers for the service are cleared.

With other providers, the service definition is what Traefik uses to create the configuration, and because this is not possible with Docker, I am wondering if this behavior should be controlled with the same `allowEmptyServices` option, WDYT?

aless3:
> Hello @acouvreur,
> 
> Thank you for this contribution!
> 
> The extension of the Docker provider `allowEmptyServices` option to include stopped containers could have a side effect that does not exist with other providers, or at least not the same meaning.

what about making `allowEmptyServices` a flag with 3 states? like on/off/also stopped or something like that? I don't know how go treats booleans and I don't know if this is feasible. 
Or maybe another variable like `allowStoppedServices`? defaulting to off to not break existing workflows and basing this pr on that instead of `allowEmptyServices`? 


> As the service definition with a docker container isn't a really "thing" (at most a docker-compose service), that Traefik can read and rely on, multiple stopped containers providing different dynamic configurations for the same resources (routers, middleware, etc...) can co-exist, which will result as being in error and dropped because of conflicting configurations found.

well, making the default behavior with `allowEmptyServices=true` the same as now, and adding a way to allow users to "opt in" to the behavior of this pr, wouldn't this be an user error more than a traefik error?

mrbreil:
I'm interested in it too. I would like to have a custom error page for a stopped container. and it only works if the container is running or unhealthy. If the container is stopped, I just get a simple 404.

acouvreur:
It's been a while!

I've reworked this all from scratch.

Basically all it does now is that the container status is now reported in the structure `dockerData`, which is then used to return a `nil` server if `AllowEmptyServices` is set to true.

We can talk about have a dedicated field for that feature specifically, but in my opinion this is what the feature is supposed to be.

`AllowEmptyServices` is about allowing services that are empty to be registered by Traefik, which is shown like that in the UI:

image

It returns a `503` no available server as expected, and you can have a middleware in front catching the request and do whatever you want with it.

acouvreur:
Any news @rtribotte ?

rtribotte:
Hi @acouvreur, sorry for the late reply.

We have given some more thought to changing the behaviour of the `allowEmptyServices` option and have come to the conclusion that we cannot introduce a breaking change.
In fact, exposing stopped containers from now on could lead to conflicting exposed routers, where the empty service router can take precedence over one with active service, and thus breaking some setups.

Also, as mentioned above, the lack of service definition in Docker is the a blocker to considering a stopped container as defining a service.

So we do not think that using this option in your use case is the best option.

However, we think the ability to discover a stopped container could be an opt-in feature per container by introducing a new label, WDYT?

Kitof:
@rtribotte Thanks for your answer. Just to be sure to understand correcly, your proposition keep the PR of @acouvreur but with an additionnal a config parameter like "visibleWhenStopped" with default value to false ?

rtribotte:
@Kitof Sure, that is indeed the suggestion.

acouvreur:
> Hi @acouvreur, sorry for the late reply.
> 
> We have given some more thought to changing the behaviour of the `allowEmptyServices` option and have come to the conclusion that we cannot introduce a breaking change. In fact, exposing stopped containers from now on could lead to conflicting exposed routers, where the empty service router can take precedence over one with active service, and thus breaking some setups.
> 
> Also, as mentioned above, the lack of service definition in Docker is the a blocker to considering a stopped container as defining a service.
> 
> So we do not think that using this option in your use case is the best option.
> 
> However, we think the ability to discover a stopped container could be an opt-in feature per container by introducing a new label, WDYT?

Let's add a new label to enable that capability.

However, I will need guidance on how to provide an empty loadbalancer when there is no IP attached.

I've been having some issues with the way I've implemented this, so any feedback or even contributions to the implementation would be really appreciated.

rtribotte:
Hi @acouvreur,

Nice!

> I've been having some issues with the way I've implemented this, so any feedback or even contributions to the implementation would be really appreciated.

When we actively review this PR, hopefully soon, we can propose to address this by pushing review commits, if you agree.

nickshanks347:
> Hi @acouvreur,
> 
> Nice!
> 
> > I've been having some issues with the way I've implemented this, so any feedback or even contributions to the implementation would be really appreciated.
> 
> When we actively review this PR, hopefully soon, we can propose to address this by pushing review commits, if you agree.



> Hi @acouvreur,
> 
> Nice!
> 
> > I've been having some issues with the way I've implemented this, so any feedback or even contributions to the implementation would be really appreciated.
> 
> When we actively review this PR, hopefully soon, we can propose to address this by pushing review commits, if you agree.

I appreciate that you, @rtribotte and @kevinpollet may be busy, but do you have any sort of timeframe as to when this could be reviewed? Also @acouvreur, I understand in a comment above you were having some issues with the way you implemented this, so it'd be useful if you expand on that so we can work towards a solution. I am super keen to see this merged!

Thanks!

Kitof:
Hi, any news about this PR ? Thanks a lot.

acouvreur:
Need some overall guidance by the Traefik developers on how they would implement a noop service that would return a 503.

nmengin:
Hey @acouvreur,

Thanks for reaching out.
Your PR is in our (maintainers') backlog.

It is a hot topic on our end, and we are interested in it.

> Need some overall guidance by the Traefik developers

Does it mean that you want to move forward on it or do you still agree to let us push on the PR as necessary?

acouvreur:
> Hey @acouvreur,
> 
> Thanks for reaching out. Your PR is in our (maintainers') backlog.
> 
> It is a hot topic on our end, and we are interested in it.
> 
> > Need some overall guidance by the Traefik developers
> 
> Does it mean that you want to move forward on it or do you still agree to let us push on the PR as necessary?

If you are going to work on such feature, please do so, otherwise I'd need some small help understanding your docker provider to be able to continue the development.

Kitof:
Is it possible to help? This PR is already 17 months old. :'(

lbenguigui:
Hey @acouvreur,
I updated your PR to address the feedback. Could you check if the changes look good to you?

acouvreur:
> Hey @acouvreur, I updated your PR to address the feedback. Could you check if the changes look good to you?

Looks good to me! Thanks a lot for such feature!

lbenguigui (approved):
LGTM

pkg/provider/docker/config.go
```diff
@@ -104,6 +104,12 @@ func (p *DynConfBuilder) build(ctx context.Context, containersInspected []docker
 }
 
 func (p *DynConfBuilder) buildTCPServiceConfiguration(ctx context.Context, container dockerData, configuration *dynamic.TCPConfiguration) error {
+	// Only allow running containers or containers with visibleWhenNotRunning=true
+	// Empty status is treated as running for backward compatibility
+	if container.Status != "" && container.Status != containertypes.StateRunning && !container.ExtraConf.VisibleWhenNotRunning {
```
sdelicata: This condition is duplicated multiple times. WDYT about extracting this condition into a `shouldIncludeContainer(container dockerData)` helper method?

pkg/provider/docker/config.go
```diff
@@ -114,7 +120,8 @@ func (p *DynConfBuilder) buildTCPServiceConfiguration(ctx context.Context, conta
 		}
 	}
 
-	if container.Health != "" && container.Health != containertypes.Healthy {
+	if (container.Health != "" && container.Health != containertypes.Healthy) ||
```
sdelicata: Same here with a `shouldAddServers(container dockerData)`?

docs/content/reference/routing-configuration/other-providers/docker.md
```diff
@@ -688,6 +688,16 @@ You can tell Traefik to consider (or not) the container by setting `traefik.enab
...
+#### `traefik.docker.visiblewhennotrunning`
+
+```yaml
+- "traefik.docker.visiblewhennotrunning=true"
+```
+
+Controls whether containers that are not in "running" state (e.g., stopped, paused, exited) should still be visible to Traefik for service discovery.
+
+By default, Traefik only considers containers in "running" state.
+
```
sdelicata, line 711: I suggest adding a bit more context:
```suggestion
When a container stops, its routers and services are removed from the configuration, resulting in 404 Not Found responses.

When this label is set to true, Traefik will:
- Keep the router and service configuration even when the container is not running
- Create services with empty backend server lists
- Return 503 Service Unavailable for requests to stopped containers (instead of 404 Not Found)
- Execute the full middleware chain, allowing middlewares to intercept requests
```
If you agree, please add it to the swarm provider documentation too.

rtribotte (approved):
Hi @acouvreur,

We pushed a review commit.
We limited the impact of the option only to the Docker provider, since we identified that the `allowEmptyService` Swarm provider option should be enough to support the Sablier use case.
However, currently the `allowEmptyService` option is not working as expected (and as you mentioned it, using the `LBSwarm` option is the current workaround). We opened an issue for that: #12196

Thank you!

vordenken:
Can't believe it finally happend... Thank you all. I'm excited to test it out when it's available

nickshanks347:
> Can't believe it finally happend... Thank you all. I'm excited to test it out when it's available

Crazy right? A year and a half in the making

Timo-1979:
I've just tested to start a single container.

PR #10648, opened 2024-04-22
Use latest Ubuntu (LTS) image consistenly across GitHub workflow

jnoordsij:
### What does this PR do?

Use Ubuntu 22.04 consistenly across GitHub workflow

### Motivation

I noticed that all but these two jobs were running on the `ubuntu-22.04` image. For consistency and easier future updates, ~~I thought using the same image for all jobs makes more sense~~ all jobs were altered to use the `ubuntu-latest` (LTS) variant.

### More

- [ ] Added/updated tests

### Additional Notes

Requires https://github.com/traefik/traefik/pull/10649 to be merged first; then `v2.11` into `v3.0` merge, after which this can be merged.

### DIFF
--- a/.github/workflows/test-conformance.yaml
+++ b/.github/workflows/test-conformance.yaml
@@ -15,7 +15,7 @@ env:
 jobs:
 
   test-conformance:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     steps:
       - name: Check out code
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,7 +12,7 @@ env:
 jobs:
 
   build-webui:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-latest
 
     steps:
       - name: Check out code
@@ -35,7 +35,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ ubuntu-22.04, macos-latest, windows-latest ]
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
     needs:
       - build-webui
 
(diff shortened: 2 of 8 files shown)


.github/workflows/codeql.yml
```diff
@@ -11,7 +11,7 @@ on:
 jobs:
   analyze:
     name: Analyze
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
ldez, line 14:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/test-conformance.yaml
```diff
@@ -15,7 +15,7 @@ env:
 jobs:
 
   test-conformance:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
ldez, line 18:
```suggestion
    runs-on: ubuntu-latest
```

ldez (changes requested):
I think we don't need an explicit version.
Also `ubuntu-latest` simplify the maintenance.

jnoordsij:
Yeah fair, using `latest` everywhere also makes sense and is even less of a burden. Theoretically this might cause unprompted breakges when the release moves, but probably slim chances of that and it's not harmful.

Updated to use this everywhere instead.

ldez:
The OS variant name is used by GitHub checks, we will have a problem with `v2.11`, `v3.0`, and `master` branches if the checks don't have the same name.

Maybe the best approach is to change that inside `v2.11` and after we will backport that on the other branches.

jnoordsij:
Created https://github.com/traefik/traefik/pull/10649 to do this on `v2.11`. 

Given that `v2.11` does not have the `test-conformance` job, this PR is still required for `v3.0` after merging that and then merging `v2.11` into `v3.0` for applying the last change here.

ldez:
Closes in favor #10649.

We will add the missing parts during the backport.

PR #10660, opened 2024-04-24
Add support for Zstandard to the Compression middleware

### LINKED ISSUE #10557: zstandard compression
### Welcome!

Any plan in adding Support for zstd compression next to gzip and brotli now that Chrome is supporting it?

https://github.com/traefik/traefik/issues/4202
https://chromestatus.com/feature/6186023867908096

Belphemur:
### What does this PR do?

This PR will add support for Zstandard compression when browser state they support it.
It will take precedence on any other compression algorithm like brotli and gzip.

I also refactored the code of the Brotli compression to have only one compression handler for Brotli and Zstandard to simplify the maintenance and reduce the change of new bug being introduced.

### Motivation

Request was made for the feature, and I'm sure this will make traefik one of the first server supporting the zstd compression algorithm.

Fixes #10557 which has the tag `contributor/wanted`.

### More

- [x] Added/updated tests


pkg/middlewares/compress/compresshandler/compression_wrapper.go
```diff
@@ -0,0 +1,106 @@
...
+	return z.brotli.Flush()
+}
+
+func (z *brotliWriter) Write(p []byte) (n int, err error) {
+	return z.brotli.Write(p)
+}
+
+func (z *brotliWriter) ContentEncoding() string {
+	return string(Brotli)
+}
```
ldez, line 106:
```suggestion
package compresshandler

import (
	"fmt"
	"io"

	"github.com/andybalholm/brotli"
	"github.com/klauspost/compress/zstd"
)

const (
	Brotli    = "br"
	Zstandard = "zstd"
)

type compression interface {
	// Write data to the encoder.
	// Input data will be buffered and as the buffer fills up
	// content will be compressed and written to the output.
	// When done writing, use Close to flush the remaining output
	// and write CRC if requested.
	Write(p []byte) (n int, err error)
	// Flush will send the currently written data to output
	// and block until everything has been written.
	// This should only be used on rare occasions where pushing the currently queued data is critical.
	Flush() error
	// Close closes the underlying writers if/when appropriate.
	// Note that the compressed writer should not be closed if we never used it,
	// as it would otherwise send some extra "end of compression" bytes.
	// Close also makes sure to flush whatever was left to write from the buffer.
	Close() error
}

type CompressionWriter struct {
	compression
	alg string
}

func NewCompressionWriter(algo string, in io.Writer) (*CompressionWriter, error) {
	switch algo {
	case Brotli:
		return &CompressionWriter{compression: brotli.NewWriter(in), alg: algo}, nil

	case Zstandard:
		writer, err := zstd.NewWriter(in)
		if err != nil {
			return nil, err
		}

		return &CompressionWriter{compression: writer}, nil

	default:
		return nil, fmt.Errorf("unknown compression algo: %s", algo)
	}
}

func (c *CompressionWriter) ContentEncoding() string {
	return c.alg
}
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -30,10 +28,16 @@ type Config struct {
 	IncludedContentTypes []string
 	// MinSize is the minimum size (in bytes) required to enable compression.
 	MinSize int
+	// Algorithmn used for the compression (currently Brotli and Zstandard)
+	Algorithm Algorithm
```
ldez, line 32:
```suggestion
	// Algorithm used for the compression (currently Brotli and Zstandard)
	Algorithm string
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -30,10 +28,16 @@ type Config struct {
...
 }
 
 // NewWrapper returns a new Brotli compressing wrapper.
 func NewWrapper(cfg Config) (func(http.Handler) http.HandlerFunc, error) {
+	if cfg.Algorithm == Undefined {
+		return nil, errors.New("need an algorithm of compression defined")
+	}
```
ldez, line 39:
```suggestion
	if cfg.Algorithm == "" {
		return nil, errors.New("compression algorithm undefined")
	}
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -66,26 +70,28 @@ func NewWrapper(cfg Config) (func(http.Handler) http.HandlerFunc, error) {
 		return func(rw http.ResponseWriter, r *http.Request) {
 			rw.Header().Add(vary, acceptEncoding)
 
-			brw := &responseWriter{
+			compressionWriter, err := NewCompressionWriter(cfg.Algorithm, rw)
+			responseWriter := &responseWriter{
 				rw:                   rw,
-				bw:                   brotli.NewWriter(rw),
+				compressionw:         compressionWriter,
```
ldez, line 76:
```suggestion
				compressionWriter:    compressionWriter,
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -66,26 +70,28 @@ func NewWrapper(cfg Config) (func(http.Handler) http.HandlerFunc, error) {
...
 	}, nil
 }
 
 // TODO: check whether we want to implement content-type sniffing (as gzip does)
 // TODO: check whether we should support Accept-Ranges (as gzip does, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Ranges)
 type responseWriter struct {
-	rw http.ResponseWriter
-	bw *brotli.Writer
+	rw           http.ResponseWriter
+	compressionw CompressionWriter
```
ldez, line 94:
```suggestion
	compressionWriter CompressionWriter
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -133,7 +144,7 @@ func (r *responseWriter) Write(p []byte) (int, error) {
 	// We are now in compression cruise mode until the end of times.
 	if r.compressionStarted {
 		// If compressionStarted we assume we have sent headers already
-		return r.bw.Write(p)
+		return r.compressionw.Write(p)
```
ldez, line 147:
```suggestion
		return r.compressionWriter.Write(p)
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -185,13 +196,13 @@ func (r *responseWriter) Write(p []byte) (int, error) {
 	// Since we know we are going to compress we will never be able to know the actual length.
 	r.rw.Header().Del(contentLength)
 
-	r.rw.Header().Set(contentEncoding, "br")
+	r.rw.Header().Set(contentEncoding, r.compressionw.ContentEncoding())
```
ldez, line 199:
```suggestion
	r.rw.Header().Set(contentEncoding, r.compressionWriter.ContentEncoding())
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -185,13 +196,13 @@ func (r *responseWriter) Write(p []byte) (int, error) {
...
 	r.rw.WriteHeader(r.statusCode)
 	r.headersSent = true
 
 	// Start with sending what we have previously buffered, before actually writing
 	// the bytes in argument.
-	n, err := r.bw.Write(r.buf)
+	n, err := r.compressionw.Write(r.buf)
```
ldez, line 205:
```suggestion
	n, err := r.compressionWriter.Write(r.buf)
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -210,7 +221,7 @@ func (r *responseWriter) Write(p []byte) (int, error) {
 	r.buf = r.buf[:0]
 
 	// Now that we emptied the buffer, we can actually write the given bytes.
-	return r.bw.Write(p)
+	return r.compressionw.Write(p)
```
ldez, line 224:
```suggestion
	return r.compressionWriter.Write(p)
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -248,15 +259,15 @@ func (r *responseWriter) Flush() {
 	// we have to do it ourselves.
 	defer func() {
 		// because we also ignore the error returned by Write anyway
-		_ = r.bw.Flush()
+		_ = r.compressionw.Flush()
```
ldez, line 262:
```suggestion
		_ = r.compressionWriter.Flush()
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -248,15 +259,15 @@ func (r *responseWriter) Flush() {
...
 
 		if rw, ok := r.rw.(http.Flusher); ok {
 			rw.Flush()
 		}
 	}()
 
 	// We empty whatever is left of the buffer that Write never took care of.
-	n, err := r.bw.Write(r.buf)
+	n, err := r.compressionw.Write(r.buf)
```
ldez, line 270:
```suggestion
	n, err := r.compressionWriter.Write(r.buf)
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -311,7 +322,7 @@ func (r *responseWriter) close() error {
 
 	if len(r.buf) == 0 {
 		// If we got here we know compression has started, so we can safely flush on bw.
-		return r.bw.Close()
+		return r.compressionw.Close()
```
ldez, line 325:
```suggestion
		return r.compressionWriter.Close()
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -329,16 +340,16 @@ func (r *responseWriter) close() error {
 
 	// There is still data in the buffer, simply because Write did not take care of it all.
 	// We flush it to the compressed writer.
-	n, err := r.bw.Write(r.buf)
+	n, err := r.compressionw.Write(r.buf)
```
ldez, line 343:
```suggestion
	n, err := r.compressionWriter.Write(r.buf)
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -329,16 +340,16 @@ func (r *responseWriter) close() error {
...
+		r.compressionw.Close()
 		return err
 	}
 	if n < len(r.buf) {
-		r.bw.Close()
+		r.compressionw.Close()
 		return io.ErrShortWrite
 	}
-	return r.bw.Close()
+	return r.compressionw.Close()
```
ldez, line 352:
```suggestion
	return r.compressionWriter.Close()
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -329,16 +340,16 @@ func (r *responseWriter) close() error {
...
-	n, err := r.bw.Write(r.buf)
+	n, err := r.compressionw.Write(r.buf)
 	if err != nil {
-		r.bw.Close()
+		r.compressionw.Close()
 		return err
 	}
 	if n < len(r.buf) {
-		r.bw.Close()
+		r.compressionw.Close()
```
ldez, line 349:
```suggestion
		r.compressionWriter.Close()
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -329,16 +340,16 @@ func (r *responseWriter) close() error {
 
 	// There is still data in the buffer, simply because Write did not take care of it all.
 	// We flush it to the compressed writer.
-	n, err := r.bw.Write(r.buf)
+	n, err := r.compressionw.Write(r.buf)
 	if err != nil {
-		r.bw.Close()
+		r.compressionw.Close()
```
ldez, line 345:
```suggestion
		r.compressionWriter.Close()
```

pkg/middlewares/compress/compresshandler/compression_handler.go
```diff
@@ -66,26 +70,28 @@ func NewWrapper(cfg Config) (func(http.Handler) http.HandlerFunc, error) {
 		return func(rw http.ResponseWriter, r *http.Request) {
 			rw.Header().Add(vary, acceptEncoding)
 
-			brw := &responseWriter{
+			compressionWriter, err := NewCompressionWriter(cfg.Algorithm, rw)
```
ldez, line 73:
```suggestion
			compressionWriter, err := NewCompressionWriter(cfg.Algorithm, rw)
			if err != nil {
				logMessage := fmt.Sprintf("create compression handler: %v", err)
				logger.Debug().Msg(logMessage)
				observability.SetStatusErrorf(req.Context(), logMessage)
				rw.WriteHeader(http.StatusInternalServerError)
				return
			}
```

ldez (changes requested):
The package `compresshandler` should be merged with `compress`.

My suggestions will produce a non-working code because more changes are required (like the remove of the package `compresshandler`)

Also `compression_handler.go` and `compression_wrapper.go` should be merged in one file `compression_handler.go`.

pkg/middlewares/compress/compresshandler/zstd_test.go
```diff
@@ -0,0 +1,864 @@
...
+	req.Header.Set(acceptEncoding, "zstd")
+
+	rw := httptest.NewRecorder()
+	h.ServeHTTP(rw, req)
+
+	// With less than 1024 bytes the response should not be compressed.
+	assert.Equal(t, http.StatusAccepted, rw.Code)
+	assert.Empty(t, rw.Header().Get(contentEncoding))
+	assert.Equal(t, smallTestBody, rw.Body.Bytes())
+}
```
ldez, line 47:
```suggestion

func Test_Vary(t *testing.T) {
	testCases := []struct {
		desc           string
		h              http.Handler
		acceptEncoding string
	}{
		{
			desc:           "brotli",
			h:              newTestBrotliHandler(t, smallTestBody),
			acceptEncoding: "br",
		},
		{
			desc:           "zstd",
			h:              newTestZStandardHandler(t, smallTestBody),
			acceptEncoding: "zstd",
		},
	}

	for _, test := range testCases {
		test := test
		t.Run(test.desc, func(t *testing.T) {
			t.Parallel()

			req, _ := http.NewRequest(http.MethodGet, "/whatever", nil)
			req.Header.Set(acceptEncoding, test.acceptEncoding)

			rw := httptest.NewRecorder()
			test.h.ServeHTTP(rw, req)

			assert.Equal(t, http.StatusAccepted, rw.Code)
			assert.Equal(t, acceptEncoding, rw.Header().Get(vary))
		})
	}
}

func Test_SmallBodyNoCompression(t *testing.T) {
	testCases := []struct {
		desc           string
		h              http.Handler
		acceptEncoding string
	}{

		{
			desc:           "brotli",
			h:              newTestBrotliHandler(t, smallTestBody),
			acceptEncoding: "br",
		},
		{
			desc:           "zstd",
			h:              newTestZStandardHandler(t, smallTestBody),
			acceptEncoding: "zstd",
		},
	}

	for _, test := range testCases {
		test := test
		t.Run(test.desc, func(t *testing.T) {
			t.Parallel()
			h := newTestBrotliHandler(t, smallTestBody)

			req, _ := http.NewRequest(http.MethodGet, "/whatever", nil)
			req.Header.Set(acceptEncoding, test.acceptEncoding)

			rw := httptest.NewRecorder()
			h.ServeHTTP(rw, req)

			// With less than 1024 bytes the response should not be compressed.
			assert.Equal(t, http.StatusAccepted, rw.Code)
			assert.Empty(t, rw.Header().Get(contentEncoding))
			assert.Equal(t, smallTestBody, rw.Body.Bytes())
		})
	}
}
```

I will add a suggestion for each test but all the tests can written on this base.
`brotli_test.go` and `zstd_test` can be "merged" inside one file based on that.

ldez (changes requested):

pkg/middlewares/compress/compresshandler/compression_wrapper.go
Belphemur: That does simplify the logic.

I'll incorporate those changes in the next commits.

pkg/middlewares/compress/compresshandler/zstd_test.go
Belphemur: I thought the same, I'll update the main test suite to be iterative following your suggestion.

Belphemur:
@ldez Ready for second round of review.

For the test suite, I don't think all test case beneficiate from iterating through all the compression algo especially when they are testing that the compression isn't used.

Other than that, I've merge together the handler and wrapper, used your suggested logic for it (way cleaner, thank you for it).

ldez:
The package `compresshandler` should be merged with `compress`.

https://github.com/traefik/traefik/pull/10660#pullrequestreview-2021128142

pkg/middlewares/compress/compress_test.go
```diff
@@ -24,6 +24,7 @@ const (
 	varyHeader            = "Vary"
 	gzipValue             = "gzip"
 	brotliValue           = "br"
+	zstdValue             = "zstd"
```
ldez, line 27:
```suggestion
```

pkg/middlewares/compress/compression_handler.go
```diff
@@ -30,10 +35,69 @@ type Config struct {
...
+	// Flush will send the currently written data to output
+	// and block until everything has been written.
+	// This should only be used on rare occasions where pushing the currently queued data is critical.
+	Flush() error
+	// Close closes the underlying writers if/when appropriate.
+	// Note that the compressed writer should not be closed if we never used it,
+	// as it would otherwise send some extra "end of compression" bytes.
+	// Close also makes sure to flush whatever was left to write from the buffer.
+	Close() error
+}
```
ldez, line 65:
```suggestion
```

pkg/middlewares/compress/compression_handler.go
```diff
@@ -10,8 +10,13 @@ import (
 	"net/http"
 
 	"github.com/andybalholm/brotli"
+	"github.com/klauspost/compress/zstd"
+	"github.com/traefik/traefik/v3/pkg/middlewares"
+	"github.com/traefik/traefik/v3/pkg/middlewares/observability"
 )
 
+const typeNameHandler = "CompressHandler"
```
ldez, line 18:
```suggestion
const (
	Brotli    = "br"
	Zstandard = "zstd"
)

type compression interface {
	// Write data to the encoder.
	// Input data will be buffered and as the buffer fills up
	// content will be compressed and written to the output.
	// When done writing, use Close to flush the remaining output
	// and write CRC if requested.
	Write(p []byte) (n int, err error)
	// Flush will send the currently written data to output
	// and block until everything has been written.
	// This should only be used on rare occasions where pushing the currently queued data is critical.
	Flush() error
	// Close closes the underlying writers if/when appropriate.
	// Note that the compressed writer should not be closed if we never used it,
	// as it would otherwise send some extra "end of compression" bytes.
	// Close also makes sure to flush whatever was left to write from the buffer.
	Close() error
}
```

pkg/middlewares/compress/compression_handler_test.go
```diff
@@ -19,44 +21,110 @@ var (
...
+		},
+		{
+			desc:           "zstd",
+			h:              newTestZstandardHandler(t, smallTestBody),
+			acceptEncoding: "zstd",
+		},
+	}
+
+	for _, test := range testCases {
+		test := test
```
ldez, line 42:
```suggestion
```

pkg/middlewares/compress/compression_handler_test.go
```diff
@@ -19,44 +21,110 @@ var (
...
+			desc:           "zstd",
+			h:              newTestZstandardHandler(t, smallTestBody),
+			acceptEncoding: "zstd",
+		},
+	}
 
-	req, _ := http.NewRequest(http.MethodGet, "/whatever", nil)
-	req.Header.Set(acceptEncoding, "br")
+	for _, test := range testCases {
+		test := test
```
ldez, line 77:
```suggestion
```

pkg/middlewares/compress/compression_handler.go
```diff
@@ -30,10 +35,69 @@ type Config struct {
...
+	compression
+	alg string
+}
+
+func NewCompressionWriter(algo string, in io.Writer) (*CompressionWriter, error) {
+	var writer compression = nil
+	switch algo {
+	case Brotli:
+		writer = brotli.NewWriter(in)
+		break
```
ldez, line 76:
```suggestion

```

pkg/middlewares/compress/compression_handler.go
```diff
@@ -30,10 +35,69 @@ type Config struct {
...
+	case Zstandard:
+		zstdWriter, err := zstd.NewWriter(in)
+		if err != nil {
+			return nil, err
+		}
+		writer = zstdWriter
+		break
```
ldez, line 83:
```suggestion

```

pkg/middlewares/compress/compression_handler_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"testing"
 
 	"github.com/andybalholm/brotli"
+
```
ldez, line 12:
```suggestion
```

ldez (changes requested):

pkg/middlewares/compress/compression_handler.go
```diff
@@ -30,10 +35,69 @@ type Config struct {
...
+	default:
+		return nil, fmt.Errorf("unknown compression algo: %s", algo)
+	}
+
+	return &CompressionWriter{compression: writer, alg: algo}, nil
+}
```
ldez, line 89:
```suggestion
func NewCompressionWriter(algo string, in io.Writer) (*CompressionWriter, error) {
	switch algo {
	case Brotli:
		return &CompressionWriter{compression: brotli.NewWriter(in), alg: algo}, nil

	case Zstandard:
		writer, err := zstd.NewWriter(in)
		if err != nil {
			return nil, err
		}

		return &CompressionWriter{compression: writer, alg: algo}, nil
	default:
		return nil, fmt.Errorf("unknown compression algo: %s", algo)
	}
}
```

Belphemur: I disagree with this suggestion. This was the previous version that was suggested, however it's easy to forget to to set the `algo` in the `CompressionWriter` (which was the case in the suggestion).

The version of the code I made account for it, since it won't be the responsibility of the developer adding a new algorithm to the CompressionHandler to not forget to set it.

Also by the time the code reach the creation of the `CompressionWriter` the `algo` has been checked to be valid.

pkg/middlewares/compress/compression_handler.go
```diff
@@ -66,26 +130,35 @@ func NewWrapper(cfg Config) (func(http.Handler) http.HandlerFunc, error) {
 		return func(rw http.ResponseWriter, r *http.Request) {
 			rw.Header().Add(vary, acceptEncoding)
 
-			brw := &responseWriter{
+			compressionWriter, err := NewCompressionWriter(cfg.Algorithm, rw)
+			if err != nil {
+				logger := middlewares.GetLogger(r.Context(), cfg.MiddlewareName, typeNameHandler)
```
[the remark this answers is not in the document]
Belphemur: Right, misread the documentation on the middleware method.
Correcting this.

pkg/middlewares/compress/compression_handler.go
ldez: I don't share your opinion, Go is not C, we return already.

pkg/middlewares/compress/compression_handler.go
ldez: both

pkg/middlewares/compress/compression_handler.go
Belphemur: Good point about the `nil` fixing that.

To be honest, I'm trying to have a match pattern like c#: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#compare-discrete-values 

One part of the code focused on building the underlying `writer` the other part of the code simply using it to wrap it with only one place setting the `algo`.

Also i'm not so sure what you mean by
> Also break is useless in a switch in Go.
From the go rfc: https://go.dev/ref/spec#Break_statements
> A "break" statement terminates execution of the innermost ["for"](https://go.dev/ref/spec#For_statements), ["switch"](https://go.dev/ref/spec#Switch_statements), or ["select"](https://go.dev/ref/spec#Select_statements) statement within the same function.

Why would it be useless in the case of a `switch` ?

ldez: > Also i'm not so sure what you mean by

a `case` always breaks, you don't need to add a `break` explicitly.

https://go.dev/ref/spec#Switch_statements

pkg/middlewares/compress/compression_handler.go
Belphemur: I moved the interface and remove the `typeNameHandler`

pkg/middlewares/compress/compression_handler.go
ldez: > To be honest, I'm trying to have a match pattern like c#

It's not a way to do it in Go, because Go doesn't follow the old rule inherited from C (it was related to memory management), could you apply my suggestion?

ldez:
I will add an extra commit to your PR, related to a refactor, but not directly related to the code of your PR.

Belphemur:
@ldez It seems GitHub still state that the PR needs some changes ?

I imagine since you have added the refactor of CompressHandler, that this is just forgetting to approve the PR ?

ldez (approved):
LGTM

pkg/middlewares/compress/compress.go
```diff
@@ -76,7 +76,13 @@ func New(ctx context.Context, next http.Handler, conf dynamic.Compress, name str
...
 
 	var err error
-	c.brotliHandler, err = c.newBrotliHandler()
+
+	c.zstdHandler, err = c.newCompressionHandler(Zstandard, name)
+	if err != nil {
+		return nil, err
+	}
+
+	c.brotliHandler, err = c.newCompressionHandler(Brotli, name)
```
rtribotte, line 85: Hi, I'm wondering why we should prioritise zstd over other encodings?
I understand there's no "good choice" since we don't parse quality values, but it seems for the same reason we could give zstd the least priority and avoid breaking/altering compress middleware behaviour. WDYT?

Belphemur:
@ldez I'd like to confirm your latest change.
By making the br > zstd > gzip, this will basically make `zstd` never used. Since every browser that support `zstd` do also support `br`.

Unless the browser decides to drop `br` which I don't think there is any plan for.

My understanding is `zstd` had better performance, compression and decompression than `br` hence why it made sense to give it priority.

ldez:
The compression depends on the `Accept-Encoding` header, I create another PR to support the weight. See #10777.

The weight allows the client to specify the compression type "priority/quality".

> By making the br > zstd > gzip, 

The default order is br > gzip > zstd, for compatibility.

The browser's usage is one of the possible use cases.

PR #10661, opened 2024-04-25
Add support to run as native windows service

fredwangwang:
### What does this PR do?

Allow traefik to run as native windows service without any service shim, like NSSM

### Motivation

This addresses the issue https://github.com/traefik/traefik/issues/10659, allowing better integration with windows systems without any third party components.

image
The screenshot shows that with the code change, traefik can be managed directly with windows service manager

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -46,6 +46,7 @@ import (
 	"github.com/traefik/traefik/v3/pkg/tracing"
 	"github.com/traefik/traefik/v3/pkg/types"
 	"github.com/traefik/traefik/v3/pkg/version"
+	"github.com/traefik/traefik/v3/pkg/winsvc"
 )
 
 func main() {
@@ -119,6 +120,11 @@ func runCmd(staticConfiguration *static.Configuration) error {
 	}
 
 	ctx, _ := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	ctx, svcCancel := context.WithCancel(ctx)
+	go func() {
+		<-winsvc.ChanExit
+		svcCancel()
+	}()
 
 	if staticConfiguration.Ping != nil {
 		staticConfiguration.Ping.WithContext(ctx)
--- a/pkg/winsvc/service_windows.go
+++ b/pkg/winsvc/service_windows.go
@@ -1,0 +1,37 @@
+//go:build windows
+// +build windows
+
+package winsvc
+
+import (
+	wsvc "golang.org/x/sys/windows/svc"
+)
+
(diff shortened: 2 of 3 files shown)


kevinpollet:
Hello @fredwangwang and thanks for the contribution,

As this is an enhancement, please could you rebase it on the master branch?

fredwangwang:
Hi @kevinpollet retargeted to master

kevinpollet:
Hello @nausicaaconsulting and sorry for the late answer, no ETA at the moment, but this pull request is on our radar.

cmd/traefik/traefik.go
```diff
@@ -119,6 +120,11 @@ func runCmd(staticConfiguration *static.Configuration) error {
 	}
 
 	ctx, _ := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	ctx, svcCancel := context.WithCancel(ctx)
+	go func() {
+		<-winsvc.ChanExit
```
kevinpollet, line 125: At first glance on other platforms, this goroutine will never end, as the channel will never receive a message. I am wondering if this mechanism could be encapsulated to only deal with ctx in the traefik cmd.

pkg/winsvc/service_windows.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	wsvc "golang.org/x/sys/windows/svc"
+)
+
+type serviceWindows struct{}
+
+func init() {
+	isService, err := wsvc.IsWindowsService()
+	if err != nil {
+		panic(err)
```
kevinpollet, line 15: If I am not mistaken, if there is an error everything will crash, right? Maybe we should ignore the error, wdyt?

pkg/winsvc/service_windows.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	if !isService {
+		return
+	}
+	go wsvc.Run("", serviceWindows{})
```
kevinpollet, line 20: Why the name is empty here?

kevinpollet (changes requested):

kevinpollet:
Hello @fredwangwang,

Are you still working on the subject?
What do you think of my previous comments?

i4j5:
error 1053

pkg/winsvc/service_windows.go
ainglese: the name passed to svc.Run is almost irrelevant, because the real service name is determined at service install time, so this would be used as a fallback, if used at all.

emilevauge:
Due to the lack of activity on this PR, we closing it.
Don't hesitate to ping us if anything changes on your end.

Issue #10638, opened 2024-04-19
Switch to Kubernetes EndpointSlices API


jnoordsij:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

In my search of figuring out how exactly Service/Pod traffic is distributed, I noticed that Traefik internally uses the Kubernetes [Endpoints API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.30/#endpoints-v1-core). As of Kubernetes 1.21, the [EndpointSlices API](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) is a stable alternative described as:
> Kubernetes' EndpointSlice API provides a way to track network endpoints within a Kubernetes cluster. EndpointSlices offer a more scalable and extensible alternative to [Endpoints](https://kubernetes.io/docs/concepts/services-networking/service/#endpoints).

Both from a performance and future-proof point of view, it might therefore be benificial to switch towards using this API.

A possible source of inspiration may be https://github.com/kubernetes/ingress-nginx/pull/8890, which implemented this for (Kubernetes) NGINX.

jnoordsij:
A WIP attempt can be found at https://github.com/jnoordsij/traefik/tree/migrate-to-endpointslices. This is far from finished, not passing tests yet, but seems to be heading in the right direction. Also this should probably rebased to `master`.

kevinpollet:
Hello @jnoordsij,

Thanks for opening the issue, 

When you have something ready feel free to open a pull request.
As this is an enhancement, the pull request has to be opened on the master branch.

jnoordsij:
This issue has been solved with #10664.

PR #10664, opened 2024-04-28
Migrate to EndpointSlices API 

jnoordsij:
### What does this PR do?

Switch to Kubernetes EndpointSlices API as described in #10638.

### Motivation

Discovered existence of this API while looking into behavior regarding non-ready service endpoints still being served traffic. Although this change will probably make little changes in that regard, it's probably still a technical improvement that allows for more features in the future and still might slightly improve performance.

### More

- [x] Added/updated tests

### To Do

- [ ] Create a list of important changes
- [ ] Update documentation; possibly with some draft for upgrade notes
- [ ] Check for additionally useful tests

### DIFF
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -644,20 +644,20 @@ func (p *Provider) loadService(client Client, namespace string, backend netv1.In
 		return svc, nil
 	}
 
-	endpoints, endpointsExists, endpointsErr := client.GetEndpoints(namespace, backend.Service.Name)
-	if endpointsErr != nil {
-		return nil, endpointsErr
+	endpointSlices, endpointSlicesExists, endpointSlicesErr := client.GetEndpointSlices(namespace, backend.Service.Name)
+	if endpointSlicesErr != nil {
+		return nil, endpointSlicesErr
 	}
 
-	if !endpointsExists {
-		return nil, errors.New("endpoints not found")
+	if !endpointSlicesExists {
+		return nil, errors.New("endpointslices not found")
 	}
 
-	for _, subset := range endpoints.Subsets {
+	for _, endpointSlice := range endpointSlices {
 		var port int32
-		for _, p := range subset.Ports {
-			if portName == p.Name {
-				port = p.Port
+		for _, p := range endpointSlice.Ports {
+			if portName == *p.Name {
+				port = *p.Port
 				break
 			}
 		}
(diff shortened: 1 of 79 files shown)


jnoordsij:
Any early comments or suggestions are welcome, but given some tests are still failing and I'm aiming to write down a more descriptive summary and list of important changes, a thorough review is probably something to only do at a later point. 

I'll hope to continue working on this soon. Yet like I said, any comments, concerns or suggestions are welcome in the meantime!

nmengin:
Hello @jnoordsij,

Thank you for the answer.
We are waiting for your go before starting the review 😄

jnoordsij:
@nmengin I think this is all ready for a first review right now!

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -668,12 +669,21 @@ func (p *Provider) loadService(client Client, namespace string, backend netv1.In
 
 		protocol := getProtocol(portSpec, portName, svcConfig)
 
-		for _, addr := range subset.Addresses {
-			hostPort := net.JoinHostPort(addr.IP, strconv.Itoa(int(port)))
+		for _, endpoint := range endpointSlice.Endpoints {
+			if !(*endpoint.Conditions.Ready) {
+				continue
+			}
```
rtribotte, line 675: Hello @jnoordsij!

Thanks for this contribution!

I'm wondering if filtering the endpoint condition here is introducing a breaking change in terms of discovery, WDYT?

jnoordsij: I did think at first when starting with this there might be a difference, but from what I've gathered it should not change anything. The endpoints with `ready=true` state on the `EndpointSlice` API should match the `addresses` on the `Endpoints` API (which also has a subsequent `notReadyAddresses` entry) and thus should return the exact same set of endpoints as before.

See also the [following note from documentation](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#serving):
> Although serving is almost identical to ready, it was added to prevent breaking the existing meaning of ready. It may be unexpected for existing clients if ready could be true for terminating endpoints, since historically terminating endpoints were never included in the Endpoints or EndpointSlice API to begin with. For this reason, ready is always false for terminating endpoints, and a new condition serving was added in v1.20 so that clients can track readiness for terminating pods independent of the existing semantics for ready.

Basically the new API has more information about state: rather than only doing a 'binary' separation between `ready` and not `ready`, there are now three conditions you can check on depending on your preference: `ready`, `serving` and `terminating`.

rtribotte: Indeed, my first assumption was that states were not coexisting but rather a value for the state of the endpoint. Thanks for putting some thoughts and providing clarification on this 👍

kevinpollet:
Hello @jnoordsij and thanks for your contribution,

To support Gateway API v1.1.0, the use of the EndpointSlices API is mandatory.

Therefore, we would like to move quickly on this pull request. Can we rebase and push some changes to it?

jnoordsij:
@kevinpollet sure thing! Feel free to further alter it where needed and if you have any questions regarding the work feel free to ask.

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks 👍

docs/content/migration/v3.md
```diff
@@ -0,0 +1,27 @@
...
+description: "Learn the steps needed to migrate to new Traefik Proxy v3 versions. Read the technical documentation."
+---
+
+# Migration: Steps needed between the versions
+
+## v3.0 to v3.1
+
+### Kubernetes Provider RBACs
+
+Starting with v3.1, the Kubernetes Providers now use the [EndpointSlices API](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) to discover service endpoint addresses.
```
rtribotte, line 12:
```suggestion
Starting with v3.1, the Kubernetes Providers now use the [EndpointSlices API](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) (Kubernetes >=v1.21) to discover service endpoint addresses.
```

mmatur (approved):

PR #10667, opened 2024-04-29
Set Gateway HTTPRoute status

kevinpollet:
### What does this PR do?

This pull request sets the Gateway HTTPRoute status when the route is correctly configured. The status set when a route is misconfigured will be addressed in another pull request.

### Motivation

To enhance the Gateway API support.

### More

- [ ] Added/updated tests

### Additional Notes


nmengin (approved):
LGTM 👍

rtribotte (approved):

mmatur (approved):
LGTM

Gitopolis:
Thank you for the change, but status now is just an empty list of parents, which is not exactly what I expected. The route works fine though, but external-dns is confused.
Screenshot 2024-06-18 at 1 18 46 PM

kevinpollet:
Hello @Gitopolis,

This change is not released and will be part of the upcoming minor version of Traefik (v3.1). Recently, we made some changes to the KubernetesGateway provider to be more compliant with the Gateway API specification.

To test them, you can use the experimental Docker image `traefik/traefik:experimental-master`.
If you still think that there is an issue, feel free to open an issue with a reproducible use case if possible.

ilmax:
Hello @kevinpollet , I'm running traefik 3.1 but external-dns still doesn't work well with HttpRoute. After installing the traefik helm chart and enabling the gateway, I still ended up with no Addresses populated hence external-dns doesn't generate any DNS records. [There's a workaround](https://github.com/kubernetes-sigs/external-dns/blob/master/source/gateway.go#L350) using the external-dns's target annotation but I would like to avoid duplicating the IP/hostname on every HTTPRoute. 
What's the proper way to end up with a populated `Gateway.Status.Addresses` that's needed [here](https://github.com/kubernetes-sigs/external-dns/blob/master/source/gateway.go#L353)?

kevinpollet:
Hello @ilmax and thanks for your interest in Traefik,

Have you tried to you use the [`statusAddress`](https://doc.traefik.io/traefik/providers/kubernetes-gateway/#statusaddress) option?

This allows to set a fixed IP, Hostname to the Gateway status address. It is also possible to define the service used to expose Traefik to replicate the service status addresses to the Gateway status addresses.

ilmax:
Thank you for getting back to me @kevinpollet but I may be missing something. 

>It is also possible to define the service used to expose Traefik to replicate the service status addresses to the Gateway status addresses.

This is exactly what I want to achieve.

What's the `static configuration` the [page mentions](https://doc.traefik.io/traefik/providers/kubernetes-gateway/#requirements)? (sentence on point 3)

 I was under the impression that it could be set using the helm values file but that has no effect and doing a quick search in the helm chart repository for the `statusAddress` only yields a comment result. Looking at the Gateway resource definition wasn't helpful either.

Can you tell me where am I supposed to configure the

```
providers:
  kubernetesGateway yml
```

as per the documentation?
Thank you

kevinpollet:
Hey @ilmax,

> What's the static configuration the [page mentions](https://doc.traefik.io/traefik/providers/kubernetes-gateway/#requirements)? (sentence on point 3)

This is the configuration used to configure the Traefik instance. If you are using Helm maybe this is a missing feature, feel free to open a feature request in the Helm chart [repository](https://github.com/traefik/traefik-helm-chart).

In the meantime, you should be able to use it by adding the additional arguments [option](https://github.com/traefik/traefik-helm-chart) to your values.yaml, something like the following (use the CLI flag matching your use case):

```yaml
additionalArguments:
  - "--providers.kubernetesgateway.statusaddress.ip=1.2.3.4"
```

ilmax:
> This is the configuration used to configure the Traefik instance. If you are using Helm maybe this is a missing feature, feel free to open a feature request in the Helm chart [repository](https://github.com/traefik/traefik-helm-chart).

Will do! I think I found that (which is not straightforward to figure out, maybe it's worth a mention in the documentation) and it looks like it's being loaded:

values.yml
```yml
additionalArguments:
  - "--log.level=DEBUG"
  - "--providers.kubernetesgateway.statusaddress.service.name=traefik"
  - "--providers.kubernetesgateway.statusaddress.service.namespace=traefik"

deployment:
  enabled: true
  replicas: 2

ports:
  web:
    redirectTo:
      port: websecure
      priority: 10
  websecure:
    http3:
      enabled: true
    advertisedPort: 4443
    tls:
      enabled: true

ingressRoute:
  dashboard:
    enabled: false

providers:
  kubernetesCRD:
    enabled: true
  kubernetesIngress:
    enabled: false
  kubernetesGateway:
    enabled: true

rbac:
  enabled: true

service:
  enabled: true
  type: LoadBalancer

gateway:
  listeners:
    web:
      namespacePolicy: All
    websecure: null
```

```console
2024-08-02T14:47:53Z DBG github.com/traefik/traefik/v3/pkg/provider/aggregator/aggregator.go:203 > *gateway.Provider provider configuration config={"statusAddress":{"service":{"name":"traefik","namespace":"traefik"}}}
````

yet something is still not working as expected:

```
k get svc -n traefik traefik
NAME      TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)                                    AGE
traefik   LoadBalancer   10.43.120.242   192.168.2.210   80:31496/TCP,443:30285/TCP,443:30285/UDP   41m
❯ k get -n traefik gateway
NAME              CLASS     ADDRESS   PROGRAMMED   AGE
traefik-gateway   traefik             True         41m

kevinpollet:
@ilmax sorry there is a misunderstanding

> Will do! I think I found that (which is not straightforward to figure out, maybe it's worth a mention in the documentation) and it looks like it's being loaded

The missing feature in the Helm chart is to be able to define the `StatusAddress` through an option in the values.yaml. Using the `additionalArguments` is a way to specify static configuration options, when they are not yet available through an Helm option in the values.yaml.

> yet something is still not working as expected:

Are the flags added to the Traefik deployments args? Are there any ERROR logs?

To keep the repository focused, we ask that all questions be asked in the [community forum](https://community.traefik.io/). It is pretty active, so you might find that your question has already been answered there. If you think that there is a bug feel free to open a bug report in the TRaefik respository.

PR #10695, opened 2024-05-05
Bump docs Docker images to Alpine 3.19

jnoordsij:
### What does this PR do?

Update both Docker images for docs to use the latest Alpine version (3.19).
Given that [in Alpine 3.19 Python's package directory is marked as externally managed](https://www.alpinelinux.org/posts/Alpine-3.19.0-released.html), the docs image was updated to use a Python venv. The requirements file was updated accordingly, to no longer include unrelated system packages and bump a few (nested) dependencies. Dependency bumps were kept to a minimum, but updating them further might definitely be a good follow-up.

### Motivation

While trying to debug docs failure job for #10664, I noticed:
- the image versions were using different versions, not being the latest
- dependencies for `mkdocs` were not installable locally using `requirements.txt`

This PR should solve both issues (albeit still not explaining the job failure 😓). It should also make Python/pip package installation more consistent, given that system installed packages are no longer used.

### More

- [ ] Added/updated tests

### Additional Notes

Also threw in a typo fix I still had in my diffs because I figured a separate PR would be too much of a thing for this.


ldez:
The `requirements.txt` should contain the exhaustive list of dependencies.

jnoordsij:
> The `requirements.txt` should contain the exhaustive list of dependencies.
> 

It does still. I assume the previous version included system installed packages that came with installing pip and/or python3-dev; I've now generated it to have all direct dependencies listed on top and then `pip freeze`-ing after installing those.

jnoordsij:
Update: bumped to Alpine 3.20, which was released yesterday

mmatur (approved):
LGTM

kevinpollet (approved):
Thanks 👍

PR #10758, opened 2024-05-24
grafana: traefik-kubernetes: fix service name label_replace

xdavidwu:
The kubernetes (ingress) provider create services named `<namespace>-<service name>-<port>`, other kubernetes providers (crd, gateway) also name their services in similiar fashion, combining names from kubernetes with '-'. Most kubernetes resources requires their name to be usable as dns subdomain names only, therefore '-' is also possible in names (actually quite common, like kube-system, kube-controller-manager).

label_replace here before extracts first two part seperated with '-'. The intention here is probably to extract `<namespace>-<service name>`, but with '-' in namespaces or service names, this breaks and usually generates labels that are not unique. If there are query operation after label_replace, it is also fragile, due to having multiple series with same labels.

This fixes it by removing only `@<providername>` instead. '@' is invalid for names in most kubernetes resources, so this should be safe.

### What does this PR do?

Fix service name extraction on dashboard for traefik on kubernetes.

### More

- [ ] Added/updated tests

### DIFF
--- a/contrib/grafana/traefik-kubernetes.json
+++ b/contrib/grafana/traefik-kubernetes.json
@@ -507,7 +507,7 @@
             "uid": "${DS_PROMETHEUS}"
           },
           "editorMode": "code",
-          "expr": "topk(15,\n    label_replace(\n        traefik_service_request_duration_seconds_sum{service=~\"$service.*\",protocol=\"http\"} / \n          traefik_service_request_duration_seconds_count{service=~\"$service.*\",protocol=\"http\"},\n        \"service\", \"$1\", \"service\", \"([^-]+-[^-]+).*\")\n)\n\n",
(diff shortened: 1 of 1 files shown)


mmatur:
Hello @xdavidwu,

Thanks for your contribution.

Could you please rebase your PR on branch v2.11?

xdavidwu:
> Could you please rebase your PR on branch v2.11?

This issue only exists on branch v3.0 and master. Which one should I target?

mmatur:
Oops my bad, you can target `v3.0` please.

mmatur (approved):
LGTM

Thanks

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #10769, opened 2024-05-29
doc: Remove reference to deprecated helm chart

corneliusroemer:
The helm chart is deprecated and severely out of date. It should not be referenced by docs anymore as a result.


corneliusroemer:
Closing in favor of #10772

PR #10770, opened 2024-05-29
Add defaultEntryPoints option

niki-timofe:
### What does this PR do?

Added option `defaultEntryPoints` to providers that have `defaultRule`.

### Motivation

We use Traefik as both forward and reverse proxy in our clusters. Reverse routes we define manually by tags. In contrast, forward routes are generated automatically using the `defaultRule` option. 

Due to the large amount of forward routes and significantly fewer amount of reverse, we encountered the issue that our proxy is vulnerable to DDoS, which can be performed by sending unroutable requests: Traefik tries to match all (including forward) routes before respond with 404 code. Moreover, for safety reasons, forward routers should not be accessible outside of the clusters. The solution of these issues is in setting `entryPoints` option of forward routers, but in case where these routers are generated automatically, this is now not possible. So this option allow us to separate routers by entry points and solve both issues.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/providers/consul-catalog.md
+++ b/docs/content/providers/consul-catalog.md
@@ -501,7 +501,7 @@ providers:
 
 _Optional, Default=```Host(`{{ normalize .Name }}`)```_
 
-The default host rule for all services.
+The default rule for all services.
 
 For a given service, if no routing rule was defined by a tag, it is defined by this `defaultRule` instead.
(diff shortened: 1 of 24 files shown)


ldez:
Hello,

have you tried the option `asDefault`?

https://doc.traefik.io/traefik/routing/entrypoints/#asdefault

niki-timofe:
@ldez, `asDefault` looks like exactly what we need 🤯

Issue #10757, opened 2024-05-24
To support HTTP Basic authentication for docker/swarm provider's endpoint


ba1dr:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Traafic config example:
```
providers:
  swarm:
    endpoint: http://username:password@docker-socket-proxy:2375
```
This is a legit URL to the service, but Traefik does not parse this correctly, not taking username/password from the URL and trying to resolve everything as a single hostname.

```
May 24 09:08:36 utils01 dc7052128204[861]: 2024-05-24T09:08:36Z ERR Failed to retrieve information of the docker client and server host error="error during connect: Get \"http://username:password%40docker-socket-proxy:2375/v1.24/version\": dial tcp: lookup username:password@docker-socket-proxy: no such host" providerName=swarm
May 24 09:08:36 utils01 dc7052128204[861]: 2024-05-24T09:08:36Z ERR Provider error, retrying in 319.082513ms error="error during connect: Get \"http://username:password%40docker-socket-proxy:2375/v1.24/version\": dial tcp: lookup username:password@docker-socket-proxy: no such host" providerName=swarm
```

--

I am trying to replace `http` to `tcp`, the host's address is parsed correctly, but authentication is lost:
```
providers:
  swarm:
    endpoint: tcp://username:password@docker-socket-proxy:2375
```
Errors:
```
May 24 09:39:34 utils01 d353b15e7944[861]: 2024-05-24T09:39:34Z ERR Failed to retrieve information of the docker client and server host error="Error response from daemon: <html>\r\n<head><title>401 Authorization
 Required</title></head>\r\n<body>\r\n<center><h1>401 Authorization Required</h1></center>\r\n<hr><center>nginx/1.25.5</center>\r\n</body>\r\n</html>" providerName=swarm
May 24 09:39:34 utils01 d353b15e7944[861]: 2024-05-24T09:39:34Z ERR Provider error, retrying in 703.423221ms error="Error response from daemon: <html>\r\n<head><title>401 Authorization Required</title></head>\r\
n<body>\r\n<center><h1>401 Authorization Required</h1></center>\r\n<hr><center>nginx/1.25.5</center>\r\n</body>\r\n</html>" providerName=swarm
```

nmengin:
[...]

PR #10776, opened 2024-06-01
[Feature] Support Http basicAuth for docker and swarm

985492783:
### What does this PR do?

1. Add docker and swarm basicAuth

### Motivation

#10757 support Http basicAuth for docker and swarm

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -34,7 +34,7 @@ pyparsing==2.4.7
 python-dateutil==2.8.2
 PyYAML==6.0.1
 pyyaml-env-tag==0.1
-requests==2.25.1
+requests==2.28.2
 retrying==1.3.3
 six==1.15.0
 toml==0.10.2
--- a/docs/content/providers/docker.md
+++ b/docs/content/providers/docker.md
@@ -132,6 +132,7 @@ You can specify which Docker API Endpoint to use with the directive [`endpoint`]
         - Accounting at container level, by exposing the socket on a another container than Traefik's.
         - Accounting at kernel level, by enforcing kernel calls with mechanisms like [SELinux](https://en.wikipedia.org/wiki/Security-Enhanced_Linux), to only allows an identified set of actions for Traefik's process (or the "socket exposer" process).
         - SSH public key authentication (SSH is supported with Docker > 18.09)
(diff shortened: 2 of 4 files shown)


nmengin:
Hello @985492783,

Thank you for this contribution.
We've set the `design-review` label to check is we agree with the implementation before reviewing it.

We'll update the label as soon as we check it.

jspdown (dismissed):
Thanks @985492783 for opening this PR.

I think that what you've done is the right approach to providing BasicAuth on this provider. 

Initially I thought that maybe we could directly put the username/password in the `endpoint` field. But after more consideration I think it would bring some confusion, especially for the `tcp` scheme.

Although, to simplify the use of this feature, I think it would be best to have 2 configuration fields instead of `basicAuth`:
- `username`
- `password`

985492783:
> Thanks @985492783 for opening this PR.
> 
> I think that what you've done is the right approach to providing BasicAuth on this provider.
> 
> Initially I thought that maybe we could directly put the username/password in the `endpoint` field. But after more consideration I think it would bring some confusion, especially for the `tcp` scheme.
> 
> Although, to simplify the use of this feature, I think it would be best to have 2 configuration fields instead of `basicAuth`:
> 
> * `username`
> * `password`

get

docs/requirements.txt
kevinpollet, line 1: I think this change is not necessary, could you revert it?

kevinpollet (changes requested):

kevinpollet:
Hello @985492783 and thanks for opening the pull request,

Would you like to make the suggested changes, or would you prefer that we handle them for you?

985492783:
> Hello @985492783 and thanks for opening the pull request,
> 
> Would you like to make the suggested changes, or would you prefer that we handle them for you?

I can finish it today, thanks

jspdown:
@985492783 Thanks for making the suggested changes. Looks good. 

You may have seen that the CI is failing. To fix this, could you just run `make generate` to generate the documentation and commit the files?

kevinpollet:
Hello @985492783,

I have rebased the pull request and generated the documentation.

kevinpollet (approved):
Thanks 👍

mmatur (approved):
LGTM

PR #10777, opened 2024-06-03
fix: compress and Accept-Encoding header weight

ldez:
### What does this PR do?

Implements the support of weight with `Accept-Encoding` header for the compression middleware.

### Motivation

https://www.rfc-editor.org/rfc/rfc9110#field.accept-encoding

### More

- [x] Added/updated tests

### Additional Notes

This is not fully compliant to avoid breaking changes.

Related to #9734

### DIFF
--- a/docs/content/middlewares/http/compress.md
+++ b/docs/content/middlewares/http/compress.md
@@ -214,3 +214,42 @@ http:
   [http.middlewares.test-compress.compress]
     minResponseBodyBytes = 1200
 ```
+
+### `defaultEncoding`
+
+_Optional_
+
+`defaultEncoding` specify the default encoding to use if the header `Accept-Encoding` is defined but empty or if there is no supported encoding.
+
+```yaml tab="Docker & Swarm"
+labels:
+  - "traefik.http.middlewares.test-compress.compress.defaultEncoding=gzip"
+```
+
+```yaml tab="Kubernetes"
+apiVersion: traefik.io/v1alpha1
+kind: Middleware
+metadata:
+  name: test-compress
+spec:
+  compress:
+    minResponseBodyBytes: 1200
+```
+
+```yaml tab="Consul Catalog"
+- "traefik.http.middlewares.test-compress.compress.defaultEncoding=gzip"
+```
+
+```yaml tab="File (YAML)"
+http:
+  middlewares:
+    test-compress:
+      compress:
+        defaultEncoding: gzip
+```
+
+```toml tab="File (TOML)"
+[http.middlewares]
+  [http.middlewares.test-compress.compress]
+  defaultEncoding = "gzip"
+```
--- a/pkg/middlewares/compress/acceptencoding.go
+++ b/pkg/middlewares/compress/acceptencoding.go
@@ -1,0 +1,126 @@
+package compress
+
+import (
+	"slices"
+	"strconv"
+	"strings"
+)
+
+const acceptEncodingHeader = "Accept-Encoding"
+
+const (
+	brotliName    = "br"
+	gzipName      = "gzip"
+	identityName  = "identity"
+	wildcardName  = "*"
+	notAcceptable = "not_acceptable"
+)
+
+type Encoding struct {
+	Type   string
+	Weight *float64
+}
+
+func getCompressionType(acceptEncoding []string, defaultType string) string {
+	accepts, hasWeight := parseAcceptsEncoding(acceptEncoding)
+
+	if hasWeight {
+		for _, a := range accepts {
+			if a.Type == identityName && a.Weight != nil && *a.Weight == 0 {
+				return notAcceptable
+			}
+
+			if a.Type == wildcardName && a.Weight != nil && *a.Weight == 0 {
+				return notAcceptable
+			}
+
+			if a.Type == wildcardName {
+				continue
+			}
+
+			return a.Type
+		}
+
+		if defaultType != "" {
+			return defaultType
+		}
+
+		// Follows the pre-existing default order inside Traefik: br > gzip.
+		return brotliName
+	}
+
+	// fallback on pre-existing order inside Traefik
+	for _, s := range []string{brotliName, gzipName} {
+		if slices.ContainsFunc(accepts, func(e Encoding) bool {
+			return e.Type == s || e.Type == wildcardName
+		}) {
+			return s
+		}
+	}
+
+	return defaultType
+}
+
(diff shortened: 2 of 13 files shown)


docs/content/middlewares/http/compress.md
```diff
@@ -214,3 +214,42 @@ http:
   [http.middlewares.test-compress.compress]
     minResponseBodyBytes = 1200
 ```
+
+### `defaultEncoding`
+
+_Optional_
```
rtribotte, line 220:
```suggestion
_Optional, Default=""_
```

docs/content/middlewares/http/compress.md
```diff
@@ -214,3 +214,42 @@ http:
...
+```
+
+```yaml tab="Kubernetes"
+apiVersion: traefik.io/v1alpha1
+kind: Middleware
+metadata:
+  name: test-compress
+spec:
+  compress:
+    minResponseBodyBytes: 1200
```
rtribotte, line 236:
```suggestion
    defaultEncoding: gzip
```

pkg/middlewares/compress/acceptencoding_test.go
```diff
@@ -0,0 +1,109 @@
...
+		values   []string
+		expected string
+	}{
+		{
+			desc:     "br > gzip (no weight)",
+			values:   []string{"gzip, br"},
+			expected: brotliName,
+		},
+		{
+			desc:     "unknown compression type (no weight)",
```
rtribotte, line 21:
```suggestion
			desc:     "no weights, known compression type",
```

pkg/middlewares/compress/acceptencoding_test.go
```diff
@@ -0,0 +1,109 @@
...
+			values:   []string{"compress, gzip"},
+			expected: gzipName,
+		},
+		{
+			desc:     "unknown compression types (no weight) use default",
```
rtribotte, line 26:
```suggestion
			desc:     "no weights, unknown compression types",
```

docs/content/middlewares/http/compress.md
```diff
@@ -214,3 +214,42 @@ http:
   [http.middlewares.test-compress.compress]
     minResponseBodyBytes = 1200
 ```
+
+### `defaultEncoding`
+
+_Optional, Default=""_
+
+`defaultEncoding` specify the default encoding to use if the header `Accept-Encoding` is defined but empty or if there is no supported encoding.
```
Belphemur, line 222: This feature would break the RFC:
> An Accept-Encoding header field with a field value that is empty implies that the user agent does not want any content coding in response.

https://www.rfc-editor.org/rfc/rfc9110#section-12.5.3-12

The compress middleware should not try to compress if the client doesn't expect it.

Belphemur:
Could it make sense to have a setting on the server side to define the preference order for the case where the client doesn't provide the `q=` weight ?

The RFC doesn't specify what to do in case of equal weight: 

> A representation could be encoded with multiple content codings. However, most content codings are alternative ways to accomplish the same purpose (e.g., data compression). When selecting between multiple content codings that have the same purpose, the acceptable content coding with the highest non-zero qvalue is preferred.

https://www.rfc-editor.org/rfc/rfc9110#section-12.5.3-11

docs/content/middlewares/http/compress.md
ldez: yes I reverted that.

ldez:
When this PR is merged, I will change the order inside your PR to `zstd` > `br` > `gzip`.
But with a default to `br` for a wildcard only (for compatibility).

pkg/middlewares/compress/acceptencoding.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+const acceptEncodingHeader = "Accept-Encoding"
+
+const (
+	brotliName    = "br"
+	gzipName      = "gzip"
+	identityName  = "identity"
+	wildcardName  = "*"
+	notAcceptable = "not_acceptable"
```
rtribotte, line 16: Couldn't we get rid of `notAcceptable` and only use identity?

ldez: The spec defines "not acceptable".
In the future, this value should be used to generate 415.

rtribotte: Ah, right!

rtribotte (approved):
LGTM

pkg/middlewares/compress/acceptencoding.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	if slices.ContainsFunc(encodings, func(e Encoding) bool { return e.Type == wildcardName }) {
+		return defaultType
+	}
+
+	return identityName
+}
+
+func parseAcceptsEncoding(acceptEncoding []string) ([]Encoding, bool) {
```
kevinpollet, line 67:
```suggestion
func parseAcceptEncoding(acceptEncoding []string) ([]Encoding, bool) {
```

pkg/middlewares/compress/acceptencoding.go
```diff
@@ -0,0 +1,137 @@
...
+	Weight *float64
+}
+
+func getCompressionType(acceptEncoding []string, defaultType string) string {
+	if defaultType == "" {
+		// Keeps the pre-existing default inside Traefik.
+		defaultType = brotliName
+	}
+
+	encodings, hasWeight := parseAcceptsEncoding(acceptEncoding)
```
kevinpollet, line 30:
```suggestion
	encodings, hasWeight := parseAcceptEncoding(acceptEncoding)
```

pkg/middlewares/compress/acceptencoding_test.go
```diff
@@ -0,0 +1,143 @@
...
+			t.Parallel()
+
+			encodingType := getCompressionType(test.values, test.defaultType)
+
+			assert.Equal(t, test.expected, encodingType)
+		})
+	}
+}
+
+func Test_parseEncodingAccepts(t *testing.T) {
```
kevinpollet, line 80:
```suggestion
func Test_parseAcceptEncoding(t *testing.T) {
```

kevinpollet (changes requested):
Nice one 👍 
Just some minor suggestions.

mmatur (approved):
LGTM

tomMoulard (approved):
LGTM :ok_hand:

kevinpollet (approved):
LGTM

PR #10779, opened 2024-06-04
optimization: struct aligment

peacewalker122:
### What does this PR do?
Some Optimization by realign struct

### Motivation
Traefik were my daily driver and when i start looking the code, i look this optimization.

### More

- [ ] Added/updated tests


ldez:
Hello,

this kind of optimization in the context of Traefik is useless: these are extremely small marginal gains.

Readability takes priority over this type of micro-optimization.

PR #10785, opened 2024-06-05
Documentation: Getting Started link to User Guides

norlinhenrik:
I wish I would have seen the User Guides after Getting Started. That would have saved me many hours.


kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks 👍

PR #10791, opened 2024-06-07
feat: adds support for newer config for wasm plugins.

jcchavezs:
Right now what we receive under config for a wasm plugin we forward it to the wasm guest without having the opportunity to config the runtime. This change allows to support a new config structure which fields under 'runtime' will configure the runtime and those under 'config' will be sent directly to the guest as well as supporting the old structure. This is related to https://github.com/traefik/traefik/issues/10739 and https://github.com/jcchavezs/coraza-http-wasm/pull/19

Both configurations will be supported but I suggest we update Traefik documents to support the later one.

```yaml
# old way
http:
# ...
  middlewares:
    waf:
      plugin:
        coraza:
          directives:
            - SecRuleEngine On
            - SecDebugLog /dev/stdout
```

```yaml
# new way
http:
# ...
  middlewares:
    waf:
      plugin:
        coraza:
          runtime: 
            rootFS: /etc/traefik 
          guestConfig:
            directives:
              - SecRuleEngine On
              - SecDebugLog /dev/stdout
```

Tests will come if the design of the feature is approved.

### DIFF
--- a/pkg/plugins/middlewarewasm.go
+++ b/pkg/plugins/middlewarewasm.go
@@ -32,6 +32,55 @@ func (b wasmMiddlewareBuilder) newMiddleware(config map[string]interface{}, midd
 	}, nil
 }
 
+func toWasmConfig(i any) ([]handler.Option, bool, error) {
+	b, err := json.Marshal(i)
+	if err != nil {
+		return nil, false, fmt.Errorf("marshaling config: %w", err)
+	}
+
+	var wc wasmConfig
+	if err := json.Unmarshal(b, &wc); err != nil {
+		return nil, false, fmt.Errorf("unmarshaling config: %w", err)
+	}
+
+	if wc.Runtime == nil && wc.Config == nil {
+		return nil, false, nil
+	}
+
+	return []handler.Option{
+		handler.GuestConfig(wc.Config),
+		handler.ModuleConfig(
+			wazero.NewModuleConfig().
+				WithSysWalltime().
+				WithFSConfig(wazero.NewFSConfig().WithReadOnlyDirMount(wc.Runtime.RootFS, "/")),
+		),
+	}, true, nil
+}
+
+func toConfig(i any) ([]handler.Option, error) {
+	im, ok := i.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("could not type assert config: %T", i)
+	}
+
+	data, err := json.Marshal(im)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling config: %w", err)
+	}
+
+	return []handler.Option{
+		handler.GuestConfig(data),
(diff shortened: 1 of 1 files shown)


mmatur:
Hi @jcchavezs,

Thanks for this PR, with @juliens we are thinking about a different approach to configure the fs mount on the host.

WDYT of configuring the runtime mounts in the static configuration like that:

```yaml
experimental:
  plugins:
    traefik-wasm-grain:
      moduleName: "github.com/jcchavezs/coraza-http-wasm-traefik"
      version: "v0.2.1"
      runtime:
        mounts:
          - /src:/dst:ro
```

To us it will ease the UX because the mounts file need to be on the same host as Traefik and we are thinking that configuring mounts on the dynamic configuration could be confusing for end user.

jcchavezs:
@mmatur that sounds very legit but I wonder if `runtime` should be a first class element in `plugin`. Is there any other plugin that has extra configuration? shall we call it runtime or something like "host_config"?

Maybe we make this generic? Also, is it possible to pass this all the way up to the middleware creation?

jcchavezs:
I guess this is unique to wasm environment as other plugins have direct access to filesystem. If that is the case maybe we call it `settings`.

mmatur:
@jcchavezs thanks for your answer

> I wonder if runtime should be a first class element in plugin

We think that it should not because with this approach you can't have multiple plugins with different fs mount like the following

```yaml
experimental:
  plugins:
    coraza:
      moduleName: "github.com/jcchavezs/coraza-http-wasm-traefik"
      version: "v0.2.1"
      runtime:
        mounts:
          - /src:/dst:ro
    coraza2:
      moduleName: "github.com/jcchavezs/coraza-http-wasm-traefik"
      version: "v0.2.1"
      runtime:
        mounts:
          - /tmp:/tmp:rw
```

> Is there any other plugin that has extra configuration?

For now there is no other plugin that require this extra configuration.

> Maybe we make this generic? Also, is it possible to pass this all the way up to the middleware creation?

It should.

> I guess this is mainly for wasm environments as other plugins have direct access to filesystem.

Today, yes it is only for WASM plugin but we can imagine introducing that for yaegi in the future for security reason (like enabling unsafe)

jcchavezs:
Ok, just one thing: maybe instead of `runtime` we call it settings.

José Carlos Chávez


ons. 12. juni 2024 kl. 14:20 skrev Michael ***@***.***>:

juliens:
Hello @jcchavezs 

During our work on this PR https://github.com/traefik/traefik/pull/10829 we integrated this feature, if it's seems ok for you, may be we can close this one?

juliens:
Hello @jcchavezs ,
As said previously, I think the config was handled in the PR #10829 so I close the PR, if you think we still need something from this PR, feel free to open a new one.

PR #10813, opened 2024-06-13
Remove interface names from ipv6 (fixes #7270)

### LINKED ISSUE #7270: [Windows] ipWhiteList middleware failed to parse IP on (local) IPv6 with zone id attached

Bug

### What did you do?
I set up an ipWhiteList middleware for our application and tried to access one of the routers with the middleware attached.

A valid service response.

### What did you see instead?
A canceled request.

```
Version:      2.2.11
Codename:     chevrotin
Go version:   go1.14.8
Built:        2020-09-07T14:12:48Z
OS/Arch:      windows/amd64
```

```toml

[http.services.nginx]
      [http.services.nginx.loadBalancer]
        [[http.services.NGINX.loadBalancer.servers]]
          url = "http://example.com:8080"

[http.routers.nginx]
      entryPoints = ["web", "web-secure"]
      priority = 10
      service = "nginx"
      rule = "PathPrefix(`/test`)"
      middlewares = ["whitelist"]

[http.middlewares.whitelist.ipWhiteList]
      sourceRange = ["0.0.0.0/0", "::/0"]
```

```
RemoteAddr:[fe80::xxxx:yyy:zzz:0123%vEthernet (Default Switch)]:58904 RequestURI:/test/ TLS:0xc0003da210 Cancel:<nil> Response:<nil> ctx:0xc0008ef530}: unable to parse address: fe80::xxxx:yyyy:zzzz:0123%vEthernet (Default Switch): can't parse IP from address fe80::xxxx:yyyy:zzzz:0123%vEthernet (Default Switch)"
```

JeroenED:
### What does this PR do?

Fixes #7270

This PR removes the interfacename from an IpAddress before it gets parsed as net.ParseIP does not allow the interfacename to be there

### Motivation

Frankly, this is pure self service. I was suffering from this issue myself and wanted to get it fixed.

### More

- [ ] Added/updated tests

### Additional Notes
I never wrote any golang code before so please allow if my code has done stuff that you should not do.

### DIFF
--- a/pkg/ip/checker.go
+++ b/pkg/ip/checker.go
@@ -22,6 +22,8 @@ func NewChecker(trustedIPs []string) (*Checker, error) {
 	checker := &Checker{}
 
 	for _, ipMask := range trustedIPs {
+		ipMask := removeInterfaceName(ipMask)
+
 		if ipAddr := net.ParseIP(ipMask); ipAddr != nil {
 			checker.authorizedIPs = append(checker.authorizedIPs, &ipAddr)
 			continue
@@ -45,6 +47,7 @@ func (ip *Checker) IsAuthorized(addr string) error {
 	if err != nil {
 		host = addr
 	}
+	host = removeInterfaceName(host)
 
 	ok, err := ip.Contains(host)
 	if err != nil {
@@ -88,6 +91,15 @@ func (ip *Checker) ContainsIP(addr net.IP) bool {
 	}
 
 	return false
+}
+
+// removeInterfaceName from IPv6 addresses (eg. 2001::64:48fe%eth0 to 2001::64:48fe)
+func removeInterfaceName(addr string) string {
+	index := strings.Index(ipv6Address, "%")
+	if index == -1 {
+		return addr
+	}
+	return addr[:index]
 }
 
 func parseIP(addr string) (net.IP, error) {


jspdown (changes requested):
Thanks @JeroenED for opening this PR.

There was indeed an issue with IPv6 zones. As you've identified, the problem comes from the `net.ParseIP` which doesn't support zones. 
Internally, `net.ParseIP` relies on `netip.ParseAddr` which does support zones. 

So, instead of removing the zone I would suggest to improve our helper function `parseIP` to do something like this, and use the helper in the `NewChecker` function.

```golang
func parseIP(addr string) (net.IP, error) {
	parsedAddr, err := netip.ParseAddr(addr)
	if err != nil {
		return nil, err
	}
	
	ip := parsedAddr.As16()
	return ip[:], nil
}
```

This would have the benefit of relying more on the standard library and providing better errors.

JeroenED:
I changed the PR as requested. I did however retain the original error line

jspdown (changes requested):
I've tested the changes and successfully fixes the issue :+1: 
One last thing before I approve the PR, could you add a new test case in `TestContainsIsAllowed`?

JeroenED:
Hi,
Fixed the import and added 2 accept and reject test case.

pkg/ip/checker.go
```diff
@@ -91,10 +92,11 @@ func (ip *Checker) ContainsIP(addr net.IP) bool {
...
+	parsedAddr, err := netip.ParseAddr(addr)
+	if err != nil {
 		return nil, fmt.Errorf("can't parse IP from address %s", addr)
 	}
 
-	return userIP, nil
-}
+	ip := parsedAddr.As16()
+	return ip[:], nil
+}
```
rtribotte, line 102:
```suggestion
}

```

JeroenED: Fixed it

JeroenED:
The FMT errors should be fixed as well now. Sorry for all these n00b-mistakes. Freaking me out as well.

jspdown (approved):
LGTM :+1:

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #10815, opened 2024-06-14
Support HTTPRoute method and query param matching

kevinpollet:
### What does this PR do?

This pull request adds support for method and query param matches for HTTPRoutes.

Note: the priority computation has been simplified as in the [spec](https://github.com/kubernetes-sigs/gateway-api/blob/v1.0.0/apis/v1/httproute_types.go#L164) regex precedence is implementation specific so simplifying is ok IMO.

### Motivation

To improve the Gateway API v1.0.0 support.

### More

- [X] Added/updated tests


rtribotte (approved):
LGTM

mmatur (approved):

nmengin (approved):
LGTM

PR #10843, opened 2024-06-24
Draft: Surface provider configuration conflicts to API for visibility

dekimsey:
### What does this PR do?

(_This is a proof-of-concept to foster a discussion on how to improve provider error behavior for configuration conflicts. It's not complete!_)

This change exposes an error field to the `Router` configuration such that when populated the runtime will automatically disable the route but leave the object defined. This is similar to how syntactically invalid route rules are handled today. This enhances visibility to operators by making the provider failure more accessible where it is otherwise only discoverable via logs.

### Motivation

Currently, when traefik loads dynamic configurations from providers and merges them together it will check to see if they conflict. When they conflict, the resultant object (typically `Router`, `Loadbalancer`, or `Middleware`) will be dropped from the configuration update. This is [by design][1].

Unfortunately, this misconfiguration is very difficult to detect in the field as it can only be discovered by observing the *lack* of the desired config in the API or examining the traefik daemon logs. Further it means a new service may cause a complete failure in the currently live service routing simply by existing (for instance canary deployments, whose purpose is to check compatibility).

#### Alternatives

I did also explore having `provider.Merge()` return an error such that the caller can choose how to proceed. But this only makes sense if the provider can opt to fall-back or make a different decision with it's inputs. Without knowing the current state, I don't see this being particularly useful. An example might be to prune the new canary instances to maintain the current "live" configuration instead of dropping altogether as that's highly disruptive. 

A different solution to this problem might be for providers to bail-out and not make an update at all. But the update payload is all-in-one, there's no way to retrieve the current configs and choose that over the new ones. Further, Providers also have no way to surface error messages to operators since they run async via consul tags or some other mechanism. And if they could, no way to correlate the update to the operator for evaluation.

### More

- [ ] Added/updated integration tests (note: I wasn't sure how this works)
- [ ] Added/updated middleware/load-balancer support

### Additional Notes

Related:
  - traefik/traefik#7738
  - traefik/traefik#8256

There are a few other issues related to surfacing providers errors more  effectively, but they active ones seem to focus on metric based detection. Which while useful, doesn't help me at the time of publishing an update to verify changes were valid.

[1]: https://github.com/traefik/traefik/issues/8256#issuecomment-882415498

### DIFF
--- a/docs/content/contributing/documentation.md
+++ b/docs/content/contributing/documentation.md
@@ -15,7 +15,7 @@ Let's see how.
 
 ### General
 
-This [documentation](https://doc.traefik.io/traefik/ "Link to the official Traefik documentation") is built with [MkDocs](https://mkdocs.org/ "Link to website of MkDocs").
+This [documentation](../../ "Link to the official Traefik documentation") is built with [MkDocs](https://mkdocs.org/ "Link to website of MkDocs").
 
 ### Method 1: `Docker` and `make`
 
--- a/docs/content/migration/v2-to-v3-details.md
+++ b/docs/content/migration/v2-to-v3-details.md
@@ -135,7 +135,7 @@ It is now unsupported and would prevent Traefik to start.
 ##### Remediation
 
 The `http3` option should be removed from the static configuration experimental section.
-To configure `http3`, please checkout the [entrypoint configuration documentation](https://doc.traefik.io/traefik/v3.0/routing/entrypoints/#http3_1).
+To configure `http3`, please checkout the [entrypoint configuration documentation](../routing/entrypoints.md#http3_1).
 
 ### Consul provider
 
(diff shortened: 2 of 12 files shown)


nmengin:
Hello @dekimsey,

Your PR has been open as a draft for a while. Have you planned to work on it soon?
Thanks in advance for your feedback.

nmengin:
Hi @dekimsey,

Thanks a lot for your contribution!
Since there hasn’t been any activity on this PR for a while, we’re going to close it for now to keep the queue manageable.

If you’d like to continue working on it or if you think we should revisit, please or let us know and we’ll reopen the PR for you. We’d be happy to pick up the discussion again when you’re ready.

Thanks again for your effort and interest in improving the project! :pray:

PR #10884, opened 2024-07-06
Draft: Add SNI routing support for MySQL

### LINKED ISSUE #10505: MySQL client cannot connect to database when using SNI routing with TLS
### Welcome!

### What did you do?

For a while Traefik could not route TCP (w/ TLS) requests for MySQL as both the client and server did not support SNI. The only option was to use a wildcard router ('HostSNI(`*`') which works fine, until you need enable TLS to host multiple MySQL instances using the same entrypoint on Traefik.

This has since been addressed by MySQL and SNI support has been added [as of version 8.1.0 ](https://dev.mysql.com/doc/relnotes/mysql/8.1/en/news-8-1-0.html)(2023/07) 

### What did you see instead?

Enabling TLS somehow confuse the mysql client (and also anything relying on the mysql library) which ends up hanging, waiting for a specific packet (?)

Example (see docker-compose.yml provided) with TLS enabled on the router, tested with passthrough on, and off:
```bash
docker run --rm -it --network host mysql:8.3.0 mysql -u root -h 127.0.0.1 --port 3306 --ssl-mode=REQUIRED 
ERROR 2013 (HY000): Lost connection to MySQL server at 'reading initial communication packet', system error: 0
```

We can confirm Traefik routes the traffic to the right container using ncat, implying that this may just be a protocol issue between the mysql client and traefik (I'm nowhere near qualified enough in either mysql or traefik to say for sure) :
```bash
➜  traefik-l4 ncat --ssl 127.0.0.1 3306 -v
Ncat: Version 7.94SVN ( https://nmap.org/ncat )
Ncat: Subject: CN=TRAEFIK DEFAULT CERT
Ncat: Issuer: CN=TRAEFIK DEFAULT CERT
Ncat: SHA-1 fingerprint: 0E49 C6BF 4852 D091 EDB7 2B6F E29A 42EC 64C5 795E
Ncat: Certificate verification failed (self-signed certificate).
Ncat: SSL connection to 127.0.0.1:3306.
Ncat: SHA-1 fingerprint: 0E49 C6BF 4852 D091 EDB7 2B6F E29A 42EC 64C5 795E
I
8.3.0   :;
          :I�����O
                  7M89Me Bcaching_sha2_password
```

And with TLS disabled on the router, it works fine:
```bash
docker run --rm -it  --network host mysql:8.3.0 mysql -h 127.0.0.1 --port 3306 --ssl-mode=REQUIRED -u root
mysql: [Warning] Using a password on the command line interface can be insecure.
Welcome to the MySQL monitor.  Commands end with ; or \g.
...
```

Maybe this is a misunderstanding on my part, and that there is additional configuration required. But I guess there may be some work needed on Traefik to improve compatibility with MySQL [as there was a similar PR done for PostgreSQL](https://github.com/traefik/traefik/pull/9377) and I would be happy to help diagnose / open a PR if that's the case. 

A community forum discussion exists for this issue with people experiencing the same behaviour: https://community.traefik.io/t/multiple-mysql-communication-through-tcp-with-tls-based-on-sni/18936/19

Version:      3.0.0-rc1
Codename:     cheddar
Go version:   go1.22.0
Built:        2024-02-13T13:41:20Z
OS/Arch:      linux/amd64

Minimal reproduction example:
```yaml
version: '2.1'

services:
  traefik:
    image: "traefik:v3.0.0-rc1"
    container_name: "traefik"
    command:
      - "--log.level=DEBUG"
      - "--api.insecure=true"
      - "--providers.docker=true"
      - "--entrypoints.mysql.address=:3306"
    ports:
      - "3306:3306"
      - "8089:8080"
    volumes:
      - "/var/run/docker.sock:/var/run/docker.sock:ro"
  db:
    image: 'mysql:8.3.0'
    labels:
      - "traefik.enable=true"
      - "traefik.tcp.routers.db.rule=HostSNI(`*`)"
      - "traefik.tcp.services.db.loadbalancer.server.port=3306"
      - "traefik.tcp.routers.db.entrypoints=mysql"
      - "traefik.tcp.routers.db.tls=true"
    environment:
      - MYSQL_ROOT_PASSWORD=magento2
      - MYSQL_DATABASE=magento2
      - MYSQL_USER=magento2
      - MYSQL_PASSWORD=magento2
```

(Obviously, to get SNI working, this docker would run "behind" a domain name, with proper certificates etc, this is only meant to show that enabling TLS on a TCP router doesn't work with MySQL)

_No response_

NoOverflow:
### What does this PR do?

This PR aims to add support for L4 Routing for MySQL using SNI. 

### Motivation

For a while Traefik could not route TCP (w/ TLS) requests for MySQL as both the client and server did not support SNI. The only option was to use a wildcard router ('HostSNI(*') which works fine, until you need enable TLS to host multiple MySQL instances using the same entrypoint on Traefik.

This has since been addressed by MySQL and SNI support has been added [as of version 8.1.0 ](https://dev.mysql.com/doc/relnotes/mysql/8.1/en/news-8-1-0.html)(2023/07)

Adding routing for MySQL would allow multiple databases to be behind a single Traefik instance.

### Draft issues 
- [X] Fake handshake packet
- [ ] Figure out how to send a proper handshake packet, as there may be multiple MySQL instances using different configuration values, we can't just send fake values, we also cannot get them from the right server since we don't know at this point which one the client is trying to reach.
- [ ] Find a proper way to identify a MySQL connection, since it uses a client-first protocol. ([An example of fingerprinting for Postgresql](https://github.com/traefik/traefik/blob/b1b4e6b918e8eeaf9e24823baf24dbc77f7d373e/pkg/server/router/tcp/postgres.go#L22))

### More

- [ ] Fix Draft issues (see above)

#### Useful links
- https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_tls.html
- https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_handshake.html
- https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_ssl_request.html
- https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase.html

Closes #10505

### DIFF
--- a/pkg/server/router/tcp/mysql.go
+++ b/pkg/server/router/tcp/mysql.go
@@ -1,0 +1,79 @@
+package tcp
+
+import (
+	"bufio"
+	"encoding/base64"
+	"encoding/binary"
+	"github.com/rs/zerolog/log"
+	"github.com/traefik/traefik/v3/pkg/tcp"
+)
+
+// isMysql determines whether the buffer contains the MySQL STARTTLS message.
+// Reference: https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_tls.html
+//            https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_handshake.html
+//            https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase_packets_protocol_ssl_request.html
+func isMysql(br *bufio.Reader) (bool, error) {
+	// There doesn't seem to be a good way to fingerprint MySQL connections
+	return true, nil
+}
+
+
+func (r *Router) serveMysql(conn tcp.WriteCloser) {
+	// MySQL is a server-first protocol, we need to send the handshake first, this is a base64 encoded string of a valid one taken from mysql:8.3.0
+	// Reference: https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_connection_phase.html
+	encodedStr := "SQAAAAo4LjMuMAALAAAAQEsOLEhSWBsA////AgD/3xUAAAAAAAAAAAAARXReNToYXyxSJW5uAGNhY2hpbmdfc2hhMl9wYXNzd29yZAA="
+	decodedBytes, _ := base64.StdEncoding.DecodeString(encodedStr)
+	log.Info().Msg("Sending handshake.")
+	_, err := conn.Write(decodedBytes)
+	if err != nil {
+		log.Error().Err(err).Msg("Error while sending handshake")
+		conn.Close()
+		return
+	}
+
+	// We expect a SSLRequest as response, which is 36 bytes long, it is like a normal handshake response but cut short before the username
+	// TODO: Parse the response and check if the client requested SSL instead of manually checking the username byte
+	br := bufio.NewReader(conn)
+	protocol_packet_length_raw := make([]byte, 3) // MySQL packet length is a 3 byte integer
+	_, err = br.Read(protocol_packet_length_raw)
+	if err != nil {
+		conn.Close()
+		return
+	}
+	protocol_packet_length_raw = append(protocol_packet_length_raw, 0) // We need to add a 0 byte to parse it as a little endian UInt32
+	protocol_packet_length := binary.LittleEndian.Uint32(protocol_packet_length_raw)
+	log.Info().Msgf("Received protocol packet length: %d", protocol_packet_length)
+
+	// Read the rest of the packet, this could probably be optimised by just skipping them since we don't need them
+	b := make([]byte, protocol_packet_length + 1) // +1 to include the packet number
+	_, err = br.Read(b)
+	if err != nil {
+		conn.Close()
+		return
+	}
+
+	// The first packet after the handshake should be a SSLRequest or a HandshakeResponse320/41
+	// If it is a SSLRequest, it should be 32 bytes long (headers + empty username)
+	// TODO: Is there a better way to check ?
+	if protocol_packet_length != 32 {
+		log.Info().Msg("MySQL client did not request SSL, we can't properly route this connection.")
+		conn.Close()
+		return
+	}
+
+	// The next packet sent should be a client hello packet
+	hello, err := clientHelloInfo(br)
+	if err != nil {
+		conn.Close()
+		return
+	}
+
+	if !hello.isTLS {
+		log.Error().Msg("MySQL client did not send a TLS Client Hello, closing connection.")
+		conn.Close()
+		return
+	}
+	log.Info().Msg("MySQL client sent a TLS Client Hello, success !")
+	log.Info().Msgf("Server Name: %s", hello.serverName)
+	log.Info().Msgf("Protocols: %v", hello.protos)
+}
--- a/pkg/server/router/tcp/mysql.go
+++ b/pkg/server/router/tcp/mysql.go
@@ -6,6 +6,9 @@ import (
 	"encoding/binary"
 	"github.com/rs/zerolog/log"
 	"github.com/traefik/traefik/v3/pkg/tcp"
+	tcpmuxer "github.com/traefik/traefik/v3/pkg/muxer/tcp"
+	"sync"
+	"errors"
 )
 
 // isMysql determines whether the buffer contains the MySQL STARTTLS message.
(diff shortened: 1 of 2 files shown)


nmengin:
Hey @NoOverflow,

Could you tell us when your PR will be in a status that allow us to start our review?

Thanks in advance.

NoOverflow:
Hey @nmengin,
I wish I could tell you, there are still many problems that need to be fixed. I opened this PR to basically kickstart the work for MySQL support (see https://github.com/traefik/traefik/issues/10505#issuecomment-2208887504)

Kwuray:
Hello,

Is it really a problem to send a fake initial handshake packet?

We could test each server flag capability and measure the impact of sending a false value?
For example, the predominant encoding is that of the client, and sending a false value does not seem very problematic.

https://dev.mysql.com/doc/refman/8.4/en/charset-connection.html#charset-connection-error-handling

It appears that it is impossible to determine the SNI before sending the initial handshake packet, or even resend it later for correction.

Kwuray:
Hi @NoOverflow 

How can I suggest changes? For example to check that it is an SSL request : `// TODO: Parse the response and check if the client requested SSL instead of manually checking the username byte` 

Can I push directly into your fork?
I have never participated in open source projects...

Thanks

NoOverflow:
> Hello,
> 
> Is it really a problem to send a fake initial handshake packet?
> 
> We could test each server flag capability and measure the impact of sending a false value? For example, the predominant encoding is that of the client, and sending a false value does not seem very problematic.
> 
> https://dev.mysql.com/doc/refman/8.4/en/charset-connection.html#charset-connection-error-handling
> 
> It appears that it is impossible to determine the SNI before sending the initial handshake packet, or even resend it later for correction.

Hi @Kwuray, this is a question that needs answering, maybe it will break or not, there are many edge cases to test

> Hi @NoOverflow
> 
> How can I suggest changes? For example to check that it is an SSL request : `// TODO: Parse the response and check if the client requested SSL instead of manually checking the username byte`
> 
> Can I push directly into your fork? I have never participated in open source projects...
> 
> Thanks

You're free to open a PR on my repo targeting the `NoOverflow:feat-mysql-support-v3.0 branch` 😄

umeshkhatiwada1:
so Any update  regarding the PR @NoOverflow

GMProg:
Hey, I've also run into this, as I see it there is not a fine and nice solution, (other than waiting for the change of mysql protocol), but what if we had options when defining the entrypoint that it is a server-first entrypoint and also the protocol, so when receiving a packet on that port we would know what to do with it. 
I know this also has it's limitations, like one type of portocol/port, but I think this is better than the current one service/port situation, and not to mention this solution would set the base of other server-first protocols.

@nmengin What do you think, would this type of solution be acceptable by traefik's end?

Kwuray:
@GMProg with your solution we would do the same thing with the request regardless the MySQL service ? So impossible to have various MySQL server (different Charest, version, ...) ? Maybe I did not understand your idea 😅

GMProg:
> @GMProg with your solution we would do the same thing with the request regardless the MySQL service ? So impossible to have various MySQL server (different Charest, version, ...) ? Maybe I did not understand your idea 😅

well yes, my suggestion would mean that all mysql servers should have the same configuration, at least to the level that's sent by the server first, and also there's the connection id parameter, which is just, really hard to fake, you're right this not the way either, sadly.

NoOverflow:
@GMProg You could indeed do that, but as @Kwuray correctly mentioned, you would be stuck with a single version of MySQL behind your traefik instance since MySQL is a server-first protocol and the first packet (even before STARTTLS negotiation) contain server properties that you cannot "get" as you don't know where the request is heading to yet. 

This port-based logic with a forged initial packet would help in (very) specific cases but would not make it into an official Traefik release.

The only way to progress further with this PR is to initiate a talk with the MySQL maintainers and propose a new protocol that would allow SNI routing (way easier said than done though...)

nmengin:
Hello @NoOverflow,

> @nmengin What do you think, would this type of solution be acceptable by traefik's end?

We have discussed with the other maintainers the opportunity to dedicate an entrypoint to the mysql protocol.
And... We haven't reached any consensus on it!

Even if the proposition is interesting, some maintainers would prefer to manage this mechanism globally.
For this reason, we'd prefer not to go this way until more discussion on this topic.

kevinpollet:
Hi @NoOverflow,

We discussed this issue during our triage session, and as mentioned in the previous message, we would like to address this globally, for example by having dedicated entry points. This would allow us to support other protocols without having to have assertions in the connection path to detect them, which can get tricky at some point and introduce incompatibilities.

And as you said, even with dedicated entry points, this may not be possible today.

So we are closing this pull request.

PR #10892, opened 2024-07-10
Docs: incorrect value in default priority computation example

rltas:
### What does this PR do?

The small table under "How default priorities are computed" states that the priority of Router-1 is 44 while it's actually 34.

### Motivation

I was confused by the docs.

### More

- [ ] Added/updated tests


rltas:
Eh? I don't think this was supposed to happen.

PR #10917, opened 2024-07-18
Allow setting service.name for OTLP metrics

### LINKED ISSUE #10908: Allow to define different serviceName for OTLP metrics
### Welcome!

It would be great if we were allowed to define a different serviceName for OTLP metrics, as when running multiple instances of Traefik (such as one for internal load balancers and one for external load balancers), there's no way to differentiate the metrics from each Traefik instance.

This is currently hard-coded here: https://github.com/traefik/traefik/blob/master/pkg/metrics/opentelemetry.go#L210

cmartell-at-m42:
### What does this PR do?
Allows to configure service.name used by OTLP metrics rather than using the hard-coded `traefik` value. This is useful for situations where you are using multiple traefik instances (e.g. an internal + an external instance) and you need to differentiate between them in the metrics.

### Motivation
Fixes #10908

### More

- [ ] Added/updated tests

### Additional Notes
I had to update .golangci.yml as it would otherwise complain about the string "traefik" being repeated too often and to make it a constant. Due to the different usages of "traefik" (some are for prefix, some are for service name, some are for entrypoint) I didn't want to make it a const, but happy to revisit.

### DIFF
--- a/pkg/types/metrics.go
+++ b/pkg/types/metrics.go
@@ -116,6 +116,7 @@ type OTLP struct {
 	AddServicesLabels    bool           `description:"Enable metrics on services." json:"addServicesLabels,omitempty" toml:"addServicesLabels,omitempty" yaml:"addServicesLabels,omitempty" export:"true"`
 	ExplicitBoundaries   []float64      `description:"Boundaries for latency metrics." json:"explicitBoundaries,omitempty" toml:"explicitBoundaries,omitempty" yaml:"explicitBoundaries,omitempty" export:"true"`
 	PushInterval         types.Duration `description:"Period between calls to collect a checkpoint." json:"pushInterval,omitempty" toml:"pushInterval,omitempty" yaml:"pushInterval,omitempty" export:"true"`
+	ServiceName          string         `description:"OTEL service name to use." json:"serviceName,omitempty" toml:"serviceName,omitempty" yaml:"serviceName,omitempty" export:"true"`
 }
 
 // SetDefaults sets the default values.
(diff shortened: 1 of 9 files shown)


rtribotte:
Hello @cmartell-at-ocp,

Thanks for opening this PR!

I see that the documentation for this new option missing, could you please add it?
Also, could you please add a note to mention the `OTEL_SERVICE_NAME` env variable and its behavior?

DrFaust92:
drive by review: i would expect to enforce same value for otel traces and metrics. so those could be correlated

pkg/types/metrics.go
```diff
@@ -127,6 +128,7 @@ func (o *OTLP) SetDefaults() {
 	o.AddServicesLabels = true
 	o.ExplicitBoundaries = []float64{.005, .01, .025, .05, .075, .1, .25, .5, .75, 1, 2.5, 5, 7.5, 10}
 	o.PushInterval = types.Duration(10 * time.Second)
+	o.ServiceName = "traefik"
```
youkoulayley, line 131: It would be great to have a test when overriding this value. There is already a test called `TestOpenTelemetry` which tests the default values. Either you add a new test based on this one, or you could add a test case to test overriding the `ServiceName`.

kevinpollet:
Hello @cmartell-at-ocp and thanks for opening this pull request,

Are you still working on this, or would you prefer that we make the changes?

lbenguigui (approved):
LGTM

youkoulayley (approved):

PR #10921, opened 2024-07-21
feat: record trace id and EntryPoint span id into access log

weijiany:
### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/middlewares/observability/entrypoint.go
+++ b/pkg/middlewares/observability/entrypoint.go
@@ -3,6 +3,7 @@ package observability
 import (
 	"context"
 	"fmt"
+	"github.com/traefik/traefik/v3/pkg/middlewares/accesslog"
 	"net/http"
 	"strconv"
 	"strings"
@@ -21,6 +22,8 @@ import (
 
 const (
 	entryPointTypeName = "TracingEntryPoint"
+	traceId            = "TraceId"
+	spanId             = "SpanId"
 )
 
 type entryPointTracing struct {
@@ -68,6 +71,12 @@ func (e *entryPointTracing) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	span.SetAttributes(attribute.String("entry_point", e.entryPoint))
 
 	e.tracer.CaptureServerRequest(span, req)
+	logData := accesslog.GetLogData(req)
+	if logData != nil {
+		spanCtx := span.SpanContext()
+		logData.Core[traceId] = spanCtx.TraceID()
+		logData.Core[spanId] = spanCtx.SpanID()
+	}
 
 	recorder := newStatusCodeRecorder(rw, http.StatusOK)
 	e.next.ServeHTTP(recorder, req)
--- a/pkg/middlewares/observability/entrypoint_test.go
+++ b/pkg/middlewares/observability/entrypoint_test.go
@@ -2,6 +2,7 @@ package observability
 
 import (
 	"context"
+	"github.com/traefik/traefik/v3/pkg/middlewares/accesslog"
 	"net/http"
 	"net/http/httptest"
 	"testing"
@@ -180,4 +181,26 @@ func TestEntryPointMiddleware_metrics(t *testing.T) {
 			metricdatatest.AssertEqual[metricdata.Metrics](t, expected, got.ScopeMetrics[0].Metrics[0], metricdatatest.IgnoreTimestamp(), metricdatatest.IgnoreValue())
 		})
 	}
+}
+
+func TestEntryPointMiddleware_tracing_info_into_log(t *testing.T) {
+	req := httptest.NewRequest(http.MethodGet, "http://www.test.com/", nil)
+	req = req.WithContext(context.WithValue(
+		req.Context(),
+		accesslog.DataTableKey,
+		&accesslog.LogData{
+			Core: accesslog.CoreLogData{},
+		}))
+	next := http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {})
+
+	tracer := &mockTracer{}
+
+	handler := newEntryPoint(context.Background(), tracing.NewTracer(tracer, []string{}, []string{}, []string{}), nil, "test", next)
+	handler.ServeHTTP(httptest.NewRecorder(), req)
+
+	expectedSpanCtx := tracer.spans[0].SpanContext()
+
+	logData := accesslog.GetLogData(req)
+	assert.Equal(t, expectedSpanCtx.TraceID(), logData.Core[traceId])
+	assert.Equal(t, expectedSpanCtx.SpanID(), logData.Core[spanId])
 }


weijiany:
Hello @rtribotte  please review it in your free time 😁

weijiany:
Hi @rtribotte I have a question about how to push this changes about lint.

I noticed that only the final PR's commit must follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/), for this fix lint commit, I am not sure whether I need to force push?

rtribotte:
Hello @weijiany, sure, you can push separate commits on your branch. When a PR is merged the first commit message is kept to squash every commit.

weijiany:
@rtribotte could you please run CI again?

rtribotte:
Hello @weijiany,

I noticed that the documentation needs to be updated to mention `TraceId` and `SpanId` fields, in this section of the documentation: https://doc.traefik.io/traefik/observability/access-logs/#limiting-the-fieldsincluding-headers

Could you please update the documentation accordingly? Thanks!

jaceknoga:
Hello @weijiany,
Sorry if it might be a kind of off topic question but this topic seems for me to be most appropriate as it's most advanced in matter documenting OpenTelemetry in Traefik v3 so far.

I'm using OpenTelemetry + OpenSearch for tracing for a while and in Traefik v3 there's no equivalent for `TraceId` header _stamp_. There was a solution in v2 where Jaeger Tracing was supported and then we had a `traceContextHeaderName` which could be stored in access logs, sent to client's browser via headers and targeted easily in tracing solution.

Coming back to subject – have You considered equivalent of `traceContextHeaderName` in OpenTelemetry for Traefik v3, extending what You've did here - I mean `TraceId` and `SpanId` in access logs. I miss the connection (bonding) between access logs and specific HTTP request which was possible previously in v2.

Of course, we have `Traceparent` header which is sent down with requests from router to bottom down service but I can't get it logged nor extracted (via middleware) from router to include it in headers sent to user. I saw a mention about using ForwardAuth to intercepting this header but I find it too complicated for this particular task as it's more a problem bypass not a solution.

@rtribotte – am I'm missing something or it's a current state on Traefik v3?

weijiany:
Hi @jaceknoga, thank you for sharing your concerns and the details on how you're planning to use OpenTelemetry with Traefik v3.

I don't want to implement a feature equivalent to traceContextHeaderName from v2. and I think approach using logging middleware or tools like Logstash, Fluentd, or Fluent Bit is indeed the recommended way to handle logging needs in Traefik v3. These tools are specifically designed to manage logs and can handle remapping of fields to include necessary tracing information, such as `TraceId` and `SpanId`.

So I adding a feature to record them into log while `otel` is enabled.

jaceknoga:
Hi @weijiany,
Thanks for Your point of view. So how You do connect exact query with exact trace otherwise? I'm using Fluentd / Fluent Bit also to ingest data to OpenSearch but how You can bond specific user request with specific trace without having some kind of shared pointer or HTTP header?

I think of a case when I have a report from user that some part of distributed "application" is not working properly. If You have a lot of traffic and lot of logs finding right one problematic request to start with consumes a lot of time. If I have a trace id header from user request I can find logs and traces regarding this particular case easily. Since otel we have it called `Traceparent` header by default, which is handled by request middleware and impossible to detect or operate with config defined middlewares.

I'd love to hear how do You deal with such case since I think we're touching a very similiar problem.

weijiany:
Hi @jaceknoga, In my project, we use Traefik as the kubernetes ingress controller. Traefik generates a trace and passes it to the backend services via the `traceparent` HTTP header. All internal service communications must carry the `traceparent` header. For each request, I include the `TraceId` in the response body, so users can report any error response using the `TraceId` to my support team. To query all logs related to a specific trace, I record both the `TraceId` and SpanId in the logs. This allows me to query all logs about a trace using the `TraceId`.

For example, Traefik will record the `TraceId` and `SpanId` once you enable the `otel` in this PR, while other internal services may use different field names for trace information. To make querying easier, it is necessary to remap these fields into your system's naming convention.

jaceknoga:
Hi, @weijiany. I'm running quite similar setup (except using Docker Swarm instead of Kubernetes) and I'm able to have request traced down even for 5-th or more service depth but I'm unable to get first `Traceparent` header, generated by Traefik, in access logs nor header in HTTP response sent to client. That is a main problem for my setup currently. All services below interact via `OTEL_PROPAGATORS: baggage,tracecontext` environment variable and it works as expected. 

Please clarify that by mentioning `TraceId` You mean default `Traceparent` request header from Traefik or do You add any other header with middleware? Maybe we're missing common point since I can't get this first `Traceparent` header logged in access logs nor in response headers.

So, summing up – I have stack trace for my services, step below request handled by Traefik but I can't "wire up" access log entry with browser request and stack trace collected via OpenTelemetry.

rtribotte:
@jaceknoga Propagation headers depend on the chosen/configured propagator, and `Traceparent` is the "default" that comes with tracecontext propagation.

This header, if initially created by Traefik (and not already present on the incoming request), is added to the request headers when the request is about to be forwarded to services.
It might be a feature/alternative behavior request to have the header added to the request before it passes through the middleware chain, and if you need this, I invite you to open an issue for that matter.

Also, my understanding is that the changes here are probably what you are looking for, regardless of the propagation model. The access log will be filled with the spanID (must be the Traefik main span) and traceID, thanks to the context, if tracing is enabled on Traefik.

jaceknoga:
@rtribotte Sure, I understand this and I'm aware. That's why I've mentioned that header passed down in requests to services and works perfectly.

At this moment I'd like just to note that `Traceparent` (I believe we don't have configuration possibility to change its name in Traefik v3?) is not accessible for middlewares. Otherwise my concept or PR from @weijiany wouldn't be needed since we can handle this purpose via configured or custom middlewares and log it accordingly. Am I right?

I've tried to insert some generated UUID header in Traefik middleware but it also doesn't work - there's no configuration option or am I missing something? It doesn't matter if I override header name with `Traceparent` either.

That behavior was different when we had Jaeger in v2 which added correlation header in response. Now we can't match "correlation header" with `Traceparent`?

rtribotte:
> At this moment I'd like just to note that Traceparent (I believe we don't have configuration possibility to change its name in Traefik v3?) is not accessible for middlewares. Otherwise my concept or PR from @weijiany wouldn't be needed since we can handle this purpose via configured or custom middlewares and log it accordingly. Am I right?

Yes, this is correct. But as I mentioned, being able to access propagation headers from within middleware could be feature request of its own.

> That behavior was different when we had Jaeger in v2 which added correlation header in response. Now we can't match "correlation header" with Traceparent?

Tracing has been reworked in v3, and doesn't rely on the same implementation as before.
If your need goes beyond access-logs, it would worth opening an feature request to have the header added to the request before it passes through the middleware chain.

weijiany:
Hi @rtribotte anything updates?

kevinpollet (approved):
Thanks 👍

Issue #10106, opened 2023-09-04
Add an option to fail startup if plugins fail to load


Desuuuu:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

It would be nice to have the option to have Traefik fail to startup if a plugin fails to load.

Currently, an error is logged and plugins are disabled which can lead to unreachable services if they require plugins to function properly.

The issue has been described in [this forum post](https://community.traefik.io/t/failed-to-download-plugin-caused-all-services-to-go-down/17717) where the user experienced a transient error downloading a plugin, requiring a manual restart to solve the issue.

nmengin:
Hello @Desuuuu,

Thank you for the suggestion.

We are unsure about the use case and the traction it will receive, so we are going to leave the status as "kind/proposal" to give the community time to let us know that they would like this.

We will reevaluate as people respond.

ldez:
> Is there a specific reason behind the current behaviour (logging an error and disabling plugins entirely)?

Yes: the goal is to be able to start Traefik if, for example, our plugin server is down.

> Because as I see it, it renders plugins unusable since any router using any plugin will be unavailable until the issue is fixed and Traefik is restarted.

It's the expected behavior.

This behavior has been added because of a failure of our plugin server: https://github.com/traefik/traefik/pull/9287

nmengin:
Hello there,

We are interested in this feature. To avoid breaking the current behavior, with the other maintainers, we think that makes sense to enable this feature using an option.

If a community member would like to build it, let us know, and we will work with you to ensure you have all the information needed to merge it.
We prefer to work with our community members at the beginning of the design process to ensure that we are aligned and can move quickly with the review and merge process. Let us know here or create a PR before you start, and we will work with you there.

Don’t forget to check out the[ contributor docs](https://github.com/traefik/contributors-guide/blob/master/pr_guidelines.md) and link the PR to this issue.

bmagic:
Hello @nmengin,

I have started something to achieve this (It's only to discuss the design, I haven't tested anything).
https://github.com/bmagic/traefik/commit/28739cbd7801071b81c917c96ac639aa2af9110b

So I've added a `Required` field in the plugin descriptor and for each plugin I'll check if it's required.   
[...]

PR #10929, opened 2024-07-24
Introduce startup fail option for critical plugin load failures

bmagic:
### What does this PR do?
This PR adds a Required field in the plugin descriptor. For each plugin, it checks if it is required. On the plugin client, it returns an error if the plugin is required and a serious issue occurs, or logs a warning if the plugin is not required. 
It's based on this issue : https://github.com/traefik/traefik/issues/10106

### Motivation
This change improves the robustness and resilience of the plugin management system. This modification allows setting certain plugins as required to ensure they are present at launch, providing greater control over critical plugin dependencies.

### More

- [ ] Added/updated tests

### Additional Notes
This implementation changes the default behavior, as currently, if a single plugin has a problem during download/check/unzip, no plugin is instantiated. With this change, all plugins that have been successfully downloaded/checked/unzipped will be activated.
If this PR is accepted, I will need to modify the Helm chart to add the required field in the deployment arguments.


bmagic:
I rebased my branch on master because tests failed. But it seems that tests also fail on master.

rtribotte:
Hello @bmagic,

Thanks for this contribution!

After some discussion among the maintainers, we agreed that having an option per plugin is not what we originally envisioned when we triaged #10106 as an enhancement.
While this was not stated clearly in the issue at the time, we prefer to introduce a static configuration option to control this behavior globally.
This would affect both local and remote plugins.

However, we are eager to get feedback on a per-plugin configuration and are open to revising our opinion on this. But for now, we do not see the benefit of a per-plugin option to address use case #10106.

Would you be inclined to revise the PR accordingly?

bmagic:
Hello, 

yes I will update this request, I don't have time right now but I should be able to propose something within a month.

PR #10935, opened 2024-07-25
Document Docker port selection on multiple exposed ports

mbrodala:
Traefik does not use the "first" exposed port since there is no such concept as a defined order of exposed ports. Instead Traefik collects all ports and uses the "lowest" / "smallest" of them.

See https://github.com/traefik/traefik/blob/87db3300d33fd5b74d793cf81d9dbcdba2092837/pkg/provider/docker/shared.go#L186-L197

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/providers/docker.md
+++ b/docs/content/providers/docker.md
@@ -117,12 +117,14 @@ When using Docker Compose, labels are specified by the directive
 
 Traefik retrieves the private IP and port of containers from the Docker API.
 
-Port detection works as follows:
+Port detection for private communication works as follows:
 
 - If a container [exposes](https://docs.docker.com/engine/reference/builder/#expose) a single port,
-  then Traefik uses this port for private communication.
+  then Traefik uses this port.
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hello @mbrodala and thanks for your contribution,

As this pull request is a bugfix could you rebase it on branch v2.11?

There is also a Port Detection section here https://doc.traefik.io/traefik/v2.11/providers/docker/#port-detection, which might need to be fixed as well, wdyt?

mbrodala:
I noticed that section but didn't change it since it currently does not even mention the port auto-selection. I can extend it accordingly but would rather not spread the same info in two locations ...

kevinpollet:
So maybe we could extend the section and in the note add something like `By default, Traefik uses the lowest exposed port of a container as detailed in the port selection section`, wdyt?

mbrodala:
OK, adjusted both locations accordingly now.

kevinpollet (approved):
Thanks :+1

rtribotte (approved):
Thanks 👍

PR #10940, opened 2024-07-26
Do not update route status when nothing changed

kevinpollet:
### What does this PR do?

This pull request improves the Gateway API routes status update by verifying whether the computed status matches the current status before making any changes. If the statuses are identical, no update is performed, thus preventing unnecessary changes to the `LastTransitionTime` when there are no actual modifications.

### Motivation

Fixes https://github.com/traefik/traefik/issues/10939

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/gateway/client.go
+++ b/pkg/provider/kubernetes/gateway/client.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"reflect"
 	"slices"
 	"time"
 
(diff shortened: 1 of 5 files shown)


juliens (approved):
LGTM

rtribotte (approved):
LGTM

pkg/provider/kubernetes/gateway/client.go
```diff
@@ -675,20 +683,20 @@ func translateNotFoundError(err error) (bool, error) {
 	return err == nil, err
 }
 
-func gatewayStatusEquals(statusA, statusB gatev1.GatewayStatus) bool {
+func gatewayStatusEqual(statusA, statusB gatev1.GatewayStatus) bool {
```
ilmax, line 686: @kevinpollet shouldn't this also check for addresses? I debugged the problem I was talking to you about [yesterday](https://github.com/traefik/traefik/pull/10667) down to this function. I debugged cloning the repo, building my image, and running Traefik with additional logs.

I ended up always having the gateway.statusAddress to null because this function avoids the update but the two gateway objects are not equal, one has the address set from the service/ip/hostname and the other does not. This function seems not to take into account that so it never updates the gateway with an address, breaking external-dns.

Can you please have a look?

kevinpollet: Hello @ilmax,

You are right, after checking the code we opened https://github.com/traefik/traefik/pull/10972 to fix it.

ilmax: Thank you for the quick fix @kevinpollet!

Issue #10953, opened 2024-07-30
WASM middleware state not kept


darkweak:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I'm actually looking at the WASM support for plugins in Træfik as @emilevauge mentioned but with this minimal snippet, we can see that the init is called when I send an HTTP request to Træfik

```go
package main

import (
	"github.com/http-wasm/http-wasm-guest-tinygo/handler"
	"github.com/http-wasm/http-wasm-guest-tinygo/handler/api"
)

func main() {
	handler.Host.Log(api.LogLevelWarn, "INIT!!!!!")

	handler.HandleRequestFn = func(r1 api.Request, r2 api.Response) (next bool, reqCtx uint32) {
		return true, 0
	}
}
```

I would like to know if that's the normal behavior or if that's a bug.  
IMHO loading WASM plugins as stateless and reconfigure them each time you go through the middleware is a performance issue.

### What did you see instead?

```
traefik-1  | 2024-07-30T21:25:08Z WRN github.com/traefik/traefik/v3/pkg/logs/wasm.go:31 > INIT!!!!! entryPointName=http middlewareName=souin@file middlewareType=wasm routerName=whoami@file
traefik-1  | 2024-07-30T21:25:08Z DBG github.com/traefik/traefik/v3/pkg/server/service/loadbalancer/wrr/wrr.go:196 > Service selected by WRR: b903963fe23550fa
traefik-1  | 192.168.65.1 - - [30/Jul/2024:21:25:08 +0000] "GET / HTTP/1.1" 200 373 "-" "-" 1 "whoami@file" "http://whoami" 9ms
traefik-1  | 2024-07-30T21:25:09Z WRN github.com/traefik/traefik/v3/pkg/logs/wasm.go:31 > INIT!!!!! entryPointName=http middlewareName=souin@file middlewareType=wasm routerName=whoami@file
traefik-1  | 2024-07-30T21:25:09Z DBG github.com/traefik/traefik/v3/pkg/server/service/loadbalancer/wrr/wrr.go:196 > Service selected by WRR: b903963fe23550fa
[...]

PR #10959, opened 2024-07-31
Draft: feat(plugins): share object between wasm instances

darkweak:
### What does this PR do?

This PR introduces a shared object to share data between WASM instances.  
It prevents the DB connection from being reinitialized or compute heavy object only once and reuse it in other instances. This can reduce the latency during the WASM plugin instantiation.

### Motivation

The discussion with @juliens on the issue #10953, the fact the pool re-instantiate the WASM plugins and recreate DB connections or create a new in-memory storage.

### More

- [x] ~~Added/updated tests~~

### DIFF
--- a/pkg/plugins/middlewarewasm.go
+++ b/pkg/plugins/middlewarewasm.go
@@ -18,6 +18,10 @@ import (
 	"github.com/traefik/traefik/v3/pkg/middlewares"
 )
 
+// WasmSharedObjects allows to share data between instances. It prevents the DB connection from being reinitialized or compute a
+// heavy object only once and reuse it in other instances. This can reduce the latency during the WASM plugin instantiation.
+var WasmSharedObjects = map[string]interface{}{}
+
 type wasmMiddlewareBuilder struct {
 	path     string
 	cache    wazero.CompilationCache


kevinpollet:
Hello @darkweak and thanks for your contribution,

As you discussed potential solutions in #10953, @juliens will handle this pull request.

juliens:
Hello @darkweak,

I'm not sure about the status of your pull request. Is it a draft?

Also, I wanted to let you know that sharing complex objects between Go and WebAssembly (WASM) is not possible. In WASM, you can only make calls with "int" values. For anything more complex, you need to use shared memory (like a []byte). This method can share strings or serialized objects, but it won't work for things like database connections.

ttys3:
I think this idea is good.

And I feel that the current wasm really needs such capabilities.

I developed https://github.com/ttys3/traefik-plugin-denyip/, which includes a multi-MB IP blacklist. If there were a read-only shared map, my plugin could save a lot of memory. In the current wasm plugin implementation, this plugin will consume a lot of memory because each connection essentially requires a separate allocated heap.

nmengin:
Hi @darkweak ,

Thanks a lot for your contribution!

Since there has been no activity on this PR for a while, we will close it for now to keep the queue manageable.
If you’d like to continue working on it or if you think we should revisit, please let us know, and we’ll reopen the PR for you. We’d be happy to pick up the discussion again when you’re ready.

Thanks again for your effort and interest in improving the project! :pray:

PR #10970, opened 2024-08-04
Add 30 day certificatesDuration step

luker983:
### What does this PR do?

Adds a new `certificatesDuration` threshold, setting renew period and renew interval to 10 days and 12 hours respectively when 30 days <= `certificatesDuration` < 90 days. 

### Motivation

There is a large gap between the 7 and 90 day `certificatesDuration` thresholds that has made Traefik difficult to adopt in environments that issue certificates with lifetimes shorter than 90 days.

With the current thresholds, 7-30 day cert lifetimes must either renew daily (straining the ACME CA), or only on the day before expiry (leaving little time to resolve an issue if renewal fails)

### More

- [x] Added/updated tests

### Additional Notes

Ideally the renewal values would be directly configurable as they are in most ACME clients, but I understand the desire for a simpler config. If maintainers are up for opening that discussion again I would prefer to go that route 😄

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -617,6 +617,7 @@ It defaults to `2160` (90 days) to follow Let's Encrypt certificates' duration.
 |----------------------|-------------------|-------------------------|
 | >= 1 year            | 4 months          | 1 week                  |
 | >= 90 days           | 30 days           | 1 day                   |
+| >= 30 days           | 10 days           | 12 hours                |
 | >= 7 days            | 1 day             | 1 hour                  |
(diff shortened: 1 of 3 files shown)


rtribotte:
Hello @luker983,

Thanks for opening this pull request!

We have marked thus as a `need-design` because we suspect it may introduce a breaking change for some users. 
Since no issue was opened beforehand to discuss the possibility of such a change, we would welcome any feedback from the community.

rtribotte (approved):
Thanks 👍

kevinpollet (approved):
Thanks 👍

PR #10974, opened 2024-08-06
Use Standard channel by default with Gateway API

mloiseleur:
### What does this PR do?

Document usage of Standard channel by default when using Gateway API.

### Motivation

Now that Gateway API support is complete, Standard channel can be used.
Experimental features like `TCPRoute` or `TLSRoute` can be enabled by installing experimental CRD and setting the experimental channel in the provider.

### More

### Additional Notes

Chart will also use Standard channel by default.


kevinpollet (approved):
Thanks 👍

rtribotte (approved):
LGTM

PR #10985, opened 2024-08-07
Allow configuring rule syntax with Kubernetes Ingress annotation

### LINKED ISSUE #10672: Traefik v3 breaks Kubernetes Ingress Prefix Path Regex
### Welcome!

### What did you do?

I upgraded from Traefik v2 to v3. Then I realized that all Kubernetes Ingress configurations which use a regex within as part of the prefix path (`pathType=Prefix`) break.

# Workaround

Add the following lines to the static config:

```yaml
core:
  defaultRuleSyntax: v2
```

# Suggestions

1. Revert to previous behavior for `pathType=Prefix`
2. Have an option for the Kubernetes Ingress provider to default to v2 behavior, by default
3. Have the Kubernetes Ingress provider to use `PathRegexp('^...')` insternally, instead
4. Add new `PathRegexp` value to `traefik.ingress.kubernetes.io/router.pathmatcher` annotation

### What did you see instead?

Kubernetes Ingress configurations which use a regex within as part of the Prefix path no longer work.

v3.0.0

```yaml
providers:
  kubernetesIngress:
    ingressClass: traefik
    allowEmptyServices: true
    throttleDuration: "3s"
```

Kubernetes Ingress Example:

```yaml
kind: Ingress
metadata:
  name: catalogservice-ingress
  namespace: uplift
  annotations:
    traefik.ingress.kubernetes.io/router.entrypoints: https
    traefik.ingress.kubernetes.io/router.tls: "true"
spec:
  ingressClassName: traefik
  rules:
    - host: 'api.xxx'
      http:
        paths:
          - path: /xxx/v1/yyyy/{customer:[a-z0-9]+}/push/
            pathType: Prefix
            backend:
              service:
                name: catalogservice-service
                port:
                  name: http
```

_No response_

rtribotte:
### What does this PR do?

This PR introduces a new annotation for Kubernetes Ingresses to be able to configure the rule syntax.

### Motivation

Fixes #10672

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/migration/v2-to-v3-details.md
+++ b/docs/content/migration/v2-to-v3-details.md
@@ -541,6 +541,19 @@ it is now unsupported and would prevent Traefik to start.
 
 All Pilot related configuration should be removed from the static configuration.
 
+### Kubernetes Ingress Path Matching
+
+In v3, the Kubernetes Ingress default path matching is strict and does not support regex syntax.
+
+#### Remediation
+
+Two level of remediation are possible:
+
+- Interpret the default path matcher `PathPrefix` with v2 syntax.
+It can done globally (for all routers) with the [static configuration](#configure-the-default-syntax-in-static-configuration) or per router by using the specific annotation: [traefik.ingress.kubernetes.io/router.rulesyntax](../routing/providers/kubernetes-ingress.md#annotations)
+
+- Adapt the path regular expression to be compatible the go regex syntax and change the default path matcher for the `PathRegexp` matcher with the annotation `traefik.ingress.kubernetes.io/router.pathmatcher`.
+
 ## Operations Changes
 
 ### Traefik RBAC Update
--- a/docs/content/routing/providers/kubernetes-ingress.md
+++ b/docs/content/routing/providers/kubernetes-ingress.md
@@ -229,10 +229,18 @@ which in turn will create the resulting routers, services, handlers, etc.
     traefik.ingress.kubernetes.io/router.priority: "42"
     ```
 
+??? info "`traefik.ingress.kubernetes.io/router.rulesyntax`"
+
+    See [rule syntax](../routers/index.md#rulesyntax) for more information.
+
+    ```yaml
+    traefik.ingress.kubernetes.io/router.rulesyntax: "v2"
+    ```
+
 ??? info "`traefik.ingress.kubernetes.io/router.pathmatcher`"
 
     Overrides the default router rule type used for a path.
-    Only path-related matcher name can be specified: `Path`, `PathPrefix`.
+    Only path-related matcher name should be specified: `Path`, `PathPrefix` or `PathRegexp`.
 
     Default `PathPrefix`
 
--- a/pkg/provider/kubernetes/ingress/annotations.go
+++ b/pkg/provider/kubernetes/ingress/annotations.go
@@ -26,6 +26,7 @@ type RouterIng struct {
 	EntryPoints []string                 `json:"entryPoints,omitempty"`
 	Middlewares []string                 `json:"middlewares,omitempty"`
 	Priority    int                      `json:"priority,omitempty"`
+	RuleSyntax  string                   `json:"ruleSyntax,omitempty"`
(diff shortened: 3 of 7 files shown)


docs/content/migration/v2-to-v3-details.md
```diff
@@ -541,6 +541,19 @@ it is now unsupported and would prevent Traefik to start.
...
+In v3, the Kubernetes Ingress default path matching does not support regexes anymore.
+
+#### Remediation
+
+Two level of remediation are possible:
+
+- Interpret the default path matcher `PathPrefix` with v2 syntax.
+It can done globally (for all routers) with the [static configuration](#configure-the-default-syntax-in-static-configuration) or per router by using the specific annotation: [traefik.ingress.kubernetes.io/router.rulesyntax](../routing/providers/kubernetes-ingress.md#annotations)
+
+- Adapt the path regular expression to be compatible the go regex syntax and change the default path matcher for the `PathRegexp` matcher with the annotation `traefik.ingress.kubernetes.io/router.pathmatcher`.
```
kevinpollet, line 555:
```suggestion
Two levels of remediation are possible:

- Interpret the default path matcher `PathPrefix` with v2 syntax.
This can done globally for all routers with the [static configuration](#configure-the-default-syntax-in-static-configuration) or on a per-router basis by using the [traefik.ingress.kubernetes.io/router.rulesyntax](../routing/providers/kubernetes-ingress.md#annotations) annotation.

- Adapt the path regex to be compatible with the Go regex syntax and change the default path matcher to use the `PathRegexp` matcher with the [`traefik.ingress.kubernetes.io/router.pathmatcher`](../routing/providers/kubernetes-ingress.md#annotations) annotation.
```

kevinpollet, line 555: Maybe we can change the pathmatcher annotation documentation to indicate that the `PathRegex` matcher can be used, here: https://doc.traefik.io/traefik/routing/providers/kubernetes-ingress/.

kevinpollet (approved):

docs/content/migration/v2-to-v3-details.md
rtribotte: I have modified the annotation documentation in that page already, I have added the mention of the `PathRegexp` matcher, don't you think it is enough?

hlascelles:
Thanks for this! We're waiting eagerly for this fix.

What follows is my [Googlewhack](https://en.wikipedia.org/wiki/Googlewhack) for [those that follow](https://xkcd.com/979/).

For those on Kubernetes upgrading from Traefik 2.x to Traefik 3.x, you may be finding your `Ingress` path Regexes no longer work. It is my understanding that this PR's change will mean you can set an annotation on the `Ingress` to fix it, but as of right now there hasn't been a new binary release yet, so the Ingress annotation won't work for you.

We have solved it using the command line switch spoken about as the workaround [here](https://github.com/traefik/traefik/issues/10672#issue-2272314209). 

In our case, we are using vanilla `Ingress` resources, not a traefik `IngressRoute`. eg:

```yaml
kind: Ingress
metadata:
  name: myapp-myapp
  annotations:
    traefik.frontend.rule.type: Path
spec:
  rules:
  - host: www.mycompany.com
    http:
      paths:
      - path: '/{regex:(somepathbit)/.*}'
        pathType: Prefix # This is a k8s keyword. It is not related to traefik.
        backend:
          service:
            name: myapp
            port:
              number: 80
```

This will match `https://www.mycompany.com/somepathbit/foo123`. Note the word `regex` in there means nothing, it is just a regex match name, I don't believe traefik uses it.

That worked on traefik 2.x. When on 3.x, you need to start traefik with `--core.defaultrulesyntax=v2`. You can do this with their helm chart by passing [in the value here](https://github.com/traefik/traefik-helm-chart/blob/f1cbfc5c046497854f1dbe13281459666fa1036e/traefik/values.yaml#L567).

With that all of our ingresses cluster wide began working again with no changes.

NB We're looking for more documentation on what the v3 change in regex style is ([others are not sure either](https://github.com/traefik/traefik/issues/10672#issuecomment-2091998030)). Can there be some more documentation on this please?

PR #10989, opened 2024-08-09
Prevent error logging when TCP WRR pool is empty

kevinpollet:
### What does this PR do?

This pull request addresses the unnecessary logging of an error when a TCP WRR pool is empty. In scenarios where the TCP WRR pool has no servers, logging an error is not appropriate since this is not indicative of an actual error condition. This also makes the behavior in line with the HTTP WRR.
### Motivation

To prevent error logging when TCP WRR pool is empty.

### More

- [ ] Added/updated tests


rtribotte (approved):
LGTM

nmengin (approved):
LGTM

PR #10997, opened 2024-08-12
Support for watching instead of polling Nomad

deverton-godaddy:
### What does this PR do?

This change modifies the Nomad provider to allow watching for service events to trigger reconfiguration.

Backwards compatibility is maintained by defaulting to polling as usual with a configurable interval. The new `watch` option can be set to `true` to enable event watching. When watching is enabled, the provider will listen for service events from the Nomad cluster and trigger reconfiguration as needed subject to the throttling setting.

The actual code change is modelled after how the Kubernetes provider handles event throttling and consumption.

### Motivation

The polling mechanism that the Nomad provider currently uses leads to longer latency for updates.

### Additional Notes

The throttling code and some structure is lifted from the Kubernetes provider. There may be a more elegant way to share the code but that seemed out of scope for this PR.

### DIFF
--- a/pkg/provider/nomad/nomad.go
+++ b/pkg/provider/nomad/nomad.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/hashicorp/nomad/api"
+	"github.com/mitchellh/hashstructure"
 	"github.com/rs/zerolog/log"
 	ptypes "github.com/traefik/paerser/types"
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
@@ -93,6 +94,8 @@ type Configuration struct {
 	ExposedByDefault   bool            `description:"Expose Nomad services by default." json:"exposedByDefault,omitempty" toml:"exposedByDefault,omitempty" yaml:"exposedByDefault,omitempty" export:"true"`
 	RefreshInterval    ptypes.Duration `description:"Interval for polling Nomad API." json:"refreshInterval,omitempty" toml:"refreshInterval,omitempty" yaml:"refreshInterval,omitempty" export:"true"`
 	AllowEmptyServices bool            `description:"Allow the creation of services without endpoints." json:"allowEmptyServices,omitempty" toml:"allowEmptyServices,omitempty" yaml:"allowEmptyServices,omitempty" export:"true"`
+	Watch              bool            `description:"Watch Nomad Service events." json:"watch,omitempty" toml:"watch,omitempty" yaml:"watch,omitempty" export:"true"`
+	ThrottleDuration   ptypes.Duration `description:"Service refresh throttle duration" json:"throttleDuration,omitempty" toml:"throttleDuration,omitempty" yaml:"throttleDuration,omitempty" export:"true"`
 }
 
 // SetDefaults sets the default values for the Nomad Traefik Provider Configuration.
@@ -118,6 +121,8 @@ func (c *Configuration) SetDefaults() {
 	c.RefreshInterval = ptypes.Duration(15 * time.Second)
 	c.DefaultRule = defaultTemplateRule
 	c.AllowEmptyServices = false
+	c.Watch = false
+	c.ThrottleDuration = ptypes.Duration(0)
 }
 
 type EndpointConfig struct {
@@ -139,6 +144,8 @@ type Provider struct {
 	namespace      string
 	client         *api.Client        // client for Nomad API
 	defaultRuleTpl *template.Template // default routing rule
+
+	lastConfiguration safe.Safe
 }
 
 // SetDefaults sets the default values for the Nomad Traefik Provider.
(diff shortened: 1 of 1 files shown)


pkg/provider/nomad/nomad.go
```diff
@@ -223,27 +256,56 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
...
+	return p.client.EventStream().Stream(ctx,
+		map[api.Topic][]string{
+			api.TopicService: {"*"},
+		},
+		0,
+		&api.QueryOptions{
+			Namespace: p.namespace,
+		},
+	)
+}
```
mmatur, line 291:
```suggestion
func (p *Provider) pollOrWatch(ctx context.Context) (<-chan *api.Events, error) {
	if p.Watch {
		return p.client.EventStream().Stream(ctx,
			map[api.Topic][]string{
				api.TopicService: {"*"},
			},
			0,
			&api.QueryOptions{
				Namespace: p.namespace,
			},
		)
	}

	serviceEventsChan := make(chan *api.Events, 1)

	go func() {
		ticker := time.NewTicker(time.Duration(p.RefreshInterval))
		defer ticker.Stop()

		for {
			select {
			case <-ctx.Done():
				return
			case t := <-ticker.C:
				serviceEventsChan <- &api.Events{
					Index: uint64(t.UnixNano()),
				}
			}
		}
	}()

	return serviceEventsChan, nil
}
```

pkg/provider/nomad/nomad.go
```diff
@@ -183,26 +190,52 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
 			ctx, cancel := context.WithCancel(ctxLog)
 			defer cancel()
 
-			// load initial configuration
-			if err := p.loadConfiguration(ctx, configurationChan); err != nil {
-				return fmt.Errorf("failed to load initial nomad services: %w", err)
+			serviceEventsChan, err := p.pollOrWatch(ctx)
+			if err != nil {
+				logger.Error().Err(err).Msg("Error watching nomad events")
+				return err
```
mmatur, line 196:
```suggestion
				return errors.New("error watching nomad events")
```

mmatur (changes requested):

pkg/provider/nomad/nomad.go
deverton-godaddy: Not logging the original error means we won't have any context to know why the watching failed. The pollOrWatch call is the first time the code attempts to connect to Nomad so it would be important to log enough context for diagnoses. We could wrap the error instead?

mmatur: Yes my bad, we have to wrap the error

mmatur (approved):
LGTM
Thanks for your contribution @deverton-godaddy  🎉

pkg/provider/nomad/nomad.go
```diff
@@ -93,6 +94,8 @@ type Configuration struct {
 	ExposedByDefault   bool            `description:"Expose Nomad services by default." json:"exposedByDefault,omitempty" toml:"exposedByDefault,omitempty" yaml:"exposedByDefault,omitempty" export:"true"`
 	RefreshInterval    ptypes.Duration `description:"Interval for polling Nomad API." json:"refreshInterval,omitempty" toml:"refreshInterval,omitempty" yaml:"refreshInterval,omitempty" export:"true"`
 	AllowEmptyServices bool            `description:"Allow the creation of services without endpoints." json:"allowEmptyServices,omitempty" toml:"allowEmptyServices,omitempty" yaml:"allowEmptyServices,omitempty" export:"true"`
+	Watch              bool            `description:"Watch Nomad Service events." json:"watch,omitempty" toml:"watch,omitempty" yaml:"watch,omitempty" export:"true"`
+	ThrottleDuration   ptypes.Duration `description:"Service refresh throttle duration" json:"throttleDuration,omitempty" toml:"throttleDuration,omitempty" yaml:"throttleDuration,omitempty" export:"true"`
```
kevinpollet, line 98: Do not miss to regenerate the documentation with the `go generate` command.

```suggestion
	ThrottleDuration   ptypes.Duration `description:"Service refresh throttle duration." json:"throttleDuration,omitempty" toml:"throttleDuration,omitempty" yaml:"throttleDuration,omitempty" export:"true"`
```

pkg/provider/nomad/nomad.go
```diff
@@ -183,26 +190,51 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
 			ctx, cancel := context.WithCancel(ctxLog)
 			defer cancel()
 
-			// load initial configuration
-			if err := p.loadConfiguration(ctx, configurationChan); err != nil {
-				return fmt.Errorf("failed to load initial nomad services: %w", err)
+			serviceEventsChan, err := p.pollOrWatch(ctx)
```
kevinpollet, line 193: In the polling, if I am not mistaken we will not load the initial configuration until the first tick, which was not the case previously.

pkg/provider/nomad/nomad.go
```diff
@@ -183,26 +190,51 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
...
+			if err != nil {
+				return fmt.Errorf("error watching nomad events: %w", err)
```
kevinpollet, line 195:
```suggestion
				return fmt.Errorf("error watching Nomad events: %w", err)
```

pkg/provider/nomad/nomad.go
```diff
@@ -223,27 +255,56 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
...
+			&api.QueryOptions{
+				Namespace: p.namespace,
+			},
+		)
+	}
+
+	serviceEventsChan := make(chan *api.Events, 1)
+
+	go func() {
+		ticker := time.NewTicker(time.Duration(p.RefreshInterval))
```
kevinpollet, line 274: Publishing an event before starting the ticker should solve the issue mentioned above.

pkg/provider/nomad/nomad.go
kevinpollet, line 98: Being able to configure a ThrottleDuration when polling sounds a bit strange. Maybe we could check that the throttle duration is always 0 when polling and return an error when that is not the case, and add a description in the doc? WDYT?

docs/content/reference/static-configuration/cli-ref.md
kevinpollet, line 1: Could you also add some documentation for added options in https://github.com/traefik/traefik/blob/master/docs/content/providers/nomad.md#provider-configuration?

kevinpollet (changes requested):

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #11001, opened 2024-08-13
Update acme.md

tired-engineer:
Fix the result of "Multiple DNS Challenge" section example.

### What does this PR do?

Adjusts documentation regarding Multiple DNS Challenge.

### Motivation

Currently, the documentation states that `_acme-challenge.example.org` CNAME'd to `challenge.example.com` will allow to issue certificate for `example.com` having access only to `example.org`, which is, I believe, is the other way around¹: the confirmation of the `example.org` domain control is being delegated with the CNAME to `example.com` where DNS TXT record can be manipulated. 

### More

- [ ] Added/updated tests
- [ x ] Added/updated documentation

### Additional Notes

¹ https://letsencrypt.org/docs/challenge-types/


kevinpollet (approved):
Thanks 👍

PR #11005, opened 2024-08-14
Improve Kubernetes CRDs

mloiseleur:
### What does this PR do?

This PR is fixing Kubernetes CRD on the following points:

- Provide _example_ value from doc.traefik.io in the CRD
- Provide _default_ value in the CRD
- Provide validation definition to check expected content: Enum, ~~UniqueItems~~, Pattern, Minimum or CEL validation expression when appropriate
- Provide printable columns for `IngressRoute`, `IngressRouteTCP`, `IngressRouteUDP`, `TLSOption` and `ServersTransport`
- Improve description with _deprecated_ when the field is not documented

### Motivation

Improve CRD help Traefik Proxy Kubernetes user when submitting or editing CR.

### More

- [ ] Added/updated tests

### Additional Notes

Documentation for CRD validation with kubebuilder is [here](https://book.kubebuilder.io/reference/markers/crd-validation.html).
It's hard to tell all possible case, so I rely on current integration test suite.
I'll need guidance if some new tests (related to CEL validation expression, for instance) needs to be written.

### DIFF
--- a/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
+++ b/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
@@ -45,9 +45,12 @@ spec:
                   Entry points have to be configured in the static configuration.
                   More info: https://doc.traefik.io/traefik/v2.11/routing/entrypoints/
                   Default: all.
+                example:
+                - web
                 items:
                   type: string
                 type: array
+                uniqueItems: true
               routes:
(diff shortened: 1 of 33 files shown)


nmengin:
Hello @mloiseleur,

We've kept the status to "need-triage" to allow us to check the PR and ensure we agree with the use case before moving forward.

mloiseleur:
This PR has been superseded by #11310 and #11311

PR #11015, opened 2024-08-19
Give valid examples for exposing dashboard with default Helm values

holysoles:
### What does this PR do?

Updates the Helm installation documentation to clarify how to enable and expose the dashboard after [recent changes](https://github.com/traefik/traefik-helm-chart/pull/1111).

### Motivation

I was trying to setup Traefik with Helm and ran into issues attempting to follow the existing documentation to access the dashboard.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/getting-started/install-traefik.md
+++ b/docs/content/getting-started/install-traefik.md
@@ -99,38 +99,6 @@ helm install traefik traefik/traefik
       - "--log.level=DEBUG"
     ```
 
-### Exposing the Traefik dashboard
-
-This Helm chart does not expose the Traefik dashboard by default, for security concerns.
-Thus, there are multiple ways to expose the dashboard.
-For instance, the dashboard access could be achieved through a port-forward:
-
-```shell
-kubectl port-forward $(kubectl get pods --selector "app.kubernetes.io/name=traefik" --output=name) 9000:9000
-```
-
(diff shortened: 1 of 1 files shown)


mloiseleur:
\o that's right, thanks :+1: 
I suggest to either (a) remove entirely this section or (b) provide a link [to the examples about dashboard exposition](https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#access-traefik-dashboard-without-exposing-it)

holysoles:
> \o that's right, thanks 👍 I suggest to either (a) remove entirely this section or (b) provide a link [to the examples about dashboard exposition](https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#access-traefik-dashboard-without-exposing-it)

I think just removing the documentation for this might be reasonable, the rest of the page is really focused on true installation steps rather than any initial configuration. We also already have a link to that examples.MD just above this section [here](https://github.com/traefik/traefik/blob/8cf506782ae6fad365142cf976561b5a3088d3c3/docs/content/getting-started/install-traefik.md?plain=1#L67) that I think is a natural place to gravitate towards for Helm configuration.

holysoles:
Updated based on my last comment, happy to squash commits if that approach is preferred

rtribotte (approved):
Thanks!

Issue #11021, opened 2024-08-19
Add support for UDP routing in systemd socket activation


tsiid:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

**Description:**

Traefik currently provides excellent support for systemd socket activation (added in #9891) enabling seamless integration with system services and allowing Traefik to be activated on demand. However, the current implementation only supports TCP sockets. Given that many services rely on UDP for communication, it would be highly beneficial if Traefik could also support UDP routing when using systemd socket activation

PR #11022, opened 2024-08-19
Add support for UDP routing in systemd socket activation

tsiid:
### What does this PR do?

This PR adds support for UDP routing in Traefik’s systemd socket activation feature. With this enhancement, Traefik can now route UDP traffic in addition to TCP when called from systemd

### Motivation

Many services rely on UDP for communication #11021

### More

- [ ] Added/updated tests

### Additional Notes

Example:

1. Create UDP socket: _~/.config/systemd/user/udp.socket_

```
[Socket]
ListenDatagram=0.0.0.0:5555
FileDescriptorName=udp
Service=traefik.service

[Install]
WantedBy=sockets.target
```

2. Create traefik service: _~/.config/systemd/user/traefik.service_

```
[Unit]
Wants=network-online.target
After=network-online.target
After=udp.socket
RequiresMountsFor=/run/user/1000/podman/podman.sock

[Service]
KillMode=mixed
Environment=PODMAN_SYSTEMD_UNIT=%n
ExecStop=/usr/bin/podman rm -v -f -i --cidfile=/run/user/1000/%N.cid
ExecStopPost=-/usr/bin/podman rm -v -f -i --cidfile=/run/user/1000/%N.cid
Delegate=yes
Type=notify
NotifyAccess=all
SyslogIdentifier=%N
ExecStartPre=/usr/bin/podman network create --ignore --internal net
#For podman below 4.4.0 use:
#ExecStartPre=-/usr/bin/podman network create --internal net
ExecStart=/usr/bin/podman run \
        --cidfile=/run/user/1000/%N.cid \
        --name=traefik \
        --replace \
        --rm \
        --network=net \
        -d \
        --security-opt label=type:container_runtime_t \
        -v /run/user/1000/podman/podman.sock:/var/run/docker.sock \
        localhost/traefik:latest \
        --log.level=DEBUG \
        --providers.docker \
        --providers.docker.exposedbydefault=false \
        --entrypoints.udp.address=/udp
        
[Install]
WantedBy=default.target
```

3. Start services:
```
systemctl --user daemon-reload && \
systemctl --user start udp.socket && \
systemctl --user start traefik.service
```

4. Launch backend container listening UDP:

For example we build alpine with netcat
```
FROM alpine:latest
RUN apk add --no-cache netcat-openbsd
ENTRYPOINT [ "nc" ]
```
and start container via podman compose file
```
services:
  nc-udp:
    image: nc:latest
    container_name: nc-udp
    networks:
      - nc
    labels:
      - "traefik.enable=true"
      - "traefik.docker.network=nc"
      - "traefik.udp.routers.nc-udp.entrypoints=udp"
      - "traefik.udp.routers.nc-udp.service=nc-udp-svc"
      - "traefik.udp.services.nc-udp-svc.loadbalancer.server.port=5555"
    command: 
      - "-vvul"
      - "-p"
      - "5555"

networks:
  nc:
    external: true
    name: nc
```

5. Send message via UDP: 
```
echo "hello" | nc -vu 127.0.0.1 5555
```

### DIFF
--- a/docs/content/routing/entrypoints.md
+++ b/docs/content/routing/entrypoints.md
@@ -1180,7 +1180,7 @@ entryPoints:
 
 Traefik supports [systemd socket activation](https://www.freedesktop.org/software/systemd/man/latest/systemd-socket-activate.html).
 
-When a socket activation file descriptor name matches an EntryPoint name, the corresponding file descriptor will be used as the TCP listener for the matching EntryPoint.
+When a socket activation file descriptor name matches an EntryPoint name, the corresponding file descriptor will be used as the TCP/UDP listener for the matching EntryPoint.
 
 ```bash
 systemd-socket-activate -l 80 -l 443 --fdname web:websecure  ./traefik --entrypoints.web --entrypoints.websecure
@@ -1188,11 +1188,7 @@ systemd-socket-activate -l 80 -l 443 --fdname web:websecure  ./traefik --entrypo
 
 !!! warning "EntryPoint Address"
 
-    When a socket activation file descriptor name matches an EntryPoint name its address configuration is ignored.     
-
-!!! warning "TCP Only"
-
-    Socket activation is not yet supported with UDP entryPoints.
+    When a socket activation file descriptor name matches an EntryPoint name its address configuration is ignored. For support UDP routing, address must have /udp suffix (--entrypoints.my-udp-entrypoint.address=/udp)
 
 !!! warning "Docker Support"
 
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -48,14 +48,7 @@ const (
 var (
 	clientConnectionStates   = map[string]*connState{}
 	clientConnectionStatesMu = sync.RWMutex{}
-
-	socketActivationListeners map[string]net.Listener
 )
-
-func init() {
-	// Populates pre-defined socketActivationListeners by socket activation.
-	populateSocketActivationListeners()
-}
 
 type connState struct {
 	State            string
--- a/pkg/server/server_entrypoint_udp.go
+++ b/pkg/server/server_entrypoint_udp.go
@@ -16,9 +16,9 @@ import (
 type UDPEntryPoints map[string]*UDPEntryPoint
 
 // NewUDPEntryPoints returns all the UDP entry points, keyed by name.
-func NewUDPEntryPoints(cfg static.EntryPoints) (UDPEntryPoints, error) {
+func NewUDPEntryPoints(config static.EntryPoints) (UDPEntryPoints, error) {
 	entryPoints := make(UDPEntryPoints)
-	for entryPointName, entryPoint := range cfg {
+	for entryPointName, entryPoint := range config {
 		protocol, err := entryPoint.GetProtocol()
 		if err != nil {
 			return nil, fmt.Errorf("error while building entryPoint %s: %w", entryPointName, err)
@@ -28,7 +28,7 @@ func NewUDPEntryPoints(cfg static.EntryPoints) (UDPEntryPoints, error) {
 			continue
 		}
 
-		ep, err := NewUDPEntryPoint(entryPoint)
+		ep, err := NewUDPEntryPoint(entryPoint, entryPointName)
 		if err != nil {
 			return nil, fmt.Errorf("error while building entryPoint %s: %w", entryPointName, err)
 		}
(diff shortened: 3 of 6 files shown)


ignisf:
@tsiid, an eagerly awaited feature, thank you for contributing it!

I have a question -- given the example for a combined TLS HTTP1/2/3 endpoint[1], how would one go about to set up a systemd socket for it?

Would adding a `ListenDatagram` directive within the systemd socket file specifying the endpoint descriptor suffice?

```
[Socket]
ListenStream=[::]:443
ListenDatagram=[::]:443
FileDescriptorName=websecure
Service=traefik.service

[Install]
WantedBy=sockets.target
```

[1]:
```yaml
entryPoints:
  web:
    address: ":80"
    http:
      redirections:
        entryPoint:
          to: websecure
          scheme: https
  websecure:
    address: :443
    http3: {}
```

tsiid:
@ignisf, thank you for your interest!

Currently, it's not possible to configure a TCP/HTTP endpoint together with a UDP endpoint in the same systemd socket file. The reason is that both would share the same FileDescriptorName, and for now, Traefik does not allow configuring two endpoints with the same name.

As a workaround, you can create two separate systemd socket files: one for HTTP and another for UDP. For example:

HTTP:
```
[Socket]
ListenStream=[::]:443
FileDescriptorName=websecure
Service=traefik.service

[Install]
WantedBy=sockets.target
```

UDP:
```
[Socket]
ListenDatagram=[::]:443
FileDescriptorName=websecure-udp
Service=traefik.service

[Install]
WantedBy=sockets.target
```

Additionally, note that in your Traefik configuration, you should also listen on port 80 to enable HTTP redirection to HTTPS. For that, you'll need to create another systemd socket file for port 80:
```
[Socket]
ListenStream=[::]:80
FileDescriptorName=web
Service=traefik.service

[Install]
WantedBy=sockets.target
```

Then, in your Traefik configuration, add the UDP endpoint:
```
entryPoints:
  web:
    address: ":80"
    http:
      redirections:
        entryPoint:
          to: websecure
          scheme: https
  websecure:
    address: ":443"
    http3: {}
  websecure-udp:
    address: ":443/udp"
```

When I tested similar solution locally, using multiple socket files for the same port with TCP/UDP worked without any problems.

ignisf:
I can't get your example above to work. It fails with:

```
2024-09-19T21:08:45Z ERR src/cmd/traefik/traefik.go:83 > Command error error="command traefik error: error while building entryPoint websecure-udp: error creating listener: listen packet: listen udp :443: bind: address already in use"
```

ignisf:
@tsiid, also, correct me if I'm wrong, but wouldn't replicating the `NewUDPEntryPoint` here achieve the sought after effect:

https://github.com/traefik/traefik/blob/a967d764805f7caa12804967f42741ecf72967a7/pkg/server/server_entrypoint_tcp_http3.go#L37-L41

ignisf:
This seem to work:

```suggestion
diff --git a/pkg/server/server_entrypoint_tcp_http3.go b/pkg/server/server_entrypoint_tcp_http3.go
index 23b23071d..db7e3df1f 100644
--- a/pkg/server/server_entrypoint_tcp_http3.go
+++ b/pkg/server/server_entrypoint_tcp_http3.go
@@ -25,7 +25,7 @@ type http3server struct {
        getter func(info *tls.ClientHelloInfo) (*tls.Config, error)
 }
 
-func newHTTP3Server(ctx context.Context, configuration *static.EntryPoint, httpsServer *httpServer) (*http3server, error) {
+func newHTTP3Server(ctx context.Context, name string, configuration *static.EntryPoint, httpsServer *httpServer) (*http3server, error) {
        if configuration.HTTP3 == nil {
                return nil, nil
        }
@@ -34,10 +34,16 @@ func newHTTP3Server(ctx context.Context, configuration *static.EntryPoint, https
                return nil, errors.New("advertised port must be greater than or equal to zero")
        }
 
-       listenConfig := newListenConfig(configuration)
-       conn, err := listenConfig.ListenPacket(ctx, "udp", configuration.GetAddress())
-       if err != nil {
-               return nil, fmt.Errorf("starting listener: %w", err)
+       var conn net.PacketConn
+
+       if socketActivatedConn, ok := socketActivationPacketConns[name]; ok {
+               conn = socketActivatedConn
+       } else {
+               listenConfig := newListenConfig(configuration)
+               conn, err = listenConfig.ListenPacket(ctx, "udp", configuration.GetAddress())
+               if err != nil {
+                       return nil, fmt.Errorf("starting listener: %w", err)
+               }
        }
 
        h3 := &http3server{
```

Pardon me, not a Golang developer

tsiid:
@ignisf You are right. I didn't focus on the fact that it's about HTTP/3. Indeed, HTTP/3 requires an additional adjustment, and for it, both ListenStream and ListenDatagram can be specified in a single socket file. I've updated the pull request to support HTTP/3.
Thank you!

youkoulayley (changes requested):
Thanks for your contribution @tsiid!
The PR looks good to me and works great. I just have a minor suggestion on the logs; we prefer to use the `.Err()` instead of wrapping the error in the message.

This PR also breaks the Windows build. To fix that I would move back the `init()` function to the `server_entrypoints_tcp.go` file along with the definition of `SocketActivation` and related methods. Finally, the `populateSocketActivationListeners` function should follow the same signature as the unix one and return an empty SocketActivation.

tsiid:
@youkoulayley thank you for your review and suggestions!

I've updated the logging to use .Err() as requested.
For the Windows build, I went with a slightly different approach. Instead of moving the init() function back to the server_entrypoints_tcp.go file, I've placed the SocketActivation declaration and related methods in a separate socket_activation.go file. I believe this keeps the code more organized, as it avoids mixing UDP socket activation logic in a file focused on TCP. Let me know if you have any concerns with this approach.

youkoulayley (approved):
LGTM

ignisf:
@youkoulayley this PR still seems to have the needs-review tag even though you reviewed it. Is this expected?

youkoulayley:
@ignisf yes, each PR needs 2 reviews from maintainers to be merged. (cf. https://doc.traefik.io/traefik/contributing/submitting-pull-requests/#the-pull-request-submit-process)

kaaax0815:
Thank you very much for this pr.
Will it be included in the next release? maybe 3.3.4?

Issue #11024, opened 2024-08-20
Support HTTP mirroring excluding body


MatteoPaier:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Often when mirroring HTTP request there is no need for the entire body of the request (e.g. only need the headers for logging or audit purposes).

As far as I know, when HTTP mirroring was introduced there was no support for body mirroring. When #6080 was merged, body mirroring became mandatory (with the option to omit the request if the body exceeds a specified size).

I propose a new boolean http/mirroring option (e.g. `mirror-body`, default `true`) to allow for the removal of the body from the mirrored requests. I can submit a PR if the feature could prove useful.

nmengin:
Hey @MatteoPaier,

Thank you for the suggestion.

> I can submit a PR if the feature could prove useful.

Be sure that if you open a PR, we'll evaluate it.
Thanks in advance.

nmengin:
Hello @MatteoPaier,

Thank you for your contribution.

Discussing again the issue with the other maintainers, we are wondering if the mirroring feature should be used to achieve this kind of use case.

Indeed, in our understanding, the primary goal of the mirroring is to replicate the same requests on alternative servers and we are wondering if a request alteration should be allowed.

Could you elaborate on your use case and explain why you cannot use the observability features provided by Traefik?

MatteoPaier:
Hi @nmengin,

My use case at the moment is to duplicate requests to get the value of some headers and take decisions based on that values. I am migrating a stack from Nginx, where the ngx_http_mirror_module (with mirror_request_body = off) was used to mirror the headers of a service to another for auditing purposes.

At the moment I'm simply mirroring the requests and ignoring the body, but this can generate a lot of unnecessary traffic.

I've looked into Traefik's access logs observability, but I feel like it does not completely apply to my use case because I need this mirroring only for some specific services, while access logs settings apply globally.
Moreover, I don't thing modifying access logs to bind them to services (i.e. enable the user to specify different options for different services) is a solution here: I think access logs are better suited for observability (indeed), while in my use case the requests headers are used also to influence the state of the application.

Maybe also body truncation up to a certain length can be useful to someone, though I don't have a real use case in mind.

Additionally, in #5893 was argued the opposite:
[...]

PR #11032, opened 2024-08-22
Add mirrorBody option to HTTP mirroring

MatteoPaier:
### What does this PR do?

Implements a new option `mirrorBody` for HTTP mirroring (default `true`) that allows for the exclusion of the request body from the mirrored request.

### Motivation

Often when mirroring HTTP request there is no need for the entire body of the request (e.g. only need the headers for logging or audit purposes).

Solves #11024.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/routing/services/index.md
+++ b/docs/content/routing/services/index.md
@@ -1207,6 +1207,7 @@ http:
 The mirroring is able to mirror requests sent to a service to other services.
 Please note that by default the whole request is buffered in memory while it is being mirrored.
 See the maxBodySize option in the example below for how to modify this behaviour.
+You can also omit the request body by setting the mirrorBody option to `false`.
 
 !!! info "Supported Providers"
 
@@ -1219,6 +1220,9 @@ http:
     mirrored-api:
       mirroring:
         service: appv1
+        # mirrorBody defines whether the request body should be mirrored.
+        # Default value is true.
+        mirrorBody: true
         # maxBodySize is the maximum size allowed for the body of the request.
         # If the body is larger, the request is not mirrored.
         # Default value is -1, which means unlimited size.
@@ -1248,6 +1252,9 @@ http:
       # If the body is larger, the request is not mirrored.
       # Default value is -1, which means unlimited size.
       maxBodySize = 1024
+      # mirrorBody defines whether the request body should be mirrored.
+      # Default value is true.
+      mirrorBody = true
     [[http.services.mirrored-api.mirroring.mirrors]]
       name = "appv2"
       percent = 10
--- a/integration/simple_test.go
+++ b/integration/simple_test.go
@@ -1106,6 +1106,95 @@ func (s *SimpleSuite) TestMirrorWithBody() {
 	assert.Equal(s.T(), int32(0), val2)
 }
 
+func (s *SimpleSuite) TestMirrorWithoutBody() {
+	var count, countMirror1, countMirror2 int32
+
+	body := make([]byte, 20)
+	_, err := rand.Read(body)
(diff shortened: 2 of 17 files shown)


MatteoPaier:
I rebased and fixed some formatting inconsistencies and CRD autogenerated code. Now the validate and validate-generate checks are passing.

I don't understand why now the integration test is failing when it was passing before, as neither the code nor the test was changed. It seems the problem is related to the mirror `percent` option, but I have no idea how to fix it.

mmatur:
Hi @MatteoPaier,

Thanks for the rebase and fixes.

It seems your change impact some integration tests as you can see [here](https://github.com/traefik/traefik/actions/runs/10616173876/job/29426309537?pr=11032#step:7:433)

MatteoPaier:
Hi @mmatur,

I've seen the logs. I can confirm that when I run the test on my machine that particular test is passing. Also [here](https://github.com/traefik/traefik/actions/runs/10615798417/job/29424972917) the test was passing, so the failure seems very strange.

Is there some nondeterminism involved with the `percent` option of HTTP mirroring? Because it seems the mirrored service got 4 requests out of 10 even if the percentage is set to 50%

MatteoPaier:
I managed to make it fail also locally (with the same error), after repeating the same test 238 times. Out of 500 iterations of the test I've ran, only one failed.

Could you try re-triggering the integration check to verify that this is indeed a nondeterministic problem?

I'm searching for the core issue, but I feel like it should be addressed in a separate PR (if identified).

mmatur:
@MatteoPaier,

integration tests have been restarted and it is working now probably a flaky test. We keep in mind that problem and we will have a look.

Thanks.

docs/content/routing/services/index.md
```diff
@@ -1219,6 +1220,9 @@ http:
     mirrored-api:
       mirroring:
         service: appv1
+        # mirrorBody defines whether the request body should be mirrored.
+        # Default value is true.
+        mirrorBody: true
```
rtribotte, line 1225:
```suggestion
        mirrorBody: false
```

docs/content/routing/services/index.md
```diff
@@ -1248,6 +1252,9 @@ http:
       # If the body is larger, the request is not mirrored.
       # Default value is -1, which means unlimited size.
       maxBodySize = 1024
+      # mirrorBody defines whether the request body should be mirrored.
+      # Default value is true.
+      mirrorBody = true
```
rtribotte, line 1257:
```suggestion
      mirrorBody = false
```

integration/simple_test.go
```diff
@@ -1106,6 +1106,95 @@ func (s *SimpleSuite) TestMirrorWithBody() {
 	assert.Equal(s.T(), int32(0), val2)
 }
 
+func (s *SimpleSuite) TestMirrorWithoutBody() {
```
rtribotte, line 1109: I thin it is better to rename and adapt TestMirrorWithBody, to check in addition the new router without body, could you please merge the two tests?

rtribotte (commented):
In the documentation, it is probably better to document the opposite of the default value.
Otherwise, LGTM, Thanks!

integration/simple_test.go
MatteoPaier: I merged the tests, let me know if there is something else to do.
I've not renamed the test, since I did not came up with a better name (my modifications also relate to body).

rtribotte: Indeed, this is ok with the same name, thanks for the rework!

rtribotte (approved):
Thanks!

mmatur (approved):
LGTM

PR #11034, opened 2024-08-23
Provide a complete k8s example on error middleware

mloiseleur:
### What does this PR do?

Provide a complete k8s example on error middleware

### Motivation

`whoami` is not tailored to return error pages.
https://github.com/tarampampam/error-pages is dedicated to this usage.
They provide [an example with Docker](https://github.com/tarampampam/error-pages?tab=readme-ov-file#http-server-starting-utilizing-either-a-binary-file-or-docker-image), not for Kubernetes.

### More

### DIFF
--- a/docs/content/middlewares/http/errorpages.md
+++ b/docs/content/middlewares/http/errorpages.md
@@ -27,21 +27,55 @@ labels:
 ```
 
 ```yaml tab="Kubernetes"
+---
 apiVersion: traefik.io/v1alpha1
 kind: Middleware
 metadata:
-  name: test-errors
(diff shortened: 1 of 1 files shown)


nmengin:
Hello @mloiseleur,

After discussing the example with the other maintainers, we have decided to close your PR.

Indeed:

- So far, in Traefik documentation we've had the stance to not provide full examples but only Traefik configuration
- We do not want to promote an image we are not maintaining

However, improving the image [`traefik/whoami`](https://github.com/traefik/whoami) to provide error pages could be a good enhancement.

PR #11084, opened 2024-09-13
Use correct default weight in Accept-Encoding

michelheusschen:
### What does this PR do?

Support for weights in the `Accept-Encoding` header was added to the compression middleware in #10777. However, the default weight is set to `0` while [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-quality-values) says: If no "q" parameter is present, the default weight is 1.

This leads to inconsistent selection of the content encoding, see this example:
| Accept-Encoding | v3.1.2 | This PR |
| - | - | - |
| gzip, br | br | br |
| gzip, br;q=0.9 | br | gzip |

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/compress/acceptencoding.go
+++ b/pkg/middlewares/compress/acceptencoding.go
@@ -1,6 +1,7 @@
 package compress
 
 import (
+	"cmp"
 	"slices"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


mmatur (approved):
LGTM
Thanks @michelheusschen 👍🏼

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #11086, opened 2024-09-14
Add tsnet so traefik can create tailscale interfaces.

JordanDeBeer:
### What does this PR do?

This PR  adds support for creating Tailscale interfaces / entrpoints via [tsnet](https://tailscale.com/kb/1244/tsnet).

### Motivation

I currently run a couple different services on a single VM, and I want traefik to create a tailscale interface for each. This allows me to have FQDN routed domain as well as take advantage of everything traefik has to offer for my tailscale services.

### More

- [ ] Added/updated tests

### Additional Notes

I managed to get this working in a few lines of code, but would like some feedback on if this kind of feature is welcome in Traefik upstream before making it more robust.

I tried to implement this via a provider or plugin, but there were a ton of challenges and changes that would have been required. It was much easier to integrate it into the Listener code as tsnet implements a `net.Listener` interface.

There are a few things that I would imagine need to be added before merged like customizing the tsnet server, passing in auth via the config, and more.


svenvg93:
@juliens Any update on this? Would love you use treafik for this. Amazing work! 

https://github.com/almeidapaulopt/tsdproxy might give you some inspirations as well 😉

juliens:
Thanks for this PR! After reviewing it with the team, we think it would be interesting to integrate it. Are you still willing to work on it? Do you think it’s ready for review? If so, could you move it out of draft?

JordanDeBeer:
Would love to rebase and add it for consideration. I'm actually out of the country this week, but happy to rebase and update.

PR #11088, opened 2024-09-16
Removes goexport dependency and adds _initialize

juliens:
### What does this PR do?

This PR fixes a license issue on the dependency https://github.com/juliens/wasm-goexport by removing it.
To handle the new wasmexport in go, we need to add `_initialize` in the startFunctions.
 
### Motivation

Because wasmexport is finally merged into the master branch of go, we don't need the wasmexport hack anymore.


mmatur (approved):
LGTM

rtribotte (approved):
LGTM

PR #11100, opened 2024-09-18
feat: Add ingress status for ClusterIP and NodePort Service Type.

mlec1:
### What does this PR do?

This pull request adds ingress status for ClusterIP and NodePort Service Type.

I modified a little bit but took 

Refs: traefik/traefik#7972

### Motivation

It is pretty useful for deployment with ArgoCD for example which check the health by checking the status of an Ingress for its health checks.

I modified a little bit but took the main code from the ticket below

See: Refs: traefik/traefik#7972

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -402,19 +402,49 @@ func (p *Provider) updateIngressStatus(ing *netv1.Ingress, k8sClient Client) err
 		return fmt.Errorf("cannot get service %s, received error: %w", p.IngressEndpoint.PublishedService, err)
 	}
 
-	if exists && service.Status.LoadBalancer.Ingress == nil {
-		// service exists, but has no Load Balancer status
-		log.Debug().Msgf("Skipping updating Ingress %s/%s due to service %s having no status set", ing.Namespace, ing.Name, p.IngressEndpoint.PublishedService)
-		return nil
-	}
-
 	if !exists {
 		return fmt.Errorf("missing service: %s", p.IngressEndpoint.PublishedService)
 	}
 
-	ingresses, err := convertSlice[netv1.IngressLoadBalancerIngress](service.Status.LoadBalancer.Ingress)
-	if err != nil {
-		return err
+	var ingresses []netv1.IngressLoadBalancerIngress
+
+	switch service.Spec.Type {
+	case corev1.ServiceTypeLoadBalancer:
+		// Use the Load Balancer status
+		if service.Status.LoadBalancer.Ingress == nil {
+			// service exists, but has no Load Balancer status
+			log.Debug().Msgf("Skipping updating Ingress %s/%s due to service %s having no status set", ing.Namespace, ing.Name, p.IngressEndpoint.PublishedService)
+			return nil
+		}
+
+		ingresses, err = convertSlice[netv1.IngressLoadBalancerIngress](service.Status.LoadBalancer.Ingress)
+		if err != nil {
+			return err
+		}
+
+	case corev1.ServiceTypeClusterIP:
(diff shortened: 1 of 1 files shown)


jemand771:
first of all, huge thanks for submitting this!

I have a small concern but I'm not sure if it will even be a problem: What happens when people use a `NodePort` service, but with an external LoadBalancer pointed to _some_, but not all of the nodes' external IPs? Could this cause problems since the list of node IPs is technically wrong?

Again, no idea if this is even a realistic problem, and I'm not trying to hold this up in any way - it just came to mind while thinking about a setup I'm currently tinkering with where argocd is acting up. I think this PR could help with that

Finally, ref: https://github.com/argoproj/argo-cd/issues/14607

mlec1:
> first of all, huge thanks for submitting this!
> 
> I have a small concern but I'm not sure if it will even be a problem: What happens when people use a `NodePort` service, but with an external LoadBalancer pointed to _some_, but not all of the nodes' external IPs? Could this cause problems since the list of node IPs is technically wrong?
> 
> Again, no idea if this is even a realistic problem, and I'm not trying to hold this up in any way - it just came to mind while thinking about a setup I'm currently tinkering with where argocd is acting up. I think this PR could help with that
> 
> Finally, ref: [argoproj/argo-cd#14607](https://github.com/argoproj/argo-cd/issues/14607)

Concerning ArgoCD, for the health of Ingress, it expects just something so it won't be an issue here: https://argo-cd.readthedocs.io/en/stable/operator-manual/health/#ingress

According to the doc: https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport, 

"_Every node in the cluster configures itself to listen on that assigned port and to forward traffic to one of the ready endpoints associated with that Service. You'll be able to contact the type: NodePort Service, from outside the cluster, by connecting to any node using the appropriate protocol (for example: TCP), and the appropriate port (as assigned to that Service)._"

So in fact all nodes listen, then it is your choice to use only some of them. (that's what I understand, feel free to correct me if I am wrong).

jemand771:
ah, yeah, that sounds reasonable. looks like my idea really was just a _very_ hypothetical (if not impossible) scenario that doesn't matter for argocd either way. thanks for clarifying :)

artuross:
Hey, thanks for submitting this PR. An issue that I see with this PR (which I tried to solve in #11242) is that Service of type ClusterIP with **only** ClusterIPs set, won't be properly handled. Here, ClusterIP is properly handled only if the ExternalIP is also set.

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -402,19 +402,49 @@ func (p *Provider) updateIngressStatus(ing *netv1.Ingress, k8sClient Client) err
...
+		for _, node := range nodes {
+			for _, address := range node.Status.Addresses {
+				if address.Type == corev1.NodeExternalIP {
+					ingresses = append(ingresses, netv1.IngressLoadBalancerIngress{IP: address.Address})
+				}
+			}
+		}
+
+	default:
+		return fmt.Errorf("unsupported service type: %s", service.Spec.Type)
```
youkoulayley, line 447: I'm not sure we should return an error here.
I think we should add a debug log like we do for the empty loadbalancer and return nil.

youkoulayley (dismissed):

nmengin:
Hello @mlec1 ,

We are reviewing this pull request, to iterate faster, can we push review commits on your branch?
If you prefer we can open a pull request on it.

mlec1:
@nmengin Yeah no problem, if you need more permissions, I can add you to my project directly

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #11102, opened 2024-09-19
Guess Datadog socket type when prefix is unix

kevinpollet:
### What does this PR do?

This pull request mimics the Datadog client to guess the socket type when using the `unix` prefix. To avoid the guessing the prefixes `unixgram` for `SOCK_DGRAM` (datagram sockets) and `unixstream` for `SOCK_STREAM` (stream sockets), can be used.

### Motivation

To fix the issue raised in https://github.com/traefik/traefik/pull/10199#issuecomment-2348492239.

### More

- [ ] Added/updated tests


rtribotte (approved):
LGTM

PR #11113, opened 2024-09-20
Ensure shellcheck failure exit code is reflected in GH job result

jnoordsij:
### What does this PR do?

In the previous implementation, the script would exit with code 0 as long as the last `shellcheck` command succeeds. This PR ensures that if any of the individual checks fail, so does the script as a whole.

### Motivation

When trying to debug the failed job at https://github.com/traefik/traefik/actions/runs/10962448670/job/30441708688?pr=11111,
I compared to https://github.com/traefik/traefik/actions/runs/10955418669/job/30419255534 to check for differences. Although still not finding why the job on my PR is failing, I noticed shellcheck reporting errors in the latter job, which were not causing the job to fail.

### More

- [x] Added/updated tests

### DIFF
--- a/script/code-gen.sh
+++ b/script/code-gen.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
-# shellcheck disable=SC2046
 
 set -e -o pipefail
 
+# shellcheck disable=SC1091 # Cannot check source of this file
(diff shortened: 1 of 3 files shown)


kevinpollet (approved):
Thanks 👍

mmatur (approved):

Issue #3986, opened 2018-10-08
Not looking for files in subdirectories dynamically


BullHacks3:
### Issue
Traefik file  handlers not looking for the files in the subdirectories
https://docs.traefik.io/configuration/backends/file/#multiple-separated-files
In documentation ,it is clearly mentioned ,that file handlers have capability to look into the subdirectories for the files dynamically

### How to reproduce the issue ###

Step 1:- Contents of the main.toml file

```sh
defaultEntryPoints = ["http"]
[entryPoints]
  [entryPoints.http]
  address = ":80"
  [api]
 entrypoint="dashboard"
  dashboard= true
  debug=true

[file]
  directory= "/home/traefik/hostRules/"
  watch=true
```
Step 2:-  Now I am dynamically creating the subdirectories inside the mounted directory 

![try](https://user-images.githubusercontent.com/34959439/46595145-a55d8280-caf4-11e8-9415-3d7ff4488a78.png)

For example the dummy configuration of one of the traefik file is:-
```sh
[file]
[frontends]
    [frontends.front_end_abc]
         backend="backend3"
        passHostHeader = true
    [frontends.front_end_abc.routes.first]
         rule="Host: testingpurpose.com"
   [backends]
    [backends.backend3]
         [backends.backend3.servers.first]
           url = "http://192.168.1.2:39832"

```
But the configurations are not loaded dynamically until ,I restart the docker image
Command used to up the docker container
```sh
docker run -d -v /home/traefik:/home/traefik -p 80:80  -p 443:443 -p 172.31.30.68:8080:8080 testing-traefik /usr/bin/traefik -c /home/traefik/main.toml
```

### What did you expect to see?
I expected that the configurations inside the subdirectories should be loaded dynamically without restarting the docker image

### What did you see instead?
I have to restart the docker image ,in order to load the configurations inside the subdirectories that are creating dynamically.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Used the latest traefik image ,for building own docker image on the top of centos
```

nmengin:
Hello @BullHacks3 ,

Many thanks for interest in the project.

I guess the problem comes from the watcher we use which is not recursive.
Indeed, Træfik can load file from subbdirectories but can not listen to events from these subdirectories.

As a workaround, by waiting for the fix, you can touch a file in the root directory when a modification is done in a subdirectory to load all the configuration (or make a modification in an existing file).
It is not perfect bu more dynamic than restarting Træfik.

[...]

PR #11117, opened 2024-09-22
Allow subfolder file watchers

centja1:
### What does this PR do?

Fix bug #3986 to allow watching configuration files in subfolders of the File Provider directory 

Removed lines 69-72 in pkg/provider/file/file.go, which intentionally prevented creation of file watchers for subfolders

### Motivation

I had a need for a hierarchical configuration file structure

### More

- [ ] Added/updated tests

### Additional Notes

Tested for several weeks in a locally built docker image


nmengin:
Hey @centja1,

After discussing the need and the solution you proposed with the other maintainers, we prefer to close it.
As explained by @juliens on the [related issue #3986,](https://github.com/traefik/traefik/issues/3986#issuecomment-2595849952) we prefer not to bring such a behavior in Traefik.

MK796:
There is new upstream context relevant to why this PR and #3986 were declined.

Traefik currently loads configuration files recursively from the configured file-provider directory. However, with `watch: true`, it only registers watches for the configured root and supported top-level configuration files. Startup discovery and runtime watching therefore behave differently for nested configuration trees.

For Traefik, the preferred behavior would be automatic recursive watching below the existing configured `providers.file.directory` whenever `watch: true` is enabled.

The existing configured `directory` should remain the root of that recursive watch. Traefik should automatically watch the complete tree below it, including directories created after startup.

This preserves the existing configuration model and makes startup discovery and runtime watching consistent. It does not require an additional recursion-specific user-facing option. It is still a behavioral change, but it matches what users reasonably expect from watching a directory.

fsnotify has active work and related implementation history targeting the missing upstream capability:

- fsnotify/fsnotify#676 provides a common recursive wrapper.
- fsnotify/fsnotify#733 originally attempted recursion across all backends.
- fsnotify/fsnotify#750 covered the remaining inotify fixes and was closed because fsnotify/fsnotify@853adeb already addressed them.
- fsnotify/fsnotify#751 adds recursive FEN support.
- fsnotify/fsnotify#752 adds recursive kqueue support.
- fsnotify/fsnotify#753 publicly enables recursive watches through the `/...` path suffix after the backend work is complete.

The common-wrapper approach in fsnotify/fsnotify#676 and the per-backend path represented by fsnotify/fsnotify#751 and fsnotify/fsnotify#752, followed by public enablement in fsnotify/fsnotify#753, appear to be alternative implementation approaches rather than a linear dependency chain.

This is not a request to merge this PR as-is. A complete implementation must define and test:

- directories present at startup
- directories created after startup
- rename, removal, and recreation
- atomic file replacement
- symlink behavior
- bind-mounted configuration
- all operating systems supported by Traefik

Would the Traefik maintainers reconsider automatic recursive file-provider watching once fsnotify exposes a stable cross-platform API with defined semantics?

nmengin:
Hey @MK796,

Thank you for your feedback. Could you please report it in the related issue, please?
It would be simpler to keep it visible.

MK796:
Hey @nmengin,

the related issue #3986 is currently locked and limits comments to collaborators, so I cannot report the context there.

Could you please unlock it temporarily? I will move the complete comment there immediately.

https://github.com/traefik/traefik/issues/3986

nmengin:
Hey @MK796,

I haven't seen the issue was closed.
Could you open a dedicated one instead or add a comment into https://github.com/traefik/traefik/issues/11512?

MK796:
Thanks. I opened a dedicated issue:

https://github.com/traefik/traefik/issues/13557

#11512 concerns multiple independent file-provider root directories. The new issue concerns automatic recursive watching below one existing `providers.file.directory`, including directories created after startup. Keeping them separate avoids mixing two different configuration semantics.

PR #11135, opened 2024-09-27
Drop response body when not allowed for status code

kevinpollet:
### What does this PR do?

This pull request drops the response body when it is not allowed for the status code in the fast proxy implementation.

### Motivation

To be consistent with the `net/httputil` implementation.

### More

- [ ] Added/updated tests


rtribotte (approved):
LGTM

PR #11139, opened 2024-09-30
Clarify only header fields may be redacted in access-logs.md

mattbnz:
The current documentation for field handling implies that redact is a supported option for both normal fields and header fields, however this is not true.

As per https://github.com/traefik/traefik/blob/master/pkg/middlewares/accesslog/logger.go#L350-L358 request fields are processed by the Keep method (https://github.com/traefik/traefik/blob/master/pkg/types/logs.go#L89) which does not perform any redaction logic. 

That logic is only available to header fields on which redactHeaders is explicit called - https://github.com/traefik/traefik/blob/master/pkg/middlewares/accesslog/logger.go#L366

Given the redaction method specifically mentions headers in its name, this seems like a documentation bug, rather than an actual logic omission.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/observability/access-logs.md
+++ b/docs/content/observability/access-logs.md
@@ -156,7 +156,8 @@ Each field can be set to:
 
 - `keep` to keep the value
 - `drop` to drop the value
-- `redact` to replace the value with "redacted"
+
+Header fields may also optionally be set to `redact` to replace the value with "redacted".
 
 The `defaultMode` for `fields.names` is `keep`.
 


rtribotte:
Hello @mattbnz,

Thanks for this contribution!

This documentation enhancement looks to be valuable for v2.11 as well, can you please rebase the changes to target the v2.11 branch?

mattbnz:
PTAL - rebased onto v2.11

docs/content/observability/access-logs.md
```diff
@@ -138,7 +138,8 @@ Each field can be set to:
 
 - `keep` to keep the value
 - `drop` to drop the value
-- `redact` to replace the value with "redacted"
+
+Header fields may also optionally be set to `redact` to replace the value with "redacted".
```
kevinpollet, line 142:
```suggestion
Header fields may also optionally be set to `redact` to replace the value with "REDACTED".
```

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

Issue #10735, opened 2024-05-17
filter query params from logs


ProfXponent:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

traefik does not provide any means of filtering sensitive query params from logs aside form redacting the entire require uri, which is not an option for many

request uris like the one below often contain sensitive data that should not be persisted to a log file

```shell
192.168.112.1 - - [17/May/2024:08:48:35 +0000] "GET /cms-api/auth/google/callback?access_token=REDACTED HTTP/1.1" 200 339 "-" "-" 1390513 "web@file" "http://web:3000" 52ms
```

this is a major issue and is likely compromising many traefik users already

hood:
+1, I’d love this

ProfXponent:
just pinging to see if this has been prioritized and put onto the roadmap yet

nmengin:
Hello @ProfXponent,

As @sdelicata mentioned in his latest message, we'd like to wait 6 months to ensure this feature gets enough traction to be added to the roadmap.

ProfXponent:
> Hello @ProfXponent,
> 
> As @sdelicata mentioned in his latest message, we'd like to wait 6 months to ensure this feature gets enough traction to be added to the roadmap.

does that process apply to security issues, because that is how I would classify this.

I feel like placing this is in the feature pipeline ignores a pretty big vulnerability that you are exposing your users to.

nmengin:
Hey @ProfXponent,

Discussing this issue with the other maintainers, we agreed that Traefik should allow filtering the query parameters (and not only filtering the path).
We move the issue as `enhancement` accordingly.

Waiting for it, note that you can filter the `RequestPath` form the access logs as described in [the documentation](https://doc.traefik.io/traefik/observability/access-logs/#limiting-the-fieldsincluding-headers).

danielfeinblatt:
Following up on this issue to see if it can be reprioritized with the extra context of Ingress Nginx retirement. We need to log the request path for investigating the requests Traefik receives, but we cannot log query parameters as they can contain sensitive data. This is a security shortcoming that is a dealbreaker for us migrating from Ingress Nginx to Traefik. We'd be interested in implementing this to unblock ourselves and others looking to migrate to Traefik.

PR #11140, opened 2024-09-30
Support filtering query params in the accesslog

mattbnz:
### What does this PR do?

Separates the request query parameters out of the RequestPath field into a new accesslog field named RequestQuery to allow configuration of whether they are kept or dropped separately from the path.

### Motivation

This Fixes: #10735, I needed this functionality now and it looks pretty simply to add, so I did it and I'm speculatively sending this PR without prior consultation - if you don't like this approach at all I'm happy to retract it, alternatively if the approach is close to OK but needs some minor tweaks, then I'm open to that too.

### More

- [X] Added/updated tests

### Additional Notes

The main design choice here seems to be around whether to expose the query as a new field (RequestQuery) or merge it back into RequestPath - I've chosen the latter for now to avoid changing the logging "API" as such which seems most likely to be the desirable choice, but open to feedback if there are other considerations since it does make the behaviour/configuration between RequestPath/RequestQuery a little bit unusual...


rtribotte:
Hello @mattbnz,

Thanks for this contribution!

At first glance, we are not sure that this would really fix #10735.
If we are not mistaken, the main value brought by #10735 was being able to redact some params, where here all query params could be dropped.
WDYT?
Would you be inclined to rework the PR to allow redacting the query params?

mattbnz:
Perhaps... depends on what the complexity of the implementation ends up looking like.

I only really need the ability to drop fields for my use-case, but if we can agree on a simple-enough design that it's not too much effort to rework I'm open to it. 

Are you thinking something like add an additional Params attribute to AccessLogFields (https://github.com/traefik/traefik/blob/master/pkg/types/logs.go#L77) which parallels how headers are handled and allows the user to specify a default mode and then a drop/keep/redact directive for arbitrary query params?

rtribotte:
> Are you thinking something like add an additional Params attribute to AccessLogFields (https://github.com/traefik/traefik/blob/master/pkg/types/logs.go#L77) which parallels how headers are handled and allows the user to specify a default mode and then a drop/keep/redact directive for arbitrary query params?

Yes, indeed, that seems to be the way to go.

rtribotte:
Hi @mattbnz,

Are you still iterating on this PR to do the suggested modifications?

mattbnz:
I am still hoping to be able to make the modifications to get this upstreamed, but I have not yet found the time to actually make it happen in the last few months - it does pop up on my radar every 2 weeks for consideration in my sprint planning!

Apologies for the recent extraneous commits pushed to the branch, which are related to our internal build carrying the original changes, but are not relevant for the PR - I will get that tided up.

If I'm blocking someone else or causing issues please let me know.

rtribotte:
Hi @mattbnz,

Thanks for the feedback!

> Apologies for the recent extraneous commits pushed to the branch, which are related to our internal build carrying the original changes, but are not relevant for the PR - I will get that tided up.

Understood, we have updated the master branch https://github.com/traefik/traefik/pull/11653, when updating rebasing this commit won't be necessary.

> If I'm blocking someone else or causing issues please let me know.

Sure, but not at all, we just wanted to make sure that this PR could be left open!

kevinpollet:
Hello @mattbnz,

Please feel free to reopen if you need to.

PR #11143, opened 2024-09-30
Only calculate basic auth hashes once for concurrent requests

michelheusschen:
### What does this PR do?
Reuse the result of basic auth credential checks for in-flight requests

### Motivation
Browsers load resources from a basic auth-protected website in parallel, sending multiple simultaneous requests with the same credentials. Traefik has to recalculate the hash for each request which slows down the page load times, especially when using bcrypt.

### More

- [ ] Added/updated tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -86,6 +86,7 @@ require (
 	golang.org/x/exp v0.0.0-20240909161429-701f63a606c0 // No tag on the repo.
 	golang.org/x/mod v0.21.0
 	golang.org/x/net v0.29.0
+	golang.org/x/sync v0.8.0
 	golang.org/x/sys v0.25.0
 	golang.org/x/text v0.18.0
 	golang.org/x/time v0.5.0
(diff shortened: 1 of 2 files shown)


michelheusschen:
@emilevauge I am aware of that issue, however this isn't caching. The plaintext password is only stored in memory temporarily, while there are pending requests. Once those are handled, the password is garbage collected. This behavior is essentially the same as the current implementation.

emilevauge:
Thanks for you additional explanations.
We will review this one, however, I can already tell you that this part should be unit tested (even though it might be tricky).

kevinpollet:
Hello @michelheusschen and thanks for adding the unit tests,

Could you also tweak the tests to make sure that the singleflightGroup is called only once when two concurrent requests are handled by the middleware for the same key?

If you prefer we can push review commits in that direction?

michelheusschen:
@kevinpollet I can take a look at that later, but feel free to commit some changes in the meantime

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

Just so you know, we removed other concurrent tests as the behavior was already tested. Thanks again for adding the hash once test.

PR #11159, opened 2024-10-03
Update go-acme/lego to v4.19.0

ldez:
### What does this PR do?

Update go-acme/lego to v4.19.0

https://github.com/go-acme/lego/compare/v4.18.0...v4.19.0

### Motivation

- **[dnsprovider]** ionos: follow CNAME
- **[lib,dnsprovider]** Reducing the lock strength of the SOA cache entry
- **[dnsprovider]** Use UTC instead of GMT when possible (remove the need to tzdata)
- **[dnsprovider]** namesilo: restrict CleanUp
- **[dnsprovider]** godaddy: fix cleanup

### More

- ~~[ ] Added/updated tests~~

### Additional Notes

- **[dnsprovider]** Add DNS provider for HuaweiCloud
- **[dnsprovider]** Add DNS provider for SelfHost.(de|eu)
- **[lib,dnsprovider]** Add `PropagationWait` function


ldez:
The CI failure is fixed inside #11160

kevinpollet:
Hello @ldez,

Why introduce the `propagationRN` and `propagationDisableANS` options in a fix release? Are they needed to fix current issues?

If this is not the case I would personally split the pull request and introduce those flags in the next version.

ldez:
This fixes situations where authoritative nameservers don't work as expected by lego: only the recursive nameservers will be used (the recursive nameservers are either those defined inside `/etc/resolv.conf` or through the `certificatesresolvers.myresolver.acme.dnschallenge.resolvers`).

There are mainly 2 situations: split DNS horizon and private PKI with custom DNS structure.

Before the only option was to disable the propagation checks and add a delay, but this is flaky and slow because of the usage of a delay.

pkg/provider/acme/provider.go
```diff
@@ -80,6 +80,8 @@ type DNSChallenge struct {
 	DelayBeforeCheck        ptypes.Duration `description:"Assume DNS propagates after a delay in seconds rather than finding and querying nameservers." json:"delayBeforeCheck,omitempty" toml:"delayBeforeCheck,omitempty" yaml:"delayBeforeCheck,omitempty" export:"true"`
 	Resolvers               []string        `description:"Use following DNS servers to resolve the FQDN authority." json:"resolvers,omitempty" toml:"resolvers,omitempty" yaml:"resolvers,omitempty"`
 	DisablePropagationCheck bool            `description:"Disable the DNS propagation checks before notifying ACME that the DNS challenge is ready. [not recommended]" json:"disablePropagationCheck,omitempty" toml:"disablePropagationCheck,omitempty" yaml:"disablePropagationCheck,omitempty" export:"true"`
+	PropagationRNS          bool            `description:"Use all the recursive nameservers to check the propagation of the TXT record." json:"propagationRNS,omitempty" toml:"propagationRNS,omitempty" yaml:"propagationRNS,omitempty" export:"true"`
```
rtribotte, line 83:
```suggestion
	PropagationRNS          bool            `description:"Use all the recursive name servers to check the propagation of the TXT record." json:"propagationRNS,omitempty" toml:"propagationRNS,omitempty" yaml:"propagationRNS,omitempty" export:"true"`
```

ldez: For the DNS word, "nameservers" is used as one word: https://developers.cloudflare.com/dns/nameservers/

pkg/provider/acme/provider.go
```diff
@@ -80,6 +80,8 @@ type DNSChallenge struct {
...
+	PropagationDisableANS   bool            `description:"Disable the need to await propagation of the TXT record to all authoritative name servers." json:"propagationDisableANS,omitempty" toml:"propagationDisableANS,omitempty" yaml:"propagationDisableANS,omitempty" export:"true"`
```
rtribotte, line 84:
```suggestion
	PropagationDisableANS   bool            `description:"Disable the need to await propagation of the TXT record to all authoritative nameservers." json:"propagationDisableANS,omitempty" toml:"propagationDisableANS,omitempty" yaml:"propagationDisableANS,omitempty" export:"true"`
```

rtribotte:
Hi @ldez,

After giving more thought to the changes, we prefer to go with the go-acme/lego upgrade only.
After reading the https://github.com/go-acme/lego/issues/2276 usage scenario for which it was introduced, we are not sure that Traefik should support it, as it seems a bit edge.
Also, since it is an extension within go-acme/lego itself, we would prefer to introduce it as an extension.
As a side note, the current naming of the options feels a bit technical, and we think it would be rather complicated to understand the meaning just by reading it. But we can think about it more when the time comes.

Could you please update the PR accordingly? (I can see that you did not check the box to allow maintainer changes, and as you know the bot will not be able to rebase the PR if needed, it would be great if you could check that box)

rtribotte (approved):
LGTM

kevinpollet (approved):

PR #11162, opened 2024-10-04
Bump github.com/klauspost/compress to dbd6c381492a

kevinpollet:
### What does this PR do?

This pull request is a follow-up to https://github.com/traefik/traefik/pull/11141 to fix another issue with the compress dependency which was adding a `content-type` header even if the response has no body.  

### Motivation

To fix a regression.

### More

- [ ] Added/updated tests


nmengin (approved):
LGTM

youkoulayley (approved):
LGTM

PR #11175, opened 2024-10-08
fix: Support `http` and `https` for Kubernetes Service `appProtocol`

WillDaSilva:
### What does this PR do?

Currently if a Kubernetes Service species the `appProtocol` field as `http` or `https`, despite Traefik having support for those protocols, Traefik will raise the error

```
Cannot load HTTPBackendRef <namespace>/<service>: unsupported application protocol http
```

### Motivation

I was trying to expose the RabbitMQ management UI, using a RabbitMQ cluster controlled by https://github.com/rabbitmq/cluster-operator. The service this operator creates sets the `appProtocol` field to `http`, which causes Traefik to error. As it stands, Traefik only works properly with `http`/`https` when the field is not set.


WillDaSilva:
Closed in favour of https://github.com/traefik/traefik/pull/11175, which targets v3.2

PR #11176, opened 2024-10-08
Support http and https appProtocol for Kubernetes Service

WillDaSilva:
### What does this PR do?

Currently if a Kubernetes Service species the `appProtocol` field as `http` or `https`, despite Traefik having support for those protocols, Traefik will raise the error

```
Cannot load HTTPBackendRef <namespace>/<service>: unsupported application protocol http
```

Documentation for the `appProtocol` field: https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol. Notably, it may be any [IANA standard service names](https://www.iana.org/assignments/service-names), such as `http` or `https`.

### Motivation

I was trying to expose the RabbitMQ management UI, using a RabbitMQ cluster controlled by https://github.com/rabbitmq/cluster-operator. The service this operator creates sets the `appProtocol` field to `http`, which causes Traefik to error. As it stands, Traefik only works properly with `http`/`https` when the field is not set.

### DIFF
--- a/pkg/provider/kubernetes/gateway/grpcroute.go
+++ b/pkg/provider/kubernetes/gateway/grpcroute.go
@@ -417,14 +417,14 @@ func getGRPCServiceProtocol(portSpec corev1.ServicePort) (string, error) {
 	}
 
 	if portSpec.AppProtocol == nil {
-		return "h2c", nil
+		return schemeH2C, nil
 	}
 
 	switch ap := *portSpec.AppProtocol; ap {
 	case appProtocolH2C:
-		return "h2c", nil
+		return schemeH2C, nil
 	case appProtocolHTTPS:
-		return "https", nil
+		return schemeHTTPS, nil
 	default:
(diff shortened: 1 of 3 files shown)


kevinpollet:
Hello @WillDaSilva and thanks for opening this pull request,

We also think that the `https` appProtocol must be supported for [GRPCRoute](https://github.com/traefik/traefik/blob/v3.2/pkg/provider/kubernetes/gateway/grpcroute.go#L337) (to be able to do gRPC over https). The server protocol should be adapted accordingly [here](https://github.com/traefik/traefik/blob/v3.2/pkg/provider/kubernetes/gateway/grpcroute.go#L353).

Would you mind doing the changes, or is it ok for you if we push review commits as we want to cut a new RC soon?

WillDaSilva:
@kevinpollet please feel free to push changes to my branch. You'd know much better than I what needs to be changed.

kevinpollet (approved):
Thanks 👍

PR #11177, opened 2024-10-09
add optional kind field to ingressroutetcp to match ingressroute

skirtan1:
### What does this PR do?

Add an optional field kind  field to IngressRouteTCP  CRD making it similar IngressRoute CRD
 

### Motivation

Issue: https://github.com/traefik/traefik/issues/11106

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/crd/traefikio/v1alpha1/ingressroutetcp.go
+++ b/pkg/provider/kubernetes/crd/traefikio/v1alpha1/ingressroutetcp.go
@@ -26,6 +26,10 @@ type RouteTCP struct {
 	// Match defines the router's rule.
 	// More info: https://doc.traefik.io/traefik/v3.2/routing/routers/#rule_1
 	Match string `json:"match"`
+	// Kind defines the kind of the route.
+	// Rule is the only supported kind.
+	// +kubebuilder:validation:Enum=Rule
+	Kind string `json:"kind"`
 	// Priority defines the router's priority.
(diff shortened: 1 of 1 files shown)


StefanTheWiz:
might need to add the same to `ServiceTCP` too

nmengin:
Hello @skirtan1,

In the [related issue](https://github.com/traefik/traefik/issues/11106#issuecomment-2363947376), we proposed only to set `Rule` as the default `kind` in `IngressRoute`to avoid specifying it every time.

>Discussing with the other maintainers, we think that could be relevant to make the option kind optional, setting the value Rule per default.

WDYT to implement this proposal?

skirtan1:
> In the [related issue](https://github.com/traefik/traefik/issues/11106#issuecomment-2363947376), we proposed only to set `Rule` as the default `kind` in `IngressRoute`to avoid specifying it every time.
> 

Yeah, this makes more sense and changes are minimal as well. 

I have made route.Kind optional and let it pass the check if it is an empty value. 

The issue also mentions about fields in ServiceTCP, I don't know if this is also to be addressed for now.

skirtan1:
@nmengin I have repurposed a unit test case which failed if the rule kind was empty and updated the crd docs at appropriate places. If there are changes which should be made please let me know.

rtribotte (approved):
Thanks

kevinpollet (approved):
Thanks 👍

PR #11193, opened 2024-10-11
fix: keep server path option

### LINKED ISSUE #10487: Keep URIs in server list
### Welcome!

I have two backends that I'd like to load balance between, but both servers are authenticated using a different URI:

```yaml
services:
  my-backend:
    loadBalancer:
      servers:
        - url: https://my-first-backend.com/my-auth-token
        - url: https://my-second-backend.com/my-other-different-auth-token
```

Since Traefik strips the URI, this configuration is impossible. Also, since the URIs are different between these servers, the add-prefix middleware will not work, nor will health checks...

The end goal is to have a request come into traefik (`https://traefik:8080`) and then get routed to either backend, as long as it's healthy. In the future, more backends could be added and I'd like Traefik to load balance between them.

Is there any solution to this with Traefik or should I find a different proxy?

mmatur:
### What does this PR do?

This PR adds an option to allow to keep the path for the upstream server

Fixes: #10487

### Motivation

Fixes #10487

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/routing/services/index.md
+++ b/docs/content/routing/services/index.md
@@ -118,11 +118,6 @@ Each service has a load-balancer, even if there is only one server to forward tr
 Servers declare a single instance of your program.
 The `url` option point to a specific instance.
 
-!!! info ""
-    Paths in the servers' `url` have no effect.
-    If you want the requests to be sent to a specific path on your servers,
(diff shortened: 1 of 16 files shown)


docs/content/routing/services/index.md
```diff
@@ -173,6 +168,33 @@ The `weight` option allows for weighted load balancing on the servers.
           weight = 1
     ```
 
+The `preservePath` option allow to preserve url path.
```
kevinpollet, line 171:
```suggestion
The `preservePath` option allows to preserve the URL path.
```

kevinpollet (approved):

juliens (approved):
LGTM

PR #11196, opened 2024-10-14
Update `traefik.docker.network` to be `traefik.swarm.network`

webash:
### What does this PR do?

Updates a reference of old `traefik.docker.network` configuration label to new `traefik.swarm.network` configuration label.

### Motivation

To prevent confusion to documentation users.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/providers/swarm.md
+++ b/docs/content/providers/swarm.md
@@ -360,7 +360,7 @@ _Optional, Default=""_
 
 Defines a default docker network to use for connections to all containers.
 
-This option can be overridden on a per-container basis with the `traefik.docker.network` label.
+This option can be overridden on a per-container basis with the `traefik.swarm.network` label.
 
 ```yaml tab="File (YAML)"
 providers:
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hello @webash and thanks for opening this pull request,

I checked the swarm provider code and we kept the label named `traefik.docker.network` to avoid a breaking change. According to this, I am closing this pull request, feel free to reopen it you think that I am missing something.

ldez:
> but the swarm provider in v3 is already breaking!

It's breaking from v2, but the label change will be breaking inside v3, it's not the same thing.

> So should I put a PR into the documentation so that its update to state to continue to use `traefik.docker.network` even when the provider in use is `swarm`?

I think you are confusing labels ([dynamic configuration](https://doc.traefik.io/traefik/getting-started/configuration-overview/#the-dynamic-configuration)) and [static configuration](https://docs.traefik.io/traefik/getting-started/configuration-overview/#the-static-configuration).

The label `traefik.docker.network` overrides the static configuration `traefik.swarm.network`.

webash:
I'm confusing it because those two things are different strings, and the documentation doesn't clarify that despite the provider for Swarm being different, the network label still uses the `docker` text. I don't think its irrational for me to confused by something that isn't clearly documented.

I'll put in a new PR that adds a note to explain that for future travellers.

Issue #11200, opened 2024-10-16
The behavior of Kubernetes Ingress pathType Prefix does not align with Kubernetes docs


charlie0129:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

Consider the following Ingress. We have two paths using Prefix:

- Backend: `/v1`
- Frontend: `/`

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: {{ include "appName" . }}
spec:
  rules:
    - host: {{ include "appName" . }}.{{ .Values.domain }}
      http:
        paths:
          - path: /v1 # Backend
            pathType: Prefix
            backend:
              service:
                name:  {{ include "appName" . }}
                port:
                  number: 80
          - path: / # Frontend
            pathType: Prefix
            backend: 
              service:
                name: {{ include "frontAppName" . }}
                port:
                  number: 80       
```

According to [Kubernetes Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/#examples) documentation:

- `/v11` should match Frontend
- `/v1/something` should match Backend

### What did you see instead?

In our testing:

- `/v11` matched Backend

This is not consistent with [Kubernetes Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/#examples) documentation.

Kind | Path | Request Path | Matches?
-- | -- | -- | --
Prefix | /aaa/bbb | /aaa/bbbxyz | No, does not match string prefix

### What version of Traefik are you using?

v3.0.0

### What is your environment & configuration?

Traefik installed via Helm traefik-28.0.0
Kubernetes v1.28.4

nmengin:
Hello @charlie0129,

Thanks for the bug report, we will work on it to reproduce it.

If you or another community member would like to build it.
We'll update this issue with our first results.

charlie0129:
Sure I can open a PR to fix this.

charlie0129:
@nmengin Hi, sorry for mentioning, but is there a possibility that my pull request can get reviewed?

PR #11203, opened 2024-10-17
fix: make the behavior of prefix matching in Ingress consistent with Kubernetes doc

charlie0129:
### What does this PR do?

Prefix matching in Kubernetes Ingress `/aaa` no longer matches `/aaabbb` to keep the Ingress behavior consistent with [Kubernetes Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/#examples) documentation.

### Motivation

Please see #11200 for background.

TLDR; According to [Kubernetes Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/#examples) documentation: a prefix matching such as `/aaa` should not match `/aaabbb`.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/routing/providers/kubernetes-ingress.md
+++ b/docs/content/routing/providers/kubernetes-ingress.md
@@ -388,8 +388,8 @@ which in turn will create the resulting routers, services, handlers, etc.
 If the Kubernetes cluster version is 1.18+,
 the new `pathType` property can be leveraged to define the rules matchers:
 
-- `Exact`: This path type forces the rule matcher to `Path`
-- `Prefix`: This path type forces the rule matcher to `PathPrefix`
+- `Exact`: This path type forces the rule matcher to `Path` such that a request path is matched exactly and with case sensitivity
+- `Prefix`: This path type forces the rule matcher to a special type of `PathRegexp` such that a request path is matched on an element by element basis and with case sensitivity
 
 Please see [this documentation](https://kubernetes.io/docs/concepts/services-networking/ingress/#path-types) for more information.
 
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -37,6 +37,7 @@ const (
 	traefikDefaultIngressClass           = "traefik"
 	traefikDefaultIngressClassController = "traefik.io/ingress-controller"
 	defaultPathMatcher                   = "PathPrefix"
+	prefixMatchRegexTemplate             = "^%s(|/.*)$"
 )
 
 // Provider holds configurations of the provider.
@@ -751,7 +752,7 @@ func loadRouter(rule netv1.IngressRule, pa netv1.HTTPIngressPath, rtConfig *Rout
 			matcher = "Path"
 		}
 
-		rules = append(rules, fmt.Sprintf("%s(`%s`)", matcher, pa.Path))
+		rules = append(rules, buildRule(matcher, pa.Path))
 	}
 
 	rt := &dynamic.Router{
@@ -771,6 +772,24 @@ func loadRouter(rule netv1.IngressRule, pa netv1.HTTPIngressPath, rtConfig *Rout
 	}
 
 	return rt
+}
+
+func buildRule(matcher string, path string) string {
+	if matcher == "PathPrefix" {
+		// We want to keep the behavior same as https://kubernetes.io/docs/concepts/services-networking/ingress/#examples.
+		// i.e. /v12 should not match /v1 , but traefik's PathPrefix matcher works as a prefix match not the
+		// element by element match as Kubernetes Ingress, so we need to use PathRegexp as a workaround.
+		// Check out TestPrefixMatchRegex() for more examples.
+		path = strings.TrimSuffix(path, "/")
+		return "PathRegexp(`" + buildPrefixMatchRegex(path) + "`)"
+	}
+
+	return fmt.Sprintf("%s(`%s`)", matcher, path)
+}
+
+func buildPrefixMatchRegex(path string) string {
+	path = strings.TrimSuffix(path, "/")
+	return fmt.Sprintf(prefixMatchRegexTemplate, regexp.QuoteMeta(path))
 }
 
 func throttleEvents(ctx context.Context, throttleDuration time.Duration, pool *safe.Pool, eventsChan <-chan interface{}) chan interface{} {
--- a/pkg/provider/kubernetes/ingress/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes_test.go
@@ -3,8 +3,10 @@ package ingress
 import (
 	"context"
 	"errors"
+	"fmt"
 	"math"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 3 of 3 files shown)


charlie0129:
Hi @rtribotte ,

I added an option (`strictPrefixMatching`) in the Ingress Provider configuration, WDYT?

charlie0129:
Looks like the some tests failed. I will proceed to fix it.

docs/content/routing/providers/kubernetes-ingress.md
```diff
@@ -438,8 +438,8 @@ Fenced servers can still process requests tied to sticky cookies, while they are
 If the Kubernetes cluster version is 1.18+,
 the new `pathType` property can be leveraged to define the rules matchers:
 
-- `Exact`: This path type forces the rule matcher to `Path` such that a request path is matched exactly and with case sensitivity
-- `Prefix`: This path type forces the rule matcher to a special type of `PathRegexp` such that a request path is matched on an element by element basis and with case sensitivity
+- `Exact`: This path type forces the rule matcher to `Path`
+- `Prefix`: This path type forces the rule matcher to `PathPrefix`. Note that if you want the matching behavior to strictly follow Kubernetes Ingress spec (request path is matched on an element by element basis), consider enabling `strictPrefixMatching` in the Ingress Provider configuration.
```
sdelicata, line 442:
```suggestion
- `Prefix`: This path type forces the rule matcher to `PathPrefix`. Note that if you want the matching behavior to strictly comply with Kubernetes Ingress specification (request path is matched on an element-by-element basis), consider enabling `strictPrefixMatching` in the Ingress Provider configuration.
```

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -57,6 +57,7 @@ type Provider struct {
 	DisableIngressClassLookup    bool `description:"Disables the lookup of IngressClasses (Deprecated, please use DisableClusterScopeResources)." json:"disableIngressClassLookup,omitempty" toml:"disableIngressClassLookup,omitempty" yaml:"disableIngressClassLookup,omitempty" export:"true"`
 	DisableClusterScopeResources bool `description:"Disables the lookup of cluster scope resources (incompatible with IngressClasses and NodePortLB enabled services)." json:"disableClusterScopeResources,omitempty" toml:"disableClusterScopeResources,omitempty" yaml:"disableClusterScopeResources,omitempty" export:"true"`
 	NativeLBByDefault            bool `description:"Defines whether to use Native Kubernetes load-balancing mode by default." json:"nativeLBByDefault,omitempty" toml:"nativeLBByDefault,omitempty" yaml:"nativeLBByDefault,omitempty" export:"true"`
+	StrictPrefixMatching         bool `description:"Make prefix matching strictly follow the Kubernetes Ingress spec (path-element-wise matching instead of string-character-wise matching)." json:"strictPrefixMatching,omitempty" toml:"strictPrefixMatching,omitempty" yaml:"strictPrefixMatching,omitempty" export:"true"`
```
sdelicata, line 60:
```suggestion
	StrictPrefixMatching         bool `description:"Make prefix matching strictly comply with the Kubernetes Ingress specification (path-element-wise matching instead of character-by-character string matching)." json:"strictPrefixMatching,omitempty" toml:"strictPrefixMatching,omitempty" yaml:"strictPrefixMatching,omitempty" export:"true"`
```

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -844,6 +846,27 @@ func makeRouterKeyWithHash(key, rule string) (string, error) {
...
+	return fmt.Sprintf("%s(`%s`)", matcher, path)
+}
+
+// buildPrefixMatchRegex returns a regex that matches the given path and any subpaths.
+// For example, if path is "/v1", the regex will match "/v1" and "/v1/anything", but not "/v12".
+func buildPrefixMatchRegex(path string) string {
+	path = strings.TrimSuffix(path, "/")
+	return fmt.Sprintf(prefixMatchRegexTemplate, regexp.QuoteMeta(path))
+}
+
```
sdelicata, line 869:
```suggestion
func buildRule(strictPrefixMatching bool, matcher string, path string) string {
	// When enabled, strictPrefixMatching ensures that prefix matching follows
	// the Kubernetes Ingress spec (path-element-wise instead of character-wise).
	if strictPrefixMatching && matcher == "PathPrefix" {
		// To mimic Kubernetes behavior (see:
		// https://kubernetes.io/docs/concepts/services-networking/ingress/#examples),
		// "/v12" should not match "/v1".
		//
		// Traefik's default PathPrefix matcher performs a character-wise prefix match,
		// unlike Kubernetes which matches path elements.
		// Use PathRegexp to replicate element-wise behavior.
		//
		// See TestPrefixMatchRegex() for examples.
		path = strings.TrimRight(path, "/")
		return "PathRegexp(`" + buildPrefixMatchRegex(path) + "`)"
	}

	return fmt.Sprintf("%s(`%s`)", matcher, path)
}

// buildPrefixMatchRegex returns a regex that matches the given path and its subpaths.
// For example, "/v1" matches "/v1" and "/v1/anything", but not "/v12".
func buildPrefixMatchRegex(path string) string {
	path = strings.TrimSuffix(path, "/")
	return fmt.Sprintf(prefixMatchRegexTemplate, regexp.QuoteMeta(path))
}
```

sdelicata (changes requested):
Thanks for the PR! Just a few suggestions to improve the clarity.

charlie0129:
Thank you for your suggestion and they have been applied.

lbenguigui (changes requested):

lbenguigui (approved):
LGTM

docs/content/routing/providers/kubernetes-ingress.md
```diff
@@ -439,7 +439,7 @@ If the Kubernetes cluster version is 1.18+,
 the new `pathType` property can be leveraged to define the rules matchers:
 
 - `Exact`: This path type forces the rule matcher to `Path`
-- `Prefix`: This path type forces the rule matcher to `PathPrefix`
+- `Prefix`: This path type forces the rule matcher to `PathPrefix`. Note that if you want the matching behavior to strictly comply with Kubernetes Ingress specification (request path is matched on an element-by-element basis), consider enabling `strictPrefixMatching` in the Ingress Provider configuration.
```
kevinpollet, line 442: I think adding a link to the `strictPrefixMatching` might be a good addition.

```suggestion
- `Prefix`: This path type forces the rule matcher to `PathPrefix`. Note that if you want the matching behavior to strictly comply with Kubernetes Ingress specification (request path is matched on an element-by-element basis), consider enabling [`strictPrefixMatching`](../../providers/kubernetes-ingress.md#strictprefixmatching) in the Ingress Provider configuration.
```

docs/content/providers/kubernetes-ingress.md
```diff
@@ -529,6 +529,29 @@ providers:
 --providers.kubernetesingress.nativeLBByDefault=true
 ```
 
+### `strictPrefixMatching`
```
kevinpollet, line 532: Please could you add this documentation to our new reference documentation page here: https://github.com/traefik/traefik/blob/master/docs/content/reference/install-configuration/providers/kubernetes/kubernetes-ingress.md?plain=1#L39

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -844,6 +846,32 @@ func makeRouterKeyWithHash(key, rule string) (string, error) {
...
+		// To mimic Kubernetes behavior (see:
+		// https://kubernetes.io/docs/concepts/services-networking/ingress/#examples),
+		// "/v12" should not match "/v1".
+		//
+		// Traefik's default PathPrefix matcher performs a character-wise prefix match,
+		// unlike Kubernetes which matches path elements.
+		// Use PathRegexp to replicate element-wise behavior.
+		//
+		// See TestPrefixMatchRegex() for examples.
+		return "PathRegexp(`" + buildPrefixMatchRegex(path) + "`)"
```
kevinpollet, line 862: I think it would be great using a matcher combination to be consistent with the Kubernetes Gateway provider (which also complies with the Ingress matching specification), something like: https://github.com/traefik/traefik/blob/master/pkg/provider/kubernetes/gateway/httproute.go#L645

pkg/provider/kubernetes/ingress/kubernetes_test.go
```diff
@@ -2256,3 +2294,94 @@ func readResources(t *testing.T, paths []string) []runtime.Object {
...
+			requestPath: "/.../",
+			match:       true,
+		},
+	}
+
+	for i, tt := range tests {
+		t.Run(fmt.Sprintf("Prefix match case #%d", i), func(t *testing.T) {
+			re := regexp.MustCompile(buildPrefixMatchRegex(tt.path))
+			assert.Equalf(t, tt.match, re.MatchString(tt.requestPath), "Kubernetes prefix matching path=%s, requestPath=%s incorrect", tt.path, tt.requestPath)
+		})
```
kevinpollet, line 2385:
```suggestion
		t.Run(fmt.Sprintf("Prefix match case #%d", i), func(t *testing.T) {
		        t.Parallel()
		        
			re := regexp.MustCompile(buildPrefixMatchRegex(tt.path))
			assert.Equalf(t, tt.match, re.MatchString(tt.requestPath), "Kubernetes prefix matching path=%s, requestPath=%s incorrect", tt.path, tt.requestPath)
		})
```

kevinpollet (changes requested):

pkg/provider/kubernetes/ingress/kubernetes.go
charlie0129: I think I can extract a common function (in gateway pkg) to build Path Prefix rules to avoid duplicating code. WDYT?

charlie0129:
@kevinpollet Hi, I changed my code according to your suggestions.

One additional change:

Since the gateway package has the same code for building a PathPrefix rule, I figured out that it can be a good idea to reuse the same code in gateway package to avoid duplicating code. So I made a public function in gateway package and call it in in ingress package. See https://github.com/traefik/traefik/pull/11203/commits/373741c29b21370aa29fbe5e79a2affc63c92a12

charlie0129:
@kevinpollet Hi, I have fixed the code. Moved the code back to Ingress package.

kevinpollet (approved):
Thanks 👍

PR #11207, opened 2024-10-21
Documentation: adding tips about the use of `docker` in dynamic configuration for `swarm` provider

webash:
### What does this PR do?

Per [my comments](https://github.com/traefik/traefik/pull/11196#issuecomment-2427844467) in #11196 , I believe it is confusing that the Docker Swarm provider has its own `swarm` static configuration keywords, but still utilises the `docker` keyword for some label /dynamic configuration. As such, I've added tips in the places where this is.

### Motivation

Hopefully help others like me who moved from v2 to v3 and were otherwise under the assumption that `swarm` was to supplant all `docker`.

### More

- [ ] Added/updated tests

### Additional Notes
I attempted another PR via #11196 - but this one seems more in keeping with the intent of the v2/v3 changes.

### DIFF
--- a/docs/content/providers/swarm.md
+++ b/docs/content/providers/swarm.md
@@ -360,7 +360,13 @@ _Optional, Default=""_
 
 Defines a default docker network to use for connections to all containers.
 
-This option can be overridden on a per-container basis with the `traefik.docker.network` label.
+!!! tip "Per-container network override"
+
+    This option can be overridden on a per-container basis with the `traefik.docker.network` label.
+    
+    Note that the dynamic configuration label uses the `docker` term, not `swarm` like the static configuration provider.
+
+    Those migrating from v2 should therefore take caution not to replace all instances of `docker` in their configuration or labels.
 
 ```yaml tab="File (YAML)"
 providers:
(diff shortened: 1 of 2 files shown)


kevinpollet (changes requested):
Hello @webash and thanks for opening this pull request,

After discussing this with the team, we think we should address this issue and rename the `traefik.docker.*` labels to `traefik.swarm.*`. As this is a breaking change, in v3 we will support both labels and log a deprecation warning when the `traefik.docker.*` labels are used. In the next major version, the support of `traefik.docker.*` labels will be removed.

Could you open a new issue in that sense?

In the meantime, we are ok to merge the warnings you added, but we are not sure about the added tip. Could you remove it?

webash:
@kevinpollet apologies for the delay - day job has nothing to do with this stuff so I sometimes go a while without checking my personal GitHub account.

I've made a tweak to my PR that is hopefully more agreeable.

I've also logged #11244

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks

PR #11212, opened 2024-10-23
Detect and drop broken conns in the fastproxy pool

kevinpollet:
### What does this PR do?

This pull request fixes the fastproxy pool to detect a broken connection and avoid returning it when a connection is acquired. It also avoids to send the body back to the client when the response status code does not allow it.

### Motivation

To avoid bad gateway errors when a backend connection is broken.

### More

- [X] Added/updated tests

### Additional Notes


juliens (approved):
LGTM

lbenguigui (approved):
LGTM

PR #11216, opened 2024-10-23
fix grafana dashboard

spagno:
### Motivation
The current dashboard uses traefik_service_requests metric with the label "service". The label that should be used in the panels is exported_service


rtribotte:
Hello @spagno,

Thanks for this contribution!

I'm not sure about the changes, it seems to me that the service label key is indeed "service".
It looks like you need to use the honor_label option in the scrape config of Prometheus: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config, in order to keep the service label key as is.

We are declining this PR proposal as it seems related to a configuration issue.

PR #11229, opened 2024-10-28
Add Compress middleware to migration guide

logica0419:
### What does this PR do?

Added “encoding” field addition to Compress middleware on v3.2 migration guide.

### Motivation

RateLimit & InFlightReq middlewares are mentioned in the guide but Compress middleware isn’t.
“encoding” field is a huge progress in Compress middleware so I think it’s important enough to mention here.

### More

- [ ] Added/updated tests


rtribotte (approved):
Thanks

kevinpollet (approved):
Thanks 👍

Issue #10794, opened 2024-06-10
Add TCP Health Check using SYN, SYN-ACK, and RST packets


eos175:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

I have implemented a new TCP health check functionality using the `tcp-shaker` library (https://github.com/tevino/tcp-shaker). This improvement uses SYN, SYN-ACK, and RST packets to perform more accurate health checks for TCP services.

You can review the implementation on my branch [here](https://github.com/eos175/traefik/tree/dev).

Would love to get feedback and discuss how this can be integrated into the project.

Thanks!

eos175:
Some related issues are #5598 and #1657.

To use it, the procedure is as follows:

```yaml
services:
  example_tcp_service:
    loadBalancer:
      servers:
        - address: 10.42.64.4:7445
        - address: 10.42.0.1:7445
        - address: 127.0.0.1:8448
      healthCheck:
        path: /health
        interval: "10s"
        timeout: "3s"
        mode: tcp
```

Note that the `path` is ignored due to the nature of the TCP protocol.

Additionally, it would be interesting to consider a `mode: passive`, which enables passive checks based on observing communication with back-end nodes. However, this is not as straightforward.

leonidas-o:
@nmengin what do you mean by "we’re unsure about the use case"? I mean it's a tcp health check, which is really needed in such a product. Just look at #5598 there you will see, they are asking for it.
Another example and real use case. You want to install OpenShift with traefik as a load balancer:

https://docs.okd.io/4.16/installing/installing_bare_metal/installing-bare-metal-network-customizations.html#installation-load-balancing-user-infra-example_installing-bare-metal-network-customizations

They use HAProxy in the example, as you can see `mode: tcp and check /readyz`, so there is a need for a tcp health check at leas if you want to setup openshift:
```bash
listen api-server-6443 
  bind *:6443
  mode tcp
  option  httpchk GET /readyz HTTP/1.0
  ...
```
I can't say anything about @eos175 's implementation, but I (and I'm pretty sure a lot of others) would really appreciate if you would accept that. Of course if the implementation quality is up to your defined standards.

nmengin:
Hello @leonidas-o,

Thank you for your feedback.

[...]

PR #11238, opened 2024-10-29
Add TCP Healthcheck

ddtmachado:
### What does this PR do?

Adds TCP Healthcheck

### Motivation

To add healthchecks for TCP forwarded connections, making TCP services more relevant and on pair with HTTP services.
Most required for the proper working of TCP features such as Weighted Round Robin.

### More

- [X] Added/updated tests

### DIFF
--- a/pkg/config/dynamic/tcp_config.go
+++ b/pkg/config/dynamic/tcp_config.go
@@ -93,7 +93,8 @@ type TCPServersLoadBalancer struct {
 	// connection. It is a duration in milliseconds, defaulting to 100. A negative value
 	// means an infinite deadline (i.e. the reading capability is never closed).
 	// Deprecated: use ServersTransport to configure the TerminationDelay instead.
-	TerminationDelay *int `json:"terminationDelay,omitempty" toml:"terminationDelay,omitempty" yaml:"terminationDelay,omitempty" export:"true"`
+	TerminationDelay *int                  `json:"terminationDelay,omitempty" toml:"terminationDelay,omitempty" yaml:"terminationDelay,omitempty" export:"true"`
+	HealthCheck      *TCPServerHealthCheck `json:"healthCheck,omitempty" toml:"healthCheck,omitempty" yaml:"healthCheck,omitempty" export:"true"`
 }
 
 // Mergeable tells if the given service is mergeable.
@@ -168,4 +169,17 @@ func (t *TCPServersTransport) SetDefaults() {
 	t.DialTimeout = ptypes.Duration(30 * time.Second)
 	t.DialKeepAlive = ptypes.Duration(15 * time.Second)
 	t.TerminationDelay = ptypes.Duration(100 * time.Millisecond)
+}
+
+// +k8s:deepcopy-gen=true
+
+// TCPServer holds a TCP Server configuration.
+type TCPServerHealthCheck struct {
+	Address  string          `json:"address,omitempty" toml:"address,omitempty" yaml:"address,omitempty" label:"-"`
+	Port     string          `json:"-" toml:"-" yaml:"-"`
+	TLS      bool            `json:"tls,omitempty" toml:"tls,omitempty" yaml:"tls,omitempty"`
+	Interval ptypes.Duration `json:"interval,omitempty" toml:"interval,omitempty" yaml:"interval,omitempty" export:"true"`
+	Timeout  ptypes.Duration `json:"timeout,omitempty" toml:"timeout,omitempty" yaml:"timeout,omitempty" export:"true"`
+	Payload  string          `json:"payload,omitempty" toml:"payload,omitempty" yaml:"payload,omitempty" export:"true"`
+	Expected string          `json:"expected,omitempty" toml:"expected,omitempty" yaml:"expected,omitempty" export:"true"`
 }
--- a/pkg/healthcheck/tcp.go
+++ b/pkg/healthcheck/tcp.go
@@ -26,12 +26,24 @@ type ServiceTCPHealthChecker struct {
 }
 
 func NewServiceTCPHealthChecker(metrics metricsHealthCheck, config *dynamic.TCPServerHealthCheck, service StatusSetter, info *runtime.TCPServiceInfo, targets map[string]*net.TCPAddr, serviceName string) *ServiceTCPHealthChecker {
+	interval := time.Duration(config.Interval)
+	if interval <= 0 {
+		log.Error().Msg("Health check interval smaller than zero")
+		interval = time.Duration(dynamic.DefaultHealthCheckInterval)
+	}
+
+	timeout := time.Duration(config.Timeout)
+	if timeout <= 0 {
+		log.Error().Msg("Health check timeout smaller than zero")
+		timeout = time.Duration(dynamic.DefaultHealthCheckTimeout)
+	}
+
 	return &ServiceTCPHealthChecker{
 		balancer:    service,
 		info:        info,
 		config:      config,
-		interval:    time.Duration(config.Interval),
-		timeout:     time.Duration(config.Timeout),
+		interval:    interval,
+		timeout:     timeout,
 		metrics:     metrics,
 		targets:     targets,
 		serviceName: serviceName,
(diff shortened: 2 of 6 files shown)


pkg/server/service/tcp/service.go
```diff
@@ -91,10 +95,28 @@ func (m *Manager) BuildTCP(rootCtx context.Context, serviceName string) (tcp.Han
...
+				conf.LoadBalancer.HealthCheck,
+				loadBalancer,
+				conf,
+				healthCheckTargets,
+				serviceQualifiedName)
+		}
+
 		return loadBalancer, nil
 
 	case conf.Weighted != nil:
```
sdelicata, line 122: I noticed the health check mechanism is not implemented for WRR. Is it intended?

sdelicata (changes requested):

pkg/config/dynamic/tcp_config.go
```diff
@@ -172,3 +173,17 @@ func (t *TCPServersTransport) SetDefaults() {
...
+
+// +k8s:deepcopy-gen=true
+
+// TCPServer holds a TCP Server configuration.
+type TCPServerHealthCheck struct {
+	Address          string          `json:"address,omitempty" toml:"address,omitempty" yaml:"address,omitempty" label:"-"`
+	Port             string          `json:"-" toml:"-" yaml:"-"`
+	TLS              bool            `json:"tls,omitempty" toml:"tls,omitempty" yaml:"tls,omitempty"`
+	ServersTransport string          `json:"serversTransport,omitempty" toml:"serversTransport,omitempty" yaml:"serversTransport,omitempty" export:"true"`
+	Interval         ptypes.Duration `json:"interval,omitempty" toml:"interval,omitempty" yaml:"interval,omitempty" export:"true"`
```
sdelicata, line 185: We could go a little further by adding an unhealthy interval to the TCP health check configuration. It has been added for HTTP health check by https://github.com/traefik/traefik/pull/10610.
WDYT?

sdelicata (changes requested):

pkg/server/service/tcp/service.go
ddtmachado: Indeed, I was thinking WRR could come with a separate PR as there might be other things to consider with the healthcheck, but didn't actually look further on it.

ddtmachado:
Thanks for the review @sdelicata , I'll attempt to address those suggestions through the week

pkg/config/dynamic/tcp_config.go
```diff
@@ -172,3 +173,17 @@ func (t *TCPServersTransport) SetDefaults() {
...
 	t.TerminationDelay = ptypes.Duration(100 * time.Millisecond)
 }
+
+// +k8s:deepcopy-gen=true
+
+// TCPServer holds a TCP Server configuration.
+type TCPServerHealthCheck struct {
+	Address          string          `json:"address,omitempty" toml:"address,omitempty" yaml:"address,omitempty" label:"-"`
+	Port             string          `json:"-" toml:"-" yaml:"-"`
+	TLS              bool            `json:"tls,omitempty" toml:"tls,omitempty" yaml:"tls,omitempty"`
```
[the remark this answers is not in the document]
ddtmachado: my bad, I was not sure those fields were needed at all when starting, then forgot to remove it later

pkg/config/dynamic/tcp_config.go
ddtmachado: Not a fan of this new separate interval config, I would prefer a new interval policy config (could include a backoff strategy as well), but I'll follow that since its merged already.

pkg/healthcheck/tcp.go
```diff
@@ -0,0 +1,141 @@
...
+				Err(err).
+				Msg("Health check failed.")
+
+			isUp = false
+			serverUpMetricValue = float64(0)
+		}
+
+		thc.balancer.SetStatus(ctx, proxyName, isUp)
+
+		thc.info.UpdateServerStatus(target.String(), isUp)
```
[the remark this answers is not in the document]
ddtmachado: We could, I defaulted to bool since we only have two statuses so far, UP and DOWN. Let me know if you still prefer the string representation and I'll change it.

ddtmachado:
@supergoudvis116 Hello, unfortunately I'm doing this only on free time and without any commitment or deadline since I'm not in need for the feature anymore, but still would like to have it for the next time. 

I just updated the PR with the review suggestions and the missing WRR healthcheck support as well, but there might be more to cover before its in an acceptable state, let's see on the next review round.

pkg/server/service/tcp/service.go
```diff
@@ -91,14 +96,33 @@ func (m *Manager) BuildTCP(rootCtx context.Context, serviceName string) (tcp.Han
...
+				continue
+			}
+			healthCheckTargets[server.Address] = tcpAddr
+
 			loadBalancer.AddServer(handler)
 			logger.Debug().Msg("Creating TCP server")
 		}
 
+		if conf.LoadBalancer.HealthCheck != nil {
+			m.healthCheckers[serviceName] = healthcheck.NewServiceTCPHealthChecker(
```
sdelicata, line 111: Unless I'm wrong, the TCP health checkers are created but never launched. While HTTP health checkers are started via `serviceManager.LaunchHealthCheck(ctx)` in `pkg/server/routerfactory.go`, there's no equivalent call for TCP services.

pkg/healthcheck/tcp.go
```diff
@@ -0,0 +1,141 @@
...
+
+		buf := make([]byte, len(config.Expected))
+		_, err = conn.Read(buf)
+		if err != nil {
+			return err
+		}
+
+		if string(buf) != config.Expected {
+			return errors.New("unexpected response")
+		}
```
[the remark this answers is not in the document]
sdelicata: I'd appreciate it if the `executeHealthCheck` method provides more context about which part of the health check failed, for example, connection timeout vs payload vs response mismatch.
Maybe we could include the server address in error messages. WDYT?

pkg/healthcheck/tcp.go
```diff
@@ -0,0 +1,142 @@
...
+	dialer, err := thc.dialerManager.Get(config.ServersTransport, config.TLS)
+	if err != nil {
+		return err
+	}
+
+	conn, err := dialer.Dial("tcp", target.String())
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
```
sdelicata:
```suggestion
	defer func() {
        if err := conn.Close(); err != nil {
            log.Ctx(ctx).Warn().Err(err).Msg("Failed to close health check connection")
        }
    }()
```

pkg/healthcheck/tcp.go
```diff
@@ -0,0 +1,142 @@
...
+		logger.Error().Msg("Health check interval smaller than zero")
+		interval = time.Duration(dynamic.DefaultHealthCheckInterval)
+	}
+
+	timeout := time.Duration(config.Timeout)
+	if timeout <= 0 {
+		logger.Error().Msg("Health check timeout smaller than zero")
+		timeout = time.Duration(dynamic.DefaultHealthCheckTimeout)
+	}
+
```
sdelicata: WDYT about limiting the payload length to prevent memory issues?
Same with validating that the expected response length is reasonable?

pkg/server/service/tcp/service.go
```diff
@@ -91,14 +96,33 @@ func (m *Manager) BuildTCP(rootCtx context.Context, serviceName string) (tcp.Han
...
+				ctx,
+				m.dialerManager,
+				nil,
```
sdelicata: `metrics` parameter is passed as `nil` but should be properly integrated like HTTP health checks.

nmengin:
Interesting follow-up to keep in mind once the PR is merged: https://github.com/traefik/traefik/issues/10794#issuecomment-3106497538

Cc @rtribotte @kevinpollet

pkg/server/service/tcp/service.go
ddtmachado: as far as I remember TCP and UDP metrics are still not integrated, https://github.com/traefik/traefik/issues/7575, so if it's ok for you I would prefer to keep this out of this PR

pkg/healthcheck/tcp.go
ddtmachado: Good point, I'm adding a max payload size constant and falling back to ignoring it if exceeded, WDYT?

nmengin:
Hey @ddtmachado,

We'd like to move forward on this PR, do you agree if we push review commits?

Thanks in advance.

ddtmachado:
@nmengin of course! thanks!

traefiker:

traefiker:

traefiker:

supergoudvis116:
Since the issue's we were having a discussion on in as #10794 is closed I will post it here. 

As discussed I did manage to build a addition for the HTTP healthcheck on a TCP service. This PR is visible here:  #12606 

:)

PR #11242, opened 2024-10-29
feature: use ClusterIPs as source of addresses for an Ingress when Service type is ClusterIP

artuross:
### What does this PR do?

When Traefik is exposed as a Service with type ClusterIP, IP addresses assigned to the service are not propagated to Ingresses (see https://github.com/traefik/traefik/issues/7972 for details). This PR changes that.

### Motivation

I use Traefik as an Ingress Controller in my homelab cluster which is behind Tailscale. Due to how the routing works, I don't need to expose the Ingress Controller's Service externally. I've tried to use ClusterIP as it felt like a natural way to "expose" my service on the internal route only to learn this is not supported.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -402,19 +402,38 @@ func (p *Provider) updateIngressStatus(ing *netv1.Ingress, k8sClient Client) err
 		return fmt.Errorf("cannot get service %s, received error: %w", p.IngressEndpoint.PublishedService, err)
 	}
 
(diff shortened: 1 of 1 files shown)


kevinpollet:
Hi @artuross and thanks for opening this pull request,

It appears that this pull request is a duplicate of https://github.com/traefik/traefik/pull/11100. This pull request is in our pipeline, but we need to find time to review it. In the meantime, could you contribute/review it to make sure it addresses your issue?

Accordingly, I decline this pull request.

Duplicate of https://github.com/traefik/traefik/pull/11100

PR #11250, opened 2024-11-04
Configurable API & Dashboard base path

### LINKED ISSUE #5853: Allow to add path preifx for both api and dashboard endpoints

rtribotte:
### What does this PR do?

This PR adds an option to configure the base path where the API and dashboard will be exposed.

### Motivation

Fixes #5853 

### More

- [ ] Added/updated tests

### Additional Notes

### DIFF
--- a/docs/content/operations/dashboard.md
+++ b/docs/content/operations/dashboard.md
@@ -87,7 +87,43 @@ rule = "Host(`traefik.example.com`) && (PathPrefix(`/api`) || PathPrefix(`/dashb
 ??? example "Dashboard Dynamic Configuration Examples"
     --8<-- "content/operations/include-dashboard-examples.md"
 
+### Custom API Base Path
+
+As shown above, by default Traefik exposes its API and Dashboard under the `/` base path,
+which means that respectively the API is served under the `/api` path,
+and the dashboard under the `/dashboard` path.
+
+However, it is possible to configure this base path:
+
+```yaml tab="File (YAML)"
+api:
+  # Customizes the base path:
+  # - Serving API under `/traefik/api`
+  # - Serving Dashboard under `/traefik/dashboard`
+  basePath: /traefik
+```
+
+```toml tab="File (TOML)"
+[api]
+  # Customizes the base path:
+  # - Serving API under `/traefik/api`
+  # - Serving Dashboard under `/traefik/dashboard`
+  basePath = "/traefik"
+```
+
+```bash tab="CLI"
+# Customizes the base path:
+# - Serving API under `/traefik/api`
+# - Serving Dashboard under `/traefik/dashboard`
+--api.basePath=/traefik
+```
+
+??? example "Dashboard Under Custom Path Dynamic Configuration Examples"
+    --8<-- "content/operations/include-dashboard-custom-path-examples.md"
+
 ## Insecure Mode
+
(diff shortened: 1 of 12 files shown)


docs/content/operations/dashboard.md
```diff
@@ -104,8 +104,44 @@ rule = "Host(`traefik.example.com`) && (PathPrefix(`/api`) || PathPrefix(`/dashb
 ??? example "Dashboard Dynamic Configuration Examples"
     --8<-- "content/operations/include-dashboard-examples.md"
 
+### Custom API Base Path
+
+As shown above, by default Traefik exposes its API and Dashboard under the `/` base path,
+which means that respectively the API is served under the `/api` path,
+and the dashboard under the `/dashboard` path.
+
+However, it is possible to configure this base path:
```
mloiseleur, line 113:
```suggestion
It is possible to configure this base path:
```

pkg/api/dashboard/dashboard.go
```diff
@@ -39,24 +92,28 @@ func Append(router *mux.Router, customAssets fs.FS) {
...
-	if assets == nil {
-		assets = webui.FS
-	}
-	// allow iframes from our domains only
-	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src
-	w.Header().Set("Content-Security-Policy", "frame-src 'self' https://traefik.io https://*.traefik.io;")
+	router.Methods(http.MethodGet).
+		PathPrefix(dashboardPath).
+		HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			// allow iframes from our domains only
```
mloiseleur, line 106:
```suggestion
			// allow iframes from traefik domains only
```

lbenguigui (approved):
LGTM

docs/content/operations/dashboard.md
rtribotte: @mloiseleur I'm not sure about this suggestion, why suppressing this adverb?

pkg/api/dashboard/dashboard.go
```diff
@@ -1,35 +1,88 @@
...
+
+	// allow iframes from traefik domains only
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-src
+	w.Header().Set("Content-Security-Policy", "frame-src 'self' https://traefik.io https://*.traefik.io;")
+
+	// The content type must be guessed by the file server.
+	// https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Content-Type-Options
+	w.Header().Del("Content-Type")
+
+	if r.URL.Path == "/" {
```
jspdown, line 41: When testing using a `--api.insecure` I don't fall into this condition and end up with a non-templated `index.html`.

jspdown (changes requested):

jspdown (approved):
LGTM :+1:

PR #11260, opened 2024-11-08
docker swarm zero downtime rolling update

5andr0:
### What does this PR do?
This PR fixes https://github.com/traefik/traefik/issues/7886 by dropping containers from the load balancer as soon as there's a kill or stop event instead of waiting for the die event.

### More
- [ ] Added/updated tests
- [X] No updates necessary

### Additional Notes
More info here https://github.com/traefik/traefik/issues/11252
Edits by maintainers are welcome.
There's no need to resync all containers for all 3 kill -> stop -> die events.
You could make a set collection to store killed containers and don't resync on stop/die when the container is in the set. Then remove it from the set after the die event.
What I would like to see in a future PR is a function to add and remove single containers on the fly without resyncing all containers, which is a ton of overhead and unnecessary redundancy.

### Minimal setup to test the patch
```yaml
version: "3.9"
services:
  app:
    image: traefik/whoami
    networks:
      - swarm
    deploy:
      placement:
        constraints:
          - "node.role == manager"
      replicas: 1
      update_config:
        order: start-first
        parallelism: 1
        #delay: 20s
      labels:
        - "traefik.enable=true"
        - "traefik.http.routers.app.rule=PathPrefix(`/`)"
        - "traefik.http.services.app.loadbalancer.server.port=80"
        - "traefik.http.routers.app.entrypoints=web"
        #- traefik.docker.lbswarm=true

  load-balancer:
    image: traefik
    command:      
      - --providers.swarm
      - --providers.swarm.exposedbydefault=false
      - --providers.swarm.network=swarm
      #- --providers.swarm.refreshSeconds=1
    ports:
      - target: 80
        published: 88
        protocol: tcp
        mode: host
    volumes:
    deploy:
      placement:
        constraints:
          - "node.role == manager"
```

Deploy traefik and restart the dummy container
```
docker stack deploy -c test.yml test 
docker service update --force test_app
```
Run ApacheBench during update:
`ab -n 5000 -s 120 localhost:88/`

Expected output:
```
Complete requests:      5000
Failed requests:        0
```

### DIFF
--- a/pkg/provider/docker/pdocker.go
+++ b/pkg/provider/docker/pdocker.go
@@ -136,6 +136,8 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
 					case event := <-eventsc:
 						if event.Action == "start" ||
 							event.Action == "die" ||
+							event.Action == "kill" ||
+							event.Action == "stop" ||
 							strings.HasPrefix(string(event.Action), "health_status") {
 							startStopHandle(event)
 						}
@@ -166,7 +168,12 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
 }
 
 func (p *Provider) listContainers(ctx context.Context, dockerClient client.ContainerAPIClient) ([]dockerData, error) {
-	containerList, err := dockerClient.ContainerList(ctx, container.ListOptions{})
+	containerList, err := dockerClient.ContainerList(ctx, container.ListOptions{Filters: filters.NewArgs(
+		filters.Arg("status", "created"),
+		filters.Arg("status", "restarting"),
+		filters.Arg("status", "running"),
+	)})
+
 	if err != nil {
 		return nil, err
 	}


rtribotte:
Hello @5andr0,

Thanks for this contribution and sorry for the late reply.
At first glance, we are wondering why the changes are only targeting the docker provider?
The exposed use case seems to be with the swarm provider.

5andr0:
> Hello @5andr0,
> 
> Thanks for this contribution and sorry for the late reply. At first glance, we are wondering why the changes are only targeting the docker provider? The exposed use case seems to be with the swarm provider.

Hi, thanks for your input. I totally rushed this and didn't realize events work differently for swarm.
https://docs.docker.com/reference/cli/docker/system/events/#services
The container events are only available on the containers node.  
For services of other nodes we only get an update event on the manager node.  
I can't think of a solution based on that. Maybe listen for an update event and from that point increase the refresh ticker checks until the services are updated or failed so we don't have to set refreshSeconds to 1 second for all the time.
I don't have time to write a PR for that.  
Docker swarm just sucks, k8s is the way to go for better support.

Nonetheless this PR for the docker provider at least fixes the situation for blue-green deployments which are based on docker containers without swarm.

kevinpollet:
Hello @5andr0,

> Nonetheless this PR for the docker provider at least fixes the situation for blue-green deployments which are based on docker containers without swarm.

Can you elaborate on this point? How does this improve the situation?

5andr0:
> Hello @5andr0,
> 
> > Nonetheless this PR for the docker provider at least fixes the situation for blue-green deployments which are based on docker containers without swarm.
> 
> Can you elaborate on this point? How does this improve the situation?

For a simple docker web server it wouldn't make sense to setup swarm. So if you want to achieve no downtime updates there's a ton of tutorials with update scripts out there for the blue green approach.

Basically if you call the update script, it spins up an updated replica of the running container (let's say blue), waits until the new container (green) runs and then stops the old blue container and vice versa.

The problem here is that the stopped container still gets requests forwarded from traefik while it's in the process of stopping.
As soon as the kill signal from the stop command is received, the web server stops handling requests and sometimes it takes up to 10 sec until docker fully kills the container and the die event is registered in traefik. During this time there will be bad request errors or request delays because traefik thinks the stopped server is still online until the die event.

That's why this PR helps recognizing a stopped server as soon as the kill signal is received instead of waiting until the die signal.

rtribotte:
Hello @5andr0,

Thanks for your feedback!

We investigated and concluded that adding `kill` and `stop` events would only bring more opportunity to fetch the container list (and build the reflecting dynamic configuration).

However, in our tests, when receiving the `kill` or `stop` events, and thus during the graceful shutdown period (before the `die` event), the container still appears to have a `running` state during this period.

Since this state is the only criteria for Traefik to expose the container, we feel there is not much improvement we can do here.

We are closing this pull request accordingly, but feel free to ask for it to be reopened if you think we are missing something.

PR #11273, opened 2024-11-14
Send 301/302 for HEAD request in a redirect

dictcp:
### What does this PR do?

The PR make a redirect send 301 / 302 for HEAD request.

existing behaviour ref: https://github.com/traefik/traefik/issues/7829#issuecomment-770210080

### Motivation

Current code will response 301/302 for GET request in a redirect and 308/307 for HEAD request in similar cases.
However,

1. HEAD is considered a GET request with partial response (header only), so they should have the same response code
2. Since the response of GET request may be cached at CDN, and CDN may response afterward HEAD request via the cache copy. so this change will make thing more consistent.

### More

- [x] Added/updated tests
- ~~Added/updated documentation~~

### Additional Notes

N/A

### DIFF
--- a/pkg/middlewares/redirect/redirect.go
+++ b/pkg/middlewares/redirect/redirect.go
@@ -97,7 +97,7 @@ func (m *moveHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 	if m.permanent {
 		status = http.StatusMovedPermanently
-		if req.Method != http.MethodGet {
+		if req.Method != http.MethodGet && req.Method != http.MethodHead  {
 			status = http.StatusPermanentRedirect
 		}
 	}
--- a/pkg/middlewares/redirect/redirect.go
+++ b/pkg/middlewares/redirect/redirect.go
@@ -91,13 +91,13 @@ func (m *moveHandler) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	rw.Header().Set("Location", m.location.String())
 
 	status := http.StatusFound
-	if req.Method != http.MethodGet {
+	if req.Method != http.MethodGet && req.Method != http.MethodHead {
 		status = http.StatusTemporaryRedirect
 	}
 
 	if m.permanent {
 		status = http.StatusMovedPermanently
-		if req.Method != http.MethodGet && req.Method != http.MethodHead  {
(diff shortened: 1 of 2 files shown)


dictcp:
https://www.rfc-editor.org/rfc/rfc9110.html

> The HEAD method is identical to GET except that the server MUST NOT send content in the response. HEAD is used to obtain metadata about the selected representation without transferring its representation data, often for the sake of testing hypertext links or finding recent modifications.

> The server SHOULD send the same header fields in response to a HEAD request as it would have sent if the request method had been GET.

Per RFC9110, The HEAD method is identical to GET except that the server MUST NOT send content in the response.

So for redirect case, traefik should send 301/302 for HEAD request, the same as GET request.

dictcp:
Say we have a config to so www redirect to non-www domain like 

```yaml
http:
  routers:
    www-redirect:
      rule: "Host(`www.abc.com`)"
      entryPoints:
        - web
      middlewares:
        - redirect-to-root
      service: dummy  # No backend service needed as this is a redirect

  middlewares:
    redirect-to-root:
      redirectRegex:
        regex: "^https?://www\\.(abc\\.com)(.*)"
        replacement: "https://abc.com$2"
        permanent: true
```

For get request to https://www.abc.com, there will be a 301 redirect

However, for head request to it, it will be a 308 redirect .

My expectation is that, it should be a 301.

rtribotte:
This behavior was introduced by https://github.com/traefik/traefik/pull/4116.

> Per RFC9110, The HEAD method is identical to GET except that the server MUST NOT send content in the response.

> So for redirect case, traefik should send 301/302 for HEAD request, the same as GET request.

According to [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-301-moved-permanently) and also [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#permanent_redirections), the 301 redirect _MAY_ lead to a change to the GET method by the user agent.
On the other hand, the 308 redirect enforces that the method should remain the same.

We kept 301 for GET requests because they are not affected and it was less breaking.

The RFC doesn't look to recommend a specific status for a method.

> My expectation is that, it should be a 301.

What is the issue with 308?

kevinpollet:
Hello @dictcp and thanks for this contribution,

Please feel free to reopen if you need to.

PR #11281, opened 2024-11-17
Fix internal handlers in GlobalServiceBuilder

juliens:
### What does this PR do?

This PR changes the way we handle internal services. Before the PR, we had a proxy around the generic service builder, now we give specific builders to the generic service builder.

### Motivation

This PR fixes a bug when we referenced internal services in WRR, mirror or fallback.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/server/service/service.go
+++ b/pkg/server/service/service.go
@@ -53,23 +53,29 @@ type Manager struct {
 	transportManager httputil.TransportManager
 	proxyBuilder     ProxyBuilder
 
-	services       map[string]http.Handler
-	configs        map[string]*runtime.ServiceInfo
-	healthCheckers map[string]*healthcheck.ServiceHealthChecker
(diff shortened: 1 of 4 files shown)


kevinpollet (approved):

pkg/server/service/service.go
```diff
@@ -78,6 +86,17 @@ func (m *Manager) BuildHTTP(rootCtx context.Context, serviceName string) (http.H
 	serviceName = provider.GetQualifiedName(ctx, serviceName)
 	ctx = provider.AddInContext(ctx, serviceName)
 
+	// Must before we get configs to handle services without config
```
rtribotte, line 89:
```suggestion
	// Must be before we get configs to handle services without config.
```

rtribotte (changes requested):

rtribotte (approved):
LGTM

PR #11312, opened 2024-12-01
Service failover option on Kubernetes CRD

jjsiv:
### What does this PR do?

This is a basic implementation of service failover feature on Kubernetes CRD provider.

### Motivation

Currently failover is only configurable on the file provider. I've created an issue last year and there's been some interest in having this option also available on the Kubernetes provider https://github.com/traefik/traefik/issues/9919.

### More

- [ ] Added/updated tests


jjsiv:
This is more of a proof of concept implementation (no tests included or documentation is currently included) than anything and I'd mostly like to get some feedback on how we can proceed with this, if at all.

1. This cannot replicate 1:1 functionality of failover on file provider without, I believe, making major and possibly breaking changes to how services are currently created in Traefik for Kubernetes provider. Without NativeLB enabled, failover will not work at all when endpoints are not available for the main service (for example in case main service pods cannot be created). It does work generally ok with NativeLB enabled, although still with some limitations compared to the file provider (e.g. won't work if wrong port is specified for main service).
2. Implementation depends on healthcheck being available for use on services. Healtcheck was implemented here https://github.com/traefik/traefik/pull/10467 although for ExternalName services only. This PR enables it for all other service types, although perhaps this should be another PR instead? Are there any considerations when enabling it?

If I could get some input on this or any other related feedback before I would be grateful.

jjsiv:
Thanks for your feedback @kevinpollet. I agree, to rely on built-in Kubernetes mechanisms would be ideal and the first approach sounds good. I will see what I can come up with.

Issue #11313, opened 2024-12-02
Middleware of ForwardAuth returns incorrect location header


cirias:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I made an auth server that returns a relative url as the Location header in the response. I'm expecting the relative url will be passed to the downstream and then be resolved by the browser or other clients.

### What did you see instead?

ForwardAuth middleware resolves the location header and return the resolved url as the location header in the response. While the correct behavior should just return the original location header.

https://github.com/traefik/traefik/blob/33c1d700c02ad09ab793673a14674b39d78fe352/pkg/middlewares/auth/forward.go#L193

For example,
- Auth server endpoint address: http://forward-auth.server.local:8080/auth
- Auth server response Location header: /index.html
- Downstream received response Location header: http://forward-auth.server.local:8080/index.html
- Expected Location header: /index.html

Note, that the location should be resolved at the custom side, like the browser, which should relative to the original forwarded domain, not the auth server.

### What version of Traefik are you using?

v3.1.6
I checked the latest code. It seems it's still the same behavior.

### What is your environment & configuration?

Not related. But can provide if being asked.

Nelwhix:
@emilevauge I made a PR to fix this, Please review

PR #11317, opened 2024-12-03
Forward auth fix

Nelwhix:
### What does this PR do?
Fixing issue #11313 where ForwardAuth middleware modifys the Location header instead of sending it as is.

### More

- [x] Added/updated tests


Nelwhix:
I forked master instead of 3.2 branch. fixing

PR #11318, opened 2024-12-03
fix: fixing issue #11313 where ForwardAuth middleware modifys the Location header instead of sending it as is.

Nelwhix:
### What does this PR do?
Fixing issue #11313 where ForwardAuth middleware modifys the Location header instead of sending it as is.

### More

- [x] Added/updated tests


kevinpollet:
Hello @Nelwhix and thanks for this contribution,

The way you address the issue is breaking because some users may rely on the current behavior of the ForwardAuth middleware. We may have not been clear enough, but the issue should be addressed in a non-breaking way. This means that an option should be added to the middleware to control whether the location header should be left as is (the new behavior) or resolved against the auth server address.

Also, as this is an enhancement, could you please rebase it on the master branch?

Nelwhix:
Hi @kevinpollet, I understand what you mean. Now to find a new name for this field. What do you think of `preserveLocationHeader` ?

```yaml
http:
  middlewares:
    test-auth:
      forwardAuth:
        preserveLocationHeader: true
```

kevinpollet:
> Hi @kevinpollet, I understand what you mean. Now to find a new name for this field. What do you think of `preserveLocationHeader` ?

@Nelwhix Looks good to me.

Nelwhix:
@kevinpollet made the changes

rtribotte:
Hello @Nelwhix,

Thanks for your changes.
This new option needs to be documented, both in configuration references (you just need to run `go generate` to update them), and in the middleware specific page.
As well, the option needs to be replicated in our CRDs by adding it to `traefikv1alpha1.ForwardAuth`, and make use of it in `createForwardAuthMiddleware` func.

rtribotte:
Hello @Nelwhix,

Thanks for last changes!
Do you mind if I rebase you branch and push a review commit?

Nelwhix:
No problem.

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #11321, opened 2024-12-05
fix:  wasm settings

juliens:
### What does this PR do?

This PR fixes a bug that prevented WASM plugins, loaded using wazero’s WASI preview 1 implementation, from properly utilizing environment variables or mounted files specified in the plugin settings.

### Motivation

Fix the bug.

### More

- [x] Added/updated tests

### Additional Notes

For now, only test on WASM Plugins using wazero and compiled with tinygo are implemented. In the futur, we should implement more cases like plugins compiled with native go 1.24 ( with wasmexport ).


rtribotte (approved):
LGTM

kevinpollet (approved):

PR #11343, opened 2024-12-14
Pass TLS bool from IngressRouteTCP to TCPService

lipmem:
### What does this PR do?

Pass TLS bool value from IngressRouteTCP to TCPServer to allow to enable TLS encrypted communication between traefik and downstream service.

### Motivation
I deployed traefik as k8s Ingress. TLS is terminated at traefik, but I want to communicate via TLS inside the cluster. This is not possible right now with the given options of IngressRouteTCP.

### More

- [x] Added/updated tests


kevinpollet (changes requested):

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👏

PR #11350, opened 2024-12-16
Replace experimental `maps` and `slices` with stdlib

Juneezee:
### What does this PR do?

This PR replaces `golang.org/x/exp/maps` and `golang.org/x/exp/slices` with `maps` and `slices` from the standard library.

The key difference is that `maps.Keys` in the `golang.org/x/exp` package return a slice, whereas `maps.Keys` in the standard library return an iterator. To work with slices, we need to use `slices.Collect` to convert the iterator into a slice.

### Motivation

We upgraded to Go 1.23 in https://github.com/traefik/traefik/pull/11035. The experimental functions are now part of the standard library as of Go 1.21 and Go 1.23.

Reference: https://go.dev/doc/go1.21#slices
Reference: https://go.dev/doc/go1.23#iterators

### More

- [ ] Added/updated tests

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -7,10 +7,11 @@ import (
 	"fmt"
 	"io"
 	stdlog "log"
+	"maps"
 	"net/http"
 	"os"
 	"os/signal"
-	"sort"
+	"slices"
 	"strings"
 	"syscall"
 	"time"
@@ -48,7 +49,6 @@ import (
 	"github.com/traefik/traefik/v3/pkg/tracing"
 	"github.com/traefik/traefik/v3/pkg/types"
 	"github.com/traefik/traefik/v3/pkg/version"
-	"golang.org/x/exp/maps"
 )
 
 func main() {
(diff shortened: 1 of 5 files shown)


cmd/traefik/traefik.go
```diff
@@ -427,7 +427,7 @@ func getDefaultsEntrypoints(staticConfiguration *static.Configuration) []string
 		}
 	}
 
-	sort.Strings(defaultEntryPoints)
+	slices.Sort(defaultEntryPoints)
```
Juneezee, line 430: As of Go 1.22, `sort.Strings` simply calls `slices.Sort`. See https://pkg.go.dev/sort#Strings

pkg/api/sort.go
```diff
@@ -357,7 +356,7 @@ func sortByName[T orderedWithName](direction string, results []T) {
 	})
 }
 
-func sortByFunc[T orderedWithName, U constraints.Ordered](direction string, results []T, fn func(int) U) {
+func sortByFunc[T orderedWithName, U cmp.Ordered](direction string, results []T, fn func(int) U) {
```
Juneezee, line 359: The `constraints.Ordered` from `golang.org/x/exp/constraints` is also included into Go 1.21 as `cmp.Ordered`. See https://pkg.go.dev/cmp#Ordered

Juneezee (commented):
Reviewing my own code changes

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #11351, opened 2024-12-17
Add passive health checks

### LINKED ISSUE #11306: Passive health check
### Welcome!

As many reverse proxy server has this function, I wonder while traefik does not.
* nginx: https://docs.nginx.com/nginx/admin-guide/load-balancer/http-health-check/
* apisix: https://apisix.apache.org/docs/apisix/2.13/health-check/

I think the design of nginx can meet most of the requirements.
[fail_timeout](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#fail_timeout) – Sets the time during which a number of failed attempts must happen for the server to be marked unavailable, and also the time for which the server is marked unavailable (default is 10 seconds).
[max_fails](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#max_fails) – Sets the number of failed attempts that must occur during the fail_timeout period for the server to be marked unavailable (default is 1 attempt).

Nelwhix:
### What does this PR do?

Adding Passive Health Checks 

### Motivation

Fixes #11306 

### More

- [ x] Added/updated tests


Nelwhix:
My goal for how this will work will be 
```yaml
http:
  services:
    my-service:
      loadBalancer:
        servers:
        - url: "http://<private-ip-server-1>:<private-port-server-1>/"
           healthCheck:
             maxFails: 3
             failTimeout: 30
```
I added the implementation to just http services for now.

pkg/healthcheck/healthcheck.go
```diff
@@ -55,6 +57,13 @@ type ServiceHealthChecker struct {
 	serviceName string
 }
 
+type PassiveHealthChecker struct {
+	mu          sync.Mutex
+	Failures    []time.Time
+	maxFail     int
+	failTimeout ptypes.Duration
+}
```
youkoulayley, line 65: Could you move this struct below, next to the `NewPassiveHealthChecker` function ?

pkg/healthcheck/healthcheck.go
```diff
@@ -263,3 +272,38 @@ func (shc *ServiceHealthChecker) checkHealthGRPC(ctx context.Context, serverURL
...
+func (p *PassiveHealthChecker) AllowRequest() bool {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	// Clean up old Failures outside the sliding window
+	now := time.Now()
+	threshold := now.Add(time.Duration(-p.failTimeout))
+	var filteredFailures []time.Time
+
+	for _, t := range p.Failures {
```
youkoulayley, line 293:
```suggestion
	threshold := now.Add(time.Duration(-p.failTimeout))

	var filteredFailures []time.Time
	for _, t := range p.Failures {
```

youkoulayley (changes requested):
Hello @Nelwhix 
Thanks for your contribution.

I have a few suggestions.

pkg/healthcheck/healthcheck.go
```diff
@@ -55,6 +57,13 @@ type ServiceHealthChecker struct {
 	serviceName string
 }
 
+type PassiveHealthChecker struct {
+	mu          sync.Mutex
+	Failures    []time.Time
```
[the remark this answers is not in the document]
Nelwhix: true, changed it

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -256,12 +305,33 @@ func (b *Balancer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+				if errors.Is(err, errNoAvailableServer) {
+					http.Error(w, errNoAvailableServer.Error(), http.StatusServiceUnavailable)
+				} else {
+					http.Error(w, err.Error(), http.StatusInternalServerError)
+				}
+				return
+			}
+		}
+
+	}
```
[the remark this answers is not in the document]
Nelwhix: added

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -4,6 +4,7 @@ import (
 	"container/heap"
 	"context"
 	"errors"
+	"github.com/traefik/traefik/v3/pkg/healthcheck"
```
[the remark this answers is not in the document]
Nelwhix: done

pkg/healthcheck/healthcheck.go
Nelwhix: okay

pkg/healthcheck/healthcheck.go
```diff
@@ -8,8 +8,11 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"sync"
 	"time"
 
+	ptypes "github.com/traefik/paerser/types"
+
```
youkoulayley, line 15:
```suggestion
```

youkoulayley (changes requested):
Just a small import to fix remaining

youkoulayley (approved):
LGTM

jspdown:
Hello @Nelwhix and thanks for your contribution.

Before reviewing the content of the PR I wanted to discuss how this feature is configured. 
First of all, I really like the `failTimeout` and `maxFails`, it's very easy to understand. In the future, if needed, we could even add an additional `recoverTimeout` to define how long to wait before recovering a server, and this without breaking change. 

Though, I'm wondering if we really want this to be configured at the server level. This would be the right spot if we wanted each server to have a different passive healthcheck config, or a mix of server with and without passive healthcheck under the same loadbalancer. I might be wrong but I have the feeling that most of the time we want to have a uniform config on the loadbalancer. 

Did you considered moving this configuration one layer up in the loadbalancer?

Nelwhix:
Hi @jspdown, I was trying to achieve nginx style passive health checks https://docs.nginx.com/nginx/admin-guide/load-balancer/http-health-check/

sdelicata (changes requested):
Hi @Nelwhix! Thanks for your contribution.
I've made a few comments/suggestions.

Once the design and configuration approach for passive health checks is finalized, we'll need to add documentation and regenerate the reference configuration accordingly.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -256,12 +306,30 @@ func (b *Balancer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		http.SetCookie(w, cookie)
 	}
 
-	server.ServeHTTP(w, req)
+	customWriter := &CustomResponseWriter{ResponseWriter: w, StatusCode: http.StatusOK}
```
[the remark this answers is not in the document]
Nelwhix: what if I modify the loop in nextServer to:

```go
for {
  // Pick handler with closest deadline.
  
  // curDeadline should be handler's deadline so that new added entry would have a fair competition environment with the old ones.
  
  heap.Push(b, handler)
  
  // Skip if:
  if _, down := b.status[handler.name]; !down {
      continue
  }

  if _, isFenced := b.fenced[handler.name]; isFenced {
      continue
  }

  if handler.passiveHealthChecker != nil && !handler.passiveHealthChecker.AllowRequest() {
      continue
  }
}
```

With this we don't need nextServerExcluding... the filtering happens when selecting a server

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -68,6 +72,16 @@ type Balancer struct {
 	fenced map[string]struct{}
 }
 
+type CustomResponseWriter struct {
```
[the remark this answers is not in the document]
Nelwhix: dibe

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -9,15 +9,19 @@ import (
...
 
 type namedHandler struct {
 	http.Handler
-	name     string
-	weight   float64
-	deadline float64
+	name                    string
+	weight                  float64
+	deadline                float64
+	wantsPassiveHealthCheck bool
```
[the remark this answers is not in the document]
Nelwhix: fixed

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -208,6 +222,42 @@ func (b *Balancer) nextServer() (*namedHandler, error) {
 	return handler, nil
 }
 
+func (b *Balancer) nextServerExcluding(handlerName string) (*namedHandler, error) {
```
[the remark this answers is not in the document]
Nelwhix: added

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -256,12 +306,30 @@ func (b *Balancer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		http.SetCookie(w, cookie)
 	}
```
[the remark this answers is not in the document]
Nelwhix: fixed, also added tests to check

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -306,8 +307,8 @@ func (b *Balancer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
 
-	customWriter := &CustomResponseWriter{ResponseWriter: w, StatusCode: http.StatusOK}
-	if server.wantsPassiveHealthCheck && !server.passiveHealthChecker.AllowRequest() {
+	customWriter := &StatusRecorder{ResponseWriter: w, StatusCode: http.StatusOK}
```
sdelicata, line 310: Could you rename the `customWriter` var accordingly, please?

pkg/server/service/loadbalancer/wrr/wrr.go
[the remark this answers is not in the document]
sdelicata: Yes, personally, I prefer this approach.

sdelicata (changes requested):

pkg/server/service/loadbalancer/wrr/wrr.go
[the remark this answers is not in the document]
Nelwhix: Done

docs/content/routing/services/index.md
```diff
@@ -533,6 +533,55 @@ Below are the available options for the health check mechanism:
...
+Passive health checks rely on real traffic to determine server health. 
+Traefik will forward requests as usual and evaluate each response or timeout, 
+incrementing a failure counter when a request fails. 
+Once the number of failures within a given period exceeds your configuration, 
+Traefik will automatically stop routing traffic to that server until it recovers.
+
+Below are the available options for passive health checks:
+- maxFails (default: 1) – the number of failed requests that must occur before the server is marked unhealthy.
+
+- failTimeout (default: 10s) – the window during which failures are counted, and the duration the server remains marked unhealthy.
```
sdelicata, line 546:
```suggestion
#### Passive Health Checks
Passive health checks rely on real traffic to assess server health.
Traefik forwards requests as usual and evaluates each response or timeout,
incrementing a failure counter whenever a request fails.
If the number of failures within a specified time window exceeds the configured threshold,
Traefik will automatically stop routing traffic to that server until it recovers.

The following options are available for passive health checks:
- maxFails (default: 1) –  The number of failed requests required before the server is considered unhealthy.

- failTimeout (default: 10s) – The time window during which failures are counted, and the duration for which the server remains marked as unhealthy.
```

pkg/config/dynamic/http_config.go
```diff
@@ -293,13 +293,17 @@ type Server struct {
...
 	Port   string `json:"-" toml:"-" yaml:"-" file:"-" kv:"-"`
-	// HealthCheck enables passive health checks for this server
+	// HealthCheck enables passive health checks for this server.
 	HealthCheck *PassiveHealthCheck `json:"healthCheck,omitempty" toml:"healthCheck,omitempty" yaml:"healthCheck,omitempty" export:"true"`
 }
 
+// +k8s:deepcopy-gen=true
+
 type PassiveHealthCheck struct {
+	// FailTimeout sets the time during which a number of failed attempts must happen for the server to be marked unavailable.
```
sdelicata, line 303:
```suggestion
	// FailTimeout defines the time window during which a certain number of failed attempts must occur for the server to be marked as unavailable.
```

pkg/config/dynamic/http_config.go
```diff
@@ -293,13 +293,17 @@ type Server struct {
...
 	FailTimeout ptypes.Duration `json:"failTimeout,omitempty" toml:"failTimeout,omitempty" yaml:"failTimeout,omitempty" export:"true"`
```
sdelicata, line 304: The current name `FailTimeout` is ambiguous, as it suggests a request timeout or delay between retries.
Renaming it to `FailureWindow` (or `FailurePeriod`) better reflects its actual role: defining the time frame during which failed attempts are counted before marking a server as unhealthy. WDYT?

pkg/config/dynamic/http_config.go
```diff
@@ -293,13 +293,17 @@ type Server struct {
...
-	MaxFails    int             `json:"maxFails,omitempty" toml:"maxFails,omitempty" yaml:"maxFails,omitempty" export:"true"`
+	// MaxFails sets the number of failed attempts that must occur during the fail timeout.
```
sdelicata, line 305:
```suggestion
	// MaxFails is the number of failed attempts allowed within the failure window before marking the server as unhealthy.
```

pkg/config/dynamic/http_config.go
```diff
@@ -293,13 +293,17 @@ type Server struct {
...
+	MaxFails int `json:"maxFails,omitempty" toml:"maxFails,omitempty" yaml:"maxFails,omitempty" export:"true"`
```
sdelicata, line 306: `MaxFails` is slightly ambiguous to me.
WDYT about replacing it with `FailureThreshold` (or `MaxFailedAttempts`) to improve clarity, making it easier to understand that it sets the limit of allowed failures within the failure window before a server is marked as unhealthy?

sdelicata (changes requested):

pkg/config/dynamic/http_config.go
Nelwhix: I changed to FailureWindow

pkg/config/dynamic/http_config.go
Nelwhix: I changed to MaxFailedAttempts

sdelicata (dismissed):
Thanks @Nelwhix for your improvements.
There are still some linting errors to fix, then we should be ready.

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -162,12 +174,17 @@ func (b *Balancer) nextServer() (*namedHandler, error) {
 		handler.deadline += 1 / handler.weight
 
 		heap.Push(b, handler)
-		if _, ok := b.status[handler.name]; ok {
-			if _, ok := b.fenced[handler.name]; !ok {
-				// do not select a fenced handler.
-				break
-			}
+		if _, down := b.status[handler.name]; !down {
```
lbenguigui, line 177: Unless I missed something, `b.status` indicates whether a server is up. Maybe renaming `down` to something like up would make the logic clearer?

pkg/server/service/loadbalancer/wrr/wrr.go
```diff
@@ -162,12 +174,17 @@ func (b *Balancer) nextServer() (*namedHandler, error) {
...
+			continue
+		}
+		if _, isFenced := b.fenced[handler.name]; isFenced {
+			continue
+		}
+		if handler.passiveHealthChecker != nil && !handler.passiveHealthChecker.AllowRequest() {
```
lbenguigui, line 183: I think there might be an issue: if there's one server up but it has a passive health check that disallows requests, we wait until `failTimeout` expires. 
Is that the intended behavior? Should we instead return an `errNoAvailableServer` in this case?

lbenguigui (dismissed):

kevinpollet:
Hello @Nelwhix and thanks for this contribution,

After some consideration, we think that the passive health check mechanism should be configured at the load balancer level, in line with the active health checker configuration (see https://doc.traefik.io/traefik/routing/services/#health-check). This means that, once configured, the passive health check will be enabled for all servers.

We also think that it should reuse the current status updater mechanism. This would make the passive health checker responsible for propagating the server status to its parent (the load balancer). It should also be responsible for updating the `ServiceInfo`, which is needed to update the server status on the UI, as well as the corresponding `ServerUpGauge` metric (see https://github.com/traefik/traefik/blob/master/pkg/healthcheck/healthcheck.go#L135). This means that nothing needs to be changed in the WRR code, as the passive health checker will be integrated in the same way as the active health checker.

Regarding the passive health check algorithm, having read the documentation (https://docs.nginx.com/nginx/admin-guide/load-balancer/http-health-check/), we propose that, once a server is down, we should wait for the `FailureWindow` duration to expire before sending another request to the server. This means that, once a server is considered down, the passive health checker will have to clear its failures slice, after which a timer should be set to update the server status to 'up' once the `FailureWindow` duration has expired.

Does it make sense?
Are you ok with updating this pull request in that direction?

Nelwhix:
@kevinpollet Okay Kevin, I will rework

Nelwhix:
Hi @kevinpollet , I was able to implement most of your suggestions. except for the passive health checker to update the ServiceInfo and serverupguage. I tried to pass in the metrics and serviceinfo to the `wrr.Balancer` and then use the healthchecker to do `p.balancer.info.update` but that will lead to cyclic import errors. Please could you point me in a better direction?

kevinpollet:
Hello @Nelwhix, and apologies for the delay.

It's a bit difficult to reason through the cyclic import issue without working directly with the code.
Would it be okay with you if we push review commits while conducting the pull request review?

Nelwhix:
Okay @kevinpollet

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #11361, opened 2024-12-20
Fix fenced server status computation

kevinpollet:
### What does this PR do?

This pull request fixes the computation of the fenced server status. A fenced server is necessarily a server with an endpoint slice condition set to `terminating` (when an endpoint has the `ready` condition set to `true` it also has the `serving` condition set to `true`). From the Go documentation:

```
serving is identical to ready except that it is set regardless of the terminating state of endpoints.
```

### Motivation

To fix a regression.

### More

- [X] Added/updated tests

### Additional Notes


nmengin (approved):
LGTM

juliens (approved):
LGTM

PR #11430, opened 2025-01-10
Optimize PNG image losslessly: 6.2 → 4.6 MiB (−25%)

maximal:


nmengin:
Hey @maximal,

Could you provide more information about your motivation, and expectations?

Thanks in advance.

maximal:
@nmengin, just faster documentation loading, nothing more. I like to optimize images.

nmengin:
Hey @maximal,

Thank you for the quick answer.

Discussing the PR with the maintainers' team, we think that would be relevant to automate the images optimization using the CI.

Could you share with us the tooling you have used, please?

maximal:
@nmengin, I used [OptiPNG](https://optipng.sourceforge.net/) tool with the maximum number of tries when bruteforcing PNG parameters:

```shell
# For one PNG file
optipng -o7 -zm1-9 file.png
# For all PNG files in the directory
find . -type f -name '*.png' -exec optipng -o7 -zm1-9 '{}' \;
```

Note, that this process can take a while, so I’d be very careful to use it in the CI/CD workflow.

emilevauge:
Dear @maximal, thanks a lot for your contribution.
After discussing this with the maintainer team, we made the decision to not merge this PR:

- The actual documentation is already served behind a CDN which already optimizes and caches images on the fly
- We would need to add this process to our CI/CD to make this change maintainable on the long term, and we don't have time to spend on this right now
- [OptiPNG](https://optipng.sourceforge.net/) does not seem maintained anymore (last version in Nov 2023)

We might get back to this topic in the future though :)

maximal:
@emilevauge, thanks for the answer. Here’s my thoughts:

> CDN which already optimizes and caches images on the fly

CDN cannot run these types of optimizations on the fly, because it takes a relatively large amount of time (a minute and longer per every image).

> [OptiPNG](https://optipng.sourceforge.net/) does not seem maintained anymore (last version in Nov 2023)

Currently, I don’t see any working alternatives. 2023 version hardly can be referred as old or discontinued (IMO).

> We would need to add this process to our CI/CD to make this change maintainable on the long term, and we don't have time to spend on this right now

That point looks reasonable to me. I just proposed the solution which can improve some things right here and right now, without waiting or investing time in long-term CI/CD supporting which may never be implemented. Anyway, you’re the host here, so it’s your decision.

Cheers!

PR #11455, opened 2025-01-16
feat: Add 'auto' webui theme option and default to it

zizzfizzix:
### What does this PR do?

This adds 'auto' as a third option to light, and dark webui theme that will adapt to the users' device automatically and defaults to it.
The user can still switch the theme in the nav bar and cycle through all three options.

### Motivation

Respect the users' device setting instead of defaulting to light and making the user switch manually every time their device changes theme (e.g. when it's getting dark).

### More

- [ ] Added/updated tests

### DIFF
--- a/webui/src/App.vue
+++ b/webui/src/App.vue
@@ -14,8 +14,10 @@ export default {
     ...mapGetters('core', { coreVersion: 'version' })
   },
   watch: {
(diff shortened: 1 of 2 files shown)


kevinpollet (approved):
Thanks 👏

rtribotte (approved):
Thanks!

PR #11473, opened 2025-01-20
Added option to preserve request method in forwardAuth

an09mous:
### What does this PR do?

This PR adds an option in forwardAuth middleware to allow preserve the original request method when the request reaches the authentication server. By default, the original request method is not respected and auth

### Motivation

Fixes: https://github.com/traefik/traefik/pull/11097

### More

- [ ] Added/updated tests


an09mous:
Hi @nmengin 
The PR is ready for review, just wanted to double check on the local validations before sending for review.

docs/content/middlewares/http/forwardauth.md
```diff
@@ -747,4 +747,46 @@ http:
 ```
 
 
+### `preserveRequestMethod`
+
+_Optional, Default=false_
+
+`preserveRequestMethod` defines whether to preserve the original request `method` while forwarding the request to the authentication server. By default, if it set to `false`, the request method is changed to `GET` when it reaches the authentication server.
```
dargmuesli, line 754:
```suggestion
`preserveRequestMethod` defines whether to preserve the original request method while forwarding the request to the authentication server. By default. when this option is set to `false`, incoming requests are always forwarded as `GET` requests to the authentication server.
```

dargmuesli (commented):
Looking good!

docs/content/middlewares/http/forwardauth.md
an09mous: There is a typo in the suggestion. Comma should be there instead of full stop after **By default**
```suggestion
`preserveRequestMethod` defines whether to preserve the original request method while forwarding the request to the authentication server. By default, when this option is set to `false`, incoming requests are always forwarded as `GET` requests to the authentication server.
```

an09mous: This is fixed

an09mous:
Updates:
1. The generate-crd check was failing due to what looks to be an issue with the formatting of the preserveRequestMethod description. Fixed it by running the check locally and accepting the formatting as suggested at the completion of the command.
2. Merged the updated `master` branch into this branch

docs/content/middlewares/http/forwardauth.md
```diff
@@ -747,4 +747,46 @@ http:
 ```
```
rtribotte, line 749:
```suggestion
```

docs/content/middlewares/http/forwardauth.md
```diff
@@ -747,4 +747,46 @@ http:
...
+        # ...
+        preserveRequestMethod: true
+```
+
+```toml tab="File (TOML)"
+[http.middlewares.test-auth.forwardAuth]
+  # ...
+  preserveRequestMethod = true
+```
+
```
rtribotte, line 790:
```suggestion
```

rtribotte (changes requested):

an09mous:
The suggested changes have been addressed in the latest commit.

kserard:
Would it be possible to set the request method explicitly? Before this CR, the `forwardAuth` method is always `GET`, with it, it preserves the original method, but would it be possible to set to use always `POST`, whatever the original method is?

an09mous:
Hi @kserard 

I can definitely see the genuine need of this behaviour when we want to use the body in the auth server to do the auth and don't want to implement APIs for all these methods in the auth server duplicating the same behaviour.

But I feel, this should be a separate enhancement to introduce this option. I would like maintainers to comment on it.
cc: @rtribotte

rtribotte:
Hello @kserard,

@an09mous said it all! It is a different feature that needs be addressed on its own. Could you please open a dedicated issue for it?

pkg/middlewares/auth/forward.go
```diff
@@ -168,6 +170,10 @@ func (fa *forwardAuth) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		}
 	}
 
+	if fa.preserveRequestMethod {
+		forwardReq.Method = req.Method
+	}
+
```
rtribotte, line 176: This is nitpicking, but an alternative could have been to create a variable for the method before giving it to `http.NewRequestWithContext`. It would have the advantage to group concerns and probably improve reading. WDYT?

rtribotte: I mean it could have had the advantage of also validating the method when building the request, but the original request method doesn't need that.

kserard:
@an09mous, @rtribotte
Thanks for your kind comments, I've opened #11482 for this new feature request.

pkg/middlewares/auth/forward.go
an09mous: Makes sense, will update it

an09mous: Updated

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

dargmuesli:
@rtribotte excuse the question, would it be possible to release a `3.4.0-rc1` containing this change? I know about the release cycle and think this will be released in about 1.5 months, but I'd like to gladly test this in an official release candidate now already if that helps.

ldez:
@dargmuesli you can use the experimental tag: [`traefik/traefik/experimental-master`](https://hub.docker.com/layers/traefik/traefik/experimental-master/images/sha256-e05b0df8f4d1cd4bf8b579a184e376fc6061016f4ae2b531c74890e434b97192)

dargmuesli:
Thank you, I can confirm it to be working well :+1:

dargmuesli:
End of month afaik. I deployed in [one of my Docker stacks](https://github.com/maevsi/stack).

sambit-mahapatra:
I tried deploying this image in my EKS cluster & it seems to throw a lot of errors saying it's unstable in the logs. I will wait for the release & meanwhile go through your docker stack. @dargmuesli

sambit-mahapatra:
@rtribotte & @ldez I know I am bothering but Can you share the timeline for v3.4 release with this commit? I mean the "preserveRequestType" changes.

rtribotte:
@dargmuesli @sambit-mahapatra [v3.4.0-rc1](https://github.com/traefik/traefik/releases/tag/v3.4.0-rc1) is available!

PR #11479, opened 2025-01-22
fix(tracing): panic when calling Tracer

basgys:
### What does this PR do?

This fixes a panic when calling `Tracer` on TracerProvider with a different name than “github.com/traefik/traefik”. This happens because `t.TracerProvider` is nil.

Instead, I replaced it with `otel.GetTracerProvider()`, which returns the default global tracer provider. If that behaviour is not desired, I would recommend using `noop.NewTracerProvider()` instead.

I created a unit test to demonstrate the problem.

### Motivation

To prevent panics.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/tracing/tracing.go
+++ b/pkg/tracing/tracing.go
@@ -117,7 +117,7 @@ func (t TracerProvider) Tracer(name string, options ...trace.TracerOption) trace
 		return t.tracer
 	}
 
-	return t.TracerProvider.Tracer(name, options...)
+	return otel.GetTracerProvider().Tracer(name, options...)
 }
 
 // Tracer is trace.Tracer with additional properties.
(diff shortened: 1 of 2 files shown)


rtribotte:
Hello @basgys,

Thanks for reporting this!

Something is indeed wrong in this code, but we don't understand the situation in which another trace provider could be requested.
We wonder what the context is for this panic?

basgys:
@rtribotte Hi! Thanks for your reply.

We have a custom middleware that calls a third-party library, which uses OpenTelemetry. Since that library invokes Tracer with a different name than `github.com/traefik/traefik`, it triggers this line, causing a panic: https://github.com/traefik/traefik/blob/master/pkg/tracing/tracing.go#L120.

This issue is unlikely to affect standard Traefik packages, as they all match this condition: https://github.com/traefik/traefik/blob/master/pkg/tracing/tracing.go#L117.

However, it remains a potential issue and will impact Traefik in the future if it interacts with third-party libraries that rely on OpenTelemetry.

sdelicata (approved):
LGTM

jspdown (approved):

PR #11487, opened 2025-01-24
Do not create a logger instance for each proxy

### LINKED ISSUE #11423: Increased memory usage in version 2.11.18
### Welcome!

### What did you do?

Upgrade from version 2.11.16 to 2.11.18. We did not expect any change in memory usage.

### What did you see instead?

It seems Træfik now uses at least 2x as much memory as before, and it may even be leaking memory.

Image
Image

2.11.18

N/A

_No response_

kevinpollet:
### What does this PR do?

This pull request fixes a memory and go routine leak introduced by https://github.com/traefik/traefik/pull/11344. After some investigation, it appears that the logrus [writer](https://github.com/sirupsen/logrus/blob/d1e6332644483cfee14de11099f03645561d55f8/writer.go#L57) goroutine is never stopped, creating a memory and go routine leak at each dynamic configuration refresh. To mitigate this, proxies now share an `errorLogger` instance. 

### Motivation

Fixes #11423

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/server/service/proxy.go
+++ b/pkg/server/service/proxy.go
@@ -26,6 +26,12 @@ const StatusClientClosedRequest = 499
 // StatusClientClosedRequestText non-standard HTTP status for client disconnection.
 const StatusClientClosedRequestText = "Client Closed Request"
 
+// errorLogger is a logger instance used to log proxy errors.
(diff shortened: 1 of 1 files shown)


nmengin (approved):
LGTM 👍

PR #11495, opened 2025-01-30
Set check-latest to true in Go setup

kevinpollet:
### What does this PR do?

This pull request sets the check-latest to true in the Go setup action. This is to ensure that we are always using the latest Go version in the CI. For example, in the Prepare Release build https://github.com/traefik/traefik/actions/runs/13029499637/job/36345477389#step:3:16 we are using a cached version of go1.23.4 but version go1.23.5 is released since mid January.

### Motivation

To ensure we are using the latest Go version in GitHub actions.

### More

- [ ] Added/updated tests


rtribotte (approved):
LGTM

PR #11497, opened 2025-01-30
fix acme write when traefik is shutting down

### LINKED ISSUE #10984: acme.json in an invalid JSON, Traefik can't read certs
### Welcome!

### What did you do?

I use Traefik as a Docker container on several servers, with both the Docker Provider and the File Provider working side by side (the File Provider will soon be deprecated). 

Each Traefik container handles about 100 services (websites in their own Docker containers). I use the HTTP challenge.

My configuration works fine, and generates certificates accordingly. The certs are written to the `acme.json` file automatically by Traefik. This is where the problem lies. 

### What did you see instead?

For unknown reasons and at (apparently) random times, Traefik fails to correctly and coimpletely write the certs to the `acme.json` file, resulting in an incomplete ans so invalid JSON file. 

If I `cat` the `acme.json` file, I can clearly see that the file is invalid, like so :

```jsonc
{
  "acmeResolver": {
    "Account": {
      "Email": "[REDACTED]",
      "Registration": {
        "body": {
          "status": "valid",
          "contact": [
            "mailto:[REDACTED]"
          ]
        },
        "uri": "https://acme-v02.api.letsencrypt.org/acme/acct/[REDACTED]"
      },
      "PrivateKey": "[REDACTED]",
      "KeyType": "4096"
    },
    "Certificates": [
      // For simplicity, only one shown here, but imagine ~100 elements in the array.
      {
        "domain": {
          "main": "[REDACTED]"
        },
        "certificate": "[REDACTED]",
        "key": "someKeyStringThatsRedacted....dISnJVMF
>>EOF

// OBVIOUSLY SOME STUFF IS MISSING HERE
// EOF is indicative, it doesn't appear when using `cat`
```
The cut certificate array element is the $`N`$th one in the array, where $`1 < N < 100`$

I have found no clue to why this occurs. Theories I've come up with up until now are :

- Concurrency issues if Traefik attempts two cert writes simultaneously, or any other concurrency issue
- Traefik crashes during the write process
- Traefik container restarts during the write process

I haven't looked over Traefik's code yet, but if the write process is streamed or chunked, a single sync write to the `acme.json` might fix this. Of course this might cause other issues.

This issue has occurred over the years since I've updated Traefik from v1.7 to v2.5.
This issue has occurred on v2.5, v2.10 & v2.11.

```yml
log:
  level: ERROR
global:
  checkNewVersion: false
  sendAnonymousUsage: false
api:
  insecure: false
  dashboard: true
  debug: false
ping:
  entryPoint: traefik
providers:
  file:
    directory: /etc/traefik/dynamic
    watch: true
  docker:
    endpoint: unix:///var/run/docker.sock
    exposedByDefault: false
certificatesResolvers:
  acmeResolver:
    acme:
      email: [REDACTED]
      storage: /etc/traefik/acme.json
      httpChallenge:
        entryPoint: web
      caServer: https://acme-v02.api.letsencrypt.org/directory
defaultEntryPoints:
  - websecure
  - web
  - traefik
entryPoints:
  web:
    address: :80
    http:
      redirections:
        entryPoint:
          to: websecure
          scheme: https
          permanent: true
  websecure:
    address: :443
    http:
      tls:
        certResolver: acmeResolver
  traefik:
    address: :8080
```

No really obvious log can be found relating to this issue except the "Can't read ACME file" once the issue has occured. 

My Traefik containers are configured to automatically restart on crashes (Docker's `restart: always` policy). Logs often get squashed by repeated crash logs.

juliens:
### What does this PR do?

This PR uses the routine pool to spawn a goroutine for writing the content of acme.json.
This ensures that Traefik waits for the file to be written before shutting down.

### Motivation

Fixes #10984 

### Additional Notes

The PR does not address the issue where each `SaveCertificates` call triggers a file write and blocks other SaveCertificates calls during the write process.
Moreover, if there is some pending `SaveCertificates` during shutdown, we will not wait for each and stop after the first one.

### DIFF
--- a/pkg/provider/acme/local_store.go
+++ b/pkg/provider/acme/local_store.go
@@ -1,6 +1,7 @@
 package acme
 
 import (
+	"context"
 	"encoding/json"
 	"io"
 	"os"
@@ -22,9 +23,9 @@ type LocalStore struct {
 }
 
 // NewLocalStore initializes a new LocalStore with a file name.
-func NewLocalStore(filename string) *LocalStore {
+func NewLocalStore(filename string, routinesPool *safe.Pool) *LocalStore {
 	store := &LocalStore{filename: filename, saveDataChan: make(chan map[string]*StoredData)}
-	store.listenSaveAction()
+	store.listenSaveAction(routinesPool)
 	return store
 }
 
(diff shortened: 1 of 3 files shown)


pkg/provider/acme/local_store.go
```diff
@@ -99,19 +100,32 @@ func (s *LocalStore) get(resolverName string) (*StoredData, error) {
...
+				if err != nil {
+					logger.Error(err)
+				}
+
+				err = os.WriteFile(s.filename, data, 0o600)
+				if err != nil {
+					logger.Error(err)
+				}
 			}
+
```
rtribotte, line 128:
```suggestion
```

kevinpollet (approved):

lbenguigui (approved):
LGTM

PR #11505, opened 2025-02-02
Fix issue where X-Forwared-For headers that contain IP's result in several middlewares not to work properly

david-garcia-garcia:
### What does this PR do?

The DepthStrategy.GetIp() will retrieve ip addresses from the X-Forwarded-For header, without stripping port information. Some reverse Proxy's such as Azure Application Gateway will append the origin port (usually an ephemeral and random port) to the source IP address.

This PR's ensure that any port, if present, is removed. "192.168.4.65:16587" will be read as "192.168.4.65"

### Motivation

Several middlewares will not work when the port is present in the IP address (ratelimit, ipwhitelist).

### DIFF
--- a/pkg/ip/strategy.go
+++ b/pkg/ip/strategy.go
@@ -52,13 +52,25 @@ func (s *DepthStrategy) GetIP(req *http.Request) string {
 		return ""
 	}
 
-	ip := strings.TrimSpace(xffs[len(xffs)-s.Depth])
+	// Extract and clean up the IP
+	rawIP := xffs[len(xffs)-s.Depth]
+	ip := cleanupIP(rawIP)
 
 	if s.IPv6Subnet != nil {
 		return getIPv6SubnetIP(ip, *s.IPv6Subnet)
 	}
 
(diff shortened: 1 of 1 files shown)


kevinpollet:
Hello @david-garcia-garcia and thanks for opening this pull request,

From our understanding, the X-Forwarded-For header should not include a port (see [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)). Instead, the port should be forwarded using the X-Forwarded-Port header.

This is the first time we’ve encountered this issue, so we’d like to discuss it further before proceeding. Could you open an issue describing your infrastructure and the specific use case for including the port in the X-Forwarded-For header? Also, do you know of any other reverse proxies that append the port to this header?

In the meantime, it looks like it’s possible to remove the port information using the approach described in [this Azure blog post](https://azure.microsoft.com/fr-fr/blog/rewrite-http-headers-with-azure-application-gateway/). Would this solution work for your case?

Issue #11509, opened 2025-02-03
Regression - Do not compress when content type text/event-stream


Igloczek:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I've just upgraded from v2.11 to v3.3 and all request with content type`text/event-stream` are broken because of enabled compression.

It wasn't part of any release notes/migration guide, wasn't marked as a breaking change, so I believe it's a regression, since it [was already part of Traefik since 2019](https://github.com/traefik/traefik/pull/5120).

### What did you see instead?

Had to manually exclude it from compression to fix the problem.

### What version of Traefik are you using?

v3.3.3

### What is your environment & configuration?

```yaml
networks:
  coolify:
    external: true
services:
  traefik:
    container_name: coolify-proxy
    image: 'traefik:v3.3'
    restart: unless-stopped
    extra_hosts:
      - 'host.docker.internal:host-gateway'
    networks:
      - coolify
    ports:
      - '80:80'
      - '443:443'
      - '443:443/udp'
      - '8080:8080'
    healthcheck:
      test: 'wget -qO- http://localhost:80/ping || exit 1'
      interval: 4s
      timeout: 2s
      retries: 5
    volumes:
      - '/var/run/docker.sock:/var/run/docker.sock:ro'
      - '/data/coolify/proxy:/traefik'
    command:
      - '--ping=true'
      - '--ping.entrypoint=http'
      - '--api.dashboard=true'
      - '--api.insecure=false'
      - '--entrypoints.http.address=:80'
      - '--entrypoints.https.address=:443'
      - '--entrypoints.http.http.encodequerysemicolons=true'
      - '--entryPoints.https.http3'
      - '--entrypoints.https.http.encodequerysemicolons=true'
      - '--entryPoints.https.http3'
      - '--providers.file.directory=/traefik/dynamic/'
      - '--providers.file.watch=true'
      - '--certificatesresolvers.letsencrypt.acme.httpchallenge=true'
      - '--certificatesresolvers.letsencrypt.acme.httpchallenge.entrypoint=http'
      - '--certificatesresolvers.letsencrypt.acme.storage=/traefik/acme.json'
      - '--api.insecure=false'
      - '--providers.docker=true'
      - '--providers.docker.exposedbydefault=false'
    labels:
[...]

PR #5120, opened 2019-07-18
Do not compress when content type text/event-stream

sylr:
Attempt to fix #2576

Co-authored-by: Joan Rieu <joan@joanrieu.com>

PR #11511, opened 2025-02-04
fix(middleware): do not compress text/event-stream responses

Igloczek:
### What does this PR do?

Fixes a regression where text/event-stream responses were being compressed in v3.x. SSE responses must not be compressed as they rely on real-time data streaming.

This behavior was previously implemented in v2.x, see #5120

### Motivation

As reported in #11509, compression of text/event-stream responses breaks SSE functionality. This was working correctly in v2.11 but broke after upgrading to v3.3, indicating a regression.

SSE responses must not be compressed to ensure proper real-time data streaming functionality.

### More

- [x] Added/updated tests
  - Added test cases for text/event-stream content type in compress middleware tests
  - Updated compress middleware documentation to clarify text/event-stream handling

### Additional Notes

This restores the previous behavior from v2.x where text/event-stream responses were automatically excluded from compression, regardless of user configuration. This ensures SSE works correctly out of the box without requiring manual exclusion configuration.

### DIFF
--- a/docs/content/middlewares/http/compress.md
+++ b/docs/content/middlewares/http/compress.md
@@ -65,7 +65,7 @@ http:
 
 ### `excludedContentTypes`
 
-_Optional, Default=""_ 
+_Optional, Default=""_
 
 `excludedContentTypes` specifies a list of content types to compare the `Content-Type` header of the incoming requests and responses before compressing.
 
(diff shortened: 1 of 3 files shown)


nmengin:
Hey @Igloczek,

We let the PR to the need-triage the time to discuss the use case in the related issue.

kevinpollet:
Hello @Igloczek and thanks for this contribution,

We made some tests and after some investigations, everything is working as expected. There is no issue with compressing `text/event-stream` responses. https://github.com/traefik/traefik/issues/11509 was addressing a different problem where the compress middleware was misbehaving and broke SSE (most likely because of connectivity issues).

Therefore we think we should not exclude `text/event-stream` responses from compression by default. According to this, we are declining this pull request.

PR #11513, opened 2025-02-05
fix(tracing): using ResourceAttributes instead of GlobalAttributes

bruno-de-queiroz:
### What does this PR do?

Fixes the custom attributes setup. The GlobalAttributes is marked as deprecated in favor of ResourceAttributes, but the `backend.setup` was still using the `GlobalAttributes`

### Motivation

Users can no longer assign custom attributes to the span via `ResourceAttributes`, and the' Helm chart' no longer allows `globalAttributes`e.

### More

- [x] Added/updated tests

### Additional Notes

The documentation is already correct, I think its just a piece of code that was forgotten when the `ResourceAttributes` were introduced.


pkg/tracing/tracing.go
```diff
@@ -44,7 +44,11 @@ func NewTracing(conf *static.Tracing) (*Tracer, io.Closer, error) {
 
 	otel.SetTextMapPropagator(autoprop.NewTextMapPropagator())
 
-	tr, closer, err := backend.Setup(conf.ServiceName, conf.SampleRate, conf.GlobalAttributes)
+	if conf.GlobalAttributes != nil && conf.ResourceAttributes == nil {
+		conf.ResourceAttributes = conf.GlobalAttributes
+	}
+
```
rtribotte, line 50:
```suggestion
```

This part is already handled in the SetEffectiveConfiguration method, so it could be removed here.

rtribotte (changes requested):
Hello @bruno-de-queiroz,

Thanks for this contribution!

As this is a fix, it should be open against the v3 branch, could you please rebase it?

pkg/tracing/tracing.go
bruno-de-queiroz: Indeed, thanks for pointing it out :) new PR pointing to the right branch: https://github.com/traefik/traefik/pull/11515

PR #11515, opened 2025-02-05
fix(tracing): using ResourceAttributes instead of GlobalAttributes

bruno-de-queiroz:
### What does this PR do?

Fixes the custom attributes setup. The `GlobalAttributes` is marked as deprecated in favor of ResourceAttributes, but the `backend.setup` was still using the `GlobalAttributes`

### Motivation

Users can't assign custom attributes to the span via `ResourceAttributes`, and the `Helm chart` no longer allows `globalAttributes`.

### More

- [x] Added/updated tests

### Additional Notes

The documentation is already correct, I think it was just a piece of code that was forgotten when the ResourceAttributes were introduced.

### DIFF
--- a/pkg/tracing/tracing.go
+++ b/pkg/tracing/tracing.go
@@ -44,7 +44,7 @@ func NewTracing(conf *static.Tracing) (*Tracer, io.Closer, error) {
 
 	otel.SetTextMapPropagator(autoprop.NewTextMapPropagator())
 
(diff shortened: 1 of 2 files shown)


bruno-de-queiroz:
Also, please review this one: https://github.com/traefik/paerser/pull/63, because currently, it's not possible to simply add `--tracing.resourcesAttributes.service.environment=test` as a flag

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #11518, opened 2025-02-07
Fix double hash in sticky cookie

juliens:
### What does this PR do?

This PR fix a bug where `server.URL` were hash two times in the cookie for sticky session.
Moreover, this PR changes the hashing mechanism for sticky cookie from `fnv` to `sha256`.
This PR keeps the compatibility with older cookie and update existing cookie with the new `sha256` hash.

### Motivation
Avoid multiple hash.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/logs/fields.go
+++ b/pkg/logs/fields.go
@@ -11,7 +11,6 @@ const (
 	ServiceName          = "serviceName"
 	MetricsProviderName  = "metricsProviderName"
 	TracingProviderName  = "tracingProviderName"
-	ServerName           = "serverName"
 	ServerIndex          = "serverIndex"
 	TLSStoreName         = "tlsStoreName"
 	ServersTransportName = "serversTransport"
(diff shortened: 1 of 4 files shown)


rtribotte (approved):
LGTM

sdelicata (approved):
LGTM

kevinpollet (approved):

niksauer:
Is there a way to extract/compute the cookie value for each of the services in a `TraefikService` CRD on Kubernetes? Right now, I need to manually clear the cookie and reload the page until I hit the desired instance. Ideally, I would be able to set the cookie to the correct instance. I've found [this post](https://community.traefik.io/t/sticky-sessions-cookie-value/25058) but I'm not sure which server URL to use without the dashboard.

niksauer:
@juliens @rtribotte @kevinpollet Pinging you as the involved hash parties 🙂

kevinpollet:
Hello @niksauer,

I'm not sure I understand everything correctly, could you open an issue explaining your use case and what you are trying to achieve?

niksauer:
@kevinpollet This might be better explained in https://github.com/traefik/traefik/issues/1207. Essentially, the goal is to have a canary frontend service always route its requests to the canary backend service. This may either be achieved based on setting a header as requested in the linked issue or setting the correct `stickyServer` cookie value in the frontend. The latter would require access to the list of server names used by Traefik [to compute the hash value](https://community.traefik.io/t/sticky-sessions-cookie-value/25058/2). Is there a better way to solve this? Sorry for responding just now. This came up again in our team last week.

PR #11557, opened 2025-02-22
Allow underscore character in hostSNI matcher

rohitlohar45:
### What does this PR do?

As per the discussion, updated the regex for including domains with underscores too by using :word.  

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/muxer/tcp/mux.go
+++ b/pkg/muxer/tcp/mux.go
@@ -310,7 +310,7 @@ func alpn(tree *matchersTree, protos ...string) error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


DeyV:
I think this fix should add both to lines 3 and 2

DeyV:
@rohitlohar45  it's look as ready.  
is there a chance this patch will be released?

rohitlohar45:
@DeyV I have opened the PR for review.

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #11582, opened 2025-03-04
Support HostRegexp as basis for TLSOptions

smerschjohann:
### What does this PR do?

This implements https://github.com/traefik/traefik/issues/11375. It allows different TLSOptions for wildcard domains by parsing HostRegExp() in addition to Host() expressions.

### Motivation

This feature was still missing and my use case requires it, as I want to host some wildcard domains with no mTLS requirement, while the default TLSOptions require it. This was not possible at all before this change. Also see the other mentioned issues other users are facing.

### More

- [x] Added/updated tests

### Additional Notes

-

### DIFF
--- a/pkg/middlewares/snicheck/snicheck.go
+++ b/pkg/middlewares/snicheck/snicheck.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"github.com/traefik/traefik/v3/pkg/middlewares/requestdecorator"
-	"github.com/traefik/traefik/v3/pkg/muxer/tcp"
 	traefiktls "github.com/traefik/traefik/v3/pkg/tls"
 )
 
@@ -23,14 +22,9 @@ type SNICheck struct {
 func New(tlsOptionsForHost map[string]string, tlsOptionsForHostRegexp map[string]string, next http.Handler) *SNICheck {
 	tlsOptionsForHostRegex := make(map[*regexp.Regexp]string)
 	for hostRegexp, tlsOptions := range tlsOptionsForHostRegexp {
-		preparePattern, err := tcp.PreparePattern(hostRegexp)
+		re, err := regexp.Compile(hostRegexp)
 		if err != nil {
-			log.Error().Err(err).Str("hostRegexp", hostRegexp).Msg("Failed to prepare pattern")
-			continue
-		}
-		re, err := regexp.Compile(preparePattern)
-		if err != nil {
-			log.Error().Err(err).Str("host", hostRegexp).Str("pattern", preparePattern).Msg("Failed to compile regex")
+			log.Error().Err(err).Str("hostRegexp", hostRegexp).Msg("Failed to compile regex")
 			continue
 		}
 		tlsOptionsForHostRegex[re] = tlsOptions
--- a/pkg/middlewares/snicheck/snicheck.go
+++ b/pkg/middlewares/snicheck/snicheck.go
@@ -1,7 +1,6 @@
 package snicheck
 
 import (
-	"github.com/traefik/traefik/v3/pkg/muxer/tcp"
 	"net"
 	"net/http"
 	"regexp"
@@ -9,6 +8,7 @@ import (
 
 	"github.com/rs/zerolog/log"
 	"github.com/traefik/traefik/v3/pkg/middlewares/requestdecorator"
+	"github.com/traefik/traefik/v3/pkg/muxer/tcp"
 	traefiktls "github.com/traefik/traefik/v3/pkg/tls"
 )
 
--- a/pkg/middlewares/snicheck/snicheck.go
+++ b/pkg/middlewares/snicheck/snicheck.go
@@ -1,8 +1,10 @@
 package snicheck
 
 import (
+	"github.com/traefik/traefik/v3/pkg/muxer/tcp"
 	"net"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/rs/zerolog/log"
@@ -12,13 +14,28 @@ import (
 
 // SNICheck is an HTTP handler that checks whether the TLS configuration for the server name is the same as for the host header.
 type SNICheck struct {
-	next              http.Handler
-	tlsOptionsForHost map[string]string
+	next                   http.Handler
+	tlsOptionsForHost      map[string]string
+	tlsOptionsForHostRegex map[*regexp.Regexp]string
 }
 
 // New creates a new SNICheck.
-func New(tlsOptionsForHost map[string]string, next http.Handler) *SNICheck {
-	return &SNICheck{next: next, tlsOptionsForHost: tlsOptionsForHost}
+func New(tlsOptionsForHost map[string]string, tlsOptionsForHostRegexp map[string]string, next http.Handler) *SNICheck {
+	tlsOptionsForHostRegex := make(map[*regexp.Regexp]string)
+	for hostRegexp, tlsOptions := range tlsOptionsForHostRegexp {
+		preparePattern, err := tcp.PreparePattern(hostRegexp)
+		if err != nil {
+			log.Error().Err(err).Str("hostRegexp", hostRegexp).Msg("Failed to prepare pattern")
+			continue
+		}
+		re, err := regexp.Compile(preparePattern)
+		if err != nil {
+			log.Error().Err(err).Str("host", hostRegexp).Str("pattern", preparePattern).Msg("Failed to compile regex")
+			continue
+		}
+		tlsOptionsForHostRegex[re] = tlsOptions
+	}
+	return &SNICheck{next: next, tlsOptionsForHost: tlsOptionsForHost, tlsOptionsForHostRegex: tlsOptionsForHostRegex}
 }
 
 func (s SNICheck) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
@@ -32,8 +49,8 @@ func (s SNICheck) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 
 	// Domain Fronting
 	if !strings.EqualFold(host, serverName) {
-		tlsOptionHeader := findTLSOptionName(s.tlsOptionsForHost, host, true)
-		tlsOptionSNI := findTLSOptionName(s.tlsOptionsForHost, serverName, false)
+		tlsOptionHeader := s.findTLSOptionName(host, true)
+		tlsOptionSNI := s.findTLSOptionName(serverName, false)
 
 		if tlsOptionHeader != tlsOptionSNI {
 			log.Debug().
@@ -68,13 +85,13 @@ func getHost(req *http.Request) string {
 	return strings.TrimSpace(host)
 }
 
-func findTLSOptionName(tlsOptionsForHost map[string]string, host string, fqdn bool) string {
-	name := findTLSOptName(tlsOptionsForHost, host, fqdn)
+func (s SNICheck) findTLSOptionName(host string, fqdn bool) string {
+	name := s.findTLSOptName(host, fqdn)
 	if name != "" {
 		return name
 	}
 
-	name = findTLSOptName(tlsOptionsForHost, strings.ToLower(host), fqdn)
+	name = s.findTLSOptName(strings.ToLower(host), fqdn)
 	if name != "" {
 		return name
 	}
@@ -82,9 +99,15 @@ func findTLSOptionName(tlsOptionsForHost map[string]string, host string, fqdn bo
 	return traefiktls.DefaultTLSConfigName
 }
 
-func findTLSOptName(tlsOptionsForHost map[string]string, host string, fqdn bool) string {
-	if tlsOptions, ok := tlsOptionsForHost[host]; ok {
+func (s SNICheck) findTLSOptName(host string, fqdn bool) string {
+	if tlsOptions, ok := s.tlsOptionsForHost[host]; ok {
 		return tlsOptions
+	}
+
+	for regexp, tlsOptions := range s.tlsOptionsForHostRegex {
+		if regexp.MatchString(host) {
+			return tlsOptions
+		}
 	}
 
 	if !fqdn {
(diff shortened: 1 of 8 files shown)


nmengin:
Hey @smerschjohann,

We have moved it to the status `design-review` to allow us the check if the implementation doe not have side-effect.

smerschjohann:
@nmengin sure, if you have anything to improve, I'm happy to change or add it.

sdelicata:
Hello @smerschjohann, and thanks for your contribution!
The ball is now in our court, and the PR is waiting for a design review.
More news soon.

pkg/middlewares/snicheck/snicheck.go
```diff
@@ -3,22 +3,39 @@ package snicheck
...
+	tlsOptionsForHostRegex := make(map[*regexp.Regexp]string)
+	for hostRegexp, tlsOptions := range tlsOptionsForHostRegexp {
+		preparePattern, err := tcp.PreparePattern(hostRegexp)
+		if err != nil {
+			log.Error().Err(err).Str("hostRegexp", hostRegexp).Msg("Failed to prepare pattern")
+			continue
+		}
+		re, err := regexp.Compile(preparePattern)
+		if err != nil {
+			log.Error().Err(err).Str("host", hostRegexp).Str("pattern", preparePattern).Msg("Failed to compile regex")
```
sdelicata:
```suggestion
			log.Error().Err(err).Str("hostRegexp", hostRegexp).Str("pattern", preparePattern).Msg("Failed to compile regex")
```

pkg/middlewares/snicheck/snicheck_test.go
```diff
@@ -10,10 +10,11 @@ import (
 
 func TestSNICheck_ServeHTTP(t *testing.T) {
 	testCases := []struct {
-		desc              string
-		tlsOptionsForHost map[string]string
-		host              string
-		expected          int
+		desc                    string
+		tlsOptionsForHost       map[string]string
+		tlsOptionsFotHostRegexp map[string]string
```
sdelicata, line 15: typo: `tlsOptionsFotHostRegexp` => `tlsOptionsForHostRegexp`

pkg/middlewares/snicheck/snicheck_test.go
```diff
@@ -24,16 +25,36 @@ func TestSNICheck_ServeHTTP(t *testing.T) {
...
 				"example.com": "foo",
 			},
+			tlsOptionsFotHostRegexp: map[string]string{},
+			expected:                http.StatusOK,
+		},
+		{
+			desc:              "with wildcard TLS options and subdomain",
+			tlsOptionsForHost: map[string]string{},
+			tlsOptionsFotHostRegexp: map[string]string{
+				".*.localhost": "foo",
```
sdelicata, line 35:
```suggestion
				".*\.localhost": "foo",
```

pkg/server/router/tcp/manager_test.go
```diff
@@ -602,6 +602,33 @@ func TestDomainFronting(t *testing.T) {
 			ServerName:     "host2.local",
 			expectedStatus: http.StatusOK,
 		},
+		{
+			desc: "Request is OK when Host is matched by HostRegexp and TLS options are different",
+			routers: map[string]*runtime.RouterInfo{
+				"router-1@file": {
+					Router: &dynamic.Router{
+						EntryPoints: entryPoints,
+						Rule:        "HostRegexp(`.*.local`)",
```
sdelicata, line 611:
```suggestion
						Rule:        "HostRegexp(`.*\.local`)",
```

pkg/middlewares/snicheck/snicheck.go
```diff
@@ -3,22 +3,39 @@ package snicheck
...
+	next                   http.Handler
+	tlsOptionsForHost      map[string]string
+	tlsOptionsForHostRegex map[*regexp.Regexp]string
 }
 
 // New creates a new SNICheck.
-func New(tlsOptionsForHost map[string]string, next http.Handler) *SNICheck {
-	return &SNICheck{next: next, tlsOptionsForHost: tlsOptionsForHost}
+func New(tlsOptionsForHost map[string]string, tlsOptionsForHostRegexp map[string]string, next http.Handler) *SNICheck {
+	tlsOptionsForHostRegex := make(map[*regexp.Regexp]string)
```
sdelicata, line 23: Naming is confusing. Could you improve it, please?

pkg/server/router/tcp/router.go
```diff
@@ -240,6 +248,26 @@ func (r *Router) AddHTTPTLSConfig(sniHost string, config *tls.Config) {
...
+func (r *Router) AddHTTPRegexpTLSConfig(hostRegexp string, config *tls.Config) {
+	preparePattern, err := tcpmuxer.PreparePattern(hostRegexp)
+	if err != nil {
+		log.Error().Err(err).Str("hostRegexp", hostRegexp).Msg("Failed to prepare pattern")
+		return
+	}
+
+	re, err := regexp.Compile(preparePattern)
+	if err != nil {
+		log.Error().Err(err).Str("host", hostRegexp).Str("pattern", preparePattern).Msg("Failed to compile regex")
```
sdelicata:
```suggestion
		log.Error().Err(err).Str("hostRegexp", hostRegexp).Str("pattern", preparePattern).Msg("Failed to compile regex")
```

pkg/server/router/tcp/router.go
```diff
@@ -297,6 +325,23 @@ func (r *Router) SetHTTPSForwarder(handler tcp.Handler) {
...
+		} else {
+			tcpHandler = &tcp.TLSHandler{
+				Next:   handler,
+				Config: tlsConf,
+			}
+		}
+
+		rule := "HostSNIRegexp(`" + sniHostRegex.String() + "`)"
+		if err := r.muxerHTTPS.AddRoute(rule, "", tcpmuxer.GetRulePriority(rule), tcpHandler); err != nil {
+			log.Error().Err(err).Msg("Error while adding route for host")
```
sdelicata, line 335:
```suggestion
			log.Error().Err(err).Msg("Error while adding route for host regexp")
```

lbenguigui (changes requested):

pkg/server/router/tcp/router.go
```diff
@@ -240,6 +248,26 @@ func (r *Router) AddHTTPTLSConfig(sniHost string, config *tls.Config) {
 	r.hostHTTPTLSConfig[sniHost] = config
 }
 
+func (r *Router) AddHTTPRegexpTLSConfig(hostRegexp string, config *tls.Config) {
+	preparePattern, err := tcpmuxer.PreparePattern(hostRegexp)
```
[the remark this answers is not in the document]
smerschjohann: I expect a pure regex Pattern now

pkg/middlewares/snicheck/snicheck.go
```diff
@@ -3,22 +3,39 @@ package snicheck
...
+	for hostRegexp, tlsOptions := range tlsOptionsForHostRegexp {
+		preparePattern, err := tcp.PreparePattern(hostRegexp)
```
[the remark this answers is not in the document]
smerschjohann: I expect a pure regex Pattern now

pkg/server/router/tcp/manager.go
```diff
@@ -208,9 +213,28 @@ func (m *Manager) buildEntryPointHandler(ctx context.Context, configs map[string
...
+				tlsOptionsForHostSNI[regex] = make(map[string]nameAndConfig)
+			}
+			tlsOptionsForHostSNI[regex][tlsOptionsName] = nameAndConfig{
+				routerName: routerHTTPName,
+				regex:      true,
+				TLSConfig:  tlsConf,
+			}
+
+			// if multiple regex matchers are defined for the same tlsOptions, we fallback to default,
+			// this could still clash if multiple regex matchers would match the same request, but that's a user error and falling back to default is not better.
```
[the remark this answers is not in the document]
smerschjohann: Maybe one possible solution would be sorting the patterns alphabetically. In that way, the same pattern would be chosen each time.

smerschjohann:
any updates on this one? I removed the v2 syntax accordingly

smerschjohann:
any updates on this one? I removed the v2 syntax accordingly

nmengin:
Hey @smerschjohann,

Thanks for reaching out.

Your PR is still on our radar.

We have put it in our backlog, and will keep you updated once we have moved forward on it.

smerschjohann:
@nmengin thank you for the feedback, I understand and will keep it on my watch list

pkg/muxer/http/mux.go
```diff
@@ -116,6 +116,36 @@ func (m *Muxer) AddRoute(rule string, syntax string, priority int, handler http.
...
+	for matcher := range httpFuncs {
+		matchers = append(matchers, matcher)
+	}
+	for matcher := range httpFuncsV2 {
+		matchers = append(matchers, matcher)
+	}
+
+	parser, err := rules.NewParser(matchers)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error while creating parser: %w", err)
```
sdelicata, line 132:
```suggestion
		return nil, nil, fmt.Errorf("creating parser: %w", err)
```

pkg/muxer/http/mux.go
```diff
@@ -116,6 +116,36 @@ func (m *Muxer) AddRoute(rule string, syntax string, priority int, handler http.
...
+	}
+
+	parse, err := parser.Parse(rule)
+	if err != nil {
+		return nil, nil, fmt.Errorf("error while parsing rule %s: %w", rule, err)
```
sdelicata, line 137:
```suggestion
		return nil, nil, fmt.Errorf("parsing rule %s: %w", rule, err)
```

pkg/server/router/tcp/manager.go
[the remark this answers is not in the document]
sdelicata: I agree with @lbenguigui, the fact that the selection is non-deterministic in case of a conflict is a serious issue for users, as it can lead to unpredictable and hard-to-debug behavior.

Using alphabetical order as a workaround seems arbitrary to me, but it's still an improvement over random selection.

That said, it's important to log a clear warning at least when such a conflict occurs. This would help users identify and fix ambiguous configurations more easily.

pkg/server/router/tcp/manager.go
```diff
@@ -96,6 +96,7 @@ func (m *Manager) BuildHandlers(rootCtx context.Context, entryPoints []string) m
 type nameAndConfig struct {
 	routerName string // just so we have it as additional information when logging
 	TLSConfig  *tls.Config
+	regex      bool
```
sdelicata, line 99: WDYT about renaming it to `isRegex`?

pkg/muxer/http/mux.go
```diff
@@ -116,6 +116,36 @@ func (m *Muxer) AddRoute(rule string, syntax string, priority int, handler http.
 	return nil
 }
 
+// ParseDomainsAndRegex extract domains and regex from rule.
+// It returns the domains and regex. If the rule is invalid, it returns an error.
+func ParseDomainsAndRegex(rule string) ([]string, []string, error) {
```
sdelicata, line 121: WDYT about renaming it to `ParseHostMatchers`?

pkg/server/router/tcp/manager.go
```diff
@@ -140,15 +144,15 @@ func (m *Manager) buildEntryPointHandler(ctx context.Context, configs map[string
 			tlsOptionsName = provider.GetQualifiedName(ctxRouter, routerHTTPConfig.TLS.Options)
 		}
 
-		domains, err := httpmuxer.ParseDomains(routerHTTPConfig.Rule)
+		domains, domainRegex, err := httpmuxer.ParseDomainsAndRegex(routerHTTPConfig.Rule)
```
sdelicata, line 147: Renaming `domainRegex` to `domainRegexps` makes it a bit clearer, WDYT?

sdelicata (changes requested):
Hi @smerschjohann, and thanks for your contribution.
I made a few comments and suggestions.

smerschjohann:
@sdelicata for some reason I cannot reply directly to the comment above about the regex issue. We cannot log a clear warning when we apply the filters, this could only be done between static host matchers and the regexes. 
We could only log if multiple regexes match while routing, which would impact performance since we have to iterate over all possible matches to print this warning, instead of just finding the first match - and this on each request. Maybe another option instead of just alphabetical sort would be "longest regex wins" which is not necessarily the intent in that situation, but maybe matches with the "longest prefix matching" in other situations. More sophisticated approaches like "shortest capture groups" etc. doesn't make it clearer for the user. I'm not sure if this edge case can really be prevented or even logged without performance degration. In addition, the same issue can already happen in the HostSNIRegexp TCP routes where no such warnings are printed as far as I know.

All other comments look fine for me. If we clarify what to do in this situation, I would try to do everything in one go.

rtribotte:
Hi @smerschjohann,

Thanks for this contribution!

Following @lbenguigui @sdelicata comments, I think the SNICheck mechanism is simply not implementable with maps with regex, as it was done previously, because the need is to select the same TLSOption as the TCP muxer will do.
I think the priority order is not the real concern, the SNICheck mechanism must now be integrated with the muxer, or use the muxer, to be reliable.
This is much more work, but I think this is mandatory.

nmengin:
Hey @smerschjohann,

Have you seen the latest comment from @rtribotte?

Have you planned to work on it?

smerschjohann:
@nmengin hi, I missed that one, thanks for the reminder.

I'm still not sure if we have overall the same understanding. The main issue is that a regular expression is not a 1:1 match and multiple regular expression can match the same domain. There is the chance that a user creates multiple entries that will all match the same domain. This cannot be resolved, as we cannot check on this beforehand. We also don't want to check on all Regex pattern all the time to see if multiple regular expressions will match, as this would lead to bad performance with the only gain that we have a log entry. This might be a debug configuration which you can activate to detect this collisions, but it shouldn't be on by default (in my opinion). As long it is not clear how we want to handle this, it is not possible to proceed here.

I'm not sure if I understand the muxer comment. I modified the findTLSOptionName accordingly so that we find the correct entry based on regex. Therefore we have two maps now. One for exact matches (faster matches) and one for regex matches. This is already done in this branch?
Hm reading your comment again, do you mean the case where we have a collision of regular expressions and then the SNICheck finds a different entry than the tcp muxer? As this should not be configured in the first place and can be prevented by "being determenistic in case of collision" we don't need to change too much here. So going back to problem 1: how do we want to handle regex collisions? (btw. I think we should have the same problem for TCP TLS muxes without the whole HTTPS part)

rtribotte:
@smerschjohann I agree that overlap detection cannot be implemented for HostRegex matchers in the same way as for Host matchers. And that runtime checking of all regular expression patterns is not doable

However, I think the muxer already handles these collisions effectively through rule priority resolution.
When multiple HostRegex rules could potentially match the same request, the muxer doesn't need to detect overlaps beforehand — it simply applies the first matching rule in priority order.

Regarding the increased risk of potential misconfiguration that would be introduced by this feature, I'm not completely convinced that it would make the situation worse.
Currently, if two routers with the same host define different TLSOptions, falling back to the default one can be confusing and require the user's attention.

I am more concerned about introducing this feature without a configuration flag and the fact that it would break things.

I'm not opposed to not implementing this feature, as it could be error-prone — but this needs more discussion.

smerschjohann:
@rtribotte I see, yes it should be possible to check for priority orders. 

As for breaking something: I don‘t see where that should happen. Either you had already a lot of warnings in your logs that the general TLSOptions will be used (which was not intended by the user) or you won‘t have an issue at all. Or am I missing a config variant here?

But I‘m also fine with introducing a experimental flag or something that I guess should become the default later on.

celebilercem:
@smerschjohann @rtribotte I haven't seen any update this year, I was wondering if this is still being worked on.

rtribotte:
Hi @smerschjohann @celebilercem,

This feature is still on our radar, we are for now focused elsewhere.

It is worth mentioning that new v3.7 introduces support for wildcards in the `Host` and `HostSNI` matchers, which allow and satisfy the attachment of TLSOptions: https://doc.traefik.io/traefik/v3.7/migrate/v3/#wildcard-host-and-hostsni.

PR #11587, opened 2025-03-06
docs: add k8s gateway integration

t3hmrman:
### What does this PR do?

This commit adds instructions related to configuration of Traefik to work with Knative via the upstream `net-gateway-api` project, which is intended to support k8s Gateways in a generic fashion.

### Motivation

As the Knative ecosystem has finally made progress towards supporting other routing providers (albeit through an additional controller that enables the Gateway API), Traefik can be used with Knative now -- and more people should know about it.

I personally have not adopted Knative because of it's restrictions on network/routing providers, and there are likely others who have been waiting for the ability to use Traefik's advanced functionality and flexibility with Knative.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/user-guides/knative-gateway.md
+++ b/docs/content/user-guides/knative-gateway.md
@@ -1,0 +1,149 @@
+---
+title: "Traefik Proxy Knative Gateway API Integration"
+description: "This section of the Traefik Proxy documentation explains how to use Traefik as reverse proxy with Knative, via the Gateway API integration provided by Knative."
+---
+# Traefik Knative Gateway API Integration
+
+Make use of [Knative][kn], a popular serverless application platform built for Kubernetes.
+{: .subtitle }
+
+[kn]: https://knative.dev/docs/
+
+## How Knative integrates with Traefik via Gateway API
+
+Traefik integrates with [Knative Serving][kn-serving] by providing a powerful and flexible routing layer,
+via the [Kubernetes Gateway API][k8s-gateway], with the use of [`net-gateway-api`][kn-net-gateway-api].
+
+Traefik supports recent versions of [Gateway API][traefik-gateway-api] along with extended features like `TCPRoute` and `TLSRoute`.
+
+!!! info
+    To get started quickly with Gateway API support in Traefik, see ["Getting started with Kubernetes Gateway API and Traefik][community-getting-started-traefik-gateway-api].
+
+[k8s-gateway]: https://gateway-api.sigs.k8s.io/
+[traefik-gateway-api]: https://doc.traefik.io/traefik/providers/kubernetes-gateway/
+[kn-serving]: https://knative.dev/docs/serving/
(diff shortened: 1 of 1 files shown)


nmengin:
Hey @t3hmrman,

From what I understood, Traefik requires the support of the service filters in Gateway API to work well with KNative.
Can you explain how you've passed through this limitation?

Thanks in advance for your feedback.

t3hmrman:
Hey @nmengin my apologies -- I was testing locally and thought this setup was working under the recent work in the other PR (separately from setting up a `provider`), but I think my local config here was wrong.

I can't get it working locally again with the new code & use of the `net-gateway-api`.

t3hmrman:
Sorry -- this *was* working, *without* the new code. I just looked back through my notes and I know what I did to get this working (I started w/ the `k3s` setup inside the repo and later switching to `kind` locally which is why I got confused -- the MetalLB setup was the one that worked, not Traefik-as-a-DaemonSet).

So, if you have Traefik set up with a MetalLB that *properly provides a IP for the relevant Gateway* (i.e. the automatically created one for Traefik), `net-gateway-api` will use *that* IP.

The relevant upstream code:

https://github.com/knative/serving/blob/636392e930c1d6b5b5619bde54bf38e3990acf88/pkg/reconciler/route/resources/service.go#L115

So what you need is a `LoadBalancer` service that looks like this:

```yaml
---
apiVersion: v1
kind: Service
metadata:
  name: example
spec:
  type: LoadBalancer
  selector:
    "serving.knative.dev/service": example
  ports:
    - port: 80
      protocol: TCP
```

And a HTTP route that looks like this:

```yaml
---
kind: HTTPRoute
metadata:
  name: example.localhost
spec:
  parentRefs:
    - name: traefik
      namespace: ingress
  hostnames:
    - example.localhost
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /
      backendRefs:
        - name: example
          namespace: ingress
          port: 80
```

I'll make a copy of my notes & local instructions so that it's easier to try.

t3hmrman:
Finally got the repo together -- please feel free to try it locally:

https://github.com/t3hmrman/traefik-kn-with-lb

t3hmrman:
Hey @nmengin just a ping here -- did you get a chance to check out the repo? I'm wondering if what I was able to do is an accepted use of Traefik and if ya'll would want it in the docs! 

If not that's totally fine too, I think I'll just write a small post about the config and however Knative support turns out I'll weigh which approach to take.

mmatur:
Hi @t3hmrman,

Thanks for your contribution. After discussion with other maintainers we think it could be better to push your repo to the following Github [organisation](https://github.com/traefik-contrib). A specific [repository](https://github.com/traefik-contrib/knative-gateway-api) has been created for you

You have been invited to this organisation.

I'm closing this issue accordingly

t3hmrman:
Hey @mmatur thanks for making the repo! Will try to push the code up there to contrib before weeks end.

As always thanks to you and the team for all the work on Traefik!

PR #11589, opened 2025-03-06
Ability to enable unsafe in yaegi through plugin manifest

Rydez:
### What does this PR do?

This PR addresses https://github.com/traefik/traefik/issues/11588. This PR provides a way to enable the use of the `unsafe` package in the go standard library within yaegi by adding a flag in the plugin manifest.

### Motivation

Otherwise, I need to create a wasm plugin which means I can't use go's native request and response types, and the request context is not preserved. 

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/plugins/builder.go
+++ b/pkg/plugins/builder.go
@@ -139,7 +139,7 @@ func newMiddlewareBuilder(ctx context.Context, goPath string, manifest *Manifest
 		return newWasmMiddlewareBuilder(goPath, moduleName, wasmPath, settings)
 
 	case runtimeYaegi, "":
-		i, err := newInterpreter(ctx, goPath, manifest.Import)
+		i, err := newInterpreter(ctx, goPath, manifest)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create Yaegi interpreter: %w", err)
 		}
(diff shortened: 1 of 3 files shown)


rtribotte:
Hi @Rydez,

Thanks for this contribution!

We have looked at the changes and we have some feedback.
Given the nature of the `unsafe` package, we are concerned about the awareness of Traefik users and we want to make sure that we give them full control over whether to allow the use of the `unsafe` package.
Also, since the plugin mechanism relies on several parts, changes are also needed in the [Piceus repository](https://github.com/traefik/piceus) and on the [catalog frontend repository](https://github.com/traefik/plugin-catalog-ui).

Considering that, we think the plan would be to:
- Add a flag to the plugin manifest to use of the `unsafe` package (as you have started to implement). This flag controls the Yaegi configuration on plugin startup in Traefik.
- The manifest flag would also be a marker for Piceus, when referencing the plugin, to mark it as "unsafe" and also bypass the usual checks (running the `New` and `CreateConfig` func). You can do this by creating a PR on the [Piceus repository](https://github.com/traefik/piceus).
- A plugin marked as "unsafe"  must be identifiable in the catalogue ui. It would also be great to provide some insight on the meaning of unsafe for the user browsing the catalog. You can do this by creating a PR on [the catalog frontend repository](https://github.com/traefik/plugin-catalog-ui).
- Add an option to the [plugin settings configuration](https://github.com/traefik/traefik/blob/ae4a00b4bcc2c9131de61265a9739e2003d36746/pkg/plugins/types.go#L13) to allow the use of the `unsafe` package. It's the opt-in part for the Traefik user. If this option is missing, the plugin should fail to start.

juliens:
Hello @Rydez 

I made the changes for the settings part, and I opened the PR on piceus and on the plugin-service.

lbenguigui (approved):
LGTM

sdelicata (approved):
Hi @Rydez, and thanks for your contribution!
LGTM

juliens:
I've merged this PR, so unsafe now works with local plugins.

To support plugins using unsafe in the catalog, we’ll need to wait for [piceus#110](https://github.com/traefik/piceus/pull/110) and [plugin-service#69](https://github.com/traefik/plugin-service/pull/69) to be merged.

PR #11592, opened 2025-03-09
Update service-by-label.md

kmanwar89:
### What does this PR do?

Documentation update that explains how default service names are derived if not already defined in a Docker label.

### Motivation

I didn't see where in the documentation it was called out exactly how a default service name is derived. When trying out different Docker label combinations to find the most efficient setup for adding to existing compose files, I was confused where and how these service names were auto-generated.  Once I discovered the answer, I thought it would be helpful to share back to the community.

### More

- [ ] Added/updated tests
- [ X ] Added/updated documentation

### Additional Notes

N/A


nmengin:
Hello @kmanwar89,

Unfortunately, we cannot accept it.

Indeed, as you have maybe seen, we are revamping the documentation and we prefer to avoid bringing modifications yet.
However, we will take your proposition into account during this revamp.

Thanks 👍

PR #11539, opened 2025-02-18
Update to go 1.24

mloiseleur:
### What does this PR do?

Update to go 1.24
Based on previous similar PR (#11035).

### Motivation

https://go.dev/doc/go1.24
Fixes #11532

### Additional Notes

1. workflows are updated to use the same source: `go.mod`
2. golangci-lint is updated to latest version: v1.64.5

PR #11601, opened 2025-03-11
enable mptcp

muhammadn:
### What does this PR do?

Add support for Multipath TCP which was already supported by Go 1.21

### Motivation

Adding MPTCP support allows clients to connect to traefik even when they lose the primary connection (think mobile phone with both WiFi and 5G/4G), or any multihomed users

Also MPTCP on server side also helps to increase throughput across multiple interfaces with different Network interface IPs

NOTE: If here are no MPTCP connection available, it will fallback to regular TCP.
### More

- [x] Added/updated tests

### DIFF
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -478,6 +478,7 @@ func buildListener(ctx context.Context, name string, config *static.EntryPoint) 
 
 	if listener == nil {
 		listenConfig := newListenConfig(config)
+                listenConfig.SetMultipathTCP(true)
 		listener, err = listenConfig.Listen(ctx, "tcp", config.GetAddress())
 		if err != nil {
 			return nil, fmt.Errorf("error opening listener: %w", err)


matttbe:
Hello,

Thank you for working on this! It would be great to have MPTCP support in Traefik!

Note that this modification should not be needed when #11539 will be merged: with Go 1.24, MPTCP support is enabled by default on the server side when available, see [here](https://go-review.googlesource.com/c/go/+/607715). (This will not affect the "plain TCP" connections, MPTCP will only be used if requested by the clients, see: [here](https://www.mptcp.dev/faq.html#why--when-should-mptcp-be-enabled-by-default))

muhammadn:
@matttbe Thanks for referencing me to those links! The first and second links makes me question if a backport of MPTCP support for previous Go version is necessary? If it doesn't make sense, then i would probably look for implementing support for other webservers.

For the third link, it gives me some hints/idea on implementing something similar to the MPTCP support in Chrome that i'm looking at. Thank you! 🙃

matttbe:
> @matttbe Thanks for referencing me to those links! The first and second links makes me question if a backport of MPTCP support for previous Go version is necessary?

Sorry, I don't know the answer, because I don't know Traefik well. (I forgot to mention for the others that I'm working on the MPTCP implementation in the Linux kernel, and I looked at this PR because @muhammadn mentioned it in a message he sent me.)

Personally, I would recommend you to test #11539 first, and check if it has the MPTCP support enabled by default as expected. If it doesn't, it might be interesting to understand why. If it does, then probably best to check with the Traefik maintainers what you can do?

kevinpollet:
Hi @muhammadn and thanks for opening this pull request, 

> The first and second links make me wonder if a backport of MPTCP support for previous Go versions is necessary?

As @matttbe pointed out, MPTCP support is enabled by default in go1.24 as mentioned in the release note, see https://tip.golang.org/doc/go1.24#minor_library_changes in the net section.

As soon as https://github.com/traefik/traefik/issues/11532 is addressed, this feature will be available in Traefik (v2.11.x and v3.x) with this bump.

I do not think it is necessary to backport this, as the go1.24 bump will be done in a near future.

PR #11603, opened 2025-03-12
Clarifies that retry middleware uses TCP, not HTTP status codes

geraldcroes:
### What does this PR do?

Clarifies that the Retry middleware only retries on TCP connection failures and does not consider HTTP status codes.

### Motivation

Improve clarity of the documentation


rtribotte (approved):
LGTM

kevinpollet (approved):

PR #11609, opened 2025-03-13
Add support to disable session ticket

avdhoot:
### What does this PR do?

Add support to disable session ticket creation 

### Motivation

Traefik utilizes TLS session tickets for session resumption, helping to reduce handshake latency. However, in multi-instance deployments, these session tickets require synchronized encryption keys across all instances. Without proper key synchronization, session tickets become invalid, forcing clients to complete a full TLS handshake. In this scenario, the client consumes bandwidth by sending an outdated ticket, which the server rejects before issuing a new one—resulting in additional bandwidth usage.

For IoT devices, where **bandwidth efficiency is critical**, unnecessary session ticket creation and transmission can sometimes **consume more bandwidth than a new handshake**.  

#### **Changes Introduced**  
- Added a configuration option to **disable TLS session tickets** (`SessionTicketsDisabled`).  Which ensures that **each TLS handshake starts fresh** without relying on potentially invalid session tickets. 

### More

- [ ] Added/updated tests


nmengin:
Hello @avdhoot,

We've set the status to "design-review" to allow us to check the PR and ensure there is no impact on the rest of Traefik before moving forward.

kevinpollet (changes requested):
Hello @avdhoot and thanks for this contribution,

At first glance, could you rename the `SessionTicketDisabled` option to `DisableSessionTickets`?

pkg/provider/kubernetes/crd/traefikio/v1alpha1/tlsoption.go
```diff
@@ -44,7 +44,8 @@ type TLSOptionSpec struct {
 	// ALPNProtocols defines the list of supported application level protocols for the TLS handshake, in order of preference.
 	// More info: https://doc.traefik.io/traefik/v3.3/https/tls/#alpn-protocols
 	ALPNProtocols []string `json:"alpnProtocols,omitempty"`
-
+	// DisableSessionTickets when set to true, disables TLS session resumption via session tickets.
+	DisableSessionTickets bool `json:"sessionTicketsDisabled,omitempty"`
```
kevinpollet, line 48: I think you forgot to change the attribute name in the struct tag. Once the tag is renamed, remember to regenerate the CRDs with the `make generate-crd` command.

```suggestion
	DisableSessionTickets bool `json:"disableSessionTickets,omitempty"`
```

kevinpollet (changes requested):

rtribotte:
Hi @avdhoot,

Thank you for the addressing the last review comment.
I'm noticing that the option still misses to be documented, and also a unit test, can you please handle it?

We can also address it for you if you want.

pkg/tls/tlsmanager.go
```diff
@@ -408,6 +408,9 @@ func buildTLSConfig(tlsOption Options) (*tls.Config, error) {
 		}
 	}
 
+	// Set SessionTicketsDisabled if set in the config
+	conf.SessionTicketsDisabled = tlsOption.DisableSessionTickets
```
kevinpollet, line 412: Could you move this initialisation to line https://github.com/traefik/traefik/pull/11609/files#diff-3f7f33621af8b7e588a5375221f22cf500b5d598ceb3f5fbb8be623c94defd41R328?

kevinpollet (changes requested):

avdhoot:
> Hi @avdhoot,
> 
> Thank you for the addressing the last review comment. I'm noticing that the option still misses to be documented, and also a unit test, can you please handle it?
> 
> We can also address it for you if you want.

I have added documentation and unit test cases to the best of my understanding. If you notice anything missing, please feel free to add it directly or let me know.

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #11619, opened 2025-03-19
Deprecate defaultRuleSyntax and ruleSyntax options

rtribotte:
### What does this PR do?

This PR deprecates the `defaultRuleSyntax` install configuration option and the `ruleSyntax` routing configuration option.

It also changes the value, used for internal routers and routers not directly defined with routing configuration, from `v3` to `default`, in order to distinguish those cases from user defined routers when checking for the deprecation log to be print.
The `default` value does not changes the behavior, as only the `v2` value matters to decide wether to use the v2 rule parser:
https://github.com/traefik/traefik/blob/50b0d772e5c0edad8bdbdb8f4013121dbad9d10c/pkg/muxer/http/mux.go#L80-L95

### Motivation

To document deprecated options.

### More

- [x] Added/updated tests

### Additional Notes


nmengin (approved):
LGTM 👍

sdelicata (approved):
LGTM

lbenguigui (approved):
LGTM

PR #11659, opened 2025-03-31
Set old certificate's tls_certs_not_after metric to 0.

### LINKED ISSUE #8606: tls_certs_not_after metric shows old renewed serials
### Welcome!

### What did you do?

The TLS Certificate got renewed automatically through acme.
The Metric for the old one did not get cleaned up and still showed up in the list.
Only after a container restart the metric for the old certificate was finally gone.

As there seems to be also no documentation for this newly introduced metric in the documentation, I can't tell if
this behavior is intentional.

To my understanding, a Certificate which is no longer present, as it was renewed, should also have it's metric cleaned up.

### What did you see instead?

multiple metrics for the same `cn`/`sans` but different `serial`.

This triggered the following alerting rule

```
name: TraefikTlsCertExpiration
expr: (traefik_tls_certs_not_after - time()) < 432000
for: 1h
labels:
severity: warning
annotations:
description: TLS Certificate for {{ $labels.sans }} on instance {{ $labels.instance }} will expire within the next 5 Days
summary: TLS Certificate will expire soon (instance {{ $labels.instance }})
```

image: `traefik:v2.5.3`

```yaml
```

_No response_

xjewer:
tls_certs_not_after metrics exports time after a certificate becomes invalid. This metrics exports 3 labels: cn, sans and serial.

When certificate is rotated, its cn and sans can stay the same, but serial is always different.

Go kit does store all label sets internally and doesn't provide interface deleting them.

Because of this, the only workaround is to reset old certificates' metric to 0.

When TLS watcher receives new dynamic configuration, it only has a set of currently available certificates. Before updating certificates in TlsManager, fetch all currently stored and calculate the diff of obsolete certificates which metrics have to be reset.

Fixes #8606

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -12,7 +12,6 @@ import (
 	"os"
 	"os/signal"
 	"slices"
-	"strings"
 	"syscall"
 	"time"
 
@@ -314,15 +313,7 @@ func setupServer(staticConfiguration *static.Configuration) (*server.Server, err
 	)
 
 	// TLS
-	watcher.AddListener(func(conf dynamic.Configuration) {
-		ctx := context.Background()
-		tlsManager.UpdateConfigs(ctx, conf.TLS.Stores, conf.TLS.Options, conf.TLS.Certificates)
-
-		gauge := metricsRegistry.TLSCertsNotAfterTimestampGauge()
-		for _, certificate := range tlsManager.GetServerCertificates() {
-			appendCertMetric(gauge, certificate)
-		}
-	})
+	watcher.AddListener(tlsUpdater(tlsManager, metricsRegistry.TLSCertsNotAfterTimestampGauge()))
 
 	// Metrics
 	watcher.AddListener(func(_ dynamic.Configuration) {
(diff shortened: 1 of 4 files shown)


nmengin:
Hello @xjewer,

rtribotte:
Hi @xjewer,

Thanks for the feedback. We understand the workaround, but making such changes just to set the zero value as a sentinel value doesn't seem satisfactory.

> I don't have other alternative solutions in mind other than meddle with go-kit or replace it which has much bigger scope.

We haven't looked further yet, but maybe there's a simple approach to exposing deletion methods for registries that support it?

xjewer:
@rtribotte I've created a ticket in go-kit project, will see if that proposal is acceptable.
 
https://github.com/go-kit/kit/issues/1306

rtribotte:
Hi @xjewer, thanks for opening this issue!
I'm not sure it will move fast, and I think just defining our own interface for this Gauge (as for the `CounterWithHeaders` one) would be quicker approach. WDYT?

xjewer:
`CounterWithHeaders` is only a wrapper on top of go-kit.
Adding an interface won't expose go-kit internal data structures.

rtribotte:
@xjewer Sure, but the registries are managed in Traefik, were instruments are build from the vendor specific libraries to satisfy the go-kit interface.
So it seems possible to extend the go-kit interface and expose the internal data structure, and only for this gauge.

PR #11311, opened 2024-11-29
Improve CEL validation on IngressCRD resources

mloiseleur:
### What does this PR do?

Provide validation definition to check expected content: Enum, UniqueItems, Pattern, Minimum or CEL validation expression when appropriate

### Motivation

Improve User XP when writing and/or submitting Traefik's CRD. 
Some IDEs are able to check those rules.
Kubernetes API Server will reject the resources, instead of applying it and forcing user to check error in logs or dashboard.

PR #11670, opened 2025-04-07
Restrict regex validation of HTTP status codes for Ingress CRD resources

jnoordsij:
### What does this PR do?

In #11311 various validations for CRD attributes were added (great change!), to provide for better user experience when creating and/or updating CRD resources. This included various regex expressions, including one for HTTP Status codes (used by the error code middleware).

Given that the RFC only allows for codes >= 100 (["Values outside the range 100..599 are invalid."](https://httpwg.org/specs/rfc9110.html#rfc.section.15)), this MR further restricts the first digit in the regex expression.

### Motivation

Reviewing changes to prepare for using v3.4 (rc) on my Kubernetes cluster.

### More

- [x] Added/updated tests

### Additional Notes

- I've targetted `v3.4` as this further enhances a change that is part of the 3.4 rc. If it's too late to make it in, then I'll retarget this on `master`.
- A similar change may be considered for the time pattern, which can be changed from `^([0-9]+(ns|us|µs|ms|s|m|h)?)+$` to `^([1-9][0-9]*(ns|us|µs|ms|s|m|h)?)+$` to disallow leading zeros, however that might be slight over-engineering and moreover would disallow a time of 0, which might be viable for some purposes

### DIFF
--- a/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
+++ b/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
@@ -1250,7 +1250,7 @@ spec:
                       as ranges by separating two codes with a dash (500-599),
                       or a combination of the two (404,418,500-599).
                     items:
-                      pattern: ^([0-5][0-9]{2}[,-]?)+$
+                      pattern: ^([1-5][0-9]{2}[,-]?)+$
                       type: string
                     type: array
                   statusRewrites:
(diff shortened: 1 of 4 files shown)


jnoordsij:
\cc @mloiseleur

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #11671, opened 2025-04-07
Remove empty (v2) CRD definition file

jnoordsij:
### What does this PR do?

In v3 the old `containo.us` CRDs from v2 were removed from support and consequently from the docs directory. However, it seems one of the files was accidentally reintroduced with no content in a merge commit (https://github.com/traefik/traefik/commit/c1d9b9ee1f59152d444b772666aefd70b11e8e60).

Given this file is empty and not relevant for v3 and higher, it should probably be removed once more.

### Motivation

Spotting it in the changeset of #11311 as being modified, but still empty.

### More

- [ ] Added/updated tests


kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #11673, opened 2025-04-08
More Descriptive Span Names

Alex-Waring:
### What does this PR do?

Fixes: https://github.com/traefik/traefik/issues/11230

Added a new attribute to the traefik tracer that defines the span number, using options defined in the tracing/types package. These are then translated into either a static values, or parts of the request.

### Motivation

Most OTEL visualisers use the SpanName as a way to identify a request. Because Traefik hard codes this to `EntryPoint`, this results in some very ugly looking span names. This PR adds in config options to change this in an expandable way.

### More

- [x] Added/updated tests

### Additional Notes

See the linked issue for a screenshot of why this PR is required.

### DIFF
--- a/docs/content/observability/tracing/overview.md
+++ b/docs/content/observability/tracing/overview.md
@@ -183,3 +183,31 @@ tracing:
 ```bash tab="CLI"
 --tracing.safeQueryParams=bar,buz
 ```
+
+#### `spanName`
+
+_Optional, Default="EntryPoint"_
+
+Configures the name of spans created as part of tracing.
+Valid values are:
+
+- static (uses "EntryPoint")
+- urlPath
+- url
+- hostName
+
+If not provided, spans will be named "EntryPoint"
+
+```yaml tab="File (YAML)"
+tracing:
+  spanName: "urlPath"
+```
+
+```toml tab="File (TOML)"
+[tracing]
+  spanName = "urlPath"
+```
+
+```bash tab="CLI"
+--tracing.spanName=urlPath
+```
--- a/pkg/config/static/static_config.go
+++ b/pkg/config/static/static_config.go
@@ -208,6 +208,7 @@ type Tracing struct {
 	SafeQueryParams         []string           `description:"Query params to not redact." json:"safeQueryParams,omitempty" toml:"safeQueryParams,omitempty" yaml:"safeQueryParams,omitempty" export:"true"`
 	SampleRate              float64            `description:"Sets the rate between 0.0 and 1.0 of requests to trace." json:"sampleRate,omitempty" toml:"sampleRate,omitempty" yaml:"sampleRate,omitempty" export:"true"`
 	AddInternals            bool               `description:"Enables tracing for internal services (ping, dashboard, etc...)." json:"addInternals,omitempty" toml:"addInternals,omitempty" yaml:"addInternals,omitempty" export:"true"`
+	SpanName                string             `description:"Span name, used for most tracing UIs." json:"spanName,omitempty" toml:"spanName,omitempty" yaml:"spanName,omitempty" label:"allowEmpty" file:"allowEmpty" export:"true"`
 	OTLP                    *types.OTelTracing `description:"Settings for OpenTelemetry." json:"otlp,omitempty" toml:"otlp,omitempty" yaml:"otlp,omitempty" label:"allowEmpty" file:"allowEmpty" export:"true"`
 
 	// Deprecated: please use ResourceAttributes instead.
--- a/pkg/middlewares/auth/forward_test.go
+++ b/pkg/middlewares/auth/forward_test.go
@@ -695,7 +695,7 @@ func TestForwardAuthTracing(t *testing.T) {
 			otel.SetTextMapPropagator(autoprop.NewTextMapPropagator())
 
 			mockTracer := &mockTracer{}
-			tracer := tracing.NewTracer(mockTracer, []string{"X-Foo"}, []string{"X-Bar"}, []string{"q"})
+			tracer := tracing.NewTracer(mockTracer, []string{"X-Foo"}, []string{"X-Bar"}, []string{"q"}, "static")
 			initialCtx, initialSpan := tracer.Start(req.Context(), "initial")
 			defer initialSpan.End()
 			req = req.WithContext(initialCtx)
--- a/pkg/middlewares/observability/entrypoint.go
+++ b/pkg/middlewares/observability/entrypoint.go
@@ -9,6 +9,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"github.com/traefik/traefik/v3/pkg/middlewares"
 	"github.com/traefik/traefik/v3/pkg/tracing"
+	"github.com/traefik/traefik/v3/pkg/types"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
 	"go.opentelemetry.io/otel/trace/noop"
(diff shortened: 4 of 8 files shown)


Alex-Waring:
@kevinpollet are you able to help triage this please?

nmengin:
Hello @Alex-Waring,

hendrikkiedrowski:
Looking into the implementation i am stoked to see active development on that. I do have one nit to this. I dont see a sensible implemented way of a path like `GET /user/4` which according to O11y docs should resolve to `GET /user/:id`

I might have overlooked smth. but if not i would be willing to help :)

Alex-Waring:
Hi @hendrikkiedrowski 

I believe that's something for the ingestion tool to handle, each individual span has no way of knowing what elements in a path are static and what represents a variable. Only when comparing many requests can you do that. It's certainly beyond the scope of this PR.

hendrikkiedrowski:
How does nodejs do that then(this is the reference i see and have running)? I struggle to find any documentation to this about an ingestion tool setting for that.

hendrikkiedrowski:
Nevermind. Nodejs knows because it knows... still do you have any idea how you would do this on ingestion?

Alex-Waring:
Yes you're right, it is high cardinality, which is why it's opt-in, however tools like Jaeger, Instana, or DataDog are designed to process this sort of data. It's also worth noting that this is within the span already, so it's not actually adding any more informed, or reducing any sort of deduplication. 

As for how you would create that, if you reviewed a list of paths, you can then identify which values in that path are dynamic, and find a regex to represent them, which is what Instana does.

hendrikkiedrowski:
I see. Okay. Thanks for the explanation. 

Still it seems off for me that we need to rely on backends to do this when [this](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/spanmetricsconnector/README.md) exists.

Alex-Waring:
Hi @nmengin , how is the review going and is there any way I can help (like hopping on a call)? I'm keen to get this in as it's a major part of our v2 to v3 migration

hendrikkiedrowski:
Just to be noted here should anyone else search for a way later to reduce cardinality, or rename stuff **before** it gets to any solution that supports OTEL. There is something calles "transform" processor. As we are using the grafana alloy stack it supports it and you can literally rename/reformat ANY trace or span. @Alex-Waring it might also be interesting to you before this issue is implemented ;) 

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/transformprocessor/README.md

Alex-Waring:
I did have a look at those tools @hendrikkiedrowski, however I would prefer to avoid having to run a whole collector just to correct the name of a span!

Alex-Waring:
Hi @nmengin, any updates on this please?

nmengin:
Hey @Alex-Waring,

Your PR is in our (maintainers') backlog.

OTel is a hot topic on our end, and we are interested in it.
However, we are focused elsewhere, and even if we have planned to move forward soon, I have no ETA to provide yet.

We'll keep you updated as soon as we move forward on it.

Alex-Waring:
Hi @nmengin,

I understand the point of wanting to re-work OTEL and do it properly, however I would suggest that this is a small PR that's adding an optional feature to your current implementation, so it doesn't need to be part of any longer work.

In any case, thank you for the response and please do keep me updated :-)

rtribotte:
Hi @Alex-Waring,

At first glance, we think it was a mistake to name the root span "Entrypoint", as it does not follow the semantic conventions. For this reason, we think this PR should fix the naming to align with the semantic conventions.

However, according to https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name, the request path should not be used; only the matched route should be used to reduce the cardinality.
In short, we think the trace should start with a root span created when a route matches, including the 404 not found case.

Would you be inclined to rework the PR in that direction?

pkg/middlewares/observability/mock_tracing_test.go
Alex-Waring, line 1: The mock tracer was broken, and was only ever reporting the initial entrypoint span, this also means that tests for details on internal spans were incorrect, which i've fixed in this PR

pkg/middlewares/observability/recorder.go
Alex-Waring, line 1: Renamed this resource, as it's now doing more than just the status code

Alex-Waring:
Hi @rtribotte,

I've tried to go in that direction, however traefik has no support for transforming a URI into a url.template and that seems like a much larger piece of work. For now I've just used the rule string. To handle 404s generically, we just report as specific as possible as we make our way down the chain (including multiple middlewares)

rtribotte:
Hi @Alex-Waring,

> I've tried to go in that direction, however traefik has no support for transforming a URI into a url.template and that seems like a much larger piece of work. For now I've just used the rule string. 

The rule string is a good choice IMHO.

> To handle 404s generically, we just report as specific as possible as we make our way down the chain (including multiple middlewares)

I'm not sure understand what this impies. The 404 is the result of the handling of the request by the muxer's default handler. In my previous comment, I was mentioning it because it shouldn't be leftover if the root span is not started at the entryPoint handler level.

Alex-Waring:
Hi @rtribotte,

Ah, I was thinking about 404s caused by no routing rule being found in the traefik config. I don't know if I really want to expand this PR to include changing where the span is started, as that's a much wide project and I think would interfere with any internal discussions on how to do OTEL better.

Alex-Waring:
Hi @rtribotte, any feedback or chance of a review here?

rtribotte:
Hi @Alex-Waring, 
Sure, but according to our previous discussion, I'm not sure about the status of the changes. Are you planning on making further iterations? Would you like us to push the review commits during the review process?

Alex-Waring:
Hi @rtribotte,

Apologies for not being clear, yes I think this PR would now benefit from a review. I don't think I can provide a more sensible solution for handling traffic that is not fully routed without majorly refactoring your OTEL implementation and given you're already working on it that's not something I'm keen on!

Alex-Waring:
Test failure looks like a flake, it passed locally for me

rtribotte:
Hi @Alex-Waring,

Don't worry about the CI for now. We do have flaky tests. We'll see when we make progress on the review.

I previously said that having the router's rule to identify the route was a good idea. However, it is not, since it's possible that the rule contains sensitive data that should not be transmitted in the span.
This needs more thought.

Alex-Waring:
Hi @rtribotte,

That's not a use case I had ever imagined so thanks for pointing it out. Is it safe enough if this option is gated behind config? Also do you have an example of the sort of sensitive information.

Another possible idea, is to allow users to pass in regexes for replacement, and then use the query path, for example:

```
regex: [a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}
identifier: :id:

/foo/64bbd37c-a26a-448a-be49-f9f864fa8fc6/bar -> /foo/:id:/bar
```

Then we skip this whole method of passing down and we always set the span name at the earliest possible time.

hendrikkiedrowski:
@Alex-Waring I would love to see the regexing in this. If there is anything I can help here would be amazing to see!

Alex-Waring:
@rtribotte are you able to provide some feedback here? Initially this PR was supposed to be a quick and easy win to meet a requirement with optional config but it's growing into something bigger.

rtribotte:
Hi @Alex-Waring,

Sorry for the late reply.

We are not convinced by the regex approach, what would then be the default behavior?
As long as users are asked to specify a pattern, we would prefer them to configure the path directly (the acceptable path to use for the span name corresponding to the router).

The semantic conventions [are straightforward](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name), if no target with low cardinality can be used, then only the HTTP method should be used for the span name. What do you think of using only the request method for the span name?

Also, after more thoughts, we think this should be addressed as a fix. We prefer following the semantic conventions, and replace the "Entrypoint" value once and for all, without providing configuration for this.
Could you please rebase the changes on the v3.5 branch and remove the configuration?

Alex-Waring:
Hi @rtribotte,

I've updated the PR to just use the method, in line with the OTEL convention. I can come back and expand this later but for now it removes the hard coding.

kevinpollet (approved):
Thanks 👍

traefiker:

PR #11674, opened 2025-04-08
feat: migrate Traefik Proxy dashboard UI to React

gndz07:
### What does this PR do?

This PR migrates the Traefik dashboard Web UI that were written using Vue to React.

Brings UI improvement:
- omit empty overview graphs
- omit fields with empty values on details pages
- look and feel improvement.

### Motivation

To update the Web UI's codebase and to refresh the UI/UX of the dashboard.

### More

- [ ] Added/updated tests

Co-authored-by: Massimiliano D'Elia [massimiliano.delia@traefik.io](mailto:massimiliano.delia@traefik.io)


rtribotte (approved):
LGTM

kaaax0815:
Just a question. Was there any reason for switching to react instead of staying on vue?

chuoke:
> Just a question. Was there any reason for switching to react instead of staying on vue?

I guess there is no vue developer submit PR

ruuda:
Hi, we package Traefik downsteam. Up to version 3.4.5, it was possible to build Traefik in a regular checkout with just `go generate && go build ./cmd/traefik`, however in v3.5.0, this no longer works; compilation fails with the following error:

```
webui/embed.go:10:12: pattern static: no matching files found
```
Bisecting shows that f16fff577a31d3836ace5c9bbeb5d6ac4a84bd61, which is part of this pull request, is the first commit for which this error occurs. Was this an intentional change?

(I understand that this is not the documented build process, however the documented build process involves third-party Docker containers, which for us defeats the purpose of building from source.)

kevinpollet:
Hello @ruuda,

This means that you are not packaging the webui, right?
If so, maybe just creating an empty static directory should work as a workaround.

PR #11676, opened 2025-04-09
Typo fix on the Explanation Section for User Guide HTTP Challenge.

YapWC:
### What does this PR do?

It fixes some typo found on the Explanation Section for User Guide HTTP Challenge.

### More

- [ ] Added/updated tests


kevinpollet (changes requested):
Hello @YapWC and thanks for this contribution,

If I am not mistaken, in the basic example: https://doc.traefik.io/traefik/v2.11/user-guides/docker-compose/basic-example/ there is only one entrypoint the `web` one.

In the acme-http example the `HTTPS` entrypoint is added to the configuration to be able to serve the TLS certificate obtained from Let's Encrypt.

The web entrypoint will be used to serve the Let's Encrypt challenge.

Therefore, I am not sure to understand the changes?
Could you explain the issue?

YapWC:
Hi @kevinpollet , ah my bad. Because when I read "We configure the HTTPS Let's Encrypt challenge:" I directly assume that there is a typo as it meant to be HTTP Let's Encrypt because it is a HTTP Challenge and not TLS Challenge. I will re-edit and see if it makes sense now. :)

rtribotte (approved):
Thanks!

mmatur (approved):
LGTM

PR #11682, opened 2025-04-11
fix: add Content-Length on response on preFlight request

lbenguigui:
### What does this PR do?

This PR ensures that a `Content-Length: 0` header is included in the response when a preflight `OPTIONS` request is handled by the `headers` middleware.

### Motivation

This change addresses an issue that occurs when the `buffer` middleware is used before in the middleware chain. Without an explicit `Content-Length`, the response may be misinterpreted as having a body, leading to unexpected error.

### More

- [x] Added/updated tests


sdelicata (approved):
LGTM

rtribotte (approved):
LGTM

PR #11698, opened 2025-04-17
Do not warn network missing if connected to a container network

### LINKED ISSUE #11645: Explicit docker network and network_mode (incorrect) warning
### Welcome!

## Context

I am using multiple docker compose services including a vpn, which several of the services use for their networking via `network_mode: service:vpn`.

I am setting a specific named network to be used by traefik (via `--providers.docker.network=<network>`), which is being routed through correctly for internal routing by all services (including all the services that use the vpn).

## Problem

The issue is that `network_mode:` prevents you from adding `networks:`, so Traefik will not be able to find any connected networks directly and warns that the networks that it can access are wrong / inaccessible (as they are not / cannot be directly connected in the config).

## Potential solutions

1. Follow the `network_mode` to discover the right service / container and check against that.
    - As `network_mode` may have [multiple values](https://docs.docker.com/reference/compose-file/services/#network_mode) this may also note that a container has no network access, or is on the host directly where traefik may not have access.
    - This affects [startup order](https://docs.docker.com/compose/how-tos/startup-order/) for `service:...`, so the linked name should already exist and be accessible (even if unhealthy) - this isn't true for `container:...` however.
1. Look at an explicit `traefik.http.routers.<name>=<service>` value and follow that.
    - In this case by *not* doing that it may believe it can route to a service that does not actually have the right network access as the actual service has access, but the service it is being pointed to might not.
    - This has no implicit order, so may refer to a service which is not up (yet) - so may cause future problems / reports if edge cases aren't taken into account.

### What did you see instead?

This is the specific warning that is seen:

```
WRN Could not find network named "<network>" for container "/<name>". Maybe you're missing the project's prefix in the label? 
```

```
Version:      3.3.4
Codename:     saintnectaire
Go version:   go1.23.6
Built:        2025-02-25T10:11:01Z
OS/Arch:      linux/amd64
```

The actual config is too large to post, but this shows the minimum *changes* needed to reproduce the (incorrect) warning:

```yaml
networks:
  random_name:
    name: random_name # so no project prefix needed

services:
  traefik:
    # ... add image and setup etc
    command:
      - "--providers.docker.network=random_name" # match the network name above
    networks:
      - random_name # and make sure we're actually connected to it

  whoami:
    # ... add image and setup etc
    network_mode: service:traefik # merge with the traefik network
    labels:
      - "traefik.http.routers.whoami=traefik@docker" # makes no difference but I like to be explicit
      - "traefik.http.routers.whoami.rule=Host(`whoami.docker.localhost`)"
```

_No response_

holysoles:
# What does this PR do?

If Traefik is provided a specific docker network to use with `--providers.docker.network`, and a service in the docker compose is configured with a `network_mode: service:traefik`, then an incorrect warning is written to the log on startup:

```
WRN Could not find network named "foo" for container "/simple-service". Maybe you're missing the project's prefix in the label?
```

This warning is incorrect because[ after it is written here](https://github.com/traefik/traefik/blob/v3.3/pkg/provider/docker/config.go#L334), alternative methods to discover the container's IP address are performed. In this case, the docker API returns the correct connected container (Traefik) and we recursively call `getIPAddress()` for that container and successfully retrieve an IP.

This PR ensures we only write this warning message after we try checking if the service is connected to another container's network.

# Motivation

Given that Traefik can still start correctly and discover the IP address of the desired service container, this warning is unnecessary and confuses users.

Fixes #11645

### More

- [ ] Added/updated tests

### Additional Notes

Some suggested test cases:

- A docker compose file that can test the change in behavior before/after:

```yaml
networks:
  foo:
    name: foo

services:
  traefik:
    image: "traefik/traefik:latest"
    container_name: "traefik"
    restart: unless-stopped
    command:
      - "--log.level=WARN"
      - "--api.insecure=true"
      - "--providers.docker=true"
      - "--entrypoints.web.address=:81"
      - "--providers.docker.network=foo"
    volumes:
    ports:
      - 8000:81
      - 8080:8080
    networks:
      - foo

  whoami:
    image: traefik/whoami
    container_name: "simple-service"
    restart: unless-stopped
    network_mode: service:traefik
    labels:
      - "traefik.http.routers.whoami.rule=PathPrefix(`/`)"
      - "traefik.http.routers.whoami.entrypoints=web"
      - "traefik.http.services.whoami.loadbalancer.server.port=80"
```

- And a docker compose file that can test that we still write this warning in cases where it makes sense:

```yaml
networks:
  foo:
    name: foo
  bar:
    name: bar

services:
  traefik:
    image: "traefik/traefik:latest"
    container_name: "traefik"
    restart: unless-stopped
    command:
      - "--log.level=WARN"
      - "--api.insecure=true"
      - "--providers.docker=true"
      - "--entrypoints.web.address=:81"
      - "--providers.docker.network=foo"
    volumes:
    ports:
      - 8000:81
      - 8080:8080
    networks:
      - foo

  whoami:
    image: traefik/whoami
    container_name: "simple-service"
    restart: unless-stopped
    networks:
      - bar
    labels:
      - "traefik.http.routers.whoami.rule=PathPrefix(`/`)"
      - "traefik.http.routers.whoami.entrypoints=web"
      - "traefik.http.services.whoami.loadbalancer.server.port=80"
```

### DIFF
--- a/pkg/provider/docker/config.go
+++ b/pkg/provider/docker/config.go
@@ -340,7 +340,6 @@ func (p *DynConfBuilder) getIPAddress(ctx context.Context, container dockerData)
 			}
 
 			netNotFound = true
-			logger.Warn().Msgf("Could not find network named %q for container %q. Maybe you're missing the project's prefix in the label?", container.ExtraConf.Network, container.Name)
 		}
 	}
 
@@ -382,6 +381,9 @@ func (p *DynConfBuilder) getIPAddress(ctx context.Context, container dockerData)
 		return p.getIPAddress(ctx, containerParsed)
 	}
 
+	if netNotFound {
+		logger.Warn().Msgf("Could not find network named %q for container %q. Maybe you're missing the project's prefix in the label?", container.ExtraConf.Network, container.Name)
+	}
 	for _, network := range container.NetworkSettings.Networks {
 		if netNotFound {
 			logger.Warn().Msgf("Defaulting to first available network (%q) for container %q.", network, container.Name)


sdelicata (approved):
Hi @holysoles, and thanks for your contribution!
LGTM

kevinpollet (changes requested):
Hello @holysoles and thanks for your contribution,

As the log exists in v2.11 and this is a bugfix, please could you rebase your pull request on branch v2.11?

holysoles:
I rebased my branch (still unfortunately named v3.3..) off v2.11.

Since there are some differences between the two branches, I've made updates to reflect the v2.11 source. If you'd like me to open a v3.3 PR as well let me know, since it won't be able to be simply cherry-picked

holysoles:
This PR should still be good to go, but after rebasing off current v2.11, I can't pass `TestRedisSentinelSuite` in the CI. I also can't pass the test locally even with an older commit checked out. Opened #11743  to followup on trying to fix that

kevinpollet:
Hello @holysoles,

I rebased your pull request on branch v3.4, as branch v2.11 is now in security maintenance mode.
See https://doc.traefik.io/traefik/deprecation/releases/#versions.

holysoles:
Thanks for rebasing @kevinpollet , I have re-tested the behavior of the update and it still looks good.

The CI here has been challenging to pass.. both because of #11699 (which I'm hoping to look into soon), but also now it seems like binary builds are terminating due to OOM, even after many retries.

Should I open a PR to enable GOGC using #11721 on GitHub runners? That does appear to fix the problem when I tested on my fork [here](https://github.com/holysoles/traefik/actions/runs/15011650923)

mmatur:
> Thanks for rebasing @kevinpollet , I have re-tested the behavior of the update and it still looks good.
> 
> The CI here has been challenging to pass.. both because of #11699 (which I'm hoping to look into soon), but also now it seems like binary builds are terminating due to OOM, even after many retries.
> 
> Should I open a PR to enable GOGC using #11721 on GitHub runners? That does appear to fix the problem when I tested on my fork [here](https://github.com/holysoles/traefik/actions/runs/15011650923)

Thanks @holysoles,

We have merged a [PR](https://github.com/traefik/traefik/pull/11756) to enable GOGC by default

holysoles:
should be set now, had issues getting the unit test job to finish

kevinpollet (approved):
Thanks 👍

PR #11710, opened 2025-04-18
Add ability to pass headers to plugin fetch action

zalbiraw:
### What does this PR do?

Adds headers support for plugins fetching action 

### Motivation

As a user I would like Traefik to be able to load plugins from my private repository

### More

- [ ] Added/updated tests

### Additional Notes

The question is, should we add support for managing headers in general, or should we only support this for auth headers?  Should we have a general headers section that sets headers for all the download calls?


zalbiraw:
Closing based on internal discussions

PR #11715, opened 2025-04-23
Add SpanID and TraceID accessLogs fields only when tracing is enabled

rtribotte:
### What does this PR do?

This PR fixes the access-logs by adding SpanID and TraceID fields only when tracing is enabled.

### Motivation

To avoid producing SpanID and TraceID fields with invalid zero IDs.

### More

- [x] Added/updated tests


sdelicata (approved):
LGTM

lbenguigui (approved):
LGTM

PR #11721, opened 2025-04-27
Adding garbage collector as variable in compilation 

borjaOrtizLlamas:
### What does this PR do?

Add the possibility of run the garbage collector of go during  compiling.  By default its off as its now. 

### Motivation

I have been experiencing errors of OOM during compilation due have the GOGC off, probably for other its not a problem but our runners do not have to much  RAM and i think it can be nice to have it.


nmengin:
Hello @borjaOrtizLlamas,

kevinpollet (approved):
Thanks 👍

PR #11727, opened 2025-05-02
Pass header identifying Traefik as the source of forward-auth requests

ItalyPaleAle:
### What does this PR do?

When Traefik makes a request to the Forward Auth endpoint, to validate the session, it includes a header `X-Traefik-Forward-Auth: 1` which identifies Traefik as the source of the request.

### Motivation

Assume that you have the Forward Auth middleware configured to query `http://example.com/authorized`, which responds with 200 if the session is valid, or redirects if the user needs to authenticate.

When Traefik makes the request to the `/authorized` endpoint, there's currently no way for the service to know if the request comes from Traefik (and should be considerd a "background" request) or if the user visited `http://example.com/authorized` directly.

Being able to differentiate between a background request made by Traefik or an explicit request from the user can allow for experiences such as responding with 204 (No Content) to Traefik, but showing a nicer page when the user invokes the endpoint directly.

### More

- [X] Added/updated tests

### Additional Notes

I am happy to accept suggestions if using a header `X-Traefik-Forward-Auth: 1` is not the recommended approach. Besides other header names, other options could include setting a specific User Agent header identifying Traefik, although that may break some scenarios (e.g. if the auth service uses the user agent as part of its session)


nmengin:
Hello @ItalyPaleAle,

Thanks for submitting this PR. 

Unfortunately, we don’t think the goal of this PR applies to enough people for us to merge it into the main branch.

After discussing the PR with the maintainers, we think that the use case you described can be satisfied using the Header middleware.
If you have a more global need, please describe it in an issue to discuss it and define a way to solve it.

I close the PR.

ItalyPaleAle:
@nmengin thanks for the review.

Could you please give an example of how to apply a header in requests made through the forward auth middleware? I can't find an option in the [docs](https://doc.traefik.io/traefik/middlewares/http/forwardauth/).

ItalyPaleAle:
I spent some time researching this, but I don't think it would do what I need.

Let me give a more concrete example. I maintain a [traefik-forward-auth](https://github.com/ItalyPaleAle/traefik-forward-auth) server that works with OAuth2 (and other identity solutions like Tailscale networks). Users can set it up with a config like this:

```yaml
http:
  middlewares:
    forwardAuth:
      forwardauth:
        address: "https://localhost:4181"
        authResponseHeaders: ["X-Forwarded-User"]
        trustForwardHeader: true
  routers:
    traefik-forward-auth:
      entryPoints: ["https"]
      rule: "Host(`auth.example.com`)"
      service: traefik-forward-auth
      tls: {}
    whoami:
      entryPoints: ["https"]
      middlewares: ["forwardAuth@file"]
      rule: "Host(`whoami.example.com`)"
      service: whoami
      tls: {}
  services:
    traefik-forward-auth:
      loadBalancer:
        healthCheck: {"interval": "5s", "path": "/healthz"}
        servers:
          - url: http://localhost:4181/
    whoami:
      loadBalancer:
        servers:
          - url: http://localhost:4545/
```

In a normal flow:

1. User visits `https://whoami.example.com`
2. Traefik makes a request to `http://localhost:4181/` (the endpoint of traefik-forwad-auth) and gets a 302 response with a Location header
3. Traefik returns the 302 response and the Location header to the user
4. User completes the authentication flow with the external IdP
5. User is redirected back to `https://auth.example.com/oauth2/callback?...` which is proxied by Traefik to TFA
6. Auth flow is complete, user is sent back to `https://whoami.example.com`
7. Traefik makes another request to TFA (`http://localhost:4181/`) and this time gets a 200 response

The problem is that the user could hit `https://auth.example.com/` directly, which would be proxied to `http://localhost:4181/`. In this case the request still comes from Traefik, but it's not a background request made by the Forward Auth middleware.

With the Forward Auth middleware, there's currently no way to be able to tell if a request hitting `http://localhost:4181/` is due to:

1. Traefik doing a background request (as part of the Forward Auth middleware)
2. The user visiting `https://auth.example.com/`

This matters because:

1. If the request is made by Traefik's Forward Auth middleware, the TFA server should respond in the quickest way as possible, such as returning an empty page (e.g. a 204 response).
    - Speed here is very important because this request is in the hot path of every request that is protected by the Forwar Auth middleware, so it can make everything slower.
3. If it's the user that is visiting `https://auth.example.com/`, ideally we'd want to show a nicer page, something that tells the user where they are (and maybe offer a logout link, etc).
   - Rendering a page like this takes time, and it also requires sending a longer response. We don't want to slow down background requests.

ItalyPaleAle:
@nmengin could you please take a look at my comment above?

nmengin:
Hello @ItalyPaleAle,

Sorry for the delay, I guess we missed your comment during our latest triages.
I will add to the plate for the next one later this week.

ItalyPaleAle:
@nmengin do you have any update for us please?

rtribotte:
Hi @ItalyPaleAle,

Sorry for the delay, did you tried configuring a Headers middleware before the ForwardAuth?
To identify that Traefik is sending the request to the auth server, adding a custom header on the request with the Headers middleware, that will be copied to the auth server request (configurable with the [`authRequestHeaders`](https://doc.traefik.io/traefik/middlewares/http/forwardauth/#authrequestheaders) option), should do the trick.

ItalyPaleAle:
@rtribotte thanks for the response.

Not really sure I understand what you're suggesting.

What I need is a header that is sent when the request is **originated by** Traefik, and not **proxied by** Traefik. `authRequestHeaders` copies headers from the request, but it doesn't allow adding new headers. And adding it to the route would mean that the downstream app would receive the header too (plus, it's additional complexity I'd need to add to the docs for traefik-forward-auth).

rtribotte:
Hi @ItalyPaleAle,

Could you please open an issue to present your proposal? That would be the best way to evaluate community interest in this.
We understand your need, but we thought it was too niche to address. Hence, our suggestion was to rely on the Headers middleware. However, we can change our minds if the proposal gets traction.

ItalyPaleAle:
I've opened https://github.com/traefik/traefik/issues/11944

It's definitely a niche thing, meaning that people who would leverage this are far and few (just maintainers of forward auth apps...). 

What's harder to measure is the number of end users that would benefit from this, given a performance improvement in the forward auth middlewares

PR #11733, opened 2025-05-05
Connect to Tailscale using tsnet

ItalyPaleAle:
Picking up from an attempt made in #11086, but updated for Traefik 3 and with broader support.

Allows Traefik to integrate with a Tailscale network directly, using [tsnet](https://tailscale.com/kb/1244/tsnet). While it's already possible to connect Traefik to Tailscale, it does require running a Tailscale daemon which then connects to Traefik using HTTP(S). Using the tsnet package, it's possible to make Traefik appear in a Tailscale network (Tailnet) directly, without the need for an additional daemon. Each Traefik instance would then appear as a standalone device in the Tailnet.

Screenshot 2025-05-04 at 21 41 39

From a user perspective, this requires setting the `tsnet` property on an entrypoint, for example:

```yaml
entryPoints:
  ts:
    address: ":8000"
    tsnet:
      hostname: "traefik-dev"
      authKey: "tskey-auth-..."
```

It's possible to pass an explicit auth key, or it will be read from the `TS_AUTHKEY` env var.

Things I've tested and confirmed working:

- HTTP/1 and HTTP/2, including TLS with the built-in certificate
- HTTP/3, including TLS (with the built-in certificate), and capable of listening on the same port as HTTP/1 or 2.
   - Only works with the IPv4 address issued by Tailscale

While this seems to be working for me, I only tested it on my local machine and it will need some more testing especially in a variety of scenarios (and I haven't tested using TLS certificates from Tailscale, as that's not enabled for my tailnet... although I don't see why it wouldn't work).

There are also a few more things that could be added:

- Support for Tailscale funnel. This requires adding a new listener (`ListenFunnel`)
- It may be nice, as an option, also to pass the user ID of the Tailscale user connecting to the endpoint in the `X-Forwarded-User` header (which would make my [traefik-forward-auth](https://github.com/ItalyPaleAle/traefik-forward-auth/blob/main/docs/01-quickstart.md#authenticate-with-tailscale) obsolete, at least for Tailscale :) )


ItalyPaleAle:
Is there any update on this? Any interest in this P?

nmengin:
Hello @ItalyPaleAle,

We're definitely interested in your contribution.

We've set the status to "needs-review" to allow us to check the PR content.
It's now on our (maintainers') plate, we'll keep you updated once the analysis is done.

To help us review the PR, could you please rebase it?

svenvg93:
@ItalyPaleAle Hi, any updates on this PR? I'm currently using TSDProxy, but the project seems abandoned. I was hoping to switch to Traefik with your PR. 😊

ItalyPaleAle:
I've rebased the PR as asked, but it's up to the maintainers to review this

apollo13:
Wow, this looks great. One question @ItalyPaleAle, is it possible to use a single use auth key and have the tailscale state stored on disk? As far as I understand the code and comments https://github.com/tailscale/tailscale/blob/d15b2312c4fb7b8ea1f98c5c80f7f72aed784b5d/tsnet/tsnet.go#L67-L77 it should do so by default. It might make sense to document the directory so users know which folder to map out of the container for a persistent experience.

pkg/config/static/entrypoints.go
```diff
@@ -163,6 +164,13 @@ func (u *UDPConfig) SetDefaults() {
 	u.Timeout = ptypes.Duration(DefaultUDPTimeout)
 }
 
+type TSNetConfig struct {
+	Hostname  string `description:"Hostname for the node; if empty, defaults to the endpoint's name" json:"hostname,omitempty" toml:"hostname,omitempty" yaml:"hostname,omitempty" export:"true"`
+	Ephemeral bool   `description:"Register the node as ephemeral." json:"ephemeral,omitempty" toml:"ephemeral,omitempty" yaml:"ephemeral,omitempty" export:"true"`
+	AuthKey   string `description:"Auth key for the node (overrides the TS_AUTHKEY env var)." json:"authKey,omitempty" toml:"authKey,omitempty" yaml:"authKey,omitempty" export:"true"`
+	// TODO: Funnel to enable listening on a funnel
```
apollo13: Maybe also add `ControlURL` to be able to run integration tests against a headscale instance in the future?

ItalyPaleAle: Done

ItalyPaleAle:
@apollo13 thanks for the review :)

> One question @ItalyPaleAle, is it possible to use a single use auth key and have the tailscale state stored on disk? As far as I understand the code and comments https://github.com/tailscale/tailscale/blob/d15b2312c4fb7b8ea1f98c5c80f7f72aed784b5d/tsnet/tsnet.go#L67-L77 it should do so by default. It might make sense to document the directory so users know which folder to map out of the container for a persistent experience.

I think it should work but need to test it later today.

Agree on the directory. I am thinking of hardcoding a specific value and then allowing it to be overridden via the config. Would that work?

Is there a preferred folder I should use as default?

ItalyPaleAle:
I added the ability to configure the directory where to store the Tailscale's state, after confirming that one-use keys can be used. I couldn't find many examples in Traefik of using persistent volumes, so I arbitrarily set `/var/tailscale` as the default folder; I can change it to whatever you recommend.

apollo13:
> so I arbitrarily set /var/tailscale as the default folder; I can change it to whatever you recommend.

`/var/tailscale` would be rather uncommon, I'd recommend `/var/lib/tailscale` instead. Btw what was the default if you don't set anything? I could imagine that being in line with user expectations?

mivalsten:
I think this might be better located in `/etc/traefik/tailscale` as acme.json is stored there by default, so most people already have this directory persisted.

ItalyPaleAle:
Both a subdir of /etc/Tailscale or /var/lib/tailscale seem reasonable to me. Let me know your preference and I can make the change

Docs say that the default is the value of os.UserConfigDir, which would likely be $HOME/.config in the container. I don't think that's the best option here IMHO

apollo13:
`/etc` seems weird to me since it is state not config. `/var/lib/tailscale` would match what a normal tailscale installation would use, so maybe not that uncommon? Though not really usable if you run traefik as non-root (but neither is `/etc/` then).

ItalyPaleAle:
Updated the PR to use /var/lib/tailscale

Zolfax:
Hi @ItalyPaleAle 

Disclaimer: I am in no means affiliated with traefik nor do I have deep Go knowledge.

I've build a docker image based on your contribution state for testing this functionality: zolfax/traefik:beta
I've tried to set traefik up with the following features enabled:
- ephemeral: false - I want to persist my machines on tailscale with their corresponding state.
- HTTPS enabled - I want traefik to automatically fetch certificates via the certificate resolver.

*Ephemeral config option:*
With my YAML approach, I tried a lot of different values for the entrypoint.
Somehow it did not work as intended and resulted in always ephemeral tailscale machines.
This happened regardless of no value added (which should result in a non-ephemeral machine), setting it explicitely to `true` or `false`.
I'm unsure whether this is a bug in tsnet or if I have misinterpreted something else, however, the configuration was good.

*HTTPS certificates:*
> Unable to fetch certificate for domain "adminer.<my-tsnet>.ts.net" error="Get \"http://local-tailscaled.sock/localapi/v0/cert/adminer.<my-tsnet>.ts.net?type=pair\": dial unix /var/run/tailscale/tailscaled.sock: connect: no such file or directory" providerName=tsresolver.tailscale
Usually traefik has access through a shared port within the tailscale container.
With this implementation, this `/var/run/tailscale/tailscaled.sock` is still required, therefore still requiring a tailscale client to run locally.

Here's an excerpt of my configuration to try it out yourself.

Static configuration: traefik.yml
```
api:
 dashboard: true

entryPoints:
  http:
    address: ":80"
    http:
      redirections:
        entrypoint:
          to: https

  https:
    address: ":443"

  adminer:
    address: ":443"
    tsnet:
      hostname: "adminer"
      authKey: "tskey-auth-<something>"
      dir: "/tmp/tailscale/adminer"
      #ephemeral: true
      #ephemeral: false

certificatesResolvers:
    tsresolver:
        tailscale: {}

serversTransport:
  insecureSkipVerify: true

tcpServersTransport:
  tls:
    insecureSkipVerify: true

providers:
  docker:
    defaultRule: "Path(`/undefined`)"
    exposedByDefault: false
    watch: true

accessLog:
  filePath: "/logs/access.log"
  fields:
    headers:
      names:
        User-Agent: "keep"

log:
  filePath: "/logs/traefik.log"
  level: "INFO"
```

And correspondingly the adminer config with docker labels:
```
services :
  adminer:
    image: adminer:5
    hostname: adminer
    container_name: adminer
    restart: no
    user: "1001"
    environment:
      ADMINER_DEFAULT_SERVER: postgres
    ports:
      - "8080:8080"
    labels:
      traefik.enable: true
      traefik.http.routers.adminer_https.rule: Host(`adminer.<my-tailscale>.ts.net`)
      traefik.http.routers.adminer_https.entrypoints: https,adminer
      traefik.http.routers.adminer_https.tls: true
      traefik.http.routers.adminer_https.tls.certResolver: tsresolver
      traefik.http.services.adminer.loadbalancer.server.port: 8080

```

The error message tells me that it tries to resolve the certificate, but was unable to, because the socket was unavailable.

If there is a configuration error on my end, feel free to point it out.
I will gladly retest it with a different configuration or changes on your side.

apollo13:
I think the current tailscale cert integration relies on the actual tailscale service to be running. When using tsnet no such socket exists at all... So I don't think that is possible currently

Zolfax:
> I think the current tailscale cert integration relies on the actual tailscale service to be running. When using tsnet no such socket exists at all... So I don't think that is possible currently

You might be right with the current implementation.
Traefik uses the tscert library of tailscale, which is a highly stripped down library of the `tailscale.com/client/tailscale`. 
See: https://github.com/tailscale/tscert 
With tsnet, this could be changed or at least extended to support both variants, as tsnet imports both `tailscale.com/client/local` and `tailscale.com/client/tailscale`.

I've seen other solutions tackle this differently without the explicit use of a shared in `tailscaled.sock`.
I am not sure exactly how TSDProxy does it, but it is working as a standalone proxy without any tailscale docker required.
Example Solution: https://github.com/almeidapaulopt/tsdproxy/blob/main/internal/proxyproviders/tailscale/proxy.go

However, I am not that fluent in go, so take this information with a grain of salt.
Furthermore, I'm unaware how much refactoring traefik would have to do to achieve the same with the tailscale client.

The general reasoning stands however:
With this implementation, the communication to fetch the https certificates goes through a shared `/var/run/tailscale/tailscaled.sock` which relies on external configuration, like a separate tailscale running with their own authkey.

At the same time, the authkey within the `traefik,yml` is used to log in to tailscale using the tsnet.

This can cause a mismatch in configuration, as this implicitely requires both to be set up with the same authkey / tailscale account, as we there should not be a possibility to check that both the socket runs on the same account as the tsnet implementation in here.

I personally am fine with this being the case, but it should be documented that this needs to be linked to the same tailscale account. See: https://github.com/tailscale/tailscale/blob/v1.84.3/tsnet/tsnet.go#L29-L30

Some other ideas:
- ListenTLS as a Listener if TLS is set on `entrypoint.<name>.tsnet.tls:  true` - https://github.com/tailscale/tailscale/blob/v1.84.3/tsnet/tsnet.go#L1040
  - Not sure if this would do the job out of the box. I do not think so, but it is documented as a different type of listener.
- LocalClient to do certain LocalClient tasks, such as Certificate generation - https://github.com/tailscale/tailscale/blob/v1.84.3/tsnet/tsnet.go#L226
  - See: https://github.com/tailscale/tailscale/blob/v1.84.3/client/local/local.go#L1039
  - Or see: https://pkg.go.dev/tailscale.com@v1.84.3/client/local#Client.CertPair

PS: As I am really not well-versed in go, I am limited in what I can contribute. If you an idea what I could contribute, feel free to point it out.

apollo13:
> This can cause a mismatch in configuration, as this implicitely requires both to be set up with the same authkey / tailscale account, as we there should not be a possibility to check that both the socket runs on the same account as the tsnet implementation in here.

Yes, this is certainly not ideal. That said with a listener exposed via tsnet you can already be sure that the connection is secured via state of the art encryption. HTTPs offers little to no benefit here from a security PoV. 

> ListenTLS as a Listener if TLS is set on entrypoint.<name>.tsnet.tls:  true

I think this might actually work from looking at the code. You could test this for now in your build by using it here https://github.com/traefik/traefik/blob/b8f364c89772639246737c3d406963d675272331/pkg/server/server_entrypoint_tcp.go#L491 (unconditionally for now, just so we see if it works).

Zolfax:
The main issue here is with the certificate resolution.

So to have a clear cut architecture, we could change the following:

```
certificatesResolvers:
    myresolver:
        tailscale: 
            useTsnet: true
            entrypoint: <tsnet-entrypoint>
```

--> Drawback: you would have to directly link your tailscale certificate resolution to an entrypoint: Design flaw

or 

```
certificatesResolvers:
    myresolver:
        tailscale: 
            tsnet:
                authKey: "ts-auth-...."
                controlurl: <custom-url-if-needed>
```

--> Drawback: we are duplicating the authentication information from one or multiple entry points

The cleanest way would be to have a seperate tsnet authentication root node that can be referenced in both entrypoint for machine declaration and certificate resolution:

```
entryPoints:
  ts:
    address: ":8000"
    tsnet: 
      hostname: "traefik-dev"
      authentication: mytsnet
      ephemeral: true
      dir: /tmp/traefik-dev # this stores unique state for each machine
      
certificatesResolvers:
    myresolver:
        tailscale: 
            tsnet: mytsnet

tsnet:
  mytsnet:
    authkey: <ts-auth-...>
    controlurl: <custom-url-if-needed>
```

Implementation Advantages:
- Storing of tailscale connection authentication in one place.
- Separation of concerns regarding different tailscale machines: auth is separate from machine definition.
- Certificate resolvers are separate as well, can refer to the tailscale configuration. Reuse of tailscale configuration with optional referral of tsnet authentication for direct certificate generation.
- Leaves dynamic configuration untouched.

What would need to be adjusted for this to work:
- Implement tsnet authentication configuration on root level.
- Fetching the authentication for `tsnet.Server` from that reference.
- Adjusting the certificate resolver for tailscale to work with the tsnet authentication reference if it is present. 

Additional features through this implementation:
- One traefik instance with https enabled 
- multiple tailscale machines in different tailscale networks 
- with correct https certificate resolution

apollo13:
FWIW personally I think tailscale certs (maybe aside from the tls flag) for the tsnet entrypoint are pretty much out of scope for this PR and can be added later on (incremental improvements, get the MVP out…)

Zolfax:
> FWIW personally I think tailscale certs (maybe aside from the tls flag) for the tsnet entrypoint are pretty much out of scope for this PR and can be added later on (incremental improvements, get the MVP out…)

I agree that getting the MVP out should be the priority. 👍 

Personally I think we should still think a little about the configuration architecture to support multiple entrypoints without duplicating the tailscale configuration authentication. Specifically since the certificate resolver will need either direct access to the socket or require a different way to fetch the certificates. A link to the entrypoint would break the loose coupling of entrypoint and certResolver.

So to get to the root of my https feedback:

Is there a design consideration why we cannot put the tsnet authentication somewhere outside of the entrypoint to prevent duplication of authentication information? It does not directly have to be a root node, but I do see a benefit in being able to refer to it from the certificate resolver. Examples could also be: `auth.tsnet` or `global.auth.tsnet`.

If putting the tsnet authentication outside of entrypoint is not the way to go, would it be accepted that we duplicate the tsnet auth configuration for the certificate resolver from an architecture design? Getting approval for this here already will make the future certificate resolver solution easier. It will use duplicated configuration for the tsnet auth, but it would keep the coupling loose.

To conclude, I'm fine with both solutions:
- Separate authentication for tsnet from entrypoint configuration to a different config setting, so that it can be reused for the certificate resolver in the future. (additional implementation changes needed)
- Accept that we will need to duplicate the tsnet authentication for certificate generation at the certResolver level if the direct tsnet certificate generation is implemented. (no changes needed, just raising awareness for this solution for certResolver)

Last, a check why the ephemeral flag does not work as intended is still something to be investigated.
If you get this to work with yaml and non-ephemeral nodes, I would be happy to see it work.
A simple "works on my machine" is enough for me here. 😄

apollo13:
>  If you get this to work with yaml and non-ephemeral nodes, I would be happy to see it work.

I didn't run it myself yet, but how did you verify that non-empheral does not work? Does the Admin UI show it as empheral, or is it just a new host every time? Did you map out the tailscale state dir?

Zolfax:
> I didn't run it myself yet, but how did you verify that non-empheral does not work? Does the Admin UI show it as empheral, or is it just a new host every time? Did you map out the tailscale state dir?

I have run several entrypoint configurations within my tailscale network.
All machines showed up with the Ephemeral information in tailscale.
It might also be an issue on tailscale end that this is displayed incorrectly for tsnet nodes?

After shutting the traefik instance down, the nodes persist for some time.
I have manually removed them to retest other configurations.
They did not get removed automatically regardless of the config.
However, they still have the ephemeral tag on them.

edit: state dir was mapped out.

edit2: my bad, this was a configuration issue on my part.
If the authkey is defined as ephemeral, it takes precendence over the actual setting in tsnet.
Status ephemeral: solved 👍

docs/content/reference/static-configuration/cli-ref.md
```diff
@@ -324,6 +324,24 @@ ReadTimeout is the maximum duration for reading the entire request, including th
 `--entrypoints.<name>.transport.respondingtimeouts.writetimeout`:  
 WriteTimeout is the maximum duration before timing out writes of the response. If zero, no timeout is set. (Default: ```0```)
 
+`--entrypoints.<name>.tsnet`:  
+Tailscale network configuration. (Default: ```false```)
+
+`--entrypoints.<name>.tsnet.authkey`:  
+Auth key for the node (overrides the TS_AUTHKEY env var).
```
Zolfax, line 331:
```suggestion
Auth key for the node (overrides the `TS_AUTHKEY` env var of tsnet library).
```

docs/content/reference/static-configuration/cli-ref.md
```diff
@@ -324,6 +324,24 @@ ReadTimeout is the maximum duration for reading the entire request, including th
...
+
+`--entrypoints.<name>.tsnet.controlurl`:  
+URL for the Tailscale control plane, when not using the default one.
+
+`--entrypoints.<name>.tsnet.dir`:  
+Path where Tailscale stores its state. Set it to a persistent volume to allow Traefik to remain authenticated with Tailscale. Defaults to /var/lib/tailscale.
```
Zolfax, line 337:
```suggestion
Path where Tailscale stores its state. For multiple entrypoints, this needs to point to different directories. Set it to a persistent volume to allow Traefik to remain authenticated with Tailscale. (Default: ```/var/lib/tailscale```)
```

docs/content/reference/static-configuration/cli-ref.md
```diff
@@ -324,6 +324,24 @@ ReadTimeout is the maximum duration for reading the entire request, including th
...
+
+`--entrypoints.<name>.tsnet.ephemeral`:  
+Register the node as ephemeral. (Default: ```false```)
+
+`--entrypoints.<name>.tsnet.hostname`:  
+Hostname for the node; if empty, defaults to the endpoint's name
```
Zolfax, line 343:
```suggestion
Hostname for the node. Be aware of Tailscale machine name conventions. (Default: entrypoint's ```<name>```)
```

docs/content/reference/static-configuration/cli-ref.md
```diff
@@ -324,6 +324,24 @@ ReadTimeout is the maximum duration for reading the entire request, including th
...
 WriteTimeout is the maximum duration before timing out writes of the response. If zero, no timeout is set. (Default: ```0```)
 
+`--entrypoints.<name>.tsnet`:  
+Tailscale network configuration. (Default: ```false```)
+
+`--entrypoints.<name>.tsnet.authkey`:  
+Auth key for the node (overrides the TS_AUTHKEY env var).
+
+`--entrypoints.<name>.tsnet.controlurl`:  
+URL for the Tailscale control plane, when not using the default one.
```
Zolfax, line 334:
```suggestion
URL for the Tailscale control plane (Default: ```https://controlplane.tailscale.com```)
```

docs/content/reference/static-configuration/env-ref.md
```diff
@@ -324,6 +324,24 @@ ReadTimeout is the maximum duration for reading the entire request, including th
 `TRAEFIK_ENTRYPOINTS_<NAME>_TRANSPORT_RESPONDINGTIMEOUTS_WRITETIMEOUT`:  
 WriteTimeout is the maximum duration before timing out writes of the response. If zero, no timeout is set. (Default: ```0```)
 
+`TRAEFIK_ENTRYPOINTS_<NAME>_TSNET`:  
+Tailscale network configuration. (Default: ```false```)
+
+`TRAEFIK_ENTRYPOINTS_<NAME>_TSNET_AUTHKEY`:  
+Auth key for the node (overrides the TS_AUTHKEY env var).
```
Zolfax, line 331:
```suggestion
Auth key for the node (overrides the `TS_AUTHKEY` env var of tsnet library).
```

docs/content/reference/static-configuration/env-ref.md
```diff
@@ -324,6 +324,24 @@ ReadTimeout is the maximum duration for reading the entire request, including th
...
+
+`TRAEFIK_ENTRYPOINTS_<NAME>_TSNET_CONTROLURL`:  
+URL for the Tailscale control plane, when not using the default one.
```
Zolfax, line 334:
```suggestion
URL for the Tailscale control plane (Default: ```https://controlplane.tailscale.com```)
```

docs/content/reference/static-configuration/env-ref.md
```diff
@@ -324,6 +324,24 @@ ReadTimeout is the maximum duration for reading the entire request, including th
...
+
+`TRAEFIK_ENTRYPOINTS_<NAME>_TSNET_DIR`:  
+Path where Tailscale stores its state. Set it to a persistent volume to allow Traefik to remain authenticated with Tailscale. Defaults to /var/lib/tailscale.
```
Zolfax, line 337:
```suggestion
Path where Tailscale stores its state. For multiple entrypoints / node configurations, this needs to point to different directories. Set it to a persistent volume to allow Traefik to remain authenticated with Tailscale. (Default: ```/var/lib/tailscale```)
```

docs/content/reference/static-configuration/env-ref.md
```diff
@@ -324,6 +324,24 @@ ReadTimeout is the maximum duration for reading the entire request, including th
...
+
+`TRAEFIK_ENTRYPOINTS_<NAME>_TSNET_EPHEMERAL`:  
+Register the node as ephemeral. (Default: ```false```)
+
+`TRAEFIK_ENTRYPOINTS_<NAME>_TSNET_HOSTNAME`:  
+Hostname for the node; if empty, defaults to the endpoint's name
```
Zolfax, line 343:
```suggestion
Hostname for the node. Be aware of Tailscale machine name conventions. (Default: entrypoint's ```<name>```)
```

pkg/config/static/entrypoints.go
```diff
@@ -163,6 +164,15 @@ func (u *UDPConfig) SetDefaults() {
 	u.Timeout = ptypes.Duration(DefaultUDPTimeout)
 }
 
+// TSNetConfig configures exposing an entry point over a Tailscale network (using tsnet).
+type TSNetConfig struct {
+	Hostname   string `description:"Hostname for the node; if empty, defaults to the endpoint's name" json:"hostname,omitempty" toml:"hostname,omitempty" yaml:"hostname,omitempty" export:"true"`
+	Ephemeral  bool   `description:"Register the node as ephemeral." json:"ephemeral,omitempty" toml:"ephemeral,omitempty" yaml:"ephemeral,omitempty" export:"true"`
+	AuthKey    string `description:"Auth key for the node (overrides the TS_AUTHKEY env var)." json:"authKey,omitempty" toml:"authKey,omitempty" yaml:"authKey,omitempty" export:"true"`
+	ControlURL string `description:"URL for the Tailscale control plane, when not using the default one." json:"controlUrl,omitempty" toml:"controlUrl,omitempty" yaml:"controlUrl,omitempty" export:"true"`
+	Dir        string `description:"Path where Tailscale stores its state. Set it to a persistent volume to allow Traefik to remain authenticated with Tailscale. Defaults to /var/lib/tailscale." json:"dir,omitempty" toml:"dir,omitempty" yaml:"dir,omitempty" export:"true"`
```
Zolfax, line 173: If any feedback regarding the documentation above gets accepted, these descriptions need to be adapted as well.

Zolfax (commented):
Optional: Some feedback regarding documentation.
Primarily to keep the consistency of the documentation.

svenvg93:
Any updates on this pr? Would love to use traefik to expose my services inside tailscale.

Zolfax:
I would love to see this in a traefik release.
I believe we are currently stuck on two things:
- PR is not updated to current main branch.
- No review from maintainers present.

ItalyPaleAle:
I would love some feedback from maintainers before updating this or making changes.

There's a lot of open comments right now and we could use some direction.

johntdyer:
oh boy... I so want this !!!

wsw70:
I know this not the proper way to signal interest (I already clicked thumbs ups, hearts and whatnot) but as @johntdyer said, oh boy... I so want this as well!! :)

offbyone:
How much work would go into adding OAuth support to this? Tailscale authkeys still can't have a lifetime longer than 90 days, which means that the key put in this configuration would require frequent rotation; OAuth allows for much more durable authentication.

ItalyPaleAle:
Not much to add OAuth2.

But I'm still waiting for the maintainers to make some key decisions...

PR #11753, opened 2025-05-13
fix: scope the rate limit counter key by source AND by middleware

aromeyer:
### What does this PR do?

Fix: adapt the rate limiter source key to also include the middleware name, ensuring that one middleware cannot affect another, regardless of the rate limiter storage.

### Motivation

While testing the new redis rate limiter solution coming in 3.4 (thank you, it's great, we've been waiting for it for a long time), it appears that one middleware configuration on a specific route was affecting another rate limit middleware on another route.

This was due to the fact that:

* both middlewares were using the same redis cluster
* the current rate limit "state by source" key is only based on the source name (`rate:10.172.0.3` for example), meaning two different ratelimit middlewares applied on two different routes will use the same "source state", one affecting the other. For a given source, both middlewares are updating the same Redis key.

Proposed solution: scope the native rate limit "counters" key per `middleware + source` allowing to have independent rate limit computations based only on the traffic handled by each middleware whatever the storage solution is.

For the in memory rate limiter, the "final" key is renamed from `<source>` to `<middleware name>:<source>`,
for the redis rate limiter, the redis "final" key is renamed from `rate:<source>` to `rate:<middleware name>:<source>`.

### DIFF
--- a/pkg/middlewares/ratelimiter/rate_limiter.go
+++ b/pkg/middlewares/ratelimiter/rate_limiter.go
@@ -146,7 +146,12 @@ func (rl *rateLimiter) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		logger.Info().Msgf("ignoring token bucket amount > 1: %d", amount)
 	}
 
-	delay, err := rl.limiter.Allow(ctx, source)
+	// Each rate limiter has its own source space,
+	// ensuring independence between rate limiters,
+	// i.e., rate limit rules are only applied based on traffic
+	// where the rate limiter is active.
+	rlSource := fmt.Sprintf("%s:%s", rl.name, source)
+	delay, err := rl.limiter.Allow(ctx, rlSource)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


lbenguigui:
Hey @aromeyer ,

Thanks for your interest in Traefik.

Since it's a fix, could you please rebase this PR on v3.4 branch?

aromeyer:
> Hey @aromeyer ,
> 
> Thanks for your interest in Traefik.
> 
> Since it's a fix, could you please rebase & change the target to v3.4 branch?

Yes sorry, it should now target the correct branch.

aromeyer:
> > Hey @aromeyer ,
> > Thanks for your interest in Traefik.
> > Since it's a fix, could you please rebase & change the target to v3.4 branch?
> 
Yes sorry, it should now target the correct branch.

lbenguigui (approved):
LGTM

rtribotte (approved):
Thanks!

PR #11762, opened 2025-05-16
fix: make IntN call safe

lbenguigui:
### What does this PR do?

This PR fixes the panic caused by concurrent calls to `Intn`.

 Creating a new `rand.Rand` with `rand.New(rand.NewSource(...))` makes calls to `Intn` unsafe (see https://pkg.go.dev/math/rand#NewSource). 

Adding a lock around the calls resolves this issue.

### Motivation

fixes: https://github.com/traefik/traefik/issues/11755

### More

- [ ] Added/updated tests


rtribotte (approved):
LGTM

PR #11768, opened 2025-05-19
Normalize request path

kevinpollet:
### What does this PR do?

This pull request normalizes the path before the routing step to ensure that the routing stays consistent when using unreserved and unallowed characters, percent-encoded.

So, before the routing step, the following steps are applied:

1. Normalization: decodes unreserved characters in the path and upper-cases percent-encoded characters, according to [RFC 3986 section 6.2.2.2](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.2) and [RFC 3986 section 6.2.2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.1).
 
2. Sanitization: removes dot dot and merges slashes from the path according to [RFC 3986 section 6.2.2.3](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.3)

3. Routing path computation: computes the routing path to inject it into the request context to be usable by the mux router. In the routing path, all characters are decoded but the reserved characters (which decoded, change the meaning of the URL).  

Please note that the path forwarded to the backend is the path after applying steps 1 and 2.

**Note:**
- Update the mux dependency once https://github.com/containous/mux/pull/11 is merged.
- Open a pull request on v3 to use the routing path from the context for routing.

### More

- [X] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/containous/alice"
+	"github.com/gorilla/mux"
 	"github.com/pires/go-proxyproto"
 	"github.com/sirupsen/logrus"
 	"github.com/traefik/traefik/v2/pkg/config/static"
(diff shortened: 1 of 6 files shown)


pkg/server/server_entrypoint_tcp.go
```diff
@@ -737,3 +742,156 @@ func sanitizePath(h http.Handler) http.Handler {
...
+	"%26": '&',
+	"%27": '\'',
+	"%28": '(',
+	"%29": ')',
+	"%2A": '*',
+	"%2B": '+',
+	"%2C": ',',
+	"%3B": ';',
+	"%3D": '=',
+	"%25": '%',
```
lbenguigui, line 849: Can you tell me why `% ` has been added in the reserved character list please?

juliens (approved):
LGTM

PR #11795, opened 2025-05-26
fix: Dashboard text h3 line height

harwck:
### What does this PR do?

Fix a very small issue where the line height is not enough and part of the port number text is being cutting out for the font in the `http/routers` page:

Before:
image

After:
image

### Motivation

Just wanna make the UI better :)

### More

- [ ] Added/updated tests


kevinpollet:
Hello @harwck and thanks for opening this pull request,

We tried to reproduce the issue bu everything is well displayed in our browsers.
How can we proceed to reproduce the issue?

We also merged https://github.com/traefik/traefik/pull/11674, which might fix your issue.
Can you try it?

Issue #11720, opened 2025-04-26
Show status rewrites of errors middleware in dashboard


sevensolutions:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

In #11520 I've implemented support for status code rewrites in the errors-middleware but unfortunately forgot about the dashboard.

It would be good if these rewrites are also be shown in the dashboard when inspecting the middleware.
I've already looked into the code and this change is pretty simple.
If you think this is useful, please tell me into which branch i should submit a PR.

Here is a screenshot what it looks like at the moment:

<img width="1800" alt="Image" src="https://github.com/user-attachments/assets/ed5215d6-ff1c-47a8-9e25-bb126d05ffd4" />

Here is what i've added:
```html
<!-- EXTRA FIELDS FROM MIDDLEWARES - [errors] - statusRewrites -->
<q-card-section v-if="middleware.errors">
  <div class="row items-start no-wrap">
    <div class="col">
      <div class="text-subtitle2">
        Status Rewrites
      </div>
      <q-chip
        v-for="(st, key) in exData(middleware).statusRewrites"
        :key="key"
        dense
        class="app-chip app-chip-green"
      >
        {{ key }} → {{ st }}
      </q-chip>
    </div>
  </div>
</q-card-section>
```

sevensolutions:
Closing this, because the PR is already merged.

PR #11806, opened 2025-05-29
Improve visualization for StatusRewrites option of errors middleware

sevensolutions:
### What does this PR do?

This PR improves the visualization of `statusRewrites` in the dashboard when using the `errors` middleware as suggested in #11720. 

### Motivation

Better visualization. In the original vue dashboard there were no visualization at all.
The new React dashboard only had a very generic one.

### More

- [ ] Added/updated tests

### Additional Notes

I don't know if this is the right approach to implement this because the component is very generic.
But there was already another special-case for `ipStrategy`.
Please tell me if I should implement it in another way.

Preview:
image


sevensolutions:
It seems that the test "TestIngressEndpointPublishedService" is a bit flaky. If i run it locally it's sometimes passing and sometimes not. Looks like the order is reversed sometimes.

kevinpollet (approved):
Thanks 👍

PR #11811, opened 2025-06-02
docs: Update supported versions

jnoordsij:
### What does this PR do?

- Adds the recently released [v3.4](https://github.com/traefik/traefik/releases/tag/v3.4.0) as currently supported v3 version
- Mark 2.11 as no longer actively supported given that Apr 29, 2025 has passed

Note that I'm not sure whether or not the v2 is indeed considered no longer actively supported just yet. If not, then probably a new date should be added instead. If it is, there might be other things that might have to be tweaked. For example, this PR template might be updated to reflect that only security fixes are accepted for v2, not 'regular' bug fixes.

### Motivation

To ensure documentation is up-to-date with releases.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/deprecation/releases.md
+++ b/docs/content/deprecation/releases.md
@@ -6,7 +6,8 @@ Below is a non-exhaustive list of versions and their maintenance status:
 
 | Version | Release Date | Active Support     | Security Support  |
 |---------|--------------|--------------------|-------------------|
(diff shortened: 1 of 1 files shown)


kevinpollet (approved):
Thanks 👍

nmengin (approved):
LGTM

PR #11812, opened 2025-06-02
Update Dockerfiles to Alpine 3.22

jnoordsij:
### What does this PR do?

Use the latest Alpine version (3.22; see also https://alpinelinux.org/posts/Alpine-3.22.0-released.html) for Dockerfiles.

Note: per #11811 I've targeted v3.4 rather than v2.11 as I assume it to be no longer supported. If however, for convenience or other reasons, it is still preferable to target v2.11 LMK and I'll adjust.

### Motivation

To be up-to-date.

### More

- [ ] Added/updated tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # syntax=docker/dockerfile:1.2
-FROM alpine:3.21
+FROM alpine:3.22
 
 RUN apk add --no-cache --no-progress ca-certificates tzdata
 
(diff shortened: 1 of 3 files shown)


rtribotte:
Hi @jnoordsij,

> Note: per https://github.com/traefik/traefik/pull/11811 I've targeted v3.4 rather than v2.11 as I assume it to be no longer supported. If however, for convenience or other reasons, it is still preferable to target v2.11 LMK and I'll adjust.

Indeed, it would be better to target v2.11 as we will maintain it for security fixes until Feb 01, 2026, thanks!

PR #11817, opened 2025-06-03
Handle context canceled in ForwardAuth middleware

bengentree:
### What does this PR do?

This PR handles context canceled errors from the forward auth request as client closed request HTTP response instead of the default internal server error response.

### Motivation

Fixes https://github.com/traefik/traefik/issues/11802

`Error calling http://xxx. Cause: Get "http://xxx": context canceled`

### More

- [x] Added/updated tests


rtribotte (approved):
Thanks!

Issue #11781, opened 2025-05-23
Show warning if mirroring percentage is set to 0


SkytAsul:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Setting up the mirroring of some requests is rather easy:
```yml
http:
  services:
    myservice-mirroring:
      mirroring:
        service: myservice
        mirrors:
        - mirror1
        - mirror2
```
This seems like a working configuration, but if you try it out, you'll notice that no traffic goes to the mirrors. This is because the `percent` option for the mirrors has not been set, and **it defaults to 0**!

In the [documentation](https://doc.traefik.io/traefik/routing/services/#mirroring-service), nothing is said about the default value of the `percent` option. One could assume that if you do not set it, that means you want all your traffic to be mirrored, not _none_.

I spent hours trying to debug my (much more complex) mirroring setup which was simply caused by this tiny misunderstanding...

Since changing the default value of the `percent` option from 0 to 100 is probably too risky as it can break existing setup relying on this default value, I suggest:
1. Writing a warning in the console in the case a mirror is set to have a percentage of 0.
2. Properly documenting that the default value is 0.

If someone _actually_ needs to mirror 0% of the requests, well they can simply remove the mirror from the list. I can't think of an instance where mirroring 0% is actually the desired behavior.

nmengin:
Hello @SkytAsul,

Thank you for the bug report.

We agreed that this behavior is unexpected.
As you mentioned, changing the value in v3 cannot be done because it leads to a breaking change.

We propose the following action plan:

1. Tag this issue a documentation bug
2. Fix the lack of documentation in v3
3. Once the documentation is fixed, re-open the issue as a bug to fix for the next major version.

Note that we would love community support to address it.
If you or another community member would like to build it, let us know, and we will work with you to make sure you have all the information needed so that it can be merged.

Alexy-vda:
👋 Hello! @nmengin I’m new to open source and would like to work on this. Is it okay if I take this on?

traefiker:
Closed by #11804.

SkytAsul:
Thanks for the quick documentation fix!

I created another issue (#11818) to change the default value in the next major version.

PR #11821, opened 2025-06-05
feat(mirroring): default percent to 100 when unset + unit test

### LINKED ISSUE #11818: Set default mirroring percentage to 100
### Welcome!

### What did you do?

Following up the discussion on #11781 

Create a `mirroring` service with some mirrors and do not set explicit `percent` values. One would expect that all traffic would be mirrored to the mirrors.

### What did you see instead?

No traffic is mirrored because the default value of the `mirror` option is 0.

In the next major version, the default value should be changed to 100.

3.3.6

```yaml
http:
  services:
    myservice-mirroring:
      mirroring:
        service: myservice
        mirrors:
        - mirror1
```

_No response_

Alexy-vda:
### What does this PR do?

* Introduces `MirroringServiceDefaultPercent` (100).
* **Breaking change :** `Percent` field is now `*int`; `SetDefaults()` sets it to 100 when `nil`.
* Adds unit tests covering: default (`nil`) / explicit `0` / explicit custom value.

### Motivation

Implements the behavior discussed in #11818 and originally raised in #11781: mirrors should receive traffic by default when `percent` is not specified.

Fixes #11818 

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/api/handler_http_test.go
+++ b/pkg/api/handler_http_test.go
@@ -467,15 +467,15 @@ func TestHandler_HTTP(t *testing.T) {
 								Mirrors: []dynamic.MirrorService{
 									{
 										Name:    "two@myprovider",
-										Percent: 10,
+										Percent: pointer(10),
 									},
 									{
 										Name:    "three@myprovider",
-										Percent: 15,
+										Percent: pointer(15),
 									},
 									{
 										Name:    "four@myprovider",
-										Percent: 80,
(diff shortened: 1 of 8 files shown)


nmengin:
Hey @Alexy-vda,

The modification makes sense.
However, it brings a breaking change.

That's why, even if we label it as "needs-review", we will wait for the next major version before merging it.

PR #11832, opened 2025-06-12
Additional reserved characters

borjaOrtizLlamas:
### What does this PR do?

Add the possibility of more reserved  encoded characters.  

### Motivation

With this pull request: https://github.com/traefik/traefik/pull/11790  its not possible to have dots as encoded characters in the path. For us is necessary to have them in some paths.  We need to add  more encoded characters and that is why I have added this new variable in the endpoint for this.  

### More
- [x] Added/updated tests


borjaOrtizLlamas:
close while i am fixing some tests,

 thanks,

Issue #10913, opened 2024-07-17
Errors on receving sigterm


msaitz:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I've been investigating some errors thrown by traefik pods (v2.11) when they happen to be terminated:

<img width="992" alt="Screenshot 2024-07-17 at 17 13 23" src="https://github.com/user-attachments/assets/390abe24-8502-43b8-8e67-1af56afe6a31">

I quickly arrived to the conclusion that the errors don't have any side effect, and that are independent of the configuration the traefik Deployment has, as I could get the same errors simply running traefik locally (in contaners):

config via cli args:
```
$ docker run -d \
  --name traefik \
  traefik:v2.11 \
  --api.insecure=true

# container logs while killing the container with sigterm via a "docker kill -s SIGTERM traefik"
$ docker logs traefik -f
time="2024-07-17T15:22:03Z" level=info msg="Configuration loaded from flags."
time="2024-07-17T15:22:07Z" level=error msg="accept tcp [::]:8080: use of closed network connection" entryPointName=traefik
time="2024-07-17T15:22:07Z" level=error msg="accept tcp [::]:80: use of closed network connection" entryPointName=http
time="2024-07-17T15:22:07Z" level=error msg="close tcp [::]:80: use of closed network connection" entryPointName=http
time="2024-07-17T15:22:07Z" level=error msg="close tcp [::]:8080: use of closed network connection" entryPointName=traefik
```

config via static config:
```
$ docker run -d \
  --name traefik \
  -v $(pwd)/traefik.yml:/etc/traefik/traefik.yml \
  traefik:v2.11

# container logs while killing the container with sigterm via a "docker kill -s SIGTERM traefik"
$ docker logs traefik -f
time="2024-07-17T15:33:05Z" level=info msg="Configuration loaded from file: /etc/traefik/traefik.yml"
time="2024-07-17T15:33:44Z" level=error msg="accept tcp [::]:80: use of closed network connection" entryPointName=web
time="2024-07-17T15:33:44Z" level=error msg="close tcp [::]:80: use of closed network connection" entryPointName=web
```

The `traefik.yml` content of the above being:
```
entryPoints:
  web:
    address: ":80"

api:
  dashboard: true
```

[...]

PR #11838, opened 2025-06-16
Errors on receving sigterm

Kwuray:
### What does this PR do?

This PR fix the bug where error are thrown on SIGTERM signal.

### Motivation

[Issue #10913](https://github.com/traefik/traefik/issues/10913)

### Additional Notes

Maybe it should be a good idea to add a test, to ensure that no error are throwns when entrypoints are gracefully stopped. But the error is not returned, just print via the Logger, so I don't how to do that.

### DIFF
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -324,14 +324,9 @@ func (e *TCPEntryPoint) Shutdown(ctx context.Context) {
 		go shutdownServer(e.httpServer.Server)
 	}
 
-	if e.httpsServer.Server != nil {
+	if e.http3Server != nil {
 		wg.Add(1)
-		go shutdownServer(e.httpsServer.Server)
-
-		if e.http3Server != nil {
-			wg.Add(1)
-			go shutdownServer(e.http3Server)
-		}
+		go shutdownServer(e.http3Server)
 	}
 
 	if e.tracker != nil {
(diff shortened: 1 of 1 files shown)


nmengin:
Hello @Kwuray,

Thank you for your contribution. We have labeled it as "need-review" to check.

> Maybe it should be a good idea to add a test

I agree, could you add this test?

Thanks in advance.

Kwuray:
Hello @nmengin,

I added the test, but I'm not sure I did it the right way.

It's difficult to test that no error are printed into stderr, so I added an error variable into the httpServer struct so we can test it. Not 100% convinced.

I would like to have your opinion on this. Thanks!

nmengin:
Hey @Kwuray.

Sorry for the delay.
Your PR is in still our (maintainers') backlog.

> I would like to have your opinion on this.

We are focused elsewhere, and even if we have planned to move forward soon, I have no ETA to provide yet.
We'll keep you updated as soon as we move forward on it.

rtribotte:
@Kwuray By looking at the changes we noticed a few things we would prefer to do differently, would you mind if for review we push commits directly on your branch?

Kwuray:
Hello @rtribotte, no problem !

Could you give me some feedback for improving myself please ? This is my first contribution. Thanks in advance :)

kevinpollet:
Hello @Kwuray and thanks for your contribution,

We updated the pull request by reinstating the HTTPS server shutdown, which was necessary. We also introduced a listener wrapper to ensure the listener closes only once. Finally, we refactored your work slightly and kept the idea of the shutdown boolean.

juliens (approved):
LGTM

kevinpollet (approved):
Thanks 👍

PR #11844, opened 2025-06-19
NGINX Ingress Provider

rtribotte:
### What does this PR do?

This PR introduces a new provider to support Ingresses with NGINX annotations.

### Motivation

To bootstrap and provide a solution for migrating from Ingresses with NGINX annotations.

### More

- [x] Added/updated tests

### Additional Notes

The PR is still missing some more work, notably on the documentation.

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/annotations.go
+++ b/pkg/provider/kubernetes/ingress-nginx/annotations.go
@@ -1,0 +1,112 @@
+package ingressnginx
+
+import (
+	"errors"
+	"reflect"
+	"strconv"
+	"strings"
+
+	netv1 "k8s.io/api/networking/v1"
+)
+
+type ingressConfig struct {
+	AuthType       *string `annotation:"nginx.ingress.kubernetes.io/auth-type"`
+	AuthSecret     *string `annotation:"nginx.ingress.kubernetes.io/auth-secret"`
+	AuthRealm      *string `annotation:"nginx.ingress.kubernetes.io/auth-realm"`
+	AuthSecretType *string `annotation:"nginx.ingress.kubernetes.io/auth-secret-type"`
+
+	AuthURL             *string `annotation:"nginx.ingress.kubernetes.io/auth-url"`
+	AuthResponseHeaders *string `annotation:"nginx.ingress.kubernetes.io/auth-response-headers"`
+
+	ForceSSLRedirect *bool `annotation:"nginx.ingress.kubernetes.io/force-ssl-redirect"`
+	SSLRedirect      *bool `annotation:"nginx.ingress.kubernetes.io/ssl-redirect"`
+
+	SSLPassthrough *bool `annotation:"nginx.ingress.kubernetes.io/ssl-passthrough"`
+
+	UseRegex *bool `annotation:"nginx.ingress.kubernetes.io/use-regex"`
+
+	Affinity              *string `annotation:"nginx.ingress.kubernetes.io/affinity"`
+	SessionCookieName     *string `annotation:"nginx.ingress.kubernetes.io/session-cookie-name"`
+	SessionCookieSecure   *bool   `annotation:"nginx.ingress.kubernetes.io/session-cookie-secure"`
+	SessionCookiePath     *string `annotation:"nginx.ingress.kubernetes.io/session-cookie-path"`
+	SessionCookieDomain   *string `annotation:"nginx.ingress.kubernetes.io/session-cookie-domain"`
+	SessionCookieSameSite *string `annotation:"nginx.ingress.kubernetes.io/session-cookie-samesite"`
+	SessionCookieMaxAge   *int    `annotation:"nginx.ingress.kubernetes.io/session-cookie-max-age"`
+
+	BackendProtocol *string `annotation:"nginx.ingress.kubernetes.io/backend-protocol"`
+
+	ProxySSLSecret     *string `annotation:"nginx.ingress.kubernetes.io/proxy-ssl-secret"`
+	ProxySSLVerify     *string `annotation:"nginx.ingress.kubernetes.io/proxy-ssl-verify"`
+	ProxySSLName       *string `annotation:"nginx.ingress.kubernetes.io/proxy-ssl-name"`
+	ProxySSLServerName *string `annotation:"nginx.ingress.kubernetes.io/proxy-ssl-server-name"`
+
+	EnableCORS                 *bool     `annotation:"nginx.ingress.kubernetes.io/enable-cors"`
+	EnableCORSAllowCredentials *bool     `annotation:"nginx.ingress.kubernetes.io/cors-allow-credentials"`
+	CORSExposeHeaders          *[]string `annotation:"nginx.ingress.kubernetes.io/cors-expose-headers"`
+	CORSAllowHeaders           *[]string `annotation:"nginx.ingress.kubernetes.io/cors-allow-headers"`
+	CORSAllowMethods           *[]string `annotation:"nginx.ingress.kubernetes.io/cors-allow-methods"`
+	CORSAllowOrigin            *[]string `annotation:"nginx.ingress.kubernetes.io/cors-allow-origin"`
+	CORSMaxAge                 *int      `annotation:"nginx.ingress.kubernetes.io/cors-max-age"`
+}
+
+// parseIngressConfig parses the annotations from an Ingress object into an ingressConfig struct.
+func parseIngressConfig(cfg *ingressConfig, ing *netv1.Ingress) error {
+	cfgType := reflect.TypeOf(*cfg)
+	cfgValue := reflect.ValueOf(cfg).Elem()
+
+	for i := range cfgType.NumField() {
+		field := cfgType.Field(i)
+		annotation := field.Tag.Get("annotation")
+		if annotation == "" {
+			continue
+		}
+
+		val, ok := ing.GetAnnotations()[annotation]
+		if !ok {
+			continue
+		}
+
+		switch field.Type.Elem().Kind() {
+		case reflect.String:
+			cfgValue.Field(i).Set(reflect.ValueOf(&val))
+		case reflect.Bool:
+			parsed, err := strconv.ParseBool(val)
+			if err == nil {
+				cfgValue.Field(i).Set(reflect.ValueOf(&parsed))
+			}
+		case reflect.Int:
+			parsed, err := strconv.Atoi(val)
+			if err == nil {
+				cfgValue.Field(i).Set(reflect.ValueOf(&parsed))
+			}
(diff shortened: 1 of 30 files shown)


pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -0,0 +1,1118 @@
...
+	for _, ingress := range ingresses {
+		logger := log.Ctx(ctx).With().Str("ingress", ingress.Name).Str("namespace", ingress.Namespace).Logger()
+		ctxIngress := logger.WithContext(ctx)
+
+		if !p.shouldProcessIngress(ingress, ingressClasses) {
+			continue
+		}
+
+		var ingressConfig ingressConfig
+		if err := parseIngressConfig(&ingressConfig, ingress); err != nil {
```
juliens: Why this take ingressConfig as first param (always with an empty ingressConfig) instead of taking only ingress and create a new ingressConfig?

pkg/provider/kubernetes/ingress-nginx/client.go
```diff
@@ -0,0 +1,385 @@
...
+
+	return true
+}
+
+// filterIngressClass return a slice containing IngressClass matching either the annotation name or the controller.
+func filterIngressClass(ingressClasses []*netv1.IngressClass, ingressClassByName bool, ingressClass, controllerClass string) []*netv1.IngressClass {
+	var filteredIngressClasses []*netv1.IngressClass
+	for _, ic := range ingressClasses {
+		if ingressClassByName && ic.Name == ingressClass {
+			filteredIngressClasses = append(filteredIngressClasses, ic)
```
lbenguigui: I believe we can return instead of continue

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -0,0 +1,1118 @@
...
+
+	IngressClass             string `description:"Name of the ingress class this controller satisfies." json:"ingressClass,omitempty" toml:"ingressClass,omitempty" yaml:"ingressClass,omitempty" export:"true"`
+	ControllerClass          string `description:"Ingress Class Controller value this controller satisfies." json:"controllerClass,omitempty" toml:"controllerClass,omitempty" yaml:"controllerClass,omitempty" export:"true"`
+	WatchIngressWithoutClass bool   `description:"Define if Ingress Controller should also watch for Ingresses without an IngressClass or the annotation specified." json:"watchIngressWithoutClass,omitempty" toml:"watchIngressWithoutClass,omitempty" yaml:"watchIngressWithoutClass,omitempty" export:"true"`
+	IngressClassByName       bool   `description:"Define if Ingress Controller should watch for Ingress Class by Name together with Controller Class." json:"ingressClassByName,omitempty" toml:"ingressClassByName,omitempty" yaml:"ingressClassByName,omitempty" export:"true"`
```
lbenguigui, line 74: Can you explain why we need this option? Isn’t the `IngressClass` resource sufficient?

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -0,0 +1,1118 @@
...
+			}
+
+			if rule.HTTP == nil {
+				continue
+			}
+
+			for pi, pa := range rule.HTTP.Paths {
+				// As NGINX we are ignoring resource backend.
+				// An Ingress backend must have se service or a resource definition.
+				if pa.Backend.Service == nil {
```
juliens, line 444: Missing logs when no service or we use resource

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -0,0 +1,1118 @@
...
+			Priority:   math.MinInt32,
+			Service:    defaultBackendName,
+			TLS:        &dynamic.RouterTLSConfig{},
+		}
+
+		conf.HTTP.Services[defaultBackendName] = svc
+	}
+
+	var ingressClasses []*netv1.IngressClass
+	if !p.k8sClient.IngressClassesIgnored() {
```
lbenguigui: Instead of exporting and using `IngressClassesIgnored`, I’d prefer using a `ListIngressClasses` that internally checks `ingressClassesIgnored` and returns an empty slice when appropriate.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -0,0 +1,1118 @@
...
+				ingressStatus = append(ingressStatus, netv1.IngressLoadBalancerIngress{IP: ip})
+			}
+		}
+	}
+
+	return p.k8sClient.UpdateIngressStatus(ing, ingressStatus)
+}
+
+func (p *Provider) shouldProcessIngress(ingress *netv1.Ingress, ingressClasses []*netv1.IngressClass) bool {
+	if !p.k8sClient.IngressClassesIgnored() && ingress.Spec.IngressClassName != nil {
```
lbenguigui: I feel `IngressClassesIgnored` call is not mandatory since if it's true `ingressClasses` will be empty

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -0,0 +1,1118 @@
...
+	}
+
+	if class, ok := ingress.Annotations[annotationIngressClass]; ok {
+		return class == p.IngressClass
+	}
+
+	return p.WatchIngressWithoutClass
+}
+
+func (p *Provider) loadCertificates(ctx context.Context, ingress *netv1.Ingress, uniqCerts map[string]*tls.CertAndStores) error {
```
lbenguigui, line 761: This comment could apply to many other parts of the provider: why did we choose not to reuse the Kubernetes ingress provider logic?

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -0,0 +1,1118 @@
...
+	WatchNamespaceSelector string `description:"Selector selects namespaces the controller watches for updates to Kubernetes objects." json:"watchNamespaceSelector,omitempty" toml:"watchNamespaceSelector,omitempty" yaml:"watchNamespaceSelector,omitempty" export:"true"`
+
+	IngressClass             string `description:"Name of the ingress class this controller satisfies." json:"ingressClass,omitempty" toml:"ingressClass,omitempty" yaml:"ingressClass,omitempty" export:"true"`
+	ControllerClass          string `description:"Ingress Class Controller value this controller satisfies." json:"controllerClass,omitempty" toml:"controllerClass,omitempty" yaml:"controllerClass,omitempty" export:"true"`
+	WatchIngressWithoutClass bool   `description:"Define if Ingress Controller should also watch for Ingresses without an IngressClass or the annotation specified." json:"watchIngressWithoutClass,omitempty" toml:"watchIngressWithoutClass,omitempty" yaml:"watchIngressWithoutClass,omitempty" export:"true"`
```
lbenguigui, line 73: I didn’t see any handling of `ingressclass.kubernetes.io/is-default-class`. To me, this is the standard way to define the default ingress controller, instead of define a bool to know if we need to handle ingress with empty ingressClass.

pkg/provider/kubernetes/ingress-nginx/annotations.go
```diff
@@ -0,0 +1,114 @@
...
+	cfgValue := reflect.ValueOf(cfg).Elem()
+
+	for i := range cfgType.NumField() {
+		field := cfgType.Field(i)
+		annotation := field.Tag.Get("annotation")
+		if annotation == "" {
+			continue
+		}
+
+		val, ok := ing.GetAnnotations()[annotation]
```
lbenguigui, line 67: nitpick: Are we sure `annotations` map is always maked?

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -0,0 +1,1118 @@
...
+			if err != nil {
+				logger.Error().
+					Str("serviceName", ingress.Spec.DefaultBackend.Service.Name).
+					Str("servicePort", ingress.Spec.DefaultBackend.Service.Port.String()).
+					Err(err).
+					Msg("Cannot create default backend service")
+			}
+		}
+
+		if len(ingress.Spec.Rules) == 0 {
```
lbenguigui: Unless I'm mistaken, when no rules are defined, defaultBackend must be set. I think we should handle this in the previous `if` block.

lbenguigui (changes requested):

juliens (changes requested):

pkg/provider/kubernetes/ingress-nginx/client.go
rtribotte: I see, updated.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
rtribotte: If we are not mistaken, this is how the Kubernetes Ingress NGINX controller works.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
rtribotte: If we are not mistaken, we expose the same options as the Kubernetes Ingress NGINX controller.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
rtribotte: As we discussed, this could be done in a second iteration.

pkg/provider/kubernetes/ingress-nginx/annotations.go
rtribotte: No need to make sure of this for read operations.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
rtribotte: Done!

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
rtribotte: Done!

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
rtribotte: Done!

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
rtribotte: Done!

juliens (approved):
LGTM

lbenguigui (approved):
LGTM

PR #11847, opened 2025-06-20
Respect service.nativelb=false annotation when nativeLBByDefault is enabled

sdelicata:
### What does this PR do?

This PR fixes an issue with the Kubernetes Gateway API provider where the `traefik.io/service.nativelb: "false"` annotation was not respected when the global `nativeLBByDefault` option was enabled.

### Motivation
Fixes https://github.com/traefik/traefik/issues/11823

### More

- [x] Added/updated tests


pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -918,7 +918,9 @@ func (p *Provider) getBackendAddresses(namespace string, ref gatev1.BackendRef)
 		return nil, corev1.ServicePort{}, fmt.Errorf("parsing service annotations config: %w", err)
 	}
 
-	if p.NativeLBByDefault || annotationsConfig.Service.NativeLB {
+	// Use the service.nativelb annotation value if explicitly set. If not, use the provider's default value.
+	_, hasNativeLBAnnotation := service.Annotations["traefik.io/service.nativelb"]
```
lbenguigui: Since `parseServiceAnnotations` is only used to get `nativelb`value, WDYT removing the call to `parseServiceAnnotation`and using the value of the annotation?

lbenguigui (changes requested):

Issue #10187, opened 2023-10-24
Support for Encrypted Client Hello (ECH, prev. ESNI)


muebau:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

As Chrome and Firefox support ECH (was ESNI) now it would be nice to get this feature in Traefik too.
I was not able to find an issue yet so I opened this one.

Information:
Chrome 117:
https://chromestatus.com/feature/6196703843581952

Firefox 118:
https://support.mozilla.org/pl/kb/understand-encrypted-client-hello

junepark678:
I feel like golang/go#68500 is more related as the issue you linked is for *client* rather than server.

ThomasCr:
The related issue is also about the Server side.
The patch is currently Work in Progress

https://go-review.googlesource.com/c/go/+/623576

savely-krasovsky:
I guess only after proper Go 1.24 release.

rtribotte:
Hello,

We marked this issue as `contributor/wanted` as this would not make it to our roadmap for a while as we are focused elsewhere. 

Once the bump to Go 1.24 will be effective, if you or another community member would like to build it, let us know, and we will work with you to make sure you have all the information needed so that it can be merged.

We prefer to work with our community members at the beginning of the design process so that we can make sure that we are aligned and can move quickly with the review and merge process. Let us know here or create a PR before you start, and we will work with you there.

Don’t forget to check out the [contributor docs](https://github.com/traefik/contributors-guide/blob/master/pr_guidelines.md) and to link the PR to this issue.
We marked this issue as `contributor/wanted` as

niffelheimba:
Just wanted to document this for anyone encountering `ERR_ECH_FALLBACK_CERTIFICATE_INVALID` or `QUIC` issues, particularly for local applications for which you have both a local DNS service and Cloudflare DNS records.

You can disable ECH on cloudflare via this guide [ERR_ECH_FALLBACK_CERTIFICATE_INVALID](https://neonode.cc/en/blog/how_to_disable_ech_cf/) and it should resolve these issues. It is a better solution than having to disable TLS1.3 entirely.

georglauterbach:
[...]

PR #11866, opened 2025-06-27
Support Encrypted Client Hello (ECH)

hasbai:
### What does this PR do?

This PR adds support for server-side Encrypted Client Hello(ECH).

### Motivation

Discussed in #10187 

### More

- [x] Added/updated tests


pkg/tls/ech.go
```diff
@@ -0,0 +1,244 @@
...
+	if s, ok := any(c.ECH).(string); ok {
+		ech, err = base64.StdEncoding.DecodeString(s)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		ech = []byte(c.ECH)
+	}
+
+	requestURL, _ := url.Parse(c.URL)
```
sdelicata: Please don't ignore the potential error here.

pkg/tls/ech.go
```diff
@@ -0,0 +1,244 @@
...
+		Host:   c.Host,
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	body = make([]byte, 1024)
+	n, _ := resp.Body.Read(body)
```
sdelicata: Same here

pkg/tls/ech.go
```diff
@@ -0,0 +1,244 @@
...
+			k.Config = block.Bytes[2:] // Skip the first two bytes (length prefix)
+		default:
+			return nil, fmt.Errorf("unknown PEM block %s", block.Type)
+		}
+
+		data = rest
+	}
+
+	if len(k.Config) == 0 || len(k.PrivateKey) == 0 {
+		return nil, fmt.Errorf("lack of ECH configuration or private key in PEM file")
```
sdelicata:
```suggestion
		return nil, errors.New("missing ECH configuration or private key in PEM file")
```

pkg/tls/ech.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	k.SendAsRetry = true
+
+	return &k, nil
+}
+
+func MarshalECHKey(k *tls.EncryptedClientHelloKey) ([]byte, error) {
+	if len(k.Config) == 0 || len(k.PrivateKey) == 0 {
+		return nil, fmt.Errorf("lack of ECH configuration or private key")
```
sdelicata:
```suggestion
		return nil, errors.New("missing ECH configuration or private key")
```

pkg/tls/ech.go
```diff
@@ -0,0 +1,244 @@
...
+
+	return &tls.EncryptedClientHelloKey{
+		Config:      configBytes,
+		PrivateKey:  privateKeyBytes,
+		SendAsRetry: true,
+	}, nil
+}
+
+// startECHServer starts a TLS server that supports Encrypted Client Hello (ECH).
+func startECHServer(bind string, cert tls.Certificate, echKey tls.EncryptedClientHelloKey) {
```
sdelicata: This function seems to be only used for testing. Please consider moving it to a test file.

cmd/ech/ech.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+		Run: func(args []string) error {
+			c := tls.ECHRequestConf[string]{}
+			fs := flag.NewFlagSet("ech request", flag.ContinueOnError)
+			c.URL = fs.Arg(0)
+			c.ECH = *fs.String("ech", "", "A base64-encoded ECH public config list.")
+			c.Host = *fs.String("h", "", "The host to use in the request. If not set, it will be derived from the URL.")
+			c.Insecure = *fs.Bool("k", false, "Allow insecure server connections when using SSL.")
+			fs.SetOutput(os.Stderr)
+			if err := fs.Parse(args); err != nil {
```
sdelicata: Maybe I'm wrong, but shouldn't the `fs.Parse(args)` function be called before assignment?

pkg/tls/ech.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return nil, err
+	}
+	privateKeyBytes, err := privateKey.MarshalBinary()
+	if err != nil {
+		return nil, err
+	}
+
+	config := echConfig{
+		Version:   0xfe0d, // ECH version 0xfe0d
```
sdelicata, line 107: Maybe this magic number should be defined as a constant. WDYT?

cmd/ech/ech.go
```diff
@@ -0,0 +1,111 @@
...
+	"flag"
+	"github.com/pkg/errors"
+	"github.com/traefik/paerser/cli"
+	"github.com/traefik/traefik/v3/pkg/tls"
+	"io"
+	stdlog "log"
+	"os"
+)
+
+// NewCmd builds a new Version command.
```
sdelicata:
```suggestion
// NewCmd builds a new ECH command.
```

cmd/ech/ech.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+
+	return &cmd
+}
+
+func generate() *cli.Command {
+	help := []byte(`Usage: ech generate SNI [SNI ...]`)
+	cmd := cli.Command{
+		Name:        "generate",
+		Description: "Generate a new ECH key with given outer sni.",
```
sdelicata:
```suggestion
		Description: "Generate a new ECH key with given outer SNI.",
```

pkg/tls/ech.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	if err := server.ListenAndServeTLS("", ""); err != nil {
+		log.Fatalf("server error: %v", err)
+	}
+}
+
+type ECHRequestConf[T []byte | string] struct {
+	URL      string `description:"The URL to request." json:"u" export:"true"`
+	Host     string `description:"The host/sni to request with." json:"h" export:"true"`
```
sdelicata:
```suggestion
	Host     string `description:"The host/sni to request." json:"h" export:"true"`
```

pkg/tls/ech.go
```diff
@@ -0,0 +1,244 @@
...
+			EncryptedClientHelloKeys: []tls.EncryptedClientHelloKey{echKey},
+		},
+	}
+
+	if err := server.ListenAndServeTLS("", ""); err != nil {
+		log.Fatalf("server error: %v", err)
+	}
+}
+
+type ECHRequestConf[T []byte | string] struct {
```
sdelicata: For the sake of consistency, could you rename it to `ECHRequestConfig`, please?

pkg/tls/ech.go
```diff
@@ -0,0 +1,244 @@
...
+	URL      string `description:"The URL to request." json:"u" export:"true"`
+	Host     string `description:"The host/sni to request with." json:"h" export:"true"`
+	ECH      T      `description:"A base64-encoded ECH public config list." json:"ech" export:"true"`
```
sdelicata:
```suggestion
	ECH      T      `description:"Base64-encoded ECH public configuration list for client use." json:"ech" export:"true"`
```

sdelicata (changes requested):
Hi @hasbai and thx for your contribution!
I made a few comments on your PR.

cmd/ech/ech.go
sdelicata, line 1: Your command is useful for testing purposes, but I wonder if Traefik is the right place to add such a command.

savely-krasovsky:
@hasbai do you need any help maybe? I am interested in merging it.

kevinpollet:
Hello @hasbai and thanks for your contribution,

Are you still working on the subject?
Would you mind if we push commits during the review process?

georglauterbach:
@hasbai any updates on this?

georglauterbach:
> Hello @hasbai and thanks for your contribution,
> 
> Are you still working on the subject? Would you mind if we push commits during the review process?

I can provide this MR again and make it available for changes (pushing commits) if this helps and in case the OP does not respond. Would this be of any assistance? I cannot help with programming itself, as I do not read nor write Go.

kevinpollet:
Hello @georglauterbach and thanks for proposing your help,

We started reviewing this pull request, but we are focused elsewhere right now.
However, we will push review commits as soon as possible.

mmatur (approved):
LGTM

Thanks @hasbai

pkg/tls/ech.go
```diff
@@ -0,0 +1,165 @@
...
+	if len(k.Config) == 0 || len(k.PrivateKey) == 0 {
+		return nil, errors.New("missing ECH configuration or private key in PEM file")
+	}
+
+	// go ecdh now only supports SHA-256 (32-byte private key)
+	if len(k.PrivateKey) < sha256PrivateKeyLength {
+		return nil, fmt.Errorf("invalid private key length: expected at least %d bytes, got %d bytes", sha256PrivateKeyLength, len(k.PrivateKey))
+	} else if len(k.PrivateKey) > sha256PrivateKeyLength {
+		k.PrivateKey = k.PrivateKey[len(k.PrivateKey)-sha256PrivateKeyLength:]
+	}
```
sdelicata, line 47: I'm not comfortable with silently truncating cryptographic material.
Why not return an error if the key is too long?
If the truncation is intentional, please explain it in a comment and add a warning log.

pkg/tls/ech.go
```diff
@@ -0,0 +1,165 @@
...
+		block, rest := pem.Decode(data)
+		if block == nil {
+			break
+		}
+
+		switch block.Type {
+		case "PRIVATE KEY":
+			k.PrivateKey = block.Bytes
+		case "ECHCONFIG":
+			k.Config = block.Bytes[2:] // Skip the first two bytes (length prefix)
```
sdelicata, line 30: I wonder if we should be more cautious here, by adding some validations.
Maybe block.Bytes is shorter than 2 bytes.
We should also check that the length prefix matches the actual content.
WDYT?

pkg/tls/ech.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+type echExtension struct {
+	Type uint16
+	Data []byte
+}
+
+type echConfig struct {
+	Version uint16
+	Length  uint16
```
sdelicata, line 80: If I'm not wrong, this property is never used.

sdelicata (changes requested):

savely-krasovsky:
Will it end up in 3.7.x? Maybe I can help somehow?

Linux-bootloader:
Is this PR still active?

nmengin:
Hey @Linux-bootloader ,

Linux-bootloader:
> Hey @Linux-bootloader ,
> 
> Thanks for reaching out! This is definitely on our radar.
> 
> We're focused on other priorities right now though, and we don't have an ETA yet—even though we plan to get to it.
> 
> We'll keep you updated as soon as we can move forward!

Thank you for clarifying 😇

savely-krasovsky:
Can you finally find some capacity and review it? AI tools will help. C'mon, this is an important feature, and exactly because of projects like Traefik we cannot gain broader adoption! Countries around the world deploy censorship, and things like ECH help fight it, at least in mild cases. We already skipped 3.7 release window for no reason, I doubt you will include it into minor release.

savely-krasovsky:
Sorry for the rant, but this has been hanging in limbo for a year already-not as an issue, but as a prepared MR.

emilevauge:
@savely-krasovsky I have another idea: could you help us fixing the last review, rebasing the code and submiting a new PR? We will merge it much faster. thanks for your contribution to the project 🙏

savely-krasovsky:
I like this idea, will try my best during this week.

savely-krasovsky:
~~I decided to give it a try today, but I have a question first.~~

~~Right now ECH connections get their TLS options from the outer SNI on TCP, but from the decrypted inner SNI on HTTP/3. The TCP path selects the config before the handshake, based on the peeked SNI, and `crypto/tls` reveals the inner name only later (ECH is processed before `GetConfigForClient`). So a protected domain with a non-default TLS option (e.g. mTLS) gets a 421 from snicheck over TCP, while the same request works over h3.~~

~~I want to align TCP with h3: add `GetConfigForClient` to the pinned config (reusing `HTTP3TLSConfigMatcherFunc`) and make snicheck take the used option from `req.TLS.ServerName`. Routing stays outer-based, because passthrough can't decrypt ECH.~~

~~I'd like to have this in the ECH MR, but it touches the TLS termination pipeline, so as far as I understand it needs a dedicated PR. Any thoughts?~~

I believe the fix ended up pretty small, and hopefully it can be merged in one go.

mmatur:
Close in favor of https://github.com/traefik/traefik/pull/13758

PR #11872, opened 2025-07-02
Improve integration tests

mmatur:
### What does this PR do?

This PR avoids to build traefik on each integration tests jobs

### Motivation

Faster CI

### More

- ~[ ] Added/updated tests~


rtribotte:
@mmatur I'm not sure why but now that the binary build time is 0s because the binary is downloaded from the artifact, the `generate go test slice` step now takes 3m43s instead of 30s before, and there is no improvement regarding the complete job time.

rtribotte (approved):
LGTM

Issue #11829, opened 2025-06-10
`{url}` placeholder in customErrors middleware is only writing the `/path` portion of the requested URL during substitution (but full URL expected)


geudrik:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### tl;dr Jun 16 2025
The CustomErrors middleware `{url}` placeholder in the `query:` spec, does _not_ replace with the full request URL. It only replaces with the request path (the scheme and hostname portions are missing). 

### What did you do?

```yaml
apiVersion: traefik.io/v1alpha1
kind: Middleware
metadata:
  name: forward-auth-catch-auth-error
spec:
  errors:
    status:
      - "401-403"
    statusRewrites:
      "401": 302
      "403": 302
    query: "/oauth2/start?rd={url}"
    service:
      name: oauth2-proxy
      namespace: traefik
      port: 80
```

This results in calls being made to only the path portion of the original requested URL. 

For example, `curl https://whoami.domain.com/api` -> `"/oauth2/start?rd=%2Fapi"` In this case, the substituted placeholder `{url}` is being replaced with `%2Fapi`, instead of the expected `https%3a%2f%2fwhoami.domain.com%2fapi`

Either the placeholder needs renaming to `{path}` and the docs updated accordingly (if the functionality is as intended), or the logic needs to be fixed to actually URL encode the entire URL being requested (what I hope is the intent here)

Functionality was introduced by @tomMoulard at https://github.com/traefik/traefik/commit/ae6e84414360a8ac2aac488349cdbceaf2bf12b2

### What did you see instead?

Only the `path` is being URL encoded, not the `url`

### What version of Traefik are you using?

3.4.1

```
traefik-b6bc779d4-g82fl traefik {"level":"info","version":"3.4.1","time":"2025-06-10T16:45:47Z","message":"Traefik version 3.4.1 built on 2025-05-27T12:45:16Z"}
```

### What is your environment & configuration?

k8's

```yaml
apiVersion: traefik.io/v1alpha1
kind: Middleware
metadata:
  name: forward-auth-catch-auth-error
spec:
  errors:
    status:
      - "401-403"
    statusRewrites:
      "401": 302
      "403": 302
    query: "/oauth2/start?rd={url}"
    service:
      name: oauth2-proxy
      namespace: traefik
[...]

PR #11876, opened 2025-07-04
Fix customerrors query url replacement

DorianBlues:
What does this PR do?
This PR addresses an issue with the {url} placeholder in the customErrors middleware where only the request path was being replaced, leaving out the scheme and hostname portions.

Motivation
Fixes https://github.com/traefik/traefik/issues/11829

More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/customerrors/custom_errors.go
+++ b/pkg/middlewares/customerrors/custom_errors.go
@@ -122,11 +122,18 @@ func (c *customErrors) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	}
 
 	var query string
+
+	scheme := "http"
+	if req.TLS != nil {
+		scheme = "https"
+	}
+	originalURL := scheme + "://" + req.Host + req.URL.RequestURI()
+
 	if len(c.backendQuery) > 0 {
 		query = "/" + strings.TrimPrefix(c.backendQuery, "/")
 		query = strings.ReplaceAll(query, "{status}", strconv.Itoa(code))
 		query = strings.ReplaceAll(query, "{originalStatus}", strconv.Itoa(originalCode))
-		query = strings.ReplaceAll(query, "{url}", url.QueryEscape(req.URL.String()))
+		query = strings.ReplaceAll(query, "{url}", url.QueryEscape(originalURL))
 	}
 
 	pageReq, err := newRequest("http://" + req.Host + query)
--- a/pkg/middlewares/customerrors/custom_errors.go
+++ b/pkg/middlewares/customerrors/custom_errors.go
@@ -127,13 +127,13 @@ func (c *customErrors) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	if req.TLS != nil {
 		scheme = "https"
 	}
-	originalURL := scheme + "://" + req.Host + req.URL.RequestURI()
+	orig := &url.URL{Scheme: scheme, Host: req.Host, Path: req.URL.Path, RawPath: req.URL.RawPath, RawQuery: req.URL.RawQuery, Fragment: req.URL.Fragment}
 
 	if len(c.backendQuery) > 0 {
 		query = "/" + strings.TrimPrefix(c.backendQuery, "/")
 		query = strings.ReplaceAll(query, "{status}", strconv.Itoa(code))
 		query = strings.ReplaceAll(query, "{originalStatus}", strconv.Itoa(originalCode))
-		query = strings.ReplaceAll(query, "{url}", url.QueryEscape(originalURL))
+		query = strings.ReplaceAll(query, "{url}", url.QueryEscape(orig.String()))
 	}
 
 	pageReq, err := newRequest("http://" + req.Host + query)
--- a/pkg/middlewares/customerrors/custom_errors_test.go
+++ b/pkg/middlewares/customerrors/custom_errors_test.go
@@ -175,6 +175,10 @@ func TestHandler(t *testing.T) {
 
 			req := testhelpers.MustNewRequest(http.MethodGet, "http://localhost/test?foo=bar&baz=buz", nil)
 
+			// Client like browser and curl will issue a relative HTTP request, which not have a host and scheme in the URL. But the http.NewRequest will set them automatically.
+			req.URL.Host = ""
+			req.URL.Scheme = ""
+
 			recorder := httptest.NewRecorder()
 			errorPageHandler.ServeHTTP(recorder, req)
 


geudrik:
This MR does not resolve the issue for the reasons outlined here: https://github.com/traefik/traefik/issues/11829#issuecomment-2977393630

DorianBlues:
> This MR does not resolve the issue for the reasons outlined here: [#11829 (comment)](https://github.com/traefik/traefik/issues/11829#issuecomment-2977393630)

@geudrik I think there is no problem here,  the req.Host here is NOT set via the service: block in the middleware spec, it is the client's original request host. The original request is forwarded to the service defined by the middleware spec through the c.backendHandler.ServeHTTP method.

pkg/middlewares/customerrors/custom_errors.go
```diff
@@ -123,11 +123,18 @@ func (c *customErrors) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	}
 
 	var query string
+
+	scheme := "http"
+	if req.TLS != nil {
+		scheme = "https"
+	}
+	originalURL := scheme + "://" + req.Host + req.URL.RequestURI()
```
geudrik: `req.Host` is **not** the HTTP `Host:` requested by the originating request in this case, as clearly seen via [line 140](https://github.com/traefik/traefik/pull/11876/commits/36eecc25784eada7b96d96680bea8baa6778b714#diff-7f1f4cc79f1369ceb5228465b95bb882b8bbfde24d5ab6b34b77dca6e5ef759cR140). It is the host defined in the `service:` block of the yaml spec (config) for this middleware (ref: [this comment](https://github.com/traefik/traefik/issues/11829#issuecomment-2977393630))

Unfortunately, this isn't a quick-fix for an AI chatbot (I tried that)

geudrik: For clarification, line 140 here is the actual request by the middleware to fetch the "error" page to return to the user. It's referencing `req.Host` to pull the value it needs to fetch from `Middleware.errors.service`

In this case (and as [outlined in the bug report](https://github.com/traefik/traefik/issues/11829)), we're hijacking the fetch to instead pull the start of login flow from an auth proxy, and combining the the 4xx from the forward-auth "is authenticated" check (munging it to be a 3xx) with the body of the request to start login flow

geudrik:
> > This MR does not resolve the issue for the reasons outlined here: [#11829 (comment)](https://github.com/traefik/traefik/issues/11829#issuecomment-2977393630)
> 
> @geudrik I think there is no problem here, the req.Host here is NOT set via the service: block in the middleware spec, it is the client's original request host. The original request is forwarded to the service defined by the middleware spec through the c.backendHandler.ServeHTTP method.

@DorianBlues see my comment on your commit.

Revolyssup:
@geudrik You said
> req.Host is not the HTTP Host: requested by the originating request 

Below is a minimal reproduction of your scenario with this fix where I can confirm that the req.Host is actually the one from originating request.


When I send the request 
```bash
 curl  -H "Host: whoami.domain.com" -i localhost:8081/headers
HTTP/1.1 302 Found
Content-Type: text/plain
Date: Tue, 29 Jul 2025 17:37:43 GMT
Server: BaseHTTP/0.6 Python/3.13.3
Transfer-Encoding: chunked

Redirect handled for /start?rd=http%3A%2F%2Fwhoami.domain.com%2Fheaders
```

The result from the mock server which just states the URI with which it was accessed has `whoami.domain.com` instead of localhost. The original Host header is being used.



```yaml
http:
  routers:
    test-router:
      rule: "PathPrefix(`/headers`)"
      service: test-service
      entryPoints: ["web"]
      middlewares:
        - auth-chain

  services:
    test-service:
      loadBalancer:
        servers:
          - url: "http://httpbin.org"

    error-service:
      loadBalancer:
        servers:
          - url: "http://localhost:9000"  # Returns "Redirect handled for <request-URI>"

  middlewares:
    auth-chain:
      chain:
        middlewares:
          - catch-errors
          - forward-auth

    forward-auth:
      forwardAuth:
        address: "http://httpbin.org/status/401"  # Returns 401
        trustForwardHeader: true

    catch-errors:
      errors:
        status: [401, 403]
        statusRewrites:
          "401": 302
          "403": 302
        query: "/start?rd={url}"  # Problematic placeholder
        service: error-service
```

geudrik:
> The result from the mock server which just states the URI with which it was accessed has `whoami.domain.com` instead of localhost. The original Host header is being used.

I have no way to test this easily, but can you explain how this works? Where I'm getting tripped up is the actual code logic (as outlined in my above comments) - [line 140](https://github.com/traefik/traefik/pull/11876/commits/36eecc25784eada7b96d96680bea8baa6778b714#diff-7f1f4cc79f1369ceb5228465b95bb882b8bbfde24d5ab6b34b77dca6e5ef759cR140) seems to suggest that `req.Host` is the upstream error call host, no? I'm not a Go coder, so perhaps this is just me being a dunce?

Revolyssup:
> line 140

In line 140, request is sent to req.Host. This request object in middleware is the one from downstream client request. So obviously the Host here will be from client. I am failing to see where you're pointing at that req.Host is upstream Host.

geudrik:
> > line 140
> 
> In line 140, request is sent to req.Host. This request object in middleware is the one from downstream client request. So obviously the Host here will be from client. I am failing to see where you're pointing at that req.Host is upstream Host.

I guess I just wasn't understanding part of the codebase (and confusing the symptom of the original bug report).

If @DorianBlues 's MR works, hell yah!

pkg/middlewares/customerrors/custom_errors.go
```diff
@@ -123,11 +123,18 @@ func (c *customErrors) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
...
+
 	if len(c.backendQuery) > 0 {
 		query = "/" + strings.TrimPrefix(c.backendQuery, "/")
 		query = strings.ReplaceAll(query, "{status}", strconv.Itoa(code))
 		query = strings.ReplaceAll(query, "{originalStatus}", strconv.Itoa(originalCode))
-		query = strings.ReplaceAll(query, "{url}", url.QueryEscape(req.URL.String()))
+		query = strings.ReplaceAll(query, "{url}", url.QueryEscape(originalURL))
```
mmatur:
```suggestion
    // Build the original URL in a robust way using net/url to avoid manual string concatenation.
	scheme := "http"
	if req.TLS != nil {
		scheme = "https"
	}
	orig := &url.URL{Scheme: scheme, Host: req.Host, Path: req.URL.Path, RawPath: req.URL.RawPath, RawQuery: req.URL.RawQuery, Fragment: req.URL.Fragment}

	if len(c.backendQuery) > 0 {
		query = "/" + strings.TrimPrefix(c.backendQuery, "/")
		query = strings.ReplaceAll(query, "{status}", strconv.Itoa(code))
		query = strings.ReplaceAll(query, "{originalStatus}", strconv.Itoa(originalCode))
		query = strings.ReplaceAll(query, "{url}", url.QueryEscape(orig.String()))
	}
```

DorianBlues: Hi, I've rebased onto the v3.5 branch and optimized the code.

mmatur (approved):
LGTM

sdelicata (approved):
Hi @DorianBlues and thx for your contribution!
LGTM 👍

traefiker:

nmengin:
Close and re-open for CI

PR #11906, opened 2025-07-16
Add k8s.pod.uid and k8s.pod.name resource attributes automatically

kevinpollet:
### What does this PR do?

This pull request adds a K8s detector to add `k8s.pod.uid` and `k8s.pod.name` resource attributes for OTel tracing/logs/metrics.

### More

- [ ] Added/updated tests

### Additional Notes

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -90,7 +90,7 @@ Complete documentation is available at https://traefik.io`,
 }
 
 func runCmd(staticConfiguration *static.Configuration) error {
-	if err := setupLogger(staticConfiguration); err != nil {
+	if err := setupLogger(context.Background(), staticConfiguration); err != nil {
 		return fmt.Errorf("setting up logger: %w", err)
 	}
 
@@ -210,8 +210,8 @@ func setupServer(staticConfiguration *static.Configuration) (*server.Server, err
 		}
 	}
 	metricsRegistry := metrics.NewMultiRegistry(metricRegistries)
-	accessLog := setupAccessLog(staticConfiguration.AccessLog)
-	tracer, tracerCloser := setupTracing(staticConfiguration.Tracing)
+	accessLog := setupAccessLog(ctx, staticConfiguration.AccessLog)
+	tracer, tracerCloser := setupTracing(ctx, staticConfiguration.Tracing)
 	observabilityMgr := middleware.NewObservabilityMgr(*staticConfiguration, metricsRegistry, semConvMetricRegistry, accessLog, tracer, tracerCloser)
 
 	// Entrypoints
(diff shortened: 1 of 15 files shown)


pkg/types/k8sdetector.go
sdelicata, line 1: WDYT about adding tests for this new `KubernetesDetector` component?

pkg/types/k8sdetector.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	// The InClusterConfig function returns a config for the Kubernetes API server
+	// when it is running inside a Kubernetes cluster.
+	config, err := rest.InClusterConfig()
+	if err != nil && errors.Is(err, rest.ErrNotInCluster) {
+		return resource.Empty(), nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("creating in cluster config: %w", err)
```
sdelicata, line 36: Nitpick for the sake of consistency:
```suggestion
		return resource.Empty(), fmt.Errorf("creating in cluster config: %w", err)
```

pkg/types/k8sdetector.go
```diff
@@ -0,0 +1,70 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kclientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// KubernetesDetector detects the metadata of the Traefik pod running in a Kubernetes cluster.
+// It reads the pod name from the hostname file and the namespace from the service account namespace file and queries the Kubernetes API to get the pod's UID.
+type KubernetesDetector struct{}
+
+func (p KubernetesDetector) Detect(ctx context.Context) (*resource.Resource, error) {
```
sdelicata:
```suggestion
func (k KubernetesDetector) Detect(ctx context.Context) (*resource.Resource, error) {
```

cmd/traefik/traefik.go
```diff
@@ -90,7 +90,7 @@ Complete documentation is available at https://traefik.io`,
 }
 
 func runCmd(staticConfiguration *static.Configuration) error {
-	if err := setupLogger(staticConfiguration); err != nil {
+	if err := setupLogger(context.Background(), staticConfiguration); err != nil {
```
sdelicata: Maybe I missed something, but is it intentional to not use the same context as the one declared few line below?

pkg/types/k8sdetector.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil && errors.Is(err, rest.ErrNotInCluster) {
+		return resource.Empty(), nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("creating in cluster config: %w", err)
+	}
+
+	client, err := kclientset.NewForConfig(config)
+	if err != nil {
+		return resource.Empty(), fmt.Errorf("creating Kubernetes client: %w", err)
```
lbenguigui: I don't get why you return an empty resource here. When an error is raised, I believe we should return nil instead of empty resource.

sdelicata: Basically resource.Empty() is nil, but conceptually I agree with you.

pkg/types/k8sdetector.go
rtribotte: @sdelicata It seems complicated to test it, as after mocking the file and K8s client, we would only test the initialization of struct.
An integration test would have been the solution to us, but it not an easy task that we preferred to delay.

traefiker:
:no_entry_sign: too many retry [1/1]: conflicts must be resolved in the PR

PR #11907, opened 2025-07-16
Redact logged static configuration

jspdown:
### What does this PR do?

This PR redacts the static configuration that is logged when Traefik starts with the DEBUG log level.

### Motivation

The static configuration can contain sensitive (data such as TLS keys, tokens, ...) that we don't want logged, even in DEBUG mode. 

### More

- [ ] ~Added/updated tests~

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"crypto/x509"
-	"encoding/json"
 	"fmt"
 	stdlog "log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


nmengin:
Hey @jspdown,

It looks like a security fix, could you rebase on the branch `v2.11`, please?

cmd/traefik/traefik.go
```diff
@@ -100,12 +100,11 @@ func runCmd(staticConfiguration *static.Configuration) error {
 
 	log.WithoutContext().Infof("Traefik version %s built on %s", version.Version, version.BuildDate)
 
-	jsonConf, err := json.Marshal(staticConfiguration)
+	redactedStaticConfiguration, err := redactor.RemoveCredentials(staticConfiguration)
 	if err != nil {
 		log.WithoutContext().Errorf("Could not marshal static configuration: %v", err)
```
rtribotte:
```suggestion
		log.WithoutContext().Errorf("Could not redact static configuration: %v", err)
```

rtribotte (approved):
LGTM

cmd/traefik/traefik.go
jspdown: I understand that we might want the term "redact" rather than "marshal" but don't you think we should keep the "Could not" form?

rtribotte: @jspdown You are right! I'm adjusting the suggestion.

PR #11911, opened 2025-07-17
Fix client arbitrary file access during archive extraction zipslip

odaysec:
https://github.com/traefik/traefik/blob/5c94bbf122408399fa9e9da3c8abb1b7fa7300c2/pkg/plugins/client.go#L263-L268

https://github.com/traefik/traefik/blob/5c94bbf122408399fa9e9da3c8abb1b7fa7300c2/pkg/plugins/client.go#L281-L290

Extracting files from a malicious zip file, or similar type of archive, is at risk of directory traversal attacks if filenames from the archive are not properly validated. archive paths. Zip archives contain archive entries representing each file in the archive. These entries include a file path for the entry, but these file paths are not restricted and may contain unexpected special elements such as the directory traversal element (..). If these file paths are used to create a filesystem path, then a file operation may happen in an unexpected location. This can result in sensitive information being revealed or deleted, or an attacker being able to influence behavior by modifying unexpected files.

fix the issue, we need to validate the file paths extracted from the zip archive to ensure they do not contain directory traversal elements (`..`) or attempt to write files outside the intended destination directory. This can be achieved by:

1. Checking if `f.Name` contains `..` or other invalid path elements before constructing the output path.
2. Using `filepath.Clean` to normalize the path and ensure it resolves within the intended destination directory.
3. Comparing the resolved absolute path of the output file with the destination directory to ensure it is a subdirectory.

The changes will be made in the `unzipFile` function, where the file paths are constructed and used.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/plugins/client.go
+++ b/pkg/plugins/client.go
@@ -288,7 +288,7 @@ func unzipFile(f *zipa.File, dest string) error {
 		return fmt.Errorf("no root directory: %s", f.Name)
 	}
 
-	// Validate and sanitize the file path
+	// Validate and sanitize the file path.
 	cleanName := filepath.Clean(pathParts[1])
 	if strings.Contains(cleanName, "..") {
 		return fmt.Errorf("invalid file path in archive: %s", f.Name)
--- a/pkg/plugins/client.go
+++ b/pkg/plugins/client.go
@@ -282,8 +282,11 @@ func unzipFile(f *zipa.File, dest string) error {
 
 	defer func() { _ = rc.Close() }()
 
-	// Split to discard the first part of the path, which is the archive name.
+	// Split to discard the first part of the path, which is [organization]-[project]-[release commit sha1] when the archive is a Yaegi go plugin with vendoring.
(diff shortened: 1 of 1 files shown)


rtribotte (approved):
Thanks!

pkg/plugins/client.go
```diff
@@ -280,24 +282,48 @@ func unzipFile(f *zipa.File, dest string) error {
...
 	defer func() { _ = rc.Close() }()
 
+	// Split to discard the first part of the path, which is [organization]-[project]-[release commit sha1] when the archive is a Yaegi go plugin with vendoring.
 	pathParts := strings.SplitN(f.Name, "/", 2)
-	p := filepath.Join(dest, pathParts[1])
+	if len(pathParts) < 2 {
+		return fmt.Errorf("no root directory: %s", f.Name)
+	}
+
+	// Validate and sanitize the file path
```
kevinpollet:
```suggestion
	// Validate and sanitize the file path.
```

kevinpollet (approved):
Thanks 👍

PR #11924, opened 2025-07-22
Remove all mentions of ordering for TLSOption `CurvePrefences` field

jnoordsij:
### What does this PR do?

This PR removes all details mentioning that the `CurvePrefences` field can be used to set the order of the curves used for key exchange in TLS, given that this no longer holds true.

As of Go 1.24, the ordering is now predetermined by the `crypto/tls` library and cannot be overridden. See also https://tip.golang.org/doc/go1.24#cryptotlspkgcryptotls and https://github.com/golang/go/issues/69393.

> Key exchange ordering is now handled entirely by the `crypto/tls` package. The order of [`Config.CurvePreferences`](https://tip.golang.org/pkg/crypto/tls#Config.CurvePreferences) is now ignored, and the contents are only used to determine which key exchanges to enable when the field is populated.

### Motivation

Making sure documentation is up-to-date.

### More

- [ ] Added/updated tests


rtribotte (approved):
LGTM 👍

Issue #11915, opened 2025-07-17
Add the option to define custom cipher suites for backend serverTransport


AliMickey:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

Add the option to define custom cipher suites for backend serverTransport.

Similar to:

```
tls:
  options:
    default:
      cipherSuites:
        - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
```

Allow configuring custom cipher suites onto the backend serverTransport configuration. Example:

```
http:
  serversTransports:
    mytransport:
      cipherSuites:
        - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
````

This allows for greater compatibility with legacy systems that do not support the default ciphers that Traefik enables.

NEwa-05:
Hi, 
I'm interested in bringing this along to Traefik.
If you would like to assign me to it, please do.

kevinpollet:
Closed by https://github.com/traefik/traefik/pull/11965

PR #11965, opened 2025-08-01
Add the option to define custom cipher suites for backend serversTransport

NEwa-05:
### What does this PR do?

This PR adds capabilities to set cipher suite and TLS version (max and min) at the serversTransport level.

### Motivation

This PR implements the feature requested in issue [#11915](https://github.com/traefik/traefik/issues/11915)

### More

- [X] Added/updated tests


pkg/server/service/transport.go
```diff
@@ -169,16 +169,54 @@ func (t *TransportManager) createTLSConfig(cfg *dynamic.ServersTransport) (*tls.
...
 
+		// map and validate the CipherSuite passed in the configuration
+		ciphersList := make([]uint16, 0)
+		if cfg.CipherSuites != nil {
+			for _, cipher := range cfg.CipherSuites {
+				if cipherID, exists := traefiktls.CipherSuites[cipher]; exists {
+					ciphersList = append(ciphersList, cipherID)
+				} else {
+					// CipherSuite listed in the configuration does not exist in our list
+					return nil, fmt.Errorf("invalid CipherSuite: %s", cipher)
```
sdelicata: I'm not sure we want to return an error here. Maybe skipping the cipher suite and logging an error is enough. WDYT?

pkg/server/service/transport_test.go
```diff
@@ -183,6 +183,47 @@ func TestKeepConnectionWhenSameConfiguration(t *testing.T) {
 	assert.EqualValues(t, 2, count)
 }
 
+func TestCipherSuites(t *testing.T) {
```
sdelicata: You only tested the happy pass. Could you please update this to verify the behavior with an invalid cipher suite and an empty suite?

pkg/server/service/transport_test.go
```diff
@@ -183,6 +183,47 @@ func TestKeepConnectionWhenSameConfiguration(t *testing.T) {
...
+		"test": {
+			ServerName:   "example.com",
+			RootCAs:      []types.FileOrContent{types.FileOrContent(LocalhostCert)},
+			CipherSuites: []string{"TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"},
+			MaxVersion:   "VersionTLS12",
+		},
+	}
+
+	transportManager.Update(dynamicConf)
+
```
sdelicata: The test doesn't verify that the TLS config was actually set correctly. Could you improve it please?

pkg/server/service/transport_test.go
```diff
@@ -183,6 +183,47 @@ func TestKeepConnectionWhenSameConfiguration(t *testing.T) {
...
+	require.NoError(t, err)
+
+	client := http.Client{Transport: tr}
+
+	resp, err := client.Get(srv.URL)
+	require.NoError(t, err)
+
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+}
+
```
sdelicata, line 266: Could you add a similar test for min and max version too?

sdelicata (changes requested):
Thanks for your contribution @NEwa-05!
I've taken the liberty of pushing a few changes directly to your branch to help move this forward
I also made a bunch of suggestions.

pkg/server/service/transport.go
NEwa-05: Make sense, I'll update it.

pkg/server/service/transport_test.go
NEwa-05: Sure, will add an invalid cipher test.

pkg/server/service/transport_test.go
NEwa-05: Indeed, I'll improve my test.

pkg/server/service/transport_test.go
NEwa-05: Yep, will do it.

NEwa-05:
@sdelicata I've updated the tests and rebased on the latest update of the master branch. I hope this will fit the need.

pkg/server/service/transport.go
```diff
@@ -169,16 +169,49 @@ func (t *TransportManager) createTLSConfig(cfg *dynamic.ServersTransport) (*tls.
...
 		}
 
+		cipherSuites := make([]uint16, 0)
+		if cfg.CipherSuites != nil {
+			for _, cipher := range cfg.CipherSuites {
+				if cipherID, exists := traefiktls.CipherSuites[cipher]; exists {
+					cipherSuites = append(cipherSuites, cipherID)
+				} else {
+					log.Error().Msgf("Invalid cipher: %v", cipher)
+					cipherSuites = nil
```
lbenguigui, line 184: Can you explain why we reset the slice instead of just skipping the cipher?

NEwa-05: When I wrote this, I wanted to be sure there was no data in the slice. 
Now that you are pointing this out to me, it is unnecessary in this case since the slice is just initialized if the cipher is invalid.
I'll update the PR.

NEwa-05: @lbenguigui, after some tests, I know why I've done this, and it may need clarification on the behavior we want to achieve here.
When setting the cipherSuites slice to nil, traefik will use the "default" set of cipherSuites, will log an error, and will connect to the backend.
If I don't set the slice to nil, the connection with the backend will fail because no cipher suites can be found.
What would be the best option here: logs and failure to contact the backend, or just an error log?

lbenguigui: In this case, if you want to set `cipherSuites` to nil, you should add a `break` to ensure no other cipher will be added

NEwa-05: Ok, thanks, I'll update my pr

lbenguigui (approved):
LGTM

pkg/server/service/transport.go
```diff
@@ -169,16 +169,50 @@ func (t *TransportManager) createTLSConfig(cfg *dynamic.ServersTransport) (*tls.
...
```
sdelicata, line 184: If I'm not wrong, setting nil here means "use Go's default cipher suites", which may include ciphers the administrator explicitly wanted to exclude.
Since the error log can easily be missed in production, we are potentially allowing weaker cipher suites.
That's something that should at least be explained in the error log, WDYT?

pkg/server/service/transport.go
```diff
@@ -169,16 +169,50 @@ func (t *TransportManager) createTLSConfig(cfg *dynamic.ServersTransport) (*tls.
...
+				}
+			}
+		}
+
+		var minVersion uint16
+		if cfg.MinVersion != "" {
+			if value, exists := traefiktls.MinVersion[cfg.MinVersion]; exists {
+				minVersion = value
+			} else {
+				log.Error().Msgf("invalid TLS minimum version: %s", cfg.MinVersion)
```
sdelicata:
```suggestion
				log.Error().Msgf("Invalid TLS minimum version: %s", cfg.MinVersion)
```

pkg/server/service/transport.go
```diff
@@ -169,16 +169,50 @@ func (t *TransportManager) createTLSConfig(cfg *dynamic.ServersTransport) (*tls.
...
+			}
+		}
+
+		var maxVersion uint16
+		if cfg.MaxVersion != "" {
+			if value, exists := traefiktls.MaxVersion[cfg.MaxVersion]; exists {
+				maxVersion = value
+			} else {
+				log.Error().Msgf("invalid TLS maximum version: %s", cfg.MaxVersion)
```
sdelicata:
```suggestion
				log.Error().Msgf("Invalid TLS maximum version: %s", cfg.MaxVersion)
```

pkg/server/service/transport.go
```diff
@@ -169,16 +169,50 @@ func (t *TransportManager) createTLSConfig(cfg *dynamic.ServersTransport) (*tls.
...
+			}
+		}
+
```
sdelicata, line 207: Shouldn't we validate that minVersion < maxVersion?
What's the behavior in such a case? Will the handshake fail?
WDYT?

pkg/server/service/transport.go
NEwa-05: make sense, I'll update the log line to mention we are falling back to the default cipher suite

pkg/server/service/transport.go
NEwa-05: Didn't think of this, but that totally makes sense. I'll add a test and modify my code accordingly

pkg/server/service/transport.go
```diff
@@ -169,16 +169,58 @@ func (t *TransportManager) createTLSConfig(cfg *dynamic.ServersTransport) (*tls.
...
 			return nil, errors.New("TLS and SPIFFE configuration cannot be defined at the same time")
 		}
 
+		cipherSuites := make([]uint16, 0)
+		if cfg.CipherSuites != nil {
+			for _, cipher := range cfg.CipherSuites {
+				if cipherID, exists := traefiktls.CipherSuites[cipher]; exists {
+					cipherSuites = append(cipherSuites, cipherID)
+				} else {
+					log.Error().Msgf("Invalid cipher: %v, falling back to default CipherSuite.", cipher)
```
NEwa-05, line 183: @sdelicata, log updated.
Do you think this is enough?

traefiker:

traefiker:

traefiker:

traefiker:

PR #11971, opened 2025-08-03
log mTLS failures and other http errors in zerolog.ErrorLevel

### LINKED ISSUE #11955: mTLS Connection failures are logged in debug mode only
### Welcome!

### What did you do?

mTLS connection failures are not uncommon and being able to see them being logged is helpful for investigating purposes, to collect metrics of failed connections and setup alerts accordingly.

When using traefik as a reverse proxy, I expected to see such failures to be logged in the error logging level

### What did you see instead?

However, traefik instead logs them in the debug logging level

https://github.com/traefik/traefik/blob/24cede62eee0a7aaf70a72f341e097452773ccd5/pkg/server/server_entrypoint_tcp.go#L648-L652

Debug level is too verbose and its easy to miss the errors, especially when multiple connections are made in a short span of time. It also puts additional load on log processing tools to parse high volume debug logs.

It would be awesome if these logs were part of the `zerolog.ErrorLevel`

Version:      3.4.5
Codename:     chaource
Go version:   go1.24.5
Built:        2025-07-23T09:26:53Z
OS/Arch:      linux/arm64

```yaml
services:
  traefik:
    image: traefik:v3.4
    container_name: traefik
    restart: unless-stopped
    security_opt:
      - no-new-privileges:true

    ports:
      - "8080:80"
      - "6443:443"
      - "8880:8080"

    volumes:
      - ./certs:/certs:ro
      - ./dynamic:/dynamic:ro

    command:
      # EntryPoints
      - "--entrypoints.web.http.redirections.entrypoint.to=websecure"
      - "--entrypoints.web.http.redirections.entrypoint.scheme=https"
      - "--entrypoints.web.http.redirections.entrypoint.permanent=true"
      - "--entrypoints.websecure.address=:443"
      - "--entrypoints.websecure.http.tls=true"

      # Attach the static configuration tls.yaml file that contains the tls configuration settings
      - "--providers.file.filename=/dynamic/tls.yaml"

      # Providers 
      - "--providers.docker=true"

      # API & Dashboard 
      - "--api.dashboard=true"

      # Observability 
      - "--log.level=INFO"
      - "--accesslog=true"
      - "--accesslog.format=json"
```

`dynamic/tls.yaml`
```yaml
tls:
  certificates:
    - certFile: /certs/fleet-cert.pem
      keyFile:  /certs/fleet-key.pem
```

_No response_

krajpur0:
### What does this PR do?
Fixes #11955 

### Motivation
When using traefix as a reverse proxy, it was surprising to see it not log mTLS connection failures as errors.
I think that's probably a bug.

Before:
```
2025-08-03T16:37:35+05:30 DBG ../../../../../opt/homebrew/Cellar/go/1.24.5/libexec/src/log/log.go:245 > http: TLS handshake error from [::1]:50461: tls: failed to verify certificate: x509: certificate signed by unknown authority
2025-08-03T16:37:43+05:30 DBG ../../../../../opt/homebrew/Cellar/go/1.24.5/libexec/src/log/log.go:245 > http: TLS handshake error from [::1]:50471: tls: client didn't provide a certificate
```

After:
```
2025-08-03T16:39:11+05:30 ERR ../../../../../opt/homebrew/Cellar/go/1.24.5/libexec/src/log/log.go:245 > http: TLS handshake error from [::1]:50526: tls: failed to verify certificate: x509: certificate signed by unknown authority
2025-08-03T16:39:21+05:30 ERR ../../../../../opt/homebrew/Cellar/go/1.24.5/libexec/src/log/log.go:245 > http: TLS handshake error from [::1]:50533: tls: client didn't provide a certificate
```
### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/logs/log.go
+++ b/pkg/logs/log.go
@@ -2,6 +2,7 @@ package logs
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/rs/zerolog"
 )
@@ -32,4 +33,18 @@ func (n NoLevelHook) Run(e *zerolog.Event, level zerolog.Level, _ string) {
 
 func msgFunc(i ...any) func() string {
 	return func() string { return fmt.Sprint(i...) }
+}
+
+func TLSErrors(logger zerolog.Logger) zerolog.Logger {
+	return logger.Hook(
+		zerolog.HookFunc(func(e *zerolog.Event, level zerolog.Level, message string) {
+			if !strings.HasPrefix(message, "http: TLS handshake error from") {
+				e.Discard()
+				return
+			}
+
+			if level == zerolog.NoLevel {
+				e.Str("level", zerolog.ErrorLevel.String())
+			}
+		}))
 }
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -647,7 +647,7 @@ func createHTTPServer(ctx context.Context, ln net.Listener, configuration *stati
 	serverHTTP := &http.Server{
 		Protocols:      &protocols,
 		Handler:        handler,
-		ErrorLog:       stdlog.New(logs.NoLevel(log.Logger, zerolog.DebugLevel), "", 0),
+		ErrorLog:       stdlog.New(logs.NoLevel(log.Logger, zerolog.ErrorLevel), "", 0),
 		ReadTimeout:    time.Duration(configuration.Transport.RespondingTimeouts.ReadTimeout),
 		WriteTimeout:   time.Duration(configuration.Transport.RespondingTimeouts.WriteTimeout),
 		IdleTimeout:    time.Duration(configuration.Transport.RespondingTimeouts.IdleTimeout),
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -18,7 +18,6 @@ import (
 	"github.com/containous/alice"
 	gokitmetrics "github.com/go-kit/kit/metrics"
 	"github.com/pires/go-proxyproto"
-	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"github.com/traefik/traefik/v3/pkg/config/static"
 	"github.com/traefik/traefik/v3/pkg/ip"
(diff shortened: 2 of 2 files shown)


kevinpollet:
Hello @krajpur0 and thanks for opening this,

Changing the log level in the `http.Server` could lead to log transient/unwanted errors at the Error level. We would prefer to identify and log only TLS errors at the Error level, but we are still unsure that this should be done. We think that someone could use that to flood the server logs.

What do you think?

krajpur0:
The documentation states the following:
```
	// ErrorLog specifies an optional logger for errors accepting
	// connections, unexpected behavior from handlers, and
	// underlying FileSystem errors.
	// If nil, logging is done via the log package's standard logger.
	ErrorLog *log.Logger
```

In my opinion, those are errors that should ideally be available to users of traefik as errors, if someone's trying flood server logs with errors then logging them as such (especially in environments with mTLS) would be important as it would suggest nefarious activity.

However, if you still want to only log TLS errors what would the approach be?
Do I create a new `logs.TLSErrors` hook and discard all events except ones that contain the bytes `TLS`?

Something like:
```golang
func TLSErrors(logger zerolog.Logger) zerolog.Logger {
	return logger.Hook(
		zerolog.HookFunc(func(e *zerolog.Event, level zerolog.Level, message string) {
			if !strings.Contains(message, "TLS") {
				e.Discard()
				return
			}

			if level == zerolog.NoLevel {
				e.Str("level", zerolog.ErrorLevel.String())
			}
		}))
}
```

krajpur0:
I've pushed a commit that only logs TLS related errors

I took a look through net/http/server.go on where `ErrorLog` is called: https://github.com/golang/go/blob/master/src/net/http/server.go

```go
// These occur when sending responses
w.conn.server.logf("http: response.WriteHeader on hijacked connection from %s (%s:%d)", caller.Function, path.Base(caller.File), caller.Line)
w.conn.server.logf("http: superfluous response.WriteHeader call from %s (%s:%d)", caller.Function, path.Base(caller.File), caller.Line)
w.conn.server.logf("http: invalid Content-Length of %q", cl)
w.conn.server.logf("http: WriteHeader called with both Transfer-Encoding of %q and a Content-Length of %d", te, w.contentLength)
w.conn.server.logf("http: response.Write on hijacked connection from %s (%s:%d)", caller.Function, path.Base(caller.File), caller.Line)

// These occur when handling an incoming connection
c.server.logf("http: panic serving %v: %v\n%s", c.remoteAddr, err, buf)
c.server.logf("http: TLS handshake error from %s: %v", c.rwc.RemoteAddr(), reason)
s.logf("http: Accept error: %v; retrying in %v", err, tempDelay)
```

all calls from `w.conn.server.logf` would come from traefik, since traefik writes the response, the only place where a connection would generate errors are the last three lines and all of those are worth logging in Error Level

juliens:
Thanks for the contribution!

After discussing this with the maintainers, we’ve decided not to move forward with this change.

mTLS handshake failures are client-driven events, not server-side errors. A client—whether misconfigured or malicious—can easily generate a high volume of invalid handshakes. Promoting these messages to the ERROR level would make Traefik excessively noisy and could lead to log-flooding or unnecessary alerts for users.

For us, the ERROR level is reserved for situations where the server itself fails to operate correctly, not for cases where the client does not meet the expected TLS requirements. The current logging level reflects that distinction.

We therefore prefer to keep the existing behavior, but we appreciate the proposal and the time you put into it.

Thanks again!

PR #11977, opened 2025-08-05
feat: add new certificatesresolvers options

### LINKED ISSUE #9439: ACME provider startup time configuration
### Welcome!

To interact (through the ACME protocol) with TLS certificates providers (such as Let's Encrypt), Traefik includes an internal (i.e. "hidden") mechanism, unfortunately also colloquially named "ACME provider". Notably, it is indirectly configured and influenced by the Certificates Resolvers part of the configuration (https://doc.traefik.io/traefik/https/acme/).

That said, a number of problems can arise when, on Traefik startup, for example this ACME provider is not yet ready (e.g. not listening yet for a Let's Encrypt challenge) and/or it is not yet reachable from the outside, or conversely cannot yet reach an external resource. In such a condition, if it then happens at the same time that the mechanism to fetch/generate a certificate is triggered, said mechanism might fail, which is quite problematic.

For that reason, it would be interesting, to be able to assert or control the moment at which said ACME provider can be considered operational.
This could be viewed as several different mechanisms, such as:
- a specific delay
- some sort of healthiness/readiness check (both of itself, and of the needed external resources)

ldez:
### What does this PR do?

Add `certificatesresolvers` options:
- `acme.tlschallenge.delay`
- `acme.disableCommonName`

### Motivation

Fixes #9439

### More

~~- [ ] Added/updated tests~~

### Additional Notes

Related to https://github.com/traefik/traefik/issues/11894

Not related to this PR, but https://github.com/traefik/traefik/issues/11558 can be closed.

### DIFF
--- a/docs/content/https/acme.md
+++ b/docs/content/https/acme.md
@@ -201,6 +201,34 @@ when using the `TLS-ALPN-01` challenge, Traefik must be reachable by Let's Encry
     --certificatesresolvers.myresolver.acme.tlschallenge=true
     ```
 
+#### `Delay`
+
+The delay between the creation of the challenge and the validation.
+A value lower than or equal to zero means no delay.
+
+```yaml tab="File (YAML)"
+certificatesResolvers:
+  myresolver:
+    acme:
+      # ...
+      tlsChallenge:
+        # ...
+        delay: 12
+```
+
+```toml tab="File (TOML)"
+[certificatesResolvers.myresolver.acme]
+  # ...
+  [certificatesResolvers.myresolver.acme.tlsChallenge]
+    # ...
+    delay = 12
+```
+
+```bash tab="CLI"
+# ...
+--certificatesresolvers.myresolver.acme.tlschallenge.delay=12
+```
+
 ### `httpChallenge`
 
 Use the `HTTP-01` challenge to generate and renew ACME certificates by provisioning an HTTP resource under a well-known URI.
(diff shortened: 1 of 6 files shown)


xxsheddy:
Hello @ldez, I've been working with the maintainers to revamp the documentation. We now have a reference section where each feature is explained in more detail and we keep a table containing all the options available so that readers can easily find them at a glance. 

please can you also update the reference section with the new options you added? 

We want to keep the reference section updated with your contribution.

This is the page you need to update: https://doc.traefik.io/traefik/reference/install-configuration/tls/certificate-resolvers/acme/

Thanks!

docs/content/https/acme.md
```diff
@@ -201,6 +201,34 @@ when using the `TLS-ALPN-01` challenge, Traefik must be reachable by Let's Encry
     --certificatesresolvers.myresolver.acme.tlschallenge=true
     ```
 
+#### `Delay`
+
+The delay between the creation of the challenge and the validation.
+A value lower than or equal to zero means no delay.
```
mmatur:
```suggestion
#### `Delay`

_Optional, Default=0_

The delay between the creation of the challenge and the validation.
A value lower than or equal to zero means no delay.
```

traefiker:

docs/content/https/acme.md
```diff
@@ -201,6 +201,36 @@ when using the `TLS-ALPN-01` challenge, Traefik must be reachable by Let's Encry
...
+
+```yaml tab="File (YAML)"
+certificatesResolvers:
+  myresolver:
+    acme:
+      # ...
+      tlsChallenge:
+        # ...
+        delay: 12
+```
```
kaaax0815, line 219: any documentation for unit is missing

PR #11981, opened 2025-08-06
feat(proxy/tcp): suppress benign "use of closed network connection" errors

xpload32004:
### What does this PR do?

Introduces a new isBenignCloseError helper in the TCP proxy to consistently classify and silence benign connection-closure errors (nil, io.EOF, net.ErrClosed, and “use of closed network connection”). Integrates this function into the proxy’s error-handling path and adds unit tests.

### Motivation

Previously, benign shutdowns of TCP connections were treated as errors which polluted the logs at the debug level making it hard to find actual errors. By centralizing and standardizing benign-error detection, we ensure normal connection closures no longer surface as errors while debug logging.

### More

- [✅] Added/updated tests
- [] Added/updated documentation

### DIFF
--- a/pkg/tcp/proxy.go
+++ b/pkg/tcp/proxy.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


hhromic:
This PR seems to be a duplicate (in goal) to <https://github.com/traefik/traefik/pull/11838>.

xpload32004:
> This PR seems to be a duplicate (in goal) to #11838.

Yes, it does look similar in nature. The main difference seems to be we're seeing these errors continuously during runtime at DEBUG log level, not just at shutdown.

kevinpollet:
Hello @xpload32004 and thanks for your contribution,

We have set the status to "design-review" to allow us to check the PR and ensure there is no deep impact on Traefik before moving forward.

We will keep you updated once the analysis is done.

PR #11990, opened 2025-08-11
fix: add missing port label in Docker setup guide for whoami service

GreggSchofield:
### What does this PR do?

This PR fixes an issue in the Docker setup guide where the whoami service configuration was missing the required port label (`traefik.http.services.whoami.loadbalancer.server.port=80`). Without this label, HTTPS requests to the whoami service would result in a 404 error, while HTTP requests would correctly show the "Moved Permanently" message. Adding this label ensures that both HTTP and HTTPS requests work as expected, providing a better experience for users following the guide.

The fix ensures that:
- HTTPS requests to whoami.docker.localhost work correctly
- The configuration example is correct, complete and explicit
- Users don't encounter unexpected 404 errors when following the guide

### Motivation

While following the Docker setup guide as a Platform Engineer evaluating Traefik for our internal developer platform, I encountered an issue where HTTPS requests to the whoami service were resulting in 404 errors. After investigating, I found that adding the port label fixed the issue. I'm submitting this PR to help other users avoid the same problem and to ensure the documentation provides a complete, working configuration example.

### More

- [ ] Added/updated tests


rtribotte:
Hi @GreggSchofield,

We suspect that you are experiencing the same issue as the one described in this GitHub issue: https://github.com/traefik/traefik/issues/11946. Are you using Docker for Mac?

GreggSchofield:
@rtribotte thank you for the speedy reply! Yes I am using Doxker for Mac, and it looks like this PR is a result of the regression in the underlying Docker for Mac repository. Happy to close this. Thanks!

PR #11991, opened 2025-08-11
Bump to github.com/pires/go-proxyproto v0.8.1

### LINKED ISSUE #11988: IPv6/TCP6 in proxy protocol header makes traefik drop connection
### Welcome!

### What did you do?

In my setup, nginx is used to route connections based on SNI into virtual machines. Inside the virtual machine, traefik is used to route traffic into docker containers.

I want to be able to handle incoming requests from connections via IPv6.

### What did you see instead?

If nginx is configured to also accept connections via IPv6, traefik stops responding to incoming requests. Firefox shows an error message like "An error occurred during a connection to host.example.com. PR_END_OF_FILE_ERROR", other tools like curl also report that the connection was closed before any data was sent back.

Traefik logs some DBG messages from inside the postgres middleware, "proxy protocol: invalid address".

Using tcpdump I can see that nginx sends a PROXY header like

```
PROXY TCP6 ::ffff:1.2.3.4 ::ffff:1.2.3.5 62123 443
```

If I disable IPv6 on nginx, everything works as expected and the PROXY headers look like

```
PROXY TCP4 1.2.3.4 1.2.3.5 62123 443
```

I have tried this with versions 3.1, 3.4.3 and 3.5.0 of traefik.

Ultimately, I have traced the problem back to the dependency github.com/pires/go-proxyproto@v0.6.1 which only support TCP4 addresses. 
In version v0.7.0 they added support for TCP6 in the PROXY header which is what nginx uses in my case.

Updating the dependency to 0.8.1 seems to work fine. I am prepared to provide a PR with an additional unit test for the TCP6 case.

relevant portion of the config is this:

```toml
[entryPoints]
  [entryPoints.http]
    address = ":80"   
    [entryPoints.http.proxyProtocol]
      insecure = false
      trustedIPs = ["127.0.0.1/32", "10.0.0.0/8"]

  [entryPoints.https]
    address = ":443"
    [entryPoints.https.proxyProtocol]
      insecure = false
      trustedIPs = ["127.0.0.1/32", "10.0.0.0/8"]
```

It's not an issue with the configured trustedIPs, setting insecure to true does not change behaviour.

```log
2025-08-08T00:10:37Z DBG github.com/traefik/traefik/v3/pkg/server/router/tcp/postgres.go:32 > Error while peeking first bytes error="proxyproto: invalid address"
```

this was the only message emitted for each failing request

rtribotte:
### What does this PR do?

This PR bumps github.com/pires/go-proxyproto to v0.8.1.

### Motivation

Fixes #11988

### More

- [ ] Added/updated tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,7 +51,7 @@ require (
 	github.com/mitchellh/hashstructure v1.0.0
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/patrickmn/go-cache v2.1.0+incompatible
-	github.com/pires/go-proxyproto v0.6.1
+	github.com/pires/go-proxyproto v0.8.1
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // No tag on the repo.
 	github.com/prometheus/client_golang v1.19.1
 	github.com/prometheus/client_model v0.6.1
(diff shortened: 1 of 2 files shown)


kevinpollet (approved):

traefiker:

PR #11998, opened 2025-08-13
feat: add global TCP access logging

### LINKED ISSUE #7800: [Feature requests] Access log for TCP routers

Feature

Access log for TCP routers

There is currently no access log for TCP routers. That would be awesome to log incoming traffic for TCP routers so I can get details on the client and the port it uses ? I have several TCP entrypoints exposed (some in plain text, some secured) and I'd like to know the port used by the different clients.

Kronusus:
Implement a TCP access log handler that logs connection start and end events, including client/server addresses, bytes in/out, duration, client certificate and errors. - Integrate the TCP access log middleware globally at the entrypoint/server level, mirroring HTTP access log behavior. - Enable TCP access logging via the global static configuration (accessLog), with no per-router or manual wiring required. - Ensure thread-safe log writes and support for both JSON and text log formats.

### What does this PR do?

This PR Implements a TCP access log handler that logs connection start and end events, including client/server addresses, bytes in/out, duration, client certificate and errors. 
- Integrate the TCP access log middleware globally at the entrypoint/server level, mirroring HTTP access log behavior. 
- Enable TCP access logging via the global static configuration (accessLog), with no per-router or manual wiring required. 
- Ensure thread-safe log writes and support for both JSON and text log formats.

### Motivation

resolves #7800 

### More

- [ ] Added/updated tests

### Additional Notes

Early Pull Request draft --> I tested the feature on my end with some demo set-ups, but would appreciate some feedback if it goes in the right direction and if something should be changed to make a merge possible.

### DIFF
--- a/cmd/traefik/tcp_accesslog.go
+++ b/cmd/traefik/tcp_accesslog.go
@@ -1,0 +1,18 @@
+package main
+
+import (
+	"github.com/traefik/traefik/v3/pkg/middlewares/accesslogtcp"
+	"github.com/traefik/traefik/v3/pkg/types"
+)
+
+// setupTCPAccessLog creates a TCP access log handler from the config, or returns nil if not enabled.
+func setupTCPAccessLog(conf *types.AccessLog) *accesslogtcp.Handler {
+	if conf == nil {
+		return nil
+	}
+	handler, err := accesslogtcp.NewHandler(conf)
+	if err != nil {
+		return nil
+	}
+	return handler
+}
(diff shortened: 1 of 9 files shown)


kevinpollet:
Hello @Kronusus and thanks for your contribution,

We have set the status to "design-review" to allow us to check the PR and ensure there is no deep impact on Traefik before moving forward.

We will keep you updated once the analysis is done.

JosiahWhite:
How is that design-review going?

zarko-a:
any chance this can move forward? 

Having access logs for TCP routes  would be very useful.

nmengin:
Hey @zarko-a 

PR #12002, opened 2025-08-17
feat(provider): log provider namespace during startup

shreealt:
https://github.com/traefik/traefik/issues/11982#issuecomment-3174788944

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/aggregator/aggregator_test.go
+++ b/pkg/provider/aggregator/aggregator_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
 	"github.com/traefik/traefik/v3/pkg/provider"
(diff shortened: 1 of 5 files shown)


rtribotte:
@shreealt Thanks for this contribution!

Would you mind if we added review commits? We have a few adjustments to make and want to extend the fix for the ConsulCatalog provider, and it is easier to push commits rather than interact here.

shreemaan-abhishek:
sure!

pkg/provider/nomad/nomad_test.go
```diff
@@ -468,3 +468,40 @@ func Test_getNomadServiceData(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			p := &Provider{
+				namespace: tt.namespace,
+			}
+
+			result := p.GetNamespace()
+			if result != tt.expected {
+				t.Errorf("GetNamespace() = %v, want %v", result, tt.expected)
```
sdelicata: Could you please use the `assert.Equal()` function instead?

pkg/provider/aggregator/aggregator_test.go
```diff
@@ -76,3 +77,36 @@ func (p *providerMock) Provide(configurationChan chan<- dynamic.Message, pool *s
...
+	var _ provider.NamespaceProvider = providerWithNamespace
+	var _ provider.Provider = providerWithNamespace
+
+	// Test GetNamespace method
+	assert.Equal(t, "test-namespace", providerWithNamespace.GetNamespace())
+
+	// Test with empty namespace
+	providerEmptyNamespace := &mockNamespaceProvider{namespace: ""}
+	assert.Empty(t, providerEmptyNamespace.GetNamespace())
+}
```
sdelicata: The test name suggests it should test the aggregator's launchProvider method with namespace providers, but it completely ignores that functionality.
To me, just testing a getter returns what was set adds no value.
Could you please test the core feature?

pkg/provider/nomad/nomad_test.go
```diff
@@ -468,3 +468,40 @@ func Test_getNomadServiceData(t *testing.T) {
 	require.NoError(t, err)
 	require.Len(t, items, 2)
 }
+
+func TestProvider_GetNamespace(t *testing.T) {
```
sdelicata: To be honest, I wonder if this test has real value.

sdelicata (changes requested):
Hi @shreealt and thanks for your contribution!
I made a few comments concerning tests and renamed the `NamespaceProvider` to `NamespacedProvider` that sounds more accurate.

pkg/provider/nomad/nomad_test.go
shreealt: my motivation behind adding this test case was to test whether the getNamespace function (being used in the log) actually works. But now I feel silly, should I remove this?

pkg/provider/aggregator/aggregator_test.go
shreealt: hey, thanks for your apt review, I made new changes to address this comments.

pkg/provider/nomad/nomad_test.go
sdelicata: No need to feel silly at all! Your instinct to test the namespace functionality was completely reasonable. I should have been clearer in my feedback - sorry for that.

The good news is that GetNamespace() already gets tested through the aggregator test we refined together. When we verify the log output contains the namespace, we're effectively testing that the getter works correctly.

In general, we don't typically write unit tests for simple getter functions because they get tested indirectly when used by other code. But there's absolutely nothing wrong with your thinking here - it's better to over-test than under-test!

So yes, we can remove this test. Thanks for being so responsive to feedback! 🙏

sdelicata (approved):
LGTM

kevinpollet (approved):
Thanks 👍

PR #12007, opened 2025-08-20
Bump github.com/docker/docker to v28.3.3

kevinpollet:
### What does this PR do?

This pull request bumps github.com/docker/docker to v28.3.3

### Motivation

To be up-to-date.

### More

- [ ] Added/updated tests


hhromic:
Hello @kevinpollet , I just stumbled upon this PR while checking/testing Traefik v3.5.1.
First, I noticed that Traefik now uses/requires the Docker Swarm Node API to inspect a Node IP address.
Then, I noticed that you are populating the Node IP unconditionally from the `node.Status.Addr` field.

I wanted to make you aware (in case you aren't already) that there is a long-standing known limitation with this field: it can get populated with `0.0.0.0` in certain scenarios, particularly when inspecting a Swarm manager node that is a leader. We have observed this at least until Docker version 24.0.x in all of our swarm clusters. We have not verified it yet in later versions.

This problem has affected other software such as Portainer and Prometheus. See my comment in the Prometheus repository for more details: <https://github.com/prometheus/prometheus/issues/11060#issuecomment-1195170923>

A common workaround is to obtain the IP address of Swarm nodes using code like this:
```go
var node swarm.Node = ... // retrive info
var nodeIP string
if node.ManagerStatus != nil {
	nodeIP = strings.SplitN(node.ManagerStatus.Addr, ":", 2)[0]
} else {
	nodeIP = node.Status.Addr
}
```

PR #12042, opened 2025-09-05
feat: add health track to ratelimit middleware

david-garcia-garcia:
### What does this PR do?

In v3.4 a redis backend was introduced for the ratelimit middleware.

This introduces a hard dependency in the critical path: if redis is down or has a transient failure, the whole middleware goes down.

This is not a problem with the in memory backend because the chance of failure when updating the bucket is zero. But not with redis where you have a remote backend and many things can glitch (network, the backend itself, etc.)

This PR introduces three new configuration settings:

* backoffDuration
* backoffTimeout
* backoffThreshold

The ratelimit will stop working (requests will go through without issues) for a period of backoffDuration when backoffThreshold requests have failed in a time window of backoffThreshold.

Related issues:

* https://github.com/traefik/traefik/issues/12036
* https://github.com/traefik/traefik/issues/8061
* https://github.com/traefik/traefik/pull/10211

When not configured, current behaviour is honored, where failure to update the bucket results in the request returning a 500 error.

### Motivation

Ingress components are critical and should be resilient to failures when possible.

### More

- [X] Added/updated tests


nmengin:
Hello @david-garcia-garcia,

PR #12050, opened 2025-09-09
Implement HTTP2 HPACK table size options

### LINKED ISSUE #12045: Config of HTTP2 HPACK cache sizes to enable caching larger headers
### Welcome!

**Context**
I want to be able to configure the HTTP2 HPACK (RFC 7541) header cache sizes that my Traefik instance uses, so that I can reduce the traffic over the wire in situations when I have large headers.

Traefik uses the default that Golang specifies in it's net package, 4096 bytes. For contrast, Firefox sets the default to 65536.

Golang's default configuration may be appropriate for many use cases (e.g., high user count public facing web services) to avoid having a lot of RAM used for caching, however for systems with low bandwidth or internal systems, it may be more desirable to use more RAM to reduce traffic.

**Design**
Golang's underlying http2 library already supports the options, `MaxDecoderHeaderTableSize` and `MaxEncoderHeaderTableSize`. I would suggest to add additional properties to the `entrypoints.web.http2` directive to configure these, with a default set to zero (use Golang's default).

GCHQDeveloper548:
### What does this PR do?

Exposes HTTP2 HPACK options for the user to customise.

The default is zero, which tells Golang to use it's default of 4096 bytes.

Configuring a different header cache size allows the user to trade memory for the opportunity to cache more headers, reducing the amount of data sent over the wire. For very large headers this might have a big saving on network traffic, especially in network constrained scenarios.

Closes #12045

### Motivation

Motivation is detailed in the #12045 issue, but to summarise: being able to have a larger cache may help reduce network traffic on applications that use large headers, which may be advantageous for network constrained environments.

### More

- [x] Added/updated tests

### Additional Notes

Right now I've set the default to 0, which tells Golang to use it's default. Happy to switch it out instead so we explicitly set the default to 4096 so it's clearer to the user what the default value actually being used is. I felt this was a bit of a tradeoff since if Go updates on their side, it's likely for a reason and will likely fit 99.99% of use cases.


integration/https_test.go
```diff
@@ -1211,3 +1211,38 @@ func (s *SimpleSuite) TestMaxConcurrentStream() {
...
+	frame, err := framer.ReadFrame()
+	require.NoError(s.T(), err)
+
+	fr, ok := frame.(*http2.SettingsFrame)
+	require.True(s.T(), ok)
+
+	maxDecodeHeaderTableSize, ok := fr.Value(http2.SettingHeaderTableSize)
+	assert.True(s.T(), ok)
+	assert.Equal(s.T(), uint32(10000), maxDecodeHeaderTableSize)
+}
```
GCHQDeveloper548: I was not able to find an easy way to test the EncodeHeaderTableSize. I attempted to write something that would create a connection from a client and then try to set two headers of varying size from the server, but the code got much too complex. Happy to take input on ideas for testing the encode side if we think it is necessary.

nmengin:
Hey @GCHQDeveloper548,

Thank you for the contribution.

We've put it on your maintainers' plate.

GCHQDeveloper548:
I have rebased and fixed the CI failure in the linter.

docs/content/reference/install-configuration/entrypoints.md
```diff
@@ -85,6 +85,7 @@ additionalArguments:
...
 |:----------------------------------------------------------------|:------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|:------------------------|:---------|
+<<<<<<< HEAD
```
sdelicata: There's a conflict that needs to be solved here.

integration/https_test.go
sdelicata: Instead of testing the complex HTTP/2 negotiation behavior, WDYT about adding a simpler unit test that verifies the configuration is correctly passed to Go's HTTP/2 server?

IMO, this would be sufficient because we don't want to test Go's HTTP/2 implementation to handle the protocol correctly.

docs/content/reference/install-configuration/entrypoints.md
GCHQDeveloper548: I've resolved this conflict.

GCHQDeveloper548:
Hi @sdelicata , afraid I cannot reply directly to your suggestion about the tests - but your approach sounds good. I have implemented it and fixed the merge conflict. This should be ready for review again now.

kevinpollet (approved):
Thanks 👍

traefiker:

PR #12065, opened 2025-09-15
feat: allow publishing services with type ExternalName

james-callahan:
### What does this PR do?

This allows pointing `PublishedService` at an external hostname (such as an externally managed CNAME) for e.g. external-dns to pick up

### Motivation

I'd like to migrate my existing cluster to use cloudflare tunnels: to do so, we are replacing our existing `type: LoadBalancer` published service with one that has `type: ExternalName, externalName: mytunnel.cfargotunnel.com`

This seems to be currently supported in the `ingress-nginx` provider, but not the normal `ingress` one!

### More

- [x] Added/updated tests


james-callahan:
Failing test seems unrelated?
>     --- FAIL: TestSimpleSuite/TestSimpleOCSP (5.11s)

pkg/provider/kubernetes/ingress/kubernetes_test.go
```diff
@@ -2297,6 +2297,14 @@ func TestIngressEndpointPublishedService(t *testing.T) {
 				},
 			},
 		},
+		{
+			desc: "Published Service ExternalName",
+			expected: []netv1.IngressLoadBalancerIngress{
+				{
+					HostName: "example.com"
```
sdelicata:
```suggestion
					Hostname: "example.com",
```

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -484,6 +484,11 @@ func (p *Provider) updateIngressStatus(ing *netv1.Ingress, k8sClient Client) err
 			}
 		}
 
+	case corev1.ServiceTypeExternalName:
```
sdelicata, line 487: Could you please update the documentation to add `ExternalName` to the supported service types list?

sdelicata (changes requested):
Hi @james-callahan and thanks for your contribution!
I just made a bunch of suggestions.

pkg/provider/kubernetes/ingress/kubernetes.go
james-callahan: done

james-callahan:
> Thanks @james-callahan for your contribution!

Can this be merged now?

sdelicata:
We're now waiting for a second review. And then we can merge it.

PR #12066, opened 2025-09-15
Fix compress excludedContentTypes option

DorianBlues:
### What does this PR do?

This PR fixes an issue with the excludedContentTypes option in the compression middleware, caused by checking the request Content-Type, which could result in ambiguity.

### Motivation

Fixes https://github.com/traefik/traefik/issues/12014

### More

- [x] Added/updated tests


rtribotte:
Hi @DorianBlues,

Thanks for this contribution!

However, as mentioned in my https://github.com/traefik/traefik/issues/12014#issuecomment-3242210598, this is a legacy behavior some may rely on, and we wanted to open a discussion for its removal.
Could you please open a dedicated issue?

PR #12073, opened 2025-09-16
Allow to extract response headers as labels for the `requests_total` metrics

jybp:
### What does this PR do?

An implementation suggestion for https://github.com/traefik/traefik/issues/12072

Adds the ability to specify specific response headers labels in `requests_total` prometheus metrics the exact same way as `metrics.prometheus.headerLabels` works for request headers.

Documentation in 2bddaa6

### Motivation

The `metrics.prometheus.headerLabels` configuration is very helpful to enrich prometheus metrics with useful labels.
Doing the same for response headers would be super helpful. Typical use cases could be `X-Cache-Status` or in our case custom response headers tied to our apps.

### More

- [x] Added/updated tests

### Additional Notes

An alternative implementation could be to introduce a `includeResponseHeaderLabels` (`true`/`false`) instead of `responseHeaderLabels`. It would mean simply merging both `http.Headers` from the `http.Request` and `http.ResponseWriter` in `pkg/middlewares/metrics/metrics.go`. 
But this is less explicit and less flexible imho: Can lead to potential confusion in case of name clashes for example whereas here you can specifically tell from which one it should be looked up.


nmengin:
Hello @jybp,

In the meantime, could you please merge the content of #12074 into this PR? We're used to managing only one PR for both implementation and documentation.
I close #12074 in favor of this one.

jybp:
Thanks @nmengin 
Documentation commit pushed 2bddaa6 on this branch 👍

jybp:
@nmengin @kevinpollet @rtribotte 👋 Hey, sorry for the radio silence on this. I've merged master to fix conflicts. 
Would appreciate if someone could have a look when available to leave feedback / approve for merge.

PR #12077, opened 2025-09-16
ServersTransport: set minimum MaxIdleConnsPerHost=-1

xe-leon:
### What does this PR do?

In version v3.4, improved CEL validation for CRDs was introduced (https://github.com/traefik/traefik/pull/11311). It set minimum value of maxIdleConnsPerHost in ServersTransport CRD to 0.
This PR changes minimum value to -1.

### Motivation

We use `nativeLB` in IngressRoutes to implement topologyAwareRouting between ingress and application. At the same time we want to use RoundRobin load balancing between each pod in the same zone. As mentioned in the comment https://github.com/traefik/traefik/issues/10303#issuecomment-1916465192 , to force traefik not to reuse established connections, maxIdleConnsPerHost should be set to -1. 
I tested the behaviour with maxIdleConnsPerHost=0, but in this case Traefik keeps reusing open connections.


nmengin:
Hello @xe-leon,

In the meantime, as it's a bugfix, could you please rebase your PR on the branch `v3.5`?

xe-leon:
> could you please rebase your PR on the branch v3.5?

done

sdelicata (changes requested):
Hi @xe-leon and thanks for your contribution!
Maybe it could be interesting to improve the `MaxIdleConnsPerHost` description in the dynamic configuration itself:
https://github.com/traefik/traefik/blob/8c23eb683339d8072eb0e4786e24a3ed221d18a6/pkg/config/dynamic/http_config.go#L350
For example: `If negative, disables connection reuse.`

sdelicata (approved):
LGTM
Thx @xe-leon!

PR #12096, opened 2025-09-25
Do not fail when pod is not found in K8sAttributesDetector

xe-leon:
### What does this PR do?

Fix tracing module startup failure in case of running with host network.

### Motivation

When running traefik with `hostNetwork: true` mode, pod inherits the hostname from the node it's running on. 
As a result, K8sAttributesDetector can't find a pod with that name, causing the entire tracing module to fail to start:

```
Unable to create tracer error="building resource: error detecting resource: 
getting pod metadata: pods \"<nodeName>\" not found"
```

Since it returns an empty resource instead of failing when [permissions are missing](https://github.com/traefik/traefik/blob/v3.5.2/pkg/types/k8sdetector.go#L56), it seems we should do the same if the pod isn't found.

As an additional way to avoid failures, I've added retrieving the pod name from the `POD_NAME` environment variable.

### More

- [ ] Added/updated tests

### Additional Notes

Maybe update some docs around [here](https://github.com/traefik/traefik/blob/v3.5.2/docs/content/migrate/v3.md?plain=1#L459), but I think it's not worth mentioning.

<details>
<summary>Current workaround</summary>

* Build `OTEL_RESOURCE_ATTRIBUTES` environment variable on your own:
```yaml
# ... pod manifest
    env:
      - name: POD_NAME
        valueFrom:
          fieldRef:
            fieldPath: metadata.name
      - name: POD_UID
        valueFrom:
          fieldRef:
            fieldPath: metadata.uid
      - name: POD_NAMESPACE
        valueFrom:
          fieldRef:
            fieldPath: metadata.namespace
      - name: OTEL_RESOURCE_ATTRIBUTES
        value: >-
          k8s.pod.name=$(POD_NAME),
          k8s.pod.uid=$(POD_UID),
          k8s.namespace.name=$(POD_NAMESPACE)
```

* All the following actions of trying to get pod metadata will be [skipped](https://github.com/traefik/traefik/blob/v3.5.2/pkg/types/k8sdetector.go#L25)
</details>


nmengin:
Hello @xe-leon,

gootik:
Hi @xe-leon we are running into the same issue. Were you able to find a workaround? until this PR is merged and released

xe-leon:
> Hi @xe-leon we are running into the same issue. Were you able to find a workaround? until this PR is merged and released

Hi, @gootik! Yes, it is hidden under `Current workaround` spoiler in PR description

gootik:
> > Hi @xe-leon we are running into the same issue. Were you able to find a workaround? until this PR is merged and released
> 
> Hi, @gootik! Yes, it is hidden under `Current workaround` spoiler in PR description

amazing! thank you, I completely missed it in the description. 

thanks again

xe-leon:
> Hello @xe-leon and thanks for this contribution,
> 
> We are fine with introducing the `IsNotFound` check to avoid blocking the user when the metadata cannot be determined because of a missing pod.
> 
> Instead of introducing another way to specify the [`POD_NAME`](https://github.com/traefik/traefik/pull/12096/files#diff-92ba44df202dd77a453dc1376739155e36b5d87b7994c1a4257e2f92503dc264R44), we would prefer to enhance the documentation (for example, in the reference documentation for metrics, tracing, and logs in the OTEL section) and keep only the standard way to declare attributes with the `OTEL_RESOURCE_ATTRIBUTES` env variable.

Reverted that things about `POD_NAME`.

Also added some info callouts to describe how resource attrubutes auto-discovery works. 

I think just the right spot for it is `docs/content/observability/tracing/opentelemetry.md`, but I didn't actually find it on [doc.traefik.io](https://doc.traefik.io/traefik), so I've added the same callout in `docs/content/reference/install-configuration/observability/tracing.md`. 

Please let me know if it's wrong.

rtribotte:
Hi @xe-leon,

Thanks for the rework! Would you mind if we push review commits for the documentation?

xe-leon:
> Hi @xe-leon,
> 
> Thanks for the rework! Would you mind if we push review commits for the documentation?

Sure thing! You can push it

Issue #12076, opened 2025-09-16
Gateway honor spec.addresses


Dbzman:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

When setting `providers.kubernetesGateway.statusAddress.hostname` in the helm chart, the Gateway resource gets a `status.addresses` populated.

My expectation was that this would also work if I set `spec.addresses` on a Gateway, so I can have multiple Gateways pointing to different hostnames.

`spec.addresses` in the GatewaySpec sound promising from the description:
> Addresses requested for this Gateway. This is optional and behavior candepend on the implementation. If a value is set in the spec and the requested address is invalid or unavailable, the implementation MUSTindicate this in the associated entry in GatewayStatus.Addresses.The Addresses field represents a request for the address(es) on the"outside of the Gateway", that traffic bound for this Gateway will use.This could be the IP address or hostname of an external load balancer orother networking infrastructure, or some other address that traffic willbe sent to.If no Addresses are specified, the implementation MAY schedule theGateway in an implementation-specific manner, assigning an appropriateset of Addresses.The implementation MUST bind all Listeners to every GatewayAddress thatit assigns to the Gateway and add a corresponding entry inGatewayStatus.Addresses.Support: Extended
--
https://gateway-api.sigs.k8s.io/reference/spec/#gatewayspec

I have a similar use case to this community thread and I use external-dns to provision my dns records:
https://community.traefik.io/t/multiple-gateways-with-traefik-helm-chart/25862

nmengin:
Hello @Dbzman,

Thanks for your suggestion, we think it makes a lot of sense.  Unfortunately, this would not make it to our roadmap for a while as we are focused elsewhere.  If you or another community member would like to build it, let us know, and we will work with you to make sure you have all the information needed so that it can be merged.  

Don’t forget to check out the [contributor docs](https://github.com/traefik/contributors-guide/blob/master/pr_guidelines.md) and to link the PR to this issue.

Dbzman:
Thanks @nmengin, I found some time to create a PR for that, now. :)

Dbzman:
Continuing the discussion here with @kevinpollet from the PR:
> Hello @Dbzman and thanks for your contribution,
> 
[...]

PR #12102, opened 2025-09-26
feat: support spec.addresses on Gateway

Dbzman:
### What does this PR do?

It allows to provide `spec.addresses` on a Gateway and populate it at `status.addresses`, taking priority over a provider-level configuration for the Traefik instance instelf.

### Motivation

We need different Gateways exposing different addresses, so that we don't need to run an additional Traefik instance for each address.

### More

- [x] Added tests
- [x] Added documentation

### Additional Notes

This should solve #12076

### DIFF
--- a/docs/content/providers/kubernetes-gateway.md
+++ b/docs/content/providers/kubernetes-gateway.md
@@ -169,6 +169,10 @@ providers:
 
 ### `statusAddress`
 
+Note: If a Gateway resource specifies `spec.addresses`, Traefik will populate `status.addresses` with the same values.
+The provider-level statusAddress options below are used only as a fallback when `spec.addresses` is empty.
+This enables different Gateways to expose different addresses.
+
 #### `ip`
 
 _Optional, Default: ""_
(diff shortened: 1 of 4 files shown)


Dbzman:
I honestly don't know why those tests fail. After spending a day on getting the tests to run with Tailscale and Docker Desktop, running the failed individual tests gets them passed. They fail though if I run more than one test. Also the gateway compliance tests passes locally. I think I can't do anything more in this PR.

nmengin:
Hey @Dbzman,

Thank you for reaching out

We've just rebased the branch to relaunch the tests.
We've put your Pr on our (maintainers') plate.

We'll keep you updated once we move forward on it.

Dbzman:
Hey @nmengin,

that's great news. Thanks for that! :)

kevinpollet:
Hello @Dbzman and thanks for your contribution,

After reviewing the Gateway API specification, we understand that the `spec.addresses` field in a Gateway is intended for allocating addresses when the corresponding gateway controller instance is provisioned. It appears that this is the required behavior in the conformance tests.

Currently, Traefik does not support this deployment model, and we do not believe using the `spec.addresses` field to achieve your goal is ideal.

Could you please describe what you are trying to achieve in the issue, as well as the tool you are using that requires this behavior?

We are therefore declining this pull request and look forward to continuing the discussion in the corresponding issue.

Issue #11992, opened 2025-08-12
otlp trace is not work on v3.5


gamerslouis:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I'm trying to upgrade Traefik from v3.4 to v3.5 using a Helm chart upgrade from v36.3.0 to v37.0.0. However, the otlp trace is no longer working.

### What did you see instead?

Traefik has stopped appending traceparent headers to requests sent to backend servers, and it's also failing to send spans to the Otel collector.

### What version of Traefik are you using?

v3.5.0

### What is your environment & configuration?

cluster create with `kind create cluster`
traefik installation: `helm install traefik -n traefik traefik/traefik -f traefik.yaml --create-namespace --version 37.0.0` (source: https://traefik.github.io/charts)

value file:
```yaml
tracing:
  addInternals: false
  otlp:
    enabled: true
    http:
      enabled: true
      endpoint: http://simplest-collector.default.svc.cluster.local:4318/v1/traces

logs:
  general:
    level: "DEBUG"
  access:
    enabled: true
    format: json
    fields:
      general:
        names:
          TraceId: keep
```

tracing is work when install with: `helm install traefik -n traefik traefik/traefik -f traefik.yaml --create-namespace --version 36.3.0`

### If applicable, please paste the log output in DEBUG level

2025-08-12T06:07:12Z INF github.com/traefik/traefik/v3/cmd/traefik/traefik.go:108 > Traefik version 3.5.0 built on 2025-07-23T13:57:30Z version=3.5.0
[...]

PR #12103, opened 2025-09-26
fix: otel not working without USER

mmatur:
### What does this PR do?

This PR sets USER env variable if not present to allow otel to start

> Unable to create tracer error="building resource: error detecting resource: user: Current requires cgo or $USER set in environment"

### Motivation

Fixes https://github.com/traefik/traefik/issues/11992

### More

- ~[ ] Added/updated tests~


pkg/logs/otel.go
```diff
@@ -18,6 +19,14 @@ func SetupOTelLogger(ctx context.Context, logger zerolog.Logger, config *types.O
 		return logger, nil
 	}
 
+	if os.Getenv("USER") == "" {
+		// If USER environment variable is not set, we are likely running in a container.
+		// Set it to "traefik" so that some libraries (like OpenTelemetry) do not fail to initialize.
```
sdelicata: This could also happen in other environments. WDYT about:
```suggestion
        // If USER environment variable is not set, set it to "traefik" so that 
        // OpenTelemetry resource detection does not fail to initialize.
```

pkg/metrics/otel.go
```diff
@@ -46,6 +47,14 @@ type SemConvMetricsRegistry struct {
...
 // NewSemConvMetricRegistry registers all stables semantic conventions metrics.
 func NewSemConvMetricRegistry(ctx context.Context, config *types.OTLP) (*SemConvMetricsRegistry, error) {
+	if os.Getenv("USER") == "" {
+		// If USER environment variable is not set, we are likely running in a container.
+		// Set it to "traefik" so that some libraries (like OpenTelemetry) do not fail to initialize.
+		if err := os.Setenv("USER", "traefik"); err != nil {
+			return nil, fmt.Errorf("could not set USER environment variable: %w", err)
+		}
+	}
+
```
sdelicata: Since the same code block is duplicated across three files WDYT about creating a common utility function such as:
```go
func EnsureUserEnvVar() error {
    if os.Getenv("USER") == "" {
        if err := os.Setenv("USER", "traefik"); err != nil {
            return fmt.Errorf("could not set USER environment variable: %w", err)
        }
    }
    return nil
}
```

This would:
- Reduce code duplication
- Make the logic easier to test
- Ensure consistency across all OpenTelemetry initialization points
- Make future changes easier to maintain

sdelicata (changes requested):

james-callahan:
> Fixes #11992

I note that I do *not* see the mentioned log message

> Unable to create tracer error="building resource: error detecting resource: user: Current requires cgo or $USER set in environment"

Yet since upgrading to v3.5.x I have no traces.
Which is to say, I think this PR alone doesn't fix #11992

rtribotte:
@james-callahan Thanks for reporting this!
Did you try the changes from this PR?
Have any other errors been reported, other than the one addressed here?

pkg/udp/conn.go
```diff
@@ -34,7 +34,7 @@ type Listener struct {
 	timeout time.Duration
 }
 
-// Creates a new listener from PacketConn.
+// ListenPacketConn a new listener from PacketConn.
```
sdelicata:
```suggestion
// ListenPacketConn creates a new listener from PacketConn.
```

lbenguigui (approved):
LGTM

PR #12115, opened 2025-10-02
Fixed example code

Dr4K4n:
### What does this PR do?

this was maybe missed when updating the documentation

### Motivation

correct documentation, we noticed it when upgrading to træfik 3

### More

- [ ] Added/updated tests


kevinpollet (approved):
Thanks 👍

PR #12129, opened 2025-10-05
docs: Fix fault markdown rendering

MaBauMeBad:
### What does this PR do?

The current [Documentation](https://doc.traefik.io/traefik/reference/install-configuration/observability/metrics/#labels_2) has a malformatted Markdown table. This PR attempt to fix tis.

### DIFF
--- a/docs/content/reference/install-configuration/observability/metrics.md
+++ b/docs/content/reference/install-configuration/observability/metrics.md
@@ -387,16 +387,16 @@ Here is a comprehensive list of labels that are provided by the metrics:
 
 Here is a comprehensive list of labels that are provided by the metrics:
 
-| <a id="Label" href="#Label" title="#Label">Label</a> | Description     | example       |
-| <a id="row" href="#row" title="#row">------  -----</a> |------------|---------------|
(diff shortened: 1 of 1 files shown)


nmengin:
Hey @MaBauMeBad,

The anchor insertion is automatically done by a script ([gendoc.go](https://github.com/traefik/traefik/blob/v3.5/internal/gendoc.go)).
Modifying only the documentation will have no impact.

Do you think you can take a look at the go script instead?

MaBauMeBad:
HI @nmengin,

thanks for your feedback.
I can take a look in the future maybe, but not at this time (have also no experience with go yet).

From where is the text generated/included ?
I assume the linked script will build the site, but not the docs itself, right ?

nmengin:
Hey @MaBauMeBad,

Indeed, I didn't link the right file.
It's [this one](https://github.com/traefik/traefik/blob/v3.5/internal/anchors.go).

If you cannot fix the issue by yourself in the near future, could you close this PR and open an issue instead?
Thanks in advance.

kevinpollet:
@nmengin @MaBauMeBad we checked the script and it is behaving well, maybe it is a relica of executing another version of the script.

traefiker:

traefiker:

traefiker:

traefiker:

PR #12134, opened 2025-10-07
Fix ingress status if type ClusterIP

walnuss0815:
### What does this PR do?

This PR fixes the Kubernetes ingress status if the traefik service is type `ClusterIP`. The field `.spec.clusterIPs` is now being used to generate the ingress status instead of the `ExternalIPs`. ExternalIPs is for IPs "for which nodes in the cluster will also accept traffic for this service".

Ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.34/#servicespec-v1-core

### Motivation

I am having the same issue as described here https://github.com/argoproj/argo-cd/issues/14607 which should be fixed by https://github.com/traefik/traefik/pull/11100 , but due to the usage of the `externalIPs` field it does not work as intended if the service type is `ClusterIP`.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/ingress/fixtures/Published-Service-ClusterIP.yml
+++ b/pkg/provider/kubernetes/ingress/fixtures/Published-Service-ClusterIP.yml
@@ -7,7 +7,7 @@ metadata:
 
 spec:
   type: ClusterIP
-  externalIPs:
+  clusterIPs:
     - 1.2.3.4
     - 5.6.7.8
 
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -459,7 +459,7 @@ func (p *Provider) updateIngressStatus(ing *netv1.Ingress, k8sClient Client) err
 			})
 		}
 
-		for _, ip := range service.Spec.ExternalIPs {
+		for _, ip := range service.Spec.ClusterIPs {
 			ingressStatus = append(ingressStatus, netv1.IngressLoadBalancerIngress{
 				IP:    ip,
 				Ports: ports,


walnuss0815:
I have built the image and tested it in my Kubernetes cluster. It works like expected and without an issues.

nmengin:
Hello @walnuss0815,

Your PR looks more like a fix. Could you rebase it on the `v3.5` branch please?

walnuss0815:
Hi @nmengin,

I have rebased the fix on the `v3.5` branch.

jkaplowitz:
I had the exact same problem, with my traefik Service running as type ClusterIP and ArgoCD being unable to see anything in the resulting Ingress objects' `status.loadBalancer.ingress` lists to determine that the Ingresses are healthy. I built an image from the current version of this PR (commit 497fe7583981f92253e5a2f049ba0c271286be62) and tried it locally. I can confirm that this fixes it for me.

In case it's relevant, my instance of ArgoCD is running in the same Kubernetes cluster as Traefik and can, if it wants to, route traffic to the traefik Service's cluster IP address without leaving the cluster. I'm not sure this matters though: even if my ArgoCD instance were running externally to the Kubernetes cluster with no ability to route to the cluster IP address of the traefik Service, ArgoCD's health check for an Ingress only cares that the `status.loadBalancer.ingress` list is non-empty with at least one value for `hostname` or `IP`, without actually attempting any form of routing.

walnuss0815:
Hi @nmengin,

is there any update regarding the analysis? because 3.5 isn't the latest release anymore, I have rebased the fix on the v3.6 branch.

kevinpollet:
Hey @walnuss0815,

PR #12144, opened 2025-10-11
Fix ticker cleanup logic

arturmelanchyk:
### What does this PR do?

In Go, time.Ticker objects were not garbage-collected before Go 1.23 — even if you lost all references to them. You had to explicitly call Stop() to release their resources (otherwise the internal goroutine would leak).

This PR fixes `ticker` cleanup logic in case of an error return. It also narrows down the scope where `ticker` is actually used.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/acme/provider.go
+++ b/pkg/provider/acme/provider.go
@@ -248,14 +248,15 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
 
 	p.renewCertificates(ctx, renewPeriod)
 
-	ticker := time.NewTicker(renewInterval)
 	pool.GoCtx(func(ctxPool context.Context) {
+		ticker := time.NewTicker(renewInterval)
+		defer ticker.Stop()
+
 		for {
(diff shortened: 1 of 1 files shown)


pkg/provider/docker/pswarm.go
arturmelanchyk, line 125: previously, on this `return` tickedr wouldn't be stopped.
now it will be stopped in `defer`

rtribotte:
Hello @arturmelanchyk,

We have set the status to 'Review' to enable us to check the pull request.
We will keep you updated once the analysis is complete.

In the meantime, could you please confirm whether the changes address an identified bug?

arturmelanchyk:
Hi @rtribotte , this PR doesn't address any specific ticket/issue.
The goal is to fix [`Ticker` leak here](https://github.com/traefik/traefik/pull/12144/files#diff-0366da00c96588150d65ca0225f09c77c77a69ab86a0e4edb7d27b9b7f5311fdR125)

PR #12167, opened 2025-10-17
Add least time load balancing strategy

sdelicata:
### What does this PR do?

Adds a new `leasttime` load balancing strategy that selects backend servers based on average response time (TTFB) and active connection count, weighted by server capacity.

### Motivation

Traefik's existing strategies (WRR, P2C, HRW) don't consider response time/latency as a selection criterion.
The `leasttime` strategy routes traffic to the fastest-responding servers while maintaining fair load distribution.
It is particularly interesting when:
- Backend servers have heterogeneous performance (different hardware, locations, or load)
- Servers experience performance degradation (sustained load, network latency changes)
- Multi-datacenter deployments with varying RTT

### More

- [x] Added/updated tests


pkg/server/service/loadbalancer/leasttime/leasttime.go
```diff
@@ -0,0 +1,389 @@
...
+func (b *Balancer) selectWRR(candidates []*namedHandler) *namedHandler {
+	if len(candidates) == 0 {
+		return nil
+	}
+
+	var selected *namedHandler
+	minDeadline := math.MaxFloat64
+
+	// Find handler with earliest deadline.
+	b.deadlineMu.RLock()
```
lbenguigui: Can you explain why you use a lock since there is no `b` attribute used

sdelicata: The lock protects the balancer state, the deadline field of each namedHandler in the candidates slice.

Even though these handlers are accessed through the loop variable h, they're shared objects that can be concurrently accessed by multiple goroutines calling ServeHTTP() => nextServer() => selectWRR().

That's why the read lock (RLock) allows multiple concurrent readers during the selection phase, while preventing any writer from modifying deadline values. Then it upgrades to a write lock (Lock) to safely update the selected handler's deadline.

There's a comment at line 103:
```
// deadlineMu protects EDF scheduling state (curDeadline and all handler deadline fields).
// Separate from handlersMu to reduce lock contention during tie-breaking.
```

Feel free to complete or modify it if you think it's not clear enough.

lbenguigui: I think it would be cleaner if the handler owned the responsibility for managing its deadline state.
We could introduce a dedicated mutex and `getDeadline/updateDeadline` methods, and rename `deadlineMu` to `curDeadlineMu`.
This would improve maintainability, keep responsibilities well defined, and align better with how `responseTime` is handled.

rtribotte: @lbenguigui Done

rtribotte (approved):
LGTM

traefiker:

traefiker:

Issue #10937, opened 2024-07-25
Internal Node IPs for NodePort services feature breaks namespace scoped Traefik deployments


gites:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I'm running my Traefik deployment on Kubernetres in a fully name-spaced way. All Traefik RBAC permissions are configured using `Roles` and `RoleBindings`. The K8s cluster is owned by an infrastructure team and host multiple namespaces for multiple teams/clients. Until now I was able to run and configure Traefik permissions without any support from the infra team. 

With the [Internal Node IPs for NodePort services](https://github.com/traefik/traefik/pull/10278) feature introduced in Treafik 3.1 I no longer can do it. Traefik 3.1 requires cluster level permissions to list/watch nodes. This makes both me (as I need to ask/deploy an external team to manage/deploy my Traefik configuration) and the infrastructure team (that operate at a big scale and don't want to be bothered with an application level configurations) unhappy.

As stated in the original PR comment by @rtribotte nodes are not namespaced so I fully understand why it was implemented this way. But the possibility to run Traefik in a fully namespaced way is/was a cool and useful feature that, I think, should be kept. 
 
@rtribotte are you open to refactoring the code to fetch node resources in the service namespace or other solution that will allow for fully namespaced Traefik setups?

Also the documentation for [3.0 -> 3.1 migration](https://doc.traefik.io/traefik/migration/v3/) is missing any information about the new permissions requirements 

### What did you see instead?

Traefik 3.1 can't run in Kubernetes without cluster level permissions for listing nodes

### What version of Traefik are you using?

```
traefik version
Version:      3.1.0
Codename:     comte
Go version:   go1.22.5
Built:        2024-07-15T14:44:04Z
OS/Arch:      linux/amd64
```

### What is your environment & configuration?

Traefik RBAC configuration:

```yaml
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: traefik
  namespace: {{ .Release.Namespace }}
---
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: traefik-service-discovery
  namespace: {{ .Release.Namespace }}
rules:
  - apiGroups: [""]
    resources: ["endpoints", "pods", "secrets", "services","nodes"]
[...]

PR #10278, opened 2023-12-03
User internal ip from nodes when service is nodeport

jorisvergeer:
### What does this PR do?

This PR causes traefik to use the ip adresses of the node insead of the pods when kubernetes services are of type NodePort. In this case it also uses the services nodePort instead of the port.

### Motivation

I have a mixed cloud environment with some micro services running in k8s and some not. I run traefik externally in a separate instance. Using the kubernetesCRD provider, treafik is able to detect the services. The only problem is that the IP addresses that are detected are the internal pod ip adresses. And those are not accessible from the instance running traefik. 

The services are accessible from the node ip when the service type is NodePort. In this case traefik should be able to access the services using the node ip and the nodeport of the service. 

This feature would help me to assist in a gradual migration from native services to traefik and k8s. 

### More

- [ ] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

This might have impact on users that accidently (or purposely) use nodeport as service type. 

Is this a feature that will even be considered?

Can this feature also be merged into the 2.x branch?

PR #12048, opened 2025-09-08
Add support for ServiceExternalTrafficPolicyTypeLocal

deveshk0:
### What does this PR do?

- Implement node filtering for Local external traffic policy in HTTP, TCP, and UDP services
- Filter nodes to only include those with ready endpoints when externalTrafficPolicy is Local
- Add EndpointSlice integration to determine node readiness for Local traffic policy
- Support multi-protocol NodePort services with Local external traffic policy
- Add comprehensive test cases covering HTTP, TCP, and UDP scenarios
- Ensure non-ready endpoints are excluded from load balancer configuration

This enables proper load balancing behavior for Kubernetes services with externalTrafficPolicy set to Local, ensuring traffic is only routed to nodes that have ready endpoints for the service.

closes https://github.com/traefik/traefik/issues/12023

### Motivation

https://github.com/traefik/traefik/issues/12023

### More

- [x] Added/updated tests
- [x] Added/updated documentation

Issue #12023, opened 2025-08-27
Kubernetes NodePort service with nodePortLB does not work with externalTrafficPolicy Local


wouterh-dev:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

When using traefik with a NodePort service with `externalTrafficPolicy: Local` and a `nodePortLB` `IngressRoute` traefik will always route traffic to all nodes, instead of only the nodes running the pods.

I am running traefik outside of a kubernetes cluster and would like to avoid an unnecessary network hop when a pod is running on 1 out of many nodes. So I set `externalTrafficPolicy: Local` (https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/):

> Local preserves the client source IP and avoids a second hop for LoadBalancer and NodePort type Services, but risks potentially imbalanced traffic spreading.

Traefik appears to balance traffic between all nodes, instead of only the nodes running pods matched by the service. But some of these nodes are not responding to traffic, so the connection just hangs until:

```
› curl http://10.250.32.225:80
Gateway Timeout
```

It would be nice if traefik would instead check if the nodes are accepting tcp connections, or was aware of which nodes to send traffic to based on the EndpointSlices for the Service with `externalTrafficPolicy: Local`

### What did you see instead?

To reproduce create a Deployment, Service & IngressRoute on a cluster with at least 2 nodes:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: internal-http-nginx-deployment
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: nginx
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx
        name: nginx
        ports:
        - containerPort: 80
          protocol: TCP
---
apiVersion: v1
kind: Service
metadata:
  name: internal-http-nginx-service
  namespace: default
spec:
  externalTrafficPolicy: Local
  internalTrafficPolicy: Cluster
  ports:
  - name: http
    port: 80
    protocol: TCP
    targetPort: 80
  selector:
    app: nginx
  type: NodePort
---
apiVersion: traefik.io/v1alpha1
[...]

PR #12175, opened 2025-10-21
NodePortLB support in kubernetes gateway provider

jorisvergeer:
### What does this PR do?

This PR adds NodePort Load Balancing capabilities to the kubernetes gateway provider. Similar to how this is implemented in kubernetes CRD provider and kubernetes Ingress provider.

### Motivation

I run traefik outside my cluster on its own instance, because i have some services running outside Kubernetes and some inside. 
Therefore back in the early days of traefik 2.10/3.0 I implemented NodePort loadbalancing #10278.

Now that the gateway api is widely available, i kind of prefer that way to defining routes. If only traefik supported NodePort load balancing in the gateway provider.... well... now it does.

### More

- [X] Added/updated tests

### Additional Notes

After the #10278 PR there was an issue with RBAC to access node resources (Issue #10937). I updated the RBAC for the gateway api, added a note in the documentation migration page and added an option to disable the lookup of node resources altogether (disabling nodeport load balancing)

I understand the RBAC in k8s, but i don't know the full extent what the links are between each part RBAC role, binding and traefik. So, I'd like a second pair of eyes to have a look at this to not have the same or similar issue this time. 

#### To consider

To mitigate the additional access to node resources, this feature might also be an opt in (providers.kubernetesgateway.enablenodeportlb), but this would be different than the other implementations of nodePort load balancing.


pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -912,6 +914,58 @@ func (p *Provider) getBackendAddresses(namespace string, ref gatev1.BackendRef)
...
+		}
+
+		if len(nodes) == 0 {
+			return nil, corev1.ServicePort{}, errors.New("nodes not found")
+		}
+
+		uniqAddresses := map[string]struct{}{}
+		var backendServers []backendAddress
+
+		for _, node := range nodes {
```
james-callahan, line 943: If `externalTrafficPolicy` is `Local` then you shouldn't include all nodes?

nmengin:
Hey @jorisvergeer,

pkg/provider/kubernetes/gateway/kubernetes.go
jorisvergeer: That is similar to #12023 and PR #12048, that has not been resolved/merged yet.
So unsure if that's a wanted feature.

PR #12193, opened 2025-10-23
Add Traefik Hub demo in dashboard

gndz07:
### What does this PR do?

Adds a Traefik Hub demo to the WebUI dashboard to showcase Hub's API management capabilities.

The demo displays Hub-specific content using a dynamically loaded remote web component. For security, the remote bundle is cryptographically signed, and signature verification must pass before the script executes. Verification runs in a Web Worker to isolate it from the main thread and prevent blocking the UI.

Improvement is also made on the Page layout rendering, to avoid the side navigation bar from getting re-rendered every time user changing the path.

<details>
<summary>Preview</summary>

Dashboard homepage:
image

Hub demo:
image

</details>

### Motivation

This demo provides users with a fully interactive live preview of Traefik Hub's API management features directly within the dashboard.

### More

- [x] Added/updated tests

### DIFF
--- a/webui/src/pages/hub-demo/HubDashboard.spec.tsx
+++ b/webui/src/pages/hub-demo/HubDashboard.spec.tsx
@@ -1,7 +1,7 @@
 import { waitFor } from '@testing-library/react'
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
 
-import HubDashboard from './HubDashboard'
+import HubDashboard, { resetCache } from './HubDashboard'
 import verifySignature from './workers/scriptVerification'
 
 import { renderWithProviders } from 'utils/test'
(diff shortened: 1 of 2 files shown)


emilevauge:
A bit more color on this.

This PR introduces a new demo section to the Traefik dashboard that showcases [Traefik Hub's advanced API management](https://traefik.io/traefik-hub) capabilities. The demo provides users with a hands-on preview of key features using sample data in a safe, non-impactful environment.
A new "API Management" menu item (collapsed by default) gives access to the demo section, which demonstrates core Traefik Hub capabilities including API publishing workflows, developer portals, gateway management, analytics and monitoring dashboards. Those views are populated with realistic sample data to provide meaningful interactions. Users can freely navigate and interact with all demo features.

This demo maintains a clean dashboard interface while keeping demo easily accessible through the collapsed menu. It runs entirely on sample data, within a separate web component, with no effect on actual Traefik resources or configurations.

→ If you are interested in more advanced API capabilities, you now have a way to discover Traefik Hub and interact with it.
→ If you are not interested in this, you can ignore this new menu item at the bottom and continue to enjoy Traefik 😎 

Side note, the job done by @gndz07 on the cryptographic aspects is amazing! More on this soon.

rtribotte (approved):
LGTM

webui/src/pages/hub-demo/workers/scriptVerificationWorker.ts
```diff
@@ -0,0 +1,176 @@
...
+      return base64Decoded
+    }
+  } catch (e) {
+    console.error(e)
+  }
+
+  // hex decoding
+  if (signatureText.length === 128 && /^[0-9a-fA-F]+$/.test(signatureText)) {
+    const hexDecoded = new Uint8Array(64)
+    for (let i = 0; i < 64; i++) {
```
emilevauge: Can't we use [Uint8Array.fromHex(string)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromHex) here?

webui/src/pages/hub-demo/workers/scriptVerificationWorker.ts
```diff
@@ -0,0 +1,176 @@
...
+  const binaryString = atob(base64)
+  const bytes = new Uint8Array(binaryString.length)
+  for (let i = 0; i < binaryString.length; i++) {
+    bytes[i] = binaryString.charCodeAt(i)
+  }
+  return bytes.buffer
+}
+
+function extractEd25519PublicKey(spkiBytes: Uint8Array): Uint8Array {
+  if (spkiBytes.length !== 44) {
```
emilevauge, line 29: This part looks a bit suspicious. Are we 100% sure this length is standard and will be always the same?

webui/src/pages/hub-demo/workers/scriptVerificationWorker.ts
```diff
@@ -0,0 +1,176 @@
...
+import { sha512 } from '@noble/hashes/sha2.js'
+
+// Set up SHA-512 for @noble/ed25519 v3.x
+ed25519.hashes.sha512 = sha512
+ed25519.hashes.sha512Async = (m) => Promise.resolve(sha512(m))
+
+function base64ToArrayBuffer(base64: string): ArrayBuffer {
+  const binaryString = atob(base64)
+  const bytes = new Uint8Array(binaryString.length)
+  for (let i = 0; i < binaryString.length; i++) {
```
emilevauge: Can't we use [Uint8Array.fromBase64()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/fromBase64) here?

webui/src/pages/hub-demo/workers/scriptVerificationWorker.ts
gndz07: We certainly could, but it's only [supported on the most recent browser versions](https://caniuse.com/?search=Uint8Array); I'm afraid that the support is pretty limited for the older versions (for example on Chrome, it's only supported from version 140). What I could try to do is to use Uint8Array but also provide a fallback in case the browser doesn't support this method yet.

webui/src/pages/hub-demo/workers/scriptVerificationWorker.ts
gndz07: For the key format the we use (ed25519), yes it's a standard length, which comprised of 32 bytes of the raw key + 12 bytes SPKI prefix ([reference](https://github.com/openssl/openssl/issues/8960#issuecomment-494041688)). This verification might not pass if we receive only the raw 32 bytes key, but as when we imported the key we asked specifically for 'spki' (line 32), this should be the standard length and gives a strict validation for our use case.

webui/src/pages/hub-demo/workers/scriptVerificationWorker.ts
gndz07: Same answer as above, we could but most probably still need a fallback for older browsers.

webui/src/pages/hub-demo/workers/scriptVerificationWorker.ts
emilevauge: Oh ok, indeed, the support is pretty new.

PR #12201, opened 2025-10-26
Clarify health check requirement for server status metric

asafm:
### What does this PR do?
Updated the description for `traefik_service_server_up` metric to specify that it is only applicable for services configured with health checks.

### Motivation

I couldn't understand why I can't see the `[traefik_service_server_up](https://doc.traefik.io/traefik/reference/install-configuration/observability/metrics/#opt-traefik-service-server-up)` metric. I had to discover why via a github issue comment.

### More


traefiker:

traefiker:

PR #12216, opened 2025-10-30
Fix default encodings in compress middleware

Belphemur:
### What does this PR do?

Seems an older version than the current one was pulled in the new documentation.

### Motivation

Because of https://github.com/traefik/traefik/issues/10859 we reverted zstd as default compression because OOM.

### More

- [x] Added/updated tests


kevinpollet (approved):
Thanks 👍

PR #12224, opened 2025-11-03
feat: add UDP Proxy Protocol support

### LINKED ISSUE #8613: Proxy Protocol for UDP
### Welcome!

I'm trying to set up a PowerDNS server running completely behind traefik proxy. From version 4.6.0 upwards, PowerDNS supports Proxy Protocol headers, which would be quite essential in the DNS UDP world (to support AXFR requests among other things). PowerDNS currently is unable to see the remote IP on UDP packets, as it only sees traefiks IP.

Similarly to #5371, I'd like to be able to set proxyProtocol on udp services.

sdelicata:
### What does this PR do?

Adds UDP Proxy Protocol v2 support to Traefik, enabling preservation of client source IP and port information when Traefik operates behind a proxy or load balancer for UDP traffic.

### Motivation

Fixes #8613 and enables use cases where client IP preservation is essential.
Without this feature, UDP services behind Traefik see only the proxy's IP address, which breaks IP-based authentication, rate limiting, etc.

### More

- [x] Added/updated tests

### Additional Notes

Implementation details:
- This PR parses Proxy Protocol header from the first UDP packet before session creation.
- It uses the client address from Proxy Protocol header for session keying while routing responses to the actual packet source.
- It also validates source IP against TrustedIPs before accepting headers (opt-in Insecure mode available for testing)

Limitations:
  - Only Proxy Protocol v2 supported (binary format, not v1 text format).
  - Header must be present in first packet only, subsequent packets in the session flow through normally.
  - Backend services must be compatible with receiving traffic without Proxy Protocol headers.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -326,6 +326,10 @@ linters:
         text: 'appendAssign: append result not assigned to the same slice'
         linters:
           - gocritic
+      - path: pkg/udp/proxyprotocol_test.go
+        text: 'appendAssign: append result not assigned to the same slice'
+        linters:
+          - gocritic
     paths:
       - pkg/provider/kubernetes/crd/generated/
 
--- a/integration/udp_test.go
+++ b/integration/udp_test.go
@@ -209,7 +209,11 @@ func proxyProtoUDPRequest(address, srcIP string, srcPort int) (string, error) {
 	}
 
 	// Read response
-	conn.SetReadDeadline(time.Now().Add(2 * time.Second))
+	err = conn.SetReadDeadline(time.Now().Add(2 * time.Second))
+	if err != nil {
+		return "", err
+	}
+
 	buf := make([]byte, 2048)
 	n, err := conn.Read(buf)
 	if err != nil {
@@ -235,8 +239,7 @@ func proxyProtoUDPRequestMultiPacket(address, srcIP string, srcPort, numPackets 
 	defer conn.Close()
 
 	responses := make([]string, 0, numPackets)
-
-	for i := 0; i < numPackets; i++ {
+	for i := range numPackets {
 		var packet []byte
 
 		if i == 0 {
@@ -274,7 +277,11 @@ func proxyProtoUDPRequestMultiPacket(address, srcIP string, srcPort, numPackets 
 		}
 
 		// Read response.
-		conn.SetReadDeadline(time.Now().Add(2 * time.Second))
+		err = conn.SetReadDeadline(time.Now().Add(2 * time.Second))
+		if err != nil {
+			return nil, err
+		}
+
 		buf := make([]byte, 2048)
 		n, err := conn.Read(buf)
 		if err != nil {
--- a/pkg/udp/proxyprotocol_test.go
+++ b/pkg/udp/proxyprotocol_test.go
@@ -61,7 +61,7 @@ func TestBuildProxyProtocolV2Header_IPv4(t *testing.T) {
 	header := buildProxyProtocolV2Header(t, "192.0.2.1", 12345, "198.51.100.1", 5000, false)
 	require.NotNil(t, header)
 	// Proxy Protocol v2 header for UDP/IPv4: 16 byte preamble + 8 bytes addresses + 4 bytes ports = 28 bytes.
-	assert.Equal(t, 28, len(header))
+	assert.Len(t, header, 28)
 }
 
 // TestBuildProxyProtocolV2Header_IPv6 verifies the helper creates valid IPv6 headers.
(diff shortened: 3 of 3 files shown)


pkg/udp/conn.go
```diff
@@ -169,70 +211,177 @@ func (l *Listener) readLoop() {
...
 }
 
 // Conn represents an on-going session with a client, over UDP packets.
 type Conn struct {
-	listener *Listener
-	rAddr    net.Addr
+	listener          *Listener
+	clientAddr        net.Addr // Client address for application logic (from Proxy Protocol header or network).
+	responseAddr      net.Addr // Network address for sending responses.
+	usedProxyProtocol bool     // Indicates if this connection used Proxy Protocol.
```
lbenguigui: `usedProxyProtocol` is only used to determine whether we need to delete an element from `connsPacketSourceToClient`.
However, since it’s safe to delete an item even if it doesn’t exist, I think `usedProxyProtocol` isn’t very useful.

pkg/udp/conn.go
```diff
@@ -327,8 +485,14 @@ func (c *Conn) close() {
 func (c *Conn) Close() error {
 	c.close()
 
-	c.listener.mu.Lock()
-	defer c.listener.mu.Unlock()
-	delete(c.listener.conns, c.rAddr.String())
+	c.listener.connsMu.Lock()
```
lbenguigui: What do you think about creating a `removeConn` method on the listener? It would fix the responsibility leak and improve encapsulation.

pkg/udp/conn.go
```diff
@@ -169,70 +211,177 @@ func (l *Listener) readLoop() {
...
+
+			if header != nil {
+				usedProxyProtocol = true
+
+				// Use header's source address for session keying.
+				srcAddr, _, _ := header.UDPAddrs()
+
+				// Record mapping of packet source to client address for subsequent packets.
+				l.connsMu.Lock()
+				l.connsPacketSourceToClient[packetSource.String()] = srcAddr.String()
```
lbenguigui: Why not use a map[string]*net.UDPAddr? It will avoid this back and forth in string

pkg/udp/conn.go
```diff
@@ -305,11 +459,15 @@ func (c *Conn) Read(p []byte) (int, error) {
 // Each call sends at most one datagram.
 // It is an error to send a message larger than the system's max UDP datagram size.
 func (c *Conn) Write(p []byte) (n int, err error) {
+	if c.responseAddr == nil {
```
lbenguigui: Unless I missed something, `responseAddr` cannot be nil. If you want to keep this, you should do it in `Close`method too

lbenguigui (changes requested):

pkg/udp/conn.go
sdelicata: You're completely right. Good catch!

4310V343k:
I'm a little clueless, and I think others would be interested in this as well
(I haven't looked at the code yet)
You specified that you key the session with the client address _from the header_, but how do you tie subsequent packets to that session, if the header is only present in the first packet? Do you rely on the source (not client) ip and port being the same?
This, and the fact that only the first packet should have the header, "requires" the source _and_ the **_udp_** communication (is there even a prolonged communication? Maybe it's just one packet?) to be **_stateful_**
In quotes because the communication _could_ be stateless, but i would expect this implementation to use more resources because of the ever-growing session map.
Maybe a compromise would be a switch between a current stateful (header only in first packet) and a stateless (header in each packet) implementation?

Everything above is just my thoughts, I just wanted to share them if they're helpful in any way

P.S. I hope I've conveyed my thoughts well, and I'm sorry if not. English is not my main language. Please, do ask me if there's anything missing, I'm happy to discuss things

pkg/udp/conn.go
```diff
@@ -169,70 +211,178 @@ func (l *Listener) readLoop() {
...
+}
+
+// getConn returns the ongoing session with clientAddr if it exists, or creates a new one otherwise.
+// The clientAddr is used for session keying, while responseAddr is used
+// for sending responses back to the packet source.
+func (l *Listener) getConn(clientAddr, responseAddr net.Addr) (*Conn, error) {
+	l.connsMu.Lock()
+	defer l.connsMu.Unlock()
+
+	conn, ok := l.conns[clientAddr.String()]
```
lbenguigui, line 342: If you look at the shutdown test, you’ll see that it tries to create a new session after the shutdown. The goal is to verify that the existing session remains usable, while creating a new one should no longer be possible.

However, since we reuse connections based on the `clientAddr`, the attempt to create a new session ends up reusing the existing connection, which causes the test to fail.
We should investigate how this test managed to pass in the previous version, to confirm whether this is regression or not.

PR #12238, opened 2025-11-07
Split Expose User Guides & Add Multi-Layer Routing Section

sheddy-traefik:
### What does this PR do?

This PR adds a new multi layer routing section to the expose guide, splits the expose guides into two levels (basic & advanced) and deletes the old pages 

### More

### DIFF
--- a/docs/content/expose/kubernetes/advanced.md
+++ b/docs/content/expose/kubernetes/advanced.md
@@ -649,15 +649,26 @@ Now create the middleware and IngressRoutes for multi-layer routing:
 
 ```yaml
 # mlr-ingressroute.yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: auth-secret
+  namespace: default
+type: Opaque
+stringData:
+  users: |
+    admin:$apr1$DmXR3Add$wfdbGw6RWIhFb0ffXMM4d0
+    user:$apr1$GJtcIY1o$mSLdsWYeXpPHVsxGDqadI.
+---
 apiVersion: traefik.io/v1alpha1
 kind: Middleware
 metadata:
-  name: auth-headers
+  name: auth-middleware
   namespace: default
 spec:
-  headers:
-    customRequestHeaders:
-      X-User-Role: "admin"  # In production, use ForwardAuth middleware
+  basicAuth:
+    secret: auth-secret
+    headerField: X-Auth-User
 ---
 apiVersion: traefik.io/v1alpha1
 kind: IngressRoute
@@ -671,7 +682,7 @@ spec:
   - match: Host(`api.docker.localhost`) && PathPrefix(`/api`)
     kind: Rule
     middlewares:
-    - name: auth-headers
+    - name: auth-middleware
   # Note: No services and no TLS config - this is a parent IngressRoute
 ---
 apiVersion: traefik.io/v1alpha1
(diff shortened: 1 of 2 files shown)


docs/content/expose/docker/advanced.md
```diff
@@ -438,28 +215,195 @@ You should see different `Hostname` values in these responses, as each request i
...
+### Setup Multi-Layer Routing with Docker
+
+To use multi-layer routing with Docker, you need to enable the File provider alongside the Docker provider.
+
+Update your Traefik service in `docker-compose.yml`:
+
+```yaml
+services:
+  traefik:
+    image: "traefik:v3.4"
```
nmengin:
```suggestion
    image: "traefik:latest"
```

nmengin: Could you please bring this modification everywhere in this page?

docs/content/expose/docker/advanced.md
```diff
@@ -438,28 +215,195 @@ You should see different `Hostname` values in these responses, as each request i
...
+      rule: "HeadersRegexp(`X-User-Role`, `user`)"
+      service: user-backend@docker  # Reference Docker service
+      parentRefs:
+        - api-parent@file  # Explicit reference to parent in file provider
+
+  middlewares:
+    auth-middleware:
+      headers:
+        customRequestHeaders:
+          X-User-Role: "admin"  # In production, use ForwardAuth middleware
```
nmengin: Could you use the BASIC Auth middleware instead, same as with other providers?

nmengin (changes requested):
Hey @sheddy-traefik,

Thank you for the PR.
Below are my comments.

docs/content/expose/docker/advanced.md
xxsheddy: Hey @nmengin the multi-layer routing example in this section relies on setting a custom X-User-Role header that child routers use to match and route requests to different backends (admin vs user).

BasicAuth middleware only handles credential validation - it doesn't inject custom headers into the request like ForwardAuth does with authResponseHeaders.

Could you clarify which approach you'd prefer?

- Keep ForwardAuth for this specific example since it demonstrates header-based routing that requires injecting X-User-Role
- Replace the entire multi-layer routing example with a simpler authentication pattern that uses BasicAuth (but this would change the purpose of the section)
- Add BasicAuth as an additional example alongside ForwardAuth, showing it as a simpler alternative when header-based routing isn't needed

nmengin: Hey @sheddy-traefik,

Thank you for your feedback.

Actually, the BasicAuth middleware allows users to set [the authenticated user in a header](https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/basicauth/#opt-headerField).
I had this scenario in mind.

It's more user-centric than role-centric, but IMO it's good enough for the example.
WDYT?

xxsheddy: Interesting, I didn't know this, I'll update the example then

xxsheddy: PR has been update here: https://github.com/traefik/traefik/pull/12238/commits/ddaf39ddd835d67779496bb92a90d493909247cc

docs/content/expose/kubernetes/advanced.md
```diff
@@ -986,29 +548,282 @@ You should see different `Hostname` values in these responses, as each request i
...
+# mlr-ingressroute.yaml
+apiVersion: traefik.io/v1alpha1
+kind: Middleware
+metadata:
+  name: auth-headers
+  namespace: default
+spec:
+  headers:
+    customRequestHeaders:
+      X-User-Role: "admin"  # In production, use ForwardAuth middleware
```
nmengin: Hey @sheddy-traefik,

As mentioned in my [latest comment](https://github.com/traefik/traefik/pull/12238#discussion_r2643451952), I'd like to use the BASIC Auth middleware in every example (to be homogeneous).

PR #12258, opened 2025-11-11
Fix multi-layer routing with models

juliens:
### What does this PR do?

This PR fixes a problem with multi-layer routing when models are used.
It updates the parentRefs to take into account the new routers created by the model engine.

### Motivation
Fix the bug.

### More

- [x] Added/updated tests


pkg/server/aggregator.go
```diff
@@ -243,6 +247,26 @@ func applyModel(cfg dynamic.Configuration) dynamic.Configuration {
...
 		}
 
+		for _, rt := range rts {
+			if rt.ParentRefs == nil {
+				continue
+			}
+
+			var parentRefs []string
+			for _, ref := range rt.ParentRefs {
+				// Only add the initial parent ref if it still exists (at least one entryPoint had no model).
```
kevinpollet:
```suggestion
				// Only add the initial parent ref if it still exists.
```

kevinpollet (approved):

traefiker:

traefiker:

PR #12278, opened 2025-11-14
Add CertificateTimeout ACME configuration option

ceko:
### What does this PR do?

This feature allows each configured `certificateResolver` to specify a finalization (sometimes referred to as certificate) timeout. The configured value is passed to the Lego ACME client library and is used to determine how long the the client waits for an order to transition from ready -> valid.

```yaml
certificatesResolvers:
  prod:
    acme:
      email: certs@acme.com
      certificateTimeout: "2m"
```

### Motivation

The default Lego ACME client timeout is 30 seconds. This is too short for some certificate authorities and can cause intermittent timeouts. Traefik already passes through several configuration values to the Lego ACME client so it seemed like an easy fix with code patterns already available.

### More

- [ ] Added/updated tests

### Additional Notes

This type of problem is present in most ACME clients that develop towards LetsEncrypt as a baseline. I created this change both to allow our Traefik instances to successfully request certificates and to contribute back a longstanding community request.

I did not include tests because it's inherently difficult to write a timeout test. It usually slows down the suite and tests for config passing seem rote instead of useful. I am willing to write tests if necessary.

I am willing to update documentation if this request gains traction. I'd like to agree on a configuration key name first.


pkg/provider/acme/provider.go
```diff
@@ -55,6 +55,7 @@ type Configuration struct {
 
 	ClientTimeout               ptypes.Duration `description:"Timeout for a complete HTTP transaction with the ACME server." json:"clientTimeout,omitempty" toml:"clientTimeout,omitempty" yaml:"clientTimeout,omitempty" label:"allowEmpty" file:"allowEmpty" export:"true"`
 	ClientResponseHeaderTimeout ptypes.Duration `description:"Timeout for receiving the response headers when communicating with the ACME server." json:"clientResponseHeaderTimeout,omitempty" toml:"clientResponseHeaderTimeout,omitempty" yaml:"clientResponseHeaderTimeout,omitempty" label:"allowEmpty" file:"allowEmpty" export:"true"`
+	FinalizationTimeout         ptypes.Duration `description:"Defines the total time the ACME client will wait for the order to finalize." json:"finalizationTimeout,omitempty" toml:"finalizationTimeout,omitempty" yaml:"finalizationTimeout,omitempty" export:"true"`
```
mmatur:
```suggestion
	FinalizationTimeout         ptypes.Duration `description:"Defines the total time the ACME client will wait for the order to finalize." json:"finalizationTimeout,omitempty" toml:"finalizationTimeout,omitempty" yaml:"finalizationTimeout,omitempty" label:"allowEmpty" file:"allowEmpty" export:"true"`
```

mmatur (changes requested):
Thanks for your contribution @ceko 

Could you also add documentation?

pkg/provider/acme/provider.go
ceko: Thanks for the reply, I've added the two suggested annotations to my configuration option.

ceko:
I changed `FinalizationTimeout` to `CertificateTimeout` to better align with the [LEGO options](https://go-acme.github.io/lego/usage/cli/options/). I also added documentation. Let me know if you'd like anything else to support this change.

docs/content/reference/install-configuration/tls/certificate-resolvers/acme.md
```diff
@@ -91,6 +91,7 @@ ACME certificate resolvers have the following configuration options:
 | <a id="opt-acme-certificatesDuration" href="#opt-acme-certificatesDuration" title="#opt-acme-certificatesDuration">`acme.certificatesDuration`</a> | The certificates' duration in hours, exclusively used to determine renewal dates. | 2160 | No |
 | <a id="opt-acme-clientTimeout" href="#opt-acme-clientTimeout" title="#opt-acme-clientTimeout">`acme.clientTimeout`</a> | Timeout for HTTP Client used to communicate with the ACME server. | 2m | No |
 | <a id="opt-acme-clientResponseHeaderTimeout" href="#opt-acme-clientResponseHeaderTimeout" title="#opt-acme-clientResponseHeaderTimeout">`acme.clientResponseHeaderTimeout`</a> | Timeout for response headers for HTTP Client used to communicate with the ACME server. | 30s | No |
+| <a id="opt-acme-certificateTimeout" href="#opt-acme-certificateTimeout" title="#opt-acme-certificateTimeout">`acme.certificateTimeout`</a> | Timeout for obtaining the certificate during the finalization request. Set this if the ACME server is slow to issue a certificate. | 30s | No |
```
rtribotte, line 94: Hello @ceko,

The file is malformed, the validate job on the CI fails, a `go generate` should fix this issue.
Could you please update the PR in that regard, and also rebase the changes? Thanks!

PR #12283, opened 2025-11-17
Fix loadbalancer doc for highest random weight

ozon2:
### What does this PR do?

The current documentation on how to use the highest random weight in a loadbalancer incorrectly mentions a "type" field where "strategy" should be used.

### Motivation

I tried using highest random weight with the example from the documentation and it did not work.


kevinpollet (approved):
Thanks 👍

PR #12223, opened 2025-11-03
Add documentation for loadbalancer.server.url in Docker provider

webash:
Added information about specifying a URL for the service in Docker configuration. This is per #11374 and #8753

### What does this PR do?

Updates the documentation to specify that `loadbalancer.server.url` is available as a label option for the Docker provider.

### Motivation

It was added in #11374 but the documentation wasn't updated. What is confusing, is that [this page](https://doc.traefik.io/traefik/reference/routing-configuration/http/load-balancing/service/#configuration-example) documents `.loadbalancer.servers[0].url` as an option in the labels example, but that doesn't work (in Swarm, at least).

### More

- [ ] Added/updated tests
- [ ] Added/updated documentation

PR #11374, opened 2024-12-27
Allow url to be used in defining a service through docker labels

yelvert:
Addresses https://github.com/traefik/traefik/issues/8753

Issue #8753, opened 2022-02-02
Allow configuring server URLs with label providers


firstdorsal:
Hi i would love to have this possibility
Configure traefik via label providers: allow setting the `traefik.http.services.<service-name>.loadBalancer.server.url`

for example in docker:
```yaml
version: "3.7"
services:
    rp:
        labels:
            traefik.http.routers.proxy.entryPoints: web
            traefik.http.routers.proxy.service: proxy-crt
            traefik.http.routers.proxy.rule: PathPrefix(`/proxy-crt`)
            traefik.http.routers.proxy.middlewares: strip-crt, cors-crt

            traefik.http.services.proxy.loadBalancer.server.url: ${A_URL_AS_VARIABLE}

            traefik.http.middlewares.strip.stripPrefix.prefixes: /proxy-crt

            traefik.http.middlewares.cors.headers.accessControlAllowMethods: GET,OPTIONS
            traefik.http.middlewares.cors.headers.accessControlAllowOriginlist: "*"
            traefik.http.middlewares.cors.headers.accessControlMaxAge: 86400

```

I have seen this question:
https://community.traefik.io/t/traefik-v2-with-docker-compose-external-services-possible/5792/2

I know that setting this exact configuration is possible with the file provider but as far as i know not with a variable.

```yaml
http:
    routers:
        proxy:
            entryPoints:
                - web
            middlewares:
                - strip
                - cors
            service: proxy
            rule: PathPrefix(`/proxy`)

    services:
        proxy:
            loadBalancer:
                passHostHeader: false
                servers:
                    - url: https://example.com # NOT CONTROLLABLE VIA VARIABLE

    middlewares:
        strip:
            stripPrefix:
                prefixes:
                    - /proxy
        cors:
            headers:
                accessControlAllowMethods: GET,OPTIONS
                accessControlAllowOriginlist: "*"
                accessControlMaxAge: 86400


```

A similar thing is possible with domain certs:
```yaml
traefik.http.routers.myrouter.tls.domains[0].main: example.org
```
could then look like this:
```yaml
 traefik.http.services.proxy.loadBalancer.servers[0].url: ${A_URL_AS_VARIABLE}
```

rtribotte:
Hello @firstdorsal,

Thanks for your interest in Traefik!

There is already a proposal opened for that feature for the ECS provider: https://github.com/traefik/traefik/issues/7568.

[...]

PR #12289, opened 2025-11-18
Add documentation for loadbalancer.server.url in Docker and Swarm providers

webash:
Added information about specifying a URL for the service in Docker configuration. This is per #11374 and #8753

### What does this PR do?

Updates the documentation to specify that `loadbalancer.server.url` is available as a label option for the Docker provider.

### Motivation

It was added in #11374 but the documentation wasn't updated. What is confusing, is that [this page](https://doc.traefik.io/traefik/reference/routing-configuration/http/load-balancing/service/#configuration-example) documents `.loadbalancer.servers[0].url` as an option in the labels example, but that doesn't work (in Swarm, at least).

### More

- [ ] Added/updated tests

### Additional Notes

Replaces #12223

### DIFF
--- a/docs/content/reference/routing-configuration/other-providers/docker.md
+++ b/docs/content/reference/routing-configuration/other-providers/docker.md
@@ -287,10 +287,11 @@ you'd add the label `traefik.http.services.<name-of-your-choice>.loadbalancer.pa
 
 ??? info "`traefik.http.services.<service_name>.loadbalancer.server.url`"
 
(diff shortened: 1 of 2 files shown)


webash:
@kevinpollet my git-fu is no good, so I did what I believed was easier: just cherry-picked the commits into a new branch based on v3.6 in a new PR

kevinpollet (approved):
Thanks 👍

PR #12291, opened 2025-11-19
Validate plugin module name

kevinpollet:
### What does this PR do?

This pull request validates that the plugin name is well-formed cause we were missing some validation. Before, a non-valid plugin module name would have led to an error when retrieving it from the registry.

### More

- [ ] Added/updated tests

### Additional Notes


juliens (approved):
LGTM

PR #12299, opened 2025-11-21
Implement waiting for Nomad health status before service registration

### LINKED ISSUE #12295: Nomad Provider: Services are registered as healthy immediately without waiting for Nomad health status (TODO in code)
### Welcome!

Hello Traefik team,

we are using Traefik with the Nomad Provider and have encountered the following issue:

The Nomad Provider registers new services immediately as healthy, without waiting for Nomad to report them as healthy.
There is a TODO in the code to address this behavior:
[TODO in config.go](https://github.com/traefik/traefik/blob/master/pkg/provider/nomad/config.go#L175
)

We also have Traefik’s own health checks enabled, but the service only gets marked as offline after the first failed health check.
Currently, we have not found a way to register a new service as offline initially until the first successful health check occurs.

Problem
We want to replace/update services/servers in Traefik without any interruption, but with the Nomad Provider this seems impossible under the current behavior. Are we missing something?

Will the TODO in the Nomad Provider be addressed?

Thank you!

kyounghoonJang:
### What does this PR do?
Fixes #12295 

This PR updates the Nomad provider to wait for the Nomad health status before marking services as healthy during registration.
Previously, services were registered as healthy immediately which could cause Traefik to route traffic to services that were not yet ready.

### Motivation

The current behavior does not respect Nomad’s health checks, leading to potential routing to unhealthy or initializing services.
This PR aligns Traefik’s Nomad provider improves reliability in Nomad-based environments.

### More

- [x] Added/updated tests


kyounghoonJang:
@lbenguigui 
Hello, I noticed that TestListenNotBlocking (conn_test.go:146) failed in the CI. Since my changes are isolated to the Nomad provider logic, I am struggling to see a direct connection between this failure and my PR. Could you please advise if there is something I might be missing?

lbenguigui:
Hi @kyounghunJang,
This failure is unrelated to your changes, this test is flaky test in the CI. We’ll rerun the workflow.

Thanks again for the contribution!

VolkerK:
@kyounghunJang Thank you very much.
Can you please continue, we are willing to test a new version with this feature.

kyounghoonJang:
@VolkerK
Hi,
Thank you for your interest! Since I don't have merge permissions, I’m waiting for the maintainers to review and approve the changes. 

@lbenguigui
Could you please take a look at this PR when you have a moment? If there are any further adjustments needed, please let me know. I'd appreciate your review!

mmatur (changes requested):
Thank you for implementing health status filtering for the Nomad provider. This is a valuable addition that addresses a long-standing TODO. However, I have some concerns that should be addressed before merging.

## Architecture Concern

The filtering logic is currently implemented in `fetchService()` within `nomad.go`, but there's an existing TODO at `pkg/provider/nomad/config.go:175-176` that suggests this filtering should be done in the `keepItem()` function:

```go
// TODO: filter on health when that information exists (nomad 1.4+)
```

The `keepItem()` function is the architectural pattern used for filtering items based on configuration and constraints. Moving the health filtering there would:
- Keep filtering logic consolidated in one place
- Follow the existing pattern used for other filtering (enable flag, constraints)
- Make the code more maintainable

Could you consider refactoring to place the health filtering in `keepItem()` or provide justification for the current placement?

## TODO Cleanup

If the filtering remains in `fetchService()`, the TODO comment at `config.go:175-176` should be removed or updated to reflect that health filtering is now implemented elsewhere.

## Nomad Version Compatibility

The TODO mentions "nomad 1.4+" which suggests health information became available in that version. A few concerns:

1. **Documentation**: What happens when this code runs against Nomad versions < 1.4? Will the allocation data be empty/nil, or will the API call fail?

2. **Graceful degradation**: Consider adding documentation or handling for older Nomad versions. Users should understand the minimum Nomad version required for health filtering to work.

3. **Configuration option**: Should there be a configuration option to disable health filtering for users who cannot upgrade their Nomad cluster?

## Performance Consideration

The current implementation makes an additional API call (`p.client.AllocationServiceIDToInfo()`) for each service. This is an N+1 query pattern. For environments with many services, this could significantly impact performance and increase load on the Nomad API.

Consider:
- Documenting this performance impact
- Potentially batching allocation lookups
- Adding a configuration option to disable health checks for performance-sensitive deployments

## Error Handling

In `fetchService()`, when `AllocationServiceIDToInfo()` returns an error, the code silently continues without the allocation info. While this prevents a single failing lookup from breaking the entire provider, it might be helpful to log a warning so users are aware that health filtering may not be working correctly.

## Minor

The fixture file `pkg/provider/nomad/fixtures/sample.json` appears to have a trailing newline added - please verify this is intentional.

kyounghoonJang:
@mmatur 
Thank you for the feedback. I’m planning to address the points below:

1.  Architecture consideration
In my view, performing the health check in `fetchService()` helps avoid creating unnecessary item objects. That said, creating a few additional items is unlikely to cause a significant performance impact. As an alternative, we can move the filtering logic into `KeepItem` if that approach is preferred.

2. Nomad version compatibility
This should not result in an error; instead, the value will likely be handled as empty or nil. In such cases, I think logging a warning would be sufficient. I’m also planning to add a `DisableHealthCheck` flag to allow enabling or disabling this behavior.

3. Performance considerations
I realized I had overlooked an optimization opportunity. There appears to be a way to fetch `allocateIds` in a single call, which would allow us to eliminate the N+1 query pattern.

shubert-wktaa:
Hey, @mmatur , @nmengin, any chance we may get a merge approval on @kyounghunJang's work? This is rather important for us since it currently causes our zero downtime deployments to cause session interruptions whenever node are routed traffic to which are started but not yet healthy.

Thank you very much in advance!

shubert-wktaa:
Salut @mmatur,

Has this been progressed into the next release already or why was it removed from "next" milestone? Can we help bringing this into the next release (we could test it for instance) somehow?

mmatur:
Hi @shubert-wktaa,

Unfortunately, we weren't able to complete the review in time for the v3.7 release, which is why it was removed from the milestone.

We understand this feature is important for your zero-downtime deployments, and we'll prioritize reviewing this PR in the coming weeks.

Thank you for your patience and for following up!

gavaga:
Hi @mmatur !

We're running into the same issue with deployments in our nomad cluster and noticed this PR hasn't had any activity since March--just wanted to check in on whether there's a plan to move forward with this or if there's anything we can do to help it along? Thank you!

PR #12303, opened 2025-11-21
Allow to decode encoded slashes in the request path

rtribotte:
### What does this PR do?
This PR adds the a `decodePathSlashes` option to the http configuration of entrypoints to allow to decode encoded slashes in the request path.

### Motivation

To allow more normalization of the request path.

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -681,7 +681,9 @@ func newHTTPServer(ctx context.Context, ln net.Listener, configuration *static.E
 		handler = sanitizePath(handler)
 	}
 
-	// Decode %2F, %2f, %5C, and %5c.
+	// Note that the decoding of slashes has to be done before the path sanitization,
+	// hence the wrapping has to be done after the sanitize path wrapping.
+	// Decodes %2F, %2f, %5C, and %5c.
 	if configuration.HTTP.DecodePathSlashes {
 		handler = decodePathSlashes(handler)
 	}


docs/content/reference/install-configuration/entrypoints.md
```diff
@@ -181,11 +182,25 @@ Behavior examples:
 | <a id="opt-false-2" href="#opt-false-2" title="#opt-false-2">false</a> | foo=bar&baz=bar;foo | foo=bar&baz=bar&foo     |
 | <a id="opt-true-2" href="#opt-true-2" title="#opt-true-2">true</a> | foo=bar&baz=bar;foo | foo=bar&baz=bar%3Bfoo   |
 
+### DecodePathSlashes
```
lbenguigui, line 185: I think we should add a section to warn users that if they don’t use this option with SanitizePath, they are exposed to path traversal risks. WDYT?

pkg/server/server_entrypoint_tcp.go
```diff
@@ -767,6 +774,71 @@ func (t *trackedConnection) Close() error {
...
 
+// slashesCharacters contains the mapping of the percent-encoded form to the ASCII form
+// of the slashes characters.
+var slashesCharacters = map[string]rune{
+	"%2F": '/',
+	"%5C": '\\',
+}
+
+// decodePathSlashes decodes encoded slashes characters from the request URL RawPath.
+func decodePathSlashes(h http.Handler) http.Handler {
```
lbenguigui, line 785: nit: since it's the same code as the normalization function, WDYT about creating a ` func decodeCharacters(h http.Handler, charMap map[string]rune) http.Handler` ?

PR #12305, opened 2025-11-22
Add http.route and server.port to OTLP metrics

### LINKED ISSUE #12247: Metrics should contain the attribute http.route
### Welcome!

Traefik emits metrics either via Prometheus or (in my case) OTLP. I receive metrics like:

* http.server.request.duration.seconds.bucket
* http.server.request.duration.seconds.count
* http.server.request.duration.seconds.sum

They have labels like

* http.request_method
* http.response_status_code

What I am missing are the labels

* http.route <- same as url.path in the spans
* server.name <- same as server.address in the spans
* server.port

According to the OpenTelemetry Semantic Conventions for HTTP Metrics (https://opentelemetry.io/docs/specs/semconv/http/http-metrics/), there should be (if available) the label http.route.

That would make getting which routes of which hostnames and routes are used much more useful.

amcheste:
### What does this PR do?

  Adds missing OpenTelemetry semantic convention attributes to HTTP server metrics emitted via OTLP:
  - `http.route` - The matched route pattern from the router
  - `server.port` - The server port extracted from the Host header

  ### Motivation

  Fixes #12247

  According to the [OpenTelemetry Semantic Conventions for HTTP Metrics](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/), the `http.route` attribute should be included when available.

  ### More

  - [x] Added/updated tests


nmengin:
Hey @amcheste,

Thanks for this!

pkg/middlewares/observability/semconv.go
```diff
@@ -76,6 +77,18 @@ func (e *semConvServerMetrics) ServeHTTP(rw http.ResponseWriter, req *http.Reque
 	attrs = append(attrs, semconv.NetworkProtocolVersion(Proto(req.Proto)))
 	attrs = append(attrs, semconv.ServerAddress(req.Host))
 
+	// Add http.route attribute if available from router context.
+	if route, ok := HTTPRoute(ctx); ok && route != "" {
+		attrs = append(attrs, semconv.HTTPRoute(route))
+	}
+
```
noa-lucent, line 84: [major] This only populates `server.port` when the Host header includes an explicit port. For the common case (`https://example.com/...`), `req.Host` is just `example.com`, so this block never runs and the OTLP metric still lacks `server.port`. Issue #12247 explicitly calls out that https traffic without an explicit port should emit `server.port = 443`, so the current change doesn’t satisfy the issue. Can we derive the listener port from `http.LocalAddrContextKey` (or another canonical source such as the entrypoint metadata) and always emit it when available, falling back to the Host header parsing only when necessary? Please cover the default-port case with a test as well.

noa-lucent (commented):
Thanks for the updates—`http.route` propagating through the router middleware looks good and the docs/tests were adjusted accordingly. However, the new `server.port` attribute is only emitted when the incoming Host header carries an explicit port. For the very common case of HTTPS on 443 (or HTTP on 80) the Host header omits the port, so the OTLP metrics will still be missing `server.port`, leaving issue #12247 unresolved. Please derive the port from a canonical source such as `http.LocalAddrContextKey` (or equivalent entrypoint metadata) so that default ports are reported, and extend the test suite to cover that scenario. I added an inline comment with details.

amcheste:
@noa-lucent thank you for the feedback!  I will work on getting htis addressed and incorporated.

PR #12318, opened 2025-11-25
NGINX Ingress Controller to Traefik Migration Guide

sheddy-traefik:
### What does this PR do?

This PR adds a new migration Guide for NGINX Ingress Controller to Traefik

### More

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
@@ -453,7 +453,7 @@ func (p *Provider) loadConfiguration(ctx context.Context) *dynamic.Configuration
 				}
 
 				// TODO: if no service, do not add middlewares and 503.
-				serviceName := provider.Normalize(ingress.Namespace + "-" + pa.Backend.Service.Name + "-" + portString)
+				serviceName := provider.Normalize(ingress.Namespace + "-" + ingress.Name + "-" + pa.Backend.Service.Name + "-" + portString)
 
 				service, err := p.buildService(ingress.Namespace, pa.Backend, ingressConfig)
 				if err != nil {
@@ -509,7 +509,7 @@ func (p *Provider) buildServersTransport(namespace, name string, cfg ingressConf
 		Name: provider.Normalize(namespace + "-" + name),
 		ServersTransport: &dynamic.ServersTransport{
 			ServerName:         ptr.Deref(cfg.ProxySSLName, ptr.Deref(cfg.ProxySSLServerName, "")),
-			InsecureSkipVerify: strings.ToLower(ptr.Deref(cfg.ProxySSLVerify, "off")) == "on",
+			InsecureSkipVerify: strings.ToLower(ptr.Deref(cfg.ProxySSLVerify, "off")) == "off",
 		},
 	}
 
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
@@ -64,7 +64,7 @@ func TestLoadIngresses(t *testing.T) {
 							Rule:        "Host(`whoami.localhost`) && Path(`/basicauth`)",
 							RuleSyntax:  "default",
 							Middlewares: []string{"default-ingress-with-basicauth-rule-0-path-0-basic-auth"},
-							Service:     "default-whoami-80",
+							Service:     "default-ingress-with-basicauth-whoami-80",
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{
@@ -78,7 +78,7 @@ func TestLoadIngresses(t *testing.T) {
 						},
 					},
 					Services: map[string]*dynamic.Service{
-						"default-whoami-80": {
+						"default-ingress-with-basicauth-whoami-80": {
 							LoadBalancer: &dynamic.ServersLoadBalancer{
 								Servers: []dynamic.Server{
 									{
@@ -119,7 +119,7 @@ func TestLoadIngresses(t *testing.T) {
 							Rule:        "Host(`whoami.localhost`) && Path(`/forwardauth`)",
 							RuleSyntax:  "default",
 							Middlewares: []string{"default-ingress-with-forwardauth-rule-0-path-0-forward-auth"},
-							Service:     "default-whoami-80",
+							Service:     "default-ingress-with-forwardauth-whoami-80",
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{
@@ -131,7 +131,7 @@ func TestLoadIngresses(t *testing.T) {
 						},
 					},
 					Services: map[string]*dynamic.Service{
-						"default-whoami-80": {
+						"default-ingress-with-forwardauth-whoami-80": {
 							LoadBalancer: &dynamic.ServersLoadBalancer{
 								Servers: []dynamic.Server{
 									{
@@ -173,7 +173,7 @@ func TestLoadIngresses(t *testing.T) {
 							Rule:       "Host(`sslredirect.localhost`) && Path(`/`)",
 							RuleSyntax: "default",
 							TLS:        &dynamic.RouterTLSConfig{},
-							Service:    "default-whoami-80",
+							Service:    "default-ingress-with-ssl-redirect-whoami-80",
 						},
 						"default-ingress-with-ssl-redirect-rule-0-path-0-redirect": {
 							Rule:        "Host(`sslredirect.localhost`) && Path(`/`)",
@@ -184,18 +184,18 @@ func TestLoadIngresses(t *testing.T) {
 						"default-ingress-without-ssl-redirect-rule-0-path-0-http": {
 							Rule:       "Host(`withoutsslredirect.localhost`) && Path(`/`)",
 							RuleSyntax: "default",
-							Service:    "default-whoami-80",
+							Service:    "default-ingress-without-ssl-redirect-whoami-80",
 						},
 						"default-ingress-without-ssl-redirect-rule-0-path-0": {
 							Rule:       "Host(`withoutsslredirect.localhost`) && Path(`/`)",
 							RuleSyntax: "default",
 							TLS:        &dynamic.RouterTLSConfig{},
-							Service:    "default-whoami-80",
+							Service:    "default-ingress-without-ssl-redirect-whoami-80",
 						},
 						"default-ingress-with-force-ssl-redirect-rule-0-path-0": {
 							Rule:       "Host(`forcesslredirect.localhost`) && Path(`/`)",
 							RuleSyntax: "default",
-							Service:    "default-whoami-80",
+							Service:    "default-ingress-with-force-ssl-redirect-whoami-80",
 						},
 						"default-ingress-with-force-ssl-redirect-rule-0-path-0-redirect": {
 							Rule:        "Host(`forcesslredirect.localhost`) && Path(`/`)",
@@ -219,7 +219,41 @@ func TestLoadIngresses(t *testing.T) {
 						},
 					},
 					Services: map[string]*dynamic.Service{
-						"default-whoami-80": {
+						"default-ingress-with-ssl-redirect-whoami-80": {
+							LoadBalancer: &dynamic.ServersLoadBalancer{
+								Servers: []dynamic.Server{
+									{
+										URL: "http://10.10.0.1:80",
+									},
+									{
+										URL: "http://10.10.0.2:80",
+									},
+								},
+								Strategy:       "wrr",
+								PassHostHeader: ptr.To(true),
+								ResponseForwarding: &dynamic.ResponseForwarding{
+									FlushInterval: dynamic.DefaultFlushInterval,
+								},
+							},
+						},
+						"default-ingress-without-ssl-redirect-whoami-80": {
+							LoadBalancer: &dynamic.ServersLoadBalancer{
+								Servers: []dynamic.Server{
+									{
+										URL: "http://10.10.0.1:80",
+									},
+									{
+										URL: "http://10.10.0.2:80",
+									},
+								},
+								Strategy:       "wrr",
+								PassHostHeader: ptr.To(true),
+								ResponseForwarding: &dynamic.ResponseForwarding{
+									FlushInterval: dynamic.DefaultFlushInterval,
+								},
+							},
+						},
+						"default-ingress-with-force-ssl-redirect-whoami-80": {
 							LoadBalancer: &dynamic.ServersLoadBalancer{
 								Servers: []dynamic.Server{
 									{
@@ -313,12 +347,12 @@ func TestLoadIngresses(t *testing.T) {
 						"default-ingress-with-sticky-rule-0-path-0": {
 							Rule:       "Host(`sticky.localhost`) && Path(`/`)",
 							RuleSyntax: "default",
-							Service:    "default-whoami-80",
+							Service:    "default-ingress-with-sticky-whoami-80",
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{},
 					Services: map[string]*dynamic.Service{
-						"default-whoami-80": {
+						"default-ingress-with-sticky-whoami-80": {
 							LoadBalancer: &dynamic.ServersLoadBalancer{
 								Servers: []dynamic.Server{
 									{
@@ -370,12 +404,12 @@ func TestLoadIngresses(t *testing.T) {
 						"default-ingress-with-proxy-ssl-rule-0-path-0": {
 							Rule:       "Host(`proxy-ssl.localhost`) && Path(`/`)",
 							RuleSyntax: "default",
-							Service:    "default-whoami-tls-443",
+							Service:    "default-ingress-with-proxy-ssl-whoami-tls-443",
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{},
 					Services: map[string]*dynamic.Service{
-						"default-whoami-tls-443": {
+						"default-ingress-with-proxy-ssl-whoami-tls-443": {
 							LoadBalancer: &dynamic.ServersLoadBalancer{
 								Servers: []dynamic.Server{
 									{
@@ -397,7 +431,7 @@ func TestLoadIngresses(t *testing.T) {
 					ServersTransports: map[string]*dynamic.ServersTransport{
 						"default-ingress-with-proxy-ssl": {
 							ServerName:         "whoami.localhost",
-							InsecureSkipVerify: true,
+							InsecureSkipVerify: false,
 							RootCAs:            []types.FileOrContent{"-----BEGIN CERTIFICATE-----"},
 						},
 					},
@@ -423,7 +457,7 @@ func TestLoadIngresses(t *testing.T) {
 							Rule:        "Host(`cors.localhost`) && Path(`/`)",
 							RuleSyntax:  "default",
 							Middlewares: []string{"default-ingress-with-cors-rule-0-path-0-cors"},
-							Service:     "default-whoami-80",
+							Service:     "default-ingress-with-cors-whoami-80",
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{
@@ -439,7 +473,7 @@ func TestLoadIngresses(t *testing.T) {
 						},
 					},
 					Services: map[string]*dynamic.Service{
-						"default-whoami-80": {
+						"default-ingress-with-cors-whoami-80": {
 							LoadBalancer: &dynamic.ServersLoadBalancer{
 								Servers: []dynamic.Server{
 									{
@@ -479,12 +513,12 @@ func TestLoadIngresses(t *testing.T) {
 						"default-ingress-with-service-upstream-rule-0-path-0": {
 							Rule:       "Host(`service-upstream.localhost`) && Path(`/`)",
 							RuleSyntax: "default",
-							Service:    "default-whoami-80",
+							Service:    "default-ingress-with-service-upstream-whoami-80",
 						},
 					},
 					Middlewares: map[string]*dynamic.Middleware{},
 					Services: map[string]*dynamic.Service{
-						"default-whoami-80": {
+						"default-ingress-with-service-upstream-whoami-80": {
 							LoadBalancer: &dynamic.ServersLoadBalancer{
 								Servers: []dynamic.Server{
 									{


docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+3. Uninstall NGINX Ingress Controller
+4. Install Traefik with the released IP and IngressClass configuration
+5. Verify traffic routing through Traefik (downtime ends)
+6. Monitor and validate
+
+**Expected Downtime:** Brief downtime (typically seconds to a few minutes) during the switchover, depending on cloud provider IP release timing. If using DNS-based migration instead of IP retention, downtime depends on DNS TTL.
+
+---
+
+## Step 1: Install Traefik with Helm
```
immanuelfodor: The step title doesn't match the migration flow step above.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+
+!!! warning "IngressClass Preservation"
+
+    The `extraObjects` section creates the `nginx` IngressClass that Traefik needs to recognize your existing Ingresses. When you uninstall the NGINX Ingress Controller via Helm, it deletes the `nginx` IngressClass. Without this IngressClass, Traefik cannot route traffic to Ingresses that specify `ingressClassName: nginx`.
+
+    **Important:** If NGINX is still installed, the IngressClass already exists and Helm will report a conflict. You must uninstall NGINX **before** installing Traefik with the `extraObjects` configuration, or install Traefik without `extraObjects` first and manually create the IngressClass after uninstalling NGINX.
+
+---
+
+## Step 2: Load Balancer IP Retention
```
immanuelfodor: Same

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+
+For other cloud providers, check their Kubernetes LoadBalancer documentation:
+
+- **DigitalOcean:** Supports `loadBalancerIP` with floating IPs
+- **Linode:** Supports `loadBalancerIP` specification
+- **Bare Metal:** Use MetalLB with IP address pools
+
+---
+
+## Step 3: Install Traefik
```
immanuelfodor: Now we even skipped one step (uninstall of Ningx was part of step 3)

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+Expected output:
+
+```bash
+NAME      TYPE           CLUSTER-IP      EXTERNAL-IP      PORT(S)
+traefik   LoadBalancer   10.0.123.45     <your-lb-ip>     80:<nodeport>/TCP,443:<nodeport>/TCP
+```
+
+---
+
+## Step 4: Test with Sample Application
```
immanuelfodor: Please check all steps, or you could switch to an UL instead of an OL in the migration flow as an alternative solution. There, you just explain a logical flow, and here, you give the concrete steps. But with the ordered list, I'd expect the same steps reflected below.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+
+Before migrating production workloads, test Traefik with a sample application using the `traefik/whoami` image.
```
immanuelfodor: I'm not sure about this step. Since the LB already points to Traefik, why test before migrating production workloads? Those have already been redirected by the switch. I could imagine such a test: Traefik is installed _besides_ Nginx, then upgraded to replace it, but here we're after the main action.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+          - path: /api
+            pathType: Prefix
+            backend:
+              service:
+                name: myapp-service
+                port:
+                  number: 8080
+```
+
+```yaml tab="After (Traefik)"
```
immanuelfodor: We should add "no change" or similar to highlight the surprise :)

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+
+Test that Traefik correctly routes traffic to the whoami service:
+
+```bash
+# Test using the LoadBalancer IP
+curl -H "Host: whoami.example.com" http://<traefik-lb-ip>/
+
+# Or add an entry to /etc/hosts for testing
+echo "<traefik-lb-ip> whoami.example.com" | sudo tee -a /etc/hosts
+curl http://whoami.example.com/
```
immanuelfodor: There is a CLI flag for curl where you can pass a custom DNS entry to resolve a specified IP:

```
Resolve a hostname to a custom IP address, with verbose output (similar to editing the /etc/hosts file for custom DNS resolution):
    curl --resolve example.com:80:127.0.0.1 http://example.com
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+# List secrets in the namespace
+kubectl get secrets -n <namespace>
+
+# Check TLS secret format
+kubectl get secret <tls-secret-name> -n <namespace> -o yaml
+```
+
+---
+
+## Step 7: Decommission NGINX Ingress Controller
```
immanuelfodor: Now I'm totally confused about what's happening :D The flow in the beginning explained it completely differently. Have we just removed the LB service so far, but not Nginx itself? Some points above said we should remove Nginx to solve some issues. If Nginx is still running, then I can probably get why the migration test was in between, but the LB was already pointing to Traefik with the new ingress class, so that point is still not clear.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+  
+If NGINX Ingress Controller was installed via Helm, run the following command to uninstall the chart and clean up chart-managed resources:
+
+```bash
+helm uninstall ingress-nginx -n ingress-nginx
+```
+
+!!! warning "IngressClass Deleted by Helm"
+
+    When you run `helm uninstall`, the `nginx` IngressClass is also deleted. If you followed the Helm values configuration in [Step 1](#step-1-install-traefik-with-helm) (which includes `extraObjects` to create the IngressClass), Traefik will continue to work. If you did not include the `extraObjects` section, you must manually create the IngressClass:
```
immanuelfodor: Since the extra objects create the same ingress class, wouldn't the Nginx uninstall delete that one, too? If there was a conflict and the nginx-originated one was deleted to add the Traefik one, I think this one will also be removed upon uninstalling nginx. So in the end, the ingress class has to be added back, even if you had it in extra objects.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+kubectl delete validatingwebhookconfiguration ingress-nginx-admission
+kubectl delete mutatingwebhookconfiguration ingress-nginx-admission --ignore-not-found
+
+# Delete the namespace (if dedicated to NGINX)
+kubectl delete namespace ingress-nginx
+```
+
+!!! note "LoadBalancer IP release timing"
+
+    Cloud providers may hold on to the NGINX LoadBalancer IP for a short time after deleting the Service. Deleting the NGINX service early (or letting Helm uninstall it) speeds up reusing that IP for Traefik, but actual release timing depends on the provider.
```
immanuelfodor: I'm not sure why the note is here, since the LB was already handled in a previous step

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,726 @@
...
+### Remove NGINX Resources
+
+!!! info
+    
+    In this section, we assume that the NGINX Ingress Controller was installed in the namespace `ingress-nginx`. If it was installed in a different namespace, you need to replace `ingress-nginx` with the namespace name.
+  
+If NGINX Ingress Controller was installed via Helm, run the following command to uninstall the chart and clean up chart-managed resources:
+
+```bash
+helm uninstall ingress-nginx -n ingress-nginx
```
immanuelfodor: Wouldn't helm uninstall everything, even those resources that Traefik is also using? Specifically: the LB service and the ingress class

immanuelfodor (changes requested):
The guide is well-written and detailed enough, but maybe this is why I also got confused about the several steps, which is my main issue. Other than that, small suggestions.

xxsheddy:
@immanuelfodor I think I get your point about the flow. I'll fix it.

xxsheddy:
@immanuelfodor i've completely restructured the guide, I think this is better now based on your feedback.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,547 @@
...
+
+!!! warning "Cloud Provider Variations"
+
+    Load balancer IP retention strategies vary by cloud provider. Follow the appropriate section for your environment.
+
+!!! warning "ServiceLB and hostPort-based environments"
+
+    On k3s/k3d (and any environment that implements `Service.Type=LoadBalancer` by binding host ports such as Klipper/ServiceLB), only one LoadBalancer service can listen on ports 80/443 at a time. Delete the NGINX LoadBalancer service before creating Traefik's service, or use different ports, otherwise Traefik's service will stay in `Pending`.
+
+#### AWS Elastic Load Balancer
```
immanuelfodor: Does Mkdocs support a collapsible section like an accordion? These vendor-specific instruction options could go into their own collapsible to reduce the length of the guide. Also, cummunity members in homelabs would not need these at all.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,547 @@
...
+
+Ensure TLS secrets are in the same namespace as the Ingress:
+
+```bash
+# List secrets in the namespace
+kubectl get secrets -n <namespace>
+
+# Check TLS secret format
+kubectl get secret <tls-secret-name> -n <namespace> -o yaml
+```
```
immanuelfodor: I'd make these collapsed, too

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,547 @@
...
+For a complete list of supported annotations, limitations, and behavioral differences, see the [Ingress NGINX Routing Configuration](../reference/routing-configuration/kubernetes/ingress-nginx.md) documentation.
+
+#### Existing TLS Secrets
+
+Existing TLS configurations continue to work with Traefik:
+
+- Keep `spec.tls` entries exactly as-is; Traefik terminates TLS using the referenced secrets
+- TLS secrets must stay in the same namespace as the Ingress
+- NGINX `ssl-redirect` / `force-ssl-redirect` annotations are honored
+- Verify with `kubectl describe ingress <name>` and `curl -k https://<host>/`
```
immanuelfodor: I'd move this section from here to the top: this is the main goal, this is what we'll want to achieve with the migration, by completing this guide, this is what you'll have. You may need to adjust the text to reflect this goal approach.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,547 @@
...
+The key benefit of the Traefik Kubernetes Ingress NGINX provider is that your existing Ingress resources work without modification:
+
+```yaml tab="Your Existing Ingress (No Changes Needed)"
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: myapp
+  namespace: production
+  annotations:
+    # These NGINX annotations are automatically translated by Traefik
```
immanuelfodor:
```suggestion
    # These NGINX annotations will be automatically translated by Traefik
```

In the sense of the goal rewrite.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,547 @@
...
+    nginx.ingress.kubernetes.io/ssl-redirect: "true"
+    nginx.ingress.kubernetes.io/enable-cors: "true"
+    nginx.ingress.kubernetes.io/cors-allow-origin: "https://example.com"
+    nginx.ingress.kubernetes.io/affinity: "cookie"
+    nginx.ingress.kubernetes.io/session-cookie-name: "route"
+spec:
+  ingressClassName: nginx  # ← Traefik watches this class
```
immanuelfodor:
```suggestion
  ingressClassName: nginx  # ← Traefik will watch this class
```

In the sense of the goal rewrite.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,547 @@
...
+- [Kubernetes Ingress NGINX Install Configuration](../reference/install-configuration/providers/kubernetes/kubernetes-ingress-nginx.md) - Detailed provider configuration
+- [Kubernetes Ingress NGINX Routing Configuration](../reference/routing-configuration/kubernetes/ingress-nginx.md) - Routing rules and annotation support
+- [HTTP Middlewares](../reference/routing-configuration/http/middlewares/overview.md) - Extend functionality beyond NGINX annotations
+- [TLS Configuration](../reference/routing-configuration/http/tls/overview.md) - Advanced TLS and certificate management
+
+**Enhance Your Setup:**
+
+- Enable [metrics](../reference/install-configuration/observability/metrics.md) and [tracing](../reference/install-configuration/observability/tracing.md)
+- Configure [access logs](../reference/install-configuration/observability/logs-and-accesslogs.md) for observability
+- Explore [Traefik Middlewares](../reference/routing-configuration/http/middlewares/overview.md) for advanced traffic management
```
immanuelfodor, line 664: Add below: Migrate from Nginx-based config to Traefik IngressRoutes or K8s Gateway API

immanuelfodor (changes requested):
Excellent rewrite! Thank you! It's a lot clearer now! Left some suggestions to make it even better 🤗

xxsheddy:
@immanuelfodor i've had to rewrite the migration guide based on new information I got from the platform team, can you take a look again?

immanuelfodor:
Just to be clear, for the 4th rewrite, I'll need some form of compensation @sheddy-traefik :P

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,577 @@
...
+By completing this migration, your existing Ingress resources will work with Traefik without any modifications. The Traefik Kubernetes Ingress NGINX Provider automatically translates NGINX annotations into Traefik configuration:
+
+```yaml tab="Your Existing Ingress (No Changes Needed)"
+
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: myapp
+  annotations:
+    # These NGINX annotations are automatically translated by Traefik
```
immanuelfodor, line 33:
```suggestion
    # NGINX annotations are automatically translated by Traefik
```

immanuelfodor: Removing "these" as it could unintentionally hint a wrong message that these are the only supported annotations.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,577 @@
...
+
+### Install Traefik
+
+```bash
+helm upgrade --install traefik traefik/traefik \
+  --namespace traefik --create-namespace \
+  --set providers.kubernetesIngressNginx.enabled=true
+```
+
+Or using a values file for more configuration:
```
immanuelfodor: A link to the options reference would be nice

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,577 @@
...
+
+# Check Traefik pods
+kubectl get pods -n traefik
+
+# Check both services have LoadBalancer IPs
+kubectl get svc -n ingress-nginx ingress-nginx-controller
+kubectl get svc -n traefik traefik
+```
+
+At this point, both NGINX and Traefik are running and can serve the same Ingress resources. Traffic is still flowing only through NGINX since DNS points to the NGINX LoadBalancer.
```
immanuelfodor, line 202: In a previous rewrite, we mentioned that some k8s envs don't allow 2 LBs with the same port, if I'm not mistaken. Shouldn't we call out this fact in the beginning of the guide, eg. in prerequisites? They should get an escape hatch, for example, the previous version with the minimal downtime and retaining the LB IP in a shorter form in a collapsible.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,577 @@
...
+### Test via Port-Forward
+
+Use port-forwarding to test Traefik directly without affecting production traffic:
+
+```bash
+# Port-forward to Traefik
+kubectl port-forward -n traefik deployment/traefik 8080:8000 &
+
+# Test an existing Ingress through Traefik
+curl --resolve myapp.example.com:8080:127.0.0.1 http://myapp.example.com:8080/
```
immanuelfodor: The port-forward is a safe way for the guide to provide a unified testing method, but this might not work with some apps that are not handling the additional port appended to the domain. I also wonder if this would work with HTTPS either.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,577 @@
...
+When NGINX is uninstalled via Helm, it deletes the `nginx` IngressClass by default. Traefik needs this IngressClass to continue recognizing your Ingresses. You must ensure the IngressClass persists after NGINX removal.
+
+### If NGINX Was Installed via Helm
+
+Upgrade the NGINX Helm release to add the `helm.sh/resource-policy: keep` annotation to the IngressClass:
+
+```bash
+helm upgrade --install ingress-nginx ingress-nginx/ingress-nginx \
+  --namespace ingress-nginx \
+  --set-json 'controller.ingressClassResource.annotations={"helm.sh/resource-policy": "keep"}'
```
immanuelfodor: Since we cannot know what were the original parameters, shouldn't we just kubectl patch or edit it? It's not at all safe to run this command verbatim.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,577 @@
...
+
+1. **Add Traefik to DNS** - Add the Traefik LoadBalancer IP to your DNS records (both IPs now receive traffic via round-robin)
+2. **Monitor** - Observe traffic patterns on both controllers
+3. **Remove NGINX from DNS** - Once confident, remove the NGINX LoadBalancer IP from DNS
+4. **Wait for DNS propagation** - Allow time for DNS caches to expire
+5. **Uninstall NGINX** - Proceed to [Step 5](#step-5-uninstall-nginx-ingress-controller)
+
+### Option B: External Load Balancer with Weighted Traffic (Best)
+
+For more control over traffic distribution, use an external load balancer (like Cloudflare, AWS ALB, or a dedicated load balancer) in front of both Kubernetes LoadBalancers.
```
immanuelfodor: It could be even Traefik! 😀

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,577 @@
...
+| Increase | 50% | 50% | 1 hour |
+| Near-complete | 10% | 90% | 1 hour |
+| Final | 0% | 100% | - |
+
+!!! tip "External Load Balancer Options"
+
+    - **Cloudflare Load Balancing** - Traffic steering with health checks
+    - **AWS Global Accelerator** - Weighted routing across endpoints
+    - **Google Cloud Load Balancing** - Traffic splitting
+    - **HAProxy / NGINX (external)** - Self-hosted option with weighted backends
```
immanuelfodor: Please list Traefik, too, not only others 🥺

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,577 @@
...
+kubectl delete mutatingwebhookconfiguration ingress-nginx-admission --ignore-not-found
+```
+
+### Uninstall NGINX
+
+```bash
+helm uninstall ingress-nginx -n ingress-nginx
+```
+
+You should see output confirming the IngressClass was preserved:
```
immanuelfodor: +"In case you somehow still managed to delete it, recreate it from Step X".

docs/content/migrate/nginx-to-traefik.md
xxsheddy: i tested this with Remi, that's the exact parameters we used.

docs/content/migrate/nginx-to-traefik.md
xxsheddy: its still mentioned, take a look at line: 451

docs/content/migrate/nginx-to-traefik.md
xxsheddy: proxy doesn't have an options list page for the helm chart yet

docs/content/migrate/nginx-to-traefik.md
xxsheddy: what do you suggest we do about this possibility? should we leave it to the readers to mitigate this?

kamilhristov:
Hey, just wanted to share something I ran into during the migration.
I’ve got ExternalDNS in the mix, and it’s watching the Ingress status to sync DNS records.

```yaml
kind: Ingress
metadata:
  name: my-app
  namespace: default
spec:
  ingressClassName: nginx
  rules: []
status:
  loadBalancer:
    ingress:
    - ip: 10.10.0.1
 ```

Now, if you install Traefik with **publishService** enabled (like the docs say), it updates the Ingress status with its own LB IP. Then ExternalDNS points traffic straight to Traefik. Nginx will try to flip it back, but Traefik usually wins because it’s faster.

What worked for me:
1. Install Traefik in “shadow” mode (disable **publishService**).
2. Test everything on the new Traefik LB.
3. Let Nginx handle the switch-over by publishing the new Traefik service [here](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml#L182)
```
controller:
  publishService:
    pathOverride: "traefik/traefik"
```
4. This way, you can roll back anytime if needed.
5. When ready, enable **publishService** on Traefik.
6. Uninstall Nginx

The above is something like the DNS-Based migration but with direct switch to one or another because of ExternalDNS.

**Comment by mloiseleur** [comment]
Thanks @kamilhristov :+1: ! 
It's widely used, so it makes sense to explain in this guide how to include it in the equation.

File: docs/content/migrate/nginx-to-traefik.md (hunk 22 above)
Comment (immanuelfodor) [reply docs/content/migrate/nginx-to-traefik.md]: I get that these are what we used somewhere, but is it the same for everybody? ingress-nginx has a values file of 1276 lines. Here we are adding just one specific option with the upgrade, but then wouldn't it reset everything else with the defaults?

Comment (xxsheddy) [reply docs/content/migrate/nginx-to-traefik.md]: Ah I see what you mean now, you raise a good point, patching/editing is a better option

File: docs/content/migrate/nginx-to-traefik.md (hunk 19 above)
Comment (immanuelfodor) [reply docs/content/migrate/nginx-to-traefik.md]: Oh, I did not know. Could we link to the helm values then? It usually has nice explanatory comments.

File: docs/content/migrate/nginx-to-traefik.md (hunk 20 above)
Comment (immanuelfodor) [reply docs/content/migrate/nginx-to-traefik.md:202]: Ahh, I missed it. But we're still in Step 1, and that's in Step 4. This sentence will not be completely true for those folks, since the LB service will not get an IP.

File: docs/content/migrate/nginx-to-traefik.md (hunk 21 above)
Comment (immanuelfodor) [reply docs/content/migrate/nginx-to-traefik.md]: If Traefik gets an LB IP (not in K3x, see the other comment), then they could just resolve the same domain to that new LB IP but it should work just the same as it would with ingress-nginx on the old/current IP. I don't fully get the point why workaround with a port-forward.

File: docs/content/migrate/nginx-to-traefik.md (hunk 22 above)
Comment (xxsheddy) [reply docs/content/migrate/nginx-to-traefik.md]: I just tested the following:

kubectl annotate -  Does not work
kubectl patch -  Does not work
kubectl edit - Does not work (same mechanism)

The reason is that Helm maintains its own internal state (stored in Secrets) and checks annotations from that internal manifest during uninstall, not from the live cluster resources. Only helm upgrade updates Helm's internal state.

Comment (xxsheddy) [reply docs/content/migrate/nginx-to-traefik.md]: i've added a `--reuse-values` flag to preserve the existing user customization

Comment (immanuelfodor) [reply docs/content/migrate/nginx-to-traefik.md]: Works for me, thank you!

File: docs/content/migrate/nginx-to-traefik.md (hunk 20 above)
Comment (xxsheddy) [reply docs/content/migrate/nginx-to-traefik.md:202]: I spoke to the team about this and we've decided not to focus on such environments in this guide since its not built to be HA and can afford downtime. 

i'll also take off the note and include a prerequisite for clusters to allow running multiple LB simultaneously

File: docs/content/migrate/nginx-to-traefik.md [hunk 26]
```diff
@@ -0,0 +1,615 @@
...
+- **Kubernetes cluster access** with `kubectl` configured 
+- **Cluster support for running multiple LoadBalancer services** on ports 80/443 simultaneously
+- **Helm**
+- **Cluster admin permissions** to create RBAC resources
+- **Backup of critical configurations** (Ingress resources, ConfigMaps, Secrets)
+
+!!! tip "Backup Recommendations"
+
+    ```bash
+    # Export all Ingress resources
```
darkweaver87, line 111: Good practice, but should we also provide a command to restore from those backups ?

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+    expose:
+      default: true
+    exposedPort: 443
+    tls:
+      enabled: true
+
+service:
+  enabled: true
+  type: LoadBalancer
+```
```
darkweaver87, line 180: I think it's confusing, what's needed, what's not. We should stick to the bare minimum for this migration:
````suggestion
Or using a [values file](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/VALUES.md) for more configuration:

```yaml tab="traefik-values.yaml"
...
providers:
  kubernetesIngressNginx:
    enabled: true
 ...
```
````

File: docs/content/migrate/nginx-to-traefik.md [hunk 28]
```diff
@@ -0,0 +1,615 @@
...
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: myapp
+                port:
+                  number: 80
+```
```
darkweaver87, line 87: Users should be able to deploy and test this procedure ... we should give a real example like here: https://github.com/traefik/traefik-helm-chart/blob/master/EXAMPLES.md#use-kubernetes-ingress-nginx-provider.

```suggestion
````yaml tab="Your Existing Ingress (No Changes Needed)"
kind: Ingress
metadata:
  name: myapp
  annotations:
    # These NGINX annotations are automatically translated by Traefik
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/enable-cors: "true"
    nginx.ingress.kubernetes.io/cors-allow-origin: "https://example.com"
    nginx.ingress.kubernetes.io/session-cookie-name: "route"
spec:
  ingressClassName: nginx  # ← Traefik will watch this class
  rules:
    - host: myapp.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: whoami
                port:
                  number: 80

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whoami
spec:
  replicas: 2
  selector:
    matchLabels:
      app: whoami
  template:
    metadata:
      labels:
        app: whoami
    spec:
      containers:
        - name: whoami
          image: traefik/whoami
          ports:
            - containerPort: 80

---
apiVersion: v1
kind: Service
metadata:
  name: whoami
spec:
  selector:
    app: whoami
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
```
````

File: docs/content/migrate/nginx-to-traefik.md [hunk 29]
```diff
@@ -0,0 +1,615 @@
...
+2. Upgrade NGINX Helm release to preserve the IngressClass on uninstall
+3. Add Traefik LoadBalancer to DNS
+4. Progressively shift traffic from NGINX to Traefik
+5. Remove NGINX from DNS and uninstall
+
+---
+
+## Step 1: Install Traefik Alongside NGINX
+
+Install Traefik with the Kubernetes Ingress NGINX provider enabled. Both controllers will serve the same Ingress resources simultaneously.
```
darkweaver87, line 155: For testing this procedure should we add the install command for nginx:

```
# Install NGINX Ingress Controller
helm upgrade --install ingress-nginx ingress-nginx \
  --repo https://kubernetes.github.io/ingress-nginx \
  --namespace ingress-nginx --create-namespace
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+
+Before adding Traefik to DNS, verify it correctly serves your Ingress resources.
+
+### Test via Traefik's LoadBalancer IP
+
+Get Traefik's LoadBalancer IP and use `--resolve` to test without changing DNS:
+
+```bash
+# Get Traefik's LoadBalancer IP
+TRAEFIK_IP=$(kubectl get svc -n traefik traefik -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
```
darkweaver87:
```suggestion
NGINX_IP=$(kubectl get svc -n ingress-nginx ingress-nginx-controller -o go-template='{{ $ing := index .status.loadBalancer.ingress 0 }}{{ if $ing.ip }}{{ $ing.ip }}{{ else }}{{ $ing.hostname }}{{ end }}')

TRAEFIK_IP=$(kubectl get svc -n traefik traefik -o go-template='{{ $ing := index .status.loadBalancer.ingress 0 }}{{ if $ing.ip }}{{ $ing.ip }}{{ else }}{{ $ing.hostname }}{{ end }}')
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+## Step 2: Verify Traefik Is Handling Traffic
+
+Before adding Traefik to DNS, verify it correctly serves your Ingress resources.
+
+### Test via Traefik's LoadBalancer IP
+
+Get Traefik's LoadBalancer IP and use `--resolve` to test without changing DNS:
+
+```bash
+# Get Traefik's LoadBalancer IP
```
darkweaver87:
```suggestion
# Get LoadBalancer IPs
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+TRAEFIK_IP=$(kubectl get svc -n traefik traefik -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+echo "Traefik IP: $TRAEFIK_IP"
```
darkweaver87:
```suggestion
echo -e "Nginx IP: $NGINX_IP\nTraefik IP: $TRAEFIK_IP"
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+
+# Test HTTP
```
darkweaver87:
```suggestion
# Test HTTP for both
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+curl --resolve myapp.example.com:80:$TRAEFIK_IP http://myapp.example.com/
```
darkweaver87:
```suggestion
# Observe HTTPS redirections:
curl --connect-to myapp.example.com:80:${NGINX_IP}:80 http://myapp.example.com/ -D -
curl --connect-to myapp.example.com:80:${TRAEFIK_IP}:80 http://myapp.example.com/ -D - # note X-Forwarded-Server which should be traefik
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+
+# Test HTTPS (use -k to skip certificate verification if using self-signed certs)
+curl --resolve myapp.example.com:443:$TRAEFIK_IP https://myapp.example.com/
```
darkweaver87:
```suggestion
curl --connect-to myapp.example.com:443:$NGINX_IP:443 https://myapp.example.com/
curl --connect-to myapp.example.com:443:$TRAEFIK_IP:443 https://myapp.example.com/
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+
+```yaml tab="Your Existing Ingress (No Changes Needed)"
+
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: myapp
+  annotations:
+    # These NGINX annotations are automatically translated by Traefik
+    nginx.ingress.kubernetes.io/ssl-redirect: "true"
```
darkweaver87, line 34: Without `force-ssl-redirect` the redirection will be sent even if a valid certificate doesn't exist:
```suggestion
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+
+Check Traefik logs to confirm it discovered your Ingress resources:
+
+```bash
+kubectl logs -n traefik deployment/traefik | grep -i "ingress"
+```
+
+---
+
+## Step 3: Preserve the IngressClass
```
darkweaver87: Uninstalling Nginx should be the last step. I don't understand why you talk about this here.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+### Option A: DNS-Based Migration
+
+Add the Traefik LoadBalancer IP to your DNS records alongside NGINX. This allows both controllers to receive traffic.
+
+**Get LoadBalancer addresses:**
+
+```bash
+# NGINX LoadBalancer
+kubectl get svc -n ingress-nginx ingress-nginx-controller \
+  -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
```
darkweaver87:
```suggestion
echo $(kubectl get svc -n ingress-nginx ingress-nginx-controller -o go-template='{{ $ing := index .status.loadBalancer.ingress 0 }}{{ if $ing.ip }}{{ $ing.ip }}{{ else }}{{ $ing.hostname }}{{ end }}')
```

it's more generic

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+
+# Traefik LoadBalancer
+kubectl get svc -n traefik traefik \
+  -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
```
darkweaver87:
```suggestion
echo $(kubectl get svc -n traefik traefik -o go-template='{{ $ing := index .status.loadBalancer.ingress 0 }}{{ if $ing.ip }}{{ $ing.ip }}{{ else }}{{ $ing.hostname }}{{ end }}')
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,615 @@
...
+                 → LoadBalancer → Traefik → Your Services
+
+Final:       DNS → LoadBalancer → Traefik → Your Services
+```
+
+**Migration Flow:**
+
+1. Install Traefik alongside NGINX (both serving traffic in parallel)
+2. Upgrade NGINX Helm release to preserve the IngressClass on uninstall
+3. Add Traefik LoadBalancer to DNS
```
darkweaver87:
```suggestion
3. Add Traefik LoadBalancer to DNS (if you choose DNS option; cf. step 3)
```

darkweaver87 (commented):
Review so far :-)

docs/content/migrate/nginx-to-traefik.md
xxsheddy: we are assuming that folks already have ingress-nginx running one way or the other. that's why it is a migration guide.

docs/content/migrate/nginx-to-traefik.md
xxsheddy: i don't think that's necessary

docs/content/migrate/nginx-to-traefik.md
xxsheddy: true, fixed

docs/content/migrate/nginx-to-traefik.md
darkweaver87: I'm not opinionated on this :-)

docs/content/migrate/nginx-to-traefik.md
darkweaver87: I'm pretty sure some people will want to test this procedure from end to end before even migrating their dev environment so having it on hand can be useful. Again, I'm not strongly opinionated on this :-)

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,626 @@
...
+
+TRAEFIK_IP=$(kubectl get svc -n traefik traefik -o go-template='{{ $ing := index .status.loadBalancer.ingress 0 }}{{ if $ing.ip }}{{ $ing.ip }}{{ else }}{{ $ing.hostname }}{{ end }}')
+echo -e "Nginx IP: $NGINX_IP\nTraefik IP: $TRAEFIK_IP"
+
+# Test HTTP for both
+# Observe HTTPS redirections:
+curl --connect-to myapp.example.com:80:${NGINX_IP}:80 http://myapp.example.com/ -D -
+curl --connect-to myapp.example.com:80:${TRAEFIK_IP}:80 http://myapp.example.com/ -D - # note X-Forwarded-Server which should be traefik
+
+# Test HTTPS (use -k to skip certificate verification if using self-signed certs)
```
darkweaver87: Here we probably should emphasize the fact that `-k` is a bad idea ... and this shouldn't be used on production. At least the behavior should be identical: if we don't run those tests with `-k` NGINX we shouldn't do it for Traefik.
As Traefik won't be exposed before migrating, either people is using an external tool (e.g. cert-manager) to manage certificates and they can use `tls.secretName` or they can use let's encrypt provider for Traefik but only DNS challenge solution will work !

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,626 @@
...
+echo $(kubectl get svc -n traefik traefik -o go-template='{{ $ing := index .status.loadBalancer.ingress 0 }}{{ if $ing.ip }}{{ $ing.ip }}{{ else }}{{ $ing.hostname }}{{ end }}')
+```
+
+**Progressive DNS migration:**
+
+1. **Add Traefik to DNS** - Add the Traefik LoadBalancer IP to your DNS records (both IPs now receive traffic via round-robin)
+2. **Monitor** - Observe traffic patterns on both controllers
+3. **Remove NGINX from DNS** - Once confident, remove the NGINX LoadBalancer IP from DNS
+4. **Wait for DNS propagation** - Allow time for DNS caches to expire
+5. **Uninstall NGINX** - Proceed to [Step 4](#step-4-uninstall-nginx-ingress-controller)
```
darkweaver87: We should warn people that some ISP don't respect DNS TTLs for making savings on traffic costs.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,626 @@
...
+    5. **[Enable `publishService` on Traefik](#step-1-install-traefik-alongside-nginx)** and [uninstall NGINX](#step-5-uninstall-nginx-ingress-controller)
+
+### Option B: External Load Balancer with Weighted Traffic
+
+For more control over traffic distribution, use an external load balancer (like Traefik, Cloudflare, AWS ALB, or a dedicated load balancer) in front of both Kubernetes LoadBalancers.
+
+**Setup:**
+
+1. Create an external load balancer pointing to the NGINX Kubernetes LoadBalancer
+2. Update DNS to point to the external load balancer
```
darkweaver87, line 327: Here again, if the infrastructure was not designed that way, they should do this change prior of doing this migration.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,626 @@
...
+| Increase | 50% | 50% | 1 hour |
+| Near-complete | 10% | 90% | 1 hour |
+| Final | 0% | 100% | - |
+
+!!! tip "External Load Balancer Options"
+
+    - **Cloudflare Load Balancing** - Traffic steering with health checks
+    - **AWS Global Accelerator** - Weighted routing across endpoints
+    - **Google Cloud Load Balancing** - Traffic splitting
+    - **Traefik / HAProxy / NGINX (external)** - Self-hosted option with weighted backends
```
darkweaver87, line 347: Some physical machines / VMs with either:
- CARP virtual IP 
- anycast BGP 
- ...

with a Traefik installed on those doing the weighted load balancing is also an option. This list looks like exhaustive :-)
```suggestion
    - **Traefik / HAProxy / NGINX (external)** - Self-hosted option with weighted backends
    - ...
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,626 @@
...
+5. Upgrade Traefik to claim the released IP
+6. (Optional) Remove Traefik's old IP from DNS once the new IP is active
+
+This way, traffic is always flowing to Traefik during the IP transfer.
+
+**Get your current NGINX LoadBalancer IP:**
+
+```bash
+kubectl get svc -n ingress-nginx ingress-nginx-controller \
+  -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
```
darkweaver87:
```suggestion
kubectl get svc -n ingress-nginx ingress-nginx-controller -o go-template='{{ $ing := index .status.loadBalancer.ingress 0 }}{{ if $ing.ip }}{{ $ing.ip }}{{ else }}{{ $ing.hostname }}{{ end }}'
```

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,626 @@
...
+```
+
+!!! tip "Zero Downtime During Helm Upgrade"
+
+    The Helm upgrade only restarts the Traefik pod, not the LoadBalancer service. Traefik uses a `RollingUpdate` deployment strategy by default, so the new pod starts before the old one terminates. For additional safety, run multiple Traefik replicas:
+
+    ```yaml
+    # In traefik-values.yaml
+    deployment:
+      replicas: 2
```
darkweaver87, line 479: Replicas is a good start but they should also consider affinity rules and PDB.

e.g. https://github.com/traefik/infra/blob/master/manifests/base/core/traefik-hub-helmchart/common/helm-values.yaml#L147 + PDB (which can be set by the helm chart values files)

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -0,0 +1,626 @@
...
+    apiVersion: networking.k8s.io/v1
+    kind: IngressClass
+    metadata:
+      name: nginx
+    spec:
+      controller: k8s.io/ingress-nginx
+    EOF
+    ```
+
+### Delete NGINX Admission Webhook
```
darkweaver87, line 557: If not installed through helm

nmengin (approved):
LGTM

Thank you @sheddy-traefik 📚

traefiker:
:no_entry_sign: error related to review: CHANGES_REQUESTED by immanuelfodor

mloiseleur (dismissed):
LGTM

PR #12319, opened 2025-11-26
Bump github.com/quic-go/quic-go to v0.57.1

GreyXor:
### What does this PR do?

Update quic-go to fix the QUIC events with session tickets disabled on Go 1.26

### Motivation

> With https://go-review.googlesource.com/c/go/+/528755 (which will be part of Go 1.26), `tls.QUICConn.SendSessionTicket` now won't return an error when session tickets are disabled via `tls.Config`. Instead, it will just return a `tls.QUICNoEvent`. The current code panics when encountering this condition. Instead, we should just proceed (and not send a session ticket to the client).

https://github.com/quic-go/quic-go/pull/5462
https://github.com/quic-go/quic-go/releases/tag/v0.57.1

### More

- [ ] Added/updated tests


rtribotte (approved):
Thanks!

PR #12330, opened 2025-11-28
Bump webui to NodeJS v24

jnoordsij:
### What does this PR do?

This pull request updates the web UI NodeJS version to v24 (current LTS).

See https://nodejs.org/en/blog/release/v24.11.0

### Motivation

Be up-to-date

### More

- [ ] Added/updated tests

### Additional Notes

The current `Dockerfile` tags both here and on `v3` are quite restrictive in their respective NodeJS versions;
using a more loose tag is probably slightly neater.
A first quick glance seems to suggest everything is still fine on the latest LTS,
but I haven't extensively tested just yet.

As `v3` differs heavily from `v2` (new web UI), different changes are required there.
If wanted I can create a dedicated MR for changes there.


rtribotte (approved):
LGTM

PR #12337, opened 2025-11-30
Add DefaultCertificate coverage for getUncheckedDomains

intojhanurag:
### What does this PR do?

Adds a new test case to validate that when a **Default (fallback) TLS certificate exists in the certificate store**, `getUncheckedDomains()` correctly returns **an empty list**, ensuring Traefik does not attempt unnecessary ACME certificate generation for covered domains.

### Motivation

The test suite already covers wildcard and domain certificate scenarios, but lacked coverage for the DefaultCertificate case. Adding this test ensures future changes do not break fallback certificate behavior and prevents regression in ACME domain-checking logic.

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/provider/acme/provider_test.go
+++ b/pkg/provider/acme/provider_test.go
@@ -162,6 +162,20 @@ func TestGetUncheckedCertificates(t *testing.T) {
 			},
 			expectedDomains: []string{"acme.wtf"},
 		},
+		{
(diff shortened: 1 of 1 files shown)


rtribotte:
Hey @intojhanurag,

Thanks for this!

PR #12355, opened 2025-12-03
Adjust debug log message for TLS sanitization

kuritka:
While debugging a TLS-related issue, I initially assumed that a certificate was missing because it did not appear in the debug configuration log. After some time I realized that Traefik intentionally hides TLS fields when logging the dynamic configuration.

To avoid similar confusions, I adjusted the debug log message to explicitly state that TLS data is sanitized. This is a small, non-functional change that improves clarity during debugging.

Signed-off-by: Michal K <kuritka@gmail.com>


nmengin:
Hey @kuritka,

Thanks for the contribution!

nmengin:
Hey @kuritka,

Thanks for the contribution and for raising this need.
After discussing the PR with the other maintainers, we agree that the feature request is valid. That said, the current implementation only partially addresses the problem, as it focuses on a single log while similar data may still appear elsewhere.

We think a more comprehensive solution would be to provide users with a redacted version of the certificates.

Would you be open to reworking the PR in that direction? Otherwise, we suggest closing this PR and opening a dedicated issue so we can track and discuss the broader requirement more effectively.

kuritka:
Hi @nmengin, Thanks for the detailed clarification!

I'm ok to rework the PR in the broader direction you suggested.
It won't be immediate, but I can take it on and work toward a more complete redaction mechanism that applies consistently across all relevant logs and configuration outputs.

I'll keep the PR open and update it once I have progress.  
Thanks again for the guidance!

kuritka:
Hi @nmengin,

configuration.TLS is not really used elsewhere from a logging perspective.
I extended the configuration class and added a TLS Summary.
Now, within a single log message, I can return both:

the sanitized TLS configuration (as it was before), and

an additional TLS summary.

To give a clearer picture, I’m including an example of how this looks in practice
(the JSON is formatted to improve readability).


```log
2025-12-25T17:37:59+01:00 DBG pkg/server/configurationwatcher.go:194 > Configuration received (TLS redacted)
config={
  "http": {
    "routers": {
      "whoami-http": {
        "entryPoints": ["web"],
        "rule": "Host(`whoami.localhost`)",
        "service": "whoami"
      },
      "whoami-https": {
        "entryPoints": ["websecure"],
        "rule": "Host(`whoami.localhost`)",
        "service": "whoami",
        "tls": {}
      }
    },
    "services": {
      "whoami": {
        "loadBalancer": {
          "passHostHeader": true,
          "responseForwarding": {
            "flushInterval": "100ms"
          },
          "servers": [
            {
              "url": "http://127.0.0.1:9090"
            }
          ],
          "strategy": "wrr"
        }
      }
    }
  },
  "tcp": {},
  "tls": {
    "stores": {
      "default": {}
    }
  },
  "udp": {}
}
providerName=file
tls_summary={
  "certificates": {
    "count": 0,
    "present": false,
    "value": "<redacted>"
  },
  "clientAuth": {
    "ca_files": "<redacted>",
    "ca_files_total": 0
  },
  "stores": {
    "default_cert": "<redacted>",
    "default_cert_count": 1,
    "names": ["default"]
  }
}
```

nmengin:
Hey @kuritka,

Thank you for your modifications.

PR #12360, opened 2025-12-04
Reject suspicious encoded characters

rtribotte:
### What does this PR do?

This PR adds a mechanism to reject request with path containing suspicious encoded characters.

### Motivation

To block requests that could induce split view scenarios between Traefik and the backends.

### More

- [x] Added/updated tests

### Additional Notes


lbenguigui (approved):
LGTM

TsengSR:
@rtribotte may I ask what triggered this PR in the first place? Its more than very odd to reject **Request paths** with %2f etc. Diese are very common, especially when redirecting or authentication  where the redirect url is passed as a query parameter and all the special character has to be encoded. 

This are the most basic things for http and absolutely required to pass these type of values as a path and still being interpreted correctly. Its even part of the http specification

rtribotte:
Hello @TsengSR,

Thanks for your question.

> @rtribotte may I ask what triggered this PR in the first place? Its more than very odd to reject Request paths with %2f etc.

Sure! We made this change to address CVE-2025-66490 (https://github.com/traefik/traefik/security/advisories/GHSA-gm3x-23wp-hc2c).

The rationale is as follows: Traefik follows [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) when evaluating routing rules. 
This means that when evaluating path matchers (`Path`, `PathPrefix`, and `PathRegex`), reserved characters (including URL delimiter characters like `/`) are kept encoded, because the encoded and decoded forms don't have the same semantic meaning per [RFC 3986 Section 2.4](https://datatracker.ietf.org/doc/html/rfc3986#section-2.4).

Unfortunately, not all backend servers behave this way, and some decode reserved characters.
This creates a "split view" situation where routing in Traefik can be bypassed to access backend paths that should be protected by another router's rules.

As a solution, and to remain secure by default, we now block requests containing encoded characters in the that can be used to create such split view situations.

> These are the most basic things for http and absolutely required to pass these type of values as a path and still being interpreted correctly. Its even part of the http specification

We agree that percent-encoded slashes are valid per [RFC 3986 Section 2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1).
However, when Traefik is used to secure specific backend paths (which is a common use case), it becomes Traefik's responsibility to prevent split view situations, even when the request URL strictly conforms to RFC 3986.
We find it unfortunate, but the security risk of access control bypass outweighs the compatibility cost in this case.

However, and it is possible that there is a confusion here, the rejection is based on what the request path contains.
These means that query parameters can contain encoded hashes characters.

There are several example of servers/framework that are also blocking the request similarly if a given set of encoded characters is present in the path.
For instance, encoded slashed (`%2F`, `%5C`) are blocked by Apache Tomcat, Apache HTTP Server, and Spring Security.

I hope this clarifies the related issue and our approach to solve it.

fpotier:
> Hello @TsengSR,
> 
> Thanks for your question.
> 
> > @rtribotte may I ask what triggered this PR in the first place? Its more than very odd to reject Request paths with %2f etc.
> 
> Sure! We made this change to address CVE-2025-66490 ([GHSA-gm3x-23wp-hc2c](https://github.com/traefik/traefik/security/advisories/GHSA-gm3x-23wp-hc2c)).
> 
> The rationale is as follows: Traefik follows [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) when evaluating routing rules. This means that when evaluating path matchers (`Path`, `PathPrefix`, and `PathRegex`), reserved characters (including URL delimiter characters like `/`) are kept encoded, because the encoded and decoded forms don't have the same semantic meaning per [RFC 3986 Section 2.4](https://datatracker.ietf.org/doc/html/rfc3986#section-2.4).
> 
> Unfortunately, not all backend servers behave this way, and some decode reserved characters. This creates a "split view" situation where routing in Traefik can be bypassed to access backend paths that should be protected by another router's rules.
> 
> As a solution, and to remain secure by default, we now block requests containing encoded characters in the that can be used to create such split view situations.
> 
> > These are the most basic things for http and absolutely required to pass these type of values as a path and still being interpreted correctly. Its even part of the http specification
> 
> We agree that percent-encoded slashes are valid per [RFC 3986 Section 2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1). However, when Traefik is used to secure specific backend paths (which is a common use case), it becomes Traefik's responsibility to prevent split view situations, even when the request URL strictly conforms to RFC 3986. We find it unfortunate, but the security risk of access control bypass outweighs the compatibility cost in this case.
> 
> However, and it is possible that there is a confusion here, the rejection is based on what the request path contains. These means that query parameters can contain encoded hashes characters.
> 
> There are several example of servers/framework that are also blocking the request similarly if a given set of encoded characters is present in the path. For instance, encoded slashed (`%2F`, `%5C`) are blocked by Apache Tomcat, Apache HTTP Server, and Spring Security.
> 
> I hope this clarifies the related issue and our approach to solve it.

I just found this PR after hours of troubleshooting and don't understand how this was flagged as a bug fix in the changelog (I'm using 3.6 where the changelog doesn't even mention it, just merge branch 2.x to 3.6) while this clearly breaks existing deployments relying on the previous behavior.

rtribotte:
Hello @fpotier,

Thank you for sharing this with us, and sorry for the inconvenience.

We have opened https://github.com/traefik/traefik/pull/12388 to make the changelog more accurate in that regard, and also edit the release on Github too.

TsengSR:
> Hello @TsengSR,
> 
> Thanks for your question.
> 
> > @rtribotte may I ask what triggered this PR in the first place? Its more than very odd to reject Request paths with %2f etc.
> 
> Sure! We made this change to address [CVE-2025-66490](https://github.com/advisories/GHSA-gm3x-23wp-hc2c) ([GHSA-gm3x-23wp-hc2c](https://github.com/traefik/traefik/security/advisories/GHSA-gm3x-23wp-hc2c)).

If I understand the issue correctly, then rejecting the request is the wrong way to handle this.

Something else is going wrong here. I mean if the matching is made for `http://mydomain.example.com/admin%2F` and the route matches `/admin/` then the previous behaviour is correct. `/admin%2F` is **NOT** the same as `/admin/`. In later one the `/ `is a path delimiter. In the former one the path segment is named `admin/` but because `/` is a special character in path the browser (or caller) encodes it, which is correct. To match this case, one would have to disable access fo `/` and then  allow for `/admin/`.

Anways, the bypassing is one thing, but that the route is allowed to the backend and is treated as `/admin/` may suggest, that traefik is unescaping it before passing ot the backend, which should not be done. The path must be sent to the backend the way it was received. 

If traefik is not unescaping the path segment when calling the upstream service, then the http server used by the sample do not correctly handle `admin%2F` and its an issue with the framework used by the user who reported it, because if the application on the othe rend do not have a `/admin%2F` endpoint, the request should reuslt in 404 not found, whether or not it triggered the pipeline. 

Under no cicumstances should such a request be blocked, because its a valid request and its valid to escape path segments that contain reserverd characters.

@rtribotte
So I think the first step would be to analyze what exactly happens in the example request and whether traefik unescapes `%2F` to `/` when it calls the target service. If not, then everything is perfectly fine and just a configuration issue by the reporting user. 

If I configure a matching for /admin/ and `/admin%2F` arrives, its working as intended, because only `/admin/` is protected and `/admin%2F` is not `/admin/`. The `%2F` is part of the segment name, not a separator! 

I highly suspect though, that the handing of `%2F` in the reporting users framework is misinterpreting it though, but I'm not familiar enough with go to debug it. I know that ASP.NET Core (and Kestrel its based on) correctly handle escaped path segments and the above is no issue, even if you bypass the middleware, you still end up with an invalid/non-existing endpoint.

TsengSR:
Just to make it clear, when I call `http://example.com/admin%2F` and have a route matching everthing and receive the path segment as a string in ASP.NET Core Minimal API, the passed parameter is `admin/` and not `admin`

```csharp
var builder = WebApplication.CreateBuilder(args);
var app = builder.Build();
app.Run();
app.MapGet("/{segment}", (string segment) => {
    return $"The segment you provided is: {segment}";
});

app.Run();
```

This prints `The segment you provided is: admin%2F` and not `The segment you provided is: admin/` or `The segment you provided is: admin`!

image

if the path would been `/admin/account/resetpassword` and the users tried to bypass the middleware with `/admin%2Faccount/resetpassword` then yes, the middleware woudl be bypassed, but the user would have received an `404 not found` as response, because ASP.NET Core would search for a `admin%2Faccount` route, but it only has a `admin/account` route on its `AccountController.cs` and would not be able to reach that endpoint. 

Thats when the http servers handles it correcltly, but again, its an security issue of the receiving http server, not of traefik. And while such a feature could be useful it should be strictly opt-in (per route, not per endpoint or per traefik instance) as it highly increases the configuration efforts (and breaks 100s of existing installations of course). So in my opinion its not an security issue in traefik itself, the way traefik handled it seems in accordance to http specification and the behavior in the Github Advisory seems to be expected behavior, as long as the upstream service http is correctly handling encoded path segments

fpotier:
> Hello @fpotier,
> 
> Thank you for sharing this with us, and sorry for the inconvenience.
> 
> We have opened #12388 to make the changelog more accurate in that regard, and also edit the release on Github too.

thanks for the quick answer, it might be bad configuration on my side but I also found the 400 error with no message confusing as I would never expect a reverse proxy to return a 400. I also couldn't see anything in the logs but again it might be misconfiguration from my side.

lastzero:
We understand that you wanted to resolve [CVE-2025-66490](https://github.com/advisories/GHSA-gm3x-23wp-hc2c) quickly, especially since Traefik is set to become the Kubernetes standard. However, please do not include **major breaking changes** like this in **minor patch releases** next time, as it essentially caused a denial of service for affected users.

If such changes are necessary and [semantic versioning](https://semver.org/)  cannot be used, consider displaying a proper error message in the server response so that the issue can be identified and resolved as quickly as possible.

**Thank you very much!** :pray:

rtribotte:
Hi @TsengSR,

Thanks for this feedback.
I understand your point about RFC compliance - you're right that `%2F` and / are semantically different, and blocking valid percent-encoded characters is unfortunate.
However, the issue isn't Traefik unescaping paths - we don't. The problem is that many backends do decode `%2F` back to `/` when routing. So Traefik correctly rejects `/admin%2F` (not matching `/admin/`), but the backend treats it as `/admin/` and grants access. That's the split-view bypass.

RFC 3986 actually supports this approach: [Section 7.3](https://datatracker.ietf.org/doc/html/rfc3986#section-7.3) and the broader security considerations establish that implementers should restrict URI elements when they could be misinterpreted by downstream systems in ways that bypass security. Encoded slashes are exactly this scenario - they're legal per the RFC, but their different interpretation between Traefik and backends creates the security risk the RFC cautions against.

Yes, ideally every backend should handle this correctly like ASP.NET Core does in your example. But many don't - Tomcat, Apache, and Spring Security all block encoded slashes by default for this exact reason. When users deploy Traefik to secure backends they don't control, we need to prevent these bypasses.

We chose secure-by-default because the risk of unauthorized access outweighs compatibility cost. But we've added allowEncodedCharacters for cases where you've verified your backend's behavior or don't use path-based security.
Also worth noting: query parameters can still contain encoded characters - only the path is restricted. So most OAuth/redirect scenarios should still work.

@fpotier Thanks for your feedback, regarding the status code, what would you have expected from Traefik?

TsengSR:
> Hi @TsengSR,
> 
> Thanks for this feedback. I understand your point about RFC compliance - you're right that `%2F` and / are semantically different, and blocking valid percent-encoded characters is unfortunate. However, the issue isn't Traefik unescaping paths - we don't. The problem is that many backends do decode `%2F` back to `/` when routing. So Traefik correctly rejects `/admin%2F` (not matching `/admin/`), but the backend treats it as `/admin/` and grants access. That's the split-view bypass.

Thats what I suspected and this would need to be addressed in the backends framework (whether it may be go, nodejs, java or whatever the reporting user was using). 

> RFC 3986 actually supports this approach: [Section 7.3](https://datatracker.ietf.org/doc/html/rfc3986#section-7.3) and the broader security considerations establish that implementers should restrict URI elements when they could be misinterpreted by downstream systems in ways that bypass security. Encoded slashes are exactly this scenario - they're legal per the RFC, but their different interpretation between Traefik and backends creates the security risk the RFC cautions against.

Yea, but its for the backends to fix it. This is the only correct place to handle it. While its a useful feature to have, for emergency, it should strictly be opt-in and **per route**, not per endpoint, so one can enable it for **one specific** application that requires this kind of extra protection, because I know there are a lot of abandomed/unmaintained frameworks out there or legacy applications that are not receiving further updates.

> Yes, ideally every backend should handle this correctly like ASP.NET Core does in your example. But many don't - Tomcat, Apache, and Spring Security all block encoded slashes by default for this exact reason. When users deploy Traefik to secure backends they don't control, we need to prevent these bypasses.

Yes, but then its not something that should be solved in traefik. At least not with opt-out. Leave it opt-in for people who want to harden their system and use deprecated/unmaintained legacy frameworks that wont get a fix in future. If they directly espose their application with nodeport they would be suspectable to the same issue. 

> We chose secure-by-default because the risk of unauthorized access outweighs compatibility cost. But we've added allowEncodedCharacters for cases where you've verified your backend's behavior or don't use path-based security. Also worth noting: query parameters can still contain encoded characters - only the path is restricted. So most OAuth/redirect scenarios should still work.

But you cant secure other applications fault, especially if it breaks a major number of working use cases. Its their job to fix these. Fixing it in traefik only helps mitigate it when traefik is used. Any other reverse proxy (HAproxy, nginx) would still let it pass and the backend would still have the same issue. 

Thats not securty by design, its security by obscurity, "removing"  the need/excuse of the actual framework maintainer to address the issue.

kevinpollet:
Hello @fpotier, we opened https://github.com/traefik/traefik/pull/12394 wdyt?

TsengSR:
> Hello @fpotier, we opened #12394 wdyt?

I still think its the wrong way to solve it by breaking it for everyone and having people to opt out of it. And the opt-out happens on the endpoint level, which is really really bad and allows no control at all. 

What speaks against opt-in per route. This allows people affected by it to opt-in the feature for legacy applications and frameworks which are likely not to get a fix ever. 

Everything else needs to be addressed by the backends http server/framework.

fpotier:
> Hello @fpotier, we opened #12394 wdyt?

Hi,

After thinking about it I'm unsure what would be the best solution.
Adding the error message would have made it easier for me to debug but on the other end in production I don't know if people really want to inform potential attackers that they are running a traefik reverse proxy and that their request is being filtered by it because it contains suspicious characters.

For instance when blocking requests based on geolocation some sites return a specific page "Not available in your country" and some simply return 404, so maybe making the response configurable would be the best.

I also think that @TsengSR has a point, this setting should have been opt-in

emilevauge:
Hey here, I understand our statement cannot please everyone, however, that's the decision we made after many conversations with the maintainers team. Envoy emitted a [similar CVE](https://github.com/envoyproxy/envoy/security/advisories/GHSA-4987-27fx-x6cf) and made the decision to add an optional check, disabled by default. But most users will never enable this options, as it's quite complex to understand and will always prefer to stick with the defaults options. That's why we prefer to be safe by default, which is a good thing for 99% of use cases, while allowing anyone to disable this behavior if you really need to handle those specific encoded characters in the path.

Regarding @lastzero 's [comment](https://github.com/traefik/traefik/pull/12360#issuecomment-3631198670), we are going to release a bug fix soon adding a log to help troubleshooting this kind of situation. Thanks for the suggestion 👍

TsengSR:
> Hey here, I understand our statement cannot please everyone, however, that's the decision we made after many conversations with the maintainers team. Envoy emitted a [similar CVE](https://github.com/envoyproxy/envoy/security/advisories/GHSA-4987-27fx-x6cf) and made the decision to add an optional check, disabled by default. 

Sorry, I fully disagre with everything said . Its not a good option. It somewhat shows me, that you are not using your own product of you don't see the issues caused by this. This issue is a massive one, and the ability to control at the endpoint level, is and will not work. 

Let's assume you have a legacy application that is affected by this and could need this type of "protection" (since its not a real one, its not fixing the underlying issue of the receiving http server handling it wrong) and another one that is blocked. There is no way to solve that since the endpoint will in 99.9% cases be 80 and 443, and you can't fix that in any way. 

Per route setting is the only one which makes sense. Breaking the protocol / specification because another software misinterpret is, is plain and simple wrong. Not only will it make your lifes a hell to maintain this with future requirements and fixes and when others http servers fix that issues, yours will still block a "non-issue"

> But most users will never enable this options, as it's quite complex to understand and will always prefer to stick with the defaults options. That's why we prefer to be safe by default, which is a good thing for 99% of use cases, while allowing anyone to disable this behavior if you really need to handle those specific encoded characters in the path.

Its not meant for most users. It's meant for the one who encounter that issue and can't upgrade their legacy application (3rd party application) or whose applications use defunct/abandoned http servers. Which is a minority. 

If you start "fixing issues" and blocking these by default that every other frameworks fucks up with their messy implementations, you'dd endup with 100s of fixes related to other frameworks flaws. That can't be your goal, can it?

After all, last time I looked up for the definition of a reverse proxy, it was task were to redirect traffic according to rules, not to act as a firewall. 

That's what Web Applcation Firewalls (WAF) like "modsecurity" (that are/were available for a lot of reverse proxyies) which focus specifically on this type of attack detection and prevention. Or [open-appsec](https://www.openappsec.io/post/open-appsec-integration-with-envoy-proxy-docker-now-available-in-beta) which offers an envoy integration. 
Anoter WAF is [Coraza](https://github.com/corazawaf/coraza). They are all on specialized to handle this type of cases or other common Web application based attacks. 

This is not something a reverse proxy has to bother with, its not one of its responsibilities.

emilevauge:
Thank you for sharing your opinion @TsengSR , we got your point.
Let's stay on this for now, and wait for more feedback from the community.

lastzero:
Here's a technical summary of CVE-2025-66490 and the changes Traefik implemented in response for anyone who doesn't want to read the original CVE and all the comments:

1. **The advisory overstates what Traefik actually does.**  
   The [CVE text suggests](https://github.com/advisories/GHSA-gm3x-23wp-hc2c) Traefik decodes the request path before forwarding it, causing `/admin%2F` to reach the backend as `/admin/`. In reality, Traefik routes on the *encoded* path and forwards it *as-is*. If the backend sees `/admin/`, it is because the backend itself chooses to decode `%2F` to `/`, not because Traefik modified the path.
2. **The real issue is a "split view" with certain backends, not Traefik alone.**  
   A bypass only exists in a specific setup:
   - Traefik is used as a **security boundary** (e.g., path-based auth or protection on `/admin/`), and  
   - There is another, less-protected router (e.g., `/`) on the same entrypoint, and  
   - The backend **decodes** reserved characters in the path (e.g., treats `/admin%2F` as `/admin/`).
   
   In that environment, a request like `/admin%2F` may skip the protected router in Traefik but still be interpreted as `/admin/` by the backend. If your backend does *not* decode this way, you are not affected by the original issue — only by the new blocking behavior.
3. **Why Traefik added a breaking "HTTP firewall-ish" fix anyway.**  
   The Traefik team chose to treat this as in-scope because many users rely on Traefik as part of their security perimeter (reverse proxy + access control). To prevent split-view bypasses in that threat model, they introduced **encoded-character filtering** at the entrypoint level and block certain encoded characters (like `%2F`) by default. This is explicitly a *security-first* design choice, not a pure HTTP transparency choice.
4. **Why this breaks legitimate, HTTP-compliant applications.**  
   Before the change, applications that legitimately expected to receive `%2F` in the path (and could decide for themselves whether to decode it) worked correctly behind Traefik. However, after the change, these applications no longer function properly:
   - These requests are blocked **before** they reach the app.  
   - The filtering is configured **per entrypoint**, not per router or service.  
   - That means on a given host/port you must choose between:
     - allowing encoded characters (restoring previous behavior and risk profile), or  
     - blocking them (protecting certain backends, but breaking apps that need strict HTTP compliance).
   
   In practice, if you have one app that requires full RFC-compliant handling of encoded paths and another that relies on Traefik as a firewall to block suspicious encodings, you cannot satisfy both on the same entrypoint without re-architecting. For many setups, this feels like an effective denial of service imposed by the mitigation, not by the original "vulnerability".
5. **Bottom line for Traefik users.**  
   - The CVE’s severity is based on a threat model where Traefik is a security gateway for path-based access control in front of backends that decode reserved characters.  
   - For deployments that use Traefik primarily as a transparent reverse proxy or load balancer, and where backends do not perform risky decoding, the "vulnerability" is largely theoretical. The main impact is the **breaking change** from the mitigation.  
   - If your apps legitimately use encoded characters in paths, you will likely need to:
     - re-enable them explicitly on the affected entrypoint(s), and/or  
     - separate "firewalled" and "HTTP-compliant" apps onto different entrypoints or layers.

In other words: Traefik behaved correctly and transparently according to HTTP before this CVE. The fix is a security-hardening measure for a specific class of backend behaviors and threat models, but it comes at the cost of compatibility for existing deployments that relied on strict HTTP compliance.

IMHO, the **6.9/10 score** for [CVE-2025-66490](https://github.com/advisories/GHSA-gm3x-23wp-hc2c) is too high because “Attack Requirements: None" is incorrect, as is the original claim that Traefik forwards requests after decoding them.

psifls:
We operate Traefik across many “production” and “testing” environments, with Watchtower handling automated patch-level updates. The recent Traefik patch release caused a large outage across multiple (partly legacy) services, including parts of our Jenkins infrastructure. Identifying the root cause took time because we initially didn’t expect a patch release to introduce breaking changes and 400 might not have been the clearest error.

We have now mitigated the issue by disabling the new checks at the entrypoint level. However, it’s unclear why this configuration is only available globally per entrypoint and not overrideable per route. Having defaults defined on the entrypoint, with the ability to override them per route (e.g. via Docker labels), would provide much finer control and help prevent similar outages in mixed environments.

emilevauge:
Thanks for your feedback @psifls, sorry for the inconvenience. 
We are WIP on making this patch less breaking. Could you tell us what kind of [rules](https://doc.traefik.io/traefik/reference/routing-configuration/http/routing/rules-and-priority/) are mostly used in your environments amongst those ones:

- Header(`key`, `value`)
- HeaderRegexp(`key`, `regexp`)
- Host(`domain`)
- HostRegexp(`regexp`)
- Method(`method`)
- Path(`path`)
- PathPrefix(`prefix`)
- PathRegexp(`regexp`)
- Query(`key`, `value`)
- QueryRegexp(`key`, `regexp`)
- ClientIP(`ip`)

macno:
FYI we've been bite too (i.e. gitlab instances behind traefik started returning 400 without body) after renovate merged the patch

> Could you tell us what kind of [rules](https://doc.traefik.io/traefik/reference/routing-configuration/http/routing/rules-and-priority/) are mostly used in your environments

```
Host(`fqdn`) -> 52%
Host(`fqdn`) && PathPrefix(`/prefix`) -> 26%
Host(`fqdn`) && PathRegexp(`regex`) -> 17%
HostRegexp(`regex`) -> 5%
```

psifls:
Most of our rules are based on 
> * Host(`domain`)
> * Path(`path`)

and some one

> * PathPrefix(`prefix`)
> * PathRegexp(`regexp`)

Az4zel99:
We had the same happen as described here. Jenkins wouldn't work, and neither would terraform accessing GitLab.
Our Traefik has access logs enabled and the 400 errors were just silently discarded. I would expect that they'd at least be noted down in the access logs when that is enabled, which would have helped us find the issue sooner.

> We operate Traefik across many “production” and “testing” environments, with Watchtower handling automated patch-level updates. The recent Traefik patch release caused a large outage across multiple (partly legacy) services, including parts of our Jenkins infrastructure. Identifying the root cause took time because we initially didn’t expect a patch release to introduce breaking changes and 400 might not have been the clearest error.
> 
> We have now mitigated the issue by disabling the new checks at the entrypoint level. However, it’s unclear why this configuration is only available globally per entrypoint and not overrideable per route. Having defaults defined on the entrypoint, with the ability to override them per route (e.g. via Docker labels), would provide much finer control and help prevent similar outages in mixed environments.

emilevauge:
Hey here, in order to inform users about this new behavior, and avoid rejecting silently requests:
- we will print a WARN log when Traefik boots if [encoded characters rejection](https://doc.traefik.io/traefik/reference/install-configuration/entrypoints/#encoded-characters) is enabled
- we will also print an access log saying that a request has been rejected

We hope this will help users to quickly troubleshoot rejections.
Let us know if you have any other suggestion to complete this effort.

psifls:
> We hope this will help users to quickly troubleshoot rejections. Let us know if you have any other suggestion to complete this effort.

Thanks, that will help what would help a lot would be a reference to traefik on the 400 page. I know that from the point of Information leakage this might not be desired by everyone but you could also make this an option that people can opt out from.

macno:
what about adding a body to the response with the reason of the 400?  or add the reason to some x- header?
it will be much easier to debug (and fix)

emilevauge:
> what about adding a body to the response with the reason of the 400? or add the reason to some x- header?
it will be much easier to debug (and fix)

It would be insecure, leaking some info to a potential intruder.

TsengSR:
> Hey here, in order to inform users about this new behavior, and avoid rejecting silently requests:
> 
> * we will print a WARN log when Traefik boots if [encoded characters rejection](https://doc.traefik.io/traefik/reference/install-configuration/entrypoints/#encoded-characters) is enabled
> * we will also print an access log saying that a request has been rejected
> 
> We hope this will help users to quickly troubleshoot rejections. Let us know if you have any other suggestion to complete this effort.

This doesn't address any of the issues. To be honest is a very lazy effort instead of fixing it properly, while fallout is spreading bigger as seen on the issues. It is still massively distruptive and its a huge effort to reverse something that wasn't a reverse proxies job in the first place.

Imho the right choice would be to revert it or disable it by default and then work on a proper and opt-in solution. I hope this case will teach you why envoy decided to disable it and have it opt-in. A lot, really really lot, applications depend on this behaviour. 

You have by far underestimated the reach and distruption of this change. Even next cloud is affected, which is used in some german governments that went opensource, i.e. Schleswig-Holstein. Or argus. Or SignalR. GitLab.

With so many affected modern and still maintained software, the fallout is huge, literally forcing majority of traefik user to opt-out of this since you can only opt-out on the endpoint its competely useless for protecting legacy applications, in most cases there is one single traefik for all services of a cluster

psifls:
> > what about adding a body to the response with the reason of the 400? or add the reason to some x- header?
> > it will be much easier to debug (and fix)
> 
> It would be insecure, leaking some info to a potential intruder.

As I said then make it opt-outable, but I think almost all admins would prefer to  quickly be able to see what caused this and react accordingly. You can change the default for this to show error information to false in the next major release or maybe even a minor release but if you potentially break that many installations with a patch release than I think the risk outweigh the gains by far. I know that all should have their Log-Monitoring-Alerting set-up in a way that they will be informed as soon as traefik sends the WARN log line but to be honest a lot of people will start to search at the downstream services first before they look at their reverse proxy.

emilevauge:
Hi @TsengSR ,

Thanks for taking the time to share your thoughts and for your interest in Traefik.

> To be honest is a very lazy effort instead of fixing it properly

>  I hope this case will teach you

That said, the tone of your recent comments (including the [latest one](https://github.com/traefik/traefik/pull/12360#issuecomment-3646215016)) comes across as disrespectful. We absolutely welcome differing opinions and critical feedback — that’s a core part of open source — but we do expect discussions to remain respectful and constructive.

Please feel free to continue sharing your point of view, but do so in a way that focuses on the technical arguments and avoids personal or dismissive language. This helps keep the conversation productive for everyone involved.

Thanks for your understanding.

emilevauge:
@psifls 

> As I said then make it opt-outable

Which option do you suggest to make opt-outable? Logging the reason in the body/response header?

TsengSR:
> Please feel free to continue sharing your point of view, but do so in a way that focuses on the technical arguments and avoids personal or dismissive language. This helps keep the conversation productive for everyone involved.
> 
> Thanks for your understanding.

But then have an outside perspective on this issue. From the issue of the people who have to maintain such k8s clusters. 

This issue blocks more and more wide spread and popular software by the hour and the only thing so far we received is "We change the log message". This do not prevent the outages. This will not fix the millions caused in damage by the outages by this change. This is not sufficient. You must think of the bigger image that this change caused. And still remember, its a minor change.

psifls:
> @psifls
> 
> > As I said then make it opt-outable
> 
> Which option do you suggest to make opt-outable? Logging the reason in the body/response header?

Yes I think with the 3.6.* this there should be an information in the 400 page that gives a hint that the issue is to be searched in the reverse proxy. If I had this the services would have been back up an running minutes after the first report came in. This way I was looking at the GitLab update that happend over the weekend then at Jenkins and finally at traefik which cost me almost an hour. From 3.7.* the config that controls this "show error" behavior could be set to false so the default settings are "secure" again. This might still bite people doing a automated update but I think minor updates are reviewed a lot more often than patch updates.

emilevauge:
Point taken @psifls. @lastzero @fpotier (and others) you were also involved on this, WDYT about this proposal? Do you have any other suggestion in mind?

lastzero:
Below are some actionable ideas on how to handle the situation:

- Publish detailed information about this issue on your social media profiles and via other channels, e.g., website, newsletter, or direct emails, so that Traefik users are aware of it (if you have not already done so).
- Update your docs to mention that encoded character filtering is an optional security feature, and that Traefik is not automatically insecure when disabled. It seems this has [already happened](https://doc.traefik.io/traefik/security/request-path/#encoded-character-filtering), at least partially, so that's great. However, I would double-check to see if [other parts of the documentation](https://doc.traefik.io/traefik/reference/install-configuration/entrypoints/#encoded-characters) need improvement or clarification, too.
- Ensure that any configuration files, examples, and/or templates you provide include the relevant configuration options, along with a brief comment or link to the documentation, to raise awareness.
- Since adding a detailed error to the 400 response would leak too much information to a potential intruder, consider using a more specific response code, such as **403 Forbidden** (policy/security refusal) or **422** (request understood but rejected due to unsafe/invalid encoding), while keeping the body generic. Then, put the actionable "reason" only in logs/metrics so that admins/developers can diagnose the issue more quickly.

schildbach:
I was also bitten by this issue and it cost me a couple of hours of debugging. It affects the Matrix Chat protocols (federation and client→server), which use the hash sign in room names and forward slashes in so-called MegaOlm IDs (encryption related). Maybe more which I didn't find yet. The issue causes encryption to fail and users unable to join rooms, amongst others.

I think:

1. This major breaking change should not have been introduced in a minor point release.
2. It should be disabled by default, at least for a point release. Maybe enabled by default for a major release.
3. It should be configurable per router/service, not per entrypoint (which is way too coarse, so I can only disable it entirely). Maybe a global switch that overrides the default.
4. Rejected requests should be logged by default. I think I had to set my log level to DEBUG to see it (not sure about this one).

Apart from terminating SSL, I don't use Traefik as a security boundary, for security reasons. Traefik is a reverse proxy. So the fuzz was entirely unnecessary in my case.

monolithed:
I spent the whole day trying to figure out why npm package publishing stopped working. Why isn’t this a major version change? It breaks backward compatibility, and honestly, I don’t even know how to build the project now because package names include / (%2f).

emilevauge:
FYI, a new release [v3.6.5](https://github.com/traefik/traefik/releases/tag/v3.6.5) is available including [more logs](https://github.com/traefik/traefik/pull/12424).

lastzero:
@emilevauge Thanks! That's great. I can confirm that the following logs appear when you start Traefik v3.6.5:

```
proxy  | 2025-12-19T21:56:31Z WRN Starting with v3.6.3, Traefik now rejects some encoded characters in the request path by default. Refer to the documentation for more details: https://doc.traefik.io/traefik/migrate/v3/#encoded-characters-in-request-path
proxy  | 2025-12-19T21:56:31Z INF Traefik version 3.6.5 built on 2025-12-16T14:56:48Z version=3.6.5
proxy  | 2025-12-19T21:56:31Z INF Version check is enabled.
proxy  | 2025-12-19T21:56:31Z INF Traefik checks for new releases to notify you if your version is out of date.
proxy  | 2025-12-19T21:56:31Z INF It also collects usage data during this process.
proxy  | 2025-12-19T21:56:31Z INF Check the documentation to get more info: https://doc.traefik.io/traefik/contributing/data-collection/
proxy  | 2025-12-19T21:56:31Z INF 
proxy  | Stats collection is disabled.
proxy  | Help us improve Traefik by turning this feature on :)
proxy  | More details on: https://doc.traefik.io/traefik/contributing/data-collection/
```

**Follow-up notes/observations:**

1. While a warning on startup raises awareness (which is good), it's perplexing from a user perspective that the world's new default proxy would emit a warning about its own defaults. However, if these aren't going to revert back to the HTTP standard, then this may indeed be the best option. It's your product, so you have more insight than I do.
2.  Have you considered [revising the below warning](https://doc.traefik.io/traefik/reference/install-configuration/entrypoints/#encoded-characters) in your docs to clarify that disabling the defaults should generally not result in a security vulnerability on its own? As I explained above, most apps should not be affected. Also, please mention that keeping them enabled **may** cause software that expects encoded characters in the path to fail:
    > :warning: **Allowing certain encoded characters may expose your application to security vulnerabilities.**(https://doc.traefik.io/traefik/reference/install-configuration/entrypoints/#encoded-characters)

lazka:
What about a `encodedCharacters.allowAll` option, or similar, to make it easy to disable all of this filtering?

this is a bit much just to get a good default behavior:

```yaml
      - "--entryPoints.web.address=:80"
      - "--entryPoints.web.http.encodedCharacters.allowEncodedSlash=true"
      - "--entryPoints.web.http.encodedCharacters.allowEncodedBackSlash=true"
      - "--entryPoints.web.http.encodedCharacters.allowEncodedNullCharacter=true"
      - "--entryPoints.web.http.encodedCharacters.allowEncodedSemicolon=true"
      - "--entryPoints.web.http.encodedCharacters.allowEncodedPercent=true"
      - "--entryPoints.web.http.encodedCharacters.allowEncodedQuestionMark=true"
      - "--entryPoints.web.http.encodedCharacters.allowEncodedHash=true"
      - "--entryPoints.web-secure.address=:443"
      - "--entryPoints.web-secure.http.encodedCharacters.allowEncodedSlash=true"
      - "--entryPoints.web-secure.http.encodedCharacters.allowEncodedBackSlash=true"
      - "--entryPoints.web-secure.http.encodedCharacters.allowEncodedNullCharacter=true"
      - "--entryPoints.web-secure.http.encodedCharacters.allowEncodedSemicolon=true"
      - "--entryPoints.web-secure.http.encodedCharacters.allowEncodedPercent=true"
      - "--entryPoints.web-secure.http.encodedCharacters.allowEncodedQuestionMark=true"
      - "--entryPoints.web-secure.http.encodedCharacters.allowEncodedHash=true"
```

florian-forestier:
Hi all,

Just to inform you that this change breaks the [GitLab namespaced API path style](https://docs.gitlab.com/api/rest/#namespaced-paths), and will definitely disrupt a lot of running services. It took me several hours to understand what's happening (and a few minutes more to find out how to remove this restriction as it is only indicated in migration guide, not directly in documentation).

I will not comment the change itself, because it's probably a good change for security and stability, but please clearly indicate when making breaking changes, for ops's sanity :smile: 

Have a good day (and still, thanks a lot for this project and product!)

francescor:
can somebody confirm please that (as @lazka [wrote above](https://github.com/traefik/traefik/pull/12360#issuecomment-3677566048)) that by setting to `true` all these 7 for each entrypoint 

```
encodedCharacters.allowEncodedSlash
encodedCharacters.allowEncodedBackSlash
encodedCharacters.allowEncodedNullCharacter
encodedCharacters.allowEncodedSemicolon
encodedCharacters.allowEncodedPercent
encodedCharacters.allowEncodedQuestionMark
encodedCharacters.allowEncodedHash
```

we can **undo** the effect of this "bug fix" ? 


I also see this one https://doc.traefik.io/traefik/reference/install-configuration/entrypoints/#opt-http-encodedCharacters
so, can we just replace the 7 above by setting to ~true~ something only this one? 

```
encodedCharacters.allowEncoded
```

emilevauge:
Hi everyone 👋

We've been carefully listening to your feedback on PR #12360 and issue #12465 regarding the CVE fix for [path traversal vulnerabilities](https://github.com/advisories/GHSA-gm3x-23wp-hc2c). We truly appreciate you taking the time to share your concerns about the breaking changes this introduced.

**Current Situation**
In our initial fix, we took a strict security-first approach by blocking potentially dangerous encoded characters by default. While this addressed the security vulnerability, we understand it's causing disruption for legitimate use cases in production environments.

**Our Proposal**
After reviewing your feedback and analyzing how other proxies handle this (particularly Envoy), we're proposing to change the default values from false to true for all encoded character options. This would make the security features opt-in rather than opt-out.
The configuration would look like this:
```yaml
entryPoints:
  websecure:
    address: ":443"
    http:
      encodedCharacters:
        allowEncodedSlash: true          # %2F - Default: true
        allowEncodedBackSlash: true      # %5C - Default: true
        allowEncodedNullCharacter: true  # %00 - Default: true
        allowEncodedSemicolon: true      # %3B - Default: true
        allowEncodedPercent: true        # %25 - Default: true
        allowEncodedQuestionMark: true   # %3F - Default: true
        allowEncodedHash: true           # %23 - Default: true
```

**Why This Change?**

1. Community feedback matters: Your deployments and use cases are real, and we don't want to break working setups
2. Flexibility: Users who need stricter security can explicitly opt-out by setting these to false
3. Backward compatibility: Existing deployments would continue working without configuration changes

**Security Considerations**
We want to be transparent: this means the vulnerable behavior would be allowed by default. However:
- Users can still harden their configurations by explicitly disabling these options
- We'll update our documentation with clear security recommendations
- The CVE mitigation path remains available for those who need it

**We Need Your Feedback**
Before we proceed, we'd love to hear your thoughts:

- Does this approach address your concerns?
- Are there specific scenarios we should consider?
- Any other proposal?

Feel free to use emoji reactions to quickly signal your support or concerns - we're watching! 👍 👎
Thank you for helping us find the right balance between security and usability. This is exactly the kind of community collaboration that makes Traefik better! 🧀

**Related:**
Original PR: #12360
Corresponding CVE: [CVE-2025-66490](https://github.com/advisories/GHSA-gm3x-23wp-hc2c)
Documentation: https://doc.traefik.io/traefik/security/request-path/

cleboo:
Hi!

Thanks for taking a second look at this. 

Coming from [this issue](https://github.com/traefik/traefik/issues/12399#issuecomment-3719531666) I still think that being able to configure this behavior per service (or even router?) would make the mitigation much more useful in environments where traefik serves multiple applications that may have different requirements.

If such an option is implemented, having the above settings default to `false` as a secure-by-default option would work for me as well.

xaabi6:
Thank you for taking the community's opinion into account, it's appreciated.

In my opinion, I still find better allowing this configuration per router or service rather than per entryPoint.

grasshide:
Sorry, but imho that is a bad solution. This makes all deployments vulnerable to the CVE. The examples of the community have shown that many of those vulnerable systems are out there and actively used. I'd rather have security by default than the "works out of the box .. with a gaping hole" solution.

ChristianCiach:
I disagree with @grasshide. Traefik is not a firewall. 

Going by that logic traefik could just as well disable access to all `/admin` endpoints by default, with the reason that this path is often exposed by accident. This is obviously absurd, but so is (in my opinion) blocking certain special characters by default.

The more I think about this, the more I believe that this kind of protection should be implemented as a middleware.

Mindar:
I have a slight preference for setting the default to true, but if the default is `true`, then you might as well just remove the option entirely.

Recently I ran into issues with the current defaults of `false` and only as a result of that even learned about this configuration option at all. Most people probably won't even know about the option if the default is `true`. Those who need it, will probably find other ways and protect their applicatons directly at the application layer and not at the proxy layer.

hu3bi:
I really do not like the change. It is better to be secure by default.

Many new users of traefik will encounter an even higher learning curve when setting it up or will fail to harden security.

I rather have something not work but at least it is secure and weaken security if needed or if possible improve/fix issues inside my application code.

lastzero:
@hu3bi see [my comment above](https://github.com/traefik/traefik/pull/12360#issuecomment-3633929685) for a summary of CVE-2025-66490 and the changes Traefik implemented in response:
- https://github.com/traefik/traefik/pull/12360#issuecomment-3633929685

Unfortunately, it is not visible by default due to the large number of comments on this issue. Alternatively, you can click the "Load more" button to view the complete discussion. In short, you are most likely **not** affected, especially if you are a new user who doesn't use split routing (multiple routing rules per application) in combination with insecure (legacy) software that relies solely on Traefik for authentication.

grasshide:
@ChristianCiach’s comment kept me thinking. You’re right: Traefik is not a firewall; it’s a reverse proxy that routes traffic. By the time traffic reaches Traefik, it has already passed the network firewall, which typically cannot differentiate requests by path.

At that point, security decisions are effectively in Traefik’s hands. If a service exposes endpoints that require different levels of protection (for example `/api` vs `/admin`), and those protections are enforced via Traefik middlewares, then Traefik must be able to reliably apply them.

If this configuration can be bypassed, it becomes a security issue, because it breaks the assumption that path-based access control defined in Traefik is authoritative and enforceable.

CybotTM:
> At that point, security decisions are effectively in Traefik’s hands. If a service exposes endpoints that require different levels of protection (for example /api vs /admin), and those protections are enforced via Traefik middlewares, then Traefik must be able to reliably apply them.

Yes, but by fixing the routing, not brute-deny legit URL parts.

As long as Traefik:

- matches router rules against path view A,
- but later forwards a differently normalized or decoded path view B,

…the system is inherently unsafe.

macno:
> but later forwards a differently normalized or decoded path view B

traefik does not do that. chars are NOT decoded

CybotTM:
The GHSA for CVE-2025-66490 explicitly states Traefik evaluates router rules before decoding/normalizing, but forwards after decoding to the backend. 

https://github.com/traefik/traefik/security/advisories/GHSA-gm3x-23wp-hc2c

> Traefik evaluates router rules before decoding or normalizing the request path, but forwards the request after decoding to the backend service. As a result, routes meant to block access to sensitive endpoints (such as internal, beta, or admin endpoints) can be trivially bypassed.

And this is the issue, which is NOT addressed by this PR.

Even if Traefik does not explicitly decode in a specific internal step, the vulnerability class is a semantic mismatch between the path representation used for routing/middleware and the effective path representation processed by the backend.

The problem is:
- Traefik determines which router matches using one path representation
- but applies middleware and routing decisions in a way that allows a request with encoded characters to bypass the intended router/middleware combination

In other words: routing intent and security enforcement are not bound to a single, canonical path representation inside Traefik

lastzero:
**I apologize for reposting this, but it seems that users are commenting on the issue without reading it first. This is ineffective and results in a flood of notifications in everyone's mailbox.**

---

Here's a technical summary of CVE-2025-66490 and the changes Traefik implemented in response for anyone who doesn't want to read the original CVE and all the comments:

1. **The advisory overstates what Traefik actually does.**  
   The [CVE text suggests](https://github.com/advisories/GHSA-gm3x-23wp-hc2c) Traefik decodes the request path before forwarding it, causing `/admin%2F` to reach the backend as `/admin/`. In reality, Traefik routes on the *encoded* path and forwards it *as-is*. If the backend sees `/admin/`, it is because the backend itself chooses to decode `%2F` to `/`, not because Traefik modified the path.
2. **The real issue is a "split view" with certain backends, not Traefik alone.**  
   A bypass only exists in a specific setup:
   - Traefik is used as a **security boundary** (e.g., path-based auth or protection on `/admin/`), and  
   - There is another, less-protected router (e.g., `/`) on the same entrypoint, and  
   - The backend **decodes** reserved characters in the path (e.g., treats `/admin%2F` as `/admin/`).
   
   In that environment, a request like `/admin%2F` may skip the protected router in Traefik but still be interpreted as `/admin/` by the backend. If your backend does *not* decode this way, you are not affected by the original issue — only by the new blocking behavior.
3. **Why Traefik added a breaking "HTTP firewall-ish" fix anyway.**  
   The Traefik team chose to treat this as in-scope because many users rely on Traefik as part of their security perimeter (reverse proxy + access control). To prevent split-view bypasses in that threat model, they introduced **encoded-character filtering** at the entrypoint level and block certain encoded characters (like `%2F`) by default. This is explicitly a *security-first* design choice, not a pure HTTP transparency choice.
4. **Why this breaks legitimate, HTTP-compliant applications.**  
   Before the change, applications that legitimately expected to receive `%2F` in the path (and could decide for themselves whether to decode it) worked correctly behind Traefik. However, after the change, these applications no longer function properly:
   - These requests are blocked **before** they reach the app.  
   - The filtering is configured **per entrypoint**, not per router or service.  
   - That means on a given host/port you must choose between:
     - allowing encoded characters (restoring previous behavior and risk profile), or  
     - blocking them (protecting certain backends, but breaking apps that need strict HTTP compliance).
   
   In practice, if you have one app that requires full RFC-compliant handling of encoded paths and another that relies on Traefik as a firewall to block suspicious encodings, you cannot satisfy both on the same entrypoint without re-architecting. For many setups, this feels like an effective denial of service imposed by the mitigation, not by the original "vulnerability".
5. **Bottom line for Traefik users.**  
   - The CVE’s severity is based on a threat model where Traefik is a security gateway for path-based access control in front of backends that decode reserved characters.  
   - For deployments that use Traefik primarily as a transparent reverse proxy or load balancer, and where backends do not perform risky decoding, the "vulnerability" is largely theoretical. The main impact is the **breaking change** from the mitigation.  
   - If your apps legitimately use encoded characters in paths, you will likely need to:
     - re-enable them explicitly on the affected entrypoint(s), and/or  
     - separate "firewalled" and "HTTP-compliant" apps onto different entrypoints or layers.

In other words: Traefik behaved correctly and transparently according to HTTP before this CVE. The fix is a security-hardening measure for a specific class of backend behaviors and threat models, but it comes at the cost of compatibility for existing deployments that relied on strict HTTP compliance.

IMHO, the **6.9/10 score** for [CVE-2025-66490](https://github.com/advisories/GHSA-gm3x-23wp-hc2c) is too high because “Attack Requirements: None" is incorrect, as is the original claim that Traefik forwards requests after decoding them.

CybotTM:
If you're referring to me, I read it. Did you read my post? The actual problem has absolutely nothing to do with the backend! It's solely due to Traefik.

To be clear: I do not doubt that the CVE is wrong or ambiguous in it's description. But the issue is real.

It seems to me you are focusing on the CVE description/explanation, but not on the real issue behind it.

This issue is a CWE-436 https://cwe.mitre.org/data/definitions/436.html

hu3bi:
> @hu3bi see [my comment above](https://github.com/traefik/traefik/pull/12360#issuecomment-3633929685) for a summary of [CVE-2025-66490](https://github.com/advisories/GHSA-gm3x-23wp-hc2c) and the changes Traefik implemented in response:
> 
>     * [Reject suspicious encoded characters #12360 (comment)](https://github.com/traefik/traefik/pull/12360#issuecomment-3633929685)
> 
> 
> Unfortunately, it is not visible by default due to the large number of comments on this issue. Alternatively, you can click the "Load more" button to view the complete discussion. In short, you are most likely **not** affected, especially if you are a new user who doesn't use split routing (multiple routing rules per application) in combination with insecure (legacy) software that relies solely on Traefik for authentication.


I have to clarify my response.

The problem is, that we are talking about 2 topics in this PR conversation.

Topic 1:  changes which shipped in [(3.6.4)](https://community.traefik.io/t/new-security-updates-for-traefik-2-11-2-11-32-and-3-6-3-6-4/29425)

Topic 2:  https://github.com/traefik/traefik/pull/12360#issuecomment-3719385044 make changes to those changes.


I am not the biggest fan of reverting some of those changes (topic 2). However, I can understand why that decision was made.

CybotTM:
I created #12524 to better explain my point of view.

emilevauge:
Hey there,
Thanks a lot for your feedback. The discussions were highly valuable and helped us reach a decision.
Based on your input, here’s the direction we’re taking:

- **v2.11.35 - v3.6.7**: We will revert Traefik’s behavior on this topic. By default, Traefik will no longer block these characters. You’ll still be able to customize the behavior using the entryPoint options introduced in v3.6.3; their default values will be reversed, as proposed in my latest comment.
- **v3.7.0**: We will introduce a new middleware that allows you to configure this behavior at the router level, or (spoiler alert) at the service level, which is a new feature we have already planned for this release.
- **v4**: Discussions are still open about whether we should take advantage of a major version to change the default behavior. 

Once again, thank you to the community for sharing your concerns, thoughts, and ideas. This was not an easy topic, and there is no perfect solution—but your feedback is essential in helping us make informed decisions.
Thanks for helping make Traefik better.

jappi00:
Spend about 16hrs on this, because of this deny of encoding (and no explicit logs). Did build the images in the wrong time frame I guess...

I still love traefik and youre contribution to the web ecosystem but that was pure hell.

PR #12370, opened 2025-12-05
Add auth-signin to unsupported nginx annotations list

fibsifan:
### What does this PR do?

Add an entry for `auth-signin` in the NGINX unsupported annotations documentation.

### Motivation

I suspect, that the annotation is not supported. I drew that conclusion from reading the note at `auth-url` and the fact that the annotation didn't occur on the entire site.

### More

- [ ] Added/updated tests


rtribotte (approved):
Thanks!

PR #12374, opened 2025-12-05
Add global option to disable X-Forwarded-For appending

### LINKED ISSUE #12001: Missing option to disable traefik's reverse proxy always appending the RemoteAddr to `x-forwarded-for` header
### Welcome!

Currently trying to migrate from ingress nginx to traefik.

Our setup is this:
Client > AWS Cloudfront > AWS NLB > Traefik > Pod

ingress nginx has a config option called [`ComputeFullForwardedFor`](https://github.com/kubernetes/ingress-nginx/blob/6b95221a22d889e38f51ade1a3c6feacf81fed49/internal/ingress/controller/config/config.go#L580-L582) by default in ingress nginx it is set to `false` but if you enable it behaves how traefik currently behaves by always appending the `RemoteAddr` to `x-forwarded-for` header.

However all our application are coded only read 1 IP from `x-forwarded-for` header and since it is read from right to left it happens to read the cloud provider load balancer ip because traefik appended it and I would like to avoid changing all our applications.

I would like traefik to have an option to not append `RemoteAddr` so either Middleware can calculate the IP or trust the cloud provider either via it's `x-forwarded-for` or proxy protocol.

I created [this plugin](https://github.com/jetersen/traefik-cloudfront-xforwarded/) to change the `x-forwarded-for` and read it from `Cloudfront-Viewer-Address`.

Basically I am looking for preserving the x-forwarded-for either from loadbalancer or middleware [without the traefik reverse proxy middleware appending RemoteAddr](https://github.com/traefik/traefik/blob/c20802b07ea8efd6638e0d126eec5df6d932f58f/pkg/proxy/fast/proxy.go#L215-L228).

https://github.com/traefik/traefik/blob/c20802b07ea8efd6638e0d126eec5df6d932f58f/pkg/proxy/fast/proxy.go#L215-L228

Considering NGINX Ingress Provider was added this makes sense to support this behavior to allow for smoother migration.

lbenguigui:
### What does this PR do?

  This PR adds a new `notAppendXForwardedFor` configuration option that prevents Traefik from automatically appending the client's `RemoteAddr` to the `X-Forwarded-For` header when forwarding requests  to backend services.

### Motivation

 Fixes #12001

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/proxy/fast/proxy_test.go
+++ b/pkg/proxy/fast/proxy_test.go
@@ -463,7 +463,7 @@ func TestXForwardedFor(t *testing.T) {
 
 			proxyHandler, err := builder.Build("", testhelpers.MustParseURL(server.URL), true, false)
 			require.NoError(t, err)
-			
+
 			ctx := t.Context()
 			if test.notAppendXFF {
 				ctx = proxyhttputil.SetNotAppendXFF(ctx)


traefiker:

naimadswdn:
Any plans to release this feature? I see that the newest release 3.6.7 do not contain it yet and I wonder when we can expect it.
CC @juliens @mmatur

PR #12396, opened 2025-12-09
Updated limitations of file provider

BaseCrusher:
### What does this PR do?

I just added an additional limitation to how file watch works. I was struggeling the last 3 days to find why it did not work. I think this is something that should be officially documented.

### Motivation

Help other to not struggle as much as i did :D

### More

- [ ] Added/updated tests


rtribotte:
Hey @BaseCrusher,

Thanks for this contribution!

We've set this to "review" and will take a look at the change.

PR #12400, opened 2025-12-10
Add Traefik ingress class annotation to IngressRoute

georgemesss:
Traefik will not pick up the ingressRoute if not annotated properly

### What does this PR do?
Small documantion fix :: Demo Deployment (of _traefik/whoami_) will need an **ingressRoute with a traefik ingress class annotation**, otherwise Traefik will not pick up the ingressRoute and the user may be confused on the root cause.

### Motivation
Was confused by current docs, unable to deploy a demo poc deployment with traefik.

### DIFF
--- a/docs/content/getting-started/kubernetes.md
+++ b/docs/content/getting-started/kubernetes.md
@@ -190,6 +190,8 @@ apiVersion: traefik.io/v1alpha1
 kind: IngressRoute
 metadata:
   name: whoami
+  annotations:
+    kubernetes.io/ingress.class: traefik
 spec:
   entryPoints:
     - web


nmengin:
Hey @georgemesssn

Reviewing it, I'm wondering if we should add a note explaining this annotation usage in the provider reference documentation (for example, in the [particularities](https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-crd/#particularities) section).

If you agree, would you mind writing such a documentation?

Thanks in advance.

nmengin:
Hey @georgemesss,

With [`ingressClassName`](https://doc.traefik.io/traefik/v3.7/reference/routing-configuration/kubernetes/crd/http/ingressroute/#opt-ingressClassName) now supported as an option in `IngressRoute`, I do not feel this modification as necessary.

I close this issue accordingly.
Please add a comment if you disagree, we'll re-open the PR is needed.

PR #12408, opened 2025-12-11
feat(tcp): allow configuring server address with label providers

### LINKED ISSUE #12386: Allow setting `tcp.services.<service-name>.loadbalancer.servers[0].address` via labels
### Welcome!

This is the only setting I have a file provider for, and I'd like to use labels for that.
Similar to other options here:
https://doc.traefik.io/traefik/reference/routing-configuration/other-providers/docker/#tcp-services
ie `traefik.tcp.services.<service_name>.loadbalancer.server.tls`

divyanshu-vashu:
### What does this PR do?

This PR enables the configuration of TCP server addresses via labels, specifically for the Docker provider.

Previously, the Docker provider would always auto-detect the container's IP and port, ignoring any manual address configuration attempted via labels. This change set the TCP behavior with the existing HTTP behavior (where server.url can be manually set).

### More

Fixes #12386

- [x] Added/updated tests

### DIFF
--- a/pkg/config/dynamic/tcp_config.go
+++ b/pkg/config/dynamic/tcp_config.go
@@ -121,7 +121,7 @@ func (l *TCPServersLoadBalancer) Mergeable(loadBalancer *TCPServersLoadBalancer)
 
 // TCPServer holds a TCP Server configuration.
 type TCPServer struct {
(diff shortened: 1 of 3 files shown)


rtribotte:
Hey @divyanshu-vashu,

Thanks for this contribution!

PR #12414, opened 2025-12-12
Support NGINX custom-headers annotation

### LINKED ISSUE #12371: Nginx migration annotation; custom-headers
### Welcome!

As per [nginx documentation](https://kubernetes.github.io/ingress-nginx/examples/customization/custom-headers/) one can use a configmap to add custom headers to responses from a configmap. 

To allow a smooth migration from nginx ingress, this could be supported. 

Goal of the annotation is to provide headers in a configmap, and then let the ingress controller add them to the responses. This is usefull for setting security headers for example.

Eventhough this is (I assume) possible with middleware, it might still be desired for migrating from nginx. For us, not having this would mean we drop these reponse headers and thus become less secure when doing nginx -> traefik migration.

**Desired output**: Adoption of the `"nginx.ingress.kubernetes.io/custom-headers"` annotation, or a very lightweight migration to add custom response headers to an ingress.

nandorKollar:
### What does this PR do?

This PR implements currently unsupported `nginx.ingress.kubernetes.io/custom-headers` annotation of Nginx Kubernetes Ingress provider by automatically configuring a response header middleware. Fixes #12371

### Motivation

Nginx Kubernetes Ingress provider is not supporting the `nginx.ingress.kubernetes.io/custom-headers` annotation. One can add a middleware manually, but supporting the aforementioned annotation could make migration from Nginx Ingress to Traefik smoother.

### More

- [x] Added/updated tests

### Additional Notes

Support for `global-allowed-response-headers` NGINX config is deferred, currently there's no option to whitelist certain headers. This means that every header is by default whitelisted.

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
@@ -804,14 +804,14 @@ func (p *Provider) applyMiddlewares(namespace, routerKey string, ingressConfig i
 
 	applyUpstreamVhost(routerKey, ingressConfig, rt, conf)
 
-	if err := p.applyCustomHeader(routerKey, ingressConfig, rt, conf); err != nil {
+	if err := p.applyCustomHeaders(routerKey, ingressConfig, rt, conf); err != nil {
 		return fmt.Errorf("applying custom headers: %w", err)
 	}
 
 	return nil
 }
 
-func (p *Provider) applyCustomHeader(routerName string, ingressConfig ingressConfig, rt *dynamic.Router, conf *dynamic.Configuration) error {
+func (p *Provider) applyCustomHeaders(routerName string, ingressConfig ingressConfig, rt *dynamic.Router, conf *dynamic.Configuration) error {
 	customHeaders := ptr.Deref(ingressConfig.CustomHeaders, "")
 	if customHeaders == "" {
 		return nil
--- a/docs/content/reference/routing-configuration/kubernetes/ingress-nginx.md
+++ b/docs/content/reference/routing-configuration/kubernetes/ingress-nginx.md
@@ -287,13 +287,13 @@ The following annotations are organized by category for easier navigation.
 
 ### Load Balancing & Backend
 
-| Annotation                                            | Limitations / Notes                                                                        |
-|-------------------------------------------------------|--------------------------------------------------------------------------------------------|
-| <a id="opt-nginx-ingress-kubernetes-ioload-balance" href="#opt-nginx-ingress-kubernetes-ioload-balance" title="#opt-nginx-ingress-kubernetes-ioload-balance">`nginx.ingress.kubernetes.io/load-balance`</a> | Only round_robin supported; ewma and IP hash not supported.                                |
-| <a id="opt-nginx-ingress-kubernetes-iobackend-protocol" href="#opt-nginx-ingress-kubernetes-iobackend-protocol" title="#opt-nginx-ingress-kubernetes-iobackend-protocol">`nginx.ingress.kubernetes.io/backend-protocol`</a> | FCGI and AUTO_HTTP not supported.                                                          |
-| <a id="opt-nginx-ingress-kubernetes-ioservice-upstream" href="#opt-nginx-ingress-kubernetes-ioservice-upstream" title="#opt-nginx-ingress-kubernetes-ioservice-upstream">`nginx.ingress.kubernetes.io/service-upstream`</a> |                                                                                            |
-| <a id="opt-nginx-ingress-kubernetes-ioupstream-vhost" href="#opt-nginx-ingress-kubernetes-ioupstream-vhost" title="#opt-nginx-ingress-kubernetes-ioupstream-vhost">`nginx.ingress.kubernetes.io/upstream-vhost`</a> |                                                      |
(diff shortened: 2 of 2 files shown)


nandorKollar:
Note that there's one limitation, a divergence from Nginx ingress controller: in this implementation one can't use allowed headers (every header is allowed). If this is required, then we can make it behave like in Nginx by reading the value of `global-allowed-response-headers` configuration from nginx configmap.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -800,6 +800,40 @@ func (p *Provider) applyMiddlewares(namespace, routerKey string, ingressConfig i
...
+}
+
+func (p *Provider) applyCustomHeader(routerName string, ingressConfig ingressConfig, rt *dynamic.Router, conf *dynamic.Configuration) error {
+	customHeaders := ptr.Deref(ingressConfig.CustomHeaders, "")
+	if customHeaders == "" {
+		return nil
+	}
+
+	customHeadersParts := strings.Split(customHeaders, "/")
+	if len(customHeadersParts) != 2 {
```
nandorKollar, line 821: Should we allow missing namespace, and default it to the Ingress namespace?

kevinpollet:
Hey @nandorKollar,

Thanks for this contribution!

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
kevinpollet: If I am not missing something, NGINX  does not allow omitting the namespace, so I would keep the same behavior and return an error if the namespace is not specified.

kevinpollet (changes requested):

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
nandorKollar: Thanks, makes sense, I did a quick check, and indeed that's my observation too.

nandorKollar:
> Note that there's one limitation, a divergence from Nginx ingress controller: in this implementation one can't use allowed headers (every header is allowed). If this is required, then we can make it behave like in Nginx by reading the value of `global-allowed-response-headers` configuration from nginx configmap.

@kevinpollet what do you think about this? This looks like an important difference between the current implementation and Nginx. We can either introduce a whitelist annotation (Traefik specific), we can read the nginx configmap and only use this config from there for now (the name and namespace of the configmap can be a template parameter when Traefik is deployed via Helm), or we can also defer this question to a different, subsequent PR and document this limitation in the current one.

kevinpollet:
> @kevinpollet what do you think about this? This looks like an important difference between the current implementation and Nginx. We can either introduce a whitelist annotation (Traefik specific), we can read the nginx configmap and only use this config from there for now (the name and namespace of the configmap can be a template parameter when Traefik is deployed via Helm), or we can also defer this question to a different, subsequent PR and document this limitation in the current one.

Hello @nandorKollar, 

We discussed this during our triage session today and concluded that supporting the 'global-allowed-response-headers' is unnecessary for now. For a first iteration, documenting this limitation is sufficient, and if the community wants this feature, it could be introduced in another pull request.

nandorKollar:
> > @kevinpollet what do you think about this? This looks like an important difference between the current implementation and Nginx. We can either introduce a whitelist annotation (Traefik specific), we can read the nginx configmap and only use this config from there for now (the name and namespace of the configmap can be a template parameter when Traefik is deployed via Helm), or we can also defer this question to a different, subsequent PR and document this limitation in the current one.
> 
> Hello @nandorKollar,
> 
> We discussed this during our triage session today and concluded that supporting the 'global-allowed-response-headers' is unnecessary for now. For a first iteration, documenting this limitation is sufficient, and if the community wants this feature, it could be introduced in another pull request.

Thanks @kevinpollet ! I addressed your review comment and rebased the brach to latest master, would you mind have a second look at this PR?

traefiker:
:no_entry_sign: error related to review: need more review [1/2]

traefiker:
:no_entry_sign: error related to review: need more review [1/2]

traefiker:
:no_entry_sign: error related to review: need more review [1/2]

mwoudstra:
Hi @nandorKollar @mmatur 



We just noticed that nginx does not reload when a `custom-headers` CM is created after the ingress resource.. E.g. when creating the ingress without the CM, it renders a 503 in the config. When adding the CM, it does not reload (and thus fix the 503). 


Can you elaborate what would happen to traefik in this situation? Perhaps this needs some explicit documentation?

nandorKollar:
> Hi @nandorKollar @mmatur
> 
> We just noticed that nginx does not reload when a `custom-headers` CM is created after the ingress resource.. E.g. when creating the ingress without the CM, it renders a 503 in the config. When adding the CM, it does not reload (and thus fix the 503).
> 
> Can you elaborate what would happen to traefik in this situation? Perhaps this needs some explicit documentation?

When you say nginx, is it Nginx ingress controller, or Traefik's ingress provider? If it is nginx, then I'm afraid Traefik will reject the entire ingress, as it looks for the configmap:
```
	configMap, err := p.k8sClient.GetConfigMap(configMapNamespace, configMapName)
	if err != nil {
		return fmt.Errorf("getting configMap %s: %w", customHeaders, err)
	}
```

Does Nginx accept the ingress, when the configmap isn't there yet?

mwoudstra:
@nandorKollar Correct, but then it will render a 503 block, to perserve that the reload does not break the entire config.

nandorKollar:
The other interesting topic is that I'm afraid the middleware isn't modified when the content of the configmap changes. When for example someone removes/adds allowed headers from/to the configmap, then the header middleware keeps it's original value. This latter sound more problematic. We need to watch the changes of this configmap, and reprogram the middleware.

nandorKollar:
> @nandorKollar Correct, but then it will render a 503 block, to perserve that the reload does not break the entire config.

@mwoudstra can you please open an issue for this case?

Issue #12358, opened 2025-12-03
Traefik Does Not Reload Configuration When a Certificate Is Removed from dynamic_conf.yml


Takalele:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

removed cert6 from dynamic_conf.yml.

When removing a certificate entry from dynamic_conf.yml, Traefik does not unload the certificate configuration. The removed certificate remains loaded and continues to be served. Only a full Traefik restart causes the certificate to be properly unloaded.

In contrast, adding a new certificate works as expected: Traefik immediately reloads the configuration, and the newly added certificate becomes active without requiring a restart.

### What did you see instead?

certificate stays loaded.

### What version of Traefik are you using?

```
Version:      3.6.2
Codename:     ramequin
Go version:   go1.24.10
Built:        2025-11-18T16:12:58Z
OS/Arch:      linux/amd64


### What is your environment & configuration?

traefik.yml:
```yaml
log:
  level: DEBUG

api:
  dashboard: True

experimental:

entryPoints:
  old_http:
    address: ":8080"
  http:
    address: ":80"
  https:
    address: ":443"
    http3:
      advertisedPort: 443

providers:
  docker:
    watch: True
    endpoint: "unix:///var/run/docker.sock"
    exposedByDefault: False
    network: proxy
  file:
    directory: "/dynamic"
    watch: True

serversTransport:
  insecureSkipVerify: True

metrics:
  prometheus:
    entryPoint: https
    manualRouting: True
    addEntryPointsLabels: True
    addRoutersLabels: True
    addServicesLabels: True
```

dynamic/dynamic_conf.yml
```yaml
tls:
  stores:
    default:
      defaultCertificate:
        certFile: /etc/letsencrypt/live/cert1/fullchain.pem
        keyFile: /etc/letsencrypt/live/cert1/privkey.pem
  certificates:
    - certFile: /etc/letsencrypt/live/cert2/fullchain.pem
      keyFile: /etc/letsencrypt/live/cert2/privkey.pem
    - certFile: /etc/letsencrypt/live/cert3/fullchain.pem
      keyFile: /etc/letsencrypt/live/cert3/privkey.pem
    - certFile: /etc/letsencrypt/live/cert4/fullchain.pem
[...]

PR #12416, opened 2025-12-13
Remove tls certificate metrics when certs are deleted from config

Takalele:
### What does this PR do?

reset all tls certificate metrics before re-adding current ones during configuration reload to ensure removed certificates don't remain in prometheus metrics.

### Motivation

certificate metrics should reflect the current configuration. when certificates are removed from dynamic_conf.yml, their metrics should be removed as well - just like they are automatically added when new certificates are configured. currently, this only works on full traefik restart, which is not acceptable for production environments.

### Additional Notes

fixes [#12358](https://github.com/traefik/traefik/issues/12358)

### DIFF
--- a/pkg/observability/metrics/prometheus.go
+++ b/pkg/observability/metrics/prometheus.go
@@ -341,7 +341,7 @@ type vector interface {
 }
 
 type prometheusState struct {
-	vectors         []vector
+	vectors          []vector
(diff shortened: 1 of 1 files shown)


nmengin:
Hey @Takalele,
Thanks for this!

Issue #11156, opened 2024-10-03
Docker Image for v3.2.0-rc1 has also been pushed as v3.2


smarrk:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

It appears that the Docker image for version v3.2.0-rc has also been tagged  as v3.2. This may lead to confusion, as it implies the release candidate is a stable version.

### What did you see instead?

The image for v3.2.0-rc has been pushed and is also accessible as v3.2, which might mislead users into thinking they are pulling a stable release instead of a release candidate.
See image checksums for v3.2.0-rc1 and v3.2 at https://hub.docker.com/_/traefik/tags?name=v3.2&ordering=-name

### What version of Traefik are you using?

n/a

### What is your environment & configuration?

n/a

### If applicable, please paste the log output in DEBUG level

n/a

smarrk:
Thank you for the quick response and for outlining your plan to address the issue.

I’m unsure whether this issue should remain open for reference until the fix is implemented or if it should be closed now.
I will leave that decision up to you—feel free to close the issue whenever you think it's appropriate.

PR #12419, opened 2025-12-13
Support new connections to terminating endpoints for zero-downtime

PartlyFluked:
### What does this PR do?

This PR fixes the WRR load balancer to route new connections to terminating Kubernetes endpoints when no other healthy endpoints are available, enabling zero-downtime deployments with single replicas or Recreate strategies.

Associated issue: https://github.com/traefik/traefik/issues/12417

### Motivation

Currently, when all backend endpoints are terminating (marked as `fenced` but still `serving` in Kubernetes EndpointSlice conditions), Traefik returns `503 Service Unavailable` instead of routing traffic to the still-serving terminating endpoints.

This breaks zero-downtime deployments in these scenarios:
- Single replica deployments
- `Recreate` deployment strategy
- Any scenario where all pods are temporarily terminating during rollout

**The Problem:**

1. Pod receives SIGTERM and starts graceful termination
2. Pod's EndpointSlice shows `{ready: false, serving: true, terminating: true}`
3. Pod's preStop hook keeps it serving traffic (e.g., 30 second grace period)
4. **Traefik returns 503 instead of routing to the still-serving pod**
5. New pod created only after old pod fully terminates
6. Users experience downtime during the gap

**Current code issues:**
```go
// Returns error immediately if all handlers are fenced
if len(b.handlers) == 0 || len(b.status) == 0 || len(b.fenced) == len(b.handlers) {
    return nil, errNoAvailableServer
}

// Infinite loop when all handlers are fenced
for {
    if _, ok := b.status[handler.name]; ok {
        if _, ok := b.fenced[handler.name]; !ok {
            break  // Never breaks if all are fenced
        }
    }
}
```

**The Solution:**

Implements a two-tier selection strategy:
1. **Prefer**: Non-terminating healthy endpoints (existing behavior maintained)
2. **Fallback**: Terminating endpoints when no non-terminating available
3. **Error**: Only when no healthy endpoints exist at all

```go
func (b *Balancer) nextServer() (*namedHandler, error) {
    if len(b.handlers) == 0 || len(b.status) == 0 {
        return nil, errNoAvailableServer
    }

    var handler *namedHandler
    var fallbackHandler *namedHandler

    // Prefer non-terminating, track terminating as fallback
    for i := 0; i < len(b.handlers); i++ {
        heap.Push(b, handler)

        if _, ok := b.status[handler.name]; ok {
            if _, isFenced := b.fenced[handler.name]; !isFenced {
                return handler, nil  // Non-terminating found
            } else if fallbackHandler == nil {
                fallbackHandler = handler  // Save first terminating
            }
        }
    }

    // Use terminating fallback if available
    if fallbackHandler != nil {
        return fallbackHandler, nil
    }

    return nil, errNoAvailableServer
}
```

**Benefits:**
- ✅ Zero-downtime deployments with single replicas
- ✅ Respects Kubernetes `serving: true` condition on terminating endpoints
- ✅ Backward compatible (non-terminating endpoints still preferred)
- ✅ Bounded loop prevents infinite iterations (O(n))
- ✅ Aligns with Kubernetes graceful termination best practices

**Manual Testing:**

Tested on k3s with Cilium CNI, Traefik v3.6.2, Nginx with 30s grace period, and Recreate strategy:

Before fix:
```
[16:02:20] Ready:false Serving:true Terminating:true | ClusterIP: 200 | Traefik: 503
[16:02:21] Ready:false Serving:true Terminating:true | ClusterIP: 200 | Traefik: 503
```

After fix:
```
[16:15:30] Ready:false Serving:true Terminating:true | ClusterIP: 200 | Traefik: 200
[16:15:31] Ready:false Serving:true Terminating:true | ClusterIP: 200 | Traefik: 200
```

### More

- [x] Added/updated tests

### Additional Notes

**Test Coverage:**

Added 6 comprehensive test cases covering:
1. `TestNextServerWithAllTerminating` - Routing when all endpoints terminating
2. `TestNextServerPreferNonTerminating` - Non-terminating preference verified
3. `TestNextServerFallbackToTerminating` - Fallback behavior
4. `TestNextServerNoHealthyHandlers` - Proper error handling
5. `TestServeHTTPWithAllTerminating` - Full HTTP request flow
6. `TestServeHTTPGracefulShutdownScenario` - Realistic rolling update scenario

Also updated existing `TestBalancerAllServersFenced` to reflect new fallback behavior.

**Documentation Updates:**

Updated Kubernetes provider documentation:
- `docs/content/routing/providers/kubernetes-ingress.md`
- `docs/content/routing/providers/kubernetes-crd.md`

Added clarification that terminating endpoints serve as fallback for new connections (not just sticky sessions).

**Use Cases:**

Particularly important for:
- Stateless applications with connection draining
- Cost-optimized deployments with minimal replicas
- Development/staging environments with single replicas
- Blue-green deployments with Recreate strategy

**Related:**

- Original terminating endpoints support: #11156 (commit 9588e5114)
- Kubernetes EndpointSlice `serving` condition: https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#serving
- Kubernetes graceful shutdown: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination

This change aligns Traefik with other ingress controllers (e.g., nginx-ingress with `externalTrafficPolicy: Local`) and Kubernetes best practices for graceful termination.

### DIFF
--- a/docs/content/routing/providers/kubernetes-crd.md
+++ b/docs/content/routing/providers/kubernetes-crd.md
@@ -992,8 +992,11 @@ More information in the dedicated [mirroring](../services/index.md#mirroring-ser
 As explained in the section about [Sticky sessions](../../services/#sticky-sessions), for stickiness to work all the way,
 it must be specified at each load-balancing level.
 
-When stickiness is enabled, Traefik uses Kubernetes [serving](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#serving) endpoints status to detect and mark servers as fenced.
-Fenced servers can still process requests tied to sticky cookies, while they are terminating.
+Traefik uses Kubernetes [serving](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#serving) endpoints status to detect and mark servers as fenced during graceful termination.
+
+**With stickiness enabled:** Fenced servers can process requests tied to sticky cookies while they are terminating.
+
+**Graceful termination fallback:** When no healthy non-terminating endpoints are available, Traefik will route new connections to terminating (fenced) endpoints as a fallback. This enables zero-downtime deployments with single replicas or Recreate deployment strategies, where all endpoints may be terminating simultaneously during pod replacement.
 
 For instance, in the example below, there is a first level of load-balancing because there is a (Weighted Round Robin) load-balancing of the two `whoami` services,
 and there is a second level because each whoami service is a `replicaset` and is thus handled as a load-balancer of servers.
--- a/docs/content/routing/providers/kubernetes-ingress.md
+++ b/docs/content/routing/providers/kubernetes-ingress.md
@@ -430,8 +430,11 @@ which in turn will create the resulting routers, services, handlers, etc.
 
 ## Stickiness and load-balancing
 
-When stickiness is enabled, Traefik uses Kubernetes [serving](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#serving) endpoints status to detect and mark servers as fenced.
-Fenced servers can still process requests tied to sticky cookies, while they are terminating.
+Traefik uses Kubernetes [serving](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#serving) endpoints status to detect and mark servers as fenced during graceful termination.
+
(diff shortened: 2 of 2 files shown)


nmengin:
Hey @PartlyFluked,

PR #12459, opened 2025-12-23
Support NGINX buffering annotations

### LINKED ISSUE #12407: ingress-nginx migration to Traefik: support for annotations like proxy-body-size and client-body-buffer-size
### Welcome!

Hi,

Coming because of the impending ingress-nginx deprecation and the hunt for a suitable ingress controller. Got motivated from the reddit posts of Emile to post an issue here. 

So...we have few unsupported annotations. I will create two more issues for them. Let's see two annotations, which I think should be relatively straightforward to fix. 

One is `nginx.ingress.kubernetes.io/client-body-buffer-size` and the other is `nginx.ingress.kubernetes.io/proxy-body-size`. 

We have some use cases where these annotations are used. For example: 

```
ingress:
  annotations:
      nginx.ingress.kubernetes.io/client-body-buffer-size: 1m
      nginx.ingress.kubernetes.io/proxy-body-size: 20m
```

If I am reading https://traefik.io/blog/migrate-from-ingress-nginx-to-traefik-now correctly, having unsupported annotation means I will not be able to seamlessly transition to Traefik. What I might have to do is to create another cluster and then install Traefik and then cut the workload. Is that a reasonable assumption? 

Thanks,

blasko03:
### What does this PR do?

- implemented annotation "nginx.ingress.kubernetes.io/proxy-request-buffering"
- implemented annotation "nginx.ingress.kubernetes.io/client-body-buffer-size"
- implemented annotation "nginx.ingress.kubernetes.io/proxy-body-size"
- implemented annotation "nginx.ingress.kubernetes.io/proxy-buffering"
- implemented annotation "nginx.ingress.kubernetes.io/proxy-buffer-size"
- implemented annotation "nginx.ingress.kubernetes.io/proxy-buffers-number"
- implemented annotation "nginx.ingress.kubernetes.io/proxy-max-temp-file-size"

### More

- [x] Added/updated tests

### Additional Notes

Fixes #12407


pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1007,6 +1011,63 @@ func applyWhitelistSourceRangeConfiguration(routerName string, ingressConfig ing
...
+	bufferingMiddlewareName := routerName + "-buffering"
+	conf.HTTP.Middlewares[bufferingMiddlewareName] = &dynamic.Middleware{
+		Buffering: buffering,
+	}
+	rt.Middlewares = append(rt.Middlewares, bufferingMiddlewareName)
+
+	return nil
+}
+
+func NginxSizeToBytes(nginxSize string) (int64, error) {
```
nandorKollar: nit: better to keep this method package private: `NginxSizeToBytes` -> `nginxSizeToBytes`

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1007,6 +1011,63 @@ func applyWhitelistSourceRangeConfiguration(routerName string, ingressConfig ing
...
+	// calculating memory bytes as defined in https://nginx.org/en/docs/syntax.html
+
+	units := map[string]int64{
+		"g": 1024 * 1024 * 1024,
+		"m": 1024 * 1024,
+		"k": 1024,
+		"":  1,
+	}
+
+	r := regexp.MustCompile(`^(?i)\s*([0-9]+)\s*([k|m|g]?)\s*$`)
```
nandorKollar: nit: it appears that `b` is in fact a valid suffix for bytes (?): https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/annotations/parser/validators.go#L57

nmengin:
Hey @blasko03,

Thanks for this!

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -892,6 +893,347 @@ func TestLoadIngresses(t *testing.T) {
...
+							},
+						},
+					},
+					ServersTransports: map[string]*dynamic.ServersTransport{},
+				},
+				TLS: &dynamic.TLSConfiguration{},
+			},
+		},
+		{
+			desc: "WhitelistSourceRange when proxy body size is invalid",
```
youkoulayley:
```suggestion
			desc: "Buffering when proxy body size is invalid",
```

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -892,6 +893,347 @@ func TestLoadIngresses(t *testing.T) {
...
+							},
+						},
+					},
+					ServersTransports: map[string]*dynamic.ServersTransport{},
+				},
+				TLS: &dynamic.TLSConfiguration{},
+			},
+		},
+		{
+			desc: "WhitelistSourceRange when proxy body size is empty",
```
youkoulayley:
```suggestion
			desc: "Buffering when proxy body size is empty",
```

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -892,6 +893,347 @@ func TestLoadIngresses(t *testing.T) {
...
+							},
+						},
+					},
+					ServersTransports: map[string]*dynamic.ServersTransport{},
+				},
+				TLS: &dynamic.TLSConfiguration{},
+			},
+		},
+		{
+			desc: "WhitelistSourceRange when client body buffer is invalid",
```
youkoulayley:
```suggestion
			desc: "Buffering when client body buffer is invalid",
```

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -892,6 +893,347 @@ func TestLoadIngresses(t *testing.T) {
...
+							},
+						},
+					},
+					ServersTransports: map[string]*dynamic.ServersTransport{},
+				},
+				TLS: &dynamic.TLSConfiguration{},
+			},
+		},
+		{
+			desc: "WhitelistSourceRange when client body buffer size is empty",
```
youkoulayley:
```suggestion
			desc: "Buffering when client body buffer size is empty",
```

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -892,6 +893,347 @@ func TestLoadIngresses(t *testing.T) {
...
+							},
+						},
+					},
+					ServersTransports: map[string]*dynamic.ServersTransport{},
+				},
+				TLS: &dynamic.TLSConfiguration{},
+			},
+		},
+		{
+			desc: "WhitelistSourceRange with proxy body size and client body buffer",
```
youkoulayley:
```suggestion
			desc: "Buffering with proxy body size and client body buffer",
```

rtribotte (dismissed):
Thanks!

rtribotte:
Hi @blasko03,

We already started the review and pushed review commits on the PR, and I even LGTM it, but after more discussions around it with other maintainers, we think we should make the default behavior configurable at the provider configuration level, and make the buffering opt-in.

Is this plan ok for you? We will push other review commits to move in that direction.

nandorKollar:
@rtribotte I guess you wanted to tag @blasko03, I only reviewed this PR. :) Btw LGTM.

rtribotte:
> @rtribotte I guess you wanted to tag @blasko03, I only reviewed this PR. :) Btw LGTM.

Oups, you are right, I edited my comment, thanks!

blasko03:
> Hi @blasko03,
> 
> Thanks again for your contribution!
> 
> We already started the review and pushed review commits on the PR, and I even LGTM it, but after more discussions around it with other maintainers, we think we should make the default behavior configurable at the provider configuration level, and make the buffering opt-in.
> 
> Is this plan ok for you? We will push other review commits to move in that direction.

yes for me is ok

MrMarkW:
It is great that y'all are trying to support more of ingress-nginx annotations, but I'm curious if consumers really need to use buffering with Traefik just because that is how ingress-nginx did things.  

What are the pros/cons of enabling buffering in Traefik?

N214:
> It is great that y'all are trying to support more of ingress-nginx annotations, but I'm curious if consumers really need to use buffering with Traefik just because that is how ingress-nginx did things.
> 
> What are the pros/cons of enabling buffering in Traefik?

What makes you think that consumers does not need those features? Isn't it true that Traefik is trying to be a drop-in replacement of nginx-ingress? If that's the case, supporting nginx annotations is critical. Can't claim to be a drop-in replacement if you're just supporting the most basic use case.

MrMarkW:
> > It is great that y'all are trying to support more of ingress-nginx annotations, but I'm curious if consumers really need to use buffering with Traefik just because that is how ingress-nginx did things.
> > What are the pros/cons of enabling buffering in Traefik?
> 
> What makes you think that consumers does not need those features? Isn't it true that Traefik is trying to be a drop-in replacement of nginx-ingress? If that's the case, supporting nginx annotations is critical. Can't claim to be a drop-in replacement if you're just supporting the most basic use case.

I agree with the need to support as much as possible to make this a drop in replacement.  

I was curious though about the default model of [Traefik acting as a streaming proxy.](https://doc.traefik.io/traefik/security/content-length/) vs ingress-nginx of proxy/buffering everything by default.   Lot of my ingress-nginx consumers have set overrides to proxy/buffer settings to go beyond the ingress-nginx defaults, but if Traefik is streaming by default, they probably don't need to enable buffering unless they need to limit multipart/form-data  or some other use case.  

Just looking to see what others have experienced.

wolszakp:
Any thoughts when it will be available?

rv0lt:
> Any thoughts when it will be available?

I guess it's missing one more review?

pkg/provider/kubernetes/ingress-nginx/annotations.go
```diff
@@ -93,10 +108,8 @@ func parseIngressConfig(ing *netv1.Ingress) (ingressConfig, error) {
 		case reflect.String:
 			cfgValue.Field(i).Set(reflect.ValueOf(&val))
 		case reflect.Bool:
-			parsed, err := strconv.ParseBool(val)
```
nandorKollar, line 97: Could you please explain this change? Is every boolean parameter now parsed as true?

mmatur: `val == "true"` is a comparison expression that evaluates to true if the string val is exactly "true", and false for anything else (including "True", "TRUE", "1", etc.).

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1188,6 +1215,83 @@ func applyWhitelistSourceRangeConfiguration(routerName string, ingressConfig ing
...
+			}
+
+			buffering.MemResponseBodyBytes = bufferSize * int64(ptr.Deref(ingressConfig.ProxyBuffersNumber, p.ProxyBuffersNumber))
+		}
+
+		proxyMaxTempFileSize := defaultProxyMaxTempFileSize
+		if ingressConfig.ProxyMaxTempFileSize != nil {
+			var err error
+			if proxyMaxTempFileSize, err = nginxSizeToBytes(*ingressConfig.ProxyMaxTempFileSize); err != nil {
+				return fmt.Errorf("proxy-body-size annotation has invalid value: %w", err)
```
```suggestion
				return fmt.Errorf("proxy-max-temp-file-size annotation has invalid value: %w", err)
```

franklouwers:
Any chance we would see those in a 3.6.x soon?

emilevauge:
Hey here, it should land in 3.7 very soon.

rv0lt:
> Hey here, it should land in 3.7 very soon.

Do you have a specific deadline? We are in march already and support for ingress-nginx is over 
https://github.com/traefik/traefik/issues/12785

PR #12329, opened 2025-11-28
Bump to go1.25

jnoordsij:
### What does this PR do?

This pull request updates the Go version to 1.25.

See https://tip.golang.org/doc/go1.25

### Motivation

Be up-to-date

### More

- [ ] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

This still requires work/decisions on the following issues:
- [x] Bump of `golangci-lint` is required for compatibility, but triggers a lot of new warnings/errors; these should either be baselined/supressed or solved; see also #12501 
- [x] Bump of tooling in `scripts/code-gen.sh` seems to cause generated code changes; these need to be reviewed and tested; see also #12460

PR #12460, opened 2025-12-24
Update code generator

jnoordsij:
### What does this PR do?

Update code-generator version to 0.19.0, including bump of k8s to latest 1.34 version (both restricted to still remain compatible with Go 1.24).

Also replace use of deprecated `NewSimpleClientset` calls with `NewClientset` in some tests.

### Motivation

My attempt at updating to Go 1.25 (see #12329) got stuck in at least two places, one of them being required to bump the code-generator version. To prevent the single PR from growing too large quickly, I've opted to extract these changes into a first PR to proceed with the other parts later.

For now I've targeted ~~`master`~~ `v3.6` as I ran into some more issues on `v2.11`, but maybe that was just me and this can be done there as well. Anyways `v2.11` does not technically require any more Go bumps (given Go 1.24 EOL is past v2 EOL), so working towards Go 1.25 could be targeted at v3/master only.

### More

- [x] Added/updated tests


nmengin:
Hey @jnoordsij,

Thanks for this!

We agree that targeting the branch `v2.11` is not relevant; however, we'd like to have this feature in the branch `v3.6` to be able to bump the Go version when necessary.

mmatur (approved):
LGTM
Thanks @jnoordsij

PR #12468, opened 2025-12-24
Allow service healthcheck body

aek:
### What does this PR do?

This allow to define a body in the healthcheck of a service in order to be able to use it for example to send jsonrpc requests like this one that works for all Odoo versions
```yaml
  services:
    odoo-2abc75a5:
      loadBalancer:
        servers:
        - url: http://localhost:8069
        healthcheck:
          path: "/jsonrpc"
          method: "POST"
          interval: "10s"
          timeout: "3s"
          headers:
            Content-Type: "application/json"
          body: '{"jsonrpc": "2.0","method": "call", "params": {"service": "db", "method": "db_exist", "args": ["your_db_name"]}}'
```

### Motivation

The need to be able to send jsonrpc requests to healthcheck Odoo instances that works for all the Odoo versions available
I have also seen that this was already requested in https://github.com/traefik/traefik/issues/10055

**More**
- [x] Added examples to the docs


nmengin:
Hey @aek,

Thanks for this!

We've set this to "review" and will review the code.

pkg/healthcheck/healthcheck.go
```diff
@@ -259,7 +261,12 @@ func (shc *ServiceHealthChecker) newRequest(ctx context.Context, target *url.URL
 		u.Host = net.JoinHostPort(u.Hostname(), strconv.Itoa(shc.config.Port))
 	}
 
-	req, err := http.NewRequestWithContext(ctx, shc.config.Method, u.String(), http.NoBody)
+	var body io.Reader = http.NoBody
+	if shc.config.Body != "" && shc.config.Method == http.MethodPost {
```
mmatur:
```suggestion
	if shc.config.Body != "" {
```

I think it is better to not limit to POST Method

pkg/config/dynamic/http_config.go
```diff
@@ -415,6 +415,7 @@ type ServerHealthCheck struct {
 	Hostname          string            `json:"hostname,omitempty" toml:"hostname,omitempty" yaml:"hostname,omitempty"`
 	FollowRedirects   *bool             `json:"followRedirects,omitempty" toml:"followRedirects,omitempty" yaml:"followRedirects,omitempty" export:"true"`
 	Headers           map[string]string `json:"headers,omitempty" toml:"headers,omitempty" yaml:"headers,omitempty" export:"true"`
+	Body              string            `json:"body,omitempty" toml:"body,omitempty" yaml:"body,omitempty" export:"true"`
```
mmatur:
```suggestion
	Body              string            `json:"body,omitempty" toml:"body,omitempty" yaml:"body,omitempty"`
```

PR #12478, opened 2025-12-29
fix(acme): Alter tls renewal period

LtHummus:
### What does this PR do?

This PR modifies the logic that determines when certificates should be renewed. Focus has been given on the Let's Encrypt `shortlived` profile as well as the changes coming in the next few years where LE will issue certificates at 45 days (and all issuers will required to issue certificates valid for 47 days at most by March 15, 2029).

### Motivation

Fixes https://github.com/traefik/traefik/issues/12461

### More

- [X] Added/updated tests
- [N/A] Added/updated documentation


LtHummus:
Forgot to base my branch off the appropriate `v3.6` ... recreating and reopening

PR #12485, opened 2025-12-30
Allow server healthchecks to run concurrently

### LINKED ISSUE #9748: Allow server healthchecks to run concurrently
### Welcome!

While reading through the source I was surprised to see that when Traefik performs a healthcheck for servers within a service, it performs them [sequentially](https://github.com/traefik/traefik/blob/46c266661c270d9732bcf1605f5a5488b5504666/pkg/healthcheck/healthcheck.go#L109). I whipped up a [quick test ](https://github.com/mbStavola/traefik-healthcheck-test) and confirmed that this is actually how it behaves. Interestingly, this isn't the case for services which seem to get launched as [separate goroutines](https://github.com/traefik/traefik/blob/56f7515ecd8572c6a4c73f35e54b33414e6efb79/pkg/server/service/service.go#L337).

This means that, in the worst-case scenario, the last server in the service's pool will get healthchecked `frequency + (timeout * poolSize)` seconds apart. In my opinion, this makes the healthcheck frequency config a bit misleading as I would expect that each server is checked `<frequency>` seconds from the last time it was checked.

What I would like to see is an option to change this behavior and run the healthchecks concurrently. This way, each server in the pool is healthchecked roughly every X seconds defined by the `frequency` parameter, regardless of the (mis)behavior of other servers in the pool. The config could be something like this:

```
traefik.http.services.my-service.loadbalancer.healthcheck.concurrent=true
```

extndr:
### What does this PR do?

Enables server healthchecks within a service to run in parallel, significantly reducing latency when monitoring multiple servers.

### Motivation

Speed up health monitoring and ensure service status is updated more accurately under load.

### Additional Notes

Setting `maxConcurrency = 1` restores the previous behavior, preserving compatibility for users who need it.

This also fixes #9748.

### DIFF
--- a/pkg/healthcheck/healthcheck.go
+++ b/pkg/healthcheck/healthcheck.go
@@ -171,8 +171,8 @@ func (shc *ServiceHealthChecker) healthcheck(ctx context.Context, targets chan t
 			}
 
 			if shc.maxConcurrency > 1 {
-				// Use semaphore to limit concurrent health checks.
-				// This prevents overwhelming the network and target servers.
+				// Limit the maximum number of concurrent health checks.
+				// The effective concurrency is bounded by the number of servers.
 				sem := make(chan struct{}, shc.maxConcurrency)
 				var wg sync.WaitGroup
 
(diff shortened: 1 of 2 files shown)


nmengin:
Hey @extndr,

PR #12488, opened 2025-12-30
Fix link description in Traefik Proxy documentation

schaerfo:
### What does this PR do?
A link in the documentation says it goes to Stack Overflow, but actually leads to Hacker News. So I think it should mention Hacker News.

### Motivation
I was confused that I was taken to Hacker News instead of Stack Overflow.

### More

- [ ] Added/updated tests


kevinpollet (approved):
Thanks 👍

traefiker:

PR #12505, opened 2026-01-06
Document negative priority support for routers

understood-the-assignment:
Currently there is no official documentation which either confirms or denies that negative router priorities are supported; the only information about this that can be found online seems to be outdated (reflecting Traefik 1.0).

Negative priority values do seem to actually be supported since Traefik v2.0.  A router with a negative priority will be evaluated after all routers with positive or default priorities, which is useful for defining fallback or catch-all routes, so let's document this explicitly.

### What does this PR do?

Clarify the documentation about negative priorities.

### Motivation

We wanted to add a negative priority route and weren't sure if it would work or not.

### More

- [ ] Added/updated tests


nmengin:
Hey @understood-the-assignment,

Where we agreed to document the negative values, we disagree on the fallback purpose.
Could you modify your PR to remove it?

understood-the-assignment:
> Could you modify your PR to remove it?

Thanks, done.

docs/content/reference/routing-configuration/http/routing/rules-and-priority.md
```diff
@@ -225,6 +225,8 @@ The priority is directly equal to the length of the rule, and so the longest len
 
 A value of `0` for the priority is ignored: `priority: 0` means that the default rules length sorting is used.
 
+Negative priority values are supported. A router with a negative priority will be evaluated after all routers with positive or default priorities.
```
rtribotte:
```suggestion
Negative priority values are supported.
```

docs/content/reference/routing-configuration/tcp/routing/rules-and-priority.md
```diff
@@ -195,6 +195,8 @@ To avoid path overlap, routes are sorted, by default, in descending order using
 The priority is directly equal to the length of the rule, and so the longest length has the highest priority.
 A value of `0` for the priority is ignored: `priority: 0` means that the default rules length sorting is used.
 
+Negative priority values are supported. A router with a negative priority will be evaluated after all routers with positive or default priorities.
```
rtribotte:
```suggestion
Negative priority values are supported.
```

docs/content/routing/routers/index.md
```diff
@@ -442,6 +442,8 @@ The priority is directly equal to the length of the rule, and so the longest len
 
 A value of `0` for the priority is ignored: `priority = 0` means that the default rules length sorting is used.
 
+Negative priority values are supported. A router with a negative priority will be evaluated after all routers with positive or default priorities.
```
rtribotte:
```suggestion
Negative priority values are supported.
```

docs/content/routing/routers/index.md
```diff
@@ -1267,6 +1269,8 @@ The priority is directly equal to the length of the rule, and so the longest len
...
```
rtribotte:
```suggestion
Negative priority values are supported.
```

understood-the-assignment:
Done.

> A router that should be evaluated last can be safely configured with the lowest positive value, 1.

If it helps, some additional context: we have had an internal deployment for some time, and wanted to add a catch-all route. The lowest value that was unambiguously documented as supported was 1; however, we discovered that some teams had already deployed services which use this priority value. This meant that we either had to search for and update all deployments (+ the machinery that created them) to use a higher priority, or figure out how to create the router with an ever lower priority than what was documented as supported.

kevinpollet (approved):
Thanks 👍

PR #12512, opened 2026-01-06
Fix health check ping

olamilekan000:
### What does this PR do?

Fixes healthcheck command failure when `http_proxy` environment variable is set in the Traefik container.

Screenshot 2026-01-06 at 20 38 20

### Motivation

Fixes: https://github.com/traefik/traefik/issues/12455

When `http_proxy` is set in the container environment, the `traefik healthcheck --ping` command fails with:
```
Bad healthcheck status: 400 Bad Request

```
This occurs because:
1. The healthcheck constructs a URL using `pingEntryPoint.GetAddress()`, which returns just the port (e.g., `:8080`) when the entrypoint address is configured as `:8080`
2. This results in a malformed URL: `http://:8080/ping` (no hostname)
3. Go's HTTP client automatically uses the `http_proxy` environment variable
4. Since the URL has no hostname, it cannot be matched against `no_proxy` settings
5. The request is sent through the proxy, which returns a 400 Bad Request for the invalid URL

> Note: While the malformed URL (http://:8080/ping) was a contributing factor, the root cause was that healthcheck shouldn't use a proxy.

### More

- [ ] Added/updated tests


nmengin:
Hey @olamilekan000,

Thanks for this!

mmatur (approved):

PR #12521, opened 2026-01-08
Fix condition used for serving and fenced endpoints

LBF38:
### What does this PR do?

This PR updates the condition used for serving and fencing endpoints, as defined by the Kubernetes docs.
The Serving condition allows to track if the Pod can receive traffic. It is set to true when the Pod is Ready or Terminating, and a nil value should be interpreted as "true".

### Motivation

Fixes https://github.com/traefik/traefik/issues/12389

### More

- [x] Added/updated tests

### Additional Notes


traefiker:
:no_entry_sign: failed to update: failed to update the pull request: failed to display git log: exit status 128

traefiker:
:no_entry_sign: failed to update: failed to clone: failed to add remote: exit status 3

traefiker:
:no_entry_sign: failed to update: failed to clone: failed to fetch upstream/v3.6 : exit status 128

der-eismann:
Hey @LBF38, I'm currently unsure if you accidentally introduced a bug or of this behavior is intended. We have a service with `publishNotReadyAddresses: true` that used to be available through the attached ingress via traefik. After upgrading from 3.4 to 3.6.11 this stopped working. The EndpointSlice looks like this:
```
    conditions:
      ready: true
      serving: false
      terminating: false
```

Since `serving` is `false` here, I can't reach the not-yet-ready pod. Also the log messages are a bit confusing because `Skipping service: no endpoints found` isn't really true, there is an Endpoint. And an EndpointSlice with an Endpoint.

nmengin:
Hey @der-eismann,

Could you please open a dedicated issue to allow us to validate the bug?
Thanks in advance.

PR #12526, opened 2026-01-08
Fix OTel k8s attributes detection when using Pod hostNetwork

### LINKED ISSUE #12525: OTel K8sAttributesDetector fails if using `hostNetwork`
### Welcome!

### What did you do?

Enable `hostNetwork: true` and OTel `tracing.otlp.enabled:true` in the Traefik Helm chart

Got error logs when pods start up:

```json
{"level":"error","error":"pods \"my-node-name\" not found","time":"2025-12-11T08:54:54Z","message":"Unable to build K8s resource attributes for Traefik pod"}
```

which is the error coming from here:

https://github.com/traefik/traefik/blob/f7280439e6378221a541910f43a01323d52db048/pkg/types/k8sdetector.go#L57

### What did you see instead?

Issue here is that when using `hostNetwork` the container's hostname is the same as the node, instead of the same name as the pod.

This code here assumes that the hostname is the pod name:

https://github.com/traefik/traefik/blob/f7280439e6378221a541910f43a01323d52db048/pkg/types/k8sdetector.go#L44-L47

I suggest using the `POD_NAME` and `POD_NAMESPACE` environment variables if they exist, as this is a common pattern in Kubernetes ecosystem to get around this issue. For example:

```go
func getPodName() (string, error) {
	if name := os.Getenv("POD_NAME"); name != "" {
		return name, nil
	}
	name, err := os.Hostname()
	if err != nil {
		return "", err
	}
	return name, nil
}

func getPodNamespace() (string, error) {
	if namespace := os.Getenv("POD_NAMESPACE"); namespace != "" {
		return namespace, nil
	}
	namespaceBytes, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
	if err != nil {
		return "", err
	}
	return string(namespaceBytes), nil
}
```

In the Helm chart you would then supply the environment variables like so:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: my-pod
spec:
  containers:
  - name: app
    env:
    - name: POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: POD_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
```

For reference, we need `hostNetwork` because we're running some sidecars that run some fingerprinting tech, similar to tcpdump, and it needs access to the original incoming packets and not some NAT'ed packets from kube-proxy.

Version:      3.6.0
Codename:     ramequin
Go version:   go1.24.10
Built:        2025-11-07T15:22:11Z
OS/Arch:      linux/amd64

Helm values:

```yaml
hostNetwork: true

tracing:
  sampleRate: 0.1

  otlp:
    enabled: false
    grpc:
      enabled: true
      endpoint: otel-collector-collector.opentelemetry.svc.cluster.local:4317
      insecure: true
```

_No response_

applejag:
### What does this PR do?

Fixes Traefik's `K8sAttributesDetector` for OpenTelemetry failing when using `hostNetwork: true` on the Traefik Pod by allowing to get Pod name and namespace from environment variables.

### Motivation

It failed when we used `hostNetwork: true`, which we require.

### More

- [ ] Added/updated tests

### Additional Notes

This could be worked around by setting the `OTEL_RESOURCE_ATTRIBUTES` environment variable instead, but that feels more like a hacky workaround.

Closes #12525

### DIFF
--- a/pkg/types/k8sdetector.go
+++ b/pkg/types/k8sdetector.go
@@ -17,7 +17,7 @@ import (
 )
 
 // K8sAttributesDetector detects the metadata of the Traefik pod running in a Kubernetes cluster.
-// It reads the pod name from the hostname file and the namespace from the service account namespace file and queries the Kubernetes API to get the pod's UID.
+// It reads the pod name from the POD_NAME env var (fallback to hostname file) and the namespace from the POD_NAMESPACE env var (fallback to service account namespace file) and queries the Kubernetes API to get the pod's UID.
 type K8sAttributesDetector struct{}
 
 func (K8sAttributesDetector) Detect(ctx context.Context) (*resource.Resource, error) {
@@ -41,16 +41,15 @@ func (K8sAttributesDetector) Detect(ctx context.Context) (*resource.Resource, er
 		return nil, fmt.Errorf("creating Kubernetes client: %w", err)
 	}
 
-	podName, err := os.Hostname()
+	podName, err := getPodName()
 	if err != nil {
 		return nil, fmt.Errorf("getting pod name: %w", err)
 	}
 
-	podNamespaceBytes, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+	podNamespace, err := getPodNamespace()
 	if err != nil {
 		return nil, fmt.Errorf("getting pod namespace: %w", err)
 	}
-	podNamespace := string(podNamespaceBytes)
 
 	pod, err := client.CoreV1().Pods(podNamespace).Get(ctx, podName, metav1.GetOptions{})
 	if err != nil && (kerror.IsForbidden(err) || kerror.IsNotFound(err)) {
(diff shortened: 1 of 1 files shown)


nmengin:
Hey @applejag,

Thanks for the contribution!

applejag:
That's completely fine.

In the meantime if anyone has this issue then here's a workaround:

```yaml
# traefik Helm chart values

env:
  # Hack around OTel k8sdetector bug in Traefik,
  # which assumes that the hostname is the pod name,
  # but that's not the case in some of our clusters where we use `hostNetwork: true`
  #
  # The POD_NAME & POD_NAMESPACE env vars are already set by the
  # Traefik Helm chart: https://github.com/traefik/traefik-helm-chart/blob/08588b97c8ea6ede92e86dcb750ec4bc4ad94e45/traefik/templates/_podtemplate.tpl#L954-L961
  #
  # See: https://github.com/traefik/traefik/blob/f7280439e6378221a541910f43a01323d52db048/pkg/types/k8sdetector.go#L57
  - name: POD_UID
    valueFrom:
      fieldRef:
        apiVersion: v1
        fieldPath: metadata.uid
  - name: OTEL_RESOURCE_ATTRIBUTES
    value: k8s.pod.name=$(POD_NAME),k8s.pod.uid=$(POD_UID),k8s.namespace.name=$(POD_NAMESPACE)
```

PR #12503, opened 2026-01-05
Replace markdown-include dependency with mkdocs-include-markdown-plugin

bpsoraggi:
### What does this PR do?

Replaces dependency with GPLv3 license

### Motivation

Comply with license obligations

### More

- [ ] Added/updated tests
- [x] Added/updated documentation

Cf. #12532

PR #12532, opened 2026-01-09
Replace markdown-include dependency with mkdocs-include-markdown-plugin

bpsoraggi:
### What does this PR do?

Replaces dependency with GPLv3 license

### Motivation

Comply with license obligations

### More

- [ ] Added/updated tests

Cf. #12503


kevinpollet (approved):
Thanks 👍

PR #12545, opened 2026-01-12
Handle empty/missing User-Agent header

### LINKED ISSUE #12527: ForwardAuth: handle empty/missing `User-Agent` header
### Welcome!

## Motivation

I'm building a simple "WAF" that is implemented as a Forward Auth middleware. One of my use-cases is to block requests with empty/missing headers. Unfortunately this is not possible for the `User-Agent` header because Traefik forwards the default Golang User-Agent in this case (which is simply the default `net.http.Client.Do()` behavior).

My current workaround is to check for empty User-Agents in my router rule and only route clients that provide a non-empty value.

This could also be useful for other Forward Auth services.

(I can prepare a PR if there is interest.)

## Expected behavior

In case the incoming request does not contain a `User-Agent` header, the Forward Auth middleware should not receive the default Golang User-Agent `Go-http-client/1.1` but an empty `User-Agent` value.

(This is already correctly handled in [net/http/httputil/reverseproxy.go](https://github.com/golang/go/blob/28147b528312055b535c6a69d0d4492bd502e1b0/src/net/http/httputil/reverseproxy.go#L533) but not in `pkg/middlewares/auth/forward.go` since it just performs a simple HTTP request. I also found an [ancient PR](https://github.com/traefik/traefik/pull/6030) that addressed a similar issue).

a-stangl:
### What does this PR do?

Adjust `writeHeader()` in `pkg/middlewares/auth/forward.go`:

- set the `User-Agent` header to an empty string in case the header is missing in the original request
- add/adapt unit tests

### Motivation

Closes #12527 

### More

- [x] Added/updated tests

### Additional Notes

I adjusted the unit test to check if the headers specified by `expectedHeaders` actually exist in the request.

### DIFF
--- a/docs/content/assets/css/content-width.css
+++ b/docs/content/assets/css/content-width.css
@@ -1,0 +1,4 @@
+/* Use a wider grid to accommodate table content and code blocks. */
+.md-grid {
+  max-width: 1650px;
+}
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -39,6 +39,7 @@ extra_javascript:
 extra_css:
   - assets/css/menu-icons.css
   - assets/css/code-copy.css
+  - assets/css/content-width.css
 
 plugins:
   - search


nmengin:
Hey @a-stangl,
Thanks for this!

mmatur (approved):
LGTM

Thanks @a-stangl for your contribution

traefiker:

Issue #12511, opened 2026-01-06
Unnecessary status updates to BackendTLSPolicy causing object modified error


steve21168:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I'm using BackendTLSPolicy and everything seems to be working fine but my traefik controller is logging:
> Unable to update BackendTLSPolicy status... the object has been modified; please apply your changes to the latest version and try again

I think this is basically the same as: https://github.com/traefik/traefik/issues/10939 .  I see the transition time is constantly updated even though nothing is changing.

### What did you see instead?

No error/warning

### What version of Traefik are you using?

v3.6.2

### What is your environment & configuration?

```yaml
# (paste your configuration here)
```

Add more configuration information here.

nmengin:
Hey @steve21168,
Thanks for reaching out! Could you share a minimal reproducible case? (For example, full Kubernetes manifests or a simple setup that shows the issue?)
If anyone else can help reproduce or fix this, we'd love the help!

steve21168:
Okay I've figured out how to reproduce.

So I think the error I saw is more of a byproduct of running multiple replicas of Traefik and a race condition updating the backendtlspolicy status.  I mentioned that the status of the backendtlspolicy is being updated frequently (every few seconds) and so that I believe is actually the relevant root cause issue.

What I found in trying to figure out how to reproduce the issue is that we have _multiple_ gateway listeners and the ancestor status (in backendtlspolicy) is being updated with different "Section Names" referencing those listeners. 


To reproduce:
```
helm upgrade --install traefik-crds traefik/traefik-crds --set gatewayAPI=true --version 1.12.0
helm upgrade --install traefik-gateway traefik/traefik --version 37.4.0 -f ./traefik-values.yaml
```
traefik-values.yaml
```
providers:
  kubernetesIngress:
    enabled: false
  kubernetesGateway:
    enabled: true

deployment:
  replicas: 2

ports:
  web:
    port: 80

gateway:
  listeners:
    web:
      protocol: HTTP
      port: 80
      hostname: foo.com
      namespacePolicy:
        from: All
    other:
      protocol: HTTP
      port: 80
      hostname: boo.com
      namespacePolicy:
        from: All

logs:
  general:
    level: DEBUG
  access:
    enabled: true
```
[...]

PR #12549, opened 2026-01-13
Remove listener name in backendtls ancestor refs

steve21168:
### What does this PR do?

Prevents a BackendTLSPolicy.status.ancestors.ancestorRef.sectionName from being overwritten by every listener.

### Motivation

There is a bug that will cause the controller to repeatedly update BackendTLSPolicy for each gateway listener.

See: https://github.com/traefik/traefik/issues/12511

### More

- [ ] Added/updated tests

### Additional Notes

Fwiw I'm not sure this is the best way to handle this but the right way may require a large rewrite of the original implementation.  My thought process is that the listener isn't specific to a backendtlspolicy.  A backendtlspolicy typically targets a kubernetes Service and tells a gateway controller when sending to that service use TLS for the connection.

In the current code implementation all listeners are getting looped through and added as the ancestor to the policy regardless of if that policy has any association whatsoever with the listener.  It makes sense to me that the Gateway itself would be the ancestor rather than a specific listener as everything passing through the gateway will already adhere to the backendtlspolicy rule.

### DIFF
--- a/pkg/provider/kubernetes/gateway/httproute.go
+++ b/pkg/provider/kubernetes/gateway/httproute.go
@@ -445,7 +445,6 @@ func (p *Provider) loadHTTPServers(ctx context.Context, namespace string, route 
 					Kind:        ptr.To(gatev1.Kind(kindGateway)),
 					Namespace:   ptr.To(gatev1.Namespace(namespace)),
 					Name:        gatev1.ObjectName(listener.GWName),
-					SectionName: ptr.To(gatev1.SectionName(listener.Name)),
 				},
 				ControllerName: controllerName,
 			}


nmengin:
Hey @steve21168,

rtribotte:
Hi @steve21168,

We successfully reproduced the problem and identified the issue. In doing so, we determined that your change is not the correct way to handle the issue.

We are closing this PR accordingly and will work on a fix.

oliverrahner:
Would be great to know where this issue is tracked now that this PR got closed. I would assume https://github.com/traefik/traefik/issues/12619?

nmengin:
Hey @oliverrahner,

We will use #12511.

PR #12562, opened 2026-01-16
Add routing configuration extension points

jspdown:
### What does this PR do?

This PR adds the ability to extend dynamic configuration structures with custom fields.

To achieve this, it introduces a separate Go module (pkg/config/dynamic/ext) containing empty placeholder structs. These structs are embedded into configuration types like HTTPConfiguration and Router. To extend these types, you can replace the module with your own implementation using a Go module replace directive.

Adding extensibility to the configuration types also required fixing how configurations get merged. The existing merge code was scattered across providers with a lot of duplication, and it could not handle custom fields. The provider.Merge function has been rewritten to use reflection to walk configuration structs, which allows it to support embedded fields.

A merge strategy concept has been introduced to handle different provider needs: one strategy merges compatible services (used by the aggregator), and another skips duplicates (used by the file provider).

### Motivation

Traefik Hub needs to attach additional metadata to routers and other resources.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/merge.go
+++ b/pkg/provider/merge.go
@@ -88,29 +88,29 @@ func Merge(ctx context.Context, configurations []NamedConfiguration, strategy Re
 
 	for _, c := range configurations {
 		if c.Configuration.HTTP != nil {
-			mergeResourceMaps(ctx, reflect.ValueOf(merged.HTTP).Elem(), reflect.ValueOf(c.Configuration.HTTP).Elem(), c.Name, tracker, strategy, resourceLogFields)
+			mergeResourceMaps(ctx, reflect.ValueOf(merged.HTTP).Elem(), reflect.ValueOf(c.Configuration.HTTP).Elem(), c.Name, tracker, strategy)
 		}
 		if c.Configuration.TCP != nil {
-			mergeResourceMaps(ctx, reflect.ValueOf(merged.TCP).Elem(), reflect.ValueOf(c.Configuration.TCP).Elem(), c.Name, tracker, strategy, resourceLogFields)
+			mergeResourceMaps(ctx, reflect.ValueOf(merged.TCP).Elem(), reflect.ValueOf(c.Configuration.TCP).Elem(), c.Name, tracker, strategy)
 		}
 		if c.Configuration.UDP != nil {
-			mergeResourceMaps(ctx, reflect.ValueOf(merged.UDP).Elem(), reflect.ValueOf(c.Configuration.UDP).Elem(), c.Name, tracker, strategy, resourceLogFields)
+			mergeResourceMaps(ctx, reflect.ValueOf(merged.UDP).Elem(), reflect.ValueOf(c.Configuration.UDP).Elem(), c.Name, tracker, strategy)
 		}
 		if c.Configuration.TLS != nil {
-			mergeResourceMaps(ctx, reflect.ValueOf(merged.TLS).Elem(), reflect.ValueOf(c.Configuration.TLS).Elem(), c.Name, tracker, strategy, resourceLogFields)
+			mergeResourceMaps(ctx, reflect.ValueOf(merged.TLS).Elem(), reflect.ValueOf(c.Configuration.TLS).Elem(), c.Name, tracker, strategy)
 
 			merged.TLS.Certificates = mergeCertificates(ctx, merged.TLS.Certificates, c.Configuration.TLS.Certificates, c.Name, strategy)
 		}
 	}
 
-	deleteConflicts(ctx, tracker, resourceLogFields)
+	deleteConflicts(ctx, tracker)
 
 	return merged
 }
 
 // mergeResourceMaps merges all the resource maps defined in the provided struct.
 // Conflicts are recorded in the given merge tracker.
-func mergeResourceMaps(ctx context.Context, dst, src reflect.Value, origin string, tracker *mergeTracker, strategy ResourceStrategy, resourceLogFields map[reflect.Type]resourceMeta) {
+func mergeResourceMaps(ctx context.Context, dst, src reflect.Value, origin string, tracker *mergeTracker, strategy ResourceStrategy) {
 	dstType := dst.Type()
 
 	for i := range dstType.NumField() {
@@ -124,12 +124,12 @@ func mergeResourceMaps(ctx context.Context, dst, src reflect.Value, origin strin
 
 		// Merge the resource maps of embedded structs.
 		if field.Anonymous {
-			mergeResourceMaps(ctx, dstField, srcField, origin, tracker, strategy, resourceLogFields)
+			mergeResourceMaps(ctx, dstField, srcField, origin, tracker, strategy)
 			continue
 		}
 
 		if dstField.Kind() == reflect.Map {
-			mergeResourceMap(ctx, dstField, srcField, origin, tracker, strategy, resourceLogFields)
+			mergeResourceMap(ctx, dstField, srcField, origin, tracker, strategy)
 		}
 	}
 }
@@ -138,7 +138,7 @@ func mergeResourceMaps(ctx context.Context, dst, src reflect.Value, origin strin
 // New keys from src are added to dst.
 // Duplicate keys are merged if the resource type implements a Merge method, otherwise
 // the values must be identical. Conflicts are recorded in the given merge tracker.
-func mergeResourceMap(ctx context.Context, dst, src reflect.Value, origin string, tracker *mergeTracker, strategy ResourceStrategy, resourceLogFields map[reflect.Type]resourceMeta) {
+func mergeResourceMap(ctx context.Context, dst, src reflect.Value, origin string, tracker *mergeTracker, strategy ResourceStrategy) {
 	if src.IsNil() {
 		return
 	}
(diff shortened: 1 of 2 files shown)


pkg/provider/merge.go
```diff
@@ -0,0 +1,320 @@
...
+			if methodType.In(0).AssignableTo(src.Type()) {
+				results := mergeMethod.Call([]reflect.Value{src})
+				return results[0].Bool()
+			}
+		}
+	}
+
+	// When Merge is not implemented, merge is not allowed; the values must be the same.
+	return reflect.DeepEqual(dstActual.Interface(), srcActual.Interface())
+}
```
lbenguigui, line 203:
```suggestion
  func tryMerge(dst, src reflect.Value) bool {                                                                                                                                                                                           
        if dst.Kind() != reflect.Ptr {                                                                                                                                                                              
                return reflect.DeepEqual(dst.Interface(), src.Interface())                                                                                                                                          
        }                                                                                                                                                                                                           
                                                                                                                                                                                                                               
        if dst.IsNil() || src.IsNil() {                                                                                                                                                                             
                return reflect.DeepEqual(dst.Interface(), src.Interface())                                                                                                                                          
        }                                                                                                                                                                                                           
                                                                                                                                                                                                                    
        	// Check if the struct has the method `func (* T) Merge(other T) bool`.
	// We use reflection to detect this method because Go's type system doesn't allow type assertions
	// on generic interfaces (Mergeable[T]) practically.                                                                                                                                    
        mergeMethod := dst.MethodByName("Merge")                                                                                                                                                                    
        if mergeMethod.IsValid() {                                                                                                                                                                                  
                methodType := mergeMethod.Type()                                                                                                                                                                    
                if methodType.NumIn() == 1 && methodType.NumOut() == 1 && methodType.Out(0).Kind() == reflect.Bool {      
                // Make sure the parameter type matches the type holding the method.                                                                                          
                        if methodType.In(0).AssignableTo(src.Type()) {                                                                                                                                              
                                results := mergeMethod.Call([]reflect.Value{src})                                                                                                                                   
                                return results[0].Bool()                                                                                                                                                            
                        }                                                                                                                                                                                           
                }                                                                                                                                                                                                   
        }                                                                                                                                                                                                           
                                                                                                                                                                                                                    
        // When Merge is not implemented, merge is not allowed; the values must be the same.                                                                                                                                           
        return reflect.DeepEqual(dst.Elem().Interface(), src.Elem().Interface())                                                                                                                                    
  } 
```

pkg/provider/kubernetes/ingress/annotations.go
```diff
@@ -22,6 +23,8 @@ type RouterConfig struct {
 
 // RouterIng is the router's configuration from annotations.
 type RouterIng struct {
+	ext.Router
```
lbenguigui: WDYT about using a dedicating struct to handle `RouterIng`?

pkg/provider/file/file.go
```diff
@@ -221,7 +221,12 @@ func (p *Provider) buildConfiguration() (*dynamic.Configuration, error) {
 	ctx := log.With().Str(logs.ProviderName, providerName).Logger().WithContext(context.Background())
 
 	if len(p.Directory) > 0 {
-		return p.loadFileConfigFromDirectory(ctx, p.Directory, nil)
+		configurations, err := p.collectFileConfigs(ctx, p.Directory, "")
+		if err != nil {
+			return nil, err
```
kevinpollet:
```suggestion
			return nil, fmt.Errorf("collecting file configs: %w")
```

pkg/provider/file/file.go
```diff
@@ -376,47 +381,28 @@ func (p *Provider) loadFileConfig(ctx context.Context, filename string, parseTem
...
-			TLS: &dynamic.TLSConfiguration{
-				Stores:  make(map[string]tls.Store),
-				Options: make(map[string]tls.Options),
-			},
-			UDP: &dynamic.UDPConfiguration{
-				Routers:  make(map[string]*dynamic.UDPRouter),
-				Services: make(map[string]*dynamic.UDPService),
-			},
-		}
+		return nil, fmt.Errorf("unable to read directory %s: %w", directory, err)
```
kevinpollet:
```suggestion
		return nil, fmt.Errorf("reading directory %s: %w", directory, err)
```

pkg/provider/file/file.go
```diff
@@ -376,47 +381,28 @@ func (p *Provider) loadFileConfig(ctx context.Context, filename string, parseTem
...
+		if prefix != "" {
+			filename = filepath.Join(prefix, item.Name())
+		}
 
 		if item.IsDir() {
-			configuration, err = p.loadFileConfigFromDirectory(logger.WithContext(ctx), filepath.Join(directory, item.Name()), configuration)
+			sub, err := p.collectFileConfigs(ctx, itemPath, filename)
 			if err != nil {
-				return configuration, fmt.Errorf("unable to load content configuration from subdirectory %s: %w", item, err)
+				return nil, fmt.Errorf("unable to load content configuration from subdirectory %s: %w", item, err)
```
kevinpollet:
```suggestion
				return nil, fmt.Errorf("loading content configuration from subdirectory %s: %w", item, err)
```

rtribotte: @jspdown @juliens I think we should keep the same behavior as before, return the partially filled configuration instead of nil.

pkg/provider/file/file.go
```diff
@@ -427,132 +413,18 @@ func (p *Provider) loadFileConfigFromDirectory(ctx context.Context, directory st
...
-			}
-		}
-
-		for name, conf := range c.TCP.Middlewares {
-			if _, exists := configuration.TCP.Middlewares[name]; exists {
-				logger.Warn().Str(logs.MiddlewareName, name).Msg("TCP middleware already configured, skipping")
-			} else {
-				configuration.TCP.Middlewares[name] = conf
-			}
+			return nil, fmt.Errorf("%s: %w", itemPath, err)
```
rtribotte, line 418: @jspdown @juliens (Same as above) I think we should keep the same behavior as before, return the partially filled configuration instead of nil.

pkg/provider/file/file.go
jspdown: Unless I missed something, `collectFileConfigs` is called by `buildConfiguration` which is called by `applyConfiguration`. Both return the error immediately and ignore the returned configuration.

pkg/provider/merge.go
```diff
@@ -0,0 +1,325 @@
...
+			}
+		}
+	}
+
+	// When Merge is not implemented, merge is not allowed; the values must be the same.
+	return reflect.DeepEqual(dst.Elem().Interface(), src.Elem().Interface())
+}
+
+// deleteConflicts removes conflicting items and logs errors.
+func deleteConflicts(ctx context.Context, tracker *mergeTracker, resourceLogFields map[reflect.Type]string) {
```
rtribotte: @jspdown @juliens Why resourceLogFields is passed down as a parameter, instead of using the package variable direclty?

jspdown: `deleteConflicts` like `mergeResourceMaps` are not specific to the `dynamic.Configuration` struct, they could work with whatever structure we pass in. When implementing it sounded natural not to hardcode dynamic config mapping in them.

pkg/provider/merge.go
```diff
@@ -0,0 +1,325 @@
...
+	"github.com/huandu/xstrings"
+	"github.com/rs/zerolog/log"
+	"github.com/traefik/traefik/v3/pkg/config/dynamic"
+	"github.com/traefik/traefik/v3/pkg/observability/logs"
+	"github.com/traefik/traefik/v3/pkg/tls"
+)
+
+var resourceLogFields = map[reflect.Type]string{
+	reflect.TypeFor[dynamic.Router]():              logs.RouterName,
+	reflect.TypeFor[dynamic.Service]():             logs.ServiceName,
```
rtribotte: @jspdown @juliens For HTTP resources, we'll miss the HTTP prefix in logs. I think we should keep it. Couldn't we simply store the name with the correct case, and get rid of xstrings.FirstRuneToUpper ?

jspdown: Yes sure 👍 For the logs I wasn't really sure if the contract needed to be maintained. I can repurpose this map to store both the resource name and the log field name. Would that work for you?

rtribotte: Yes I think we should, just to make the log message clearer.
> Would that work for you?

Absolutely 👍

pkg/provider/merge.go
rtribotte: Mmmh maybe I overlook something here, but it looks like the `resourceLogFields` maps is passed down from `Merge` to `deleteConflicts` and `mergeResourceMaps`, and is the `resourceLogFields` package variable.

pkg/provider/file/file.go
rtribotte: Indeed, I didn't see that. All good then.

pkg/provider/file/file.go
```diff
@@ -221,7 +221,12 @@ func (p *Provider) buildConfiguration() (*dynamic.Configuration, error) {
 	ctx := log.With().Str(logs.ProviderName, providerName).Logger().WithContext(context.Background())
 
 	if len(p.Directory) > 0 {
-		return p.loadFileConfigFromDirectory(ctx, p.Directory, nil)
+		configurations, err := p.collectFileConfigs(ctx, p.Directory, "")
+		if err != nil {
+			return nil, fmt.Errorf("collecting file configs: %w")
```
kevinpollet:
```suggestion
			return nil, fmt.Errorf("collecting file configs: %w", err)
```

pkg/provider/merge.go
```diff
@@ -0,0 +1,330 @@
...
+
+		if !found {
+			certificates = append(certificates, certificate)
+		}
+	}
+
+	return certificates
+}
+
+// mergeStores merges two store slices, deduplicating entries while. Order is preserved.
```
sdelicata, line 249:
```suggestion
// mergeStores merges two store slices, deduplicating entries. Order is preserved.
```

traefiker:

PR #12563, opened 2026-01-17
feat: enhance headers middleware to support regex replacement

### LINKED ISSUE #5809: replaceResponseHeaders middleware feature request

Feature

Would be great  to have the option to replace header the same way as it's done for `replacepathregex`, example : 

````
http:
  middlewares:
    test-replaceheaders:
      replaceResponseHeaders: // new middleware feature
        location:   //header to replace
          regex: "^/foo/(.*)"
          replacement: "/bar/$1"
 
````

jayl1e:
### What does this PR do?

Enhance headers middleware to support regex-based header replacement.

### Motivation

The ability to replace headers is required in various situations (see #5809). Since the headers middleware is already designed to add and delete headers, adding a 'replace' function is a natural progression. Importantly, this implementation remains lightweight and introduces no further dependencies.

### More

- [x] Added/updated tests

### Additional Notes

Closes #5809


kevinpollet:
Hey @jayl1e,

Thanks for this!

jayl1e:
any update?

nmengin:
Hey @jayl1e,

CISSRUPF:
This change does sound like something close to what I am looking for. @jayl1e if the header I'd target with this middleware were not present in the response, would it get added?

My scenario is that some paths on the backends that I target respond with a 30x HTTP status and "Location" header for the redirect and unfortunately those legacy services respond with their own, per-backend hostname instead of the frontend hostname and don't support modern conventions such as using the X-Forwarded-Host header value that I pass them. If I were to use this middleware on the "Location" header, would it only apply the regex replace, if Location were present or add the header also for non-redirecting requests that don't contain that header?

It would simplify my traefik configuration a lot if I could apply this middleware to all requests, at the cost of the overhead it adds, if I would not have to find all possible redirect locations from access logs and create different pathprefix routes for all of them.

jayl1e:
It just replace, not add
There are already other ways to add header,  we should maintain orthogonality here.

CISSRUPF:
Ok, thats perfect and exactly what I would need (replace the header, only if present). I'll see if I can reach someone at Traefik to make them aware of this pre-existing work. Had discussed it with some of their devs at KubeCon as a missing functionality.

PR #12565, opened 2026-01-17
Remove extraneous dots

dathbe:
### What does this PR do?

Removes extraneous `.` from documentation

### Motivation

There was an extraneous `.` when copying

### More

- [ ] Added/updated tests
- [X ] Added/updated documentation

PR #12571, opened 2026-01-19
Remove extraneous dots in migration guide

dathbe:
This is #12565 with the change made against the v3.6 branch.  Removes extraneous `.` from documentation.


rtribotte (approved):
Thanks!

PR #12574, opened 2026-01-20
Remove invalid private key in log

juliens:
### What does this PR do?

This PR removes the invalid private key from the logs. Even though the key is not valid, it may still contain sensitive information and should not be logged.

### Motivation

Improve log security and reduce the risk of leaking sensitive data.


pkg/provider/acme/account.go
```diff
@@ -58,7 +58,7 @@ func (a *Account) GetPrivateKey() crypto.PrivateKey {
 	privateKey, err := x509.ParsePKCS1PrivateKey(a.PrivateKey)
 	if err != nil {
 		log.Error().Str(logs.ProviderName, "acme").
-			Err(err).Msgf("Cannot unmarshal private key %+v", a.PrivateKey)
+			Err(err).Msgf("Cannot unmarshal private key")
```
kevinpollet:
```suggestion
			Err(err).Msg("Cannot unmarshal private key")
```

kevinpollet (approved):

PR #12596, opened 2026-01-26
Failover according to response status code

lbenguigui:
### What does this PR do?

This PR adds the ability to trigger failover based on HTTP response status codes from the main handler, in addition to the existing health check mechanism. 
When configured with an errors block, the failover service will intercept responses matching specified status codes (e.g., 500-504) and automatically route the request to the fallback handler

### More

- [x] Added/updated tests

Co-authored-by: juliens <julien.salleyron@gmail.com>

### DIFF
--- a/pkg/server/service/loadbalancer/failover/failover.go
+++ b/pkg/server/service/loadbalancer/failover/failover.go
@@ -41,7 +41,7 @@ func New(config *dynamic.Failover) (*Failover, error) {
 		if len(config.Errors.Status) > 0 {
 			httpCodeRanges, err := types.NewHTTPCodeRanges(config.Errors.Status)
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("creating HTTP code ranges: %w", err)
 			}
 			f.statusCode = httpCodeRanges
 		}
@@ -85,8 +85,8 @@ func (f *Failover) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		// TODO: move reusable request to a common package at some point.
 		rr, _, err := mirror.NewReusableRequest(req, f.maxRequestBodyBytes)
 		if err != nil && !errors.Is(err, mirror.ErrBodyTooLarge) {
-			http.Error(w, fmt.Sprintf("%s: creating reusable request: %v",
-				http.StatusText(http.StatusInternalServerError), err), http.StatusInternalServerError)
+			log.Ctx(req.Context()).Debug().Err(err).Msg("Error while creating reusable request for failover")
+			http.Error(w, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
 			return
 		}
 
--- a/pkg/server/service/loadbalancer/mirror/mirror.go
+++ b/pkg/server/service/loadbalancer/mirror/mirror.go
@@ -74,8 +74,8 @@ func (m *Mirroring) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		var bytesRead []byte
 		rr, bytesRead, err = NewReusableRequest(req, m.maxBodySize)
 		if err != nil && !errors.Is(err, ErrBodyTooLarge) {
-			http.Error(rw, fmt.Sprintf("%s: creating reusable request: %v",
-				http.StatusText(http.StatusInternalServerError), err), http.StatusInternalServerError)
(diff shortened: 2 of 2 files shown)


pkg/server/service/loadbalancer/failover/failover.go
```diff
@@ -138,3 +186,27 @@ func (f *Failover) SetFallbackHandlerStatus(ctx context.Context, up bool) {
 		fn(f.handlerStatus || f.fallbackStatus)
 	}
 }
+
+var errBodyTooLarge = errors.New("request body too large")
+
+func (f *Failover) readBodyBytes(req *http.Request) ([]byte, error) {
+	defer req.Body.Close()
+
+	if f.maxRequestBodyBytes <= 0 {
```
rtribotte: I think I would make it consistent with mirroring and -1 would mean infinite size: https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/forwardauth/#maxbodysize

pkg/server/service/loadbalancer/failover/failover.go
```diff
@@ -53,8 +71,38 @@ func (f *Failover) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
-		return
+		if len(f.statusCode) == 0 {
+			f.handler.ServeHTTP(w, req)
+
+			return
+		}
+
+		request := req.Clone(req.Context())
+		if req.Body != nil && req.Body != http.NoBody {
+			bodyBytes, err := f.readBodyBytes(req)
```
rtribotte: The logic already exist for the mirroring service through the `reusableRequest` type, WDYT of reusing it?

pkg/server/service/loadbalancer/failover/response_writer.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (r *responseWriter) Header() http.Header {
+	if r.header == nil {
+		r.header = make(http.Header)
+	}
+
+	return r.header
+}
+
+func (r *responseWriter) WriteHeader(statusCode int) {
```
rtribotte, line 42: I'm thinking about informational headers, would it be correct to replay the request if the client has already received one them?

integration/simple_test.go
```diff
@@ -2365,6 +2365,205 @@ func (s *SimpleSuite) TestEncodedCharactersDifferentEntryPoints() {
...
+	err := try.GetRequest("http://127.0.0.1:8080/api/rawdata", 10*time.Second, try.BodyContains("PathPrefix"))
+	require.NoError(s.T(), err)
+
+	// Verify the router is registered
+	err = try.GetRequest("http://127.0.0.1:8080/api/http/routers", 2*time.Second, try.BodyContains("router1"))
+	require.NoError(s.T(), err)
+
+	// Verify the services are registered
+	err = try.GetRequest("http://127.0.0.1:8080/api/http/services", 2*time.Second, try.BodyContains("fallback-service"))
+	require.NoError(s.T(), err)
```
rtribotte:
```suggestion
```

integration/simple_test.go
```diff
@@ -2365,6 +2365,205 @@ func (s *SimpleSuite) TestEncodedCharactersDifferentEntryPoints() {
...
 }
 
+func (s *SimpleSuite) TestFailoverService() {
+	s.createComposeProject("base")
+
+	s.composeUp()
+	defer s.composeDown()
+
+	whoami1IP := s.getComposeServiceIP("whoami1")
+	whoami2IP := s.getComposeServiceIP("whoami2")
```
rtribotte, line 2375: Considering the other test didn't need an actual whoami server; WDYT of relying only on local go server?

docs/content/reference/routing-configuration/http/load-balancing/service.md
```diff
@@ -940,15 +944,15 @@ http:
...
     [http.services.app.failover]
       service = "main"
       fallback = "backup"
+      [http.services.app.failover.healthCheck]
 
   [http.services.main]
     [http.services.main.loadBalancer]
       [http.services.main.loadBalancer.healthCheck]
-        path = "/health"
+        path = "/status"
```
rtribotte, line 955: @sheddy-traefik Why did you changed the path in this example?

xxsheddy: For consistency with the rest of the examples.

PR #12597, opened 2026-01-26
Add configuration transformer mechanism to the ConfigurationWatcher

sdelicata:
### What does this PR do?

Adds a configuration transformer mechanism to the ConfigurationWatcher that allows modifying dynamic configurations before they are applied. Transformers receive the triggering provider name and the full set of configurations from all providers.

### Motivation

This mechanism complements #12562  (dynamic config extension) by providing a way to compute and populate extended dynamic configuration fields. While #12562 allows extending configuration structures with custom fields, this PR provides the hook to compute and populate those fields based on provider configurations.

This enables dynamic configuration transformations from external components without requiring changes to individual providers. 

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/server/configurationwatcher.go
+++ b/pkg/server/configurationwatcher.go
@@ -28,7 +28,7 @@ type ConfigurationWatcher struct {
 	requiredProvider       string
 	configurationListeners []func(dynamic.Configuration)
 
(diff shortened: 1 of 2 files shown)


juliens (approved):
LGTM

kevinpollet (approved):

PR #12598, opened 2026-01-26
Fix incorrect hostname matching for kubernetes adapter

TheColorman:
### What does this PR do?

Fix `findMatchingHostnames` returning the wrong hostname specificity during comparison.
That behavior is fixed by this PR.

### Motivation

If a listener hostname is `*.foo.com`, and the HTTPRoute hostname is `*.bar.foo.com`, the returned domain would be `*.foo.com`, causing all requests matching `*.foo.com` to be routed to the service that specifically advertised `*.bar.foo.com`.

This is because, when comparing the hostnames, the returned hostname is determined by the number of wildscards in each domain. The one with the least wildcards is picked, and if both hostnames have the same number of wildcards, the listener is picked.
This has been changed to just return the second hostname (HTTPRoute) as previous checks have already verified that it is has higher specificity.

I've also added a test for this behavior.

### More

- [x] Added/updated tests

### Additional Notes

I considered opening an issue, but the fix for this is so small that I thought I'd open a PR, hope that's okay! I'm also not a Go developer, but the logic change is quite simple.


TheColorman:
crap, opened from the wrong source branch 😭

PR #12599, opened 2026-01-26
Fix incorrect hostname matching between listener and route

TheColorman:
### What does this PR do?

Fix `findMatchingHostnames` returning the wrong hostname specificity during comparison.
That behavior is fixed by this PR.

### Motivation

If a listener hostname is `*.foo.com`, and the HTTPRoute hostname is `*.bar.foo.com`, the returned domain would be `*.foo.com`, causing all requests matching `*.foo.com` to be routed to the service that specifically advertised `*.bar.foo.com`.

This is because, when comparing the hostnames, the returned hostname is determined by the number of wildcards in each domain. The one with the fewer wildcards is picked, and if both hostnames have the same number of wildcards, the listener is picked.
This has been changed to just return the second hostname (HTTPRoute) as previous checks have already verified that it is has higher specificity.

I've also added a test checking this behavior.

### More

- [x] Added/updated tests

### Additional Notes

I considered opening an issue, but the fix for this is so small that I thought I'd open a PR, hope that's okay! I'm also not a Go developer, but the logic change is quite simple.

### DIFF
--- a/pkg/provider/kubernetes/gateway/kubernetes.go
+++ b/pkg/provider/kubernetes/gateway/kubernetes.go
@@ -1093,10 +1093,6 @@ func findMatchingHostname(h1, h2 gatev1.Hostname) gatev1.Hostname {
 	}
 
 	trimmedH1 := strings.TrimPrefix(string(h1), "*")
-	// root domain doesn't match subdomain wildcard.
-	if trimmedH1 == string(h2) {
-		return ""
-	}
 
 	if !strings.HasSuffix(string(h2), trimmedH1) {
 		return ""


nmengin:
Hey @TheColorman,

Thanks for the contribution!
We've marked this for "need-triage" to confirm the bug before we move forward.

TheColorman:
Here's some sample k8s resources that demonstrate the bug with an example S3 route:
```yaml
---
# Sample S3 route. Every domain under *.s3.example.com should be forwarded to this service, as each subdomain is an S3 bucket.
kind: HTTPRoute
metadata:
  name: wildcard-s3-route
spec:
  parentRefs:
    - name: traefik-gateway
      sectionName: https
  hostnames:
    - "s3.example.com"
    - "*.s3.example.com"
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /
      backendRefs:
        - name: my-s3-service
          port: 80
---
# Other service that runs on the same example.com domain
kind: HTTPRoute
metadata:
  name: normal-route
spec:
  parentRefs:
    - name: traefik-gateway
      sectionName: https
  hostnames:
    - "simple-service.example.com"
  rules:
    - matches:
        - path:
            type: PathPrefix
            value: /
      backendRefs:
        - name: my-simple-service
          port: 80
---
# And the Traefik Gateway resource with a single listener for *.example.com
kind: GatewayClass
metadata:
  name: traefik
spec:
  controllerName: traefik.io/gateway-controller
---
kind: Gateway
metadata:
  name: traefik-gateway
spec:
  gatewayClassName: traefik
  listeners:
    - name: https
      protocol: HTTPS
      port: 8443
      hostname: "*.example.com"
      tls: ...
```

Using the above config, the Traefik dashboard will report the `httproute-...-wildcard-s3-route-...` router to have the rule
```
Host(`*.example.com`) && PathPrefix(`/`)
```
both preventing any traffic from going to `normal-service`, and causing unexpected traffic to arrive at `wildcard-s3-service`.

rtribotte:
Hi @TheColorman,

Thanks for you feedback!

We agree that there is a bug here, and marked this PR as bug fix.

However, the change feels limited, as `findMatchingHostname` is called in mirror, always return `h2` would only work because of the comparison order, but we are thinking this part should be simply rewritten, would you mind if during review we push commits directly?

TheColorman:
Of course not, go ahead! I'm happy as long as the bug gets resolved :D

TheColorman:
Though even called in mirror, the line last statement before the return should, as far as I understand, ensure that the correct domain is returned:
```go
		return ""
	}
```
e.g. if the order is changed, `h1` is no longer a suffix of `h2`, stopping execution.

rtribotte:
@TheColorman Indeed, but it feels unnecessarily complex. As far as we understand, if both the route and the listener define a host, the listener's host should be used for validation to decide whether the route host is valid. Considering this, we think the code should more clearly reflect that so the question of whether the listener host could be used for the router will never arise.

rtribotte:
Hi @TheColorman,

After trying the current implementation with the following test case:
```
{
	desc:             "Matching wildcard subsubdomain with listener wildcard subdomain",
	listenerHostname: ptr.To(gatev1.Hostname("*.foo.com")),
	routeHostnames:   []gatev1.Hostname{"*.bar.foo.com"},
	want:             []gatev1.Hostname{"*.bar.foo.com"},
	wantOk:           true,
},
```

We determined that it is already working without the modifications you are proposing, so now we are unsure what the problem is.

Also in the reproduction case you shared with us, the generated rule ```Host(`*.example.com`) && PathPrefix(`/`)``` does not seem possible to be produced, as the `buildHostRule` func unit tests indicate, a wildcard hostname rule is always built with the `HostRegexp` matcher.

Could you double check on your side?

TheColorman:
> Also in the reproduction case you shared with us, the generated rule <code>Host(\`*.example.com\`) && PathPrefix(\`/\`)</code> does not seem possible to be produced, as the buildHostRule func unit tests indicate, a wildcard hostname rule is always built with the HostRegexp matcher.

Oops, you're right about that, sorry. Seems I was a bit too quick copying over the rules by eye. I've created a new clean reproduction here.

1. Created a new Ubuntu WSL instance.
2. Set up a 1-node kubernetes cluster using `kubeadm` along with Flannel CNI.
3. Installed Traefik with `helm install traefik traefik/traefik -f values.yaml --wait` using chart/app version `traefik-39.0.5`/`v3.6.10` and the following values.yaml:
   <details>
   <summary><code>values.yaml</code></summary>

   ```yaml
   providers:
     kubernetesGateway:
       enabled: true
   api:
     dashboard: true
   ingressRoute:
     dashboard:
       enabled: true
       matchRule: Host(`dash.localhost`)
       entryPoints: ["web"]
   gateway:
     enabled: false
   service:
     type: NodePort
   ports:
     web:
       exposedPort: 30080
       nodePort: 30080
       port: 30080
     websecure:
       exposedPort: 30443
       nodePort: 30443
       port: 30443
    ```

   </details>
4. Applied the following configuration using `kubectl apply -f reproduction.yaml`:
   <details>
   <summary><code>reproduction.yaml</code></summary>
   
   ```yaml
   ---
   # Sample S3 route. Every domain under *.bar.foo.com should be forwarded to this service.
   kind: HTTPRoute
   metadata:
     name: wildcard-s3-route
   spec:
     parentRefs:
       - name: traefik-gateway
         sectionName: http
     hostnames:
       - "*.bar.foo.com"
     rules:
       - matches:
           - path:
               type: PathPrefix
               value: /
         backendRefs:
           - name: my-s3-service
             port: 80
   ---
   # Other service that runs under the same foo.com domain
   kind: HTTPRoute
   metadata:
     name: normal-route
   spec:
     parentRefs:
       - name: traefik-gateway
         sectionName: http
     hostnames:
       - "baz.foo.com"
     rules:
       - matches:
           - path:
               type: PathPrefix
               value: /
         backendRefs:
           - name: my-simple-service
             port: 80
   ---
   # And the Traefik Gateway resource with a single listener for *.foo.com
   kind: GatewayClass
   metadata:
     name: traefik
   spec:
     controllerName: traefik.io/gateway-controller
   ---
   kind: Gateway
   metadata:
     name: traefik-gateway
   spec:
     gatewayClassName: traefik
     listeners:
       - name: http
         protocol: HTTP
         port: 30080
         hostname: "*.foo.com"
   ---
   # Not sure I needed these
   apiVersion: v1
   kind: Service
   metadata:
     name: my-s3-service
   spec:
     selector:
       app: s3-service
     ports:
       - port: 80
   ---
   apiVersion: v1
   kind: Service
   metadata:
     name: my-simple-service
   spec:
     selector:
       app: simple-service
     ports:
       - port: 80
   ```
   </details>
5. Forwarded the Traefik HTTP port to my host with `kubectl port-forward svc/traefik 30080:30080 30443:30443` so that I could access the dashboard at `dash.localhost:30080`.

This generated 2 HTTP Routers in Traefik with the following Rules (copied directly from the dashboard this time 😅) (expand to see the raw json response from Traefik API):
<ul><li><details>
  <summary><code>Host("baz.foo.com") && PathPrefix("/")</code></summary>
  
  ```json
  {
    "entryPoints": [
      "web"
    ],
    "service": "httproute-default-normal-route-gw-default-traefik-gateway-ep-web-0-8655affa5fcb02c10fe6-wrr",
    "rule": "Host(\"baz.foo.com\") && PathPrefix(\"/\")",
    "ruleSyntax": "default",
    "priority": 13,
    "observability": {
      "accessLogs": true,
      "metrics": true,
      "tracing": true,
      "traceVerbosity": "minimal"
    },
    "status": "enabled",
    "using": [
      "web"
    ],
    "name": "httproute-default-normal-route-gw-default-traefik-gateway-ep-web-0-8655affa5fcb02c10fe6@kubernetesgateway",
    "provider": "kubernetesgateway",
    "priorityStr": "13"
  }
  ```
  
  </details></li>
<li><details>
  <summary><code>HostRegexp("^[a-z0-9-\\.]+\\.foo\\.com$") && PathPrefix("/")</code></summary>
  
  ```json
  {
    "entryPoints": ["web"],
    "service": "httproute-default-wildcard-s3-route-gw-default-traefik-gateway-ep-web-0-6e55a4edfb06fa4a1e17-wrr",
    "rule": "HostRegexp(\"^[a-z0-9-\\\\.]+\\\\.foo\\\\.com$\") && PathPrefix(\"/\")",
    "ruleSyntax": "default",
    "priority": 11,
    "observability": {
      "accessLogs": true,
      "metrics": true,
      "tracing": true,
      "traceVerbosity": "minimal"
    },
    "status": "enabled",
    "using": ["web"],
    "name": "httproute-default-wildcard-s3-route-gw-default-traefik-gateway-ep-web-0-6e55a4edfb06fa4a1e17@kubernetesgateway",
    "provider": "kubernetesgateway",
    "priorityStr": "11"
  }
  ```
  
  </details></li></ul>
  
  
The second route essentially gets assigned the `*.foo.com` wildcard, which means `baz.foo.com` gets no traffic.

This seems awfully similar to your successful test case. Is there a different version of Traefik/the chart I should be trying this with?

PR #12600, opened 2026-01-26
Fix the errors middleware's document for Kubernetes CRD

yuito-it:
### What does this PR do?

Fixes a syntax error in the documentation for the errors middleware configuration that caused validation failures when applying the manifest.

### Motivation

While following the documentation, the provided example caused a Kubernetes validation error when running:

```bash
kubectl apply -f traefik/error.yaml
```

The error message was:

```
The Middleware "custom-error-page" is invalid: 
* spec.errors.statusRewrites.418: Invalid value: "string": spec.errors.statusRewrites.418 in body must be of type integer: "string"
* <nil>: Invalid value: null: some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
```

This PR corrects the configuration example so that it uses the proper data types and no longer produces a syntax error.

### More

- [x] Added/updated tests

### Additional Notes

This change only affects the documentation example and does not modify runtime behavior.

### DIFF
--- a/docs/content/reference/routing-configuration/http/middlewares/errorpages.md
+++ b/docs/content/reference/routing-configuration/http/middlewares/errorpages.md
@@ -19,8 +19,8 @@ http:
           - "503"
           - "505-599"
         statusRewrites:
-          "418": "404"
-          "502-504": "500"
+          "418": 404
+          "502-504": 500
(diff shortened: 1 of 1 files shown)


traefiker:

PR #12606, opened 2026-01-27
HTTP Healthcheck for TCP services

supergoudvis116:
### What does this PR do?

Adds support for performing HTTP health checks for TCP services in Traefik.
This allows TCP routing decisions to be based on the result of an HTTP endpoint rather than just a TCP-level check.

### Motivation

For clustered services such as Patroni, etcd, PostgreSQL, or similar setups, a simple TCP check is often insufficient.
These systems usually expose an HTTP API that indicates whether a node is healthy, functional, or currently the primary/leader.

This change enables Traefik to:

Poll an HTTP endpoint for health information

Route traffic only to nodes that are healthy or in the correct role (e.g. primary)

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/reference/routing-configuration/tcp/service.md
+++ b/docs/content/reference/routing-configuration/tcp/service.md
@@ -73,6 +73,44 @@ labels:
 }
 ```
 
+#### HTTP Health Check Configuration Example
+
+You can also configure health checks using HTTP mode:
+
+```yaml tab="Structured (YAML)"
+tcp:
+  services:
+    my-service:
+      loadBalancer:
+        servers:
+        - address: "xx.xx.xx.xx:8080"
+        healthCheck:
+          mode: "HTTP"
+          httpHealthCheck:
+            path: "/health"
+            method: "GET"
+            status: 200
+          interval: "10s"
+          timeout: "3s"
+```
+
+```toml tab="Structured (TOML)"
+[tcp.services]
+  [tcp.services.my-service.loadBalancer]
+    [[tcp.services.my-service.loadBalancer.servers]]
+      address = "xx.xx.xx.xx:8080"
+
+    [tcp.services.my-service.loadBalancer.healthCheck]
+      mode = "HTTP"
+      interval = "10s"
+      timeout = "3s"
+
+      [tcp.services.my-service.loadBalancer.healthCheck.httpHealthCheck]
+        path = "/health"
+        method = "GET"
+        status = 200
+```
+
 ### Configuration Options
 
 | Field | Description                                 | Default |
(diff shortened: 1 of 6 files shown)


nmengin:
Hey @supergoudvis116,

Thanks for this!

supergoudvis116:
Hey @nmengin,

If there is any feedback I can easily fix myself to make things easier for you all let me know! 

:)

mneundorfer:
@supergoudvis116 Would this also work if I'd configure the TCP router for TLS passthrough...?

Reason I'm asking: I would assume that in that case, the healthcheck would fail as it would try to reach the health check path via HTTP instead of HTTPS?

supergoudvis116:
> @supergoudvis116 Would this also work if I'd configure the TCP router for TLS passthrough...?

Currently it only supports HTTP, it does not go to HTTPS if you enabled TLS passtrough. We could easily modify the config so it allows for a scheme in the HTTP config so you can force the check to request over HTTPS.

I dont think its beneficial to automatically make the healthcheck go over HTTPS when TLs passtrough is enabled.

SimonMTS:
Hey @nmengin ,

Do you have any indication as to when this might be reviewed?

kevinpollet:
Hey,

ahus1:
Hi, this @ahus1, one of the maintainers of the @keycloak project which is offering identity and access management with protocols like SAML or OIDC. 

We're currently attempting to document a TLS passthrough setup for Keycloak leveraging Traefik for our community.

As Keycloak can expose a health check on a separate HTTP management endpoint, the best we could do with the current version v3.6 is to use TCP check that talks HTTP - see below for the configuration snippet [1]. 

This was crude to configure, and is probably brittle as any character change in the request or response will break the check. It is also not encrypted. 

Do you have any feedback for us how this could be simplified with today's options? 

So why are we not using simple TCP check? The readiness probe via HTTP allows us a more specific handling like a graceful shutdown of Keycloak nodes where they first mark themselves as not ready, then draining TCP connections, and then eventually shutting down and restarting. They can also mark themselves not ready when they lose connection to other cluster nodes or to the database. This already works quite well on infrastructures like Kubernetes where Pods mark themselves as ready via readiness probe. 

We're looking forward for Traefik to eventually support a HTTP(S) probe for TCP services, as it would make our setup more robust. 

Thanks!

[1] Configuration snippet for attempting to do HTTP over TCP. 
 

```
  services:
    keycloak-service:
      loadBalancer:
        serversTransport: kc-passthrough
        # Use the JSON syntax to have explicit CR/LF encoding for send/expect.
        healthCheck: {
          interval: "5s",
          timeout: "3s",
          port: 9000,
          send: "HEAD /health/ready HTTP/1.0\r\n\r\n",
          expect: "HTTP/1.0 200 OK\r\ncontent-type: application/json; charset=UTF-8\r\ncache-control: no-store\r\n\r\n"
        }
        servers:
          - address: "keycloak1:8443"
          - address: "keycloak2:8443"
```

kevinpollet:
Hi @ahus1,

> Do you have any feedback for us how this could be simplified with today's options?

Thanks for the feedback. Today, this is the right way to do it, and this pull request will address this use case by adding an HTTP health check for TCP backends. This feature is on our roadmap, but for the moment, we are focused elsewhere.

PR #12610, opened 2026-01-27
Add support for from-to-www-redirect NGINX annotation

LBF38:
### What does this PR do?

This PR adds support for `from-to-www-redirect` nginx annotation ([see nginx docs](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#redirect-fromto-www)).

### Motivation

To add more annotations.

### More

- [x] Added/updated tests


pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -813,6 +826,8 @@ func (p *Provider) applyMiddlewares(namespace, routerKey, rulePath string, ingre
 
 	applyRewriteTargetConfiguration(rulePath, routerKey, ingressConfig, rt, conf)
 
+	applyFromToWwwRedirect(hosts, ruleHost, routerKey, ingressConfig, rt, conf)
```
rtribotte: @LBF38 Shouldn't this be added just below appRoot (as well as the other redirect)?

LBF38: Thanks for noticing it !
I've tested other annotation combinations and reordered the middlewares accordingly.

mmatur (approved):
LGTM

Issue #6156, opened 2020-01-11
Feature request : SSL/TLS certificate status in web UI


erichiller:
I'd love to see SSL/TLS certificate status in the web UI. For example:
- see each domain's certificate
- status of each certificate, date of last renewal, expiry
- any to messages pertinent to the certificate's renewal.

Thanks for all the work!

msonowal:
It should list the certs from the resolvers and show information for each of them so that we know

Dandrum:
Ability to revoke certs would be nice and showing a list of Certs based on ACME/Certbot list from LE server instead of ACME.json
If ACME.json is empty or dosnt have a Cert but LE Server contains Certs for Domain and User (Email + Priv Key) Then reuse the Server Cert instead of trying to request a new one (prevent LE limits)

b0r1sp:
I'd like to kindly add:

- ability to force-renew certs
- ability to revoke/delete certs

lucasff:
having an issue right now that a certificate is issued with staging LE API, and I want to force renew to production.

heitorPB:
It would also be nice to see the configuration for each resolver: min TLS version, list of ciphers, ...

JoeHitchen:
I tried to have a quick look at this (as an outsider) since I think it's an important thing to have displayed but it seems the API only really describes the dynamic configuration, with a bit of information about the static configuration. I guess the certificate information is buried deep down in the internals and would need a bit of work to make accessible?

bluepuma77:
It would be great to see more information about TLS in the dashboard than what's currently provided:

<img width="490" alt="Traefik-Dashboard-TLS" src="https://user-images.githubusercontent.com/5221187/190997467-1de6b720-cf9c-4dbb-969d-0b7d043f6640.png">

The only info available is that TLS is enabled. But there is zero information regarding the used certificate.

TLS is even green when no dedicated static certificate nor a LetsEncrypt certificate is found. But from an end-user perspective a Traefik self-generated certificate is not really "green" when it brings up browser warning messages.

We try to supply static TLS certificates via HTTP provider with 15sec pollInterval, so even more problems can occur.

Therefore it would be great to see the actual used certificate for the router/host in the dashboard, specifically the issuer, included domains and expiry date.

ddtmachado:
Hey everyone, I see many different feature requests embedded in comments and altought the original request is generic it doesn't mean we can fit everything in it.

So let's draw some limits on what could be achieved or not:

[...]

PR #12628, opened 2026-02-05
Add certificates menu and overview

holomekc:
Fixes: #6156
At least partially

### What does this PR do?

This PR creates a dedicated view for certificate information. I locally already have more versions, which would allow renewal or revocation / deletion of ACME certificates, but the changes are too many and too wild. So I thought, maybe first start with showing some information.

It was more difficult than I thought. Reason one is that I am not so familiar with go and vite. The second "issue" I had is that a certificate does not have a key to identify it. At least I could not find one. I think at the moment you use the domains to find the best perfect match. So I thought I do something similar.

Nevertheless, here the things which should be checked:
1. certKey. base64(domains[]) where domains are sorted. No idea if this is a good idea, but this what came to my mind.
2. Changes in tlsManager.go. I think using Leaf is safe, but as said I am not familiar with go. This should prevent extracting cert info over and over again
3. Due to my changes I changed the domains[] representation in the UI. Well I kind of replaced it
4. Showing the cert info in the HTTP, TCP Router section is kind of difficult. If domains are configured it is easy, but if not it tries to extract the info from the rule. Which can be kind of difficult.
5. I still have the "resolver" (ACME resolver/provider) information in the API, which is not set. The reason for that is that it is kind of complicated in the backend to do. I would need to iterate over all certs (which is done anyway for mapping), but also then use the domains to check via acme provider to find a suitable one. As said I have another branch where I have all this, but it is really messy. I think it would be a great info, but I am not sure if there is a better way to find the proper one.

I cannot explain the UI so some pictures:

1. Running in docker locally on my raspi
image
image

2. webui dev
image

3. HTTP Routers on my raspi
image
4. webui dev when multiple domains are defined 
image

I hope this helps somehow. If not it was at least helpful to learn a little bit for me. Besides that feel free to tell me what needs to be changed or just edit yourself. You should have the permissions.

### Motivation

#6156

### More
-

### Additional Notes
-

### DIFF
--- a/webui/src/pages/certificates/Certificates.tsx
+++ b/webui/src/pages/certificates/Certificates.tsx
@@ -29,6 +29,11 @@ export const CertificateRenderRow: RenderRowType = (row: unknown) => {
       <AriaTd>
         <TooltipText text={cert.commonName || '-'} />
       </AriaTd>
+      <AriaTd css={{ maxWidth: '240px' }}>
+        <Text css={{ wordBreak: 'break-word', whiteSpace: 'normal' }}>
+          {cert.sans?.length > 0 ? cert.sans.join(', ') : '-'}
+        </Text>
+      </AriaTd>
       <AriaTd>
         <TooltipText text={cert.issuerOrg || cert.issuerCN || 'Unknown'} />
       </AriaTd>
@@ -64,6 +69,7 @@ export const CertificatesRender = ({
           <AriaTr>
             <SortableTh label="Status" isSortable sortByValue="status" css={{ width: '36px' }} />
             <SortableTh label="Common Name" isSortable sortByValue="cn" />
+            <SortableTh label="SANs" css={{ maxWidth: '240px' }} />
             <SortableTh label="Issuer" isSortable sortByValue="issuer" />
             <SortableTh label="Valid Until" isSortable sortByValue="validUntil" css={{ width: '100px' }} />
             <SortableTh label="Expiry" css={{ width: '100px' }} />


nmengin:
Hey @holomekc,
Thanks for this!

webui/src/components/certificates/CertificateDetails.tsx
```diff
@@ -0,0 +1,152 @@
+import { Badge, Box, Flex, Grid, Text } from '@traefiklabs/faency'
+
+// Matches the API response structure exactly
+export interface CertificateInfo {
```
gndz07: There is a `types` directory that holds all the resources types that could be reused across the project. I think it'd be great to have this type declaration on the same place, so that we could follow the same structure with other components. Also, we wouldn't need to explicitly export the types.

webui/src/components/certificates/CertificateDetails.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+export const CertificateDetails = ({ certificate }: CertificateDetailsProps) => {
+  const validFrom = new Date(certificate.notBefore)
+  const validUntil = new Date(certificate.notAfter)
+
+  return (
+    <Flex direction="column" gap={4}>
+      <Box>
+        <Text css={{ fontWeight: 600, mb: '$3' }}>Issued To</Text>
+        <Grid css={{ gridTemplateColumns: '200px 1fr', gap: '$3' }}>
```
gndz07: For the sake of style consistency for all of the pages across the dashboard, I think instead of building a new `Grid` from scratch here we can reuse the same `<DetailsCard />` component, WDYT? Here is an [example of the similar implementation on TlsSection component](https://github.com/traefik/traefik/blob/4a4be524bb38334b427849a184aa92141dd3973d/webui/src/components/routers/TlsSection.tsx#L51).

webui/src/components/certificates/CertificateDetails.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <Text variant="subtle">Common Name:</Text>
+          <div>
+            <a href={`//${certificate.commonName}`} target="_blank" rel="noopener noreferrer" style={{ textDecoration: 'none', display: 'inline' }}>
+              <Text css={{ color: '$primary', fontWeight: 600, '&:hover': { textDecoration: 'underline' } }}>{certificate.commonName}</Text>
+            </a>
+          </div>
```
gndz07: We can use Link component from faency here, and the Text component wouldn't be needed.
```suggestion
          <Link variant="blue" href={`//${certificate.commonName}`} target="_blank" rel="noopener noreferrer">
              {certificate.commonName}
          </Link>
```

webui/src/components/certificates/CertificateDetails.tsx
```diff
@@ -0,0 +1,152 @@
...
+          
+          <Text variant="subtle">Status:</Text>
+          <div>
+            <Badge size="small" variant={getCertStatusColor(certificate.daysLeft).variant}>
+              {getCertStatusColor(certificate.daysLeft).label}
```
gndz07: I think `getCertStatusColor(certificate.daysLeft)` could benefit from memoization.

webui/src/components/certificates/CertificateDetails.tsx
```diff
@@ -0,0 +1,152 @@
...
+                {san.startsWith('*.') ? (
+                  <Text>{san}</Text>
+                ) : (
+                  <a href={`//${san}`} target="_blank" rel="noopener noreferrer" style={{ textDecoration: 'none', display: 'inline-block' }}>
+                    <Text css={{ color: '$primary', '&:hover': { textDecoration: 'underline' } }}>{san}</Text>
+                  </a>
+                )}
+              </div>
+            ))}
+          </div>
```
gndz07:
```suggestion
          <Box>
            {certificate.sans.map((san, idx) => (
              <Box key={idx}>
                {san.startsWith('*.') ? (
                  <Text>{san}</Text>
                ) : (
                  <Link variant="blue" href={`//${san}`} target="_blank" rel="noopener noreferrer" >
                   {san}
                  </Link>
                )}
              </Box>
            ))}
          </Box>
```

webui/src/pages/certificates/Certificate.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { useCertificate } from '../../hooks/use-certificates'
+
+import PageTitle from 'layout/PageTitle'
+
+export const Certificate = () => {
+  const { name } = useParams<{ name: string }>()
+  const { certificate, isLoading, error } = useCertificate(name || '')
+
+  if (isLoading) {
+    return <div>Loading certificate...</div>
```
gndz07: By wrapping it with faency's component instead of a plain `div`, the style will be inherited. Another pattern is that on other pages, the loading state is marked with skeleton ([example on RouterDetails page](https://github.com/traefik/traefik/blob/4a4be524bb38334b427849a184aa92141dd3973d/webui/src/components/routers/RouterDetail.tsx#L32)).
```suggestion
    return <Box>Loading certificate...</Box>
```

webui/src/pages/certificates/Certificate.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  if (error) {
+    return <div>Error loading certificate: {error.message}</div>
```
gndz07:
```suggestion
    return (
      <>
        <PageTitle title={certificate?.commonName || name} />
        <Text data-testid="error-text">
          Sorry, we could not fetch detail information for this Certificate right now. Please, try again later.
        </Text>
      </>
    )
```

webui/src/pages/certificates/Certificate.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  if (!certificate) {
+    return <div>Certificate not found</div>
```
gndz07: There is a `<NotFound />` component that could be reused.
```suggestion
    return <NotFound />
```

webui/src/components/routers/TlsSection.spec.tsx
```diff
@@ -0,0 +1,316 @@
...
+    vi.spyOn(useCertificates, 'useCertificate').mockReturnValue({
+      certificate: mockCertificate,
+      isLoading: false,
+      error: null,
+    })
+
+    const tlsData: Router.TLS = {
+      certResolver: 'letsencrypt',
+      domains: [{ main: 'example.com', sans: ['www.example.com'] }],
+    }
```
gndz07: This invokes TS errors, if it makes sense `options` and `passthrough` could be made as optional inside the `Router.TLS` types.

webui/src/components/certificates/CertificateDetails.tsx
```diff
@@ -0,0 +1,152 @@
...
+  notBefore: string
+  notAfter: string
+  daysLeft: number
+  version?: string
+  keyType?: string
+  keySize?: number
+  signatureAlgorithm?: string
+  certFingerprint?: string
+  publicKeyFingerprint?: string
+  status?: string
```
gndz07: This invokes TS error.
```suggestion
  status?: 'enabled' | 'disabled' | 'warning'
```

webui/src/components/routers/TlsSection.tsx
```diff
@@ -9,46 +12,98 @@ import DetailsCard, { SectionTitle } from 'components/resources/DetailsCard'
...
+  let match
+  while ((match = hostRegex.exec(rule)) !== null) {
+    const domain = match[1]
+    // Filter out regex patterns and wildcards
+    if (!domain.includes('{') && !domain.includes('*') && !domain.includes('[')) {
+      domains.push(domain)
+    }
+  }
+  return domains
 }
```
gndz07: Refactor to fit ES2020.
```suggestion
const extractDomainsFromRule = (rule: string): string[] =>
  [...rule.matchAll(/Host(?:SNI|Regexp)?\(`([^`]+)`\)/g)]
    .map(([, domain]) => domain)
    .filter((domain) => !/[{*[]/.test(domain))
```

webui/src/hooks/use-certificates.tsx
```diff
@@ -0,0 +1,84 @@
+import { useMemo } from 'react'
+import useSWR from 'swr'
+
+export interface Certificate {
```
gndz07: Is it completely different than the type inside CertificateDetails.tsx? Same kind of suggestion to unify all the types inside the `types` directory.

webui/src/pages/certificates/Certificates.tsx
```diff
@@ -0,0 +1,134 @@
...
+import TooltipText from 'components/TooltipText'
+import useFetchWithPagination, { pagesResponseInterface, RenderRowType } from 'hooks/use-fetch-with-pagination'
+import { EmptyPlaceholderTd } from 'layout/EmptyPlaceholder'
+import PageTitle from 'layout/PageTitle'
+
+const getExpiryColor = (daysLeft: number) => {
+  if (daysLeft < 0) return 'red'
+  if (daysLeft < 14) return 'orange'
+  return 'green'
+}
```
gndz07: This function is slightly different with `getCertStatusColor` in CertificateDetails, but I believe the end goals are the same, which is to construct the Badge component that will show how many days left until expiration and its corresponding color. To keep consistency of this Badge component, I think they could be refactored so they'll use the same component/function, WDYT?

webui/src/pages/certificates/Certificates.tsx
```diff
@@ -0,0 +1,134 @@
...
+  })
+
+  return (
+    <>
+      <AriaTable>
+        <AriaThead>
+          <AriaTr>
+            <SortableTh label="Status" isSortable sortByValue="status" css={{ width: '36px' }} />
+            <SortableTh label="Common Name" isSortable sortByValue="cn" />
+            <SortableTh label="Issuer" isSortable sortByValue="issuer" css={{ width: '120px' }} />
```
gndz07: A question, why is this column limited to 120px? In some cases this could make it wrapped while there's still a lot of space available.

Image

webui/src/pages/certificates/Certificate.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <div>
+      <PageTitle title={`Certificate: ${certificate.commonName}`} />
+      
+      <Card css={{ p: '$4' }}>
+        <H2 css={{ mb: '$4' }}>Certificate</H2>
+        <CertificateDetails certificate={certificate} />
+      </Card>
+    </div>
```
gndz07: A suggestion, to keep the layout inline with other details pages, WDYT of this?

Image

```suggestion
     <>
      <PageTitle title={`Certificate: ${certificate.commonName}`} />
      <Flex gap={2} align="center">
        <H1 css={{ mb: '$4' }}>{certificate.commonName}</H1>
        <ResourceStatus status={certificate.status || 'disabled'} />
      </Flex>
      <Card css={{ p: '$4' }}>
        <CertificateDetails certificate={certificate} />
      </Card>
     </>
```

gndz07 (changes requested):
Hi @holomekc,

Thank you for opening the PR! I left some review comments, mainly about the typing and the preference to use components from Faency so that the global style could be inherited automatically.

Please let me know if you have some questions or if you disagree with some of my comments 🙂

holomekc:
@gndz07 
Thx for all the feedback <3. I need some time to check everything. I will most likely start this weekend.

webui/src/components/certificates/CertificateDetails.tsx
holomekc: I like the idea, but I am not sure how to do that properly.

I can do something like this:
```tsx
...
  const fingerprintItems = [
    { key: 'Certificate', val: certificate.certFingerprint || 'N/A' },
    { key: 'Public Key', val: certificate.publicKeyFingerprint || 'N/A' },
  ]

  return (
    <Flex direction="column" gap={2}>
      <SectionTitle title="Issued To" />
      <DetailsCard items={issuedToItems} keyColumns={1} />

      <SectionTitle title="Issued By" />
      <DetailsCard items={issuedByItems} keyColumns={1} />

      <SectionTitle title="Validity" />
      <DetailsCard items={validityItems} keyColumns={1} />

      <SectionTitle title="Technical Details" />
      <DetailsCard items={technicalItems} keyColumns={1} />

      <SectionTitle title="SHA-256 Fingerprints" />
      <DetailsCard items={fingerprintItems} keyColumns={1} />
    </Flex>
  )
```
this looks like this:
image

What I don’t like is that we end up with too many vertical lines. It makes the design look too busy.

Variant 1: above
Variant 2: Put everything in one DetailCard
image

Variant 3: Similar to variant 1, but with minKeyWidth=200px
image


I like three the most, but I am not sure regarding the hardcoded 200px though. What do you think?

webui/src/components/certificates/CertificateDetails.tsx
holomekc: Sounds good. I need to use ValText though, otherwise the text has 14px instead of 16px as all other DetailsCard values. I posted some images already at the comment regarding the usage suggestion of DetailsCard.

webui/src/hooks/use-certificates.tsx
holomekc: This is my bad. Not sure if this is wise, but I tried the following in resources.d.ts:
```tsx
declare namespace Certificate {
  /** Raw API response shape */
  type Raw = {
    name?: string
    commonName: string
    sans: string[]
    issuer?: string
    issuerOrg?: string
    issuerCN?: string
    issuerCountry?: string
    organization?: string
    country?: string
    subject?: string
    serialNumber?: string
    notBefore: string
    notAfter: string
    version?: string
    keyType?: string
    keySize?: number
    signatureAlgorithm?: string
    certFingerprint?: string
    publicKeyFingerprint?: string
    status?: 'enabled' | 'disabled' | 'warning'
    resolver?: string
    usedBy?: string[]
  }

  /** Enriched certificate with computed fields */
  type Info = Raw & {
    daysLeft: number
  }
}

```

webui/src/pages/certificates/Certificates.tsx
holomekc: Moved it to a dedicated ExpiryBadge.

webui/src/pages/certificates/Certificates.tsx
holomekc: This is without the width limitation:

I am unsure. My brain said ok common name should take all the space it can get. 

commonName: 45%, issuer: 30%
image

no css so 50:50
image

What do you think? I am fine with either.

webui/src/pages/certificates/Certificate.tsx
holomekc: Yeah good thinking.
image

Not sure about the ResourceStatus? I could not find it on the other pages. I moved the mb to the flex element:
```tsx
      <Flex gap={2} align="center"  css={{ mb: '$4' }}>
        <H1>{certificate.commonName}</H1>
        <ResourceStatus status={certificate.status || 'disabled'} />
      </Flex>
```
Otherwise, the status was not aligned.

webui/src/components/certificates/CertificateDetails.tsx
gndz07: I agree, I like the 3rd option best as well 👍

webui/src/pages/certificates/Certificates.tsx
gndz07: I think no css is fine, it should adjust based on the content if needed. Thanks for checking this!

webui/src/pages/certificates/Certificate.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const { name } = useParams<{ name: string }>()
+  const { certificate, isLoading, error } = useCertificate(name || '')
+
+  if (isLoading) {
+    return (
+      <Box>
+        <PageTitle title={name || ''} />
+        <Skeleton css={{ height: '$7', width: '320px', mb: '$7' }} data-testid="skeleton" />
+        <Flex direction="column" gap={6}>
+          <DetailsCardSkeleton />
```
gndz07: Just to make it closer to the actual UI.
```suggestion
          <DetailsCardSkeleton keyColumns={1} rows={5} />
```

webui/src/pages/certificates/Certificate.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return (
+    <>
+      <PageTitle title={`Certificate: ${certificate.commonName}`} />
+      <Flex gap={2} align="center"  css={{ mb: '$4' }}>
+        <H1>{certificate.commonName}</H1>
+        <ResourceStatus status={certificate.status || 'disabled'} />
+      </Flex>
+      <Card css={{ p: '$4' }}>
+        <CertificateDetails certificate={certificate} />
+      </Card>
```
gndz07:
```suggestion
        <CertificateDetails certificate={certificate} />
```

pkg/api/certificate.go
```diff
@@ -0,0 +1,263 @@
...
+		normalized[i] = strings.ToLower(strings.TrimSpace(domain))
+	}
+	sort.Strings(normalized)
+
+	return normalized, nil
+}
+
+// buildCertKey creates a base64-encoded certKey from a list of domains
+// Deduplicates, lowercases, sorts, and encodes domains
+func buildCertKey(domains []string) string {
```
gndz07: You've raised a doubt about this `certKey` as identifier that is built from sorted domains, and maybe I share the same concern. Please correct me if I misunderstood, but seems to me that there's a possibility that this `certKey` might not be really unique (e.g. original cert vs a renewed version), hence it might not be a very reliable identifier. WDYT of using fingerprint as identifier?
However, this will have impact on how the Router page on the UI fetch certificate details.

webui/src/components/certificates/CertificateDetails.tsx
holomekc: Ok awesome. Three it is. Due to other changes below there is one tiny question left. Certificate.tsx -> Card variant normal or ghost.

Normal:
image

Ghost:
image

I like ghost more, because no overlapping card elevation. I pushed ghost variant now (or in a few secs), but I can change it.

webui/src/pages/certificates/Certificates.tsx
holomekc: No problem. Sounds good. No css it is.

webui/src/components/certificates/CertificateDetails.tsx
holomekc: Ok forget it. I see below. No card it is :)

pkg/api/certificate.go
holomekc: ## Certificates Page
I need to check. fingerprint sounds so much better for the certificates table. If I remember correctly (I am getting old...) I did this to define only one endpoint, which is usable for the certificates page as for the router page. But maybe this is not that great.

## Router page
The current idea is based on the data, which was already present in the backend. So the chain is:
GET /api/certificates/{certKey} → getDomainsFromURLEncodedCertKey(certKey) → tlsManager.GetStore("default") → store.GetCertificate(domains) → ...

So the benefit was that I could rely on the existing logic and I did not need to come up with my own solution. But I am not sure if this is the proper way. Because this is not what incoming requests use. We will see that in a sec.

### Problem of the router page
Yeah... challenging. The first thing is that domains are not always available. So why the host rule approach?

As far as I understood it the selection of a cert is:
Incoming HTTPS request (SNI: "app.example.com") → GetBestCertificate → cache lookup by "app.example.com" → miss → iterate DynamicCerts keys (e.g. "*.example.com,example.com") → split → matchDomain("app.example.com", "*.example.com") → match → cache result → return cert → TLS handshake completes

This happens before and independently from the router selection.

Now I am a bit wild guessing. The complete certificate approach seems to be domains related. For ACME a new cert is requested based on domains → stored if not already exists. Similar for "static" certificates. parse cert → extract SANs → stored. So all certs end up in the same store.

So at the moment there is no reference between a router and a certificate. The "trick" I tried to use is to guess the domain. if the domain is not already defined for the router. Depending on how complicated the Host rules are the guessing can fail. But if I guess a domain properly I can get the cert from the store. I cannot use the same logic incoming HTTP request use, because the logic works with the TLS Handshake and I cannot really simulate that. I think.


Ok sorry. A lot of text. Long story short. I think fingerprint for certificates table is a valid option. For router page I see no other approach without major changes.

holomekc:
Hi @gndz07 I think I applied the latest requested UI changes now. I hope at least. I think the biggest topic is now "only" the how to find the proper certificate issue.

pkg/api/certificate.go
holomekc: @gndz07 
I think we have multiple options:
1. keep it as it is
2. Use fingerprint for table, keep certkey for router
3. Use fingerprint for table, in router only shown when domains are specified
4. Use fingerprint for table, remove changes in router

gndz07: Hi @holomekc,

Sorry for the delay. We just discussed with the team, and we think that it doesn't make much sense to have the certificate information on the router page, so if you don't mind we can just keep the new certificate list and details page and remove the changes from the router page.

gndz07:
Hi @holomekc,

To help us move this forward faster instead of going back and forth on small changes, would you mind if we push some commits directly to your branch?
Let us know if that works for you!

traefiker:

holomekc:
Hi @nmengin @gndz07 . Sorry I am still on vacation. Sure feel free to edit whatever you like. Well you already did ;). All changes are fine with me. Thx for pushing this. I will be back next week. If you need something urgent from me. Spam me messages.

Happy Easter
Chris

traefiker:

schildbach:
See #12946, LetsEncrypt shortlived profiles are all marked with a "Expires soon" warning despite there is nothing wrong with them.

nmengin:
Hey @schildbach,

Thank you for the feedback. Could you please open a dedicated issue to allow us to keep track of it?

Thanks in advance.

schildbach:
> Thank you for the feedback. Could you please open a dedicated issue to allow us to keep track of it?

It's https://github.com/traefik/traefik/issues/12946

zalbiraw:
Thank you @holomekc, this is such a wonderful addition!!!

PR #12630, opened 2026-02-06
Add support for proxy-read-timeout and proxy-send-timeout NGINX annotations

LBF38:
### What does this PR do?

This PR adds support for `proxy-read-timeout` and `proxy-send-timeout` NGINX annotations

### Motivation

To add more NGINX annotations.

### More

- [x] Added/updated tests

### Additional Notes

Closes https://github.com/traefik/traefik/issues/12627

Co-authored-by: Romain Tribotte <rtribotte@users.noreply.github.com>


rv0lt:
when could this be merged ?

nmengin:
Hey @rv0lt,

Thanks for reaching out.
We're actively working on NGINX annotations to embed them in the next minor version of Traefik (which should happen soon).

pkg/server/service/transport.go
```diff
@@ -233,6 +233,54 @@ func (t *TransportManager) createTLSConfig(cfg *dynamic.ServersTransport) (*tls.
...
+	}
+
+	return n, nil
+}
+
+func customDialContext(d *net.Dialer, cfg *dynamic.ForwardingTimeouts) func(ctx context.Context, network string, address string) (net.Conn, error) {
+	return func(ctx context.Context, network, address string) (net.Conn, error) {
+		conn, err := d.DialContext(ctx, network, address)
+
+		if cfg.ReadTimeout > 0 && cfg.WriteTimeout > 0 {
```
juliens: We may have one to zero but not the other?

LBF38: I agree. What should be the default value for each timeout? Or should I move the zero value check in the connWithTimeout?

Also, I can add some tests for this func if necessary

PR #12649, opened 2026-02-11
Simplify Go version refs in CI templates

jnoordsij:
### What does this PR do?

To ever so slightly simplify the amount of bumps and changes needed to update the Go version, this PR proposes to use the `go-version-file` functionality of the `setup-go` GitHub action, which allows extracting the version from a specified file. Consequently this means that only the `go.mod` file in this project contains the (single source of truth) Go version.
See also https://github.com/actions/setup-go?tab=readme-ov-file#go-version-file.

It also ensures that the `check-latest: true` flag is applied consistently; if this PR would not make it entirely, this should probably be included as individual fix, to ensure all jobs always run on latest patch version rather than the (latest) cached version.

### Motivation

Simplifying Go version bump process.

### More

- [ ] Added/updated tests

### Additional Notes

If a need ever arises, the `go-version` identifier can still be set on individual jobs, taking precedence over the version file directive. 

Also there are no longer any explicit Go version mentioned in CI job step names, but this should be clear enough anyways.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,10 @@
 module github.com/traefik/traefik/v2
 
+// Only specify the major.minor version here.
+// CI workflows read this value via go-version-file and use check-latest
+// to automatically pick up the newest patch release.
+// Pinning a patch version would defeat that mechanism and require manual
+// bumps for every Go patch release.
 go 1.25
 
 require (


.github/workflows/test-integration.yaml
```diff
@@ -42,7 +41,7 @@ jobs:
         with:
           path: |
             ~/.cache/go-build
-          key: ${{ runner.os }}-go-build-cache-${{ env.GO_VERSION }}-${{ hashFiles('**/go.sum') }}
```
jnoordsij, line 44: I'm not really sure if this was really necessary given that including `go.sum` should already be sufficient AFAICS, however adding in `go.mod` is probably harmless and the amount of times the Go version there will be updated is very limited anyways.

.github/workflows/codeql.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       if: ${{ matrix.language == 'go' }}
       with:
         go-version-file: 'go.mod'
```
jnoordsij, line 37: FWIW this job already used this pattern; if the 'old' pattern is preferable this one should probably be adjusted to match.

mmatur:
Hi @jnoordsij 

Thanks for your contribution. 

Could you please rebase your PR on v2.11

jnoordsij:
@mmatur done! FWIW some of the changes for the `v3.6` branch have been dropped due to files not existing on `v2.11` (`.github/workflows/test-gateway-api-conformance.yaml` and `.github/workflows/test-knative-conformance.yaml`) and should be 'redone' on merging `v2.11` to `v3.6`.

.github/workflows/build.yaml
```diff
@@ -55,12 +54,12 @@ jobs:
...
           fetch-depth: 0
 
-      - name: Set up Go ${{ env.GO_VERSION }}
+      - name: Set up Go
         uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
         env:
           ImageOS: ${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.goarm }}
         with:
-          go-version: ${{ env.GO_VERSION }}
+          go-version-file: 'go.mod'
```
thaJeztah, line 62: FWIW, this action now supports a separate file for the go-version; see, for example https://github.com/docker/compose/pull/13418


This allows updating the go version without _enforcing_ the minimum (patch) version to be updated by projects using this module as a dependency.

Go modules, including the version of go defined in `go.mod` are designed around MVS (minimum version selection) and should be kept at the lowest version that can be used, not the "highest" version;

- https://github.com/docker/compose/pull/13418#discussion_r2588179894

jnoordsij: That would definitely make sense to use potentially. Although the `go.mod` version bumping was already happening prior to this PR, it still would be neater to distinguish the Go version used by the build vs. the minimal supported version. Although I'm not sure on how much of 'use this as a module' is formally supported, it [clearly is being used at such](https://github.com/search?q=github.com%2Ftraefik%2Ftraefik+path%3A**%2Fgo.mod&type=code) in some places at least.

It would probably require some thought as well on supporting 'multiple' Go versions, given that currently of course all CI tests etc. only run on the 'single' supported minor.

thaJeztah: It was really a peanut-gallery comment from my side, because the PR title caught my eye; was already happy to see the patch version being removed from `go.mod`.

Yes, testing "old" versions is always the tricky bit; Go's recommendation usually is "latest" and "latest-1" (`stable`, `oldstable`), but for some modules _intended_ as library, we try to keep it really low, but of course for any project "within reason".

PR #12656, opened 2026-02-11
Remove :ro from docker.sock

bluepuma77:
Documentation: Remove `:ro` from `docker.sock`, as it creates a false sense of security. The socket is always read-write.


rtribotte:
Hey @bluepuma77,

Thanks for this!

We've set this to "review," and we want to ensure that there are no nuances that would make the read-only mount relevant.

We'll update you once the review is done, but we can already say that the read only socket mount is documented in other places, could you address those to make the PR consistent?

mloiseleur:
It makes sense to me.

Since it's a socket file, _write_ permission is required to [connect](https://www.man7.org/linux/man-pages/man2/connect.2.html) (see EPERM). 
=> All calls sent to the Docker API through this socket require this _write_ permission.

With or without `:ro`, it's possible to remove it from the container, easily tested like this:

```yaml
services:
  debian:
    image: debian
    restart: unless-stopped
    security_opt:
      - no-new-privileges:true

    volumes:

    command:
      - sh
      - -c
      - ls -lh /var/run/docker.sock
      - rm -f /var/run/docker.sock
      - ls -lh /var/run/docker.sock
```

In both cases, it did not delete the socket from the host, only inside the container.

kevinpollet (approved):

rtribotte (approved):
Thanks!

Issue #12631, opened 2026-02-06
Implement more NGINX Ingress Annotations


nmengin:
Hey there,

We're looking for contributors to help implement more and more **NGINX Ingress annotations**.

If you're interested in contributing, this is a great way to get started!

The idea is simple: **pick one or more annotations you'd like to implement**, leave a comment on this issue, and we'll assign them to you. Once your PR is merged, we'll check the box to track progress.

> Tip: Start with **Basic** annotations if you're new, then move on to **Medium** or **Complex** ones as you get familiar.
> 

---

## Basic

- [x]  [nginx.ingress.kubernetes.io/default-backend](http://nginx.ingress.kubernetes.io/default-backend)
- [ ]  [nginx.ingress.kubernetes.io/enable-access-log](http://nginx.ingress.kubernetes.io/enable-access-log) @ris-tlp 
- [x]  [nginx.ingress.kubernetes.io/limit-rps](http://nginx.ingress.kubernetes.io/limit-rps) @amazon7737 
- [x]  [nginx.ingress.kubernetes.io/server-alias](http://nginx.ingress.kubernetes.io/server-alias) @amazon7737 
- [x]  [nginx.ingress.kubernetes.io/limit-rpm](http://nginx.ingress.kubernetes.io/limit-rpm) @Ph4rell
- [x]  [nginx.ingress.kubernetes.io/proxy-http-version](http://nginx.ingress.kubernetes.io/proxy-http-version) @KshitijBharde 
- [x]  [nginx.ingress.kubernetes.io/upstream-keepalive-timeout](http://nginx.ingress.kubernetes.io/upstream-keepalive-timeout)
- [ ]  [nginx.ingress.kubernetes.io/proxy-redirect-from](http://nginx.ingress.kubernetes.io/proxy-redirect-from) @sagarrvala
- [x]  [nginx.ingress.kubernetes.io/allowlist-source-range](http://nginx.ingress.kubernetes.io/allowlist-source-range) @ris-tlp 

---

## Medium

- [ ]  [nginx.ingress.kubernetes.io/load-balance](http://nginx.ingress.kubernetes.io/load-balance) @KshitijBharde 
- [x]  [nginx.ingress.kubernetes.io/x-forwarded-prefix](http://nginx.ingress.kubernetes.io/x-forwarded-prefix) @nandorKollar 
[...]

PR #12659, opened 2026-02-12
Support nginx.ingress.kubernetes.io/allowlist-source-range

ris-tlp:
### What does this PR do?

- Completes part of #12631. 
- Implemented `nginx.ingress.kubernetes.io/allowlist-source-range` annotation
- Configured `IPAllowList` middleware similar to how the `whitelist-source-range` annotation implements it

### Motivation

#12631 

### More

- [x] Added/updated tests

### Additional Notes

The `applyAllowlistSourceRangeConfiguration` and `applyWhitelistSourceRangeConfiguration` functions are identical with the only difference being the `middlewareName` configured according to the annotation name. I've kept the functions separate for now but if needed, I can extract out the common parts into a helper and reuse it across `applyAllowlistSourceRangeConfiguration` and `applyWhitelistSourceRangeConfiguration` if preferred.

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
@@ -860,9 +860,7 @@ func (p *Provider) applyMiddlewares(namespace, routerKey, rulePath, ruleHost str
 		return fmt.Errorf("applying forward auth configuration: %w", err)
 	}
 
-	applyWhitelistSourceRangeConfiguration(routerKey, ingressConfig, rt, conf)
-
-	applyAllowlistSourceRangeConfiguration(routerKey, ingressConfig, rt, conf)
+	applyAllowedSourceRangeConfiguration(routerKey, ingressConfig, rt, conf)
 
 	applyCORSConfiguration(routerKey, ingressConfig, rt, conf)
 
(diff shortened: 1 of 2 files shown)


rtribotte:
Hi @ris-tlp,

Thanks for this contribution!

> The applyAllowlistSourceRangeConfiguration and applyWhitelistSourceRangeConfiguration functions are identical with the only difference being the middlewareName configured according to the annotation name. I've kept the functions separate for now but if needed, I can extract out the common parts into a helper and reuse it across applyAllowlistSourceRangeConfiguration and applyWhitelistSourceRangeConfiguration if preferred.

Indeed, we would rather just have an alias mechanism. We could parse the two annotations into two variables that would setup the same middleware. Could you rework the PR in that direction? Or do you prefer that we handle this in a review commit ?

ris-tlp:
Hi @rtribotte, I've reworked the PR so that both the Allowlist and Whitelist annotations create the same middleware: `routerName + "-allowed-source-range"`, and I've used `allowed` as an umbrella term for both those annotations. Let me know if this matches what you had in mind, thanks!

pepordev:
is this available already on the v3.6.9? i do not see it in the changelog. Thanks for the effort!

nmengin:
Hey @pepordev,

This Pr is an enhancement for the next minor version (v3.7.0).
The RC will be released shortly, so stay tuned 😄

PR #12664, opened 2026-02-12
Add K8s config settings (QPS and Burst).

Robert-Stam:
This allows to override the defaults

```
    // QPS indicates the maximum QPS to the master from this client.
	// If it's zero, the created RESTClient will use DefaultQPS: 5
	//
	// Setting this to a negative value will disable client-side ratelimiting
	// unless `Ratelimiter` is also set.
	QPS float32

	// Maximum burst for throttle.
	// If it's zero, the created RESTClient will use DefaultBurst: 10.
	Burst int
```

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/crd/client.go
+++ b/pkg/provider/kubernetes/crd/client.go
@@ -105,7 +105,7 @@ func newClientImpl(csKube kclientset.Interface, csCrd traefikclientset.Interface
 
 // newInClusterClient returns a new Provider client that is expected to run
 // inside the cluster.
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hey @Robert-Stam,

Thanks for this!

Robert-Stam:
The PR should have been pushed in a fork and was for an internal experiment only, which worked, but didn't gave the desired performance boost (with 6k ingress-routes).

We deleted the fork, which closed this PR.

PR #12679, opened 2026-02-13
Correct encoded characters allowance in entrypoints.md

Apflkuacha:
### What does this PR do?

After back an forth in the last versions they are allowed by default again?!

### Motivation

It confused me what the default is now. But a few lines above it is written: By default, Traefik do not reject requests with path containing certain encoded characters

### More

- [ ] Added/updated tests


gndz07 (approved):
LGTM, thanks 👍

mmatur (approved):

PR #12684, opened 2026-02-14
Fix HasSecureHeadersDefined returning false when stsSeconds is 0

### LINKED ISSUE #11839: Cannot set stsSeconds=0 (HasSecureHeadersDefined() logic bug)
### Welcome!

### What did you do?

* i want to have "strict-transport-security" response header removed or alternatively set to "max-age=0". 
* As the documentation states (https://doc.traefik.io/traefik/middlewares/http/headers/#adding-headers-to-the-request-and-the-response):
 stsSeconds
 The stsSeconds is the max-age of the Strict-Transport-Security header. If set to 0, the header is not set.

when i set `traefik.http.middlewares.no-hsts.headers.stsSeconds=0` in config, i am getting:
 "router.go:135 > error="headers configuration not valid"

which i think is caused by 

https://github.com/traefik/traefik/blob/b39ee8ede5b64e7e1c6e78c276b79d3dd13e7a34/pkg/config/dynamic/middlewares.go#L399 and
https://github.com/traefik/traefik/blob/b39ee8ede5b64e7e1c6e78c276b79d3dd13e7a34/pkg/middlewares/headers/headers.go#L34

```
...
	if !hasSecureHeaders && !hasCustomHeaders && !hasCorsHeaders {
		return nil, errors.New("headers configuration not valid")
	}
...
```

hasSecureHeaders will be true only if STSSeconds  is != 0

so if STSSeconds = 0 then `cfg.HasSecureHeadersDefined()` is returning false and finally the summary condition of
`if !hasSecureHeaders && !hasCustomHeaders && !hasCorsHeaders` will trigger and we get the error.
if i set 
`traefik.http.middlewares.no-hsts.headers.customresponseheaders.strict-transport-security=max-age=0`
then the header is set as wanted.

### What did you see instead?

"router.go:135 > error="headers configuration not valid"

3.4.1

```yaml
 # HSTS middleware
      #- "traefik.http.middlewares.no-hsts.headers.stsSeconds=0"
      #- "traefik.http.middlewares.no-hsts.headers.stsPreload=false"
      - "traefik.http.middlewares.no-hsts.headers.customresponseheaders.strict-transport-security=max-age=0"
```

traefik         | 2025-06-16T18:12:59Z ERR github.com/traefik/traefik/v3/pkg/server/router/router.go:135 > error="headers configuration not valid" entryPointName=websecure routerName=trilium@docker

veeceey:
Fixes #11839

When configuring `stsSeconds=0` as the only header middleware option, traefik rejects the config with `"headers configuration not valid"`. This happens because `HasSecureHeadersDefined()` checks `h.STSSeconds != 0`, which is false for zero — making the middleware think nothing was configured.

The root cause is that `STSSeconds` is `int64`, so we can't distinguish between "not set" and "explicitly set to 0".

**Fix**: Change `STSSeconds` from `int64` to `*int64`. Now `nil` means "not configured" and `0` means "user explicitly set it to zero". The check in `HasSecureHeadersDefined()` becomes `h.STSSeconds != nil`.

Steps to reproduce:
```yaml
http:
  middlewares:
    no-hsts:
      headers:
        stsSeconds: 0
```

Before this fix: `error="headers configuration not valid"`
After this fix: middleware is created successfully

Changes:
- `pkg/config/dynamic/middlewares.go`: field type `int64` -> `*int64`, check `!= 0` -> `!= nil`
- `pkg/config/dynamic/zz_generated.deepcopy.go`: handle pointer deep copy
- `pkg/middlewares/headers/secure.go`: dereference pointer when passing to `unrolled/secure`
- Tests added in `middleware_test.go` and `headers_test.go` covering the stsSeconds=0 case
- Existing tests updated to use pointer values


jspdown:
Hello! Thanks for opening this PR.

Could you open this PR against the branch v3.6 as it's a bug fix?

veeceey:
Done, retargeted to v3.6. Thanks for the heads up!

PR #12691, opened 2026-02-17
Allow entry points to be specified on Nginx Ingresses

### LINKED ISSUE #12646: Be able to limit nginx ingresses to a specific entrypoint(s)
### Welcome!

I have deployed Traefik and added several custom TCP entrypoints (e.g. imap, pop3) to proxy to a TCP backend. I enabled the kubernetesNginxIngress provider, but now all my ingresses are being exposed on my custom TCP entrypoints (as per the Traefik dashboard.) 

I would like the ability to limit the NGINX ingress to a specific entry point to make it easier to migrate to Traefik.

I attempted to add the `traefik.ingress.kubernetes.io/router.entrypoints=web,websecure` annotation to the ingress, however it had no effect until I changed the ingress class to traefik. I'd be happy to send a pull request if this is an acceptable change

ajacques:
### What does this PR do?

Fixes #12646 

Users can now explicitly specify which entry points an NGINX compatible ingress get exposed on. Instead of exposing it on every single entry point. There's no reason for my web server to exposed on my mail ports.

Now users can specify add an annotation, just like the non-NGINX ingresses: `traefik.ingress.kubernetes.io/router.entrypoints: websecure` to define the entry point to use.

### Motivation

I'm migrating from ingress-nginx to Traefik and the KubernetesIngressNginx provider makes this a lot easier. However, I also use Traefik to proxy TCP, non-HTTP based traffic and I noticed that Traefik appeared to expose my web servers over those TCP endpoints:

image

I added support for the same annotation that the non-NGINX ingress uses to make it trivial to then switch to a non-NGINX ingress.

### More

- [X] Added/updated tests

### Additional Notes

I had a few different options for implementing this. I could define it at the provider level (i.e. all NGINX ingresses use entry point X, Y, Z), or I could do a Traefik annotation (`traefik.ingress.kubernetes.io/router.entrypoints: websecure`), or use something that fits with the rest of the NGINX annotations (`nginx.ingress.kubernetes.io/traefik.entrypoints: websecure`). I opted for the second option because it seemed to be the least surprising for a user.

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes.go
+++ b/pkg/provider/kubernetes/crd/kubernetes.go
@@ -1422,10 +1422,9 @@ func buildTLSStores(ctx context.Context, client Client) (map[string]tls.Store, m
 			}
 		}
 
-		if err := buildCertificates(client, id, t.Namespace, t.Spec.Certificates, tlsConfigs); err != nil {
-			logger.Error().Err(err).Msg("Failed to load certificates")
-			continue
-		}
+		// buildCertificates now handles missing secrets gracefully and continues processing
+		// other certificates, so we don't fail the entire TLS store if one certificate is missing
+		buildCertificates(ctx, client, id, t.Namespace, t.Spec.Certificates, tlsConfigs)
 
 		tlsStores[id] = tlsStore
 	}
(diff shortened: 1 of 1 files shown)


kevinpollet:
Hello @ajacques and thanks for opening this,

As said in https://github.com/traefik/traefik/issues/12646#issuecomment-3912980123, we would prefer two new options at the provider level.

Could you adapt this pull request?

ajacques:
> we would prefer two new options at the provider level.

OK with me. I'll find some time to rework the PR in the next couple of days. Thanks for your feedback

PR #12694, opened 2026-02-17
Add maxResponseBodySize configuration to forwardAuth middleware

gndz07:
### What does this PR do?

This PR add an option to configure `maxResponseBodySize` in the `ForwardAuth` middleware.

### Motivation

Currently, the `ForwardAuth` middleware will forward the entirety of the response body without any measure to limit the size. The `maxResponseBodySize` configuration will allow user to define the maximum size of the response body to be forwarded to avoid memory exhaustion.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/migration/v2.md
+++ b/docs/content/migration/v2.md
@@ -764,3 +764,14 @@ in [RFC3986 section-3](https://datatracker.ietf.org/doc/html/rfc3986#section-3).
 
 Please check out the entrypoint [encodedCharacters option](../routing/entrypoints.md#encoded-characters) documentation
 for more details.
+
+## v2.11.38
+
+### `maxResponseBodySize` configuration on ForwardAuth middleware
+
+In `v2.11.38`, a new `maxResponseBodySize` option has been added to the ForwardAuth middleware configuration.
+The default value for this option is -1, which means there is no limit to the response body size.
+However, it is strongly recommended to set this option to a suitable value to avoid performance and security issues,
+such as DoS attacks and memory exhaustion.
+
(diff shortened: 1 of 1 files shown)


pkg/provider/kubernetes/crd/kubernetes.go
```diff
@@ -678,6 +678,10 @@ func createForwardAuthMiddleware(k8sClient Client, namespace string, auth *traef
 		forwardAuth.TLS.Key = authSecretKey
 	}
 
+	if auth.MaxResponseBodySize != nil {
+		forwardAuth.MaxResponseBodySize = auth.MaxResponseBodySize
+	}
+
```
mmatur: It should be placed before 

```go
	if auth.TLS == nil {
		return forwardAuth, nil
	}
```

docs/content/middlewares/http/forwardauth.md
```diff
@@ -637,4 +637,65 @@ http:
...
       insecureSkipVerify: true
 ```
+
+### `maxResponseBodySize`
+
+_Optional, Default=-1_
+
+The `maxResponseBodySize` option defines the maximum allowed response body size in bytes from the authentication server.
+If the response body exceeds the configured limit, the request is rejected with a 401 (Unauthorized) status.
+If left unset, the request body size is unrestricted which can have performance or security implications
```
mmatur:
```suggestion
If left unset, the request body size is unrestricted which can have performance or security implications.
```

mmatur (changes requested):

PR #12697, opened 2026-02-17
Nginx x-forwarded-prefix annotation

nandorKollar:
### What does this PR do?

This PR adds support for `x-forwarded-prefix` Nginx annotation.

### Motivation

Support for `x-forwarded-prefix` Nginx annotation.

### More

- [x] Added/updated tests


pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -945,6 +949,29 @@ func applyRedirect(routerName string, ingressConfig ingressConfig, rt *dynamic.R
 	rt.Middlewares = append(rt.Middlewares, redirectMiddlewareName)
 }
 
+const (
```
nandorKollar: This validation rule is copy-pasted from Nginx, therefore the exact same validation is applied to the value of this annotation in Traefik.

nandorKollar:
There's still two area which requires some clarification:
- It appears to me that Nginx sets this header ONLY when rewrite target is also [used](https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/controller/template/template.go#L780) (the ingress controller doc doesn't mention it), and the path in the ingress rule is [not equals](https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/controller/template/template.go#L776) to the rewrite target. This logic is not implemented in this PR yet.
- Not sure what happens if both `custom-headers` and `x-forwarded-prefix` is present with conflicting target value. Which one wins?

nmengin:
Hey @nandorKollar,

Thanks for this!

NGINX annotations support is a hot topic for us, and we've planned to review it shortly.

gndz07:
Hello @nandorKollar,

Thank you for opening the PR! I agree with your first point, seems to me that this header is only set when `rewrite-target` is configured, even though this condition isn’t explicitly documented.

> Not sure what happens if both custom-headers and x-forwarded-prefix is present with conflicting target value. Which one wins?

My understanding is that there shouldn’t be a conflict between these two annotations. Based on the documentation, [`custom-headers` applies to response headers](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-headers) while [`x-forwarded-prefix` applies to request headers](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-headers).
That said, if I’m misunderstanding the behavior or missing something, please feel free to correct me 🙂

nandorKollar:
> Hello @nandorKollar,
> 
> Thank you for opening the PR! I agree with your first point, seems to me that this header is only set when `rewrite-target` is configured, even though this condition isn’t explicitly documented.
> 
> > Not sure what happens if both custom-headers and x-forwarded-prefix is present with conflicting target value. Which one wins?
> 
> My understanding is that there shouldn’t be a conflict between these two annotations. Based on the documentation, [`custom-headers` applies to response headers](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-headers) while [`x-forwarded-prefix` applies to request headers](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#custom-headers). That said, if I’m misunderstanding the behavior or missing something, please feel free to correct me 🙂

Great, thanks @gndz07 for clarifying, I'm afraid that I overlooked this important detail! After reading the doc, my understanding is the same as yours.

As of the other topic, do you think we should do the same as Nginx?

gndz07:
@nandorKollar yes, we prefer to keep the same behavior as nginx 🙂

nandorKollar:
> @nandorKollar yes, we prefer to keep the same behavior as nginx 🙂

Okay! I'll try to adjust the PR accordingly!

nandorKollar:
@gndz07 it seems that "when rewrite target is also [used](https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/controller/template/template.go#L780)" part was easy to add, however I've some problems with the other precondition, with "the path in the ingress rule is [not equals](https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/controller/template/template.go#L776) to the rewrite target". Could you maybe help with it?  Do you happen to know where can we fetch the path, which we should compare with?

gndz07:
@nandorKollar we had an internal discussion about this annotation, and there are two subjects to discuss with you:
- could you share if you have a particular use case for this annotation?
- we discussed about taking another approach of implementing this annotation, which is by modifying the behavior of the `ReplacePath` middleware that is used to implement `rewrite-target` annotation. We can help with this part. Would you mind if we push some commits directly to your branch? It'll help us move this forward faster instead of going back and forth on small changes.
Let us know if that works for you!

kamilhristov:
@gndz07 Jumping in as we’re migrating to Traefik and waiting on `x-forwarded-prefix`. Our use case is `/$1` (forwarding the first regex path segment to the backend).

```yaml
kind: Ingress
metadata:
  name: ingress-sample-api
  namespace: team-env
  annotations:
    nginx.ingress.kubernetes.io/use-regex: "true"
    nginx.ingress.kubernetes.io/rewrite-target: /$3
    nginx.ingress.kubernetes.io/x-forwarded-prefix: /$1
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
spec:
  ingressClassName: nginx
  rules:
    - host: gateway.example.com
      http:
        paths:
          - path: /(sample-api)(/|$)(.*)
            pathType: ImplementationSpecific
            backend:
              service:
                name: sample-api-service
                port:
                  number: 8080
  tls:
    - hosts:
        - gateway.example.com
      secretName: tls-secret
```

nandorKollar:
> @nandorKollar we had an internal discussion about this annotation, and there are two subjects to discuss with you:
> 
> * could you share if you have a particular use case for this annotation?
> * we discussed about taking another approach of implementing this annotation, which is by modifying the behavior of the `ReplacePath` middleware that is used to implement `rewrite-target` annotation. We can help with this part. Would you mind if we push some commits directly to your branch? It'll help us move this forward faster instead of going back and forth on small changes.
>   Let us know if that works for you!

@gndz07 
* I personally can't tell a use-case for this one, we don't use this annotation. I just implemented this for fun in my spare time to learn something new, and hope someone appreciates that this is working in Traefik. :) What matters for us is `upstream-vhost`, which was required for Traefik to work with Istio.
* Sure no problem, feel free to push changes to my branch. Let me know if you need assistance.

dingyuanli:
@gndz07  Similar to @kamilhristov, my team is also waiting on `x-forwarded-prefix` annotation with a similar usecase while migrating from ingress-nginx. /$1 (forwarding the first regex path segment to the backend). The backend is Spring Boot applications, and the applications api needs to know the full URL to serve static resources. Since the URL prefix is stripped and stored in the `x-forwarded-prefix` header to be extracted to form a full URL, it would be great if Traefik could recognize that annotation and convert it. 

After moving away from ingress-nginx, we probably can use `stripprefix` middlewares to strip and store the prefix into `x-forwarded-prefix` header automatically.

https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/stripprefix/

The support of  `nginx.ingress.kubernetes.io/rewrite-target: /$3` and `nginx.ingress.kubernetes.io/x-forwarded-prefix: /$1` annotations seems to be important, as their priority is higher than ingressroute that I manually created. So, the ingressroute I created with `stripprefix` middlewares cannot take effect, but I cannot remove the existing ingress object before completedly migrated to Traefik

PR #12706, opened 2026-02-19
Add docker provider polling

karaolidis:
### What does this PR do?

This PR enables the use of optional polling on the Docker provider, which can be useful in non-standard setups, e.g. podman (https://community.traefik.io/t/need-to-restart-traefik-whenever-a-converter-is-restarted-even-if-compose-yml-configuration-does-not-change/21166),

### Motivation

I got tired of restarting traefik every time a podman container restarted, and I couldn't figure it out on the podman side :)

### More

- [ ] Added/updated tests

### Additional Notes

If this is something that seems merge-able, I will look into updating docs and tests. Thanks for the great work on traefik!


nmengin:
Hey @karaolidis,

PR #12707, opened 2026-02-20
Support server-alias annotation for Ingress NGINX

amazon7737:
### What does this PR do?
Completes part of Implement more NGINX Ingress Annotations #12631.
Implemented the following NGINX Ingress annotations:

- ~~`nginx.ingress.kubernetes.io/limit-rps`: Maps to Traefik's RateLimit middleware.~~ (Moved to a separate PR for better review alignment with `limit-rpm`)
- ~~`nginx.ingress.kubernetes.io/limit-methods`: Integrates into the router rule using the Method() matcher.~~ (Removed, as it is not an official NGINX annotation)
- `nginx.ingress.kubernetes.io/server-alias`: Supports multiple hostnames by expanding the [Host()](cci:1://file:///Users/amazon/lunch.cancelled/cloud-native-go/traefik/pkg/provider/kubernetes/ingress-nginx/kubernetes.go:1368:0-1375:1) rule.

*(Per the reviewer's request, this PR has been narrowed down to include only `server-alias`.)*

### Motivation
#12631

### More
- [x] Added/updated tests

### Additional Notes
- **Implementation Detail**: 
  - `server-alias`: Multiple aliases are joined with `||` in the [Host()](cci:1://file:///Users/amazon/lunch.cancelled/cloud-native-go/traefik/pkg/provider/kubernetes/ingress-nginx/kubernetes.go:1368:0-1375:1) matcher.


nmengin:
Hey @amazon7737,

Thanks for the PR, great to see more annotation support coming in!
Before we dive into the review, a few things to sort out first:

- **limit-methods**: this annotation doesn't actually exist in the NGINX spec. It was mistakenly included in the related issue based on annotations reported by users through our migration tool. We caught most of them during review, but missed this one — our bad! Could you remove its support from this PR?
- **Split the PR**: once limit-methods is removed, you'd be left with limit-rps and server-alias. It would be really helpful to split those into two separate PRs — one per annotation. It'll make the review cleaner and the changelog easier to manage.
- **Coordinate on rate limiting**: as mentioned in the related issue, another contributor (@Ph4rell) has already implemented limit-rpm. Since limit-rpm and limit-rps are closely related, it would be great if you two could align and build on each other's work to make sure rate limiting is handled consistently. I'm happy to help facilitate that if it would be useful! 😄

Thanks again for jumping on this!

amazon7737:
@nmengin Theanks for the feedback! I have updated this PR to only include the `server-alias` annotation and completely removed `limit-methods`. I will open a separate PR for `limit-rps` so that it can be aligned with @Ph4rell's `limit-rpm` implementation.

docs/content/reference/routing-configuration/kubernetes/ingress-nginx.md
kevinpollet, line 1: Could you please keep the original formatting?

Removing spaces can affect the documentation rendering, and it is harder to review your changes.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1694,7 +1694,21 @@ func basicAuthUsers(secret *corev1.Secret, authSecretType string) (dynamic.Users
 func buildRule(host string, pa netv1.HTTPIngressPath, config ingressConfig) string {
 	var rules []string
 	if len(host) > 0 {
-		rules = append(rules, buildHostRule(host))
+		hosts := []string{host}
+		if config.ServerAlias != nil {
+			hosts = append(hosts, *config.ServerAlias...)
```
kevinpollet: From the `ingress-nginx` [documentation](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#server-alias):

> A server-alias name cannot conflict with the hostname of an existing server. If it does, the server-alias annotation will be ignored. If a server-alias is created and later a new server with the same hostname is created, the new server configuration will take place over the alias configuration.

This means that, before adding the server alias to the host list, we should check whether the alias domain exists in an ingress rule. Perhaps we could reuse the current `hosts` map, which contains all the hosts configured in all the Ingress rules? What do you think?

kevinpollet (changes requested):

docs/content/reference/routing-configuration/kubernetes/ingress-nginx.md
amazon7737: Hello, @kevinpollet, sorry for the formatting mess !
formatting issues caused by my IDE formatter in the previous commit.
I've reverted the documentation to the original master style to fix the spacing and make the review easier.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
amazon7737: Hello, @kevinpollet ,
Thanks for the review! I realized that checking against the `hosts` map instead of just appending is exactly what's needed to follow the spec. I this it's a great way to resolve conflicts while keeping `buildRule` independent.

I'm also considering these additional details:

- Case-insensitive checks : `ToLower` for safer collision detection.
- Warning logs : ignored aliases to help users debug.

While looking at the `hosts` map code, I noticed the current `hosts` map doesn't account for case-sensitivity.
Since DNS hostnames are case-insensitive (e.g. `example.com` vs `EXAMPLE.COM`), I think adding a check using `strings.ToLower()` would be much safer for proper collision detection.
Also, I'm thinking of adding **warning logs** when an alias is ignored so users can understand why it wasn't applied

What do you think about these points? I'll proceed with the implementation once confirmed. Thanks!"

kevinpollet: Hello @amazon7737,

You are right that the comparison should be case-insensitive. Adding a log is also a good idea, but setting the log level to DEBUG should suffice.

Could you adapt the pull request?

We would like to move quickly with this pull request so that it is included in the next release, if you prefer, we can push review commits.

amazon7737: Hello @kevinpollet ,
Thank you for the review! I've updated the PR to address your feedback:

- **Case-insensitive Host Comparison** : I've optimized the host mapping to use lowercase keys, ensuring O(1) performance for case-insensitive comparisons when checking for duplicated aliases. (Performance-friendly approach)
- - **Debug Level Logging** : the log level for skipped server-alias has been set to `DEBUG` as requested.

Please let me know if there are any further improvements needed!

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1691,10 +1691,38 @@ func basicAuthUsers(secret *corev1.Secret, authSecretType string) (dynamic.Users
...
+						aliasAlreadyExists = true
+						break
+					}
+				}
+
+				if !aliasAlreadyExists {
+					hosts = append(hosts, alias)
+				} else {
+					log.Debug().Str("alias", alias).Msg("Skipping server-alias because it is already defined as a host")
+				}
```
amazon7737: Added  a `DEBUG` level log to nitify when a `server-alias` is skipped due to duplication. This keeps the logs clean while providing necessary info for debugging.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
kevinpollet: Thanks! Rather than using a loop to iterate over the hosts map, I would change the host indexation to use only lowercase and adapt the code accordingly. I will submit a review commit.

traefiker:

PR #12708, opened 2026-02-20
Support upstream-keepalive-timeout

jcob-sikorski:
### What does this PR do?

This PR introduces support for upstream-keepalive-timeout configuration for the Kubernetes Ingress NGINX provider.

~This PR introduces support for the `nginx.ingress.kubernetes.io/upstream-keepalive-timeout` annotation within the Kubernetes Ingress NGINX provider.~

~It reads the integer value provided in the annotation (in seconds) and maps it directly to Traefik's internal `ServersTransport.ForwardingTimeouts.IdleConnTimeout` configuration for that specific route.~

### Motivation

Related to #12631

~With the impending retirement of the Kubernetes NGINX Ingress Controller, many users are migrating to Traefik. To ensure a seamless migration, it's crucial that they can maintain their existing routing configurations without rewriting their manifests.~

~Previously, Traefik ignored this specific annotation, which prevented users in high-traffic environments from properly tuning their upstream TCP keep-alive connection behavior. This PR bridges that gap.~

### More

- [x] Added/updated tests

### Additional Notes

~- The annotation expects a standard integer value representing seconds, which precisely mirrors the native NGINX Ingress Controller behavior.~ 
~- A new YAML fixture and test case have been added to `kubernetes_test.go` to ensure this translation is strictly enforced and regressions are prevented.~


nmengin:
Hey @jcob-sikorski,

Thanks for this!

kevinpollet:
Hello @jcob-sikorski,

The `nginx.ingress.kubernetes.io/upstream-keepalive-timeout` annotation does not actually exist in the NGINX spec. It was mistakenly included in the related issue based on the annotations reported by users through our migration tool. We caught most of these annotations during review, but missed this one — our mistake!

The `upstream-keepalive-timeout` configuration is a ConfigMap setting, as described here: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#upstream-keepalive-timeout.

Therefore, a new configuration option with a default value of `60s` has to be added at the ingress-nginx provider level. The ServersTransport added to every service must be adapted to take this value into account.

Could you adapt this pull request?

kevinpollet:
Hello @jcob-sikorski,

Can we push review commits to the pull request to move it forward and include it in the next Traefik release?

PR #12714, opened 2026-02-23
Add TLSKeyExchange access log field

jacobsee:
### What does this PR do?

Adds a new `TLSKeyExchange` access log field

TLS 1.3 decouples the key exchange mechanism from the cipher suite. TLS 1.2 cipher suite names encode the key exchange (e.g. `TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256`, see https://pkg.go.dev/crypto/tls#pkg-constants), but TLS 1.3 cipher suites only identify the cipher and hash function (e.g. `TLS_AES_128_GCM_SHA256`), making it impossible to determine the key exchange from the `TLSCipher` log field alone. Go 1.25+ exposes the negotiated key exchange via `tls.ConnectionState.CurveID`, which is now logged as the new `TLSKeyExchange` field (e.g. "X25519", "CurveP256", "X25519MLKEM768").

Additionally, the existing TLS test case demonstrated using `TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256` (a TLS 1.2 cipher) alongside `Version: VersionTLS13`, which is not a valid combination. I've taken the opportunity to correct this to `TLS_AES_128_GCM_SHA256`, a proper TLS 1.3 cipher suite.

### Motivation

My personal motivation was to allow users who wish to audit usage of quantum-safe key exchange to do so... but I'm sure others may have their own reasons for wanting to know key exchange information.

### More

- [x] Added/updated tests

### Additional Notes

Thanks for looking :slightly_smiling_face:


rtribotte:
Hey @jacobsee,

Thanks for this contribution!

PR #12716, opened 2026-02-23
Add new forwarding timeouts in ServersTransport

### LINKED ISSUE #10962: Setting timeout per service/router
### Welcome!

I am hosting a service that needs a higher (or better disabled) read timeout for its requests. Currently I can't find a possibility to set this.
It would be very helpful, so I can host those services alongside with others, that shouldnt be exposed to [CVE-2024-28869](https://github.com/traefik/traefik/security/advisories/GHSA-4vwx-54mw-vqfw) vulnerability. (#10622)

Is there any plan to integrate this feature? NGINX supports this out of the box, I just switched to traefik a few months ago, because I like those comfort features it provides, but feels its lacking some details.

The same issue already existed in Apr. 2020 and got mentioned in the traefik community forum, but never had a reaction:
https://community.traefik.io/t/setting-timeout-per-service-router/5650

Hope its possible to implement that feature in the future.

Have a good one!

LBF38:
### What does this PR do?

This PR adds two new options, `readTimeout` and `writeTimeout`, for configuring the forwarding timeouts in the `ServersTransport`.

### Motivation

To complete #12630

### More

- [x] Added/updated tests

### Additional Notes

Requires #12630 to be merged first.

Closes https://github.com/traefik/traefik/issues/10962


boindil:
@LBF38 thank you very much for your work on this and I would love to see this merged! If I could be of any help, I'd be glad to help out. 

Since there are currently test failing which - to me - seem to require more knowledge of the traefik coding, I guess there's not much I can do? If there is something, please let me know!

Jeroen0494:
Hello, thank you for working on this feature! Is there the possibility of this feature becoming available through Middleware so I can apply it to Kubernetes ingress definitions?

nmengin:
Hey @Jeroen0494,

You can apply serversTransport on a Service using the dedicated annotation as described in [the documentation](https://doc.traefik.io/traefik/reference/routing-configuration/kubernetes/ingress/#on-service).

boindil:
@rtribotte @nmengin any chance this makes it into 3.7.0? Any idea on when this actually might be merged? I think this is a great addition to the available features of traefik! <3

Sorry for being a bit impatient!

rtribotte:
Hi @boindil,

> @rtribotte @nmengin any chance this makes it into 3.7.0? Any idea on when this actually might be merged? I think this is a great addition to the available features of traefik! <3
> Sorry for being a bit impatient!

The v3.7 is currently at release candidate stage, so unfortunately we won't be adding any features to it. However, this should be included in the next version.

gndz07 (approved):
LGTM

boindil:
> Hi @boindil,
> 
> > @rtribotte @nmengin any chance this makes it into 3.7.0? Any idea on when this actually might be merged? I think this is a great addition to the available features of traefik! <3
> > Sorry for being a bit impatient!
> 
> The v3.7 is currently at release candidate stage, so unfortunately we won't be adding any features to it. However, this should be included in the next version.

Hi @rtribotte,

we already got a 3.7.1 and this was sadly not included. I hope you did not mean 3.8 ... Was this PR missed for 3.7.1?

kevinpollet:
Hello @boindil,

> we already got a 3.7.1 and this was sadly not included. I hope you did not mean 3.8 ... Was this PR missed for 3.7.1?

As @rtribotte mentioned, this pull request introduces a new feature, so it should be included in a future minor version, likely v3.8.

bestseller-rsj:
Any update on this one?

pierinho13:
When is planned to be merged? Thanks

lewis-jackson-bots:
I noticed that this sadly missed v3.8, we'd be keen to use it

mmatur (changes requested):
Thanks for the PR @LBF38! 

Before merging, I'd like us to agree on three design points, because they change the default behavior for existing users.

**1. The 30s default is a breaking change.** `SetDefaults()` now sets `readTimeout = writeTimeout = 30s`. Before, these were `0` for everyone except the NGINX provider. The CRD provider calls `SetDefaults()` unconditionally, and paerser calls it for any file/label/KV config with a `forwardingTimeouts:` block. So after upgrade, every named ServersTransport gets a 30s per-read/per-write deadline: idle WebSockets, SSE/long-polling, quiet gRPC/HTTP2 streams, and backends slower than 30s to first byte will start failing with i/o timeout / 502. `serversTransport@internal` is not affected, which makes it hard to diagnose.

**2. The default transport is inconsistent.** Static `forwardingTimeouts` has no `readTimeout`/`writeTimeout`, and `internal.go` doesn't copy them. So the default transport stays unlimited while any named transport gets 30s, with no way to set or disable it globally.

**3. Fast proxy ignores these options.** `pkg/proxy/fast/builder.go` only reads dial/idle/responseHeader timeouts, so with `experimental.fastProxy` the new fields do nothing. Not documented.

**Proposal:** default `readTimeout`/`writeTimeout` to `0` (opt-in, non-breaking). If we really want a default, add the fields to the static config + default transport and write a migration note. And either support them in fast proxy or document that they only apply to the standard proxy.

What do you think?

LBF38:
Hi @mmatur, 
Thanks for your review!

I agree with you. It makes sense to keep the default from previous versions so that we don't break anything.
I've applied your suggestions as follows:
- keep non-breaking defaults (0 for both read/write timeouts)
- make it consistent in the fast proxy and static config

mmatur (approved):

PR #12717, opened 2026-02-23
Fix Go version pinning in CI workflows

mmatur:
### What does this PR do?

- Add a `.go-version` file to pin CI builds to a specific Go patch version (1.25.7)
- Switch all CI workflows from `go-version-file: 'go.mod'` to `go-version-file: '.go-version'`
- Revert `go.mod` to `go 1.25.0`, keeping it as the minimum required version

This decouples the CI Go version from the `go.mod` directive — `go.mod` declares the minimum language version the module needs, while `.go-version` controls exactly which toolchain CI uses.

### Motivation

Traefik is used as a library. (cf: https://github.com/traefik/traefik/pull/12649#discussion_r2841062608)

### More

- ~[ ] Added/updated tests~


go.mod
```diff
@@ -1,11 +1,6 @@
 module github.com/traefik/traefik/v2
 
-// Only specify the major.minor version here.
-// CI workflows read this value via go-version-file and use check-latest
-// to automatically pick up the newest patch release.
-// Pinning a patch version would defeat that mechanism and require manual
-// bumps for every Go patch release.
-go 1.25
+go 1.25.0
```
thaJeztah, line 3: oh! this can probably still be `go 1.25` without a `.patch` specified


edit: patch, not minor

jnoordsij, line 3: Per https://github.com/traefik/traefik/pull/12649#discussion_r2841163010 (and also the discussion on https://github.com/docker/compose/pull/13418): shouldn't it be preferable to keep this minor-only anyways?

Adding a patch just adds additional restrictions for any consuming module (most notably disallowing any prereleases) which are probably not required anyways.

.go-version
```diff
@@ -0,0 +1 @@
+1.25.7
```
jnoordsij, line 1: Do we really want to manually pin the patch here, which would also require updating it manually to stay up-to-date?

go.mod
mmatur: We are using traefik as a library and to avoid issues we still want to keep it for now.

mmatur: We are using traefik as a library and to avoid issues we still want to keep it for now.
We will pass some time to be sure that it does not impact our commercial product.

.go-version
mmatur: For now yes to avoid this type of [PR](https://github.com/traefik/traefik/pull/12712)

PR #12724, opened 2026-02-23
Fix build with GOPROXY=direct by replacing nhooyr.io/websocket

### LINKED ISSUE #12705: Traefik is not buildable from source with GOPROXY=direct
### Welcome!

### What did you do?

Tried to build traefik 3.6.8 from source on a box that doesn't have any previously downloaded go modules cached.

The build fails when downloading the go dependencies because several versions listed in go.mod are no longer available in the locations traefik expects to find them - the direct dependency nhooyr.io/websocket has moved to github.com/coder/websocket

### What did you see instead?

Errors while fetching dependencies.
It works with GOPROXY=https://proxy.golang.org/ because the proxy still has the old location's files stored.

Also, the new location has never versions that have potentially relevant fixes. It may make sense to update to 1.8.14 (and potentially update some other dependencies too)

3.6.8

OpenMandriva Cooker install with go 1.26.0

_No response_

mehrdadbn9:
## What does this PR do?

This PR fixes the build issue when using `GOPROXY=direct` by adding a `replace` directive to redirect `nhooyr.io/websocket` to `github.com/coder/websocket`.

## Motivation

The `nhooyr.io/websocket` module has been archived and moved to `github.com/coder/websocket`. When building with `GOPROXY=direct` (without using the Go module proxy), the build fails because the old module path is no longer accessible.

Fixes #12705

## More

- [x] Added/updated tests

### How to test

1. Clone the repository
2. Run `GOPROXY=direct go build ./cmd/traefik`
3. Verify the build succeeds

Before this fix:
```
go: nhooyr.io/websocket@v1.8.7: reading nhooyr.io/websocket/go.mod at revision v1.8.7: unknown revision v1.8.7
```

After this fix:
Build succeeds with `GOPROXY=direct`.

### Additional Notes

- The `github.com/coder/websocket` fork is actively maintained by Coder
- Version v1.8.7 is used to maintain API compatibility with the original `nhooyr.io/websocket v1.8.7` that traefik's `grpc-web` dependency requires
- All websocket-related unit tests pass

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -425,6 +425,7 @@ replace github.com/traefik/traefik/dynamic/ext => ./pkg/config/dynamic/ext
 replace (
 	github.com/abbot/go-http-auth => github.com/containous/go-http-auth v0.4.1-0.20260804094822-b975dcaa8c48
 	github.com/gorilla/mux => github.com/containous/mux v0.0.0-20250523120546-41b6ec3aed59
+	nhooyr.io/websocket => github.com/coder/websocket v1.8.7
 )
 
 // ambiguous import: found package github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/common/http in multiple modules
--- a/go.sum
+++ b/go.sum
@@ -861,6 +861,8 @@ github.com/cncf/xds/go v0.0.0-20211011173535-cb28da3451f1/go.mod h1:eXthEFrGJvWH
 github.com/cncf/xds/go v0.0.0-20220314180256-7f1daf1720fc/go.mod h1:eXthEFrGJvWHgFFCl3hGmgk+/aYT6PnTQLykKQRLhEs=
 github.com/cncf/xds/go v0.0.0-20230105202645-06c439db220b/go.mod h1:eXthEFrGJvWHgFFCl3hGmgk+/aYT6PnTQLykKQRLhEs=
 github.com/cncf/xds/go v0.0.0-20230310173818-32f1caf87195/go.mod h1:eXthEFrGJvWHgFFCl3hGmgk+/aYT6PnTQLykKQRLhEs=
+github.com/coder/websocket v1.8.7 h1:jiep6gmlfP/yq2w1gBoubJEXL9gf8x3bp6lzzX8nJxE=
+github.com/coder/websocket v1.8.7/go.mod h1:B70DZP8IakI65RVQ51MsWP/8jndNma26DVA/nFSCgW0=
 github.com/containerd/errdefs v1.0.0 h1:tg5yIfIlQIrxYtu9ajqY42W3lpS19XqdxRQeEwYG8PI=
 github.com/containerd/errdefs v1.0.0/go.mod h1:+YBYIdtsnF4Iw6nWZhJcqGSg/dwvV7tyJ/kCkyJ2k+M=
 github.com/containerd/errdefs/pkg v0.3.0 h1:9IKJ06FvyNlexW690DXuQNx2KA2cUJXx151Xdx3ZPPE=
@@ -3106,8 +3108,6 @@ modernc.org/token v1.0.0/go.mod h1:UGzOrNV1mAFSEB63lOFHIpNRUVMvYTc6yu1SMY/XTDM=
 modernc.org/z v1.5.1/go.mod h1:eWFB510QWW5Th9YGZT81s+LwvaAs3Q2yr4sP0rmLkv8=
 mvdan.cc/xurls/v2 v2.5.0 h1:lyBNOm8Wo71UknhUs4QTFUNNMyxy2JEIaKKo0RWOh+8=
 mvdan.cc/xurls/v2 v2.5.0/go.mod h1:yQgaGQ1rFtJUzkmKiHYSSfuQxqfYmd//X6PxvholpeE=
-nhooyr.io/websocket v1.8.7 h1:usjR2uOr/zjjkVMy0lW+PPohFok7PCow5sDjLgX4P4g=
-nhooyr.io/websocket v1.8.7/go.mod h1:B70DZP8IakI65RVQ51MsWP/8jndNma26DVA/nFSCgW0=
 pgregory.net/rapid v1.2.0 h1:keKAYRcjm+e1F0oAuU5F5+YPAWcyxNNRK2wud503Gnk=
 pgregory.net/rapid v1.2.0/go.mod h1:PY5XlDGj0+V1FCq0o192FdRhpKHGTRIWBgqjDBTrq04=
 rsc.io/binaryregexp v0.2.0/go.mod h1:qTv7/COck+e2FymRvadv62gMdZztPaShugOCi3I+8D8=


go.mod
```diff
@@ -420,6 +420,7 @@ replace (
 	github.com/gorilla/mux => github.com/containous/mux v0.0.0-20250523120546-41b6ec3aed59
 	github.com/mailgun/minheap => github.com/containous/minheap v0.0.0-20190809180810-6e71eb837595
 	github.com/vulcand/oxy/v2 => github.com/traefik/oxy/v2 v2.0.0-20260126093803-fb11d60e0fdf
+	nhooyr.io/websocket => github.com/coder/websocket v1.8.7
```
thaJeztah: From https://github.com/nhooyr/websocket-old/commit/f29be9fb88edf3dd51afba3b277e7643b1628c74 it looks like his is a permanent move, so may be better to update all import paths instead of using a `replace`.

thaJeztah: Oh! Looks like it's an indirect dependency; so needs update in `traefik/grpc-web`?

```bash
go mod graph | grep ' nhooyr.io/websocket'
github.com/traefik/traefik/v3 nhooyr.io/websocket@v1.8.7
github.com/traefik/grpc-web@v0.16.0 nhooyr.io/websocket@v1.8.7
```

https://github.com/traefik/grpc-web/blob/b664acc3fb778af638715e30358fb70262a1a916/go.mod#L21

nmengin:
Hey @mehrdadbn9,

Thanks for this!
We've set this to "review" and will review the code.
We'll make sure to update you once the review is done.

mehrdadbn9:
Rebased onto current master and resolved the go.mod/go.sum conflict. The fix is still needed: master still pins `nhooyr.io/websocket v1.8.7` as an indirect dependency, which is archived and not fetchable via `GOPROXY=direct`, breaking offline/air-gapped builds.

Verified: `GOPROXY=direct go build ./cmd/traefik` now succeeds with the `replace nhooyr.io/websocket => github.com/coder/websocket v1.8.7` directive in place.

amazon7737:
FYI: This fixes the websocket download issue, but I encountered two other checksum mismatches when building with `GOPROXY=direct` and an empty module cache: `github.com/liquidweb/liquidweb-cli@v0.7.0` and the `go.mod` for `software.sslmate.com/src/go-pkcs12@v0.7.3`. Both also occur on the base branch. These need further investigation to determine whether they require dependency updates in Traefik or fixes upstream.

thaJeztah:
I would recommend fixing it in `github.com/traefik/grpc-web`. It's a ticking time-bomb, because once the cache in Google's module proxy expires, both this module, and `github.com/traefik/grpc-web` will fail to build. The `replace` rule also is non-transitive, so any user of this module will still attempt to use the `nhooyr.io/websocket` module, and fail.

I see there's a similar PR in `github.com/traefik/grpc-web`, but it also attempts to fix it with a replace rule;

- https://github.com/traefik/grpc-web/pull/2

That PR should be updated (or replaced with another PR) to update the imports to the correct name (`github.com/coder/websocket`).

rtribotte (approved):
Thanks!

go.mod
```diff
@@ -423,6 +420,14 @@ replace (
 	github.com/gorilla/mux => github.com/containous/mux v0.0.0-20250523120546-41b6ec3aed59
 )
 
+// The upstream v0.7.0 tag was moved after publication, so it no longer matches the checksum database and GOPROXY=direct builds fail.
+// The pinned commit precedes the retag and leaves the imported packages byte-identical to the notarized v0.7.0.
+replace github.com/liquidweb/liquidweb-cli => github.com/liquidweb/liquidweb-cli v0.6.11-0.20231013174031-ea0acb475a1d
+
+// The canonical software.sslmate.com host serves the repository over dumb HTTP with a stale objects/info/packs, so GOPROXY=direct cannot fetch the objects behind the v0.7.3 tag.
+// The SSLMate GitHub mirror carries the same commit and the same go.mod bytes.
+replace software.sslmate.com/src/go-pkcs12 => github.com/SSLMate/go-pkcs12 v0.7.3
```
thaJeztah, line 429: Has this been reported upstream? This project looks to be still actively maintained.

rtribotte: @thaJeztah Not yet. But I will open an issue to report the issue we had.

go.mod
```diff
@@ -423,6 +420,14 @@ replace (
 	github.com/gorilla/mux => github.com/containous/mux v0.0.0-20250523120546-41b6ec3aed59
 )
 
+// The upstream v0.7.0 tag was moved after publication, so it no longer matches the checksum database and GOPROXY=direct builds fail.
+// The pinned commit precedes the retag and leaves the imported packages byte-identical to the notarized v0.7.0.
+replace github.com/liquidweb/liquidweb-cli => github.com/liquidweb/liquidweb-cli v0.6.11-0.20231013174031-ea0acb475a1d
```
thaJeztah, line 425: This is concerning; looks indeed like an attempt to "fix up" a tag, but that doesn't really work with go modules; https://github.com/liquidweb/liquidweb-cli/compare/ea0acb475a1d...v0.7.0

Ideally, they'd tag a v0.7.1 and `retract` v0.7.0 as faulty release; project doesn't look actively maintained though, but https://www.liquidweb.com/ is still up.

cc @Eric-Hegel @desmcdermitt looks like you are both active within the @liquidweb org; do you know who to reach out to to fix that repository?

kevinpollet (approved):
Thanks 👍

go.mod
rtribotte: @thaJeztah https://github.com/SSLMate/go-pkcs12/issues/84

go.mod
ldez: As this is a dependency related to lego, I will handle the problem.

https://github.com/liquidweb/liquidweb-cli/issues/90

ldez: `github.com/liquidweb/liquidweb-go` only uses `github.com/liquidweb/liquidweb-cli` for one structure `apiTypes.Subaccnt`: https://github.com/liquidweb/liquidweb-go/blob/v1.6.4/asset/asset_client.go

This structure has no dependencies, I can safely downgrade the version to v0.6.10 inside lego.

Note: the package `apiTypes` doesn't have the same name as the directory and contains uppercase letter, which are a bad practices.

go.mod
AGWA: @thaJeztah thanks for asking that this be reported upstream. As explained [here](https://github.com/SSLMate/go-pkcs12/issues/84#issuecomment-5785885384), the diagnosis that objects/info/packs is stale is incorrect, and this is likely an issue on @rtribotte's end. No guarantees that the GitHub mirror will always exist.

thaJeztah: Thanks @AGWA - I saw the discussion on that ticket; good to hear it's not an issue (the `replace` added in this PR caught my eye, and I was looking at ways to avoid them, because they should only be needed in exceptional cases).

thaJeztah: Thanks @AGWA - I saw the discussion on that ticket; good to hear it's not an issue (the `replace` added in this PR caught my eye, and I was looking at ways to avoid them, because they should only be needed in exceptional cases).

go.mod
thaJeztah: Thx @ldez ! Looks like it's then (probably?) possible to remove the replace rules again with that change 👍

PR #12725, opened 2026-02-23
Add RFC 7239 Forwarded header support

### LINKED ISSUE #4099: Support "Forwarded" header instead "X-Forwarded-For"

Feature

RFC 7239 now defines the ```Forwarded:``` header, which is intended to replace ```X-Forwarded-*```. If you care about standards, I would recommend using that instead.

At that moment Traefik support ```X-Forwarded-For```, ```X-Forwarded-Host```, ```X-Forwarded-Port```, ```X-Forwarded-Proto``` and ```X-Forwarded-Server```, but not follow RFC 7239 for share that information.

First of all, I wanted to point out that it is the RFC that exists.

sammiee5311:
### What does this PR do?

Adds RFC 7239 `Forwarded` header support to the `forwardedheaders` middleware.

- Adds `Forwarded` to the `xHeaders` list so it gets stripped from untrusted clients (same as `X-Forwarded-For` and other headers)
- Generates a proper RFC 7239 `Forwarded` header in `rewrite()` using the client IP, protocol, and host
- Handles IPv6 addresses with quoting and bracketing per RFC 7239 Section 6
- Appends to existing trusted `Forwarded` values for proper proxy chaining
- Validates `proto` against known values and escapes quotes in `host` to prevent header injection

### Motivation

Closes #4099

The `Forwarded` header (RFC 7239) is the standardized replacement for `X-Forwarded-For`/`X-Forwarded-Proto`/`X-Forwarded-Host`. Currently, Traefik strips `X-Forwarded-*` headers from untrusted clients but doesn't touch `Forwarded`, meaning it gets passed through to backends unsanitized. Several users raised this concern in #4099.

This PR adds both the sanitization (stripping from untrusted clients) and the generation (building the header from real client info), consistent with how `X-Forwarded-*` headers are handled.

### More

- [x] Added/updated tests

### Additional Notes

Tests cover: IPv4, IPv6, trusted chaining (appending), untrusted stripping, and host header injection prevention. All existing tests pass.

The `Forwarded` header is generated alongside the existing `X-Forwarded-*` headers, not as a replacement, so nothing breaks for backends that still read `X-Forwarded-For`.


rtribotte:
Hey @sammiee5311,

Issue #12673, opened 2026-02-13
[Doc] Compress middleware: missing SSE stream behaviour in 3.X


Chostakovitch:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

PR #11583 solved an issue with SSE streams compression, which enables users (like me) to remove `text/event-stream` from `excludedContentTypes`.

However I was puzzled not to find this information in the documentation (whether 2.X or 3.X online docs), whereas the PR did documented it.

>  When data is sent to the client on flush, the `minResponseBodyBytes` configuration is ignored and the data is compressed. This is particularly the case when data is streamed to the client when using `Transfer-encoding: chunked` response.

I am not sure to understand why `docs/content`middlewares and `docs/content/reference/routing-configuration/http/middlewares` both exist and still receive updates; anyway, the former has the information from 474ab23fe920243d2b7def642c4be53656178074, but the later was created by b7170df2c3a37caaf8e68de5919a2013d9c7b928 just a few days later, so maybe they overlapped.

Sorry for being lengthy for such a small detail; it will also help me understand how docs are organized in newer version ☺️

I'll submit a mini PR to fix that if I understood it right.

### What did you see instead?

https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/compress/

No mention of bypassing compressing settings on flush.

### What version of Traefik are you using?

3.6

### What is your environment & configuration?

Not relevant.

prasadf18:
hey @nmengin  can pls you assign this to me

nmengin:
Hey @prasad-firame, thank you for the proposition.
I've just assigned you. 👍

prasadf18:
Hi @nmengin,
I’ve raised a PR. Could you please review it at your convenience?
Thank you. I’m waiting for your feedback.

Chostakovitch:
@nmengin  @prasad-firame Is there anything wrong with the PR I submitted when I created the issue i.e. #12674? Is it supposedly under review atm.

Also I probably know the code way less that you, but I feel you could double check the part in #12728 where you state:

> To avoid breaking real-time streams, it is recommended to exclude SSE content types from compression: [...] Without excluding text/event-stream, compressed SSE streams may not work correctly in clients.

[...]

PR #11583, opened 2025-03-05
Compress data on flush when compression is not started

kevinpollet:
### What does this PR do?

This pull request fixes the compress middleware to compress the data by default in streaming mode even if we haven't received enough data to decide if we should compress (see https://doc.traefik.io/traefik/middlewares/http/compress/#minresponsebodybytes config option) or not.

### Motivation

Fixes #11509

### More

- [X] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

Co-authored-by: Romain <rtribotte@users.noreply.github.com>

PR #12674, opened 2026-02-13
Add a note about bypassing of minResponseBodyBytes in streaming mode

Chostakovitch:
Fixes #12673.

Feel free to check/reword, I am not 100% sure of understanding correctly. I tried to make to original wording a bit more explanatory.

PR #12728, opened 2026-02-25
Clarify SSE behavior with compress middleware

prasadf18:
### What does this PR do?

Clarifies the behavior of Server-Sent Events (SSE) in the Compress middleware for Traefik 3.X.  
Documents that `minResponseBodyBytes` is bypassed for SSE and chunked responses, and shows how to exclude `text/event-stream` from compression.

### Motivation

Without this clarification, users may experience broken real-time streams when SSE is compressed.  
Aligns documentation with the fix introduced in PR #11583.

### More

- [ ] Added/updated tests

### Additional Notes

This change applies to Traefik v3.6.  
Fixes issue #12673.


rtribotte:
Hey @prasad-firame,

mmatur:
Thanks for the contribution! Closing as a duplicate of #12674, which documents the same behavior with a much smaller diff. Let's continue the discussion there.

PR #12729, opened 2026-02-25
Fix basePath validation for dashboard template

gndz07:
### What does this PR do?

Replace the imported `text/template` with `html/template` for the dashboard template and add validation for the `basePath` configuration.

### Motivation

A stricter regex validation will allow user to be warned early if the URL in their `basePath` is not a valid URL.
`text/template` provides no automatic escaping for HTML or JavaScript contexts. Switching to `html/template` ensures values rendered into the dashboard template are properly escaped, preventing potential injection issues.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/config/static/static_config.go
+++ b/pkg/config/static/static_config.go
@@ -3,7 +3,6 @@ package static
 import (
 	"errors"
 	"fmt"
-	"path"
 	"regexp"
 	"strings"
 	"time"
@@ -59,7 +58,7 @@ const (
 )
 
 // Allowed characters in URL following RFC 3986 (https://www.rfc-editor.org/rfc/rfc3986#section-2)
-var validURLRegex = regexp.MustCompile(`^[a-zA-Z0-9\-._~!$&'()*+,;=:@/%]+$`)
+var validBasePath = regexp.MustCompile(`^/[a-zA-Z0-9/_.-]*$`)
 
 // Configuration is the static configuration.
 type Configuration struct {
@@ -468,14 +467,8 @@ func (c *Configuration) ValidateConfiguration() error {
 		}
 	}
 
-	if c.API != nil {
-		if !path.IsAbs(c.API.BasePath) {
-			return errors.New("API basePath must be a valid absolute path")
-		}
-
-		if !validURLRegex.MatchString(c.API.BasePath) {
-			return errors.New("API basePath must be a valid URL path")
-		}
+	if c.API != nil && !validBasePath.MatchString(c.API.BasePath) {
+		return errors.New("API basePath must be a valid absolute URL path")
 	}
 
 	if c.OCSP != nil {
--- a/pkg/config/static/static_config.go
+++ b/pkg/config/static/static_config.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


pkg/config/static/static_config.go
```diff
@@ -57,6 +58,9 @@ const (
 	DefaultUDPTimeout = 3 * time.Second
 )
 
+// Allowed characters in URL following RFC 3986 (https://www.rfc-editor.org/rfc/rfc3986#section-2)
+var validURLRegex = regexp.MustCompile(`^[a-zA-Z0-9\-._~!$&'()*+,;=:@/%]+$`)
```
mmatur:
```suggestion
var validBasePath = regexp.MustCompile(`^/[a-zA-Z0-9/_.-]*$`)
```

pkg/config/static/static_config.go
```diff
@@ -464,8 +468,14 @@ func (c *Configuration) ValidateConfiguration() error {
...
-	if c.API != nil && !path.IsAbs(c.API.BasePath) {
-		return errors.New("API basePath must be a valid absolute path")
+	if c.API != nil {
+		if !path.IsAbs(c.API.BasePath) {
+			return errors.New("API basePath must be a valid absolute path")
+		}
+
+		if !validURLRegex.MatchString(c.API.BasePath) {
+			return errors.New("API basePath must be a valid URL path")
+		}
```
mmatur:
```suggestion
	if c.API != nil && !validBasePath.MatchString(c.API.BasePath) {
		return errors.New("API basePath must be a valid absolute URL path")
	}
```

pkg/config/static/static_config_test.go
```diff
@@ -282,3 +282,64 @@ func TestConfiguration_SetEffectiveConfiguration(t *testing.T) {
...
+		{
+			desc:      "path with spaces",
+			basePath:  "/path with spaces",
+			expectErr: true,
+		},
+		{
+			desc:      "path with angle brackets",
+			basePath:  "/path/<evil>",
+			expectErr: true,
+		},
```
mmatur, line 326:
```suggestion
		},
		{
			desc:      "path with query string",
			basePath:  "/api?foo=bar",
			expectErr: true,
		},
		{
			desc:      "path with fragment",
			basePath:  "/api#section",
			expectErr: true,
		},
		{
			desc:      "valid root path",
			basePath:  "/",
			expectErr: false,
		},
		{
			desc:      "path with quote",
			basePath:  "/api/'onclick=alert(1)",
			expectErr: true,
		},
		{
			desc:      "path with encoded character",
			basePath:  "/api%2Ftoto",
			expectErr: true,
		},
```

mmatur (changes requested):

mmatur (approved):
LGTM

rtribotte (approved):
LGTM

pkg/config/static/static_config.go
```diff
@@ -57,6 +57,9 @@ const (
 	DefaultUDPTimeout = 3 * time.Second
 )
 
+// Allowed characters in URL following RFC 3986 (https://www.rfc-editor.org/rfc/rfc3986#section-2)
+var validBasePath = regexp.MustCompile(`^/[a-zA-Z0-9/_.-]*$`)
```
thaJeztah, line 61: Only glancing over, but I noticed `.` in the accepted range; is path traversing allowed? (`/api/../../`, `/api/./foo/`); not sure if those are problematic or handled elsewhere; mostly curious!

gndz07: Hello @thaJeztah,

The `.` character is valid in a URL and may legitimately be used by some users when defining their `basePath`. Additionally, this value can only be set through the static configuration, meaning it is controlled by a Traefik administrator. Given this condition, we think that path traversal is not a concern in this context at the moment 🙂

PR #12730, opened 2026-02-25
Fix nginx rewrite target

mmatur:
### What does this PR do?

This PR fixes the nginx rewrite target implementation

### Motivation

To have the same behaviour as ingress nginx

### More

- [x] Added/updated tests


gndz07 (approved):
LGTM

PR #12721, opened 2026-02-23
Log Kubernetes resource namespace, kind and name

sathieu:
### What does this PR do?

This PR adds the following fields to access logs:

- namespace
- kind
- name

For the following providers:

- [Kubernetes Ingress NGINX](https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-ingress-nginx/)
- [Kubernetes Ingress](https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-ingress/)

If the pattern is good, I will propose PRs for the following providers:

- [Kubernetes IngressRoute](https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-crd/)
- [Kubernetes Gateway API](https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-gateway/)

I will also use those three fields for metrics in a future PR.

### Motivation

We need better information to map access logs to Kubernetes resources.

See also https://github.com/traefik/traefik/issues/10105

### More

- [x] Added/updated tests
- [x] Added/updated documentation

PR #12732, opened 2026-02-25
Use Kubernetes labels in Prometheus metrics

### LINKED ISSUE #6864: Include namespace label in prometheus metrics
This is a **feature** request

Traefik can be configured to expose [a metrics endpoint in the Prometheus format](https://docs.traefik.io/observability/metrics/prometheus/). An end-user is free to choose between:
* an entry point format:
`traefik_entrypoint_request_duration_seconds_bucket{code="404",entrypoint="traefik",method="GET",protocol="http",le="0.1"} 14`
* and a service format:
`traefik_service_request_duration_seconds_bucket{code="200",method="GET",protocol="http",service="whoami-whoami-9ab4060701404e59ffcd@kubernetescrd",le="0.1"} 2`

Typically, those metrics are produced by a Traefik controller according to how [IngressRoute](https://docs.traefik.io/reference/dynamic-configuration/kubernetes-crd/) is defined (how entryPoints and services properties are set). If we have, for example, the following IngressRoute definition:
```
kind: IngressRoute
metadata:
  name: myservice
  namespace: mynamespace
spec:
  entryPoints:
    - web
  routes:
  - match: PathPrefix(`/myservice`)
    kind: Rule
    services:
    - name: myservice
      port: 80
```

Traefik configured with `--metrics.prometheus=true --metrics.prometheus.buckets=0.100000, 0.300000, 1.200000, 5.000000 --metrics.prometheus.addEntryPointsLabels=true --metrics.prometheus.addServicesLabels=true` will produce the following metrics:
```
traefik_entrypoint_request_duration_seconds_bucket{code="200",entrypoint="web",method="GET",protocol="http",le="0.1"} 2
traefik_entrypoint_request_duration_seconds_bucket{code="200",entrypoint="web",method="GET",protocol="http",le="0.3"} 2
traefik_entrypoint_request_duration_seconds_bucket{code="200",entrypoint="web",method="GET",protocol="http",le="1.2"} 2
traefik_entrypoint_request_duration_seconds_bucket{code="200",entrypoint="web",method="GET",protocol="http",le="5"} 2
traefik_entrypoint_request_duration_seconds_bucket{code="200",entrypoint="web",method="GET",protocol="http",le="+Inf"} 2
traefik_entrypoint_request_duration_seconds_sum{code="200",entrypoint="web",method="GET",protocol="http"} 0.002430748
traefik_entrypoint_request_duration_seconds_count{code="200",entrypoint="web",method="GET",protocol="http"} 2

traefik_service_request_duration_seconds_bucket{code="200",method="GET",protocol="http",service="mynamespace-myservice-9ab4060701404e59ffcd@kubernetescrd",le="0.1"} 2
traefik_service_request_duration_seconds_bucket{code="200",method="GET",protocol="http",service="mynamespace-myservice-9ab4060701404e59ffcd@kubernetescrd",le="0.3"} 2
traefik_service_request_duration_seconds_bucket{code="200",method="GET",protocol="http",service="mynamespace-myservice-9ab4060701404e59ffcd@kubernetescrd",le="1.2"} 2
traefik_service_request_duration_seconds_bucket{code="200",method="GET",protocol="http",service="mynamespace-myservice-9ab4060701404e59ffcd@kubernetescrd",le="5"} 2
traefik_service_request_duration_seconds_bucket{code="200",method="GET",protocol="http",service="mynamespace-myservice-9ab4060701404e59ffcd@kubernetescrd",le="+Inf"} 2
```

While having all relevant context data, the output still lacks a clear notion of what exact namespace the related resources (IngressRoute, Service, TraefikService) are defined in.

**Why is namespace so important?**

So far, I've come up with several use cases:
* Grafana dashboards usually contain namespace selectors. Users are willing to distinguish their services between different namespaces nicely.
* Integration: to integrate with HPA through [k8s-prometheus-adapter](https://github.com/DirectXMan12/k8s-prometheus-adapter) it requires at least to know a namespace when [the HPA scaling is based on an object type](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#autoscaling-on-multiple-metrics-and-custom-metrics).
* Further automation: if there's some consumer service that makes decisions based on observed metrics, it somehow needs to know where to find the original resource to request its specs or state.

**But the service latency metric already contains the namespace entry in its `service` label.**

That's true. But the way [it's implemented](https://github.com/containous/traefik/blob/master/pkg/provider/kubernetes/crd/kubernetes_http.go#L91-L126) ([`namespace + "-" + name`](https://github.com/containous/traefik/blob/master/pkg/provider/kubernetes/crd/kubernetes.go#L628-L634)) doesn't cover all possible scenarios.

Imagine if you have the `my-namespace` namespace and the `the-resource` resource. In this case, even a powerful Prometheus [metric_relabel_configs](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs) won't help you much. With the relabeling rule like this:
```
    metric_relabel_configs:
      - source_labels:
          - service
        regex: ^([^-]+)-.*$
        replacement: ${1}
        target_label: namespace
      - source_labels:
          - service
        regex: "^[^-]+-(.*)-[^@]+@.*$"
        replacement: ${1}
        target_label: entrypoint
```
You'll get a processed metric:
```
traefik_service_request_duration_seconds_bucket{code="2000",entrypoint="namespace-the-resource",instance="10.6.152.9:8082",job="traefik",le="0.3",method="GET",namespace="my",protocol="http",service="my-namespace-the-resource-df34efd3b36ac5e9a216@kubernetescrd"}
```
Which is not exactly the original intention.

**All of this just comes down to a proper convention**

Having a standardized number of dashes in your namespaces would totally resolve the issue. Still, nothing beats **a distinct self-explanatory label name `namespace`** in a metric description.
```
traefik_entrypoint_request_duration_seconds_count{code="200",entrypoint="web",namespace="mynamespace",method="GET",protocol="http"} 2
traefik_service_request_duration_seconds_bucket{code="200",method="GET",protocol="http",namespace="mynamespace",service="mynamespace-myservice-9ab4060701404e59ffcd@kubernetescrd",le="0.1"} 2
```

**It would be better to have pod-level metrics**

It totally would! Instead of measuring latency on a Traefik's load balancer and averaging it, how about measuring it independently on every individual pod and producing the metrics that would include a pod label that later can be consumed by [k8s-prometheus-adapter](https://github.com/DirectXMan12/k8s-prometheus-adapter) and handed over to the HPA which is quite successful at balancing the scaling parameters by evaluating [the pod-level metrics](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/#autoscaling-on-multiple-metrics-and-custom-metrics).

sathieu:
### What does this PR do?

It allows to add Kubernetes resource namespace, kind and name to Prometheus metrics (when `--metrics.prometheus.addKubernetesLabels` cli flag is passed).

This is build on top of https://github.com/traefik/traefik/pull/12721 and would fix https://github.com/traefik/traefik/issues/6864.

### Motivation

Per-ingress metrics is very useful for enhanced dashboards. It is also needed for HPA (HorizontalPodAutoscaler).

### More

- [ ] Added/updated tests


nmengin:
Hey @sathieu,

rtribotte:
Hi @sathieu,

We discussed the changes with the other maintainers and were wondering about your motivations. Is it related to the ingress-nginx migration?

sathieu:
> Hi @sathieu,
> 
> We discussed the changes with the other maintainers and were wondering about your motivations. Is it related to the ingress-nginx migration?

@rtribotte Yes, it is related to ingress-nginx migration. We need metrics to be properly associated to the Kubernetes resource that is generating it.

In ingress-nginx, the following Prometheus labels are available (for `nginx_ingress_controller_requests`):

- container="controller"
- controller_class="k8s.io/ingress-nginx"
- controller_namespace="ingress-nginx"
- controller_pod="ingress-nginx-controller-42n42"
- endpoint="metrics"
- host="foo.bar.example.org" ✅ already present
- ingress="foo" 👈 needed
- method="GET" ✅ already present
- namespace="bar" 👈 needed
- path="/"
- pod="ingress-nginx-controller-42n42"
- service="foosvc"
- status="200" ✅ already present

We also need namespace and ingress name for [Horizontal Pod Autoscaling](https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/).

rtribotte:
Hi @sathieu,

Thanks again for this contribution!

We have closed #12721 in benefit of #12913, which provides an alternative mechanism to collecting per provider context metadata for observability.
For this reason this PR would have to be rebased, to use that mechanism.

However, while reviewing the current changes, we discussed about the impact of adding such labels, to all of our metrics.

We noticed that the actual request metric with Ingress-Nginx is providing more labels than the one you implemented, and we are wondering if we should bring them as well, and it would be preferable to know that set before adding them, to not introduce a little  breaking change by modifying the timeserie afterward.

We can keep the current PR open waiting for more discussions on our end, but we think it won't be ready in time to be shipped with v3.7

shreddedbacon:
> We noticed that the actual request metric with Ingress-Nginx is providing more labels than the one you implemented, and we are wondering if we should bring them as well, and it would be preferable to know that set before adding them, to not introduce a little breaking change by modifying the timeserie afterward.

It would be preferable to have the additional metrics on `Ingress` and all metrics traefik produces. Part of migrating away from `ingress-nginx` has us using Traefik now, but the missing labels on metrics that ingress-nginx provided is sorely missed here.

Is there a plan or timeline for getting the additional labels on all metrics?

kevinpollet:
Hello @shreddedbacon,

This is definitely on our radar.

azorahai3724:
This also concerns us as we are planning a migration from Ingress-NGINX, and we would need to do crazy workarounds from what we see now to get around it.

shreddedbacon:
G'day @kevinpollet. Just checking in, this is becoming a bit more pressing for us for being able to isolate by namespace. Regex matching on the service name is not reliable enough.

sathieu:
@kevinpollet @nmengin @rtribotte Please take a look again at this PR. I've rebased it on top of https://github.com/traefik/traefik/pull/12913. WDYT?

PR #12739, opened 2026-02-27
Support NGINX canary annotations

kevinpollet:
### What does this PR do?

This pull request adds the support of the following annotations:

* nginx.ingress.kubernetes.io/canary
* nginx.ingress.kubernetes.io/canary-by-cookie
* nginx.ingress.kubernetes.io/canary-by-header
* nginx.ingress.kubernetes.io/canary-by-header-value
* nginx.ingress.kubernetes.io/canary-by-header-pattern
* nginx.ingress.kubernetes.io/canary-weight
* nginx.ingress.kubernetes.io/canary-weight-total
* nginx.ingress.kubernetes.io/affinity-canary-behavior

Note that for the `nginx.ingress.kubernetes.io/affinity-canary-behavior` annotation only the `sticky` behavior is supported.  

### Motivation

To support canary annotations.

### More

- [X] Added/updated tests

### Additional Notes
Co-authored-by: Mathis Urien <contact.lbf38@gmail.com>

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
@@ -119,8 +119,9 @@ func (c canaryBackend) RequiresCanaryRouter() bool {
 
 // RequiresNonCanaryRouter returns true if the canary backend requires a non-canary router configuration for Cookie or Header routing.
 // This is the case when only the Header/Cookie options are configured, as a "never" value should forward the request to the production service.
+// When the canary weight is 0, no canary router should be created as all the traffic will be handled by the non-canary router based on the weight configuration.
 func (c canaryBackend) RequiresNonCanaryRouter() bool {
-	return (c.Header != "" && c.HeaderValue == "" && c.HeaderPattern == "") || c.Cookie != ""
+	return c.Weight > 0 && ((c.Header != "" && c.HeaderValue == "" && c.HeaderPattern == "") || c.Cookie != "")
 }
 
 // AppendCanaryRule appends the canary condition to the given rule based on the canary configuration for Cookie or Header routing.
@@ -135,11 +136,17 @@ func (c canaryBackend) AppendCanaryRule(rule string) string {
 			rules = append(rules, fmt.Sprintf("Header(`%s`, `%s`)", c.Header, c.HeaderValue))
 
 		default:
-			rules = append(rules, fmt.Sprintf("Header(`%s`, `%s`)", c.Header, "always"))
+			rules = append(rules, fmt.Sprintf("Header(`%s`, `always`)", c.Header))
 		}
 	}
+
 	if c.Cookie != "" {
-		rules = append(rules, fmt.Sprintf("HeaderRegexp(`Cookie`, `%s`)", fmt.Sprintf("(^|;\\s*)%s=always(;|$)", c.Cookie)))
+		cookieRule := fmt.Sprintf("HeaderRegexp(`Cookie`, `%s`)", fmt.Sprintf("(^|;\\s*)%s=always(;|$)", c.Cookie))
+		if c.Header != "" && c.HeaderValue == "" && c.HeaderPattern == "" {
+			cookieRule = fmt.Sprintf("(%s && !%s)", cookieRule, fmt.Sprintf("Header(`%s`, `never`)", c.Header))
+		}
+
+		rules = append(rules, cookieRule)
 	}
 
 	return fmt.Sprintf("(%s) && (%s)", rule, strings.Join(rules, " || "))
--- a/pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingresses-with-canary-by-header-and-cookie-and-weight.yml
+++ b/pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingresses-with-canary-by-header-and-cookie-and-weight.yml
@@ -1,0 +1,46 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: ingress-with-canary-by-header-and-cookie-and-weight
+  namespace: default
+
+spec:
+  ingressClassName: nginx
+  rules:
+    - host: production.localhost
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: whoami
+                port:
+                  number: 80
+
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: canary
+  namespace: default
+  annotations:
+    nginx.ingress.kubernetes.io/canary: "true"
+    nginx.ingress.kubernetes.io/canary-by-header: "Foo"
+    nginx.ingress.kubernetes.io/canary-by-cookie: "foo"
+    nginx.ingress.kubernetes.io/canary-weight: "10"
+
+spec:
+  ingressClassName: nginx
+  rules:
+    - host: production.localhost
+      http:
+        paths:
+          - path: /
+            pathType: Prefix
+            backend:
+              service:
+                name: canary
+                port:
+                  number: 80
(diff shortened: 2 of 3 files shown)


pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1893,3 +2112,45 @@ func nginxSizeToBytes(nginxSize string) (int64, error) {
...
+			Expires:  ptr.Deref(cfg.SessionCookieExpires, 0),
+			Path:     ptr.To(ptr.Deref(cfg.SessionCookiePath, "/")),
+			Domain:   ptr.Deref(cfg.SessionCookieDomain, ""),
+		},
+	}
+}
+
+func ingressPathKey(namespace, host string, pa netv1.HTTPIngressPath) string {
+	pathType := ptr.Deref(pa.PathType, netv1.PathTypePrefix)
+	return namespace + host + pa.Path + string(pathType)
```
```suggestion
	return namespace + "/" + host + "/" + pa.Path + "/" + string(pathType)
```

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1893,3 +2112,45 @@ func nginxSizeToBytes(nginxSize string) (int64, error) {
...
+}
+
+func canaryBackendKey(namespace string, backend netv1.IngressServiceBackend) string {
+	return namespace + backend.Name + portString(backend.Port)
```
```suggestion
	return namespace + "/" + backend.Name + "/" + portString(backend.Port)
```

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1561,6 +1730,56 @@ func applyForwardAuthConfiguration(routerName string, ingressConfig ingressConfi
...
+
+		weight := ptr.Deref(canaryIngressConfig.CanaryWeight, 0)
+		weightTotal := ptr.Deref(canaryIngressConfig.CanaryWeightTotal, 100)
+		canaryBackends[canaryBackendKey(namespace, *ingressPath.Backend.Service)] = &canaryBackend{
+			IngressBackend: &pa.Backend,
+			Weight:         min(max(weight, 0), weightTotal),
+			WeightTotal:    max(weightTotal, 0),
+			Header:         ptr.Deref(canaryIngressConfig.CanaryHeader, ""),
+			HeaderValue:    ptr.Deref(canaryIngressConfig.CanaryHeaderValue, ""),
+			HeaderPattern:  ptr.Deref(canaryIngressConfig.CanaryHeaderPattern, ""),
```
mmatur, line 1847: Do we want to use a regexp.Compile() validation before using the pattern

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -333,16 +403,76 @@ func (p *Provider) loadConfiguration(ctx context.Context) *dynamic.Configuration
...
 
+		i := ingress{
+			Ingress:       ing,
+			IngressConfig: parseIngressConfig(ing),
+		}
+
+		// Canary ingresses should be processed separately after processing all the ingresses
+		// to ensure that the canary rules are matching the ingress rules.
+		if ptr.Deref(i.IngressConfig.Canary, false) {
+			canaryIngresses = append(canaryIngresses, i)
```
mmatur, line 446: WDYT adding a Warn message saying that only sticky is supported for `nginx.ingress.kubernetes.io/affinity-canary-behavior`

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
kevinpollet: This has to be addressed globally, because in the provider, we are also using patterns in other annotations, and we should also check the NGINX behavior in that regard.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
kevinpollet: As discussed with @rtribotte, we think that this should be addressed when the ingressconfig is parsed to warn the user that some annotation values/annotations are not supported.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1619,6 +1790,56 @@ func applyForwardAuthConfiguration(routerName string, ingressConfig ingressConfi
...
+		// In case the service cannot be retrieved, or it has no endpoints we should ignore this matching canary rule.
+		// Here we are using the original Ingress configuration as a canary ingress inherit the original Ingress configuration.
+		if addresses, err := p.getBackendAddresses(namespace, pa.Backend, ingressPath.IngressConfig); err != nil || len(addresses) == 0 {
+			continue
+		}
+
+		canaryPaths[pathKey] = struct{}{}
+
+		weight := ptr.Deref(canaryIngressConfig.CanaryWeight, 0)
+		weightTotal := ptr.Deref(canaryIngressConfig.CanaryWeightTotal, 100)
```
gndz07: Value guard

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1619,6 +1790,56 @@ func applyForwardAuthConfiguration(routerName string, ingressConfig ingressConfi
...
+		canaryBackends[canaryBackendKey(namespace, *ingressPath.Backend.Service)] = &canaryBackend{
+			IngressBackend: &pa.Backend,
+			Weight:         min(max(weight, 0), weightTotal),
+			WeightTotal:    max(weightTotal, 0),
```
gndz07: On ingress-nginx, `weight-total` has a minimum value of 100.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -594,6 +709,39 @@ func (p *Provider) loadConfiguration(ctx context.Context) *dynamic.Configuration
...
+						logger.Error().
+							Str("serviceName", canaryBackend.IngressBackend.Service.Name).
+							Str("servicePort", canaryBackend.IngressBackend.Service.Port.String()).
+							Err(err).
+							Msg("Cannot create canary service")
+						continue
+					}
+
+					wrrServiceName = serviceName + "-wrr"
+					wrrService = &dynamic.Service{
```
gndz07, line 741: WDYT of creating the WRR service only when the canary-weight is more than 0?

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -1619,6 +1790,56 @@ func applyForwardAuthConfiguration(routerName string, ingressConfig ingressConfi
...
+
+		canaryPaths[pathKey] = struct{}{}
+
+		weight := ptr.Deref(canaryIngressConfig.CanaryWeight, 0)
+		weightTotal := ptr.Deref(canaryIngressConfig.CanaryWeightTotal, 100)
+		canaryBackends[canaryBackendKey(namespace, *ingressPath.Backend.Service)] = &canaryBackend{
+			IngressBackend: &pa.Backend,
+			Weight:         min(max(weight, 0), weightTotal),
+			WeightTotal:    max(weightTotal, 0),
+			Header:         ptr.Deref(canaryIngressConfig.CanaryHeader, ""),
```
gndz07, line 1845: Should we add value validation here against `headerRegexp`?

gndz07 (changes requested):

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
kevinpollet: The code becomes quite complex if doing that thus I think this is ok to keep it for now.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
kevinpollet: As said in a previous comment, the annotation validation should be addressed when parsing the annotations, and this has to be addressed globally.

PR #12740, opened 2026-02-27
Fix priority display in dashboard and ACME bypass redirect

mmatur:
### What does this PR do?

- Fix priority display in dashboard: Expose router priority as a string field (priorityStr) in the HTTP and TCP router API responses, since JavaScript cannot accurately represent large int64 values.
- Fix ACME bypass for entrypoint redirections: When allowACMEByPass is enabled on an entrypoint with HTTP-to-HTTPS redirection, exclude /.well-known/acme-challenge/ paths from the redirect rule so ACME HTTP-01 challenges can succeed.
- Fix incorrect TOML examples in entrypoints docs.

### More

- [x] Added/updated tests


gndz07 (approved):
LGTM

PR #12741, opened 2026-02-27
Use unicode.MaxASCII for clearer ASCII check

1911860538:
### What does this PR do?

Using unicode.MaxASCII with a greater-than check (>) more explicitly conveys the intent: "check if the character is outside the standard ASCII range". This aligns better with the function name IsASCII and makes the code more self-documenting for readers.

### More

- [ ] Added/updated tests


kevinpollet:
Hey @1911860538,

Thanks for this!

PR #12760, opened 2026-03-05
Remove path parsing with grpc healthcheck

rtribotte:
### What does this PR do?

This PR removes the path configuration parsing operation when executing grpc healthcheck.

### Motivation

To fix grpc healthcheck that should not be impacted by the path configuration.

### More

- [ ] Added/updated tests


kevinpollet (approved):

mmatur (approved):
LGTM

PR #12768, opened 2026-03-06
Bump sigs.k8s.io/gateway-api to v1.5.1

mmatur:
### What does this PR do?

Bump `sigs.k8s.io/gateway-api` from v1.4.0 to v1.5.1, promoting TLSRoute support from `v1alpha2` (experimental) to `v1` (stable GA).

**Key changes:**

- **TLSRoute promoted to GA**: TLSRoute no longer requires the `experimentalChannel` option. All TLSRoute types migrated from `gatev1alpha2` to `gatev1`.
- **TLS `Terminate` mode support**: Added `TLSRouteModeTerminate` as a supported extended feature, enabling TLS termination on TLS protocol listeners (previously only Passthrough was supported).
- **Deny unknown TLS hosts**: Added a default low-priority catch-all TCP router (`deny-unknown-host`) that rejects TLS connections for undefined hostnames, preventing traffic leaking to unintended backends.
- **Mixed TLS mode conflict detection**: New validation rejects listeners that mix Passthrough and Terminate modes on the same port, setting `ProtocolConflict` status conditions.
- **Wildcard hostname matching fix**: Fixed `findMatchingHostname` to correctly return the more specific hostname when both listener and route use wildcards (e.g., `*.bar.foo.com` is now correctly matched against `*.foo.com`).
- **HostSNIRegexp fix**: Corrected the wildcard-to-regex conversion to use `[a-z0-9-]+` instead of `[a-z0-9-\.]+`, preventing wildcard labels from matching across dot boundaries.
- **BackendTLSPolicy hardening**: Added empty string validation for `ca.crt` in ConfigMap-based CA certificates.
- **TLS protocol decoupled from TCPRoute**: TLS listeners now only support TLSRoute (no longer TCPRoute), aligning with the upstream spec.
- Updated CRD manifests, conformance test fixtures, and documentation for v1.5.1.

### Motivation

Align with the latest Gateway API specification. TLSRoute has been promoted to GA in v1.5.0, and v1.5.1 includes bugfixes to the conformance test suite.

### More

- [x] Added/updated tests

#### Conformance Test Results (v1.5.1)

| Profile | Core | Extended | Result |
|---------|------|----------|--------|
| **GATEWAY-HTTP** | 33/33 | 20/20 | All passed |
| **GATEWAY-GRPC** | 13/13 | — | All passed |
| **GATEWAY-TLS** | 18/18 | 3/3 (`TLSRouteModeTerminate`) | All passed |

### DIFF
--- a/integration/gateway-api-conformance-reports/v1.5.1/experimental-v3.7-default-report.yaml
+++ b/integration/gateway-api-conformance-reports/v1.5.1/experimental-v3.7-default-report.yaml
@@ -81,10 +81,11 @@ profiles:
     result: success
     statistics:
       Failed: 0
-      Passed: 3
+      Passed: 4
       Skipped: 0
     supportedFeatures:
     - GatewayPort8080
+    - TLSRouteModeMixed
     - TLSRouteModeTerminate
     unsupportedFeatures:
     - GatewayAddressEmpty
@@ -96,8 +97,8 @@ profiles:
     - GatewayInfrastructurePropagation
     - GatewayStaticAddresses
     - ListenerSet
-    - TLSRouteModeMixed
   name: GATEWAY-TLS
   summary: Core tests succeeded. Extended tests succeeded.
 succeededProvisionalTests:
+- TLSRouteMixedTerminationSameNamespace
 - TLSRouteTerminateSimpleSameNamespace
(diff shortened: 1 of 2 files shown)


fsimmend:
does this also mean, that Traefik now supports CORS configuration in HTTPRoute

nmengin:
Hey @ fsimmend,

Improving our Gateway API support, including CORS, is on our short-term roadmap.
Could you please open an issue for the CORS support, so that we can follow up on it once it is implemented.

PR #12773, opened 2026-03-06
Fix use-regex annotation behavior and add strictValidatePathType config for ingress-nginx provider

### LINKED ISSUE #12560: Rulesyntax for Ingress objects when using kubernetesIngressNginx provider
### Welcome!

### What did you do?

Hello!

Migrating from nginx ingress controller to traefik v3.6.7 using the helm chart.
I have certain Ingress Objects that have regexes in the path with pathType Prefix
E.g

```
  rules:
      host: my-app
      http:
          paths:
            - backend:
                service:
                  name:  my-app
                  port:
                    name: http-service
              path: /my-app(/|$)(.*)
              pathType: Prefix
```

Given the [Dynamic Configuration Changes](https://doc.traefik.io/traefik/v3.0/migration/v2-to-v3-details/#dynamic-configuration-changes), I understand that this will not work out the box. But I have tried a number of things - none of which work (what I've tried outlined at the bottom of this block). 

I suspect that the remediation actions suggested are built for IngressRoute objects and not necessarily for whatever happens under the hood to translate raw Ingress objects governed by the NGINX provider - but that's just my assumption here. So I've opened this ticket in the event that this is the case. If I should relog this elsewhere please let me know!

Things I've tried:
- Setting the router rules back to v2 globally - core.defaultRuleSyntax -  https://doc.traefik.io/traefik/v3.6/migrate/v3/#rule-syntax-configuration
- Setting `traefik.ingress.kubernetes.io/router.pathmatcher` on my Ingress Object
- Setting `traefik.ingress.kubernetes.io/router.rulesyntax` on my Ingress Object

[Issue I've found very helpful](https://github.com/traefik/traefik/issues/10672#issue-2272314209) for context, related to [this PR](https://github.com/traefik/traefik/pull/10985) 

I should also mention that it is only a problem for regexes. Static paths work as expected and the nginx ingress provider works as expected.

### What did you see instead?

Traefik returns a default 404 instead of routing to my backend

3.6.7

Using the helm chart v38.0.2

```yaml
providers:
  kubernetesIngressNginx:
    enabled: true
    controllerClass: "k8s.io/my-ingress-class"
    ingressClass: "private"

ports:
  websecure:
    http:
      sanitizepath: true
  web:
    redirections:
      entryPoint:
        to: websecure
        scheme: https
        permanent: true
```

_No response_

gndz07:
### What does this PR do?

This PR brings a fix for `use-regex` annotation for the ingress-nginx provider. When this annotation is set, it should apply to all ingresses on the same host. Additionally, location modifier regex should be case insensitive.

This PR also brings `strictValidatePathType` provider configuration, which allows user to configure whether or not they want to reject ingress paths containing regex characters when pathType is Prefix or Exact.

Fixes #12560 

### Motivation

To match the behavior of ingress-nginx when applying `use-regex` annotation.

### More

- [x] Added/updated tests


pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -1261,14 +1262,14 @@ func TestLoadIngresses(t *testing.T) {
 						},
 						"default-ingress-with-x-forwarded-prefix-regex-rule-0-path-0": {
 							EntryPoints: []string{"http"},
-							Rule:        "Host(`x-forwarded-prefix-regex.localhost`) && PathRegexp(`^(/something)(/.+)`)",
+							Rule:        "Host(`x-forwarded-prefix-regex.localhost`) && PathRegexp(`(?i)^(/something)(/.+)`)",
```
LBF38:
```suggestion
							Rule:        "Host(`x-forwarded-prefix-regex.localhost`) && PathRegexp(`(?i)^/(something)(/.+)`)",
```
Otherwise, it isn't a valid path for k8s ingress spec. WDYT?
(The fixture also needs to be changed.)

kevinpollet (approved):

rtribotte (approved):
LGTM

PR #12784, opened 2026-03-09
Fix unnecessary escaping of pipe in regexp examples

diegmonti:
### What does this PR do?

Removes unnecessary backslash escaping of the pipe character (`|`) in regexp examples in the HTTP routing rules & priority documentation. `\|` matches a literal pipe in Go regexp, while `|` is the alternation operator, which is what the examples intend to demonstrate.

### Motivation

The regexp examples in the [rules-and-priority docs](https://doc.traefik.io/traefik/reference/routing-configuration/http/routing/rules-and-priority/) escape the pipe as `\|` inside inline code blocks in Markdown tables.

### More

- [ ] Added/updated tests

### Additional Notes

None.


rtribotte (approved):
Thanks!

PR #12787, opened 2026-03-10
Support fragmented TLS client hello

rtribotte:
### What does this PR do?

This PR changes the TLS client hello read operations to delegate most of it to the go TLS implementation to support TLS fragmented client hello, and fail if an unexpected error happen.

### Motivation

To be compliant with https://datatracker.ietf.org/doc/html/rfc5246#section-6.2.1

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/server/router/tcp/router.go
+++ b/pkg/server/router/tcp/router.go
@@ -397,7 +397,7 @@ func clientHelloInfo(conn net.Conn) (*clientHello, error) {
 		GetConfigForClient: func(hello *tls.ClientHelloInfo) (*tls.Config, error) {
 			sni = hello.ServerName
 			protos = hello.SupportedProtos
-            // This error prevents unnecessary additional steps in the TLS ClientHello message processing.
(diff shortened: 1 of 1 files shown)


pkg/server/router/tcp/router.go
```diff
@@ -386,73 +381,48 @@ func clientHelloInfo(br *bufio.Reader) (*clientHello, error) {
...
+	var (
+		sni    string
+		protos []string
+	)
+	server := tls.Server(readOnlyConn{r: br}, &tls.Config{
 		GetConfigForClient: func(hello *tls.ClientHelloInfo) (*tls.Config, error) {
 			sni = hello.ServerName
 			protos = hello.SupportedProtos
-			return nil, nil
+			return nil, errClientHelloRead
```
juliens, line 401:
```suggestion
            // This error prevents unnecessary additional steps in the TLS ClientHello message processing.
			return nil, errClientHelloRead
```

PR #12789, opened 2026-03-10
Add configurable metrics temporality for OTLP exporter

JensPauwels:
## What does this PR do?

This PR adds a new `metrics.otlp.temporality` configuration option that allows users to choose between `cumulative`, `delta`, and `lowmemory` temporality for the OpenTelemetry metrics exporter. Previously, the OTel SDK default (cumulative) was always used with no way to override it from Traefik's configuration.

## Motivation

Some monitoring backends (e.g., Datadog) prefer delta temporality over cumulative. With cumulative temporality, counters report the running total since process start and never reset to zero, which can be confusing and requires rate functions on the query side. This change gives users direct control over temporality via Traefik's static configuration, without needing to rely on the `cumulativetodelta` OTel Collector processor or undocumented SDK environment variables.

## Usage

```yaml
metrics:
  otlp:
    temporality: delta
```

```bash
--metrics.otlp.temporality=delta
```

Accepted values: `cumulative` (default), `delta`, `lowmemory`. When empty or unrecognized, the SDK default (cumulative) is used, preserving full backward compatibility.

## More

- [x] Added/updated tests

## Additional Notes

The implementation uses the OTel Go SDK's built-in `DeltaTemporalitySelector`, `CumulativeTemporalitySelector`, and `LowMemoryTemporalitySelector` — no custom temporality logic was needed.


kevinpollet:
Hey @JensPauwels,

Thanks for this contribution!

PR #12794, opened 2026-03-11
Add FilterTLSClientCert middleware

nerzhul:
### What does this PR do?

This middleware permits to filter which certificates Subject/Issuers are allowed.

This permits some parity with `nginx ingress nginx.ingress.kubernetes.io/auth-tls-match-cn` annotation.

it's very useful in mTLS+HTTP ingress setup

### Motivation

We switch recently to traefik and we have a implement a WASM plugin in order to achieve this feature. It has to be implemented in rust as WASM+Golang has nasty memory issues for this usecase in traefik.

We think it's better to have a native feature in order to implement this.

### More

- [x] Added/updated tests


rtribotte:
Hey @nerzhul,

We've set this to "design-review" and will take a look at the code.

nerzhul:
@rtribotte when this design review is done generally ?

PhilippeChepy:
@rtribotte : is there any updates on the status of this work ?

PR #12805, opened 2026-03-13
Fix Host header not being modified by RequestHeaderModifier

### LINKED ISSUE #12793: Gateway API RequestHeaderModifier filter does not modify the Host header
### Welcome!

### What did you do?

Setting Host via RequestHeaderModifier via kubernetes gateway provider should modify the Host header seen by the upstream backend.

```
  kind: HTTPRoute
  metadata:
    name: example
  spec:
    parentRefs:
      - kind: Gateway
        name: my-gateway
    rules:
      - backendRefs:
          - kind: Service
            name: my-service
            port: 80
        filters:
          - type: RequestHeaderModifier
            requestHeaderModifier:
              set:
                - name: Host
                  value: desired-backend-host.example.com
        matches:
          - path:
              type: PathPrefix
              value: /
```
Similar to what customRequestHeaders middleware does.

### What did you see instead?

Host header is not modified. 

The regular Traefik Headers middleware (customRequestHeaders) handles this correctly in https://github.com/traefik/traefik/blob/master/pkg/middlewares/headers/header.go#L128-L129:
```
  case strings.EqualFold(header, "Host"):
      req.Host = value
```
  But the Gateway API middleware in https://github.com/traefik/traefik/blob/master/pkg/middlewares/gatewayapi/headermodifier/request_header_modifier.go#L42-L44 lacks this:
```
  for headerName, headerValue := range r.set {
      req.Header.Set(headerName, headerValue)  // no-op for Host
  }
```

3.6.8

kubernetes, installed via helmchart

_No response_

mihuross:
### What does this PR do?

In Go's net/http, setting the Host header via req.Header.Set is a no-op. The Host header must be set directly on req.Host, similar to what is already done in the customRequestHeaders middleware.

### Motivation

Fixes #12793

### More

- [x] Added/updated tests

To my understanding this does not need extra doc, because it is more of fix than extra feature.

### DIFF
--- a/pkg/middlewares/gatewayapi/headermodifier/request_header_modifier.go
+++ b/pkg/middlewares/gatewayapi/headermodifier/request_header_modifier.go
@@ -3,6 +3,7 @@ package headermodifier
 import (
 	"context"
 	"net/http"
+	"strings"
 
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
 	"github.com/traefik/traefik/v3/pkg/middlewares"
@@ -40,6 +41,10 @@ func (r *requestHeaderModifier) GetTracingInformation() (string, string) {
 
 func (r *requestHeaderModifier) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	for headerName, headerValue := range r.set {
+		if strings.EqualFold(headerName, "Host") {
+			req.Host = headerValue
+			continue
+		}
 		req.Header.Set(headerName, headerValue)
 	}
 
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hey @mihuross ,

Thanks for this contribution!

Could you please rebase this pull request on branch v3.6 as this is a fix?

mihuross:
> Could you please rebase this pull request on branch v3.6 as this is a fix?

sure, I was at the impression that this has been there from the start for few years already, so it is not something that should be merged as hot bug.

mihuross:
@mmatur any else that needs to be done from side?

nmengin:
Hey @mihuross,

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

traefiker:

PR #12811, opened 2026-03-14
AllowEmptyServices for Swarm provider

### LINKED ISSUE #12196: AllowEmptyServices option does not work with the Swarm provider
### Welcome!

### What did you do?

While reviewing GitHub pull request #10645, we found that the `AllowEmptyServices` option does not work with the Swarm provider.

Consequently, when a Swarm service is scaled down to zero, the corresponding router becomes inaccessible in Traefik, resulting in a `404` error.

v3.5.3

clifmo:
### What does this PR do?

Fixes the `AllowEmptyServices` option for the Swarm provider, which was previously non-functional despite being exposed in the configuration.

### Motivation

I'm using [Sablier](https://github.com/acouvreur/sablier) in a homelab context to scale Swarm services to zero when idle and wake them on demand. Without this fix, scaled-down services return 404s instead of allowing Sablier's middleware to intercept and wake the service.

### More

Fixes #12196 

- [x] Added/updated tests

### Additional Notes

Adds 2 new test cases validating both `true` and `false` behaviors. The implementation follows the exact same pattern as the Docker provider. Flag defaults to false for backwards compatibility.

### DIFF
--- a/pkg/provider/docker/pswarm.go
+++ b/pkg/provider/docker/pswarm.go
@@ -197,12 +197,12 @@ func (p *SwarmProvider) listServices(ctx context.Context, dockerClient client.AP
 		}
 
 		if dData.ExtraConf.LBSwarm {
-			if len(dData.NetworkSettings.Networks) > 0 {
+			if len(dData.NetworkSettings.Networks) > 0 || p.AllowEmptyServices {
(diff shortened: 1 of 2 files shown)


amazon7737 (approved):
LGTM

PR #12812, opened 2026-03-14
fix: handle Host header in Gateway API RequestHeaderModifier

### LINKED ISSUE #12793: Gateway API RequestHeaderModifier filter does not modify the Host header

mango766:
## What does this PR do?

Fixes #12793

The Gateway API `RequestHeaderModifier` filter does not correctly modify the `Host` header. In Go's `net/http`, the `Host` header is special: calling `req.Header.Set("Host", value)` only modifies the `Header` map but does **not** update `req.Host`, which is the field actually used when forwarding requests to backend services. This means setting the `Host` header via the Gateway API `RequestHeaderModifier` filter is silently ignored.

## Root Cause

The existing `headers` middleware (`customRequestHeaders`) already handles this correctly in [`pkg/middlewares/headers/header.go`](https://github.com/traefik/traefik/blob/master/pkg/middlewares/headers/header.go#L128-L129):

```go
case strings.EqualFold(header, "Host"):
    req.Host = value
```

But the Gateway API middleware in [`pkg/middlewares/gatewayapi/headermodifier/request_header_modifier.go`](https://github.com/traefik/traefik/blob/master/pkg/middlewares/gatewayapi/headermodifier/request_header_modifier.go) does not apply this pattern, so `Host` modifications are silently dropped.

## Changes

- Added special handling for the `Host` header in both `set` and `add` operations of the `requestHeaderModifier.ServeHTTP` method, using `strings.EqualFold` for case-insensitive matching and setting `req.Host` directly
- Added two new test cases (`set Host header modifies request host` and `add Host header modifies request host`) to verify the fix
- All existing tests continue to pass

## Test Plan

- [x] `go test ./pkg/middlewares/gatewayapi/headermodifier/...` -- all 12 tests pass (10 existing + 2 new)
- [x] Verified the fix matches the pattern used by the existing `headers` middleware

### DIFF
--- a/pkg/middlewares/gatewayapi/headermodifier/request_header_modifier.go
+++ b/pkg/middlewares/gatewayapi/headermodifier/request_header_modifier.go
@@ -3,6 +3,7 @@ package headermodifier
 import (
 	"context"
 	"net/http"
+	"strings"
 
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
 	"github.com/traefik/traefik/v3/pkg/middlewares"
@@ -40,10 +41,21 @@ func (r *requestHeaderModifier) GetTracingInformation() (string, string) {
 
 func (r *requestHeaderModifier) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	for headerName, headerValue := range r.set {
+		// In Go's net/http, the Host header is not stored in the Header map
+		// but in the Request.Host field. Setting it via Header.Set is a no-op
+		// for the actual host used in upstream requests.
+		if strings.EqualFold(headerName, "Host") {
+			req.Host = headerValue
+			continue
+		}
 		req.Header.Set(headerName, headerValue)
 	}
 
 	for headerName, headerValue := range r.add {
+		if strings.EqualFold(headerName, "Host") {
+			req.Host = headerValue
+			continue
+		}
 		req.Header.Add(headerName, headerValue)
 	}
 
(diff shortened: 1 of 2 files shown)


nmengin:
Hey @mango766,

Thank you for this, but it looks like it's a duplicate of #12805.
I close the PR, please leave a comment if you disagree; we'll reopen it if necessary.

PR #12824, opened 2026-03-17
kubernetes/ingress-nginx: add defaultTLS compatibility flag

Ph4rell:
### What does this PR do?

Adds a `defaultTLS` option to the Kubernetes Ingress NGINX provider.
When enabled, ingresses without `spec.tls` are treated as TLS if only TLS entrypoints exist,
so the default TLS store certificate can be used.

### Motivation

Upgrading to v3.7 changes entrypoint attachment for ingresses without `spec.tls`.
This option preserves the previous behavior for users relying on a default TLS certificate,
without forcing a spec.tls update on every ingress.

### More

- [x] Added/updated tests

### Additional Notes

- Tests not run locally because `go` is not available in this environment.
- Generated configuration options doc was not regenerated for the same reason.

### DIFF
--- a/docs/content/reference/install-configuration/providers/kubernetes/kubernetes-ingress-nginx.md
+++ b/docs/content/reference/install-configuration/providers/kubernetes/kubernetes-ingress-nginx.md
@@ -73,6 +73,9 @@ providers:
     globalAllowedResponseHeaders:
       - "X-Custom-Header1"
       - "X-Custom-Header2"
+    # Compatibility: treat ingresses without spec.tls as TLS
+    # when only TLS entrypoints exist
+    defaultTLS: false
 ```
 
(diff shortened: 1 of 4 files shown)


Ph4rell:
Potential fix for https://github.com/traefik/traefik/issues/12820

sathieu:
Maybe the flag should default to true, to have the same behavior as ingress-nginx?

gndz07:
Hi @Ph4rell,

We've investigated all issues related to TLS in the ingress-nginx provider and we decided to address the root cause. Specifically, we aim to align with the behavior of ingress-nginx.
We have opened a PR #12831 that takes this approach to fix the issues.

We appreciate your contribution, but we will close this PR in favor of the aforementioned PR .

PR #12831, opened 2026-03-18
Fix TLS behavior in ingress-nginx provider

LBF38:
### What does this PR do?

This PR fixes the TLS behavior in the ingress-nginx provider to match the original behavior:
- If an ingress doesn't have a TLS section, default certificate will be provided but redirect will not be forced.
- If an ingress has an invalid certificate, it will fallback to default certificate.

### Motivation

- Fixes https://github.com/traefik/traefik/issues/12820
- Fixes https://github.com/traefik/traefik/issues/12748

### More

- [x] Added/updated tests

### Additional Notes


sathieu:
@LBF38 @gndz07 Please rebase on top of https://github.com/traefik/traefik/pull/12808 or the CI will fail.

PR #12832, opened 2026-03-19
Handle empty rewrite-target like unset rewrite-target

sathieu:
### What does this PR do?

Handle the following annotation:

```yaml
nginx.ingress.kubernetes.io/rewrite-target: ""
```

like no annotation at all.

### Motivation

Fixes: https://github.com/traefik/traefik/issues/12829

### More

- [ ] Added/updated tests


rtribotte:
Hey @sathieu,

Thanks for this contribution!

PR #12834, opened 2026-03-19
fix: preserve health check status updater when service has headers

jspdown:
### What does this PR do?

Wrap the headers handler in a `statusUpdaterHandler` that preserves the `StatusUpdater` interface from the original service handler. 

### Motivation

When a WRR child service has both headers and health checks configured, the headers wrapping was losing the `StatusUpdater` interface, preventing health check status propagation to the parent load balancer.

### More

- [x] Added/updated tests

### Additional Notes

I'm targeting v3.6 since this piece of code has been introduced in v3.6.0.

### DIFF
--- a/pkg/server/service/service.go
+++ b/pkg/server/service/service.go
@@ -330,13 +330,18 @@ func (m *Manager) getServiceHandler(ctx context.Context, service dynamic.WRRServ
 	}
 
 	if service.Headers != nil {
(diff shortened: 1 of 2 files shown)


juliens:
This headers feature works only in Knative which has no healthcheck.
I close this, the best thing to do will be in the futur to remove this specific Headers feature to use the new middleware on service feature.

PR #12863, opened 2026-03-23
Fix StripPrefix and StripPrefixRegex to slice the prefix using encoded prefix length

gndz07:
### What does this PR do?

`StripPrefixRegex` matches the regex against the decoded `URL.Path` to extract the prefix, then uses `len(prefix)` to strip it from the encoded `URL.RawPath`. This could produce a corrupted RawPath when the prefix region contains percent-encoded characters.

This PR fixes the offset by scanning RawPath and advancing 3 bytes per %XX sequence, so the slice point correctly matches the encoded length of the prefix.

This PR also addresses the similar problem with `StripPrefix` middleware.

### Motivation

To avoid corrupted RawPath when the prefix contains encoded characters.

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: Mathis Urien <contact.lbf38@gmail.com>


rtribotte (approved):
LGTM

kevinpollet (approved):

PR #12864, opened 2026-03-23
Remove duplicate step in CI

mloiseleur:
### What does this PR do?

1. Remove duplicated step when checking diff on go mod tidy.
   * It's already done with `make validate-files` from `validate` job, that calls `script/validate-vendor.sh`

### Motivation

Avoid duplication when it's possible.


.github/workflows/validate.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
           version: "${{ env.GOLANGCI_LINT_VERSION }}"
 
   validate:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
```
mloiseleur:
```suggestion
    runs-on: ubuntu-latest
```

PR #12879, opened 2026-03-26
Make duplicate slash normalization optional

### LINKED ISSUE #12666: "duplicated slashes" normalization should be optional
### Welcome!

Since #11768 Traefik normalizes requests paths according to the rules described in [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.2).
But the "merge slashes" sanitization is actually neither mentioned, nor recommended in RFC 3986. According to [Section 3.3](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) empty path segments are allowed (it's not explicitly mentioned in the text, but according to the ABNF empty segments are allowed).

As this "merge slashes" sanitization actually changes the semantic of the path, the **feature should be optional** (i don't care about the default value).
i.e. it should be possible to disable the "merge slashes" sanitization but keep the RFC-conform normalization (e.g. the dot-segment replacement)

just for reference:
* In Apache httpd this can be enabled/disabled with the [MergeSlashes](https://httpd.apache.org/docs/current/mod/core.html#mergeslashes) directive (enabled by default)
* In nginx this can be enabled/disabled with [merge_slashes](https://nginx.org/en/docs/http/ngx_http_core_module.html#merge_slashes) (enabled by default)
* In Istion this can be configured with the `MERGE_SLASHES` [pathNormalization](https://istio.io/latest/docs/ops/best-practices/security/#customize-your-system-on-path-normalization) (**disabled** by default)

boinger:
### What does this PR do?

Adds a `mergeSlashes` entrypoint option (default `true`) that controls whether consecutive slashes are collapsed. Only takes effect when `sanitizePath` is enabled.

When `mergeSlashes` is `false`, `//` is preserved in forwarded paths while `..`/`.` segments are still resolved. The default codepath is unchanged — both-true still uses the existing `JoinPath` implementation.

### Motivation

Fixes #12666

`sanitizePath` currently bundles dot-segment resolution with slash merging via `URL.JoinPath()`. Users who need to preserve empty path segments (which RFC 3986 §3.3 allows) have to disable all sanitization, losing path traversal protection. This is a problem in multi-proxy setups where upstream gateways rely on the original path structure for authorization.

### More

- [x] Added/updated tests

Also fixed a minor `WriteString(string())` → `WriteByte` in `normalizePath` while I was in there.


nmengin:
Hey @boinger,

Thanks for this contribution!

PR #12881, opened 2026-03-26
Use 2^53-1 for CRD priority maximum to avoid float64 rounding

### LINKED ISSUE #12693: F64 rounding error in IngressRoute* CRD definitions
### Welcome!

### What did you do?

I attempted to apply Traefik CRDs to my Kubernetes cluster using ArgoCD. However, `IngressRouteTCP` and `IngressRoute` (traefik.io) always stayed out-of-sync. No diff was shown in the UI.

A bit of digging revealed that the [`Priority` field](https://github.com/traefik/traefik/blob/4c9c70b7f5ff0e58cd9445f25997a608ad11e0b9/pkg/provider/kubernetes/crd/traefikio/v1alpha1/ingressroutetcp.go#L33) on `RouteTCP` has a kubebuilder annotation setting the maximum value to `9223372036854774807`. The field is defined as an `int` and the value is `<2^64` so it should be fine™

### What did you see instead?

At multiple points in the CRD journey this value may be converted into a float64 but its handling depends on the tooling in use.

1. kubebuilder already [fudges the value](https://github.com/traefik/traefik/blob/4c9c70b7f5ff0e58cd9445f25997a608ad11e0b9/integration/fixtures/k8s/01-traefik-crd.yml#L140) to the nearest representable f64 `9223372036854775000` by rounding up to the nearest float-representable value, so the CRD deployed alongside Traefik _does not match the expected value_
2. Running the CRD YAML through Go at any point (e.g. by using `go-jsonnet` or ArgoCD) strangely yields a **different** value of `9223372036854774784` which is non-representable as a float64. This can be reproduced using the snippet below
3. Submitting the original value of `9223372036854774807` to K8s yields the same rounded value as `kubebuilder`

While it is not (technically speaking) a problem with Traefik, the used maximum value proves as highly problematic since different tools fudge it in unpredictable ways, presumably by round-tripping through JSON where many implementations use float64 for all numbers. I would question the benefit of having this maximum defined, especially since it is some arbitrary-seeming value instead of the actual boundary `2^64-1`. It causes more trouble than it is worth in its current form IMO.

---

Code example showing how Go fudges an actually representable f64 to something weird in between:
```go
package main

import "fmt"

func main() {
	value := 9223372036854775000.0
	fmt.Printf("%f", value) // -> 9223372036854774784
}
```

This can be reproduced using `go-jsonnet` which is used by ArgoCD among other things:

```bash
go-jsonnet -e 9223372036854775000 # -> 9223372036854774784

# The C jsonnet implementation behaves "correctly"
jsonnet -e 9223372036854774784.000000 # -> 9223372036854775000
```

The same example [in Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=9c3783be66ca4bf8908379ed37cae213) yields the expected representable value of `9223372036854775000` which matches the behaviour of [this float playground](https://evanw.github.io/float-toy/).

v3.6.7

- K8s `1.34.0`
- ArgoCD `v3.2.5+c56f440`
- Traefik `v3.6.7`
  - Deployed via ArgoCD
  - Round-trip through jsonnet, rendered using [standard ArgoCD tooling](https://argo-cd.readthedocs.io/en/stable/user-guide/jsonnet/)

boinger:
### What does this PR do?

Lower the kubebuilder maximum validation on IngressRoute and IngressRouteTCP
priority fields from `9223372036854774807` to `9007199254740991` (2^53 - 1).

The old value does not survive float64 round-tripping — different JSON/YAML
parsers produce different results (Go gives 9223372036854774784, kubebuilder
outputs 9223372036854775000, go-jsonnet gives yet another value). This causes
tools like ArgoCD to see the CRD as perpetually out-of-sync.

2^53 - 1 is the largest integer exactly representable as a float64. Still
absurdly large for a router priority.

### Motivation

Fixes #12693

### More


nmengin:
Hey @boinger,

Thanks for this contribution!

PR #12892, opened 2026-03-26
Restore disableClusterScopeResources option in current documentation

mloiseleur:
### What does this PR do?

Restore a missed option (example [here](https://doc.traefik.io/traefik/v3.2/providers/kubernetes-ingress/#disableclusterscoperesources) in v3.2) in the current version of Proxy documentation.
Since disableIngressClassLookup is deprecated since v3.1, I just switched the option with the current one.

### Motivation

Accurate documentation

### More

### Additional Notes

I discovered it when reviewing and updating links in Helm chart values.


mloiseleur:
It did not appear in the search engine, but the option is here.
https://doc.traefik.io/traefik/reference/install-configuration/providers/kubernetes/kubernetes-ingress/#opt-providers-kubernetesIngress-disableClusterScopeResources

PR #12901, opened 2026-03-27
Add custom json log format

thomaspeitz:
Hey folks,
first and foremost thanks for your tool!
We currently migrate away from nginx and we love it!

### What does this PR do?
Allows to set custom json log format.

Adds a jsonTemplate field to accessLog configuration that accepts a Go text/template string.

When format: json is set and jsonTemplate is non-empty, Traefik renders each access log entry by executing the template against a flat map[string]any of all log fields (plus time, level, and msg built-ins).
The template is compiled once at startup.
An invalid template causes Traefik to fail fast with a descriptive error.

### Motivation
Motivation is written down here:
https://github.com/traefik/traefik/issues/12896

### More

- [x] Added/updated tests

### Additional Notes

  - Fully opt-in and backward-compatible. Empty jsonTemplate (the default) preserves existing behavior exactly.
  - Startup-time validation. An invalid Go template string causes NewHandler to return an error, so misconfiguration is caught before Traefik begins serving traffic.
  - Format mismatch is warned, not fatal. If jsonTemplate is set alongside a non-JSON format, a warning is logged and the field is ignored, consistent with how Traefik handles other format mismatches.
  - Fields are accessed via {{ index . "FieldName" }} using the same field name constants defined in pkg/middlewares/accesslog/fields.go (e.g. "ClientHost", "DownstreamStatus",  "Duration"). Method calls on values work as expected ((index . "Duration").Milliseconds, (index . "StartUTC").Format "...").

Watch out ‼️
- I added helm docs, but did not look into this. Want this one at least verified before moving forward and adding and documenting it there.


thomaspeitz:
If you could give this free as an image build as well, we test it for a week on nonprod.
Currently we are bit short on resources to build it ourselves.


Here is a performance review as well (ai vibe coded) - Can add to the PR as well but would not add it to the codebase to be honest.
```

  ┌───────────────────────┬─────────┬──────────┬───────────┐
  │       Formatter       │ Time/op │ Bytes/op │ Allocs/op │
  ├───────────────────────┼─────────┼──────────┼───────────┤
  │ logrus.JSONFormatter  │ ~2.0 µs │ 1968 B   │ 36        │
  ├───────────────────────┼─────────┼──────────┼───────────┤
  │ TemplateJSONFormatter │ ~4.7 µs │ 3195 B   │ 96        │
  └───────────────────────┴─────────┴──────────┴───────────┘

  ~2.4× slower, ~1.6× more memory, ~2.7× more allocations than the built-in JSON formatter.

  The cost comes from three things:
  1. Map copy — every log line copies entry.Data into a new map (necessary to add time/level/msg and wrap strings)
  2. text/template execution — reflection-heavy, allocates per field access
  3. json.Marshal per string field — called once each time a string field is rendered, to escape content

  In absolute terms: ~5 µs per log line. At that rate a single core can format ~200k lines/sec. In practice it doesn't matter — access logging is almost always async (bufferingSize)
  and I/O-bound, not CPU-bound. The formatter is not on the request hot path.
```

nmengin:
Hey @thomaspeitz ,
Thanks for this contribution!

thomaspeitz:
@nmengin  - Had some issues on the pipelines.
Should be fixed by now. At least "works locally" :D

LSGUNDL:
Hi! I wanted to ask, why the PR got closed? 🥺 We where happily waiting that this feature gets it's way into the main

PR #12902, opened 2026-03-27
Allow docker image to bind to privileged ports

sathieu:
### What does this PR do?

Allows traefik to bind to ports below 1024, like the standards http and https ports (80 and 443).

### Motivation

While migrating from ingress-nginx to traefik, we got timeout issues in some applications, typically those authenticating with OpenID Connect.

This is due to how network policies are handled (at least with Cilium):

- we have an *egress* network policy in the source namespace allowing connections to any destination on port 443
- when the source pod targets a website in the same cluster published thru a LoadBalancer service (for example Keycloak, served by traefik), Cilium sends the packets directly to the service's backend (NB : we also use [Cilium to manage LoadBalancer services](https://docs.cilium.io/en/stable/network/l2-announcements/))
- in this case, the target port in 8443 (websecure in the pods) and not 443 (in the service)
- for this reason, the connection is not allowed and leads to timeout

To fix this, we use and image including this PR, and the following values in the Helm chart:

```yaml
ports:
  web:
    port: 80
  websecure:
    port: 443

securityContext:
  capabilities:
    add:
      - NET_BIND_SERVICE
``` 

Related issues:

- https://github.com/traefik/traefik/issues/11679
- https://github.com/traefik/traefik/issues/8595

Related refs:

- [Unable to listen on port (80/443)](https://github.com/kubernetes/ingress-nginx/blob/dbb11b92ddb77bee9c35e462479129354c84f939/docs/troubleshooting.md#unable-to-listen-on-port-80443) in ingress-nginx *Troubleshooting* doc.

### More

- [x] Added/updated tests (with `setcap -v`)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,13 @@
 # syntax=docker/dockerfile:1.2
 FROM alpine:3.23
 
-RUN apk add --no-cache --no-progress ca-certificates tzdata
 
 ARG TARGETPLATFORM
 COPY ./dist/$TARGETPLATFORM/traefik /
+
+RUN apk add --no-cache --no-progress ca-certificates libcap-setcap tzdata && \
+  setcap    cap_net_bind_service=+ep /traefik && \
+  setcap -v cap_net_bind_service=+ep /traefik
 
 EXPOSE 80
 VOLUME ["/tmp"]


rtsui-mk:
I'm facing the same issue with traefik running on hostnetwork to listen for 80 and 443
even with NET_BIND_SERVICE, this should help

nmengin:
Hey @sathieu,

Thanks for the contribution!

darkweaver87:
The use case is valid, but I don't think this belongs in the upstream Dockerfile. Adding `cap_net_bind_service` and `libcap-setcap` to every published image changes the default security posture for all users to solve a niche deployment scenario (Cilium + port-based network policies).

Also, `setcap` on the binary shouldn't even be necessary here. In Kubernetes, capabilities are granted at the process level by the container runtime. Just setting `securityContext.capabilities.add: [NET_BIND_SERVICE]` in your Helm values (along with the port remapping) should be enough — no image change needed.

This should already work with the following chart `values.yaml` (tested with [v40.0.0-ea.3](https://github.com/traefik/traefik-helm-chart/releases/tag/v40.0.0-ea.3)) and the standard `traefik:v3.7.0-ea.3` image:

```yaml
# Traefik Helm chart v40.0
# Bind on privileged ports 80/443 (e.g. migration from ingress-nginx with Cilium)

ports:
  web:
    port: 80
  websecure:
    port: 443

securityContext:
  capabilities:
    drop: [ALL]
    add: [NET_BIND_SERVICE]
```

Here's the resulting deployed spec confirming it works (pod running, non-root, no `setcap`):

```yaml
spec:
  template:
    spec:
      containers:
      - args:
        - ...
        - --entryPoints.web.address=:80/tcp
        - --entryPoints.websecure.address=:443/tcp
        - ...
        image: docker.io/traefik:v3.7.0-ea.3
        ports:
        - ...
        - containerPort: 80
          name: web
        - containerPort: 443
          name: websecure
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            add:
            - NET_BIND_SERVICE
            drop:
            - ALL
          readOnlyRootFilesystem: true
      ...
      securityContext:
        runAsGroup: 65532
        runAsNonRoot: true
        runAsUser: 65532
        ...
status:
  availableReplicas: 1
  readyReplicas: 1
```

sathieu:
@darkweaver87 

>  Just setting securityContext.capabilities.add: [NET_BIND_SERVICE] in your Helm values (along with the port remapping) should be enough

No. This doesn't work. I started with this, without success.

When the image is running as root, setting this capability is enough, because root has all caps; but when running the image as non-root (the default from the chart) we need both : the cap on the file **and** in the bounding set. This is clearly explained [in the ingress-nginx docs](https://github.com/kubernetes/ingress-nginx/blob/dbb11b92ddb77bee9c35e462479129354c84f939/docs/troubleshooting.md#unable-to-listen-on-port-80443)

I don't agree that this is a niche:

- using hostNetwork is common on small clusters (where there is no load balancer).
- using Cilium is very common (60% of Kubernetes according to [this](https://www.cncf.io/blog/2025/12/18/cilium-releases-2025-annual-report-a-decade-of-cloud-native-networking/)), and using netpols is probably not as common as it should, but adding more barriers won't help

darkweaver87:
I tested the full scenario (Cilium + network policy allowing only egress on port 443) on a kind cluster, both with and without `hostNetwork`. It works with the standard image and Helm values only — no `setcap` needed.

### Setup

kind cluster with Cilium as CNI (no kube-proxy):

```yaml
# kind-cilium.yaml
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
networking:
  disableDefaultCNI: true
  kubeProxyMode: none
nodes:
- role: control-plane
- role: worker
```

```bash
kind create cluster --config kind-cilium.yaml
helm install cilium cilium/cilium --namespace kube-system \
  --set kubeProxyReplacement=true \
  --set k8sServiceHost=kind-control-plane \
  --set k8sServicePort=6443
```

Traefik with privileged ports (`values.yaml`):

```yaml
ports:
  web:
    port: 80
  websecure:
    port: 443

securityContext:
  capabilities:
    drop: [ALL]
    add: [NET_BIND_SERVICE]
```

```bash
helm install traefik traefik/traefik -f values.yaml --devel
```

### Test scenario

Backend app + network policy allowing only egress on port 443 (and DNS):

```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: backend
---
apiVersion: v1
kind: Namespace
metadata:
  name: client
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whoami
  namespace: backend
spec:
  replicas: 1
  selector:
    matchLabels:
      app: whoami
  template:
    metadata:
      labels:
        app: whoami
    spec:
      containers:
      - name: whoami
        image: traefik/whoami:latest
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: whoami
  namespace: backend
spec:
  selector:
    app: whoami
  ports:
  - port: 80
    targetPort: 80
---
kind: IngressRoute
metadata:
  name: whoami
  namespace: backend
spec:
  entryPoints:
  - websecure
  routes:
  - match: Host(`whoami.local`)
    kind: Rule
    services:
    - name: whoami
      port: 80
  tls: {}
---
kind: NetworkPolicy
metadata:
  name: allow-egress-443-only
  namespace: client
spec:
  podSelector: {}
  policyTypes:
  - Egress
  egress:
  - to: []
    ports:
    - protocol: TCP
      port: 443
  - to: []
    ports:
    - protocol: UDP
      port: 53
---
apiVersion: v1
kind: Pod
metadata:
  name: client
  namespace: client
spec:
  containers:
  - name: curl
    image: curlimages/curl:latest
    command: ["sleep", "3600"]
```

### Results (ClusterIP)

Port 443 (allowed by network policy) — **passes**:

```bash
$ kubectl exec -n client client -- curl -sk --connect-timeout 5 -H "Host: whoami.local" https://traefik.default.svc.cluster.local:443
Hostname: whoami-866c4f9b89-6ch5c
...
X-Forwarded-Port: 443
X-Forwarded-Proto: https
```

Port 80 (not allowed by network policy) — **blocked**:

```bash
$ kubectl exec -n client client -- curl -s --connect-timeout 5 -H "Host: whoami.local" http://traefik.default.svc.cluster.local:80
# exit code 28 — connection timeout
```

### Results (hostNetwork)

Also tested with `hostNetwork: true`:

```bash
helm upgrade traefik traefik/traefik -f values.yaml --devel \
  --set hostNetwork=true \
  --set service.enabled=false
```

Traefik binds on ports 80/443 on the node IP and runs fine — no `setcap` needed either.

### Conclusion

No image change required. The `securityContext.capabilities.add: [NET_BIND_SERVICE]` Helm value is sufficient in both ClusterIP and hostNetwork modes.

darkweaver87:
Fair point regarding the "niche" wording — that was probably too dismissive. To clarify:

- **hostNetwork** is common on small/bare-metal clusters, agreed.
- **Cilium** adoption is significant ([~60% according to Datadog's 2024 report](https://www.datadoghq.com/container-report/)).

That said, the issue described in this PR requires a very specific combination to manifest: Cilium with kube-proxy replacement + direct pod routing (bypassing Service DNAT) + network policies filtering on destination port + intra-cluster traffic hitting a Traefik-fronted service. Each of these is common individually, but the intersection is narrower than it appears.

Regardless of how common the scenario is, the key point remains: the fix doesn't require an image change. As demonstrated in my [previous comment](https://github.com/traefik/traefik/pull/12902#issuecomment-4155091234), both ClusterIP and hostNetwork modes work with the standard Traefik image — `securityContext.capabilities.add: [NET_BIND_SERVICE]` in Helm values is sufficient, no `setcap` needed.

@sathieu Out of curiosity, did you try using only the Helm values (`securityContext.capabilities.add: [NET_BIND_SERVICE]` + port remapping) without the Dockerfile change? If so, what was the error? Would be helpful to know your runtime environment (container runtime, Kubernetes version, Cilium version) in case there's an edge case we're missing.

sathieu:
@darkweaver87 Your test is using kind which has no privileged ports:

```console
$ # outside of kind
$ cat /proc/sys/net/ipv4/ip_unprivileged_port_start
1024

$ # in kind:
$ docker exec  -ti kind-control-plane bash
root@kind-control-plane:/# cat proc/sys/net/ipv4/ip_unprivileged_port_start
0
``` 

So no port is considered privileged. This is not the case with standard Kubernetes installs.

darkweaver87:
@sathieu thanks for the detailed context. I'll try this out.

sathieu:
> did you try using only the Helm values (securityContext.capabilities.add: [NET_BIND_SERVICE] + port remapping) without the Dockerfile change? If so, what was the error?

Yes. The message was `listen tcp :80: bind: permission denied`

> Would be helpful to know your runtime environment (container runtime, Kubernetes version, Cilium version) in case there's an edge case we're missing.

- Kubernetes v1.34
- containerd v2.2
- runc
- cilium v1.19

---

It looks like docker is setting `net.ipv4.ip_unprivileged_port_start `to 0. See https://github.com/moby/moby/issues/42728

mloiseleur:
@sathieu Wdyt of setting this sysctl as you expect on your cluster?

Something like this in the `Deployment`:

```yaml
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    spec:
      securityContext:
        sysctls:
        - name: "net.ipv4.ip_unprivileged_port_start"
          value: "0"
```

sathieu:
@mloiseleur We have an [admission policy preventing that](https://github.com/open-policy-agent/gatekeeper-library/tree/master/library/pod-security-policy/forbidden-sysctls):

```
The sysctl net.ipv4.ip_unprivileged_port_start is not allowed, pod: traefik-65498b5988-wbl9f. Forbidden sysctls: ["*"]
```

I don't understand why this change has so much friction. Have you identified any risk with it?

mloiseleur:
Sorry if you felt any friction; that was not intended. 

We are security-first oriented. Adding _cap_net_bind_service_ and _libcap-setcap_ to every published image changes the default security posture. 

We have a clear interest in keeping this _default_ security posture for _all_ users. The same is true for OPA, which prevents that by default.

Since it's your use case, may I invite you to configure a tailored exception in your admission policy?

sathieu:
I found the reason: recent containerd (>= 2.0) allows unprivileged ports by default ([ref](https://github.com/containerd/containerd/blob/131286b17755467322f5fcc2bf90842d4c414b79/docs/cri/config.md?plain=1#L436-L441)), but on clusters deployed with Kubespray, this is disabled by default ([ref](https://github.com/kubernetes-sigs/kubespray/blob/90ce8004b7c773432cbb8a048d32bfb89601ca76/roles/container-engine/containerd/defaults/main.yml#L71)).

We will change this Kubespray value instead: `containerd_enable_unprivileged_ports: true`

darkweaver87:
@sathieu I managed to reproduce your issue by installing k3s + Cilium directly on a VM (GCP, `ip_unprivileged_port_start=1024`, Cilium with kube-proxy replacement). My previous kind-based tests were invalid as you pointed out — kind sets `ip_unprivileged_port_start=0`.

**ClusterIP mode**: works with just Helm values, no image change needed. containerd >= 2.0 sets `enable_unprivileged_ports=true` by default in the pod's network namespace:

```
Host:  ip_unprivileged_port_start = 1024
Pod:   ip_unprivileged_port_start = 0
```

**hostNetwork mode**: `NET_BIND_SERVICE` via securityContext alone is indeed not enough for a non-root process. Two options, both working without any image change:

1. **Run as root** — consistent with the privilege level (you already have full access to the host's network stack):

```yaml
ports:
  web:
    port: 80
  websecure:
    port: 443

hostNetwork: true

podSecurityContext:
  runAsUser: 0
  runAsNonRoot: false
  runAsGroup: 0
```

2. **Set `net.ipv4.ip_unprivileged_port_start=0` on the host** (or `containerd_enable_unprivileged_ports: true` in Kubespray, which is actually what you ended up doing) — allows non-root to bind privileged ports in the host network namespace.

Both tested and working on k3s + Cilium (kube-proxy replacement) + network policy allowing only egress 443: port 443 passes, port 80 blocked — all with the standard image, no `setcap`.

Given the above, I believe this PR is not necessary — the solutions already exist at deploy time. WDYT?

rtsui-mk:
Hi @darkweaver87 
> We are security-first oriented.

Regards the two workarounds, both (1) and (2) seems to defeat the purpose.

1 would grant unnecessary privileges compared to ingress-nginx's configuration, and 2 grant not just the traefik pod, but everything on the host to bind to ports below 1024? 

https://github.com/kubernetes/ingress-nginx/blob/main/deploy/static/provider/baremetal/deploy.yaml

rtsui-mk:
Hey @mloiseleur
> Wdyt of setting this sysctl as you expect on your cluster?
> Something like this in the Deployment:
```
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    spec:
      securityContext:
        sysctls:
        - name: "net.ipv4.ip_unprivileged_port_start"
          value: "0"
```

This doesn't work even without the admission policy.

On my clean kubespray 2.30 setup with hostnetwork=true for traefik daemonset, I get the error below

```
 daemonsets.apps "traefik" was not valid:
# * spec.template.spec.securityContext.sysctls[0].name: Invalid value: "net.ipv4.ip_unprivileged_port_start": may not be specified when 'hostNetwork' is true
#
```

Sorry a bit confused.  Is the PR/suggested change supposed to be made in here instead

https://github.com/traefik/traefik-library-image/blob/v3.7.0-ea.3/v3.7/alpine/Dockerfile

rtsui-mk:
also confirmed `containerd_enable_unprivileged_ports: true` doesn't work, 
as described in containerd's doc it is not for hostnetwork

https://github.com/containerd/containerd/blob/main/docs/containerd-2.0.md#unprivileged-ports-and-icmp-by-default-for-cri

darkweaver87:
Thanks everyone for the thorough discussion — it helped clarify what's actually going on under the hood.

The root cause here is a known Kubernetes limitation, not a Traefik issue. When you set `securityContext.capabilities.add: ["NET_BIND_SERVICE"]`, containerd adds the capability to the bounding, effective, and permitted sets — but not the ambient set ([`pkg/oci/spec_opts.go:1084`](https://github.com/containerd/containerd/blob/main/pkg/oci/spec_opts.go#L1084)). Ambient capabilities are actually [explicitly cleared](https://github.com/containerd/containerd/blob/main/internal/cri/server/container_create.go#L847-L852) for all containers, referencing [kubernetes#56374](https://github.com/kubernetes/kubernetes/issues/56374). Without ambient capabilities, the kernel clears permitted+effective on `execve()` for non-root processes, so the capability is granted but unusable. On a traditional Linux server, systemd solves this with `AmbientCapabilities=CAP_NET_BIND_SERVICE`. The Kubernetes equivalent is [KEP-2763](https://github.com/kubernetes/enhancements/issues/2763), but it's been stalled since 2024 — containerd closed the implementation as ["not planned"](https://github.com/containerd/containerd/issues/5644).

The image size impact of adding `setcap` would be negligible (~64 KiB). However, file capabilities set via `setcap` require `allowPrivilegeEscalation: true` at runtime — the kernel's `NO_NEW_PRIVS` bit blocks them otherwise. This is true whether the `setcap` is baked into the image or applied at deploy time. For this reason, we'd rather not change the upstream image and instead provide a deploy-time workaround.

For users who need `hostNetwork` + non-root + privileged ports, here's an init container approach that doesn't require a custom image. Tested on k3s v1.34 with `ip_unprivileged_port_start=1024`:

```yaml
podSecurityContext:
  runAsGroup: 65532
  runAsNonRoot: false
  runAsUser: 65532
  seccompProfile:
    type: RuntimeDefault

securityContext:
  runAsNonRoot: true
  runAsUser: 65532
  allowPrivilegeEscalation: true
  capabilities:
    drop: [ALL]
    add: [NET_BIND_SERVICE]

hostNetwork: true

service:
  enabled: false

deployment:
  initContainers:
    - name: copy-binary
      image: traefik:v3.6.12
      command: ["cp", "/usr/local/bin/traefik", "/shared/traefik"]
      volumeMounts:
        - name: traefik-bin
          mountPath: /shared
    - name: setcap
      image: alpine:3.21
      command:
        - sh
        - -c
        - apk add --no-cache libcap && setcap cap_net_bind_service=+ep /shared/traefik
      securityContext:
        runAsUser: 0
        runAsNonRoot: false
      volumeMounts:
        - name: traefik-bin
          mountPath: /shared
  additionalVolumes:
    - name: traefik-bin
      emptyDir: {}

additionalVolumeMounts:
  - name: traefik-bin
    mountPath: /usr/local/bin
```

The init containers copy the traefik binary to an `emptyDir`, apply `setcap` as root, then mount it over `/usr/local/bin` so the default entrypoint picks it up. The main container still runs as non-root (65532).

@sathieu @ICHx would this approach work for your use cases?

sathieu:
@darkweaver87 

>  However, file capabilities set via setcap require allowPrivilegeEscalation: true at runtime 

This is not what I see. This works with:

- `Dockerfile`having `setcap cap_net_bind_service=+ep .../traefik` (i.e. this PR)
- `allowPrivilegeEscalation: false`
- `capabilities.add: [NET_BIND_SERVICE]`
- `ports.web.port=80` and `ports.websecure.port=443`

Maybe the process setting `NO_NEW_PRIVS` has `NET_BIND_SERVICE` capability?

darkweaver87:
@sathieu You're right, I was wrong about `NO_NEW_PRIVS` — thanks for the correction. `setcap` does work with `allowPrivilegeEscalation: false` when the capability is already in the permitted set :+1:

That said, I still think that this is a Kubernetes-level limitation, not something to work around in the default upstream image. Let's see what maintainers think about this :-)

rtsui-mk:
@darkweaver87 
Thank you for the reply.  No, the init container approach wouldn't work for me as some of our k8s deployment would be air gapped and apk wouldn't be available. 

My copilot coincidentally gave us the same piece of init container code.  It is working as expected though

mloiseleur:
> this is a Kubernetes-level limitation, not something to work around in the default upstream image. 

I truly agree with that. To me, this kind of system-level workaround for a specific use case should not be part of the default Traefik distribution.

rtsui-mk:
@mloiseleur I understand that this might not be for the default traefik image.  But this is still highly desirable for a k8s (a new specific image?) if traefik intends to be an ingress-nginx replacement

Thx

emilevauge:
Hey @sathieu, and thanks to everyone who contributed to this discussion — it was really thorough and helped clarify the underlying issue.

After reviewing this internally, we've decided not to merge this PR. Here's our reasoning:

**The core issue is a Kubernetes/containerd limitation, not a Traefik one.** Kubernetes doesn't propagate capabilities to the ambient set for non-root processes (see [kubernetes#56374](https://github.com/kubernetes/kubernetes/issues/56374) and the stalled [KEP-2763](https://github.com/kubernetes/enhancements/issues/2763)), which is why `securityContext.capabilities.add` alone isn't enough when `ip_unprivileged_port_start=1024`.

Baking `setcap` + `libcap` into the default image would change the security posture for *all* users to work around a platform-level limitation that only affects a subset of deployments. We'd rather keep the default image as minimal and least-privileged as possible.

**Deploy-time alternatives exist** that don't require changing the upstream image:
- Adjusting `containerd_enable_unprivileged_ports` (as you ended up doing via Kubespray)
- Using an init container to copy the binary and apply `setcap` at runtime (as @darkweaver87 demonstrated)
- Setting `ip_unprivileged_port_start=0` on the host for hostNetwork scenarios

We've also opened [traefik/traefik-helm-chart#1781](https://github.com/traefik/traefik-helm-chart/issues/1781) to improve documentation on how to bind to privileged ports, this should make it easier for others hitting the same issue.

We had already declined a similar change in [traefik-library-image#51](https://github.com/traefik/traefik-library-image/pull/51) for the same reasons, so we want to stay consistent on this.

Thanks again for the contribution and the detailed context, it's genuinely useful and will help us improve our docs. Closing this one out.

rtsui-mk:
sorry this is a bit misleading, I mentioned containerd_enable_unprivileged_ports is not working for hostnetwork in my comment
but thx

PR #12908, opened 2026-03-29
Support nginx.ingress.kubernetes.io/enable-access-log annotation

ris-tlp:
### What does this PR do?

- Completes part of #12631. 
- Implemented the `nginx.ingress.kubernetes.io/enable-access-log` annotation

### Motivation

#12631 

### More

- [x] Added/updated tests

### Additional Notes
When the annotation is unset, the router’s default/inherited access-log behavior is kept. If set to `true`, access logs are explicitly enabled for the router, and when set to `false`, they are explicitly disabled. This only controls per-router access logging.


kevinpollet:
Hey @ris-tlp,

Thanks for this contribution!

PR #12913, opened 2026-03-30
Add Kubernetes Ingress logs fields

rtribotte:
### What does this PR do?

This PR adds 4 new Kubernetes Ingress logs fields to the JSON access log format.

To do so, it introduces a new metadata struct type to the router observability configuration, that is unexposed to user configuration, but could be used by providers to add context.

### Motivation

Superseedes #12721

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
@@ -5388,6 +5388,14 @@ func TestLoadIngresses(t *testing.T) {
 							Service:     "default-ingress-with-access-log-enabled-whoami-80",
 							Observability: &dynamic.RouterObservabilityConfig{
 								AccessLogs: ptr.To(true),
+								Metadata: &dynamic.ObservabilityMetadata{
+									Ingress: &dynamic.KubernetesIngressMetadata{
+										Namespace:   "default",
+										IngressName: "ingress-with-access-log-enabled",
+										ServiceName: "whoami",
+										ServicePort: "80",
+									},
+								},
 							},
 						},
 						"default-ingress-with-access-log-enabled-rule-0-path-0-tls": {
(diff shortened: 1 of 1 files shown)


pkg/config/dynamic/http_config.go
```diff
@@ -177,6 +181,23 @@ func (r *RouterObservabilityConfig) SetDefaults() {
 
 // +k8s:deepcopy-gen=true
 
+// ObservabilityMetadata holds the observability metadata configuration.
+type ObservabilityMetadata struct {
+	Ingress *KubernetesIngressMetadata `json:"ingress,omitempty" toml:"-" yaml:"-" label:"-" file:"-" kv:"-"`
```
sathieu, line 186: This could be useful for other resource kinds:

- Gateway API Gateways, HTTPRoutes, TCPRoutes, ...
- Traefik's IngressRoute, IngressRouteTCP, ...

So, maybe the Ingress term is too small?

pkg/config/dynamic/http_config.go
```diff
@@ -177,6 +181,23 @@ func (r *RouterObservabilityConfig) SetDefaults() {
...
+}
+
+// +k8s:deepcopy-gen=true
+
+// KubernetesIngressMetadata holds the Kubernetes Ingress metadata.
+type KubernetesIngressMetadata struct {
+	Namespace   string `json:"namespace,omitempty"`
+	IngressName string `json:"ingressName,omitempty"`
```
sathieu, line 194: Same remark as above.

Two ways to solve this:

- using two fields : Kind and Name
- using one field per type : IngressName, HTTPRouteName, IngressRouteName, ...

pkg/config/dynamic/http_config.go
rtribotte: Hi @sathieu,

Indeed, and we purposely specialized it in a KubernetesIngressMetadata struct, so that later on for instance for Gateway API we would need to create a dedicated struct.

pkg/config/dynamic/http_config.go
rtribotte: Please see my comment above!

PR #12914, opened 2026-03-31
Add optional X-Forwarded-Scheme and X-Scheme headers in forwarded headers middleware

### LINKED ISSUE #12903: X-Forwarded-Scheme: https and  X-Scheme: https headers
### Welcome!

While moving from ingress-nginx, we lost the two following headers when the frontend is https and the backend is http:

- `X-Forwarded-Scheme: https`
- `X-Scheme: https`

This is used by Spring ([ref](https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilter.html)) to use proper redirect URI on OIDC login.

qwerty8811:
### What does this PR do?

Adds optional support for emitting the legacy `X-Forwarded-Scheme` and `X-Scheme` headers from Traefik's core forwarded headers middleware at the entrypoint level.

When `entryPoints.<name>.forwardedHeaders.addXForwardedSchemeHeaders=true` is enabled, Traefik sets these compatibility headers alongside `X-Forwarded-Proto` using the same resolved scheme value.

This allows applications that still depend on those legacy headers to keep working consistently across providers during migration to Traefik.

### Motivation

Closes #12903

### More

- [x] Added/updated tests


kevinpollet:
Hey @qwerty8811,

rtribotte:
Hi @qwerty8811,

While we understand the approach you have taken, we believe the fix should be implemented at a different level.

Traefik already sets the X-Forwarded-Proto header for all TLS connections in the core forwarded headers middleware (pkg/middlewares/forwardedheaders/forwarded_header.go), at the 'entrypoint' level.
As X-Forwarded-Scheme and X-Scheme essentially contain the same information under different header names, they should be set next to X-Forwarded-Proto in the same middleware.

Setting them in the ingress-nginx provider would mean that anyone migrating to Traefik through another provider (IngressRoute, Docker, File, etc.) would still encounter the same issue.
Making it part of the core means it works everywhere and for everyone.

Could you therefore move the change to the forwarded headers middleware? This would mean that a single entrypoint-level change would cover all providers, and we wouldn't need per-route middleware injection.
We could also use a feature flag to control the addition of those headers.

Additionally, the migration documentation should be improved in this regard to indicate the existence of the flag.

If you would prefer not to make these changes, we could handle superseding this PR.

Thanks!

qwerty8811:
Hi @rtribotte,

Thanks for the review. I moved the legacy `X-Forwarded-Scheme` / `X-Scheme` handling into the core forwarded headers middleware at the entrypoint level, added an opt-in `forwardedHeaders.addXForwardedSchemeHeaders` flag, removed the ingress-nginx-specific middleware injection, and updated the migration/docs notes. Please take another look when you have time.

pkg/middlewares/forwardedheaders/forwarded_header.go
```diff
@@ -168,6 +174,14 @@ func (x *XForwarded) rewrite(outreq *http.Request) {
 		unsafeHeader(outreq.Header).Set(XForwardedPort, forwardedPort(outreq))
 	}
 
+	if x.addXForwardedSchemeHeaders {
+		scheme := unsafeHeader(outreq.Header).Get(XForwardedProto)
+		if scheme != "" {
```
youkoulayley: XForwardedProto is always set before (L153-171), so this condition is unnecessary.

pkg/middlewares/forwardedheaders/forwarded_header_test.go
```diff
@@ -257,6 +289,21 @@ func TestServeHTTP(t *testing.T) {
 				XForwardedProto: "wss",
 			},
 		},
+		{
+			desc:             "xForwardedScheme headers overwrite trusted values",
```
youkoulayley:
```suggestion
			desc:             "xForwardedScheme headers overwrite in insecure mode",
```

pkg/middlewares/forwardedheaders/forwarded_header_test.go
youkoulayley, line 1: The tests only iterates over `expectedHeaders` and checks values, it never asserts that unlisted headers are absent. 
Could you add a test that verifies that if `addSchemeHeaders` is false, the `xForwardedScheme` and `xScheme` are not present?

youkoulayley (changes requested):

docs/content/reference/install-configuration/configuration-options.md
youkoulayley, line 1: There is also an error on the CI related to this file. Could you run `make generate` to fix this?

pkg/middlewares/forwardedheaders/forwarded_header.go
qwerty8811: removed

pkg/middlewares/forwardedheaders/forwarded_header_test.go
qwerty8811: updated

pkg/middlewares/forwardedheaders/forwarded_header_test.go
qwerty8811: makes sense, added a dedicated test case with addSchemeHeaders disabled. also extended the test table with an absentHeaders field.

docs/content/reference/install-configuration/configuration-options.md
qwerty8811: done

youkoulayley (approved):
LGTM 👍

gndz07 (approved):
LGTM

PR #12918, opened 2026-04-01
Support nginx.ingress.kubernetes.io/limit-allowlist annotation

foxcool:
### What does this PR do?

Adds support for the `nginx.ingress.kubernetes.io/limit-allowlist` annotation, and for its legacy alias `nginx.ingress.kubernetes.io/limit-whitelist`, in the Kubernetes Ingress NGINX provider. The annotation takes a comma-separated list of IPs and CIDR ranges exempted from the limits of the Ingress path.

Rather than teaching the limit middlewares about exempted clients, the provider builds a companion router next to every router it creates for the path: the same rule extended with `ClientIP(...)`, the same service, and the same middleware chain minus `limit-rps`, `limit-rpm` and `limit-connections`. An allowlisted request never reaches a limiter.

The companion router is pinned to the priority of the router it exempts, plus one, so its longer rule does not outrank the other routers of the host.

### Motivation

Part of #12631.

In NGINX the allowlist empties the counting key shared by `limit_req` and `limit_conn`, which exempts all three limits at once. A router carrying a chain without the limits reproduces that, `limit-connections` included, and `ClientIP` matches the connection remote address just like `geo $remote_addr` does.

The first revision of this PR added an `excludedIPs` option to the RateLimit middleware instead. It was dropped after the design review: it covered two limits out of three, it collided with the existing `sourceCriterion.ipStrategy.excludedIPs`, and the same option had already been declined in #7911 because exempting a set of IPs from a middleware belongs to the router. Nothing outside the provider is touched now: no new dynamic configuration and no CRD change.

### More

- [x] Added/updated tests

### Additional Notes

The cost of the approach, as raised during the review:

- one extra router per router of an annotated path, so up to twice the routers on a path that also has TLS and a canary;
- exempted traffic is reported in the metrics under the companion router name, which splits the graphs of that path in two.

Pinning the priority leaves one narrow case open: the companion router still ties with a router whose rule is exactly one character longer, and ties fall back to provider precedence. Closing it entirely would mean touching how the muxer sorts, which is out of scope here. Happy to open a separate issue if you think it is worth it.


nmengin:
Hey @foxcool,

We've set the status to "design-review" to allow us to check the PR.
There are internal discussions about the best way to tackle this topic, and we need more time to dig on it.

For this reason, we won't be able to merge it for v3.7.

foxcool:
Hi! I see you mermeg 3.7 and updated this branch.

rubarcelk:
Hey @nmengin , are there any updates on this?

nmengin:
Hey @rubarcelk,

sdelicata:
Hi @foxcool! 👋 Thanks a lot for the contribution, and sorry this one has been sitting for so long.

Since we flagged it for design review, my feedback is on the shape rather than the code.

What bugs me is that we're adding a public option to the RateLimit middleware to answer what is really a routing question: which requests are subject to this location's limits. Two things follow from that:

- In nginx the allowlist also exempts `limit-connections`, not only `limit-rpm` / `limit-rps`: the three limits share the same counting key, and the allowlist empties it (an empty key means nginx doesn't count at all). So an option on `RateLimit` only covers two of the three, and we'd need the same one on `InFlightReq` right after.
- `excludedIPs` collides with the `sourceCriterion.ipStrategy.excludedIPs` we already have on that same middleware, which is about finding the client IP in `X-Forwarded-For`, not about exempting anyone.

We can already express this with a bypass router, and this provider does exactly that for canary in `translator.go`: same rule plus `&& ClientIP("10.0.0.0/24", "192.168.1.1")`, and the middleware stack applied without the limits. The longer rule wins on its own since the default priority is the rule length, and the `ClientIP` matcher resolves on `RemoteAddr` through `ip.Checker`, so we get the same semantics and the same CIDR syntax as nginx's `geo $remote_addr`. No new dynamic config, no CRD change, and `limit-connections` comes for free.

That's not a new conclusion either: the same option was already proposed and [declined](https://github.com/traefik/traefik/pull/7911#issuecomment-806707612) because the need would spread to every middleware, and because bypassing one for a set of IPs belongs to the router and its middleware chain. @rtribotte

The cost is more routers on the locations carrying the annotation, plus allowlisted traffic showing up under a second router name in the metrics.

I'd suggest keeping this PR to the annotation itself, implemented with a bypass router (and handling `limit-allowlist`, which is the canonical name, `limit-whitelist` being only its alias in ingress-nginx), and removing the generic `excludedIPs` option.

foxcool:
Hi @sdelicata, thanks for fair points, especially the shared counting key. Reworked as suggested: ExcludedIPs is gone, the diff now only touches the ingress-nginx provider and its docs. The annotation builds a bypass router the same way canary does, so limit-connections is covered too. limit-allowlist is the canonical name, limit-whitelist the alias.

One thing worth flagging: with a single bypass router, a long CIDR list could make its rule longer than the canary rule and steal allowlisted canary traffic. So each router of the location gets its own allowlist twin built from its own rule -- rule-length ordering then just works, no explicit priorities (except the default backend, where priority is pinned and the twin gets +1). Worst case is 8 routers per location; there's a fixture asserting that.                                                        
The ClientIP-matches-RemoteAddr behavior and the metrics split are documented as-is, matching nginx.

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -405,6 +422,41 @@ func buildSticky(cfg IngressConfig, nameSuffix string) *dynamic.Sticky {
...
+		TLS:           rt.TLS,
+		Observability: rt.Observability,
+	}
+
+	// The default backend routers pin an explicit lowest priority: their allowlist
+	// router has to stay below every other router, while still winning over its own
+	// base router. Elsewhere the default priority (the rule length) already does it.
+	if rt.Priority != 0 {
+		allowlistRouter.Priority = rt.Priority + 1
+	}
```
sdelicata, line 452: I'm afraid the rule-length ordering doesn't fully work here. Wrapping the rule and appending `ClientIP(...)` adds ~29 chars, so the twin doesn't just outrank its own base router, it outranks any router on the same host whose rule is up to 29 chars longer.

Two plain Ingresses on `app.example.com`, actual lengths of the generated rules:

```
Host("app.example.com") && PathPrefix("/")                              ->  42   front, has the limits
Host("app.example.com") && Path("/api/v1/users")                        ->  48   api, no annotations
(Host("app.example.com") && PathPrefix("/")) && ClientIP("10.0.0.0/24") ->  71   front's twin
```

A request to `/api/v1/users` goes to the api service for everyone, and to the front service for allowlisted clients only. Nothing is logged, and nginx has no such side effect since the geo/map pair plays no part in location selection.

The default backend branch just below already does the right thing. WDYT about applying it everywhere and dropping the special case?

```suggestion
	priority := rt.Priority
	if priority == 0 {
		priority = httpmuxer.GetRulePriority(rt.Rule)
	}
	allowlistRouter.Priority = priority + 1
```

`GetRulePriority` is exported and providers already import the muxer (`acme`, `tailscale`). One thing to keep in mind: `+1` can still tie with a router whose rule is exactly one char longer, and ties fall back to provider precedence then an unstable sort. Much narrower than today, but worth a thought.

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -16892,6 +17155,423 @@ func TestLoadIngresses(t *testing.T) {
...
+								IdleConnTimeout: ptypes.Duration(60 * time.Second),
+							},
+						},
+					},
+				},
+				TLS: &dynamic.TLSConfiguration{},
+			},
+		},
+		{
+			desc: "Limit allowlist without any limit",
```
sdelicata, line 17479: Could you add a case with a `spec.defaultBackend` Ingress carrying `limit-allowlist` please? That's the only path where the priority is computed by hand (`math.MinInt32 + 1`), and it's currently untested.

It becomes more useful once the priorities are pinned everywhere: asserting the exact `Priority` on the twins is what would catch a later refactor silently bringing the ordering problem back.

sdelicata (changes requested):
Thanks @foxcool for all of your changes!
Here are my second review feedbacks.

pkg/provider/kubernetes/ingress-nginx/translator.go
foxcool: Thank you, I agree with both points. You are right about the priorities: I was resolving a collision within a single location and missed the fact that the muxer sorts all entrypoint routers into a single list.

I applied the suggestion and removed the special case for the default backend. One deviation: I moved the applyLimitAllowlist call inside applyMiddlewares instead of eight calls in translate.

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
foxcool: Yes. Case added. The red Unit tests are not from this diff. Could you please rerun the check?

sdelicata (approved):
LGTM
Thx for your contribution @foxcool!

amazon7737 (approved):
LGTM 👍

PR #12920, opened 2026-04-01
Add reportNodeInternalIPs option to report node internal IPs in Ingress status

### LINKED ISSUE #12897: Feature Request: Support reporting Node Internal IPs for Ingress status (bare-metal Kubernetes)
### Welcome!

### Background

In bare-metal Kubernetes environments (on-prem), it is common to run ingress controllers without a cloud LoadBalancer or LoadBalancer controller.

With `ingress-nginx` (now deprecated/EOL), this use case was supported via the flag:

```
--report-node-internal-ip-address=true
```
Or the equivalent Helm value: https://github.com/kubernetes/ingress-nginx/blob/dbb11b92ddb77bee9c35e462479129354c84f939/charts/ingress-nginx/values.yaml#L83-L85

This allowed the controller (typically running as a DaemonSet) to populate the `status.loadBalancer.ingress` field of Ingress resources with the internal IPs of the nodes running the ingress pods.

This made it straightforward to:

* Access services via node IPs
* Integrate with external DNS solutions
* Support environments without external load balancers

### Problem

Traefik currently does not provide an equivalent mechanism to automatically report node internal IPs in the Ingress status.

In bare-metal setups, this results in:

* Empty or incorrect `status.loadBalancer` fields
* Additional manual work or custom controllers to update Ingress status
* Friction when migrating from ingress-nginx to Traefik

### Proposal

Add support for a configuration option (flag and/or Helm value) that allows Traefik to:

* Detect the node(s) it is running on (especially in DaemonSet mode)
* Retrieve the node internal IP(s)
* Populate the `status.loadBalancer.ingress` field of managed Ingress resources accordingly

Example:

```
--providers.kubernetesIngress.reportNodeInternalIP=true
```

or similar configuration via Helm values.

### Expected Behavior

When enabled:

* Traefik updates Ingress status with the internal IPs of nodes running Traefik pods
* Works especially well when deployed as a DaemonSet
* Mirrors the behavior previously available in ingress-nginx

### Use Case

* Bare-metal / on-prem Kubernetes clusters
* No cloud provider integration
* No LoadBalancer controller (e.g., MetalLB not used)
* Migration path from ingress-nginx to Traefik

### Additional Context

This feature would significantly improve Traefik’s usability in non-cloud environments and reduce friction for users transitioning from ingress-nginx.

Happy to help test or provide more details if needed.

#### References

1. https://kubernetes.github.io/ingress-nginx/deploy/baremetal/#via-the-host-network
2. https://www.marketwatch.com/press-release/traefik-becomes-the-de-facto-standard-for-kubernetes-networking-as-major-platform-vendors-migrate-from-ingress-nginx-65d83e09

qwerty8811:
### What does this PR do?

Adds a `reportNodeInternalIPs` option to the Kubernetes Ingress provider that populates the `status.loadBalancer.ingress` field of each managed Ingress resource with the internal IPs of all cluster nodes. This is the equivalent of ingress-nginx's `--report-node-internal-ip-address` flag, intended for bare-metal deployments where Traefik runs as a DaemonSet without a cloud LoadBalancer or MetalLB.

The option is validated at startup (`Init()`) and is mutually exclusive with `ingressEndpoint` and `disableClusterScopeResources`. An error is returned if no nodes are found or if no node has an `InternalIP` address.

### Motivation
Closes #12897

### More

- [x] Added/updated tests


rtribotte:
Hey @qwerty8811,

Thanks for this contribution!

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -406,6 +415,34 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 }
 
 func (p *Provider) updateIngressStatus(ing *netv1.Ingress, k8sClient Client) error {
+	if p.ReportNodeInternalIPs {
+		nodes, nodesExists, err := k8sClient.GetNodes()
+		if err != nil {
+			return fmt.Errorf("getting nodes: %w", err)
+		}
+
+		if !nodesExists || len(nodes) == 0 {
```
youkoulayley:
```suggestion
		if len(nodes) == 0 {
```

I think this condition can be simplified.

pkg/provider/kubernetes/ingress/fixtures/Node-Internal-IP.yml
```diff
@@ -0,0 +1,74 @@
...
+  ports:
+    - port: 80
+
+  clusterIP: 10.0.0.1
+
+---
+kind: EndpointSlice
+apiVersion: discovery.k8s.io/v1
+metadata:
+  name: service1-abc
```
youkoulayley, line 62:
```suggestion
  name: service1-abc
  namespace: default
```

pkg/provider/kubernetes/ingress/kubernetes_test.go
```diff
@@ -2379,6 +2379,114 @@ func readResources(t *testing.T, paths []string) []runtime.Object {
...
+		ReportNodeInternalIPs:        true,
+		DisableClusterScopeResources: true,
+	}
+	assert.EqualError(t, p2.Init(), "reportNodeInternalIPs and disableClusterScopeResources are mutually exclusive")
+
+	p3 := Provider{ReportNodeInternalIPs: true}
+	assert.NoError(t, p3.Init())
+}
+
+func TestIngressEndpointReportNodeInternalIPs(t *testing.T) {
```
youkoulayley: This test doesn't need testcases.

youkoulayley (changes requested):

pkg/provider/kubernetes/ingress/kubernetes_test.go
qwerty8811: updated

pkg/provider/kubernetes/ingress/fixtures/Node-Internal-IP.yml
qwerty8811: default namespace added

pkg/provider/kubernetes/ingress/kubernetes.go
qwerty8811: done

qwerty8811:
Hello @youkoulayley 
Thank you for the feedback. I have addressed your comments. Please let me know if any other changes are needed.

pkg/provider/kubernetes/ingress/kubernetes_test.go
```diff
@@ -3069,6 +3069,100 @@ func readResources(t *testing.T, paths []string) []runtime.Object {
...
+	t.Parallel()
+
```
youkoulayley:
```suggestion
```

youkoulayley (changes requested):

pkg/provider/kubernetes/ingress/kubernetes_test.go
qwerty8811: done

youkoulayley (approved):
LGTM 👍

pkg/provider/kubernetes/ingress/kubernetes.go
```diff
@@ -436,6 +445,34 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
...
+					ingressStatus = append(ingressStatus, netv1.IngressLoadBalancerIngress{
+						IP: address.Address,
+					})
+				}
+			}
+		}
+
+		if len(ingressStatus) == 0 {
+			return errors.New("no nodes with internal IP address found")
+		}
```
gndz07: WDYT of moving the `ingressStatus[]` compilation out of `updateIngressStatus` and computing it once before the ingress loop? That way we avoid recalculating it per ingress, and the result can be passed in as a parameter.

gndz07 (changes requested):

gndz07 (approved):
LGTM

traefiker:

Issue #12500, opened 2026-01-02
Gateway API: multiple Gateways backed by different Services with different IPs/hostnames


daggaz:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

This is related to #12076 but I think more general in scope.

## Summary

When running Traefik as a Kubernetes Gateway API controller with **multiple `Gateway` objects**, Traefik currently reports the **same address** in `.status.addresses` for all Gateways, as configured statically through the `providers.kubernetesGateway.
statusAddress.*` CLI arguments.

For example if traefik is configured with:
```
--providers.kubernetesgateway.statusaddress.service.name=bar
--providers.kubernetesgateway.statusaddress.service.namespace=foo
```

It will determine the status addresses for _all_ `Gateways` using the IP of that one named `Service`. The same is true for the `hostname/ip` options.

This means we cannot create a 2nd `Gateway` (e.g. `internal` and `public`) and have them appear on different IPs.

For example, I have configured traefik with two entrypoints (`websecure:8443` and `public:8444`) that are exposed via different `Services` or type `LoadBalancer` (mapped to 443 on each external IP), but I cannot create two `Gateway` configurations, each referencing 8443 and 8444 respectively, and get the non-statically-configured IP of the corresponding services.

The Gateway API explicitly supports:

- Multiple Gateways per cluster
- Different exposure models (internal, public, edge, etc.)
- Distinct addresses per Gateway

## Observed context

- Traefik runs once (DaemonSet or Deployment)
- Multiple LoadBalancer `Services` front the same pods:
  - `Service/traefik`: 443 -> 8443
  - `Service/traefik-public`: 443 -> 8444
- Multiple entryPoints exist:
  - `websecure`: 8443
  - `public`: 8444
- Multiple Gateways:
  - `internal`: port 8443
  - `public`: port 8444

Traefik already has access to:
- Gateway listener ports
- EntryPoint configuration
- Service definitions and selectors

so should have enough information to lookup the correct `Service` (correct labels and targetport) and publish that `Service`'s IP(s) as the `Gateway` address.

## Relevant code

Gateway status addresses are currently derived from a single configured Service:

- `Provider.gatewayAddresses()`  
[...]

PR #12921, opened 2026-04-01
feat(gateway): support per-Gateway status address annotations

jscaltreto:
### What does this PR do?

Adds per-Gateway status address annotations that allow overriding the static `statusAddress` configuration on individual `Gateway` resources. This enables multiple Gateways backed by different Services (e.g. internal vs public) to report distinct addresses in `.status.addresses`.

### Motivation

When running Traefik with multiple `Gateway` objects fronted by different `LoadBalancer` Services, all Gateways currently report the same address in `.status.addresses` because the `statusAddress` is a single static configuration. This breaks tools like external-dns that read Gateway status to create DNS records, since every Gateway gets the same IP regardless of which Service actually fronts it. See #12500 for the full discussion.

Rather than an auto-discovery mechanism (which would require heuristic matching of listener ports to Services), this PR takes an explicit annotation-based approach that reuses the existing annotation infrastructure (`convertAnnotations` + `label.Decode`) and mirrors the static config options:

- `traefik.io/statusaddress.ip` - direct IP override
- `traefik.io/statusaddress.hostname` - direct hostname override
- `traefik.io/statusaddress.service.name` + `traefik.io/statusaddress.service.namespace` - copy addresses from a specific Service's LoadBalancer status

If `service.namespace` is omitted, it defaults to the Gateway's own namespace. Cross-namespace service references require a `ReferenceGrant` in the target namespace, consistent with how TLS certificate cross-namespace refs are already handled in this provider. When no annotations are present, behavior is identical to before (fully backward compatible).

### More

- [x] Added/updated tests

### Additional Notes

- The `resolveStatusAddress` method was extracted from the former `gatewayAddresses` to make the core resolution logic reusable for both static config and annotation-sourced overrides.
- Partial `ServiceRef` (name without namespace or vice versa) now returns a clear error message instead of the misleading "empty Gateway status address configuration".
- If annotation-based address resolution fails for a Gateway, the status update for that Gateway is skipped entirely rather than silently falling back to the default address.

Ref: #12500

### DIFF
--- a/docs/content/reference/routing-configuration/kubernetes/gateway-api.md
+++ b/docs/content/reference/routing-configuration/kubernetes/gateway-api.md
@@ -734,6 +734,61 @@ Once everything is deployed, sending the WHO command should return the following
     IP: fe80::d873:20ff:fef5:be86
     ```
 
+## Per-Gateway Status Addresses
+
+By default, Traefik copies the same status addresses (configured via `statusAddress`) to all `Gateway` resources it manages.
+When running multiple `Gateway` objects backed by different `Service` resources (e.g. internal and public), each `Gateway` needs its own address in `.status.addresses` so that tools like external-dns can resolve the correct IP per `Gateway`.
+
+It is possible to override the static `statusAddress` configuration on a per-`Gateway` basis using annotations.
+
+The following annotations are supported on `Gateway` resources:
+
+| Annotation | Description |
+|:-----------|:------------|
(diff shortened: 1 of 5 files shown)


kevinpollet:
Hey @jscaltreto,

jscaltreto:
Thanks @kevinpollet! Just following-up to check on the status of the design review or if there was anything else you needed from me at this point. Thanks, again!

PR #12932, opened 2026-04-03
Add ipAllowListStrategy option for allowlist/whitelist annotations

### LINKED ISSUE #12928: Ingress NGINX Annotation Support - IP Allowlists / Whitelists and Proxies
### Welcome!

Like many of you, our team is currently embarking on a migration from ingress-nginx to Traefik, and we're experiencing a blocker with the support that's been added for our most prevalent annotation  - `nginx.ingress.kubernetes.io/whitelist-source-range`. Almost all of our hosts sit behind a Cloudflare proxy, so client addresses appear as they are coming from one of the Cloudflare IP ranges.

In ingress-nginx, we can accommodate Cloudflare by using a variety of global [configuration options](https://community.cloudflare.com/t/nginx-ingress-whitelisting-doesnt-work-with-proxy-enabled/197282/3) so the "real client IP" is analyzed when it comes to these whitelist ranges.

In Traefik, it has converted these annotations to an IPAllowList middleware, but I have not found a way to add the necessary ipStrategy.Depth configuration (with a value of 1) to allow it to match our IP ranges against the X-Forwarded-For header. I have seen [solutions](https://blog.lrvt.de/solving-traefiks-ipstrategy-dilemma-while-using-cdns/) of how to resolve this when using native Traefik IPAllowList middleware, and it looks like the traefik-warp plugin will prove to be very useful for us, but our goal is to not refactor hundreds of Ingress configurations as middlewares if we don't have to.

Do you think there would be any way to manipulate ipStrategy.Depth values for these annotations that are being converted to middlewares behind the scenes? Ideally, we could just have a global configuration flag that directs IPAllowList middlewares to, by default, look at the X-Forwarded-For header. Since the traefik-warp plugin is sanitizing and appending non-proxied client addresses to that header, that would be the key to unblocking from our migration without a need for refactoring hundreds of Ingresses.

mathieuherbert:
### What does this PR do?

Add a global configuration field `ipAllowListDepth` under `kubernetesingressnginx` provider that configure the ipAllowListDepth when `[nginx.ingress.kubernetes.io/whitelist-source-range](http://nginx.ingress.kubernetes.io/whitelist-source-range)` is enabled. It does not change the default behaviour.

### Motivation
When using Traefik's `kubernetesingressnginx` provider, the
[nginx.ingress.kubernetes.io/whitelist-source-range](http://nginx.ingress.kubernetes.io/whitelist-source-range) annotation is translated into an `ipAllowList` middleware **without** `ipStrategy`. Traefik then checks `RemoteAddr` (the internal LB IP) instead of the real client IP from `X-Forwarded-For`, making the whitelist ineffective.

### More

- [X] Added/updated tests

### Additional Notes

Fixes #12928


rtribotte:
Hey @mathieuherbert,

Thanks for this contribution!

mathieuherbert:
@rtribotte any chance we can have it review for the next version ?
Thanks !

Persi:
Looks good for me.

rtribotte:
Hello @mathieuherbert,

> @rtribotte any chance we can have it review for the next version ?

We took a look at the changes, and while we understand the intent, we think the solution may lack thoroughness. For instance, it seems that with ingress-nginx, the whitelist-source-range is checked against the remote_addr variable, which is rewritten by the realip module and resolved thanks to the trusted-ips strategy, and also used in variable interpolation, observability, etc...

This raises questions about whether this solution (fixed depth) would be sufficient. To be transparent with you, we don't currently think that this PR will make it into v3.7.0 GA.

mathieuherbert:
Hi @rtribotte ,
Thank you for the feedback. 
It is true that the check is done differently between Traefik and Ingress-Nginx. This lack of compatibility on the whitelist-source-range IP makes our migration to Traefik via the ingress-nginx annotation incompatible for the moment.
If you think another approach is needed, do not hesitate to close this PR.
thanks!

bblier:
I won't argue that there could be a more thorough solution out there, but with ingress-nginx already archived, we are looking for at least a stop-gap solution to move to before it becomes a critical need. Without some ability to manipulate ipStrategy settings, the annotation support is effectively useless for all users of Cloudflare (or other proxy/CDN services). As it stands, we are blocked from migrating without changing all Ingress classes to Traefik, and changing annotations to their middleware counterparts.

I understand the amount of work that has been done for this project over the last few months but I need to know if this PR does not make it into 3.7.0 GA, is there an opportunity for it to make it in a patch release for 3.7.x, are we stuck until a 3.8.0 release, or is there no interest at all in making the annotation support work for us?

Could you at least provide some more detailed feedback as to what would make this PR more suitable solution?

luisdeniz:
I'd also like to point out that the proposed solution in this PR would unblock us to use Traefik as a replacement for Ingress NGINX. We deploy Traefik behind an AWS ALB — similar to how we deploy Ingress NGINX today — and have also found that the default `ipStrategy` behavior is a show-stopper for us, since we heavily rely on the `whitelist-source-range` annotation in production environments. Having at least the ability to customize the `ipAllowListDepth` value at the provider level would work for our use case. 

There may be more elegant solutions, as others have pointed out, but we would appreciate it if you could consider this implementation to allow us to move away from Ingress NGINX as soon as possible.

dfeinblatt:
@rtribotte @nmengin Following up on this PR echoing what @bblier and @luisdeniz have said. This functionality is necessary for us to be able to move off ingress-nginx, so we are currently blocked from migrating. Given that ingress-nginx is already archived, being unblocked as soon as possible is vital. I think we are all in agreement that there can be improvement added to this solution, but it provides meaningful and necessary value now that makes it worth including and it can be iterated on thereafter. I'd request you to strongly consider expediting the review and approval accordingly given the critical nature of this functionality. Thank you for your consideration.

schneider-f:
This functionality is critical for our migration as well. We’re currently blocked and would greatly appreciate expediting the review and approval. Thank you for pushing this forward!

kevinpollet:
Hello,

After further discussion, we decided to introduce this option at the provider level, allowing the configuration of all related strategies (depth, excludedIPs, and IPv6Subnet).

Since Traefik PR #13062 has now been merged, I have rebased this pull request accordingly to incorporate those changes.

kevinpollet (approved):
Thanks 👍

juliens (approved):
LGTM

PR #12969, opened 2026-04-13
Fix custom headers annotation with 503 Service Unavailable

### LINKED ISSUE #12737: [Nginx] Traefik might break when configmaps are non-existent but used in 'custom-headers'
### Welcome!

### What did you do?

See discussion here: https://github.com/traefik/traefik/pull/12414

Nginx renders a 503 when a configmap is not existent on the cluster and the ingress is using a configmap for 'custom-headers'. According to the discussion, it is possible that traefik is not able to properly apply the configmap when it does not exist. It might be needed to render a placeholder when the configmap does not exist.

Please note aswell: nginx does not reload to replace the 503 entry when the configmap becomes available. This would be very much desired, to not have to restart the controller to remove the 503 and apply the actual route.

Please note that this is not confirmed behaviour, but reasoned behaviour. 

### What did you see instead?

-

-

-

-

LBF38:
### What does this PR do?

This PR fixes the custom headers annotation in the ingress-nginx provider. It returns a 503 Service Unavailable when there is a configuration error with this annotation.

### Motivation

Feature parity with ingress-nginx.
Fixes https://github.com/traefik/traefik/issues/12737

### More

- [x] Added/updated tests

### Additional Notes

See also the minimal [e2e tests](https://github.com/traefik/ingress-nginx-migration/pull/20) (migration tool) for this fix.

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
@@ -635,6 +635,9 @@ func (p *Provider) loadConfiguration(ctx context.Context) *dynamic.Configuration
 
 			if err := p.applyMiddlewares(ingress, defaultBackendName, "", "", ingress.Spec.DefaultBackend, hosts, rt, conf, ""); err != nil {
 				logger.Error().Err(err).Msg("Error applying middlewares")
+				if strings.Contains(err.Error(), "custom headers") {
+					unavailableService(rt, conf)
+				}
 			}
 
 			conf.HTTP.Routers[defaultBackendName] = rt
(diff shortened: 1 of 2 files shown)


mwoudstra:
Question. In the original issue, it was mentioned that

`Please note aswell: nginx does not reload to replace the 503 entry when the configmap becomes available. This would be very much desired, to not have to restart the controller to remove the 503 and apply the actual route.`

Is this in scope of this PR? If not, is it in scope to fix for the issue, or should I create a followup after this is merged?

LBF38:
Hi @mwoudstra,

By default, Traefik reloads the dynamic configuration based on Kubernetes events. Therefore, it will reload the router when the configmap becomes available.

However, there is a known limitation regarding the static configuration. If the custom headers config map defines headers that are not allowed by the [`global-allowed-response-headers`](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#global-allowed-response-headers), you will need to update Traefik's static configuration and specify them in the [provider's config](https://doc.traefik.io/traefik/v3.7/reference/install-configuration/providers/kubernetes/kubernetes-ingress-nginx/#opt-providers-kubernetesIngressNGINX-globalAllowedResponseHeaders).

Feel free to open an issue or share more context if you need to.

juliens (approved):
LGTM

kevinpollet (approved):

PR #12972, opened 2026-04-14
fix(ratelimiter): support FIPS 140-3 mode via server-side script hashing

chaitanyabodlapati:
### What does this PR do?

Fixes a startup panic when Traefik runs in strict FIPS 140-3 mode (`GODEBUG=fips140=only`). The rate limiter middleware calls `redis.NewScript()` at package init time, which internally uses `crypto/sha1`. In FIPS mode, SHA-1 is blocked, causing Traefik to panic before it can serve any requests.

This PR switches the rate limiter's Lua script construction to use `redis.NewScriptServerSHA()` when FIPS mode is detected at runtime. That function (added in [redis/go-redis#3700](https://github.com/redis/go-redis/pull/3700), merged upstream) avoids client-side SHA-1 by using Redis's `SCRIPT LOAD` command to compute the script digest server-side.

Changes:
- Bump `github.com/redis/go-redis/v9` to a version containing `NewScriptServerSHA()`
- Add `isFIPSMode()` helper that detects FIPS via `GODEBUG`
- Add `newAllowTokenBucketScript()` wrapper that picks the correct constructor based on the runtime environment
- Non-FIPS users see zero behavior change; the existing `NewScript()` path is preserved

### Motivation

Running Traefik in strict FIPS 140-3 environments is a hard requirement for many regulated deployments. Today, any FIPS-mode deployment that uses the rate limiter middleware crashes at startup with:
panic: crypto/sha1: use of SHA-1 is not allowed in FIPS 140-only mode goroutine 1 [running]: crypto/sha1.(*digest).checkSum(...) github.com/redis/go-redis/v9.NewScript(...) github.com/traefik/traefik/v3/pkg/middlewares/ratelimiter.init() pkg/middlewares/ratelimiter/lua.go:66

The same binary now works in both FIPS and non-FIPS environments, with the runtime environment deciding which path is taken.

### More

- [x] Added/updated tests

### Additional Notes

Tested end-to-end against a real Redis instance:

- Existing tests pass: `go test ./pkg/middlewares/ratelimiter/...`
- Non-FIPS startup: clean, no behavior change
- FIPS startup (`GODEBUG=fips140=only`): no panic
- Rate limiting verified in both modes (HTTP 200 under limit, 429 when exceeded)
- `SCRIPT FLUSH` recovery works in FIPS mode (script auto-reloads on `NOSCRIPT`)
- Verified via `redis-cli MONITOR` that FIPS mode uses `SCRIPT LOAD` + `EVALSHA`, and non-FIPS mode uses `EVALSHA` directly

## Related

- Upstream go-redis fix: redis/go-redis#3700 (merged)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,7 @@ require (
 	github.com/prometheus/client_golang v1.23.0
 	github.com/prometheus/client_model v0.6.2
 	github.com/quic-go/quic-go v0.59.0
-	github.com/redis/go-redis/v9 v9.8.0
+	github.com/redis/go-redis/v9 v9.18.1-0.20260409133122-462843a27550
 	github.com/rs/zerolog v1.33.0
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spiffe/go-spiffe/v2 v2.6.0
@@ -187,7 +187,6 @@ require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/deepmap/oapi-codegen v1.9.1 // indirect
 	github.com/desertbit/timer v0.0.0-20180107155436-c41aec40b27f // indirect
-	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
 	github.com/dimchansky/utfbom v1.1.1 // indirect
 	github.com/distribution/reference v0.6.0 // indirect
 	github.com/dnsimple/dnsimple-go/v4 v4.0.0 // indirect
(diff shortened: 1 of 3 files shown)


emilevauge:
Hi, and thank you for taking the time to open this PR and work on FIPS support!

After reviewing this contribution, we've decided to close it for three reasons:

1. **FIPS 140-3 support in Go is not yet finalized.** The upstream Go implementation is still in progress (see golang/go#69536), which means any FIPS-related integration in Traefik OSS would be built on an unstable foundation.

2. While rebuilding Traefik images from source is of course possible, the resulting images won't be identical to those produced by Traefik Labs. Our official images rely on proprietary build configuration with build-time credentials that cannot be replicated externally, which means some features will be missing from custom-built images.

3. **FIPS support is part of our commercial offering.** FIPS compliance is a feature we maintain and ship through Traefik Hub, our paid product. We intentionally keep this outside of the open-source project.

For these reasons, we don't think merging this PR is the right path forward for the project at this time.
Thanks again for your contribution and your interest in Traefik!

PR #12973, opened 2026-04-14
Use QuoteMeta for cookie name when building canary rules

kevinpollet:
### What does this PR do?

This pull request escapes the Cookie name when building the canary rules.

### More

- [X] Added/updated tests


rtribotte (approved):
LGTM

PR #12976, opened 2026-04-14
Honor allowCrossNamespace with chain middleware CRD

rtribotte:
### What does this PR do?

This PR changes the way chain middleware CRD is discovered to honor the allowCrossNamespace configuration with the chain middleware CRD.

If `allowCrossNamespace` is set to `false` (the default) and a `Chain` middleware references a middleware in a different namespace from its own,
the whole `Chain` is now rejected and an error is logged.

### Motivation

To honor allowCrossNamespace configuration.

### More

- [x] Added/updated tests


kevinpollet (approved):

gndz07 (approved):
LGTM

PR #12977, opened 2026-04-14
Avoid 302 redirect when rewrite-target value is not an absolute URL for ingress-nginx provider

gndz07:
### What does this PR do?

When the rewrite-target is configured as an absolute URL, Traefik issues a 302 redirect. However, when using a capture group replacement (e.g. $1), a user-controlled path segment containing an absolute URL could also trigger an open redirect.

This fix ensures a 302 redirect is only issued when the replacement template itself is an absolute URL, not when one is injected via a capture group.

### Motivation

To avoid open redirect via user-injected capture group content.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/ingressnginx/rewritetarget/rewrite_target.go
+++ b/pkg/middlewares/ingressnginx/rewritetarget/rewrite_target.go
@@ -28,6 +28,7 @@ type rewriteTarget struct {
 	replacement      string
 	xForwardedPrefix string
 	name             string
+	redirectAbsolute bool
 }
 
 // New creates a new rewrite target middleware.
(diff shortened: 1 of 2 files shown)


pkg/middlewares/ingressnginx/rewritetarget/rewrite_target.go
```diff
@@ -28,6 +28,9 @@ type rewriteTarget struct {
 	replacement      string
 	xForwardedPrefix string
 	name             string
+	// absoluteURLRedirect  is true when the replacement template is an absolute URL,
```
mmatur:
```suggestion
	// absoluteURLRedirect is true when the replacement template is an absolute URL,
```

mmatur (approved):
LGTM

kevinpollet (approved):

PR #12987, opened 2026-04-15
Fix casing of 'defaultMode' to 'defaultmode'

niklasfyi:
### What does this PR do?

This PR fixes a typo in access logs. Helm did not set the correct args for logging headers. Now it is aligned with the values.yml in traefiks helm chart repo.

### Motivation

We were unable to change the values for logging headers in the access log.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/observe/logs-and-access-logs.md
+++ b/docs/content/observe/logs-and-access-logs.md
@@ -55,7 +55,7 @@ accessLog:
     names:
       ClientUsername: drop
     headers:
-      defaultMode: keep
+      defaultmode: keep
       names:
         User-Agent: redact
         Content-Type: keep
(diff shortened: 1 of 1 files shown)


mmatur:
Hi @niklasfyi, thanks for the contribution

However, defaultMode (camelCase) is actually the correct casing for Traefik's configuration — it matches the struct tags in the code (pkg/observability/types/logs.go) and the YAML/TOML serialization format. Changing the documentation to defaultmode would make it incorrect.

The issue you're experiencing is actually in the Helm chart's values.yaml which uses incorrect casing. I'd suggest opening an issue (or a PR) on traefik/traefik-helm-chart instead.

Closing this PR as the fix belongs on the Helm chart side. Thanks for looking into this!

PR #13002, opened 2026-04-17
Add load balancer strategy annotation for Ingress

### LINKED ISSUE #12983: P2C for Ingress resource
### Welcome!

Hello,

we have recently switched to Traefik from Nginx and we are experiencing some issues when it comes to our setup.

We quickly found out that we need to use P2C instead of rr so we had to switch from Ingress to IngressRoute.
That has caused issues with our cert-manager, which relies on Ingresses to renew certificates.

There are some potential ways to combat this, like using Traefik's certificate feature or switching to Gateway API. We don't like these 2 options for various reasons. We can also simply add a Certificate resource next to the IngressRoute resource.

My question is, is it possible to have P2C, even as a default load balancing algorithm, for the Ingress resource? I guess it is not that simple, but thought it was worth asking, since it will solve all our problems.

PantelisK:
## What does this PR do?

Adds support for configuring the load balancer strategy via a service annotation on Kubernetes Ingress resources.

This introduces a new annotation `traefik.ingress.kubernetes.io/service.strategy` that can be set on Kubernetes Service resources to configure the load balancing algorithm.

**Supported values:**
- `wrr` (weighted round-robin) - default
- `p2c` (power of two choices)
- `hrw` (highest random weight)
- `leasttime`

**Example usage:**
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-service
  annotations:
    traefik.ingress.kubernetes.io/service.strategy: "p2c"
spec:
  ports:
    - port: 80
```

## Motivation

This brings feature parity between standard Kubernetes Ingress resources and IngressRoute/TraefikService CRDs for load balancer strategy configuration.

As discussed in #12983, users migrating from Nginx to Traefik need P2C load balancing but want to continue using standard Ingress resources (e.g., for cert-manager compatibility). Previously, the only way to configure load balancer strategy was through IngressRoute CRDs.

## More

- [x] Added/updated tests

## Additional Notes

As far as I can understand all the code around the load balancer already exists. I've just added the annotation in the Ingress resource.

Closes #12983


kevinpollet:
Hey @PantelisK,

Thanks for this contribution!

PR #13004, opened 2026-04-18
feat: add reusePrivateKey option

### LINKED ISSUE #10103: Support disabling private key reuse in certificate renewal
### Welcome!

We are getting `urn:ietf:params:acme:error:badCSR :: Encountered errors validating CSR; errors=[CSR_PUBLIC_KEY_REUSE_NOT_ALLOWED]` error while traefik tries to renew certs. Basically acme provider refuses to do renewal due to key reuse.  Looking at the code, it seems like there is no option to avoid that:
https://github.com/traefik/traefik/blob/3fd5c747a2e16c1a337da733a84bcf159c56599b/pkg/provider/acme/provider.go#L820-L824

I think there should be an option to control this.

ldez:
### What does this PR do?

Add an option to disable the fact to reuse the certificate private key.

### Motivation

Fixes #10103

### More

- ~~[ ] Added/updated tests~~

### Additional Notes

I recommend changing this default to false in the future.


pkg/provider/acme/provider.go
```diff
@@ -936,10 +940,13 @@ func (p *Provider) renewCertificates(ctx context.Context, renewPeriod time.Durat
 		res := certificate.Resource{
 			ID:          cert.Domain.Main,
 			Domains:     cert.Domain.ToStrArray(),
-			PrivateKey:  cert.Key,
 			Certificate: cert.Certificate.Certificate,
 		}
 
+		if ptr.Deref(p.ReusePrivateKey, true) {
```
jnoordsij: From what I could gather from lego logic (but correct me if I'm wrong, you're the expert!) we should pass `KeyType` here if not passing an existing private key, to ensure we use the Traefik-configuration value rather than lego default, right? See also https://github.com/traefik/traefik/pull/13910/changes#diff-4f93ba21a09614847c01e2f916eba5dcc78e000cb9c076900dc033aa5d2bfc58R942-R945.

jnoordsij (commented):
Thanks for this!

As per my notes on #13910:
- totally agree this should default to `false` at some point in the future
- given how Traefik & lego currently work, it might be worth adding some note to `keyType` that changing this value will only affect new certificates, unless this option is toggled

pkg/provider/acme/provider.go
ldez: Yes but there is another problem related to the `Renew` method, I will fix it inside lego.

ldez:
> given how Traefik & lego currently work, it might be worth adding some note to `keyType` that changing this value will only affect new certificates, unless this option is toggled

If changed the behavior: if there is a private key and the key type of this private key is different than the key type defined on the provider, the private key is ignored.

I think there are several issues inside Traefik related to that (ex: https://github.com/traefik/traefik/issues/12146#issuecomment-3562146162)

pkg/provider/acme/provider.go
```diff
@@ -936,8 +941,19 @@ func (p *Provider) renewCertificates(ctx context.Context, renewPeriod time.Durat
...
 			Certificate: cert.Certificate.Certificate,
+			KeyType:     GetKeyType(ctx, p.KeyType),
+		}
+
+		if ptr.Deref(p.ReusePrivateKey, true) && len(cert.Key) > 0 {
+			keyType, err := getPrivateKeyType(cert)
+			if err != nil {
+				logger.Info().Err(err).Msgf("Error renewing ACME certificate: %+v", cert.Domain)
+
+				res.PrivateKey = cert.Key
```
jnoordsij, line 953:
```suggestion
				continue
```

Shouldn't we abort here?

ldez: I don't abort to keep the previous behavior (no failure) in case of the usage of a specific private key type.
The goal is to avoid any breaking changes.

jnoordsij: Ah fair. It will still err in lego (https://github.com/go-acme/lego/blob/01673cc5c504e0ebb41f652af7f35fb0eac3df4f/certificate/certificates.go#L244) at some point right? As renewing will not succeed in any way without getting the type?

ldez: Yes that's true, I forgot this part, so I will change the case.

ldez: :thinking: I will not abort but fallback: if the key type cannot be extracted form the private key, the private key will be ignored and the key type defined by the configuration will be used.

jnoordsij: Sounds great! Then it is a change of behavior compared to previous, however I'd argue it to be an improvement/bug fix to choose "renewal with new private key" over "failure with unrecognized existing key", given the latter should be very unlikely to occur in practice.

jnoordsij (approved):
LGTM

PR #12035, opened 2025-09-02
refactor: configurable HTTP client and base URL

jspdown:
### What does this PR do?

This PR updates the `plugins.Client` to allow the configuration of the underlying `http.Client` and base URL. 

### Motivation

Opening these 2 new options makes it possible to cover the implementation with tests and allow better extensibility of the plugin system.

### More

- [x] Added/updated tests
- [ ] ~~Added/updated documentation~~

PR #13006, opened 2026-04-19
Reuse cached plugin archive when plugins.traefik.io is unreachable

### LINKED ISSUE #13005: Plugin install/validation on every startup couples runtime to plugins.traefik.io; transient failure disables all plugins (regression since v3.5.3)
#### What did you do?

Ran Traefik **v3.5.3** in a minimal docker-compose stack with one experimental plugin (`github.com/sablierapp/sablier-traefik-plugin` v1.1.0) and one file-provider middleware referencing it. Standard configuration, no custom network tweaks.

Full reproduction: https://github.com/5queezer/sablier-traefik-repro

#### What did you expect to see?

A one-time download of the plugin archive on first start, then subsequent starts using the cached archive from the plugins storage volume. Plugin validation issues should not make all plugins and all plugin-backed middlewares unusable.

#### What did you see instead?

On every start, the plugin manager makes a blocking HTTP call to `plugins.traefik.io/public/download/...` (and subsequently `/validate/...`). If that call fails for **any** reason (transient DNS, registry rate-limit, CDN hiccup, captive portal), the failure path in `pkg/plugins/manager.go` calls `manager.ResetAll()`, disables **all** configured plugins, and cascades into:

```
ERR  invalid middleware "<name>@file" configuration: invalid middleware type or middleware does not exist
```

for every middleware that references a plugin. The user-facing symptom is sporadic router breakage after restarts (nightly cron, backup jobs, planned reboots), with manual retry usually fixing it.

#### Deterministic reproduction

`./reproduce-deterministic.sh` in the linked repo maps `plugins.traefik.io` to `127.0.0.1` inside the Traefik container via Docker `extra_hosts`, which forces a `connection refused` on the plugin install call. 100% trigger rate, no restart loop required:

```
INF  Loading plugins...  plugins=["sablier"]
ERR  Plugins are disabled because an error has occurred.
     error="unable to set up plugins environment: unable to install plugin sablier:
     unable to download plugin github.com/sablierapp/sablier-traefik-plugin:
     failed to call service: Get \"https://plugins.traefik.io/public/download/github.com/sablierapp/sablier-traefik-plugin/v1.1.0\":
     dial tcp 127.0.0.1:443: connect: connection refused"
ERR  error="invalid middleware \"whoami-sablier@file\" configuration:
     invalid middleware type or middleware does not exist"
```

#### Output of `traefik version`

```
Version:      3.5.3
Built:        2025-09-26T09:20:06Z
```

Also observed in released 3.6.x builds per user reports in #12137 and sablierapp/sablier#773.

#### What is your environment & configuration?

Minimal; full config in the linked repo. Key pieces:

```yaml
# traefik.yml (static)
experimental:
  plugins:
    sablier:
      moduleName: github.com/sablierapp/sablier-traefik-plugin
      version: v1.1.0

# dynamic.yml (file provider)
http:
  middlewares:
    whoami-sablier:
      plugin:
        sablier:
          sablierUrl: http://sablier:10000
          group: whoami
          sessionDuration: 30s
```

Docker Engine 28.5.2 on macOS (OrbStack). Error is network-path dependent, not host-OS dependent.

#### Root cause

[#12035](https://github.com/traefik/traefik/pull/12035) — "Refactor plugins system" — moved plugin installation into a new `Manager`/`Downloader` split and added a `Check()` call per plugin on top of the existing `Download()`. Both hit `plugins.traefik.io/public/...`. The aggregated error path treats any failure from either call as fatal and triggers `manager.ResetAll()`. This is why a single transient network blip disables every plugin-backed middleware in the configuration, even when the plugin archive is already cached locally and would otherwise load fine.

Pre-v3.5.3, plugin startup had lower coupling to `plugins.traefik.io` runtime reachability — this change regressed operational robustness for any operator whose Traefik restarts occasionally happen during unlucky moments (nightly cron, backup windows, registry rate-limit events).

#### Related

- #12137 — earlier report, closed as `frozen-due-to-age`, locked.
- #12035 — PR that introduced the coupling.
- sablierapp/sablier#773 — the most-cited downstream symptom report.

#### Proposed fix (happy to PR)

The minimal change that keeps the integrity check meaningful for fresh installs while restoring operational robustness: **when the plugin archive is already present and extractable in the local plugins storage directory, treat `downloader.Check()` failure as a logged warning, not a fatal error.** Fresh installs still fail loudly if the registry is unreachable (because there's nothing cached to fall back to), but ongoing restarts don't get held hostage by registry uptime.

Alternatives, in case this shape isn't preferred:

1. Config flag `experimental.plugins.offline: true` — operator opt-in, safe default unchanged, no implicit behavior change.
2. Per-plugin cached validation result with a configurable TTL, so restart cadence doesn't depend on `plugins.traefik.io` uptime.
3. Move `Check()` to an async post-startup goroutine — log subsequent validation failures at warning level and do not affect middleware registration.

Happy to open a PR for option 1 (non-fatal-when-cached) or whichever the team prefers. Want to check first that there's appetite to fix this before investing in implementation and tests.

5queezer:
### What does this PR do?

When `Downloader.Download` fails and a previously-downloaded archive for the same plugin+version is already on disk, fall back to the cached archive instead of returning an error (which cascades into `manager.ResetAll()` and disables every plugin-backed middleware). In that fallback path, a subsequent `Downloader.Check` failure against `plugins.traefik.io` is also tolerated, because the cached archive was validated on the prior successful install.

Fresh installs (no cached archive on disk) remain fatal as before. Integrity-check failures that follow a *successful* `Download` also remain fatal, so a tampered or corrupted freshly-downloaded archive is still rejected. Hash pinning (`plugin.Hash`) is always enforced, including in the cached-fallback path.

### Motivation

Since v3.5.3 (#12035 restructured the plugin manager), every Traefik restart makes blocking HTTP calls to `plugins.traefik.io` for each configured plugin. Any failure on those calls — transient DNS, CDN hiccup, registry rate-limit, captive portal — currently causes `manager.ResetAll()` to wipe the whole plugin environment and cascades into:

```
ERR  invalid middleware "<name>@file" configuration: invalid middleware type or middleware does not exist
```

for every middleware that references a plugin. The user-facing symptom is sporadic router breakage after otherwise-routine restarts (nightly cron, backups, planned reboots), where a manual retry usually fixes it.

Filed as #13005 with a deterministic reproduction (https://github.com/5queezer/sablier-traefik-repro — maps `plugins.traefik.io` to `127.0.0.1` via Docker `extra_hosts` and gets a 100% trigger rate). Also surfaced downstream in traefik/traefik#12137 and sablierapp/sablier#773.

Pre-v3.5.3 releases did not have this level of runtime coupling to registry reachability on every start, so this restores that operational property without reverting the integrity-check work.

### More

**Scope of the change**

In `pkg/plugins/manager.go`, `InstallPlugin` now attempts `Download` as before. On download error, it stats the target archive path. If the archive exists, it logs a warning, flips an internal `fallback` flag, and recomputes the archive hash locally. If the archive does not exist, the download error remains fatal.

After hash resolution, the PR still runs `plugin.Hash` comparison and `Downloader.Check` exactly as before. `Check` failure is tolerated *only* when `fallback` is set — that is, only when we have explicitly chosen the cached archive because `Download` was unreachable. `Check` failure after a successful `Download` remains fatal, preserving the integrity guarantee for any content newly served by the registry.

**Tests**

Three new subtests added to `TestPluginManager_InstallPlugin`:

- `download fails but cached archive is reused` — asserts install succeeds when the network is unreachable but the archive is cached
- `integrity check failure after a successful download stays fatal` — asserts that a `Check` failure on freshly-downloaded content is still fatal, preventing a regression in the integrity gate
- `pinned hash still enforced when falling back to cached archive` — asserts install fails if `plugin.Hash` doesn't match the cached archive, even when download fails

All four existing subtests pass unchanged. A small `writeValidPluginArchive` helper was extracted so the new cases don't duplicate the zip-writing boilerplate.

**Alternatives considered (and why not)**

Listed these in #13005 for design discussion — happy to switch if maintainers prefer a different shape:

1. Config flag `experimental.plugins.offline: true` — operator opt-in, safe default unchanged. Less invasive but requires config changes from affected operators.
2. Per-plugin cached validation result with a configurable TTL — more complex, more state to manage.
3. Move `Check()` to an async post-startup goroutine — moves the complexity to concurrency handling and changes semantics (plugins are "up" before integrity is confirmed).

Option 1 (this PR) is the minimal change that restores robustness without changing the default integrity-check behaviour for fresh installs or newly-downloaded content.

**Local verification**

- `go build ./pkg/plugins/...` — clean
- `go vet ./pkg/plugins/...` — clean
- `gofmt -l pkg/plugins/` — clean
- `go test ./pkg/plugins/...` — all 7 subtests pass, including 3 new ones

Happy to run `make validate` / `make test` on top if anything is unexpectedly red in CI — let me know.

### Additional Notes

- Fixes #13005
- Related: traefik/traefik#12137 (closed as `frozen-due-to-age`), sablierapp/sablier#773
- Targets `v3.6` per CONTRIBUTING guidance for bug fixes
- "Allow edits from maintainers" enabled

### DIFF
--- a/pkg/plugins/manager.go
+++ b/pkg/plugins/manager.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/rs/zerolog/log"
 	"golang.org/x/mod/module"
 	"golang.org/x/mod/zip"
 	"gopkg.in/yaml.v3"
@@ -77,10 +78,38 @@ func NewManager(downloader PluginDownloader, opts ManagerOptions) (*Manager, err
 }
 
 // InstallPlugin download and unzip the given plugin.
+//
+// When a previously downloaded archive is present on disk and the registry
+// download fails, the install falls back to the cached archive instead of
+// wiping the plugin environment. This decouples ongoing Traefik restarts from
+// plugins.traefik.io reachability. In that fallback path, a subsequent
+// integrity check failure is also tolerated, because the cached archive was
+// validated on the prior successful install.
+//
+// Fresh installs (no cached archive) and integrity check failures that follow
+// a successful download remain fatal, so security guarantees for new or
+// updated plugins are unchanged. Hash pinning (plugin.Hash) is always enforced,
+// including in the cached-fallback path.
 func (m *Manager) InstallPlugin(ctx context.Context, plugin Descriptor) error {
+	archivePath := m.buildArchivePath(plugin.ModuleName, plugin.Version)
+
+	var fallback bool
 	hash, err := m.downloader.Download(ctx, plugin.ModuleName, plugin.Version)
 	if err != nil {
-		return fmt.Errorf("unable to download plugin %s: %w", plugin.ModuleName, err)
+		if _, statErr := os.Stat(archivePath); statErr != nil {
+			return fmt.Errorf("unable to download plugin %s: %w", plugin.ModuleName, err)
+		}
+
+		log.Ctx(ctx).Warn().Err(err).
+			Str("module", plugin.ModuleName).
+			Str("version", plugin.Version).
+			Msg("Plugin download failed; falling back to previously cached archive")
+
+		fallback = true
+		hash, err = computeHash(archivePath)
+		if err != nil {
+			return fmt.Errorf("unable to hash cached plugin archive %s: %w", archivePath, err)
+		}
 	}
 
 	if plugin.Hash != "" {
@@ -88,9 +117,15 @@ func (m *Manager) InstallPlugin(ctx context.Context, plugin Descriptor) error {
 			return fmt.Errorf("invalid hash for plugin %s, expected %s, got %s", plugin.ModuleName, plugin.Hash, hash)
 		}
 	} else {
-		err = m.downloader.Check(ctx, plugin.ModuleName, plugin.Version, hash)
-		if err != nil {
-			return fmt.Errorf("unable to check archive integrity of the plugin %s: %w", plugin.ModuleName, err)
+		if checkErr := m.downloader.Check(ctx, plugin.ModuleName, plugin.Version, hash); checkErr != nil {
+			if !fallback {
+				return fmt.Errorf("unable to check archive integrity of the plugin %s: %w", plugin.ModuleName, checkErr)
+			}
+
+			log.Ctx(ctx).Warn().Err(checkErr).
+				Str("module", plugin.ModuleName).
+				Str("version", plugin.Version).
+				Msg("Plugin integrity check against registry failed; using previously validated cached archive")
 		}
 	}
 
--- a/pkg/plugins/manager_test.go
+++ b/pkg/plugins/manager_test.go
@@ -33,6 +33,27 @@ func (m *mockDownloader) Check(ctx context.Context, pName, pVersion, hash string
 	return nil
 }
 
+// writeValidPluginArchive writes a minimal but structurally valid zip archive at
+// archivePath, suitable for InstallPlugin's unzip step in tests.
+func writeValidPluginArchive(t *testing.T, archivePath string) {
+	t.Helper()
+
+	err := os.MkdirAll(filepath.Dir(archivePath), 0o755)
+	require.NoError(t, err)
+
+	file, err := os.Create(archivePath)
+	require.NoError(t, err)
+	defer file.Close()
+
+	writer := zipa.NewWriter(file)
+	defer writer.Close()
+
+	fileWriter, err := writer.Create("test-module-v1.0.0/main.go")
+	require.NoError(t, err)
+	_, err = fileWriter.Write([]byte("package main\n\nfunc main() {}\n"))
+	require.NoError(t, err)
+}
+
 func TestPluginManager_ReadManifest(t *testing.T) {
 	tempDir := t.TempDir()
 	opts := ManagerOptions{Output: tempDir}
@@ -297,6 +318,51 @@ func TestPluginManager_InstallPlugin(t *testing.T) {
 			},
 			expectError: true,
 			errorMsg:    "unable to unzip plugin",
+		},
+		{
+			name: "download fails but cached archive is reused",
+			plugin: Descriptor{
+				ModuleName: "github.com/test/plugin",
+				Version:    "v1.0.0",
+			},
+			downloadFunc: func(ctx context.Context, pName, pVersion string) (string, error) {
+				return "", assert.AnError
+			},
+			checkFunc: func(ctx context.Context, pName, pVersion, hash string) error {
+				return assert.AnError
+			},
+			setupArchive: writeValidPluginArchive,
+			expectError:  false,
+		},
+		{
+			name: "integrity check failure after a successful download stays fatal",
+			plugin: Descriptor{
+				ModuleName: "github.com/test/plugin",
+				Version:    "v1.0.0",
+			},
+			downloadFunc: func(ctx context.Context, pName, pVersion string) (string, error) {
+				return "test-hash", nil
+			},
+			checkFunc: func(ctx context.Context, pName, pVersion, hash string) error {
+				return assert.AnError
+			},
+			setupArchive: writeValidPluginArchive,
+			expectError:  true,
+			errorMsg:     "unable to check archive integrity",
+		},
+		{
+			name: "pinned hash still enforced when falling back to cached archive",
+			plugin: Descriptor{
+				ModuleName: "github.com/test/plugin",
+				Version:    "v1.0.0",
+				Hash:       "a-hash-that-will-not-match-the-archive-contents",
+			},
+			downloadFunc: func(ctx context.Context, pName, pVersion string) (string, error) {
+				return "", assert.AnError
+			},
+			setupArchive: writeValidPluginArchive,
+			expectError:  true,
+			errorMsg:     "invalid hash for plugin",
 		},
 	}
 


5queezer:
Self-review follow-up before anyone reads this: I tightened the integrity-check semantics and force-pushed.

The earlier version tolerated `Downloader.Check` failure whenever the archive path existed on disk at the start of `InstallPlugin`. But `Downloader.Download` overwrites the archive on success, so an 'archive existed at start' flag does not prove the currently-on-disk archive is the previously-validated one. That would have let a post-download integrity mismatch pass as a warning, which is not the intended security property.

Replaced with an internal `fallback` flag set only when we explicitly fall back to the cached archive because `Download` failed. `Check` failure is tolerated only in that fallback path. A `Check` failure after a successful `Download` is fatal again, as before the refactor. Test that previously asserted this as a success is renamed and flipped to assert failure (`integrity check failure after a successful download stays fatal`).

Happy to split the tightening into a separate commit if that is easier to review — currently single commit because no one had looked yet.

kevinpollet:
Hey @5queezer,

berglh:
@kevinpollet Any update on this review?

It's impacting a deployment I have whenever I lose power at home, the server starts before the Internet is up and my middle-ware gets nuked. I've now added an initContainer on my Traefik deployment to check for access to the plugin URL resource using curl, which should hopefully deal with the power outage recovery situation. Another situation would be that Traefik won't be available while my Internet is offline and my Pod/server is restarted then I'll have no middle-ware, meaning the routes depending on it won't be available.

nmengin:
Hey @5queezer,

TrapStoner:
> @kevinpollet Any update on this review?
> 
> It's impacting a deployment I have whenever I lose power at home, the server starts before the Internet is up and my middle-ware gets nuked. I've now added an initContainer on my Traefik deployment to check for access to the plugin URL resource using curl, which should hopefully deal with the power outage recovery situation. Another situation would be that Traefik won't be available while my Internet is offline and my Pod/server is restarted then I'll have no middle-ware, meaning the routes depending on it won't be available.

Same! I was bummed to find out the hard way that download failure results in fatal and disabling of plugins instead of cache fallback when power went down and server booted up before router and half my routes were returning 404. Can't wait for this to get merged and until then i will just clone plugins to local plugins folder i guess because i don't exactly like the idea of making a plugin recovery service which detects `Plugins are disabled` > polls a domain > restarts traefik when internet is back up.

dierochade:
thanks for the PR. 

In my view, current situation clearly qualifies as bug, not enhancement. 

its a bad/broken design if a single unsuccessful call for an online resource at startup breaks all routers with plugins and there is not even a retry logic or whatever. 

Would really appreciate if this could be reviewed and could advance soon.

berglh:
> its a bad/broken design if a single unsuccessful call for an online resource at startup breaks all routers with plugins and there is not even a retry logic or whatever.

I appreciate the sentiment. I myself would not go as far to call it a bug, more so an intentional design choice as a definitive way to prevent routes being published without their plugins available, even if the implementation may have been short sighted.

From memory, it is possible to install the plugins in development mode in a local folder where they do persist. They will need to be manually updated by file placement and is a significant challenge in actual Kubernetes clusters with replicas. If running Docker, it can be completed with a simple host path mount for the local plugins. If you're running it via K8S in Helm you can do something like this in the values YAML which will wait for the plugin servers to be reachable before starting the actual Traefik container:

> [!note]
> You can just replace the curl URIs with the plugin paths you want to check for.

<details>
  <summary>Click to expand Helm values example</summary>

```yaml
deployment:
  initContainers:
    - name: wait-for-plugin-registry
      image: curlimages/curl:8.12.1
      command:
        - sh
        - -ec
        - |
          until \
            curl -fsS --head https://plugins.traefik.io/public/download/github.com/andrewkroh/google-oidc-auth-middleware/v0.3.0 >/dev/null \
            && curl -fsS --head https://plugins.traefik.io/public/download/github.com/PascalMinder/geoblock/${geoblock_middleware_version} >/dev/null; do
            echo "Waiting for external DNS/internet so Traefik plugin downloads can succeed..."
            sleep 5
          done`
```
</details>

owpac:
Hello!

Hit this again last night on a Docker host, same reboot-after-power-loss scenario @berglh and @TrapStoner described. Adding it because of one aspect I don't think has been raised: **the outage is silent**.

My resolver (AdGuard) is itself a container. On boot it reports healthy, so `depends_on: condition: service_healthy` is satisfied, but its upstream isn't usable yet:

```
INF Loading plugins... plugins=["crowdsec-bouncer","geoblock"]
ERR Plugins are disabled because an error has occurred.
  error="... unable to install plugin crowdsec-bouncer: ... 
  dial tcp: lookup plugins.traefik.io on 127.0.0.11:53: server misbehaving"
```

Only the CrowdSec download failed, but `ResetAll()` takes out geoblock too — note both on the error line. Every chain I have references `crowdsec@file`, so every router is dropped and my catch-all error page answers for all of them: total outage, not degradation.

The part worth weighing in the design review: Traefik doesn't crash, doesn't exit non-zero, and `traefik healthcheck --ping` keeps returning healthy. The container sits at `Up (healthy)` throughout, so `restart: unless-stopped` has nothing to act on and container-level monitoring sees a green stack. A few seconds of DNS unavailability at boot become an indefinite outage that only ends when a human notices. `docker restart traefik` fixed it instantly.

The cached-archive fallback looks like the right shape to me. It doesn't weaken the "no routes without their plugins" property: the cached archive is the same validated artifact that was running before the restart, so the route comes back with the code it had five minutes earlier. Fresh installs stay fatal, so the integrity gate still covers anything new. All that changes is refusing to start with bytes already on disk because a network call couldn't confirm them.

On status: I've followed this since June expecting 3.6. The milestone moved to 3.7, and 3.7 has now shipped (I'm on 3.7.13) without it. Last maintainer update is 2026-05-18.

@kevinpollet @nmengin: is there anything blocking the design review that contributors could help with?

PR #13008, opened 2026-04-20
Inject active span context into request headers for plugin middlewares

### LINKED ISSUE #12891: traceparent not propagated to plugin middleware context — breaks distributed tracing for plugins making outbound requests
### Welcome!

This is a follow-up to the frozen proposal #11816, highlighting a concrete use case that makes this gap especially impactful.

A common middleware pattern is to have a Traefik plugin validate incoming requests by calling out to a sidecar or microservice — for example, verifying an ECDSA signature before proxying upstream:
Client → Traefik → [Plugin: verify ECDSA sig → calls /verify service] → Upstream

Currently, traceparent is not present in the request headers when ServeHTTP is called on the plugin, even though Traefik has already created a trace span. This results in orphaned traces from all microservices involved in the validation flow — they cannot be correlated back to the main Traefik trace.

Expected Behaviour:
The active traceparent / tracestate should be available in the request headers at the plugin layer, so plugins can propagate trace context to any outbound calls they make and have them correctly nested under the main trace.

Version:
Traefik v3.6.8

JoeGruffins:
Before this change, traefik created a trace span for each incoming request but did not propagate the resulting traceparent/tracestate through the request headers. Downstream middlewares - particularly plugins that read raw headers to propagate trace context on outbound calls - never saw the active span, causing orphaned traces in distributed tracing setups.

Inject the active span context into the request headers in the tracing entry point and in the per-middleware tracing wrapper so plugins making outbound calls can propagate trace context correctly.

closes #12891

### What does this PR do?

Injects the active span context into request headers inside the tracing
entry point and the per-middleware tracing wrapper, so plugin middlewares
that read raw headers (e.g. to propagate trace context on outbound calls
they make themselves) see the correct `traceparent` / `tracestate`.

Previously, traefik started a span per request but only injected its
headers on the outbound proxied request in the RoundTripper
(`pkg/proxy/httputil/observability.go`). Plugin middlewares making
side-band HTTP calls never saw the active span, producing orphaned
traces in distributed setups.

### Motivation

Looking to get commits into golang repos with high traffic.

### More

- [x] Added/updated tests

### Additional Notes

**Behavioral change to be aware of:** any inbound `traceparent` /
`tracestate` headers on the request are now overwritten before
downstream middlewares run. This is the correct W3C semantics
(downstream should see the current span, not the client's parent
span), and matches what the outbound RoundTripper already does - but
it is a visible change for any middleware that introspects those
headers directly.

The middleware-side injection is gated on `DetailedTracingEnabled`,
so when detailed tracing is off, plugins see the entry-point span's
`traceparent` (not a per-middleware one).

### DIFF
--- a/pkg/middlewares/observability/entrypoint.go
+++ b/pkg/middlewares/observability/entrypoint.go
@@ -67,6 +67,11 @@ func (e *entryPointTracing) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 
 	req = req.WithContext(loggerCtx)
 
+	// Inject the active span context into the request headers so downstream
+	// middlewares (notably plugins) can propagate trace context on outbound
+	// requests they make.
+	tracing.InjectContextIntoCarrier(req)
+
 	span.SetAttributes(attribute.String("entry_point", e.entryPoint))
 
 	e.tracer.CaptureServerRequest(span, req)
--- a/pkg/middlewares/observability/entrypoint_test.go
+++ b/pkg/middlewares/observability/entrypoint_test.go
@@ -8,7 +8,9 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/traefik/traefik/v3/pkg/observability/tracing"
+	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/propagation"
 )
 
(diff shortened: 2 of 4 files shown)


kevinpollet:
Hey @JoeGruffins,

Thanks for this contribution!

pkese:
Bump.

This would be a really valuable addition (I'm having the same problem).

PR #13009, opened 2026-04-20
Make FLAGS Make variable usable

twz123:
### What does this PR do?

Strip the [*] suffix, so it's usable in the expected way when invoking make, e.g. like so: `make FLAGS=-trimpath`.

### Motivation

Previously, it was using the bash array syntax, which has no special meaning at all in a Makefile. Using something like `make FLAGS[*]=-trimpath` worked, but that doesn't seem intentional.

### More

- [ ] Added/updated tests

### Additional Notes

Fixes: 39b0aa665 ("Improve makefile")


mmatur (approved):
LGTM

PR #13018, opened 2026-04-21
Support middleware with TraefikService CRD

sdelicata:
### What does this PR do?

This PR fixes a bug where middlewares declared on a weighted / HRW / mirroring / failover child of a TraefikService are silently dropped at runtime when that child targets another TraefikService (kind: TraefikService). 

The CRD schema has always accepted the `middlewares:` field on every reference site, so manifests apply cleanly and look valid, but the chain is never attached.

This fix:
- adds Middlewares to the dynamic-config reference types (WRRService, HRWService, MirrorService, Mirroring, Failover) and FallbackMiddlewares to Failover
- wires the CRD builder to populate them whenever the child is a TraefikService
- wraps the child handler with the per-child chain at runtime while preserving the `healthcheck.StatusUpdater` interface so balancer healthcheck registration keeps working

### Motivation

The only middleware-on-service mechanism in the dynamic config is `dynamic.Service.Middlewares`, applied on the callee regardless of the parent. 

For `kind: Service` children the CRD provider accidentally satisfies per-child semantics, because `buildServersLB` synthesizes an exclusive inner `dynamic.Service` for each caller and the per-child middlewares collapse onto that synthesized node. 

But for `kind: TraefikService` children the reference resolves to an existing, potentially shared `dynamic.Service`, so `LoadBalancerSpec.Middlewares` no-ops.

Example:

```yaml
kind: TraefikService
metadata:
  name: svc
spec:
  weighted:
    services:
      - kind: TraefikService
        name: child-svc
        weight: 1
        middlewares:
          - name: addprefix   # silently dropped before this fix
```

Dashboard rawdata confirmed the weighted parent had no middleware attached to the child.

### More

- [x] Added/updated tests

Fixes https://github.com/traefik/hub-issues/issues/2797


juliens:
I don't think this is the right approach. Middlewares should only be allowed directly on LoadBalancerSpec when it references a Kubernetes Service — not a TraefikService. If you want to attach service middlewares to a TraefikService, they should be defined on the target service itself.

We could also emit a warning when a TraefikService is used with middlewares set on it, to avoid the "bug effect."

PR #13020, opened 2026-04-22
Preserve request query on absolute-URL redirect

### LINKED ISSUE #12991: Preserve query parameters when using rewrite-target annotation
### Welcome!

### What did you do?

I'm using the `rewrite-target` annotation support from the Ingress NGINX provider to perform a redirect using an absolute URL (which is already supported by the Traefik provider:
nginx.ingress.kubernetes.io/rewrite-target: "{{ include "namespace.PortalBaseURL" . }}/site/$1"

The capture group `$1` contains both additional path segments, as well as query parameters. I am expecting parity with Ingress NGINX behavior which retains both the path segments and query parameters within the final generated URL.

### What did you see instead?

The redirect successfully happens with an HTTP 302, but because the [rewrite_target.go file](https://github.com/traefik/traefik/blob/master/pkg/middlewares/ingressnginx/rewritetarget/rewrite_target.go#L62) uses currentPath := req.URL.RawPath as the starting point for the rewrites, the If statement that checks for resolving to an absolute URL never has an opportunity to see the query parameters.

This is an incompatibility compared to Ingress NGINX, which handles query parameters properly.

`tag: experimental-v3.7`

Example ingress configuration:

```yaml
apiVersion: "networking.k8s.io/v1"
kind: "Ingress"
metadata:
  name: "web-subreport-redirect"
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: "{{ include "namespace.PortalBaseURL" . }}/site$1"
spec:
  ingressClassName: nginx
  tls:
    - hosts:
        - website.domain.com
      secretName: "website-secret"
  rules:
    - host: website.domain.com
      http:
        paths:
            # This path will scope redirects to URIs with at least this structure.
            # Notably, this includes ALL subpaths to .../view.
          - path: "/(report/view/.*)"
            pathType: "ImplementationSpecific"
            # The backend is not used here, but required by the specification
            backend:
              service:
                name: "defaultbackend"
                port:
                  number: 80
```

_No response_

SAY-5:
## Summary

Fixes #12991.

ingress-nginx's `rewrite-target` annotation keeps the incoming query on both relative rewrites and absolute-URL redirects. In the Traefik port, `ServeHTTP` takes `currentPath = req.URL.RawPath` (or `req.URL.EscapedPath()`), which doesn't carry the query string, so when the replacement resolves to an absolute URL and Traefik issues a 302 the `Location` header is query-less. The reporter's capture group intentionally expected `$1` to cover both additional path segments and query; this fix reaches parity on the absolute-URL redirect path.

## Fix

After parsing `newTarget` in the absolute-URL branch of `pkg/middlewares/ingressnginx/rewritetarget/rewrite_target.go`, if the parsed URL has no query and the inbound request does, carry `req.URL.RawQuery` over. If the rewrite target already has a query (e.g. `Replacement: "https://bar.example.org/some/path?tenant=acme"`), the rewrite wins — matching nginx precedence.

## Tests

Two table-driven cases added to `TestRewriteTarget`:

- `full URL replacement preserves incoming query` — request `/some/path?foo=bar&baz=qux` with replacement `https://bar.example.org/some/path` now redirects to `https://bar.example.org/some/path?foo=bar&baz=qux` (previously the query was dropped).
- `full URL replacement keeps rewrite-supplied query over request query` — pins the precedence rule so the rewrite's query wins when both exist.

`go test ./pkg/middlewares/ingressnginx/rewritetarget/...` green; `go vet` clean. Relative (non-absolute) rewrites and the existing 302 test cases are untouched.

### DIFF
--- a/pkg/middlewares/ingressnginx/rewritetarget/rewrite_target.go
+++ b/pkg/middlewares/ingressnginx/rewritetarget/rewrite_target.go
@@ -112,16 +112,21 @@ func (rt *rewriteTarget) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	// Only issue a 302 redirect if the replacement template itself is an absolute URL.
 	// Prevent user-controlled capture group content from injecting an absolute URL redirect.
 	if rt.absoluteURLRedirect {
+		newTargetURL, err := url.Parse(newTarget)
+		if err != nil {
+			http.Error(rw, http.StatusText(http.StatusInternalServerError), http.StatusInternalServerError)
+			return
+		}
 		// Carry the incoming query string through the redirect so an
 		// absolute-URL rewrite behaves like ingress-nginx, which preserves
 		// query parameters on rewrite-target redirects. Only fill it in
 		// when the rewrite itself didn't supply one, matching the nginx
 		// precedence of rewrite-set query over request query.
-		if parsed.RawQuery == "" && req.URL.RawQuery != "" {
-			parsed.RawQuery = req.URL.RawQuery
-			newTarget = parsed.String()
+		if newTargetURL.RawQuery == "" && req.URL.RawQuery != "" {
+			newTargetURL.RawQuery = req.URL.RawQuery
 		}
-		http.Redirect(rw, req, newTarget, http.StatusFound)
+
+		http.Redirect(rw, req, newTargetURL.String(), http.StatusFound)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


kevinpollet (approved):
Thanks 👍

mmatur (approved):
LGTM

Issue #12869, opened 2026-03-24
Expose disableRequestBuffer and disableResponseBuffer in config providers


V-Legendre:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

The buffering middleware has two options — `DisableRequestBuffer` and `DisableResponseBuffer` — that are present in the config struct and fully wired to the oxy layer, but excluded from all file-based configuration providers via struct tags:

From https://github.com/traefik/traefik/blob/master/pkg/config/dynamic/middlewares.go#L149 :
```go
DisableResponseBuffer bool `json:"disableResponseBuffer,omitempty" toml:"-" yaml:"-" label:"-" file:"-" kv:"-" export:"true"`
```

Today, enabling the buffering middleware (e.g. to enforce maxRequestBodyBytes for request size protection) implicitly buffers the entire response before forwarding it to the client. There is no way to opt out of response buffering through YAML, TOML, Docker labels, or KV providers.

This breaks streaming/chunked responses: the bufferWriter in oxy captures every byte from the upstream handler and only forwards them to the client via io.Copy after the handler has fully returned — defeating chunked transfer encoding and causing timeouts for long-running transfers.

## Use case

A common pattern is wanting request buffering (slow loris protection, request size limits) on all routes, while allowing specific routes to stream responses. Today the only workaround is removing the buffering middleware entirely from those routes and reimplementing request size limits at the application layer.

Being able to set `disableResponseBuffer: true` in a file-provider middleware definition would allow:
```yaml
http:
  middlewares:
    request-only-buffer:
      buffering:
        maxRequestBodyBytes: 8000000
        disableResponseBuffer: true
```

This would let users apply request buffering universally while preserving streaming on routes that need it.

## Proposed change

Change the struct tags from `toml:"-" yaml:"-" label:"-" file:"-" kv:"-"` to their actual field names, matching the existing pattern of the other fields in the same struct.

### Github issues 
- #7930 — Buffering middleware breaks chunked encoded requests
- #11568 — Buffering middleware generates 500 on chunked 302 response
- #2264 — Buffering/corruption of gRPC stream response

### Community forum threads describing the same gap (both stale, from 2024, unresolved):

[...]

PR #13029, opened 2026-04-22
feat(middleware/buffering): expose disableRequestBuffer and disableResponseBuffer

zalbiraw:
### What does this PR do?

Exposes DisableRequestBuffer and DisableResponseBuffer from oxy library to the buffer middleware config.

### Motivation

The buffering middleware already plumbs DisableRequestBuffer and DisableResponseBuffer through to the underlying oxy Buffer, but the two fields on dynamic.Buffering are tagged to be stripped from every user-facing configuration format (TOML, YAML, file, labels, KV), and the Kubernetes CRD type does not include them at all. The ingress-nginx compatibility provider was the only caller that could set them.

As a result, users cannot configure the buffering middleware to enforce maxRequestBodyBytes while still passing streaming responses (Server-Sent Events, gRPC streaming, Kubernetes watch long-polls) through untouched. With default settings, the middleware buffers the entire response body before forwarding it, which stalls streams until the upstream closes the connection.

Expose the two flags in all providers:

- Remove the '-' directives on dynamic.Buffering so file, TOML, YAML, label, and KV providers parse them.
- Add the fields to the traefikio/v1alpha1.Buffering CRD type and copy them in createBufferingMiddleware.
- Regenerate the BufferingApplyConfiguration and the CRD schema YAML.
- Document the new options and the streaming use case.
- Extend the KV and label round-trip tests.

### More

- [ ] Added/updated tests


V-Legendre (commented):
Thanks for picking this up @zalbiraw !  Happy to help since it's the feature I'd opened #12869 for.

I checked out the branch and reproduced the two failing checks locally. Here's what I found:

**`validate-generate`** : this is the one to fix. Running `make generate && make generate-crd` on the branch produces changes in three generated files that aren't committed yet:

- `docs/content/reference/routing-configuration/other-providers/file.yaml` — missing the `disableRequestBuffer` / `disableResponseBuffer` lines
- `docs/content/reference/routing-configuration/other-providers/file.toml` — same two lines
- `integration/fixtures/k8s/01-traefik-crd.yml` — missing the CRD schema block for both fields (this file is `cp`'d from the CRD def by `script/code-gen.sh`)

The `dynamic-configuration/` reference files were updated, but these three (the `routing-configuration/other-providers/` pair + the integration CRD fixture copy) get regenerated too.

**`test-unit`** : looks like an unrelated flaky test: `TestServiceTCPHealthChecker_Launch/connection-only_server_toggling_unhealthy_to_healthy` in `pkg/healthcheck` (a timing-sensitive TCP health-check, fails on "connection refused").

The rest of the change looks good to me  (the tag fix, the CRD type + conversion and the unit coverage).

zalbiraw:
Thank you @V-Legendre,  I will get that done now!

PR #13030, opened 2026-04-23
Add limit-connections support

amazon7737:
### What does this PR do?
Adds support for the `nginx.ingress.kubernetes.io/limit-connections` annotation in the Kubernetes Ingress-Nginx provider. This maps directly to Traefik's `InFlightReq` middleware to limit the number of simultaneous requests allowed from a single IP.

### Motivation
NGINX's `limit-connections` is designed to restrict concurrent connections per IP address. To replicate this behavior natively, I mapped the annotation to the `InFlightReq` struct. 
- I configured the `SourceCriterion` with an empty `IPStrategy`. This ensures the in-flight requests are tracked per client IP and is configured to align with the NGINX spec.
- If the parsed value is `<= 0` or missing, it just returns early. This prevents Traefik from accidentally generating empty middlewares or blocking traffic.
- It only attaches the middleware to the HTTP router when explicitly requested by the user.

Issue #12631

### More
- [x] Added/updated unit tests (verified via `ingress-with-limit-connections.yml` fixture)
- [x] Added/updated code generation & linting (`make generate fmt lint test-unit` passed)
- [x] Added documentation for the annotation in the NGINX Ingress provider reference


amazon7737:
The CI failure in test-unit for `github.com/traefik/traefik/v3/pkg/provider/kubernetes/ingress` seems completely unrelated to my changes, as this PR only modifies the `ingress-nginx` provider. Also, the tests pass consistently on my local environment. Could you please take a look ? Thanks!

kevinpollet:
Hey @amazon7737,

Thanks for this contribution!

At first glance, limiting the number of HTTP requests and TCP connections (multiple HTTP requests can be made in a single TCP connection) does not seem to fit well, or maybe I missed something?

amazon7737:
Hey  @kevinpollet,

The NGINX Ingress `limit-connections` annotation maps directly to the standard NGINX `limit_conn` directive. According to the [official NGINX documentation](https://nginx.org/en/docs/http/ngx_http_limit_conn_module.html), `limit_conn` does not count all TCP connections:

> "Not all connections are counted. A connection is counted only if it has a request being processed by the server and the whole request header has already been read."
> "In HTTP/2 and HTTP/3, each concurrent request is considered a separate connection."

If we break down what this actually means in practice:
1. By excluding idle or Keep-Alive TCP connections, it only counts active requests.
2. The requirement that a "header has already been read" confirms this is tracking HTTP-level traffic, not pure TCP sockets.
3. With HTTP/2 multiplexing, if 10 concurrent requests are flowing through a single TCP connection, NGINX counts them as 10 connections, not 1.

This means NGINX's implementation is functionally limiting the number of active, concurrent HTTP requests per IP, rather than idle TCP sockets.

Traefik's `InFlightReq` middleware, as seen in [`inflight_req.go` (L45-L50)](https://github.com/traefik/traefik/blob/v3.0/pkg/middlewares/inflightreq/inflight_req.go#L45-L50), similarly tracks and restricts the number of simultaneous HTTP requests currently being processed per IP:

```go
// pkg/middlewares/inflightreq/inflight_req.go
handler, err := connlimit.New(next, sourceMatcher, config.Amount, 
    connlimit.Logger(logs.NewOxyWrapper(*logger)),
    connlimit.Verbose(logger.GetLevel() == zerolog.TraceLevel))
```

As a result, because both mechanisms ultimately restrict concurrent active HTTP requests, mapping this annotation to InFlightReq provides a 1:1 functional equivalent for overload protection.

kevinpollet (approved):
Thanks 👍

juliens (approved):
LGTM

Issue #192, opened 2016-02-11
link to Project Mantl from Cisco contains extra 'http://'


ookhoi:
The link to "Project Mantl from Cisco" contains a small error:

```
<li>Project <a href="http://http://mantl.io/">Mantl</a> from Cisco</li>
```

PR #13038, opened 2026-04-24
Fix duplicate logging fields [#192]

pvbouwel:
### What does this PR do?

It avoids using the logger from the context to avoid that logger being mutated.

### Motivation

Before this change log lines with "message": "Skipping service: no endpoints found" could have an unbounded amount of "namespace" and "ingress" fields creating a log processing problem. 

The logging problem grows with the number of services without endpoints. The problems caused include:
 - Log lines being truncated causing invalid JSON
 - Extreme log size growth

### More

- [ ] Added/updated tests => I haven't been able to craft a test to reproduce the issue.
- [ ] Added/updated documentation => Does not seem to apply

### Additional Notes

I had the issue on a live cluster and by doing similar changes (cluster was not yet on 3.6) and build a custom container the log issue was solved.

Both parts were needed namely:
 1)  ensures each call to loadConfigurationFromIngresses starts with a fresh context, preventing field accumulation across multiple Kubernetes events
  2) Direct use of log.With() avoids retrieving a potentially mutated logger from context
  
I do not understand why I need (2) because there it only seems to mutate a child logger but if I only have (1) then the logging still occurred.

This bug seems to be present from Traefik v3 onwards.

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
@@ -151,7 +151,7 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
 					// dropped events. This is fine, because we don't treat different
 					// event types differently. But if we do in the future, we'll need to
 					// track more information about the dropped events.
-					conf := p.loadConfiguration(ctxLog)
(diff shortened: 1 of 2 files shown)


nmengin:
Hey @pvbouwel,
Thanks for the contribution!

PR #13059, opened 2026-04-28
fix: avoid UDP timeout before first activity

### LINKED ISSUE #13044: UDP: new session Conn is closed before first packet forwards when udp.timeout is small (lastActivity zero-time race in pkg/udp/conn.go)
**Welcome:**

---

**What did you do?**

Configured a UDP entrypoint with a small `udp.timeout` value and sent UDP packets through it.

Minimal Traefik static config:

```yaml
entryPoints:
  testudp:
    address: ":19003/udp"
    udp:
      timeout: 1ms
```

Or via CLI flags:

```
--entrypoints.testudp.address=:19003/udp
--entrypoints.testudp.udp.timeout=1ms
```

A UDP backend listening on `:19003` is wired up via any provider (file/docker/etc.), then a client sends packets at any cadence:

```
nc -u <traefik-host> 19003
```

---

**What did you expect to see?**

Each UDP packet from the client is forwarded to the backend. The session's idle teardown (governed by `udp.timeout`) is irrelevant for forwarding the very first packet — that packet should always reach the backend regardless of how small `udp.timeout` is.

---

**What did you see instead?**

With `udp.timeout=1ms`, the overwhelming majority of client packets never reach the backend — `tcpdump` on the backend host shows nothing on the entrypoint port even though the client is sending steadily. Increasing `udp.timeout` to `100ms` or higher makes the symptom disappear and packets forward normally.

The dropped-packet rate scales with how short `udp.timeout` is (or how slow the host's goroutine scheduling / syscall latency is). Sub-millisecond and low-single-digit-millisecond timeouts are unusable; a few tens of ms is the safe floor on a typical host.

---

**Root cause**

Two things combine to produce a deterministic race for the first packet of every new UDP session:

1. `newConn` does not initialize `lastActivity`, so it starts as the Go zero time (`time.Time{}` = January 1, year 1 AD):

   https://github.com/traefik/traefik/blob/master/pkg/udp/conn.go#L194-L204

   ```go
   func (l *Listener) newConn(rAddr net.Addr) *Conn {
       return &Conn{
           listener:  l,
           rAddr:     rAddr,
           receiveCh: make(chan []byte),
           readCh:    make(chan []byte),
           sizeCh:    make(chan int),
           doneCh:    make(chan struct{}),
           timeout:   l.timeout,
           // lastActivity intentionally? unset — Go zero value (year 1 AD)
       }
   }
   ```

2. `readLoop` ticks every `c.timeout / 10` and closes the Conn whenever `time.Now().After(c.lastActivity.Add(c.timeout))`:

   https://github.com/traefik/traefik/blob/master/pkg/udp/conn.go#L267-L300

   ```go
   ticker := time.NewTicker(c.timeout / 10)
   ...
   case <-ticker.C:
       c.muActivity.RLock()
       deadline := c.lastActivity.Add(c.timeout)
       c.muActivity.RUnlock()
       if time.Now().After(deadline) {
           c.Close()
           return
       }
   ```

`lastActivity` is only updated *after* `Conn.Read` completes its `readCh`/`sizeCh` handshake (`pkg/udp/conn.go` `Read`, the `c.lastActivity = time.Now()` line) or after `Conn.Write`. For the first packet, that update is gated behind the proxy chain: `Listener.acceptCh` pickup → `Proxy.ServeUDP` → `net.Dial("udp", target)` → spawn two `connCopy` goroutines → `io.CopyBuffer(connBackend, conn, buffer)` → `conn.Read`. On a normal host this chain takes hundreds of microseconds to a few milliseconds (goroutine scheduling, UDP socket setup, syscalls).

Until that chain completes its first `Read`, the deadline check is `(year 1 AD) + timeout`, which `time.Now().After(...)` always reports as true. With `timeout=1ms` the ticker fires every 100µs, so it virtually always wins the race against the proxy chain and `c.Close()` evicts the Conn from `Listener.conns` before the buffered packet is ever copied to the backend socket. The packet is silently lost.

With `timeout=100ms` the ticker period is 10ms, comfortably longer than typical proxy-chain setup latency, so `lastActivity` gets stamped before any ticker check runs.

---

**Suggested fix**

Make the idle-timeout ticker a no-op until the session has registered at least one `Read` or `Write`. This guarantees first-packet delivery and aligns the configured `udp.timeout` with its documented purpose — idle teardown of *established* sessions, not setup-latency censorship of new ones.

Minimal change using `lastActivity`'s zero value as the sentinel for "no activity yet":

```go
// pkg/udp/conn.go — readLoop ticker cases (both selects)
case <-ticker.C:
    c.muActivity.RLock()
    last := c.lastActivity
    c.muActivity.RUnlock()
    if last.IsZero() {
        // No Read/Write has completed yet; session is mid-setup.
        // Idle timeout does not apply.
        continue
    }
    if time.Now().After(last.Add(c.timeout)) {
        c.Close()
        return
    }
```

Equivalently, an explicit `started bool` on `Conn` set in `Read`/`Write` alongside the `lastActivity` stamp. Either way, the proxy chain's setup latency (Accept → `net.Dial` → spawn `connCopy` → first `Read`) can no longer evict a Conn before its first packet has been forwarded, regardless of how small `udp.timeout` is set.

---

**Regression analysis**

The fix shifts setup-window cleanup off the ticker and onto the proxy chain's existing deferred-close cascade. That cascade is already load-bearing for every other failure mode and is independently guaranteed to fire in bounded time:

- `Proxy.ServeUDP` registers `defer conn.Close()` immediately on entry (`pkg/udp/proxy.go:26`). Fires on any return path, including panics — removes the Conn from `Listener.conns` and closes `doneCh`.
- `net.Dial("udp", ...)` cannot block indefinitely. UDP is connectionless, so Dial does no network round-trip; the only meaningful latency is DNS resolution, which is bounded by the platform resolver timeout.
- Once Dial returns, the two `connCopy` goroutines spawn immediately. Each calls `io.CopyBuffer`, whose first action is a `Read`. Either:
  - the `Read` lands → `lastActivity` is stamped → ticker is "armed" and resumes its normal idle-teardown role; or
  - the `Read` (or its paired `Write`) errors → `connCopy` returns → `defer dst.Close()` (`pkg/udp/proxy.go:57`) closes the Conn → cascade.

So `ServeUDP` is guaranteed to either reach first-`Read`-or-`Write` activity (which arms the ticker) or terminate via a deferred `Close`, in bounded time. The ticker is not load-bearing for setup-window cleanup; it just incidentally also fires today because of the year-1-AD math.

Dead-backend behavior is unchanged. Once first activity stamps `lastActivity`, `udp.timeout` governs idle teardown exactly as documented. ICMP `ECONNREFUSED` from a dead backend still propagates through `connCopy`'s read error → `defer dst.Close()` cascade. A continuously-sending client to a dead backend keeps the session alive indefinitely — but that is the documented behavior of an idle timeout (which `udp.timeout` is) and is true today as well.

The single invariant the maintainers should preserve: `Proxy.ServeUDP` (and any future code path that calls `Listener.Accept`) must reach either first `Read`/`Write` or a deferred `Close` in bounded time. Today that is true. If a future change introduced a blocking path between `Accept` and the first `connCopy` spawn that bypassed `defer conn.Close()`, the fix would let such a Conn leak — but that would represent a separate regression in the proxy code, not in the timeout logic.

---

**Output of `traefik version`**

Reproduced on `v2.8.5` (the version we run); the affected code is unchanged on every released branch I checked — `v2.11.43`, `v3.6.14`, `v3.7.0-rc.2`, and current `master` — `lastActivity` is uninitialized in `newConn` and the ticker uses `c.timeout / 10` in all of them.

---

**What is your environment & configuration?**

```yaml
# Static configuration (relevant parts only)
entryPoints:
  testudp:
    address: ":19003/udp"
    udp:
      timeout: 1ms

log:
  level: DEBUG
```

Any UDP backend (e.g. `socat -u UDP-RECV:19003,reuseaddr OPEN:/dev/null`) wired via any provider is sufficient to reproduce.

---

**If applicable, please paste the log output in DEBUG level**

I can attach DEBUG logs from a clean repro on request, but the bug is fully explained from the code paths above and reproduces deterministically with the minimal config shown.

cyphercodes:
### What does this PR do?

Prevents a new UDP session from being closed by the idle-timeout ticker before any read or write activity has been recorded. Once the first read or write completes, the existing idle timeout behavior applies as before.

Fixes #13044

### Motivation

With a very small `udp.timeout`, `lastActivity` starts as the Go zero time, so the first timeout tick can close the `Conn` before the first queued packet is read and forwarded.

### More

- [x] Added/updated tests

### Additional Notes

Tested with:

- `go test ./pkg/udp -count=1`

### DIFF
--- a/pkg/udp/conn.go
+++ b/pkg/udp/conn.go
@@ -273,10 +273,7 @@ func (c *Conn) readLoop() {
 			case msg := <-c.receiveCh:
 				c.msgs = append(c.msgs, msg)
 			case <-ticker.C:
-				c.muActivity.RLock()
-				deadline := c.lastActivity.Add(c.timeout)
-				c.muActivity.RUnlock()
-				if time.Now().After(deadline) {
+				if c.hasTimedOut() {
 					c.Close()
 					return
 				}
@@ -293,15 +290,24 @@ func (c *Conn) readLoop() {
 		case msg := <-c.receiveCh:
 			c.msgs = append(c.msgs, msg)
 		case <-ticker.C:
-			c.muActivity.RLock()
-			deadline := c.lastActivity.Add(c.timeout)
-			c.muActivity.RUnlock()
-			if time.Now().After(deadline) {
+			if c.hasTimedOut() {
 				c.Close()
 				return
 			}
 		}
 	}
+}
+
+func (c *Conn) hasTimedOut() bool {
+	c.muActivity.RLock()
+	lastActivity := c.lastActivity
+	c.muActivity.RUnlock()
+
+	if lastActivity.IsZero() {
+		return false
+	}
+
+	return time.Now().After(lastActivity.Add(c.timeout))
 }
 
 func (c *Conn) close() {
--- a/pkg/udp/conn_test.go
+++ b/pkg/udp/conn_test.go
@@ -160,12 +160,89 @@ func TestListenWithZeroTimeout(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestTimeoutDoesNotCloseBeforeFirstRead(t *testing.T) {
+	ln, err := Listen(net.ListenConfig{}, "udp", ":0", time.Millisecond)
+	require.NoError(t, err)
+	defer func() {
+		err := ln.Close()
+		require.NoError(t, err)
+	}()
+
+	accepted := make(chan *Conn)
+	go func() {
+		conn, err := ln.Accept()
+		require.NoError(t, err)
+		accepted <- conn
+	}()
+
+	udpConn, err := net.Dial("udp", ln.Addr().String())
+	require.NoError(t, err)
+
+	_, err = udpConn.Write([]byte("TEST"))
+	require.NoError(t, err)
+
+	conn := <-accepted
+	time.Sleep(20 * time.Millisecond)
+
+	type readResult struct {
+		n   int
+		err error
+	}
+	resultCh := make(chan readResult)
+	go func() {
+		buf := make([]byte, 2048)
+		n, err := conn.Read(buf)
+		if err == nil {
+			assert.Equal(t, "TEST", string(buf[:n]))
+		}
+		resultCh <- readResult{n: n, err: err}
+	}()
+
+	select {
+	case result := <-resultCh:
+		require.NoError(t, result.err)
+		require.Equal(t, 4, result.n)
+	case <-time.Tick(time.Second):
+		t.Fatal("Timeout during first read")
+	}
+}
+
 func TestTimeoutWithRead(t *testing.T) {
 	testTimeout(t, true)
 }
 
-func TestTimeoutWithoutRead(t *testing.T) {
-	testTimeout(t, false)
+func TestTimeoutAfterFirstRead(t *testing.T) {
+	ln, err := Listen(net.ListenConfig{}, "udp", ":0", 50*time.Millisecond)
+	require.NoError(t, err)
+	defer func() {
+		err := ln.Close()
+		require.NoError(t, err)
+	}()
+
+	accepted := make(chan *Conn)
+	go func() {
+		conn, err := ln.Accept()
+		require.NoError(t, err)
+		accepted <- conn
+	}()
+
+	udpConn, err := net.Dial("udp", ln.Addr().String())
+	require.NoError(t, err)
+
+	_, err = udpConn.Write([]byte("TEST"))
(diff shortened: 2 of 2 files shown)


Hossy:
Watching 👀
Thank you @cyphercodes !

kevinpollet:
Hey @cyphercodes,

Thanks for this contribution!

PR #13060, opened 2026-04-28
Add error log deduplication for Kubernetes providers

george-angel:
### What does this PR do?

When broken IngressRoute or Ingress resources reference missing services or services without endpoints, Traefik emits the same error logs on every configuration refresh cycle. This creates noise and can overwhelm log rate limiting.

Introduce a shared ErrorDeduper with a hardcoded 1-minute TTL. Duplicate errors within the window are suppressed. When the TTL expires and the error is allowed through again, a suppressed=N field indicates how many occurrences were skipped.

This covers:
- Ingress provider: Cannot create service, Skipping service
- CRD provider (HTTP/TCP/UDP): Cannot create service, kubernetes service not found, no servers found

### Motivation

I want to get a feel for appetite for this change. We run heavily multi-tenant cluster. A lot of business teams that self-manage Namespaces.

We don't care about broken Service / Ingress config until a team comes looking to debug. Currently there is a lot of noise that is not very useful, we strive for correct config but inevitably we run invalid config and this can be costly logs wise.

### More

- [x] Added/updated tests

### Additional Notes

This is a first light stab, I wan to mainly see if this is something you would consider. Thanks!

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes.go
+++ b/pkg/provider/kubernetes/crd/kubernetes.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/cenkalti/backoff/v4"
 	"github.com/mitchellh/hashstructure"
+	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	ptypes "github.com/traefik/paerser/types"
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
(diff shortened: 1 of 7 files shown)


mmatur:
Hi @george-angel, thanks for the PR.

Since you're looking to gauge appetite first, the right path is to open a proposal issue so we can discuss the design and the broader use case before any implementation work.

Closing this for now, feel free to link back here once the proposal is up.

PR #13065, opened 2026-04-29
fix(ingress-nginx): make trailing slash optional in rewrite-target regex

### LINKED ISSUE #12982: rewrite-target full URL still appends path suffix on Prefix ingress without regex (regression case after #12905)
### Welcome!

### What did you do?

Running the ingress-nginx compatibility E2E tests on the `v3.7` branch (commit `2b9ffc4`, image `traefik/traefik:experimental-v3.7`). Two ingresses with the same annotation:

```
nginx.ingress.kubernetes.io/rewrite-target: https://bar.example.org/$1
```

1. **Case A** — `pathType: Prefix`, `path: /original` (no regex in path)
2. **Case B** — `pathType: ImplementationSpecific`, `path: /original/(.*)` (regex capture)

Sending requests to paths below the prefix (e.g. `/original/other`, `/original/a/b/c`).

### What did you see instead?

**Case A** (Prefix, no regex) — Traefik still appends the request path suffix with a leading extra slash, while NGINX strips it to `/` (no capture group so `$1` is empty):

| Request | NGINX `Location` | Traefik `Location` |
|---|---|---|
| `/original` | `https://bar.example.org/` | `https://bar.example.org/` ✓ |
| `/original/other` | `https://bar.example.org/` | `https://bar.example.org//other` ✗ |
| `/original/a/b/c` | `https://bar.example.org/` | `https://bar.example.org//a/b/c` ✗ |

**Case B** (regex path `/original/(.*)`) — request `/original` returns **404** on Traefik (expected `302` matching NGINX).

Same pattern as #12931 (closed as duplicate of #12911, fixed by #12905). #12905 solves the case where the request path **equals** the ingress path, but not the case where the request has additional path segments under a `Prefix` match without a regex capture.

```console
$ docker run traefik/traefik:experimental-v3.7 version
Version:      2b9ffc426122b89bf38139cebfd7dfb8d5007634
Codename:     cheddar
Go version:   go1.25.9
Built:        2026-04-14_07:54:14AM
OS/Arch:      linux/arm64
```

Ingress (Case A):

```yaml
kind: Ingress
metadata:
  name: full-path-test
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: "https://bar.example.org/\$1"
spec:
  ingressClassName: traefik
  rules:
    - host: full-path-test.traefik.local
      http:
        paths:
          - path: /original
            pathType: Prefix
            backend:
              service:
                name: whoami
                port:
                  number: 80
```

Traefik Helm values:

```yaml
providers:
  kubernetesIngressNginx: true
  kubernetesCRD: true
  kubernetesGateway: true
```

Reproduction available in https://github.com/traefik/ingress-nginx-migration — suite \`TestRewriteTargetSuite\`, tests \`TestFullPathRewriteNoRegexPath\` and \`TestFullPathRewriteWithRegexPath\`. Can attach full DEBUG logs on request.

blackwell-systems:
## What does this PR do?

When `pathType: ImplementationSpecific` with a regex capture path like `/original/(.*)` is used alongside `rewrite-target: https://bar.example.org/$1`, requests to the base path `/original` (without trailing slash) return 404. Requests to `/original/other` work correctly.

nginx handles this because it does a prefix match before the regex. Traefik only does regex matching, so the regex `/original/(.*)` requires the trailing slash to match, and `/original` doesn't have one.

### Root cause

In `applyRewriteTargetConfiguration`, `rulePath` (the raw ingress path) is passed directly as `Regex`. For `/original/(.*)`, the regex is strict about the slash before the capture group.

### Fix

Detect path regexes that end with a literal slash followed by a capture group (e.g. `/(.*)`), and make the slash optional: `/original/(.*)` becomes `/original/?(.*)`.

A compiled regex (`trailingSlashCaptureRegex`) matches only the trailing `/(...)` pattern, so interior path segments like `/(prefix)/(sub)` and alternation groups like `(/|$)(.*)` are not affected.

### Reproduction from #12982

| Request | Before (404) | After |
|---|---|---|
| `/original` | 404 (regex doesn't match) | 302 to `https://bar.example.org/` |
| `/original/other` | 302 to `https://bar.example.org/other` | 302 to `https://bar.example.org/other` (unchanged) |

Fixes #12982

### DIFF
--- a/pkg/middlewares/ingressnginx/rewritetarget/rewrite_target_test.go
+++ b/pkg/middlewares/ingressnginx/rewritetarget/rewrite_target_test.go
@@ -220,6 +220,26 @@ func TestRewriteTarget(t *testing.T) {
 			expectedStatusCode:  http.StatusFound,
 			expectedRedirectURL: "https://portal.example.org/site/foo/bar?tenant=acme&id=42",
 		},
+		{
+			desc: "regex with optional slash matches base path without trailing slash",
+			path: "/original",
+			config: dynamic.RewriteTarget{
+				Regex:       `/original/?(.*)`,
+				Replacement: "https://bar.example.org/$1",
+			},
+			expectedStatusCode:  http.StatusFound,
+			expectedRedirectURL: "https://bar.example.org/",
+		},
+		{
+			desc: "regex with optional slash matches path with suffix",
+			path: "/original/other",
+			config: dynamic.RewriteTarget{
+				Regex:       `/original/?(.*)`,
+				Replacement: "https://bar.example.org/$1",
+			},
+			expectedStatusCode:  http.StatusFound,
+			expectedRedirectURL: "https://bar.example.org/other",
+		},
 	}
 
 	for _, test := range testCases {
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
@@ -83,6 +83,10 @@ var (
 	strictPathTypeRegexp = regexp.MustCompile(`(?i)^/[[:alnum:]._\-/]*$`)
 	// The same regexp used in ingress-nginx: https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/annotations/parser/validators.go#L77
 	regexPathWithCapture = regexp.MustCompile(`^/?[-._~a-zA-Z0-9/$:]*$`)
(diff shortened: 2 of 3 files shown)


kevinpollet:
Hey @blackwell-systems,

Thanks for this contribution!

As this pull request is a fix, please could you rebase this pull request on branch v3.7?

blackwell-systems:
Closing to resubmit against v3.7 branch as requested by @kevinpollet.

PR #13062, opened 2026-04-28
Use a metamodel to generate dynamic configuration in ingress-nginx

juliens:
### What does this PR do?

This PR changes the way we create dynamic configuration from ingresses in the ingress-nginx provider.
It uses 2 phases, one to construct the metamodel, and one to create the dynamic configuratino from this metamodel.

### Motivation

Improve the maintainability.

### More

~- [ ] Added/updated tests~
~- [ ] Added/updated documentation~

### Additional Notes

Assisted-by: Claude
Co-authored-by: Kevin Pollet <pollet.kevin@gmail.com>

PR #13066, opened 2026-04-29
Add regression test for ingress default backend without rules

mmatur:
### What does this PR do?

Adds a regression test that exercises an Ingress with `spec.defaultBackend` set and no `spec.rules`. The test asserts that the provider emits the expected catch-all `default-backend` and `default-backend-tls` routers (with `PathPrefix("/")` and `Priority: math.MinInt32`), the corresponding `default-backend` service backed by the ingress default backend, and the per-ingress `ServersTransport`.

A new fixture `fixtures/ingresses/ingress-with-default-backend-no-rules.yml` is introduced to drive the case.

### Motivation

This code path was not covered by any existing fixture. Adding a test locks in the current behavior so that any future refactor (e.g. the metamodel rework in #13062) cannot silently drop the catch-all routers for an ingress whose only routing intent is its default backend.

### More

- [x] Added/updated tests


gndz07 (approved):

juliens (approved):
LGTM

PR #13071, opened 2026-04-29
Add test for ingress with default backend and other annotations for ingress-nginx provider

gndz07:
### What does this PR do?

Add a test for an ingress that has a default backend on an ingress level and other annotations (like backend-protocol and session affinity), make sure that they apply these annotations.

### Motivation

Add more test coverage.

### More

- [x] Added/updated tests


rtribotte (approved):
LGTM

PR #13077, opened 2026-04-29
Make ipAllowList lookup fast for large lists

giner:
Replace O(n) linear scan with O(1) hash map for single IPs and O(log n) binary search for CIDR ranges. Ranges are sorted and merged during initialization to minimize lookup time.

The original linear scan becomes a bottleneck with large allow lists, especially on the no-match path which scans every entry. The optimized lookup is essentially constant time regardless of list size.

Benchmark comparison (10K-100K entries):
```
goos: linux
goarch: amd64
pkg: github.com/traefik/traefik/v3/pkg/ip
cpu: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
```

BenchmarkContainsIP Before the change:
```
MatchInSingleIPList_100k-8         4446   248490 ns/op  0 B/op  0 allocs/op
NoMatchInSingleIPList_100k-8       2042   514240 ns/op  0 B/op  0 allocs/op
MatchInCIDRList_10k-8             25490    44886 ns/op  0 B/op  0 allocs/op
NoMatchInCIDRList_10k-8           13095    87495 ns/op  0 B/op  0 allocs/op
MatchInCIDRList_100k-8             2222   549185 ns/op  0 B/op  0 allocs/op
NoMatchInCIDRList_100k-8            777  1534752 ns/op  0 B/op  0 allocs/op
MatchInMixedList_SingleIP_100k-8  10090   123441 ns/op  0 B/op  0 allocs/op
MatchInMixedList_CIDR_100k-8       2258   552287 ns/op  0 B/op  0 allocs/op
NoMatchInMixedList_100k-8          1359   846405 ns/op  0 B/op  0 allocs/op
```

BenchmarkContainsIP After the change:
```
MatchInSingleIPList_100k-8        46649944   25.64 ns/op  0 B/op  0 allocs/op
NoMatchInSingleIPList_100k-8      50158794   23.51 ns/op  0 B/op  0 allocs/op
MatchInCIDRList_10k-8              8956461  121.5  ns/op  0 B/op  0 allocs/op
NoMatchInCIDRList_10k-8            9507897  118.0  ns/op  0 B/op  0 allocs/op
MatchInCIDRList_100k-8             9087830  133.1  ns/op  0 B/op  0 allocs/op
NoMatchInCIDRList_100k-8           8594448  136.2  ns/op  0 B/op  0 allocs/op
MatchInMixedList_SingleIP_100k-8  47374381   24.36 ns/op  0 B/op  0 allocs/op
MatchInMixedList_CIDR_100k-8       8320868  133.4  ns/op  0 B/op  0 allocs/op
NoMatchInMixedList_100k-8          9360322  130.1  ns/op  0 B/op  0 allocs/op
```

### What does this PR do?

Improves IP lookup performance when working with large allow lists 

### Motivation

The current (linear lookup) implementation quickly becomes a bottleneck when IP allow lists grow

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/ip/checker.go
+++ b/pkg/ip/checker.go
@@ -5,13 +5,19 @@ import (
 	"fmt"
 	"net"
 	"net/netip"
-	"strings"
+	"slices"
 )
 
 // Checker allows to check that addresses are in a trusted IPs.
 type Checker struct {
-	authorizedIPs    []*net.IP
-	authorizedIPsNet []*net.IPNet
+	authorizedIPs    map[netip.Addr]struct{}
+	authorizedRanges []ipRange
+}
+
+// ipRange represents a contiguous range of IP addresses.
+type ipRange struct {
+	start netip.Addr
+	end   netip.Addr
 }
 
 // NewChecker builds a new Checker given a list of CIDR-Strings to trusted IPs.
@@ -20,28 +26,44 @@ func NewChecker(trustedIPs []string) (*Checker, error) {
 		return nil, errors.New("no trusted IPs provided")
 	}
 
-	checker := &Checker{}
+	checker := &Checker{
+		authorizedIPs: make(map[netip.Addr]struct{}),
+	}
 
 	for _, ipMask := range trustedIPs {
-		if ipAddr := net.ParseIP(ipMask); ipAddr != nil {
-			checker.authorizedIPs = append(checker.authorizedIPs, &ipAddr)
+		if addr, err := netip.ParseAddr(ipMask); err == nil {
+			checker.authorizedIPs[canonicalAddr(addr)] = struct{}{}
 			continue
 		}
 
-		_, ipAddr, err := net.ParseCIDR(ipMask)
+		prefix, err := netip.ParsePrefix(ipMask)
 		if err != nil {
-			return nil, fmt.Errorf("parsing CIDR trusted IPs %s: %w", ipAddr, err)
+			return nil, fmt.Errorf("parsing CIDR trusted IPs %s: %w", ipMask, err)
 		}
-		checker.authorizedIPsNet = append(checker.authorizedIPsNet, ipAddr)
+
+		startAddr := canonicalAddr(prefix.Masked().Addr())
+
+		if prefix.IsSingleIP() {
+			checker.authorizedIPs[startAddr] = struct{}{}
+			continue
+		}
+
+		endAddr := canonicalAddr(networkLastAddress(prefix))
+
+		checker.authorizedRanges = append(checker.authorizedRanges, ipRange{
+			start: startAddr,
+			end:   endAddr,
+		})
 	}
+
+	// Sort and merge overlapping ranges for efficient binary search
+	checker.mergeRanges()
 
 	return checker, nil
 }
 
 // IsAuthorized checks if provided request is authorized by the trusted IPs.
 func (ip *Checker) IsAuthorized(addr string) error {
-	var invalidMatches []string
-
 	host, _, err := net.SplitHostPort(addr)
 	if err != nil {
 		host = addr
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hey @giner,

Thanks for this contribution!

tjps:
Mirroring my thoughts here for visibility on the active PR:

https://github.com/traefik/traefik/pull/13246#issuecomment-4566536448

giner:
I've realized that benchmark tests for CIRDs were not exactly correct because adjacent CIDRs were merged and that resulted in very fast lookups. I've pushed a new commit with the updated tests.

Is it fine if I rebase on master,  merge these two commits and update the original commit message?

kevinpollet:
Hello @giner,

> Is it fine if I rebase on master, merge these two commits and update the original commit message?

Yes, you can rebase it and merge the commits.

pkg/ip/checker.go
```diff
@@ -5,13 +5,19 @@ import (
...
-	"strings"
+	"slices"
 )
 
 // Checker allows to check that addresses are in a trusted IPs.
 type Checker struct {
-	authorizedIPs    []*net.IP
-	authorizedIPsNet []*net.IPNet
+	authorizedIPs    map[netip.Addr]struct{}
+	authorizedRanges []ipRange
```
tjps, line 14: Instead of maintaining and checking two separate structures, consider using a prefix trie like so: 

https://github.com/traefik/traefik/pull/13246/changes#diff-10ec15a35787e5e713f6260a53016f8d8db61ff30187d887ea9253344d8366e1R15

Tries are O(n) in the length of the checked string, which is a fixed 4 or 16 bytes in this case (IPv4 or IPv6 address). In practice this will be on the same order as the O(1) lookup in the hashmap for single addresses, and faster than the O(log(n)) in the case of checking many address ranges.

It also leads to fewer lines of code, as the entirety of the check is handled by the prefix trie datastructure.

I don't care about attribution, if you'd like to use the commits from the linked branch please feel free. 

Thanks for considering.

giner: Hi @tjps, thank you for the suggestion. I've run the benchmarks and found no significant difference in performance (~trie is slightly faster on no-match but the same or slower on match~ trie shows the same or lower performance on the benchmarks added in this PR) ~however~ and it introduces an external dependency. I find both implementations suitable in solving the problem and I don't have a strong opinion on which one to choose so I'd devote this decision to the maintainers.

giner: I've run the updated tests and found trie being the same or slower in all cases so I've updated the comment above:
```
goos: linux
goarch: amd64
pkg: github.com/traefik/traefik/v3/pkg/ip
cpu: Intel(R) Core(TM) i7-10610U CPU @ 1.80GHz
BenchmarkContainsIP_MatchInSingleIPList_100k-8         	 7360812	       160.6 ns/op
BenchmarkContainsIP_NoMatchInSingleIPList_100k-8       	24588618	        48.86 ns/op
BenchmarkContainsIP_MatchInCIDRList_10k-8              	 9772454	       122.3 ns/op
BenchmarkContainsIP_NoMatchInCIDRList_10k-8            	 9813638	       122.2 ns/op
BenchmarkContainsIP_MatchInCIDRList_100k-8             	 7480167	       160.0 ns/op
BenchmarkContainsIP_NoMatchInCIDRList_100k-8           	 7462056	       160.6 ns/op
BenchmarkContainsIP_MatchInMixedList_SingleIP_100k-8   	 4368682	       274.9 ns/op
BenchmarkContainsIP_MatchInMixedList_CIDR_100k-8       	 8122760	       148.4 ns/op
BenchmarkContainsIP_NoMatchInMixedList_100k-8          	 4372549	       267.6 ns/op
```

PR #13084, opened 2026-04-30
Expose DisableCompression on ServersTransport

### LINKED ISSUE #13075: Expose DisableCompression on ServersTransport to allow upstream gzip passthrough
### Welcome!

**Problem**
When a client sends a request without an Accept-Encoding header, Traefik should forward the upstream response as-is to the client — including the original Content-Encoding: gzip header and the original gzipped body — matching nginx and other passthrough reverse proxies.

**What did you see instead?**

Traefik silently:

1. Adds Accept-Encoding: gzip to the upstream request that the client never asked for
2. Receives a gzipped response from the backend
3. Decompresses the body in-memory
4. Strips the Content-Encoding: gzip response header
5. Returns the decompressed bytes to the client with no indication the upstream had compressed the response

This is Go's net/http.Transport default behavior (DisableCompression: false):

"If the Transport requests gzip on its own and gets a gzipped response, it's transparently decoded in the Response.Body."

It is appropriate for an HTTP **client**. It is incorrect for a **reverse proxy** because it changes the response payload that the client sees relative to what the backend produced, and strips a header that may carry semantic meaning for the client. Crucially, it creates an asymmetry: clients that do send Accept-Encoding get passthrough; clients that don't get rewritten responses.

**Reproduction (minimal)**

Backend: any service that returns Content-Encoding: gzip regardless of Accept-Encoding (Spring Boot with server.compression.enabled=true, etc.).

```
# Client without Accept-Encoding → Traefik strips the header and decompresses
$ curl -sv http://traefik/path -o /tmp/a.bin 2>&1 | grep -i content-encoding
# (no content-encoding header)
$ xxd /tmp/a.bin | head -1
00000000: 1ae0 5d0a ...   # raw payload, NOT gzip — bytes are different from upstream

# Client with Accept-Encoding → passthrough
$ curl -sv -H "Accept-Encoding: gzip" http://traefik/path -o /tmp/b.bin 2>&1 | grep -i content-encoding
< content-encoding: gzip
$ xxd /tmp/b.bin | head -1
00000000: 1f8b 0800 ...   # real gzip — passes through unchanged
```

**Root cause**

In [pkg/server/service/transport.go:212-222](https://github.com/traefik/traefik/blob/v3.6.13/pkg/server/service/transport.go#L212-L222) the http.Transport is built without setting DisableCompression: true. Go therefore applies its default "automatic Accept-Encoding + transparent decompression" behavior. There is no ServersTransport field, no static-config option, and no CLI flag to opt out.

**Proposed change**

Expose a boolean field on dynamic.ServersTransport that maps directly to http.Transport.DisableCompression. Naming follows the existing disableHTTP2 field on the same struct:

```yaml
kind: ServersTransport
metadata:
  name: passthrough
  namespace: default
spec:
  # When true, prevents Traefik from requesting compression with
  # "Accept-Encoding: gzip" on its own when the client did not send one,
  # and forwards upstream Content-Encoding/body unchanged. Maps directly
  # to http.Transport.DisableCompression. Default: false.
  # (Same naming convention as the existing disableHTTP2 field.)
  disableCompression: true
```

File provider equivalent:

```yaml
http:
  serversTransports:
    passthrough:
      disableCompression: true
```

The flag would directly map to http.Transport.DisableCompression. Default false to preserve current behavior.

**Why it matters**

- Compatibility with backends that always emit Content-Encoding: gzip
- Compatibility with HTTP clients that don't advertise compression but still want the encoded payload — curl without --compressed, libcurl with CURLOPT_ACCEPT_ENCODING unset, native C/C++ HTTP clients, embedded device clients, custom scripted tooling. Today these clients silently receive a different payload than the backend produced
- Migration parity with nginx (and other passthrough proxies), where this scenario works without configuration

Per libcurl's own documentation:

"The CURLOPT_ACCEPT_ENCODING must be set to any non-NULL value for content to be automatically decoded. If it is not set and the server still sends encoded content (despite not having been asked), the data is returned in its raw form and the Content-Encoding type is not checked."

i.e. clients are designed to receive encoded payloads opaquely when they don't opt in to decoding. Traefik's current behavior breaks this contract by removing the encoding before the client ever sees it.

**Environment**

Traefik version: v3.6.13 (also confirmed in v3.4+)
Provider: Kubernetes Gateway API + Kubernetes CRD
Deployment: Helm chart traefik-helm-chart v39.0.8

rupertgti:
### What does this PR do?

Adds a new `disableCompression` boolean field to `ServersTransport` (Kubernetes CRD, file provider and labels) that maps directly to Go's `http.Transport.DisableCompression`. When set to `true`, Traefik no longer requests gzip compression on its own to the upstream and no longer transparently decompresses upstream responses, so the original `Content-Encoding` header and response body are forwarded to the client as-is.

The new field follows the existing `disableHTTP2` naming convention on the same struct. Default is `false`, preserving current behavior.

### Motivation

Closes #13075.

Today, when a client sends a request without an `Accept-Encoding` header, Traefik silently:

1. Adds `Accept-Encoding: gzip` to the upstream request,
2. Receives a gzipped response from the backend,
3. Decompresses the body in-memory,
4. Strips the `Content-Encoding: gzip` response header,
5. Returns the decompressed bytes to the client with no indication the upstream had compressed the response.

This is Go's `net/http.Transport` default behavior (`DisableCompression: false`), which is appropriate for an HTTP client but surprising for a reverse proxy: the response payload that the client sees no longer matches what the backend produced. Clients that *do* send `Accept-Encoding` get pure passthrough; clients that don't get rewritten responses (curl without `--compressed`, libcurl with `CURLOPT_ACCEPT_ENCODING` unset, native C/C++ clients, embedded device clients, etc.).

There was no way to opt out of this behavior — `ServersTransport` did not expose `DisableCompression`, and there was no static-config or CLI flag. This PR exposes it.

### More

- [x] Added/updated tests

### Additional Notes

- Code changes are minimal: one new field on `dynamic.ServersTransport`, one new field on the CRD `ServersTransportSpec`, one extra line in `createRoundTripper` that wires `DisableCompression: cfg.DisableCompression` on the underlying `http.Transport`, and one extra line in the Kubernetes CRD provider that copies the spec value into the dynamic configuration.
- A new test (`TestDisableCompression` in `pkg/server/service/transport_test.go`) covers both the default and the new opt-in behavior, asserting both the response body content and the `Content-Encoding` header.
- The auto-generated files were regenerated with `make generate` and `make generate-crd` (CRD YAMLs, deepcopy, applyconfiguration with the `WithDisableCompression()` builder).
- Documentation updated in both the file-provider reference (`serverstransport.md`, `file.yaml`, `file.toml`) and the Kubernetes CRD reference (`crd/http/serverstransport.md`).
- An existing workaround for users today is to apply the `compress` middleware with `defaultEncoding: gzip` so that Traefik re-compresses the body that the Go transport had transparently decoded. It works but does redundant CPU work (decompress in transport, recompress in middleware) on every response — this PR removes the need for that workaround.


mmatur:
Hey @rupertgti ,

Thanks for this contribution!

rupertgti:
I fixed the check failed

rupertgti:
@mmatur Could you set to "review" again to check the test after my fixes? Thank you

PR #13087, opened 2026-04-30
Remove cross-provider sanitization for Kubernetes service loading

### LINKED ISSUE #13041: Add support for the 'Kind' attribute in ErrorPages middleware's service property on Kubernetes
### Welcome!

It would be nice if the ErrorPages middleware would support the `Kind` attribute in the `service` definition. Enabling the use of both Kubernetes Service and TraefikService to be used with ErrorPages.

I see that the CRD does define this field, and accepts the expected values in https://github.com/traefik/traefik/blob/f5efe1e69b38d8a98c16f8839c6dde3b5e79b52a/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml#L1248-L1253

But when Traefik parses the `service` field it does not seem to consider the `Kind` field and goes straight into looking it up as a Kubernetes service in 
- https://github.com/traefik/traefik/blob/9893e89628425c91faa1fa7768eff8f1ccdc596f/pkg/provider/kubernetes/crd/kubernetes.go#L673 calls
- https://github.com/traefik/traefik/blob/9893e89628425c91faa1fa7768eff8f1ccdc596f/pkg/provider/kubernetes/crd/kubernetes_http.go#L415 calls
- https://github.com/traefik/traefik/blob/9893e89628425c91faa1fa7768eff8f1ccdc596f/pkg/provider/kubernetes/crd/kubernetes_http.go#L555 is where it unconditionally looks up the Kubernetes service

TraefikService provides lot more flexibility (like configuring TLS options) compared to Kubernetes services which would be useful for errorpages just as well. Also in it's current state the CRD is confusing because it does define the Kind field but it has no effect.

rtribotte:
### What does this PR do?

This PR removes cross-provider sanitization for Kubernetes service loading in the Kubernetes CRD provider, and also refactors a bit the services loaders.
The refactoring opens for TraefikService service kind to be loaded with the ErrorPage CRD middleware.

### Motivation

To cleanup the code, and remove a unexpected handling of cross-provider reference.
And:
Fixes #13041

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/provider/kubernetes/crd/fixtures/with_error_page_traefik_service.yml
+++ b/pkg/provider/kubernetes/crd/fixtures/with_error_page_traefik_service.yml
@@ -1,0 +1,29 @@
+apiVersion: traefik.io/v1alpha1
+kind: TraefikService
+metadata:
+  name: errorpage-wrr
+  namespace: default
+
+spec:
+  weighted:
+    services:
+      - name: whoami
+        port: 80
+        weight: 1
+
+---
+apiVersion: traefik.io/v1alpha1
+kind: Middleware
+metadata:
+  name: errorpage
+  namespace: default
+
+spec:
+  errors:
+    status:
+      - "404"
+      - "500"
+    query: query
+    service:
+      name: errorpage-wrr
+      kind: TraefikService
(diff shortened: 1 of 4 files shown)


kevinpollet (approved):

juliens (approved):
LGTM

PR #13089, opened 2026-04-30
fix(ingress-nginx): make trailing slash optional in rewrite-target regex

### LINKED ISSUE #12982: rewrite-target full URL still appends path suffix on Prefix ingress without regex (regression case after #12905)

blackwell-systems:
Rebased onto v3.7 per @kevinpollet's request in #13065.

## What does this PR do?

When `pathType: ImplementationSpecific` with a regex capture path like `/original/(.*)` is used alongside `rewrite-target: https://bar.example.org/$1`, requests to the base path `/original` (without trailing slash) return 404.

nginx handles this because it does a prefix match before the regex. Traefik only does regex matching, so `/original/(.*)` requires the trailing slash to match.

### Fix

Detect path regexes ending with a literal slash followed by a capture group (e.g. `/(.*)`), and make the slash optional: `/original/(.*)` becomes `/original/?(.*)`. A compiled regex (`trailingSlashCaptureRegex`) ensures only the trailing `/(...)` is affected.

| Request | Before (404) | After |
|---|---|---|
| `/original` | 404 (regex doesn't match) | 302 to `https://bar.example.org/` |
| `/original/other` | 302 (unchanged) | 302 (unchanged) |

Fixes #12982

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes.go
@@ -73,6 +73,10 @@ var (
 	strictPathTypeRegexp = regexp.MustCompile(`(?i)^/[[:alnum:]._\-/]*$`)
 	// The same regexp used in ingress-nginx: https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/annotations/parser/validators.go#L77
 	regexPathWithCapture = regexp.MustCompile(`^/?[-._~a-zA-Z0-9/$:]*$`)
+	// trailingSlashCaptureRegex matches a trailing /(...) at the end of a path regex.
+	// Used to make the slash optional so "/original/(.*)" also matches "/original".
+	// See https://github.com/traefik/traefik/issues/12982
+	trailingSlashCaptureRegex = regexp.MustCompile(`/\(([^)]*)\)$`)
 )
 
 // Provider holds configurations of the provider.
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
@@ -3164,14 +3164,14 @@ func TestLoadIngresses(t *testing.T) {
 					Middlewares: map[string]*dynamic.Middleware{
 						"default-ingress-with-x-forwarded-prefix-three-groups-rule-0-path-0-rewrite-target": {
 							RewriteTarget: &dynamic.RewriteTarget{
-								Regex:            "/(prefix)/(sub)/(.*)",
+								Regex:            "/(prefix)/(sub)/?(.*)",
 								Replacement:      "/$3",
 								XForwardedPrefix: "/$1/$2",
 							},
 						},
 						"default-ingress-with-x-forwarded-prefix-three-groups-rule-0-path-0-tls-rewrite-target": {
 							RewriteTarget: &dynamic.RewriteTarget{
-								Regex:            "/(prefix)/(sub)/(.*)",
+								Regex:            "/(prefix)/(sub)/?(.*)",
 								Replacement:      "/$3",
 								XForwardedPrefix: "/$1/$2",
 							},
--- a/pkg/provider/kubernetes/ingress-nginx/middleware.go
+++ b/pkg/provider/kubernetes/ingress-nginx/middleware.go
@@ -256,7 +256,13 @@ func (p *Provider) buildRewriteTarget(loc *location) {
 		return
 	}
 
-	regex := loc.Path
+	// When the path ends with a trailing slash + capture group like "/(.*)"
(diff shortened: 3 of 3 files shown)


nmengin:
Hey @blackwell-systems,

Thanks for this contribution!

blackwell-systems:
Any updates on the review? Happy to address feedback.

nmengin:
Hey @blackwell-systems,
We're focused on other priorities right now, though, and we don't have an ETA yet—even though we plan to get to it.

gndz07:
Hi @blackwell-systems,

Thank you for opening the PR. After walking through the core issue and traced through the nginx-generated config to understand exactly why the behavior differs, we don't think that this is a behavior that Traefik should replicate.

When rewrite-target is set on an ingress with a regex path like `/original/(.*)`, ingress-nginx generates two location blocks for that server:

```
location ~* "^/original/(.*)" {
    rewrite "(?i)/original/(.*)" https://bar.example.org/$1 break;
    ...
}

location ~* "^/" {
    rewrite "(?i)/" https://bar.example.org/$1 break;
    proxy_pass http://upstream-default-backend;
    ...
}
```

The request to `/original` fails to match the first block (the regex requires a literal `/` before `(.*)`), so it falls through to the catch-all `~* "^/"`. There,` (?i)/` matches any URI, and because the replacement is an absolute URL, nginx immediately issues a 302. That is where the redirect comes from, as a side effect of the catch-all, not because of prefix match. This is not intentional routing behavior; it is a known bug in ingress-nginx and there is an [issue](https://github.com/kubernetes/ingress-nginx/issues/9485) on this matter. The catch-all picks the rewrite-target from one of the ingresses on the host arbitrarily. With two ingresses on the same host pointing to different absolute URLs as their rewrite-target, unmatched paths will silently redirect to the wrong destination.

Replicating this in Traefik would mean inheriting the same bug. Additionally, this behavior only occurs with absolute URL rewrite-targets. For relative ones (e.g. `/$1`), nginx returns 404 for `/original`, so making the slash unconditionally optional would also introduce a parity difference there.

Therefore, I'm closing this PR and the related issue as this is not a fixable behavior.

Don't hesitate to leave a comment or reopen this PR if you don't agree with the decision. Thank you for your contribution to help make Traefik a better product.

PR #13092, opened 2026-05-03
feat: Affinity Load Balancer

### LINKED ISSUE #12782: Feature request: path-based session affinity for HTTP load balancer
### Welcome!

I would like Traefik HTTP load balancing to support affinity based on a request-derived key, especially a path segment.

My use case is Xray XHTTP (SplitHTTP), where a single logical session is split into multiple HTTP requests:

- `GET /hello/<session-id>`
- `POST /hello/<session-id>/<seq>`

All requests sharing the same `<session-id>` must be routed to the same backend server.

Today, Traefik provides strategies such as WRR and HRW, and sticky sessions based on cookies. However:

- WRR does not preserve request affinity for this protocol
- cookie-based stickiness is not reliable for this type of non-browser client
- HRW is based on client IP, not on an application-level session key in the request path

What I expected is a built-in way to select the backend using a request-derived value, for example:

- a full request path
- a regex capture group from the path
- a header value
- or a custom hash key extracted from the request

For this use case, the ideal behavior would be something like:

- extract `<session-id>` from `/hello/<session-id>` and `/hello/<session-id>/<seq>`
- consistently hash on that extracted value
- route all requests with the same `<session-id>` to the same backend

StillNotUsefull:
### What does this PR do?

Adds a new `affinity` load balancing strategy to the `ServersLoadBalancer` that routes
requests to a consistent backend based on a key extracted from the incoming request,
rather than from client IP or a cookie.

The affinity key is resolved in the following priority order:
1. First capture group of a regex matched against the request path (`affinity.regex`)
2. Value of a named request header (`affinity.headerName`)
3. Full request path as fallback (deterministic, no session pinning)

Once a key is first seen, it is pinned to a backend via an in-memory ring using jump
consistent hashing. If the pinned backend is marked unhealthy, the session is evicted
and re-assigned on the next request.

### Motivation

Closes #12782.

Protocols like XHTTP (SplitHTTP) split a single logical session across multiple HTTP
requests that share a session identifier in the path, for example:

- `GET /hello/<session-id>`  opens the read stream
- `POST /hello/<session-id>/<seq>` carries sequential data chunks

All requests sharing the same `<session-id>` must reach the same backend. Existing
strategies do not support this:

- `wrr` and `p2c` have no affinity at all
- `hrw` hashes on client IP, not on an application-level session key
- Cookie-based sticky sessions are not reliable for non-browser clients

The `affinity` strategy solves this by deriving the routing key directly from the
request itself, making it suitable for any non-browser protocol where the session
identifier is carried in the path or a header.

### More
- [x] Added/updated tests

### Additional Notes

The strategy is wired into the existing `getLoadBalancerServiceHandler` switch in
`pkg/server/service/service.go` alongside `wrr`, `p2c`, `hrw`, and `leasttime`,
and follows the same structural patterns as those balancers (`namedHandler`, `status`
map, `RegisterStatusUpdater`, `wantsHealthCheck` flag).

Configuration example:

```yaml
http:
  services:
    my-service:
      loadBalancer:
        strategy: "affinity"
        affinity:
          regex: "^/hello/([^/]+)"      # extract session-id from path
          headerName: "X-Session-Id"    # fallback: use header if regex does not match
        servers:
          - url: "http://backend-1/"
          - url: "http://backend-2/"
          - url: "http://backend-3/"
```

Or via Docker labels:

```
traefik.http.services.my-service.loadbalancer.strategy=affinity
traefik.http.services.my-service.loadbalancer.affinity.regex=^/hello/([^/]+)
traefik.http.services.my-service.loadbalancer.affinity.headername=X-Session-Id
```


nmengin:
Hey @StillNotUsefull,

In the meantime, coudl rebase the PR on the master branch please?

PR #13093, opened 2026-05-04
Fix retry state leaking into mirror requests

### LINKED ISSUE #12988: Mirror + Retry: mirror goroutine triggers shouldRetry on main request (data race, response body corruption)
### Welcome!

### What did you do?

We identified this bug while running Traefik v3.6.11 with a Mirroring service (100% traffic mirrored to a shadow backend) combined with the built-in Retry middleware (3 attempts). Under moderate load, we started observing non-deterministic HTTP response body corruption on the main (non-mirror) path: JSON responses were being returned with garbage appended ({...valid JSON...}Internal Server Error), or with the body duplicated entirely. The issue was load-dependent and completely absent when either the Retry middleware or the Mirroring service was disabled, which pointed immediately to an interaction between the two.
To reproduce: configure a router with a Retry middleware (attempts: 3) and a Mirroring TraefikService (percent: 100) pointing to any backend. Send concurrent requests under load (we used ~50 parallel clients). A significant fraction of responses will have corrupted bodies — either trailing garbage or doubled content — despite the backend returning clean 200 OK responses.

### What did you see instead?

Instead of seeing regular JSON responses, we saw responses of the kind

{...valid JSON...}Internal Server Error

or

{...valid JSON1...}{...valid JSON2...}

v3.6.11

here's the list of arguments we pass to traefik:

```yaml
  0         - --global.checkNewVersion=false
  1         - --entryPoints.metrics.address=:9100/tcp
  2         - --entryPoints.traefik.address=:8080/tcp
  3         - --entryPoints.web.address=:8000/tcp
  4         - --entryPoints.websecure.address=:8443/tcp
  5         - --ping=true
  6         - --metrics.datadog=true
  7         - --metrics.datadog.address=unix:///var/run/datadog/dsd.socket
  8         - --metrics.datadog.addRoutersLabels=true
  9         - --providers.kubernetescrd
 10         - --providers.kubernetescrd.allowCrossNamespace=true
 11         - --providers.kubernetescrd.allowExternalNameServices=true
 12         - --providers.kubernetescrd.allowEmptyServices=true
 13         - --providers.kubernetescrd.nativeLBByDefault=true
 14         - --providers.kubernetesingress
 15         - --providers.kubernetesingress.allowEmptyServices=true
 16         - --providers.kubernetesingress.ingressendpoint.publishedservice=ingress-traefik/traefik
 17         - --providers.kubernetesingress.nativeLBByDefault=true
 18         - --providers.kubernetesgateway
 19         - --providers.kubernetesgateway.statusaddress.service.name=traefik
 20         - --providers.kubernetesgateway.statusaddress.service.namespace=ingress-traefik
 21         - --providers.kubernetesgateway.nativeLBByDefault=true
 22         - --entryPoints.web.http.redirections.entryPoint.to=:443
 23         - --entryPoints.web.http.redirections.entryPoint.scheme=https
 24         - --entryPoints.web.http.redirections.entryPoint.permanent=true
 25         - --entryPoints.websecure.http.tls=true
 26         - --log.format=json
 27         - --log.level=INFO
 28         - --accesslog=true
 29         - --accesslog.format=json
 30         - --accesslog.fields.defaultmode=keep
 31         - --accesslog.fields.headers.defaultmode=drop
 32         - --experimental.localPlugins.traefik-internal-jwt.moduleName=cognex.com/traefik-internal-jwt
 33         - --experimental.localPlugins.rewrite-body.moduleName=github.com/traefik/plugin-rewritebody
 34         - --experimental.localPlugins.traefik-rewrite-body.moduleName=cognex.com/traefik-rewrite-body
 35         - --experimental.localPlugins.static-response.moduleName=github.com/edgeworx/static-response-plugin
 36         - --experimental.localPlugins.corsmiddleware.moduleName=github.com/SergioFloresG/corsmiddleware
 37         - --experimental.localPlugins.traefik-status-retry.moduleName=cognex.com/traefik-status-retry
```

We run traefik on kubernetes.

The fix should be two small changes:

1. retry.go — add a StripRetryContext helper that returns a copy of the context with shouldRetryContextKey set to nil:
      func StripRetryContext(ctx context.Context) context.Context {
       return context.WithValue(ctx, shouldRetryContextKey{}, nil)
   }
   
2. mirror.go — call it in the mirror goroutine, right after the existing accesslog.DataTableKey strip:
      ctx := context.WithValue(r.Context(), accesslog.DataTableKey, nil)
   ctx = retry.StripRetryContext(ctx)  // ← the fix
   
The idea is symmetrical with what Traefik already does for the access log datatable: any context key that must not be shared between the main goroutine and the mirror goroutine needs to be explicitly stripped before the mirror runs. shouldRetryContextKey was simply missed.

We have not yet tried that fix, we have only tried to disable the built-in retry middleware entirely, which seems to work.

cyphercodes:
### What does this PR do?

Strips retry middleware state from mirrored request contexts before dispatching mirror backend calls. This prevents a mirror request from reusing the main request's retry callback and mutating the main response writer from the mirror goroutine.

A regression test now covers that mirrored requests do not receive a retry context when a mirroring service is served behind the retry middleware.

### Motivation

Fixes #12988.

Mirror requests already strip access-log context state to avoid concurrent mutation of main-request data. Retry context state has the same requirement: it is bound to the current retry attempt/response writer and must not be shared with mirror goroutines.

### More

- [x] Added/updated tests

### Additional Notes

Local verification:

- `go test ./pkg/server/service/loadbalancer/mirror -run TestMirrorRequestStripsRetryContext -count=1` (failed before the fix, passes after)
- `go test ./pkg/middlewares/retry ./pkg/server/service/loadbalancer/mirror -count=1`
- `git diff --check`

### DIFF
--- a/pkg/middlewares/retry/retry.go
+++ b/pkg/middlewares/retry/retry.go
@@ -56,6 +56,11 @@ func ContextShouldRetry(ctx context.Context) ShouldRetry {
 	return f
 }
 
+// StripRetryContext returns a context that hides any retry state from its parent context.
+func StripRetryContext(ctx context.Context) context.Context {
+	return context.WithValue(ctx, shouldRetryContextKey{}, nil)
+}
+
 // WrapHandler wraps a given http.Handler to inject the httptrace.ClientTrace in the request context when it is needed
 // by the retry middleware.
 func WrapHandler(next http.Handler) http.Handler {
--- a/pkg/server/service/loadbalancer/mirror/mirror.go
+++ b/pkg/server/service/loadbalancer/mirror/mirror.go
@@ -15,6 +15,7 @@ import (
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
 	"github.com/traefik/traefik/v3/pkg/healthcheck"
 	"github.com/traefik/traefik/v3/pkg/middlewares/accesslog"
+	retrymiddleware "github.com/traefik/traefik/v3/pkg/middlewares/retry"
 	"github.com/traefik/traefik/v3/pkg/safe"
 )
 
@@ -98,6 +99,7 @@ func (m *Mirroring) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			// Especially since it would result in unguarded concurrent reads/writes on the datatable.
 			// Therefore, we reset any potential datatable key in the new context that we pass around.
 			ctx := context.WithValue(r.Context(), accesslog.DataTableKey, nil)
+			ctx = retrymiddleware.StripRetryContext(ctx)
 
 			// When a request served by m.handler is successful, req.Context will be canceled,
 			// which would trigger a cancellation of the ongoing mirrored requests.
--- a/pkg/server/service/loadbalancer/mirror/mirror_test.go
+++ b/pkg/server/service/loadbalancer/mirror/mirror_test.go
@@ -7,12 +7,47 @@ import (
 	"net/http/httptest"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/traefik/traefik/v3/pkg/config/dynamic"
+	retrymiddleware "github.com/traefik/traefik/v3/pkg/middlewares/retry"
 	"github.com/traefik/traefik/v3/pkg/safe"
 )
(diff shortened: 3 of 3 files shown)


mmatur:
Hey @cyphercodes ,

Thanks for this contribution!

PR #13094, opened 2026-05-04
Add CrossProviderNamespaces option 

rtribotte:
### What does this PR do?

This PR adds a new CrossProviderNamespaces option to Kubernetes providers to control from which namespace Kubernetes resources are allowed to use cross-provider references.

Traefik offers the possibility to make references from one resource to another.

This is often used to reference routing resources from one provider to another (cross-provider references), 
but this mechanism also works from within one provider.

However, in the context of Kubernetes providers, those direct references (e.g.: `myservice@kubernetescrd`) allow a user to cross namespace boundaries, as well as exposing `@internal` services, that only the operator should be able to expose.

This new `crossProviderNamespaces` option restricts in which namespaces Kubernetes resources are allowed to use direct references.

### Motivation

To provide a mechanism for operators running multi-tenant clusters to avoid privilege escalation across namespace boundaries.

### More

- [x] Added/updated tests

### Additional Notes


gndz07 (approved):
LGTM

kevinpollet (approved):

PR #13096, opened 2026-05-04
Bump webui dependencies

gndz07:
### What does this PR do?

This PR updates dependencies for webui dashboard.
This PR also bump faency to v12.3.3, which removes some dependencies as peer dependencies, as they are not used directly in the components and we don't need to have them here. As consequence, we have to rename the package from `@traefiklabs/faency` to `@traefik-labs/faency`.

### Motivation

To have updated dependencies.

### More

- [ ] Added/updated tests

### DIFF
--- a/.github/workflows/template-webui.yaml
+++ b/.github/workflows/template-webui.yaml
@@ -2,7 +2,7 @@ name: Build Web UI
 on:
   workflow_call: {}
 env:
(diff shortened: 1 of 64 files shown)


rtribotte (approved):
LGTM

kevinpollet (approved):

PR #13099, opened 2026-05-04
Bump rollup to the latest version to allow building on loong64-musl

arrowd:
### What does this PR do?

This change bumps versions of rollup-related packages in yarn.lock

### Motivation

rollup featured loong64-musl support since v4.55.0, related PR is https://github.com/rollup/rollup/commit/72b0f0ee02f6766f8e59c1724700da6647c6b51f

### Additional Notes

This allowed me to successfully build traefik on linux-loong64, `make test-unit` passes.

### DIFF
--- a/webui/yarn.lock
+++ b/webui/yarn.lock
@@ -4470,296 +4470,177 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@rollup/rollup-android-arm-eabi@npm:4.39.0":
-  version: 4.39.0
-  resolution: "@rollup/rollup-android-arm-eabi@npm:4.39.0"
+"@rollup/rollup-android-arm-eabi@npm:4.60.2":
+  version: 4.60.2
+  resolution: "@rollup/rollup-android-arm-eabi@npm:4.60.2"
   conditions: os=android & cpu=arm
   languageName: node
   linkType: hard
 
-"@rollup/rollup-android-arm-eabi@npm:4.52.5":
-  version: 4.52.5
-  resolution: "@rollup/rollup-android-arm-eabi@npm:4.52.5"
-  conditions: os=android & cpu=arm
-  languageName: node
-  linkType: hard
-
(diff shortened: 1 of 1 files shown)


gndz07:
Hello @arrowd,

Thank you for opening this PR. However, we don’t think manually editing `yarn.lock` is the right approach for upgrading dependencies, since the lockfile should be generated by the package manager.

While we generally agree that keeping dependencies up to date is beneficial, the use case described in this PR does not currently seem strong enough to justify this change as-is.

If you’d like, you can open a proposal issue to discuss the use case further and gather community feedback so we can better assess its priority.

In any case, we should look for another way to upgrade the Rollup version.

arrowd:
> However, we don’t think manually editing yarn.lock is the right approach for upgrading dependencies, since the lockfile should be generated by the package manager.

That's exactly what I did - `yarn up -R rollup`. I'm not a JS developer, so I'm sorry if that's not the intended way to do it.

gndz07:
@arrowd thank you, we put this PR on our maintainer's plate. We will update you once we have a chance to dig deeper into it.

arrowd:
Any progress in this?

gndz07:
Hi @arrowd,

As mentioned previously, while we generally agree that keeping dependencies up to date is beneficial, the use case outlined in this PR does not currently provide sufficient justification for this change in its current form.

Additionally, we would need more time to properly evaluate and implement an upgrade for this dependency. In practice, this would likely involve upgrading Vite rather than bumping Rollup directly.

If you'd like to pursue this further, please consider opening a proposal issue to discuss the use case in more detail and gather community feedback. That would help us better understand the demand for this change and assess its priority.

gndz07:
Hi @arrowd,

We've upgraded the vite version in this [PR](https://github.com/traefik/traefik/pull/13794), where they moved on from rollup, so it is not a dependency anymore. Therefore, I'm closing this PR. Thank you for opening this PR!

PR #13114, opened 2026-05-06
fix: suppress unmatched ping entrypoint requests from access log

### LINKED ISSUE #12751: Ping Internal Appearing in Access Log
### Welcome!

### What did you do?

While implementing observability, I was going over Traefik Access Log and noticed and high amount of localhost call After a bit of digging I narrowed it down to the ping@internal (as they disapear when ping is disabled and they don't register inside the metrics). I use the /ping endpoint to verify the reverse proxy state (which is done from another machine on the same network) and it's implemented in a Dynamic configuration.

### What did you see instead?

A higher baseline (see first and second image) of logs looking like the following:

```json
{"ClientAddr":"127.0.0.1:50036","ClientHost":"127.0.0.1","ClientPort":"50036","ClientUsername":"-","DownstreamContentSize":19,"DownstreamStatus":404,"Duration":49121,"GzipRatio":0,"OriginContentSize":0,"OriginDuration":0,"OriginStatus":0,"Overhead":49121,"RequestAddr":":443","RequestContentSize":0,"RequestCount":158,"RequestHost":"","RequestMethod":"HEAD","RequestPath":"/ping","RequestPort":"443","RequestProtocol":"HTTP/1.1","RequestScheme":"http","RetryAttempts":0,"StartLocal":"2026-03-02T15:18:49.563546864Z","StartUTC":"2026-03-02T15:18:49.563546864Z","entryPointName":"websecure","level":"info","msg":"","time":"2026-03-02T15:18:49Z"}
```

The actual head scratcher being that this log line is alone and is missing the actual "caller" (the machine calling the /ping endpoint) whom seems to be correctly filtered. Following are two images, the first a log baseline with ping enabled and the second one without ping

Ping enabled Log Quantity

Ping disabled Log Quantity

Version:      3.6.9
Codename:     ramequin
Go version:   go1.25.7
Built:        2026-02-23T17:12:19Z
OS/Arch:      linux/amd64

Treafik.yml

```yaml
global:
  checkNewVersion: true
  sendAnonymousUsage: false

entryPoints:
  web:
    address: ":80"
    http:
      redirections:
        entryPoint:
          to: websecure
          scheme: https
      sanitizePath: true
      encodedCharacters:
        allowEncodedSlash: false
        allowEncodedBackSlash: false
        allowEncodedNullCharacter: false
        allowEncodedSemicolon: false
        allowEncodedPercent: false
        allowEncodedQuestionMark: false
        allowEncodedHash: false
  websecure:
    address: ":443"
    http:
      tls:
        certResolver: letsencrypt
        domains:
          - main: REDACTED
            sans:
              - REDACTED
      sanitizePath: true
      encodedCharacters:
        allowEncodedSlash: false
        allowEncodedBackSlash: false
        allowEncodedNullCharacter: false
        allowEncodedSemicolon: false
        allowEncodedPercent: false
        allowEncodedQuestionMark: false
        allowEncodedHash: false
    forwardedHeaders:
      trustedIPs:
        - "173.245.48.0/20"
        - "103.21.244.0/22"
        - "103.22.200.0/22"
        - "103.31.4.0/22"
        - "141.101.64.0/18"
        - "108.162.192.0/18"
        - "190.93.240.0/20"
        - "188.114.96.0/20"
        - "197.234.240.0/22"
        - "198.41.128.0/17"
        - "162.158.0.0/15"
        - "104.16.0.0/13"
        - "104.24.0.0/14"
        - "172.64.0.0/13"
        - "131.0.72.0/22"
        - 2400:cb00::/32
        - 2606:4700::/32
        - 2803:f800::/32
        - 2405:b500::/32
        - 2405:8100::/32
        - 2a06:98c0::/29
        - 2c0f:f248::/32
        - "127.0.0.1/32"
        - "10.0.0.0/8"
        - "192.168.0.0/16"
        - "172.16.0.0/12"

  ssh:
    address: ":22"

log:
  filePath: /var/log/traefik/traefik.log
  format: json
  level: INFO

accessLog:
  filePath: /var/log/traefik/access.log
  format: json
  filters:
    statusCodes:
      - "200-201"
      - "400-404"
      - "500-503"
  addInternals: false

metrics:
  otlp:
    grpc:
      insecure: true
  addInternals: false

providers:
  file:
    directory: /etc/traefik/conf.d
    watch: true

certificatesResolvers:
  letsencrypt:
    acme:
      email: REDACTED
      storage: /etc/traefik/acme.json
      profile: "shortlived"
      certificatesDuration: 160
      dnsChallenge:
        provider: cloudflare
        resolvers:
          - "1.1.1.1:53"
          - "8.8.8.8:53"
api: {}

ping:
  entryPoint: websecure
  manualRouting: true

```

Custom Ping Router

```yaml                               
http:
  routers:
    ping:
      rule: Host(`REDACTED`) && PathPrefix(`/ping`)
      service: ping@internal
      tls:
        certResolver: letsencrypt
        domains:
          - main: REDACTED
          - sans:
              - REDACTED
      entryPoints:
        - web
        - websecure
      middlewares:
        - trusted-ips
```

VM running Ubuntu 24.04, and Traefik running inside a SystemD Service

_No response_

satulurideepak4:
---
Title: fix: suppress ping entrypoint unmatched requests from access log

---

**What does this PR do?**

When `ping.entryPoint` is set to a TLS entry point with `manualRouting: true`, plain HTTP requests to that port match no router (the ping router requires TLS). The fallback 404 handler was built with `internal=false`, causing these unmatched requests to bypass `addInternals: false` and appear in the access log as noise.

This fix treats the default (unmatched) handler on the ping entry point as internal, so `addInternals: false` correctly suppresses it. Matched router handlers are unaffected - they continue to derive the `internal` flag from the service name suffix.

Fixes #12751

**Motivation**

Unmatched requests on a TLS ping entry point (e.g. plain HTTP hitting a port configured for TLS-only ping with `manualRouting: true`) were leaking into access logs even with `accessLog.addInternals: false`. This produced spurious log noise for operators using the ping health check on a dedicated TLS entry point.

**More**

- Added/updated tests
- Added/updated documentation

**Additional Notes**

Only the fallback `defaultHandler` on the ping entry point is affected. The `IsPingEntryPoint` helper checks `o.config.Ping.EntryPoint` and returns `false` if ping is unconfigured, so there is no behavioral change for setups that don't use `ping.entryPoint`.

---

### DIFF
--- a/pkg/server/middleware/observability.go
+++ b/pkg/server/middleware/observability.go
@@ -136,6 +136,17 @@ func (o *ObservabilityMgr) observabilityContextHandler(next http.Handler, intern
 	})
 }
 
+// IsPingEntryPoint reports whether the given entry point name is configured as the ping entry point.
+// This is used to treat unmatched requests on the ping entry point as internal,
+// so that addInternals=false suppresses them (e.g. plain HTTP hitting a TLS ping entry point).
+func (o *ObservabilityMgr) IsPingEntryPoint(entryPointName string) bool {
+	if o == nil || o.config.Ping == nil {
+		return false
+	}
+
+	return o.config.Ping.EntryPoint == entryPointName
+}
+
 // shouldAccessLog returns whether the access logs should be enabled for the given serviceName and the observability config.
 func (o *ObservabilityMgr) shouldAccessLog(internal bool, observabilityConfig dynamic.RouterObservabilityConfig) bool {
 	if o == nil {
--- a/pkg/server/middleware/observability_test.go
+++ b/pkg/server/middleware/observability_test.go
@@ -1,0 +1,69 @@
+package middleware
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/traefik/traefik/v3/pkg/config/static"
+	"github.com/traefik/traefik/v3/pkg/ping"
+)
+
+func TestIsPingEntryPoint(t *testing.T) {
+	testCases := []struct {
+		desc           string
+		mgr            *ObservabilityMgr
+		entryPointName string
+		expected       bool
+	}{
+		{
+			desc:           "nil ObservabilityMgr receiver",
+			mgr:            nil,
+			entryPointName: "websecure",
+			expected:       false,
+		},
+		{
+			desc:           "ping config is nil",
+			mgr:            &ObservabilityMgr{config: static.Configuration{}},
+			entryPointName: "websecure",
+			expected:       false,
+		},
+		{
+			desc: "entry point matches configured ping entry point",
+			mgr: &ObservabilityMgr{
+				config: static.Configuration{
+					Ping: &ping.Handler{EntryPoint: "websecure"},
+				},
+			},
+			entryPointName: "websecure",
+			expected:       true,
+		},
+		{
+			desc: "entry point does not match configured ping entry point",
+			mgr: &ObservabilityMgr{
+				config: static.Configuration{
+					Ping: &ping.Handler{EntryPoint: "websecure"},
+				},
+			},
+			entryPointName: "web",
+			expected:       false,
+		},
+		{
+			desc: "non-ping entry point on same traefik instance",
+			mgr: &ObservabilityMgr{
+				config: static.Configuration{
+					Ping: &ping.Handler{EntryPoint: "ping"},
+				},
+			},
+			entryPointName: "websecure",
+			expected:       false,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.desc, func(t *testing.T) {
+			t.Parallel()
+
+			assert.Equal(t, test.expected, test.mgr.IsPingEntryPoint(test.entryPointName))
+		})
+	}
+}
--- a/pkg/server/router/router.go
+++ b/pkg/server/router/router.go
@@ -208,7 +208,7 @@ func (m *Manager) BuildHandlers(rootCtx context.Context, entryPoints []string, t
 			epObsConfig = model.Observability
 		}
 
-		defaultHandler, err := m.observabilityMgr.BuildEPChain(ctx, entryPointName, false, epObsConfig).Then(http.NotFoundHandler())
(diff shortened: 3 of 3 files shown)


nmengin:
Thank you for this PR! This fix addresses the issue where ping entrypoint requests were appearing in access logs even with `addInternals: false` enabled. The approach of treating the default handler on the ping entrypoint as internal is sound and the scope is appropriately limited.

Please ensure that:
- Tests cover both matched and unmatched requests on the ping entrypoint
- Documentation is updated if needed to clarify this behavior
- The change has been validated with the scenario described in #12751 (TLS entrypoint with manualRouting: true)

We'll review the implementation details to ensure this doesn't affect other entrypoints or handlers.

satulurideepak4:
Hi @nmengin thanks for the quick review!

Just to clarify, I kept this scoped to ping only since that's exactly what #12751 is about. Prometheus would be a separate thing and can be tackled in its own PR if needed.

For the points you raised:

Tests: Added TestIsPingEntryPoint in pkg/server/middleware/observability_test.go. It covers the nil receiver case, nil ping config, matching entry point, non-matching entry point, and an unrelated entry point on the same instance.

Validation: Tested locally with the exact setup from the issue. ping.entryPoint set to websecure, manualRouting true, TLS-only ping router. HTTPS ping returns 200 and doesn't appear in the log. Plain HTTP to the TLS port gets a 404 but also doesn't appear in the log. Access log stays empty.

Other entrypoints: Only the default unmatched handler on the ping entry point is affected. Matched router handlers still work exactly as before, deriving internal from the service name suffix. Everything else is untouched.

Docs: I don't think a doc change is needed here. The addInternals field already says it suppresses internal services, this just makes the code actually do what it says.

Let me know if you want anything changed!

satulurideepak4:
@nmengin can you assign anyone to review this pr, as this is in review state for very long time?

rtribotte:
Hi @satulurideepak4,

We're focused on other priorities right now though, we'll keep you updated as soon as we can move forward!

PR #13118, opened 2026-05-07
Add v3.7 to supported version docs

jnoordsij:
### What does this PR do?

Add the v3.7.0 release to the docs in the listed supported releases.

It also marks v2.11 as no longer receiving official security support.
Although still maintained for internal purposes, I think this reads more clearly,
as an 'Ends on' with a past date looks slightly weird.
If another rephrasing of the current maintenance state is more appropriate,
feel free to amend accordingly, or drop the latter commit entirely.

### Motivation

Be up-to-date on supported version.

(Side note: is there some release checklist this action can be added to?)

### More

- [ ] Added/updated tests


rtribotte (approved):
LGTM

PR #13126, opened 2026-05-07
feat(metrics): export Go runtime metrics via OTLP

### LINKED ISSUE #13125: OTLP metrics exporter does not include Go runtime / process metrics
### Welcome!

### What did you do?

Configure Traefik with OTLP metrics enabled, no Prometheus endpoint:

```yaml
metrics:
  otlp:
    enabled: true
    http:
      enabled: true
      endpoint: http://otel-collector:4318/v1/metrics
```

Send metrics to an OTLP-compatible backend (Prometheus via remote-write, Thanos, Datadog, etc.) and look for Go runtime / process metrics.

### What did you see instead?

Only the Traefik application meter is exported (`traefik_entrypoint_*`, `traefik_service_*`, `traefik_config_*`, `traefik_open_connections_ratio`, `traefik_tls_certs_not_after_seconds`).

No Go runtime metrics are emitted via OTLP — no `process.runtime.go.goroutines`, `process.runtime.go.mem.heap_alloc`, `process.runtime.go.gc.pause_total_ns`, etc.

These metrics are still available on the `--metrics.prometheus` endpoint (because `prometheus/client_golang` registers a `go_collector` by default). Operators who only enable OTLP have no visibility into the runtime.

### Why this matters

Go runtime telemetry is essential for diagnosing memory pressure on Traefik, especially with `GOMEMLIMIT` set (which the official Helm chart sets to ~90 % of the container memory limit). Working-set memory naturally parks near `GOMEMLIMIT`, which trips generic memory alerts; the only way to distinguish "Go GC is doing its job" from "real heap growth" is to look at `heap_alloc` / `heap_inuse` / `next_gc` / `gc.pause` — exactly the metrics that aren't shipped via OTLP today.

Confirmed by reading [`pkg/observability/metrics/otel.go`](https://github.com/traefik/traefik/blob/master/pkg/observability/metrics/otel.go): only application meters are registered on the `MeterProvider`; the upstream `go.opentelemetry.io/contrib/instrumentation/runtime` package is not started.

3.6.13 (also reproducible on master).

Helm chart `traefik/traefik:39.0.x`, OTLP HTTP exporter to OpenTelemetry Collector, no Prometheus scrape config.

### Workarounds

- Enable `--metrics.prometheus=true` alongside OTLP and add a `ServiceMonitor` to scrape `/metrics` with a relabel that keeps only `go_*|process_*`. Adds a second metrics path; duplicates `traefik_*` if not relabeled.

### Proposal

Register `go.opentelemetry.io/contrib/instrumentation/runtime` against the OTLP `MeterProvider` once it's built in `newOpenTelemetryMeterProvider`. ~5 lines of change. Emits the standard `process.runtime.go.*` namespace via the existing OTLP pipeline. No new flags; if a user wants to opt-out we can gate it on a future config field.

Happy to send a PR.

ElfoLiNk:
### What does this PR do?

Registers the upstream Go runtime instrumentation
(`go.opentelemetry.io/contrib/instrumentation/runtime`) against the OTLP
`MeterProvider` once it is created in `newOpenTelemetryMeterProvider`.

After this change, `process.runtime.go.*` metrics — goroutines, heap
allocations, GC pause, CGO calls — are emitted via the same OTLP
exporter that already ships the Traefik application meter.

### Motivation

When OTLP metrics are enabled and the Prometheus endpoint is left
disabled (a common deployment shape: send everything to an OTel
Collector), there is currently no way to observe the Go runtime
through the configured OTLP pipeline. The only available path today
is to also enable `--metrics.prometheus` and have the operator scrape
`/metrics`, which:

- splits the metrics path into two backends,
- duplicates the `traefik_*` series unless an explicit relabel is
  configured, and
- forces operators to maintain a `ServiceMonitor` (or equivalent)
  alongside an otherwise pure-OTLP setup.

Runtime visibility matters for memory triage in particular. The
official Helm chart sets `GOMEMLIMIT` to ~90 % of the container memory
limit, which causes the working-set to park near the limit by design
— a behavior that trips generic `working_set / cgroup_limit`
container alerts. The only way to discriminate "Go is GC'ing
correctly" from "real heap growth" is to look at `heap_alloc`,
`heap_inuse`, `next_gc`, and `gc.pause` over time, and those metrics
need a transport.

### More

- Fixes #13125
- Bug fix: no
- Enhancement: yes
- Issue: #13125
- Documentation:

### Additional Notes

- Failure to start the runtime instrumentation is logged at error
  level but does not prevent Traefik from starting (consistent with
  the rest of the OTLP setup, which logs and continues on partial
  failure).
- New direct dependency:
  `go.opentelemetry.io/contrib/instrumentation/runtime v0.68.0`
  (compatible with the otel core already pinned in `go.mod`).
- No new flags, no breaking changes. Operators who do not enable
  OTLP metrics see no behavioural change. Operators who do will see
  additional `process.runtime.go.*` series in their backend.

### Tests

- `go test ./pkg/observability/metrics/...` passes locally on master.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -124,6 +124,8 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+require go.opentelemetry.io/contrib/instrumentation/runtime v0.68.0
+
 require (
 	cloud.google.com/go/auth v0.20.0 // indirect
 	cloud.google.com/go/auth/oauth2adapt v0.2.8 // indirect
--- a/go.sum
+++ b/go.sum
@@ -2112,6 +2112,8 @@ go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.6
 go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.67.0/go.mod h1:NoUCKYWK+3ecatC4HjkRktREheMeEtrXoQxrqYFeHSc=
 go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.67.0 h1:OyrsyzuttWTSur2qN/Lm0m2a8yqyIjUVBZcxFPuXq2o=
 go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.67.0/go.mod h1:C2NGBr+kAB4bk3xtMXfZ94gqFDtg/GkI7e9zqGh5Beg=
+go.opentelemetry.io/contrib/instrumentation/runtime v0.68.0 h1:jhVIQEprwUTV+KfzzliLidclhoTOoHTgdz96kAyR8mU=
+go.opentelemetry.io/contrib/instrumentation/runtime v0.68.0/go.mod h1:4HsdbLUbernaTnA8CNaNE+1g026SciXb3juRYe3l8EY=
 go.opentelemetry.io/contrib/propagators/autoprop v0.63.0 h1:S3+4UwR3Y1tUKklruMwOacAFInNvtuOexz4ZTmJNAyw=
 go.opentelemetry.io/contrib/propagators/autoprop v0.63.0/go.mod h1:qpIuOggbbw2T9nKRaO1je/oTRKd4zslAcJonN8LYbTg=
 go.opentelemetry.io/contrib/propagators/aws v1.38.0 h1:eRZ7asSbLc5dH7+TBzL6hFKb1dabz0IV51uUUwYRZts=
--- a/pkg/observability/metrics/otel.go
+++ b/pkg/observability/metrics/otel.go
@@ -15,6 +15,7 @@ import (
 	otypes "github.com/traefik/traefik/v3/pkg/observability/types"
 	"github.com/traefik/traefik/v3/pkg/types"
 	"github.com/traefik/traefik/v3/pkg/version"
+	"go.opentelemetry.io/contrib/instrumentation/runtime"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc"
@@ -252,6 +253,14 @@ func newOpenTelemetryMeterProvider(ctx context.Context, config *otypes.OTLP) (*s
 	)
 
 	otel.SetMeterProvider(meterProvider)
+
+	// Register Go runtime instrumentation so that process.runtime.go.*
+	// metrics (goroutines, heap, GC, etc.) are exported via OTLP alongside
+	// the Traefik application meter. Failure here is non-fatal: a missing
+	// runtime gauge should not prevent Traefik from starting.
+	if err := runtime.Start(runtime.WithMeterProvider(meterProvider)); err != nil {
+		log.Ctx(ctx).Err(err).Msg("Unable to start Go runtime instrumentation for OTLP metrics")
+	}
 
 	return meterProvider, nil
 }


nmengin:
Thank you for this contribution! This PR addresses a real operational need by enabling Go runtime metrics export via OTLP, eliminating the requirement for parallel Prometheus endpoints in pure-OTLP deployments.

**Summary of the change:**
- Registers upstream Go runtime instrumentation against the OTLP MeterProvider
- Exports `process.runtime.go.*` metrics (goroutines, heap allocations, GC pause, CGO calls) via OTLP
- Non-breaking change; no new configuration flags required
- Adds direct dependency on `go.opentelemetry.io/contrib/instrumentation/runtime v0.68.0`

**Observations:**
- The implementation follows Traefik's existing pattern of logging errors without blocking startup
- The motivation is well-documented, particularly regarding memory triage visibility with `GOMEMLIMIT`
- Tests have been run locally (`go test ./pkg/observability/metrics/...`)

This is ready for review. Please ensure the new dependency is compatible with Traefik's dependency management strategy before merging.

ElfoLiNk:
Rebased onto current `master` (the branch was 251 commits behind and conflicting on `go.mod`/`go.sum`) and bumped the dependency. `newOpenTelemetryMeterProvider` has moved since May, so the registration now sits after `otel.SetMeterProvider` at its current location in `pkg/observability/metrics/otel.go`.

On the dependency question from the review:

- **It is not a new dependency in the module graph.** `go.opentelemetry.io/contrib/instrumentation/runtime` is already there as an indirect requirement — `go get` reported `upgraded v0.64.0 => v0.69.0`. This PR promotes it to a direct requirement, it does not add a new one.
- **v0.69.0 is the version that matches `master`.** The contrib instrumentation modules pin the core release exactly:

  | version | requires `go.opentelemetry.io/otel` |
  |---|---|
  | v0.68.0 (previously pinned here) | v1.43.0 — below master |
  | **v0.69.0** | **v1.44.0 — same as master** |
  | v0.70.0 | v1.45.0 — would force a core bump |

  So v0.69.0 keeps the core untouched: `git diff go.mod go.sum` shows no change to any `go.opentelemetry.io/otel*` version. v0.70.0 would drag the whole project to otel v1.45.0, which seemed like the opposite of what the dependency-management concern was about.
- The diff is now `+15/-1` across `go.mod`, `go.sum` and `otel.go`.

Verified with `go build ./pkg/observability/...` and `go test ./pkg/observability/metrics/...` (ok, 27s), `gofmt` clean.

### Why this got more useful since May

The missing runtime metrics turned out to hide a real leak. In #13584 a Traefik pod reached 7087 goroutines and 1.49 GB of live heap after 50 days of uptime, 6411 of those goroutines being leaked `go-cache` janitors from the TLS store rebuild. Diagnosing it required `SIGQUIT` on a production pod, because the OTLP pipeline carried no `process.runtime.go.goroutines` and `--api.debug` was not enabled.

Concretely, on nine production clusters shipping Traefik metrics over OTLP, the leak is invisible: `traefik_open_connections_ratio` and `traefik_config_reloads_total` arrive, `go_goroutines` does not, and the only workaround is enabling `--metrics.prometheus` alongside OTLP purely to scrape the Go collector — which is exactly the duplication this PR removes.

ElfoLiNk:
Rebased again onto current `master`: the branch had drifted 125 commits behind and `go.mod`/`go.sum` went back into conflict. The code change itself is unchanged (9 lines in `pkg/observability/metrics/otel.go`); only the dependency lines were regenerated.

Dependency status is the same as before, re-verified against today's `master`:

- Not a new dependency in the module graph — `go get` reports `upgraded go.opentelemetry.io/contrib/instrumentation/runtime v0.64.0 => v0.69.0`. This promotes an existing indirect requirement to a direct one.
- v0.69.0 still matches `master`'s core: it requires `go.opentelemetry.io/otel v1.44.0`, exactly what `master` pins. `git diff go.mod go.sum` shows no change to any `go.opentelemetry.io/otel*` version. v0.70.0 and v0.71.0 would pull the core to v1.45.0+.

Verified with `go build ./pkg/observability/...` and `go test ./pkg/observability/metrics/...` (66 tests, ok), `gofmt` clean. Now 1 ahead / 0 behind `master`.

PR #13127, opened 2026-05-07
Display dashboardName on the UI 

t0mtaylor:
### What does this PR do?

Add the dashboard name badge to the dashboard sidebar to assist with identifying multiple Traefik instances (see below example).

| External Traefik Badge | Internal Traefik Badge |
| ------------- | ------------- |
| image  |   image |

  ### Motivation

To give a clearer differentiator of different instances, on top of the dashboard name as the title page metadata.

  ### More

  - [x] Added/updated tests

  ### Additional Notes

 ~- Backend: new `DashboardName` and `DashboardNamePosition` on `static_config.API`, threaded into `pkg/version` so `/api/version` exposes them. 5 unit tests including concurrent-read coverage of the package-level vars.~
  - Frontend: `contexts/version.tsx` adds `dashboardNamePosition` with a `normalizePosition()` helper that coerces unknown values to `'side'`.`SideNavBar.tsx` renders the badge inside the logo `Flex` when position is `side`, in a separate `Flex` (with `data-testid="instance-badge-below"`) when position is `below`.
  - Default value is empty → badge is hidden, so no visual change for users who don't configure it.
  - 3 new spec files cover both render branches; existing specs (`InstanceBadge.spec.tsx`, `PageTitle.spec.tsx`, `use-hub-upgrade-button.spec.tsx`) updated for the new context field.
  - Soaked 24h+ on both ext and int tier dashboards; no regressions.
  - No matching open issue — submitting as a clean enhancement.

  Both checkboxes (tests + docs) are now ticked since the docs commit landed on each branch.


gndz07:
Hey @t0mtaylor,

Thank you for opening the PR.

Just to clarify, this PR only adds support for displaying the previously available api.dashboardName in the UI; it does not introduce the option itself. If you don’t mind, I’ll update the PR description to make this clearer and avoid confusion.

Also, after discussing this with the other maintainers, we feel that adding a position configuration is not necessary at the moment. We would prefer to keep it simple and display the dashboard name below the logo by default.

Could you please update the PR accordingly?

t0mtaylor:
Hey @gndz07

Thank you for reviewing the PR

Sure i can update based on your comments, are you happy with the below image of the badge and position below the logo?

image

gndz07:
@t0mtaylor yes, below the logo is fine.

t0mtaylor:
> @t0mtaylor yes, below the logo is fine.

@gndz07 Any update on target release for this change before i rebase again? Thanks

webui/src/components/InstanceBadge.tsx
```diff
@@ -0,0 +1,51 @@
+import { Box, Text } from '@traefik-labs/faency'
```
gndz07: Is there a particular reason why you didn't choose to use the Badge component from Faency instead of making a custom one here? As we already use Badge in several other places, I think it'd be best to use the same component so that they style will always be aligned.

pkg/version/version_test.go
```diff
@@ -0,0 +1,90 @@
+package version
```
gndz07: I think it's better to limit the scope of this PR to UI changes.

gndz07 (changes requested):
Hi @t0mtaylor, sorry for the delay. I left some review, thank you again for your contribution!

t0mtaylor:
> Hi @t0mtaylor, sorry for the delay. I left some review, thank you again for your contribution!

No problem @gndz07 thank you for reviewing, I shall make those changes as requested and let you know when this MR is ready again for review 🚀

PR #13135, opened 2026-05-09
feat: add uid options pass tls client cert

### LINKED ISSUE #13104: Add UID (Unique Identifier) to passtlscert middleware
### Welcome!

Similar request: #7951
With the passtlscert middleware I cannot get the UID from a certificate passed in the info section. My workaround is to pass the pem cert but as noted in the traefik docs this can cause [issues with large header](https://doc.traefik.io/traefik/middlewares/passtlsclientcert/) sizes. It also forces a second PEM decode later, when all I really need is the UID.

My dirty fix:
```yaml
experimental:
  localPlugins:
    cert-subject-plugin:
      moduleName: github.com/SwiftCTRL/traefik-plugin-cert-subject
      mountPath: /plugins-local/src/github.com/SwiftCTRL/traefik-plugin-cert-subject
      type: inlinePlugin
      source:
        go.mod: |
          module github.com/SwiftCTRL/traefik-plugin-cert-subject

          go 1.24
        .traefik.yml: |
          displayName: Certificate Subject DN
          type: middleware

          import: github.com/SwiftCTRL/traefik-plugin-cert-subject

          summary: |
            Reads X-Forwarded-Tls-Client-Cert (URL-encoded PEM), extracts DC, O, OU, CN,
            and UID from the certificate subject in order, and sets Ssl-Client-Subject-Dn.

          testData: {}
        main.go: |
          package traefik_plugin_cert_subject

          import (
            "context"
            "crypto/x509"
            "encoding/asn1"
            "encoding/base64"
            "encoding/pem"
            "fmt"
            "log"
            "net/http"
            "net/url"
            "strings"
          )

          type Config struct{}

          func CreateConfig() *Config {
            return &Config{}
          }

          type CertSubjectPlugin struct {
            next http.Handler
          }

          func New(ctx context.Context, next http.Handler, config *Config, name string) (http.Handler, error) {
            return &CertSubjectPlugin{next: next}, nil
          }

          func (p *CertSubjectPlugin) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
            log.Printf("[cert-subject-plugin] handling request: method=%s host=%s path=%s", req.Method, req.Host, req.URL.Path)
            certHeader := req.Header.Get("X-Forwarded-Tls-Client-Cert")
            if certHeader == "" {
              dbgMsg := "no X-Forwarded-Tls-Client-Cert header"
              log.Printf("[cert-subject-plugin] %s", dbgMsg)
              req.Header.Set("X-Cert-Subject-Debug", dbgMsg)
              p.next.ServeHTTP(rw, req)
              return
            }

          bf, _, found := strings.Cut(certHeader, ",")

          var cert string

          if !found {
            cert = certHeader
          } else {
            cert = bf
          }

            dn, dbg, err := extractSubjectDN(cert)
            if err != nil {
              dbgMsg := fmt.Sprintf("error: %v", err)
              log.Printf("[cert-subject-plugin] %s", dbgMsg)
              req.Header.Set("X-Cert-Subject-Debug", dbgMsg)
              p.next.ServeHTTP(rw, req)
              return
            }

            if dn == "" {
              dbgMsg := fmt.Sprintf("no DC/O/OU/CN/UID found; %s", dbg)
              log.Printf("[cert-subject-plugin] %s", dbgMsg)
              req.Header.Set("X-Cert-Subject-Debug", dbgMsg)
              p.next.ServeHTTP(rw, req)
              return
            }

            log.Printf("[cert-subject-plugin] %s", dbg)
            req.Header.Set("X-Cert-Subject-Debug", dbg)
            req.Header.Set("Ssl-Client-Subject-Dn", dn)
            p.next.ServeHTTP(rw, req)
          }

          func matchOID(oid asn1.ObjectIdentifier, parts ...int) bool {
            if len(oid) != len(parts) {
              return false
            }
            for i, v := range parts {
              if oid[i] != v {
                return false
              }
            }
            return true
          }

          // extractSubjectDN decodes the cert from X-Forwarded-Tls-Client-Cert,
          // parses it, and returns a comma-separated DN of DC, O, OU, CN, UID in subject order,
          // plus a debug summary string.
          // The header may arrive as URL-encoded PEM (Traefik default) or as raw base64 DER
          // (some proxy chains strip PEM headers). Format is detected by prefix.
          func extractSubjectDN(encoded string) (string, string, error) {
            var der []byte

            if strings.HasPrefix(encoded, "-----") || strings.HasPrefix(encoded, "%2D") {
              // URL-encoded or plain PEM
              unescaped, err := url.QueryUnescape(encoded)
              if err != nil {
                return "", "", fmt.Errorf("url decode: %w", err)
              }
              block, _ := pem.Decode([]byte(unescaped))
              if block == nil {
                return "", "", fmt.Errorf("pem decode: no PEM block found")
              }
              der = block.Bytes
            } else {
              // Raw base64-encoded DER (strip padding before decoding)
              var err error
              der, err = base64.StdEncoding.DecodeString(encoded)
              if err != nil {
                return "", "", fmt.Errorf("base64 decode: %w", err)
              }
            }

            cert, err := x509.ParseCertificate(der)
            if err != nil {
              return "", "", fmt.Errorf("x509 parse: %w", err)
            }

            var parts, msgs []string
            msgs = append(msgs, fmt.Sprintf("%d subject attribute(s)", len(cert.Subject.Names)))
            for _, attr := range cert.Subject.Names {
              switch {
              case matchOID(attr.Type, 0, 9, 2342, 19200300, 100, 1, 25): // DC
                parts = append(parts, fmt.Sprintf("DC=%v", attr.Value))
              case matchOID(attr.Type, 2, 5, 4, 10): // O
                parts = append(parts, fmt.Sprintf("O=%v", attr.Value))
              case matchOID(attr.Type, 2, 5, 4, 11): // OU
                parts = append(parts, fmt.Sprintf("OU=%v", attr.Value))
              case matchOID(attr.Type, 2, 5, 4, 3): // CN
                parts = append(parts, fmt.Sprintf("CN=%v", attr.Value))
              case matchOID(attr.Type, 0, 9, 2342, 19200300, 100, 1, 1): // UID
                parts = append(parts, fmt.Sprintf("UID=%v", attr.Value))
              default:
                msgs = append(msgs, fmt.Sprintf("skipped OID %v=%v", attr.Type, attr.Value))
              }
            }

            return strings.Join(parts, ","), strings.Join(msgs, "; "), nil
          }
```

n0ta26:
### What does this PR do?

This PR adds support for forwarding the UID attribute from the client certificate Subject DN in the `passTLSClientCert` middleware.

The UID attribute is mapped from OID `0.9.2342.19200300.100.1.1` and can be enabled with the following configuration:
```yaml
http:
  middlewares:
    pass-client-cert:
      passTLSClientCert:
        info:
          subject:
            uid: true
```

When enabled, the UID attribute is included in the X-Forwarded-Tls-Client-Cert-Info header.
Example decoded header value:
```text
Subject="UID=user-001,C=JP,O=Client Org,OU=Engineering,CN=test-client"
```

### Motivation

Currently, users who need the UID attribute from the client certificate Subject DN have to enable pem: true and parse the full client certificate downstream.

This has a few drawbacks:

* The full PEM certificate increases the forwarded header size.
* Downstream services need to parse the certificate again.
* Users cannot selectively forward only the UID attribute through X-Forwarded-Tls-Client-Cert-Info.

This change allows users to forward only the selected UID attribute directly from Traefik, similar to other supported Subject DN attributes.

Closes #13104

### More

- [ ] Added/updated tests

### Additional Notes

Unit tests were added to verify that the UID attribute is included only when info.subject.uid is enabled.
Manual verification was also performed with an mTLS client certificate containing UID=user-001.
Forwarded header:
```text
X-Forwarded-Tls-Client-Cert-Info: Subject%3D%22UID%3Duser-001%2CC%3DJP%2CO%3DClient+Org%2COU%3DEngineering%2CCN%3Dtest-client%22
```

Decoded value:
```text
Subject="UID=user-001,C=JP,O=Client Org,OU=Engineering,CN=test-client"
```

Tests run:
```text
go test -v ./pkg/middlewares/passtlsclientcert
make test-unit
make binary
```

### DIFF
--- a/docs/content/reference/dynamic-configuration/docker-labels.yml
+++ b/docs/content/reference/dynamic-configuration/docker-labels.yml
@@ -126,6 +126,7 @@
 - "traefik.http.middlewares.middleware16.passtlsclientcert.info.subject.organizationalunit=true"
 - "traefik.http.middlewares.middleware16.passtlsclientcert.info.subject.province=true"
 - "traefik.http.middlewares.middleware16.passtlsclientcert.info.subject.serialnumber=true"
+- "traefik.http.middlewares.middleware16.passtlsclientcert.info.subject.uid=true"
 - "traefik.http.middlewares.middleware16.passtlsclientcert.pem=true"
 - "traefik.http.middlewares.middleware17.plugin.pluginconf0.name0=foobar"
 - "traefik.http.middlewares.middleware17.plugin.pluginconf0.name1=foobar"
--- a/docs/content/reference/dynamic-configuration/file.toml
+++ b/docs/content/reference/dynamic-configuration/file.toml
@@ -303,6 +303,7 @@
             commonName = true
             serialNumber = true
             domainComponent = true
+            uid = true
           [http.middlewares.Middleware16.passTLSClientCert.info.issuer]
             country = true
             province = true
--- a/docs/content/reference/dynamic-configuration/file.yaml
+++ b/docs/content/reference/dynamic-configuration/file.yaml
@@ -349,6 +349,7 @@ http:
             commonName: true
             serialNumber: true
             domainComponent: true
+            uid: true
           issuer:
             country: true
             province: true
--- a/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
+++ b/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
@@ -1950,6 +1950,10 @@ spec:
                             description: SerialNumber defines whether to add the serialNumber
                               information into the subject.
                             type: boolean
+                          uid:
+                            description: UID defines whether to add the UID information
+                              into the subject.
+                            type: boolean
                         type: object
                     type: object
                   pem:
--- a/docs/content/reference/dynamic-configuration/traefik.io_middlewares.yaml
+++ b/docs/content/reference/dynamic-configuration/traefik.io_middlewares.yaml
@@ -1083,6 +1083,10 @@ spec:
                             description: SerialNumber defines whether to add the serialNumber
                               information into the subject.
                             type: boolean
+                          uid:
+                            description: UID defines whether to add the UID information
+                              into the subject.
+                            type: boolean
                         type: object
                     type: object
                   pem:
--- a/docs/content/reference/routing-configuration/http/middlewares/passtlsclientcert.md
+++ b/docs/content/reference/routing-configuration/http/middlewares/passtlsclientcert.md
@@ -71,6 +71,7 @@ spec:
                 commonName: true
                 serialNumber: true
                 domainComponent: true
+                uid: true
               issuer:
                 country: true
                 province: true
@@ -98,6 +99,7 @@ spec:
             commonName = true
             serialNumber = true
             domainComponent = true
+            uid = true
           [http.middlewares.test-passtlsclientcert.passTLSClientCert.info.issuer]
             country = true
             province = true
@@ -123,6 +125,7 @@ spec:
       - "traefik.http.middlewares.test-passtlsclientcert.passtlsclientcert.info.subject.organizationalunit=true"
       - "traefik.http.middlewares.test-passtlsclientcert.passtlsclientcert.info.subject.province=true"
       - "traefik.http.middlewares.test-passtlsclientcert.passtlsclientcert.info.subject.serialnumber=true"
+      - "traefik.http.middlewares.test-passtlsclientcert.passtlsclientcert.info.subject.uid=true"
       - "traefik.http.middlewares.test-passtlsclientcert.passtlsclientcert.info.issuer.commonname=true"
       - "traefik.http.middlewares.test-passtlsclientcert.passtlsclientcert.info.issuer.country=true"
       - "traefik.http.middlewares.test-passtlsclientcert.passtlsclientcert.info.issuer.domaincomponent=true"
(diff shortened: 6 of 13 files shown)


nmengin:
Thank you for this PR! This is a well-crafted implementation that directly addresses the feature request in #13104. The addition of UID support to the `passTLSClientCert` middleware is valuable for users who need to extract specific certificate attributes without forwarding the entire PEM certificate.

A few observations:
- The implementation is focused and minimal, following the existing pattern for other Subject DN attributes
- Unit tests have been added to verify the functionality
- Manual testing with mTLS certificates has been performed
- The solution addresses the original pain points (header size, downstream parsing, selective forwarding)

Please ensure all tests pass (`make test-unit`) and consider adding/updating documentation if not already done. The maintainers will review the code for any adjustments needed before merging.

n0ta26:
@nmengin Thank you for the prompt review and guidance.

I have completed the requested follow-ups:
- Ran `make test-unit`
- Updated the documentation

Please let me know if anything else needs to be adjusted.

n0ta26:
Hi maintainers, just checking in on this PR.

Please let me know if any additional changes, tests, or documentation updates are needed from my side. Thank you!

rtribotte:
Hey @n0ta26,

PR #13141, opened 2026-05-11
Make resolveReference method as a function

kevinpollet:
### What does this PR do?

This pull request makes the resolveReference method a function to be aligned with the code in v3.7, which introduces the middlewares at the service level.

### More

- [ ] Added/updated tests


juliens (approved):
LGTM

PR #13153, opened 2026-05-11
Node.js v26 preparations

jnoordsij:
### What does this PR do?

The new Node.js v26 major has been recently released and comes with some backward incompatible changes.
Most notably for this project is the removal of the corepack integration from Node.js itself (see https://github.com/nodejs/TSC/pull/1697) and the removal of Yarn from the Node.js official Docker image (see https://github.com/nodejs/docker-node/pull/2422).

This PR aims to prepare for these changes in a backwards compatible manner, so that we can remain on the current active LTS major (v24) while already being forward compatible (see also #13129).

In the effort, we also bump to the latest available Yarn and Node.js v24 versions.
Note for the update of Yarn, the migrated old default value of `npmMinimalAgeGate: 0` was not kept in `.yarnrc.yml` automigration, as the new default should be prefered (more secure).

### Motivation

Be up-to-date.

### More

- [ ] Added/updated tests

### Additional Notes

See #13129 for the follow-up.


jnoordsij:
@gndz07 with v26 becoming LTS sometime soon, it might be nice to look into this sometime soon. I do admit I've thrown quite a few distinct things in here all at once, so if you prefer for me to split this up further and/or drop parts even, LMK!

.github/workflows/template-webui.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
           persist-credentials: false
 
       - name: Enable corepack
-        run: corepack enable
+        run: npm i -g corepack
```
gndz07, line 19: I have a doubt here whether we still have to enable the corepack after installing it? Seems to me that this command just installs the missing package but we still need to enable the shims.
```suggestion
        run: npm i -g corepack && corepack enable
```

jnoordsij: This should be sufficient (per installation instructions and succeeding CI). It works as the package itself already globally declares it bins and shims: https://github.com/nodejs/corepack/blob/main/package.json#L63-L92.

gndz07: @jnoordsij got it, thank you for the confirmation!

gndz07 (approved):
LGTM, thanks!

jnoordsij:
Rebased to solve conflicts and further bumped Yarn and Node.js to their latest minor versions.

Issue #13090, opened 2026-04-30
ingress-nginx provider CORs behavior not consistent with Nginx


mliang2:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

The ingress-nginx Traefik provider's CORs behavior is not matching Nginx's behavior.
With `nginx.ingress.kubernetes.io/enable-cors: "true"` annotation:

* Nginx response
```
$ curl -sI --resolve httpbin.local:80:$nginx_ip http://httpbin.local|sort
Access-Control-Allow-Credentials: true
Access-Control-Allow-Headers: DNT,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range,Authorization
Access-Control-Allow-Methods: GET, PUT, POST, DELETE, PATCH, OPTIONS
Access-Control-Allow-Origin: *
Access-Control-Max-Age: 1728000
Content-Type: text/html; charset=utf-8
```

* Traefik's response
```
$ curl -sI --resolve httpbin.local:80:$traefik_ip http://httpbin.local|sort
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: *
Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' camo.githubusercontent.com
Content-Type: text/html; charset=utf-8
```

The  Traefik response is missing the `Access-Control-Allow-Headers`, `Access-Control-Allow-Methods`, and `Access-Control-Max-Age` headers.

If I set any custom values for `nginx.ingress.kubernetes.io/cors-allow-headers` or `nginx.ingress.kubernetes.io/cors-allow-methods`, they show up in Nginx, but not in Traefik

The [Traefik docs](https://doc.traefik.io/traefik/reference/routing-configuration/kubernetes/ingress-nginx/#limitations) states:
> CORS: NGINX responds with all configured headers unconditionally; Traefik handles headers differently between pre-flight and regular requests.

Is the behavior I've described expected?

### What did you see instead?

CORs headers are not set in Traefik response

### What version of Traefik are you using?

Traefik v3.7.0-rc3
Helm chart: 40.0.0-rc.2

### What is your environment & configuration?

Using this k8s configuration:

```yml
---
apiVersion: apps/v1
[...]

PR #13161, opened 2026-05-12
fix(middleware): add CORS headers to regular responses

aliamerj:
### What does this PR do?

Adds configured CORS headers to regular non-preflight responses in the headers middleware.

Currently:

1.  preflight `OPTIONS` responses include:  `Access-Control-Allow-Headers`, `Access-Control-Allow-Methods` and `Access-Control-Max-Age`

2. regular `GET/POST` responses only include: `Access-Control-Allow-Origin` and  `Access-Control-Allow-Credentials`

This change makes regular responses include the missing configured CORS headers as well, matching nginx-ingress behavior more closely.

### Motivation

Related to #13090.

I investigated the issue and found the behavior comes from the core headers middleware rather than the ingress-nginx provider translation itself.

This aligns Traefik behavior more closely with nginx-ingress, which returns the configured CORS headers on regular responses as well.

### More

- [x] Added/updated tests

### Additional Notes

The existing preflight behavior is unchanged. This only adds the configured CORS headers to regular non-preflight responses.

This PR is opened mainly for discussion and feedback first, as I am not fully sure whether this behavior change is desired globally or if there is a better approach for nginx-ingress compatibility.

### DIFF
--- a/pkg/middlewares/headers/header.go
+++ b/pkg/middlewares/headers/header.go
@@ -95,6 +95,20 @@ func (s *Header) PostRequestModifyResponseHeaders(res *http.Response) error {
 		res.Header.Set("Access-Control-Expose-Headers", exposeHeaders)
 	}
 
+	allowHeaders := strings.Join(s.headers.AccessControlAllowHeaders, ",")
+	if allowHeaders != "" {
+		res.Header.Set("Access-Control-Allow-Headers", allowHeaders)
+	}
+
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hello @aliamerj and thanks for your contribution,

According to https://github.com/traefik/traefik/issues/13090#issuecomment-4590617423, I am closing this pull request.

PR #13163, opened 2026-05-12
Add ingress-nginx upstream keepalive timeout annotation

koresha:
## Summary

- parse `nginx.ingress.kubernetes.io/upstream-keepalive-timeout` in the ingress-nginx provider
- let the per-Ingress annotation override the existing provider-level `upstreamKeepaliveTimeout` default
- add a fixture, regression test, and documentation entry for the timeout annotation matrix

## Motivation

`#12708` added the provider-level `upstreamKeepaliveTimeout` option, but the ingress-nginx provider still ignored the per-Ingress `nginx.ingress.kubernetes.io/upstream-keepalive-timeout` annotation.

This change completes that migration path so users moving from ingress-nginx to Traefik can preserve backend idle keep-alive timeout behavior without changing their global provider configuration.

Related to #12631.


nmengin:
Hey @koresha,

Thank you for this PR.
There was a misunderstanding in issue #12631: this option exists only in the ingress-nginx [ConfigMap](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/configmap.md#upstream-keepalive-timeout), but it's not a valid [annotation](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md).

Sorry for that, but I have to close this PR.

Thanks again for making Traefik a better project.

PR #13183, opened 2026-05-16
feat: fallback to RemoteAddr in PoolStrategy when X-Forwarded-For is absent

### LINKED ISSUE #7884: no strategy fallback in case of X-Forwarded-For's absence

firasmosbehi:
### What does this PR do?

When the `X-Forwarded-For` header is not present, `PoolStrategy` now falls back to checking the request's `RemoteAddr` against the configured IP pool.

### Motivation

This improves the `IPAllowList` middleware usability for traffic paths that do not set the `X-Forwarded-For` header (e.g., internal traffic, VPNs). Previously, the middleware would return an empty client IP when `X-Forwarded-For` was absent, causing allow-listing to fail for legitimate direct connections.

- `PoolStrategy.GetIP` falls back to `RemoteAddr` when XFF is absent
- `RemoteAddr` IP is checked against the `Checker` pool before returning
- Existing XFF behavior is preserved: fallback only happens when the XFF header is completely absent, not when all XFF IPs are in the pool

Fixes #7884

### More

- [x] Added/updated tests

### Additional Notes

This is a non-breaking change. The fallback only activates when `X-Forwarded-For` is absent. When the header is present (even if all IPs are in the excluded pool), the existing behavior is preserved.

### DIFF
--- a/docs/content/reference/routing-configuration/http/middlewares/ipallowlist.md
+++ b/docs/content/reference/routing-configuration/http/middlewares/ipallowlist.md
@@ -5,7 +5,9 @@ description: "Learn how to use IPAllowList in HTTP middleware for limiting clien
 
 `ipAllowList` accepts / refuses requests based on the client IP.
 
-## Configuration Example
+## Configuration Examples
+
+### Allowing Defined IPs
 
 ```yaml tab="Structured (YAML)"
 # Accepts request from defined IP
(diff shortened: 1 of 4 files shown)


nmengin:
Hey @firasmosbehi,

Could you please rebase on the master branch instead of your previous commits?

PR #13184, opened 2026-05-16
feat: expose DisableCompression on ServersTransport

### LINKED ISSUE #13075: Expose DisableCompression on ServersTransport to allow upstream gzip passthrough

firasmosbehi:
### What does this PR do?

Exposes a `disableCompression` option on `ServersTransport` that maps directly to `http.Transport.DisableCompression`.

### Motivation

Go's `http.Transport` defaults to automatically adding `Accept-Encoding: gzip` to requests and transparently decompressing responses. For reverse proxies, this is problematic when backends always return gzipped content regardless of the request's `Accept-Encoding` header (e.g., Spring Boot with `server.compression.enabled=true`). In those cases, clients that did not request compression receive decompressed responses, creating an asymmetry.

Fixes #13075

### More

- [x] Added/updated tests

### Additional Notes

When `disableCompression: true` is set on a `ServersTransport`, Traefik will:
- Not automatically add `Accept-Encoding: gzip` to upstream requests
- Not transparently decompress responses from the backend


nmengin:
Hey @firasmosbehi,

Could you please rebase on the master branch instead of your previous commits?

PR #13188, opened 2026-05-17
fix: handle lowercase percent-encoded reserved characters in muxer routing path

### LINKED ISSUE #13173: Muxer routing path: reservedCharacters map is uppercase-hex only, percent-encoding case asymmetry on reserved characters
## Summary

The `reservedCharacters` map in `pkg/muxer/http/mux.go:102-122` contains only uppercase hex entries (`%2A`, `%2F`, `%3A`, `%3B`, `%3D`, `%3F`, `%5B`, `%5D`, etc.). Per RFC 3986 section 2.1, `%2a` and `%2A` are equivalent, but Go's `url.URL.EscapedPath()` preserves the original case. Result : lowercase percent-encoded reserved characters fall through `url.PathUnescape` and are decoded into the routing path, while uppercase variants stay encoded. Two RFC-equivalent paths can route to different routers.

```go
var reservedCharacters = map[string]rune{
    "%3A": ':',
    "%2F": '/',
    "%3F": '?',
    "%23": '#',
    "%5B": '[',
    "%5D": ']',
    "%40": '@',
    "%21": '!',
    "%24": '$',
    "%26": '&',
    "%27": '\'',
    "%28": '(',
    "%29": ')',
    "%2A": '*',
    "%2B": '+',
    "%2C": ',',
    "%3B": ';',
    "%3D": '=',
    "%25": '%',
}
```

The companion deny middleware `RouterDeniedEncodedPathCharacters.Map()` at `pkg/config/dynamic/http_config.go:115-146` registers both upper and lower variants for the characters it gates (`%2F` + `%2f`, `%5C` + `%5c`, etc.), so slash and backslash specifically are not bypassable for routes that default-deny encoded slashes. The asymmetry remains for the other reserved characters (`* + , : = ; ? [ ] # @ !`) or when an operator explicitly opts into allowing semicolon / question-mark.

## Impact (correctness, narrow practical surface)

Practical exploit windows are narrow but the behaviour deviates from RFC 3986. Sibling pattern of Caddy CVE-2026-27587 (path %xx case-normalisation skip).

## Suggested fix

Either :

1. Replace `reservedCharacters` with a `func(string) (rune, bool)` that case-folds the hex characters before lookup.
2. Or expand the map at init time to include lower-case variants.

Option 1 is more concise and matches the pattern used by the deny middleware.

Reported as part of a periodic security review by the security advisor tooling ; surfacing here as a routing-path correctness bug, not a security advisory.

firasmosbehi:
### What does this PR do?

Normalizes percent-encoded character lookups in the HTTP muxer to be case-insensitive for reserved characters.

### Motivation

Fixes #13173

The reservedCharacters map in pkg/muxer/http/mux.go only contained uppercase hex entries (e.g. %2F). Per RFC 3986, %2f and %2F are equivalent, but Go's url.URL.EscapedPath() preserves the original case. This caused lowercase percent-encoded reserved characters to fall through to url.PathUnescape and get decoded into the routing path, while uppercase variants stayed encoded.

### How does it work?

A single-line change: strings.ToUpper(encodedCharacter) before looking up reservedCharacters.

### Testing

- [x] Unit tests added for lowercase reserved characters (%2f, %2a)
- [x] Existing tests pass

### Breaking Changes

- [x] None

### DIFF
--- a/pkg/muxer/http/mux.go
+++ b/pkg/muxer/http/mux.go
@@ -152,7 +152,7 @@ func withRoutingPath(req *http.Request) (*http.Request, error) {
 		}
 
 		encodedCharacter := escapedPath[i : i+3]
-		if _, reserved := reservedCharacters[encodedCharacter]; reserved {
+		if _, reserved := reservedCharacters[strings.ToUpper(encodedCharacter)]; reserved {
 			routingPathBuilder.WriteString(encodedCharacter)
 		} else {
 			// This should never happen as the standard library will reject requests containing invalid percent-encodings.
--- a/pkg/muxer/http/mux_test.go
+++ b/pkg/muxer/http/mux_test.go
@@ -694,6 +694,16 @@ func TestRoutingPath(t *testing.T) {
 			path:                "/foo%20bar%2Fbaz%23qux",
 			expectedRoutingPath: "/foo bar%2Fbaz%23qux",
 		},
+		{
+			desc:                "lowercase reserved percent-encoded slash is kept encoded",
+			path:                "/foo%2fbar",
+			expectedRoutingPath: "/foo%2fbar",
+		},
+		{
+			desc:                "lowercase reserved percent-encoded asterisk is kept encoded",
+			path:                "/foo%2abar",
+			expectedRoutingPath: "/foo%2abar",
+		},
 	}
 
 	for _, test := range tests {


nmengin:
Thank you for this fix! 

kevinpollet:
Hello @firasmosbehi and thanks for your contribution,

As explained in #13173, since `v3.4.1`, paths are normalized before reaching the muxer and percent-encoded sequences are [uppercased](https://github.com/traefik/traefik/blob/master/pkg/server/server_entrypoint_tcp.go#L839) per RFC 3986 [§6.2.2.1](https://datatracker.ietf.org/doc/html/rfc3986#section-6.2.2.1) ([migration notes](https://doc.traefik.io/traefik/migrate/v3/#request-path-normalization)). By the time `reservedCharacters` is consulted, `%2f` is already `%2F`, so the uppercase-only map is correct by design.

If you think that I missed something, feel free to leave a comment or reopen it.

PR #13189, opened 2026-05-17
fix: reject multi-label subdomains for Gateway API wildcard hostname listeners

### LINKED ISSUE #13171: Gateway API: HTTPRoute / TLSRoute wildcard hostname matcher accepts multi-label subdomains
## Summary

`findMatchingHostname` in `pkg/provider/kubernetes/gateway/kubernetes.go:1066-1084` uses `strings.HasSuffix` to compare a listener wildcard against a route hostname. This allows multi-label subdomains to attach to a single-label wildcard listener, deviating from the Gateway API spec.

## Spec reference

The Gateway API specification states that wildcard hostnames are interpreted as strict one-label DNS wildcards : a listener `*.foo.com` should accept `bar.foo.com` and reject `baz.bar.foo.com`. See [Gateway API spec, HTTPRoute hostnames](https://gateway-api.sigs.k8s.io/reference/spec/#httproute).

```go
func findMatchingHostname(h1, h2 gatev1.Hostname) gatev1.Hostname {
    if h1 == h2 {
        return h1
    }
    if !strings.HasPrefix(string(h1), "*.") {
        return ""
    }
    trimmedH1 := strings.TrimPrefix(string(h1), "*")
        return ""
    }
    return h2
}
```

`HasSuffix` matches `baz.bar.foo.com` against `*.foo.com` because both end with `.foo.com`. The existing unit test `Test_findMatchingHostnames / "Matching subsubdomain with listener wildcard"` at `pkg/provider/kubernetes/gateway/kubernetes_test.go:7588` explicitly pins this non-conformant behaviour.

## Impact (Gateway API conformance, not a CVE class)

In multi-tenant shared-Gateway setups, an operator who delegates `*.foo.com` to a tenant namespace via a listener wildcard is silently also delegating every deeper subdomain. A tenant in namespace X can claim `attacker.controlled.foo.com` even though the operator's mental model is "one DNS label only". Affects both HTTPRoute and TLSRoute.

## Suggested fix

When `h1` starts with `*.`, accept `h2` only when the number of labels in `h2` equals the number of labels in `h1`. Equivalent : count dots and require equality, or strip the wildcard label and the next single label from `h2` then compare. Update the existing test from "subsubdomain matches wildcard" to "subsubdomain rejected by wildcard", and add a release note since this is a behaviour change.

Reported as part of a periodic security review by the security advisor tooling ; surfacing here as a spec-conformance bug, not a security advisory.

firasmosbehi:
### What does this PR do?

Fixes findMatchingHostname to reject multi-label subdomains when matching against wildcard listener hostnames, aligning with the Gateway API specification.

### Motivation

Fixes #13171

The Gateway API specification states that wildcard hostnames are interpreted as strict one-label DNS wildcards: a listener *.foo.com should accept bar.foo.com and reject baz.bar.foo.com.

The previous implementation used strings.HasSuffix, which allowed baz.bar.foo.com to match *.foo.com because both end with .foo.com.

### How does it work?

After confirming the suffix match, we check that the prefix (the part before the matched suffix) contains no dots — unless it is itself a wildcard pattern. This ensures:
- bar.foo.com matches *.foo.com ✅
- baz.bar.foo.com does NOT match *.foo.com ✅
- *.bar.foo.com still matches *.foo.com ✅ (wildcard route hostnames)

### Testing

- [x] Updated existing unit tests to expect rejection for multi-label subdomains
- [x] Wildcard-wildcard matching test preserved

### Breaking Changes

This is a behavior change that brings Traefik into compliance with the Gateway API spec. Existing deployments relying on the non-conformant behavior (multi-label subdomains matching single-label wildcards) will be affected.

### DIFF
--- a/pkg/provider/kubernetes/gateway/kubernetes.go
+++ b/pkg/provider/kubernetes/gateway/kubernetes.go
@@ -1078,6 +1078,14 @@ func findMatchingHostname(h1, h2 gatev1.Hostname) gatev1.Hostname {
 		return ""
 	}
 
+	// Gateway API spec: wildcard matches exactly one label.
+	// Reject multi-label subdomains (e.g. *.foo.com must not match baz.bar.foo.com).
+	// Wildcard route hostnames (e.g. *.bar.foo.com) are still allowed.
+	prefix := strings.TrimSuffix(string(h2), trimmedH1)
+	if !strings.HasPrefix(prefix, "*.") && strings.Contains(prefix, ".") {
+		return ""
+	}
+
 	// h1 is a wildcard that encompasses h2, so h2 is always
 	// the more specific hostname (the correct intersection).
 	return h2
--- a/pkg/provider/kubernetes/gateway/kubernetes_test.go
+++ b/pkg/provider/kubernetes/gateway/kubernetes_test.go
@@ -7589,8 +7589,6 @@ func Test_findMatchingHostnames(t *testing.T) {
 			desc:             "Matching subsubdomain with listener wildcard",
 			listenerHostname: ptr.To(gatev1.Hostname("*.foo.com")),
 			routeHostnames:   []gatev1.Hostname{"baz.bar.foo.com"},
-			want:             []gatev1.Hostname{"baz.bar.foo.com"},
-			wantOk:           true,
 		},
 		{
 			desc:             "Matching subdomain with route hostname wildcard",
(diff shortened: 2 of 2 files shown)


nmengin:
Thank you for this fix! 

filip2mac:
Independently arrived at the same fix — good to see this already addressed. The approach here (checking that the concrete-hostname prefix contains no dot, while skipping the check for wildcard route hostnames) is exactly right per the spec. Left a comment to show support.

kevinpollet:
Hello @firasmosbehi and thanks for the contribution,

The underlying issue #13171 misreads the spec. Gateway API defines wildcards as a suffix match over **one or more** labels, not single-label. From [Listener.Hostname](https://gateway-api.sigs.k8s.io/reference/api-spec/1.5/spec/#gateway.networking.k8s.io/v1.Listener):

> a match for `*.example.com` would match both `test.example.com`, and `foo.test.example.com`, but not `example.com`.

This PR would reject `foo.test.example.com` on a `*.example.com` listener — the exact case the spec gives as a match. Merging would move us from conformant to non-conformant and break valid multi-tenant patterns.

Feel free to reopen or leave a comment if you think that I missed something.

PR #13190, opened 2026-05-17
fix: return catch-all rule when nginx ingress buildRule has no host or path

### LINKED ISSUE #13151: kubernetesIngressNGINX: error while parsing rule : parsing rule : 1:1: expected operand, found 'EOF'
### Welcome!

### What did you do?

Updated from Traefik 3.7.0-rc2 to 3.7.0 GA.

### What did you see instead?

After updating, I saw a lot of error messages in the log like:

`2026-05-11T12:36:33Z ERR error="error while parsing rule : parsing rule : 1:1: expected operand, found 'EOF'" entryPointName=web routerName=default-nginx-test01-ingress-default-backend-default-backend@kubernetesingressnginx`

Traefik v.3.7.0
Helm chart v40.0.0

```yaml
deployment:
  replicas: 2
 service:
   enabled: true
   spec:
     type: LoadBalancer
     externalTrafficPolicy: Local
api:
  dashboard: true
  insecure: true
providers:
  kubernetesIngressNGINX:
    enabled: true
  kubernetesGateway:
    enabled: false

kind: Ingress
metadata:
  name: nginx-test01-ingress-default-backend
spec:
  ingressClassName: nginx
  defaultBackend:
    service:
      name: nginx-test01-service
      port:
        number: 80
  rules:
  - http:
      paths:
      - path: /web
        pathType: Prefix
        backend:
          service:
            name: nginx-test01-service
            port:
              number: 80

_No response_

firasmosbehi:
### What does this PR do?

Ensures buildRule returns a valid catch-all rule (PathPrefix("/")) when both host and path are empty, instead of returning an empty string that crashes the rule parser.

### Motivation

Fixes #13151

When a Kubernetes ingress has both a defaultBackend and rules with no host field, the nginx ingress provider creates fallback default-backend locations for each rule. When buildRule is called with an empty host and empty path, it returned an empty string, causing the rule parser to fail with:
  parsing rule : 1:1: expected operand, found 'EOF'

### How does it work?

Added a guard at the end of buildRule: if the rules slice is empty, return PathPrefix("/") instead of joining an empty slice.

### Testing

- [x] Added unit tests for buildRule covering empty host/path, host-only, path-only, and host+path combinations
- [x] Existing provider tests pass

### Breaking Changes

- [x] None

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/translator.go
+++ b/pkg/provider/kubernetes/ingress-nginx/translator.go
@@ -606,6 +606,10 @@ func buildRule(host string, loc *location) string {
 		}
 	}
 
+	if len(rules) == 0 {
+		return `PathPrefix("/")`
+	}
+
 	return strings.Join(rules, " && ")
 }
 
--- a/pkg/provider/kubernetes/ingress-nginx/translator_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/translator_test.go
@@ -1,0 +1,56 @@
+package ingressnginx
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	netv1 "k8s.io/api/networking/v1"
+	"k8s.io/utils/ptr"
+)
+
+func TestBuildRule(t *testing.T) {
+	tests := []struct {
+		desc     string
+		host     string
+		loc      *location
+		expected string
+	}{
+		{
+			desc:     "empty host and path returns catch-all",
+			host:     "",
+			loc:      &location{},
+			expected: `PathPrefix("/")`,
+		},
+		{
+			desc:     "host only",
+			host:     "foo.localhost",
+			loc:      &location{},
+			expected: `Host("foo.localhost")`,
+		},
+		{
+			desc: "path only",
+			host: "",
+			loc: &location{
+				Path:     "/foo",
+				PathType: ptr.To(netv1.PathTypePrefix),
+			},
+			expected: `(Path("/foo") || PathPrefix("/foo/"))`,
+		},
+		{
+			desc: "host and path",
+			host: "foo.localhost",
+			loc: &location{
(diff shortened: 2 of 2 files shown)


nmengin:
Thank you for this fix! 

amazon7737:
The fix makes sense for the EOF error, but I wonder if the scope is a bit broader than needed.

Since the empty rule seems to come from the generated ingress default-backend fallback location for a hostless rule, would it make sense to scope the `PathPrefix("/")` fallback to that case instead of applying it to any empty host/path rule built by `buildRule`?

A provider-level regression test with the original manifest shape would also be useful to cover the full build/translate path.

amazon7737:
Hey, @firasmosbehi. I was looking through this PR together with #13151 and found one small test coverage idea that might be worth adding.

The current test covers `buildRule` directly, but the original issue happens through the full ingress loading path: `spec.defaultBackend` plus a hostless rule with a `/web` Prefix path.

I tried adding a fixture for that shape locally and checking that the generated default-backend HTTP/TLS routers use `PathPrefix("/")` instead of an empty rule. It passes on this PR, and fails with the original empty rule if the `buildRule` fallback is removed.

Might be useful to add that as a provider-level regression test here.

pkg/provider/kubernetes/ingress-nginx/translator.go
gndz07, line 210: I think we should set low priority on these routers when `loc.IsIngressDefaultBackend` is true, as has been done for cluster-level default backend at lines 75 and 83 (which sets the priority to `math.MinInt32`). Without it, the `PathPrefix("/")` catch-all gets an auto-computed priority of 13 (rule string length), which can shadow more specific routes from other ingresses.

gndz07 (changes requested):
Hi @firasmosbehi, thank you for opening the PR! Just a small review point.

PR #13194, opened 2026-05-18
docs: add 3.7 to the release documentation

vincentbernat:
### What does this PR do?

Add 3.7 to the release documentation.

### Motivation

Also note that 3.6 has no longer an active support. Don't put a security support end date as I am not sure if the 1-year support is enforced.

Note that https://endoflife.date/traefik information seems incorrect, hence this update.

Also, the PR template says 3.6 and 2.11. I think it should be updated too?

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/deprecation/releases.md
+++ b/docs/content/deprecation/releases.md
@@ -4,16 +4,17 @@
 
 Below is a non-exhaustive list of versions and their maintenance status:
 
-| Version | Release Date | Active Support     | Security Support  |
-|---------|--------------|--------------------|-------------------|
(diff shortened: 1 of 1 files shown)


nmengin:
Hello @vincentbernat, thank you for your contribution! However, it appears that PR #13118 is already open with the same purpose of adding v3.7 to the release documentation. To avoid duplicate efforts, we recommend closing this PR in favor of #13118. If #13118 needs updates or improvements (such as the PR template update you mentioned for versions 3.6 and 2.11), please feel free to comment on that PR or open a separate issue for the template updates. Thank you for your understanding!

PR #13198, opened 2026-05-18
muxer: normalise percent-encoding case before reservedCharacters lookup

### LINKED ISSUE #13173: Muxer routing path: reservedCharacters map is uppercase-hex only, percent-encoding case asymmetry on reserved characters

c-tonneslan:
Closes #13173.

`reservedCharacters` is keyed in uppercase, but `url.URL.EscapedPath` preserves the case the client sent, so `%2f` slips past the reserved check while `%2F` doesn't. Per RFC 3986 2.1 those two are equivalent. Uppercase `encodedCharacter` before the map lookup so two RFC-equivalent paths route to the same router.

### DIFF
--- a/pkg/muxer/http/mux.go
+++ b/pkg/muxer/http/mux.go
@@ -152,7 +152,10 @@ func withRoutingPath(req *http.Request) (*http.Request, error) {
 		}
 
 		encodedCharacter := escapedPath[i : i+3]
-		if _, reserved := reservedCharacters[encodedCharacter]; reserved {
+		// Per RFC 3986 section 2.1, %2A and %2a are equivalent. The map
+		// is keyed in uppercase, so normalise before lookup or two
+		// RFC-equivalent paths would route differently.
+		if _, reserved := reservedCharacters[strings.ToUpper(encodedCharacter)]; reserved {
 			routingPathBuilder.WriteString(encodedCharacter)
 		} else {
 			// This should never happen as the standard library will reject requests containing invalid percent-encodings.
--- a/pkg/muxer/http/mux_test.go
+++ b/pkg/muxer/http/mux_test.go
@@ -694,6 +694,13 @@ func TestRoutingPath(t *testing.T) {
 			path:                "/foo%20bar%2Fbaz%23qux",
 			expectedRoutingPath: "/foo bar%2Fbaz%23qux",
 		},
+		// #13173: lowercase percent-encoding of a reserved character is
+		// equivalent to uppercase per RFC 3986 section 2.1.
+		{
+			desc:                "lowercase reserved percent-encoded character is kept encoded",
+			path:                "/foo%2fbar",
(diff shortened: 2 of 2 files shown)


nmengin:
Thank you for this contribution! However, we already have PR #13188 open that addresses this exact issue with an identical solution. Both PRs normalize percent-encoded character lookups to be case-insensitive for reserved characters in the HTTP muxer, fixing issue #13173.

To avoid duplicate effort, we recommend:
1. Closing this PR in favor of #13188
2. If you'd like to contribute further, you could review or help test #13188 to ensure it properly handles all edge cases

We appreciate your attention to RFC 3986 compliance in Traefik!

c-tonneslan:
Sorry for the duplicate, I should've checked for an open PR before opening this. #13188 clearly got there first, thanks for the heads up.

PR #12808, opened 2026-03-13
Fix ingress router's rule

gndz07:
### What does this PR do?

This PR fixes router's rules generation for Knative, ingress-nginx, and ingress providers.
With the changes, the rules will now use the double quote syntax, which allow to escape any double quote in the rule value.

### Motivation

To make sure to use matcher values that are escaped.

### More

- [x] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

Co-authored-by: Mathis Urien <contact.lbf38@gmail.com>

PR #13201, opened 2026-05-19
Escape exact gRPC method matches

### LINKED ISSUE #13172: Gateway API: GRPCRoute Exact method matcher always builds PathRegexp, allowing dot-as-wildcard semantic drift
## Summary

`buildGRPCMethodRule` in `pkg/provider/kubernetes/gateway/grpcroute.go:392-408` unconditionally builds a `PathRegexp` rule, regardless of whether the `GRPCMethodMatch.Type` is `Exact` (default) or `RegularExpression`. Service names that contain a literal dot are therefore matched as regex wildcards at runtime.

```go
func buildGRPCMethodRule(method *gatev1.GRPCMethodMatch) string {
    if method == nil {
        return `PathPrefix("/")`
    }

    sExpr := "[^/]+"
    if s := ptr.Deref(method.Service, ""); s != "" {
        sExpr = s
    }

    mExpr := "[^/]+"
    if m := ptr.Deref(method.Method, ""); m != "" {
        mExpr = m
    }

    return fmt.Sprintf("PathRegexp(%q)", fmt.Sprintf("/%s/%s", sExpr, mExpr))
}
```

`method.Type` is never inspected. The Gateway API CEL validation permits `.` in Exact service names (the dot is a literal separator in the protobuf-style service name), but at runtime the dot is a regex wildcard. So a route advertising `service = "auth.api", type = Exact` will also match paths under `/authXapi/...`.

## Impact (Gateway API conformance)

The CEL alphabet is restricted enough that the practical attack surface is narrow (only the dot is permitted as a regex-meaningful character in Exact service names), but the route-matching semantics drift from the spec. In a multi-tenant cross-namespace setup, a tenant who registers an Exact-type GRPCRoute with a dot in the service name can silently capture traffic intended for a neighbour tenant whose service name differs only in the position of the dot.

## Suggested fix

Inspect `method.Type`. For Exact, wrap each side with `regexp.QuoteMeta` before format, or build a literal `Path` rule. Same logic for Header / QueryParam Exact matches if any other helper has the same shape (worth a sweep).

Add a unit test asserting that `GRPCRoute service = "foo.bar" Exact` does not match path `/fooXbar/Method`.

Reported as part of a periodic security review by the security advisor tooling ; surfacing here as a matcher-correctness bug.

nickmnt:
### What does this PR do?

Fixes `buildGRPCMethodRule` so exact gRPC service and method matches are escaped before building the route rule.

Previously, exact values were inserted into `PathRegexp` without escaping, so `.` in service names was treated as a regex wildcard.

This PR preserves regex behavior for `RegularExpression` matches and adds tests for exact escaping.

### Motivation

Fixes the behavior reported in #13172.

Gateway API allows `.` in exact gRPC service names, and those dots should be matched literally. Without escaping, route matching can be broader than intended.

### More

- [x] Added/updated tests

Closes #13172.

### DIFF
--- a/pkg/provider/kubernetes/gateway/grpcroute.go
+++ b/pkg/provider/kubernetes/gateway/grpcroute.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -394,14 +395,25 @@ func buildGRPCMethodRule(method *gatev1.GRPCMethodMatch) string {
 		return `PathPrefix("/")`
 	}
 
+	// Determine if we should treat values as literals or regex patterns.
+	isExact := method.Type == nil || *method.Type == gatev1.GRPCMethodMatchExact
+
 	sExpr := "[^/]+"
 	if s := ptr.Deref(method.Service, ""); s != "" {
-		sExpr = s
+		if isExact {
+			sExpr = regexp.QuoteMeta(s)
+		} else {
+			sExpr = s
+		}
 	}
 
 	mExpr := "[^/]+"
 	if m := ptr.Deref(method.Method, ""); m != "" {
-		mExpr = m
+		if isExact {
+			mExpr = regexp.QuoteMeta(m)
+		} else {
+			mExpr = m
+		}
 	}
 
 	return fmt.Sprintf("PathRegexp(%q)", fmt.Sprintf("/%s/%s", sExpr, mExpr))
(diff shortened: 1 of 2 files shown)


nmengin:
Thank you for this fix, @nickmnt! This PR correctly addresses the issue reported in #13172 where dots in gRPC service names were being treated as regex wildcards instead of literal matches in exact gRPC method rules.

The fix looks good:
- It properly escapes exact matches while preserving regex behavior for `RegularExpression` type matches
- Tests have been added to cover the escaping behavior
- The change is minimal and focused

This aligns well with similar fixes in the Gateway API provider (like #12753 and #12808) that ensure proper rule escaping. The PR is ready for technical review.

filip2mac:
Independently implemented the same fix for this one too. `regexp.QuoteMeta` on Exact-type values is exactly the right approach — keeps RegularExpression matches unchanged while making literal service/method names safe in PathRegexp rules.

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #13203, opened 2026-05-19
Trim quotes from proxy_set_header header name

crisbal:
### What does this PR do?

Apply `trimQuote` to the header name parameter in the `proxy_set_header` directive, so that quoted header names (e.g. `proxy_set_header "X-Custom-Header" "my-value"`) are handled correctly.

### Motivation

* Previously, the quotes were stripped from the value but not from the header name, causing the header to be set with literal quote characters in its name.
* This could lead to errors like: `net/http: invalid header field name "\"Authorization\""`
* `ingress-nginx` support quotes in `proxy_set_header` header names. [Reference 1](https://github.com/kubernetes/ingress-nginx/blob/main/test/e2e/annotations/auth.go#L311) [Reference 2](https://github.com/kubernetes/ingress-nginx/blob/main/internal/ingress/controller/template/template_test.go#L561)

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/ingressnginx/snippet/action.go
+++ b/pkg/middlewares/ingressnginx/snippet/action.go
@@ -393,7 +393,7 @@ func createProxySetHeaderAction(d config.IDirective) (action, error) {
 		return nil, errors.New("proxy_set_header directive requires 2 parameters (header and value)")
 	}
 
-	key := params[0].String()
+	key := trimQuote(params[0].String())
 	val := trimQuote(params[1].String())
 
 	return func(rw http.ResponseWriter, req *http.Request, ctx *actionContext) (bool, error) {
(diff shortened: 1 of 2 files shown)


nmengin:
Thank you for this PR! This fix addresses a clear bug where quoted header names in the `proxy_set_header` directive are not properly handled, resulting in invalid HTTP headers with literal quote characters. The change aligns well with how ingress-nginx handles quotes in similar directives, and the included tests demonstrate that the fix works correctly.

With that said, could you rebase on branch v3.7?

Thanks in advance

crisbal:
> Thank you for this PR! This fix addresses a clear bug where quoted header names in the `proxy_set_header` directive are not properly handled, resulting in invalid HTTP headers with literal quote characters. The change aligns well with how ingress-nginx handles quotes in similar directives, and the included tests demonstrate that the fix works correctly.
> 
> With that said, could you rebase on branch v3.7?
> 
> Thanks in advance

@nmengin Thanks for the review. How does this look now?  I should have rebased.

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #13205, opened 2026-05-19
Surface the Ingress status race condition during NGINX coexistence

emilevauge:
### What does this PR do?

Improves the Ingress NGINX to Traefik migration guide to make the status-field race condition between the two controllers visible and actionable.

Concretely, in `docs/content/migrate/nginx-to-traefik.md`:

- Promotes the previous collapsed `??? info "ExternalDNS Users"` note into an open `!!! warning "Ingress Status Race Condition During Coexistence"` block in Step 3.
- Names the affected field explicitly: `status.loadBalancer.ingress[]`.
- Broadens the impact section beyond ExternalDNS: kube-state-metrics, monitoring dashboards, ArgoCD/Flux (permanent drift on every affected Ingress), and custom operators that reconcile on Ingress status.
- States plainly that routing is not impacted by the race, only observability and anything driven by Ingress status.
- Adds a second mitigation option: the transitional-IngressClass approach SUSE documents for RKE2, with a link to the [SUSE migration guide](https://documentation.suse.com/cloudnative/rke2/latest/en/reference/ingress_migration.html).
- Adds an early pointer in Step 1 (Install Traefik) so readers see the trade-off before configuring `publishService`.

### Motivation

Users running both controllers in parallel with the same `ingressClassName: nginx` (the strategy the current migration guide promotes) hit a tight reconciliation loop where Traefik's `updateIngressStatus` and ingress-nginx's status writer keep overwriting each other's value in `status.loadBalancer.ingress[]`. The Traefik provider does skip the write when the observed status already equals its target (`isLoadBalancerIngressEquals` in `pkg/provider/kubernetes/ingress-nginx/client.go`), but as long as the other controller publishes a different IP, the loop never settles.

Today the issue is mentioned only as a collapsed ExternalDNS note, with the keyword "race condition" absent. Users discovering the problem (recent community thread: [Potential race condition when migrating ingress-nginx to Traefik](https://community.traefik.io/t/potential-race-condition-when-migrating-ingress-nginx-to-traefik/29704)) do not find the existing guidance from the symptoms they observe (kube-state-metrics flapping, ArgoCD drift, ExternalDNS bouncing).

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/migrate/nginx-to-traefik.md
+++ b/docs/content/migrate/nginx-to-traefik.md
@@ -154,6 +154,10 @@ Final:       DNS → LoadBalancer → Traefik → Your Services
     ```
 Install Traefik with the Kubernetes Ingress NGINX provider enabled. Both controllers will serve the same Ingress resources simultaneously.
 
+!!! warning "Read the status race condition note first"
+
+    Running both controllers against the same Ingresses creates contention on the `status.loadBalancer.ingress[]` field. Before installing, review the [Ingress Status Race Condition](#status-race) section in Step 3 and decide which mitigation to apply (disable `publishService` on Traefik, or use a transitional IngressClass).
+
 ### Add Traefik Helm Repository
 
 ```bash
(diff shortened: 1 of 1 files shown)


nmengin (approved):
LGTM

kevinpollet (approved):

PR #13206, opened 2026-05-19
Add ExcludeRequestContentType option to compress middleware

### LINKED ISSUE #12070: Proposal: Remove request content-type based filtering from excludedContentTypes behavior
### Welcome!

### Background
The excludedContentTypes option of Traefik’s compress middleware is intended to exclude specific content types from being compressed. Unlike includeContentTypes, which considers only the response header, the current behavior filters based on both request and response Content-Type. I believe this is not reasonable.

It should suffice for excludedContentTypes to only check the response Content-Type, just as includeContentTypes does. Adding a check on the request Content-Type introduces ambiguity for users.

As @rtribotte reminded in #12014, this is a legacy behavior that some users may rely on, so more discussion is needed before removing it.

KaanSimsek:
### What does this PR do?

Adds an excludeRequestContentType boolean option to the compress middleware. When set to false, the request's Content Type header is no longer checked before deciding whether to compress — only the response Content-Type is evaluated.

### Motivation

Currently the compress middleware skips compression entirely when the request Content-Type matches an excluded type (e.g. text/event-stream). This is semantically wrong — the request body's content type has no bearing on whether the response should be compressed. This option makes that behavior opt-in rather than forced, allowing users to compress responses regardless of what the client sends in the request.

### More

- [x] Added/updated tests

### Additional Notes

Closes #12070


nmengin:
Thank you for this PR! This addresses a valid semantic issue with the compress middleware - request Content-Type should not determine whether the response gets compressed.

Great work, including tests and documentation updates!

PR #13214, opened 2026-05-20
Move snicheck to ctx instead of simulated routing

juliens:
### What does this PR do?

This PR changes how domain fronting verification is handled.

Before this PR, we attempted to reproduce the TCP and HTTP routing logic to determine whether the same TLS options were selected. However, because the muxers are more complex, this could lead to inconsistent or incorrect behavior.

With this change, the TLS layer now stores the selected TLS options in the request context. A dedicated HTTP middleware then verifies that the TLS options associated with the matched router are the same as those selected during the TLS phase.

### Motivation

To fix cases where SNICheck and the routing logic are inconsistent.


kevinpollet (approved):

youkoulayley (approved):
LGTM

PR #13215, opened 2026-05-20
Reject requests with different paths after StripPrefix and StripPrefixRegex normalisation

rtribotte:
### What does this PR do?

This PR adds the rejection of requests with different paths after the StripPrefix and StripPrefixRegex normalisation.

### Motivation

To avoid the stripped path from being interpreted as a different resource by the upstream service.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/stripprefixregex/strip_prefix_regex.go
+++ b/pkg/middlewares/stripprefixregex/strip_prefix_regex.go
@@ -73,9 +73,6 @@ func (s *stripPrefixRegex) ServeHTTP(rw http.ResponseWriter, req *http.Request) 
 				req.URL = req.URL.JoinPath()
 			}
 
-			if path != "" && !strings.HasPrefix(path, "/") {
-				path = "/" + path
-			}
 			// Stop here if the normalization of the path produces a different path.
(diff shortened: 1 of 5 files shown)


pkg/middlewares/stripprefixregex/strip_prefix_regex.go
```diff
@@ -68,10 +68,17 @@ func (s *stripPrefixRegex) ServeHTTP(rw http.ResponseWriter, req *http.Request)
...
 			// to be aligned with ensureLeadingSlash behavior.
-			if req.URL.Path != "" {
+			path := req.URL.Path
+			if path != "" {
 				req.URL = req.URL.JoinPath()
 			}
 
+			// Stop here if the normalization of the path produces a different path.
+			if path != req.URL.Path {
+				http.Error(rw, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
```
youkoulayley, line 81: Since this is a new behavior and can be a bit obscure why the path is rejected, should we add a debug log here? (like in basic auth for example)

youkoulayley (approved):

kevinpollet (approved):

PR #13219, opened 2026-05-21
Add nginx.ingress.kubernetes.io/enable-global-auth to the list of supported annotations

filip2mac:
### What does this PR do?

Fills every empty **Limitations / Notes** cell in `docs/content/reference/routing-configuration/kubernetes/ingress-nginx.md` — the ingress-nginx annotation compatibility reference table.

Also corrects a misclassification: `nginx.ingress.kubernetes.io/enable-global-auth` was listed in the Unsupported table but is fully implemented (`pkg/provider/kubernetes/ingress-nginx/middleware.go` reads it to opt an Ingress out of `globalAuthUrl`). It has been moved to the Authentication supported section.

### Motivation

The reference table had ~65 empty Notes cells across all sections (Authentication, SSL/TLS, Session Affinity, Load Balancing, Buffering, CORS, Routing, IP Allowlist, and the Unsupported table groups). Operators migrating from NGINX Ingress Controller need these notes to understand behavioral differences, accepted values, and defaults.

All notes were verified against the Go source in `pkg/provider/kubernetes/ingress-nginx/` before being written.

### More

- [ ] Added/updated tests

### Additional Notes

Changes are documentation-only — no Go code modified. The 11 commits are organized one per table section for easier review.


rtribotte:
Hi @filip2mac, 

Thank you for the effort here.

The goal of this reference table is to tell readers which annotations Traefik supports, which it doesn't, and to flag any subtleties where the support has caveats or behaves differently than they might expect.

It's deliberately a compatibility map, not a behavioral manual. Filling every Notes cell with descriptions of accepted values, defaults, and behavioral details starts to mirror the ingress-nginx documentation itself, and that's something we want to avoid. For anything beyond "is this supported," the upstream ingress-nginx documentation is the right source of truth.

So our preference is to keep most of those cells empty by design, and reserve notes specifically for the cases where there's a real subtlety worth calling out — a caveat, a partial implementation, or a difference in behavior an operator would otherwise be surprised by.

One genuinely useful thing did surface in here, though: the `nginx.ingress.kubernetes.io/enable-global-auth` misclassification. That's a real bug in the docs worth fixing. Would you be willing to change this PR for a small, focused change for just that correction?

Thanks again for digging into this.

filip2mac:
Thanks for the clear explanation — that makes sense. I've slimmed the PR down to just the `enable-global-auth` misclassification fix: removed from Unsupported, added to the Authentication table.

rtribotte (approved):
Thanks!

Issue #13078, opened 2026-04-29
Traefik v3.7.0-rc.2 ingress-nginx provider proxy-body-size annotation requires buffering


mliang2:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

In order for the `nginx.ingress.kubernetes.io/proxy-body-size` Nginx annotation to work, Traefik requires either `nginx.ingress.kubernetes.io/proxy-request-buffering: "on"` annotation or setting `.providers.kubernetesIngressNGINX.proxyRequestBuffering = true` on helm.

If neither are set, the size limit from the `proxy-body-size` is not respected.

This behavior is different from ingress-nginx where `proxy-body-size` is always respected, regardless if `nginx.ingress.kubernetes.io/proxy-request-buffering` is `on` or `off`

### What did you see instead?

`nginx.ingress.kubernetes.io/proxy-body-size` only works if `nginx.ingress.kubernetes.io/proxy-request-buffering: "on"` or setting `.providers.kubernetesIngressNGINX.proxyRequestBuffering = true` at the NGINX provider level

### What version of Traefik are you using?

Traefik v3.7.0-rc.2
Helm chart version: 40.0.0-rc.2

### What is your environment & configuration?

Steps to reproduce:

* Create `traefik-helm-values.yaml`
```yml
deployment:
  replicas: 1

ingressClass:
  isDefaultClass: false
  name: traefik

providers:
  kubernetesIngressNGINX:
    enabled: true
    ingressClassByName: true
    controllerClass: k8s.io/ingress-nginx
    ingressClass: nginx

    allowSnippetAnnotations: true

logs:
  access:
    enabled: true
```
* create the `httpbin.yaml` ingress config
```yml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: httpbin
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app: httpbin
  template:
    metadata:
      labels:
        app: httpbin
    spec:
      hostname: default
      containers:
      - name: httpbin
        image: ghcr.io/mccutchen/go-httpbin
        ports:
        - containerPort: 8080
        env:
        - name: USE_REAL_HOSTNAME
          value: "true"
        livenessProbe:
          httpGet:
            path: /status/200
            port: 8080
[...]

PR #13220, opened 2026-05-21
fix(ingress-nginx): enforce proxy-body-size when proxy-request-buffering is off

filip2mac:
### What does this PR do?

Fixes `nginx.ingress.kubernetes.io/proxy-body-size` being silently ignored when `nginx.ingress.kubernetes.io/proxy-request-buffering` is set to `"off"`.

**Root cause — two compounding issues in `buildBuffering`:**

1. `proxy-body-size` was parsed inside the `if !disableReq` block, so it was never applied when request buffering was disabled.
2. If both request and response buffering flags were off, the function returned early before creating any middleware — discarding the body size annotation entirely.
3. The underlying `oxy/buffer` library only calls `checkLimit` (which enforces `MaxRequestBodyBytes`) when `disableRequest` is false, so setting the field with `DisableRequestBuffer: true` would have no effect anyway.

**Fix:** Parse `proxy-body-size` before the early-return check. When a non-zero body size limit is configured, force request buffering on so the oxy library actually enforces the limit. This matches NGINX's behaviour where `client_max_body_size` (the underlying directive) is enforced independently of `proxy_request_buffering`.

### Motivation

In NGINX, `proxy-body-size` maps to `client_max_body_size`, which is a server-layer directive that always applies regardless of buffering. Traefik's implementation was not matching this behaviour, causing body size limits to be silently dropped for ingresses that disabled request buffering.

Reported in #13078.

### More

- [x] Added/updated tests

### Additional Notes

The fix adds a test fixture (`ingress-with-proxy-body-size-buffering-off.yml`) and a test case that explicitly asserts `DisableRequestBuffer` is `false` when `proxy-body-size` is set, confirming the body size limit will be enforced.

Trade-off: if an ingress sets both `proxy-request-buffering: "off"` (for streaming) and `proxy-body-size`, request buffering will now be enabled for that ingress. This is necessary because there is no streaming body-size-enforcement mechanism in the current buffering middleware. The alternative — silently ignoring the size limit — is worse.

### DIFF
--- a/pkg/middlewares/buffering/buffering.go
+++ b/pkg/middlewares/buffering/buffering.go
@@ -76,7 +76,7 @@ func (b *buffer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		}
 		// For streaming requests (chunked or unknown length), wrap the body so the
 		// limit is enforced as bytes are read, matching NGINX's client_max_body_size
-		// behaviour without enabling request buffering.
+		// behavior without enabling request buffering.
 		req.Body = http.MaxBytesReader(rw, req.Body, b.maxBodyBytes)
 	}
 	b.buffer.ServeHTTP(rw, req)
--- a/pkg/middlewares/buffering/buffering.go
+++ b/pkg/middlewares/buffering/buffering.go
@@ -16,8 +16,10 @@ const (
 )
 
 type buffer struct {
-	name   string
-	buffer *oxybuffer.Buffer
+	name                 string
+	buffer               *oxybuffer.Buffer
+	maxBodyBytes         int64
+	disableRequestBuffer bool
 }
 
 // New creates a buffering middleware.
@@ -54,8 +56,10 @@ func New(ctx context.Context, next http.Handler, config dynamic.Buffering, name 
 	}
 
 	return &buffer{
-		name:   name,
-		buffer: oxyBuffer,
+		name:                 name,
+		buffer:               oxyBuffer,
+		maxBodyBytes:         config.MaxRequestBodyBytes,
+		disableRequestBuffer: config.DisableRequestBuffer,
 	}, nil
 }
 
@@ -64,5 +68,16 @@ func (b *buffer) GetTracingInformation() (string, string) {
 }
 
 func (b *buffer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	if b.disableRequestBuffer && b.maxBodyBytes > 0 && req.Body != nil && req.Body != http.NoBody {
+		// Reject immediately when Content-Length is known and exceeds the limit.
+		if req.ContentLength > b.maxBodyBytes {
+			http.Error(rw, http.StatusText(http.StatusRequestEntityTooLarge), http.StatusRequestEntityTooLarge)
+			return
+		}
+		// For streaming requests (chunked or unknown length), wrap the body so the
+		// limit is enforced as bytes are read, matching NGINX's client_max_body_size
+		// behaviour without enabling request buffering.
+		req.Body = http.MaxBytesReader(rw, req.Body, b.maxBodyBytes)
+	}
 	b.buffer.ServeHTTP(rw, req)
 }
--- a/pkg/middlewares/buffering/buffering_test.go
+++ b/pkg/middlewares/buffering/buffering_test.go
@@ -3,6 +3,8 @@ package buffering
 import (
 	"bytes"
 	"crypto/rand"
+	"errors"
+	"io"
 	"math"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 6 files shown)


nmengin:
Hey @filip2mac,

nandorKollar:
I'm not sure that silently turning request buffering on, when `proxy-request-buffering` is set to `off` is a good idea. I think this is also a different behaviour when we compare against Nginx.

It seems to me, that buffering and request body size check should be handled separately. Instead of handling the request body size limit in buffering middleware, we might need a dedicated middleware?

filip2mac:
You're right that NGINX handles these independently — `client_max_body_size` is enforced at the HTTP layer via a streaming read limit, completely separate from `proxy_request_buffering`. My fix doesn't replicate that.

The constraint I hit is that `oxy/buffer` only enforces `MaxRequestBodyBytes` when request buffering is enabled, so there's no way to get size enforcement through that library without enabling buffering.

That said, I'd argue the tradeoff cuts in favour of this fix: the current bug silently drops a body size limit entirely — a security-relevant annotation that fails open. Silently enabling buffering (with the behaviour documented in comments and the PR description) seems like the lesser of the two evils. A user who set `proxy-body-size` clearly wants a limit enforced.

That said, a dedicated middleware using `http.MaxBytesReader` would be the architecturally correct solution — it enforces the limit as a streaming check without buffering, matching NGINX's actual semantics. Happy to go that route if the team prefers it. Would you rather I rewrite the fix that way, or treat that as a follow-up?

nandorKollar:
> You're right that NGINX handles these independently — `client_max_body_size` is enforced at the HTTP layer via a streaming read limit, completely separate from `proxy_request_buffering`. My fix doesn't replicate that.
> 
> The constraint I hit is that `oxy/buffer` only enforces `MaxRequestBodyBytes` when request buffering is enabled, so there's no way to get size enforcement through that library without enabling buffering.
> 
> That said, I'd argue the tradeoff cuts in favour of this fix: the current bug silently drops a body size limit entirely — a security-relevant annotation that fails open. Silently enabling buffering (with the behaviour documented in comments and the PR description) seems like the lesser of the two evils. A user who set `proxy-body-size` clearly wants a limit enforced.
> 
> That said, a dedicated middleware using `http.MaxBytesReader` would be the architecturally correct solution — it enforces the limit as a streaming check without buffering, matching NGINX's actual semantics. Happy to go that route if the team prefers it. Would you rather I rewrite the fix that way, or treat that as a follow-up?

The final recommendation is provided by the core maintainers, let's wait for their verdict. I personally think, that accidentally turning on buffering is not less confusing, than the original problem we'd like to solve here. It causes a regression (setting `nginx.ingress.kubernetes.io/proxy-request-buffering` to `"off"` works well now, while with combining with `nginx.ingress.kubernetes.io/proxy-body-size` buffering will silently turned on), thus possibly can have an unexpected memory impact. If this is acceptable, at least we should document it as a note for `nginx.ingress.kubernetes.io/proxy-body-size` annotation.

gndz07:
Hi @filip2mac,

Thank you for opening the PR. After reviewing, I agree with @nandorKollar that silently turning request buffering on is not the best approach. Considering that `proxy-body-size` has a default value, users that hasn't set this value but explicitly set buffering to "off" will have their buffering on. This also diverge from ingress-nginx's behavior.

As you mentioned, the fix should enforce the body size limit as a streaming check, without forcing request buffering on; this will be the architecturally correct solution and matches ingress-nginx's behavior. If you don't mind to update this PR to take this approach, that would be great. Also, could you please base this PR on v3.7 as this is categorized as a fix?

filip2mac:
Thanks for the clear direction, @gndz07 and @nandorKollar the PR has been updated.

**What changed:**

The silent buffering-on hack is gone. Instead, the fix is split across two layers:

1. **`pkg/provider/kubernetes/ingress-nginx/middleware.go`**  `proxy-body-size` is now parsed before the early-return check, and `DisableRequestBuffer` is left as-is (no longer forced to `false`). The early-return condition is tightened to `disableReq && disableResp && maxRequestBodyBytes == 0`, so a middleware is still created when a body size limit needs to be enforced in streaming mode.

2. **`pkg/middlewares/buffering/buffering.go`**  when `DisableRequestBuffer` is `true` and `MaxRequestBodyBytes > 0`, `ServeHTTP` now:
   - rejects immediately with 413 if `Content-Length` is known and exceeds the limit (before touching the upstream)
   - otherwise wraps `req.Body` with `http.MaxBytesReader` so the limit is enforced as bytes stream through, without buffering

The branch has also been rebased onto `v3.7`.

pkg/middlewares/buffering/buffering.go
```diff
@@ -64,5 +68,16 @@ func (b *buffer) GetTracingInformation() (string, string) {
...
+	if b.disableRequestBuffer && b.maxBodyBytes > 0 && req.Body != nil && req.Body != http.NoBody {
+		// Reject immediately when Content-Length is known and exceeds the limit.
+		if req.ContentLength > b.maxBodyBytes {
+			http.Error(rw, http.StatusText(http.StatusRequestEntityTooLarge), http.StatusRequestEntityTooLarge)
+			return
+		}
+		// For streaming requests (chunked or unknown length), wrap the body so the
+		// limit is enforced as bytes are read, matching NGINX's client_max_body_size
+		// behaviour without enabling request buffering.
+		req.Body = http.MaxBytesReader(rw, req.Body, b.maxBodyBytes)
```
gndz07, line 80: In case of error, `*http.MaxBytesError` isn't handled in `ComputeStatusCode` (pkg/proxy/httputil/proxy.go), so the streaming path returns 500 instead of 413. Could you add a case for it there?

gndz07 (changes requested):
Hi @filip2mac, thank you for the update! Just a quick feedback about the error handling.

pkg/middlewares/buffering/buffering.go
```diff
@@ -64,5 +68,16 @@ func (b *buffer) GetTracingInformation() (string, string) {
 }
 
 func (b *buffer) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
+	if b.disableRequestBuffer && b.maxBodyBytes > 0 && req.Body != nil && req.Body != http.NoBody {
```
nandorKollar, line 71: Not critical, but you might consider adding an test case to cover at least partially the new changes in this middleware. I don't think the current tests for buffering middleware has coverage for this, no? What do you think?

pkg/middlewares/buffering/buffering.go
filip2mac: Fixed. Added a  case to  in `pkg/proxy/httputil/proxy.go` so the streaming path now returns 413 instead of falling through to 500.

pkg/middlewares/buffering/buffering.go
filip2mac: Added  covering four cases: no limit, Content-Length within limit, Content-Length exceeding limit (immediate 413), and streaming body exceeding limit (413 on read via MaxBytesReader).

filip2mac:
Both review comments addressed in the latest push:

- `ComputeStatusCode` in `pkg/proxy/httputil/proxy.go` now handles `*http.MaxBytesError` and returns 413, so the streaming body-size limit path no longer falls through to 500.
- Added `TestBufferingDisabledRequestBuffer` covering the Content-Length early-rejection path and the streaming read-limit path.

@gndz07 @nandorKollar — happy to iterate further if anything else comes up.

pkg/middlewares/buffering/buffering_test.go
```diff
@@ -68,3 +70,90 @@ func TestBuffering(t *testing.T) {
 		})
 	}
 }
+
+func TestBufferingDisabledRequestBuffer(t *testing.T) {
+	payload := make([]byte, 128)
+	_, _ = rand.Read(payload)
```
nandorKollar: nit: if none of the return values are needed, just omit them: `rand.Read(payload)`

nandorKollar (commented):
Overall, LGTM.

gndz07:
Thank you for the update @filip2mac, I fixed the linter and updated the branch, but many tests are failing. I believe the issue is that the new enforcement of `proxy-body-size` when buffering is off uses `p.ProxyBodySize`, which defaults to 1MB. This means the Buffering middleware is now attached to every ingress regardless of whether the `proxy-body-size` annotation is explicitly set. Could you please update the tests where necessary?

PR #12685, opened 2026-02-14
Add router-level open connections metric for TCP routers

pavanreddy07385:
### What does this PR do?

Reintroduces a router-level open connections gauge metric (`traefik_router_open_connections`) for TCP routers, with `router` and `service` labels. This restores per-service connection visibility that was available in Traefik v2 but removed in v3.

Fixes #11006

### Motivation

In Traefik v3, `traefik_open_connections` only exposes entrypoint-level aggregation and no longer includes router/service labels. This makes it impossible to attribute open connections to specific backends, which is critical for:

- Per-service load visibility
- Accurate autoscaling signals (e.g. KEDA, HPA)
- Detecting stuck or overloaded backends

This change restores observability parity with Traefik v2 while aligning with v3's metric model.

### Implementation Details

- Added `RouterOpenConnectionsGauge()` to the metrics `Registry` interface
- Implemented `metricsMiddleware` in the TCP router manager that wraps handlers to track open connections (increment on connect, decrement on close via `defer`)
- Applied metrics wrapping to all three TCP routing paths: non-TLS, TLS passthrough, and TLS terminating
- Registered the metric across all 5 backends:
  - **Prometheus**: `traefik_router_open_connections` with labels `["router", "service"]`
  - **Datadog**: `router.open.connections`
  - **StatsD**: `router.open.connections`
  - **InfluxDB2**: `traefik.router.open.connections`
  - **OpenTelemetry**: `traefik_router_open_connections`
- Gated behind the existing `AddRoutersLabels` config flag, consistent with other router metrics
- Existing `traefik_open_connections` (entrypoint-level) remains unchanged — fully backward compatible

### More

- [x] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

- Low cardinality: labels are limited to `router` and `service` names from the dynamic configuration
- Stale metric cleanup is handled via `promState.vectors` for Prometheus
- Tests cover both handler construction with correct label wiring and direct middleware increment/decrement behavior
- This issue is tagged `contributor/wanted` on the upstream tracker

Issue #11006, opened 2024-08-14
Incomplete metrics in Traefik v3


bryanhorstmann:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

In Traefik v3.1.0, the new `traefik_open_connections` has replaced `traefik_service_open_connections`. But this metric is missing crucial labels. Namely: `exported_service`, `method` and `protocol`.

I've looked at https://github.com/traefik/traefik/issues/10928, but this does not address the missing labels.

### What did you see instead?

`traefik_open_connections` has limited labels:
```traefik_open_connections{container="traefik", endpoint="metrics", entrypoint="web", instance="10.1.204.153:9100", job="traefik-metrics", namespace="ingress", pod="traefik-75d56bbd64-kb62z", protocol="TCP", service="traefik-metrics"}```

### What version of Traefik are you using?

```
→ kubectl exec -it deploy/traefik -- traefik version
Version:      3.1.0
Codename:     comte
Go version:   go1.22.5
Built:        2024-07-15T14:44:04Z
OS/Arch:      linux/amd64
```

### What is your environment & configuration?

Helm `values.yaml`
```yaml
  ports:
    metrics:
      expose:
        default: false

  metrics:
    addInternals: false
    prometheus:
      service:
        enabled: true
      serviceMonitor:
        enabled: true
        namespace: ingress
```

Add more configuration information here.

rtribotte:
Hello @bryanhorstmann,

Thanks for your interest in Traefik!

The labels are not missing, this change has been done because the implementation was producing misleading figures (counting of HTTP requests instead of TCP connections), and the metric being unrelated to services, it has been reimplemented at the entryPoint level: https://doc.traefik.io/traefik/migration/v2-to-v3-details/#open-connections-metric

If this issue is a feature request, could you please elaborate on your need?

bryanhorstmann:
Its not truely a feature request if this was how it used to work. The documentation says those metrics have all been combined into `traefik_open_connections`. But relevant labels are no longer present.

> The equivalent to traefik_entrypoint_open_connections, traefik_router_open_connections and traefik_service_open_connections is now traefik_open_connections.

`traefik_open_connections` only gives amount of connections per entrypoint. Previosuly I could get metrics on protocol, service and method.

[...]

Issue #10733, opened 2024-05-17
traefik_open_connections metric drifts down until negative


navaati:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I configured Prometheus monitoring on a traefik instance with `--metrics.prometheus=true` on a separate entrypoint (for a separate port) with no other Prometheus specific configuration, so I got the default set of metrics. I can query the /metrics allright and I would expect traefik_open_connections to stay at 5 (as that’s my load, 5 clients from a load testing tool).

### What did you see instead?

The value of the traefik_open_connections metric goes down over time, until it even gets negative, which doesn’t make sense for a connection count (how can you have -1 open connections…).

See this excerpt from the /metrics:
```
# HELP traefik_open_connections How many open connections exist, by entryPoint and protocol
# TYPE traefik_open_connections gauge
traefik_open_connections{entrypoint="metrics",protocol="TCP"} -1
traefik_open_connections{entrypoint="traefik",protocol="TCP"} 0
traefik_open_connections{entrypoint="web",protocol="TCP"} -1
```
and this screenshot from Prometheus:
![image](https://github.com/traefik/traefik/assets/1193282/83b8dcf6-4502-4ee1-a4ca-6284148930d0)

In the screenshot, the metric is supposed to stay at a constant 5 connections.

As a proof that there is actually still 5 connections, here is the output of `ss` (where we also see the connections to the backend service):
```
State           Recv-Q           Send-Q                           Local Address:Port                                 Peer Address:Port           Process
ESTAB           0                0                                   172.18.0.3:42488                                  172.18.0.2:8000            users:(("traefik",pid=2506,fd=19))
ESTAB           0                0                                   172.18.0.3:32842                                  172.18.0.2:8000            users:(("traefik",pid=2506,fd=18))
ESTAB           0                0                                   172.18.0.3:38010                                  172.18.0.2:8000            users:(("traefik",pid=2506,fd=17))
ESTAB           0                0                                   172.18.0.3:36600                                  172.18.0.2:8000            users:(("traefik",pid=2506,fd=16))
ESTAB           0                0                                   172.18.0.3:35230                                  172.18.0.2:8000            users:(("traefik",pid=2506,fd=23))
[...]

PR #9656, opened 2023-01-10
Fix open connections metric

mpl:
### What does this PR do?

This PR removes the (broken) "Open connections" metric from the entryPoint, router, and service, and replaces them with a new (correct) metric at the server/global level.

As a notable side-effect, this is now the first metric that can be considered a TCP metric.

### Motivation

The previous metric Open connections was incorrect as it was computed. It was done at the ServeHTTP level, and was therefore basically computing the number of inflight requests instead of the number of connections.

This PR therefore fixes the implementation so that it matches the intended meaning, and computes the metric at the TCP level, where/when the connections are opened and closed.

Consequently, it does not make any sense to keep labels related to the application/HTTP layer, such as the protocol (as it was before) and the method. There is still a protocol label, but its meaning will now be to differentiate the transport (TCP VS UDP).

### More

- [ ] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

Addresses #7575 
Fixes #4226

PR #13221, opened 2026-05-21
Add per-service in-flight requests gauge

eugenetaranov:
### What does this PR do?

Adds a new gauge metric `traefik_service_inflight_requests` (with equivalent dotted variants for Datadog/StatsD/InfluxDB2 and an OTLP gauge of the same name) that tracks the current number of in-flight HTTP requests per service, labelled by `service`, `method`, and `protocol`. The metric is wired into the existing per-service metrics middleware: increment on entry, decrement deferred to handler return. It is registered across every metrics backend (Prometheus, StatsD, Datadog, InfluxDB2, OTLP) and gated by the existing `AddServicesLabels` flag.

### Motivation

Issue #11006 reports that `traefik_open_connections` - the v3 replacement (added in #9656) for the v2 `traefik_service_open_connections` - lacks the `service`, `method`, and `protocol` labels needed for per-service load visibility. PR #12685 reintroduces per-router visibility for TCP routers; this PR is the complementary piece for HTTP services.

The name `inflight_requests` is deliberate. #9656 removed the v2 `service_open_connections` because the implementation measured in-flight `ServeHTTP` calls rather than connections - the semantics didn't match the name. This PR keeps the same measurement point and renames the metric so the name reflects what is actually counted.

A key driver for restoring this signal is upgraded connections. For WebSocket and SSE the fast-proxy upgrade path blocks until the session closes, so the deferred decrement only fires at close — meaning the gauge correctly accounts for long-lived sessions. Those sessions never increment `traefik_service_request_duration_seconds_count` during their lifetime, so `rate(...request_duration_seconds_count[1m])` cannot expose them; this gauge is the only signal that does.

Inc/dec share a snapshotted label slice so the inc-dec label incoherence fixed in #10733 cannot recur here.

### More

- [x] Added/updated tests

### Additional Notes

- **Behaviour change**: when `AddServicesLabels` is enabled, a new metric appears in the exposition. No existing metric is renamed or removed; the v3 entrypoint-level `traefik_open_connections` is untouched.
- **Tests**: `pkg/middlewares/metrics/inflight_requests_test.go` covers two cases — a deterministic HTTP request (release-channel-synchronised mid-flight scrape asserts the gauge is 1, post-completion poll asserts it returns to 0), and a WebSocket session that the test holds open across a scrape to assert the gauge stays at 1 for the session lifetime and only decrements at close. The WebSocket case is the one that justifies the metric over the histogram-rate alternative.
- **Docs**: `docs/content/reference/install-configuration/observability/metrics.md` — one row added to each of the five backend tables (OTLP, Prometheus, Datadog, InfluxDB2, StatsD).
- **Scope**: HTTP services only. TCP-router visibility is addressed by #12685.


rtribotte:
Hi @eugenetaranov,

PR #13222, opened 2026-05-21
fix(gateway): spec-correct GRPCRoute exact matching and wildcard hostname enforcement

### LINKED ISSUE #13171: Gateway API: HTTPRoute / TLSRoute wildcard hostname matcher accepts multi-label subdomains

### LINKED ISSUE #13172: Gateway API: GRPCRoute Exact method matcher always builds PathRegexp, allowing dot-as-wildcard semantic drift

filip2mac:
### What does this PR do?

Fixes two Gateway API spec violations in the Kubernetes Gateway API provider.

**1. GRPCRoute Exact method matcher — regex metachar escape (#13172)**

`buildGRPCMethodRule` inserted service and method name strings verbatim into a `PathRegexp` rule. A service named `my.service.example` produced the pattern `/my.service.example/[^/]+`, where each `.` is a regex wildcard — so `myXserviceYexample` would also match. Per the spec, `GRPCMethodMatchExact` (the default) is a literal match.

Fix: wrap Exact-type values with `regexp.QuoteMeta` before inserting into the pattern. `RegularExpression` values pass through unchanged.

**2. Wildcard hostname one-label enforcement (#13171)**

`findMatchingHostname` used `strings.HasSuffix` to test whether a hostname falls under a wildcard listener. This allowed `baz.bar.foo.com` to match a `*.foo.com` listener. Per the [Gateway API spec](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.Hostname), `*.foo.com` is a strict single-label wildcard — only `<one-label>.foo.com` should match.

Fix: after the suffix check, when `h2` is a concrete hostname (not itself a wildcard), verify the non-suffix prefix contains no dot. Wildcard-under-wildcard matching (`*.bar.foo.com` under `*.foo.com`) is unaffected.

### Motivation

Both bugs cause Traefik's Gateway API routing to silently accept requests it should reject, producing false-positive route matches that deviate from the Gateway API specification.

### More

- [x] Added/updated tests

### Additional Notes

The two fixes are independent and touch different files; they are grouped in one PR since both are small spec-compliance corrections in the same package.

Closes #13172
Closes #13171


filip2mac:
Closing as duplicate of #13189 (wildcard hostname fix) and #13201 (GRPCRoute regex escaping). Both issues were already covered — leaving these to the original authors.

PR #13223, opened 2026-05-21
fix(healthcheck): run first check immediately instead of waiting one interval

### LINKED ISSUE #13218: Traefik v3 WRR load balancer occasionally dispatches to DOWN servers in steady state
### Welcome!

### What did you do?

I have a blue/green deployment using the file provider, with two servers per service load balancer. In steady state, only the `green` container is deployed; the `blue` container does not exist (removed via `docker compose down` over 24 hours ago, no DNS record, no container alias).

Dynamic configuration:

```yaml
http:
  routers:
    to-myservice:
      rule: "Host(`myservice.mydomain.local`)"
      service: myservice-lb
      entryPoints:
        - websecure
      tls: true
      middlewares:
        - myservice-retry

  middlewares:
    myservice-retry:
      retry:
        attempts: 2
        initialInterval: 100ms

  services:
    myservice-lb:
      loadBalancer:
        servers:
          - url: http://myservice-blue:80
          - url: http://myservice-green:80
        passHostHeader: true
        healthCheck:
          path: /healthz
          interval: 5s
          timeout: 3s
          scheme: http
```

The `blue` hostname does not resolve from inside the Traefik container:

```
$ docker exec traefik getent hosts myservice-blue
(no output)

$ docker exec traefik nslookup myservice-blue
Server:         127.0.0.11
Address:        127.0.0.11:53
** server can't find myservice-blue.mydomain.local: SERVFAIL

$ docker exec traefik wget -qO- -S http://myservice-blue:80/
wget: bad address 'myservice-blue:80'
```

DNS is restricted to Docker's embedded resolver only (`dns: [127.0.0.11]` with `extra_hosts` for the syslog target), so there is no external DNS path that could resolve `myservice-blue` to anything.

## What I expected to see

Traefik never dispatches requests to a server that:

1. Is listed as `DOWN` in the API (`serverStatus`)
2. Has a hostname that does not resolve

All requests should go to the only healthy server (`myservice-green:80`).

### What did you see instead?

Despite the API reporting blue as DOWN, occasional requests are dispatched to `http://myservice-blue:80` and return 502 to the client.

Traefik API output confirming blue is DOWN at the time of inspection:

```json
{
  "loadBalancer": {
    "servers": [
      {"url": "http://myservice-blue:80"},
      {"url": "http://myservice-green:80"}
    ],
    "strategy": "wrr",
    "healthCheck": {
      "scheme": "http",
      "mode": "http",
      "path": "/healthz",
      "interval": "5s",
      "timeout": "3s",
      "followRedirects": true
    },
    "passHostHeader": true,
    "responseForwarding": {"flushInterval": "100ms"}
  },
  "status": "enabled",
  "usedBy": ["to-myservice@file"],
  "name": "myservice-lb@file",
  "provider": "file",
  "type": "loadbalancer",
  "serverStatus": {
    "http://myservice-blue:80": "DOWN",
    "http://myservice-green:80": "UP"
  }
}
```

Access log of a 502 dispatched to blue (single representative example out of many; this happens in bursts):

```json
{
  "ClientHost": "...",
  "DownstreamStatus": 502,
  "Duration": 130044066,
  "OriginContentSize": 11,
  "OriginDuration": 1806464,
  "OriginStatus": 502,
  "RequestAddr": "myservice.mydomain.local",
  "RequestMethod": "GET",
  "RequestPath": "/loads/3b09c963-b130-45db-bf7b-308d06e91c80?fields=all",
  "RequestProtocol": "HTTP/1.1",
  "RequestScheme": "https",
  "RetryAttempts": 0,
  "RouterName": "to-myservice@file",
  "ServiceAddr": "myservice-blue:80",
  "ServiceName": "myservice-lb@file",
  "ServiceURL": "http://myservice-blue:80",
  "StartUTC": "2026-05-20T18:00:56.481148303Z",
  "TLSCipher": "TLS_AES_128_GCM_SHA256",
  "TLSVersion": "1.3",
  "entryPointName": "websecure"
}
```

Notable details:

- `OriginDuration: 1806464` ns = **1.8 ms** — far too fast for a real DNS lookup or TCP connection attempt to a non-resolving host. Suggests the dispatch fails internally and is synthesized as a 502 response (rather than propagated as a dial error).
- `RetryAttempts: 0` — the retry middleware does not fire, presumably because the failure is wrapped as an HTTP response (502) rather than a transport-level error.
- This is **not** a transition-window issue. Blue has been completely absent for 24+ hours.

### Pattern observed

The bursty nature is visible at scale. Graylog query over 24h for `Application:traefik AND e_OriginStatus:502 AND e_ServiceAddr:rda-blue\:80` shows clusters of 502s — quiet for many hours, then dozens-to-hundreds within a short window, then quiet again. Multiple endpoints affected, not one client retrying. The dashboard and API consistently report blue as DOWN throughout. Out of thousands of requests per minute, a small fraction lands on blue during a burst.

This rules out:

- DNS resolution loops (verified — DNS does not resolve blue, restricted to embedded Docker resolver)
- Connection pool / stale connection reuse (24h since Traefik restart, blue has never existed in that window)
- Restart races (no Traefik or container restart correlated with the bursts)
- Health-check transition windows (steady state, blue solidly DOWN)

`traefik:latest` (Docker image, pulled within the last 14 days).

```
Version:      3.7.1
Codename:     langres
Go version:   go1.25.10
Built:        2026-05-11T13:15:43Z
OS/Arch:      linux/amd64
```

- Two-host cluster: `docker-prod1`, `docker-prod2`, each running Docker + Traefik
- Azure Application Gateway round-robins traffic across both hosts on `*.mydomain.local`
- File provider with `providers.file.directory=/etc/traefik/dynamic` and `providers.file.watch=true`
- Bug reproduces on both hosts independently — observed on docker-prod1 (default DNS config) and docker-prod2 (after restricting `dns: [127.0.0.11]` with `extra_hosts` for syslog only). Restricting DNS did not eliminate the bug on docker-prod2.

Traefik static configuration (CLI flags):

```yaml
command:
  - --api.dashboard=true
  - --log.level=INFO
  - --accesslog=true
  - --accesslog.format=json
  - --accesslog.bufferingsize=100
  - --api.insecure=false
  - --api=true
  - --providers.docker.network=traefik-network
  - --providers.docker.exposedByDefault=false
  - --entrypoints.web.address=:80
  - --entrypoints.websecure.address=:443
  - --entrypoints.websecure.http.tls=true
  - --providers.file.directory=/etc/traefik/dynamic
  - --providers.file.watch=true
```

DEBUG logs were not captured during a burst (bug is rare and bursty; and traffic is very high, we did not have DEBUG enabled at the time). If requested by maintainers, I can enable DEBUG and wait for the next occurrence to capture `wrr.go:207 > Service selected by WRR:` log lines for the affected service around the time of a 502 dispatch.

## Notes on the WRR code

Reading [`pkg/server/service/loadbalancer/wrr/wrr.go`](https://github.com/traefik/traefik/blob/master/pkg/server/service/loadbalancer/wrr/wrr.go), the `nextServer()` function should skip any handler not present in `b.status` (the map of UP-only servers):

```go
for {
    heap.Push(b, handler)
    if _, ok := b.status[handler.name]; ok {
        if _, ok := b.fenced[handler.name]; !ok {
            break
        }
    }
}
```

For this bug to occur, either:

1. Blue's name is briefly present in `b.status` at the moment of selection (transient health-check flap, or a race between `SetStatus` and `nextServer`), or
2. There is a code path that bypasses `nextServer()` (sticky sessions are not configured, so that path is not relevant here)

DEBUG logs would confirm which. Without them, I can only observe that the API and dashboard report blue as DOWN both before and after each burst, and that the dispatches do occur per the access logs.

## Related

This is similar to but not the same as #11622 (which concerns Kubernetes CRD `TraefikService` weighted services where health check propagation was missing). My case is the simpler file provider `loadBalancer.servers` form, and the health check is correctly being performed and reflected in `serverStatus` — the bug is that the LB selector dispatches to a server its own status map reports as DOWN.

filip2mac:
## Problem

Every dynamic configuration reload rebuilds all load balancers from scratch,
with every server initially marked UP. Health checks are then scheduled to
run after the first full `interval` elapses. This opens a window — up to
`interval` long — during which DOWN servers receive real traffic and produce
5xx errors visible to clients.

The window is especially pronounced when:
- The health check interval is long (e.g. the default 10s or user-configured 5s)
- Config reloads are frequent (e.g. `providers.file.watch=true` + Docker provider
  triggering reloads on every container event)

The reporter in #13218 observed occasional 502s to a server that had been DOWN
for 24+ hours. The API reported the server as DOWN both before and after the
burst because the UP window (during which traffic was dispatched) was short
enough to be missed by manual API checks.

## Fix

Replace `time.NewTicker(interval)` with `time.NewTimer(0)` so the first health
check fires immediately after `Launch`. After each check cycle, `timer.Reset(interval)`
schedules the next check. This eliminates the initial warmup window.

The same fix is applied to both HTTP and TCP health checkers, which share the
same structural pattern.

## Test

`TestServiceHealthChecker_ImmediateFirstCheck` proves the fix: it starts a
server that always returns 503, configures a 1-minute interval, and asserts
the server is marked DOWN within 5 seconds. Without this fix the test would
time out.

All existing `TestServiceHealthChecker_Launch` cases continue to pass because
they wait for health-check events via a channel; an earlier first check only
makes the tests faster, not incorrect.

Closes #13218


kevinpollet:
Hey @filip2mac,

PR #13226, opened 2026-05-22
Fix X-Forwarded-For append handling

### LINKED ISSUE #13185: Optimize X-Forwarded-For append handling and centralize related logic
### Welcome!

In PR #12374, a new option `NotAppendXForwardedFor` was introduced to skip appending the client IP to X-Forwarded-For (XFF). I noticed that it stores a `notAppendXFFKey` of type `string` in the request context.

After reading through the proxy request flow, I have the following thoughts:

1. As a context key, using `struct{}` instead of `string` seems like a better choice, and it may also offer a performance improvement.
2. Why is XFF not handled in the always-executed, unified `forwardedheaders` middleware, just like the other X-Forwarded-* headers? At the moment, there are two separate XFF implementations in `/pkg/proxy`, which feels somewhat repetitive and hard to maintain. I wonder whether this is because `http.ReverseProxy.Director` was used originally, and it already includes built-in XFF handling, which led to the current design.

In addition, there was already a `NotAppendXForwardedFor` option in the global configuration section from a previous PR, but it does not actually have any effect. The only effective option is the `NotAppendXForwardedFor` under the `ForwardedHeaders` configuration section.

choyri:
### What does this PR do?

Centralizes `X-Forwarded-For` append handling in the `forwardedheaders` middleware.

This removes the duplicated XFF append logic from both proxy implementations and lets the existing `forwardedHeaders.notAppendXForwardedFor` option control the behavior in one place. It also removes the ineffective global `notAppendXForwardedFor` configuration option.

### Motivation

Closes #13185.

`X-Forwarded-For` was previously handled separately in `/pkg/proxy/fast` and `/pkg/proxy/httputil`, while the other forwarded headers are handled by the shared `forwardedheaders` middleware. This made the behavior harder to maintain and required passing a context flag through the request flow.

The global `notAppendXForwardedFor` option was also exposed in the static configuration, but it was not actually wired to the effective behavior.

### More

- [x] Added/updated tests

### Additional Notes

The effective option remains the forwarded headers configuration option: `entryPoints.<name>.forwardedHeaders.notAppendXForwardedFor`

The removed global option did not affect runtime behavior.


kevinpollet:
Hey @choyri,

Note that this is a breaking change: currently, middlewares only see the XFF header without the latest remote IP appended, so existing middleware configurations will break. 

PR #13228, opened 2026-05-22
kubernetes ingress-nginx: deduplicate status updates per rebuild

MatteoManzoni:
### What does this PR do?

Deduplicates Kubernetes Ingress NGINX status updates during each provider rebuild.

Before this PR, status publishing walked every generated model location and scanned all watched Ingresses for each location. Multiple locations can originate from the same Ingress, so Traefik repeated the same lookup and unchanged-status check several times in one rebuild.

With this change, the provider builds an Ingress index once and updates each distinct Ingress at most once per rebuild. It also skips the status phase when neither `publishService` nor `publishStatusAddress` is configured.

### Motivation

The Kubernetes Ingress NGINX provider can rebuild frequently in large clusters. Repeating the status lookup/update path for every generated location adds avoidable work and can amplify provider reconcile cost without changing status output.

### More

- [x] Added/updated tests

### Additional Notes

Added a regression test with one Ingress containing two paths that asserts only one `ingresses/status` update is attempted.

This PR keeps existing status semantics; it only deduplicates repeated status work for Ingresses that already appear in the built model.

### DIFF
--- a/docs/content/migrate/v3.md
+++ b/docs/content/migrate/v3.md
@@ -80,7 +80,7 @@ kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/downloa
 
 ### Kubernetes CRD Provider
 
-To use the new options of the `retry` middleware with the Kubernetes CRD provider, you need to update your CRDs.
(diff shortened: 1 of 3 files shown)


nmengin:
Hey @MatteoManzoni,

PR #13246, opened 2026-05-27
refactor: use prefix trie for IP authorization checks

tjps:
### What does this PR do?

Switch the IP set checker from a list that is O(n)  in number of entries to a prefix trie that is O(k) in number of bytes of the IP (4 or 16), effectively constant.

### Motivation

For very large IP allowlists or blocklists, the lookup time per request can in practice grow to larger than the request servicing time. This change caps the lookup time in the IP checker.

### More

- Added a new unit test demonstrating a large number of IPs in the set, and showing impact on lookup time.

### Additional Notes

There are several Go prefix trie implementations, I chose the one that looked best (code quality, then performance). I am open to suggestions for alternatives.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -53,6 +53,7 @@ require (
 	github.com/moby/moby/api v1.54.1
 	github.com/moby/moby/client v0.4.0
 	github.com/patrickmn/go-cache v2.1.0+incompatible
+	github.com/phemmer/go-iptrie v0.0.0-20240326174613-ba542f5282c9 // No tag on the repo.
 	github.com/pires/go-proxyproto v0.8.1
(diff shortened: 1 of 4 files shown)


kevinpollet:
Hello @tjps and thanks for opening this,

We've already accepted a pull request on the same topic: https://github.com/traefik/traefik/pull/13077.
Feel free to review it or leave a comment if you think it can be improved.

I'm therefore closing this pull request as a duplicate.

tjps:
Hi @kevinpollet - I didn't think to check for this already being in the PR queue, thanks for pointing out.

I'd still like to advocate for this solution:
- one data structure vs. two (prefix trie vs. hash map and list of ranges that are binary searched)
- fewer lines of code
- simpler check path

Issue #13235, opened 2026-05-26
WASM plugin memory leak on dynamic config reload persists in v3.4.4 — per-router WASM middleware instances are not freed (despite #11119/#11151)


Yaminyam:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any. (Found related: #11119 (closed/fixed), #11436 (open), #10959 (draft) — this appears to be a residual case not covered by the #11119 fix.)
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

We run a WASM middleware plugin (http-wasm, Rust → `wasm32-wasip1`, ~311 KB release build) applied **per-router**: each router references its own middleware instance because the plugin receives router-specific config via `get_config()` (per-route auth secrets / data). A KV (etcd) provider re-publishes the route table on a periodic reconcile, so the dynamic configuration changes frequently.

In our deployment there are ~50–80 routers, each with its own WASM middleware instance, and the provider triggers a config reload roughly every ~8–17 s (~8 reloads/min observed via `traefik_config_reloads_total`).

### What did you expect to see?

Stable memory. On each dynamic-config reload that recreates the per-router WASM middlewares, the **previous** instances should be released (their wazero runtime / compiled module closed and GC-eligible), per the fix in #11119 / #11151 ("Close wasm middleware to prevent memory leak") and http-wasm-host-go #86.

### What did you see instead?

traefik RSS grows **linearly and unbounded** (~150 MB → 6–22 GB over a few days, ~1.4 GB/h on a busy node) until it is **OOM-killed**, then resets on restart and climbs again (sawtooth). It is **reload-driven, not request-driven**: RSS climbs in step with config reloads, and the `traefik_config_reloads_total` rate jumped from ~0 to ~0.1–0.6/s exactly when the periodic reconcile was enabled; before that, with reloads ≈ 0, RSS was flat for months.

A heap profile (`/debug/pprof/heap`, diff between two snapshots ~23 min apart; RSS 215 MB → 761 MB) shows **~93% of the growth is wazero** — the WASM module is **re-decoded and re-instantiated on every reload and the previous instances are retained**:

```
go tool pprof -top -inuse_space -base t0.prof t1.prof
  +156.9 MB (53%)  github.com/tetratelabs/wazero/internal/wasm.NewMemoryInstance      (<- ModuleInstance.buildMemory)
[...]

Issue #11436, opened 2025-01-10
Memory allocation panic in WASM plugin during normal operation


skosch:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I use Traefik v3 with [this Traefik OIDC auth plugin](https://github.com/xzzpig/traefik-oidc-wasm/blob/88208714b7f5d7ccaf81234f86bdad06359cdd0e/main.go#L107). The plugin uses http-wasm-host-go with the forked [traefik/http-wasm-guest-tinygo](https://github.com/traefik/http-wasm-guest-tinygo).

I opened an issue [here](https://github.com/http-wasm/http-wasm-host-go/issues/92) weeks ago, but have not received a response. I hope it's okay to crosspost here, in case the problem lies with Traefik; otherwise please feel free to close and redirect me elsewhere.

This is unrelated to the memory leak problems discussed in #11116. I experienced those as well, but more recent versions of Traefik have been fine.

### What did you see instead?

Certain requests cause the runtime to crash, causing downtime until I manually restart Traefik. I have not been able to detect what, if anything, the problematic requests have in common.

The plugin calls `GetURI` via [GetString](https://github.com/http-wasm/http-wasm-guest-tinygo/blob/e57ec90ac6a122642b466d20f63db31d0f56da1f/handler/internal/mem/mem.go#L39), which fails when trying to allocate memory for the returned `string(...)`:

```go
func GetString(fn func(ptr uint32, limit imports.BufLimit) (len uint32)) (result string) {
	size := fn(ReadBufPtr, ReadBufLimit)
	if size == 0 {
		return // If nothing was read, return an empty string.
	} else if size <= ReadBufLimit {
		return string(ReadBuf[:size]) // string will copy the buffer.  <-------  THIS FAILS!
	}
	...
}
```
Here's a stacktrace:

```
runtime.throw({0xab3e1, 0x2a})
	/home/runner/sdk/gotip/src/runtime/panic.go:1074 +0x3 fp=0x48c1d8 sp=0x48c1b0 pc=0x14a00003
runtime.sigpanic()
	/home/runner/sdk/gotip/src/runtime/os_wasm.go:26 +0xe fp=0x48c1f8 sp=0x48c1d8 pc=0x121a000e
[...]

PR #11151, opened 2024-10-02
fix: closes wasm middleware to prevent memory leak

ttys3:
### What does this PR do?
Adds a finalizer to close middleware when not in use due to a config change.

### Motivation

Traefik does not Close the middleware when creating a new instance on a configuration change.
When the middleware is marked to be GC, we need to close it so the wasm instance is properly closed.

Fixes https://github.com/traefik/traefik/issues/11119

Issue #11119, opened 2024-09-24
wasm plugin memory leak in Traefik v3


ttys3:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

add a simple wasm traefik middleware plugin to a router. 
run ab benchmark to make some request
you will see memory increase
stop the benchmark and wait
you will see the memory keep leaks, no matter how long you waited

the more concurrent requests make, the more memory leak.

@juliens  @zetaab  could you plz take a look at this?

here's the traefik memory usage grafana graph from our real server,  after enabling wasm based plugin:

![image](https://github.com/user-attachments/assets/21f2a491-0066-4e63-aab5-2f3125cb7280)

----------------------------------------------

steps to reproduce

go version 1.23.1
under Linux amd64 machine

```shell
git clone https://github.com/ttys3/traefik-wasm-plugin-memory-leak-demo.git

cd traefik-wasm-plugin-memory-leak-demo

pushd plugins-local/src/github.com/traefik/plugindemowasm/
# build wasm plugin
./build.sh

# you will see generated plugin.wasm now

# now we go back to traefik-wasm-plugin-memory-leak-demo dir
popd

# download newest traefik
curl -LO https://github.com/traefik/traefik/releases/download/v3.1.4/traefik_v3.1.4_linux_amd64.tar.gz
tar xvf traefik_v3.1.4_linux_amd64.tar.gz

# run the demo backend service, it is a simple http static server
./run-downstream-service.sh

# run traefik
./run.sh
```

now install ps_mem if you do not have it

```shell
sudo pip install ps_mem
```

now we open a new terminal to watch the memory

```shell
./watch-mem.sh
```

we 'll see that the init memory usage is: 

![image](https://github.com/user-attachments/assets/0abfede5-c5ab-457e-8c12-fea266431139)

go we run `./bench.sh` in the repo to make some requests (you should install `ab` - the Apache HTTP server benchmarking tool, if you os does not installed it, or other bench tools also OK)

> adjust params of `ab` if you system does not have too much free memory, in case OOM

```shell
./bench.sh

Benchmarking localhost (be patient)
Completed 100 requests
Completed 200 requests
Completed 300 requests
Completed 400 requests
Completed 500 requests
Completed 600 requests
Completed 700 requests
[...]

PR #13253, opened 2026-05-29
plugins: cache wasm middleware instances across reloads to avoid runtime leaks

### LINKED ISSUE #86: Add PassHostHeader in frontends configuration
We need to add PassHostHeader option in frontend configuration.

Yaminyam:
### What does this PR do?

Memoizes the built http-wasm `Middleware` (and the wazero runtime and compiled guest module behind it) per middleware name + guest config hash inside `wasmMiddlewareBuilder`, so that re-publishing the same middleware on a dynamic configuration reload reuses the existing wazero runtime instead of allocating a fresh one.

Lifetime is reference-counted:

* On every `buildMiddleware` call the cached instance's refcount is incremented and the returned handler is wrapped in a tiny `instanceHandler` struct that carries a finalizer releasing that reference on GC.
* When the guest config for a middleware name changes, the previous entry is marked **evicted**. Its underlying runtime is closed deterministically once its last outstanding handler is released — no longer relying on the previous best-effort finalizer to keep up under churn.
* If no entry was cached for that name, behavior is unchanged: a new runtime is built and cached.

The existing `runtime.SetFinalizer` on the returned handler that was the only cleanup mechanism (added in https://github.com/traefik/traefik/pull/11151) is removed: with deterministic eviction-close that finalizer was both redundant and the reason why frequent reloads accumulated runtimes (one new `wazero.Runtime` per build, only ever reaped on a successful future GC of the entire handler graph).

The validation runtime built inside `newWasmMiddlewareBuilder` is now also explicitly `Close`d (it was previously leaked — minor, but tidies up the same code path).

### Motivation

Several reports — most recently #13235 (see also #11119, #11436, #10959) — describe the wasm plugin's memory footprint growing without bound when a dynamic provider re-publishes middlewares frequently.

I traced an in-the-wild OOM (traefik 3.4.4, RSS ~150 MB → > 20 GB over a few days, ending in OOM kill) to this code path. The provider in question (etcd) re-publishes every route on a ~30 s reconcile loop, which means the upstream router builder rebuilds the wasm middleware chain on every reload. Two pprof heaps taken 23 minutes apart showed ~93 % of the growth in wazero — concentrated in `NewMemoryInstance` and `decodeCode` — exactly the symptom you'd expect if a fresh `wazero.Runtime` is created on every rebuild and the previous one is held alive long enough that the finalizer never gets to run. `go_goroutines` was flat, ruling out goroutine leaks.

The root cause is the per-rebuild `wazero.NewRuntimeWithConfig` + `runtime.SetFinalizer` pattern in `buildMiddleware`: under many wasm-using routers and frequent reloads, finalizers cannot keep up and the runtimes accumulate. Notably http-wasm-host-go is already at the latest v0.7.0 (which includes the upstream finalizer fix #86), so the problem is not in the host runtime — it's the lifetime model in this builder.

Memoizing the built `Middleware` per name + guest-config hash means the steady-state cost of "same middleware re-published" is one wazero runtime per middleware (not one per reload), and the rare actual-config-change rebuilds clean up the previous runtime deterministically. This should also remove the residual leak path that #11151's finalizer was working around.

### Tests

Added four tests covering the new behavior:

* `TestWasmMiddlewareInstanceReuse` — unchanged guest config reuses the cached instance and its wazero runtime.
* `TestWasmMiddlewareConfigChangeRebuilds` — changed guest config produces a fresh instance (single cache entry per middleware name).
* `TestWasmMiddlewareConfigChangeEvictsOldInstance` — when the config changes, the previous instance is marked evicted (deterministic; no GC dependency).
* `TestWasmMiddlewareNoRuntimeLeakOnReload` — reproduces #13235's reload churn: 50 rebuilds of the same middleware create exactly **one** cached instance.

All pre-existing tests in `pkg/plugins/` still pass; `TestSettingsWithoutSocket` was updated only to construct the builder with the new (now mandatory) `instances` cache field.

### More

- [x] Added/updated tests
- [ ] Added/updated documentation — none needed, behavior change is invisible to plugin authors and to users (no new config knobs).

### Additional Notes

* I'm happy to iterate on the design — open to alternative shapes (e.g., a hard `Close` on cache eviction without the refcount + finalizer hybrid, or wiring this through `pluginMiddleware` rather than at the builder layer). The refcount-and-handler-wrapper design here was chosen because it keeps the `pluginMiddleware` / `WasmMiddleware` API surface unchanged and never closes a runtime that might still serve an in-flight request — at the cost of one extra small object allocation per built handler.
* Targeting `v3.6` per CONTRIBUTING guidance for bug fixes; happy to retarget if you'd rather take this only on `master`.
* Related issue: #13235.

### DIFF
--- a/pkg/plugins/middlewarewasm.go
+++ b/pkg/plugins/middlewarewasm.go
@@ -2,6 +2,7 @@ package plugins
 
 import (
 	"context"
+	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


rtribotte:
Hey @Yaminyam,

AndrewGearhart:
@rtribotte - We're seeing this pop-up as an issue for us as well... we're looking for what sort of timeline might be upcoming for this analysis and fix?

rtribotte:
Hi @AndrewGearhart,

We're focused on other priorities right now though, we'll keep you updated as soon as we can move forward!

Yaminyam:
A long time has passed. When might this issue be resolved?
@rtribotte

PR #13260, opened 2026-06-01
Clear Ssl-Client-* headers when no client certificate is present

gndz07:
### What does this PR do?

As a hardening improvement, this PR clears `Ssl-Client-Cert`, `Ssl-Client-Subject-Dn`, and `Ssl-Client-Issuer-Dn` headers on `AuthTLSPassCertificateToUpstream` middleware for ingress-nginx provider when no client certificate is provided, mirroring the behaviour of the mTLS branch which already overwrites all four headers from the verified peer certificate.

### Motivation

Ensures client-supplied values for those headers never reach the upstream when no certificate is present.

### More

- [x] Added/updated tests


juliens (approved):

rtribotte (approved):
LGTM

PR #13263, opened 2026-06-01
docs: add common middleware stack example

mfolofy:
Real-world middleware chain combining basicAuth + rateLimit + CORS headers. This is the most common pattern for API gateways and wasn't documented as a composite example.


nmengin:
Hey @mfolofy,

Thanks for this.

From what I understand, the example is more of an advanced use case.
Could you move it into the different ["Expose" guide](https://doc.traefik.io/traefik/expose/overview/) instead of the reference documentation?

PR #13264, opened 2026-06-01
add source-IP sticky routing as a TCP load balancer option

### LINKED ISSUE #10265: Sticky IngressRouteTCP based on source IP
### Welcome!

Hello, 

I would like the ability of IngressRouteTCP to allow sticky routing based on source IP (compatible with PROXY protocol), this affinity ensures the client is served on the same pod.

In our use case, the client established multiple connections (mTLS, TLS handshake) to 3rd party services. Since not all services support Session Ticket/TLS PSK yet, it's costly and slow to restart connections (Require full handshake). When you have undred pods, it's just loop to warm each pods, previous warmed become cold (connection are closed) so it's very inneficient. We would like to reroute the same user to this pod.

KaanSimsek:
### What does this PR do?

Adds an ip-hash load balancing strategy for TCP services. When enabled, connections from the same source IP are consistently routed to the same backend server using Rendezvous Hashing (HRW).
A new strategy field is added to TCPServersLoadBalancer (and exposed via the IngressRouteTCP CRD) with two options:
wrr — existing Weighted Round Robin behavior (default, no breaking change)
ip-hash — routes by source IP hash

### Motivation

Closes #10265.
When clients establish multiple TCP connections to Traefik, each connection can land on a different backend pod. This breaks TLS session reuse — every new connection requires a full TLS handshake, which is expensive for workloads with many short-lived connections. Routing connections from the same source IP to the same backend allows the backend to reuse TLS sessions across connections.
Cookies cannot be used at the TCP layer, so IP-based consistent hashing is the natural equivalent of HTTP sticky sessions.

### More

- [x] Added/updated tests

### Additional Notes
* Rendezvous Hashing is used instead of simple modulo hashing for stability: when a server goes down, only the connections that were mapped to that server are redistributed. All other clients continue routing to their original server.
* A LoadBalancer interface has been introduced in pkg/tcp/handler.go to allow the strategy switch in BuildTCP to hold either LB implementation as an interface variable.


juliens:
Hey @KaanSimsek ,

PR #13255, opened 2026-05-29
fix(k8s): reduce provider churn from EndpointSlice and node updates

jaygridley:
## Summary

Fixes the Kubernetes provider CPU regression reported in https://github.com/traefik/traefik/issues/13011 by reducing unnecessary config rebuild churn and hot-path cost during informer updates.

## What changed

- Added a fast internal configuration hashing implementation in `pkg/config/dynamic/confighash` and switched Kubernetes providers, collector, and nomad provider hashing call sites to it.
- Added EndpointSlice service-name index support in `pkg/provider/kubernetes/k8s/indexers.go` and integrated it in provider clients to avoid repeated selector scans.
- Improved Kubernetes event filtering in `pkg/provider/kubernetes/k8s/event_handler.go`:
  - Added `nodeChanged()` to ignore non-load-bearing Node updates.
  - Expanded EndpointSlice change detection to only react to data relevant for routing/service resolution.
- Added/extended tests for hash determinism/safety and event change detection behavior.

## Benchmark evidence

Source files:
[summary-churn.txt](https://github.com/user-attachments/files/28390790/summary-churn.txt)
[summary-no-churn.txt](https://github.com/user-attachments/files/28390746/summary-no-churn.txt)

Benchmark files: https://github.com/jaygridley/traefik/tree/bench-issue-13011/bench/issue-13011

### Churn scenario (node + EndpointSlice churn)

- Mean CPU: `209.6 mCPU` -> `119.6 mCPU` (**-43%**)
- P95 CPU: `431 mCPU` -> `246 mCPU`
- Max CPU: `477 mCPU` -> `268 mCPU`

### No-churn scenario

- Mean CPU: `1.5 mCPU` -> `1.4 mCPU` (**-7%**)

## Testing

```
go test ./pkg/config/dynamic/confighash ./pkg/provider/kubernetes/... ./pkg/provider/nomad ./pkg/collector           
ok      github.com/traefik/traefik/v3/pkg/config/dynamic/confighash     0.606s
ok      github.com/traefik/traefik/v3/pkg/provider/kubernetes/crd       5.496s
?       github.com/traefik/traefik/v3/pkg/provider/kubernetes/crd/generated/applyconfiguration  [no test files]
?       github.com/traefik/traefik/v3/pkg/provider/kubernetes/crd/generated/applyconfiguration/internal [no test files]
[...]

Issue #13011, opened 2026-04-20
High CPU regression since v3.1 — EndpointSlice migration + Nodes heartbeat on k8s providers


jaygridley:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

Reopening the investigation from #11256 (closed as stale after being misdiagnosed as a Compress middleware issue — multiple users confirmed it reproduces without Compress: @matuszeman, @devopsbarbarian, @fangker).

Bisected the CPU regression on a production EKS cluster:

1. Started with Traefik v3.0.4 (CPU baseline acceptable).
2. Upgraded version by version through v3.1 → v3.6 (CPU regressed at v3.1, stayed elevated through v3.6).
3. On a separate local `v3.1` branch, reverted the single commit `a8a92eb2a` "Migrate to EndpointSlices API" (Jun 2024, shipped in v3.1.0). CPU neared to the pre-v3.1 baseline. This is the main regression point.
5. Captured `pprof` CPU profiles during reconfig bursts (Deployment rollouts) to identify remaining hotspots on top of the EndpointSlice migration.
6. Applied targeted mitigations on a fork and measured: combined, they bring CPU back to pre-v3.1 levels and substantially cut spike amplitude.
This is not a hyperscale cluster, yet the regression reproduces clearly — which is probably why prior reproduction attempts against a minimal setup didn't trigger it.

### What did you see instead?

**Observed:** CPU steady-state is significantly increased (post 04/13 below) and spikes severely during cluster events (rollouts, scale changes, node autoscaler activity). The pattern is a dense series of short high-amplitude spikes, not a single peak:

<img width="472" height="244" alt="Image" src="https://github.com/user-attachments/assets/b61a530f-acaa-498e-9826-516f916300af" />
* CPU utilization. Traefik upgraded to v3.6 after 04/13.

**Root causes identified:**

**(1) O(N) per-call label-selector scan in `GetEndpointSlicesForService`.** Every K8s provider does:

```go
c.factoriesKube[...].Discovery().V1().EndpointSlices().Lister().
    EndpointSlices(namespace).List(serviceSelector)
```

[...]

PR #13266, opened 2026-06-02
kubernetes: strip managedFields in informers

ishioni:
### What does this PR do?

Registers a `SetTransform` on every Kubernetes informer factory across the providers that drops `metadata.managedFields` from objects as they enter the informer cache, via a shared `k8s.StripManagedFields` transform.

### Motivation

Part of the CPU regression investigation in #13011. Traefik never reads `managedFields`, yet they ride along on every watched object (Secrets, Ingresses, Services, EndpointSlices, Nodes, …) and inflate cached object size — which in turn raises the cost of the `DeepCopy` and `DeepEqual` work Traefik performs on every event. Stripping them at ingest cuts that cost for every resource kind.

### More

- [x] Added/updated tests

### Additional Notes

One of a few independent follow-ups addressing the tertiary hotspots listed in #13011; complementary to and non-overlapping with #13255. Targeting v3.7 since the regression is present there (issue labeled `kind/bug`) — happy to retarget to `master` if preferred.

### DIFF
--- a/pkg/provider/kubernetes/crd/client.go
+++ b/pkg/provider/kubernetes/crd/client.go
@@ -177,7 +177,7 @@ func (c *clientWrapper) WatchAll(namespaces []string, stopCh <-chan struct{}) (<
 	}
 
 	for _, ns := range namespaces {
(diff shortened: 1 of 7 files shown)


kevinpollet:
Hey @ishioni,

Thanks for this contribution!

ishioni:

kevinpollet:
Closed in favor of https://github.com/traefik/traefik/pull/13396

PR #13267, opened 2026-06-02
kubernetes ingress: cache parsed router annotations

ishioni:
### What does this PR do?

Caches the parsed router-annotation config (`parseRouterConfig`) per Ingress, keyed by `(namespace, name, resourceVersion)`, pruning entries not seen during a rebuild. Cached entries are deep-copied on read so each rebuild gets independent values.

### Motivation

Part of #13011. The pprof in that issue attributes ~19% of rebuild CPU to `ingress.parseRouterConfig` — `paerser`'s reflection-based `label.Decode` runs for every Ingress on every rebuild even when annotations are unchanged. Keying on `resourceVersion` lets unchanged Ingresses skip the decode entirely on subsequent rebuilds.

### More

- [x] Added/updated tests

### Additional Notes

Independent follow-up to a tertiary hotspot in #13011, complementary to #13255. The deep copy on read preserves the pre-cache semantics (each rebuild gets fresh slices/pointers, which are aliased into the emitted dynamic config). Targeting v3.7 (regression present there, labeled `kind/bug`).

### DIFF
--- a/pkg/provider/kubernetes/ingress/annotations.go
+++ b/pkg/provider/kubernetes/ingress/annotations.go
@@ -2,6 +2,7 @@ package ingress
 
 import (
 	"regexp"
+	"slices"
 	"strings"
 
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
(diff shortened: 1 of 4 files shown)


kevinpollet:
Hey @ishioni,

Thanks for this contribution!

ishioni:

PR #13268, opened 2026-06-02
kubernetes: deduplicate backend service loads per rebuild

ishioni:
### What does this PR do?

Deduplicates backend service resolution within a single provider rebuild. Many routes reference the same backend; this caches the resolved result keyed on the inputs that determine it (namespace, service name, port — plus `service-upstream`/`default-backend` for ingress-nginx), so the Service lookup, EndpointSlice lookup, annotation parsing, and endpoint iteration run once per unique backend instead of once per route. Applied to the Ingress and ingress-nginx providers.

### Motivation

Part of #13011. The pprof shows `loadService`/`loadServers` (and the equivalent `getBackendEndpoints` in ingress-nginx) called per path rather than per unique Service, repeating identical work across routes that share a backend. A per-rebuild cache collapses these.

### More

- [x] Added/updated tests

### Additional Notes

Independent follow-up to a tertiary hotspot in #13011, complementary to #13255. The cache lives for a single rebuild (created fresh each time) and returns cloned/deep-copied values so callers can't corrupt it. This touches two providers with the same optimization — happy to split into two PRs if maintainers prefer. Targeting v3.7.

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/build.go
+++ b/pkg/provider/kubernetes/ingress-nginx/build.go
@@ -48,6 +48,11 @@ type certBlocks struct {
 	key  []byte
 }
 
+type backendEndpointsCacheEntry struct {
+	endpoints []endpoint
+	err       error
+}
+
 // build reads all Ingress resources visible to the client and produces a model.
 //
 //nolint:funlen // multi-phase ingress processing kept inline for readability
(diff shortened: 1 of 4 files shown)


kevinpollet:
Hey @ishioni,

Thanks for this contribution!

ishioni:

PR #13269, opened 2026-06-02
kubernetes: index EndpointSlices by service name

ishioni:
### What does this PR do?

Adds an informer indexer on `discoveryv1.LabelServiceName` and replaces the per-call `EndpointSlices(ns).List(serviceSelector)` label-selector scan with an indexed `GetIndexer().ByIndex(...)` lookup, across the ingress, crd, gateway, and ingress-nginx providers.

### Motivation

The primary root-cause fix for #13011: the EndpointSlice migration in v3.1 (`a8a92eb2a`) replaced an O(1) Endpoints lookup with an O(N) scan of every EndpointSlice in the namespace on every `loadService` call, which dominates CPU during reconfig bursts.

### More

- [x] Added/updated tests

### Additional Notes

Heads-up: this overlaps #13255, which implements the same indexer. We're mainly opening it so there's a concrete diff to compare — ours benchmarks a little lighter in our runs, but we're entirely happy to defer to #13255, fold our version in, or close this, whichever is least work for review. Targeting v3.7 (regression present there).

### DIFF
--- a/pkg/provider/kubernetes/crd/client.go
+++ b/pkg/provider/kubernetes/crd/client.go
@@ -21,7 +21,6 @@ import (
 	kerror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
 	kinformers "k8s.io/client-go/informers"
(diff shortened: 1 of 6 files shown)


kevinpollet:
Hey @ishioni,

Thanks for this contribution!

ishioni:

kevinpollet:
Closed in favor of #13395

PR #13270, opened 2026-06-02
kubernetes: detect EndpointSlice condition changes

ishioni:
### What does this PR do?

Adds content-aware change detection for EndpointSlice update events, so a rebuild only fires when fields Traefik consumes change (ports, endpoint addresses, and the Ready/Serving/Terminating tri-state) — ignoring updates that only bump metadata/resourceVersion.

### Motivation

Part of #13011. Without this, any EndpointSlice update (annotation ticks from co-resident controllers, condition transitions with no address change) triggers a full O(Ingresses × paths) rebuild. The condition comparison treats nil/true/false as distinct, because Serving/Terminating drive server filtering + `Fenced` and Ready drives TCP/UDP/Gateway — dropping a condition-only transition would be a correctness bug.

### More

- [x] Added/updated tests

### Additional Notes

Overlaps #13255, which also expands EndpointSlice change detection. Opening for comparison; happy to defer to #13255, fold in, or close — whichever you prefer. Targeting v3.7.

### DIFF
--- a/pkg/provider/kubernetes/k8s/event_handler.go
+++ b/pkg/provider/kubernetes/k8s/event_handler.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	discoveryv1 "k8s.io/api/discovery/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 )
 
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hey @ishioni,

Thanks for this contribution!

ishioni:

kevinpollet:
Closed in favor of #13405

PR #13271, opened 2026-06-02
kubernetes: ignore Node updates that do not affect addresses

ishioni:
### What does this PR do?

Adds `nodeChanged()` so Node updates only trigger a rebuild when the set of `InternalIP`/`ExternalIP` addresses changes, ignoring the ~10s kubelet status heartbeat (conditions/images/capacity/allocatable) that Traefik never reads.

### Motivation

Part of #13011. With cluster-scope resources enabled (default, needed for NodePortLB), kubelet rewrites `Node.Status` on a heartbeat; each rewrite currently triggers a full rebuild even though Traefik only reads `Node.Status.Addresses` of type `InternalIP`/`ExternalIP`.

### More

- [x] Added/updated tests

### Additional Notes

Overlaps #13255, which also adds `nodeChanged`. Opening for comparison; fully open to deferring to #13255, folding in, or closing. Targeting v3.7.

### DIFF
--- a/pkg/provider/kubernetes/k8s/event_handler.go
+++ b/pkg/provider/kubernetes/k8s/event_handler.go
@@ -1,6 +1,9 @@
 package k8s
 
 import (
+	"maps"
+
+	corev1 "k8s.io/api/core/v1"
 	discoveryv1 "k8s.io/api/discovery/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hey @ishioni,

Thanks for this contribution!

ishioni:
@kevinpollet so these would not hit the 3.7 branch? Only a future 3.8?

kevinpollet:
Hello @ishioni,

> so these would not hit the 3.7 branch? Only a future 3.8?

This will be released in a future v3.8. Since those changes might have side effects, we prefer to ship them in a later version so there's an RC period that lets users test them.

ishioni:

PR #13273, opened 2026-06-02
Fix redis write timeout option configuration

bzyy1024:
### What does this PR do?

Fixes a copy-paste bug in Redis rate limiter configuration where the 
WriteTimeout condition incorrectly checks `*config.Redis.ReadTimeout` 
instead of `*config.Redis.WriteTimeout`.

When `ReadTimeout` is set to `0` (disabled) and `WriteTimeout` is set to 
a positive value (e.g. `5s`), the configured WriteTimeout is silently 
ignored and set to `-1` (disabled).

### Motivation

Found while reading the code. The ReadTimeout block directly above uses 
the correct variable, indicating this is a copy-paste error.


juliens (approved):

rtribotte (approved):
Thanks!

PR #13285, opened 2026-06-03
feat: write status conditions to Traefik CRDs after config load

Alex-Waring:
### What does this PR do?

This PR addresses https://github.com/traefik/traefik/issues/10015.

This PR adds status fields to all CRDs provided by the Traefik CRD provider. On configuration reload, traefik will update the status of every resource so it's clear to anyone on the API whether they have been loaded in. This action is non-blocking, if it fails the routes will still be loaded but we will log a warning to indicate the failure.

The helm chart has been updated with addition RBAC required for this.

I've split this out into two commits, the first is the change to the CRDs and the autogenerated changes, the second id the functional change.

### Motivation

See https://github.com/traefik/traefik/issues/10015. In essence it is kube standard to update the status field with details on the status of that resource, this PR implements that, improving visibility into deployed routes.

### More

- [x] Added/updated tests
- [ ] Added/updated documentation - I don't believe there is any relevant documentation to update, but please let me know if this is wrong.

### Additional Notes

🤖 Claude has helped me understand the code base, and write the tests. I wrote the non test code.


nmengin:
Hey @Alex-Waring,

Alex-Waring:
@nmengin @kevinpollet any updates on this?

nmengin:
Hey @Alex-Waring ,

Alex-Waring:
Thanks for the update @nmengin, I only nudge as this is fairly small and self contained, and in my opinion is just a value add to a current feature you have.

Alex-Waring:
Hi @nmengin, any update please?

PR #13316, opened 2026-06-06
Do not reuse the consumed request body for mirrors when mirrorBody is false

### LINKED ISSUE #13292: Mirroring with `mirrorBody: false` fails (502 "invalid Read on closed Body") for any request with a body
# Mirroring with `mirrorBody: false` fails (502) for any request that carries a body

### What did you do?

When a `mirroring` service is configured with `mirrorBody: false`, every mirrored request that has a request body (POST/PUT/PATCH with a non-empty body) fails on the mirror leg with:

```
http: invalid Read on closed Body
```

Requests **without** a body (GET/HEAD, or even an empty-body POST) are mirrored correctly. So the trigger is the *presence of a body*, not the HTTP method.

`mirrorBody: false` is documented as "the request body should not be mirrored", i.e. the mirror should receive a body-less copy of the request. Instead the mirrored request shares the *same, already-consumed* body reader as the main request.

Looking at `pkg/server/service/loadbalancer/mirror/mirror.go`: with `mirrorBody: false` the body is never buffered, so `clone()` returns `rr.req.Clone(ctx)` which shallow-copies the `Body` reader and keeps `Content-Length`. The main handler runs first, reads and closes that shared body, and by the time the mirror goroutine clones the request again, reading the body yields `invalid Read on closed Body` → 502 on the mirror. (Confirmed the same code path exists on `master`.)

#### Minimal reproduction

`docker-compose.yml`:

```yaml
services:
  traefik:
    image: traefik:v3.6.13
    command:
      - --log.level=DEBUG
      - --entryPoints.web.address=:80
      - --providers.file.filename=/etc/traefik/dynamic.yml
      - --api.insecure=true
    ports: ["8000:80"]
    volumes: ["./dynamic.yml:/etc/traefik/dynamic.yml:ro"]
  backend:
    image: traefik/whoami
    command: ["--port=80", "--name=MAIN-BACKEND"]
  mirrortarget:
    image: traefik/whoami
    command: ["--port=80", "--name=MIRROR-TARGET"]
```

`dynamic.yml`:

```yaml
http:
  routers:
    test:
      rule: PathPrefix(`/`)
      entryPoints: [web]
      service: mirror-svc
  services:
    mirror-svc:
      mirroring:
        service: main
        mirrorBody: false      # <-- the trigger
        mirrors:
          - name: mirrortarget
            percent: 100
    main:
      loadBalancer:
        servers: [{ url: "http://backend:80" }]
    mirrortarget:
      loadBalancer:
        servers: [{ url: "http://mirrortarget:80" }]
```

Then:

```bash
docker compose up -d

curl -s -XPOST -d 'hello=world' http://localhost:8000/   # mirror -> 502 in logs
curl -s              http://localhost:8000/               # mirror -> OK
curl -s -XPOST -H 'Content-Length: 0' http://localhost:8000/  # mirror -> OK (empty body)
```

Observed (per request type):

| Request              | Client status | Mirror leg |
|----------------------|---------------|------------|
| GET (no body)        | 200           | OK         |
| POST with body       | 200           | **502**    |
| POST empty body      | 200           | OK         |
| PUT with body        | 200           | **502**    |
| PATCH with body      | 200           | **502**    |
| DELETE no body       | 200           | OK         |

With `mirrorBody: false`, the mirror should receive a body-less copy of the request and succeed, exactly as it does for GET. No `invalid Read on closed Body` errors.

### What did you see instead?

The mirror leg returns 502 for every body-carrying request:

```
DBG github.com/traefik/traefik/v3/pkg/proxy/httputil/proxy.go:121 > 502 Bad Gateway error="readfrom tcp 172.19.0.4:35592->172.19.0.3:80: http: invalid Read on closed Body"
```

(The client itself is unaffected — the main service still responds 200 — but the mirror leg always errors, which pollutes error metrics/traces.)

Setting `mirrorBody: true` makes the same requests succeed (the body is buffered and each clone gets its own reader), which is the current workaround.

### Output of `traefik version`

```
Version:      3.6.13
Codename:     ramequin
Go version:   go1.25.8
OS/Arch:      linux/arm64
```

Also reproduced by reading the source on `master` — the `mirrorBody: false` path is unchanged.

Traefik 3.6.13, file provider, see reproduction above. Originally observed in production behind Kubernetes `TraefikService` mirroring (audit-log mirror) where ~100% of POST/PUT/PATCH mirror legs returned 502 while GET returned the mirror target's normal response.

JATINKUMAR1008:
### What does this PR do?

Ensures mirrored requests carry no body when `mirrorBody` is set to `false`.

Previously, mirrored requests shared the original body reader and retained its content length. Reading that body after the main handler had consumed and closed it caused the mirror leg to fail with `http: invalid Read on closed Body`.

The mirrored clones now use `http.NoBody`, reset `ContentLength` to zero, clear `TransferEncoding`, and remove the `Content-Length` header.

Regression tests cover both known-length and chunked request bodies, including reads after close and preservation of the main request's body and transfer information.

### Motivation

Fixes #13292.

Requests with a body could produce 502 errors on the mirror leg despite `mirrorBody: false`, polluting error metrics and traces.

### More

- [x] Added/updated tests

### Additional Notes

Validated with the mirroring package tests using `-race`, repository lint, and file validation.

The strengthened known-length test fails without the fix. The chunked test also fails when `TransferEncoding` is not cleared.

### DIFF
--- a/pkg/server/service/loadbalancer/mirror/mirror.go
+++ b/pkg/server/service/loadbalancer/mirror/mirror.go
@@ -91,6 +91,17 @@ func (m *Mirroring) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 			// prepare request, update body from buffer
 			r := rr.clone(req.Context())
 
+			if !m.mirrorBody {
+				// As the body has not been buffered,
+				// the clone shares the original body,
+				// which has already been consumed (and closed) by the main handler.
+				// The mirrored request must neither carry nor advertise a body.
+				r.Body = http.NoBody
+				r.ContentLength = 0
+				r.TransferEncoding = nil
+				r.Header.Del("Content-Length")
+			}
+
 			// In ServeHTTP, we rely on the presence of the accessLog datatable found in the request's context
 			// to know whether we should mutate said datatable (and contribute some fields to the log).
 			// In this instance, we do not want the mirrors mutating (i.e. changing the service name in)
--- a/pkg/server/service/loadbalancer/mirror/mirror_test.go
+++ b/pkg/server/service/loadbalancer/mirror/mirror_test.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
 	"sync/atomic"
 	"testing"
 
@@ -176,13 +177,31 @@ func TestMirroringWithBody(t *testing.T) {
 	assert.Equal(t, numMirrors, int(val))
 }
 
+// serverRequestBody mimics the net/http server request body,
+// whose reads fail with http.ErrBodyReadAfterClose once it has been closed.
+type serverRequestBody struct {
+	reader *bytes.Reader
+	closed bool
+}
+
+func (b *serverRequestBody) Read(p []byte) (int, error) {
+	if b.closed {
+		return 0, http.ErrBodyReadAfterClose
+	}
+	return b.reader.Read(p)
+}
+
+func (b *serverRequestBody) Close() error {
+	b.closed = true
+	return nil
+}
+
 func TestMirroringWithIgnoredBody(t *testing.T) {
 	const numMirrors = 10
 
 	var (
 		countMirror atomic.Int32
 		body        = []byte(`body`)
-		emptyBody   = []byte(``)
 	)
 
 	pool := safe.NewPool(t.Context())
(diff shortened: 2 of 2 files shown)


rtribotte:
Hey @JATINKUMAR1008,

Thanks for this contribution!

amazon7737 (approved):
Thanks 👍

PR #13323, opened 2026-06-08
Skip ingress when auth-secret resolution fails

gndz07:
### What does this PR do?

This PR brings a fix to skip the ingress entirely instead of only skipping the middleware if the configured auth-secret could not be resolved.

### Motivation

To have the same behavior as ingress-nginx.

### More

- [x] Added/updated tests


rtribotte (approved):
LGTM

kevinpollet (approved):

PR #12986, opened 2026-04-15
Fix app-root with query params redirect

LBF38:
### What does this PR do?

This PR fixes the `app-root` redirection when using query params in the request URL.

### Motivation

Closes https://github.com/traefik/traefik/issues/12955

### More

- [x] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

More tests have been added to the related e2e tests. See https://github.com/traefik/ingress-nginx-migration/pull/22.

PR #13330, opened 2026-06-09
Fix nginx.ingress.kubernetes.io/app-root annotation not preserving query parameters

### LINKED ISSUE #13315: [ingress-nginx] #12955 not fixed: Preserve query parameters when using app-root annotation
### Welcome!

### What did you do?

When using the ingress-nginx provider with the `nginx.ingress.kubernetes.io/app-root annotation`, query parameters for requests to the root are not preserved in the redirect. This is a re-report of the issue originally reported in #12955, as the resolution #12986 was incomplete. Specifically, the PR updated the regex to add a new capture group for the query parameters, but did not update the `Replacement` to include the captured query parameters.

For example, take an ingress with annotation `nginx.ingress.kubernetes.io/app-root: /login`

**Expected behavior**
* `https://site.example.com/` redirects to `https://site.example.com/login/`
* `https://site.example.com/?foo=bar` redirects to `https://site.example.com/login/?foo=bar`

### What did you see instead?

**Observed behavior**
* `https://site.example.com/ ` redirects to `https://site.example.com/login` (as expected)
* `https://site.example.com/?foo=bar` redirects to `https://site.example.com/login` (query parameters are dropped)

3.7.3

Using the `kubernetesingressnginx`

_No response_

dfeinblatt:
### What does this PR do?
Ingress Nginx provider bug fix for `nginx.ingress.kubernetes.io/app-root` annotation to include the captured query parameters in the redirect replacement so that requests to the root with query parameters are preserved in the redirect to the app-root path. PR #12986 updated the regex to capture query parameters, but did not update the `Replacement` to include them, causing query parameters to be dropped on redirect.

### Motivation
Fixes #13315 

### More

- [x] Added/updated tests

### Additional Notes
Support for `nginx.ingress.kubernetes.io/app-root` annotation was added in v3.7, so targeting v3.7 rather than v3.6

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/translator.go
+++ b/pkg/provider/kubernetes/ingress-nginx/translator.go
@@ -440,8 +440,8 @@ func (p *Provider) applyMiddlewares(mc *model, loc *location, routerKey string, 
 		name := routerKey + "-app-root"
 		conf.HTTP.Middlewares[name] = &dynamic.Middleware{
 			RedirectRegex: &dynamic.RedirectRegex{
-				Regex:       `^(https?://[^/]+)/(\?.*)?$`,
-				Replacement: "$1" + *loc.AppRoot,
+				Regex:       `^(https?://[^/?]+)/?(\?.*)?$`,
+				Replacement: "$1" + *loc.AppRoot + "$2",
 			},
 		}
 		rt.Middlewares = append(rt.Middlewares, name)
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
@@ -4635,14 +4635,14 @@ func TestLoadIngresses(t *testing.T) {
 					Middlewares: map[string]*dynamic.Middleware{
 						"default-ingress-with-app-root-rule-0-path-0-app-root": {
 							RedirectRegex: &dynamic.RedirectRegex{
-								Regex:       `^(https?://[^/]+)/(\?.*)?$`,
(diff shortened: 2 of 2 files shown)


pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -441,7 +441,7 @@ func (p *Provider) applyMiddlewares(mc *model, loc *location, routerKey string,
 		conf.HTTP.Middlewares[name] = &dynamic.Middleware{
 			RedirectRegex: &dynamic.RedirectRegex{
 				Regex:       `^(https?://[^/]+)/(\?.*)?$`,
```
gndz07: WDYT of updating this regex, to make the slash before the query params optional? The current state won't match a URL without slash before the query params, e.g. `https://site.example.com?foo=bar`.
```suggestion
				Regex:       `^(https?://[^/]+)/?(\?.*)?$`,
```

gndz07 (changes requested):

pkg/provider/kubernetes/ingress-nginx/translator.go
dfeinblatt: Good point, we should also account for the example you provided. To make the slash before the query parameters optional, I also needed to exclude `?` from the host capture group, otherwise the host capture group would consume the query string as well.

gndz07 (approved):
LGTM, thanks 👍

kevinpollet:
Hello @dfeinblatt,

According to https://github.com/traefik/traefik/issues/13315#issuecomment-4742571402, 
I am closing this pull request.

PR #13329, opened 2026-06-09
Fix routers with same host, different tlsoptions on different entryPoint

juliens:
### What does this PR do?

This PR fixes the behaviour with multiple TLSOptions on two different entryPoints.

### Motivation

Fixes #13314 

### More

- [x] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes
Co-authored-by: Romain <rtribotte@users.noreply.github.com>

PR #13333, opened 2026-06-10
Fix snicheck for routers with no hosts

rtribotte:
### What does this PR do?

This PR fixes the SNICheck mechanism for routers with no hosts that are already configured to use the default TLSOptions. It also improves logging for detected conflicts, and mitigate (not fully) possible collisions in naming of the conflicted routers.

### Motivation

Follow up to #13329

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/server/aggregator.go
+++ b/pkg/server/aggregator.go
@@ -150,11 +150,8 @@ func mergeConfiguration(configurations dynamic.Configurations, defaultEntryPoint
 //
 // A router keeps its original name, and its resolved TLS options, for the entryPoints
 // on which it does not conflict. For each entryPoint on which it conflicts, that
-// entryPoint is removed from the router and a dedicated copy is emitted, named
-// "ep-conflicted-name@provider". Prefixing with "ep-conflicted-" keeps the original
-// "name@provider" suffix intact, so provider parsing (Split on "@", index 1) still
-// resolves the original provider, and the prefix keeps the copy human-readable. The
-// copy's TLS options are reset to the default ones.
+// entryPoint is removed from the router and a dedicated copy is emitted,
+// and named following this pattern: "ep-conflicted-name@provider".
 func resolveHTTPTLSOptions(routers map[string]*dynamic.Router) map[string]*dynamic.Router {
 	if len(routers) == 0 {
 		return routers
--- a/pkg/server/aggregator.go
+++ b/pkg/server/aggregator.go
@@ -150,8 +150,8 @@ func mergeConfiguration(configurations dynamic.Configurations, defaultEntryPoint
 //
 // A router keeps its original name, and its resolved TLS options, for the entryPoints
 // on which it does not conflict. For each entryPoint on which it conflicts, that
-// entryPoint is removed from the router and a dedicated copy is emitted,
-// and named following this pattern: "ep-conflicted-name@provider".
+// entryPoint is removed from the router and a dedicated copy is emitted, with its
+// TLSOptions reset to the default one, named following the "ep-conflicted-name@provider" pattern.
 func resolveHTTPTLSOptions(routers map[string]*dynamic.Router) map[string]*dynamic.Router {
 	if len(routers) == 0 {
 		return routers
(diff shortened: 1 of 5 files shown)


pkg/server/aggregator.go
```diff
@@ -235,7 +236,7 @@ func findConflictingRouters(routers map[string]*dynamic.Router) []string {
 
 		// A router without a domain in its rule cannot be matched against an SNI,
 		// so it always falls back to the default TLS options.
-		if len(domains) == 0 {
+		if len(domains) == 0 && len(router.TLS.Options) > 0 && router.TLS.Options != traefiktls.DefaultTLSConfigName {
```
emilevauge: Two remarks on this condition:

- `len(router.TLS.Options) > 0 && router.TLS.Options != traefiktls.DefaultTLSConfigName` re-derives what the first loop of `resolveHTTPTLSOptions` already computed into `ResolvedOptions`. Testing `router.TLS.ResolvedOptions != traefiktls.DefaultTLSConfigName` would keep a single source of truth and cannot drift if the resolution logic evolves.
- The comment above is now stale: it says the router "always falls back to the default TLS options", but with this change the fallback only happens when non-default options are configured.

pkg/server/aggregator.go
```diff
@@ -150,9 +150,8 @@ func mergeConfiguration(configurations dynamic.Configurations, defaultEntryPoint
 //
 // A router keeps its original name, and its resolved TLS options, for the entryPoints
 // on which it does not conflict. For each entryPoint on which it conflicts, that
-// entryPoint is removed from the router and a dedicated copy is emitted, prefixed with
-// the entryPoint name the same way applyModel does (ep-name), with its TLS options reset
-// to the default ones.
+// entryPoint is removed from the router and a dedicated copy is emitted,
+// and named following this pattern: "ep-conflicted-name@provider".
```
emilevauge: Nit: the sentence reads "a dedicated copy is emitted, and named following this pattern" (double "and"), and it lost the mention that the copy gets its TLS options reset to the default ones, which is the actual point of the copy. Suggestion:

```suggestion
// entryPoint is removed from the router and a dedicated copy is emitted, with its
// TLS options reset to the default ones, named following the "ep-conflicted-name@provider" pattern.
```

pkg/server/aggregator.go
```diff
@@ -249,14 +250,18 @@ func findConflictingRouters(routers map[string]*dynamic.Router) []string {
...
 			continue
 		}
 
+		var routersInConflict []string
 		for _, names := range routersByOption {
 			conflicting = append(conflicting, names...)
+			routersInConflict = append(routersInConflict, names...)
 		}
+
+		log.WithoutContext().Errorf("On EntryPoint %q, Host %q is served by multiple routers with different TLS options, default TLSOptions will be applied for the following routers: %v", ep, domain, routersInConflict)
```
emilevauge, line 264: Before #13291 this situation was logged at `Warn` level. The routers still serve traffic (with the default options), so this is a degraded-but-functional state rather than a failure. Is `Error` deliberate? Users alerting on error-level logs will start paging on a misconfiguration that used to be a warning.

pkg/server/aggregator_test.go
```diff
@@ -716,22 +716,32 @@ func Test_resolveHTTPTLSOptions(t *testing.T) {
...
 				"router-a@file": {EntryPoints: []string{"ep-a"}, Rule: "PathPrefix(`/foo`)", TLS: &dynamic.RouterTLSConfig{Options: "optsA"}},
 			},
 			expected: map[string]string{
-				"ep-a-router-a@file": "default",
+				"ep-a-conflicted-router-a@file": "default",
 			},
 			unexpectedRouters: []string{"router-a@file"},
 		},
+		{
+			desc: "no domain in rule, default options: not conflicting, keeps its name",
```
emilevauge: Nit: this covers `Options: ""`, but not an explicit `Options: "default"`, which goes through the other half of the condition. A second case (or a second router in this one) would pin that behavior down.

emilevauge (commented):
Thanks, this addresses the follow-ups from #13329 precisely.

Two points to settle before merging, in my opinion:

1. **Log severity.** The conflict is now logged at `Error` level, while before #13291 it was a `Warn`. The behavior is a functional fallback (the routers still serve traffic with the default options), not a failure, and users monitoring the error level may start alerting on it. If `Error` is the deliberate choice (it is a misconfiguration after all), fine, but it should be a conscious decision.
2. **Migration note.** The `ep-conflicted-name` renaming is visible in the API, the dashboard, metrics labels and access logs for conflicted routers. The docs checkbox is unchecked; a short note under `docs/content/migration/` would save support some tickets.

pkg/server/aggregator.go
rtribotte: It is indeed delibarate.

juliens (approved):

kevinpollet (approved):

Issue #13317, opened 2026-06-06
Missing `v`-prefixed container tags for Traefik v3.7.4 on GHCR


iamspido:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I checked the available Traefik image tags in GitHub Container Registry for the newly published 3.7.4 images.

Registry:

ghcr.io/traefik/traefik

I expected the 3.7.4 images to follow the same tag pattern as the previous v3.7.3 images.

For v3.7.3, GHCR exposed v-prefixed tags, for example:

v3.7.3
3.7.3
v3.7

The same was true for the Windows/Nanoserver variants, for example:

v3.7.3-windowsservercore-ltsc2025
3.7.3-windowsservercore-ltsc2025
v3.7-windowsservercore-ltsc2025

v3.7.3-windowsservercore-ltsc2022
3.7.3-windowsservercore-ltsc2022
v3.7-windowsservercore-ltsc2022

v3.7.3-nanoserver-ltsc2025
3.7.3-nanoserver-ltsc2025
v3.7-nanoserver-ltsc2025

v3.7.3-nanoserver-ltsc2022
3.7.3-nanoserver-ltsc2022
v3.7-nanoserver-ltsc2022

After the 3.7.4 release, I expected the equivalent v-prefixed tags to exist, for example:

v3.7.4
v3.7

and for the variants:

v3.7.4-windowsservercore-ltsc2025
v3.7-windowsservercore-ltsc2025

v3.7.4-windowsservercore-ltsc2022
v3.7-windowsservercore-ltsc2022

v3.7.4-nanoserver-ltsc2025
v3.7-nanoserver-ltsc2025

v3.7.4-nanoserver-ltsc2022
v3.7-nanoserver-ltsc2022

### What did you see instead?

For 3.7.4, GHCR currently shows the following tags for the main image:

3.7.4
3.7
langres
v3
3

The expected v3.7.4 and v3.7 tags are missing from that image entry.

For the Windows/Nanoserver variants, GHCR currently shows tags such as:

3.7.4-windowsservercore-ltsc2025
3.7-windowsservercore-ltsc2025
v3-windowsservercore-ltsc2025
langres-windowsservercore-ltsc2025
3-windowsservercore-ltsc2025

[...]

PR #13343, opened 2026-06-11
Increase timeout for Docker image sync job to 30 minutes

jnoordsij:
### What does this PR do?

Increase the timeout on the Docker image sync (GitHub) job.

### Motivation

Seeing #13317 caused by a recent sync job timing out: https://github.com/traefik/traefik/actions/runs/27047058215/job/79835019788.

When multiple (re-)releases happen on the same day, the current value may not suffice as observed. The result is impartial synchronization, which may cause consumer confusion.

As the risk and impact of increasing this value seem relatively minor, I'd say it would be preferable to err on the side of increasing this value rather than keeping it really tight.

### More

- [ ] Added/updated tests

### DIFF
--- a/.github/workflows/sync-docker-images.yaml
+++ b/.github/workflows/sync-docker-images.yaml
@@ -8,7 +8,7 @@ on:
 jobs:
   sync:
     runs-on: ubuntu-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
     permissions:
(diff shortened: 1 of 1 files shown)


rtribotte (approved):
LGTM

gndz07 (approved):

traefiker:

PR #13347, opened 2026-06-11
Fix request scheme derivation when Gateway API RequestRedirect omits scheme

gndz07:
### What does this PR do?

Fix `RequestRedirect` middleware producing scheme-relative `Location` headers (e.g. `//example.com/baz`) when no scheme is configured in the Gateway API `HTTPRequestRedirectFilter`.

In Go's `net/http`, real server requests always have `req.URL.Scheme = ""` because clients send requests in origin-form (`GET /path HTTP/1.1`), which carries no scheme on the wire ([RFC 7230, section 5.3](https://datatracker.ietf.org/doc/html/rfc7230#section-5.3)).

The fix is to derive the scheme from `req.TLS`, which is the reliable indicator of the transport-layer protocol in a Go HTTP handler.

### Motivation

To be inline with Gateway API spec.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/gatewayapi/redirect/request_redirect.go
+++ b/pkg/middlewares/gatewayapi/redirect/request_redirect.go
@@ -59,8 +59,14 @@ func (r redirect) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	redirectURL := *req.URL
 	redirectURL.Host = req.Host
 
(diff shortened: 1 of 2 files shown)


rtribotte (approved):
LGTM

nmengin (approved):
LGTM

PR #13348, opened 2026-06-11
Fix nondeterministic TLS certificate selection on shared SAN

### LINKED ISSUE #13286: Nondeterministic TLS certificate selection when multiple certs share the same SAN
### Welcome!

### What did you do?

We run Traefik with the default TLS store that loads multiple certificates.
Two certificates overlap on the same exact domain:

- Cert A (wildcard): SANs include `*.app.example.test` and `app.example.test`
- Cert B (exact): SANs include `app.example.test` only

Observed in debug logs that both certs are loaded into the store.

**Repro:** repeat TLS handshakes to `app.example.test` and compare the served certificate over multiple requests.

**What we expected:** deterministic selection (ideally always choose the exact match when both match the same SNI). The exact-only cert should be served for `app.example.test`, and the wildcard cert for `page.app.example.test`.

### What did you see instead?

The certificate served for the SNI `app.example.test` is **nondeterministic**: sometimes the wildcard cert is served, sometimes the exact-only cert.

**Effects:**

- clients intermittently see a different certificate chain for the same hostname
- SNI-based routing and monitoring becomes inconsistent over time

**Hypothesis**: selection becomes nondeterministic when multiple certificates share the same SAN. In `GetBestCertificate`, matched candidates are stored in a map keyed by `certDomain`. If two certs include the same `certDomain`, the last one overwrites the previous entry, and map iteration order is randomized. The final choice is based on sorted keys only, so the chosen cert for identical `certDomain` can vary between runs. See: https://github.com/traefik/traefik/blob/29406d42898547f1ffabd904f66af06c212740cf/pkg/tls/certificate_store.go#L75

**Testing command:**
We can execute `watch 'curl -kv https://app.example.test:30443 --resolve app.example.test:30443:127.0.0.1 2>&1 | grep -i "subject:"''` and wait or restart traefik pods (recommanded). 

Version:      3.7.1
Codename:     langres
Go version:   go1.25.10
Built:        2026-05-11T13:15:43Z
OS/Arch:      linux/amd64

Tested on Kubernetes >= 1.31

Traefik's v40.1.0 chart values:
```yaml
rbac:
  enabled: true
deployment:
  kind: DaemonSet
updateStrategy:
  type: RollingUpdate
  rollingUpdate:
    maxUnavailable: 2
    maxSurge: 0
service:
  enabled: true
  spec:
    type: NodePort
    externalTrafficPolicy: Local
ports:
  web:
    nodePort: 30080
  websecure:
    nodePort: 30443
logs:
  general:
    level: DEBUG
    format: json
  access:
    enabled: true
    format: json
    fields:
      headers:
        defaultMode: drop
        names:
          User-Agent: keep
          Referer: keep

```

**Can be demonstrated with:**
[demo.yaml](https://github.com/user-attachments/files/28554315/demo.yaml)

_No response_

rtribotte:
### What does this PR do?

When two distinct certificates share a SAN matching the requested server name, GetBestCertificate keyed the matched candidates by the single matched SAN instead of the certificate identifier (the full SANs concatenation). Distinct certificates then collided on the same map key, and the randomized map iteration order made the served certificate nondeterministic.

Keying by the certificate identifier restores the documented behavior: candidates are sorted by identifier and the last one is served.

### Motivation

Fixes #13286 
To be in conformance with https://doc.traefik.io/traefik/getting-started/faq/#serving-tls-certificates.

### More

- [x] Added/updated tests

### Additional Notes

Assisted-by: Claude Opus 4.8

### DIFF
--- a/pkg/tls/certificate_store.go
+++ b/pkg/tls/certificate_store.go
@@ -104,7 +104,9 @@ func (c *CertificateStore) GetBestCertificate(clientHello *tls.ClientHelloInfo) 
 		for domains, cert := range c.DynamicCerts.Get().(map[string]*CertificateData) {
 			for certDomain := range strings.SplitSeq(domains, ",") {
 				if matchDomain(serverName, certDomain) {
-					matchedCerts[certDomain] = cert
+					// Key the matched candidates by the certificate identifier (the full SANs concatenation).
+					matchedCerts[domains] = cert
+					break
 				}
 			}
 		}
--- a/pkg/tls/certificate_store.go
+++ b/pkg/tls/certificate_store.go
@@ -3,7 +3,9 @@ package tls
 import (
 	"crypto/tls"
 	"fmt"
+	"maps"
 	"net"
+	"slices"
 	"sort"
 	"strings"
 	"time"
@@ -76,6 +78,7 @@ func (c *CertificateStore) GetBestCertificate(clientHello *tls.ClientHelloInfo) 
 	if c == nil {
 		return nil
 	}
+
 	serverName := strings.ToLower(strings.TrimSpace(clientHello.ServerName))
 	if len(serverName) == 0 {
 		// If no ServerName is provided, Check for local IP address matches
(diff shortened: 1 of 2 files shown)


gndz07 (approved):
LGTM

juliens (approved):

PR #13352, opened 2026-06-12
Webui certificate status improvements

### LINKED ISSUE #12946: v3.7.0-rc.1: all certs are marked with warning in new certificate dashboard
### Welcome!

### What did you do?

All statuses are "Expiring soon". However, that's normal for Lets Encrypt "shortlived" certificates. They are expiring soon by design. No need to warn about that, it makes warnings useless.

I recommend either

- using ARI (ACME Renewal Information) to determine when a cert should be renewed (which would then warrant a warning). Or,
- use the [hardcoded values](https://github.com/traefik/traefik/blob/174e5d81111d8e9fb3d3c81cf6d22f3e33eb4f78/pkg/provider/acme/provider.go#L807). Or,
- at least use a small constant value such as 1-2 days.

### What did you see instead?

see above

3.7.0-rc.1

Running in Docker, ACME provider with LetsEncrypt shortlived profile (160 hours lifetime).

_No response_

jnoordsij:
### What does this PR do?

Recently certificate information was added to the webui (dashboard).
Within the implementation, it was opted to consider certificates with a remaining lifetime less than (static) 30 days as 'expiring soon' / 'warning'. However, depending on your ACME provider and/or settings, this may cause user confusion, as certificates may be intended to be more shortlived.

For example the Let's Encrypt shortlived profile offers 7 day lifetime certificates, where the tlsserver profile has been updated to provide 45 day lifetime certificates, which is the intended future default. See also https://letsencrypt.org/docs/profiles/ and https://letsencrypt.org/2025/12/02/from-90-to-45.

To better accomodate for such usage, this PR aims to switch towards checking for certificates with a remaining lifetime of less than one third of the entire lifetime, rather than comparing to a static value. Although still not perfect, this should be a much more accurate check for a larger group of certificates, while still 'complying' with the previous behavior for 90 day certificates.

Fixes #12946.

### Motivation

Noting a lot of my certificates being ("incorrectly") marked as warning/expiring soon,
which was also reported in #12946.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/api/certificate.go
+++ b/pkg/api/certificate.go
@@ -84,7 +84,7 @@ func buildCertificateRepresentation(cert *x509.Certificate) certificateRepresent
 		SignatureAlgorithm:   cert.SignatureAlgorithm.String(),
 		CertFingerprint:      certFingerprint,
 		PublicKeyFingerprint: pubKeyFingerprint,
-		Status:               getCertificateStatus(cert.NotAfter),
+		Status:               getCertificateStatus(cert.NotBefore, cert.NotAfter),
 	}
 }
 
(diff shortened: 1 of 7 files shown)


jnoordsij:
Note: I'll hold off on this pending potential (ACME) ARI support as discussed in https://github.com/traefik/traefik/issues/12461, as that would allow for using actual renewal time data for display, rather than estimated dates.

PR #13354, opened 2026-06-12
Add disableHTTPEntryPoint option

### LINKED ISSUE #13243: providers.kubernetesIngressNGINX.httpEntrypoint: Allow disabling HTTP entrypoint completely
### Welcome!

### What did you do?

When providers.kubernetesIngressNGINX.httpEntrypoint is left empty, Traefik populates ingress routes to all HTTP entrypoints. The current default is web, which binds ingress routes to the web entrypoint only.

However, there is no way to completely disable the HTTP entrypoint association for NGINX ingress resources. Setting it to empty triggers the "all entrypoints" behavior rather than disabling it.

**Use Case**

We want ingress resources managed by kubernetesIngressNGINX to only be exposed on HTTPS (e.g., websecure) without any HTTP route being created. Currently this isn't achievable without a workaround.

Suggested Solution
One of:
 - Support a sentinel value like "" or "none" to explicitly disable
  - Add a boolean field like disableHTTPEntrypoint: true

### What did you see instead?

```
entryPoint "web" doesn't exist
no valid entryPoint for this router
```

3.7.1

```
providers:
  kubernetesIngressNGINX:
    enabled: true
```

_No response_

sinashahoveisi:
### What does this PR do?
Adds a `disableHTTPEntryPoint` boolean option to the `kubernetesIngressNGINX` provider. When set to `true`, no HTTP routes are created for Ingress resources, enabling HTTPS-only configurations.

### Motivation
Currently, when `httpEntryPoint` is left empty, Traefik binds ingress routes to all HTTP entrypoints. There is no way to completely disable HTTP route creation for NGINX ingress resources. This is needed for HTTPS-only environments (e.g. OpenShift/Kubernetes clusters that want to enforce TLS at the ingress level).

Fixes #13243

### More
- [x] Added/updated tests

### Additional Notes
Changes made:
- Added `DisableHTTPEntryPoint` field to the Provider config struct
- Updated `Init()` to skip HTTP entrypoint assignment when disabled
- Updated `SetEffectiveConfiguration` to skip HTTP entrypoint auto-detection when disabled
- Updated `translator.go` to skip HTTP router creation when disabled


kevinpollet:
Hey @sinashahoveisi,

Thanks for this contribution!

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -218,13 +220,17 @@ func (p *Provider) translate(ctx context.Context, mc *model) *dynamic.Configurat
 				rtTLS.Service = unavailableServiceName
 			}
 
-			conf.HTTP.Routers[routerKey] = rt
+			if !p.DisableHTTPEntryPoint {
+				conf.HTTP.Routers[routerKey] = rt
+			}
 			conf.HTTP.Routers[routerKey+"-tls"] = rtTLS
```
sdelicata: Hi @sinashahoveisi! 👋 Thanks a lot for taking this one.

I'm afraid the guard misses the canary routers below (lines 237-285): `canaryRouter` and `nonCanaryRouter` are built from `rt.EntryPoints` and registered unconditionally. With the option on, `NonTLSEntryPoints` is nil, and an empty `EntryPoints` doesn't mean "no entryPoint" in Traefik, `mergeConfiguration` replaces it with the default ones (see `pkg/server/aggregator.go:53`). So an Ingress with a `canary-by-header`/`canary-by-cookie` annotation keeps a plaintext router on `web`, and gets an extra one on `websecure` too.

WDYT about routing every non-TLS router through a small helper so no site can be missed?

```go
// addNonTLSRouter registers a router on the non-TLS entryPoints,
// unless HTTP routes are disabled.
func (p *Provider) addNonTLSRouter(conf *dynamic.Configuration, key string, rt *dynamic.Router) bool {
	if p.DisableHTTPEntryPoint {
		return false
	}
	conf.HTTP.Routers[key] = rt
	return true
}
```

A test case with `disableHTTPEntryPoint: true` plus a canary annotation would cover it.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
sdelicata, line 1: Could you add the new option to the provider reference table please? It should go in `docs/content/reference/install-configuration/providers/kubernetes/kubernetes-ingress-nginx.md`, with its default (`false`) and a word on the fact that it wins over `httpEntryPoint`.

It's also probably worth a line in `docs/content/migrate/nginx-to-traefik.md`, since this is the Traefik equivalent of the `controller.service.enablehttp: false` chart value @dl-mai mentioned in #13243.

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -98,7 +98,9 @@ func (p *Provider) translate(ctx context.Context, mc *model) *dynamic.Configurat
 			p.applyMiddlewares(mc, loc, defaultBackendTLSName, rtTLS, conf)
 		}
 
-		conf.HTTP.Routers[defaultBackendName] = rt
+		if !p.DisableHTTPEntryPoint {
+			conf.HTTP.Routers[defaultBackendName] = rt
+		}
```
sdelicata: The `applyMiddlewares` call just above still runs for the non-TLS default backend router, so its middlewares end up in the dynamic config with no router referencing them, and they surface in the API and the dashboard. The main location loop below skips them, so the two paths behave differently.

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -15842,12 +15917,18 @@ func TestLoadIngresses(t *testing.T) {
 				<-eventCh
 			}
 
+			nonTLSEntryPoints := []string{"http"}
+			if test.disableHTTPEntryPoint {
+				nonTLSEntryPoints = []string{}
+			}
```
sdelicata, line 19465: Since the test sets `NonTLSEntryPoints` by hand, the `Init()` and SetEffectiveConfiguration()` changes aren't covered by anything. Maybe a small table test on `SetEffectiveConfiguration` asserting the field stays empty when the option is on?

And could you add a case with a canary annotation and one with `defaultBackendService` set please? Those are the paths where HTTP routers are still built today.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -167,7 +168,7 @@ func (p *Provider) Init() error {
 		return fmt.Errorf("validating %s provider configuration: %w", ProviderName, err)
 	}
 
-	if p.HTTPEntryPoint != "" {
+	if !p.DisableHTTPEntryPoint && p.HTTPEntryPoint != "" {
```
sdelicata: Setting both `disableHTTPEntryPoint` and `httpEntryPoint` silently drops the latter. To me they're mutually exclusive, so maybe `validateConfiguration()` could reject the combination?

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -100,8 +100,9 @@ type Provider struct {
...
 
-	HTTPEntryPoint  string `description:"Defines the EntryPoint to use for HTTP requests." json:"httpEntryPoint,omitempty" toml:"httpEntryPoint,omitempty" yaml:"httpEntryPoint,omitempty" export:"true"`
-	HTTPSEntryPoint string `description:"Defines the EntryPoint to use for HTTPS requests." json:"httpsEntryPoint,omitempty" toml:"httpsEntryPoint,omitempty" yaml:"httpsEntryPoint,omitempty" export:"true"`
+	HTTPEntryPoint  		string `description:"Defines the EntryPoint to use for HTTP requests." json:"httpEntryPoint,omitempty" toml:"httpEntryPoint,omitempty" yaml:"httpEntryPoint,omitempty" export:"true"`
+	HTTPSEntryPoint 		string `description:"Defines the EntryPoint to use for HTTPS requests." json:"httpsEntryPoint,omitempty" toml:"httpsEntryPoint,omitempty" yaml:"httpsEntryPoint,omitempty" export:"true"`
+	DisableHTTPEntryPoint 	bool   `description:"Disables the creation of HTTP routes for Ingress resources. Takes precedence over httpEntryPoint." json:"disableHTTPEntryPoint,omitempty" toml:"disableHTTPEntryPoint,omitempty" yaml:"disableHTTPEntryPoint,omitempty" export:"true"`
```
sdelicata: The option name says "EntryPoint" while the description says "routes".

```suggestion
	DisableHTTPEntryPoint 	bool   `description:"Disables the HTTP EntryPoint, so that no HTTP router is created for Ingress resources. Takes precedence over httpEntryPoint." json:"disableHTTPEntryPoint,omitempty" toml:"disableHTTPEntryPoint,omitempty" yaml:"disableHTTPEntryPoint,omitempty" export:"true"`
```

Unless we rename the option into `DisableHTTPRouters` which is more accurate to what is does. It doesn't disable an entryPoint (the entryPoint stays up and keeps serving every other provider), it stops the provider from creating non-TLS routers. WDYT?

sdelicata (changes requested):

pkg/provider/kubernetes/ingress-nginx/translator.go
sinashahoveisi: Hi @sdelicata, Good catch, thanks. Added a shared ```addNonTLSRouter``` helper and routed the main, canary, and non-canary routers all through it, so none can be missed going forward. Also gated the canary/non-canary middleware application on successful registration. Added a test case with ```disableHTTPEntryPoint: true``` + a canary-by-header ingress asserting only the ```-tls``` routers are present.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
sinashahoveisi: Added a row for ```disableHTTPEntryPoint``` to the reference table (default ```false```, notes it takes precedence over ```httpEntryPoint```), and a note in the nginx-to-traefik migration guide linking it to the ingress-nginx Helm chart's ```controller.service.enablehttp: false```.

pkg/provider/kubernetes/ingress-nginx/translator.go
sinashahoveisi: the ```applyMiddlewares``` call for the default-backend's non-TLS router is now skipped when ```disableHTTPEntryPoint``` is true, matching how the main location loop already handles this.

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
sinashahoveisi: Added a dedicated test in ```pkg/config/static/static_config_test.go``` (```TestSetEffectiveConfiguration_DisableHTTPEntryPoint```) since that logic actually lives in ```SetEffectiveConfiguration``` on the static config, not on the Provider itself. Also added canary-by-header + disable and defaultBackendService + disable cases to ```kubernetes_test.go```.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
sinashahoveisi: Agreed they should be mutually exclusive. added a check in ```validateConfiguration()```: setting both now returns an error (```httpEntryPoint cannot be set when disableHTTPEntryPoint is true```). Added test cases for both the conflict and the disable-only case.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
sinashahoveisi: Updated the description to ```"Disables the HTTP EntryPoint, so that no HTTP router is created for Ingress resources. Takes precedence over httpEntryPoint."```. kept the field name as-is to stay consistent with the existing ```httpEntryPoint```/```httpsEntryPoint``` naming in the same struct, but happy to rename to ```DisableHTTPRouters``` if you still feel that's clearer.

sinashahoveisi:
@sdelicata Thanks for the review. Pushed a commit addressing all the points. Let me know if anything else needs adjusting.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -193,7 +187,7 @@ This label selector locates the controller ConfigMap regardless of the namespace
...
+| `ssl-redirect` / `force-ssl-redirect` | `nginx.ingress.kubernetes.io/ssl-redirect` and `nginx.ingress.kubernetes.io/force-ssl-redirect` annotations, or cluster-wide [entryPoint redirection](../reference/install-configuration/entrypoints.md#configuration-example) | Traefik translates the annotations when they are present. For a global default, configure HTTP-to-HTTPS redirection on the `web` entryPoint and set `providers.kubernetesIngressNGINX.httpEntryPoint` / `httpsEntryPoint` if you need explicit entryPoint selection. To serve Ingresses on HTTPS only (no plaintext HTTP routers), set `providers.kubernetesIngressNGINX.disableHTTPEntryPoint: true` — the Traefik equivalent of the ingress-nginx Helm chart value `controller.service.enablehttp: false`. |
```
mmatur:
```suggestion
| `ssl-redirect` / `force-ssl-redirect` | `nginx.ingress.kubernetes.io/ssl-redirect` and `nginx.ingress.kubernetes.io/force-ssl-redirect` annotations, or cluster-wide [entryPoint redirection](../reference/install-configuration/entrypoints.md#configuration-example) | Traefik translates the annotations when they are present. For a global default, configure HTTP-to-HTTPS redirection on the `web` entryPoint and set `providers.kubernetesIngressNGINX.httpEntryPoint` / `httpsEntryPoint` if you need explicit entryPoint selection. To serve Ingresses on HTTPS only (no plaintext HTTP routers), set `providers.kubernetesIngressNGINX.disableHTTPEntryPoint: true` — the Traefik equivalent of the ingress-nginx Helm chart value `controller.service.enableHttp: false`. |
```

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -100,8 +101,9 @@ type Provider struct {
...
 
-	HTTPEntryPoint  string `description:"Defines the EntryPoint to use for HTTP requests." json:"httpEntryPoint,omitempty" toml:"httpEntryPoint,omitempty" yaml:"httpEntryPoint,omitempty" export:"true"`
-	HTTPSEntryPoint string `description:"Defines the EntryPoint to use for HTTPS requests." json:"httpsEntryPoint,omitempty" toml:"httpsEntryPoint,omitempty" yaml:"httpsEntryPoint,omitempty" export:"true"`
+	HTTPEntryPoint        string `description:"Defines the EntryPoint to use for HTTP requests." json:"httpEntryPoint,omitempty" toml:"httpEntryPoint,omitempty" yaml:"httpEntryPoint,omitempty" export:"true"`
+	HTTPSEntryPoint       string `description:"Defines the EntryPoint to use for HTTPS requests." json:"httpsEntryPoint,omitempty" toml:"httpsEntryPoint,omitempty" yaml:"httpsEntryPoint,omitempty" export:"true"`
+	DisableHTTPEntryPoint bool   `description:"Disables the HTTP EntryPoint, so that no HTTP router is created for Ingress resources. Takes precedence over httpEntryPoint." json:"disableHTTPEntryPoint,omitempty" toml:"disableHTTPEntryPoint,omitempty" yaml:"disableHTTPEntryPoint,omitempty" export:"true"`
```
```suggestion
	DisableHTTPEntryPoint bool   `description:"Disables the HTTP EntryPoint, so that no HTTP router is created for Ingress resources. Cannot be used together with httpEntryPoint." json:"disableHTTPEntryPoint,omitempty" toml:"disableHTTPEntryPoint,omitempty" yaml:"disableHTTPEntryPoint,omitempty" export:"true"`
```

pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingress-disable-http-entrypoint.yml
```diff
@@ -0,0 +1,23 @@
...
+    - host: whoami.localhost
+      http:
+        paths:
+          - backend:
+              service:
+                name: service1
+                port:
+                  number: 80
+            path: /
+            pathType: Prefix
```
mmatur:
```suggestion
            pathType: Prefix

```

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -167,7 +169,7 @@ func (p *Provider) Init() error {
 		return fmt.Errorf("validating %s provider configuration: %w", ProviderName, err)
 	}
 
-	if p.HTTPEntryPoint != "" {
+	if !p.DisableHTTPEntryPoint && p.HTTPEntryPoint != "" {
```
```suggestion
	if p.HTTPEntryPoint != "" {
```

Init calls validateConfiguration() first, which already errors when DisableHTTPEntryPoint && HTTPEntryPoint != ""

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -94,11 +94,13 @@ func (p *Provider) translate(ctx context.Context, mc *model) *dynamic.Configurat
...
-			p.applyMiddlewares(mc, loc, defaultBackendName, rt, conf)
+			if !p.DisableHTTPEntryPoint {
+				p.applyMiddlewares(mc, loc, defaultBackendName, rt, conf)
+			}
 			p.applyMiddlewares(mc, loc, defaultBackendTLSName, rtTLS, conf)
 		}
 
-		conf.HTTP.Routers[defaultBackendName] = rt
+		p.addNonTLSRouter(conf, defaultBackendName, rt)
 		conf.HTTP.Routers[defaultBackendTLSName] = rtTLS
```
mmatur:
```suggestion
         registered := p.addNonTLSRouter(conf, defaultBackendName, rt)
		conf.HTTP.Routers[defaultBackendTLSName] = rtTLS

		// Apply the full middleware stack from the ingress location annotations to
		// both catch-all routers, so options like enable-cors, custom-headers,
		// rate-limits, redirects, etc. configured on a "spec.defaultBackend only"
		// ingress reach its catch-all routers.
		if loc := mc.DefaultBackendLocation; loc != nil {
			if registered {
				p.applyMiddlewares(mc, loc, defaultBackendName, rt, conf)
			}
			p.applyMiddlewares(mc, loc, defaultBackendTLSName, rtTLS, conf)
		}
```

Keeping same pattern.

mmatur (changes requested):
Thanks for you contribution

Could you please rebase your PR and fix conflicts.
Also you will have to do a `make generate`

docs/content/migrate/nginx-to-traefik.md
sinashahoveisi: Fixed

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
sinashahoveisi: Updated the wording.

pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingress-disable-http-entrypoint.yml
sinashahoveisi: Added the trailing newline.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
sinashahoveisi: You're right, removed the redundant check.

pkg/provider/kubernetes/ingress-nginx/translator.go
sinashahoveisi: Applied exactly as suggested.

sinashahoveisi:
@mmatur Thanks! Rebased onto master, ran make generate, and addressed all the review points.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
sdelicata: Thanks for pushing back, but I'm not convinced by the consistency argument. `httpEntryPoint` answers "which entryPoint", while this one changes what gets built, so they're not on the same axis.

What bothers me more is that every other `disable*` option in the codebase names something that actually stops (`disableHTTP2`, `disableSvcExternalName`, `disableIngressClassLookup`, ...). Here the entryPoint keeps running untouched, it still serves the other providers, and it even keeps serving these Ingresses over TLS when `httpsEntryPoint` isn't set.

I suggest to find a better name. Nothing is released yet, so it's free now, but implies a deprecation cycle later.

Having a second thought, `disableHTTPRouters` can be a bit misleading since In Traefik's config vocabulary `http.routers` covers TLS routers too (`http.routers.<name>.tls`), so the name promises no HTTP routers at all while the TLS ones are still HTTP routers. Trades one inaccuracy for another.

WDYT about `disableNonTLSRouters`? It mirrors the internal `NonTLSEntryPoints` field it actually controls, and "non-TLS router" is wording we already use in the docs.

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -17155,6 +17156,290 @@ func TestLoadIngresses(t *testing.T) {
...
+			},
+		},
+		{
+			desc:                           "Disable HTTP entrypoint with default backend",
+			defaultBackendServiceName:      "whoami",
+			defaultBackendServiceNamespace: "default",
+			disableHTTPEntryPoint:          true,
+			paths: []string{
+				"services.yml",
+			},
```
sdelicata: If I'm not wrong, this case doesn't reach the guard it was added for. `mc.DefaultBackendLocation` is only set from an Ingress with `spec.defaultBackend` and no rules (`build.go:527`), and there's no Ingress loaded here, so the `if registered { applyMiddlewares(...) }` block never runs. The empty `Middlewares` map comes from nothing being generated at all, and the test still passes if you drop the guard.

Could you use `ingresses/ingress-with-default-backend-no-rules.yml` instead please? It has no rules plus an `enable-cors` annotation, so the CORS middleware should show up for `default-backend-tls` only.

sdelicata (changes requested):
Thanks @sinashahoveisi for your changes.
Just a few comments before I leave my LGTM.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
sinashahoveisi: Renamed to `disableNonTLSRouters` as suggested. makes sense, thanks for the detailed reasoning.

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
sinashahoveisi: Thanks, fixed. now uses `ingress-with-default-backend-no-rules.yml` and asserts the CORS/retry middlewares only attach to the TLS router.

sinashahoveisi:
@sdelicata Thanks. Renamed the option to `disableNonTLSRouters` and fixed the default-backend test to actually exercise the guard.

docs/content/migrate/nginx-to-traefik.md
```diff
@@ -193,7 +193,7 @@ This label selector locates the controller ConfigMap regardless of the namespace
...
+| `ssl-redirect` / `force-ssl-redirect` | `nginx.ingress.kubernetes.io/ssl-redirect` and `nginx.ingress.kubernetes.io/force-ssl-redirect` annotations, or cluster-wide [entryPoint redirection](../reference/install-configuration/entrypoints.md#configuration-example) | Traefik translates the annotations when they are present. For a global default, configure HTTP-to-HTTPS redirection on the `web` entryPoint and set `providers.kubernetesIngressNGINX.httpEntryPoint` / `httpsEntryPoint` if you need explicit entryPoint selection. To serve Ingresses on HTTPS only (no non-TLS routers), set `providers.kubernetesIngressNGINX.disableNonTLSRouters: true` — the Traefik equivalent of the ingress-nginx Helm chart value `controller.service.enableHttp: false`. |
```
sdelicata: I'm afraid the equivalence with `controller.service.enableHttp: false` isn't accurate. That chart value removes the port from the Service, so nothing reaches :80 at all. Here the entryPoint stays up, keeps serving the other providers, and even keeps serving these Ingresses over TLS when `httpsEntryPoint` isn't set.

Someone migrating will read that row and believe plaintext is impossible. Could you drop the comparison and just say what it does? Something like "stops this provider from creating non-TLS routers for its Ingresses. The HTTP entryPoint itself is unaffected, remove it from the install configuration if you want the port closed."

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
```diff
@@ -100,9 +100,9 @@ type Provider struct {
...
-	DisableHTTPEntryPoint bool   `description:"Disables the HTTP EntryPoint, so that no HTTP router is created for Ingress resources. Cannot be used together with httpEntryPoint." json:"disableHTTPEntryPoint,omitempty" toml:"disableHTTPEntryPoint,omitempty" yaml:"disableHTTPEntryPoint,omitempty" export:"true"`
+	HTTPEntryPoint       string `description:"Defines the EntryPoint to use for HTTP requests." json:"httpEntryPoint,omitempty" toml:"httpEntryPoint,omitempty" yaml:"httpEntryPoint,omitempty" export:"true"`
+	HTTPSEntryPoint      string `description:"Defines the EntryPoint to use for HTTPS requests." json:"httpsEntryPoint,omitempty" toml:"httpsEntryPoint,omitempty" yaml:"httpsEntryPoint,omitempty" export:"true"`
+	DisableNonTLSRouters bool   `description:"Disables the creation of non-TLS routers for Ingress resources. Cannot be used together with httpEntryPoint." json:"disableNonTLSRouters,omitempty" toml:"disableNonTLSRouters,omitempty" yaml:"disableNonTLSRouters,omitempty" export:"true"`
```
sdelicata, line 105: One thing this doesn't cover: `TLSEntryPoints` is only filled from `httpsEntryPoint` in `Init()`, so with `disableNonTLSRouters: true` and no `httpsEntryPoint`, the `-tls` routers get an empty `EntryPoints` and `mergeConfiguration` hands them every default entryPoint, `web` included. Plaintext on :80 is gone, but `https://host:80` still answers.

Pre-existing, nothing you introduced. Since this is the option where people expect :80 to go quiet, WDYT about a line in the reference table saying to set `httpsEntryPoint` too when you want everything on the HTTPS entryPoint?

sdelicata (changes requested):
Thanks for your reactivity! I think these are my last feedbacks.
Could you please also rebase your branch and resolve conflicts? 🙏

docs/content/migrate/nginx-to-traefik.md
sinashahoveisi: Fixed, dropped the comparison, states plainly what it does now.

pkg/provider/kubernetes/ingress-nginx/kubernetes.go
sinashahoveisi: Added a note recommending `httpsEntryPoint` be set alongside it.

sinashahoveisi:
@sdelicata Thanks, Docs updated and branch rebased onto master again.

sdelicata (approved):
LGTM
Thanks @sinashahoveisi for your contribution

mmatur (approved):
LGTM
Thanks

Issue #10303, opened 2023-12-15
`nativeLB: true` breaks round robin load balancing


DaazKu:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

# Context
We're using traefik in EKS and started using `nativeLB: true` in our `IngressRoute`s to fix 502 being returned when deploying.

# The problem
Since then we started to see that some pods were receiving WAY MORE traffic than others for the same deployment.

## Tests

Here are some tests we did on a deployment with 15 replica that were sent 10k requests using different configurations.
We used the following to perform the requests
```
seq 10000 | parallel 'curl -s "{URL}" -o /dev/null'
```

Requesting the k8s service (from inside the cluster) directly and completely bypassing traefik. 
```
POD    REQUESTS RECEIVED
POD-01 718
POD-02 695
POD-03 688
POD-04 680
POD-05 675
POD-06 673
POD-07 672
POD-08 664
POD-09 662
POD-10 652
POD-11 651
POD-12 650
POD-13 649
POD-14 639
POD-15 632
```

Requesting with `nativeLB: false` on the route's service
```
POD    REQUESTS RECEIVED
POD-01 673
POD-02 671
POD-03 671
POD-04 670
POD-05 670
POD-06 670
POD-07 668
POD-08 667
POD-09 666
POD-10 665
POD-11 665
POD-12 664
POD-13 662
POD-14 662
POD-15 656
```

Requesting with `nativeLB: true` on the route's service
```
POD    REQUESTS RECEIVED
POD-01 1517
POD-02 1348
POD-03 1230
POD-04 1097
POD-05 791
POD-06 631
POD-07 522
POD-08 501
POD-09 442
POD-10 441
POD-11 424
POD-12 402
POD-13 309
POD-14 204
POD-15 141
```

Calling with `nativeLB: true` on the route's service + custom server transport `ServersTransport -> maxIdleConnsPerHost: 0`
The idea was to try to force traefik to init a new connection on each request because k8s doesn't handle long lived connections well for load balancing. See https://learnk8s.io/kubernetes-long-lived-connections
```
POD    REQUESTS RECEIVED
POD-01 1727
POD-02 1283
POD-03 1193
POD-04 869
POD-05 790
POD-06 532
POD-07 521
POD-08 514
[...]

PR #13358, opened 2026-06-12
Clarify nativeLB load balancing behavior in Kubernetes reference docs

zarko-a:
### What does this PR do?

Clarifies the `nativeLB` documentation in the Kubernetes reference pages (the Ingress annotation and the TraefikService CRD field). The previous wording only said whether the load-balancer's children are pod IPs or the Service clusterIP, which left it ambiguous whether "native" meant Kubernetes-native or Traefik-native routing.

The updated text explains:

- `nativeLB: true` routes traffic to the Kubernetes Service, so load balancing is handled by Kubernetes (kube-proxy / Cilium / Calico / etc.).
- `nativeLB: false` routes directly to the pods, with Traefik making load balancing decisions.
- Traefik reuses backend connections regardless of the setting, and combining connection reuse with `nativeLB: true` can cause uneven load balancing — which can be mitigated with `maxIdleConnsPerHost: -1`.

### Motivation

Follow-up to #10303. Several users (myself included) found the existing wording ambiguous about what "native" refers to. @nmengin reviewed the proposed wording in the issue and asked for it to be applied to the reference pages.

### More

- [ ] Added/updated tests

### Additional Notes

Applied to the reference pages as @nmengin requested in the issue (the previously-referenced `routing/providers/kubernetes-ingress` page is being removed in #11945).


zarko-a:
Hey @nmengin . I'm submitting this small PR based on our prior conversation [here](https://github.com/traefik/traefik/issues/10303#issuecomment-3210711770)

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #13364, opened 2026-06-16
Clarify entryPoint middleware reference format

mloiseleur:
### What does this PR do?

Fixes the EntryPoint `http.middlewares` examples and documents the per-provider name format (table).

EntryPoint middleware names are used as-is, so kubernetescrd references need the Namespace prefix (`default-auth@kubernetescrd`, not `auth@kubernetescrd`); the old form never matched and silently failed.

### Motivation

The documented format is currently inaccurate. Confirmed against the v3.6 code (`pkg/provider/kubernetes/crd/kubernetes.go:252`): kubernetescrd middlewares are keyed `<namespace>-<name>@kubernetescrd`, and EntryPoint names are qualified without auto-prefixing. Matches the form already used for Ingress, ServersTransport and TLSOption references.

### More

- [ ] Added/updated tests

### Additional Notes

* Documentation-only change, targeting `v3.6` per the PR template.

### DIFF
--- a/docs/content/reference/install-configuration/entrypoints.md
+++ b/docs/content/reference/install-configuration/entrypoints.md
@@ -246,7 +246,7 @@ Behavior examples:
 | <a id="opt-false-2" href="#opt-false-2" title="#opt-false-2">false</a> | foo=bar&baz=bar;foo | foo=bar&baz=bar&foo     |
 | <a id="opt-true-2" href="#opt-true-2" title="#opt-true-2">true</a> | foo=bar&baz=bar;foo | foo=bar&baz=bar%3Bfoo   |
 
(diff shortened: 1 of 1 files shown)


mloiseleur:
I also fixed minor inconsistencies on section headers, in a separate commit. 

:thinking: It's kinda weird to have a mix _Configuration Details_ (like "Encoded Characters" or "Forwarded Headers") under the same header as real _Configuration Options_ (like "encodeQuerySemicolons" or this "http.middlewares").

Wdyt of creating a third h3 header named _Configuration Details_ on this page, and move all related sections under this header ?

rtribotte (approved):
Thanks!

PR #13367, opened 2026-06-16
Avoid collisions for Gateway API services names

rtribotte:
### What does this PR do?

This PR changes the naming for generated Traefik services, and associated middlewares, to avoid collisions.

### Motivation

To prevent services/middewares configurations to collide.

### More

- [x] Added/updated tests

### Additional Notes

Assisted-by: Claude Opus 4.8

### DIFF
--- a/integration/testdata/rawdata-gateway.json
+++ b/integration/testdata/rawdata-gateway.json
@@ -120,7 +120,7 @@
 				"dashboard@internal"
 			]
 		},
-		"default-whoami-http-80@kubernetesgateway": {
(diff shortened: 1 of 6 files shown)


juliens (approved):

kevinpollet (approved):

zarko-a:
This PR changed the `exported_service` metrics label format in significant way that it breaks alerting if anyone has filters to include/exclude `exported_service` based on the value. We had filters for example `linux-backend-api-*` that broke because after the change that service got prepended with httproute, e.g. `httproute-linux-backend-api-...`

I'm sure we are not the only ones who got bit by this, it would be very beneficial to add this to release notes or [upgrade guide](https://doc.traefik.io/traefik/master/migrate/v3/).

PR #13368, opened 2026-06-16
Fix CORS wildcard when allow-credentials is true

AnatoleLucet:
### What does this PR do?

This PR fixes two misalignment with the W3C specs on CORS:

1. [[spec](https://fetch.spec.whatwg.org/#http-responses)] `Access-Control-Expose-Headers` is now also returned in preflight responses. Before that, it was only returned in non-preflight responses, even though it's supposed to be a preflight-only header.
2. [[spec](https://fetch.spec.whatwg.org/#cors-protocol-and-credentials)] when `AC-Allow-Credentials` is configured to `true`, the CORS headers in the backend responses shouldn't include wildcards. So when `AC-Allow-Credentials` is configured as `true`, the middleware now echos the values from the request in the response's CORS headers instead of returning wildcards.

### Motivation

W3C conformance for the upcoming Gateway API CORS filter in the K8s provider.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/headers/header.go
+++ b/pkg/middlewares/headers/header.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -78,10 +79,8 @@ func (s *Header) PostRequestModifyResponseHeaders(res *http.Response) error {
 	}
 
 	if res != nil && res.Request != nil {
-		originHeader := res.Request.Header.Get("Origin")
-		allowed, match := s.isOriginAllowed(originHeader)
-
-		if allowed {
+		match, ok := s.matchOrigin(res.Request.Header.Get("Origin"))
+		if ok {
 			res.Header.Set("Access-Control-Allow-Origin", match)
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/middlewares/headers/header.go
```diff
@@ -154,41 +155,63 @@ func (s *Header) processCorsHeaders(rw http.ResponseWriter, req *http.Request) b
...
+				rw.Header().Set("Access-Control-Allow-Headers", acrh)
+			}
+		} else if allowHeaders != "" {
 			rw.Header().Set("Access-Control-Allow-Headers", allowHeaders)
 		}
 
 		allowMethods := strings.Join(s.headers.AccessControlAllowMethods, ",")
-		if allowMethods != "" {
+		if slices.Contains(s.headers.AccessControlAllowMethods, "*") && s.headers.AccessControlAllowCredentials {
+			if acrm := req.Header.Get("Access-Control-Request-Method"); acrm != "" {
```
gndz07: Seems to me this is already in a block where `Access-Control-Request-Method` shouldn't be empty.

mmatur (approved):

pkg/middlewares/headers/header.go
```diff
@@ -154,41 +155,63 @@ func (s *Header) processCorsHeaders(rw http.ResponseWriter, req *http.Request) b
...
+func (s *Header) matchOrigin(origin string) (string, bool) {
 	for _, item := range s.headers.AccessControlAllowOriginList {
-		if item == "*" || item == origin {
-			return true, item
+		switch item {
+		case origin:
+			return item, true
+		case "*":
+			if s.headers.AccessControlAllowCredentials {
+				return origin, true // Can't use wildcard with credentials
```
gndz07:
```suggestion
				return origin, true // Can't use wildcard with credentials.
```

pkg/middlewares/headers/header.go
AnatoleLucet: good catch :pray:

gndz07 (approved):
LGTM

PR #13369, opened 2026-06-16
Create the redis rate limiter script lazily

jspdown:
### What does this PR do?

This moves the token bucket Lua script used by the Redis rate limiter from an exported package-level variable into the Redis limiter constructor, guarded by a `sync.Once` so it is still built once and shared across limiters. The script is now created the first time a Redis rate limiter is instantiated rather than when the package is imported, and `AllowTokenBucketScript`, which was only referenced within the package, is no longer exported.

### Motivation

The script is only used by the Redis rate limiter, so building it during package initialization is unnecessary for setups that only use the in-memory limiter. Creating it lazily co-locates the script with the limiter that uses it and avoids building it on import.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/middlewares/ratelimiter/lua.go
+++ b/pkg/middlewares/ratelimiter/lua.go
@@ -2,6 +2,7 @@ package ratelimiter
 
 import (
 	"context"
+	"sync"
 
 	"github.com/redis/go-redis/v9"
 )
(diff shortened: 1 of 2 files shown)


juliens (approved):

rtribotte (approved):
LGTM

traefiker:

traefiker:

PR #13376, opened 2026-06-18
Expose disableRequestBuffer and disableResponseBuffer buffering options (Issue #12869)

### LINKED ISSUE #12869: Expose disableRequestBuffer and disableResponseBuffer in config providers

V-Legendre:
### What does this PR do?

Exposes the `disableRequestBuffer` and `disableResponseBuffer` options of the buffering middleware to the file (YAML/TOML), Docker label, and KV providers.

These fields were already wired to the oxy buffer layer but were excluded from the configuration providers via `toml:"-" yaml:"-" label:"-" file:"-" kv:"-"` struct tags. This PR changes the tags to match the other fields of the `Buffering` struct so the options can be set from configuration.

### Motivation

Closes #12869.

Enabling the buffering middleware (e.g. to enforce `maxRequestBodyBytes`) also buffers the entire response, which breaks streaming/chunked responses (SSE, gRPC). There was no way to keep request buffering while letting responses stream.

This allows, for example:

```yaml
http:
  middlewares:
    request-only-buffer:
      buffering:
        maxRequestBodyBytes: 8000000
        disableResponseBuffer: true
```

### More

- [x] Added/updated tests

### Additional Notes

Scope is limited to the file/label/KV providers, matching the issue's proposed change. The Kubernetes CRD uses a separate `Buffering` type and is left untouched.


kevinpollet:
Thanks a lot for this contribution, @V-Legendre, and for following up on #12869!

We noticed that PR #13029 already addresses the same change — exposing `disableRequestBuffer` and `disableResponseBuffer` to the configuration providers — and it's currently in our review queue. To avoid duplicated effort, we'll consolidate the work there.

Would you be open to reviewing #13029 and sharing any improvements (for example, your approach around the file/label/KV providers, or any test coverage that's missing there)? Your input would be very welcome to help get that PR merged. We'll likely close this one as a duplicate, but we really appreciate you taking the time to put it together.

V-Legendre:
> Thanks a lot for this contribution, @V-Legendre, and for following up on #12869!
> 
> We noticed that PR #13029 already addresses the same change — exposing `disableRequestBuffer` and `disableResponseBuffer` to the configuration providers — and it's currently in our review queue. To avoid duplicated effort, we'll consolidate the work there.
> 
> Would you be open to reviewing #13029 and sharing any improvements (for example, your approach around the file/label/KV providers, or any test coverage that's missing there)? Your input would be very welcome to help get that PR merged. We'll likely close this one as a duplicate, but we really appreciate you taking the time to put it together.

Hey @kevinpollet , for sure i'd be open to review it :)

PR #13378, opened 2026-06-18
feat(otel): add traefik.service.name attribute to client spans

### LINKED ISSUE #13365: Feature Request: Add internal service name to client spans in opentelemetry
### Welcome!

When Traefik acts as a client (outbound requests to Kubernetes Pods) and a request faces a timeout or network drop, the client span populates `server.address` with the ephemeral Pod IP.
Because the connection is dropped, no child span is created by the destination service. When sending these traces to an APM, the opentelemetry service graph generator cannot aggregate the nodes properly. Instead of a clean, unified service node, it draws hundreds of unique numeric IP nodes.
Since Traefik internally knows the logical name of the dynamic service it's routing traffic to before selecting the specific backend Pod IP, it would be extremely useful if Traefik could inject this logical name as an attribute in the client span. For example, using the OpenTelemetry standard `service.peer.name` or a custom `traefik.service.name` attribute.
This would allow APM tools and Service Graphs to correctly group dropped and slow outbound traffic under the correct logical service without needing to process or map individual IP addresses at the collector level.

sinashahoveisi:
### What does this PR do?

When Traefik acts as a client (outbound requests to backend Pods), the `ReverseProxy` 
client span only populates `server.address` with the ephemeral Pod IP. This causes APM 
tools and service graph generators to draw hundreds of unique numeric IP nodes instead 
of a unified logical service node.

This PR adds the logical service name as a `traefik.service.name` attribute on the 
client span by:

1. Storing the service name in the request context inside `serviceTracing.ServeHTTP`
2. Reading it in the `RoundTripper` (`pkg/proxy/httputil/observability.go`) and setting 
   it as an attribute on the `ReverseProxy` client span

### Motivation

Fixes #13365

### More
- [ ] Added/updated tests

### Additional Notes

The `traefik.service.name` attribute is already used on the internal `Service` span 
(in `serviceTracing`), so this PR reuses the same attribute name for consistency on 
the client span as well.

### DIFF
--- a/pkg/middlewares/observability/service.go
+++ b/pkg/middlewares/observability/service.go
@@ -15,6 +15,19 @@ const (
 	serviceTypeName = "TracingService"
 )
 
+type serviceNameKey struct{}
+
+// WithServiceName stores the logical service name in the context.
+func WithServiceName(ctx context.Context, service string) context.Context {
+	return context.WithValue(ctx, serviceNameKey{}, service)
+}
+
+// ServiceNameFromContext retrieves the logical service name from the context.
+func ServiceNameFromContext(ctx context.Context) (string, bool) {
+	name, ok := ctx.Value(serviceNameKey{}).(string)
+	return name, ok && name != ""
+}
+
 type serviceTracing struct {
 	service string
 	next    http.Handler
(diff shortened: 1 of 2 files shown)


rtribotte:
Hey @sinashahoveisi,

Thanks for this contribution!

PR #13379, opened 2026-06-18
Fix ability to preserve query parameters in nginx.ingress.kubernetes.io/app-root annotation

dfeinblatt:
This PR fixes a bug with the `kubernetesingressnginx` provider support of the `nginx.ingress.kubernetes.io/app-root` annotation where the app-root redirect is not able to be configured to keep the query parameters on the redirect. With ingress nginx, this functionality was accomplished by appending `$is_args$args` to the annotation path value, so the `kubernetesingressnginx` provider should support the same functionality and configuration

I also added update to the regex as per suggested in https://github.com/traefik/traefik/pull/13330#discussion_r3403643829. Although that PR was closed, the suggestion is still valid I believe.

### Motivation
Fix https://github.com/traefik/traefik/issues/13377 so that traefik supports same functionality for this annotation as ingress nginx.

### More

- [x] Added/updated tests

### Additional Notes

With this change Traefik should now be fully supporting the annotation's functionality, so the docs can stay as they currently are with no notes/limitations for this annotation.

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingress-with-app-root-query-params.yml
+++ b/pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingress-with-app-root-query-params.yml
@@ -1,0 +1,22 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: ingress-with-app-root-query-params
+  namespace: default
+  annotations:
+    nginx.ingress.kubernetes.io/app-root: /foo$is_args$args
+
+spec:
+  ingressClassName: nginx
+  rules:
+    - host: app-root-query-params.localhost
+      http:
+        paths:
+          - path: /bar
+            pathType: Prefix
+            backend:
+              service:
+                name: whoami
+                port:
+                  number: 80
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
@@ -4635,13 +4635,13 @@ func TestLoadIngresses(t *testing.T) {
 					Middlewares: map[string]*dynamic.Middleware{
 						"default-ingress-with-app-root-rule-0-path-0-app-root": {
 							RedirectRegex: &dynamic.RedirectRegex{
-								Regex:       `^(https?://[^/]+)/(\?.*)?$`,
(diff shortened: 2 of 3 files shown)


kevinpollet:
Hey @dfeinblatt,

Thanks for this contribution!

This totally makes sense that it requires `$is_args$args` to be set.

gndz07:
Hi @dfeinblatt, thanks for the PR!

We discussed this internally and would prefer a different approach.

Traefik already has a general-purpose NGINX variable interpolation system (`ReplaceVariables` in `pkg/middlewares/ingressnginx/interpolation.go`) that handles `$is_args`, `$args`, and other NGINX variables natively at request time. The `upstreamvhost` middleware uses exactly this pattern.

We'd prefer the app-root annotation follow the same approach, which is to intercepts root requests, calls `ReplaceVariables` on the annotation value against the live request, then issues the redirect. This would make the full range of NGINX variables work in app-root values, not just the `$is_args$args` suffix.

Happy to discuss the implementation shape if that would be helpful.

kevinpollet:
Hello @dfeinblatt,

We tried to rebase your pull request to move forward, but the PR was opened from an organization-owned fork. GitHub doesn't let maintainers push commits/rebase PRs from organization forks — the "Allow edits by maintainers" option only works for forks on a personal user account. As a result, we're also unable to run the parts of our CI/CD needed to merge.

Could you reopen the pull request from a fork on your personal user account and address the comments?

dfeinblatt:
> Hi @dfeinblatt, thanks for the PR!
> 
> We discussed this internally and would prefer a different approach.
> 
> Traefik already has a general-purpose NGINX variable interpolation system (`ReplaceVariables` in `pkg/middlewares/ingressnginx/interpolation.go`) that handles `$is_args`, `$args`, and other NGINX variables natively at request time. The `upstreamvhost` middleware uses exactly this pattern.
> 
> We'd prefer the app-root annotation follow the same approach, which is to intercepts root requests, calls `ReplaceVariables` on the annotation value against the live request, then issues the redirect. This would make the full range of NGINX variables work in app-root values, not just the `$is_args$args` suffix.
> 
> Happy to discuss the implementation shape if that would be helpful.

Thanks for the feedback. I will read up on that part of the code and how `upstreamvhost` handles this and then will take a look at implementing this approach.

dfeinblatt:
Closing this PR as I have created new PR from personal fork as requested

PR #13380, opened 2026-06-19
Bump axios to v1.18.0

gndz07:
### What does this PR do?

Bump Axios version to 1.18.0.
`form-data` is a transitive dependency from axios, and it should use the latest patch of v4.0. Therefor, we won't need an explicit version pinning on the resolution.

### Motivation

To be up to date.

### More

~- [ ] Added/updated tests~


rtribotte (approved):
LGTM

PR #13383, opened 2026-06-19
Add regression test coverage ensuring request trailers are not forwarded to the backend

juliens:
## What does this PR do?
This PR adds regression test coverage ensuring that HTTP request trailer values are not forwarded to the backend by the reverse proxy.

The coverage spins up a backend that captures req.Trailer after fully reading the body, places the real Traefik proxy in front of it, and sends a client request carrying a trailer (X-Test-Trailer: trailer-value, chunked body via io.Pipe). It then asserts the backend never observes the trailer value.

## Motivation
Request trailers arrive after the body, once routing and security decisions have already been made on the headers. Forwarding them could let a late field bypass earlier checks, which raises security concerns in Traefik. The current behavior is to not propagate trailer values to the backend, and we want to keep it that way.

Today this behavior is implicit and untested: the stdlib ReverseProxy clones the request (req.Clone) before the trailer values are populated, so the announced trailer key may transit but its value is always empty.

This coverage locks that behavior in, so any future change that starts propagating trailer values to the backend will fail CI.


rtribotte (approved):
LGTM

PR #13386, opened 2026-06-19
Fix force-ssl-redirect routing 418 when TLS is terminated upstream

### LINKED ISSUE #13385: `force-ssl-redirect` returns 418 when TLS is terminated at the load balancer
### Welcome!

## Summary

When using the `nginx.ingress.kubernetes.io/force-ssl-redirect` annotation with TLS terminated upstream (e.g. a cloud load balancer), Traefik returns HTTP 418 for all requests instead of routing them to the backend.

The root cause is that the ingress-nginx translator unconditionally sets `noop@internal` as the service for any router with `force-ssl-redirect`, assuming every request will be redirected. When `X-Forwarded-Proto: https` is present from a trusted source, the `RedirectScheme` middleware correctly skips the redirect but the request then falls through to `noop@internal`, which always returns 418.

## Environment

- Traefik with `kubernetesIngressNGINX` provider
- TLS terminated by an upstream load balancer (not by Traefik)
- Traefik receives requests on the plain HTTP entrypoint
- Load balancer sets `X-Forwarded-Proto: https`
- Entrypoint configured with `forwardedHeaders.trustedIPs` including the load balancer's IP

## Steps to reproduce

> **Skip the first block if you already have a Kubernetes cluster available.**

```bash
# Optional: create a local cluster with k3d
k3d cluster create repro --k3s-arg '--disable=traefik@server:0' -p "8080:80@loadbalancer"
```

Install Traefik with the `kubernetesIngressNGINX` provider. `forwardedHeaders.insecure: true`
stands in for a real load balancer passing `X-Forwarded-Proto: https` from a trusted IP.

```bash
helm repo add traefik https://traefik.github.io/charts && helm repo update

helm install traefik traefik/traefik -n traefik --create-namespace -f - <<'EOF'
providers:
  kubernetesIngress:
    enabled: false
  kubernetesIngressNGINX:
    enabled: true
additionalArguments:
  - "--entryPoints.web.forwardedHeaders.insecure=true"
EOF

kubectl rollout status -n traefik deployment/traefik
```

Deploy a backend and an Ingress with `force-ssl-redirect: "true"` and no TLS section (the
topology where TLS is terminated by an upstream load balancer, not by Traefik):

```bash
kubectl apply -f - <<'EOF'
kind: IngressClass
metadata:
  name: nginx
spec:
  controller: k8s.io/ingress-nginx
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: echo
spec:
  replicas: 1
  selector:
    matchLabels:
      app: echo
  template:
    metadata:
      labels:
        app: echo
    spec:
      containers:
        - name: echo
          image: ealen/echo-server:latest
          ports:
            - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: echo
spec:
  selector:
    app: echo
  ports:
    - port: 80
      targetPort: 80
---
kind: Ingress
metadata:
  name: echo
  annotations:
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
spec:
  ingressClassName: nginx
  rules:
    - host: echo.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: echo
                port:
                  number: 80
EOF

kubectl rollout status deployment/echo
```

Verify that plain HTTP is redirected correctly, then simulate a request from an upstream load
balancer that already terminated TLS:

```bash
$ curl -si http://localhost:8080/ -H "Host: echo.example.com" | head -1
HTTP/1.1 308 Permanent Redirect


$ curl -si http://localhost:8080/ -H "Host: echo.example.com" -H "X-Forwarded-Proto: https" | head -1
HTTP/1.1 418 I'm a teapot
```

## Expected behavior

Requests arriving with `X-Forwarded-Proto: https` from a trusted IP are forwarded to the backend service. Requests without it (plain HTTP from a client, no LB in between) are redirected to HTTPS.

## Actual behavior

All requests with `X-Forwarded-Proto: https` return HTTP 418.

```
$ curl -si http://localhost:8080/ -H "Host: echo.example.com" -H "X-Forwarded-Proto: https" | head -1
HTTP/1.1 418 I'm a teapot
```

v3.7.5

```
$ kubectl get deploy/traefik -n traefik -oyaml | grep image:
        image: docker.io/traefik:v3.7.5
```

## Environment

- Traefik with `kubernetesIngressNGINX` provider
- TLS terminated by an upstream load balancer (not by Traefik)
- Traefik receives requests on the plain HTTP entrypoint
- Load balancer sets `X-Forwarded-Proto: https`
- Entrypoint configured with `forwardedHeaders.trustedIPs` including the load balancer's IP

## Configuration

```
providers:
  kubernetesIngress:
    enabled: false
  kubernetesIngressNGINX:
    enabled: true
additionalArguments:
  - "--entryPoints.web.forwardedHeaders.insecure=true"
```

_No response_

carloslima:
### What does this PR do?

Remove the `noop@internal` overwrite and the early return from `applyMiddlewares` in `ingress-nginx/translator.go`. The real backend service is already set on the router before `applyMiddlewares` is called, so the redirect path is unchanged: when the redirect fires it short-circuits the chain and the service is never reached. When `X-Forwarded-Proto: https` bypasses the redirect, the request now reaches the real backend through the full middleware chain.

### Motivation

When an Ingress uses `force-ssl-redirect` (or `ssl-redirect` via a TLS section) and TLS is terminated by an upstream load balancer, Traefik receives plain HTTP with `X-Forwarded-Proto: https`. The RedirectScheme middleware correctly skips the redirect in this case, but the request then hit `noop@internal` (which always returns HTTP 418) because `applyMiddlewares` unconditionally overwrote the router service with `noop@internal` and returned early, discarding all other middlewares.

### More

- [x] Added/updated tests

Closes #13385

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingress-with-force-ssl-redirect-no-tls.yml
+++ b/pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingress-with-force-ssl-redirect-no-tls.yml
@@ -1,0 +1,22 @@
+---
+kind: Ingress
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: ingress-with-force-ssl-redirect-no-tls
+  namespace: default
+  annotations:
+    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
+
+spec:
+  ingressClassName: nginx
+  rules:
+    - host: forcesslredirectnotls.localhost
+      http:
+        paths:
+          - path: /
+            pathType: Exact
+            backend:
+              service:
+                name: whoami
+                port:
+                  number: 80
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
@@ -1150,8 +1150,11 @@ func TestLoadIngresses(t *testing.T) {
 							EntryPoints: []string{"http"},
 							Rule:        `Host("whoami.localhost") && PathPrefix("/")`,
 							RuleSyntax:  "default",
-							Middlewares: []string{"default-ingress-with-no-annotation-rule-0-path-0-redirect-scheme"},
-							Service:     "noop@internal",
+							Middlewares: []string{
+								"default-ingress-with-no-annotation-rule-0-path-0-redirect-scheme",
+								"default-ingress-with-no-annotation-rule-0-path-0-retry",
+							},
+							Service: "default-ingress-with-no-annotation-whoami-80",
 							Observability: &dynamic.RouterObservabilityConfig{
 								Metadata: &dynamic.ObservabilityMetadata{
 									Ingress: &dynamic.KubernetesIngressMetadata{
@@ -1170,6 +1173,9 @@ func TestLoadIngresses(t *testing.T) {
 								Scheme:                 "https",
 								ForcePermanentRedirect: true,
 							},
+						},
+						"default-ingress-with-no-annotation-rule-0-path-0-retry": {
+							Retry: &dynamic.Retry{Attempts: 3},
 						},
 						"default-ingress-with-no-annotation-rule-0-path-0-tls-retry": {
 							Retry: &dynamic.Retry{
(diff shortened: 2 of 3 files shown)


rtribotte:
Hey @carloslima,

gndz07 (approved):
LGTM, thanks!

mmatur (approved):

PR #13387, opened 2026-06-20
feat: Server HealthCheck Retry Mechanism

### LINKED ISSUE #12390: Server HealthCheck Retry Mechanism
### Welcome!

We have implemented a HealthCheck mechanism in the production environment to enable failover. However, after several days of operation, we encountered occasional HealthCheck timeouts. Since the dependent service is a third-party one, occasional request timeouts are inevitable. Therefore, we hope to add a retry mechanism to the HealthCheck, similar to the configuration example in Nginx Plus:

```
location / {
    proxy_pass   http://backend;
    health_check interval=10 fails=3 passes=2;
}
```

fails: Failure threshold. Nginx will mark the upstream node as unavailable and stop routing traffic to it only after N consecutive failed health checks.

passes: Recovery threshold. A node marked as unavailable will be restored to the available state and resume receiving traffic only after M consecutive successful health checks.

Reference documentation: https://docs.nginx.com/nginx/admin-guide/load-balancer/http-health-check/

This enhancement will make the feature more robust.

hanxiantao:
### What does this PR do?

  Adds `fails` and `passes` thresholds to active health checks for HTTP and TCP services.

  `fails` controls how many consecutive failed probes are required before a server is marked unhealthy. `passes` controls how many consecutive successful probes
  are required before an unhealthy server is marked healthy again. Both default to `1`, preserving the previous behavior.

  The change also updates File, Label, KV, Kubernetes CRD configuration support, generated CRD manifests, documentation, and WebUI display.

  ### Motivation

  A single transient health check failure can currently remove a healthy server from rotation immediately. This PR makes active health checks less sensitive to
  short-lived network or backend hiccups by allowing users to require consecutive failures or successes before changing server status.

  Closes #12390.

  ### More

  - [x] Added/updated tests

  ### Additional Notes

  The implementation covers both HTTP and TCP active health checks. Passive health checks are unchanged.

  `fails` and `passes` must be greater than zero; invalid runtime values fall back to `1`, and Kubernetes CRD schema enforces a minimum of `1`.

#### Configuration
```yaml
http:
  routers:
    healthcheck-demo:
      entryPoints:
        - web
      rule: PathPrefix(`/`)
      service: backend

  services:
    backend:
      loadBalancer:
        servers:
          - url: http://127.0.0.1:8082
        healthCheck:
          path: /hello
          interval: 5s
          timeout: 2s
          fails: 3
          passes: 2
```          
Shut down the local HTTP service on port 8082. The service status will only be changed to down after 3 consecutive failed health check probes. After restarting the HTTP service, the service status will be restored to up only after 2 consecutive successful health check probes.

image

#### Verification Result
The actual verification is consistent with the expected effect. The configured threshold rules for consecutive failures and successes take effect normally. The service status will not be frequently switched due to occasional single health check probe exceptions, which effectively avoids service jitter caused by transient network fluctuations and short-term backend service anomalies, and optimizes the stability of active health check scheduling.


kevinpollet:
Hey @hanxiantao,

Thanks for this contribution!

crtvmn:
Hello! Any update about this feature? We would be interesting by it

PR #13395, opened 2026-06-24
Index Kubernetes EndpointSlice by service name

kevinpollet:
### What does this PR do?

Adds an informer indexer on `discoveryv1.LabelServiceName` and replaces the per-call `EndpointSlices(ns).List(serviceSelector)` label-selector scan with an indexed `GetIndexer().ByIndex(...)` lookup, across the ingress, crd, gateway, and ingress-nginx providers.

### Motivation

The primary root-cause fix for #13011: the EndpointSlice migration in v3.1 (`a8a92eb2a`) replaced an O(1) Endpoints lookup with an O(N) scan of every EndpointSlice in the namespace on every `loadService` call, which dominates CPU during reconfig bursts.

Supercedes #13269

### More

- [X] Added/updated tests

### Additional Notes

Co-authored-by: Piotr Maksymiuk <piotr.maksymiuk@docplanner.com>


rtribotte (approved):
LGTM

nandorKollar:
FYI, to prove that the obvious, I created a quick benchmark with AI for this change, and there's a significant speedup too:
```
Slices   Before (ns/op)   After (ns/op)   Speedup
10       2,286            520             ~4.4x
100      6,668            1,613           ~4.1x
1000     53,854           13,350          ~4.0x

```
This should also manifest in drop in CPU usage.

PR #13396, opened 2026-06-24
Strip managedFields in informers

kevinpollet:
### What does this PR do?

Registers a SetTransform on every Kubernetes informer factory across the providers that drops metadata.managedFields from objects as they enter the informer cache, via a shared `k8s.StripManagedFields` transform.

### Motivation

Supercedes #13266

### More

- [X] Added/updated tests

### Additional Notes

Co-authored-by: Piotr Maksymiuk <piotr.maksymiuk@docplanner.com>


rtribotte (approved):
LGTM

PR #13398, opened 2026-06-24
Add app-root middleware with nginx variable interpolation

### LINKED ISSUE #13377: [ingress-nginx] Unable to preserve query parameters when using app-root annotation
### Welcome!

### What did you do?

This Issue corrects the inaccuracy in the explanation of ingress nginx behavior in the description of Issue #13315 and the associated PR #13330 , as was pointed out in https://github.com/traefik/traefik/issues/13315#issuecomment-4742571402 leading to its closure.

With ingress nginx, when using the annotation `nginx.ingress.kubernetes.io/app-root`, you can specify that query parameters should be kept in the redirect by using the nginx variables `$is_args$args`.

**Example**

You want an app-root redirect to path `/login`. With ingress nginx, you can choose whether you want that redirect to pass along the query parameters with it or not.
* **Without** query parameters kept on the redirect, use: `nginx.ingress.kubernetes.io/app-root: /login`
* **With** query parameters kept on the redirect, use: `nginx.ingress.kubernetes.io/app-root: /login$is_args$args`

### What did you see instead?

Using the annotation `nginx.ingress.kubernetes.io/app-root: /login$is_args$args` does not cause query parameters to be kept in the redirect. It yields the same exact functionality as using `nginx.ingress.kubernetes.io/app-root: /login`. 

This is a bug with Traefik's support of this annotation. When the dynamic redirect regex middleware is created [here](https://github.com/traefik/traefik/blob/ec80d1145c5f5b06e6cb6eab7dd04f9e4024a6fd/pkg/provider/kubernetes/ingress-nginx/translator.go#L439-L448), it should check if `$is_args$args` is in the annotation value, and if so, then specify the replacement for the regex as `"$1" + *loc.AppRoot + "$2"` so that query parameters are kept on the redirect. `$is_args$args` is the standard way of specifying that query parameters should be passed along with nginx.

3.7.3

Using the `kubernetesingressnginx` provider

_No response_

dfeinblatt:
### What does this PR do?
Replace the static RedirectRegex middleware for the app-root annotation with an ingress-nginx middleware that resolves nginx variables at request time using ReplaceVariables. This enables $is_args$args to preserve query parameters on the app-root redirect, fixing the bug #13377 to match ingress-nginx behavior.

### Motivation
Fix #13377 so that traefik supports same functionality for this annotation as ingress-nginx.

### More

- [x] Added/updated tests

### Additional Notes

Reopened PR of #13379 as requested

With this change Traefik should now be fully supporting the annotation's functionality by supporting the nginx variables, so the docs can stay as they currently are with no notes/limitations for this annotation.


gndz07 (approved):
LGTM, thanks!

dfeinblatt:
> LGTM, thanks!

Thanks for the updates you made, they make sense to me too!

kevinpollet (approved):
Thanks 👍

PR #13399, opened 2026-06-25
Fix handle empty unknown-length bodies in mirroring

### LINKED ISSUE #12079: mirroring failed for HTTP/3 requests when maxBodySize is set
### Welcome!

### What did you do?

I configured traefik with HTTP/3 enabled and setup a service with 100% mirroring.

When I access this service with my browser, everything is normal at first, but after a period of time, it responds with `Internal Server Error: creating reusable request: EOF`.

I guessed this might be because the browser upgrades the connection from HTTP/2 to HTTP/3. So I checked the access log, and all the request logs that have this type of response show `"RequestProtocol":"HTTP/3.0"`.

I also tried curl with HTTP/3 only (`curl --http3-only -k -v https://mirrored.service`) and the result was the same error. Using HTTP/2 gets normal response.

There was a similar issue #8045.

### What did you see instead?

It responds with `Internal Server Error: creating reusable request: EOF`.

```
Version:      3.5.0
Codename:     chabichou
Go version:   go1.24.5
Built:        2025-07-23T13:57:30Z
OS/Arch:      linux/amd64
```

```yaml
entryPoints:
  web:
    address: ":80"
    asDefault: true
    http:
      redirections:
        entryPoint:
          to: websecure
          scheme: https

  websecure:
    address: ":443"
    asDefault: true
    http3: {}
    http:
      tls: {}
```

```yaml
http:
  services:
    mirroring-service:
      loadBalancer:
        servers:
          - url: "https://x.x.x.x/"
```

```yaml
http:
  routers:
    whoami-service:
      rule: "Host(`mirrored.service`)"
      service: mirrored-service

  services:
    mirrored-service:
      mirroring:
        service: whoami-service
        maxBodySize: 0x100000
        mirrors:
          - name: mirroring-service
            percent: 100

    whoami-service:
      loadBalancer:
        servers:
          - url: "https://x.x.x.x/"
```

_No response_

amazon7737:
### What does this PR do?

This handles the empty request-body case when the incoming request has an unknown content length (`ContentLength == -1`). Instead of treating all unknown-length requests as bodyless, it only treats `io.ReadFull` returning `io.EOF` with no bytes read as an empty body.

The regression coverage also checks that unknown-length requests with an actual body are still mirrored.

### Motivation

Fixes #12079.

### More

Tests:
- test success
  - `go test ./pkg/server/service/loadbalancer/mirror`
  - `go test ./pkg/server/service/loadbalancer/failover`

### DIFF
--- a/pkg/server/service/loadbalancer/mirror/mirror.go
+++ b/pkg/server/service/loadbalancer/mirror/mirror.go
@@ -231,7 +231,7 @@ func newReusableRequest(req *http.Request, mirrorBody bool, maxBodySize int64) (
 	// level rather than declared via Content-Length, so a bodyless request arrives
 	// with a non-nil Body and ContentLength == -1. The same shape is possible for a
 	// chunked request that sends no chunks.
-	if errors.Is(err, io.EOF) && n == 0 {
+	if errors.Is(err, io.EOF) {
 		return &reusableRequest{
 			req:  req,
 			body: body[:n],
--- a/pkg/server/service/loadbalancer/mirror/mirror.go
+++ b/pkg/server/service/loadbalancer/mirror/mirror.go
@@ -228,8 +228,9 @@ func newReusableRequest(req *http.Request, mirrorBody bool, maxBodySize int64) (
 	body := make([]byte, maxBodySize+1)
 	n, err := io.ReadFull(req.Body, body)
 	// This happens with HTTP/3, where the end of the body is framed at the stream
-	// level instead of declared with Content-Length. A chunked request with no
-	// chunks has the same shape.
+	// level rather than declared via Content-Length, so a bodyless request arrives
+	// with a non-nil Body and ContentLength == -1. The same shape is possible for a
+	// chunked request that sends no chunks.
 	if errors.Is(err, io.EOF) && n == 0 {
 		return &reusableRequest{
 			req:  req,
--- a/pkg/server/service/loadbalancer/mirror/mirror.go
+++ b/pkg/server/service/loadbalancer/mirror/mirror.go
@@ -227,6 +227,12 @@ func newReusableRequest(req *http.Request, mirrorBody bool, maxBodySize int64) (
 	// the request body is larger than what we allow for the mirrors.
 	body := make([]byte, maxBodySize+1)
 	n, err := io.ReadFull(req.Body, body)
+	if errors.Is(err, io.EOF) && n == 0 {
+		return &reusableRequest{
+			req:  req,
+			body: body[:n],
+		}, nil, nil
+	}
 	if err != nil && !errors.Is(err, io.ErrUnexpectedEOF) {
 		return nil, nil, err
 	}
--- a/pkg/server/service/loadbalancer/mirror/mirror.go
+++ b/pkg/server/service/loadbalancer/mirror/mirror.go
@@ -227,6 +227,9 @@ func newReusableRequest(req *http.Request, mirrorBody bool, maxBodySize int64) (
 	// the request body is larger than what we allow for the mirrors.
 	body := make([]byte, maxBodySize+1)
 	n, err := io.ReadFull(req.Body, body)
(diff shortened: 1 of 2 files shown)


amazon7737:
this PR initially had an incorrect diff because the branch was based on master while targeting v3.7. It looks like some of the current labels may have been applied from the earlier incorrect diff and no longer match the current PR scope.

kevinpollet:
Hey @amazon7737,

Thanks for this contribution!

As this is a fix, please could you rebase your pull request on branch v3.6?

pkg/server/service/loadbalancer/mirror/mirror.go
```diff
@@ -227,6 +227,12 @@ func newReusableRequest(req *http.Request, mirrorBody bool, maxBodySize int64) (
 	// the request body is larger than what we allow for the mirrors.
 	body := make([]byte, maxBodySize+1)
 	n, err := io.ReadFull(req.Body, body)
+	if errors.Is(err, io.EOF) && n == 0 {
```
kevinpollet: Here I would add a comment to explain this case, something like the following, WDYT? 
Feel free to rephrase.

```
// This happens with HTTP/3, where the end of the body is framed at the stream
// level rather than declared via Content-Length, so a body-less request arrives
// with a non-nil Body and ContentLength == -1. The same shape is possible, 
// for a chunked request that sends no chunks.
```

kevinpollet (changes requested):

pkg/server/service/loadbalancer/mirror/mirror.go
amazon7737: How about this?

```
// This happens with HTTP/3, where the end of the body is framed at the stream
// level rather than declared via Content-Length, so a bodyless request arrives
// with a non-nil Body and ContentLength == -1. The same shape is possible for a
// chunked request that sends no chunks.
```

I only adjusted the last sentence to avoid the comma. I kept `ContentLength == -1` and `non-nil Body` explicit because they explain why this case does not hit the existing `ContentLength == 0` or `req.Body == nil` fast paths.

kevinpollet (approved):
Thanks 👍

pkg/server/service/loadbalancer/mirror/mirror.go
```diff
@@ -227,6 +227,16 @@ func newReusableRequest(req *http.Request, mirrorBody bool, maxBodySize int64) (
 	// the request body is larger than what we allow for the mirrors.
 	body := make([]byte, maxBodySize+1)
 	n, err := io.ReadFull(req.Body, body)
+	// This happens with HTTP/3, where the end of the body is framed at the stream
+	// level rather than declared via Content-Length, so a bodyless request arrives
+	// with a non-nil Body and ContentLength == -1. The same shape is possible for a
+	// chunked request that sends no chunks.
+	if errors.Is(err, io.EOF) && n == 0 {
```
rtribotte:
```suggestion
	if errors.Is(err, io.EOF) {
```

I think it is unnecessary to assert `n==0`, as the `EOF` error is always produced with `n==0`.

pkg/server/service/loadbalancer/mirror/mirror.go
rtribotte:
```suggestion
	if req.Body == nil || !mirrorBody {
```

It is probably better to not rely on the `req.ContentLength` at all, the read operations will catch the no body case anyway.

pkg/server/service/loadbalancer/mirror/mirror.go
amazon7737: Okay I understand. This is a duplicate check.

pkg/server/service/loadbalancer/mirror/mirror.go
amazon7737: I agree 😄

rtribotte (approved):
LGTM

Issue #11323, opened 2024-12-09
Incorrect doc for maxRequestBodyBytes?


mrypdm:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

We sending large POST request (~100 MB) through Traefik v3.1.2 and it is fails with `EOF occurred in violation of protocol` on client side

According to docs ([Traefik Buffering Documentation - Traefik](https://doc.traefik.io/traefik/middlewares/http/buffering/#maxrequestbodybytes)) `maxRequestBodyBytes=0` by default which means that there is no limitation for body size.

But if we add `maxRequestBodyBytes=0` explicitly, error does not happening.

Also, if we add `memRequestBodyBytes=115343360` (110 MB), which is by default 1 MB, the error does not occur either.

Also, if we send requests smaller than 1 MB, there is no error, but for larger sizes the error appears.

### What did you see instead?

There is a suspicion that `maxRequestBodyBytes` is actually not 0 by default, but is equal to or slightly greater than `memRequestBodyBytes`

### What version of Traefik are you using?

3.1.2

### What is your environment & configuration?

main config:

```yaml
---
log:
  level: "INFO"

global:
  checkNewVersion: false
  sendAnonymousUsage: false

entryPoints:
  http:
    address: ":80"
    transport:
      respondingTimeouts:
        readTimeout: "0s"
    forwardedHeaders:
      trustedIPs:
        - "***"
  https:
    address: ":443"
    http:
      tls: true
    transport:
      respondingTimeouts:
        readTimeout: "0s"
    forwardedHeaders:
      trustedIPs:
        - "***"
  traefik:
    address: ":8080"

api:
  dashboard: true
  disabledashboardad: true

ping:
  entryPoint: "traefik"

metrics:
  datadog:
    address: ***

providers:
  file:
    directory: "/etc/traefik.d/dynamic"
  consulCatalog:
    prefix: "traefik"
    exposedByDefault: false
    defaultRule: "Host(`***`)"
    refreshInterval: "5s"
    endpoint:
      address: "***"
    stale: true
```

dynamic config:

```yaml
---
http:
  routers:
    catchall:
      entryPoints:
        - "http"
        - "https"
      rule: "PathPrefix(`/`)"
      service: unavailable
      priority: 1
  services:
[...]

PR #13401, opened 2026-06-25
Clarify buffering middleware defaults

amazon7737:
### What does this PR do?

Clarifies that buffering middleware options only apply when the middleware is explicitly attached to a router.

It also explains that `maxRequestBodyBytes=0` keeps the request body size unlimited, but enabling the middleware still buffers the request before forwarding it upstream.

### Motivation

Related to #11323.

In that issue, setting `maxRequestBodyBytes=0` appeared to change large POST behavior. The important distinction is that configuring the option also enables the buffering middleware for the router, which can change how the request is forwarded upstream.

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/reference/routing-configuration/http/middlewares/buffering.md
+++ b/docs/content/reference/routing-configuration/http/middlewares/buffering.md
@@ -7,14 +7,12 @@ The `buffering` middleware limits the size of requests that can be forwarded to 
 
 With buffering, Traefik reads the entire request into memory (possibly buffering large requests into disk), and rejects requests that are over a specified size limit.
 
-The buffering middleware is not enabled by default.
-Its options only apply to routers that explicitly use this middleware.
-
-Setting `maxRequestBodyBytes` to `0` keeps the request body size unlimited.
-However, when the middleware is attached, Traefik still buffers the request body before forwarding it.
-As a result, Traefik can send the request upstream with a fixed `Content-Length` instead of streaming the original chunked body.
-
 This can help services avoid large amounts of data (`multipart/form-data` for example), and can minimize the time spent sending data to a Service
+
+!!! info
+
+    When the middleware is attached, Traefik still buffers the request body before forwarding it.
+    As a result, Traefik can send the request upstream with a fixed `Content-Length` instead of streaming the original chunked body.
 
 ## Configuration Examples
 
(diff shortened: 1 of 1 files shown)


rtribotte:
Hey @amazon7737,
Thanks for this contribution!

docs/content/reference/routing-configuration/http/middlewares/buffering.md
```diff
@@ -7,6 +7,13 @@ The `buffering` middleware limits the size of requests that can be forwarded to
 
 With buffering, Traefik reads the entire request into memory (possibly buffering large requests into disk), and rejects requests that are over a specified size limit.
 
+The buffering middleware is not enabled by default.
+Its options only apply to routers that explicitly use this middleware.
```
kevinpollet: I would remove these lines. A middleware only being active once attached to a router is the general contract for all middlewares, so documenting it specifically for buffering could suggest it behaves differently from the others.

docs/content/reference/routing-configuration/http/middlewares/buffering.md
```diff
@@ -7,6 +7,13 @@ The `buffering` middleware limits the size of requests that can be forwarded to
...
+
+Setting `maxRequestBodyBytes` to `0` keeps the request body size unlimited.
```
kevinpollet: I would move this clarification to the Configuration options section by adding a node in the `maxRequestBodyBytes` and `maxResponseBodyBytes`option descriptions, saying that 0 means unlimited.

docs/content/reference/routing-configuration/http/middlewares/buffering.md
```diff
@@ -7,6 +7,13 @@ The `buffering` middleware limits the size of requests that can be forwarded to
...
+However, when the middleware is attached, Traefik still buffers the request body before forwarding it.
+As a result, Traefik can send the request upstream with a fixed `Content-Length` instead of streaming the original chunked body.
```
kevinpollet: I would move those lines just before the Configuration examples, and add those lines in an info block, something like:

```
!!! info

    When the middleware is attached, Traefik still buffers the request body before forwarding it.
    As a result, Traefik can send the request upstream with a fixed `Content-Length` instead of streaming the original chunked body.

```

kevinpollet (changes requested):
Hello @amazon7737,

I reviewed your pull request and added some comments. 
As this pull request is a fix, could you rebase it on branch v3.6?

However, I tried to figure out whether something was wrong with the buffering middleware causing the issue, but on my side the defaults look good.
Did you tried to reproduce the issue?

amazon7737:
> I reviewed your pull request and added some comments. As this pull request is a fix, could you rebase it on branch v3.6?
> 
> However, I tried to figure out whether something was wrong with the buffering middleware causing the issue, but on my side the defaults look good. Did you tried to reproduce the issue?

Hello @kevinpollet,

No, I did not reproduce it in the reporter's exact environment.

What I checked locally was narrower. With a temporary focused Go test against v3.1.2, I verified that adding the buffering middleware with `maxRequestBodyBytes=0` keeps the request size unlimited, but still changes how the request is forwarded upstream. 

Without buffering, the upstream received the large request as chunked. With buffering enabled, the upstream received it with a fixed `Content-Length`.

So the intent of this PR was to document that buffering side effect. I will update the wording based on your comments.

docs/content/reference/routing-configuration/http/middlewares/buffering.md
```diff
@@ -9,6 +9,11 @@ With buffering, Traefik reads the entire request into memory (possibly buffering
 
 This can help services avoid large amounts of data (`multipart/form-data` for example), and can minimize the time spent sending data to a Service
 
+!!! info
+
+    When the middleware is attached, Traefik still buffers the request body before forwarding it.
```
kevinpollet:
```suggestion
    When the middleware is attached, Traefik buffers the request body before forwarding it.
```

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
LGTM

PR #13407, opened 2026-06-26
Optimize loadService calls for multiple paths sharing the same service host:port

nandorKollar:
### What does this PR do?

When multiple Ingress path rules reference the same backend service and port, loadService was called for each occurrence — fetching the Service, resolving EndpointSlices, and building the server list from scratch every time, only to overwrite the same map entry with an identical result.

This change checks whether conf.HTTP.Services[serviceName] already exists before calling loadService. If the service was already resolved by an earlier path rule, the redundant work is skipped entirely.

In clusters where many Ingress rules route to a shared backend (e.g. multiple hosts fronting the same API service), this eliminates O(N-1) unnecessary GetService + GetEndpointSlicesForService calls per shared service per configuration rebuild.

### Motivation

Minor and simple optimization when multiple paths are sharing the same service host:port.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/ingress/kubernetes.go
+++ b/pkg/provider/kubernetes/ingress/kubernetes.go
@@ -403,26 +403,29 @@ func (p *Provider) loadConfigurationFromIngresses(ctx context.Context, client Cl
 					continue
 				}
 
-				service, err := p.loadService(client, ingress.Namespace, pa.Backend)
-				if err != nil {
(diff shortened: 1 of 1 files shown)


rtribotte (approved):
Thanks!

juliens (approved):
LGTM

traefiker:

Issue #13321, opened 2026-06-08
Docker provider with exposedbydefault=true causes intermittent upstream unreachability and config churn on multi-stack hosts (~40 containers)


loneObserver1:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I run Traefik v3.3.7 as the only reverse proxy on a single Docker host (~43 containers, multiple Compose stacks on a shared external network web). Public sites sit behind Cloudflare (proxied DNS, Full Strict).

I used the Docker provider with:

providers.docker=true
providers.docker.network=web
providers.docker.exposedbydefault=true (initial setting)
Entrypoints web (:80) and websecure (:443)
forwardedHeaders.trustedIPs set to Cloudflare IPv4/IPv6 ranges on both entrypoints
ACME DNS challenge (Cloudflare provider), certs in a bind-mounted acme.json
Public services labeled with traefik.enable=true, Host(...), tls.certresolver, etc.
Many internal containers (DB, Redis, mail workers, monitoring agents) without routing labels
I monitored public URLs through Cloudflare (HTTP every 60s) and compared with local origin checks (curl to 127.0.0.1 with Host:).

After intermittent failures, I changed only providers.docker.exposedbydefault to false (kept explicit traefik.enable=true on public services) and redeployed Traefik without other stack changes.

### What did you see instead?

With exposedbydefault=true:

Intermittent HTTP 522 from Cloudflare on several unrelated public hostnames (often one at a time), recovering in ~30–60s
External monitoring yo-yo (DOWN/UP ~every minute) with message Request failed with status code 522
Same-minute local origin checks OK (curl via 127.0.0.1 + Host: → 302/200 in < 5 ms) while public/CF checks timed out (curl (28), 0 bytes) or returned 522
No host saturation (load < 1, ~35% RAM, no OOM)
Heavy Traefik DEBUG noise on routine Docker events (health_status: healthy, healthcheck exec_*): full dynamic config JSON dumps, ACME “Looking for provided certificate(s) to validate [...]” across many routers on each refresh
After switching to exposedbydefault=false (same Cloudflare, same backends, same labels on public services):

522 / monitoring flapping stopped and stayed stable 24+ hours
DEBUG logs showed expected Filtering disabled container for internal services only; public routers unchanged and working

[...]

PR #13412, opened 2026-06-28
skip health_status: healthy events to prevent config churn

KirylJazzSax:
### What does this PR do?

This PR prevents unnecessary configuration regeneration when Traefik receives a Docker event with `health_status: healthy`. When such an event is received, and it does not represent an actual state change relevant to routing or configuration, Traefik now skips rebuilding the dynamic configuration.

### Motivation

As described in #13321, Docker emits `health_status: healthy` events even when a container is already in a stable healthy state.

Currently, these events trigger the Docker provider to recompute configuration, even though no meaningful change has occurred. This leads to redundant configuration updates and unnecessary processing overhead.

This change ensures that only relevant health transitions trigger configuration regeneration, improving efficiency and reducing noise from Docker health event churn.

### More

* [x] Added/updated tests
* [ ] Added/updated documentation

### Additional Notes

This change is scoped specifically to Docker `health_status: healthy` events and does not affect other Docker event types or health state transitions.

### DIFF
--- a/pkg/provider/docker/pdocker.go
+++ b/pkg/provider/docker/pdocker.go
@@ -119,9 +119,7 @@ func (p *Provider) Provide(configurationChan chan<- dynamic.Message, pool *safe.
 				for {
 					select {
 					case event := <-res.Messages:
-						if event.Action == "start" ||
-							event.Action == "die" ||
-							strings.HasPrefix(string(event.Action), "health_status") {
+						if shouldHandleEvent(event.Action) {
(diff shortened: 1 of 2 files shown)


rtribotte:
Hey @KirylJazzSax,

Thanks for this contribution!

In the mean, as this is a bug fix, could you please rebase the changes on branch v3.6 ?

KirylJazzSax:
Hi!
Rebased!

PR #13417, opened 2026-06-29
Add a ServersTransport option to prefer IPv6 for backend connections

### LINKED ISSUE #10677: Prefer IPv6 option for Service Loadbalancer URL
### Welcome!

We use Dual Stack at our Company for all the Backend Services and use Traefik as a Reverse Proxy. Traefik by default on a EKS Cluster with IPv4 and IPv6 setup uses IPv4. IPv4 connections costs a lot more to use then IPv6 (NAT Gateway Costs) so we would really love an option to prefer IPv6 whenever available. HA Proxy for example has such an option and we could not find any possible way to use it on Traefik.

RamazanKara:
### What does this PR do?

Adds a `preferIPv6` option to `ServersTransport`. When enabled, Traefik resolves
backend server hostnames to IPv6 first and only falls back to IPv4 when no IPv6
address is reachable. Literal IP addresses and connections that already pin an
address family are left untouched.

The option is available on the dynamic `ServersTransport` (file, labels, KV), the
Kubernetes `ServersTransport` CRD, and the static `serversTransport` used for the
`default@internal` transport.

### Motivation

Closes #10677.

On dual-stack setups the backend service often resolves to both A and AAAA records,
and the dialer ends up using IPv4. In environments such as dual-stack EKS this is
more expensive than IPv6 (NAT gateway costs for IPv4 egress), so being able to
prefer IPv6 when it is available is desirable. HAProxy offers a similar
`resolve-prefer ipv6` option.

### More

- [x] Added/updated tests

### Additional Notes

The preference is applied at the dialer level, so it covers HTTP/1.1 and HTTP/2 over
TLS. Explicit h2c (cleartext HTTP/2) backends dial directly and are therefore not
affected, which is consistent with the existing behavior of the read/write timeouts.


kevinpollet:
Hey @RamazanKara,

Thanks for the contribution!

PR #13421, opened 2026-06-29
Fix HTTP/GRPC routes conflict

AnatoleLucet:
### What does this PR do?

Currently, both an HTTP and a GRPC route can coexist on the same hostname.
This PRs fixes this by setting the `ConditionStatus` to `false` on routes with a hostname that conflicts with another existing route ([spec](https://gateway-api.sigs.k8s.io/reference/api-types/grpcroute/#cross-serving)).

### Motivation

Fix unclear behavior when two routes are used on the same hostname.

### More

- [x] Added/updated tests


pkg/provider/kubernetes/gateway/grpcroute.go
```diff
@@ -69,6 +68,12 @@ func (p *Provider) loadGRPCRoutes(ctx context.Context, gateways []gatewayWithLis
 					accepted = false
 				}
 
+				listenerKey := match.gatewayNamespace + "/" + match.gatewayName + "/" + listener.Name
+				if accepted && attached.HasConflict(listenerKey, kindGRPCRoute, hostnames) {
+					acceptedCondition.Reason = routeReasonHostnameConflict
+					accepted = false
+				}
```
gndz07: Shouldn't we have a condition here?
```suggestion
				if accepted && attached.HasConflict(listenerKey, kindGRPCRoute, hostnames) {
					
					if acceptedCondition.Status == metav1.ConditionFalse {
					       acceptedCondition.Reason = routeReasonHostnameConflict
					}
					accepted = false
				}
```

pkg/provider/kubernetes/gateway/httproute.go
```diff
@@ -71,6 +71,12 @@ func (p *Provider) loadHTTPRoutes(ctx context.Context, gateways []gatewayWithLis
 					accepted = false
 				}
 
+				listenerKey := match.gatewayNamespace + "/" + match.gatewayName + "/" + listener.Name
+				if accepted && attached.HasConflict(listenerKey, kindHTTPRoute, hostnames) {
+					acceptedCondition.Reason = routeReasonHostnameConflict
```
gndz07: Same as above about the condition.

gndz07 (changes requested):

pkg/provider/kubernetes/gateway/attachment.go
```diff
@@ -0,0 +1,40 @@
...
+type routeClaim struct {
+	kind      string
+	hostname  gatev1.Hostname
+	timestamp time.Time
+}
+
+// attachedRoutes tracks which route kind owns each hostname on each listener.
+type attachedRoutes map[string][]routeClaim // listenerKey → []routeClaim
+
+func (a attachedRoutes) Claim(listenerKey, kind string, hostname gatev1.Hostname, timestamp time.Time) {
```
nandorKollar, line 19: nit: we can keep both methods (`Claim` and `HasConflict`) private

pkg/provider/kubernetes/gateway/attachment.go
```diff
@@ -0,0 +1,40 @@
...
+	for i, claim := range a[listenerKey] {
+		if hostnamesIntersect([]gatev1.Hostname{hostname}, []gatev1.Hostname{claim.hostname}) {
+			// Only claim if the timestamp is older than the existing claim to ensure the older route always wins.
+			if timestamp.Before(claim.timestamp) {
+				a[listenerKey][i] = routeClaim{hostname: hostname, kind: kind, timestamp: timestamp}
+			}
+			return
```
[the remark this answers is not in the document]
AnatoleLucet: fixed by the new Claim methods. also added tests for this since the logic is now a bit more subtle.

pkg/provider/kubernetes/gateway/kubernetes.go
```diff
@@ -1099,6 +1104,84 @@ func findMatchingHostnames(listenerHostname *gatev1.Hostname, routeHostnames []g
 	return matches, len(matches) > 0
 }
 
+func (p *Provider) collectAttachedRoutes(ctx context.Context, gateways []gatewayWithListeners) attachedRoutes {
```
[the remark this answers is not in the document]
AnatoleLucet: fixed with the new `attachmentForListener` fn.

pkg/provider/kubernetes/gateway/attachment.go
AnatoleLucet: we generally keep methods and (some) fields public on private types :slightly_smiling_face:

kevinpollet (approved):

gndz07 (approved):
LGTM

PR #13444, opened 2026-06-30
Fix(tracing): use X-Forwarded-For for client.address when trusted

### LINKED ISSUE #13408: Tracing attribute `client.address` use X-Forwarded-For when trusted
### Welcome!

#### When 

  - using [OTEL Tracing](https://doc.traefik.io/traefik/reference/install-configuration/observability/tracing/) 
  - and setting `entryPoints.websecure.forwardedHeaders.trustedIPs` such that the `X-Forwarded-For` headers of real requests are trusted

#### Then

I would expect the span attribute `client.address` to contain the IP present in the `X-Forwarded-For` header instead of the TCP source IP.

#### Because

Semconv states

>[1]: When observed from the server side, and when communicating through an intermediary, client.address SHOULD represent the client address behind any intermediaries, for example proxies, if it's available.

See https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/docs/attributes-registry/client.md

#### See also

- PR comment by @rtribotte, which - I think - is discussing the same thing: https://github.com/traefik/traefik/pull/10408#issuecomment-1945680323

aliamerj:
### What does this PR do?

Sets the `client.address` span attribute from a trusted `X-Forwarded-For`
header instead of always using the raw TCP peer address (`r.RemoteAddr`).
`network.peer.address`/`network.peer.port` are untouched and still always
reflect the actual TCP connection.

No new trust logic is added: `X-Forwarded-For` only reaches this code
already-trusted, since `forwardedheaders.XForwarded` strips it earlier
in the chain for untrusted peers.

### Motivation

Fixes #13408. Per OTel semantic conventions, `client.address` should
represent the real client behind a trusted intermediary when available,
but it was always set from the TCP source IP, ignoring
`entryPoints.<name>.forwardedHeaders.trustedIPs` entirely.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/observability/entrypoint_test.go
+++ b/pkg/middlewares/observability/entrypoint_test.go
@@ -18,10 +18,11 @@ func TestEntryPointMiddleware_tracing(t *testing.T) {
 	}
 
 	testCases := []struct {
-		desc       string
-		entryPoint string
-		method     string
-		expected   expected
+		desc         string
+		entryPoint   string
+		method       string
+		forwardedFor string
+		expected     expected
 	}{
 		{
 			desc:       "GET",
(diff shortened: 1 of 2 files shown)


aliamerj:
Reproduced and verified locally (not via kind, not needed for this fix —
pure HTTP/tracing logic): built the binary, ran it against a `whoami`
backend with Jaeger as the OTLP trace backend, and
`entryPoints.web.forwardedHeaders.trustedIPs: ["127.0.0.1/32"]`. Sent
`curl -4 -H "X-Forwarded-For: 203.0.113.7" http://localhost:8000/` and
confirmed in the Jaeger UI that the entrypoint's server span
(`span.kind=server`) showed `client.address=127.0.0.1` before the fix
and `client.address=203.0.113.7` after, with `network.peer.address`
correctly remaining `127.0.0.1` throughout.
Screenshot from 2026-06-30 20-37-27

rtribotte:
Hey @aliamerj,

Thanks for this contribution!

As this is a fix, could you please rebase the changes on the v3.6 branch?

aliamerj:
> Hey @aliamerj,
> 
> Thanks for this contribution!
> 
> We've set this to "review" and will take a look at the code. We'll update you once the review is done.
> 
> As this is a fix, could you please rebase the changes on the v3.6 branch?

Sure

Issue #13334, opened 2026-06-10
Double dashes in Middleware names do not work for Kubernetes Ingress annotations


lewis-jackson-bots:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

Ingress with

```yaml
metadata:
  annotations:
    traefik.ingress.kubernetes.io/router.middlewares: prod-host-rewrite--portal--abc@kubernetescrd
```

### What did you see instead?

Error logs like

```
middleware "prod-host-rewrite--portal--abc@kubernetescrd" does not exist
```

See issue here https://github.com/traefik/traefik/issues/6570.

The error log here should be way clearer. It's very hard to debug. Even if this ticket is closed, I hope that "double dashes" helps someone Googling later.

### What version of Traefik are you using?

v3.7.0

### What is your environment & configuration?

Kubernetes provider

lewis-jackson-bots:
If double dashes are going to be "normalized" into single dashes, I want to opt in to a mode where Traefik errors upon loading the Middleware with an error like:

```
Middleware name  "host-rewrite--portal--abc" not valid!
```

This would preserve any backwards compatibility for anyone who was relying on this behaviour. It's just a bad pattern to silently transform my names without telling me.

rtribotte:
Thank you for reporting this issue!
This appears to be a duplicate of #6570, which documents that Traefik normalizes resource names by replacing all non-alphanumeric characters with single dashes. When you use double dashes (`--`) in middleware names, they get normalized to single dashes, causing the lookup to fail.

To resolve this, you should either:
1. Update your middleware name to use single dashes instead of double dashes in your Kubernetes resources
2. Update the annotation reference to match the normalized name shown in the Traefik dashboard

We are keeping this issue as a documentation enhancement.

lewis-jackson-bots:
I see this as a bad pattern with Kubernetes. If we can name Middleware resources a certain way - the names should not be transformed without a warning. It's just as bad as removing dashes all together, or shifting a to b, b to c.

Documentation would help, but I reckon this should be validated and create a warning log.

rtribotte:
Thank you for your reply!

[...]

PR #13447, opened 2026-07-01
doc: clarify resource name normalization for cross-provider references

amarkdotdev:
### What does this PR do?

Documents Traefik's resource name normalization in the **Provider Namespace** section of the providers reference, next to where the `<namespace>-<name>@provider` cross-provider reference syntax is described.

It adds a warning callout explaining that Traefik normalizes resource names by replacing every sequence of non-alphanumeric characters with a single dash (`-`), so names containing special characters, including consecutive dashes (`--`), dots, or underscores, are collapsed. It gives a concrete example (a middleware `strip--prefix` in the `prod` namespace must be referenced as `prod-strip-prefix@kubernetescrd`, not `prod-strip--prefix@kubernetescrd`) and notes that referencing the non-normalized name produces the confusing `middleware "..." does not exist` error.

### Motivation

Addresses #13334. Users naming Kubernetes resources with double dashes (or other special characters) hit a cryptic `middleware "..." does not exist` error because the reference in the `router.middlewares` annotation is not normalized while the registered resource ID is. As discussed in that issue, the agreed first step is to document the normalization rules clearly (a behavior change is deferred to a future major release since it would be breaking). This PR does exactly that.

### More

- [ ] Added/updated tests

### Additional Notes

Targeted at `master` as a documentation enhancement; happy to retarget to a version branch (v3.7, etc.) if preferred.

### DIFF
--- a/docs/content/reference/install-configuration/providers/overview.md
+++ b/docs/content/reference/install-configuration/providers/overview.md
@@ -45,6 +45,20 @@ For the list of the providers names, see the [supported providers](#supported-pr
     On the other hand, if you were to declare a middleware as a Custom Resource in Kubernetes and use the non-CRD Ingress objects,
     you would have to add the Kubernetes Namespace of the middleware to the annotation like this `<middleware-namespace>-<middleware-name>@kubernetescrd`.
 
+!!! warning "Resource Name Normalization"
+
+    Traefik normalizes resource names by replacing every sequence of non-alphanumeric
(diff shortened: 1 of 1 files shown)


amarkdotdev:
Rebased onto latest `master`.

amarkdotdev:
rebased

amarkdotdev:
Maintainer ping when someone on the team has a free slot. Thanks!

amarkdotdev:
Closing this PR. Thanks for the review time.

Issue #13394, opened 2026-06-24
Error message such as "route not found for path" when running in ingress-nginx compatibility mode with Traefik 3.7.1


SohamChakraborty:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

* I am [one of the folks](https://github.com/traefik/traefik/issues/12327#issuecomment-3652879120) who need/needed the `configuration-snippet` annotation support to move to Traefik. Now that `configuration-snippet` has been supported, I tried to use Traefik in ingress-nginx compatibility mode.
* For one particular service, which has ingress written like this:
```
nginx.ingress.kubernetes.io/configuration-snippet: |
  if ($arg_db = "foo") {
    rewrite ^/write$ /api/v1 last;
  }
  if ($arg_db = "bar") {
    rewrite ^/write$ /api/v1/bar last;
  }
if ($arg_db = "baz") {
  rewrite ^/write$ /api/v1/baz last;
}
if ($arg_db = "foobar") {
  rewrite ^/write$ /api/v1/foo/bar last;
}
  return 404;
```

we are getting these messages in logs: 

```
route not found for path /api/v1/foo/bar?abc=foobar&v=foobar-1.2.3.4
```

Looking into the official documentation, I see this https://doc.traefik.io/traefik/reference/routing-configuration/kubernetes/ingress-nginx/#opt-nginx-ingress-kubernetes-ioconfiguration-snippet, which mentions the directives that are supported. 

I want to know whether `rewrite` can be supported? And if not, what should be the course of action for folks who are mid-way in the migration to Traefik? Should we enable `KubernetesCRDs` and try to write yaml that is natively supported by Traefik? Or something else? I wanted to get a feedback since this will be used in production. 

The image tag being used is:

```
image:
  tag:  v3.7.1
```

### What did you see instead?

we are getting these messages in logs: 

```
route not found for path /api/v1/foo/bar?abc=foobar&v=foobar-1.2.3.4
```

### What version of Traefik are you using?

```
~ $ wget -qO- http://localhost:8080/api/version
{"Version":"3.7.1","Codename":"langres","startDate":"2026-06-23T18:27:16.482132294Z"}~ $ 
~ $ 
```

### What is your environment & configuration?

Existing ingress-nginx template:

```yaml
[...]

PR #13448, opened 2026-07-01
Document ingress-nginx snippet rewrite behavior

michaelsam94:
### What does PR do?

Documents that ingress-nginx snippet `rewrite` directives are supported but run after Traefik has selected the router/service, so they rewrite the request forwarded upstream without restarting Traefik router matching.

Adds a regression test for a `configuration-snippet` rewrite chain that branches on `$arg_db` and preserves the original query string.

### Motivation

Related to #13394. The reported migration case depends on `configuration-snippet` rewrite behavior, and the existing documentation did not list `rewrite` as a supported snippet directive or describe the post-routing limitation.

### More

- [x] Added/updated tests

### Additional Notes

Test run:

```console
go test ./pkg/middlewares/ingressnginx/snippet
```


rtribotte:
Hey @michaelsam94,

Thanks for this contribution!

We've set this to "review" and will take a look at the changes.

PR #13449, opened 2026-07-01
Fix KV provider silently dropping slice values like caFiles when a folder-marker key exists at the same path

### LINKED ISSUE #12746: Traefik does not accept caFiles field as list in Consul kv store
### Welcome!

### What did you do?

I was migrating my traefik configuration from file provider to consul kv provider and found that `caFiles` field of `tls.options` has a different behavior in kv provider comparing to file provider. The consul kv provider one would break mtls services that requires multiple ca files.

### What did you see instead?

```yaml
tls:
  options:
    custom-mtls:
      clientAuth:
        caFiles:
          - /etc/traefik/certs/custom.crt
        clientAuthType: RequireAndVerifyClientCert
```

The same configuration in consul kv store as 
```
traefik/tls/options/custom-mtls/clientAuth/caFiles/0=/etc/traefik/certs/custom.crt
```
would result in an error `building router handler: building TLS config: invalid clientAuthType: RequireAndVerifyClientCert, CAFiles is required` and break any router with this tls options.

However, changing caFiles field to
```
traefik/tls/options/custom-mtls/clientAuth/caFiles=/etc/traefikc/certs/custom.crt
```
would somehow work but it won't accept multiple CA files anymore.

```
Version:      3.6.9
Codename:     ramequin
Go version:   go1.25.7
Built:        2026-02-23T17:12:19Z
OS/Arch:      linux/amd64
```

The example static config:
```yaml
api:
  dashboard: true
  debug: false
providers:
  consul:
    endpoints:
      - http://consul.example.com # behind a traefik reverse proxy on another server
    rootKey: traefik
    token: xxx
#  file:
#    directory: /etc/traefik/config
#    watch: true
entryPoints:
  web:
    address: :80
    http:
      redirections:
        entryPoint:
          to: websecure
          scheme: https
  websecure:
    address: ":443"
    asDefault: true
certificatesResolvers:
  letsencrypt:
    acme:
      email: xxx@example.com
      storage: /etc/traefik/letsencrypt/acme.json
      httpChallenge:
        entryPoint: web
```

The minimum dynamic config:
```yaml
http:
  routers:
    dashboard:
      entryPoints:
        - websecure
      rule: Host(`traefik.example.com`)
      service: api@internal
      tls:
        certResolver: letsencrypt
        options: custom-mtls
```

_No response_

aliamerj:
### What does this PR do?

Skips empty-value KV keys ending in `/` before the KV provider buckets keys for slice decoding, so a folder-marker key (e.g. `.../caFiles/`) can't collide with the real indexed slice key (`.../caFiles/0`) at the same tree path.

### Motivation

Fixes #12746. When an empty-value key sits at the parent path of an indexed slice, `decodeToNode` collapses trailing separators and treats both keys as the same tree node. Since pairs are processed in sorted order, the empty marker is applied last and silently overwrites the real value with `""`. For `ClientAuth.CAFiles` this breaks mTLS with `CAFiles is required`, even though the underlying data looks fine.

This kind of marker key isn't specific to one workflow — it shows up whenever a nested KV path gets built one level at a time (Consul UI folder creation, `consul kv put path/ ""`, Terraform, import/export tooling), since they all write the same empty-value/trailing-slash shape through Consul's KV API. Verified the CLI and raw HTTP API produce identical output, so the fix isn't tied to any particular tool.

### More

- [x] Added/updated tests

### Additional Notes

Reproduced end-to-end against a real Consul instance and Traefik binary, before and after the fix. Existing `pkg/config/kv` and `pkg/provider/kv` test suites pass unchanged.

### DIFF
--- a/pkg/config/kv/kv_node.go
+++ b/pkg/config/kv/kv_node.go
@@ -85,6 +85,10 @@ func filterPairs(pairs []*store.KVPair, filters []string) []*store.KVPair {
 	slicePairs := map[string][]string{}
 
 	for _, pair := range pairs {
+		if strings.HasSuffix(pair.Key, "/") && len(pair.Value) == 0 {
+			continue
+		}
+
 		if len(filters) == 0 {
 			// Slice of simple type
 			if exp.MatchString(pair.Key) {
--- a/pkg/config/kv/kv_node_test.go
+++ b/pkg/config/kv/kv_node_test.go
@@ -241,6 +241,54 @@ func TestDecodeToNode(t *testing.T) {
 				},
 			}},
 		},
+		{
+			desc: "several entries, slice syntax with empty-value folder marker",
+			in: map[string]string{
+				"traefik/foo/caFiles/":  "",
+				"traefik/foo/caFiles/0": "bar0",
+				"traefik/foo/caFiles/1": "bar1",
+			},
+			expected: expected{node: &parser.Node{
+				Name: "traefik",
+				Children: []*parser.Node{
+					{Name: "foo", Children: []*parser.Node{
+						{Name: "caFiles", Value: "bar0,bar1"},
+					}},
+				},
+			}},
+		},
+		{
+			desc: "single entry, slice syntax with empty-value folder marker",
+			in: map[string]string{
+				"traefik/foo/caFiles/":  "",
+				"traefik/foo/caFiles/0": "bar0",
+			},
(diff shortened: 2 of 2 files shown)


kevinpollet:
Hey @aliamerj,

Thanks for this contribution!

As this pull request is a fix, could you rebase it on branch v3.6?

PR #13453, opened 2026-07-02
Fix(ingress-nginx): respect explicit ingress classes

Andrei-Predoiu:
### What does this PR do?

This PR fixes a bug in the `kubernetesIngressNGINX` provider where it incorrectly processes Ingress resources belonging to other controllers (such as `traefik`) when `WatchIngressWithoutClass` is set to `true`.

### Motivation

When `WatchIngressWithoutClass: true` is configured, the Ingress-NGINX provider is intended to process Ingresses that do not explicitly define an `ingressClassName` or `kubernetes.io/ingress.class` annotation. However, due to a logic flaw in the `shouldProcess` method, the provider was ignoring the explicitly defined class on the Ingress and incorrectly adopting resources explicitly targeted at other controllers.

This caused routing conflicts and overlapping configurations when running multiple Ingress controllers in the same cluster. 

This fix ensures that if an Ingress resource explicitly defines a class, and that class does not match the provider's configured NGINX ingress class, the resource will be correctly ignored regardless of the `WatchIngressWithoutClass` setting.

### More

- [x] Added/updated tests

### Additional Notes

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


pkg/provider/kubernetes/ingress-nginx/build.go
```diff
@@ -955,15 +955,15 @@ func (p *Provider) resolveHTTPErrorBackend(namespace string, cfg IngressConfig)
 func (p *Provider) shouldProcess(ing *netv1.Ingress, ingressClasses []*netv1.IngressClass) bool {
 	if len(ingressClasses) > 0 && ing.Spec.IngressClassName != nil {
 		return slices.ContainsFunc(ingressClasses, func(ic *netv1.IngressClass) bool {
-			return *ing.Spec.IngressClassName == ic.ObjectMeta.Name
```
nandorKollar, line 958: This change is not strictly necessary to fix the problem, is it?

nandorKollar (commented):
Overall, LGTM. It would be nice to add a test case to `kubernetes_test.go` which covers this scenario, though.

pkg/provider/kubernetes/ingress-nginx/build.go
Andrei-Predoiu: No, just saw it as a stylistic improvement. I can revert it if you prefer.  @nandorKollar let me know.

kevinpollet:
Hey @Andrei-Predoiu,

As this pull request is a fix, could you rebase this pull request on branch v3.6?

PR #13454, opened 2026-07-02
fix(ingress-nginx): preserve query string in app-root variable interpolation

mmatur:
### What does this PR do?

Fixes `$args`, `$query_string` and `$is_args` variable resolution dropping the query string in `app-root` redirects. When a middleware earlier in the pipeline rebuilds `req.URL`, `RawQuery` is left empty. The resolver now falls back to parsing the query from `req.RequestURI`, which is set directly from the wire and is never modified by the pipeline.

### Motivation

Annotations such as `nginx.ingress.kubernetes.io/app-root: /login$is_args$args` are expected to carry the original query string into the redirect target. Because `req.URL.RawQuery` can be cleared by upstream middleware, `/?foo=bar` was silently redirected to `/login` instead of `/login?foo=bar`.

### More

- [x] Added/updated tests

### Additional Notes

Four new test cases cover the fallback by deliberately clearing `req.URL.RawQuery` while keeping `req.RequestURI` intact, reproducing the middleware-rebuild scenario.

### DIFF
--- a/pkg/middlewares/ingressnginx/interpolation.go
+++ b/pkg/middlewares/ingressnginx/interpolation.go
@@ -120,7 +120,15 @@ func variableValue(rawVariable, variable string, req *http.Request, responseHead
 		return "http", nil
 
 	case args, queryString:
-		return req.URL.RawQuery, nil
+		if req.URL.RawQuery != "" {
+			return req.URL.RawQuery, nil
+		}
+		// Fallback: req.RequestURI is immutable from the wire and retains the query string
+		// even when req.URL is rebuilt by middleware without copying RawQuery.
+		if _, q, found := strings.Cut(req.RequestURI, "?"); found {
+			return q, nil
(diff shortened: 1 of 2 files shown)


sdelicata:
I was unable to reproduce the issue.
After investigating, I have the impression that the issue is already solved by https://github.com/traefik/traefik/pull/12986
To me, we don't need anything else. WDYT @mmatur ?

mmatur:
Hi @sdelicata,

Good catch 👍 After digging into it, the issue was on my side: my e2e runs were using stale experimental-v3.7 / experimental-master images built before #13398 was merged, which is the PR that actually fixed the $is_args$args interpolation.

I re-ran the full AppRoot e2e suite against the current v3.7 image and everything works very well, including query string preservation.

Closing this PR as it's no longer needed. Thanks! 🙏

Issue #12931, opened 2026-04-03
ingress-nginx provider: rewrite-target full URL appends path suffix


emilevauge:
### Description

When using `rewrite-target` with a full URL (e.g. `https://bar.example.org/$1`) and a non-regex path (e.g. `/original` with `pathType: Prefix`), the ingress-nginx provider incorrectly appends the unmatched path suffix to the redirect URL.

### Expected behavior

A request to `/original/other` should redirect to `https://bar.example.org/` (the `$1` capture group is empty since the path has no regex).

This is the behavior of the real nginx ingress controller.

### Actual behavior

A request to `/original/other` redirects to `https://bar.example.org//other` (double slash + unmatched suffix).

### Root cause

In `pkg/middlewares/ingressnginx/rewritetarget/rewrite_target.go:73`:

```go
newTarget = rt.regexp.ReplaceAllString(currentPath, rt.replacement)
```

The regex is `(?i)/original` (set in `pkg/provider/kubernetes/ingress-nginx/kubernetes.go:1593`). When applied to `/original/other`:

1. `ReplaceAllString` matches `/original` (positions 0-9)
2. Replaces it with `https://bar.example.org/` (`$1` → empty, no capture group)
3. **Keeps the unmatched suffix** `/other`
4. Result: `https://bar.example.org/` + `/other` = `https://bar.example.org//other`

### Ingress configuration

```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: "https://bar.example.org/$1"
spec:
  rules:
    - host: example.com
      http:
        paths:
          - path: /original
            pathType: Prefix
            backend:
              service:
                name: backend
                port:
                  number: 80
```

### Possible fix

When the replacement resolves to a full URL, the regex should match the entire path to avoid suffix leaking. In `kubernetes.go:1593`, append `.*` when the rewrite-target is a full URL:

```go
regex := "(?i)" + rulePath
if isFullURL(rewrite) {
    regex += ".*"
}
```

Or alternatively, in the middleware itself, skip `ReplaceAllString` when the replacement is an absolute URL and the regex has no capture groups.

rtribotte:
Duplicate of https://github.com/traefik/traefik/issues/12911

PR #13455, opened 2026-07-02
Make trailing capture group optional in ImplementationSpecific paths

mmatur:
### What does this PR do?

Introduces `makeTrailingGroupOptional`, which rewrites an `ImplementationSpecific` path like `/original/(.*)` to `/original(?:/(.*))? ` so that the bare path `/original` also matches. The transformation applies to both the router `PathRegexp` rule and the `RewriteTarget` regex. Paths whose first character is a capture group are left unchanged.

### Motivation

The nginx ingress controller handles the trailing-slash edge case implicitly. Without this fix, a request to `/original` returned 404 because the generated `PathRegexp` required a literal `/` after the prefix.

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/middleware.go
+++ b/pkg/provider/kubernetes/ingress-nginx/middleware.go
@@ -14,6 +14,7 @@ import (
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
 	"github.com/traefik/traefik/v3/pkg/tls"
 	"github.com/traefik/traefik/v3/pkg/types"
+	netv1 "k8s.io/api/networking/v1"
 	"k8s.io/utils/ptr"
 )
 
@@ -278,6 +279,9 @@ func (p *Provider) buildRewriteTarget(loc *location) {
 	}
 
 	regex := loc.Path
+	if ptr.Deref(loc.PathType, netv1.PathTypePrefix) == netv1.PathTypeImplementationSpecific {
+		regex = makeTrailingGroupOptional(loc.Path)
+	}
 
 	xfp := ""
 	if loc.Config.XForwardedPrefix != nil && regexPathWithCapture.MatchString(*loc.Config.XForwardedPrefix) {
--- a/pkg/provider/kubernetes/ingress-nginx/middleware.go
+++ b/pkg/provider/kubernetes/ingress-nginx/middleware.go
@@ -279,7 +279,7 @@ func (p *Provider) buildRewriteTarget(loc *location) {
 	}
 
 	regex := loc.Path
-	if ptr.Deref(loc.PathType, netv1.PathTypePrefix) == netv1.PathTypeImplementationSpecific {
+	if ptr.Deref(loc.PathType, netv1.PathTypePrefix) == netv1.PathTypeImplementationSpecific && hasAbsoluteRewriteTarget(loc) {
 		regex = makeTrailingGroupOptional(loc.Path)
 	}
 
--- a/pkg/provider/kubernetes/ingress-nginx/translator.go
+++ b/pkg/provider/kubernetes/ingress-nginx/translator.go
@@ -6,6 +6,7 @@ import (
 	"maps"
 	"math"
 	"net/http"
+	"net/url"
 	"regexp"
 	"slices"
 	"strings"
@@ -663,7 +664,9 @@ func buildRule(host string, loc *location) string {
 		regexPath := loc.Path
 		if pathType == netv1.PathTypeImplementationSpecific {
 			pathType = netv1.PathTypePrefix
-			regexPath = makeTrailingGroupOptional(loc.Path)
+			if hasAbsoluteRewriteTarget(loc) {
+				regexPath = makeTrailingGroupOptional(loc.Path)
+			}
 		}
 
 		switch pathType {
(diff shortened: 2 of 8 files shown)


pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -620,3 +623,18 @@ func buildPrefixRule(path string) string {
 	path = strings.TrimSuffix(path, "/")
 	return fmt.Sprintf("(Path(%q) || PathPrefix(%q))", path, path+"/")
 }
+
+// makeTrailingGroupOptional converts an ImplementationSpecific regex path like /foo/(.*)
+// into /foo(?:/(.*))? so it also matches the bare /foo (without trailing slash).
+// Only applies when the path starts with a literal prefix (not a capture group),
+// i.e. when the path does not begin with "/(".
+func makeTrailingGroupOptional(path string) string {
```
sdelicata, line 669: Could you add a small table-driven test for `makeTrailingGroupOptional` directly? A few edge cases would be nice to lock down: `/(.*)` left untouched, multiple groups like `/foo/(bar)/(.*)`, and whatever we decide for the `/originalfoo` case above.

pkg/provider/kubernetes/ingress-nginx/middleware.go
```diff
@@ -261,6 +262,9 @@ func (p *Provider) buildRewriteTarget(loc *location) {
 	}
 
 	regex := loc.Path
+	if ptr.Deref(loc.PathType, netv1.PathTypePrefix) == netv1.PathTypeImplementationSpecific {
+		regex = makeTrailingGroupOptional(loc.Path)
```
sdelicata, line 276: Maybe I'm missing something, but I think this reopens the suffix leak from #12931 for a request like `/originalfoo` (the prefix immediately followed by non-slash chars).

The router rule `^/original(?:/(.*))?` matches `/originalfoo`, so it gets routed to the rewrite middleware, and there `ReplaceAllString` only matches `/original` (the optional group fails on the `f`) and leaks the trailing `foo`, so we redirect to `https://bar.example.org/foo`. nginx routes that same request to its implicit `^/` catch-all and redirects to `https://bar.example.org/`. And before this PR `/originalfoo` was a 404, so it's a behavior change too.

The #12931 guard that appends `.*` to consume the whole path lives in `rewritetarget.New` (the `if re.NumSubexp() == 0 && absoluteURLRedirect` branch in `pkg/middlewares/ingressnginx/rewritetarget/rewrite_target.go`), but it doesn't fire here: `makeTrailingGroupOptional` always emits a capturing group, so `NumSubexp()` is 1, not 0.

Relaxing that condition to also append `.*` in the absolute-URL case when there are capture groups fixes it, and the captures stay intact:
`/originalfoo` -> `.../`, `/original/foo` -> `.../foo`, `/original` -> `.../`.
WDYT?

sdelicata (changes requested):

pkg/provider/kubernetes/ingress-nginx/middleware.go
mmatur: You are right, I am moving with this approach. By the way I have added some test on [ingress-nginx-migration](https://github.com/traefik/ingress-nginx-migration/pull/33)

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -620,3 +623,36 @@ func buildPrefixRule(path string) string {
...
+			if depth == 0 && i != len(path)-1 {
+				return path
+			}
+		}
+	}
+	if depth != 0 {
+		return path
+	}
+
+	return path[:idx] + "(?:" + path[idx:] + ")?"
```
sdelicata, line 694: Maybe I'm wrong, but I think `/originalfoo` still diverges for the non-absolute rewrite case (the one the new `ingress-with-rewrite-target-trailing-group.yml` fixture generates).

`^/original(?:/(.*))?` matches `/originalfoo`, and in the middleware `ReplaceAllString` only matches `/original`, so `/newpath/$1` gives `/newpath/` plus the leftover `foo`, and we proxy `/newpath/foo` upstream. The `.*` guard fixed this for absolute URLs only. nginx wouldn't match `/originalfoo` at all with `location ~* "^/original/(.*)"`, and before this PR it was a 404, so it's a behavior change too.

WDYT about emitting `/original(?:$|/(.*))` instead? It's the pattern ingress-nginx documents itself in https://kubernetes.github.io/ingress-nginx/examples/rewrite/ (`/something(/|$)(.*)`), the capture indexes stay the same, `/original` and `/original/foo` still match, and `/originalfoo` doesn't.

sdelicata (changes requested):

pkg/provider/kubernetes/ingress-nginx/translator.go
mmatur: Good catch, the non-absolute case was indeed still leaking.

I tried your pattern first, but the `ingress-nginx-migration` e2e suite caught a regression on the absolute-URL case: for `/originalfoo`, nginx answers `302 https://bar.example.org/` (the request falls through to the implicit `^/` catch-all, whose rewrite still issues the redirect), while `/original(?:$|/(.*))` makes Traefik answer 404.

It turns out nginx behaves differently depending on the rewrite-target:

- absolute URL: unmatched paths hit the catch-all and still get the 302 redirect
- path rewrite: unmatched paths hit the catch-all and get a 404 from the default backend

So instead of changing the pattern, I now only apply `makeTrailingGroupOptional` when the rewrite-target is an absolute URL. For the non-absolute case the regex path is kept as-is, so `/original` and `/originalfoo` return 404 exactly like nginx, and the suffix leak is gone without touching the middleware.

With this, the whole `TestRewriteTargetSuite` [e2e suite](https://github.com/traefik/ingress-nginx-migration/pull/33) passes against a real ingress-nginx, including the `/originalfoo` cases in both modes.

sdelicata (approved):
Thanks for the changes!
LGTM

juliens (approved):

PR #13456, opened 2026-07-02
Preserve leading dot in sticky session cookie Domain attribute

mmatur:
### What does this PR do?

Preserves the leading dot in the `Domain` attribute of sticky session cookies when `session-cookie-domain` starts with `.`. Go's `http.SetCookie` silently strips the leading dot per RFC 6265 §4.1.2.3. The fix serializes the cookie with `cookie.String()`, restores the dot, then writes the header directly via `rw.Header().Add`.

### Motivation

nginx-ingress-controller emits `Domain=.example.com` as-is. Traefik was emitting `Domain=example.com`, so clients carrying a cookie set by nginx-ingress during a migration window would lose session affinity immediately.

### More

- [x] Added/updated tests

### Additional Notes

Only domains that explicitly start with `.` go through this path. Everything else continues to use `http.SetCookie` unchanged.

### DIFF
--- a/pkg/server/service/loadbalancer/sticky.go
+++ b/pkg/server/service/loadbalancer/sticky.go
@@ -147,7 +147,15 @@ func (s *Sticky) WriteStickyCookie(rw http.ResponseWriter, name string) error {
 	if s.cookie.expires > 0 {
 		cookie.Expires = time.Now().Add(s.cookie.expires)
 	}
-	http.SetCookie(rw, cookie)
+
+	cookieStr := cookie.String()
+	if strings.HasPrefix(s.cookie.domain, ".") {
+		// Go stdlib http.SetCookie strips the leading dot from the Domain attribute per RFC 6265.
+		// nginx-ingress-controller preserves it (Domain=.example.com), so we restore it here
+		// to maintain migration compatibility.
+		cookieStr = strings.Replace(cookieStr, "Domain="+s.cookie.domain[1:], "Domain="+s.cookie.domain, 1)
+	}
+	rw.Header().Add("Set-Cookie", cookieStr)
 
 	return nil
 }
--- a/pkg/server/service/loadbalancer/sticky.go
+++ b/pkg/server/service/loadbalancer/sticky.go
@@ -24,14 +24,15 @@ type NamedHandler struct {
 
 // stickyCookie represents a sticky cookie.
 type stickyCookie struct {
-	name     string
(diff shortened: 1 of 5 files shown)


pkg/server/service/loadbalancer/sticky.go
```diff
@@ -145,7 +145,15 @@ func (s *Sticky) WriteStickyCookie(rw http.ResponseWriter, name string) error {
...
 	}
-	http.SetCookie(rw, cookie)
+
+	cookieStr := cookie.String()
+	if strings.HasPrefix(s.cookie.domain, ".") {
+		// Go stdlib http.SetCookie strips the leading dot from the Domain attribute per RFC 6265.
+		// nginx-ingress-controller preserves it (Domain=.example.com), so we restore it here
+		// to maintain migration compatibility.
+		cookieStr = strings.Replace(cookieStr, "Domain="+s.cookie.domain[1:], "Domain="+s.cookie.domain, 1)
+	}
```
sdelicata: Maybe it's worth flagging that, since this sits in the shared load balancer sticky path, the leading dot is now preserved for every provider, not just the ingress-nginx one. Is it intentional? Should we consider it as breaking?
Apart from this consideration, the fix LGTM.

sdelicata (changes requested):

pkg/server/service/loadbalancer/sticky_test.go
```diff
@@ -139,6 +139,47 @@ func TestSticky_WriteStickyCookie(t *testing.T) {
...
+
+	for _, test := range testCases {
+		t.Run(test.desc, func(t *testing.T) {
+			t.Parallel()
+
+			cookieConfig := dynamic.Cookie{
+				Name:               "test",
+				Domain:             ".example.com",
+				PreserveLeadingDot: test.preserveLeadingDot,
+			}
```
sdelicata, line 176: Maybe it's worth setting all the attributes here, like in `TestSticky_WriteStickyCookie` just above?

IMO, the risk with the string replace is damaging the rest of the header, and with only Name+Domain set we're testing the shortest possible cookie string. An `assert.Equal` on the full `Set-Cookie` value (minus `Expires`) would highlight it. WDYT?

sdelicata (approved):
Just a nitpick regarding tests.
LGTM anyway.

kevinpollet (approved):

PR #13457, opened 2026-07-02
Create HTTP redirect router for ssl-passthrough with force-ssl-redirect

mmatur:
### What does this PR do?

Emits an HTTP router with a `RedirectScheme` middleware (308) when an ingress sets both `ssl-passthrough: "true"` and `force-ssl-redirect: "true"`. The `force-ssl-redirect` flag is now carried through `sslPassthroughBackend` and checked in `translator.go` after the passthrough TCP router is created.

### Motivation

The ssl-passthrough code path in `build.go` hits a `continue` before the main annotation-processing loop, so `force-ssl-redirect` was never applied to passthrough ingresses. HTTP requests to those hosts returned 404 instead of a redirect to HTTPS.

### More

- [ ] Added/updated tests

### Additional Notes

The HTTP redirect router is wired to `noop@internal` — no real upstream is needed since the `RedirectScheme` middleware short-circuits the response before it reaches a backend.

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/build.go
+++ b/pkg/provider/kubernetes/ingress-nginx/build.go
@@ -265,7 +265,7 @@ func (p *Provider) build(ctx context.Context, ingressClasses []*netv1.IngressCla
 			Logger()
 		ctxIng := logger.WithContext(ctx)
 
-		// ssl-passthrough: handled per-rule. serversTransport is not needed for passthrough.
+		// ssl-passthrough: handled per-rule.
 		if ptr.Deref(ing.config.SSLPassthrough, false) {
 			// Even with ssl-passthrough, the Spec.TLS section's certificates are still loaded so they remain available as the default certificate.
 			if len(ing.Spec.TLS) > 0 {
@@ -311,12 +311,22 @@ func (p *Provider) build(ctx context.Context, ingressClasses []*netv1.IngressCla
 					}
 				}
 
+				nst, err := p.buildServersTransport(ctxIng, ing.Namespace, ing.Name, ing.config)
+				if err != nil {
+					logger.Error().Err(err).Msgf("Cannot build serversTransport for ssl-passthrough on host %q", rule.Host)
+					continue
+				}
+
 				routerKey := strings.TrimPrefix(provider.Normalize(ing.Namespace+"-"+ing.Name+"-"+rule.Host), "-")
 				mc.PassthroughBackends = append(mc.PassthroughBackends, &sslPassthroughBackend{
-					BackendName:      ptBackendName,
-					Hostname:         rule.Host,
-					RouterKey:        routerKey,
-					ForceSSLRedirect: ptr.Deref(ing.config.ForceSSLRedirect, false),
+					BackendName:          ptBackendName,
+					Hostname:             rule.Host,
+					RouterKey:            routerKey,
+					ForceSSLRedirect:     ptr.Deref(ing.config.ForceSSLRedirect, false),
+					HTTPServiceName:      provider.Normalize(ing.Namespace + "-" + ing.Name + "-" + ingBackend.Service.Name + "-" + portString(ingBackend.Service.Port)),
+					ServersTransportName: nst.name,
+					ServersTransport:     nst.ServersTransport,
+					Config:               ing.config,
 				})
 				markProcessedIngress(ing.Ingress)
 			}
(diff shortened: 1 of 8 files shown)


nandorKollar:
Should we add an extra test case to our fixtures with an ingress using both annotations to cover this scenario?

nandorKollar (approved):
LGTM

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -122,6 +122,20 @@ func (p *Provider) translate(ctx context.Context, mc *model) *dynamic.Configurat
 			Service:     pt.BackendName,
 			TLS:         &dynamic.RouterTCPTLSConfig{Passthrough: true},
 		}
+
+		if pt.ForceSSLRedirect {
+			redirectMWName := pt.RouterKey + "-redirect-scheme"
+			conf.HTTP.Middlewares[redirectMWName] = &dynamic.Middleware{
+				RedirectScheme: &dynamic.RedirectScheme{Scheme: "https", ForcePermanentRedirect: true},
+			}
+			conf.HTTP.Routers[pt.RouterKey+"-http"] = &dynamic.Router{
```
sdelicata: Maybe I miss something, but the sibling `force-ssl-redirect` path (the `Force SSL Redirect without Ingress TLS section` regression test in `kubernetes_test.go:2396`) deliberately keeps the real backend service here, so a request arriving with `X-Forwarded-Proto: https` doesn't bypass the redirect and hit noop with a 418. 
This one points to `noop@internal`, so the same request would get a 418. 
Probably fine since a passthrough host has no HTTP backend to fall back to anyway, but is it intentional? Maybe worth a one-line comment referencing that regression rationale so nobody wonders why the two paths differ. WDYT?

sdelicata (changes requested):

pkg/provider/kubernetes/ingress-nginx/translator.go
mmatur: Good catch! I compared the actual behaviors side by side (traefik vs ingress-nginx e2e), and the `noop@internal` shortcut was indeed a real divergence, not just a cosmetic one:

- HTTP request with `X-Forwarded-Proto: https` on a passthrough host with `force-ssl-redirect`: nginx skips the redirect and proxies the request to the backend (over TLS with `backend-protocol: HTTPS`) → `200`, while we returned a `418` from `noop@internal`.
- Plain HTTP request without `force-ssl-redirect`: nginx keeps an HTTP server for the passthrough host and proxies to the TLS backend, which rejects it with `400`, while we returned a `404`.

So I reworked the fix: every passthrough host now gets an HTTP router pointing to a real service built from the backend (honoring `backend-protocol`), with the redirect-scheme middleware only added when `force-ssl-redirect` is set. The service is named per ingress (like the regular path) since `backend-protocol` is a per-ingress annotation.

All three cases now match nginx, verified e2e: `308` on plain HTTP with `force-ssl-redirect`, `200` when the redirect is bypassed via `X-Forwarded-Proto`, and `400` on plain HTTP without the redirect.

pkg/provider/kubernetes/ingress-nginx/build.go
```diff
@@ -310,11 +310,22 @@ func (p *Provider) build(ctx context.Context, ingressClasses []*netv1.IngressCla
 					}
 				}
 
+				nst, err := p.buildServersTransport(ctxIng, ing.Namespace, ing.Name, ing.config)
+				if err != nil {
+					logger.Error().Err(err).Msgf("Cannot build serversTransport for ssl-passthrough on host %q", rule.Host)
+					continue
```
sdelicata: I'm afraid this continue also kills the TCP passthrough router. Since we skip the rule entirely, nothing is appended to `mc.PassthroughBackends`.

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -122,6 +122,34 @@ func (p *Provider) translate(ctx context.Context, mc *model) *dynamic.Configurat
...
+		conf.HTTP.Services[pt.HTTPServiceName] = buildServiceWithLocConfig(backend, pt.ServersTransportName, pt.Config)
+
+		rt := &dynamic.Router{
+			EntryPoints: p.NonTLSEntryPoints,
+			Rule:        fmt.Sprintf("Host(%q)", pt.Hostname),
+			RuleSyntax:  "default",
+			Service:     pt.HTTPServiceName,
+		}
+
+		if pt.ForceSSLRedirect {
```
sdelicata: Maybe I miss something, but I have the feeling that we only implement the first half of the redirect semantics compared to `buildSSLRedirect` in middleware.go.

sdelicata (changes requested):

sdelicata (approved):
LGTM

traefiker:

PR #13458, opened 2026-07-02
Fix cross-provider ref check for TCP ServersTransport in Kubernetes CRD provider

gndz07:
### What does this PR do?

Adds the missing `crossProviderNamespaces` allowlist check to `makeTCPServersTransportKey` in the Kubernetes CRD provider.

### Motivation

Follow up to https://github.com/traefik/traefik/pull/13094.
To align the `crossProviderNamespaces` check behavior; this allowlist is enforced for HTTP `serversTransport`, TLS options, and middlewares on both HTTP and TCP, but the TCP `serversTransport` path never got the same check.

### More

- [x] Added/updated tests


rtribotte (approved):
LGTM

PR #13459, opened 2026-07-03
Supports configuring a minimum number of requests for circuit breaker

### LINKED ISSUE #10963: circuit breaker supports configuring a minimum number of requests
### Welcome!

When using Traefik, we noticed that the circuit breaker functionality in the middleware is implemented using cbreaker. However, currently, cbreaker does not support configuring a minimum number of requests to determine whether to trip the circuit breaker, a feature that is commonly provided in popular circuit breaker components.
For instance, in Hystrix, there is a configuration parameter called requestVolumeThreshold (defaulting to 20), which stipulates that Hystrix will only begin calculating failure rates and decide whether to trip the circuit based on this rate when the number of requests reaches or exceeds this threshold within the sliding time window. Even if all requests fail within the sliding window, if the number of requests is below this threshold, the circuit breaker will not open.
We would like to know if there are plans to add this feature to Traefik's circuit breaker functionality, or if there are plans to further enhance the capabilities of circuit breaking and rate limiting, such as providing cluster-wide rate limiting, rate limiting based on specific request headers, etc.

hanxiantao:
### What does this PR do?

Adds support for the `RequestThreshold()` circuit breaker expression by upgrading `github.com/vulcand/oxy/v2` to `v2.2.0`.

This allows users to combine the existing circuit breaker metrics with a minimum request count condition, for example:

```yaml
expression: "ResponseCodeRatio(500, 600, 0, 600) > 0.30 && RequestThreshold() >= 20"
```

Fixes #10963

### Motivation

Traefik's circuit breaker previously could trip based on ratios even when only a small number of requests had been recorded. This change enables users to require a minimum number of requests before the circuit breaker opens, matching the behavior requested in #10963 and the RequestThreshold() support added in vulcand/oxy#256.

### More

- [x] Added/updated tests

### Additional Notes

The new threshold is exposed through the existing expression option. No new Traefik configuration field is added.

### DIFF
--- a/docs/content/reference/routing-configuration/http/middlewares/circuitbreaker.md
+++ b/docs/content/reference/routing-configuration/http/middlewares/circuitbreaker.md
@@ -73,13 +73,14 @@ spec:
 
 ### expression
 
-The `expression` option can check three different metrics:
+The `expression` option can check four different metrics:
 
 | Metrics | Description | Example |
 |:------|:------------|:--------|
 | <a id="opt-NetworkErrorRatio" href="#opt-NetworkErrorRatio" title="#opt-NetworkErrorRatio">`NetworkErrorRatio`</a> | The network error ratio to open the circuit breaker. | `NetworkErrorRatio() > 0.30` opens the circuit breaker at a 30% ratio of network errors |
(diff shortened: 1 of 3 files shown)


kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

Issue #12504, opened 2026-01-06
Dynamic config that is not dynamic


D0wn3r:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

Hi,

I have certificate in the dynamic directory grouped in a file named "certificates.toml".
When I add a new certificate or when I renew one, Traefik doesn't take it in consideration. To try to fix this, I execute the command `touch` on the certificates's file on each machine (with ansible).
But sometimes it's not enough, or I have to do it multiple times.

I'm using glusterfs, so I guess it can have an impact with this issue.

### What did you see instead?

no dynamic changes

### What version of Traefik are you using?

v3.5.3

### What is your environment & configuration?

```yaml
global:
  checkNewVersion: false
  sendAnonymousUsage: false

log:
  level: INFO

api:
  dashboard: true

accessLog: {}

metrics:
  prometheus:
    entryPoint: websecure
    manualRouting: true
    addRoutersLabels: true
    addServicesLabels: true

entryPoints:
  web:
    address: :80
    http:
      redirections:
        entryPoint:
          to: websecure
          scheme: https
  websecure:
    address: :443
    http3: {}
    proxyProtocol:
      trustedIPs: 192.168.0.8

providers:
  swarm:
    endpoint: "tcp://dockerproxy:2375"
    exposedByDefault: false
  file:
    directory: "/etc/traefik/dynamic"
    watch: tru
```

nmengin:
Hey @D0wn3r,

Thanks for reaching out! Could you share a minimal reproducible case? (For example, full Kubernetes manifests or a simple setup that shows the issue?)
If anyone else can help reproduce or fix this, we'd love the help!

giorgiga:
I noticed this too (this or something similar...).

I didn't report because I'm not 100% sure of the details (the problem disappears restarting and every time I keep repeating "ok, let's just restart. But next time I'll pay proper attention and report this").

I have the certs managed by the nixos' "security.acme" system (which is built around lego) and so I generate a dynamic configuration file with the paths where the certs/keys, once generated, will be stored.

When I reboot after adding a new service, traefik starts before the security.acme service managed to ready the keys/certs and so the files mentioned in the dynamic config don't exist.

[...]

PR #13463, opened 2026-07-04
fix: watch externally referenced certificate files in the file provider

aliamerj:
### What does this PR do?

The file provider watches the dynamic config file(s) for changes, but never
watched the certificate, key, or CA files those configs reference. Renewing
a certificate in place never triggered a reload, the only way was to touch
the dynamic config file itself.

This adds automatic watching of every certificate/key/CA path referenced
from the dynamic configuration, kept in sync as the configuration changes.

### Motivation

#12504 reports two related symptoms of the same root cause: a renewed
certificate isn't picked up without manually touching the config file, and
a certificate that doesn't exist yet at startup is never noticed once it's
issued. Both come down to the same thing — the file provider was never
watching the certificate files at all, only the dynamic config around them.

### More

- [x] Added/updated tests

### Additional Notes

Manually reproduced and verified against both scenarios from the issue
(in-place renewal and write-then-rename renewal, run repeatedly; delete
and recreate a referenced cert with no config file changes) before and
after this fix. `go test ./pkg/provider/file/...` passes.

You may see a transient `private key does not match public key` warning
if a certificate and its key are renewed via two separate rename operations
a few seconds apart — that's an existing, expected consequence of reloading
on every change, not something introduced by this PR. It self-corrects the
moment the second file lands.

### DIFF
--- a/pkg/provider/file/file.go
+++ b/pkg/provider/file/file.go
@@ -254,7 +254,7 @@ func (p *Provider) isRelevantEvent(filename string) bool {
 
 // syncExternalFileWatches updates the fsnotify watcher so that it watches the parent
 // directories of exactly the external files (certificates, keys, CA bundles) currently
-// referenced by the dynamic configuration
+// referenced by the dynamic configuration.
 func (p *Provider) syncExternalFileWatches(refFiles []string) {
 	p.watcherMu.Lock()
 	defer p.watcherMu.Unlock()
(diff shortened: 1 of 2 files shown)


pkg/provider/file/file.go
```diff
@@ -212,39 +225,117 @@ func (p *Provider) addWatcher(pool *safe.Pool, items []string, configurationChan
...
+		_, confFileName := filepath.Split(p.Filename)
+		if evtFileName == confFileName {
+			return true
+		}
+	}
+
+	p.watcherMu.Lock()
+	defer p.watcherMu.Unlock()
+
+	_, ok := p.externalDirs[filepath.Dir(filename)]
```
nandorKollar, line 251: nit: since this is a read-only operation, you might want to consider replacing the mutex with a read-write lock. It probably won't make a big difference though, as certificate file changes are likely infrequent.

kevinpollet:
Hey @aliamerj,

As this pull request is a fix, could you please rebase it on branch v3.6?

Issue #13464, opened 2026-07-05
docs: warn that YAML parse errors drop entire file provider configuration


magnus919:
## Summary

The File Provider documentation at `docs/content/reference/install-configuration/providers/others/file.md` warns about filesystem notification issues with mounted volumes, but does not warn about a more dangerous failure mode: **when a single YAML/TOML file has a parse error, Traefik silently discards the entire provider's configuration** — every router, middleware, and service from all files in the directory disappears.

## Real-World Impact

I hit this on a production deployment using `providers.file.directory` with ~40 service files. A single config file had a subtle indentation issue. Python's `yaml.safe_load()` accepted the file as valid, but Traefik's parser rejected it. The result: every middleware reference across all services failed simultaneously with "middleware does not exist" errors. The only signal was a burst of identical errors at the same timestamp in the access log. Recovery required restoring a backup of the entire config directory.

The current "Limitations" admonition only discusses filesystem notification quirks. The parse-error-drop behavior is undocumented and operationally dangerous.

## Proposed Change

Add a paragraph to the "Limitations" warning in `docs/content/reference/install-configuration/providers/others/file.md`, after the existing paragraph about filesystem notification issues:

```markdown
    If a configuration file contains a YAML or TOML syntax error, Traefik
    discards **all** configuration from the file provider — every router,
    middleware, and service from the entire directory disappears, not just
    the broken file. Validate configuration files before deploying
    (e.g., using `traefik healthcheck` or an external YAML/TOML validator),
    and keep backups of known-good configurations.
```

## Why This Matters

- The failure mode is silent (no startup error in most cases — it happens on hot-reload)
- Python's `yaml.safe_load()` is not a sufficient validator (Traefik's parser may reject files that pass Python's parser)
- A single typo in one service file takes down routing for *all* services
- The recovery path (restore backup) is not obvious to operators who encounter this for the first time

I'm happy to submit a PR for this if the maintainers agree it's worth documenting.

Filed by Jasper (AI agent on behalf of Magnus Hedemark)

magnus919:
Closed — filed against wrong repository.

PR #13465, opened 2026-07-05
docs: warn that YAML parse errors drop entire file provider configuration

magnus919:
### What does this PR do?

Adds a warning to the File Provider documentation about a dangerous failure mode: when a single YAML/TOML file has a syntax error, Traefik silently discards **all** configuration from the file provider — not just the broken file. Every router, middleware, and service from the entire directory is dropped.

### Motivation

This happened in production. A single config file had a subtle indentation issue that Python's `yaml.safe_load()` accepted but Traefik's parser rejected. Every middleware reference across all services failed simultaneously. The behavior is undocumented and the recovery path (restore from backup) is not obvious to operators encountering it for the first time.

Related issue: #13464

### More

- [ ] Added/updated tests — N/A (docs-only change)

### Additional Notes

The change adds 4 lines to the existing "Limitations" admonition in `docs/content/reference/install-configuration/providers/others/file.md`, after the paragraph about filesystem notification issues.


magnus919:
Closed — filed against wrong repository.

PR #13470, opened 2026-07-06
Silence encoded characters warning when some are disallowed

### LINKED ISSUE #13169: persistent spurious warning at startup: Traefik can reject some encoded characters in the request path
### Welcome!

### What did you do?

When starting Traefik with a reasonable configuration, I expect no warning to be printed. Please do not add feature announcements as persistent warnings that cannot be turned off: it confuses operators.

### What did you see instead?

When starting Traefik, it prints the following warning:
```
Traefik can reject some encoded characters in the request path.When your backend is not fully compliant with [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986),it is recommended to set these options to `false` to avoid split-view situation.Refer to the documentation for more details: https://doc.traefik.io/traefik/v3.6/migrate/v3/#encoded-characters-configuration-default-values
```

This is actually a feature announcement and there is no way to turn it off. It looks like there was a [plan to remove that warning](https://github.com/traefik/traefik/commit/1e0e03edc7f76e1e8a7c0b22baba337ce1b4a9e1) in a future release but that deprecation comment has since been [removed](https://github.com/traefik/traefik/pull/12540/changes).

3.6.9 but AFAICT the unconditional warning is present in 3.7 and master.

That is not relevant.

_No response_

Adel-Ayoub:
### What does this PR do?

The warning about encoded characters in the request path is currently logged on every startup, whatever the entrypoints configuration looks like. With this change it is only logged when no entrypoint disallows any of the encoded characters, so operators who already set some `allowEncoded*` options to `false` are not warned about something they took care of.

UDP-only configurations do not trigger the warning either, since UDP entrypoints never parse HTTP request paths.

### Motivation

Setting the options to `false` is exactly what the warning recommends, but doing so didn't make it go away. That made people think their configuration wasn't applied.

Fixes #13169

### More

- [x] Added/updated tests

### Additional Notes

Verified with a local build. With `allowEncodedSlash=false` on an entrypoint the warning is gone, and a request with `%2F` in the path is still rejected with a 400. With a default configuration the warning is still logged at startup.


rtribotte:
Hey @Adel-Ayoub,

Thanks for this contribution!

pkg/config/static/static_config.go
```diff
@@ -482,6 +482,28 @@ func (c *Configuration) ValidateConfiguration() error {
...
+		}
+
+		if !encodedCharacters.AllowEncodedSlash ||
+			!encodedCharacters.AllowEncodedBackSlash ||
+			!encodedCharacters.AllowEncodedNullCharacter ||
+			!encodedCharacters.AllowEncodedSemicolon ||
+			!encodedCharacters.AllowEncodedPercent ||
+			!encodedCharacters.AllowEncodedQuestionMark ||
+			!encodedCharacters.AllowEncodedHash {
+			return true
```
amazon7737, line 522: Just to confirm that this matches the behavior agreed on in issue.

 this method returns `true` as soon as any single entrypoint denies one encoded character. 

As a result, `runCmd` suppresses the startup warning globally, even if another entrypoint still allows all encoded characters. Is this global “any entrypoint” behavior intentional?

pkg/config/static/static_config.go
```diff
@@ -482,6 +482,28 @@ func (c *Configuration) ValidateConfiguration() error {
 	return nil
 }
 
+// HasDeniedEncodedCharacters reports whether at least one entrypoint disallows an encoded character in the request path.
+func (c *Configuration) HasDeniedEncodedCharacters() bool {
+	for _, ep := range c.EntryPoints {
+		encodedCharacters := ep.HTTP.EncodedCharacters
+		if encodedCharacters == nil {
+			continue
+		}
```
amazon7737, line 513: Could we account for UDP-only configurations here? 

When all entrypoints use the UDP protocol, `EncodedCharacters` is `nil` for every entrypoint, so this method returns `false` and `runCmd` emits an HTTP request-path warning even though no HTTP-capable entrypoint exists. 

Would it make sense to skip UDP entrypoints and suppress the warning when there are no TCP entrypoints?

pkg/config/static/static_config.go
Adel-Ayoub: Yes, intentional, that's the behavior agreed on in #13169. Once some characters are disallowed the operator clearly knows about these options, no point warning them again.

pkg/config/static/static_config.go
Adel-Ayoub: Thanks, makes sense. Added a test so a denied character on a UDP entrypoint still warns.

amazon7737 (approved):
Thanks 😄

gndz07 (approved):
LGTM, thanks 👍

PR #13476, opened 2026-07-07
Fix redirect with use-regex in IngressNGINX provider

### LINKED ISSUE #13474: Redirect Behaviour in IngressNGINX Provider Different from Ingress-Nginx
### Welcome!

### What did you do?

We have tried to use the same Ingress in both Ingress-Nginx and Traefik's IngressNGINX provider. We expected the behaviour of the two to be the same but they were different. In Traefik we can't use the regex in the paths to specify a different redirect target. 

For example using the annotations:
```
nginx.ingress.kubernetes.io/permanent-redirect: https://hub.example.com/$1$2$3
nginx.ingress.kubernetes.io/permanent-redirect-code: "308"
nginx.ingress.kubernetes.io/use-regex: "true"
```

And the path:
```
      - path: /(example)(/|$)(.*)
        pathType: Prefix
```

If we send a HTTP request like https://spoke.example.com/example/foo/bar in Ingress-Nginx we get redirected to https://hub.example.com/example/foo/bar, and we expected the same behaviour in Traefik.

### What did you see instead?

In Traefik instead we are redirected to https://spoke.example.com/. 

I believe this problem comes from the `buildRedirect` function in: https://github.com/traefik/traefik/blob/db21cbae1ad4d2b0e8a8d3648adecdeadcb57ffb/pkg/provider/kubernetes/ingress-nginx/middleware.go#L149-L153
It uses a regex of `.*` which means that we can't use regex capture groups from the paths.

v3.7.1

```yaml
providers:
  kubernetesIngressNGINX:
    enabled: true
    publishService:
      enabled: false
    ingressClass: "nginx"
    controllerClass: "k8s.io/ingress-nginx"
    ingressClassByName: true
    watchIngressWithoutClass: true
    strictValidatePathType: false
    allowCrossNamespaceResources: true
    ipAllowListStrategy:
      depth: 1
```

_No response_

AmariahAK:
### What does this PR do?
Fixes the `buildRedirect` function in the IngressNGINX provider so that `$1`, `$2`, `$3` (regex capture group references) in `permanent-redirect` / `temporal-redirect` annotation values resolve correctly against the path regex when `use-regex: "true"` is set, matching ingress-nginx behavior.
### Motivation
Fixes #13474
When `use-regex: "true"` and the path has regex capture groups (e.g. `/(example)(/|$)(.*)`), the redirect annotation `permanent-redirect: https://hub.example.com/$1$2$3` should capture the path groups and substitute them into the redirect target. Instead, the `RedirectRegex` middleware used a hardcoded `".*"` regex (zero capture groups), so `$n` references resolved to empty strings.
For example, `https://spoke.example.com/example/foo/bar` should redirect to `https://hub.example.com/example/foo/bar`, but was redirecting to `https://hub.example.com/`.
**Where:** `pkg/provider/kubernetes/ingress-nginx/middleware.go` — `buildRedirect` (line 149)
**Fix:** When `loc.UseRegex` is true, the regex is built as `^https?://[^/]+<loc.Path>` instead of `".*"`. The prefix matches the scheme+host portion non-capturing, so the path's own capture groups (`(example)`, `(/|$)`, `(.*)`) are at positions `$1`, `$2`, `$3` as expected. When `use-regex` is false, the regex stays `".*"` — fully backward compatible.
**Why this approach:** `loc.UseRegex` was already computed by the builder from the `use-regex` annotation — no new parsing, no new fields. The fix is a 5-line change that mirrors ingress-nginx's `location ~` semantics exactly.
### More
- [x] Added/updated tests
Assisted-by: DeepSeek V4 Pro <agent@atlarix.dev>

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingress-with-redirect-use-regex.yml
+++ b/pkg/provider/kubernetes/ingress-nginx/fixtures/ingresses/ingress-with-redirect-use-regex.yml
@@ -1,0 +1,23 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: ingress-with-redirect-use-regex
+  namespace: default
+  annotations:
+    nginx.ingress.kubernetes.io/use-regex: "true"
+    nginx.ingress.kubernetes.io/permanent-redirect: "https://redirect.example.com/$1$2"
+
+spec:
+  ingressClassName: nginx
+  rules:
+    - host: redirect-regex.localhost
+      http:
+        paths:
+          - path: /(example)(/.*)
+            pathType: ImplementationSpecific
+            backend:
+              service:
+                name: whoami
+                port:
+                  number: 80
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
@@ -8054,26 +8054,26 @@ func TestLoadIngresses(t *testing.T) {
 							RedirectRegex: &dynamic.RedirectRegex{
 								Regex:       `^https?://[^/]+/(example)(/.*)`,
 								Replacement: "https://redirect.example.com/$1$2",
-								StatusCode:  ptr.To(http.StatusMovedPermanently),
+								StatusCode:  new(http.StatusMovedPermanently),
 							},
 						},
 						"default-ingress-with-redirect-use-regex-rule-0-path-0-tls-redirect": {
 							RedirectRegex: &dynamic.RedirectRegex{
 								Regex:       `^https?://[^/]+/(example)(/.*)`,
 								Replacement: "https://redirect.example.com/$1$2",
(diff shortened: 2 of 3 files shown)


kevinpollet:
Hey @AmariahAK,

Thanks for this contribution!

As this is a fix, please could you rebase your pull request on v3.7?

AmariahAK:
@kevinpollet  thanks for the feedback, branch should now be on latest code

justdan96:
It looks like the tests for this do not pass, `ptr.To(http.StatusMovedPermanently)` is not valid but `new(http.StatusMovedPermanently)` does work.

AmariahAK:
@justdan96  should be fixed now

AmariahAK:
@kevinpollet thanks a lot for the assist, i forgot to do this so thanks again

kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

traefiker:

traefiker:

traefiker:

traefiker:

PR #13481, opened 2026-07-07
Fix panic when endpointslice port value or name is nil

kevinpollet:
### What does this PR do?

This pull request fixes a panic when the endpointslice port name or port is nil.

### More

- [X] Added/updated tests


rtribotte (approved):
LGTM

PR #13482, opened 2026-07-07
Report the invalid value when trusted IP CIDR parsing fails

archy-rock3t-cloud:
### What does this PR do?

When a trusted IP entry fails to parse as a CIDR, the error interpolated the parsed `*net.IPNet`, which is `nil` on failure, so the prefix always read `parsing CIDR trusted IPs <nil>`. It now shows the invalid entry.

### Motivation

Names the invalid entry in the error prefix, including empty entries where the wrapped error shows nothing.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/ip/checker.go
+++ b/pkg/ip/checker.go
@@ -30,7 +30,7 @@ func NewChecker(trustedIPs []string) (*Checker, error) {
 
 		_, ipAddr, err := net.ParseCIDR(ipMask)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


rtribotte:
Hey @archy-rock3t-cloud,

Thanks for this contribution!

We'll update you once the review is done, in the mean time, since it is a fix, could you please rebase the changes on branch v3.6?

archy-rock3t-cloud:
sure, done

archy-rock3t-cloud:
@rtribotte please take a look

PR #13492, opened 2026-07-08
Disable persisted Git credentials in CI workflow checkouts

darkweaver87:
# Motivation

The `actions/checkout` steps persist the GitHub token in `.git/config` for the whole job, which is unnecessary here and needlessly widens the credential's exposure window (CWE-522/200). This sets `persist-credentials: false` on every workflow checkout — none rely on the persisted credential: `documentation.yaml` publishes via `mixtus` using its own token, and `release.yaml` runs goreleaser, which authenticates through `GITHUB_TOKEN`.


rtribotte (approved):
Thanks!

PR #13496, opened 2026-07-08
Fix sticky cookie expiration per request

### LINKED ISSUE #13485: kubernetesIngressNGINX: session-cookie-expires sets a fixed Expires, so the affinity cookie ends up in the past
### Welcome!

### What did you do?

Enabled the kubernetesIngressNGINX provider and used cookie affinity with session-cookie-expires on an Ingress whose backend has 2 replicas:

annotations:
  nginx.ingress.kubernetes.io/session-cookie-name: "sticky"
  nginx.ingress.kubernetes.io/session-cookie-expires: "30"

Then requested it, waited 45s, and requested again:

curl -sD- -H 'Host: app.test' http://<traefik>/ | grep -iE '^(date|set-cookie):'

### What did you see instead?

Expires is computed once and frozen. It does not track the response Date, so after 30s it is already in the past and the browser discards the cookie:

t=0     date: ...11:37:27    set-cookie: sticky=...; Expires=...11:37:45
t=45s   date: ...11:38:12    set-cookie: sticky=...; Expires=...11:37:45   (unchanged, now in the past)

The cookie value changes on each request, only Expires is stuck. It refreshes only when the Ingress endpoints change (for example a pod restart), which then works for expires seconds before going stale again.

Effect: session affinity silently breaks expires seconds after the last reconcile. Anything session-bound behind it then breaks across replicas.

reproduced on v3.7.0 through v3.7.6.

RKE2: [v1.34.9+rke2r1](https://github.com/rancher/rke2/releases/tag/v1.34.9%2Brke2r1)

_No response_

makaiver:
### What does this PR do?

A suggestion for how to fix the issue where the kubernetesIngressNGINX provider _session-cookie-expires_ sets a fixed Expires, so the affinity cookie ends up in the past as per issue #13485.

### Motivation

Cookie-based session affinity via the kubernetesIngressNGINX provider breaks ~`N` seconds after each config reload, only recovering on the next reconcile/restart. Ingress NGINX calculates this value per request, so Traefik Kubernetes Ingress NGINX
should ideally behave the same way.

### More

- [x] Added/updated tests

### Additional Notes

The added regression test fails on the current code and will start to pass with this PR.

### DIFF
--- a/pkg/server/service/loadbalancer/sticky.go
+++ b/pkg/server/service/loadbalancer/sticky.go
@@ -29,7 +29,7 @@ type stickyCookie struct {
 	httpOnly bool
 	sameSite http.SameSite
 	maxAge   int
-	expires  time.Time
+	expires  time.Duration
 	path     string
 	domain   string
 }
@@ -63,7 +63,7 @@ func NewSticky(cookieConfig dynamic.Cookie) *Sticky {
 		cookie.path = *cookieConfig.Path
 	}
 	if cookieConfig.Expires > 0 {
-		cookie.expires = time.Now().Add(time.Duration(cookieConfig.Expires) * time.Second)
+		cookie.expires = time.Duration(cookieConfig.Expires) * time.Second
 	}
 
 	return &Sticky{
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hey @makaiver,

Thanks for this contribution!

amazon7737 (approved):
LGTM, Thanks !

mmatur (approved):
LGTM

PR #13504, opened 2026-07-10
Set http.route on root server span and update span name to include route

### LINKED ISSUE #11230: Spans should have more senseful names
### Welcome!

Hi,

I activated OpenTelemetry tracing in the current version of Traefik (installed via Helm chart. When I take a look in Tempo I see the following:

Image

I would prefer that the Span names would be something like

```
GET https://<hostname>/path
```

Ideally the pattern is configurable.

Same for the outgoing Span "ReverseProxy". It should be named like

```
GET http://<servicename>/path
```

Additionally I think all Spans between EntryPoint and ReverseProxy should be optional, they don't contain valueful information and just bloat the trace.

newdogwang-netizen:
## What does this PR do?

Fixes #11230

The root server span created by the entrypoint middleware only used the HTTP method as its name (e.g. `"GET"`), making it impossible to distinguish requests by path in tracing backends like Tempo, Jaeger, or Datadog. This is a known limitation discussed in #11230 and acknowledged by the Traefik team.

This PR enriches the root server span when a router matches the request:

- Sets the `http.route` attribute on the root server span with the extracted path from the router rule
- Updates the root server span name to `"{method} {route}"` (e.g. `"GET /api/v1/ml-scribe"`) per the [OTel HTTP semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name)

### Route extraction

The route value is extracted from the router rule:

| Rule | Extracted `http.route` | Span name |
|------|----------------------|-----------|
| `Path(\`/healthz\`)` | `/healthz` | `GET /healthz` |
| `PathPrefix(\`/api/v1/ml-scribe\`)` | `/api/v1/ml-scribe` | `GET /api/v1/ml-scribe` |
| `Host(\`example.com\`) && PathPrefix(\`/grafana\`)` | `/grafana` | `GET /grafana` |
| `Host(\`example.com\`)` | `Host(\`example.com\`)` (fallback to raw rule) | `GET Host(\`example.com\`)` |

For `Path()` and `PathPrefix()` matchers, the path argument is used directly. For complex rules that cannot be reduced to a simple path (`PathRegexp`, `Host`-only, etc.), the raw rule string is used as a fallback so the `http.route` attribute is always populated when a router matches.

If no router matches (404), the span name remains `"{method}"` and no `http.route` is set — preserving the current behavior for unmatched requests.

### Implementation

The router tracing middleware (`router.go`) captures the root server span from the request context **before** creating its own internal span (which would replace the context's active span). It then sets `http.route` and calls `SetName()` on the root server span.

The router's own internal span keeps the raw rule string as its `http.route` attribute (unchanged behavior).

### Testing

- Updated `router_test.go` with 4 test cases covering Path, PathPrefix, complex (Host + PathPrefix), and Host-only rules
- Fixed `mock_tracing_test.go`: `IsRecording()` now returns `true` so span modifications are testable
- Fixed the test to use `tracing.NewTracer()` wrapper and `WithObservability()` context, which are required for `TracerFromContext` and `DetailedTracingEnabled` to work
- All existing tests in the `observability` package continue to pass

### Notes

- The `http.route` attribute was already being set on the router's internal span (line 56 of `router.go`), but not on the root server span. This PR moves it to the server span where tracing backends expect it for endpoint grouping.
- `Span.SetName()` is used to update the name after span creation. While the OTel Go SDK recommends setting the name at `Start()` time, the route is not known at the entrypoint level (where the server span is created) — it's only known when a router matches. `SetName()` is the correct approach for this use case.

### DIFF
--- a/pkg/middlewares/observability/entrypoint.go
+++ b/pkg/middlewares/observability/entrypoint.go
@@ -57,7 +57,12 @@ func (e *entryPointTracing) ServeHTTP(rw http.ResponseWriter, req *http.Request)
 	start := time.Now()
 
 	// Follow semantic conventions defined by OTEL: https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name
-	// At the moment this implementation only gets the {method} as there is no guarantee {target} is low cardinality.
+	// The span name initially uses only {method} (e.g. "GET") because the route
+	// is not yet known at the entrypoint level. When a router matches the
+	// request, the router tracing middleware enriches this server span with
+	// the http.route attribute and updates the span name to "{method} {route}"
+	// (e.g. "GET /api/v1/ml-scribe"). If no router matches (404), the span
+	// name remains "{method}".
 	tracingCtx, span := e.tracer.Start(tracingCtx, req.Method, trace.WithSpanKind(trace.SpanKindServer), trace.WithTimestamp(start))
 
 	// Associate the request context with the logger.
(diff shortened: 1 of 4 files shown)


kevinpollet:
Hey @newdogwang-netizen,

PR #13506, opened 2026-07-10
Sanitize rewritten target on ingress-nginx provider

gndz07:
### What does this PR do?

Sanitize the rewritten path applied for ingress-nginx provider (either via rewrite-target annotation or configuration snippets) to avoid split view.

### Motivation

To fix the behavior.

### More

- [x] Added/updated tests


LBF38 (approved):
LGTM

kevinpollet (approved):

PR #13509, opened 2026-07-11
feat(middleware): add IPAllowGroup middleware for OR-chaining

### LINKED ISSUE #12249: Optimize ipAllowList middleware in the config (OR-chaining)
### Welcome!

eg. I have 20 routers, with different IPs whitelisted, but all of them have the same management IPs whitelisted.

The way traefik currently works, I have to create 20 whitelists, with all management IPs + special IPs for that router:
```
  routers:
    first-router:
      rule: Host(`domain.tld`)
      entryPoints:
        - web-secure
      middlewares:
        - security-headers
        - first-whitelist # <-----------
      service: first

  middlewares:
    first-whitelist:
      ipAllowList:
        sourceRange:
          - "1.1.1.1" # mgmt IP 1
          - "2.2.2.2" # mgmt IP 2
          - "3.3.3.3" # special request IP 3 for "first" service
          ...
```
And imagine having this for 20 routers, if I want to add a device to the management IPs whitelist, I have to edit all 20 whitelists.

---

Instead of having a single management whitelist and special ones on a per case basis. Like this:
```
  routers:
    first-router:
      rule: Host(`domain.tld`)
      entryPoints:
        - web-secure
      middlewares:
        - security-headers
        - management-whitelist # <--------
        - first-whitelist # <-----------
      service: first

  middlewares:
    management-whitelist:
      ipAllowList:
        sourceRange:
          - "1.1.1.1" # mgmt IP 1
          - "2.2.2.2" # mgmt IP 2
          ...

    first-whitelist:
      ipAllowList:
        sourceRange:
          - "3.3.3.3" # special use case IP 3 for "first" service
        ...
```

So now if I need to update management IPs, I do that in one place, and it is applied to all the routers.

Related: https://github.com/traefik/traefik/issues/6007

Aisu2635:
## What does this PR do?

This PR adds a new `IPAllowGroup` middleware which allows multiple `IPAllowList` rules to be combined using OR logic.

Currently, when multiple IP allowlist conditions are required, there isn't a simple way to allow a request if it matches **any one** of the configured allowlists.

With `IPAllowGroup`, the request is allowed as soon as the client IP matches one of the configured rules. If none of the rules match, the request is rejected.

This is mainly based on the use case discussed in #12249.

Fixes #12249

## Why?

In our organization, we have multiple partners where each partner has their own set of IPs that need to be allowlisted.

As the number of partners and IPs increased, managing all of them within a single `IPAllowList` started becoming difficult and unmanagable. Updating or removing IPs also becomes harder since all partner IPs are maintained together.

The idea behind `IPAllowGroup` is to keep these allowlists separate and partner-specific, while still allowing a request if it matches any one of them.

For example, separate middlewares can be maintained for different partner rules and combined with OR behaviour:

`Partner A matches OR Partner B matches OR Partner C matches -> allow request`

Currently, chaining IP allowlist rules results in AND behaviour, while this use case requires OR behaviour.

This also keeps the flexibility of having different `ipStrategy` configurations for individual rules when required.

## Changes

- Added the `IPAllowGroup` middleware configuration.
- Added support for multiple `IPAllowList` rules inside a group.
- Each rule builds and uses its own IP checker and strategy.
- Rules are checked sequentially and the request is allowed on the first match.
- Returns `403` by default when no rule matches.
- Added support for a custom `rejectStatusCode`.
- Added the middleware to the middleware builder.
- Regenerated the required deepcopy code.
- Added unit tests for validation and OR matching behaviour.

## Testing

I added tests covering:

- Empty rule configuration.
- Invalid CIDR configuration.
- Invalid reject status codes.
- IP matching the first rule.
- IP matching only the second rule.
- IP matching the third rule in a group.
- IP not matching any rule.
- Single rule behaviour.
- Custom reject status code.

Tested using:

```bash
go test -v ./pkg/middlewares/ipallowlist/...
```

## AI Assistance

Gemini 2.0 was used during the implementation to help with code exploration, implementation and test generation.

Assisted-by: Gemini 2.0


kevinpollet:
Hey @Aisu2635,

PR #13512, opened 2026-07-12
fix(ratelimiter): allow all requests when the Redis backend has rate limiting disabled (average 0)

TowyTowy:
### What does this PR do?

Fixes a Redis rate-limiter backend regression relative to the in-memory backend: with `average: 0` (documented as "no rate limiting", and the default) the Redis backend rejects **every** request with `429 "No bursty traffic allowed"`, while the in-memory backend with the same config lets all traffic through.

### Motivation

When `average` is `0`, `New` sets the rate to `rate.Inf`. The limiter's `Allow` contract uses a `nil` delay to mean "reject this request" (that's how the in-memory limiter signals an over-capacity bucket, and the caller answers `429`). The in-memory limiter returns a zero delay for `rate.Inf` (allow immediately), which matches the in-code note in `rate_limiter.go` that the reservation delay is `<= 0` in the `Inf` case. The Redis limiter's `evaluateScript`, however, returned `nil` in the `rate.Inf` branch — i.e. the reject signal — so all traffic got a 429.

The two backends are meant to be interchangeable, and the reference docs list `average: 0` → "0 means no rate limiting", so this is a divergence from both the sibling backend and the documented semantics.

### Reproduce

A `rateLimit` middleware with `average: 0` and a `redis` backend returns `429` for every request; swap it to the in-memory backend and the same requests pass.

### Fix

Return a zero delay (allow, no wait) from the Redis limiter in the `rate.Inf` case instead of `nil`. Adds `TestRedisRateLimitAverageZeroDisablesLimiting`, which fails before the change and passes after; it needs no live Redis because the `Inf` branch short-circuits before any Redis call.

I found this while auditing the two rate-limiter backends against each other; the change was made and tested with AI assistance and reviewed by me.

Assisted-by: Claude Fable 5

### DIFF
--- a/pkg/middlewares/ratelimiter/rate_limiter_test.go
+++ b/pkg/middlewares/ratelimiter/rate_limiter_test.go
@@ -551,6 +551,39 @@ func TestRedisRateLimit(t *testing.T) {
 	}
 }
 
+// An average of 0 means no rate limiting (see the ratelimit reference), so every request
+// must be allowed. The Redis backend must behave like the in-memory one in this case.
+func TestRedisRateLimitAverageZeroDisablesLimiting(t *testing.T) {
+	config := dynamic.RateLimit{
+		Average: 0,
+		Burst:   1,
+		Redis: &dynamic.Redis{
+			Endpoints: []string{"localhost:6379"},
+		},
+	}
+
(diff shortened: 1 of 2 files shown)


kevinpollet:
Hey @TowyTowy,

Thanks for this contribution!

Issue #13503, opened 2026-07-09
Suggestion: log a specific reason when respondingTimeouts closes a connection


gustavcke:
### Search check
I searched existing issues first and found #11788 ("File uploads... fail after a certain
threshold"), which covers a related symptom but a different ask (changing the default
`readTimeout` value) and was closed as declined/not-planned. This suggestion is
deliberately narrower and doesn't ask to change any default - only to improve what gets
logged when one of these timeouts fires.

### Context / grounding for this suggestion
We run Traefik in front of Immich (self-hosted photo/video server) on a home LAN. Large
video uploads from the Immich iOS app started failing intermittently with a bare `502
Bad Gateway`. Nothing in Immich's own logs indicated a problem - the request simply never
completed on the Immich side. Diagnosing this required cross-referencing Traefik's JSON
access logs across many requests before noticing that every failure shared an almost
suspiciously exact `Duration` of ~60000ms with `DownstreamStatus`/`OriginStatus` = 502,
which eventually led us to the entryPoints default `respondingTimeouts.readTimeout` (60s)
as the actual cause - a value we had never explicitly set.

Once we knew what to look for, confirming it took minutes. Before that, it took
considerably longer purely because nothing in Traefik's own logs said anything like
"closing connection: read timeout exceeded" - we only had a generic 502 to go on, and had
to correlate logs across two separate hosts to find the real cause.

### What we'd love to see
When Traefik closes a connection because one of the `respondingTimeouts` values
(`readTimeout`, `writeTimeout`, or `idleTimeout`) has been hit, log a specific message
identifying which timeout fired (and the entrypoint/router involved, if known at that
point in the request lifecycle).

Even a single log line like `entrypoint websecure: closing connection, readTimeout (60s)
exceeded` **would let anyone hitting this exact scenario - which seems fairly common for
self-hosted apps with large uploads**, since Immich's own docs already warn Traefik users
about this default - **go from "mysterious 502, no idea why" to "oh, it's the read
timeout" in seconds instead of extensive cross-host log correlation.** Since this kind of
scenario is exactly where Traefik often gets used, this could potentially save a lot of
people considerable headaches.

We're not asking to change any default (understanding why #11788 was declined) - purely
a diagnostics/observability suggestion. Thank you for considering it, and for
maintaining Traefik!

kevinpollet:
Hey @gustavcke,

[...]

PR #13516, opened 2026-07-14
Log which responding timeout closed the connection

k-adm:
### What does this PR do?

adds a debug log saying which respondingTimeout (read or write) closed a connection, so a timeout is diagnosable instead of only showing up as a bare gateway error. the read/write deadlines are already set on the wrapped conn in server_entrypoint_tcp.go - this just reports which one fired, tagged with the entrypoint. no defaults or behaviour change, only an extra log line.

### Motivation

follows up on #13503. right now when a readTimeout/writeTimeout fires nothing in the logs says so - you only get a generic 502/504 and have to correlate logs across hosts to find the cause. the issue describes a large-upload case where the default 60s readTimeout was the culprit and took a while to track down.

### More

- [x] Added/updated tests

### Additional Notes

kept it conservative (read/write only). a couple of things i'd like your call on before finalising, hence draft:

- log level: i put it at Debug to not be noisy on busy proxies (matches the already-swallowed http.Server ErrorLog level). downside is Debug is off by default, so someone hitting this still wouldn't see it without turning the level up. if it should be visible by default, Info/Warn might fit better.
- idleTimeout is left out on purpose. net/http enforces IdleTimeout with a read deadline on keep-alive conns, so at the conn wrapper an idle close looks identical to a readTimeout. writeTimeout is unambiguous. didn't want to mislabel idle closes as readTimeout - happy to add it if you have a preferred way to tell them apart.
- timeout detection uses errors.Is(err, os.ErrDeadlineExceeded) (the deadlines we set produce exactly that) instead of the net.Error.Timeout() idiom used elsewhere, to stay specific to the responding deadlines. can switch if you'd rather be consistent.

### DIFF
--- a/pkg/server/server_entrypoint_tcp.go
+++ b/pkg/server/server_entrypoint_tcp.go
@@ -284,7 +284,13 @@ func (e *TCPEntryPoint) Start(ctx context.Context) {
 				}
 			}
 
-			e.switcher.ServeTCP(newTrackedConnection(writeCloser, e.tracker))
+			e.switcher.ServeTCP(newTrackedConnection(
+				writeCloser,
+				e.tracker,
+				logger,
+				time.Duration(e.transportConfiguration.RespondingTimeouts.ReadTimeout),
+				time.Duration(e.transportConfiguration.RespondingTimeouts.WriteTimeout),
+			))
 		})
 	}
 }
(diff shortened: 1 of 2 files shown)


gustavcke:
Thank you for taking this on. A few thoughts on the three open points:

### Log level

I understand the concern about noise on busy public proxies, but I do not think a DEBUG-only message would have materially helped in the case that prompted me to open #13503.

The difficult part was discovering that Traefik's readTimeout was responsible for the unexplained 502. Nothing in Traefik's normal logs or in Immich's logs indicated a timeout, so there was no reason to enable debug logging or even to focus the investigation on Traefik's responding timeouts.

The cause only became apparent after correlating multiple access-log entries and noticing that each failure occurred at almost exactly 60 seconds. By then, the diagnosis was effectively complete. DEBUG logging would have confirmed an existing hypothesis, but it would not have made the cause discoverable in the first place.
(The signal that would tell an operator to enable debug logging is itself only available after debug logging has been enabled. It is circular.)

This also seems different from routine connection housekeeping: readTimeout or writeTimeout can terminate an active HTTP transaction and produce a client-visible failure. For those cases, WARN seems defensible.

If WARN would be too noisy, a structured termination reason in the access log might be the best long-term compromise. That would associate the reason with the failed request without adding every timeout event to the normal application log.

### idleTimeout

Leaving idleTimeout out seems reasonable if net/http applies it internally as a read deadline and the connection wrapper cannot reliably distinguish it from readTimeout.

It also appears operationally different from the reported problem. A normal idleTimeout expiry closes an unused keep-alive connection while waiting for another request; it does not normally abort an active transaction or produce the unexplained 502 seen in #13503.

I would prefer accurate readTimeout/writeTimeout reporting over guessing and occasionally labelling an idleTimeout as readTimeout. It may be useful to document explicitly that idleTimeout is not currently classified.

### Timeout detection

`errors.Is(err, os.ErrDeadlineExceeded)` seems preferable here because the intended signal is specifically expiration of an I/O deadline, rather than any network error that broadly reports `Timeout() == true`.

The narrower check should reduce the risk of classifying another timeout-like network error as one of the responding timeouts. As I understand it, the remaining ambiguity is not whether a deadline expired, but which read-side setting installed it, particularly readTimeout versus idleTimeout.

k-adm:
the circularity point is the one i hadn't weighed properly. Debug is the wrong default here, agreed.

before picking a level though, a correction on my idleTimeout bullet, and its worse than what i wrote there. with readTimeout set, the guard in trackedConnection.Read fires on any os.ErrDeadlineExceeded coming back from that Read, and net/http sets read deadlines on the same conn for things that aren't the responding read timeout. two of them land here. one is the idle deadline: idleTimeout() returns IdleTimeout when it's set, so on an http/1.1 keep-alive waiting for the next request that's the deadline in force, and when it expires the line comes out tagged readTimeout with the wrong duration. the other one is worse - abortPendingRead cancels the speculative background read with SetReadDeadline(aLongTimeAgo) every time a handler returns with that read in flight, and that goes down the same branch.

i ran it against the branch to be sure. three http/1.1 keep-alive requests on an entrypoint with stock defaults produced three "read timeout exceeded" lines before the connection had gone idle at all, then a fourth once idleTimeout fired. so most of what it logs today is noise on a connection that isn't closing and where nothing timed out, and the read side needs rework before this is worth merging.

the level question is still open either way. at Warn, as the code stands, that's about a warning per request instead of one per idle close.

@kevinpollet a couple of these i'd rather have your call on than guess at.

level: whether the reason belongs at Warn in the app log, or as a termination reason in the access log the way gustavcke suggests.

attribution: how you want a real responding-timeout expiry told apart from net/http's own use of the deadline, since the sentinel doesn't say who set it. skipping read attribution when idleTimeout is set is the cheap option, but idleTimeout is 180s by default on every entrypoint, so that switches the read log off out of the box and wouldn't have helped the case in #13503. tracking whether bytes arrived is closer, but it has to be "no bytes since the last write on this conn" rather than "no bytes at all", or it swallows the genuine no-first-byte timeout that TestReadTimeoutWithoutFirstByte covers. happy to implement whichever shape you want.

detection: no argument from me, staying on errors.Is(err, os.ErrDeadlineExceeded). i want it to catch a deadline we set expiring, and Timeout() would widen that to anything timeout-ish.

PR #13304, opened 2026-06-05
Fix inaccuracies in reference documentation

sheddy-traefik:
### What does this PR do?

Corrects inaccuracies across 83 reference documentation pages under docs/content/reference/, identified by cross-referencing every documented option against the Go source structs, JSON/YAML tags, SetDefaults() calls, and validation logic.

  Fix categories:

  - Wrong option names
  - Wrong defaults 
  - Phantom options — fields that don't exist in the struct and would be silently ignored: 
  - Wrong examples — broken JSON (missing colons/commas), TOML bare-word strings
  - Misleading descriptions
  - Missing options

Hub middleware pages (distributed-ratelimit, jwt, oidc, oauth2-client-credentials, oauth2-token-introspection, opa) were individually verified against the traefik-hub codebase before committing.

### Motivation

Community members reported a growing number of configuration errors tracing back to the docs revamp, where content volume outpaced review. This PR addresses the backlog in bulk.

### More

- [x] Added/updated documentation

### Additional Notes

  Each fix was verified against the current Go source — not just the audit output. A second verification pass was done on all hub middleware pages and a subset of the highest-severity non-hub changes (field name renames, phantom option removals, default value corrections) before committing.

Assisted-by: Claude Sonnet 4.6

PR #13518, opened 2026-07-15
Fix duplicated options table in ServersTransport CRD reference

rachana5:
### What does this PR do?

Fixes the Kubernetes CRD `ServersTransport` reference page (`docs/content/reference/routing-configuration/kubernetes/crd/http/serverstransport.md`), which currently renders its entire "Configuration Options" table twice, back-to-back. Because the two copies disagree, `maxIdleConnsPerHost` shows up with two different defaults on the same page — `0` in the first copy, `200` in the second.

This PR removes the duplicated rows and keeps a single table, using `0` as the default for `maxIdleConnsPerHost` (the accurate value — see Motivation) and also fixes the stale `rootcas` anchor/casing left over in the duplicate copy so it matches the `rootCAs` JSON tag used everywhere else on the page.

### Motivation

Discrepancy in docs reported by a customer. Traced it back to [#11965](https://github.com/traefik/traefik/pull/11965), which added `cipherSuites`/`minVersion`/`maxVersion` but duplicated the whole table instead of adding just those rows. [#13304](https://github.com/traefik/traefik/pull/13304) later corrected the *first* copy's `maxIdleConnsPerHost` default from `200` to `0` (to match `pkg/config/dynamic/http_config.go` and the explicit non-default behavior documented in `pkg/server/service/transport.go`), but missed the second, duplicate copy — leaving the stale `200` value in place.

### More

- [ ] Added/updated tests

### Additional Notes

`master` has since been restructured (#13416) and no longer has the duplicate table, but it still carries the incorrect `200` default for `maxIdleConnsPerHost` — opening a separate PR against `master` for that.


kevinpollet (approved):
Thanks 👍

PR #13519, opened 2026-07-15
Fix incorrect maxIdleConnsPerHost default in ServersTransport reference docs

rachana5:
### What does this PR do?

Corrects the documented default for `maxIdleConnsPerHost` from `200` to `0` on the two `master` reference pages that describe the `ServersTransport` options:

- `docs/content/reference/routing-configuration/kubernetes/crd/http/serverstransport.md` (Kubernetes CRD)
- `docs/content/reference/routing-configuration/http/load-balancing/serverstransport.md` (file/label providers)

### Motivation

`MaxIdleConnsPerHost` (`pkg/config/dynamic/http_config.go`) has no `SetDefaults()` call and no default struct tag, so it keeps Go's zero value (`0`) unless a user sets it explicitly. `pkg/server/service/transport.go` even has a comment explaining this is deliberate: *"Setting this value to the default of 100 could lead to confusing behavior and backwards compatibility issues."*

Both pages used to be correct. The Kubernetes CRD page briefly had its options table duplicated (see [#11965](https://github.com/traefik/traefik/pull/11965)), with `maxIdleConnsPerHost` showing `0` in one copy and `200` in the other; when that duplicate table was collapsed back to one on `master` (#13416), it kept the `200` copy instead of the `0` copy. The file/label page was correctly fixed to `0` on `v3.6`/`v3.7` in [#13304](https://github.com/traefik/traefik/pull/13304), but that fix was never forward-ported to `master`, which still had the old `200`.

A companion fix for the still-duplicated Kubernetes CRD table on `v3.7` is at [#13518](https://github.com/traefik/traefik/pull/13518) (that branch's file/label page was already correct, no change needed there).

### More

- [ ] Added/updated tests

### Additional Notes

Both changes are default-value/description corrections only, no behavior change.


kevinpollet (approved):

traefiker:
:no_entry_sign: error related to review: need more review [1/2]

PR #13031, opened 2026-04-23
feat: third-party attribution

bpsoraggi:
### What does this PR do?

Workflow that adds third-party attribution files using [Assimilis](https://github.com/traefik/assimilis) 🐜

Edits release workflow to include third-party attributions

Refactor of #12582

[[Cf.](https://github.com/bpsoraggi/traefik/pull/5)] & [[Cf.](https://github.com/traefik/assimilis/pull/5)]

### Motivation

Comply with license obligations

### More

- [ ] Added/updated tests
- [ ] Added/updated documentation

### Additional Notes

Release tested [here](https://github.com/bpsoraggi/traefik/actions/runs/24836145603), release notes can be found [here](https://github.com/bpsoraggi/traefik/releases/tag/v0.0.1-test)

PR #13522, opened 2026-07-15
feat: third-party attribution

bpsoraggi:
### What does this PR do?

Adds manual command for third-party attribution files using [Assimilis](https://github.com/traefik/assimilis) 🐜

Edits release workflow to validate and include third-party attribution files

Refactor of #13031, upon discussion

### Motivation

Comply with license obligations

### More

- [ ] Added/updated tests

### Additional Notes

Release tested [here](https://github.com/bpsoraggi/traefik/actions/runs/29410832336/job/87337252991) (_failure_) and [here](https://github.com/bpsoraggi/traefik/actions/runs/29410291261) (_success_)

### DIFF
--- a/script/third-party.sh
+++ b/script/third-party.sh
@@ -1,154 +1,137 @@
 #!/usr/bin/env bash
-set -eo pipefail
+set -euo pipefail
 
-ENV_FILE="${ENV_FILE:-$(pwd)/script/.env}"
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+ROOT_DIR="$(cd -- "${SCRIPT_DIR}/.." && pwd)"
+
+ENV_FILE="${ENV_FILE:-${SCRIPT_DIR}/.env}"
 ASSIMILIS_VERSION="${ASSIMILIS_VERSION:-v1.0.2}"
 AIKIDO_BASE_URL="${AIKIDO_BASE_URL:-https://app.aikido.dev}"
-SBOM_DIR="${SBOM_DIR:-$(pwd)/third_party/sbom}"
+SBOM_DIR="${SBOM_DIR:-${ROOT_DIR}/third_party/sbom}"
+TIMESTAMP_FILE="${ROOT_DIR}/third_party/.last_generated_at"
 REPO_NAME="traefik"
-ASSIMILIS_ERROR_LOG="$(pwd)/third_party/assimilis.log"
-TIMESTAMP_FILE="$(pwd)/third_party/.last_generated_at"
 
-if [[ ! -f "$ENV_FILE" ]]; then
-    printf "Credentials file not found: %s
+TMP_DIR="$(mktemp -d)"
+trap 'rm -rf "$TMP_DIR"' EXIT
+
+log() { printf '==> %s\n' "$*"; }
+
+die() { printf 'error: %s\n' "$*" >&2; exit 1; }
+
+load_env() {
+    if [[ ! -f "$ENV_FILE" ]]; then
+        die "Credentials file not found: ${ENV_FILE}
+
 Create it with:
 
-  cp .env.example .env
+  cp script/.env.example script/.env
 
-Then add your Aikido credentials." "$ENV_FILE"
-    exit 1
-fi
-
-set -a
-# shellcheck source=/dev/null
-source "$ENV_FILE"
-set +a
-
-TMP_DIR=$(mktemp -d)
-
-log() {
-    printf '==> %s\n' "$*"
-}
-
-format_timestamp_utc() {
-    local timestamp="$1"
-
-    if date --version >/dev/null 2>&1; then
-        # GNU date: Linux
-        formatted_scan_time="$(
-            date -u -d "@${timestamp}" '+%Y-%m-%d %H:%M:%S UTC'
-        )"
-    else
-        # BSD date: macOS
-        formatted_scan_time="$(
-            date -u -r "${timestamp}" '+%Y-%m-%d %H:%M:%S UTC'
-        )"
+Then add your Aikido credentials."
     fi
+
+    set -a
+    # shellcheck source=/dev/null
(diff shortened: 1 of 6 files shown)


script/third-party.sh
```diff
@@ -0,0 +1,154 @@
...
+cleanup() {
+  rm -rf "$TMP_DIR"
+  rm -f "$ASSIMILIS_ERROR_LOG"
+}
+
+trap cleanup EXIT
+
+log "Requesting access token from Aikido"
+
+OAUTH_RESPONSE="${TMP_DIR}/oauth-response.json"
```
darkweaver87: This 12–15-line "curl → capture `%{http_code}` → check 2xx → print error → exit" block repeats three times (here, SBOM, repo metadata). Extracting one helper and orchestrating with a `main` would cut the script by roughly half and make it read top-down:

```bash
die() { printf 'error: %s\n' "$*" >&2; exit 1; }

aikido_api() {  # aikido_api METHOD PATH [curl-args...]
  local method="$1" path="$2"; shift 2
  local body code; body="$(mktemp)"
  code="$(curl -sS -X "$method" "$AIKIDO_BASE_URL$path" \
            -w '%{http_code}' -o "$body" "$@")" || die "cannot reach Aikido ($path)"
  [[ "$code" =~ ^2 ]] || die "Aikido $path → HTTP $code: $(cat "$body")"
  cat "$body"
}

get_token()     { aikido_api POST /api/oauth/token ... | jq -er .access_token; }
download_sbom() { aikido_api GET ".../export?format=sbom" ... > "$SBOM_DIR/$REPO_NAME.cdx.json"; }

main() {
  load_env
  local token; token="$(get_token)" || die 'no token'
  download_sbom "$token"
  assimilis --repo-name "$REPO_NAME" || die 'assimilis failed'
  date -u '+%Y-%m-%dT%H:%M:%SZ' > "$TIMESTAMP_FILE"
}
main "$@"
```

The single `aikido_api` also fixes the path-vs-body error below, and dropping the `last_scanned_at` block + log-file machinery (other comments) is what gets you to ~60 lines.

script/third-party.sh
```diff
@@ -0,0 +1,154 @@
...
+        # BSD date: macOS
+        formatted_scan_time="$(
+            date -u -r "${timestamp}" '+%Y-%m-%d %H:%M:%S UTC'
+        )"
+    fi
+}
+
+cleanup() {
+  rm -rf "$TMP_DIR"
+  rm -f "$ASSIMILIS_ERROR_LOG"
```
darkweaver87: `cleanup` also `rm -rf`s `$TMP_DIR`, so the log survives nowhere — yet the failure message points the user to it. Simplest fix: drop the log-file mechanism entirely (`ASSIMILIS_ERROR_LOG` + `tee` + this `rm`). assimilis already writes to stdout, which the terminal/CI capture.

script/third-party.sh
```diff
@@ -0,0 +1,154 @@
...
+        --header 'content-type: application/json' \
+        --data '{"grant_type": "client_credentials"}' \
+        -o "${OAUTH_RESPONSE}"
+)"; then
+    printf "Could not connect to the Aikido authentication endpoint"
+    exit 1
+fi
+
+if [[ ! "$oauth_http_code" =~ ^2[0-9][0-9]$ ]]; then
+    printf "Aikido API error: %s" "$OAUTH_RESPONSE"
```
darkweaver87: This prints the temp-file path, not the response body — use `cat "$OAUTH_RESPONSE"`. Same in the SBOM branch. Also no `\n`, so this line and the next run together. (Subsumed if you adopt the `aikido_api` helper above.)

script/third-party.sh
```diff
@@ -0,0 +1,154 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+ENV_FILE="${ENV_FILE:-$(pwd)/script/.env}"
+ASSIMILIS_VERSION="${ASSIMILIS_VERSION:-v1.0.2}"
```
darkweaver87, line 8: `ASSIMILIS_VERSION` is never used — the script just calls `assimilis` from `$PATH`, so the version isn't pinned and the install step is undocumented. Either `go install …@$ASSIMILIS_VERSION` here or drop the var and document the prerequisite.

script/third-party.sh
```diff
@@ -0,0 +1,154 @@
...
+
+if [[ ! "$sbom_http_code" =~ ^2[0-9][0-9]$ ]]; then
+    printf "Aikido API error: %s/%s" "$SBOM_DIR" "${REPO_NAME}.cdx.json"
+    printf "Aikido SBOM export failed with HTTP status %s" "$sbom_http_code"
+    exit 1
+fi
+
+log "SBOM saved to ${SBOM_DIR}/${REPO_NAME}.cdx.json"
+
+REPOSITORY_RESPONSE="${TMP_DIR}/repository-response.json"
```
darkweaver87: This repository-metadata block (plus the `format_timestamp_utc` GNU/BSD helper above) exists only to print one cosmetic `last scanned at` log line — no functional effect. Dropping it and the log-file machinery would roughly halve the script.

script/validate-third-party.sh
```diff
@@ -0,0 +1,31 @@
...
+    exit 1
+fi
+
+generated_date="${generated_at%%T*}"
+release_date="$(date -u '+%Y-%m-%d')"
+
+echo "Generated at: ${generated_date}"
+echo "Release date: ${release_date}"
+
+if [[ "${generated_date}" != "${release_date}" ]]; then
```
darkweaver87: Exact same-day (UTC) equality is brittle: a tag pushed just after midnight fails, and it forces regeneration on every release day. A freshness window (≤ N days) would be more robust.

script/third-party.sh
bpsoraggi: Yes, I forgot a `TODO` comment 😅 I needed it like this because the fixes on Assimilis haven't been released yet

Issue #13517, opened 2026-07-14
Add IP masking


mpldr:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you expect to see?

For GDPR compliance, it's helpful to be able to mask IPs at ingress. Currently this is only possible using janky hacks. It would be nice to be able to support IP masking straight inside traefik.

Having taken a look, I think I have an idea on how to implement it and would be interested in doing so (in the access log middleware).

PR #13526, opened 2026-07-15
Allow IP anonymization

mpldr:
### What does this PR do?

Under GDPR, an IP is PII, but a subnet is not. As such, operators may want to redact access logs in order to not be required to have short retention times. Add configuration and logic to drop a configurable number of bits from logged IPs. Document legally established /24 and /48 subnets.

### Motivation

I am one of those operators

### More

- [x] Added/updated tests

### Additional Notes

I did not follow semantic linebreaks in the documentation updates, because the rest of the file doesn't use them either. If desired, I'll happily update that.

Implements #13517


kevinpollet:
Hey @mpldr,

Thanks for this contribution!

mpldr:
size/L seems a bit overblown… its not that big a changeset

PR #13529, opened 2026-07-16
feat(rate_limiter): add denyOnError option to Redis rate limiter

ayoubbts:
### Motivation

When Redis is unavailable, the rate limiter currently rejects all
requests with a 500. Add a denyOnError option on the Redis configuration
that, when false, logs the error and forwards the request to the next
handler instead (fail-open behavior).

The option defaults to true to preserve the existing fail-closed behavior.

Fixes this issue: https://github.com/traefik/traefik/issues/12043

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
+++ b/docs/content/reference/dynamic-configuration/kubernetes-crd-definition-v1.yml
@@ -2010,6 +2010,12 @@ spec:
                         description: DB defines the Redis database that will be selected
                           after connecting to the server.
                         type: integer
+                      denyOnError:
+                        description: |-
+                          DenyOnError controls the middleware behavior when Redis is unavailable or returns an error.
+                          When true (the default), the request is rejected with a 500 status code.
+                          When false, the error is logged and the request is forwarded to the next handler (fail-open).
+                        type: boolean
                       dialTimeout:
                         anyOf:
                         - type: integer
--- a/docs/content/reference/dynamic-configuration/traefik.io_middlewares.yaml
+++ b/docs/content/reference/dynamic-configuration/traefik.io_middlewares.yaml
@@ -1143,6 +1143,12 @@ spec:
                         description: DB defines the Redis database that will be selected
                           after connecting to the server.
                         type: integer
+                      denyOnError:
+                        description: |-
+                          DenyOnError controls the middleware behavior when Redis is unavailable or returns an error.
+                          When true (the default), the request is rejected with a 500 status code.
+                          When false, the error is logged and the request is forwarded to the next handler (fail-open).
+                        type: boolean
                       dialTimeout:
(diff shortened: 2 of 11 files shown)


kevinpollet:
Hey @ayoubbts,

Thanks for this contribution!

deyanp:
Adding an operator data point, since this one is blocking adoption for us.

We run Traefik as a DaemonSet in front of a payments platform — several hundred routes, ten `rateLimit` middlewares, roughly 250 router-to-middleware references. We want the Redis backend, because the in-memory limiter is per-pod and our effective limits are therefore multiplied by node count. But we cannot adopt it as things stand: with fail-closed as the only behaviour, any blip in the store becomes a 500 on every rate-limited route, which for us is essentially all ingress. That trades a rate-limiting problem for an availability one, and we would rather be briefly unprotected than briefly down.

So this PR is precisely the thing that makes the Redis backend usable for us, and I suspect for anyone whose rate limits are abuse protection rather than metering. The default of `true` keeps existing deployments exactly as they are, which seems right.

One small thing we noticed while reading it:

```go
denyOnError := config.Redis != nil && config.Redis.DenyOnError
```

When `Redis` is nil this evaluates to `false`, so the in-memory limiter switches from fail-closed to fail-open as a side effect. In practice that path looks unreachable — `inMemoryRateLimiter.Allow` only returns an error from `buckets.Set`, which errors when `ttlSeconds <= 0`, and the computed `ttl` is always at least 1 — so nothing changes today. But it makes the in-memory default the opposite of the Redis one, which would be surprising if the TTL computation ever changed.

Inverting it would keep both backends fail-closed by default:

```go
denyOnError := config.Redis == nil || config.Redis.DenyOnError
```

Happy to test the branch against our setup if that is useful.

deyanp:
@kevinpollet , when can this be reviewed?

deyanp:
@ayoubbts are you still planning to take this forward? The branch has not moved since
21 July, and this option is what currently blocks Redis-backed rate limiting for us.

If you would rather hand it on, I am happy to pick it up — either by pushing to your branch
if you add me to your fork, or as a separate PR that credits this one. I would keep your
design as it is and add three things:

- A fix for `denyOnError := config.Redis != nil && config.Redis.DenyOnError`. With `Redis`
  nil this is `false`, which flips the in-memory limiter from fail-closed to fail-open. That
  path looks unreachable today, but it leaves the two backends with opposite defaults.
- Tests. #13707 was closed partly for not having any, so they seem to be expected here.
- The documentation entry for the new option.

I do not want to step on your work, so I will leave the branch alone unless you say
otherwise.

@kevinpollet would that be a useful route, or would you rather this branch were reviewed as
it stands? Either is fine by me — I am asking because it has been in `status/2-needs-review`
since 20 July and I would rather help than keep bumping the thread.

PR #13533, opened 2026-07-17
Disable Zstd support in the gzhttp wrapper

kevinpollet:
### What does this PR do?

This pull request disables the zstd support added in https://github.com/klauspost/compress/pull/1121 by default in the gzhttp wrapper because we want to support only Gzip compression in this handler. This is a regression introduced by https://github.com/traefik/traefik/pull/12937

### More

- [X] Added/updated tests


rtribotte (approved):
LGTM

PR #13534, opened 2026-07-17
Return parser errors for malformed ingress nginx snippets

amazon7737:
### What does this PR do?

Updates `github.com/tufanbarisyildirim/gonginx` to `v0.0.0-20260220081509-8e17ce617db3`, so malformed ingress nginx snippets now return parser errors instead of reaching Traefik's panic recovery fallback.

The regression test covers malformed quoted strings, Lua blocks, and invalid `include` directives.

### Motivation

Traefik already recovers from snippet parsing panics to avoid crashing, but that recovery is a fallback path. This update uses the parser safety hardening added upstream in tufanbarisyildirim/gonginx#74 and keeps malformed snippets on the normal error path.

### More

- [x] Added/updated tests

### Additional Notes

The existing Traefik [`recover()` guard](https://github.com/amazon7737/traefik/blob/0b33e4178d79297be97c197fc39cc785c3e51f3f/pkg/middlewares/ingressnginx/snippet/snippet.go#L57-L62) is kept in place to protect the broader snippet setup path, while the updated parser reduces the malformed parser inputs that reach it.


kevinpollet:
Hello @amazon7737 and thanks for the pull request,

As this is a fix, please could you rebase it on branch v3.7?

PR #13536, opened 2026-07-18
Remove go-retryablehttp dependency from plugin downloader

andrewbytecoder:
### What does this PR do?

Replace the `go-retryablehttp` HTTP client with a standard `http.Client` in the plugin downloader initialization, and remove the direct dependency on `github.com/hashicorp/go-retryablehttp`.

### Motivation

The `initPlugins` function previously used a `retryablehttp.Client` with custom logger and retry settings for downloading plugin archives. This added an unnecessary direct dependency for a straightforward HTTP GET operation. A standard `http.Client` with a reasonable timeout is sufficient and keeps the dependency graph leaner.

### More

- [ ] Added/updated tests

### Additional Notes

The `go-retryablehttp` dependency remains in `go.mod` as an indirect dependency (still required by other packages in the dependency tree), so no `go mod tidy` changes are needed.

### DIFF
--- a/cmd/traefik/plugins.go
+++ b/cmd/traefik/plugins.go
@@ -6,10 +6,7 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/hashicorp/go-retryablehttp"
-	"github.com/rs/zerolog/log"
 	"github.com/traefik/traefik/v3/pkg/config/static"
-	"github.com/traefik/traefik/v3/pkg/observability/logs"
 	"github.com/traefik/traefik/v3/pkg/plugins"
 )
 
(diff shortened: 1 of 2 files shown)


nmengin:
Hey @andrewbytecoder,

Thanks for the contribution!
Could you elaborate on your motivations to change the current behavior?

In the meantime, we keep the status "need-triage".

andrewbytecoder:
This part of the code doesn’t do anything useful. During further development or when contributing code, it can easily lead to misunderstandings.

nmengin:
Hey @andrewbytecoder,

Thank you for your answer.

PR #13539, opened 2026-07-20
Prefer Host(*) over HostRegexp-catch-all

jnoordsij:
### What does this PR do?

As of v3.7.7, the expression ``Host(`*`)`` can be used to match any host. This PR changes the internal (https) redirect rule to leverage this syntax.

### Motivation

Comply with suggestions in documentation; from https://doc.traefik.io/traefik/v3.7/reference/routing-configuration/http/routing/rules-and-priority/#host-and-hostregexp:
> The Host matcher supports a single-level wildcard prefix (*.example.com) to match any direct subdomain of example.com. It should be preferred over the HostRegexp matcher as it allows attaching a TLS option and is more efficient.

### More

- [ ] Added/updated tests

### Additional Notes

This may change behavior for requests without URL or Host header, as documentation says this expression will also match requests without hosts, where the previous regex seems to match only those where Host contains at least one character. I think this behavior is more consistent and could be considered a bug fix if anything, although I think it will be unlikely to have impact in any practical situation.

### DIFF
--- a/pkg/provider/traefik/fixtures/redirection.json
+++ b/pkg/provider/traefik/fixtures/redirection.json
@@ -9,7 +9,7 @@
           "redirect-web-to-websecure"
         ],
         "service": "noop@internal",
-        "rule": "HostRegexp(`^.+$`)",
+        "rule": "Host(`*`)",
         "ruleSyntax": "default"
       }
     },
--- a/pkg/provider/traefik/fixtures/redirection_port.json
+++ b/pkg/provider/traefik/fixtures/redirection_port.json
@@ -9,7 +9,7 @@
           "redirect-web-to-443"
         ],
         "service": "noop@internal",
-        "rule": "HostRegexp(`^.+$`)",
+        "rule": "Host(`*`)",
         "ruleSyntax": "default"
       }
     },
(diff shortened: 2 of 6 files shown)


nmengin:
Hey @jnoordsij,

Thank you for this contribution.

> This may change behavior

Do yo uhave any idea of the behavior modifications that could bring?

jnoordsij:
> Do yo uhave any idea of the behavior modifications that could bring?

I've just tested this locally, for requests without `Host` header this will result in a redirect rather than (likely) 404:

```
# old behavior
$ curl --http2-prior-knowledge -I -H "User-Agent:" -H "Host:" http://192.168.178.1/
HTTP/2 404 
content-type: text/plain; charset=utf-8
x-content-type-options: nosniff
content-length: 19
date: Thu, 23 Jul 2026 12:50:54 GMT

# new behavior
$ curl --http2-prior-knowledge -I -H "User-Agent:" -H "Host:" http://192.168.178.1/
HTTP/2 308 
location: https:///
content-length: 18
date: Thu, 23 Jul 2026 12:53:42 GMT
```

I'm not sure if there is any real world example where such requests make sense or may occur, as they are relatively hard to create. For 'regular' IP-based queries, it seems the IP is also used as host. Maybe there is someone out there adding their own logic for such requests on a redirected port; in that case this would catch it first where previously it wouldn't.

PR #13545, opened 2026-07-21
Fix typo in nginx annotation proxy-buffer-numbers

fischerman:
### What does this PR do?

This PR fixes a typo in the documentation. The annotation is currently named in the first column but incorrectly in the notes. 

### Motivation

One might think that the annotation key needs to be changed during the migration from nginx.

### More

- [ ] Added/updated tests


kevinpollet (approved):

kevinpollet:
Thanks 👍

PR #13548, opened 2026-07-21
Redact duplicate TLS certificates in provider merge logs

xsergos:
### What?

`mergeCertificates` (`pkg/provider/merge.go`) logged the whole embedded `Certificate` struct with `%v` when skipping a duplicate certificate under `ResourceStrategySkipDuplicates`:

```go
Msgf("TLS certificate %v already configured, skipping", certificate.Certificate)
```

Because the file provider's `flattenCertificates` (`pkg/provider/file/file.go`) reads `CertFile`/`KeyFile` from disk and replaces the field values with the file content before the configuration reaches `mergeCertificates`, this wrote the raw certificate **and private key** to the WARN log — even when the certificate is referenced by file path, not inline content.

`KeyFile` is tagged `loggable:"false"` in `pkg/tls/certificate.go`, but the raw `%v` formatting bypasses that redaction mechanism. The sibling helper `logSkippedDuplicate` in the same file already logs only a resource key via `Str(...)`, so the certificate branch was the odd one out.

Example leaked line (private key redacted here):

```json
{"level":"warn","origin":"b.yaml","message":"TLS certificate {-----BEGIN CERTIFICATE-----\n...\n-----BEGIN PRIVATE KEY-----\n...\n} already configured, skipping"}
```

### How?

Log the truncated certificate name instead of the raw struct, matching how `pkg/tls/tlsmanager.go` already logs certificates:

```go
Msgf("TLS certificate %s already configured, skipping", certificate.Certificate.GetTruncatedCertificateName())
```

`GetTruncatedCertificateName()` returns only a truncated certificate identifier and no key material.

### Reproduce

1. File provider, one certificate referenced by file path.
2. Two dynamic-config files defining the same certificate:
   ```yaml
   tls:
     certificates:
       - certFile: /ssl/example.com.crt
         keyFile: /ssl/example.com.key
   ```
3. Start Traefik → the WARN "already configured, skipping" line contains the full PEM including the private key.

### Checklist

- [x] `gofmt` clean
- [x] `go vet ./pkg/provider/` clean
- [x] `go build ./pkg/provider/` ok


nmengin:
Hey @xsergos,

As it's a bugfix, could you please rebase on the branch `v3.6`.

Thanks

xsergos:
Done, rebased onto `v3.6`.

Note: on `v3.6` the duplicate-skip logging lives in `pkg/provider/file/file.go` (the `mergeCertificates` refactor in `pkg/provider/merge.go` only exists on `v3.7`+), so the fix is applied there. Same issue in both: the `%v` formatting dumped the whole `CertAndStores`/`Certificate` struct — and since `flattenCertificates` inlines the on-disk PEM into `CertFile`/`KeyFile` before this point, the private key was written to the WARN log even when the cert is referenced by file path. Now logs the truncated certificate name via `GetTruncatedCertificateName()`, consistent with `logSkippedDuplicate` and `pkg/tls/tlsmanager.go`.

xsergos:
Hey @nmengin, hope you can check and merge the fix if it's not fixed in upstream yet.

sdelicata:
Hi @xsergos! 👋

Thanks for this fix, and sorry for the ping-pong on the target branch. v3.6 reached end of support on August 16, so this now targets `v3.7`. You were right in your earlier comment: on `v3.7` the duplicate-skip logging lives in `mergeCertificates` (`pkg/provider/merge.go`), not in `pkg/provider/file/file.go` anymore.

I pushed a few changes on top of your commit:

- rebased on `v3.7` and moved the fix to `mergeCertificates`
- routed it through the existing `logSkippedDuplicate` helper, so the certificate name lands in a structured field like every other resource
- added a regression test asserting the warning carries no key material
- hardened `GetTruncatedCertificateName()`, which now decodes the PEM block and redacts anything whose first block isn't a certificate, since a `certFile` can be a bundle that also holds the key

Your analysis was spot on. Thanks again for the contribution!

sdelicata (approved):
LGTM

rtribotte (approved):
Thanks!

traefiker:

PR #13561, opened 2026-07-23
fix(provider/docker): don't drop container config when an unreferenced auto-created service can't resolve a port

### LINKED ISSUE #13532: Traefik dashboard works, but only if loadbalancer.server.port is included
### Welcome!

### What did you do?

Traefik dashboard works, but only if label `loadbalancer.server.port` is included.

However, it is expected that including the label `traefik.http.routers.traefik-secure.service: "api@internal"` alone should be sufficient.

### What did you see instead?

Without the `traefik.http.services.dashboard.loadbalancer.server.port: 80` label the browser shows error `404 page not found`.

/ # traefik version
```text
Version:      3.7.7
Codename:     langres
Go version:   go1.26.5
Built:        2026-07-08T15:31:52Z
OS/Arch:      linux/arm6
```

## Information to investigate why api@internal is not working

Investigated intensively with [Traefik community](https://community.traefik.io/t/traefik-works-but-only-if-loadbalancer-server-port-is-included/29970); no reason found.

Traefik itself is working, the dashboard isn't (without port label).

Traefik dashboard from the same Traefik instance url: https://traefik01.web.example.org/dashboard/

Traefik version 3.7.7  (Docker)

This is already for quite some, not specific for a certain Treafik release.

Only one port exposed:

```text
docker inspect traefik01 --format '{{json .Config.ExposedPorts}}'

{"80/tcp":{}}
```

### Compose.yaml

```text
services:
  traefik:
    image: traefik:v3.7
    container_name: traefik01
    restart: unless-stopped
    security_opt:
      - no-new-privileges:true
    networks:
      macvlan01:
        ipv4_address: 192.168.1.231

    environment:
      CF_API_EMAIL: ${CF_API_EMAIL}
      CF_DNS_API_TOKEN: ${CF_DNS_API_TOKEN}
      MIJNHOST_API_KEY: ${MIJNHOST_API_KEY}
      TRAEFIK_DASHBOARD_CREDENTIALS: ${TRAEFIK_DASHBOARD_CREDENTIALS}

    volumes:
      - /etc/localtime:/etc/localtime:ro
      - ./config/traefik.yml:/traefik.yml:ro
      - ./config/config.yml:/config.yml:ro

      - /srv/docker/swarm/.volumes/router/traefik01/pki:/etc/pki
      - /srv/docker/swarm/.volumes/router/traefik01/logs:/var/logs

    labels:
      dockhand.url: "[traefik](https://traefik01.web.example.org)"

      traefik.enable: true

      traefik.http.routers.traefik01-web-example-org-secure.entrypoints: "https"
      traefik.http.routers.traefik01-web-example-org-secure.rule: "Host(`traefik01.web.example.org`)"   
      traefik.http.routers.traefik01-web-example-org-secure.tls: true
      traefik.http.routers.traefik01-web-example-org-secure.tls.certresolver: "mijnhost"
      traefik.http.routers.traefik01-web-example-org-secure.middlewares: "traefik-auth"
      traefik.http.routers.traefik01-web-example-org-secure.service: "api@internal"
      traefik.http.services.dashboard.loadbalancer.server.port: 80   ### without this the dashboard is accessible

networks:
  macvlan01:
    external: true
```

### config/Traefik.yml

```text
api:
  dashboard: true
  debug: true

log:
   level: DEBUG
   format: json
   filePath: /var/logs/traefik.log
 
accesslog:
  filePath: /var/logs/access.log
  fields:
    names:
      StartUTC: drop

entryPoints:

  ssh:
    address: ":22"

  ssh24:
    address: ":24"

  http:
    address: ":80"

  http-external:
    address: ":81"

  https:
    address: ":443"
  
  https-external:
    address: ":444"

serversTransport:
  insecureSkipVerify: true

providers:
  docker:
    exposedByDefault: false

  file:
    filename: /config.yml
    watch: true

  redis:
    endpoints:
      - "redis:6379"

certificatesResolvers:

...
```

### config/config.yml

```text
http:

  routers:

    grafana-cloud:
      entryPoints:
         - https
      rule: "Host(`db01.web.mydomain.org`)"
      tls: true
      service: grafana-cloud

  services:
      

    grafana-cloud:
      loadbalancer:
        servers:
          - url: "http://cloud.mydomain.org/:3000"

  middlewares:

    grafana-db01-example-to-mydomain-rewrite:
      redirectRegex:
        regex: "^https://db01.example.org:444/(.*)"
        replacement: "https://db01.mydomain.org/${1}"

    grafana-db01-example-headers-manipulaton:
      headers:
        customrequestheaders:
          X-Script-Name: "X-Forwarded-For"

    redirect-http-to-https-once:
      redirectScheme:
        scheme: https

    redirect-http-to-https-permanent:
      redirectScheme:
        scheme: https
        permanent: true

    default-headers:
      headers:
        frameDeny: true
        browserXssFilter: true
        contentTypeNosniff: true
        forceSTSHeader: true
        stsIncludeSubdomains: true
        stsPreload: true
        stsSeconds: 15552000
        customFrameOptionsValue: SAMEORIGIN
        customRequestHeaders:
          X-Forwarded-Proto: https

    default-allowlist:
      ipAllowList:
        sourceRange:
        - "192.168.1.0/16"
        - "172.16.0.0/12"
        - "10.0.0.0/8"

    secured:
      chain:
        middlewares:
        - default-allowlist
        - default-headers

tcp:
  routers:

    owncloud:
      entryPoints: "https-external"
      rule: "HostSNI(`owncloud.mydomain.org`)"
      service: owncloud
      tls:
        passthrough: true

  services:
    owncloud:
      loadbalancer:
        servers:
          - address: "cloud.mydomain.org:443"
```

I think that the dashboard is not reachable may be due to the fact that I've a macvlan configured.
Going to the dashboard (in the fault situation results in) `404 page not found`.

When going to the dashboard url, the logs provide the following information.

access.log:

```text
192.168.1.xxx - - \[16/Jul/2026:11:16:04 +0200\] "GET /dashboard/ HTTP/2.0" 404 19 "-" "-" 241 "-" "-" 0ms 
192.168.1.xxx - - \[16/Jul/2026:11:16:04 +0200\] "GET /favicon.ico HTTP/2.0" 404 19 "-" "-" 242 "-" "-" 0ms
```

treafik.log has nothing. loglevel is set to debug:

```text
{"level":"**debug**",....
```

Why is there nothing in traefik.log? What does that tell?
Is noting routed by Traefik?

I if add the `.port` configuration label, (only) the following is logged:

```text
{"level":"debug","middlewareName":"traefik-auth@docker","middlewareType":"BasicAuth","time":"2026-07-16T11:46:57+02:00","caller":"github.com/traefik/traefik/v3/pkg/middlewares/auth/basic_auth.go:102","message":"Authentication succeeded"}
```

Could it be that Traefik is listening to 0.0.0.0 but, because of macvlan it should be listening to an IP address (or the other way around)?

Is this expected behavior?

Any idea what's (not) happening?

Included in previous section.

RahulRingane:
### What does this PR do?

When a container declares no service via labels, Traefik auto-creates a default service named after the container and tries to attach a backend server to it. If port resolution fails for that auto-created service, the error is currently treated as fatal for the entire container's HTTP configuration, including routers that never referenced that service at all (for example a router pointing at api@internal).

This PR makes that specific failure non-fatal, but only when it's provably safe: the service must be one Traefik invented on its own (not declared via labels), and every router already declared on the container must already have an explicit Service set. That second check mirrors the same condition BuildRouterConfiguration itself uses to decide whether to auto-wire a router to the default service, so it shouldn't introduce new judgment calls, just reuse an existing one slightly earlier. If either condition doesn't hold, behavior is unchanged and the container still fails loudly as before.

Fixes #13532

### Motivation

On macvlan networks, Docker doesn't populate NetworkSettings.Ports even for ports declared via EXPOSE, so the auto-created service reliably fails to resolve a port, and any router on that container gets silently dropped along with it, even one pointing at an internal service like api@internal that never needed that service to begin with. Full investigation and live repro are in the linked issue.

I verified this end to end against a real macvlan container built from this branch: without the fix, GET /api/http/routers returns nothing for the router and the container logs "port is missing"; with the fix, the router shows up correctly with status enabled and the correct service reference, and the error no longer appears.

### On the approach

I went with fixing this at the point where the service is built (buildServiceConfiguration), since that's the earliest place it's possible to tell whether the failing service is actually needed by anything. Open to other directions if a different layer makes more sense, a couple I considered but didn't take:

- Handling this further up in build(), by inspecting the whole container's config after a failure instead of deciding per-service. Went with the current approach since it keeps the decision local to the function that owns the service map, but happy to move it if that's preferred.
- Being more permissive and just building services with no servers at all rather than dropping them, similar to how allowEmptyServices already works for non-running containers. Didn't go this route since it changes behavior for every unresolvable-port case rather than only the orphaned/unreferenced ones, but wanted to flag it as an alternative in case that's actually closer to the intended semantics here.

Happy to adjust the approach based on what fits the codebase's conventions best, this is more of a starting point for discussion than something I'm attached to.

### More
- [x] Added/updated tests

### Additional Notes

Added two tests in pkg/provider/docker/config_test.go:
- TestDynConfBuilder_buildServiceConfiguration_orphanedAutoServicePortFailure confirms the router survives and the dead service is dropped when nothing references it.
- TestDynConfBuilder_buildServiceConfiguration_requiredAutoServicePortFailure confirms the original fail-hard behavior is unchanged when the default service is genuinely needed.

Ran the full pkg/provider/docker test suite locally, all existing tests pass unchanged.

### DIFF
--- a/pkg/provider/docker/config.go
+++ b/pkg/provider/docker/config.go
@@ -166,7 +166,9 @@ func (p *DynConfBuilder) buildUDPServiceConfiguration(ctx context.Context, conta
 func (p *DynConfBuilder) buildServiceConfiguration(ctx context.Context, container dockerData, configuration *dynamic.HTTPConfiguration) error {
 	serviceName := getServiceName(container)
 
+	autoCreatedService := false
 	if len(configuration.Services) == 0 {
+		autoCreatedService = true
 		configuration.Services = make(map[string]*dynamic.Service)
 		lb := &dynamic.ServersLoadBalancer{}
 		lb.SetDefaults()
@@ -187,11 +189,45 @@ func (p *DynConfBuilder) buildServiceConfiguration(ctx context.Context, containe
 	for name, service := range configuration.Services {
 		ctx := log.Ctx(ctx).With().Str(logs.ServiceName, name).Logger().WithContext(ctx)
 		if err := p.addServer(ctx, container, service.LoadBalancer); err != nil {
+			// If this service was auto-created (the user declared no service
+			// via labels) and every declared router already has an explicit
+			// Service set (e.g. pointing at an internal service such as
+			// api@internal), nothing in this container's configuration
+			// actually needs this default service. In that case, don't take
+			// down the whole container's configuration (routers included)
+			// over a backend that was never going to be used; drop the
+			// unusable service and keep going.
+			if autoCreatedService && !p.defaultServiceRequired(configuration) {
+				log.Ctx(ctx).Warn().Err(err).
+					Msgf("Ignoring auto-created service %q: no router references it", name)
+				delete(configuration.Services, name)
+				continue
+			}
 			return fmt.Errorf("service %q error: %w", name, err)
 		}
 	}
 
 	return nil
+}
+
+// defaultServiceRequired reports whether the (as yet unbuilt) default router
+// wiring could still end up depending on the container's auto-created,
+// name-based default service. This is true whenever there are no routers
+// declared at all (provider.BuildRouterConfiguration will create one and
+// auto-assign it the container's sole service), or when at least one
+// declared router has no explicit Service set (same auto-assignment logic
+// applies to it). If every declared router already has an explicit Service,
+// the default service is unreachable dead weight and safe to drop on error.
+func (p *DynConfBuilder) defaultServiceRequired(configuration *dynamic.HTTPConfiguration) bool {
+	if len(configuration.Routers) == 0 {
+		return true
+	}
+	for _, router := range configuration.Routers {
+		if router.Service == "" {
+			return true
+		}
+	}
+	return false
 }
 
 func (p *DynConfBuilder) keepContainer(ctx context.Context, container dockerData) bool {
--- a/pkg/provider/docker/config_test.go
+++ b/pkg/provider/docker/config_test.go
@@ -4749,3 +4749,80 @@ func TestDynConfBuilder_getIPAddress_swarm(t *testing.T) {
 		})
 	}
 }
+
+// TestDynConfBuilder_buildServiceConfiguration_orphanedAutoServicePortFailure
+// covers the fix for the macvlan/no-port-label bug: when a container declares
+// no service label (so a default service gets auto-created from the
+// container name) but every declared router already has an explicit Service
+// set (e.g. a dashboard router pointing at api@internal), a port-resolution
+// failure on that unused auto-created service must not take the router down
+// with it.
+func TestDynConfBuilder_buildServiceConfiguration_orphanedAutoServicePortFailure(t *testing.T) {
+	container := dockerData{
+		ServiceName: "traefik13532",
+		Name:        "traefik13532",
+		NetworkSettings: networkSettings{
+			// Empty Ports mirrors what Docker reports for a macvlan-attached
+			// container even when a port is declared via EXPOSE.
+			Ports: networktypes.PortMap{},
+			Networks: map[string]*networkData{
+				"macvlan01": {
+					Name: "macvlan01",
+					Addr: "192.168.1.231",
+				},
+			},
+		},
+	}
+
+	configuration := &dynamic.HTTPConfiguration{
+		Routers: map[string]*dynamic.Router{
+			"dashboard": {
+				Rule:        "PathPrefix(`/dashboard`)",
+				EntryPoints: []string{"web"},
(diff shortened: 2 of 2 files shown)


nmengin:
Hey @RahulRingane,

We've set the status to "design-review" to allow us to check the PR and ensure there is no deep impact on Traefik before moving forward, and provide you with feedback.

PR #13569, opened 2026-07-26
Abort the response when the fast proxy fails after the response has started

### LINKED ISSUE #13568: FastProxy appends "Internal Server Error" to the response body when the backend fails mid-response
### Welcome!

### What did you do?

With `experimental.fastProxy` enabled, I proxied a backend that starts a chunked response and then goes away before finishing it, which is what a pod being rolled during a long-lived stream looks like.

Here is a self-contained reproduction. It runs against the package's own test helpers, so it needs only a checkout: drop it in `pkg/proxy/fast/` as `repro_test.go` and run `go test -run TestRepro -v ./pkg/proxy/fast/`.

```go
package fast

import (
	"bufio"
	"io"
	"net"
	"net/http/httptest"
	"testing"

	"github.com/stretchr/testify/require"
	"github.com/traefik/traefik/v3/pkg/config/static"
	"github.com/traefik/traefik/v3/pkg/testhelpers"
)

func TestRepro(t *testing.T) {
	// A backend that commits a chunked 200, writes one chunk,
	// then goes away without the terminating chunk.
	listener, err := net.Listen("tcp", "127.0.0.1:0")
	require.NoError(t, err)
	t.Cleanup(func() { _ = listener.Close() })

	go func() {
		for {
			conn, err := listener.Accept()
			if err != nil {
				return
			}

			go func() {
				defer conn.Close()

				reader := bufio.NewReader(conn)
				for {
					line, err := reader.ReadString('\n')
					if err != nil || line == "\r\n" {
						break
					}
				}

				_, _ = io.WriteString(conn, "HTTP/1.1 200 OK\r\nTransfer-Encoding: chunked\r\n\r\n")
				_, _ = io.WriteString(conn, "5\r\nchunk\r\n")
			}()
		}
	}()

	builder := NewProxyBuilder(&transportManagerMock{}, static.FastProxyConfig{})

	proxyHandler, err := builder.Build("", testhelpers.MustParseURL("http://"+listener.Addr().String()), true, true)
	require.NoError(t, err)

	proxyServer := httptest.NewServer(proxyHandler)
	t.Cleanup(proxyServer.Close)

	res, err := proxyServer.Client().Get(proxyServer.URL)
	require.NoError(t, err)

	t.Cleanup(func() { _ = res.Body.Close() })

	body, readErr := io.ReadAll(res.Body)
	t.Logf("body=%q readErr=%v", string(body), readErr)
}
```

I expected the client to receive the five bytes the backend managed to send, plus an error telling it the response is incomplete. That is what the default proxy does: the same backend behind `httputil.NewProxyBuilder(...).Build(..., 0)` gives:

```
body="chunk" readErr=unexpected EOF
```

### What did you see instead?

```
body="chunkInternal Server Error" readErr=<nil>
```

Two things go wrong:

1. The string `Internal Server Error` is appended to the response body, inside the payload the client is already reading.
2. `readErr` is `nil`. The response is terminated as if it were complete, so the client has no way to detect that it is truncated.

A JSON parser will choke on the trailing bytes. Anything more forgiving just accepts a payload that is missing data.

The cause is in `ReverseProxy.ServeHTTP` ([`pkg/proxy/fast/proxy.go#L237`](https://github.com/traefik/traefik/blob/v3.6/pkg/proxy/fast/proxy.go#L237-L239)):

```go
if err := p.roundTrip(rw, req, outReq, reqUpType); err != nil {
	proxyhttputil.ErrorHandler(rw, req, err)
}
```

`roundTrip` returns errors from both before and after the response has started, but `ErrorHandler` is only valid before. `handleResponse` calls `r.RW.WriteHeader(res.StatusCode())` ([`connpool.go#L212`](https://github.com/traefik/traefik/blob/v3.6/pkg/proxy/fast/connpool.go#L212)) and only then copies the body, so a copy failure arrives with the status line already sent. `ErrorHandler`'s `WriteHeader(500)` is ignored at that point, but its `Write([]byte("Internal Server Error"))` is not, so those bytes land in the body. `ServeHTTP` then returns normally, and the server finishes the response cleanly.

`net/http/httputil.ReverseProxy` handles the same case by aborting the request instead ([`reverseproxy.go#L613`](https://github.com/golang/go/blob/go1.26.5/src/net/http/httputil/reverseproxy.go#L613)), and Traefik's own recovery middleware documents the same rule ([`recovery.go#L123`](https://github.com/traefik/traefik/blob/v3.6/pkg/middlewares/recovery/recovery.go#L123-L124)):

> If headers have been sent this is not possible to respond with an HTTP error, and we let the server abort the response silently thanks to the http.ErrAbortHandler sentinel panic value.

The same path is hit when it is the client that goes away instead of the backend, which on an SSE route is just a browser closing a tab. The injected bytes go to a stream nobody is reading, so the client is unaffected, but the 500 still reaches `captureResponseWriter.WriteHeader`, which keeps the last status it is given ([`capture.go#L180`](https://github.com/traefik/traefik/blob/v3.6/pkg/middlewares/capture/capture.go#L180)). Traefik then logs and counts the request as a 500. On a route serving long-lived streams, that is every ordinary disconnect.

None of this needs HTTP/2 or SSE: the reproduction above is HTTP/1.1 end to end.

v3.6.24 and v3.7.9. The code path is unchanged on `v3.6`, `v3.7` and `master` (`fd5a576df`), so all three are affected.

```yaml
experimental:
  fastProxy: true
```

The reproduction needs nothing else: no TLS and no middleware. Only `experimental.fastProxy` matters, since `smart_builder.go` otherwise routes the request to the `httputil` proxy, which is unaffected.

At `--log.level=DEBUG` the exchange produces two lines and nothing at INFO or above. `ErrorHandler` logs the underlying error as `500 Internal Server Error`, and net/http reports the write that was ignored:

```
http: superfluous response.WriteHeader call from github.com/traefik/traefik/v3/pkg/proxy/httputil.ErrorHandlerWithContext (proxy.go:168)
```

With the access log enabled the request is recorded as a 500, which is the only signal at default log level, and it does not match the 200 the client received.

### Related

- #10807 and #10819 covered access log correctness for streaming responses aborted by the client, in the access log middleware. The wrong status here has a different origin: the fast proxy writes a 500 after the response has started.
- #11728 reports a fastProxy streaming failure that may share this root cause; I have not been able to confirm it.

FedeBev:
### What does this PR do?

`fast.ReverseProxy.ServeHTTP` passes every error from `roundTrip` to `proxyhttputil.ErrorHandler`, including errors that happen after the status line has been forwarded. `ErrorHandler` is only usable before that point: once the response has started, its `WriteHeader(500)` is ignored, but its `Write([]byte("Internal Server Error"))` is appended to the body the client is reading. `ServeHTTP` then returns normally, so the response is terminated as if it were complete.

`roundTrip` now reports whether the response had started, and `ServeHTTP` aborts with `http.ErrAbortHandler` in that case instead of calling `ErrorHandler`. `net/http/httputil.ReverseProxy` does the same on a copy error, and so does the recovery middleware when it recovers a panic after the headers have gone out.

Behavior on a backend that dies mid-response, HTTP/1.1 end to end:

|  | before | after |
| --- | --- | --- |
| body | `chunk` + `Internal Server Error` | `chunk` |
| client-side read error | none | `unexpected EOF` |

The `httputil` proxy already produces the second column, so this also removes a difference in behavior between the two proxies.

The same path is taken when it is the client that goes away, and there the 500 was landing in the access log and the request metrics for what is a normal disconnect. That now stays the 200 the client actually got.

Errors that happen before the response starts are unaffected and still go through `ErrorHandler`.

### Motivation

Fixes #13568.

The stray bytes in the body are the visible half. The half that made me dig in is that a lenient client cannot tell a stream that ended from a stream that broke.

### More

- [x] Added/updated tests

Three tests in `pkg/proxy/fast/proxy_test.go`:

- `TestUpstreamErrorAfterResponseStarted`: backend stops mid-chunked-response. The body must be left untouched and the client must see a read error. Without the fix it gets `"chunkInternal Server Error"` and no error.
- `TestDownstreamCancelAfterResponseStarted`: HTTP/2 client resets an in-flight stream. Nothing may be written downstream afterwards. Without the fix the recorded status is 500.
- `TestUpstreamErrorBeforeResponseStarted` is the guard on the other side: a dial failure must still surface as a 502.

No documentation change: this is not a behavior users configure.

### Additional Notes

`ServeHTTP`'s error handling and `connpool.go` are unchanged between `v3.6`, `v3.7` and `master` (v3.6 only differs earlier in `ServeHTTP`, in the X-Forwarded-For block), so the patch applies unchanged to all three.

The new marker is a per-attempt `atomic.Bool` on `conn`, set on the read loop goroutine and read on the request goroutine after `ErrCh`. It is reset before each attempt because connections are pooled and retried on. There is no added work on the success path beyond the two atomic stores.

`go test -race ./pkg/proxy/fast/` already fails on an unmodified `v3.6` tree, in the websocket and trailer tests and with a different set each run. Unrelated to this change, and CI does not run the suite with `-race`.

### DIFF
--- a/pkg/proxy/fast/connpool.go
+++ b/pkg/proxy/fast/connpool.go
@@ -43,6 +43,11 @@ type conn struct {
 	broken           atomic.Bool
 	upgraded         atomic.Bool
 
+	// responseStarted is set once the status line has been committed downstream,
+	// after which a forwarding error can no longer be turned into an error status
+	// (see ReverseProxy.ServeHTTP).
+	responseStarted atomic.Bool
+
 	closeMu  sync.Mutex
 	closed   bool
 	closeErr error
@@ -209,6 +214,7 @@ func (c *conn) handleResponse(r rwWithUpgrade) error {
 		r.RW.Header().Add(string(key), string(value))
 	})
 
+	c.responseStarted.Store(true)
 	r.RW.WriteHeader(res.StatusCode())
 
 	if noResponseBodyExpected(r.ReqMethod) {
--- a/pkg/proxy/fast/proxy.go
+++ b/pkg/proxy/fast/proxy.go
@@ -234,7 +234,20 @@ func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 		}
 	}
 
-	if err := p.roundTrip(rw, req, outReq, reqUpType); err != nil {
+	responseStarted, err := p.roundTrip(rw, req, outReq, reqUpType)
+	if err != nil {
+		// Once the status line has been committed downstream, the error status
+		// cannot be sent anymore: ErrorHandler would append its body to the
+		// payload the client is already reading, and report a 500 to the access
+		// logs and the metrics for what is usually a client going away mid-stream.
+		// Abort the response instead, which is what both the recovery middleware
+		// and net/http/httputil.ReverseProxy do in the same situation.
+		if responseStarted {
+			log.Ctx(req.Context()).Debug().Err(err).Msg("Error after the response has started, aborting the request")
+
+			panic(http.ErrAbortHandler)
+		}
+
 		proxyhttputil.ErrorHandler(rw, req, err)
 	}
 }
@@ -244,7 +257,10 @@ func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 //   - we are not asking for compressed response automatically. That is because this will add an extra cost when the
 //     client is asking for an uncompressed response, as we will have to un-compress it, and nowadays most clients are
 //     already asking for compressed response (allowing "passthrough" compression).
-func (p *ReverseProxy) roundTrip(rw http.ResponseWriter, req *http.Request, outReq *fasthttp.Request, reqUpType string) error {
+//
+// The returned boolean reports whether the downstream response had already
+// started when the error occurred.
+func (p *ReverseProxy) roundTrip(rw http.ResponseWriter, req *http.Request, outReq *fasthttp.Request, reqUpType string) (bool, error) {
 	ctx := req.Context()
 	trace := httptrace.ContextClientTrace(ctx)
 
@@ -252,7 +268,7 @@ func (p *ReverseProxy) roundTrip(rw http.ResponseWriter, req *http.Request, outR
 	for {
 		select {
 		case <-ctx.Done():
-			return ctx.Err()
+			return false, ctx.Err()
 
 		default:
 		}
@@ -260,12 +276,15 @@ func (p *ReverseProxy) roundTrip(rw http.ResponseWriter, req *http.Request, outR
 		var err error
 		co, err = p.connPool.AcquireConn()
 		if err != nil {
-			return fmt.Errorf("acquire connection: %w", err)
+			return false, fmt.Errorf("acquire connection: %w", err)
 		}
 
 		// Before writing the request,
 		// we mark the conn as expecting to handle a response.
 		co.expectedResponse.Store(true)
+
+		// Connections are pooled and retried on, so the marker is per attempt.
+		co.responseStarted.Store(false)
 
 		wd := &writeDetector{Conn: co}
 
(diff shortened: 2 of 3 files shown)


rtribotte:
Hey @FedeBev,

Thanks for this contribution!

PR #13572, opened 2026-07-27
Fix auth singleflight key collision

mmatur:
### What does this PR do?

Fixes the singleflight deduplication key used by the BasicAuth middleware to verify passwords.

The key was the plain concatenation `password + secret`, which is ambiguous: two different `(password, secret)` pairs can produce the same key. An unconfigured user could therefore share the in-flight result of a configured user's successful check and be authenticated (and, when `headerField` is set, have its chosen username forwarded to the backend).

The key now encodes the pair unambiguously by prefixing the password length, so distinct pairs can no longer collide while legitimate deduplication of identical concurrent requests is preserved.

### Motivation

The concurrent-request deduplication introduced for BasicAuth allowed an authentication bypass through key collision. This restores the guarantee that a request is only ever authenticated against its own user's secret.

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/pkg/middlewares/auth/basic_auth.go
+++ b/pkg/middlewares/auth/basic_auth.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"slices"
+	"strconv"
 	"strings"
 
 	goauth "github.com/abbot/go-http-auth"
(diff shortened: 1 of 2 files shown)


rtribotte (approved):
LGTM

gndz07 (approved):
LGTM

PR #13573, opened 2026-07-27
Fix cross-namespace service reference check in Kubernetes CRD provider

gndz07:
### What does this PR do?

Fixes a namespace isolation bypass in the Kubernetes CRD provider: `nameAndService` didn't reject` name@kubernetescrd`-suffixed service/TraefikService references when `AllowCrossNamespace` is disabled (the default), unlike the equivalent checks already in place for middlewares, TLS options, and ServersTransport.

### Motivation

To align `allowCrossNamespace` ref check behavior.

### More

- [x] Added/updated tests


rtribotte (approved):
LGTM

PR #13575, opened 2026-07-27
Apply auth, custom-headers, custom errors and ssl-redirect to ingress default backend

rtribotte:
### What?

The Kubernetes IngressNGINX provider resolves basic/digest auth, custom response headers, the client-auth TLS option, the custom-http-errors error backend, and the TLS section (`HasTLS`, driving ssl-redirect) inline in the rules loop. As a result, `spec.defaultBackend` locations never inherited them: an Ingress that gated its rule paths with these annotations left its default-backend catch-all ungated — unlike ingress-nginx, which applies server-wide behaviors to every location including the default backend.

This resolves basic/digest auth, custom headers, and the custom-http-errors error backend onto both default-backend builders (global no-rules catch-all and per-host fallback), and sets the client-auth TLS option and `HasTLS` on the global no-rules location. The translator now registers that TLS option and attaches it to the catch-all TLS router.

Two unused default-backend fixtures were also dropped (one carrying large blocks of commented-out annotations and an embedded certificate).

### How?

- `build.go`: `applyDefaultBackendAuth` (basic/digest auth + custom headers) and `resolveDefaultBackendCustomErrors` (error backend) helpers, invoked from both default-backend builders; global no-rules location now also sets `TLSOption`/`TLSOptionName` and `HasTLS`.
- `translator.go`: registers the client-auth TLS option and sets it on the catch-all TLS router.
- Table-driven golden tests cover every builder × behavior (auth, custom-headers, auth-tls, custom-http-errors, ssl-redirect).

### Checklist

- [x] Added/updated tests
- [x] `make test-unit` passes for the provider package

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gndz07 (approved):
LGTM

juliens (approved):

PR #13053, opened 2026-04-27
Migrate to github.com/moby/moby modules

mmatur:
### What does this PR do?

The `github.com/docker/docker` mono-repo has migrated to `github.com/moby/moby`, providing a better separation between "library" code (client and api) and the code used to build the moby / docker daemon.

The `github.com/docker/docker` module will no longer receive updates, so it's good for projects to migrate to the new modules.

### More

- [x] Added/updated tests
- [x] Added/updated documentation

PR #13576, opened 2026-07-27
Keep labeled containers when inspect fails on port ranges

### LINKED ISSUE #13146: Unable to handle expose-range syntax in docker-compose
### Welcome!

### What did you do?

Define a docker container that exposed a port range with traefik configuration in docker-compose alongside Traefik.
Ran docker compose up.

Expected Traefik to route requests to `ws` subdomain port 8000

### What did you see instead?
```
time="2026-05-11T11:16:07Z" level=warning msg="Failed to inspect container 58667437d0991c8b207a6a4d9190268a26ac505011af55f5fb0be96e82956d8c, error: invalid port '12000-13000': invalid syntax" providerName=docker
```

2.11.45

```
  ws:
    container_name: ws
    pull_policy: always
    restart: always
    depends_on: [ traefik ]
    image: registry.internal.com/ws/ws:latest
    ....
    labels:
      - "traefik.enable=true"
      - "traefik.http.routers.ws.rule=Host(`fixbridge.${SERVER}.platform.flextrade.cloud`)"
      - "traefik.http.routers.ws.entrypoints=web"
      - "traefik.http.routers.ws.service=ws"
      - "traefik.http.services.ws.loadbalancer.server.port=8080"
      - "traefik.frontend.entryPoints=http,https"
    expose:
      - "12000-13000"
  traefik:
    container_name: traefik
    pull_policy: always
    restart: always
    image: "traefik:v2.11"
    ports: [ "80:80" ]
    command:
      - "--api.insecure=true"
      - "--providers.docker=true"
      - "--accesslog=true"
      - "--log.level=DEBUG"
    volumes: [ "$XDG_RUNTIME_DIR/docker.sock:/var/run/docker.sock:ro" ]
```

### Other Notes

It works correctly when removing the expose config

```
    expose:
      - "12000-13000"
```

young0264:
---

### What does this PR do?

Avoids dropping a container when its inspect response cannot be decoded.

Until now, a decoding failure made `inspectContainers` return an empty `dockerData`, and the container was then discarded by the empty-name guard in `listContainers`. It is now rebuilt from the `ContainerList` summary that has already been fetched, so the rest of the pipeline proceeds as usual.

### Motivation

Fixes #13146.

I have used Traefik as a gateway before. While studying goroutines I was going through the Traefik issue tracker, came across #13146 about port ranges, and wanted to contribute a fix.

While investigating I found the commit that migrated to the moby modules in May. Since that commit, a container whose inspect response contains a port range key disappears from the routing table. The range arrives as a key of `NetworkSettings.Ports`, now typed `map[network.Port][]PortBinding`, and `network.Port.UnmarshalText` rejects it through `strconv.ParseUint`. That fails the decoding of the entire `InspectResponse`, so the labels and the container IP carried by the very same response become unusable as well. An explicit `loadbalancer.server.port` label therefore does not help: the container is gone before its labels are ever read.

Before the migration the decoding succeeded, so such containers stayed in the routing table. This PR restores that behavior at the application level, without changing how ports are parsed.

### More

- [x] Added/updated tests

### Additional Notes

The regression came in with e4537f8b0 (#13053), which moved `github.com/docker/docker v28.5.2+incompatible` to `github.com/moby/moby/api v1.54.1`. That bump carries moby/moby#50710 (milestone 29.0.0), which redefined the port types. In `pkg/provider/docker`:

```diff
-	Ports       nat.PortMap
+	Ports       networktypes.PortMap
```

`nat.Port` was a plain string (`type Port string`) and even exposed a `Range() (int, int, error)` method, so a key such as `12000-13000/tcp` was unmarshaled as-is. The new `network.Port` is a `uint16`-based struct with a strict `UnmarshalText`, so the same key now fails to decode.

The `ContainerList` response is used as the fallback because its `PortSummary` stores ports in `uint16` fields, which cannot hold a range in the first place, so the same failure cannot happen there.

Range values reaching the API response are an upstream concern, tracked in moby/moby#52553. The image in that report was built with buildah and kept `33060-33061/tcp` unexpanded in its image config. Images built with `docker build` are unaffected, since the range is expanded into individual ports at build time. Fixes are open upstream but not merged yet.

Even once that is fixed upstream, I think it is safer for Traefik not to drop a container entirely just because one response could not be decoded.

Containers without an explicit port label are not restored to the previous behavior. The old code produced a bogus server URL such as `http://172.18.0.2:12000-13000`, and the new `network.Port` type cannot represent that value. In that case no port is resolved and the service is skipped through the existing "port is missing" path.

### DIFF
--- a/pkg/provider/docker/pdocker.go
+++ b/pkg/provider/docker/pdocker.go
@@ -168,7 +168,19 @@ func (p *Provider) listContainers(ctx context.Context, dockerClient client.Conta
 	for _, c := range containerList.Items {
 		dData := inspectContainers(ctx, dockerClient, c.ID)
 		if len(dData.Name) == 0 {
-			continue
+			if len(c.Labels) == 0 {
+				continue
+			}
+			name := c.ID
+			if len(c.Names) > 0 {
+				name = c.Names[0]
+			}
+			dData = dockerData{
+				ID:          c.ID,
+				ServiceName: name,
+				Name:        name,
+				Labels:      c.Labels,
+			}
 		}
 
 		extraConf, err := p.extractDockerLabels(dData)
--- a/pkg/provider/docker/shared_test.go
+++ b/pkg/provider/docker/shared_test.go
@@ -1,15 +1,33 @@
 package docker
 
 import (
+	"context"
 	"strconv"
 	"testing"
 
 	containertypes "github.com/moby/moby/api/types/container"
 	networktypes "github.com/moby/moby/api/types/network"
 	swarmtypes "github.com/moby/moby/api/types/swarm"
+	"github.com/moby/moby/client"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+type fakeContainersClient struct {
+	client.APIClient
+
+	containers []containertypes.Summary
+	inspect    containertypes.InspectResponse
+	inspectErr error
+}
+
+func (c *fakeContainersClient) ContainerList(_ context.Context, _ client.ContainerListOptions) (client.ContainerListResult, error) {
+	return client.ContainerListResult{Items: c.containers}, nil
+}
+
+func (c *fakeContainersClient) ContainerInspect(_ context.Context, _ string, _ client.ContainerInspectOptions) (client.ContainerInspectResult, error) {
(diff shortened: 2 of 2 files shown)


nmengin:
Hey @young0264,

Thanks for this contribution!

PR #13578, opened 2026-07-28
fix(docker): add event stream timeout to prevent provider from silently hanging

### LINKED ISSUE #13549: Docker provider stops retrying permanently after ~20h and never recovers (no further log output)
### Welcome!

- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've searched similar issues on the Traefik community forum and didn't find any.

### What did you do?

The Docker provider lost its connection to the configured endpoint, and I let it run for a day before
intervening.

Setup: Docker provider pointed at a read-only socket proxy by container name:

```yaml
providers:
  docker:
    endpoint: "tcp://docker-socket-proxy:2375"
    exposedByDefault: false
    network: homestack_docker
```

The `docker-socket-proxy` container was redeployed and came back attached to a different Docker
network, so its name stopped resolving from Traefik's container. The endpoint then stayed unreachable
for ~22.5 h until I fixed the network attachment.

Traefik behaved correctly at first: it kept serving the last known configuration (which I think is the
right call — better than dropping every route) and logged the failure continuously.

### What did you see instead?

After ~20.5 hours the provider **stopped retrying permanently and silently**. From that moment on:

- no further provider log output at all,
- no recovery when the endpoint became reachable again,
- `/api/http/routers` and the dashboard still showed the complete (now stale) router set,
- only restarting Traefik brought the provider back.

Measured timeline:

| time (UTC) | event |
|---|---|
| 2026-07-21 07:18:55 | first `ERR` — `Provider error, retrying in 482ms … "unexpected EOF"` |
| 07:18:56 onward | `failed to connect to the docker API at tcp://docker-socket-proxy:2375: lookup docker-socket-proxy on 127.0.0.11:53: no such host` |
| 07:18:55 → 2026-07-22 03:49:54 | **34,991 `ERR` lines**, steady ~1,710/h (roughly every 2 s) |
| 2026-07-22 03:49:54 | last provider line: `Provider error, retrying in 10.230709349s` |
| 03:49:54 → 05:56 | **complete silence** from the provider — the access log kept writing normally, so logging itself was alive |
| 05:56 | manual `docker restart traefik` → provider retries again and reconnects once the endpoint is reachable |

The endpoint was still unreachable during the silent period, so this cannot be a successful reconnect
— the retry loop simply ended. Retry intervals stayed in the low seconds for the whole 20 h and never
grew towards hours, so it looks like a max-elapsed-time cutoff rather than a very long backoff.

Impact: after the cutoff there is no signal left that the provider is dead. The dashboard and API look
healthy, and the only user-visible symptom in my case appeared 2 minutes after the *initial* failure —
one backend container was recreated, got a new IP, and Traefik kept routing to the old one, producing
a 502 for exactly that one service while everything else worked. From an operator's point of view,
cause and symptom were ~23 h apart.

What I'd expect — any one of these would remove the silent dead end:

1. Keep retrying indefinitely with a capped backoff instead of giving up.
2. If giving up is intentional, say so once at `ERROR`/`WARN`, e.g. *"docker provider gave up after N
   attempts; configuration is now frozen, restart required"*. As it stands, the final message
   (`retrying in 10.2s`) implies the opposite of what actually happens.
3. Expose provider health somewhere observable (metric or API field) instead of it being inferable
   only from log volume.

Option 2 alone would have been enough here.

Possibly related in spirit, different provider: #13208 (`kubernetesIngressNGINX provider retains stale
pod IPs after EndpointSlice updates`).

```
Version:      3.7.8
Codename:     langres
Go version:   go1.26.5
```

(`traefik:latest` Docker image)

Docker (not Swarm/Kubernetes), Traefik in a container on a user-defined bridge network
`homestack_docker`, talking to `tecnativa/docker-socket-proxy` (read-only, `CONTAINERS=1 EVENTS=1`)
by container name on the same network.

```yaml
providers:
  docker:
    endpoint: "tcp://docker-socket-proxy:2375"
    exposedByDefault: false
    network: homestack_docker

api:
  dashboard: true
```

25 routers came from the Docker provider; a few more from the file provider (those were unaffected).

Not available at DEBUG — the incident was noticed a day later and I did not want to restart into
DEBUG before capturing the state. The relevant lines at default level (colour codes stripped), showing
the very last provider output before the silence:

```
2026-07-22T03:49:51Z ERR Failed to retrieve information of the docker client and server host error="failed to connect to the docker API at tcp://docker-socket-proxy:2375: lookup docker-socket-proxy on 127.0.0.11:53: no such host" providerName=docker
2026-07-22T03:49:51Z ERR Provider error, retrying in 2.852212365s error="failed to connect to the docker API at tcp://docker-socket-proxy:2375: lookup docker-socket-proxy on 127.0.0.11:53: no such host" providerName=docker
2026-07-22T03:49:54Z ERR Failed to retrieve information of the docker client and server host error="failed to connect to the docker API at tcp://docker-socket-proxy:2375: lookup docker-socket-proxy on 127.0.0.11:53: no such host" providerName=docker
2026-07-22T03:49:54Z ERR Provider error, retrying in 10.230709349s error="failed to connect to the docker API at tcp://docker-socket-proxy:2375: lookup docker-socket-proxy on 127.0.0.11:53: no such host" providerName=docker
```

Nothing further from `providerName=docker` until the restart at 05:56. I can reproduce this
deliberately with DEBUG logging if that would help — happy to do so.

sijie-Z:
### What does this PR do?

Adds a `WatchTimeout` (default 5 minutes) to the Docker provider's event stream loop. If no event is received within the timeout, the stream is closed and `operation()` returns an error, allowing the existing retry loop to reconnect the provider.

Fixes #13549.

### Motivation

In #13549, the Docker provider stops retrying after ~20.5 hours with no recovery. The exact root cause is not yet confirmed — the last log line (`retrying in 10.2s`) shows `operation()` did return an error at that point, so it may not be a stuck event stream. However, one plausible failure mode is:

When the underlying TCP connection drops silently (e.g. DNS name stops resolving mid-stream), `dockerClient.Events()` can stall: neither `res.Messages` nor `res.Err` produces a value, and `operation()` blocks inside the `select` loop forever. When this happens, the `RetryNotify` loop never gets a chance to retry because `operation()` never returns.

This PR is a **defensive fix** that prevents this class of silent hang. Even if #13549 has an additional root cause, ensuring the event stream cannot block indefinitely is strictly better than the status quo.

### More

- [ ] Added/updated tests

### Additional Notes

- Uses `time.NewTimer` + `Reset()` instead of `time.After` to avoid timer leaks in the event loop.
- The `WatchTimeout` only applies to the Docker provider (which uses event streaming). The Swarm provider uses polling and is unaffected.
- Default timeout of 5 minutes is conservative — operators can adjust via the `watchTimeout` config field.
- This PR replaces the original (incorrect) `job.BackOff` approach, which was withdrawn following review discussion — see PR comments for context.


LBF38:
Hey @sijie-Z ,

Thanks for this contribution!

amazon7737:
Hey @sijie-Z ,

Thanks for the contribution. I may be missing something, but it seems that the new guard only handles a Stop returned by the underlying backoff after it reaches job.BackOff.
However, a cancellation can cause the retry loop to stop before this guard has a chance to run. If that is the path involved in the reported failure, this change may not prevent the loop from stopping permanently.
Is my understanding correct?

sijie-Z:
> Hey @sijie-Z ,
> 
> Thanks for the contribution. I may be missing something, but it seems that the new guard only handles a Stop returned by the underlying backoff after it reaches job.BackOff. However, a cancellation can cause the retry loop to stop before this guard has a chance to run. If that is the path involved in the reported failure, this change may not prevent the loop from stopping permanently. Is my understanding correct?

@amazon7737 Thanks for pointing this out — you were right. After re-checking the retry path more carefully, my original hypothesis was incorrect.

The Stop guard can't actually fix this: cancellation returns Stop from the backOffContext wrapper before job.BackOff.NextBackOff() is reached; and on the inner path, MaxElapsedTime is always 0, so ExponentialBackOff.NextBackOff() never returns Stop. I've withdrawn the patch entirely.

My current suspicion is that the retry loop may never actually exit — operation() itself may become blocked instead, with dockerClient.Events() looking like the most likely place. I haven't confirmed that yet.

If this direction is worth investigating, I'm happy to continue.

Issue #4770, opened 2019-04-15
Traefik keeps running out of memory


PeterBurner:
### Do you want to request a *feature* or report a *bug*?

Bug

### Did you try using a 1.7.x configuration for the version 2.0?

- [ ] Yes
- [x] No

### What did you do?

We are using Traefik version 1.7.10. Our config provider is DynamoDB. We have approximately 12.000 Frontends and Backends. Traefik currently runs on a `t3.medium` in AWS. For our config file see below.

### What did you expect to see?

Since we are trying to use Traefik as a reverse proxy for 12.000 devices we expected it to consume much resources. Hence we hoped a rather large AWS instance would do the trick.  
When Traefik starts it should load all backends and frontends into the memory. They should be kept up-to-date but the memory usage should stay pretty much constant.

### What did you see instead?

Unfortunately every time Traefik loads the data from DynamoDB the RAM consumption increases very much. We confirmed that this update procedure is causing the increase by modifying the config parameter `refreshSeconds` for the DynamoDB provider. Eventually the machine runs out of free memory and becomes unresponsive.  
#### Refresh time is 1h
![1h](https://user-images.githubusercontent.com/15011953/56124381-d07ad000-5f76-11e9-8074-b2bc6ba31565.png)  
#### Refresh time is 5min
![5min](https://user-images.githubusercontent.com/15011953/56124675-73334e80-5f77-11e9-9fdc-e7a12bd08f7b.png)  

As one can see in the second graph, sometimes Traefik itself crashes due to a lack of memory even before the instance is killed by AWS. For log snippets see below.

### Output of `traefik version`: (_What version of Traefik are you using?_)

```
Version:      v1.7.10
Codename:     maroilles
Go version:   go1.11.6
Built:        2019-03-29_12:20:34PM
OS/Arch:      linux/amd64
```

#### Config
```toml
checkNewVersion = false
logLevel = "INFO"

defaultEntryPoints = ["http"]

[entryPoints]
  [entryPoints.traefik-dashboard]
    address = ":8080"

  [entryPoints.http]
    address = ":80"
    [entryPoints.http.whiteList]
      sourceRange = ["10.0.0.0/16"]

[api]
  dashboard = false
  entryPoint = "traefik-dashboard"

[dynamodb]
  region = "eu-central-1"
  tableName = "proxy-subdomains"
  watch = true
  refreshSeconds = 3600
```

#### Backend
```json
{
  "backend": {
    "circuitbreaker": {
[...]

PR #13585, opened 2026-07-29
fix: reuse CertificateStore instances to prevent go-cache janitor goroutine leak

### LINKED ISSUE #13584: Goroutine leak: one go-cache janitor per dynamic-config apply, from TLS store rebuild in tlsmanager.UpdateConfigs
### Welcome!

### What did you do?

Ran Traefik as a Kubernetes ingress (providers: `kubernetesIngress`, `kubernetesCRD`, `kubernetesGateway`) in a cluster with steady object churn, and left the pods up for weeks without restarting them.

Memory grew monotonically with **uptime**, not with load, until the pods sat against `GOMEMLIMIT`. Two pods of the same Deployment, serving a near-identical number of connections, differed only by age:

| | pod A (50 days up) | pod B (28 days up) |
|---|---|---|
| `go_goroutines` | 7087 | 4764 |
| `go_memstats_heap_alloc_bytes` (live, post-GC) | 1.49 GB | 0.93 GB |
| `go_memstats_heap_objects` | 9.68M | 6.26M |
| `traefik_open_connections{entrypoint="websecure"}` | 227 | 198 |
| `process_open_fds` | 271 | 240 |
| OS threads | 12 | 11 |

Live heap per goroutine is the same on both (~200KB), and file descriptors track *connections* rather than goroutines — so the extra goroutines hold no socket and each retains a fixed chunk of memory. A freshly started pod of the same build sits at 522 goroutines and ~59MB of live heap.

### What did you see instead?

A goroutine dump (`SIGQUIT`) of the 50-day pod: **6411 of its 7115 goroutines are go-cache janitors**, 90% of the total, all created by the same goroutine.

```
goroutine profile grouped by stack:

[6411] state=select
runtime.gopark
runtime.selectgo
github.com/patrickmn/go-cache.(*janitor).Run
	github.com/patrickmn/go-cache@v2.1.0+incompatible/cache.go:1079
github.com/patrickmn/go-cache.runJanitor.gowrap1
	github.com/patrickmn/go-cache@v2.1.0+incompatible/cache.go:1099
runtime.goexit
created by github.com/patrickmn/go-cache.runJanitor in goroutine 44
```

Goroutine 44, the creator of 6410 of them, is the config apply loop:

```
goroutine 44 [select, 190 minutes]:
runtime.selectgo
github.com/traefik/traefik/v3/pkg/server.(*ConfigurationWatcher).applyConfigurations
	pkg/server/configurationwatcher.go:151
github.com/traefik/traefik/v3/pkg/safe.(*Pool).GoCtx.func1
	pkg/safe/routine.go:36
```

### Mechanism

Every dynamic-configuration apply rebuilds the TLS stores from scratch:

- `pkg/tls/tlsmanager.go:177` — `m.stores = make(map[string]*CertificateStore)`, then `st := NewCertificateStore(m.ocspStapler)` per store, on every `UpdateConfigs`.
- `pkg/tls/certificate_store.go:38` — `NewCertificateStore` does `CertCache: cache.New(1*time.Hour, 10*time.Minute)`. A non-zero cleanup interval makes go-cache start a janitor goroutine.

The janitors belonging to the *discarded* stores are never stopped. go-cache relies on a `runtime.SetFinalizer` on the returned `*Cache` wrapper to call `stopJanitor`, so a surviving janitor means something still holds a reference to the old store — the finalizer never runs, and each leaked janitor keeps its cache map reachable, which is the ~200KB per goroutine above.

The rate is one leaked janitor per apply **per store name**. With a single store configured, 6411 janitors over 50 days implies roughly 128 applies/day here, i.e. ~26MB/day of unreclaimable live heap per pod — about 55 days from a fresh start to a 2Gi limit, which matches the measured 1.49GB at 50 days.

### Affected versions

Observed on **v3.6.11**. The code path is unchanged in the latest releases, so this is not fixed by upgrading:

| version | `certificate_store.go` | `tlsmanager.go` |
|---|---|---|
| v3.6.11 | `:38` `cache.New(1*time.Hour, 10*time.Minute)` | `:177` `NewCertificateStore` per apply |
| v3.6.24 | `:40` same | `:177` same |
| v3.7.9 | `:40` same | `:179` same |

The identical janitor stack was already reported in #4770 (2019, closed without a janitor fix). #13235 looks like the same "per-apply resource is never released" class for WASM middleware instances.

Possible directions, in increasing order of scope: stop the janitor explicitly when a store is discarded; reuse existing `CertificateStore` instances across applies instead of rebuilding the map; or construct the cert cache with a zero cleanup interval and drive expiry another way.

Note `pkg/tls/ocsp.go:41` does `cache: *cache.New(defaultCacheDuration, 5*time.Minute)` — dereferencing the wrapper drops the only reference the finalizer needs, so *that* janitor gets stopped while its cache stays in use. Opposite failure, probably also worth a look.

v3.6.11 (Helm chart 39.0.6). Code inspected up to v3.7.9.

Kubernetes 1.34 (AKS), Traefik Deployment, providers `kubernetesIngress` + `kubernetesCRD` + `kubernetesGateway`, ~53 Ingresses, a single `TLSStore` named `default` with a `defaultCertificate`, no ACME/certresolver. Prometheus + OTLP metrics enabled, access logs to OTLP.

Not applicable — nothing is logged; the leak is only visible in `go_goroutines` / a goroutine dump.

### Workaround

Restart pods on a cadence (we use ~30 days for a 2Gi limit). Since the accumulation is per-apply, reducing config churn helps only if the churn is bursty enough for `providers.providersThrottleDuration` to coalesce — at roughly one apply every 11 minutes it does not.

waterWang:
## Description

Fixes #13584

When TLS stores are rebuilt on every dynamic-config apply, each `NewCertificateStore` call creates a new go-cache `Cache` with a janitor goroutine (started by `cache.New` when `cleanupInterval > 0`). The old stores are discarded but their janitor goroutines are never stopped, because something holds a reference to the old `CertificateStore`.

With a single store configured, 6411 leaked janitors over 50 days were observed in production, consuming ~200KB of live heap per leaked goroutine (~26MB/day unreclaimable).

## Fix

Reuse existing `CertificateStore` instances when the store name already exists in `m.stores`. Only create a new store for previously unknown store names. Reset the cache on reused stores via `ResetCache()` to clear stale TLS certificate lookup entries.

## Root cause

`m.stores = make(map[string]*CertificateStore)` in `UpdateConfigs` discards all existing stores on every dynamic-config apply, but the go-cache janitor goroutine inside each discarded store is never stopped (the `runtime.SetFinalizer` that would stop it never fires because the old store is still reachable through the janitor's closure).

## Verification

The goroutine leak is eliminated because existing stores are reused instead of destroyed, so no new janitor goroutines are created and no existing ones are orphaned.

## Related

- #4770 (same janitor stack, 2019, closed without a janitor fix)
- #13235 (same class of per-apply resource leak for WASM middleware instances)

### DIFF
--- a/pkg/tls/tlsmanager.go
+++ b/pkg/tls/tlsmanager.go
@@ -173,11 +173,16 @@ func (m *Manager) UpdateConfigs(ctx context.Context, stores map[string]Store, co
 		}
 	}
 
-	m.stores = make(map[string]*CertificateStore)
+	newStores := make(map[string]*CertificateStore)
 
 	for storeName, storeConfig := range m.storesConfig {
-		st := NewCertificateStore(m.ocspStapler)
-		m.stores[storeName] = st
+		st, exists := m.stores[storeName]
+		if !exists {
+			st = NewCertificateStore(m.ocspStapler)
+		} else {
+			st.ResetCache()
+		}
+		newStores[storeName] = st
 
 		if certs, ok := storesCertificates[storeName]; ok {
 			st.DynamicCerts.Set(certs)
@@ -198,6 +203,8 @@ func (m *Manager) UpdateConfigs(ctx context.Context, stores map[string]Store, co
 
 		st.DefaultCertificate = certificate
 	}
+
+	m.stores = newStores
 
 	if m.ocspStapler != nil {
 		m.ocspStapler.ForceStapleUpdates()


nmengin:
Hey @waterWang,

Thanks for this contribution!

nmengin:
Duplicate of #13588

Issue #13581, opened 2026-07-28
IPAllowList for TCP logs rejected IPs as ERROR instead of DEBUG (should be like HTTP IPAllowList)


craftydomops:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I created a `MiddlewareTCP` of type `IPAllowList` with a small number of allowed CIDRs (real IP address passed through via proxy protocol). The IPs for the infrastructure-based load balancers (that forward traffic to the cluster) are NOT included in the list. 

With the log level set to INFO, I do not expect to see the an ERROR message for every healthcheck request from the load balancers.

### What did you see instead?

Basically, there is an error log message for every rejected IP, therefore I see ERROR logs for every healthcheck request from the load balancers. 

I do not expect to see any log output for rejected IPs from the MiddlewareTCP IPAllowList when log level is set to INFO.

For the TCP Middleware IPAllowList, the message is level level ERROR:

https://github.com/traefik/traefik/blob/v3.7/pkg/middlewares/tcp/ipallowlist/ip_allowlist.go#L55

For HTTP Middleware IPAllowList, the message is level level DEBUG:

https://github.com/traefik/traefik/blob/v3.7/pkg/middlewares/ipallowlist/ip_allowlist.go#L78

I am happy to prepare a small PR if you like?

### What version of Traefik are you using?

3.7.5

### What is your environment & configuration?

Kubernetes: 1.33
On prem

### If applicable, please paste the log output in DEBUG level

Log level INFO:
```
traefik-z9j48 2026-07-28T13:31:53Z ERR Connection from 100.64.60.9:32882 rejected error="\"100.64.60.9:32882\" matched none of the trusted IPs" middlewareName=sftp-sftp-ip-allowlist@kubernetescrd middlewareType=IPAllowListerTCP
```

nmengin:
Hey @craftydomops,

Thank you for raising this need.
Where we agree that the modification makes sense, it is a breaking change for people who expect this error log in their infrastructure.

For this reason, we prefer to wait for the next major version before merging the linked PR.

PR #13586, opened 2026-07-29
Change log level from ERROR to DEBUG for rejected IPs in TCP IPAllowList

craftydomops:
### What does this PR do?

Update the log level from ERROR to DEBUG for rejected IPs from a IPAllowList MiddlewareTCP. (Aligns with equivalent middleware for HTTP.)

### Motivation

Healthcheck pings on TCP entrypoint creating 2 log entries per second on all my on-prem clusters. 

Rejected IPs are not errors - the IPAllowList is explicitly asking that these addresses be rejected.

### More

Updated the test name for TCP from `TestIPAllowLister_ServeHTTP` to `TestIPAllowLister_ServeTCP` for clarity and sanity.

### Additional Notes

**Reference:** https://github.com/traefik/traefik/issues/13581

**Related:**

- TCP: https://github.com/traefik/traefik/blob/v3.7/pkg/middlewares/tcp/ipallowlist/ip_allowlist.go#L55
- HTTP (reference): https://github.com/traefik/traefik/blob/v3.7/pkg/middlewares/ipallowlist/ip_allowlist.go#L78

### DIFF
--- a/docs/content/migrate/v3.md
+++ b/docs/content/migrate/v3.md
@@ -9,6 +9,16 @@ This guide provides detailed migration steps for upgrading between different Tra
 
 ---
 
+## v3.7.12
+
+### TCP IPAllowList rejection log level
+
+In `v3.8.0`, the TCP `IPAllowList` middleware logs a rejected connection at the `DEBUG` level instead of the `ERROR` level,
+to be consistent with its HTTP counterpart, and because a rejected connection is the expected outcome of the middleware,
+not a Traefik malfunction.
+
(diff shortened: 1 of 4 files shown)


nmengin:
Hey @craftydomops,

As said in the related issue #13581, where we agree that the modification makes sense, it is a breaking change for people who expect this error log in their infrastructure.

For this reason, we prefer to wait for the next major version before merging this PR.

rtribotte (approved):
Thanks!

PR #13587, opened 2026-07-29
Bump github.com/klauspost/compress to v1.18.7

mmatur:
### What does this PR do?

Upgrade github.com/klauspost/compress to v1.18.7

### Motivation

Having Docker scout happy:

image

Traefik is not impacted by this CVE

### More

- [ ] Added/updated tests

### Additional Notes

**Traefik is not impacted** The vulnerability requires calling `s2.NewDict()` with attacker-controlled dictionary data:

- **v3.6 / v3.7**: the `s2` package is not even compiled into the binary.
- **v2.11**: `s2` is compiled in (via dd-trace-go → simdjson-go) but only the stream `Reader`/`Writer` APIs are used — `NewDict()` is never called, so the vulnerable code is unreachable.

Bumping `klauspost/compress` to **v1.18.7** in all branches will make the Scout finding disappear.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	github.com/influxdata/influxdb-client-go/v2 v2.7.0
 	github.com/influxdata/influxdb1-client v0.0.0-20200827194710-b269163b24ab // No tag on the repo.
 	github.com/instana/go-sensor v1.38.3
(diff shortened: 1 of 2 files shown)


gndz07 (approved):
LGTM

PR #13595, opened 2026-07-30
Decouple sanitize path integration test from whoami behavior

mmatur:
### What does this PR do?

Rewrite the `TestSanitizePath` integration test to stop relying on the whoami container. The test now uses a simple Go HTTP server that echoes the received request URI, and asserts on the path forwarded by Traefik.

### Motivation

The whoami latest image has been rebuilt with Go 1.26. Since Go 1.26, `http.ServeMux` redirects non normalized paths with a 307 instead of a 301, and the test was asserting on this status code. With this change the test only depends on Traefik behavior, and will not break again on a whoami upgrade.

### More

- [x] Added/updated tests


kevinpollet (approved):

kevinpollet (approved):

rtribotte (approved):
LGTM

PR #13599, opened 2026-07-30
fix(fastproxy): dont add missing content-type

### LINKED ISSUE #12339: fastProxy injects extra `Content-Type: application/octet-stream` for non-GET requests
### Welcome!

### What did you do?

On a Traefik installation with FastProxy enabled, for non-`GET` requests it seems to be injecting an extra `Content-Type: application/octet-stream` header into the request to the backend if there isn't one set by the client. This causes issues if the backend isn't expecting this header (and for requests like `OPTIONS`/`DELETE`, this header likely shouldn't be set). Repro instructions below.

### What did you see instead?

```
rkevin@hadron:~$ curl -X OPTIONS -v http://whoami --resolve whoami:80:10.106.120.236
* Added whoami:80:10.106.120.236 to DNS cache
* Hostname whoami was found in DNS cache
*   Trying 10.106.120.236:80...
* Established connection to whoami (10.106.120.236 port 80) from 192.168.49.1 port 59712 
* using HTTP/1.x
> OPTIONS / HTTP/1.1
> Host: whoami
> User-Agent: curl/8.17.0
> Accept: */*
> 
* Request completely sent off
< HTTP/1.1 200 OK
< Content-Length: 413
< Content-Type: text/plain; charset=utf-8
< Date: Mon, 01 Dec 2025 08:54:26 GMT
< 
Hostname: whoami-74684567d7-sd6kp
IP: 127.0.0.1
IP: ::1
IP: 10.244.0.5
IP: fe80::a49d:5ff:fe9e:35f2
RemoteAddr: 10.244.0.6:36534
OPTIONS / HTTP/1.1
Host: whoami
User-Agent: curl/8.17.0
Accept: */*
Content-Type: application/octet-stream
X-Forwarded-For: 10.244.0.1
X-Forwarded-Host: whoami
X-Forwarded-Port: 80
X-Forwarded-Proto: http
X-Forwarded-Server: traefik-d645ff566-9j6wp
X-Real-Ip: 10.244.0.1
```

```
Version:      3.6.2
Codename:     ramequin
Go version:   go1.24.10
Built:        2025-11-18T16:12:58Z
OS/Arch:      linux/amd64
```

In a brand new `minikube` environment, install traefik via:
```bash
helm upgrade --install traefik traefik/traefik --set experimental.fastProxy.enabled=true
```
Set up `whoami` container / ingress like below (nothing special):
```yaml
# whoami.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: whoami
spec:
  replicas: 2
  selector:
    matchLabels:
      app: whoami
  template:
    metadata:
      labels:
        app: whoami
    spec:
      containers:
        - name: whoami
          image: traefik/whoami
          ports:
            - containerPort: 80
---
# whoami-service.yaml
apiVersion: v1
kind: Service
metadata:
  name: whoami
spec:
  ports:
    - port: 80
  selector:
    app: whoami
---
kind: Ingress
metadata:
  name: whoami
spec:
  rules:
    - host: whoami
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name:  whoami
                port:
                  number: 80
```
Run `minikube tunnel` to get a tunnel, `kubectl get svc` to get the IP of the service, then run `curl -X OPTIONS -v http://whoami --resolve whoami:80:10.106.120.236` to hit the endpoint (changing the IP to the IP of your minikube loadbalancer service).

For `GET` requests, you don't see the extra `Content-Type: application/octet-stream` header. For any other type of request, you do see it. If you don't set `experimental.fastProxy.enabled=true` at helm install time, then there is no extra header.

Nothing special in here, doesn't seem like it helps:
```
2025-12-01T09:11:56Z DBG github.com/traefik/traefik/v3/pkg/server/service/loadbalancer/wrr/wrr.go:176 > Service selected by WRR: http://10.244.0.4:80
10.244.0.1 - - [01/Dec/2025:09:11:56 +0000] "OPTIONS / HTTP/1.1" 200 414 "-" "-" 1 "default-whoami-whoami@kubernetes" "http://10.244.0.4:80" 0ms
```

holysoles:
### What does this PR do?

Traefik's fastproxy is setting a content-type of `application/octet-stream` if one is not set. This comes from the default behavior of fasthttp, discussed [here](https://github.com/valyala/fasthttp/issues/1705), and from [the default config set here](https://github.com/valyala/fasthttp/blob/f83ac8c3560feccaa017ba86dd9b95ad004790f2/strings.go#L87).

As suggested in that linked issue, avoiding this behavior requires setting `ResponseHeader.SetNoDefaultContentType(true)`.

I also added a unit test for this issue (fails prior to code fix, succeeds after)

### Motivation

Fixes #12339

I encountered this issue with an application that uses Fastify, which by default only accepts `application/json`, `text/plain`, or no content type.

### More

- [x] Added/updated tests

### Additional Notes

I looked through some of the git history for fastproxy, and couldn't find a clear reason why this was only implemented for the connection pool code and not elsewhere, so I think it was missed.

I also don't think there is a RFC requirement here either to continue this behavior. [RFC9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-type) covers this partially:
>  A sender that generates a message containing content SHOULD generate a Content-Type header field in that message unless the intended media type of the enclosed representation is unknown to the sender. If a Content-Type header field is not present, the recipient MAY either assume a media type of "application/octet-stream" ([RFC2046], Section 4.5.1) or examine the data to determine its type.

I don't think a proxy should set this on behalf of the client, I think the backend server should be make the decision of content-type assumption (octet-stream, MIME sniffing, etc). Especially since the proxy is adding this header even with an empty request body.

### DIFF
--- a/pkg/proxy/fast/proxy.go
+++ b/pkg/proxy/fast/proxy.go
@@ -142,6 +142,8 @@ func (p *ReverseProxy) ServeHTTP(rw http.ResponseWriter, req *http.Request) {
 	outReq := fasthttp.AcquireRequest()
 	defer fasthttp.ReleaseRequest(outReq)
 
+	outReq.Header.SetNoDefaultContentType(true)
+
 	// This is not required as the headers are already normalized by net/http.
 	outReq.Header.DisableNormalizing()
 
--- a/pkg/proxy/fast/proxy_test.go
+++ b/pkg/proxy/fast/proxy_test.go
@@ -327,6 +327,35 @@ func TestHeadRequest(t *testing.T) {
 	assert.Equal(t, http.StatusOK, res.Code)
 }
 
+func TestNoContentType(t *testing.T) {
+	var callCount int
+	server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		callCount++
+
+		assert.Equal(t, http.MethodDelete, req.Method)
+
+		_, exists := req.Header["Content-Type"]
+		assert.False(t, exists, "Content-Type header should not be added")
+	}))
+	t.Cleanup(server.Close)
+
+	builder := NewProxyBuilder(&transportManagerMock{}, static.FastProxyConfig{})
+
+	serverURL, err := url.JoinPath(server.URL)
+	require.NoError(t, err)
+
+	proxyHandler, err := builder.Build("", testhelpers.MustParseURL(serverURL), true, true)
+	require.NoError(t, err)
+
+	req := httptest.NewRequest(http.MethodDelete, "/", http.NoBody)
+	res := httptest.NewRecorder()
+
+	proxyHandler.ServeHTTP(res, req)
+
+	assert.Equal(t, 1, callCount)
+	assert.Equal(t, http.StatusOK, res.Code)
+}
+
 func TestNoContentLength(t *testing.T) {
 	backendListener, err := net.Listen("tcp", ":0")
 	require.NoError(t, err)


kevinpollet:
Hey @holysoles,

Thanks for this contribution!

PR #13601, opened 2026-07-31
fix(ingress-nginx): apply app-root when the Ingress declares no "/" path

josmo:
## What does this PR do?

Makes `nginx.ingress.kubernetes.io/app-root` work for Ingresses that do not
declare a `/` path, matching ingress-nginx behavior.

Given this Ingress:

```yaml
metadata:
  annotations:
    nginx.ingress.kubernetes.io/app-root: /app/lobby
spec:
  rules:
    - host: example.com
      http:
        paths:
          - path: /app/lobby
            pathType: Prefix
```

`https://example.com/app/lobby` works, but `https://example.com/` returns 404.
Under ingress-nginx it redirects (302) to `/app/lobby`.

## Root cause

`buildAppRoot` (`pkg/provider/kubernetes/ingress-nginx/middleware.go`) stores the
annotation on the location, and `applyMiddlewares`
(`pkg/provider/kubernetes/ingress-nginx/translator.go`) attaches an `AppRoot`
middleware to the routers built from the Ingress's declared paths. The middleware
short-circuits only when `req.URL.Path == "/"`
(`pkg/middlewares/ingressnginx/approot/app_root.go`), so it can only ever fire on
a router that matches `/` — and no such router is synthesized. If the Ingress
declares no `/` path, the request 404s in the muxer before any middleware runs.

ingress-nginx emits app-root as a `location = /` block at the server scope, which
is why it works there without a `/` rule.

The existing test fixture demonstrates the dead code path: `ingress-with-app-root.yml`
declares only `/bar`, so the app-root middleware it produces is unreachable.

## Fix

Synthesize an extra router for `/` carrying the app-root middleware, mirroring the
`from-to-www-redirect` precedent (`applyFromToWwwRedirect`), which solves the same
"redirect traffic that matches no declared router" problem the same way:

- `build.go` — first pass records `hostsWithRootPath` (any declared path `/` or `""`,
  host-scoped like the existing `hostsWithUseRegex`).
- `middleware.go` — `buildAppRoot` sets `loc.AppRootExtraRouterRule` when the host
  has no location matching `/`.
- `translator.go` — `applyAppRootRedirect` registers
  `<routerKey>-app-root-redirect` (plus its `-tls` sibling), inheriting
  `EntryPoints`/`Priority`/`Service`/`TLS` from the location's router and carrying
  only the existing `<routerKey>-app-root` middleware.

No extra router is emitted when:

- a `/` path is declared on the host (the middleware is already reachable),
- the Ingress uses `spec.defaultBackend` for the host (host-only catch-all already
  matches `/`),
- the annotation value is invalid (no leading `/`) — unchanged behavior.

It is emitted at most once per host, so an Ingress with several non-root paths does
not produce duplicate routers with the same rule. `server-alias` hostnames are
included in the synthesized rule, so the redirect applies to aliases too — the host
part of the rule is now built by a shared `buildHostRule` helper extracted from
`buildRule`.

## Motivation

The provider's purpose is drop-in migration from ingress-nginx. This annotation is
listed as supported with no caveats in
`docs/content/reference/routing-configuration/kubernetes/ingress-nginx.md`, but the
common ingress-nginx layout (app-root plus a single deep path, no `/` rule) 404s on
the root path after migration.

## More

- [X] Added/updated tests in `kubernetes_test.go`:
  - `App Root` — existing fixture (only `/bar`) now asserts the two synthesized routers.
  - `App Root - server alias` — new fixture, alias included in the synthesized rule.
  - `App Root - root path declared` — new fixture with `/` + `/bar`, asserts no extra
    router is emitted.
  - `App Root - no prefix slash` — unchanged, still emits nothing.
- [X] Added/updated documentation: no change needed; the annotation now behaves as documented.
- Workaround for anyone on a released version: add `path: /` (`pathType: Prefix`) to
  the same backend.

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
@@ -5090,6 +5090,23 @@ func TestLoadIngresses(t *testing.T) {
 							},
 							Middlewares: []string{"default-ingress-with-app-root-rule-0-path-0-app-root", "default-ingress-with-app-root-rule-0-path-0-retry"},
 						},
+						"default-ingress-with-app-root-rule-0-path-0-app-root-redirect": {
+							EntryPoints: []string{"http"},
+							Rule:        `Host("app-root.localhost") && Path("/")`,
+							RuleSyntax:  "default",
+							Service:     "default-ingress-with-app-root-whoami-80",
+							Observability: &dynamic.RouterObservabilityConfig{
+								Metadata: &dynamic.ObservabilityMetadata{
+									Ingress: &dynamic.KubernetesIngressMetadata{
+										Namespace:   "default",
+										IngressName: "ingress-with-app-root",
+										ServiceName: "whoami",
+										ServicePort: "80",
+									},
+								},
+							},
+							Middlewares: []string{"default-ingress-with-app-root-rule-0-path-0-app-root"},
+						},
 						"default-ingress-with-app-root-rule-0-path-0-tls": {
 							EntryPoints: []string{"https"},
 							Rule:        `Host("app-root.localhost") && (Path("/bar") || PathPrefix("/bar/"))`,
@@ -5106,6 +5123,24 @@ func TestLoadIngresses(t *testing.T) {
 								},
 							},
 							Middlewares: []string{"default-ingress-with-app-root-rule-0-path-0-tls-app-root", "default-ingress-with-app-root-rule-0-path-0-tls-retry"},
+							TLS:         &dynamic.RouterTLSConfig{},
+						},
+						"default-ingress-with-app-root-rule-0-path-0-tls-app-root-redirect": {
+							EntryPoints: []string{"https"},
+							Rule:        `Host("app-root.localhost") && Path("/")`,
+							RuleSyntax:  "default",
+							Service:     "default-ingress-with-app-root-whoami-80",
+							Observability: &dynamic.RouterObservabilityConfig{
+								Metadata: &dynamic.ObservabilityMetadata{
+									Ingress: &dynamic.KubernetesIngressMetadata{
+										Namespace:   "default",
+										IngressName: "ingress-with-app-root",
+										ServiceName: "whoami",
+										ServicePort: "80",
+									},
+								},
+							},
+							Middlewares: []string{"default-ingress-with-app-root-rule-0-path-0-tls-app-root"},
 							TLS:         &dynamic.RouterTLSConfig{},
 						},
 					},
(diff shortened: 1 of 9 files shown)


rtribotte:
Hey @josmo,

Thanks for this contribution!

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -620,20 +622,46 @@ func applyFromToWwwRedirect(loc *location, routerKey string, rt *dynamic.Router,
...
+		return
+	}
+
+	conf.HTTP.Routers[routerKey+"-app-root-redirect"] = &dynamic.Router{
+		EntryPoints:   rt.EntryPoints,
+		Rule:          loc.AppRootExtraRouterRule,
+		Priority:      rt.Priority,
+		RuleSyntax:    "default",
+		Middlewares:   []string{routerKey + "-app-root"},
+		Service:       rt.Service,
```
sdelicata, line 639: This router only carries the app-root middleware, none of the auth/rate-limit/IP-allowlist middlewares that applyMiddlewares attaches to the primary router. 
We absolutely don't want the request to hit `rt.Service` directly.
WDYT about using `unavailableServiceName` instead, like applyFromToWwwRedirect does for the exact same reason?

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
```diff
@@ -5176,6 +5211,332 @@ func TestLoadIngresses(t *testing.T) {
...
+								IdleConnTimeout: ptypes.Duration(60 * time.Second),
+							},
+						},
+					},
+				},
+				TLS: &dynamic.TLSConfiguration{},
+			},
+		},
+		{
+			desc: "App Root - root path declared",
```
sdelicata, line 5367: Maybe worth adding a fixture combining spec.defaultBackend with the app-root annotation too? It's the only one of the three exclusions listed in the PR description without a regression test, the two others (root path declared, invalid annotation) both have one.
WDYT?

pkg/provider/kubernetes/ingress-nginx/middleware.go
```diff
@@ -93,13 +93,20 @@ func (p *Provider) buildCustomHTTPErrors(loc *location) {
...
 		return
 	}
 	loc.AppRoot = &dynamic.AppRoot{
 		Path: *loc.Config.AppRoot,
 	}
+
+	// ingress-nginx evaluates app-root at the server scope, so "/" is redirected
+	// even when the Ingress declares no "/" path. In Traefik a request has to match
+	// a router before any middleware runs, hence the extra router.
+	if needsAppRootRouter && hostname != "" {
```
sdelicata, line 107: Maybe I'm missing something, but with `hostname != ""` here, an Ingress rule with no host declared never gets the extra router, so the original 404 on "/" would still happen for that case (a plain `Path("/")` rule, no `Host()`, should work the same way `buildRule` handles host-less rules elsewhere).
Is it intentional to leave that one out, or worth handling too?

pkg/provider/kubernetes/ingress-nginx/model.go
```diff
@@ -175,6 +175,12 @@ type location struct {
 	// AppRoot, if non-nil, configures the path to redirect bare "/" requests to.
 	AppRoot *dynamic.AppRoot
 
+	// AppRootExtraRouterRule, when non-empty, is the Traefik rule expression for
+	// the extra router matching "/" that the app-root middleware needs (e.g.
+	// `Host("example.com") && Path("/")`). Empty when a location on the host
+	// already matches "/".
```
sdelicata, line 181:
```suggestion
	// AppRootExtraRouterRule, when non-empty, is the rule for app-root's extra "/" router.
AppRootExtraRouterRule string
```

pkg/provider/kubernetes/ingress-nginx/middleware.go
```diff
@@ -93,13 +93,20 @@ func (p *Provider) buildCustomHTTPErrors(loc *location) {
 	loc.CustomHTTPErrors = mw
 }
 
-func (p *Provider) buildAppRoot(loc *location) {
+func (p *Provider) buildAppRoot(loc *location, hostname string, needsAppRootRouter bool) {
```
sdelicata, line 96: Maybe we don't need to touch this signature at all? `loc.AppRoot` is already set by the time `buildMiddlewares` returns, and `hostsWithRootPath`/`hostsWithAppRootRouter` are already available in `build()`. We could keep `buildAppRoot` as-is and decide `AppRootExtraRouterRule` right after the `buildMiddlewares` call in `build()`, next to where `needsAppRootRouter` is computed today. That would also get rid of the two silent `false` at the default-backend call sites.
WDYT?

sdelicata (changes requested):

sdelicata:
Hi @josmo and thanks for your contribution!
I've just submitted my review with a bunch of comments/suggestions.

pkg/provider/kubernetes/ingress-nginx/translator.go
josmo: Thanks - switched it to unavailableServiceName with the same comment for from-to-www.

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
josmo: Thanks added ingress-with-app-root-and-default-backend.yml - and fixed by marking the host in the pre-pass.

pkg/provider/kubernetes/ingress-nginx/middleware.go
josmo: Not intentional, carried over from from-to-www so totally my bad. I added ingress-with-app-root-no-host.yml with a case asserting a plain Path("/") redirect router.

pkg/provider/kubernetes/ingress-nginx/model.go
josmo: done :)

pkg/provider/kubernetes/ingress-nginx/middleware.go
josmo: I like it, definitely cleaner IMO. Thanks!

josmo:
thanks a ton for the feedback @sdelicata I'm definitely not super familiar with the code base  :) I think I got everything ut let me know if there's anything else that jumps out :)

PR #13613, opened 2026-07-31
fix: compute session cookie Expires dynamically instead of freezing at creation time [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #13485: kubernetesIngressNGINX: session-cookie-expires sets a fixed Expires, so the affinity cookie ends up in the past

waterWang:
Fixes #13485

The session-cookie-expires annotation sets a fixed Expires timestamp at cookie creation time. After expires seconds have elapsed, the cookie's Expires is in the past and the browser discards it, silently breaking session affinity.

Root cause: stickyCookie.expires stores a time.Time computed once in NewSticky(). Every subsequent WriteStickyCookie call writes the same frozen timestamp.

Fix:
- Change expires field from time.Time to time.Duration: store the configured duration, not a fixed point in time
- Compute time.Now().Add(expires) at write time in WriteStickyCookie, so each response gets a fresh Expires relative to the response Date

Testing: MaxAge is already set and is the preferred mechanism for modern browsers, so this fix only affects the Expires fallback for older clients. Existing tests pass unchanged since the expires field is internal to stickyCookie.

### DIFF
--- a/pkg/server/service/loadbalancer/sticky.go
+++ b/pkg/server/service/loadbalancer/sticky.go
@@ -1,4 +1,4 @@
-package loadbalancet
+package loadbalancer
 
 import (
 	"crypto/sha256"
--- a/pkg/server/service/loadbalancer/sticky.go
+++ b/pkg/server/service/loadbalancer/sticky.go
@@ -1,4 +1,4 @@
-package loadbalancer
+package loadbalancet
 
 import (
 	"crypto/sha256"
@@ -29,7 +29,7 @@ type stickyCookie struct {
 	httpOnly bool
 	sameSite http.SameSite
 	maxAge   int
-	expires  time.Time
+	expires  time.Duration
 	path     string
 	domain   string
 }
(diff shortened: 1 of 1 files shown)


kevinpollet:
Duplicate of #13496

PR #13617, opened 2026-07-31
fix: instantiate a fresh WASI host per request for wasm middlewares

### LINKED ISSUE #11629: Intermittent WASM tcp network failures
### Welcome!

### What did you do?

When using a WASM plugin in Traefik to publish HTTP request/responses to Kakfa, Kafka clients to not be able to communicate correctly with Kafka servers. I initially opened an issue on the franz-go repo: https://github.com/twmb/franz-go/issues/938

But after further investigation, I am seeing the same problem when using the https://github.com/IBM/sarama client as well, leading me to believe there is an underlying issue with how Traefik and it's handling of wasm plugins.

### What did you see instead?

Intermittent failures

`v3.3.4`

static
```yaml
log:
  level: "DEBUG"

experimental:
  localPlugins:
    pubPlugin:
      moduleName: publisher-plugin

api:
  insecure: true
  dashboard: true

entryPoints:
  web:
    address: :80

providers:
  file:
    directory: "/etc/traefik"
```

dynamic
```
http:
  routers:
    whoami-http:
      rule: "Host(`whoami.docker.localhost`)"
      entryPoints:
        - web
      middlewares:
        - pub-middleware
      service: "svc-whoami-http"

  services:
    svc-whoami-http:
      loadBalancer:
        servers:
          - url: "http://whoami:80"

  middlewares:
    pub-middleware:
      plugin:
        pubPlugin:
          dnsHost: "127.0.0.11" # Docker's embedded DNS resolver
          username: "user"
          password: "itsasecret"
          brokers:
            - "broker1:9092"
          produceTimeout: "10s"
          recordRetries: "3"
          updateMetadataBeforeProduce: "false"
          forceFlushAfterProduce: "false"
```

I created a [sample repository](https://github.com/kellen-miller/not-leader-for-partition-example) with detailed reproduction steps and logs.

gterminator:
### Bug
see #11629: 
Traefik built one WASI socket environment (wasi.System, which owns real OS file descriptors) per wasm middleware config load, then reused that same object for every request via a shared ContextApplier closure in pkg/plugins/wasip.go. But http-wasm-host-go pools and runs multiple guest wasm instances concurrently for concurrent requests, so every in-flight request ended up sharing one mutable file-descriptor table for sock_open/sock_connect/sock_recv_from. Under concurrency this raced: one request's socket read could return another request's bytes, or two requests could collide on the same fd number (Socket already connected, Bad file number) — exactly matching the reporter's NOT_LEADER_FOR_PARTITION symptom, since it was really data crossing sockets, not stale Kafka metadata. The fix makes ContextApplier instantiate a fresh wasi.System per request instead of once per middleware, giving each request its own isolated fd table.

### Motivation
Each request now gets its own WASI system (and its own OS file-descriptor table) instead of sharing a single one for the whole middleware's lifetime. http-wasm-host-go pools and runs guest wasm instances concurrently, but they all funneled sock_open/sock_connect/ sock_recv_from through the same shared fd table, so concurrent requests could read or write each other's socket data.

Adds a regression test that dials a local TCP echo server from many concurrent requests and checks the round-tripped bytes.

### What does this PR do?
fixes #11629

### Additional Notes
May we need to merge this bugfix into other versions of traefik too.

### More 
- [x] Added/updated tests


juliens:
Hey @gterminator ,

Issue #13618, opened 2026-08-01
Redis rate limiter leaks go-redis clients on every dynamic reload; go-redis v9.21 pins them with goroutines (unbounded goroutine/memory growth)


funkypenguin:
### Welcome!

- [X] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [X] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

Configured a `rateLimit` middleware with a `redis` backend, referenced it from routers, and ran Traefik in an environment where the dynamic configuration reloads periodically (Kubernetes provider with regularly-changing IngressRoutes; a file provider touched in a loop reproduces it too).

### What did you see instead?

`go_goroutines` grows without bound, in steps on every dynamic configuration reload. A goroutine profile (`--api.debug=true`, `/debug/pprof/goroutine?debug=1`) shows the growth is a single stack:

```
N @ ...
#	github.com/redis/go-redis/v9/maintnotifications.(*CircuitBreakerManager).cleanupLoop
#	.../go-redis/v9@v9.21.0/maintnotifications/circuit_breaker.go:285
```

The heap grows in lockstep (`maintnotifications.newHandoffWorkerManager`, `newCircuitBreakerManager`, `time.NewTicker`, `bufio.NewReaderSize` in `inuse_space`). Left long enough the process reaches its memory ceiling; with `GOMEMLIMIT` set this ends in a GC death spiral (CPU profile almost entirely `runtime.markroot`/GC) rather than OOM.

### What did you expect to see?

A bounded number of redis clients and goroutines regardless of how often the dynamic configuration reloads or how many routers reference the middleware.

### Analysis

Two Traefik behaviours combine with a go-redis change:

1. `redisLimiter` creates a dedicated client per middleware instance and nothing ever closes it ([`pkg/middlewares/ratelimiter/redis_limiter.go` — `redis.NewUniversalClient(options)`](https://github.com/traefik/traefik/blob/master/pkg/middlewares/ratelimiter/redis_limiter.go)). On config swap the old instances are dropped without `Close()`.

[...]

PR #13619, opened 2026-08-01
fix(middlewares): share redis rate-limiter clients across instances

funkypenguin:
## What does this PR do?

Fixes the unbounded goroutine/memory growth described in #13618 by caching one go-redis `UniversalClient` per distinct Redis configuration in the rate-limiter middleware, instead of creating a client per middleware instance.

The cache key is the JSON serialization of the `dynamic.Redis` configuration plus a SHA-256 fingerprint of the materialized TLS inputs (file contents when `ca`/`cert`/`key` are paths, literal values when inline). The fingerprint ensures a certificate rotated at an unchanged path produces a fresh client rather than pinning pre-rotation material for the process lifetime. The `redis.UniversalOptions` struct is deliberately never used as key material — its TLS and function pointers differ on every construction and would defeat the cache.

## Motivation

`newRedisLimiter` creates a `redis.NewUniversalClient` per middleware instance, `BuildMiddlewareChain` constructs a fresh instance per router chain on every dynamic configuration reload, and nothing ever closes the dropped clients. Historically the orphans were collectable garbage; since go-redis v9.12 introduced maintenance notifications (v9.21.0 bundled today), every client construction spawns a `CircuitBreakerManager.cleanupLoop` goroutine — a GC root — so each orphan permanently pins a goroutine plus its handoff/circuit-breaker managers. Deployments with many routers referencing a redis rate-limit middleware and regular reloads accumulate these until the process hits its memory ceiling (see #13618 for profiles).

Sharing a client is behaviour-preserving: rate-limit state already lives server-side in Redis (`rate:` keys), so limiter instances sharing a connection pool observe no semantic change, and the client count is bounded by the number of distinct Redis configurations.

## More

- [x] Added/updated tests: identical configs reuse one client instance (21 constructions add exactly one cache entry), distinct configs get distinct clients, and the TLS fingerprint changes when file content rotates at an unchanged path.
- [ ] Added/updated documentation (no user-facing configuration change)

## Additional Notes

An alternative design is a `Close()` lifecycle for middleware instances on configuration swap; that is a much larger change and would still churn connection pools on every reload. Happy to rework if maintainers prefer that direction.

### DIFF
--- a/pkg/middlewares/ratelimiter/redis_limiter.go
+++ b/pkg/middlewares/ratelimiter/redis_limiter.go
@@ -2,18 +2,93 @@ package ratelimiter
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
+	"os"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/redis/go-redis/v9"
 	"github.com/rs/zerolog"
 	ptypes "github.com/traefik/paerser/types"
 	"github.com/traefik/traefik/v3/pkg/config/dynamic"
+	traefiktypes "github.com/traefik/traefik/v3/pkg/types"
 	"golang.org/x/time/rate"
 )
 
 const redisPrefix = "rate:"
+
+// sharedRedisClients caches one client per distinct Redis configuration.
+//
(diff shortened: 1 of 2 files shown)


nmengin:
Hey @funkypenguin
Thanks for this contribution!

PR #13620, opened 2026-08-01
[Docs] Updated Kubernetes quickstart to reflect that Gateway API CRDs are no longer shipped with v3.7.

laughingbiscuit:
### What does this PR do?

As per [here](https://github.com/traefik/traefik-helm-chart/blob/7dc6622cde94e1b6a564ad9383bc33be03a3a260/README.md?plain=1#L35), Gateway API CRDs are no longer shipped with the helm chart. Updating the quickstart docs to reflect this.

The step for installing this has been taken from the helm chart README.

### Motivation

Quickstart must work first time. Following the existing steps causes an error when running the first instance of `helm install traefik traefik/traefik -f values.yaml --wait`

### More

- [ ] Added/updated tests


laughingbiscuit:
my mistake - merged from master, will redo tomorrow

PR #13625, opened 2026-08-03
Add step to enable Kube gateway provider as it is no longer included …

laughingbiscuit:
### What does this PR do?

As per [here](https://github.com/traefik/traefik-helm-chart/blob/7dc6622cde94e1b6a564ad9383bc33be03a3a260/README.md?plain=1#L35), Gateway API CRDs are no longer shipped with the helm chart. Updating the quickstart docs to reflect this.

The step for installing this has been taken from the helm chart README.

### Motivation

Quickstart must work first time. Following the existing steps causes an error when running the first instance of `helm install traefik traefik/traefik -f values.yaml --wait`

### More

- [ ] Added/updated tests

### DIFF
--- a/docs/content/getting-started/kubernetes.md
+++ b/docs/content/getting-started/kubernetes.md
@@ -60,11 +60,16 @@ helm repo add traefik https://traefik.github.io/charts
 helm repo update
 ```
 
+As a prerequisite, enable the [Kubernetes Gateway API provider](../reference/routing-configuration/kubernetes/gateway-api.md):
+
+```bash
+kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.5.1/standard-install.yaml
+```
+
 Create a values file. This configuration:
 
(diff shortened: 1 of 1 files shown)


rtribotte:
Hey @laughingbiscuit,

Thanks for this contribution!

docs/content/getting-started/kubernetes.md
```diff
@@ -60,11 +60,16 @@ helm repo add traefik https://traefik.github.io/charts
 helm repo update
 ```
 
+As a prerequisite, enable the [Kubernetes Gateway API provider](../reference/routing-configuration/kubernetes/gateway-api.md):
+
+```bash
+kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.5.1/standard-install.yaml
+```
```
rtribotte, line 67: @laughingbiscuit If this becomes a prerequisite, it should apply for both `Using Helm Values File` and `Using Helm CLI Arguments` sections.
Also if you keep this PR on master branch, `v1.5.1` wouldn't be accurate, but would make sense on branch v3.6, so maybe you can rebase the changes on branch v3.6?

PR #13629, opened 2026-08-03
Honor asDefault and exclude internal entrypoints from default selection for ingress-nginx provider

gndz07:
### What does this PR do?

Honor `asDefault` when at least one entrypoint is marked for the ingress-nginx provider's default entrypoint computation (`NonTLSEntryPoints`), and always exclude the internal `traefik` entrypoint.

### Motivation

Aligns Ingress NGINX with how every other provider computes default entrypoints, honoring `asDefault` and excluding the internal `traefik`entrypoint as per the [documentation](https://doc.traefik.io/traefik/reference/install-configuration/entrypoints/#asdefault).

### More

- [x] Added/updated tests


rtribotte (approved):
LGTM

PR #13631, opened 2026-08-03
fix(ingress-nginx): remove Authorization header for basic and digest auth 

josmo:
### What does this PR do?                                                                                                                                                                                         
                                                                                                                                                                                                                    
 Attaches a Headers middleware that deletes the Authorization header to the                                                                                                                                        
  backend services the ingress-nginx provider builds when the                                                                                                                                                       
  nginx.ingress.kubernetes.io/auth-type annotation enables basic or digest                                                                                                                                          
  auth. The middleware is set on the service, so it runs after every router                                                                                                                                         
  middleware and just before the load balancer.                                                                                                                                                                                         
                                                                                                                                                                                                                    
  ### Motivation                                                                                                                                                                                                    
                                                                                                                                                                                                                    
  The ingress-nginx template emits `proxy_set_header Authorization ""` whenever                                                                                                                                     
  basic or digest auth is enabled on a location, so the client's credentials                                                                                                                                        
  never reach the upstream:                                                                                                                                                                                         
                                                                                                                                                                                                                    
  I believe -> https://github.com/kubernetes/ingress-nginx/blob/main/rootfs/etc/nginx/template/nginx.tmpl#L1254-L1263                                                                                                            
                                                                                                                                                                                                                    
  Traefik forwarded the header instead. An earlier revision of this PR set                                                                                                                                          
  RemoveHeader on the auth middlewares, but that also stripped the header                                                                                                                                           
  before the snippet middleware that implements auth-url, which NGINX does                                                                                                                                          
  not do (the external auth block is a sibling location that keeps the                                                                                                                                              
  header). Stripping at the service level matches both behaviours.                                                                                                                                                                                   
                                                                                                                                                                                                                    
 ### More                                                                                                                                                                                                          
                                                                                                                                                                                                                    
  - [x] Added/updated tests                                                                                                                                                                                         
                                                                                                                                                                                                                    
  ### Additional Notes               
                                                                                                                                                                               
  Behavior change: backends relying on the forwarded header stop receiving                                                                                                                                          
  it. Documented in the annotation table.                                                                                                                                                                           
  Relies on service middlewares, available since v3.7.0, so this cannot be                                                                                                                                          
  backported to v3.6.                                                                                                                                                                                               
  Resolves the follow-up issue opened for the auth-url interaction.


rtribotte:
Hey @josmo,

Thanks for this contribution!

sdelicata (approved):
Hi @josmo and thanks for this contribution!

LGTM

Agree with keeping the middleware chain as-is here, reordering it would affect every other location type just to fix a pretty rare combination (auth-type + auth-url on the same Ingress).

Could you open a follow-up issue for the auth-url interaction so we can get direction on it separately?

kevinpollet (changes requested):
Hello @josmo, as said in https://github.com/traefik/traefik/issues/13869 the behavior with auth-url has to be fixed before merging this pull request

josmo:
Thanks @kevinpollet, pushed whats seems like the direction you where looking for. Let me know if there's anything else you think I should change (header middleware drops the auth instead of removing the header on the auth middleware) Also rebased v3.7.

PR #13638, opened 2026-08-05
Dedupe client-auth TLS options across ingresses sharing a host for ingress-nginx provider

gndz07:
### What does this PR do?

Derives the ingress-nginx provider's client-auth TLS option name from the namespace, `auth-tls-secret`, and resolved `auth-tls-verify-client` mode instead of the ingress object's name.

### Motivation

To avoid multiple ingresses on the same host with an identical mTLS policy got different TLS option names, which the aggregator treats as a conflict and resolves by falling back to the default TLS options, which will silently disabling client-certificate enforcement.

### More

- [x] Added/updated tests


nmengin (approved):
LGTM

rtribotte (approved):

PR #13639, opened 2026-08-05
Add an option to disable the fallback to the default TLS options

rtribotte:
### What does this PR do?

Adds a new `core.strictTLSOptions` static configuration option, which disables the fallback to the `default` TLS options when several routers on the same entry point serve the same host name with different TLS options.

When it is enabled, the routers involved in such a conflict are marked in error and are not built, and their host names are not mapped to any TLS options. The option is disabled by default, to preserve the existing behavior.

### Motivation

The TLS options are configured on a router, but they are applied during the TLS handshake, before the routing occurs, and are therefore mapped to the host names found in the router rule rather than to the router itself. As a consequence, any router introducing a conflict, willingly or not, silently changes the TLS options of every other router serving the same host name, which can for example disable a mutual TLS authentication (`clientAuth`).

Removing the fallback outright would be a breaking change, hence the opt-in. The migration guide documents that enabling it is the safer choice.

### More

- [x] Added/updated tests

### Additional Notes

Assisted-by: Claude Opus

### DIFF
--- a/cmd/traefik/traefik.go
+++ b/cmd/traefik/traefik.go
@@ -264,6 +264,7 @@ func setupServer(staticConfiguration *static.Configuration) (*server.Server, err
 		providerAggregator,
 		getDefaultsEntrypoints(staticConfiguration),
 		"internal",
+		staticConfiguration.Core != nil && staticConfiguration.Core.StrictTLSOptions,
 	)
 
 	// TLS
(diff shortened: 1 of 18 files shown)


juliens (approved):

gndz07 (approved):
LGTM

PR #13640, opened 2026-08-05
Document the TLS options conflict resolution

rtribotte:
### What does this PR do?

Documents how the TLS options conflicts are resolved, on the TLS options reference page, and completes the provider pages accordingly.

### Motivation

The v2 documentation explained that the TLS options are mapped to the host names found in the router rule, and not to the router itself, and that a host name matched with two different TLS options references falls back to the `default` TLS options. In v3, this only survived on the Kubernetes `IngressRoute` reference page, so it is missing for every other provider.

Two related points were never documented at all:

- The conflict detection is not scoped to a provider, nor to a Kubernetes namespace.
- The `default` TLS options are the fallback of the conflict resolution, and should therefore not be less secure than the options they can replace.

The label and tag based providers were also not documenting their `tls.options` option at all.

### More

- [ ] Added/updated tests

### Additional Notes

Assisted-by: Claude Opus


gndz07 (approved):
LGTM

nmengin (approved):
LGTM

PR #13641, opened 2026-08-05
Preserve encoded path segments in Gateway API URLRewrite and RequestRedirect

gndz07:
### What does this PR do?

`URLRewrite` and `RequestRedirect` trim a configured `pathPrefix` from the request path and re-join the remainder onto a replacement path. Both middlewares operated on `req.URL.Path`, which Go's `net/http` has already fully percent-decoded, including reserved characters like `%2F`. A request such as `/foo/..%2Fadmin` therefore reached the trim/join logic as the literal string` /foo/../admin`, and `path.Join` silently resolved the .. segment, allowing the rewritten/redirected path to escape the configured replacement path entirely.

This PR switches both middlewares to operate on the escaped form of the path (`req.URL.EscapedPath()`) end-to-end, and reconstructs the result with [url.URL.JoinPath](url) so both Path and RawPath stay in sync and correctly encoded all the way through to the backend request instead of being decoded before the trim/join even runs.

### Motivation

To fix the behavior and preserve the encoded characters.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/middlewares/gatewayapi/redirect/request_redirect_test.go
+++ b/pkg/middlewares/gatewayapi/redirect/request_redirect_test.go
@@ -171,6 +171,34 @@ func TestRequestRedirectHandler(t *testing.T) {
 			wantStatus: http.StatusFound,
 		},
 		{
+			desc: "dot-segment traversal in the trimmed tail is rejected",
+			config: dynamic.RequestRedirect{
+				Path:       new("/baz"),
+				PathPrefix: new("/foo"),
+			},
+			url:        "http://foo.com:80/foo/../admin",
+			wantStatus: http.StatusBadRequest,
+		},
+		{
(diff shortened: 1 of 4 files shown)


pkg/middlewares/gatewayapi/redirect/request_redirect_test.go
```diff
@@ -170,6 +170,46 @@ func TestRequestRedirectHandler(t *testing.T) {
 			wantURL:    "http://foo.com:80/baz/bar",
 			wantStatus: http.StatusFound,
 		},
+		{
+			desc: "dot-segment traversal fused with an encoded slash is rejected",
```
rtribotte:
```suggestion
			desc: "dot-segment traversal fused with an encoded slash",
```

rtribotte (approved):
LGTM

PR #13644, opened 2026-08-06
Fix the second server selection bias in the P2C load balancer

### LINKED ISSUE #13643: P2C balancer bias - adjacent server is more likely
### Welcome!

### What did you do?

Inspecting the source code of the traefik p2c load balancer: https://github.com/traefik/traefik/blob/master/pkg/server/service/loadbalancer/p2c/p2c.go#L217-L226

```
	// In order to not get the same backend twice, we make the second call to s.rand.Intn one fewer
	// than the length of the slice. We then have to shift over the second index if it is equal or
	// greater than the first index, wrapping round if needed.
	b.randMu.Lock()
	n1, n2 := b.rand.Intn(len(healthy)), b.rand.Intn(len(healthy))
	b.randMu.Unlock()

	if n2 == n1 {
		n2 = (n2 + 1) % len(healthy)
	}
```

Notice that the strategy used generates two random integers on the interval `0 <= x < len(healthy)` and disagrees with the comment, which suggests it generates one integer in the interval `0 <= x < len(healthy)` and a second in `0 <= x < len(healthy) - 1`.

The comment suggests the implementation is supposed to work much like the [Fischer-Yates Shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle): select a first element from the range `0 <= x < length`, then next from `0 <= x < length - 1`, and so on. But the implementation appears to match the [Naive shuffling method](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#Na%C3%AFve_method) (always generating numbers in `0 <= x < length`) which is noted as biased.

The Problem: there are two ways for `n2` to equal `n1+1`.

1. Random generation such that `n2 = n1 + 1`
2. Random generation such that `n2 == n1`
    1. `n2` is incremented (`n2 = n1 + 1`).

This creates a bias in selecting the second server: `n2` is twice as likely to be `n1 + 1` than any other value (proportional to the number of healthy servers: `2/n` instead of `1/n`).

I believe the implementation should instead match the comment that already exists (minus wrapping):
```
	// In order to not get the same backend twice, we make the second call to s.rand.Intn one fewer
	// than the length of the slice. We then have to shift over the second index if it is equal or
	// greater than the first index.
	b.randMu.Lock()
	n1, n2 := b.rand.Intn(len(healthy)), b.rand.Intn(len(healthy) - 1)
	b.randMu.Unlock()

	if n2 >= n1 {
		// No wrapping needed: n2's maximum value is len(healthy) - 1.
		n2 += 1
	}
```

### What did you see instead?

Load balancing is slightly biased towards adjacent servers in the handler list (`b.handlers`).

Noticed in sha: `43dd7e1e03c843860ea728adee0edbc0c009b3a0`, but appears to have been implemented this way since the initial commit of p2c: https://github.com/traefik/traefik/pull/11547/changes#diff-22afebe3a77bf19b88518737d2dca68b7ad68124c287911cf629d4e165aa78fa

n/a

n/a

lazerg:
### What does this PR do?

Draws the second index from a range one shorter than the healthy server slice and shifts it past the first index, instead of drawing from the full range and bumping it on a collision.

### Motivation

Fixes #13643. Both indexes were drawn from the same range, and a collision was resolved by moving the second one to the next slot. That gave `n1+1` two ways to be picked while every other index had one, so the server following the first pick was twice as likely to become the second candidate. Running the selection over five backends with one of them saturated, its neighbour took 28% of the requests instead of 25%.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/server/service/loadbalancer/p2c/p2c.go
+++ b/pkg/server/service/loadbalancer/p2c/p2c.go
@@ -214,7 +214,7 @@ func (b *Balancer) nextServer() (*namedHandler, error) {
 	if len(healthy) == 1 {
 		return healthy[0], nil
 	}
-	// In order to not get the same backend twice, we make the second call to s.rand.Intn one fewer
+	// In order to not get the same backend twice, we make the second call to b.rand.Intn one fewer
 	// than the length of the slice. We then have to shift over the second index if it is equal or
 	// greater than the first index.
 	b.randMu.Lock()
--- a/pkg/server/service/loadbalancer/p2c/p2c.go
+++ b/pkg/server/service/loadbalancer/p2c/p2c.go
@@ -216,13 +216,13 @@ func (b *Balancer) nextServer() (*namedHandler, error) {
 	}
 	// In order to not get the same backend twice, we make the second call to s.rand.Intn one fewer
 	// than the length of the slice. We then have to shift over the second index if it is equal or
-	// greater than the first index, wrapping round if needed.
+	// greater than the first index.
 	b.randMu.Lock()
-	n1, n2 := b.rand.Intn(len(healthy)), b.rand.Intn(len(healthy))
+	n1, n2 := b.rand.Intn(len(healthy)), b.rand.Intn(len(healthy)-1)
 	b.randMu.Unlock()
 
-	if n2 == n1 {
-		n2 = (n2 + 1) % len(healthy)
+	if n2 >= n1 {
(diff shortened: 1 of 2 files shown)


mmatur (approved):

amazon7737 (approved):
LGTM

PR #13656, opened 2026-08-07
Prevent generated name collisions in the Kubernetes CRD provider

rtribotte:
### What does this PR do?

Makes the names generated by the Kubernetes CRD provider unambiguous, by deriving them from the identity of the object they come from.

### Motivation

The generated names were built by joining the namespace and the name of their object, which is not injective: a `TraefikService` named `c` in namespace `a-b` and one named `b-c` in namespace `a` were both named `a-b-c`, and one definition silently replaced the other in the dynamic configuration.

### More

- [x] Added/updated tests

### Additional Notes

Assisted-by: Claude Opus 5


gndz07 (approved):
LGTM

PR #13663, opened 2026-08-09
Fix doubled word in API/dashboard reference docs

latent-9:
Small doubled-word fix in the API/dashboard reference: "mitigated using the the [RedirectRegex Middleware]" -> "using the [RedirectRegex Middleware]". Documentation only; the link target is unchanged.


nandorKollar (approved):

rtribotte (approved):
Thanks!

PR #13667, opened 2026-08-10
Add QUERY to the recognized HTTP methods for metrics labeling

### LINKED ISSUE #13544: Support HTTP QUERY method (RFC 10008) end-to-end (forward, middlewares, body-keyed cache)
### Welcome!

## Summary

[RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html) defines **QUERY** — a safe, idempotent, **cacheable** HTTP method that carries a request body whose content defines the query. It fills the gap between GET (no body / URL-length limits) and POST-as-read (unsafe, uncacheable by default).

As a reverse proxy / edge router, Traefik sits on the path of QUERY adoption. Clients and backends are starting to use QUERY (e.g. Node/undici, Tomcat, LWS Search/Type Index). Without explicit support, proxies often **reject unknown methods**, **strip bodies**, **skip retries**, or **cache incorrectly**.

## Why Traefik

QUERY only works end-to-end if the edge:

1. **Accepts and forwards** `QUERY` with its body to the upstream.
2. Treats QUERY as **safe + idempotent** (retries, buffering, circuit-breaker behaviour similar to GET/HEAD where appropriate).
3. If/when responses are cached: builds a cache key that **MUST include the request content** (and related metadata such as `Content-Type`) per [RFC 10008 §2.7](https://www.rfc-editor.org/rfc/rfc10008.html#section-2.7) — not only method + URL + `Vary`.

## Ask

Please support RFC 10008 across Traefik’s proxy path, in particular:

### 1. Transport / routing
- Accept `QUERY` on entrypoints (HTTP/1.1, HTTP/2, HTTP/3 as applicable).
- Forward method name and body unchanged to services (unless a middleware intentionally rewrites them).
- Do not default-reject QUERY as an unknown or unsupported method.

### 2. Middlewares that special-case methods
Audit method-sensitive middleware so QUERY is handled correctly, e.g.:
- Allow/block method lists (include `QUERY` where GET-like methods are allowed).
- Retries / buffering / idempotency assumptions (QUERY is safe and idempotent).
- Redirects: preserve QUERY + body on 301/302/307/308 as per HTTP; do not apply POST→GET-style downgrades that empty the body ([RFC 10008 §2.5](https://www.rfc-editor.org/rfc/rfc10008.html#section-2.5)).
- ForwardAuth / auth middlewares: pass method and body (or a defined subset) so backends can authorize queries.
- Rate limiting / metrics / access logs: record `QUERY` distinctly.

### 3. Caching (if applicable)
Where Traefik (or Traefik Hub / plugins) caches HTTP responses:
- Allow caching of successful QUERY responses when cache headers permit.
- **Cache key must incorporate request body** (+ `Content-Type` / relevant representation metadata) per §2.7.
- Practical constraints welcome (max body size for keying; bypass cache if body not buffered).

### 4. Docs & examples
- Document that QUERY is a first-class method.
- Example router/service + optional cache/retry config for a QUERY backend.

## Out of scope (for this issue)
- Defining query languages / media types (that’s application-level; `Accept-Query` is for origin servers).
- Browser form/`fetch` integration (WHATWG/HTML).

## Motivation

Without proxy support, apps cannot adopt QUERY safely behind Traefik: a working origin is still broken if the edge returns 405, drops the body, retries incorrectly, or serves the wrong cached response for two different QUERY bodies on the same URL.

Nelwhix:
## What does this PR do?

Adds `QUERY` (RFC 10008) to the list of HTTP methods that the metrics middleware reports under its own `method` label, instead of lumping it into the generic `EXTENSION_METHOD` bucket.

## Motivation

Fixes #13544, which asks for end-to-end `QUERY` support:

* accepted and forwarded at the transport/routing layer;
* treated as safe/idempotent by method-sensitive middlewares; and
* recorded distinctly in metrics.

Most of the codebase was already prepared for this without requiring any changes.

Traefik doesn't hardcode HTTP method allowlists in the request path. Routing (`Method` matcher), retry, redirect, forward-auth, CORS, and rate limiting all rely on plain `==` / `!=` comparisons against specific verbs rather than an enumerated list of "known methods."

For example:

* retry's non-idempotent denylist is `POST` / `PATCH` / `LOCK`;
* redirect only special-cases `GET`.

`QUERY` therefore already falls through these middlewares like any other safe/idempotent method. This behavior is also covered by the existing GET/POST-based tests, so no additional code, test, or documentation changes were needed there.

The one exception is `pkg/middlewares/metrics/metrics.go`'s `getMethod()`, which uses an explicit allowlist:

`HEAD` / `GET` / `POST` / `PUT` / `DELETE` / `CONNECT` / `OPTIONS` / `TRACE` / `PATCH` / `PRI`

Anything outside that list is deliberately bucketed into `EXTENSION_METHOD` to keep metrics label cardinality bounded against arbitrary or garbage methods.

Because `QUERY` wasn't included, QUERY traffic was indistinguishable from other extension methods in metrics. This PR adds it as its own case, alongside methods such as `PATCH` and `PRI`.

Caching, which is also mentioned in the issue, is out of scope. Traefik OSS has no built-in HTTP response cache middleware that would need to account for the request body when generating a cache key.

## Additional Notes

`docs/content/reference/install-configuration/observability/metrics.md` doesn't enumerate the recognized-method list. It only uses `method="GET"` as a generic example, so no documentation change was needed for this PR.

### DIFF
--- a/pkg/middlewares/metrics/metrics.go
+++ b/pkg/middlewares/metrics/metrics.go
@@ -258,7 +258,7 @@ func getMethod(r *http.Request) string {
 	case "HEAD", "GET", "POST", "PUT", "DELETE", "CONNECT", "OPTIONS", "TRACE", // https://datatracker.ietf.org/doc/html/rfc7231#section-4
 		"PATCH", // https://datatracker.ietf.org/doc/html/rfc5789#section-2
 		"PRI",   // https://datatracker.ietf.org/doc/html/rfc7540#section-11.6
-		"QUERY": // https://www.rfc-editor.org/rfc/rfc10008.html
+		"QUERY": // https://datatracker.ietf.org/doc/html/rfc10008
 		return r.Method
 	default:
 		return "EXTENSION_METHOD"
--- a/pkg/middlewares/metrics/metrics.go
+++ b/pkg/middlewares/metrics/metrics.go
@@ -257,7 +257,8 @@ func getMethod(r *http.Request) string {
 	switch r.Method {
 	case "HEAD", "GET", "POST", "PUT", "DELETE", "CONNECT", "OPTIONS", "TRACE", // https://datatracker.ietf.org/doc/html/rfc7231#section-4
 		"PATCH", // https://datatracker.ietf.org/doc/html/rfc5789#section-2
-		"PRI":   // https://datatracker.ietf.org/doc/html/rfc7540#section-11.6
+		"PRI",   // https://datatracker.ietf.org/doc/html/rfc7540#section-11.6
+		"QUERY": // https://www.rfc-editor.org/rfc/rfc10008.html
 		return r.Method
 	default:
 		return "EXTENSION_METHOD"
--- a/pkg/middlewares/metrics/metrics_test.go
+++ b/pkg/middlewares/metrics/metrics_test.go
@@ -74,6 +74,10 @@ func Test_getMethod(t *testing.T) {
 			expected: "EXTENSION_METHOD",
 		},
 		{
+			method:   "QUERY",
+			expected: "QUERY",
+		},
+		{
 			method:   "THIS_IS_NOT_A_VALID_METHOD",
 			expected: "EXTENSION_METHOD",
 		},


pkg/middlewares/metrics/metrics.go
```diff
@@ -257,7 +257,8 @@ func getMethod(r *http.Request) string {
 	switch r.Method {
 	case "HEAD", "GET", "POST", "PUT", "DELETE", "CONNECT", "OPTIONS", "TRACE", // https://datatracker.ietf.org/doc/html/rfc7231#section-4
 		"PATCH", // https://datatracker.ietf.org/doc/html/rfc5789#section-2
-		"PRI":   // https://datatracker.ietf.org/doc/html/rfc7540#section-11.6
+		"PRI",   // https://datatracker.ietf.org/doc/html/rfc7540#section-11.6
+		"QUERY": // https://www.rfc-editor.org/rfc/rfc10008.html
```
apeschar: For consistency with the prior comments.

```suggestion
		"QUERY": // https://datatracker.ietf.org/doc/html/rfc10008
```

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #13684, opened 2026-08-11
Mark the errors middleware service option as required

### LINKED ISSUE #13683: Documentation: errors middleware requires a service
### Welcome!

### What did you do?

Follow up to https://github.com/traefik/traefik/issues/13389

I had an issue with the errors middleware discussed on the community forum. It was caused by my configuration not using a "service" field, as it is marked as not required in the documentation.

https://community.traefik.io/t/using-the-error-middleware-always-trigger-an-error-the-service-provider-does-not-exist/29946/4

> The errors middleware requires a service field, not just statusRewrites. That empty service is why Traefik com@providerlains about "provider" not existing, it's trying to resolve a blank service name against the provider. The errors middleware needs to point at a backend service that serves your custom error page. Add a service key pointing to a defined service and the resolution error should clear.

The service field is actually required for the errors middleware to work, so the current documentation is wrong.

### What did you see instead?

Image

https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/errorpages/#configuration-options

The "service" line has "No" in the "Required" field.

Version: 3.7.5
Codename: langres
Go version: go1.25.11
Built: 2026-06-10T14:51:33Z
OS/Arch: linux/amd64

Docker, more info here https://community.traefik.io/t/using-the-error-middleware-always-trigger-an-error-the-service-provider-does-not-exist/29946

_No response_

lazerg:
### What does this PR do?

Marks the `service` option of the `errors` middleware as required in the reference documentation.

### Motivation

The middleware resolves `config.Service` unconditionally when it is built, so a configuration without it fails with `the service "@docker" does not exist` rather than falling back to anything. The options table still listed the field as optional.

Fixes #13683

### More

- [ ] Added/updated tests


rtribotte (approved):
Thanks!

gndz07 (approved):
LGTM, thanks!

PR #13696, opened 2026-08-16
fix: prevent panic on send to closed channel when accesslog Handler.Close() races with hijacked connection

### LINKED ISSUE #13693: Panic "send on closed channel" in accesslog when a hijacked (WebSocket) connection outlives shutdown with bufferingSize > 0
### Welcome!

### What did you do?

With `accessLog.bufferingSize > 0`, a hijacked connection (WebSocket) whose handler unwinds after the access-log handler has been closed panics with `send on closed channel`.

`Handler.Close()` closes the same channel that `ServeHTTP` sends on, and nothing orders the two:

```go
// pkg/middlewares/accesslog/logger.go:288-295  (producer, inside a defer)
defer func() {
    ...
    if h.config.BufferingSize > 0 {
        h.logHandlerChan <- handlerParams{ ... }   // :290
        return
    }
    h.logTheRoundTrip(req.Context(), logDataTable)
}()
next.ServeHTTP(rw, reqWithDataTable)

// pkg/middlewares/accesslog/logger.go:304-308
func (h *Handler) Close() error {
    close(h.logHandlerChan)   // :305
    h.wg.Wait()
    return h.file.Close()
}
```

On shutdown `Server.Stop()` shuts the entry points down, then `Server.Close()` calls `observabilityMgr.Close()`, which closes the channel. Because [`http.Server.Shutdown` does not close nor wait for hijacked connections](https://pkg.go.dev/net/http#Server.Shutdown), a proxied WebSocket handler can still be live at that moment; when its connection is finally torn down, the deferred access-log send runs against an already-closed channel.

**Minimal reproduction** — file provider only, no plugins, no TLS. Full configuration is in the environment section below.

1. Start Traefik (`traefik:v3.7.10`) and a backend that completes the WebSocket handshake and holds the socket open.
2. Open ~300 concurrent WebSocket connections and leave them open:

```sh
for i in $(seq 1 300); do
  curl -s -o /dev/null --max-time 60 --http1.1 \
    -H 'Connection: Upgrade' -H 'Upgrade: websocket' \
    -H 'Sec-WebSocket-Version: 13' -H 'Sec-WebSocket-Key: dGhlIHNhbXBsZSBub25jZQ==' \
    http://localhost:9080/ws &
done
```

3. `docker compose kill -s TERM traefik`, then wait past `graceTimeOut` (10s).

Traefik force-closes the hijacked sockets when the grace period expires, so many handlers unwind at once and race `observabilityMgr.Close()`. Concurrency matters: at ~40 connections it did not reproduce; at 300 it reproduced on every attempt (4/4 runs, 8–48 panics per run).

I originally hit this on a production instance of the same version proxying a WebSocket endpoint with `accessLog.bufferingSize: 100`, where it fired once during a restart.

### What did you see instead?

```
{"level":"error","middlewareName":"traefik-internal-recovery","middlewareType":"Recovery",
 "caller":"github.com/traefik/traefik/v3/pkg/middlewares/recovery/recovery.go:49",
 "message":"Recovered from panic in HTTP handler [172.27.0.1:48968 - /ws]: send on closed channel"}
```

```
Stack: goroutine 1986 [running]:
github.com/traefik/traefik/v3/pkg/middlewares/recovery.recoverFunc(...)
	github.com/traefik/traefik/v3/pkg/middlewares/recovery/recovery.go:52 +0x225
panic({0x64961e0?, 0x756c220?})
	runtime/panic.go:860 +0x13a
github.com/traefik/traefik/v3/pkg/middlewares/accesslog.(*Handler).ServeHTTP.func1()
	github.com/traefik/traefik/v3/pkg/middlewares/accesslog/logger.go:290 +0x238
github.com/traefik/traefik/v3/pkg/middlewares/accesslog.(*Handler).ServeHTTP(...)
	github.com/traefik/traefik/v3/pkg/middlewares/accesslog/logger.go:301 +0x18d7
github.com/traefik/traefik/v3/pkg/server/middleware.(*ObservabilityMgr).BuildEPChain.(*Handler).AliceConstructor.func4.1(...)
	github.com/traefik/traefik/v3/pkg/middlewares/accesslog/logger.go:181 +0x3a
net/http.HandlerFunc.ServeHTTP(...)
	net/http/server.go:2286 +0x29
github.com/traefik/traefik/v3/pkg/middlewares/observability.(*entryPointTracing).ServeHTTP(...)
	github.com/traefik/traefik/v3/pkg/middlewares/observability/entrypoint.go:52 +0x237
github.com/traefik/traefik/v3/pkg/middlewares/capture.Wrap.func1(...)
	github.com/traefik/traefik/v3/pkg/middlewares/capture/capture.go:50 +0x162
net/http.HandlerFunc.ServeHTTP(...)
	net/http/server.go:2286 +0x29
github.com/traefik/traefik/v3/pkg/server/middleware.(*ObservabilityMgr).observabilityContextHandler.WithObservabilityHandler.func1(...)
	github.com/traefik/traefik/v3/pkg/middlewares/observability/observability.go:30 +0x9b
github.com/traefik/traefik/v3/pkg/muxer/http.(*Muxer).ServeHTTP(...)
	github.com/traefik/traefik/v3/pkg/muxer/http/mux.go:60 +0x151
github.com/traefik/traefik/v3/pkg/middlewares/recovery.(*recovery).ServeHTTP(...)
	github.com/traefik/traefik/v3/pkg/middlewares/recovery/recovery.go:36 +0x147
github.com/traefik/traefik/v3/pkg/middlewares.(*HTTPHandlerSwitcher).ServeHTTP(...)
	github.com/traefik/traefik/v3/pkg/middlewares/handler_switcher.go:23 +0x91
github.com/traefik/traefik/v3/pkg/middlewares/requestdecorator.(*RequestDecorator).ServeHTTP(...)
	github.com/traefik/traefik/v3/pkg/middlewares/requestdecorator/request_decorator.go:47 +0x349
```

Isolating the two variables — 300 connections, everything else identical:

| `accessLog.bufferingSize` | in-flight connection at SIGTERM | panics |
| --- | --- | --- |
| `100` | WebSocket (hijacked) | **48** |
| unset (`0`) | WebSocket (hijacked) | 0 |
| `100` | plain HTTP request, never answered (not hijacked) | 0 |

Both conditions are required: `bufferingSize > 0` **and** a hijacked connection. Plain in-flight requests are unaffected, because `Shutdown` does wait for those. In all three runs the shutdown took the full 10s grace period, so the two control runs were equivalent tests.

No plugins were loaded in this reproduction, and no plugin frames appear anywhere in the stacks.

**Secondary observation.** `recoverFunc` logs the panic message and the stack as two separate writes:

```go
logger.Error().Msgf("Recovered from panic in HTTP handler [%s - %s]: %+v", ...)   // recovery.go:49
const size = 64 << 10
buf := make([]byte, size)
buf = buf[:runtime.Stack(buf, false)]
logger.Error().Msgf("Stack: %s", buf)                                            // recovery.go:53
```

Because this happens as the process is exiting, the second write is frequently lost — **22 of the 48 panics above had no `Stack:` line**. On the production instance the single panic lost its stack entirely, which is what made it undiagnosable there. Happy to split this out into its own issue if you would prefer.

```console
$ docker run traefik:v3.7.10 version
Version:      3.7.10
Codename:     langres
Go version:   go1.26.5
Built:        2026-07-31T12:42:19Z
OS/Arch:      linux/amd64
```

Image digest `sha256:9c3b91d5fb7770853ca5c1124a23c34bf2d9b47ffaebeab2614cbaf410dcb2ac`.

Docker Compose on linux/amd64, file provider only, no plugins, no TLS, default `graceTimeOut` (10s).

Static configuration:

```yaml
entryPoints:
  web:
    address: ":80"

providers:
  file:
    filename: /etc/traefik/dynamic.yml

log:
  level: DEBUG
  format: json

accessLog:
  filePath: /tmp/access.log
  format: json
  bufferingSize: 100      # trigger; omitting this (default 0) makes the panic disappear

global:
  checkNewVersion: false
  sendAnonymousUsage: false
```

Dynamic configuration:

```yaml
http:
  routers:
    catchall:
      rule: "PathPrefix(`/`)"
      entryPoints: [web]
      service: backend

  services:
    backend:
      loadBalancer:
        servers:
          - url: "http://backend:8080"
```

`docker-compose.yml`:

```yaml
services:
  traefik:
    image: traefik:v3.7.10
    command: ["--configFile=/etc/traefik/traefik.yml"]
    ports:
      - "9080:80"
    volumes:
      - ./traefik.yml:/etc/traefik/traefik.yml:ro
      - ./dynamic.yml:/etc/traefik/dynamic.yml:ro
    depends_on:
      - backend

  backend:
    image: node:24-alpine
    command: ["node", "/app/ws-server.mjs"]
    volumes:
      - ./ws-server.mjs:/app/ws-server.mjs:ro
```

Backend (`ws-server.mjs`) — dependency-free; `/ws` completes the handshake and holds the socket open, `/hang` is an in-flight request that is never answered (used for the third row of the table):

```js
import http from "node:http";
import crypto from "node:crypto";

const WS_GUID = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";

const server = http.createServer((req, res) => {
  if (req.url.startsWith("/hang")) return;      // in-flight, NOT hijacked
  res.writeHead(200, { "content-type": "text/plain" });
  res.end("ok\n");
});

server.on("upgrade", (req, socket) => {
  const accept = crypto.createHash("sha1")
    .update(req.headers["sec-websocket-key"] + WS_GUID).digest("base64");
  socket.write("HTTP/1.1 101 Switching Protocols\r\nUpgrade: websocket\r\n" +
               `Connection: Upgrade\r\nSec-WebSocket-Accept: ${accept}\r\n\r\n`);
});

server.listen(8080);
```

```json
{"level":"info","time":"2026-08-15T09:25:32Z","caller":"github.com/traefik/traefik/v3/pkg/server/server.go:51","message":"I have to go..."}
{"level":"info","time":"2026-08-15T09:25:32Z","caller":"github.com/traefik/traefik/v3/pkg/server/server.go:52","message":"Stopping server gracefully"}
{"level":"debug","entryPointName":"web","time":"2026-08-15T09:25:32Z","caller":"github.com/traefik/traefik/v3/pkg/server/server_entrypoint_tcp.go:306","message":"Waiting 10s seconds before killing connections"}
{"level":"info","time":"2026-08-15T09:25:42Z","caller":"github.com/traefik/traefik/v3/cmd/traefik/traefik.go:170","message":"Shutting down"}
{"level":"error","middlewareName":"traefik-internal-recovery","middlewareType":"Recovery","time":"2026-08-15T09:25:42Z","caller":"github.com/traefik/traefik/v3/pkg/middlewares/recovery/recovery.go:49","message":"Recovered from panic in HTTP handler [172.27.0.1:48968 - /ws]: send on closed channel"}
```

Note that every panic is logged *after* `Shutting down`.

### Suggested fix

Guard the producer against the closed channel rather than closing a channel that live handlers may still write to — for example by tracking handler lifecycle and skipping the send once `Close()` has begun (an `RWMutex` around send/close, or a `done` channel selected on alongside the send).

Separately, emitting the panic message and its stack as a single log write would stop the trace being lost when a panic happens during shutdown.

waterWang:
### Problem

When `accessLog.bufferingSize > 0`, a hijacked connection (WebSocket) whose handler
unwinds after `Handler.Close()` panics with `send on closed channel`.

`Handler.Close()` closes the same channel that the deferred `ServeHTTP` code sends on,
and nothing orders the two. Because `http.Server.Shutdown` does not close nor wait for
hijacked connections, a proxied WebSocket handler can still be live when `Close()` is
called; when its connection is finally torn down, the deferred send runs against an
already-closed channel.

### Fix

1. Add a `done` channel to the `Handler` struct.
2. In `Close()`, close `h.done` **before** `h.logHandlerChan` — the consumer goroutine
   still drains `logHandlerChan` via `for range`, and the `done` channel signals the
   producer to stop.
3. In the deferred `ServeHTTP` send, wrap the channel send in a `select` with
   `<-h.done` — if the handler is closing, the log entry is silently dropped instead
   of panicking.

### Race-free ordering

- `Close()` closes `h.done` → producers see `<-h.done` ready in the select
- Then closes `h.logHandlerChan` → consumer goroutine exits via `for range`
- `h.wg.Wait()` waits for the consumer to drain remaining entries

Fixes #13693

### DIFF
--- a/pkg/middlewares/accesslog/logger.go
+++ b/pkg/middlewares/accesslog/logger.go
@@ -153,6 +153,7 @@ func NewHandler(ctx context.Context, config *otypes.AccessLog, hooks ...logrus.H
 		logger:         logger,
 		file:           file,
 		logHandlerChan: logHandlerChan,
+		done:           make(chan struct{}),
 	}
 
 	if config.Filters != nil {
--- a/pkg/middlewares/accesslog/logger.go
+++ b/pkg/middlewares/accesslog/logger.go
@@ -72,6 +72,7 @@ type Handler struct {
 	httpCodeRanges types.HTTPCodeRanges
 	logHandlerChan chan handlerParams
 	wg             sync.WaitGroup
+	done           chan struct{}
 }
 
 // NewHandler creates a new Handler.
@@ -291,9 +292,12 @@ func (h *Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next http
 		}
 
 		if h.config.BufferingSize > 0 {
-			h.logHandlerChan <- handlerParams{
+			select {
+			case h.logHandlerChan <- handlerParams{
 				ctx:          req.Context(),
 				logDataTable: logDataTable,
+			}:
+			case <-h.done:
 			}
 			return
 		}
@@ -306,6 +310,7 @@ func (h *Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next http
 
 // Close closes the Logger (i.e. the file, drain logHandlerChan, etc).
 func (h *Handler) Close() error {
+	close(h.done)
 	close(h.logHandlerChan)
 	h.wg.Wait()
 	return h.file.Close()


pkg/middlewares/accesslog/logger.go
```diff
@@ -291,9 +293,12 @@ func (h *Handler) ServeHTTP(rw http.ResponseWriter, req *http.Request, next http
 		}
 
 		if h.config.BufferingSize > 0 {
-			h.logHandlerChan <- handlerParams{
+			select {
+			case h.logHandlerChan <- handlerParams{
 				ctx:          req.Context(),
 				logDataTable: logDataTable,
+			}:
+			case <-h.done:
```
juliens, line 301: Here we will drop every accesslogs after the Close() call.

I think we should wait for all ServeHTTP call to be sure that there is no more accesslog waiting to be push.

juliens (changes requested):

PR #13708, opened 2026-08-18
Fix redirect www host with a non-numeric port

mmatur:
### What does this PR do?

Fixes the `from-to-www-redirect` router generated by the ingress-nginx provider so that it redirects requests whose `Host` header carries a non-numeric or empty port (for example `www.example.com:x` or `www.example.com:`).

The `RedirectRegex` pattern only matched an authority with a numeric port or no port at all, so a request carrying a malformed port did not match and fell through to the next handler instead of being redirected. The pattern now matches the whole authority and captures only a numeric port.

### Motivation

The redirect router is expected to redirect every request it receives for the `www` host. Requests with a malformed port in the `Host` header were not redirected, diverging from the expected behavior.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
+++ b/pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
@@ -5051,7 +5051,7 @@ func TestLoadIngresses(t *testing.T) {
 							EntryPoints: []string{"http"},
 							Rule:        `Host("host.localhost")`,
 							RuleSyntax:  "default",
-							Service:     "noop@internal",
+							Service:     unavailableServiceName,
 							Observability: &dynamic.RouterObservabilityConfig{
 								Metadata: &dynamic.ObservabilityMetadata{
 									Ingress: &dynamic.KubernetesIngressMetadata{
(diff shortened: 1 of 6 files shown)


LBF38 (dismissed):
LGTM

pkg/provider/kubernetes/ingress-nginx/kubernetes_test.go
sdelicata, line 1: Tests actually rely on no-op next handlers.
WDYT about adding a test with a next handler that flip a boolean and ensures that the next handler is never called?

sdelicata, line 1: In `pkg/middlewares/redirect/redirect_regex_test.go`, it could be a good idea to add a dedicated test with a malformed authorities such as:
- www.example.com:
- www.example.com:x
- www.example.com:8080;x
WDYT?

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -564,7 +564,7 @@ func applyFromToWwwRedirect(loc *location, routerKey string, rt *dynamic.Router,
 	mwName := routerKey + "-from-to-www-redirect"
 	conf.HTTP.Middlewares[mwName] = &dynamic.Middleware{
 		RedirectRegex: &dynamic.RedirectRegex{
-			Regex:       `(https?)://[^/:]+(:[0-9]+)?/(.*)`,
+			Regex:       `(https?)://[^/]*?(:[0-9]+)?/(.*)`,
```
sdelicata: IMO there's still an issue since we still point to the protected backend.
We should point to the `unavailableServiceName` which is already used in this exact "must not proxy" situation. See line `229` for ex.

sdelicata (changes requested):

sdelicata (approved):
LGTM

gndz07 (approved):
LGTM

PR #13717, opened 2026-08-18
Apply read timeout, idle timeout, and max header bytes for HTTP/3

gndz07:
### What does this PR do?

Apply `readTimeout` as a per-request read deadline on the request's QUIC stream via `http.ResponseController.SetReadDeadline` and apply `IdleTimeout`/`MaxHeaderBytes`, which are set directly, using two fields that already exist on `http3.Server` but were never populated.

### Motivation

`entryPoints.<name>.transport.respondingTimeouts.readTimeout`, `.idleTimeout`, and `entryPoints.<name>.http.maxHeaderBytes` were never applied to HTTP/3; `newHTTP3Server` built `http3.Server` without referencing any of them.
`readTimeout` defaults to 60s and is documented as bounding "the maximum duration for reading the entire request, including the body". An operator enabling `http3` on an existing entrypoint would have it silently ignored and not protected with a timeout, leaving a request body read unbounded.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/server/server_entrypoint_tcp_http3.go
+++ b/pkg/server/server_entrypoint_tcp_http3.go
@@ -48,11 +48,13 @@ func newHTTP3Server(ctx context.Context, configuration *static.EntryPoint, https
 		},
 	}
 
+	handler := httpsServer.Server.(*http.Server).Handler
+	if readTimeout := time.Duration(configuration.Transport.RespondingTimeouts.ReadTimeout); readTimeout != 0 {
+		handler = withHTTP3ReadTimeout(ctx, handler, readTimeout)
+	}
+
 	h3.Server = &http3.Server{
-		Addr:      configuration.GetAddress(),
-		Port:      configuration.HTTP3.AdvertisedPort,
-		Handler:   httpsServer.Server.(*http.Server).Handler,
-		TLSConfig: &tls.Config{GetConfigForClient: h3.getTLSConfigForClient},
+		Handler:        handler,
 		QUICConfig: &quic.Config{
 			Allow0RTT: false,
 		},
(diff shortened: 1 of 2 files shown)


pkg/server/server_entrypoint_tcp_http3.go
```diff
@@ -48,11 +49,18 @@ func newHTTP3Server(ctx context.Context, configuration *static.EntryPoint, https
 		},
 	}
 
+	handler := httpsServer.Server.(*http.Server).Handler
+	if readTimeout := time.Duration(configuration.Transport.RespondingTimeouts.ReadTimeout); readTimeout != 0 {
+		handler = withHTTP3ReadTimeout(ctx, handler, readTimeout)
```
rtribotte:
```suggestion
		handler = withReadTimeout(ctx, handler, readTimeout)
```

pkg/server/server_entrypoint_tcp_http3.go
```diff
@@ -79,6 +87,18 @@ func newHTTP3Server(ctx context.Context, configuration *static.EntryPoint, https
 	return h3, nil
 }
 
+func withHTTP3ReadTimeout(ctx context.Context, next http.Handler, timeout time.Duration) http.Handler {
+	return http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		if req.ContentLength != 0 {
```
rtribotte, line 96:
```suggestion
		if req.ContentLength != 0 || req.Body != nil || req.Body != http.NoBody {
```

rtribotte (approved):
LGTM

mmatur (approved):

PR #13720, opened 2026-08-19
Add an entry point option to handle request headers with aliasing names

rtribotte:
### What does this PR do?

Adds a new `aliasHeadersStrategy` entry point option, and deprecates `underscoreHeadersStrategy`.

### Motivation

Go canonicalizes header names on dashes only, whereas the backends deriving variable names from the header names (CGI, WSGI, PHP, NGINX, ...) replace every character that is neither a letter nor a digit with an underscore.

The underscore is not the only character building such an alias: `!`, `#`, `$`, `%`, `&`, `'`, `*`, `+`, `.`, `^`, `` ` ``, `|` and `~` do as well, and all of them reach the handlers in both HTTP/1 and HTTP/2. The `underscoreHeadersStrategy` option only covers one of them.

### More

- [x] Added/updated tests

### Additional Notes

The `underscoreHeadersStrategy` behavior is left unchanged.

Assisted-by: Claude Opus


gndz07 (approved):
LGTM

mmatur (approved):

pkg/server/server_entrypoint_tcp.go
```diff
@@ -209,6 +213,26 @@ func NewTCPEntryPoint(ctx context.Context, configuration *static.EntryPoint, hos
...
+			"whereas the aliasHeadersStrategy option handles every header name aliasing another one.")
+	}
+
+	if configuration.HTTP.AliasHeadersStrategy == "" &&
+		(configuration.HTTP.UnderscoreHeadersStrategy == "" || configuration.HTTP.UnderscoreHeadersStrategy == static.UnderscoreHeadersStrategyKeep) {
+		log.FromContext(ctx).Warn("aliasHeadersStrategy is not configured: the request headers whose name aliases another header name " +
+			"(e.g. X_Auth_User or X.Auth.User for X-Auth-User) are forwarded as is. The backends deriving variable names from the header " +
+			"names (CGI, WSGI, PHP, NGINX, ...) read them as the header they alias, which allows a client to spoof the headers Traefik manages. " +
+			"Please set it to delete or reject on the entry points fronting such backends.")
+	}
```
sdelicata, line 233: Maybe I'm reading this wrong, but this skips the warning for people who set `underscoreHeadersStrategy` to delete or reject. Those are the ones who read the v2.11.51 advisory and hardened, and they're still fully exposed to the dot form and the 12 other alias chars. They get the deprecation notice, which reads more like housekeeping than "you're still vulnerable".

What about gating only on `AliasHeadersStrategy == ""`, or spelling out the exposure in the deprecation message?

sdelicata (changes requested):

pkg/server/server_entrypoint_tcp.go
sdelicata: My bad, the deprecated warning is already clear enough.

sdelicata (approved):

PR #13721, opened 2026-08-19
Fix TLS option name collision across namespaces in the ingress-nginx provider

gndz07:
### What does this PR do?

Length-suffixes both the namespace and the secret name when deriving the `auth-tls-secret` TLS option key, so two Ingresses in different namespaces can no longer collide on the same cached TLS option.

### Motivation

To avoid possible collision between two distinct objects.

### More

- [x] Added/updated tests


rtribotte (approved):
LGTM

nandorKollar (approved):

PR #13723, opened 2026-08-19
Update redirections block reference in basic.md

Larzenegger:
### What does this PR do?

Add "http" to info block so that it contains the entire path to the attribute

### Motivation

Avoid confusion when cross-checking with https://doc.traefik.io/traefik/setup/kubernetes/

### More

- [ ] Added/updated tests


rtribotte (approved):
Thanks!

PR #13725, opened 2026-08-19
Ignore negated matchers when parsing rule domains

rtribotte:
### What does this PR do?

Excludes negated matchers from the domains deduced from a router rule.

### Motivation

A router with a rule such as ``Host(`a.com`) && !Host(`b.com`)`` was reported as serving `b.com`.
As the deduced domains are used to bind a router TLS configuration to an SNI, such a router was hijacking the TLS options of the router actually serving `b.com`.

Both routers being reported as serving `b.com`, they also conflict on it: the TLS options of the router actually serving it fall back to the `default` ones, or, with `core.strictTLSOptions` enabled, both routers are marked in error and are not built.

### More

- [x] Added/updated tests

### Additional Notes

Assisted-by: Claude Opus

### DIFF
--- a/pkg/rules/parser.go
+++ b/pkg/rules/parser.go
@@ -99,12 +99,19 @@ func notFunc(elem TreeBuilder) TreeBuilder {
 	}
 }
 
-// ParseMatchers returns the subset of matchers in the Tree matching the given matchers.
+// ParseMatchers returns the subset of matchers in the Tree matching the given matchers,
+// skipping the negated ones. A negated matcher excludes its values from what the rule matches,
+// they are therefore not part of the values the rule can match.
 func (tree *Tree) ParseMatchers(matchers []string) []string {
 	switch tree.Matcher {
 	case and, or:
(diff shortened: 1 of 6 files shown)


mmatur (approved):

pkg/rules/parser.go
```diff
@@ -99,12 +99,19 @@ func notFunc(elem TreeBuilder) TreeBuilder {
 	}
 }
 
-// ParseMatchers returns the subset of matchers in the Tree matching the given matchers.
+// ParseMatchers returns the subset of matchers in the Tree matching the given matchers,
+// skipping the negated ones. A negated matcher excludes its values from what the rule matches,
+// they are therefore not part of the values the rule can match.
 func (tree *Tree) ParseMatchers(matchers []string) []string {
```
kevinpollet: IMHO, this method should be renamed, as it is not obvious that it will not parse non-matchers. Plus, I would add a comment on ParseDomains and ParseHostSNI methods to explicit that not matcher will not be parsed in the given rule.

kevinpollet (changes requested):

kevinpollet (approved):

PR #13726, opened 2026-08-19
Support wildcard origins in cors-allow-origin for the ingress-nginx provider

### LINKED ISSUE #13724: kubernetesIngressNGINX: wildcard origins in cors-allow-origin (https://*.example.com) silently never match
### Welcome!

### What did you do?

We are migrating from ingress-nginx to Traefik using the `kubernetesIngressNGINX` provider.

Some of our Ingresses use a single-level wildcard origin in the CORS annotation, which is an
officially documented ingress-nginx feature
(https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#enable-cors):

```yaml
kind: Ingress
metadata:
  name: whoami
  annotations:
    nginx.ingress.kubernetes.io/enable-cors: "true"
    nginx.ingress.kubernetes.io/cors-allow-origin: "https://*.example.com"
spec:
  ingressClassName: nginx
  rules:
    - host: api.example.com
      http:
        paths:
          - path: /
            pathType: Prefix
            backend:
              service:
                name: whoami
                port:
                  number: 80
```

Then sent a CORS preflight request from a subdomain origin:

```
curl -s -D- -o /dev/null \
  -H "Origin: https://app.example.com" \
  -H "Access-Control-Request-Method: GET" \
  -X OPTIONS http://api.example.com/
```

### What did you see instead?

Neither the preflight response nor the actual response contains any `Access-Control-Allow-Origin`
header, so the browser blocks the request. There is no error or warning in the logs — the
wildcard entry silently never matches any origin.

Root cause: `buildCORS()` puts all annotation values into `AccessControlAllowOriginList`, which
only supports exact string comparison or a literal `*`:

- provider: https://github.com/traefik/traefik/blob/v3.7.10/pkg/provider/kubernetes/ingress-nginx/middleware.go#L247
- matcher (`matchOrigin`): https://github.com/traefik/traefik/blob/v3.7.10/pkg/middlewares/headers/header.go#L195

The headers middleware already supports `AccessControlAllowOriginListRegex`, and `matchOrigin`
already checks both lists — but the provider never populates the regex list.

**Proposed fix**: route single-level wildcard entries to the regex matcher, mirroring
ingress-nginx semantics where `*` matches exactly one DNS label. Exact values keep the current
behavior, and no configuration surface changes:

```diff
 import (
  "context"
  "fmt"
  "net/http"
+ "regexp"
  "slices"
  "strconv"
  "strings"
  ...
 )

+// wildcardOriginRegex converts a single-level wildcard origin (e.g. https://*.example.com)
+// into a regular expression, mirroring ingress-nginx semantics where '*' matches
+// exactly one DNS label.
+func wildcardOriginRegex(origin string) string {
+ quoted := regexp.QuoteMeta(origin)
+ return "^(?i)" + strings.Replace(quoted, `\*`, `[A-Za-z0-9-]+`, 1) + "$"
+}
+
 func (p *Provider) buildCORS(loc *location) {
  if !ptr.Deref(loc.Config.EnableCORS, false) {
    return
  }

+ // Split configured origins into exact values and single-level wildcards.
+ var originList, originRegexList []string
+ for _, origin := range ptr.Deref(loc.Config.CORSAllowOrigin, []string{"*"}) {
+   if origin != "*" && strings.Contains(origin, "*") {
+     originRegexList = append(originRegexList, wildcardOriginRegex(origin))
+     continue
+   }
+   originList = append(originList, origin)
+ }
+
  loc.CORS = &dynamic.Headers{
    AccessControlAllowCredentials: ptr.Deref(loc.Config.EnableCORSAllowCredentials, true),
    AccessControlExposeHeaders:    ptr.Deref(loc.Config.CORSExposeHeaders, []string{}),
    AccessControlAllowHeaders:     ptr.Deref(loc.Config.CORSAllowHeaders, []string{ /* defaults */ }),
    AccessControlAllowMethods:     ptr.Deref(loc.Config.CORSAllowMethods, []string{ /* defaults */ }),
-   AccessControlAllowOriginList:  ptr.Deref(loc.Config.CORSAllowOrigin, []string{"*"}),
+   AccessControlAllowOriginList:      originList,
+   AccessControlAllowOriginListRegex: originRegexList,
    AccessControlMaxAge:           new(int64(ptr.Deref(loc.Config.CORSMaxAge, 1728000))),
  }
 }
```

I'm happy to submit a PR with this change and tests if the approach is acceptable.
Alternatively, if wildcard support is out of scope, it would help migrating users if the
annotations support table documented this limitation, since the failure is silent.

v3.7.10 (helm chart 41.2.0). The same code is present on master.

Kubernetes (EKS), Traefik installed via the official helm chart, running alongside
ingress-nginx during migration.

Relevant static configuration:

```yaml
providers:
  kubernetesIngressNGINX:
    enabled: true
    ingressClass: nginx
    ingressClassByName: true
  kubernetesIngress:
    enabled: false
```

_No response_

kimsehyoung:
### What does this PR do?

Routes single-level wildcard origins (e.g. `https://*.example.com`) in the `nginx.ingress.kubernetes.io/cors-allow-origin` annotation to the headers middleware's `AccessControlAllowOriginListRegex`, converting the wildcard into a regex where `*` matches exactly one DNS label. Exact origins and the literal `*` keep going to `AccessControlAllowOriginList`, so existing behavior is unchanged and no configuration surface is added. The matcher already checks both lists, so only the provider wiring changes.

### Motivation

Fixes #13724

While migrating from ingress-nginx, wildcard origins — an officially documented ingress-nginx feature — silently never match with the `kubernetesIngressNGINX` provider: no error, no log, the browser just blocks the request. The annotations support table lists `cors-allow-origin` as supported with no caveat, so this is easy to hit without noticing.

### More

- [x] Added/updated tests

### Additional Notes

- The generated regex is case-insensitive (`(?i)`) to mirror ingress-nginx's `~*` origin matching. Happy to drop it if case-sensitive matching is preferred for consistency with the exact-match list.
- `*` matches exactly one DNS label (`[A-Za-z0-9-]+`), matching the documented ingress-nginx limitation (multi-level subdomains like `a.b.example.com` are not matched).
 - Origins with a wildcard outside the documented protocol://*.domain[:port] position are skipped rather than converted, matching ingress-nginx's own annotation validation.
- The default (no annotation → `["*"]`) path is unchanged.

*note: This implementation intentionally follows ingress-nginx’s permissive corsOriginRegex validation rules to preserve migration compatibility.*


pkg/provider/kubernetes/ingress-nginx/middleware.go
```diff
@@ -244,18 +245,38 @@ func (p *Provider) buildIPAllowList(loc *location) {
...
 func (p *Provider) buildCORS(loc *location) {
 	if !ptr.Deref(loc.Config.EnableCORS, false) {
 		return
 	}
 
+	// ingress-nginx supports single-level wildcard origins, which the exact-match origin
+	// list cannot express; route them to the regex matcher to preserve behavior on migration.
+	var originList, originRegexList []string
+	for _, origin := range ptr.Deref(loc.Config.CORSAllowOrigin, []string{"*"}) {
+		if origin != "*" && strings.Contains(origin, "*") {
```
amazon7737, line 276: Could you restrict wildcard conversion to the documented `protocol://*.domain[:port]` form and add regression cases for unsupported wildcard placement? 

`parseIngressConfig` only splits the raw annotation, so `strings.Contains(origin, "*")` currently turns unsupported values into active allow-origin regexes, unlike ingress-nginx which skips them.

amazon7737 (changes requested):
Thanks for working on this!
I left a comment suggesting that wildcard handling be limited to the syntax supported by ingress-nginx and that tests be added for unsupported formats.

pkg/provider/kubernetes/ingress-nginx/middleware.go
kimsehyoung: @amazon7737
Good catch, thanks — fixed in f6fea83.

Now using ingress-nginx's own corsOriginRegex pattern with the wildcard group made mandatory, so only protocol://*.domain[:port] is converted and anything else is dropped as ingress-nginx does. The check only applies to values containing a *, to avoid newly dropping exact origins that work today.

Added middleware_test.go with a table test over each supported and unsupported form, and extended the existing CORS fixture with one unsupported value so the skip is covered end to end.

docs/content/reference/routing-configuration/kubernetes/ingress-nginx.md
```diff
@@ -342,7 +342,7 @@ The following annotations are organized by category for easier navigation.
...
 | <a id="opt-nginx-ingress-kubernetes-iocors-allow-methods" href="#opt-nginx-ingress-kubernetes-iocors-allow-methods" title="#opt-nginx-ingress-kubernetes-iocors-allow-methods">`nginx.ingress.kubernetes.io/cors-allow-methods`</a> |                                                                                            |
-| <a id="opt-nginx-ingress-kubernetes-iocors-allow-origin" href="#opt-nginx-ingress-kubernetes-iocors-allow-origin" title="#opt-nginx-ingress-kubernetes-iocors-allow-origin">`nginx.ingress.kubernetes.io/cors-allow-origin`</a> |                                                                                            |
+| <a id="opt-nginx-ingress-kubernetes-iocors-allow-origin" href="#opt-nginx-ingress-kubernetes-iocors-allow-origin" title="#opt-nginx-ingress-kubernetes-iocors-allow-origin">`nginx.ingress.kubernetes.io/cors-allow-origin`</a> | Single-level wildcard origins are supported in the `protocol://*.domain[:port]` form, where `*` matches exactly one DNS label. Wildcards in any other position are ignored, as with NGINX. |
```
amazon7737: Nit: Could we say “as with ingress-nginx” here? The annotation validation and skipping behavior comes from the ingress-nginx controller rather than NGINX itself.

kimsehyoung: You're right — the validation and skipping happen in the controller, not NGINX. Fixed in b1dc2a65.

amazon7737 (approved):
LGTM 👍

LBF38 (approved):
LGTM
Thanks for your contribution @kimsehyoung !

PR #13738, opened 2026-08-20
Isolate the Kubernetes Gateway API listeners

rtribotte:
### What does this PR do?

Isolates the Kubernetes Gateway API listeners: a request is only served by the routes of the listener that is the most specific match for its host name.

### Motivation

The routes of a listener were served for host names another listener of the Gateway is a more specific match for, which the listener isolation forbids.

Adds the support for the `GatewayHTTPListenerIsolation` and `GatewayHTTPSListenerDetectMisdirectedRequests` features.

### More

- [x] Added/updated tests

### Additional Notes

The host names a listener router rule excludes are matched with `HostRegexp` to keep them out of the domains deduced from the rule. They can be rewritten as `Host` matchers once #13725 is merged.

Assisted-by: Claude Opus

### DIFF
--- a/docs/content/migrate/v3.md
+++ b/docs/content/migrate/v3.md
@@ -34,6 +34,21 @@ For the experimental channel:
 kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.6.1/experimental-install.yaml
 ```
 
+#### Listener Isolation
+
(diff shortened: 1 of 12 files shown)


gndz07 (approved):
LGTM

juliens (approved):

PR #13740, opened 2026-08-21
Document the security threat model and settled security decisions

emilevauge:
### What does this PR do?

Adds two things to the security documentation:

- a **threat model** on `submitting-security-issues.md`: what Traefik trusts, what crosses the boundary, and what is handled as a bug without an advisory;
- a **Security Decisions** page recording the positions we already apply, grouped by the surface a report touches, each with the neighbouring variant we do treat as a vulnerability and the advisories that prove it.

Also adds three submission requirements: check the finding against the threat model, one finding per report, disclose AI tooling use.

### Motivation

Our most frequent report class is not a new vulnerability, it is a position the team has already settled, rediscovered and resubmitted. That cost is paid again for every resubmission, and reporters cannot check their finding against a bar that is not published.

Publishing the bar answers those reports with a reference instead of a fresh analysis, and lets a reporter self-check before submitting. The pages are written from our own triage history: every position corresponds to decisions already applied, and the 52 referenced advisories are all published.

The threat model states the boundary; the decisions page applies it. Keeping them separate is deliberate, so the same claim is not maintained in two places.

### More

- [ ] Added/updated tests

### Additional Notes

The threat model carries an effective date of 1 September 2026 and applies to reports submitted from then on. It describes existing practice rather than new rules, but retroactive application is what turns a policy change into a dispute, so the date is explicit.

Two paragraphs are worth reviewing closely, both stating a line publicly for the first time: the Ingress-NGINX provider section, where upstream compatibility is kept even when the upstream outcome is insecure but does not extend to authentication failing open; and the statement that the internal configuration keyspace is not a tenant isolation boundary on a shared instance.

### DIFF
--- a/docs/content/contributing/security-decisions.md
+++ b/docs/content/contributing/security-decisions.md
@@ -76,9 +76,6 @@ that does not repeat a strip is **not automatically** an incomplete fix.
 
 - A path that **reintroduces or reconstructs a trusted value after entrypoint sanitisation**, so the
   entrypoint's decision no longer holds.
-- **Header name handling** that lets an untrusted client reach a header the entrypoint believes it
-  sanitised: underscore variants, alias forms, non-canonical casing, or any normalization difference
-  between Traefik and the backend.
 - A middleware that makes a **pre-authentication decision** on a header the operator declared
   untrusted.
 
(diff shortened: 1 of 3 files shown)


amazon7737 (approved):
LGTM

docs/content/contributing/security-decisions.md
```diff
@@ -0,0 +1,287 @@
...
+was expected to protect.
+
+**Our position.** This is our most productive surface and we treat it seriously. A normalization or
+encoding differential that reaches the backend **with stock entrypoint defaults** is a vulnerability,
+including when it only manifests as a desynchronisation between the decoded and raw forms of the path.
+
+**Where the line is.** We decline variants whose escape depends on the operator having widened handling,
+or on one backend's own re-interpretation of a character the standard permits:
+
+- The escape works only with `allowEncodedSlash` or encoded-character handling left permissive. That
```
rtribotte:
```suggestion
- The escape works only with encoded-character handling left permissive. That
```

docs/content/contributing/security-decisions.md
```diff
@@ -0,0 +1,315 @@
...
```
rtribotte:
```suggestion
- The escape works only with encoded-character handling left permissive. That
```

rtribotte (approved):
LGTM

gndz07 (approved):
LGTM

mmatur (approved):

PR #13741, opened 2026-08-21
Build a collision-free item key in the Consul Catalog and Nomad providers

rtribotte:
### What does this PR do?

Makes the per-instance configuration key of the Consul Catalog and Nomad providers collision-free, so that two service instances with distinct identities no longer share a key, which discarded the configuration of all but one of them.

### Motivation

`(node=n1, name=web, id=a-b)` and `(node=n1, name=web-a, id=b)` currently produce the same key.

### More

- [x] Added/updated tests

### Additional Notes

Assisted-by: Claude Opus 5


gndz07 (approved):
LGTM

sdelicata (approved):
LGTM

PR #13746, opened 2026-08-21
Build the configuration copy once per change

jspdown:
### What does this PR do?

Builds the configuration copy in `receiveConfigurations` once per change, instead of inside the two `select` send cases.

### Motivation

A `select` evaluates the right hand side of its send cases when it enters the statement, even for a case that cannot run. Both send cases call `DeepCopy` inline, so most of those copies are built and dropped: 3 map copies per message where 1 is enough, and 2 more for an unchanged configuration that is skipped.

Here is a chart showing the measured impact before and after:

deepcopy-cost

The tested scenarios were:
  - `providers-1/2/5/10`: one changed message at a time, drained before the next, with 1 to 10 providers in the map.
  - `size-10/50/300`: the same, varying routers per provider instead of provider count.
  - `transformers-1/3`: the same, with configuration transformers registered.
  - `burst`: messages arrive faster than a  consumer takes them.
  - `resync`: provider re-sends the configuration it already sent, which the dedup skips.

### More

- [ ] Added/updated tests

### DIFF
--- a/pkg/server/configurationwatcher.go
+++ b/pkg/server/configurationwatcher.go
@@ -141,14 +141,15 @@ func (c *ConfigurationWatcher) receiveConfigurations(ctx context.Context) {
 
 				newConfigurations[configMsg.ProviderName] = configMsg.Configuration.DeepCopy()
 
-				transformedConfigurations := newConfigurations
+				// DeepCopy is necessary because newConfigurations gets modified later by the consumer of c.newConfigs.
+				transformedConfigurations := newConfigurations.DeepCopy()
 				for _, transform := range c.configurationTransformers {
+					// Each transformer gets its own copy because a transformer could keep a reference to the one it received.
 					transformedConfigurations = transform(logger.WithContext(ctx), transformedConfigurations.DeepCopy())
 				}
 
 				output = c.newConfigs
-				// DeepCopy is necessary because newConfigurations gets modified later by the consumer of c.newConfigs.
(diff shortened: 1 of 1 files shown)


pkg/server/configurationwatcher.go
```diff
@@ -140,16 +141,19 @@ func (c *ConfigurationWatcher) receiveConfigurations(ctx context.Context) {
 
 				newConfigurations[configMsg.ProviderName] = configMsg.Configuration.DeepCopy()
 
-				transformedConfigurations = newConfigurations
+				// DeepCopy is necessary because newConfigurations gets modified later by the consumer of c.newConfigs.
+				transformedConfigurations := newConfigurations.DeepCopy()
 				for _, transform := range c.configurationTransformers {
+					// Each transformer gets its own copy because a transformer could keep a reference to the one it received.
 					transformedConfigurations = transform(logger.WithContext(ctx), transformedConfigurations.DeepCopy())
```
rtribotte: @jspdown maybe there is a better form but I would guarantee that pending receives a fresh copy, and not the last copy passed to the last transform

```suggestion
					transformedConfigurations = transform(logger.WithContext(ctx), transformedConfigurations)
					transformedConfigurations = transformedConfigurations.DeepCopy()
```

jspdown: Good catch 👍

rtribotte (approved):
LGTM

mmatur (approved):

PR #13749, opened 2026-08-21
Reject negative weights in TCP and UDP weighted services

rtribotte:
### What does this PR do?

Rejects negative weights in TCP and UDP weighted services, and reports a child service build error on the parent weighted service.

### Motivation

A negative weight made the TCP and UDP weighted round-robin selector loop endlessly while holding its lock, consuming a CPU core and blocking every subsequent connection to that service. Correcting the configuration did not stop an already running loop, so recovery required restarting Traefik.

While fixing it, a child service failing to build left the parent weighted service without an error nor a status, unlike its HTTP counterpart.

### More

- [x] Added/updated tests

### Additional Notes

Assisted-by: Claude Opus 5


gndz07 (approved):
LGTM

mmatur (approved):
LGTM

PR #13753, opened 2026-08-24
Tell scanning agents to read the security policy and decisions pages

emilevauge:
### What does this PR do?

Adds a "Reporting a security finding" section to `AGENTS.md`, pointing at the two pages an agent should read before filing anything: the [security policy](https://doc.traefik.io/traefik/contributing/submitting-security-issues/) for the threat model and submission requirements, and [security decisions](https://doc.traefik.io/traefik/contributing/security-decisions/) for the positions already settled.

It then states what to do with the result: a finding matching a settled position should not be filed; one matching the in-scope variant named in that same entry should be, saying how it differs; a non-GA-only finding belongs in an issue rather than an advisory.

Also updates the maintenance branch reference from `v3.6` to `v3.7`.

### Motivation

Our most frequent report class is a position the team has already settled, rediscovered and resubmitted, and an increasing share arrives from automated tooling. The two pages exist to answer those with a reference instead of a fresh analysis, but only if the tool reads them first. `AGENTS.md` is where agents look, and `CLAUDE.md` already imports it, so this is the one place that reaches them before they file.

The branch reference is a separate correction: v3.6 security support ended 16 August 2026.

### More

- [ ] Added/updated tests

### Additional Notes

Also restates the two submission requirements most often missed, one finding per report and disclosing AI tooling use, since an agent that reads only this file would otherwise not see them.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -92,7 +92,7 @@ User-facing features need matching documentation updates under `docs/content/`. 
 
 ## Contributing etiquette
 
-- **Target the right branch** (the [PR template](./.github/PULL_REQUEST_TEMPLATE.md) is authoritative): enhancements go to `master`; bug fixes and documentation updates go to the current maintenance branches (`v3.6` for v3, `v2.11` for v2, security-fixes only). Forward-ports from the maintenance branches up to `master` are handled by maintainers.
+- **Target the right branch** (the [PR template](./.github/PULL_REQUEST_TEMPLATE.md) is authoritative): enhancements go to `master`; bug fixes and documentation updates go to the current maintenance branches (`v3.7` for v3, `v2.11` for v2, security-fixes only). Forward-ports from the maintenance branches up to `master` are handled by maintainers.
 - Keep pull requests small and focused; one logical change per PR.
 - For anything beyond a bug fix, open an issue first and wait for a maintainer to confirm the direction before investing significant work.
 - Follow the full guide in [`docs/content/contributing/submitting-pull-requests.md`](./docs/content/contributing/submitting-pull-requests.md).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,11 +110,14 @@ Traefik welcomes AI-assisted contributions, provided a few simple rules are foll
 If you are scanning this repository for vulnerabilities, read these two pages **before** filing anything,
 and cite them in whatever you file:
 
-1. [Security policy](https://doc.traefik.io/traefik/contributing/submitting-security-issues/) : the threat
-   model, what Traefik trusts, what crosses the security boundary, and the submission requirements.
-2. [Security decisions](https://doc.traefik.io/traefik/contributing/security-decisions/) : the positions
-   already settled, grouped by the surface a report touches, each naming the neighbouring variant that
-   *is* a vulnerability.
+1. [`docs/content/contributing/submitting-security-issues.md`](./docs/content/contributing/submitting-security-issues.md)
+   : the threat model, what Traefik trusts, what crosses the security boundary, and the submission
+   requirements. Published at
+   [doc.traefik.io](https://doc.traefik.io/traefik/contributing/submitting-security-issues/).
+2. [`docs/content/contributing/security-decisions.md`](./docs/content/contributing/security-decisions.md)
+   : the positions already settled, grouped by the surface a report touches, each naming the
+   neighbouring variant that *is* a vulnerability. Published at
+   [doc.traefik.io](https://doc.traefik.io/traefik/contributing/security-decisions/).
 
 Then apply the outcome:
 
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -105,6 +105,35 @@ Traefik welcomes AI-assisted contributions, provided a few simple rules are foll
...
+Then apply the outcome:
+
+- **The finding matches a settled position on the decisions page : do not file it.** It will be closed
+  with a link to that entry. The most common report we receive is a position already decided,
+  rediscovered.
+- **It matches the in-scope variant named in that same entry : file it**, and say which entry and how
+  your finding differs from the declined shape.
+- **The surface is not covered : check it against the threat model.** A report whose precondition is
+  configuration-write access, operator or cluster-admin privilege, or a non-GA version, is not a
+  vulnerability here.
```
sdelicata: Maybe I'm misreading, but this bullet and the next one disagree about non-GA findings: here it's "not a vulnerability here" (so don't file), and the next one says open a normal issue. An agent applying this list top-down stops on the first match and drops the report.

The decisions page is pretty explicit that we do want those (`Reporting during a release candidate is also the most useful moment to reach us`), so I think non-GA is a routing question, not a scope one, and it doesn't belong next to config-write access and operator privilege.

WDYT about just removing `or a non-GA version` here and letting the next bullet own it?

AGENTS.md
```diff
@@ -105,6 +105,35 @@ Traefik welcomes AI-assisted contributions, provided a few simple rules are foll
...
+- **Non-GA only** (release candidate, beta, development branch) : open a normal issue, not an advisory.
+
+Two requirements that are easy to miss, both on the policy page: **one finding per report**, and
```
sdelicata: Since the point is that an agent reading only this file still gets it right, maybe the working PoC + repro steps requirement should be in here too? It's the first `must` on the policy page and the first indicator in the AI-generated section, and it's the one these reports actually miss.

Also `Undisclosed unvalidated` reads like disclosing buys you something: the policy closes unvalidated AI-generated reports immediately whether they're disclosed or not.

```suggestion
Three requirements that are easy to miss, all on the policy page: a **working proof of concept with
reproduction steps**, **one finding per report**, and **disclose your use of AI tooling**. Unvalidated
AI-generated reports are closed immediately.
```

AGENTS.md
```diff
@@ -105,6 +105,35 @@ Traefik welcomes AI-assisted contributions, provided a few simple rules are foll
...
+## Reporting a security finding
+
+If you are scanning this repository for vulnerabilities, read these two pages **before** filing anything,
+and cite them in whatever you file:
+
+1. [Security policy](https://doc.traefik.io/traefik/contributing/submitting-security-issues/) : the threat
+   model, what Traefik trusts, what crosses the security boundary, and the submission requirements.
+2. [Security decisions](https://doc.traefik.io/traefik/contributing/security-decisions/) : the positions
+   already settled, grouped by the surface a report touches, each naming the neighbouring variant that
+   *is* a vulnerability.
```
sdelicata: An agent reading AGENTS.md already has the checkout, so a relative path is one read away, while doc.traefik.io is a fetch it may not be allowed to make. That matters a bit here since reading them is a precondition for the whole section.

AGENTS.md
```diff
@@ -92,7 +92,7 @@ User-facing features need matching documentation updates under `docs/content/`.
 
 ## Contributing etiquette
 
-- **Target the right branch** (the [PR template](./.github/PULL_REQUEST_TEMPLATE.md) is authoritative): enhancements go to `master`; bug fixes and documentation updates go to the current maintenance branches (`v3.6` for v3, `v2.11` for v2, security-fixes only). Forward-ports from the maintenance branches up to `master` are handled by maintainers.
+- **Target the right branch** (the [PR template](./.github/PULL_REQUEST_TEMPLATE.md) is authoritative): enhancements go to `master`; bug fixes and documentation updates go to the current maintenance branches (`v3.7` for v3, `v2.11` for v2, security-fixes only). Forward-ports from the maintenance branches up to `master` are handled by maintainers.
```
sdelicata, line 95: Since this line says the template is authoritative, it's now disagreeing with it: `.github/PULL_REQUEST_TEMPLATE.md` still offers `v3.6` for both docs and bugs. Same reasoning as this change, v3.6 security support ended Aug 16 (releases.md).

sdelicata (changes requested):
Adding this new section is definitively a good idea. Thanks for that!

AGENTS.md
emilevauge: You are not misreading, the two bullets did contradict each other, and top-down first-match is exactly how an agent would read them. Removed `or a non-GA version` and left the next bullet to own the routing.

AGENTS.md
emilevauge: Agreed on both. Taken your suggestion as-is: the PoC requirement is in, and `Undisclosed unvalidated` is gone, since disclosure does not change the outcome for an unvalidated report.

AGENTS.md
emilevauge: Good point, and it applies to the whole section since reading them is the precondition. Both entries now lead with the in-repo path and keep the published URL after it, so an agent that cannot fetch still has a file to open.

AGENTS.md
emilevauge: Right, this change made AGENTS.md contradict the file it calls authoritative. Dropped both `v3.6` lines from the template in a separate commit.

sdelicata (approved):
LGTM

jspdown (approved):
LGTM 👍

PR #13764, opened 2026-08-25
Merge v2.11 into v3.7

mmatur:
### What does this PR do?

Merge current v2.11 into v3.7

### Motivation

Be up to date

### More

- [ ] Added/updated tests


go.mod
```diff
@@ -378,7 +378,7 @@ require (
 	github.com/youmark/pkcs8 v0.0.0-20240726163527-a2c0da244d78 // indirect
 	github.com/yusufpapurcu/wmi v1.2.4 // indirect
 	go.etcd.io/etcd/api/v3 v3.6.8 // indirect
-	go.etcd.io/etcd/client/pkg/v3 v3.6.8 // indirect
+	go.etcd.io/etcd/client/pkg/v3 v3.6.14 // indirect
 	go.etcd.io/etcd/client/v3 v3.6.8 // indirect
```
sdelicata: Shouldn't we align the 3 etcd client modules as it was previously the case?

gndz07 (approved):
LGTM

pkg/server/service/tcp/service.go
sdelicata: We have two more `return nil, err` that v2.11 didn't have. We should probably add an error to the conf here as well:
```suggestion
                conf.AddError(err, true)
				return nil, fmt.Errorf("child service %v of %v not a healthcheck.StatusUpdater (%T)", service.Name, serviceName, handler)
```

sdelicata: And here:
```suggestion
                conf.AddError(err, true)
				return nil, fmt.Errorf("cannot register %v as updater for %v: %w", service.Name, serviceName, err)
```

pkg/tcp/wrr_load_balancer.go
sdelicata, line 172: I'm afraid we have a bug here since we have a `status` map in v3 that didn't exist in v2.11.
If every UP server has a weight of 0 and some DOWN server has a positive weight, we will stay stuck forever.
I suggest to simply skip DOWN servers:
```suggestion
        // Skipped, as they are never selected: a maximum taken from a down server would
        // set a threshold no selectable server can reach, making nextServer spin forever.
        if _, ok := b.status[s.name]; !ok {
			continue
		}
		
		if s.weight > maximum {
```
WDYT?

pkg/tcp/wrr_load_balancer.go
```diff
@@ -172,6 +173,11 @@ func (b *WRRLoadBalancer) maxWeight() int {
 func (b *WRRLoadBalancer) weightGcd() int {
 	divisor := -1
 	for _, s := range b.servers {
+		// Skipped, as they are never selected and would make the divisor negative.
```
sdelicata, line 188: Same here:
```suggestion
        // Skipped, as they are never selected: the divisor has to cover the same servers
        // as the maximum, or the weight steps no longer match the selectable servers.
		if _, ok := b.status[s.name]; !ok {
			continue
		}
		// Skipped, as they are never selected and would make the divisor negative.
```

sdelicata (changes requested):

sdelicata (approved):
Thanks for the changes.
LGTM

PR #13767, opened 2026-08-25
Set access log entry level and time before formatting the OTLP body

### LINKED ISSUE #13766: Access log records shipped over OTLP have level=panic and a zero time in their body
### Welcome!

- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've searched similar issues on the Traefik community forum and didn't find any.

### What did you do?

Configured the access log to be shipped over OTLP, with `dualOutput` so the same records also go to a file:

```yaml
accessLog:
  filePath: "/logs/access.log"
  format: json
  otlp:
    grpc:
      endpoint: collector:4317
      insecure: true
  dualOutput: true

experimental:
  otlpLogs: true
```

Sent traffic through a router, then compared the records written to the file with the ones received by the OTLP collector.

### What did you see instead?

Every access log record shipped over OTLP has a JSON body stamped with `"level":"panic"` and `"time":"0001-01-01T00:00:00Z"`, while the exact same record written to the file has the correct `"level":"info"` and a real timestamp.

Received over OTLP (body, abridged):

```json
{"DownstreamStatus":200,"RequestHost":"whoami.localhost","RouterName":"whoami@docker","ServiceName":"whoami@docker","level":"panic","msg":"","time":"0001-01-01T00:00:00Z"}
```

Same record in the file:

```json
{"DownstreamStatus":200,"RequestHost":"whoami.localhost","RouterName":"whoami@docker","ServiceName":"whoami@docker","level":"info","time":"2026-08-25T08:28:52.141714591Z"}
```

The OTLP `severityText` field itself is correct (`info`), so only the log body is affected. But the body is what most backends index, display and alert on, so in practice every access log record looks like a panic. It also makes the body's `time` field useless for anything reading it.

The OTLP body to carry the same `level` and `time` as the file output.

### Root cause

In `pkg/middlewares/accesslog/logger.go`, the OTLP body is produced by calling the formatter on the entry before the entry is actually logged:

```go
entry := h.logger.WithContext(ctx).WithFields(fields)

var message string
if h.config.OTLP != nil {
    mBytes, err := h.logger.Formatter.Format(entry)
    ...
    message = string(mBytes)
}

entry.Println(message)
```

`logrus.Entry.Level` and `logrus.Entry.Time` are only populated by logrus when the log method is called (`Entry.log` sets `newEntry.Level = level`, and `newEntry.Time = time.Now()` if it is zero). At `Format` time both are still zero values: `logrus.PanicLevel` is `0`, hence `"level":"panic"`, and the zero `time.Time` renders as `0001-01-01T00:00:00Z`.

v3.7.8, and reproduced on the current `v3.7` branch.

See above. Reproduced with the OTLP/gRPC exporter against an `otel/opentelemetry-collector-contrib` receiver, and with a unit test on the `v3.7` branch.

Not applicable, the issue is in the access log body itself.

emilevauge:
### What does this PR do?

Sets the access log entry level and time before the formatter computes the OTLP log body.

`logrus.Entry.Level` and `logrus.Entry.Time` are only populated by logrus when the log method is called: `Entry.log` sets `newEntry.Level = level`, and `newEntry.Time = time.Now()` only if it is zero. The OTLP body is produced by calling `Formatter.Format(entry)` before `entry.Println`, so at format time both fields are still zero values. `logrus.PanicLevel` is `0`, which is why every record shipped over OTLP carries `"level":"panic"` and `"time":"0001-01-01T00:00:00Z"` in its body.

Because `Entry.log` only overrides `Time` when it is zero, logrus reuses the timestamp set here, so the OTLP body and the regular output now carry the same level and the same time rather than two independently computed values.

Fixes #13766

### Motivation

Any Traefik instance shipping access logs over OTLP has every access log body stamped as a panic, with an unusable timestamp inside the body. The OTLP `severityText` field itself is correct (`info`), so only the body is affected, but the body is what most backends index, display and alert on.

Found while comparing the file output and the OTLP output of the same records against an `otel/opentelemetry-collector-contrib` receiver.

### More

- [X] Added/updated tests

### Additional Notes

The new assertions are on the JSON-format case of `TestOTelAccessLogWithBodyAndDualOutput`. They fail on the current `v3.7` branch and pass with the fix.

Targeting `v3.7` as a bug fix, per the branch guidance in the PR template.

### DIFF
--- a/pkg/middlewares/accesslog/logger.go
+++ b/pkg/middlewares/accesslog/logger.go
@@ -393,7 +393,11 @@ func (h *Handler) logTheRoundTrip(ctx context.Context, logDataTable *LogData) {
 		// Setting them here avoids formatting a body with the zero values, and makes logrus reuse
 		// this timestamp, so the OTLP body and the regular output carry the same level and time.
 		entry.Level = logrus.InfoLevel
+
 		entry.Time = time.Now()
+		if t, ok := core[StartUTC].(time.Time); ok {
+			entry.Time = t
+		}
 
 		mBytes, err := h.logger.Formatter.Format(entry)
 		if err != nil {
--- a/pkg/middlewares/accesslog/logger.go
+++ b/pkg/middlewares/accesslog/logger.go
@@ -389,6 +389,12 @@ func (h *Handler) logTheRoundTrip(ctx context.Context, logDataTable *LogData) {
 	if h.config.OTLP != nil {
 		// If the logger is configured to use OpenTelemetry,
 		// we compute the log body with the formatter.
+		// The formatter reads the entry level and time, which logrus only sets when the log method is called.
+		// Setting them here avoids formatting a body with the zero values, and makes logrus reuse
+		// this timestamp, so the OTLP body and the regular output carry the same level and time.
+		entry.Level = logrus.InfoLevel
+		entry.Time = time.Now()
+
 		mBytes, err := h.logger.Formatter.Format(entry)
 		if err != nil {
 			message = fmt.Sprintf("Failed to format access log entry: %v", err)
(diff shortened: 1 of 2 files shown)


sdelicata (approved):
LGTM

kevinpollet (approved):

PR #13689, opened 2026-08-13
Add safe naming option to avoid collisions for Kubernetes CRD provider

gndz07:
### What does this PR do?

Introduces a new `safeNaming (*bool)` option on the Kubernetes CRD provider, defaulting to false/unset, which preserves the legacy naming behavior. When enabled, routers, middlewares, and services are named by joining their identity components (namespace, name, route index, etc.) with a `_` separator instead of a rule-derived hash, preventing collisions.

### Motivation

The naming system to prevent collisions that were introduced in #13656 and #13668 was applied unconditionally and might be breaking for some users. With this new option, user can choose to apply the collision-free naming system or to stay with the current one (the default).

### More

- [x] Added/updated tests
- [x] Added/updated documentation

PR #13768, opened 2026-08-25
Document the safeNaming impact on cross-provider references

### LINKED ISSUE #13765: Wrong docs for kubernetescrd when `safeName=true`
### Welcome!

### What did you do?

I've enabled `kubernetescrd.safeName` and now the middlewares have the new name on runtime.

`<namespace>-<name>@kubernetescrd` now becomes `<namespace>_<name>@kubernetescrd` when referencing it from `traefik.ingress.kubernetes.io/router.middlewares`, so the middlewares and routers are no longer working.

### What did you see instead?

kubernetes ingress no longer works because of it and the docs don't mention it anywhere

- https://doc.traefik.io/traefik/reference/install-configuration/providers/overview/#provider-namespace

Version:      3.7.11
Codename:     langres
Go version:   go1.26.6
Built:        2026-08-19T06:38:56Z
OS/Arch:      linux/amd64

```yaml
providers:
  kubernetesIngress:
    allowEmptyServices: true
    publishedService:
      enabled: false
  kubernetesCRD:
    allowEmptyServices: true
    safeNaming: true # this is newly added with traefik v3.7.11
```

manually changing the reference from `<namespace>-<name>@kubernetescrd` to `<namespace>_<name>@kubernetescrd` works.

_No response_

lazerg:
### What does this PR do?

Documents that `safeNaming` also changes how Kubernetes CRD objects are referenced from other providers, not only how they appear in the dashboard, logs and metrics.

The `_` separator note already existed on the two ServersTransport pages, so this adds the same note to the remaining pages that state the reference format: the provider namespace section, the Ingress annotations reference, the entry point middleware format table, the TLSOption annotation, the service middlewares annotation, and the `safeNaming` option itself. The v3.7.11 migration guide gets the same warning, since that is the page read when deciding to enable the option.

### Motivation

With `safeNaming` enabled, a middleware in namespace `default` named `stripprefix` is generated as `default_stripprefix`, so an Ingress annotation still pointing at `default-stripprefix@kubernetescrd` resolves to nothing and the router is not created. The docs only mentioned the naming change under Observability, which reads as cosmetic.

Fixes #13765

### More

- [ ] Added/updated tests

### Additional Notes

The wording is the sentence introduced by #13689 on the ServersTransport pages, reused as is on the pages it did not cover.


viceice (commented):
❤️

PR #13779, opened 2026-08-26
Fix displayed number on details pages

gndz07:
### What does this PR do?

Wrap numbers as a text inside the details card, so that it will inherit the text's styling.

### Motivation

Numbers were not taken into account as a string, so the they were displayed unstyled.

Before:
image

After:
image

### More

- [ ] Added/updated tests


rtribotte (approved):
LGTM

kevinpollet (approved):

PR #13780, opened 2026-08-26
Downgrade default TLS resources namespace mismatch log to warning

### LINKED ISSUE #13778: Remove error logs for ignoring default TLS store when defining defaultTLSResourcesNamespace
### Welcome!

### What did you do?

Configured Traefik Helm Chart with
``` yaml
providers:
  kubernetesCRD:
    defaultTLSResourcesNamespace: traefik-internal
```

### What did you see instead?

Traefik pod is generating a lot of error logs because it cannot use a default TLS store in another namespace.
This is expected because we configure a specific namespace to retrieve the TLSStore from.

From our perspective these error logs are obsolete as we expect Traefik to not use TLSStores from other namespaces.

Image

3.7.11

``` yaml
providers:
  kubernetesCRD:
    defaultTLSResourcesNamespace: traefik-internal
```

_No response_

lazerg:
### What does this PR do?

Logs the ignored default TLSStore and TLSOption resources at warning level instead of error, when `defaultTLSResourcesNamespace` restricts them to a single namespace.

### Motivation

Setting `defaultTLSResourcesNamespace` is an explicit choice to honor the default TLS resources from one namespace only, so finding them elsewhere is the expected result of that option rather than a failure. The message also runs on every watched Kubernetes event, not only when the resulting configuration changes, which is what floods the logs. Warning is the level `addToConfig` already uses to report a resource being dropped, and it keeps the message reachable for anyone running below the default ERROR level.

Fixes #13778

### More

- [ ] Added/updated tests

### Additional Notes

The existing `buildTLSStores` and `buildTLSOptions` tests assert the resulting dynamic configuration only, so the level change leaves them unaffected.

Tell me if you would rather have debug here. I kept warning because a `default` TLSStore showing up outside the reserved namespace can also be a tenant trying to replace the cluster-wide TLS policy, which is the case the option was added to close.

### DIFF
--- a/pkg/provider/kubernetes/crd/kubernetes.go
+++ b/pkg/provider/kubernetes/crd/kubernetes.go
@@ -1354,7 +1354,7 @@ func (p *Provider) buildTLSOptions(ctx context.Context, client Client) map[strin
 		// When a namespace is explicitly configured, the default TLS options can only be defined in this namespace.
 		if tlsOptionsCRD.Name == tls.DefaultTLSConfigName &&
 			p.DefaultTLSResourcesNamespace != "" && tlsOptionsCRD.Namespace != p.DefaultTLSResourcesNamespace {
-			logger.Error().Msgf("Ignoring default TLS options: they can only be defined in the %q namespace", p.DefaultTLSResourcesNamespace)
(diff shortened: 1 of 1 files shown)


mmatur (approved):

rtribotte (approved):
Thanks

kevinpollet (approved):
Thanks 👍

PR #13796, opened 2026-08-28
Deny request with an opaque request target

sdelicata:
### What does this PR do?

Rejects with a `400 Bad Request` any request whose target is rootless, that is a scheme followed by something that does not start with a slash (`http:example.com/admin`). The new `denyOpaque` handler sits on the entry point chain, before the path handling ones. As a second barrier, the outgoing proxy request now clears `URL.Opaque`.

### Motivation

Go stores such a target in `URL.Opaque` and leaves `URL.Path`, `URL.RawPath` and `URL.Host` empty, so `denyFragment`, `normalizePath`, `sanitizePath` and the routing never see it. But `URL.RequestURI()` gives `Opaque` precedence over the path, so the target comes back when forwarding.

None of the four request target forms of RFC 9112 section 3.2 can populate `Opaque`, so the rejection is not configurable.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/migration/v2.md
+++ b/docs/content/migration/v2.md
@@ -1177,10 +1177,5 @@ A rootless target, for instance `http:example.com/admin`, carries no path,
 hence escaping the path handling and the routing rules while still designating a resource on the backend.
 This rejection is not configurable, as allowing such a target would reintroduce the bypass.
 
-Clients only emit this form when the application overrides the request target explicitly,
-for instance with `curl --request-target`.
-
 As the validation happens before the routing, the rejected requests are not reported in the access logs.
 They are only visible at the `DEBUG` log level.
-
-Please check out the [Request Path](../security/request-path.md) documentation for more details.
--- a/docs/content/migration/v2.md
+++ b/docs/content/migration/v2.md
@@ -1166,3 +1166,21 @@ The providers exposing a scheme instead of a server URL take it the same way,
 with the `traefik.http.services.<service_name>.loadbalancer.server.scheme=h2c` label,
 the `scheme: h2c` option of the Kubernetes CRD,
 or the `traefik.ingress.kubernetes.io/service.serversscheme: h2c` Kubernetes Ingress annotation.
+
+### Request Target Validation
+
+Starting with `v2.11.57`, a request whose target is rootless, that is a target made of a scheme followed by
+something that does not start with a slash, is rejected with a `400 Bad Request` response.
+Such a target is none of the four forms allowed by [rfc9112#section-3.2](https://datatracker.ietf.org/doc/html/rfc9112#section-3.2).
+
+A rootless target, for instance `http:example.com/admin`, carries no path,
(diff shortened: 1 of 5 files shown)


docs/content/migration/v2.md
```diff
@@ -1166,3 +1166,21 @@ The providers exposing a scheme instead of a server URL take it the same way,
...
+Starting with `v2.11.57`, a request whose target is rootless, that is a target made of a scheme followed by
+something that does not start with a slash, is rejected with a `400 Bad Request` response.
+Such a target is none of the four forms allowed by [rfc9112#section-3.2](https://datatracker.ietf.org/doc/html/rfc9112#section-3.2).
+
+A rootless target, for instance `http:example.com/admin`, carries no path,
+hence escaping the path handling and the routing rules while still designating a resource on the backend.
+This rejection is not configurable, as allowing such a target would reintroduce the bypass.
+
+Clients only emit this form when the application overrides the request target explicitly,
+for instance with `curl --request-target`.
```
kevinpollet: This should be removed

docs/content/migration/v2.md
```diff
@@ -1166,3 +1166,21 @@ The providers exposing a scheme instead of a server URL take it the same way,
...
+
+As the validation happens before the routing, the rejected requests are not reported in the access logs.
+They are only visible at the `DEBUG` log level.
+
+Please check out the [Request Path](../security/request-path.md) documentation for more details.
```
kevinpollet: This should be removed

docs/content/security/request-path.md
kevinpollet, line 1: Here I would add anything about request target validation.

pkg/server/service/proxy.go
```diff
@@ -73,6 +73,8 @@ func buildProxy(passHostHeader *bool, responseForwarding *dynamic.ResponseForwar
 
 			pr.Out.URL.Path = u.Path
 			pr.Out.URL.RawPath = u.RawPath
+			// URL.RequestURI gives Opaque precedence over the path, an opaque outgoing URL would discard the path set above.
+			pr.Out.URL.Opaque = ""
```
kevinpollet: I would move this just after or before removing the RequestURI.

pkg/server/server_entrypoint_tcp.go
```diff
@@ -625,6 +625,10 @@ func createHTTPServer(ctx context.Context, ln net.Listener, configuration *stati
 
 	handler = denyFragment(handler)
 
+	// An opaque URL has to be rejected before any handler deriving the path or the request URI from it,
+	// hence the wrapping has to be done after the path handling ones.
+	handler = denyOpaque(handler)
```
kevinpollet: Maybe this should be the first handler executed.

kevinpollet (changes requested):

mmatur (approved):

kevinpollet (approved):

PR #13798, opened 2026-08-28
Preserve ingress-nginx negative lookahead path exclusions

omkar619-dev:
### What does this PR do?

Translates ingress-nginx paths that exclude subpaths with a negative lookahead into a form Go's regexp can compile, so the router is created instead of silently dropped.

    /api/licensing/((?!_internal).*)  -> PathRegexp("(?i)^/api/licensing/(.*)") && !PathRegexp("(?i)^/api/licensing/(?:_internal)")

The assertion is zero-width *and* non-capturing, so capture numbering and spans are unchanged, which is what keeps `rewrite-target: /$1` working. All three sites that compile the path get the cleaned form: `buildRule`, `buildRewriteTarget`,
and `buildRedirect`.

The supported shape is a literal prefix, one negative lookahead opening the capture group, a literal alternation inside it, and no top-level alternation in the tail. Paths outside it are emitted verbatim exactly as today and now log a warning naming the path and the ingress.

### Motivation

Fixes https://github.com/traefik/traefik/issues/13606

### More

- [x] Added/updated tests

### Additional Notes

Priority. Traefik derives router priority from rule length, and this provider sets an explicit `Priority` only on default-backend routers, so the negated arm would have promoted every translated router above its neighbors on the same host. Each of the six path-carrying routers is now pinned to the priority its own rule would have scored before translation, which keeps the existing canary and the non-canary ordering.

That score comes from the verbatim rule, not the keep arm as I proposed in the issue thread. ingress-nginx sorts locations by path length, so the verbatim length is the closer match to the ordering a migrating operator already has.

The from-to-www redirect router no longer inherits `rt.Priority`, since its `ExtraRouterRule` is independent of the path rule.

*The code now enforces the requested shape. Please update the PR description to state the exact `X((?!A|B|C).*)` restriction and mention the unsupported absolute URL rewrite combination.*


pkg/provider/kubernetes/ingress-nginx/lookahead_test.go
```diff
@@ -0,0 +1,341 @@
...
+		{
+			desc:                "unsupported shape is left verbatim",
+			path:                `/a/((?!b).*)/((?!c).*)`,
+			expectedRule:        `Host("example.localhost") && PathRegexp("(?i)^/a/((?!b).*)/((?!c).*)")`,
+			expectedPreNegation: `Host("example.localhost") && PathRegexp("(?i)^/a/((?!b).*)/((?!c).*)")`,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.desc, func(t *testing.T) {
```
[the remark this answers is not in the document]
omkar619-dev: Done — added at both the top-level functions and the subtests, matching translator_test.go and middleware_test.go.

pkg/provider/kubernetes/ingress-nginx/middleware.go
```diff
@@ -271,6 +271,9 @@ func (p *Provider) buildRewriteTarget(loc *location) {
 	}
 
 	regex := loc.Path
+	if keep, _, ok := splitNegativeLookahead(loc.Path); ok {
```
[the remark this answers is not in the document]
omkar619-dev: ok, Good catch, that's a real gap i feel . Applied the keep form in buildRedirect too, inside the same loc.UseRegex guard the other two sites use, and added Test_buildRedirect_negativeLookahead — it asserts the resulting regex and then runs regexp.Compile on it, which is the assertion that fails on the old code with invalid or unsupported Perl syntax: (?!.

I did went back and enumerated the rest while I was there: three regex compile sites in total — buildRule, buildRewriteTarget, buildRedirect. The remaining loc.Path uses can't compile a lookahead — a log message, the rewrite == loc.Path identity comparison, the $location_path snippet variable, the literal Path() matcher for PathTypeExact, and buildPrefixRule.

pkg/provider/kubernetes/ingress-nginx/lookahead.go
```diff
@@ -0,0 +1,150 @@
...
+
+	if !literalPathPrefix.MatchString(prefix) {
+		return "", "", false
+	}
+
+	if !literalAlternation.MatchString(alternation) {
+		return "", "", false
+	}
+
+	keep = path[:start] + path[end+1:]
```
amazon7737, line 109: Maybe we are talking before issue, The prefix and the alternation are constrained, but the tail after the lookahead is not. A top-level in the tail lets a branch bypass the assertion in ingress-nginx, so keep `&&` `!` exclude under-matches. With /a/((?!b)c|b), nginx serves /a/b through the second branch, while `PathRegexp("(?i)^/a/(c|b)") && !PathRegexp("(?i)^/a/(?:b)")` refuses it. In the issue we agreed to start from a literal prefix, one lookahead, and a .`*` tail. Could you constrain the tail to the shapes the tests cover, or at least reject a depth-0 in it, and add that case to the rejection table?

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -644,17 +671,37 @@ func buildRule(host string, loc *location) string {
...
-				rules = append(rules, fmt.Sprintf("PathRegexp(%q)", "(?i)^"+loc.Path))
+				verbatim := fmt.Sprintf("PathRegexp(%q)", nginxRegexPrefix+loc.Path)
+				if keep, exclude, ok := splitNegativeLookahead(loc.Path); ok {
+					pathRules = []string{
+						fmt.Sprintf("PathRegexp(%q)", nginxRegexPrefix+keep),
+						fmt.Sprintf("!PathRegexp(%q)", nginxRegexPrefix+exclude),
+					}
+					prePathRules = []string{verbatim}
+				} else {
+					pathRules = []string{verbatim}
```
amazon7737, line 700: When the path contains `(?!` but falls outside the supported shape, it is still emitted verbatim and the only signal is the generic regexp parse error from the router manager. The issue asks for a clear log when a pattern cannot be translated safely, and you proposed a warning for that case in the thread. Could you log a warning here that names the path and says the lookahead shape is unsupported?

pkg/provider/kubernetes/ingress-nginx/translator.go
```diff
@@ -230,7 +231,19 @@ func (p *Provider) translate(ctx context.Context, mc *model) *dynamic.Configurat
...
 
-			rule := buildRule(srv.Hostname, loc)
+			rule, preNegationRule := buildRule(srv.Hostname, loc)
+
+			// Traefik derives router priority from rule length, so the negated arm that
+			// lookahead translation adds would silently promote a translated router over
+			// its neighbors on the same host. Pin each router to the priority its own
+			// rule would have scored before translation. Zero means unset, which leaves
+			// the muxer's own length-based fallback in place where nothing was translated.
+			priorityOf := func(r string) int {
```
amazon7737: In the issue you proposed pinning to the keep arm. This pins to the verbatim rule, which is longer by the lookahead text. ingress-nginx sorts locations by path length, so the verbatim length is the closer match and I think this is the right choice. Could you state that change in the PR description so the reasoning is on record?

amazon7737 (changes requested):
Thanks for the careful work on this. Two blocking items are inline. Could you add an ingress fixture under fixtures/ingresses/ with a TestLoadIngresses case, at least for a lookahead path combined with rewrite-target? That is where the pinned priority on the generated routers and the rewrite regex get asserted end to end, and the current unit tests do not go through translate. 🙂

pkg/provider/kubernetes/ingress-nginx/lookahead.go
omkar619-dev: You're right, and thanks for the concrete counter-example. That's a genuine under-match, and the grammar had drifted wider than what we actually agreed in the issue: the prefix and the alternation were constrained but the tail never was, so /a/((?!b)c|b) passed all three checks and produced a translation that refuses /a/b.

so, splitNegativeLookahead now rejects a | at the top level of the tail, via a small depth scan that skips escapes and character classes the same way matchingParen does. It refuses at depth ≤ 0 rather than exactly 0, so a top-level alternation after the capture group closes is caught too — conservative, and simpler to reason about than tracking the two cases apart. Your case is in the rejection table as R7 btw.

pkg/provider/kubernetes/ingress-nginx/translator.go
omkar619-dev: Added at the verbatim branch, guarded on the path actually containing (?! so ordinary regex paths stay quiet — that branch is the normal route for all of them. The message names both the path and the ingress, since an operator with a lot of ingresses needs to know which one to fix.

I used the bare log.Warn() rather than log.Ctx(ctx): buildRule has no context, and kubernetes.go already uses the bare form for the analogous "this annotation value is invalid and will be ignored" case. Happy to thread a context through if you'd prefer it scoped.

pkg/provider/kubernetes/ingress-nginx/translator.go
omkar619-dev: Done, and it's in the commit message as well as the description so it survives a squash. Thank you for the reasoning, too — that ingress-nginx sorts locations by path length, so the verbatim length is the closer match to the ordering a migrating operator already has, is a better justification than the one I had in the issue thread. I've written it up that way rather than as my own.

pkg/provider/kubernetes/ingress-nginx/lookahead.go
amazon7737: The tail is still not restricted to the agreed `.* form. /a/((?!b).*)?b$` is still accepted and changes the matching result. Please support only `X((?!A|B|C).*)` and add rejection tests for other tails and expressions after the outer capture.

pkg/provider/kubernetes/ingress-nginx/lookahead_test.go
```diff
@@ -0,0 +1,388 @@
...
+			path:                `/api/licensing/((?!_internal).*)`,
+			expectedRule:        `Host("example.localhost") && PathRegexp("(?i)^/api/licensing/(.*)") && !PathRegexp("(?i)^/api/licensing/(?:_internal)")`,
+			expectedPreNegation: `Host("example.localhost") && PathRegexp("(?i)^/api/licensing/((?!_internal).*)")`,
+		},
+		{
+			desc:                "several excluded prefixes stay in a single negated arm",
+			path:                `/api/example/v1/((?!invitations|session|self-registration).*)`,
+			expectedRule:        `Host("example.localhost") && PathRegexp("(?i)^/api/example/v1/(.*)") && !PathRegexp("(?i)^/api/example/v1/(?:invitations|session|self-registration)")`,
+			expectedPreNegation: `Host("example.localhost") && PathRegexp("(?i)^/api/example/v1/((?!invitations|session|self-registration).*)")`,
+		},
```
amazon7737, line 337: Translation still runs separately in three places. As [agreed in the issue](https://github.com/traefik/traefik/issues/13606#issuecomment-5398783175), please run it once in the builder before buildMiddlewares, preserve loc.Path, and have the router and middleware builders consume the stored result.

amazon7737:
> I rebased onto [bc4f6fc](https://github.com/traefik/traefik/commit/bc4f6fc2816a96ff566c55140dd239aa97dac16d). The two new transforms compose in sequence rather than competing: in buildRule the split now runs on regexPath, and in buildRewriteTarget on regex after makeTrailingGroupOptional has had its turn, so the widening is never discarded. One consequence worth naming — once a path is widened its prefix stops being a literal, so splitNegativeLookahead rejects it and the path falls through to verbatim plus the new warning. That's the outcome I'd want, since widening breaks the fixed-length-prefix property the exclusion arm anchors on, but tell me if you'd rather they composed differently.
> 
> Also: v3.7 has moved into these two functions a few times in the last day. Happy to keep rebasing, but if you'd rather I hold until the rewrite-target work settles, just say.

Please document and test that an absolute URL rewrite combined with a lookahead path remains unsupported.

amazon7737 (changes requested):

pkg/provider/kubernetes/ingress-nginx/lookahead.go
omkar619-dev: Narrowed to exactly that. literalTail at lookahead.go:40 is now ^\.\*\)$, checked at lookahead.go:115, so the only shape that translates is X((?!A|B|C).*) with the outer capture closing at the end of the path.

Your counter-example falls out of that directly — the tail there is .*)?b$, which the anchored pattern rejects, so /a/((?!b).*)?b$ now goes through to verbatim plus the warning rather than being translated.

I should be straight that my first pass at this was wider than you asked for: it also accepted .+ and [^/]* tails, which I'd written acceptance tests for earlier. Those are rejections now. The rejection tests are in Test_splitNegativeLookahead, grouped under R7 since a single rule covers them all — a character-class tail, a one-or-more tail, a trailing literal inside the capture, a nested group, a trailing literal after the outer capture, and your /a/((?!b).*)?b$.

Worth noting why one check covers both halves of what you asked: it matches everything from the end of the assertion to the end of the path rather than just the capture body, so a tail inside the group and an expression after it are rejected by the same anchor.

If .+ or [^/]* turn out to be worth supporting later, I'd rather add them back one at a time with their own semantics tests than loosen the pattern again — but happy either way.

pkg/provider/kubernetes/ingress-nginx/lookahead_test.go
omkar619-dev: It runs once now, in the builder at build.go:509 and build.go:619, right after loc.UseRegex is resolved and before buildMiddlewares, and stores the two halves as PathKeep/PathExclude on location (model.go:115). loc.Path is never rewritten, so it's still what an untranslatable path compiles from and what a translated router's priority is pinned to. The consumers just read the stored fields — translator.go:690 for the rule, middleware.go:314 for the rewrite target, middleware.go:167 for the redirect. splitNegativeLookahead has exactly one caller outside the tests now, at lookahead.go:188.

One thing I hadn't noticed until I did this: the three sites weren't only duplicating the call, they were feeding it different inputs. buildRule and buildRewriteTarget split the widened path while buildRedirect split the raw one, so they could disagree about whether a given path was translatable. I pulled the widening out into nginxRegexPath (translator.go:747) so that widen-then-split is structural rather than something each call site has to remember.

That does shift buildRedirect slightly, since it now consumes the widened split. As far as I can tell it can only differ when a location carries an absolute rewrite-target, a redirect annotation and a lookahead at once — and in that case the router rule is left verbatim and doesn't compile, so the router is dropped and the redirect regex was unreachable anyway. Flagging it in case you read that differently

omkar619-dev:
> > I rebased onto [bc4f6fc](https://github.com/traefik/traefik/commit/bc4f6fc2816a96ff566c55140dd239aa97dac16d). The two new transforms compose in sequence rather than competing: in buildRule the split now runs on regexPath, and in buildRewriteTarget on regex after makeTrailingGroupOptional has had its turn, so the widening is never discarded. One consequence worth naming — once a path is widened its prefix stops being a literal, so splitNegativeLookahead rejects it and the path falls through to verbatim plus the new warning. That's the outcome I'd want, since widening breaks the fixed-length-prefix property the exclusion arm anchors on, but tell me if you'd rather they composed differently.
> > Also: v3.7 has moved into these two functions a few times in the last day. Happy to keep rebasing, but if you'd rather I hold until the rewrite-target work settles, just say.
> 
> Please document and test that an absolute URL rewrite combined with a lookahead path remains unsupported.

All three are pushed btw.

On the last one — documented on resolveNegativeLookahead (lookahead.go:183) and tested in Test_resolveNegativeLookahead_absoluteRewriteTarget (lookahead_test.go:422).

The reason the two can't compose is that the widening runs first and removes exactly what the split needs. makeTrailingGroupOptional turns /api/licensing/((?!_internal).*) into /api/licensing(?:/((?!_internal).*))?, so the prefix ahead of the assertion becomes /api/licensing(?:/, which is no longer a literal — literalPathPrefix rejects it and the translation declines. The path is then left verbatim, Go's regexp can't compile it, the router is dropped and the warning fires. So it's unsupported by construction rather than by an explicit guard, and that's what the test pins: the exact widened string, both halves empty, and the rule and the rewrite regex both still carrying the assertion.

Happy to call this out on the ingress-nginx provider docs page too if you'd rather it were user-visible — there's no limitations section there at the moment, so I didn't want to add one without asking.

amazon7737 (approved):
LGTM

PR #13812, opened 2026-08-31
Dedicate a transport per HTTP/3 client connection

sdelicata:
### What does this PR do?

This PR dedicates a backend connection pool to each HTTP/3 client connection when a backend answers with an NTLM or Negotiate (Kerberos) challenge, as the TCP entry point already does.

The HTTP/3 entry point `ConnContext` now installs the holder that `KerberosRoundTripper` looks up to keep a client on its own backend connection.

### Motivation

NTLM and Negotiate bind the authentication to the connection instead of the request, so Traefik dedicates a connection pool to each client connection for such backends. HTTP/3 requests go through the HTTPS handler chain, hence through the same round tripper, but the HTTP/3 entry point never set up the holder that mechanism relies on, so those backends did not get a dedicated pool over HTTP/3.

### More

- [x] Added/updated tests

### DIFF
--- a/docs/content/migration/v2.md
+++ b/docs/content/migration/v2.md
@@ -1179,13 +1179,3 @@ This rejection is not configurable, as allowing such a target would reintroduce 
 
 As the validation happens before the routing, the rejected requests are not reported in the access logs.
 They are only visible at the `DEBUG` log level.
-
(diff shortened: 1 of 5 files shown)


mmatur (approved):
LGTM

kevinpollet (approved):

PR #13814, opened 2026-08-31
feat(acme): add certificate renewal tracking through ARI

### LINKED ISSUE #13808: Add support for certificate renewal tracking through ARI
### Welcome!

ACME Renewal Information (ARI) allows ACME clients to contact the CA about suggested renewal windows. This mainly benefits scenarios in which a mass revocation occurs at a certificate authority, as they can consequently reflect this in their ARI responses to trigger automated renewal on the client side. It also offers some additional benefits to clients, for example not having to manually implement handling logic for certificates with distinct lifetimes, but instead relying on the provided renewal window.

Moreover, the suggested renewal window could also be used in the Traefik dashboard to warn about 'expiring soon' certificates, rather than relying on fixed durations as the current implementation does. As such, it could help tackling issues like #12461. 

A suggestion for implementation was already posted in https://github.com/traefik/traefik/issues/12461#issuecomment-4700342346; this may serve as a starting point.

References:
- https://letsencrypt.org/2025/09/16/ari-rfc

amarkdotdev:
## Summary

- Query ACME Renewal Information when renewing certificates
- Schedule the next renewal check from ARI retry windows when available
- Fall back to the existing time-based renewal logic when ARI is disabled or unsupported

Implementation based on the earlier work of @LtHummus (see also https://github.com/LtHummus/traefik/commit/2ea78e7).

Fixes #13808

## Test plan

- [x] go test ./pkg/provider/acme/...


pkg/provider/acme/provider_acme_renewal.go
```diff
@@ -0,0 +1,274 @@
...
+
+	retryAfter := renewInterval
+	if info.RetryAfter > 0 {
+		retryAfter = info.RetryAfter
+	}
+
+	return true, certID, retryAfter, nil
+}
+
+func (p *Provider) renewCertificateWithARI(ctx context.Context, client *lego.Client, crt *certRenewalInfo) (*certificate.Resource, error) {
```
jnoordsij, line 115: Do we really need an entire new handler here? I do realize that ARI introduced the new `ReplacesCertID` field for better replacement, but I just wonder if `lego` is not taking care of this automatically? If not, I wonder why and if there's some way to make them do it, as to me it would make a lot of sense to just use the renew API everywhere...

pkg/provider/acme/provider_acme_renewal.go
```diff
@@ -0,0 +1,274 @@
...
+	if renewAt == nil {
+		retryAfter := renewInterval
+		if info.RetryAfter > 0 {
+			retryAfter = info.RetryAfter
+		}
+		return false, "", retryAfter, nil
+	}
+
+	if renewAt.After(time.Now()) {
+		return false, "", renewAt.Sub(time.Now()), nil
```
jnoordsij: Why are we not using `info.RetryAfter` here? Per https://github.com/go-acme/lego/blob/0671c9ba21ff91cfe6faa50cca0f3c26827cdbb0/acme/commons.go#L361-L365:
> 	// RetryAfter header indicating the polling interval that the ACME server recommends.
	// Conforming clients SHOULD query the renewalInfo URL again after the RetryAfter period has passed,
	// as the server may provide a different suggestedWindow.
	// https://www.rfc-editor.org/rfc/rfc9773.html#section-4.2

I think this is particularly important, as the suggested renew window may change before it has reached it's start; in particular when a mass revocation happens and early renewals are scheduled, which is one of the main benefits of using ARI

jnoordsij (changes requested):
This is very much not an extensive review yet, just an early glance. I already have a few questions. Also as pointed out on https://github.com/traefik/traefik/issues/13808#issuecomment-5484326344, the initial implementation already had tests and those are gone. I think having those is very much a must if reasonably possible.

amarkdotdev:
@jnoordsij thanks for the lego issue link, I will keep an eye on that.

On splitting the PR: fair point about the per-certificate renew duration change being separate from ARI. I am fine leaving them combined for now unless you would strongly prefer a split before review. The lifetime-based durations were needed so ARI and legacy paths share the same scheduling logic, but I can peel that into its own PR first if that makes review easier for you.

amarkdotdev:
@jnoordsij thanks for the lego issue link, I will keep an eye on that.

On splitting the PR: fair point about the per-certificate renew duration change being separate from ARI. I am fine leaving them combined for now unless you would strongly prefer a split before review. The lifetime-based durations were needed so ARI and legacy paths share the same scheduling logic, but I can peel that into its own PR first if that makes review easier for you.

jnoordsij:
> I've opened https://github.com/go-acme/lego/issues/3272 to see if there's interest in adding support for the renewing in lego itself, as that would save quite a bit of boilerplate and asymmetrical handling here.

Thanks to the quick upstream response, this should now be available in go-acme/lego (see https://github.com/go-acme/lego/pull/3273). If you use `github.com/go-acme/lego/v5 v5.4.2-0.20260915193632-d638164dc4f1` you should be able to try it out.

PR #13816, opened 2026-09-01
Prevent user enumeration through the basic auth singleflight key

sdelicata:
### What does this PR do?

Derive the BasicAuth singleflight key from the submitted credentials (user and password) instead of the submitted password and the stored secret, and move it to a dedicated `singleflightKey` function.

### Motivation

The deduplication key was built from the stored secret, so it depended on the server state and not only on what the client sent. Concurrent requests carrying different usernames could end up sharing the same call. Keying on the credentials keeps the deduplication on what it's meant for, collapsing identical concurrent checks.

### More

- [x] Added/updated tests


mmatur (approved):
LGTM

kevinpollet (approved):

PR #13825, opened 2026-09-03
Avoid deprecated table.maxn in the rate limiter Lua script

### LINKED ISSUE #13824: Support Dragonfly for ratelimit
### Welcome!

When attempting to use Dragonfly instead of redis, the following happens in traefik:v3.7.9
`ERR Could not insert/update bucket error="evaluating script: running script: ERR Error running script (call to 3d93177e9b9a18f241edddc8d7e7fa1c47ae4eb6): @user_script:16: user_script:16: attempt to call a nil value (field 'maxn')" entryPointName=websecure middlewareName=keycloak-keycloak-token-ratelimit@kubernetescrd middlewareType=RateLimiter routerName=keycloak-...@kubernetes`

This pairs with the log in dragonfly:v1.40.1
`15 script_mgr.cc:363] Error running script (call to 3d93177e9b9a18f241edddc8d7e7fa1c47ae4eb6): @user_script:16: user_script:16: attempt to call a nil value (field 'maxn')"`

When attempting the following
```
kind: Middleware
metadata:
  name: keycloak-token-ratelimit
  namespace: {{ template "keycloak.namespace" . }}
spec:
  rateLimit:
    average: 60
    period: 60s
    burst: 5
    sourceCriterion:
      requestHeaderName: CF-Connecting-IP
    redis:
      endpoints:
        - "df-traefik.traefik:6379"
      db: 1
```

https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/ratelimit/

lazerg:
### What does this PR do?

Replaces `table.maxn(rl_source)` with `#rl_source` in the rate limiter token bucket script.

### Motivation

`table.maxn` only exists in Lua 5.1. Redis still embeds that version, but Redis-compatible servers built on newer Lua do not, so the script fails there with `attempt to call a nil value (field 'maxn')` and every request gets rate limited. `hgetall` returns a plain array, so the length operator gives the same result on both.

Fixes #13824

### More

- [x] Added/updated tests

### Additional Notes

The mock Redis client in the unit tests now clears `table.maxn` before running the script, so the test suite exercises it the way a Lua 5.4 server would.

I don't have a Dragonfly instance to test against, so confirmation from the reporter would be worth having before this gets merged.

### DIFF
--- a/pkg/middlewares/ratelimiter/lua.go
+++ b/pkg/middlewares/ratelimiter/lua.go
@@ -33,7 +33,7 @@ local bucket = {
 
 local rl_source = redis.call('hgetall', key)
 
-if table.maxn(rl_source) == 4 then
+if #rl_source == 4 then
     -- Get bucket state from redis
     bucket.last = tonumber(rl_source[2])
     bucket.tokens = tonumber(rl_source[4])
--- a/pkg/middlewares/ratelimiter/rate_limiter_test.go
+++ b/pkg/middlewares/ratelimiter/rate_limiter_test.go
@@ -568,6 +568,9 @@ func (m *mockRedisClient) EvalSha(ctx context.Context, _ string, keys []string, 
 	state := lua.NewState()
 	defer state.Close()
 
+	// gopher-lua provides table.maxn, unlike the Lua 5.4 that Dragonfly and similar servers embed.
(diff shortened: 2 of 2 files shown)


rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #13838, opened 2026-09-04
Dedicate a sticky transport per ServersTransport

sdelicata:
### What does this PR do?

Scopes the Kerberos/NTLM sticky round tripper to the ServersTransport that created it, instead of keeping a single one per client connection.

Every ServersTransport shared one slot in the ConnContext, so the stickiness was scoped to the client connection and to nothing else. Once a backend answered `WWW-Authenticate: NTLM|Negotiate`, every later request on that connection went through that clone, including requests routed to a service relying on another ServersTransport, which was then not served with its own transport configuration.

The slot becomes a mutex guarded map keyed by the `KerberosRoundTripper` that created the clone. The map is lazily allocated, as most client connections never reach an NTLM backend.

The mutex also fixes a data race: the slot was read and written with no synchronization, while HTTP/2 and HTTP/3 client connections multiplex concurrent requests.

### Motivation

The transport configuration an operator sets on a ServersTransport should govern the backend leg of that service, and of that service only.

### More

- [x] Added/updated tests

### DIFF
--- a/pkg/server/service/roundtripper.go
+++ b/pkg/server/service/roundtripper.go
@@ -177,8 +177,38 @@ type KerberosRoundTripper struct {
 	OriginalRoundTripper http.RoundTripper
 }
 
-type stickyRoundTripper struct {
-	RoundTripper http.RoundTripper
+// stickyRoundTrippers is keyed by owner, as there is one KerberosRoundTripper per ServersTransport and a dedicated
+// round tripper carries the whole transport configuration (TLS client configuration, timeouts, connection pool) of
+// the ServersTransport that created it, hence cannot be shared with another one.
+// The map is lazily allocated, as most client connections never reach a Kerberos or NTLM backend.
+type stickyRoundTrippers struct {
+	mu            sync.Mutex
+	roundTrippers map[*KerberosRoundTripper]http.RoundTripper
+}
+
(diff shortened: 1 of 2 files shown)


juliens (approved):

pkg/server/service/roundtripper.go
```diff
@@ -183,14 +186,14 @@ type KerberosRoundTripper struct {
 // The map is lazily allocated, as most client connections never reach a Kerberos or NTLM backend.
 type stickyRoundTrippers struct {
 	mu            sync.Mutex
-	roundTrippers map[*KerberosRoundTripper]http.RoundTripper
+	roundTrippers map[string]http.RoundTripper
```
sdelicata: @juliens This looks like my first implementation, but I changed my mind because keying on the config name loses the invalidation we were getting for free from the instance.
`Update()` builds a new `KerberosRoundTripper` as soon as a ServersTransport config changes. With the pointer as key, a client connection that already stuck one simply doesn't find it anymore and re-sticks with the new config. With the name, it keeps serving the stale clone until the client disconnects, so the new rootCAs/serverName/certificates just don't apply.

juliens: Ok, I remove this part!

kevinpollet (approved):

PR #13849, opened 2026-09-07
Use `EC256` as default key type for certificates requested via ACME providers

jnoordsij:
### What does this PR do?

Switch to using `EC256` as default key type for certificates requested via ACME providers, as opposed to the current default of `RSA4096`. 

### Motivation

Improve performance and reduce network traffic for default setups; with effectively little-to-none impact on compatibility and security.

In the current state of the web, practically any viable client should be compatible with this setup. Moreover, due to the considerably smaller key sizes and operation timings, certificates with ECDSA keys generally seem to outperform RSA certificates, hence using them by default makes a lot of sense.

#### References

- Let's Encrypt switching defaults on certbot mid-2022: https://community.letsencrypt.org/t/ecdsa-certificates-by-default-and-other-upcoming-changes-in-certbot-2-0/177013
- Server-Side TLS recommendations by TLSRef (formerly Mozilla Server Side TLS – Recommended Configurations): https://docs.tlsref.org/server-side-tls.html

### More

- [ ] Added/updated tests

### Additional Notes

Before this lands, it should be considered:

- Whether or not this type of change is suitable for a minor version release (I'd argue it is, given historic precedent on changes with similar or more impact landing on minor or patch releases)
- Determine whether or not the reduction in security strength is relevant (I'd argue the new default is reasonably secure for default setups based on references above; those with stricter requirements have likely already ensured their own value instead of relying on defaults)
- What kind of release notes and details should be added to help users in their migration


integration/acme_test.go
jnoordsij, line 301: FWIW I just figured this is likely because the related conf-file is configured to use a fixture certificate file, rather than a resolved one, which is also why it remains using RSA instead of ECDSA here.

Refs: 
- https://github.com/traefik/traefik/blob/d48621ce0b6fd221b20bdb6c22e652e7498e5db6/integration/fixtures/acme/acme_tls.toml#L53-L56
- https://github.com/traefik/traefik/blob/d48621ce0b6fd221b20bdb6c22e652e7498e5db6/integration/fixtures/acme/ssl/wildcard.key (https://github.com/traefik/traefik/pull/1772)

jnoordsij:
Cross-ref to ~~#13910~~ #13004: this change will, as far as I can gather, currently only impact newly request certificates, given that old ones will reuse their existing key (and thus type). As there's currently no way for users to roll this over in a way, ~~#13910~~ #13004 is required to also offer changing the key of existing certificates (at the next renewal time).

PR #13850, opened 2026-09-07
Remove the remaining lego v4 dependency

amazon7737:
### What does this PR do?

Use lego v5 in the Ingress NGINX provider tests and remove the remaining lego v4 dependency with `go mod tidy`.

### Motivation

The tests still import lego v4 only for `tlsalpn01.ACMETLS1Protocol`, while the rest of Traefik uses lego v5. The constant has the same value in both versions, so this dependency is no longer needed.

### More

- [x] Added/updated tests

### Additional Notes

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -125,8 +125,6 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
-require github.com/go-acme/lego/v4 v4.35.2
-
 require (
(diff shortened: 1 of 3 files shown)


kevinpollet (approved):
Thanks 👍

rtribotte (approved):
Thanks!

PR #13862, opened 2026-09-08
Add consistent Dockerfile syntax versioning

jnoordsij:
### What does this PR do?

Ensure Dockerfiles always use the latest major-compatible syntax version on build.

### Motivation

To be up-to-date instead of using a [very outdated version](https://github.com/moby/buildkit/releases/tag/dockerfile%2F1.2.1) when building Docker images. This should help in prevention of potential future issues. Spotted when looking through some CI logs, where I noticed this (very) old Dockerfile syntax version being pulled by BuildKit.

This makes us comply with [upstream recommendation by Docker](https://docs.docker.com/build/buildkit/frontend/#stable-channel):
> We recommend using `docker/dockerfile:1`, which always points to the latest stable release of the version 1 syntax, and receives both "minor" and "patch" updates for the version 1 release cycle. BuildKit automatically checks for updates of the syntax when performing a build, making sure you are using the most current version.

### More

- [ ] Added/updated tests

### Additional Notes

Out of an abundance of caution, I've now targeted `master`, although I can't see any reason this would cause (relevant/noticeable) behavioral changes of any kind. As such, I'm fine with rebasing to `v2.11` or `v3.7` if more appropriate.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-# syntax=docker/dockerfile:1.2
+# syntax=docker/dockerfile:1
 FROM alpine:3.24
 
 RUN apk add --no-cache --no-progress ca-certificates tzdata
--- a/docs/check.Dockerfile
+++ b/docs/check.Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
 FROM alpine:3.24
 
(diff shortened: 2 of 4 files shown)


rtribotte (approved):
LGTM

kevinpollet (approved):

PR #13892, opened 2026-09-12
Return a deterministic 504 when the request read deadline expires

### LINKED ISSUE #12634: Getting HTTP 504 Gateway Timeout when `readTimeout` is set
### Welcome!

### What did you do?

I've setup the latest version of Traefik with minimal configuration in a Compose file to act as a proxy between a client and my back-end server.

Then, I've configured a `respondingTimeouts.readTimeout` to be 2 seconds, and made some requests with a large file throttling the upload speed so that the timeout is reached.

### What did you see instead?

I've got inconsistent results: sometimes `HTTP 504 Gateway Timeout` is returned, the other times `HTTP 499 Client Closed Request`.

```
Version:      3.6.7
Codename:     ramequin
Go version:   go1.24.11
Built:        2026-01-14T14:04:03Z
OS/Arch:      linux/amd64
```

Compose file:

```yaml
services:
  traefik:
    image: traefik:v3.6.7
    command:
      - --api.insecure=true
      - --log.level=DEBUG
      - --accesslog=true

      - --providers.docker=true
      - --providers.docker.exposedbydefault=false

      - --entryPoints.web.address=:80

      - --entryPoints.web.transport.respondingTimeouts.readTimeout=2s
    ports:
      - "8080:80"
    volumes:

  upload:
    build: ./upload
    labels:
      - traefik.enable=true
      - traefik.http.routers.upload.rule=PathPrefix(`/upload`)
      - traefik.http.routers.upload.entrypoints=web
      - traefik.http.services.upload.loadbalancer.server.port=8000
```

Application source code:

```python
from fastapi import FastAPI, UploadFile, File
from fastapi.responses import JSONResponse

app = FastAPI()

@app.post("/upload")
async def upload(file: UploadFile = File(...)):
    total = 0
    while True:
        chunk = await file.read(1024 * 1024)  # 1 MiB
        if not chunk:
            break
        total += len(chunk)
    return JSONResponse({"filename": file.filename, "bytes": total})
```

Dockerfile (to replicate the issue faster):

```
FROM python:3.12-slim

WORKDIR /app
RUN pip install --no-cache-dir fastapi uvicorn python-multipart

COPY app.py /app/app.py

EXPOSE 8000
CMD ["uvicorn", "app:app", "--host=0.0.0.0", "--port=8000", "--log-level=info"]
```

The throttled curl command that produces an error:

> Create a file with the following command: `dd if=/dev/zero of=big.bin bs=1M count=50`

```sh
curl -v --limit-rate 200k -F "file=@big.bin" http://localhost:8080/upload
```

The curl command that works (no throttling):

```sh
curl -v -F "file=@big.bin" http://localhost:8080/upload
```

```
traefik-1  | 2026-02-06T18:29:07Z DBG github.com/traefik/traefik/v3/pkg/server/service/loadbalancer/wrr/wrr.go:176 > Service selected by WRR: http://172.18.0.2:8000
traefik-1  | 2026-02-06T18:29:09Z DBG github.com/traefik/traefik/v3/pkg/proxy/httputil/proxy.go:121 > 504 Gateway Timeout error="readfrom tcp 172.18.0.3:35172->172.18.0.2:8000: read tcp 172.18.0.3:80->172.18.0.1:42564: i/o timeout"
traefik-1  | 172.18.0.1 - - [06/Feb/2026:18:29:07 +0000] "POST /upload HTTP/1.1" 504 15 "-" "-" 2 "upload@docker" "http://172.18.0.2:8000" 2000ms
traefik-1  | 2026-02-06T18:29:10Z DBG github.com/traefik/traefik/v3/pkg/server/service/loadbalancer/wrr/wrr.go:176 > Service selected by WRR: http://172.18.0.2:8000
traefik-1  | 2026-02-06T18:29:12Z DBG github.com/traefik/traefik/v3/pkg/proxy/httputil/proxy.go:121 > 499 Client Closed Request error="context canceled"
traefik-1  | 172.18.0.1 - - [06/Feb/2026:18:29:10 +0000] "POST /upload HTTP/1.1" 499 21 "-" "-" 3 "upload@docker" "http://172.18.0.2:8000" 2000ms
```

abrahamgracef:
Fixes #12634

### What does this PR do?

Makes the status returned when the entrypoint `respondingTimeouts.readTimeout` fires while a request body is still being transferred deterministic.

Today the client can get a `504 Gateway Timeout` **or** a `499 Client Closed Request` for the exact same slow upload, depending on a race between two cancellation paths inside Go's `net/http`:

- the connection reader surfaces the raw `i/o timeout` error, which maps to `504`, or
- the canceled request context surfaces `context.Canceled`, which currently maps to `499`.

This PR stamps the read-timeout deadline into the request context when the entrypoint receives the request, and makes `ComputeStatusCode` reclassify a `context.Canceled` error as a `504 Gateway Timeout` when that deadline has already passed. A genuine client disconnect (deadline not expired) still maps to `499 Client Closed Request`.

### Motivation

The issue reports inconsistent status codes for the same scenario, which is confusing and misleading: a `499` makes it look like the client hung up when Traefik actually timed out. This was reproducible from master with a throttled upload and a 2s `readTimeout` (1x `504` / 4x `499` in five runs).

### More

- [x] Added/updated tests

### Additional Notes

- `ComputeStatusCode` is exported; its signature changed from `ComputeStatusCode(err error)` to `ComputeStatusCode(ctx context.Context, err error)` to know whether the read deadline expired.
- Verified locally:
  - unit test `TestComputeStatusCode` covers the new deadline-based reclassification;
  - `TestReadTimeoutDuringBodyForwarding` (entrypoint + reverse proxy, trickled body past a 300ms `readTimeout`) returns `504` with the fix, and returned `499` 20 times out of 20 against the unpatched code;
  - live reproduction against master with a throttled 50MB upload: 6/6 deterministic `504` (was mixed `504`/`499`);
  - `golangci-lint` v2.13.2 on the touched packages: 0 issues.

### DIFF
--- a/pkg/proxy/httputil/proxy.go
+++ b/pkg/proxy/httputil/proxy.go
@@ -29,7 +29,23 @@ const (
 	StatusClientClosedRequestText = "Client Closed Request"
 
 	notAppendXFFKey key = "NotAppendXFF"
+	// readDeadlineKey is the context key used to store the read deadline
+	// enforced by the entrypoint's respondingTimeouts.readTimeout for the request.
+	readDeadlineKey key = "ReadDeadline"
 )
+
+// WithReadDeadline stores the read deadline enforced by the entrypoint's
+// respondingTimeouts.readTimeout for the request into the context.
+func WithReadDeadline(ctx context.Context, deadline time.Time) context.Context {
+	return context.WithValue(ctx, readDeadlineKey, deadline)
+}
+
+// ReadDeadlineFromContext returns the read deadline stored in the context by
+// WithReadDeadline, and whether it was present.
+func ReadDeadlineFromContext(ctx context.Context) (time.Time, bool) {
+	deadline, ok := ctx.Value(readDeadlineKey).(time.Time)
+	return deadline, ok
+}
 
 // SetNotAppendXFF indicates xff should not be appended.
 func SetNotAppendXFF(ctx context.Context) context.Context {
@@ -194,7 +210,7 @@ func ErrorHandler(w http.ResponseWriter, req *http.Request, err error) {
 
 // ErrorHandlerWithContext is the http.Handler called when something goes wrong when forwarding the request.
 func ErrorHandlerWithContext(ctx context.Context, w http.ResponseWriter, err error) {
-	statusCode := ComputeStatusCode(err)
+	statusCode := ComputeStatusCode(ctx, err)
 
 	logger := log.Ctx(ctx)
 
@@ -234,11 +250,23 @@ func isTLSConfigError(err error) bool {
 }
 
 // ComputeStatusCode computes the HTTP status code according to the given error.
-func ComputeStatusCode(err error) int {
+//
+// A context.Canceled error is ambiguous: both a client disconnecting and the
+// entrypoint read deadline (respondingTimeouts.readTimeout) firing cancel the
+// request context. That race made the status returned for a gateway timeout
+// depend on which cancellation surfaced first (499 or 504). When the read
+// deadline has already passed by the time the request is handled, the timeout
+// is the cause, so a 504 Gateway Timeout is returned regardless of the error
+// that surfaced.
+func ComputeStatusCode(ctx context.Context, err error) int {
 	switch {
 	case errors.Is(err, io.EOF):
 		return http.StatusBadGateway
 	case errors.Is(err, context.Canceled):
+		if deadline, ok := ReadDeadlineFromContext(ctx); ok && !time.Now().Before(deadline) {
+			return http.StatusGatewayTimeout
+		}
+
 		return StatusClientClosedRequest
 	default:
 		if netErr, ok := errors.AsType[net.Error](err); ok {
--- a/pkg/proxy/httputil/observability.go
+++ b/pkg/proxy/httputil/observability.go
@@ -49,7 +49,7 @@ func (t *wrapper) RoundTrip(req *http.Request) (*http.Response, error) {
 	var headers http.Header
 	response, err := t.rt.RoundTrip(req)
 	if err != nil {
-		statusCode = ComputeStatusCode(err)
+		statusCode = ComputeStatusCode(req.Context(), err)
 	}
 	if response != nil {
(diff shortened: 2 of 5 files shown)


pkg/proxy/httputil/proxy.go
```diff
@@ -234,11 +250,23 @@ func isTLSConfigError(err error) bool {
...
+// depend on which cancellation surfaced first (499 or 504). When the read
+// deadline has already passed by the time the request is handled, the timeout
+// is the cause, so a 504 Gateway Timeout is returned regardless of the error
+// that surfaced.
+func ComputeStatusCode(ctx context.Context, err error) int {
 	switch {
 	case errors.Is(err, io.EOF):
 		return http.StatusBadGateway
 	case errors.Is(err, context.Canceled):
+		if deadline, ok := ReadDeadlineFromContext(ctx); ok && !time.Now().Before(deadline) {
```
zerkms, line 266: `time.Now().Before(deadline)` would create another race condition: if a client disconnects one nanosecond before timeout, and if it takes more than 1ns to run to this line - then it will be returned as `http.StatusGatewayTimeout`.

abrahamgracef (commented):
Thanks for the review, @zerkms! You are absolutely right—there is a microscopic race window there. 

If a client disconnects 1ns before the deadline, and the scheduler takes 2ns to wake up the goroutine and evaluate `time.Now()`, it will incorrectly log a `504` instead of a `499`.

Unfortunately, because Go's `net/http` server simply cancels the request context when the connection drops (regardless of whether it dropped due to a `ReadTimeout` expiry or a genuine TCP FIN from the client), we don't get a specific `net.Error` timeout surfaced to the middleware—we only get a generic `context.Canceled`. 

Because of this, comparing the deadline against `time.Now()` is currently the only heuristic I could find to differentiate between a client disconnect and a server timeout. 

I'll defer to the core maintainers on whether this heuristic is acceptable given the microscopic race window, or if there is a better approach (perhaps utilizing `http.ResponseController` deeper in the stack)! Thanks again for catching that.

PR #13902, opened 2026-09-14
Isolate preauthenticated NTLM and Kerberos backend connections

kevinpollet:
### What does this PR do?

Dispatches requests carrying an NTLM or Kerberos (Negotiate) credential on a dedicated, connection-scoped transport before they reach the shared backend connection pool.

### Motivation

NTLM and Kerberos are connection-bound authentication schemes, so a credential-bearing request must not be served by the globally shared backend pool.

### More

- [x] Added/updated tests

### Additional Notes

Assisted-by: Claude Opus


rtribotte (approved):
LGTM

PR #13903, opened 2026-09-14
Fix Docker provider reconnect after event refresh errors

herefindalex:
### What does this PR do?

Propagates container-list errors encountered while handling Docker `start`, `die`, and health-status events.

Returning the error exits the current watch operation and allows the existing retry/backoff mechanism to recreate the Docker client and event subscription.

A regression test simulates a transient Docker API failure during an event-triggered refresh and verifies that the provider reconnects.

### Motivation

Previously, an event-triggered container-list failure was logged but otherwise ignored. As a result, the provider did not enter its existing retry and reconnection path, potentially leaving its configuration stale until another matching event triggered a successful refresh.

### More

- [x] Added/updated tests

### Additional Notes

Validation performed:

- `go test -race ./pkg/provider/docker -count=1`
- Full repository lint passed
- `go vet ./pkg/provider/docker`


rtribotte:
Hey @herefindalex,

Thanks for this contribution!

As this is a bug fix, could you please rebase the changes on branch v3.7?

PR #13910, opened 2026-09-15
Add renewKey option to ACME provider configuration

jnoordsij:
### What does this PR do?

Add an option to trigger (private) key renewal upon certificate renewal.

### Motivation

Figuring out this was not an option yet while looking into the code in the scope of another PR,
then subsequently realizing this option is available on various other ACME related applications.

In particular it is relevant/related to https://github.com/traefik/traefik/pull/13849,
as currently changing the new default key type will only apply to new certificates,
with no way to automatically roll over existing ones upon renewal.

### More

- [ ] Added/updated tests

### Additional Notes

There's some implementation choices here:
- Currently this is an 'enable' flag with default `false` to not impact existing flows without user intervention (non-breaking); however arguably the reasonable more secure default would be `true`. I'd suggest starting with this feature disabled by default is best, then waiting at least a few minor versions or potentially even a new major version before toggling it on by default.
- As such, it might as well already be designed as a disable flag (or e.g. `ReuseKey`), with a default of `false` being applied, where a future Traefik version will remove this backwards compatible default.


jnoordsij:
Ideally this would add some kind of (integration) test, but I could not find any tests related to renewal yet.

ldez:
FYI #13004

jnoordsij:
Ah thanks for the heads-up! Totally forgot to search for this first, just started straight into it coming from #13849. I'll closed this in favor of your attempt.

Issue #12569, opened 2026-01-19
Using Kubernetes Gateway API with Cert Manager continuously reloads controller


asardaes:
### Welcome!

- [x] Yes, I've searched similar issues on [GitHub](https://github.com/traefik/traefik/issues) and didn't find any.
- [x] Yes, I've searched similar issues on the [Traefik community forum](https://community.traefik.io) and didn't find any.

### What did you do?

I have 2 Traefik controllers in Kubernetes, same Helm chart but different values. In both of them I enabled the Kubernetes Gateway API and configured a TLS secret that is created and managed by Cert Manager.

### What did you see instead?

Based on the logs, it seems to me that each Traefik controller receives updates for both TLS secrets and reloads everything periodically. See the attached logs below.

### What version of Traefik are you using?

v3.6.7

### What is your environment & configuration?

Kubernetes with Helm chart v38.0.2

```yaml
providers:
  kubernetesCRD:
    enabled: true
    allowCrossNamespace: true
    ingressClass: mbcon
    labelSelector: app.kubernetes.io/instance=mbcon-ingress
  kubernetesIngress:
    enabled: false
    allowEmptyServices: false
    ingressClass: mbcon
  kubernetesGateway:
    enabled: true
gatewayClass:
  enabled: true
  name: mbcon
gateway:
  enabled: true
  name: mbcon
  listeners:
    web:
      port: 8000
      protocol: HTTP
      hostname: svc.domain.com
      namespacePolicy:
        from: Same
    websecure:
      port: 8443
      protocol: HTTPS
      mode: Terminate
      hostname: svc.domain.com
      namespacePolicy:
        from: All
      certificateRefs:
        - kind: Secret
          name: mbcon-cert-secret
          group: ""
```

### If applicable, please paste the log output in DEBUG level

[traefik.log](https://github.com/user-attachments/files/24717692/traefik.log)

nmengin:
Hello @asardaes,

Thank you for reaching out.

After an initial review with the maintainer team, we’re not entirely sure we’re interpreting the issue correctly. Could you clarify whether you’re experiencing an actual problem with how Traefik is behaving in your setup?

It’s expected that both Traefik instances receive an event whenever a secret is created or updated. However, if that change does not result in a modification of their routing configuration, no routing update is applied, and there should be no impact on traffic.

[...]

PR #13911, opened 2026-09-15
Stable sort Gateways to keep the TLS certificate order stable

rl-io:
### What does this PR do?

Sorts the filtered Gateways by namespace and name before building the dynamic
configuration, so the TLS certificates appended per Gateway always come out in
the same order.

### Motivation

`ListGateways()` ranges over `c.factoriesGateway`, so with more than one Gateway
the combined `conf.TLS.Certificates` slice comes back shuffled on most rebuilds.
`getTLSConfig` only sorts within a single Gateway. Both dedup checks are order
sensitive on the slice (`hashstructure.Hash` in the provider, `reflect.DeepEqual`
in `ConfigurationWatcher`), so an unrelated event compares as changed and
triggers a full reload. Related to #12569.

### Measured effect

Tested on a staging cluster with ~20 Gateways sharing a wildcard certificate,
2 Traefik pods, all three Kubernetes providers enabled. Both images built from
614f87e, one with this commit and one without, deployed to the same cluster with
no other changes:

| build | config reloads |
|---|---|
| 614f87e | ~0.67/sec per pod |
| 614f87e + this commit | ~0.11/sec per pod |

Routing was idle for both runs, so nearly all of the reloads on the unpatched
build were the shuffled certificate order comparing as a change. v3.7.1, v3.7.5,
v3.7.6 and v3.7.13 all measured at the same ~0.67/sec, so this is not something
the recent EndpointSlice or route ordering fixes address.

### More

- [x] Added/updated tests

### Additional Notes

The new test builds the configuration ten times and asserts the certificate order
is identical each time. It fails consistently without the sort. No existing test
asserts more than one certificate, which is why the randomisation has never shown
up in CI.


kevinpollet:
Hey @rl-io,

Thanks for this contribution!

rtribotte (approved):
Thanks!

kevinpollet (approved):
Thanks 👍

PR #13915, opened 2026-09-16
Build the ssl-passthrough HTTP route like any other ingress

rtribotte:
### What does this PR do?

Makes an ssl-passthrough Ingress go through the same routing path as any other Ingress, its host keeping the TCP router forwarding the connection to the backend without decrypting it.

### Motivation

Aligning with ingress-nginx, which serves that host on the HTTP port from the regular location, with the annotation directives applied.

### More

- [x] Added/updated tests

### Additional Notes

The routers serving the HTTP part are renamed, hence the migration guide entry.

Assisted-by: Claude Opus


traefiker:
:no_entry_sign: error related to review: need more review [0/2]

kevinpollet (approved):

PR #13916, opened 2026-09-16
Add denyOnError option to the Redis rate limiter

### LINKED ISSUE #12043: Add health track to ratelimit middleware
### Welcome!

In v3.4 a redis backend was introduced for the ratelimit middleware.

This introduces a hard dependency in the critical path: if redis is down or has a transient failure, the whole middleware goes down.

This is not a problem with the in memory backend because the chance of failure when updating the bucket is zero. 

But not with redis where you have a remote backend and many things can glitch (network, the backend itself, etc.)

deyanp:
### What does this PR do?

Adds a `denyOnError` option to the Redis rate limiter. It sets what happens when Redis cannot be reached or returns an error:

- `true` (default): the request is rejected with a 500, as today.
- `false`: the error is logged and the request goes through without rate limiting.

This picks up #13529 by @ayoubbts. The first two commits are theirs, unchanged. The third fixes two cases where the default did not hold, and adds the documentation. @ayoubbts, if you want to finish your PR instead, I am happy to close this one.

Fixes #12043.

### Motivation

With a Redis store, any store problem rejects every rate-limited request. When the limit is there to stop abuse rather than to meter usage, a few seconds without rate limiting is better than a few seconds of 500s. We have one rate limit on every route of an API, so for us a store problem is an outage of the whole API.

The third commit fixes two things in the original change:

- **REST provider.** `denyOnError` was a `bool` that `Redis.SetDefaults` sets to `true`. The REST provider decodes with `encoding/json` and never calls `SetDefaults`, so a `redis` block without the option became `false` and the middleware failed open. The same configuration fails closed today. The field is now a `*bool`, like `passHostHeader`, and an unset value means `true`.
- **In-memory limiter.** The option was read as `config.Redis != nil && config.Redis.DenyOnError`, so it was `false` whenever there was no `redis` block. The in-memory limiter cannot return an error from any configuration today, but it now stays fail-closed if that changes.

There are test cases for both, and they fail on the first two commits.

### More

- [x] Added/updated tests

### Additional Notes

This PR and #13823 both change `ratelimiter.New`. Whichever is merged second needs a small rebase.

I ran `golangci-lint` (v2.13.2), `script/validate-misspell.sh`, `go build ./...`, `go vet`, the unit tests for the touched packages, and `make generate-crd`, which leaves the tree clean.


kevinpollet:
Hey @deyanp,

Thanks for this contribution!

deyanp:
I think the three `new` comments are wrong.

`new` can take a value since Go 1.26, and this repo is on Go 1.26. `master` already uses it
in 32 files — `PassHostHeader: new(true)` in `pkg/api/handler_http_test.go:415`, for
example. And the second commit here, from #13529, is the change from a `boolPtr` helper to
`new`.

The branch builds and the tests pass on go1.26.6. I also ran `go vet` and golangci-lint
v2.13.2. The checks list is empty because the workflows still need approval, not because
anything failed.

The third comment does raise something real, though. Assigning the pointer straight
through:

```go
rl.Redis.DenyOnError = rateLimit.Redis.DenyOnError
```

would leave the dynamic config pointing at the same bool as the Kubernetes object from the
informer cache. I copy it so the two stay separate.

Re-adding the review label gives a fresh run, but I can't do that myself. Happy to fix
anything that turns out to be real.

PR #13918, opened 2026-09-16
Feat: Add sampleRate config for otlp access log

### LINKED ISSUE #13885: Supports configuring `sampleRate` to limit the number of reported logs when `--experimental.otlpLogs=true` is enabled
### Welcome!

- Supports configuring `sampleRate` to limit the number of reported logs when `--experimental.otlpLogs=true` is enabled.
- Currently, Traefik supports sending access logs to an OTLP backend, but it does not support limiting the volume of logs sent via a `sampleRate` attribute—unlike the mechanism used for traces. This means that access logs are reported unconditionally for every single request, which is not always the behavior I desire. See https://doc.traefik.io/traefik/reference/install-configuration/observability/logs-and-accesslogs/#configuration-options_3 .

Thanhphan1147:
### What does this PR do?

Adds an `accesslog.sampleRate` option to limit the volume of access logs exported to an OpenTelemetry Collector.

The value is between `0.0` and `1.0` (default `1.0`, meaning every access log is exported). When set below `1.0`, each access log record is randomly kept with a probability equal to `sampleRate` before being handed to the OTLP batch processor.

We implement sampling by wrapping the log processor. We're also using a basic random sampling as I suspect that this should be enough for access logs. A more deterministic sampling can be used instead if necessary. This option only applies to access logs (`accesslog.otlp`).

### Motivation

Fixes #13885.

### More

- [x] Added/updated tests

PR made with AI assistance


kevinpollet:
Hey @Thanhphan1147,

Thanks for this contribution!

